PR #280, opened 2025-08-10
add reference to writing a new plugin

### LINKED ISSUE #260: For the release, integrate "create a new filter" doc into the scheduler architecture doc
The scheduler being configurable is a key 0.2 release outcome and we want to reference it from the readme and blog.  The current [architecture.md](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/docs/architecture.md) doc is in good shape, but we should also describe how to [extend with a new filter](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/docs/create_new_filter.md) inside the doc somewhere.

elevran:
Add a reference to creating a new plugin guide in architecture.md

In addition:
- break long lines and introduce leading space where a continuation is needed.
- fix typos
- add blank line before/after code snippets (e.g., YAML examples)

Fix #260

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -4,7 +4,9 @@
 
 ## Overview
 
-**llm-d** is an extensible architecture designed to route inference requests efficiently across model-serving pods. A central component of this architecture is the **Inference Gateway**, which builds on the Kubernetes-native **Gateway API Inference Extension (GIE)** to enable scalable, flexible, and pluggable routing of requests.
+**llm-d** is an extensible architecture designed to route inference requests efficiently across model-serving pods.
+ A central component of this architecture is the **Inference Gateway**, which builds on the Kubernetes-native
+ **Gateway API Inference Extension** to enable scalable, flexible, and pluggable routing of requests.
 
 The design enables:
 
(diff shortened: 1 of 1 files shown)


docs/architecture.md
```diff
@@ -51,6 +53,8 @@ Routing decisions are governed by dynamic components:
 - **Scrapers**: Collect pod metadata and metrics for scorers
 
 These components are maintained in the `llm-d-inference-scheduler` repository and can evolve independently.
+A [sample filter plugin guide](./create_new_filter.md) is provided to illustrate how one could extend the
+ Inference Gateway functionality to address unique requirements.
```
kfswain, line 57: non-blocking: Do we think its worth mentioning/linking to the other scheduler extension points?

They should all be in a single file here: https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/pkg/epp/scheduling/framework/plugins.go

kfswain (approved):
lgtm, one non blocking comment

shmuelk:
In general oter changes are needed to the architecture.md file. But that isn't connected to the changes in this PR.

nirrozenbaum:
> Since extensibility is a feature of IGW, wouldn't this better be documented in upstream IGW?

@ahg-g yes. going forward, this will move to GIE along with more extensive documentation on the optional extension point.

it's a bit awkward to put currently in GIE how to write a plugin with no explanation about pluggability in general.
we should address the in depth documentation of the extension points. I'll take care of it after this release. 
after that we will move this doc to GIE as well.

PR #289, opened 2025-08-12
Add workflows to automate aged issues management

anoruxylene:
The PR is a result of https://github.com/llm-d/llm-d-inference-scheduler/issues/274.

### Description

Explaining each of the workflows added:
* **stale.yaml** implementing the '**(90d) lifecycle/stale -> (30d) lifecycle/rotten -> (30d) close an issue**' issue flow via [actions/stale](https://github.com/actions/stale). The most practical way to run the workflow is via a cron job scheduled at some time.
* **unstale.yaml** my handwritten workflow created as an add-on to **stale.yaml** to have a real-time reaction to whenever a 'stale' or 'rotten' issue becomes active again. So, it quickly removes the irrelevant label. It's small, focused, and efficient in terms of GitHub API calls. The current definition of 'active' refers to the issue being reopened or any comment being added to it. We can extend it with more GitHub events if you wish (e.g., issue: labeled, assigned, etc). See the [full list](https://docs.github.com/en/webhooks/webhook-events-and-payloads#issues) of issue events. 

### Things to consider
1. **stale.yaml** is configured to run daily at 1:00 UTC. Please, pick the cron value that would work best for you.
2. **operations-per-run** parameter of actions/stale. If I understand it correctly, it's essentially the number of API calls made to GitHub (such as get batch issue, add/remove label, and so on). If it fails to process all issues in one run, it will proceed on the next run. The default value is 30.
3. The workflows mentioned above have been **tested** in my personal repo - https://github.com/anoruxylene/github-issue-managing-test. Feel free to check it out if that makes any sense.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,38 @@
+name: 'Mark stale issues'
+
+on:
+  schedule:
+    - cron: '0 1 * * *'
+
+jobs:
+  stale-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: 'Mark stale issues'
+        uses: actions/stale@v9
+        with:
+          days-before-issue-stale: 90
+          days-before-pr-stale: -1
+          days-before-close: -1
+          stale-issue-label: 'lifecycle/stale'
+          exempt-issue-labels: 'lifecycle/rotten'
+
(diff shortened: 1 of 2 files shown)


elevran (approved):
Thanks @anoruxylene!
We may want to tweak the staleness values later on, but let's run with these defaults for now.

Issue #279, opened 2025-08-10
load aware scorer unit tests should be updated


nirrozenbaum:
The current [load_aware_scorer_test](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/plugins/scorer/load_aware_scorer_test.go) is not testing the unit.
Instead, it uses a scheduler, with the scorer and a picker. This is testing much more than just the scorer, it tests the scheduler behavior and the picker behavior and in case of a bug in any of them, this unit test will fail although it shouldn't.

The load aware unit-test should be re-written to test the single unit. 
A good example can be seen in the [session_affinity_test](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/plugins/scorer/session_affinity_test.go).

PR #291, opened 2025-08-13
update load aware scorer unit tests

jairuigou:
Fix: #279

### DIFF
--- a/pkg/plugins/scorer/load_aware_scorer_test.go
+++ b/pkg/plugins/scorer/load_aware_scorer_test.go
@@ -10,104 +10,60 @@ import (
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/backend"
 	backendmetrics "sigs.k8s.io/gateway-api-inference-extension/pkg/epp/backend/metrics" // Import config for thresholds
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/scheduling/framework"
-	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/scheduling/framework/plugins/picker"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/scheduling/types"
 
 	"github.com/llm-d/llm-d-inference-scheduler/pkg/plugins/scorer"
 )
 
 func TestLoadBasedScorer(t *testing.T) {
+	podA := &types.PodMetrics{
+		Pod: &backend.Pod{NamespacedName: k8stypes.NamespacedName{Name: "pod-a"}},
+		MetricsState: &backendmetrics.MetricsState{
+			WaitingQueueSize: 2,
+		},
+	}
+	podB := &types.PodMetrics{
+		Pod: &backend.Pod{NamespacedName: k8stypes.NamespacedName{Name: "pod-b"}},
+		MetricsState: &backendmetrics.MetricsState{
+			WaitingQueueSize: 0,
+		},
+	}
+	podC := &types.PodMetrics{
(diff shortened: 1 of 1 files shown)


pkg/plugins/scorer/load_aware_scorer_test.go
```diff
@@ -72,42 +70,24 @@ func TestLoadBasedScorer(t *testing.T) {
...
-
-			if test.err != (err != nil) {
-				t.Errorf("Unexpected error, got %v, want %v", err, test.err)
+			wantScores := map[types.Pod]float64{}
+			for _, pod := range test.input {
+				wantScores[pod] = test.want[pod.GetPod().NamespacedName.Name]
 			}
 
-			if diff := cmp.Diff(test.wantRes, got); diff != "" {
+			got := test.scorer.Score(context.TODO(), nil, nil, test.input)
```
nirrozenbaum: can we use context.Background instead of TODO?
```suggestion
			got := test.scorer.Score(context.Background(), nil, nil, test.input)
```

pkg/plugins/scorer/load_aware_scorer_test.go
```diff
@@ -72,42 +70,24 @@ func TestLoadBasedScorer(t *testing.T) {
...
-				WithScorers(framework.NewWeightedScorer(test.scorer, 1)).
-				WithPicker(picker.NewMaxScorePicker(picker.DefaultMaxNumOfEndpoints))
-
-			got, err := schedulerProfile.Run(context.Background(), test.req, nil, test.input)
-
-			if test.err != (err != nil) {
-				t.Errorf("Unexpected error, got %v, want %v", err, test.err)
+			wantScores := map[types.Pod]float64{}
+			for _, pod := range test.input {
+				wantScores[pod] = test.want[pod.GetPod().NamespacedName.Name]
```
nirrozenbaum: why do we need this loop and copy? can we define wantScores directly in the test?
I'm expecting to see at the end of a test a diff between `test.wantScores` and `got`.

pkg/plugins/scorer/load_aware_scorer_test.go
```diff
@@ -62,7 +60,7 @@ func TestLoadBasedScorer(t *testing.T) {
 				&types.PodMetrics{
 					Pod: &backend.Pod{NamespacedName: k8stypes.NamespacedName{Name: "pod3"}},
 					MetricsState: &backendmetrics.MetricsState{
-						WaitingQueueSize:    5,
+						WaitingQueueSize:    15,
```
nirrozenbaum: can we keep only WaitingQueueSize metric and remove the others?
as far as I know, none of the other metrics are used in the load scorer so it will focus the test on the relevant metric.

pkg/plugins/scorer/load_aware_scorer_test.go
jairuigou: Fixed

pkg/plugins/scorer/load_aware_scorer_test.go
jairuigou: Oh right, I revised it again based on session_affinity_test.

pkg/plugins/scorer/load_aware_scorer_test.go
jairuigou: Fixed

nirrozenbaum (approved):
/lgtm
/approve

Thanks @jairuigou, this is great! 🙏🏼

nirrozenbaum:
@jairuigou can you rebase to sync with main so we can merge your PR?

PR #295, opened 2025-08-18
Code coverage report

### LINKED ISSUE #287: Add test coverage to `test-unit` Makefile target
The ability to track unit test coverage.

**Why is this needed**:
Clearer indication of the test coverage status (i.e., not only pass/fail/no test files).

carlory:
Fix #287

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -60,3 +60,18 @@ jobs:
         shell: bash
         run: |
           make build sidecar-build
+
+  code_coverage:
+    name: "Code coverage report"
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    needs: lint-and-test # Depends on the artifact uploaded by the "lint-and-test" job
+    permissions:
+      contents:      read
+      actions:       read  # to download code coverage results from "lint-and-test" job
+      pull-requests: write # write permission needed to comment on PR
+    steps:
+      - uses: fgrosse/go-coverage-report@v1.2.0
+        with:
+          coverage-artifact-name: "code-coverage"
+          coverage-file-name: "coverage.out"


.github/workflows/ci-pr-checks.yaml
```diff
@@ -47,6 +47,16 @@ jobs:
         run: |
           make test
 
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24  # v5.4.3
```
elevran: any reason we can't use a tagged release instead of a SHA?

.github/workflows/ci-pr-checks.yaml
```diff
@@ -47,6 +47,16 @@ jobs:
...
+        with:
+          fail_ci_if_error: true
```
elevran: what's considered an error for coverage?

Makefile
```diff
@@ -63,7 +63,7 @@ test: test-unit
 .PHONY: test-unit
 test-unit: download-tokenizer download-zmq
 	@printf "\033[33;1m==== Running Unit Tests ====\033[0m\n"
-	go test -ldflags="$(LDFLAGS)" -v ./...
+	go test -ldflags="$(LDFLAGS)" -v -covermode=atomic -coverprofile=go-test-coverage.out ./...
```
elevran: if we go with and pay the runtime penalty for `covermode=atomic` due to concurrent code execution, might be worthwhile to also enable `-race` detection in tests.

carlory:
@elevran Sorry for the delay. I forgot to update this PR. I create a demo project in my account and try to add some changes to verify the codecov behavior because I'm not familiar with the codecov behavior.

It seems that the repo/organization owner needs to configure this project in the codecov dashboard before this workflow can work. I'm unsure whether we want to use codecov in our project. If yes, I will update this PR after I check it on my demo project.

carlory:
> Hi, I was wondering about the rationale behind using an external service. What benefits would it bring? Since we already have the coverage file from go test, it seems possible to use tools like go tool cover—so I’m curious about which specific features we’re looking to leverage through CodeDev.

@Jooho FYI:

image

.github/workflows/ci-pr-checks.yaml
carlory: No special reason. I copied it from the project envoyproxy/ai-gateway: https://github.com/envoyproxy/ai-gateway/blob/main/.github/workflows/build_and_test.yaml#L70-L82

Makefile
carlory: will do

.github/workflows/ci-pr-checks.yaml
carlory:
```
On error, exit with non-zero code
```
For example, if there is an issue with uploading the coverage file (e.g., network issues, invalid file paths, or errors with the Codecov service), the CI process will be marked as failed. This is helpful in environments where you want to ensure that coverage reports are always successfully uploaded to Codeco

elevran:
@carlory happy to see you're still interested and will work on the PR.

Regarding the use of an external service such (as [codecov](https://about.codecov.io/)), let me think about this a bit.
The other option is to integrate this via a GitHub action. Seems [go-coverage-report](https://github.com/fgrosse/go-coverage-report) GitHub Action can automatically detect and report coverage changes in Go pull requests. It posts a detailed comment to the PR including which lines and files are newly uncovered in the PR compared to the base branch.
That might be more self contained solution.

Makefile
elevran: the latest commit dropped `covermode` and does not enable `-race`. 
Perhaps make them an optional coverage makefile parameter that can be set (but is empty by default)?

carlory:
@elevran fixed. When I was resolving a conflict, the same result file was uploaded twice.

shmuelk:
@elevran said:

> Regarding the use of an external service such (as [codecov](https://about.codecov.io/)), let me think about this a bit.
The other option is to integrate this via a GitHub action. Seems [go-coverage-report](https://github.com/fgrosse/go-coverage-report) GitHub Action can automatically detect and report coverage changes in Go pull requests. It posts a detailed comment to the PR including which lines and files are newly uncovered in the PR compared to the base branch.
That might be more self contained solution.

Looking at the [go-coverage-report](https://github.com/fgrosse/go-coverage-report) GitHub Action, the documentation there says:

Support for forks is limited since the necessary GITHUB_TOKEN permissions don't allow to post comments to the pull request of the base repository (see https://github.com/fgrosse/go-coverage-report/issues/15). If forks are important for you, this action might not be the best solution.

I think that applies to us.

elevran:
@carlory - yes please, thanks

PR #297, opened 2025-08-18
[feature] Added Active-Request-Scorer

vMaroon:
## Summary

Introduced a new load-based scorer that looks at the active requests being served through the epp instance. Each request is tracked individually with its own TTL to ensure accurate timeout handling. Pods with fewer active requests receive higher scores.

The scorer maintains two data structures for efficiency:
- A cache of individual requests with TTL for automatic cleanup
- A count map for fast O(1) lookup during scoring

Scores are normalized to a range of 0-1, where pods with fewer active requests get higher scores.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -361,10 +361,6 @@ Scores pods based on the number of active requests being served per pod. Each re
 individually with its own TTL to ensure accurate timeout handling. Pods with fewer active 
 requests receive higher scores.
 
-The scorer maintains two data structures for efficiency:
-- A cache of individual requests with TTL for automatic cleanup
-- A count map for fast O(1) lookup during scoring
-
 Scores are normalized to a range of 0-1, where pods with fewer active requests get higher scores.
 
 - **Type**: `active-request-scorer`
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -355,6 +355,25 @@ Pods with requests in the queue will get score between 0.5 and 0.
 
 ---
 
+#### ActiveRequestScorer
+
+Scores pods based on the number of active requests being served per pod. Each request is tracked 
+individually with its own TTL to ensure accurate timeout handling. Pods with fewer active 
+requests receive higher scores.
+
+The scorer maintains two data structures for efficiency:
+- A cache of individual requests with TTL for automatic cleanup
+- A count map for fast O(1) lookup during scoring
+
+Scores are normalized to a range of 0-1, where pods with fewer active requests get higher scores.
+
+- **Type**: `active-request-scorer`
+- **Parameters**:
+  - `requestTimeout`: specifies the timeout for requests in seconds. Once a request is "in-flight" 
+    for this duration, it is considered timed out and automatically removed.
+
+---
+
 #### SessionAffinity
 
 Scores the candidate pods by giving a higher score to the pods that were previously
--- a/pkg/plugins/scorer/active_request.go
+++ b/pkg/plugins/scorer/active_request.go
@@ -21,8 +21,9 @@ const (
 	// ActiveRequestScorerType is the type of the ActiveRequestScorer
 	ActiveRequestScorerType = "active-request-scorer"
 
-	// defaultRequestTimeout defines the default timeout for requests in seconds
-	defaultRequestTimeout = 300
+	// defaultRequestTimeout defines the default timeout for open requests to be
+	// considered stale and removed from the cache.
+	defaultRequestTimeout = 2 * time.Minute
 )
 
 // ActiveRequestScorerParameters defines the parameters for the
@@ -31,7 +32,8 @@ type ActiveRequestScorerParameters struct {
 	// RequestTimeout defines the timeout for requests in seconds.
 	// Once the request is "in-flight" for this duration, it is considered to
 	// be timed out and dropped.
-	RequestTimeout int `json:"requestTimeout"`
+	// This field accepts duration strings like "30s", "1m", "2h".
+	RequestTimeout string `json:"requestTimeout"`
 }
 
 // requestEntry represents a single request in the cache
@@ -50,7 +52,7 @@ var _ framework.Scorer = &ActiveRequestScorer{}
 
 // ActiveRequestScorerFactory defines the factory function for the ActiveRequestScorer.
 func ActiveRequestScorerFactory(name string, rawParameters json.RawMessage, handle plugins.Handle) (plugins.Plugin, error) {
-	parameters := ActiveRequestScorerParameters{RequestTimeout: defaultRequestTimeout}
+	parameters := ActiveRequestScorerParameters{}
 	if rawParameters != nil {
 		if err := json.Unmarshal(rawParameters, &parameters); err != nil {
 			return nil, fmt.Errorf("failed to parse the parameters of the '%s' scorer - %w", ActiveRequestScorerType, err)
@@ -63,15 +65,21 @@ func ActiveRequestScorerFactory(name string, rawParameters json.RawMessage, hand
 // NewActiveRequestScorer creates a new ActiveRequestScorer scorer.
 func NewActiveRequestScorer(ctx context.Context, params *ActiveRequestScorerParameters) *ActiveRequestScorer {
 	requestTimeout := defaultRequestTimeout
+	logger := log.FromContext(ctx)
 
-	if params != nil && params.RequestTimeout > 0 {
-		requestTimeout = params.RequestTimeout
-		log.FromContext(ctx).V(logutil.DEFAULT).Info("Request timeout should be positive, using default request timeout")
+	if params != nil && params.RequestTimeout != "" {
+		paramsRequestTimeout, err := time.ParseDuration(params.RequestTimeout)
+		if err != nil || paramsRequestTimeout <= 0 {
+			logger.Error(err, "Invalid request timeout duration, using default request timeout")
+		} else {
+			requestTimeout = paramsRequestTimeout
+			logger.Info("Using request timeout", "requestTimeout", requestTimeout)
+		}
 	}
 
 	// cache for individual requests with their own TTL
 	requestCache := ttlcache.New[string, *requestEntry](
-		ttlcache.WithTTL[string, *requestEntry](time.Duration(requestTimeout)*time.Second),
+		ttlcache.WithTTL[string, *requestEntry](requestTimeout),
 		ttlcache.WithDisableTouchOnHit[string, *requestEntry](),
 	)
 
@@ -223,8 +231,8 @@ func (s *ActiveRequestScorer) decrementPodCount(podName string) {
 	}
 }
 
-func cleanCachePeriodically(ctx context.Context, cache *ttlcache.Cache[string, *requestEntry], requestTimeout int) {
-	ticker := time.NewTicker(time.Duration(requestTimeout) * time.Second)
+func cleanCachePeriodically(ctx context.Context, cache *ttlcache.Cache[string, *requestEntry], requestTimeout time.Duration) {
+	ticker := time.NewTicker(requestTimeout)
 	defer ticker.Stop()
 
 	for {
--- a/pkg/plugins/scorer/active_request.go
+++ b/pkg/plugins/scorer/active_request.go
@@ -48,27 +48,23 @@ func (r *requestEntry) String() string {
 // compile-time type assertion
 var _ framework.Scorer = &ActiveRequestScorer{}
 
-// ActiveRequestScorerFactory defines the factory function for the LoadAwareScorer
-func ActiveRequestScorerFactory(name string, rawParameters json.RawMessage,
-	handle plugins.Handle) (plugins.Plugin, error) {
+// ActiveRequestScorerFactory defines the factory function for the ActiveRequestScorer.
+func ActiveRequestScorerFactory(name string, rawParameters json.RawMessage, handle plugins.Handle) (plugins.Plugin, error) {
 	parameters := ActiveRequestScorerParameters{RequestTimeout: defaultRequestTimeout}
 	if rawParameters != nil {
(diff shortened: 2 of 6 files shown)


docs/architecture.md
```diff
@@ -355,6 +355,25 @@ Pods with requests in the queue will get score between 0.5 and 0.
...
+#### ActiveRequestScorer
+
+Scores pods based on the number of active requests being served per pod. Each request is tracked 
+individually with its own TTL to ensure accurate timeout handling. Pods with fewer active 
+requests receive higher scores.
+
+The scorer maintains two data structures for efficiency:
+- A cache of individual requests with TTL for automatic cleanup
+- A count map for fast O(1) lookup during scoring
+
```
elevran: nit: this might be TMI for an architecture doc.

pkg/plugins/scorer/active_request.go
```diff
@@ -0,0 +1,236 @@
...
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/scheduling/types"
+	logutil "sigs.k8s.io/gateway-api-inference-extension/pkg/epp/util/logging"
+)
+
+const (
+	// ActiveRequestScorerType is the type of the ActiveRequestScorer
+	ActiveRequestScorerType = "active-request-scorer"
+
+	// defaultRequestTimeout defines the default timeout for requests in seconds
+	defaultRequestTimeout = 300
```
elevran: consider using a time.Duration for readability?
```suggestion
	defaultRequestTimeout = 5 * time.Minute
```

pkg/plugins/scorer/active_request.go
```diff
@@ -0,0 +1,236 @@
...
+func (r *requestEntry) String() string {
+	return fmt.Sprintf("%s.%s", r.PodName, r.RequestID)
+}
+
+// compile-time type assertion
+var _ framework.Scorer = &ActiveRequestScorer{}
+
+// ActiveRequestScorerFactory defines the factory function for the LoadAwareScorer
+func ActiveRequestScorerFactory(name string, rawParameters json.RawMessage,
+	handle plugins.Handle) (plugins.Plugin, error) {
```
elevran: nit: consider line breaks at longer column count. It could make it easier to read and no need to limit to 80 characters wide. Also in L57, L69, etc.

pkg/plugins/scorer/active_request.go
```diff
@@ -0,0 +1,236 @@
...
+		if err := json.Unmarshal(rawParameters, &parameters); err != nil {
+			return nil, fmt.Errorf("failed to parse the parameters of the '%s' scorer - %w",
+				ActiveRequestScorerType, err)
+		}
+	}
+
+	return NewActiveRequestScorer(handle.Context(), parameters.RequestTimeout).WithName(name), nil
+}
+
+// NewActiveRequestScorer creates a new local-serving-queue scorer.
```
elevran: comment refers to an old scorer name?

pkg/plugins/scorer/active_request.go
```diff
@@ -0,0 +1,236 @@
...
+func NewActiveRequestScorer(ctx context.Context, requestTimeout int) *ActiveRequestScorer {
```
elevran: Consider using `*ActiveRequestScorerParameters` as input instead of raw timeout

pkg/plugins/scorer/active_request.go
```diff
@@ -0,0 +1,236 @@
...
+	// cache for individual requests with their own TTL
+	requestCache := ttlcache.New[string, *requestEntry](
+		ttlcache.WithTTL[string, *requestEntry](time.Duration(requestTimeout)*time.Second),
+		ttlcache.WithDisableTouchOnHit[string, *requestEntry](),
+	)
+
+	scorer := &ActiveRequestScorer{
+		typedName:    plugins.TypedName{Type: ActiveRequestScorerType},
+		requestCache: requestCache,
+		podCounts:    make(map[string]int),
```
elevran, line 89: there seems to be a preference in IGW for sync,Map over map+sync.Mutex

pkg/plugins/scorer/active_request.go
```diff
@@ -0,0 +1,236 @@
...
+// per pod.
+type ActiveRequestScorer struct {
+	typedName plugins.TypedName
+
+	// requestCache stores individual request entries with unique composite keys (podName.requestID)
+	requestCache *ttlcache.Cache[string, *requestEntry]
+
+	// podCounts maintains fast lookup for request counts per pod
+	podCounts map[string]int
+	mutex     *sync.RWMutex
```
elevran, line 117: See earlier note regarding use of sync.Map

pkg/plugins/scorer/active_request.go
```diff
@@ -0,0 +1,236 @@
...
+	s.mutex.RUnlock()
+
+	scoredPodsMap := make(map[types.Pod]float64, len(pods))
+	for _, pod := range pods {
+		podName := pod.GetPod().NamespacedName.String()
+		if count, exists := scoredPods[podName]; exists {
+			if count == 0 {
+				scoredPodsMap[pod] = 1.0 // no requests means highest score
+			} else {
+				scoredPodsMap[pod] = float64(sum-count) / float64(sum)
```
elevran: how sensitive is this to the total sum of all pods?
For example, consider a max request count of 8, 100 pods only one has 8 requests and rest have just one each.
Sum = 106 
Pod(8, essentially full): 8/106
Pod(1) = 1/106, almost the same score as 8/106 (**0.01** vs **0.08**, both are very close and that's not what you want, IMHO).

Maybe normalize over max or some other formula?

pkg/plugins/scorer/active_request.go
```diff
@@ -0,0 +1,236 @@
...
+	if _, found := s.requestCache.GetAndDelete(entry.String()); found {
+		s.decrementPodCount(entry.PodName)
+		debugLogger.Info("Removed request from cache", "requestEntry", entry.String())
+	} else {
+		debugLogger.Info("Request not found in cache", "requestEntry", entry.String())
+	}
+}
+
+// incrementPodCount increments the request count for a pod.
+func (s *ActiveRequestScorer) incrementPodCount(podName string) {
```
elevran, line 212: nit: consider using `defer s.mutex.Unlock()` in all functions.

pkg/plugins/scorer/active_request_test.go
```diff
@@ -0,0 +1,303 @@
...
+			name: "some pods in cache",
+			setupCache: func(s *ActiveRequestScorer) {
+				s.mutex.Lock()
+				s.podCounts["default/pod-a"] = 3
+				s.podCounts["default/pod-c"] = 1
+				// pod-b not in cache
+				s.mutex.Unlock()
+			},
+			input: []types.Pod{podA, podB, podC},
+			wantScores: map[types.Pod]float64{
```
elevran, line 80: I don't think the test  needs to confirm the actual scores - if the formula for scores should change, the values won't be correct but the relative order should be maintained.

pkg/plugins/scorer/active_request_test.go
```diff
@@ -0,0 +1,303 @@
...
+	scorer.mutex.RLock()
+	_, exists := scorer.podCounts["default/pod-a"]
+	scorer.mutex.RUnlock()
+	if exists {
+		t.Errorf("Pod should be removed from podCounts when count reaches 0")
+	}
+}
+
+func TestActiveRequestScorer_TTLExpiration(t *testing.T) {
+	ctx := context.Background()
```
elevran, line 228: consider passing in a ticker to control time in tests (e.g., you can advance it as slowly or quickly as needed). Note that it can be tricky with some go routines since the starting time of the routine is not under your control and can vary under loaded test machines.

pkg/plugins/scorer/active_request.go
vMaroon: The problem here is then it would make the config field awkward since either that remains `int` and we do some casting magic to match, or that changes to duration but then its in nanoseconds on config level (which is less readable for the users). Am I missing something?

pkg/plugins/scorer/active_request.go
vMaroon: 80 chars limit is burnt in my brain from other projects - but I don't mind.

pkg/plugins/scorer/active_request.go
vMaroon: > The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.
> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

I think our use does not fit this categorization - what do you think?

pkg/plugins/scorer/active_request.go
vMaroon: Agreed.

pkg/plugins/scorer/active_request_test.go
vMaroon: I think for small scale tests it's ok - is this fine by you?

pkg/plugins/scorer/active_request_test.go
vMaroon: I would rather this is not done in this iteration.

pkg/plugins/scorer/active_request.go
elevran: Not necessarily. There's a `time.ParseDuration` that can be used with a custom (Un)marshaller from string type.

pkg/plugins/scorer/active_request.go
elevran: I think I made the same argument on IGW and ultimately decided it wasn't worth it.
It matters if you intend to contribute to IGW later, but can delay the switch (if at all) to that point in time.

pkg/plugins/scorer/active_request_test.go
elevran: yes - no objections

elevran (dismissed):
Leave to your discretion whether you'd like to add duration parsing or not.
/approve

pkg/plugins/scorer/active_request.go
vMaroon: Switched.

pkg/plugins/scorer/active_request.go
```diff
@@ -0,0 +1,246 @@
...
+	debugLogger := log.FromContext(ctx).V(logutil.DEBUG).WithName("ActiveRequestScorer.PostResponse")
+	if targetPod == nil {
+		debugLogger.Info("Skipping PostResponse because targetPod is nil")
+		return
+	}
+
+	entry := requestEntry{targetPod.NamespacedName.String(), request.RequestId}
+
+	if _, found := s.requestCache.GetAndDelete(entry.String()); found {
+		s.decrementPodCount(entry.PodName)
```
VadimEisenberg, line 204: @vMaroon From reading the code, it seems `PreRequest` adds first target pods of each of _ProfileResults_, for example of both decode and prefill, while `PostResponse` removes only the primary target pod (of the decode). 

The target pod in `PostResponse` is set to be the target pod of `RequestCtx`, which is the first target pod from the primary `ProfileResult`:

https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/8b154baffd35e1c4ad1dcd131f1cbcac04ddc304/pkg/epp/requestcontrol/director.go#L253C2-L263C34

vMaroon: Ahh - apologies, I misunderstood. Then you're right, it does not track that granularity in P/D.

We can definitely bump this in priority if you provide info. For example, given 2-points PostResponse hooks (cc @kfswain), the start can set prefill as done and the end updates decode.

VadimEisenberg: @vMaroon @kfswain maybe `PostResponse` should get target pods of all the profiles, and not only of the primary profile? Or can`PostResponse` receive `SchedulingResult` as a parameter, symmetrically to `PreRequest`?

nirrozenbaum: @VadimEisenberg PostResponse intentionally receives the target pod who actually SERVED the request.
There is another factor that wasn't taken into consideration in this discussion - 
EPP protocol allows to define multiple (prioritized) endpoints as candidates for serving.
more details here:
https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/docs/proposals/004-endpoint-picker-protocol#destination-endpoint

this essentially tells Envoy that if the first endpoint in the list failed in serving the request, envoy should try to next endpoint in the list. 
for all scorers (or other plugins) that maintain a state per request, it is useful to know which endpoint was serving the request and not which one was ranked first (and didn't necessarily served successfully).
This information should be reported by Gateways back to EPP:
https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/docs/proposals/004-endpoint-picker-protocol#destination-endpoint-served

The above is still under development and therefore we currently use picker with configuration of `MaxEndpoints=1`. 

let me know if it makes sense.

VadimEisenberg: @nirrozenbaum I see. My problem currently is that `active-request-scorer` will not work correctly with regard to prefill pods. Prefill pods will be incremented in the `PreRequest`, but will not be decremented in `PostResponse`. They will be removed by TTL eventually.

nirrozenbaum: @VadimEisenberg right. 
I think that until PostResponse issue is fixed, this scorer should either count only decode results, or it cannot be used as is. 
cc: @vMaroon

vMaroon: For PD I think we can do with the start/end PostResponse endpoints and the scorer can use request-id to maintain happy-path association of 1 P and 1 D per request.

Issue #293, opened 2025-08-13
env-dev-kind on Mac M1 (Apple silicone)


davidbreitgand:
**What happened**:

when `make-dev-kind` on Mac M1 with Colima:

<img width="1716" height="584" alt="Image" src="https://github.com/user-attachments/assets/61560dea-4d8e-40d9-adf7-84b40074d9af" />

**What you expected to happen**:

I expected the platform targeted in the Docker build to match the Docker image manifest. As shown below, I inspected the manifest and indeed the platforms do not match. This is not a problem per se, because Colima is a Linux VM and it should run `Linux/arm64` images just fine. I tested that I can run the Docker container of the base image in Colima's Docker context just fine.

The problem is that the Docker image build is attempted for `Darwin/arm64` while the base image is `ubi8`, which is Linux. When Colima is used, Darwin should be ignored, because the real target is Linux. The platform value in the `Makefile` comes from the `Go` environment variables. They indicate `Darwin` because Go aims at running natively on Mac M1. 

When building for Mac M1 native containers, a different base image should be used. However, in the IBM environment, Docker Desktop on Mac is not licensed and using Colima is common. 

**How to reproduce it (as minimally and precisely as possible)**:

`make env-dev-kind` on Mac M1 with Colima.

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): N/A
- llm-d-scheduler version (use `git describe --tags --dirty --always` if you built from source, or specify the tag if you used a tagged version or image): N/A
- Cloud provider or hardware configuration: 

<img width="267" height="475" alt="Image" src="https://github.com/user-attachments/assets/72f69d2d-453d-42b8-b5dc-b7d90fffa6ba" />

- Install tools: N/A
- Others: N/A

** Additional **

Docker image manifest inspect:

<img width="1367" height="996" alt="Image" src="https://github.com/user-attachments/assets/a7fe264d-1e74-472d-a521-1e412656aaf0" />

Base Docker image inspect

<img width="1268" height="1042" alt="Image" src="https://github.com/user-attachments/assets/69211f17-0ee6-43ec-9100-ec0dd4815e14" />

davidbreitgand:
@shmuelk , yes, the platform should not be determined automatically

PR #304, opened 2025-08-19
fix: Makefile fixes for MacOS

shmuelk:
This PR fixes issues with building images and downloading the tokenizer library on MacOS.

In particular:

1. It hard codes the TARGETOS passed into the image builder as linux
2. It overrides the architecture used when downloading the toenizer to be x86_64 and not amd64 on Intel based Macs

**Note:** this uses GNU Makefile extensions. Is this ok?

fix: #293


elevran (approved):
/lgtm

PR #305, opened 2025-08-19
Support for installation of inference-sim with kv-cache enabled

mayabar:
- add example of scheduler and simulator configuration to work with kv-cache
- update kind-dev-env.sh to support model names with '/'
- disable kind environment creation with both PD_ENABLED and KV_CACHE_ENABED true

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -321,12 +321,18 @@ KIND_CLUSTER_NAME ?= llm-d-inference-scheduler-dev
 KIND_GATEWAY_HOST_PORT ?= 30080
 
 .PHONY: env-dev-kind
-env-dev-kind: image-build  ## Run under kind ($(KIND_CLUSTER_NAME))
-	CLUSTER_NAME=$(KIND_CLUSTER_NAME) \
-	GATEWAY_HOST_PORT=$(KIND_GATEWAY_HOST_PORT) \
-	IMAGE_REGISTRY=$(IMAGE_REGISTRY) \
-	EPP_TAG=$(EPP_TAG) \
-		./scripts/kind-dev-env.sh
+env-dev-kind: ## Run under kind ($(KIND_CLUSTER_NAME))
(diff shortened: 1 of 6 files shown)


deploy/components/inference-gateway/deployments.yaml
```diff
@@ -32,11 +32,10 @@ spec:
         - -grpcHealthPort
         - "9003"
         env:
-        - name: PD_ENABLED
-          value: '${PD_ENABLED}'
-        - name: PD_PROMPT_LEN_THRESHOLD
-          value: '${PD_PROMPT_LEN_THRESHOLD}'
+        - name: PYTHONHASHSEED
+          value: "42"
```
kfswain, line 36: Is this a Hitchhikers Guide Reference? 😂 (Just wondering why we chose 42)

mayabar: Copied from kv-cache-manager ;)

Makefile
```diff
@@ -321,12 +321,18 @@ KIND_CLUSTER_NAME ?= llm-d-inference-scheduler-dev
 KIND_GATEWAY_HOST_PORT ?= 30080
 
 .PHONY: env-dev-kind
-env-dev-kind: image-build  ## Run under kind ($(KIND_CLUSTER_NAME))
-	CLUSTER_NAME=$(KIND_CLUSTER_NAME) \
-	GATEWAY_HOST_PORT=$(KIND_GATEWAY_HOST_PORT) \
-	IMAGE_REGISTRY=$(IMAGE_REGISTRY) \
-	EPP_TAG=$(EPP_TAG) \
-		./scripts/kind-dev-env.sh
+env-dev-kind: #image-build  ## Run under kind ($(KIND_CLUSTER_NAME))
```
shmuelk: Please remove the #image-build as it looks at firt like a mistaken commit. Which it's not as you explicitly run that step.

mayabar: you. are right, removing

mayabar: done

shmuelk (approved):
/lgtm

/approve

PR #309, opened 2025-08-21
chore: enable golangci-lint auto fix

yyzxw:


elevran (changes requested):
Should not override the default `lint` target.

PR #310, opened 2025-08-21
chore: Added end to end tests

shmuelk:
This PR adds an initial set of end to end tests for the llm-d-inference-scheduler. The test causes the llm-d-inference-scheduler to be built. The tests are run under a Kubernetes Kind cluster that is setup and torn down as part of the tests. The llm-d-inference-sim is used for the vLLM servers in all of the tests.

These tests include:

1. A simple EPP configuration with a single vLLM simulator.
2. A setup that uses Prefill/Decode disambiguation with one prefill instance and four decode instances.
3. A setup that uses the precise prefix scorer exploiting KV events with a single vLLM simulator.

There is a set of utility functions that possibly can be upstreamed for common use in tests both in the llm-d-inference-scheduler and in the upstream Inference Gateway. This will investigated at a later time.


test/e2e/inference-scheduler/e2e_suite_test.go
elevran, line 1: Q:
- do you expect additional Go packages under `test/e2e` besides `inference-scheduler`?
- what is the reasoning for *not* keeping test related files (e.g. YAMLs) under this directory

test/e2e/inference-scheduler/e2e_suite_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	k8slog "sigs.k8s.io/controller-runtime/pkg/log"
+
+	infextv1a2 "sigs.k8s.io/gateway-api-inference-extension/api/v1alpha2"
```
elevran:
```suggestion
	k8slog "sigs.k8s.io/controller-runtime/pkg/log"
	infextv1a2 "sigs.k8s.io/gateway-api-inference-extension/api/v1alpha2"
```
They should belong in the same import group.

test/e2e/inference-scheduler/e2e_suite_test.go
```diff
@@ -0,0 +1,200 @@
...
+	nsName = "default"
+)
+
+var (
+	ctx       = context.Background()
+	k8sClient client.Client
+	port      string
+	scheme    = runtime.NewScheme()
+
+	vllmSimTag        = getEnvWithDefault("VLLM_SIMULATOR_TAG", "dev")
```
elevran: nit: I think there are IGW functions for interacting with the environment. Up to you if you wan to reuse or not

test/e2e/inference-scheduler/e2e_suite_test.go
```diff
@@ -0,0 +1,200 @@
...
+	applyYAMLFile(serviceAccountManifest)
+	applyYAMLFile(servicesManifest)
+
+	infPoolYaml := readYaml(inferExtManifest)
+	infPoolYaml = substituteMany(infPoolYaml, map[string]string{"${POOL_NAME}": modelName + "-inference-pool"})
+	createObjsFromYaml(infPoolYaml)
+})
+
+var _ = ginkgo.AfterSuite(func() {
+	command := exec.Command("kind", "delete", "cluster", "--name", "e2e-tests")
```
elevran: Q: if destroying and recreating the KIND cluster takes a while, might make sense to run on the same cluster (but with a different namespace per test to ensure we start with a fresh copy)?

test/e2e/inference-scheduler/e2e_suite_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"io"
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
```
elevran: Defer to your preference but request that unless you think it is required, I would prefer not to add/use ginkgo and the related machinery.
- the BDD style feels heavy and our tests are are relatively scoped.
- all the Describe/Context/It/Should seems overly verbose
- requires the ginkgo CLI for the full experience (though you can run tests with `go test`).
- we already use google/cmp and testify/assert elsewhere in the code for matching, not sure how gomega is helping beyond that

test/e2e/inference-scheduler/e2e_suite_test.go
```diff
@@ -0,0 +1,200 @@
...
+	gomega.Eventually(session).WithTimeout(600 * time.Second).Should(gexec.Exit(0))
+
+	command = exec.Command("kind", "--name", "e2e-tests", "load", "docker-image",
+		"ghcr.io/llm-d/llm-d-inference-scheduler:dev")
+	session, err = gexec.Start(command, ginkgo.GinkgoWriter, ginkgo.GinkgoWriter)
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+	gomega.Eventually(session).WithTimeout(600 * time.Second).Should(gexec.Exit(0))
+
+	command = exec.Command("kind", "--name", "e2e-tests", "load", "docker-image",
+		"ghcr.io/llm-d/llm-d-routing-sidecar:v0.2.0")
```
elevran: allow overriding the sidecar image tag via env?

test/e2e/inference-scheduler/e2e_suite_test.go
```diff
@@ -0,0 +1,200 @@
...
+	gomega.Eventually(session).WithTimeout(600 * time.Second).Should(gexec.Exit(0))
+
+	command = exec.Command("kind", "--name", "e2e-tests", "load", "docker-image",
+		"ghcr.io/llm-d/llm-d-inference-sim:"+vllmSimTag)
+	session, err = gexec.Start(command, ginkgo.GinkgoWriter, ginkgo.GinkgoWriter)
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+	gomega.Eventually(session).WithTimeout(600 * time.Second).Should(gexec.Exit(0))
+
+	command = exec.Command("kind", "--name", "e2e-tests", "load", "docker-image",
+		"ghcr.io/llm-d/llm-d-inference-scheduler:dev")
```
elevran: allow setting the image tag or do you always want `dev`?

test/e2e/inference-scheduler/e2e_suite_test.go
```diff
@@ -0,0 +1,200 @@
...
+// createCRDs creates the Inference Extension CRDs used for testing.
+func createCRDs() {
+	crds := runKustomize(gieCrdsKustomize)
+	createObjsFromYaml(crds)
+}
+
+func createEnvoy() {
+	manifests := readYaml(envoyManifest)
+	ginkgo.By("Creating envoy proxy resources from manifest: " + envoyManifest)
+	createObjsFromYaml(manifests)
```
elevran: are there any k8s test helpers that can be leveraged?
Maybe https://github.com/dlespiau/kube-test-harness or github.com/gruntwork-io/terratest or something similar?
For another project we use K8s API by first constructing abstract [resource.Info](http://resource.info/) objects from the YAML files, as in [here](https://github.com/np-guard/netpol-analyzer/blob/main/pkg/manifests/fsscanner/manifests.go) , then parse them into concrete objects as in [here](https://github.com/np-guard/netpol-analyzer/blob/main/pkg/manifests/parser/parser.go).
For reference - your call.

test/e2e/inference-scheduler/e2e_test.go
```diff
@@ -0,0 +1,339 @@
...
+			deleteObjects(modelServers)
+		})
+	})
+
+	ginkgo.When("Running simple non-PD KV enabled configuration", func() {
+		ginkgo.It("should run successfully", func() {
+			epp := createEndPointPicker(kvConfig)
+
+			modelServers := createModelServers(false, true, 1, 0, 0)
+			time.Sleep(5 * time.Second) // wait for model server(s) to become ready
```
elevran: nit: consider using Eventually?

test/e2e/inference-scheduler/e2e_test.go
```diff
@@ -0,0 +1,339 @@
...
+	client := healthPb.NewHealthClient(conn)
+	healthCheckReq := &healthPb.HealthCheckRequest{}
+	checkCount := 20
+	serving := false
+
+	// Wait for EPP to be really ready
+	for checkCount > 0 {
+		time.Sleep(2 * time.Second)
+
+		resp, err := client.Check(ctx, healthCheckReq)
```
elevran:
```suggestion
		resp, err := client.Check(ctx, &healthPb.HealthCheckRequest{})
```

test/e2e/inference-scheduler/utils_test.go
```diff
@@ -0,0 +1,225 @@
...
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	apilabels "k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/gateway-api-inference-extension/api/v1alpha2"
+	testutils "sigs.k8s.io/gateway-api-inference-extension/test/utils"
```
elevran:
```suggestion
	"k8s.io/apimachinery/pkg/types"
	"sigs.k8s.io/controller-runtime/pkg/client"
	"sigs.k8s.io/gateway-api-inference-extension/api/v1alpha2"	
	testutils "sigs.k8s.io/gateway-api-inference-extension/test/utils"
```

test/e2e/inference-scheduler/utils_test.go
elevran, line 1: nit: The `_test.go` extension to a file name usually means it contains `TestXYZ` functions.
Perhaps `utils.go` or `test_utils.go`?

test/e2e/yaml/inference-pools.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: inference.networking.x-k8s.io/v1alpha2
+kind: InferencePool
+metadata:
+  name: ${POOL_NAME}
+spec:
+  targetPortNumber: 8000
+  selector:
+    app: ${POOL_NAME}
+  extensionRef:
+    name: e2e-epp
```
elevran: missing newline at end of file

test/scripts/run_e2e.sh
```diff
@@ -0,0 +1,42 @@
...
+fi
+
+docker pull ghcr.io/llm-d/llm-d-routing-sidecar:${ROUTING_SIDECAR_TAG}
+
+tput bold
+echo "Running end to end tests"
+tput sgr0
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+go test -v ${DIR}/../e2e/inference-scheduler/ -ginkgo.v
```
elevran: missing newline

test/scripts/run_e2e.sh
```diff
@@ -0,0 +1,42 @@
...
+
+EPPTAG=$(docker images | grep ghcr.io/llm-d/llm-d-inference-scheduler | awk '{print $2}' | grep dev)
+if [[ "${EPPTAG}" != "dev" ]]; then
+  tput setaf 1
+  echo "Missing image ghcr.io/llm-d/llm-d-inference-scheduler:dev"
+  tput sgr0
+  exit 1
+fi
+
+docker pull ghcr.io/llm-d/llm-d-routing-sidecar:${ROUTING_SIDECAR_TAG}
```
elevran: consider allowing sim and sidecar image specification (registry and tag) to be defined locally, to support tests where the end-to-end involves changes to sim and/or sidecar as well

elevran (changes requested):

test/e2e/inference-scheduler/e2e_suite_test.go
elevran: side note:
if you do decide to use them, please "dot import" - lines like `gomega.Expect(err).ShouldNot(gomega.HaveOccurred())` throw off the BDD style (i.e. vs. `Expect(err).ShouldNot(HaveOccurred())`)

test/e2e/inference-scheduler/e2e_suite_test.go
shmuelk: All tests in the end to end test suite run under one cluster.

What is created for every test is the deployment of the scheduler and theappropriate vLLM simuators

test/e2e/inference-scheduler/e2e_suite_test.go
shmuelk: The linter fails on dot imports

test/e2e/inference-scheduler/e2e_suite_test.go
elevran: thanks for clarifying.

test/e2e/inference-scheduler/e2e_suite_test.go
shmuelk: IGW uses ginkgo for its end to end tests. Many of the test helper functions that I have used from IGW use ginkgo. I will also point out that the llm-d-inference-scheduler's linter configuration includes one for ginkgo.

elevran: sure - up to you.

test/e2e/inference-scheduler/e2e_suite_test.go
shmuelk: I simply copied the pattern in IGW. I will flatten the package structure.

The source files for the test are now directly under test/e2e and the YAML files which were already under test/e2e, remain there.

test/e2e/inference-scheduler/e2e_suite_test.go
shmuelk: Done

test/e2e/inference-scheduler/e2e_suite_test.go
shmuelk: Done

test/e2e/inference-scheduler/utils_test.go
shmuelk: It needs the _test suffix to be considered the same package as the rest of the test code.

test/e2e/inference-scheduler/utils_test.go
shmuelk: Done

test/e2e/yaml/inference-pools.yaml
shmuelk: Done

test/scripts/run_e2e.sh
shmuelk: The code allows for the specification of the tags of both the sim and the sidecar.

The lines in `run_e2e.sh`:

>> export VLLM_SIMULATOR_TAG="${VLLM_SIMULATOR_TAG:-v0.4.0}"
>> export ROUTING_SIDECAR_TAG="${ROUTING_SIDECAR_TAG:-v0.2.0}"

set defaults if one wasn't specified.

test/scripts/run_e2e.sh
shmuelk: Done

test/scripts/run_e2e.sh
elevran: Some YAML files in this PR have an explicit setting of the tag, not derived from the `_TAG` environment variables.

There's also a registry usage aspect in the comment.
Using a full path (e.g., `ghcr.io/llm-d/...`, including the regisitry) implies that the image should be pulled from a remote container registry. If the image isn’t pushed there, we'll likely get a runtime failure (unless you’ve explicitly overridden the registry behavior, which is neither common nor recommended).
Do/Should we allow use of the e2e scripts and setup to run tests inference-scheduler tests that require a **locally** modified simulator or sidecar?

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,192 @@
...
+	gomega.Eventually(session).WithTimeout(600 * time.Second).Should(gexec.Exit(0))
+
+	command = exec.Command("kind", "--name", "e2e-tests", "load", "docker-image",
+		"ghcr.io/llm-d/llm-d-inference-scheduler:dev")
+	session, err = gexec.Start(command, ginkgo.GinkgoWriter, ginkgo.GinkgoWriter)
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+	gomega.Eventually(session).WithTimeout(600 * time.Second).Should(gexec.Exit(0))
+
+	command = exec.Command("kind", "--name", "e2e-tests", "load", "docker-image",
+		"ghcr.io/llm-d/llm-d-routing-sidecar:v0.2.0")
```
elevran: Please consider using TAG overrides for all images, not just the simulator as in L122, but also for inference-scheduler (L128) and sidecar (L134).
If this does not make sense, feel free to disregard, just please ack since I do not see this addressed from the previous comments.

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,339 @@
...
+	}()
+
+	client := healthPb.NewHealthClient(conn)
+	healthCheckReq := &healthPb.HealthCheckRequest{}
+	checkCount := 20
+	serving := false
+
+	// Wait for EPP to be really ready
+	for checkCount > 0 {
+		time.Sleep(2 * time.Second)
```
elevran: Eventually?

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,339 @@
...
+	conn, err := grpc.NewClient("localhost:30081",
+		grpc.WithTransportCredentials(insecure.NewCredentials()))
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+	defer func() {
+		err := conn.Close()
+		gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+	}()
+
+	client := healthPb.NewHealthClient(conn)
+	healthCheckReq := &healthPb.HealthCheckRequest{}
```
elevran, line 203: any reason not to define/instantiate where used?

test/scripts/run_e2e.sh
shmuelk: All of our images when built locally for testing are tagged `ghcr.io/llm-d/....` usually with the tag dev.

This allows the use of locally built images, I do this all of the time for the simulator

test/e2e/e2e_test.go
shmuelk: I don't see a need to create the handles on every call.

test/e2e/e2e_test.go
shmuelk: I have changed the code to use ginkgo.Eventually

test/e2e/e2e_suite_test.go
shmuelk: Tags can now be specified in environment variables for all of the images.

elevran (approved):

PR #315, opened 2025-08-26
ByLabelSelector filter tests

### LINKED ISSUE #220: Should the prefill and decode instances role selector configurable?
I'm trying to understand better how the pods selection work but I think there is a problem with hardcoded prefill/decode labels

Currently, the selection is based on hardcoded role labels + `InferencePool`'s selector:
```golang
	RoleLabel = "llm-d.ai/role"
	// RolePrefill set for designated prefill workers
	RolePrefill = "prefill"
	// RoleDecode set for designated decode workers
	RoleDecode = "decode"
```

However, prefill and decode can have different topologies.

for example, in the multi node MoE DP topology using LWS, the InferencePool label selector will need to be something like this:

```yaml
leaderworkerset.sigs.k8s.io/worker-index=0
app.kubernetes.io/component=worker
app.kubernetes.io/name=<object-name>
```

so to select the decode instances the full selector would be:
```yaml
leaderworkerset.sigs.k8s.io/worker-index=0
app.kubernetes.io/component=worker
app.kubernetes.io/name=<object-name>
llm-d.ai/role=decode
```

but then prefill can have a different topology that maybe doesn't have 
```yaml
leaderworkerset.sigs.k8s.io/worker-index=0
```
(because it's using just raw deployments and doesn't need to use multi node and LWS)

So to select the prefill instances the full selector would be:
```yaml
app.kubernetes.io/component=worker
app.kubernetes.io/name=<object-name>
llm-d.ai/role=prefill
```

should we decouple prefill and decode topology and make the scheduler configurable on how to select prefill and decode instances?

elevran:
Test code for filtering based on k8s LabelSelector.
Besides tests, there's example code to show selection of Pods using ByLabelSelector filter with configurations to match #220.

Fix #220

### DIFF
--- a/pkg/plugins/filter/by_label_selector_test.go
+++ b/pkg/plugins/filter/by_label_selector_test.go
@@ -1,0 +1,409 @@
+package filter_test
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/backend"
+	backendmetrics "sigs.k8s.io/gateway-api-inference-extension/pkg/epp/backend/metrics"
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/scheduling/types"
+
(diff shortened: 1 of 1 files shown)


elevran:
agree - the sample config would be added in a future filter/scorer guide in IGW

PR #320, opened 2025-08-29
sync with IGW release 1.0.0-rc

nirrozenbaum:
sync with IGW release 1.0.0-rc


cmd/epp/main.go
```diff
@@ -34,15 +34,10 @@ import (
 )
 
 func main() {
-	ctx := ctrl.SetupSignalHandler()
-
-	// Register GIE plugins
-	runner.RegisterAllPlugins()
```
nirrozenbaum, line 40: this function doesn't exist anymore. all built in plugins are automatically registered inside runner (internal function)

PR #321, opened 2025-08-29
fixed a bug where typos check is checking go.mod and go.sum

nirrozenbaum:
This is blocking PR #320 to sync with IGW latest RC.
we should exclude go.sum and go.mod from typos check


kfswain:
/lgtm
/approve

I think llm-d uses prow now so I think this works?

PR #329, opened 2025-09-02
Add configurable prefill port support to prefill-header-handler plugin

pallavijaini0525:
In prefill/decode (PD) disaggregation deployments, the prefill-header-handler plugin using the same targetPort for both the routing sidecar and prefill nodes. This resulted in incorrect x-prefiller-host-port headers being generated when prefill is running on a different port, preventing proper communication between decode and prefill workers.

Added an optional prefillTargetPort parameter to the prefill-header-handler plugin configuration. When specified, this parameter overrides the generic targetPort when constructing the x-prefiller-host-port header.

### DIFF
--- a/pkg/plugins/pre-request/pd_prerequest.go
+++ b/pkg/plugins/pre-request/pd_prerequest.go
@@ -24,7 +24,7 @@ const (
 
 type prefillHeaderHandlerParameters struct {
 	PrefillProfile string `json:"prefillProfile"`
-	PrefillTargetPort int `json:"prefillTargetPort,omitempty"`  // ← Optional field
+	PrefillTargetPort int `json:"prefillTargetPort,omitempty"`  // Optional field
 }
 
 // compile-time type assertion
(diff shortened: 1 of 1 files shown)


elevran:
@pallavijaini0525 the PR is failing CI (linting) - kindly rebase and fix before we can move it forward.

pallavijaini0525:
@elevran  -> Lint issues are fixed. Thanks

pallavijaini0525:
@elevran - I extended this code for two reasons:

1. For the prefill configuration, we can use a different port by updating ms-pd/values.yaml, but the epp/inference scheduler is currently fixed to port 8000, with no option to specify a different port in the epp config.

2. About a month ago, we faced some networking issues, which required setting hostNetwork to true. This caused a conflict, preventing the use of port 8000 for both sidecar and prefill. Since then, we’ve fixed the networking issues with RDMA devices, and there's no longer a need for hostNetwork to be true.

I thought this fix would be useful for overriding the ports, so I left the PR open.

PR #353, opened 2025-09-16
sync gic to latest v1.0.0 release

nirrozenbaum:


pkg/scheduling/pd/scheduler_test.go
```diff
@@ -238,6 +239,7 @@ func TestPDSchedule(t *testing.T) {
 			if test.wantRes2 != nil { // Checking the prefix match in the decode pod.
 				// make sure prefix plugin stores the prefix hit in cache, so we can test it in the following schedule call
 				prefixScorer.PreRequest(ctx, test.req, got, 0)
+				time.Sleep(time.Second)
```
nirrozenbaum, line 242: this is a temp solution until it's solved appropriately in IGW.
PreRequest is adding the information to the cache asynchronously, so we need to wait.
this should be solved in next version and then we should remove this line from our tests.

Issue #347, opened 2025-09-14
Create scale up/down test


elevran:
End to end test to validate that scaling the model server deployment up/down affects scheduling.
- start with 1 replica (EPP configuration should be minimal, if any, so it's easier to rationalize how traffic is split)
- generate traffic going to replica 1
- add second replica, eventually traffic should be split
- scale down back to 1 and ensure no errors as a replica is removed from EPP state

PR #354, opened 2025-09-17
test: Scale up and down the model server during an end to end test

shmuelk:
This PR adds a test to the end to tests.

In particular it adds a test in which:
   1. The "system" is brought up
   2. Inference requests are sent.
   3. The model server deployment is scaled up by one pod
   4. Additional inference requests are made
   5. The test validates that some of the requests go to the added pod.
   6. The model server deployment is scaled down by one
   7. Additional inference requests are made

Fixes: #347

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -179,8 +179,8 @@ var _ = ginkgo.Describe("Run end to end tests", ginkgo.Ordered, func() {
 				gomega.Expect(podHdr).Should(gomega.Equal(scaledDownDecodePods[0]))
 			}
 
-			deleteObjects(epp)
-			deleteObjects(modelServers)
+			testutils.DeleteObjects(testConfig, epp)
+			testutils.DeleteObjects(testConfig, modelServers)
 		})
 	})
 })


test/e2e/e2e_test.go
```diff
@@ -128,6 +128,62 @@ var _ = ginkgo.Describe("Run end to end tests", ginkgo.Ordered, func() {
 			deleteObjects(modelServers)
 		})
 	})
+
+	ginkgo.When("Scaling up and down the model servers", func() {
+		ginkgo.It("work should be distributed across all model servers", func() {
```
elevran: nit: if using `ginkgo.It` the text should probably align with that...
```suggestion
		ginkgo.It("should distribute inference requests across all model servers", func() {
```

test/e2e/e2e_test.go
```diff
@@ -128,6 +128,62 @@ var _ = ginkgo.Describe("Run end to end tests", ginkgo.Ordered, func() {
...
+			modelServers := createModelServers(false, false, 1, 0, 0)
+
+			epp := createEndPointPicker(scaleConfig)
+
+			prefillPods, decodePods := getModelServerPods(podSelector, prefillSelector, decodeSelector)
+			gomega.Expect(prefillPods).Should(gomega.BeEmpty())
```
elevran, line 140: when is the goal of validating that there are no prefill Pods?
Can it somehow fail and if so, how does a failure affect the test if at all?

test/e2e/e2e_test.go
```diff
@@ -128,6 +128,62 @@ var _ = ginkgo.Describe("Run end to end tests", ginkgo.Ordered, func() {
...
+			for range 5 {
+				nsHdr, podHdr = runCompletion(simplePrompt, modelName)
+				gomega.Expect(nsHdr).Should(gomega.Equal(nsName))
+				gomega.Expect(podHdr).Should(gomega.Equal(decodePods[0]))
+			}
+
+			scaleDeployment(modelServers, 1)
+
+			scaledUpPrefillPods, scaledUpDecodePods := getModelServerPods(podSelector, prefillSelector, decodeSelector)
+			gomega.Expect(scaledUpPrefillPods).Should(gomega.BeEmpty())
```
elevran, line 153: ditto

test/e2e/e2e_test.go
```diff
@@ -128,6 +128,62 @@ var _ = ginkgo.Describe("Run end to end tests", ginkgo.Ordered, func() {
...
+			gomega.Expect(scaledUpDecodePods).Should(gomega.HaveLen(2))
+
+			time.Sleep(time.Second)
```
elevran: might it be worthwhile to check that pods are in ready state and not rely on the 1s sleep?
Or maybe there's a different reason for sleeping...?

test/e2e/e2e_test.go
```diff
@@ -128,6 +128,62 @@ var _ = ginkgo.Describe("Run end to end tests", ginkgo.Ordered, func() {
...
+				if scaledPodHdr != podHdr {
+					break
+				}
+			}
+			gomega.Expect(scaledPodHdr).ShouldNot(gomega.Equal(podHdr))
+
+			scaleDeployment(modelServers, -1)
+
+			scaledDownPrefillPods, scaledDownDecodePods := getModelServerPods(podSelector, prefillSelector, decodeSelector)
+			gomega.Expect(scaledDownPrefillPods).Should(gomega.BeEmpty())
```
elevran, line 171: ditto

test/e2e/e2e_test.go
```diff
@@ -128,6 +128,62 @@ var _ = ginkgo.Describe("Run end to end tests", ginkgo.Ordered, func() {
...
+			gomega.Expect(scaledDownDecodePods).Should(gomega.HaveLen(1))
+			gomega.Expect(scaledDownDecodePods[0]).Should(gomega.BeElementOf(scaledUpDecodePods))
+
+			time.Sleep(time.Second)
```
elevran: same. Expect on a condition (with Eventually...) would be quicker.

test/e2e/utils_test.go
```diff
@@ -15,15 +15,49 @@ import (
...
+	for _, kindAndName := range objects {
+		split := strings.Split(kindAndName, "/")
+		if strings.ToLower(split[0]) == deploymentKind {
+			ginkgo.By(fmt.Sprintf("Scaling the deployment %s %s by %d", split[1], direction, absIncrement))
+			scale, err := client.AppsV1().Deployments(nsName).GetScale(ctx, split[1], v1.GetOptions{})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			scale.Spec.Replicas += int32(increment)
+			_, err = client.AppsV1().Deployments(nsName).UpdateScale(ctx, split[1], scale, v1.UpdateOptions{})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
```
elevran, line 46: the scale change can have succeeded but the replica count not have increased or the new pods not ready (or deleted) yet...

test/e2e/e2e_test.go
shmuelk: I'm just checking that it didn't happen. Looking for test bugs

test/e2e/e2e_test.go
shmuelk: Ditto

test/e2e/e2e_test.go
shmuelk: Ditto

test/e2e/utils_test.go
shmuelk: The code in the utility function podsInDeploymentsReady checks that the number of replicas in the spec equals the number of pods in the status, which in turn is checked to be equal to the number of ready pods as noted in the status.

test/e2e/e2e_test.go
shmuelk: The code in the utility function podsInDeploymentsReady checks that the number of replicas in the spec equals the number of pods in the status, which in turn is checked to be equal to the number of ready pods as noted in the status.

I have removed the sleep.

test/e2e/e2e_test.go
shmuelk: The code in the utility function podsInDeploymentsReady checks that the number of replicas in the spec equals the number of pods in the status, which in turn is checked to be equal to the number of ready pods as noted in the status.

I have removed the sleep.

test/e2e/e2e_test.go
shmuelk: Fixed.

PR #357, opened 2025-09-22
feat: Add a scoring plugin to distribute new groups evenly

usize:
The plugin uses a LRU cache to assign higher scores to pods that have less recently handled a request matching no existing prefixes (thus, a new group). This should help even cache growth, as new groups are likely to add more new blocks to the cache than existing groups.

For simplicity, this implementation relies on state produced by the prefix-cache-scorer plugin in order to track whether a request hits or misses the cache. Because of that, prefix-cache-scorer is required for this to work properly in production.

Fixes: llm-d/llm-d-inference-scheduler#346


pkg/plugins/scorer/no_hit_lru.go
```diff
@@ -0,0 +1,274 @@
...
+			prefixPluginName = params.PrefixPluginName
+		}
+		if params.LRUSize > 0 {
+			lruSize = params.LRUSize
+		}
+	}
+
+	lruCache, err := lru.New[string, struct{}](lruSize)
+	if err != nil {
+		panic(fmt.Sprintf("failed to create LRU cache: %v", err))
```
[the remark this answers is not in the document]
vMaroon: `panic` should never be called - an error log is sufficient. Scorers should only attempt to enhance scheduling but not block it on failures (in setup/while running).

pkg/plugins/scorer/no_hit_lru.go
```diff
@@ -0,0 +1,274 @@
...
+// For cache hits, returns neutral scores (0.5) for all pods.
+// For cache misses, ranks pods by their LRU order.
+// - LRU ordering is with respect to when a pod last received a cold request.
+// - Least recently used (or never used) pods get highest score (1.0)
+// - Most recently used pods get lowest score (approaching 0.0)
+func (s *NoHitLRU) Score(ctx context.Context, cycleState *types.CycleState, request *types.LLMRequest, pods []types.Pod) map[types.Pod]float64 {
+	logger := log.FromContext(ctx).V(logutil.DEBUG)
+
+	// Read prefix cache state to determine if this is a cold request
+	prefixState, err := types.ReadCycleStateKey[*prefix.SchedulingContextState](cycleState, plugins.StateKey(s.prefixPluginName))
```
vMaroon, line 126: I think this can be treated as an optimization of accuracy rather than a dependency.
It would be a necessity for the `precise-prefix-cache-scorer` to also implement it IMO.

pkg/plugins/scorer/no_hit_lru.go
```diff
@@ -0,0 +1,263 @@
...
+	return s
+}
+
+// Score scores the given pods based on LRU for cold requests.
+// For cache hits, returns neutral scores (0.5) for all pods.
+// For cache misses, ranks pods by their LRU order.
+// - LRU ordering is with respect to when a pod last received a cold request.
+// - Least recently used (or never used) pods get highest score (1.0)
+// - Most recently used pods get lowest score (approaching 0.0)
+func (s *NoHitLRU) Score(ctx context.Context, cycleState *types.CycleState, request *types.LLMRequest, pods []types.Pod) map[types.Pod]float64 {
```
vMaroon: I think it might help readability if the main logical blocks in this call are separated to different functions.

pkg/plugins/scorer/no_hit_lru.go
```diff
@@ -0,0 +1,263 @@
...
+	logger := log.FromContext(ctx).V(logutil.DEBUG)
+
+	// Read prefix cache state to determine if this is a cold request
+	// This is treated as an optimization - if the state isn't available, we assume cold request
+	prefixState, err := types.ReadCycleStateKey[*prefix.SchedulingContextState](cycleState, plugins.StateKey(s.prefixPluginName))
```
vMaroon, line 126: Not a blocker but would be great: do you think you can extend and support the same optimization for the `precise-prefix-cache-scorer` too? I personally am eager to use this scorer with the latter.

It is also possible that this optimization can be achieved with unequal weighting (e.g., ratio of 2:1 between prefix-cache scorers and this one) - but could also be fragile in some cases.

pkg/plugins/scorer/no_hit_lru.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+}
+
+// NoHitLRU scorer that favors pods that were least recently used for cold requests.
+// This can help evenly distribute cache growth, since cold requests result in more
+// new KV blocks.
+type NoHitLRU struct {
+	typedName        plugins.TypedName
+	lruCache         *lru.Cache[string, struct{}] // pod name -> dummy value (we only care about order)
+	mutex            *sync.RWMutex
```
vMaroon: the `lru.Cache` package is thread-safe, the `Add` and `Keys` calls are protected by an internal lock. The use of this mutex is therefore purely for snapshot semantics: If `Keys()` is called, the mutex makes sure no other request will progress before the scoring. Might be good to test the effect of this locking and profile the overhead as well.

vMaroon (commented):
Overall looks good - a few comments.

pkg/plugins/scorer/no_hit_lru.go
usize: Great idea. Maybe I can add this in a follow up PR? I'm a bit worried about mixing concerns.

pkg/plugins/scorer/no_hit_lru.go
usize: Only with unit tests, but I did do a benchmark and added a test with multiple threads. Results support removing mutex imo. Being stale by 1ms on rare occasions shouldn't significantly break the intent behind this -- cache balancing. Here are benchmark results:

https://gist.github.com/usize/b936f67655029c05ba701218c1bfe5da

pkg/plugins/scorer/no_hit_lru.go
vMaroon: Sounds good.

vMaroon:
@usize looks good overall, thank you for the contribution! I think we're just missing updates to the architecture doc which contains info about plugins and their configuration - then good to do.

docs/architecture.md
```diff
@@ -364,6 +364,30 @@ used for the same session.
...
+The LRU tracking is specific to cold requests only - pods are added to the LRU cache when they serve
+a cold request, not when they serve requests with cache hits.
+
+- **Type**: `no-hit-lru-scorer`
+- **Parameters**:
+  - `prefixPluginName` (optional): The name of the prefix cache plugin to read state from. Defaults to `prefix-cache-scorer`.
+  - `lruSize` (optional): The maximum number of pods to track in the LRU cache. Defaults to 1024.
+
+**Note:** This scorer is designed to work alongside a prefix cache scorer (such as `prefix-cache-scorer` or
+`precise-prefix-cache-scorer`). If no prefix cache state is available, all requests are treated as cold.
```
vMaroon, line 413: I think following with a yaml example of what a configuration looks like is helpful here. Additionally, should state that when integrating with a prefix-cache scorer, the latter should be defined first in the scheduling profile.

PR #378, opened 2025-10-17
Supports the ResponseComplete plugin

learner0810:
fix: https://github.com/llm-d/llm-d-inference-scheduler/issues/375, https://github.com/llm-d/llm-d-inference-scheduler/issues/376


pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -139,3 +145,16 @@ func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, _ *types.CycleStat
...
 	return indexedScoresToNormalizedScoredPods(pods, podToKey, scores)
 }
+
+func getUserInput(request *types.LLMRequest) (string, error) {
+	if request.Body.Completions != nil { // assumed to be valid if not nil
+		return request.Body.Completions.Prompt, nil
+	}
+
+	// must be chat-completions request at this point, return string of entire messages
+	data, err := json.Marshal(request.Body.ChatCompletions.Messages)
```
vMaroon, line 155: This will not work, but it doesn't work right now either. We'll follow-up with a PR to integrate chat-completions support.

vMaroon: cc @guygir

vMaroon (approved):

PR #387, opened 2025-10-24
Initial drop of the activator code

lionelvillard:
This PR introduces a new component, called the activator, scaling from zero and to zero Kubernetes object attached to InferencePool.

Overview
---

The code consists of:
- `charts/activator-filter`: a helm chart for inserting the activator into the gateway HTTP filter chain (Istio only). Needs to be applied only once.
- `charts/activator`:  a helm chart configuring the activator for a given HTTPRoute and InferencePool
- `cmd/activator`: the activator main
- `pkg/activator`: 
  - a ext-proc server handling header requests and ensuring there is at least one replica in the InferencePool.  
  - a monitor scaling down to zero the object attached to the InferencePool after a configurable idleness period.
  
Control-knobs
---

Here are the control-knobs attached to InferencePool objects as annotations.

- `activator.llm-d.ai/target-apiversion`: APIVersion of the target object to scale
- `activator.llm-d.ai/target-kind`: Kind of the target object to scale
- `activator.llm-d.ai/target-name`: Name of the target object to scale
- `activator.llm-d.ai/scale-from-zero-grace-period`: time (in seconds) to wait for scale-from-zero decision to complete before aborting. 
- `activator.llm-d.ai/scale-down-delay`: the minimum of time (in seconds) before the system decides to scale down the last replica to zero

 
Here are the control-knobs specified on the command line:
- `--enable-scale-to-zero`: Enable scaling down InferencePool to zero replicas after a period of idleness (Default: false)
 
Testing
---

- e2e testing is done manually following [these](https://github.com/lionelvillard/llm-d-experiments/tree/main/scale-zero-activator) recipes.
  - recipe 1: multiple models using url-rewriting method
  - recipe 2: multiple models using BBR method
- @dumb0002 recipes 

Future Work
---
- better documentation other than helm charts
- load testing 
- more control-knobs 
 
### Co-authors
- @dumb0002  

@dumb0002 @nilig @elevran @kfswain @shmuelk

### DIFF
--- a/Dockerfile.activator
+++ b/Dockerfile.activator
@@ -1,5 +1,5 @@
-# Build Stage: using Go 1.24.1 mage
-FROM quay.io/projectquay/golang:1.24.1 AS builder
+# Build Stage: using Go 1.24 image
+FROM quay.io/projectquay/golang:1.24 AS builder
 ARG TARGETOS
 ARG TARGETARCH
 
--- a/Dockerfile.epp
+++ b/Dockerfile.epp
@@ -1,6 +1,6 @@
 ## Minimal runtime Dockerfile (microdnf-only, no torch, wrapper in site-packages)
-# Build Stage: using Go 1.24.1 image
-FROM quay.io/projectquay/golang:1.24.1 AS builder
+# Build Stage: using Go 1.24 image
+FROM quay.io/projectquay/golang:1.24 AS builder
 ARG TARGETOS
 ARG TARGETARCH
 ARG KVCACHE_MANAGER_VERSION=v0.3.2


Makefile
```diff
@@ -134,6 +137,14 @@ sidecar-build: check-go ## Build the Sidecar
...
 	go build -o bin/$(SIDECAR_NAME) cmd/$(SIDECAR_NAME)/main.go
 
+##@ Build Activator
+
+.PHONY: activator-build
+activator-build: check-go install-dependencies download-tokenizer ## Build the project
+	@printf "\033[33;1m==== Building ====\033[0m\n"
+	go build -ldflags="$(LDFLAGS)" -o bin/activator cmd/activator/main.go
+
+
```
shmuelk, line 153: See PR #397 and remove the above target after adding the appropriate variables (activator_NAME)

Makefile
```diff
@@ -167,6 +178,22 @@ sidecar-image-push: check-container-tool load-version-json ## Push Sidecar Docke
...
+		--build-arg COMMIT_SHA=${GIT_COMMIT_SHA} \
+		--build-arg BUILD_REF=${BUILD_REF} \
+		 -t $(ACTIVATOR_IMG) \
+		 -f Dockerfile.activator .
+
+.PHONY: activator-image-push
+activator-image-push: check-container-tool load-version-json ## Push Activator Docker image $(ACTIVATOR_IMG) to registry
+	@printf "\033[33;1m==== Pushing Activator Docker image $(ACTIVATOR_IMG) ====\033[0m\n"
+	$(CONTAINER_TOOL) push $(ACTIVATOR_IMG)
+
```
shmuelk, line 182: See PR https://github.com/llm-d/llm-d-inference-scheduler/pull/397 and remove the above targets after adding the appropriate variables (activator_IMAGE)

charts/activator-filter/templates/rbac.yaml
```diff
@@ -0,0 +1,81 @@
...
+  namespace: {{ .Release.Namespace }}
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ .Values.name }}
+  namespace: {{ .Release.Namespace }}
+rules: # TODO: These can probably be trimmed down
+- apiGroups:
+  - "inference.networking.x-k8s.io"
```
shmuelk, line 14: You should support InferencePool v1 (inference.networking.k8s.io/v1)

charts/activator/README.md
```diff
@@ -0,0 +1,44 @@
...
+| `activator.suffix`                         | Suffix to append to the name of the activator deployment and service. Defaults to `-activator`.    |
+| `activator.port`                            | Port serving ext_proc. Defaults to `9004`.  |
+| `activator.healthCheckPort`                 | Port for health checks. Defaults to `9005`. |
+| `activator.image.name`                      | Name of the container image used. |
+| `activator.image.registry`                  | Registry URL and namespace where the image is hosted. |
+| `activator.image.tag`              | Image tag. |
+| `activator.image.pullPolicy`       | Image pull policy for the container. Possible values: `Always`, `IfNotPresent`, or `Never`. Defaults to `Always`. |
+| `inferenceGateway.port`            | The port of the Gateway. Defaults to `80`.    |
+| `inferencePool.name`               | The name of the InferencePool to target.  |
+| `inferencePool.apiVersion`         | The API version of the InferencePool. Defaults to `inference.networking.x-k8s.io`.  |
```
shmuelk: I think InferencePool API version should default to inference.networking.k8s.io

Dockerfile.activator
```diff
@@ -0,0 +1,28 @@
...
+ARG COMMIT_SHA=unknown
+ARG BUILD_REF
+ARG TARGETOS=linux
+ARG TARGETARCH=amd64
+ENV CGO_ENABLED=0
+ENV GOOS=$TARGETOS
+ENV GOARCH=$TARGETARCH
+
+# Dependencies
+WORKDIR /src
```
shmuelk: Change this line to:
`WORKDIR /workspace`

Dockerfile.activator
```diff
@@ -0,0 +1,28 @@
...
+COPY go.mod go.sum ./
+RUN go mod download
+
+# Sources
+COPY cmd/activator ./cmd/activator
+COPY pkg/activator ./pkg/activator
+WORKDIR /src/cmd/activator
```
shmuelk: Remove this line

Dockerfile.activator
```diff
@@ -0,0 +1,28 @@
...
+RUN go build -ldflags="-X sigs.k8s.io/gateway-api-inference-extension/version.CommitSHA=${COMMIT_SHA} -X sigs.k8s.io/gateway-api-inference-extension/version.BuildRef=${BUILD_REF}" -o /activator
```
shmuelk: Please make this Dockerfile more like the others in this repo:

Change:  `-o /activator` to `-o bin/activator cmd/activator/main.go`

Dockerfile.activator
```diff
@@ -0,0 +1,28 @@
...
+
+## Multistage deploy
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
+
+WORKDIR /
+COPY --from=builder /activator /activator
```
shmuelk: Change line to:

`COPY --from=builder /workspace/bin/activator /app/activator

Dockerfile.activator
```diff
@@ -0,0 +1,28 @@
...
+
+ENTRYPOINT ["/activator"]
```
shmuelk: Change this line to:

`ENTRYPOINT ["/app/activator"]`

Makefile
dumb0002: Maybe we should make this change after PR https://github.com/llm-d/llm-d-inference-scheduler/pull/397 is merged?

Makefile
dumb0002: Maybe we should make this change after PR https://github.com/llm-d/llm-d-inference-scheduler/pull/397 is merged?

charts/activator/README.md
dumb0002: Could you explain the reason why this API version `inference.networking.k8s.io` should be the default? thanks!

shmuelk: First of all it is the llm-d-inference-scheduler and GIE EPP default.

Second of all it is not clear how much longer latest versions of Gateways will support the old v1alpha2 version of the InferencePool.

charts/activator-filter/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+The following table list the configurable parameters of the chart.
+
+| **Parameter Name**                          | **Description**                                                                                    |
+|---------------------------------------------|----------------------------------------------------------------------------------------------------|
+| `name`                   | Name of the activator RBAC resources. Defaults to `activator`.  |
+
+## Notes
+
+This chart should only be deployed once.
```
shmuelk: In cmd/activator/runner/runner.go, you have the following code:

    metricsServerOptions := metricsserver.Options{
		BindAddress:    fmt.Sprintf(":%d", *metricsPort),
		FilterProvider: filters.WithAuthenticationAndAuthorization,
	}

That code forces the deployer to set up security to access the /metrics path of the Activator. You should probably add something in one of the README.md files (at a minimum) telling the user this fact witha pointer to the K8S documentation on how to configure this.

cmd/activator/runner/runner.go
```diff
@@ -0,0 +1,259 @@
...
+		return err
+	}
+
+	// Start Deactivator
+	go deactivator.MonitorInferencePoolIdleness(ctx)
+
+	// --- Setup Metrics Server ---
+	metricsServerOptions := metricsserver.Options{
+		BindAddress:    fmt.Sprintf(":%d", *metricsPort),
+		FilterProvider: filters.WithAuthenticationAndAuthorization,
```
shmuelk, line 125: See my above comment related to security on the README.md file

pkg/activator/handlers/server.go
```diff
@@ -0,0 +1,202 @@
...
+					},
+				},
+			},
+		}
+	default:
+		return nil, status.Errorf(status.Code(err), "failed to handle request: %v", err)
+	}
+
+	if err.Error() != "" {
+		resp.Response.(*extProcPb.ProcessingResponse_ImmediateResponse).ImmediateResponse.Body = []byte(err.Error())
```
pierDipi, line 168: Should we use OpenAI compatible error response body? https://platform.openai.com/docs/api-reference/responses-streaming/error

pkg/activator/requestcontrol/activator.go
```diff
@@ -0,0 +1,295 @@
+// Package requestcontrol implements the activator logic for controlling request flow
```
pierDipi: Missing copyright header

pkg/activator/server/controller_manager.go
```diff
@@ -0,0 +1,119 @@
...
+	opt, err := defaultManagerOptions(gknn, metricsServerOptions)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create controller manager options: %v", err)
+	}
+
+	if leaderElectionEnabled {
+		opt.LeaderElection = true
+		opt.LeaderElectionResourceLock = "leases"
+		// The lease name needs to be unique per EPP deployment.
+		opt.LeaderElectionID = fmt.Sprintf("epp-%s-%s.gateway-api-inference-extension.sigs.k8s.io", gknn.Namespace, gknn.Name)
```
pierDipi: to avoid conflicts?
```suggestion
		// The lease name needs to be unique per activator deployment.
		opt.LeaderElectionID = fmt.Sprintf("activator-%s-%s.llm-d.ai", gknn.Namespace, gknn.Name)
```

pkg/activator/requestcontrol/activator.go
```diff
@@ -0,0 +1,295 @@
...
+		a.datastore.ResetTicker(DefaultScaleDownDelay) // turn off the deactivator during scale from zero events
+
+		unstructuredObj, err := a.DynamicClient.Resource(gvr).Namespace(namespace).Get(ctx, objname, metav1.GetOptions{})
+		if err != nil {
+			logger.Error(err, "Error getting unstructured object")
+			return false, nil // continue polling
+		}
+
+		// NOTE: this assumes that the target object has a status.readyReplicas field
+		if readyReplicas, ok := unstructuredObj.Object["status"].(map[string]any)["readyReplicas"].(int64); ok {
```
pierDipi: should we use `readyReplicas, ok, err := unstructured.NestedFieldNoCopy(unstructuredObj, "status", "readyReplicas")`?

pierDipi: As currently if status is empty/nil, it would panic

charts/activator-filter/templates/rbac.yaml
lionelvillard: done

Makefile
lionelvillard: done

charts/activator/README.md
lionelvillard: done

pkg/activator/handlers/server.go
lionelvillard: I don't think we need to conform to OpenAI here as this error is internal to Envoy and not exposed to the end-user.

Makefile
lionelvillard: done

pkg/activator/requestcontrol/activator.go
lionelvillard: fixed.

lionelvillard:
rebase done @shmuelk

Dockerfile.activator
elevran: be consistent in build command and container paths across dockerfiles

cmd/activator/main.go
```diff
@@ -0,0 +1,35 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/llm-d/llm-d-inference-scheduler/cmd/activator/runner"
+)
+
+func main() {
+	// For adding out-of-tree plugins to the plugins registry, use the following:
+	// plugins.Register(my-out-of-tree-plugin-name, my-out-of-tree-plugin-factory-function)
+
+	if err := runner.Run(ctrl.SetupSignalHandler()); err != nil {
```
elevran, line 32: future: consider using IGW runner (if it is configurable / refactored to allow better reuse)?

cmd/activator/runner/health.go
elevran, line 1: future: consider using more of IGW "infra" (if it is configurable / refactored to allow better reuse)?

cmd/activator/runner/runner.go
```diff
@@ -0,0 +1,262 @@
...
+// controllers, and servers, and blocks until the manager exits or the context is cancelled.
+func Run(ctx context.Context) error {
+	opts := zap.Options{
+		Development: true,
+	}
+	opts.BindFlags(flag.CommandLine)
+	flag.Parse()
+	initLogging(&opts)
+
+	setupLog.Info("GIE build", "commit-sha", version.CommitSHA, "build-ref", version.BuildRef)
```
elevran: this is not GIE build (sha, etc) but the activator's build

cmd/activator/runner/runner.go
```diff
@@ -0,0 +1,262 @@
...
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
+
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/common"
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/util/logging"
+	"sigs.k8s.io/gateway-api-inference-extension/version"
+
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/activator/datastore"
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/activator/requestcontrol"
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/activator/runnable"
+	runserver "github.com/llm-d/llm-d-inference-scheduler/pkg/activator/server"
```
elevran, line 46: Ideally, we can reuse more of IGW by importing instead of reimplementing

cmd/activator/runner/runner.go
```diff
@@ -0,0 +1,262 @@
...
+	if *enableScaleToZero {
+		deactivator, err := requestcontrol.DeactivatorWithConfig(cfg, &datastore)
+		if err != nil {
+			setupLog.Error(err, "Failed to setup Deactivator")
+			return err
+		}
+
+		// Start Deactivator
+		go deactivator.MonitorInferencePoolIdleness(ctx)
+	}
```
elevran, line 120: shall be removed once WVA handles all scaling and does not require collaboration from activator to scale to zero.
at that point, activtor handles queueing when we need to scale from zero

pkg/activator/requestcontrol/activator.go
```diff
@@ -0,0 +1,298 @@
+// Package requestcontrol implements the activator logic for controlling request flow
+package requestcontrol
+
+import (
```
elevran, line 4: nit: import grouping

pkg/activator/requestcontrol/types.go
elevran, line 1: Is this needed? Is response used ?
Asking since the ext_proc only sends request headers.

pkg/activator/server/controller_manager.go
```diff
@@ -0,0 +1,119 @@
...
+	opt := ctrl.Options{
+		Scheme: scheme,
+		Cache: cache.Options{
+			ByObject: map[client.Object]cache.ByObject{
+				&corev1.Pod{}: {
+					Namespaces: map[string]cache.Config{
+						gknn.Namespace: {},
+					},
+				},
+				&v1alpha2.InferenceObjective{}: {
```
elevran: Q: why is inferenceObjective needed?

pkg/activator/server/runserver.go
```diff
@@ -0,0 +1,156 @@
...
+// AsRunnable returns a Runnable that can be used to start the ext-proc gRPC server.
+// The runnable implements LeaderElectionRunnable with leader election disabled.
+func (r *ExtProcServerRunner) AsRunnable(logger logr.Logger) manager.Runnable {
+	return runnable.NoLeaderElection(manager.RunnableFunc(func(ctx context.Context) error {
+
+		var srv *grpc.Server
+		if r.SecureServing {
+			var cert tls.Certificate
+			var err error
+			if r.CertPath != "" {
```
elevran, line 121: Q: is TLS needed?
If so, should probably include docs on how to create self-signed certificates, etc.
Same as for EPP

pkg/activator/requestcontrol/deactivator.go
```diff
@@ -0,0 +1,108 @@
...
+		datastore:     datastore,
+		DynamicClient: dynamicClient,
+		Mapper:        mapper,
+		ScaleClient:   scaleClient}, nil
+}
+
+// MonitorInferencePoolIdleness monitors the InferencePool for idleness and scales it down to zero replicas
+func (da *Deactivator) MonitorInferencePoolIdleness(ctx context.Context) {
+	logger := log.FromContext(ctx)
+	ds := *(da.datastore)
```
elevran: Q: why deref?

pkg/activator/requestcontrol/deactivator.go
```diff
@@ -0,0 +1,108 @@
...
+			gr := gvr.GroupResource()
+
+			scaleObject, err := da.ScaleClient.Scales(pool.Namespace).Get(ctx, gr, pool.Annotations[objectNameKey], metav1.GetOptions{})
+			if err != nil {
+				logger.Error(err, "Error getting scale subresource object")
+				continue
+			}
+
+			// Scale inferencePool to zero replicas
+			scaleObject.Spec.Replicas = 0
```
elevran, line 100: Q: what is the condition checked to determine that the pool is idle before scaling it to zero?

Do I understand correctly that the timer is rearmed in Activiator on every request and gets to expire here only when there are no requests in flight for delay?
Would be worth documenting this clearly.

pkg/activator/datastore/datastore.go
```diff
@@ -0,0 +1,114 @@
...
+	return ds.pool != nil
+}
+
+func (ds *datastore) Clear() {
+	ds.PoolSet(nil)
+}
+
+func (ds *datastore) ResetTicker(t time.Duration) {
+	ds.poolMu.RLock()
+	defer ds.poolMu.RUnlock()
```
elevran: Lock/Unlock needed, instead of RLock/RUnlock?
Same on L111-112 for StopTicker.

pkg/activator/datastore/datastore.go
```diff
@@ -0,0 +1,114 @@
...
+func (ds *datastore) GetTicker() *time.Ticker {
+	ds.poolMu.RLock()
+	defer ds.poolMu.RUnlock()
+	return ds.ticker
+}
+
+func (ds *datastore) StopTicker() {
+	ds.poolMu.RLock()
+	defer ds.poolMu.RUnlock()
+	ds.ticker.Stop()
```
elevran, line 109: nit: consider draining the ticker's channel unless the deactivator can handle the pending events safely

pkg/activator/server/runserver.go
lionelvillard: > Q: is TLS needed?

Why wouldn't be needed?

pkg/activator/datastore/datastore.go
lionelvillard: no pending events (i.e. tick) will be sent after the ticker is stopped.

Dockerfile.activator
lionelvillard: done.

pkg/activator/requestcontrol/deactivator.go
lionelvillard: fixed.

pkg/activator/requestcontrol/deactivator.go
lionelvillard: I added a comment.

pkg/activator/datastore/datastore.go
lionelvillard: Fixed!

pkg/activator/server/runserver.go
elevran: we might be able to workaround it as only headers are passed to the avtivator.
Not against using TLS, trying to see if can simplify

pkg/activator/datastore/datastore.go
elevran: True. I was asking a slightly different question:
`After Stop, no more ticks will be sent. Stop does not close the channel, to prevent a concurrent goroutine reading from the channel from seeing an erroneous "tick".` 
Could there be ticks on the channel not yet read, when the close happens?
It is probably a rare corner case, if at all. Wanted to bring to attention.

pkg/activator/server/runserver.go
lionelvillard: Let's keep in mind this code is temporary as eventually the activator will be embedded into the EPP. 

Meanwhile, it doesn't hurt to keep our options open regarding TLS.

pkg/activator/datastore/datastore.go
lionelvillard: Thanks. `StopTicker` is only called when the activator terminates and after the scale-down-to-zero monitoring loop. It's a none-issue.

That said, the ticker should not be part of the datastore IMO. Only the ticker channel should be shared among the 2 components, activator and deactivator. We can make this a bit cleaner in a follow-up PR, if that's ok with you.

charts/activator-filter/templates/istio.yaml
```diff
@@ -0,0 +1,25 @@
...
+  configPatches:
+  - applyTo: HTTP_FILTER
+    match:
+      # context omitted so that this applies to both sidecars and gateways
+      listener:
+        filterChain:
+          filter:
+            name: "envoy.filters.network.http_connection_manager"
+    patch:
+      operation: INSERT_FIRST # TODO: insert before EPP
```
nirrozenbaum: OOC: what will happen if this is deployed together with BBR - which ext proc will be invoked first?

charts/activator-filter/README.md
```diff
@@ -0,0 +1,31 @@
...
+$ helm uninstall activator-filter
+```
+
+## Configuration
+
+The following table list the configurable parameters of the chart.
+
+| **Parameter Name**                          | **Description**                                                                                    |
+|---------------------------------------------|----------------------------------------------------------------------------------------------------|
+| `name`                   | Name of the activator RBAC resources. Defaults to `activator`.  |
```
nirrozenbaum, line 29: default should be a combination of the inferencepool name + activator, assuming we're going to have multiple inference pools at some point (near future)

charts/activator-filter/templates/istio.yaml
lionelvillard: it depends on the order of which filters are applied. It really does not matter which one run first, maybe more optimal to have BBR before the activator as the BBR can reject invalid requests.

charts/activator-filter/README.md
```diff
@@ -0,0 +1,31 @@
+# Activator Chart
+
+A chart to deploy the activator HTTP filter for an InferenceGateway and RBAC for all per route activator deployments.
+
+## Install
+
+To install an activator-filter named `activator-filter`, you can run the following command:
```
nirrozenbaum, line 7: whoever is going to use activator isn't necessarily familiar with what is filter. we can call it just `activator`.

charts/activator-filter/templates/rbac.yaml
```diff
@@ -0,0 +1,82 @@
...
+  - "watch"
+  - "list"
+- apiGroups:
+  - "discovery.k8s.io"
+  resources:
+  - "endpointslices"
+  verbs:
+  - "get"
+  - "watch"
+  - "list"
```
nirrozenbaum, line 37: this is legacy. please remove.
```suggestion
```

charts/activator-filter/README.md
lionelvillard: this filter is global, applied only once.

charts/activator-filter/templates/istio.yaml
nirrozenbaum: exactly. you are touching exactly the point I was trying to highlight.
It might be worth adding this information so some section in the deployment instructions.
we should recommend in which order to deploy them (and I agree as you mentioned that BBR should run first).

charts/activator-filter/README.md
nirrozenbaum: to avoid conflicts when deploying multi pools

charts/activator-filter/README.md
lionelvillard: there is already a chart named `activator`. What about `activator-istio`?

nirrozenbaum: is there a use case for deploying one chart without the other?

charts/activator-filter/README.md
lionelvillard: do you mean when deploying multiple gateways? As configured this filter applies to all gateways in the namespace.

nirrozenbaum: ok, then we should be fine 👍

charts/activator-filter/README.md
lionelvillard: both are needed. The only difference is one is applied once per namespace, the other one once per HTTPRoute.

pkg/activator/datastore/datastore.go
elevran: yeah, not critical to do so in this PR

charts/activator-filter/templates/istio.yaml
lionelvillard: I added a note in the readme.

charts/activator-filter/templates/rbac.yaml
```diff
@@ -0,0 +1,82 @@
...
+- apiGroups:
+  - "authorization.k8s.io"
+  resources:
+  - "subjectaccessreviews"
+  verbs:
+  - "create"
+- apiGroups:
+  - "apps"
+  resources:
+  - "deployments"
```
elevran, line 53: The assumption is that only deployments are used, never a replicaset?
Same question regarding deployments/scale,

charts/activator-filter/templates/istio.yaml
```diff
@@ -0,0 +1,25 @@
...
+  configPatches:
+  - applyTo: HTTP_FILTER
+    match:
+      # context omitted so that this applies to both sidecars and gateways
+      listener:
+        filterChain:
+          filter:
+            name: "envoy.filters.network.http_connection_manager"
+    patch:
+      operation: INSERT_FIRST
```
elevran, line 16: do I understand correctly (from `INSERT_FIRST`) that activator runs before BBR?
If so, there could be cases where model (and therefore InferencePool and EPP) are not known yet,

charts/activator-filter/templates/rbac.yaml
lionelvillard: There is no restrictions in the code, other than the target object must have the field `status.readyReplicas`. The RBAC rules can be adjusted as needed. As future work it should be a helm chart parameter.

charts/activator-filter/templates/istio.yaml
lionelvillard: My understanding is `INSERT_FIRST` tells istio/envoy to insert this filter at the very beginning of the HTTP connection manager filter chain. Depending on when the BBR is applied, it will be either before or after. See the helm chart documentation for a recommendation to when to apply the activator filter in relation to the bbr envoy filter. 

Note that the `activator-filter` Envoy filter is a placeholder. It is global and does nothing. It is not associated with any particular InferencePool.

charts/activator-filter/templates/rbac.yaml
elevran: Understood. I realize the GKV received by Activator can be anything, hence was wondering why it has permissions to deployments only. 
It's likely OK for now.

elevran (dismissed):

elevran:
/hold
@lionelvillard please sign commits - missing [verified signature](https://docs.github.com/en/authentication/managing-commit-signature-verification)

pkg/activator/server/controller_manager.go
shmuelk: @lionelvillard Please respond?

Dockerfile.activator
```diff
@@ -0,0 +1,42 @@
+# Build Stage: using Go 1.25 image
+FROM quay.io/projectquay/golang:1.25 AS builder
```
shmuelk: Please use the same GO version as in the rest of the repo. A change slipped in that wasn't suppose to have.

```suggestion
# Build Stage: using Go 1.24.1 image
FROM quay.io/projectquay/golang:1.24.1 AS builder
```

Dockerfile.epp
```diff
@@ -1,4 +1,4 @@
-# Build Stage: using Go 1.24.1 image
+# Build Stage: using Go 1.25 image
 FROM quay.io/projectquay/golang:1.25 AS builder
```
shmuelk: Please set the GO version to 1.24.1. A prior change (not your fault) changed this...

```suggestion
# Build Stage: using Go 1.24.1 image
FROM quay.io/projectquay/golang:1.24.1 AS builder
```

pkg/activator/handlers/server.go
```diff
@@ -0,0 +1,202 @@
...
+				ImmediateResponse: &extProcPb.ImmediateResponse{
+					Status: &envoyTypePb.HttpStatus{
+						Code: envoyTypePb.StatusCode_NotFound,
+					},
+				},
+			},
+		}
+	default:
+		return nil, status.Errorf(status.Code(err), "failed to handle request: %v", err)
+	}
```
shmuelk: Please replace the above code with the following more compact code:

```suggestion
	var code envoyTypePb.StatusCode

	switch errutil.CanonicalCode(err) {
	// This code can be returned by scheduler when there is no capacity for sheddable
	// requests.
	case errutil.InferencePoolResourceExhausted:
		code = envoyTypePb.StatusCode_TooManyRequests
	// This code can be returned by when EPP processes the request and run into server-side errors.
	case errutil.Internal:
		code = envoyTypePb.StatusCode_InternalServerError
	// This code can be returned by the director when there are no candidate pods for the request scheduling.
	case errutil.ServiceUnavailable:
		code = envoyTypePb.StatusCode_ServiceUnavailable
	// This code can be returned when users provide invalid json request.
	case errutil.BadRequest:
		code = envoyTypePb.StatusCode_BadRequest
	case errutil.BadConfiguration:
		code = envoyTypePb.StatusCode_NotFound
	default:
		return nil, status.Errorf(status.Code(err), "failed to handle request: %v", err)
	}

	resp := &extProcPb.ProcessingResponse{
		Response: &extProcPb.ProcessingResponse_ImmediateResponse{
			ImmediateResponse: &extProcPb.ImmediateResponse{
				Status: &envoyTypePb.HttpStatus{
					Code: code,
				},
			},
		},
	}
```

Dockerfile.activator
lionelvillard: All other images are using 1.25. Are you going to open a PR to fix the other Dockerfile?

lionelvillard: My point is we can fix all Dockerfile in this PR if that's ok with you

pkg/activator/server/controller_manager.go
lionelvillard: Sorry missed this comment. I removed it.

Dockerfile.activator
lionelvillard: did .epp and .activator, not .sidecar.

Dockerfile.epp
lionelvillard: done

Dockerfile.activator
shmuelk: Thanks

lionelvillard:
@shmuelk I resolved your recent comments

Issue #380, opened 2025-10-23
[sidecar] Implement data parallel based routing using port hints


elevran:
Upstream gateway support of multiport inference pool is not ready yet (e.g., [this](https://github.com/istio/istio/issues/57638)).

While above are resolved, can use the sidecar as the entrypoint to multiple 
- InferencePool defined with multiple ports => EPP uses data parallel mode
- Each of the vLLM endpoints (i.e., multiple ports on the same Pod IP) is tracked by the inference scheduler independently. This is aligned with current implementation. Each vLLM process has its own `/metrics` endpoint, etc.
- All inference requests are sent to the sidecar which listens on the primary IP:port (i.e., first port only). See below on how accomplished.
- The sidecar receives a new header, set in the EPP, that instructs it to forward to a specific port of the relevant vLLM. The sidecar can do sanity checks (e.g., same IP, port is in range defined in its startup, etc.) on the header and it can also be cleansed by the EPP on receiving client request (same as it should do for the prefill header being injected by clients).
- EPP processes request normally (including disaggregated P/D, I think). We add a new PreRequest plugin that takes the target pod and places it in the new sidecar header and nserts the primary IP:port (i.e., sidecar) into the request so Envoy can route to the "default primary cluster" always. If prerequest hook is insufficient, can achieve the same behavior emulating separate cycles as in P/D with the second cycle only haveing a Picker.

Changes are needed in the EPP and sidecar.

PR #392, opened 2025-10-28
feat: Add vLLM Data Parallel support to llm-d-inference-scheduler

shmuelk:
This PR adds support to the llm-d-inference-scheduler for vLLM's Data Parallel (DP) feature. This vLLM feature enables the running of multiple HTTP Servers in the same container.

In particular this PR attempts to work around the Istio issue [57638](https://github.com/istio/istio/issues/57638)

This PR adds a new profile handler plugin `data-parallel-profile-handler` which needs to be used to enable DP when Disambiguated  Prefill/Decode (PD) isn't used. This plugin changes the scheduling result to always schedule the request on the "base" port on the vLLM decode pod and adds a header for the sidecar.

Additionally the sidecar has been extended to work with the header added by the new `data-parallel-profile-handler` plugin.  This is a work around to deal with Istio issue [57638](https://github.com/istio/istio/issues/57638) in which we can only route to teh "base" port of vLLM pods. It also always proxies all direct HTTP calls to the "extra" HTTP servers started by vLLM.

Ref #380

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,7 @@ test-integration: download-tokenizer install-dependencies ## Run integration tes
 	go test -ldflags="$(LDFLAGS)" -v -tags=integration_tests ./test/integration/
 
 .PHONY: test-e2e
-test-e2e: image-build sidecar-image-build ## Run end-to-end tests against a new kind cluster
+test-e2e: image-build ## Run end-to-end tests against a new kind cluster
 	@printf "\033[33;1m==== Running End to End Tests ====\033[0m\n"
 	./test/scripts/run_e2e.sh
 


deploy/components/crds-gie/kustomization.yaml
```diff
@@ -10,4 +10,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-- https://github.com/kubernetes-sigs/gateway-api-inference-extension/config/crd?ref=v1.0.0
+- https://github.com/kubernetes-sigs/gateway-api-inference-extension/config/crd?ref=v1.1.0-rc.1
```
nirrozenbaum, line 13: v1.1.0 is out. let's use formal release:
```suggestion
- https://github.com/kubernetes-sigs/gateway-api-inference-extension/config/crd?ref=v1.1.0
```

deploy/components/inference-gateway/rbac.yaml
```diff
@@ -12,6 +12,14 @@ rules:
   - "get"
   - "watch"
   - "list"
+- apiGroups:
+  - "inference.networking.k8s.io"
```
nirrozenbaum, line 16: can we remove inferencepools from the `inference.networking.x-k8s.io` group?

deploy/components/istio-control-plane/rbac.yaml
```diff
@@ -330,6 +330,21 @@ rules:
   verbs:
   - update
   - patch
+- apiGroups:
```
nirrozenbaum, line 333: ditto - remove from `inference.networking.x-k8s.io` group.

deploy/components/vllm-sim/deployments.yaml
```diff
@@ -27,10 +27,32 @@ spec:
         - "--zmq-endpoint=tcp://${EPP_NAME}.default.svc.cluster.local:5557"
         - "--event-batch-size=16"
         - "--tokenizers-cache-dir=/tokenizer-cache"
+        - "--data-parallel-size=${VLLM_DATA_PARALLEL_SIZE}"
```
nirrozenbaum, line 71: I'm a bit confused. 
the flag `--data-parallel-size` is reading from env var `VLLM_DATA_PARALLEL_SIZE` which makes this configureable (which is great).

but the ports list below is hardcoded.

deploy/components/vllm-sim-pd/deployments.yaml
```diff
@@ -55,21 +55,74 @@ spec:
         - "--port=8000"
         - "--vllm-port=8200"
         - "--connector=lmcache"
+        - "--secure-proxy=false"
+        - "--data-parallel-size=${VLLM_DATA_PARALLEL_SIZE}"
```
nirrozenbaum, line 59: I thought dp is not used together with pd.
(this file is vllm-sim-pd).
what am I missing?

deploy/environments/dev/base-kind-istio/patch-deployments.yaml
```diff
@@ -13,7 +13,7 @@ spec:
         - --pool-namespace
         - "default"
         - "--pool-group"
-        - "inference.networking.x-k8s.io"
+        - "inference.networking.k8s.io"
```
nirrozenbaum: I think this is the default si maybe we can omit these lines.
(should be checked, but almost sure about that)

pkg/plugins/profile/dp_profile_handler.go
```diff
@@ -0,0 +1,120 @@
+package profile
```
nirrozenbaum, line 1: rename file to dp_profile_handler?

pkg/plugins/profile/dp_profile_handler.go
```diff
@@ -0,0 +1,120 @@
...
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/scheduling/types"
+)
+
+const (
+	// DataParallelProfileHandlerType is the type of the DataParallelProfileHandler
+	DataParallelProfileHandlerType = "data-parallel-profile-handler"
+)
+
+type dataParallelProfileHandlerParameters struct {
+	TargetPort int `json:"targetPort"`
```
nirrozenbaum: maybe we should name this parameters to be more self explained?
TargetPort could be a bit confusing because it's not really the target.. it's more of a `PrimaryPort` (that serves as a proxy, not the target).

pkg/plugins/profile/dp_profile_handler.go
```diff
@@ -0,0 +1,120 @@
...
+
+	request.Headers[common.DataParallelPodHeader] = net.JoinHostPort(targetPod.Address, targetPod.Port)
+
+	for _, target := range profileResult.TargetPods {
+		newPodInfo := target.GetPod().Clone()
+		newPodInfo.Port = h.targetPort
+		targetPod := &types.PodMetrics{Pod: newPodInfo, MetricsState: target.GetMetrics().Clone()}
+		newResult.TargetPods = append(newResult.TargetPods, targetPod)
+	}
+	modifiedResults := map[string]*types.ProfileRunResult{singleProfileName: &newResult}
```
nirrozenbaum, line 114: not sure I got the idea of why we need `newResult` or `modifiedResults`.

can't we iterate over range profileResult.TargetPods and update only the port to `PrimaryPort`?

deploy/config/dp-epp-config.yaml
```diff
@@ -0,0 +1,18 @@
+# Sample EPP configuration for running with Data Parallel
+#
+apiVersion: inference.networking.x-k8s.io/v1alpha1
+kind: EndpointPickerConfig
+plugins:
+- type: prefix-cache-scorer
+- type: decode-filter
```
nirrozenbaum, line 7: this is a bit confusing cause dp is currently used without pd. 
I understand why it's used, but maybe there is a less confusing alternative.

deploy/components/inference-gateway/rbac.yaml
shmuelk: I left it in in case someone wants to use the v1alpha2 version...

deploy/components/vllm-sim/deployments.yaml
shmuelk: The ports just need to be exposed. There can't be more than eight ports. The flag simply says how many we want to use.

deploy/components/vllm-sim-pd/deployments.yaml
shmuelk: I have updated the non-pd deployment files to include the sidecar. The script has been changed to launch using the non-pd deployment files even if VLLM_DATA_PARALLEL_SIZE is greater than one.

A subsequent PR will add DP support for PD deployments. For now we'll leave this change here

pkg/plugins/profile/dp_profile_handler.go
shmuelk: I don't want to depend on the fact that somewhere up the stack the PodInfo that the plugin received was a clone....

deploy/config/dp-epp-config.yaml
shmuelk: That filter is in all of the configuration files here. I actually don't know why....

deploy/components/crds-gie/kustomization.yaml
shmuelk: The go.mod file still refers to v1.1.0-rc.1. I'd prefer to keep things in sync.

The move to GIE 1.1.0 should be in a separate PR

pkg/plugins/profile/dp_profile_handler.go
shmuelk: done

pkg/plugins/profile/dp_profile_handler.go
shmuelk: Done

deploy/environments/dev/base-kind-istio/patch-deployments.yaml
shmuelk: Done. It is the default.

deploy/components/istio-control-plane/rbac.yaml
shmuelk: The kgateway deployment files still use the v1alpha2 InferencePool

deploy/components/inference-gateway/rbac.yaml
shmuelk: The kgateway deployment files still use the v1alpha2 InferencePool

deploy/components/vllm-sim-pd/deployments.yaml
vMaroon: The primary use for DP will be with PD deployments, I think supporting it is a feature blocker. Today's hot deployments are TP+DP+EP with PD 😆.

PR #404, opened 2025-10-29
Allow the sidecar to sample from a list of prefill host ports

smarterclayton:
In some benchmarking and test environments dynamic prefill selection
may be difficult and random selection among a set of hosts is
sufficient.

Add a new `--enable-prefiller-sampling` flag that instructs the
sidecar to select a random prefill host from the provided list
instead of the first one. Make the behavior opt-in to prevent
users from accidentally depending on the new behavior, and
keep the existing default behavior (first header value) consistent.

E.g.:

    curl -H 'x-prefiller-host-port: server1:8000` -H 'x-prefiller-host-port: server2:8000'

will randomly choose one of the two values.

Moved from the original repo

### DIFF
--- a/pkg/sidecar/proxy/chat_completions_test.go
+++ b/pkg/sidecar/proxy/chat_completions_test.go
@@ -1,0 +1,158 @@
+/*
+Copyright 2025 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package proxy
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/common"
+)
+
+func TestServer_chatCompletionsHandler(t *testing.T) {
+	tests := []struct {
+		name     string
+		sampling bool
+		r        *http.Request
+
+		expectedCode             int
+		expectedPrefillHostPorts []string
+		expectedPassthrough      bool
+	}{
+		{
+			name: "passthrough by default",
+			r:    &http.Request{},
+
+			expectedPassthrough: true,
+		},
+		{
+			name: "passthrough with no header value",
+			r:    &http.Request{Header: http.Header{http.CanonicalHeaderKey(common.PrefillPodHeader): []string{}}},
+
+			expectedPassthrough: true,
+		},
+		{
+			name: "default prefill to one header value",
+			r:    &http.Request{Header: http.Header{http.CanonicalHeaderKey(common.PrefillPodHeader): []string{"a"}}},
+
+			expectedCode:             200,
+			expectedPrefillHostPorts: []string{"a"},
+		},
+		{
+			name: "default prefill to first header value",
+			r:    &http.Request{Header: http.Header{http.CanonicalHeaderKey(common.PrefillPodHeader): []string{"a,b"}}},
+
+			expectedCode:             200,
+			expectedPrefillHostPorts: []string{"a"},
+		},
+		{
+			name:     "sample from comma delimited header",
+			r:        &http.Request{Header: http.Header{http.CanonicalHeaderKey(common.PrefillPodHeader): []string{"a,b"}}},
+			sampling: true,
+
+			expectedCode:             200,
+			expectedPrefillHostPorts: []string{"a", "b"},
+		},
+		{
+			name:     "sample from comma delimited header with whitespace",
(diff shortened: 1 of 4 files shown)


pkg/sidecar/proxy/chat_completions_test.go
```diff
@@ -0,0 +1,88 @@
...
+		name     string
+		port     string
+		sampling bool
+		r        *http.Request
+
+		expectedCode        int
+		expectedPrefillerIn []string
+		expectedPassthrough bool
+	}{
+		{r: &http.Request{}, expectedPassthrough: true},
```
shmuelk: Please fill in the name field in each of the tests. Also print the name field in any error messages. That way we know which test failed.

pkg/sidecar/proxy/chat_completions_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"slices"
+	"testing"
+
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/common"
+)
+
+func TestServer_chatCompletionsHandler(t *testing.T) {
+	tests := []struct {
+		name     string
+		port     string
```
shmuelk: As you never use the port field and don't really need it, please delete it from the struct.

smarterclayton:
PTAL.  Moved the server handler types back to `http.Handler` so they could be replaced at test time, not sure whether that was a change you were making ahead of time.  At minimum the test needs to mock the handlers, suggest if we need to call other things on decode we switch to a local interface.

pkg/sidecar/proxy/chat_completions_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+			expectedPassthrough: true,
+		},
+		{
+			name:     "sample from multiple empty header values",
+			r:        &http.Request{Header: http.Header{http.CanonicalHeaderKey(common.PrefillPodHeader): []string{"", ""}}},
+			sampling: true,
+
+			expectedPassthrough: true,
+		},
```
shmuelk, line 109: How about a test without the PrefillPodHeader at all?

pkg/sidecar/proxy/chat_completions_test.go
```diff
@@ -0,0 +1,150 @@
...
+				var hostPort string
+				s.runConnectorProtocol = func(_ http.ResponseWriter, _ *http.Request, selectedHostPort string) { hostPort = selectedHostPort }
+				var passthrough bool
+				s.decoderProxy = http.HandlerFunc(func(_ http.ResponseWriter, _ *http.Request) {
+					passthrough = true
+				})
+				s.dataParallelProxies = make(map[string]http.Handler)
+				recorder := httptest.NewRecorder()
+				recorder.Code = 0
+				s.chatCompletionsHandler(recorder, tt.r)
```
shmuelk, line 130: While recorder is technically an http.ResponseWriter, one is not suppose to reference it's fields directly.

Please add:

`resp := recorder.Response()`

and in the following code replace `recorder` with `resp`

pkg/sidecar/proxy/chat_completions_test.go
```diff
@@ -0,0 +1,150 @@
...
+				if passthrough {
+					if !tt.expectedPassthrough {
+						t.Errorf("unexpected passthrough to decode")
+					}
+					if recorder.Body.Len() > 0 || recorder.Code != 0 || len(recorder.Header()) > 0 {
```
shmuelk: See my comment above. recoder.Body, may not actually have a value....

shmuelk (changes requested):

pkg/sidecar/proxy/chat_completions_test.go
smarterclayton: Not sure I understand your comment, here's the ResponseRecorder godoc:

```
// ResponseRecorder is an implementation of [http.ResponseWriter] that
// records its mutations for later inspection in tests.
```

Why can I not pass an implementation of a ResponseWriter to a handler that accepts a ResponseWriter?

pkg/sidecar/proxy/chat_completions_test.go
smarterclayton: That is the first test, `passthrough by default`.  Or are you suggesting testing with an initialized but empty `http.Header`?

pkg/sidecar/proxy/chat_completions_test.go
smarterclayton: Ah, you meant "call Response() before checking fields of recorder **after** the handler".

smarterclayton: In this case, what the test is verifying is that the handler does nothing to the ResponseWriter, and using Response doesn't quite provide that.  The test is check the invariant that no fields on ResponseWriter were invoked, and so using Recorder directly is appropriate for the pass through scenario.

pkg/sidecar/proxy/chat_completions_test.go
smarterclayton: `recorder.Body` is initialized when using `NewRecorder`

pkg/sidecar/proxy/chat_completions_test.go
shmuelk: It turns out that if Request.Header is nil, Request.Header.Get(???) return "".

shmuelk (approved):

PR #410, opened 2025-10-30
Chat Completions support; WIP

guygir:
1. Dockerfile
  - Minimal Python-enabled runtime to support chat templating.
  - Builder: installs Python 3.12 headers; fetches preprocessing requirements and `render_jinja_template_wrapper.py`.
  - Runtime: use of microdnf because a package manager is required to install Python 3.12 and zeromq at runtime, and microdnf is lighter than dnf (fewer dependencies), so it keeps the image smaller.; copies the wrapper into `/usr/local/lib/python3.12/site-packages/`.
  - Python deps installed to a single site-packages, `--no-cache-dir` to keep image size smaller; `torch` filtered out manually for now.
  - Env set for reliable imports and HF cache (`PYTHONPATH`, `HF_HOME`).
  
2. Precise Prefix Cache (scorer)
  - Adds real chat-completions preprocessing before scoring.
  - Builds `RenderJinjaTemplateRequest` from messages (uses `msg.Content.Raw`), initializes processor, fetches model chat template, renders, returns flattened prompt to KV-cache scoring.
  - Regular completions use prompt directly.
  - Includes `PREPROCESSING:` INFO/DEBUG logs; will be trimmed post-WIP.

3. Submodules
  - Adds empty placeholders for `llm-d-kv-cache-manager` and `gateway-api-inference-extension`.
  - Purpose: keep upstream references during integration without pulling large repos.
  - Final build relies on Go modules and fetches Python files during Docker build; submodule content not required.
  - Documents where things came from and makes it easier to work with if needed later; can be removed if it becomes unnecessary later.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,37 +1,45 @@
 ## Minimal runtime Dockerfile (microdnf-only, no torch, wrapper in site-packages)
-# Build stage
+# Build Stage: using Go 1.25 image
 FROM quay.io/projectquay/golang:1.25 AS builder
 ARG TARGETOS
 ARG TARGETARCH
 
+# Install build tools
+# The builder is based on UBI8, so we need epel-release-8.
 RUN dnf install -y 'https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm' && \
     dnf install -y gcc-c++ libstdc++ libstdc++-devel clang zeromq-devel pkgconfig python3.12-devel python3.12-pip git && \
     dnf clean all
+# python3.12-devel needed for CGO compilation (Python headers and python3.12-config for linker flags)
 
 WORKDIR /workspace
 
+# Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
+
+# Copy the go source
 COPY cmd/ cmd/
 COPY pkg/ pkg/
+COPY scripts/fetch-python-wrapper.sh scripts/fetch-python-wrapper.sh
 
 # Fetch requirements only (install in runtime)
 RUN curl -L https://raw.githubusercontent.com/llm-d/llm-d-kv-cache-manager/v0.3.2/pkg/preprocessing/chat_completions/requirements.txt -o /workspace/requirements.txt
 
 # Prepare Python wrapper file for runtime
-RUN mkdir -p /tmp/kv-cache-manager && \
-    cd /tmp/kv-cache-manager && \
-    git clone --depth 1 --branch v0.3.2 https://github.com/llm-d/llm-d-kv-cache-manager.git . && \
-    mkdir -p /workspace/llm-d-kv-cache-manager/pkg/preprocessing/chat_completions && \
-    cp pkg/preprocessing/chat_completions/render_jinja_template_wrapper.py /workspace/llm-d-kv-cache-manager/pkg/preprocessing/chat_completions/ && \
-    rm -rf /tmp/kv-cache-manager
+RUN scripts/fetch-python-wrapper.sh v0.3.2 /workspace/llm-d-kv-cache-manager/pkg/preprocessing/chat_completions
 
-# Tokenizers static lib
+# HuggingFace tokenizer bindings (static lib)
 RUN mkdir -p lib
+# Ensure that the RELEASE_VERSION matches the one used in the imported llm-d-kv-cache-manager version
 ARG RELEASE_VERSION=v1.22.1
 RUN curl -L https://github.com/daulet/tokenizers/releases/download/${RELEASE_VERSION}/libtokenizers.${TARGETOS}-${TARGETARCH}.tar.gz | tar -xz -C lib
 RUN ranlib lib/*.a
 
+# Build
+# the GOARCH has not a default value to allow the binary be built according to the host where the command
+# was called. For example, if we call make image-build in a local env which has the Apple Silicon M1 SO
+# the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
+# by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
 ENV CGO_ENABLED=1
 ENV GOOS=${TARGETOS:-linux}
 ENV GOARCH=${TARGETARCH}
@@ -45,12 +53,16 @@ RUN export CGO_CFLAGS="$(python3.12-config --cflags) -I/workspace/lib" && \
     go build -a -o bin/epp -ldflags="-extldflags '-L$(pwd)/lib' -X sigs.k8s.io/gateway-api-inference-extension/version.CommitSHA=${COMMIT_SHA} -X sigs.k8s.io/gateway-api-inference-extension/version.BuildRef=${BUILD_REF}" cmd/epp/main.go
 
 # Runtime stage
+# Use ubi9 as a minimal base image to package the manager binary
+# Refer to https://catalog.redhat.com/software/containers/ubi9/ubi-minimal/615bd9b4075b022acc111bf5 for more details
 FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
 WORKDIR /
 COPY --from=builder /workspace/bin/epp /app/epp
 
 USER root
-# microdnf only; add EPEL and install Python runtime
+# Install zeromq runtime library and Python runtime needed by the manager.
+# The final image is UBI9, so we need epel-release-9.
+# Using microdnf for minimal image size
 RUN curl -L -o /tmp/epel-release.rpm https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm && \
     rpm -i /tmp/epel-release.rpm && \
     rm /tmp/epel-release.rpm && \
@@ -59,6 +71,7 @@ RUN curl -L -o /tmp/epel-release.rpm https://dl.fedoraproject.org/pub/epel/epel-
     rm -rf /var/cache/yum /var/lib/yum && \
     ln -sf /usr/bin/python3.12 /usr/bin/python3 && \
     ln -sf /usr/bin/python3.12 /usr/bin/python
+# Note: python3.12 package does not automatically create python3/python symlinks - they must be created manually
 
 # Install wrapper as a module in site-packages
 RUN mkdir -p /usr/local/lib/python3.12/site-packages/
@@ -82,9 +95,11 @@ ENV HF_HOME="/tmp/.cache"
 
 USER 65532:65532
 
+# expose gRPC, health and metrics ports
 EXPOSE 9002
 EXPOSE 9003
 EXPOSE 9090
+# expose port for KV-Events ZMQ SUB socket
 EXPOSE 5557
 
 ENTRYPOINT ["/app/epp"]
--- a/pkg/plugins/scorer/precise_prefix_cache.go
+++ b/pkg/plugins/scorer/precise_prefix_cache.go
@@ -79,7 +79,6 @@ func New(ctx context.Context, config PrecisePrefixCachePluginConfig) (*PrecisePr
 	if err != nil {
 		return nil, fmt.Errorf("failed to create `kvcache.Indexer`: %w", err)
 	}
-	// init the preProcessor, maroon fork let him handle
 
 	go kvCacheIndexer.Run(ctx)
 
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -1,66 +1,91 @@
...
+# Build stage
 FROM quay.io/projectquay/golang:1.25 AS builder
 ARG TARGETOS
 ARG TARGETARCH
 
-# Install build tools
-# The builder is based on UBI8, so we need epel-release-8.
 RUN dnf install -y 'https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm' && \
-    dnf install -y gcc-c++ libstdc++ libstdc++-devel clang zeromq-devel pkgconfig && \
+    dnf install -y gcc-c++ libstdc++ libstdc++-devel clang zeromq-devel pkgconfig python3.12-devel python3.12-pip git && \
```
elevran, line 10: Q: not familiar with python packages. Is `-devel` needed or just a runtime package? I guess the `-devel` is for writing extensions in other languages?

Dockerfile.epp
```diff
@@ -1,66 +1,91 @@
...
-# Install build tools
-# The builder is based on UBI8, so we need epel-release-8.
 RUN dnf install -y 'https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm' && \
-    dnf install -y gcc-c++ libstdc++ libstdc++-devel clang zeromq-devel pkgconfig && \
+    dnf install -y gcc-c++ libstdc++ libstdc++-devel clang zeromq-devel pkgconfig python3.12-devel python3.12-pip git && \
     dnf clean all
 
 WORKDIR /workspace
 
-# Copy the Go Modules manifests
```
elevran, line 14: Q: what was the goal of removing the comments?

Dockerfile.epp
```diff
@@ -1,66 +1,91 @@
...
+    rm -rf /tmp/kv-cache-manager
+
+# Tokenizers static lib
 RUN mkdir -p lib
-# Ensure that the RELEASE_VERSION matches the one used in the imported llm-d-kv-cache-manager version
 ARG RELEASE_VERSION=v1.22.1
 RUN curl -L https://github.com/daulet/tokenizers/releases/download/${RELEASE_VERSION}/libtokenizers.${TARGETOS}-${TARGETARCH}.tar.gz | tar -xz -C lib
 RUN ranlib lib/*.a
 
-# Build
```
elevran, line 29: ditto - why remove the comments? no longer true?

Dockerfile
```diff
@@ -1,66 +1,91 @@
...
-# Copy the go source
 COPY cmd/ cmd/
 COPY pkg/ pkg/
 
-# HuggingFace tokenizer bindings
+# Fetch requirements only (install in runtime)
+RUN curl -L https://raw.githubusercontent.com/llm-d/llm-d-kv-cache-manager/v0.3.2/pkg/preprocessing/chat_completions/requirements.txt -o /workspace/requirements.txt
+
+# Prepare Python wrapper file for runtime
+RUN mkdir -p /tmp/kv-cache-manager && \
```
elevran: it might be better to create a script that does all that since it's likely needed in local builds as well.

Dockerfile
elevran, line 1: Please restore all removed comments unless there's a valid reason (e.g., no longer true, misleading...)

Dockerfile
```diff
@@ -1,66 +1,91 @@
...
-    rm -rf /var/cache/dnf /var/lib/dnf
+# microdnf only; add EPEL and install Python runtime
+RUN curl -L -o /tmp/epel-release.rpm https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm && \
+    rpm -i /tmp/epel-release.rpm && \
+    rm /tmp/epel-release.rpm && \
+    microdnf install -y --setopt=install_weak_deps=0 zeromq python3.12 python3.12-libs python3.12-pip && \
+    microdnf clean all && \
+    rm -rf /var/cache/yum /var/lib/yum && \
+    ln -sf /usr/bin/python3.12 /usr/bin/python3 && \
+    ln -sf /usr/bin/python3.12 /usr/bin/python
```
elevran, line 73: isn't that part of the microdnf install? typical install sets up the paths.

gateway-api-inference-extension
elevran, line 1: what is the rationale of adding this as a submodule? it was not needed before.

llm-d-kv-cache-manager
elevran, line 1: same - why submodule and not a runtime copy/import?

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -78,6 +79,7 @@ func New(ctx context.Context, config PrecisePrefixCachePluginConfig) (*PrecisePr
 	if err != nil {
 		return nil, fmt.Errorf("failed to create `kvcache.Indexer`: %w", err)
 	}
+	// init the preProcessor, maroon fork let him handle
```
elevran: `maroon fork let him handle`?

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -115,23 +117,40 @@ func (s *PrecisePrefixCacheScorer) WithName(name string) *PrecisePrefixCacheScor
 // Score scores the provided pod based on the KVCache index state.
 // The returned scores are normalized to a range of 0-1.
 func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, _ *types.CycleState, request *types.LLMRequest, pods []types.Pod) map[types.Pod]float64 {
-	loggerDebug := log.FromContext(ctx).WithName(s.typedName.String()).V(logutil.DEBUG)
+	logger := log.FromContext(ctx).WithName(s.typedName.String())
+	loggerDebug := logger.V(logutil.DEBUG)
+	
 	if request == nil {
+		logger.Info("PREPROCESSING: Request is nil, skipping scoring")
```
elevran: - seems redundant with next log line?
- Also the addition of `PRPROCESSING` prefix everywhere is inconsistent with other log lines, please drop or add to regular message where applicable.
- please review message levels throughout. Debug should be used for developer and Info are for user.

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -115,23 +117,40 @@ func (s *PrecisePrefixCacheScorer) WithName(name string) *PrecisePrefixCacheScor
 // Score scores the provided pod based on the KVCache index state.
 // The returned scores are normalized to a range of 0-1.
 func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, _ *types.CycleState, request *types.LLMRequest, pods []types.Pod) map[types.Pod]float64 {
-	loggerDebug := log.FromContext(ctx).WithName(s.typedName.String()).V(logutil.DEBUG)
+	logger := log.FromContext(ctx).WithName(s.typedName.String())
```
elevran, line 119: the two seem to be duplicating messages at different levels.
can you do with a single logger and adjust the log level?

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -115,23 +117,40 @@ func (s *PrecisePrefixCacheScorer) WithName(name string) *PrecisePrefixCacheScor
...
 	}
 
-	prompt, err := getUserInput(request)
+	// Preprocess the request to get the flattened prompt
+	logger.Info("PREPROCESSING: Starting preprocessing", 
+		"target_model", request.TargetModel,
+		"has_chat_completions", request.Body != nil && request.Body.ChatCompletions != nil,
+		"has_completions", request.Body != nil && request.Body.Completions != nil)
+	
+	prompt, err := s.preprocessRequest(ctx, request)
```
elevran: is it doing more than extracting the user prompt?
If so, make the function name reflect that instead of the more generic `preprocessRequest`

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -146,15 +165,118 @@ func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, _ *types.CycleStat
...
+		// Create preprocessing processor
+		processor := chat_completions.NewChatTemplatingProcessor()
+
+		// Initialize the processor (this starts Python interpreter if not already started)
+		// The C code handles global initialization tracking to prevent multiple initializations
+		if err := processor.Initialize(); err != nil {
+			return "", fmt.Errorf("failed to initialize chat templating processor: %w", err)
+		}
+
+		// First, fetch the chat template from the model
```
elevran: Is there a reason to do all these intermediate steps in Go and call out to python multiple times?
what are the tradeoffs vs shuttling all to python in one call?

elevran: Do we have a processor per goroutine or reusing one? If reusing, where is the Python GIL managed? I suspect it may serliaze all requests in the system due to the GIL.
Have you tested with many concurrent requests?

vMaroon: The GIL is managed here https://github.com/llm-d/llm-d-kv-cache-manager/blob/main/pkg/preprocessing/chat_completions/cgo_functions.c.

Benchmarks were ran, one is in this doc: https://github.com/llm-d/llm-d-kv-cache-manager/tree/main/pkg/preprocessing/chat_completions#performance-analysis.

Dockerfile.epp
guygir: No goal - I rewrote the script and accidentally left some of them out. I’ve now restored/adjusted all previously removed comments.

Dockerfile
guygir: Good idea. Added a script (scripts/fetch-python-wrapper.sh) to handle fetching the Python wrapper file from the kv-cache-manager.

pkg/plugins/scorer/precise_prefix_cache.go
guygir: Oops, some comments were accidentally left out and others accidentally left in... Removed.

pkg/plugins/scorer/precise_prefix_cache.go
guygir: Agree, these were part of the WIP and used for my debugging, so they were messy. Removed the “PREPROCESSING:” prefixes, cleaned up duplicate logs, and adjusted log levels - developer details are now at DEBUG, for user messages at INFO/ERROR.

pkg/plugins/scorer/precise_prefix_cache.go
guygir: For regular completions, it only extracts the prompt, while for chat completions it also pre-processes by fetching and rendering the messages. Since the core idea is to extract the finalized prompt, I renamed the function from preprocessRequest to extractPrompt. It could potentially be split into separate functions for completions extraction and chat completions preprocessing plus extraction, but that seems redundant.

pkg/plugins/scorer/precise_prefix_cache.go
guygir: The separation is intentional to support template caching. The kv-cache manager maintains an in memory cache per model/revision/token, allowing templates to be reused across requests. Both the module and Python functions are cached at the C level, so subsequent calls have very low overhead - only minimal GIL acquire/release and JSON marshaling, while module import and function lookup are cached.

Regarding trade-offs: multiple calls add negligible overhead but preserve caching, reduce repeated HF API calls, and allow error granularity between fetch and render. Combining into a single python call would remove caching, require repeated template fetches, and degrade performance in high throughput scenarios.

Note, this is all based on the benchmarks we ran. As we've discussed, maybe the workload wasn't representing enough... will benchmark again with a different workload to verify.

EDIT: Actually I think we can still keep the caching even if we combine fetching and rendering into a single call... need to measure the overhead on the separate calls, if it isn't very low maybe it's worth to try and combine them.

Dockerfile
guygir: From my testing, it seems only /usr/bin/python3.12 exists and the usual python3 and python symlinks are missing. Directly running /usr/bin/python3 or /usr/bin/python fails, so seems like the python3.12 package does not create these symlinks automatically like the standard python3 package.

gateway-api-inference-extension
guygir: These empty directories are leftover "artifacts" from my earlier integration work in the branch...
They were added as placeholders, are not actual submodules, and serve no functional purpose now - so they have been removed.

Dockerfile
guygir: Not very familiar with this myself, but from what I understand, python3.12-devel is needed in the builder stage, not at runtime. It provides the Python headers and the python3.12-config tool required for CGO compilation when go code links against python.
The runtime stage only needs python3.12 and python3.12-libs, which are already included.
Without -devel, the build will fail because of missing headers or the python3.12-config command.
Added a comment in the Dockerfile to clarify this in case we revisit this in the future...

guygir:
Updated the PR, and all previous comments have been addressed.
Regarding the last comment by @vMaroon - I will re-run the benchmark (including a more reliable workload, as discussed with @elevran) to produce a detailed doc similar to the one in kv-cache manager.

Note on DCO: Two commits in the branch (ec6d849 by learner0810, c40cc15 by Morgan Foster) are currently failing DCO checks. These are upstream commits that came into this branch by merging upstream, and they have the same signoff issues in upstream/main itself (missing/incorrect signoffs). Guidance on how to handle these upstream commits - should they be addressed in upstream/main, or is there another approach that I should take in my branch?

PR #412, opened 2025-10-31
Enhancement: return 503 instead of 502 when decode node is not ready

Phil-OSophy-42:
## Description

Enhancement https://github.com/llm-d/llm-d-inference-scheduler/issues/402 Improve error handling in the sidecar proxy to provide clearer feedback when
the decode node (vLLM) is not ready.

## Changes

- Return `503 Service Unavailable` instead of `502` when ECONNREFUSED occurs
- Add structured error response with clear message
- Enhance error logging

### DIFF
--- a/pkg/sidecar/proxy/errors.go
+++ b/pkg/sidecar/proxy/errors.go
@@ -30,6 +30,8 @@ type errorResponse struct {
 	Code    int    `json:"code"`
 }
 
+const decoderServiceUnavailableResponseJSON = `{"object":"error","message":"The decode node is not ready. Please check that the vLLM service is running and the port configuration is correct.","type":"ServiceUnavailable","param":"","code":503}`
+
 func errorJSONInvalid(err error, w http.ResponseWriter) error {
 	// Simulate vLLM error
 
--- a/pkg/sidecar/proxy/base_proxy.go
+++ b/pkg/sidecar/proxy/base_proxy.go
@@ -120,11 +120,19 @@ func (s *BaseServer) createDecoderProxyHandler(decoderURL *url.URL, decoderInsec
 		// Log errors from the decoder proxy
 		switch {
 		case errors.Is(err, syscall.ECONNREFUSED):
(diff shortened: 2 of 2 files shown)


pkg/sidecar/proxy/errors.go
```diff
@@ -76,3 +76,21 @@ func errorBadGateway(err error, w http.ResponseWriter) error {
...
+
+func errorServiceUnavailable(message string, w http.ResponseWriter) error {
+	er := errorResponse{
+		Object:  "error",
+		Message: message,
+		Type:    "ServiceUnavailable",
+		Code:    http.StatusServiceUnavailable,
+	}
+
+	b, err := json.Marshal(er)
```
elevran: nit: can inline the `er` struct into the parameter to avoid the intermediate variable.

pkg/sidecar/proxy/proxy.go
```diff
@@ -260,11 +260,18 @@ func (s *Server) createRoutes() *http.ServeMux {
 		// Log errors from the decoder proxy
 		switch {
 		case errors.Is(err, syscall.ECONNREFUSED):
-			s.logger.Error(err, "waiting for vLLM to be ready")
+			s.logger.Error(err, "failed to connect to vLLM decoder",
+				"decoderURL", s.decoderURL.String())
+			if err := errorServiceUnavailable("The decode node is not ready. Please check that the vLLM service is running and the port configuration is correct.", res); err != nil {
```
elevran: why is the `if` needed (instead of a direct assignment of `err := ...`)?
Also, since the message (and thus error return) is fixed, you might define it once as constant and avoid calling functions and marshalling on requests.

Phil-OSophy-42: Done. Changed err := ... to a direct assignment and defined a constant for decodeServiceUnavailable. The if statement is needed to log errors when writing the response to the client fails.

elevran (approved):

PR #426, opened 2025-11-05
Added chat completions preprocessing support

guygir:
This PR adds support for chat completions preprocessing in the inference scheduler, enabling proper handling of chat template formats.

#### 1. **Dockerfile.epp**

The Dockerfile has been updated to support chat completions preprocessing. It now fetches the Python wrapper from `llm-d-kv-cache-manager` (v0.3.2) during the build process, installs the necessary chat completions preprocessing dependencies, and integrates the Python wrapper script into the runtime image. Documentation has been adjusted to clarify the Dockerfile structure and purpose.

#### 2. **precise_prefix_cache.go**

The precise prefix cache scorer now includes chat completions preprocessing support. This implementation handles chat template formats using Jinja2 templates, allowing the system to properly process OpenAI Chat Completions API format requests. The prompt extraction and flattening logic has been enhanced to work with chat completions requests while maintaining backward compatibility with existing request formats.

#### 3. **scripts/fetch-python-wrapper.sh**

A new script has been added to fetch the Python wrapper from `llm-d-kv-cache-manager`. This script enables consistent retrieval of the chat template rendering wrapper (`render_jinja_template_wrapper.py`), is used during the Docker build process, and supports version pinning and flexible output directory configuration.

### Remaining Tasks

From all requests and questions in the previous PR, all have been answered and the requested changes have been implemented. The only remaining item is:
- **Tests and documentation** - Benchmarks will be run on various workloads, followed by comprehensive documentation.

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -32,6 +32,14 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libzmq3-dev pkg-config python3-dev
 
+      - name: Configure CGO for Python
+        run: |
+          PYTHON_INCLUDE=$(python3 -c "import sysconfig; print(sysconfig.get_path('include'))")
+          echo "CPATH=${PYTHON_INCLUDE}:${CPATH}" >> $GITHUB_ENV
+          echo "CGO_ENABLED=1" >> $GITHUB_ENV
+          echo "CGO_CFLAGS=$(python3-config --cflags --embed)" >> $GITHUB_ENV
+          echo "CGO_LDFLAGS=$(python3-config --ldflags --embed)" >> $GITHUB_ENV
+
       - name: Set PKG_CONFIG_PATH
         run: echo "PKG_CONFIG_PATH=/usr/lib/pkgconfig" >> $GITHUB_ENV
 
@@ -43,6 +51,12 @@ jobs:
         with:
           version: 'v2.1.6'
           args: "--config=./.golangci.yml"
+        env:
+          CGO_ENABLED: ${{ env.CGO_ENABLED }}
+          CGO_CFLAGS: ${{ env.CGO_CFLAGS }}
+          CGO_LDFLAGS: ${{ env.CGO_LDFLAGS }}
+          CPATH: ${{ env.CPATH }}
+          PKG_CONFIG_PATH: ${{ env.PKG_CONFIG_PATH }}
 
       - name: Run make test
         shell: bash 


Dockerfile.epp
```diff
@@ -18,8 +20,15 @@ COPY go.sum go.sum
...
 COPY cmd/ cmd/
 COPY pkg/ pkg/
+COPY scripts/fetch-python-wrapper.sh scripts/fetch-python-wrapper.sh
 
-# HuggingFace tokenizer bindings
+# Fetch requirements only (install in runtime)
+RUN curl -L https://raw.githubusercontent.com/llm-d/llm-d-kv-cache-manager/v0.3.2/pkg/preprocessing/chat_completions/requirements.txt -o /workspace/requirements.txt
+
+# Prepare Python wrapper file for runtime
+RUN scripts/fetch-python-wrapper.sh v0.3.2 /workspace/llm-d-kv-cache-manager/pkg/preprocessing/chat_completions
```
elevran: nit: consider making `v0.3.2` a build arg with default.
Not blocking this PR.

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/llm-d/llm-d-kv-cache-manager/pkg/kvcache"
 	"github.com/llm-d/llm-d-kv-cache-manager/pkg/kvcache/kvevents"
+	chat_completions "github.com/llm-d/llm-d-kv-cache-manager/pkg/preprocessing/chat_completions"
```
elevran: I don't think `chat_completions` import rename is needed. Identical to package name so no-op

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -115,24 +116,36 @@ func (s *PrecisePrefixCacheScorer) WithName(name string) *PrecisePrefixCacheScor
...
+		logger.V(logutil.DEBUG).Info("Request is nil, skipping scoring")
 		return nil
 	}
 
-	prompt, err := getUserInput(request)
+	// Extract the flattened prompt from the request
+	logger.V(logutil.DEBUG).Info("Extracting prompt from request", 
+		"target_model", request.TargetModel,
+		"has_chat_completions", request.Body != nil && request.Body.ChatCompletions != nil,
+		"has_completions", request.Body != nil && request.Body.Completions != nil)
```
elevran, line 131: Q: is it valid to have both and if so, which should you use?

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -146,15 +159,102 @@ func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, _ *types.CycleStat
...
-		return "", fmt.Errorf("failed to marshal chat-completions messages: %w", err)
+	// For regular completions, use the prompt directly
+	if request.Body != nil && request.Body.Completions != nil {
+		prompt := request.Body.Completions.Prompt
+		logger.V(logutil.DEBUG).Info("Using completion prompt directly", "prompt_length", len(prompt))
+		return prompt, nil
 	}
-	return string(data), nil
+
+	// Fallback: try to extract prompt from request body if available
```
elevran: Q: was this introduced in response to a specific failure you observed or as defensive measure?
Reason for asking is that if in response to failure it might be worwhile to make this available in IGW for other components to use when the request object is created and not keep it private to scorer.

scripts/fetch-python-wrapper.sh
```diff
@@ -0,0 +1,46 @@
...
+# This script replicates the original Dockerfile logic:
+# 1. Creates a temporary directory
+# 2. Clones the repo into that directory
+# 3. Creates the output directory structure
+# 4. Copies the wrapper file to the output location
+# 5. Cleans up the temporary directory
+
+set -euo pipefail
+
+VERSION="${1:-v0.3.2}"
```
elevran: see earlier comment on makefile, where `v0.3.2` is hardcoded and not a build arg.

elevran:
thanks @guygir - looks good and much cleaner than previous, thanks for handling that.
Please see minor comments above. Your call if to respond or fix (with exception of import rename which should be fixed).
This can be merged after CICD passes - currently lint is failing. You may want to run `make lint` locally to verify your fix.

Dockerfile.epp
guygir: Updated Dockerfile to use argument KVCACHE_MANAGER_VERSION so the version is configurable and consistent.

pkg/plugins/scorer/precise_prefix_cache.go
guygir: Indeed no need for the alias.
The package actually declares itself as package preprocessing. I dropped the chat_completions rename entirely and now import it as preprocessing, matching the package’s own name.

pkg/plugins/scorer/precise_prefix_cache.go
guygir: Upstream guarantees only one body.
I log defensively and default to chat completions so that, if the scheduler or an external client ever regresses and sends both payloads (for some reason), we still follow the richer chat path and get a debug trace instead of silently misrouting the request. Might be an overkill.

pkg/plugins/scorer/precise_prefix_cache.go
guygir: Defensive measure, but not for something I personally encountered.
Its referencing legacy IGW releases (≤ gateway-api-inference-extension v0.5.x) that read prompts directly from raw prompt fields.

Dockerfile.epp
```diff
@@ -1,13 +1,17 @@
-# Build Stage: using Go 1.24 image
-FROM quay.io/projectquay/golang:1.24 AS builder
+## Minimal runtime Dockerfile (microdnf-only, no torch, wrapper in site-packages)
+# Build Stage: using Go 1.25 image
+FROM quay.io/projectquay/golang:1.25 AS builder
```
nirrozenbaum, line 3: please revert to 1.24 (we haven't updated golang to 1.25 yet).

pkg/common/common.go
```diff
@@ -1,5 +1,7 @@
 // Package common contains items common to both the
 // EPP/Inference-Scheduler and the Routing Sidecar
+//
+//revive:disable:var-naming
```
nirrozenbaum, line 4: why did we add the lint disable here?

Dockerfile.epp
```diff
@@ -18,8 +22,15 @@ COPY go.sum go.sum
 # Copy the go source
 COPY cmd/ cmd/
 COPY pkg/ pkg/
+COPY scripts/fetch-python-wrapper.sh scripts/fetch-python-wrapper.sh
+
+# Fetch requirements only (install in runtime)
+RUN curl -L https://raw.githubusercontent.com/llm-d/llm-d-kv-cache-manager/${KVCACHE_MANAGER_VERSION}/pkg/preprocessing/chat_completions/requirements.txt -o /workspace/requirements.txt
```
nirrozenbaum, line 28: we should do better with dependencies management. 
not a blocker, but we should probably handle that in the next iterations.

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -115,24 +117,36 @@ func (s *PrecisePrefixCacheScorer) WithName(name string) *PrecisePrefixCacheScor
 // Score scores the provided pod based on the KVCache index state.
 // The returned scores are normalized to a range of 0-1.
 func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, _ *types.CycleState, request *types.LLMRequest, pods []types.Pod) map[types.Pod]float64 {
-	loggerDebug := log.FromContext(ctx).WithName(s.typedName.String()).V(logutil.DEBUG)
+	logger := log.FromContext(ctx).WithName(s.typedName.String())
```
nirrozenbaum, line 120: suggestion for improvement, not a blocker:
I noticed all logging lines are sepcifying targetModelName as key value pair.
this can be done one time at the beginning of the function to avoid repeating it in all logger calls. e.g.,
```suggestion
	logger := log.FromContext(ctx).WithName(s.typedName.String()).WithValues("target_model", request.TargetModel)
```
then you won't need to specify in in the subsequent logger calls in this function

test/sidecar/utils/utils.go
```diff
@@ -15,6 +15,8 @@ limitations under the License.
 */
 
 // Package utils provides various utilities
+//
+//revive:disable:var-naming
```
nirrozenbaum, line 19: same question here

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -146,15 +160,114 @@ func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, _ *types.CycleStat
...
-	// must be chat-completions request at this point, return string of entire messages
-	data, err := json.Marshal(request.Body.ChatCompletions.Messages)
-	if err != nil {
-		return "", fmt.Errorf("failed to marshal chat-completions messages: %w", err)
+	// The upstream parser guarantees exactly one body is populated, but we defensively prioritize chat completions.
+	// If an unexpected dual payload slips through (parser regression/new client), log it and use chat semantics.
+	if request.Body != nil && request.Body.ChatCompletions != nil {
+		if request.Body.Completions != nil {
+			logger.V(logutil.DEBUG).Info("Both chat_completions and completions present; defaulting to chat completions", "target_model", request.TargetModel)
+		}
```
nirrozenbaum, line 181: This was already checked in line 172, duplicated check

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -146,15 +160,114 @@ func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, _ *types.CycleStat
...
+		renderReq.ChatTemplate = chatTemplate
+		renderReq.ChatTemplateKWArgs = chatTemplateKWArgs
+
+		// Render the template to get flattened prompt
+		logger.V(logutil.DEBUG).Info("Rendering chat template",
+			"conversations_count", len(renderReq.Conversations))
+		resp, err := processor.RenderChatTemplate(ctx, renderReq)
+		if err != nil {
+			logger.Error(err, "Failed to render chat template")
+			return "", fmt.Errorf("failed to render chat template: %w", err)
```
nirrozenbaum, line 237: if we return an error there is no need to call logger.Error just before that. this error will be logged by the upper layers.

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -146,15 +160,114 @@ func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, _ *types.CycleStat
...
+		}
+
+		if len(resp.RenderedChats) == 0 {
+			logger.Error(nil, "No rendered chat returned from template rendering")
+			return "", errors.New("no rendered chat returned from template rendering")
```
nirrozenbaum, line 242: ditto

Issue #402, opened 2025-10-29
Improved Error output when vLLM pod is not ready


elevran:
Spawned from: https://llm-d.slack.com/archives/C08SBNRRSBD/p1758306792046299

TL;DR - The sidecar seems to emit a 502 Bad Gateway when the decode node is not ready.

Is it possible to update the sidecars error to be a 503 with an error message? Other users may hit this issue and assume (reasonably so) that its an issue with their gateway configuration. If we improve the error handling it would, hopefully, help in future triage

copied from llm-d/llm-d-routing-sidecar#61

Issue #401, opened 2025-10-29
feat: Extend PD support to include when vLLM is configured to run in Data Parallel mode


elevran:
vLLM has added a feature called Data Parallel (DP). When DP is active vLLM runs many inference engines in the same Kubernetes pod. There are several different configurations all have multiple vLLM engines on one pod.

One of the DP configurations has multiple vLLMs riunning on one pod each with it's own inference port and metrics port.

When running Disambiguated Prefill/Decode (PD) we run with an llm-d-routing-sidecar in front of the vLLM engine in the decode pods. When running PD with DP enabled, we need the llm-d-routing-sidepar to listen on N ports and have each port work with one specific vLLM engine in the pod.

To simplify things:
 
* We can add a command line argument `--data-parallel-size=N`, where N indicates how many vLLM engines are running in parallel. This is the same command line argument used by vLLM. `N` must be greater than zero and for now less then or equal to eight, with a default of one.
* When --data-parallel-size is used with a value greater than one:
   * The sidecar will listen on `N` ports starting with the port specified by the `--port` command line argument
   * The sidecar will communicate with `N` vLLM engines all listening on separate ports starting with the  port number specified by the --vllm-port` coommad line argument

PR #432, opened 2025-11-06
Added support for Data Parallel in a Disagregated Prefil/Decode setup

shmuelk:
This PR enables the use of the vLLM Data Parallel feature with multiple public vLLM endpoints (ranks) in a situation where Disaggregated Prefill/Decode is also configured.

**Note:** To run the code in this PR using Istio as the gateway, a fix to Istio issue [57638](https://github.com/istio/istio/issues/57638). Istio PR [58068](https://github.com/istio/istio/pull/58068) is a potential partial solution, that will work with the workaround in this PR.

This PR also refactors error handling and tests in the Sidecar.

Fixes: #401
Fixes: #402

### DIFF
--- a/pkg/plugins/profile/pd_profile_handler.go
+++ b/pkg/plugins/profile/pd_profile_handler.go
@@ -9,13 +9,14 @@ import (
 	"net"
 	"strconv"
 
-	"github.com/llm-d/llm-d-inference-scheduler/pkg/common"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/plugins"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/scheduling/framework"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/scheduling/framework/plugins/multi/prefix"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/scheduling/types"
 	logutil "sigs.k8s.io/gateway-api-inference-extension/pkg/epp/util/logging"
+
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/common"
 )
 
 const (
@@ -165,7 +166,6 @@ func (h *PdProfileHandler) ProcessResults(_ context.Context, _ *types.CycleState
 	updatedResults := map[string]*types.ProfileRunResult{}
 
 	// Add decode profile to result
-
 	if h.primaryPort != "" {
 		// Data Parallel is active
 
--- a/pkg/plugins/profile/dp_profile_handler.go
+++ b/pkg/plugins/profile/dp_profile_handler.go
@@ -8,10 +8,11 @@ import (
 	"net"
 	"strconv"
 
-	"github.com/llm-d/llm-d-inference-scheduler/pkg/common"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/plugins"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/scheduling/framework"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/scheduling/types"
+
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/common"
 )
 
 const (


pkg/plugins/profile/pd_profile_handler.go
```diff
@@ -6,7 +6,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
+	"strconv"
 
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/common"
```
elevran: wrong import order?

pkg/plugins/profile/pd_profile_handler.go
```diff
@@ -51,19 +56,24 @@ func PdProfileHandlerFactory(name string, rawParameters json.RawMessage, _ plugi
 	}
 
 	return NewPdProfileHandler(parameters.PrefillProfile, parameters.DecodeProfile, parameters.PrefixPluginName,
-		parameters.Threshold, parameters.HashBlockSize).WithName(name), nil
+		parameters.Threshold, parameters.HashBlockSize, parameters.PrimaryPort).WithName(name), nil
```
elevran, line 60: Q: is it possible to determine the primary port from the InferencePool object instead of requiring it from user config (whcih could be missing, out of sync, ...)?

pkg/plugins/profile/pd_profile_handler.go
```diff
@@ -143,27 +154,48 @@ func (h *PdProfileHandler) Pick(ctx context.Context, cycleState *types.CycleStat
...
 	}
 	// otherwise, decode ran successfully
 
+	updatedResults := map[string]*types.ProfileRunResult{}
+
+	// Add decode profile to result
+
+	if h.primaryPort != "" {
+		// Data Parallel is active
+
```
elevran, line 171:
```suggestion
	// Add decode profile to result
	if h.primaryPort != "" { // Data Parallel is active
```

pkg/plugins/profile/pd_profile_handler.go
```diff
@@ -143,27 +154,48 @@ func (h *PdProfileHandler) Pick(ctx context.Context, cycleState *types.CycleStat
...
+		targetPod := decodeRunResults.TargetPods[0].GetPod()
+		request.Headers[common.DataParallelPodHeader] = net.JoinHostPort(targetPod.Address, targetPod.Port)
+
+		updatedResult := types.ProfileRunResult{
+			TargetPods: []types.Pod{},
+		}
+
+		for _, target := range decodeRunResults.TargetPods {
```
elevran, line 179: Q: I presume the pod with the primary port is also part of the pod array passed-in, would it be easier to swap that in for the result?
Just clarification, no change needed.

pkg/plugins/profile/pd_profile_handler.go
```diff
@@ -143,27 +154,48 @@ func (h *PdProfileHandler) Pick(ctx context.Context, cycleState *types.CycleStat
...
+			updatedPodInfo := target.GetPod().Clone()
```
elevran, line 180: Q: can this cange be applied in [pd_prerequest.go](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/plugins/pre-request/pd_prerequest.go) when we handle `PrefillHeaderHandler.PreRequest()`?
We extract the Prefill result and place it into a header, I wonder if we can mutate the scheduling result there too?

pkg/sidecar/proxy/errors.go
```diff
@@ -30,51 +30,51 @@ type errorResponse struct {
...
-	}
-
-	b, err := json.Marshal(er)
-	if err != nil {
-		return err
+		Message: "The decode node is not ready. Please check that the vLLM service is running and the port configuration is correct.",
+		Type:    "ServiceUnavailable",
+		Code:    http.StatusServiceUnavailable,
 	}
+	decoderServiceUnavailableResponseJSON, _ = json.Marshal(response)
```
elevran, line 42: you may need to make the linter happy with a comment since ignoring error return. Nothing we can do with the error reagrdless, but may be add a comment?

pkg/plugins/profile/pd_profile_handler.go
shmuelk: Unfortunately plugins do not have access to the InferencePool.

They have minimal access to the Datastore. In particular the ability to list the pods.

pkg/plugins/profile/pd_profile_handler.go
shmuelk: Not necessarily. If the "target pod" is only rank-1 for example, then there will be no pod in the result with the primary port

pkg/plugins/profile/pd_profile_handler.go
shmuelk: We can't. The target pods and endpoints are selected before the PreRequest extension point is called.

See the function preRequest at https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/pkg/epp/requestcontrol/director.go#L217

pkg/sidecar/proxy/errors.go
shmuelk: Lint passed with the code as is

pkg/plugins/profile/pd_profile_handler.go
elevran: Got it. Plugins are called after the result is set.
Do you know why the logic is that way and not plugins called before?

shmuelk: No.

PR #437, opened 2025-11-07
Remove code coverage from CI workflow

carlory:
See https://github.com/llm-d/llm-d-inference-scheduler/pull/295#issuecomment-3477978918

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -49,12 +49,6 @@ jobs:
         run: |
           make test
 
-      - name: Archive code coverage results
-        uses: actions/upload-artifact@v5
-        with:
-          name: code-coverage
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-pr-checks.yaml
```diff
@@ -49,8 +49,6 @@ jobs:
         run: |
           make test
 
-      - name: Archive code coverage results
-        uses: actions/upload-artifact@v5
         with:
           name: code-coverage
           path: coverage-*.out # Make sure to use the same file name you chose for the "-coverprofile" in the "make test" step
```
shmuelk: Shouldn't these lines be deleted as well?

shmuelk (changes requested):

.github/workflows/ci-pr-checks.yaml
carlory: removed.

shmuelk (approved):

PR #439, opened 2025-11-07
[WIP] SLO Aware Routing support

RishabhSaini:
go.mod: update GAIE to slo aware routing branch
pkg/plugins/scorer: Make it compile with the new interface

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	sigs.k8s.io/gateway-api-inference-extension v1.0.0
 )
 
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -24,6 +24,8 @@ require (
 	sigs.k8s.io/gateway-api-inference-extension v1.0.0
 )
 
+replace sigs.k8s.io/gateway-api-inference-extension => github.com/RishabhSaini/gateway-api-inference-extension v0.0.0-20251107192434-0e0db125e842
```
zdtsw: since we've got https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/1839 in, should we use that one instead

PR #442, opened 2025-11-10
test: End to End test for Data Parallel support

shmuelk:
This PR adds an additional test to the set of end to end tests. In particular it adds an end to end test of the Inference-Scheduler and Sidecar support of vLLM's Data Parallel feature.

Fixes: #441

### DIFF
--- a/test/utils/network.go
+++ b/test/utils/network.go
@@ -12,7 +12,9 @@ func GetFreePort() (string, error) {
 	if listener, err = net.Listen("tcp", ":0"); err == nil {
 		var port string
 		_, port, err = net.SplitHostPort(listener.Addr().String())
-		defer listener.Close()
+		defer func() {
+			_ = listener.Close()
+		}()
 		return port, err
 	}
 	return "", err


pkg/sidecar/proxy/proxy_test.go
```diff
@@ -61,14 +61,15 @@ var _ = Describe("Reverse Proxy", func() {
 				proxy := NewProxy("0", targetURL, cfg) // port 0 to automatically choose one that's available.
 
 				ctx, cancelFn := context.WithCancel(ctx)
-				defer cancelFn()
+				stoppedCh := make(chan bool)
```
elevran: nit: if you'll never send `false` down the channel then no need for a bool. Can use `chan struct{}` if only interested in read-readiness.

pkg/sidecar/proxy/proxy_test.go
```diff
@@ -116,16 +120,13 @@ var _ = Describe("Reverse Proxy", func() {
 	})
 
 	When("x-prefiller-url is present", func() {
-		var ctx context.Context
 		var decodeBackend *httptest.Server
 		var decodeHandler *mock.ChatCompletionHandler
 		var prefillBackend *httptest.Server
 		var prefillHandler *mock.ChatCompletionHandler
```
elevran, line 126: unrelated to this PR, curious why a `mock.ChatCompletionHandler` is used for `decode` and `prefill` handlers.

test/e2e/e2e_suite_test.go
```diff
@@ -180,6 +179,28 @@ func createEnvoy() {
 	testutils.CreateObjsFromYaml(testConfig, manifests)
 }
 
+func createInferencePool(targets int, toDelete bool) {
```
elevran: nit:
targets could mean pods or endpoint (ip:port) for the inference pool.
Based on usage below, seems to be neither, but rather a port count for the pool

test/e2e/e2e_test.go
```diff
@@ -25,6 +25,9 @@ const (
 	// simPdDeployment references the YAML file for the deployment
 	// running the vLLM simulator with PD
 	simPdDeployment = "./yaml/vllm-sim-pd.yaml"
+	// simDpDeployment references  the YAML file for the deployment
```
elevran: nit: idiomatic Go should capitalize the acronym as `DP` not `Dp` (same as `HTTP`, `TCP`, etc.)

test/utils/network.go
```diff
@@ -0,0 +1,22 @@
+// Package utils contains utilities for testing
+package utils
+
+import (
+	"net"
+)
+
+// GetFreePort finds a free port to listen on
+func GetFreePort() (string, error) {
```
elevran, line 9: nit: is there a cleaner version of this function (e.g., no err2, use defer)?

test/e2e/yaml/inference-pools.yaml
```diff
@@ -11,5 +11,4 @@ spec:
     kind: Service
     port:
       number: 9002
-  targetPorts:
-  - number: 8000
\ No newline at end of file
+  targetPorts: ${TARGET_PORTS}
```
elevran: missing newline at end of file

pkg/sidecar/proxy/data_parallel_test.go
```diff
@@ -0,0 +1,100 @@
...
+	When("configured with --data-parallel-size > 1", func() {
+		It("should create an extra proxy", func() {
+			_, ctx := ktesting.NewTestContext(GinkgoT())
+			ctx, cancel := context.WithCancel(ctx)
+			grp, ctx := errgroup.WithContext(ctx)
+
+			// proxy.startDataParallel starts listeners on the ports following
+			// the proxy's main port. To avoid problems, get a free port and
+			// tell the proxy that it is listening on that port minus one.
+			freePort, err := testutils.GetFreePort()
```
elevran, line 34: if you plan to open the port, there's a potential race condition.
AFAIK, the port can be in TIMED_WAIT or might have been open by another server in the meantime. Short window, but might need the user of the port (sidecar?) to retry

test/utils/network.go
```diff
@@ -0,0 +1,22 @@
...
+	var listener net.Listener
+	var err error
+	if listener, err = net.Listen("tcp", "127.0.0.1:0"); err == nil {
```
elevran:
```suggestion
	if listener, err = net.Listen("tcp", ":0"); err == nil {
```
localhost is implied.

pkg/sidecar/proxy/proxy_test.go
shmuelk: I didn't write that test. The mock handlers simply record that there was a call and return a HTTP Status of 200.

The test doesn't actually test if the K/V is transferred from the prefiller to the decoder.

pkg/sidecar/proxy/data_parallel_test.go
shmuelk: The port involved is the extra port opened on the sidecar for the /metrics calls, sent to the inference engine. The test doesn't actually send anything to that port. It does listen on that port....

elevran: it could fail in listener open, seems ok so far but maybe leave a note if it starts showing up in future runs?

test/utils/network.go
shmuelk: Done

test/e2e/yaml/inference-pools.yaml
shmuelk: fixed

test/utils/network.go
shmuelk: Done

test/e2e/e2e_test.go
shmuelk: Done

test/e2e/e2e_suite_test.go
shmuelk: Changed to numTargetPorts

pkg/sidecar/proxy/proxy_test.go
shmuelk: Done

elevran (approved):

PR #453, opened 2025-11-13
test: Enable running the end to end tests on K8S clusters other than Kind

shmuelk:
This PR enables the running of the End to End tests on clusters other than kind.

If the environment variable `K8S_CONTEXT` is set the end to end test code will not create a kind cluster for the test. Instead the cluster pointed to by the specified context in the current Kubernetes Config will be used. All created Kubernetes objects will be deleted at the end of the test. Additionally `kubectl port-forward` will be started to enable the tests to be run from "outside" the target cluster. The user running the tests, needs to be a cluster wide admin for the cluster in question.

In addition the following environment variables can "configure" the tests if needed:
- `E2E_PORT` - the local port used to send the inference requests to.
- `NAMESPACE` - the Kubernetes namespace to create all of the namespaced objects in.
- `EPP_IMAGE` - the full image name to be used for the llm-d-inference-scheduler
- `SIDECAR_IMAGE` - the full image name for the llm-d-routing-sidecar
- `VLLM_SIMULATOR_IMAGE` - the full name for the llm-d-inference-sim

### DIFF
--- a/docs/create_new_filter.md
+++ b/docs/create_new_filter.md
@@ -127,7 +127,7 @@ Once a filter is defined, it can be used to modify llm-d-inference-scheduler
  
 - Add the relevant import path (if defined outside this repository);
 - Add any desired configuration knobs (e.g., environment variables); and
-- Listing the new filter in the `LoadConfig()` function's `cfg.loadPluginInfo`
+- Listing the new filter in the `LoadConfigPhaseTwo()` function's `cfg.loadPluginInfo`
  list of available plugins.
 
 In the case of the llm-d-inference-scheduler, filters can be hooked into the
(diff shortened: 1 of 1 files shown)


hdefazio (approved):

docs/create_new_filter.md
elevran, line 1: nit: this seems like an unrelated change?

test/config/prefix_cache_mode_test.go
elevran, line 1: nit: unrelated?

shmuelk: You're fundamentally correct. But upstream, the function LoadConfig no longer exists causing lint to fail.

docs/create_new_filter.md
shmuelk: You're fundamentally correct. But upstream, the function LoadConfig no longer exists causing lint to fail in the file `test/config/prefix_cache_mode_test.go`. I then searched for LoadConfig and updated all references.

PR #456, opened 2025-11-17
Add SGLang Connector for Prefill/Decode Disaggregation (migrated from llm-d-routing-sidecar#64)

bongwoobak:
## Summary

This PR is the same work as:
https://github.com/llm-d/llm-d-routing-sidecar/pull/64

It is moved here because the reviewer asked for the PR to be relocated to this repository.

## Changes

- Added connector_sglang.go implementing P/D disaggregation  
- Integrated bootstrap configuration (host, port, room)  
- Updated cmd/llm-d-routing-sidecar/main.go and internal/proxy/proxy.go  

## Features

- Room-based communication for coordinating prefill/decode  
- Configurable bootstrap via env `SGLANG_BOOTSTRAP_PORT` (default: 8668)  
- Prefill requests are sent asynchronously, decode requests are sent synchronously and processed upon receiving the decode response  

## Test

- Tested with SGLang prefill/decode services  
- Confirmed asynchronous prefill & synchronous decode execution  
- Successfully tested in a Kubernetes cluster with AMD MI250 GPUs  
- Tested successfully on gateway-api-inference-extension v1.1.0 with inference-scheduler v0.3.2

### DIFF
--- a/cmd/pd-sidecar/main.go
+++ b/cmd/pd-sidecar/main.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"net/url"
 	"os"
+	"strings"
 
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"


pkg/sidecar/proxy/connector_sglang.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (s *Server) generateSGLangRoomID() int64 {
+	return time.Now().UnixNano() + int64(rand.Intn(1000))
+}
+
+func (s *Server) getBootstrapHost(prefillHostPort string) (string, int) {
+	// Extract hostname from prefill host
+	parts := strings.Split(prefillHostPort, ":")
+	hostname := parts[0]
+	// Get bootstrap port from environment variable
```
shmuelk: This code should be done once. Perhaps in an init() function, with the port stored in a local global variable.

pkg/sidecar/proxy/connector_sglang.go
```diff
@@ -0,0 +1,151 @@
...
+
+	if err != nil {
+		if err := errorJSONInvalid(err, w); err != nil {
+			s.logger.Error(err, "failed to send error response to client")
+		}
+		return
+	}
+
+	// Validate prefill host
+	if prefillPodHostPort == "" {
```
shmuelk: There is no need to check for this. The code in proxy.go, checks for this situation and if it occurs simply sends the request to the decode node.

pkg/sidecar/proxy/connector_sglang.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (s *Server) sendSGLangConcurrentRequests(w http.ResponseWriter, r *http.Request, body []byte, prefillHost string) {
+	// Create separate requests for prefill and decode
+	prefillReq := cloneWithJSONBody(r, body)
+	decodeReq := cloneWithJSONBody(r, body)
+
+	// Send prefill request asynchronously
+	go func() {
+		prefillHandler, err := s.prefillerProxyHandler(prefillHost)
```
shmuelk: This should be done outside the GO routine.

bongwoobak: Fixed. Moved handler creation outside.

Thank you for the thorough review!

pkg/sidecar/proxy/connector_sglang.go
bongwoobak: Removed. When prefill host is empty, it will be treated as non-PD mode.

pkg/sidecar/proxy/connector_sglang.go
bongwoobak: I've moved the bootstrap port parsing to init() as suggested. 

Because the bootstrap host is derived from the prefill host and varies per request,
only the port is initialized once in init(). The hostname extraction remains 
in getBootstrapHost() as it depends on the dynamic prefillHostPort parameter.

pkg/sidecar/proxy/connector_sglang.go
```diff
@@ -0,0 +1,155 @@
...
+	for k, v := range requestData {
+		modifiedRequest[k] = v
+	}
+
+	// Generate bootstrap host from prefill host
+	bootstrapHost, bootstrapPort := s.getBootstrapHost(prefillHostPort)
+
+	// Add bootstrap information
+	modifiedRequest[requestFieldBootstrapHost] = bootstrapHost
+	modifiedRequest[requestFieldBootstrapPort] = bootstrapPort
```
shmuelk: Why not simply set this to `sglangBootstrapPort` ?

pkg/sidecar/proxy/connector_sglang.go
```diff
@@ -0,0 +1,155 @@
...
+func (s *Server) generateSGLangRoomID() int64 {
+	return time.Now().UnixNano() + int64(rand.Intn(1000))
+}
+
+func (s *Server) getBootstrapHost(prefillHostPort string) (string, int) {
+	// Extract hostname from prefill host
+	parts := strings.Split(prefillHostPort, ":")
+	hostname := parts[0]
+
+	return hostname, sglangBootstrapPort
```
shmuelk: Why return two values as the second value, the port, is constant through the execution of the server?

bongwoobak: You're right, I completely missed this. That was silly of me. 
I've fixed both the function signature and its usage as you suggested.

Thank you.

cmd/pd-sidecar/main.go
```diff
@@ -28,11 +28,20 @@ import (
...
+		proxy.ConnectorSGLang,
+	}
+)
+
 func main() {
 	port := flag.String("port", "8000", "the port the sidecar is listening on")
 	vLLMPort := flag.String("vllm-port", "8001", "the port vLLM is listening on")
 	vLLMDataParallelSize := flag.Int("data-parallel-size", 1, "the vLLM DATA-PARALLEL-SIZE value")
-	connector := flag.String("connector", "nixlv2", "the P/D connector being used. Either nixl, nixlv2 or lmcache")
+	connector := flag.String("connector", "nixlv2", "the P/D connector being used. Either nixl, nixlv2, lmcache, or sglang")
```
hhk7734:
```suggestion
	connector := flag.String("connector", proxy.ConnectorNIXLV2, "the P/D connector being used. Supported: "+strings.Join(supportedConnectors, ", "))
```

It may be a good idea to update it this way.
And since `nixl` is no longer used, we should remove it from the list.

bongwoobak: That would be better. Thank you

cmd/pd-sidecar/main.go
```diff
@@ -57,8 +66,16 @@ func main() {
...
+	// Validate connector
+	isValidConnector := false
+	for _, validConnector := range supportedConnectors {
+		if *connector == validConnector {
+			isValidConnector = true
+			break
+		}
+	}
+	if !isValidConnector {
+		logger.Info("Error: --connector must be one of: nixlv2, lmcache, sglang")
```
hhk7734:
```suggestion
		logger.Info("Error: --connector must be one of: " + strings.Join(supportedConnectors, ", "))
```

Here too.

shmuelk:
@bongwoobak You are missing an import for "strings. Please add it.

bongwoobak:
@shmuelk I’ve added the import string and ran another test just to be sure. Thank you.

PR #467, opened 2025-11-20
Bump kv-cache-manager to v0.4.0-rc2

pierDipi:
Bump + fix breaking changes

### DIFF
--- a/pkg/plugins/scorer/precise_prefix_cache.go
+++ b/pkg/plugins/scorer/precise_prefix_cache.go
@@ -164,10 +164,9 @@ func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, _ *types.CycleStat
 // For regular completions, it uses the prompt directly.
 func (s *PrecisePrefixCacheScorer) getScores(ctx context.Context, request *types.LLMRequest) (map[string]float64, error) {
 	logger := log.FromContext(ctx).WithName(s.typedName.String())
-	debugLogger := logger.V(logutil.DEBUG)
 	traceLogger := logger.V(logutil.TRACE)
 
-	debugLogger.Info("Getting scores",
+	traceLogger.Info("Getting scores",
 		"target_model", request.TargetModel,
 		"has_chat_completions", request.Body != nil && request.Body.ChatCompletions != nil,
 		"has_completions", request.Body != nil && request.Body.Completions != nil)


pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -41,14 +41,23 @@ var _ framework.Scorer = &PrecisePrefixCacheScorer{}
...
 
 	// read hugging face token from environment variable if set
-	if token := os.Getenv("HF_TOKEN"); token != "" {
-		parameters.IndexerConfig.TokenizersPoolConfig.HuggingFaceToken = token
+	if token := os.Getenv("HF_TOKEN"); token != "" &&
+		parameters.IndexerConfig != nil &&
+		parameters.IndexerConfig.TokenizersPoolConfig != nil &&
+		parameters.IndexerConfig.TokenizersPoolConfig.HFTokenizerConfig != nil {
+		parameters.IndexerConfig.TokenizersPoolConfig.HFTokenizerConfig.HuggingFaceToken = token
 	}
```
nirrozenbaum, line 61: these lines look like internal implementation details of kvcache manager code. 
can we move this to kvcache repo?
I'd expect the call `indexerConfig, err := kvcache.NewDefaultConfig()` in L45 to initialize that internally.

vMaroon: The kvcache library defines a configuration with default values. This code piece is that of a user, permitting configuration through the env-var `HF_TOKEN`. It is arguable that this is a special env-var that is widely accepted but generally in the kvcache library we attempt to contain all configuration in the referenced structure, leaving customized UX to the users.

nirrozenbaum: right. it's a user defined env var (with the user's token). 
HF_TOKEN as env var is very acceptable.

the question I was trying to answer is why do we read that env var here and not in kvcache code.
this part looks very not natural, the scorer factory function writes to an internal config of the kvcache indexer parameters.
I was expecting to have the `os.getenv` call inside `kvcache.NewDefaultConfig()`.

pierDipi: Since this discussion is applicable to the current setup, should we discuss it separately?

pierDipi: I agree that kv cache manager library can do this since the use can still opt-out from the default env-var injection by not using the NewDefaultConfig function if they want to have different defaults

vMaroon: Let's track this in a new issue, I think this should not be a blocker.

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -161,20 +162,23 @@ func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, _ *types.CycleStat
...
 // For chat completions, it renders the messages using the model's chat template.
 // For regular completions, it uses the prompt directly.
-func (s *PrecisePrefixCacheScorer) extractPrompt(ctx context.Context, request *types.LLMRequest) (string, error) {
-	traceLogger := log.FromContext(ctx).V(logutil.TRACE).WithName(s.typedName.String())
+func (s *PrecisePrefixCacheScorer) getScores(ctx context.Context, request *types.LLMRequest) (map[string]float64, error) {
+	logger := log.FromContext(ctx).WithName(s.typedName.String())
+	debugLogger := logger.V(logutil.DEBUG)
+	traceLogger := logger.V(logutil.TRACE)
+
+	debugLogger.Info("Getting scores",
```
vMaroon: This should be trace, otherwise will be too spammy. TBH can be dropped.

pierDipi: Done

pkg/plugins/scorer/precise_prefix_cache.go
nirrozenbaum: not a blocker 👍

PR #474, opened 2025-11-21
feat(allowlist): support both v1 and v1alpha2 InferencePool APIs with flag

googs1025:
add new `pool-group` flag to choose which inferencepool to support

fix: https://github.com/llm-d/llm-d-inference-scheduler/issues/462

### DIFF
--- a/pkg/sidecar/proxy/allowlist.go
+++ b/pkg/sidecar/proxy/allowlist.go
@@ -38,11 +38,15 @@ import (
 )
 
 const (
-	inferencePoolGroup    = "inference.networking.x-k8s.io"
-	inferencePoolVersion  = "v1alpha2"
 	inferencePoolResource = "inferencepools"
 	resyncPeriod          = 30 * time.Second
 )
+
+// InferencePool API group to version mapping
+var inferencePoolGroupToVersion = map[string]string{
+	DefaultPoolGroup: "v1",
+	LegacyPoolGroup:  "v1alpha2",
+}
 
 // AllowlistValidator manages allowed prefill targets based on InferencePool resources
 type AllowlistValidator struct {
@@ -51,6 +55,8 @@ type AllowlistValidator struct {
 	namespace     string
 	poolName      string
 	enabled       bool
+
+	gvr schema.GroupVersionResource // detected GVR
 
 	// allowedTargets maps hostport -> bool for allowed prefill targets
 	allowedTargets   set.Set[string]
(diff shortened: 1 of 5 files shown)


pkg/sidecar/proxy/allowlist.go
```diff
@@ -87,44 +107,79 @@ func NewAllowlistValidator(enabled bool, namespace string, poolName string) (*Al
...
 }
 
+func isGVRSupported(discoveryClient discovery.DiscoveryInterface, gvr schema.GroupVersionResource) (bool, error) {
+	apiResourceList, err := discoveryClient.ServerResourcesForGroupVersion(gvr.GroupVersion().String())
+	if err != nil {
+		// If the group/version doesn't exist, Kubernetes returns a "NotFound" error
+		if errors.IsNotFound(err) {
+			return false, nil // GroupVersion not supported
+		}
+		return false, fmt.Errorf("failed to discover resources for %s: %w", gvr.GroupVersion(), err)
```
pierDipi: nit: as the function is generic and doesn't necessarily assume InfPool resource
```suggestion
		return false, fmt.Errorf("failed to discover resources for %s: %w", gvr.String(), err)
```

pierDipi (approved):
/lgtm

pkg/sidecar/proxy/allowlist.go
googs1025: done

shmuelk:
@googs1025 If I understand this PR correctly, it is trying to figure which version of the InferencePool CRD is installed.

What happens if both CRDs are installed? I believe `make env-dev-kind` installs both versions of the InferencePool CRD. It actually uses the v1 version, but both CRDs are installed.

pkg/sidecar/proxy/allowlist.go
```diff
@@ -38,12 +40,28 @@ import (
...
+//
+// 1. inference.networking.k8s.io/v1         ← Preferred (new official API group)
+// 2. inference.networking.x-k8s.io/v1alpha2 ← Fallback (legacy experimental API group)
+
+// The validator automatically detects which API is available by using the Kubernetes
+// discovery API, selecting the first supported GVR in this list.
+//
+// This approach aligns with upstream Ingress Gateway (IGW) behavior, which also supports
+// both API versions concurrently (see issue #462).
+var candidateGVRs = []schema.GroupVersionResource{
```
zdtsw: `isGVRSupported()` is generic function, should `candidateGVRs` be more specific as it is only on `inferencepools` for now

googs1025: change to `inferencePoolGVRCandidates`

pkg/sidecar/proxy/allowlist.go
```diff
@@ -99,32 +140,45 @@ func NewAllowlistValidator(enabled bool, namespace string, poolName string) (*Al
...
 	av.logger = log.FromContext(ctx).WithName("allowlist-validator")
-	av.logger.Info("starting SSRF protection allowlist validator", "namespace", av.namespace, "poolName", av.poolName)
-
-	gvr := schema.GroupVersionResource{
-		Group:    inferencePoolGroup,
-		Version:  inferencePoolVersion,
-		Resource: inferencePoolResource,
-	}
+	av.logger.Info("starting SSRF protection allowlist validator",
+		"namespace", av.namespace, "poolName", av.poolName, "gvr", av.gvr.String())
```
shmuelk, line 138: av.gvr is never set

shmuelk (changes requested):

pkg/sidecar/proxy/allowlist.go
googs1025: oh... my bad. done

Issue #470, opened 2025-11-20
Installation of build dependencies via Makefile is broken


elevran:
**What happened**:
Running make on Linux.
```sh
$ make test-unit
Installing dependencies with apt...
Reading package lists... Done
E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
E: Unable to lock directory /var/lib/apt/lists/
W: Problem unlinking the file /var/cache/apt/pkgcache.bin - RemoveCaches (13: Permission denied)
W: Problem unlinking the file /var/cache/apt/srcpkgcache.bin - RemoveCaches (13: Permission denied)
make: *** [Makefile:461: install-dependencies] Error 100
```

**What you expected to happen**:
dependencies install cleanly (or message to install manually).

**How to reproduce it (as minimally and precisely as possible)**:
As above. Run on system without dependencies (python, zmq, g++) preinstalled.

**Anything else we need to know?**:
Running as non root user by default (e.g., `sudo` required). 

**Environment**:
Linux/Ubuntu (WSL)

elevran:
cc @shmuelk @vMaroon

elevran:
@acardace SGTM.
Would you like to take a stab at the container image for unit tests?

PR #481, opened 2025-11-24
fix: github action missing Trivy scan on sidecar image

zdtsw:
- Trivy scan should be done on both images
- move "make build" before "make test" to have quick exit
- use "make" target to install k-v cache dependency
- use Github Action than download lychee binary

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -27,10 +27,8 @@ jobs:
           go-version: "${{ env.GO_VERSION }}"
           cache-dependency-path: ./go.sum
 
-      - name: Install kvcache dependencies
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y libzmq3-dev pkg-config python3-dev
+      - name: Install dependencies
+        run: sudo make install-dependencies
 
       - name: Configure CGO for Python
         run: |
@@ -58,12 +56,12 @@ jobs:
           CPATH: ${{ env.CPATH }}
           PKG_CONFIG_PATH: ${{ env.PKG_CONFIG_PATH }}
 
-      - name: Run make test
-        shell: bash 
-        run: |
-          make test
-
       - name: Run make build
         shell: bash
         run: |
           make build
+
+      - name: Run make test
+        shell: bash
+        run: |
+          make test
--- a/.github/workflows/ci-release.yaml
+++ b/.github/workflows/ci-release.yaml
@@ -14,11 +14,12 @@ jobs:
       - name: Checkout source
         uses: actions/checkout@v6
 
-      - name: Set project name from repository
+      - name: Set image names
         id: version
         run: |
           repo="${GITHUB_REPOSITORY##*/}"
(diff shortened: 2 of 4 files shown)


elevran:
@zdtsw DCO is missing, Please see https://github.com/llm-d/llm-d-inference-scheduler/pull/481/checks?check_run_id=56199228166

.github/workflows/ci-pr-checks.yaml
```diff
@@ -27,10 +27,8 @@ jobs:
           go-version: "${{ env.GO_VERSION }}"
           cache-dependency-path: ./go.sum
 
-      - name: Install kvcache dependencies
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y libzmq3-dev pkg-config python3-dev
+      - name: Install dependencies
+        run: sudo make install-dependencies
```
elevran, line 31: Q: 
the makefile installs Python dependencies outside the `install-dependencies` target (it is handled in L64-L92). 
in the previous version it was being installed explicitly via apt-get. 

- does Python get installed and set up correctly by calling make?
- if not, can you please move the Python install to the same make target?

zdtsw: you are right.
the current make target "install-dependencies" does not install python3-dev
but i am wondering if it really need it or not.  see, runs from https://github.com/llm-d/llm-d-inference-scheduler/actions/runs/19634069811
"ubuntu-latest" runner itself does have python-is-python3 pre-installed.
https://github.com/llm-d/llm-d-inference-scheduler/actions/runs/19634069811

but on the safe side, i think it is better have python3-dev included in the  "install-dependencies"

elevran: What we need are the Python header files so we can use them (via CGO).
The makefile in local dev-env would would not install python and only print a message (`Please install with: 'sudo apt install python$(PYTHON_VERSION)-dev'`) if missing.
If python-dev is always preinstalled on `ubuntu-latest` runner then we're good to go. Perhaps add an explicit step to check if the assumption is valid? Or maybe the makefile would already fail if python headers are not available?

cc @vMaroon for python dependencies.

zdtsw: After some debug, I think I know the issue here
in Github Ubuntu-latest runner, python3-dev is already  pre-installed, and Python.h is there, see https://github.com/zdtsw/llm-d-inference-scheduler/actions/runs/19637556798/job/56232027808
so to have a "apt-get install" is not necessary here.
But for the local dev it does require python3-dev/python3-devel, which now is added in the "install-dependencies"
it might be a duplicated in the Github Action case to have both pre-install and then calling make again. unless we wanna break python3-dev/python3-devel out of "install-dependencies"

zdtsw:
> > if we need more confirmation regarding python header file, should I split this PR into two so get the "missing Trivy scan" out first?
> 
> No further confirmation is needed and not need to split PR. I think it is safest to go back to what you originally had and remove Python from Makefile's install-dependencies target. We have #470 to track Makefile issue on dependencies. /unhold

Thanks,  I've reverted the changes for python3-dev and updated the commit message

PR #484, opened 2025-11-24
Use kv-cache-manager based on Go mod version instead of hardcoded

pierDipi:
Currently the KVCACHE_MANAGER_VERSION is out of sync with the go mod which could cause bugs.


vMaroon:
Thanks @pierDipi - I tested locally 👍🏻 
Hopefully in v0.5 we will break all these dependencies.

/lgtm
/approve

PR #488, opened 2025-11-25
Enhance Dockerfile to manage tokenizers binary download

vivek8123:
Updated Dockerfile to handle missing prebuilt binaries for tokenizers by attempting to download them first, and falling back to building from source if not found.

### DIFF
--- a/Dockerfile.epp
+++ b/Dockerfile.epp
@@ -34,7 +34,22 @@ RUN scripts/fetch-python-wrapper.sh ${KVCACHE_MANAGER_VERSION} /workspace/llm-d-
 RUN mkdir -p lib
 # Ensure that the RELEASE_VERSION matches the one used in the imported llm-d-kv-cache-manager version
 ARG RELEASE_VERSION=v1.22.1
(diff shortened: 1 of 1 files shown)


Dockerfile.epp
```diff
@@ -34,7 +34,22 @@ RUN scripts/fetch-python-wrapper.sh ${KVCACHE_MANAGER_VERSION} /workspace/llm-d-
 RUN mkdir -p lib
 # Ensure that the RELEASE_VERSION matches the one used in the imported llm-d-kv-cache-manager version
 ARG RELEASE_VERSION=v1.22.1
-RUN curl -L https://github.com/daulet/tokenizers/releases/download/${RELEASE_VERSION}/libtokenizers.${TARGETOS}-${TARGETARCH}.tar.gz | tar -xz -C lib
+# Handle ppc64le or missing prebuilt binaries
```
zdtsw, line 37: isnt the avaiable [release](https://github.com/daulet/tokenizers/releases/tag/v1.22.1) for v1.22.1 only for arm64 and amd64 ?
so you should uplift RELEASE_VERSION to v1.24.0 instead adding build-from-source

zdtsw: @vMaroon do we have plan uplift to 1.24.0 https://github.com/llm-d/llm-d-kv-cache-manager/blob/main/go.mod#L8

PR #497, opened 2025-11-28
Error if PYTHON_CONFIG is empty

### LINKED ISSUE #496: Unguarded shell invocation in Makefile
**What happened**:
```sh
$ make test-unit
bash: --: invalid option
Usage:  bash [GNU long option] [option] ...
        bash [GNU long option] [option] script-file ...
```

no bash error.

On system without Python installed, run make.

**Environment**:
NA

elevran:
Fix #496 

Previously, if PYTHON_CONFIG was empty, makefile would invoke shell with `--cflags`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ ifeq ($(PYTHON_CONFIG),)
 endif
 
 ifeq ($(PYTHON_CONFIG),)
-    $(error Python configuration tool not found or invalid)
+    $(error ${PYTHON_ERROR})
 endif
 
 ifeq ($(shell $(PYTHON_CONFIG) --cflags 2>/dev/null),)


Makefile
```diff
@@ -81,14 +81,17 @@ ifeq ($(PYTHON_CONFIG),)
 	endif
 endif
 
-ifneq ($(shell $(PYTHON_CONFIG) --cflags 2>/dev/null),)
-    PYTHON_CFLAGS := $(shell $(PYTHON_CONFIG) --cflags)
-    # Use --ldflags --embed to get all necessary flags for linking
-    PYTHON_LDFLAGS := $(shell $(PYTHON_CONFIG) --ldflags --embed)
-else
-    $(error $(PYTHON_ERROR))
```
shmuelk, line 89: You have removed the use of the platform specific error message, set somewhat earlier in the code. Those messages told the user how to install the Python config utility.

If we don't want to isse a message of that nature, please remove the code that sets the variable PYTHON_ERROR

shmuelk (changes requested):

Makefile
elevran: Restored.

PR #498, opened 2025-11-30
fix: Fixes for Data Parallel support when also running with Prefix Disaggregation

shmuelk:
This PR fixes issues with the ability to run Data Parallel (DP) and Prefix Dis-aggregation (PD) together.

In particular it better enables the case where the special DP header isn't used.

### DIFF
--- a/pkg/sidecar/proxy/chat_completions.go
+++ b/pkg/sidecar/proxy/chat_completions.go
@@ -57,7 +57,7 @@ func (s *Server) chatCompletionsHandler(w http.ResponseWriter, r *http.Request) 
 	if len(prefillHostPort) == 0 {
 		s.logger.V(4).Info("skip disaggregated prefill")
 
-		if s.forwardDataParallel && !s.dataParallelHandler(w, r) {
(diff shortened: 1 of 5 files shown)


pkg/sidecar/proxy/chat_completions.go
```diff
@@ -57,7 +57,7 @@ func (s *Server) chatCompletionsHandler(w http.ResponseWriter, r *http.Request)
 	if len(prefillHostPort) == 0 {
 		s.logger.V(4).Info("skip disaggregated prefill")
 
-		if s.forwardDataParallel && !s.dataParallelHandler(w, r) {
+		if !s.forwardDataParallel || !s.dataParallelHandler(w, r) {
```
googs1025, line 60: This seems to be correct 🤔 ; it ensures that decode requests are handled properly when Data Parallel is disabled or not applicable.

Additionally, should we also modify the deprecated connect function simultaneously? 🤔

nirrozenbaum:
@shmuelk can you please rebase?

PR #505, opened 2025-12-04
test: add precise_prefix_cache_test

### LINKED ISSUE #218: add unit tests for kvcache aware scorer

evacchi:
Fixes #218. Partially based on initial work in #311 and @vMaroon's advice at https://github.com/llm-d/llm-d-inference-scheduler/pull/311#pullrequestreview-3153091022 (if I have understood it correctly).

~~Opening as draft to see if I am on the right track, early feedback is welcome.~~

~~At this time, this is successfully testing a couple of edge cases (nil request, empty request body), just to make sure that the test case run correctly; then there is one proper test case ("test normalized scores with different kv-block hits"), **failing at this time**~~

1. I have lifted pkg/plugins/scorer/testdata/test-model from https://github.com/llm-d/llm-d-kv-cache-manager/tree/98db134628e59dfe527c40000f8d3829c002d99b/pkg/tokenization/testdata/test-model
2. applying the templates requires Python with the right packages, so we have to setup PYTHONPATH correctly; thus I have also lifted a few sections from https://github.com/llm-d/llm-d-kv-cache-manager/blob/98db134628e59dfe527c40000f8d3829c002d99b/Makefile as well

~~The reason the test fails is all scores are zero, indicating there might be still some issues initializing the tokenizer. I have to investigate.~~

Notes:
- Because of (2) this is sort-of an integration test. I can mark this as a "long" test (`if testing.Short() { t.Skip() }`) in the same fashion of other tokenization tests in the cache manager.
- Also notice that there is apparently [ongoing work](https://github.com/llm-d/llm-d-kv-cache-manager/issues/194) to refactor `kvcache.Indexer` into an interface, and that would probably simplify this; e.g. by allowing us to inject a mock tokenizer
- Finally, a misconfigured tokenizer can easily hang; from a cursory glance, I'd say it's because at least _some_ errors fail to bubble up from the task queue; maybe I'll follow up with some fixes around that

### DIFF
--- a/pkg/plugins/scorer/precise_prefix_cache.go
+++ b/pkg/plugins/scorer/precise_prefix_cache.go
@@ -8,7 +8,6 @@ import (
 	"os"
 
 	"github.com/llm-d/llm-d-kv-cache-manager/pkg/kvcache"
-	"github.com/llm-d/llm-d-kv-cache-manager/pkg/kvcache/kvblock"
 	"github.com/llm-d/llm-d-kv-cache-manager/pkg/kvcache/kvevents"
 	preprocessing "github.com/llm-d/llm-d-kv-cache-manager/pkg/preprocessing/chat_completions"
 	"sigs.k8s.io/controller-runtime/pkg/log"
@@ -122,11 +121,6 @@ func (s *PrecisePrefixCacheScorer) TypedName() plugins.TypedName {
 func (s *PrecisePrefixCacheScorer) WithName(name string) *PrecisePrefixCacheScorer {
 	s.typedName.Name = name
 	return s
-}
-
-// KVBlockIndex returns the underlying kvblock.Index for testing purposes.
-func (s *PrecisePrefixCacheScorer) KVBlockIndex() kvblock.Index {
-	return s.kvCacheIndexer.KVBlockIndex()
 }
 
 // Score scores the provided pod based on the KVCache index state.
--- a/pkg/plugins/scorer/precise_prefix_cache_test.go
+++ b/pkg/plugins/scorer/precise_prefix_cache_test.go
@@ -1,4 +1,4 @@
-package scorer_test
+package scorer
 
 import (
 	"os"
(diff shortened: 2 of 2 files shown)


evacchi:
notes: 
- the tokenizer contains "typos" and CI complains about it; I could either add excludes to CI, or use an alternative approach
- the local tokenizer is now working but `db.TokensToKVBlockKeys()` returns nil 👀

evacchi:
this is now working e2e for Completions. It still hangs for ChatCompletions, I will investigate later. This is still open for comments.

again, hangs are 90% due to errors not bubbling up from the tasks queues of the tokenizer; the response queue should receive errors as well as completed tasks. I'll follow up with an issue and/or PR there once this is done.

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -123,6 +124,11 @@ func (s *PrecisePrefixCacheScorer) WithName(name string) *PrecisePrefixCacheScor
 	return s
 }
 
+// KVBlockIndex returns the underlying kvblock.Index for testing purposes.
+func (s *PrecisePrefixCacheScorer) KVBlockIndex() kvblock.Index {
+	return s.kvCacheIndexer.KVBlockIndex()
```
evacchi: note: this is only necessary because the test follows the pattern `package score_test`. If the test were under `package score` then we could just access `s.kvCacheIndexer`.

vMaroon: I think going for `package score` and adding a `nolint` directive wherever necessary is cleaner. @elevran what do you think?

In general this scorer is not expected to understand the internals of the kvcache library, it is only concerned with the API exposed by the indexer.

evacchi: moved to `package score` and removed the public method. I don't think the linter is complaining, I get an unrelated issue on `test/e2e/e2e_suite_test.go`

PR #506, opened 2025-12-05
Add otel tracing instrumentation

sallyom:
This PR satisfies the llm-d distributed tracing proposal https://github.com/llm-d/llm-d/pull/119 implementation plan.

### DIFF
--- a/pkg/sidecar/proxy/chat_completions.go
+++ b/pkg/sidecar/proxy/chat_completions.go
@@ -54,9 +54,14 @@ func (s *Server) chatCompletionsHandler(w http.ResponseWriter, r *http.Request) 
 	ctx = context.WithValue(ctx, requestStartTimeKey, requestStart)
 	r = r.WithContext(ctx)
 
+	// Set span attributes with safe defaults for nil values
+	requestPath := ""
+	if r.URL != nil {
+		requestPath = r.URL.Path
+	}
 	span.SetAttributes(
 		attribute.String("llm_d.pd_proxy.connector", s.config.Connector),
-		attribute.String("llm_d.pd_proxy.request_path", r.URL.Path),
+		attribute.String("llm_d.pd_proxy.request_path", requestPath),
 	)
 
 	var prefillHostPorts []string
--- a/pkg/sidecar/proxy/proxy_helpers.go
+++ b/pkg/sidecar/proxy/proxy_helpers.go
@@ -32,7 +32,11 @@ func (s *Server) startHTTP(ctx context.Context, cert *tls.Certificate) error {
 	// Wrap handler with OpenTelemetry middleware to extract trace context from incoming requests
 	handler := otelhttp.NewHandler(s.handler, "llm-d-pd-proxy",
 		otelhttp.WithSpanNameFormatter(func(_ string, r *http.Request) string {
-			return "llm_d.pd_proxy." + r.Method + " " + r.URL.Path
+			path := ""
+			if r.URL != nil {
+				path = r.URL.Path
+			}
+			return "llm_d.pd_proxy." + r.Method + " " + path
 		}),
 	)
 


Dockerfile.localdev.epp
```diff
@@ -0,0 +1,110 @@
+# Build Stage: using Go 1.25 image
+FROM quay.io/projectquay/golang:1.25 AS builder
```
zdtsw: should we keep using 1.24 to align with other places?

Dockerfile.localdev.epp
```diff
@@ -0,0 +1,110 @@
...
+WORKDIR /workspace
+
+# Copy the Go Modules manifests
+COPY llm-d-inference-scheduler/go.mod go.mod
+COPY llm-d-inference-scheduler/go.sum go.sum
+
+# Copy the go source
+COPY llm-d-inference-scheduler/cmd/ cmd/
+COPY llm-d-inference-scheduler/pkg/ pkg/
+COPY llm-d-inference-scheduler/test/ test/
```
zdtsw: are the "src" part correct here? or the intention is to build image from one level up

cmd/epp/main.go
```diff
@@ -27,20 +27,56 @@ package main
...
+	if err != nil {
+		// Log error but don't fail - tracing is optional
+		ctrl.Log.Error(err, "Failed to initialize tracing")
+	}
+
+	// Add startup span to verify tracing is working
+	tracer := telemetry.Tracer()
+	ctx, span := tracer.Start(ctx, "llm_d.epp.startup")
+	span.SetAttributes(
+		attribute.String("component", "llm-d-inference-scheduler"),
```
zdtsw: should we use "epp" than "llm-d-inference-scheduler?

cmd/pd-sidecar/main.go
```diff
@@ -70,6 +73,29 @@ func main() {
...
+	}
+
+	// Add startup span to verify tracing is working
+	tracer := telemetry.Tracer()
+	ctx, span := tracer.Start(ctx, "llm_d.pd_proxy.startup")
+	span.SetAttributes(
+		attribute.String("component", "llm-d-pd-proxy"),
+		attribute.String("operation", "startup"),
+	)
+	defer span.End()
```
zdtsw: should be a "defer" on the "epp" as well?

pkg/telemetry/tracing.go
```diff
@@ -0,0 +1,128 @@
...
+// Configuration is done via environment variables:
+// - OTEL_SERVICE_NAME: Service name for tracing (default: llm-d-inference-scheduler)
+// - OTEL_EXPORTER_OTLP_ENDPOINT: OTLP collector endpoint (default: http://localhost:4317)
+// - OTEL_TRACES_SAMPLER: Sampling strategy (default: parentbased_traceidratio)
+// - OTEL_TRACES_SAMPLER_ARG: Sampling ratio (default: 0.1 for 10%)
+func InitTracing(ctx context.Context) (func(context.Context) error, error) {
+	logger := log.FromContext(ctx)
+
+	// Get service name from environment, fallback to default
+	svcName := os.Getenv("OTEL_SERVICE_NAME")
```
zdtsw: seems `svcName` is not really needed?
either `serviceName` set to `os.Getenv("OTEL_SERVICE_NAME") `or  use `defaultServiceName`

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -127,21 +132,47 @@ func (s *PrecisePrefixCacheScorer) WithName(name string) *PrecisePrefixCacheScor
...
+		attribute.Int("llm_d.scorer.candidate_pods", len(pods)),
+	)
+
+	if request != nil && request.TargetModel != "" {
+		span.SetAttributes(attribute.String("gen_ai.request.model", request.TargetModel))
+	}
+	if request != nil && request.RequestId != "" {
+		span.SetAttributes(attribute.String("gen_ai.request.id", request.RequestId))
+	}
+
```
wseaton: nit:
```suggestion
if request != nil {
    if request.TargetModel != "" {
		span.SetAttributes(attribute.String("gen_ai.request.model", request.TargetModel))
	}
    if request.RequestId != "" {
		span.SetAttributes(attribute.String("gen_ai.request.id", request.RequestId))
	}
}
```

pkg/sidecar/proxy/chat_completions.go
```diff
@@ -32,6 +43,21 @@ var (
 )
 
 func (s *Server) chatCompletionsHandler(w http.ResponseWriter, r *http.Request) {
+	requestStart := time.Now()
+	tracer := telemetry.Tracer()
+	ctx, span := tracer.Start(r.Context(), "llm_d.pd_proxy.request",
```
zdtsw, line 48: read the [proposal part ](https://github.com/llm-d/llm-d/pull/119/files )seem iit has decided to use llm_d.pd_proxy.

not for this PR,  but should we change the folder to `pkg/pd/proxy `than `pkg/sidecar/proxy` for better naming?

Dockerfile.localdev.epp
```diff
@@ -0,0 +1,110 @@
+# Build Stage: using Go 1.25 image
```
zdtsw: is this intended to check-in in this PR?

pkg/telemetry/tracing.go
```diff
@@ -0,0 +1,128 @@
...
+	return otel.Tracer(serviceName)
+}
+
+// stripScheme removes http:// or https:// prefix from endpoint URL
+// OpenTelemetry gRPC exporter expects host:port format only
+func stripScheme(endpoint string) string {
+	endpoint = strings.TrimPrefix(endpoint, "http://")
+	endpoint = strings.TrimPrefix(endpoint, "https://")
+	return endpoint
+}
```
zdtsw: looks like a common func, should it be moved to` pkg/common/common.go` ?

Dockerfile.localdev.epp
sallyom: this Dockerfile I was mostly using in order to build & test from local sources- I'll clean it up!

Dockerfile.localdev.epp
sallyom: probably not - unless it's useful and people want it :)

cmd/pd-sidecar/main.go
sallyom: added

pkg/telemetry/tracing.go
sallyom: fixed this, ty!

Dockerfile.localdev.epp
sallyom: I will add this local-build Dockerfile in a follow-up PR if people want it - I removed it from this PR

Dockerfile.localdev.epp
sallyom: I will add the localdev Dockerfile in a follow-up PR if others want it, but I've removed it from this PR

Dockerfile.localdev.epp
sallyom: removed!

pkg/plugins/scorer/precise_prefix_cache.go
sallyom: fixed, ty!

sallyom:
> The GAIE entry request span PR is [kubernetes-sigs/gateway-api-inference-extension#2057](https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2057)

THis PR merged!

pkg/plugins/scorer/precise_prefix_cache.go
sallyom: updated, but I went with a check & handle nil/error early instead

cmd/epp/main.go
```diff
@@ -27,20 +27,55 @@ package main
...
 	if err := runner.NewRunner().
 		WithCustomCollectors(metrics.GetCollectors()...).
-		Run(ctrl.SetupSignalHandler()); err != nil {
+		Run(ctx); err != nil {
+		span.SetStatus(codes.Error, "startup failed")
+		if shutdownTracing != nil {
+			if err := shutdownTracing(ctx); err != nil {
+				ctrl.Log.Error(err, "Failed to shutdown tracing")
+			}
+		}
```
zdtsw: yeah, this is valid.
we could wrap up the code in a function. 
and from main to call this function.
```
func main() {
	os.Exit(run())
}
func run() int {
// the current logic
// move os.Exit(1) to return 1
}
```
then both defer and os.Exit(1) should work.

sallyom: I agree but I'd rather leave that to a follow-up - since it's not related so much to tracing and I'm already out of my neighborhood here in this repo :) I'm trying to keep my changes as minimal as can be.

cmd/epp/main.go
```diff
@@ -27,20 +27,55 @@ package main
...
+	ctx := ctrl.SetupSignalHandler()
+
+	// Initialize tracing before creating any spans
+	shutdownTracing, err := telemetry.InitTracing(ctx)
+	if err != nil {
+		// Log error but don't fail - tracing is optional
+		ctrl.Log.Error(err, "Failed to initialize tracing")
+	}
+
+	// Add startup span to verify tracing is working
```
elevran: Q: how does the below verify that tracing is working? Is there an expected error returned anywhere?

pkg/common/common.go
```diff
@@ -4,10 +4,20 @@
...
+
 const (
 	// PrefillPodHeader is the header name used to indicate Prefill worker <ip:port>
 	PrefillPodHeader = "x-prefiller-host-port"
 
 	// DataParallelPodHeader is the header name used to indicate the worker <ip:port> for Data Parallel
 	DataParallelPodHeader = "x-data-parallel-host-port"
 )
+
+// StripScheme removes http:// or https:// prefix from endpoint URL
```
elevran: nit: This could be improved with `url.Parse()` to better handle corner cases?

pkg/common/common_test.go
```diff
@@ -0,0 +1,87 @@
+/*
+Copyright 2025 The llm-d Authors.
```
elevran: nit: probably (c) 2026-...?

pkg/common/common_test.go
```diff
@@ -0,0 +1,87 @@
...
+*/
+
+package common
+
+import "testing"
+
+func TestStripScheme(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
```
elevran, line 12: nit: additional test cases that may make sense?
- `//192.168.1.1:80`
- `HTTP://localhost:4317`
- `:9090`

pkg/plugins/pre-request/pd_prerequest.go
```diff
@@ -67,17 +70,46 @@ func (p *PrefillHeaderHandler) WithName(name string) *PrefillHeaderHandler {
 }
 
 // PreRequest wires prefill SchedulerProfile result into a header to indicate prefill worker
-func (p *PrefillHeaderHandler) PreRequest(_ context.Context, request *types.LLMRequest, schedulingResult *types.SchedulingResult) {
+func (p *PrefillHeaderHandler) PreRequest(ctx context.Context, request *types.LLMRequest, schedulingResult *types.SchedulingResult) {
+	tracer := telemetry.Tracer()
+	_, span := tracer.Start(ctx, "llm_d.epp.prerequest.pd_disaggregation",
```
elevran, line 75: nit: consider making consistent use of `llm_d.epp` (or `llm-d.epp`) in all span names (e.g., the "startup" span in main.go)?

pkg/plugins/pre-request/pd_prerequest.go
```diff
@@ -67,17 +70,46 @@ func (p *PrefillHeaderHandler) WithName(name string) *PrefillHeaderHandler {
...
+	}
+
 	if _, found := request.Headers[common.PrefillPodHeader]; found {
 		request.Headers[common.PrefillPodHeader] = "" // clear header, if already set
 	}
 
 	prefillProfileRunResult, exists := schedulingResult.ProfileResults[p.prefillProfile]
 	if !exists {
+		span.SetAttributes(
+			attribute.Bool("llm_d.epp.pd.disaggregation_enabled", false),
```
elevran: nit: it is enabled, just not used for this request

pkg/plugins/profile/pd_profile_handler.go
```diff
@@ -124,8 +128,35 @@ func (h *PdProfileHandler) WithName(name string) *PdProfileHandler {
 // previously executed cycles along with their results.
 func (h *PdProfileHandler) Pick(ctx context.Context, cycleState *types.CycleState, request *types.LLMRequest, profiles map[string]*framework.SchedulerProfile,
 	profileResults map[string]*types.ProfileRunResult) map[string]*framework.SchedulerProfile {
+	// Start tracing span for profile picking operation
+	tracer := telemetry.Tracer()
+	ctx, span := tracer.Start(ctx, "llm_d.epp.profile_handler.pick",
```
elevran: nit: do you want to span to explicitly state this is a P/D profile handler Pick() function? Missing P/D designation - all profile handler plugins have a Pick function...

pkg/sidecar/proxy/chat_completions.go
elevran: Down the road we may have the sidecar proxy handle more than P/D disagg so I would suggest not renaming.

cmd/pd-sidecar/main.go
```diff
@@ -70,6 +71,20 @@ func main() {
 	ctx := ctrl.SetupSignalHandler()
 	log.IntoContext(ctx, logger)
 
+	// Initialize tracing before creating any spans
+	shutdownTracing, err := telemetry.InitTracing(ctx)
```
kfswain, line 75: would we ever want to be able to disable this?

pkg/sidecar/proxy/chat_completions.go
sallyom: ok leaving as/is for this PR

cmd/epp/main.go
sallyom: simplified and removed the startup span, no longer relevant

cmd/epp/main.go
sallyom: Updated!

cmd/epp/main.go
sallyom: removed this

pkg/common/common.go
sallyom: added!

pkg/plugins/pre-request/pd_prerequest.go
sallyom: updated, thanks!

pkg/common/common_test.go
sallyom: added!

pkg/plugins/profile/pd_profile_handler.go
sallyom: updated to add pd

pkg/common/common_test.go
sallyom: it was 2025 when it was created :) but yea, updated

cmd/pd-sidecar/main.go
sallyom: rather than introduce a config option, InitTracing is safe to always call - the conditional is within the OTEL SDK so if no collector -> no generated spans (tracer.Start is a no-op) -> no cost. If the default collector is unreachable or doesn't exist, there's no error & negligible overhead.

pkg/sidecar/proxy/chat_completions.go
```diff
@@ -32,6 +43,22 @@ var (
...
+	)
+	defer span.End()
+
+	// Update request context with span and start time
+	ctx = context.WithValue(ctx, requestStartTimeKey, requestStart)
+	r = r.WithContext(ctx)
+
+	span.SetAttributes(
+		attribute.String("llm_d.pd_proxy.connector", s.config.Connector),
+		attribute.String("llm_d.pd_proxy.request_path", r.URL.Path),
```
elevran: `pkg/sidecar/proxy/chat_completions_test.go` sends requests with just the minimum needed info. Specifically, it does **not** set the request's URL, so `r.URL.Path` segfaults.
You should either fix the test or guard against the `nil` `r.URL` here

elevran: I _think_ `s.config` does have a default value (or it is set up by the test), but might be worthwhile to guard against that as well.

PR #509, opened 2025-12-09
feat(lmcache): implement decode first flow on lmcache connector when cache_hit_threshold field is present

kyanokashi:
This PR updates the lmcache connector on the sidecar to decode first when `cache_hit_threshold` is present in the completion request. 

The new flow goes as follows:
1. If `cache_hit_threshold` present; then decode first
2. If decode is successful write success response; return
3. If finish reason is `cache_threshold` that means the decode node didn't meet the cache hit threshold
4. Then we continue with the original flow of prefill, then decode 

# Test Report

**Environment:**
- Kind Cluster: `llm-d-inference-scheduler-dev`
- vLLM Simulator Image: `kyanokashi/llm-d-inference-sim:dev`
- Sidecar Connector: `lmcache`
- Log Level: `--zap-log-level=4`

---

### Scenario 1: Non-Streaming WITHOUT `cache_hit_threshold`
**Description**: Basic request without lmcache protocol activation.

**Result**: ✅ PASS - Request proxied directly to decoder.

---

### Scenario 2: Non-Streaming WITH `cache_hit_threshold`
**Description**: Request with `cache_hit_threshold` field, no `cache_threshold` triggered.

**Result**: ✅ PASS - LMCache protocol runs, response returned successfully.

---

### Scenario 3: Non-Streaming WITH `cache_hit_threshold` AND `X-Cache-Threshold: true`
**Description**: Request configured to trigger `cache_threshold` finish_reason.

**Result**: ✅ PASS - LMCache protocol detects `cache_threshold`, triggers prefill→decode.

---

### Scenario 4: Streaming WITHOUT `cache_hit_threshold`
**Description**: Streaming request without lmcache protocol activation.

**Result**: ✅ PASS - SSE chunks streamed directly from decoder.

---

### Scenario 5: Streaming WITH `cache_hit_threshold`
**Description**: Streaming request with `cache_hit_threshold`, no header.

**Result**: ✅ PASS - LMCache protocol parses SSE chunks, streams through when no `cache_threshold` found.

---

### Scenario 6: Streaming WITH `cache_hit_threshold` AND `X-Cache-Threshold: true`
**Description**: Streaming request with full lmcache configuration.

**Result**: ✅ PASS - LMCache protocol detects `cache_threshold` in SSE, triggers prefill→decode.

---

## Failure Injection Test

To enable failure injection on the simulator:
```bash
--failure-injection-rate=100
--failure-types=rate_limit
```

### Scenario 7: Streaming Decode Error (tryDecodeStreaming)

**Request**:
```bash
curl -s http://localhost:30080/v1/chat/completions \
  -d '{
    "model": "food-review",
    "messages": [{"role": "user", "content": "Hello"}],
    "max_tokens": 5,
    "cache_hit_threshold": 0.5,
    "stream": true
  }'
```

**Path**: `tryDecodeStreaming → error`

**Result**: ✅ PASS
- HTTP Status: 429
- Error returned as JSON (not SSE)

**Response**:
```json
{
  "error": {
    "code": 429,
    "message": "Rate limit reached for food-review...",
    "type": "RateLimitError"
  }
}
```

---

## Summary

| Scenario | Streaming | cache_hit_threshold | X-Cache-Threshold | Result |
|----------|-----------|---------------------|-------------------|--------|
| 1 | No | No | No | ✅ PASS |
| 2 | No | Yes | No | ✅ PASS |
| 3 | No | Yes | Yes | ✅ PASS |
| 4 | Yes | No | No | ✅ PASS |
| 5 | Yes | Yes | No | ✅ PASS |
| 6 | Yes | Yes | Yes | ✅ PASS |
| 7 | Yes | Yes | No (error) | ✅ PASS |

All scenarios passed successfully.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -617,7 +617,10 @@ func runStreamingCompletion(prompt string, theModel openai.CompletionNewParamsMo
 
 	resp, err := http.DefaultClient.Do(req)
 	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-	defer resp.Body.Close()
+	defer func() {
+		err := resp.Body.Close()
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	}()
 
 	gomega.Expect(resp.StatusCode).Should(gomega.Equal(http.StatusOK))
 
@@ -644,7 +647,10 @@ func runStreamingChatCompletion(prompt string) (string, string) {
 
 	resp, err := http.DefaultClient.Do(req)
 	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-	defer resp.Body.Close()
+	defer func() {
+		err := resp.Body.Close()
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	}()
 
 	gomega.Expect(resp.StatusCode).Should(gomega.Equal(http.StatusOK))
 
@@ -678,7 +684,10 @@ func runCompletionWithCacheThreshold(prompt string, cacheHitThreshold float64, f
 
 	resp, err := http.DefaultClient.Do(req)
 	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-	defer resp.Body.Close()
+	defer func() {
+		err := resp.Body.Close()
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	}()
 
 	gomega.Expect(resp.StatusCode).Should(gomega.Equal(http.StatusOK))
 
@@ -712,7 +721,10 @@ func runStreamingCompletionWithCacheThreshold(prompt string, cacheHitThreshold f
 
 	resp, err := http.DefaultClient.Do(req)
 	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-	defer resp.Body.Close()
+	defer func() {
+		err := resp.Body.Close()
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	}()
 
 	gomega.Expect(resp.StatusCode).Should(gomega.Equal(http.StatusOK))
 
@@ -768,7 +780,7 @@ func extractFinishReasonFromStreaming(sseData string) string {
 // getPrefillRequestCount gets the total request count from a prefill pod's metrics endpoint.
 // This is used to verify whether a request was processed by the prefill pod.
 func getPrefillRequestCount(prefillPodName string) int {
-	ginkgo.By(fmt.Sprintf("Getting request count from prefill pod: %s", prefillPodName))
+	ginkgo.By("Getting request count from prefill pod: " + prefillPodName)
 
 	// Use Kubernetes API proxy to access the metrics endpoint
 	output, err := testConfig.KubeCli.CoreV1().RESTClient().


pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -26,16 +27,14 @@ import (
 func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, prefillPodHostPort string) {
 	s.logger.Info("running LMCache protocol")
 
-	// Read and parse request body
 	defer r.Body.Close() //nolint:all
 	original, err := io.ReadAll(r.Body)
 	if err != nil {
-		w.WriteHeader(http.StatusBadRequest) // TODO: check FastAPI error code when failing to read body
-		w.Write([]byte(err.Error()))         //nolint:all
+		w.WriteHeader(http.StatusBadRequest)
```
elevran: Q: does that mean the `TODO` in the removed line is done and this is the right error code?

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -44,11 +43,117 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+		return
+	}
+
+	// If decode succeeded (cache hit was sufficient), we're done
+	if !needsPrefill {
+		s.logger.V(4).Info("decode succeeded without prefill")
+		return
+	}
+
+	// Step 2: Cache threshold not met, execute prefill
```
elevran: is there a way to refactor so that if decde first has insufficient cache (and therefore prefill should run), we can call into the `prefillThenDecode` function and not duplicate some of the code here?

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -44,11 +43,117 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+		w.WriteHeader(dw.statusCode)
+		if dw.buffer.Len() > 0 {
+			w.Write([]byte(dw.buffer.String())) //nolint:all
+		}
+		return false, fmt.Errorf("decode request failed with status code: %d", dw.statusCode)
+	}
+
+	// Parse response to check finish_reason
+	var response map[string]any
+	if err := json.Unmarshal([]byte(dw.buffer.String()), &response); err != nil {
```
elevran: Q: if we fail to unmarshal we pass the reposne back as is, with original status and body, **and** no prefill needed? Can you please explain the error handling path? Not saying it's wrong - looking to understand.

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -44,11 +43,117 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+
+		}
+	}
+
+	// Decode succeeded, write response to client
+	for k, v := range dw.headers {
+		for _, val := range v {
+			w.Header().Add(k, val)
+		}
+	}
```
elevran: nit: maybe copy `v` into `w.Header()` instead of appending each `val` entry?

pkg/sidecar/proxy/connector_lmcache.go
kyanokashi: No, my bad

pkg/sidecar/proxy/connector_lmcache.go
kyanokashi: Yes, done

pkg/sidecar/proxy/connector_lmcache.go
kyanokashi: hmmm yeah, I overlooked this part. Should we send a 500? The reason I skipped prefill is that this likely indicates a fault in the decode node that would still fail after prefill. What do you think would be the approach in this scenario?

elevran: if we failed to unmarshal to map[string]any, it's likely not valid JSON.
I don't know if this is a possible response from the inference server. 
Debating with myself whether it's best to surface the reponse back to the caller or return an error as unexpected.
We should likely be logging these for further investigation (or may produce a sidecar metrics for errors...)

elevran: @kfirwolfson any thoughts?

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -18,15 +18,16 @@ package proxy
...
 	"io"
+	"maps"
 	"net/http"
 	"strings"
 )
 
 func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, prefillPodHostPort string) {
 	s.logger.Info("running LMCache protocol")
 
-	// Read and parse request body
```
kfirwolfson, line 29: Suggest keeping informational comments

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -35,7 +36,6 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
 		return
 	}
 
-	// Parse completion request
```
kfirwolfson, line 38: Suggest keeping informational comments

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -44,11 +44,82 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
 		return
 	}
 
-	// Create prefiller request. Set max_tokens to 1.
+	if s.forwardDataParallel && s.dataParallelHandler(w, r) {
```
kfirwolfson: This whole "if" code block and its contents should be removed. We don't want to start with "prefill". We want to start with "tryDecode" flow.

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -44,11 +44,82 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+	if _, hasCacheHitThreshold := completionRequest[requestFieldCacheHitThreshold]; hasCacheHitThreshold {
+		needsPrefill, err := s.tryDecode(w, r)
+		if err != nil {
+			return
+		}
+		if !needsPrefill {
+			s.logger.V(4).Info("decode succeeded without prefill")
+			return
+		}
+		s.logger.Info("decode failed due to insufficient cache hit threshold.")
```
kfirwolfson: This can be a common scenario. Should be V(4). Let's also print the threshold (rename from "_")

pkg/sidecar/proxy/connector_lmcache.go
kfirwolfson: Suggest changing text to "failed to send Invalid JSON error response to Client".

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -44,11 +44,82 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
 	}
 
-	// Create prefiller request. Set max_tokens to 1.
+	if s.forwardDataParallel && s.dataParallelHandler(w, r) {
+		if err := s.prefill(w, r, prefillPodHostPort, completionRequest); err != nil {
+			s.logger.Error(err, "prefill failed")
+		}
+		return
+	}
+
```
kfirwolfson: Suggest adding a comment explaining the logic - if cache_hit_threshold appears in the request, we take it as a sign to issue decode-first flow with this threshold. If not - then a prefill-first flow. Possibly link to RFC for more details.

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -44,11 +44,82 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+}
+
+// prefill routes a request to a preill node
+func (s *Server) prefill(w http.ResponseWriter, r *http.Request, prefillPodHostPort string, completionRequest map[string]any) error {
 	ctx := r.Context()
 	preq := r.Clone(ctx)
 
+	// Prepare prefill request
 	completionRequest[requestFieldMaxTokens] = 1
 	completionRequest[requestFieldMaxCompletionTokens] = 1
```
kfirwolfson: should also set cache_threshold = 0, since we want prefiller to perform the prefill without any cache condition.

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -44,11 +44,82 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+	}
+
+	// Decode succeeded, write response to client
+	maps.Copy(w.Header(), dw.headers)
+	w.Write([]byte(dw.buffer.String())) //nolint:all
+
+	return false, nil
+}
+
+// prefill routes a request to a preill node
```
kfirwolfson: typo: preill -> prefill.

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -57,35 +128,33 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
 		if err := errorJSONInvalid(err, w); err != nil {
 			s.logger.Error(err, "failed to send error response to client")
```
kfirwolfson: Suggest changing text to "failed to send Invalid JSON error response to Client".

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -57,35 +128,33 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
 	}
 	preq.Body = io.NopCloser(strings.NewReader(string(pbody)))
 	preq.ContentLength = int64(len(pbody))
 
-	// Forward request to prefiller
-
 	prefillHandler, err := s.prefillerProxyHandler(prefillPodHostPort)
 	if err != nil {
 		if err := errorBadGateway(err, w); err != nil {
 			s.logger.Error(err, "failed to send error response to client")
```
kfirwolfson: Suggest changing text to "failed to send Bad Gateway error response to Client".

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -57,35 +128,33 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
 	}
+
+	// send prefill request
 	s.logger.V(4).Info("sending prefill request", "to", prefillPodHostPort)
 	pw := &bufferedResponseWriter{}
 	prefillHandler.ServeHTTP(pw, preq)
 
 	if pw.statusCode < 200 || pw.statusCode >= 300 {
-		s.logger.Error(err, "request failed", "code", pw.statusCode)
+		s.logger.Error(nil, "prefill request failed", "code", pw.statusCode)
```
kfirwolfson: good catch of "err" being unrelated!

pkg/sidecar/proxy/connector_lmcache.go
kfirwolfson: I don't think it's a common scenario - the decoder returning an invalid JSON body (dw.buffer) with a 2XX statusCode. I suggest:
1. logging the error for future debug including the contents of dw.buffer, and 
2. returning an error statusCode like 500, I mean adding "func errorInternalServerError" to errors.go and calling it here.

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -44,11 +44,82 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+}
+
+// tryDecode attempts to decode and returns whether prefill is needed
+func (s *Server) tryDecode(w http.ResponseWriter, r *http.Request) (bool, error) {
+	dw := &bufferedResponseWriter{}
+	s.decoderProxy.ServeHTTP(dw, r)
 
+	// Check for non-success status codes
+	if dw.statusCode < 200 || dw.statusCode >= 300 {
+		w.WriteHeader(dw.statusCode)
```
kfirwolfson: contemplating if the correct thing in case Decode fails is to return the error to the user or jump to the Prefill-first flow.
I guess we can start with the implementation as it is here (return error to the user) and fix if needed

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -44,11 +44,82 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+					return true, nil
+				}
+			}
+
+		}
+	}
+
+	// Decode succeeded, write response to client
+	maps.Copy(w.Header(), dw.headers)
+	w.Write([]byte(dw.buffer.String())) //nolint:all
```
kfirwolfson: do we also want to call `w.WriteHeader(http.StatusOK)` for good measure?

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -44,11 +44,82 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+			return
+		}
+		s.logger.Info("decode failed due to insufficient cache hit threshold.")
+	}
+
+	if err := s.prefill(w, r, prefillPodHostPort, completionRequest); err != nil {
+		s.logger.Error(err, "prefill failed")
+		return
+	}
+
```
kfirwolfson: I suggest handing the Decode preemption problem (explained in detail in [vLLM ticket 24256](https://github.com/vllm-project/vllm/issues/24256) ), in **another PR**.

So for now, we should also set cache_threshold = 0 in the request we send to the decoder a second time. This is ensure the response to the user will fit expectations (no Status 200 for incomplete requests).

In a separate PR for handling preemption, we should set the cache_threshold for Decoder requests and handle responses accordingly (e.g. start Prefill-first flow from scratch)

kfirwolfson:
I suggest focusing on d-first use-case in this PR, and handing the Decode preemption problem (explained in detail in [vLLM issue 24256](https://github.com/vllm-project/vllm/issues/24256) ), in separate PR.

pkg/sidecar/proxy/connector_lmcache.go
kyanokashi: That's the default status if nothing else is specified

pkg/sidecar/proxy/connector_lmcache.go
kyanokashi: This is to account for the temporary workaround @shmuelk mentioned he put in place for something related to istio. In that case we would only prefill, therefore, we don't want to call `tryDecode`.

kfirwolfson: I believe that if DP is in use, it's only for the Decode phase, and prefill works normally. @shmuelk please correct me if I am mistaken. Assuming this code will be removed soon, maybe we can avoid having a special handling for  s.forwardDataParallel == true

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -51,6 +51,9 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
 		return
 	}
 
+	// If "cache_hit_threshold" is present in the request, we try to decode first. The decode node must mit the cache hit threshold in order to execute.
```
kfirwolfson: typo: mit --> meet

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -51,6 +51,9 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+	// If the decode node is below the threshold, it won't process the request and return a "cache_threshold" finish reason. In that case, we need to prefill.
```
kfirwolfson: "In that case, we fall back to P/D disaggregation: perform prefill and then decode"

pkg/sidecar/proxy/connector_lmcache.go
kfirwolfson: @kyanokashi i think i wrote too much text above and it's confusing :-) . Bottom line - we should set requestFieldCacheHitThreshold to 0 before sending "r" to decodeProxy here.

pkg/sidecar/proxy/connector_lmcache.go
kfirwolfson: Not sure how this would work with Streaming (vLLM online inference with partial responses). Would it?

pkg/sidecar/proxy/connector_lmcache.go
shmuelk: This code is for the workaround we have for Istio 1.28.0.

Istio 1.28.1 has a fix for this issue. We will be removing this code in the future.

kfirwolfson: Should it be when sending to prefill or decode, @shmuelk ?

kyanokashi: So the istio workaround specifically intends to avoid decoding and only prefill for some scenarios. 

I guess the question is whether the condition caused from the istio workaround and the `cache_hit_threshold` being present in the request could exist. Technically it shouldn't because that field is only relevant for decoding correct me if I'm wrong @kfirwolfson

pkg/sidecar/proxy/connector_lmcache.go
kyanokashi: was meant to be `hit` but `meet` is indeed better haha

pkg/sidecar/proxy/connector_lmcache.go
kfirwolfson: I am not sure anyone will try running the connector_lmcache.py code with s.forwardDataParallel==True, let alone with cache_hit_threshold>0, before the s.forwardDataParallel code will be removed, so I don't think it matters much.
But for now, let's just remove this whole "if" statement. It's inaccurate for when s.forwardDataParallel==True and irrelevant for when s.forwardDataParallel==False.

pkg/sidecar/proxy/connector_lmcache.go
kyanokashi: Hmmm this is tricky. 

Previously we were passing a pass through writer to the decode proxy which was responsible for writing responses back to the client. 

Now, because we need to parse the response to determine the finish reason, we use a buffered writer so we could read the first choice. 

Let me explore some options. I'm thinking of either updating `bufferedResponseWriter` to support streaming or implementing a new writer type that handles the `cache_threshold` case specifically

pkg/sidecar/proxy/connector_lmcache.go
kyanokashi: https://github.com/llm-d/llm-d-inference-scheduler/pull/509/commits/4c8659e5c349c99f38ef258d2fb26a1cd76ef15b

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -130,7 +141,6 @@ func (s *Server) prefill(w http.ResponseWriter, r *http.Request, prefillPodHostP
 	// Prepare prefill request
 	completionRequest[requestFieldMaxTokens] = 1
 	completionRequest[requestFieldMaxCompletionTokens] = 1
-	completionRequest[requestFieldCacheHitThreshold] = 0
```
kfirwolfson: @kyanokashi we actually also need this set to 0 here, when sending to prefill.
The cache threshold is only used in the first "try decode" step. After that - it should be zeroed out.

pkg/sidecar/proxy/connector_lmcache.go
kyanokashi: Here's what I ended up doing https://github.com/llm-d/llm-d-inference-scheduler/pull/509/commits/88739c66b8ff35eed085f1b1197c09a4e702b651

It's a bit complex, but couldn't think of a better way to do it. 

Still need to test this

@kfirwolfson

kyanokashi:
@kfirwolfson @elevran testing is complete. I updated the PR description with a full test report

pkg/sidecar/proxy/connector_lmcache.go
kfirwolfson: nice work.

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -39,53 +44,224 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+	}
+
+	if s.hasCacheThresholdFinishReason(response) {
+		s.logger.V(4).Info("finish reason cache_threshold detected, needs prefill")
+		return true, nil
+	}
+
+	// No cache_threshold found, flush buffer and switch to direct mode
+	// to let the rest of the response stream through.
+	s.logger.V(4).Info("no cache_threshold, switching to direct mode")
```
kfirwolfson: "no cache_threshold" is a bit misleading (I read it as no incoming threshold in the request). Suggest rephrasing to something like "first response for request shows success without cache_threshold reason"

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -39,53 +44,224 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+	}
+
+	// If "cache_hit_threshold" is present in the request, we try to decode first. The decode node must meet the cache hit threshold in order to execute.
+	// If the decode node is below the threshold, it won't process the request and return a "cache_threshold" finish reason. In that case,
+	// we fall back to P/D disaggregation: perform prefill and then decode.
+	// For more infromation refer to the RFC https://github.com/vllm-project/vllm/issues/24256
+	if cacheHitThreshold, hasCacheHitThreshold := completionRequest[requestFieldCacheHitThreshold]; hasCacheHitThreshold {
+		decodeReq := r.Clone(ctx)
+		decodeReq.Body = io.NopCloser(bytes.NewReader(original))
+		decodeReq.ContentLength = int64(len(original))
```
kfirwolfson: suggest adding a V(4) debug log, indicating that cache hit threshold exists in the request, so trying decode first. can also mention the threshold itself, like we do in the failure case

kfirwolfson (commented):
Very nice. Tests look good to me. 
@kyanokashi, Is it possible to add tests to cover both streaming and non-streaming modes?

After seeing the tests output, I added a couple of suggestions for the debug logs to make the flow slightly clearer.

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -39,53 +44,232 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
 	var completionRequest map[string]any
 	if err := json.Unmarshal(original, &completionRequest); err != nil {
 		if err := errorJSONInvalid(err, w); err != nil {
-			s.logger.Error(err, "failed to send error response to client")
+			s.logger.Error(err, "failed to send Invalid JSON error response to Client")
```
elevran:
```suggestion
			s.logger.Error(err, "failed to send Invalid JSON error response to client")
```
nit (do not capitalize `Client`  in error message)

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -39,53 +44,232 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+	// If the decode node is below the threshold, it won't process the request and return a "cache_threshold" finish reason. In that case,
+	// we fall back to P/D disaggregation: perform prefill and then decode.
+	// For more information refer to the RFC https://github.com/vllm-project/vllm/issues/24256
+	if cacheHitThreshold, hasCacheHitThreshold := completionRequest[requestFieldCacheHitThreshold]; hasCacheHitThreshold {
+		s.logger.V(4).Info("cache_hit_threshold field found in the request, trying to decode first", requestFieldCacheHitThreshold, cacheHitThreshold)
+		decodeReq := r.Clone(ctx)
+		decodeReq.Body = io.NopCloser(bytes.NewReader(original))
+		decodeReq.ContentLength = int64(len(original))
+		isStreaming, _ := completionRequest[requestFieldStream].(bool)
+		needsPrefill, err := s.tryDecode(w, decodeReq, isStreaming)
```
elevran: would it be worthwhile to log the error from `tryDecode` for context?

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -39,53 +44,232 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+		s.logger.Error(err, "prefill failed")
+		return
+	}
+
+	s.logger.V(4).Info("forwarding to decoder after prefill")
+	completionRequest[requestFieldCacheHitThreshold] = 0
+	decodeRequestBody, err := json.Marshal(completionRequest)
+	if err != nil {
+		if err := errorJSONInvalid(err, w); err != nil {
+			s.logger.Error(err, "failed to send Invalid JSON error response to Client")
```
elevran:
```suggestion
			s.logger.Error(err, "failed to send Invalid JSON error response to client")
```

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -39,53 +44,232 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+}
+
+// tryDecodeBuffered handles non-streaming decode attempts.
+// It buffers the entire response before inspecting it.
+func (s *Server) tryDecodeBuffered(w http.ResponseWriter, r *http.Request) (bool, error) {
+	dw := &bufferedResponseWriter{}
+	s.decoderProxy.ServeHTTP(dw, r)
+
+	// Check for non-success status codes
+	if dw.statusCode < 200 || dw.statusCode >= 300 {
```
elevran: nits: 
- use `http.StatusOK` and `http.StatusMultipleChoices`? 
- move comment in same line as `if`, not above it.
- consider making this check a function since also used below (L166)

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -39,53 +44,232 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+		if dw.buffer.Len() > 0 {
+			w.Write([]byte(dw.buffer.String())) //nolint:all
+		}
+		return false, fmt.Errorf("decode request failed with status code: %d", dw.statusCode)
+	}
+
+	// Parse response to check finish_reason
+	var response map[string]any
+	if err := json.Unmarshal([]byte(dw.buffer.String()), &response); err != nil {
+		s.logger.Error(err, "failed to unmarshal decode response", "response", dw.buffer.String())
```
elevran, line 123: nit: any danger in this leaking sensitive information into the logs?

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -39,53 +44,232 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+
+	// Check buffered SSE content for cache_threshold finish reason.
+	if s.checkBufferedResponseForCacheThreshold(w.buffered()) {
+		s.logger.V(4).Info("finish reason cache_threshold detected, needs prefill")
+		return true, nil
+	}
+
+	// No cache_threshold finish reasonfound, flush buffer and switch to direct mode
+	// to let the rest of the response stream through.
+	s.logger.V(4).Info("first response for request shows success without cache_threshold finishreason")
```
elevran: should `finishreason` be two words?

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -39,53 +44,232 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+	// Prepare prefill request
 	completionRequest[requestFieldMaxTokens] = 1
 	completionRequest[requestFieldMaxCompletionTokens] = 1
+	completionRequest[requestFieldCacheHitThreshold] = 0
 
 	pbody, err := json.Marshal(completionRequest)
 	if err != nil {
 		if err := errorJSONInvalid(err, w); err != nil {
-			s.logger.Error(err, "failed to send error response to client")
+			s.logger.Error(err, "failed to send Invalid JSON error response to Client")
```
elevran:
```suggestion
			s.logger.Error(err, "failed to send Invalid JSON error response to client")
```

pkg/sidecar/proxy/status_response_writer.go
```diff
@@ -45,3 +47,187 @@ func (w *bufferedResponseWriter) Write(b []byte) (int, error) {
...
+	return rw
+}
+
+// FirstChunkReady returns a channel that receives nil when the first chunk of
+// body data is available in the buffer, or an error if the write failed.
+func (w *responseWriterWithBuffer) FirstChunkReady() <-chan struct{} {
+	return w.ready
+}
+
+func (w *responseWriterWithBuffer) Header() http.Header {
```
elevran, line 91: Q: isn't `Header` auto-exposed due to Go embedding?

pkg/sidecar/proxy/status_response_writer.go
```diff
@@ -114,11 +114,18 @@ func (w *responseWriterWithBuffer) Write(b []byte) (int, error) {
...
-	w.signalReady(err)
-	return n, err
+	// Write() always returns a nil error
+	n, _ := w.buffer.Write(b)
+
+	// Signal ready when buffer contains at least 2 complete SSE events.
+	// For SSE streaming, the first chunk is just the role announcement with
+	// finish_reason:null. We need the second chunk to see if cache_threshold
+	// was returned (early abort) or if decode is proceeding normally.
+	if strings.Count(w.buffer.String(), "\n\n") >= 2 {
```
kfirwolfson: Consider extracting the logic `strings.Count(w.buffer.String(), "\n\n") >= 2` to an external function and reusing it. Define whether locking should be done inside or outside of the function.

Also consider defining "\n\n" as a constant.

pkg/sidecar/proxy/connector_lmcache.go
```diff
@@ -39,53 +44,232 @@ func (s *Server) runLMCacheProtocol(w http.ResponseWriter, r *http.Request, pref
...
+		return false, fmt.Errorf("decode request failed with status code: %d", statusCode)
+	}
+
+	// Check buffered SSE content for cache_threshold finish reason.
+	if s.checkBufferedResponseForCacheThreshold(w.buffered()) {
+		s.logger.V(4).Info("finish reason cache_threshold detected, needs prefill")
+		return true, nil
+	}
+
+	// No cache_threshold finish reasonfound, flush buffer and switch to direct mode
```
kfirwolfson: typo: reasonfound --> reason found

pkg/sidecar/proxy/status_response_writer.go
kyanokashi: Yes, but it isn't embedded on the `responseWriterWithBuffer` struct, it's a named field

```
writerFlusher flushableResponseWriter
```

pkg/sidecar/proxy/connector_lmcache.go
kyanokashi: I already log a few of the errors from within `tryDecode`. We have different logs for different errors. 

I added though more log for failure points that were previously not covered.

pkg/sidecar/proxy/connector_lmcache.go
kyanokashi: https://github.com/llm-d/llm-d-inference-scheduler/pull/509/commits/bd114fa64222ad6b2ade5713cb23883dbe11cbaf

pkg/sidecar/proxy/connector_lmcache.go
kyanokashi: yes

pkg/sidecar/proxy/status_response_writer.go
kyanokashi: Locking should definitely be done outside of the function IMO since it depends on the usage

pkg/sidecar/proxy/connector_lmcache.go
kyanokashi: I'm unsure about that. 

To begin with, it would already be unusual to encounter this log, as it indicates a fault in the VLLM instance responding with an unformatted response. However, there is a benefit to logging it, because if we don't, it would be hard to know what's going on if we encounter this error.

pkg/sidecar/proxy/status_response_writer.go
```diff
@@ -121,7 +123,7 @@ func (w *responseWriterWithBuffer) Write(b []byte) (int, error) {
 	// For SSE streaming, the first chunk is just the role announcement with
 	// finish_reason:null. We need the second chunk to see if cache_threshold
 	// was returned (early abort) or if decode is proceeding normally.
-	if strings.Count(w.buffer.String(), "\n\n") >= 2 {
+	if countSSEEvents(w.buffer.String()) >= 2 {
```
kfirwolfson: nit. to make sure we use the same logic for signaling, how about extracting " countSSEEvents(w.buffer.String()) >= 2 " to a bool function shouldSignal(w)?

kyanokashi:
> > @kfirwolfson @elevran I was also thinking about the need to better abstract the connector technology used and to enable plug-and-play with different connector types. I did notice some slight differences, though, between the two connector runners in how we prepare the prefill/decode requests. Perhaps @shmuelk can explain the differences more clearly.
> 
> To better understand what you mean, can you point to the two connector runners with the different code, @kyanokashi ?

@kfirwolfson 

They diverge in the request preparation. Here are some examples from the NIXL connector:

https://github.com/kyanokashi/llm-d-inference-scheduler/blob/ea60bf0ecd9364a7081bc5afa8c0c0a71176029d/pkg/sidecar/proxy/connector_nixlv2.go#L72-L80

https://github.com/kyanokashi/llm-d-inference-scheduler/blob/ea60bf0ecd9364a7081bc5afa8c0c0a71176029d/pkg/sidecar/proxy/connector_nixlv2.go#L141-L157

kfirwolfson:
> > > @kfirwolfson @elevran I was also thinking about the need to better abstract the connector technology used and to enable plug-and-play with different connector types. I did notice some slight differences, though, between the two connector runners in how we prepare the prefill/decode requests. Perhaps @shmuelk can explain the differences more clearly.
> > 
> > 
> > To better understand what you mean, can you point to the two connector runners with the different code, @kyanokashi ?
> 
> @kfirwolfson
> 
> They diverge in the request preparation. Here are some examples from the NIXL connector:
> 
> https://github.com/kyanokashi/llm-d-inference-scheduler/blob/ea60bf0ecd9364a7081bc5afa8c0c0a71176029d/pkg/sidecar/proxy/connector_nixlv2.go#L72-L80
> 
> https://github.com/kyanokashi/llm-d-inference-scheduler/blob/ea60bf0ecd9364a7081bc5afa8c0c0a71176029d/pkg/sidecar/proxy/connector_nixlv2.go#L141-L157

@kyanokashi, NIXL connector requires KVTransferParams and returns such params from the Prefiller which should be propagated to the Decoder. I assume those are the main differences.

pkg/sidecar/proxy/proxy.go
```diff
@@ -61,8 +61,8 @@ const (
 	// ConnectorNIXLV2 enables the P/D NIXL v2 protocol
 	ConnectorNIXLV2 = "nixlv2"
 
-	// ConnectorLMCache enables (now deprecated) P/D LMCache protocol
-	ConnectorLMCache = "lmcache"
+	// ConnectorSharedStorage enables (now deprecated) P/D Shared Storage protocol
```
kfirwolfson, line 64: suggest removing the "(now deprecated)" text. We're working to enable it

kfirwolfson:
Overall, both logic and code look good to me, @elevran.

pkg/sidecar/proxy/proxy.go
kyanokashi: I'll remove it in the follow up refactor PR. It has massive refactoring and I prefer not changing the base if I can.

kyanokashi:
@kfirwolfson @elevran @vMaroon what's left before we can merge this?

kyanokashi:
@elevran I think you were referring to the test errors? Anyways they are fixed now

PR #510, opened 2025-12-10
Miscellaneous dependency updates

shmuelk:
This PR updates "dependencies" used in the deployments for development.

This includes:
   1. Use the GIE CRDs from 1.2.0
   2. Use Istio 1.28.1

**Note:** The upgrades required other changes to keep things working. This includes:
   1. Added RBAC for the new InferenceModelRewrite CRD
   2. Added a label to get the HttpRoute to work with the deployed Istio.

### DIFF
--- a/deploy/components/crds-gie/kustomization.yaml
+++ b/deploy/components/crds-gie/kustomization.yaml
@@ -10,4 +10,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-- https://github.com/kubernetes-sigs/gateway-api-inference-extension/config/crd?ref=v1.2.0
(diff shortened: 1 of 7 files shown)


deploy/components/crds-gie/kustomization.yaml
```diff
@@ -10,4 +10,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-- https://github.com/kubernetes-sigs/gateway-api-inference-extension/config/crd?ref=v1.1.0
+- https://github.com/kubernetes-sigs/gateway-api-inference-extension/config/crd?ref=v1.2.0
```
nirrozenbaum: can we update to v1.2.1?

shmuelk: Done

PR #512, opened 2025-12-15
Fix running `make env-dev-kind`

acardace:
Running `make env-dev-kind` will fail if the vllm simulator image hasn't been already pulled.

This fixes it by skipping the manual load & save of the image unless we're dealing with a custom locally built image (using the dev tag).

The kubelet will anyway pull the right image when deploying the pod.

### DIFF
--- a/scripts/kind-dev-env.sh
+++ b/scripts/kind-dev-env.sh
@@ -179,14 +179,16 @@ kubectl --context ${KUBE_CONTEXT} -n local-path-storage wait --for=condition=Rea
 # Load Container Images
 # ------------------------------------------------------------------------------
 
-# Load the vllm simulator image into the cluster
-if [ "${CONTAINER_RUNTIME}" == "podman" ]; then
-	podman save ${VLLM_SIMULATOR_IMAGE} -o /dev/stdout | kind --name ${CLUSTER_NAME} load image-archive /dev/stdin
-else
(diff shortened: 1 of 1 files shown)


scripts/kind-dev-env.sh
```diff
@@ -179,14 +179,16 @@ kubectl --context ${KUBE_CONTEXT} -n local-path-storage wait --for=condition=Rea
...
-# Load the vllm simulator image into the cluster
-if [ "${CONTAINER_RUNTIME}" == "podman" ]; then
-	podman save ${VLLM_SIMULATOR_IMAGE} -o /dev/stdout | kind --name ${CLUSTER_NAME} load image-archive /dev/stdin
-else
-	if docker image inspect "${VLLM_SIMULATOR_IMAGE}" > /dev/null 2>&1; then
-		echo "INFO: Loading image into KIND cluster..."
-		kind --name ${CLUSTER_NAME} load docker-image ${VLLM_SIMULATOR_IMAGE}
-	fi
+# Load the vllm simulator image into the cluster (only if it's a locally built image)
+if [ "${VLLM_SIMULATOR_TAG}" == "dev" ]; then
```
shmuelk:
```suggestion
if [ -n "$(${CONTAINER_RUNTIME} images -q "${VLLM_SIMULATOR_IMAGE}")" ]; then
```

This isn't tied to any specific image tag

acardace: This looks better, thank you!

PR #516, opened 2025-12-19
feat: allow pd_profile_handler to handle diverse plugin types

hyeongyun0916:
Summary
This PR extends the functionality of `pd_profile_handler` to support various plugin types beyond the existing PrefixCachePluginType. It also ensures that scoring data from `precise_prefix_cache` is store in cycleState.

Key Changes
Enhanced Type Compatibility: Refactored `pd_profile_handler` to be more generic, allowing it to accept and process multiple plugin types. This removes the previous limitation of only supporting PrefixCachePluginType.

Persistence of `precise_prefix_cache` score: Modified the logic to store the precise prefix cache score within cycleState.

Information Sharing: This change allows pd_profile_handler to retrieve and utilize the score information directly from cycleState, enabling more detailed and accurate profiling.

Why is this change necessary?
Previously, the `pd_profile_handler` was tightly coupled with a specific plugin type. By generalizing the handler and leveraging cycleState for data persistence, we can now provide a more flexible profiling infrastructure that supports a wider range of scheduling behaviors and more granular performance insights.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ go.work.sum
 .vscode
 
 vendor
+
+# Build output
+/build
--- a/pkg/plugins/scorer/precise_prefix_cache_test.go
+++ b/pkg/plugins/scorer/precise_prefix_cache_test.go
@@ -583,7 +583,7 @@ func TestPrefixCacheTracking_Score(t *testing.T) {
 				}
 			}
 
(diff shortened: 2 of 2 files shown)


mayabar:
> urrent work includes implementing the PrepareData interface, would it be a good id...

Agree

vMaroon:
This looks good as an intermediate step before switching to the data layer. Thanks @hyeongyun0916!

/lgtm
/approve

PR #521, opened 2025-12-23
Run unit, integration tests and builds in a container

acardace:
## Summary

Containerize the entire build system so that the only host requirements are a container runtime (docker or podman) and git. All Go tooling (build, test, lint, format, typos) runs inside a builder container image.

### Builder image (`Dockerfile.builder`)
- Based on `quay.io/projectquay/golang:1.25` (same Go version as production images)
- Includes all build tools at pinned versions: golangci-lint, typos, kind, kubectl, kustomize, docker CLI + buildx, podman
- Go module and build caches use named volumes for persistence across container runs and image rebuilds
- Stamp file (`.builder.stamp`) tracks when the image needs rebuilding

### Makefile
- All build, test, lint, and format targets run inside the builder container via `BUILDER_RUN` / `sh -c`
- Three flag profiles for different target needs:
  - `BUILDER_RUN` — basic targets (build, test-unit, lint, format, clean, go-mod-check)
  - `BUILDER_RUN_CLUSTER` — pre-existing cluster access (test-integration, bench-tokenizer): adds kubeconfig mount and host network
  - `BUILDER_E2E_FLAGS` — e2e tests: adds host network, container socket, and `CONTAINER_RUNTIME` env (no kubeconfig — kind manages its own)
- Docker/podman compatibility: automatic user mapping, socket detection, GID passthrough for docker socket access, SELinux label=disable for podman socket access
- `LINT_NEW_ONLY` wired through to containerized golangci-lint via `LINT_ARGS`
- `image-build-uds-tokenizer` builds inside the container using `$CONTAINER_RUNTIME` (docker or podman) via socket mount
- Added `builder-shell`, `builder-cluster-shell`, `builder-e2e-shell` targets for debugging
- Removed all host-based tool installation (go-install-tool, install-ginkgo, install-kustomize, check-go, check-ginkgo, etc.)
- Removed dead code: typos arch mapping block, `CGO_ENABLED`, `SRC`, `BUILDER`, `LOCALBIN`, `LOCALLIB`

### CI workflow
- Removed Go setup, go mod tidy, and golangci-lint-action steps
- All CI steps now delegate to containerized make targets
- Expanded paths filter to cover `Dockerfile.*`, `go.sum`, `.golangci.yml`

### Makefile.tools.mk
- Only retains host checks needed for cluster deployment: `check-container-tool`, `check-kubectl`, `check-kustomize`, `check-envsubst`

Fixes https://github.com/llm-d/llm-d-inference-scheduler/issues/470

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"io"
 	"os/exec"
-	"runtime"
 	"strings"
 	"testing"
 	"time"
@@ -176,28 +175,10 @@ func setupK8sCluster() {
 }
 
 func kindLoadImage(image string) {
-	tempDir := ginkgo.GinkgoT().TempDir()
-	target := tempDir + "/container.tar"
-
 	ginkgo.By(fmt.Sprintf("Loading %s into the cluster %s using %s", image, kindClusterName, containerRuntime))
 
-	_, err := exec.LookPath(containerRuntime)
-	gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "Could not find %s in PATH", containerRuntime)
-
-	saveArgs := []string{"save", "--output", target}
-	if containerRuntime == "docker" {
-		// The platform flag is required for docker save to work but it is an unsupported flag for podman
-		saveArgs = append(saveArgs, "--platform", "linux/"+runtime.GOARCH)
-	}
-	saveArgs = append(saveArgs, image)
-
-	command := exec.Command(containerRuntime, saveArgs...)
+	command := exec.Command("kind", "--name", kindClusterName, "load", "docker-image", image)
 	session, err := gexec.Start(command, ginkgo.GinkgoWriter, ginkgo.GinkgoWriter)
-	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-	gomega.Eventually(session).WithTimeout(600 * time.Second).Should(gexec.Exit(0))
-
-	command = exec.Command("kind", "--name", kindClusterName, "load", "image-archive", target)
-	session, err = gexec.Start(command, ginkgo.GinkgoWriter, ginkgo.GinkgoWriter)
 	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
 	gomega.Eventually(session).WithTimeout(600 * time.Second).Should(gexec.Exit(0))
 }


elevran:
Container build fails:
```==== Building image ghcr.io/llm-d/llm-d-builder:dev ====
docker build --quiet \
	-f Dockerfile.builder \
	-t ghcr.io/llm-d/llm-d-builder:dev
ERROR: docker: 'docker buildx build' requires 1 argument

Usage:  docker buildx build [OPTIONS] PATH | URL | -

Run 'docker buildx build --help' for more information
make: *** [Makefile:135: image-build-builder] Error 1
Error: Process completed with exit code 2.```

File: .gitignore [hunk 1]
```diff
@@ -39,3 +39,4 @@ vendor
 
 # Build output
 /build
+.builder.stamp
```
elevran: is it possible to put all stamps under the build directory so they're automatically ignored? We may want to do so for various other PHONY targets.

Makefile
```diff
@@ -156,44 +181,45 @@ test: test-unit test-e2e ## Run all tests (unit and e2e)
...
 .PHONY: test-integration
-test-integration: ## Run integration tests
+test-integration: image-build-builder ## Run integration tests (requires KUBECONFIG and running cluster)
 	@printf "\033[33;1m==== Running Integration Tests ====\033[0m\n"
-	go test -v -tags=integration_tests ./test/integration/
+	$(BUILDER_RUN_CLUSTER) 'go test -v -tags=integration_tests ./test/integration/'
 
 .PHONY: test-e2e
-test-e2e: image-build image-build-uds-tokenizer image-pull ## Run end-to-end tests against a new kind cluster
+test-e2e: image-build-builder image-build image-build-uds-tokenizer image-pull ## Run end-to-end tests against a new kind cluster
```
elevran: See #739 
We would want to consume UDS as a dependency image and not build it in inference-scheduler.

Makefile
```diff
@@ -37,66 +34,102 @@ UDS_TOKENIZER_TAG ?= dev
...
+BUILDER_RUN = $(CONTAINER_RUNTIME) run $(BUILDER_RUN_FLAGS) $(BUILDER_IMAGE) sh -c
+BUILDER_RUN_CLUSTER = $(CONTAINER_RUNTIME) run $(BUILDER_RUN_FLAGS) $(BUILDER_CLUSTER_FLAGS) $(BUILDER_IMAGE) sh -c
 
 # Linker flags for go build inside Docker images.
 # Default strips debug symbols for smaller production images.
 # Override locally to build a debuggable image: LDFLAGS="" make image-build-epp
 LDFLAGS ?= -s -w
 
+# test packages
+epp_TEST_PACKAGES = $$(go list ./... | grep -v /test/ | grep -v ./pkg/sidecar/)
```
elevran: I think `epp_TEST_PACKAGES = $$(go list ./... | grep -v ...)` stores the value `$(go list ./... | grep -v ...)` in the variable (i.e., the `$$` collapses to `$`)?

If that's correct, then when referenced in a recipe as `$($*_TEST_PACKAGES)`, make expands it to `$(go list ./... | ...)` and then looks it up, The variable name would not exist and result in empty string, and the resulting command would be `docker run ... sh -c 'go test -v '`. The invocation of `go test -v` with no packages only test the root, skipping all other test packages. This is likely not what we would want for test coverage.

A possible fix tould be to inline in the recipe string (i.e., where `go test` is called) or consider using the same syntax as before:
```diff
-       go test -v -run "$(PATTERN)" $$($(epp_TEST_FILES) | tr '\n' ' '); \
+		$(BUILDER_RUN) 'go test -v -run "$(PATTERN)" $(epp_TEST_PACKAGES)'; \. 
```

Makefile
```diff
@@ -37,66 +34,102 @@ UDS_TOKENIZER_TAG ?= dev
...
+ifeq ($(CONTAINER_RUNTIME),podman)
+CONTAINER_SOCK ?= $(or $(shell podman info --format '{{.Host.RemoteSocket.Path}}' 2>/dev/null),/run/podman/podman.sock)
+BUILDER_SOCK_FLAGS = --security-opt label=disable \
+	-v $(CONTAINER_SOCK):$(CONTAINER_SOCK) \
+	-e CONTAINER_HOST=unix://$(CONTAINER_SOCK) \
+	-e CONTAINER_RUNTIME=podman \
+	-e KIND_EXPERIMENTAL_PROVIDER=podman
+else
+CONTAINER_SOCK ?= /var/run/docker.sock
+DOCKER_SOCK_GID := $(shell stat -c '%g' $(CONTAINER_SOCK) 2>/dev/null)
```
elevran, line 86: nit: `-c` might be Linux or GNU specific. Does work on MacOS?

.github/workflows/ci-pr-checks.yaml
```diff
@@ -43,24 +44,8 @@ jobs:
       - name: Sanity check repo contents
         run: ls -la
 
-      - name: Extract Go version from go.mod
-        run: sed -En 's/^go (.*)$/GO_VERSION=\1/p' go.mod >> $GITHUB_ENV
-
-      - name: Set up Go with cache
```
elevran, line 44: Q: consider if this has a side effect of removing Go module caching via `actions/cache` (AFAIK the setup Go and go mod gave us the caching). The Docker volumes only persist on the same physical host and not on ephemeral GH runners, so volumes are lost every run. If so, consider adding `--cache-from`/`--cache-to` for the builder image build, or use `actions/cache` to persist the Docker volumes explicitly.

Dockerfile.builder
```diff
@@ -0,0 +1,64 @@
...
+ARG TYPOS_VERSION=v1.34.0
+ARG KIND_VERSION=v0.27.0
+ARG GOLANGCI_LINT_VERSION=v2.8.0
+ARG KUBECTL_VERSION=v1.35.3
+ARG KUSTOMIZE_VERSION=v5.6.0
+ARG DOCKER_VERSION=29.3.0
+ARG DOCKER_BUILDX_VERSION=v0.32.1
+
+RUN dnf install -y \
+    gcc-c++ podman && \
```
elevran: `gcc-c++` no longer needed as we set `CGO=0` in builds. The dependency was removed.

Makefile
```diff
@@ -119,35 +152,27 @@ signed-commits-check:
...
+	rm -rf bin build $(BUILDER_STAMP)
+	-$(BUILDER_RUN) 'go clean -testcache -cache'
 
 .PHONY: format
-format: check-golangci-lint ## Format Go source files
+format: image-build-builder ## Format Go source files
 	@printf "\033[33;1m==== Running go fmt ====\033[0m\n"
-	@gofmt -l -w $(SRC)
-	$(GOLANGCI_LINT) fmt
+	$(BUILDER_RUN) 'gofmt -l -w . && golangci-lint fmt'
```
elevran: missing config file in `golangci-lint fmt` means we use the default configuration not project level config.

Makefile
```diff
@@ -156,44 +181,45 @@ test: test-unit test-e2e ## Run all tests (unit and e2e)
...
 	@printf "\033[33;1m==== Running Integration Tests ====\033[0m\n"
-	go test -v -tags=integration_tests ./test/integration/
+	$(BUILDER_RUN_CLUSTER) 'go test -v -tags=integration_tests ./test/integration/'
 
 .PHONY: test-e2e
-test-e2e: image-build image-build-uds-tokenizer image-pull ## Run end-to-end tests against a new kind cluster
+test-e2e: image-build-builder image-build image-build-uds-tokenizer image-pull ## Run end-to-end tests against a new kind cluster
 	@printf "\033[33;1m==== Running End to End Tests ====\033[0m\n"
-	PATH=$(LOCALBIN):$$PATH ./test/scripts/run_e2e.sh
+	$(CONTAINER_RUNTIME) run $(BUILDER_RUN_FLAGS) $(BUILDER_E2E_FLAGS) \
```
elevran, line 223: Q: `test-e2e` container invocation doesn't pass `CONTAINER_RUNTIME`, so any e2e test code would use `docker` even when the host is using `podman`. Intentional?

Makefile
```diff
@@ -203,35 +229,23 @@ post-deploy-test: ## Run post deployment tests
...
 
 .PHONY:	image-build
 image-build: image-build-epp image-build-sidecar image-build-uds-tokenizer ## Build Container image using $(CONTAINER_RUNTIME)
 
-# Path to kv-cache repo for UDS tokenizer image build (can be overridden)
-KV_CACHE_PATH ?= $(shell go list -m -f '{{.Dir}}' github.com/llm-d/llm-d-kv-cache 2>/dev/null)
-
 .PHONY: image-build-uds-tokenizer
-image-build-uds-tokenizer: check-container-tool ## Build UDS tokenizer image from kv-cache
+image-build-uds-tokenizer: check-container-tool image-build-builder ## Build UDS tokenizer image from kv-cache
```
elevran: #739

.gitignore
acardace: Yes, will do.

Makefile
acardace: very good catch, Will just use double quotes instead which fixes it.

Makefile
acardace: `CONTAINER_RUNTIME` is passed down, it's in the `BUILDER_SOCK_FLAGS`.

.github/workflows/ci-pr-checks.yaml
acardace: You're right, this is CI regression. I've fixed it now by caching the volumes.

Makefile
acardace: Fixed.

.gitignore
acardace: Done.

Makefile
acardace: Fixed.

Makefile
acardace: Fixed.

Dockerfile.builder
acardace: Good catch, removed it.

Makefile
acardace: Fixed.

Makefile
acardace: Removed it.

Issue #515, opened 2025-12-17
No Hit LRU plugin needs to become P/D aware


kfswain:
Discovered as part of: https://github.com/llm-d/llm-d/issues/550

No Hit LRU assumes only the [`PrimaryProfileName`](https://github.com/llm-d/llm-d-inference-scheduler/blob/12a5902522c87f88bef0033bb6694740e0448068/pkg/plugins/scorer/no_hit_lru.go#L283) results should be considered. This does not allow the plugin to work in P/D scenarios

To work with P/D we should instead:

- Instead include the `prefill` profile and the PrimaryProfileName to be added to the LRU (this should work fine since the prefill and decode workers are filtered seperately)

evacchi:
attempting with https://github.com/llm-d/llm-d-inference-scheduler/pull/522

elevran:
Fixed by #522

PR #522, opened 2025-12-23
feat: make no-hit-lru P/D-aware

evacchi:
an attempt at addressing #515. Still navigating the codebase, this has been written with obvious assistance (mostly the test cases), let me know if I got it right.

Essentially, 
- we now accept an optional `prefillProfile` config parameter in the style of [pd_prerequest.go](https://github.com/llm-d/llm-d-inference-scheduler/blob/5176573e4df75b9868b1da8bc6421cf5198e7aac/pkg/plugins/pre-request/pd_prerequest.go#L25), defaulting to `"prefill"`.
- we _unconditionally_ check and move to the front of the LRU cache the pod for the `prefillProfile` key in the `schedulingResult` after we have checked for `schedulingResult.PrimaryProfileName`

not 100% sure this is what was requested originally 😬

### DIFF
--- a/pkg/plugins/scorer/no_hit_lru.go
+++ b/pkg/plugins/scorer/no_hit_lru.go
@@ -292,17 +292,16 @@ func (s *NoHitLRU) PreRequest(ctx context.Context, request *types.LLMRequest, sc
 		return
 	}
 
-	s.moveTargetPodToFront(ctx, request, schedulingResult.ProfileResults[schedulingResult.PrimaryProfileName], schedulingResult.PrimaryProfileName)
-	s.moveTargetPodToFront(ctx, request, schedulingResult.ProfileResults[defaultPrefillProfile], defaultPrefillProfile)
+	if targetProfile, ok := schedulingResult.ProfileResults[schedulingResult.PrimaryProfileName]; ok && targetProfile != nil && len(targetProfile.TargetPods) != 0 {
+		s.moveTargetPodToFront(ctx, request, targetProfile, schedulingResult.PrimaryProfileName)
+	}
+	if targetProfile, ok := schedulingResult.ProfileResults[defaultPrefillProfile]; ok && targetProfile != nil && len(targetProfile.TargetPods) != 0 {
+		s.moveTargetPodToFront(ctx, request, targetProfile, defaultPrefillProfile)
+	}
 }
 
 func (s *NoHitLRU) moveTargetPodToFront(ctx context.Context, request *types.LLMRequest, targetProfile *types.ProfileRunResult, profileName string) {
 	logger := log.FromContext(ctx).V(logutil.DEBUG)
-
-	// Get the target profile's target pod
-	if targetProfile == nil || len(targetProfile.TargetPods) == 0 {
-		return
-	}
 
 	targetPod := targetProfile.TargetPods[0]
 	podName := targetPod.GetPod().NamespacedName.String()
--- a/pkg/plugins/scorer/no_hit_lru_test.go
+++ b/pkg/plugins/scorer/no_hit_lru_test.go
@@ -550,37 +550,6 @@ func TestNoHitLRUPrefillDecodeTracking(t *testing.T) {
 		}
 	})
 
-	t.Run("custom prefill profile name", func(t *testing.T) {
-		// Create scorer with custom prefill profile name
-		scorer := scorer.NewNoHitLRU(ctx, &scorer.NoHitLRUParameters{
-			PrefillProfile: "custom-prefill",
-		})
-
-		req := &types.LLMRequest{RequestId: "custom-prefill-request"}
-		scorer.Score(ctx, coldPrefixState, req, prefillPods)
-
-		result := &types.SchedulingResult{
-			PrimaryProfileName: "decode",
-			ProfileResults: map[string]*types.ProfileRunResult{
-				"custom-prefill": {
-					TargetPods: []types.Pod{prefillPodA},
-				},
-				"decode": {
-					TargetPods: []types.Pod{decodePodA},
-				},
-			},
-		}
-		scorer.PreRequest(ctx, req, result)
-
-		// Verify custom prefill profile was tracked
-		req2 := &types.LLMRequest{RequestId: "custom-prefill-request-2"}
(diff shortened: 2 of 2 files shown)


evacchi:
squashed, rebased

pkg/plugins/scorer/no_hit_lru.go
```diff
@@ -36,6 +39,10 @@ type NoHitLRUParameters struct {
 	// Defaults to "prefix-cache-scorer".
 	PrefixPluginName string `json:"prefixPluginName"`
 
+	// PrefillProfile defines the name of the prefill profile to track in LRU.
+	// Defaults to "prefill".
+	PrefillProfile string `json:"prefillProfile"`
```
vMaroon: I think this requirement may degrade UX (need more opinions @elevran @kfswain) - what if instead we add a switch to bump selected pods in all scheduling profiles, and turn it on by default?

I think currently the only scenario in which multiple profiles are used is P/D. I could be mistaken though.

vMaroon:
No problem at all - the proposed approach is very legitimate, I only worry about introducing more manual configuration especially if can be avoided. The latter may be questionable though, let's see what others think.

pkg/plugins/scorer/no_hit_lru.go
ahg-g: In igw, for the [approximate prefix cache, we just hardcoded the profile name](https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2104/changes) for now until we have a proper interface. I suggest we do the same here.

pkg/plugins/scorer/no_hit_lru.go
```diff
@@ -286,19 +292,25 @@ func (s *NoHitLRU) PreRequest(ctx context.Context, request *types.LLMRequest, sc
...
+	s.moveTargetPodToFront(ctx, request, schedulingResult.ProfileResults[schedulingResult.PrimaryProfileName], schedulingResult.PrimaryProfileName)
+	s.moveTargetPodToFront(ctx, request, schedulingResult.ProfileResults[defaultPrefillProfile], defaultPrefillProfile)
+}
+
+func (s *NoHitLRU) moveTargetPodToFront(ctx context.Context, request *types.LLMRequest, targetProfile *types.ProfileRunResult, profileName string) {
+	logger := log.FromContext(ctx).V(logutil.DEBUG)
+
+	// Get the target profile's target pod
+	if targetProfile == nil || len(targetProfile.TargetPods) == 0 {
+		logger.Info("No target pod in profile", "profile", profileName)
```
ahg-g: this will cause a lot of log spam in non-disagg mode, I would remove it

pkg/plugins/scorer/no_hit_lru.go
```diff
@@ -286,19 +292,24 @@ func (s *NoHitLRU) PreRequest(ctx context.Context, request *types.LLMRequest, sc
 		return
 	}
 
-	// Get the primary profile's target pod
-	primaryProfile := schedulingResult.ProfileResults[schedulingResult.PrimaryProfileName]
-	if primaryProfile == nil || len(primaryProfile.TargetPods) == 0 {
-		logger.Info("No target pod in primary profile")
+	s.moveTargetPodToFront(ctx, request, schedulingResult.ProfileResults[schedulingResult.PrimaryProfileName], schedulingResult.PrimaryProfileName)
+	s.moveTargetPodToFront(ctx, request, schedulingResult.ProfileResults[defaultPrefillProfile], defaultPrefillProfile)
```
kfswain: I think we should check if the `defaultPrefillProfile` key exists

pkg/plugins/scorer/no_hit_lru.go
```diff
@@ -286,19 +292,24 @@ func (s *NoHitLRU) PreRequest(ctx context.Context, request *types.LLMRequest, sc
...
+}
+
+func (s *NoHitLRU) moveTargetPodToFront(ctx context.Context, request *types.LLMRequest, targetProfile *types.ProfileRunResult, profileName string) {
+	logger := log.FromContext(ctx).V(logutil.DEBUG)
+
+	// Get the target profile's target pod
+	if targetProfile == nil || len(targetProfile.TargetPods) == 0 {
```
kfswain: Related to the other comment: doing the defensive check here is great, I think it may be more readable to bubble it up.

evacchi: ok I think it should be better now!

evacchi:
> @evacchi note the lint failure

The lint was actually due to a test I forgot to delete: it configured a custom name for `prefillProfile`!

pkg/plugins/scorer/no_hit_lru.go
```diff
@@ -286,19 +292,24 @@ func (s *NoHitLRU) PreRequest(ctx context.Context, request *types.LLMRequest, sc
...
-	if primaryProfile == nil || len(primaryProfile.TargetPods) == 0 {
-		logger.Info("No target pod in primary profile")
-		return
+	if targetProfile, ok := schedulingResult.ProfileResults[schedulingResult.PrimaryProfileName]; ok && targetProfile != nil && len(targetProfile.TargetPods) != 0 {
+		s.moveTargetPodToFront(ctx, request, targetProfile, schedulingResult.PrimaryProfileName)
 	}
+	if targetProfile, ok := schedulingResult.ProfileResults[defaultPrefillProfile]; ok && targetProfile != nil && len(targetProfile.TargetPods) != 0 {
+		s.moveTargetPodToFront(ctx, request, targetProfile, defaultPrefillProfile)
+	}
+
```
elevran: nit: extra empty line
```suggestion
```

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -31,7 +31,7 @@ type PrecisePrefixCachePluginConfig struct {
 	IndexerConfig *kvcache.Config `json:"indexerConfig"`
 	// KVEventsConfig holds the configuration for the `kvevents.Pool` which is
 	// used to subscribe to KV-cache events and update the internal KV-cache
-	// index state.
+	// index st	ate.
```
elevran: This seems like an unintended change.
Please revert

evacchi: whoops

Issue #524, opened 2025-12-25
Improve documentation on multiple GenAI models and LoRAs


davidbreitgand:
Update the [architecture description](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/docs/architecture.md) about supporting multiple GenAI models in the same cluster through GIE. 

At the moment, the architecture Readme states that this is not possible. 

Suggestion: to mention BBR and the updated IGW Guide on supporting multiple GenAI models and LoRAs

This is also related to issue #415

PR #525, opened 2025-12-25
Updates the architecture description 

davidbreitgand:
To reflect support for multiple GenAI models and LoRAs. This will remove confusion about llm-d supporting only one model per cluster

Solves issue #524


davidbreitgand:
@nirrozenbaum , can you please take a look at this? Thanks.

PR #550, opened 2026-01-10
implement context-length-aware plugin (scorer/filter)

vMaroon:
## Summary

This PR adds a **Context Length Aware** scheduling plugin that routes inference requests based on their **estimated context length**.  
It enables efficient use of **heterogeneous context-length pools** by matching requests to pods configured for specific context ranges.

An engine's context length is determined by multiple factors that essentially boil down to the available KV-cache size and batching/concurrency limitations given explicit/implicit max-input-length.

This plugin has two immediate use-cases:
1. Avoid request rejection in environments where different pods may serve different max prompt lengths
2. Better place requests based on hardware/acceleration

### Key Points

- Supports **scorer** (soft preference) and **filter** (hard constraint) modes
- Uses simple **pod labels** to declare supported context length ranges
- Unlabeled pods are handled safely

### Example Configuration

**As a Scorer (default):**
```yaml
plugins:
  - type: context-length-aware
    parameters:
      mode: score
      label: llm-d.ai/context-length-range
schedulingProfiles:
  - name: default
    plugins:
      - pluginRef: context-length-aware
        weight: 3
```

**Pod Labels:**
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: vllm-short-context
  labels:
    llm-d.ai/context-length-range: "0-2048"
---
apiVersion: v1
kind: Pod
metadata:
  name: vllm-long-context
  labels:
    llm-d.ai/context-length-range: "8192-32768"
```

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -578,9 +578,10 @@ schedulingProfiles:
 
 #### ContextLengthAware
 
-A multi-purpose plugin that can operate as both a **Filter** and a **Scorer** to route inference requests
-based on context length (token count). This enables optimized resource allocation by directing requests to
-pods configured for specific context length ranges.
+A **Scorer** plugin that routes inference requests based on context length (token count), with optional
+**filtering** gated behind `enableFiltering`. Scoring is always applied; filtering is off by default.
+This enables optimized resource allocation by directing requests to pods configured for specific
+context length ranges.
 
 **Use Cases:**
 - Route short prompts to pods with smaller GPU memory
@@ -589,11 +590,11 @@ pods configured for specific context length ranges.
 - Support heterogeneous deployments with different GPU configurations
 
 The plugin scores all pods based on how well their ranges match the request:
-- **In-range match (0.5–1.0):** Higher scores for tighter/more specific ranges (specialized pods), lower scores for very wide ranges (generalist pods)
-- **Out-of-range fallback (0.0–0.3):** When no range matches, pods are ranked by proximity to the request. For example, a 9000-token request prefers a pod with `max=8192` over one with `max=2048`. Fallback scores never exceed 0.3, so an in-range match always wins.
-- **Neutral score (0.5):** Pods without the context length label
+- **In-range match (0.3–1.0]:** Higher scores for tighter/more specific ranges (specialized pods), lower scores for very wide ranges (generalist pods). In-range scores are always strictly above 0.3, guaranteeing they beat any out-of-range fallback.
+- **Out-of-range fallback [0.0–0.3):** When no range matches, pods are ranked by proximity to the request. For example, a 9000-token request prefers a pod with `max=8192` over one with `max=2048`.
+- **Neutral score (0.5):** Pods without the context length label.
 
-If `enableFiltering` is set to true, the plugin also filters out pods that do not match the request's context length.
+When `enableFiltering` is set to true, the plugin also filters out pods whose range does not contain the request's context length.
 
 **Configuration:**
 
@@ -620,13 +621,13 @@ When the tokenizer scorer is not configured, the plugin falls back to character-
 
 **Label Format:**
 
-Each pod should be labeled with a single context length range using the format `"min-max"`, where _min_ and _max_ are both positive integers:
+Pods should be labeled with context length ranges using the format `"min-max"`, where _min_ and _max_ are both positive integers:
 
 ```yaml
 llm-d.ai/context-length-range: "0-2048"
 ```
 
-**Example - Scorer with Precise Tokenization:**
+**Example Configuration - Scorer:**
 
 ```yaml
 plugins:
@@ -652,7 +653,7 @@ schedulingProfiles:
       - pluginRef: max-score-picker
 ```
 
-**Example - Filter with Estimation Fallback (no tokenizer):**
+**Example Configuration - Scorer with Filtering Enabled:**
 
 ```yaml
 plugins:
--- a/pkg/plugins/multi/context_length_aware.go
+++ b/pkg/plugins/multi/context_length_aware.go
@@ -29,7 +29,7 @@ const (
 	charToTokenMultiplier = 0.25
 )
 
-type contextLengthAwareParams struct {
+type contextLengthAwareParameters struct {
 	// Label is the pod label name to check for context length range.
 	// Format expected: "min-max" (e.g., "0-2048" or "2048-8192"), where min and max are positive integers.
 	Label string `json:"label"`
@@ -52,7 +52,7 @@ var _ scheduling.Scorer = &ContextLengthAware{} // validate interface conformanc
 
 // ContextLengthAwareFactory defines the factory function for the ContextLengthAware plugin.
 func ContextLengthAwareFactory(name string, rawParameters json.RawMessage, _ plugin.Handle) (plugin.Plugin, error) {
-	parameters := &contextLengthAwareParams{
+	parameters := &contextLengthAwareParameters{
 		Label:           DefaultContextLengthLabel,
 		EnableFiltering: false,
 	}
@@ -71,7 +71,7 @@ func ContextLengthAwareFactory(name string, rawParameters json.RawMessage, _ plu
 }
 
 // NewContextLengthAware creates and returns an instance of the ContextLengthAware plugin.
-func NewContextLengthAware(name string, params *contextLengthAwareParams) *ContextLengthAware {
+func NewContextLengthAware(name string, params *contextLengthAwareParameters) *ContextLengthAware {
 	return &ContextLengthAware{
 		typedName:       plugin.TypedName{Type: ContextLengthAwareType, Name: name},
 		labelName:       params.Label,
@@ -119,17 +119,19 @@ func (p *ContextLengthAware) Filter(ctx context.Context, cycleState *scheduling.
 	contextLength, usedTokenizer := p.getContextLength(ctx, cycleState, request)
 	logger.V(logutil.TRACE).Info("Filtering endpoints by context length", "contextLength", contextLength, "usedTokenizer", usedTokenizer)
 
-	var filteredEndpoints []scheduling.Endpoint
+	filteredEndpoints := []scheduling.Endpoint{}
 
 	for _, endpoint := range endpoints {
 		metadata := endpoint.GetMetadata()
 		if metadata == nil {
+			// Endpoints without metadata are included (treated as no-label).
+			filteredEndpoints = append(filteredEndpoints, endpoint)
 			continue
 		}
 
 		rangeStr, hasLabel := metadata.Labels[p.labelName]
 		if !hasLabel {
-			// Endpoints without the label are included (they accept any context length)
+			// Endpoints without the label are included (they accept any context length).
 			filteredEndpoints = append(filteredEndpoints, endpoint)
 			continue
 		}
@@ -276,31 +278,36 @@ func parseContextRange(rangeStr string) (contextRange, error) {
 // calculateRangeScore calculates a score for how well a pod's context range matches the request.
 //
 // Scoring tiers (higher tier always wins over lower):
-//   - In-range match (0.5–1.0): exact or tight matches score highest, wide ranges lower.
-//   - Out-of-range fallback (0.0–0.3): scored by proximity to the nearest boundary,
-//     so the pod with the largest max is preferred for oversized requests.
+//   - In-range match (0.3–1.0]: tighter ranges and more headroom score higher.
+//     The minimum in-range score is strictly above maxFallbackScore, so an in-range
+//     match always beats any out-of-range fallback.
+//   - Out-of-range fallback [0.0–0.3): scored by proximity to the nearest boundary,
+//     so the pod closest to the request is preferred for out-of-range requests.
 //
 // This ensures that out-of-range requests still get routed to the most reasonable pod
 // rather than being scored equally (0.0) and selected arbitrarily.
 func calculateRangeScore(contextLength int, r contextRange) float64 {
-	// In-range match
+	const maxFallbackScore = 0.3
+
+	// In-range match: score in (maxFallbackScore, 1.0].
 	if contextLength >= r.min && contextLength <= r.max {
 		rangeWidth := r.max - r.min
 		if rangeWidth == 0 {
 			return 1.0
 		}
 
+		// rawScore is in [0.0, 1.0]: tighter ranges and more headroom score higher.
 		widthScore := 1.0 / (1.0 + float64(rangeWidth)/10000.0)
 		headroom := float64(r.max - contextLength)
 		positionScore := headroom / float64(rangeWidth)
+		rawScore := 0.7*widthScore + 0.3*positionScore
 
-		return 0.7*widthScore + 0.3*positionScore
+		// Map rawScore from [0.0, 1.0] into (maxFallbackScore, 1.0].
+		return maxFallbackScore + rawScore*(1.0-maxFallbackScore)
 	}
 
 	// Out-of-range fallback: score by proximity to the nearest range boundary.
-	// Capped at 0.3 so a fallback never outscores an in-range match.
-	const maxFallbackScore = 0.3
-
+	// Strictly below maxFallbackScore so a fallback never outscores an in-range match.
 	var distance int
 	if contextLength > r.max {
 		distance = contextLength - r.max
--- a/pkg/plugins/multi/context_length_aware_test.go
+++ b/pkg/plugins/multi/context_length_aware_test.go
@@ -99,7 +99,7 @@ func TestContextLengthAwareFilter(t *testing.T) {
 			map[string]string{}),
 	}
 
-	params := &contextLengthAwareParams{
+	params := &contextLengthAwareParameters{
 		Label:           DefaultContextLengthLabel,
 		EnableFiltering: true,
 	}
@@ -136,7 +136,7 @@ func TestContextLengthAwareScore(t *testing.T) {
 			map[string]string{}),
 	}
 
-	params := &contextLengthAwareParams{
+	params := &contextLengthAwareParameters{
 		Label:           DefaultContextLengthLabel,
 		EnableFiltering: false,
 	}
@@ -146,13 +146,15 @@ func TestContextLengthAwareScore(t *testing.T) {
 	scores := plugin.Score(ctx, nil, request, endpoints)
 
 	// With context length 0:
-	// - tight-range (0-20): should score high (in-range match)
-	// - wide-range (0-10000): should score lower than tight (in-range but wide)
-	// - no-match (500-1000): out-of-range fallback, scored by proximity (0 < score <= 0.3)
-	// - no-label: should score 0.5 (neutral)
+	// - tight-range (0-20): in-range, should score high (> 0.3)
+	// - wide-range (0-10000): in-range but wide, should score lower than tight but still > 0.3
+	// - no-match (500-1000): out-of-range fallback (0 < score < 0.3)
+	// - no-label: neutral (0.5)
 	assert.Greater(t, scores[endpoints[0]], scores[endpoints[1]], "tight range should score higher than wide range")
+	assert.Greater(t, scores[endpoints[0]], 0.3, "in-range score must be strictly above 0.3")
+	assert.Greater(t, scores[endpoints[1]], 0.3, "in-range score must be strictly above 0.3")
 	assert.Greater(t, scores[endpoints[2]], 0.0, "out-of-range should get a fallback score > 0")
-	assert.LessOrEqual(t, scores[endpoints[2]], 0.3, "out-of-range fallback should not exceed 0.3")
+	assert.Less(t, scores[endpoints[2]], 0.3, "out-of-range fallback must be strictly below 0.3")
 	assert.Greater(t, scores[endpoints[1]], scores[endpoints[2]], "in-range match should outscore out-of-range fallback")
 	assert.Equal(t, 0.5, scores[endpoints[3]], "no label should score 0.5")
 }
@@ -210,9 +212,9 @@ func TestCalculateRangeScoreFallback(t *testing.T) {
 		largeMax := calculateRangeScore(9000, contextRange{min: 0, max: 8192})
 
 		assert.Greater(t, smallMax, 0.0)
-		assert.LessOrEqual(t, smallMax, 0.3)
+		assert.Less(t, smallMax, 0.3)
 		assert.Greater(t, largeMax, 0.0)
-		assert.LessOrEqual(t, largeMax, 0.3)
+		assert.Less(t, largeMax, 0.3)
 		assert.Greater(t, largeMax, smallMax, "pod with larger max should score higher")
 	})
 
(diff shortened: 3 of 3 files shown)


pkg/plugins/multi/context_length_aware_test.go
```diff
@@ -0,0 +1,442 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tokens := estimateContextLength(tt.request)
+			assert.Equal(t, tt.expected, tokens)
+		})
+	}
+}
```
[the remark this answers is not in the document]
vMaroon: Overkill, out of scope.

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+
+		maxVal, err := strconv.Atoi(strings.TrimSpace(bounds[1]))
+		if err != nil {
+			return nil, fmt.Errorf("invalid max value: %s", bounds[1])
+		}
+
+		if minVal < 0 || maxVal < 0 {
+			return nil, fmt.Errorf("negative values not allowed: min=%d, max=%d", minVal, maxVal)
+		}
```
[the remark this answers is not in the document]
vMaroon: No negative context lengths.

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,315 @@
...
+
+	if rawParameters != nil {
+		if err := json.Unmarshal(rawParameters, &parameters); err != nil {
+			return nil, fmt.Errorf("failed to parse the parameters of the '%s' plugin - %w", ContextLengthAwareType, err)
+		}
+	}
+
+	if name == "" {
+		return nil, fmt.Errorf("invalid configuration for '%s' plugin: name cannot be empty", ContextLengthAwareType)
+	}
```
shmuelk: Name will always have a value. If one isn't specified in the configuration, it is defaulted to the type.

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,315 @@
...
+
+	if parameters.Label == "" {
+		return nil, fmt.Errorf("invalid configuration for '%s' plugin: 'label' must be specified", ContextLengthAwareType)
```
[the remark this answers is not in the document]
shmuelk: If the label field was forced to be empty in the JSON, unlikely I think, I'd just use the default value rather than issue an error message.

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+	}
+
+	logger.Info("Filtered pods", "originalCount", len(pods), "filteredCount", len(filteredPods))
+	return filteredPods
+}
+
+// Score scores pods based on how well their context length ranges match the request
+// Pods with tighter/more specific ranges matching the request get higher scores.
+func (p *ContextLengthAware) Score(ctx context.Context, _ *types.CycleState, request *types.LLMRequest, pods []types.Pod) map[types.Pod]float64 {
```
shmuelk: The code scores even if in Filter mode?

pkg/plugins/multi/context_length_aware.go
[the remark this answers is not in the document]
shmuelk: You are checking for negative numbers in the range, which will never happen.

A split on "-", will never leave negative numbers to parse.

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,315 @@
...
+			// Score based on how tight the range is
+			// Normalize by position within range and range width
+			// Tighter ranges get higher scores (up to 1.0)
+			// Wider ranges get lower scores (approaching 0.5)
+
+			// Calculate range width score (narrower is better)
+			// Use log scale to handle very large ranges
+			widthScore := 1.0 / (1.0 + float64(rangeWidth)/10000.0)
+
+			// Calculate position score (closer to middle is better)
```
shmuelk: Why is closer to the middle better and no closer to the minimum?

pkg/plugins/multi/context_length_aware_bench_test.go
```diff
@@ -0,0 +1,236 @@
...
+		case 3:
+			label = map[string]string{DefaultContextLengthLabel: "0-1024,4096-16384"}
+		case 4:
+			label = map[string]string{} // no label
+		}
+		pods[i] = &types.PodMetrics{
+			Pod: &backend.Pod{
+				NamespacedName: k8stypes.NamespacedName{
+					Namespace: "default",
+					Name:      "pod-" + string(rune(i)),
```
shmuelk:
```suggestion
					Name:      fmt.Sprintf("pod-%d", i),
```

shmuelk (changes requested):

pkg/plugins/multi/context_length_aware.go
vMaroon: Yes, after filtering, there may be several groups with fitting boundaries - the scoring makes it so that the tighter groups are preferred.

pkg/plugins/multi/context_length_aware.go
vMaroon: The hard user requirement is fitting between disjoint groups, the scoring here is one algorithm for placement across intersecting ranges. I think closer to the minimum may be better, but I'll phrase it as "distance from maximum, more headroom is better".

pkg/plugins/multi/context_length_aware.go
shmuelk: Then the "mode" is a misnomer. It is not either a Filter or a Scorer. It is always a Scorer and sometimes a Filter.

More than that, as it implements the Scorer interface it will always be used as a Scorer.

I think it would be better if this was two separate structs, one the Filter and another the Scorer, that used common helper functions to reuse as much code as possible.

pkg/plugins/multi/context_length_aware_test.go
```diff
@@ -0,0 +1,442 @@
...
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/backend"
+	backendmetrics "sigs.k8s.io/gateway-api-inference-extension/pkg/epp/backend/metrics"
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/scheduling/types"
+
+	"github.com/llm-d/llm-d-inference-scheduler/test/utils"
+)
+
+// Helper functions
+
+func createPod(nsn k8stypes.NamespacedName, ipaddr string, labels map[string]string) types.Pod {
```
shmuelk: Why not just pass in a Name, and default the NameSpace to default?

docs/architecture.md
```diff
@@ -462,6 +462,125 @@ When integrating with a prefix-cache scorer, the prefix-cache scorer should be d
 
 ---
 
+#### ContextLengthAware
+
+A multi-purpose plugin that can operate as both a **Filter** and a **Scorer** to route inference requests
```
elevran: nit: based on some of the review comments, seems it is always used as a **Scorer** and **filtering** is optional?
if that's indeed the case, should be better reflected in the doc as well as in configuration items.

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,302 @@
...
+	DefaultContextLengthLabel = "llm-d.ai/context-length-range"
+
+	// charToTokenMultiplier defines the multiplier to convert characters to tokens
+	// This is an approximate value and may vary based on the tokenizer used
+	charToTokenMultiplier = 0.25
+)
+
+type contextLengthAwareParams struct {
+	// Label is the pod label name to check for context length ranges
+	// Format expected: "min-max" (e.g., "0-2048" or "2048-8192")
```
elevran: nit: consider allowing support for open ended ranges as well? That is, `-<number>` (for lengths up to `<number>`) and `<number>-` (for lengths above the `<number>`).

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,302 @@
...
+	// Multiple ranges can be specified with comma separation (e.g., "0-2048,8192-16384")
```
elevran: what is the motivation for multple non-contiguous ranges? Based on the architecture description it seems less likely.

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+var _ framework.Filter = &ContextLengthAware{} // validate interface conformance
+var _ framework.Scorer = &ContextLengthAware{} // validate interface conformance
+
+// ContextLengthAwareFactory defines the factory function for the ContextLengthAware plugin.
+func ContextLengthAwareFactory(name string, rawParameters json.RawMessage, _ plugins.Handle) (plugins.Plugin, error) {
+	parameters := contextLengthAwareParams{
+		Label: DefaultContextLengthLabel,
+		Mode:  "score", // default to scoring mode
```
elevran: nit: if scroing is always applied, consider renaming this to "enableFiltering" or similar with default disabled.

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,302 @@
...
+	return p.typedName
+}
+
+// WithName sets the name of the plugin.
+func (p *ContextLengthAware) WithName(name string) *ContextLengthAware {
+	p.typedName.Name = name
+	return p
+}
+
+// Filter filters out pods that don't have a context length range matching the request
```
elevran:
```suggestion
// Filter filters out pods that don't have a context length range matching the request.
```

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+
+		ranges, err := parseContextRanges(rangeStr)
+		if err != nil {
+			logger.Error(err, "Failed to parse context range label", "pod", pod.GetPod().NamespacedName, "rangeStr", rangeStr)
+			continue
+		}
+
+		// Check if any range matches
+		if matchesAnyRange(contextLength, ranges) {
```
elevran: nit: consider combining parsing and matching?

pkg/plugins/multi/context_length_aware.go
elevran: 👍 also change the configuration mode to enable filtering (off by default)?

pkg/plugins/multi/context_length_aware.go
elevran: nit: In general, try to be consistent with using `.` at end of comments.

pkg/plugins/multi/context_length_aware.go
vMaroon: Delegating to later followups.

pkg/plugins/multi/context_length_aware.go
vMaroon: For now setting filtering through a field, hope that is acceptable.

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,296 @@
...
+		}
+
+		maxVal, err := strconv.Atoi(strings.TrimSpace(bounds[1]))
+		if err != nil {
+			return nil, fmt.Errorf("invalid max value: %s", bounds[1])
+		}
+
+		if minVal < 0 || maxVal < 0 {
+			return nil, fmt.Errorf("negative values not allowed: min=%d, max=%d", minVal, maxVal)
+		}
```
shmuelk: This check can be deleted as it will never happen

docs/architecture.md
```diff
@@ -462,6 +462,122 @@ When integrating with a prefix-cache scorer, the prefix-cache scorer should be d
...
+- **Type**: `context-length-aware`
+- **Parameters**:
+  - `label` (optional): Pod label name containing context length range(s). 
+    Default: `llm-d.ai/context-length-range`
+  - `enableFiltering` (optional): If true, the plugin operates as a filter, excluding non-matching pods. 
+    Default: false
+
+**Label Format:**
+
+Pods should be labeled with context length ranges using the format `"min-max"`:
```
```suggestion
Pods should be labeled with context length ranges using the format `"min-max"`, where _min_ and _max_ are both positive integers:
```

docs/architecture.md
```diff
@@ -462,6 +462,122 @@ When integrating with a prefix-cache scorer, the prefix-cache scorer should be d
...
+  - name: default
+    plugins:
+      - pluginRef: context-length-aware
+        weight: 3
+      - pluginRef: load-aware-scorer
+        weight: 1
+      - pluginRef: max-score-picker
+```
+
+**Example - Filter Configuration:**
```
```suggestion
**Example Configuration - Scorer with Filtering Enabled:**
```

docs/architecture.md
```diff
@@ -462,6 +462,122 @@ When integrating with a prefix-cache scorer, the prefix-cache scorer should be d
...
+llm-d.ai/context-length-range: "0-2048"
+```
+
+Multiple ranges can be specified with comma separation:
+
+```yaml
+llm-d.ai/context-length-range: "0-2048,8192-16384"
+```
+
+**Example - Scorer Configuration:**
```
```suggestion
**Example Configuration - Scorer:**
```

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,296 @@
...
+const (
+	// ContextLengthAwareType is the type of the ContextLengthAware plugin
+	ContextLengthAwareType = "context-length-aware"
+
+	// DefaultContextLengthLabel is the default label name used to identify context length ranges on pods
+	DefaultContextLengthLabel = "llm-d.ai/context-length-range"
+
+	// charToTokenMultiplier defines the multiplier to convert characters to tokens
+	// This is an approximate value and may vary based on the tokenizer used
+	charToTokenMultiplier = 0.25
```
elevran, line 29: nit: might make sense to have this factor shared across different places in the code? Assuming this is used in multiple places

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,296 @@
...
+
+	// EnableFiltering determines whether the plugin also filters pods that don't match a
+	// request's context length.
+	// If false, the plugin only scores pods.
+	// Default is false.
+	EnableFiltering bool `json:"enableFiltering"`
+}
+
+// contextRange represents a single context length range.
+type contextRange struct {
```
elevran, line 45: nit: probably fine as `range`, nothing `context` specific about this

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,296 @@
...
+	totalChars := 0
+
+	// Handle chat completions
+	if request.Body.ChatCompletions != nil {
+		for _, msg := range request.Body.ChatCompletions.Messages {
+			totalChars += len(msg.Content.Raw)
+		}
+	}
+
+	// Handle regular completions
```
elevran, line 238: nit: be defensive and use "else if", or is it guaranteed to never have both set?

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	// Convert characters to approximate token count
+	estimatedTokens := int(float64(totalChars) * charToTokenMultiplier)
+	return estimatedTokens
+}
+
+// parseContextRanges parses a label value into context ranges
+// Expected format: "min-max" or "min-max,min-max,..."
+// Examples: "0-2048", "2048-8192", "0-2048,8192-16384"
```
elevran: nit: consider simplifying to a single range only?
Still not clear based on description why multiple ranges are possible. Given that this is driven by "available memory", one could expect labels to be a single range (often "0-max" unless there are other deployments with their own max set so other Pods become "min-max" to allow splitting between different pods).

docs/architecture.md
```diff
@@ -531,6 +531,129 @@ When integrating with a prefix-cache scorer, the prefix-cache scorer should be d
...
+
+**Label Format:**
+
+Pods should be labeled with context length ranges using the format `"min-max"`, where _min_ and _max_ are both positive integers:
+
+```yaml
+llm-d.ai/context-length-range: "0-2048"
+```
+
+Multiple ranges can be specified with comma separation:
```
elevran: Still not clear on the requirement to support multiple ranges on an endpoint.
Unless there's a clear requirement, suggest simplifying the use and code (less parsing, less checking, ...) by having a single range per endpoint.

vMaroon: There is none at the moment, agree that dropping support would simplify code. Done.

docs/architecture.md
```diff
@@ -531,6 +531,116 @@ When integrating with a prefix-cache scorer, the prefix-cache scorer should be d
 
 ---
 
+#### ContextLengthAware
+
+A multi-purpose plugin that can operate as both a **Filter** and a **Scorer** to route inference requests
```
elevran: Still seems to suggest that role of scoring and filtering are on par. I would emphasize that the Scorer is always applied and Filter is optional behavior, gated behind `enableFiltering`.

docs/architecture.md
```diff
@@ -531,6 +531,116 @@ When integrating with a prefix-cache scorer, the prefix-cache scorer should be d
...
+
+**Use Cases:**
+- Route short prompts to pods with smaller GPU memory
+- Direct long-context requests to specialized high-memory pods
+- Optimize performance by matching workload characteristics to hardware capabilities
+- Support heterogeneous deployments with different GPU configurations
+
+The plugin scores all pods based on how well their ranges match the request:
+- **In-range match (0.5–1.0):** Higher scores for tighter/more specific ranges (specialized pods), lower scores for very wide ranges (generalist pods)
+- **Out-of-range fallback (0.0–0.3):** When no range matches, pods are ranked by proximity to the request. For example, a 9000-token request prefers a pod with `max=8192` over one with `max=2048`. Fallback scores never exceed 0.3, so an in-range match always wins.
```
elevran: Q: I wonder if this is reasonable: can a 9000 token request be handled by **anything** that has a `max` < 9000?

```md
Fallback scores never exceed 0.3, so an in-range match always wins.
```
Not always true, unless filtering is enabled.
For wide ranges (> 13K) and a request at the top of the width, this fails (scores < 0.3).
A pod with range [0-15000] routing a 14,999-token request scores 0.280, losing to a nearby out-of-range pod with distance=1 scoring 0.300. This becomes worse for wider ranges (e.g., ~0.16 for 32K range).

You should make sure that the in-range score are (0.3, 1.0] so they always score higher (e.g. `0.3 + 0.7 * (widthScore*0.7 + positionScore*0.3))`.
Noted here but should be fixed in code.

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,316 @@
...
+	// DefaultContextLengthLabel is the default label name used to identify context length ranges on pods
+	DefaultContextLengthLabel = "llm-d.ai/context-length-range"
+
+	// charToTokenMultiplier defines the multiplier to convert characters to tokens
+	// This is an approximate value and may vary based on the tokenizer used
+	// Used as a fallback when no tokenizer is configured
+	charToTokenMultiplier = 0.25
+)
+
+type contextLengthAwareParams struct {
```
elevran: nit:
```suggestion
type contextLengthAwareParameters struct {
```
Be consistent with other filters/scorers.

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,316 @@
...
+
+	logger := log.FromContext(ctx).V(logutil.DEBUG).WithName("ContextLengthAware.Filter")
+	contextLength, usedTokenizer := p.getContextLength(ctx, request)
+	logger.V(logutil.TRACE).Info("Filtering endpoints by context length", "contextLength", contextLength, "usedTokenizer", usedTokenizer)
+
+	var filteredEndpoints []scheduling.Endpoint
+
+	for _, endpoint := range endpoints {
+		metadata := endpoint.GetMetadata()
+		if metadata == nil {
```
elevran, line 126: unlikely that there would ever be an endpoint without metadata (outside of tests).
However, if it did happen, why not treat it the same as a pod without a label (i.e. include it, do not filter out).

pkg/plugins/multi/context_length_aware.go
```diff
@@ -0,0 +1,316 @@
...
+func (p *ContextLengthAware) Filter(ctx context.Context, _ *scheduling.CycleState, request *scheduling.LLMRequest, endpoints []scheduling.Endpoint) []scheduling.Endpoint {
+	if !p.enableFiltering {
+		return endpoints // pass through if not in filter mode
+	}
+
+	logger := log.FromContext(ctx).V(logutil.DEBUG).WithName("ContextLengthAware.Filter")
+	contextLength, usedTokenizer := p.getContextLength(ctx, request)
+	logger.V(logutil.TRACE).Info("Filtering endpoints by context length", "contextLength", contextLength, "usedTokenizer", usedTokenizer)
+
+	var filteredEndpoints []scheduling.Endpoint
```
elevran: nit: functionally the same but empty slice form is used in other filters/scorers
```suggestion
	filteredEndpoints := []scheduling.Endpoint{}
```

elevran (commented):
/lgtm

Minor edits mostly (with the exception of wrong scoring for wide ranges). Please correct scoring (wide ranges **and** out of range?).
Leaving approval to @shmuelk as he has `requested changes` anyway.

pkg/plugins/multi/context_length_aware.go
vMaroon: Should be guaranteed.

docs/architecture.md
```diff
@@ -576,6 +576,119 @@ schedulingProfiles:
...
+
+This plugin reads tokenized prompt data from CycleState, as written by the `tokenizer` plugin.
+When the tokenizer plugin is configured in the same scheduling profile, the context-length-aware plugin
+uses the exact token count for routing decisions. This avoids double-tokenization with other plugins
+that also consume tokens (e.g., `precise-prefix-cache-scorer`).
+
+When the tokenizer scorer is not configured, the plugin falls back to character-based estimation
+(characters × 0.25).
+
+> [!NOTE]
```
elevran, line 618: is this still true? Isn't the `tokenizer` running in `PrepareData`?

docs/architecture.md
```diff
@@ -576,6 +576,119 @@ schedulingProfiles:
...
+
+**Use Cases:**
+- Route short prompts to pods with smaller GPU memory
+- Direct long-context requests to specialized high-memory pods
+- Optimize performance by matching workload characteristics to hardware capabilities
+- Support heterogeneous deployments with different GPU configurations
+
+The plugin scores all pods based on how well their ranges match the request:
+- **In-range match (0.3–1.0]:** Higher scores for tighter/more specific ranges (specialized pods), lower scores for very wide ranges (generalist pods). In-range scores are always strictly above 0.3, guaranteeing they beat any out-of-range fallback.
+- **Out-of-range fallback [0.0–0.3):** When no range matches, pods are ranked by proximity to the request. For example, a 9000-token request prefers a pod with `max=8192` over one with `max=2048`.
```
elevran, line 594: agree that `max=8192` seems like a better choice than `max=2048`, still wondering if a request with context length over max can be served...

elevran:
/lgtm

@shmuelk should approve via UX to remove `changes requested`

pkg/plugins/multi/context_length_aware_test.go
shmuelk: I guess this is OK.

docs/architecture.md
vMaroon: Addressed.

pkg/plugins/multi/context_length_aware.go
vMaroon: Addressed.

pkg/plugins/multi/context_length_aware.go
vMaroon: Addressed.

docs/architecture.md
vMaroon: It works when the range is a preference, which will actually be the Spyre use-case I think.

docs/architecture.md
vMaroon: It is still true, yes. For the release, the tokenizer is a no-op scorer. Since we lost the `TokenizedPrompt` field in `LLMRequest`.

PR #553, opened 2026-01-12
Models extractor

### LINKED ISSUE #466: Enable collection of configured / loaded models in each inference serving endpoint
The ability to collect and use information from `/v1/models` to direct traffic to specific endpoints.

**Why is this needed**:
Models can be loaded at runtime and the set of available models changed at runtime.
To support divergent model availability we would need to collect the set of models on each endpoint and confirm model availability on serving inference.

irar2:
This PR adds an ability to collect information from /v1/models and store it in endpoint's attributes.

Closes #466

### DIFF
--- a/pkg/plugins/datalayer/models/factories.go
+++ b/pkg/plugins/datalayer/models/factories.go
@@ -35,6 +35,9 @@ func ModelDataSourceFactory(name string, parameters json.RawMessage, _ plugin.Ha
 			return nil, err
 		}
 	}
+	if cfg.Scheme != "http" && cfg.Scheme != "https" {
+		return nil, fmt.Errorf("unsupported scheme: %s", cfg.Scheme)
+	}
 
 	ds := http.NewHTTPDataSource(cfg.Scheme, cfg.Path, cfg.InsecureSkipVerify, ModelsDataSourceType,
 		name, parseModels, ModelsResponseType)


pkg/plugins/datalayer/models/extractor.go
```diff
@@ -0,0 +1,96 @@
...
+
+const modelsAttributeKey = "/v1/models"
+
+// ModelInfoCollection defines models' data returned from /v1/models API
+type ModelInfoCollection []ModelInfo
+
+// ModelInfo defines model's data returned from /v1/models API
+type ModelInfo struct {
+	ID     string `json:"id"`
+	Parent string `json:"parent,omitempty"`
```
nirrozenbaum, line 21: parent field is not part of OpenAI standardization. 
it's specific to vllm and might not work with other model servers.
I also don't think it's used (or should be used) anywhere.
I recommend removing this field.

OpenAI standard here:
https://platform.openai.com/docs/api-reference/models/list

elevran: A few comments
- If not present, the `omitempty` kicks in so I don't see the downside of having it.
- For use cases that need the parent information for Base/LoRA relations, if it is not provided by model extraction then one must assume the base model name is provided elsewhere. There is currently no other source of truth...

I think it is fine to rely on vLLM specific for that. 
1. It can be treated as part of the "contract" (same as the case when other model servers are expected to provide the MSP metrics even if by a different name).
1. configuration of data sources is per EPP so you can always **not** enable this for other model servers . This is valid usage as long as we use homogeneous model server in a pool (other code breaks as well when this is not the case...)

pkg/plugins/datalayer/models/datasource_test.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func TestDatasource(t *testing.T) {
+	source := http.NewHTTPDataSource("https", "/models", true, ModelsDataSourceType,
+		"models-data-source", parseModels, ModelsResponseType)
+	extractor, err := NewModelExtractor()
+	assert.Nil(t, err, "failed to create extractor")
+
+	dsType := source.TypedName().Type
+	assert.Equal(t, ModelsDataSourceType, dsType)
```
elevran: nit: not sure there's value in testing this

pkg/plugins/datalayer/models/datasource_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	extractors := source.Extractors()
+	assert.Len(t, extractors, 1)
+	assert.Equal(t, extractor.TypedName().String(), extractors[0])
+
+	err = datalayer.RegisterSource(source)
+	assert.Nil(t, err, "failed to register")
+
+	ctx := context.Background()
+	factory := datalayer.NewEndpointFactory([]fwkdl.DataSource{source}, 100*time.Millisecond)
```
elevran: nit: since calling `Collect()` directly in L50, the timeout passed can be "infinite" to indicate it's not used

pkg/plugins/datalayer/models/extractor.go
```diff
@@ -0,0 +1,96 @@
...
+		return nil
+	}
+	clone := make([]ModelInfo, len(m))
+	copy(clone, m)
+	return (*ModelInfoCollection)(&clone)
+}
+
+func (m ModelInfoCollection) String() string {
+	if m == nil {
+		return ""
```
elevran: nit: `[]` is moreidiomatic for a collection than empty string

pkg/plugins/datalayer/models/extractor.go
```diff
@@ -0,0 +1,96 @@
...
+	ModelsResponseType = reflect.TypeOf(ModelResponse{})
+)
+
+// ModelExtractor implements the models extraction.
+type ModelExtractor struct {
+	typedName fwkplugin.TypedName
+}
+
+// NewModelExtractor returns a new model extractor.
+func NewModelExtractor() (*ModelExtractor, error) {
```
elevran, line 67: nit: at least in theory, the plugin could have a name...

pkg/plugins/datalayer/models/extractor_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	if exType := extractor.TypedName().Type; exType == "" {
+		t.Error("empty extractor type")
+	}
+
+	if exName := extractor.TypedName().Name; exName == "" {
+		t.Error("empty extractor name")
+	}
+
+	if inputType := extractor.ExpectedInputType(); inputType != ModelsResponseType {
```
elevran, line 28: nit: unless tested elsewher (eg, in model data source), may want to validate that extractor can be added to the source (which validates compatible source output and extractor input types)

pkg/plugins/datalayer/models/factories.go
```diff
@@ -0,0 +1,67 @@
...
+// models data source plugins specified in a configuration.
+func ModelDataSourceFactory(name string, parameters json.RawMessage, _ plugin.Handle) (plugin.Plugin, error) {
+	cfg := defaultDataSourceConfigParams()
+	if parameters != nil { // overlay the defaults with configured values
+		if err := json.Unmarshal(parameters, cfg); err != nil {
+			return nil, err
+		}
+	}
+
+	ds := http.NewHTTPDataSource(cfg.Scheme, cfg.Path, cfg.InsecureSkipVerify, ModelsDataSourceType,
```
elevran, line 42: Q; does `NewHTTPDataSource` validate the scheme?

pkg/plugins/datalayer/models/factories.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// ModelServerExtractorFactory is a factory function used to instantiate data layer's models
+// Extractor plugins specified in a configuration.
+func ModelServerExtractorFactory(name string, _ json.RawMessage, _ plugin.Handle) (plugin.Plugin, error) {
+	extractor, err := NewModelExtractor()
+	if err != nil {
+		return nil, err
+	}
+	extractor.typedName.Name = name
```
elevran: nit: IIRC most plugins implement a `WithName(name string)`. if so, this should be consistent

pkg/plugins/datalayer/models/datasource_test.go
irar2: removed

pkg/plugins/datalayer/models/datasource_test.go
irar2: Changed to 100 hours

pkg/plugins/datalayer/models/extractor.go
irar2: fixed

pkg/plugins/datalayer/models/extractor.go
irar2: What do you mean?

pkg/plugins/datalayer/models/extractor_test.go
irar2: Tested in datasource_test.go

pkg/plugins/datalayer/models/factories.go
irar2: No, there is only a check if it's https

pkg/plugins/datalayer/models/factories.go
irar2: Added

pkg/plugins/datalayer/models/extractor.go
elevran: ModelExtractor is a plugin. A plugin has a type and an optional name.
The code does not support setting a plugin name and it should.

pkg/plugins/datalayer/models/factories.go
elevran: Since we use the scheme passed in by the user it should at least sanitize it to ensure it's one one of a known set of acceptable values (e.g., "http" and "https").
Can be in this PR or separate adding scheme validation to the HTTPDataSource

pkg/plugins/datalayer/models/extractor.go
irar2: There is the WithName() method now

pkg/plugins/datalayer/models/factories.go
irar2: Added

pkg/plugins/datalayer/models/extractor.go
elevran: thanks.
I was also thinking `NewModelExtractor()` should be extended with a `name string` parameter. If empty it is set to the type and `WithName()` is called internally.
I think that that would have been more consistent with other plugins.

pkg/plugins/datalayer/models/factories.go
elevran: thanks.
Please open a tracking issue to move this check into HTTPDataSource in GAIE. It should not be up to each data source, IMO.

PR #555, opened 2026-01-12
chore(test): update API version for nixl test

zdtsw:
# Changes
- Update InferencePool  from v1alpha2 to v1
- Remove InfereceModel
- extentionRef was in old v1alpha2, in v1 it should be updated to endpointPickerRef
- Update docs


test/sidecar/config/nixl/inferencemodel.yaml
```diff
@@ -1,10 +1,10 @@
-apiVersion: inference.networking.x-k8s.io/v1alpha2
+apiVersion: inference.networking.k8s.io/v1
```
shmuelk: There is no longer an InferenceModel CRD

zdtsw: Thanks, I've remove InferenceModel in the change

PR #564, opened 2026-01-14
LatencyPredictionScorer for PD

RishabhSaini:
Blocked on GAIE
1) https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2145


RishabhSaini:
For PD:
| Metric | WITHOUT Latency Predictor | WITH Latency Predictor | Difference |
  |--------|------------------|----------------|-------------|
  | **Success Rate** | 97.1% (4854/4999) | 98.4% (4919/5000) | +1.3%  |
  | **TTFT Mean** | 2105 ms | 911 ms | **-56.7%**  |
  | **TTFT P50** | 1397 ms | 836 ms | **-40.2%**  |
  | **TTFT P99** | 9687 ms | 3608 ms | **-62.8%** |
  | **TPOT Mean** | 22.4 ms | 13.0 ms | **-41.9%**  |
  | **TPOT P50** | 17.4 ms | 12.3 ms | **-28.9%**  |
  | **TPOT P99** | 74.8 ms | 27.0 ms | **-63.9%**  |
  | **Throughput** | 7862 tok/s | 7462 tok/s | -5.1% |

  ### Summary

  The SLO-aware latency prediction scorer delivers **significant latency improvements** across all percentiles:
  -  **~57% faster** mean TTFT
  -  **~42% faster** mean TPOT
  -  **~63% reduction** in P99 tail latencies
  -  **Higher success rate** (98.4% vs 97.1%)

  The slight throughput reduction (-5%) is expected as the predictor prioritizes **load-balanced, low-latency pods** over maximizing absolute throughput, resulting in better QoS and reliability.

ahg-g:
Can we summarize the approach we are taking here? is it using ttft prediction to pick the prefill endpoint and the itl to pick the decode endpoint? 

Also, can we have the scorers in IGW instead of here?

pkg/sidecar/proxy/timing_writer.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"fmt"
+	"net/http"
+)
+
+const (
+	// responseHeaderPrefillTTFTMs reports the actual prefill TTFT in milliseconds to EPP
+	// for training data collection. EPP's SLOAwareRouter extracts this header in the
+	// ResponseReceived hook and records training data for the prefill pod.
+	responseHeaderPrefillTTFTMs = "x-prefill-ttft-ms"
```
kaushikmitr: i am little confused why this is necessary. Cant we just reuse the requestcontroller hooks to measure ttft and tpot just as we are doing in epp. We already should have the information as to which pods (prefill or decode) the TTFT/TPOT values are coming from?

RishabhSaini: The EPP from its hooks only sees 
PreRequest(decodeEndpoint): sending request to the decode pod
ResponseReceived(decodeEndpoint): receiving from the decode pod

EPP never directly communicates with the prefill pod. That happens inside the routing-proxy sidecar for the P/D llm-d deployment.

This diagram explains it well: https://github.com/llm-d/llm-d-inference-scheduler/blob/main/docs/disagg_pd.md#architectural-details

With this header, the routing-proxy measures prefill latency internally and injects into response: "x-prefill-ttft-ms". EPP extracts this and records training data for the prefill endpoint

kaushikmitr: i see, so the requestcontrol plugins in GAIE cannot handle dissagg scenario? 
https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/pkg/epp/framework/interface/requestcontrol/plugins.go
Actually the requestcontrol hooks only needs to grab the metrics from the right endpoint: https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/pkg/epp/framework/plugins/scheduling/scorer/predictedlatency/requestcontrol_hooks.go

Only caveat is we can only grab TTFT as returned by decode pod whch includes network hop, but i think that i sthe true TTFT and we should include that in out training.

RishabhSaini: EPP can query the prefill pod's aggregate metrics (via Prometheus/vLLM metrics endpoint). But since EPP does not track individual Prefill Pod requests, per request timing is not possible with the current P/D architecture. Only the routing-proxy sidecar has access to the local per-request measurements.
The reason we need to have a routing sidecar on the decode pod is to coordinate the KV Cache transfer between the Decode->Prefill->Decode for a request.

RishabhSaini: @kaushikmitr Ok as discussed in today's sync, I explored the `Experimental_DefaultPrefillProfile` and was able to get rid of handling the addRequestToQueue and removeRequestFromQueue for the Prefill endpoint type. So extneding the request control hooks PreRequest and ResponseComplete are no longer needed in the llm-d-inferende-scheduler. These are now handled within the GAIE.
However, you would still need to return `x-prefill-ttft-ms` so it can be read by the ResponseReceived hook. This will be included in the training entry for the Prefill profile

PR #567, opened 2026-01-15
Add P/D coordinator Prometheus metrics

sallyom:
Add Prometheus histogram metrics to track P/D disaggregation performance from the coordinator's perspective. These metrics capture the true end-to-end latency that vLLM instances cannot see individually.


JeffLuoo:
cc: @liu-cong as you mentioned to me about metrics from the sidecar. Can you help take a look of this draft PR?

PR #569, opened 2026-01-18
fix(scorer): make active request pd aware

kyanokashi:


pkg/plugins/scorer/active_request.go
```diff
@@ -166,47 +169,62 @@ func (s *ActiveRequest) Score(ctx context.Context, _ *types.CycleState, _ *types
...
 	debugLogger := log.FromContext(ctx).V(logutil.DEBUG)
 
-	for _, profileResult := range schedulingResult.ProfileResults { // schedulingResult guaranteed not to be nil
-		if profileResult == nil || profileResult.TargetPods == nil || len(profileResult.TargetPods) == 0 {
-			continue
-		}
+	targetProfiles := []string{schedulingResult.PrimaryProfileName}
+	if schedulingResult.PrimaryProfileName != "prefill" {
+		targetProfiles = append(targetProfiles, "prefill")
+	}
```
kyanokashi: AFAIU if there's a secondary "prefill" profile, that means we are in P/D

kfswain: I think this is too universal of a conditional. There may be cases where this scorer is used, with no relation to P/D, and the primary profile name will not be P/D.

I think we can instead use: https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/c68bbe56ee6684060d3b2184d85326f75aa64096/pkg/epp/scheduling/types/types.go#L232

and make the conditional `if schedulingResult.ProfileResults["prefill"] = ok, _; ok {...`

kyanokashi: I mean, now that I'm thinking about it, because we now create 1 request entry per request id, that is enough to sort the issue. We can just iterate through all of the profile results as before.

kfswain (approved):

pkg/plugins/scorer/active_request.go
```diff
@@ -166,47 +169,61 @@ func (s *ActiveRequest) Score(ctx context.Context, _ *types.CycleState, _ *types
...
+	// add to request cache
+	s.requestCache.Set(request.RequestId, &requestEntry{PodNames: podNames, RequestID: request.RequestId}, 0) // Use default TTL
 }
 
 // ResponseComplete is called after a response is sent to the client.
 // It removes the specific request entry from the cache and decrements
 // the pod count.
-func (s *ActiveRequest) ResponseComplete(ctx context.Context, request *types.LLMRequest,
-	_ *requestcontrol.Response, targetPod *backend.Pod) {
+func (s *ActiveRequest) ResponseComplete(
```
kyanokashi, line 203: @vMaroon just to make sure, this is called once per request correct? This PR makes this assumption

nirrozenbaum: yes 👍🏻

Issue #552, opened 2026-01-11
Update llm-d-kv-cache dependency to use the new disaggregated tokenizer service.


delavet:
`llm-d-kv-cache` has just been renamed and a new disaggregated tokenizer service is proposed for better tokenization.

This issue is responsible for tracking the integration of these changes into llm-d-inference-scheduler in the next release.

cc @vMaroon

PR #570, opened 2026-01-19
refactor: kv cache manager repo

sagearc:
This PR includes updates from llm-d-kv-cache v0.5.
The main changes are:
- lora support:
  - https://github.com/llm-d/llm-d-kv-cache/pull/220
  - https://github.com/llm-d/llm-d-kv-cache/pull/237
  - https://github.com/llm-d/llm-d-kv-cache/pull/243
- vllm wrapper for chat template rendering & tokenization:
  - https://github.com/llm-d/llm-d-kv-cache/pull/234
- active-active high availability (HA):
  - https://github.com/llm-d/llm-d-kv-cache/pull/212
- single tokenizer initialization at startup
  - https://github.com/llm-d/llm-d-kv-cache/pull/192

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -308,12 +308,14 @@ Configuration:
 
 - **Type**: `precise-prefix-cache-scorer`
 - **Parameters**:
+  - `tokenProcessorConfig`: Configuration for the `kvblock.TokenProcessor`.
   - `indexerConfig`: Configuration for the `kvcache.Indexer`.
   - `kvEventsConfig`: Configuration for the `kvevents.Pool`.
 
 See list of parameters at [llm-d-kv-cache/docs/configuration.md](https://github.com/llm-d/llm-d-kv-cache/blob/fa85b60207ba0a09daf23071e10ccb62d7977b40/docs/configuration.md).
 
 Note that in most cases you will only need to set:
+- Model name in the `tokenizersPoolConfig` to match the model used in the vLLM deployment.
 - HuggingFace token for the `tokenizersPoolConfig` or the `tokenizersCacheDir` to a mounted directory containing the tokenizers.
   - For the HuggingFace token, the inference-scheduler also accepts the environment variable `HF_TOKEN` - this is the practical option for security. 
 - **IMPORTANT**: Token processor's block-size and hash-seed to match those used in the vLLM deployment.
@@ -325,16 +327,16 @@ Example configuration with the above parameters set:
 plugins:
   - type: precise-prefix-cache-scorer
     parameters:
+      tokenProcessorConfig:
+        blockSize: 64                    # must match vLLM block size
+        hashSeed: "12345"                # must match vLLM PYTHONHASHSEED env var
       indexerConfig:
-        tokenProcessorConfig:
-          blockSize: 64
-          hashSeed: "12345"
-      tokenizersPoolConfig:
-        modelName: hf-repo/model-name
-        hf:
-          huggingFaceToken: your_hf_token_here    # automatically set by `HF_TOKEN` environment variable
-      kvBlockIndexConfig:
-        enableMetrics: true
+        kvBlockIndexConfig:
+          enableMetrics: true    
+        tokenizersPoolConfig:
+          modelName: hf-repo/model-name
+          hf:
+            huggingFaceToken: your_hf_token_here    # automatically set by `HF_TOKEN` environment variable
 ```
 
 Example configuration with all parameters set:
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -138,3 +138,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.1 // indirect
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
+
+replace github.com/llm-d/llm-d-kv-cache => github.com/sagearc/llm-d-kv-cache-manager v0.0.0-20260105163019-e73ed03ca6ff
```
nirrozenbaum: this should be changed 😃

sagearc: This was temporary since the PR wasn't merged yet. Now that it is merged, reverted to kv cache manager upstream instead of my fork :)

scripts/fetch-python-wrapper.sh
```diff
@@ -16,32 +16,32 @@ set -euo pipefail
 VERSION="${1:-${KVCACHE_MANAGER_VERSION:-v0.3.2}}"
 OUTPUT_DIR="${2:-llm-d-kv-cache-manager/pkg/preprocessing/chat_completions}"
```
hyeongyun0916: It looks like we should also remove `-manager` here.

scripts/fetch-python-wrapper.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 # fetch-python-wrapper.sh
-# Fetches the Python wrapper file (render_jinja_template_wrapper.py) from llm-d-kv-cache-manager
+# Fetches the Python wrapper file (render_jinja_template_wrapper.py) from llm-d-kv-cache
```
hyeongyun0916: What is `fetch-python-wrapper.sh` used for? If it's not being used, should we remove it? I think it would be better to use install-python-deps for local setups instead.

vMaroon: I think it's not in use anymore

hyeongyun0916:
I think it need test. but skipped because this pr also changed docs
https://github.com/llm-d/llm-d-inference-scheduler/actions/runs/21184070368/workflow?pr=570#L28

scripts/fetch-python-wrapper.sh
hyeongyun0916: [delete fetch-python-wrapper.sh](https://github.com/llm-d/llm-d-inference-scheduler/pull/570/commits/a50eb9d6b98896b9d99994e68e7a5c02a878790c)

elevran:
Please split this into diffrerent PRs to allow early merge (e.g., LoRA and vLLM are independent)

.dockerignore
```diff
@@ -0,0 +1,21 @@
...
+*.DS_Store
+
+# Local virtual environments
+venv
+
+# Python cache files
+__pycache__
+
+# Docker files
+Dockerfile
```
elevran, line 21: nit: newline missing

.github/workflows/ci-pr-checks.yaml
```diff
@@ -12,22 +12,27 @@ jobs:
...
+            src:
+              - '**/*.go'
+              - '**/*.py'
   lint-and-test:
     needs: check-changes
-    if: ${{ needs.check-changes.outputs.docs == 'false' }}
+    if: ${{ needs.check-changes.outputs.src == 'true' }}
     runs-on: ubuntu-latest
     steps:
+      - name: Free Disk Space (Ubuntu)
```
elevran, line 31: Q: can you briefly explain why this is needed and was not needed before?

.github/workflows/ci-pr-checks.yaml
```diff
@@ -57,13 +59,16 @@ jobs:
       - name: Set PKG_CONFIG_PATH
         run: echo "PKG_CONFIG_PATH=/usr/lib/pkgconfig" >> $GITHUB_ENV
 
-      - name: go mod tidy
-        run: go mod tidy
+      - name: Install dependencies
+        run: |
+          go mod tidy
+          sudo -E env "PATH=$PATH" make install-dependencies
```
elevran, line 65: Q: any limitation in running `sudo -E env "PATH=$PATH" make install-dependencies install-python-deps` in one statement?

.github/workflows/ci-pr-checks.yaml
vMaroon: Because now the disk use for dependencies is larger - and it now runs out of space. This was discussed with Greg at some point, we can increase capacity by upgrading tier, but this approach was preferred.

It is temporary in all cases.

elevran: since this is the first step, this clears up disk space of packages that are installed (and hopefully not used) before anything from inference-scheduler is actually run?
It is run in the hope that afterwards there will be sufficient disk space on the worker to run the build.

elevran:
/lgtm
/approve

under the assumption that #552 is top priority and shall be completed ASAP after v0.5 is cut.
The two remaining items in this PR can be done in a quick follow up to allow cutting 0.5 RC now

PR #578, opened 2026-01-21
Enable prefix-cache awareness in active-active multi-replica scheduler deployments

vMaroon:
### active-active multi-replica scheduler support https://github.com/llm-d/llm-d-kv-cache/pull/212

> ## Summary
> Implemented a pod reconciler controller that manages per-pod ZMQ subscribers for KVEvents processing, and the required logic. Also moved the `kvevents` to the same level of `kvcache` library, as should have been.
> 
> Components:
> - **PodReconciler**: Kubernetes controller that watches pods and manages subscribers
>   - This is one standalone approach, actual integration with the scheduler depends on https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2017
> - **SubscriberManager**: Thread-safe manager for multiple ZMQ subscribers
> - Updated `kvevents.Pool` - Now works with SubscriberManager instead of a global socket
> 
> Added integration + unit tests for all new functionality, updated documentation + examples.

The current integration manages pod discovery through looking at the available pods on every `Score` call. A pod that does not appear there in any request for 10 minutes is assumed dead.

A proper integration through the data-layer will be implemented once ready in IGW. Tracker: https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2017

### DIFF
--- a/pkg/plugins/scorer/precise_prefix_cache.go
+++ b/pkg/plugins/scorer/precise_prefix_cache.go
@@ -118,7 +118,7 @@ func New(ctx context.Context, config PrecisePrefixCachePluginConfig) (*PrecisePr
 	var subscribersCache *ttlcache.Cache[string, struct{}]
 
 	// initialize the subscribers cache only if pod discovery is enabled
-	if config.KVEventsConfig.DiscoverPods == true {
+	if config.KVEventsConfig.DiscoverPods {
 		// initialize the subscribers TTL cache
 		subscriptionTimeout := 10 * time.Minute
 		subscribersCache = ttlcache.New[string, struct{}](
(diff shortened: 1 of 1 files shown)


docs/architecture.md
```diff
@@ -339,6 +339,31 @@ plugins:
...
+            tokenizersCacheDir: "/tmp/tokenizers"
+      kvEventsConfig:
+        topicFilter: "kv@"
+        concurrency: 4 
+        discoverPods: true      # enables automatic pod discovery for active-active HA
+        podDiscoveryConfig:
+          socketPort: 5556
+```
+
+Where the vLLM engines are configured to emite KV-Events on port `5556` as follows:
```
```suggestion
Where the vLLM engines are configured to emit KV-Events on port `5556` as follows:
```

elevran (approved):

PR #531, opened 2026-01-05
Extend support for different ways to decide if disaggregated PD is required

mayabar:
Update PD profile handler to allow easier support for different ways to decide whether disaggregated PD is required.
Use prefix data stored in prepare data stage.
Currently supports prefix plugin, next step is to add support for precise prefix scorer plugin (#530)

Fixes #519

This PR should be merges after [PR](https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2053) in GIE which stores in prefix scorer data number of tokens instead of number of blocks.

The GIE PR contains change in prefix plugin configuration structure, waiting with this PR until it will be merged.

PR #581, opened 2026-01-21
feat: use Tinyllama as the "model" for kind test and switch to use precise-prefix-cache-score in config

zdtsw:
# Description
- in order to test precies-prefix-cache-score we cannot use `fool-review` since it need call kv-cache-manager to get tokenizer by getting a real model from HF
- the change is to switch the "default model" to TinyLlama
- also to make tokenizer folder writable need change permission to the USER in Dockerfile
- rename dp-epp-config.yaml sim-dp-epp-config.yaml as it is used for local test
- change to use `precise-prefix-cache-score `than previous `prefix-cache-score`

# Test
 test sim-epp-config.yaml
`make env-dev-kind`

test sim-dp-epp-config.yaml (data parallel, multiple replicas)
`VLLM_DATA_PARALLEL_SIZE=2 make env-dev-kind`

test sim-pd-epp-config.yaml (prefill/decode disaggregation)
`PD_ENABLED=true make env-dev-kind`

test sim-epp-kvcache-config.yaml (KV cache tracking enabled)
`KV_CACHE_ENABLED=true make env-dev-kind`

test sim-epp-no-hit-lru.yaml (custom config with no-hit-lru-scorer)
`EPP_CONFIG=deploy/config/sim-epp-no-hit-lru.yaml make env-dev-kind`
```
kubectl --context kind-llm-d-inference-scheduler-dev get configmap epp-config -o yaml
apiVersion: v1
data:
  epp-config.yaml: |
    # Sample EPP configuration for running without P/D
    # with small hash block size for simulation purposes
    kind: EndpointPickerConfig
    plugins:
    - type: precise-prefix-cache-scorer
      parameters:
        indexerConfig:
          tokenProcessorConfig:
            blockSize: 5
          kvBlockIndexConfig:
            maxPrefixBlocksToMatch: 256
    - type: no-hit-lru-scorer
      parameters:
        lruSize: 2048
    - type: decode-filter
    - type: max-score-picker
    schedulingProfiles:
    - name: default
      plugins:
      - pluginRef: decode-filter
      - pluginRef: max-score-picker
      - pluginRef: precise-prefix-cache-scorer
        weight: 2
      - pluginRef: no-hit-lru-scorer
        weight: 1
kind: ConfigMap
...
> curl -s -w '\n' http://localhost:30080/v1/completions -H 'Content-Type: application/json' -d '{"model":"TinyLlama/TinyLlama-1.1B-Chat-v1.0","prompt":"hi","max_tokens":10,"temperature":0}' | jq
{
  "choices": [
    {
      "finish_reason": "stop",
      "index": 0,
      "text": "The temperature here is twenty-five degrees centigrade"
    }
  ],
  "created": 1769006945,
  "do_remote_decode": false,
  "do_remote_prefill": false,
  "id": "chatcmpl-863806e1-e462-4945-bd4e-a0e035e70a3d",
  "model": "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
  "object": "text_completion",
  "remote_block_ids": null,
  "remote_engine_id": "",
  "remote_host": "",
  "remote_port": 0,
  "usage": {
    "completion_tokens": 9,
    "prompt_tokens": 1,
    "total_tokens": 10
  }
}
```


deploy/config/epp-config.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: inference.networking.x-k8s.io/v1alpha1
 kind: EndpointPickerConfig
 plugins:
-- type: prefix-cache-scorer
+- type: precise-prefix-cache-scorer
```
shmuelk: I would prefer that this was a simpler deployment without the precise-prefix-cache-scorer

Also this will not work unless the KV_CACHE_ENABLED environment variable is set to true.

Plewase revert this 
```suggestion
- type: prefix-cache-scorer
```

deploy/config/epp-config.yaml
```diff
@@ -12,5 +12,5 @@ schedulingProfiles:
   plugins:
   - pluginRef: decode-filter
   - pluginRef: max-score-picker
-  - pluginRef: prefix-cache-scorer
+  - pluginRef: precise-prefix-cache-scorer
```
shmuelk: Here as well:

```suggestion
  - pluginRef: prefix-cache-scorer
```

deploy/config/sim-epp-config.yaml
```diff
@@ -3,11 +3,13 @@
 apiVersion: inference.networking.x-k8s.io/v1alpha1
 kind: EndpointPickerConfig
 plugins:
-- type: prefix-cache-scorer
+- type: precise-prefix-cache-scorer
```
shmuelk: Again the same is true here:

```suggestion
- type: prefix-cache-scorer
```

deploy/config/sim-epp-config.yaml
```diff
@@ -3,11 +3,13 @@
...
   parameters:
-    hashBlockSize: 5
-    maxPrefixBlocksToMatch: 256
-    lruCapacityPerServer: 31250
+    indexerConfig:
+      tokenProcessorConfig:
+        blockSize: 5
+      kvBlockIndexConfig:
+        maxPrefixBlocksToMatch: 256
```
shmuelk: Please revert this as well

deploy/config/sim-epp-config.yaml
```diff
@@ -16,5 +18,5 @@ schedulingProfiles:
   plugins:
   - pluginRef: decode-filter
   - pluginRef: max-score-picker
-  - pluginRef: prefix-cache-scorer
+  - pluginRef: precise-prefix-cache-scorer
```
shmuelk: Here too:

```suggestion
  - pluginRef: prefix-cache-scorer
```

scripts/kind-dev-env.sh
```diff
@@ -89,7 +89,7 @@ else
         PRIMARY_PORT="8000"
       fi
     else
-      DEFAULT_EPP_CONFIG="deploy/config/dp-epp-config.yaml"
+      DEFAULT_EPP_CONFIG="deploy/config/sim-dp-epp-config.yaml"
```
shmuelk: When running under kind there is no real need to use the dp-epp-config.yaml configuration, as Istio 1.28.1 is used.

In general this profile is only useful for deployments with Istio prior to Istio 1.28.1

deploy/config/sim-pd-epp-config.yaml
```diff
@@ -4,11 +4,13 @@ apiVersion: inference.networking.x-k8s.io/v1alpha1
 kind: EndpointPickerConfig
 plugins:
 - type: prefill-header-handler
-- type: prefix-cache-scorer
+- type: precise-prefix-cache-scorer
```
shmuelk: PD doesn't work with the precise-prefix-cache-scorer, PR #531 comes to fix that issue,

As a result the kind-dev-env.sh script doesn't allow the use of KV_CACHE_ENABLED=true with PD_ENABLED=true

```suggestion
- type: prefix-cache-scorer
```

deploy/config/sim-pd-epp-config.yaml
```diff
@@ -22,11 +24,11 @@ schedulingProfiles:
   plugins:
   - pluginRef: prefill-filter
   - pluginRef: max-score-picker
-  - pluginRef: prefix-cache-scorer
+  - pluginRef: precise-prefix-cache-scorer
```
shmuelk: Again revert here:

```suggestion
  - pluginRef: prefix-cache-scorer
```

deploy/config/sim-pd-epp-config.yaml
```diff
@@ -22,11 +24,11 @@ schedulingProfiles:
...
     weight: 2
 - name: decode
   plugins:
   - pluginRef: decode-filter
   - pluginRef: max-score-picker
-  - pluginRef: prefix-cache-scorer
+  - pluginRef: precise-prefix-cache-scorer
```
shmuelk: Again revert here:

```suggestion
  - pluginRef: prefix-cache-scorer
```

deploy/config/sim-pd-epp-config.yaml
```diff
@@ -4,11 +4,13 @@ apiVersion: inference.networking.x-k8s.io/v1alpha1
...
+- type: precise-prefix-cache-scorer
   parameters:
-    hashBlockSize: 5
-    maxPrefixBlocksToMatch: 256
-    lruCapacityPerServer: 31250
+    indexerConfig:
+      tokenProcessorConfig:
+        blockSize: 5
+      kvBlockIndexConfig:
+        maxPrefixBlocksToMatch: 256
```
shmuelk: Again revert here:

```suggestion
    hashBlockSize: 5
    maxPrefixBlocksToMatch: 256
    lruCapacityPerServer: 31250
```

scripts/kind-dev-env.sh
zdtsw: should I keep the old name `dp-epp-config.yaml` or just skip set  `DEFAULT_EPP_CONFIG` in  `kind-dev-env.sh` at all ?

shmuelk: I'd leave the old name for now.

I'll try and rework the code there to no longer use the config altogether in subsequent PR

PR #582, opened 2026-01-21
Switch to pre-built vLLM wheels for CPU builds

### LINKED ISSUE #579: Local builds on Ubuntu broken following #550
**What happened**:
Errors in running make locally on Ubuntu:
- failed to copy `build/kv_cache_setup.sh` (previous make copied it over without `w` permission
- installation of tools in `build/kv_cache_setup.sh` (i.e., `git gcc-12 g++-12 libnuma-dev`) and running `update-alternatives` require root.

Clearer error messages, do not attempt to execute actions already done (e.g., existing valid copy of file, alternatives set correctly, asking for permission to install or prompting to manually install.

Run on local Ubuntu install without root privileges (CICD actions run with root so not an issue. Also not an issue on Mac).

cc: @vMaroon @sagearc this is a release blocker

sagearc:
Description as in https://github.com/llm-d/llm-d-kv-cache/pull/268

closes #579


Dockerfile.epp
```diff
@@ -18,17 +18,27 @@ RUN go mod download
...
 COPY --from=go-deps /workspace/kv-cache /workspace/kv-cache
 WORKDIR /workspace/kv-cache
-# llm-d-kv-cache's Makefile. not llm-d-inference-scheduler's
-RUN KV_CACHE_PKG=/workspace/kv-cache make install-python-deps
+
+# Create venv and install vLLM based on architecture using pre-built wheels
+RUN python3.12 -m venv /workspace/kv-cache/build/venv && \
+    . /workspace/kv-cache/build/venv/bin/activate && \
+    pip install --upgrade pip && \
+    VLLM_VERSION="0.14.0" && \
```
elevran, line 34: nit: might be worthwhile to make this an input arg so it can be set and updated consistently in different places (e.g., if we need to be consistent in tools makefile, want to experiment with newer versions, etc.)

Makefile.tools.mk
```diff
@@ -194,14 +195,30 @@ install-python-deps: setup-venv ## installs dependencies.
...
 	@KV_CACHE_PKG=$${KV_CACHE_PKG:-$$(go list -m -f '{{.Dir}}' github.com/llm-d/llm-d-kv-cache 2>/dev/null)}; \
-	if [ -n "$$KV_CACHE_PKG" ] && [ -f "$$KV_CACHE_PKG/pkg/preprocessing/chat_completions/setup.sh" ]; then \
-		echo "Running kv-cache setup script..."; \
-		cp "$$KV_CACHE_PKG/pkg/preprocessing/chat_completions/setup.sh" build/kv-cache-setup.sh; \
-		chmod +x build/kv-cache-setup.sh; \
-		cd build && PATH=$(VENV_BIN):$$PATH ./kv-cache-setup.sh && cd ..; \
-	else \
-		echo "ERROR: kv-cache package not found or setup script missing."; \
+	if [ -z "$$KV_CACHE_PKG" ]; then \
+		echo "ERROR: kv-cache package not found."; \
```
elevran, line 199: Q: how is one supposed to make it available? Is it automated or manual intervention needed? Don't let the user guess unless it is trivial.

elevran:
/hold
/lgtm
/approve

Placing hold in case you want to address comments. Otherwise you can cancel the hold.

Dockerfile.epp
vMaroon: The next time this field is updated should be for its removal along with all the python packages. It does exist here and in the setup script for Mac.

Makefile.tools.mk
vMaroon: `go mod download` or `go mod vendor` does it. A followup PR?

elevran: `go mod` with download/tidy will be run (automatically in make or by normal dev process) so I think that's fine.
/hold cancel

PR #600, opened 2026-02-05
add cookie-based affinity support

### LINKED ISSUE #28: [bug] SessionAffinity should use a session cookie
Currently the code uses an arbitrary `x-session-id` header which the HTTP client is not required to relay back in subsequent requests.
It should be using the standard [`Set-Cookie` mechanism](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie)

roytman:
Migrate from proprietary header-based affinity to standard cookie-based affinity to enable native browser support for session persistence.

/fix #28


DEVELOPMENT.md
```diff
@@ -288,6 +288,16 @@ export VLLM_TENSOR_PARALLEL_SIZE=2
 export VLLM_GPU_COUNT_PER_INSTANCE=2
 ```
 
+**4. Enabling Session Affinity:**
```
elevran: I don't think we're recommending any other scorers or filters in the doc so should not be emphasizing this either.
It should be enabled via configuration file, as any other plugin.

README.md
```diff
@@ -14,7 +14,15 @@ This provides an "Endpoint Picker (EPP)" component to the llm-d inference
 framework which schedules incoming inference requests to the platform via a
 [Kubernetes] Gateway according to scheduler plugins. For more details on the
 llm-d inference scheduler architecture, routing logic, and different plugins
-(filters and scorers), including plugin configuration, see the [Architecture Documentation]).
+(filters and scorers), including plugin configuration, see the [Architecture Documentation].
+
```
elevran: Q: what is the rationale for adding the below section?
It seems unrelated to this PR and there are likeluy other features we could highlight.

deploy/config/session-affinity-config.yaml
elevran, line 1: I don't think this file and the next should be incldued as part of the code base.
It should suffice to define the configuration options and let users roll out their own config files.
The sample values should instead be listed as examples in the configuration description of the Plugin.

deploy/config/session-affinity-production-config.yaml
elevran, line 1: See previous file comment.

docs/architecture.md
```diff
@@ -437,10 +437,79 @@ Scores are normalized to a range of 0-1, where pods with fewer active requests g
...
+- **Path**: `/`
+- **HttpOnly**: `true` (prevents JavaScript access for security)
+- **SameSite**: `Lax` (provides CSRF protection while allowing normal navigation)
+
+**Configuration:**
 - **Type**: `session-affinity-scorer`
-- **Parameters**: None
+- **Parameters**:
+  - `maxAge` (int, optional): Cookie's Max-Age in seconds. Default: 0 (session cookie that expires when browser closes)
+  - `secure` (bool, optional): Whether cookie should only be sent over HTTPS. Default: false. **Should be true in production**
```
elevran: Q: Does that value depend on the protocol scheme used by the client?
That is, TLS should be used in production inference and is not a cookie specific configuration?

docs/architecture.md
```diff
@@ -437,10 +437,79 @@ Scores are normalized to a range of 0-1, where pods with fewer active requests g
...
+  - `sameSite` (string, optional): SameSite attribute. Valid values: "Lax" (default), "Strict", "None"
```
elevran: Q: this makes sense for Web traffic, do we expect it to matter in local inference serving?

docs/architecture.md
```diff
@@ -437,10 +437,79 @@ Scores are normalized to a range of 0-1, where pods with fewer active requests g
...
+- **Name**: `llm-d-session`
+- **Value**: Base64-encoded pod identifier (namespace/name)
+- **Path**: `/`
+- **HttpOnly**: `true` (prevents JavaScript access for security)
+- **SameSite**: `Lax` (provides CSRF protection while allowing normal navigation)
+
+**Configuration:**
 - **Type**: `session-affinity-scorer`
-- **Parameters**: None
+- **Parameters**:
```
elevran: Consider if we can minimize the configuration values exposed to the user.
Is it something we really expect admin to configure?

docs/session_affinity.md
elevran, line 1: I believe the description in the architecture.md file is sufficient for most use cases.

pkg/plugins/scorer/session_affinity.go
```diff
@@ -17,22 +19,56 @@ const (
...
 )
 
 // compile-time type assertion
 var _ scheduling.Scorer = &SessionAffinity{}
-var _ requestcontrol.ResponseComplete = &SessionAffinity{}
+var _ requestcontrol.ResponseReceived = &SessionAffinity{}
+
+// SessionAffinityConfig holds configuration for the SessionAffinity scorer.
+type SessionAffinityConfig struct {
+	// MaxAge specifies the cookie's Max-Age attribute in seconds.
```
elevran: See earlier comments on simplifying configuration options.

pkg/plugins/scorer/session_affinity.go
```diff
@@ -17,22 +19,56 @@ const (
...
 }
 
-// NewSessionAffinity returns a scorer
-func NewSessionAffinity() *SessionAffinity {
+// NewSessionAffinity returns a scorer with the given configuration
+func NewSessionAffinity(config *SessionAffinityConfig) *SessionAffinity {
+	if config == nil {
+		config = &SessionAffinityConfig{
+			SameSite: "Lax",
+		}
```
elevran: consider a `defaultConfig()` function to be used here and in Factory?

pkg/plugins/scorer/session_affinity.go
```diff
@@ -63,17 +100,22 @@ func (s *SessionAffinity) Category() scheduling.ScorerCategory {
...
+	// Extract session cookie from Cookie header
+	cookieHeader := request.Headers["cookie"]
+	if cookieHeader != "" {
+		sessionToken := extractCookieValue(cookieHeader, sessionCookieName)
+		if sessionToken != "" {
+			decodedBytes, err := base64.StdEncoding.DecodeString(sessionToken)
+			if err != nil {
+				log.FromContext(ctx).Error(err, "Error decoding session cookie")
+			} else {
+				podName = string(decodedBytes)
```
elevran:
```suggestion
				target = string(decodedBytes)
```
nit: we no longer use Pods as the term for the target but rather endpoint, since `endpoint` is used below, perhaps `target` is a fitting name.

pkg/plugins/scorer/session_affinity.go
```diff
@@ -98,9 +138,59 @@ func (s *SessionAffinity) ResponseComplete(ctx context.Context, _ *scheduling.LL
...
 	}
 
-	if response.Headers == nil { // TODO should always be populated?
+	if response.Headers == nil {
 		response.Headers = make(map[string]string)
 	}
 
-	response.Headers[sessionTokenHeader] = base64.StdEncoding.EncodeToString([]byte(targetPod.NamespacedName.String()))
+	// Create session cookie with encoded pod name
+	sessionToken := base64.StdEncoding.EncodeToString([]byte(targetPod.NamespacedName.String()))
```
elevran: Q: should this cookie be set on every response or only on the first response (or when the target selected changes)?
IIRC, the Set-Cookie can be applied once per session and no need to set it again on every repsonse. I think you may still have the original request available so can validate if present or not.

scripts/kind-dev-env.sh
```diff
@@ -67,6 +67,9 @@ export PD_ENABLED="\"${PD_ENABLED:-false}\""
 # By default we are not setting up for KV cache
 export KV_CACHE_ENABLED="${KV_CACHE_ENABLED:-false}"
 
+# By default session affinity is disabled
```
elevran: See earlier comment on special casing specific plugins - it should be up to the developer to have a suitable config file and use it in testing/deployment.

scripts/kubernetes-dev-env.sh
```diff
@@ -83,7 +83,17 @@ export PD_ENABLED="\"${PD_ENABLED:-false}\""
 # Token length threshold to trigger P/D logic
 export PD_PROMPT_LEN_THRESHOLD="\"${PD_PROMPT_LEN_THRESHOLD:-10}\""
 
-export EPP_CONFIG="${EPP_CONFIG:-deploy/config/epp-prefix-cache-tracking-config.yaml}"
+# By default session affinity is disabled
```
elevran: ditto - developers can set `$EPP_CONFIG` to their preferred local configuration before deployment.

deploy/config/session-affinity-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+- type: decode-filter
+- type: max-score-picker
+- type: single-profile-handler
+schedulingProfiles:
+- name: default
+  plugins:
+  - pluginRef: decode-filter
+  - pluginRef: max-score-picker
+  - pluginRef: session-affinity-scorer
+    weight: 100  # High weight ensures session affinity takes precedence when cookie is present
```
kfswain: @elevran the provided config shows using KV & session, which would hard-bind when system prompts are used.

kfswain:
> Is it multiple sessions with the same/similar initial prompt gravitating to the single server due to kv affinity and then locking to it due to this scorer?

Yes. Tagged you in a comment wiht the relevant config

Personally, I think we need some reasonable evidence that a user _should_ walk this path before we advertise/adopt.

docs/architecture.md
```diff
@@ -437,10 +437,79 @@ Scores are normalized to a range of 0-1, where pods with fewer active requests g
 #### SessionAffinity
 
 Scores the candidate pods by giving a higher score to the pods that were previously
-used for the same session.
-
+used for the same session. Session affinity is implemented using standard HTTP cookies,
+which are automatically handled by browsers and HTTP clients.
```
elevran: Consider adding a short explanation of relation between this and prefix caching (e.g., expected reuse, recommending this as a simpler and more limited alternative, not running both in the same configuration).
The example configurations should follow that (e.g., not include with prefix scorer)

elevran (commented):
@roytman - thanks for the PR.
Can you please simplify it by focusing on fixing the session affinity scorer bugs only?
I would prefer removing most of the shell, markdown and YAML content for now. We should make usage recommendations after validating in benchmarking.

docs/architecture.md
```diff
@@ -437,10 +437,79 @@ Scores are normalized to a range of 0-1, where pods with fewer active requests g
...
+
+**Configuration:**
 - **Type**: `session-affinity-scorer`
-- **Parameters**: None
+- **Parameters**:
+  - `maxAge` (int, optional): Cookie's Max-Age in seconds. Default: 0 (session cookie that expires when browser closes)
+  - `secure` (bool, optional): Whether cookie should only be sent over HTTPS. Default: false. **Should be true in production**
+  - `sameSite` (string, optional): SameSite attribute. Valid values: "Lax" (default), "Strict", "None"
+
+**Example Configuration (Basic):**
```
elevran: I think it would be sufficient to include only one example and show only relevant snippet for this scorer instead of a full configuration file.

README.md
roytman: I removed the changes in the file

DEVELOPMENT.md
roytman: I reverted the changes in the file.

scripts/kubernetes-dev-env.sh
roytman: I reverted the changes in the configuration file.

scripts/kind-dev-env.sh
roytman: I reverted the changes in the configuration file.

roytman:
> @roytman - thanks for the PR. Can you please simplify it by focusing on fixing the session affinity scorer bugs only? I would prefer removing most of the shell, markdown and YAML content for now. We should make usage recommendations after validating in benchmarking.

OK, I’ve reverted the changes in the shell, Markdown, and YAML files. I kept only docs/session_affinity.md; it serves as a guide for working with the plugin and can later be removed or integrated into docs/architecture.md or other plugin-related sections.

docs/architecture.md
roytman: This is a cookie security attribute that instructs browsers to transmit cookies only over encrypted connections (TLS/HTTPS). In our use case, it isn't required.

pkg/plugins/scorer/session_affinity.go
```diff
@@ -17,22 +19,45 @@ const (
...
+var _ requestcontrol.ResponseReceived = &SessionAffinity{}
 
 // SessionAffinityFactory defines the factory function for SessionAffinity scorer.
-func SessionAffinityFactory(name string, _ json.RawMessage, _ plugin.Handle) (plugin.Plugin, error) {
-	return NewSessionAffinity().WithName(name), nil
+func SessionAffinityFactory(name string, rawConfig json.RawMessage, _ plugin.Handle) (plugin.Plugin, error) {
+	maxAge := defaultMaxAge
+
+	if len(rawConfig) > 0 {
+		var config map[string]interface{}
```
elevran: Unless there is some reason, perhaps define a config structure with json struct tags so it can be Unmarshalled directly instead of using a generic map?

pkg/plugins/scorer/session_affinity.go
```diff
@@ -63,44 +89,102 @@ func (s *SessionAffinity) Category() scheduling.ScorerCategory {
...
+	expectedSessionToken := base64.StdEncoding.EncodeToString([]byte(targetEndpoint.NamespacedName.String()))
+
+	// Check if client already has the correct cookie
+	if request != nil && request.Headers != nil {
+		cookieHeader := request.Headers[cookieHeaderName]
+		if cookieHeader != "" {
+			existingSessionToken := extractCookieValue(cookieHeader, sessionCookieName)
+			if existingSessionToken == expectedSessionToken {
+				// Cookie already exists with correct value, no need to set it again
+				logger.V(logutil.DEBUG).Info("Session cookie already exists with correct value, skipping Set-Cookie",
```
elevran: nit: to aovid overloading the logm consider only logging when the cookie is set.

pkg/plugins/scorer/session_affinity.go
```diff
@@ -63,44 +89,102 @@ func (s *SessionAffinity) Category() scheduling.ScorerCategory {
...
 		response.Headers = make(map[string]string)
 	}
 
-	response.Headers[sessionTokenHeader] = base64.StdEncoding.EncodeToString([]byte(targetPod.NamespacedName.String()))
+	cookie := &http.Cookie{
+		Name:     sessionCookieName,
+		Value:    expectedSessionToken,
+		Path:     "/",
+		HttpOnly: true,
+		Secure:   false,
```
elevran, line 169: Q: Is it possible to set these based on the request? Assuming we have the full URL passed in from the gateway.

pkg/plugins/scorer/session_affinity.go
```diff
@@ -63,44 +89,102 @@ func (s *SessionAffinity) Category() scheduling.ScorerCategory {
...
+	cookieStr := cookie.String()
+
+	logger.V(logutil.DEBUG).Info("Setting session cookie",
+		"sessionToken", expectedSessionToken,
+		"targetEndpoint", targetEndpoint.NamespacedName.String())
+
+	// Append to existing Set-Cookie headers if any
+	existingSetCookie := response.Headers[setCookieHeaderName]
+	if existingSetCookie != "" {
+		response.Headers[setCookieHeaderName] = existingSetCookie + "\n" + cookieStr
```
elevran: Q: is newline the expected separator in a cookie?
Asking since you split on `;` L181.
Is there a standard lib way of parsing and concatenating cookie values?

pkg/plugins/scorer/session_affinity_test.go
```diff
@@ -1,147 +1,369 @@
-package scorer_test
+package scorer
```
elevran: Q: do you need access to package variables or methods? if not, can ermain with a `_test` package.

pkg/plugins/scorer/session_affinity_test.go
```diff
@@ -1,147 +1,369 @@
...
 
-	// valid session token for endpointB
-	validSessionTokenForEndpointB := base64.StdEncoding.EncodeToString([]byte(endpointB.GetMetadata().NamespacedName.String()))
+	// Create mock endpoints
+	endpoint1 := newTestEndpointForSession("pod-1", "default")
+	endpoint2 := newTestEndpointForSession("pod-2", "default")
+	endpoints := []scheduling.Endpoint{endpoint1, endpoint2}
 
-	sessionAffinityScorer := scorer.NewSessionAffinity()
+	t.Run("No cookie - all endpoints get zero score", func(t *testing.T) {
```
elevran: can these be written as table driven tests instead of calling `t.Run()` for each? That's how we typcially handle multiple test cases for the same type of test.
The same applies to other Test functions.

elevran (commented):
Some questions for clarification and style changes requested.

pkg/plugins/scorer/session_affinity.go
roytman: Done. (One of the advantages of resetting Set-Cookie is updating its Max-Age)

pkg/plugins/scorer/session_affinity_test.go
roytman: done

pkg/plugins/scorer/session_affinity.go
roytman: OK, removed

pkg/plugins/scorer/session_affinity.go
roytman: done

pkg/plugins/scorer/session_affinity.go
roytman: I don't believe this is necessary. Since the cookie isn't tied to the application logic, there's no significant risk in transmitting it over a non-secure connection. Let me know if you disagree.

pkg/plugins/scorer/session_affinity.go
roytman: HTTP Specification (RFC 6265):

> Multiple Set-Cookie headers should be sent as separate header lines
> Each cookie needs its own Set-Cookie header

Correct HTTP Response:
```
HTTP/1.1 200 OK
Set-Cookie: session=abc123; Path=/; HttpOnly
Set-Cookie: llm-d-session=xyz789; Path=/; HttpOnly
```
When building HTTP headers in a map structure (like response.Headers), multiple values for the same header name are typically concatenated with newlines. The HTTP framework/proxy then splits on "\n" to create separate header lines.
Example:
```go
response.Headers["Set-Cookie"] = "cookie1=value1\ncookie2=value2"
```
Gets sent as:
```
Set-Cookie: cookie1=value1
Set-Cookie: cookie2=value2
```

L181 deals with the "Cookie" header, where different values of the same header are separated by ";"

pkg/plugins/scorer/session_affinity_test.go
roytman: done

docs/session_affinity.md
```diff
@@ -0,0 +1,224 @@
...
+      { role: 'user', content: 'Hello' },
+      { role: 'assistant', content: 'Hi there!' },
+      { role: 'user', content: 'How are you?' }
+    ]
+  })
+});
+```
+Note: modern browsers check CORS (Cross-Origin Resource Sharing) settings and can prevent cookies sending even if "Set-Cookie" exist. Please check your browser configuration.
+
+
```
elevran:
```suggestion
```

docs/session_affinity.md
```diff
@@ -0,0 +1,224 @@
...
+- type: prefix-cache-scorer
+- type: max-score-picker
+- type: single-profile-handler
+schedulingProfiles:
+- name: default
+  plugins:
+  - pluginRef: max-score-picker
+  - pluginRef: session-affinity-scorer
+    weight: 70  # Adjust based on how strongly you want session affinity
+  - pluginRef: prefix-cache-scorer
```
elevran: Based on previous discussions, it seems that we might not want to run session affinity and prefix caching.
They have overlapping functionality and can have adverse effects when used together.

elevran: Based on previous discussions, it seems that we might not want to run session affinity and prefix caching.
They have overlapping functionality and can have adverse effects when used together.

roytman: I removed the file

elevran:
@kfswain given #815, can we push this one through (@roytman after conflicts are resolved)?

pkg/plugins/scorer/session_affinity.go
```diff
@@ -63,44 +99,101 @@ func (s *SessionAffinity) Category() scheduling.ScorerCategory {
...
 		return
 	}
 
-	if response.Headers == nil { // TODO should always be populated?
+	// Create session token for the target endpoint
+	expectedSessionToken := base64.StdEncoding.EncodeToString([]byte(targetEndpoint.NamespacedName.String()))
+
+	// Check if client already has the correct cookie
+	if request != nil && request.Headers != nil {
+		cookieHeader := request.Headers[CookieHeaderName]
```
zetxqx, line 149: what if a request doesn't want to use session affinity? Reading the logic, if we have this Plugin enabled, we always set the cookie for the caller.

elevran: fair point. It should be ok to operate in either use-if-present only more.

elevran:
@roytman thank you for this contribution.
Unfortunately the code base has been refactored due to the GAIE migration and the introduction of per request attributes (e.g., see discussion on #815).
Should you wish to implement the plugin based on the attribute in #1372, please self assign #815.

Issue #594, opened 2026-01-31
`make install-rbac` fails with "no objects passed"


DhritiShikhar:
```sh
> make install-rbac
Applying RBAC configuration from deploy/rbac...
/home/dhritishikhar/Documents/repo/github/llm/redhat/llm-d-inference-scheduler/bin/kustomize build deploy/environments/kubernetes-base/rbac | envsubst '$PROJECT_NAME $NAMESPACE $IMAGE_TAG_BASE $VERSION' | kubectl apply -f -
Error: unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization' in directory '/home/dhritishikhar/Documents/repo/github/llm/redhat/llm-d-inference-scheduler/deploy/environments/kubernetes-base/rbac'
error: no objects passed to apply
make: *** [Makefile.cluster.mk:19: install-rbac] Error 1
```

PR #601, opened 2026-02-08
fix: Add kustomization file for rbac

albertoperdomo2:
## Summary

- Fixes `make install-rbac` failing with "no objects passed" error by adding the missing `kustomization.yaml` to `deploy/environments/kubernetes-base/rbac/`

  ## Test plan
  - [x] Run `make install-rbac` and verify it no longer fails with "unable to find one of 'kustomization.yaml'" error
  - [x] Verify the applied RBAC resources are correct (`kubectl get roles`, `kubectl get rolebindings`)

## Related issues
- #594

### DIFF
--- a/deploy/config/pd-epp-config.yaml
+++ b/deploy/config/pd-epp-config.yaml
@@ -1,13 +1,25 @@
 # Sample EPP configuration for tunning with P/D
 apiVersion: inference.networking.x-k8s.io/v1alpha1
 kind: EndpointPickerConfig
+featureGates:
+- prepareDataPlugins
 plugins:
 - type: prefill-header-handler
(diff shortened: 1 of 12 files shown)


albertoperdomo2:
@shmuelk given that we expect `make install-rbac` to work on its own (which only applies **rbac**), if I add the `rbac/` directory instead of the individual files we will end up with double `namePrefix` (set to `${PROJECT_NAME}-`) if we ever apply the entire `kubernetes-base` - the one from the `kubernetes-base/kustomization.yaml` and the one from the `rbac/kustomization.yaml`.

shmuelk (approved):

PR #603, opened 2026-02-09
Enable chunked decode in the routing proxy

andreyod:
Add in the routing proxy sidecar ability to run the decode request in chunks by setting max_completion_tokens as the output chunk size. This will allow to break a long decode requests in to multiple smaller requests.
When all the iterations are done, single response will be returned. 
This will allow more fairness and prevent head-of-line blocking.


andreyod:
closing in favor of #822

PR #625, opened 2026-02-16
Cleanup End to End test framework

shmuelk:
This PR cleans up the "framework" used by the end to end tests.

In particular it:

1. Deletes a duplicated load of the vLLM Simulator image
2. Deletes a forced Pull of the images needed from the code. This code is not needed for two reasons:
     1. The Makefile target test-e2e has a pre-requisite target of image-pull, which runs a scripts that verifies that the needed images are in the local directory.
     2. The code causes a pull of images from the remote repository even if an image with the appropriate tag exists locally.


nirrozenbaum (approved):
/lgtm
/approve

PR #632, opened 2026-02-18
Fix panic in SGLang proxy handling of concurrent requests

### LINKED ISSUE #631: SGLang connector crash when sending concurrent requests
**What happened**:
When dispatching requests to an SGLang prefill/decode disaggregated setup via the `pd-sidecar` proxy with a concurrency $>1$, the sidecar proxy crashes and restarts, shutting down the connection. The crash produces a `panic: net/http: abort Handler` stack trace originating from `connector_sglang.go` where `ReverseProxy.ServeHTTP` is called inside a background goroutine. 

The sidecar proxy should successfully proxy concurrent traffic to the prefill and decode backends without crashing. The `net/http` server should not panic when a single overlapping request is closed or cancelled.

1. Deploy `llm-d-inference-scheduler` with the SGLang connector enabled (`--connector=sglang`).
2. Point traffic to the `pd-sidecar` proxy at `concurrency=1` and observe that requests succeed without errors.
3. Increase traffic benchmarking to `concurrency=2` or higher.
4. The sidecar proxy will immediately panic with `http.ErrAbortHandler` and the pod will restart.

The root cause is a context cancellation race condition:
The `sendSGLangConcurrentRequests` method in `connector_sglang.go` clones the incoming HTTP request context (`r.Clone(r.Context())`) for both the synchronous decode request and the asynchronous background prefill request. When the synchronous decode proxy completes, the main `net/http` request handler returns, which automatically cancels the shared base context. If the background prefill proxy is still in the middle of reading the `200 OK` response stream when this context finishes, `httputil.ReverseProxy` interprets it as an aborted connection and throws a panic.

Because this call is in a raw `go func()` without an outer handler catching it, it bubbles up and takes down the binary.

The fix I have implemented is decoupling the prefill context and wrapping the async call in a `defer recover`:
```go
// Decouple background prefill lifecycle from the synchronous decode return
prefillReq := cloneWithJSONBody(context.WithoutCancel(r.Context()), r, body)
decodeReq := cloneWithJSONBody(r.Context(), r, body)
```

**Environment**:
- Kubernetes version (use `kubectl version`): `v1.34.3-gke.1318000`
- Cloud provider or hardware configuration: GKE
- Install tools: Helmfile
- Others: SGLang backend 

**Attachment**:
Attached debug log of the prefill node, decode node, and the sidecar.

[sglang_crash.log](https://github.com/user-attachments/files/25383023/sglang_crash.log)

yangligt2:
Fix #631 

### Description
This PR addresses a critical connection termination bug where the `pd-sidecar` panics with `net/http: abort Handler` under concurrent load (`concurrency > 1`) when running with the SGLang connector.

**Root Cause:**
In the SGLang architecture, the sidecar splits incoming requests and submits them to both the prefill and decode backends concurrently. The decode pipeline runs synchronously, while the prefill pipeline is dispatched asynchronously in a background goroutine. 

Both requests were previously inheriting the identical request context using `r.Clone(r.Context())`. Because the prefill `200 OK` response stream frequently arrives *after* the decode processing has successfully completed and returned, the standard Go HTTP server automatically cancels the parent context `r.Context()`. 

When this cancellation occurs before the background prefill proxy has fully read its response stream, `httputil.ReverseProxy` interprets it as an aborted connection and fatals with `http.ErrAbortHandler`, bringing down the sidecar pod. 

As seen in the sidecar debug logs under load, the prefill request completions (`prefill request completed`) log significantly later than the target forwarding:
```json
...
{"level":"Level(-4)","ts":"2026-02-18T05:45:31Z","logger":"proxy server on port 8000","msg":"SSRF protection: prefill target allowed","target":"10.120.11.7:8000"}
{"level":"Level(-4)","ts":"2026-02-18T05:45:31Z","logger":"proxy server on port 8000","msg":"running SGLang protocol","url":"10.120.11.7:8000"}
{"level":"Level(-5)","ts":"2026-02-18T05:45:31Z","logger":"proxy server on port 8000","msg":"bootstrap info added","bootstrap_host":"10.120.11.7","bootstrap_port":8998,"bootstrap_room":1771393531315488403}
{"level":"Level(-5)","ts":"2026-02-18T05:45:31Z","logger":"proxy server on port 8000","msg":"prefill request completed","status":200}
{"level":"Level(-5)","ts":"2026-02-18T05:45:31Z","logger":"proxy server on port 8000","msg":"prefill request completed","status":200}
{"level":"Level(-5)","ts":"2026-02-18T05:45:31Z","logger":"proxy server on port 8000","msg":"prefill request completed","status":200}
{"level":"Level(-5)","ts":"2026-02-18T05:45:31Z","logger":"proxy server on port 8000","msg":"prefill request completed","status":200}
{"level":"Level(-5)","ts":"2026-02-18T05:45:31Z","logger":"proxy server on port 8000","msg":"prefill request completed","status":200}
```

**Fix:**
This PR properly unties the background prefill context lifecycle from the main synchronous request using `context.WithoutCancel`. A `defer recover()` was additionally added to the async call stack for extra safety against uncaught internal proxy panics propagating to the main service loop. Added a `TestSGLang` Ginkgo test suite to validate end-to-end functionality.

### Validation
Validated in a real GKE environment with SGLang backends:
- All benchmark queries succeeded smoothly at `concurrency=512`.
- `go test -v -run SGLang` tests cleanly pass.

### DIFF
--- a/pkg/sidecar/proxy/connector_sglang_test.go
+++ b/pkg/sidecar/proxy/connector_sglang_test.go
@@ -19,6 +19,8 @@ package proxy
 import (
 	"io"
 	"net/http"
+	"net/http/httptest"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/sidecar/proxy/connector_sglang_test.go
bongwoobak, line 1: How about adding a test case for concurrency>1?

yangligt2: Added test case "should not panic when prefill response is slower than decode response", which was the direct root cause for the  concurrency>1 crash

PR #643, opened 2026-02-23
initial E/PD extension of the sidecar

roytman:
This PR is part of the implementation for https://github.com/llm-d/llm-d-inference-scheduler/issues/608. It extends the sidecar to support Encode disaggregation. 

Currently, it does not support advanced Encoder selection optimizations; it simply selects a random Encoder from the available options. More sophisticated scheduling strategies will be introduced in future updates.

### DIFF
--- a/.github/actions/trivy-scan/action.yml
+++ b/.github/actions/trivy-scan/action.yml
@@ -20,8 +20,6 @@ runs:
       uses: aquasecurity/trivy-action@0.35.0
       env:
         GITHUB_TOKEN: ${{ inputs.github-token }}
-        TRIVY_USERNAME: ${{ github.actor }}
-        TRIVY_PASSWORD: ${{ inputs.github-token }}      
       with:
         image-ref: ${{ inputs.image }}
         format: 'table'
--- a/pkg/plugins/pre-request/pd_prerequest_test.go
+++ b/pkg/plugins/pre-request/pd_prerequest_test.go
@@ -1,0 +1,233 @@
+package prerequest
+
+import (
+	"encoding/json"
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+	fwkdl "sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/datalayer"
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/scheduling"
+
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/common"
+	"github.com/llm-d/llm-d-inference-scheduler/test/utils"
+)
+
+const (
+	testAddr     = "10.0.0.5"
+	testPort     = "8000"
+	testIPv6Addr = "fd00::1"
+)
+
+func makeEndpoint(addr string) scheduling.Endpoint {
+	return scheduling.NewEndpoint(
+		&fwkdl.EndpointMetadata{
+			NamespacedName: k8stypes.NamespacedName{Namespace: "default", Name: "prefill-pod"},
+			Address:        addr,
+			Port:           testPort,
+		},
+		&fwkdl.Metrics{},
+		nil,
+	)
+}
+
+func TestPrefillHeaderHandlerFactory(t *testing.T) {
+	tests := []struct {
+		name          string
+		pluginName    string
+		rawParams     string
+		expectErr     bool
+		expectProfile string
+		expectName    string
+	}{
+		{
+			name:          "default parameters",
+			pluginName:    "my-handler",
+			rawParams:     "",
+			expectErr:     false,
+			expectProfile: "prefill",
+			expectName:    "my-handler",
+		},
+		{
+			name:          "custom prefill profile",
+			pluginName:    "custom-handler",
+			rawParams:     `{"prefillProfile": "my-prefill"}`,
+			expectErr:     false,
+			expectProfile: "my-prefill",
+			expectName:    "custom-handler",
+		},
+		{
+			name:       "invalid json",
+			pluginName: "bad-handler",
+			rawParams:  `{invalid}`,
+			expectErr:  true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var raw json.RawMessage
+			if tt.rawParams != "" {
+				raw = json.RawMessage(tt.rawParams)
+			}
+
+			p, err := PrefillHeaderHandlerFactory(tt.pluginName, raw, nil)
+			if tt.expectErr {
+				assert.Error(t, err)
+				assert.Nil(t, p)
+				return
+			}
+
+			require.NoError(t, err)
+			require.NotNil(t, p)
+
+			handler, ok := p.(*PrefillHeaderHandler)
+			require.True(t, ok)
+			assert.Equal(t, tt.expectName, handler.TypedName().Name)
+			assert.Equal(t, PrefillHeaderHandlerType, handler.TypedName().Type)
+			assert.Equal(t, tt.expectProfile, handler.prefillProfile)
+		})
+	}
+}
+
+func TestPreRequestPrefillProfileExists(t *testing.T) {
+	ctx := utils.NewTestContext(t)
+	handler := NewPrefillHeaderHandler("prefill").WithName("test")
+
+	request := &scheduling.LLMRequest{
+		TargetModel: "test-model",
+		RequestId:   "req-123",
+		Headers:     map[string]string{},
+	}
+
+	result := &scheduling.SchedulingResult{
+		PrimaryProfileName: "decode",
+		ProfileResults: map[string]*scheduling.ProfileRunResult{
+			"prefill": {
+				TargetEndpoints: []scheduling.Endpoint{
+					makeEndpoint(testAddr),
+				},
+			},
+		},
+	}
+
+	handler.PreRequest(ctx, request, result)
+
+	assert.Equal(t, net.JoinHostPort(testAddr, testPort), request.Headers[common.PrefillPodHeader])
+}
+
+func TestPreRequestPrefillProfileNotExists(t *testing.T) {
+	ctx := utils.NewTestContext(t)
+	handler := NewPrefillHeaderHandler("prefill").WithName("test")
+
+	request := &scheduling.LLMRequest{
+		Headers: map[string]string{},
+	}
+
+	result := &scheduling.SchedulingResult{
+		PrimaryProfileName: "decode",
+		ProfileResults:     map[string]*scheduling.ProfileRunResult{},
+	}
+
+	handler.PreRequest(ctx, request, result)
+
+	_, exists := request.Headers[common.PrefillPodHeader]
+	assert.False(t, exists)
+}
+
+func TestPreRequestClearsExistingHeader(t *testing.T) {
+	ctx := utils.NewTestContext(t)
+	handler := NewPrefillHeaderHandler("prefill").WithName("test")
+
+	request := &scheduling.LLMRequest{
+		Headers: map[string]string{
+			common.PrefillPodHeader: "old-host:9999",
+		},
+	}
+
+	result := &scheduling.SchedulingResult{
+		PrimaryProfileName: "decode",
+		ProfileResults: map[string]*scheduling.ProfileRunResult{
+			"prefill": {
+				TargetEndpoints: []scheduling.Endpoint{
+					makeEndpoint(testAddr),
+				},
+			},
+		},
+	}
+
+	handler.PreRequest(ctx, request, result)
+
+	assert.Equal(t, net.JoinHostPort(testAddr, testPort), request.Headers[common.PrefillPodHeader])
+}
+
+func TestPreRequestClearsHeaderWhenNoPrefillResult(t *testing.T) {
+	ctx := utils.NewTestContext(t)
+	handler := NewPrefillHeaderHandler("prefill").WithName("test")
+
+	request := &scheduling.LLMRequest{
+		Headers: map[string]string{
+			common.PrefillPodHeader: "stale-host:9999",
+		},
+	}
+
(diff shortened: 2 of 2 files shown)


cmd/pd-sidecar/main.go
```diff
@@ -32,22 +32,30 @@ import (
...
-		proxy.ConnectorSGLang,
+	// supportedKVConnectors defines all valid KV (Prefiller-Decoder) connector types
+	supportedKVConnectors = []string{
+		proxy.KVConnectorNIXLV2,
+		proxy.KVConnectorSharedStorage,
+		proxy.KVConnectorSGLang,
+	}
+
+	// supportedECConnectors defines all valid EC (Encoder-Prefiller) connector types
+	supportedECConnectors = []string{
```
elevran: Q: if you run EP/D does that require an EC connector or only KV connector?

cmd/pd-sidecar/main.go
```diff
@@ -32,22 +32,30 @@ import (
...
 
 func main() {
 	port := flag.String("port", "8000", "the port the sidecar is listening on")
 	vLLMPort := flag.String("vllm-port", "8001", "the port vLLM is listening on")
 	vLLMDataParallelSize := flag.Int("data-parallel-size", 1, "the vLLM DATA-PARALLEL-SIZE value")
-	connector := flag.String("connector", proxy.ConnectorNIXLV2, "the P/D connector being used. Supported: "+strings.Join(supportedConnectors, ", "))
+	kvConnector := flag.String("kv-connector", proxy.KVConnectorNIXLV2, "the KV connector between Prefiller and Decoder. Supported: "+strings.Join(supportedKVConnectors, ", "))
+	ecConnector := flag.String("ec-connector", proxy.ECExampleConnector, "the EC connector between Encoder and Prefiller (optional, for EPD mode). Supported: "+strings.Join(supportedECConnectors, ", "))
 	prefillerUseTLS := flag.Bool("prefiller-use-tls", false, "whether to use TLS when sending requests to prefillers")
+	encoderUseTLS := flag.Bool("encoder-use-tls", false, "whether to use TLS when sending requests to encoders")
```
elevran: See #682

cmd/pd-sidecar/main.go
```diff
@@ -32,22 +32,30 @@ import (
...
+		proxy.ECExampleConnector,
 	}
 )
 
 func main() {
 	port := flag.String("port", "8000", "the port the sidecar is listening on")
 	vLLMPort := flag.String("vllm-port", "8001", "the port vLLM is listening on")
 	vLLMDataParallelSize := flag.Int("data-parallel-size", 1, "the vLLM DATA-PARALLEL-SIZE value")
-	connector := flag.String("connector", proxy.ConnectorNIXLV2, "the P/D connector being used. Supported: "+strings.Join(supportedConnectors, ", "))
+	kvConnector := flag.String("kv-connector", proxy.KVConnectorNIXLV2, "the KV connector between Prefiller and Decoder. Supported: "+strings.Join(supportedKVConnectors, ", "))
```
elevran: For backward compatibility and migration, please leave `--connector` as is (marking it deprecated) and use the new `--kv-connector` when available. If both are defined, prefer `--kv-*` and log this.

cmd/pd-sidecar/main.go
```diff
@@ -86,19 +94,37 @@ func main() {
...
 	if !isValidConnector {
-		logger.Info("Error: --connector must be one of: " + strings.Join(supportedConnectors, ", "))
+		logger.Info("Error: --kv-connector must be one of: " + strings.Join(supportedKVConnectors, ", "))
 		return
 	}
-	logger.Info("p/d connector validated", "connector", connector)
+	logger.Info("KV connector (prefiller-decoder) validated", "kvConnector", kvConnector)
+
+	// Validate EC connector (Encoder-Prefiller) if specified
+	if *ecConnector != "" {
```
elevran: nit: realize you're being consistent with code above, but a map/set would work better. Made similar comment on #682

elevran: Note that `if *ecConnector != ""` is a less likely state since you set a a non-empty default for the flag. You may want to check if the flag was set by the user or not (in addition or instead of this check).
By using a default EC connector , the sidecar always starts with it, even when not operating in EPD disaggregation mode.

pkg/common/common.go
```diff
@@ -12,6 +12,9 @@ const (
 
 	// DataParallelPodHeader is the header name used to indicate the worker <ip:port> for Data Parallel
 	DataParallelPodHeader = "x-data-parallel-host-port"
+
+	// EncoderHostsPortsHeader is the header name used to indicate Encoder workers <ip:port> list
```
elevran, line 16: nit: clarify list format (e.g., comma separated?). Ensure it is HTTP header compatible.
consider adding convenience functions to encode/decode the list.

Q: (unrelated to this PR, raising for awareness/feedback) - we could extend the host port with a protocol (e.g., `https://<ip>:<port>` and rid of the tls configuration. It would be up to the EPP to send this in the correct format. For backward compatibility, can use `http` if protocol is ommitted.

pkg/sidecar/proxy/chat_completions.go
```diff
@@ -60,7 +60,8 @@ func (s *Server) chatCompletionsHandler(w http.ResponseWriter, r *http.Request)
 		requestPath = r.URL.Path
 	}
 	span.SetAttributes(
-		attribute.String("llm_d.pd_proxy.connector", s.config.Connector),
+		attribute.String("llm_d.pd_proxy.kv_connector", s.config.KVConnector),
```
elevran, line 65: this could break existing dashboards. Suggest keeping the attribute unchanged (can consider adding another and then deprecating the attribute in a later release).

pkg/sidecar/proxy/connector_epd_shared_storage.go
```diff
@@ -0,0 +1,250 @@
...
+	// Create a deep copy of the original request
+	encoderRequest := make(map[string]any)
+	for k, v := range originalRequest {
+		encoderRequest[k] = v
+	}
+
+	// Build messages with only the MM item (no text)
+	messages := []map[string]any{
+		{
+			"role": "user",
```
elevran, line 89: Q: is this always the `user` role? What happens with (e.g.,) images created by the agent in a previous reponse, would/could they be included as well in the body with role set to `agent`?

pkg/sidecar/proxy/connector_epd_shared_storage.go
```diff
@@ -0,0 +1,250 @@
+/*
+Copyright 2025 The llm-d Authors.
```
elevran: nit 😉 
```suggestion
Copyright 2026 The llm-d Authors.
```

pkg/sidecar/proxy/connector_epd_shared_storage.go
elevran, line 1: is this the default EC connector?

pkg/sidecar/proxy/connector_epd_shared_storage.go
```diff
@@ -0,0 +1,250 @@
...
+
+	encoderRequest["messages"] = messages
+	encoderRequest["max_tokens"] = 1
+	encoderRequest["stream"] = false
+	delete(encoderRequest, "stream_options")
+
+	return encoderRequest
+}
+
+// fanoutEncoderPrimer sends concurrent requests to encoder cluster for all MM items. We assume that there is no identical MM items in the same request.
```
elevran, line 104: nit: at least in theory you could validate no identical items in the request, not sure if you'd want to but you might not need to assume.

pkg/sidecar/proxy/connector_epd_shared_storage.go
```diff
@@ -0,0 +1,250 @@
...
+
+	return nil
+}
+
+// runEPDProtocol implements the Encoder-Prefiller-Decoder disaggregation protocol
+func (s *Server) runEPDProtocol(w http.ResponseWriter, r *http.Request, prefillEndPoint string, encodeEndPoints []string) {
+	s.logger.V(4).Info("running EPD protocol", "prefiller", prefillEndPoint, "encoderCount", len(encodeEndPoints))
+
+	// Read request body
+	defer r.Body.Close() //nolint:all
```
elevran: nit: prefer not using linter supression.

pkg/sidecar/proxy/connector_epd_shared_storage.go
```diff
@@ -0,0 +1,250 @@
...
+	original, err := io.ReadAll(r.Body)
```
elevran, line 182: Q: confirming: is the request (headers and body) required for P and D requests later? If so, saved so it can reconstructed?

cmd/pd-sidecar/main.go
roytman: Only KV connector. I think I see your point

pkg/sidecar/proxy/connector_epd_shared_storage.go
roytman: wow, time is running

cmd/pd-sidecar/main.go
roytman: I see, will update this PR when #682 is merged.

cmd/pd-sidecar/main.go
roytman: Will do that after merging #682. The pflag library has built-in deprecation support

pkg/sidecar/proxy/connector_epd_shared_storage.go
roytman: Linter Warning: errcheck - error return value not checked
In a deferred statement, checking the error from Close() is often impractical since you're already in cleanup/error handling code. It's a common practice to ignore Close() errors in defer statements

elevran: Agree with the analysis but prefer making it explicit instead of bypassing the linter:
```go
defer func() {
    _ = r.Body.Close()
}()
```

roytman: Another option is to be more specific - only suppress errcheck (more compact syntax) 
```golang
defer r.Body.Close() //nolint:errcheck
```
Actually, it is the same. Which option do you prefer?

elevran: Prefer not supressing linter, if possible.

```go
   defer func() { _ = r.Body.Close() }()
```
is still a one liner if that matters

cmd/pd-sidecar/main.go
roytman: I changed the supported connectors in #689, plus added the `deprecation` check.  
Please take a look. 
As an alternative, I can use a simple Map as in #682. 

I'll update `ecConnector` validation accordingly.

cmd/pd-sidecar/main.go
roytman: done

cmd/pd-sidecar/main.go
roytman: done

roytman: fixed

pkg/sidecar/proxy/chat_completions.go
roytman: done, added another one

pkg/sidecar/proxy/connector_epd_shared_storage.go
roytman: It is the only EC connector that we have for now

pkg/sidecar/proxy/connector_epd_shared_storage.go
roytman: It is a request for encoding only, so the role doesn't matter.

pkg/sidecar/proxy/connector_epd_shared_storage.go
roytman: I'll do it in a separate PR

pkg/sidecar/proxy/connector_epd_shared_storage.go
roytman: done

pkg/sidecar/proxy/connector_epd_shared_storage.go
roytman: In the current shared storage implementation, we don't need the request/response (headers and body)

pkg/sidecar/proxy/connector_epd_shared_storage.go
roytman: I think we will move part of the shared functionality to a base class

pkg/common/common.go
roytman: An HTTP header value can include a comma (,). In fact, the comma is the standard character used to separate multiple values within a single header field.

According to the HTTP specification (RFC 9110), many headers are defined as a comma-separated list. This allows a server or client to send multiple pieces of information under the same header name.

And the values can include a colon (:) as well

roytman: I don't think we need protocol; it probably should be based on `EncoderUseTLS`. I'll check how to include it

elevran: right now it uses XYZUseTLS, I was suggesting we move it to EPP which knows about E/P/D anyway so there's no need to configure the sidecar for it (as long as we do send a protocol header). Keep as is for now, created #711 to discuss.

pkg/sidecar/proxy/connector_epd_shared_storage.go
```diff
@@ -0,0 +1,250 @@
...
+			req.Header.Set(requestHeaderRequestID, fmt.Sprintf("%s-enc-%d", requestID, itemIdx))
+
+			s.logger.V(4).Info("sending encoder request", "item", itemIdx, "to", hostPort, "requestID", requestID)
+
+			pw := &bufferedResponseWriter{}
+			encoderHandler.ServeHTTP(pw, req)
+
+			if isHTTPError(pw.statusCode) {
+				errChan <- fmt.Errorf("encoder request failed for item %d with status %d: %s", itemIdx, pw.statusCode, pw.buffer.String())
+				return
```
roytman, line 156: The error handler will be added as part of #712

Issue #608, opened 2026-02-10
Feature: Extend P/D Disaggregation to support Encode-Prefill-Decode (EPD)


roytman:
**What would you like to be added**:
This issue proposes extending our existing Prefill/Decode (P/D) disaggregation logic to support full Encode–Prefill–Decode (EPD) disaggregation.

Building on the foundation of llm-d-scheduler's current P/D capabilities, we aim to incorporate a dedicated Encoding stage. This follows the architectural pattern recently introduced by vLLM, which decouples the multimodal encoder (Vision, Audio, etc.) from the subsequent LLM generation phases.

**Why is this needed**:
While our current P/D support effectively separates compute-bound prefill from memory-bound decoding, multimodal workloads (like LLaVA or Whisper) introduce a heavy Encoding overhead.

By upgrading from P/D to E/P/D , we can:

- **Optimize Multimodal Pipelines**: Avoid idling expensive LLM GPUs while the Vision/Encoder modules are processing raw data.

- **Specialized Resource Pools**: Route the 'Encoder' phase to nodes optimized for non-autoregressive tasks, while reserving P/D pools for token generation and its associated prefill operations.

- **Cache Granularity**: Implement caching at the Encoder Cache level, allowing the output of a processed image or document to be reused across different prompt variations without re-encoding.

**Note**: This architecture allows us to evaluate various disaggregation configurations, including E/P/D (fully independent), EP/D (combined Encode-Prefill), and E/PD (combined Prefill-Decode)

Initial implementation [proposal](https://docs.google.com/document/d/1EoqmVX9VeJCi2_ZcSxYBm6OtRoyaCK41FCSMm9mqkRU/edit?usp=sharing)

**References**:
- **llm-d Documentation**: [P/D Disaggregation Guide](https://llm-d.ai/docs/guide/Installation/pd-disaggregation)
- **llm-d Related issue**: [[Feature Request] Disaggregation scheme for Vision Language Models](https://github.com/llm-d/llm-d/issues/231)
- **llm-d**: [llm-d Disaggregated Serving Northstar](https://docs.google.com/document/d/1FNN5snmipaTxEA1FGEeSH7Z_kEqskouKD1XYhVyTHr8/edit?tab=t.0#heading=h.ycwld2oth1kj)
[...]

PR #648, opened 2026-02-24
Add tests for disaggregated Encoder

revit13:
This PR is part of the implementation for #608. It extends the kind deployment to support Encode disaggregation and adds e2e tests.


scripts/kind-dev-env.sh
```diff
@@ -23,7 +23,7 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 : "${IMAGE_REGISTRY:=ghcr.io/llm-d}"
 
 # Set a default VLLM_SIMULATOR_TAG if not provided
-export VLLM_SIMULATOR_TAG="${VLLM_SIMULATOR_TAG:-latest}"
+export VLLM_SIMULATOR_TAG="${VLLM_SIMULATOR_TAG:-v0.8.1}"
```
roytman, line 26: we don't need this change

PR #652, opened 2026-02-25
Optimized request prefill error messages

learner0810:
fix: https://github.com/llm-d/llm-d-inference-scheduler/issues/637

The complete error message should be returned to the upstream rather than just an HTTP code.

- Before

```bash
~# curl -s http://inference-gateway-istio.lzj.svc.cluster.local:80/v1/chat/completions       -H "Content-Type: application/json"       -d '{
                "model": "Qwen/Qwen3-0.6B",
                "messages": [
                        {
                                "role": "user",
                                "content": [
                                        {
                                                "type": "text",
                                                "text": "Describe this image in one sentence."
                                        },
                                        {
                                                "type": "image_url",
                                                "image_url": {
                                                        "url": "https://cdn2.thecatapi.com/images/dui.jpg"
                                                }
                                        }
                                ]
                        }
                ]
        }'|jq .

```

- Now

```bash
curl -s http://inference-gateway-istio.lzj.svc.cluster.local:80/v1/chat/completions       -H "Content-Type: application/json"       -d '{
                "model": "Qwen/Qwen3-0.6B",
                "messages": [
                        {
                                "role": "user",
                                "content": [
                                        {
                                                "type": "text",
                                                "text": "Describe this image in one sentence."
                                        },
                                        {
                                                "type": "image_url",
                                                "image_url": {
                                                        "url": "https://cdn2.thecatapi.com/images/dui.jpg"
                                                }
                                        }
                                ]
                        }
                ]
        }'|jq .
{
  "error": {
    "code": 400,
    "message": "/model-cache/Qwen3-0.6B is not a multimodal model None",
    "param": null,
    "type": "BadRequestError"
  }
}
```

### DIFF
--- a/pkg/sidecar/proxy/connector_nixlv2.go
+++ b/pkg/sidecar/proxy/connector_nixlv2.go
@@ -130,10 +130,26 @@ func (s *Server) runNIXLProtocolV2(w http.ResponseWriter, r *http.Request, prefi
 	)
 
 	if isHTTPError(pw.statusCode) {
-		s.logger.Error(err, "request failed", "code", pw.statusCode)
+		s.logger.Error(err, "request failed", "code", pw.statusCode, "body", pw.buffer.String())
 		prefillSpan.SetStatus(codes.Error, "prefill request failed")
 		prefillSpan.End()
-		w.WriteHeader(pw.statusCode)
+
+		if shouldFallbackToDecode(pw) {
+			s.logger.Info("fallback to decode", "request_id", uuidStr)
+			r.Body = io.NopCloser(strings.NewReader(string(original)))
+			s.decoderProxy.ServeHTTP(w, r)
+		} else {
+			for key, values := range pw.Header() {
+				for _, v := range values {
+					w.Header().Add(key, v)
+				}
+			}
+			w.WriteHeader(pw.statusCode)
+			_, err := w.Write([]byte(pw.buffer.String()))
+			if err != nil {
+				s.logger.Error(err, "failed to send error response to client")
+			}
+		}
 		return
 	}
 	prefillSpan.End()
--- a/pkg/sidecar/proxy/proxy_helpers.go
+++ b/pkg/sidecar/proxy/proxy_helpers.go
@@ -142,3 +142,11 @@ func (s *Server) createDecoderProxyHandler(decoderURL *url.URL, decoderInsecureS
 func isHTTPError(statusCode int) bool {
 	return statusCode < http.StatusOK || statusCode >= http.StatusMultipleChoices
 }
+
+// shouldFallbackToDecode returns false for client error 4xx status codes (400–451). For all other status codes, it returns true.
+func shouldFallbackToDecode(pw *bufferedResponseWriter) bool {
+	if pw.statusCode >= http.StatusBadRequest && pw.statusCode <= http.StatusUnavailableForLegalReasons {
+		return false
+	}
+	return true
+}


pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -130,10 +130,19 @@ func (s *Server) runNIXLProtocolV2(w http.ResponseWriter, r *http.Request, prefi
 	)
 
 	if isHTTPError(pw.statusCode) {
-		s.logger.Error(err, "request failed", "code", pw.statusCode)
+		s.logger.Error(err, "request failed", "code", pw.statusCode, "body", pw.buffer.String())
```
elevran, line 133: The prefill request can fail for multiple reasons and returning an error here might not be optimal).

What do you think about approaching this in the following way instead: 
- ignore prefill on failure (and not return its error to the caller) and 
- process the request with the decode node (removing the KV cache transfer or any other P related artifacts from the request)?

Since P/D disaggregation is an optimization, this would sidestep any prefiller issues and (might?) still return a response to the user.
I think this would be a safer way to handle prefill errors/failures (e.g., when only the prefiller does not run the base model but the decoder does).

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -130,10 +130,19 @@ func (s *Server) runNIXLProtocolV2(w http.ResponseWriter, r *http.Request, prefi
...
+			}
+		}
 		w.WriteHeader(pw.statusCode)
+		_, err := w.Write([]byte(pw.buffer.String()))
+		if err != nil {
+			s.logger.Error(err, "failed to send error response to client")
+		}
 		return
 	}
 	prefillSpan.End()
```
elevran, line 155: unrelated to this PR:
if there is an error, is the `prefillerSpan` reported and does it carry any valid and useful information back?
We might want to have the span close as error for observability.


/cc: @sallyom

pkg/sidecar/proxy/connector_nixlv2.go
learner0810: I've adjusted some of the logic. Please take another look. If it's a client-side error, there's no need to forward the request to decode. In other cases, forward the request to decode.

shmuelk: A better overall solution might entail that we send multiple prefill targets in the "prefill header", if onbe fails we try another prefill node.

This is apparently very true with E/P/D, where the decode nodes are not allowed to do prefill.

@robertgshaw2-redhat

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -130,10 +130,26 @@ func (s *Server) runNIXLProtocolV2(w http.ResponseWriter, r *http.Request, prefi
...
 	if isHTTPError(pw.statusCode) {
-		s.logger.Error(err, "request failed", "code", pw.statusCode)
+		s.logger.Error(err, "request failed", "code", pw.statusCode, "body", pw.buffer.String())
 		prefillSpan.SetStatus(codes.Error, "prefill request failed")
 		prefillSpan.End()
-		w.WriteHeader(pw.statusCode)
+
+		if shouldFallbackToDecode(pw) {
+			s.logger.Info("fallback to decode", "request_id", uuidStr)
+			r.Body = io.NopCloser(strings.NewReader(string(original)))
```
elevran, line 139: Q: does the original request already contain P/D parameters? If so, should be cleared.

learner0810: The original request does not include P/D parameters.

pkg/sidecar/proxy/connector_nixlv2.go
robertgshaw2-redhat: In the specific case I was looking at, the error will occur in every pod because the model name was incorrect in my input.

I think we should take a simplicity approach in error handling at the current stage of the feature

cmd/pd-sidecar/main.go
elevran, line 1: The changes in this file seem like they are specific to a test environment and should not be included in the PR?

Dockerfile.sidecar
```diff
@@ -11,7 +11,7 @@ COPY go.mod go.mod
 COPY go.sum go.sum
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
-RUN go mod download
+RUN go env -w GOPROXY=https://goproxy.cn && go mod download
```
elevran: is this change specific to your dev environment or do you think it needs to be set always?
if there's a need for `GOPROXY` - please submit an issue or separate PR allowing (but not enforcing) the setting  and use of a GOPROXY server.

learner0810: ditto

cmd/pd-sidecar/main.go
learner0810: Yes, I'm very sorry. I accidentally submitted it. I've already deleted it.

pkg/sidecar/proxy/connector_nixlv2.go
elevran: @Gregory-Pereira - please see above comment, observability related. Can we merge in inference-scheduler or will it break otel?

PR #659, opened 2026-02-27
feat: speculative indexing for PrecisePrefixCacheScorer

bongwoobak:
Part of https://github.com/llm-d/llm-d-kv-cache/issues/353

## Summary
- Implement `PrepareDataPlugin` interface for pre-computed block keys and prefix cache match info
- Implement `PreRequest` hook to inject speculative entries into Index after routing decision
- Add TTL-based speculative cache with automatic eviction on expiry
- Fix PodIdentifier format to `ip:port` to match KV event topic format — speculative entries previously used IP-only, causing lookup mismatches with confirmed entries from KV events (which use `kv@ip:port@model` topic format)

## Details

Part 2 of Speculative Indexing (https://github.com/llm-d/llm-d-kv-cache/issues/353).
Depends on kv-cache changes: (https://github.com/llm-d/llm-d-kv-cache/pull/369)

### PrepareDataPlugin (PrepareRequestData)
- Computes block keys via `Indexer.ComputeBlockKeys()`
- Looks up KV-cache index to find which pods have matching blocks
- Stores `PrefixCacheMatchInfo` on each endpoint for downstream consumers
- Saves block keys + scores to `PluginState` for reuse by Score/PreRequest

### Score() Optimization
- Reuses pre-computed scores from PluginState when PrepareRequestData ran first
- Falls back to full computation for backward compatibility (no PrepareRequestData)

### PreRequest (Speculative Entry Injection)
- Adds speculative `PodEntry` (`Annotation: "speculative"`) to Index immediately after routing
- Handles both decode and prefill endpoints for P/D disaggregation
- Uses `ip:port` format to match KV event topic (`kv@${POD_IP}:${PORT}@${MODEL}`)

### TTL Cache for Speculative Entries
- Tracks speculative entries per request (default TTL: 30s)
- `OnEviction` callback auto-removes speculative entries from Index on expiry
- Confirmed entries from KV events are unaffected (different `Annotation` value)

### PodIdentifier ip:port Format Fix
- The scheduler constructs PodIdentifier from endpoint metadata's `Address` and `Port` fields as `ip:port`
- Previously only `Address` (IP) was used, which didn't match the PodIdentifier from KV events (`kv@ip:port@model` topic → PodIdentifier `ip:port`)
- This mismatch meant speculative entries and confirmed entries had different PodIdentifiers for the same pod, so the LongestPrefixScorer treated them as different pods and scored them separately
- Now both speculative and confirmed entries use the same `ip:port` format, enabling proper deduplication and score merging
- This is especially critical for data parallelism (DP) deployments where multiple DP ranks run on the same pod IP but with different ports (e.g., `10.0.0.1:8000` for rank 0, `10.0.0.1:8001` for rank 1). Without the port, all DP ranks would be indistinguishable and collapse into a single PodIdentifier

### Data Flow
1. PrepareRequestData --> compute blockKeys, lookup index, store scores/matchInfo
2. Score --> reuse pre-computed scores (or fallback to direct computation)
3. PreRequest --> inject speculative entries for selected pod(s), register in TTL cache
4. [KV event arrives] --> confirmed entry added (Annotation: "")
5. [TTL expires] --> speculative entry evicted, confirmed entry remains

## Test plan
- PrepareRequestData: block key computation + PluginState storage
- Score with PrepareRequestData: pre-computed data reuse
- PreRequest: speculative entries added to Index for selected endpoints
- TTL eviction: automatic cleanup after expiry, confirmed entries preserved
- E2E flow: PrepareData --> Score --> PreRequest --> KV event --> TTL eviction
- Backward compatibility: Score without PrepareRequestData works as before
- E2E cluster test with PD disaggregation (DP8): back-to-back requests with same prefix --> cache hit on 2nd request

### Breaking Change: PodIdentifier Format
- PodIdentifier now requires `ip:port` format (e.g., `10.0.0.1:8000`) to match KV event topic format (`kv@ip:port@model`)
- Previously IP-only format (`10.0.0.1`) was used, which caused mismatches with confirmed entries from KV events
- vLLM KV events already publish topics in `kv@${POD_IP}:${PORT}@${MODEL}` format, so this aligns the scheduler with the engine behavior
- **Action required**: Ensure all KV event publishers include the port in the topic string


pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -27,8 +30,42 @@ import (
 const (
 	// PrecisePrefixCachePluginType is the type-name of the PrecisePrefixCacheScorer plugin.
 	PrecisePrefixCachePluginType = "precise-prefix-cache-scorer"
+
+	// defaultSpeculativeTTL is the default TTL for speculative entries.
+	// This should be long enough for the vLLM engine to process the request
+	// and publish BlockStored events, after which confirmed entries replace them.
+	defaultSpeculativeTTL = 30 * time.Second
```
vMaroon: I think this should be lower - perhaps around 2s? In order to maintain high confidence in speculations.

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -27,8 +30,42 @@ import (
...
+
+	// stateKey is the PluginState key used to share data between
+	// PrepareRequestData, Score, and PreRequest.
+	stateKey = plugin.StateKey("precise-prefix-cache-state")
```
vMaroon: For immediate integration with other scorers, and since `prefix-cache-scorer` and `precise-prefix-cache-scorer` wont generally be active together, the key can be `prefix-cache-state` or whatever the IGW scorer has.

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -42,10 +79,58 @@ type PrecisePrefixCachePluginConfig struct {
 	// used to subscribe to KV-cache events and update the internal KV-cache
 	// index state.
 	KVEventsConfig *kvevents.Config `json:"kvEventsConfig"`
+	// SpeculativeIndexing enables speculative indexing. When true, the plugin
```
vMaroon, line 69: These updates should be reflected in `docs/configuration.md`.

bongwoobak: Do you mean `docs/architecture.md`?

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -135,12 +221,47 @@ func New(ctx context.Context, config PrecisePrefixCachePluginConfig) (*PrecisePr
...
+			entries := item.Value()
+			for _, reqKey := range entries.blockKeys {
+				// Evict speculative entries from the index.
+				// Speculative entries were added without engineKey mapping (nil engineKeys),
+				// so Evict() falls back to treating the key as a requestKey directly.
+				//nolint:errcheck // best-effort cleanup on TTL expiry
+				kvCacheIndexer.KVBlockIndex().Evict(context.Background(), reqKey, entries.podEntries)
+			}
+		})
+		go speculativeCache.Start()
```
hhk7734:
```suggestion
		go cleanCachePeriodically(ctx, speculativeCache, speculativeTTL)
```

What's the difference between it and the call on line 214?

bongwoobak: Actually same I will change it to `cleanCachePeriodically()`. Thanks

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -27,8 +30,43 @@ import (
...
 )
 
+// prefixCacheMatchInfo mirrors the upstream attrprefix.PrefixCacheMatchInfo.
+// Replace with the upstream type once the GIE dependency is updated.
+type prefixCacheMatchInfo struct {
+	matchBlocks     int
+	totalBlocks     int
+	blockSizeTokens int
+}
+
```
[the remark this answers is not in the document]
bongwoobak: this type is internal to the scorer plugin, not consumed by other packages.

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -179,8 +321,95 @@ func (s *PrecisePrefixCacheScorer) Category() scheduling.ScorerCategory {
 	return scheduling.Affinity
 }
 
+// --- PrepareDataPlugin implementation ---
+
+// Produces declares the data keys this plugin writes to endpoints.
+func (s *PrecisePrefixCacheScorer) Produces() map[string]any {
+	return map[string]any{
+		prefixCacheMatchInfoKey: prefixCacheMatchInfo{},
```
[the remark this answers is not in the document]
bongwoobak: PrepareRequestData stores a pointer via ep.Put(). No runtime issue.

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -293,6 +542,235 @@ func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, cycleState *schedu
...
+
+	if len(state.blockKeys) == 0 {
+		return
+	}
+
+	// 2. Get target endpoint from scheduling result
+	primaryResult := schedulingResult.ProfileResults[schedulingResult.PrimaryProfileName]
+	if primaryResult == nil || len(primaryResult.TargetEndpoints) == 0 {
+		return
+	}
```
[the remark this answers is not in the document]
bongwoobak: Already handled at line 577. schedulingResult itself is guaranteed non-nil by the EPP framework.

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -293,6 +546,235 @@ func (s *PrecisePrefixCacheScorer) Score(ctx context.Context, cycleState *schedu
...
+		}
+	}
+	return podSet
+}
+
+// computeScoresFromKeyToPods computes per-pod longest prefix match scores
+// from the index lookup result. This mirrors the intersection logic in the
+// LongestPrefixScorer: a pod's prefix chain ends as soon as it is missing
+// from any block key, even if other pods still have that key.
+func computeScoresFromKeyToPods(blockKeys []kvblock.BlockHash,
```
vMaroon: Wonder if we can reuse the `kvblock.Scorer` implementation directly here.

pkg/plugins/scorer/precise_prefix_cache.go
```diff
@@ -179,8 +321,99 @@ func (s *PrecisePrefixCacheScorer) Category() scheduling.ScorerCategory {
...
+	}
+
+	// 4. Compute per-pod scores (longest prefix match)
+	scores := computeScoresFromKeyToPods(blockKeys, keyToPods)
+
+	// 4b. Compute confirmed-only scores (excluding speculative entries).
+	// This is used by the no-hit-lru-scorer to correctly identify cold requests:
+	// when only speculative entries exist, confirmedScores will be empty,
+	// causing PrefixCacheServers to be empty, so isColdRequest returns true.
+	confirmedScores := computeConfirmedScoresFromKeyToPods(blockKeys, keyToPods)
```
vMaroon: While I get the logic, this would make the `no-hit-lru-scorer` contradict a speculative entry. If this is needed, then the `no-hit-lru-scorer` would not work with the approximate `prefix-cache-scorer` at all. 

The point of speculative indexing is consolidation: steer same-prefix requests to fewer pods during the blind spot. If `no-hit-lru-scorer` keeps distributing despite the speculative entries, it partially counteracts that consolidation.

I think this step, and all its supporting logic should be removed. What do you think?

pkg/plugins/scorer/precise_prefix_cache.go
bongwoobak: Do you mean using LongestPrefixScorer directly, or exposing the scorer from Indexer? Either way it needs some interface work. How about to address it in follow-up PR?

pkg/plugins/scorer/precise_prefix_cache.go
bongwoobak: Agree. The confirmedScores split was added to keep isColdRequest() working for LRU distribution, but you're right that it contradicts the consolidation goal. Removed confirmedScores and all supporting logic.

docs/configuration.md
```diff
@@ -0,0 +1,134 @@
+# Speculative Indexing Configuration
```
vMaroon: A separate documentation `configuration.md` file just for this might be confusing, maybe better update the `precise-prefix-cache-scorer` section in the `architecture.md` file, or name/place differently.

docs/configuration.md
```diff
@@ -0,0 +1,134 @@
...
+featureGates:
+  - prepareDataPlugins                    # required for speculative indexing
+plugins:
+  - type: precise-prefix-cache-scorer
+    parameters:
+      tokenProcessorConfig:
+        blockSize: 32                     # must match vLLM --block-size
+        hashSeed: "12345"
+      speculativeIndexing: true           # enable speculative indexing
+      speculativeTTL: 2000000000          # 2s in nanoseconds
```
vMaroon: Can we make this human-readable?

pkg/plugins/scorer/precise_prefix_cache.go
vMaroon: yes, my bad.

PR #664, opened 2026-03-01
fix(makefile): use shell variable for kv-cache path in tokenizer build

### LINKED ISSUE #663: make image-build-uds-tokenizer failed
**What happened**:

Clone the code and make image-build-uds-tokenizer

```console
$ make image-build-uds-tokenizer
  go: downloading go1.25.7 (darwin/arm64)
  Container tool 'docker' found.
  ==== Building UDS Tokenizer image ghcr.io/llm-d/llm-d-uds-tokenizer:dev ====
  kv-cache module not found, downloading Go modules...
  docker build \
          --platform linux/arm64 \
          -t ghcr.io/llm-d/llm-d-uds-tokenizer:dev \
          -f /services/uds_tokenizer/Dockerfile \
          /services/uds_tokenizer
  [+] Building 0.0s (0/0)                                                    docker:desktop-linux
  ERROR: failed to build: unable to prepare context: path "/services/uds_tokenizer" not found
  make: *** [image-build-uds-tokenizer] Error 1
```

**Environment**:
- Install tools:
- Others:

gyliu513:
Fixed #663 

`$(KV_CACHE_PATH)` is a Make variable evaluated during parse phase. When the kv-cache module wasn't present at parse time, the variable remained empty even after `go mod download` executed in the recipe, resulting in invalid paths like `/services/uds_tokenizer`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -190,12 +190,12 @@ image-build-uds-tokenizer: check-container-tool ## Build UDS tokenizer image fro
 	if [ -z "$$KV_CACHE_PATH_CHECK" ]; then \
 		echo "Error: Could not find kv-cache module even after download."; \
 		exit 1; \
-	fi
+	fi; \
 	$(CONTAINER_RUNTIME) build \
 		--platform linux/$(TARGETARCH) \
 		-t $(UDS_TOKENIZER_IMAGE) \
-		-f $(KV_CACHE_PATH)/services/uds_tokenizer/Dockerfile \
-		$(KV_CACHE_PATH)/services/uds_tokenizer
+		-f $$KV_CACHE_PATH_CHECK/services/uds_tokenizer/Dockerfile \
+		$$KV_CACHE_PATH_CHECK/services/uds_tokenizer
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -190,12 +190,12 @@ image-build-uds-tokenizer: check-container-tool ## Build UDS tokenizer image fro
 	if [ -z "$$KV_CACHE_PATH_CHECK" ]; then \
```
elevran, line 190: who's responsible for setting this? 
If the user, provide a hint that `KV_CACHE_PATH_CHECK` (or similar) should be set.
If by automation, ensure it is set once ut has run.

While we can consider a short term workaround, long term these should not be in inference-scheduler at all.

Makefile
gyliu513, line 189: @elevran KV_CACHE_PATH_CHECK is set here, am I missing anything here? It is some existing logic.

elevran: no you are not missing anything. I was :-)

zdtsw:
thanks for the fix @gyliu513 !

PR #669, opened 2026-03-02
Add idle pod config to active-request-scorer

dagrayvid:
This PR adds features to the active-requests-scorer to enable us to prefer idle Pods in the scoring. 

The two features, idleThreshold and maxBusyScore parameters create a scoring gap between idle and busy pods, helping distribute the prefix caches during cold starts / warmup
- idleThreshold: max requests to be considered idle (default: 0)
- maxBusyScore: max score for busy pods (default: 1.0 for backwards compat)

Examples:
- Binary mode: idleThreshold=0, maxBusyScore=0 (idle=1.0, busy=0.0)
- Hybrid mode: idleThreshold=0, maxBusyScore=0.5 (idle=1.0, busy=0-0.5)
- Flexible: idleThreshold=1, maxBusyScore=0.5 (≤1 req = idle)

This solves an issue with the default configuration of the precise-prefix-cache scorer guide.

**Explanation of issue:**
When using the default settings with the precise-prefix-cache guide, if we benchmark with a dataset which has 1 large common prefix, the prefix cache scorer dominates routing decisions, and can result in all requests going to 1 Pod. With the current scores in the guide, requests won't be routed to another Pod until the load is high enough to cause queueing (num_requests_waiting > 0). 

**Performance tests:**
_**Note: Originally this PR added an entirely new scorer "idle-pod-scorer" which behaved the same way as active-requests-scorer in the Binary mode described above. The performance results below are with that idle-pod-scorer.**_ 

In the case mentioned in the issue, where all requests share a common prefix, you can see the issue by looking at the num_requests_running.

With upstream defaults:
with-upstream-defaults-num-requests-running

With idle-pod-scorer with weight 3:
with-idle-pod-scorer-3-num-requests-running

And the performance results reflect this difference, in particular in ITL and request latency. This benchmark was with consstant RPS 2, 4, and 6 with the following dataset settings: `prefix_tokens=5000,prompt_tokens=512,prompt_tokens_stdev=250,output_tokens=1000,output_tokens_stdev=500`
Interactivity vs throughput:
newplot - 2026-03-02T072338 565

p95 TTFT:
newplot - 2026-03-02T072402 269

From my testing, the addition of the idle-pod-scorer does not seem to impact performance in most other cases such as multi-turn chat with no common prefix / system prompt, or cases with 0 prefix cache hits.


ahg-g:
> I did experiment with using the active requests scorer with weights 2 and 3. With 2 we start routing requests to other Pods once kv cache utilization is ~50% on the one busy Pod, and with 3 we would start as soon as there are idle Pods. The problem with this is that it impacts routing decisions at other times, and from my performance tests it seemed to make TTFT slightly worse, in particular p95 and p99. I believe this is because the active-requests, kv-cache, and queue scorers together can outweigh the prefix cache scorer when it doesn't make sense to do so.

Was the test done using all three load scorers, kv-cache, queue and active-requests scorers? or using active-requests scorer only (along with the prefix-cache one)?

dagrayvid:
> > I did experiment with using the active requests scorer with weights 2 and 3. ... The problem with this is that it impacts routing decisions at other times, and from my performance tests it seemed to make TTFT slightly worse, in particular p95 and p99. ...
> 
> Was the test done using all three load scorers, kv-cache, queue and active-requests scorers? or using active-requests scorer only (along with the prefix-cache one)?

That testing was done with all 4 scorers:
```
            - pluginRef: precise-prefix-cache-scorer
              weight: 3.0
              weight: 2.0
            - pluginRef: queue-scorer
              weight: 2.0
            - pluginRef: active-request-scorer
              weight: 2.0
            - pluginRef: max-score-picker
```

Results were mixed across experiments with this config, so I wouldn't suggest it as a solution to the underlying issue. The idle-pod-scorer seems to be a safer solution which solves the underlying issue without making performance worse in any other cases. 

Here are some example results of a guidellm test simulating multi-turn chat (**without** a common shared system prompt)
```
turns=5,turns_stdev=2 # each multi-turn conversation goes for this many steps
prompt_tokens=500,prompt_tokens_stdev=250, # plus the previous prompts + outputs in the conversation 
output_tokens=500,output_tokens_stdev=500,
```
p95-ttft-with-active-requests-scorer-added

dagrayvid:
I was able to repeat the experiment with these customizations to active-request-scorer instead of an entirely new scorer. The results are right on par with the idle-pod-scorer originally proposed

newplot - 2026-03-03T162130 015

ahg-g:
thanks, what is the full scorers config and weights used?

pkg/plugins/scorer/active_request.go
elevran, line 31: unrelated to this PR and not to be included in it.
If the main expectation is that Plugins are created via their factory method, this can be kept as unexported.
Similarly, the use of `scorer.NewActiveRequest` can be confined to local pacakge and its tests.

PR #682, opened 2026-03-06
feat(sidecar): simplify TLS command line options with StringSlice flags

### LINKED ISSUE #635: Simplify sidecar proxy TLS command line options
Currently the sidecar has multiple boolean flags to enable TLS, certification verification, etc, per each stage (`prefill`, `decode`).
Consider replacing with a (e.g.,) string slice flags that accepts the stage code as parameter (e.g., `--enable-tls <stage>`).
Can use `pflag` library with StringSlice flag type to allow multiple options.

**Why is this needed**:
Simpler command line.

gyliu513:
fixed #635 

@elevran

### DIFF
--- a/cmd/pd-sidecar/main.go
+++ b/cmd/pd-sidecar/main.go
@@ -47,9 +47,9 @@ var (
 	}
 
 	// supportedTLSStages defines all valid stages for TLS configuration
-	supportedTLSStages = map[string]bool{
-		prefillStage: true,
-		decodeStage:  true,
+	supportedTLSStages = map[string]struct{}{
+		prefillStage: {},
+		decodeStage:  {},
 	}
 )
 
@@ -79,6 +79,12 @@ func main() {
 	connector := pflag.String("connector", proxy.ConnectorNIXLV2, "the P/D connector being used. Supported: "+strings.Join(supportedConnectors, ", "))
 	enableTLS := pflag.StringSlice("enable-tls", []string{}, "stages to enable TLS for. Supported: "+strings.Join(supportedTLSStagesNames(), ", ")+". Can be specified multiple times or as comma-separated values.")
 	tlsInsecureSkipVerify := pflag.StringSlice("tls-insecure-skip-verify", []string{}, "stages to skip TLS verification for. Supported: "+strings.Join(supportedTLSStagesNames(), ", ")+". Can be specified multiple times or as comma-separated values.")
+
+	// Deprecated flags - kept for backward compatibility, will be removed in a future release
+	prefillerUseTLS := pflag.Bool("prefiller-use-tls", false, "Deprecated: use --enable-tls=prefiller instead. Whether to use TLS when sending requests to prefillers.")
+	decoderUseTLS := pflag.Bool("decoder-use-tls", false, "Deprecated: use --enable-tls=decoder instead. Whether to use TLS when sending requests to the decoder.")
+	prefillerInsecureSkipVerify := pflag.Bool("prefiller-tls-insecure-skip-verify", false, "Deprecated: use --tls-insecure-skip-verify=prefiller instead. Skip TLS verification for requests to prefiller.")
+	decoderInsecureSkipVerify := pflag.Bool("decoder-tls-insecure-skip-verify", false, "Deprecated: use --tls-insecure-skip-verify=decoder instead. Skip TLS verification for requests to decoder.")
 	secureProxy := pflag.Bool("secure-proxy", true, "Enables secure proxy. Defaults to true.")
 	certPath := pflag.String(
 		"cert-path", "", "The path to the certificate for secure proxy. The certificate and private key files "+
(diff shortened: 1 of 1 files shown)


cmd/pd-sidecar/main.go
```diff
@@ -31,38 +32,61 @@ import (
 	"github.com/llm-d/llm-d-inference-scheduler/pkg/telemetry"
 )
 
+const (
+	// TLS stages
+	stagePrefiller = "prefiller"
+	stageDecoder   = "decoder"
```
elevran: Nit: original options used prefiller and decoder, perhaps prefer

```suggestion
	prefillStage = "prefill"
	decodDStage   = "decode"
```

cmd/pd-sidecar/main.go
```diff
@@ -31,38 +32,61 @@ import (
...
-	inferencePoolName := flag.String("inference-pool-name", os.Getenv("INFERENCE_POOL_NAME"), "the specific InferencePool name to watch (defaults to INFERENCE_POOL_NAME env var)")
-	enablePrefillerSampling := flag.Bool("enable-prefiller-sampling", func() bool { b, _ := strconv.ParseBool(os.Getenv("ENABLE_PREFILLER_SAMPLING")); return b }(), "if true, the target prefill instance will be selected randomly from among the provided prefill host values")
-	poolGroup := flag.String("pool-group", proxy.DefaultPoolGroup, "group of the InferencePool this Endpoint Picker is associated with.")
+	enableSSRFProtection := pflag.Bool("enable-ssrf-protection", false, "enable SSRF protection using InferencePool allowlisting")
+	inferencePoolNamespace := pflag.String("inference-pool-namespace", os.Getenv("INFERENCE_POOL_NAMESPACE"), "the Kubernetes namespace to watch for InferencePool resources (defaults to INFERENCE_POOL_NAMESPACE env var)")
+	inferencePoolName := pflag.String("inference-pool-name", os.Getenv("INFERENCE_POOL_NAME"), "the specific InferencePool name to watch (defaults to INFERENCE_POOL_NAME env var)")
```
elevran, line 95: nit: unrelated to this PR but can be simplified to use standard namespced-name syntax (`ns/name`)

cmd/pd-sidecar/main.go
```diff
@@ -31,38 +32,61 @@ import (
...
+	enablePrefillerSampling := pflag.Bool("enable-prefiller-sampling", func() bool { b, _ := strconv.ParseBool(os.Getenv("ENABLE_PREFILLER_SAMPLING")); return b }(), "if true, the target prefill instance will be selected randomly from among the provided prefill host values")
+	poolGroup := pflag.String("pool-group", proxy.DefaultPoolGroup, "group of the InferencePool this Endpoint Picker is associated with.")
 
 	opts := zap.Options{}
 	opts.BindFlags(flag.CommandLine) // optional to allow zap logging control via CLI
-	flag.Parse()
+
+	// Add Go flags to pflag (for zap options compatibility)
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
```
elevran, line 103: confirming: is this not switched to `pflag` due to the `zap.Options.BindFlags()`?

cmd/pd-sidecar/main.go
```diff
@@ -100,6 +124,35 @@ func main() {
 	}
 	logger.Info("p/d connector validated", "connector", connector)
 
+	// Validate TLS stages
```
elevran, line 168: seems like the below checks could be simplified using a map of the allowed values for`supportedTLSStages` instead of an array?

cmd/pd-sidecar/main.go
```diff
@@ -31,38 +32,61 @@ import (
...
 var (
 	// supportedConnectors defines all valid P/D connector types
 	supportedConnectors = []string{
 		proxy.ConnectorNIXLV2,
 		proxy.ConnectorSharedStorage,
 		proxy.ConnectorSGLang,
 	}
+
+	// supportedTLSStages defines all valid stages for TLS configuration
+	supportedTLSStages = []string{
```
elevran: consier using a map for later checks (e.g., contains, isvalid, ...)  instead of slice iteration.

cmd/pd-sidecar/main.go
gyliu513: Done

cmd/pd-sidecar/main.go
gyliu513: Done

cmd/pd-sidecar/main.go
gyliu513: Opened a follow up issue at https://github.com/llm-d/llm-d-inference-scheduler/issues/684

cmd/pd-sidecar/main.go
gyliu513: Yes

cmd/pd-sidecar/main.go
gyliu513: Done

cmd/pd-sidecar/main.go
```diff
@@ -31,38 +32,70 @@ import (
...
 var (
 	// supportedConnectors defines all valid P/D connector types
 	supportedConnectors = []string{
 		proxy.ConnectorNIXLV2,
 		proxy.ConnectorSharedStorage,
 		proxy.ConnectorSGLang,
 	}
+
+	// supportedTLSStages defines all valid stages for TLS configuration
+	supportedTLSStages = map[string]bool{
```
elevran: nit/FYI: can use `map[string]struct{}` (or use `k8s.io/apimachinery/pkg/util/sets` if you prefer a higher level API)

cmd/pd-sidecar/main.go
```diff
@@ -31,38 +32,70 @@ import (
...
-	decoderUseTLS := flag.Bool("decoder-use-tls", false, "whether to use TLS when sending requests to the decoder")
-	prefillerInsecureSkipVerify := flag.Bool("prefiller-tls-insecure-skip-verify", false, "configures the proxy to skip TLS verification for requests to prefiller")
-	decoderInsecureSkipVerify := flag.Bool("decoder-tls-insecure-skip-verify", false, "configures the proxy to skip TLS verification for requests to decoder")
-	secureProxy := flag.Bool("secure-proxy", true, "Enables secure proxy. Defaults to true.")
-	certPath := flag.String(
+	port := pflag.String("port", "8000", "the port the sidecar is listening on")
+	vLLMPort := pflag.String("vllm-port", "8001", "the port vLLM is listening on")
+	vLLMDataParallelSize := pflag.Int("data-parallel-size", 1, "the vLLM DATA-PARALLEL-SIZE value")
+	connector := pflag.String("connector", proxy.ConnectorNIXLV2, "the P/D connector being used. Supported: "+strings.Join(supportedConnectors, ", "))
+	enableTLS := pflag.StringSlice("enable-tls", []string{}, "stages to enable TLS for. Supported: "+strings.Join(supportedTLSStagesNames(), ", ")+". Can be specified multiple times or as comma-separated values.")
```
elevran, line 80: As noted on #685 we may want to allow deprecation cycle of the existing flags instead of removing them altogether.

gyliu513: Done

elevran (commented):
/lgtm
/approve

Issue #634, opened 2026-02-19
Add configuration file for sidecar proxy


elevran:
**What would you like to be added**:
The ability to configure the sidecar proxy using a configuration file or text (e.g., YAML, JSON).

**Why is this needed**:
As the set of command line options continues to grow, conifguration file can reduce some of the that.

elevran:
resolved by #683

PR #683, opened 2026-03-06
Adds configuration file for sidecar proxy

DhritiShikhar:
### Resolves 

1. #634 

### Changes

1. Adds ability to configure sidecar proxy through YAML.
 
2. YAML configuration can be provided via command line flags `--configuration` and/or `--configuration-file`.

3. YAML configuration provided via `--configuration` is given higher priority.

4. `--configuration` enables sidecar configuration YAML to be provided as inline specification. 
Example: `--configuration={port: 8085, vllm-port: 8203, connector: sglang}`.

6. `--configuration-file` enables sidecar configuration YAML to be provided as file. 
Example: `--configuration-file=/etc/config/sidecar-config.yaml`.

```sidecarconfig.yaml
port: 8083
vllm-port: 8201
connector: "sglang"
secure-proxy: false
```

8. Configuration file is mounted as file at /etc/config/sidecar-config.yaml.

9. Sidecar configuration priority (as specified [here](https://github.com/llm-d/llm-d-inference-scheduler/pull/683/changes#r2901536829)):
    - Command-line flags (highest precedence)
    - Inline specification / Configuration file
    - Environment variables
    - Built-in defaults (lowest precedence)

### Tests

This PR contains unit tests

```release-note
Add sidecar configuration specified via either `--configuration-file <path>` or  `--configuration "<text>"`. Both YAML and JSON formats are allowed. Configuration field names are the same as the CLI flags. CLI has priority over the configuration file or text.
```

### DIFF
--- a/pkg/sidecar/proxy/options.go
+++ b/pkg/sidecar/proxy/options.go
@@ -33,7 +33,77 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	logutil "github.com/llm-d/llm-d-inference-scheduler/pkg/common/observability/logging"
+	"sigs.k8s.io/yaml"
 )
+
+const (
+	// Flags
+	port                    = "port"
+	vllmPort                = "vllm-port"
+	dataParallelSize        = "data-parallel-size"
+	kvConnector             = "kv-connector"
+	ecConnector             = "ec-connector"
+	enableSSRFProtection    = "enable-ssrf-protection"
+	enablePrefillerSampling = "enable-prefiller-sampling"
+	enableTLS               = "enable-tls"
+	tlsInsecureSkipVerify   = "tls-insecure-skip-verify"
+	secureServing           = "secure-proxy"
+	certPath                = "cert-path"
+	inferencePool           = "inference-pool"
+	poolGroup               = "pool-group"
+	maxIdleConnsPerHost     = "max-idle-conns-per-host"
+	inlineConfiguration     = "configuration"
+	configurationFile       = "configuration-file"
+
+	// Deprecated flags
+	connector                      = "connector"
+	prefillerUseTLS                = "prefiller-use-tls"
+	decoderUseTLS                  = "decoder-use-tls"
+	encoderUseTLS                  = "UseTLSForEncoder"
+	prefillerTLSInsecureSkipVerify = "prefiller-tls-insecure-skip-verify"
+	decoderTLSInsecureSkipVerify   = "decoder-tls-insecure-skip-verify"
+	inferencePoolNamespace         = "inference-pool-namespace"
+	inferencePoolName              = "inference-pool-name"
+
+	// Environment variables
+	envInferencePool           = "INFERENCE_POOL"
+	envInferencePoolNamespace  = "INFERENCE_POOL_NAMESPACE"
+	envInferencePoolName       = "INFERENCE_POOL_NAME"
+	envEnablePrefillerSampling = "ENABLE_PREFILLER_SAMPLING"
+
+	// Defaults
+	defaultPort             = "8000"
+	defaultVLLMPort         = "8001"
+	defaultDataParallelSize = 1
+
+	// TLS stages
+	prefillStage = "prefiller"
+	decodeStage  = "decoder"
+	encodeStage  = "encoder"
+)
+
+// yamlConfiguration represents structure of YAML configuration for sidecar proxy
+type yamlConfiguration struct {
+	Port                           int      `json:"port,omitempty"`
+	VLLMPort                       int      `json:"vllm-port,omitempty"`
+	DataParallelSize               int      `json:"data-parallel-size,omitempty"`
+	KVConnector                    string   `json:"kv-connector,omitempty"`
+	Connector                      string   `json:"connector,omitempty"`
+	ECConnector                    string   `json:"ec-connector,omitempty"`
+	EnableSSRFProtection           *bool    `json:"enable-ssrf-protection,omitempty"`
+	EnablePrefillerSampling        *bool    `json:"enable-prefiller-sampling,omitempty"`
+	SecureServing                  *bool    `json:"secure-proxy,omitempty"`
+	CertPath                       string   `json:"cert-path,omitempty"`
+	EnableTLS                      []string `json:"enable-tls,omitempty"`
+	TLSInsecureSkipVerify          []string `json:"tls-insecure-skip-verify,omitempty"`
+	PrefillerUseTLS                *bool    `json:"prefiller-use-tls,omitempty"`
+	DecoderUseTLS                  *bool    `json:"decoder-use-tls,omitempty"`
+	PrefillerTLSInsecureSkipVerify *bool    `json:"prefiller-tls-insecure-skip-verify,omitempty"`
+	DecoderTLSInsecureSkipVerify   *bool    `json:"decoder-tls-insecure-skip-verify,omitempty"`
+	InferencePool                  string   `json:"inference-pool,omitempty"`
+	PoolGroup                      string   `json:"pool-group,omitempty"`
+	MaxIdleConnsPerHost            int      `json:"max-idle-conns-per-host,omitempty"`
+}
 
 // Options holds the CLI-facing configuration for the pd-sidecar proxy.
 // It embeds Config which represents the complete processed runtime configuration.
@@ -60,15 +130,11 @@ type Options struct {
 	prefillerInsecureSkipVerify bool   // Deprecated: use --tls-insecure-skip-verify=prefiller instead
 	decoderInsecureSkipVerify   bool   // Deprecated: use --tls-insecure-skip-verify=decoder instead
 
-	loggingOptions zap.Options // loggingOptions holds the zap logging configuration
+	loggingOptions      zap.Options // loggingOptions holds the zap logging configuration
+	pflagSet            *pflag.FlagSet
+	inlineConfiguration string
+	fileConfiguration   string
 }
-
-const (
-	// TLS stages
-	prefillStage = "prefiller"
-	decodeStage  = "decoder"
-	encodeStage  = "encoder"
-)
 
 var (
 	// supportedKVConnectors defines all valid P/D KV connector types
@@ -98,23 +164,23 @@ var (
 // NewOptions returns a new Options struct initialized with default values.
 func NewOptions() *Options {
 	enablePrefillerSampling := false
-	if val, err := strconv.ParseBool(os.Getenv("ENABLE_PREFILLER_SAMPLING")); err == nil {
+	if val, err := strconv.ParseBool(os.Getenv(envEnablePrefillerSampling)); err == nil {
 		enablePrefillerSampling = val
 	}
 
 	return &Options{
 		Config: Config{
-			Port:                    "8000",
-			DataParallelSize:        1,
+			Port:                    defaultPort,
+			DataParallelSize:        defaultDataParallelSize,
 			SecureServing:           true,
 			EnablePrefillerSampling: enablePrefillerSampling,
 			MaxIdleConnsPerHost:     defaultMaxIdleConnsPerHost,
 			PoolGroup:               DefaultPoolGroup,
-			InferencePoolNamespace:  os.Getenv("INFERENCE_POOL_NAMESPACE"),
-			InferencePoolName:       os.Getenv("INFERENCE_POOL_NAME"),
+			InferencePoolNamespace:  os.Getenv(envInferencePoolNamespace),
+			InferencePoolName:       os.Getenv(envInferencePoolName),
 		},
-		vllmPort:      "8001",
-		inferencePool: os.Getenv("INFERENCE_POOL"),
+		vllmPort:      defaultVLLMPort,
+		inferencePool: os.Getenv(envInferencePool),
 		connector:     KVConnectorNIXLV2,
 	}
 }
@@ -122,47 +188,51 @@ func NewOptions() *Options {
 // AddFlags binds the Options fields to command-line flags on the given FlagSet.
 // It also sets up zap logging flags and integrates Go flags with pflag.
 func (opts *Options) AddFlags(fs *pflag.FlagSet) {
+	if opts.pflagSet == nil {
+		opts.pflagSet = fs
+	}
+	goFlagSet := flag.NewFlagSet("goFlagSet", flag.ContinueOnError)
 	// Add logging flags to the standard flag set
-	opts.loggingOptions.BindFlags(flag.CommandLine)
-
+	opts.loggingOptions.BindFlags(goFlagSet)
 	// Add Go flags to pflag (for zap options compatibility)
-	fs.AddGoFlagSet(flag.CommandLine)
-
-	fs.StringVar(&opts.Port, "port", opts.Port, "the port the sidecar is listening on")
-	fs.StringVar(&opts.vllmPort, "vllm-port", opts.vllmPort, "the port vLLM is listening on")
-	fs.IntVar(&opts.DataParallelSize, "data-parallel-size", opts.DataParallelSize, "the vLLM DATA-PARALLEL-SIZE value")
-	fs.StringVar(&opts.KVConnector, "kv-connector", opts.KVConnector,
+	fs.AddGoFlagSet(goFlagSet)
+	fs.StringVar(&opts.Port, port, opts.Port, "the port the sidecar is listening on")
+	fs.StringVar(&opts.vllmPort, vllmPort, opts.vllmPort, "the port vLLM is listening on")
+	fs.IntVar(&opts.DataParallelSize, dataParallelSize, opts.DataParallelSize, "the vLLM DATA-PARALLEL-SIZE value")
+	fs.StringVar(&opts.KVConnector, kvConnector, opts.KVConnector,
 		"the KV protocol between prefiller and decoder. Supported: "+supportedKVConnectorNamesStr)
-	fs.StringVar(&opts.ECConnector, "ec-connector", opts.ECConnector,
+	fs.StringVar(&opts.ECConnector, ecConnector, opts.ECConnector,
 		"the EC protocol between encoder and prefiller (for EPD mode). Supported: "+supportedECConnectorNamesStr+". Leave empty to skip encoder stage.")
-	fs.BoolVar(&opts.SecureServing, "secure-proxy", opts.SecureServing, "Enables secure proxy. Defaults to true.")
-	fs.StringVar(&opts.CertPath, "cert-path", opts.CertPath, "The path to the certificate for secure proxy. The certificate and private key files are assumed to be named tls.crt and tls.key, respectively. If not set, and secureProxy is enabled, then a self-signed certificate is used (for testing).")
-	fs.BoolVar(&opts.EnableSSRFProtection, "enable-ssrf-protection", opts.EnableSSRFProtection, "enable SSRF protection using InferencePool allowlisting")
-	fs.BoolVar(&opts.EnablePrefillerSampling, "enable-prefiller-sampling", opts.EnablePrefillerSampling, "if true, the target prefill instance will be selected randomly from among the provided prefill host values")
-	fs.StringVar(&opts.PoolGroup, "pool-group", opts.PoolGroup, "group of the InferencePool this Endpoint Picker is associated with.")
+	fs.BoolVar(&opts.SecureServing, secureServing, opts.SecureServing, "Enables secure proxy. Defaults to true.")
+	fs.StringVar(&opts.CertPath, certPath, opts.CertPath, "The path to the certificate for secure proxy. The certificate and private key files are assumed to be named tls.crt and tls.key, respectively. If not set, and secureProxy is enabled, then a self-signed certificate is used (for testing).")
+	fs.BoolVar(&opts.EnableSSRFProtection, enableSSRFProtection, opts.EnableSSRFProtection, "enable SSRF protection using InferencePool allowlisting")
+	fs.BoolVar(&opts.EnablePrefillerSampling, enablePrefillerSampling, opts.EnablePrefillerSampling, "if true, the target prefill instance will be selected randomly from among the provided prefill host values")
+	fs.StringVar(&opts.PoolGroup, poolGroup, opts.PoolGroup, "group of the InferencePool this Endpoint Picker is associated with.")
 
-	fs.StringSliceVar(&opts.enableTLS, "enable-tls", opts.enableTLS, "stages to enable TLS for. Supported: "+supportedTLSStageNamesStr+". Can be specified multiple times or as comma-separated values.")
-	fs.StringSliceVar(&opts.tlsInsecureSkipVerify, "tls-insecure-skip-verify", opts.tlsInsecureSkipVerify, "stages to skip TLS verification for. Supported: "+supportedTLSStageNamesStr+". Can be specified multiple times or as comma-separated values.")
-	fs.StringVar(&opts.inferencePool, "inference-pool", opts.inferencePool, "InferencePool in namespace/name or name format (e.g., default/my-pool or my-pool). A single name implies the 'default' namespace. Can also use INFERENCE_POOL env var.")
+	fs.StringSliceVar(&opts.enableTLS, enableTLS, opts.enableTLS, "stages to enable TLS for. Supported: "+supportedTLSStageNamesStr+". Can be specified multiple times or as comma-separated values.")
+	fs.StringSliceVar(&opts.tlsInsecureSkipVerify, tlsInsecureSkipVerify, opts.tlsInsecureSkipVerify, "stages to skip TLS verification for. Supported: "+supportedTLSStageNamesStr+". Can be specified multiple times or as comma-separated values.")
+	fs.StringVar(&opts.inferencePool, inferencePool, opts.inferencePool, "InferencePool in namespace/name or name format (e.g., default/my-pool or my-pool). A single name implies the 'default' namespace. Can also use INFERENCE_POOL env var.")
 
 	// Deprecated flags - kept for backward compatibility
-	fs.StringVar(&opts.connector, "connector", opts.connector, "Deprecated: use --kv-connector instead. The P/D connector being used. Supported: "+supportedKVConnectorNamesStr)
-	_ = fs.MarkDeprecated("connector", "use --kv-connector instead")
+	fs.StringVar(&opts.connector, connector, opts.connector, "Deprecated: use --kv-connector instead. The P/D connector being used. Supported: "+supportedKVConnectorNamesStr)
+	_ = fs.MarkDeprecated(connector, "use --kv-connector instead")
 
-	fs.BoolVar(&opts.prefillerUseTLS, "prefiller-use-tls", opts.prefillerUseTLS, "Deprecated: use --enable-tls=prefiller instead. Whether to use TLS when sending requests to prefillers.")
-	_ = fs.MarkDeprecated("prefiller-use-tls", "use --enable-tls=prefiller instead")
+	fs.BoolVar(&opts.prefillerUseTLS, prefillerUseTLS, opts.prefillerUseTLS, "Deprecated: use --enable-tls=prefiller instead. Whether to use TLS when sending requests to prefillers.")
+	_ = fs.MarkDeprecated(prefillerUseTLS, "use --enable-tls=prefiller instead")
 	fs.BoolVar(&opts.decoderUseTLS, "decoder-use-tls", opts.decoderUseTLS, "Deprecated: use --enable-tls=decoder instead. Whether to use TLS when sending requests to the decoder.")
-	_ = fs.MarkDeprecated("decoder-use-tls", "use --enable-tls=decoder instead")
-	fs.BoolVar(&opts.prefillerInsecureSkipVerify, "prefiller-tls-insecure-skip-verify", opts.prefillerInsecureSkipVerify, "Deprecated: use --tls-insecure-skip-verify=prefiller instead. Skip TLS verification for requests to prefiller.")
-	_ = fs.MarkDeprecated("prefiller-tls-insecure-skip-verify", "use --tls-insecure-skip-verify=prefiller instead")
-	fs.BoolVar(&opts.decoderInsecureSkipVerify, "decoder-tls-insecure-skip-verify", opts.decoderInsecureSkipVerify, "Deprecated: use --tls-insecure-skip-verify=decoder instead. Skip TLS verification for requests to decoder.")
-	_ = fs.MarkDeprecated("decoder-tls-insecure-skip-verify", "use --tls-insecure-skip-verify=decoder instead")
+	_ = fs.MarkDeprecated(decoderUseTLS, "use --enable-tls=decoder instead")
+	fs.BoolVar(&opts.prefillerInsecureSkipVerify, prefillerTLSInsecureSkipVerify, opts.prefillerInsecureSkipVerify, "Deprecated: use --tls-insecure-skip-verify=prefiller instead. Skip TLS verification for requests to prefiller.")
+	_ = fs.MarkDeprecated(prefillerTLSInsecureSkipVerify, "use --tls-insecure-skip-verify=prefiller instead")
+	fs.BoolVar(&opts.decoderInsecureSkipVerify, decoderTLSInsecureSkipVerify, opts.decoderInsecureSkipVerify, "Deprecated: use --tls-insecure-skip-verify=decoder instead. Skip TLS verification for requests to decoder.")
+	_ = fs.MarkDeprecated(decoderTLSInsecureSkipVerify, "use --tls-insecure-skip-verify=decoder instead")
 
-	fs.StringVar(&opts.InferencePoolNamespace, "inference-pool-namespace", opts.InferencePoolNamespace, "Deprecated: use --inference-pool instead. The Kubernetes namespace for the InferencePool (defaults to INFERENCE_POOL_NAMESPACE env var)")
-	_ = fs.MarkDeprecated("inference-pool-namespace", "use --inference-pool instead")
-	fs.StringVar(&opts.InferencePoolName, "inference-pool-name", opts.InferencePoolName, "Deprecated: use --inference-pool instead. The specific InferencePool name (defaults to INFERENCE_POOL_NAME env var)")
-	_ = fs.MarkDeprecated("inference-pool-name", "use --inference-pool instead")
+	fs.StringVar(&opts.InferencePoolNamespace, inferencePoolNamespace, opts.InferencePoolNamespace, "Deprecated: use --inference-pool instead. The Kubernetes namespace for the InferencePool (defaults to INFERENCE_POOL_NAMESPACE env var)")
+	_ = fs.MarkDeprecated(inferencePoolNamespace, "use --inference-pool instead")
+	fs.StringVar(&opts.InferencePoolName, inferencePoolName, opts.InferencePoolName, "Deprecated: use --inference-pool instead. The specific InferencePool name (defaults to INFERENCE_POOL_NAME env var)")
+	_ = fs.MarkDeprecated(inferencePoolName, "use --inference-pool instead")
 	fs.IntVar(&opts.MaxIdleConnsPerHost, "max-idle-conns-per-host", opts.MaxIdleConnsPerHost, "max idle keep-alive connections per host for reverse proxy transports; set to at least the expected concurrency")
+	fs.StringVar(&opts.inlineConfiguration, inlineConfiguration, "", "Sidecar configuration in YAML provided as inline specification. Example `--configuration={port: 8085, vllm-port: 8203}. Inline configuration and file configuration are mutually exclusive.`")
+	fs.StringVar(&opts.fileConfiguration, configurationFile, "", "Path to file which contains sidecar configuration in YAML. Example `--configuration-file=/etc/config/sidecar-config.yaml`. Inline configuration and file configuration are mutually exclusive.")
 }
 
 // validateStages checks if all stages in the slice are valid according to the supportedStages map
@@ -176,10 +246,14 @@ func validateStages(stages []string, supportedStages map[string]struct{}, flagNa
 }
 
 // Complete performs post-processing of parsed command-line arguments.
-// It handles migration from deprecated flags, parses the InferencePool field,
-// computes boolean TLS fields, and builds Config.DecoderURL.
+// It extracts YAML configuration (if provided), handles migration from deprecated flags,
+// parses the InferencePool field, computes boolean TLS fields, and builds Config.DecoderURL.
 // After Complete(), opts.Config is fully populated.
 func (opts *Options) Complete() error {
+	if err := opts.extractYAMLConfiguration(); err != nil {
+		return err
+	}
+
 	// Migrate deprecated connector flag to KVConnector
 	if opts.connector != "" && opts.KVConnector == "" {
 		opts.KVConnector = opts.connector
@@ -322,4 +396,124 @@ func (opts *Options) NewLogger() logr.Logger {
 		zap.UseFlagOptions(&opts.loggingOptions),
 		zap.Encoder(zapcore.NewJSONEncoder(config)),
 	)
+}
+
+// extractYAMLConfiguration extracts sidecar configuration (if provided)
+// from `--configuration` and `--configuration-file` parameters
+func (opts *Options) extractYAMLConfiguration() error {
+	var yamlConfiguration yamlConfiguration
+	var yamlData []byte
+	var err error
+
+	switch {
+	case opts.inlineConfiguration != "" && opts.fileConfiguration != "":
+		return fmt.Errorf("flags --%s and --%s are mutually exclusive", inlineConfiguration, configurationFile)
+
+	case opts.inlineConfiguration != "":
+		yamlData = []byte(opts.inlineConfiguration)
+
+	case opts.fileConfiguration != "":
+		yamlData, err = os.ReadFile(opts.fileConfiguration)
+		if err != nil {
+			return fmt.Errorf("failed to read sidecar configuration from file: %w", err)
+		}
+	}
+
+	if yamlData == nil {
+		return nil
+	}
+
+	// fail on unknown YAML fields
+	if err := yaml.UnmarshalStrict(yamlData, &yamlConfiguration); err != nil {
+		return fmt.Errorf("failed to unmarshal sidecar configuration: %w", err)
+	}
+
+	opts.mergeYAMLConfiguration(yamlConfiguration)
+	return nil
+}
+
+// mergeYAMLConfiguration merges provided yamlConfiguration into Options struct,
+// respecting precedence of command-line flags (i.e., YAML values are only applied if corresponding flag was not explicitly set by user)
+func (opts *Options) mergeYAMLConfiguration(cfg yamlConfiguration) {
+	if cfg.Port != 0 && !opts.isFlagSet(port) {
+		opts.Port = strconv.Itoa(cfg.Port)
+	}
+	if cfg.VLLMPort != 0 && !opts.isFlagSet(vllmPort) {
+		opts.vllmPort = strconv.Itoa(cfg.VLLMPort)
+	}
+	if cfg.DataParallelSize != 0 && !opts.isFlagSet(dataParallelSize) {
+		opts.DataParallelSize = cfg.DataParallelSize
+	}
+	if cfg.MaxIdleConnsPerHost != 0 && !opts.isFlagSet(maxIdleConnsPerHost) {
+		opts.MaxIdleConnsPerHost = cfg.MaxIdleConnsPerHost
+	}
+
+	if cfg.KVConnector != "" && !opts.isFlagSet(kvConnector) {
+		opts.KVConnector = cfg.KVConnector
+	}
+	if cfg.Connector != "" && !opts.isFlagSet(connector) {
+		opts.connector = cfg.Connector
+	}
+	if cfg.ECConnector != "" && !opts.isFlagSet(ecConnector) {
+		opts.ECConnector = cfg.ECConnector
+	}
+
+	if cfg.EnableSSRFProtection != nil && !opts.isFlagSet(enableSSRFProtection) {
+		opts.EnableSSRFProtection = *cfg.EnableSSRFProtection
+	}
+	if cfg.EnablePrefillerSampling != nil && !opts.isFlagSet(enablePrefillerSampling) {
+		opts.EnablePrefillerSampling = *cfg.EnablePrefillerSampling
+	}
+
+	if cfg.SecureServing != nil && !opts.isFlagSet(secureServing) {
+		opts.SecureServing = *cfg.SecureServing
+	}
+	if cfg.CertPath != "" && !opts.isFlagSet(certPath) {
+		opts.CertPath = cfg.CertPath
+	}
+
+	if len(cfg.EnableTLS) > 0 && !opts.isFlagSet(enableTLS) {
+		opts.enableTLS = cfg.EnableTLS
+	}
+	if len(cfg.TLSInsecureSkipVerify) > 0 && !opts.isFlagSet(tlsInsecureSkipVerify) {
+		opts.tlsInsecureSkipVerify = cfg.TLSInsecureSkipVerify
+	}
+
+	// update prefiller/decoder TLS settings from deprecated YAML fields if corresponding new fields are not set by user via flags
+	// (i.e., prefillerUseTLS only applies if --enable-tls and --prefiller-use-tls are not set,
+	// and decoderUseTLS only applies if --enable-tls and --decoder-use-tls are not set)
+	if cfg.PrefillerUseTLS != nil && !opts.isFlagSet(enableTLS) && !opts.isFlagSet(prefillerUseTLS) {
+		opts.prefillerUseTLS = *cfg.PrefillerUseTLS
+	}
+	if cfg.DecoderUseTLS != nil && !opts.isFlagSet(enableTLS) && !opts.isFlagSet(decoderUseTLS) {
+		opts.decoderUseTLS = *cfg.DecoderUseTLS
+	}
+
+	// update prefiller/decoder TLS insecure skip verify settings from deprecated YAML fields if corresponding new fields are not set by user via flags
+	// (i.e., prefillerTLSInsecureSkipVerify only applies if --tls-insecure-skip-verify and --prefiller-tls-insecure-skip-verify are not set,
+	// and decoderTLSInsecureSkipVerify only applies if --tls-insecure-skip-verify and --decoder-tls-insecure-skip-verify are not set)
+	if cfg.PrefillerTLSInsecureSkipVerify != nil && !opts.isFlagSet(prefillerTLSInsecureSkipVerify) && !opts.isFlagSet(tlsInsecureSkipVerify) {
+		opts.prefillerInsecureSkipVerify = *cfg.PrefillerTLSInsecureSkipVerify
+	}
+	if cfg.DecoderTLSInsecureSkipVerify != nil && !opts.isFlagSet(decoderTLSInsecureSkipVerify) && !opts.isFlagSet(tlsInsecureSkipVerify) {
+		opts.decoderInsecureSkipVerify = *cfg.DecoderTLSInsecureSkipVerify
+	}
+
+	if cfg.InferencePool != "" && !opts.isFlagSet(inferencePool) {
+		opts.inferencePool = cfg.InferencePool
+	}
+	if cfg.PoolGroup != "" && !opts.isFlagSet(poolGroup) {
+		opts.PoolGroup = cfg.PoolGroup
+	}
+}
+
+// isFlagSet returns true if flag was set by user
+func (opts *Options) isFlagSet(parameter string) bool {
+	if opts.pflagSet != nil {
+		flag := opts.pflagSet.Lookup(parameter)
+		if flag != nil && flag.Changed {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/sidecar/proxy/options_test.go
+++ b/pkg/sidecar/proxy/options_test.go
@@ -17,8 +17,506 @@ limitations under the License.
 package proxy
 
 import (
+	"errors"
+	"flag"
+	"fmt"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
 	"testing"
+
+	"github.com/spf13/pflag"
+	"github.com/stretchr/testify/require"
 )
+
+func writeTempYAML(t *testing.T, name, content string) string {
+	t.Helper()
+	path := filepath.Join(t.TempDir(), name)
+	require.NoError(t, os.WriteFile(path, []byte(content), 0o644))
+	return path
+}
+
+func createConfigWithValidYAML(t *testing.T) string {
+	t.Helper()
+	return writeTempYAML(t, "valid.yaml", `
+port: 8100
+vllm-port: 8200
+data-parallel-size: 5
+kv-connector: "sglang"
+connector: "nixlv2"
+ec-connector: "ec-example"
+enable-ssrf-protection: true
+enable-prefiller-sampling: true
+enable-tls:
+- prefiller
+- decoder
+prefiller-use-tls: false
+tls-insecure-skip-verify:
+- prefiller
+decoder-tls-insecure-skip-verify: true
+secure-proxy: false
+cert-path: "/etc/certificates-file"
+inference-pool: "file-ns/inference-pool-file"
+pool-group: "pool-group-file"
+max-idle-conns-per-host: 300
+`)
+}
+
+func createConfigWithUnknownKeys(t *testing.T) string {
+	t.Helper()
+	return writeTempYAML(t, "valid.yaml", `
+port: 8100
+vllm-port: 8200
+unknown-key: 1001
+`)
+}
+
+func createConfigWithInvalidYAML(t *testing.T) string {
+	t.Helper()
+	return writeTempYAML(t, "invalid.yaml", `
+port: 8100
+invalid-yaml,
+`)
+}
+
+func TestSidecarConfiguration(t *testing.T) {
+	// --- inline YAML for testing ---
+	inlineYAML :=
+		`{
+		port: 8011,
+		vllm-port: 8021,
+		data-parallel-size: 3,
+		kv-connector: sglang,
+		connector: nixlv2,
+		ec-connector: ec-example,
+		enable-ssrf-protection: true,
+		enable-prefiller-sampling: true,
+		enable-tls: ['prefiller', 'decoder'],
+		prefiller-use-tls: false,
+		decoder-use-tls: true,
+		tls-insecure-skip-verify: ['decoder'],
+		prefiller-tls-insecure-skip-verify: true,
+		secure-proxy: false,
+		cert-path: '/etc/certificates-inline',
+		inference-pool: inline-ns/inference-pool-inline,
+		pool-group: pool-group-inline,
+		max-idle-conns-per-host: 200
+	}`
+	invalidInlineYAML := "{port: 8200, invalid-yaml}"
+
+	// -- file YAML for testing ---
+	validYAMLPath := createConfigWithValidYAML(t)
+	invalidYAMLPath := createConfigWithInvalidYAML(t)
+	unknownKeysYAMLPath := createConfigWithUnknownKeys(t)
+
+	tests := []struct {
+		name          string
+		expected      func(*Options)
+		expectedError error
+		inputFlags    map[string]any
+		inputEnvVar   map[string]any
+	}{
+		{
+			name: "inline YAML overrides default",
+			inputFlags: map[string]any{
+				inlineConfiguration: &inlineYAML,
+			},
+			expected: func(o *Options) {
+				o.Port = "8011"
+				o.vllmPort = "8021"
+				o.DataParallelSize = 3
+				o.MaxIdleConnsPerHost = 200
+
+				o.KVConnector = KVConnectorSGLang
+				o.connector = KVConnectorNIXLV2
+				o.ECConnector = ECExampleConnector
+
+				o.EnableSSRFProtection = true
+				o.EnablePrefillerSampling = true
+
+				o.enableTLS = []string{prefillStage, decodeStage}
+				o.UseTLSForPrefiller = true
+				o.UseTLSForDecoder = true
+				o.UseTLSForEncoder = false
+
+				o.tlsInsecureSkipVerify = []string{prefillStage, decodeStage}
+				o.InsecureSkipVerifyForPrefiller = true
+				o.InsecureSkipVerifyForDecoder = true
+				o.InsecureSkipVerifyForEncoder = false
+
+				o.SecureServing = false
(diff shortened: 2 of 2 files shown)


cmd/pd-sidecar/main.go
```diff
@@ -29,6 +29,7 @@ import (
 	"github.com/llm-d/llm-d-inference-scheduler/pkg/sidecar/proxy"
 	"github.com/llm-d/llm-d-inference-scheduler/pkg/sidecar/version"
 	"github.com/llm-d/llm-d-inference-scheduler/pkg/telemetry"
+	"github.com/stretchr/testify/assert/yaml"
```
elevran: Please do not use a package intended for testing for the implementation.
Either define the configuration using JSON (and use the standard `encoding/json`)  or use the k8s package supporting both YAML and JSON (`sigs.k8s.io/yaml`)

cmd/pd-sidecar/main.go
```diff
@@ -40,25 +41,78 @@ var (
...
+		EnableSSRFProtection:        false,
+		InferencePoolNamespace:      os.Getenv("INFERENCE_POOL_NAMESPACE"),
+		InferencePoolName:           os.Getenv("INFERENCE_POOL_NAMESPACE"),
+		EnablePrefillerSampling:     func() bool { b, _ := strconv.ParseBool(os.Getenv("ENABLE_PREFILLER_SAMPLING")); return b }(),
+		PoolGroup:                   proxy.DefaultPoolGroup,
+	}
+}
+
 func main() {
-	port := flag.String("port", "8000", "the port the sidecar is listening on")
```
elevran: Please revert. 
Since there are deployments already using command line we should not be removing these but rather augmenting the command line with a configuration file. 
When the same setting can come from multiple sources is, the precedence order should be (from highest to lowest):
1. Command-line flags (highest precedence)
1. Environment variables
1. Configuration file
1. Built-in defaults (lowest precedence)

cmd/pd-sidecar/main.go
```diff
@@ -40,25 +41,78 @@ var (
...
-			"then a self-signed certificate is used (for testing).")
-	enableSSRFProtection := flag.Bool("enable-ssrf-protection", false, "enable SSRF protection using InferencePool allowlisting")
-	inferencePoolNamespace := flag.String("inference-pool-namespace", os.Getenv("INFERENCE_POOL_NAMESPACE"), "the Kubernetes namespace to watch for InferencePool resources (defaults to INFERENCE_POOL_NAMESPACE env var)")
-	inferencePoolName := flag.String("inference-pool-name", os.Getenv("INFERENCE_POOL_NAME"), "the specific InferencePool name to watch (defaults to INFERENCE_POOL_NAME env var)")
-	enablePrefillerSampling := flag.Bool("enable-prefiller-sampling", func() bool { b, _ := strconv.ParseBool(os.Getenv("ENABLE_PREFILLER_SAMPLING")); return b }(), "if true, the target prefill instance will be selected randomly from among the provided prefill host values")
-	poolGroup := flag.String("pool-group", proxy.DefaultPoolGroup, "group of the InferencePool this Endpoint Picker is associated with.")
+	sidecarConfig := NewSidecarConfig()
+	configFile := flag.String("config", "", "sidecar config file")
```
elevran: We may also want to support `--config-text` option to allow inline specification (files require volume mounts in k8s so it might be easier to test using inline configuration text). Not critical for this PR.

```suggestion
	configFile := flag.String("config", "", "sidecar configuration file")
```

elevran: Is there a default configuration path that makes sense?

cmd/pd-sidecar/main.go
```diff
@@ -70,6 +124,14 @@ func main() {
 	ctx := ctrl.SetupSignalHandler()
 	log.IntoContext(ctx, logger)
 
+	rawFile, err := os.ReadFile(*configFile)
```
elevran: Since there's no default specified for the flag, `os.ReadFile("")` does not make sense.
Test the flag values and log when empty (e.g., to indicate using the defaults). Do not try to read it.
If a value was specified and any error occurred (e.g., in reading or marshalling) we should abort execution and not run with the defaults (which were not intended in this case).

cmd/pd-sidecar/main.go
```diff
@@ -70,6 +124,14 @@ func main() {
...
+	if err != nil {
+		logger.Error(err, "Failed to read sidecar config file")
+	}
+	if err := yaml.Unmarshal(rawFile, sidecarConfig); err != nil {
+		logger.Error(err, "Failed to unmarshal sidecar config")
+	}
+
```
elevran: If _not already done_, please log the final configuration used (indicating if from file or default) here.

cmd/pd-sidecar/main.go
```diff
@@ -98,52 +160,52 @@ func main() {
 		logger.Info("Error: --connector must be one of: " + strings.Join(supportedConnectors, ", "))
 		return
 	}
-	logger.Info("p/d connector validated", "connector", connector)
+	logger.Info("p/d connector validated", "connector", sidecarConfig.Connector)
 
 	// Determine namespace and pool name for SSRF protection
-	if *enableSSRFProtection {
-		if *inferencePoolNamespace == "" {
+	if sidecarConfig.EnableSSRFProtection {
```
elevran: Note the overlap with #684. Will likely need a rebase of the other merges before.

cmd/pd-sidecar/main.go
```diff
@@ -40,25 +41,78 @@ var (
...
+func NewSidecarConfig() *SidecarConfig {
+	return &SidecarConfig{
+		Port:                        8000,
+		VLLMPort:                    8001,
+		VLLMDataParallelSize:        1,
+		Connector:                   proxy.ConnectorNIXLV2,
+		PrefillerUseTLS:             false,
+		DecoderUseTLS:               false,
+		PrefillerInsecureSkipVerify: false,
+		DecoderInsecureSkipVerify:   false,
```
elevran: These are being changed in #682. Would require a change here.

deploy/components/vllm-sim-pd/deployments.yaml
```diff
@@ -74,11 +74,7 @@ spec:
         image: ${SIDECAR_IMAGE}
         imagePullPolicy: IfNotPresent
         args:
-        - "--port=8000"
-        - "--vllm-port=8200"
-        - "--connector=nixlv2"
-        - "--secure-proxy=false"
-        - "--data-parallel-size=${VLLM_DATA_PARALLEL_SIZE}"
+        - "--config=/etc/config/sidecar-config.yaml"
```
elevran: Revert. Please see earlier note on adding, not replacing, command line.

elevran (commented):
Thanks for working on this @DhritiShikhar 

A couple of high level comments in addition to feedback directly on code:
- we allow deprecation of user facing API/conifguration for at least one release cycle. Thus the PR should be adding configuration support without removing command line flags. In fact, we may want to keep both as options for the future.
- YAML files are missing newline at the end of file.

cmd/pd-sidecar/main.go
DhritiShikhar: DONE.
Now, using `sigs.k8s.io/yaml`

cmd/pd-sidecar/main.go
DhritiShikhar: Thanks for letting me know. When I read the issue description, I thought it is a replacement task. I have updated my PR.

This PR considers:
1. Individual CLI flags (`--port`, `--vllm-port`) 
2. Inline YAML Configuration file `--configuration`
3. YAML Configuration file `--configuration-file`
in the precedence specified above.

I have not addressed configuration through env var in this PR.

cmd/pd-sidecar/main.go
DhritiShikhar: DONE. I have updated PR to [check for missing values](https://github.com/llm-d/llm-d-inference-scheduler/pull/683/changes#diff-896bae7ae5d80e8064ca36010b0bac5f4eec8bc492f4b56fcf18b64fff8daf7cR208).

cmd/pd-sidecar/main.go
DhritiShikhar: This PR combines values from different sources - built-in, individual flags `--port`, file `--config-file`, inline YAML `--config`. If a key is missing from one source, it picks up from a different source (if it is provided). So, the final config can be a mix of settings from different sources. 

I log the final sidecar configuration [here](https://github.com/llm-d/llm-d-inference-scheduler/pull/683/changes#diff-896bae7ae5d80e8064ca36010b0bac5f4eec8bc492f4b56fcf18b64fff8daf7cR139).

deploy/components/vllm-sim-pd/deployments.yaml
DhritiShikhar: DONE

cmd/pd-sidecar/main.go
DhritiShikhar: DONE - rebased

cmd/pd-sidecar/main.go
DhritiShikhar: DONE

pkg/sidecar/proxy/options.go
```diff
@@ -311,3 +326,275 @@ func (opts *Options) Validate() error {
...
+	}
+	if configurationFile != "" {
+		configurationMap2, err = YAMLConfigurationFromFile(configurationFile)
+		if err != nil {
+			return err
+		}
+	}
+	switch {
+	case configurationMap1 != nil && configurationMap2 != nil:
+		opts.updateSidecarConfiguration(mergeYAMLConfigurations(configurationMap2, configurationMap1))
```
elevran: All three call sites ignore the returned error. `updateSidecarConfiguration` returns an error on type-assertion failures (e.g. `port: "not-a-number"`). Curently these errors are dropped and the process continues with partial/corrupt configuration.

Each call site must propagate the error up.

pkg/sidecar/proxy/options.go
```diff
@@ -311,3 +326,275 @@ func (opts *Options) Validate() error {
...
+	if configurationMap["decoder-use-tls"] != nil {
+		if v, ok := configurationMap["decoder-use-tls"].(bool); ok {
+			if opts.isDefault("decoder-use-tls") {
+				opts.DecoderUseTLS = v
+			}
+		} else {
+			return errors.New("Type assertion failed for decoder-use-tls: " + fmt.Sprintf("%v", configurationMap["decoder-use-tls"]))
+		}
+	}
+	if configurationMap["tls-insecure-skip-verify"] != nil {
```
elevran: The flag `--tls-insecure-skip-verify` is registered as a `StringSlice` (list of stage names). Here it is treated as `bool`?
Note that the Options/Configuration structures been cleaned up considerably on main and this PR shuold be rebased.

pkg/sidecar/proxy/options.go
```diff
@@ -311,3 +326,275 @@ func (opts *Options) Validate() error {
...
+	if configurationMap["ec-connector"] != nil {
+		if v, ok := configurationMap["ec-connector"].(string); ok {
+			if opts.isDefault("ec-connector") {
+				opts.ECConnector = v
+			}
+		} else {
+			return errors.New("Type assertion failed for ec-connector: " + fmt.Sprintf("%v", configurationMap["ec-connector"]))
+		}
+	}
+	if configurationMap["enable-tls"] != nil {
```
elevran: Other field check `opts.isDefault(flagName)` before applying the YAML value. `enable-tls` unconditionally appends to `opts.EnableTLS`. If a user sets `--enable-tls=decoder` on the CLI and also has `enable-tls: [decoder, encoder]` in YAML, both are applied.

pkg/sidecar/proxy/options.go
```diff
@@ -311,3 +326,275 @@ func (opts *Options) Validate() error {
...
+	flag := o.FlagSet.Lookup(parameter)
+	return flag == nil || !flag.Changed
+}
+
+// YAMLConfigurationFromInlineSpecification extracts YAML configuration provided as inline specification
+// "--configuration={port: 8085, vllm-port: 8203}"
+func YAMLConfigurationFromInlineSpecification(config string) (map[string]any, error) {
+	var temp map[string]any
+	if err := yaml.Unmarshal([]byte(config), &temp); err != nil {
+		return nil, errors.New("Failed to unmarshal sidecar configuration")
```
elevran: Both `YAMLConfigurationFromInlineSpecification` (here) and `YAMLConfigurationFromFile` wrap errors with a fixed string, losing the underlying cause. Use `fmt.Errorf` with `%w` so the operator is aware of the underlying cause.

pkg/sidecar/proxy/options.go
```diff
@@ -311,3 +326,275 @@ func (opts *Options) Validate() error {
 
 	return nil
 }
+
+// extractYAMLConfiguration extracts sidecar configuration (if provided)
+// from `--configuration` and `--configuration-file` parameters
+func (opts *Options) extractYAMLConfiguration(configuration string, configurationFile string) error {
+	var configurationMap1, configurationMap2 configurationMap
```
elevran: Non-descriptive variable names make the merge logic hard to follow.
I think `configurationMap1` = inline spec, `configurationMap2` = file, so consider rename to `inlineConfig` / `fileConfig` to make code self-documenting.

pkg/sidecar/proxy/options_test.go
```diff
@@ -17,9 +17,404 @@ limitations under the License.
...
+var validSidecarfilePath, invalidSidecarfilePath string
+
+// createSidecarConfigurationYAMLFiles creates sidecar configuration files with valid and invalid YAML
+func createSidecarConfigurationYAMLFiles(t *testing.T) {
+	tempDir := t.TempDir()
+	validYAML := []byte(`
+port: 8083
+vllm-port: 8201
+data-parallel-size: 5
+kv-connector: "sglang"
```
elevran, line 47: Duplicate key `data-parallel-size` in test YAML fixture - behavior is undefined/misleading.
The `sigs.k8s.io/yaml` parsing converts to JSON first; last-key-wins semantics means this resolves to `6`, yet `fileDataParallelSize = 5` and case 4 asserts that value.

Remove the duplicate. If `6` was the intended value, update `fileDataParallelSize` to match.

pkg/sidecar/proxy/options_test.go
```diff
@@ -17,9 +17,404 @@ limitations under the License.
...
+	ecConnector := fmt.Sprintf("%v", ECExampleConnector)
+	secureProxyEnabled := true
+	secureProxyDisabled := false
+	ssrfProtectionEnabled := true
+	ssrfProtectionDisabled := false
+	inferencePool := "pool1"
+	fileInferencePool := "pool3"
+	enableTLSDecode := []string{decodeStage}
+	enableTLSPrefill := []string{prefillStage}
+	configuration := "{port: 8200, vllm-port: 7200, kv-connector: sglang, enableTLS: 'prefiller,decoder'}"
```
elevran: `enableTLS` (camelCase) key in inline config is silently ignored. The TLS assertions for cases 3/5/7/8 are never actually verified since `updateSidecarConfiguration` looks for the key `enable-tls` (kebab-case). The key `enableTLS` does not match, so TLS is never set from the inline config.

Further, the test assertions for `expectedUseTLSForPrefiller/Decoder/Encoder` are inside `if tt.inputFlags["enable-tls"] != nil`, which is false for cases 3/5/7/8, so those expectations are never checked.

pkg/sidecar/proxy/options_test.go
```diff
@@ -17,9 +17,404 @@ limitations under the License.
...
+			name: "case 8: case 2 + case 3 + case 4 i.e. configuration provided through individual flags + YAML inline specification + file",
+			inputFlags: map[string]any{
+				"port":               port,
+				"vllm-port":          vllmPort,
+				"configuration":      configuration,
+				"configuration-file": validSidecarfilePath,
+			},
+			expectedPort:                 port,
+			expectedVLLMPort:             vllmPort,
+			expectedDataParallelSize:     defaultDataParallelSize,
```
elevran: Case 7 (inline + file) correctly expects `fileDataParallelSize`. Case 8 adds only `port` and `vllm-port` CLI flags on top and neither affects `data-parallel-size`. Yet case 8 expects `defaultDataParallelSize (1)` instead of `fileDataParallelSize`.
This looks like a copy-paste error?

pkg/sidecar/proxy/options_test.go
```diff
@@ -17,9 +17,404 @@ limitations under the License.
...
+	"slices"
+	"strconv"
+	"strings"
 	"testing"
+
+	"github.com/spf13/pflag"
+	"github.com/stretchr/testify/require"
 )
 
+var validSidecarfilePath, invalidSidecarfilePath string
```
elevran: `createSidecarConfigurationYAMLFiles` writes package-level globals, so any two tests calling it concurrently would race. It would be safer to return the paths from the helper instead.

pkg/sidecar/proxy/options_test.go
```diff
@@ -17,9 +17,404 @@ limitations under the License.
...
+				case int:
+					require.NoError(t, testFlagSet.Set(k, strconv.Itoa(v)))
+				case float64:
+					require.NoError(t, testFlagSet.Set(k, fmt.Sprintf("%v", v)))
+				case string:
+					require.NoError(t, testFlagSet.Set(k, v))
+				case *string:
+					require.NoError(t, testFlagSet.Set(k, *v))
+				case *[]string:
+					require.NoError(t, testFlagSet.Set(k, strings.Join(*v, " ")))
```
elevran: `*[]string` flags joined with space instead of comma. The `pflag` package will treat the whole string as one value. Since pflag's `StringSlice` splits on commas, change to `strings.Join(*v, ",")`.

cmd/pd-sidecar/main.go
DhritiShikhar: This is done.

pkg/sidecar/proxy/options.go
DhritiShikhar: DONE

pkg/sidecar/proxy/options.go
DhritiShikhar: DONE

pkg/sidecar/proxy/options.go
DhritiShikhar: This is fixed. DONE

pkg/sidecar/proxy/options.go
DhritiShikhar: This is fixed. DONE

If a user sets `--enable-tls=decoder` on the CLI and also has `enable-tls: [decoder, encoder]` in YAML, both are NOT applied.

Individual flag is given precedence, then inline YAML specification, then file YAML.

pkg/sidecar/proxy/options.go
DhritiShikhar: DONE

pkg/sidecar/proxy/options.go
DhritiShikhar: DONE

pkg/sidecar/proxy/options_test.go
DhritiShikhar: DONE

pkg/sidecar/proxy/options_test.go
DhritiShikhar: Fixed this. DONE.

pkg/sidecar/proxy/options_test.go
DhritiShikhar: Fixed this. DONE.

pkg/sidecar/proxy/options_test.go
DhritiShikhar: Thanks for pointing out. Added a test case for [this](https://github.com/llm-d/llm-d-inference-scheduler/pull/683/changes#diff-2db7b36a7f34202a28506fe7f31c53cda6ceb546d336b5f2e64521131a296977R1272). 
DONE.

pkg/sidecar/proxy/options_test.go
DhritiShikhar: Fixed this. DONE.

pkg/sidecar/proxy/options.go
```diff
@@ -30,10 +30,65 @@ import (
 	"github.com/spf13/pflag"
 	uberzap "go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+	"gopkg.in/yaml.v3"
```
elevran: Please use `sigs.k8s.io/yaml` (already a transitive dependency of the project).
Note that `gopkg.in/yaml.v3` repository was archived by the owner on Apr 1, 2025. It is now read-only and no longer maintained.

pkg/sidecar/proxy/options.go
```diff
@@ -59,15 +114,14 @@ type Options struct {
 	prefillerInsecureSkipVerify bool   // Deprecated: use --tls-insecure-skip-verify=prefiller instead
 	decoderInsecureSkipVerify   bool   // Deprecated: use --tls-insecure-skip-verify=decoder instead
 
-	loggingOptions zap.Options // loggingOptions holds the zap logging configuration
-}
+	loggingOptions                       zap.Options // loggingOptions holds the zap logging configuration
+	FlagSet                              *pflag.FlagSet
```
elevran: Exporting `FlagSet *pflag.FlagSet` leaks implementation detail. Since `FlagSet` is only set to allow `isParsed` to work, the cleanest fix is to pass it through `AddFlags`:

```go
func (opts *Options) AddFlags(fs *pflag.FlagSet) {
    opts.flagSet = fs   // store as an unexported reference
    // ... rest unchanged
}
```

Then `isParsed` can use `opts.flagSet`. The `main.go` assignment `opts.FlagSet = pflag.CommandLine` and the test assignment can both go away.

pkg/sidecar/proxy/options.go
```diff
@@ -320,3 +380,529 @@ func (opts *Options) NewLogger() logr.Logger {
...
+		}
+	}
+	return nil
+}
+
+// YAMLConfigurationFromInlineSpecification extracts YAML configuration provided as inline specification
+// "--configuration={port: 8085, vllm-port: 8203}"
+func (opts *Options) YAMLConfigurationFromInlineSpecification() (map[string]any, error) {
+	var temp map[string]any
+	if err := yaml.Unmarshal([]byte(opts.configurationFromInlineSpecification), &temp); err != nil {
```
elevran: Error from `yaml.Unmarshal` is discarded, losing the actual parse failure message.
You should return a wrapped error instead:
```go
// wrap with %w so the cause propagates:
return nil, fmt.Errorf("failed to unmarshal inline sidecar configuration: %w", err)
```

`YAMLConfigurationFromFile` was correctly fixed with `%w`; this function was not.

pkg/sidecar/proxy/options.go
```diff
@@ -320,3 +380,529 @@ func (opts *Options) NewLogger() logr.Logger {
...
+		return nil, errors.New("failed to unmarshal sidecar configuration")
+	}
+	for key := range temp {
```
elevran: This colon-splitting loop is dead code in practice and potentially harmful.

After `yaml.Unmarshal`, keys in the resulting `map[string]any` are plain strings without embedded colons, as the YAML parser handles all syntax. A key like `"port: 8085"` would only appear if the raw YAML string was not valid YAML and the library returned it as a literal string key, which would already have caused an unmarshal error above.
BTW, if the loop runs on a valid YAML key that legitimately contains a colon (e.g., `"http://something"`), it would silently corrupt the key. 
Please remove this block entirely.

pkg/sidecar/proxy/options.go
```diff
@@ -320,3 +380,529 @@ func (opts *Options) NewLogger() logr.Logger {
...
+			return err
+		}
+	}
+	if opts.configurationFromFile != "" {
+		fileConfiguration, err = opts.YAMLConfigurationFromFile()
+		if err != nil {
+			return err
+		}
+	}
+	switch {
```
elevran, line 408: The three `switch` arms are nearly identical and can be collapsed to ~15 lines, eliminating `mergeYAMLConfigurations` entirely.

Suggested replacement for the entire `extractYAMLConfiguration` body:
```go
func (opts *Options) extractYAMLConfiguration() error {
    var fileCfg, inlineCfg configurationMap

    if opts.configurationFromFile != "" {
        var err error
        fileCfg, err = opts.YAMLConfigurationFromFile()
        if err != nil {
            return err
        }
    }
    if opts.configurationFromInlineSpecification != "" {
        var err error
        inlineCfg, err = opts.YAMLConfigurationFromInlineSpecification()
        if err != nil {
            return err
        }
    }

    // Merge: file is the base; inline overrides with higher priority.
    merged := make(configurationMap)
    for k, v := range fileCfg {
        merged[k] = v
        appendItem(&opts.ConfigurationState.FromFile, k)
    }
    for k, v := range inlineCfg {
        merged[k] = v // inline wins
        appendItem(&opts.ConfigurationState.FromInline, k)
        removeItems(&opts.ConfigurationState.FromFile, []string{k})
    }

    if len(merged) == 0 {
        return nil
    }
    if err := opts.updateSidecarConfiguration(merged); err != nil {
        return err
    }

    // Higher-priority sources shadow lower ones in the state log.
    removeDuplicates(&opts.ConfigurationState.FromFile, opts.ConfigurationState.FromFlags)
    removeDuplicates(&opts.ConfigurationState.FromFile, opts.ConfigurationState.FromEnv)
    removeDuplicates(&opts.ConfigurationState.FromInline, opts.ConfigurationState.FromFlags)
    removeDuplicates(&opts.ConfigurationState.FromInline, opts.ConfigurationState.FromEnv)
    removeDuplicates(&opts.ConfigurationState.Defaults, opts.ConfigurationState.FromInline)
    removeDuplicates(&opts.ConfigurationState.Defaults, opts.ConfigurationState.FromFile)
    return nil
}
```

This also removes the need for `mergeYAMLConfigurations` (delete that function).

pkg/sidecar/proxy/options.go
```diff
@@ -320,3 +380,529 @@ func (opts *Options) NewLogger() logr.Logger {
...
+
+	}
+	return temp, nil
+}
+
+// mergeYAMLConfigurations merges following:
+// 1. YAML configuration from file path `--configuration-file`
+// 2. YAML configuration provided as inline specification `--configuration“,
+// and gives higher priority to configuration provided in inline specification `--configuration`
+func (opts *Options) mergeYAMLConfigurations(fileYAML, parameterYAML map[string]any) map[string]any {
```
elevran: `mergeYAMLConfigurations` mixes two unrelated concerns: map merging and `ConfigurationState` mutation.

When this function is called from `extractYAMLConfiguration`, `ConfigurationState` entries are populated *again* (lines 559–565), and then `extractYAMLConfiguration` also iterates over the same maps to append the same keys again (lines 479–489). This causes double-population of the state lists and requires the subsequent `removeDuplicates` calls to clean up.

The function should either:
- Only merge maps and return the result (pure function, no state side-effects), leaving all state tracking to the caller, or
- Be eliminated entirely — see the simplification suggestion on `extractYAMLConfiguration`.

pkg/sidecar/proxy/options.go
```diff
@@ -320,3 +380,529 @@ func (opts *Options) NewLogger() logr.Logger {
...
+			continue
+		}
+		removeItems(&opts.ConfigurationState.Defaults, []string{flagConfiguration.flag})
+		if opts.isParsed(flagConfiguration.flag) {
+			appendItem(&opts.ConfigurationState.FromFlags, flagConfiguration.flag)
+			continue
+		}
+		if err := flagConfiguration.update(opts, value); err != nil {
+			return fmt.Errorf("update failed for: %v. %w", value, err)
+		}
```
elevran: `inferencePool` is handled as a special case outside `flagConfigurationList`, creating an inconsistency.

All other flags are processed uniformly via `flagConfigurationList` with an `update` closure. `inferencePool` is handled in a separate block at the bottom with different logic (checks `opts.inferencePool == ""` rather than `!opts.isParsed(inferencePool)`). 

This means:
1. The `inferencePool` key in YAML is silently ignored if the process was started with `INFERENCE_POOL` env var (env var sets `opts.inferencePool` at `NewOptions` time), even though the env var has higher precedence and should win. The current check `opts.inferencePool == ""` prevents YAML from setting it, but this is a different check than the `isParsed` approach used for other fields.
2. It's invisible to `getDefaults()` / `getFlags()` state tracking.

Add an entry to `flagConfigurationList` for `inferencePool` consistent with the other fields.

pkg/sidecar/proxy/options.go
```diff
@@ -320,3 +380,529 @@ func (opts *Options) NewLogger() logr.Logger {
...
+			opts.inferencePool = v
+			removeItems(&opts.ConfigurationState.FromEnv, []string{inferencePool, inferencePoolNamespace, inferencePoolName})
+			removeItems(&opts.ConfigurationState.FromFlags, []string{inferencePool, inferencePoolNamespace, inferencePoolName})
+		}
+	}
+	return nil
+}
+
+// isParsed returns true if flag contains parsed value (and not default value)
+func (opts *Options) isParsed(parameter string) bool {
```
elevran: `isParsed` returns `true` when the flag is not found (`flag == nil`), which silently prevents YAML values from being applied.

```go
func (opts *Options) isParsed(parameter string) bool {
    flag := opts.FlagSet.Lookup(parameter)
    return flag == nil || flag.Changed  // ← if flag not found, treated as "CLI-set"
}
```

If `FlagSet` is nil (e.g., `AddFlags` was never called, or in a test that forgot to set it), `FlagSet.Lookup` will panic. And if a flag name in `flagConfigurationList` doesn't match an entry in the FlagSet (e.g., a deprecated flag like `connector` that is in the list but wasn't added to `flagConfigurationList`), `isParsed` returns `true` and the YAML value for that key is silently dropped.

The logic should be: if the flag is unknown, treat it as "not set by CLI" (return `false`). Only return `true` when `flag != nil && flag.Changed`. BTW, ocnsider renaming to `isFlagSet` which is closer to the comment.

pkg/sidecar/proxy/options.go
```diff
@@ -320,3 +380,529 @@ func (opts *Options) NewLogger() logr.Logger {
 		zap.Encoder(zapcore.NewJSONEncoder(config)),
 	)
 }
+
+// GetConfigurationState determines if configuration was obtained from:
+// 1. Default values
+// 2. Environment variables
+// 3. Individual flags
+func (opts *Options) GetConfigurationState() {
```
elevran: `GetConfigurationState` and `extractYAMLConfiguration` share responsibility for populating `ConfigurationState`, but are called at different points in `main.go`, creating a fragile ordering contract.

`GetConfigurationState` (called before `Complete` in main.go) handles Defaults/Flags/Env. Then `Complete` calls `extractYAMLConfiguration` handles FromInline/FromFile and also mutates Defaults/Flags/Env via `removeDuplicates`. 

**Suggestion**: fold `GetConfigurationState` into `Complete`, so there is one call that populates the full state correctly in sequence. This removes the ordering constraint from callers entirely:

```go
func (opts *Options) Complete() error {
    // populate flags/env/defaults state first
    opts.getDefaults()
    opts.getEnvVars()
    opts.getFlags()

    // then process YAML (which can further narrow the lists)
    if err := opts.extractYAMLConfiguration(); err != nil {
        return err
    }
    // ... rest of Complete unchanged
}
```

Remove the exported `GetConfigurationState` method, and update `main.go` to remove the explicit call.

pkg/sidecar/proxy/options.go
```diff
@@ -320,3 +380,529 @@ func (opts *Options) NewLogger() logr.Logger {
...
+	opts.getDefaults()
+	opts.getEnvVars()
+	opts.getFlags()
+}
+
+// getDefaults updates `ConfigurationState` with configuration keys which contain default values
+func (opts *Options) getDefaults() {
```
elevran: `enablePrefillerSampling` is missing from `getDefaults()`?

Every other non-env-var-only flag is listed here, but `enablePrefillerSampling` is absent. When the binary starts without `ENABLE_PREFILLER_SAMPLING` set and `--enable-prefiller-sampling` not passed, the flag uses a `false` default but won't appear in `ConfigurationState.Defaults` log output. Add it to the list for consistent logging.

cmd/pd-sidecar/main.go
```diff
@@ -52,6 +53,8 @@ func main() {
 			}
 		}()
 	}
+	// GetConfigurationState calculates whether configuration is from default values, flags or environment variables
+	opts.GetConfigurationState()
```
elevran: `GetConfigurationState()` must be called before `Complete()` but this constraint is invisible to callers and not checked.

The YAML configuration (inline/file) state is populated *inside* `Complete()` by calling `extractYAMLConfiguration()`, which then calls `removeDuplicates` on the Defaults/Flags/Env lists that `GetConfigurationState` populated. If the order is reversed (or `GetConfigurationState` is called twice), the state will be wrong with no error.

See the comment on `options.go` for the suggested fix: fold `GetConfigurationState` into `Complete` so there's a single, safe call sequence.

cmd/pd-sidecar/main.go
```diff
@@ -65,6 +68,27 @@ func main() {
 		return
 	}
 
+	if len(opts.ConfigurationState.Defaults) != 0 {
```
elevran: There are five separate `logger.Info` calls, each checking a different `len(...) != 0` condition, is noisy at the call site and generates multiple log lines for what is conceptually a single "here is the configuration provenance" event.

Consider a single structured log call after `Complete()` that logs all non-empty categories together:

```go
logger.Info("Sidecar configuration provenance",
    "defaults", opts.ConfigurationState.Defaults,
    "from_flags", opts.ConfigurationState.FromFlags,
    "from_env", opts.ConfigurationState.FromEnv,
    "from_inline", opts.ConfigurationState.FromInline,
    "from_file", opts.ConfigurationState.FromFile,
)
```

This also removes the per-category `if len(...) != 0` guards, an empty slice renders as `[]` which is clear and unambiguous in structured logs.
Side note: might be too much information. Might be sufficient to only log the final configuration, regadless of source.

pkg/sidecar/proxy/options_test.go
```diff
@@ -17,9 +17,1819 @@ limitations under the License.
...
+		return "", err
+	}
+	return path, nil
+}
+
+// newTestOptions creates:
+// 1. new flag set for test
+// 2. new Options struct initialized with default values
+func newTestOptions(t *testing.T) (*Options, *pflag.FlagSet) {
+	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ContinueOnError)
```
elevran: `newTestOptions` resets the global `flag.CommandLine`:

```go
flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ContinueOnError)
```

This is a global mutation. If tests using `newTestOptions` ever run in parallel (or if any other test in the process relies on `flag.CommandLine`), this causes a data race and unpredictable failures. If `t.Parallel()` is ever added to this test or a sibling test, this will break silently.

Use `flag.NewFlagSet` only for the local test scope and don't assign it to `flag.CommandLine`.

elevran (commented):
Thanks @DhritiShikhar!

Good progress addressing the previous round of comments. Most prior feedback is resolved, but a few items remain unaddressed, and the new implementation has grown considerably larger than initially expected. See inline comments for specifics.

I think it would be ok to simplfy by:
1. drop provenance (tracking from where each value is set)
1. allow only one of inline config or file (not both) and 
1. simplify so that CLI > config file/string > defaults (drop ENV, do not allow both direct string and file based YAML)

pkg/sidecar/proxy/options.go
DhritiShikhar: > The inferencePool key in YAML is silently ignored if the process was started with INFERENCE_POOL env var (env var sets opts.inferencePool at NewOptions time), even though the env var has higher precedence and should win.

If the process starts with `INFERENCE_POOL` env var set or `--inference-pool` flag set, then `opts.inferencePool` will be set. Thus, `inferencePool` key in YAML **SHOULD BE IGNORED**. Individual flag and env Var have higher precedence over YAML. This is expected behaviour. And this is what is implemented. Only when `opts.inferencePool` doesnot have a value set, then value is picked up from YAML.

> The current check opts.inferencePool == "" prevents YAML from setting it, but this is a different check than the isParsed approach used for other fields.

This is the priority list:
1. Individual flag `--inference-pool`
2. Env Var `INFERENCE_POOL`
3. YAML Inline specification
4. YAML file
5. Default value

In the case of inference pool, default value is picked up from env var which means if env var is empty (and value is not provided through individual flag `--inference-pool`) then `opts.inferencePool` value is not set. `isParsed()` checks if value is actually provided by user through flag.

If `opts.inferencePool` has value, it either comes from flag or env var. There is no need to call `isParsed()`. If `opts.inferencePool` value is set, it will have higher precedence and YAML configuration for inferencePool WILL BE IGNORED.  

> It's invisible to getDefaults() / getFlags() state tracking.

`getDefaults()` and other provenance tracking code has been removed completely now

> Add an entry to flagConfigurationList for inferencePool consistent with the other fields.

This is done.

pkg/sidecar/proxy/options.go
DhritiShikhar: This is a good one. I missed it. Thanks for this review.

DhritiShikhar:
> Thanks @DhritiShikhar!
> 
> Good progress addressing the previous round of comments. Most prior feedback is resolved, but a few items remain unaddressed, and the new implementation has grown considerably larger than initially expected. See inline comments for specifics.
> 
> I think it would be ok to simplfy by:
> 
> 1. drop provenance (tracking from where each value is set)
> 2. allow only one of inline config or file (not both) and
> 3. simplify so that CLI > config file/string > defaults (drop ENV, do not allow both direct string and file based YAML)

1. Provenance has been removed.
2. Inline and File Config are mutually exclusive now.
3. CLI > inline/ file > defaults is followed in the PR

pkg/sidecar/proxy/options.go
DhritiShikhar: DONE

pkg/sidecar/proxy/options.go
DhritiShikhar: This has been updated to:
`return fmt.Errorf("failed to unmarshal sidecar configuration: %w", err)`

pkg/sidecar/proxy/options.go
DhritiShikhar: That block has been removed

pkg/sidecar/proxy/options.go
DhritiShikhar: `extractYAMLConfiguration` has been updated.

Removed merge and configuration source.

pkg/sidecar/proxy/options.go
DhritiShikhar: `mergeYAMLConfigurations` has been removed completely.

pkg/sidecar/proxy/options.go
DhritiShikhar: ConfigurationState has been removed completely.

pkg/sidecar/proxy/options.go
DhritiShikhar: `getDefaults() ` has been removed completely.

cmd/pd-sidecar/main.go
DhritiShikhar: `GetConfigurationState()` has been removed.

cmd/pd-sidecar/main.go
DhritiShikhar: Sidecar configuration provenance and associated log has been removed from this PR.

pkg/sidecar/proxy/options_test.go
DhritiShikhar: DONE

pkg/sidecar/proxy/options.go
```diff
@@ -97,71 +143,88 @@ var (
...
+	_ = fs.MarkDeprecated(decoderTLSInsecureSkipVerify, "use --tls-insecure-skip-verify=decoder instead")
+
+	fs.StringVar(&opts.InferencePoolNamespace, inferencePoolNamespace, opts.InferencePoolNamespace, "Deprecated: use --inference-pool instead. The Kubernetes namespace for the InferencePool (defaults to INFERENCE_POOL_NAMESPACE env var)")
+	_ = fs.MarkDeprecated(inferencePoolNamespace, "use --inference-pool instead")
+	fs.StringVar(&opts.InferencePoolName, inferencePoolName, opts.InferencePoolName, "Deprecated: use --inference-pool instead. The specific InferencePool name (defaults to INFERENCE_POOL_NAME env var)")
+	_ = fs.MarkDeprecated(inferencePoolName, "use --inference-pool instead")
+	fs.StringVar(&opts.inlineConfiguration, inlineConfiguration, "", "Sidecar configuration in YAML provided as inline specification. Example `--configuration={port: 8085, vllm-port: 8203}`")
+	fs.StringVar(&opts.fileConfiguration, configurationFile, "", "Path to file which contains sidecar configuration in YAML. Example `--configuration-file=/etc/config/sidecar-config.yaml`")
```
elevran: `--max-idle-conns-per-host` flag is removed and should be restored.

The original `AddFlags` registered `--max-idle-conns-per-host` and bound it to `opts.MaxIdleConnsPerHost`. That line was deleted and the flag is not re-added anywhere in the new code even though `MaxIdleConnsPerHost` is still initialised in `NewOptions()` and used by the proxy transport.

pkg/sidecar/proxy/options.go
```diff
@@ -322,3 +392,354 @@ func (opts *Options) NewLogger() logr.Logger {
...
+	}
+	return v, nil
+}
+
+// extractInt extracts int from interface
+func extractInt(value any) (int, error) {
+	switch v := value.(type) {
+	case int:
+		return v, nil
+	case float64:
```
elevran: **Silent truncation and no range validation in `extractInt`**

`sigs.k8s.io/yaml` decodes YAML integers as `float64` (via the JSON path), so this branch is always taken for port/size values. Two problems:

1. `port: 8100.7` silently truncates to `8100` — no error, wrong value.
2. Negative values are accepted: `port: -1` or `data-parallel-size: -5` pass through and cause obscure failures later in `Validate()` or at `net.Listen`.

Suggested fix:
```go
case float64:
    if v != math.Trunc(v) {
        return 0, fmt.Errorf("expected integer, got fractional value %v", v)
    }
    if v < 0 {
        return 0, fmt.Errorf("value must be non-negative, got %v", v)
    }
    return int(v), nil
```

pkg/sidecar/proxy/options.go
```diff
@@ -322,3 +392,354 @@ func (opts *Options) NewLogger() logr.Logger {
...
+	{
+		flag: enableTLS,
+		update: func(opts *Options, v any) error {
+			switch t := v.(type) {
+			case string:
+				opts.enableTLS = strings.Split(t, ",")
+			case []any:
+				temp := make([]string, 0, 10)
+				for _, val := range t {
+					temp = append(temp, fmt.Sprintf("%v", val))
```
elevran: Inconsistent slice-string handling between `enable-tls` and `tls-insecure-skip-verify`
Best if you can eliminate it altogther (see comment on L432).

`enable-tls` accepts non-string list elements and coerces them silently via `fmt.Sprintf` (e.g., `enable-tls: [1, 2, 3]` results in `["1","2","3"]`). The parallel field `tls-insecure-skip-verify` uses `extractSliceString`, which returns an error for non-string elements.

pkg/sidecar/proxy/options.go
```diff
@@ -322,3 +392,354 @@ func (opts *Options) NewLogger() logr.Logger {
...
+		opts.yamlConfiguration = opts.inlineConfiguration
+	case opts.fileConfiguration != "":
+		data, err := os.ReadFile(opts.fileConfiguration)
+		if err != nil {
+			return fmt.Errorf("failed to read sidecar configuration from file: %w", err)
+		}
+		opts.yamlConfiguration = string(data)
+	}
+	if opts.yamlConfiguration != "" {
+		if err := yaml.Unmarshal([]byte(opts.yamlConfiguration), &configuration); err != nil {
```
elevran: Typos in YAML keys are silently ignored.

The YAML is unmarshaled into `map[string]any`. Keys not matching any entry in `flagConfigurationList` are quietly discarded, so a user writing:
```yaml
kv-connectors: sglang   # typo: should be kv-connector
```
gets no warning and the sidecar silently starts with the default connector. This is risky.

Consider warning (at least at log level) when the parsed map contains keys that do not appear in `flagConfigurationList`. A better option is to use a typed struct with strict unmarshalling (see comments below).

pkg/sidecar/proxy/options.go
```diff
@@ -322,3 +392,354 @@ func (opts *Options) NewLogger() logr.Logger {
...
+
+// flagConfiguration contains
+// 1. flag name
+// 2. function to update value when flag contains default value
+type flagConfiguration struct {
+	flag   string
+	update func(opts *Options, value any) error
+}
+
+var flagConfigurationList = []flagConfiguration{
```
elevran: The whole `flagConfiguration` / `flagConfigurationList` can be eliminated.

The entire dispatch table exists to solve one problem: unmarshal a YAML map into `Options` fields, but only when the corresponding CLI flag was not explicitly set. The idiomatic solution is a **typed intermediary struct**:

```go
type sidecarFileConfig struct {
    Port             int      `json:"port,omitempty"`
    VLLMPort         int      `json:"vllm-port,omitempty"`
    DataParallelSize int      `json:"data-parallel-size,omitempty"`
    KVConnector      string   `json:"kv-connector,omitempty"`
    EnableTLS        []string `json:"enable-tls,omitempty"`
    // ... one field per configurable flag
}
```

`yaml.Unmarshal` (or `yaml.UnmarshalStrict`) into this struct handles all type coercion correctly and the manual `extractInt`/`extractBool`/`extractString`/`extractSliceString` helpers can also be removed.

Then `updateSidecarConfiguration` becomes:
```go
func (opts *Options) mergeFileConfig(cfg sidecarFileConfig) {
    if cfg.Port != 0 && !opts.isFlagSet(port) { opts.Port = strconv.Itoa(cfg.Port) }
    // ... one line per field
}
```

So can simplify by removing `flagConfiguration` struct, `flagConfigurationList` variable, `configurationMap` type, the various `extractXYZ` helpers.
This would also eliminiate a `float64`-truncation bug, and the silent unknown key due to typo bug (if you use `yaml.UnmarshalStrict`).

pkg/sidecar/proxy/options.go
```diff
@@ -59,16 +107,14 @@ type Options struct {
 	prefillerInsecureSkipVerify bool   // Deprecated: use --tls-insecure-skip-verify=prefiller instead
 	decoderInsecureSkipVerify   bool   // Deprecated: use --tls-insecure-skip-verify=decoder instead
 
-	loggingOptions zap.Options // loggingOptions holds the zap logging configuration
+	loggingOptions      zap.Options // loggingOptions holds the zap logging configuration
+	goFlagSet           *flag.FlagSet
+	pflagSet            *pflag.FlagSet
+	inlineConfiguration string
+	fileConfiguration   string
+	yamlConfiguration   string
```
elevran: `yamlConfiguration` should be a local variable, not a struct field.

This field is assigned once inside `extractYAMLConfiguration()` and never read after the function returns. Storing it on the struct leaks an implementation detail and holds the raw configuration string for the lifetime of the `Options` object.

The replacement with a local variable inside `extractYAMLConfiguration` also removes the `[]byte => string => []byte` round-trip conversions (the file bytes can be passed directly to `yaml.Unmarshal` without converting through a string).

pkg/sidecar/proxy/options.go
```diff
@@ -97,71 +143,88 @@ var (
...
+	_ = fs.MarkDeprecated(inferencePoolNamespace, "use --inference-pool instead")
+	fs.StringVar(&opts.InferencePoolName, inferencePoolName, opts.InferencePoolName, "Deprecated: use --inference-pool instead. The specific InferencePool name (defaults to INFERENCE_POOL_NAME env var)")
+	_ = fs.MarkDeprecated(inferencePoolName, "use --inference-pool instead")
+	fs.StringVar(&opts.inlineConfiguration, inlineConfiguration, "", "Sidecar configuration in YAML provided as inline specification. Example `--configuration={port: 8085, vllm-port: 8203}`")
+	fs.StringVar(&opts.fileConfiguration, configurationFile, "", "Path to file which contains sidecar configuration in YAML. Example `--configuration-file=/etc/config/sidecar-config.yaml`")
+}
+
+// validateMutuallyExclusiveFlags validates that user has not provided both --configuration and --configuration-file flags, returning an error if both are explicitly set
+func (opts *Options) validateMutuallyExclusiveFlags() error {
```
elevran: `extractYAMLConfiguration` already handles "both flags set" via the `switch` where inline wins and the file is ignored. You can move the warning / error reporting there and eliminate this function.

Also document in `AddFlags` flags' descriptions that `--configuration` takes priority over `--configuration-file` when both are provided.

pkg/sidecar/proxy/options.go
```diff
@@ -322,3 +392,354 @@ func (opts *Options) NewLogger() logr.Logger {
...
+	if !ok {
+		return false, fmt.Errorf("type assertion failed for value: %v", value)
+	}
+	return v, nil
+}
+
+// extractInt extracts int from interface
+func extractInt(value any) (int, error) {
+	switch v := value.(type) {
+	case int:
```
elevran: The `case int:` is dead code.

`sigs.k8s.io/yaml` decodes YAML by first converting to JSON and then using the standard JSON decoder. For `map[string]any`, the JSON decoder represents all numbers as `float64`, never as ints. This means that only the `case float64:` is ever hit. It has silent truncation and no range validation (e.g., `port: 8100.7` silently truncates to `8100` with no error, even though it is the wrong value type; negative values are accepted: `port: -1` or `data-parallel-size: -5` pass through and cause obscure failures later in `Validate()` or at `net.Listen`).

This is another argument for the typed-struct approach (see comment on `flagConfigurationList`): unmarshaling into a typed struct delivers values in the correct Go type, making the entire `extractInt` switch unnecessary.

pkg/sidecar/proxy/options.go
elevran: Unnecessary `[]byte => string => []byte` conversions

```go
data, _ := os.ReadFile(...)                          // []byte
opts.yamlConfiguration = string(data)                // alloc: []byte → string
yaml.Unmarshal([]byte(opts.yamlConfiguration), ...)  // alloc: string → []byte
```

Note that `yaml.Unmarshal` accepts `[]byte` directly, so can pass `data` in without converting. See comment on line 115 for making this a local variable and not a field on Options.

pkg/sidecar/proxy/options.go
```diff
@@ -97,71 +143,88 @@ var (
...
+	fs.StringVar(&opts.InferencePoolName, inferencePoolName, opts.InferencePoolName, "Deprecated: use --inference-pool instead. The specific InferencePool name (defaults to INFERENCE_POOL_NAME env var)")
+	_ = fs.MarkDeprecated(inferencePoolName, "use --inference-pool instead")
+	fs.StringVar(&opts.inlineConfiguration, inlineConfiguration, "", "Sidecar configuration in YAML provided as inline specification. Example `--configuration={port: 8085, vllm-port: 8203}`")
+	fs.StringVar(&opts.fileConfiguration, configurationFile, "", "Path to file which contains sidecar configuration in YAML. Example `--configuration-file=/etc/config/sidecar-config.yaml`")
+}
+
+// validateMutuallyExclusiveFlags validates that user has not provided both --configuration and --configuration-file flags, returning an error if both are explicitly set
+func (opts *Options) validateMutuallyExclusiveFlags() error {
+	if opts.pflagSet != nil {
+		inlineChanged := opts.pflagSet.Lookup(inlineConfiguration).Changed
```
elevran: nit: if we ever deviate from this and either of these flags is not registered, `.Changed` will deref a nil and panic. It would be safer to check lookup for nil return first. However, I prefer this entire function is simplified away as commented above.

pkg/sidecar/proxy/options.go
```diff
@@ -175,10 +238,17 @@ func validateStages(stages []string, supportedStages map[string]struct{}, flagNa
 }
 
 // Complete performs post-processing of parsed command-line arguments.
-// It handles migration from deprecated flags, parses the InferencePool field,
-// computes boolean TLS fields, and builds Config.DecoderURL.
+// It validates mutually exclusive flags, extracts YAML configuration (if provided), handles migration from deprecated flags,
+// parses the InferencePool field, computes boolean TLS fields, and builds Config.DecoderURL.
 // After Complete(), opts.Config is fully populated.
 func (opts *Options) Complete() error {
+	if err := opts.validateMutuallyExclusiveFlags(); err != nil {
```
elevran: see earlier comment on function removal.

pkg/sidecar/proxy/options.go
```diff
@@ -32,8 +32,56 @@ import (
 	"go.uber.org/zap/zapcore"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	logutil "sigs.k8s.io/gateway-api-inference-extension/pkg/common/observability/logging"
```
elevran: note that following #870 we should not longer be importing GAIE but using the corresponding local path from llm-d-inference-scheduler.

elevran:
additional note: #822 introduces a new flag and option. If it merges before, needs to be rebased/accounted for here.

pkg/sidecar/proxy/options.go
DhritiShikhar: This is fixed.

pkg/sidecar/proxy/options.go
DhritiShikhar: Got it!

pkg/sidecar/proxy/options.go
DhritiShikhar: DONE. This is fixed.

pkg/sidecar/proxy/options.go
DhritiShikhar: > Consider warning (at least at log level) when the parsed map contains keys that do not appear in flagConfigurationList

DONE

DhritiShikhar: DONE

pkg/sidecar/proxy/options.go
DhritiShikhar: This is fixed. Removed unnecessary code. Thanks for this input. This is much better.

pkg/sidecar/proxy/options.go
DhritiShikhar: Restored `--max-idle-conns-per-host` flag - https://github.com/llm-d/llm-d-inference-scheduler/pull/683/changes#diff-d21d44dfeac0b1e61c9684c2d425e9e54890171194f3f5913804dfd5d3a57336R235

pkg/sidecar/proxy/options.go
DhritiShikhar: > You can move the warning / error reporting there and eliminate this function

1. Removed `validateMutuallyExclusiveFlags()` function completely.
2. Error is reported here https://github.com/llm-d/llm-d-inference-scheduler/pull/683/changes#diff-d21d44dfeac0b1e61c9684c2d425e9e54890171194f3f5913804dfd5d3a57336R412

pkg/sidecar/proxy/options.go
DhritiShikhar: I have removed `validateMutuallyExclusive()` function completely

pkg/sidecar/proxy/options.go
DhritiShikhar: > gets no warning and the sidecar silently starts with the default connector. This is risky.

Agreed.

> Consider warning (at least at log level) when the parsed map contains keys that do not appear in flagConfigurationList. A better option is to use a typed struct with strict unmarshalling (see comments below).

Skipped logging. Decided to throw error. This is handled here. Please take a look - https://github.com/llm-d/llm-d-inference-scheduler/pull/683/changes#diff-d21d44dfeac0b1e61c9684c2d425e9e54890171194f3f5913804dfd5d3a57336R428

DhritiShikhar: Removed unnessecary conversions.. back and forth.

Right now, there is only one conversion:

```
yamlData = []byte(opts.inlineConfiguration)
```

pkg/sidecar/proxy/options.go
DhritiShikhar: Thanks for the input.

1. Removed `flagConfiguration` list
2. Removed helper functions `extractInt`, `extractBool`, etc
3. I use intermediary struct `yamlConfiguration` now to unmarshal
4. `mergeYAMLConfiguration` to update values

cmd/pd-sidecar/main.go
DhritiShikhar: Let's not add it unless we really need it.

pkg/sidecar/proxy/options.go
DhritiShikhar: This has been updated:
1. Added `pflagSet` and `goFlagSet` which is unexported.
2. Passed pflagSet through `AddFlags()`

```go
func (opts *Options) AddFlags(fs *pflag.FlagSet) {
	if opts.pflagSet == nil {
		opts.pflagSet = fs
	}
	if opts.goFlagSet == nil {
		opts.goFlagSet = flag.NewFlagSet("goFlagSet", flag.ContinueOnError)
		// Add logging flags to the standard flag set
		opts.loggingOptions.BindFlags(opts.goFlagSet)
		fs.AddGoFlagSet(opts.goFlagSet)
	}
```

pkg/sidecar/proxy/options.go
DhritiShikhar: This code has been removed.

pkg/sidecar/proxy/options.go
DhritiShikhar: This code has been removed.

pkg/sidecar/proxy/options.go
DhritiShikhar: Removed this function.

pkg/sidecar/proxy/options.go
```diff
@@ -98,71 +165,77 @@ var (
...
+	_ = fs.MarkDeprecated(inferencePoolNamespace, "use --inference-pool instead")
+	fs.StringVar(&opts.InferencePoolName, inferencePoolName, opts.InferencePoolName, "Deprecated: use --inference-pool instead. The specific InferencePool name (defaults to INFERENCE_POOL_NAME env var)")
+	_ = fs.MarkDeprecated(inferencePoolName, "use --inference-pool instead")
 	fs.IntVar(&opts.MaxIdleConnsPerHost, "max-idle-conns-per-host", opts.MaxIdleConnsPerHost, "max idle keep-alive connections per host for reverse proxy transports; set to at least the expected concurrency")
+	fs.StringVar(&opts.inlineConfiguration, inlineConfiguration, "", "Sidecar configuration in YAML provided as inline specification. Example `--configuration={port: 8085, vllm-port: 8203}. Inline configuration takes precedence over file configuration if both are provided.`")
+	fs.StringVar(&opts.fileConfiguration, configurationFile, "", "Path to file which contains sidecar configuration in YAML. Example `--configuration-file=/etc/config/sidecar-config.yaml`. File configuration is ignored if inline configuration is provided.")
```
elevran: The description text (and PR description text) contradict the actual implementation: inline and file are mutually exclusive and there is no precedence. CLI description (and the PR description) should reflect that. 

nit: PR description also says `CLI flags > env vars > inline spec > file > defaults` whereas the actual precedence is `CLI flags > inline/file YAML > env vars > defaults` (i.e., env vars are applied in NewOptions() before flags are parsed).
We have started documenting user facing changes in PR descriptions. I've updated on your behalf.

pkg/sidecar/proxy/options.go
```diff
@@ -33,8 +33,78 @@ import (
...
+	ecConnector             = "ec-connector"
+	enableSSRFProtection    = "enable-ssrf-protection"
+	enablePrefillerSampling = "enable-prefiller-sampling"
+	enableTLS               = "enable-tls"
+	tlsInsecureSkipVerify   = "tls-insecure-skip-verify"
+	secureServing           = "secure-proxy"
+	certPath                = "cert-path"
+	inferencePool           = "inference-pool"
+	poolGroup               = "pool-group"
+	MaxIdleConnsPerHost     = "max-idle-conns-per-host"
```
elevran: No need to export the symbol. Use `maxIdleConnsPerHost` to have same exposure as other constants

pkg/sidecar/proxy/options_test.go
```diff
@@ -17,9 +17,502 @@ limitations under the License.
...
+}
+
+func baseExpected() expected {
+	return expected{
+		port:                    defaultPort,
+		vLLMPort:                defaultVLLMPort,
+		dataParallelSize:        defaultDataParallelSize,
+		kVConnector:             KVConnectorNIXLV2,
+		eCConnector:             ECExampleConnector,
+		secureServing:           false,
```
elevran: `NewOptions()` defaults `SecureServing: true`, but here it is set `secureServing: false`. Every test case that inherits from baseExpected() and doesn't explicitly set a YAML value would silently assert false against a true default. The base struct should mirror the actual defaults, and the YAML test cases should explicitly override.

pkg/sidecar/proxy/options.go
```diff
@@ -98,71 +165,77 @@ var (
...
-	fs.AddGoFlagSet(flag.CommandLine)
-
-	fs.StringVar(&opts.Port, "port", opts.Port, "the port the sidecar is listening on")
-	fs.StringVar(&opts.vllmPort, "vllm-port", opts.vllmPort, "the port vLLM is listening on")
-	fs.IntVar(&opts.DataParallelSize, "data-parallel-size", opts.DataParallelSize, "the vLLM DATA-PARALLEL-SIZE value")
-	fs.StringVar(&opts.KVConnector, "kv-connector", opts.KVConnector,
+	if opts.pflagSet == nil {
+		opts.pflagSet = fs
+	}
+	if opts.goFlagSet == nil {
```
elevran: `goFlagSet` is effectively a local variable and need not be stored on the Options struct. Only pflagSet is used outside AddFlags.

```go
func (opts *Options) AddFlags(fs *pflag.FlagSet) {
      if opts.pflagSet == nil {
          opts.pflagSet = fs
      }
      goFlagSet := flag.NewFlagSet("goFlagSet", flag.ContinueOnError)
      opts.loggingOptions.BindFlags(goFlagSet)
      fs.AddGoFlagSet(goFlagSet)
      // ... rest of flag registrations
  }
```

pkg/sidecar/proxy/options_test.go
DhritiShikhar: I have removed `baseExpected()`

pkg/sidecar/proxy/options.go
DhritiShikhar: This is done. Removed `goFlagSet` from Options struct.

pkg/sidecar/proxy/options.go
DhritiShikhar: `maxIdleConnsPerHost` is unexported now.

pkg/sidecar/proxy/options.go
DhritiShikhar: I have updated the flag description text.

elevran (commented):
Previous feedback rounds are all addressed - thanks! Please fix conflict, otherwise /lgtm

elevran:
thanks for persevering with this one @DhritiShikhar 
/lgtm
/approve

DhritiShikhar:
Thanks @elevran for all the help with this PR!

PR #685, opened 2026-03-06
fix: simplify InferencePool flag to namespace/name format

### LINKED ISSUE #684: Simplify InferencePool CLI flags to use ns/name syntax
Replace `--inference-pool-namespace` and `--inference-pool-name` with single `--inference-pool` flag accepting `namespace/name` format.

This is a review comment from https://github.com/llm-d/llm-d-inference-scheduler/pull/682#discussion_r2894396484

gyliu513:
Fixed #684

### DIFF
--- a/cmd/pd-sidecar/main.go
+++ b/cmd/pd-sidecar/main.go
@@ -90,10 +90,14 @@ func main() {
 		"cert-path", "", "The path to the certificate for secure proxy. The certificate and private key files "+
 			"are assumed to be named tls.crt and tls.key, respectively. If not set, and secureProxy is enabled, "+
 			"then a self-signed certificate is used (for testing).")
-	enableSSRFProtection := flag.Bool("enable-ssrf-protection", false, "enable SSRF protection using InferencePool allowlisting")
-	inferencePool := flag.String("inference-pool", os.Getenv("INFERENCE_POOL"), "InferencePool to watch in namespace/name format (e.g., default/my-pool). Can also use INFERENCE_POOL env var.")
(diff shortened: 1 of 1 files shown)


cmd/pd-sidecar/main.go
```diff
@@ -100,16 +99,21 @@ func main() {
...
-			logger.Info("Error: --inference-pool-namespace or INFERENCE_POOL_NAMESPACE environment variable is required when --enable-ssrf-protection is true")
+		if *inferencePool == "" {
+			logger.Info("Error: --inference-pool is required when --enable-ssrf-protection is true")
 			return
 		}
-		if *inferencePoolName == "" {
-			logger.Info("Error: --inference-pool-name or INFERENCE_POOL_NAME environment variable is required when --enable-ssrf-protection is true")
+
+		parts := strings.SplitN(*inferencePool, "/", 2)
+		if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
```
elevran: AFAIK, k8s also accepts a single element namespaced name which implies the use of "default" namespace.

elevran (commented):
Thanks @gyliu513 - please consider allowing `<name>` (with an implied "default" used for the namespace).

Also, since users might already be using the current flags, we would like to allow for some migration time by deprecating (and not removing) the existing flags. Thus, suggest the following for handling during the deprecation period:
- add deprecation mark on the flags being replaced (might require an update to `pflag` package or manually testing if a deprecated flag was set)
- prefer `--inference-pool` over the name and namespace flags if both are set (and log a message of this)
- if no inference flags are set, use `INFERENCE_POOL` environment or `INFERENCE_POOL_NAMESPACE`/`INFERENCE_POOL_NAME` if these are set.

cmd/pd-sidecar/main.go
gyliu513: Done

PR #687, opened 2026-03-08
Add presubmit makefile target for local verification before pushing a PR

elevran:
Add Makefile target to allow consistent manual check before pushing PR, if desired. 
The goal is to capture some common issues before submitting a PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,34 @@ help: ## Print help
 
 ##@ Development
 
+.PHONY: install-hooks
+install-hooks: ## Install git hooks
+	git config core.hooksPath hooks
+
+.PHONY: presubmit
+presubmit: LINT_NEW_ONLY=true
+presubmit: git-branch-check signed-commits-check go-mod-check format lint
+
+.PHONY: git-branch-check
+git-branch-check:
+	@branch=$$(git rev-parse --abbrev-ref HEAD); \
+	if [ "$$branch" = "main" ]; then \
+		echo "ERROR: Direct push to 'main' is not allowed."; \
+		echo "Create a branch and open a PR instead."; \
+		exit 1; \
+	fi
+
+.PHONY: signed-commits-check
+signed-commits-check:
+	@./scripts/check-commits.sh upstream/main
+
+.PHONY: go-mod-check
+go-mod-check:
+	@echo "Checking go.mod/go.sum are clean..."
+	@go mod tidy
+	@git diff --exit-code go.mod go.sum || \
+	( echo "ERROR: go.mod/go.sum are not tidy. Run 'go mod tidy' and commit."; exit 1 )
+
 .PHONY: clean
 clean: ## Clean build artifacts, tools and caches
 	go clean -testcache -cache
@@ -115,10 +143,6 @@ lint: check-golangci-lint check-typos ## Run lint (use LINT_NEW_ONLY=true to onl
 		$(GOLANGCI_LINT) run; \
 	fi
 	$(TYPOS)
-
-.PHONY: install-hooks
-install-hooks: ## Install git hooks
-	git config core.hooksPath hooks
 
 .PHONY: test
 test: test-unit test-e2e ## Run all tests (unit and e2e)
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -20,6 +20,12 @@ Documentation for developing the inference scheduler.
 [ZeroMQ]:https://zeromq.org/
 
 > [!NOTE]
+> Before committing and pushing changes to an upstream repository, you may want to 
+> explicitly run the `make presubmit` target to avoid failing PR checks. The checks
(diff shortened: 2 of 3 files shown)


.githooks/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+
+cd "$(git rev-parse --show-toplevel)"
+make precommit
```
ahg-g: when does precommit get executed exactly? when one commits to their own branch?

.githooks/pre-push
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -e
+
+# Move to repo root
+cd "$(git rev-parse --show-toplevel)"
+make presubmit
```
ahg-g: when does presubmit get executed? on an open PR and after /approve is applied?

.githooks/pre-commit
elevran: On git commit. It is in the PR description

.githooks/pre-push
elevran: On git push

ahg-g: Thanks, does this mean both hooks execute locally? if so, do we see value in having both hooks? what do we miss if all the tests execute on git push for example? 

Another question, if git push indeed executes locally, what is the path to having public logs of the tests that run on `git push`? I am thinking about test flakes, having a public record of the failed test is useful to address them.

elevran: both hooks execute locally and are opt-in (`make install hooks`). 
The same checks will run in GH actions, after the push, but this gives the developer to check earlier.
We could move all checks to the pre-push hook and remove the pre-commit. I intentionally tried to make commit hook light weight, but this would reduce it to zero.

I`m thinking of additional check we can do locally (as long as it does not have noticeable delays), such as checking for signed commits. The checks are based on common PR failings that can be easily be avoided by running this before it arrives on GH.

Regarding the record of test flakes - any idea how to distinguish flakes from failed tests that have to do with the current commit? Right now we only capture failures on GH actions so those are public.

Looking for feedback on:
- should we keep both or just `pre-push`
- would it make sense to rename to `pre-commit` and `pre-push`
- removing the hooks altogther and leaving only `make` that a user **may** run if they so wish

Makefile
```diff
@@ -93,13 +93,40 @@ help: ## Print help
 
 ##@ Development
 
+.PHONY: install-hooks
+install-hooks: ## Install git hooks
+	git config core.hooksPath .githooks
+
+.PHONY: precommit ## git commit hook
+precommit:
+	@test -z "$$(gofmt -l .)"
```
shmuelk: Do you expect people to do one large commit? Otherwise this becomes a pain.

elevran: Can you explain what you see as the pain?
Running `gofmt` on the code base takes ~0s

.githooks/pre-push
ahg-g: > Regarding the record of test flakes - any idea how to distinguish flakes from failed tests that have to do with the current commit? 

Is there a way in GH actions to trigger running the tests periodically? a few times a day? 

> Right now we only capture failures on GH actions so those are public.

Those will help too because a reviewer sees them and ideally creates issues to track irrelevant test failures.

> removing the hooks altogther and leaving only make that a user may run if they so wish

I would remove them since a developer can run the tests manually if they wish to, at the end of the day they will trigger on GH actions. But I don't have strong opinion here, keeping them is fine as well.

ahg-g: btw, I assume we haven't yet set up the GH actions to run presubmit tests as I don't see them in https://github.com/llm-d/llm-d-inference-scheduler/tree/main/.github/actions, right?

elevran: > Is there a way in GH actions to trigger running the tests periodically? a few times a day?

Yes. Can run actions on Cron like schedule.

> btw, I assume we haven't yet set up the GH actions to run presubmit tests as I don't see them in https://github.com/llm-d/llm-d-inference-scheduler/tree/main/.github/actions, right?

We have actions with mandatory passing tests that include everything in the pre-submit hook (lint, unit test, etc.).
They are not called presubmit, though. See the [ci-pr-checks workflow](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/.github/workflows/ci-pr-checks.yaml).
Unless you misunderstood your comment, the functionality is already present in GH actions and runs on every PR. The hooks were meant to allow running the same **before** creating the PR. Since local hooks can not be enforced, I will remove the hooks and leave a `presubmit` makefile target to run quick sanity locally.

Makefile
elevran: Either way, I will remove the hooks and make these regular makefile targets that developers may wish to run before creating a PR (nothing on commits).

shmuelk: My bad. I read that somehow as running the tests.

PR #689, opened 2026-03-08
Replace supportedConnectors as a list with a map

roytman:
Replace `supportedConnectors` as a list with a map, plus add `depreciation` test during the validation.

### DIFF
--- a/cmd/pd-sidecar/main.go
+++ b/cmd/pd-sidecar/main.go
@@ -31,20 +31,11 @@ import (
 	"github.com/llm-d/llm-d-inference-scheduler/pkg/telemetry"
 )
 
-var (
-	// supportedConnectors defines all valid P/D connector types
-	supportedConnectors = []string{
-		proxy.ConnectorNIXLV2,
-		proxy.ConnectorSharedStorage,
-		proxy.ConnectorSGLang,
-	}
-)
-
 func main() {
 	port := flag.String("port", "8000", "the port the sidecar is listening on")
 	vLLMPort := flag.String("vllm-port", "8001", "the port vLLM is listening on")
 	vLLMDataParallelSize := flag.Int("data-parallel-size", 1, "the vLLM DATA-PARALLEL-SIZE value")
-	connector := flag.String("connector", proxy.ConnectorNIXLV2, "the P/D connector being used. Supported: "+strings.Join(supportedConnectors, ", "))
+	connector := flag.String("connector", proxy.ConnectorNIXLV2, "the P/D connector being used. Supported: "+strings.Join(proxy.GetConnectorNames(), ", "))
 	prefillerUseTLS := flag.Bool("prefiller-use-tls", false, "whether to use TLS when sending requests to prefillers")
 	decoderUseTLS := flag.Bool("decoder-use-tls", false, "whether to use TLS when sending requests to the decoder")
 	prefillerInsecureSkipVerify := flag.Bool("prefiller-tls-insecure-skip-verify", false, "configures the proxy to skip TLS verification for requests to prefiller")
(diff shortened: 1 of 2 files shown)


pkg/sidecar/proxy/proxy.go
```diff
@@ -73,6 +73,43 @@ const (
 	LegacyPoolGroup = "inference.networking.x-k8s.io"
 )
 
+// ConnectorInfo holds metadata about a connector
+type ConnectorInfo struct {
+	Description string
+	Deprecated  bool
```
elevran, line 79: None of the kV connectors is deprecated. The shared storage is used in several use cases (e.g., decode-first)

cmd/pd-sidecar/main.go
```diff
@@ -87,18 +78,22 @@ func main() {
...
 
 	// Validate connector
-	isValidConnector := false
-	for _, validConnector := range supportedConnectors {
-		if *connector == validConnector {
-			isValidConnector = true
-			break
-		}
-	}
+	connectorInfo, isValidConnector := proxy.IsValidConnector(*connector)
```
elevran, line 81: nit: since this is a new function and does not require backward compatibility, you could use `IsValidKV/ECConnector` anticipating future changes

cmd/pd-sidecar/main.go
```diff
@@ -31,20 +31,11 @@ import (
...
-		proxy.ConnectorSGLang,
-	}
-)
-
 func main() {
 	port := flag.String("port", "8000", "the port the sidecar is listening on")
 	vLLMPort := flag.String("vllm-port", "8001", "the port vLLM is listening on")
 	vLLMDataParallelSize := flag.Int("data-parallel-size", 1, "the vLLM DATA-PARALLEL-SIZE value")
-	connector := flag.String("connector", proxy.ConnectorNIXLV2, "the P/D connector being used. Supported: "+strings.Join(supportedConnectors, ", "))
+	connector := flag.String("connector", proxy.ConnectorNIXLV2, "the P/D connector being used. Supported: "+strings.Join(proxy.GetConnectorNames(), ", "))
```
elevran, line 38: nit: new function `GetConnectorNames()` could use `KV`/`EC` in name for known future expansion. Alternatively, pass the connector type as parameter so you could key off it for testing either KV or EC.

cmd/pd-sidecar/main.go
```diff
@@ -87,18 +78,22 @@ func main() {
...
+	connectorInfo, isValidConnector := proxy.IsValidConnector(*connector)
 	if !isValidConnector {
-		logger.Info("Error: --connector must be one of: " + strings.Join(supportedConnectors, ", "))
+		logger.Info("Error: --connector must be one of: " + strings.Join(proxy.GetConnectorNames(), ", "))
 		return
 	}
-	logger.Info("p/d connector validated", "connector", connector)
+
+	// Warn if connector is deprecated
+	if connectorInfo.Deprecated {
```
elevran, line 88: See note later - no Connector deprecation currently. The previously deprecated Connector has been repurposed and is now used.

cmd/pd-sidecar/main.go
```diff
@@ -87,18 +78,22 @@ func main() {
...
+		logger.Info("WARNING: connector is deprecated",
+			"connector", *connector,
+			"description", connectorInfo.Description)
+	}
+
+	logger.Info("p/d connector validated",
```
elevran, line 94: nit: In preparation for EPD, consider using KV as connector name and not `p/d`

pkg/sidecar/proxy/proxy.go
```diff
@@ -73,6 +73,43 @@ const (
 	LegacyPoolGroup = "inference.networking.x-k8s.io"
 )
 
+// ConnectorInfo holds metadata about a connector
```
elevran, line 76: consider a new file for these functions, instead of in proxy.go

pkg/sidecar/proxy/proxy.go
roytman: See https://github.com/llm-d/llm-d-inference-scheduler/blob/c910eebbd53f39094599c27c7f01c0230a7312fb/pkg/sidecar/proxy/proxy.go#L64

PR #691, opened 2026-03-09
NonCachedTokens defines the minimum number of non-cached tokens required

modassarrana89-new:
Fixed https://github.com/llm-d/llm-d-inference-scheduler/issues/254

### DIFF
--- a/pkg/plugins/profile/prefix_based_pd_decider.go
+++ b/pkg/plugins/profile/prefix_based_pd_decider.go
@@ -70,6 +70,10 @@ func NewPrefixBasedPDDecider(config PrefixBasedPDDeciderConfig) (*PrefixBasedPDD
 		return nil, err
 	}
 
+	if config.NonCachedTokens == 0 {
+		log.Log.Info("Prefix-based PD disabled (NonCachedTokens=0)")
+	}
+
 	return &PrefixBasedPDDecider{
 		config: config,
 	}, nil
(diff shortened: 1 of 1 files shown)


pkg/plugins/profile/prefix_based_pd_decider.go
```diff
@@ -90,8 +90,11 @@ func (d *PrefixBasedPDDecider) disaggregate(ctx context.Context, inputTokens int
 	logger := log.FromContext(ctx)
 	debugLogger := log.FromContext(ctx).V(logutil.DEBUG)
 
-	if d.config.NonCachedTokens <= 0 { // always use disaggregation in case of non cached tokens number is 0
-		return true
+	// NonCachedTokens defines the minimum number of non-cached tokens required
```
elevran, line 97: @modassarrana89-new the configuration is valid for the duration of the decider's lifetime, and (AFAIK) this endpont is called for each PD cycle (@mayabar - true?).
If that's the case, would suggest moving this message to the Factory or New methods so it is printed once.
Will let @mayabar make the final call whether it is better here or in the factory/new.

mayabar: I agree there is no need to log this message for every request. It will be enough to log this message in`pd_profile_handler` where decider is created.
In case of NonCachedTokens==0, it will be only one cycle but it will still execute for all requests.

modassarrana89-new: sure @mayabar @elevran I will move it to pd_profile_handler

elevran (commented):
/lgtm
/approve

elevran (approved):

PR #696, opened 2026-03-10
replace map[string]bool with map[string]struct{}

roytman:
Replaced the list of supported connectors with a map to achieve $O(1)$ lookup time.

This change is based on the [comment](https://github.com/llm-d/llm-d-inference-scheduler/pull/643#discussion_r2902050488)


elevran:
@roytman _might_ need a rebase/merge as #682 merges

Issue #636, opened 2026-02-19
Implement Options pattern for sidecar proxy


elevran:
**What would you like to be added**:
Follow the options pattern for sidecar proxy.
See for example usage in [IGW EPP](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/pkg/epp/server/options.go)

**Why is this needed**:
Centralize and make comand line handling and validation for sidecar simpler and more consistent.

nirrozenbaum:
please pay attention to this issue:
xref https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2378

once that one is merged you might be able to reuse logging options ootb.

PR #697, opened 2026-03-10
Implement Options pattern for sidecar proxy

Mohamedma96:
**Summary**
Refactored pd-sidecar to use a centralized Options struct for cleaner, more maintainable configuration management.

**Changes**
1. Added: pkg/sidecar/proxy/options.go
2. Added: pkg/sidecar/proxy/options_test.go
3. Modified: cmd/pd-sidecar/main.go - replaced manual flag handling with Options struct

Why we need this PR:
- Centralized configuration in single struct
- Comprehensive validation logic
- Reduced code complexity.

Fixes:
issue #636


cmd/pd-sidecar/main.go
```diff
@@ -32,78 +29,15 @@ import (
...
-	poolGroup := pflag.String("pool-group", proxy.DefaultPoolGroup, "group of the InferencePool this Endpoint Picker is associated with.")
-
-	opts := zap.Options{}
-	opts.BindFlags(flag.CommandLine) // optional to allow zap logging control via CLI
-
+	// Initialize options with defaults
+	opts := proxy.NewOptions()
+	opts.AddFlags(pflag.CommandLine)
 	// Add Go flags to pflag (for zap options compatibility)
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
```
elevran: Consider if zap logging flags can be handled inside `AddFlags` instead of in main.
L36-40

cmd/pd-sidecar/main.go
```diff
@@ -123,107 +57,73 @@ func main() {
 		}()
 	}
 
-	// Migrate deprecated boolean TLS flags to new StringSlice flags
-	if *prefillerUseTLS {
+	// Log deprecation warnings for old flags
```
elevran: shouldn't the pflag deprecation mark handle that automatically?

elevran: Also, try to avoid introducing linter bypasses unless there is no other reasonable avenue.

cmd/pd-sidecar/main.go
```diff
@@ -123,107 +57,73 @@ func main() {
...
+	logger.Info("p/d connector validated", "connector", opts.Connector)
+
+	if opts.EnableSSRFProtection {
+		logger.Info("SSRF protection enabled", "namespace", opts.InferencePoolNamespace, "poolName", opts.InferencePoolName)
 	}
 
 	// start reverse proxy HTTP server
 	scheme := "http"
-	if containsStage(*enableTLS, decodeStage) {
+	if opts.GetDecoderUseTLS() {
```
elevran: Any reason this can't be just direct access to an option's field?
The pattern is that you can separate having a flag field (e.g., the []string) from the actual configuration fields you set (e.g., `UseTLSForDecoder bool` or however you end up calling it).
The flag field is used for the CLI parsing and then converted to specific flags in `Complete()`.

cmd/pd-sidecar/main.go
```diff
@@ -123,107 +57,73 @@ func main() {
...
 	// Create SSRF protection validator
-	validator, err := proxy.NewAllowlistValidator(*enableSSRFProtection, *poolGroup, *inferencePoolNamespace, *inferencePoolName)
+	validator, err := proxy.NewAllowlistValidator(opts.EnableSSRFProtection, opts.PoolGroup, opts.InferencePoolNamespace, opts.InferencePoolName)
 	if err != nil {
 		logger.Error(err, "failed to create SSRF protection validator")
 		return
 	}
 
-	proxyServer := proxy.NewProxy(*port, targetURL, config)
+	proxyServer := proxy.NewProxy(opts.Port, targetURL, config)
```
hexfusion, line 119: question: port feels important why isn't it part of proxy.Config?

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,305 @@
+/*
+Copyright 2025 The llm-d Authors.
```
elevran:
```suggestion
Copyright 2026 The llm-d Authors.
```
Time sure flies :-)

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,305 @@
...
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/spf13/pflag"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+// Options holds all configuration options for the pd-sidecar proxy.
+type Options struct {
```
elevran, line 32: Prefer using `field // comment` in the same line.

cmd/pd-sidecar/main.go
```diff
@@ -123,107 +57,73 @@ func main() {
...
 	}
 
 	// start reverse proxy HTTP server
 	scheme := "http"
-	if containsStage(*enableTLS, decodeStage) {
+	if opts.GetDecoderUseTLS() {
 		scheme = "https"
 	}
-	targetURL, err := url.Parse(scheme + "://localhost:" + *vLLMPort)
+	targetURL, err := url.Parse(scheme + "://localhost:" + opts.VLLMPort)
```
elevran: the `targetURL` can also be an option field and set in `Complete()` as above line

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,305 @@
...
+	names := make([]string, 0, len(supportedTLSStages))
+	for stage := range supportedTLSStages {
+		names = append(names, stage)
+	}
+	return names
+}
+
+// containsStage checks if a stage is present in the slice
+func containsStage(stages []string, stage string) bool {
+	for _, s := range stages {
```
elevran, line 104: can't this be a lookup on a map instead of iterating on the slice? would you need a function then or can do it in place...?

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,305 @@
...
+	return false
+}
+
+// NewOptions returns a new Options struct initialized with default values.
+func NewOptions() *Options {
+	// Get default value for EnablePrefillerSampling from environment
+	enablePrefillerSampling := false
+	if val := os.Getenv("ENABLE_PREFILLER_SAMPLING"); val != "" {
+		enablePrefillerSampling, _ = strconv.ParseBool(val)
+	}
```
elevran:
```suggestion
	if val, err := strconv.ParseBool(os.Getenv("ENABLE_PREFILLER_SAMPLING")); err != nil {
		enablePrefillerSampling = val
	}
```

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,305 @@
...
+		EnablePrefillerSampling:     enablePrefillerSampling,
+		PoolGroup:                   DefaultPoolGroup,
+		LoggingOptions:              zap.Options{},
+	}
+}
+
+// AddFlags binds the Options fields to command-line flags on the given FlagSet.
+// It also binds logging flags to the standard flag.CommandLine FlagSet.
+func (opts *Options) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&opts.Port, "port", opts.Port,
```
elevran: nit: do not fold lines unless required, remove extra white space.

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,305 @@
...
+		"stages to enable TLS for. Supported: "+strings.Join(supportedTLSStagesNames(), ", ")+". Can be specified multiple times or as comma-separated values.")
+
+	fs.StringSliceVar(&opts.TLSInsecureSkipVerify, "tls-insecure-skip-verify", opts.TLSInsecureSkipVerify,
+		"stages to skip TLS verification for. Supported: "+strings.Join(supportedTLSStagesNames(), ", ")+". Can be specified multiple times or as comma-separated values.")
+
+	// Deprecated flags - kept for backward compatibility
+	fs.BoolVar(&opts.PrefillerUseTLS, "prefiller-use-tls", opts.PrefillerUseTLS,
+		"Deprecated: use --enable-tls=prefiller instead. Whether to use TLS when sending requests to prefillers.")
+
+	fs.BoolVar(&opts.DecoderUseTLS, "decoder-use-tls", opts.DecoderUseTLS,
```
elevran: please add `pflag.MarkDeprecated()` for the deprecated flags explicitly. It allows setting a message and automatically displays it to the use if the flags is set.

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,305 @@
...
+			return errors.New("--inference-pool-name or INFERENCE_POOL_NAME environment variable is required when --enable-ssrf-protection is true")
+		}
+	}
+
+	return nil
+}
+
+// Complete performs post-processing of parsed command-line arguments.
+// This handles migration from deprecated boolean flags to new StringSlice flags.
+func (opts *Options) Complete() error {
```
elevran, line 197: Please move `Complete()` before `Validate()` - as that's the order of calling.

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+	}
+
+	for _, stage := range opts.TLSInsecureSkipVerify {
+		if _, ok := supportedTLSStages[stage]; !ok {
+			return fmt.Errorf("--tls-insecure-skip-verify stages must be one of: %s", strings.Join(supportedTLSStagesNames(), ", "))
+		}
+	}
+
+	// Validate SSRF protection requirements
```
elevran: Please confirm these checks are not duplicated or logged by main as well.

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,305 @@
...
+
+	// Add logging flags to the standard flag set
+	opts.LoggingOptions.BindFlags(flag.CommandLine)
+}
+
+// Validate checks the Options for invalid or conflicting values.
+func (opts *Options) Validate() error {
+	// Validate connector
+	isValidConnector := false
+	for _, validConnector := range supportedConnectors {
```
elevran: map lookup?

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,305 @@
...
+		if opts.Connector == validConnector {
+			isValidConnector = true
+			break
+		}
+	}
+	if !isValidConnector {
+		return fmt.Errorf("--connector must be one of: %s", strings.Join(supportedConnectors, ", "))
+	}
+
+	// Validate TLS stages
```
elevran: See earlier comment on flag fields vs configuration fields. Convert the string slices of the flags into boolean configuration options

pkg/sidecar/proxy/options_test.go
```diff
@@ -0,0 +1,253 @@
...
+package proxy
+
+import (
+	"os"
+	"testing"
+
+	"github.com/spf13/pflag"
+)
+
+func TestNewOptions(t *testing.T) {
```
elevran: Not sure testing defaults make sense. If you change the default you will change the test as well.

pkg/sidecar/proxy/options_test.go
```diff
@@ -0,0 +1,253 @@
...
+		t.Error("Expected SecureProxy to be true")
+	}
+	if opts.PoolGroup != DefaultPoolGroup {
+		t.Errorf("Expected PoolGroup to be '%s', got '%s'", DefaultPoolGroup, opts.PoolGroup)
+	}
+}
+
+func TestNewOptionsWithEnvVars(t *testing.T) {
+	// Set environment variables
+	if err := os.Setenv("INFERENCE_POOL_NAMESPACE", "test-namespace"); err != nil {
```
elevran: consider using `t.Setenv(<var>, <value>)` instead

pkg/sidecar/proxy/options_test.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+	defer func() {
+		_ = os.Unsetenv("INFERENCE_POOL_NAMESPACE")
+		_ = os.Unsetenv("INFERENCE_POOL_NAME")
+		_ = os.Unsetenv("ENABLE_PREFILLER_SAMPLING")
+	}()
+
+	opts := NewOptions()
+
+	if opts.InferencePoolNamespace != "test-namespace" {
```
elevran, line 31: #685 will affect this (and all references for inference pool namespace and name). Consider waiting on it to merge so you can handle that correctly as well.

pkg/sidecar/proxy/options_test.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+	if opts.InferencePoolName != "test-pool" {
+		t.Errorf("Expected InferencePoolName to be 'test-pool', got '%s'", opts.InferencePoolName)
+	}
+	if !opts.EnablePrefillerSampling {
+		t.Error("Expected EnablePrefillerSampling to be true")
+	}
+}
+
+func TestAddFlags(t *testing.T) {
```
elevran: this seems to be a low value test.

pkg/sidecar/proxy/options_test.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+}
+
+func TestValidateConnector(t *testing.T) {
+	tests := []struct {
+		name      string
+		connector string
+		wantErr   bool
+	}{
+		{"valid nixlv2", ConnectorNIXLV2, false},
```
elevran: Please use explicit setup of field names and not rely on the order defined,

pkg/sidecar/proxy/options_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+	if !containsStage(opts.TLSInsecureSkipVerify, "prefiller") {
+		t.Error("Expected 'prefiller' to be in TLSInsecureSkipVerify after Complete()")
+	}
+	if !containsStage(opts.TLSInsecureSkipVerify, "decoder") {
+		t.Error("Expected 'decoder' to be in TLSInsecureSkipVerify after Complete()")
+	}
+}
+
+func TestGetPrefillerUseTLS(t *testing.T) {
```
elevran: Suggest combining the TLS string slice to bool into a single test (e.g., test none, one stage only, both stages, using comma separated or separate repeated flags, invalid values, etc.)

pkg/sidecar/proxy/options.go
Mohamedma96: it's checked here but logged in main- if EnableSSRFProtection is enabled, an info log is shown.

pkg/sidecar/proxy/options.go
Mohamedma96: done

pkg/sidecar/proxy/options_test.go
Mohamedma96: true, removed it.

pkg/sidecar/proxy/options_test.go
Mohamedma96: done

pkg/sidecar/proxy/options_test.go
Mohamedma96: Merged and considered the changes made there and worked on top of them.

pkg/sidecar/proxy/options_test.go
Mohamedma96: yes, I think it's actually redundant, removed it as well.

pkg/sidecar/proxy/options_test.go
Mohamedma96: done

pkg/sidecar/proxy/options.go
Mohamedma96: fixed, btw, what is the number of chars limit on the lines ?

pkg/sidecar/proxy/options.go
Mohamedma96: done

pkg/sidecar/proxy/options.go
Mohamedma96: it can, the map is basically 2-3 items, so it won't make much of a difference, but changed it to the right way, thanks

pkg/sidecar/proxy/options.go
Mohamedma96: done

pkg/sidecar/proxy/options.go
Mohamedma96: it does :( , fixed

cmd/pd-sidecar/main.go
Mohamedma96: removed this since MarkDeprecated was added

cmd/pd-sidecar/main.go
Mohamedma96: I also agree, but the port was already handled in main.go and I moved it to options with the other configs, we can make it part of proxy.Config, @elevran what do you think?

pkg/sidecar/proxy/options.go
Mohamedma96: done

pkg/sidecar/proxy/options.go
Mohamedma96: changed, done

cmd/pd-sidecar/main.go
Mohamedma96: done

cmd/pd-sidecar/main.go
Mohamedma96: no specific reason, I just continued what was done in the prevous PR for the string slices.

But I agree, changed it to how you proposed.

done

cmd/pd-sidecar/main.go
Mohamedma96: fixed

pkg/sidecar/proxy/options.go
Mohamedma96: done

pkg/sidecar/proxy/options_test.go
Mohamedma96: test enhanced and added a test case as you proposed.
Done

cmd/pd-sidecar/main.go
elevran: Agree with @hexfusion 
I think we could consider replacing the current Config with Options in a follow up PR.
It can either be a full replacement (remove Config) or keep Config and embed it into Options. Don't feel strongly either way (though embedding would separate the inetrmediate CLI flag fields from the actual configuration consumed by proxy which means we can pass just the Config subset).

cmd/pd-sidecar/main.go
```diff
@@ -133,100 +56,51 @@ func main() {
...
-	}
 
-	for _, stage := range *tlsInsecureSkipVerify {
-		if _, ok := supportedTLSStages[stage]; !ok {
-			logger.Info("Error: --tls-insecure-skip-verify stages must be one of: " + strings.Join(supportedTLSStagesNames(), ", "))
-			return
-		}
-	}
+	logger.Info("Proxy starting", "Built on", version.BuildRef, "From Git SHA", version.CommitSHA)
+	logger.Info("p/d KV connector validated", "connector", opts.Connector)
```
elevran: nit: any reason to special case the `Connector`? Perhaps just dump the entire actual Config used by the proxy?

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,289 @@
...
+	"strings"
+
+	"github.com/spf13/pflag"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+// Options holds all configuration options for the pd-sidecar proxy.
+type Options struct {
+	Port string // Port is the port the sidecar is listening on
+
```
elevran: nit: consider condensing by dropping the empty line between fields.

pkg/sidecar/proxy/options.go
elevran: Resolved? still using array, not map...

elevran: agree about the performance of iterating on 2-3 items being "zero", but if `containsStage` is only called in this file I would drop the function and just do a map lookup in its stead.

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,289 @@
...
+// NewOptions returns a new Options struct initialized with default values.
+func NewOptions() *Options {
+	// Get default value for EnablePrefillerSampling from environment
+	enablePrefillerSampling := false
+	if val, err := strconv.ParseBool(os.Getenv("ENABLE_PREFILLER_SAMPLING")); err == nil {
+		enablePrefillerSampling = val
+	}
+
+	return &Options{
```
elevran, line 120: nit: default value initializations (e.g., false bool, empty string) can be safely dropped from this, unless you prefer explicit settings.

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// AddFlags binds the Options fields to command-line flags on the given FlagSet.
+// It also sets up zap logging flags and integrates Go flags with pflag.
+func (opts *Options) AddFlags(fs *pflag.FlagSet) {
+	// Add logging flags to the standard flag set
+	opts.LoggingOptions.BindFlags(flag.CommandLine)
+
+	// Add Go flags to pflag (for zap options compatibility)
+	fs.AddGoFlagSet(flag.CommandLine)
```
elevran, line 143: Q: is there something similar done on main too?

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,289 @@
...
+
+	fs.StringVar(&opts.Port, "port", opts.Port, "the port the sidecar is listening on")
+
```
elevran: nit: please consider condensing. The empty lines are not needed.

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,289 @@
...
+	fs.StringVar(&opts.VLLMPort, "vllm-port", opts.VLLMPort, "the port vLLM is listening on")
+
+	fs.IntVar(&opts.DataParallelSize, "data-parallel-size", opts.DataParallelSize, "the vLLM DATA-PARALLEL-SIZE value")
+
+	fs.StringVar(&opts.Connector, "connector", opts.Connector, "the P/D connector being used. Supported: "+strings.Join(supportedKVConnectorsNames(), ", "))
+
+	fs.StringSliceVar(&opts.EnableTLS, "enable-tls", opts.EnableTLS, "stages to enable TLS for. Supported: "+strings.Join(supportedTLSStagesNames(), ", ")+". Can be specified multiple times or as comma-separated values.")
```
elevran: nit: line length of ~100-110 characters should fit fine on most monitors. We don't enforce limits, but would consider wrapping if you exceed that.

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,289 @@
...
+	// Add Go flags to pflag (for zap options compatibility)
+	fs.AddGoFlagSet(flag.CommandLine)
+
+	fs.StringVar(&opts.Port, "port", opts.Port, "the port the sidecar is listening on")
+
+	fs.StringVar(&opts.VLLMPort, "vllm-port", opts.VLLMPort, "the port vLLM is listening on")
+
+	fs.IntVar(&opts.DataParallelSize, "data-parallel-size", opts.DataParallelSize, "the vLLM DATA-PARALLEL-SIZE value")
+
+	fs.StringVar(&opts.Connector, "connector", opts.Connector, "the P/D connector being used. Supported: "+strings.Join(supportedKVConnectorsNames(), ", "))
```
elevran: nit: since `string.Join` is called multiple times, you could add it to the vars section `supportedKVConnectorsNames = strings.Join(...)`.  Same for `supportedTLSStagesNames` 

nit on nit: I think you only need one plural in English - `stageNames`, `connectorNames`...)

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+	return nil
+}
+
+// Complete performs post-processing of parsed command-line arguments.
+// This handles migration from deprecated boolean flags to new StringSlice flags,
+// sets configuration fields from flag fields, and computes the target URL.
+func (opts *Options) Complete() error {
+	// Migrate deprecated boolean TLS flags to new StringSlice flags
+	if opts.PrefillerUseTLS {
```
elevran, line 218: ultimately you should have a boolean for the configuration value.
the value can come from either command line option (new or deprecated).
I think the directionality of these is reversed (ie it should be from flags to configuration used by the proxy. This sets the slice and then extracts the value from the slice in main using `containsStage`, which seems reversed).
The code is kind of already doing part of that in L246-250.

It can be done in a separate PR that merges the proxy config and options.
Same applies to other string slices / boolean conversions.

pkg/sidecar/proxy/options.go
```diff
@@ -0,0 +1,289 @@
...
+		return err
+	}
+
+	if err := validateStages(opts.TLSInsecureSkipVerify, supportedTLSStages, "--tls-insecure-skip-verify"); err != nil {
+		return err
+	}
+
+	// Validate SSRF protection requirements
+	if opts.EnableSSRFProtection {
+		if opts.InferencePoolNamespace == "" {
```
elevran, line 304: this assumes that you did not set an empty namespace to "default" as in #684. We allow empty namespace to mean "default" namespace

Mohamedma96: why not? in complete(), if I got opts.InferencePool = "Something", without /, then we set opts.InferencePoolNamespace = "default"
And in validate we make sure we got the namespace from somewhere, either inferencepool flag as explained above, or if it was set via env variable as before (for the to be deprecated InferencePoolNamespace)

pkg/sidecar/proxy/options.go
Mohamedma96: done and done

pkg/sidecar/proxy/options.go
Mohamedma96: ok done, I appreciate doing so, but I think there should be a makefile command that lints and formats the code down to the level of newlines, spaces/tabs, line length, etc.. is there a command other than make lint? I run make lint but it shows no issue for this case.

pkg/sidecar/proxy/options.go
Mohamedma96: Here is what is done, let me know if I am getting something wrong:

in AddFlags():
Zap adds flags to flag.CommandLine - with BindFlags().
we copy flag.CommandLine to pflag.CommandLine - with AddGoFlagSet()

in main:
we call AddfFlags() explained above and then we parse pflag.CommandLine

pkg/sidecar/proxy/options.go
Mohamedma96: done

pkg/sidecar/proxy/options.go
Mohamedma96: done

cmd/pd-sidecar/main.go
Mohamedma96: These logs were before and I kept them,  I added a log  for the entire config

pkg/sidecar/proxy/options.go
elevran: `gofmt` automatically compresses a **series** of empty lines down to a maximum of one empty line to maintain a consistent style, and it removes leading/trailing blank lines in function bodies. Thus the single empty line is not being condensed.

elevran: So it is a stylistic preference, not something mandated by code convention tooling.

pkg/sidecar/proxy/options.go
elevran: ok, I probably mixed up different PRs or iterations.

pkg/sidecar/proxy/options.go
elevran: fair point. It can be done in `Complete()`

cmd/pd-sidecar/main.go
```diff
@@ -136,129 +56,63 @@ func main() {
...
+		PrefillerUseTLS:             opts.UseTLSForPrefiller,
+		PrefillerInsecureSkipVerify: opts.InsecureSkipVerifyForPrefiller,
+		DecoderInsecureSkipVerify:   opts.InsecureSkipVerifyForDecoder,
+		DataParallelSize:            opts.DataParallelSize,
+		EnablePrefillerSampling:     opts.EnablePrefillerSampling,
+		SecureServing:               opts.SecureProxy,
+		CertPath:                    opts.CertPath,
+	}
+
+	logger.Info("Proxy configuration",
```
elevran, line 94: nit: or maybe have a method on Options to print it?
It can be `Stringer` interface (e.g., if you want to filter out some fields and control the output), or using something like the below once you combine Options and Config (**different PR**).

```go
    cfg, err := o.Config() // convert options to Config required by the server
    if err != nil {
        return err
    }
    
    settings, _ := json.Marshal(cfg) // or can use MarshalIndent for readability, can also use struct tags on Config
    log.Info("Starting with configuration: %s", string(settings))

    server.RunWith(cfg)
```

PR #701, opened 2026-03-10
Optimize TTFT: send first token immediately after prefill for streaming

RishabhSaini:
Reduces Time To First Token (TTFT) for streaming clients by sending the first token immediately after prefill completes, before KV cache transfer to decode.

  - Convert non-streaming prefill response to SSE and forward first token to streaming clients
  - Fix token budget: decrement max_tokens by 1 for decode stage
  - Fix TTFT metrics to measure when first token actually reaches user (streaming vs non-streaming)
  - Strip internal kv_transfer_params from user-facing responses
  - Add headersSentWriter to prevent duplicate WriteHeader errors

### DIFF
--- a/pkg/sidecar/proxy/connector_nixlv2.go
+++ b/pkg/sidecar/proxy/connector_nixlv2.go
@@ -86,12 +86,8 @@ func (s *Server) runNIXLProtocolV2(w http.ResponseWriter, r *http.Request, prefi
 	maxCompletionTokensValue, maxCompletionTokensOk := completionRequest[requestFieldMaxCompletionTokens]
 
 	// Determine if client wants streaming
-	clientWantsStreaming := false
-	if streamOk {
-		if streamBool, ok := streamValue.(bool); ok {
-			clientWantsStreaming = streamBool
-		}
-	}
+	streamBool, streamBoolOk := streamValue.(bool)
+	clientWantsStreaming := streamOk && streamBoolOk && streamBool
 
 	completionRequest[requestFieldKVTransferParams] = map[string]any{
 		requestFieldDoRemoteDecode:  true,
@@ -182,7 +178,7 @@ func (s *Server) runNIXLProtocolV2(w http.ResponseWriter, r *http.Request, prefi
 
 	// 4. Send first token to client if streaming
 	var firstTokenSentTime time.Time
-	var decodeResponseWriter http.ResponseWriter = w
+	decodeResponseWriter := w
 	if clientWantsStreaming {
 		// Convert non-streaming response to SSE format and send first chunk
 		// Remove kv_transfer_params before sending to user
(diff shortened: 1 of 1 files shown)


pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -85,6 +85,14 @@ func (s *Server) runNIXLProtocolV2(w http.ResponseWriter, r *http.Request, prefi
 	maxTokensValue, maxTokensOk := completionRequest[requestFieldMaxTokens]
 	maxCompletionTokensValue, maxCompletionTokensOk := completionRequest[requestFieldMaxCompletionTokens]
 
+	// Determine if client wants streaming
+	clientWantsStreaming := false
+	if streamOk {
+		if streamBool, ok := streamValue.(bool); ok {
+			clientWantsStreaming = streamBool
```
kfswain: These conditionals seem unnecessary, cant you just `streamBool && streamOk`?
And beyond that, if we need two bools (one to validate is streaming is capable and one to validate if the client has opted in). I would think that validation would occur when `streamBool` is set.

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -172,6 +180,54 @@ func (s *Server) runNIXLProtocolV2(w http.ResponseWriter, r *http.Request, prefi
 
 	s.logger.V(5).Info("received prefiller response", requestFieldKVTransferParams, pKVTransferParams)
 
+	// 4. Send first token to client if streaming
+	var firstTokenSentTime time.Time
+	var decodeResponseWriter http.ResponseWriter = w
+	if clientWantsStreaming {
```
kfswain, line 189: Any reason not to do this in a goroutine?

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -204,11 +260,21 @@ func (s *Server) runNIXLProtocolV2(w http.ResponseWriter, r *http.Request, prefi
 	}
 	delete(completionRequest, requestFieldMaxTokens)
 	if maxTokensOk {
-		completionRequest[requestFieldMaxTokens] = maxTokensValue
+		// Decrement by 1 since prefill already generated 1 token
+		if val, ok := maxTokensValue.(float64); ok && val > 0 {
```
kfswain: This is outside of the conditional on line 186, right? Wouldn't this decrement for _all_ cases (even streaming off)?

RishabhSaini: agreed, updated

pkg/sidecar/proxy/connector_nixlv2.go
RishabhSaini: agreed, updated to a single expression

pkg/sidecar/proxy/connector_nixlv2.go
RishabhSaini: There needs to be synchronization around preventing decode from writing before first prefill token flush. Essentially the "highly unlikely" degenerate case where decode instance responds before first token send is complete creating ordering issues

kfswain: A channel should get you there, right? Your prefiller goroutine can send the signal on completion, and your decode routine can wait to send tokens until that signal is recieved, but still allow the two to run in parallel until the coordination point

RishabhSaini: agreed, updated

PR #707, opened 2026-03-11
Add program aware plugin

praveingk:
This PR introduces **program-aware plugin** would enable identifying a request from an agentic program based on its program-id (_x-gateway-inference-fairness-id_)  and performing scheduling decisions based on program-level metrics. Additionally, this plugin captures program-level metrics and exports to the prometheus endpoint.

Program aware plugin implements following plugin interfaces:
1) **Prepare Data Interface:** Extracts program information from request headers, subsequently updates the relevant program metrics and request metadata. The plugin assumes that a request arrives with a fairness ID (x-gateway-inference-fairness-id) to identify an agentic program.

2) **Fairness Interface (Flow Control):** We implement flow-control fairness plugin's `Pick` interface. Here we enable multiple strategies which can be configured. Currently, we are implemented two strategies, a simple EWMA based, and Deficit Round Robin (DRR).

3) **Pre Request Interface:** Updates program metrics immediately before dispatch. For example, this hook is used to calculate the wait time (time spent in EPP/queue) of requests per program, and keep track of requests sent to the vLLM inference pod. In future, this hook could also be used to add a vLLM priority to the request. 

4) **Response Received Interface:** Updates program metrics like deficit counters (for DRR) for tokens used.

We are currently evaluating the scheduling strategies using [inference-perf](https://github.com/kubernetes-sigs/inference-perf) benchmarks.


pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
```diff
@@ -0,0 +1,196 @@
...
+func (p *ProgramAwarePlugin) NewState(_ context.Context) any {
+	return nil
+}
+
+// Pick selects which program queue to service next.
+//
+// For each queue in the band, the configured ScoringStrategy is given a chance
+// to update its per-program state (OnPickStart), then the queue with the highest
+// score is selected for dispatch.
+func (p *ProgramAwarePlugin) Pick(_ context.Context, band flowcontrol.PriorityBandAccessor) (flowcontrol.FlowQueueAccessor, error) {
```
hexfusion, line 271: question: while it's true you have no consumers of the ctx in this method. should we perform this work if ctx.Err()?

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
```diff
@@ -0,0 +1,196 @@
...
+	start := time.Now()
+	defer func() {
+		pickLatencyUs.Observe(float64(time.Since(start).Microseconds()))
+	}()
+
+	if band == nil {
+		return nil, nil //nolint:nilnil
+	}
```
hexfusion, line 279: question: nil band is not an error? this means the caller even when the err == nil must nil check the return is nil. also won't this give inaccurate latency metrics for work that was not done?

pkg/plugins/program-aware/programmetrics.go
```diff
@@ -0,0 +1,103 @@
+package programaware
+
+import (
+	"sync"
+	"sync/atomic"
+)
+
+const ewmaAlpha = 0.2
```
hexfusion: question: how did you arrive at 0.2 for the EWMA alpha? is this based on benchmarking, or should it be configurable via the plugin config alongside strategy?

praveingk: We are still experimenting with the parameters. Ideally, we can make it a plugin config once we finalize the programmable parameters for each strategy.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
D-Sai-Venkatesh: We return (nil, nil) instead of (nil, err) because this is how the [contract](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/af2fac7acbd17b57580d2971f8750ba1c4530e78/pkg/epp/flowcontrol/framework/plugins/fairness/functional_test.go#L90) for fairness policy has been defined. The existing Fairness plugins also follow the same convention for example, [round-robin](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/af2fac7acbd17b57580d2971f8750ba1c4530e78/pkg/epp/flowcontrol/framework/plugins/fairness/roundrobin.go#L76). 

From the behaviour point of view, both cases have same behaviour, that is no item will be dispatched from that band. But in error case the error will be logged.

praveingk:
@elevran @nilig @vMaroon  @hexfusion  This is ready for review. 
We have implemented three scheduling strategies : 1) Round-Robin, 2) Least Attained Service, and 3) Deficit Round-Robin.  A brief overview and micro-benchmarks are documented in the [proposal doc](https://docs.google.com/document/d/1Pc9TjZolaJtxg2hLpP8PKG4LmZ98qtP72IzdcxAdDsw/edit?tab=t.0#heading=h.hwlyfrbgmgft)

We are conducting inference-perf based benchmarks using HotPotQA traces and Otel replay.  However, we can proceed with this PR, since we have validated correctness and performance in a small-scale setup.

praveingk:
@elevran @nilig @ahg-g  : We have resolved the conflicts after the GAIE migration, and now this PR is ready to be looked at.
We performed some benchmarks with agentic traces (MCTS based on HotPotQA) replayed using inference-perf. The evaluation report can be accessed [here](https://docs.google.com/document/d/1Pc9TjZolaJtxg2hLpP8PKG4LmZ98qtP72IzdcxAdDsw/edit?tab=t.0)

TLDR; We observe program latency reduction of  10-21 % for LAS (Least Attained Service) & DRR (Deficit Round Robin) for light and medium programs compared to naive Round Robin scheduling. For heavy programs, we observe close to 10% reduction in program latency for both LAS and DRR.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
```diff
@@ -0,0 +1,260 @@
...
+// Program identity comes from the x-gateway-inference-fairness-id request header.
+//
+//nolint:revive
+type ProgramAwarePlugin struct {
+	name     string
+	strategy ScoringStrategy
+
+	// programMetrics stores aggregated metrics per program.
+	// Key: program ID (string), Value: *ProgramMetrics.
+	programMetrics sync.Map
```
roytman, line 238: `ProgramMetrics` are created in `getOrCreateMetrics` but never evicted. In a long-running scheduler with many short-lived programs, this map will grow indefinitely.

Suggestion: Add a TTL-based eviction or periodic cleanup of programs with no recent activity.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// requestTimestamps tracks when Pick() dispatched each request,
+	// used to compute flow-control queue wait time in PreRequest.
+	// Key: request ID (string), Value: time.Time.
+	requestTimestamps sync.Map
```
roytman: A similar comment. 
If a request is picked but never completes (e.g., connection dropped, pod eviction), the timestamp entry is never cleaned up. Only `ResponseBody` deletes the entry.

Suggestion: Add a periodic sweep or TTL for stale entries in requestTimestamps.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/programmetrics.go
```diff
@@ -0,0 +1,255 @@
...
+// Deficit returns the current deficit counter value in tokens.
+func (m *ProgramMetrics) Deficit() int64 {
+	return m.deficitTokens.Load()
+}
+
+// DecayDeficitTimed applies time-based exponential decay to the deficit counter
+// using a half-life. The decay factor is 0.5^(elapsed/halfLifeSeconds), so the
+// deficit halves every halfLifeSeconds regardless of call frequency.
+func (m *ProgramMetrics) DecayDeficitTimed(halfLifeSeconds float64, now time.Time) {
+	m.mu.Lock()
```
roytman: `DecayDeficitTimed` locks `mu` to protect `lastDeficitDecay`, but the calls `m.deficitTokens.Load()` and `m.deficitTokens.Store()` outside the atomic operation semantics. While each individual call is atomic, the read-modify-write sequence (Load, multiply, Store) is not atomic with respect to concurrent AddDeficit or DeductTokens calls. A concurrent `AddDeficit` between Load and Store would be lost.

Possible suggestions:

- Make AddDeficit/DeductTokens also acquire mu - correct, but adds lock contention on the hot path.
- Use a compare-and-swap loop in DecayDeficitTimed - keeps the writers lock-free, see below.
- Do nothing - since decay is an approximation anyway (the factor is based on wall-clock time, rounding to int64 loses precision), the occasional lost update may be tolerable. But it's worth noting because under high concurrency, the lost updates compound, and in pathological cases, a program could retain stale credit it shouldn't have.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy.go
```diff
@@ -0,0 +1,477 @@
...
+// QueueInfo bundles read-only data for each queue passed to Pick.
+type QueueInfo struct {
+	Queue   flowcontrol.FlowQueueAccessor
+	Metrics *ProgramMetrics
+	Len     int
+}
+
+// newStrategy constructs a ScoringStrategy from the plugin config.
+func newStrategy(cfg Config) (ScoringStrategy, error) {
+	switch cfg.Strategy {
```
roytman: No validation on config values like:

- weightService + weightServiceHeadWait should sum to 1.0 (I'm not sure about it)
- quantumTokens should be > 0
- serviceDecayFactor should be in (0, 1)
- serviceHalfLifeSeconds should be > 0 when set

Suggestion: Add validation in newStrategy() and return errors for invalid configurations.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
```diff
@@ -0,0 +1,260 @@
...
+const (
+	// ProgramAwarePluginType is the registered type name for this plugin.
+	ProgramAwarePluginType = "program-aware-fairness"
+
+	// fairnessIDHeader is the standard header used to identify the program.
+	fairnessIDHeader = "x-gateway-inference-fairness-id"
+
+	// defaultFairnessID is the flow key assigned by the upstream framework when
+	// no x-gateway-inference-fairness-id header is present on the request.
+	// Matches the constant in github.com/llm-d/llm-d-inference-scheduler/pkg/epp/handlers/request.go.
```
roytman: Now, after the repositories merge, the constant is in `pkg/epp/metadata/consts.go`. Should we just use it instead of redefining it?

pkg/metrics/metrics.go
```diff
@@ -49,7 +50,12 @@ var (
 
 // GetCollectors returns all custom collectors for the llm-d-inference-scheduler.
 func GetCollectors() []prometheus.Collector {
-	return []prometheus.Collector{SchedulerPDDecisionCount, SchedulerDisaggDecisionCount}
+	base := []prometheus.Collector{SchedulerPDDecisionCount, SchedulerDisaggDecisionCount}
+	pa := programaware.GetCollectors()
```
roytman: nit: unconditionally registers all program-aware metrics even when the plugin isn't configured, adding overhead to Prometheus scrapes in deployments that don't use this plugin.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/programmetrics.go
```diff
@@ -0,0 +1,255 @@
...
+		return
+	}
+	elapsed := now.Sub(m.lastDeficitDecay).Seconds()
+	if elapsed <= 0 {
+		return
+	}
+	factor := math.Pow(0.5, elapsed/halfLifeSeconds)
+	current := m.deficitTokens.Load()
+	decayed := int64(float64(current) * factor)
+	m.deficitTokens.Store(decayed)
```
roytman: The compare-and-swap loop from the suggestion above 
```suggestion
    for {
	    current := m.deficitTokens.Load()
	    decayed := int64(float64(current) * factor)
	    if m.deficitTokens.CompareAndSwap(current, decayed) {
            break
        }
```

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/programmetrics.go
praveingk: Thanks @roytman for the suggestions. This is a good catch, and I am leaning towards the 2nd option which is very simple;  1st option increases contention, and the last option (current implementation) might decrease fairness under high load.

praveingk:
Thanks @roytman for the review. Will work on addressing them.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
```diff
@@ -0,0 +1,235 @@
+# Program-Aware Fairness Plugin
+
+The program-aware fairness plugin introduces program-level scheduling to `llm-d-inference-scheduler`. Rather than treating each inference request independently, this plugin recognizes that requests belong to higher-level **agentic programs** — workflows composed of multiple LLM calls — and makes scheduling decisions based on aggregated program-level metrics.
```
revit13: We have an ongoing effort to standardize the plugin README formats. Could you please update this README to align with the template outlined in [issue #917](https://github.com/llm-d/llm-d-inference-scheduler/issues/917)? Thanks

praveingk: Definitely @revit13 👍🏽

roytman:
You probably will need an e2e test(s)

elevran:
question: what identifies a stream of requests / responses as being part of a specific "program"?
Consider:
- same user, multiple sessions
- same session, sub-agents
- etc.

How would we identify related requests and responses that are under the policy?
I do not think that using _x-gateway-inference-fairness-id_ HTTP header is a common practice of agentic programs.
There should be code (ideally as part of existing plugin points) to provide the session, agent and subagent-id, type of work (e.g., `plan`). Then "downstream" plugins such as this one can consume that information for optimization purposes.

praveingk:
> question: what identifies a stream of requests / responses as being part of a specific "program"? Is the **program-id** ( the _x-gateway-inference-fairness-id_ HTTP header) a common practice of agentic programs? Consider:
> 
> * same user, multiple sessions
> * same session, sub-agents
> * etc.
>   How would we identify related requests and responses that are under the policy?

@elevran Valid point. Initially, we wanted to focus on the scheduler, Hence we assumed a particular header as an abstraction and demonstrate that the scheduler can deliver performance benefits to agentic workloads.

I have just create an issue to bridge the gap you have mentioned to introduce a plugin to provide normalized agent ID across providers https://github.com/llm-d/llm-d-inference-scheduler/issues/1031

This can be consumed by fairness plugin (this PR), and in future filter/scorer plugin enhancements.

For now, we will work on the suggestions by @roytman and @revit13 in this PR, but hold off on merging until we have a uniform agent ID support.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
```diff
@@ -0,0 +1,260 @@
...
+	_ requestcontrol.PreRequest            = &ProgramAwarePlugin{}
+	_ requestcontrol.ResponseBodyProcessor = &ProgramAwarePlugin{}
+)
+
+// ProgramAwarePluginFactory creates a new ProgramAwarePlugin from JSON config.
+// Example config: {"strategy": "drr"}
+//
+//nolint:revive
+func ProgramAwarePluginFactory(name string, rawCfg json.RawMessage, _ plugin.Handle) (plugin.Plugin, error) {
+	cfg := Config{Strategy: "las"}
```
hexfusion: should we define the DefaultConfig() so we can ensure this is stable over time?

```suggestion
 cfg := DefaultConfig()
```

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
```diff
@@ -0,0 +1,260 @@
...
+	_ flowcontrol.FairnessPolicy           = &ProgramAwarePlugin{}
+	_ requestcontrol.DataProducer          = &ProgramAwarePlugin{}
+	_ requestcontrol.PreRequest            = &ProgramAwarePlugin{}
+	_ requestcontrol.ResponseBodyProcessor = &ProgramAwarePlugin{}
+)
+
+// ProgramAwarePluginFactory creates a new ProgramAwarePlugin from JSON config.
+// Example config: {"strategy": "drr"}
+//
+//nolint:revive
```
hexfusion: since we are in programaware pkg the linter is trying to remove the echo.  its probably good practice to give a comment why we circumventing the linter.

```gp
p, err := programaware.ProgramAwarePluginFactory()
vs 
p, err := programaware.PluginFactory()
```

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// computeFairnessIndex returns Jain's Fairness Index over the service rate
+// (weighted tokens/sec) for each program. Equal service rates = perfect fairness.
+// Returns 1.0 when fewer than 2 programs have rate data.
+func (p *ProgramAwarePlugin) computeFairnessIndex() float64 {
+	var sum, sumSq float64
+	var n float64
+	p.programMetrics.Range(func(_, value any) bool {
+		m := value.(*ProgramMetrics)
```
hexfusion: we bake assumptions that value type assertion will always be ok. but I think we should always be a bit more defensive.

```suggestion
        m, ok := value.(*ProgramMetrics)
```

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/prometheus.go
```diff
@@ -0,0 +1,130 @@
...
+	outputTokensTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: programAwareSubsystem,
+			Name:      "output_tokens_total",
+			Help:      metricsutil.HelpMsgWithStability("Total output (completion) tokens produced per program", compbasemetrics.ALPHA),
+		},
+		[]string{"program_id"},
+	)
+
+	pickLatencyUs = prometheus.NewHistogram(
```
hexfusion, line 50: we need to be very careful on the cardinality of the metrics we are adding here.  histograms emit buckets + 3 series per label set. with 11 buckets and program_id as a label, that's 14 series per program, vs 1 series per program for each of the gauges/counters. the plugin already exports ewmaWaitTimeMs as a gauge that gives per-program average at 1/14th the cost. please review the overhead here.

ahg-g:
A couple of questions/comments:
1) Do we have an issue to document the goals of this PR?
2) https://github.com/llm-d/llm-d-router/pull/1244 merged, I assume it was needed by this.

praveingk:
> A couple of questions/comments:
> 
> 1. Do we have an issue to document the goals of this PR?

We haven't created an issue regarding this PR. It was mostly tracked by this [proposal doc](https://docs.google.com/document/d/1Pc9TjZolaJtxg2hLpP8PKG4LmZ98qtP72IzdcxAdDsw/edit?tab=t.0#heading=h.hwlyfrbgmgft) . I can create one issue to document. 
> 2. [Pre admission processor #1244](https://github.com/llm-d/llm-d-router/pull/1244) merged, I assume it was needed by this.

Yes.

ahg-g:
Yes pls, let's create an issue. Is this planned for the next release? June 14th is the code freeze.

Also, do we have evaluations showing the impact of this approach?

praveingk:
@ahg-g : Yes, we can make it by June 14, The main prerequisite Issue #1031 
Since, that is done, We will work on addressing the comments. 
Regarding evaluation, We performed some benchmarks with agentic traces (MCTS based on HotPotQA) replayed using inference-perf.

We observe program latency reduction of  10-24 % for LAS & 10-16% for  DRR for ultra-light and light programs compared to naive Round Robin scheduling. For medium and heavy programs, both LAS and DRR deliver consistent ~10-11% reduction in program latency. We observe LAS consistently outperforms DRR with the gap being more pronounced for ultra-light programs

The More details on the  evaluation can be accessed [here](https://docs.google.com/document/d/1Pc9TjZolaJtxg2hLpP8PKG4LmZ98qtP72IzdcxAdDsw/edit?tab=t.0#heading=h.hwlyfrbgmgft)

We are now in the process of conducting large scale evaluation using the Otel trace replay [published](https://huggingface.co/datasets/Exgentic/agent-llm-traces). However, that is not a dependency, and the PR can progress in parallel.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/programmetrics.go
D-Sai-Venkatesh: Thanks @roytman — the deficit state moved off `ProgramMetrics` into a per-strategy [`drrState`](https://github.com/llm-d/llm-d-inference-scheduler/blob/e6ac876a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_drr.go#L21-L25) struct with its own mutex. Pick's [decay branch](https://github.com/llm-d/llm-d-inference-scheduler/blob/e6ac876a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_drr.go#L104-L118) and [`OnCompleted`](https://github.com/llm-d/llm-d-inference-scheduler/blob/e6ac876a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_drr.go#L190-L192) both hold `st.mu`, so the lost-update race should not be present now.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
D-Sai-Venkatesh: added a TTL sweep. A background goroutine ([`runEviction`](https://github.com/llm-d/llm-d-inference-scheduler/blob/e6ac876a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go#L345-L358)) runs on a timer and [`evictIdle`](https://github.com/llm-d/llm-d-inference-scheduler/blob/e6ac876a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go#L360-L411) drops the metrics entry, strategy state, and metric labels for programs idle past TTL — so the map and metric cardinality stay bounded. Defaults: TTL 3600s, sweep every 300s ([`plugin.go:125-126`](https://github.com/llm-d/llm-d-inference-scheduler/blob/e6ac876a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go#L125-L126)).

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
D-Sai-Venkatesh: Side map is gone. Enqueue time is stored on the request's attribute map ([`plugin.go:317`](https://github.com/llm-d/llm-d-inference-scheduler/blob/e6ac876a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go#L317), read in [`request_hooks.go:71`](https://github.com/llm-d/llm-d-inference-scheduler/blob/e6ac876a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/request_hooks.go#L71)), so it's freed when the request is dropped — no leak even if the request never completes.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy.go
D-Sai-Venkatesh: Config validation added in [`Config.validate()`](https://github.com/llm-d/llm-d-inference-scheduler/blob/e6ac876a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go#L132-L166): weights must be ≥ 0, `quantumTokens > 0`, decay factors in their valid ranges, half-lives ≥ 0, eviction values bounded. Called from the factory at [`plugin.go:191`](https://github.com/llm-d/llm-d-inference-scheduler/blob/e6ac876a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go#L191) so bad config fails plugin construction.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
D-Sai-Venkatesh: Done — using `metadata.DefaultFairnessID` (e.g. [`request_hooks.go:35`](https://github.com/llm-d/llm-d-inference-scheduler/blob/e6ac876a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/request_hooks.go#L35), and other call sites).

pkg/metrics/metrics.go
D-Sai-Venkatesh: Registration moved into the plugin factory ([`plugin.go:202-212`](https://github.com/llm-d/llm-d-inference-scheduler/blob/e6ac876a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go#L202-L212)) — collectors are only registered when the plugin is actually instantiated, so deployments that don't configure it pay nothing.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
D-Sai-Venkatesh: Tried to align the [README](https://github.com/llm-d/llm-d-inference-scheduler/blob/e6ac876a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md) with the #917 template — has the title/metadata, What It Does, How It Works, Inputs Consumed, Configuration (with parameter table and examples), Limitations, Related Documentation. Please let us know if anything else is off.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
D-Sai-Venkatesh: Done — `DefaultConfig` is now a function returning a fresh `Config` value ([`plugin.go:115`](https://github.com/llm-d/llm-d-inference-scheduler/blob/405ee1d0/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go#L115)). Updated the call sites accordingly.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
D-Sai-Venkatesh: added a rationale on both the docstring and the `//nolint` directive ([`plugin.go:180-186`](https://github.com/llm-d/llm-d-inference-scheduler/blob/405ee1d0/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go#L180-L186)). The factory name matches sibling fairness plugins (`roundrobin.RoundRobinFairnessPolicyFactory`, `globalstrict.GlobalStrictFairnessPolicyFactory`).

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
D-Sai-Venkatesh: `*ProgramMetrics` assertions use comma-ok ([`plugin.go:389`](https://github.com/llm-d/llm-d-inference-scheduler/blob/f1621a6d/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go#L389)), and same for the strategy `sync.Map` loaders.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/prometheus.go
D-Sai-Venkatesh: dropped the per-program wait-time histogram.

ahg-g:
It seems the rebase is completely off, it is not clear what logic this PR adds

D-Sai-Venkatesh:
hey @ahg-g, as this is a long running PR, the commit history for this PR grew very large. Due to this for last few weeks we are using merge instead of rebase, that is why we are seeing few merge commits in the PR. Even though commit history is huge the diff itself is contained to a single folder for the plugin and 1 change in runner.go and a sample config yaml. Happy to walk through the diff if that helps and [README](https://github.com/praveingk/llm-d-inference-scheduler/blob/program-aware-plugin/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md) contains the info regarding the plugin. In the last merge (main -> this pr branch) approximately 49 commits were merged which included movement of flow control plugin resolution into the config loader, addition of support for multiple parsers and ci/docs fixes, these commits did not affect program aware plugin's behavior.

Please let us know if any further clarification/information is required from our side.

ahg-g:
If the code is fairly contained, why not do a rebase then? What is the best way to view the diff compared to main right now?

praveingk:
Hi @ahg-g , We had opted to merge since @D-Sai-Venkatesh and I were making changes together, and to preserve the existing reviews so that we could keep them in context while we address.
The entire diff compared to main can be viewed here : https://github.com/llm-d/llm-d-router/pull/707/changes

Overall, This PR's changes are in `pkg/epp/framework/plugins/flowcontrol/fairness/program-aware` and it introduces three strategies to perform program-aware scheduling: 1) Round-robin (`strategy_rr.go`), 2) LAS (Least Attained Service) (`strategy_las.go`) , and 3) DRR (Deficit Round Robin) (`strategy_drr.go`). Each program's requests are maintained in a separate queue based on the fairness-id (populated by the agent identity plugin - introduced in https://github.com/llm-d/llm-d-router/pull/1244).
`Plugin.go` acts as the entry-point of the plugin, implementing the flow controller's `Pick` logic to select which queue (program) to dispatch an item from. It internally invokes the one of the above strategies based on the configuration.
We maintain common program metrics in `programmetrics.go`. `requesthooks.go` additionally implements plugins for PreRequest, and Response to update metrics appropriately.

ahg-g:
Sorry, this is not our standard review process, the reviewer needs to clearly see the diff and what it will eventually get merged.

ahg-g:
One more thing, the PR is fairly large, can this be broken into smaller incremental PRs pls

D-Sai-Venkatesh:
yes @ahg-g, we can split this PR into smaller PRs. Will tag the new PR here once it is ready.

D-Sai-Venkatesh:
hey @ahg-g, created #1534 which is the first of 2 PRs splitting this PR. Please take a look and let us know your comments.

praveingk:
Closing this PR, since it's being broken down into a smaller PR in https://github.com/llm-d/llm-d-router/pull/1534
@roytman @hexfusion Thanks a lot for your comments,  We have addressed them in #1534

PR #714, opened 2026-03-12
Enable major version updates to gh actions

elevran:
Update the dependabot configuration to allow major semver updates to GH actions. Manually upgrade docker/setup-buildx-action to v4 until the update cycle runs.


hexfusion (commented):
LGTM while major bumps can break that would be visible in CI.

PR #715, opened 2026-03-12
fix(test): Increase test coverage for prefix_based_pd_decider.go

gyliu513:


pkg/plugins/profile/prefix_based_pd_decider_test.go
```diff
@@ -0,0 +1,288 @@
...
+
+func (n *notPrefixCacheMatchInfo) Clone() fwkdl.Cloneable { return &notPrefixCacheMatchInfo{} }
+
+const (
+	testTotalTokens = 10
+	testBlockSize   = 1
+)
+
+func makeTestEndpoint(cachedTokens int) scheduling.Endpoint {
+	ep := scheduling.NewEndpoint(
```
elevran: Q: would a datalayer.NewEndpoint satisfy the expected interface?
It would set up the attributes for you.

pkg/plugins/profile/prefix_based_pd_decider_test.go
```diff
@@ -0,0 +1,288 @@
...
+
+func TestDisaggregate(t *testing.T) {
+	ctx := utils.NewTestContext(t)
+
+	tests := []struct {
+		name            string
+		nonCachedTokens int
+		inputTokens     int
+		endpoint        scheduling.Endpoint
+		expected        bool
```
elevran: nit: what's expected?

pkg/plugins/profile/prefix_based_pd_decider_test.go
```diff
@@ -0,0 +1,288 @@
...
+			result := decider.disaggregate(ctx, tt.inputTokens, tt.endpoint)
+			assert.Equal(t, tt.expected, result)
+		})
+	}
+}
+
+func TestDisaggregateNoPrefixInfo(t *testing.T) {
+	ctx := utils.NewTestContext(t)
+
+	ep := scheduling.NewEndpoint(
```
elevran: nit: can use `makeTestEndpoint`? same in next test function

pkg/plugins/profile/prefix_based_pd_decider_test.go
gyliu513: Done, updated to `expectDisaggregate`

pkg/plugins/profile/prefix_based_pd_decider_test.go
gyliu513: Done

PR #717, opened 2026-03-14
Check for uniqueness of media URLs

roytman:
User requests can occasionally include redundant references to identical multimedia content. This PR ensures that media URLs are unique before they are dispatched to encoders to avoid unnecessary processing.

Based on the comment https://github.com/llm-d/llm-d-inference-scheduler/pull/643#discussion_r2902078093

@elevran ^^

### DIFF
--- a/pkg/sidecar/proxy/connector_epd_shared_storage.go
+++ b/pkg/sidecar/proxy/connector_epd_shared_storage.go
@@ -130,15 +130,15 @@ func (s *Server) fanoutEncoderPrimer(originalRequest map[string]any, encoderHost
 	}
 
 	// Deduplicate URL-based items; keep all non-URL items (e.g. inline audio).
-	seenURLs := make(map[string]bool)
+	seenURLs := make(map[string]struct{})
 	var mmItems []map[string]any
 	for _, item := range allItems {
 		if url := mmItemURL(item); url != "" {
-			if seenURLs[url] {
+			if _, seen := seenURLs[url]; seen {
 				s.logger.V(4).Info("skipping duplicate multimodal URL", "url", url, "requestID", requestID)
 				continue
 			}
-			seenURLs[url] = true
+			seenURLs[url] = struct{}{}
 		}
 		mmItems = append(mmItems, item)
 	}
--- a/pkg/sidecar/proxy/connector_epd_shared_storage_test.go
+++ b/pkg/sidecar/proxy/connector_epd_shared_storage_test.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"sigs.k8s.io/controller-runtime/pkg/log"
-	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
 func TestExtractMMItems(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/sidecar/proxy/connector_epd_shared_storage_test.go
```diff
@@ -171,3 +177,175 @@ func TestBuildEncoderRequest(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.expected, mmItemURL(tt.item))
+		})
+	}
+}
+
+func TestFanoutEncoderPrimerDeduplication(t *testing.T) {
+	// Set up a logger so the Server doesn't panic on s.logger calls.
+	logger := zap.New(zap.UseDevMode(true))
```
elevran: nit: might want to use `zap.NewNop()` to discard instead.

pkg/sidecar/proxy/connector_epd_shared_storage_test.go
```diff
@@ -171,3 +177,175 @@ func TestBuildEncoderRequest(t *testing.T) {
...
+	encoderHostPort := encoderURL.Host
+
+	tests := []struct {
+		name          string
+		request       map[string]any
+		expectedCalls int32
+	}{
+		{
+			name: "no duplicates — all items sent",
+			request: map[string]any{
```
elevran: consider using a test helper to compose the request as many fields are shared between these three samples.

pkg/sidecar/proxy/connector_epd_shared_storage.go
```diff
@@ -101,14 +101,48 @@ func buildEncoderRequest(originalRequest map[string]any, mmItem map[string]any)
...
-	mmItems := extractMMItems(originalRequest)
-	if len(mmItems) == 0 {
+	allItems := extractMMItems(originalRequest)
+	if len(allItems) == 0 {
 		s.logger.V(4).Info("no multimodal items, skipping encoder", "requestID", requestID)
 		return nil
 	}
 
+	// Deduplicate URL-based items; keep all non-URL items (e.g. inline audio).
+	seenURLs := make(map[string]bool)
```
elevran: nit: debating myself if `k8s.Set`, `map[string]struct{}` or bool values as you have it...

pkg/sidecar/proxy/connector_epd_shared_storage_test.go
roytman: done

pkg/sidecar/proxy/connector_epd_shared_storage.go
roytman: done with map[string]struct{}

pkg/sidecar/proxy/connector_epd_shared_storage_test.go
roytman: The Kubernetes controller-runtime Zap Package doesn't have zap.NewNop(),

removed this initialization

elevran: then maybe `WriteTo(io.Writer)` unless you want to see the logs?

roytman: I removed it at all, please take a look now.

PR #720, opened 2026-03-15
Basic implementation of dynamic LoRA adapters placement, based on shuffle sharding algorithm

dmitripikus:
Here is a basic implementation of dynamic LoRA adapters placement, based on shuffle sharding algorithm
**This implementation is for experimentation and feedback.**

At this point:
* Solution is implemented as a scorer that causes LoRA adapters to be loaded (and serve requests) at specific vLLM endpoints
* Re-balancing of LoRA adapters in a case of endpoints restarts, scaling up/down will be added in the following PR
* Currently base model is received by scorer as a parameter. Base model discovery will be added in the following PR

### DIFF
--- a/pkg/plugins/scorer/lora_aware.go
+++ b/pkg/plugins/scorer/lora_aware.go
@@ -1,0 +1,354 @@
+package scorer
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"hash/fnv"
+	"math/rand"
+	"sort"
+	"sync"
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	logutil "sigs.k8s.io/gateway-api-inference-extension/pkg/common/util/logging"
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/plugin"
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/scheduling"
+)
+
+const (
+	// LoRAAwareType is the type of the LoRAAware scorer.
+	LoRAAwareType = "lora-aware-scorer"
+
+	// minShardSize is the minimum number of endpoints per LoRA adapter shard.
+	minShardSize = 2
+)
+
+// LoRAAwareParameters defines the parameters for the LoRAAware scorer.
+type LoRAAwareParameters struct {
+	// ShardSize defines the number of endpoints to assign to each LoRA adapter.
+	// If not specified or set to 0, it will be calculated dynamically based on the
+	// number of available endpoints using the formula: ceil(N / 2), where N is the
+	// number of endpoints. This provides a conservative default that balances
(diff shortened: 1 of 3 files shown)


pkg/plugins/scorer/lora_aware.go
```diff
@@ -0,0 +1,354 @@
...
+// them on every request, improving performance for frequently used adapters.
+type LoRAAware struct {
+	typedName           plugin.TypedName
+	configuredShardSize int                 // 0 means auto-calculate based on endpoint count
+	baseModel           string              // Base model name
+	shardCache          map[string][]string // Cache of adapter -> endpoint names
+	shardCacheMu        sync.RWMutex        // Protects shardCache
+	cachedShardSize     int                 // Cached calculated shard size
+	cachedEndpointCount int                 // Number of endpoints for cached shard size
+	shardSizeMu         sync.RWMutex        // Protects shard size cache
```
hexfusion: nit: these two caches are coupled, hard assignments depend on shard size. a single mutex would make invalidation simpler when you add rebalancing.

pkg/plugins/scorer/lora_aware.go
```diff
@@ -0,0 +1,354 @@
...
+	// Shuffle the endpoints deterministically based on the adapter name
+	shuffled := make([]scheduling.Endpoint, len(sortedEndpoints))
+	copy(shuffled, sortedEndpoints)
+	rng.Shuffle(len(shuffled), func(i, j int) {
+		shuffled[i], shuffled[j] = shuffled[j], shuffled[i]
+	})
+
+	// Take the top shardSize endpoints from the shuffled list
+	result := make([]scheduling.Endpoint, shardSize)
+	copy(result, shuffled[:shardSize])
```
hexfusion: nit: logically we are making 3 allocations and 2 full copies of the endpoint list, how big could this list get?

hexfusion (commented):
i understand this PR is experimental for benchmarking added a few things to consider as you iterate.

pkg/plugins/scorer/lora_aware.go
dmitripikus: Fixed. Thank you!

pkg/plugins/scorer/lora_aware.go
dmitripikus: Fixed. The fix reduces getShardForAdapter from 3 allocations + 2 full copies down to 1 allocation + 1 copy by sorting and shuffling in-place on a single copy, then returning a sub-slice.
The list could reach hundreds of vLLM pods in a production cluster, so redundant allocations are wasteful.
Thank you!

dmitripikus:
working on this

kfswain:
@dmitripikus sg, feel free to tag me when ready for another pass, thanks!

dmitripikus:
Hi @kfswain , yes, sure, I'll do so soon

PR #721, opened 2026-03-16
Rename EncoderPodsHeader according to other constants

roytman:
Rename the `common.EncoderPodsHeader` constant according to other constant names.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,8 @@ lint: check-golangci-lint check-typos ## Run lint (use LINT_NEW_ONLY=true to onl
 		printf "\033[33mChecking all code (LINT_NEW_ONLY=false, default)\033[0m\n"; \
 		$(GOLANGCI_LINT) run; \
 	fi
-	$(TYPOS)
+	@echo "Checking for spelling errors with typos..."
+	@$(TYPOS) --format brief
 
 .PHONY: test
 test: test-unit test-e2e ## Run all tests (unit and e2e)
--- a/Makefile.tools.mk
+++ b/Makefile.tools.mk
@@ -200,5 +200,3 @@ check-typos:
 	  echo "ERROR: typos is not installed."; \
 	  echo "Run: make install-typos (or install-tools)"; \
 	  exit 1; }
-	@echo "Checking for spelling errors with typos..."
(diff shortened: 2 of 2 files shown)


pkg/common/common.go
```diff
@@ -13,8 +13,8 @@ const (
 	// DataParallelPodHeader is the header name used to indicate the worker <ip:port> for Data Parallel
 	DataParallelPodHeader = "x-data-parallel-host-port"
 
-	// EncoderHostsPortsHeader is the header name used to indicate Encoder workers <ip:port> list
-	EncoderHostsPortsHeader = "x-encoder-hosts-ports"
+	// EncoderPodsHeader is the header name used to indicate Encoder workers <ip:port> list
+	EncoderPodsHeader = "x-encoder-hosts-ports"
```
hexfusion, line 17: question: should these consts be infrastructure agnostic? the header carries <ip:port> pairs which is a network level concept. renaming to "Pods" couples this to kubernetes when the data itself is really transport agnostic. absolutely get your intention of making these more aligned just an observation.

roytman: Hi Sam, I believe HostPort was widely used in the past, but it has largely been replaced by Endpoints or, as in this case, Pods. I view `llm-d-interface-scheduler` as an extension of the Kubernetes Gateway API, making it infrastructure-dependent. That said, I’m open to any suggestions you might have.

hexfusion: that is fair in general my concern is we need to consider that llm-d could/should exist without a kube dependency but I see your point on GIE, saying that no need to block this PR.

shmuelk: As the person who did quite of the the pod to endpoint refactoring in many of the names in IGW, perhaps we should rename these constants here as well. Here are my suggestions:

DataParallelPodHeader -> DataParallelEndpointHeader
EncoderHostsPortsHeader -> EncoderEndpointHeader
PrefillPodHeader > PrefillEndpointHeader

If any are changed, they all should be changed, consistency is very important.

roytman: I'll rename the constants according to Samuel's suggestion
/hold

PR #727, opened 2026-03-17
Deprecate the workaround used to support vLLM Data Parallel on Istio 1.28

shmuelk:
Istio 1.28 did not properly support Inference Gateway Extension InferencePool CRs with multiple targetPorts specified.

A work around was added to the llm-d-inference-scheduler and the llm-d-routing-sidecar.

With Istio 1.28.1 GA for some time, this work around is no longer needed. This PR marks the use of the various parts of that workaround as being deprecated and will be removed in a future release of the llm-d-inference-scheduler.

The deprecated items include:
  - The data-parallel-profile-handler. Replaced by the standard simple-profile-handler
  - The use of the targetPort parameter of the pd-profile-handler
  - The use of x-data-parallel-host-port header with the llm-d-routing-sidecar
 
 In addition support for these items has been removed from the env-dev-kind make target and its shell script.

### DIFF
--- a/pkg/plugins/profile/dp_profile_handler.go
+++ b/pkg/plugins/profile/dp_profile_handler.go
@@ -29,7 +29,7 @@ var _ scheduling.ProfileHandler = &DataParallelProfileHandler{}
 
 // DataParallelProfileHandlerFactory defines the factory function for the DataParallelProfileHandler
 func DataParallelProfileHandlerFactory(name string, rawParameters json.RawMessage, handle plugin.Handle) (plugin.Plugin, error) {
-	log.FromContext(handle.Context()).Info("Deprecated: Use simple-profile-handler with Istio 1.28.1")
+	log.FromContext(handle.Context()).Info("Deprecated: Use simple-profile-handler with Istio >= 1.28.1")
 	parameters := dataParallelProfileHandlerParameters{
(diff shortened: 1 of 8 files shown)


pkg/plugins/profile/dp_profile_handler.go
```diff
@@ -28,7 +28,8 @@ type dataParallelProfileHandlerParameters struct {
 var _ scheduling.ProfileHandler = &DataParallelProfileHandler{}
 
 // DataParallelProfileHandlerFactory defines the factory function for the DataParallelProfileHandler
-func DataParallelProfileHandlerFactory(name string, rawParameters json.RawMessage, _ plugin.Handle) (plugin.Plugin, error) {
+func DataParallelProfileHandlerFactory(name string, rawParameters json.RawMessage, handle plugin.Handle) (plugin.Plugin, error) {
+	log.FromContext(handle.Context()).Info("Deprecated: Use simple-profile-handler with Istio 1.28.1")
```
hexfusion: nit: Istio 1.28.1 should this be >= 1.28.1 or 1.28.x? current wording reads like it only applies to that specific patch version.

shmuelk: Fixed

Issue #729, opened 2026-03-17
Fix test-e2e for podman environments


hexfusion:
Related: #605

**What happened**:

`make test-e2e` fails on systems with `podman-docker`. The Makefile detects `docker` first, but the podman wrapper doesn't support `--platform` on save.

```
 STEP: Loading ghcr.io/llm-d/llm-d-inference-sim:latest into the cluster e2e-tests using docker @ 03/18/26 07:19:51.93
  Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
  Error: unknown flag: --platform
  See 'podman save --help'
  [FAILED] in [BeforeSuite] - /home/sbatsche/projects/llm-d/llm-d-inference-scheduler/test/e2e/e2e_suite_test.go:197 @ 03/18/26 07:19:51.952
[BeforeSuite] [FAILED] [16.685 seconds]
[BeforeSuite] 
/home/sbatsche/projects/llm-d/llm-d-inference-scheduler/test/e2e/e2e_suite_test.go:97

  [FAILED] No future change is possible.  Bailing out early after 0.020s.
  Expected
      <int>: 125
  to match exit code:
      <int>: 0
  In [BeforeSuite] at: /home/sbatsche/projects/llm-d/llm-d-inference-scheduler/test/e2e/e2e_suite_test.go:197 @ 03/18/26 07:19:51.952
````

Additionally, `kustomize` is required by the test suite but missing from the `test-e2e` dependency chain.

**What you expected to happen**:

e2e to run while using podman or docker  and all dependencies installed by default.

**Environment:** Fedora 43, podman 5.8.1

Issue #605, opened 2026-02-09
Improve developer onboarding and ongoing contribution experience


elevran:
This is a request for feedback on out new project contributor flows. Looking for feedback and suggested improvements in
- understanding the repo structure
- knowing where to ask for help
- processes for setting up dev env
- submitting issues and PR's
- etc.

Would be happy to get feedback (what's missing, what needs changes, what's working well?), suggestions for improvements as well as updates to guides and onboarding material.
If you're a first time contributor to llm-d-inference-scheduler, please your feedback (and feel free to create PRs to address pain points and streamline onboarding for those that will follow :-) )

elevran:
cc @Griffin-Sullivan

revit13:
Based on my initial local deployment, It seems that `DEVELOPMENT.md` page should be updated: the context name mentioned seems incorrect -its name in the deployment is `kind-llm-d-inference-scheduler-dev` instead of the mentioned `llm-d-inference-scheduler-dev`, in addition the service name should be changed as well (inference-gateway -> inference-gateway-istio).

roytman:
As a newcomer to the project, I find the list of slash commands (such as /assign) along with their options and explanations, to be very useful.

PR #730, opened 2026-03-17
fix: support podman-docker in e2e tests and Makefile

hexfusion:
Fix make test-e2e on podman systems where podman-docker is installed by preferring podman in container runtime detection.

Fixes: #729
Related: #605

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ else
 	TYPOS_ARCH = $(TYPOS_TARGET_ARCH)-unknown-linux-musl
 endif
 
-CONTAINER_RUNTIME := $(shell { command -v docker >/dev/null 2>&1 && echo docker; } || { command -v podman >/dev/null 2>&1 && echo podman; } || echo "")
+CONTAINER_RUNTIME := $(shell { command -v podman >/dev/null 2>&1 && echo podman; } || { command -v docker >/dev/null 2>&1 && echo docker; } || echo "")
 export CONTAINER_RUNTIME
 BUILDER := $(shell command -v buildah >/dev/null 2>&1 && echo buildah || echo $(CONTAINER_RUNTIME))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -63,7 +63,7 @@ else
 	TYPOS_ARCH = $(TYPOS_TARGET_ARCH)-unknown-linux-musl
 endif
 
-CONTAINER_RUNTIME := $(shell { command -v docker >/dev/null 2>&1 && echo docker; } || { command -v podman >/dev/null 2>&1 && echo podman; } || echo "")
+CONTAINER_RUNTIME := $(shell { command -v podman >/dev/null 2>&1 && echo podman; } || { command -v docker >/dev/null 2>&1 && echo docker; } || echo "")
```
elevran, line 66: Just making sure I understand this: if only one is installed, it would be selected.
If both are installed, it defaulted to `docker` but now defaults to `podman`?
When would one want one over the other? Seem more of a personal preference, even when both are installed.

shmuelk: See the issue.

> make test-e2e fails on systems with podman-docker. The Makefile detects docker first, but the podman wrapper doesn't support --platform on save.

Apparently podman-docker provides a docker wrapper that invokes podman under the covers for systems that don't explicitly support podman. However, it is apparently incomplete.

hexfusion: podman-docker is a fairly common package. As a makefile is developer focused it feels reasonable that it just works without friction.

hexfusion: I can come up with a change that does not change base assumptions if we feel that this change itself causes friction. I will also look to resolve the issue upstream in the podman repo.

elevran: I'm fine with it as is.
Not sure how common installing both directly is and it seems that podman-docker is common for podman users.

PR #732, opened 2026-03-17
Unified Disaggregate Handler

roytman:
We use pd_disagg_handler to cover PD and P/D disaggregation cases. The introduction of a disaggregated Encoder significantly increases complexity, introducing EPD, EP/D (the current P/D case), E/PD, and E/P/D scenarios. To avoid defining a separate handler for every permutation, this PR introduces a more flexible approach.

This PR implements a generic handler that supports optional E and P profiles. Users can now use the same unified handler while defining or skipping specific profiles as needed.

Additionally, each profile now includes a decider plugin. This plugin determines whether a profile should be executed (returning true) or skipped (returning false) based on the specific request and/or system context.

### DIFF
--- a/pkg/plugins/profile/disagg_profile_handler.go
+++ b/pkg/plugins/profile/disagg_profile_handler.go
@@ -57,34 +57,56 @@ type disaggProfileHandlerParameters struct {
 	DeprecatedDeciderPluginName string `json:"deciderPluginName,omitempty"`
 }
 
-// applyDeprecatedFields promotes flat legacy fields into the nested Profiles/Deciders
-// fields when the nested fields are absent, and logs a deprecation warning for each
-// legacy field in use.
-func (p *disaggProfileHandlerParameters) applyDeprecatedFields(logger interface{ Info(string, ...any) }) {
-	if p.DeprecatedDecodeProfile != "" && p.Profiles.Decode == "" {
+// hasDeprecatedFields reports whether any deprecated flat field is set.
+func (p *disaggProfileHandlerParameters) hasDeprecatedFields() bool {
+	return p.DeprecatedDecodeProfile != "" ||
+		p.DeprecatedPrefillProfile != "" ||
+		p.DeprecatedEncodeProfile != "" ||
+		p.DeprecatedPrefillDeciderPluginName != "" ||
+		p.DeprecatedDeciderPluginName != "" ||
+		p.DeprecatedEncodeDeciderPluginName != ""
+}
+
+// hasNestedFields reports whether any nested profiles/deciders field is set.
+func (p *disaggProfileHandlerParameters) hasNestedFields() bool {
+	return p.Profiles.Decode != "" || p.Profiles.Prefill != "" || p.Profiles.Encode != "" ||
+		p.Deciders.Prefill != "" || p.Deciders.Encode != ""
+}
+
+// validateAndApplyDeprecatedFields returns an error when both deprecated flat fields
+// and nested profiles/deciders fields are set simultaneously, otherwise promotes the
+// deprecated flat fields into their nested equivalents and logs a warning for each.
+func (p *disaggProfileHandlerParameters) validateAndApplyDeprecatedFields(logger interface{ Info(string, ...any) }) error {
+	if p.hasDeprecatedFields() && p.hasNestedFields() {
+		return errors.New("cannot mix deprecated flat parameters (decodeProfile, prefillProfile, encodeProfile, " +
+			"deciderPluginName, prefillDeciderPluginName, encodeDeciderPluginName) " +
+			"with nested parameters (profiles, deciders): use one format or the other")
+	}
+	if p.DeprecatedDecodeProfile != "" {
 		logger.Info("Deprecated parameter 'decodeProfile', use 'profiles.decode' instead")
 		p.Profiles.Decode = p.DeprecatedDecodeProfile
 	}
-	if p.DeprecatedPrefillProfile != "" && p.Profiles.Prefill == "" {
+	if p.DeprecatedPrefillProfile != "" {
 		logger.Info("Deprecated parameter 'prefillProfile', use 'profiles.prefill' instead")
 		p.Profiles.Prefill = p.DeprecatedPrefillProfile
 	}
-	if p.DeprecatedEncodeProfile != "" && p.Profiles.Encode == "" {
+	if p.DeprecatedEncodeProfile != "" {
 		logger.Info("Deprecated parameter 'encodeProfile', use 'profiles.encode' instead")
 		p.Profiles.Encode = p.DeprecatedEncodeProfile
 	}
-	if p.DeprecatedPrefillDeciderPluginName != "" && p.Deciders.Prefill == "" {
+	if p.DeprecatedPrefillDeciderPluginName != "" {
 		logger.Info("Deprecated parameter 'prefillDeciderPluginName', use 'deciders.prefill' instead")
 		p.Deciders.Prefill = p.DeprecatedPrefillDeciderPluginName
 	}
-	if p.DeprecatedDeciderPluginName != "" && p.Deciders.Prefill == "" {
+	if p.DeprecatedDeciderPluginName != "" {
 		logger.Info("Deprecated parameter 'deciderPluginName', use 'deciders.prefill' instead")
 		p.Deciders.Prefill = p.DeprecatedDeciderPluginName
 	}
-	if p.DeprecatedEncodeDeciderPluginName != "" && p.Deciders.Encode == "" {
+	if p.DeprecatedEncodeDeciderPluginName != "" {
 		logger.Info("Deprecated parameter 'encodeDeciderPluginName', use 'deciders.encode' instead")
 		p.Deciders.Encode = p.DeprecatedEncodeDeciderPluginName
 	}
+	return nil
 }
 
 // DisaggProfileHandlerFactory is the unified factory for all disaggregation profile handlers.
@@ -104,7 +126,9 @@ func DisaggProfileHandlerFactory(name string, rawParameters json.RawMessage, han
 	logger := log.FromContext(handle.Context())
 
 	// Promote any deprecated flat fields that are still in use.
-	parameters.applyDeprecatedFields(logger)
+	if err := parameters.validateAndApplyDeprecatedFields(logger); err != nil {
+		return nil, err
+	}
 
 	// Apply profile name defaults for any fields still unset.
 	if parameters.Profiles.Decode == "" {
--- a/pkg/plugins/profile/disagg_profile_handler_test.go
+++ b/pkg/plugins/profile/disagg_profile_handler_test.go
@@ -256,11 +256,14 @@ func TestDisaggProfileHandlerFactory_DeprecatedFlatParams(t *testing.T) {
 			"encodeProfile":            "my-encode",
 			"prefillDeciderPluginName": PrefixBasedPDDeciderPluginType,
 		}, false},
-		{"deprecated fields overridden by nested", map[string]any{
-			"decodeProfile": "ignored-decode",
+		{"mixing deprecated and nested fields is an error", map[string]any{
+			"decodeProfile": "my-decode",
 			"profiles":      map[string]any{"decode": "my-decode"},
-			"deciders":      map[string]any{"prefill": AlwaysDisaggPDDeciderPluginType},
-		}, false},
+		}, true},
+		{"mixing deprecated decider and nested deciders is an error", map[string]any{
+			"prefillDeciderPluginName": PrefixBasedPDDeciderPluginType,
+			"deciders":                 map[string]any{"prefill": AlwaysDisaggPDDeciderPluginType},
+		}, true},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {


pkg/plugins/filter/pd_role.go
```diff
@@ -20,6 +20,8 @@ const (
 	DecodeRoleType = "decode-filter"
 	// PrefillRoleType is the type of the PrefillFilter
 	PrefillRoleType = "prefill-filter"
+	// RolePrefillDecode set for workers that can handle prefill and decode stages
```
mayabar: "prefill-decode" is the same as "both", do we need both?
maybe add depreciation comment for "both" role?

pkg/plugins/profile/always_disagg_decider.go
```diff
@@ -43,6 +45,6 @@ func (d *AlwaysDisaggPDDecider) WithName(name string) *AlwaysDisaggPDDecider {
 	return d
 }
 
-func (d *AlwaysDisaggPDDecider) disaggregate(ctx context.Context, inputTokens int, endpoint scheduling.Endpoint) bool {
+func (d *AlwaysDisaggPDDecider) decide(ctx context.Context, request *scheduling.LLMRequest, endpoint scheduling.Endpoint) bool {
```
mayabar: This function answers the question "should the processing be disaggregated"
If the class name is deciderPlugin and function name decide - there is not way to understand which decision is done.

pkg/plugins/profile/decider-plugin.go
```diff
@@ -0,0 +1,15 @@
+// Package profile provides profile handler plugins for the epp.
```
mayabar, line 1: please rename to decider-plugin.go

pkg/plugins/profile/always_encode_decider.go
```diff
@@ -0,0 +1,65 @@
+package profile
```
mayabar: consider to make it consistent with pd decider: has "disagg" in the name in both or without "disagg" in both

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,321 @@
...
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/metrics"
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/telemetry"
+)
+
+// ── Constants ───────────────────────────────────────────────────────────────
+
+const (
+	// DisaggProfileHandlerType is the canonical type for the unified disaggregation profile handler.
+	DisaggProfileHandlerType = "disagg-profile-handler"
+	defaultDeciderPluginName = AlwaysDisaggDeciderPluginType
```
mayabar: not in use?

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,321 @@
...
+
+	defaultDecodeProfile  = "decode"
+	defaultPrefillProfile = "prefill"
+	defaultEncodeProfile  = "encode"
+
+	// AverageCharactersPerToken is an estimated average characters per token,
+	// used since the request we cache is not tokenized.
+	AverageCharactersPerToken = 4
```
mayabar: used in prefix prefilDecider, consider to move it there

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,321 @@
...
+			info.Port = h.primaryPort
+			updated.TargetEndpoints = append(updated.TargetEndpoints,
+				scheduling.NewEndpoint(info, target.GetMetrics().Clone(), nil))
+		}
+		updatedResults[h.decodeProfile] = &updated
+	} else {
+		updatedResults[h.decodeProfile] = decodeRunResults
+	}
+
+	if h.pdDecider != nil {
```
mayabar: is this if required?

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,321 @@
...
+		if prefillRes, ok := profileResults[h.prefillProfile]; ok && prefillRes != nil {
+			updatedResults[h.prefillProfile] = prefillRes
+		}
+	}
+
+	if h.encodeDecider != nil {
```
mayabar: same as above

pkg/plugins/profile/always_encode_decider.go
revit13: Changed to always_disagg_encode_decider.go and  Thanks

pkg/plugins/profile/decider-plugin.go
revit13: Done. Thanks

pkg/plugins/profile/always_disagg_decider.go
revit13: I changed decide to disaggregate. Thanks

pkg/plugins/profile/disagg_profile_handler.go
revit13: Removed. Thanks

pkg/plugins/profile/disagg_profile_handler.go
revit13: moved to prefix_based_pd_decider.go Thanks

pkg/plugins/profile/disagg_profile_handler.go
revit13: I think it's safer, because pdDecider is an optional plugin.

pkg/plugins/filter/pd_role.go
revit13: It's kept in for backward compatibility. I've added a comment to make this clearer. Thanks!

pkg/plugins/profile/disagg_profile_handler.go
mayabar: in case pdDecider is not defined, profileResults[h.prefillProfile] will be nil and we will not enter the next "if"

docs/disagg_pd.md
```diff
@@ -181,8 +183,9 @@ plugins:
   - type: prefix-based-pd-decider
     parameters:
       nonCachedTokens: 8
+  # Note: pd-profile-handler is deprecated, use disagg-profile-handler instead
```
shmuelk: If it's deprecated, please update the doc with the correct profile(s)

pkg/plugins/filter/epd_role.go
```diff
@@ -0,0 +1,40 @@
...
+	// RoleEncode set for designated encode workers (first stage in pipeline)
+	RoleEncode = "encode"
+
+	// RoleEncodePrefill set for workers that can handle encode+prefill (EP/D disaggregation)
+	RoleEncodePrefill = "encode-prefill"
+
+	// RoleEncodeDecode set for workers that can handle encode+decode (E/PD disaggregation - rare)
+	RoleEncodeDecode = "encode-decode"
+	// RoleAll set for workers that can handle all stages (encode, prefill, and decode)
+	RoleAll = "all"
```
shmuelk: While convenient a role of all, considering that the purpose of this PR is generalize the profile handler for disaggregation, seems a bit over powerful. Tomorrow there will X, Y, and Z disaggregation....

pkg/plugins/pre-request/encode_prerequest.go
```diff
@@ -0,0 +1,125 @@
...
+		span.SetAttributes(attribute.String("llm_d.epp.encode.reason", "scheduling_result_is_nil"))
+		return
+	}
+
+	if request.TargetModel != "" {
+		span.SetAttributes(attribute.String("gen_ai.request.model", request.TargetModel))
+	}
+	if request.RequestId != "" {
+		span.SetAttributes(attribute.String("gen_ai.request.id", request.RequestId))
+	}
```
shmuelk: This check isn't needed, there will always be a RequestId set.

pkg/plugins/pre-request/encode_prerequest.go
```diff
@@ -0,0 +1,125 @@
...
+			attribute.String("llm_d.epp.encode.reason", "no_encode_profile_result"),
+		)
+		return // encode profile failed to run or we chose not to run it, no-op in this case
+	}
+
+	// Collect all target endpoints as comma-separated host:port pairs
+	var encodeHostPorts []string
+	for _, endpoint := range encodeProfileRunResult.TargetEndpoints {
+		targetPod := endpoint.GetMetadata()
+		encodeHostPort := net.JoinHostPort(targetPod.Address, targetPod.Port)
```
shmuelk: We should avoid using the word Pod unless it really refers to an entire Pod. Here it doesn;t.

```suggestion
		targetEndpoint := endpoint.GetMetadata()
		encodeHostPort := net.JoinHostPort(targetEndpoint.Address, targetEndpoint.Port)
```

pkg/plugins/pre-request/encode_prerequest_test.go
```diff
@@ -0,0 +1,292 @@
...
+
+	val := request.Headers[common.EncoderEndpointsHeader]
+	assert.Equal(t, "", val)
+}
+
+func TestPreRequestEncodeCustomProfile(t *testing.T) {
+	ctx := utils.NewTestContext(t)
+	handler := NewEncodeHeaderHandler("my-custom-encode").WithName("test")
+
+	request := &scheduling.LLMRequest{
```
shmuelk, line 224: Add the RequestId field

pkg/plugins/pre-request/encode_prerequest_test.go
```diff
@@ -0,0 +1,292 @@
...
+	handler.PreRequest(ctx, request, result)
+
+	assert.Equal(t, net.JoinHostPort(encodeTestAddr, encodeTestPort), request.Headers[common.EncoderEndpointsHeader])
+}
+
+func TestPreRequestEncodeClearsHeaderWhenNoEncodeResult(t *testing.T) {
+	ctx := utils.NewTestContext(t)
+	handler := NewEncodeHeaderHandler("encode").WithName("test")
+
+	request := &scheduling.LLMRequest{
```
shmuelk, line 202: Add the RequestId field

pkg/plugins/pre-request/encode_prerequest_test.go
```diff
@@ -0,0 +1,292 @@
...
+
+	_, exists := request.Headers[common.EncoderEndpointsHeader]
+	assert.False(t, exists)
+}
+
+func TestPreRequestEncodeClearsExistingHeader(t *testing.T) {
+	ctx := utils.NewTestContext(t)
+	handler := NewEncodeHeaderHandler("encode").WithName("test")
+
+	request := &scheduling.LLMRequest{
```
shmuelk, line 175: Add the requestId field

pkg/plugins/pre-request/encode_prerequest_test.go
```diff
@@ -0,0 +1,292 @@
...
+	handler.PreRequest(ctx, request, result)
+
+	assert.Equal(t, net.JoinHostPort(encodeTestAddr, encodeTestPort), request.Headers[common.EncoderEndpointsHeader])
+}
+
+func TestPreRequestEncodeProfileNotExists(t *testing.T) {
+	ctx := utils.NewTestContext(t)
+	handler := NewEncodeHeaderHandler("encode").WithName("test")
+
+	request := &scheduling.LLMRequest{
```
shmuelk, line 155: Add the RequestId field

pkg/plugins/pre-request/encode_prerequest_test.go
```diff
@@ -0,0 +1,292 @@
...
+	handler.PreRequest(ctx, request, result)
+
+	assert.Equal(t, net.JoinHostPort(encodeTestAddr, encodeTestPort), request.Headers[common.EncoderEndpointsHeader])
+}
+
+func TestPreRequestEncodeIPv6Address(t *testing.T) {
+	ctx := utils.NewTestContext(t)
+	handler := NewEncodeHeaderHandler("encode").WithName("test")
+
+	request := &scheduling.LLMRequest{
```
shmuelk, line 249: Add the RequestId field

pkg/plugins/pre-request/encode_prerequest_test.go
```diff
@@ -0,0 +1,292 @@
...
+	handler.PreRequest(ctx, request, result)
+
+	assert.Equal(t, net.JoinHostPort(encodeTestIPv6Addr, encodeTestPort), request.Headers[common.EncoderEndpointsHeader])
+}
+
+func TestPreRequestEncodeMultipleEndpoints(t *testing.T) {
+	ctx := utils.NewTestContext(t)
+	handler := NewEncodeHeaderHandler("encode").WithName("test")
+
+	request := &scheduling.LLMRequest{
```
shmuelk, line 299: Add the RequestId field

pkg/plugins/profile/always_disagg_decider.go
```diff
@@ -29,7 +29,9 @@ func AlwaysDisaggPDDeciderPluginFactory(name string, _ json.RawMessage,
 }
 
 func newAlwaysDisaggPDDecider() *AlwaysDisaggPDDecider {
-	return &AlwaysDisaggPDDecider{}
+	return &AlwaysDisaggPDDecider{
+		typedName: plugin.TypedName{Type: AlwaysDisaggDeciderPluginType},
```
shmuelk: If this is a PD Decider, should the type have PD in there some where?

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,307 @@
...
+	defaultEncodeProfile  = "encode"
+)
+
+// ── Factory & constructor ────────────────────────────────────────────────────
+
+type disaggProfileHandlerParameters struct {
+	DecodeProfile            string `json:"decodeProfile"`
+	PrefillProfile           string `json:"prefillProfile"`
+	EncodeProfile            string `json:"encodeProfile"`
+	PrimaryPort              int    `json:"primaryPort"`
```
shmuelk: I would leave out all of the PrimaryPort support. That was a workaround for Istio 1.28 and below, which has been deprecated. This is new code, there is no need for backwards compatibility here.

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,307 @@
...
+		pdDecider, encodeDecider,
+	)
+	return handler.WithName(name), nil
+}
+
+// NewDisaggProfileHandler creates a DisaggProfileHandler directly.
+// Active stages are determined by which profile names are non-empty.
+func NewDisaggProfileHandler(
+	decodeProfile, prefillProfile, encodeProfile string,
+	primaryPort int,
```
shmuelk: See above. Remove

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,307 @@
...
+	pdDecider deciderPlugin,
+	encodeDecider deciderPlugin,
+) *DisaggProfileHandler {
+	return newDisaggProfileHandler(
+		DisaggProfileHandlerType,
+		decodeProfile, prefillProfile, encodeProfile,
+		primaryPort,
```
shmuelk: See above. Remove

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,307 @@
...
+// All three handler types (P/D, E/PD, E/P/D) share this single implementation;
+// active stages are selected by setting encodeProfile / prefillProfile.
+type DisaggProfileHandler struct {
+	typedName      plugin.TypedName
+	decodeProfile  string
+	prefillProfile string
+	encodeProfile  string
+	pdDecider      deciderPlugin
+	encodeDecider  deciderPlugin
+	primaryPort    string
```
shmuelk: See above. Remove

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,307 @@
...
+// WithName sets the instance name of the plugin.
+func (h *DisaggProfileHandler) WithName(name string) *DisaggProfileHandler {
+	h.typedName.Name = name
+	return h
+}
+
+func newDisaggProfileHandler(
+	handlerType string,
+	decodeProfile, prefillProfile, encodeProfile string,
+	primaryPort int,
```
shmuelk: See above. Remove

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,307 @@
...
+		typedName:      plugin.TypedName{Type: handlerType},
+		decodeProfile:  decodeProfile,
+		prefillProfile: prefillProfile,
+		encodeProfile:  encodeProfile,
+		pdDecider:      pdDecider,
+		encodeDecider:  encodeDecider,
+	}
+	if primaryPort != 0 {
+		h.primaryPort = strconv.Itoa(primaryPort)
+	}
```
shmuelk: See above. Remove

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,307 @@
...
+		span.SetAttributes(attribute.String("llm_d.profile_handler.decision", "complete_nil_request"))
+		return map[string]scheduling.SchedulerProfile{}
+	}
+
+	if request.TargetModel != "" {
+		span.SetAttributes(attribute.String("gen_ai.request.model", request.TargetModel))
+	}
+	if request.RequestId != "" {
+		span.SetAttributes(attribute.String("gen_ai.request.id", request.RequestId))
+	}
```
shmuelk: RequestId is always set, the check isn't needed

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,307 @@
...
+
+		updated := scheduling.ProfileRunResult{}
+		for _, target := range decodeRunResults.TargetEndpoints {
+			info := target.GetMetadata().Clone()
+			info.Port = h.primaryPort
+			updated.TargetEndpoints = append(updated.TargetEndpoints,
+				scheduling.NewEndpoint(info, target.GetMetrics().Clone(), nil))
+		}
+		updatedResults[h.decodeProfile] = &updated
+	} else {
```
shmuelk: This is new code, theer is no need to support the Istio 1.28 and lower, workaround that has been deprecated.

pkg/plugins/profile/multimodal_helpers.go
```diff
@@ -0,0 +1,23 @@
...
+		for _, block := range msg.Content.Structured {
+			if block.Type == "image_url" || block.Type == "video_url" || block.Type == "input_audio" {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// Made with Bob
```
shmuelk: Please remove

pkg/plugins/profile/pd_profile_handler.go
```diff
@@ -125,13 +117,15 @@ func NewPdProfileHandler(prefillProfile, decodeProfile, prefixPluginType, prefix
 }
 
 // PdProfileHandler handles scheduler profiles for PD.
+//
+// Deprecated: Use DisaggProfileHandler instead.
```
shmuelk, line 121: The factory should log a deprecation message. Most users will not read the code....

pkg/plugins/profile/disagg_profile_handler.go
roytman: removed the check

pkg/plugins/profile/disagg_profile_handler.go
roytman: Some as above, removed the check

pkg/plugins/profile/disagg_profile_handler.go
roytman: OK, done

pkg/plugins/profile/pd_profile_handler.go
roytman: it is printed at L:50

pkg/plugins/pre-request/encode_prerequest.go
roytman: It should be set, but since it is a user request, we cannot assume it is always correct. 

`pd_prerequest.go` has the same check.

pkg/plugins/filter/epd_role.go
roytman: renamed to `RoleEncodePrefillDecode`

pkg/plugins/pre-request/encode_prerequest_test.go
roytman: done

pkg/plugins/pre-request/encode_prerequest_test.go
roytman: done

pkg/plugins/pre-request/encode_prerequest_test.go
roytman: done

pkg/plugins/pre-request/encode_prerequest_test.go
roytman: done

pkg/plugins/pre-request/encode_prerequest_test.go
roytman: done

pkg/plugins/pre-request/encode_prerequest_test.go
roytman: done

pkg/plugins/profile/always_disagg_decider.go
roytman: Good point

pkg/plugins/profile/disagg_profile_handler.go
roytman: done

pkg/plugins/profile/multimodal_helpers.go
roytman: done

pkg/plugins/profile/disagg_profile_handler.go
roytman: done

pkg/plugins/profile/disagg_profile_handler.go
roytman: done

pkg/plugins/profile/disagg_profile_handler.go
roytman: done

pkg/plugins/profile/disagg_profile_handler.go
roytman: done

pkg/plugins/profile/disagg_profile_handler.go
roytman: done

pkg/plugins/profile/disagg_profile_handler.go
roytman: It should be set, but since it is a user request, we cannot assume it is always correct.

dp_profile_handler.go has the same check.

shmuelk: Actually, if the user doesn't provide the correct header with a requestId one is generated by the IGW.

See: https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/87207aa8c3ca93dfc9a8d02de8858558f93d665e/pkg/epp/handlers/server.go#L212

shmuelk: dp_profile_handler is now deprecated and will be going away eventually...

docs/disagg_pd.md
roytman: done

pkg/plugins/profile/disagg_profile_handler.go
roytman: thank you, I remove the check.

pkg/plugins/pre-request/encode_prerequest.go
roytman: done

pkg/plugins/pre-request/encode_prerequest_test.go
roytman: Why, if the test doesn't use this field, is it so important to add it? 
BTW, [pd_prerequest_test.go](https://github.com/llm-d/llm-d-inference-scheduler/blob/4cd7046e2cf9121be6cdc2fc0815dbeeba721c9f/pkg/plugins/pre-request/pd_prerequest_test.go#L125) doesn't include it either.

shmuelk: I was concerned tat there was code that referenced it. I guess an empty string is a valid value here for the purposes of the tests

pkg/plugins/pre-request/encode_prerequest.go
```diff
@@ -0,0 +1,127 @@
...
+	if request.TargetModel != "" {
+		span.SetAttributes(attribute.String("gen_ai.request.model", request.TargetModel))
+	}
+	span.SetAttributes(attribute.String("gen_ai.request.id", request.RequestId))
+
+	if _, found := request.Headers[common.EncoderEndpointsHeader]; found {
+		request.Headers[common.EncoderEndpointsHeader] = "" // clear header, if already set
+	}
+
+	encodeProfileRunResult, exists := schedulingResult.ProfileResults[p.encodeProfile]
```
elevran: `Pick()` in `disagg_profile_handler.go` (L207) sets `profileResults[h.encodeProfile] = nil` when encode is skipped (i.e., decider returned false). This means the key exists in the map with a nil value.
L112 can then deref a nil pointer and panic. 
Suggest changing check on L102 to `if !exists || encodeProfileRunResult == nil { ... }` so we return in either case before L112.

pkg/plugins/filter/epd_role.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// EncodeRoleFactory defines the factory function for the Encode filter.
+func EncodeRoleFactory(name string, _ json.RawMessage, _ plugin.Handle) (plugin.Plugin, error) {
+	return NewEncodeRole().WithName(name), nil
+}
+
+// NewEncodeRole creates and returns an instance of the Filter configured for encode role.
+// Encode is the first stage in the pipeline: Encode → Prefill → Decode
+// Accepts pods with roles: encode, encode-prefill, encode-decode, or encode-prefill-decode.
```
elevran: Q: is `encode-decode` actually used? I might have missed it or maybe it is an old comment?
If so, is it accepted by both encode role filter and the decode role filter?
Otherwise, a pod labeled `encode-decode` would pass the encode filter but be dropped by the decode filter result in failed routing and no decode workers.

pkg/plugins/profile/always_disagg_encode_decider.go
```diff
@@ -0,0 +1,49 @@
...
+	return d.typedName
+}
+
+// WithName sets the name of the plugin.
+func (d *AlwaysDisaggEncodeDecider) WithName(name string) *AlwaysDisaggEncodeDecider {
+	d.typedName.Name = name
+	return d
+}
+
+func (d *AlwaysDisaggEncodeDecider) disaggregate(_ context.Context, request *scheduling.LLMRequest, _ scheduling.Endpoint) bool {
```
elevran: nit: `AlwaysDisaggEncodeDecider` is misnamed since `hasMultimodalContent()` can return false. Maybe `AlwaysDisaggMultimodalRequests`?

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,269 @@
...
+		if p == nil {
+			return nil, fmt.Errorf("prefillDeciderPluginName not found: %s", parameters.PrefillDeciderPluginName)
+		}
+		var ok bool
+		pdDecider, ok = p.(deciderPlugin)
+		if !ok {
+			return nil, fmt.Errorf("plugin %s does not implement prefillDeciderPlugin", parameters.PrefillDeciderPluginName)
+		}
+	}
+	// Resolve encode decider (optional).
```
elevran, line 159: nit: log clear message/error also when the reverse is true to help diagnose configuration issues?
That is, when someone configures an EncodeProfile but omits EncodeDeciderPluginName.

pkg/plugins/profile/prefix_based_pd_decider.go
```diff
@@ -136,3 +146,21 @@ func (d *PrefixBasedPDDecider) disaggregate(ctx context.Context, inputTokens int
...
 	return true
 }
+
+// getUserInputLenInTokens returns an estimated token count for the user input.
+func getUserInputLenInTokens(request *scheduling.LLMRequest) (int, error) {
+	if request == nil || request.Body == nil {
+		return 0, errors.New("request or request body is nil")
+	}
+	if request.Body.Completions != nil {
+		return len([]byte(request.Body.Completions.Prompt)) / AverageCharactersPerToken, nil
```
elevran:
```suggestion
		return len(request.Body.Completions.Prompt) / AverageCharactersPerToken, nil
```
Calling `len()` on a string already returns the byte count. The `[]byte(...)` conversion allocates a copy which is not needed.

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,269 @@
...
+	// ── Stage 3: Prefill (optional) ────────────────────────────────────────
+	if _, hasPrefillProfile := profiles[h.prefillProfile]; hasPrefillProfile {
+		if _, executed := profileResults[h.prefillProfile]; !executed {
+			if h.pdDecider != nil && h.pdDecider.disaggregate(ctx, request, decodeRes.TargetEndpoints[0]) {
+				span.SetAttributes(attribute.String("llm_d.profile_handler.decision", "run_prefill"))
+				return map[string]scheduling.SchedulerProfile{h.prefillProfile: profiles[h.prefillProfile]}
+			}
+			// Decider rejected prefill — mark as evaluated so we don't re-run the decider.
+			profileResults[h.prefillProfile] = nil
+			span.SetAttributes(attribute.String("llm_d.profile_handler.decision", "skip_prefil"))
```
elevran:
```suggestion
			span.SetAttributes(attribute.String("llm_d.profile_handler.decision", "skip_prefill"))
```
typo

pkg/plugins/pre-request/encode_prerequest.go
```diff
@@ -0,0 +1,127 @@
...
+		span.SetAttributes(attribute.String("llm_d.epp.encode.reason", "scheduling_result_is_nil"))
+		return
+	}
+
+	if request.TargetModel != "" {
+		span.SetAttributes(attribute.String("gen_ai.request.model", request.TargetModel))
+	}
+	span.SetAttributes(attribute.String("gen_ai.request.id", request.RequestId))
+
+	if _, found := request.Headers[common.EncoderEndpointsHeader]; found {
```
elevran: nit: if it wasn't already set, is there harm in setting it to empty? Or just calling `delete()` on it? Either option would be safe and avoid the lookup.

docs/architecture.md
```diff
@@ -319,7 +321,16 @@ Filters out pods that are not marked either as decode or both prefill and decode
 
 #### PrefillFilter
 
-Filters out pods that are not marked as prefill. The filter looks for the label `llm-d.ai/role`, with a value of `prefill`.
+Filters out pods that are not marked as prefill. The filter looks for the label `llm-d.ai/role`, with a value of `prefill`, `encode-prefil`, `prefill-decode` or `encode-prefill-decode`.
```
```suggestion
Filters out pods that are not marked as prefill. The filter looks for the label `llm-d.ai/role`, with a value of `prefill`, `encode-prefill`, `prefill-decode` or `encode-prefill-decode`.
```
typo. 

Can you please review the different role labels in code and docs to make sure they're aligned betwen doc/code and also consistently used by the different role filters?

pkg/plugins/profile/prefix_based_pd_decider.go
```diff
@@ -136,3 +146,21 @@ func (d *PrefixBasedPDDecider) disaggregate(ctx context.Context, inputTokens int
...
+	}
+	if request.Body.ChatCompletions == nil {
+		return 0, errors.New("request has neither completions nor chat completions body")
+	}
+	prompt, err := json.Marshal(request.Body.ChatCompletions.Messages)
```
elevran, line 161: not part of this PR, carry over code:
but I wonder if we need the actual `json.Marshal` in the request path or can use `len()` as a good enough estimate and avoid the processing and allocation overheads.

docs/architecture.md
```diff
@@ -208,16 +209,19 @@ Sets a header for use in disaggregated prefill/decode
...
-- **Type**: `pd-profile-handler`
+Selects the profiles to use when running with disaggregation
+
+- **Type**: `disagg-profile-handler`
 - **Parameters**:
   - `decodeProfile`: specifies the name of the profile used for the decode scheduling. Only needed if the decode profile is not named `decode`.
   - `prefillProfile`: specifies the name of the profile used for the prefill scheduling. Only needed if the prefill profile is not named `prefill`.
-  - `deciderPluginName`: specifies the name of the decider plugin. Decider determines whether disaggregated PD should be executed
-  - `primaryPort`: the base port number used for data parallel communication.
+  - `encodeProfile`: specifies the name of the profile used for the encode scheduling. Only needed if the encode profile is not named `encode`.
```
elevran: can you enable the other profile names use the same type of defaults (i.e., the "only needed if ..." section).
Should be consistent across the different profiles and not specific to encode

docs/architecture.md
```diff
@@ -208,16 +209,19 @@ Sets a header for use in disaggregated prefill/decode
...
-#### PdProfileHandler
+#### DisaggProfileHandler
 
-Selects the profiles to use when running with disaggregated prefill/decode
 
-- **Type**: `pd-profile-handler`
+Selects the profiles to use when running with disaggregation
+
+- **Type**: `disagg-profile-handler`
 - **Parameters**:
```
elevran, line 218: side (not necessarily in this PR). May want to refactor the parameter section (and allow backward compatibility for just the PD case), so that the new format is closer to a map (with value set to the key name if missing so you get the defaults from below):
```YAML
parameters:
   profiles:
   - encode: <name>
   - prefill: <name>
   - decode: <name>
  deciders: # not the best name?
  - encode: <name>
  - prefill <name>
```
If you choose to update the parameters section, you can marshal new format first and fallback to previous if that fails so you get backward compatibility (and log an info/warning for configuration debug)

docs/architecture.md
```diff
@@ -319,7 +321,16 @@ Filters out pods that are not marked either as decode or both prefill and decode
...
+Filters out pods that are not marked as prefill. The filter looks for the label `llm-d.ai/role`, with a value of `prefill`, `encode-prefil`, `prefill-decode` or `encode-prefill-decode`.
+
+- **Type**: `prefill-filter`
+- **Parameters**: None
+
+---
+
+#### EncodeFilter
+
+Filters out pods that are not marked as encode. The filter looks for the label `llm-d.ai/role`, with a value of `encode`, `encode-prefill` or `encode-prefill-decode`.
```
elevran: nit: pods missing a label altogether are not be filtered out but kept as candidates. Same as you had in the previous filter description?

docs/architecture.md
```diff
@@ -595,20 +609,22 @@ Requires the `prepareDataPlugins` feature gate and KV events from vLLM engines.
...
-## Disaggregated Prefill/Decode (P/D)
+## Disaggregated Encode/Prefill/Decode (E/P/D)
 
 When enabled, the router:
 
 - Selects one pod for **Prefill** (prompt processing)
 - Selects another pod for **Decode** (token generation)
 
-The **vLLM sidecar** handles orchestration between Prefill and Decode stages. It allows:
+*Note:* Note: Encode disaggregation is an experimental feature. When enabled, the router identifies all pods capable of encoding, and the vLLM sidecar distributes multimedia requests to randomly selected pods from that subset. More sophisticated selection strategies are planned for future versions.
```
elevran: - double `Note` at start
- consider using markdown NOTE as in other markdowns (e.g, DEVELOPMENT.md)
- break line, it's too long

docs/metrics.md
```diff
@@ -10,16 +10,31 @@ Metrics defined in llm-d-scheduler are in addition to Inference Gateway metrics.
...
 
 *   **Type:** Counter
 *   **Labels:**
-    *   `decision_type`: string ("decode-only" or "prefill-decode")
+    *   `model_name`: string (the target model name, or "unknown" if empty)
+    *   `decision_type`: string — one of:
+        *   `decode-only` — the request used the decode-only path (no disaggregation)
+        *   `prefill-decode` — the request was split into prefill and decode stages (P/D or EP/D)
+        *   `encode-decode` — the request used encode disaggregation with local prefill+decode (E/PD)
+        *   `encode-prefill-decode` — the request used the full three-stage pipeline (E/P/D)
```
elevran: nit: lose the em-dash'es?

docs/metrics.md
```diff
@@ -10,16 +10,31 @@ Metrics defined in llm-d-scheduler are in addition to Inference Gateway metrics.
...
+
+### `pd_decision_total` (deprecated)
+
+> **Deprecated:** Use `disagg_decision_total` instead.
+
+*   **Type:** Counter
+*   **Labels:**
+    *   `model_name`: string (the target model name, or "unknown" if empty)
+    *   `decision_type`: string ("decode-only" or "prefill-decode")
+*   **Release Stage:** ALPHA
```
elevran, line 38: Should there be an `ALPHA` designation associated with the `disagg_decisiton_total`?

pkg/metrics/metrics.go
```diff
@@ -27,22 +33,56 @@ var (
...
-	}
+	return []prometheus.Collector{SchedulerPDDecisionCount, SchedulerDisaggDecisionCount}
 }
 
 // RecordPDDecision increments the counter for a specific P/D routing decision.
-// The decisionType must be one of the DecisionType* constants (e.g., DecisionTypeDecodeOnly).
-// The model parameter should be the target model name (e.g., from request.TargetModel);
-// if empty, the caller should pass a placeholder like "unknown" to avoid empty labels.
+//
+// Deprecated: Use RecordDisaggDecision instead.
```
elevran, line 56: once merged, consider adding a tracking issue for the deprecation.

pkg/plugins/pre-request/encode_prerequest.go
```diff
@@ -0,0 +1,127 @@
...
+// compile-time type assertion
+var _ requestcontrol.PreRequest = &EncodeHeaderHandler{}
+
+// EncodeHeaderHandlerFactory defines the factory function for the EncodeHeaderHandler
+func EncodeHeaderHandlerFactory(name string, rawParameters json.RawMessage, _ plugin.Handle) (plugin.Plugin, error) {
+	parameters := encodeHeaderHandlerParameters{
+		EncodeProfile: defaultEncodeProfile,
+	}
+	if rawParameters != nil {
+		if err := json.Unmarshal(rawParameters, &parameters); err != nil {
```
elevran, line 40: Q: will this retain the default `EncodeProfile` value when the raw parameters do not have the field at all?
I think this is intended, confirming.

pkg/plugins/filter/epd_role.go
roytman: No, there is no such role. It's not a valid use case. The comment is updated.

pkg/plugins/profile/always_disagg_encode_decider.go
roytman: OK, renamed it

pkg/plugins/profile/decider-plugin.go
```diff
@@ -0,0 +1,15 @@
...
+	"context"
+
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/plugin"
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/scheduling"
+)
+
+// deciderPlugin decides whether the disaggregated stage should run for the request.
+type deciderPlugin interface {
+	plugin.Plugin
+	disaggregate(ctx context.Context, request *scheduling.LLMRequest, endpoint scheduling.Endpoint) bool
```
elevran, line 14: nit: shuold we consider exported interface and function or expect this to **always** implement inside this package?

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,269 @@
...
+	handler := NewDisaggProfileHandler(
+		parameters.DecodeProfile, parameters.PrefillProfile, parameters.EncodeProfile,
+		pdDecider, encodeDecider,
+	)
+	return handler.WithName(name), nil
+}
+
+// NewDisaggProfileHandler creates a DisaggProfileHandler directly.
+// Active stages are determined by which profile names are non-empty.
+func NewDisaggProfileHandler(
```
elevran: when possible, try to avoid breaking up function signature across multiple lines when the parameters can be bunched.

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,269 @@
...
+	decodeProfile, prefillProfile, encodeProfile string,
+	pdDecider deciderPlugin,
+	encodeDecider deciderPlugin,
+) *DisaggProfileHandler {
```
elevran:
```suggestion
func NewDisaggProfileHandler(decodeProfile, prefillProfile, encodeProfile string, pdDecider, encodeDecider deciderPlugin) *DisaggProfileHandler {
```

elevran: note that both pdDecider and encodeDecider are of the same type so no need to repeat (i.e. correctly done for the strings)

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -0,0 +1,269 @@
...
+// TypedName returns the typed name of the plugin.
+func (h *DisaggProfileHandler) TypedName() plugin.TypedName { return h.typedName }
+
+// WithName sets the instance name of the plugin.
+func (h *DisaggProfileHandler) WithName(name string) *DisaggProfileHandler {
+	h.typedName.Name = name
+	return h
+}
+
+func newDisaggProfileHandler(
```
elevran: try to avoid multiple lines in function signatures (throughout the files)

pkg/plugins/profile/multimodal_helpers.go
elevran, line 1: nit: if you don't expect additional helpers, consider folding this into another file (e.g., disagg_profile_handler.go)

pkg/plugins/profile/pd_profile_handler.go
```diff
@@ -25,24 +23,13 @@ import (
 )
 
 const (
-	// PdProfileHandlerType is the type of the PdProfileHandler
-	PdProfileHandlerType = "pd-profile-handler"
-
-	defaultDecodeProfile     = "decode"
-	defaultPrefillProfile    = "prefill"
+	// PdProfileHandlerType is a legacy alias for DisaggProfileHandlerType.
+	PdProfileHandlerType     = "pd-profile-handler"
```
elevran, line 27: Q: is it possible to have disagg profile actually a Go type alias for it and completely replace it? Just need to keep the legacy type name

pkg/plugins/profile/prefix_based_pd_decider.go
```diff
@@ -16,6 +16,10 @@ import (
 const (
 	// PrefixBasedPDDeciderPluginType is the type-name of the prefixBasedPDDecider plugin.
 	PrefixBasedPDDeciderPluginType = "prefix-based-pd-decider"
+
+	// AverageCharactersPerToken is an estimated average characters per token,
+	// used since the request we cache is not tokenized.
+	AverageCharactersPerToken = 4
```
elevran, line 22: Please see if this is already defines somewhere and can be reused instead of having the same constant in multiple places.

pkg/plugins/pre-request/encode_prerequest.go
roytman: Oh, thank you, fixed
Added the same check to `pd_prerequest.go`

pkg/plugins/profile/prefix_based_pd_decider.go
roytman: fixed

docs/architecture.md
roytman: Sorry, I don't understand you. 
All profile definitions have the "Only needed if the " section

pkg/plugins/profile/prefix_based_pd_decider.go
roytman: We did not define a new const; we just moved it from [dp_profile_handler](https://github.com/llm-d/llm-d-inference-scheduler/blob/fce0fff1e64df043080c3352d9cdda2807b539f9/pkg/plugins/profile/pd_profile_handler.go#L37) because:
- The handler is deprecated
- It is used by `PrefixBasedDecider` only.

docs/architecture.md
roytman: fixed. 
Yes, the roles are correct.

docs/architecture.md
roytman: done

pkg/plugins/pre-request/encode_prerequest.go
roytman: Actually, here we can check that `encodeProfileRunResult == nil ` only, and we don't have to check existence.

docs/architecture.md
elevran: Based on the description (and code?) it is ok to leave the name of the encode profile empty and it will defualt to `encode`. The same is not true for the other profile names? They should all have the same behavior.

pkg/plugins/profile/disagg_profile_handler.go
roytman: fixed

pkg/plugins/pre-request/encode_prerequest.go
roytman: Yes, you're right, updated `pd_prerequest.go` also.

docs/architecture.md
roytman: All profile names can be empty if they are the default. 
Here is the code
```goLang
func DisaggProfileHandlerFactory(name string, rawParameters json.RawMessage, handle plugin.Handle) 
(plugin.Plugin, error) {
   
  parameters := disaggProfileHandlerParameters{
	DecodeProfile:  defaultDecodeProfile,
	PrefillProfile: defaultPrefillProfile,
	EncodeProfile:  defaultEncodeProfile,
  }
  if rawParameters != nil {
		return nil, fmt.Errorf("failed to parse parameters of the disagg-profile-handler - %w", err)
	}
  }
  ...
```

The method description is updated

docs/architecture.md
roytman: The current implementation does support PD disaggregation. If the encodder decider is not defined but the prefill decider is, the system operates in P-D mode.

roytman: Here is an example from `architecture.md`
```yaml
kind: EndpointPickerConfig
plugins:
- type: prefill-header-handler
- type: prefix-cache-scorer
  parameters:
    maxPrefixBlocksToMatch: 256
    lruCapacityPerServer: 31250
- type: prefill-filter
- type: decode-filter
- type: max-score-picker
- type: prefix-based-pd-decider
    parameters:
      nonCachedTokens: 8
- type: disagg-profile-handler
  parameters:
    prefillDeciderPluginName: prefix-based-pd-decider
schedulingProfiles:
- name: prefill
  plugins:
  - pluginRef: prefill-filter
  - pluginRef: max-score-picker
    weight: 50
- name: decode
  plugins:
  - pluginRef: decode-filter
  - pluginRef: max-score-picker
    weight: 50
```

docs/architecture.md
roytman: added.

docs/metrics.md
roytman: What do you mean?

pkg/plugins/profile/disagg_profile_handler.go
roytman: EncodeProfile as all other profiles are always defined either by a user or with the default values. The factory doesn't have a logger or a Context to print a message.

pkg/plugins/profile/prefix_based_pd_decider.go
roytman: I'll check it separately.

docs/metrics.md
roytman: It is there - L:23

pkg/plugins/pre-request/encode_prerequest.go
roytman: Yes, it is for ALL profiles

pkg/plugins/profile/decider-plugin.go
roytman: I cannot answer this now. What is your opinion?

pkg/plugins/profile/disagg_profile_handler.go
roytman: fixed

pkg/plugins/profile/disagg_profile_handler.go
roytman: fixed

pkg/plugins/profile/multimodal_helpers.go
roytman: I do expect other encoding handlers.

pkg/plugins/profile/pd_profile_handler.go
roytman: Let's do it in a separate PR when e2e tests are ready

docs/architecture.md
elevran: thanks - the comment was more about changing the `parameters` object, not PD support. Maintaining backward compatible support for PD was in relation to changes in the config - allow both new and previous formats for a transition period

docs/metrics.md
elevran: The `—` (em-dash) and `-` (dash) are not the same character.

pkg/plugins/profile/decider-plugin.go
elevran: Leave it for now as private. We can export on first need with more confidence of the necessity,

pkg/plugins/pre-request/pd_prerequest.go
```diff
@@ -77,18 +77,23 @@ func (p *PrefillHeaderHandler) PreRequest(ctx context.Context, request *scheduli
...
+		)
+		return
 	}
-	if _, found := request.Headers[common.PrefillEndpointHeader]; found {
-		request.Headers[common.PrefillEndpointHeader] = "" // clear header, if already set
+
+	if request.TargetModel != "" {
+		span.SetAttributes(attribute.String("gen_ai.request.model", request.TargetModel))
 	}
+	span.SetAttributes(attribute.String("gen_ai.request.id", request.RequestId))
```
elevran, line 91: previous L83 checked that `RequestId != ""` before adding the attribute.

pkg/plugins/profile/disagg_profile_handler.go
shmuelk: Actually the plugin factory functions have a context with a logger. Look in the PluginHandle they receive

roytman: Thank you @shmuelk , added the log messages

pkg/plugins/pre-request/pd_prerequest.go
elevran: always present since GW adds it, so no need to check

Issue #725, opened 2026-03-17
Add scheduler disruption e2e tests


hexfusion:
**What would you like to be added:**

Disruption e2e tests for the scheduler. The existing e2e tests cover the happy path but not failure conditions. Using Ginkgo with `[Disruptive]` labels and `llm-d-inference-sim` on Kind.

Scenarios:
1. **Decode pod kill mid-request**: kill a decode pod while it's streaming. clean error or hang? does the EPP stop routing to it? (variant: OOMKill)
2. **EPP failure**: kill the EPP while requests are in flight. do routed requests finish? How long until recovery?
3. **Slow model server**:  one slow pod alongside normal ones. does the scheduler deprioritize it as the queue grows?
4. **Pool exhaustion**: delete all prefill pods in a P/D deployment. do warm cache requests still work? cold cache fall back or error? (variants: all decode gone, all pods gone)
5. **Graceful termination**:  rolling update while requests are in flight. does the pod drain cleanly? do clients see 503s?

**Why is this needed:**

There's no e2e coverage for scheduler behavior under failure. These scenarios cover the most common real-world disruptions OOM, rolling updates, slow pods, pool loss and will surface undefined behavior in the EPP's routing and recovery paths.

Related: #338

PR #735, opened 2026-03-18
test: add disruption e2e tests for scheduler failure scenarios

hexfusion:
Initial disruption test suite for the scheduler. Covers the most basic failure scenarios pod kills, pool exhaustion, EPP restart, and scale tozero under load. This is a starting point; more scenarios can be added incrementally.

Fixes: #725

## Summary

* Decode pod kill: force-delete mid-request, verify reroute to survivor and recovery
* Decode pod kill mid-stream: synchronize on response headers, kill the routed pod, assert no hang and routing recovery
* Pool exhaustion (503): scale to zero, poll until EPP returns 503, verify recovery
* EPP restart: kill EPP pod, verify requests fail during gap, verify recovery
* Scale-to-zero under load: background traffic loop during scale-down/up, verify failures then recovery

### DIFF
--- a/test/e2e/disruption_test.go
+++ b/test/e2e/disruption_test.go
@@ -102,7 +102,6 @@ var _ = ginkgo.Describe("Disruption tests", ginkgo.Ordered, ginkgo.Label("Disrup
 			gomega.Expect(nsHdr).Should(gomega.Equal(nsName))
 
 			targetPod := decodePods[0]
-			survivingPod := decodePods[1]
 
 			ginkgo.By("Force-deleting decode pod " + targetPod)
 			deletePodByName(targetPod, 0)
@@ -110,11 +109,11 @@ var _ = ginkgo.Describe("Disruption tests", ginkgo.Ordered, ginkgo.Label("Disrup
 			ginkgo.By("Waiting for killed pod to be replaced")
 			gomega.Eventually(podGone(targetPod, 1), podRemovalTimeout, 1*time.Second).Should(gomega.BeTrue())
 
-			ginkgo.By("Verifying new requests route to the surviving pod")
+			ginkgo.By("Verifying new requests route to a pod other than the killed one")
 			for range 5 {
 				nsHdr, podHdr, _ := runCompletion(simplePrompt, simModelName)
 				gomega.Expect(nsHdr).Should(gomega.Equal(nsName))
-				gomega.Expect(podHdr).Should(gomega.Equal(survivingPod))
+				gomega.Expect(podHdr).ShouldNot(gomega.Equal(targetPod))
 			}
 
 			ginkgo.By("Waiting for replacement pod to become ready")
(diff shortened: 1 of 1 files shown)


test/e2e/disruption_test.go
```diff
@@ -0,0 +1,424 @@
...
+			}
+		}
+		return len(currentDecode) >= minRemaining
+	}
+}
+
+// eppPodReady returns true when a new EPP pod (not oldPodName) is Running and Ready.
+func eppPodReady(oldPodName string) func() bool {
+	return func() bool {
+		pods := getPods(map[string]string{"app": "e2e-epp"})
```
elevran, line 67: nit: consider defining selector as package-level vars (e.g., `podSelector` and `prefillSelector` etc in e2e_test.go)

test/e2e/disruption_test.go
```diff
@@ -0,0 +1,424 @@
...
+			gomega.Eventually(func() int {
+				_, currentDecode := getModelServerPods(podSelector, prefillSelector, decodeSelector)
+				return len(currentDecode)
+			}, podRemovalTimeout, 1*time.Second).Should(gomega.Equal(0))
+
+			ginkgo.By("Waiting for traffic to observe failures")
+			gomega.Eventually(tc.failures, trafficProbeTimeout, 500*time.Millisecond).Should(gomega.BeNumerically(">", 0))
+
+			ginkgo.By("Scaling back to 1")
+			scaleDeployment(modelServers, 1)
```
elevran, line 316: is `gomega.Eventually(...` needed after scaleDeployment when scaling up?
Internally calls `podsInDeploymentsReady` before returning, so possibly the `getModelServerPods` is redundant in such cases?

test/e2e/disruption_test.go
```diff
@@ -0,0 +1,424 @@
...
+			targetPod := decodePods[0]
+			survivingPod := decodePods[1]
+
+			ginkgo.By(fmt.Sprintf("Force-deleting decode pod %s", targetPod))
+			deletePodByName(targetPod, 0)
+
+			ginkgo.By("Waiting for killed pod to be replaced")
+			gomega.Eventually(podGone(targetPod, 1), podRemovalTimeout, 1*time.Second).Should(gomega.BeTrue())
+
+			ginkgo.By("Verifying new requests route to the surviving pod")
```
elevran: I'm concerned we may have some flakiness here if the deployment controller creates a replacement pod before the requests are sent. `podGone` checks for `>= 1` which allows for either the survivor and a new replacement or just the surviving pod. 
If a repacement is started and ready "too quickly", the check on L117 might fail.
Probably not likely but perhaps be more on the the safe side?

test/e2e/disruption_test.go
```diff
@@ -0,0 +1,424 @@
...
+			scaleDeployment(modelServers, 1)
+
+			ginkgo.By("Waiting for pod to become ready")
+			gomega.Eventually(func() int {
+				_, currentDecode := getModelServerPods(podSelector, prefillSelector, decodeSelector)
+				return len(currentDecode)
+			}, readyTimeout, 2*time.Second).Should(gomega.Equal(1))
+
+			ginkgo.By("Verifying requests succeed after recovery")
+			nsHdr, _, _ = runCompletion(simplePrompt, simModelName)
```
elevran: small window of vulnerability (test sees Pod in L227 does not mean k8s API notification to EPP has completed or that the Pod was added to the EPP pool...).
Consider using some form of retry/eventually around the `runCompletion`

test/e2e/disruption_test.go
```diff
@@ -0,0 +1,424 @@
...
+
+			targetPod := <-connected
+			gomega.Expect(targetPod).ShouldNot(gomega.BeEmpty(),
+				"streaming request failed to connect")
+
+			ginkgo.By(fmt.Sprintf("Force-deleting decode pod %s during the stream", targetPod))
+			deletePodByName(targetPod, 0)
+
+			select {
+			case err := <-errCh:
```
elevran: should we inspect the error?
The error received from `io.ReadAll` is either:
  - io.EOF / connection reset by peer when pod died and TCP was severed (correct behavior under test)
  - context.DeadlineExceeded when client deadline fired while pod was still alive (incorrect behavior?)

test/e2e/disruption_test.go
```diff
@@ -0,0 +1,424 @@
...
+				if err != nil {
+					ginkgo.By(fmt.Sprintf("Stream terminated with error: %v", err))
+				} else {
+					ginkgo.By("Stream completed before pod was killed")
+				}
+			case <-time.After(disruptionClient.Timeout + 5*time.Second):
```
elevran: I think this would never be reached (`disruptionClient.Timeout` would always fire before `disruptionClient.Timeout` + 5s

PR #744, opened 2026-03-20
feat: precise-prefix-cache-scorer consumes tokenizer plugin

### LINKED ISSUE #743: `precise-prefix-cache-scorer` plugin should consume the new `tokenizer` plugin
With the recent merge of:
1. #659 
2. #694

We have a tokenizer plugin that extracts what the `precise-prefix-cache-scorer` plugin does internally. The first is an effort towards disaggregating tokenization and moving it to the router level, therefore, the latter should now use and depend on it, while supporting backwards compatibility incase the tokenizer plugin is not configured.

The `precise-prefix-cache-scorer` plugin should also be moved to the `multi` package, since it no longer only implements `Scorer`.

RishabhSaini:
resolves #743 

Move precise-prefix-cache-scorer from scorer to multi/precise_prefix_cache package, as it implements Scorer, PrepareDataPlugin, and PreRequest.

- Declare TokenizedPromptKey in Consumes() for tokenizer plugin dependency
- Add tokenProcessor field to use pre-tokenized data in computeBlockKeys()
- Fall back to internal tokenization when tokenizer plugin isn't configured
- Export CleanCachePeriodically and IndexedScoresToNormalizedScoredPods
- Add tests for tokenized fast path and fallback behavior


pkg/plugins/multi/precise_prefix_cache/precise_prefix_cache.go
```diff
@@ -114,24 +119,24 @@ func (s *precisePluginState) Clone() plugin.StateData {
...
 }
 
-// PrecisePrefixCachePluginFactory defines the factory function for creating
-// a new instance of the PrefixCacheTrackingPlugin.
-func PrecisePrefixCachePluginFactory(name string, rawParameters json.RawMessage,
+// PluginFactory defines the factory function for creating
+// a new instance of the PrecisePrefixCacheScorer plugin.
+func PluginFactory(name string, rawParameters json.RawMessage,
 	handle plugin.Handle,
 ) (plugin.Plugin, error) {
```
elevran:
```suggestion
func PluginFactory(name string, rawParameters json.RawMessage, handle plugin.Handle) (plugin.Plugin, error) {
```
nit: prefer not to break parameter list unless necessary

pkg/plugins/multi/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -311,40 +326,46 @@ type PrecisePrefixCacheScorer struct {
...
+func (s *Plugin) Produces() map[string]any {
 	return map[string]any{
 		dl_prefix.PrefixCacheMatchInfoKey: dl_prefix.PrefixCacheMatchInfo{},
 	}
 }
 
 // Consumes declares the data keys this plugin requires from other plugins.
-func (s *PrecisePrefixCacheScorer) Consumes() map[string]any {
-	return map[string]any{}
+// It declares an optional dependency on the tokenizer plugin's TokenizedPrompt.
```
elevran, line 352: Q: when we use `Produces()` for token stream generation and declare `Conusmes()` here, is it an optional relation or validated by data layer in GIE and fails when missing since the producer<-consumer contract is not valid? If enforced then we do not enable fallback.
Has this been vetted/tested?

@kfswain - what is the produces/consumes contract semantics?

pkg/plugins/multi/preciseprefixcache/precise_prefix_cache_test.go
```diff
@@ -1,6 +1,6 @@
 //go:build embedded_tokenizers
 
-package scorer
+package preciseprefixcache
```
elevran, line 3: nit: try to align the package and directory names?

pkg/plugins/multi/preciseprefixcache/precise_prefix_cache.go
vMaroon: @RishabhSaini can you please test?

pkg/plugins/multi/precise_prefix_cache/precise_prefix_cache.go
RishabhSaini: agreed, updated

pkg/plugins/multi/preciseprefixcache/precise_prefix_cache_test.go
RishabhSaini: agreed, updated

pkg/plugins/multi/preciseprefixcache/precise_prefix_cache.go
RishabhSaini: verified the three cases: 
Producer and Consumer Present: tokenizer ordered before scorer
Consumer Only (no tokenizer): no error
Producer Only (no consumer): no error

RishabhSaini: buildDAG only creates an edge and validates iff both producer AND consumer are configured
https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/pkg/epp/datalayer/data_graph.go#L126

elevran:
@RishabhSaini @vMaroon what's the status of this one? 
If still active, please resolve conflicts

ahg-g:
Addressed already

PR #749, opened 2026-03-22
[build] Add test coverage reporting

elevran:
Allows tracking test coverage (unit and integration).
Results are accessible in github job summary.


DEVELOPMENT.md
```diff
@@ -26,6 +26,69 @@ Documentation for developing the inference scheduler.
...
+make test-unit-sidecar  # sidecar only
+```
+
+### Unit Tests with Coverage
+
+```bash
+make test-coverage          # run all unit tests and report coverage
+make test-coverage-epp      # epp only
+make test-coverage-sidecar  # sidecar only
+```
```
hexfusion: question: the cover overhead should be pretty light. did you find it expensive to run unconditionally?

Makefile
```diff
@@ -155,10 +155,17 @@ test: test-unit test-e2e ## Run all tests (unit and e2e)
...
+# Set COVERAGE=1 to collect coverage profiles during unit tests (always set in CI via test-coverage-*)
+COVERAGE ?=
+
 .PHONY: test-unit-%
-test-unit-%: ## Run unit tests
+test-unit-%: ## Run unit tests (set COVERAGE=1 to also collect coverage profiles)
+	@$(if $(COVERAGE),mkdir -p $(COVERAGE_DIR))
 	@printf "\033[33;1m==== Running Unit Tests ====\033[0m\n"
-	@go test -v $$($($*_TEST_FILES) | tr '\n' ' ')
+	@go test -v \
```
hexfusion: nit: should we add `-race` ?

DEVELOPMENT.md
elevran: not really. Just wanted to allow both options if someone does not care for the coverage while iterating. Can remove and declutter

PR #750, opened 2026-03-22
DP-aware routing with X-data-parallel-rank header injection

satyamg1620:
## Summary

Enables the scheduler to route requests to a **specific DP rank** in vLLM's Internal LB and Hybrid LB modes by injecting the `X-data-parallel-rank` HTTP header. Previously, per-rank KV cache scores were collapsed to pod-level and the rank information was lost, relying on vLLM's internal queue-based LB to pick a rank. This change preserves the winning rank through the scoring pipeline and injects it as a header so vLLM pins the request to the rank with the best KV cache match.

Relates to: https://github.com/llm-d/llm-d-kv-cache/issues/357
Companion PR: https://github.com/llm-d/llm-d-kv-cache/pull/370

**NOTE:** First this PR https://github.com/llm-d/llm-d-kv-cache/pull/370 needs to be merged

### DIFF
--- a/pkg/common/dp.go
+++ b/pkg/common/dp.go
@@ -2,6 +2,8 @@ package common
 
 import (
 	"encoding/json"
+	"errors"
+	"fmt"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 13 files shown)


pkg/common/dp.go
```diff
@@ -0,0 +1,110 @@
...
+// and data parallel rank. Scoring keys from the KV cache indexer use the
+// format "<podIdentifier>@dp<rank>".
+//
+// Examples:
+//
+//	"10.0.0.1:8080"       -> ("10.0.0.1:8080", -1)
+//	"10.0.0.1:8080@dp0"   -> ("10.0.0.1:8080", 0)
+//	"10.0.0.1:8080@dp3"   -> ("10.0.0.1:8080", 3)
+func ParseDPScoringKey(scoringKey string) (podIdentifier string, dpRank int) {
+	idx := strings.LastIndex(scoringKey, DPRankSuffix)
```
modassarrana89-new, line 49: This will incorrectly parse if:
"pod@dp-service:8080"

Suggestion : to use strict suffix check

pkg/common/dp.go
```diff
@@ -0,0 +1,110 @@
...
+	if idx < 0 {
+		return scoringKey, NoDataParallelRank
+	}
+
+	rankStr := scoringKey[idx+len(DPRankSuffix):]
+	rank, err := strconv.Atoi(rankStr)
```
modassarrana89-new, line 59: No validation of negative rank

pkg/common/dp.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return "", err
+	}
+	return string(data), nil
+}
+
+// DecodeWinningRanks deserializes a JSON string from an HTTP header back into a winning ranks map.
+func DecodeWinningRanks(encoded string) (map[string]int, error) {
+	var ranks map[string]int
+	if err := json.Unmarshal([]byte(encoded), &ranks); err != nil {
```
modassarrana89-new, line 145: No validation of empty map

pkg/plugins/pre-request/dp_rank_header_handler.go
```diff
@@ -0,0 +1,131 @@
...
+		return
+	}
+
+	targetPod := primaryResult.TargetEndpoints[0].GetMetadata()
+	if targetPod == nil {
+		span.SetAttributes(attribute.Bool("llm_d.epp.dp.rank_header_set", false))
+		return
+	}
+
+	podAddress := fmt.Sprintf("%s:%s", targetPod.GetIPAddress(), targetPod.GetPort())
```
modassarrana89-new: This assumes exact string match with scorer
But earlier you defined:
```
"<podIdentifier>@dp<rank>"
```
And scorer might use:
```
hostname
IP
service DNS
port formatting differences
```

pkg/plugins/pre-request/dp_rank_header_handler.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	// Get the selected pod's address from the primary profile result.
+	primaryResult := schedulingResult.ProfileResults[schedulingResult.PrimaryProfileName]
+	if primaryResult == nil || len(primaryResult.TargetEndpoints) == 0 {
+		span.SetAttributes(attribute.Bool("llm_d.epp.dp.rank_header_set", false))
+		return
+	}
+
+	targetPod := primaryResult.TargetEndpoints[0].GetMetadata()
```
modassarrana89-new, line 124: It Assumes:

- only 1 endpoint OR
- first is always chosen

modassarrana89-new:
@satyamg1620 
You have defined:
```
DPModeInternalLB
DPModeHybridLB
DPModeExternalLB
```
But plugin ignores it.

ahg-g:
Is this still relevant?

satyamg1620:
> Is this still relevant?

It is required for specific DP rank in vLLM's Internal LB and Hybrid LB modes. This is follow up PR of https://github.com/llm-d/llm-d-kv-cache/pull/370

ahg-g:
Can you explain why we need this? For DP-aware routing, don't we route to a specific port? each engine running in the pod runs on a separate port and is treated as an independent endpoint.

satyamg1620:
> Can you explain why we need this? For DP-aware routing, don't we route to a specific port? each engine running in the pod runs on a separate port and is treated as an independent endpoint.

That holds for External LB mode — each DP rank is its own port/endpoint, the scheduler already picks it directly, and this PR is a no-op there (kv-cache returns NoDataParallelRank, so no header is injected).

This targets Internal LB and Hybrid LB, where multiple DP ranks share a single HTTP port and vLLM balances across ranks internally. There's no per-rank port to route to — the only way to steer a request to the rank that already holds the prefix's KV cache is the x-data-parallel-rank header, which vLLM honors to bypass its internal LB.

This is the scheduler-side half of llm-d/llm-d-kv-cache#370 (resolves llm-d/llm-d-kv-cache#357), which threads data_parallel_rank through the index and scorer so a pod's ranks get independent KV scores.

PR #751, opened 2026-03-22
[refactor] Unify sidecar `Config` and `Options`

elevran:
Follows up on review feedback in #697.
This PR makes `Config` the single source of truth for proxy runtime configuration, and embeds it in `Options`.

As Config was also incomplete (e.g., port, target URL, and SSRF fields were passed as separate arguments to NewProxy and NewAllowlistValidator), so main.go had to manually assemble them after Complete(). Since Config is now part of Options, Complete() populates it directly.                                                                                                                        
                                                                                                                                                         
Key changes:
- Config now includes Port, TargetURL, SSRF fields, and consistently named TLS fields
 - Options embeds Config; flags that need no transformation bind directly to the embedded fields
 - NewProxy(Config) and Start(ctx) instead of passing no separate args
 - main.go reduces to NewProxy(opts.Config) + Start(ctx)
 - all tests changed to new structure


cmd/pd-sidecar/main.go
```diff
@@ -70,55 +68,11 @@ func main() {
...
-	// Create SSRF protection validator
-	validator, err := proxy.NewAllowlistValidator(opts.EnableSSRFProtection, opts.PoolGroup, opts.InferencePoolNamespace, opts.InferencePoolName)
-	if err != nil {
-		logger.Error(err, "failed to create SSRF protection validator")
-		return
-	}
+	logger.Info("Proxy configuration", "config", opts.Config)
 
-	proxyServer := proxy.NewProxy(opts.Port, targetURL, config)
+	proxyServer := proxy.NewProxy(opts.Config)
```
hexfusion, line 71: 👍

cmd/pd-sidecar/main.go
```diff
@@ -70,55 +68,11 @@ func main() {
...
 
-	if err := proxyServer.Start(ctx, validator); err != nil {
+	if err := proxyServer.Start(ctx); err != nil {
```
hexfusion, line 72: 👍

pkg/sidecar/proxy/proxy.go
```diff
@@ -142,14 +177,14 @@ type Server struct {
...
+// NewProxy creates a new routing reverse proxy from the given Config.
+func NewProxy(config Config) *Server {
 	prefillerCache, _ := lru.New[string, http.Handler](16) // nolint:all
 	encoderCache, _ := lru.New[string, http.Handler](16)   // nolint:all
 
 	server := &Server{
-		port:                port,
-		decoderURL:          decodeURL,
+		port:                config.Port,
+		decoderURL:          config.TargetURL,
```
hexfusion: nit: we pass the full config to the server do we need these fields?

elevran: good catch - removed.

PR #752, opened 2026-03-22
pickup Gateway API 1.5.1 and GIE 1.4.0

Gregory-Pereira:


elevran:
GAIE 1.4 brought in #770

PR #756, opened 2026-03-24
Prevent mismatch between new and deprecated APIs

roytman:
Prevent mismatch between new and legacy `DisaggProfileHandler` APIs.
One of the comments that was provided to #732


pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -42,94 +43,87 @@ type disaggDecidersParameters struct {
...
-func (p *disaggProfileHandlerParameters) validateAndApplyDeprecatedFields(logger interface{ Info(string, ...any) }) error {
-	if p.hasDeprecatedFields() && p.hasNestedFields() {
-		return errors.New("cannot mix deprecated flat parameters (decodeProfile, prefillProfile, encodeProfile, " +
-			"deciderPluginName, prefillDeciderPluginName, encodeDeciderPluginName) " +
-			"with nested parameters (profiles, deciders): use one format or the other")
-	}
-	if p.DeprecatedDecodeProfile != "" {
+// toDisaggParams copies legacy flat fields into the nested format, logging a
+// deprecation warning for each field in use.
+func (l *legacyDisaggProfileHandlerParameters) toDisaggParams(logger interface{ Info(string, ...any) }) disaggProfileHandlerParameters {
```
shmuelk: Please import `"github.com/go-logr/logr"`


```suggestion
func (l *legacyDisaggProfileHandlerParameters) toDisaggParams(logger logr.Logger) disaggProfileHandlerParameters {
```

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -42,94 +43,87 @@ type disaggDecidersParameters struct {
...
+			}
+			// Detect mixed format: also check whether any nested fields were present.
+			var nested disaggProfileHandlerParameters
+			if _ = json.Unmarshal(rawParameters, &nested); nested.Profiles != (disaggProfilesParameters{}) || nested.Deciders != (disaggDecidersParameters{}) {
+				return nil, errors.New("cannot mix deprecated flat parameters (decodeProfile, prefillProfile, encodeProfile, " +
+					"deciderPluginName, prefillDeciderPluginName, encodeDeciderPluginName) " +
+					"with nested parameters (profiles, deciders): use one format or the other")
+			}
+			logger.Info("Deprecated: using flat parameter format, migrate to nested profiles/deciders format")
+			parameters = legacy.toDisaggParams(logger)
```
shmuelk: I would use something simpler here:
legacy := legacyDisaggProfileHandlerParameters{}

   return nil, fmt.Errorf("failed to parse parameters of the disagg-profile-handler - %w", err)
}
if err := json.Unmarshal(rawParameters, &legacy); err != nil {
   return nil, fmt.Errorf("failed to parse parameters of the disagg-profile-handler - %w", err)
}

if parameters.Profiles != (disaggProfilesParameters{}) ||
                parameters.Deciders != (disaggDecidersParameters{}) {
    // Make sure the legacy parameters were not used
    if legacy != (legacyDisaggProfileHandlerParameters{}) {
       return nil, errors.New("cannot mix deprecated flat parameters (decodeProfile, prefillProfile, encodeProfile, " +
					"deciderPluginName, prefillDeciderPluginName, encodeDeciderPluginName) " +
					"with nested parameters (profiles, deciders): use one format or the other")
	}
} else {
    logger.Info("Deprecated: using flat parameter format, migrate to nested profiles/deciders format")
	parameters = legacy.toDisaggParams(logger)

pkg/plugins/profile/disagg_profile_handler.go
roytman: done

pkg/plugins/profile/disagg_profile_handler.go
roytman: done, thank you

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -42,92 +43,87 @@ type disaggDecidersParameters struct {
...
 	}
-	if p.DeprecatedPrefillDeciderPluginName != "" {
+	if l.PrefillDeciderPluginName != "" {
 		logger.Info("Deprecated parameter 'prefillDeciderPluginName', use 'deciders.prefill' instead")
-		p.Deciders.Prefill = p.DeprecatedPrefillDeciderPluginName
+		p.Deciders.Prefill = l.PrefillDeciderPluginName
 	}
-	if p.DeprecatedDeciderPluginName != "" {
+	// DeciderPluginName is a lower-priority alias for prefill decider (from pd-profile-handler).
+	if l.DeciderPluginName != "" && p.Deciders.Prefill == "" {
```
elevran, line 86: nit: consider adding a test with both set to make sure any future changes do not break the assumption that `DeciderPluginName` has lower priority?

pkg/plugins/profile/disagg_profile_handler.go
```diff
@@ -42,92 +43,87 @@ type disaggDecidersParameters struct {
...
-
-	logger := log.FromContext(handle.Context())
+		if err := json.Unmarshal(rawParameters, &legacy); err != nil {
+			return nil, fmt.Errorf("failed to parse parameters of the disagg-profile-handler - %w", err)
+		}
 
-	// Promote any deprecated flat fields that are still in use.
-	if err := parameters.validateAndApplyDeprecatedFields(logger); err != nil {
-		return nil, err
+		if parameters.Profiles != (disaggProfilesParameters{}) ||
```
elevran, line 115: nit: consider adding a comment on comparing with zero value parameters in case we add some field with "non-zero" default in the future?

pkg/plugins/profile/disagg_profile_handler.go
roytman: The current implementation doesn't give old params lower priority — it errors when both are set simultaneously. The test that covers this is in disagg_profile_handler_test.go:263

```goLang
{"mixing deprecated and nested fields is an error", map[string]any{
    "decodeProfile": "my-decode",
    "profiles":      map[string]any{"decode": "other-decode"},
}, true},
{"mixing deprecated decider and nested deciders is an error", map[string]any{
    "prefillDeciderPluginName": PrefixBasedPDDeciderPluginType,
    "deciders":                 map[string]any{"prefill": AlwaysDisaggPDDeciderPluginType},
}, true},
```

pkg/plugins/profile/disagg_profile_handler.go
roytman: We set non-zero parameters (profile names) after this test. In the future, we will probably remove support the legacy API.

PR #757, opened 2026-03-25
[cicd] Compare coverage against `main` and latest `release-*`

elevran:
- Upload release coverage as artifact for future reference
- On every PR merge to main, compare coverage vs main and latest release branch. Both reports are avaialble in the action output.


shmuelk:
@elevran  Please fix conflicts

PR #765, opened 2026-03-26
Replace time.Sleep in the e2e tests with deterministic pod-discovery wait

### LINKED ISSUE #764: fix(e2e): replace time.Sleep with deterministic pod-discovery wait
The e2e tests have a race condition between the EPP's gRPC health check reporting SERVING and the EPP's InferencePool controller completing its initial pod discovery.

The EPP sets its health status to SERVING the moment the gRPC server starts (in `runserver.go`), before the Kubernetes controller manager has finished reconciling the InferencePool and populating the internal pod datastore. The tests detect SERVING and then rely on a fixed time.Sleep (2 s in most cases, 5 s in the KV tests) to give the EPP time to discover pods. When the pod discovery takes longer than the sleep — which is non-deterministic and depends on CI load — the first completion request returns:

```
503 Service Unavailable
inference error: ServiceUnavailable - failed to find candidate pods for serving the request
```

roytman:
Replace all fixed sleeps with a gomega.Eventually loop that sends a warmup POST /v1/completions request and retries until the response is not the "failed to find candidate pods" 503. This directly tests the condition the sleep was approximating, making the tests deterministic regardless of cluster or CI load.

The polling logic is extracted into a shared waitForEPPToDiscoverPods(modelName string) helper in utils_test.go and used in three places:

- createEndPointPicker — EPP created after model servers (standard flow)
- KV enabled configuration test — EPP created before model servers
- KV external tokenizer test — same as above

The `gomega.Eventually(func, 30*time.Second, time.Second)` in `waitForEPPToDiscoverPods` has a 30-second timeout with a 1-second polling interval. If the EPP doesn't discover pods within 30 seconds, the test fails with a clear timeout error.

/close #764

### DIFF
--- a/test/e2e/utils_test.go
+++ b/test/e2e/utils_test.go
@@ -45,7 +45,10 @@ func waitForEPPToDiscoverPods(modelName string) {
 		if resp.StatusCode != http.StatusServiceUnavailable {
 			return true
 		}
-		body, _ := io.ReadAll(resp.Body)
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return false
+		}
 		return !strings.Contains(string(body), "failed to find candidate pods")
 	}, 30*time.Second, time.Second).Should(gomega.BeTrue())
 }
--- a/test/e2e/utils_test.go
+++ b/test/e2e/utils_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net/http"
 	"os/exec"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


test/e2e/utils_test.go
```diff
@@ -21,6 +23,31 @@ const (
...
+// controller has finished its initial pod discovery.
+// This is necessary because the EPP sets its health status to SERVING as soon
+// as the gRPC server starts, before pod discovery is complete.
+func waitForEPPToDiscoverPods(modelName string) {
+	ginkgo.By("Waiting for EPP to discover pool members")
+	gomega.Eventually(func() bool {
+		resp, err := http.Post(
+			fmt.Sprintf("http://localhost:%s/v1/completions", port),
+			"application/json",
+			strings.NewReader(`{"model":"`+modelName+`","prompt":"warmup"}`),
```
nirrozenbaum, line 39: looks good. 
a question (not really related to this PR, but the solution is similar) - 
atm, when a new pod is started it's marked as ready before it can actually serve requests.
what are your thoughts about adding readiness check that does exactly this warmup (could be just `/v1/models` to avoid prompt need).
then the e2e tests can just check if the pod is ready.

roytman: I'll check this option separately.

shmuelk: The E2E tests use the llm-d-inference-sim, which if used with "fake" models should be serving almost immediately.

test/e2e/utils_test.go
```diff
@@ -21,6 +23,31 @@ const (
 	deploymentKind = "deployment"
 )
 
+// waitForEPPToDiscoverPods polls the EPP's completions endpoint until it no
+// longer returns "failed to find candidate pods", indicating the InferencePool
+// controller has finished its initial pod discovery.
+// This is necessary because the EPP sets its health status to SERVING as soon
+// as the gRPC server starts, before pod discovery is complete.
+func waitForEPPToDiscoverPods(modelName string) {
```
acardace, line 33: By checking the epp health code I actually see that the probe doesn't switch to serving until the endpoints are discovered by the InferencePool https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/cmd/epp/runner/health.go#L47 https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/pkg/epp/controller/inferencepool_reconciler.go#L90. So maybe just removing the sleep is fine?

nirrozenbaum:
Thanks @roytman. 
I was actually asking about having a readiness probe doing a warm up in real deployments, not in the tests themselves. 
as @shmuelk mentioned, in e2e we usually use sim. but with real vllm, the warmup is a real problem

PR #770, opened 2026-03-26
Revert IGW import to 1.4.0 and update tokenizer plugin accordingly

vMaroon:
## Summary
- import IGW 1.4.0
- turn tokenizer PrepareData plugin into a scorer and wrap the latter in no-build flag

### DIFF
--- a/pkg/plugins/preparedata/tokenizer_scorer.go
+++ b/pkg/plugins/preparedata/tokenizer_scorer.go
@@ -37,9 +37,8 @@ func (p *TokenizerPlugin) Category() scheduling.ScorerCategory {
 }
 
 // Score implements scheduling.Scorer. It tokenizes the request prompt, writes
-// the result to CycleState using the plugin's TypedName as key (standard IGW
-// pattern — same as how NoHitLRU reads prefix-cache scorer data), and returns
-// zero scores for all endpoints since this scorer only produces data.
+// the result to CycleState under the TokenizedPromptStateKey constant, and
+// returns zero scores for all endpoints since this scorer only produces data.
 func (p *TokenizerPlugin) Score(ctx context.Context, cycleState *scheduling.CycleState, request *scheduling.LLMRequest, pods []scheduling.Endpoint) map[scheduling.Endpoint]float64 {
 	logger := log.FromContext(ctx).WithName(p.typedName.String())
 	traceLogger := logger.V(logutil.TRACE)
--- a/pkg/plugins/preparedata/tokenizer_scorer.go
+++ b/pkg/plugins/preparedata/tokenizer_scorer.go
@@ -1,0 +1,66 @@
+//go:build !gaie_tokenized_prompt
+
+/*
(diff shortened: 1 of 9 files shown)


pkg/plugins/preparedata/tokenizer_scorer.go
```diff
@@ -0,0 +1,66 @@
...
+// Category implements scheduling.Scorer.
+// The tokenizer scorer is an Affinity scorer since it produces data consumed
+// by cache-locality scorers.
+func (p *TokenizerPlugin) Category() scheduling.ScorerCategory {
+	return scheduling.Affinity
+}
+
+// Score implements scheduling.Scorer. It tokenizes the request prompt, writes
+// the result to CycleState using the plugin's TypedName as key (standard IGW
+// pattern — same as how NoHitLRU reads prefix-cache scorer data), and returns
```
elevran: Inaccurate doc comment: key is `TokenizedPromptStateKey`, not the plugin's type name.

pkg/plugins/preparedata/tokenizer_scorer_test.go
```diff
@@ -0,0 +1,180 @@
...
+func TestTokenizerScorer_SkipsWhenAlreadyInCycleState(t *testing.T) {
+	ctx := utils.NewTestContext(t)
+	cycleState := scheduling.NewCycleState()
+
+	// Pre-populate CycleState.
+	existing := &TokenizedPromptState{TokenIDs: []uint32{1, 2, 3}}
+	cycleState.Write(TokenizedPromptStateKey, existing)
+
+	// Use nil tokenizer — would panic if called.
+	p := newTestPlugin(nil)
```
elevran: nit: nil tokenizer panics on failure instead of producing a clear assertion. The panic gives a cryptic backtrace rather than a clear test failure, and is fragile to refactors of the skip guard. Perhaps use a mock tokenizer that records whether it was called? This could give a readable failure message and makes the invariant explicit.

elevran (commented):
Thanks @vMaroon!
/lgtm
Not approving in case you want to address copilot feedback

pkg/plugins/preparedata/tokenizer_scorer_test.go
vMaroon: Addressed (and other comments).

hexfusion:
@vMaroon @elevran what was the reason for this revert can we provide some clarity?

elevran:
Some days after v1.4 was branched, the GAIE Request object was extended with tokenized prompt (`[]int`). The inference scheduler imported a pseudo version that includes the tokenized prompt and a prepare-data plugin point that generates tokens for use by Scorers.
For the Scehduler's v0.7 release we had to pick up
1. a **tested** v1.4 (which meant reverting back to request structure without tokenized prompt); or
1. an untested pseudo version (with the updated struct).

We opted for the release version that was tested and will reenable tokenized prompt when we get v1.5.

PR #791, opened 2026-04-01
[cicd] Add release notes action

elevran:
Capture release note in PULL_REQUEST_TEMPLATE and add them to a release notes file.

### DIFF
--- a/.github/workflows/ci-signed-commits.yaml
+++ b/.github/workflows/ci-signed-commits.yaml
@@ -7,7 +7,7 @@ jobs:
   check-signed-commits:
     name: Check signed commits in PR
     if: |
-      !startsWith(github.head_ref, 'changelog/pr-') ||
+      !startsWith(github.head_ref, 'release-notes/pr-') ||
       github.event.pull_request.user.login != 'github-actions[bot]'
     runs-on: ubuntu-latest
     permissions:
--- a/.github/workflows/ci-signed-commits.yaml
+++ b/.github/workflows/ci-signed-commits.yaml
@@ -6,6 +6,9 @@ on:
 jobs:
   check-signed-commits:
     name: Check signed commits in PR
+    if: |
+      !startsWith(github.head_ref, 'changelog/pr-') ||
+      github.event.pull_request.user.login != 'github-actions[bot]'
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/.github/workflows/release-notes-update.yaml
+++ b/.github/workflows/release-notes-update.yaml
@@ -1,4 +1,4 @@
-name: Update Changelog on PR Merge
+name: Update Release Notes on PR Merge
 
 on:
   pull_request:
@@ -7,7 +7,7 @@ on:
       - main
 
 jobs:
-  update-changelog:
+  update-release-notes:
     if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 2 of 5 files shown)


.github/workflows/release-notes-update.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+          BRANCH="changelog/pr-${PR_NUMBER}"
+          git checkout -b "$BRANCH"
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add CHANGELOG.md
+          git commit -s -m "docs: update CHANGELOG for PR #${PR_NUMBER}"
+          git push origin "$BRANCH"
+
+          gh pr create \
```
shmuelk, line 49: This doubles the number of PRs that need to be reviewed. Is that ok?

.github/workflows/release-notes-update.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Update CHANGELOG.md and open PR
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_BODY: ${{ github.event.pull_request.body }}
+          PR_URL: ${{ github.event.pull_request.html_url }}
+          PR_MERGED_AT: ${{ github.event.pull_request.merged_at }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          NOTE=$(printf '%s' "$PR_BODY" | tr -d '\r' | \
+            sed -n '/^```release-note/,/^```/p' | sed '1d;$d' | xargs)
```
shmuelk, line 31: Does this strip white space around the release note to insure tat it found `NONE`?

.github/workflows/release-notes-update.yaml
elevran: the changelog/release notes require a review of their own (instead of just taking the PR content as is). 
If you find a way to add the change log file to the current PR that would be preferred. I wasn't sure how to do so cleanly.

.github/workflows/release-notes-update.yaml
elevran: extracts the text in the back tick block, folding lines, if needed. It does not strip extra whitespece. Since the default already is set to `NONE` in the PR comment, I expect it be retained when not needed. The changelog file is also reviewed so we'll capture the off chance it slips through (i.e. a use replacing the default NONE with their own version that has extra whitespace).

.github/workflows/ci-signed-commits.yaml
```diff
@@ -6,6 +6,9 @@ on:
 jobs:
   check-signed-commits:
     name: Check signed commits in PR
+    if: |
```
elevran, line 9: this avoids checking signed commits on action generated review of the changelog

shmuelk:
This PR is not building a changelog. It is building release notes, they are not the same thing.

elevran:
> This PR is not building a changelog. It is building release notes, they are not the same thing.

That's a valid distinction.
What is the required change? Replace terminology to use `release notes` consistently or something else?

elevran:
@shmuelk changed to release notes throughout
PTAL

.github/workflows/release-notes-update.yaml
elevran: (and the added file would be missing a signature...)

shmuelk:
nit: I think you should add something to `.github/workflows/ci-release.yaml` that adds text to the top of the release notes file, indicating the release involved.

I will place a hold and let you decide

Issue #792, opened 2026-04-01
Add disaggregation decider evaluation metrics


wenhug:
## Summary

The `prefix-based-pd-decider` plugin makes disaggregation decisions with multiple rejection paths, but today there is no visibility into **why** a request was or was not disaggregated. The existing `disagg_decision_total` metric records the final routing outcome (decode-only, prefill-decode, etc.) but not the decider's reasoning.

## Problem

When operators see a high rate of `decode-only` decisions, they cannot distinguish between:
- Disaggregation is **disabled** (`NonCachedTokens=0`)
- Input is **too short** to justify disaggregation
- The prompt is **sufficiently cached**, making disaggregation unnecessary
- An **error** prevented the decision (nil endpoint, missing prefix cache state)

This makes it difficult to tune `NonCachedTokens` thresholds or diagnose misconfiguration.

## Proposal

Add a new Prometheus counter `llm_d_inference_scheduler_decider_evaluation_total` with labels:
- `model_name` — target model
- `decider` — decider plugin name (e.g. `prefix-based-pd-decider`)
- `reason` — one of: `disabled`, `input_too_short`, `suffix_cached`, `error`, `disaggregated`

This complements `disagg_decision_total` (which records *what* happened) by exposing *why* the decider accepted or rejected disaggregation.

## Scope

- Add metric definition and recording helper in `pkg/metrics/`
- Instrument `PrefixBasedPDDecider.disaggregate()` at each return path
- Add unit tests
- No behavior change — observability only

PR #793, opened 2026-04-01
feat: add disaggregation decider evaluation metrics

wenhug:
## Summary

Add a new Prometheus counter `llm_d_inference_scheduler_decider_evaluation_total` that records the reason for each disaggregation decider evaluation.

The existing `disagg_decision_total` metric captures the **final routing outcome** (decode-only, prefill-decode, etc.) but not **why** the decider accepted or rejected disaggregation. This makes it difficult for operators to tune `NonCachedTokens` thresholds or diagnose misconfiguration.

The new metric complements `disagg_decision_total` with reason labels:
- `disabled` — disaggregation is turned off (`NonCachedTokens=0`)
- `input_too_short` — total input tokens below threshold
- `suffix_cached` — non-cached suffix below threshold (cache hit is sufficient)
- `error` — nil endpoint, missing prefix cache state, or input parse failure
- `disaggregated` — decider accepted disaggregation

### Changes
- Add `DeciderEvaluationCount` counter and `RecordDeciderEvaluation()` helper in `pkg/metrics/`
- Instrument `PrefixBasedPDDecider.disaggregate()` at each return path
- Add unit tests for the new metric
- **No behavior change — observability only**

Ref #792

## Test plan
- [x] `go build ./...` passes
- [x] `go test -race ./pkg/metrics/...` passes
- [x] `go test -race ./pkg/plugins/profile/...` passes
- [x] New metric tests validate all 5 reason labels and empty model fallback


pkg/metrics/metrics.go
```diff
@@ -44,11 +57,24 @@ var (
...
+	// decider accepted or rejected disaggregation, enabling operators to
+	// distinguish threshold misses from errors or disabled configurations.
+	DeciderEvaluationCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: SchedulerSubsystem,
+			Name:      "decider_evaluation_total",
+			Help:      metrics.HelpMsgWithStability("Total number of disaggregation decider evaluations by reason", compbasemetrics.ALPHA),
+		},
+		[]string{"model_name", "decider", "reason"},
+	)
```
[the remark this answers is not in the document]
wenhug: Fixed in 82a3578 — renamed to `SchedulerDeciderEvaluationCount` for consistency.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -141,9 +155,11 @@ func (d *PrefixBasedPDDecider) disaggregate(ctx context.Context, request *schedu
 
 	if nonCachedTokens < d.config.NonCachedTokens {
 		debugLogger.Info("Non-cached suffix is smaller than threshold, using decode profile only")
+		metrics.RecordDeciderEvaluation(modelName, deciderName, metrics.DeciderReasonSuffixCached)
 		return false // do not run prefill
 	}
 
+	metrics.RecordDeciderEvaluation(modelName, deciderName, metrics.DeciderReasonDisaggregated)
 	return true
```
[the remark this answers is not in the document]
wenhug: Fixed in 82a3578 — added `TestDisaggregateMetricReasons` covering all 6 return paths (disabled, error, input_too_short, suffix_cached, disaggregated, missing prefix info).

pkg/metrics/metrics.go
```diff
@@ -44,11 +57,24 @@ var (
...
 	)
+
+	// DeciderEvaluationCount records the reason for each disaggregation decider
+	// evaluation. This complements disagg_decision_total by exposing *why* the
+	// decider accepted or rejected disaggregation, enabling operators to
+	// distinguish threshold misses from errors or disabled configurations.
+	SchedulerDeciderEvaluationCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: SchedulerSubsystem,
+			Name:      "decider_evaluation_total",
```
hexfusion: question: would it make sense to extend the existing `disagg_decision_total` with a reason label instead of a new counter?  `disagg_decision_total` and `decider_evaluation_total` record two facets of the same event

```
disagg_decision_total{decision_type="decode-only", reason="suffix_cached"}
disagg_decision_total{decision_type="decode-only", reason="input_too_short"}
```

shmuelk: +!

wenhug: My thinking was that `disagg_decision_total` is the final routing result, while this is the decider-level reason.
Since we can have multiple deciders/stages, putting `reason` on the final result may be ambiguous. But if you prefer one metric, I can rework it

ahg-g:
can you pls rebase?

pkg/metrics/llm_d_router_metrics.go
```diff
@@ -34,6 +34,16 @@ var (
 		[]string{"model_name", "decision_type"},
 	)
 
+	// LlmdDeciderEvaluationCount records disaggregation decider evaluations by reason.
+	LlmdDeciderEvaluationCount = prometheus.NewCounterVec(
```
ahg-g: This metric should be defined and registered with the plugin that emits it, pls see the metrics emitted by the approximate prefix cache producer as an example: https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/metrics.go
```diff
@@ -0,0 +1,73 @@
...
+	// deciderReasonError indicates disaggregation was skipped due to an error.
+	deciderReasonError = "error"
+	// deciderReasonDisaggregated indicates the decider chose to disaggregate.
+	deciderReasonDisaggregated = "disaggregated"
+)
+
+var (
+	schedulerDeciderEvaluationCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: schedulerSubsystem,
```
ahg-g: If those are new metrics, we don't want to emit them using the deprecated `llm_d_inference_scheduler` subsystem, we just need the new one.

ahg-g (dismissed):

wenhug:
@ahg-g gentle ping — I addressed the previous metrics feedback and rebased this PR. Could you take another look when you have a chance?

PR #795, opened 2026-04-02
[feat] [cicd] Use distroless/static as default runtime base image

elevran:
Use distroless:static as the base image for EPP and sidecar. The distroless/static image includes CA certs and has minimal CVE surface.

The base can be easily replaced (if needed by downstream or for a particular reason) by passing (e.g.) `--build-arg BASE_IMAGE=registry.access.redhat.com/ubi9/ubi-micro:9.7` at runtime.
Note: if using `scratch`, CA certificates must be copied from the builder stage:
```dockerfile
COPY --from=go-builder /etc/ssl/certs/ca-bundle.crt /etc/ssl/certs/ca-bundle.crt
```

For ad-hoc debugging (e.g., filesystem inspection) use ephemeral debug containers:
```console
kubectl debug -it <pod-name> -n <namespace> \
    --image=busybox \
    --target=epp \
    -- sh
```
This attaches a throwaway container with a full shell, sharing the pod's PID/network/filesystem namespace, without baking anything into the image. The `--target` flag requires the pod to have shareProcessNamespace: true or be on a recent Kubernetes (1.25+) version.

### DIFF
--- a/Dockerfile.epp
+++ b/Dockerfile.epp
@@ -1,3 +1,10 @@
+# BASE_IMAGE can be overridden at build time, e.g.:
+#   --build-arg BASE_IMAGE=registry.access.redhat.com/ubi9/ubi-micro:9.7
+# Default is distroless/static which includes CA certs and has minimal CVE surface.
+# NOTE: if using scratch, CA certificates must be copied from the builder stage:
+#   COPY --from=go-builder /etc/ssl/certs/ca-bundle.crt /etc/ssl/certs/ca-bundle.crt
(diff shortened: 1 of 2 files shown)


shmuelk:
Please add a Makefile target to run the debug container. This can be a separate PR

elevran:
> Please add a Makefile target to run the debug container. This can be a separate PR

and just use whatever k8s context and config kubectl happens to point to?
I think it is better addressed as a comment in the DEVELOPMENT.md for debug and changing the base

PR #803, opened 2026-04-08
migrate: import pkg/epp from gateway-api-inference-extension as pkg/epp/igw

elevran:
Migrates `pkg/epp` from `sigs.k8s.io/gateway-api-inference-extension` into this repo under `pkg/epp/igw`, preserving full git history (556 commits).

## What's included
- All files under `pkg/epp/` from GAIE, landed at `pkg/epp/igw/`
- Import path rewrites throughout the destination repo: `sigs.k8s.io/gateway-api-inference-extension/pkg/epp/*` → `github.com/llm-d/llm-d-inference-scheduler/pkg/epp/igw/*`

## Notes
- Build is not clean: moved code imports other GAIE packages (`pkg/common/...`) not yet migrated
- **Merge via "Create a merge commit"** — squash or rebase will discard the migrated history

PR #804, opened 2026-04-08
Script to automate migration for GAIE code into llm-d

elevran:
Script support initial move and subsequent updates (via `--since` flag).
The script creates a PR to merge mgrated files from GAIE (can be easily modified to support other migrations such as bbr, latency sidecar, etc).
In addition to the source commits. the PR includes two commits:
- `migrate: import <GAIE paths> from sigs.k8s.io/gateway-api-inference-extension as <local-paths> at <SHA>`
- `chore: rewrite imports <from> -> <to>`

The SHA can be used to do incremental migrations using `--since` (can be either SHA ref or tag. Prefer adding clear GAIE tags for `llm-migration-<date>`). Ideally we should strictly limit PRs for code already migrated (e.g., bug fixes only, request that they are applied here as well, etc.).

Some caveats (see #803 for example output . Note that PR will be closed - only done as test):
1. need to investigate missing Signed-Off-By
1. need to merge without crypto verified signatures (currently required on llm-d-inference-scheduler)
1. fix typos or linting on migration

I think the first two can be temporarily relaxed for the merge window in our local settings. The last one should be applied.

Looking at the entire GAIE code base under `cmd`, `pkg` and `test`

| Path | Conflict? | Notes |                                                                                                                
|------|-----------|-------|                                                                                                              
| `cmd/epp/main.go` | **Yes** | Both repos have `main.go` at this exact path; a direct move would clobber the dest entrypoint. `cmd/epp/runner/` (4 files) exists only in source so safe. |                                                                                 
| `pkg/common/` | No | Source: `certs.go`, `envoy/`, `error/`, `observability/`, `request/`. Dest: `common.go`, `common_test.go`. Different filenames; co-exist safely, though they'd share the same Go package name. |                                                                 
| `test/e2e/` | No | Different filenames on each side; no clobbering. |
| `test/integration/` | No | Different filenames on each side; no clobbering. |
| `test/utils/` | No | Different filenames on each side; no clobbering. |

So there is only one true conflict: `cmd/epp/main.go`. Any migration of `cmd/epp` from source should be done manually.
Everything else under `pkg/common`, `test/e2e`, `test/integration`, and `test/utils` shares directory names but not file names, so they would merge without clobbering (though we'd want to verify the content is compatible and code from GAIE and llm-d co-exist under the same package name and import path if not renamed).

### DIFF
--- a/scripts/migrate-gaie-paths.sh
+++ b/scripts/migrate-gaie-paths.sh
@@ -145,19 +145,18 @@ for i in "${!SRC_PATHS[@]}"; do
   [[ "${SRC_PATHS[$i]}" != "${DEST_PATHS[$i]}" ]] && FILTER_ARGS+=(--path-rename "${SRC_PATHS[$i]}:${DEST_PATHS[$i]}")
 done
 
-# Clone source and filter to target paths
+# Clone source, filter to target paths, and rewrite bare #NNN issue references to
+# SOURCE_ORG/SOURCE_REPO_NAME#NNN so they link to the original repo rather than the
+# destination. Matches only #NNN preceded by start-of-line, space, '(' or ',' to
+# avoid hex literals, code-block references, and already-qualified org/repo#NNN refs.
+MSG_CALLBACK="import re
+return re.sub(rb'(?m)(^|[ (,])#(\\d+)', lambda m: m.group(1) + b'${SOURCE_ORG}/${SOURCE_REPO_NAME}#' + m.group(2), message)"
+
 rm -rf "${FILTER_WORK_DIR}"
 git clone "file://${SOURCE_DIR}" "${FILTER_WORK_DIR}"
-git -C "${FILTER_WORK_DIR}" filter-repo "${FILTER_ARGS[@]}" --force
-
-# Rewrite bare #NNN references to SOURCE_ORG/SOURCE_REPO_NAME#NNN in commit messages so
-# references point to original repo links, and not arbitrary destination references.
-# Matches only #NNN preceded by start-of-line, space, '(' or ','. This avoids
-# hex literals, code-block references, and already-qualified org/repo#NNN refs.
-FILTER_BRANCH_SQUELCH_WARNING=1 git -C "${FILTER_WORK_DIR}" \
-  filter-branch --msg-filter \
-  "sed -E 's/(^|[ (,])#([0-9]+)/\1${SOURCE_ORG}\/${SOURCE_REPO_NAME}#\2/g'" \
-  -- --all
+git -C "${FILTER_WORK_DIR}" filter-repo "${FILTER_ARGS[@]}" \
+  --message-callback "${MSG_CALLBACK}" \
+  --force
 
 if [[ -n "${SINCE_REF}" ]]; then
   # Use filter-repo's commit-map to translate SINCE_REF into the filtered history.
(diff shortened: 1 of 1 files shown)


elevran:
@hexfusion implemented your suggestion: commit message rewrite GH references qualified by org/repo. It does not resolve all of the aspects (e.g., git blame) but I think it does make sense to include it. WDYT?

hexfusion:
Right I think the main question is what is reasonably correct. I would argue in the case that code is moving ownership from one repo to another that a reference to the source of truth is valid. Git blame would reference the commit to the repo, the comments would provide details on the source of truth. From a simplicity and correctness perspective it feels appropriate. 

With regards to contribution credit. I as a contributor do not lose my contribution to GIE. But also I did not contribute directly to llm-d-scheduler so again this feels reasonable that the git blame does not reference me directly but instead references the migration PR from GIE.

Agents and humans can still understand original ownership either way. But is the complexity cost worth it. Open question for input.

scripts/migrate-gaie-paths.sh
```diff
@@ -0,0 +1,282 @@
...
+git -C "${FILTER_WORK_DIR}" filter-repo "${FILTER_ARGS[@]}" --force
+
+# Rewrite bare #NNN references to SOURCE_ORG/SOURCE_REPO_NAME#NNN in commit messages so
+# references point to original repo links, and not arbitrary destination references.
+# Matches only #NNN preceded by start-of-line, space, '(' or ','. This avoids
+# hex literals, code-block references, and already-qualified org/repo#NNN refs.
+FILTER_BRANCH_SQUELCH_WARNING=1 git -C "${FILTER_WORK_DIR}" \
+  filter-branch --msg-filter \
+  "sed -E 's/(^|[ (,])#([0-9]+)/\1${SOURCE_ORG}\/${SOURCE_REPO_NAME}#\2/g'" \
+  -- --all
```
shmuelk: It would be better to use `git filter-repo..` with `--message-callback`.

See https://htmlpreview.github.io/?https://github.com/newren/git-filter-repo/blob/docs/html/git-filter-repo.html#CALLBACKS

scripts/migrate-gaie-paths.sh
```diff
@@ -0,0 +1,282 @@
...
+    echo "no commits in source before ${SINCE_REF} touching the specified paths; nothing to sync"
+    exit 0
+  fi
+  COMMIT_MAP="${FILTER_WORK_DIR}/.git/filter-repo/commit-map"
+  SINCE_FILTERED=$(awk -v sha="${SINCE_SRC_SHA}" 'NR>1 && $1==sha {print $2; exit}' "${COMMIT_MAP}")
+  if [[ -z "${SINCE_FILTERED}" || "${SINCE_FILTERED}" == "0000000000000000000000000000000000000000" ]]; then
+    echo "error: could not map ${SINCE_SRC_SHA} to filter-work; the commit may not touch the specified paths"
+    exit 1
+  fi
+  readarray -t COMMITS < <(git -C "${FILTER_WORK_DIR}" rev-list --reverse "${SINCE_FILTERED}..HEAD" 2>/dev/null || true)
```
shmuelk: For future reference, this requires Bash 4.0 or above (MacOS comes with Bash 3.0)

elevran: thanks. will push a more portable variant

PR #811, opened 2026-04-11
Refactoring and unification of e2e tests and deployment yaml files

### LINKED ISSUE #798: Implement component-based deployment manifests to reduce YAML duplication
Since extending support for Encode disaggregation, we now have many different deployment YAML files, especially for E2E tests.

We should consider breaking down the single, monolithic deployment files into smaller YAMLs that include only specific components, such as Decode, Prefill, and Encode. Depending on the testing environment, we can then combine the required parts.

This approach will simplify the deployment process and make maintenance easier by eliminating a significant amount of duplicated directives.

roytman:
**Summary**

- Replaces per-scenario deployment files with 3 atomic Kustomize components (vllm-encode, vllm-prefill, vllm-decode) that compose to form any disaggregation scenario
- Eliminates test/e2e/yaml/ entirely - e2e tests now consume the same deploy/ overlays used by kind-dev-env.sh
- Separates simulator and real vLLM concerns via Kustomize component overlays
- Introduces DISAGG_E / DISAGG_P independent boolean flags, replacing the combined PD_ENABLED / EPD_ENABLED flags

**Motivation**
The project maintained two separate YAML locations (deploy/components/ and test/e2e/yaml/) with duplicated, drifting deployment definitions. Each disaggregation scenario (EPD, P/D, E/PD, E/P/D, DP) had its own complete file despite sharing identical Encode and Prefill definitions. Adding a new scenario required changes in multiple places.

**Changes**

- New atomic components (deploy/components/):
   - vllm-encode/ - Encoder deployment (multimodal, --mm-encoder-only)
   - vllm-prefill/ - Prefill deployment
   - vllm-decode/ - Decode deployment with routing sidecar (if required)

- Simulator/real vLLM overlays (deploy/components/overlays/):
   - simulator/ — Kustomize component that adds --mode=${VLLM_MODE}, UDS tokenizer sidecar, and KV cache args via JSON patches (avoids conflicts with scenario strategic merge patches)
   - real-vllm/ — Kustomize component that adds --ec-transfer-config and shared PVC for encoder embeddings, will be part of another PR

- Unified dev scenario overlays (deploy/environments/dev/):
   - epd/ - No disaggregation (default), no routing sidecar
   - p-d/ - Prefill/Decode
   - e-pd/ - Encode/Prefill-Decode
   - e-p-d/ - Encode/Prefill/Decode

   Each overlay selects which components to deploy
- Disaggregation flags - two independent booleans replacing the old combined flag:

   |**Flag**|**Default**|**Meaning**|
   |--------|-----------|-------------|
   |DISAGG_E|false|Deploy separate Encoder pods|
   | DISAGG_P|false|Deploy separate Prefill pods|

- Additional flags:
  |**Flag**|**Default**|**Meaning**|
  |--------|-----------|-------------|
  |VLLM_SIM_MODE|echo|Acts as a loopback. The response content mirrors the input, see [Simulator modes](https://github.com/llm-d/llm-d-inference-sim#modes-of-operation)|
  |VLLM_DATA_PARALLEL_SIZE|1| Number of data-parallel ranks per vLLM pod

**Breaking changes**

 - DISAGG_E / DISAGG_P replace PD_ENABLED / EPD_ENABLED (the legacy flags still work but print a deprecation warning)
- Deployment names changed: ${MODEL_NAME_SAFE}-vllm-sim-* → vllm-e, vllm-p, vllm-d
- EPP config renamed: sim-epd-epp-config.yaml → sim-e-p-d-epp-config.yaml

/close #798


roytman:
@elevran, @shmuelk, @mayabar, before going into detailed reviews, what do you think about this PR?

mayabar:
> @elevran, @shmuelk, @mayabar, before going into detailed reviews, what do you think about this PR?

I completely support the first three bullets.
I have some doubts about the last one.

I think keeping a single environment variable to define a combination of different disaggregated parts is more complicated than using a separate boolean flag for each disaggregation decision.

Comparison of options:

image

In addition, we could introduce a DP flag to define whether the D pod operates in DP mode.

What do you think?

roytman:
@mayabar , thank you for your comment, yes I replaced the `DISAGG_MODE` env variable with two flags as you suggested. 
For DP, we have another flag - `VLLM_DATA_PARALLEL_SIZE`, which is relevant for all Pods.

deploy/environments/dev/README.md
```diff
@@ -0,0 +1,68 @@
...
+| `e2e-infra/` | Test-specific infrastructure: NodePort services for health/metrics and envoy proxy config |
+| `kubernetes-kgateway/` | Alternative gateway configuration using Kubernetes Gateway API (instead of Istio) |
+
+## Scenario Selection
+
+Set `DISAGG_MODE` environment variable when running `kind-dev-env.sh`:
+
+```bash
+# Default (no disaggregation) - uses vllm-decode component directly
+DISAGG_MODE=epd ./scripts/kind-dev-env.sh
```
shmuelk: Aren't there now two separate flags?

scripts/kind-dev-env.sh
```diff
@@ -251,10 +280,10 @@ fi
 for IMAGE in "${VLLM_SIMULATOR_IMAGE}" "${EPP_IMAGE}" "${SIDECAR_IMAGE}" "${UDS_TOKENIZER_IMAGE}"; do
     if ! "${CONTAINER_RUNTIME}" image inspect "${IMAGE}" > /dev/null 2>&1; then
         echo "Image ${IMAGE} not found locally, pulling..."
-        "${CONTAINER_RUNTIME}" pull "${PLATFORM_ARGS[@]}" "${IMAGE}"
+        "${CONTAINER_RUNTIME}" pull ${PLATFORM_ARGS[@]+"${PLATFORM_ARGS[@]}"} "${IMAGE}"
```
shmuelk: ??

scripts/kind-dev-env.sh
```diff
@@ -251,10 +280,10 @@ fi
...
     fi
     echo "Loading ${IMAGE} into kind cluster..."
-    "${CONTAINER_RUNTIME}" save "${PLATFORM_ARGS[@]}" "${IMAGE}" | kind --name "${CLUSTER_NAME}" load image-archive /dev/stdin
+    "${CONTAINER_RUNTIME}" save ${PLATFORM_ARGS[@]+"${PLATFORM_ARGS[@]}"} "${IMAGE}" | kind --name "${CLUSTER_NAME}" load image-archive /dev/stdin
```
shmuelk: ??

scripts/kind-dev-env.sh
```diff
@@ -291,12 +322,21 @@ kubectl --context ${KUBE_CONTEXT} delete configmap epp-config --ignore-not-found
...
+# Deploy Istio base (shared infrastructure)
+kubectl kustomize --enable-helm deploy/environments/dev/base-kind-istio \
+  | envsubst '${POOL_NAME} ${MODEL_NAME} ${MODEL_NAME_SAFE} ${EPP_NAME} ${EPP_IMAGE} ${VLLM_IMAGE} ${VLLM_SIMULATOR_IMAGE} \
+  ${SIDECAR_IMAGE} ${UDS_TOKENIZER_IMAGE} ${TARGET_PORTS} ${POOL_NAMESPACE} ${METRICS_ENDPOINT_AUTH} \
   ${VLLM_REPLICA_COUNT} ${VLLM_REPLICA_COUNT_E} ${VLLM_REPLICA_COUNT_P} ${VLLM_REPLICA_COUNT_D} ${VLLM_DATA_PARALLEL_SIZE}' \
   | kubectl --context ${KUBE_CONTEXT} apply -f -
 
+# Deploy scenario-specific vLLM components
+kubectl kustomize --enable-helm ${KUSTOMIZE_DIR} \
+  | envsubst '${POOL_NAME} ${MODEL_NAME} ${MODEL_NAME_SAFE} ${EPP_NAME} ${EPP_IMAGE} ${VLLM_IMAGE} ${VLLM_SIMULATOR_IMAGE} \
```
shmuelk: I don't think the variable `VLLM_SIMULATOR_IMAGE` exists in any of the YAMLs any more.

scripts/kind-dev-env.sh
```diff
@@ -291,12 +322,21 @@ kubectl --context ${KUBE_CONTEXT} delete configmap epp-config --ignore-not-found
 envsubst '$MODEL_NAME' < ${EPP_CONFIG} > ${TEMP_FILE}
 kubectl --context ${KUBE_CONTEXT} create configmap epp-config --from-file=epp-config.yaml=${TEMP_FILE}
 
-kubectl kustomize --enable-helm  ${KUSTOMIZE_DIR} \
-	| envsubst '${POOL_NAME} ${MODEL_NAME} ${MODEL_NAME_SAFE} ${EPP_NAME} ${EPP_IMAGE} ${VLLM_SIMULATOR_IMAGE} \
-  ${PD_ENABLED} ${KV_CACHE_ENABLED} ${SIDECAR_IMAGE} ${UDS_TOKENIZER_IMAGE} ${TARGET_PORTS} \
+# Deploy Istio base (shared infrastructure)
+kubectl kustomize --enable-helm deploy/environments/dev/base-kind-istio \
+  | envsubst '${POOL_NAME} ${MODEL_NAME} ${MODEL_NAME_SAFE} ${EPP_NAME} ${EPP_IMAGE} ${VLLM_IMAGE} ${VLLM_SIMULATOR_IMAGE} \
```
shmuelk: I don't think the variable VLLM_SIMULATOR_IMAGE exists in any of the YAMLs any more.

scripts/kind-dev-env.sh
```diff
@@ -291,12 +322,21 @@ kubectl --context ${KUBE_CONTEXT} delete configmap epp-config --ignore-not-found
...
+  ${SIDECAR_IMAGE} ${UDS_TOKENIZER_IMAGE} ${TARGET_PORTS} ${POOL_NAMESPACE} ${METRICS_ENDPOINT_AUTH} \
   ${VLLM_REPLICA_COUNT} ${VLLM_REPLICA_COUNT_E} ${VLLM_REPLICA_COUNT_P} ${VLLM_REPLICA_COUNT_D} ${VLLM_DATA_PARALLEL_SIZE}' \
   | kubectl --context ${KUBE_CONTEXT} apply -f -
 
+# Deploy scenario-specific vLLM components
+kubectl kustomize --enable-helm ${KUSTOMIZE_DIR} \
+  | envsubst '${POOL_NAME} ${MODEL_NAME} ${MODEL_NAME_SAFE} ${EPP_NAME} ${EPP_IMAGE} ${VLLM_IMAGE} ${VLLM_SIMULATOR_IMAGE} \
+  ${SIDECAR_IMAGE} ${UDS_TOKENIZER_IMAGE} ${TARGET_PORTS} \
+  ${VLLM_REPLICA_COUNT} ${VLLM_REPLICA_COUNT_E} ${VLLM_REPLICA_COUNT_P} ${VLLM_REPLICA_COUNT_D} ${VLLM_DATA_PARALLEL_SIZE} \
```
shmuelk: I don't think the variable `VLLM_REPLICA_COUNT` exists in any of the YAMLs any more

DEVELOPMENT.md
```diff
@@ -395,7 +531,12 @@ kubectl --context kind-e2e-tests get pods
...
 | `READY_TIMEOUT` | `3m` | How long to wait for resources to become ready |
 | `EPP_IMAGE` | `ghcr.io/llm-d/llm-d-inference-scheduler:dev` | EPP image loaded into the Kind cluster |
-| `VLLM_SIMULATOR_IMAGE` | `ghcr.io/llm-d/llm-d-inference-sim:v0.8.2` | vLLM simulator image loaded into the Kind cluster |
+| `DISAGG_E` | `false` | Deploy a separate Encoder pod. See [Inference Disaggregation Modes](#inference-disaggregation-modes) |
+| `DISAGG_P` | `false` | Deploy a separate Prefill pod. See [Inference Disaggregation Modes](#inference-disaggregation-modes) |
+| `VLLM_DATA_PARALLEL_SIZE` | `1` | Number of data-parallel ranks per vLLM pod. Applies to all pod types. Set to `2`+ to enable multi-rank inference. See [Combining Scenarios with Data Parallel and KV Cache](#combining-scenarios-with-data-parallel-and-kv-cache) |
+| `VLLM_IMAGE` | `${VLLM_SIMULATOR_IMAGE}` | vLLM container image used in deployment templates. Can be a simulator or a real vLLM image (e.g., `vllm/vllm-openai:v0.16.0`) |
+| `VLLM_SIMULATOR_IMAGE` | `ghcr.io/llm-d/llm-d-inference-sim:v0.8.2` | Simulator image. Used as the default for `VLLM_IMAGE` and loaded into the Kind cluster |
```
shmuelk: Why both and not just `VLLM_IMAGE` ?

shmuelk (changes requested):
In addition, I think the environment variable defaults in the Makefile need to be reviewed.

deploy/environments/dev/README.md
roytman: Fixed, thank you.

scripts/kind-dev-env.sh
roytman: This is a Parameter Expansion trick. It is used to handle an array (PLATFORM_ARGS) that might be empty. In some versions of Bash, simply writing "${PLATFORM_ARGS[@]}" when the array was empty could cause an error or pass an "empty string" argument to the command.
So here is:

- `PLATFORM_ARGS` is an array defined earlier — it's empty (()) for Podman, or `("--platform" "linux/${LINUX_ARCH}")` for docker
- The ${array[@]+"${array[@]}"} pattern expands to nothing when the array is empty (avoids the "unbound variable" error with set -u), or expands to all array elements when non-empty
- So for Podman: the pull command becomes just `podman pull "${IMAGE}"`
- For Docker on Apple Silicon: `docker pull --platform linux/arm64 "${IMAGE}"` or  `docker pull --platform linux/amd64 "${IMAGE}"` for Intel/AMD

scripts/kind-dev-env.sh
roytman: see above

scripts/kind-dev-env.sh
roytman: it is transferred to VLLM_IMAGWE
scripts/kind-dev-env.sh:34:export VLLM_IMAGE="${VLLM_IMAGE:-${VLLM_SIMULATOR_IMAGE}}"

I left it for backward compatibility, but I can remove it if you think that we don't need it.

scripts/kind-dev-env.sh
roytman: I removed VLLM_SIMULATOR_IMAGE.

scripts/kind-dev-env.sh
roytman: removed it, thank you

DEVELOPMENT.md
roytman: I left `VLLM_SIMULATOR_IMAGE` for some kind of backward compatibility, but I can remove it if you think we don't need it.

deploy/environments/dev/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+| Variable | Description | Example |
+|----------|-------------|---------|
+| `DISAGG_E` | Deploy a separate Encoder pod (`true`/`false`) | `false` |
+| `DISAGG_P` | Deploy a separate Prefill pod (`true`/`false`) | `false` |
+| `VLLM_SIM_MODE` | Simulator response mode: `echo` (returns input) or `random` (random sentences) | `echo` |
+| `VLLM_IMAGE` | vLLM container image (simulator or real) | `ghcr.io/llm-d/llm-d-inference-sim:v0.8.2` |
+| `SIDECAR_IMAGE` | Routing sidecar image | `ghcr.io/llm-d/llm-d-routing-sidecar:dev` |
+| `UDS_TOKENIZER_IMAGE` | UDS tokenizer sidecar image | `ghcr.io/llm-d/llm-d-uds-tokenizer:dev` |
+| `MODEL_NAME` | HuggingFace model name | `food-review` |
```
mayabar: maybe explain here that model can be real HF model (provide example) or "fake" name like "food-review"
+ default value of the model_name in the kind deployment shell script is Qwen/Qwen3-VL-2B-Instruct or TinyLlama/TinyLlama-1.1B-Chat-v1.0

scripts/kind-dev-env.sh
roytman: removed it

DEVELOPMENT.md
roytman: Now we have only `VLLM_IMAGE`

roytman:
> In addition, I think the environment variable defaults in the Makefile need to be reviewed.

I removed some redundant environment variable settings. Plus, @elevran and I discussed to review/re-implementing all Makefiles as a separate effort.

scripts/kind-dev-env.sh
```diff
@@ -294,10 +352,21 @@ kubectl --context ${KUBE_CONTEXT} delete configmap epp-config --ignore-not-found
...
+# Deploy Istio base (shared infrastructure)
+kubectl kustomize --enable-helm deploy/environments/dev/base-kind-istio \
+  | envsubst '${POOL_NAME} ${MODEL_NAME} ${MODEL_NAME_SAFE} ${EPP_NAME} ${EPP_IMAGE} ${VLLM_IMAGE} ${VLLM_IMAGE} \
+  ${SIDECAR_IMAGE} ${UDS_TOKENIZER_IMAGE} ${TARGET_PORTS} ${NAMESPACE} ${METRICS_ENDPOINT_AUTH} \
+${VLLM_REPLICA_COUNT_E} ${VLLM_REPLICA_COUNT_P} ${VLLM_REPLICA_COUNT_D} ${VLLM_DATA_PARALLEL_SIZE}' \
+  | kubectl --context ${KUBE_CONTEXT} apply -f -
+
+# Deploy scenario-specific vLLM components
+kubectl kustomize --enable-helm ${KUSTOMIZE_DIR} \
+  | envsubst '${POOL_NAME} ${MODEL_NAME} ${MODEL_NAME_SAFE} ${EPP_NAME} ${EPP_IMAGE} ${VLLM_IMAGE} ${VLLM_IMAGE} \
```
shmuelk: ${VLLM_IMAGE} twice...

```suggestion
  | envsubst '${POOL_NAME} ${MODEL_NAME} ${MODEL_NAME_SAFE} ${EPP_NAME} ${EPP_IMAGE} ${VLLM_IMAGE} \
```

test/e2e/e2e_suite_test.go
```diff
@@ -192,6 +197,27 @@ func setupK8sCluster() {
...
+			out, err := exec.Command("uname", "-m").Output()
+			if err != nil {
+				return "amd64"
+			}
+			m := strings.TrimSpace(string(out))
+			if m == "aarch64" || m == "arm64" {
+				return "arm64"
+			}
+			return "amd64"
+		}()
```
shmuelk: Why not use `runtime.GOARCH` ?

test/e2e/setup_test.go
```diff
@@ -15,53 +15,66 @@ import (
...
-func createModelServersFromYaml(yaml string, extra map[string]string) []string {
+func createModelServersFromKustomize(kustomizeDir string, extra map[string]string) []string {
 	subs := map[string]string{
-		"${MODEL_NAME}":           simModelName,
-		"${MODEL_NAME_SAFE}":      simModelName,
-		"${POOL_NAME}":            poolName,
-		"${VLLM_SIMULATOR_IMAGE}": vllmSimImage,
-		"${UDS_TOKENIZER_IMAGE}":  udsTokenizerImage,
+		"${MODEL_NAME}":              simModelName,
+		"${MODEL_NAME_SAFE}":         simModelName,
```
shmuelk: Is ${MODEL_SAFE_NAME} actually used anymore?

shmuelk (changes requested):

shmuelk (approved):

PR #812, opened 2026-04-12
test: add fuzz test for parseContextRange

zzdpk2:
## What
Add fuzz test for `parseContextRange` in the context-length-aware scheduling plugin.

## Why
`parseContextRange` processes pod label values to determine context length routing.
While unit tests cover 5 cases, fuzz testing validates robustness against
arbitrary/malformed inputs at scale (500K+ executions, zero panics, zero invariant violations).

Fuzz test verifies:
- No panics on any input
- Parsed results maintain min <= max invariant
- Non-negative context length values

## Testing
- `go test -fuzz=FuzzParseContextRange -fuzztime=60s ./pkg/plugins/multi/` — PASS (533K executions)
- All existing tests pass: `go test ./pkg/plugins/multi/` — PASS

### DIFF
--- a/pkg/plugins/multi/context_length_aware_fuzz_test.go
+++ b/pkg/plugins/multi/context_length_aware_fuzz_test.go
@@ -1,0 +1,48 @@
+package multi
+
+import (
+	"testing"
+)
+
+func FuzzParseContextRange(f *testing.F) {
+	// Seed corpus: valid formats
+	f.Add("0-2048")
+	f.Add("2048-8192")
+	f.Add("0-0")
+	f.Add("100-100")
+
+	// Seed corpus: edge cases and invalid inputs
+	f.Add("")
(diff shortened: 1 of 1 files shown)


elevran:
thanks @zzdpk2!
Did you encounter a specific issue or input failures for the parsing code, or is this more for being proactive and adding it to the test matrix?

Fuzz tests can be important, and we need to consider them from an overall project perspective. Due to the time needed in fuzz, it probably makes more sense as regression testing done for a release and not as unit test (e.g., consider when the nuber of fuzzed functions grow).

Would you be interested in making this (regression testing) type of contribution? If so, please open an issue to discuss
- when and how fuzz tests are to be run
- how the makefile and CICD would change to support that 
- an initial prioritization of functions that shuold be fuzzed (e.g., based on risk and criticallity).

PR #822, opened 2026-04-14
feat(sidecar): add chunked decode

andreyod:
Adds experimental chunked decode to the pd-sidecar proxy. When enabled,
the decode stage is split into a sequence of shorter decode calls, each
capped at a configurable token budget (`--decode-chunk-size`). After each
chunk, the generated text is appended to the conversation context so the
next chunk continues seamlessly.

The feature applies at the decode stage regardless of whether P/D
disaggregation is in use.

## Test and support plan
- Tested e2e for decode-only setups with vLLM and SGLang inference servers.
- Tested e2e for PD disaggregation setup with vLLM (kv-connector=nixlv2).
- Not supported for PD disaggregation setup with SGLang (kv-connector=sglang) due to a known SGLang limitation that may be addressed in the future.

```release-note
Introduce chunked decode via a sidecar CLI flag, splitting long token generation into multiple bounded iterations.
To enable add `--decode-chunk-size=<size>`. The chunk size should relatively large (512 or more) and be a multiple of the block size.
```

## How to enable
```yaml
--decode-chunk-size=512   # should be a multiple of the block size
```

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -227,7 +227,7 @@ chunk continues seamlessly from where the previous one left off.
 
 ### How it works
 
-1. The sidecar receives a `/v1/chat/completions` or `/v1/completions` request at the decode stage.
+1. The sidecar receives a `/v1/chat/completions` request at the decode stage.
 2. Each chunk is dispatched as a separate request to the local decoder with `max_tokens` capped
    at `decode-chunk-size`.
 3. From the second chunk onward, `continue_final_message=true` and `add_generation_prompt=false` are
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -243,15 +243,14 @@ chunk continues seamlessly from where the previous one left off.
 
 ### Configuration
 
-Enable chunked decode via the pd-sidecar flags:
+Enable chunked decode via the pd-sidecar flag:
 
 | Flag | Default | Description |
 |---|---|---|
-| `--enable-chunked-decode` | `false` | Enable chunked decode mode |
-| `--decode-chunk-size` | `512` | Chunk size in tokens; for best performance use a multiple of the KV cache block size |
+| `--decode-chunk-size` | `0` (disabled) | Token budget per chunk. Set to a positive integer to enable chunked decode. For best performance use a multiple of the KV cache block size. |
 
 > [!NOTE]
-> If the request's `max_tokens` / `max_completion_tokens` is less than or equal to `decode-chunk-size`,
+> If the request's `max_tokens` / `max_completion_tokens` is less than or equal to `--decode-chunk-size`,
 > the sidecar falls back to a single regular decode call without chunking.
 
 ---
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -211,6 +211,51 @@ The **vLLM sidecar** handles orchestration between Encode, Prefill and Decode st
 
 ---
 
+## Chunked Decode (Experimental)
+
+Chunked decode is an experimental feature of the pd-sidecar that splits the decode stage into a
+sequence of shorter decode calls, each capped at a configurable token budget.
+It applies at the decode stage regardless of whether P/D disaggregation is in use.
+After each chunk the generated text is appended to the conversation context so the next
+chunk continues seamlessly from where the previous one left off.
+
+### Why to use it
+
+- Improve average Time-To-First-Token among all requests
+- Prevent head-of-line blocking by long requests in run-to-completion
+- Get more predictable execution time
+
+### How it works
+
+1. The sidecar receives a `/v1/chat/completions` or `/v1/completions` request at the decode stage.
+2. Each chunk is dispatched as a separate request to the local decoder with `max_tokens` capped
+   at `decode-chunk-size`.
+3. From the second chunk onward, `continue_final_message=true` and `add_generation_prompt=false` are
+   set so the model continues the existing assistant turn rather than starting a new one. The
+   generated text from the previous chunk is also appended to the request context.
+4. Generation stops when the model returns a terminal `finish_reason` (anything other than `length`),
+   or when the original token budget is exhausted.
+5. For **non-streaming** requests, all chunk outputs are concatenated and returned as a single
+   response. The `usage` field reports the original `prompt_tokens` (from the first chunk) and the
+   total `completion_tokens` across all chunks.
+6. For **streaming** requests, each chunk's tokens are re-emitted as SSE delta events in real time,
+   and a `[DONE]` sentinel closes the stream once all chunks are complete.
+
+### Configuration
+
+Enable chunked decode via the pd-sidecar flags:
+
+| Flag | Default | Description |
+|---|---|---|
+| `--enable-chunked-decode` | `false` | Enable chunked decode mode |
+| `--decode-chunk-size` | `512` | Chunk size in tokens; for best performance use a multiple of the KV cache block size |
+
+> [!NOTE]
+> If the request's `max_tokens` / `max_completion_tokens` is less than or equal to `decode-chunk-size`,
+> the sidecar falls back to a single regular decode call without chunking.
+
+---
+
 ## InferencePool & InferenceModel Design
 
 ### Current Assumptions
--- a/pkg/sidecar/proxy/decode.go
+++ b/pkg/sidecar/proxy/decode.go
@@ -39,22 +39,15 @@ const (
 	sseDataPrefix = "data: "
 	sseDone       = "data: [DONE]"
 
-	// legacyChunkPrefix is prepended to each chunk's generated text when
-	// appending to a legacy completions prompt for continuation.
-	legacyChunkPrefix = " Assistant: "
-
 	responseFieldUsage            = "usage"
 	responseFieldCompletionTokens = "completion_tokens"
 	responseFieldPromptTokens     = "prompt_tokens"
 	responseFieldTotalTokens      = "total_tokens"
 	responseFieldMessage          = "message"
-	responseFieldText             = "text"
-	responseFieldLogprobs         = "logprobs"
 	responseFieldIndex            = "index"
 	responseFieldDelta            = "delta"
 
 	requestFieldMessages = "messages"
-	requestFieldPrompt   = "prompt"
 	requestFieldRole     = "role"
 	requestFieldContent  = "content"
 
@@ -62,11 +55,12 @@ const (
 )
 
 // dispatchDecode routes a fully-prepared decode request to either chunked
-// decode or the regular decoder proxy. Chunked decode is used when
-// s.config.DecodeChunkSize > 0. completionRequest is the already-parsed JSON
-// map; callers that hold it should use this instead of calling s.decoderProxy directly.
+// decode or the regular decoder proxy. Chunked decode is only used for
+// chat completions requests when s.config.DecodeChunkSize > 0.
+// completionRequest is the already-parsed JSON map; callers that hold it
+// should use this instead of calling s.decoderProxy directly.
 func (s *Server) dispatchDecode(w http.ResponseWriter, r *http.Request, completionRequest map[string]any) {
-	if s.config.DecodeChunkSize > 0 {
+	if s.config.DecodeChunkSize > 0 && r.URL.Path == ChatCompletionsPath {
 		s.runChunkedDecodeFromMap(w, r, completionRequest)
 		return
 	}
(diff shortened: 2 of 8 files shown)


shmuelk:
@andreyod Why do you need both an enable flag and chunk size parameters? Is a default chunk size useful? If not, why not simply set the default to zero or minus one, eliminate the enable flag and enable automatically if the chunk size is greater than zero?

andreyod:
> @andreyod Why do you need both an enable flag and chunk size parameters? Is a default chunk size useful? If not, why not simply set the default to zero or minus one, eliminate the enable flag and enable automatically if the chunk size is greater than zero?

Right. I will make the change to use a single flag that defaults to 0 (off).

docs/architecture.md
```diff
@@ -782,6 +782,50 @@ The detailed E/P/D design is available in this document:
...
 
+## Chunked Decode (Experimental)
+
+Chunked decode is an experimental feature of the pd-sidecar that splits the decode stage into a
+sequence of shorter decode calls, each capped at a configurable token budget.
+It applies at the decode stage regardless of whether P/D disaggregation is in use.
+After each chunk the generated text is appended to the conversation context so the next
+chunk continues seamlessly from where the previous one left off.
+
+### Why to use it
```
hexfusion, line 222: thanks for the contribution!

still building my mental model of this, just have a few questions
 
- i thought vLLM's continuous batching already runs decode steps for all active requests each iteration, so wouldn't a long decode already not block other requests' TTFT? curious what engine mode or workload this is targeting.

- do you have any benchmark numbers for the claimed benefits? especially curious about per-chunk overhead I'd assume each chunk adds a round-trip plus some work on the engine side.

andreyod: @hexfusion good questions.
- vLLM's continuous batching focuses on request scheduling and prefill chunking. However, each decode request still runs to completion, meaning a series of long decode requests can still block other requests.
- We tested exactly what you are describing and found that the overall completion time for an array of requests does not change. This indicates that the "context switch" is very fast and the overhead is minimal.

We presented this feature in the SIG meeting on Feb 11; you may find more answers by watching the [[recording]](https://drive.google.com/file/d/1HKAjrXs4A2odZ_0sIHdYVijpaRYzjeK3/view)

pkg/sidecar/proxy/decode_test.go
```diff
@@ -0,0 +1,349 @@
...
+	ctx := newTestContext()
+	ctx, cancelFn := context.WithCancel(ctx)
+	stoppedCh := make(chan struct{})
+
+	go func() {
+		defer GinkgoRecover()
+		_ = proxy.Start(ctx)
+		stoppedCh <- struct{}{}
+	}()
+	time.Sleep(200 * time.Millisecond)
```
elevran: Use `<-proxy.readyCh` to wait for proxy start, which is deterministic and eliminates the 200 ms overhead per test.
Same issue at lines 223 and 261.

pkg/sidecar/proxy/decode_test.go
```diff
@@ -0,0 +1,349 @@
...
+
+		It("appends assistant message to messages on second chunk", func() {
+			var secondReqBody map[string]any
+			var reqIdx int
+			responses := []string{
+				chatResponse("hello ", "length", 5, 5),
+				chatResponse("world", "stop", 9, 5),
+			}
+
+			backend := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
elevran: This test inlines 14 lines of proxy setup (backend server, config, proxy, context, goroutine) that `newChunkedTestSetup` already provides. The only thing specific to this test is the backend handler that captures the second request body.

Consider extending `newChunkedTestSetup` to accept an optional `http.Handler` override, or pass the capture logic as a wrapper around the standard `decodeResponses`-based handler. Same duplication exists in the `propagates decode backend error` test at line 242.

pkg/sidecar/proxy/decode.go
```diff
@@ -0,0 +1,461 @@
...
+			if !ok {
+				continue
+			}
+			text := extractChoiceText(choice)
+			streamChoice := map[string]any{
+				responseFieldIndex:        choice[responseFieldIndex],
+				responseFieldFinishReason: choice[responseFieldFinishReason],
+			}
+			if _, isChatCompletion := choice[responseFieldMessage]; isChatCompletion {
+				streamChoice[responseFieldDelta] = map[string]any{requestFieldContent: text, requestFieldRole: "assistant"}
```
elevran: The role string `"assistant"` is used as a raw literal here and again at line 438 (in `appendChunkToRequest`). A `roleAssistant = "assistant"` constant alongside `finishReasonLength` would make both sites consistent and safer in refactors.

pkg/sidecar/proxy/decode.go
```diff
@@ -0,0 +1,461 @@
...
+	if lastResponse == nil {
+		if err := errorInternalServerError(errors.New("no chunks produced"), w); err != nil {
+			s.logger.Error(err, "failed to send error response to client")
+		}
+		return
+	}
+
+	// Fix usage: prompt_tokens from first chunk, completion_tokens summed across all chunks.
+	if usage, ok := lastResponse[responseFieldUsage].(map[string]any); ok {
+		usage = maps.Clone(usage)
```
elevran: `maps.Clone(usage)` is unnecessary here. `lastResponse` is a local variable that is JSON-encoded immediately after and not referenced again. The `usage` sub-map can be modified in place:

```go
if usage, ok := lastResponse[responseFieldUsage].(map[string]any); ok {
    usage[responseFieldPromptTokens] = originalPromptTokens
    usage[responseFieldCompletionTokens] = totalTokens
    usage[responseFieldTotalTokens] = originalPromptTokens + totalTokens
    // no Clone or reassignment needed
}
```

pkg/sidecar/proxy/options.go
```diff
@@ -111,6 +111,7 @@ func NewOptions() *Options {
 			PoolGroup:               DefaultPoolGroup,
 			InferencePoolNamespace:  os.Getenv("INFERENCE_POOL_NAMESPACE"),
 			InferencePoolName:       os.Getenv("INFERENCE_POOL_NAME"),
+			DecodeChunkSize:         0, // 0 disables chunked decode
```
elevran: nit: comment `// 0 disables chunked decode` restates what the flag description already documents and can be safely removed.

elevran (commented):
thanks @andreyod 
Some comments below, otherwise please resolve conflicts and ensure `make presubmit` is clean.

elevran:
#683 modifies the options/command line. Please ensure the additional CLI flag introduced in this PR is correctly reflected.

pkg/sidecar/proxy/decode_test.go
andreyod: done

pkg/sidecar/proxy/decode_test.go
andreyod: done

pkg/sidecar/proxy/decode.go
andreyod: done

pkg/sidecar/proxy/decode.go
andreyod: done

pkg/sidecar/proxy/options.go
andreyod: removed

pkg/sidecar/proxy/decode.go
```diff
@@ -0,0 +1,461 @@
...
+			} else {
+				streamChoice[responseFieldText] = text
+				streamChoice[responseFieldLogprobs] = choice[responseFieldLogprobs]
+			}
+			streamChoices = append(streamChoices, streamChoice)
+		}
+		streamChunk[responseFieldChoices] = streamChoices
+	}
+
+	// Omit usage on intermediate chunks (finish_reason == length or absent).
```
elevran: This correctly strips usage from intermediate chunks, but the final chunk (terminal `finish_reason`) still carries raw usage from the last backend response: `prompt_tokens` reflects the extended context (inflated) and `completion_tokens` is only the last chunk's output (not the cumulative total). Streaming clients that track token spend will get wrong numbers.

The non-streaming path already has the right fix at L271–275. Consider injecting a corrected usage event here for the final chunk, using same arithmetic, just wrapped in an SSE `data:` event, before emitting `[DONE]`.

pkg/sidecar/proxy/decode.go
```diff
@@ -0,0 +1,461 @@
...
+		totalTokens          int
+		chunkIndex           int
+		lastResponse         map[string]any
+		originalPromptTokens int
+		textAccum            strings.Builder
+	)
+
+	decodeStart := time.Now()
+
+	for {
```
elevran, line 136: No `ctx.Err()` check at the top of the loop. If the client disconnects between two backend calls, the loop keeps running until the next `ServeHTTP` call propagates the error. For a large token budget this can mean several unnecessary backend requests. A `if err := ctx.Err(); err != nil { return }` at the start of each iteration would be enough.

pkg/sidecar/proxy/decode.go
```diff
@@ -0,0 +1,461 @@
...
+const (
+	// finishReasonLength is the finish reason when max_tokens was reached.
+	finishReasonLength = "length"
+
+	sseDataPrefix = "data: "
+	sseDone       = "data: [DONE]"
+
+	// legacyChunkPrefix is prepended to each chunk's generated text when
+	// appending to a legacy completions prompt for continuation.
+	legacyChunkPrefix = " Assistant: "
```
elevran: `" Assistant: "` hardcodes an instruction-tuned chat format. General legacy completions (code completion, fill-in-the-middle) won't use that delimiter, so the stitched prompt will be malformed. Worth either restricting chunked decode to the chat completions path, or at minimum adding a prominent caveat in the docs/flag description.

pkg/sidecar/proxy/decode.go
andreyod: done as part of [dfc0ac4](https://github.com/llm-d/llm-d-inference-scheduler/pull/822/commits/dfc0ac43f7986400b3c89956b2e91783cb067edb)

pkg/sidecar/proxy/decode.go
andreyod: done as part of [dfc0ac4](https://github.com/llm-d/llm-d-inference-scheduler/pull/822/commits/dfc0ac43f7986400b3c89956b2e91783cb067edb)

pkg/sidecar/proxy/decode.go
andreyod: Restricted chunked decode to the chat completions path as it's also problematic for other models at small chunk size. Done as part of [dfc0ac4](https://github.com/llm-d/llm-d-inference-scheduler/pull/822/commits/dfc0ac43f7986400b3c89956b2e91783cb067edb)

pkg/sidecar/proxy/decode.go
```diff
@@ -0,0 +1,458 @@
...
+				responseFieldFinishReason: choice[responseFieldFinishReason],
+			}
+			streamChoice[responseFieldDelta] = map[string]any{requestFieldContent: text, requestFieldRole: roleAssistant}
+			streamChoices = append(streamChoices, streamChoice)
+		}
+		streamChunk[responseFieldChoices] = streamChoices
+	}
+
+	// Omit usage on intermediate chunks (finish_reason == length or absent).
+	if fr := extractFinishReason(chunkResponse); fr == "" || fr == finishReasonLength {
```
elevran: In a multi-chunk streaming scenario the terminal chunk (i.e., finish_reason=`stop`) keeps its backend usage here, but that usage only reflects the last chunk's `completion_tokens`, not the cumulative total. 
Suggestion to always strip usage in `emitSSEChunk` and unconditionally emit the separate usage event (dropping the `chunkIndex > 1` guard) in L266.

elevran (commented):
/lgtm

pkg/sidecar/proxy/decode.go
andreyod: changed in [dfa564f](https://github.com/llm-d/llm-d-inference-scheduler/pull/822/commits/dfa564f5f68f4d8ed08cc82b60b98e6d17ffebd0)

elevran:
@andreyod #683 merged. Can you please resolve conflicts and add the chunked decode config accordingly (Options, CLI, etc.)

andreyod:
> @andreyod #683 merged. Can you please resolve conflicts and add the chunked decode config accordingly (Options, CLI, etc.)

Thanks. Done in [7b8cde7](https://github.com/llm-d/llm-d-router/pull/822/commits/7b8cde7430a9dbf5bd8d34b149f06fcf1dfb5ddc)

PR #827, opened 2026-04-14
Refactoring plugins according to GAIE structure

roytman:
Part of https://github.com/llm-d/llm-d-inference-scheduler/issues/810

- README files will be added as separate PRs.
- Renaming plugin types requires updating other repositories, and probably requires the deprecation process. It will check separately.


ahg-g:
Thanks @roytman ; this doesn't match the structure we have in https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/pkg/epp/framework/plugins

Note that for each subcomponent (datalayer, requestcontrol, requesthandling, flowcontrol and scheduling), we also have a subdirectory for the the exact functionality (like scorer, filter, dataproducer etc.) and then under it we create a pkg per plugin.

roytman:
> Thanks @roytman ; this doesn't match the structure we have in https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/pkg/epp/framework/plugins
> 
Thank you @ahg-g. I hope it now follows the GAIE directory structure.

pkg/epp/framework/plugins/datalayer/dataproducer/tokenizer/tokenizer.go
```diff
@@ -14,8 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package preparedata provides PrepareData plugins for the scheduler.
-package preparedata
+// Package tokenizer provides PrepareData plugins for the scheduler.
```
ahg-g: the sub pkg is called `dataproducer`, not `preparedata`, the plugin name changed to `DataProducer`

pkg/epp/framework/plugins/scheduling/disagg_profile/always_disagg_mm_decider.go
```diff
@@ -1,4 +1,4 @@
-package profile
+package disaggprofile
```
ahg-g: The sub pkg is called [profile](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/pkg/epp/framework/plugins/scheduling/profile); I think we should call it `profilehandler` and under it we add `disagg` (and `single` after the upstream merges with this repo)

pkg/epp/framework/plugins/scheduling/disagg_profile/dp_profile_handler.go
```diff
@@ -1,4 +1,4 @@
-package profile
+package disaggprofile
```
ahg-g: This is not a disagg profile, it is a dp (data parallel) one, pls place under `profilehandler/dataparallel`

pkg/epp/framework/plugins/scheduling/scorer/multi/context_length_aware.go
```diff
@@ -13,7 +13,7 @@ import (
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/plugin"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/scheduling"
 
-	"github.com/llm-d/llm-d-inference-scheduler/pkg/plugins/preparedata"
+	tokenizer "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/datalayer/preparedata/tokenizer"
```
ahg-g: pls rename multi to something else that matches the name of the scorer

ahg-g (commented):
Thanks!

The convention for pkg names is to not have underscore, so `activerequests`, `bylabel`, `loadaware`, `nothitlru` and `preciseprefix`

pkg/epp/framework/plugins/datalayer/dataproducer/tokenizer/tokenizer.go
roytman: I'm not sure that DataProducer is the correct name for the plugin. It is a tokenizer plugin and probably one of the DataProducers. 
I'm slightly confused by the content of the `dataproducer` dir, it contains go files and a subdir with another plugin.

ahg-g: I meant the name of the pluggable interface changed to DataProducer, and so the the subpkg name should follow that.

ahg-g: The `dataproducer` pkg is under `requestcontrol` not `datalayer`, I know it is a bit confusing, but datalayer is about async data management, while requestcontrol/dataproducer is about per request data processing.

so pls move `datalayer/dataproducer/tokenizer` to `requestcontrol/dataproducer/tokenizer`

ahg-g (commented):
I can't comment on the file, but why is [disagg_headers_handler.go](https://github.com/llm-d/llm-d-inference-scheduler/pull/827/changes#diff-006bd31a3ca45c4a67a7d9c9e4233870993c0c64f64b6605697a95ced3c18ae0) a separate plugin? this should be tied to the disagg profile handler (just like some dataproducer plugins use PreRequest extension point in some cases)

roytman:
> I can't comment on the file, but why is [disagg_headers_handler.go](https://github.com/llm-d/llm-d-inference-scheduler/pull/827/changes#diff-006bd31a3ca45c4a67a7d9c9e4233870993c0c64f64b6605697a95ced3c18ae0) a separate plugin? this should be tied to the disagg profile handler (just like some dataproducer plugins use PreRequest extension point in some cases)

I like the idea of combining the handlers, as it will simplify the YAML files. I am not entirely sure why they were originally separated, though it was likely to enforce stricter flow separation.

My focus in this PR is on refactoring the plugins; therefore, I have moved the files to the `pkg/epp/framework/plugins/scheduling/profilehandler/disagg/` directory for now. 
I will explore combining the handlers in a subsequent PR (#833), as this is an end-user-facing feature, it may require a formal deprecation process.

pkg/epp/framework/plugins/register.go
```diff
@@ -0,0 +1,49 @@
...
+	activerequest "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/scheduling/scorer/activerequest"
+	contextlengthaware "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/scheduling/scorer/contextlengthaware"
+	loadaware "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/scheduling/scorer/loadaware"
+	nohitlru "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/scheduling/scorer/nohitlru"
+	preciseprefixcache "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache"
+	sessionaffinity "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity"
+)
+
+// RegisterAllPlugins registers the factory functions of all plugins in this repository.
+func RegisterAllPlugins() {
```
elevran, line 21: nit: full name includes package `plugins` 
```suggestion
func RegisterAll() {
```

pkg/epp/framework/plugins/scheduling/filter/bylabel/selector_test.go
```diff
@@ -12,7 +12,7 @@ import (
 	fwkdl "sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/datalayer"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/scheduling"
 
-	"github.com/llm-d/llm-d-inference-scheduler/pkg/plugins/filter"
+	by_label_filter "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/scheduling/filter/bylabel"
```
elevran: Q: rationale for having the package alias instead of using `bylabel` directly?

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_headers_handler.go
```diff
@@ -1,5 +1,5 @@
-// Package prerequest provides pre-request plugins for GIE.
-package prerequest
+// Package disagg provides pre-request plugins for GIE.
```
elevran, line 1: 1. Q: is it sufficient to have just one file with package level comment or must it be reapeated in all?
2. I _think_ it might be best to have a `doc.go` serving as package level description and see if we can use automation to generate the README from it (assuming we want a README in the repo - end users should refer to documentation, developers can use the doc.go and we should have a way to generate docs from so it is not duplicated).

pkg/scheduling/pd/scheduler_test.go
```diff
@@ -19,9 +19,9 @@ import (
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/plugins/scheduling/scorer/prefix"
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/scheduling"
 
-	"github.com/llm-d/llm-d-inference-scheduler/pkg/plugins/filter"
-	"github.com/llm-d/llm-d-inference-scheduler/pkg/plugins/profile"
-	"github.com/llm-d/llm-d-inference-scheduler/pkg/plugins/scorer"
+	by_label_filter "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/scheduling/filter/bylabel"
```
elevran: Q: is there an issue with `bylabel`?

elevran (commented):
/lgtm

pkg/epp/framework/plugins/register.go
```diff
@@ -0,0 +1,49 @@
+// Package plugins registers all llm-d inference scheduler plugins.
+package plugins
+
+import (
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/plugin"
+
+	tokenizer "github.com/llm-d/llm-d-inference-scheduler/pkg/app/framework/plugins/requestcontrol/dataproducer/tokenizer"
```
elevran: no need to rename these imports, especially not to the same name as the package.

pkg/epp/framework/plugins/register.go
roytman: I think `RegisterAllPlugins` is clearer.

pkg/epp/framework/plugins/scheduling/filter/bylabel/selector_test.go
roytman: I removed the alias, but there is another question. Why are some of the unit tests in a separate package, like this one (`bylabel_test`), but others are in the plugin package e.g. `filter_test.go` from the same directory? 
(Maybe a reason for a new issue)

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_headers_handler.go
roytman: Good point, will check it separately

pkg/epp/framework/plugins/scheduling/filter/bylabel/selector_test.go
elevran: they serve (or should serve) two different purposes. The `_test` package is supposed to test the exported API only and since it does not have access to the unexported implementation it can't test that and test have to be in the same package. 
In general, both are needed, not sure if this s the case here as well.

pkg/epp/framework/plugins/register.go
roytman: I removed all of them

elevran:
/lgtm
/approve
@roytman please resolve conflct

pkg/epp/framework/plugins/register.go
```diff
@@ -4,34 +4,34 @@ package plugins
 import (
 	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/plugin"
 
-	tokenizer "github.com/llm-d/llm-d-inference-scheduler/pkg/app/framework/plugins/requestcontrol/dataproducer/tokenizer"
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/app/framework/plugins/requestcontrol/dataproducer/tokenizer"
```
elevran: is this the right location for the import?
shouldn't `plugins/requestcontrol/dataproducer/tokenizer` be placed after `plugins/datalayer/extractor/models`?

PR #830, opened 2026-04-14
update image-build-builder rule

### LINKED ISSUE #829: Missing builder image
The current build and test process relies on the BUILDER_IMAGE (`ghcr.io/llm-d/llm-d-builder`). This image is not published to a remote repository and exists only locally.

To avoid rebuilding the image for every `make lint` or `make test` execution, we use a local file check called [BUILDER_STAMP](https://github.com/llm-d/llm-d-inference-scheduler/blob/bd4eb93e88cfb0e540d5e5c3ab32646cff933179/Makefile#L347). 

The problem occurs when the local image is deleted while the `BUILDER_STAMP` file still exists. In this case, subsequent commands like `make lint` attempt to pull the image from the remote repository and fail. 

The logic should be:

1. Image doesn't exist locally → build
2. Stamp file missing → build
3. Dockerfile.builder is newer than the stamp → build

All three must be false to skip the build.

roytman:
Update `image-build-builder` Makefile rule, to check that `BUILDER_IMAGE` exsists locally.

/fix https://github.com/llm-d/llm-d-inference-scheduler/issues/829

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -346,16 +346,16 @@ image-build-%: check-container-tool ## Build Container image using $(CONTAINER_R
 
 BUILDER_STAMP = build/.builder.stamp
 
-$(BUILDER_STAMP): Dockerfile.builder | check-container-tool
-	@printf "\033[33;1m==== Building image $(BUILDER_IMAGE) ====\033[0m\n"
-	$(CONTAINER_RUNTIME) build \
-		-f Dockerfile.builder \
-		-t $(BUILDER_IMAGE) .
-	@mkdir -p $(dir $(BUILDER_STAMP))
-	@touch $(BUILDER_STAMP)
-
 .PHONY: image-build-builder
-image-build-builder: $(BUILDER_STAMP)
+image-build-builder: | check-container-tool ## Build builder image if missing locally or Dockerfile.builder changed
+	@if ! $(CONTAINER_RUNTIME) image inspect $(BUILDER_IMAGE) >/dev/null 2>&1 || \
+	    [ ! -f $(BUILDER_STAMP) ] || \
+	    [ Dockerfile.builder -nt $(BUILDER_STAMP) ]; then \
+		printf "\033[33;1m==== Building image $(BUILDER_IMAGE) ====\033[0m\n"; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -346,16 +346,16 @@ image-build-%: check-container-tool ## Build Container image using $(CONTAINER_R
...
-	@printf "\033[33;1m==== Building image $(BUILDER_IMAGE) ====\033[0m\n"
-	$(CONTAINER_RUNTIME) build \
-		-f Dockerfile.builder \
-		-t $(BUILDER_IMAGE) .
-	@mkdir -p $(dir $(BUILDER_STAMP))
-	@touch $(BUILDER_STAMP)
-
 .PHONY: image-build-builder
-image-build-builder: $(BUILDER_STAMP)
+image-build-builder: | check-container-tool ## Build builder image if missing locally or Dockerfile.builder changed
```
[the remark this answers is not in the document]
roytman: done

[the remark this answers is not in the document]
roytman: done

Makefile
```diff
@@ -346,16 +346,16 @@ image-build-%: check-container-tool ## Build Container image using $(CONTAINER_R
...
-		-f Dockerfile.builder \
-		-t $(BUILDER_IMAGE) .
-	@mkdir -p $(dir $(BUILDER_STAMP))
-	@touch $(BUILDER_STAMP)
-
 .PHONY: image-build-builder
-image-build-builder: $(BUILDER_STAMP)
+image-build-builder: check-container-tool ## Build builder image if missing locally, stamp missing, or Dockerfile.builder newer than stamp
+	@if ! $(CONTAINER_RUNTIME) image inspect $(BUILDER_IMAGE) >/dev/null 2>&1 || \
+	    [ ! -f $(BUILDER_STAMP) ] || \
```
shmuelk, line 352: Why are you keeping the `$(BUILDER_STAMP)` file?

roytman: to guarantee that if Dockerfile.builder is updated, we will rebuild the image.

shmuelk: ok

Issue #813, opened 2026-04-12
Refactor label-based filters


ahg-g:
**What would you like to be added**:

1) Deprecate `by-label` filter, the `by-label-selector` filter should cover all its use cases
2) Rename `by-label-selector` to `label-selector-filter`
3) The role instances of the `by-label` should be based on the `label-selector-filter`
4) The role filters should be named `prefill-filter` and `decode-filter` etc.

**Why is this needed**:

Remove redundant filter and align naming with the conventions established upstream.

ahg-g:
@elevran PTAL and let me know what you think.

elevran:
IIRC there was some caveat on selector not being to express OR relations, so having a default role when no label is used did not work. If we assume role labels are mandatory then all can be folded into the label selector, otherwise allowing defaults should wrap around the selector.
No issue with the renames.

szedan-rh:
hi @elevran / @ahg-g - Would like to work on this issue if no objection from your side

PR #842, opened 2026-04-19
Refactor label-based filters: rename, deprecate, and add runtime warn…

szedan-rh:
/kind bug

Addresses #813.
- Rename `by-label-selector` to `label-selector-filter` as the canonical type
- Deprecate `by-label` filter in favor of `label-selector-filter` and role-specific filters
- Introduce `RoleFilter` struct backed by k8s label selectors, replacing `ByLabel` for role filters
- Register legacy type names (`by-label`, `by-label-selector`) as aliases for backward compatibility
- Add runtime deprecation log warnings in legacy factories, matching existing project conventions
- Add Go type/function aliases (`LabelSelectorFilter`, `NewLabelSelectorFilter`) for discoverability
- Preserve legacy `TypedName.Type` for plugins created through deprecated factories
- Update docs and YAML examples to reflect new canonical names


pkg/epp/framework/plugins/scheduling/filter/bylabel/roles.go
```diff
@@ -37,14 +41,78 @@ const (
...
+	return &RoleFilter{
+		typedName:     plugin.TypedName{Type: typeName, Name: typeName},
+		selector:      selector,
+		labelKey:      labelKey,
+		allowsNoLabel: allowsNoLabel,
+	}
+}
+
 // DecodeRoleFactory defines the factory function for the Decode filter.
 func DecodeRoleFactory(name string, _ json.RawMessage, _ plugin.Handle) (plugin.Plugin, error) {
```
ahg-g, line 41: @elevran  brain storming here: do we really need to have those hardcoded per-role filters at all? Everyone of them can easily be expressed in the config as a separate instance of the `label-selector-filter`, so why do we need to assume the burden of these special filters?

pkg/epp/framework/plugins/scheduling/filter/bylabel/selector.go
```diff
@@ -8,29 +8,56 @@ import (
...
 	}
 	return NewSelector(name, &parameters)
 }
 
+// DeprecatedSelectorFactory creates a Selector but preserves the legacy TypedName.Type
+// so that plugins created through this factory report "by-label-selector"
+// rather than the canonical "label-selector-filter". It also logs a deprecation warning.
+//
+// Deprecated: Use SelectorFactory instead.
+func DeprecatedSelectorFactory(name string, rawParameters json.RawMessage, handle plugin.Handle) (plugin.Plugin, error) {
```
hexfusion, line 49: question: since there are no callers do we need this at all?

pkg/epp/framework/plugins/scheduling/filter/bylabel/roles.go
```diff
@@ -37,14 +41,78 @@ const (
...
 )
 
+// compile-time interface assertion
+var _ scheduling.Filter = &RoleFilter{}
+
+// RoleFilter filters endpoints by their role label value, with optional
+// fallback to accept endpoints that lack the role label entirely.
+// It composes a labels.Selector for value matching and adds the
+// "allowsNoLabel" OR-fallback that cannot be expressed with a pure k8s LabelSelector.
+type RoleFilter struct {
```
ahg-g: pls remove this for now and all new role-based filters created based on it below. I think we should try to use the `label-selector-filter` as suggested in https://github.com/llm-d/llm-d-inference-scheduler/pull/842/changes#r3111853481

szedan-rh: Done.

szedan-rh:
@ahg-g / @elevran  - could you please take another look?

docs/disaggregation.md
```diff
@@ -244,7 +244,10 @@ role: prefill
 role: decode
 ```
 
-To accommodate this **without code changes**, you can configure the **EndpointPickerConfig** to use the generic `by-label` filter plugin instead of the hardcoded `encode-filter` / `prefill-filter` / `decode-filter`.
+To accommodate this **without code changes**, you can configure the **EndpointPickerConfig** to use the generic `label-selector-filter` plugin instead of the hardcoded `encode-filter` / `prefill-filter` / `decode-filter`.
+
+> [!NOTE]
+> The previous filter type `by-label` is deprecated. Use `label-selector-filter` with standard Kubernetes label selector syntax instead.
```
ahg-g, line 250:
```suggestion

```

ahg-g (commented):
Thanks, a couple of nits

ahg-g:
> Addresses https://github.com/llm-d/llm-d-inference-scheduler/issues/813.

We still need to keep the issue open to track removing the older version in two releases.

pkg/epp/framework/plugins/scheduling/filter/bylabel/README.md
```diff
@@ -23,33 +26,31 @@ Retains only candidate pods that carry a specific Kubernetes label with one of a
...
 ```
 
-### Limitations
-
-- Only exact string equality is checked — no wildcards, prefix matching, or regular expressions.
-
 ---
 
-## ByLabelSelector
+## ByLabel (Deprecated)
```
ahg-g, line 44: pls indicate which release this will be removed, it should be v0.10.0

szedan-rh: Thank you, updated.

PR #843, opened 2026-04-19
migration scripts: improve tooling robustness

elevran:
Two independent fixes to the migration scripts:

- `check-commits.sh`: restrict the signed-commits check to first-parent history to avoid false failures after merging upstream GAIE history.
- `migrate-gaie-paths.sh`: replace the coarse directory-exists check with Go package conflict detection, failing only when the destination already contains .go files.

### DIFF
--- a/scripts/check-commits.sh
+++ b/scripts/check-commits.sh
@@ -10,12 +10,12 @@ BASE_BRANCH=${1:-upstream/main}
 BASE=$(git merge-base HEAD "$BASE_BRANCH")
 RANGE="$BASE..HEAD"
 
-echo "Checking $(git rev-list --count "$RANGE") commits ($BASE..HEAD)..."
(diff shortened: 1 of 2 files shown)


ahg-g:
/approve

We are including the charts in the move, right?

elevran:
the first move will be everything under apix/config/, pkg/epp/, pkg/common/, internal/ and test/util/ (only one not under the same path as the source since there is already a test/util directory with Go files).
we will need to do follow ups to move other bits and pieces (e.g., charts, integration and e2e, makefile. etc.) in separate PRs as these have more direct conflicts

nirrozenbaum:
please keep the existing code in k8s upstream “frozen” and don’t delete, as bbr currently has some references to epp (e.g., plugin.TypedName). 
I’m planning to start working on migration on the week of April 27th. 
once we complete both migrations we can cleanup

elevran:
👍 no code is being deleted at this time.

Issue #538, opened 2026-01-06
[Tracking] Flow Control Tasks


LukeAVanDrie:
Tracks tasks related to Flow Control (request buffering for online inference with priority, fairness, and ordering features).

- Draft UG (to be replaced with dev site link soon): https://docs.google.com/document/d/1sesId2WOw6QnKTGQqnc4_b9us6zmQb6x90NqmZ-p9Lk/edit?tab=t.0
- [Open tasks](https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues?q=is%3Aissue%20state%3Aopen%20milestone%3A%22Flow%20Control%22)
- Roadmap: coming soon

PR #860, opened 2026-04-21
fix: Handle string array prompts in /v1/completions scoring

### LINKED ISSUE #848: Handle array of prompts passed in `/completions`
Ability to process different prompt types in tokenization and scheduling. Currently only a single string prompt is supported.
Need to 
- determne how this is treated by the model servers (e.g., parts of a single prompt or unrelated prompts)
- address use in code (e.g., scorers, tokenization sidecar, etc)

**Why is this needed**:
The OpenAI API spec allows different prompt types - `string` (prompt), `[]string` (prompt array), `[]int` (tokenized prompt) and `[][]int` (multiple tokenized prompts).
This was introduced in GAIE v1.5. - https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2724.

cc: @vMaroon

### LINKED ISSUE #1388: Refactor `Prompt` and `TokenizedPrompt` types to uniform slice representation
### Context

From PR review discussion on #860 only `/v1/completions` has the "array-of" prompt format, but the current type design forces every consumer to branch on which field is populated (`Raw` vs `Strings`, `TokenIDs` vs `PerPromptTokens`).

### Problem

The `Prompt` struct uses three separate fields to represent what is fundamentally the same data at different cardinalities:

```go
type Prompt struct {
    Raw      string    // single string prompt
    Strings  []string  // array-of-strings prompt
    TokenIDs []uint32  // array-of-token-IDs prompt
}
```

Similarly, `TokenizedPrompt` stores a redundant flat representation alongside the per-prompt breakdown:

```go
type TokenizedPrompt struct {
    TokenIDs        []uint32     // flat concatenation
    PerPromptTokens [][]uint32   // per-prompt breakdown (only when >1 prompt)
    // ...
}
```

This causes:
- Branching in every consumer (tokenizer, precise prefix cache scorer, context-length-aware scorer) to check which field is populated
- A redundant invariant (`concat(PerPromptTokens...) == TokenIDs`) that must be validated at runtime
- A dedicated validation function (`perPromptTokenGroups`) that exists only to check this invariant

### Proposal

Collapse both types to always use slices, treating single-prompt as a length-1 slice:

```go
type Prompt struct {
    Strings  []string    // ["hello"] or ["hello", "world"]
    TokenIDs [][]uint32  // [[1,2,3]] or [[1,2,3],[4,5,6]]
}

type TokenizedPrompt struct {
    TokenIDs           [][]uint32  // [[tokens_prompt_1], [tokens_prompt_2]]
    MultiModalFeatures []MultiModalFeature
}
```

Apply the same change to `EmbeddingsInput` which mirrors `Prompt`.

### Dependencies

This should be done after #860 lands, since that PR adds the multi-prompt support that this refactor simplifies.

cc: @elevran

albertoperdomo2:
**What type of PR is this?**

/kind bug
/kind feature

## What this PR does

Fixes remaining call sites that read `Prompt.Raw` directly instead of handling `Prompt.Strings`, which broke scoring and disaggregation decisions when clients send `/v1/completions` requests with an array of strings (e.g., `"prompt": ["prompt1", "prompt2"]`).

Replaces the dual-field `TokenizedPrompt` representation (`TokenIDs []uint32` + `PerPromptTokens [][]uint32`) with a single `PerPromptTokens [][]uint32` field. Single-prompt requests use a length-1 outer slice; multi-string completions use one inner slice per prompt string. Two accessor methods are added:

- `TokenCount() int`: returns the total token count across all prompts (zero-alloc, nil-safe).
- `FlatTokenIDs() []uint32`: returns the flat concatenation. For single-prompt (the common case), returns the inner slice directly with no allocation.

The prefix cache scorer iterates `PerPromptTokens` directly for independent per-prompt scoring. All other consumers use `TokenCount()` or `FlatTokenIDs()`.

## Why we need it

The OpenAI `/v1/completions` API allows `prompt` to be either a `string` or a `string[]`. GAIE v1.5 added support for both via the `Prompt` struct (`Raw` and `Strings` fields), but two consumers still read `Prompt.Raw` directly:

- **precise-prefix-cache scorer**: `GetPodScores`, `ComputeBlockKeys`, and `PrepareRequestData` received an empty string for array prompts, producing no cache keys and no scores.
- **prefix-based PD decider**: `getUserInputLenInTokens` returned 0 for array prompts, so the disaggregation threshold was never met.

vLLM and other model servers treat `string[]` prompts as **independent batch prompts**, so each string is a separate inference request with its own KV-cache prefix chain starting from token 0.

This means different consumers need different handling:

- **PD decider / context-length-aware**: Use `TokenCount()`. These consumers estimate total work, so combined token count is the right signal.

- **Prefix cache scorer**: Score each prompt string **independently** and sum scores per endpoint. This is necessary because `LongestPrefixScorer` walks block keys from position 0 and uses set intersection, and once a pod drops out, it never comes back. With concatenated block keys from multiple prompts, the intersection breaks at prompt boundaries (different prefix chains), causing pods with only non-first prompts cached to get **zero score** despite real cache hits.

Example with `["hello", "foo", "world"]`:

#### Concatenated scoring:
Pod A (has "hello"): score = 3 (drops at "foo" boundary)
Pod B (has "foo"+"world"): score = 0 (not in keys[0] = hello's first block)
Pod C (has all three): score = 9

#### Independent scoring:
Pod A: score("hello")=3 + score("foo")=0 + score("world")=0 = 3
Pod B: score("hello")=0 + score("foo")=3 + score("world")=3 = 6
Pod C: score("hello")=3 + score("foo")=3 + score("world")=3 = 9

Pod B correctly outscores Pod A, since it has more cached.

### Why a single `PerPromptTokens` field instead of dual `TokenIDs` + `PerPromptTokens`

Storing both a flat concatenation and a per-prompt split creates a drift invariant: the two must always be consistent. This required `len > 1` gates in producers to decide when to populate the second field, and would eventually need a validator to assert `Concat(PerPromptTokens) == TokenIDs`.

Collapsing to a single `PerPromptTokens [][]uint32` eliminates the invariant entirely. Single-prompt is a length-1 slice, so no special case, no gate, no drift. The `FlatTokenIDs()` method provides the concatenated view on demand, and returns the inner slice directly (zero-alloc) for the single-prompt common case.

Fixes #848
Fixes #1388 

```release-note
NONE
```


albertoperdomo2:
cc: @elevran

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -172,18 +186,40 @@ func (p *Plugin) tokenize(ctx context.Context, request *scheduling.LLMRequest) (
...
+				if renderErr != nil {
+					logger.Error(renderErr, "String tokenization failed, skipping")
+					continue
+				}
+				allTokenIDs = append(allTokenIDs, sTokenIDs)
+			}
+		} else {
+			traceLogger.Info("Calling Render for completions", "prompt", prompt)
+			var tokenIDs []uint32
+			tokenIDs, _, err = p.tokenizer.Render(prompt.PlainText())
```
elevran: @vMaroon can we optimize this to avoid multiple roundtrips? For example, have the tokenizer always accept an array?

albertoperdomo2: IMHO it is feasible, since vLLM supports batching natively. I can work on that, proposing a new RPC method that supports `repeated string`.

elevran: thanks. Please coordinate with @vMaroon it's an area I'm less familiar with

vMaroon: We can add a **new** API yes, I don't expect any significant benefit over sending requests **in parallel** - but this would be aligned with the incoming vLLM-/SGLang- direct tokenization plugins that do support such an API.

elevran: Likely correct from a performance perspective. Speaking from a networking service point of view:
- single call that either fails or succeeds is much is easier to manage and reason about than N independent calls to the sidecar, and deciding how to react when only 1 is slow or fails (retries, timeouts, ...)
- N requests do have a higher load on the sidecar (connection states, orderring, serialization, ...)

So would suggest we target a new API to support multiple inputs/outputs.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -174,18 +188,40 @@ func (p *Plugin) tokenize(ctx context.Context, request *scheduling.InferenceRequ
...
 	switch {
 	case request.Body.Completions != nil:
-		traceLogger.Info("Calling Render for completions", "prompt", request.Body.Completions.Prompt)
-		tokenIDs, _, err = p.tokenizer.Render(request.Body.Completions.Prompt.Raw)
+		prompt := request.Body.Completions.Prompt
+		if len(prompt.Strings) > 0 {
+			allTokenIDs = make([][]uint32, 0, len(prompt.Strings))
+			for _, promptStr := range prompt.Strings {
+				traceLogger.Info("Calling Render for completions string", "promptLength", len(promptStr))
+				sTokenIDs, _, renderErr := p.tokenizer.Render(promptStr)
```
hexfusion: minor: prefer the conventional err here for consistency 
```suggestion
				sTokenIDs, _, err := p.tokenizer.Render(promptStr)
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -174,18 +188,40 @@ func (p *Plugin) tokenize(ctx context.Context, request *scheduling.InferenceRequ
...
+				if renderErr != nil {
+					logger.Error(renderErr, "String tokenization failed, skipping")
+					continue
+				}
+				allTokenIDs = append(allTokenIDs, sTokenIDs)
+			}
+		} else {
+			traceLogger.Info("Calling Render for completions", "prompt", prompt)
+			var tokenIDs []uint32
```
hexfusion: minor: no need to allocate here if you close the scope below
```suggestion
    tokenIDs, _, err := p.tokenizer.Render(prompt.PlainText())
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -174,18 +188,40 @@ func (p *Plugin) tokenize(ctx context.Context, request *scheduling.InferenceRequ
...
+			tokenIDs, _, err = p.tokenizer.Render(prompt.PlainText())
+			if err == nil {
+				allTokenIDs = [][]uint32{tokenIDs}
+			}
+		}
 	case request.Body.ChatCompletions != nil:
 		renderReq := ChatCompletionsToRenderChatRequest(request.Body.ChatCompletions)
 		traceLogger.Info("Calling RenderChat for chat completions", "messageCount", len(request.Body.ChatCompletions.Messages))
+		var tokenIDs []uint32
```
hexfusion: same here the scope of var is tied to the switch case right so no need to allocate
```suggestion
       tokenIDs, mmFeatures, err := p.tokenizer.RenderChat(renderReq)
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -174,18 +188,40 @@ func (p *Plugin) tokenize(ctx context.Context, request *scheduling.InferenceRequ
...
 		tokenIDs, mmFeatures, err = p.tokenizer.RenderChat(renderReq)
+		if err == nil {
```
hexfusion: what about if the err != nil should we log that?:

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -196,8 +232,12 @@ func (p *Plugin) tokenize(ctx context.Context, request *scheduling.InferenceRequ
 		return nil, nil
 	}
 
-	traceLogger.Info("Tokenization succeeded", "tokenCount", len(tokenIDs))
-	return tokenIDs, mmFeatures
+	totalTokens := 0
+	for _, ids := range allTokenIDs {
+		totalTokens += len(ids)
+	}
+	traceLogger.Info("Tokenization succeeded", "tokenCount", totalTokens, "promptCount", len(allTokenIDs))
```
hexfusion: if you really want the total, accumulate inside the branches instead of re-iterating?

```suggestion
   traceLogger.Info("Tokenization succeeded", "promptCount", len(allTokenIDs))
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer_preparedata.go
```diff
@@ -47,13 +47,18 @@ func (p *Plugin) PrepareRequestData(ctx context.Context, request *scheduling.Inf
...
 	}
 
+	// Flatten per-prompt tokens into a single slice for the GAIE TokenizedPrompt
+	var flat []uint32
+	for _, ids := range allTokenIDs {
+		flat = append(flat, ids...)
+	}
 	request.TokenizedPrompt = &scheduling.TokenizedPrompt{
-		TokenIDs: tokenIDs,
+		TokenIDs: flat,
```
hexfusion: take a look at slices.Concat it does this for you and pre allocates
```suggestion
	TokenIDs: slices.Concat(allTokenIDs...)
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer_scorer.go
```diff
@@ -49,12 +49,24 @@ func (p *Plugin) Score(ctx context.Context, cycleState *scheduling.CycleState, r
...
+			totalLen := 0
+			for _, ids := range allTokenIDs {
+				totalLen += len(ids)
+			}
+			flat := make([]uint32, 0, totalLen)
+			for _, ids := range allTokenIDs {
+				flat = append(flat, ids...)
+			}
+			state := &TokenizedPromptState{
+				TokenIDs:   flat,
```
hexfusion:
```suggestion
  state := &TokenizedPromptState{
      TokenIDs:   slices.Concat(allTokenIDs...),
      MMFeatures: mmFeatures,
  }
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
albertoperdomo2: It's a bit counterintuitive, but at the line 230 we log if it errors.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
albertoperdomo2: I think promptCount suffices here.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer_preparedata.go
albertoperdomo2: right, great catch, thanks! 🙏🏽

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
albertoperdomo2: See [llm-d-kv-cache/pull/538](https://github.com/llm-d/llm-d-kv-cache/pull/538)

vMaroon: @albertoperdomo2 @elevran I think we should do multiple round-trips in this PR, and update to one call in a follow-up with benchmarks.

albertoperdomo2: agree, we can work on [#538](https://github.com/llm-d/llm-d-kv-cache/pull/538) after this one is merged and we properly benchmark it.

elevran:
@albertoperdomo2 can you please resolve conflicts and lint error?

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -358,6 +358,61 @@ func (s *Scorer) PrepareRequestData(ctx context.Context,
...
+			}
+			addr := fmt.Sprintf("%s:%s", md.Address, md.Port)
+			matchLen := int(aggregatedScores[addr])
+			ep.Put(approxprefix.PrefixCacheMatchInfoKey,
+				approxprefix.NewPrefixCacheMatchInfo(matchLen, totalBlocks, blockSize))
+		}
+
+		// Save ALL block keys (concatenated) for PreRequest speculative insertion
+		s.pluginState.Write(request.RequestID, stateKey, &precisePluginState{
+			blockKeys: allBlockKeys,
```
vMaroon: Speculative insertion assumes that the blocks form a prefix-chain of one sequence:
```
    // Pass nil engineKeys: speculative entries only need requestKey -> PodEntry mapping.
	// Engine keys will be linked later when confirmed KV events arrive.
	if err := index.Add(ctx, nil, state.blockKeys, []kvblock.PodEntry{speculativePod}); err != nil {
		logger.Error(err, "Failed to add speculative entries to index",
			"pod", speculativePod.PodIdentifier)
	}
```

It also needs to be aware of the multiple prompts, and to add an entry per prompt.

albertoperdomo2:
@elevran could you kick off the PR checks? thanks 🙏🏽

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -91,7 +91,13 @@ type InferenceRequestBody struct {
 // actual token data such as prefix-cache awareness.
 type TokenizedPrompt struct {
 	// TokenIDs are the token IDs for the prompt, including multimodal placeholder tokens.
+	// For multi-prompt completions this is the concatenation of all per-prompt tokens.
```
elevran: Only `/completions` has the "array-of" format so we need to look at two fields (raw prompt or multi-string-prompt; TokenIDs or PerPromptTokens, ...).

The use of arrays could simplify some of the code paths
Does it make sense to have the prompt (and tokenized prompt) types as 
- []string and
- [][]uint32?
making the case of a single entry a specical case ([1]string or [1][]int32).

albertoperdomo2: Good point, collapsing to `[]string` / `[][]uint32` with single-entry as the special case would clean up the branching in the scoring paths. Happy to do that refactor in this PR since I'm already touching these code paths, or I can split it into a follow-up if you prefer to keep this one scoped to the fix.

elevran: either way works, but different PR feels more scoped to me.
Would you be replacing with array for both the prompt strings and tokens?
@ahg-g - any objection to assuming plural inputs via arrays as the request prompt and tokenized prompt internal representations?

albertoperdomo2: got it, then I'll do it in a separate PR. And yeah, the plan would be to replace both: `Prompt` would use `[]string` instead of` Raw`/`Strings`, and `TokenizedPrompt` would use `[][]uint32` instead of `TokenIDs`/`PerPromptTokens`, with single-entry slices for the single-prompt case.

ahg-g: > @ahg-g - any objection to assuming plural inputs via arrays as the request prompt and tokenized prompt internal representations?

No particular objections, but we do need to revisit the whole `InferenceRequest` API to decouple from the embedded OpenAI fields and move towards a more generic one that allows us to also support messages api.

elevran:
needs conflict resolution and green CI run.
Assuming resolution does not introduce new issues, we can likely mnerged based on previous reviews)

ahg-g:
Please be aware that https://github.com/llm-d/llm-d-router/pull/1380 will standardize prompt representation in the InferenceRequestBody, it will be api-agnostic, and all downstream plugins are expected to use this new type (the plan is to remove all the api-specific types from InferenceRequestBody)

albertoperdomo2:
@ahg-g so this PR should be held until #1380 is merged and adapted to the new `InferenceRequestBody` afterwards? or how can I help in that effort?

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -97,7 +97,13 @@ type InferenceRequestBody struct {
...
 	TokenIDs []uint32
+	// PerPromptTokens holds the token IDs for each individual prompt string.
+	// Set only when the request carries a []string prompt with more than one element.
+	// Consumers that need per-prompt granularity (e.g. prefix-cache scoring) use
+	// this field; all others use the flat TokenIDs.
+	PerPromptTokens [][]uint32
```
hyeongyun0916, line 110: could PerPromptTokens replace the flat TokenIDs instead of sitting beside it?

Now that PerPromptTokens carries the per-prompt split, would it make sense to drop the flat TokenIDs and derive it on demand (slices.Concat(perPromptTokens...))? A single prompt is just a one-element slice (N=1).

The reason this matters: storing both a flat and a split form means they can drift, which is exactly why this PR needs the perPromptTokenGroups validator (the Concat(PerPromptTokens) == TokenIDs check) and the len > 1 gate before populating the field.

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
hyeongyun0916: Changing blockKeys from [] to [][] means totalBlocks = len(pluginStateData.blockKeys) now counts prompts, not blocks.

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -716,8 +788,24 @@ func (s *Scorer) computeBlockKeys(ctx context.Context,
...
+			for _, promptStr := range prompt.Strings {
+				//nolint:staticcheck // SA1019: legacy path retained for tokenizersPoolConfig configs.
+				pKeys, pErr := s.kvCacheIndexer.ComputeBlockKeys(ctx, nil, promptStr, request.TargetModel)
+				if pErr != nil {
+					if errors.Is(pErr, kvcache.ErrInternalTokenizationDisabled) {
+						return nil, nil
+					}
+					return nil, pErr
+				}
+				allBlockKeys = append(allBlockKeys, pKeys...)
```
elevran: Q: `allBlockKeys = append(allBlockKeys, pKeys...)` concatenates block keys across prompts into a flat slice.
The multi-prompt path in `Produce()` explicitly avoids this with the comment "Flattening here would create cross-prompt block adjacency that never exists in the model server cache." However, `getScores()` already handles `prompt.Strings` before it reaches `computeBlockKeys`, so this branch appears unreachable through the normal call path. Either remove it (if dead code) or add a comment explaining why the legacy concatenation is acceptable on this path and that the normal call path does not reach here.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -232,11 +259,22 @@ func (p *Plugin) tokenize(ctx context.Context, request *scheduling.InferenceRequ
 		return nil, fmt.Errorf("tokenization failed: %w", err)
 	}
 
-	traceLogger.Info("Tokenization succeeded", "tokenCount", len(tokenIDs))
-	return &fwkrh.TokenizedPrompt{
+	tokenIDs := slices.Concat(allTokenIDs...)
+
+	totalTokens := 0
```
elevran: nit: `totalTokens` is accumulated in a loop but equals `len(tokenIDs)` since `tokenIDs = slices.Concat(allTokenIDs...)`. The loop and variable can be removed; use `len(tokenIDs)` in the log line directly.

elevran:
@albertoperdomo2 can you take a look at the conflicts please?
@vMaroon will #1445 handle this input as well?

pkg/epp/framework/interface/requesthandling/types.go
albertoperdomo2: This makes complete sense, but we agreed on tackling it in a separate PR.

ahg-g: Why do you think this should be addressed in a separate PR? I think is a core part of the solution.

ahg-g (commented):
Thanks @albertoperdomo2; https://github.com/llm-d/llm-d-router/pull/1445 merged, can you pls rebase?

albertoperdomo2:
@ahg-g yeah I'm fixing a couple of conflicts and I'll ping when ready.

pkg/epp/framework/interface/requesthandling/types.go
albertoperdomo2: We agreed on addressing it in a separate PR previously, but I'm happy with taking care of it in this one.

ahg-g: I recommend to address it in this PR.

albertoperdomo2:
@ahg-g @elevran just pushed the changes of the refactor, dropping `TokenIDs` in favor of `PerPromptTokens`.

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -98,15 +99,49 @@ type InferenceRequestBody struct {
...
+	for _, pp := range tp.PerPromptTokens {
+		n += len(pp)
+	}
+	return n
+}
+
+// FlatTokenIDs returns all tokens as a single slice. For single-prompt
+// requests (the common case) it returns the inner slice directly with no
+// allocation. For multi-prompt requests it concatenates.
+func (tp *TokenizedPrompt) FlatTokenIDs() []uint32 {
```
ahg-g: I don't think we should provide this function, it makes a copy of the tokens in all cases, even in the most common case where the prompt is a single string. The consumers should simply iterate over the `PerPromptTokens` list.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -243,8 +243,10 @@ func (p *Plugin) Produce(ctx context.Context, request *scheduling.InferenceReque
 	if err != nil {
 		return err
 	}
+	if tp == nil || tp.TokenCount() == 0 {
+		return nil
+	}
```
ahg-g, line 266: Why adding this? unless there is an error, I think we should simply accept the produce output.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/blockkeys.go
```diff
@@ -31,25 +32,54 @@ type kvCacheIndexer interface {
...
+	for _, tokens := range tp.PerPromptTokens {
+		if len(tokens) == 0 {
+			continue
+		}
+		// MM features apply only to single-prompt requests (chat); multi-prompt
+		// completions never carry multimodal content.
+		var mmf []fwkrh.MultiModalFeature
+		if len(tp.PerPromptTokens) == 1 {
+			mmf = tp.MultiModalFeatures
+		}
```
ahg-g, line 63: We should not make any assumptions about how different protocols work, if there is something special about "multi-prompt completions", then it needs to be captured in the `TokenizedPrompt` api and here we just work with whatever `TokenizedPrompt` offers.

/cc @vMaroon

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
```diff
@@ -129,7 +135,8 @@ func (p *Producer) PreRequest(ctx context.Context,
 	}
 	p.pluginState.Delete(request.RequestID)
 
-	if len(state.blockKeys) == 0 {
+	allBlockKeys := slices.Concat(state.perPromptKeys...)
```
ahg-g: We are being too liberal with copying lists and I am not sure it is necessary in all cases, this impacts the epp's memory footprint, something we need to be very cognizant about as we expand to support agentic workloads which carry > 100k tokens per request. Is this concat necessary?

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -57,12 +57,12 @@ func getBlockHashes(ctx context.Context, request *scheduling.InferenceRequest, b
...
 
 	tp := request.Body.TokenizedPrompt
-	if tp == nil || len(tp.TokenIDs) == 0 {
+	if tp == nil || tp.TokenCount() == 0 {
 		loggerDebug.Info("TokenizedPrompt is empty, skipping hashing")
 		return nil
 	}
 
-	seq := getKVCacheBlocksFromTokens(tp.TokenIDs, blockSizeTokens)
+	seq := getKVCacheBlocksFromTokens(tp.FlatTokenIDs(), blockSizeTokens)
```
ahg-g: Doesn't this mean we are treating the separate prompts as if they are a single long prompt (meaning the prompts preceding a specific prompt are treated as a prefix to it)? is this actually accurate? Aren't the separate prompts literally independent with respect to prefix caching?

@liu-cong @zetxqx for thoughts

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
```diff
@@ -46,27 +46,40 @@ func (b estimateBackend) produce(_ context.Context, body *fwkrh.InferenceRequest
 	switch {
 	case body.Generate != nil:
 		return &fwkrh.TokenizedPrompt{
-			TokenIDs:           body.Generate.TokenIDs,
+			PerPromptTokens:    [][]uint32{body.Generate.TokenIDs},
 			MultiModalFeatures: convertMMFeaturesToUpstream(body.Generate.Features),
 		}, nil
 	case body.Completions != nil && len(body.Completions.Prompt.TokenIDs) > 0:
```
ahg-g, line 56: Completions support both a single and list of strings, does this also apply to tokens? a single array and a list of token arrays?

@kaushikmitr @liu-cong for thoughts.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
```diff
@@ -46,27 +46,40 @@ func (b estimateBackend) produce(_ context.Context, body *fwkrh.InferenceRequest
...
-		return &fwkrh.TokenizedPrompt{TokenIDs: body.Completions.Prompt.TokenIDs}, nil
+		return &fwkrh.TokenizedPrompt{PerPromptTokens: [][]uint32{body.Completions.Prompt.TokenIDs}}, nil
 	case body.Embeddings != nil && len(body.Embeddings.Input.TokenIDs) > 0:
-		return &fwkrh.TokenizedPrompt{TokenIDs: body.Embeddings.Input.TokenIDs}, nil
+		return &fwkrh.TokenizedPrompt{PerPromptTokens: [][]uint32{body.Embeddings.Input.TokenIDs}}, nil
 	}
```
ahg-g, line 60: @liu-cong @zetxqx  a switch statement with a fallback handling makes it easy to miss adding support for new protocols/messages. I think this switch statement should explicitly list all the cases, and we adda lint check that fails if a new type is added and didn't have a case here. This will be possible once https://github.com/llm-d/llm-d-router/issues/1472 is addressed.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
```diff
@@ -46,27 +46,40 @@ func (b estimateBackend) produce(_ context.Context, body *fwkrh.InferenceRequest
...
-	return &fwkrh.TokenizedPrompt{TokenIDs: packBytes(raw)}, nil
+	return &fwkrh.TokenizedPrompt{PerPromptTokens: [][]uint32{packBytes(raw)}}, nil
+}
+
+func estimateMultiStringCompletions(req *fwkrh.CompletionsRequest) (*fwkrh.TokenizedPrompt, error) {
+	allTokenIDs := make([][]uint32, 0, len(req.Prompt.Strings))
+	for _, s := range req.Prompt.Strings {
+		ids := packBytes([]byte(s))
+		allTokenIDs = append(allTokenIDs, ids)
```
ahg-g, line 96: @zetxqx @liu-cong is this correct? can we simply concat the strings here? I recall there is a case where the different prompts represent batch of independent prompts. which one is that?

ahg-g (commented):
Thanks @albertoperdomo2 !

pkg/epp/framework/interface/requesthandling/types.go
albertoperdomo2: In the common case it does return the slice directly, but I agree on the simplification, I'll fix it.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
albertoperdomo2: yes, as per the [official docs](https://developers.openai.com/api/reference/resources/completions/methods/create#(resource)%20completions%20%3E%20(method)%20create%20%3E%20(params)%200.non_streaming%20%3E%20(param)%20prompt%20%3E%20(schema)):
> prompt: string or array of string or array of number or array of array of number
> The prompt(s) to generate completions for, encoded as a string, array of strings, array of tokens, or array of token arrays.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
albertoperdomo2: Your observation is correct. I forgot to fix this after dropping the flat `TokenIDs`. Just pushed the changes for this and matching multi-prompt tests.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
albertoperdomo2: In this case we are not concatenating the strings, each string is estimated independently into its own `PerPromptTokens` entry (note the missing `...`).

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
ahg-g: Should we address it?

albertoperdomo2: Now that we are at it, I think it will be an easy change and needs to be done at some point, so yes.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
```diff
@@ -179,22 +179,27 @@ func (p *dataProducer) Produce(ctx context.Context, request *fwksched.InferenceR
...
+	perPromptHashes := getBlockHashes(ctx, request, blockSize, maxBlocks)
+
+	prefixCacheServers := make(map[ServerID]int)
+	totalBlocks := 0
+	for _, hashes := range perPromptHashes {
+		for server, matchLen := range p.matchLongestPrefix(ctx, hashes) {
+			prefixCacheServers[server] += matchLen
+		}
+		totalBlocks += len(hashes)
+	}
```
zetxqx, line 191: @liu-cong here it's using all the prefix cache match among the batched prompts as the matchLen for each server. I think it makes sense, just want to flag this for you to confirm

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
zetxqx: Add a note here, to completely accept array of tokens input, we need to add a new field to the Completions.Prompt for array of tokenization input. https://github.com/llm-d/llm-d-router/blob/c45f49d00401b9d1805eb66f0f52a37d2b79c5fc/pkg/epp/framework/interface/requesthandling/types.go#L132-L136 I think we should scope this PR for only handling array of strings frist given handling array of strings already touches a lot of places.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
zetxqx: Agree on this is error-prone, but the above switch case is only for request that already have token input it should be only for `Generate` `Completion` and `Embeddings`.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
```diff
@@ -181,3 +177,27 @@ func CacheSaltFromBody(body *fwkrh.InferenceRequestBody) string {
 		return ""
 	}
 }
+
+// renderCompletions tokenizes a completions prompt. When the prompt is a
+// string array, each element is rendered independently and PerPromptTokens
+// is populated so downstream consumers can score per-prompt.
+func (b renderBackend) renderCompletions(ctx context.Context, body *fwkrh.InferenceRequestBody) (*fwkrh.TokenizedPrompt, error) {
+	prompt := body.Completions.Prompt
+	if len(prompt.Strings) <= 1 {
```
zetxqx: should this just be `if len(prompt.Raw) > 0` ? Ref: https://github.com/llm-d/llm-d-router/blob/c45f49d00401b9d1805eb66f0f52a37d2b79c5fc/pkg/epp/framework/interface/requesthandling/types.go#L132-L136

zetxqx: or do we want to completely deprecate the prompt.Raw and using strings. I think we can leave it for now to make this PR simple, other wise we need to remove the `Prompt.Raw` field and do a lot of additional changes.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/blockkeys.go
```diff
@@ -31,29 +32,58 @@ type kvCacheIndexer interface {
...
+	if tp == nil || len(tp.PerPromptTokens) == 0 {
 		return nil, nil
 	}
+
+	var result [][]kvblock.BlockHash
+	for _, tokens := range tp.PerPromptTokens {
+		if len(tokens) == 0 {
+			continue
+		}
+		// MM features apply only to single-prompt requests (chat); multi-prompt
```
liu-cong, line 58: How is this guaranteed?

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
albertoperdomo2: Exactly @zetxqx - I checked and even though it is a small change (<50 loc + tests) I think it might be out of the scope of this PR and should go in a new PR. I have opened #1573

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
albertoperdomo2: @zetxqx `len(prompt.Strings) <= 1` covers both Raw (where Strings is nil, so len is 0) and a single-element Strings slice (both go through `PlainText()` -> single Render call). Using `len(prompt.Raw) > 0` would "miss" the single-element Strings case, sending it through the multi-string loop. That still works (one iteration, same result), but the `<= 1` check keeps the intent clearer (it answers the question: do we need per-element rendering?)

albertoperdomo2: and yeah, I agree on leaving Raw as it is for now. Collapsing it into a length-1 Strings slice would clean things up and it looks like the right call, but keeps increasing the scope of this PR. I'll create an issue for it, or just submit a follow up PR with the changes.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/blockkeys.go
albertoperdomo2: It's a protocol level invariant: only `/v1/completions` produces multi-prompt (`PerPromptTokens` with len > 1), and the `completions` protocol doesn't support multimodal content, so `MultiModalFeatures` is always nil when `len(PerPromptTokens) > 1`. That said, nothing in the code enforces this, it relies on the parsers not populating both.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/blockkeys.go
```diff
@@ -31,29 +32,58 @@ type kvCacheIndexer interface {
...
+		// completions never carry multimodal content.
+		var mmf []fwkrh.MultiModalFeature
+		if len(tp.PerPromptTokens) == 1 {
```
albertoperdomo2, line 61: @liu-cong the fix/guarantee could be changing `len(tp.PerPromptTokens) == 1` to `i == 0`, because if offsets computed during tokenization point to positions in the first prompt's tokens, applying MM to the prompt whose tokens the offsets reference is correct regardless of how many prompts exist.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
ahg-g: make sense for now

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
ahg-g: I don't think this is correct. We should make a single Render call using `body.Payload` if it is not nil and the payload is a map even in the case where there is more than one String. So we do multiple calls only in the case where `body.Payload` is nil.

ahg-g: in fact, we shouldn't need to do multiple calls at all, in the case where `body.Payload` is nil, we can create a minimal PayloadMap with multiple strings in the "prompt", right?

albertoperdomo2: @ahg-g we could definitely build `PayloadMap{"prompt": ["alpha", "beta"]}` for a single call, but `Render` still returns a flat `[]uint32`, so there is no per-prompt boundary info in the return value. Without knowing where each prompt's tokens end, we can't populate `PerPromptTokens` correctly, which the precise prefix cache scorer needs for independent per-prompt scoring.

Something like this [llm-d-kv-cache/pull/538](https://github.com/llm-d/llm-d-kv-cache/pull/538) would help to solve this cleanly.

ahg-g: I think it is best if we handle it properly here and not make assumptions about render's behavior, which as you pointed may (should) change.

ahg-g: Basically this logic should move to `completionsPayload`, while `renderCompletions` shouldn't be changed, wdyt?

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/blockkeys.go
ahg-g: I think the issue is in the fact that `MultiModalFeature` is singleton, we should probably make it per token array, otherwise having the logic here making assumptions about how the upstream plugins work is not ideal. But that is a follow up, do you mind opening an issue to track this?

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
albertoperdomo2: Sounds good to me. So the idea is that `completionsPayload` builds the right payload (original `body.Payload` if available, otherwise a `PayloadMap` with the full prompt, string or array) and `renderCompletions` remains just one `Render` call always, right?

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/blockkeys.go
albertoperdomo2: Done in #1584 👍🏽

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/blockkeys.go
vMaroon: This is generally a legacy API that should die out eventually, from experience, while correction enablement is one thing (what this PR achieves) - a first class support for this API should wire through live batching or splitting requests at a coordinator+ level. Any one endpoint that gets bombarded with a heavy multi-prompt completions request will choke.

I hate that enablement has a deep and wide blast radius, but we can consider options and refactor later. It is possible that prefix-cache optimizations for this scenario are not effective, but don't have enough data.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
ahg-g: Correct.

albertoperdomo2: @ahg-g I updated `completionsPayload` (haven't pushed yet) to build the right payload shape: pass through `body.Payload` when available, otherwise a `PayloadMap` with the full prompt (string or string array). `renderCompletions` is now always a single `Render` call, with no per-string loop.

Just wanted to mention the following before pushing the change (which is small and scoped): In this case per-prompt scoring in the render path is still flat. Since `Render` returns a single `[]uint32`, multi-string completions land as one `PerPromptTokens` entry. vLLM's `/v1/completions/render` does return a `[]renderResponse` (one per prompt string), but [postCompletionsRender](https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go#L137) only reads `resp[0]`. Parsing all response elements to populate `PerPromptTokens` per-prompt would require changing the `Render` signature (or adding a new method) to return `[][]uint32`.

ahg-g: > but [postCompletionsRender](https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go#L137) only reads resp[0]. Parsing all response elements to populate PerPromptTokens per-prompt would require changing the Render signature (or adding a new method) to return [][]uint32.

lets create an issue and fix this in a follow up PR?

ahg-g: unless the change to [postCompletionsRender](https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go#L137) is also limited, we can do it in this PR as well.

ahg-g: I will leave it to you to decide, but I think we need to get the above fix for this release.

albertoperdomo2: Given the scope of the change, I have just included it in this PR. 

The only additional change on top of this is the offset return value. Due to the change to return `[][]uint32` by `Render`, the current offset return value ended up being semantically ambiguous for multi-prompt (`[]tokenizerTypes.Offset`, are offsets relative to which prompt?), so I changed it to `[][]tokenizerTypes.Offset`.

ahg-g (approved):

ahg-g:
@liu-cong @vMaroon @albertoperdomo2 we should run some benchmarks to ensure this PR does not cause a regression somehow.

PR #879, opened 2026-04-24
Fix: dispatch endpoint events when no PollingDataSource is configured

### LINKED ISSUE #876: [bug] data-layer endpoint events don't dispatch without a `PollingDataSource`
`Runtime.NewEndpoint` short-circuits when no `PollingDataSource` is
configured, returning before reaching `dispatchEndpointEvent`:

https://github.com/llm-d/llm-d-inference-scheduler/blob/ca32e657603d03ba4f8880fccf89a9d171a834b6/pkg/epp/datalayer/runtime.go#L180

```go
if len(pollers) == 0 {
    logger.Info("No polling sources configured, creating endpoint without collector")
    return fwkdl.NewEndpoint(endpointMetadata, nil)   // ← returns BEFORE dispatchEndpointEvent
}
// ... only here does dispatchEndpointEvent run
```

Effect: a config that wires endpoint-notification-source and an EndpointExtractor but no PollingDataSource looks correctly configured and silently never fires EndpointEvents. The precise-prefix-cache-scorer hit this in #862.

Compounding: the config loader's `ensureDataLayer` skips auto-injecting the default metrics source when any explicit dataLayer: block is present https://github.com/llm-d/llm-d-inference-scheduler/blob/ca32e657603d03ba4f8880fccf89a9d171a834b6/pkg/epp/config/loader/defaults.go#L300 
so wiring just the endpoint source actively disables what would otherwise have been auto-provided.

Expected: endpoint-only data-layer wiring should work standalone - endpoint add/delete events should dispatch regardless of whether any polling source is configured.

elevran:
**What type of PR is this?**
/kind bug

`Runtime.NewEndpoint` returned early before `dispatchEndpointEvent` when no `PollingDataSource` was configured, so a config with an `endpoint-notification-source` and an `EndpointExtractor` but no polling source silently never fired endpoint add/delete events. Added test and added notification call site.

Fixes #876.

cc @vMaroon 

```release-note
NONE
```

### DIFF
--- a/pkg/epp/datalayer/runtime.go
+++ b/pkg/epp/datalayer/runtime.go
@@ -179,7 +179,9 @@ func (r *Runtime) NewEndpoint(ctx context.Context, endpointMetadata *fwkdl.Endpo
 
 	if len(pollers) == 0 {
 		logger.Info("No polling sources configured, creating endpoint without collector")
-		return fwkdl.NewEndpoint(endpointMetadata, nil)
+		endpoint := fwkdl.NewEndpoint(endpointMetadata, nil)
+		r.dispatchEndpointEvent(ctx, logger, fwkdl.EndpointEvent{Type: fwkdl.EventAddOrUpdate, Endpoint: endpoint})
+		return endpoint
 	}
 
 	extractors := make(map[string][]fwkdl.Extractor, len(pollers))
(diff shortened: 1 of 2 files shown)


hexfusion (commented):
LGTM

Issue #833, opened 2026-04-16
Evaluate possibility of combining disagg_profile_handler and disagg_headers_handler


roytman:
For historical reasons, we have two separate handlers:`disagg_profile_handler` and `disagg_headers_handler`.

Since they are always used together, we should explore combining them into a single handler. This would improve code maintainability and, more importantly, simplify the configuration YAML files.

ahg-g:
I strongly support this as it simplifies the config experience for users.

noalimoy:
Hi @roytman ,

I'm working on this issue, and before I open a PR:

the merge is ready — `Handler` now implements both `ProfileHandler` and `PreRequest`, and configs no longer need a separate `disagg-headers-handler` entry.

one constraint I found: the legacy type strings (`disagg-headers-handler`, `prefill-header-handler`) can't simply point to `HandlerFactory`, because it now returns a `ProfileHandler`. 

configs that use `pd-profile-handler` with `prefill-header-handler` would fail with "multiple profile handlers found" — since `pd-profile-handler` depends on a separate PreRequest plugin for routing headers.

What's the preferred way to handle this?

roytman:
Hi @noalimoy, can you create a draft so I'll be able to see your code. 

I don't think we will face the `multiple profile handlers found` issue, because users with the deprecated `pd-profile-handler` and `prefill-header-handler` will continue using them.

noalimoy:
Hi @roytman,

Draft PR is up: #905

Since `prefill-header-handler` was previously registered with `HeadersHandlerFactory` — which was deleted as part of the merge — pointing it to `HandlerFactory` would introduce a second `ProfileHandler` and break `pd-profile-handler` configs with `multiple profile handlers found`.

To avoid this, the legacy type strings are registered as **PreRequest-only** factories, so they preserve the original behavior without conflicting.

Let me know if this approach works for you or if you'd prefer a different direction.

PR #880, opened 2026-04-24
refactor: introduce ExtractorBase to remove forced no-op Extract from event-driven extractors

### LINKED ISSUE #877: `EndpointExtractor` implementers shouldn't have to implement an empty `Extract`
`EndpointExtractor`
(https://github.com/llm-d/llm-d-inference-scheduler/blob/ca32e657603d03ba4f8880fccf89a9d171a834b6/pkg/epp/framework/interface/datalayer/plugin.go#L146)
extends `Extractor`
(https://github.com/llm-d/llm-d-inference-scheduler/blob/ca32e657603d03ba4f8880fccf89a9d171a834b6/pkg/epp/framework/interface/datalayer/plugin.go#L60),
which requires an `Extract(ctx, data, ep)` method. The Runtime never calls `Extract` on endpoint extractors, it dispatches via `ExtractEndpoint` instead. Yet every `EndpointExtractor` implementation must still provide a
no-op `Extract` to satisfy the interface. Example from #862:
```go
// Extract is the base Extractor entrypoint and is unused for endpoint
// extractors — the Runtime calls ExtractEndpoint instead.
func (s *Scorer) Extract(_ context.Context, _ any, _ fwkdl.Endpoint) error {
    return nil
}

### PR DESCRIPTION
**Pull request by elevran**
**What type of PR is this?**
/kind cleanup

`EndpointExtractor` and `NotificationExtractor` both embedded `Extractor`, which requires every implementor to provide `Extract(ctx, data, ep)`, which the Runtime never calls on these types. 

Introduce `ExtractorBase`, the minimal common base for all extractor variants:
```go
type ExtractorBase interface {
    plugin.Plugin
    ExpectedInputType() reflect.Type
}
```

The poll-based `Extractor` keeps its existing shape, now embedding `ExtractorBase` instead of duplicating the base methods. `NotificationExtractor` and `EndpointExtractor` embed `ExtractorBase` instead of `Extractor`.
Removed unused Extract from real and mock plugins.

Fixes #877.

```release-note
NONE
```

### DIFF
--- a/pkg/epp/datalayer/collector.go
+++ b/pkg/epp/datalayer/collector.go
@@ -96,7 +96,7 @@ func NewCollector() *Collector {
 
 // Start initiates data source collection for the endpoint.
 // All sources must implement PollingDataSource. Validation is performed by the caller.
-func (c *Collector) Start(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint, pollers []fwkdl.PollingDataSource, extractors map[string][]fwkdl.Extractor) error {
+func (c *Collector) Start(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint, pollers []fwkdl.PollingDataSource, extractors map[string][]fwkdl.ExtractorBase) error {
 	// Validate sources slice is not empty
 	if len(pollers) == 0 {
 		return errors.New("cannot start collector with empty sources")
@@ -111,7 +111,7 @@ func (c *Collector) Start(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint,
 	return c.startCollection(ctx, ticker, ep, pollers, extractors)
 }
 
-func (c *Collector) startCollection(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint, pollers []fwkdl.PollingDataSource, extractors map[string][]fwkdl.Extractor) error {
+func (c *Collector) startCollection(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint, pollers []fwkdl.PollingDataSource, extractors map[string][]fwkdl.ExtractorBase) error {
 	var ready chan struct{}
 	started := false
 
@@ -121,7 +121,7 @@ func (c *Collector) startCollection(ctx context.Context, ticker Ticker, ep fwkdl
 		started = true
 		ready = make(chan struct{})
 
-		go func(endpoint fwkdl.Endpoint, sources []fwkdl.PollingDataSource, exts map[string][]fwkdl.Extractor) {
+		go func(endpoint fwkdl.Endpoint, sources []fwkdl.PollingDataSource, exts map[string][]fwkdl.ExtractorBase) {
 			logger.V(logging.DEFAULT).Info("starting collection")
 
 			defer func() {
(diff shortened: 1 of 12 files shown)


pkg/epp/datalayer/runtime.go
```diff
@@ -133,7 +133,7 @@ func (r *Runtime) Start(ctx context.Context, mgr ctrl.Manager) error {
 
 		var extractors []fwkdl.NotificationExtractor
 		if rawExts, ok := r.sourceExtractors.Load(srcName); ok {
-			raw := rawExts.([]fwkdl.Extractor)
+			raw := rawExts.([]fwkdl.ExtractorBase)
 			extractors = make([]fwkdl.NotificationExtractor, len(raw))
 			for i, e := range raw {
 				extractors[i] = e.(fwkdl.NotificationExtractor)
```
hexfusion, line 139: i know this isn't part of you change but is that inner assertion safe?

elevran: `Configure()` validates the full compatibility between the source and its extractrors, including interface assertions.
Start is called after the validation happened so I believe it is safe.

pkg/epp/framework/interface/datalayer/plugin.go
```diff
@@ -56,11 +57,20 @@ type PollingDataSource interface {
...
+type ExtractorBase interface {
 	plugin.Plugin
 	// ExpectedInputType defines the type expected by the extractor.
 	ExpectedInputType() reflect.Type
+}
+
+// Extractor transforms raw data from a PollingDataSource into structured
+// attributes. Used only with poll-based sources; the Runtime calls Extract
+// on each tick with the data returned by the source's Poll method.
+type Extractor interface {
```
hexfusion, line 72: I looked at this for a while and what was sort of burning in my head is what is an extractor that doesn't Extract. The problem IMO is ExtractEndpoint is really just Extract right? What and how you are extracting is implementation details. 

What about this, it covers the usage patterns while being extensible

```go
 type Extractor[T any] interface {
      plugin.Plugin
      Extract(ctx context.Context, input T, opts ...ExtractOption) error
  }
```

hexfusion: take a look https://github.com/llm-d/llm-d-inference-scheduler/pull/883

elevran: I think #883 could be a step in the right direction. Will review once it is ready.
How close do you think is #883 to review readiness? Asking to assess if we want to have this one merged for now or eliminiate it waiting for PR#883 to resolve all.

shmuelk: Please see my comment on PR #833 https://github.com/llm-d/llm-d-inference-scheduler/pull/883#discussion_r3143401386

I don't think that code is a good solution. Also Generics are a compile time only feature. They may not help here, where one is trying to do runtime binding.

pkg/epp/datalayer/collector.go
```diff
@@ -153,8 +153,10 @@ func (c *Collector) startCollection(ctx context.Context, ticker Ticker, ep fwkdl
 						if srcExtractors, ok := exts[tn.Name]; ok && data != nil {
 							for _, ext := range srcExtractors {
 								extKey := ext.TypedName().String()
-								extErr := ext.Extract(ctx, data, endpoint)
-								logErrorTransition(logger, c.lastExtractErrors, extKey, "extract", "extractor", extErr)
+								if poller, ok := ext.(fwkdl.Extractor); ok {
+									extErr := poller.Extract(ctx, data, endpoint)
+									logErrorTransition(logger, c.lastExtractErrors, extKey, "extract", "extractor", extErr)
+								}
```
shmuelk: Wouldn't it be better to create a map of arrays of fwkdl.Extractor up front in the beggining of `startCollector`?

pkg/epp/framework/interface/datalayer/plugin.go
hexfusion: the available plugins are known at compile time, why do we need to do runtime assertion?

shmuelk: The wiring happens at runtime via the config file.

A user can easily have a source that references a profile handler for example....

hexfusion: > The wiring happens at runtime via the config file.

both designs validate at config load the same runtime assertion. the difference is after: reflect kept doing untyped interface{} assertions on every tick because the data flow was never typed. typed variants pay the cost once at the boundary downstream is typed and compiler-checked.

> A user can easily have a source that references a profile handler for example....

that case is caught by ResolveSource one type assertion at config load returns source plugin %q does not implement DataSource.

PR #883, opened 2026-04-24
refactor(datalayer): type extractor dispatch; remove reflect machinery 

hexfusion:
**What type of PR is this?**

/kind cleanup

Reshapes the data-layer extractor surface and runtime around typed generics, dropping the reflect-based validation it replaced.                                              
                                                                   
- Extractor[T any] with typed per-variant storage on `Runtime`: `*pollingManager` /  `*notificationManager` / `*endpointManager` (generic `sourceManager[S, E]`, plus a small notification wrapper for GVK uniqueness).                                        
- Removes reflect-based validation (`OutputType`, `ExpectedInputType`, `validateInputTypeCompatible`) and per-tick type assertions.
- Unifies the extractor method to `Extract` (was `ExtractNotification` / `ExtractEndpoint`).                                              
- A typed `sourceVariant` enum drives variant dispatch; `findSourceByType` walks one `Runtime.variants` lookup table built once in `NewRuntime` (single source of truth no parallel hard-coded list to drift).                                                   
- Cross-variant SourceType collisions now surface as configuration errors when a pending extractor references the colliding type. (Exhaustive startup-time validation for consumer-less collisions is tracked as a TODO in Configure.)                                              
- Drops the dead `outputType reflect.Type` plumbing from `HTTPDataSource` and the orphaned `PrometheusMetricType` / `ModelsResponseType` values. 

Fixes # https://github.com/llm-d/llm-d-inference-scheduler/issues/1053

```release-note
NONE
```

### DIFF
--- a/pkg/epp/config/loader/configloader.go
+++ b/pkg/epp/config/loader/configloader.go
@@ -33,7 +33,6 @@ import (
 	"github.com/llm-d/llm-d-inference-scheduler/pkg/epp/config"
 	"github.com/llm-d/llm-d-inference-scheduler/pkg/epp/datalayer"
 	"github.com/llm-d/llm-d-inference-scheduler/pkg/epp/flowcontrol"
-	fwkdl "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/datalayer"
 	fwkfc "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/flowcontrol"
 	fwkplugin "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/plugin"
 	fwkrh "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/requesthandling"
@@ -310,22 +309,24 @@ func buildDataLayerConfig(rawDataConfig *configapi.DataLayerConfig, handle fwkpl
 	}
 
 	for _, source := range rawDataConfig.Sources {
-		if sourcePlugin, ok := handle.Plugin(source.PluginRef).(fwkdl.DataSource); ok {
-			sourceConfig := datalayer.DataSourceConfig{
-				Plugin:     sourcePlugin,
-				Extractors: []fwkdl.ExtractorBase{},
-			}
-			for _, extractor := range source.Extractors {
-				if extractorPlugin, ok := handle.Plugin(extractor.PluginRef).(fwkdl.ExtractorBase); ok {
-					sourceConfig.Extractors = append(sourceConfig.Extractors, extractorPlugin)
-				} else {
-					return nil, fmt.Errorf("the plugin %s is not a fwkdl.ExtractorBase", source.PluginRef)
-				}
-			}
-			cfg.Sources = append(cfg.Sources, sourceConfig)
-		} else {
-			return nil, fmt.Errorf("the plugin %s is not a fwkdl.DataSource", source.PluginRef)
+		src, err := datalayer.ResolveSource(handle, source.PluginRef)
+		if err != nil {
+			return nil, err
 		}
+		extractors := make([]fwkplugin.Plugin, 0, len(source.Extractors))
+		for _, e := range source.Extractors {
+			p := handle.Plugin(e.PluginRef)
+			if p == nil {
+				return nil, fmt.Errorf("extractor plugin %s not found (required by source %s)",
+					e.PluginRef, source.PluginRef)
+			}
+			extractors = append(extractors, p)
+		}
+
+		cfg.Sources = append(cfg.Sources, datalayer.DataSourceConfig{
+			Plugin:     src,
+			Extractors: extractors,
+		})
 	}
 	return &cfg, nil
 }
--- a/pkg/epp/datalayer/config.go
+++ b/pkg/epp/datalayer/config.go
@@ -23,23 +23,19 @@ import (
 	fwkplugin "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/plugin"
 )
 
-// Config defines the configuration of EPP data layer, as the set of DataSources
-// and Extractors defined on them.
+// Config holds the data layer's configured sources and their extractors.
 type Config struct {
 	Sources []DataSourceConfig
 }
 
-// DataSourceConfig defines the configuration of a specific DataSource.
-//
-// Extractors are held as plugin.Plugin and type-asserted at Configure time to
-// the variant interface matching Plugin (PollingExtractor / NotificationExtractor
-// / EndpointExtractor). Mismatches surface as config-load errors.
+// DataSourceConfig pairs a source plugin with its extractor plugins. Extractors
+// are typed at Configure time against Plugin's variant; mismatches error out.
 type DataSourceConfig struct {
 	Plugin     fwkdl.DataSource
 	Extractors []fwkplugin.Plugin
 }
 
-// ResolveSource looks up ref via handle and asserts the plugin implements DataSource.
+// ResolveSource looks up ref and asserts the plugin implements DataSource.
 func ResolveSource(handle fwkplugin.Handle, ref string) (fwkdl.DataSource, error) {
 	p := handle.Plugin(ref)
 	if p == nil {
@@ -52,9 +48,9 @@ func ResolveSource(handle fwkplugin.Handle, ref string) (fwkdl.DataSource, error
 	return src, nil
 }
 
-// assertAll type-asserts each plugin to T and returns the typed slice.
-// T is resolved at compile time at the call site; per-element conformance is
-// checked at runtime and surfaced as an error rather than a panic.
+// assertAll types each plugin to T. T is bound at compile time at the call
+// site; per-element conformance is a runtime check that returns an error
+// rather than panicking, because plugins arrive as the erased plugin.Plugin.
 func assertAll[T fwkplugin.Plugin](plugins []fwkplugin.Plugin, variant string) ([]T, error) {
 	out := make([]T, 0, len(plugins))
 	for _, p := range plugins {
--- a/pkg/epp/datalayer/config.go
+++ b/pkg/epp/datalayer/config.go
@@ -54,6 +54,9 @@ func ResolveSource(handle fwkplugin.Handle, ref string) (fwkdl.DataSource, error
 func assertAll[T fwkplugin.Plugin](plugins []fwkplugin.Plugin, variant string) ([]T, error) {
 	out := make([]T, 0, len(plugins))
 	for _, p := range plugins {
+		if p == nil {
+			return nil, fmt.Errorf("nil plugin in %s", variant)
+		}
 		e, ok := p.(T)
 		if !ok {
 			return nil, fmt.Errorf("plugin %s is not a %s", p.TypedName(), variant)
--- a/pkg/epp/datalayer/config.go
+++ b/pkg/epp/datalayer/config.go
@@ -17,19 +17,52 @@ limitations under the License.
 package datalayer
 
 import (
+	"fmt"
+
 	fwkdl "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/datalayer"
+	fwkplugin "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/plugin"
 )
 
 // Config defines the configuration of EPP data layer, as the set of DataSources
-// and Extractors defined on them. Both poll-based and event-driven (notification)
-// sources are stored in Sources. Differentiation by type of source is handled during
-// the set-up phase.
+// and Extractors defined on them.
 type Config struct {
-	Sources []DataSourceConfig // the data sources configured in the data layer
+	Sources []DataSourceConfig
 }
 
-// DataSourceConfig defines the configuration of a specific DataSource
+// DataSourceConfig defines the configuration of a specific DataSource.
+//
+// Extractors are held as plugin.Plugin and type-asserted at Configure time to
+// the variant interface matching Plugin (PollingExtractor / NotificationExtractor
+// / EndpointExtractor). Mismatches surface as config-load errors.
 type DataSourceConfig struct {
-	Plugin     fwkdl.DataSource      // the data source plugin instance
-	Extractors []fwkdl.ExtractorBase // extractors defined for the data source
+	Plugin     fwkdl.DataSource
+	Extractors []fwkplugin.Plugin
+}
+
+// ResolveSource looks up ref via handle and asserts the plugin implements DataSource.
+func ResolveSource(handle fwkplugin.Handle, ref string) (fwkdl.DataSource, error) {
+	p := handle.Plugin(ref)
+	if p == nil {
+		return nil, fmt.Errorf("source plugin %q not registered", ref)
+	}
+	src, ok := p.(fwkdl.DataSource)
+	if !ok {
+		return nil, fmt.Errorf("source plugin %q does not implement DataSource", ref)
+	}
+	return src, nil
+}
+
+// assertAll type-asserts each plugin to T and returns the typed slice.
+// T is resolved at compile time at the call site; per-element conformance is
+// checked at runtime and surfaced as an error rather than a panic.
+func assertAll[T fwkplugin.Plugin](plugins []fwkplugin.Plugin, variant string) ([]T, error) {
+	out := make([]T, 0, len(plugins))
+	for _, p := range plugins {
+		e, ok := p.(T)
+		if !ok {
+			return nil, fmt.Errorf("plugin %s is not a %s", p.TypedName(), variant)
+		}
+		out = append(out, e)
+	}
+	return out, nil
 }
--- a/pkg/epp/datalayer/manager.go
+++ b/pkg/epp/datalayer/manager.go
@@ -2,6 +2,7 @@ package datalayer
 
 import (
 	"fmt"
+	"sort"
 	"sync"
 
 	fwkdl "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/datalayer"
@@ -27,14 +28,20 @@ func newSourceManager[S fwkdl.DataSource, E fwkplugin.Plugin](name string) *sour
 	}
 }
 
-// Register installs src and (if non-empty) exts under src's name.
-func (m *sourceManager[S, E]) Register(src S, exts []E) {
+// Register installs src and (if non-empty) exts under src's name. Returns an
+// error if a source with the same name is already registered.
+func (m *sourceManager[S, E]) Register(src S, exts []E) error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
-	m.sources[src.TypedName().Name] = src
-	if len(exts) > 0 {
-		m.extractors[src.TypedName().Name] = exts
+	srcName := src.TypedName().Name
+	if _, exists := m.sources[srcName]; exists {
+		return fmt.Errorf("duplicate %s source name %q", m.name, srcName)
 	}
+	m.sources[srcName] = src
+	if len(exts) > 0 {
+		m.extractors[srcName] = exts
+	}
+	return nil
 }
 
 // AppendExtractor appends ext to srcName's extractor list, deduping by type.
@@ -96,12 +103,19 @@ func (m *sourceManager[S, E]) IsEmpty() bool {
 	return len(m.sources) == 0
 }
 
-// FindByType returns the first source whose TypedName.Type matches sourceType.
+// FindByType returns the first matching source (in sorted-by-name order, so
+// behavior is stable across runs even when multiple sources share a Type).
 // match (if non-nil) is an additional filter applied to candidates.
 func (m *sourceManager[S, E]) FindByType(sourceType string, match func(S) bool) (string, S, bool) {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
-	for name, src := range m.sources {
+	names := make([]string, 0, len(m.sources))
+	for n := range m.sources {
+		names = append(names, n)
+	}
+	sort.Strings(names)
+	for _, name := range names {
+		src := m.sources[name]
 		if src.TypedName().Type != sourceType {
 			continue
 		}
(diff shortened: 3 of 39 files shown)


elevran:
@hexfusion this sounds like a good idea. Ping when ready for review?

shmuelk:
I don't think this is such a good idea. It makes the configuration code tied tightly to the various types of Sources and Extractors. At the end of the day some piece of code has to validate that the user didn't wire things up incorrectly.

elevran:
I might have misunderstood, but I think the idea is to have templated Go code handle the validation instead of using `reflect` and type assertions. We'll need to see once the PR is ready - it should provide better guarantees with less handwritten code.

pkg/epp/config/loader/configloader.go
```diff
@@ -288,22 +288,47 @@ func buildDataLayerConfig(rawDataConfig *configapi.DataLayerConfig, handle fwkpl
 	}
 
 	for _, source := range rawDataConfig.Sources {
-		if sourcePlugin, ok := handle.Plugin(source.PluginRef).(fwkdl.DataSource); ok {
-			sourceConfig := datalayer.DataSourceConfig{
-				Plugin:     sourcePlugin,
-				Extractors: []fwkdl.Extractor{},
+		sourcePlugin, ok := handle.Plugin(source.PluginRef).(fwkdl.DataSource)
```
shmuelk: This should have been handled in the validation code.

pkg/epp/config/loader/configloader.go
```diff
@@ -288,22 +288,47 @@ func buildDataLayerConfig(rawDataConfig *configapi.DataLayerConfig, handle fwkpl
...
+		if !ok {
+			return nil, fmt.Errorf("the plugin %s is not a fwkdl.DataSource", source.PluginRef)
+		}
+
+		sourceConfig := datalayer.DataSourceConfig{Plugin: sourcePlugin}
+
+		switch sourcePlugin.(type) {
```
shmuelk: This code is fragile and will have to be updated every time a new source and/or extrator is added.

pkg/epp/config/loader/configloader.go
hexfusion: fixed via 8ceb7a4c8f8d94f99db836624fb02da459e8cad7

pkg/epp/config/loader/configloader.go
hexfusion: good callout, the prior approach used reflection to walk extractors and dispatch by interface tag, runtime checks, error paths only exercised in tests, and no help from the compiler when adding a variant. this change makes the variant set a typed contract. each variant is an interface in framework/interface/datalayer, dispatch is a switch on c.Plugin.(type) in ResolveExtractors, and the typed extractor slices live on DataSourceConfig so the Runtime consumes them without re-asserting.

now the generic assertAll[T] collapses the per-variant cast loop to one line, the tradeoff feels reasonable trading runtime reflection for compile-checked typed slices and load-time variant validation.

pkg/epp/datalayer/runtime.go
```diff
@@ -35,10 +34,13 @@ import (
...
-	notifiers        sync.Map // Map of k8s notification sources (key=source name, value=NotificationSource)
-	endpointSources  sync.Map // Map of endpoint sources (key=source name, value=EndpointSource)
-	sourceExtractors sync.Map // Map sources to extractors (key=source name. value=[]Extractor)
+	pollers         sync.Map // Map of polling sources (key=source name, value=PollingDataSource)
+	notifiers       sync.Map // Map of k8s notification sources (key=source name, value=NotificationSource)
+	endpointSources sync.Map // Map of endpoint sources (key=source name, value=EndpointSource)
+
+	pollingExtractors      sync.Map // Map of polling extractors (key=source name, value=[]PollingExtractor)
+	notificationExtractors sync.Map // Map of notification extractors (key=source name, value=[]NotificationExtractor)
+	endpointExtractors     sync.Map // Map of endpoint extractors (key=source name, value=[]EndpointExtractor)
```
hexfusion: note: looking into reducing sync.Map sprawl in followup

hexfusion:
> @hexfusion this sounds like a good idea. Ping when ready for review?

I think its in a reasonable spot @elevran

pkg/epp/framework/interface/datalayer/plugin.go
```diff
@@ -56,22 +40,60 @@ type PollingDataSource interface {
...
-	Extract(ctx context.Context, data any, ep Endpoint) error
+	Extract(ctx context.Context, input T, opts ...ExtractOption) error
+}
+
+// PollingExtractor is an Extractor paired with a PollingDataSource.
+//
+// TODO: parametrize PollingDataSource.Poll over T and drop the `any` here so
+// polling extractors are typed end-to-end.
+type PollingExtractor = Extractor[any]
+
```
elevran, line 64: `PollingExtractor = Extractor[any]` is a type alias, not a type constraint.

The alias provides no type safety over the old untyped interface. Callers still pass `any` and must internally type-assert. The TODO comment acknowledges this explicitly.

If the goal of the PR is to eliminate reflect-based dispatch, the polling path is not actually fixed, it just hides the problem. The current alias creates false confidence that the generic machinery is doing work here.

pkg/epp/framework/interface/datalayer/plugin.go
```diff
@@ -56,22 +40,60 @@ type PollingDataSource interface {
...
+// See PollingExtractor, EndpointExtractor, and NotificationExtractor for variants.
+type Extractor[T any] interface {
 	plugin.Plugin
-	// ExpectedInputType defines the type expected by the extractor.
-	ExpectedInputType() reflect.Type
-	// Extract transforms the raw data source output into a concrete structured
-	// attribute, stored on the given endpoint.
-	Extract(ctx context.Context, data any, ep Endpoint) error
+	Extract(ctx context.Context, input T, opts ...ExtractOption) error
+}
```
elevran, line 60: The variadic `opts ...ExtractOption` weakens the API contract relative to the old explicit `ep Endpoint` parameter defined only where used.
`WithEndpoint` is consumed only by polling extractors and `EndpointExtractor` and `NotificationExtractor` don't use it (the event carries its own payload). Nothing prevents a caller from accidentally passing `WithEndpoint` to the wrong variant. There is no compile-time rejection and no runtime warning; the option is silently ignored. A caller that routes an endpoint to a `NotificationExtractor` via `WithEndpoint` gets a silent no-op that will surface only in tests or production.

For polling extractors, `WithEndpoint` is always *required* and both `metrics.Extractor` and `models.ModelExtractor` return a runtime error if it is absent. Making a required parameter optional trades a compile-time guarantee for a runtime nil check. Consider keeping `ep Endpoint` explicit on `PollingExtractor` and excluding it from the event-driven variants, even if that means the three variants don't share an identical `Extract` signature.

pkg/epp/framework/interface/datalayer/plugin.go
```diff
@@ -56,22 +40,60 @@ type PollingDataSource interface {
...
-// ValidatingDataSource is an optional interface that DataSources can implement
-// to perform additional custom validation when adding extractors.
-type ValidatingDataSource interface {
-	// ValidateExtractor allows the DataSource to perform additional validation
-	// beyond the standard type compatibility checks. Return an error if validation fails.
-	ValidateExtractor(extractor Extractor) error
+// Validator is an optional interface that DataSources can implement to
+// perform additional custom validation when adding extractors.
+type Validator interface {
+	// Validate allows the DataSource to perform additional validation
```
elevran: `Validator.Validate(p plugin.Plugin)` re-opens the dispatch problem this PR removes from `runtime.go`.

The interface takes an untyped `plugin.Plugin`, which means every `DataSource` that implements `Validator` must re-implement its own type switch to distinguish between `PollingExtractor`, `NotificationExtractor`, and `EndpointExtractor`. This is the same dispatch problem that was centralised in `validateSourceExtractors` and is now removed.

The old `ValidatingDataSource.ValidateExtractor(extractor Extractor)` accepted the specific extractor interface — a typed call. The new `Validate(p plugin.Plugin)` is strictly weaker. If at least one `DataSource` in tree has a `Validate` implementation, it is also re-implementing the type switch.

The fix is to make the method variant-specific — either three separate optional interfaces (`PollingValidator`, `NotificationValidator`, `EndpointValidator`) or a generic `Validator[T]` bound to the extractor type.

pkg/epp/datalayer/config.go
```diff
@@ -17,19 +17,78 @@ limitations under the License.
...
 
-// DataSourceConfig defines the configuration of a specific DataSource
+// DataSourceConfig defines the configuration of a specific DataSource.
+//
+// Each source pairs with extractors of exactly one variant (polling,
+// notification, or endpoint) — the variant is determined by the source's
+// interface type. Exactly one of the *Extractors fields is populated by
+// ResolveExtractors; the Runtime reads the field matching the source's variant.
 type DataSourceConfig struct {
-	Plugin     fwkdl.DataSource  // the data source plugin instance
```
elevran: `DataSourceConfig` parallel-slice invariant is comment-enforced only.

The comment says "Exactly one slice is populated." Nothing in the type system enforces this. Code that constructs `DataSourceConfig` directly (outside `configloader.go`) can silently populate multiple slices. `runtime.go`'s `Configure()` will then run two separate validate-and-wire paths for the same source.

pkg/epp/datalayer/config.go
```diff
@@ -17,19 +17,78 @@ limitations under the License.
...
+	if !ok {
+		return nil, fmt.Errorf("source plugin %q does not implement DataSource", ref)
+	}
+	return src, nil
+}
+
+// ResolveExtractors routes plugins into the field matching this source's
+// variant, type-asserting each to the variant's extractor interface.
+func (c *DataSourceConfig) ResolveExtractors(plugins []fwkplugin.Plugin) error {
+	var err error
```
elevran: The `switch c.Plugin.(type)` in `ResolveExtractors` is first-match-only, which is a silent correctness hazard for future sources.

A `DataSource` that implements multiple types (e.g., both `NotificationSource` and `EndpointSource`  - structurally valid given the interfaces) will hit only the first matching `case` and the second variant's extractors are silently dropped without error. This may be intentional today (no such source exists yet), but if a future source crosses interface boundaries this will wire only the first-matched extractor type.

An explicit guard would make the intent clear.

pkg/epp/datalayer/config.go
```diff
@@ -17,19 +17,78 @@ limitations under the License.
...
+			c.Plugin.TypedName())
+	}
+	return err
+}
+
+// assertAll type-asserts each plugin to the variant interface E and returns
+// the typed slice. Returns an error naming the first plugin that doesn't
+// implement T.
+func assertAll[T fwkplugin.Plugin](plugins []fwkplugin.Plugin, variant string) ([]T, error) {
+	out := make([]T, 0, len(plugins))
```
elevran, line 55: `assertAll[T]` is clean and avoids per-variant code repetition, but the safety it provides is **runtime**, **not compile-time**: the input is `[]fwkplugin.Plugin` (an erased interface), `p.(T)` is a runtime type assertion, and a mismatched plugin produces an error at load time rather than a compiler rejection. Worth being explicit that the generic parameter is resolved at compile time at the *call site* while the assertion itself executes at runtime.
Callers expecting compile-time safety from the generic will be surprised if this still panics at config load.

pkg/epp/datalayer/runtime.go
```diff
@@ -77,48 +79,127 @@ func (r *Runtime) Configure(cfg *Config, enableNewMetrics bool, disallowedExtrac
...
+			"endpoint", len(srcCfg.EndpointExtractors))
+	}
+
+	logger.Info("Datalayer runtime configured",
+		"pollers", pollersCount, "notifiers", notifiersCount, "endpointSources", endpointSourcesCount)
+	return nil
+}
+
+func (r *Runtime) validatePolling(src fwkdl.PollingDataSource, exts []fwkdl.PollingExtractor, disallowedType string) error {
+	validator, hasValidator := src.(fwkdl.Validator)
```
elevran: `validatePolling` and `validateEndpoint` are structurally identical — the disallowed-type check and `Validator.Validate` dispatch are copy-pasted verbatim, differing only in the extractor slice type. `validateNotification` adds only the GVK check on top of the same body. A generic helper would collapse the shared logic:

```go
func validateExtractors[E fwkplugin.Plugin](
    src fwkdl.DataSource, exts []E, disallowedType string,
) error {
    validator, hasValidator := src.(fwkdl.Validator)
    for _, ext := range exts {
        if disallowedType != "" && ext.TypedName().Type == disallowedType {
            return fmt.Errorf("disallowed Extractor %s configured for source %s",
                ext.TypedName(), src.TypedName())
        }
        if hasValidator {
            if err := validator.Validate(ext); err != nil {
                return fmt.Errorf("extractor %s failed validation for datasource %s: %w",
                    ext.TypedName(), src.TypedName(), err)
            }
        }
    }
    return nil
}
```

`validateNotification` calls this helper then appends its GVK check.

pkg/epp/config/loader/validation.go
```diff
@@ -40,6 +42,20 @@ func validateConfig(cfg *configapi.EndpointPickerConfig) error {
...
 
+// validateDataLayer checks that every source plugin ref resolves and implements
+// DataSource, so buildDataLayerConfig can construct sources without re-checking.
+func validateDataLayer(cfg *configapi.EndpointPickerConfig, handle fwkplugin.Handle) error {
+	if cfg.DataLayer == nil {
+		return nil
+	}
+	for _, source := range cfg.DataLayer.Sources {
+		if _, err := datalayer.ResolveSource(handle, source.PluginRef); err != nil {
+			return err
```
elevran: `validateDataLayer` calls `ResolveSource` for each source and discards the result. `buildDataLayerConfig` immediately calls `ResolveSource` again for the same set. Every source is looked up and type-asserted twice with no benefit since the validation pass does not cache its results, and errors from the build phase are possible even after validation passes (e.g., extractor resolution failure).

Consider either removing the pre-pass and letting `buildDataLayerConfig` be the single authority on errors, or have `validateDataLayer` return the resolved sources so the build path can reuse them.

elevran (commented):
The reflect-machinery removal is a genuine improvement — `OutputType()`, `ExtractorType()`, `ExpectedInputType()`, and `validateInputTypeCompatible`/`validateExtractorCompatible` were complex and potentially error-prone, and the hot-dispatch paths are noticeably cleaner without them. A few design issues below.

**On the generics premise**

The PR claims compile-time type safety in place of reflect-based validation.
- That claim is accurate for *code-registered* plugins - a concrete `Extract` implementation is type-checked against its variant's `T`, and the dispatch loops make typed calls without per-tick assertions. 
- the config-file binding path remains fully runtime checked

```
YAML PluginRef → handle.Plugin(ref) → p.(fwkdl.DataSource)   // runtime
               → switch c.Plugin.(type)                       // runtime
               → assertAll[T](plugins, variant) ≡ p.(T)      // runtime
```

What the PR delivers at the config boundary is *earlier* error detection (load-time vs. first-dispatch), which is valuable — but it is not compile-time safety. See inline comments for some of the examples.

pkg/epp/framework/interface/datalayer/plugin.go
hexfusion: right the TODO outlines this, resolving that gap could be a followup but felt like a larger change. open for debate.

pkg/epp/framework/interface/datalayer/plugin.go
hexfusion: > Nothing prevents a caller from accidentally passing WithEndpoint to the wrong variant

fair, validation could be added.

> For polling extractors, WithEndpoint is always required and both metrics.Extractor and models.ModelExtractor 

fair this needs a deeper look as to why this conditional requirement is needed at all.

pkg/epp/datalayer/config.go
hexfusion: updated the assertAll doc to spell out that T types the returned slice at compile time while per-element conformance is a runtime check surfaced as an error, not a panic.

pkg/epp/datalayer/config.go
hexfusion: i'll file that as a follow-up so this pr stays scoped

pkg/epp/framework/interface/datalayer/plugin.go
hexfusion: on the dispatch framing, a Validator impl only ever sees its own variant's extractors, so it's one typed handoff not a switch. you're right the API was strictly weaker though, so switched it to Validator[T plugin.Plugin] and the runtime now asserts Validator[E].

pkg/epp/datalayer/manager.go
```diff
@@ -0,0 +1,165 @@
+package datalayer
```
elevran, line 17: nit: every other `.go` file in the package starts with the Apache-2.0 license header which is is missing here.
I'm not a fan for including this baggage in every file, but for now please be consistent.

pkg/epp/datalayer/runtime.go
```diff
@@ -190,144 +119,187 @@ func (r *Runtime) Register(reg fwkdl.PendingRegistration) error {
...
 		}
+	}
+	return nil
+}
 
-		if bindErr := BindNotificationSource(ns, extractors, mgr); bindErr != nil {
-			err = fmt.Errorf("failed to bind notification source %s: %w", ns.TypedName(), bindErr)
-			return false
+// Stop halts all per-endpoint collectors.
+func (r *Runtime) Stop() error {
```
elevran: _Suggestion:_ `Stop()` always returns `nil` and the `error` return is never used. Either drop it or add a comment explaining it's reserved for future cleanup hooks.

pkg/epp/datalayer/runtime.go
```diff
@@ -190,144 +119,187 @@ func (r *Runtime) Register(reg fwkdl.PendingRegistration) error {
...
-		}
-		return true
-	})
-	if foundSrc != nil {
-		return foundName, foundSrc
+	return r.appendPendingExtractor(srcName, matchedSrc, pending.Extractor, disallowedType)
+}
+
+// findSourceByType walks all managers; gvkFilter narrows notification matches.
+func (r *Runtime) findSourceByType(sourceType string, gvkFilter *schema.GroupVersionKind) (string, fwkdl.DataSource) {
```
elevran: _Question_: The search order is now deterministic: polling > notification > endpoint. Is a cross-variant type collision is a configuration error? I think it seems likely, so consider returning an error instead of silently preferring polling. But perhaps that's being over defensive.

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache_extractor_test.go
```diff
@@ -58,11 +58,11 @@ func TestScorer_EndpointExtractor_InterfaceContract(t *testing.T) {
 	s := newExtractorScorer(true)
 	defer s.subscribersManager.Shutdown(ctx)
 
-	assert.Equal(t, fwkdl.EndpointEventReflectType, s.ExpectedInputType(),
-		"ExpectedInputType must report EndpointEvent for data-layer compatibility checks")
-
 	var _ fwkdl.EndpointExtractor = s
 	assert.True(t, reflect.TypeOf(s).Implements(reflect.TypeFor[fwkdl.EndpointExtractor]()))
```
elevran: nit: The `var _ fwkdl.EndpointExtractor = s` on the preceding line already enforces this contract at compile time so the `assert.True(reflect.TypeOf(s).Implements(...))` can be removed safely (it would have failed compilation).

elevran (commented):
Thanks @hexfusion 

The change leaves dead code in `HTTPDataSource`, the `outputType` field is no longer read and can be removed along with the `reflect` import. It can also be removed from `func NewHTTPDataSource(..., outputType reflect.Type)` 
Some additional nits below.

Otherwise can be approved once tests pass.
/lgtm

pkg/epp/datalayer/runtime.go
```diff
@@ -190,144 +156,190 @@ func (r *Runtime) Register(reg fwkdl.PendingRegistration) error {
...
+		matchedSrc     fwkdl.DataSource
+		matchedVariant sourceVariant
+	)
+	for _, l := range r.variants {
+		name, src, ok := l.find(sourceType, gvkFilter)
+		if !ok {
+			continue
+		}
+		if matchedSrc != nil {
+			return "", nil, fmt.Errorf("source type %q is registered across variants: %s (%s) and %s (%s)",
```
elevran: The PR description says *"Cross-variant sourceType collisions now surface as configuration errors"*, but this path (reachable only via `resolvePending`) has no test. Please add a test in `runtime_registrar_test.go` that registers the same `SourceType` across two different variant managers (e.g. one polling + one endpoint source) and asserts this error fires.

pkg/epp/datalayer/config.go
```diff
@@ -17,19 +17,48 @@ limitations under the License.
...
+	if !ok {
+		return nil, fmt.Errorf("source plugin %q does not implement DataSource", ref)
+	}
+	return src, nil
+}
+
+// assertAll types each plugin to T. T is bound at compile time at the call
+// site; per-element conformance is a runtime check that returns an error
+// rather than panicking, because plugins arrive as the erased plugin.Plugin.
+func assertAll[T fwkplugin.Plugin](plugins []fwkplugin.Plugin, variant string) ([]T, error) {
```
elevran, line 54: nit: `p.TypedName()` will panic if a nil element reaches this slice. Current callers guard against it, but a local check would protect against future caller.

pkg/epp/framework/interface/datalayer/plugin.go
```diff
@@ -57,31 +40,43 @@ type PollingDataSource interface {
...
+// See PollingExtractor, EndpointExtractor, and NotificationExtractor for variants.
+type Extractor[T any] interface {
 	plugin.Plugin
-	// ExpectedInputType defines the type expected by the extractor.
-	ExpectedInputType() reflect.Type
+	Extract(ctx context.Context, input T) error
+}
+
+// PollingInput pairs a poll result with the endpoint it belongs to.
+// TODO: parametrize PollingDataSource.Poll over T to drop this `any`.
```
elevran: nit: this TODO would be more discoverable on `PollingDataSource.Poll` rather than on the `PollingInput` struct.

elevran (commented):
/lgtm

- Can you please add a test for "Cross-variant sourceType collisions now surface as configuration errors"?
- the `Fixes #` in the PR description body needs an issue number or `NONE`.

pkg/epp/datalayer/config.go
hexfusion: feels a bit overly defensive but doesn't hurt

pkg/epp/datalayer/runtime.go
hexfusion: added todo, test, and updated description

hexfusion:
@elevran I appreciate the review and I think the exploration was a good learning experience but I dont think the churn here makes sense.

PR #891, opened 2026-04-26
Second step in refactoring the FlowControl component

shmuelk:
**What type of PR is this?**
/kind cleanup

This PR is the second step in a process of simplifying the Flow Control component's code base.

In particular this PR:
   1. Removes the number of shards from the FlowControl configuration
       **Note:** The number of shards was never externally exposed to the user.
   2. Implements suggested changes and simplification from the review of the previous PR
   3. Eliminates. various slices of shards that were in the code.
   4. Pre-allocates the priorityBandAccessor rather than constantly creating a new one.

**Note:** The previous PR mentioned above was an Inference-Gateway PR that was previously merged.

Refs #889

```release-note
NONE
```

### DIFF
--- a/pkg/epp/flowcontrol/benchmark/benchmark.go
+++ b/pkg/epp/flowcontrol/benchmark/benchmark.go
@@ -90,9 +90,6 @@ func init() {
 // egressConcurrencyLimit (L) defines the maximum capacity of the simulated pool.
 type egressConcurrencyLimit int64
 
-// shardCount (S) dictates the data parallelism of the Flow Control layer.
-type shardCount int
-
 // priorityLevels (P) dictates the number of priority bands.
 type priorityLevels int
 
@@ -107,7 +104,6 @@ type ingressConcurrency int
 // benchMatrix defines a single coordinate in the performance hypercube.
 type benchMatrix struct {
 	limit       egressConcurrencyLimit
-	shards      shardCount
 	priorities  priorityLevels
 	flows       flowCount
 	concurrency ingressConcurrency
@@ -115,8 +111,8 @@ type benchMatrix struct {
 
 // name returns a human-readable string representation of the matrix coordinate.
 func (m benchMatrix) name() string {
-	return fmt.Sprintf("L=%03d/S=%03d/P=%03d/F=%06d/W=%05d",
-		m.limit, m.shards, m.priorities, m.flows, m.concurrency)
+	return fmt.Sprintf("L=%03d/P=%03d/F=%06d/W=%05d",
+		m.limit, m.priorities, m.flows, m.concurrency)
 }
 
 // testDetector exposes an API to manually release downstream capacity during a test run.
@@ -195,13 +191,11 @@ func setupRegistry(
 	ctx context.Context,
 	b *testing.B,
 	handle plugin.Handle,
-	s shardCount,
 	p priorityLevels,
 ) contracts.FlowRegistry {
 	b.Helper()
 
 	cfgOpts := []registry.ConfigOption{
-		registry.WithInitialShardCount(int(s)),
 		registry.WithMaxBytes(0), // Capacity restricted strictly via concurrency (L).
 	}
 
@@ -234,7 +228,6 @@ func setupRegistry(
 func setupBenchmarkHarness(
 	ctx context.Context,
 	b *testing.B,
-	s shardCount,
 	p priorityLevels,
 	limit egressConcurrencyLimit,
 	customDetector testDetector,
@@ -255,7 +248,7 @@ func setupBenchmarkHarness(
 	}
 	handle.AddPlugin(registry.DefaultOrderingPolicyRef, oPolicy)
 
-	reg := setupRegistry(ctx, b, handle, s, p)
+	reg := setupRegistry(ctx, b, handle, p)
 
 	detector := customDetector
 	if detector == nil {
@@ -266,8 +259,7 @@ func setupBenchmarkHarness(
 
 	cfg := customCfg
 	if cfg == nil {
-		// Buffer size dynamically scales down with parallel shards to isolate queue sorting overhead.
-		bufferSize := max(2000/int(s), 10)
+		bufferSize := max(2000, 10)
 		cfg = &controller.Config{
 			DefaultRequestTTL:               5 * time.Minute,
 			ProcessorReconciliationInterval: 1 * time.Hour, // Effectively disabled
--- a/pkg/epp/flowcontrol/benchmark/benchmark.go
+++ b/pkg/epp/flowcontrol/benchmark/benchmark.go
@@ -259,12 +259,11 @@ func setupBenchmarkHarness(
 
 	cfg := customCfg
 	if cfg == nil {
-		bufferSize := max(2000, 10)
 		cfg = &controller.Config{
 			DefaultRequestTTL:               5 * time.Minute,
 			ProcessorReconciliationInterval: 1 * time.Hour, // Effectively disabled
 			ExpiryCleanupInterval:           1 * time.Hour, // Effectively disabled
-			EnqueueChannelBufferSize:        bufferSize,
+			EnqueueChannelBufferSize:        2000,
 		}
 	}
 
--- a/pkg/epp/flowcontrol/benchmark/benchmark_test.go
+++ b/pkg/epp/flowcontrol/benchmark/benchmark_test.go
@@ -125,8 +125,8 @@ func runMatrixCoordinate(b *testing.B, m benchMatrix) {
 
 		// Offset the starting index per thread to prevent identical striding over the array.
 		// Multiply by a prime to guarantee threads start at different offsets.
-		threadId := globalThreadID.Add(1)
-		localIdx := int(threadId) * 9973
+		threadID := globalThreadID.Add(1)
+		localIdx := int(threadID) * 9973
 
 		for pb.Next() {
 			localIdx++
--- a/pkg/epp/flowcontrol/benchmark/benchmark_test.go
+++ b/pkg/epp/flowcontrol/benchmark/benchmark_test.go
@@ -40,44 +40,29 @@ func BenchmarkFlowController_PerformanceMatrix(b *testing.B) {
 		b.Skip("skipping PerformanceMatrix in short mode")
 	}
 	limits := []egressConcurrencyLimit{0, 1, 100}
-	shards := []shardCount{1, 8}
 	priorities := []priorityLevels{1, 8}
-	flows := []flowCount{10, 50000}
-	concurrencies := []ingressConcurrency{10, 5000, 50000}
+	flows := []flowCount{10, 5000}
+	concurrencies := []ingressConcurrency{10, 5000}
 
 	for _, L := range limits {
-		for _, S := range shards {
-			for _, P := range priorities {
-				for _, F := range flows {
-					for _, W := range concurrencies {
-						// Skip illogical boundaries.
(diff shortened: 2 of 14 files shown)


pkg/epp/flowcontrol/benchmark/benchmark_test.go
```diff
@@ -40,52 +40,37 @@ func BenchmarkFlowController_PerformanceMatrix(b *testing.B) {
...
-	flows := []flowCount{10, 50000}
-	concurrencies := []ingressConcurrency{10, 5000, 50000}
+	flows := []flowCount{10, 5000}
+	concurrencies := []ingressConcurrency{10, 5000}
 
 	for _, L := range limits {
-		for _, S := range shards {
-			for _, P := range priorities {
-				for _, F := range flows {
-					for _, W := range concurrencies {
```
hexfusion, line 52: 🎉

pkg/epp/flowcontrol/registry/connection.go
```diff
@@ -31,13 +31,9 @@ type connection struct {
...
 
-// Shards returns a stable snapshot of accessors for all internal state shards.
-func (c *connection) ActiveShards() []contracts.RegistryShard {
-	// Return a copy to ensure the caller cannot modify the registry's internal slice.
-	shardsCopy := make([]contracts.RegistryShard, 1)
-	shardsCopy[0] = c.registry.shard
-
-	return shardsCopy
+// GetShard returns the shard this connection is pinned to.
+func (c *connection) GetShard() contracts.RegistryShard {
```
hexfusion, line 35: if we are only returning one item is shard the correct term?

pkg/epp/flowcontrol/contracts/registry.go
```diff
@@ -56,8 +56,8 @@ type FlowRegistryObserver interface {
 	// Stats returns a near-consistent snapshot globally aggregated statistics for the entire `FlowRegistry`.
 	Stats() AggregateStats
 
-	// ShardStats returns a near-consistent slice of statistics snapshots, one for each `RegistryShard`.
-	ShardStats() []ShardStats
+	// ShardStats returns a near-consistent statistics snapshot for the `RegistryShard`.
+	ShardStats() *ShardStats
```
hexfusion, line 60: since there are no shards really, can we remove this method and  `AggregateStats` -> `Stats`.

pkg/epp/flowcontrol/registry/connection.go
shmuelk: This PR is the second in a series of PRs that will eventually remove the term shard from the EPP.

This PR cleans up some stuff from the previous PR that took a step in the direction of having only one shard.

The next PR will remove the shard struct altogether.

pkg/epp/flowcontrol/contracts/registry.go
shmuelk: This PR is the second in a series of PRs that will eventually remove the term shard from the EPP.

This PR cleans up some stuff from the previous PR that took a step in the direction of having only one shard.

The next PR will remove the shard struct altogether.

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -456,22 +441,19 @@ func (fc *FlowController) selectDistributionCandidates(conn contracts.ActiveFlow
...
-	for _, c := range candidates {
-		if err := c.processor.Submit(item); err == nil {
-			return types.QueueOutcomeNotYetFinalized, nil
-		}
-		fc.logger.V(logutil.TRACE).Info("Processor busy during fast failover, trying next candidate",
-			"shardID", c.shardID, "requestID", reqID)
+	if err := candidate.processor.Submit(item); err == nil {
+		return types.QueueOutcomeNotYetFinalized, nil
 	}
+	fc.logger.V(logutil.TRACE).Info("Processor busy during fast failover, trying next candidate",
```
elevran: This log (lines 450–451) is a copy-paste remnant from the old `for _, c := range candidates` body. With a single candidate there is no "next candidate"; `Submit()` failing goes straight to `SubmitOrBlock` below. This line fires on every busy processor, emitting a misleading trace message unconditionally.

Delete lines 450–451. The correct log on line 454 is the one to keep.
Also: line 453 has a grammar error — `// processor are busy.` should be `// Processor is busy.`

pkg/epp/flowcontrol/benchmark/benchmark.go
```diff
@@ -266,8 +259,7 @@ func setupBenchmarkHarness(
 
 	cfg := customCfg
 	if cfg == nil {
-		// Buffer size dynamically scales down with parallel shards to isolate queue sorting overhead.
-		bufferSize := max(2000/int(s), 10)
+		bufferSize := max(2000, 10)
```
elevran: This was originally `max(2000/int(s), 10)` so the buffer size scaled down with shard count and the lower bound of 10 was reachable. Now that `s` is gone, `max(2000, 10)` is a constant expression and can be replaced with `bufferSize := 2000` (or a named constant) to remove the misleading hint that the lower bound is meaningful.

elevran (commented):
Reviewed for simplification and correctness. Two issues below couldn't be posted as inline diff comments because they're in unchanged context lines; they're called out here with exact file:line references.


`pkg/epp/flowcontrol/controller/controller.go:254` - Stale comment referencing removed `ActiveShards()` method
The comment reads:
> "We can safely retry within this loop using the same 'conn' object because `conn.ActiveShards()` provides a live view of the topology."

`ActiveShards()` was replaced by `GetShard()` in this PR. The comment now contradicts the implementation: `GetShard()` returns a **fixed** single shard, not a live topology view. The safety argument it makes is no longer valid (see next issue). Update or remove.

In ``pkg/epp/flowcontrol/controller/controller.go:278` - `ErrShardDraining` retry is now an infinite spin loop

The retry-on-draining logic at line 278 was safe with `ActiveShards()` because the registry would have already removed the draining shard from the live list,  so the next `continue` landed on a healthy shard.

With `GetShard()` the shard is **fixed** for the lifetime of the connection. If the shard drains after acceptance, every retry produces:
1. `tryDistribution` > `GetShard()` > same draining shard > accepted by processor
2. Processor finalizes the item with `ErrShardDraining`
3. `continue` back to step 1

The only exit is TTL expiry, which is a silent worst-case latency of `DefaultRequestTTL` instead of a fast failure. With a single shard, `ErrShardDraining` should break out of the loop and return a terminal error.

elevran:
This can be approved for my point if view. Holding lgtm and approval until @LukeAVanDrie has a chance to review

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -406,36 +388,23 @@ type candidate struct {
 	byteSize  uint64
 }
 
-// selectDistributionCandidates identifies all Active shards for the leased flow and ranks them by the current byte size
+// selectDistributionCandidate identifies all Active shards for the leased flow and ranks them by the current byte size
```
LukeAVanDrie, line 391: Since multi-shard sorting (JSQ-Bytes) was removed, `shardID` and `byteSize` are completely unused in `distributeRequest`. Remove them from struct candidate and eliminate the `ManagedQueue` lookup and `ByteSize()` calculation in `selectDistributionCandidate`.

E.g.,

```
type candidate struct {
	processor shardProcessor
}

func (fc *FlowController) selectDistributionCandidate(conn contracts.ActiveFlowConnection) (*candidate, error) {
	shard := conn.GetShard()
	if shard == nil {
		return nil, fmt.Errorf("%w for flow %s", errNoShards, conn.FlowKey())
	}
	worker := fc.getOrStartWorker(shard)
	return &candidate{processor: worker.processor}, nil
}
```

pkg/epp/flowcontrol/controller/controller.go
LukeAVanDrie, line 116: This is no longer necessary. Consider replacing `workers sync.Map` with a static `processor shardProcessor` field initialized once at startup, and remove` reconcileProcessors` entirely. If this is planned for the next PR that removes the shard struct altogether, feel free to defer.

pkg/epp/flowcontrol/registry/config.go
LukeAVanDrie, line 642: Since `totalShards` is always 1 in the new architecture, the division math and remainder distribution logic are redundant. Remove `partitionUint64` and simplify `partition()` to pass configured limits directly.

pkg/epp/flowcontrol/controller/controller.go
LukeAVanDrie, line 289:
```suggestion
var errNoShard = errors.New("no viable active shard available")
```

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -406,36 +388,23 @@ type candidate struct {
...
-		}
-		candidates = append(candidates, candidate{worker.processor, shard.ID(), mq.FlowQueueAccessor().ByteSize()})
-	}
-
-	if len(candidates) == 0 {
+	worker := fc.getOrStartWorker(shard)
+	mq, err := shard.ManagedQueue(conn.FlowKey())
+	if err != nil {
+		fc.logger.Error(err,
+			"Invariant violation. Failed to get ManagedQueue for a leased flow on an Active shard. Skipping shard.",
```
LukeAVanDrie, line 403: nit: Update the log message to reflect terminal failure since there are no other shards in a loop to fail over to:

```
fc.logger.Error(err, "Invariant violation. Failed to get ManagedQueue for a leased flow. Aborting request distribution.", "flowKey", conn.FlowKey())
```

pkg/epp/flowcontrol/controller/controller.go
LukeAVanDrie, line 391: nit: Update docstrings to accurately reflect single-shard submission and blocking fallback:

```
// selectDistributionCandidate identifies the Active shard for the leased flow.
func (fc *FlowController) selectDistributionCandidate(...)

...

// distributeRequest attempts to submit an item to the shard processor.
// If non-blocking submission fails, it performs a single blocking submission providing backpressure.
func (fc *FlowController) distributeRequest(...)
```

LukeAVanDrie (approved):
All my feedback is non-blocking as this may be better addressed in followup PRs after more code is simplified. Just calling out some areas for further cleanup, thanks!

LukeAVanDrie:
We also need to update package-level docstrings and README.md to describe the simplified single-shard supervisor/processor architecture.

Though, once the dynamic shard topology is completely removed, the distinction between the `FlowController` (supervisor) and `ShardProcessor` (worker) becomes purely artificial.

Currently, the `FlowController` acts as the synchronous entrypoint (handling ext_proc stream blocking, TTL/context cancellation enforcement, and metrics), while the `ShardProcessor` runs the asynchronous background dispatch loop. In a strict single-shard design, these responsibilities can and should be collapsed into a single unified engine, eliminating the supervisor layer entirely. Does this align with your approach to this refactoring?

If so, maybe its best to wait on these doc changes until everything is done.

LukeAVanDrie:
> This can be approved for my point if view. Holding lgtm and approval until @LukeAVanDrie has a chance to review

/lgtm on my end. Left some non-blocking feedback for things we will need to remember to clean up, but all can be deferred. It may be easier to just re-write the docs on the simplified end state rather than trying to keep a perfectly coherent file state between PRs, especially if these are landing soon.

elevran:
/lgtm
/approve

@LukeAVanDrie thanks for the review!
@shmuelk can you please incorporate Luke's suggestions in a follow up (as this is 2 of N, anyway...) 🙏

PR #894, opened 2026-04-26
refactor: align eviction plugin Type names with naming conventions

### LINKED ISSUE #893: Rename eviction filters to align with plugin naming convictions
1) Rename `eviction-sheddable-filter` to `sheddable-eviction-filter` in https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/epp/framework/plugins/flowcontrol/eviction/filtering/sheddable.go; all eviction filters should be suffixed with `eviction-filter` (helps distinguish from endpoint filters which we suffix with `filter`
2) Rename `eviction-priority-then-time-ordering` to `priority-then-time-eviction-order-policy` in https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/epp/framework/plugins/flowcontrol/eviction/ordering/priority_time.go; all eviction ordering plugins should be suffixed with `-eviction-order-policy`

jia-gao:
Fixes #893.

Renames the two eviction plugin Type strings to match the suffix conventions @ahg-g requested in the issue:

- Eviction filters use the `-eviction-filter` suffix (distinct from endpoint filters, which use `-filter`).
- Eviction ordering plugins use the `-eviction-order-policy` suffix.

Changes:
- `eviction-sheddable-filter` → `sheddable-eviction-filter`
- `eviction-priority-then-time-ordering` → `priority-then-time-eviction-order-policy`

Only the constant string values change; the Go identifiers (`SheddableFilterType`, `PriorityThenTimeOrderingType`) and the tests are unchanged since they reference the constants by name.

## Verification

- `go build ./...` clean
- `go vet ./...` clean
- `go test ./pkg/...` all pass
- Skipped: `test/e2e`, `test/sidecar/e2e`, `test/profiling/tokenizerbench` (require a live K8s cluster + running EPP)

### DIFF
--- a/pkg/epp/framework/plugins/flowcontrol/eviction/filtering/sheddable.go
+++ b/pkg/epp/framework/plugins/flowcontrol/eviction/filtering/sheddable.go
@@ -26,7 +26,7 @@ import (
 
 // SheddableFilterType only admits sheddable requests (priority < 0) into the eviction queue.
 // This uses the project-wide definition of sheddable from requtil.IsSheddable.
-const SheddableFilterType = "eviction-sheddable-filter"
+const SheddableFilterType = "sheddable-eviction-filter"
 
 func init() {
 	plugin.Register(SheddableFilterType, SheddableFilterFactory)
--- a/pkg/epp/framework/plugins/flowcontrol/eviction/ordering/priority_time.go
+++ b/pkg/epp/framework/plugins/flowcontrol/eviction/ordering/priority_time.go
@@ -25,7 +25,7 @@ import (
 
 // PriorityThenTimeOrderingType evicts the lowest priority request first,
 // breaking ties by newest dispatch time (least KV-cache investment).
-const PriorityThenTimeOrderingType = "eviction-priority-then-time-ordering"
+const PriorityThenTimeOrderingType = "priority-then-time-eviction-order-policy"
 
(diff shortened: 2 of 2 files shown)


pkg/epp/framework/plugins/flowcontrol/eviction/filtering/sheddable.go
```diff
@@ -26,7 +26,7 @@ import (
 
 // SheddableFilterType only admits sheddable requests (priority < 0) into the eviction queue.
 // This uses the project-wide definition of sheddable from requtil.IsSheddable.
-const SheddableFilterType = "eviction-sheddable-filter"
+const SheddableFilterType = "sheddable-eviction-filter"
```
hexfusion, line 29: nit: this makes sense but it also is breaking for anyone whose yaml declared the old string should we handle deprecation gracefully?

ahg-g: this is not used yet

ahg-g:
can you please rebase and have the commit signed, using `commit -s` pls

jia-gao:
Rebased onto latest `main` and signed off the commit (`eb8bef2`). Ready for re-review @ahg-g — thanks!

PR #898, opened 2026-04-27
fix: make eviction runtime a builtin, not a plugin

### LINKED ISSUE #818: Eviction runtime should not be modeled as a plugin, it should be builtin

The evictor subsystem is currently implemented as a plugin. This logic represents the runtime that executes the eviction logic and it should not be modeled as a plugin. 

Context: https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2737#discussion_r3066591083

**Why is this needed**:

This is a misuse of the pluggable framework.  We need to put this guidance in a framework's dev guide: Plugins represent entities that a user can enable/disable via the config api, and they only reside under framework/epp/plugins.

abdallahsamabd:
fix: make eviction runtime a builtin, not a plugin

Remove plugin.Plugin interface (TypedName) from RequestEvictor so it is
no longer modeled as a user-configurable plugin. Rename plugin.go to
request_evictor.go. Add developer guide documenting the plugin vs builtin
rule.

**What type of PR is this?**

/kind cleanup
/kind documentation

The `RequestEvictor` (the subsystem that tracks in-flight inference requests
and executes eviction when the system is overloaded) was incorrectly
implementing the `plugin.Plugin` interface via a `TypedName()` method
returning `{Type: "EvictionPlugin", Name: "eviction"}`. This made it appear
as a user-configurable plugin that could be enabled/disabled via YAML config,
when in reality it is core EPP infrastructure that must always be present.

This PR:

1. **Removes `TypedName()`** and the `plugin.Plugin` interface from
   `RequestEvictor`, making it a proper builtin component.
2. **Removes the compile-time interface assertions** (`var _`) that forced
   `RequestEvictor` to satisfy plugin-based `requestcontrol.PreRequest` and
   `requestcontrol.ResponseBodyProcessor` interfaces.
3. **Renames** `plugin.go` → `request_evictor.go` and `plugin_test.go` →
   `request_evictor_test.go` to eliminate naming confusion.
4. **Updates comments** in `eviction_registry.go` replacing "eviction plugin"
   with "RequestEvictor".
5. **Adds `docs/plugins-vs-builtins.md`** — a developer guide documenting the
   rule that plugins are user-selectable components under `framework/plugins/`,
   while builtins are always-on infrastructure wired directly by the EPP.

Note: eviction **policies** (`EvictionOrderingPolicy`, `EvictionFilterPolicy`)
remain as real plugins — they are user-selectable strategies. Only the
eviction **runtime** (the machinery that executes those policies) is changed
to a builtin.

No behavioral changes — all 21 eviction tests pass, full repo builds clean.

Fixes #818

NONE


docs/plugins-vs-builtins.md
```diff
@@ -0,0 +1,93 @@
...
+- A user **chooses** which plugins to activate in their scheduling profiles.
+- Multiple instances of the same plugin type may coexist with different names
+  and parameters.
+- Removing a plugin from config disables its behavior without breaking the EPP.
+
+## Builtins
+
+A **builtin** is a component that:
+
+- **Is always present** and wired directly by the EPP runner or request-control
```
hexfusion: nit: since core and plugins are compiled into the binary they are technically all present, is it more that core are always enabled, I guess that is covered in the next line.

abdallahsamabd: done, thanks

ahg-g:
/approve

Excellent!

PR #899, opened 2026-04-27
refactor(datalayer): simplify Collector lifecycle, add error metrics

hexfusion:
**What type of PR is this?**

/kind cleanup
/kind bug

Simplifies `datalayer.Collector` lifecycle and replaces in-binary error tracking with Prometheus counters.

- Drop stored `ctx`; goroutine captures it via closure. Mutex + cancel func replace `startOnce`/`stopOnce`/`done`.
- `Stop()` is fire-and-forget and idempotent; misuse logged. Shutdown flows through parent ctx cancellation.
- Per-call timeouts for both Poll and each Extract. Fixes a latent bug where Extract received an already-cancelled ctx.
- Bail on `ctx.Err()` between iterations.
- Remove `Runtime.Stop()`: unused.
- Replace error-state maps + transition log with `datalayer_poll_errors_total{source_type}` and `datalayer_extract_errors_total{source_type, extractor_type}`. Cardinality bounded by plugin Type.

Fixes #

**Release note**:
  
```release-note
  Add `datalayer_poll_errors_total` and `datalayer_extract_errors_total`
  Prometheus counters; remove transition-only error logging.
```

### DIFF
--- a/pkg/epp/datalayer/collector.go
+++ b/pkg/epp/datalayer/collector.go
@@ -41,8 +41,6 @@ const (
 	defaultCollectionTimeout = time.Second
 )
 
-var collectorLogger = log.Log.WithName("datalayer-collector")
-
 // Ticker implements a time source for periodic invocation.
 // The Ticker is passed in as parameter a Collector to allow control over time
 // progress in tests, ensuring tests are deterministic and fast.
@@ -69,15 +67,20 @@ func (t *TimeTicker) Channel() <-chan time.Time {
 }
 
 // Collector runs data collection for a single endpoint.
+//
+// Lifecycle contract: any in-flight write the collection goroutine performs
+// against the endpoint completes before Stop returns. Callers may therefore
+// mutate or release endpoint state immediately after Stop returns without
+// racing the collection goroutine.
 type Collector struct {
-	mu      sync.Mutex
-	cancel  context.CancelFunc
-	stopped bool
+	mu     sync.Mutex
+	cancel context.CancelFunc
+	done   chan struct{}
 }
 
 // NewCollector returns a new collector.
 func NewCollector() *Collector {
-	return &Collector{}
+	return &Collector{done: make(chan struct{})}
 }
 
 // Start launches the collection goroutine.
@@ -115,27 +118,22 @@ func (c *Collector) Start(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint,
 	return nil
 }
 
-// Stop signals the collection goroutine to exit. Idempotent.
+// Stop cancels the collection goroutine and blocks until it has exited. Idempotent.
 func (c *Collector) Stop() {
-	var msg string
 	c.mu.Lock()
-	switch {
-	case c.cancel == nil:
-		msg = "Stop called before successful Start"
-	case c.stopped:
-		msg = "Stop called more than once"
-	default:
-		c.stopped = true
-		c.cancel()
-	}
+	cancel := c.cancel
 	c.mu.Unlock()
-	if msg != "" {
-		collectorLogger.Info(msg)
+	if cancel != nil {
+		cancel()
+		<-c.done
 	}
 }
 
 func (c *Collector) run(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint, pollers []fwkdl.PollingDataSource, extractors map[string][]fwkdl.Extractor) {
-	defer ticker.Stop()
+	defer func() {
+		close(c.done)
+		ticker.Stop()
+	}()
 	logger := log.FromContext(ctx).WithValues("endpoint", ep.GetMetadata().GetIPAddress())
 
 	for {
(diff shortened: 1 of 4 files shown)


pkg/epp/datalayer/runtime.go
```diff
@@ -149,18 +149,6 @@ func (r *Runtime) Start(ctx context.Context, mgr ctrl.Manager) error {
 	return err
 }
 
-// Stop is called to terminate the Runtime's data collection. It terminates all
-// go routines used for polling data sources.
-func (r *Runtime) Stop() error {
```
hexfusion, line 299: note: unused, per-endpoint shutdown runs via ReleaseEndpoint process shutdown runs via parent ctx cancellation to each collector's ctx.

pkg/epp/datalayer/collector.go
```diff
@@ -65,153 +68,107 @@ func (t *TimeTicker) Channel() <-chan time.Time {
...
-		lastPollErrors:    make(map[string]error),
-		lastExtractErrors: make(map[string]error),
-	}
+	return &Collector{}
 }
 
-// Start initiates data source collection for the endpoint.
-// All sources must implement PollingDataSource. Validation is performed by the caller.
+// Start launches the collection goroutine.
 func (c *Collector) Start(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint, pollers []fwkdl.PollingDataSource, extractors map[string][]fwkdl.Extractor) error {
```
hexfusion: note:  pretty sure we can remove Ticker from the signature but will do as a followup

elevran: Possibly used in testing to controlling time progression.

pkg/epp/datalayer/collector.go
```diff
@@ -65,153 +68,107 @@ func (t *TimeTicker) Channel() <-chan time.Time {
...
-			defer func() {
-				logger.V(logging.DEFAULT).Info("terminating collection")
-				ticker.Stop()
-				close(c.done)
-			}()
-
-			close(ready) // signal ready to accept ticks
+// Stop signals the collection goroutine to exit. Idempotent.
+func (c *Collector) Stop() {
+	var msg string
```
elevran: `Stop()` is now fire-and-forget with no guarantee the goroutine has exited. This changes behavior.

The old `Stop()` blocked on `<-c.done`, giving callers a synchronization point. That channel is gone. After `collector.Stop()` returns, the `run` goroutine may still be in the middle of `Poll` or `Extract` and still holding a reference to `ep` and writing to its metrics.

This matters in `ReleaseEndpoint`: the collector is removed from the runtime's map and then `Stop()` is called, but the goroutine can keep accessing `ep` after `ReleaseEndpoint` returns. If the endpoint's backing state is cleaned up or reused concurrently, that is a data race (the `-race` detector might not catch here because the goroutine holds only the `ep` interface value (reachable), not the collector).

For example, `TestCollectorRapidStartStopRaceFree` reuses the same `endpoint` package-level variable across 100 iterations while previous iterations' goroutines may still be running. This is testing the collector lifecycle but not the endpoint-access window.

Consider restoring a `Wait()` method (or a `done <-chan struct{}`) so callers that need a hard synchronization point can get one, without forcing all callers to block:

```go
func (c *Collector) Wait() <-chan struct{} { return c.done }
```

`ReleaseEndpoint` would then call `collector.Stop()` and optionally `<-collector.Wait()` if the endpoint's data can be freed immediately after.

pkg/epp/datalayer/collector.go
```diff
@@ -65,153 +68,107 @@ func (t *TimeTicker) Channel() <-chan time.Time {
...
-	stopped := false
-	c.stopOnce.Do(func() {
-		stopped = true
-		c.cancel()
-	})
-
-	if !stopped {
-		return errors.New("collector stop called multiple times")
+	pollCtx, cancel := context.WithTimeout(ctx, defaultCollectionTimeout)
+	data, err := src.Poll(pollCtx, ep)
```
elevran, line 159: `cancel()` should be deferred, not called inline.

Both timeout contexts in `pollOne` call `cancel()` synchronously and again for `extCtx` / `Extract`.
If `Poll` or `Extract` panics, `cancel()` is never called. The `context.WithTimeout` goroutine then leaks (at least until the deadline fires).

It would be better to `defer cancel()` immediately after each `context.WithTimeout`:

elevran (commented):
Good cleanup overall - the lifecycle simplification and Prometheus counters are an improvement - thanks!
Two items below.

pkg/epp/datalayer/collector.go
hexfusion: yeah defer is more idiomatic in general I for some reason thought this was in a for loop but that is not the case looking at it now.

pkg/epp/datalayer/collector.go
hexfusion: contract: collector's write to an endpoint completes before Stop() so the next operation against that endpoint observes a consistent state.

it is reasonable. will document contract and ensure enforcement

elevran (commented):
/lgtm
/approve

Issue #900, opened 2026-04-27
Add multimodal encoder-cache affinity routing


guygir:
Multimodal requests can repeat the same image/video/audio content across different prompts. Today, routing primarily optimizes for text/kv-prefix cache reuse, but repeated multimodal content can also benefit from being routed back to endpoints that have already processed the same media through the model's preprocessor path.

We should add a scheduler scoring signal for multimodal encoder-cache affinity:

- detect stable multimodal content hashes from tokenized prompt state
- prefer endpoints that recently handled the same multimodal hashes
- weight affinity by multimodal placeholder length when available
- keep this separate from KV-block prefix-cache routing, which remains owned by the precise-prefix-cache scorer
- avoid unbounded request-state retention or stale routing history

This should improve routing for repeated multimodal workloads while preserving the existing precise-prefix-cache scorer as the source of truth for kv-block affinity.

Related PR: #901

rahulgurnani:
Thanks for creating the issue. I also had similar issue in https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2665

guygir:
Progress update on multimodal encoder-cache affinity routing:

- #901 covers the lightweight/unweighted MM cache affinity path. It keeps routing tokenizer-free by using stable MM hashes with unit weights, and remains the low-overhead scorer for repeated MM content.
- #890 changed the shape of the follow-up weighted work: token-producer now exposes vLLM-rendered multimodal metadata, including placeholder lengths, so the weighted scorer can consume authoritative metadata instead of doing custom request traversal.
- I have a follow-up PR ready to open for a separate `weighted-mm-cache-affinity-scorer`. It keeps #901's unweighted scorer intact, adds a separate weighted producer/scorer/data key, consumes `TokenizedPrompt.MultiModalFeatures`, and weights matches by `MultiModalFeature.Length`.
- Latest VisionArena validation used a 4-pod Qwen deployment with cold model restarts per scorer. At 30 RPS for 180s, weighted MM had the best tail/average latency: weighted avg 15.194s / p95 25.663s vs unweighted avg 19.187s / p95 32.933s and precise-prefix avg 24.020s / p95 42.720s. At 10 RPS, unweighted and precise-prefix were slightly faster, which matches the expected tradeoff: weighted MM pays token-producer `/render` overhead and helps most under higher pressure.
[...]

PR #890, opened 2026-04-26
tokenizer: add vLLM HTTP /render backend as default

vMaroon:
## Summary

- Added `vllmHTTP` as a peer backend on the existing tokens-producer plugin
- Turned new backend to default
  - Pulled `llm-d-kv-cache` v0.8.0 to fully execute on the above and not require configuring uds-tokenizer on precise configurations

Includes documentation and example configuration.

**Release note** _(write `NONE` if no user-facing change)_:
```release-note
Adds new configuration option for the `tokens-producer` plugin. Existing configurations are unaffected.
```

A followup PR is formally deprecating the UDS tokenizer.

Builds on #863 (for docs correctness)

PR #901, opened 2026-04-27
Add multimodal encoder-cache affinity scorer

guygir:
This PR adds a single `mm-cache-affinity-scorer` focused on multimodal encoder-cache affinity. It routes requests with repeated multimodal content toward endpoints that recently processed the same MM hashes, while leaving kv-block affinity to the existing `precise-prefix-cache-scorer`.

What changed:

- adds one MM scorer plugin: `mm-cache-affinity-scorer`
- extracts weighted MM hash state from `TokenizedPromptState.MMFeatures`
- weights affinity by `MMPlaceholders.Length` when available, with a fallback weight of 1
- records successful request placement in an LRU `mm_hash -> endpoint` history
- uses TTL-backed request handoff state between `Score` and `PreRequest`
- registers the scorer in the central plugin registry
- adds an example E2E config that composes tokenizer, MM encoder-affinity scoring, and existing precise-prefix/kv affinity

Validation:

- `go test ./pkg/plugins/...`
- `go test -race ./pkg/plugins/scorer/...`

Note: full `go test ./...` still has unrelated environment-dependent failures in config/e2e/profiling tests.

Related issue: #900


rahulgurnani:
Thanks for sending out this change :) 

A high-level comment the scorers reside in pkg/epp/framework/plugins/scheduling/scorer presently.

pkg/plugins/scorer/mm_cache_affinity_test.go
```diff
@@ -0,0 +1,328 @@
...
+
+	"github.com/jellydator/ttlcache/v3"
+	"github.com/llm-d/llm-d-kv-cache/pkg/kvcache/kvblock"
+	"github.com/llm-d/llm-d-kv-cache/pkg/tokenization"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+	fwkdl "sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/datalayer"
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/plugin"
+	"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/interface/scheduling"
```
rahulgurnani: We have migrated epp to llm-d-inference-scheduler repo, this dependency should come from this repo then.

rahulgurnani:
I wrote a writeup with details about implementation: https://github.com/llm-d/llm-d-inference-scheduler/issues/912#issue-4344535094 
We should split the scorer into data producer and a scorer. The data produced by data producer can be reused across other scorers like latency predictor etc.

pkg/plugins/scorer/mm_cache_affinity.go
```diff
@@ -0,0 +1,324 @@
...
+}
+
+// MMCacheAffinityScorer scores endpoints based on multimodal encoder-cache
+// affinity. It intentionally does not reason about KV-block affinity — that
+// remains the responsibility of precise-prefix-cache-scorer. This scorer only
+// tracks which endpoint most recently processed a given mm_hash and weighs that
+// affinity by the encoder token cost implied by MMPlaceholders.
+type MMCacheAffinityScorer struct {
+	typedName         plugin.TypedName
+	lruCache          *lru.Cache[string, string]                 // mm_hash -> endpoint namespaced name
```
rahulgurnani: a multimodal content can be present in multiple pods

guygir: Correct, changed it so that the producer now tracks mm_hash -> set[pod] instead of a single pod

pkg/plugins/scorer/mm_cache_affinity_test.go
guygir: Correct, changed it

guygir:
> I wrote a writeup with details about implementation: [#912 (comment)](https://github.com/llm-d/llm-d-inference-scheduler/issues/912#issue-4344535094) We should split the scorer into data producer and a scorer. The data produced by data producer can be reused across other scorers like latency predictor etc.

Thanks!

I updated this PR to follow the Issue #912 producer/consumer design (and the needed migration and imports). The MM cache logic is now split into multimodal-encoder-cache-data-producer and mm-cache-affinity-scorer: the producer extracts multimodal hashes, keeps an lru of mm_hash -> set[pod], attaches reusable info to endpoints, and the scorer consumes that data for his computation

This PR keeps encoder-cache aware routing tokenizer-free: each unique multimodal hash contributes unit weight 1, so no tokenizer sidecar is required. Weighted/image-size-aware scoring is deferred to a follow-up, where we’ll need a reliable source for MM placeholder counts, ideally derived cheaply from image dimensions/model config (can be done) rather than forcing tokenizer/UDS preprocessing on the scheduling path, because forcing tokenizer preprocessing into the scheduling path adds latency and can block routing on image fetch/decode/preprocess work (observed in tests)

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,480 @@
...
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/requestcontrol"
+	fwkrh "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/requesthandling"
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/scheduling"
+	attrmm "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/datalayer/attribute/multimodal"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	// ProducerType is the type name used to register the multimodal data producer.
+	ProducerType = "multimodal-encoder-cache-data-producer"
```
rahulgurnani: nit: rename to "multimodal-encoder-producer"

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
rahulgurnani, line 1: Suggesting to rename this file to plugin.go for consistency across the directory structure.

rahulgurnani:
Thanks for addressing the earlier comments. Can you also do some micro benchmark of this plugin, like comparing this against a k8s service? We have a  micro-benchmark which we are using https://github.com/capri-xiyue/multi-modal-benchmark-helper , the prefix shared images should just work with encoder cache as well, since requests with same images shared in the prefix have indeed the same images.

rahulgurnani:
> Thanks for addressing the earlier comments. Can you also do some micro benchmark of this plugin, like comparing this against a k8s service? We have a micro-benchmark which we are using https://github.com/capri-xiyue/multi-modal-benchmark-helper , the prefix shared images should just work with encoder cache as well, since requests with same images shared in the prefix have indeed the same images.

@guygir feel free to connect with me on llm-d slack, I go by the same name there. 
I am also benchmarking this code presently. We can discuss results/gaps there if needed for faster turn around. Thanks!

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,480 @@
...
+	}
+	return p.WithName(name), nil
+}
+
+// Producer tracks multimodal content hashes and the pods that likely hold their
+// encoder-cache entries.
+type Producer struct {
+	typedName       plugin.TypedName
+	cache           *lru.Cache[string, map[string]struct{}]
+	requestStates   *ttlcache.Cache[string, map[string]int]
```
rahulgurnani: Thanks for exploring ttl cache here.

 The prefix cache plugin uses PluginState for storing the request state, refer https://github.com/llm-d/llm-d-inference-scheduler/blob/e52311b7d6a0184b79a5ae2fb1383643041c0268/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go#L50

The plugin state also has a cleanup after 5 minutes to avoid memory leaks. I think its better to continue using plugin state since it avoids external dependency and is already being used in the code base. 
Would also let  @elevran and @liu-cong  also chime in on this.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,480 @@
...
+		if err == nil && string(bytes) != "null" {
+			return string(bytes)
+		}
+	}
+	return ""
+}
+
+func contentHash(kind, identifier string) string {
+	sum := sha256.Sum256([]byte(kind + "\x00" + identifier))
+	return kind + ":" + hex.EncodeToString(sum[:])
```
liu-cong: Why adding the kind to the hash output? For debugging purpose?

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,480 @@
...
+}
+
+func addMaxWeight(hashToWeight map[string]int, hash string, weight int) {
```
liu-cong: Let's avoid over defensive code, the caller is this file and we control it completely. Just make sure the caller sends sane inputs (hash is non empty, weight is non zero, etc.).

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,480 @@
...
+	hashToWeight := map[string]int{}
+	for _, message := range request.Messages {
+		for _, block := range message.Content.Structured {
+			addBlockHash(hashToWeight, block)
+		}
+	}
+	return emptyToNil(hashToWeight)
+}
+
+func hashesFromAny(value any) map[string]int {
```
liu-cong: We shouldn't support "any". This logic should only apply to mm inputs. If we don't know if the input contains mm or not, we should just ignore.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
liu-cong: We should not maintain a counter per hash I think - even if we have the same image url multiple times in the request, on the server side it should only be encoded once (I think), so we shouldn't double counting.

guygir:
Thanks @liu-cong, @rahulgurnani for the feedback. I'll be addressing it later on. Updating on the benchmark requested:

I ran a small in-cluster micro benchmark for the scorer, using the helper repo against two Qwen2-VL-2B-Instruct models. I compared three paths: direct k8s service routing, an EPP with random-picker only, and the current MM-affinity scorer path.

The workload was 100 multimodal chat payloads with 207 total image references and 42 unique repeated images. All images were 640x360 JPEGs.

| Path | 1 RPS p50 | 1 RPS p95 | 1 RPS max | 1 RPS avg | 4 RPS p50 | 4 RPS p95 | 4 RPS max | 4 RPS avg |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Kubernetes Service | 57 ms | 105 ms | 691 ms | 64 ms | 27 ms | 90 ms | 696 ms | 47 ms |
| Random EPP | 65 ms | 151 ms | 608 ms | 82 ms | 41 ms | 104 ms | 1.680 s | 66 ms |
| MM affinity EPP | 69 ms | 136 ms | 631 ms | 80 ms | 34 ms | 79 ms | 653 ms | 47 ms |

The MM-affinity result at 4 RPS is the main signal I wanted to validate: it stays in the same latency band as direct service routing and is better than the random EPP baseline on p50, p95, max, and average latency, for this reuse heavy workload, verifying that no meaningful scheduling overhead is introduced.

rahulgurnani:
> Thanks @liu-cong, @rahulgurnani for the feedback. I'll be addressing it later on. Updating on the benchmark requested:
> 
> I ran a small in-cluster micro benchmark for the scorer, using the helper repo against two Qwen2-VL-2B-Instruct models. I compared three paths: direct k8s service routing, an EPP with random-picker only, and the current MM-affinity scorer path.
> 
> The workload was 100 multimodal chat payloads with 207 total image references and 42 unique repeated images. All images were 640x360 JPEGs.
> 
> Path	1 RPS p50	1 RPS p95	1 RPS max	1 RPS avg	4 RPS p50	4 RPS p95	4 RPS max	4 RPS avg
> Kubernetes Service	57 ms	105 ms	691 ms	64 ms	27 ms	90 ms	696 ms	47 ms
> Random EPP	65 ms	151 ms	608 ms	82 ms	41 ms	104 ms	1.680 s	66 ms
> MM affinity EPP	69 ms	136 ms	631 ms	80 ms	34 ms	79 ms	653 ms	47 ms
> The MM-affinity result at 4 RPS is the main signal I wanted to validate: it stays in the same latency band as direct service routing and is better than the random EPP baseline on p50, p95, max, and average latency, for this reuse heavy workload, verifying that no meaningful scheduling overhead is introduced.

I ran a more comprehensive benchmark against a k8s service and EPP with a profile of equal weight to  mm-cache-affinity-scorer and queue-scorer. I saw significant improvement with EPP.


RPS | Metric | EPP (Affinity Enabled) | Direct Service | EPP Improvement
-- | -- | -- | -- | --
20 | Success Rate | 100% (1200/1200) | 99.9% (1199/1200) | Zero Failures
  | p50 Latency | 0.080 s | 0.179 s | 2.2x Faster
  | p95 Latency | 0.605 s | 0.816 s | 1.3x Faster
30 | Success Rate | 100% (1800/1800) | 99.9% (1799/1800) | Zero Failures
  | p50 Latency | 0.074 s | 0.236 s | 3.2x Faster
  | p95 Latency | 0.363 s | 1.018 s | 2.8x Faster
40 | Success Rate | 100% (2400/2400) | 94.4% (2265/2400) | Zero Failures
  | p50 Latency | 0.089 s | 0.304 s | 3.4x Faster
  | p95 Latency | 0.764 s | 3.338 s | 4.4x Faster

Full report: https://github.com/rahulgurnani/llm-d-inference-scheduler/blob/0535be27219ea74af359273ee6815a306c29fc02/comprehensive_performance_report.md

pkg/epp/framework/plugins/datalayer/attribute/multimodal/data_types.go
```diff
@@ -0,0 +1,73 @@
...
+
+const (
+	// EncoderCacheMatchInfoKey is attached to endpoints by the multimodal data
+	// producer and consumed by scorer/latency plugins that need encoder-cache locality.
+	EncoderCacheMatchInfoKey = "MultiModalEncoderCacheMatchInfoKey"
+)
+
+// EncoderCacheMatchInfo summarizes how much of a request's multimodal encoder
+// work is likely already cached on one endpoint.
+type EncoderCacheMatchInfo struct {
```
rahulgurnani, line 39: Instead of computing the weights here, it would be better to keep per item information here i.e.

```
EncoderCacheMatchInfo {
  matchItems MatchItem
  totalWeight int // 
}

MatchItem {
 hash string
 size int
}
```
and compute the weights in the scorer. This enables us to try out multiple scoring logics and provide data for smarter routing to latency predictor as well.

liu-cong:
> Full report: https://github.com/rahulgurnani/llm-d-inference-scheduler/blob/0535be27219ea74af359273ee6815a306c29fc02/comprehensive_performance_report.md

Thanks! Can you put this to a google doc so people can comment on?

rahulgurnani:
> > Full report: https://github.com/rahulgurnani/llm-d-inference-scheduler/blob/0535be27219ea74af359273ee6815a306c29fc02/comprehensive_performance_report.md
> 
> Thanks! Can you put this to a google doc so people can comment on?

Sure @liu-cong , here is a public doc to comment
https://docs.google.com/document/d/1cX4neQtH8YBWZ0ck4DFo4rdJ7zGrPooVxWoTK_mtENA/edit?tab=t.0

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
guygir: switched to framework PluginState, matching the approximate-prefix producer pattern, removing the private TTL cache.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
guygir: Yes, it made the stored value easier to inspect, but it is not needed by consumers. I removed it from the externally stored value, so the stored hash is now just the hex digest.

I still include the modality kind when calculating the hash for domain separation, so the same identifier string used as image/video/audio input cannot collide across modalities.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
guygir: Makes sense, simplified the helper and moved the sanity checks to the controlled callers.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
guygir: Removed broad arbitrary 'any'/map traversal. The producer now extracts from tokenized multimodal features when available, and from known typed chat-completions structured content blocks. If the request is only available as a generic parsed payload map, we'll skip it instead of inferring multimodal content from arbitrary key names.

I checked the current OpenAI parser path: responses and conversations are parsed as first-class request types, but their nested content remains 'any' after JSON unmarshal if I understand correctly. So this PR is scoped to the typed paths that are actually available today. MM content sent through responses/conversations will participate in this affinity only if an upstream step has already converted it into typed TokenizedPrompt.MultiModalFeatures (can be handled separately if needed).

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
guygir: Makes sense, request multimodal items are unique by hash now, so repeated references within one request count once.

pkg/epp/framework/plugins/datalayer/attribute/multimodal/data_types.go
guygir: Sounds good - the data producer now exposes item-level MatchItem data and the scorer computes weights locally, keeping the producer reusable for other consumers / future scoring logic.

guygir:
I addressed the review feedback in the latest commit and replied to the inline comments.
The multimodal producer now uses framework PluginState, emits item-level MatchItem data, and leaves weight computation to mm-cache-affinity-scorer.
I also tightened multimodal extraction to avoid arbitrary payload inference, cleaned up the stored hash format, and made repeated media references count once per request.

Current PR remains scoped to tokenizer free, unit weight encoder cache affinity, while broader shared multimodal extraction/parser-level metadata remains follow-up scope tracked by #1009 .

rahulgurnani:
Thanks for addressing them @guygir . One thing, you commits don't seem to have been signed off, so the presubmit is failing: https://github.com/llm-d/llm-d-inference-scheduler/pull/901/checks?check_run_id=75330934306 , you can use git rebase HEAD~4 --signoff to make the PR check pass.

Also there are some merge conflicts.

capri-xiyue:
Have we evaluated how the encode cache eviction in the plugin works when compared to real encode cache eviction in the model server? For example, in vllm, users can customize encode cache size by GB. Seems like it will only trigger LRU eviction when encode cache size exceeds the specified cache size.

guygir:
> Have we evaluated how the encode cache eviction in the plugin works when compared to real encode cache eviction in the model server? For example, in vllm, users can customize encode cache size by GB. Seems like it will only trigger LRU eviction when encode cache size exceeds the specified cache size.

I've not validated against real model server encoder cache eviction yet.

The current EPP cache is a "best effort" routing signal, not an exact mirror of vLLM’s encoder cache state. If the model server evicts an entry that EPP still remembers, the impact should be a stale routing preference rather than a correctness issue.

furthermore - since EPP does not currently receive model server encoder cache eviction events, it cannot precisely know when a cached MM item was evicted from a pod anyway, but adding eviction/event feedback to keep this state more accurate is part of a broader follow up architecture discussion.

capri-xiyue:
> > Have we evaluated how the encode cache eviction in the plugin works when compared to real encode cache eviction in the model server? For example, in vllm, users can customize encode cache size by GB. Seems like it will only trigger LRU eviction when encode cache size exceeds the specified cache size.
> 
> I've not validated against real model server encoder cache eviction yet.
> 
> The current EPP cache is a "best effort" routing signal, not an exact mirror of vLLM’s encoder cache state. If the model server evicts an entry that EPP still remembers, the impact should be a stale routing preference rather than a correctness issue.
> 
> furthermore - since EPP does not currently receive model server encoder cache eviction events, it cannot precisely know when a cached MM item was evicted from a pod anyway, but adding eviction/event feedback to keep this state more accurate is part of a broader follow up architecture discussion.

 I think there are two paths:
-  approximate encode cache eviction - try to mimic how model server does encode cache eviction, not sure whether it is feasible as you need to know the encode cache size of each image and then see whether eviction is needed 
-  precise encode cache eviction - see whether model server support encoder cache eviction events and EPP listen to it

I'm fine with either path. But I think at least we should investigate and have some cache eviction mechanism for the encoder-cache affinity scorer.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
guygir: @liu-cong heads up, I saw #1072 and I think it may affect the longer-term shape here.

If #1072 moves plugins toward consuming the original lossless payload instead of the typed request structs, I think the right direction would be a shared/schema-aware MM extractor over the raw payload, rather than reintroducing 'any'/map traversal in this producer. That also seems aligned with #1009.

rahulgurnani:
> I think there are two paths:
> 
> * approximate encode cache eviction - try to mimic how model server does encode cache eviction, not sure whether it is feasible as you need to know the encode cache size of each image and then see whether eviction is needed
> * precise encode cache eviction - see whether model server support encoder cache eviction events and EPP listen to it
> 
> I'm fine with either path. But I think at least we should investigate and have some cache eviction mechanism for the encoder-cache affinity scorer.

I agree. I think this can be a follow up, created https://github.com/llm-d/llm-d-router/issues/1144 as a follow up

rahulgurnani:
@liu-cong looks good to me for now. We can make caching improvements in a follow up.

liu-cong:
Thank you! I will review this ASAP.

pkg/epp/framework/plugins/scheduling/scorer/mmcacheaffinity/scorer.go
```diff
@@ -0,0 +1,137 @@
...
+
+// Consumes returns the endpoint data consumed by this scorer.
+func (s *Scorer) Consumes() map[string]any {
+	return map[string]any{attrmm.EncoderCacheMatchInfoKey: attrmm.EncoderCacheMatchInfo{}}
+}
+
+// Score scores endpoints by matched multimodal encoder-cache item size divided
+// by total multimodal request item size.
+func (s *Scorer) Score(ctx context.Context, _ *scheduling.CycleState, req *scheduling.InferenceRequest, endpoints []scheduling.Endpoint) map[scheduling.Endpoint]float64 {
+	logger := log.FromContext(ctx).V(logging.DEBUG)
```
vMaroon: The logging below looks more like that of tracing - I think the logger name should be changed to `traceLogger` and level to `logging.TRACE`.

test/e2e/configs_test.go
```diff
@@ -238,6 +238,49 @@ schedulingProfiles:
     weight: 10
 `
 
+// EPP config for multimodal encoder-cache affinity plus precise KV-prefix affinity.
+// The multimodal data producer runs before scheduling and the scorer consumes its endpoint match info.
+const mmCacheAffinityConfig = `apiVersion: inference.networking.x-k8s.io/v1alpha1
+kind: EndpointPickerConfig
+plugins:
```
vMaroon: Pls use the up to date recommended configuration, done in #890.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/plugin.go
```diff
@@ -0,0 +1,383 @@
...
+func (p *Producer) Consumes() map[string]any {
+	return nil
+}
+
+// PluginState returns request-scoped state shared between producer extension points.
+func (p *Producer) PluginState() *plugin.PluginState {
+	return p.pluginState
+}
+
+// PrepareRequestData attaches multimodal encoder-cache match data to endpoints.
```
vMaroon: s/PrepareRequestData/Produce.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,383 @@
...
+	if len(requestItems) == 0 {
+		logger.Info("No multimodal content found, skipping encoder-cache match data")
+		return nil
+	}
+
+	if request != nil && request.RequestID != "" {
+		p.pluginState.Write(request.RequestID, plugin.StateKey(ProducerType), &requestState{items: requestItems})
+	}
+	// TODO(#1144): Removal of stale pods should happen in background for better performance.
+	p.removeStalePods()
```
vMaroon: You can subscribe to endpoint lifecycle events through the datalayer endpoint-extractors, see https://github.com/llm-d/llm-d-router/pull/1112/changes#diff-0fcaa5a021d63526d456e879ee156c743d991fd208e8094a0d90e2f4ee9cf429. It is already in the codebase, but the linked refactor makes it clear.

Since it already exists in the codebase, importing it here is low-risk.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,383 @@
...
+
+// Produces returns the data keys this plugin produces.
+func (p *Producer) Produces() map[string]any {
+	return map[string]any{ProducedKey: attrmm.EncoderCacheMatchInfo{}}
+}
+
+// Consumes returns the data keys this plugin requires.
+func (p *Producer) Consumes() map[string]any {
+	return nil
+}
```
vMaroon:
```suggestion
// Consumes declares the TokenizedPrompt dependency so the data-layer DAG
// orders `token-producer` first. No-op when TokenizedPrompt is absent.
func (p *Producer) Consumes() map[string]any {
	return map[string]any{tokenproducer.TokenizedPromptKey: scheduling.TokenizedPrompt{}}
}
```

vMaroon (changes requested):
Overall looks good, thanks @guygir.

Added a few comments. In addition, these are missing:
1. Add a plugin README.md that is descriptive of what it does and its configuration. See others for example.
2. A `deploy/config` config example.
3. Since the `dataproducer` also implements `PreRequest`, separate the latter into a `prerequest.go` and rename `plugin.go` to `producer.go` to align with others.

pkg/epp/framework/plugins/scheduling/scorer/mmcacheaffinity/scorer.go
```diff
@@ -0,0 +1,137 @@
...
+var (
+	_ scheduling.Scorer     = &Scorer{}
+	_ plugin.ConsumerPlugin = &Scorer{}
+)
+
+// Factory creates a multimodal encoder-cache affinity scorer.
+func Factory(name string, rawParameters json.RawMessage, _ plugin.Handle) (plugin.Plugin, error) {
+	if len(rawParameters) > 0 {
+		var parameters map[string]any
+		if err := json.Unmarshal(rawParameters, &parameters); err != nil {
```
vMaroon: params are parsed but not used anywhere.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/plugin.go
```diff
@@ -0,0 +1,383 @@
...
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/requestcontrol"
+	fwkrh "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/requesthandling"
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/scheduling"
+	attrmm "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/datalayer/attribute/multimodal"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	// ProducerType is the type name used to register the multimodal data producer.
+	ProducerType = "multimodal-encoder-producer"
```
vMaroon:
```suggestion
	ProducerType = "mm-embeddings-cache-producer"
```

`mm-embeddings-cache-producer` is more in line with the naming convention we're carrying.

pkg/epp/framework/plugins/scheduling/scorer/mmcacheaffinity/scorer.go
```diff
@@ -0,0 +1,137 @@
...
+
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/common/observability/logging"
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/plugin"
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/scheduling"
+	attrmm "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/datalayer/attribute/multimodal"
+)
+
+const (
+	// Type is the type name used to register the multimodal encoder-cache scorer.
+	Type = "mm-cache-affinity-scorer"
```
vMaroon:
```suggestion
	Type = "mm-embeddings-cache-scorer"
```

`mm-embeddings-cache-scorer` is more analogous to the established `prefix-cache-scorer` (producers too).

guygir:
Thanks @vMaroon. I addressed the review comments in the latest update:

Renamed the multimodal embeddings cache plugins, split producer hooks into dedicated files, added endpoint lifecycle cleanup, and documented the producer/scorer configuration.

Validated with focused tests for the multimodal producer, MM scorer, plugin registration, and data layer packages.

Also adding the VisionArena benchmark numbers for the unweighted scorer path in this PR:

Scorers:

- Random: EPP baseline that randomly picks among candidate decode endpoints.
- Unweighted MM: this PR's multimodal embeddings cache affinity path; each unique MM hash counts once and no `/render` metadata is required for weighting.
- Precise Prefix: existing KV-prefix cache scorer using token-producer / vLLM render metadata for precise prefix-block affinity.

The first two tables use a smaller VisionArena multi-turn workload: 100 conversations converted to Qwen requests with `min_turns=2`, `max_turns_per_conversation=20`, unshuffled, `max_tokens=1`. 288 valid requests. The stress tables use a larger in-cluster VisionArena workload with `min_turns=4`, `max_turns_per_conversation=20`, `max_tokens=512`, filtered to prompt length `<=3584`, leaving 5,536 valid requests. All runs used 4 model pods.

VisionArena multi-turn, 288 valid requests, 10 RPS:

| Scorer | Success | Actual RPS | p50 | p95 | Max | Avg |
|---|---:|---:|---:|---:|---:|---:|
| Unweighted MM | 288/288 | 10.00 | 93 ms | 373 ms | 892 ms | 138 ms |
| Precise Prefix | 288/288 | 9.99 | 137 ms | 453 ms | 915 ms | 182 ms |

VisionArena random vs. unweighted verification, same 288-request payload, 10 RPS:

| Scorer | Success | Actual RPS | p50 | p95 | Max | Avg |
|---|---:|---:|---:|---:|---:|---:|
| Random | 288/288 | 9.99 | 132 ms | 367 ms | 879 ms | 165 ms |
| Unweighted MM | 288/288 | 10.00 | 96 ms | 305 ms | 875 ms | 132 ms |

VisionArena stress payload, 5,536 valid requests, 30 RPS for 180s:

| Scorer | Success | Actual RPS | p50 | p75 | p95 | Max | Avg |
|---|---:|---:|---:|---:|---:|---:|---:|
| Unweighted MM | 5399/5400 | 24.96 | 20.160 s | 27.512 s | 32.933 s | 38.744 s | 19.187 s |
| Precise Prefix | 5399/5400 | 23.61 | 24.877 s | 34.425 s | 42.720 s | 49.779 s | 24.020 s |

Same VisionArena stress payload, 10 RPS for 180s:

| Scorer | Success | Actual RPS | p50 | p75 | p95 | Max | Avg |
|---|---:|---:|---:|---:|---:|---:|---:|
| Unweighted MM | 1800/1800 | 9.90 | 0.942 s | 1.664 s | 2.336 s | 3.489 s | 1.097 s |
| Precise Prefix | 1800/1800 | 9.89 | 0.960 s | 1.596 s | 2.389 s | 3.094 s | 1.107 s |

guygir:
Weighted follow-up is split into [guygir/llm-d-inference-scheduler#1](https://github.com/guygir/llm-d-inference-scheduler/pull/1) .

pkg/epp/framework/plugins/scheduling/scorer/mmcacheaffinity/scorer.go
```diff
@@ -0,0 +1,129 @@
...
+
+	"github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
+	attrmm "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/attribute/multimodal"
+)
+
+const (
+	// Type is the type name used to register the multimodal encoder-cache scorer.
+	Type = "mm-embeddings-cache-scorer"
```
liu-cong, line 36: pls also rename the package to mmembeddingscacheaffinity

pkg/epp/framework/plugins/scheduling/scorer/mmcacheaffinity/scorer.go
```diff
@@ -0,0 +1,129 @@
...
+	return &Scorer{typedName: plugin.TypedName{Type: Type}}
+}
+
+// TypedName returns the plugin type/name.
+func (s *Scorer) TypedName() plugin.TypedName {
+	return s.typedName
+}
+
+// WithName sets the plugin instance name.
+func (s *Scorer) WithName(name string) *Scorer {
```
liu-cong: Pls remove WithName and add `name` to the New() directly. Plugin should always have a name (the framework will default to the type if not configured)

pkg/epp/framework/plugins/scheduling/scorer/mmcacheaffinity/scorer.go
```diff
@@ -0,0 +1,129 @@
...
+	return New().WithName(name), nil
+}
+
+// Scorer computes normalized endpoint affinity from produced multimodal match data.
+type Scorer struct {
+	typedName plugin.TypedName
+}
+
+// New creates a Scorer.
+func New() *Scorer {
```
liu-cong: A "producerName" config needs to be introdcued, see example https://github.com/llm-d/llm-d-router/blob/25b258433b2909070499976a209b9a99e08bdfba/pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go#L35

pkg/epp/framework/plugins/scheduling/scorer/mmcacheaffinity/scorer.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// Category returns the scorer category.
+func (s *Scorer) Category() scheduling.ScorerCategory {
+	return scheduling.Affinity
+}
+
+// Consumes returns the endpoint data consumed by this scorer.
+func (s *Scorer) Consumes() map[string]any {
+	return map[string]any{attrmm.EncoderCacheMatchInfoKey: attrmm.EncoderCacheMatchInfo{}}
```
liu-cong: Now it needs the DataKey, See example in https://github.com/llm-d/llm-d-router/blob/25b258433b2909070499976a209b9a99e08bdfba/pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go#L35

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/README.md
```diff
@@ -0,0 +1,71 @@
...
+- `MultiModalEncoderCacheMatchInfoKey` (`EncoderCacheMatchInfo`)
+
+## Configuration
+
+The producer supports the following runtime parameters:
+
+- `cacheSize` (integer, default: `10000`): maximum number of multimodal hash entries
+  retained in the best-effort pod-affinity cache.
+
+**Configuration Example:**
```
liu-cong: Can you also provide an example without tokenizer?

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/prerequest.go
```diff
@@ -0,0 +1,77 @@
...
+		return
+	}
+
+	targets := targetEndpoints(schedulingResult)
+	if len(targets) == 0 {
+		logger.Info("No target endpoints found, skipping encoder-cache update")
+		return
+	}
+
+	p.mutex.Lock()
```
liu-cong: Cache update can be done async to avoid bottleneck, see example in the prefix cache producer.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,367 @@
...
+var (
+	_ requestcontrol.DataProducer = &Producer{}
+	_ requestcontrol.PreRequest   = &Producer{}
+	_ fwkdl.EndpointExtractor     = &Producer{}
+)
+
+// Parameters configures the multimodal encoder-cache data producer.
+type Parameters struct {
+	// CacheSize defines the maximum number of mm_hash -> pod-set entries to track.
+	CacheSize int `json:"cacheSize"`
```
liu-cong, line 65: This can be a follow up but we need to document some guidance on how to choose this number. Perhaps use an example model deployment, and explain how this maps to the actual memory footprint on the model servers

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,367 @@
...
+	}, nil
+}
+
+// TypedName returns the plugin type/name.
+func (p *Producer) TypedName() plugin.TypedName {
+	return p.typedName
+}
+
+// WithName sets the plugin instance name.
+func (p *Producer) WithName(name string) *Producer {
```
liu-cong: nit: Remove this and add name to the New function.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,367 @@
...
+	return p
+}
+
+// Produces returns the data keys this plugin produces.
+func (p *Producer) Produces() map[string]any {
+	return map[string]any{ProducedKey: attrmm.EncoderCacheMatchInfo{}}
+}
+
+// Consumes returns the data keys this plugin requires.
+func (p *Producer) Consumes() map[string]any {
```
liu-cong: We have a "auto create default producer" feature where the framework will try to create the producer if not configured explicitly. The "Consumes" declaration enforces the framework to do that.

So here this data is optional. Please just remove "Consumes".

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,367 @@
...
+	requestItems := ExtractMMItems(request)
+	if len(requestItems) == 0 {
+		logger.Info("No multimodal content found, skipping encoder-cache match data")
+		return nil
+	}
+
+	if request != nil && request.RequestID != "" {
+		p.pluginState.Write(request.RequestID, plugin.StateKey(ProducerType), &requestState{items: requestItems})
+	}
+	// TODO(#1144): Removal of stale pods should happen in background for better performance.
```
liu-cong: can we just run a periodic loop for the cleanup?

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
liu-cong: I am actually in favor of the simple periodic cleanup loop approach, it's simple and reliable, and don't need to worry about "what if I miss an event".

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,367 @@
...
+func contentHash(kind, identifier string) string {
+	sum := sha256.Sum256([]byte(kind + "\x00" + identifier))
+	return hex.EncodeToString(sum[:])
+}
+
+func addItem(itemsByHash map[string]attrmm.MatchItem, hash string) {
+	itemsByHash[hash] = attrmm.MatchItem{Hash: hash, Size: 1}
+}
+
+func sortedItems(itemsByHash map[string]attrmm.MatchItem) []attrmm.MatchItem {
```
liu-cong: why do we need to sort?

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,367 @@
...
+		delete(pods, pod)
+		if len(pods) == 0 {
+			p.cache.Remove(hash)
+			continue
+		}
+		p.cache.Add(hash, pods)
+	}
+}
+
+func (p *Producer) cacheSnapshot() map[string]map[string]struct{} {
```
liu-cong: why do we need this?

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -0,0 +1,367 @@
...
+	snapshot := make(map[string]map[string]struct{}, p.cache.Len())
+	for _, hash := range p.cache.Keys() {
+		if pods, ok := p.cache.Get(hash); ok {
+			snapshot[hash] = maps.Clone(pods)
+		}
+	}
+	return snapshot
+}
+
+func (p *Producer) putCacheEntry(hash string, pods ...k8stypes.NamespacedName) {
```
liu-cong: why do we need this?

pkg/epp/framework/plugins/scheduling/scorer/mmcacheaffinity/scorer.go
rahulgurnani: mmembeddingscacheaffinity is verbose for a Go package name — Go convention favors short, lowercase names. mmcacheaffinity is more idiomatic. Happy to use a shorter variant if you have one in mind : https://go.dev/blog/package-names

pkg/epp/framework/plugins/scheduling/scorer/mmcacheaffinity/scorer.go
rahulgurnani: updated, thanks!

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
rahulgurnani: removed, thanks!

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
rahulgurnani: This seems to be only used in tests. Moved it to a separate file.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
rahulgurnani: This seems to be only used in tests, moved it

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
rahulgurnani: I created https://github.com/llm-d/llm-d-router/issues/1144 to track this (point 3). Thanks!

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/README.md
rahulgurnani: Added an example

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
rahulgurnani: It seems the sorting is done to help with the tests because tests were using assert.Equal . I updated the tests to use assert.ElementsMatch and removed sorting the items. Thanks!

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
rahulgurnani: Updated the code to use periodic cleanup

pkg/epp/framework/plugins/scheduling/scorer/mmcacheaffinity/scorer.go
rahulgurnani: updated to use the new format. thanks!

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/prerequest.go
rahulgurnani: made this async in this PR. thanks!

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
rahulgurnani: removed, thanks!

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
rahulgurnani: yes, updated. thanks!

PR #905, opened 2026-04-27
refactor(disagg): add PreRequest to disagg-profile-handler, deprecate disagg-headers-handler

### LINKED ISSUE #833: Evaluate possibility of combining disagg_profile_handler and disagg_headers_handler

noalimoy:
**What type of PR is this?**

/kind cleanup
/kind deprecation

`disagg-profile-handler` (`Handler`) now implements both `ProfileHandler` and `PreRequest`, so configs no longer need a separate `disagg-headers-handler` entry.

`disagg_headers_handler.go` and its tests are preserved with all exported symbols marked `// Deprecated:` — existing configs that reference `disagg-headers-handler` or `prefill-header-handler` continue to work without changes. 
This introduces temporary code duplication that can be cleaned up once the deprecated plugin is removed in a future release.

**Key changes:**
- Add `PreRequest` method to `Handler` in `disagg_profile_handler.go`
- Mark `disagg_headers_handler.go` and all its exported symbols as `Deprecated`
- Remove `disagg-headers-handler` entries from deploy configs, e2e test configs, and documentation examples
- Add `TestBothProfileAndHeadersHandlerPreRequest` to verify both handlers can coexist safely (redundant but not conflicting)

**Backward compatibility:**
When both `disagg-profile-handler` and `disagg-headers-handler` appear in the same config, `HeadersHandler` implements only `PreRequest` (not `ProfileHandler`), so the "multiple profile handlers" validation does not trigger. The prefill/encode headers are written redundantly by both handlers — safe and idempotent.

Fixes #833

**Release note (write NONE if no user-facing change):**
```release-note
disagg-headers-handler and prefill-header-handler plugins are now deprecated.
their PreRequest functionality is built into disagg-profile-handler.
existing configs continue to work; the deprecated plugin will be removed in a future release

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/README.md
+++ b/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/README.md
@@ -1,6 +1,6 @@
 # Disaggregated Profile Handler, PreRequest, and Decider Plugins
 
-Plugins for disaggregated inference scheduling: a profile handler that selects the active stages: EPD (no disaggregation), P/D (Prefill/Decode), E/P/D (Encode/Prefill/Decode), or E/PD (Encode/Prefill-Decode), a headers handler that routes pods across stages, and decider plugins that control whether each disaggregation stage runs per request.
+Plugins for disaggregated inference scheduling: a profile handler that selects the active stages: EPD (no disaggregation), P/D (Prefill/Decode), E/P/D (Encode/Prefill/Decode), or E/PD (Encode/Prefill-Decode), legacy headers handlers (deprecated) kept for backward compatibility, and decider plugins that control whether each disaggregation stage runs per request.
 
 ## Contents
 
@@ -8,7 +8,7 @@ Plugins for disaggregated inference scheduling: a profile handler that selects t
   - [DisaggProfileHandler](#disaggprofilehandler)
   - [PdProfileHandler (Deprecated)](#pdprofilehandler-deprecated)
 - [PreRequest Plugins](#prerequest-plugins)
-  - [DisaggHeadersHandler](#disaggheadershandler)
+  - [DisaggHeadersHandler (Deprecated)](#disaggheadershandler-deprecated)
   - [PrefillHeaderHandler (Deprecated)](#prefillheaderhandler-deprecated)
 - [Decider Plugins](#decider-plugins)
   - [PrefixBasedPDDecider](#prefixbasedpddecider)
@@ -100,10 +100,16 @@ plugins:
 
 ## PreRequest Plugins
 
-### DisaggHeadersHandler
+### DisaggHeadersHandler (Deprecated)
 
 **Type:** `disagg-headers-handler`
 **Interfaces**: `requestcontrol.PreRequest`
+
+> **Deprecated:** Use `disagg-profile-handler` instead.
+>
+> `disagg-profile-handler` now implements `requestcontrol.PreRequest` natively.
+>
+> Planned removal: `v0.11`.
 
 Sets HTTP routing headers on the outgoing request so the inference proxy can forward prefill and encode work to the selected disaggregated pods.
 
@@ -152,7 +158,9 @@ plugins:
 **Type:** `prefill-header-handler`
 **Interfaces**: `requestcontrol.PreRequest`
 
-> **Deprecated:** Use `disagg-headers-handler` instead.
+> **Deprecated:** Use `disagg-profile-handler` instead.
+>
+> Planned removal: `v0.11`.
 
 ---
 
--- a/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler_test.go
+++ b/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler_test.go
@@ -4,11 +4,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	k8stypes "k8s.io/apimachinery/pkg/types"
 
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by roytman** [comment]
Thank you @noalimoy.
I think it will be better to restore `disagg_headers_handler.go` and `disagg_headers_handler_test.go, and define them as deprecated. Instead of creating the legacyHeadersPlugin. 

Just to be on the safe side, can you check what happens if a user defines both `disagg-profile-handler` and `disagg-headers-handler` ?

**Comment by noalimoy** [comment]
> I think it will be better to restore disagg_headers_handler.go and `disagg_headers_handler_test.go, and define them as deprecated. Instead of creating the legacyHeadersPlugin.

done - updated in the latest push:

>  can you check what happens if a user defines both disagg-profile-handler and disagg-headers-handler ?

checked - when both are defined, `HeadersHandler` only implements `PreRequest` (not `ProfileHandler`), so no "multiple profile handlers" conflict. The prefill/encode headers are written twice (redundant but not conflicting and can be cleaned up once the deprecated plugin is removed in a future release.).

File: pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler_test.go [hunk 1]
```diff
@@ -1290,3 +1291,44 @@ func TestHandler_Factory_NilDeciders(t *testing.T) {
...
+		PrimaryProfileName: "decode",
+		ProfileResults: map[string]*scheduling.ProfileRunResult{
+			"prefill": {TargetEndpoints: []scheduling.Endpoint{ep}},
+		},
+	}
+
+	profileHandler.PreRequest(ctx, request, result)
+	headersHandler.PreRequest(ctx, request, result)
+
+	expected := fmt.Sprintf("%s:%s", podAddr, podPort)
```
roytman: Better to use `net.JoinHostPort`
For IPv6 addresses, the production code would produce [::1]:8080 but the test would assert ::1:8080
Other tests use net.JoinHostPort

roytman:
Several comments:

- Please update the PR definition, it does have user-facing change.
- docs/architecture.md L:781 says PrefillHeader (i.e., disagg-headers-handler) "must be in the list of plugins instantiated," but the PR deprecates it and moves that functionality into disagg-profile-handler. The note was not updated.
- Duplicate OTel span name — Both Handler.PreRequest (new) and HeadersHandler.PreRequest (deprecated) use the identical span name "llm_d.epp.prerequest.disaggregation". During migration, when both coexist (a scenario the PR explicitly tests), every trace has two indistinguishable sibling spans. Probably, we are not going to face this situation, and the deprecated code will be removed soon.
- Personally, I don't like the duplicated code `unc (p *HeadersHandler) PreRequest(ctx context.Context, request *scheduling.InferenceRequest, schedulingResult *scheduling.SchedulingResult)`. Theoretically, we can change the call method to a function with 2 additional arguments (`p.prefillProfile` and `p.encodeProfile`) and call it from both `disagg-headers-handler` and `disagg-profile-handler`, but probably we will remove the deprecated classes soon.

noalimoy:
> Several comments:
> 
> * Please update the PR definition, it does have user-facing change.
> * docs/architecture.md L:781 says PrefillHeader (i.e., disagg-headers-handler) "must be in the list of plugins instantiated," but the PR deprecates it and moves that functionality into disagg-profile-handler. The note was not updated.
> * Duplicate OTel span name — Both Handler.PreRequest (new) and HeadersHandler.PreRequest (deprecated) use the identical span name "llm_d.epp.prerequest.disaggregation". During migration, when both coexist (a scenario the PR explicitly tests), every trace has two indistinguishable sibling spans. Probably, we are not going to face this situation, and the deprecated code will be removed soon.
> * Personally, I don't like the duplicated code `unc (p *HeadersHandler) PreRequest(ctx context.Context, request *scheduling.InferenceRequest, schedulingResult *scheduling.SchedulingResult)`. Theoretically, we can change the call method to a function with 2 additional arguments (`p.prefillProfile` and `p.encodeProfile`) and call it from both `disagg-headers-handler` and `disagg-profile-handler`, but probably we will remove the deprecated classes soon.

3 & 4 — agreed on both. the duplication is temporary and code will be removed in a future release, not worth refactoring something that gets discarded soon.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler_test.go
noalimoy: Fixed - switched to `net.JoinHostPort`. initially used `fmt.Sprintf` since the test address is IPv4, but `JoinHostPort` is the right choice regardless for consistency and IPv6 correctness..

elevran:
can you please add an issue to release milestone N+2 (we are at 0.9 (or more accurately post 0.8) so to 0.11) to deprecate and actually remove the code?

noalimoy:
@roytman @elevran
addressed the feedback and pushed the latest changes.
could you please re-review?

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/README.md
```diff
@@ -100,11 +100,15 @@ plugins:
 
 ## PreRequest Plugins
 
-### DisaggHeadersHandler
+### DisaggHeadersHandler (Deprecated)
```
ahg-g, line 103: indicate which release this will be removed

noalimoy: updated — planned removal in v0.11

PR #906, opened 2026-04-28
feat(sidecar): allow injecting custom proxy handlers

### LINKED ISSUE #896: Allow injecting custom proxy handlers (encode/prefill/decode) in sidecar proxy
Allow users to inject custom `http.Handler` (or a handler factory) for encode/prefill/decode proxy handlers in the sidecar proxy.

Currently, these handlers are constructed internally(using `httputil.ReverseProxy`).
It would be helpful to provide an extension point so that users can replace or customize the handler logic.

A minimal approach could be:
- Add optional handler or handler factory fields
- If provided, use them instead of constructing the default reverse proxy
- Otherwise, fall back to the existing behavior

Example (conceptual):
```go
// ProxyHandlerFactory creates an http.Handler for the given target URL.
type ProxyHandlerFactory func(targetURL *url.URL) http.Handler
type Server struct {
	// ...
	decoderHandlerFactory   ProxyHandlerFactory   // optional custom decoder handler
	prefillerHandlerFactory ProxyHandlerFactory   // optional custom prefiller handler
	encoderHandlerFactory   ProxyHandlerFactory   // optional custom encoder handler
}
func (s *Server) createProxyHandler(
	hostPort string,
	cache *lru.Cache[string, http.Handler],
	urlPrefix string,
	insecureSkipVerify bool,
	proxyFactory ProxyHandlerFactory,
) (http.Handler, error) {
	// ... existing URL parsing logic ...
	var newProxy http.Handler
	if proxyFactory != nil {
		newProxy = proxyFactory(u)
	} else {
 		rp = httputil.NewSingleHostReverseProxy(u)
		rp.Transport = s.newProxyTransport(u.Scheme, insecureSkipVerify)
		newProxy = rp
	}
    	...
}
```

**Why is this needed**:
In some use cases, the proxy needs to do more than simple reverse proxing.
For example:
- intergrating external services (e.g. Redis-based token usage collecting per project/user)
- performing request/response transforming before forwarding(e.g. content filtering)

There use cases require replacing or extending the handler logic itself, not just tweaking reverse proxy settings
At the moment, this required modifiying internal code or maintaining a fork which makes it harder to keep up with upstream changes.

questcollector:
Add ProxyHandlerFactory type to Config for decoder, prefiller, and encoder proxies. Falls back to default reverse proxy when not set.

**What type of PR is this?**

/kind feature

Add ProxyHandlerFactory type to Config for decoder, prefiller, and encoder proxies. This allows users to inject custom http.Handler for use cases like:
- **Token usage tracking**: During prefill/decode disaggregation, the prefill stage has access to exact prompt tokens and cached tokens values (decode processes as 100% cached tokens). Custom handlers can capture this information for accurate per-project/user token accounting (e.g., Redis-based tracking).
- Request/response transformation and content filtering.
Falls back to default reverse proxy when not set.

Fixes #896 

```release-note
NONE
```

### DIFF
--- a/pkg/sidecar/proxy/proxy.go
+++ b/pkg/sidecar/proxy/proxy.go
@@ -174,7 +174,7 @@ type Config struct {
 	InferencePoolName string
 	// PoolGroup is the API group of the InferencePool resource.
 	PoolGroup string
-	
+
 	// DecoderHandlerFactory is an optional custom handler factory for the decoder proxy.
 	// If set, it replaces the default reverse proxy handler.
 	DecoderHandlerFactory ProxyHandlerFactory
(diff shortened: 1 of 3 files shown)


elevran:
Suggest addressing the question of whether this functionality belongs in the sidecar or in EPP on #896 first.

elevran:
@questcollector can this be closed in light of discussion on #896?

questcollector:
Okay

Issue #907, opened 2026-04-28
Add pluggable DiscoveryPlugin interface


ezrasilvera:
**What would you like to be added**:

  A DiscoveryPlugin interface that decouples backend endpoint discovery from Kubernetes. The interface defines
  two responsibilities: an initial enumeration of all known endpoints, and a continuous watch for lifecycle
  changes (add/delete). Discovery plugins communicate endpoint state to the datastore through a Notifier
  callback, which enforces strict event ordering so that add/delete sequences are never reordered.

**Why is this needed**:
The current discovery mechanism is tightly coupled to Kubernetes, which limits the scheduler to Kubernetes-native deployments. As llm-d expands to support reinforcement learning (RL) workloads, inference servers increasingly run on non-Kubernetes infrastructure such as Slurm clusters and Ray clusters. A pluggable discovery abstraction allows the scheduler to work in these environments without changes to its core logic.

The proposed next steps are:
1. Refactor the existing Kubernetes-based discovery into a DiscoveryPlugin implementation, preserving current behavior.
2. Introduce a file-based DiscoveryPlugin as the first non-Kubernetes implementation, enabling static or externally-managed endpoint lists — a simple and portable starting point for Slurm and Ray environments.

elevran:
xref: https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2310

ahg-g:
This is done, right? can we close it if so pls?

ezrasilvera:
yes.

PR #908, opened 2026-04-28
feat: Add pluggable DiscoveryPlugin interface

ezrasilvera:
**What type of PR is this?**

This PR addresses #907 

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/interface/datalayer/discovery.go
+++ b/pkg/epp/framework/interface/datalayer/discovery.go
@@ -14,16 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package discovery defines the EndpointDiscovery abstraction for populating
-// the datastore with inference server endpoints independently of Kubernetes.
-package discovery
+package datalayer
 
 import (
 	"context"
 
 	"k8s.io/apimachinery/pkg/types"
 
-	fwkdl "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/datalayer"
 	fwkplugin "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/plugin"
 )
 
@@ -42,23 +39,23 @@ type EndpointDiscovery interface {
 	// datastore at startup. Implementations that guarantee no missed events
 	// through their watch mechanism (e.g. a Kubernetes list+watch) may fold the
 	// initial enumeration into the watch sequence instead.
-	Start(ctx context.Context, notifier Notifier) error
+	Start(ctx context.Context, notifier DiscoveryNotifier) error
 }
 
-// Notifier is the callback through which EndpointDiscovery communicates endpoint state
-// to the datastore.
+// DiscoveryNotifier is the callback through which EndpointDiscovery communicates
+// endpoint state to the datastore.
 //
-// Notifier is NOT goroutine-safe. All calls must be made sequentially from a
-// single goroutine. This is the source of the ordering contract below.
+// DiscoveryNotifier is NOT goroutine-safe. All calls must be made sequentially
+// from a single goroutine. This is the source of the ordering contract below.
 //
 // Ordering contract: the datastore processes Upsert and Delete calls in the order
 // they are received. Plugin implementations MUST preserve event order -- do not
 // buffer or dispatch calls concurrently in a way that could reorder them. For
 // example, an Upsert followed by a Delete for the same endpoint must arrive in
 // that order, or the endpoint will be incorrectly left in the datastore.
-type Notifier interface {
+type DiscoveryNotifier interface {
 	// Upsert adds or updates an endpoint in the datastore.
-	Upsert(endpoint *fwkdl.EndpointMetadata)
+	Upsert(endpoint *EndpointMetadata)
 	// Delete removes an endpoint from the datastore by its namespaced name.
 	Delete(id types.NamespacedName)
 }
--- a/pkg/epp/framework/interface/discovery/discovery.go
+++ b/pkg/epp/framework/interface/discovery/discovery.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package discovery defines the BackendDiscovery abstraction for populating
+// Package discovery defines the DiscoveryPlugin abstraction for populating
 // the datastore with inference server endpoints independently of Kubernetes.
 package discovery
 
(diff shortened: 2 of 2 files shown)


pkg/epp/framework/interface/discovery/discovery.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"k8s.io/apimachinery/pkg/types"
+
+	fwkdl "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/datalayer"
+)
+
+// DiscoveryPlugin discovers inference endpoints and drives their lifecycle in the datastore.
+// Implementations must also implement fwkplugin.Plugin so they can be registered in the
+// plugin registry and selected via EndpointPickerConfig.discovery.pluginRef.
+type DiscoveryPlugin interface {
```
elevran: Instead of noting this in the comment, it would be better to embed the `Plugin` interface here so the requirement is enforced by the compiler.

pkg/epp/framework/interface/discovery/discovery.go
```diff
@@ -0,0 +1,53 @@
...
+	// Start begins discovery. It should:
+	//   1. Enumerate all known endpoints, calling notifier.Upsert for each.
+	//   2. Continue watching for changes until ctx is cancelled.
```
elevran: I think these are implementation details and specific plugin implementations shuold be allowed to deviate. For example, is (1) really required or can it be handled as part of normal life cycle updates as in (2)?

pkg/epp/framework/interface/discovery/discovery.go
```diff
@@ -0,0 +1,53 @@
...
+	// Blocks until ctx is cancelled or a fatal error occurs.
+	Start(ctx context.Context, notifier Notifier) error
```
elevran: who manages the concurrency - is `Start()` expected to be called in a new goroutine or should it create one internally?

pkg/epp/framework/interface/datalayer/discovery.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Notifier is the callback through which DiscoveryPlugin communicates endpoint state
+// to the datastore.
+//
+// Ordering contract: the datastore processes Upsert and Delete calls in the order
+// they are received. Plugin implementations MUST preserve event order -- do not
+// buffer or dispatch calls concurrently in a way that could reorder them. For
+// example, an Upsert followed by a Delete for the same endpoint must arrive in
+// that order, or the endpoint will be incorrectly left in the datastore.
```
elevran, line 55: Add a note that notifier is not expected to be go routine safe? Hence the ordering requirement.

pkg/epp/framework/interface/datalayer/discovery.go
```diff
@@ -0,0 +1,53 @@
...
+type Notifier interface {
+	// Upsert adds or updates an endpoint in the datastore.
+	Upsert(endpoint *fwkdl.EndpointMetadata)
+	// Delete removes an endpoint from the datastore by its namespaced name.
+	Delete(id types.NamespacedName)
```
elevran, line 60: I wonder if this is not too k8s centric and we should include an explicit `UniqueID` in the endpoint metadata?
Not part of this PR.

pkg/epp/framework/interface/discovery/discovery.go
ezrasilvera: The requirement is here because without initial enumeration, the datastore starts empty and requests fail until the first event arrives. I will loosened from a hard requirement to a guideline

pkg/epp/framework/interface/discovery/discovery.go
ezrasilvera: Agree. Done.

pkg/epp/framework/interface/discovery/discovery.go
ezrasilvera: Agree. Added comment

pkg/epp/framework/interface/datalayer/discovery.go
ezrasilvera: Added

pkg/epp/framework/interface/datalayer/discovery.go
ezrasilvera: I agree. But it's wider than this instance.  several other places using NamespacedName  and we should switch to ID

ezrasilvera:
@elevran Thanks for the review. I addressed all comments.

pkg/epp/framework/interface/discovery/discovery.go
```diff
@@ -0,0 +1,64 @@
...
+	"k8s.io/apimachinery/pkg/types"
+
+	fwkdl "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/datalayer"
+	fwkplugin "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/plugin"
+)
+
+// DiscoveryPlugin discovers inference endpoints and drives their lifecycle in the datastore.
+// Implementations are registered in the plugin registry and selected via
+// EndpointPickerConfig.discovery.pluginRef.
+type DiscoveryPlugin interface {
```
ahg-g: We don't need to suffix it with a "Plugin"

ahg-g (commented):
Suggestion: 

1) we name it EndpointManager (in the spirit of naming plugins using agent nouns). 
2) place it under the datalayer acomponent. Which we defined as the layer that performs async data syncing.

ezrasilvera:
> 1. we name it EndpointManager (in the spirit of naming plugins using agent nouns).

EndpointManager implies it directly manages endpoints (deploy/remove them etc...).  How about EndpointDiscovery ?

ahg-g:
> > 1. we name it EndpointManager (in the spirit of naming plugins using agent nouns).
> >    EndpointManager implies it directly manages endpoints (deploy/remove them etc...).  How about EndpointDiscovery ?

Sounds good

ezrasilvera:
@ahg-g renamed to EndpointDiscovery. 

> Regarding the suggestion to place it under the datalayer component, I have some concerns:
> 
> 1. The datalayer and discovery have different responsibilities: the datalayer answers "what is the state of **known** endpoints?" (metrics, KV cache, queue depth).  Discovery answers "which endpoints exist?". These are fundamentally different concerns -- topology management vs gather endpoint data.
> 2. We aim to also support non k8s in the future so we need to keep the datalayer clean: we want K8s, file, Slurm, and Ray discovery all as plugins. If discovery lives inside the datalayer, the datalayer gets contaminated with infrastructure-specific code (ctrl.Manager, fsnotify, etc.),  which defeats the whole point of the abstraction.
> 3. Discovery is upstream of the datalayer, not a peer, the datalayer only activates per endpoint AFTER discovery creates it. Putting the thing that creates endpoints inside the thing that depends on those endpoints existing is an inversion. the fact that discovery triggers datalayer activity is expected and correct: any data pipeline has a source that feeds into a processor. That does not mean the source and the processor should live in the same component. Keeping them separate is what allows each to evolve independently
> 4. EndpointSource vs EndpointDiscovery are opposite in direction Same word "endpoint events", opposite flow.
>     EndpointSource:   datastore event --> outward --> consumers
>     EndpointDiscovery: outside world --> inward --> datastore
> 

@ahg-g @elevran After some more thinking, this will help in the plugin implementation for k8s. 
I moved the interface to under the datalayer.   With this I addressed all review comments

ezrasilvera:
@ahg-g @elevran After some more thinking,   I moved the interface to under the datalayer, this will also help in the plugin implementation for k8s.  With this I addressed all review comments. Thanks for your feedback

elevran:
/lgtm
will leave approval to @ahg-g

PR #915, opened 2026-04-28
add disk cleanup before e2e tests

### LINKED ISSUE #914: Add a disk-cleanup step on a CI runner before execution make test-e2e
I have observed a disk space issue on the GitHub Actions runner — no space left on device.

```
STEP: Loading ghcr.io/llm-d/llm-d-uds-tokenizer:dev into the cluster e2e-tests using docker @ 04/28/26 19:33:30.383
  ctr: failed to ingest "blobs/sha256/a37a0c04e996efed448d97610515ca94b0af0cef6c2b92aeffa850656a0aa743": failed to copy: failed to send write: write /var/lib/containerd/io.containerd.content.v1.content/ingest/1979762a8a9b300078d87f996b851d2302b3a25a88b97b89ced50d324901c0a0/data: no space left on device: unknown
```
I suggest adding a disk-cleanup step in `ci-pr-checks.yaml` before make test-e2e

```yaml
- name: Free disk space
  run: |
    sudo rm -rf /usr/local/lib/android
    sudo rm -rf /usr/share/dotnet
    docker image prune -af
```
The cleanup runs immediately before make test-e2e, after all the build/lint/test steps that need those tools. It frees ~17GB ( **around 25%**): Android SDK (~14GB), .NET (~2GB), and any dangling Docker layers.

roytman:
**What type of PR is this?**
/kind test

Prevents the "no space left on device" error during e2e tests

/Fixes #914

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -31,6 +31,7 @@ jobs:
               - Makefile*
               - go.mod
               - scripts/**
+              - .github/workflows/ci-pr-checks.yaml
   lint-and-test:
     needs: check-changes
     if: ${{ needs.check-changes.outputs.src == 'true' }}
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -181,10 +181,13 @@ jobs:
 
       - name: Free disk space
         run: |
+          df -h /
           sudo rm -rf /usr/local/lib/android
           sudo rm -rf /usr/share/dotnet
-          sudo rm -rf /opt/hostedtoolcache
           docker image prune -af
+          docker builder prune -af
+          df -h /
+
 
       - name: Run make test-e2e
         shell: bash
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -183,6 +183,10 @@ jobs:
         run: |
           df -h /
           sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /opt/microsoft
+          sudo rm -rf /usr/local/julia* 
+          sudo rm -rf /usr/local/share/powershel
+          sudo rm -rf /usr/share/swift
           sudo rm -rf /usr/share/dotnet
           docker image prune -af
           docker builder prune -af
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -40,6 +40,24 @@ jobs:
       - name: Checkout source
         uses: actions/checkout@v6
 
+      # Custom profile with directory-only removals (run in parallel, no apt).
+      # The cleanup-profile: max is much more aggressive, it cleanups around 41GB, 
+      # but runs 3.5 mins
+      - name: Free disk space
+        uses: justinthelaw/maximize-github-runner-space@v0.9.0
+        with:
+          cleanup-profile: custom
+          remove-android: 'true'
+          remove-dotnet: 'true'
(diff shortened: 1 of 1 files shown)


zetxqx (approved):

.github/workflows/ci-pr-checks.yaml
```diff
@@ -178,6 +179,16 @@ jobs:
...
           overwrite: true
 
+      - name: Free disk space
+        run: |
+          df -h /
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /usr/share/dotnet
+          docker image prune -af
+          docker builder prune -af
+          df -h /
```
revit13: Please also see https://github.com/actions/runner-images/issues/2840#issuecomment-790492173 for additional cleanups

roytman: thank you

.github/workflows/ci-pr-checks.yaml
```diff
@@ -178,6 +179,20 @@ jobs:
           retention-days: 400
           overwrite: true
 
+      - name: Free disk space
+        run: |
+          df -h /
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /opt/microsoft
+          sudo rm -rf /usr/local/julia* 
+          sudo rm -rf /usr/local/share/powershel
```
yankay: Nit: typo — `/usr/local/share/powershel` is missing an `l`, the actual directory is `/usr/local/share/powershell`. As-is the line silently no-ops (`rm -rf` on a non-existent path succeeds without error), so PowerShell isn't actually being freed.

ahg-g:
```suggestion
          sudo rm -rf /usr/local/share/powershell
```

ahg-g:
The e2e test took over 25min so far and still running, is that expected?

.github/workflows/ci-pr-checks.yaml
roytman: Thank you

roytman:
> The e2e test took over 25min so far and is still running, is that expected?

It took 28 mins. Yes, it takes a lot of time.

elevran:
there are several GH actions to clean / free disk space. Can you please see if it is possible to use one instead of manually maintaining the paths in the action?
For example https://github.com/easimon/maximize-build-space (I think we used another one action in the past, when python installs filled the disk).
Also, consider if we should be pruning the docker builder cache.

ahg-g:
DCO seems to also be stuck

elevran:
> DCO seems to also be stuck

Yes, seen across multiple repo's - likely a github issue.

zetxqx:
@elevran @ahg-g  is there any way we can bypass DCO check?

elevran:
If the outage is not resolved in the next 12h, will temporarily remove dco as a required check.
Will check tomorrow my morning

roytman:
> there are several GH actions to clean / free disk space. Can you please see if it is possible to use one instead of manually maintaining the paths in the action? For example https://github.com/easimon/maximize-build-space (I think we used another one action in the past, when python installs filled the disk). 

I replaced the rm commands with the [Maximize GitHub Runner Space](https://github.com/marketplace/actions/maximize-github-runner-space) action. Its custom profile frees ~30GB of disk space. The action can be much more aggressive with the `max` cleanup profile (frees 41 GB), but it takes 3.5 mins. 

> Also, consider if we should be pruning the docker builder cache.

It prunes the Docker builder cache.

roytman:
> If the outage is not resolved in the next 12h, will temporarily remove dco as a required check. Will check tomorrow my morning

DCO still doesn't work.

PR #916, opened 2026-04-29
fix: prefill cached tokens for NIXL decode usage

### LINKED ISSUE #1012: [Bug] PD Disaggregation: cached_tokens reported from Decoder side is misleading for billing
## Bug Description

In PD-disaggregated deployments, the final `usage.prompt_tokens_details.cached_tokens`
reported to the user comes from the **Decoder node's response**,  which equals to ISL, which kv-cache coming from Prefiller.

This value does not accurately reflect the true caching situation across the full P→D pipeline.

When we are billing the cached_tokens, this problem makes prefiller computing efforts all becoming "cached" wrongly.

Below content are co-authored by ClaudeCode

## Root Cause

When a request is processed in PD mode:

1. **Prefill node** processes the prompt and reports `prompt_tokens_details.cached_tokens`
   from its own prefix cache perspective (which may include disk/HBM cache hits).

2. **KV blocks are transferred** from Prefiller to Decoder.

3. **Decoder node** processes the continuation. When the Decoder consumes KV blocks
   transferred from the Prefiller, vLLM's `cached_tokens` metric on the Decoder side
   may report **nearly all prompt tokens as cached** — even though:
   - The Prefiller already did GPU computation for compute-miss tokens
   - The Prefiller's cache hit/miss breakdown is more accurate

4. The **final response** (from Decoder) overwrites the more accurate Prefiller
   `cached_tokens`, leading to incorrect billing.

## Billing Impact

**Scenario:**

```
Prompt: 200 tokens
- Prefiller: 80 tokens hit disk cache, 120 tokens compute miss (GPU ran)
- KV transferred to Decoder
- Decoder: 50 tokens hit local prefix cache

Current (broken):
  final cached_tokens = 50  ← only Decoder's local cache, ignores Prefiller's disk cache hit
  billing = 50 * cache_price + 150 * full_price

  Problem: Prefiller's 80 disk cache hit is not credited.
  Prefiller's 120 compute-miss was already charged at INPUT price.

Correct:
  cached_tokens should reflect what actually happened at prefill time,
  not the Decoder's perspective on transferred KV blocks.
```

## Expected Behavior

The final `cached_tokens` reported to the user should reflect the **Prefiller's**
cache situation, because:

1. Prefiller's `cached_tokens` accurately reflects prefill-stage cache hits
2. Prefiller already performed (and was charged for) GPU computation on compute-miss tokens
3. Decoder-side `cached_tokens` on transferred KV is misleading — it conflates
   "tokens whose KV was transferred from Prefiller" with "tokens that hit a cache"

## Proposed Fix

Within the **NIXL v2 sidecar / Inference Gateway** (no vLLM changes required):

1. **Intercept Prefiller response**: Extract `usage.prompt_tokens_details.cached_tokens`
   from the Prefiller response before KV transfer.

2. **Intercept Decoder response**: Compare with Decoder's `cached_tokens`.

3. **Prefer Prefiller's value**: If Prefiller reports `cached_tokens`, use that value
   in the final response instead of the Decoder's value.

4. **Handle both formats**: Support `prompt_tokens_details.cached_tokens` (OpenAI format)
   and `prompt_token_details.cached_tokens` (legacy/capitalization variant).

5. **Streamed responses**: Apply the same logic for SSE streamed usage chunks.

This is the approach taken by PR #916.

## Related Issues

- Related to #896
- Related to #906 (custom proxy handler extension points)

## Labels

bug, pd-disaggregation, billing

kebe7jun:
**What type of PR is this?**

/kind bug

In NIXL P/D mode, the sidecar currently returns the decoder's usage response directly. When the decoder consumes KV blocks transferred from the prefiller, vLLM may report nearly all prompt tokens as cached from the decoder perspective.

This updates the NIXL v2 sidecar flow to cached token usage reported by the prefiller. If the prefiller response includes `usage.prompt_tokens_details.cached_tokens`, the sidecar rewrites the decoder response usage with that value. If the prefiller does not report cached tokens, set to 0.

The rewrite supports both JSON responses and streamed SSE usage chunks.

Related to #896 and #906.

This PR implements the built-in NIXL v2 sidecar response usage correction for the default path, while #906 adds custom proxy handler extension points for similar accounting needs.

Fixes #1012

```release-note
Use prefiller cached token usage for NIXL P/D responses.
```

### DIFF
--- a/pkg/sidecar/proxy/cached_tokens_usage_rewriter.go
+++ b/pkg/sidecar/proxy/cached_tokens_usage_rewriter.go
@@ -39,18 +39,19 @@ type cachedTokensUsageRewriter struct {
 // See: https://platform.openai.com/docs/guides/prompt-caching
 const promptTokensDetailsField = "prompt_tokens_details"
 
-func newCachedTokensResponseWriter(w http.ResponseWriter, cachedTokens int, enabled bool) http.ResponseWriter {
-	if !enabled {
-		// Prefill did not report cached_tokens, so keep the decoder response intact.
-		return w
-	}
+func newCachedTokensResponseWriter(w http.ResponseWriter, cachedTokens int) http.ResponseWriter {
+	writer, _ := newCachedTokensResponseWriterWithFinalize(w, cachedTokens)
+	return writer
+}
+
+func newCachedTokensResponseWriterWithFinalize(w http.ResponseWriter, cachedTokens int) (http.ResponseWriter, func() error) {
 	rewriter := &cachedTokensUsageRewriter{
 		header:       w.Header(),
 		cachedTokens: cachedTokens,
 	}
 	// httpsnoop preserves optional ResponseWriter interfaces such as
 	// http.Flusher, http.Hijacker, http.Pusher, and io.ReaderFrom.
-	return httpsnoop.Wrap(w, httpsnoop.Hooks{
+	writer := httpsnoop.Wrap(w, httpsnoop.Hooks{
 		WriteHeader: func(next httpsnoop.WriteHeaderFunc) httpsnoop.WriteHeaderFunc {
 			return func(statusCode int) {
 				rewriter.writeHeader(next, statusCode)
@@ -68,6 +69,9 @@ func newCachedTokensResponseWriter(w http.ResponseWriter, cachedTokens int, enab
 			}
 		},
 	})
+	return writer, func() error {
+		return rewriter.flushSSEBuffer(w.Write)
+	}
 }
 
 func (r *cachedTokensUsageRewriter) writeHeader(next httpsnoop.WriteHeaderFunc, statusCode int) {
--- a/pkg/sidecar/proxy/cached_tokens_usage_rewriter.go
+++ b/pkg/sidecar/proxy/cached_tokens_usage_rewriter.go
@@ -1,0 +1,308 @@
+/*
+Copyright 2025 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package proxy
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/felixge/httpsnoop"
+)
+
+type cachedTokensUsageRewriter struct {
+	header       http.Header
+	cachedTokens int
+	wroteHeader  bool
+	streaming    bool
+	streamBuffer []byte
+}
+
+// OpenAI-compatible chat usage reports prompt cache hits at
+// usage.prompt_tokens_details.cached_tokens.
+// See: https://platform.openai.com/docs/guides/prompt-caching
+const promptTokensDetailsField = "prompt_tokens_details"
+
+func newCachedTokensResponseWriter(w http.ResponseWriter, cachedTokens int, enabled bool) http.ResponseWriter {
+	if !enabled {
+		// Prefill did not report cached_tokens, so keep the decoder response intact.
+		return w
+	}
+	rewriter := &cachedTokensUsageRewriter{
+		header:       w.Header(),
+		cachedTokens: cachedTokens,
+	}
+	// httpsnoop preserves optional ResponseWriter interfaces such as
+	// http.Flusher, http.Hijacker, http.Pusher, and io.ReaderFrom.
+	return httpsnoop.Wrap(w, httpsnoop.Hooks{
+		WriteHeader: func(next httpsnoop.WriteHeaderFunc) httpsnoop.WriteHeaderFunc {
+			return func(statusCode int) {
+				rewriter.writeHeader(next, statusCode)
+			}
+		},
+		Write: func(next httpsnoop.WriteFunc) httpsnoop.WriteFunc {
+			return func(body []byte) (int, error) {
+				return rewriter.write(next, body)
+			}
+		},
+		ReadFrom: func(_ httpsnoop.ReadFromFunc) httpsnoop.ReadFromFunc {
+			return func(src io.Reader) (int64, error) {
+				// ReadFrom would otherwise bypass Write and skip usage rewriting.
+				return rewriter.readFrom(w.Write, src)
+			}
+		},
+	})
+}
+
+func (r *cachedTokensUsageRewriter) writeHeader(next httpsnoop.WriteHeaderFunc, statusCode int) {
+	// Rewriting may change the body size, so any upstream Content-Length is stale.
+	r.header.Del("Content-Length")
+	r.wroteHeader = true
+	next(statusCode)
+}
+
+func (r *cachedTokensUsageRewriter) write(next httpsnoop.WriteFunc, body []byte) (int, error) {
+	updated := r.rewrite(body)
+	if !r.wroteHeader {
+		r.header.Del("Content-Length")
+	}
+	n, err := next(updated)
+	if err != nil {
+		return n, err
+	}
+	return len(body), nil
+}
+
+func (r *cachedTokensUsageRewriter) readFrom(next httpsnoop.WriteFunc, src io.Reader) (int64, error) {
+	if r.isSSE(nil) {
+		// SSE can be long-lived, so keep it streaming and rewrite complete lines.
+		n, err := io.Copy(cachedTokensStreamWriter{
+			write:   r.write,
+			forward: next,
+		}, src)
+		if err != nil {
+			return n, err
+		}
+		if err := r.flushSSEBuffer(next); err != nil {
+			return n, err
+		}
+		return n, nil
+	}
+
+	// Non-streaming JSON must be complete before we can safely rewrite usage.
+	body, err := io.ReadAll(src)
+	if err != nil {
+		return 0, err
+	}
+	_, err = r.write(next, body)
+	if err != nil {
+		return 0, err
+	}
+	return int64(len(body)), nil
+}
+
+func (r *cachedTokensUsageRewriter) rewrite(body []byte) []byte {
+	if r.isSSE(body) {
+		return r.rewriteSSEChunk(body)
+	}
+	return replaceCachedTokens(body, r.cachedTokens)
+}
+
+type cachedTokensStreamWriter struct {
+	write   func(httpsnoop.WriteFunc, []byte) (int, error)
+	forward httpsnoop.WriteFunc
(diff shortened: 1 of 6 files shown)


pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -281,3 +283,174 @@ func (s *Server) runNIXLProtocolV2(w http.ResponseWriter, r *http.Request, prefi
...
+	changed := false
+	processed := false
+
+	for _, line := range lines {
+		if len(line) == 0 {
+			continue
+		}
+		trimmedLine := bytes.TrimRight(line, "\r\n")
+		lineEnding := line[len(trimmedLine):]
+		data, ok := bytes.CutPrefix(trimmedLine, []byte("data: "))
```
hexfusion: question: would this cause an issue if upstream splits a chunk mid event? in this case we wold miss the data prefix. an option could be buffer to newline as a possible regression hedge.

kebe7jun: Thank you for pointing out the problem. I’ve fixed this.

pkg/sidecar/proxy/cached_tokens_usage_rewriter.go
```diff
@@ -0,0 +1,289 @@
...
+	cachedTokens int
+	wroteHeader  bool
+	streaming    bool
+	streamBuffer []byte
+}
+
+// OpenAI-compatible chat usage reports prompt cache hits at
+// usage.prompt_tokens_details.cached_tokens.
+// See: https://platform.openai.com/docs/guides/prompt-caching
+const promptTokensDetailsField = "prompt_tokens_details"
```
elevran, line 40: The PR description says this handles both `usage.prompt_tokens_details.cached_tokens` (plural) _and_ `usage.prompt_token_details.cached_tokens` (singular). The code only defines and checks the plural form.

If `prompt_token_details` (singular) is a real field variant in some vLLM builds, the code needs to handle it. If it's a PR description typo, please correct the description.

pkg/sidecar/proxy/cached_tokens_usage_rewriter.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+func (r *cachedTokensUsageRewriter) write(next httpsnoop.WriteFunc, body []byte) (int, error) {
+	updated := r.rewrite(body)
+	if !r.wroteHeader {
+		r.header.Del("Content-Length")
+	}
+	_, err := next(updated)
+	if err != nil {
+		return 0, err
```
elevran: Suggestion: when `next(updated)` fails, returning `0` misrepresents how many input bytes were consumed. Returning the actual `n` from `next` is more accurate:

pkg/sidecar/proxy/cached_tokens_usage_rewriter.go
```diff
@@ -0,0 +1,289 @@
...
+		lineEnd := bytes.IndexByte(r.streamBuffer, '\n')
+		if lineEnd < 0 {
+			break
+		}
+		// Upstream may split chunks mid-event; only rewrite complete SSE lines.
+		line := r.streamBuffer[:lineEnd+1]
+		replacedLine, _ := replaceCachedTokensSSELine(line, r.cachedTokens)
+		updated = append(updated, replacedLine...)
+		r.streamBuffer = r.streamBuffer[lineEnd+1:]
+	}
```
elevran, line 166: **suggestion**:
If the upstream closes before sending a final `\n` (e.g., truncated connection, server-side bug), bytes in `r.streamBuffer` are silently dropped. While well-formed vLLM SSE always ends with `data: [DONE]\n\n`, so this is benign on the happy path, might be worth guarding and flushing as well.

pkg/sidecar/proxy/cached_tokens_usage_rewriter_test.go
```diff
@@ -0,0 +1,176 @@
...
+		updated := replaceCachedTokens(body, 7)
+
+		var response map[string]any
+		Expect(json.Unmarshal(updated, &response)).To(Succeed())
+		usage := response["usage"].(map[string]any)
+		details := usage["prompt_tokens_details"].(map[string]any)
+		Expect(details["cached_tokens"]).To(BeNumerically("==", 7))
+	})
+
+	It("should not extract cached tokens when prefill response has none", func() {
```
elevran, line 84: This test only covers the case where the prefiller has no usage field at all. The case where the prefiller explicitly reports `"cached_tokens": 0` is not tested. `extractCachedTokens` returns `(0, true)` for that input, so the rewriter activates and overwrites the decoder's non-zero value, which is the correct behaviour. A regression that collapsed `(0, true)` to the disabled path would go undetected.

Please add a test to confirm extract zero cached tokens when prefiller reports zero, and an integration test for prefiller with `cached_tokens:0`, decoder with `cached_tokens:<number>`, asserting output has `cached_tokens:0`.

pkg/sidecar/proxy/connector_nixlv2_test.go
```diff
@@ -49,22 +60,84 @@ var _ = Describe("NIXL Connector (v2)", func() {
...
+		defer rp.Body.Close()
+
+		responseBody, err := io.ReadAll(rp.Body)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(rp.StatusCode).To(Equal(http.StatusOK), string(responseBody))
+		Expect(rp.Header.Get("Content-Type")).To(ContainSubstring(eventStreamContentType))
+		return string(responseBody)
+	}
+
+	cachedTokensFromResponse := func(response map[string]any) float64 {
```
elevran, line 115: nit: there are three consecutive naked type assertions on `usage`, `details` and `cachedTokens` will `panic` if any field is absent or mis-typed (e.g., a typo in a `RawResponse` fixture). Replace with "field, ok := "  + `Expect(ok)`

pkg/sidecar/proxy/connector_nixlv2_test.go
```diff
@@ -49,22 +60,84 @@ var _ = Describe("NIXL Connector (v2)", func() {
...
+
+			testInfo.proxy.allowlistValidator = &AllowlistValidator{enabled: false}
+			err := testInfo.proxy.Start(testInfo.ctx)
+			Expect(err).ToNot(HaveOccurred())
+
+			testInfo.stoppedCh <- struct{}{}
+		}()
+
+		<-testInfo.proxy.readyCh
+		proxyBaseAddr := "http://" + testInfo.proxy.addr.String()
```
elevran: test was moved here but still duplicates the proxy-start goroutine inline and calls `testInfo.cancelFn()` manually at the end, while the six new tests added in this PR all use the `startProxy()` helper with `DeferCleanup`. Consider refactoring this test to use `startProxy()` for consistency within the same `Describe` block.

test/sidecar/mock/chat_completions_handler.go
```diff
@@ -148,11 +150,20 @@ func (cc *ChatCompletionHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
...
 	var completionResponse map[string]any
-	if err := json.Unmarshal([]byte(rawResponse), &completionResponse); err != nil {
-		w.WriteHeader(http.StatusBadRequest)
-		w.Write([]byte(err.Error())) //nolint:all
-		return
+	if cc.RawResponseType != "text/event-stream" {
+		if err := json.Unmarshal([]byte(rawResponse), &completionResponse); err != nil {
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte(err.Error())) //nolint:all
+			return
```
elevran, line 167: nit: 
When `RawResponseType == "text/event-stream"`, the `json.Unmarshal` block is skipped and `completionResponse` stays `nil`. The `CompletionResponses = append(…, completionResponse)` call further down stores `nil` in the slice, possibly leading to a future test reading `decodeHandler.CompletionResponses[0]` for an SSE scenario and getting a nil map and `panic`.

Consider skipping the append for streaming responses.

elevran (commented):
overall looks good - thanks @kebe7jun!
A few suggestions and nits (mostly inlined as comments).

There is a discrepancy between the PR description and the implementation around the singular `prompt_token_details` field name but the code handles the plural (`prompt_tokens_details`). As far as I can tell the PR description is a typo.

Can you please indicate `NONE` in the PR description or set an explicit issue in `Fixes #`?

pkg/sidecar/proxy/cached_tokens_usage_rewriter.go
kebe7jun: I had collected incorrect info earlier, after careful investigation, I found that there is no version with `prompt_token_details`, so I removed it, and updated the description.

yankay:
Hi @vMaroon, when you get a chance, could you take a look? Thanks!

yankay (commented):
Technical review notes on `cachedTokensUsageRewriter`: `Write` error-path semantics vs. `io.Writer` expectations, and SSE line buffering asymmetry between `ReadFrom` and chunked `Write` (upstream reverse proxy behavior).

pkg/sidecar/proxy/cached_tokens_usage_rewriter.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+func (r *cachedTokensUsageRewriter) write(next httpsnoop.WriteFunc, body []byte) (int, error) {
+	updated := r.rewrite(body)
+	if !r.wroteHeader {
+		r.header.Del("Content-Length")
+	}
+	n, err := next(updated)
+	if err != nil {
+		return n, err
```
yankay, line 91: Bug 1: current `write` semantics are OK (`0, err` on failure, `io.ErrShortWrite` on short write), so no code change needed here.

Follow-up commit: https://github.com/yankay/llm-d-inference-scheduler/commit/3c828284062e4271fa29e7854760e5c86f7ab10a

pkg/sidecar/proxy/cached_tokens_usage_rewriter.go
```diff
@@ -0,0 +1,304 @@
...
+		if lineEnd < 0 {
+			break
+		}
+		// Upstream may split chunks mid-event; only rewrite complete SSE lines.
+		line := r.streamBuffer[:lineEnd+1]
+		replacedLine, _ := replaceCachedTokensSSELine(line, r.cachedTokens)
+		updated = append(updated, replacedLine...)
+		r.streamBuffer = r.streamBuffer[lineEnd+1:]
+	}
+	return updated
```
yankay, line 167: Bug 2: `Write` path could leave trailing SSE bytes in `streamBuffer`; fixed by finalizing the cached-token writer after decode.

Follow-up commit: https://github.com/yankay/llm-d-inference-scheduler/commit/3c828284062e4271fa29e7854760e5c86f7ab10a

pkg/sidecar/proxy/cached_tokens_usage_rewriter.go
kebe7jun: See https://github.com/llm-d/llm-d-inference-scheduler/pull/916#discussion_r3215017286

elevran:
@kebe7jun kindly please fix the linter error so this can be merged 🙏

kebe7jun:
@elevran fixed.

PR #933, opened 2026-04-30
add promptTokens-based PD decision to prefix-based-pd-decider

learner0810:
**What type of PR is this?**

/kind feature

This PR introduces a prompt-length-based PD decider that enables P/D disaggregation when the estimated prompt token length reaches a configured threshold.

This provides a simple threshold-based decision strategy for scenarios where P/D disaggregation should be determined solely by prompt length, without depending on additional cache-related signals.

As an alternative, this logic could also be incorporated into the existing `prefix-based-pd-decider` plugin if we prefer to consolidate PD decision logic in a single plugin.

Fixes #

```release-note
Added promptTokens support to prefix-based-pd-decider so P/D disaggregation now requires the request prompt to meet a minimum estimated token-length threshold before prefix-cache-based logic is applied. Defining promptTokens as 0 disables this guard.
```


shmuelk:
The code looks good.

Please update the PR description to add relase notes. You are adding a new disambiguation decider. This is a User facing change.

It also needs to be documented, I just can't tell you exactly where to do so, as the the documentation is in flux.

roytman:
I'd vote to incorporate into the existing `prefix-based-pd-decider plugin`. 

- It will simplify users' configuration and understanding; all related functionality is concentrated in a single plugin. Therefore, we have a trend to reduce the number of plugins. (e.g. #833)
- Some of the suggested plugin functionality lives in `prefix_based_pd_decider.go`, e.g. `getUserInputLenInTokens()` and `makeRequestWithTokens`

We have not finalized our documentation, but in the meantime, please add this plugin to the architecture.md (see [prefix-based-pd-decider plugin](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/docs/architecture.md#prefix-based-decider-plugin))  

There is no e2e test and EPP config example. 
BTW, combining the plugins will simplify the documentation and e2e tests.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prompt_length_based_pd_decider.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+const (
+	// PromptLengthBasedPDDeciderPluginType is the type-name of the promptLengthBasedPDDecider plugin.
+	PromptLengthBasedPDDeciderPluginType = "prompt-length-based-pd-decider"
+)
+
+// PromptLengthBasedPDDeciderConfig holds the configuration for the promptLengthBasedPDDecider plugin.
+type PromptLengthBasedPDDeciderConfig struct {
+	// PromptTokens minimum prompt tokens that triggers disaggregated PD.
```
roytman: nit: Token count is an estimate, but the config name implies exactness. The name `promptTokens` doesn't indicate it's approximate, which will mislead operators tuning the threshold. At a minimum, the config comment should say "estimated."

```suggestion
	// PromptTokens is the minimum estimated prompt length in tokens (approximated
    // as character-count / 4) that triggers disaggregated PD. 
    // Zero disables disaggregation.
    PromptTokens int `json:"promptTokens"`
```

vMaroon:
Thanks @learner0810 - I agree with @roytman and the proposed alternative in the PR description. I think this PR should pivot to that.

learner0810:
> Thanks @learner0810 - I agree with @roytman and the proposed alternative in the PR description. I think this PR should pivot to that.

Thanks everyone for the reviews. I’ll integrate the changes into the existing  prefix-based-pd-decider  plugin and submit the revised version shortly.

learner0810:
> Thanks @learner0810 - I agree with @roytman and the proposed alternative in the PR description. I think this PR should pivot to that.

Thanks, I’ve updated the PR accordingly. Could you please take another look?

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -27,13 +27,22 @@ const (
 type PrefixBasedPDDeciderConfig struct {
 	// NonCachedTokens non cached minimum tokens that triggers disaggregated PD
 	NonCachedTokens int `json:"nonCachedTokens"`
+
+	// PromptTokens is the minimum estimated prompt length in tokens (approximated
+	// as character-count / 4) that triggers disaggregated PD.
+	// Zero disables disaggregation.
```
shmuelk: I assume that you are adding an extra decision point, the prompt length is checked in addition to any other attributes/information that is checked in this decider.

If I am correct, then specifying zero for PromptTokens, which is the default, should disable the prompt length check.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -79,6 +89,10 @@ func NewPrefixBasedPDDecider(config PrefixBasedPDDeciderConfig) (*PrefixBasedPDD
 		log.Log.Info("Prefix-based PD disabled (NonCachedTokens=0)")
 	}
 
+	if config.PromptTokens == 0 {
+		log.Log.Info("Prefix-based PD disabled (PromptTokens=0)")
+	}
```
shmuelk: Again, I assume that you are adding an extra decision point, the prompt length is checked in addition to any other attributes/information that is checked in this decider.

If I am correct, then specifying zero for PromptTokens, which is the default, should disable the prompt length check.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -105,6 +119,13 @@ func (d *PrefixBasedPDDecider) disaggregate(ctx context.Context, request *schedu
 	if d.config.NonCachedTokens == 0 {
 		return false
 	}
+
+	// PromptTokens is the minimum prompt length required to enable
+	// disaggregated PD. A value of 0 disables the decider.
+	if d.config.PromptTokens == 0 {
+		return false
+	}
```
shmuelk: See my comment above

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -114,6 +135,12 @@ func (d *PrefixBasedPDDecider) disaggregate(ctx context.Context, request *schedu
 		logger.Error(err, "prefix decider: failed to get user input length in tokens")
 		return false
 	}
+
+	if inputTokens < d.config.PromptTokens {
+		debugLogger.Info("Input is shorter than the promptTokens, no disaggregated PD")
+		return false
+	}
```
shmuelk: See my comment above

shmuelk (changes requested):

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
learner0810: Thanks for the review. I’ve updated the PR based on your comments. Could you please take another look when you get a chance?

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
learner0810: ditto

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
learner0810: ditto

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
learner0810: ditto

learner0810:
Gentle ping @shmuelk @vMaroon  @roytman  for review when you get a chance.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
roytman: Actually, `PromptTokens` equals to zero does not disable disaggregation; it disables this guard. Disaggregation can still happen via the prefix-cache path. Should read something like:
```golang
// Zero disables this threshold check (prompt-length gate is not applied). 
```

roytman (commented):
Both NonCachedTokens and PromptTokens operate on inputTokens but at different conceptual levels:

- PromptTokens: total prompt must be long enough to bother disaggregating
- NonCachedTokens: the non-cached suffix must be long enough

The ordering matters (promptTokens is checked first). A brief comment near the check or in the config docs explaining the evaluation order would help operators understand what combination of values to set.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
learner0810: Thanks for the review. I’ve updated the PR based on your comments. Could you please take another look when you get a chance?

roytman:
> Both NonCachedTokens and PromptTokens operate on inputTokens but at different conceptual levels:
> 
>     * PromptTokens: total prompt must be long enough to bother disaggregating
> 
>     * NonCachedTokens: the non-cached suffix must be long enough
> 
> 
> The ordering matters (promptTokens is checked first). A brief comment near the check or in the config docs explaining the evaluation order would help operators understand what combination of values to set.

Hi @learner0810 , thank you for the fixes, just asmall nit:
In the Release Notes, you wrote "Added promptTokens support to prefix-based-pd-decider so P/D disaggregation now requires the request prompt to meet a minimum estimated token-length threshold before prefix-cache-based logic is applied." I would add that defining `promptTokens to 0` disables this guard.

I'm OK with the rest, by @shmuelk requested changes, and he should approve.

learner0810:
> > Both NonCachedTokens and PromptTokens operate on inputTokens but at different conceptual levels:
> > ```
> > * PromptTokens: total prompt must be long enough to bother disaggregating
> > 
> > * NonCachedTokens: the non-cached suffix must be long enough
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The ordering matters (promptTokens is checked first). A brief comment near the check or in the config docs explaining the evaluation order would help operators understand what combination of values to set.
> 
> Hi @learner0810 , thank you for the fixes, just asmall nit: In the Release Notes, you wrote "Added promptTokens support to prefix-based-pd-decider so P/D disaggregation now requires the request prompt to meet a minimum estimated token-length threshold before prefix-cache-based logic is applied." I would add that defining `promptTokens to 0` disables this guard.
> 
> I'm OK with the rest, by @shmuelk requested changes, and he should approve.

Thanks for the review. I’ve updated the release notes to add: “Defining promptTokens as 0 disables this guard.”

@shmuelk Could you please take another look when you have a chance?

learner0810:
Hi @shmuelk and @roytman, could you please review this when you have a chance?

roytman:
Hi @ learner0810, sorry for the late response. 
In addition to the Copilot comments, `docs/disaggregation.md` lists the parameters for `prefix-based-pd-decider` (currently only `nonCachedTokens`) but does not mention the new promptTokens. Please add it there, including that 0 disables the gate and that it is evaluated before the `nonCachedTokens` / prefix-cache logic (this is also the evaluation-order point raised in the inline discussion).

learner0810:
> Hi @ learner0810, sorry for the late response.
> In addition to the Copilot comments, `docs/disaggregation.md` lists the parameters for `prefix-based-pd-decider` (currently only `nonCachedTokens`) but does not mention the new promptTokens. Please add it there, including that 0 disables the gate and that it is evaluated before the `nonCachedTokens` / prefix-cache logic (this is also the evaluation-order point raised in the inline discussion).

Thanks for the clarification. I’ve updated docs/disaggregation.md to document the new promptTokens parameter for prefix-based-pd-decider, including that 0 disables the gate and that it is evaluated before the nonCachedTokens / prefix-cache logic.

roytman:
Thank you for your PR 
/lgtm

ahg-g (approved):

PR #942, opened 2026-04-30
feat(datalayer): plugin-driven source/extractor registration via Registrant interface

### LINKED ISSUE #963: [feat] Allow Plugins to register data layer extractors in code, without needing configuration changes as well
Ability to register data source and extractors via code, without depending on config file content.
Add methods to datalayer.Runtime, and make it available to Plugins via (e.g.,) the Handle struct.

**Why is this needed**:
Per https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2749#discussion_r3033350234:

> @ahg-g: I think extractors don't necessarily need to be configured via the config api. In some cases extractors will live with a sophisticated plugin (like a data producer one). I wonder if we should allow a plugin to register an extractor via code, perhaps we offer an interface in the framework handle that a data producer plugin can call to register its extractor.
> @elevran: It would need to register a data source as well. Can easily add register source / extractor to the runtime.
> @ahg-g: I think registering the source via config is fine since it is common across multiple extractors. A plugin should fail its instantiation if it tries to register an extractor that depends on an unregistered source.
> @elevran: This would mix different aspects: a user configuration with plugin code dependency. The user is not expected to know which data source a (e.g.) scheduling scorer plugin depends on.
Making extractor registration available through code maks perfect sense to me, but I would complement it with source registration (or at least a source presence verification so the plugin can issue a clear log)

xref https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2784

elevran:
**What type of PR is this?**
/kind feature

Introduce the `Registrar`/`Registrant` interfaces so plugins can declare their datalayer dependencies at init time rather than hard-coding them in configuration.

- `fwkdl.Registrar` / `fwkdl.Registrant` / `PendingRegistration` live in `pkg/epp/framework/interface/datalayer/registrar.go`
- `Runtime` implements `Registrar`; `Configure()` resolves pending registrations after processing user-config sources: wires extractors to matching sources, auto-registers a `DefaultSource` when absent, respects Fail/Warn policy, and deduplicates by extractor type
- Runner iterates all plugins and calls `RegisterDependencies` on any that implement `Registrant`, immediately before `Runtime.Configure()`
- `InFlightLoadProducer` is the first adopter: it registers itself as an `EndpointExtractor` against the endpoint-notification source (auto-creating the source when absent from user config) so endpoint-deletion events deterministically clean up its concurrency trackers

**Depends on**: #941 — the auto-created endpoint source is only useful when `ensureDataLayer()` is additive (so the metrics source is not silently dropped). Please merge #941 first.

Fixes #963 

```release-note
NONE
```
Deferred registrations are only allowed in code and are not a user exposed feature.

### DIFF
--- a/pkg/epp/datalayer/runtime.go
+++ b/pkg/epp/datalayer/runtime.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	"github.com/llm-d/llm-d-inference-scheduler/pkg/common/observability/logging"
@@ -39,6 +40,9 @@ type Runtime struct {
 	notifiers        sync.Map // Map of k8s notification sources (key=source name, value=NotificationSource)
 	endpointSources  sync.Map // Map of endpoint sources (key=source name, value=EndpointSource)
 	sourceExtractors sync.Map // Map sources to extractors (key=source name. value=[]Extractor)
+
+	pendingMu            sync.Mutex
+	pendingRegistrations []fwkdl.PendingRegistration // code-registered (source-type, extractor) pairs, resolved by Configure()
 
 	collectors sync.Map    // Per-endpoint poller (key=namespaced name, value=*Collector)
 	logger     logr.Logger // Set in Configure; used where no context is available (e.g. ReleaseEndpoint).
@@ -64,7 +68,8 @@ func NewRuntime(pollingInterval time.Duration) *Runtime {
 // Configure is called to transform the configuration information into the Runtime's
 // internal fields.
 func (r *Runtime) Configure(cfg *Config, enableNewMetrics bool, disallowedExtractorType string, logger logr.Logger) error {
-	if cfg == nil || len(cfg.Sources) == 0 {
+	hasPending := len(r.pendingRegistrations) > 0
+	if (cfg == nil || len(cfg.Sources) == 0) && !hasPending {
 		if enableNewMetrics {
 			return errors.New("data layer enabled but no data sources configured")
 		}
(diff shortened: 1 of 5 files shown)


pkg/epp/datalayer/runtime.go
```diff
@@ -40,6 +41,8 @@ type Runtime struct {
 	endpointSources  sync.Map // Map of endpoint sources (key=source name, value=EndpointSource)
 	sourceExtractors sync.Map // Map sources to extractors (key=source name. value=[]Extractor)
 
+	pendingRegistrations sync.Map // Pending code-registered (source-type, extractor) pairs (key="owner/sourceType", value=fwkdl.PendingRegistration)
```
shmuelk: Only one pending extrator per source is allowed?

pkg/epp/datalayer/runtime.go
```diff
@@ -64,64 +67,211 @@ func NewRuntime(pollingInterval time.Duration) *Runtime {
...
 }
 
+// Register stores a pending source/extractor dependency declared by a plugin.
+// Called by plugins implementing Registrant.RegisterDependencies() before Configure() runs.
+func (r *Runtime) Register(reg fwkdl.PendingRegistration) error {
+	if reg.Extractor == nil {
+		return fmt.Errorf("plugin %s: PendingRegistration.Extractor must not be nil", reg.Owner)
+	}
+	key := reg.Owner.String() + "/" + reg.SourceType
+	r.pendingRegistrations.Store(key, reg)
```
shmuelk: Only one pending extrator per source is allowed?

pkg/epp/datalayer/runtime.go
elevran: no, missed it. Changed to support multiple

pkg/epp/datalayer/runtime.go
```diff
@@ -40,6 +41,9 @@ type Runtime struct {
 	endpointSources  sync.Map // Map of endpoint sources (key=source name, value=EndpointSource)
 	sourceExtractors sync.Map // Map sources to extractors (key=source name. value=[]Extractor)
 
+	pendingMu            sync.Mutex
```
shmuelk, line 44: Doesn't the initialization and setup logic all run in the same goroutine?

Why the mutex?

elevran: Provide additional guarantee if that ever changes. The mutex overhead, when uncontended, is very low. Single go routine means no contention

shmuelk: ok

PR #944, opened 2026-04-30
feature: add slo metrics to flowcontrol

### LINKED ISSUE #945: [Flow Control] Add SLO values to flow control metrics
If SLO values are provided (e.g. through request headers), it should be added to the flow control metrics. This can be used by the autoscaler.

The issue was initially opened on the GIE repo. See https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2532. It needs to be re-opened in this repo following the migration of EPP.

loicmarchal:
**What type of PR is this?**
/kind feature

Adds the following metrics: 
* request duration per SLO TTFT class
* adds SLO TTFT class to `queue size` and `queue bytes` metrics 
* counter for incoming requests per SLO class.

The TTFT SLO is passed with the request in the headers (`x-slo-ttft-ms`).

These metrics will allow tracking the percentage of requests with a duration below or above their associated SLO class as well as the number of incoming and queued requests per SLO class.

_The SLO classes might need to have smaller intervals (currently the interval is 200 ms), but it would increase cardinality of the Prometheus histogram_

Fixes #945 

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```


loicmarchal:
@kfswain I reopened the PR in this repo. We still have the ongoing discussion about defining the SLO classes for the metrics in the InferenceObjective (https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2685#discussion_r2993501190).

One simple approach could be to simply add an SLO class label on the InferenceObjective, but it ties the InferenceObjective to the SLO ms target, and it relies on the user setting the right headers (`x-gateway-inference-objective` and `x-slo-ttft-ms`)

```yaml
apiVersion: inference.networking.x-k8s.io/v1alpha2
kind: InferenceObjective
metadata:
  name: realtime
spec:
  poolRef:
    name: my-pool
  priority: 10
  sloClass: fast
```

We could add SLO class bucket min and max values, and check if the value in `x-slo-ttft-ms` is within the bucket defined in the InferenceObjective

```yaml
apiVersion: inference.networking.x-k8s.io/v1alpha2
kind: InferenceObjective
metadata:
  name: realtime
spec:
  poolRef:
    name: my-pool
  priority: 10
  sloClass:
    label: fast
    minMs: 0
    maxMs: 100
```

But what to do if it's not within the bucket? Send the metrics with the wrong label anyway? Fix the label by finding the right label from another InferenceObjective with the right min/max? Consider the request has the wrong InferenceObjective given the SLO deadline defined in `x-slo-ttft-ms`? We also need to validate the InferenceObjectives, as SLO class buckets should not overlap?

If we enable the SLO deadline request ordering, requests will be effectively prioritized based on the values defined in the `x-slo-ttft-ms` header. So it is in fact an InferenceObjective?

What's your take on this? Thank you!

Another question: the InferenceObjective type is not yet migrated from GIE. Should we migrate it in this repo?

loicmarchal:
I have moved the SLO metrics range definition to the InferenceObjective. Here is a summary of the design choices.

The SLO class seems semantically tied to the objective's contract: a high-priority `realtime` objective implies tight TTFT deadlines, and a low-priority `batch` objective implies loose ones. A request shouldn't realistically declare a `realtime` objective with a 500ms deadline.

__InferenceObjective name = SLO class label__

The InferenceObjective name (e.g., `realtime`, `batch`) is used directly as the SLO class label in metrics. No separate `sloClass.label` field is needed.

__Per-metric ranges__

The spec uses a nested structure with separate TTFT and TPOT ranges, since both `x-slo-ttft-ms` and `x-slo-tpot-ms` headers already exist in the system:

```yaml
spec:
  slo:
    ttft:
      minMs: 0
      maxMs: 100
    tpot:
      maxMs: 20
```

This avoids a breaking API change when TPOT support is needed later.

__Ranges are optional and can be open-ended__

- slo is optional — objectives without it still work for priority-based scheduling (SLO class defaults to "none" in metrics).
- minMs and maxMs are independently optional: omitting minMs implies no lower bound (0), omitting maxMs implies no upper bound.
- Semantics are [minMs, maxMs) (inclusive lower, exclusive upper).

__Overlapping ranges are allowed__

The SLO class is determined by the objective name (set via the x-gateway-inference-objective header), not by matching the TTFT value against ranges. So overlapping ranges between objectives don't cause ambiguity. The range is a validation hint, not a classification key. This avoids the complexity of admission-time overlap checking.

__Warning logs for mismatches__

When a request's x-slo-ttft-ms (or x-slo-tpot-ms) value falls outside the range defined on its target InferenceObjective, a warning is logged. The request is still processed normally — the mismatch indicates a client misconfiguration, not a rejection condition.

@kfswain does it align with what you had in mind?

pkg/epp/flowcontrol/controller/internal/item.go
```diff
@@ -159,16 +161,33 @@ func (fi *FlowItem) finalizeInternal(outcome types.QueueOutcome, err error) {
...
 }
 
+func extractHeader(req flowcontrol.FlowControlRequest, name string) string {
+	infReq := req.InferenceRequest()
+	if infReq == nil || infReq.Headers == nil {
+		return ""
+	}
+	return request.GetHeader(infReq.Headers, name)
+}
+
```
shmuelk: I don't think this function is needed.

1. req.InferenceRequest() should never be nil
2. Even if req.InferenceRequest().Headers is nil, request.Headers will not crash. If one does `val, ok := m[k]`, where m is map will a nil value, ok will be set to false, correctly

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -331,6 +342,13 @@ func (fc *FlowController) withConnectionWithFallback(
 		return fn(conn, fallback)
 	})
 }
+func extractHeader(req flowcontrol.FlowControlRequest, name string) string {
+	infReq := req.InferenceRequest()
+	if infReq == nil || infReq.Headers == nil {
+		return ""
+	}
+	return fwkrequest.GetHeader(infReq.Headers, name)
+}
```
shmuelk: See above comment about the same function in the other package.

pkg/epp/flowcontrol/controller/internal/item.go
loicmarchal: Good point, `nil` for req.InferenceRequest().Headers is covered by the `val, ok := headers[key]` access to the map.
req.InferenceRequest() should not be nil, I would still check for nil to be safe, but that does not require an extra function. I'll fix it and remove the function

loicmarchal:
This PR has been opened for awhile. Also it seems the metrics in the `pkk/epp/metrics` package have been deprecated, so it's probably not where we should define the new ones.
I suggest we close this PR and organize some discussion around the exact metrics requirements for the auto-scaling so we can open a new PR. Happy to reopen this later if we want to continue with this particular development.
cc @shmuelk @lionelvillard @shuynh2017

PR #946, opened 2026-04-30
Updated owners file as per the code migration alignment

ahg-g:


OWNERS
```diff
@@ -21,9 +17,12 @@ reviewers:
...
   - elevran
-  - kfswain
   - nilig
   - nirrozenbaum
   - shmuelk
   - vMaroon
   - ahg-g
+
+
+emeritus:
```
elevran, line 27: should we add 
```yaml
- nilig
- nirrozenbaum
- shmuelk
```
here as well?

ahg-g: They are still closely involved with the project and listed under reviewers, right?

elevran: True. Interpreted emeritus differently.
/lgtm

Issue #917, opened 2026-04-29
[RFC] Template for plugins README.md


revit13:
/kind documentation

Currently, the README files for the plugins located in `pkg/epp/framework/plugins` lack a consistent format. To address this, we have drafted a proposed template below to standardize our documentation. Please review and share any feedback or suggestions. Once we agree on a format, we will create follow-up PRs to apply it across the directory.

```md
# <Plugin display name>

**Type:** `<plugin-type-string>`
**Interfaces**:  `scheduling.Filter`  etc (Note: Use a middle dot · to separate multiple interfaces)
**Category**: relevant for Scorer (Affinity, Distribution, and Balance)

> [!NOTE]
>  This plugin is framework-injected by default. You do not need to explicitly declare it in your configuration. (omit if not applicable)

One-sentence summary of the plugin's purpose.

## What It Does
Explain what the plugin is meant to do and when it should be used.

## How It Works
(omit for simple plugins)
Describe the internal logic in plain language, including the main steps, rules, or scoring/selection strategy.

## Inputs Consumed
- What objects, fields, or context it receives.
- Which inputs are required or optional.
- Any assumptions about input format or ordering.

## Output Produced
(relevant for DataProducer plugins for example)

## Configuration

### Parameters
| Name | Type | Required | Default | Description |
|------|------|----------|---------|-------------|
| `...` | `...` | `yes/no` | `...` | `...` |

### Examples

## Limitations
List important tradeoffs, edge cases, or cases where the plugin should not be used.

## Related Documentation
list other relevant plugins/links
```

Guidelines:

- Avoid linking to code unless mentioned in `Related Documentation` section.
- README.md files that document multiple plugins (such as ByLabel Filter Plugins) can share a single 'Related Documentation' and 'Interfaces' section.

cc @mayabar @elevran @ahg-g

elevran:
Overall looks good to me.

I'm still confused on the intended audience for the README. Users are not (IMO) expected to read repo code, especialy 5 levels deep. Users should be getting inforamtion and guidance from the documentation site. It would be best if these are consumable as input to doc scripts  that generate documentation. A potential (overengineered...) solution would be to create some kind of template that can be used to generate the code and documentation from the same source (with makefile dependencies).

**nits**
> **Interfaces**:  `scheduling.Filter`  etc (Note: Use a middle dot · to separate multiple interfaces)

any reason to use the (uncommon) middle dot instead of a comma which is easier to find and type?

[...]

PR #947, opened 2026-04-30
cleanup: remove deprecated prefill-header-handler alias

### LINKED ISSUE #936: cleanup: remove deprecated disagg-headers-handler and prefill-header-handler

Remove the deprecated `disagg-headers-handler` and `prefill-header-handler` plugins that were superseded by `disagg-profile-handler` in #905.

**Why is this needed**:

After #905, `disagg-profile-handler` handles PreRequest natively. The deprecated files were kept for backward compatibility during the transition period. This issue tracks their removal to eliminate code duplication.
**Scope:**
- Delete `pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_headers_handler.go`
- Delete `pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_headers_handler_test.go`
- Remove `DisaggHeadersHandlerType` and `PrefillHeaderHandlerType` registration from `register.go`
- Remove the `DisaggHeadersHandler` section from `docs/architecture.md`
- Remove `deprecatedPdConfig` from `test/e2e/configs_test.go` (or migrate it to use `disagg-profile-handler`)

/kind cleanup

yukthapriya:
**What type of PR is this?**
/kind cleanup

Removes the deprecated `prefill-header-handler` plugin type, which was a backward-compatibility alias for `disagg-headers-handler`. All remaining usages have been updated to use the canonical type.

Fixes #936

```release-note
NONE
```


revit13:
We have an ongoing effort to standardize the plugin README formats (see [issue #917](https://github.com/llm-d/llm-d-inference-scheduler/issues/917)). Could you please update the plugin README with your changes while trying to keep the suggested format? Thanks

yukthapriya:
Added README.md for the disagg profilehandler plugins following the existing format. Also rebased on main to resolve conflicts. Thanks for the pointer to #917.

ahg-g:
We are looking into deprecation policy, we should deprecate this after two releases.

yukthapriya:
Thanks @ahg-g! I can revert the removal and instead mark prefill-header-handler as deprecated, keeping the alias registered but adding a deprecation comment pointing users to disagg-headers-handler. We could then plan the actual removal for whichever release fits the two-release window. Does that work?

ahg-g:
> Thanks @ahg-g! I can revert the removal and instead mark prefill-header-handler as deprecated, keeping the alias registered but adding a deprecation comment pointing users to disagg-headers-handler. We could then plan the actual removal for whichever release fits the two-release window. Does that work?

@elevran @LukeAVanDrie are putting a deprecation policy for the project; pls advise.

ahg-g:
we should do this in 0.11.0 as per the two release deprecation policy

Issue #857, opened 2026-04-21
[migration] GAIE changes and migration


elevran:
Following merge of #870, need to attend to remaining migration tasks (can be done as separate issues)

Migrate from GAIE into this repo, preserving history:
- [ ] cmd/epp, if needed: determine the llm-d cmd/epp has all functionality and code needed by GAIE cmd/epp (@hexfusion)
- [ ] integration test (@hexfusion)
- [ ] the e2e test. (@roytman)
- [ ] Makefile targets for moved functionality and code (@shmuelk #887)
- [ ] Deployment YAMLs and helm charts (@danehans)
- [x] Imports to GAIE sidecar code (after sidecard migrate to their own repo) (@kaushikmitr #881)
- [ ] CRDs other than InferencePool types (@zhouyou9505 #923)
- [ ] Any remaining GAIE imports outside of `/api` (i.e. only InferencePool remains on GAIE)
- [ ] gRPC Go Package in [vllm_engine.proto](pkg/epp/framework/plugins/requesthandling/parsers/vllmgrpc/api/proto/vllm_engine.proto) changed to [sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/plugins/requesthandling/parsers/vllmgrpc/api/gen](http://sigs.k8s.io/gateway-api-inference-extension/pkg/epp/framework/plugins/requesthandling/parsers/vllmgrpc/api/gen) (@shmuelk #887)
- [x] runtime strings embedded in exe (@evacchi #873 #922)
- [ ] consistent use of import aliases (e.g., `fwrh` for `framework/interfaces/requesthandler` or not) (@evacchi #920)
- [ ] Cleanup golintci-configuration and `//nolint` bypasses done as part of migration along with affected code (@shmuelk #934)
- [ ] Any cross reference links and examples pointing back to GAIE's (e.g., references to `doc/proposals` and `site-src`) - decide which move and which should remain referencing GAIE.
- [ ] release script if still relevant, and any useful scripts under https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/hack in general

> [!NOTE]
> 1. When picking up an iterm, please analyze which files should move together (e.g., e2e tests may rely on YAMLs). Once it is understood, please create an issue, optionally combining several tasks. If unsure, or this grows larger than expected, please discuss here or on the created issue.
[...]

PR #954, opened 2026-05-01
[Migration] Move apix v1alpha2 resources from GAIE

### LINKED ISSUE #923: [migration] Move InferenceObjective and InferenceModelRewrite CRDs from GAIE
## Background

As discussed in #857 , this migration task covers the remaining CRDs other than InferencePool types.

Per maintainer clarification, the scope for this issue is:

* InferenceObjective
* InferenceModelRewrite

InferencePool will remain in IGW.

## Proposed Work

Migrate the relevant resources for these CRDs from GAIE into this repository while preserving functionality.

Planned items:

* Move API type definitions
* Move generated deepcopy / related files if required
* Move CRD manifests / YAMLs
* Update imports and references
* Ensure no remaining GAIE references outside allowed api/docs paths
* Validate build/tests after migration

## Notes

If additional dependent files are discovered during migration, they will be included in scope or split into follow-up issues as needed.

zhouyou9505:
## Summary

This is a migration PR for issue #923, split from the broader GAIE migration tracker in #857.

This PR moves the `InferenceObjective` and `InferenceModelRewrite` API families from GAIE into `llm-d-inference-scheduler`, while keeping `InferencePool` in IGW/GIE.

Changes included in this PR:

- move `apix/v1alpha2` API type definitions into this repo
- move generated API support files (`deepcopy` / `register`)
- add local CRD manifests for:
  - `InferenceObjective`
  - `InferenceModelRewrite`
- update internal imports and references to use local `apix/v1alpha2`
- keep `InferencePool` references on the existing GIE API path
- avoid introducing new GAIE references outside the allowed `api/` / docs surface

## Scope

Included:

- `InferenceObjective`
- `InferenceModelRewrite`

Not included:

- `InferencePool` migration
- broader GAIE dependency cleanup outside this API slice
- deployment/chart migration beyond the CRD manifests for these two resources

## Validation

Validated with:

- focused Go test runs for affected packages:
  - `./apix/...`
  - `./pkg/epp/controller`
  - `./pkg/epp/datastore`
  - `./pkg/epp/requestcontrol`
  - `./pkg/epp/server`
  - `./pkg/epp/util/testing`
  - `./test/utils/igw`
- `kubectl kustomize deploy/components/crds-apix`

Notes from local validation:

- full `go test ./...` on this repo still has pre-existing failures outside this migration scope
- `make test-unit` also depends on the local Docker/builder environment

## Migration Notes

This is a **migration PR**.

Files were moved from GAIE into this repo as part of the `InferenceObjective` / `InferenceModelRewrite` migration work. Please use the special maintainer handling for migration PRs so history/signoff expectations are handled correctly, as noted in #857.

## Related Issues

- Closes #923
- Related to #857


danehans:
@zhouyou9505 please DCO-sign your commit.

apix/v1alpha2/doc.go
```diff
@@ -0,0 +1,24 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package v1alpha2 contains API Schema definitions for the
+// inference.networking.x-k8s.io API group.
+//
+// +k8s:openapi-gen=true
+// +kubebuilder:object:generate=true
+// +groupName=inference.networking.x-k8s.io
```
shmuelk, line 22: Don't we want to change the group name at this time?

We would need to support the old group name as well, deprecating it.

shmuelk (changes requested):

shmuelk:
Why was all of the code under apix simply copied and not migrated using the migration utility, which would maintain accredation?

Also you need to migrate pkg/generator...

apix/v1alpha2/doc.go
ahg-g: I think step one is to migrate the code and then in a followup PR we rename and add support both groups in epp.

elevran: as long as all are done in one release can be separate PRs. The goal is to avoid double deprecation (i.e., do move and rename at the same release).

ahg-g: There is no double deprecation if we don't do it in this release because moving the CRDs as is don't impact users as long as we didn't diverge from the source. But we should certainly try to rename the group in the next release.

Other than renaming the group to inference.llm-d.ai, I recommend graduating the inferenceObjective API to v1, this will be used frequently now and we should try to avoid having users migrate twice.

zhouyou9505:
> Why was all of the code under apix simply copied and not migrated using the migration utility, which would maintain accredation?
> 
> Also you need to migrate pkg/generator...



> Why was all of the code under apix simply copied and not migrated using the migration utility, which would maintain accredation?
> 
> Also you need to migrate pkg/generator...

Thanks, that makes sense.

I will update this PR to follow the migration utility flow so history/attribution are preserved, instead of keeping the current direct-copy approach.

I will also review the related `pkg/generator` pieces and include the required migration work here if they are part of this API move.

zhouyou9505:
Thanks for the feedback.

I reworked this change using the migration utility so history/attribution are preserved, and included the related `pkg/generator` migration as requested.

Please continue review on #989. I will close this PR in favor of #989 .

Issue #956, opened 2026-05-01
[RFC][CI]: CI/CD Improvements for `llm-d-inference-scheduler`


hickeyma:
## Summary

This RFC proposes a phased set of improvements to the `llm-d-inference-scheduler` CI/CD pipeline. The current pipeline has a solid foundation (hermetic builds via a pinned builder container, multi-arch image publishing, Trivy security scanning, full E2E tests on every PR, coverage tracking with baseline comparison, and a prow-style review workflow) but there are gaps in supply-chain security, test coverage enforcement, pipeline reliability, and operational hygiene.

## Motivation

An improved, secure, and reliable CI/CD pipeline.

## Current State

| Workflow | Trigger | Covers |
|---|---|---|
| `ci-pr-checks.yaml` | push / PR | Lint, build, unit tests, E2E, coverage comparison |
| `ci-build-images.yaml` | `workflow_call` | Build + push multi-arch container images (reusable) |
| `ci-dev.yaml` | push to main/release-* | Publishes dev-tagged images on merge |
| `ci-release.yaml` | `v*` tag / release published | Publishes release-tagged images |
| `check-typos.yaml` | all PRs + push | Spell-check via `crate-ci/typos` |
| `ci-signed-commits.yaml` | `pull_request_target` | GPG/DCO commit signature enforcement |
| `md-link-check.yml` | push/PR to main | Broken markdown link detection |
| `auto-assign.yaml` | PR opened | Random reviewer assignment from OWNERS |
| `pr-kind-label.yaml` | PR opened/edited | Parses `/kind` directive → GitHub label |
| `non-main-gatekeeper.yml` | PR synchronized | Adds hold label when base ≠ main |
| `prow-github.yml` | `issue_comment` | Slash commands (`/lgtm`, `/approve`, etc.) |
| `prow-pr-automerge.yml` | cron every 5 min | Merges PRs with `lgtm` label, no `hold` |
| `prow-pr-remove-lgtm.yml` | PR events | Re-evaluates LGTM label on PR update |
| `re-run-action.yml` | `issue_comment` | Re-triggers CI from a PR comment |
| `release-notes-update.yaml` | PR merged to main | Extracts `release-note` block → opens PR |
| `stale.yaml` | daily cron | Labels and closes stale issues/PRs |
| `unstale.yaml` | `issue_comment` / reopen | Removes stale labels on activity |

**Custom composite actions:**
[...]

PR #960, opened 2026-05-02
ci: implement Phase 1 security and reliability hardening

Jwrede:
**What type of PR is this?**

/kind cleanup

Implements the Phase 1 "Quick-Win" checklist from the CI/CD Improvements RFC (#956). These are all small, no-new-infrastructure changes that close the highest-risk gaps in the pipeline:

1. **Supply chain**: Pin `estroz/rerun-actions@main` to a verified commit SHA (`6660c16`). The `@main` ref is mutable -- any push to that repo would take immediate effect in this pipeline.

2. **Security**: Add explicit `permissions:` blocks to all workflows that were missing them (`check-typos`, `ci-build-images`, `ci-dev`, `ci-release`, `md-link-check`, `non-main-gatekeeper`, `prow-pr-automerge`, `prow-pr-remove-lgtm`, `re-run-action`). Each workflow now declares minimum required scopes.

3. **Reliability**: Add `timeout-minutes: 30` to the `lint-and-test` job in `ci-pr-checks.yaml`. Without this, a hung kind cluster holds a runner for up to 6 hours.

4. **Reliability**: Add `concurrency:` groups to all PR-triggered workflows. Rapid force-pushes no longer queue redundant parallel runs.

5. **Cost**: Reduce `prow-pr-automerge.yml` cron from `*/5` (~4,300 runs/month) to event-driven (`pull_request: [labeled]` + `check_suite: [completed]`) with a `*/30` fallback cron.

6. **Visibility**: Switch Trivy scan output from `table` to `sarif` and upload results to the GitHub Security tab via `github/codeql-action/upload-sarif@v3`. Findings are now tracked over time. `exit-code: 1` is preserved for build blocking.

7. **Vulnerability**: Add `govulncheck ./...` step after lint in `ci-pr-checks.yaml`. This queries the Go vulnerability database (vuln.go.dev) for known CVEs in the dependency tree -- distinct from Trivy's container-layer scan.

Ref: https://github.com/llm-d/llm-d-inference-scheduler/issues/956

```release-note
NONE
```


.github/workflows/prow-pr-automerge.yml
```diff
@@ -1,10 +1,18 @@
...
-# If the hold label is present, it will block automatic merging.
+# Merges PRs with the lgtm label. Blocks if hold label is present.
+# Event-driven triggers handle immediate merges; cron fallback catches edge cases.
 
 name: "Prow merge on lgtm label"
 on:
+  pull_request:
+    types: [labeled]
+  check_suite:
+    types: [completed]
```
roytman: This event fires for every completed check suite across the entire repo -- not just on PRs that have the `lgtm` label. This action will spin up, scan for eligible PRs, find nothing to merge, and exit. It's harmless but wasteful.

I suggest dropping `check_suite", and maybe set a tighter */10 cron which is quieter and more predictable than `check_suite`

.github/workflows/ci-pr-checks.yaml
```diff
@@ -36,12 +40,13 @@ jobs:
     needs: check-changes
     if: ${{ needs.check-changes.outputs.src == 'true' }}
     runs-on: ubuntu-latest
+    timeout-minutes: 30
```
roytman: Sometimes the `lint-and-test` job takes 40-50 minutes. And this is without e2e tests from the upstream repository. 
We definitely need to check how to reduce the time, but in the meantime, the 30 minutes are not enough.

.github/workflows/prow-pr-automerge.yml
Jwrede: Good point, dropped check_suite. The cron is now */10 so eligible PRs get picked up quickly without the noise of firing on every suite completion.

.github/workflows/ci-pr-checks.yaml
Jwrede: Makes sense. Bumped to 90 minutes to give headroom for e2e tests landing on top of the current 40-50 min runs.

.github/actions/trivy-scan/action.yml
```diff
@@ -12,6 +12,13 @@ runs:
...
         image-ref: ${{ inputs.image }}
-        format: 'table'
+        format: 'sarif'
+        output: 'trivy-results.sarif'
         severity: 'HIGH,CRITICAL'
         exit-code: '1'
+
+    - name: Upload Trivy SARIF to Security tab
+      uses: github/codeql-action/upload-sarif@v3
+      if: always()
```
elevran, line 22: is the `if` needed when it always runs?

.github/workflows/check-typos.yaml
```diff
@@ -4,6 +4,13 @@ on:
   pull_request:
   push:
 
+permissions:
+  contents: read
+
+concurrency:
```
elevran, line 10: Q: what does `concurrency` mean? 
If controlling the run of `typos` - is it unnecessary (really light and short run)?

.github/workflows/ci-pr-checks.yaml
```diff
@@ -91,6 +96,11 @@ jobs:
       - name: Run make lint
         run: make lint
 
+      - name: Run govulncheck
+        run: |
+          go install golang.org/x/vuln/cmd/govulncheck@latest
+          govulncheck ./...
```
elevran, line 102: Q: how/where are results saved and should they also be uploaded same as Trivy scan?
what is the overlap with `dependabot` running and auto PR'ing pacth releases (i.e., bug and secirity related)?

Jwrede:
Bumped the Go toolchain in 618f2ad to fix the new govulncheck failure. The 4 vulnerabilities (GO-2026-4870 in crypto/tls, GO-2026-4865 in html/template, GO-2026-4946 and GO-2026-4947 in crypto/x509) are all in the Go 1.25.8 stdlib and are fixed in 1.25.9. They were silent before this PR; the new govulncheck step is what surfaces them. Bumping the toolchain is the actual fix.

.github/workflows/ci-pr-checks.yaml
```diff
@@ -91,6 +96,11 @@ jobs:
       - name: Run make lint
         run: make lint
 
+      - name: Run govulncheck
+        run: |
+          go install golang.org/x/vuln/cmd/govulncheck@latest
```
elevran: does it make sense to add this to list of Makefile installed tools and running the check as part of (e.g.) `make presubmit`?
is it a good idea to pull in the latest vs some up-to-date version?

.github/workflows/check-typos.yaml
elevran: same concurrency question applies to other files/locations.

.github/actions/trivy-scan/action.yml
Jwrede: The if: always() is needed so the SARIF upload still runs when Trivy exits with code 1 (vulnerabilities found). Without it, GitHub Actions skips subsequent steps after a failure, and the results would never reach the Security tab.

.github/workflows/check-typos.yaml
Jwrede: Fair point. The concurrency group cancels older in-flight runs for the same PR when a new push arrives. For a lightweight job like typos it is not strictly necessary. Happy to drop it here if you prefer keeping it simple.

.github/workflows/ci-pr-checks.yaml
Jwrede: Right now govulncheck only fails the build on findings; results are in the job log but not uploaded as an artifact. We could add SARIF output similar to Trivy if that is useful. On overlap with Dependabot: Dependabot covers dependency updates but does not scan for reachable call paths in the stdlib or transitive modules the way govulncheck does, so they complement each other.

.github/workflows/ci-pr-checks.yaml
Jwrede: Good call on both. Pinning to a specific version (e.g. govulncheck@v1.3.0) is safer to avoid surprise breakage. And adding a make vulncheck target would let developers run it locally too. I will pin the version and can add the Makefile target if you want.

.github/workflows/check-typos.yaml
Jwrede: Acknowledged. If we drop concurrency from typos, I will remove it from the other lightweight workflows too and only keep it on ci-pr-checks where cancelling stale runs actually saves compute.

elevran: leave it to your discretion. I can see the benefit of consistency - if any new action is added in the future, the concurrency pattern would be apparent. 
It would probably not kick in for short runs anyway (you'd need to push more commit before it manages to run...), so it's probably best to leave all with the concurrency constraint.

.github/workflows/ci-pr-checks.yaml
elevran: Please do add the makefile target and make `presubmit` depend on it.

.github/workflows/check-typos.yaml
elevran: I see you already removed it... your call if to add to all or not.

Jwrede: Makes sense, restored the concurrency groups on all workflows for consistency.

.github/workflows/ci-pr-checks.yaml
Jwrede: Done. Added a make vulncheck target (govulncheck@v1.3.0, runs inside the builder container) and wired it into presubmit.

elevran:
/approve
/lgtm
please fix the conflict in workflows/check-typos.yaml

elevran (commented):
/lgtm
/approve

PR #972, opened 2026-05-03
migration: Change the Group Name in the configuration pseudo CRD to llm-d.ai

### LINKED ISSUE #935: migration: Update the Group name of the pseudo CRD used by the text based configuration
The text based configuration uses K8S machinery to load the YAML text. This requires the use of of pseudo CRD.

Currently that pseudo CRD has a group name of  `inference.networking.x-k8s.io` as it had in the Inference Gateway. This group name is being changed to `llm-d.ai`.

This issue will track:
   1. Code updates
   2. Documentation updates in this repo
   3. Various deployment/test YAML files
   4. Documentation and YAML updates in the llm-d/ll-d repository.

shmuelk:
**What type of PR is this?**
/kind cleanup

/kind deprecation

With the merge of the Inference Gateway code into the llm-d-inference-scheduler repo, the various CRDs and pseudo CRDs should reflect their new home.

This PR deprecates the old EndpoointPickerConfig group name. However, it maintains support for the old Group name while adding support for the new one. All of the configurations and documentation in the llm-d-inference-scheduler repo has been updated to use the new Group name. A test was added that uses the old group name to insure backwards compatibility for now.

Fixes #935

```release-note
The use of inference.networking.x-k8s.io/v1alpha1/EndpointPickerConfig is deprecated. Please use llm-d.ai/v1alpha1/EndpointPickerConfig instead
```


pkg/epp/config/loader/configloader.go
```diff
@@ -42,12 +44,25 @@ import (
...
+	var builder runtime.SchemeBuilder
+	(&builder).Register(func(scheme *runtime.Scheme) error {
+		scheme.AddKnownTypes(deprecatedSchemeGroupVersion,
+			&configapi.EndpointPickerConfig{},
+		)
+		// AddToGroupVersion allows the serialization of client types like ListOptions.
+		v1.AddToGroupVersion(scheme, deprecatedSchemeGroupVersion)
+		return nil
+	})
+	_ = (&builder).AddToScheme(scheme)
```
roytman: I don't know if it was done intentionally, but in this registration, the deprecated GVK takes precedence over the canonical.
The k8s scheme stores GVKs per Go type as a slice (typeToGVK), appending in registration order. 
Several encoding code paths call `ObjectKinds()` and take `gvks[0]`. 
With the current order, `gvks[0]` for `EndpointPickerConfig` is the deprecated group — so serialization paths that don't specify a target GroupVersion will stamp objects with `inference.networking.x-k8s.io/v1alpha1` instead of `llm-d.ai/v1alpha1`.

If you want to change it, please replace the registration order. 

Another comment, we ignore the possible errors in `_ = (&builder).AddToScheme(scheme)` 


```suggestion
	utilruntime.Must(configapi.Install(scheme))
    utilruntime.Must((&builder).AddToScheme(scheme))
```

pkg/epp/config/loader/configloader.go
```diff
@@ -68,6 +83,11 @@ func LoadRawConfig(configBytes []byte, logger logr.Logger) (*configapi.EndpointP
 		if err != nil {
 			return nil, nil, err
 		}
+
+		if rawConfig.GroupVersionKind().Group == deprecatedSchemeGroupVersion.Group {
+			logger.Info("The use of inference.networking.x-k8s.io/v1alpha1/EndpointPickerConfig is deprecated. Please use llm-d.ai/v1alpha1/EndpointPickerConfig instead")
```
roytman: nit: Deprecation warnings logged at `Info` level are easily lost in production.

```suggestion
			logger.Info("DEPRECATION: apiVersion inference.networking.x-k8s.io/v1alpha1/EndpointPickerConfig is deprecated",
    "replacement", "llm-d.ai/v1alpha1/EndpointPickerConfig")
```

pkg/epp/config/loader/configloader.go
```diff
@@ -68,6 +83,11 @@ func LoadRawConfig(configBytes []byte, logger logr.Logger) (*configapi.EndpointP
 		if err != nil {
 			return nil, nil, err
 		}
+
+		if rawConfig.GroupVersionKind().Group == deprecatedSchemeGroupVersion.Group {
```
roytman: The comparison checks only .Group — comparing the full GroupVersion string is more precise:

```suggestion
		if rawConfig.GroupVersionKind().GroupVersion() == deprecatedSchemeGroupVersion {
```

roytman (commented):
`configloader_test.go` at line 258 checks:
```go
diff := cmp.Diff(tc.want, got, cmpopts.IgnoreFields(configapi.EndpointPickerConfig{}, "TypeMeta"))
```
`cmpopts.IgnoreFields` tells `cmp.Diff` to skip the `TypeMeta` field entirely when computing the diff. So no matter what value `got.TypeMeta.APIVersion` has - `llm-d.ai/v1alpha1`, `inference.networking.x-k8s.io/v1alpha1`, `garbage`, or even an empty string, the test passes.

Because `IgnoreFields drops TypeMeta`, this test case only verifies: "LoadRawConfig doesn't return an error." 
It does not verify:

- That the parsed object actually retains `inference.networking.x-k8s.io/v1alpha1` in TypeMeta
- That the `deprecation warning logger.Info(...)` was actually called

So if someone accidentally deleted the entire backward-compat registration block in `init()`, or the deprecation `logger.Info call`, this test would still pass, because it only checks that `err == nil`.

In order to test the depreciation, we probably should remove the `IgnoreFields`.

pkg/epp/config/loader/configloader.go
shmuelk: Fixed

pkg/epp/config/loader/configloader.go
shmuelk: Fixed

pkg/epp/config/loader/configloader.go
shmuelk: Fixed

PR #974, opened 2026-05-03
feat(picker): add limited sampling (topK) to `weighted-random-picker`

vMaroon:
This PR adds optional candidate-pool restriction before A-Res sampling. Default behavior (no top-K knobs) is unchanged.

  - topK: absolute cap on candidates by score
  - topKPercent: percentage cap, ceil(N * P / 100)
  - minTopK: floor (default 2) so small fleets don't collapse to argmax

Useful on large fleets where the long low-score tail otherwise pulls non-trivial traffic. Bounds the worst case while preserving sampling variance within the top tier.

If we had perfect scorers expression and composability, this would not be needed.

< Note: Alternative is a new picker.

```release-note
New configuration knobs added.
```

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/picker/weightedrandom/README.md
+++ b/pkg/epp/framework/plugins/scheduling/picker/weightedrandom/README.md
@@ -27,6 +27,12 @@ The plugin config supports:
 
 - `maxNumOfEndpoints` (default 1)
   - The maximum number of endpoints to pick and return. Must be > 0. If more candidates are available than this limit, only the top subset is returned.
+- `topK` (optional, integer; default `0` = disabled)
+  - Absolute cap on the candidate-pool size before A-Res sampling. Candidates are sorted by score, descending. Combinable with `topKPercent` — when both are set, the more restrictive bound wins.
+- `topKPercent` (optional, 1-100; default `0` = disabled)
+  - Percentage cap on the candidate-pool size before A-Res sampling, computed as `ceil(N * P / 100)`. Useful when fleet size varies (autoscaling).
+- `minTopK` (default 2)
+  - Floor on the post-filter pool size when `topK`/`topKPercent` is set. Prevents collapse to a single candidate (which would degenerate into argmax). Set to `1` to allow argmax. Ignored when neither `topK` nor `topKPercent` is set.
 
 > [!TIP]
 > In most production scenarios, `maxNumOfEndpoints` is left at its default value of `1` to select a single target endpoint for the request.
--- a/pkg/epp/framework/plugins/scheduling/picker/weightedrandom/picker.go
+++ b/pkg/epp/framework/plugins/scheduling/picker/weightedrandom/picker.go
@@ -40,7 +40,79 @@ import (
 const (
 	// WeightedRandomPickerType is the registered name of the weighted random picker plugin.
 	WeightedRandomPickerType = "weighted-random-picker"
+
+	// DefaultMinTopK is the lower bound applied after topK / topKPercent. Without a floor,
+	// small fleets or aggressive percentages can collapse to a single candidate (degenerating
+	// into argmax). A default of 2 keeps the picker actually sampling. Set MinTopK = 1 to
+	// allow degeneration to argmax.
+	DefaultMinTopK = 2
 )
+
+// Parameters extends picker.PickerParameters with optional top-K filtering applied
+// before A-Res sampling. When neither TopK nor TopKPercent is set, the picker
+// considers every candidate (default behavior).
+type Parameters struct {
+	picker.PickerParameters
+
+	// TopK is the absolute cap on the candidate pool size after sorting by score.
+	// 0 (the zero value) means disabled. Combinable with TopKPercent — when both
+	// are set, the more restrictive bound wins.
+	TopK int `json:"topK,omitempty"`
+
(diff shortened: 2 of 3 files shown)


pkg/epp/framework/plugins/scheduling/picker/weightedrandom/picker.go
```diff
@@ -40,8 +40,80 @@ import (
...
+	TopK int `json:"topK,omitempty"`
+
+	// TopKPercent is the percentage cap, computed as ceil(N * P / 100). Valid
+	// values are 1-100; 0 (the zero value) means disabled. Combinable with TopK.
+	TopKPercent int `json:"topKPercent,omitempty"`
+
+	// MinTopK is the floor on the post-filter pool size. Default DefaultMinTopK.
+	// Set to 1 to allow degeneration to argmax. Ignored when neither TopK nor
+	// TopKPercent is set.
+	MinTopK int `json:"minTopK,omitempty"`
```
nirrozenbaum, line 69: do you think this could be useful for random picker as well? 
e.g., choose randomly out of the top k? (without taking into consideration their weight).

pkg/epp/framework/plugins/scheduling/picker/weightedrandom/README.md
```diff
@@ -27,6 +27,12 @@ The plugin config supports:
 
 - `maxNumOfEndpoints` (default 1)
   - The maximum number of endpoints to pick and return. Must be > 0. If more candidates are available than this limit, only the top subset is returned.
+- `topK` (optional, integer; default `0` = disabled)
```
ahg-g, line 30: I would have expected that only one of topK or topKPercent can be set, if both can be, pls explain the semantics.

pkg/epp/framework/plugins/scheduling/picker/weightedrandom/README.md
```diff
@@ -27,6 +27,12 @@ The plugin config supports:
...
+  - Absolute cap on the candidate-pool size before A-Res sampling. Candidates are sorted by score, descending. Combinable with `topKPercent` — when both are set, the more restrictive bound wins.
+- `topKPercent` (optional, 1-100; default `0` = disabled)
+  - Percentage cap on the candidate-pool size before A-Res sampling, computed as `ceil(N * P / 100)`. Useful when fleet size varies (autoscaling).
+- `minTopK` (default 2)
+  - Floor on the post-filter pool size when `topK`/`topKPercent` is set. Prevents collapse to a single candidate (which would degenerate into argmax). Set to `1` to allow argmax. Ignored when neither `topK` nor `topKPercent` is set.
```
ahg-g, line 35: I didn't read the code yet, trying to see if a user would get how this works without knowledge of the code: this parameter should only be applicable when `topKPercent` is set, no?

pkg/epp/framework/plugins/scheduling/picker/weightedrandom/README.md
revit13, line 30: We have an ongoing effort to standardize the plugin README formats. Could you please update the README with you changes and make sure its aligned with the template outlined in [issue #917](https://github.com/llm-d/llm-d-inference-scheduler/issues/917)? Thanks

nirrozenbaum:
@vMaroon do you think it makes sense to limit by score distance from the max?
e.g., think about the following use case-

we have 10 endpoints with scores (sorted by score):
| endpoint | score |
|----------|----------|
| A  | 100 |
| B  | 99  |
| C  | 98  |
| D  | 30  |
| E  | 25  |
| F  | 20  |
....

now, if the topK is set to 6, picker will choose from A-F. but if we add a `max-distance-from-max` (e.g., 20%) that would eliminate D-F, and make sure we choose endpoint from A-C, which according to the scores is either optimal or very close to optimal.

ahg-g:
> @vMaroon do you think it makes sense to limit by score distance from the max? e.g., think about the following use case-
> 
> we have 10 endpoints with scores (sorted by score):
> 
> endpoint	score
> A	100
> B	99
> C	98
> D	30
> E	25
> F	20
> ....	
> now, if the topK is set to 6, picker will choose from A-F. but if we add a `max-distance-from-max` (e.g., 20%) that would eliminate D-F, and make sure we choose endpoint from A-C, which according to the scores is either optimal or very close to optimal.

This will practically revert back to max, the reservoir sampling already gives lower probabilities to the low score candidates, and selecting them every now and then is actually important and desired to avoid hotspotting.

nirrozenbaum:
> > @vMaroon do you think it makes sense to limit by score distance from the max? e.g., think about the following use case-
> > we have 10 endpoints with scores (sorted by score):
> > endpoint	score
> > A	100
> > B	99
> > C	98
> > D	30
> > E	25
> > F	20
> > ....
> > now, if the topK is set to 6, picker will choose from A-F. but if we add a `max-distance-from-max` (e.g., 20%) that would eliminate D-F, and make sure we choose endpoint from A-C, which according to the scores is either optimal or very close to optimal.
> 
> This will practically revert back to max, the reservoir sampling already gives lower probabilities to the low score candidates, and selecting them every now and then is actually important and desired to avoid hotspotting.

fair enough.

ahg-g:
@vMaroon should we move forward with this?

vMaroon:
The new latency-based approach is better, more generalizable. This is no longer needed.

PR #870, opened 2026-04-23
migrate: sigs.k8s.io/gateway-api-inference-extension @ a048a796

elevran:
Migrates the following paths from `sigs.k8s.io/gateway-api-inference-extension` (@ a048a796) with full git history:

- `apix/config` → `apix/config`
- `cmd/epp/runner` → `cmd/epp/runner`
- `pkg/epp` → `pkg/epp`
- `pkg/common` → `pkg/common`
- `internal` → `internal`
- `test/utils` → `test/utils/igw`
- `version` → `version`

To pick up future upstream changes: `migrate-gaie-paths.sh --since a048a796 apix/config apix/config cmd/epp/runner cmd/epp/runner pkg/epp pkg/epp pkg/common pkg/common internal internal test/utils test/utils/igw version version`

**Merge via "Create a merge commit"** — squash or rebase discards the migrated history.

PR #989, opened 2026-05-05
migrate: sigs.k8s.io/gateway-api-inference-extension @ ed1af5d8

zhouyou9505:
Migrates the following paths from `sigs.k8s.io/gateway-api-inference-extension` (@ ed1af5d8) with full git history:

- `apix/v1alpha2` → `apix/v1alpha2`
- `pkg/generator` → `pkg/generator`

To pick up future upstream changes: `migrate-gaie-paths.sh --since ed1af5d8 apix/v1alpha2 apix/v1alpha2 pkg/generator pkg/generator`

**Merge via "Create a merge commit"** — squash or rebase discards the migrated history.


shmuelk:
@zhouyou9505 Thank you for this PR. How is it related to your other PR #954

zhouyou9505:
Thanks. This PR supersedes #954.

#954 was the initial version of the same work for issue #923. I reworked it here using the migration utility to preserve history/attribution, and included the related `pkg/generator` migration as requested.

Please continue review on #989. I will close #954. @shmuelk

elevran:
@zhouyou9505 there seems to be something wrong with the way this PR was created. While it has all commits, they're all under your name. See difference vs #987 and #870 where each commit is associated and shows the original author.

roytman:
Hi @zhouyou9505 , I had a similar issue. If your git is configured to automatically sign commits (you can check it by `git config --global commit.gpgsign`). 

The solution is  `git -c commit.gpgsign=false rebase main --force-rebase` and after that sign your last commit by `git commit --amend --no-edit -S`

pkg/generator/main.go
```diff
@@ -0,0 +1,107 @@
...
+	"fmt"
+	"log"
+	"os"
+
+	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"sigs.k8s.io/controller-tools/pkg/crd"
+	"sigs.k8s.io/controller-tools/pkg/loader"
+	"sigs.k8s.io/controller-tools/pkg/markers"
+	"sigs.k8s.io/yaml"
+
```
roytman, line 29: We don't need this line separation

pkg/generator/main.go
```diff
@@ -0,0 +1,107 @@
...
+	"sigs.k8s.io/gateway-api-inference-extension/version"
+)
+
+// This generation code is largely copied from
+// github.com/kubernetes-sigs/gateway-api/blob/main/pkg/generator/main.go
+func main() {
```
roytman, line 35: Not for this PR, but should it be moved to something like `cmd/generator/` or `tools/generator/`?

test/e2e/e2e_suite_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	k8slog "sigs.k8s.io/controller-runtime/pkg/log"
 	infextv1 "sigs.k8s.io/gateway-api-inference-extension/api/v1"
-	infextv1a2 "sigs.k8s.io/gateway-api-inference-extension/apix/v1alpha2"
+	infextv1a2 "github.com/llm-d/llm-d-inference-scheduler/apix/v1alpha2"
```
roytman, line 22: We don't need this line separation, but we do need it between lines 22 and 23.

pkg/epp/controller/inferenceobjective_reconciler_test.go
```diff
@@ -31,7 +31,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	v1 "sigs.k8s.io/gateway-api-inference-extension/api/v1"
-	"sigs.k8s.io/gateway-api-inference-extension/apix/v1alpha2"
+	"github.com/llm-d/llm-d-inference-scheduler/apix/v1alpha2"
```
roytman: can you combine the imports from this project in a separate group

roytman (commented):
Please organize imports in all files

zhouyou9505:
Addressed in the latest update:

1. removed the unnecessary line separation in `pkg/generator/main.go`
2. adjusted the import grouping in `test/e2e/e2e_suite_test.go`
3. grouped the project-local imports together in `pkg/epp/controller/inferenceobjective_reconciler_test.go`

For the `pkg/generator` location, I agree that `cmd/` or `tools/` may be a better long-term home. I’d like to keep that out of this migration PR and open a separate follow-up PR for it. @roytman

pkg/generator/main.go
zhouyou9505: Agree , I agree that `cmd/` or `tools/` may be a better long-term home. I’d like to keep that out of this migration PR and open a separate follow-up PR for it.

ahg-g:
Thanks @zhouyou9505 ; this needs a rebase and resolving conflicts

pkg/generator/main.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"fmt"
+	"log"
+	"os"
+
+	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"sigs.k8s.io/controller-tools/pkg/crd"
+	"sigs.k8s.io/controller-tools/pkg/loader"
+	"sigs.k8s.io/controller-tools/pkg/markers"
+	"sigs.k8s.io/gateway-api-inference-extension/version"
```
ahg-g: This points to the gateway-api-inference-extension version, should we have our own?

@zetxqx you copied this file for igw to add "CRD annotation for gateway-api-inferecence-extention CRD", do we need it for InfObj and InfModelRewrite CRDs if they are not directly related to gateway api and its conformance tests?

ahg-g: Related PR: https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/1134

zhouyou9505:
ok , I rebased onto the latest `main` and resolved the conflicts. @ahg-g 

I resolved the rebase/conflicts issue.

The remaining failing check is the signed-commits workflow, which appears to require signatures on all migrated historical commits in the PR. For this migration PR, should I re-sign the full migrated history, or is there an exception/expected handling for imported historical commits?

pkg/generator/main.go
zhouyou9505: I agree we should not keep this tied to the GAIE version package.

This repo already has its own `version` package, so I’ll switch the generator to use the local llm-d version package instead. If we decide these migrated CRDs do not need that annotation at all, we can follow up by removing that logic entirely.

shmuelk:
@zhouyou9505 While you are no longer the only one on all of the commits in the PR, you still are on all of them.

Something is still wrong here.

zhouyou9505:
I checked the current PR commit list after the latest force-push. The migrated commits now appear to preserve the original authors/committers in the GitHub Commits tab, and only the final import-rewrite commit is authored/signed by me @shmuelk

pkg/generator/main.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/llm-d/llm-d-inference-scheduler/version"
+)
+
+// This generation code is largely copied from
+// github.com/kubernetes-sigs/gateway-api/blob/main/pkg/generator/main.go
+func main() {
+	roots, err := loader.LoadRoots(
+		"k8s.io/apimachinery/pkg/runtime/schema", // Needed to parse generated register functions.
+		"sigs.k8s.io/gateway-api-inference-extension/apix/v1alpha1",
```
roytman, line 38: nit: should we combine `sigs.k8s.io/gateway-api-inference-extension` imports together?

zhouyou9505: Good catch, I can group the remaining `sigs.k8s.io/gateway-api-inference-extension/...` roots together.

version/version.go
```diff
@@ -25,10 +25,10 @@ var (
 )
 
 const (
-	// BundleVersionAnnotation is the annotation key used in the Gateway API inference extension CRDs to specify
-	// the installed Gateway API inference extension version.
+	// BundleVersionAnnotation is the annotation key used in llm-d CRDs to specify
+	// the installed bundle version.
```
roytman, line 29: Should the `BundleVersionAnnotation` be `llm-d.ai/bundle-version` ?

apix/v1alpha2/inferenceobjective_types.go
```diff
@@ -0,0 +1,121 @@
...
+	// However, implementations that consume this field (such as the Endpoint Picker) will treat an unset value as '0'.
+	// Priority is used in flow control, primarily in the event of resource scarcity(requests need to be queued).
+	// All requests will be queued, and flow control will _always_ allow requests of higher priority to be served first.
+	// Fairness is only enforced and tracked between requests of the same priority.
+	//
+	// Example: requests with Priority 10 will always be served before
+	// requests with Priority of 0 (the value used if Priority is unset or no InferenceObjective is specified).
+	// Similarly requests with a Priority of -10 will always be served after requests with Priority of 0.
+	// +optional
+	Priority *int `json:"priority,omitempty"`
```
roytman, line 73: It is less related to this PR, but the type should be *int32
```suggestion
	Priority *int32`json:"priority,omitempty"`
```

apix/v1alpha2/inferencemodelrewrite_types.go
```diff
@@ -0,0 +1,199 @@
...
+	// implementation supports. Weight is not a percentage and the sum of
+	// weights does not need to equal 100.
+	//
+	// If a weight is set for any targetModel, it must be set for all targetModels.
+	// Conversely weights are optional, so long as ALL targetModels do not specify a weight.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=1000000
+	Weight int32 `json:"weight"`
```
roytman, line 117: Probably for another PR, but it should be 
```go
Weight *int32 `json:"weight,omitempty"`
```
The field should be *int32 with json:"weight,omitempty" to properly support the optional/all-or-nothing contract described. This is the same class of bug fixed in PR #919.

-  When a user submits YAML without weight, the field is absent from the request JSON. Since the field is // +optional, it is not in the CRD schema's required list. JSON Schema minimum: 1 only applies when the field is actually present, so the initial CREATE passes admission.
- int32 without omitempty always serializes -- the zero value 0 is written to etcd. So after the CREATE, the stored object has weight: 0.

The breakage happens on the next write:

Any subsequent UPDATE that does a read-modify-write (e.g., a controller reconciling the object, or kubectl apply touching another field) will serialize the stored struct back to JSON, producing weight: 0 in the request body. Now the field IS present in the request, minimum: 1 applies, and the UPDATE is rejected.

So in practice:

- Initial create without weight: works
 - Any controller-driven update on that object: rejected by validation
- The "all targets unweighted" case is effectively unusable without working around the validation

roytman:
Thank you @zhouyou9505, I left a comment about `BundleVersionAnnotation`. Two other comments can be resolved separately.

apix/v1alpha2/inferenceobjective_types.go
zhouyou9505: Good catch. I agree with both points, but they are a bit broader than this migration PR.

I’d prefer to keep this PR focused on moving the existing API surface from GAIE into this repo, and open a separate follow-up PR to address:
1. changing `Priority` to `*int32`
2. changing `Weight` to `*int32` with `omitempty` and updating the related logic/tests accordingly

version/version.go
zhouyou9505: Yes, that makes sense. I updated `BundleVersionAnnotation` to `llm-d.ai/bundle-version` so the annotation namespace matches the llm-d ownership of these migrated CRDs.

shmuelk:
Thank you very much. This looks very very good now. Thanks for all of the effort here.

/lgtm
/approve

shmuelk:
@zhouyou9505 Unfortunately CI failed due to `go mod tidy` issues.

Please run `go mod tidy` and push the updated go.mod and go.sum files.

zhouyou9505:
Thanks, I ran `go mod tidy` and pushed the updated dependency files.
@shmuelk

shmuelk:
@zhouyou9505 Please update the failing end to end test

zhouyou9505:
CI is currently blocked with: “This workflow is awaiting approval from a maintainer”.
Could a maintainer please approve and run the workflows for this PR? @shmuelk

ahg-g:
Need to update https://github.com/llm-d/llm-d-inference-scheduler/blob/main/hack/update-codegen.sh#L22

zhouyou9505:
> Need to update https://github.com/llm-d/llm-d-inference-scheduler/blob/main/hack/update-codegen.sh#L22

Do we want CRDs to be generated via `hack/update-codegen.sh` (enabling `go run ./pkg/generator`), or should we rely on `controller-gen crd ...` only and remove the generator call (and/or drop line 22 entirely)?

Enabling the generator tends to bring additional generated artifacts (e.g. `config/crd/bases/`, `client-go/`, and updated `zz_generated.*`), so I want to align on the intended source of truth before changing the codegen flow. @ahg-g

ahg-g:
As a starting point, we should mirror the exact set up we had in the gateway-api-inference-extension repo.

> Enabling the generator tends to bring additional generated artifacts (e.g. config/crd/bases/, client-go/, and updated zz_generated.*), so I want to align on the intended source of truth before changing the codegen flow. @ahg-g

But that is the intent, this repo now owns those apis and is the source of truth for them. Note that as a follow up, we need to change the api group to `llm-d.ai`.

zhouyou9505:
> As a starting point, we should mirror the exact set up we had in the gateway-api-inference-extension repo.
> 
> > Enabling the generator tends to bring additional generated artifacts (e.g. config/crd/bases/, client-go/, and updated zz_generated.*), so I want to align on the intended source of truth before changing the codegen flow. @ahg-g
> 
> But that is the intent, this repo now owns those apis and is the source of truth for them. Note that as a follow up, we need to change the api group to `llm-d.ai`.

the codegen script has been updated so the generator runs from the repo root instead of depending on
  the caller’s working directory, and the full codegen flow was rerun successfully. @ahg-g

Issue #934, opened 2026-04-30
migration: Re-enable the disabled lint rules


shmuelk:
The lint rules in llm-d-inference-scheduler were stricter than those in the Inference Gateway. In order to enable the merge of the code the rules in llm-d-inference-scheduler were relaxed.

This issue will track the re-enabling of those disabled rules and the needed source code changes.

ahg-g:
anything left on this issue?

shmuelk:
Tere is one revive rule left `exported`. I ran it and it complains about the names of our exported items. For example, it doesn't like the type LoggingOptions in pkg/common/observability/logging, because it gets used as logging.LoggingOptions. This use sort of stutters. There are others like that.

There are also these rules which have been excluded, if understand correctly:

```yaml
      # Package names inherited from GAIE upstream (common, types, utils, http) are intentional
      - linters: [revive]
        text: "avoid meaningless package names"
      - linters: [revive]
        text: "avoid package names that conflict with Go standard library package names"
```

ahg-g:
Do we still anything left to do with this issue?

ahg-g:
pls open if this is still an issue.

Issue #832, opened 2026-04-16
Evaluate Go linter configurations to consider GAIE


elevran:
As efforts and code are concentrated into llm-d, we should consider aligning the linter configurations so code is subject to the same standards.
Currently, inference scheduler has more strict checks which merged code will not pass.
Specifically, the following failures are seen using the llm-d cofiguration to lint GAIE `pkg/epp/...` and `pkg/common/...`:

| Linter / Rule | Count |
|---------------|------:|
| `errcheck` | 34 |
| `goconst` | 2 |
| `nilnil` | 24 |
| `revive/context-as-argument` | 8 |
| `revive/exported` | 380 |
| `revive/package-comments` | 68 |
| `revive/if-return` | 1 |
| `revive/indent-error-flow` | 1 |
| `revive/receiver-naming` | 16 |
| `revive/unexported-return` | 1 |
| `revive/var-declaration` | 1 |
| `revive/var-naming` | 26 |
| **Total** | **562** |

The majority of these are suppressed by the GAIE configuration of:
- exclude on `comments` (suppresses all "missing comment" violations on packages or exported symbols) - this accounts for the 448 `revive/exported` + `revive/package-comments`.
- exclude on `std-error-handling` (suppresses unchecked errors for well-known stdlib patterns, such as unchecked errors on `*.Close()`) - this accounts for most of the 34 `errcheck` 

The full report is atteched to this issue.

**Plan** (seeking feedback):
- [ ] relax llm-d inference scheduler linting configuration, specifically including the above two exclusions. This can be temporary or permanant change. #834 
- [ ] address other lint errors during code migration as part of the merge
- [ ] evaluate and decide on long term linter configuration (based on common practices and minimizing false positives). Any changes to the configurations should be merged alongside modified existing code files to linting on **all** code is clean (and not just new code from the change onwards)
- [ ] go over all code to fix any issues marked with `//nolint` comments. This requires an understanding of callers, their expectations and impact for fixing or keeping the lint suppression.

elevran:
Full lint report (562 violations across 91 directories, with per-directory breakdown split by linter/rule): https://gist.github.com/elevran/5ff0e8c0ae58a4f6cb9f6f4fac3b07fc

ahg-g:
The plan looks good to me, thanks for driving this!

elevran:
disabled relevant linters on merge. Some have been re-enabled since.

PR #990, opened 2026-05-05
Enabled more lint rules and corrected related code issues - part three

shmuelk:
What type of PR is this?
/kind cleanup

What this PR does / why we need it:
The Inference Gateway Extension (IGW) project had a less strict configuration for linting the code using golangci-lint than was used in the llm-d-inference-scheduler. A number of the lint rules were removed to enable the migration of the code to the llm-d-inference-scheduler repo.

This PR is an additional step in re-enabling the golangci-lint rules that were removed. It re-enables several rules and corrected the issues in the IGW code to enable CI to pass.

Which issue(s) this PR fixes:
refs #832 and #934

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -58,7 +58,6 @@ linters:
         - name: error-strings
         - name: error-naming
         - name: errorf
-        #- name: exported
         - name: if-return
         - name: increment-decrement
         - name: indent-error-flow
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -26,6 +26,7 @@ linters:
     - makezero
     - misspell
     - nakedret
+    - nilnil
     - perfsprint
     - prealloc
     - revive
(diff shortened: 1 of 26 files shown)


.golangci.yml
```diff
@@ -57,11 +58,13 @@ linters:
         - name: error-strings
         - name: error-naming
         - name: errorf
+        #- name: exported
```
elevran: if commented out - just remove

pkg/epp/framework/interface/datalayer/endpoint_metadata.go
```diff
@@ -42,22 +42,22 @@ func (e *EndpointMetadata) String() string {
 }
 
 // Clone returns a full copy of the object.
-func (p *EndpointMetadata) Clone() *EndpointMetadata {
-	if p == nil {
+func (e *EndpointMetadata) Clone() *EndpointMetadata {
+	if e == nil {
```
elevran: if already changing names, consider picking a more aligned shorthand (e.g., ep for Endpoint, pl for PredicatedLatency, etc.)

PR #998, opened 2026-05-06
refactor(tokenizer): return errors from PrepareRequestData instead of swallowing them

acardace:
## Summary

Changes the tokenizer plugin's error handling from fail-open (silently swallow errors inside the plugin) to fail-closed (return errors to the caller). The Director at `director.go:200-204` already handles `PrepareRequestData` errors gracefully — it logs and continues — so this change surfaces tokenizer failures in logs without breaking request flow.

### `tokenizer.go`

- `tokenize()` signature changed from `([]uint32, *tokenization.MultiModalFeatures)` to `(*fwkrh.TokenizedPrompt, error)`
- Nil body → returns `errors.New("request body is nil")` instead of silent `(nil, nil)`
- Already-tokenized request → returns existing prompt unchanged (idempotent short-circuit)
- Tokenizer failure → returns `fmt.Errorf("tokenization failed: %w", err)` instead of logging and returning `(nil, nil)`
- Unsupported body type → still returns `(nil, nil)` (passthrough, no change)
- `tokenize()` now constructs the `*fwkrh.TokenizedPrompt` struct internally
- `PrepareRequestData()` is now a thin wrapper: calls `tokenize()`, propagates errors, assigns result

### `tokenizer_test.go`

- Removed `TestPrepareRequestData_NilRequest` (nil request now caught by nil-body error)
- Updated `TestPrepareRequestData_NilBody` to expect an error
- Added `TestPrepareRequestData_TokenizerError` — verifies tokenizer failures propagate
- Added `TestPrepareRequestData_UnsupportedBodyType` — verifies passthrough returns nil without error

## Why

Previously, tokenizer errors were silently swallowed (`logger.Error(err, "..."); return nil, nil`). This hid failures from operators — downstream plugins would silently get `nil` TokenizedPrompt and fall back without any signal that tokenization failed. By returning errors, the Director logs them at the appropriate level, and operators can detect and debug tokenizer issues.

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
@@ -22,6 +22,7 @@ package tokenizer
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"sort"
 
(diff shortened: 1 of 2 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -146,34 +147,34 @@ func (p *Plugin) Consumes() map[string]any {
 
 // PrepareRequestData tokenizes the request prompt and stores the result on
 // InferenceRequestBody.TokenizedPrompt (TokenIDs + MultiModalFeatures in flat shape).
-// Fail-open: errors are logged; TokenizedPrompt is left nil. If the request
-// already carries a TokenizedPrompt, tokenization is skipped.
+// Returns an error when tokenization fails; the caller (Director) decides the
+// policy (currently: log and continue). If the request already carries a
+// TokenizedPrompt, tokenization is skipped.
 func (p *Plugin) PrepareRequestData(ctx context.Context, request *scheduling.InferenceRequest, _ []scheduling.Endpoint) error {
-	if request == nil || request.Body == nil || request.Body.TokenizedPrompt != nil {
```
hexfusion, line 152: @vMaroon I am trying to understand why we bailed early here without signal. by glancing here it seems an error makes sense and caller can handle that how they want but I could be missing something.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -146,34 +147,34 @@ func (p *Plugin) Consumes() map[string]any {
...
+	request.Body.TokenizedPrompt = tp
 	return nil
 }
 
 // tokenize extracts token IDs and optional multimodal features from the request.
-// Returns (nil, nil) on error or unsupported type.
-func (p *Plugin) tokenize(ctx context.Context, request *scheduling.InferenceRequest) ([]uint32, *tokenization.MultiModalFeatures) {
+// Returns the existing TokenizedPrompt unchanged if one is already set.
+// Returns a non-nil error if the request body is nil, has an unsupported type,
+// or if the tokenizer fails.
```
hexfusion, line 166: this contract makes sense to me but its a change in behavior for every caller. are we certain the callers are handling this signal correctly now? meaning on error does this mean execution stops or is the error signal just logged?

acardace: @hexfusion you're right, the director does log the error without stopping but one plugin returing an error causes other configured plugins which have not been executed yet to not run. I'll fix it. Thanks for catching it!

acardace: too late I guess, I'll just open a new PR.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
acardace: I think this is mainly due to the fact that is plugin is completely optional so we don't want to block anything in case tokenization (through the plugin) cannot be done.

PR #1001, opened 2026-05-06
Correct issues related to pulling images in tests

shmuelk:
**What type of PR is this?**
/kind bug

1. The current end to end test suite setup code, always attemps from code to pull a set of images. This is after a pre-requisite Makefile target pulled the images, if necessary. This causes issues if trying to use locally built images for the test.
2. The shell script used to launch the development kind environment, when using docker, trys to pull all needed images, without first checking if they are available locally.
3. Upgraded the patch level of the K8S image used by kind. Includes containerd fixes which may help in loading multi-architecture images.

```release-note
NONE
```

### DIFF
--- a/scripts/kind-dev-env.sh
+++ b/scripts/kind-dev-env.sh
@@ -304,20 +304,7 @@ fi
 
 pull_image() {
     local image="$1"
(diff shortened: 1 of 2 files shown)


roytman (commented):
/lgtm

PR #1002, opened 2026-05-06
fix(tokenizer): log errors instead of returning them to avoid blocking plugin DAG

acardace:
## Summary

Follow-up to #998. That PR changed `tokenize()` to return errors instead of swallowing them, and `PrepareRequestData` propagated those errors to the caller. However, `executePluginsAsDAG` aborts the entire plugin chain on the first error — so a tokenizer failure would prevent all other `DataProducer` plugins (e.g., `inflightload`, `approximateprefix`, `predictedlatency`) from running.

This fix changes `PrepareRequestData` to log the error and return `nil`, so tokenization failures are visible but do not block unrelated plugins.

## Changes

- `PrepareRequestData`: log the error from `tokenize()` and return `nil` instead of propagating it
- `TokenizedPrompt` is left `nil` on failure; downstream scorers (`contextlengthaware`, `preciseprefixcache`) already fall back gracefully when it is absent


pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -153,7 +153,9 @@ func (p *Plugin) Consumes() map[string]any {
 func (p *Plugin) PrepareRequestData(ctx context.Context, request *scheduling.InferenceRequest, _ []scheduling.Endpoint) error {
 	tp, err := p.tokenize(ctx, request)
 	if err != nil {
-		return err
+		log.FromContext(ctx).WithName(p.typedName.String()).Error(err, "tokenization failed")
+		// skip error otherwise other PrepareRequestData plugins will not run
+		return nil
```
hexfusion: question: there is never an error that can come from tokenize that should stop execution?

acardace: I think not while this is completely optional but I'd also ping @vMaroon about this. My understanding was that we can't yet fully enforce tokenization only through the dataproducer plugin.

hexfusion: can we look at the failing test and then in terms of regression can we look at adding something gating here if possible? I would expect a test failure from your last PR.

acardace: The tests now fail because they expect an error but of course this is now swallowed by PrepareRequestData().

Conceptually we don't need this PR because a pluging that consumes a `TokenizedPrompt' should not run if this plugin fails but I'd like @vMaroon or @elevran to weigh in on this before making a decision.

ahg-g: @vMaroon pls advise so that we either close or merge this PR.

ahg-g: Masking errors is not a good practice, I think we should not move forward with this PR

vMaroon: I think this should be a question about whether `executePluginsAsDAG` should halt if any producer fails. I think a failed producer should only stop the progression of its sub-graph, and not others.

So I don't think this is the right place to solve this.

acardace: @vMaroon @ahg-g agree with both.

ahg-g:
What is the status of this PR?

PR #1007, opened 2026-05-06
 Discovery Plugins: K8s and File-based Endpoint Discovery

ezrasilvera:
## Summary

This PR builds on the _EndpointDiscovery_ interface (#908) by delivering the concrete plugin implementations: 

1. Two Kubernetes-based plugins that migrate the existing pod-watching logic into the new framework, 
2. A new file-based plugin that allows llm-d to run entirely without Kubernetes.

### K8s discovery plugins 
Introduced two new K8s discovery plugins under (pkg/epp/framework/plugins/datalayer/discovery/k8s). The existing Kubernetes pod-watching logic has been refactored into two explicit plugins:
  - **inference-pool-discovery** : watches pods matching an InferencePool CRD, equivalent to the previous default behavior
  - **static-selector-discovery** : watches pods matching a fixed label selector, equivalent to the previous --endpoint-selector CLI path

Both plugins use the existing datalayer.BindNotificationSource with a shared new podDiscoveryExtractor to translate pod lifecycle events into DiscoveryNotifier calls, eliminating duplicated controller-runtime wiring.

### File based discovery
Introduced a new file-discovery plugin (pkg/epp/framework/plugins/datalayer/discovery/file). The plugin reads endpoints from a YAML file and optionally watches it for live changes via fsnotify. 

When this plugin is used the EPP has no dependency on Kubernetes and can run on bare-metal, Slurm, Ray, or any other non-Kubernetes infrastructure -- enabling llm-d inference scheduling in environments such as RL training clusters where a Kubernetes cluster is unavailable or undesirable.

### Runner wiring (cmd/epp/runner/runner.go):
The runner now resolves the active discovery plugin at startup and drives it through NewDiscoveryNotifier. The discovery.pluginRef field in EndpointPickerConfig selects the plugin; when absent, the runner falls back to the legacy K8s behavior described below.

### Backward compatibility
No breaking changes. Existing deployments require no config changes:

The discovery section is only needed when explicitly opting into a non-default plugin such as file-discovery.

  | Config | Behavior |
  |---|---|
  | No `discovery` section, `--pool-name` set | Uses `inference-pool-discovery` -- identical to the previous default |
  | No `discovery` section, `--endpoint-selector` set | Uses `static-selector-discovery` -- identical to the previous `--endpoint-selector` path |
  | Neither flag set, no `discovery` section | EPP fails to start with a clear error |

**Documentation**
A detailed description can be found in:  docs/discovery.md . It covers all three built-in plugins, the backward compatibility behavior, and runnable examples for both standard and P/D disaggregated setups using file-discovery.

discovery-architecture drawio


apix/config/v1alpha1/endpointpickerconfig_types.go
```diff
@@ -67,6 +67,14 @@ type EndpointPickerConfig struct {
...
 	// If unspecified, default parsing behavior will be applied.
 	Parser *ParserConfig `json:"parser,omitempty"`
+
+	// +optional
+	// Discovery specifies which EndpointDiscovery plugin to use for populating the
+	// endpoint datastore. When set, the EPP bypasses Kubernetes CRD reconcilers and
+	// relies entirely on the referenced plugin to enumerate and track inference
+	// endpoints. This enables running the EPP without a Kubernetes cluster.
+	// If omitted, the EPP uses the default Kubernetes-based discovery.
+	Discovery *DiscoveryConfig `json:"discovery,omitempty"`
```
elevran: I wonder if this does not belong as a section under `DataLayer`?

cmd/epp/runner/runner.go
```diff
@@ -41,7 +41,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
+	nhpprof "net/http/pprof"
```
elevran: please fix the import order/groupnig (std lib, 3rd party, local, each group alpha sorted and empty line between groups).

cmd/epp/runner/runner.go
```diff
@@ -186,51 +195,234 @@ func (r *Runner) Run(ctx context.Context) error {
...
 	}
 
-	pmc, err := backendmetrics.NewPodMetricsClientImpl(setupLog, backendmetrics.Config{
-		ModelServerMetricsScheme:        opts.ModelServerMetricsScheme,
-		ModelServerMetricsHTTPSInsecure: opts.ModelServerMetricsHTTPSInsecure,
-		ModelServerMetricsPath:          opts.ModelServerMetricsPath,
+	namespace := resolvePoolNamespace(opts.PoolNamespace)
+	poolName := opts.PoolName
+	if poolName == "" {
+		if v := os.Getenv("POD_NAME"); v != "" {
```
elevran: `POD_NAME`?

cmd/epp/runner/rungroup.go
elevran, line 1: perhaps under internal/runner package instead of a new package?

pkg/epp/framework/plugins/datalayer/discovery/k8s/k8s_discovery.go
```diff
@@ -0,0 +1,355 @@
...
+	if err != nil {
+		return fmt.Errorf("static-selector-discovery: invalid endpointSelector: %w", err)
+	}
+
+	pool := datalayer.NewEndpointPool(s.namespace, StaticSelectorPluginType)
+	pool.Selector = selectorMap
+	pool.TargetPorts = s.targetPorts
+	s.ds.WithEndpointPool(pool)
+
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
```
elevran, line 349: The `ctrl.NewManager` is created in this path with no `Cache.Options` to filter pods by the pool's namespace. Thus, Pods are watched cluster-wide across all namespaces. `InferencePoolDiscoveryPlugin.Start` correctly scopes via `ByObject`. Since `podDiscoveryExtractor` filters by label match only (not namespace), pods from other namespaces with matching labels will be upserted as endpoints.

pkg/epp/framework/plugins/datalayer/discovery/k8s/pod_extractor.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil
+	}
+
+	for _, meta := range podToEndpointMetadata(pod, e.pool.PoolTargetPorts()) {
+		e.notifier.Upsert(meta)
+	}
+	return nil
+}
+
+func podToEndpointMetadata(pod *corev1.Pod, targetPorts []int) []*fwkdl.EndpointMetadata {
```
elevran, line 106: `podToEndpointMetadata` creates an endpoint for every target port unconditionally. The existing `PodUpdateOrAddIfNotExist` path calls `extractActivePorts(pod, targetPorts)` which reads labels to skip ports the pod declared inactive. Any deployment using that annotation will now have traffic forwarded to inactive ports via the new discovery path. 
It would be preferred to have the new Backend paths follow the same logic.
Either consolidate into shared code path or add tests that validate the same behavior across all known variations  (e.g., labels and annotations, different number of ports (1 and >1), inactive or not-ready Pods, ...)

pkg/epp/framework/plugins/datalayer/discovery/file/file_discovery.go
```diff
@@ -0,0 +1,185 @@
...
+			if !ok {
+				return nil
+			}
+			logger.Error(err, "watcher error")
+		}
+	}
+}
+
+func (f *FileDiscovery) load(notifier fwkdl.DiscoveryNotifier) error {
+	data, err := os.ReadFile(f.path)
```
elevran: `EndpointMetadata.MetricsHost` built from this file flows to the HTTP metrics poller (`NewEndpoint` -> `HTTPDataSource` -> `client.Get`), and `Address`+`Port` is set in `x-gateway-destination-endpoint` which Envoy uses to route real user traffic. There is no validation of address and thus potentially SSRF.

Can validate `net.ParseIP(e.Address)` and confirm port is a valid int in range, before `notifier.Upsert` in `load()`'s entry loop. 
You can also add a netmask in the config to capture accidental typos or later "hijack" of the file content itself.

pkg/epp/framework/plugins/datalayer/discovery/k8s/k8s_discovery.go
```diff
@@ -0,0 +1,355 @@
...
+		mgrOpts.LeaderElectionNamespace = namespace
+		mgrOpts.LeaderElectionReleaseOnCancel = true
+	}
+
+	mgr, err := ctrl.NewManager(cfg, mgrOpts)
+	if err != nil {
+		return fmt.Errorf("inference-pool-discovery: failed to create manager: %w", err)
+	}
+
+	if err := (&controller.InferencePoolReconciler{
```
elevran, line 240: `InferencePoolReconciler` is registered directly here while Pod events go through `BindNotificationSource` + `NotificationExtractor` in `runManager`. Both watch K8s events on the same manager; using two different patterns is inconsistent and bypasses the datalayer notification machinery.

The reconciler logic maps directly onto a `NotificationExtractor` - `EventDelete`/deletion-timestamp -> `ds.Clear()`, `EventAddOrUpdate` -> convert unstructured to `v1.InferencePool` -> `ds.PoolSet(ctx, reader, endpointPool)`. A new `inferencePoolSelectorExtractor` alongside `podDiscoveryExtractor` in `pod_extractor.go`, wired via:

```go
poolSrc := dlnotifications.NewK8sNotificationSource("k8s-notification", "pool-selector-sync", inferencePoolGVK)
datalayer.BindNotificationSource(poolSrc,
    []fwkdl.NotificationExtractor{newInferencePoolSelectorExtractor(k.ds, mgr.GetClient())},
    mgr)
```

`InferenceObjectiveReconciler` and `InferenceModelRewriteReconciler` follow the same pattern. `mgr.GetClient()` is available before `mgr.Start()` so `client.Reader` injection is straightforward.

pkg/epp/framework/plugins/datalayer/discovery/k8s/k8s_discovery.go
```diff
@@ -0,0 +1,355 @@
...
+	return &InferencePoolDiscoveryPlugin{
+		typedName:     fwkplugin.TypedName{Type: InferencePoolPluginType, Name: InferencePoolPluginType},
+		PoolName:      poolName,
+		PoolNamespace: poolNamespace,
+		PoolGroup:     poolGroup,
+		LeaderElect:   leaderElect,
+	}
+}
+
+func InferencePoolFactory(name string, parameters json.RawMessage, _ fwkplugin.Handle) (fwkplugin.Plugin, error) {
```
elevran, line 155: `StaticSelectorFactory` validates `endpointSelector` and `endpointTargetPorts` at factory time and returns an error immediately on missing required params. This factory accepts an empty `poolName` without complaint and the check only fires in `Start()` after the EPP has already started. Add check for empty pool and return and error here.

pkg/epp/framework/plugins/datalayer/discovery/k8s/pod_extractor.go
```diff
@@ -0,0 +1,127 @@
...
+
+func (e *podDiscoveryExtractor) TypedName() fwkplugin.TypedName { return e.typedName }
+
+func (e *podDiscoveryExtractor) GVK() schema.GroupVersionKind { return podDiscoveryGVK }
+
+func (e *podDiscoveryExtractor) ExpectedInputType() reflect.Type {
+	return reflect.TypeFor[fwkdl.NotificationEvent]()
+}
+
+func (e *podDiscoveryExtractor) ExtractNotification(ctx context.Context, event fwkdl.NotificationEvent) error {
```
elevran, line 69: `e.pool.PoolTargetPorts()` is called multiple times in this function (`EventDelete`, the not-ready branch, and inside `podToEndpointMetadata`). If `TargetPorts` changes between calls (e.g. `InferencePoolReconciler` fires `PoolSet` with a new port list concurrently), different port sets are used for the delete and upsert sides of the same event. Snapshot once at the top of the function and pass `targetPorts` through all three call sites.

pkg/epp/framework/interface/datalayer/discovery_test.go
```diff
@@ -0,0 +1,83 @@
...
+	id := types.NamespacedName{Name: "ep-0", Namespace: "default"}
+	notifier.Delete(id)
+
+	assert.Len(t, store.deleted, 1)
+	assert.Equal(t, id, store.deleted[0])
+	assert.Empty(t, store.upserted)
+}
+
+func TestNewDiscoveryNotifier_Order(t *testing.T) {
+	store := &fakeBackendStore{}
```
elevran: This test asserts `len(store.upserted) == 1` and `len(store.deleted) == 1` but does not verify the order. An implementation that batched or reversed the calls would pass. Consider using the `trackingNotifier` pattern from `file_discovery_test.go` to record operation order and assert the upsert index is less than the delete index.

cmd/epp/runner/runner.go
ezrasilvera: Yes. That's the exact same logic as was before. I just pushed a small fix as originally the POD_NAME was stripped

pkg/epp/datastore/datastore.go
```diff
@@ -87,6 +87,14 @@ type Datastore interface {
 	PodUpdateOrAddIfNotExist(ctx context.Context, pod *corev1.Pod) bool
 	PodDelete(podName string)
 
+	// PoolTargetPorts returns the target ports of the current pool, or nil if not synced.
+	PoolTargetPorts() []int
+
+	// BackendUpsert adds or updates an endpoint from a non-Kubernetes discovery source.
```
elevran, line 80: Does it make sense to allow both `Backend` and `Pod` functions? This leads to duplicate/diverging paths. Can one be implemented via the others? For example: `PodUpdateOrAddIfNotExist` converts Pod to `fwkdl.EndpointMetadata` and then calls `BackendUpsert()`? Same for delete path.

cmd/epp/runner/runner.go
ezrasilvera: fixed

apix/config/v1alpha1/endpointpickerconfig_types.go
ezrasilvera: One issue with that is that DataLayer *DataLayerConfig is already an optional pointer. If discovery moves inside it a user who only wants a custom discovery plugin (no custom sources) would need to write:

dataLayer:
  discovery:
    pluginRef: my-file-discovery

instead of the current flat:

discovery:
  pluginRef: my-file-discovery

cmd/epp/runner/rungroup.go
ezrasilvera: do you mean internal/runnable ?  
It us used only by  cmd/epp/runner/runner.go  so maybe move it to  cmd/epp/runner ?

elevran (commented):
Thanks @ezrasilvera!

The overall architecture is sound: `baseK8sDiscovery.runManager` cleanly factors out shared pod-watching code, the file-based plugin is well-tested, and the `DiscoveryNotifier`/`BackendUpsert`/`BackendDelete` additions are clean. Several correctness issues need fixing before merging. See inline comments for four of these.

**pprof and `/metrics` are served unauthenticated. This is a regression from old path**
Check `cmd/epp/runner/runner.go`  new `serveMetrics` function.
The old `setup()` path registered pprof via `profiling.SetupPprofHandlers(mgr)` on the controller-runtime manager, which consults `--secure-serving` and `MetricsEndpointAuth`. The new `serveMetrics` uses a plain `http.Server` with no auth regardless of `opts.MetricsEndpointAuth`. With `EnablePprof=true` heap/goroutine profiles are exposed unauthenticated. Also, importing `"net/http/pprof"` as `nhpprof` registers all standard pprof endpoints on `http.DefaultServeMux` via its `init()`. The new path needs to honor `opts.MetricsEndpointAuth` for `/metrics` and gate pprof behind the same auth filter as before.

**`ExtractNotification` which is the core event-handling logic, is entirely untested**
`pkg/epp/framework/plugins/datalayer/discovery/k8s/pod_extractor_test.go`

All branches unexercised: `PoolHasSynced()==false` early return; `EventDelete`; pod not-ready/label-mismatch delete; happy-path upsert. A fake `PoolInfoProvider` is trivial; `NotificationEvent` wraps `*unstructured.Unstructured`.

**`BackendUpsert` and `BackendDelete` have no tests**
`pkg/epp/datastore/datastore_test.go`

These are the new entry points for the entire discovery layer. No test covers: new endpoint creation, update-on-existing, `NewEndpoint` returning `nil`, delete-existing, delete-missing (no-op).

**CLI `--pool-name` silently ignored when `discovery.pluginRef` is set without `poolName` param**
`cmd/epp/runner/runner.go` - `resolveDiscovery`

The docs table says `poolName` defaults to "from `--pool-name` flag" for `inference-pool-discovery`. The code doesn't implement this: when `discovery.pluginRef: inference-pool-discovery` is in config with no `poolName` param, `--pool-name` is silently ignored and `Start()` fails at runtime with "poolName is required". Either implement the fallback (`ip.PoolName = opts.PoolName` when factory left it empty) or correct the docs and emit a warning log when both are provided.

**`DatalayerBinder`/`DatastoreProvider` defined in the K8s plugin package, not the framework interface layer**
`pkg/epp/framework/plugins/datalayer/discovery/k8s/k8s_discovery.go` - `DatastoreProvider`/`DatalayerBinder` interfaces

These are framework-level contracts but defined in a leaf plugin package. The runner imports `k8sdiscovery` solely for these type assertions, and any non-K8s plugin author who needs datalayer injection must import an internal K8s package. Move to `pkg/epp/framework/interface/datalayer/discovery.go`.

**`InferencePoolDiscoveryPlugin` exports struct fields; all other plugins in this codebase don't**
In `k8s_discovery.go` the `PoolName`, `PoolNamespace`, `PoolGroup`, `LeaderElect` are exported only because tests access them via type cast. Every other plugin uses unexported fields.

**`rungroup` package has no tests**
`internal/rungroup/rungroup.go` — new package with no `_test.go`. The error-propagation contract (one failure cancels context for all) is untested.

- **File size limit**: `os.ReadFile` in `file_discovery.go:load()` has no upper bound; with `watchFile: true` a large file write can exhaust heap. Use `io.LimitReader` (e.g. 1 MiB cap).
- **Validate selector at factory time**: `StaticSelectorFactory` defers `ConvertSelectorToLabelsMap` to `Start()`; a malformed selector passes config load but fails at runtime. Move to `StaticSelectorFactory`.
- **`serveMetrics` shutdown timeout**: `srv.Shutdown(context.Background())` can block indefinitely; use a 5-second timeout context.

apix/config/v1alpha1/endpointpickerconfig_types.go
elevran: Saves a single configuration line. If we implement discovery in the datalayer's framework hierarchy I think it makes sense to have it follow the same in config,

cmd/epp/runner/runner.go
elevran: I see.
I wonder if it is not a typo that should has escaped. The use of `POD_NAME` right after `if poolName == ""` suggested a typo. What do you think?

cmd/epp/runner/rungroup.go
elevran: yes - `internal/runnable`. But your suggestion is better.

cmd/epp/runner/runner.go
ezrasilvera: Fixed in a5e31c4a

pkg/epp/framework/plugins/datalayer/discovery/k8s/k8s_discovery.go
ezrasilvera: Fixed in a169f94e

pkg/epp/framework/plugins/datalayer/discovery/k8s/k8s_discovery.go
ezrasilvera: I Agree, but prefer to defer to follow-up PR. Will open an issue to track it. 
The three reconcilers each have non-trivial logic. Converting them is a major refactor that would require separate review. 
Adding  a TODO comment above each SetupWithManager call:

// TODO: convert to NotificationExtractor via BindNotificationSource for
// consistency with pod-discovery.

ezrasilvera: Added TODO comments on all three reconcilers in 68bbdef3. Full conversion to NotificationExtractor will be done in a follow-up.

pkg/epp/framework/plugins/datalayer/discovery/file/file_discovery.go
ezrasilvera: Fixed in 81f2ce78

pkg/epp/framework/plugins/datalayer/discovery/k8s/k8s_discovery.go
ezrasilvera: k8s discovery: validate StaticSelector endpoint selector at construction time. Fixed in 618e5026

pkg/epp/framework/interface/datalayer/discovery_test.go
ezrasilvera: operation order in TestNewDiscoveryNotifier_Order. Fixed in 732df12b

cmd/epp/runner/rungroup.go
ezrasilvera: runner: move rungroup into cmd/epp/runner package. Fixed in 3a300ce9

pkg/epp/datastore/datastore.go
ezrasilvera: datastore: extract upsertEndpoint helper shared by BackendUpsert and podUpdateOrAddIfNotExist. Fixed in 611cc04d

ezrasilvera:
datastore: add tests for BackendUpsert and BackendDelete. Fixed in a6d120c0

pkg/epp/framework/plugins/datalayer/discovery/k8s/pod_extractor.go
ezrasilvera: k8s discovery: snapshot PoolTargetPorts once per ExtractNotification to prevent race. Fixed in a610197b

pkg/epp/framework/plugins/datalayer/discovery/k8s/pod_extractor.go
ezrasilvera: k8s discovery: use active-ports annotation in podToEndpointMetadata. Fixed in 9ad8fbaf.

ezrasilvera:
@elevran Thanks for the deep review.  I replied your inline comments, and this is for the additional list of issues.

> **pprof and `/metrics` are served unauthenticated. This is a regression from old path**
> Check `cmd/epp/runner/runner.go`  new `serveMetrics` function.
> The old `setup()` path registered pprof via `profiling.SetupPprofHandlers(mgr)` on the controller-runtime manager, which consults `--secure-serving` and `MetricsEndpointAuth`. The new `serveMetrics` uses a plain `http.Server` with no auth regardless of `opts.MetricsEndpointAuth`. With `EnablePprof=true` heap/goroutine profiles are exposed unauthenticated. Also, importing `"net/http/pprof"` as `nhpprof` registers all standard pprof endpoints on `http.DefaultServeMux` via its `init()`. The new path needs to honor `opts.MetricsEndpointAuth` for `/metrics` and gate pprof behind the same auth filter as before.

runner: honor MetricsEndpointAuth in serveMetrics, fix shutdown timeout. Fixed in d40587ac

> **`ExtractNotification` which is the core event-handling logic, is entirely untested**
> `pkg/epp/framework/plugins/datalayer/discovery/k8s/pod_extractor_test.go`
>
> All branches unexercised: `PoolHasSynced()==false` early return; `EventDelete`; pod not-ready/label-mismatch delete; happy-path upsert. A fake `PoolInfoProvider` is trivial; `NotificationEvent` wraps `*unstructured.Unstructured`.

k8s discovery: add ExtractNotification tests covering all branches. Fixed in 07014848

> **`BackendUpsert` and `BackendDelete` have no tests**
> `pkg/epp/datastore/datastore_test.go`
>
> These are the new entry points for the entire discovery layer. No test covers: new endpoint creation, update-on-existing, `NewEndpoint` returning `nil`, delete-existing, delete-missing (no-op).

datastore: add tests for BackendUpsert and BackendDelete. Fixed in a6d120c0

> **CLI `--pool-name` silently ignored when `discovery.pluginRef` is set without `poolName` param**
> `cmd/epp/runner/runner.go` - `resolveDiscovery`
>
> The docs table says `poolName` defaults to "from `--pool-name` flag" for `inference-pool-discovery`. The code doesn't implement this: when `discovery.pluginRef: inference-pool-discovery` is in config with no `poolName` param, `--pool-name` is silently ignored and `Start()` fails at runtime with "poolName is required". Either implement the fallback (`ip.PoolName = opts.PoolName` when factory left it empty) or correct the docs and emit a warning log when both are provided.

runner: inject --pool-name into InferencePoolDiscoveryPlugin when poolName param is omitted. Fixed in 0e3388ea

> **`DatalayerBinder`/`DatastoreProvider` defined in the K8s plugin package, not the framework interface layer**
> `pkg/epp/framework/plugins/datalayer/discovery/k8s/k8s_discovery.go` - `DatastoreProvider`/`DatalayerBinder` interfaces
>
> These are framework-level contracts but defined in a leaf plugin package. The runner imports `k8sdiscovery` solely for these type assertions, and any non-K8s plugin author who needs datalayer injection must import an internal K8s package. Move to `pkg/epp/framework/interface/datalayer/discovery.go`.

discovery: move DatastoreProvider and DatalayerBinder to inject package. Fixed in 3f94d112

> **`InferencePoolDiscoveryPlugin` exports struct fields; all other plugins in this codebase don't**
> In `k8s_discovery.go` the `PoolName`, `PoolNamespace`, `PoolGroup`, `LeaderElect` are exported only because tests access them via type cast. Every other plugin uses unexported fields.

k8s discovery: unexport InferencePoolDiscoveryPlugin fields, add GetPoolName/SetPoolName accessors. Fixed in a47abe05

> **`rungroup` package has no tests**
> `internal/rungroup/rungroup.go` — new package with no `_test.go`. The error-propagation contract (one failure cancels context for all) is untested.

runner: add tests for RunnableGroup error propagation contract. Fixed in c37ba409

> - **File size limit**: `os.ReadFile` in `file_discovery.go:load()` has no upper bound; with `watchFile: true` a large file write can exhaust heap. Use `io.LimitReader` (e.g. 1 MiB cap).

file-discovery: cap endpoints file reads at 1 MiB to prevent heap exhaustion. Fixed in 0f216451

> - **Validate selector at factory time**: `StaticSelectorFactory` defers `ConvertSelectorToLabelsMap` to `Start()`; a malformed selector passes config load but fails at runtime. Move to `StaticSelectorFactory`.

k8s discovery: validate StaticSelector endpoint selector at construction time. Fixed in 618e5026

> - **`serveMetrics` shutdown timeout**: `srv.Shutdown(context.Background())` can block indefinitely; use a 5-second timeout context.

runner: honor MetricsEndpointAuth in serveMetrics, fix shutdown timeout. Fixed in d40587ac

apix/config/v1alpha1/endpointpickerconfig_types.go
ezrasilvera: You're right that in the code it lives under pkg/epp/framework/plugins/datalayer/discovery, so conceptually it's a datalayer concern. But I think there are a few reasons to keep it top-level beyond just the one config line:

The bigger issue for me is as I pointed previously is that DataLayer is itself an optional pointer. If someone just wants a file-discovery and no custom metrics sources, they currently have no reason to write a dataLayer block at all. For most users dataLayer mainly means metrics collection It would look like discovery is a metrics concern, which it isn't.
Also on the runner , we currently just need to check rawConfig.Discovery != nil. If discovery moves under DataLayer we need two checks  rawConfig.DataLayer != nil &&  rawConfig.DataLayer.Discovery != nil . While this is a small change it adds a null-dereference risk if someone forgets it.
Looking at the other top-level fields, Parser, FlowControl, SaturationDetector they are all subsystems that live in their own implementation packages too, but they are kept as top-level because they are primary EPP-level knobs. Discovery seems the same to me.
I just wanted to make sure you are aware of the concerns, nevertheless if you still prefer to move it I have no problem doing it.

ezrasilvera:
Hi @elevran I accepted all your comments and suggestions  (except 2: one I suggested to defer and one is waiting for your response).

elevran:
I am fine with leaving discovery a top-level configuration aspect, not under the data layer.

ahg-g:
I was planning to send out a dev guide, but didn't get a chance because we are still working on the llm-d release.

I wanted to suggest to have the configuration API aligned with the four core components of EPP: RequestHandler, FlowControl, Scheduler and Datalayer. We need to refactor the existing structure to align with this suggestion as well.

elevran:
@ezrasilvera note DCO status.

ezrasilvera:
> I was planning to send out a dev guide, but didn't get a chance because we are still working on the llm-d release.
> 
> I wanted to suggest to have the configuration API aligned with the four core components of EPP: RequestHandler, FlowControl, Scheduler and Datalayer. We need to refactor the existing structure to align with this suggestion as well.

@ahg-g sounds reasonable. Do you think it's a blocking issue for this PR, or we can make any needed change in following PRs ?  We have have been asked for this ability for couple of concrete use cases that are waiting for it.

ahg-g:
> @ahg-g sounds reasonable. Do you think it's a blocking issue for this PR, or we can make any needed change in following PRs ? We have have been asked for this ability for couple of concrete use cases that are waiting for it.

It has to wait for a release anyway for any one to use it. I am fine if you would like to do that in a follow up, but we should aim to do that before the next release.

ahg-g:
Created https://github.com/llm-d/llm-d-inference-scheduler/issues/1041

ezrasilvera:
> > @ahg-g sounds reasonable. Do you think it's a blocking issue for this PR, or we can make any needed change in following PRs ? We have have been asked for this ability for couple of concrete use cases that are waiting for it.
> 
> It has to wait for a release anyway for any one to use it. I am fine if you would like to do that in a follow up, but we should aim to do that before the next release.

@ahg-g In your opinion the discovery should be under DataLayer or a top level ?  See my comment to @elevran here https://github.com/llm-d/llm-d-inference-scheduler/pull/1007#discussion_r3213056439
If you strongly prefer to move the discovery to under datalayer I have no problem making that change as part of this PR.

ezrasilvera:
> It has to wait for a release anyway for any one to use it. 

@ahg-g Good that you mentioned the release, it reminded me that we might need to also add the appropriate guides and documentation. @elevran We probably need to define an issue covering everything you think is required for the release to expose this new capability so we can plan for it. Thanks.

ahg-g:
Thanks @ezrasilvera , yes it would be good to do it in this PR.

apix/config/v1alpha1/endpointpickerconfig_types.go
ezrasilvera: Done. Discovery was moved to under Datalayer in the config and is no longer top level  (cc: @elevran @ahg-g )

cmd/epp/runner/runner.go
```diff
@@ -240,6 +464,8 @@ func (r *Runner) Run(ctx context.Context) error {
 //
 // The returned Datastore is **only** meant to be used in the integration test.
 // Optional managerOverrides are applied to the controller manager options before creation.
+// setup is kept for integration tests that need to run against a live K8s cluster.
+// Production code uses Run() instead.
```
ahg-g, line 468: This divergence is dangerous, integration tests should as much as possible test the exact same path as the one executed in reality. We worked carefully to make that happen in the past after a number of regressions and we need to maintain this principle.

cmd/epp/runner/runner.go
```diff
@@ -186,51 +194,267 @@ func (r *Runner) Run(ctx context.Context) error {
...
 
-	pmc, err := backendmetrics.NewPodMetricsClientImpl(setupLog, backendmetrics.Config{
-		ModelServerMetricsScheme:        opts.ModelServerMetricsScheme,
-		ModelServerMetricsHTTPSInsecure: opts.ModelServerMetricsHTTPSInsecure,
-		ModelServerMetricsPath:          opts.ModelServerMetricsPath,
+	namespace := resolvePoolNamespace(opts.PoolNamespace)
+	poolName := opts.PoolName
+	if poolName == "" {
+		poolName = "epp"
+	}
```
ahg-g, line 215: I recommend against defaulting in random places, if this is read from a flag, then the flag itself should set  the default

cmd/epp/runner/runner.go
```diff
@@ -186,51 +194,267 @@ func (r *Runner) Run(ctx context.Context) error {
...
 
-		TotalQueuedRequestsMetric:    opts.TotalQueuedRequestsMetric,
-		TotalRunningRequestsMetric:   opts.TotalRunningRequestsMetric,
-		KVCacheUsagePercentageMetric: opts.KVCacheUsagePercentageMetric,
-		LoRAInfoMetric:               opts.LoRAInfoMetric,
-		CacheInfoMetric:              opts.CacheInfoMetric,
-	})
+	useNewMetrics := !r.featureGates[datalayer.EnableLegacyMetricsFeatureGate]
```
ahg-g, line 217: I suggest we deprecate this path in a separate PR and not bother with it here. @elevran wdyt?

ahg-g (commented):
Thanks @ezrasilvera for working on this. This PR is very large and I advise against merging it as is and recommend to break it into smaller PRs. Before we add new discovery plugins, we better have a couple of PRs that refactor the code to allow adding them, for example: 
- A PR to refactor the startup code
- A PR to refactor the existing discovery mechanism into the new plugin architecture
- A PR to add the new discovery plugins.

ezrasilvera:
> Thanks @ezrasilvera for working on this. This PR is very large and I advise against merging it as is and recommend to break it into smaller PRs. Before we add new discovery plugins, we better have a couple of PRs that refactor the code to allow adding them, for example:
> 
> * A PR to refactor the startup code
> * A PR to refactor the existing discovery mechanism into the new plugin architecture
> * A PR to add the new discovery plugins.

Hi @ahg-g , We do need the file discovery for the coming release, so my plan is to hold with this PR, and instead push several small PRs that will add a filediscovery plugin without changing the k8s behavior at all (not moving the k8s into plugins).  The actuall wiring of the filediscovey plugin in the runner will be in the last PR to minimize the impact even further. 
We can then start slowly as you suggested refactoring the existing k8s code into the new discovery plugin framework through smaller PRs.   Thanks for the input.

ezrasilvera:
Don't merge this PR for now. Instead we will interduce the change in stages starting from the file discovery. 
The first two PRs in the path to support filediscovery can be found in #1080 and #1081

cmd/epp/runner/runner.go
elevran: this is to be removed in v0.10, tracked separately. Part of removing the whole `backend.metrics` package.

ezrasilvera:
Replace with teh file discovery PRs

Issue #988, opened 2026-05-05
Fix flaky test "Verifying that all expected metrics are present."


roytman:
After migration, GAIE e2e tests, the "Verifying that all expected metrics are present" is flaky, and requires a separate investigation.
```
[FAILED] in [It] - /app/test/e2e/epp/e2e_test.go:516 @ 05/04/26 12:55:35.582
  STEP: Deleting the InferenceObjective test resource. @ 05/04/26 12:55:35.582
  STEP: Restoring vLLM Deployment and InferencePool. @ 05/04/26 12:55:35.847
  STEP: Waiting for Deployment inf-ext-e2e/vllm-qwen3-32b to complete rollout @ 05/04/26 12:55:35.858
  STEP: Deployment rollout complete @ 05/04/26 12:55:39.396
• [FAILED] [199.479 seconds]
InferencePool when The Inference Extension is running [It] Should expose EPP metrics after generating traffic
/app/test/e2e/epp/e2e_test.go:198

  [FAILED] Timed out after 180.003s.
  Expected success, but got an error:
      <*errors.errorString | 0xc000a80090>: 
      expected metric inference_objective_request_total not found in metrics output
      {
          s: "expected metric inference_objective_request_total not found in metrics output",
      }
  In [It] at: /app/test/e2e/epp/e2e_test.go:516 @ 05/04/26 12:55:35.582
------------------------------

```

elevran:
@roytman can you please point to specific failure run? It helps in diagnosing the issue.

elevran:
#992 fixes a race condition but do not know if it was the same root cause. If it is the same root cause - please close the issue.

roytman:
@elevran, I saw this error once, too, but it was not reproduced. I added the error log

elevran:
Understood. If this (or another flake) comes up again, please include a link to the action's run so we have as much context as possible

revit13:
Investigating two potential bugs that could trigger this flaky error.  Both bugs scenarios are now verified by new unit tests, which correctly fail without a fix.

## Failure Flow

```
BeforeEach (runs before every It)
│
├─ DeleteAllOf(EPP pods)                     [line 116]
│   └─ Pod gets deletionTimestamp; starts graceful shutdown
│      BUT remains Ready=True for several seconds
│
├─ waitForDeploymentReady()                  [line 120]
│   └─ Checks ReadyReplicas > 0
│      BUG 1: The old terminating pod still counts as Ready
[...]

PR #1013, opened 2026-05-07
fix(e2e/epp): treat non-HTTP-200 as failure in generateTraffic

revit13:
This PR:

- Fix `generateTraffic` validation: Previously, the function only checked if the curl command executed successfully (err == nil). Because curl exits with a 0 even for HTTP errors (like 503s), this caused false positives. The function now explicitly validates the HTTP response codes.

- Test structure refactor: Separated the core test logic and utility functions into distinct files to improve code organization and maintainability.

**What type of PR is this?**
Add one of the following kinds:
/kind bug
/kind test

A bug reveled while investigating #988 

```release-note
NONE
```


roytman:
There are a lot of repetitive strings in these files. Could you define them as constants instead?

test/e2e/epp/utils_test.go
```diff
@@ -0,0 +1,566 @@
...
+
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	v1 "sigs.k8s.io/gateway-api-inference-extension/api/v1"
+	"sigs.k8s.io/gateway-api-inference-extension/apix/v1alpha2"
```
roytman, line 37: Once PR #989 merges, this import will need to be updated to github.com/llm-d/llm-d-inference-scheduler/apix/v1alpha2

Issue #1003, opened 2026-05-06
Approximate Prefix Cache Aware Plugin optimization for multimodal


capri-xiyue:
**What would you like to be added**:
We need to optimize it with Optimization Phase 2: Guess MultiModal 
Token Counts , see
https://docs.google.com/document/d/1eFetVjTQR22IkcW4CYvtY0KnVRX0jbwnUrfzzkugtRA/edit?tab=t.0#bookmark=id.in3e8jy9j768 which mitigate the above two limitation

Basically have Approximate Prefix Cache Aware Plugin multimodal support for two patterns:
1. Dimension-based scaling: Calculating the token count by dividing the image width and height by a specific factor (e.g., Qwen VL 2.5 and 3.5). Factor can vary among different models
2. Fixed count: Allowing the user to specify a hard limit for the number of tokens (e.g., Gemma 4).

This issue is to support both modes. Users will have the flexibility to customize either the scaling factor or the fixed token count, and the system will apply default values if no preference is set. As a future enhancement, EPP could monitor recent multimodal requests to learn and cache the appropriate token counts for specific image dimensions.

We will mimic how vllm does using 
1. multimodal hash
2. number of placeholder tokens
for the approximate prefix cache aware plugin.  See details [here](https://docs.vllm.ai/en/latest/design/prefix_caching/)

**Why is this needed**:

Currently EPP approximate cache aware plugin just use image/audio/video url/content when calculating in prefix match score. It has some limitation
1. may accelerate hot-spotting issue either because of fail to consider correct image/audio/video token weight or because of image content partial metadata match even the whole image are different
2. lru eviction behavior will have large difference when compared to real model server eviction behavior

Guess MultiModal Token Counts approach will address such limitations, you can see micro-benchmarking here https://docs.google.com/document/d/1Xbnu0G3FbG_RXtQnCRENfvHwxbBS0iOGIUZm3lKEf0w/edit?usp=sharing&resourcekey=0-AgTY4KsPeOmD8u-JPdcUdA

hacky poc pr: https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2904 (Please ignore the codestyle)

capri-xiyue:
assign @capri-xiyue

capri-xiyue:
cc @nicolexin

rahulgurnani:
nit: Rename to "Approximate prefix cache"

[...]

PR #1018, opened 2026-05-07
feat: added tokenplaceholder guess algorithm with image content in chat completion api for approximate prefix cache aware plugin

capri-xiyue:
**What type of PR is this?**

part of #1003 
This PR just handle image type in chat completion api. Video or Audio support will get supported later.
I use the same concept in https://docs.vllm.ai/en/latest/design/prefix_caching/.

This branch introduces multimodal input support (specifically image_url blocks in Chat Completions requests) into the Approximate Prefix Cache Scheduler routing plugin.

It allows the scheduler to estimate token counts for multimodal assets (images), serialize them into stable binary token signatures, and perform token-boundary aligned prefix caching index lookups.

Key Features Implemented
1. Multimodal Token Estimator 
Purpose: Estimates the number of tokens consumed by structured content blocks (like images) before actual tokenization occurs.

2. Multimodal Hash 
For structured image_url inputs, calculates a stable xxhash of the URL, use the hash as a 1-token placeholder repeated for numPlaceholders times in the PseudoTokens buffer.

Content-Boundary Alignment: Appends null-byte (0x00) padding to the text buffer right before appending any multimodal hash bytes. This ensures image hashes always align perfectly with 4-byte token boundaries and never get mixed with preceding text characters inside a token 

Fixes #

```release-note
Approximate prefix cache aware plugin has new config parameters related to multimodal

for model that the image tokens are fixed(can be set by user) like gemma4,

multiModalTokenEstimator:
  image:
    mode: static
    defaultResolution:
      width: 1024
      height: 768
    static:
      staticToken: 512
  
  This config means default image width and height is 1024*768 when epp can't fetch the image dimension from image url/content , and for each image, it has 512 tokens
  
multiModalTokenEstimator:
  image:
    mode: dynamic
    defaultResolution:
      width: 1024
      height: 768
    dynamic:
      factor: 1024  
   
  This config means default image width and height is 1024*768 when epp can't fetch the image dimension from image url/content , and for each image, it has image_width*image_height/factor tokens
```


capri-xiyue:
cc @nicolexin @liu-cong @ahg-g for early review for the major direction. Want to gather early feedback before I jump to more UT.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -118,19 +119,163 @@ func getUserInputBytes(request *scheduling.InferenceRequest) ([]byte, error) {
...
+	return blocks
+}
+
+func getContentBlocksFromChatCompletions(ctx context.Context, request *scheduling.InferenceRequest, blockSizeTokens int, multimodalConfig *multiModalConfig) []ContentBlock {
+	loggerDebug := log.FromContext(ctx).V(logutil.DEBUG)
+	messages := request.Body.ChatCompletions.Messages
+	var allTokens []string
+	var tokenImageHashes []string
+
+	for _, msg := range messages {
```
capri-xiyue, line 194: For this part, I will try to simplify the code and make it more readable, you can refer to UT to see its effect instead of reviewing the code style here

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -118,19 +120,166 @@ func getUserInputBytes(request *scheduling.InferenceRequest) ([]byte, error) {
...
 
 	default:
 		return nil, errors.New("invalid request body: no recognized API format found")
 	}
 }
+
+func getContentBlocksFromRawBytes(rawBytes []byte, blockSizeTokens int) []ContentBlock {
+	rawStr := string(rawBytes)
+
+	var tokens []string
```
liu-cong: calling this tokens is misleading

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -118,19 +120,166 @@ func getUserInputBytes(request *scheduling.InferenceRequest) ([]byte, error) {
...
+	for i := 0; i < len(rawStr); i += averageCharactersPerToken {
+		end := i + averageCharactersPerToken
+		if end > len(rawStr) {
+			end = len(rawStr)
+		}
+		tokens = append(tokens, rawStr[i:end])
+	}
+
+	var blocks []ContentBlock
+	for i := 0; i < len(tokens); i += blockSizeTokens {
```
liu-cong: I think we should be able to do this in just one loop, with loop step = blockSizeTokens*averageChractersPerToken

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -118,19 +120,166 @@ func getUserInputBytes(request *scheduling.InferenceRequest) ([]byte, error) {
...
+		}
+		return getContentBlocksFromRawBytes(rawBytes, blockSizeTokens), nil
 
 	default:
 		return nil, errors.New("invalid request body: no recognized API format found")
 	}
 }
+
+func getContentBlocksFromRawBytes(rawBytes []byte, blockSizeTokens int) []ContentBlock {
+	rawStr := string(rawBytes)
```
liu-cong: let's avoid new memory allocation. we should not introduce additional memory footprint than  the original input and the new hash array

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -118,19 +120,166 @@ func getUserInputBytes(request *scheduling.InferenceRequest) ([]byte, error) {
...
+		contentBlocks = append(contentBlocks, ContentBlock{
+			TokenIDs:  blockTokens,
+			ExtraHash: extraHashes,
+		})
+	}
+
+	return contentBlocks
+}
+
+func getImageDimensionsFromBase64(url string) (*resolution, error) {
```
liu-cong: Please consider reusing the TokenEstimator interface https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/token_estimator.go

And create a MultiModalTokenEstimator implementation

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -29,44 +38,42 @@ import (
 	"github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/scheduling"
 )
 
+const (
+	multiModalPlaceHolder = "P"
+)
+
+type ContentBlock struct {
```
liu-cong: Suggestions:
We model this as a 2-step process: 1. Break input (text or token) into Blocks and 2. hash each block.

After step1, text or token input should share the same step 2.

For step1, we should define ContentBlock to support both text (`[]byte`) and tokens (`[]uint32`). Do we need `ExtraHash` to be an array, or just  a single vlaue? Another option is to just get rid of the placeholder, and simply fill the content with the `ExtraHash` value for multi modal

capri-xiyue: Wanna clarify what you meant here.

>  we should define ContentBlock to support both text ([]byte) and tokens ([]uint32)

Currently what I did is to convert either text(using 4 character to represent one token) or image tokens to TokenIDs  []string at blocksize and then hash each block?  So what's your suggestion here?

>  Do we need ExtraHash to be an array

I just make it an array to be more clear and human readable. In vllm, they also make it a tuple instead of a string https://github.com/vllm-project/vllm/blob/4140faa4a51d42cb9618949bee28fd47682f611c/vllm/v1/core/kv_cache_utils.py#L393
Is there any cons to make ExtraHash to be an array here?

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: Do you have any suggestions, here I just make four characters as one token, so I call it token? Do you think tokenID is better?

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: Discussed offline. 
For the contentBlock, it will have tokens([]uint32) when we have tokenized prompt or pseudoTokens([]string) when we just have raw prompt

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: Discussed offline,  https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/token_estimator.go is not a interface that suit here. I will write a new token_estimator interface that estimate tokens given text or image or video

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: called it pseudoTokens

rahulgurnani:
Do you happen to have the micro-benchmark numbers handy? Adding them to the PR description would provide some great context for the performance improvements. Thanks!

capri-xiyue:
> Do you happen to have the micro-benchmark numbers handy? Adding them to the PR description would provide some great context for the performance improvements. Thanks!

Previously had some micro-benchmark as below. And I will work on generate more representative datasets to have a full benchmark

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-3a9609e3-7fff-1af0-8b17-c6fd4337e8e4"><div dir="ltr" style="margin-left:-63.75pt;" align="left">
Target RPS | Configuration | p75 TTFT (s) | P95 TTFT (s) | Total Prompt Tokens | Cached Prompt Tokens | Uncached (Raw) Prompts
-- | -- | -- | -- | -- | -- | --
30 | EPP POC  (Image Token Guess) | 0.725 | 1.639 | 1,966,961 | 946,640 | 361,109
  | K8s Service | 0.942 | 1.552 | 1,965,871 | 395,728 | 254,895
  | EPP  v1.5.0 (image url/content directly) | 0.727 | 2.510 | 1,966,961 | 994,752 | 412,684
40 | EPP POC  (Image Token Guess) | 0.726 | 1.418 | 2,482,727 | 1,213,664 | 572,187
  | K8s Service | 0.963 | 1.328 | 2,554,807 | 483,472 | 413,911
  | EPP  v1.5.0 (image url/content directly) | 1.094 | 3.720 | 2,482,727 | 1,249,712 | 501,585
50 | EPP POC  (Image Token Guess) | 1.001 | 1.595 | 3,261,332 | 1,554,784 | 750,247
  | K8s Service | 1.713 | 2.510 | 3,259,184 | 605,920 | 516,040
  | EPP  v1.5.0 (image url/content directly) | 1.227 | 2.066 | 3,261,332 | 1,584,240 | 899,559
60 | EPP POC  (Image Token Guess) | 1.992 | 3.237 | 3,783,409 | 1,635,872 | 827,166
  | K8s Service | 3.339 | 3.745 | 3,783,377 | 684448 | 568254
  | EPP  v1.5.0 (image url/content directly) | 4.122 | 5.573 | 3,783,409 | 1,639,056 | 805,182

</div><br /><br /><span style="font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:italic;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">Note: In the table, the sum of Cached Prompt Tokens and Uncached (Raw) Prompts does not equal the Total Prompt Tokens. This discrepancy happens because vLLM may fail to report cached token details, even when the </span><span style="font-size:11pt;font-family:'Roboto Mono',monospace;color:#188038;background-color:transparent;font-weight:400;font-style:italic;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">--enable-prompt-tokens-details</span><span style="font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:italic;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;"> flag is enabled. However, for requests where cached tokens are successfully reported, the figures align perfectly with the dataset I provided. Therefore, the overall data still provides a valid and useful signal.</span></b>

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -21,52 +21,61 @@ import (
...
 )
 
-// hashPrompt divides the prompt into blocks and calculates a prefix cache hash for each block.
+const (
+	multiModalPlaceHolder = "P"
+)
+
+type KVCacheBlock struct {
+	// PseudoTokens are used when we just have prompts
+	PseudoTokens []string `json:"PseudoTokens"`
```
nicolexin: How about calling it EstimatedTokens? Also can we have more description on "what it is" in addition to "when it is used".

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -21,52 +21,61 @@ import (
...
+	// PseudoExtraHash are used when we just have prompts
+	PseudoExtraHash []string `json:"PseudoExtraHash"`
```
nicolexin: Shall we call it MultimodalHashes?

Similar to above, can we have more descriptions?

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
```diff
@@ -113,6 +113,60 @@ const (
...
+type multiModalConfig struct {
+	Image *imageConfig `json:"image,omitempty"`
+	Video *videoConfig `json:"video,omitempty"`
+}
+
+// defaultMultimodalConfig provides default configuration for multimodal inputs.
+var defaultMultimodalConfig = multiModalConfig{
+	Image: &imageConfig{
+		Mode: ModeFixed,
+		DefaultResolution: resolution{
```
nicolexin, line 156: Have we done some research on the default numbers here? Since it's configurable we don't need to make it perfect but let's start with some estimates. 

I think as a follow up we need to have a multimodel guide on how this can be configured and where to find information on whether a model is fixed vs dynamic and how to configure / find out the factor.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: I keep it consistent with https://docs.vllm.ai/en/latest/design/prefix_caching/, in theory, it should carry the hash info needed for this kv cache block, not just multimodal hash.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: I will add more more description on "what it is" in addition to "when it is used". I want to keep the name consistent with extra hash. If we changed the name to estimatedToken here, I feel estimatedExtraHash is also needed, but I feel the name is a little bit confusing.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
capri-xiyue: I will write a multimodal guide regarding configuration and how to figure out whether a model is fixed vs dynamic and how to configure the factor.

For the default numbers, do you think which model we should target in our example guide, seems like qwen3 vl is commonly used.  I think for the default config, we can just let it match our example guide.

nicolexin: Yeah that makes sense, qwen3 vl sounds good to me!

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: added more desp

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/token_estimator.go
```diff
@@ -0,0 +1,120 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// TokenEstimator estimates the number of tokens for different content types.
+type TokenEstimator interface {
+	Estimate(block fwkrh.ContentBlock) int
+}
+
+type approximatePrefixCacheTokenEstimator struct {
+	ctx              context.Context
```
liu-cong, line 44: is this needed?

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
```diff
@@ -113,6 +113,61 @@ const (
...
+// imageConfig defines the configuration for image modality.
+type imageConfig struct {
+	Mode              mode           `json:"mode"`
+	DefaultResolution resolution     `json:"defaultResolution"`
+	DynamicCfg        *dynamicConfig `json:"dynamic,omitempty"`
+	FixedCfg          *fixedConfig   `json:"fixed,omitempty"`
+}
+
+// videoConfig defines the configuration for video modality.
+type videoConfig struct {
```
liu-cong: let's remove until we have something to add

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
```diff
@@ -113,6 +113,61 @@ const (
...
+}
+
+// multiModalConfig defines the configuration for multimodal inputs.
+type multiModalConfig struct {
```
liu-cong:
```suggestion
type multiModalTokenEstimatorConfig struct {
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
```diff
@@ -113,6 +113,61 @@ const (
...
+type fixedConfig struct {
+	FixedToken int `json:"fixedToken"`
+}
+
+type dynamicConfig struct {
+	Factor int `json:"factor"`
+}
+
+// imageConfig defines the configuration for image modality.
+type imageConfig struct {
```
liu-cong:
```suggestion
type imageTokenEstimatorConfig struct {
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
```diff
@@ -113,6 +113,61 @@ const (
...
+// mode defines the mode for image processing.
+type mode string
+
+// resolution defines the width and height of an image.
+type resolution struct {
+	Width  int `json:"width"`
+	Height int `json:"height"`
+}
+
+type fixedConfig struct {
```
liu-cong:
```suggestion
type staticImageTokenEstimatorConfig struct {
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
```diff
@@ -113,6 +113,61 @@ const (
...
+	FixedToken int `json:"fixedToken"`
+}
+
+type dynamicConfig struct {
```
liu-cong:
```suggestion
type dynamicImageTokenEstimatorConfig struct {
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
```diff
@@ -130,6 +185,8 @@ type config struct {
 	MaxPrefixTokensToMatch int `json:"maxPrefixTokensToMatch"`
 	// Max capacity size of the LRU indexer in number of entries per server (pod).
 	LRUCapacityPerServer int `json:"lruCapacityPerServer"`
+	// Multimodal configuration for the plugin.
+	Multimodal *multiModalConfig `json:"multimodal,omitempty"`
```
liu-cong:
```suggestion
	MultimodalTokenEstimator *multiModalConfig `json:"multimodal,omitempty"`
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -76,26 +89,18 @@ func hashPrompt(ctx context.Context, request *scheduling.InferenceRequest, block
...
-		prevBlockHash = res[len(res)-1]
-	}
-
-	// 2. Process any remaining bytes as a partial block
-	if i < len(userInput) {
+	for _, block := range kvCacheBlocks {
 		h.Reset()
-
-		_, _ = h.Write(userInput[i:])
+		blockBytes, _ := json.Marshal(block)
```
liu-cong: we should avoid json marshal here, it can be expensive. We can define a `Hash` method on the block itself and deal with the primitive data types directly (e..g, if pseudo token field is already a byte slice)

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -118,19 +126,145 @@ func getUserInputBytes(request *scheduling.InferenceRequest) ([]byte, error) {
...
+		kvCacheBlocks = append(kvCacheBlocks, KVCacheBlock{
+			PseudoTokens:    blockTokens,
+			PseudoExtraHash: extraHashes,
+		})
+	}
+
+	return kvCacheBlocks
+}
+
+func appendTextTokens(text string, allTokens *[]string, tokenImageHashes *[]string) {
```
liu-cong: Can we pass the slice and return the slice instead of a pointer?

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -21,52 +21,65 @@ import (
...
+	fwkrh "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requesthandling"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
 )
 
-// hashPrompt divides the prompt into blocks and calculates a prefix cache hash for each block.
+const (
+	multiModalPlaceHolder = "P"
+)
+
+type KVCacheBlock struct {
```
liu-cong: To achieve convergence between token and text input, and maintain high performance I suggest the following changes:
 
1. Define a `TokenWrapper` that represents 1 token.
 // TokenWrapper converts either token or text input to a token equivalent.
// For tokenized input, TokenWrapper simply represents one token.
// For text input, TokenWrapper is a slice of the input bytes roughly mapped to one token.
type TokenWrapper  struct {
   Token *uint32 // Set for token input
    PseudoToken []byte // Set for text input
    ExtraHash string // In multi-modality inputs, this is the content hash of the multi-modality input. In such case the token is hard coded as a placeholder value. ExtraHash should be used for prefix matching.
}

Why `TokenWrapper` instead of `TokenBlock`? `TokenWrapper` makes it cleaner when a block consists of non-mm and mm tokens, in such a case it's unclear which `PseudoExtraHash` matches to which token.

cc @vMaroon please also comment here if this aligns with your expectation. I think we can achieve convergence via the `TokenWrapper` approach.

2. Use the generator pattern: https://gobyexample.com/range-over-iterators. This allows the conversion to happen on the fly and significantly reduce the memory footprint. Consider a 1M token input, directly converting it to a list of `TokenWrapper` can easily incur 10s of MB+ memory.  EPP has a memory problem today in high load.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: Using pointer is more efficient as you don't need to do additional struct copy

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: Why do you need `PseudoExtraHash` info to match to which token? Even in vllm, https://docs.vllm.ai/en/latest/design/prefix_caching/, they just have block hash instead of token hash?

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
capri-xiyue: done

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
capri-xiyue: done

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
capri-xiyue: done

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
capri-xiyue: same above

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
capri-xiyue: I changed it to staticTokenEstimatorConfig as multimodal will include image, audio or video

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
capri-xiyue: done

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: synced offline. I will keep the same kv cache block structure here.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: done. Let me know whether it is what you want

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/token_estimator.go
capri-xiyue: Yep, I need it to create a logger to logger some debug info

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -21,52 +21,122 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+	fwkrh "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requesthandling"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
 )
 
-// hashPrompt divides the prompt into blocks and calculates a prefix cache hash for each block.
+const (
+	multiModalPlaceHolder = "MULTIMODAL_P"
```
capri-xiyue: It won't have collision with text value as for text token, it will only include at most 4 chars.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -21,52 +21,89 @@ import (
...
+	var pseudoBytesHash uint64
+	if len(b.PseudoBytes) > 0 {
+		pseudoBytesHash = xxhash.Sum64(b.PseudoBytes)
+	}
+
+	// 2. Hash Tokens
+	var tokensHash uint64
+	if len(b.Tokens) > 0 {
+		h.Reset()
+		for _, t := range b.Tokens {
```
liu-cong: This should be much more performant: `byteSlice := unsafe.Slice((*byte)(unsafe.Pointer(&data[0])), len(data)*4)`

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -21,52 +21,89 @@ import (
...
+			// A uint32 fits perfectly into the first 4 bytes of our 8-byte buffer
+			binary.LittleEndian.PutUint32(buf[:4], t)
+			_, _ = h.Write(buf[:4])
+		}
+		tokensHash = h.Sum64()
+	}
+
+	// 3. Combine the two hash values together stably
```
liu-cong: It should either be text or tokens, so no need to combine, we should return early

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -21,52 +21,89 @@ import (
...
+	// 2. Hash Tokens
+	var tokensHash uint64
+	if len(b.Tokens) > 0 {
+		h.Reset()
+		for _, t := range b.Tokens {
+			// A uint32 fits perfectly into the first 4 bytes of our 8-byte buffer
+			binary.LittleEndian.PutUint32(buf[:4], t)
+			_, _ = h.Write(buf[:4])
+		}
+		tokensHash = h.Sum64()
```
liu-cong: simply return here.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -21,52 +21,89 @@ import (
...
 	"github.com/cespare/xxhash/v2"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+	fwkrh "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requesthandling"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
 )
 
-// hashPrompt divides the prompt into blocks and calculates a prefix cache hash for each block.
+type KVCacheBlock struct {
```
liu-cong:
```suggestion
// HashBlock wraps text or tokens which is later used for calculating prefix hashes.
// Only one of text or tokens should be specified.
type HashBlock struct {
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -21,52 +21,89 @@ import (
...
+	// PseudoBytes represents estimated tokens derived directly from raw prompt.
+	// It is used when the EPP only has access to raw prompts and actual tokenization hasn't occurred.
+	PseudoBytes []byte `json:"PseudoBytes"`
```
liu-cong:
```suggestion
	PseudoTokens []byte `json:"PseudoTokens"`
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: Yep, but is it safe we use `unsafe` here?

go.mod
```diff
@@ -1,9 +1,6 @@
 module github.com/llm-d/llm-d-router
 
-go 1.25.7
-
-// Pinned to go1.25.10 to pick up stdlib fixes reported by govulncheck.
-toolchain go1.25.10
+go 1.25.10
```
capri-xiyue, line 3: This is needed to fix govulncheck

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: done

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: done

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: done

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: done

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
capri-xiyue: done

PR #1025, opened 2026-05-07
fix(e2e): waitForDeploymentReady races on terminating pods after DeleteAllOf.

### LINKED ISSUE #988: Fix flaky test "Verifying that all expected metrics are present."

revit13:
**What type of PR is this?**
Add one of the following kinds:
/kind bug
/kind test

Problem: The existing waitForDeploymentReady function was susceptible to a race condition. It previously only verified that Status.Replicas == Status.ReadyReplicas (and > 0). When triggering a config reload by deleting EPP pods via DeleteAllOf, terminating pods remain in the Ready=True state throughout their graceful shutdown period (default 30s). As a result, the Deployment still counted them as ready, allowing the check to pass prematurely. Subsequent tests (like metrics verification and traffic routing) would then fail because they were connecting to dying pods.

Fix: `deploymentReadyCondition` (extracted from `waitForDeploymentReady` for testability) now enforces pod-level health in addition to replica counts: after the count checks pass, it lists all pods matching the Deployment's label selector and returns an error if any pod has a non-nil DeletionTimestamp. Because the caller wraps the condition in `gomega.Eventually`, this error triggers a retry — the test blocks until every terminating pod is fully removed and the replacement is genuinely healthy.

This PR also:

- Re-enabled the previously flaky metrics spec by removing `ginkgo.Skip(...)` in verifyMetrics. With the readiness fix in place, the scrape now lands against the new EPP pod deterministically.
- Added TestDeploymentReadyCondition (table-driven, fake k8s client) covering the regression case plus the happy path, rollout-starting, and replicas-mismatch cases.

Fixes #988

Closes #988 

```release-note
NONE
```


test/e2e/epp/utils_test.go
```diff
@@ -463,28 +462,50 @@ func waitForDeploymentRollout(tc *igwtestutils.TestConfig, deploy *appsv1.Deploy
...
 
+// deploymentReadyCondition is the testable core of waitForDeploymentReady.
+// It checks replica counts AND that no pod owned by the deployment is still terminating,
+// since terminating pods count toward ReadyReplicas during their graceful-shutdown window.
+func deploymentReadyCondition(tc *igwtestutils.TestConfig, key types.NamespacedName) error {
+	current := &appsv1.Deployment{}
+	if err := tc.K8sClient.Get(tc.Context, key, current); err != nil {
+		return err
+	}
+
```
roytman, line 473: The current check doesn't verify *current.Spec.Replicas == current.Status.Replicas. During a scale-up (or just after the deployment controller starts reconciling), Status.Replicas might not yet reflect the desired count. For e2e stability, you might want:

```go
if current.Spec.Replicas != nil && *current.Spec.Replicas != current.Status.Replicas {
    return fmt.Errorf("status replicas (%d) has not converged to spec (%d)",
        current.Status.Replicas, *current.Spec.Replicas)
}
```
This is a nice-to-have for broader robustness.

roytman:
a nit comment.
/lgtm

PR #1027, opened 2026-05-07
feat: add x-llm-d-request-dropped-reason header to flow control responses

### LINKED ISSUE #1022: Differentiate flow control eviction reasons in HTTP response
## Summary

When flow control rejects or evicts a request, EPP sends an `ImmediateResponse(429)` with body `"request evicted by flow control"` ([server.go:486-495](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/epp/handlers/server.go#L486-L495)). This is the same 429 status and same body text for all rejection/eviction reasons, even though the underlying causes are distinct.

## Problem

Downstream consumers (e.g., batch-gateway) need to handle these cases differently:
- **Capacity rejection** → retry the request (transient backpressure)
- **TTL expiration** → mark as expired, do not retry (the request already waited and timed out)
- **Context cancellation** → caller-initiated, no retry needed
- **Post-dispatch eviction** → request was sent to a model server and then killed (active queue management, priority-based shedding, etc.)

Currently there is no way to distinguish these from the HTTP response alone — they all return `429` with identical body text.

## Proposal

Add an `x-llm-d-request-dropped-reason` response header to the `ImmediateResponse` with a structured reason value.

### Header

`x-llm-d-request-dropped-reason` — follows the `x-llm-d-` prefix convention for all llm-d component headers. Uses "dropped" rather than "error" because these are intentional flow control decisions, not errors.

### Reason naming convention

Reasons use a two-prefix scheme that identifies whether the request was dropped **before** or **after** being dispatched to an inference server:

| Prefix | Meaning |
|--------|---------|
| `rejected-*` | Request never reached an inference server (admission block or removed from queue before dispatch) |
| `evicted-*` | Request was dispatched to an inference server and then killed |

This distinction matters for retry decisions: `rejected-*` means no inference work was done (cheap to retry); `evicted-*` means GPU cycles were consumed (factor into backoff).

### Reason values

| Prefix | Reason value | When |
|--------|-------------|------|
| `rejected-` | `rejected-saturated` | System at capacity, request rejected before queueing |
| `rejected-` | `rejected-ttl-expired` | Request entered the queue but its TTL expired before dispatch |
| `rejected-` | `rejected-context-cancelled` | Client disconnected while request was queued |
| `evicted-` | `evicted` | Generic post-dispatch eviction (default when no specific sub-reason applies) |
| `evicted-` | `evicted-queue-pressure` | Evicted after dispatch due to backpressure from new arrivals |
| `evicted-` | `evicted-priority` | Evicted after dispatch, preempted by higher-priority request |

### Extensibility

The `evicted-*` namespace is reserved for future post-dispatch eviction reasons. The generic `evicted` reason serves as a fallback until specific callers of `EvictN()` exist and can specify why they're evicting. The specific sub-reasons (`evicted-queue-pressure`, `evicted-priority`) are forward-looking constants that document intent.

The two-prefix convention (`rejected-`, `evicted-`) makes it easy to add new reasons without ambiguity and gives consumers a simple grouping rule for retry logic.

## Implementation

Two response paths produce flow-control drop responses:

**Path 1: Flow control admission (pre-dispatch)**
`FlowControlAdmissionController.Admit()` → `EnqueueAndWait()` → `translateFlowControlOutcome()` → `errcommon.Error` with `Headers` map → `BuildErrResponse()` → `ImmediateResponse` with `HeaderMutation`

**Path 2: Post-dispatch eviction via evictCh**
`EvictN()` → sets reason on `EvictionRegistry` → `close(evictCh)` → server.go select → `RequestEvicted` → `updateStateAndSendIfNeeded()` → `ImmediateResponse` with `HeaderMutation`

### Type definition

```go
const RequestDroppedReasonHeaderKey = "x-llm-d-request-dropped-reason"

type RequestDroppedReason string

const (
    RequestDroppedReasonSaturated             RequestDroppedReason = "rejected-saturated"
    RequestDroppedReasonTTLExpired            RequestDroppedReason = "rejected-ttl-expired"
    RequestDroppedReasonContextCancelled      RequestDroppedReason = "rejected-context-cancelled"
    RequestDroppedReasonEvicted               RequestDroppedReason = "evicted"
    RequestDroppedReasonEvictedQueuePressure  RequestDroppedReason = "evicted-queue-pressure"
    RequestDroppedReasonEvictedPriority       RequestDroppedReason = "evicted-priority"
)
```

## Context

- Types are defined in [`pkg/common/error/error.go`](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/common/error/error.go)
- Implementation PR: #1027

lioraron:
## Summary

- Adds an `x-llm-d-request-dropped-reason` response header so downstream consumers (e.g. batch-gateway) can differentiate flow control rejection/eviction reasons and decide whether to retry.
- Covers both response paths: admission-time rejection via `BuildErrResponse` headers, and post-dispatch eviction via `EvictionRegistry` reason storage.
- Introduces a typed `RequestDroppedReason` string with two-prefix constants:
  - `rejected-saturated` — system at capacity, rejected before queueing
  - `rejected-ttl-expired` — TTL expired while queued before dispatch
  - `rejected-context-cancelled` — client disconnected while queued
  - `evicted` — generic post-dispatch eviction
  - `evicted-queue-pressure` — evicted after dispatch due to backpressure
  - `evicted-priority` — preempted by higher-priority request after dispatch
- The `rejected-*` / `evicted-*` distinction tells consumers whether inference work was done (retry cost).
- Adds a generic `Headers map[string]string` field to `errcommon.Error` so `BuildErrResponse` can attach arbitrary response headers without domain-specific fields.

Closes #1022

## Test plan

- [x] `TestBuildErrResponse` — verifies header presence/absence in ImmediateResponse
- [x] `TestFlowControlAdmissionController_Admit` — verifies each `QueueOutcome` maps to correct dropped reason header
- [x] `TestUpdateStateAndSendIfNeeded_Evicted` — verifies post-dispatch eviction path includes header
- [x] `TestEvictionRegistry_SetAndGetReason` — verifies typed reason storage and retrieval
- [x] `TestImmediateResponseEvictor_SetsReasonOnRegistry` — verifies evictor sets reason on registry
- [x] `TestRequestEvictor_EvictN_ClosesEvictChannel` — verifies EvictN propagates reason
- [x] `TestEvictionRegistry_Concurrency` — exercises SetReason/GetReason under concurrent load
- [x] `go build ./...` and `go test` pass for all affected packages


hexfusion:
cc @RishabhSaini

lioraron:
@elevran @nirrozenbaum @shmuelk Appreciate if you can review of this PR.

pkg/epp/flowcontrol/eviction/eviction_registry.go
```diff
@@ -30,34 +40,56 @@ import "sync"
 //
 // All methods are goroutine-safe.
 type EvictionRegistry struct {
-	mu       sync.RWMutex
-	channels map[string]chan struct{} // requestID → eviction channel
+	mu      sync.RWMutex
+	entries map[string]*evictionEntry // requestID → eviction entry
```
shmuelk: Use a sync.Map instead of a mutex and a map

pkg/epp/flowcontrol/eviction/eviction_registry.go
```diff
@@ -30,34 +40,56 @@ import "sync"
...
 }
 
 // NewEvictionRegistry creates a new EvictionRegistry.
 func NewEvictionRegistry() *EvictionRegistry {
 	return &EvictionRegistry{
-		channels: make(map[string]chan struct{}),
+		entries: make(map[string]*evictionEntry),
```
shmuelk: Ditto

pkg/epp/flowcontrol/eviction/eviction_registry.go
```diff
@@ -30,34 +40,56 @@ import "sync"
...
 	}
 }
 
 // Register stores an eviction channel for the given request ID.
 func (r *EvictionRegistry) Register(requestID string, ch chan struct{}) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	r.channels[requestID] = ch
+	r.entries[requestID] = &evictionEntry{ch: ch}
```
shmuelk: Dito

pkg/epp/flowcontrol/eviction/eviction_registry.go
```diff
@@ -30,34 +40,56 @@ import "sync"
...
 }
 
 // Get returns the eviction channel for the given request ID, or nil if not found.
 func (r *EvictionRegistry) Get(requestID string) chan struct{} {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
-	return r.channels[requestID]
```
shmuelk: Ditto

pkg/epp/flowcontrol/eviction/eviction_registry.go
```diff
@@ -30,34 +40,56 @@ import "sync"
...
+	if e := r.entries[requestID]; e != nil {
+		return e.ch
+	}
+	return nil
+}
+
+// SetReason records the eviction reason for a request before the channel is closed.
+func (r *EvictionRegistry) SetReason(requestID string, reason errcommon.RemovalReason) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
```
shmuelk: Ditto

pkg/epp/flowcontrol/eviction/eviction_registry.go
```diff
@@ -30,34 +40,56 @@ import "sync"
...
+	if e := r.entries[requestID]; e != nil {
+		e.reason = reason
+	}
+}
+
+// GetReason returns the eviction reason for a request, or empty string if not found.
+func (r *EvictionRegistry) GetReason(requestID string) errcommon.RemovalReason {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
```
shmuelk: Ditto

pkg/epp/flowcontrol/eviction/eviction_registry.go
```diff
@@ -30,34 +40,56 @@ import "sync"
...
+	return ""
 }
 
-// Deregister removes the eviction channel for the given request ID.
+// Deregister removes the eviction entry for the given request ID.
 func (r *EvictionRegistry) Deregister(requestID string) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	delete(r.channels, requestID)
+	delete(r.entries, requestID)
```
shmuelk: Ditto

shmuelk (dismissed):

pkg/epp/flowcontrol/eviction/eviction_registry.go
lioraron: Good idea, thanks.
Refactored `EvictionRegistry` to use `sync.Map` in dbb73ed7.

pkg/epp/flowcontrol/eviction/eviction_registry.go
lioraron: See https://github.com/llm-d/llm-d-inference-scheduler/pull/1027#discussion_r3219350954.

pkg/epp/flowcontrol/eviction/eviction_registry.go
lioraron: See https://github.com/llm-d/llm-d-inference-scheduler/pull/1027#discussion_r3219350954.

pkg/epp/flowcontrol/eviction/eviction_registry.go
lioraron: See https://github.com/llm-d/llm-d-inference-scheduler/pull/1027#discussion_r3219350954.

pkg/epp/flowcontrol/eviction/eviction_registry.go
lioraron: See https://github.com/llm-d/llm-d-inference-scheduler/pull/1027#discussion_r3219350954.

pkg/epp/flowcontrol/eviction/eviction_registry.go
lioraron: See https://github.com/llm-d/llm-d-inference-scheduler/pull/1027#discussion_r3219350954.

pkg/epp/flowcontrol/eviction/eviction_registry.go
lioraron: See https://github.com/llm-d/llm-d-inference-scheduler/pull/1027#discussion_r3219350954.

pkg/common/error/error.go
```diff
@@ -19,15 +19,38 @@ package error
...
 
+	configPb "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	extProcPb "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
 	envoyTypePb "github.com/envoyproxy/go-control-plane/envoy/type/v3"
 	"google.golang.org/grpc/status"
 )
 
+// RemovalReasonHeaderKey is the HTTP response header that communicates the specific
+// reason a request was rejected or evicted by flow control.
+const RemovalReasonHeaderKey = "x-removal-reason"
```
ahg-g: A couple of comments:

- We need to align on a common header prefix for all of llm-d components: `x-llm-d`
- The name `removal` is a bit unorthodox when describing request errors, it is either client or server error as per http codes, so perhaps `request-error` is more appropriate?

Putting them together, perhaps the header should be `x-llm-d-request-error-reason`?

pkg/common/error/error.go
```diff
@@ -19,15 +19,38 @@ package error
...
+
+// RemovalReason is the reason a request was rejected, removed from queue, or evicted after dispatch.
+type RemovalReason string
```
ahg-g: This will practically become an API, so worth some scrutiny. I think we need a concrete naming convention here, currently the reasons below are not fully aligned in how they are named.

pkg/common/error/error.go
```diff
@@ -19,15 +19,38 @@ package error
...
+const (
+	// Admission — rejected before entering a queue.
+	RemovalReasonAdmissionCapacity RemovalReason = "admission-capacity"
+
+	// Queue — entered the queue but removed before dispatch.
+	RemovalReasonQueueTTLExpired       RemovalReason = "queue-ttl-expired"
+	RemovalReasonQueueContextCancelled RemovalReason = "queue-context-cancelled"
+
+	// Evicted — dispatched to a model server and then killed.
+	RemovalReasonEvicted              RemovalReason = "evicted"
```
ahg-g: When will this catch all evicted reason be applied vs the detailed ones below?

pkg/common/error/error.go
```diff
@@ -19,15 +19,38 @@ package error
...
+// RemovalReasonHeaderKey is the HTTP response header that communicates the specific
+// reason a request was rejected or evicted by flow control.
+const RemovalReasonHeaderKey = "x-removal-reason"
+
+// RemovalReason is the reason a request was rejected, removed from queue, or evicted after dispatch.
+type RemovalReason string
+
+const (
+	// Admission — rejected before entering a queue.
+	RemovalReasonAdmissionCapacity RemovalReason = "admission-capacity"
```
ahg-g: `saturated` perhaps aligns better with all the past sentence form of the other reasons?

pkg/common/error/error.go
```diff
@@ -19,15 +19,38 @@ package error
...
+
+	// Queue — entered the queue but removed before dispatch.
+	RemovalReasonQueueTTLExpired       RemovalReason = "queue-ttl-expired"
+	RemovalReasonQueueContextCancelled RemovalReason = "queue-context-cancelled"
```
ahg-g: Perhaps using "dequeued" as the top level reason, so : `dequeued-tll-expired` and `dequeued-context-cancelled`

pkg/common/error/error.go
lioraron: Agreed on both points. Renamed the header to `x-llm-d-request-error-reason` and the type to `RequestErrorReason` in 67d62ee7.

pkg/common/error/error.go
lioraron: Adopted a two-prefix convention: `rejected-*` for pre-dispatch (request never reached an inference server) and `evicted-*` for post-dispatch (request was dispatched and then killed). This gives consumers a simple grouping rule for retry logic. See 67d62ee7.

pkg/common/error/error.go
lioraron: `evicted` is the current default set by `ImmediateResponseEvictor.Evict()`. The specific sub-reasons (`evicted-queue-pressure`, `evicted-priority`) are forward-looking — they'll be used when `EvictN()` callers exist and can specify why they're evicting. Added a comment explaining this.

pkg/common/error/error.go
lioraron: Done — renamed to `rejected-saturated`.

pkg/common/error/error.go
lioraron: Used `rejected-` instead of `dequeued-` as the prefix. "Dequeued" is ambiguous in this codebase since `PQDequeue` means "taken for processing." The `rejected-` prefix clearly communicates that the request never reached an inference server, which is the distinction downstream consumers care about (retry cost). Applied as `rejected-ttl-expired` and `rejected-context-cancelled`.

pkg/common/error/error.go
```diff
@@ -19,15 +19,38 @@ package error
...
 
+	configPb "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	extProcPb "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
 	envoyTypePb "github.com/envoyproxy/go-control-plane/envoy/type/v3"
 	"google.golang.org/grpc/status"
 )
 
+// RequestErrorReasonHeaderKey is the HTTP response header that communicates the specific
+// reason a request was rejected or evicted by flow control.
+const RequestErrorReasonHeaderKey = "x-llm-d-request-error-reason"
```
ahg-g: on a second thought, error is more about something went wrong, how about we call it `x-llm-d-request-dropped-reason`?

lioraron: Good call — "dropped" is more accurate since these aren't errors in the traditional sense; the system intentionally chose not to process them. Applied the rename: header is now `x-llm-d-request-dropped-reason`, type is `RequestDroppedReason`, and all constants/fields updated accordingly. See 912de9e5.

pkg/common/error/error.go
```diff
@@ -19,15 +19,38 @@ package error
...
+const RequestDroppedReasonHeaderKey = "x-llm-d-request-dropped-reason"
+
+// RequestDroppedReason is the reason a request was rejected before dispatch or evicted after dispatch.
+type RequestDroppedReason string
+
+const (
+	// Rejected — request never reached an inference server.
+	RequestDroppedReasonSaturated        RequestDroppedReason = "rejected-saturated"
+	RequestDroppedReasonTTLExpired       RequestDroppedReason = "rejected-ttl-expired"
+	RequestDroppedReasonContextCancelled RequestDroppedReason = "rejected-context-cancelled"
```
ahg-g, line 39: We should have a shedding reason as well, right? it is when the queues reach their limits.

ahg-g: oh, so the `QueueOutcomeRejectedCapacity`, which is the saturated reason above, is actually that.

ahg-g:
Please update https://github.com/llm-d/llm-d/blob/main/docs/api-reference/epp-http-headers.md with this new header.

/approve

lioraron:
@shmuelk can you review again pls?

PR #1037, opened 2026-05-10
fix: use int32 for InferenceObjective priority

### LINKED ISSUE #1036: [InferenceObjective] Use *int32 for the priority type
It is less related to this PR, but the type should be *int32
```suggestion
	Priority *int32`json:"priority,omitempty"`
```

_Originally posted by @roytman in https://github.com/llm-d/llm-d-inference-scheduler/pull/989#discussion_r3199623592_

 In Kubernetes API types, using int is discouraged because its size is platform-dependent (32-bit on some architectures, 64-bit on others). That can lead to inconsistent API schemas and client code generation. Using an      
  explicit width like int32 makes the CRD/OpenAPI schema stable (format: int32) and ensures the field behaves consistently across platforms and tooling.

zhouyou9505:
**What type of PR is this?**
/kind bug
/kind cleanup

This PR standardizes InferenceObjective.spec.priority to use an explicit-width integer type.
  - Updates the API field InferenceObjectiveSpec.Priority from *int to *int32. In Kubernetes APIs, int is platform-dependent (32/64-bit) and can lead to inconsistent schemas and generated clients; int32 is stable and matches 
    API conventions.
  - Regenerates dependent artifacts so everything stays in sync:
      - CRD schema (config/crd/bases/...inferenceobjectives.yaml) now reflects format: int32.
      - Apply-configuration client code (client-go/applyconfiguration/.../inferenceobjectivespec.go) now uses *int32 and WithPriority(int32).
      - Deepcopy output updated accordingly.
  - Fixes internal code/tests that assumed int for priority (e.g., director default priority handling and test wrappers), while keeping the scheduler/internal objective representation as int where needed via explicit
    conversion.

Fixes #1036 

NONE

### DIFF
--- a/apix/v1alpha2/inferenceobjective_types.go
+++ b/apix/v1alpha2/inferenceobjective_types.go
@@ -70,7 +70,7 @@ type InferenceObjectiveSpec struct {
 	// requests with Priority of 0 (the value used if Priority is unset or no InferenceObjective is specified).
 	// Similarly requests with a Priority of -10 will always be served after requests with Priority of 0.
 	// +optional
-	Priority *int `json:"priority,omitempty"`
+	Priority *int32 `json:"priority,omitempty"`
 
 	// PoolRef is a reference to the inference pool, the pool must exist in the same namespace.
 	//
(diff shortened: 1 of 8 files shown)


zhouyou9505:
I have fixed my DCO signature. Could you please help me confirm again @ahg-g

zhouyou9505:
I have fixed lint warning. Could you please help me confirm again @ahg-g

PR #1039, opened 2026-05-10
refactor(requestcontrol): remove ConsumerPlugin from DataProducer

### LINKED ISSUE #836: Consistently signal no data consumption from plugins
Different plugin can (and do) declare no dependency on data producers. 
This is not currently standardized
- not implement `Consumes()` at all (this is the clearest and simplest)
- `InFlightLoadProducer` and some test mocks use `return nil`  from `Consumes()` (see `inflightload/producer.go` and two occurrences in `pkg/epp/requestcontrol/plugin_executor_test.go`)
- `approximateprefix/plugin.go` uses `return map[string]any{}` (an empty map) which is semantically the same intent but with a different return value.

We should have a standard and documented way which all plugins implement.

yehuditkerido:
**What type of PR is this?**
/kind cleanup

Removes `ConsumerPlugin` from the `DataProducer` interface. DataProducers that don't consume data no longer need to implement an empty `Consumes()` method. Plugins that actually consume data can still implement `ConsumerPlugin` independently - the data dependency system detects consumers via runtime type assertion.

Fixes #836

**Release note**:
```release-note
NONE

### DIFF
--- a/pkg/epp/datalayer/data_graph_test.go
+++ b/pkg/epp/datalayer/data_graph_test.go
@@ -79,7 +79,6 @@ func (m *typedMockPlugin) TypedName() fwkplugin.TypedName {
 }
 
 func (m *typedMockPlugin) Produces() map[string]any { return m.produces }
-func (m *typedMockPlugin) Consumes() map[string]any { return nil }
 func (m *typedMockPlugin) PrepareRequestData(ctx context.Context, request *fwksched.InferenceRequest, endpoints []fwksched.Endpoint) error {
 	return nil
 }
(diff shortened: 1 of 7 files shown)


vMaroon:
While currently not in use, we do need this to manage dependencies between producers such as the `vllm-tokens-producer` that provides the tokens and MM features, and the `precise-prefix-cache-scorer`'s in-progress breakdown into at least one dependent producer: tokens -> kv-block hashes -> hits per endpoint -> scores.

elevran:
> While currently not in use, we do need this to manage dependencies between producers such as the `vllm-tokens-producer` that provides the tokens and MM features, and the `precise-prefix-cache-scorer`'s in-progress breakdown into at least one dependent producer: tokens -> kv-block hashes -> hits per endpoint -> scores.

@vMaroon please open an issue (if not already done) for ensuring the production DAG takes these relations into consideration when scheduling plugins. 
This PR does not change the current behavior and merely cleans up the code.

elevran:
> Error: pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go:340:1: File is not properly formatted (gofmt)

@yehuditkerido

PR #1043, opened 2026-05-10
update: pull GIE CRD from upstream instead of locally

zdtsw:
**What type of PR is this?**

/kind cleanup

- keep only the llm-d owned llm-d.ai CRDs in `config/crd/bases/`
-  add new `make sync-upstream-versions` target that reads GIE and Gateway API versions from `go.mod` and updates all references as input for  `make generate`
- update hack scripts to fetch `inferencepoolimports` CRD from upstream (was missing)

Fixes #

```release-note
NONE
```


elevran:
@zdtsw seems this needs a rebase to pick up the corrected import reference from llm-d-router

zdtsw:
> @zdtsw seems this needs a rebase to pick up the corrected import reference from llm-d-router

done rebase, thanks!

zdtsw:
@elevran mind if take another look?

config/crd/bases/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- inference.networking.k8s.io_inferencepools.yaml
+- inference.networking.x-k8s.io_inferencemodelrewrites.yaml
+- inference.networking.x-k8s.io_inferenceobjectives.yaml
+- inference.networking.x-k8s.io_inferencepoolimports.yaml
```
ahg-g: This repo shouldn't be hosting the definitions for those crds. We need to remove [inference.networking.x-k8s.io_inferencemodelrewrites.yaml](https://github.com/llm-d/llm-d-router/blob/main/config/crd/bases/inference.networking.x-k8s.io_inferencemodelrewrites.yaml) and [inference.networking.x-k8s.io_inferenceobjectives.yaml](https://github.com/llm-d/llm-d-router/blob/main/config/crd/bases/inference.networking.x-k8s.io_inferenceobjectives.yaml) and simply point to the latest released crds from the gaei repo, specifically https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/download/v1.5.0/v1-manifests.yaml

zdtsw: right, let me update

ahg-g:
One more thing, the path we should use to pull the GAIE manifests in https://github.com/llm-d/llm-d-router/blob/main/deploy/components/crds-gie/kustomization.yaml#L11 should be the one from the release artifact, not the branch, which is: `https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/download/v1.5.0/v1-manifests.yaml`

zdtsw:
> One more thing, the path we should use to pull the GAIE manifests in https://github.com/llm-d/llm-d-router/blob/main/deploy/components/crds-gie/kustomization.yaml#L11 should be the one from the release artifact, not the branch, which is: `https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/download/v1.5.0/v1-manifests.yaml`

thanks! updated

zdtsw:
Think this is finally passed e2e test, with updates in rbac for the new API group

PR #1046, opened 2026-05-11
fix: preserve assistant tool_calls through tokenizer render path

### LINKED ISSUE #1045: Assistant tool_calls are dropped before reaching the tokenizer
**What happened**:

Chat-completions requests with assistant `tool_calls` lose that field before being passed to the tokenizer. The rendered prompt diverges from vLLM, breaking KV-cache prefix hits for tool-calling workloads.

Assistant `tool_calls` should be parsed from the request and forwarded to the tokenizer's render path, so the rendered prompt matches vLLM and KV-cache prefix indexing stays consistent.

See the reproducer in https://github.com/llm-d/llm-d-kv-cache/issues/573.

Scheduler-side counterpart to:
- https://github.com/llm-d/llm-d-kv-cache/issues/573 (root cause)
- https://github.com/llm-d/llm-d-kv-cache/pull/577 (kv-cache side fix)

Per the discussion on llm-d/llm-d-kv-cache#577, the kv-cache fix alone is not enough — the scheduler also needs to parse `tool_calls` and forward it to the tokenizer.

**Environment**:
- Kubernetes version: n/a
- llm-d-scheduler version: main
- Install tools: n/a
- Others: n/a

yankay:
**What type of PR is this?**

/kind bug

Parse assistant `tool_calls` from chat-completions requests and forward them to the tokenizer's `RenderChatRequest`, then on to both rendering backends, so the rendered prompt used for KV-cache block indexing matches vLLM's rendering for tool-calling workloads.

Scheduler-side counterpart to the kv-cache fix in https://github.com/llm-d/llm-d-kv-cache/pull/577. Per the discussion there, the kv-cache change alone is not enough — the scheduler must also parse `tool_calls` and pass it through.

Paths covered:

- OpenAI request parser populates `Message.ToolCalls`.
- `ChatCompletionsToRenderChatRequest` writes it to `tokenizerTypes.Conversation.ToolCalls` (v0.8.1 of `llm-d-kv-cache` already exposes this field via https://github.com/llm-d/llm-d-kv-cache/pull/577).
- UDS path: `Conversation` is serialized as-is and consumed by the Python tokenizer service.
- vLLM HTTP path: `buildChatRenderRequest` now projects `ToolCalls` onto the outgoing `chatMessage`, so `/v1/chat/completions/render` receives the field.

Verified with the Qwen2.5 tokenizer that dropping `tool_calls` from the assistant message changes the rendered token stream (differs starting at token 32, 21-token delta), so the missing field would otherwise misalign KV-cache block hashes on tool-calling workloads.

Fixes #1045

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/interface/requesthandling/types.go
+++ b/pkg/epp/framework/interface/requesthandling/types.go
@@ -440,6 +440,8 @@ type Message struct {
 	Role string `json:"role,omitempty"`
 	// Content defines text of this message
 	Content Content `json:"content"`
+	// ToolCalls contains assistant tool calls for chat template rendering.
+	ToolCalls []any `json:"tool_calls,omitempty"`
 }
 
 type Content struct {
--- a/pkg/epp/framework/interface/requesthandling/types.go
+++ b/pkg/epp/framework/interface/requesthandling/types.go
@@ -555,6 +555,8 @@ type Message struct {
 	Role string `json:"role,omitempty"`
 	// Content defines text of this message
 	Content Content `json:"content"`
+	// ToolCalls contains assistant tool calls for chat template rendering.
+	ToolCalls []any `json:"tool_calls,omitempty"`
 }
 
 type Content struct {
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
@@ -24,6 +24,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
 	"sort"
 
 	"github.com/llm-d/llm-d-kv-cache/pkg/kvcache/kvblock"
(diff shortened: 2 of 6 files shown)


shashwatj07:
I tried this along with the other PR for the tokenizer. It still doesn't seem to be tokenizing the tool calls. @yankay @vMaroon 

Edit: Works. I was building it incorrectly.

elevran:
@yankay is this ready for review? https://github.com/llm-d/llm-d-kv-cache/pull/577 merged. Not sure if a release also been cut and updated here.

elevran:
@yankay still interested in working on this?
Need to rebase on main to avoid the vulnerability.

yankay:
Thanks @elevran — still on it. Just rebased onto main and added the missing piece on the vLLM HTTP render path: `buildChatRenderRequest` was dropping `ToolCalls` when projecting `Conversation` into `chatMessage`, so `/v1/chat/completions/render` never saw `tool_calls`. Fixed in the latest commit.

Verified end-to-end against vLLM v0.11.0 + Qwen2.5-0.5B-Instruct via `/tokenize` (same chat-template path as `/render`):

|  | tokens |
|---|---|
| with `tool_calls` (fixed) | 58 |
| without `tool_calls` (bug) | 38 |

20-token delta from the `<tool_call>...</tool_call>` block. Dropping WIP now.

PR #1048, opened 2026-05-11
ci: wip

sagearc:


.github/workflows/ci-lint.yaml
```diff
@@ -0,0 +1,80 @@
...
+        uses: actions/checkout@v6
+      - uses: dorny/paths-filter@v4
+        id: filter
+        with:
+          filters: |
+            src:
+              - '**/*.go'
+              - Dockerfile.*
+              - Makefile*
+              - go.mod
```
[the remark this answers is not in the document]
sagearc: fixed

.github/workflows/ci-lint.yaml
```diff
@@ -0,0 +1,80 @@
...
+    branches:
+      - main
+      - 'release-*'
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  contents: read
+  actions: read
```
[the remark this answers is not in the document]
sagearc: The original workflow used `actions: read`

.github/actions/e2e-build/action.yml
```diff
@@ -0,0 +1,51 @@
+name: Build e2e images
```
elevran: Is it possible to reuse the existing build image actions or a new action required for a specific reason?
If this provides a better alternative, consider enhancing the existing action instead of adding a new build action that needs to be maintained over time.

.github/workflows/ci-test.yaml
```diff
@@ -44,47 +44,19 @@ jobs:
...
     needs: check-changes
     if: ${{ needs.check-changes.outputs.src == 'true' }}
     runs-on: ubuntu-latest
-    timeout-minutes: 90
+    timeout-minutes: 60
     steps:
       - name: Checkout source
         uses: actions/checkout@v6
 
-      # Custom profile with directory-only removals (run in parallel, no apt).
```
elevran: Is the cleanup no longer needed or moved elsewhere?

.github/workflows/ci-lint.yaml
elevran, line 1: these changes require repo level settings change (e.g., changes in required tests)

.github/workflows/ci-test.yaml
```diff
@@ -221,10 +181,137 @@ jobs:
...
+        uses: ./.github/actions/e2e-build
+        with:
+          commit-sha: ${{ github.sha }}
+
+      - name: Run make test-e2e-gaie
         shell: bash
-        run: docker system prune -af
+        run: make test-e2e-gaie-run
+
+  e2e-scheduler:
```
elevran:
```suggestion
  e2e-router:
```

.github/workflows/ci-test.yaml
```diff
@@ -221,10 +181,137 @@ jobs:
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v6
+
+      # Custom profile with directory-only removals (run in parallel, no apt).
+      # The cleanup-profile: max is much more aggressive, it cleanups around 41GB,
+      # but runs 3.5 mins
+      - name: Free disk space
```
elevran: This 4-step preamble (`Free disk space` → `Create and resolve Go cache volumes` → `Cache Go modules and build cache` → `Build e2e images`) is copy-pasted verbatim into `e2e-scheduler` (lines 275–311). Since `.github/actions/e2e-build` was already extracted to eliminate duplication, the same pattern fits here — a `.github/actions/e2e-runner-setup` composite action would collapse each e2e job's boilerplate to two `uses:` lines:

```yaml
- name: Set up e2e runner
  uses: ./.github/actions/e2e-runner-setup
- name: Build e2e images
  uses: ./.github/actions/e2e-build
  with:
    commit-sha: ${{ github.sha }}
```

One requirement for composite actions: every `run` step must declare `shell:`, so `Create and resolve Go cache volumes` would need `shell: bash` added.

.github/workflows/ci-test.yaml
sagearc: Moved to each e2e job — since the suites now run in parallel as independent jobs, each one needs its own disk cleanup before spinning up the kind cluster.

elevran (commented):
## Timing (latest run 26031300926)

| Job | Duration |
|-----|----------|
| e2e-build (cache warm) | 5m 41s |
| e2e-gaie | 7m 01s |
| e2e-scheduler (disruption) | 12m 21s |
| e2e-scheduler (extended) | 14m 18s |
| e2e-scheduler (pd) | **23m 54s** |
| unit-and-integration | 12m 30s |

Total wall time: **~30 min** (critical path: `check-changes` → `e2e-build` 5m41s → `e2e-scheduler (pd)` 23m54s). Down from the stated 43m — the parallelisation is working.

`e2e-scheduler (pd)` at 24m is now the bottleneck. Based on `e2e_suite_test.go`, `BeforeSuite` creates a kind cluster and image-loads 5 images — that's probably 4–6 min of cluster setup. The remaining ~18–20 min is actual test execution.

## Could namespace-based parallelism cut it further?

The suite hardcodes host ports via NodePort services in `kindClusterConfig` (30080 / 32090). Those ports are cluster-global, not namespace-scoped, so running two shards in separate namespaces on the **same** kind cluster would conflict. To enable namespace parallelism you would need either:

1. Pre-map N unique host-port pairs in the kind cluster config and pass distinct `E2E_PORT`/`E2E_METRICS_PORT` per shard, or
2. Switch the shards to the existing port-forward code path (already used when `K8S_CONTEXT` is set), which is namespace-clean.

Both are follow-up work. Option 2 is likely smaller — the k8s-context path already handles namespace-scoped cleanup and avoids the NodePort constraint entirely.

.github/actions/e2e-build/action.yml
sagearc: Before this PR, e2e tests had no layer caching. While adding it I missed the release workflow already uses the same buildx GHA cache (layer resolution), so both action can be consolidated now. Thanks!

PR #1052, opened 2026-05-11
Upgraded Istio to 1.29.2

shmuelk:
**What type of PR is this?**
/kind bug

Due to a bug in the Istio control plane (istiod/pilot) one can not have two HttpRoute objects in the same namespace that route by path to two difference InferencePool objects.

This is Istio issue https://github.com/istio/istio/issues/58392. It is fixed in Istio 1.28.5 and in 1.29.1.

This PR upgrades the version of Istio used by the development script `make env-dev-kind` to Istio 1.29.2. That release was chosen as it is the release mentioned in the llm-d Guides.

The upgrade was performed by run `helm template .....` with a namespace of llm-d-istio-system in lieu of running `helm install ....` as documented in the Istio Installation guide. The generated text was broken up into separate files and edited to reflect several names that are changed in the development deployment.

Please note that this PR does not touch the End to End tests as they don't use Istio. Instead they use Envoy directly with a hard coded configuration.

```release-note
NONE
```

### DIFF
--- a/deploy/components/crds-istio/istio.yaml
+++ b/deploy/components/crds-istio/istio.yaml
@@ -7057,7 +7057,7 @@ spec:
                               type: integer
                               x-kubernetes-validations:
                               - message: port must be between 1-65535
-                                rule: 0 < self && self <= 6553
+                                rule: 0 < self && self <= 65535
                             route:
                               description: Match a specific route.
                               properties:
--- a/deploy/components/crds-istio/istio.yaml
+++ b/deploy/components/crds-istio/istio.yaml
@@ -4,11 +4,11 @@ metadata:
   annotations:
     helm.sh/resource-policy: keep
   labels:
-    app.kubernetes.io/instance: istio
+    app.kubernetes.io/instance: istio-base
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/part-of: istio
-    app.kubernetes.io/version: 1.28.0
-    helm.sh/chart: base-1.27.1
+    app.kubernetes.io/version: 1.29.2
+    helm.sh/chart: base-1.29.2
   name: authorizationpolicies.security.istio.io
 spec:
   group: security.istio.io
@@ -391,7 +391,7 @@ spec:
             x-kubernetes-preserve-unknown-fields: true
         type: object
     served: true
-    storage: false
+    storage: true
     subresources:
       status: {}
   - additionalPrinterColumns:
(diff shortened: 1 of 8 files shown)


deploy/components/crds-istio/istio.yaml
```diff
@@ -4,11 +4,11 @@ metadata:
   annotations:
     helm.sh/resource-policy: keep
   labels:
-    app.kubernetes.io/instance: istio
+    app.kubernetes.io/instance: istio-base
```
hexfusion, line 7: question: are these yaml auto generated or why for example is this changed for a version bump. more for my own understanding

shmuelk: I simply ran `helm template ....` and modified the various files in this repo.

The original files seemed to have been a dump of a running system (the order of elements is alphabetic and not in the order most people would use)

hexfusion: Makes sense, I think the process is an important takeaway for future efforts. Do we have it documented or could we add it as part of this PR? IE how to bump istio.

roytman:
Thank you @shmuelk .
Two small comments:

- `ValidatingWebhookConfiguration` objectSelector removed
The `revision-scoping` selector (istio.io/rev: llm-d-gateway) was removed. Is it intentional?
- storage: true swapped between CRD versions - safe for a fresh dev cluster (make env-dev-kind implies kind cluster), but would need a migration step on any reused cluster.

deploy/components/crds-istio/istio.yaml
```diff
@@ -6843,7 +7032,46 @@ spec:
...
+                                  type: object
+                              type: object
+                            portNumber:
+                              description: The service port to match on.
+                              maximum: 4294967295
+                              minimum: 0
+                              type: integer
+                              x-kubernetes-validations:
+                              - message: port must be between 1-65535
+                                rule: 0 < self && self <= 6553
```
roytman:
```suggestion
                                rule: 0 < self && self <= 65535
```

deploy/components/istio-control-plane/webhooks.yaml
```diff
@@ -3,75 +3,84 @@ kind: ValidatingWebhookConfiguration
...
+    # endpoint is ready.
+    failurePolicy: Ignore
+    sideEffects: None
+    admissionReviewVersions: ["v1"]
+    objectSelector:
+      matchExpressions:
+        - key: istio.io/rev
+          operator: In
+          values:
+          - "default"
```
roytman: Should it be 

```suggestion
          - "default"
          - "llm-d-gateway"
```

shmuelk:
@roytman Wrote:
> storage: true swapped between CRD versions - safe for a fresh dev cluster (make env-dev-kind implies kind cluster), but would need a migration step on any reused cluster.

The updates here are really only for kind. The Well lit paths in llm-d/llm-d do not reference the stuff here.

roytman:
thank you /lgtm

PR #1057, opened 2026-05-12
refactor(datalayer): wrap sync.Maps in typed managers

hexfusion:
**What type of PR is this?**

/kind bug
/kind cleanup

Configurations that register two sources with the same SourceType in different variants today silently first-match by variant iteration order (polling -> notification -> endpoint). This PR makes that case fail with ErrSourceTypeCollision at config time.

The detection is enabled by a small refactor: Runtime's four sync.Maps are wrapped in typed managers backed by a generic typedSyncMap[V]. Storage remains sync.Map the change is API shape. The commit body has the manager layout.

GVK uniqueness scope and all non-collision behavior are bit-for-bit identical, no external API changes.

Fixes # https://github.com/llm-d/llm-d-router/issues/1131
Related To: https://github.com/llm-d/llm-d-inference-scheduler/issues/1051

```release-note
NONE
```

Followup from # https://github.com/llm-d/llm-d-inference-scheduler/pull/883

### DIFF
--- a/pkg/epp/datalayer/collector.go
+++ b/pkg/epp/datalayer/collector.go
@@ -93,6 +93,9 @@ func (c *Collector) Start(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint,
 			return errors.New("cannot add nil data source")
 		}
 	}
+	if extractors == nil {
+		extractors = newExtractorMap()
+	}
 	if err := ctx.Err(); err != nil {
 		return err
 	}
--- a/pkg/epp/datalayer/collector.go
+++ b/pkg/epp/datalayer/collector.go
@@ -84,7 +84,7 @@ func NewCollector() *Collector {
 }
 
 // Start launches the collection goroutine.
-func (c *Collector) Start(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint, pollers []fwkdl.PollingDataSource, extractors map[string][]fwkdl.ExtractorBase) error {
+func (c *Collector) Start(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint, pollers []fwkdl.PollingDataSource, extractors *extractorMap) error {
 	if len(pollers) == 0 {
 		return errors.New("cannot start collector with empty sources")
 	}
@@ -98,14 +98,15 @@ func (c *Collector) Start(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint,
 	}
 
 	// Filter to poll-capable extractors up front so the hot loop avoids per-tick type assertions.
-	pollingExtractors := make(map[string][]fwkdl.Extractor, len(extractors))
-	for name, exts := range extractors {
+	pollingExtractors := make(map[string][]fwkdl.Extractor, extractors.Count())
+	extractors.Range(func(name string, exts []fwkdl.ExtractorBase) bool {
 		for _, ext := range exts {
 			if e, ok := ext.(fwkdl.Extractor); ok {
 				pollingExtractors[name] = append(pollingExtractors[name], e)
 			}
 		}
-	}
+		return true
+	})
 
 	c.mu.Lock()
 	defer c.mu.Unlock()
--- a/pkg/epp/datalayer/manager.go
+++ b/pkg/epp/datalayer/manager.go
@@ -1,0 +1,248 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package datalayer
+
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	fwkdl "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/datalayer"
+)
+
+// sourceHit identifies a matched source by its variant, registered name, and value.
+type sourceHit struct {
+	variant sourceVariant
+	name    string
+	src     fwkdl.DataSource
+}
+
+// variantSourceMap stores DataSources of one variant, keyed by TypedName.Name.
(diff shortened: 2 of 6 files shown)


hexfusion:
@elevran ptal :pray:

pkg/epp/datalayer/runtime.go
```diff
@@ -122,14 +125,21 @@ func (r *Runtime) Configure(cfg *Config, enableNewMetrics bool, disallowedExtrac
...
 
+	if err := r.validateNoCrossVariantCollisions(); err != nil {
+		return err
+	}
+
 	// Resolve code-registered pending registrations after processing user config.
 	for _, pending := range r.pendingRegistrations {
 		var gvkFilter *schema.GroupVersionKind
 		if ns, ok := pending.DefaultSource.(fwkdl.NotificationSource); ok {
 			gvk := ns.GVK()
```
elevran: this `gvk := ns.GVK()` shadows the outer `gvk := newGvk()` tracker declared at line 101. The call to `r.registerSource(pending.DefaultSource, gvk)` at L154 correctly picks up the outer (the inner scope with shadow has closed), but the shadow gvk could affect futre modifications of this function. Please rename one or the other.

pkg/epp/datalayer/manager.go
```diff
@@ -0,0 +1,248 @@
...
+func (e *extractorMap) Range(f func(name string, exts []fwkdl.ExtractorBase) bool) {
+	e.m.Range(func(k, raw any) bool {
+		return f(k.(string), raw.([]fwkdl.ExtractorBase))
+	})
+}
+
+// Append adds ext to srcName's extractor slice, deduping by extractor type so
+// code-registered extractors are a no-op when an equivalent type is already
+// wired via config.
+func (e *extractorMap) Append(srcName string, ext fwkdl.ExtractorBase) {
```
elevran, line 232: `Append` performs a `Get` and then a `Set` with no lock between them - is not safe if called concurrently for the same name (last writer wins). This is curerntly fine (Configure's pending-resolution loop is sequential), but `sync.Map` backing and the otherwise concurrent-looking API could mislead a future caller. Consider adding a constraint note ("Must not be called concurrently for the same srcName").

pkg/epp/datalayer/collector.go
```diff
@@ -84,7 +84,7 @@ func NewCollector() *Collector {
 }
 
 // Start launches the collection goroutine.
-func (c *Collector) Start(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint, pollers []fwkdl.PollingDataSource, extractors map[string][]fwkdl.ExtractorBase) error {
+func (c *Collector) Start(ctx context.Context, ticker Ticker, ep fwkdl.Endpoint, pollers []fwkdl.PollingDataSource, extractors *extractorMap) error {
```
elevran, line 87: nit: The old `map[string][]fwkdl.ExtractorBase` was nil-safe (for range etc) but `*extractorMap` is not and would panding on `nil`. All current callers pass `newExtractorMap()`, but a nil guard at the top of `Start` would match the previous safety.

pkg/epp/datalayer/manager_test.go
```diff
@@ -0,0 +1,41 @@
...
+	fwkdl "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/datalayer"
+	fwkplugin "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/plugin"
+	srcmocks "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/datalayer/source/mocks"
+)
+
+// TestVariantSourceMap_ConcurrentReadsRaceFree verifies variantSourceMap's
+// sync.Map backing permits concurrent read access from many goroutines
+// without data races. Run under -race to catch regressions if the storage
+// switches to a primitive that requires explicit locking on reads.
+func TestVariantSourceMap_ConcurrentReadsRaceFree(t *testing.T) {
```
elevran: nit: the commit message says this test exists so "a future refactor that drops the sync.Map...can't ship silent races." Only `variantSourceMap` is exercised here. Consider adding tests for `extractorMap` as well (parallel `Get`/`Range`/`Count` goroutines).

elevran (commented):
minor changes needed.
generally /lgtm

pkg/epp/datalayer/manager.go
hexfusion: yeah cleaned this up so there is no future concerns, good catch.

hexfusion: one by one I think sync.Map will go away but I will let that happen organically :)

hexfusion:
flakey test sent fix via https://github.com/llm-d/llm-d-router/pull/1143

hexfusion:
cc @elevran feedback addressed

hexfusion:
@ahg-g ptal

pkg/epp/datalayer/runtime.go
```diff
@@ -98,20 +109,12 @@ func (r *Runtime) Configure(cfg *Config, enableNewMetrics bool, disallowedExtrac
...
-				pollersCount++
-			case fwkdl.NotificationSource:
-				notifiersCount++
-			default:
-				endpointSourcesCount++
-			}
 
-			if len(srcCfg.Extractors) > 0 { // Store extractors mapped to source
-				r.sourceExtractors.Store(srcName, srcCfg.Extractors)
+			for _, ext := range srcCfg.Extractors {
```
elevran, line 119: The old code stored the entire user-supplied extractor slice as-is
After the change, user-config extractors are added through Append, which dedups by Type.
The extractorMap.Append returns silently on a Type collision. A config that lists two extractors of the same Type for the same source will silently drop the second one with no error or log. 
The original code only deduped when **code-registered** (i.e., pending/optional) extractors collided with config-wired ones. Intent is documented in the old comment "code registration is a no-op if already wired via config." Applying the same dedup to the user-config possibly changes user-visible config semantics.

Two acceptable resolutions (I'm fine either way: need to think if it makes sense to run the same extractor from user config or not...):
  - Restore the old "store as-is" semantics for user-config (keep Append-with-dedup only for the pending-resolution loop), or
  - Make duplicate-Type-per-source a hard config error (similar to ErrSourceTypeCollision) so silent loss is impossible.

Either way, please add a regression test (TestConfigure_DuplicateExtractorTypePerSource) that pins the chosen behavior.

elevran:
/lgtm
Please address the change in extractor dedup behavior (choose an option and add a test to guard against regressions) before approval.

elevran (commented):
/approve

elevran (approved):

PR #609, opened 2026-02-12
feat: update to use new kv-cache UDS tokenizer

zdtsw:
# Changes
- change `preprocessing` to `types` from kv-cache
- add new unit test case: same tests as the old ones
- keep old test case but mark it wont be used(for now can be removed later)
- add new make target` image-build-uds-tokenizer `to build UDS image
- update image to use the one from llm-d in deploy
- simplify Dockerfile, no more need python, install vllm wheels, preprocessing from old kv-cache-manager

# Ref
https://github.com/llm-d/llm-d-kv-cache/pull/303 
relate to https://github.com/llm-d/llm-d-kv-cache/pull/307 and https://github.com/llm-d/llm-d/pull/740

# Test
since the new testcase is a unit-test which only runs if we have a UDS service running locally
to verify this below steps were used
- `make image-build-uds-tokenizer `# to build UDS image from kv-cache repo by version stated in the go.mod,  in order to create ghcr.io/llm-d/llm-d-uds-tokenizer:dev locally
- `mkdir -p /tmp/tokenizer;  chmod 777 /tmp/tokenizer`  # prepare to run UDS image locally
- `podman run -d --name uds-tokenizer --user=0:0  -v /tmp/tokenizer:/tmp/tokenizer:z ghcr.io/llm-d/llm-d-uds-tokenizer:dev` # we use root for now , set 0:0
-` ls -la /tmp/tokenizer/tokenizer-uds.socket` # check socket is working
- update `const testModelName = "Qwen/Qwen2.5-0.5B-Instruct" `# we need a real model for the test, update precise_prefix_cache_uds_test.go, here can be any model, but Qwen was mentioned in UDS readme thus use this
-  `go test -v -run "TestPrefixCacheTracking_Score_UDS" ./pkg/plugins/scorer`
-```
 --- PASS: TestPrefixCacheTracking_Score_UDS (2.52s)
    --- PASS: TestPrefixCacheTracking_Score_UDS/nil_request (0.20s)
    --- PASS: TestPrefixCacheTracking_Score_UDS/empty_request_body (0.19s)
    --- PASS: TestPrefixCacheTracking_Score_UDS/longest_prefix_scorer_(default_scorer) (0.39s)
    --- PASS: TestPrefixCacheTracking_Score_UDS/chat_completions_request (0.40s)
    --- PASS: TestPrefixCacheTracking_Score_UDS/partial_prefix (0.38s)
[...]

PR #1060, opened 2026-05-12
test(preciseprefixcache): replace removed chat_completions deps with in-test fake

### LINKED ISSUE #1059: test(preciseprefixcache): drop dependency on removed kv-cache chat_completions package
The test file `pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache_test.go`
imports `github.com/llm-d/llm-d-kv-cache/pkg/preprocessing/chat_completions`,
which was removed in llm-d/llm-d-kv-cache#473 (merged 2026-04-18) as part
of the embedded-tokenizer deprecation tracked in llm-d/llm-d-kv-cache#295.

The scheduler's `go.mod` is currently pinned to kv-cache `v0.7.1`, which
still ships this package, so production code is unaffected today.
However, the next bump past the deletion point (`v0.8.0` and beyond,
including current `main`) will break the test build with:

```
module github.com/llm-d/llm-d-kv-cache@latest found
(v0.8.0, replaced by ...), but does not contain package
github.com/llm-d/llm-d-kv-cache/pkg/preprocessing/chat_completions
```

Discovered by the kv-cache upstream smoke job introduced in #1058.

### Fix

The production code path already runs over the UDS tokenizer; only the
test still reaches into the removed embedded path. Replace the in-test
use of `preprocessing.NewChatTemplatingProcessor` etc. with the UDS
tokenizer path (see the sibling `precise_prefix_cache_uds_test.go`),
or a small in-test fake.

### Refs

- llm-d/llm-d-kv-cache#473 — removed the package
- llm-d/llm-d-kv-cache#295 — RFC: remove embedded tokenizers
- #1058 — kv-cache upstream smoke job that surfaced this

yankay:
## What this PR does / why we need it

kv-cache main removed `pkg/preprocessing/chat_completions` and
`tokenization.NewCachedLocalTokenizer`, which broke `go mod tidy` against
kv-cache main in the upstream smoke job (see #1058).

Rewrite `precise_prefix_cache_test.go` to use an in-test fake
`KVCacheIndexer` instead of the embedded tokenizer build path:

- Extend `mockKVCacheIndexer` with a `computeBlockKeysFunc` hook and an
  injectable `kvBlockIndex`.
- Drop the `//go:build embedded_tokenizers` tag. The tag was effectively
  dead code: nothing in the repo (Makefile, CI workflows, scripts) ever
  passed `-tags=embedded_tokenizers` to `go build`/`go test`, so the
  tagged file was already skipped in every build. Removing it has no
  effect on what currently runs and unblocks bumping kv-cache past the
  package deletion. A grep confirms no other occurrences remain in the
  tree.
- Remove `TestPrefixCacheTracking_Score`; the `_UDS` variant added in
  #609 covers the same sub-cases end-to-end against the real UDS
  tokenizer.
- Keep speculative-path coverage (Produce/PreRequest/TTL eviction) via
  the fake.
- Use `require.Eventually` for the TTL eviction assertion because
  ttlcache v3 runs `OnEviction` asynchronously.

No production code changes.

## Which issue(s) this PR fixes

Resolves https://github.com/llm-d/llm-d-inference-scheduler/issues/1059

## Does this PR introduce a user-facing change?

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache_test.go
+++ b/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache_test.go
@@ -1,0 +1,295 @@
+package preciseprefixcache
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/jellydator/ttlcache/v3"
+	"github.com/llm-d/llm-d-kv-cache/pkg/kvcache"
+	"github.com/llm-d/llm-d-kv-cache/pkg/kvcache/kvblock"
+	"github.com/llm-d/llm-d-kv-cache/pkg/kvevents"
+	"github.com/llm-d/llm-d-kv-cache/pkg/tokenization/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+	fwkrh "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requesthandling"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
+	"github.com/llm-d/llm-d-router/test/utils"
+)
+
+// newSpeculativeFakeScorer builds a Scorer with an in-test fake kvCacheIndexer
+// (real in-memory kvblock.Index + fixed ComputeBlockKeys), avoiding the
+// embedded tokenizer path removed in llm-d/llm-d-kv-cache#473.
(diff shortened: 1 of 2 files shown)


shmuelk:
Is this PR needed? See PR #890

yankay:
Good point — #890 deletes this test file entirely as part of the vLLM HTTP tokenizer work + kv-cache v0.8.0 bump, which resolves #1059.

#1060 is a test-only stopgap to unblock the kv-cache upstream smoke job (#1058) in case #890 takes a while to land. Happy to close if #890 is merging soon — WDYT?

vMaroon:
We do need to recover effective tests, I suggest rebasing on the other PR which should merge soon, and pushing the tests. Thanks @yankay

yankay:
Closing as obsolete. PR #890 has landed and removed `precise_prefix_cache_test.go` entirely (together with the `pkg/preprocessing/chat_completions` and `NewCachedLocalTokenizer` usages this PR was meant to drop), so the original `go mod tidy` breakage against kv-cache main is already resolved upstream. The speculative-path unit coverage added here can be revisited as a separate, focused PR if useful. Thanks!

PR #1061, opened 2026-05-12
fix(disagg): use unweighted match-block count in P/D decider

### LINKED ISSUE #1047: [Feature] correct RAM-cached  usage in P/D disaggregation
OLD TITLE:  prefix-based-pd-decider misroutes RAM-cached requests to remote prefill due to weight/match-count semantic conflation

Problem Statement
⚠️ NOTE: this issue comes from real usage but context generated by Opus 4.7

## Summary
prefix-based-pd-decider decides whether to run remote prefill (P/D) by computing nonCachedTokens = inputTokens − matchBlocks × blockSize. The matchBlocks it reads is sourced from precise-prefix-cache-scorer, which assigns matchLen := int(score), where score is the device-tier-weighted sum produced by LongestPrefixScorer (GPU=1.0, CPU=0.8 by default).

The weight is meant for ranking endpoints during scoring, not for representing actual cached token count. Conflating these two semantics causes RAM-cached blocks to be undercounted by 20%, which inflates nonCachedTokens and pushes requests with large local-RAM hits onto the remote-prefill path — the opposite of what an LMCache / CPU-offload deployment wants.

**Why is this needed**:

## User story
As an operator running disaggregated llm-d with vLLM CPU offload (LMCache or built-in offloading), I expect that when most of my prompt is already cached in the decoder pod's RAM, the request stays as a plain decoder-only request — at most a host-to-device copy plus a short suffix prefill. Today, the more I cache in RAM, the more likely the EPP is to misroute my request to a remote prefiller, doubling latency.

## issue
Configuration: blockSize=16, nonCachedTokens=512, inputTokens=4096, contiguous prefix hit on 240 blocks (3840 tokens, ~94% of prompt), all hits on the selected D pod.

Hit location on D | score | matchLen | hitPrefixTokens | nonCachedTokens | Decision
-- | -- | -- | -- | -- | --
HBM (weight 1.0) | 240.0 | 240 | 3840 | 256 | skip prefill ✅
RAM (weight 0.8) | 192.0 | 192 | 3072 | 1024 | trigger P/D ❌

Real uncached suffix is 256 tokens in both cases; the RAM case is misclassified.

## Where it happens
pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go — assigns weighted score to matchLen:
```
matchLen := int(scores[addr])
ep.Put(attrprefix.PrefixCacheMatchInfoKey,
    attrprefix.NewPrefixCacheMatchInfo(matchLen, len(blockKeys), blockSize))
```

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go — uses matchBlocks as a literal cached-block count:
```
hitPrefixTokens := prefixCacheMatchInfo.MatchBlocks() * prefixCacheMatchInfo.BlockSizeTokens()
nonCachedTokens := inputTokens - hitPrefixTokens
```

LongestPrefixScorer in llm-d-kv-cache is correct in isolation — the misuse is on the inference-scheduler side.

## Why it gets worse with more cached blocks
Bias ≈ (1 − w) × m × blockSize tokens, i.e. every additional RAM-cached block adds ~3.2 fictitious uncached tokens to the decider's view. With more local cache, the misrouting rate goes up, not down.

### Proposed Solution

# Proposed fix
Decouple "ranking weight" from "match count":

Minimal fix: have precise-prefix-cache-scorer populate PrefixCacheMatchInfo.matchBlocks with the unweighted number of contiguous prefix hits on the selected pod, while still using the weighted score for endpoint ranking.
Better: extend PrefixCacheMatchInfo with per-tier hit counts (matchBlocksHBM, matchBlocksRAM) so deciders can apply tier-aware thresholds and cost models (HBM hit ≈ free, RAM hit ≈ H2D copy, remote P hit ≈ NIXL RDMA).
Optional: make nonCachedTokens configurable per tier, e.g.:

```
nonCachedTokens:
  afterLocalHBM: 0       # always skip P/D if HBM has it
  afterLocalAny: 512     # treat HBM+RAM identically vs threshold
```

yankay:
**What type of PR is this?**

/kind bug

The prefix-based P/D decider computes cached tokens as `MatchBlocks() * BlockSizeTokens()`. The precise-prefix-cache scorer fills `MatchBlocks()` with `int(weighted_score)` from `LongestPrefixScorer` (default weights: gpu=1.0, cpu=0.8), so RAM-cached prefixes get undercounted and large local-RAM hits are misrouted to remote prefill — the opposite of what an LMCache / CPU-offload deployment expects.

Example (blockSize=16, inputTokens=4096, 240/256 blocks hit, threshold=512):

| Hit tier | matchBlocks | hitTokens | nonCachedTokens | Decision |
| --- | --- | --- | --- | --- |
| HBM | 240 | 3840 | 256 | skip prefill |
| RAM | 192 | 3072 | 1024 | trigger P/D (wrong) |

Fix: add `MatchBlocksUnweighted()` on `PrefixCacheMatchInfo` (raw block hit count, populated by precise-scorer from `keyToPods`), and switch the decider to use it. Other consumers keep reading the weighted `MatchBlocks()` so ranking behaviour is unchanged. Legacy producers without a tier model are unaffected: `MatchBlocksUnweighted()` falls back to `MatchBlocks()` when unset.

Additionally guard `precise_prefix_cache.Score()` so it no longer overwrites the real attributes produced by `Produce()` with a `(matchLen=1, hitBlocks=1, blockSize=1)` placeholder — without this, layer (a) alone is not sufficient to fix the bug end-to-end.

Out of scope (follow-up):
- Renaming `MatchBlocks()` to make the weighted semantics explicit; touches every consumer.

Fixes https://github.com/llm-d/llm-d-inference-scheduler/issues/1047

**Release note**:

```release-note
Fix prefix-based P/D decider misrouting RAM-cached requests to remote prefill.
```

---

## Reproduction & Verification

Reproduced end-to-end on a real KIND deployment (llm-d gateway + EPP + vLLM simulator P/D pair, `precise-prefix-cache-scorer` with `speculativeIndexing` enabled), comparing two EPP images built from the same tree with only this fix toggled.

### Setup

- Two pods: `vllm-d` (10.244.0.11:8000), `vllm-p` (10.244.0.8:8000)
- EPP config: `precise-prefix-cache-scorer` weight=2, `nonCachedTokens=16` (= 1 block @ `BlockSizeTokens=16`)
- Injected a single **cpu-tier** KV-cache event via ZMQ for pod `10.244.0.11:8000` covering 2 full blocks (32 tokens)
- Sent one 43-token prompt; captured EPP logs by request ID

### Results

| Signal | buggy (before) | fixed (after) |
| --- | --- | --- |
| Decider caller line | `prefix_based_pd_decider.go:139` | `:140` + `:145` |
| `hit` tokens reported to decider | **1** | **32** (= 2 blocks × 16) |
| Profiles run | `decode` + `prefill` | **`decode` only** |
| Prefill skipped | 0 | 1 (`"Non-cached suffix is smaller than threshold, using decode profile only"`) |
| Prefill picker target | `10.244.0.8` (vllm-p) — wrong | n/a (skipped) |
| Decode picker target | `10.244.0.11` (vllm-d) | `10.244.0.11` (vllm-d) |

The buggy build, despite the scorer correctly identifying a 32-token cache hit on `vllm-d`, fed `hit=1` into the decider (weighted match-blocks collapsed by the cpu-tier weight), failed the skip-prefill threshold, and dispatched a needless prefill to `vllm-p`. The fixed build feeds the true unweighted block count (`hit=32`), passes the threshold, and correctly skips prefill — eliminating the spurious P/D disaggregation.

### Root cause confirmed in two layers

Both had to be fixed for the bug to disappear; fixing only (a) was not sufficient:

- (a) `prefix_based_pd_decider` was consuming the **weighted** match-block count. Switched to `MatchBlocksUnweighted()`.
- (b) `precise_prefix_cache.Score()` was unconditionally writing a `(matchLen=1, hitBlocks=1, blockSize=1)` placeholder, overwriting the real `Produce()` attributes. Guarded by `producePopulated` so real values survive.

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
@@ -24,39 +24,49 @@ const (
 	PrefixCacheMatchInfoKey = "PrefixCacheMatchInfoKey"
 )
 
+// PrefixCacheMatchInfo carries per-endpoint prefix-cache match information.
+//
+// matchBlocks may be a device-weighted score (precise-prefix-cache) or a raw
+// block count (approximate-prefix). Consumers translating blocks to tokens
+// (e.g. P/D decider) must use MatchBlocksUnweighted. See issue #1047.
 type PrefixCacheMatchInfo struct {
-	// matched prefix length in blocks
-	matchBlocks int
-	// total length in blocks
-	totalBlocks int
-	// block length in tokens
-	blockSizeTokens int
+	matchBlocks           int
+	matchBlocksUnweighted int
+	totalBlocks           int
+	blockSizeTokens       int
 }
 
-func NewPrefixCacheMatchInfo(matchBlocks int, totalBlocks int, blockSizeTokens int) *PrefixCacheMatchInfo {
+func NewPrefixCacheMatchInfo(matchBlocks, totalBlocks, blockSizeTokens int) *PrefixCacheMatchInfo {
 	return &PrefixCacheMatchInfo{
-		matchBlocks:     matchBlocks,
-		totalBlocks:     totalBlocks,
-		blockSizeTokens: blockSizeTokens,
+		matchBlocks:           matchBlocks,
+		matchBlocksUnweighted: matchBlocks, // tier-agnostic default
+		totalBlocks:           totalBlocks,
+		blockSizeTokens:       blockSizeTokens,
 	}
 }
 
-func (p *PrefixCacheMatchInfo) MatchBlocks() int {
-	return p.matchBlocks
+// WithMatchBlocksUnweighted sets the raw block-hit count, distinct from the
+// weighted ranking score in matchBlocks. Used by tier-aware producers.
+func (p *PrefixCacheMatchInfo) WithMatchBlocksUnweighted(n int) *PrefixCacheMatchInfo {
+	p.matchBlocksUnweighted = n
+	return p
 }
 
-func (p *PrefixCacheMatchInfo) TotalBlocks() int {
-	return p.totalBlocks
-}
+// MatchBlocks returns the producer-supplied score. For tier-aware producers
+// this is a device-weighted value, not a raw block count — use
+// MatchBlocksUnweighted when you need the physical hit count.
+func (p *PrefixCacheMatchInfo) MatchBlocks() int { return p.matchBlocks }
 
-func (p *PrefixCacheMatchInfo) BlockSizeTokens() int {
-	return p.blockSizeTokens
-}
+// MatchBlocksUnweighted returns the raw count of prefix blocks hit.
+func (p *PrefixCacheMatchInfo) MatchBlocksUnweighted() int { return p.matchBlocksUnweighted }
+
+func (p *PrefixCacheMatchInfo) TotalBlocks() int     { return p.totalBlocks }
+func (p *PrefixCacheMatchInfo) BlockSizeTokens() int { return p.blockSizeTokens }
 
 func (p *PrefixCacheMatchInfo) Clone() fwkdl.Cloneable {
-	return &PrefixCacheMatchInfo{
-		matchBlocks:     p.matchBlocks,
-		totalBlocks:     p.totalBlocks,
-		blockSizeTokens: p.blockSizeTokens,
-	}
+	// All fields are value types (int), so a shallow struct copy is a full
+	// deep copy. If a reference-type field (slice, map, pointer) is added
+	// later, this must be updated to copy it explicitly.
+	c := *p
+	return &c
 }
--- a/pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
@@ -24,39 +24,49 @@ import (
 
 var PrefixCacheMatchInfoDataKey = plugin.NewDataKey("PrefixCacheMatchInfoDataKey", approxprefixconstants.ApproxPrefixCachePluginType)
 
+// PrefixCacheMatchInfo carries per-endpoint prefix-cache match information.
+//
+// matchBlocks may be a device-weighted score (precise-prefix-cache) or a raw
+// block count (approximate-prefix). Consumers translating blocks to tokens
+// (e.g. P/D decider) must use MatchBlocksUnweighted. See issue #1047.
 type PrefixCacheMatchInfo struct {
-	// matched prefix length in blocks
-	matchBlocks int
-	// total length in blocks
-	totalBlocks int
-	// block length in tokens
-	blockSizeTokens int
+	matchBlocks           int
+	matchBlocksUnweighted int
+	totalBlocks           int
+	blockSizeTokens       int
 }
 
-func NewPrefixCacheMatchInfo(matchBlocks int, totalBlocks int, blockSizeTokens int) *PrefixCacheMatchInfo {
+func NewPrefixCacheMatchInfo(matchBlocks, totalBlocks, blockSizeTokens int) *PrefixCacheMatchInfo {
 	return &PrefixCacheMatchInfo{
-		matchBlocks:     matchBlocks,
-		totalBlocks:     totalBlocks,
-		blockSizeTokens: blockSizeTokens,
+		matchBlocks:           matchBlocks,
+		matchBlocksUnweighted: matchBlocks, // tier-agnostic default
+		totalBlocks:           totalBlocks,
+		blockSizeTokens:       blockSizeTokens,
 	}
 }
 
-func (p *PrefixCacheMatchInfo) MatchBlocks() int {
-	return p.matchBlocks
+// WithMatchBlocksUnweighted sets the raw block-hit count, distinct from the
+// weighted ranking score in matchBlocks. Used by tier-aware producers.
+func (p *PrefixCacheMatchInfo) WithMatchBlocksUnweighted(n int) *PrefixCacheMatchInfo {
+	p.matchBlocksUnweighted = n
+	return p
 }
 
-func (p *PrefixCacheMatchInfo) TotalBlocks() int {
-	return p.totalBlocks
-}
+// MatchBlocks returns the producer-supplied score. For tier-aware producers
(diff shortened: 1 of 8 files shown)


pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
```diff
@@ -24,39 +24,46 @@ const (
...
+func (p *PrefixCacheMatchInfo) TotalBlocks() int     { return p.totalBlocks }
+func (p *PrefixCacheMatchInfo) BlockSizeTokens() int { return p.blockSizeTokens }
 
 func (p *PrefixCacheMatchInfo) Clone() fwkdl.Cloneable {
-	return &PrefixCacheMatchInfo{
-		matchBlocks:     p.matchBlocks,
-		totalBlocks:     p.totalBlocks,
-		blockSizeTokens: p.blockSizeTokens,
-	}
+	c := *p
```
hexfusion, line 70: is it worth a comment that all  fields are value types which makes this shallow copy ok?

yankay: Good call, thanks! Added a note above the shallow copy in 35b08f29 documenting that all fields are value types today, with a reminder to update `Clone()` if a reference-type field is added later.

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -721,6 +728,48 @@ func extractPodSet(endpoints []scheduling.Endpoint) sets.Set[string] {
 	return podSet
 }
 
+// computeUnweightedMatchBlocks returns the longest consecutive prefix match
+// count per pod, ignoring device-tier weights. Mirrors the intersection
+// loop in kvcache.LongestPrefixScorer but adds 1 per block instead of a
+// weight. Pods absent from the returned map have 0 hits.
+func computeUnweightedMatchBlocks(
```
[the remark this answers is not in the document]
yankay: Fair point — `computeUnweightedMatchBlocks` is effectively a weight=1 specialization of `LongestPrefixScorer.Score`'s intersection, so the two will need to be kept aligned on edge cases. I'd prefer to keep the refactor out of this PR to keep its scope on the #1047 misrouting fix; happy to follow up with either a shared helper in this repo or extending the upstream `kvBlockScorer` API to return `(weighted, raw)` in a single pass.

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -386,7 +389,11 @@ func (s *Scorer) Produce(ctx context.Context,
 		}
 		addr := fmt.Sprintf("%s:%s", md.Address, md.Port)
 		matchLen := int(scores[addr])
-		ep.Put(attrprefix.PrefixCacheMatchInfoKey, attrprefix.NewPrefixCacheMatchInfo(matchLen, len(blockKeys), blockSize))
+		ep.Put(
+			attrprefix.PrefixCacheMatchInfoKey,
+			attrprefix.NewPrefixCacheMatchInfo(matchLen, len(blockKeys), blockSize).
+				WithMatchBlocksUnweighted(unweighted[addr]),
+		)
```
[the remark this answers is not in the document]
yankay: Good idea — added `TestProduce_AttachesUnweightedMatchBlocksToEndpoint` in b11be6e6 that asserts the unweighted value survives the builder chain onto the endpoint attribute, alongside the existing `TestUnweightedMatchBlocks_DecouplesFromTierWeight` covering the semantic side. The full tokenizer-backed `Produce()` path stays under the `embedded_tokenizers` build tag.

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -482,15 +497,22 @@ func (s *Scorer) Score(ctx context.Context, cycleState *scheduling.CycleState, r
 
 	// Write per-endpoint prefix-cache match info as endpoint attributes so downstream
 	// scorers (e.g. nohitlru) can determine whether any cache hits were found.
-	for _, endpoint := range endpoints {
```
liu-cong: @vMaroon Can you help understand why is this fallback path needed? I see Produce returns nil when speculativeEnabled is false, but I don't understand why. 

And we should not use "fake" `totalBlocks` and `blockSizeTokens` of (1,1) here, this can cause unintended consequences to downstream consumers,

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -386,7 +389,11 @@ func (s *Scorer) Produce(ctx context.Context,
 		}
 		addr := fmt.Sprintf("%s:%s", md.Address, md.Port)
 		matchLen := int(scores[addr])
-		ep.Put(attrprefix.PrefixCacheMatchInfoKey, attrprefix.NewPrefixCacheMatchInfo(matchLen, len(blockKeys), blockSize))
+		ep.Put(
```
liu-cong: I agree that the ideal fix for this is that the plugin should public per tier match tokens and let consumers decide how to use that info. But before we get there, the current implementation of counting a weighted token count seems reasonable to me. Loading from CPU is more costly, so undercounting is expected.

ahg-g:
@yankay  @vMaroon @liu-cong what is blocking this? can we pls move it forward? is this a must for the next release?

yankay:
> @yankay @vMaroon @liu-cong what is blocking this? can we pls move it forward? is this a must for the next release?

Hi @ahg-g, thanks for checking in.

There is some overlap with #1413, which addresses the same underlying issue on current main. Both approaches look fine to me.
I would appreciate another look when folks have bandwidth. Thanks!

yankay:
Closing this since #1413 has merged and covers the same #1047 fix on current main. Thanks everyone for moving it forward.

PR #1064, opened 2026-05-12
Fix/pretokenized prompt handling

### LINKED ISSUE #1062: [Bug] EPP silently ignores string-array and pre-tokenized prompt inputs in /v1/completions and /v1/embeddings requests
**Summary**                                                                                                                        
                                                                                                                                 
  Several scheduling and data-producer plugins in the Endpoint Picker handle only the simplest shape of the OpenAI prompt field —a single string ("prompt": "hello").
The two other shapes that are valid per the OpenAI spec are silently dropped:            
                                                            
  - Array of strings: "prompt": ["hello", "world"]                                                                               
  - Pre-tokenized input: "prompt": [1, 2, 3]                
                                                                                                                                 
  The same issue applies to /v1/embeddings where input accepts the same three shapes.                                            
                                                                                                                                 
  **Impact**                                                                                                                         
                                                            
  When a client sends either of the two unsupported shapes, the affected plugins see an empty or zero-length prompt, which leads to degraded (or silently wrong) scheduling decisions:
                                                                                                                                 
  - Prefix-cache scorers (approximate and precise) treat every such request as having no content, so prefix-cache hits are missed and all pre-tokenized requests collide on the same empty-input hash.
  - Tokenizer data producer tokenizes an empty string, discarding the token IDs the client already supplied.                     
  - Context-length-aware scorer estimates zero tokens, causing incorrect range-based endpoint filtering.                         
  - P/D (prefill/decode) decider sees length zero and routes sub-optimally.                                                      
  - Predicted-latency plugin computes an input token count of zero, throwing off its latency prediction and in-flight token counters.                                                                                                                      
  - Embeddings requests additionally fail to serialize at all in the approximate prefix hasher, so every embeddings request hashes to the same digest.                                                                                                     
                                                            
  **Expected behavior**                                                                                                              
                                                            
  All three prompt shapes should flow through the scheduling pipeline with equivalent accuracy:                                  
   
  - Pre-tokenized inputs should be used as-is, bypassing redundant tokenization and preserving the exact token count.            
  - String arrays should be treated as their concatenation for hashing, tokenization, and estimation.
  - Embeddings requests should participate in prefix-cache hashing and tokenization on equal footing with completions.           
                                                                                                                                 
  **Reproduction**                                                                                                                   
                                                                                                                                 
```
# Pre-tokenized: scheduler logs show tokenCount=0, prefix-cache scores are empty.                                              
  curl -X POST $GATEWAY/v1/completions \                                                                                         
    -H 'Content-Type: application/json' \                                                                                        
    -d '{"model":"<model>","prompt":[1,2,3,4,5,6,7,8]}'                 
```                                                         
                                                                                                                                 
```
# String array: same symptom.                                                                                                  
  curl -X POST $GATEWAY/v1/completions \                                                                                         
    -H 'Content-Type: application/json' \                                                                                        
    -d '{"model":"<model>","prompt":["hello, ","world"]}'                
```                                                        
                                                                                                                                 
  Inspect EPP logs — context-length, prefix-cache, and P/D decisions all behave as if the prompt were empty.                     
                                                                                                                                 
  **Scope of fix**                                                                                                                   
                                                            
  The request-body parser already decodes all three shapes correctly; the gap is entirely in downstream consumers. 
Affected plugins:                                     
                                                                                                                                 
  - tokenizer data producer                                                                                                      
  - approximateprefix hasher
  - preciseprefixcache scorer                                                                                                    
  - contextlengthaware scorer                                                                                                    
  - prefix_based_pd_decider profile handler                                                                                      
  - predictedlatency data producer                                                                                               
                                                                                                                                 
  A PR with the fix and unit-test coverage is in progress.

Mohamedma96:
**What type of PR is this?**
/kind bug
/kind test

Fixes #1062 

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
@@ -124,13 +124,30 @@ func getUserInputBytes(request *scheduling.InferenceRequest) ([]byte, error) {
 		return json.Marshal(request.Body.ChatCompletions.Messages)
 
 	case request.Body.Completions != nil:
+		if ids := request.Body.Completions.Prompt.TokenIDs; len(ids) > 0 {
+			return tokenIDsToBytes(ids), nil
+		}
 		return []byte(request.Body.Completions.Prompt.PlainText()), nil
 
 	case request.Body.Embeddings != nil:
-		// Handle embeddings API - marshal input for cache key generation
-		return json.Marshal(request.Body.Embeddings.Input)
+		input := request.Body.Embeddings.Input
+		if ids := input.TokenIDs; len(ids) > 0 {
+			return tokenIDsToBytes(ids), nil
+		}
+		return []byte(input.PlainText()), nil
 
 	default:
 		return nil, errors.New("invalid request body: no recognized API format found")
 	}
+}
+
+// tokenIDsToBytes encodes a uint32 token ID slice as little-endian bytes so
+// pre-tokenized prompts contribute real content to the prefix hash (rather
+// than collapsing to an empty byte slice).
+func tokenIDsToBytes(ids []uint32) []byte {
+	out := make([]byte, 4*len(ids))
+	for i, id := range ids {
+		binary.LittleEndian.PutUint32(out[i*4:(i+1)*4], id)
+	}
+	return out
 }
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
@@ -147,6 +147,9 @@ func getKVCacheBlocksFromRawPrompt(ctx context.Context, request *scheduling.Infe
 		return getKVCacheBlocksFromChatCompletions(ctx, request, blockSizeTokens, tokenEstimator), nil
 
 	case request.Body.Completions != nil:
+		if ids := request.Body.Completions.Prompt.TokenIDs; len(ids) > 0 {
+			return getKVCacheBlocksFromTokens(ids, blockSizeTokens), nil
+		}
 		rawBytes := []byte(request.Body.Completions.Prompt.PlainText())
 		return getKVCacheBlocksFromRawBytes(rawBytes, blockSizeTokens), nil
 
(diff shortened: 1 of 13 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -124,13 +124,30 @@ func getUserInputBytes(request *scheduling.InferenceRequest) ([]byte, error) {
...
 		return []byte(request.Body.Completions.Prompt.PlainText()), nil
 
 	case request.Body.Embeddings != nil:
-		// Handle embeddings API - marshal input for cache key generation
-		return json.Marshal(request.Body.Embeddings.Input)
+		input := request.Body.Embeddings.Input
+		if ids := input.TokenIDs; len(ids) > 0 {
+			return tokenIDsToBytes(ids), nil
+		}
+		return []byte(input.PlainText()), nil
```
hexfusion: nit: []byte(prompt.PlainText()) is a double allocation for string-array prompts (join then copy). a Prompt.Bytes() accessor  in hashing.go would let it go straight to bytes once.

roytman:
@Mohamedma96 , can you add Copilot as a reviewer

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
Mohamedma96: fixed, thanks

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -124,13 +124,30 @@ func getUserInputBytes(request *scheduling.InferenceRequest) ([]byte, error) {
 		return json.Marshal(request.Body.ChatCompletions.Messages)
```
vMaroon: This path should also consume tokens **+ mm_features** if present in a sensible manner.

ezrasilvera:
Great. Need it for RL.

ahg-g:
This is great, but similar to https://github.com/llm-d/llm-d-router/pull/1088 and https://github.com/llm-d/llm-d-router/pull/860, we are requiring each plugin to be aware of the all the possible ways of how a prompt or tokens will be available in the InferenceRequest type. We need to come up with a canonical and protocol agnostic way of passing down the prompts / tokens to the plugins and not have them aware of every single api we add (which some will likely miss to support).

/cc @zetxqx

vMaroon:
@ahg-g I think the `tokens-producer` can become the canonical transformer of user-input, as well as the single point that understands the concept of "estimations". Pls see https://github.com/llm-d/llm-d-router/issues/1303.

I've been discussing the latter with @liu-cong, but the first is also a solid possibility.

ahg-g:
I think https://github.com/llm-d/llm-d-router/pull/1445 will also address the issue this PR is trying to fix.

Mohamedma96:
> I think #1445 will also address the issue this PR is trying to fix.

I have spoken with @vMaroon and agreed that we'll merge this one in and rebase the new "bigger" PR on top of the changes here.

ahg-g:
> I think #1445 will also address the issue this PR is trying to fix.

this merged, closing this, please open again if you think there is still a gap

PR #1073, opened 2026-05-12
refactor(datalayer): typed Extractor[T] and source-as-dispatcher polling

### LINKED ISSUE #1053: [refactor] Typed, language-portable surface for data-layer extractors and dispatchers
## Goal

A typed data-layer plugin surface that is:

1. **Reusable within EPP** — one set of interfaces across every data-layer
   plugin in the inference scheduler, so authors write extractors once and
   bind them to polling, notification, or endpoint sources without
   re-implementing dispatch wiring per source type.
2. **Designed without Go-specific machinery** — the generic shape
   (`Extractor[T]`, `PollingDispatcher`, `PollInput[T]`) contains no
   `reflect.Type` tags or per-variant method names. A port to another
   language (e.g. Rust) could match the structure directly rather than
   designing a parallel runtime-type-tagging system.
3. **Small for plugin authors** — variant dispatch and type erasure live
   behind the interface; an extractor implements two methods
   (`TypedName` + `Extract`) and never touches `reflect`, `any`, or
   per-variant method names.

## Why this matters

The legacy `ExtractorBase` surface couples plugin authors to Go-specific
reflection (`OutputType()`, `ExpectedInputType()`, per-tick
`reflect.TypeOf().AssignableTo()`) and a 3-flavor method hierarchy
(`Extract` / `ExtractNotification` / `ExtractEndpoint`). Two costs:

- **Errors that should be compile-time become runtime errors** at first
  request, with no help from the type system when adding a new source or
  extractor variant.
- **Reflection on the dispatch hot path** every tick, regardless of how
  rarely the types could ever drift apart in practice.

The shape also doesn't survive a port to another language: languages
without `reflect.Type` (e.g. Rust) have no convention for runtime type
tagging on trait objects, so a port would need a parallel `Any`-based shim
with a different shape from the Go side.

The typed surface removes the reflection, lets the type system catch
misconfiguration at compile time for Go-wired extractors and at config
load for YAML-wired ones, and keeps the shape free of Go-specific
machinery a future port would have to design around.

## The surface this issue lands

- **`Extractor[T any]`** — single generic interface with
  `Extract(ctx, T) error`. Variant identity (polling vs notification vs
  endpoint) is carried by `T`, not by method name.
- **`PollingDispatcher`** — polling sources own their bound extractors and
  drive them themselves (`Dispatch(ctx, ep) error`). The runtime tick is
  `dispatcher.Dispatch(ctx, ep)` — no central `extractorMap` walk, no
  per-tick `any → T` cast.
- **`PollInput[T any]`** — typed poll payload (`{Payload T, Endpoint}`)
  handed to polling extractors. Replaces the `(data any, ep Endpoint)`
  pair.
- **Type erasure at one point** —
  `PollingDispatcher.AppendExtractor(plugin.Plugin)` asserts
  `Extractor[PollInput[T]]` once at config time. The dispatch hot path
  is fully typed.

## Portability sketch (Rust as a concrete example)

| Concept | Go | Rust |
|---|---|---|
| `Extractor[T any]` | generic interface | `trait Extractor<T>` |
| `PollInput[T]` | generic struct | `struct PollInput<T> { payload: T, endpoint: ... }` |
| `PollingDispatcher` | interface (`Dispatch` + `AppendExtractor`) | trait, same two methods |
| Type erasure boundary | `plugin.Plugin` + assertion | `Box<dyn Plugin>` + downcast |

No Go-isms inside the interface definitions — erasure only at the
registry/config boundary where dynamic loading needs it regardless of
language.

## Byproducts

- Reflect machinery (`OutputType`, `ExpectedInputType`,
  `validateInputTypeCompatible`, six reflect type vars) removed as dead
  code.
- Cross-variant `SourceType` collisions surface as configuration errors
  (`ErrSourceTypeCollision`).
- Per-step timeouts and per-extractor error metrics preserved against
  prior behavior.

## Scope and follow-ups

- Notification and endpoint variants keep the legacy `extractorMap` for
  now; bringing them into the dispatcher pattern is a follow-up so the
  asymmetry doesn't ossify.
- Hot-reload is not in scope. Configure runs once at startup today; a
  reload spike can design against the new typed surface when needed.
- Plugin-author docs (`docs/create_new_data_source.md`) are a separate
  PR.

hexfusion:
[Design doc](https://docs.google.com/document/d/1Ch78yADpIly3LTanH9Gv7BludwtqQrYrNQhvCQmh60w/edit?usp=sharing)

**What type of PR is this?**
/kind cleanup

Replaces the legacy `ExtractorBase` hierarchy with a generic `Extractor[T any]` interface and makes polling sources own their bound extractors via a `PollingDispatcher` contract. The runtime tick simplifies to `dispatcher.Dispatch(ctx, ep)`; per-call `any -> T` casts and reflect-based input-type checks are gone.

**Plugin before:**
```go
func (ext *Extractor) Extract(ctx context.Context, data any, ep fwkdl.Endpoint) error {
  metrics, ok := data.(sourcemetrics.PrometheusMetricMap)
  if !ok {
    return fmt.Errorf("unexpected input in Extract: %T", data)
  }
  // ... business logic
}
```

**Plugin after:**
```go
func (ext *Extractor) Extract(ctx context.Context, in fwkdl.PollInput[sourcemetrics.PrometheusMetricMap]) error {
  metrics := in.Payload
  ep := in.Endpoint
  // ... business logic
}
```

**Behavior changes** (intentional; not behavior-preserving):

- **Pure-append at binding layer; dedup enforced at config-validation layer.** `HTTPDataSource.AppendExtractor` and `extractorMap.Append` are pure appends. Duplicate-Type detection lives in `runtime.Configure` (via `validateSourceExtractors` for user-config slices, plus an explicit `boundTypes` map that lets code-registered pending extractors yield to user-config of the same Type). Single enforcement layer, no per-binding-call defense.
- **Per-extractor panic recovery in `HTTPDataSource.Dispatch`.** A buggy extractor (e.g. nil-deref) no longer kills the EPP collector goroutine. Panics increment the same `*ExtractErrorsTotal` counters as regular failures and log with stack trace.
- **Multi-variant source rejection.** `registerSource` errors when a source implements more than one variant interface (`PollingDispatcher` + `NotificationSource`, etc.); previously the first match silently won.
- **Typed parser contract.** `HTTPDataSource[T]`'s `parser` field doc states "MUST NOT return (zero, nil) for nilable T; the dispatcher does not validate." Parsers signal failure with a non-nil error; the typed signature makes `(nil, nil)` an explicit programming bug rather than an accidental untyped fall-through.

**New plugin-author-visible surface**:

- `fwkdl.Extractor[T any]` interface (replaces `ExtractorBase` + reflect-based dispatch)
- `fwkdl.PollInput[T]`, `fwkdl.PollingExtractor[T]`, `fwkdl.PollingDispatcher`
- `http.ErrExtractorTypeMismatch` sentinel
- `httptest.ParserContract[T]` test helper for parser implementations to verify their non-nil-on-success contract against golden inputs
- `LlmdDataLayer{Poll,Extract}ErrorsTotal` increments added at the dispatch path for parity with the upstream metric-deprecation cycle

Fixes https://github.com/llm-d/llm-d-router/issues/1053

**Release note**:
```release-note
The data layer's extractor API has been refactored to use Go generics. Out-of-tree plugin authors writing custom extractors must implement `Extract(ctx, fwkdl.PollInput[T]) error` (replacing the untyped `Extract(ctx, data any, ep Endpoint) error`); parsers passed to `NewHTTPDataSource[T]` must not return `(nil, nil)` for nilable T. A new `httptest.ParserContract[T]` helper validates parser implementations against this contract.
```

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/extractor/models/extractor.go
+++ b/pkg/epp/framework/plugins/datalayer/extractor/models/extractor.go
@@ -3,8 +3,6 @@ package models
 import (
 	"context"
 	"encoding/json"
-	"fmt"
-	"reflect"
 
 	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
 	fwkplugin "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
@@ -15,18 +13,13 @@ import (
 // asynchronously and not tied to incoming requests
 var _ fwkplugin.ProducerPlugin = &ModelExtractor{}
 
-var _ fwkdl.Extractor = &ModelExtractor{}
+var _ fwkdl.PollingExtractor[*ModelResponse] = &ModelExtractor{}
 
-// ModelResponse is the response from /v1/models API
+// ModelResponse is the response from /v1/models API.
 type ModelResponse struct {
 	Object string                 `json:"object"`
 	Data   []attrmodels.ModelData `json:"data"`
 }
-
-// ModelsResponseType is the type of models response
-var (
-	ModelsResponseType = reflect.TypeOf(ModelResponse{})
-)
 
 // ModelExtractor implements the models extraction.
 type ModelExtractor struct {
@@ -50,11 +43,6 @@ func (me *ModelExtractor) TypedName() fwkplugin.TypedName {
 	return me.typedName
 }
 
-// ExpectedInputType defines the type expected by ModelExtractor.
-func (me *ModelExtractor) ExpectedInputType() reflect.Type {
-	return ModelsResponseType
-}
-
 // ModelServerExtractorFactory is a factory function used to instantiate data layer's
 // models extractor plugins specified in a configuration.
 func ModelServerExtractorFactory(name string, _ *json.Decoder, _ fwkplugin.Handle) (fwkplugin.Plugin, error) {
@@ -63,21 +51,13 @@ func ModelServerExtractorFactory(name string, _ *json.Decoder, _ fwkplugin.Handl
 	return extractor, nil
 }
 
-// Extract transforms the data source output into a concrete attribute that
-// is stored on the given endpoint.
-func (me *ModelExtractor) Extract(_ context.Context, data any, ep fwkdl.Endpoint) error {
-	models, ok := data.(*ModelResponse)
-	if !ok {
-		return fmt.Errorf("unexpected input in Extract: %T", data)
-	}
-
-	ep.GetAttributes().Put(me.dk.String(), attrmodels.ModelDataCollection(models.Data))
+// Extract stores the model list as an endpoint attribute.
+func (me *ModelExtractor) Extract(_ context.Context, in fwkdl.PollInput[*ModelResponse]) error {
+	in.Endpoint.GetAttributes().Put(me.dk.String(), attrmodels.ModelDataCollection(in.Payload.Data))
 	return nil
 }
 
 // Produces returns data produced by the producer.
-// This is a map from data key (string) produced to
-// the data type of the key (represented as data with default value casted as any field).
 func (me *ModelExtractor) Produces() map[fwkplugin.DataKey]any {
 	return map[fwkplugin.DataKey]any{me.dk: attrmodels.ModelDataCollection{}}
 }
--- a/pkg/epp/framework/plugins/datalayer/source/http/datasource.go
+++ b/pkg/epp/framework/plugins/datalayer/source/http/datasource.go
@@ -19,83 +19,153 @@ package http
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io"
 	"net/url"
 	"reflect"
+	"runtime/debug"
+	"slices"
+	"sync"
+	"time"
 
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/llm-d/llm-d-router/pkg/common/observability/logging"
 	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
 	fwkplugin "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+	"github.com/llm-d/llm-d-router/pkg/metrics"
 )
 
-// HTTPDataSource is a data source that receives its data using HTTP client.
-type HTTPDataSource struct {
-	typedName fwkplugin.TypedName
-	scheme    string // scheme to use
-	path      string // path to use
+var ErrExtractorTypeMismatch = errors.New("extractor type mismatch")
 
-	client     Client // client (e.g. a wrapped http.Client) used to get data
-	parser     func(io.Reader) (any, error)
-	outputType reflect.Type
+// defaultStepTimeout bounds each Poll and each Extract independently so a slow
+// extractor cannot starve sibling extractors of their tick budget.
+const defaultStepTimeout = time.Second
+
+// HTTPDataSource is a typed polling dispatcher. T is the data type the source
+// produces; bound extractors must implement Extractor[PollInput[T]].
+type HTTPDataSource[T any] struct {
+	typedName fwkplugin.TypedName
+	scheme    string
+	path      string
+
+	client Client
+	// parser converts the response body to T. MUST NOT return (zero, nil) for nilable T;
+	// the dispatcher does not validate.
+	parser func(io.Reader) (T, error)
+
+	mu   sync.RWMutex
+	exts []fwkdl.PollingExtractor[T]
 }
 
-// NewHTTPDataSource returns a new data source, configured with
-// the provided scheme, path and certificate verification parameters.
-func NewHTTPDataSource(scheme string, path string, skipCertVerification bool, pluginType string,
-	pluginName string, parser func(io.Reader) (any, error), outputType reflect.Type) (*HTTPDataSource, error) {
+// NewHTTPDataSource constructs a typed polling dispatcher.
+func NewHTTPDataSource[T any](scheme, path string, skipCertVerification bool,
+	pluginType, pluginName string, parser func(io.Reader) (T, error)) (*HTTPDataSource[T], error) {
 	if scheme != "http" && scheme != "https" {
 		return nil, fmt.Errorf("unsupported scheme: %s", scheme)
 	}
 	if scheme == "https" {
 		httpsTransport := baseTransport.Clone()
-		httpsTransport.TLSClientConfig = &tls.Config{
-			InsecureSkipVerify: skipCertVerification,
-		}
+		httpsTransport.TLSClientConfig = &tls.Config{InsecureSkipVerify: skipCertVerification}
 		defaultClient.Transport = httpsTransport
 	}
+	return &HTTPDataSource[T]{
+		typedName: fwkplugin.TypedName{Type: pluginType, Name: pluginName},
+		scheme:    scheme,
+		path:      path,
+		client:    defaultClient,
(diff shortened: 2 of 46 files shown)


hexfusion:
cc @ahg-g would love your thoughts on direction

hexfusion:
@elevran this is rebased and cleaned up

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -23,79 +23,151 @@ import (
...
+	pollCtx, cancelPoll := context.WithTimeout(ctx, defaultStepTimeout)
+	data, err := s.Poll(pollCtx, ep)
+	cancelPoll()
+	if err != nil {
+		return err
+	}
+	// Behavior preservation: the legacy collector short-circuited the extractor
+	// loop when Poll returned nil data. Preserve that for nilable T (pointer,
+	// map, slice, chan, func, interface). Value T's (int, struct) skip the check.
+	if isNilData(data) {
```
elevran: suggestion: cache nilability at construction

`isNilData` keeps a `reflect.ValueOf(v)` call on every Dispatch tick — even for value `T` (where the kind switch falls through and returns false). Whether `T` is nilable is fixed at type-instantiation time, so the reflect call is recomputing a constant.

A single bool cached at `NewHTTPDataSource` time would make the hot path reflect-free.

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -23,79 +23,151 @@ import (
...
+	if !ok {
+		return fmt.Errorf("extractor %s: expected %s, got %T",
+			ext.TypedName(), reflect.TypeFor[fwkdl.PollingExtractor[T]](), ext)
+	}
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	extType := typed.TypedName().Type
+	for _, existing := range s.exts {
+		if existing.TypedName().Type == extType {
+			return nil
```
elevran: Silent dedup hides config bugs; inconsistent with `pollingDispatchers.Register`.

`pollingDispatchers.Register` (manager.go:150) errors on a duplicate source **Name**, but `AppendExtractor` here silently drops a duplicate extractor **Type** (different Names allowed). A user who configures two extractors of Type `"metrics"` under different Names against the same source will get only the first one bound, with no log line and no error.
Either return an error here, or at minimum log a warning so the dropped binding is visible. (The contract comment on `PollingDispatcher` says "AppendExtractor dedups by TypedName().Type", even keeping that contract, surfacing the drop seems worth the small cost.)

The existing test `TestAppendExtractor_DuplicateType_SilentlyDropped` pins the current behavior; flipping to an error would need that test updated.

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -23,79 +23,151 @@ import (
...
+	})
+	if err != nil {
+		var zero T
+		return zero, err
+	}
+	typed, ok := raw.(T)
+	if !ok {
+		var zero T
+		return zero, fmt.Errorf("HTTPDataSource %s: parser returned %T, expected %T", s.typedName, raw, zero)
+	}
```
elevran, line 99: The wrapper closure on line 81 has signature `func(io.Reader) (any, error)` and returns `s.parser(r)` (typed `(T, error)`), so `raw` is always a non-nil `any` boxing a `T`. `Client.Get` doesn't transform `raw`; it just propagates the parser's return. So `raw.(T)` cannot fail with `ok == false`.

If you want to keep this as a pure defensive guard, a one-line comment to that effect would help. Otherwise dropping the branch (and just `return raw.(T), nil`) simplifies. The cleaner long-term fix would be a typed `Client[T]` so the wrapper closure goes away entirely , but that's a bigger change not in scope of this PR.

pkg/epp/framework/plugins/datalayer/extractor/models/extractor.go
```diff
@@ -51,17 +50,15 @@ func (m ModelDataCollection) String() string {
...
-// ModelResponse is the response from /v1/models API
+// ModelResponse is the response from /v1/models API.
+// Data is typed as ModelDataCollection (not []ModelData) so the Cloneable
+// methods are attached directly, removing the per-call ModelDataCollection
+// conversion the extractor would otherwise need.
 type ModelResponse struct {
-	Object string      `json:"object"`
-	Data   []ModelData `json:"data"`
+	Object string              `json:"object"`
+	Data   ModelDataCollection `json:"data"`
```
elevran: the attribute is being moved. FYI

pkg/epp/framework/plugins/datalayer/extractor/models/extractor.go
```diff
@@ -95,14 +87,9 @@ func ModelServerExtractorFactory(name string, _ json.RawMessage, _ fwkplugin.Han
...
-// Extract transforms the data source output into a concrete attribute that
-// is stored on the given endpoint.
-func (me *ModelExtractor) Extract(_ context.Context, data any, ep fwkdl.Endpoint) error {
-	models, ok := data.(*ModelResponse)
-	if !ok {
-		return fmt.Errorf("unexpected input in Extract: %T", data)
-	}
-
-	ep.GetAttributes().Put(ModelsAttributeKey, ModelDataCollection(models.Data))
+// Extract requires non-nil Payload + Endpoint per the PollingDispatcher contract.
```
elevran: Question:
the doc comment says "Direct callers bypassing the dispatcher will panic on nil Payload" and the previous explicit `data.(*ModelResponse)` nil check (and the `nil data` test case) were removed. The polling dispatcher's `isNilData` short-circuit covers the production path, but is the new "panic on misuse" contract intentional? An explicit early `return nil` (or a panic with a clear message) would surface misuse more loudly than a bare nil-pointer deref on `in.Payload.Data`. Not blocking, happy with either answer.

elevran (commented):
Thanks @hexfusion!
Following the refactor the per-tick `any => T` cast and reflect-based input-type checks are gone, the variant-collision and multi-variant-source guards are well covered by new tests, and the `notification source requires NotificationExtractor` / `endpoint source requires EndpointExtractor` rules now fail at Configure rather than panicking at the first event. Nice work!
Please resolve the conflict.
/lgtm

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
hexfusion: honestly we should ensure proper returns. returning nil, nil breaks golangs error handling patterns and imo should be avoided. i would much rather see us enforce nil, nil as an anti-pattern.

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
hexfusion: adding comment and will properly resolve in followup

pkg/epp/framework/plugins/datalayer/extractor/models/extractor.go
hexfusion: intentional. dropped the panic warning from Extract since that was the source of "feels strange". deleted isNilData since audited parsers don't return (nil, nil) in practice. contract now lives on PollInput + the parser field godoc, Extract trusts it. defer recover() in Dispatch catches panics, increments extract-errors metrics, and logs with stack trace violations surface loudly without killing the goroutine.

will continue to harden these contracts down to Parsers in future work.

ahg-g:
This is great, but huge change, @elevran we should merge it soon to give it enough soak time before the freeze!

PR #1079, opened 2026-05-13
deprecate UDS-backend in `token-producer` and suggest configuring the vllm render instead

vMaroon:
Christmas came early

```release-note
deprecated UDS-backend in `token-producer`
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/README.md
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/README.md
@@ -24,6 +24,12 @@ The plugin calls vLLM's `/v1/completions/render` and
 to `vllm` with `http://localhost:8000`. Future protocol fields (e.g. `grpc`)
 can be added alongside `http` under the same `vllm` block.
 
+> [!WARNING]
+> The `udsTokenizerConfig` backend (gRPC-over-UDS sidecar) is **deprecated**
+> and will be removed in a future release. Existing configs continue to work
+> but emit a deprecation warning at startup. Migrate to `vllm.http`. See
+> [Migration](#migration-from-udstokenizerconfig) below.
+
 ## Config
 
 | Parameter        | Default                 | Description                                                       |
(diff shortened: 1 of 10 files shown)


vMaroon:
@pierDipi @zdtsw

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/README.md
```diff
@@ -82,6 +88,36 @@ A complete sample config that pairs this with `precise-prefix-cache-scorer`
...
+```
+
+After:
+
+```yaml
+- type: token-producer
+  parameters:
+    modelName: "${MODEL_NAME}"
+    vllm:
+      http: "http://localhost:8000"   # or a shared render Service
```
elevran: nit: do you need `http://` as well? If so, calling the key `http` is redundant or misleading.
If this is a URL, consider using `vllm.url` or `vllm.endpoint`. If it is a host-port pair, do not use protocol in examples (i.e., `localhost:8000`, dropping the `http://` prefix).
Parsing in the code should be similar (i.e., do not `url.Parse()`, but a `net.SplitHostPort()`)

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -112,11 +115,15 @@ func LegacyPluginFactory(name string, rawParameters json.RawMessage, handle plug
...
 // NewPlugin creates a new tokenizer plugin instance and constructs the
 // configured backend. vllm is the default; udsTokenizerConfig is selected
-// only when explicitly enabled (its socketFile is set).
+// only when explicitly enabled (its socketFile is set) and is deprecated.
 func NewPlugin(ctx context.Context, config *tokenizerPluginConfig) (*Plugin, error) {
 	var tk tokenizer
 	switch {
 	case config.TokenizerConfig.IsEnabled():
+		log.FromContext(ctx).Info(
+			"DEPRECATION: 'udsTokenizerConfig' backend is deprecated and will be removed in a future release; migrate to the 'vllm' HTTP /render backend",
```
elevran: nit: be more explicit about the user configuration needed, not the system behind it.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/README.md
elevran: change the examples and be specific in the description that only HTTP is used (e.g., not TLS config)

ahg-g:
pls rebase

vMaroon:
@ahg-g done.

ahg-g:
/approve

Can we create an issue to track the deprecation pls and mark it for the release after next?

PR #1080, opened 2026-05-13
file-discovery-1: add BackendUpsert/Delete interface methods and extract upertEndpoint helper

ezrasilvera:
This PR is the first of 4 PRs adding file based discovery allowing the epp to run without Kubernetes. The file based discovery is the first phase in re-creating the functionality of #1007 

Two methods are added to the Datastore interface to give non-Kubernetes discovery giving a direct route to push endpoints into the store:

- BackendUpsert(ctx context.Context, meta *fwkdl.EndpointMetadata)
- BackendDelete(id types.NamespacedName)

The store/update logic that was inline in podUpdateOrAddIfNotExist is extracted into a private upsertEndpoint helper, which both the K8s path and BackendUpsert now share. The K8s code path behavior is unchanged.

### DIFF
--- a/pkg/epp/datastore/datastore.go
+++ b/pkg/epp/datastore/datastore.go
@@ -87,6 +87,11 @@ type Datastore interface {
 	PodUpdateOrAddIfNotExist(ctx context.Context, pod *corev1.Pod) bool
 	PodDelete(podName string)
 
+	// BackendUpsert adds or updates an endpoint from a non-Kubernetes discovery source.
+	BackendUpsert(ctx context.Context, meta *fwkdl.EndpointMetadata)
+	// BackendDelete removes the endpoint with the given namespaced name.
+	BackendDelete(id types.NamespacedName)
+
 	// Clears the store state, happens when the pool gets deleted.
 	Clear()
 }
@@ -334,23 +339,9 @@ func (ds *datastore) podUpdateOrAddIfNotExist(ctx context.Context, pod *corev1.P
 	existingEpSet := sets.Set[types.NamespacedName]{}
 	for _, endpointMetadata := range pods {
 		existingEpSet.Insert(endpointMetadata.NamespacedName)
-		var ep fwkdl.Endpoint
-		existing, ok := ds.pods.Load(endpointMetadata.NamespacedName)
-		if !ok {
-			ep = ds.epf.NewEndpoint(ds.parentCtx, endpointMetadata, ds)
-			if ep == nil {
-				// NewEndpoint returns nil when a collector is already running for this
-				// endpoint (duplicate reconcile race). The existing entry in ds.pods
-				// is still valid; skip re-registering it.
-				continue
-			}
-			ds.pods.Store(endpointMetadata.NamespacedName, ep)
+		if ds.upsertEndpoint(endpointMetadata) {
 			result = false
-		} else {
-			ep = existing.(fwkdl.Endpoint)
 		}
-		// Update endpoint properties if anything changed.
-		ep.UpdateMetadata(endpointMetadata)
 	}
 
 	// remove endpoints that are no longer active in the pool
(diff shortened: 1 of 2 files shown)


pkg/epp/datastore/datastore.go
```diff
@@ -87,6 +87,11 @@ type Datastore interface {
 	PodUpdateOrAddIfNotExist(ctx context.Context, pod *corev1.Pod) bool
 	PodDelete(podName string)
 
+	// BackendUpsert adds or updates an endpoint from a non-Kubernetes discovery source.
```
elevran, line 90: Q: can this be used (eventually), instead of `PodUpdateOrAddIfNotExist`, or do you expect that k8s endpoints would need special handling and can't be folded into this call directly from a Pod reconciler?

pkg/epp/datastore/datastore.go
```diff
@@ -380,6 +371,37 @@ func (ds *datastore) PodDelete(podName string) {
...
+		ds.epf.ReleaseEndpoint(v.(fwkdl.Endpoint))
+	}
+}
+
+// upsertEndpoint stores or updates a single endpoint in the pods map.
+// Returns true if the endpoint was newly created, false if it already existed
+// or if NewEndpoint returned nil (duplicate-start race).
+// Shared by BackendUpsert and podUpdateOrAddIfNotExist.
+func (ds *datastore) upsertEndpoint(meta *fwkdl.EndpointMetadata) bool {
+	existing, ok := ds.pods.Load(meta.NamespacedName)
```
elevran, line 389: not part of this PR: eventually need to replace the use of `pods` with endpoints since not every endpoint is a pod...

pkg/epp/datastore/datastore_test.go
```diff
@@ -47,6 +47,21 @@ import (
 	testutil "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/util/testing"
 )
 
+// simpleEndpointFactory is a minimal EndpointFactory for BackendUpsert/Delete tests.
```
elevran: nit: it is a `mock` object not a simple one.

pkg/epp/datastore/datastore_test.go
```diff
@@ -1330,3 +1345,60 @@ func TestExtractActivePorts(t *testing.T) {
...
 }
+
+// ---- BackendUpsert / BackendDelete tests -----------------------------------
+
+func TestBackendUpsert_NewEndpoint(t *testing.T) {
+	ctx := context.Background()
+	ds := NewDatastore(ctx, &simpleEndpointFactory{}, 0)
+	id := types.NamespacedName{Name: "ep1", Namespace: "default"}
+
+	ds.BackendUpsert(ctx, &fwkdl.EndpointMetadata{NamespacedName: id, Address: "10.0.0.1", Port: "8000"})
```
elevran: nit: suggest using constants for the fields that you set and validate on the endpoint to avoid cut-n-paste errors.

Unrelated: confused why is `EndpointMetadata.Port` a string... not related to this PR I presume.

pkg/epp/datastore/datastore_test.go
```diff
@@ -1330,3 +1345,60 @@ func TestExtractActivePorts(t *testing.T) {
...
+}
+
+func TestBackendDelete_Missing(t *testing.T) {
+	ctx := context.Background()
+	ds := NewDatastore(ctx, &simpleEndpointFactory{}, 0)
+
+	assert.NotPanics(t, func() {
+		ds.BackendDelete(types.NamespacedName{Name: "nonexistent", Namespace: "default"})
+	})
+}
```
elevran, line 1406: can you please add a test that the endpoint notification data source works regardless of how endpoints are added?

pkg/epp/datastore/datastore_test.go
ezrasilvera: Fixed in 0ef6919b

pkg/epp/datastore/datastore_test.go
ezrasilvera: Constants fixed in 41f6a7f7
Regarding the port as a string - it's stored as a string throughout the datalayer because it is always used in net.JoinHostPort(addr, port) and passed to strconv.Itoa and immediately stored as a string.

pkg/epp/datastore/datastore_test.go
ezrasilvera: In the next PR I'm implementing a NewDiscoveryNotifier, writing this test now would require importing or duplicating that constructor, this test will be included in the next PR (filediscovery-3)

pkg/epp/datastore/datastore.go
ezrasilvera: I prefer that the K8s path should stay separate for now - Pod->EndpointMetadata conversion (multi-port, labels, prune stale ports) has to live somewhere. BackendUpsert is the non-K8s generic path. Merging them in a future refactor is possible but require major changes and is out of scope here.

pkg/epp/datastore/datastore_test.go
ezrasilvera: Fixed in 4c317ddd (on filediscovery-3, which builds on this PR)

PR #1081, opened 2026-05-13
filediscovery-2: add DiscoveryBackendStore interface, NewDiscoveryNotifier …

ezrasilvera:
This PR is the second of 4 PRs adding file based discovery allowing the EPP to run
without Kubernetes. The file based discovery is the first phase in re-creating the
functionality of #1007.

This should be merged after #1080 so that Datastore already satisfies DiscoveryBackendStore when the forth PR lands.

Three additions, each in its own file:

- DiscoveryBackendStore (framework interface package): a narrow two-method interface that any store satisfying BackendUpsert/BackendDelete can implement. Lives in the framework interface package to avoid a circular import with pkg/epp/datastore.

- NewDiscoveryNotifier(store DiscoveryBackendStore) DiscoveryNotifier: wraps the store as a DiscoveryNotifier so discovery plugins can call Upsert/Delete without importing the datastore package directly.

- DiscoveryConfig / Discovery field on DataLayerConfig: YAML config support for selecting a discovery plugin by name. When omitted, the EPP default K8s path is unchanged.


elevran (commented):
/lgtm
/approve

PR #1082, opened 2026-05-13
filediscovery-3: add FileDiscovery plugin with YAML/JSON file loading and fsnotify watch

ezrasilvera:
This PR is the third of 4 PRs adding file based discovery allowing the EPP to run without Kubernetes. The file based discovery is the first phase in re-creating the functionality of #1007.

Depends on: #1081  (imports fwkdl.EndpointDiscovery and fwkdl.DiscoveryNotifier).  Must be merged after #1081 is merged

Adds the file-discovery plugin (plugin type: "file-discovery"). It implements EndpointDiscovery by reading a YAML or JSON file that lists inference endpoints, calling notifier.Upsert for each entry on load, and calling notifier.Delete for any entry that disappears between reloads.

Factory parameters:
- path (required): path to the endpoints file.
- watchFile (default false): if true, re-reads the file on every Write/Create event via fsnotify without restarting the plugin.


ezrasilvera:
Will re-open after #1080 and #1081 #1082

PR #1086, opened 2026-05-13
refactor(precise-prefix-cache): split into producer + slim scorer

vMaroon:
## Summary

Splits the `precise-prefix-cache-scorer` (Scorer + DataProducer + PreRequest + EndpointExtractor on one type) into:

- `precise-prefix-cache-producer` — new DataProducer; owns the KV-block index, kvevents pool, ZMQ subscribers, speculative TTL cache. Consumes `TokenizedPrompt`, writes `PrefixCacheMatchInfo`.
- `precise-prefix-cache-scorer` (slim) — pure attribute reader, returns `matchBlocks / totalBlocks`.

Old YAML keeps working: when parameters are passed to the scorer, the factory instantiates an internal Producer, logs a deprecation notice, and returns a wrapper fulfilling all four interfaces.

## Behavior changes

- **Attribute carries real values.** `PrefixCacheMatchInfo` is now `(matchLen, totalBlocks, blockSizeTokens)` end-to-end. Old code overwrote it with `(1,1,1)` in `Score`, silently breaking `prefix-based-pd-decider` (computed `1*1=1` token instead of real cached-prefix length) and `latency` / `prefixcacheaffinity` / `predictedlatency` (saw 0/1 binary ratios). `nohitlru` unaffected.
- **Scorer returns `match/total`** (matches `prefix-cache-scorer`), replacing min-max normalization. Fixes cold clusters scoring 1.0 everywhere and small absolute hits stretched to 1.0. Score distribution shape changes — re-tune the scorer weight if calibrated against the old output.
- **Producer always runs in DataProducer phase**, no longer gated on `speculativeIndexing`.

## Test plan

- [x] Unit tests for Produce paths (tokens vs prompt fallback, MM features, no-op exits), PreRequest, EndpointExtractor lifecycle, slim scorer, and legacy factory routing.
- [x] `go test -race ./...` and `golangci-lint run` clean.
- [x] End-to-end on cluster (Qwen3-32B, 2 vLLM replicas with ZMQ subscribers wired): 6 same-prefix requests all routed to the warmed pod with the expected score gap (`precise-prefix-cache-scorer` weight × match ratio).
- [ ] Benchmarks.

```release-note
split `precise-prefix-cache-scorer` into a scorer and a producer
```


shmuelk:
Please be more specific in the release notes section as to which plugins needed to eb added or deleted in one's configuration

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
```diff
@@ -0,0 +1,98 @@
...
+		_ = p.ensureSubscriber(ctx, ep.GetMetadata())
+	}
+}
+
+func (p *Producer) ensureSubscriber(ctx context.Context, meta *fwkdl.EndpointMetadata) error {
+	if meta == nil || meta.Address == "" {
+		return nil
+	}
+	endpointKey := meta.NamespacedName.String()
+	zmqEndpoint := fmt.Sprintf("tcp://%s:%d", meta.Address, p.kvEventsConfig.PodDiscoveryConfig.SocketPort)
```
shmuelk, line 86: How does this support DataParallel? Or does only vLLM instance fire the KV events?

vMaroon:
Closing in favor of:

- #1107
- #1112  
- #1121

PR #1098, opened 2026-05-13
rename docker images to router-endpoint-picker and router-disagg-sidecar

### LINKED ISSUE #1085: [Migration]: rename the docker images
As per #929, the images should be renamed to `router-endpoint-picker` and `router-disagg-sidecar`

tessapham:
**What type of PR is this?**

/kind cleanup

Fixes #1085.

```release-note
NONE
```


tessapham:
@ahg-g Lmk if I'm changing the right images.

.github/workflows/ci-dev.yaml
```diff
@@ -23,9 +23,8 @@ jobs:
       - name: Set image names
         id: version
         run: |
-          repo="${GITHUB_REPOSITORY##*/}"
-          echo "project_name=${repo}-dev" >> "$GITHUB_OUTPUT"
-          echo "sidecar_name=llm-d-routing-sidecar-dev" >> "$GITHUB_OUTPUT"
+          echo "project_name=router-endpoint-picker-dev" >> "$GITHUB_OUTPUT"
```
hexfusion: I believe we want to maintain the llm-d prefix here

```suggestion
          echo "project_name=llm-d-router-endpoint-picker-dev" >> "$GITHUB_OUTPUT"
```

config/charts/inferencepool/values.yaml
```diff
@@ -2,7 +2,7 @@ inferenceExtension:
   replicas: 1
   image:
     registry: ghcr.io/llm-d
-    repository: llm-d-inference-scheduler-dev
+    repository: router-endpoint-picker-dev
```
hexfusion: in this context we are describing the router holistically so this would be  `s/llm-d-inference-scheduler/llm-d-router/`?

tessapham: Oh good point. Does that mean we want to keep the `llm-d` prefix for all the image names in this PR?

tessapham: Updated all images with `llm-d-` prefix.

hexfusion: right to be clear we're not dropping the llm-d- prefix that's the org-level prefix in the ghcr path and stays. we're reflecting the project/repo rename from inference-scheduler to router in the image names, same as the repo, slack channel etc.

ahg-g: @elevran do we want to keep the llm-d prefix if we already have llm-d in the repository name?

hexfusion: possible I misunderstood intent, good to get consensus from maintainers.

elevran: > @elevran do we want to keep the llm-d prefix if we already have llm-d in the repository name?

My preference would be to drop the duplicate `llm-d/llm-d-*` everwhere but it seems to be an `llm-d` org decision.
I think it is redundant but removing it might cause inconsistency with other images/repos in the org,

tessapham: If there's concern with inconsistency maybe we can keep it for now and clean up later when there is an org-wide decision.

.github/workflows/ci-dev.yaml
```diff
@@ -23,9 +23,8 @@ jobs:
       - name: Set image names
         id: version
         run: |
-          repo="${GITHUB_REPOSITORY##*/}"
-          echo "project_name=${repo}-dev" >> "$GITHUB_OUTPUT"
-          echo "sidecar_name=llm-d-routing-sidecar-dev" >> "$GITHUB_OUTPUT"
+          echo "project_name=llm-d-router-endpoint-picker-dev" >> "$GITHUB_OUTPUT"
```
zdtsw: why not if we wanna keep the project name as prefix
```suggestion
          repo="${GITHUB_REPOSITORY##*/}"
          echo "project_name=${repo}-endpoint-picker-dev" >> "$GITHUB_OUTPUT"
          echo "sidecar_name=${repo}-disagg-sidecar-dev" >> "$GITHUB_OUTPUT"
```

tessapham: Good point. From my understanding `llm-d-router` is the project name but the image name should be `llm-d-endpoint-picker`. Could @ahg-g or @elevran confirm where to use which?

ahg-g: correct, the image name is llm-d-endpoint-picker, the image name is per component of llm-d router

tessapham: @zdtsw For here, we can add `epp_name` to be more explicit:

```suggestion
          echo "project_name=${repo}-dev" >> "$GITHUB_OUTPUT"
          echo "epp_name=${repo}-endpoint-picker-dev" >> "$GITHUB_OUTPUT"
```

Or just replace `project_name` -> `epp_name`.

ahg-g:
Thanks @tessapham , sorry for the conflict, can you pls address that?

tessapham:
@ahg-g Resolved conflicts!

.github/workflows/ci-dev.yaml
zdtsw: looks like `project_name` is not in need any more if `project_name` == `repo_name `== "llm-d-router" 
and two image name are `llm-d-endpoint-picker-dev` and `llm-d-disagg-sidecar-dev`
we could, hardcode image name as done in the first commit, or introduce a variable(but not much gain from that)

Makefile
```diff
@@ -18,9 +18,9 @@ include Makefile.gen.mk
 TARGETOS ?= $(shell command -v go >/dev/null 2>&1 && go env GOOS || uname -s | tr '[:upper:]' '[:lower:]')
 TARGETARCH ?= $(shell command -v go >/dev/null 2>&1 && go env GOARCH || uname -m | sed 's/x86_64/amd64/; s/aarch64/arm64/; s/armv7l/arm/')
 PROJECT_NAME ?= llm-d-router
-# TODO: rename EPP_IMAGE_NAME to router-endpoint-picker
-EPP_IMAGE_NAME ?= llm-d-inference-scheduler
-SIDECAR_IMAGE_NAME ?= llm-d-routing-sidecar
+# TODO: decide whether EPP_IMAGE_NAME shoudl be router-endpoint-picker or llm-d-router-endpoint-picker
```
ahg-g:
```suggestion
```

PR #1106, opened 2026-05-14
feat(precise-prefix-cache): DP support in discovery mode

vMaroon:
## Summary

Carry the rank index through endpoint metadata so multi-port pod deployments (one `TargetPort` per rank) can be distinguished downstream.

- `EndpointMetadata.RankIndex` = position in `pool.TargetPorts`, populated in the datastore.
- `precise-prefix-cache-scorer` consumes it: KV-events ZMQ subscriber now dials `SocketPort + RankIndex`, matching vLLM's [`offset_endpoint_port`](https://github.com/vllm-project/vllm/blob/main/vllm/distributed/kv_events.py) (`base + dp_rank`).
- Single-rank pods (`RankIndex=0`) hit the base port unchanged — no operator-side change required.

## Test plan

- [x] Unit: `go test ./...` green, `go vet` clean.
- [x] Cluster: 1 pod, DP=2, TP=2. EPP logs show two subscribers at distinct ports against the same pod IP:

  ```
  Ensured KV-events subscriber  endpoint=.../-rank-0  zmq=tcp://10.129.7.198:5556
  Ensured KV-events subscriber  endpoint=.../-rank-1  zmq=tcp://10.129.7.198:5557
  ```

```release-note
NONE
```

### DIFF
--- a/pkg/epp/datastore/datastore.go
+++ b/pkg/epp/datastore/datastore.go
@@ -326,6 +326,7 @@ func (ds *datastore) podUpdateOrAddIfNotExist(ctx context.Context, pod *corev1.P
 				Port:           strconv.Itoa(port),
 				MetricsHost:    net.JoinHostPort(pod.Status.PodIP, strconv.Itoa(metricsPort)),
 				Labels:         labels,
+				RankIndex:      idx,
 			})
 	}
 
--- a/pkg/epp/framework/interface/datalayer/endpoint_metadata.go
+++ b/pkg/epp/framework/interface/datalayer/endpoint_metadata.go
@@ -31,6 +31,9 @@ type EndpointMetadata struct {
 	Port           string
 	MetricsHost    string
 	Labels         map[string]string
+	// RankIndex is this endpoint's position in the pool's TargetPorts,
+	// identifying the pod-local rank in multi-port deployments.
+	RankIndex int
 }
 
 // String returns a string representation of the endpoint.
@@ -59,7 +62,17 @@ func (epm *EndpointMetadata) Clone() *EndpointMetadata {
 		Port:        epm.Port,
 		MetricsHost: epm.MetricsHost,
 		Labels:      clonedLabels,
+		RankIndex:   epm.RankIndex,
 	}
+}
+
+// GetRankIndex returns the rank index of this endpoint within the pool's
+// TargetPorts list.
+func (epm *EndpointMetadata) GetRankIndex() int {
+	if epm == nil {
+		return 0
+	}
+	return epm.RankIndex
 }
 
 // GetNamespacedName gets the namespace name of the Endpoint.
--- a/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
+++ b/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
@@ -654,7 +654,12 @@ func (s *Scorer) ensureSubscriber(ctx context.Context, meta *fwkdl.EndpointMetad
 		return nil
 	}
 	endpointKey := meta.NamespacedName.String()
-	zmqEndpoint := fmt.Sprintf("tcp://%s:%d", meta.Address, s.kvEventsConfig.PodDiscoveryConfig.SocketPort)
+	// Mirror vLLM's `offset_endpoint_port` rule: rank N binds at
+	// SocketPort + N. For single-rank pods (RankIndex=0) this collapses to
+	// the legacy single-port-per-pod behaviour; multi-rank wide-EP / DP
+	// pods get one subscriber per rank automatically.
+	port := s.kvEventsConfig.PodDiscoveryConfig.SocketPort + meta.RankIndex
+	zmqEndpoint := fmt.Sprintf("tcp://%s:%d", meta.Address, port)
 
 	logger := log.FromContext(ctx).WithName(s.typedName.String())
 	if err := s.subscribersManager.EnsureSubscriber(s.subscriberCtx, endpointKey,
--- a/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
+++ b/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
@@ -658,7 +658,7 @@ func (s *Scorer) ensureSubscriber(ctx context.Context, meta *fwkdl.EndpointMetad
 	// SocketPort + N. For single-rank pods (RankIndex=0) this collapses to
 	// the legacy single-port-per-pod behaviour; multi-rank wide-EP / DP
 	// pods get one subscriber per rank automatically.
-	port := s.kvEventsConfig.PodDiscoveryConfig.SocketPort + meta.RankIndex
+	port := s.kvEventsConfig.PodDiscoveryConfig.SocketPort + meta.GetRankIndex()
 	zmqEndpoint := fmt.Sprintf("tcp://%s:%d", meta.Address, port)
 
 	logger := log.FromContext(ctx).WithName(s.typedName.String())
--- a/pkg/epp/datastore/datastore_test.go
+++ b/pkg/epp/datastore/datastore_test.go
@@ -670,6 +670,7 @@ func TestEndpointMetadata(t *testing.T) {
 					Port:        inferencePoolMultiTargetPort1,
 					MetricsHost: net.JoinHostPort(pod1.Status.PodIP, inferencePoolMultiTargetPort1),
 					Labels:      map[string]string{},
+					RankIndex:   1,
 				},
 			},
 			op: func(ctx context.Context, ds Datastore) {
@@ -704,6 +705,7 @@ func TestEndpointMetadata(t *testing.T) {
 					Port:        inferencePoolMultiTargetPort1,
 					MetricsHost: net.JoinHostPort(pod1.Status.PodIP, inferencePoolMultiTargetPort1),
 					Labels:      map[string]string{},
+					RankIndex:   1,
 				},
 				{
 					NamespacedName: types.NamespacedName{
@@ -728,6 +730,7 @@ func TestEndpointMetadata(t *testing.T) {
 					Port:        inferencePoolMultiTargetPort1,
 					MetricsHost: net.JoinHostPort(pod1.Status.PodIP, inferencePoolMultiTargetPort1),
 					Labels:      map[string]string{},
+					RankIndex:   1,
 				},
 			},
 			op: func(ctx context.Context, ds Datastore) {
@@ -762,6 +765,7 @@ func TestEndpointMetadata(t *testing.T) {
 					Port:        inferencePoolMultiTargetPort1,
 					MetricsHost: net.JoinHostPort(pod1.Status.PodIP, inferencePoolMultiTargetPort1),
 					Labels:      map[string]string{},
+					RankIndex:   1,
 				},
 			},
 			op: func(ctx context.Context, ds Datastore) {
--- a/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache_extractor_test.go
+++ b/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache_extractor_test.go
@@ -233,6 +233,75 @@ func TestScorer_LegacyInScoreDiscovery_DiscoverPodsDisabled(t *testing.T) {
 	assert.Empty(t, ids)
 }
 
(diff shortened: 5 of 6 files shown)


pkg/epp/framework/interface/datalayer/endpoint_metadata.go
```diff
@@ -31,6 +31,9 @@ type EndpointMetadata struct {
 	Port           string
 	MetricsHost    string
 	Labels         map[string]string
+	// RankIndex is this endpoint's position in the pool's TargetPorts,
+	// identifying the pod-local rank in multi-port deployments.
+	RankIndex int
```
hexfusion, line 36: question: would TargetPortIndex be more in line with what this actually is, literally a position in pool.TargetPorts?  consumers like your precise-prefix-cache scorer could derive dpRank from it, matching vLLM's [new_port = base_port + data_parallel_rank](https://github.com/vllm-project/vllm/blob/ae4f59f0ece88ca25ba1064ae6d3b512c3bfc606/vllm/distributed/kv_events.py#L467-L499)

vMaroon: The port this adds up to is the ZMQ port used for the KVEvents pub-sub socket. If I go with target-port, I'd need to know base-port to derive the index. The index is the raw information that signals that an endpoint is a multi-rank one, which is very common in inference deployments. From there consumers can do what they need with it.

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/jellydator/ttlcache/v3 v3.4.0
+	github.com/llm-d/llm-d-inference-scheduler v0.8.0
```
vMaroon: The sidecar still imports llm-d-inference-scheduler https://github.com/llm-d/llm-d-router/blob/821dbc3a21e5ba7ff1364e5264f6427490491095/pkg/sidecar/proxy/decode.go#L29 (cc @zdtsw)

elevran:
Please consider using endpoint Attributes instead of expanding the metadata state directly

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -657,7 +657,12 @@ func (s *Scorer) ensureSubscriber(ctx context.Context, meta *fwkdl.EndpointMetad
 		return nil
 	}
 	endpointKey := meta.NamespacedName.String()
-	zmqEndpoint := fmt.Sprintf("tcp://%s:%d", meta.Address, s.kvEventsConfig.PodDiscoveryConfig.SocketPort)
+	// Mirror vLLM's `offset_endpoint_port` rule: rank N binds at
+	// SocketPort + N. For single-rank pods (RankIndex=0) this collapses to
+	// the legacy single-port-per-pod behaviour; multi-rank wide-EP / DP
+	// pods get one subscriber per rank automatically.
+	port := s.kvEventsConfig.PodDiscoveryConfig.SocketPort + meta.RankIndex
```
zdtsw:
```suggestion
	port := s.kvEventsConfig.PodDiscoveryConfig.SocketPort + meta.GetRankIndex()
```
if we define a new GetRankIndex() already

pkg/epp/datastore/datastore.go
```diff
@@ -38,6 +38,7 @@ import (
 	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
 	"github.com/llm-d/llm-d-router/pkg/epp/datalayer"
 	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+	attrtopology "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/attribute/topology"
```
elevran: `topology` is misleading, IMO. We have had requests for topology aware routing (think: DC, AZ, rack, host, ....) - this seems to be focused on the rank.

pkg/epp/datastore/datastore.go
```diff
@@ -308,7 +309,11 @@ func (ds *datastore) podUpdateOrAddIfNotExist(ctx context.Context, pod *corev1.P
 	if len(pool.TargetPorts) == 1 {
 		modelServerMetricsPort = int(ds.modelServerMetricsPort)
 	}
-	pods := []*fwkdl.EndpointMetadata{}
+	type rankedMeta struct {
```
elevran: what's the reason for changing datastore to support a new attribute?
Other attributes are handled from outside the store and do not require changes here (i.e., solely use the Get/Put interface)

pkg/epp/datastore/datastore.go
```diff
@@ -337,11 +344,14 @@ func (ds *datastore) podUpdateOrAddIfNotExist(ctx context.Context, pod *corev1.P
...
-	for _, endpointMetadata := range pods {
-		existingEpSet.Insert(endpointMetadata.NamespacedName)
-		if ds.upsertEndpoint(endpointMetadata) {
+	for _, rm := range pods {
+		existingEpSet.Insert(rm.meta.NamespacedName)
+		if ds.upsertEndpoint(rm.meta) {
 			result = false
 		}
+		if ep, ok := ds.pods.Load(rm.meta.NamespacedName); ok {
+			seedRankIndex(ep.(fwkdl.Endpoint), rm.rank)
```
elevran: The `rank` attribute is seeded after `ds.upsertEndpoint`. The `upsertEndpoint` calls `epf.NewEndpoint` synchronously dispatches `EventAddOrUpdate` before returning. So the extractor in `precise_prefix_cache.go` runs `ensureSubscriber` with the rank attribute **still unset for any endpoint with idx >= 1**, and falls back to rank = 0.

In practice this self-corrects on the first `Score` call via the backwards-compat `ensureSubscribersForEndpoints` path where `kvevents.SubscriberManager.EnsureSubscriber` does tear down and recreate. However:

1. KV events arriving in the window between endpoint creation and the first Score are misattributed (rank >=1 all listen on the rank-0 socket).
2. Any deployment that runs the data-layer path without the in-Score legacy path never gets corrected.
3. Tests pass because TestScorer_ExtractEndpoint_OffsetsZMQPortByRankIndex sets the attribute on the endpoint before invoking ExtractEndpoint manually, bypassing the production ordering.
4. This is resource wasteful as sockets are created and destroyed.

A minimal regression test that goes through epf.NewEndpoint (not just direct ExtractEndpoint) with idx > 0 would catch this.

pkg/epp/framework/interface/scheduling/types.go
```diff
@@ -71,6 +71,7 @@ func (r *InferenceRequest) String() string {
 type Endpoint interface {
 	GetMetadata() *fwkdl.EndpointMetadata
 	GetMetrics() *fwkdl.Metrics
+	GetAttributes() fwkdl.AttributeMap
```
elevran: why is this needed? If you need a specific attribute, use `Get`

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -642,19 +643,40 @@ func (s *Scorer) ExtractEndpoint(ctx context.Context, event fwkdl.EndpointEvent)
 	return nil
 }
 
+// subscriberEndpoint is the minimal endpoint shape ensureSubscriber needs:
+// metadata for dialing and attributes for reading the rank index. Both
+// fwkdl.Endpoint (data-layer path) and scheduling.Endpoint (legacy in-Score
+// path) satisfy it, so callers pass what they already have.
+type subscriberEndpoint interface {
+	GetMetadata() *fwkdl.EndpointMetadata
+	GetAttributes() fwkdl.AttributeMap
```
elevran: why is `Get(RankIndexKey)` for the specific attribute not sufficient?
Seems too broad access for a single attribute.

pkg/epp/datastore/datastore.go
elevran: Suggestion:
if using the (default/zero) rank attribute value to mean unset, the whole path could be simplified.
Use rank-1 as the base and leave `0` as undefined?

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -642,19 +643,40 @@ func (s *Scorer) ExtractEndpoint(ctx context.Context, event fwkdl.EndpointEvent)
...
 		return nil
 	}
 	endpointKey := meta.NamespacedName.String()
-	zmqEndpoint := fmt.Sprintf("tcp://%s:%d", meta.Address, s.kvEventsConfig.PodDiscoveryConfig.SocketPort)
+	// Mirror vLLM's `offset_endpoint_port` rule: rank N binds at
+	// SocketPort + N. For single-rank pods (no rank attribute, treated as 0)
+	// this collapses to the legacy single-port-per-pod behaviour; multi-rank
+	// wide-EP / DP pods get one subscriber per rank automatically.
+	rank := 0
+	if attrs := ep.GetAttributes(); attrs != nil {
```
elevran: Just do a Get with the specific attribute you need. You can still type-assert the return but there should not be a reason to iterate and skip all other attributes.

pkg/epp/framework/plugins/datalayer/attribute/topology/data_types.go
```diff
@@ -0,0 +1,35 @@
...
+// (e.g. KV-events subscribers in the precise-prefix-cache scorer) read off
+// the endpoint's AttributeMap.
+package topology
+
+import (
+	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+)
+
+// RankIndexKey identifies the endpoint's position in its pool's TargetPorts.
+const RankIndexKey = "RankIndex"
```
elevran: whichever component produces this should implement `ProducerPlugin` and its `Produces`. A consumer should similarly define Consumes so that the DAG can be created correctly.

pkg/epp/datastore/datastore.go
ahg-g: We should add Rank as a property of EndpointMetadata, it is already embedded as part of `NamespacedName`, which I think is misleading/confusing.

pkg/epp/framework/plugins/datalayer/attribute/topology/data_types.go
ahg-g: I think we should drop this attribute completely and just make Rank a builtin attribute of EndpointMetadata.

pkg/epp/framework/interface/scheduling/types.go
vMaroon: `fwkdl.Endpoint` (datalayer) has it. Pieces that look at both endpoint interfaces have a weird view right now.

pkg/epp/datastore/datastore.go
elevran: @ahg-g if it is an optional attribute, prefer that we make non-essential attributes part of the AttributeMap and not load into metadata state.

pkg/epp/framework/plugins/datalayer/attribute/topology/data_types.go
vMaroon: I think reverting back to `EndpointMetadata` makes sense as rank is key to the identity. Today we don't do much with it, but DP endpoints are coupled in health and performance.

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -654,7 +654,12 @@ func (s *Scorer) ensureSubscriber(ctx context.Context, meta *fwkdl.EndpointMetad
 		return nil
 	}
 	endpointKey := meta.NamespacedName.String()
-	zmqEndpoint := fmt.Sprintf("tcp://%s:%d", meta.Address, s.kvEventsConfig.PodDiscoveryConfig.SocketPort)
+	// Mirror vLLM's `offset_endpoint_port` rule: rank N binds at
+	// SocketPort + N. For single-rank pods (RankIndex=0) this collapses to
+	// the legacy single-port-per-pod behaviour; multi-rank wide-EP / DP
+	// pods get one subscriber per rank automatically.
+	port := s.kvEventsConfig.PodDiscoveryConfig.SocketPort + meta.GetRankIndex()
```
elevran, line 661: nit: this assumes that the port range is always consecutive.
Is it possible to deploy non-continguous port ranges (it is possible in the InferencePool since that's an array, not base + size)?
If possible, then the documentation should clearly reflect the assumed constraint.

elevran (commented):
/lgtm

PR #1107, opened 2026-05-14
[1/3] feat(precise-scorer): switch to absolute matched/total normalization

vMaroon:
## Summary

Switch `precise-prefix-cache-scorer` from cross-endpoint min-max normalization to per-endpoint `matchBlocks / totalBlocks`, clipped to `[0, 1]`. Matches the approximate `prefix-cache-scorer` semantics.

## Tests
- [x] `go test -race ./pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/...`
- [x] `golangci-lint run` clean.

```release-note
The precise-prefix-cache-scorer now emits absolute matched_blocks/total_blocks (clamped to [0,1]) instead of cross-endpoint min-max normalization. If you have tuned the scorer's weight for a specific workload, you may need to re-tune it.
```

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
+++ b/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
@@ -749,8 +749,23 @@ func (s *Scorer) getBlockSizeTokens() int {
 	return s.blockSizeTokens
 }
 
+// scoreBlockKeys computes per-pod scores from precomputed block keys, avoiding
+// re-tokenization in the legacy prompt/chat fallback paths. Empty input
+// returns an empty score map.
+func (s *Scorer) scoreBlockKeys(ctx context.Context, blockKeys []kvblock.BlockHash) (map[string]float64, error) {
+	if len(blockKeys) == 0 {
+		return map[string]float64{}, nil
+	}
+	keyToPods, err := s.kvCacheIndexer.KVBlockIndex().Lookup(ctx, blockKeys, nil)
+	if err != nil {
+		return nil, fmt.Errorf("lookup: %w", err)
+	}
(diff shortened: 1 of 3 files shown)


elevran:
overall, /lgtm, leaving copilot review comments to your discretion.
Two things that are worth addressing:
- the release notes say `NONE` but PR description explicitly states "_If the scorer's weight was tuned against the old output, you may need to re-tune_". This suggests a possible change visible to operators and may warrant a release note with some guidance.
- `absoluteScoredPods` has no direct unit test. It's covered indirectly by the UDS integration test, but (1) that is being deprecated and (2) the `totalBlocks <= 0` (cold-cluster yields all zeros) and `ratio > 1.0` (clip) paths are non-trivial and only tested through the UDS integration test, which (1) is being deprecated and (2) requires the full scorer stack with a running socket. A small table-driven unit test in a `utils_test.go` would likely cover these branches deterministically.

Consider tuning the release notes along these lines:
```
The precise-prefix-cache-scorer now emits absolute matched_blocks/total_blocks (clamped to [0,1]) instead of cross-endpoint min-max normalization. If you have tuned the scorer's weight for a specific workload, you may need to re-tune it.
```

vMaroon:
> the release notes say NONE but PR description explicitly states "If the scorer's weight was tuned against the old output, you may need to re-tune". This suggests a possible change visible to operators and may warrant a release note with some guidance.

Updated note.

/approve

nirrozenbaum:
joining Etai’s comment: 
IMO this is a good and correct change. 
but since prefix cache scorer **is the most dominant** one in default configuration, I believe this change requires re-running benchmarks cause it may affect performance.

vMaroon:
Benchmarks were ran, and this behavior went through an IBM internal "production environment". The 2:2:3 weights still work as-is which makes sense as we ship the same defaults for approx and precise paths. Users who tune their weights based on the previous signal are the audience of the note, though I doubt they exist.

PR #1112, opened 2026-05-14
[2/3] feat(precise-prefix-cache): add precise-prefix-cache-producer

vMaroon:
## Summary

New `precise-prefix-cache-producer` DataProducer plugin. Purely additive — existing `precise-prefix-cache-scorer` untouched.

Owns: KV-block index, KV-events pool, per-pod ZMQ subscriber lifecycle (`EndpointExtractor` on `endpoint-notification-source`), speculative TTL cache. Writes `PrefixCacheMatchInfo` per endpoint under its name-bound key, consuming `TokenizedPrompt` from `token-producer`. No internal tokenization fallback — `tokenizersPoolConfig` is rejected at the factory.

Pairs with the existing generic `prefix-cache-scorer` via explicit name reference:

```yaml
- type: prefix-cache-scorer
  parameters:
    prefixMatchInfoProducerName: precise-prefix-cache-producer
```

Without the field, the scorer keeps reading the auto-spawned approx producer as today.

## Test plan

- [x] `go test -race ./pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/...`
- [x] `golangci-lint run` clean.

```release-note
Adds `precise-prefix-cache-producer` DataProducer plugin. Configure alongside `prefix-cache-scorer` with `prefixMatchInfoProducerName: precise-prefix-cache-producer` to opt into precise KV-block scoring; the existing monolithic `precise-prefix-cache-scorer` is unaffected.
```

### DIFF
--- a/deploy/config/epp-precise-prefix-cache-split-config.yaml
+++ b/deploy/config/epp-precise-prefix-cache-split-config.yaml
@@ -1,0 +1,47 @@
+# Sample EPP config: split precise-prefix-cache pipeline
+#   token-producer -> precise-prefix-cache-producer -> prefix-cache-scorer
+apiVersion: llm-d.ai/v1alpha1
+kind: EndpointPickerConfig
+plugins:
+  - type: token-producer
+    parameters:
+      modelName: hf-repo/model-name           # set to the model used in the vLLM deployment
+      vllm:
+        http: http://localhost:8000
+  - type: endpoint-notification-source
+  - type: metrics-data-source
+  - type: core-metrics-extractor
+  - type: single-profile-handler
+  - type: decode-filter
+  - type: precise-prefix-cache-producer
+    parameters:
+      tokenProcessorConfig:
+        blockSize: 64
+      indexerConfig:
+        kvBlockIndexConfig:
+          enableMetrics: true       # enable kv-block index metrics (prometheus)
+  - type: prefix-cache-scorer
+    parameters:
+      prefixMatchInfoProducerName: precise-prefix-cache-producer
+  - type: kv-cache-utilization-scorer
+  - type: queue-scorer
+  - type: max-score-picker
+dataLayer:
+  sources:
+    - pluginRef: metrics-data-source
+      extractors:
+        - pluginRef: core-metrics-extractor
+    - pluginRef: endpoint-notification-source
+      extractors:
+        - pluginRef: precise-prefix-cache-producer
+schedulingProfiles:
+  - name: default
+    plugins:
+      - pluginRef: decode-filter
+      - pluginRef: prefix-cache-scorer
+        weight: 2.0
+      - pluginRef: kv-cache-utilization-scorer
+        weight: 1.0
+      - pluginRef: queue-scorer
+        weight: 1.0
+      - pluginRef: max-score-picker
--- a/pkg/epp/datalayer/data_graph.go
+++ b/pkg/epp/datalayer/data_graph.go
@@ -117,6 +117,10 @@ func CreateMissingDataProducers(ctx context.Context, defaultProducerRegistry map
 			return fmt.Errorf("auto-created default entry %q is not a ProducerPlugin, this is required by datakey: %v, which is consumed by: %v", defaultProducerNameOrType, key, consumerName)
 		}
 		handle.AddPlugin(plg.TypedName().Name, plg)
+		logger.Info("auto-created default producer",
+			"producer", plg.TypedName().String(),
+			"dataKey", key,
+			"consumer", consumerName)
 	}
 
 	return nil
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package preciseprefixcache
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+)
+
+var _ fwkdl.EndpointExtractor = &Producer{}
+
+func (p *Producer) ExpectedInputType() reflect.Type {
+	return fwkdl.EndpointEventReflectType
+}
+
+// ExtractEndpoint: add/update installs a per-pod ZMQ subscriber, delete tears
+// it down. Sole per-pod discovery path; global socket mode is the alternative.
+func (p *Producer) ExtractEndpoint(ctx context.Context, event fwkdl.EndpointEvent) error {
+	if !p.kvEventsConfig.DiscoverPods || p.kvEventsConfig.PodDiscoveryConfig == nil {
+		return nil
+	}
+	meta := event.Endpoint.GetMetadata()
+	if meta == nil || meta.NamespacedName.Name == "" {
+		return nil
+	}
+
+	logger := log.FromContext(ctx).WithName(p.typedName.String())
+	endpointKey := meta.NamespacedName.String()
+
+	switch event.Type {
+	case fwkdl.EventAddOrUpdate:
+		if err := p.ensureSubscriber(ctx, meta); err != nil {
+			return err
+		}
+		logger.V(logging.DEBUG).Info("Adding subscriber", "endpoint", endpointKey)
+	case fwkdl.EventDelete:
+		p.subscribersManager.RemoveSubscriber(ctx, endpointKey)
+		logger.V(logging.DEBUG).Info("Removed KV-events subscriber", "endpoint", endpointKey)
+	}
+	return nil
+}
+
+func (p *Producer) ensureSubscriber(ctx context.Context, meta *fwkdl.EndpointMetadata) error {
+	if meta == nil || meta.Address == "" {
+		return nil
+	}
+	endpointKey := meta.NamespacedName.String()
+	// vLLM offset_endpoint_port: rank N binds at SocketPort + N.
+	port := p.kvEventsConfig.PodDiscoveryConfig.SocketPort + meta.GetRankIndex()
+	zmqEndpoint := fmt.Sprintf("tcp://%s:%d", meta.Address, port)
+
+	logger := log.FromContext(ctx).WithName(p.typedName.String())
+	// Use subscriberCtx (plugin-lifetime), not the caller's ctx.
+	if err := p.subscribersManager.EnsureSubscriber(p.subscriberCtx, endpointKey,
+		zmqEndpoint, p.kvEventsConfig.TopicFilter, true); err != nil {
+		logger.Error(err, "Failed to ensure KV-events subscriber for endpoint",
+			"endpoint", endpointKey, "address", meta.Address)
+		return fmt.Errorf("ensure subscriber for %s: %w", endpointKey, err)
+	}
+	logger.V(logging.DEBUG).Info("Ensured KV-events subscriber", "endpoint", endpointKey, "zmq", zmqEndpoint)
+	return nil
+}
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
@@ -37,7 +37,7 @@ func (p *Producer) ExpectedInputType() reflect.Type {
 // ExtractEndpoint processes endpoint lifecycle events emitted by the
 // endpoint-notification-source: add/update installs a per-pod ZMQ KV-events
 // subscriber, delete tears one down. No-op unless per-pod discovery is
-// enabled; the alternative is global socket mode (KVEventsConfig.ZMQEndpoint).
+// enabled.
 func (p *Producer) ExtractEndpoint(ctx context.Context, event fwkdl.EndpointEvent) error {
 	if !p.kvEventsConfig.DiscoverPods || p.kvEventsConfig.PodDiscoveryConfig == nil {
 		return nil
@@ -64,7 +64,7 @@ func (p *Producer) ExtractEndpoint(ctx context.Context, event fwkdl.EndpointEven
 }
 
 // ensureSubscriber idempotently installs a KV-events subscriber for the given
-// endpoint, dialing SocketPort + RankIndex to match vLLM's offset_endpoint_port
+// endpoint, dialing SocketPort + RankIndex to match standard inference-engine port offsetting
 // (one ZMQ PUB socket per DP rank on the same pod IP).
 func (p *Producer) ensureSubscriber(ctx context.Context, meta *fwkdl.EndpointMetadata) error {
 	if meta == nil || meta.Address == "" {
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
@@ -29,12 +29,15 @@ import (
 
 var _ fwkdl.EndpointExtractor = &Producer{}
 
+// ExpectedInputType reports the data-layer event type this extractor consumes.
 func (p *Producer) ExpectedInputType() reflect.Type {
 	return fwkdl.EndpointEventReflectType
 }
 
-// ExtractEndpoint: add/update installs a per-pod ZMQ subscriber, delete tears
-// it down. Sole per-pod discovery path; global socket mode is the alternative.
+// ExtractEndpoint processes endpoint lifecycle events emitted by the
+// endpoint-notification-source: add/update installs a per-pod ZMQ KV-events
+// subscriber, delete tears one down. No-op unless per-pod discovery is
+// enabled; the alternative is global socket mode (KVEventsConfig.ZMQEndpoint).
 func (p *Producer) ExtractEndpoint(ctx context.Context, event fwkdl.EndpointEvent) error {
 	if !p.kvEventsConfig.DiscoverPods || p.kvEventsConfig.PodDiscoveryConfig == nil {
 		return nil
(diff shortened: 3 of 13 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
```diff
@@ -0,0 +1,174 @@
...
+	) {
+		if reason != ttlcache.EvictionReasonExpired {
+			return
+		}
+		entries := item.Value()
+		for _, reqKey := range entries.blockKeys {
+			// Speculative entries were added without engineKey mapping, so
+			// evict by RequestKey directly.
+			//nolint:errcheck // best-effort cleanup on TTL expiry
+			index.Evict(context.Background(), reqKey, kvblock.RequestKey, entries.podEntries)
```
hexfusion: question: why not use the real ctx here?

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
```diff
@@ -0,0 +1,98 @@
...
+		_ = p.ensureSubscriber(ctx, ep.GetMetadata())
+	}
+}
+
+func (p *Producer) ensureSubscriber(ctx context.Context, meta *fwkdl.EndpointMetadata) error {
+	if meta == nil || meta.Address == "" {
+		return nil
+	}
+	endpointKey := meta.NamespacedName.String()
+	zmqEndpoint := fmt.Sprintf("tcp://%s:%d", meta.Address, p.kvEventsConfig.PodDiscoveryConfig.SocketPort)
```
bongwoobak: PR #1106 added SocketPort + meta.RankIndex in the scorer's ensureSubscriber to support data-parallel deployments where each DP rank binds ZMQ at base + dp_rank. The producer's ensureSubscriber here still dials the bare SocketPort, so in DP/wide-EP deployments every rank will be subscribed on the same port - collapsing all ranks onto one subscriber.
```suggestion
	port := p.kvEventsConfig.PodDiscoveryConfig.SocketPort + meta.RankIndex
    zmqEndpoint := fmt.Sprintf("tcp://%s:%d", meta.Address, port)
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+	})
+	go cleanCachePeriodically(ctx, cache, ttl)
+
+	return cache, ttl, nil
+}
+
+// PreRequest seeds speculative entries for the selected endpoint(s).
+// No-op when speculative indexing is disabled.
+func (p *Producer) PreRequest(ctx context.Context,
```
elevran, line 114: `PreRequest` implements non-trivial speculative-indexing logic (state read + delete, `index.Add` for primary and prefill endpoints, TTL cache entry) but there are no tests for it.

Please consider adding test cases for at least:
- `speculativeEnabled=true` with populated `blockKeys`: assert `index.Add` called with expected pod identifier and keys; assert TTL entry inserted in cache.
- `speculativeEnabled=true` with empty `blockKeys`: assert `index.Add` **not** called.
- P/D prefill-profile branch: `schedulingResult` has both primary and `"prefill"` profiles; assert `index.Add` called twice.
- `speculativeEnabled=false`: assert early return, nothing written.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -0,0 +1,248 @@
...
+	return p
+}
+
+func (p *Producer) Produces() map[string]any {
+	return map[string]any{attrprefix.PrefixCacheMatchInfoKey: attrprefix.PrefixCacheMatchInfo{}}
+}
+
+// Consumes declares the TokenizedPrompt dependency so the data-layer DAG
+// orders `token-producer` first. Configs without a token-producer fall back
+// to the indexer's internal tokenization in computeBlockKeys.
```
elevran: Stale comment after this PR.

```suggestion
// Consumes declares the TokenizedPrompt dependency so the data-layer DAG
// orders `token-producer` first. No-op when TokenizedPrompt is absent.
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
```diff
@@ -0,0 +1,174 @@
...
+	if len(state.blockKeys) == 0 {
+		return
+	}
+
+	primary := schedulingResult.ProfileResults[schedulingResult.PrimaryProfileName]
+	if primary == nil || len(primary.TargetEndpoints) == 0 {
+		return
+	}
+	targetEndpoint := primary.TargetEndpoints[0]
+	targetMeta := targetEndpoint.GetMetadata()
```
elevran, line 141: `GetMetadata()` can return nil; `targetMeta.Address` on a nil pointer panics.

```suggestion
	targetMeta := targetEndpoint.GetMetadata()
	if targetMeta == nil {
		return
	}
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
```diff
@@ -0,0 +1,174 @@
...
+	if err := index.Add(ctx, nil, state.blockKeys, []kvblock.PodEntry{speculativePod}); err != nil {
+		logger.Error(err, "Failed to add speculative entries to index",
+			"pod", speculativePod.PodIdentifier)
+	}
+
+	allPodEntries := []kvblock.PodEntry{speculativePod}
+
+	// P/D disaggregation: also seed the prefill endpoint.
+	if pr, exists := schedulingResult.ProfileResults[experimentalPrefillProfile]; exists && len(pr.TargetEndpoints) > 0 {
+		prefillMeta := pr.TargetEndpoints[0].GetMetadata()
```
elevran: Same nil-guard missing for the prefill endpoint's metadata here.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer_test.go
```diff
@@ -0,0 +1,294 @@
...
+
+func TestProduces_DeclaresPrefixCacheMatchInfo(t *testing.T) {
+	p := &Producer{typedName: plugin.TypedName{Type: PluginType, Name: "x"}}
+	_, ok := p.Produces()[attrprefix.PrefixCacheMatchInfoKey]
+	require.True(t, ok)
+}
+
+func TestConsumes_DeclaresTokenizedPrompt(t *testing.T) {
+	p := &Producer{typedName: plugin.TypedName{Type: PluginType, Name: "x"}}
+	_, ok := p.Consumes()["TokenizedPrompt"]
```
elevran: nit: consider a string literal instead of the constant. If `TokenizedPromptKey` is ever renamed, this test silently passes while `Consumes()` drifts.

deploy/config/epp-precise-prefix-cache-split-config.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - type: single-profile-handler
+  - type: decode-filter
+  - type: precise-prefix-cache-producer
+    parameters:
+      tokenProcessorConfig:
+        blockSize: 64
+      indexerConfig:
+        kvBlockIndexConfig:
+          enableMetrics: true       # enable kv-block index metrics (prometheus)
+  - type: precise-prefix-cache-scorer
```
elevran: nit: this is the old monolithic scorer, duplicating the producer's work as noted in the PR description. While the slim scorer (step 3/3) will replace it here, a short comment would prevent users from treating this file as the final recommended config in the meantime.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create %s plugin: %w", PluginType, err)
+	}
+
+	return p.WithName(name), nil
+}
+
+// New initializes a Producer. The kvcache indexer and KV-events pool are
+// started in background goroutines bound to ctx.
+func New(ctx context.Context, config PluginConfig) (*Producer, error) {
```
elevran: Calling `New()` on the producer creates a complete independent stack: its own `kvcache.Indexer`, KV-events pool, and `SubscriberManager`. The existing `precise-prefix-cache-scorer.New()` does the same. In the sample split-pipeline config, registering both plugins means **two separate ZMQ pools, two kvcache indexes, and two subscriber managers run simultaneously**, the producer's index is populated but its `PrefixCacheMatchInfo` is immediately overwritten by the scorer, which scores from its own independent index. The split config is therefore a no-op at runtime until the slim scorer (step 3/3) replaces the monolithic one entirely.

Consider either (a) omitting the sample split config until the slim scorer exists, or (b) having the scorer's `PluginFactory` detect a co-registered producer and delegate to it instead of creating its own stack?

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/utils.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+	"time"
+
+	"github.com/jellydator/ttlcache/v3"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/scheduling"
+)
+
+func extractPodSet(endpoints []scheduling.Endpoint) sets.Set[string] {
```
elevran: `extractPodSet` is identical to `pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/utils.go`. Consider moving to a shared package to eliminate the duplication and prevent future drift between the two copies.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		entries := item.Value()
+		for _, reqKey := range entries.blockKeys {
+			// Speculative entries were added without engineKey mapping, so
+			// evict by RequestKey directly.
+			//nolint:errcheck // best-effort cleanup on TTL expiry
+			index.Evict(context.Background(), reqKey, kvblock.RequestKey, entries.podEntries)
+		}
+	})
+	go cleanCachePeriodically(ctx, cache, ttl)
```
elevran: `ttlcache` already provides expiration cleanup via `cache.Start()`. Replace with `go cache.Start()` (and `cache.Stop()` in a shutdown path). This removes a custom goroutine per plugin instance; the identical fix applies to the legacy scorer's copy of this function.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
vMaroon: ~These can kick off outside a request's lifecycle.~ right, updated.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
vMaroon: I think workarounds for the legacy path to work should happen in the slim scorer, and not here. I'll address in the 3rd PR.

The producer should be the first-class entity in the pipeline.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/utils.go
vMaroon: The slim-scorer will import and initialize components from here for the legacy path.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
vMaroon: I'll rebase this or that one once one merges.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
ahg-g: A couple of questions:
1) how long do we aim to keep the legacy path? are we treating this as a deprecation and so two releases?
2) can the legacy scorer be kept as is? for the new path, we add a new slim scorer, and so the "legacy path" is well encapsulated and used as is, and the new path is encapsulated in a new set of plugins. Perhaps this is already the plan, but just checking.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/utils.go
ahg-g: pls drop the `Pod` terminology in favor of endpoint

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
```diff
@@ -0,0 +1,98 @@
...
+		p.subscribersManager.RemoveSubscriber(ctx, endpointKey)
+		logger.V(logging.DEBUG).Info("Removed KV-events subscriber", "endpoint", endpointKey)
+	}
+	return nil
+}
+
+// EnsureSubscribersForEndpoints is the backwards-compat hook for configs that
+// don't wire the endpoint-notification-source: the legacy scorer wrapper
+// calls this at Produce time. No-op once ExtractEndpoint has fired.
+func (p *Producer) EnsureSubscribersForEndpoints(ctx context.Context, endpoints []scheduling.Endpoint) {
```
ahg-g: I don't recommend we mix the legacy with the new mode. I recommend to keep the legacy path as is and separate even if it has duplicate code and focus on the new path. The legacy will be removed eventually and ideally should not be used unless there is a regression in the new path.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
vMaroon: 1. It's a deprecation, in a couple of releases, the scorer will no longer support the params moved to the producer
2. I've been contemplating dropping the scorer completely, we can hook the `prefix-cache-scorer` to the `precise-prefix-cache-producer` and it should work

ahg-g: That would be ideal, the same scorer but different prefix match producers.

pkg/epp/datalayer/data_graph.go
```diff
@@ -72,12 +72,24 @@ func CreateMissingDataProducers(ctx context.Context, defaultProducerRegistry map
 	logger := log.FromContext(ctx)
 
 	// Collect all keys already produced by existing plugins.
+	// Warn on duplicate producers per key: last-writer-wins on the attribute
```
ahg-g: Probably worth noting that this is totally fine since the key includes the producer name and so different plugins may consume the same data but from different producers.

pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
```diff
@@ -19,10 +19,10 @@ package prefix
 import (
 	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
-	approxprefixconstants "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/constants"
 )
 
-var PrefixCacheMatchInfoDataKey = plugin.NewDataKey("PrefixCacheMatchInfoDataKey", approxprefixconstants.ApproxPrefixCachePluginType)
+// Shared by approx and precise producers. Configure exactly one.
+var PrefixCacheMatchInfoDataKey = plugin.NewDataKey("PrefixCacheMatchInfoDataKey", "prefix-cache-info")
```
ahg-g: I think we need to keep using a true default as the producer name, using the approximate may be the most appropriate as a default since it always works (it doesn't require a tokenizer sidecar)

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
```diff
@@ -94,7 +94,7 @@ func newDataProducer(ctx context.Context, name string, config config, handle plu
 		config:      config,
 		indexerInst: indexer,
 		pluginState: plugin.NewPluginState(ctx),
-		dk:          attrprefix.PrefixCacheMatchInfoDataKey.WithNonEmptyProducerName(name),
+		dk:          attrprefix.PrefixCacheMatchInfoDataKey,
```
ahg-g: This is not right, we need to be able to create more than one producer with different configs for the same key to support tiered kv-cache

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -0,0 +1,270 @@
...
+		return nil, err
+	}
+
+	return &Producer{
+		typedName:          plugin.TypedName{Type: PluginType},
+		kvCacheIndexer:     indexer,
+		kvBlockScorer:      kvBlockScorer,
+		subscribersManager: subscribersManager,
+		kvEventsConfig:     config.KVEventsConfig,
+		dk:                 attrprefix.PrefixCacheMatchInfoDataKey,
```
ahg-g: ditto, in https://github.com/llm-d/llm-d-router/pull/1183 we introduced the producer name as part of the key, I think it is best if we are consistent across producers and avoid using static keys. The prefix and predicted latency scorers already compose a key using a named prefix info producer.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/utils.go
ahg-g: not addressed

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
```diff
@@ -68,7 +69,13 @@ func PrefixCachePluginFactory(name string, rawParameters json.RawMessage, handle
...
 
 // New initializes a new prefix Plugin.
-func New(_ context.Context, name string, producerName string) (*Plugin, error) {
+func New(ctx context.Context, name string, producerName string) (*Plugin, error) {
+	// Back-compat: the historic default was approx; map to the current
+	// neutral default so existing configs keep matching the producer.
+	if producerName == approxconstants.ApproxPrefixCachePluginType {
+		log.FromContext(ctx).Info("prefixMatchInfoProducerName=approx-prefix-cache-producer is deprecated; remove the field (default now matches both approx and precise producers)")
+		producerName = ""
+	}
```
ahg-g: This is a bit a of hack. I think we should support defaulting to any producer for the key as long as there is exactly one producer configured for it. If more than one producer is configured for the key, we error out.

vMaroon:
Updated the direction, the main consumer will be the `prefix-cache-scorer` through:

```yaml
- type: prefix-cache-scorer
  parameters:
    prefixMatchInfoProducerName: precise-prefix-cache-producer
```

Ideally we get to a state where defining one of the producers suffices, but I gave it a try in a previous state of this PR and it is not trivial. Will followup.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -0,0 +1,271 @@
...
+		return nil, err
+	}
+
+	return &Producer{
+		typedName:          plugin.TypedName{Type: PluginType},
+		kvCacheIndexer:     indexer,
+		kvBlockScorer:      kvBlockScorer,
+		subscribersManager: subscribersManager,
+		kvEventsConfig:     config.KVEventsConfig,
+		dk:                 attrprefix.PrefixCacheMatchInfoDataKey.WithNonEmptyProducerName(PluginType),
```
ahg-g: should be plugin name here not type

Issue #1072, opened 2026-05-12
Align request data between EPP plugins and vLLM


sagearc:
This affects routing decisions today, and will cause tokens-in/tokens-out accounting to break as we expand coverage.

### Context

When a request arrives, the EPP produces two representations from the body:
- A lossless `Payload` (`map[string]any`, all fields preserved)
- A typed `ChatCompletionsRequest` (lossy)

The lossy typed struct is what gets passed to the precise-prefix scorer and the tokenizer sidecar, while vLLM receives the lossless `Payload`. Model-specific fields like `reasoning` and others are present in what vLLM processes but absent in what the tokenizer sees.

### Problem

The token IDs produced by the sidecar don't match what vLLM actually processes, so the precise-prefix scorer computes wrong cache keys and misses valid hits. The approximate-prefix scorer is also affected — it hashes the same lossy `Messages` slice, so it can't distinguish requests that differ only in tool-calling fields.

### Proposed Fix

1. Pass the original payload directly to the data plugins (tokenizer, precise-prefix scorer, approximate-prefix scorer) instead of converting through the typed struct
2. Remove both `ChatCompletionsRequest` and `RenderChatRequest` entirely — they are maintenance overhead with no real benefit once (1) is in place
3. Keep only a minimal read of `messages` (content block types and length) and `cache_salt` for multimodal content detection, context-length estimation, and cache key salting — all directly from the raw payload, no typed struct for the entire request needed.

This also gives us native support for all multimodal types vLLM already covers, without any additional work.

cc @nilig @vMaroon @liu-cong @yankay

liu-cong:
I don't see a problem of converting to a typed struct, if the typed struct is indeed needed (which I believe so). The raw payload is already available in the InferenceRequest right? If a plugin needs to access the raw payload, it can already do.

sagearc:
The fix isn't to remove typed structs, it's to scope them to what we actually use. The struct should only expose fields that EPP components genuinely need (content block types for MM detection, raw content for length estimation), not try to be a full OpenAI message mirror. Everything else should come from the raw payload.

liu-cong:
@sagearc Generally agree we can clean up fields that aren't actually used. We still need to parse fields such as image url because we estimate tokens differently than text. There is already an open PR for this.

PR #1115, opened 2026-05-14
fix(tokenizer): forward raw payload to vLLM render endpoint

sagearc:
## Summary

Use the parsed request payload when the vLLM HTTP tokenizer calls `/render`, so tokenization sees the same request shape that vLLM serves.

This preserves backend/model-specific fields such as `reasoning` and multimodal content blocks that are not represented in the local typed OpenAI structs.

## Changes

- Forward `fwkrh.PayloadMap` for vLLM HTTP completions and chat-completions tokenization.
- Inject the configured tokenizer `model` into a cloned payload.
- Keep existing tokenizer interfaces and non-vLLM/UDS behavior unchanged.
- Add tests proving arbitrary payload fields and multimodal content blocks reach the mock vLLM render endpoint.

## Multimodal impact

Multimodal chat content blocks such as `input_audio` and `video_url` are now forwarded unchanged to `/v1/chat/completions/render`. This lets the vLLM render endpoint return `token_ids` plus multimodal hash/placeholder metadata from the same request shape that vLLM will serve.

This does not move media encoding or embedding computation into the router; those remain vLLM inference-side behavior.

## Scope

This is a narrow vLLM HTTP tokenizer-path alignment for #1072. It does not implement EPP-native multimodal tokenization and does not complete the broader API-agnostic `InferenceRequestBody` standardization tracked in #1364.

## Validation

- `go test ./pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/...`

Partially handles #1072.


pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -209,11 +209,27 @@ func (p *Plugin) tokenize(ctx context.Context, request *scheduling.InferenceRequ
...
+		renderReq := ChatCompletionsToRenderChatRequest(request.Body.ChatCompletions)
+		if renderer, ok := p.tokenizer.(*vllmHTTPRenderer); ok {
+			payload, ok := request.Body.Payload.(fwkrh.PayloadMap)
+			if !ok {
+				return nil, errors.New("vLLM HTTP tokenizer requires request PayloadMap")
+			}
+			tokenIDs, mmFeatures, err = renderer.renderChatPayload(ctx, payload, renderReq)
+		} else {
+			tokenIDs, mmFeatures, err = p.tokenizer.RenderChat(ctx, renderReq)
+		}
```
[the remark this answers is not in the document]
sagearc: fixed

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
```diff
@@ -136,8 +154,20 @@ func (r *vllmHTTPRenderer) Render(ctx context.Context, prompt string) ([]uint32,
 // kvcache map shape expected by the upstream interface.
 func (r *vllmHTTPRenderer) RenderChat(ctx context.Context, req *tokenizerTypes.RenderChatRequest) ([]uint32, *tokenization.MultiModalFeatures, error) {
 	body := buildChatRenderRequest(r.modelName, req)
+
+	return r.postChatRender(ctx, body, r.chatTimeout(req))
+}
+
+func (r *vllmHTTPRenderer) renderChatPayload(ctx context.Context, payload fwkrh.PayloadMap, req *tokenizerTypes.RenderChatRequest) ([]uint32, *tokenization.MultiModalFeatures, error) {
+	body := maps.Clone(payload)
+	body["model"] = r.modelName
```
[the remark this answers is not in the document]
sagearc: Not changed: helpers are only called with non-nil PayloadMap.

sagearc:
A temporary local `httptest` benchmark was run against `main` to measure renderer payload-forwarding overhead:

| Path | `main` | this branch | delta |
| --- | ---: | ---: | ---: |
| completions | ~126.9 us/op, 8.97 KB/op, 115 allocs/op | ~129.3 us/op, 9.53 KB/op, 124 allocs/op | ~+2%, +9 allocs |
| chat completions | ~130.7 us/op, 9.13 KB/op, 116 allocs/op | ~133.3 us/op, 9.84 KB/op, 133 allocs/op | ~+2%, +17 allocs |

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -205,15 +205,26 @@ func (p *Plugin) tokenize(ctx context.Context, request *scheduling.InferenceRequ
 	var tokenIDs []uint32
 	var mmFeatures *tokenization.MultiModalFeatures
 	var err error
+	renderer, isVLLMHTTP := p.tokenizer.(*vllmHTTPRenderer)
```
liu-cong: Can we just encapsulate this logic to the vllm_http.go? And we can update the Render method signature to take the InferenceRequest if needed.

sagearc: Done, updated render methods to accept `RequestPayload` directly, routing logic is now contained in each tokenizer.

sagearc:
@liu-cong routing logic is now contained in each tokenizer as suggested in the review, is this good to go?

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http_test.go
```diff
@@ -77,9 +77,37 @@ func TestVLLMHTTPRenderer_Render(t *testing.T) {
...
+	var sent map[string]any
+	require.NoError(t, json.Unmarshal(cap.completions, &sent))
+	assert.Equal(t, "kept", sent["dummy_field"])
+	assert.Equal(t, testHTTPModel, sent["model"])
+}
+
+// TestVLLMHTTPRenderer_RenderChat_Multimodal covers the chat endpoint: the raw
+// payload is forwarded directly and multimodal features are converted from wire
+// format to kvcache map shape.
 func TestVLLMHTTPRenderer_RenderChat_Multimodal(t *testing.T) {
```
hexfusion, line 111: nit: please add tests asserting input_audio and video_url content blocks are forwarded verbatim

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
```diff
@@ -131,115 +144,39 @@ func (r *vllmHTTPRenderer) Render(ctx context.Context, prompt string) ([]uint32,
...
 		return nil, nil, err
 	}
 	return resp.TokenIDs, toKVCacheMM(resp.Features), nil
 }
 
-func (r *vllmHTTPRenderer) chatTimeout(req *tokenizerTypes.RenderChatRequest) time.Duration {
-	for _, msg := range req.Conversation {
+func (r *vllmHTTPRenderer) chatTimeout(payload fwkrh.PayloadMap) time.Duration {
+	var chat fwkrh.ChatCompletionsRequest
+	if data, err := json.Marshal(payload); err == nil {
```
hexfusion: nit: chatTimeout does json.Marshal and json.Unmarshal per chat request just to check whether any message has structured content. since this is already a map can we get the same signal directly? this would save an extra allocation.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
```diff
@@ -117,10 +120,20 @@ func parseHTTPDuration(s string, def time.Duration) (time.Duration, error) {
...
-	body := completionsRenderRequest{Model: r.modelName, Prompt: prompt}
+// Render calls /v1/completions/render. The PayloadMap is forwarded verbatim
+// (preserving backend-specific fields such as reasoning) with the configured
+// model name stamped in. Char offsets are not returned by vLLM's render endpoint.
+func (r *vllmHTTPRenderer) Render(ctx context.Context, payload fwkrh.RequestPayload) ([]uint32, []tokenizerTypes.Offset, error) {
+	pm, ok := payload.AsMap()
+	if !ok {
+		return nil, nil, errors.New("vLLM HTTP tokenizer requires a parsed PayloadMap")
+	}
+	body := maps.Clone(pm)
```
hexfusion, line 132: nit: lets add a comment that this is a shallow copy and why that is ok?

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/uds.go
```diff
@@ -38,10 +43,35 @@ func newUDSTokenizer(ctx context.Context, cfg *tokenization.UdsTokenizerConfig,
 	return &udsTokenizerAdapter{t: uds}, nil
 }
 
-func (a *udsTokenizerAdapter) Render(_ context.Context, prompt string) ([]uint32, []tokenizerTypes.Offset, error) {
+func (a *udsTokenizerAdapter) Render(_ context.Context, payload fwkrh.RequestPayload) ([]uint32, []tokenizerTypes.Offset, error) {
+	pm, ok := payload.AsMap()
+	if !ok {
+		return nil, nil, errors.New("UDS tokenizer requires a parsed PayloadMap")
+	}
+	prompt, _ := pm["prompt"].(string)
```
hexfusion: nit: this is a behavior change, before this change prompt was ensured to be a string by the type system. so dropping the !ok check could lead to a possible regression.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
hexfusion:
```suggestion
func (r *vllmHTTPRenderer) chatTimeout(payload fwkrh.PayloadMap) time.Duration {
    msgs, ok := payload["messages"].([]any)
    if !ok {
        return r.timeout
    }   
    for _, m := range msgs {
        msg, ok := m.(map[string]any)
        if !ok {
            continue
        }   
        // Array-shaped content is multimodal; use the longer timeout.
        if arr, ok := msg["content"].([]any); ok && len(arr) > 0 { 
            return r.mmTimeout
        }   
    }   
    return r.timeout                                                                                                                                                                                
}
```

vMaroon:
cc @zetxqx - for rendering the best path is passing the raw body, FYI. Thoughts?

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
sagearc: done

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/uds.go
sagearc: added the explicit error. UDS completions only accepts a string prompt (never truly supported arrays)

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http_test.go
sagearc: done

sagearc:
Thanks @hexfusion 
Applied the suggested changes, added you as coauthor

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/uds.go
```diff
@@ -38,10 +43,38 @@ func newUDSTokenizer(ctx context.Context, cfg *tokenization.UdsTokenizerConfig,
...
+	}
 	return a.t.RenderChat(req)
 }
+
+func renderChatRequestFromPayload(pm fwkrh.PayloadMap) (*tokenizerTypes.RenderChatRequest, error) {
+	data, err := json.Marshal(pm)
+	if err != nil {
+		return nil, fmt.Errorf("marshal payload: %w", err)
+	}
+	var chat fwkrh.ChatCompletionsRequest
```
zetxqx, line 75: Just note here I'm planning to unexported  `ChatCompletionsRequest` from fwkrh in https://github.com/llm-d/llm-d-router/issues/1364#issuecomment-4560679710. the payload body will be still passed. I think we can either

1. define the `ChatCompletionsRequest` here, I think it's unmarshaled the just for struct conversion for: ChatCompletionsToRenderChatRequest. 
or
2.I still keep the `ChatCompletionsRequest` exported for this limited usage.

zetxqx:
> cc @zetxqx - for rendering the best path is passing the raw body, FYI. Thoughts?

Thank you @vMaroon @sagearc , I think this is aligned with https://github.com/llm-d/llm-d-router/issues/1364#issuecomment-4560679710 . Mainly for stage 6.  I see we still need to unmarshal to ChatCompletion to do ChatCompletionsToRenderChatRequest in udsTokenizer. But it should be easy to define a struct inside udsTokenizer just for struct conversion while I'm working on stage 8.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -39,8 +39,8 @@ import (
 )
 
 type tokenizer interface {
-	Render(ctx context.Context, prompt string) ([]uint32, []tokenizerTypes.Offset, error)
-	RenderChat(ctx context.Context, req *tokenizerTypes.RenderChatRequest) ([]uint32, *tokenization.MultiModalFeatures, error)
+	Render(ctx context.Context, payload fwkrh.RequestPayload) ([]uint32, []tokenizerTypes.Offset, error)
```
vMaroon, line 42: If implementers are always going to use the map representation, then the latter should be passed instead.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -213,11 +213,10 @@ func (p *Plugin) tokenize(ctx context.Context, request *scheduling.InferenceRequ
 	switch {
 	case request.Body.Completions != nil:
```
ahg-g: What about `/v1/conversations`, `/v1/messages` and `/v1/responses`? what is the path for these?

Also, note that the grpc protocols (vllmgrpc and vertexai), populate `Completions` or `ChatCompletions`, this change breaks them.

Should we have a fallback to the older approach if the payload was not originally a Completions or ChatCompletions type?

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/uds.go
sagearc: Thanks for raising this, for this PR I'd keep the UDS conversion as compatibility-only and not make it part of the vllm http renderer, as will get deprecated soon either way

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
sagearc: Partially agree. I kept the shared interface wider because not all existing parser paths are map-backed, and we may want to support tokenization for them without converting them to a json map. Is this covered by #1445?

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -41,27 +41,32 @@ const ModalityImage Modality = "image"
 type RequestPayload interface {
 	isRequestPayload()
 	IsParsed() bool
+	// AsMap returns the parsed JSON map
+	AsMap() (PayloadMap, bool)
 }
 
 // PayloadMap represents a JSON request body unmarshaled into a map.
 type PayloadMap map[string]any
```
elevran, line 49: two suggestions:
- unless you need the entire payload parsed, `map[string]json.RawMessage` would be considerably cheaper to construct and GC (avoid the full parsing descent into sub-fields)
- nit: **if** the order of keys matters, consider using an ordered map struct instead.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -213,11 +213,10 @@ func (p *Plugin) tokenize(ctx context.Context, request *scheduling.InferenceRequ
...
 		traceLogger.Info("Calling Render for completions", "prompt", request.Body.Completions.Prompt)
-		tokenIDs, _, err = p.tokenizer.Render(ctx, request.Body.Completions.Prompt.Raw)
+		tokenIDs, _, err = p.tokenizer.Render(ctx, request.Body.Payload)
 	case request.Body.ChatCompletions != nil:
-		renderReq := ChatCompletionsToRenderChatRequest(request.Body.ChatCompletions)
 		traceLogger.Info("Calling RenderChat for chat completions", "messageCount", len(request.Body.ChatCompletions.Messages))
-		tokenIDs, mmFeatures, err = p.tokenizer.RenderChat(ctx, renderReq)
+		tokenIDs, mmFeatures, err = p.tokenizer.RenderChat(ctx, request.Body.Payload)
```
elevran: Q: `request.Body.Payload` is passed without a nil check. If Payload is the zero value of the `RequestPayload` interface (i.e., nil), calling  `.AsMap()` on it in both `vllmHTTPRenderer` and the tokenizer will panic. Before this change, the tokenizer received typed values derived from already-validated fields and not the raw payload directly. Add a nil check and error here or handle in the callees.

elevran (commented):
two suggestions and a question left.
From my point of view this can be merged once @ahg-g's comment is addressed.
/lgtm

hexfusion:
@sagearc  code freeze is on the 14th and I would really like to see this make it :+1:

sagearc:
> @sagearc code freeze is on the 14th and I would really like to see this make it 👍

@hexfusion This is planned for next release. Sorry for the delay, was digging deeper into @ahg-g review.
cc @vMaroon

hexfusion:
@sagearc please resolve lint failure 
```sh
go fmt ./...
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
sagearc: Thanks for raising this!

> Also, note that the grpc protocols (vllmgrpc and vertexai), populate `Completions` or `ChatCompletions`, this change breaks them.

Fixed, falls back to constructing the render oai request body from the typed structs, same wire body as before this PR.

> What about `/v1/conversations`, `/v1/messages` and `/v1/responses`? what is the path for these?

pre-existing gap, they'll keep hitting the default error case in `renderBackend.produce()` until proper support is added.

ahg-g:
/lgtm

I will leave it to @vMaroon to approve

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
sagearc: Fixed, `completionsPayload()` and `chatPayload()` in `backend.go` both guard with body.Payload != nil before calling .AsMap() and fall through to a typed-struct fallback when absent or not a map.

pkg/epp/framework/interface/requesthandling/types.go
sagearc: Appreciate the insights!

> if the order of keys matters, consider using an ordered map struct instead.

order is not needed here, the only mutation is stamping model at the top level in `Render`/`RenderChat` before marshaling.

> unless you need the entire payload parsed, `map[string]json.RawMessage` would be considerably cheaper to construct and GC (avoid the full parsing descent into sub-fields)

Good point. Changing `PayloadMap`'s type breaks ~10 sites across parsers, tokenizer, and director (casts from `map[string]any`, type assertions on values, and direct writes). bounded but a standalone refactor, happy to track separately.

vMaroon (approved):
great work

vMaroon:
We need to support `/v1/messages` and `/v1/responses` (stateless) as well.

PR #1121, opened 2026-05-14
[3/3] feat(precise-prefix-cache): scorer deprecation + maximal implicit use of the new producer

vMaroon:
## Summary

Replace the all-in-one `precise-prefix-cache-scorer` with a thin wrapper that composes the `precise-prefix-cache-producer` (landed in #1112) and the generic `prefix-cache-scorer` behind the same plugin type. Existing YAML keeps working unchanged; behavior tracks the new path because the two paths are now the same code.

## Factory modes

- **Self-host (default):** instantiate an internal producer + scorer pair from the legacy parameters; expose Scorer + DataProducer + PreRequest + EndpointExtractor by delegation.
- **Defer-to-existing:** when a `precise-prefix-cache-producer` is already in the handle, return a `prefix-cache-scorer` pointed at it and ignore the legacy parameters. Multiple precise producers raise an error since the wrapper cannot disambiguate.

## Legacy `tokenizersPoolConfig` preserved

The historical `indexerConfig.tokenizersPoolConfig` path is kept alive through an isolated `legacyProducer` that owns its own `tokenization.Pool` and pre-tokenizes prompts before delegating to the embedded producer. The pool config is stripped from the inner indexer to avoid double-construction. No upstream `token-producer` is required in this mode (the wrapper's `Consumes` drops `TokenizedPrompt` accordingly).

## Behavior preserved

- OTel span `llm_d.epp.scorer.prefix_cache` on `Score` with the original attribute schema.
- PreRequest nil-guards on target/prefill endpoint metadata.
- Speculative TTL eviction bound to the callback's ctx.
- `ttlcache.Start()` in place of `cleanCachePeriodically`.

## Lost behavior

- The legacy in-`Score` subscriber discovery path (for YAMLs without `endpoint-notification-source`) is gone. Operators upgrading must wire `endpoint-notification-source` or use global socket mode (`KVEventsConfig.ZMQEndpoint`).
  - This is OK because paths without endpoint-notification-source were kind of hidden and experimental. Active-active HA went clearly visible in llm-d v0.7.0 with the notification source configured.

## Test coverage

- Heavy scorer's `extractor_test.go` and `tokenized_test.go` are removed; equivalent coverage lives in the producer's package.
- Heavy scorer's UDS integration tests (`uds_test.go` + `testdata/test-model/*`) are removed; replaced by a stub-tokenizer end-to-end test in this package that verifies the pool → tokens → block keys → endpoint attribute pipeline without a real UDS server.

## Test plan

- [x] `go test ./pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/...`
- [x] `go build ./...`
- [ ] Smoke-test legacy YAML (`precise-prefix-cache-scorer` only, no token-producer) against a vLLM pool

**Release note**
```release-note
`precise-prefix-cache-scorer` is now a thin compatibility wrapper around `precise-prefix-cache-producer` and the `prefix-cache-scorer`. Existing configurations continue to work. Deployments without `endpoint-notification-source` wired must add it (or use global socket mode); the legacy in-Score subscriber discovery path is removed. The plugin is deprecated; configure `precise-prefix-cache-producer` + `prefix-cache-scorer` with `prefixMatchInfoProducerName: precise-prefix-cache-producer` directly for new deployments.
```


vMaroon:
@elevran @ahg-g PTAL

Considering there are open PRs that touch this code, I chose to not keep it as-is and to defer to the producer as much as possible. The scorer itself initializes a producer + a `prefix-cache-scorer` when possible. 

The legacy path of internal tokenization becomes a producer wrapper that handles text -> tokens and calls the new producer.

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -1,858 +1,236 @@
...
-	for _, ep := range endpoints {
-		md := ep.GetMetadata()
-		if md == nil {
+// findExistingPreciseProducer returns the precise-prefix-cache-producer
+// registered in the handle, or nil if none is configured. Errors when more
+// than one is found, since the legacy plugin has no way to choose between
+// multiple producers.
+func findExistingPreciseProducer(handle plugin.Handle) (*preciseproducer.Producer, error) {
+	var found *preciseproducer.Producer
+	for _, p := range handle.GetAllPlugins() {
```
ahg-g, line 144: What if the producer isn't instantiated yet because it is later in the plugins list? If this is true, then I think GetAllPlugins is dangerous and should probably return an error until the full instantiation is done.

vMaroon: This logic attempts to not initialize an internal producer in configurations that use the producer and the precise scorer. Technically such deployments should not exist as they were never and are still not recommended anywhere.

In the worst case here, we would have 2 valid producers, duplicating resource use and logic, but still working correctly. This would be purely an optimization for if a user configures such a setup.

ahg-g: it is fine here since this is getting deprecated, but the behavior of GetAllPlugins is I think a bug waiting to happen (not related to this PR).

vMaroon: @ahg-g any concrete action you are requesting here? I can followup on this in a different PR.

ahg-g: all good, I think this is sufficient.

PR #1126, opened 2026-05-14
Reject unknown approximate prefix config fields

### LINKED ISSUE #1068: EPP should signal invalid plugin config fields via error/warning
Add a flag `--ignore-invalid-config-fields` to EPP,
* If `true`, EPP should continue to function, but print a `warning` log
* If `false`, EPP should exit with an error 

I propose defaulting this to `false` - this is the safest. 

**Why is this needed**:
If provided with invalid plugin config (e.g., a non-existing field, or a deprecated field), EPP ignores those fields and starts normally. 
This is dangerous: users may have a typo, or use a deprecated field, without knowing that the intended config is not taking effect.

A concrete example is the recent change of the approximate prefix cache data producer plugin. In llm-d tired prefix cache guides, we still have the old configuration, and there is no clear signal of detecting that.

mzl2233:
This updates the approximate prefix cache data producer to decode its plugin parameters with unknown-field rejection, so misspelled or removed configuration keys fail during plugin initialization instead of being silently ignored. I added a focused regression test that exercises an invalid approximate prefix configuration field and verified the targeted package with `go test ./pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix`.

Closes #1068

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
@@ -17,6 +17,7 @@ limitations under the License.
 package approximateprefix
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


elevran:
@mzl2233 would this still be needed after #1134 is merged? The other PR performs strict decoding which would reject unknown fields across the approximate prefix config as well.

ahg-g:
@mzl2233 pls feel free to reopen in case you think https://github.com/llm-d/llm-d-router/pull/1134 doesn't address the problem

PR #1128, opened 2026-05-14
Third step in refactoring the FlowControl component

shmuelk:
**What type of PR is this?**
/kind cleanup

This PR is the third step in a process of simplifying the Flow Control component's code base.

In particular this PR:
1. Removes the RegistryShard interface and the Shard struct.
2. The managed queues and the PriorityBands are now managed directly by the FlowRegistry
3. The initial number of shards was removed from the FlowControl configuration
4. Some simplifications were done to the code base
5. Test that were no longer appropriate were removed

**Notes** to the reviewers:
1. To keep the files registry.go and registry_test.go in the directory pkg/epp/flowcontrol/registry reasonably sized, the parts of the RegistryShard that were merged into the FlowRegistry the files containing that code shard.go and shard_test.go were renamed to registry_helper.go and registry_helper_test.go respectively.
2. The word "shard" has not yet been removed completely from the code. This PR is large enough as it is. This will be cleaned up in a follow on PR.
3. There is still some lazy initialization of some things that will be cleaned up in a follow on PR.
4. As @LukeAVanDrie suggested the documentation will be updated at the end of the process.

Refs: #889

```release-note
NONE
```


shmuelk:
cc: @LukeAVanDrie

pkg/epp/flowcontrol/contracts/registry.go
```diff
@@ -78,36 +75,6 @@ type FlowRegistryDataPlane interface {
...
-// A shard acts as an independent, parallel execution unit, allowing the system's dispatch logic to scale horizontally.
-//
-// # Conformance: Implementations MUST be goroutine-safe.
-type RegistryShard interface {
-	// ID returns a unique identifier for this shard, which must remain stable for the shard's lifetime.
-	ID() string
-
-	// IsActive returns true if the shard should accept new requests for enqueueing. A false value indicates the shard is
-	// being gracefully drained and should not be given new work.
-	IsActive() bool
```
hexfusion, line 110: 👍

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -253,14 +259,17 @@ func (fr *FlowRegistry) ensureFlowInfrastructure(key flowcontrol.FlowKey) error
 	// 2. Synchronize shards.
 	// Acquire Read Lock to iterate the shard topology safely.
```
hexfusion, line 260: there are no more shards right do these docs need updtated?

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -253,14 +259,17 @@ func (fr *FlowRegistry) ensureFlowInfrastructure(key flowcontrol.FlowKey) error
...
 	fr.mu.RLock()
```
hexfusion, line 261: if we are holding the read lock on FlowRegistry  just for buildFlowComponents why not move it internal to the method?

Jwrede (commented):
Nice cleanup, one comment:

**Nil pointer risk in new code in `propagateStatsDelta`** (`registry.go`, new lines in the `propagateStatsDelta` function):

```go
rawBand, _ := fr.priorityBands.Load(priority)
band := rawBand.(*priorityBand)
```

The `ok` return from `Load` is ignored. If a priority band doesn't exist (e.g., during a race with GC eviction), `rawBand` is `nil` and the cast panics. The pre-existing `perPriorityBandStats.Load` a few lines below has the same pattern, but the new `priorityBands.Load` duplicates the risk for a second map.

Same pattern appears in `synchronizeFlow` (`registry_helpers.go`):
```go
val, _ := fr.priorityBands.Load(key.Priority)
band := val.(*priorityBand)
```

Might be worth a guard or at least a comment explaining the invariant that guarantees the band exists at call time

pkg/epp/flowcontrol/registry/registry.go
shmuelk: As mentioned in the PR description, the docs will updated in a followup PR. This is already a very large PR.

pkg/epp/flowcontrol/registry/registry.go
LukeAVanDrie: +1

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -477,10 +469,16 @@ func (fr *FlowRegistry) buildFlowComponents(key flowcontrol.FlowKey) (*flowCompo
...
-	stats.len.Add(lenDelta)
-	stats.byteSize.Add(byteSizeDelta)
-	fr.totalLen.Add(lenDelta)
-	fr.totalByteSize.Add(byteSizeDelta)
+	if rawBand, ok := fr.priorityBands.Load(priority); rawBand != nil && ok {
+		band := rawBand.(*priorityBand)
+		band.len.Add(lenDelta)
+		band.byteSize.Add(byteSizeDelta)
+
+		val, _ := fr.perPriorityBandStats.Load(priority)
```
LukeAVanDrie: The `ok` return from `Load` is ignored and the result is cast directly. If a priority band is garbage collected (`cleanupPriorityBandResources`) concurrently with stats propagation from a queue, `val` will be `nil`, causing a cast panic.

This was a preexisting issue and is not caused by your refactoring efforts. We can easily fix it with:

```
if val, ok := fr.perPriorityBandStats.Load(priority); ok && val != nil {
    stats := val.(*bandStats)
    stats.len.Add(lenDelta)
    stats.byteSize.Add(byteSizeDelta)
}
```

pkg/epp/flowcontrol/registry/registry_helpers.go
```diff
@@ -0,0 +1,259 @@
...
+	a.registry.mu.RLock()
+	ids := make([]string, 0, len(a.band.queues))
+	for id := range a.band.queues {
+		ids = append(ids, id)
+	}
+	a.registry.mu.RUnlock()
+
+	flowKeys := make([]flowcontrol.FlowKey, len(ids))
+	for i, id := range ids {
+		flowKeys[i] = flowcontrol.FlowKey{ID: id, Priority: a.Priority()}
```
LukeAVanDrie: `a.Priority()` acquires and releases `fr.mu.RLock()`, this results in N redundant global read lock acquisitions for a band containing N flows. Since priority is immutable, let's move this out of the loop:

```
priority := a.band.config.Priority // Or a.Priority() once
for i, id := range ids { ... }
```

pkg/epp/flowcontrol/registry/registry.go
LukeAVanDrie, line 334: This map isn't concurrency safe (JIT registration and GC act on it). We should wrap `Stats` in a read lock.

pkg/epp/flowcontrol/registry/registry_helpers.go
```diff
@@ -0,0 +1,259 @@
...
+	band.queues[key.ID] = mq
+}
+
+// deleteFlow removes a queue instance.
+// Must be called with the registry write lock held
+func (fr *FlowRegistry) deleteFlow(key flowcontrol.FlowKey) {
+	fr.logger.V(logging.DEFAULT).Info("Deleting queue instance.", "flowKey", key)
+	if val, ok := fr.priorityBands.Load(key.Priority); ok {
+		band := val.(*priorityBand)
+		delete(band.queues, key.ID)
```
LukeAVanDrie, line 188: When a request in a queue is asynchronously finalized (e.g., due to client stream cancellation or context timeout), `awaitFinalization()` in `controller.go` immediately wakes up and returns from `WithConnection`. This unpins the flow lease, dropping the lease count `state.leaseCount` to `0`. 
At this point, the flow is considered "idle" by `collectLeasedResources` and is scheduled for logical GC. 

However, the physical `FlowItem` still sits in the queue as a "zombie" waiting for the processor's periodic cleanup sweep (`sweepFinalizedItems()`). Because the leasing/GC logic (`leasing.go`) is completely decoupled from the physical queue length, `gcFlows()` can run and trigger `deleteFlow(key)` before the processor sweeps the zombie items.

Before deleting the queue in `deleteFlow`, we must verify if there are any remaining (unswept) items in the queue and, if so, deduct their metrics from the band and registry totals:

```
func (fr *FlowRegistry) deleteFlow(key flowcontrol.FlowKey) {
	fr.logger.V(logging.DEFAULT).Info("Deleting queue instance.", "flowKey", key)
	if val, ok := fr.priorityBands.Load(key.Priority); ok {
		band := val.(*priorityBand)
		if mq, ok := band.queues[key.ID]; ok && mq != nil {
			// Safe-guard: Deduct any unswept capacity before destroying the queue
			if mqLen := int64(mq.Len()); mqLen > 0 {
				fr.logger.V(logging.WARN).Info("Deregistering non-empty queue during GC, flushing stats", 
					"flowKey", key, "unsweptCount", mqLen)
				fr.propagateStatsDelta(key.Priority, -mqLen, -int64(mq.ByteSize()))
			}
		}
		delete(band.queues, key.ID)
	}
}
```

This is another preexisting issue. Found a few of these now that I'm looking at this code with fresh eyes and under a simpler concurrency / locking model.

In practice this issue is nearly impossible to trigger given the defaults for idleness window (5m) and our processor sweep interval (1s), so maybe we can defer.

LukeAVanDrie (approved):
No blocking issues introduced by this PR. I just uncovered some preexisting ones now that the locking and concurrency model is simpler and I am looking at the implementation with fresh eyes. We can address those here or create an issue to track them. Similarly, we will need to do this for stale docs.

This is a very nice change; thanks @shmuelk.

pkg/epp/flowcontrol/registry/registry.go
shmuelk: Done

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -334,11 +341,6 @@ func (fr *FlowRegistry) Stats() contracts.AggregateStats {
 	return stats
 }
```
elevran, line 345: Looks like a data race in `Stats()` above: `bandCfg := fr.config.PriorityBands[priority]` is read with no lock, while `ensurePriorityBand` and `cleanupPriorityBandResources` write to that same map under `fr.mu`. Either take `mu.RLock()` here, or move the per-band capacities onto `*bandStats` so the read path is genuinely lock-free as the doc claims.

pkg/epp/flowcontrol/registry/registry_helpers.go
```diff
@@ -0,0 +1,273 @@
...
+	// The priority is implicit from the parent priorityBand.
+	queues map[string]*managedQueue
+
+	// priorityBandAccessor is a preallocated flowcontrol.PriorityBandAccessor for this priorityBand
+	priorityBandAccessor *priorityBandAccessor
+
+	// --- Concurrent-Safe State (Atomics) ---
+
+	// Band-level statistics, updated via lock-free propagation from child queues.
+	byteSize atomic.Int64
```
elevran: `priorityBand.len` / `priorityBand.byteSize` are written in `propagateStatsDelta` but never read anywhere: `Stats()` reads from `bandStats` instead. Looks like leftover duplication from the shard merge; safe to drop along with the corresponding `band.len.Add` / `band.byteSize.Add` calls.

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -383,28 +358,24 @@ func (fc *FlowController) createRequestContext(
 
 // candidate holds the information needed to evaluate a shard as a potential target for a request.
 type candidate struct {
-	processor shardProcessor
-	shardID   string
+	processor processor
 	byteSize  uint64
```
elevran: `candidate.byteSize` is set here but never read. There's only one worker now, so there's nothing to rank against. The whole `candidate` struct collapses to `processor`, and `selectDistributionCandidate` could likely fold into `getOrStartWorker`.

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -383,28 +358,24 @@ func (fc *FlowController) createRequestContext(
...
-			"flowKey", conn.FlowKey(), "shardID", shard.ID())
+			"Invariant violation. Failed to get ManagedQueue for a leased flow.",
+			"flowKey", conn.FlowKey())
 		return nil, fmt.Errorf("%w for flow %s", errNoShards, conn.FlowKey())
 	}
-	return &candidate{worker.processor, shard.ID(), mq.FlowQueueAccessor().ByteSize()}, nil
+	return &candidate{worker.processor, mq.FlowQueueAccessor().ByteSize()}, nil
 }
 
 // distributeRequest implements a flow-aware, two-phase "Join-Shortest-Queue-by-Bytes" (JSQ-Bytes) distribution strategy
```
elevran, line 357: The "two-phase JSQ-Bytes ranked candidates" docstring describes the old multi-shard algorithm. After the refactor it's just "Submit; on busy, SubmitOrBlock to the single worker." Worth rewriting now since this comment actively misleads readers about what the function does.

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -138,14 +146,13 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
...
+		fr.initPriorityBand(bandConfig)
 	}
 
-	if err := fr.createShard(); err != nil {
-		return nil, fmt.Errorf("failed to initialize shard: %w", err)
-	}
-	fr.logger.V(logging.DEFAULT).Info("FlowRegistry initialized successfully")
+	fr.logger.V(logging.DEFAULT).Info("FlowRegistry initialized successfully",
+		"orderedPriorities", fr.orderedPriorityLevels)
 	return fr, nil
```
elevran: `NewFlowRegistry` returns `(*FlowRegistry, error)` but never produces a non-nil error — every `return` is `fr, nil`. Either drop the error return, or wire up the validation that should fail here.

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -477,10 +469,17 @@ func (fr *FlowRegistry) buildFlowComponents(key flowcontrol.FlowKey) (*flowCompo
 
 // propagateStatsDelta is the top-level, lock-free aggregator for all statistics.
 func (fr *FlowRegistry) propagateStatsDelta(priority int, lenDelta, byteSizeDelta int64) {
-	val, _ := fr.perPriorityBandStats.Load(priority)
-	stats := val.(*bandStats)
-	stats.len.Add(lenDelta)
-	stats.byteSize.Add(byteSizeDelta)
-	fr.totalLen.Add(lenDelta)
-	fr.totalByteSize.Add(byteSizeDelta)
+	if rawBand, ok := fr.priorityBands.Load(priority); rawBand != nil && ok {
```
elevran: `rawBand != nil && ok` here and `ok && val != nil` on line 477. Calls to `sync.Map.Load` already returns `ok=false` when there's no entry, so the nil checks are redundant.

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -138,14 +146,13 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
 		fr.clock = &clock.RealClock{}
 	}
 
-	for prio := range cfg.PriorityBands {
+	for prio, bandConfig := range cfg.PriorityBands {
 		fr.perPriorityBandStats.Store(prio, &bandStats{})
```
elevran: `Store` here vs `LoadOrStore` at line 294 for the same map. Please be consistnt (`LoadOrStore` works in both places).

elevran (commented):
Nice consolidation - thanks @shmuelk!
Collapsing `RegistryShard` into `FlowRegistry` reads cleanly, and the leasing/JIT path is easier to follow without the shard topology.

Some nits for a follow up PR as to not expand this further:
- `managedWorker` (controller.go:73-76) is now a one-field wrapper around processor and could be eliminated.
- `flowComponents` (registry.go:447-450) wraps two values used in exactly one call site, could just as well be a (policy, queue) return.
- `RegistryOption` functional-options pattern carries only one option (test-only...) and can likely be removed (heavy machinery for one entry).
- the processor interface in `controller.go` has a single concrete implenebtation and is a possible candidate for inlining (though can keep for its test seams).

Issue #1122, opened 2026-05-14
`make presubmit` fails locally on rootless container runtimes (`vulncheck` target)


gkneighb:
## What happened

On a clean clone, `make presubmit` on a feature branch fails at the `vulncheck` step:

```
==== Running govulncheck ====
podman run --rm --userns=keep-id -v $(pwd):/app:Z -w /app \
    -v llm-d-gomodcache:/go/pkg/mod -v llm-d-gobuildcache:/go/cache \
    ghcr.io/llm-d/llm-d-builder:dev \
    sh -c 'go install golang.org/x/vuln/cmd/govulncheck@v1.3.0 && govulncheck ./...'
go: downloading golang.org/x/vuln v1.3.0
golang.org/x/vuln/cmd/govulncheck: go install golang.org/x/vuln/cmd/govulncheck:
    copying /tmp/go-build…/b001/exe/a.out: open /usr/local/bin/govulncheck: permission denied
make: *** [vulncheck] Error 1
```

All other presubmit steps pass cleanly (`go fmt`, `golangci-lint`, `typos`).

## Root cause

The `vulncheck` target does `go install` at runtime inside the builder container:

```makefile
vulncheck: image-build-builder
	$(BUILDER_RUN) 'go install golang.org/x/vuln/cmd/govulncheck@v1.3.0 && govulncheck ./...'
```

Two facts collide:
- The base image (`quay.io/projectquay/golang:1.25`) bakes in `GOBIN=/usr/local/bin`, which is root-owned.
- `BUILDER_RUN_FLAGS` runs the container as a non-root user in two common dev setups:
  - **Rootless podman**: `--userns=keep-id` (Makefile L70–73)
  - **Docker**: `-u $(id -u):$(id -g)` (Makefile L78)

So `go install` cannot write to `GOBIN`. The error only surfaces locally; CI sidesteps the Makefile target — #1048 runs `govulncheck` directly on the GHA runner (which is root) — so this has stayed invisible upstream.

## Suggested fix

Pre-install `govulncheck` in `Dockerfile.builder`, matching the pattern used for every other tool (setup-envtest, golangci-lint, typos, kubectl, etc.):

```dockerfile
ARG GOVULNCHECK_VERSION=v1.3.0
RUN GOBIN=/usr/local/bin go install golang.org/x/vuln/cmd/govulncheck@${GOVULNCHECK_VERSION}
```

And simplify the Makefile target:

```diff
[...]

PR #1134, opened 2026-05-14
feat(epp/plugins): strict plugin-config parsing via *json.Decoder

### LINKED ISSUE #1068: EPP should signal invalid plugin config fields via error/warning

gkneighb:
**What type of PR is this?**
/kind feature

Implements #1068. EPP previously accepted plugin configs with unknown fields, silently dropping typos and stale-doc references — notably the recent `BlockSize` → `BlockSizeTokens` rename in the approximate prefix cache plugin, which left tiered-prefix-cache guides pointing at a non-functional field.

Changes `FactoryFunc` so the framework owns strict decoding:

```go
// Before:
type FactoryFunc func(name string, parameters json.RawMessage, handle Handle) (Plugin, error)

// After:
type FactoryFunc func(name string, parameters *json.Decoder, handle Handle) (Plugin, error)
```

The framework wraps raw plugin parameters in a `json.Decoder` with `DisallowUnknownFields()` before invoking each factory (or passes nil when no parameters were supplied). A new `plugin.StrictDecoder` helper centralizes decoder construction so production code and tests share one path.

All ~30 factories migrated to the new signature; bodies that consumed parameters now call `decoder.Decode(&cfg)` in place of `json.Unmarshal`.

**Reference deprecation pattern (approximate prefix cache):**
- `BlockSize` remains on the config struct so strict decode accepts it.
- Factory post-processes: if `BlockSize` is set, log a deprecation warning and map to `BlockSizeTokens` when the latter is at its default.
- The now-redundant runtime `BlockSize` check in `newDataProducer` is removed; deprecation handling is consolidated at the factory boundary.

**Disagg profile handler refactor:**
The factory previously relied on `json.Unmarshal`'s re-readable bytes to try both new and legacy schemas in succession. With the one-shot `Decoder` this no longer works (second `Decode` returns EOF). Refactored to capture raw bytes once and strict-decode each schema independently — surfaces a clear error when neither matches (mixed format or unknown field).

**Tests:**
- Two new tests on `approximateprefix` capture the desired behavior: `TestFactory_RejectsUnknownField` and `TestFactory_DeprecatedBlockSizeMapped`.
- Tests that previously asserted "unknown fields silently accepted" were updated to assert rejection (the behavior #1068 reverses).
- All existing tests constructing factory arguments via `json.RawMessage` now wrap with `plugin.StrictDecoder`.
- `make test-unit-epp` passes locally; lint/format/typos all clean. (`make vulncheck` hits the pre-existing local-only issue tracked by #1122 / fixed by #1132 — unrelated to this change.)

**Notes:**
- Single atomic PR per [shmuelk's feedback](https://github.com/llm-d/llm-d-router/issues/1068#issuecomment-4452389806) — Go static typing requires the signature change and all factory migrations to land together.
- Per [elevran's guidance](https://github.com/llm-d/llm-d-router/issues/1068#issuecomment-4451660760), deprecated fields are accepted with a warning, not rejected; only *removed* (truly unknown) fields fail strict parsing.

Fixes #1068

**Release note**:
```release-note
EPP now strictly parses plugin configurations — unknown fields cause plugin initialization to fail with a clear error rather than being silently ignored. Deprecated fields continue to be accepted with a warning until they are removed.
```

### DIFF
--- a/cmd/epp/runner/test_runner.go
+++ b/cmd/epp/runner/test_runner.go
@@ -39,7 +39,7 @@ func NewTestRunnerSetup(ctx context.Context, cfg *rest.Config, opts *runserver.O
 
 	if mockDataSource != nil {
 		mockType := mockDataSource.TypedName().Type
-		fwkplugin.Register(mockType, func(name string, _ json.RawMessage, _ fwkplugin.Handle) (fwkplugin.Plugin, error) {
+		fwkplugin.Register(mockType, func(name string, _ *json.Decoder, _ fwkplugin.Handle) (fwkplugin.Plugin, error) {
 			return mockDataSource, nil
 		})
 		defer delete(fwkplugin.Registry, mockType)
--- a/deploy/config/pd-epp-config.yaml
+++ b/deploy/config/pd-epp-config.yaml
@@ -2,10 +2,11 @@
 apiVersion: llm-d.ai/v1alpha1
 kind: EndpointPickerConfig
 plugins:
-- type: prefix-cache-scorer
+- type: approx-prefix-cache-producer
   parameters:
     maxPrefixBlocksToMatch: 256
     lruCapacityPerServer: 31250
+- type: prefix-cache-scorer
 - type: queue-scorer
 - type: prefill-filter
 - type: decode-filter
--- a/deploy/config/sim-epp-config.yaml
+++ b/deploy/config/sim-epp-config.yaml
@@ -3,10 +3,11 @@
 apiVersion: llm-d.ai/v1alpha1
 kind: EndpointPickerConfig
 plugins:
-- type: prefix-cache-scorer
+- type: approx-prefix-cache-producer
   parameters:
     maxPrefixBlocksToMatch: 256
     lruCapacityPerServer: 31250
+- type: prefix-cache-scorer
 - type: decode-filter
 - type: max-score-picker
 - type: single-profile-handler
--- a/deploy/config/sim-pd-epp-config.yaml
+++ b/deploy/config/sim-pd-epp-config.yaml
@@ -3,12 +3,13 @@
 apiVersion: llm-d.ai/v1alpha1
 kind: EndpointPickerConfig
 plugins:
-- type: prefix-cache-scorer
+- type: approx-prefix-cache-producer
   parameters:
     blockSizeTokens: 16
     autoTune: false
     maxPrefixBlocksToMatch: 256
     lruCapacityPerServer: 31250
+- type: prefix-cache-scorer
 - type: queue-scorer
 - type: prefill-filter
 - type: decode-filter
--- a/pkg/epp/README.md
+++ b/pkg/epp/README.md
@@ -1,7 +1,7 @@
 # The EndPoint Picker (EPP)
-This package provides the reference implementation for the Endpoint Picker (EPP). As demonstrated in the diagram below, it implements the [extension protocol](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/v1.5.0/docs/proposals/004-endpoint-picker-protocol), enabling a proxy or gateway to request endpoint hints from an extension, and interacts with the model servers through the defined [model server protocol](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/v1.5.0/docs/proposals/003-model-server-protocol).
+This package provides the reference implementation for the Endpoint Picker (EPP). As demonstrated in the diagram below, it implements the [extension protocol](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/docs/proposals/004-endpoint-picker-protocol.md), enabling a proxy or gateway to request endpoint hints from an extension, and interacts with the model servers through the defined [model server protocol](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/docs/proposals/003-model-server-protocol.md).
 
-![Architecture Diagram](https://github.com/kubernetes-sigs/gateway-api-inference-extension/raw/v1.5.0/docs/endpoint-picker.svg)
+![Architecture Diagram](https://github.com/kubernetes-sigs/gateway-api-inference-extension/raw/main/docs/endpoint-picker.svg)
 
 
 ## Core Functions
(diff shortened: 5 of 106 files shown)


liu-cong:
@copilot resolve the merge conflicts in this pull request

gkneighb:
Force-pushed a clean rebase: squashed everything to a single signed commit on top of current `main`, preserved @liu-cong as co-author + DCO sign-off for the conflict resolution work from `5ec8304d`. This unblocks the signed-commits check (and is also why the prior CI failed).

Verified locally: `make test-unit-epp` + `make presubmit` green (DCO, signed-commits-check, go-mod-check, gofmt, golangci-lint, typos, govulncheck). Diff vs. main: 100 files, +339/−243.

Thanks @liu-cong for taking the conflict resolution off my plate — sorry the rewrite was needed for the signing gate.

liu-cong:
FIXES https://github.com/llm-d/llm-d-router/issues/1068

/lgtm
/approve

gkneighb:
Pushed a fixup. The `lint-and-test` failure was strict parsing correctly catching two orphan fields in `test/integration/epp/wellknown_configs_test.go` that were silently dropped before:

- `tiered-prefix-cache-cpu` used `producer:` on `prefix-cache-scorer`; the struct field is `prefixMatchInfoProducerName` (`pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go:35`). Renamed.
- `wide-ep-lws-experimental-dp-aware` set `threshold: 0` and `hashBlockSize: 5` on `pd-profile-handler`; neither field exists on `pdProfileHandlerParameters` (`pkg/epp/framework/plugins/scheduling/profilehandler/disagg/pd_profile_handler.go:32`). Dropped.

Verified locally: `make test-integration-hermetic PATTERN='TestWellKnownConfigs'` — all 7 subtests pass. This is exactly the kind of latent bug the PR was supposed to surface, so I think the test-fixture fix is the right call rather than adding deprecation shims.

ahg-g:
Is this a flaky test?

```
[FAIL] Disruption tests when A decode pod is killed mid-request [It] should recover and route to surviving pods [Disruptive]
  /app/test/utils/igw/utils.go:278
```

gkneighb:
@ahg-g That's the `[Disruptive]` e2e test that kills a decode pod, waits for routing to recover, then waits for the replacement to become Ready. The failure is at `DeploymentReadyReplicas` in `test/utils/igw/utils.go:278` — the timing-sensitive last step, not a routing assertion. Nothing in this PR's diff touches the EPP routing path, the data layer's pod discovery, or any deployment-readiness logic — it's all `*json.Decoder` config parsing in plugin factories. Almost certainly flaky.

Just rebased onto current main to clear the merge conflict (single one-line URL drift in `pkg/epp/framework/plugins/README.md`, `core/epp` → `core/router/epp`). Hermetic integration suite still green locally. Re-LGTM at your convenience.

elevran:
The same Ginkgo spec (Disruption tests › when A decode pod is killed mid-request › should recover and route to surviving pods) times out after 600s in test/utils/igw/utils.go:278, which is the Eventually(...) waiting for the e2e-epp deployment to become Available. That deploy never becomes `Available` because the e2e-epp pod is in `CrashLoopBackOff` (multiple restarts).

From the dumped pod logs (e.g., [here](https://github.com/llm-d/llm-d-router/actions/runs/26366955002/job/77615622324#step:22:1484)):
```console
Failed to parse configuration","error":"failed to load the configuration - plugin instantiation failed: \
  failed to create plugin 'prefix-cache-scorer' (type: prefix-cache-scorer): \
  failed to decode prefix cache scorer parameters: json: unknown field \"lruCapacityPerServer\"
```

The prefix-cache-scorer plugin's Config (pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go:33-36) only declares one field (a string `PrefixMatchInfoProducerName` with `json:"prefixMatchInfoProducerName,omitempty"`) but the e2e fixtures in test/e2e/configs_test.go (lines 10, 33, 92, 129, 163) feed it:
```yaml
  - type: prefix-cache-scorer
    parameters:
      maxPrefixBlocksToMatch: 256
      lruCapacityPerServer: 256
```
  
Both `lruCapacityPerServer` and `maxPrefixBlocksToMatch` actually belong to the `approximate-prefix` data producer (`pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go`:182), not to the scorer. The Pre-PR, lenient `json.Unmarshal` silently dropped the unknown fields; post-PR, strict decode hard-fails and the EPP exits in startup.

The same issue is in the (local) deploy configs
  - deploy/config/sim-epp-config.yaml:9 — lruCapacityPerServer: 31250 under prefix-cache-scorer
  - deploy/config/sim-pd-epp-config.yaml:11
  - deploy/config/pd-epp-config.yaml:8

Previous failures in this PR:
| Run | Job | Head SHA | Date | Failed step | EPP CrashLoopBackOff (`lruCapacityPerServer`) |
|---|---|---|---|---|---|
| [26366955002](https://github.com/llm-d/llm-d-router/actions/runs/26366955002) | [77615622324](https://github.com/llm-d/llm-d-router/actions/runs/26366955002/job/77615622324) | 2c3ad5d | 2026-05-24 16:43Z | `Run make test-e2e` | yes |
| [26323303763](https://github.com/llm-d/llm-d-router/actions/runs/26323303763) | [77606828778](https://github.com/llm-d/llm-d-router/actions/runs/26323303763/job/77606828778) | 6e1419d | 2026-05-23 04:20Z | `Run make test-e2e` | yes |
| [26319463886](https://github.com/llm-d/llm-d-router/actions/runs/26319463886) | — | ea0aa5b | 2026-05-23 01:12Z | `Run hermetic integration tests` (e2e skipped) | n/a |

So on both Disruption runs:
- e2e-epp pod in CrashLoopBackOff, multiple restarts, container epp not ready
- EPP setup log: Failed to parse configuration ... failed to decode prefix cache scorer parameters: json: unknown field
  "lruCapacityPerServer"
- Ginkgo [FAILED] Timed out after 600.000s in the disruption spec waiting on the deployment in Eventually()

**Suggested fix**:
Either drop those keys from every prefix-cache-scorer parameters block (test fixtures + deploy configs), or move them under the plugin that actually owns them (the approximate-prefix data producer). The PR's whole point is that this kind of mismatch should fail loudly, so the test failure is the strict decoder doing its job; the work to land alongside it is updating the in-tree configs.

gkneighb:
@elevran — you nailed it, not flaky at all. I owe @ahg-g and you a clearer initial response; the disruption suite was failing because strict parsing was doing its job and the in-tree configs hadn't caught up. Sorry for the misdirection.

Pushed a fix that:
- Moves `maxPrefixBlocksToMatch`, `lruCapacityPerServer`, `blockSizeTokens`, `autoTune` from `prefix-cache-scorer` `parameters` to an explicit `approx-prefix-cache-producer` entry (the plugin that actually owns them, per `pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go`). Same pattern the bundled `tiered-prefix-cache-cpu` well-known config already uses.
- Drops the truly-orphan `hashBlockSize` field from `decodeOnlyConfig` (no field of that name on any plugin).
- Covers all 5 `test/e2e/configs_test.go` fixtures + all 3 `deploy/config/*-epp-config.yaml` files.
- Adds a new hermetic smoke test `TestE2EConfigs_StartupParseSmoke` that loads each e2e config through `NewTestHarness` so the next round of orphan-field mismatch is caught locally in <5s instead of via opaque CI CrashLoopBackOff.

Verified: `TestE2EConfigs_StartupParseSmoke` and `TestWellKnownConfigs` both pass; `make presubmit` + `make test-unit-epp` green. Re-LGTM at your convenience and please drop the `/hold` if you're happy.

PR #1135, opened 2026-05-14
file-discovery (3/4):  add FileDiscovery plugin with YAML/JSON file loading and fsnotify watch

ezrasilvera:
This PR is the third of 4 PRs adding file based discovery allowing the EPP to run without Kubernetes. The file based discovery is the first phase in re-creating the functionality of #1007.

This PR implements the FileDiscovery plugin, the first EndpointDiscovery backend for  non-Kubernetes deployments. It reads a YAML or JSON file that lists inference endpoints and drives their lifecycle in the datastore via DiscoveryNotifier.

**Note**: The FileDiscovery plugin is fully implemented and tested in isolation but is not yet usable end-to-end. The runner wiring plugin registration, config detection, and the runWithFileDiscovery execution path will land in the next (and final) PR. Until then setting dataLayer.discovery.pluginRef in the config will have no effect.

- Loads endpoints from a YAML/JSON file at startup, emitting one Upsert per entry so the datastore is populated before the EPP begins serving traffic
- Optionally watches the file with fsnotify and reconciles additions/removals on every write, allowing zero-downtime endpoint updates without restarting the EPP
- Removes stale endpoints on reload by diffing the previous snapshot against the new file contents
- Configured via EndpointPickerConfig.dataLayer.discovery.pluginRef pointing to a plugin instance of type file-discovery with path and watchFile params


pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+		meta := &fwkdl.EndpointMetadata{
+			NamespacedName: types.NamespacedName{Name: e.Name, Namespace: ns},
+			PodName:        e.Name,
+			Address:        e.Address,
+			Port:           e.Port,
+			MetricsHost:    e.MetricsHost,
+			Labels:         e.Labels,
+		}
+		if meta.MetricsHost == "" {
```
elevran: `notifier.Upsert(meta)` is called inside the validation loop. If endpoint N fails validation (bad IP, bad port), endpoints 0…N-1 have already been Upserted, but `f.current = incoming` is never reached. Those endpoints are now in the datastore but absent from `f.current`. The delete-reconciliation loop at the end can never emit a Delete for them and they stay in the datastore indefinitely, even if removed from the file.

Suggest validate all entries first, then apply all Upserts.

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if err := watcher.Add(f.path); err != nil {
+		return fmt.Errorf("file-discovery: failed to watch %s: %w", f.path, err)
+	}
+
+	logger.Info("watching endpoints file for changes")
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
```
elevran, line 134: More context on @hexfusion's comment - this can happend outside of k8s.

`watcher.Add(f.path)` at startup watches the file's *inode*. Many editors (vim, etc.) and Kubernetes ConfigMap volume mounts use atomic renames: write to a temp file, then `rename()` over the original. After the rename, the inode at `f.path` changes; fsnotify fires a `Create` event for the new file, but the old-inode watch is now dead. Subsequent writes generate no events since the watcher is silently broken after the first atomic update.

The `Create` branch correctly reloads for the rename itself, but needs to re-attach the watcher to the new inode:

```suggestion
			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) {
				if event.Has(fsnotify.Create) {
					// Re-attach after atomic rename (new inode at same path).
					_ = watcher.Add(f.path)
				}
				logger.Info("endpoints file changed, reloading")
				if err := f.load(notifier); err != nil {
					logger.Error(err, "failed to reload endpoints file")
				}
			}
```

pkg/epp/framework/plugins/datalayer/discovery/file/file_discovery.go
```diff
@@ -0,0 +1,206 @@
...
+	logger.Info("watching endpoints file for changes")
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return nil
+			}
+			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) {
```
hexfusion: i know your intention is to not depend on K8s, but my understanding is that, for example, with a ConfigMap  rotation is a symlink swap on data via atomic rename, but from a fsnotify perspective it shows up as Remove + Create. If we want to support both, it might make sense to align with controller-runtime/pkg/certwatcher or document the gap more explicitly.

ref. https://github.com/kubernetes-sigs/controller-runtime/blob/589525a057391c2be40fd0de7c31ad3bac69a5d7/pkg/certwatcher/certwatcher.go#L227

elevran (commented):
Thanks @ezrasilvera 
Two issues in `load()` worth fixing before merging: 
- a partial-Upsert atomicity bug that can leave zombie endpoints in the datastore; and 
- missing watcher re-add that breaks `watchFile` mode after atomic renames
See inline.

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
ezrasilvera: @elevran good catch Before implementing, we can take 2 approaches what do you think is better ?

1. All-or-nothing (two-pass fix): If any entry is bad, the whole update is rejected and the datastore keeps the last good state. This is a safe and predictable state, but a single typo blocks all other valid updates.
2. Best-effort (skip bad, apply good): valid entries are upserted, invalid ones are skipped with a logged error. 

I prefer option 2 uless you have strong preferences to 1.  For example, if  you have 10 vLLM instances and one has a bad IP, the other 9 still get updated seems like a more resilient operationally.

pkg/epp/framework/plugins/datalayer/discovery/file/file_discovery.go
ezrasilvera: Thanks @elevran @hexfusion, I addressed both comments together: `Remove` re-adds the watcher (covers ConfigMap rotation where the watch dies before `Create` arrives), `Create` re-adds and reloads (covers editor atomic rename). This is now aligned with certwatcher. Fixed in 82c39a73.

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
ezrasilvera: Fixed. See https://github.com/llm-d/llm-d-router/pull/1135#discussion_r3246436906

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
ezrasilvera: Thanks. Fixed in abcc755c

ezrasilvera:
Once #1150 is merged I'll remove my docs commit and rebase.

ezrasilvera:
@elevran I addressed all comments. Thanks

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
```diff
@@ -0,0 +1,216 @@
...
+	"os"
+	"strconv"
+
+	"github.com/fsnotify/fsnotify"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/yaml"
+
+	fwkdl "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/datalayer"
+	fwkplugin "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/interface/plugin"
```
ahg-g: pls update the path `llm-d-router`

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// EndpointsFile is the top-level structure of the endpoints YAML/JSON file.
+type EndpointsFile struct {
+	Endpoints []EndpointEntry `json:"endpoints" yaml:"endpoints"`
+}
+
+type params struct {
+	Path      string `json:"path"`
+	WatchFile bool   `json:"watchFile"`
```
ahg-g: pls document the parameters, this is a user facing api

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// FileDiscovery implements EndpointDiscovery by reading a static endpoints file.
+type FileDiscovery struct {
+	typedName fwkplugin.TypedName
+	path      string
+	watchFile bool
+	current   map[types.NamespacedName]struct{}
```
ahg-g: pls add a comment documenting what this map is for and the value it tracks.

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
```diff
@@ -0,0 +1,216 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package file provides a file-based EndpointDiscovery implementation that reads
+// a YAML (or JSON) file listing inference endpoints.
+package file
```
ahg-g, line 19: nit: we could drop the file_ prefix from file_discovery.go, the pkg is already named file; in fact we can just name it plugin.go

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
```diff
@@ -0,0 +1,216 @@
...
+
+const PluginType = "file-discovery"
+
+// EndpointEntry is the YAML/JSON representation of a single endpoint.
+type EndpointEntry struct {
+	Name        string            `json:"name"                  yaml:"name"`
+	Namespace   string            `json:"namespace,omitempty"   yaml:"namespace,omitempty"`
+	Address     string            `json:"address"               yaml:"address"`
+	Port        string            `json:"port"                  yaml:"port"`
+	MetricsHost string            `json:"metricsHost,omitempty" yaml:"metricsHost,omitempty"`
```
ahg-g: What is a "Metrics Host"?

ahg-g (commented):
pls add a README.md file to document the plugin, see other plugins for examples or https://github.com/llm-d/llm-d-router/issues/917 for a suggested template..

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
ahg-g: if this is tracking the endpoints, perhaps name it `endpoints`?

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
ezrasilvera: Added a comment to explain c6901f69. This is documented in the discovery readme but I was planning on adding it only when I actually wire in the feature in the next PR. 
A host:port string that tells the EPP where to scrape Prometheus metrics from for this endpoint. Allow to indicate a different port than they accept inference requests on (e.g., vLLM running inference on :8000 but /metrics on :9090).

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
ezrasilvera: Sounds reasonable. Fixed in  624dc48a

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
ezrasilvera: Done. 2295a193

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
ezrasilvera: Oh that happened over the last couple of days. Rebased and fixed in 37b5139

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
ezrasilvera: Done both in 06986a05

ezrasilvera:
> pls add a README.md file to document the plugin, see other plugins for examples or #917 for a suggested template..

I'm planning to have a discovery.md (next PR) describing the overall discovery plugin framework.   I also added a README to the filediscovery folder according to the template.  93421cc8

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
```diff
@@ -0,0 +1,108 @@
...
+## Inputs Consumed
+
+A YAML or JSON file with the schema below. The path is supplied via the
+plugin's `path` parameter.
+
+```yaml
+endpoints:
+  - name: <string>              # required -- unique within the file
+    namespace: <string>         # optional -- defaults to "default"
+    address: <IPv4 or IPv6>     # required -- must be a valid IP address
```
elevran: nit: not sure the EPP can cope with IPv6. Please remove and validate that only IPv4 is passed.

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Inputs Consumed
+
+A YAML or JSON file with the schema below. The path is supplied via the
+plugin's `path` parameter.
+
+```yaml
+endpoints:
+  - name: <string>              # required -- unique within the file
+    namespace: <string>         # optional -- defaults to "default"
```
elevran, line 44: nit: consider moving away from k8s heritage and using `scope` instead of `namespace`, `tags` instead of `labels`, etc.

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+```yaml
+endpoints:
+  - name: vllm-0
+    address: "10.0.0.1"
+    port: "8000"
+  - name: vllm-1
+    address: "10.0.0.2"
+    port: "8000"
+    metricsHost: "10.0.0.2:9090"
```
elevran: nit: suggest removing it from the example or documenting when is it useful (i.e., in which deployment is the port and host different for serving and metrics, can you really have a completely different metric host name or is it just a port change?)

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
```diff
@@ -0,0 +1,108 @@
...
+- The endpoints file is capped at 1 MiB.
+- `address` must be a literal IPv4 or IPv6 address. Hostnames are not
+  resolved.
+- File-discovery mode runs the EPP without a Kubernetes controller manager,
+  so several K8s-only features are inactive: the `InferenceModelRewrite`
+  and `InferenceObjective` reconcilers do not run, and any
+  `k8s-notification-source` plugin in the data layer config will not bind.
+  The runner emits a startup log naming the inactive features.
+- A single bad entry on initial load is logged and skipped, not fatal. If
+  the entire file is unreadable or unparseable, startup fails.
```
elevran:
```suggestion
  the entire file is not readable or fails to parse, startup fails.
```
Or fix the typo based on suggestion below.

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
elevran: sounds like only a port change makes sense. The EPP can not scrape and serve on **different IPs**, only same IP and differenet port is feasible (and _deprecated_).

elevran:
please do not use Sleep in tests, use Eventually instead.

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
ezrasilvera: fixed in 5ae49ca9

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
ezrasilvera: Fixed in  5ae49ca9  change to support Port only

ezrasilvera:
> please do not use Sleep in tests, use Eventually instead.

Done. 73eaa820

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
```diff
@@ -0,0 +1,237 @@
...
+
+	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+	fwkplugin "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+)
+
+const PluginType = "file-discovery"
+
+// EndpointEntry is the YAML/JSON representation of a single endpoint.
+type EndpointEntry struct {
+	Name      string `json:"name"                  yaml:"name"`
```
elevran: Q: IIRC, the `sigs.k8s.io/yaml` converts all to `json` so you might be able to get away with json tags alone

elevran (commented):
/lgtm 
from my point of view (nits can be addressed or not at your discretion).
Leaving final approval to @ahg-g

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
ezrasilvera: Fixed in bdf8ebd8

ezrasilvera:
@elevran as you suggested I removed the MetricsPort and HOST completely so scrapping always from address:port fixed in d1151603

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
ezrasilvera: IN general it makes sense  however we will have runtime discrepancy for the user: For example in the YAML he will use scope: my-scope  and tags: but the EPP logs and Prometheus metrics display those values under their internal labels:
1) endpoint_namespace="my-scope" (not endpoint_scope)
2) selector / filter docs talk about labels, not tags
I personally would prefer not to change unless you think it's better even with these discrepancies

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
```diff
@@ -0,0 +1,222 @@
...
+// FileDiscovery implements EndpointDiscovery by reading a static endpoints file.
+type FileDiscovery struct {
+	typedName fwkplugin.TypedName
+	path      string
+	watchFile bool
+	// endpoints is the set of endpoint identities applied to the datastore
+	// from the last successful load. Used as a key set only -- values are
+	// zero-byte structs. Compared against the entries parsed during a
+	// reload to compute which endpoints to delete from the datastore.
+	endpoints map[types.NamespacedName]struct{}
```
ahg-g, line 77: nit, I think for a set `any` is typically being used as the value

ahg-g (commented):
/approve

PR #1136, opened 2026-05-14
ci: trigger check workflows on pull_request_target for bot-created PRs

elevran:
Add `pull_request_target` trigger to `md-link-check.yml` and `ci-pr-checks.yaml`, with a job-level skip condition and an updated concurrency key.

GitHub suppresses `pull_request` events when `GITHUB_TOKEN` is the actor creating the PR. The `release-notes-update.yaml` workflow uses `github.token` to open its PRs, so the `pull_request`-triggered workflows never fire and their required checks stay pending indefinitely.

`ci-signed-commits.yaml` was already correct (uses `pull_request_target` + skip condition) and is unchanged.

- `pull_request_target` is added alongside `pull_request` so the workflow is triggered for bot-created PRs.
- Each gated job has `if: github.event_name != 'pull_request_target'`, so the job is **skipped** (not run) when triggered this way. GitHub treats a skipped conclusion as satisfying branch protection.
- The concurrency group key now includes `${{ github.event_name }}` to prevent the `pull_request_target` run from cancelling an in-flight `pull_request` run for human PRs (which also receive both events).

See for example #1117 and #1129

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -8,17 +8,22 @@ on:
   pull_request:
     branches:
       - main
+  pull_request_target:
+    branches:
+      - main
+    types: [ opened, synchronize, reopened ]
 
 permissions:
   contents: read
   actions: read
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
   check-changes:
+    if: github.event_name != 'pull_request_target'
     runs-on: ubuntu-latest
     outputs:
       src: ${{ steps.filter.outputs.src }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-pr-checks.yaml
```diff
@@ -8,17 +8,22 @@ on:
...
   actions: read
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
   check-changes:
+    if: github.event_name != 'pull_request_target'
```
hexfusion, line 31: minor: could we add a comment on what this does I needed to look it up :)

elevran:
@hexfusion Good call. Added a 5-line block comment above `if:` conditions (in both `ci-pr-checks.yaml` and `md-link-check.yml`) explaining the mechanism: GitHub suppresses `pull_request` events when `GITHUB_TOKEN` (the Bot) creates the PR, so `pull_request_target` is registered purely to surface a `skipped` conclusion that satisfies branch protection; the skip itself prevents duplicate runs for human PRs.

hexfusion:
LGTM

PR #1108, opened 2026-05-14
fix: resolve Trivy CVEs and CI action deprecation warnings

elevran:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Address trivy scan results in building images

- upgrades `aquasecurity/trivy-action` to v0.36.0 and `github/codeql-action/upload-sarif` to v4.35.4, eliminating the Node.js 20 and CodeQL v3 deprecation warnings on CI jobs
- bump Prometheus to v0.311.3, fixing CVE-2026-42154, CVE-2026-42151. Changed code to the breaking parser API change in prometheus v0.311.x (`ParseExpr` moved from a package-level function to a `Parser` interface method)
- bump `go-openapi/swag/loading` to v0.26.0 to fix a broken transitive test-dependency that blocked `go mod tidy`

**Which issue(s) this PR fixes**:

NONE

**Release note** _(write `NONE` if no user-facing change)_:
```release-note
NONE
```

PR #1138, opened 2026-05-14
refactor(metrics): replace prometheus server parser with regex

elevran:
**What type of PR is this?**
/kind cleanup

`parseStringToSpec` in the metrics extractor used `prometheus/prometheus/promql/parser` and `prometheus/prometheus/model/labels` only to parse strings of the form `metric_name{k=v,...}` and extract equality label pairs. Two compiled regexes cover the same grammar with no behaviour change; all existing tests pass.

Removing `prometheus/prometheus v0.311.3` results in 84 lines removed from `go.sum`.

NONE.

Ref https://github.com/llm-d/llm-d-router/pull/1108#discussion_r3242066431

```release-note
NONE
```


go.mod
```diff
@@ -26,7 +26,6 @@ require (
 	github.com/prometheus/client_golang v1.23.2
 	github.com/prometheus/client_model v0.6.2
 	github.com/prometheus/common v0.67.5
-	github.com/prometheus/prometheus v0.311.3
```
hexfusion, line 29: ❤️

pkg/epp/framework/plugins/datalayer/extractor/metrics/spec.go
```diff
@@ -23,12 +23,20 @@ import (
...
 	dto "github.com/prometheus/client_model/go"
-	"github.com/prometheus/prometheus/model/labels"
-	"github.com/prometheus/prometheus/promql/parser"
 
 	sourcemetrics "github.com/llm-d/llm-d-inference-scheduler/pkg/epp/framework/plugins/datalayer/source/metrics"
 )
 
+// metricSpecRe matches: optional-space metric-name optional-space optional-{labels} optional-space end.
+// Prometheus metric names may contain letters, digits, underscores, and colons.
+var metricSpecRe = regexp.MustCompile(`^\s*([a-zA-Z_:][a-zA-Z0-9_:]*)\s*(?:\{([^}]*)\})?\s*$`)
```
hexfusion: since we still have prometheus/common i wonder if there is anything here that could help maybe IsValidMetricName?

ref. https://github.com/prometheus/common/blob/ce9215c53d8c8e507c5f72a69d80568c072be3d9/model/metric.go#L151

elevran: Thanks for pointing this out.
Will check if we can similarly replace the other regexps with a prom subset

elevran: RegExps used to valdate the structure, actual parsing validation delegated to prometheus/commeon/model

elevran:
@hexfusion @ahg-g - thoughts? Anything else needed or can we drop the larger Prom dependency?

hexfusion:
LGTM nice clean up thanks!

PR #1145, opened 2026-05-15
fix: --inference-pool-namespace/--inference-pool-name silently ignored when YAML config present

weizhoublue:
**What type of PR is this?**

/kind bug

Trigger Scenario: 
User sets InferencePool via deprecated CLI flags (--inference-pool-namespace myns --inference-pool-name mypool) while also mounting a YAML config that contains  inference-pool: yaml-ns/yaml-pool.

The YAML value (yaml-ns/yaml-pool) silently takes effect instead of the CLI value (myns/mypool). The user's explicit CLI intent is ignored with no warning.

 Expected Behavior: 
CLI flags always take precedence over YAML configuration, regardless of whether the user uses the new --inference-pool flag or the deprecated  --inference-pool-namespace / --inference-pool-name flags.

 Root Cause: 
mergeYAMLConfiguration()  only guards YAML merge with !opts.isFlagSet(inferencePool), omitting checks for the two deprecated flags. When YAML wins the merge,  Complete()  unconditionally parses opts.inferencePool back into Namespace/Name, overwriting the deprecated flags' values. The TLS deprecated fields have identical  protection (checking both old and new flags) — InferencePool simply missed it.

strengthen https://github.com/llm-d/llm-d-router/pull/683

### DIFF
--- a/pkg/sidecar/proxy/options.go
+++ b/pkg/sidecar/proxy/options.go
@@ -499,7 +499,8 @@ func (opts *Options) mergeYAMLConfiguration(cfg yamlConfiguration) {
 		opts.decoderInsecureSkipVerify = *cfg.DecoderTLSInsecureSkipVerify
 	}
 
-	if cfg.InferencePool != "" && !opts.isFlagSet(inferencePool) {
+	if cfg.InferencePool != "" && !opts.isFlagSet(inferencePool) &&
+		!opts.isFlagSet(inferencePoolNamespace) && !opts.isFlagSet(inferencePoolName) {
 		opts.inferencePool = cfg.InferencePool
 	}
 	if cfg.PoolGroup != "" && !opts.isFlagSet(poolGroup) {


weizhoublue:
hi, @DhritiShikhar  , could you help check whether this pr make sense ? thanks

elevran:
thanks the fix @weizhoublue.
The two flags were marked as deprecated in the 0.7 release and will thus be removed in v0.9 - the next release, making this contribution obsolete.

@ahg-g @vMaroon @liu-cong what do you think?

ahg-g:
+1 to deprecating the flags;

btw, why does the sidecar watches on the infpool?

elevran:
> +1 to deprecating the flags;
> 
> btw, why does the sidecar watches on the infpool?

For SSRF (Server-Side Request Forgery) protection for prefill targets.

It receives the prefill URL it should call out to. Without validation, an attacker could craft a request pointing the sidecar at any internal service. So the sidecar watches the InferencePool and Pods (combined by the pool.Selector) to determine valid targets. Targets not part of the Pool are rejected.
The informers keep the AllowList "live" with pod changes

ahg-g:
> > +1 to deprecating the flags;
> > btw, why does the sidecar watches on the infpool?
> 
> For SSRF (Server-Side Request Forgery) protection for prefill targets.
> 
> It receives the prefill URL it should call out to. Without validation, an attacker could craft a request pointing the sidecar at any internal service. So the sidecar watches the InferencePool and Pods (combined by the pool.Selector) to determine valid targets. Targets not part of the Pool are rejected. The informers keep the AllowList "live" with pod changes

I know this is only for InfPool objects, but my concern is that now every pod is creating a watch on the api server, workload pods don't usually do that. 

What capabilities do we assume the attacker can have? The EPP can scrape any headers that it uses as part of the disagg protocol, wouldn't that be sufficient?

elevran:
> I know this is only for InfPool objects, but my concern is that now every pod is creating a watch on the api server, workload pods don't usually do that.
> 
> What capabilities do we assume the attacker can have? The EPP can scrape any headers that it uses as part of the disagg protocol, wouldn't that be sufficient?

1. **Not** a security expert.
2. IIRC, the protection was put in place assuming direct access by other entities in the cluster. These won't necessarily go through the EPP and could hit the sidecar directly. It's a question of the threat model and what we think is reasonable to assume regarding attackers.

DhritiShikhar:
@weizhoublue `inference-pool` in YAML config takes precedence over deprecated flags `--inference-pool-name` and `--inference-pool-namespace` provided via CLI.

weizhoublue:
> @weizhoublue `inference-pool` in YAML config takes precedence over deprecated flags `--inference-pool-name` and `--inference-pool-namespace` provided via CLI.

Thanks. If that was the original intention, this PR is unnecessary.

PR #1148, opened 2026-05-15
Add plugin state debug endpoint

### LINKED ISSUE #1074: [Operations] Add a debug endpoint to dump all plugin internal states
This issue moved from https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2901 and was created by https://github.com/ChenChuang

1. Define a StateDumper interface
```go
// StateDumper is an optional interface that plugins can implement to expose
// their internal state for debugging purposes.
type StateDumper interface {
    // DumpState returns a JSON-serializable representation of the plugin's
    // current internal state. This is intended for debugging only.
    DumpState() any
}
```
2. Add a new package pkg/common/observability/debug to maintain all plugins that implement StateDumper and register a new handler on the metrics/admin server (e.g., /debug/plugins/state) that:  
    - Iterates over all plugins via Handle.GetAllPluginsWithNames()
    - For each plugin that implements StateDumper, calls DumpState()
    - Returns a JSON response with all plugin states:
```json
{
  "plugins": {
    "approx-prefix-cache-producer": {
      "state": { "blocks": 1024 ... }
    },
    "inflight-load-producer": {
      "state": { "requests": {"pod-1": {"concurrency": 5}, "pod-2": {"concurrency": 3}} }
    }
  },
  "timestamp": "2025-01-01T00:00:00Z"
}
```

**Why is this needed**:
Currently, there is no mechanism to inspect or dump the internal state of all running plugins at runtime. This makes it difficult to debug issues in production, especially for stateful plugins like inflight-load-producer or approx-prefix-cache.

wenhug:
## Summary
- Add optional `plugin.StateDumper` interface for plugins to expose sanitized debug state
- Register `/debug/plugins/state` on the existing metrics/debug server and return deterministic JSON for named dumpers
- Add a safe in-flight load dumper with per-endpoint request/token counters

Fixes #1074

## Testing
- `go test ./pkg/epp/server ./pkg/epp/framework/interface/plugin ./pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload ./cmd/epp/runner`
- `go list ./... | grep -Ev 'github.com/llm-d/llm-d-inference-scheduler/test/(config|e2e|integration|profiling|sidecar/e2e)(/|$)' | xargs go test`

Note: full `go test ./...` reaches repo suites that require a local Kubernetes/kubebuilder environment; those failed here because `/usr/local/kubebuilder/bin/etcd` and a local Kubernetes API were unavailable.

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -137,6 +137,7 @@ type Runner struct {
 	customCollectors     []prometheus.Collector
 	parser               fwkrh.Parser
 	dlRuntime            *datalayer.Runtime
+	pluginHandle         fwkplugin.HandlePlugins
 }
 
 // WithExecutableName sets the name of the executable containing the runner.
@@ -321,6 +322,11 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 			setupLog.Error(err, "Failed to setup pprof handlers")
 			return nil, nil, err
 		}
+	}
+
+	if err = runserver.SetupPluginStateDebugHandler(mgr, r.pluginHandle); err != nil {
+		setupLog.Error(err, "Failed to setup plugin state debug handler")
+		return nil, nil, err
 	}
 
 	// --- Initialize Core EPP Components ---
@@ -616,6 +622,7 @@ func (r *Runner) parseConfigurationPhaseTwo(ctx context.Context, rawConfig *conf
 	// The plugins will be executed in topologically sorted order to ensure that data is produced before it is consumed.
 	r.requestControlConfig.OrderDataProducerPlugins(dag)
 
+	r.pluginHandle = handle
 	r.parser = handlers.NewParser(cfg.ParserConfig)
 	logger.Info("loaded configuration from file/text successfully")
 
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -324,9 +324,13 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 		}
 	}
 
-	if err = runserver.SetupPluginStateDebugHandler(mgr, r.pluginHandle); err != nil {
-		setupLog.Error(err, "Failed to setup plugin state debug handler")
-		return nil, nil, err
+	if opts.EnablePluginStateDebug {
+		if r.pluginHandle == nil {
+			setupLog.Info("Plugin state debug handler not registered: plugin handle unavailable")
+		} else if err = runserver.SetupPluginStateDebugHandler(mgr, r.pluginHandle); err != nil {
+			setupLog.Error(err, "Failed to setup plugin state debug handler")
+			return nil, nil, err
+		}
 	}
 
 	// --- Initialize Core EPP Components ---
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -347,13 +347,11 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 		}
 	}
 
-	if opts.EnablePluginStateDebug {
-		if r.PluginHandle == nil {
-			setupLog.Info("Plugin state debug handler not registered: plugin handle unavailable")
-		} else if err = runserver.SetupPluginStateDebugHandler(mgr, r.PluginHandle); err != nil {
-			setupLog.Error(err, "Failed to setup plugin state debug handler")
-			return nil, nil, err
-		}
+	if r.PluginHandle == nil {
+		setupLog.Info("Plugin state debug handler not registered: plugin handle unavailable")
+	} else if err = runserver.SetupPluginStateDebugHandler(mgr, r.PluginHandle); err != nil {
+		setupLog.Error(err, "Failed to setup plugin state debug handler")
+		return nil, nil, err
 	}
 
 	// --- Initialize Core EPP Components ---
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -347,9 +347,13 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 		}
 	}
 
-	if err = runserver.SetupPluginStateDebugHandler(mgr, r.PluginHandle); err != nil {
-		setupLog.Error(err, "Failed to setup plugin state debug handler")
-		return nil, nil, err
+	if opts.EnablePluginStateDebug {
+		if r.PluginHandle == nil {
+			setupLog.Info("Plugin state debug handler not registered: plugin handle unavailable")
+		} else if err = runserver.SetupPluginStateDebugHandler(mgr, r.PluginHandle); err != nil {
+			setupLog.Error(err, "Failed to setup plugin state debug handler")
+			return nil, nil, err
+		}
 	}
 
 	// --- Initialize Core EPP Components ---
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -347,6 +347,11 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 		}
 	}
 
+	if err = runserver.SetupPluginStateDebugHandler(mgr, r.PluginHandle); err != nil {
+		setupLog.Error(err, "Failed to setup plugin state debug handler")
+		return nil, nil, err
+	}
+
 	// --- Initialize Core EPP Components ---
 	if r.schedulerConfig == nil {
 		err := errors.New("scheduler config must be set either by config api or through code")
--- a/pkg/epp/framework/interface/plugin/plugins.go
+++ b/pkg/epp/framework/interface/plugin/plugins.go
@@ -23,6 +23,14 @@ type Plugin interface {
 	TypedName() TypedName
 }
 
+// StateDumper is an optional interface for plugins that can expose sanitized
+// internal state through operational debug endpoints.
+type StateDumper interface {
+	// DumpState returns JSON-serializable plugin state. Implementations must not
+	// include request payloads, credentials, or other sensitive values.
+	DumpState() any
+}
+
 // ConsumerPlugin defines the interface for a consumer.
 type ConsumerPlugin interface {
 	Plugin
--- a/pkg/epp/framework/interface/plugin/plugins.go
+++ b/pkg/epp/framework/interface/plugin/plugins.go
@@ -16,6 +16,8 @@ limitations under the License.
 
 package plugin
 
+import "encoding/json"
+
 // Plugin defines the interface for a plugin.
 // This interface should be embedded in all plugins across the code.
 type Plugin interface {
@@ -32,6 +34,21 @@ type DataDependencies struct {
 	// Optional keys — the framework logs a warning at init time but does NOT error if no producer exists.
 	// The plugin must handle the case where this data is absent at runtime.
 	Optional map[DataKey]any
+}
+
+// StateDumper is an optional interface for plugins that can expose sanitized,
+// bounded internal state through operational debug endpoints.
+//
+// DumpState is intended for on-demand debugging snapshots of dynamic runtime
+// state that is difficult to understand from metrics alone. Prefer metrics for
+// numeric time series, alerting, dashboards, and aggregation over time. Dumped
+// state should stay reasonably small; large or high-cardinality state should be
+// summarized, capped, or omitted.
+type StateDumper interface {
+	// DumpState returns a JSON-encoded representation of plugin state.
+	// Implementations own serialization and must not include request payloads,
+	// credentials, or other sensitive values.
+	DumpState() (json.RawMessage, error)
 }
 
 // ConsumerPlugin defines the interface for a consumer.
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"reflect"
+	"sort"
 	"sync"
 	"sync/atomic"
 
@@ -54,6 +55,7 @@ var (
 	_ requestcontrol.DataProducer          = &InFlightLoadProducer{}
 	_ datalayer.EndpointExtractor          = &InFlightLoadProducer{}
 	_ datalayer.Registrant                 = &InFlightLoadProducer{}
+	_ fwkplugin.StateDumper                = &InFlightLoadProducer{}
 )
 
 type InFlightLoadProducer struct {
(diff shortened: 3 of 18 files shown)


pkg/epp/server/plugin_state_debug.go
```diff
@@ -0,0 +1,104 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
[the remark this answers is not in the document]
wenhug: Fixed in 26f878cc by aligning the new pkg/epp/server files to the package-local 2025 copyright year.

pkg/epp/server/plugin_state_debug.go
```diff
@@ -0,0 +1,104 @@
...
+
+type pluginStateDebugResponse struct {
+	Plugins []pluginStateDebugEntry `json:"plugins"`
+}
+
+type pluginStateDebugEntry struct {
+	Name  string `json:"name"`
+	Type  string `json:"type"`
+	State any    `json:"state"`
+}
```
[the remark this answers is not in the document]
wenhug: Fixed in 26f878cc by changing the response to the issue #1074 shape: a top-level timestamp plus plugins keyed by plugin name, with each entry containing type and state. Tests freeze the clock so the timestamp remains deterministic.

cmd/epp/runner/runner.go
```diff
@@ -323,6 +324,11 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 		}
 	}
 
+	if err = runserver.SetupPluginStateDebugHandler(mgr, r.pluginHandle); err != nil {
```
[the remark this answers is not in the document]
wenhug: Fixed in 26f878cc by making registration tolerant of a missing plugin handle: setup now logs and skips the handler instead of aborting.

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
```diff
@@ -63,10 +65,59 @@ type InFlightLoadProducer struct {
...
+	for _, endpointID := range endpointIDs {
+		state.Endpoints = append(state.Endpoints, EndpointInFlightLoadState{
+			Endpoint: endpointID,
+			Requests: requestCounts[endpointID],
+			Tokens:   tokenCounts[endpointID],
+		})
+	}
+
+	return state
+}
```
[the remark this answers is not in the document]
wenhug: Addressed in 26f878cc with a DumpState doc comment that explicitly calls out the best-effort, non-atomic semantics of the separate request/token tracker snapshots.

cmd/epp/runner/runner.go
```diff
@@ -323,6 +324,11 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
...
+		setupLog.Error(err, "Failed to setup plugin state debug handler")
+		return nil, nil, err
+	}
```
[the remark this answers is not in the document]
wenhug: Fixed in 26f878cc by adding EnablePluginStateDebug / --enable-plugin-state-debug so operators can disable this debug endpoint, mirroring the existing pprof-style configuration surface.

cmd/epp/runner/runner.go
```diff
@@ -323,6 +324,15 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
...
 
+	if opts.EnablePluginStateDebug {
+		if r.pluginHandle == nil {
+			setupLog.Info("Plugin state debug handler not registered: plugin handle unavailable")
+		} else if err = runserver.SetupPluginStateDebugHandler(mgr, r.pluginHandle); err != nil {
+			setupLog.Error(err, "Failed to setup plugin state debug handler")
+			return nil, nil, err
+		}
+	}
+
```
shmuelk: As a new HTTP endpoint is being added, why does this need to be enabled?

pkg/epp/server/options.go
```diff
@@ -91,6 +91,7 @@ type Options struct {
 	MetricsPort            int    // The metrics port exposed by EPP. (TODO: uint16)
 	GRPCHealthPort         int    // The port used for gRPC liveness and readiness probes. (TODO: uint16)
 	EnablePprof            bool   // Enables pprof handlers.
+	EnablePluginStateDebug bool   // Enables the /debug/plugins/state handler.
```
shmuelk: Why is this needed?

pkg/epp/server/options.go
```diff
@@ -195,6 +196,8 @@ func (opts *Options) AddFlags(fs *pflag.FlagSet) {
 		"The port used for gRPC liveness and readiness probes.")
 	fs.BoolVar(&opts.EnablePprof, "enable-pprof", opts.EnablePprof,
 		"Enables pprof handlers. Defaults to true. Set to false to disable pprof handlers.")
+	fs.BoolVar(&opts.EnablePluginStateDebug, "enable-plugin-state-debug", opts.EnablePluginStateDebug,
+		"Enables the /debug/plugins/state handler on the metrics server. Defaults to false.")
```
shmuelk: Ditto?

cmd/epp/runner/runner.go
wenhug: It dumps live plugin state, so I kept it opt-in instead of exposing it by default on the metrics server.

pkg/epp/server/options.go
wenhug: Same reason — this keeps the debug endpoint explicitly opt-in.

pkg/epp/server/options.go
wenhug: Same here, this flag is the opt-in for that debug endpoint.

cmd/epp/runner/runner.go
wenhug: Keeping this as an explicit opt-in: the endpoint can expose any future `StateDumper` plugin state, so operators should intentionally enable it while still relying on the existing metrics/admin server auth controls when configured.

pkg/epp/server/options.go
wenhug: This option is the explicit opt-in switch for `/debug/plugins/state`; it defaults off so upgrades do not silently expose a new debug surface.

pkg/epp/server/options.go
wenhug: Same rationale as above: this flag gives operators an explicit way to enable the plugin-state debug endpoint when they want it.

wenhug:
Hi @elevran @nilig, gentle ping for review when you have a chance. Thanks!

pkg/epp/server/options.go
shmuelk: @wenhug A better approach would be to add code that checks that the client is from the address 127.0.0.1 (i.e. localhost), this would limit acces sto this endpoint to anyone who can run `kubectl port-forward ...` which is only admin of the cluster. Then this could always be added and an operator could get this debug information anytime they wanted to without the need to plan in advance.

pkg/epp/framework/interface/plugin/plugins.go
```diff
@@ -23,6 +23,14 @@ type Plugin interface {
 	TypedName() TypedName
 }
 
+// StateDumper is an optional interface for plugins that can expose sanitized
+// internal state through operational debug endpoints.
+type StateDumper interface {
+	// DumpState returns JSON-serializable plugin state. Implementations must not
+	// include request payloads, credentials, or other sensitive values.
+	DumpState() any
```
elevran: consider making the requrn value explicit by using `json.RawMessage` - the serialization should be the responsibility of the the plugin. 
- Can use internal structs for serialization (no need to rely on exported structs fields and reflection).
- Localizes responsibility to the plugin writer.
- Explicit typing

Cosider adding some guidance with repsect to what should be in the state dump (e.g., configuration paramters, dynamic runtime state, what's the expected size, ...)

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
```diff
@@ -95,10 +97,68 @@ type InFlightLoadProducer struct {
 	dk          fwkplugin.DataKey
 }
 
+type InFlightLoadState struct {
+	Endpoints []EndpointInFlightLoadState `json:"endpoints"`
+}
+
+type EndpointInFlightLoadState struct {
```
elevran: Is there a reasonable limit on endpoint count?
If we have 100's or 1000's of endpoints, maybe only dump histograms or top-k?

pkg/epp/framework/interface/plugin/plugins.go
```diff
@@ -23,6 +23,14 @@ type Plugin interface {
 	TypedName() TypedName
 }
 
+// StateDumper is an optional interface for plugins that can expose sanitized
```
elevran: Is it clear when one should use DumpState and when expose metrics of differnet kinds?
Should be clear when each makes sense so we do not overlap, IMO.

pkg/epp/server/plugin_state_debug.go
```diff
@@ -0,0 +1,98 @@
...
+	Timestamp string                           `json:"timestamp"`
+	Plugins   map[string]pluginStateDebugEntry `json:"plugins"`
+}
+
+type pluginStateDebugEntry struct {
+	Type  string `json:"type"`
+	State any    `json:"state"`
+}
+
+func SetupPluginStateDebugHandler(mgr ctrl.Manager, plugins fwkplugin.HandlePlugins) error {
```
elevran: is there a way to avoid the `ctlr.Manager` here so it is not k8s specific?
/cc @ezrasilvera for non-Kubernetes envionments.

pkg/epp/framework/interface/plugin/plugins.go
wenhug: Updated in `707a2240`: `StateDumper.DumpState` now returns `(json.RawMessage, error)`, so plugins own serialization and can use internal structs. I also expanded the interface docs with guidance about sanitized, bounded debug state and when to prefer metrics.

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
wenhug: Added a bounded dump in `707a2240`: the in-flight producer now reports `totalEndpoints`, `maxEndpoints`, and `truncated`, and includes only the busiest endpoints up to the cap.

pkg/epp/framework/interface/plugin/plugins.go
wenhug: Clarified this in the `StateDumper` docs in `707a2240`: dumps are for on-demand debugging snapshots of dynamic state that is hard to understand from metrics alone; metrics remain preferred for time series, alerting, dashboards, and aggregation.

pkg/epp/server/plugin_state_debug.go
wenhug: Adjusted in `707a2240`: `SetupPluginStateDebugHandler` now takes a small `MetricsHandlerRegistrar` interface instead of `ctrl.Manager`, so the handler setup is no longer tied to controller-runtime.

pkg/epp/server/options.go
wenhug: Good point — updated in `707a2240` to remove the separate enable flag, always register the endpoint, and restrict direct access to loopback clients so it is usable via local/port-forward access without pre-planning.

pkg/epp/server/plugin_state_debug.go
```diff
@@ -0,0 +1,135 @@
...
+		Timestamp: nowFunc().UTC().Format(time.RFC3339Nano),
+		Plugins:   make(map[string]pluginStateDebugEntry, len(allPlugins)),
+	}
+	for name, plugin := range allPlugins {
+		if plugin == nil {
+			continue
+		}
+		dumper, ok := plugin.(fwkplugin.StateDumper)
+		if !ok {
+			continue
```
ahg-g, line 104: how about we add an entry to `response.Plugins` with a message that the plugin doesn't support state collection?

ahg-g (commented):
pls add a doc to docs/ on this feature; we nee to re-work the docs directory, but for now a docs/plugin_debug.md would be good.

pkg/epp/server/plugin_state_debug.go
wenhug: Done in `df32ad5a`: non-`StateDumper` plugins are now included with their type and a message explaining that state collection is unsupported, so operators can distinguish unsupported dumping from a missing plugin.

ahg-g:
great, pls fix the conflict and we can merge this

ahg-g:
pls fix the build issue

pkg/epp/server/plugin_state_debug.go
```diff
@@ -0,0 +1,128 @@
...
+		if !ok {
+			response.Plugins[name] = pluginStateDebugEntry{
+				Type:    plugin.TypedName().Type,
+				Message: pluginStateUnsupportedText,
+			}
+			continue
+		}
+		state, err := dumper.DumpState()
+		if err != nil {
+			return pluginStateDebugResponse{}, fmt.Errorf("plugin %q failed to dump state: %w", name, err)
```
elevran: A single plugin's `DumpState` error (and the invalid-JSON check at line 120) aborts collection and returns 500 for the entire endpoint. For an operational debug tool, one faulty plugin should not black out all others. Consider recording the error as the entry's `message` and continuing, so the response remains useful even when one plugin misbehaves.

cmd/epp/runner/runner.go
```diff
@@ -348,6 +348,13 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 		}
 	}
 
+	if r.PluginHandle == nil {
+		setupLog.Info("Plugin state debug handler not registered: plugin handle unavailable")
+	} else if err = runserver.SetupPluginStateDebugHandler(mgr, r.PluginHandle); err != nil {
```
elevran, line 352: The debug handler is registered only here, in the manager/k8s path. `runWithFileDiscovery` starts its own standalone `serveMetrics()` mux that does not implement `MetricsHandlerRegistrar`, so file-discovery users never get this endpoint. Either extend the implementation to cover that path or document the limitation in `docs/plugin_debug.md`.

elevran (commented):
Previous feedback has been addressed cleanly - thanks!

Two issues remain and are noted inline. 
Also: the test and lint jobs are skipping on the latest commit (8a6fa02), which is a merge-from-main commit. The path filter sees no source changes in that merge alone. Please rebase to trigger a full CI run.

pkg/epp/server/plugin_state_debug.go
wenhug: Updated in `193612a1`: per-plugin `DumpState` errors and invalid JSON are now reported in that plugin's `message` entry while collection continues for the rest of the response.

cmd/epp/runner/runner.go
wenhug: Documented the current limitation in `docs/plugin_debug.md` in `193612a1`: the endpoint is available in the EPP controller-manager server path, while standalone file-discovery mode uses a separate metrics mux and does not expose plugin debug state.

ahg-g:
can you pls rebase, I think this is ready to go

shmuelk:
@wenhug Please fix lint error. The rest looks good

elevran (approved):

PR #1155, opened 2026-05-15
Add gRPC transport controls

LukeAVanDrie:
**What type of PR is this?**

/kind feature

When evaluating high-load multi-modal workloads (e.g. many inlined base64 images totaling ~25MB per request), EPP rejected incoming streams due to the default 4MB gRPC receive message limit (`grpc: received message larger than max`). This PR addresses this issue by:

*   gRPC Controls: Added CLI flags `--grpc-max-recv-msg-size`, `--grpc-max-send-msg-size`, and `--grpc-enable-compression` to `Options` to allow overriding transport limits and enabling over-the-wire compression.
*   Compression Support: Registered the gzip compressor via blank import in `runserver.go`.
*   Memory Pool (`sync.Pool`): Implemented a `sync.Pool` of `*bytes.Buffer` in `StreamingServer` to recycle buffers across incoming request streams, capping memory allocations and mitigating GC thrashing.
*   Buffer Decoupling: Decoupled request and response buffering slices to prevent cross-phase contamination.

```release-note
Add CLI flags --grpc-max-recv-msg-size, --grpc-max-send-msg-size, and --grpc-enable-compression to EPP to support large multi-modal payloads and over-the-wire compression.
```

### DIFF
--- a/pkg/epp/server/options.go
+++ b/pkg/epp/server/options.go
@@ -53,8 +53,11 @@ type Options struct {
 	//
 	// ext_proc configuration.
 	//
-	GRPCPort             int  // gRPC port used for communicating with Envoy proxy. (TODO: uint16?)
-	EnableLeaderElection bool // Enables leader election for high availability
+	GRPCPort              int  // gRPC port used for communicating with Envoy proxy. (TODO: uint16?)
+	EnableLeaderElection  bool // Enables leader election for high availability
+	GRPCMaxRecvMsgSize    int  // Maximum size of a gRPC message to receive.
+	GRPCMaxSendMsgSize    int  // Maximum size of a gRPC message to send.
+	GRPCEnableCompression bool // Enables gRPC compression.
 	//
 	// InferencePool.
 	//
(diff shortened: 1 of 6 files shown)


LukeAVanDrie:
/hold

Wrapping up some manual testing (as I just ported this code from my GAIE fork here after the EPP repo migration).

pkg/epp/server/options.go
```diff
@@ -53,8 +53,11 @@ type Options struct {
 	//
 	// ext_proc configuration.
 	//
-	GRPCPort             int  // gRPC port used for communicating with Envoy proxy. (TODO: uint16?)
-	EnableLeaderElection bool // Enables leader election for high availability
+	GRPCPort              int  // gRPC port used for communicating with Envoy proxy. (TODO: uint16?)
+	EnableLeaderElection  bool // Enables leader election for high availability
+	GRPCMaxRecvMsgSize    int  // Maximum size of a gRPC message to receive.
+	GRPCMaxSendMsgSize    int  // Maximum size of a gRPC message to send.
```
ahg-g: can we make them a string to accept units, easier to set that way.

LukeAVanDrie: I accept string and parse using resource quantity now. I made a few changes which you can see in the latest commit.

pkg/epp/server/runserver.go
```diff
@@ -173,13 +178,27 @@ func (r *ExtProcServerRunner) AsRunnable(logger logr.Logger) manager.Runnable {
...
+		if creds != nil {
+			grpcOpts = append(grpcOpts, grpc.Creds(creds))
+		}
+		if r.GRPCMaxRecvMsgSize > 0 {
+			grpcOpts = append(grpcOpts, grpc.MaxRecvMsgSize(r.GRPCMaxRecvMsgSize))
+		}
+		if r.GRPCMaxSendMsgSize > 0 {
+			grpcOpts = append(grpcOpts, grpc.MaxSendMsgSize(r.GRPCMaxSendMsgSize))
+		}
+		// Note: gzip compressor is registered via blank import above.
```
ahg-g, line 193: I am not following, how is this going to be configured? is this enabled by default now?

LukeAVanDrie: In Go grpc-go, simply importing this package registers the gzip codec with the global gRPC registry at initialization time (init()). There is no manual configuration required on the EPP container. It relies entirely on automatic standard protocol negotiation between the client (Envoy) and EPP.

- Decompression: If the downstream client (or Envoy) compresses the incoming request payload and sends it to EPP, EPP's gRPC server will automatically detect the compression header and decompress it.
- Compression: By default, the grpc-go server responds using the same compressor that was used for the incoming request. If Envoy sends a compressed request, EPP will automatically compress the response back to Envoy.
- No Compression: If the client does not request compression (i.e., does not send the header or doesn't include gzip), EPP falls back to identity (uncompressed) automatically.

In short, my flag did nothing. We make gzip support universally available, and let the L7 proxy (Envoy) decide whether to negotiate and use it.

LukeAVanDrie: This piece won't help much with base64 encoded images anyways as there is not much left to compress. Might help a bit with text prompts.

Issue #866, opened 2026-04-22
[Feature Request] A Prefetch Plugin for Reducing latency when reading KV cache from disk in Tiered Storage


leipanhz:
# What would you like to be added:

A component in inference scheduler that prefetches KV-Cache from remote storage before vllm starts to process the inference work request. 

# Why is this needed:

### Current Challenge

The llm-d-inference-scheduler can be configured to read KV cache from slower storage tiers (remote cloud storage) when KV-cache information is stored in the index. On the other hand, reading from the remote cloud storage leads to high latency and directly impacts user experience and system performance. In addition, the performance is unpredictable as cache hits are fast, cache misses are slow. 

### Tiered Storage Architecture

The llm-d inference scheduler can be configured to use tiered caching:

```
Tier 1: GPU Memory (fastest, most expensive)
   ↓
Tier 2: CPU Memory (fast, moderate cost)
   ↓
Tier 3: High-performance Storage (moderate, lower cost)
   ↓
Tier 4: Remote Storage (slowest, cheapest)
```

## Desired Solution

We need a **proactive prefetch mechanism** that 

1. Identifies which KV-cache files will be needed for incoming requests and uses request metadata (prompt, model, etc.) to determine file paths. 

2. Runs prefetching in the background and not delay request routing or processing.

3. Gracfully hanles failure (best-effort)
 

## Expected Benefits

1. **Reduced TTFT** with more consistent user experience

2. **Better Resource Utilization and performance under load** by leveraging idle I/O capacity during request processing

3. **Cost Optimization** that reduces need for expensive high-tier storage

### Core Functionality

1. **Request Interception**
   - Start to prefetch before sending request to vllm pod
   - Extract information needed to identify KV-cache files

2. **File Path Resolution**
   - From the request metadata, derive filesystem paths
  

3. **Async Prefetch Workers**
   - Worker pool with configurable concurrency
   - Pre-allocated buffers to avoid allocations

4. **Configuration**
   - Enable/disable prefetch per deployment
   - Tune worker count and  buffer sizes
   - Adapt to different storage characteristics

elevran:
a few technical questions regarding the use case and suggested approach:
- is there a (standard or non standard) KV cache API we can call directly from the router? If so, what are the specific calls and functionality provided?
[...]

PR #1156, opened 2026-05-15
Add KV-cache file prefetch plugin

leipanhz:
Add KV-cache file prefetch plugin for inference requests (experimental feature)
Part 1: [changes in KV-Cache](https://github.com/llm-d/llm-d-kv-cache/pull/587)
Part 2: changes in llm-d-router (current PR)

PR Description:
Introduces a new experimental feature that aims to proactively prefetch KV-cache blocks across different storage tiers before inference requests are processed by the GPU pod. The plugin extends the precise prefix cache scorer with engine key calculation to determine the storage location (file names) of KV-cache blocks that will be needed and arrange for them to be promoted to a closer storage tier to improve inference latency. The current implementation is intended for a shared file system that includes transparent access to a remote storage tier, such as IBM Storage Scale configured to off-load cold data to remote object storage. The prefetch plugin uses a concurrent worker thread pool architecture to efficiently prefetch multiple (configurable) files in parallel from remote storage to the shared file system. In a future version of the plugin this could be extended, for example, to prefetch KV-cache blocks from the file system to CPU memory on the worker node that the request is being routed to.

For this to work correctly, the plugin must be configured to use a hash algorithm for generating engine keys that matches the algorithm used by vLLM when offloading KV-cache blocks to storage. For this purpose, this work adds a configurable hashing algorithm SHA256-CBOR to the token processor as an alternative for vLLM compatibility. The SHA256-CBOR implementation supports extra keys (multimodal features) in block hash computation. In addition, this feature relies on logic derived from the llm-d-fs-connector to generate KV file names, so it currently only works with the llm-d-fs-connector.

Changes include:

New Prefetch Plugin (prefetch_prerequest_experimental.go):

Implements PreRequest interface for pre-inference file prefetching
Converts engine keys to filesystem paths using llm-d-fs-connector format
Manages worker thread pool for concurrent file prefetching (configurable workers)
Each worker reads configurable number of blocks (BlockSize x BlockCount bytes) from KV-cache files to trigger prefetch of the rest of the file from remote storage.
Supports configurable prefetch parameters (block size, concurrency, queue size)
Precise Prefix Cache Scorer Enhancement (precise_prefix_cache.go):

Add GetEngineKeysForRequest() method to extract engine keys from requests
Support multimodal features in engine key computation
Add SHA256-CBOR hashing algorithm for token processor with extra keys support

Add configuration to choose hashing function via the field name “hashAlgorithm”: FNV64a default, SHA256-CBOR for vLLM
Implement SHA256-CBOR hashing matching vLLM engine-key computation
Extend BlockExtraFeatures for multimodal content support

**What type of PR is this?**
/kind feature

See description in https://github.com/llm-d/llm-d-router/issues/866 

https://github.com/llm-d/llm-d-router/issues/866 

Fixes #

```release-note
NONE
```


elevran:
@leipanhz can we please discuss the requirements/approach on #866 before continuing work on this PR?

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -197,6 +199,29 @@ func New(ctx context.Context, name string, config PluginConfig) (*Producer, erro
...
+}
+
+// GetEngineKeysAndDigestsForRequest returns the per-block uint64 engine keys
+// (used for index lookup and wire compatibility) alongside the full-width hash
+// digests (used by the prefetch plugin to build vLLM fs-connector filenames:
+// 32 bytes for sha256_cbor, 8 bytes for fnv64a). Both slices are flattened
+// across prompts and aligned index-for-index.
+func (p *Producer) GetEngineKeysAndDigestsForRequest(_ context.Context,
+	request *scheduling.InferenceRequest) ([]uint64, [][]byte, error) {
+	engineKeys, digests, err := computeBlockKeysWithDigests(p.tokenProcessor, request, p.blockSizeTokens)
```
bongwoobak, line 231: Enabling prefetch forces `hashAlgorithm: sha256-cbor` on the shared producer tokenProcessor, which also switches the requestKey/scoring path to sha256. 

That collapses the requestKey<-->engineKey separation and puts sha256 on the scoring/event hot path even for non-prefetch traffic. Since the router only needs vLLM-matching hashes for prefetch filenames (the index already bridges engineKey-->requestKey for lookup), 

could the engine-key/digest computation use a dedicated sha256-cbor hasher, leaving the scoring requestKey hasher independent (e.g. FNV)?

pkg/epp/framework/plugins/scheduling/scorer/prefetch/prefetch_prerequest_experimental.go
```diff
@@ -0,0 +1,454 @@
...
+//     config (Digest). The router cannot reliably recompute it, so the
+//     operator reads it from the vLLM pod and supplies it directly.
+//   - <groupIdx> is the KV-cache group index (GroupIdx, typically 0).
+//   - <rank> ranges over [0, TpSize*PpSize*PcpSize*DcpSize); every worker
+//     shares the same base prefix (file_mapper.py:117), so each rank folder
+//     is the base with "_r<rank>" appended.
+type KVFilePathBaseParams struct {
+	RootDir          string `json:"rootDir"`
+	ModelName        string `json:"modelName"`
+	Digest           string `json:"digest"`
```
bongwoobak: The digest is the fingerprint of the running vLLM's cache config, pinned statically in config here. 

Isn't it effectively per-deployment it shifts whenever the backend's model/dtype/parallelism or vLLM version changes (e.g. a pod rolled out with a new config/version)? A stale value then silently no-ops with no error. 

Could we read it dynamically when an endpoint is (re)discovered from a vLLM metric/label or similar?

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
leipanhz: Hi @bongwoobak, you are correct that "the router only needs vLLM-matching hashes for prefetch filenames". The intention of using sha256-cbor as hashingAlgorithm is to offer user options to choose a different hashing algorithm other than the default FNV and reduce the complexity of using two hashing algorithms in the router. We had similar discussions with @vMaroon earlier, adding him here for additional comments or corrections if needed. Thank you

pkg/epp/framework/plugins/scheduling/scorer/prefetch/prefetch_prerequest_experimental.go
leipanhz: > The digest is the fingerprint of the running vLLM's cache config, pinned statically in config here.
> 
> Isn't it effectively per-deployment it shifts whenever the backend's model/dtype/parallelism or vLLM version changes (e.g. a pod rolled out with a new config/version)? A stale value then silently no-ops with no error.
> 
> Could we read it dynamically when an endpoint is (re)discovered from a vLLM metric/label or similar?

You suggestion is what we are looking for in the long run, having API support from vllm to provide config information. Currently there is no API (or metrics) available to retrieve the configuration content, a separately efffor is working to address this: https://github.com/vllm-project/vllm/issues/38147. We are actively tracking this and 
following up.

Issue #1157, opened 2026-05-15
precise_prefix_cache fails to read the updated configuration: BlockSizeTokens


leipanhz:
**What happened**:
During the cluster testing, I hit a [bug in precise_prefix_cache](https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go#L255): deprecated config name BlockSize and new config name BlockSizeTokens are both used in the code. The prefix plugin is created using the old config name. Therefore, if the deployment uses new field name, it gets wrong block size.

**What you expected to happen**:
The prefix plugin should be created using BlockSizeTokens if it is configured by the user. TokenProcessor.BlockSize() should be used to retrieve block size. 

**How to reproduce it (as minimally and precisely as possible)**:
Deployment the router using the latest config name
blockSizeTokens: 64
Observe the value is not taken by the scorer, instead, it take values from old config "blockSize"

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- llm-d-scheduler version (use `git describe --tags --dirty --always` if you built from source, or specify the tag if you used a tagged version or image):
- Cloud provider or hardware configuration:
- Install tools:
- Others:

gkneighb:
Reproduced with a unit test against `PluginConfig` on current `main`: setting `blockSizeTokens: 64` at the top level of the plugin's parameters silently drops the value (the field doesn't exist on `PluginConfig`), and the scorer falls through to the default 16-token block size from `kvblock.DefaultTokenProcessorConfig()`. Setting `tokenProcessorConfig.blockSize: 64` works because that wraps the underlying `kvblock.TokenProcessorConfig` directly.

Opened #1283 which adds `BlockSizeTokens` as a top-level field that overrides `TokenProcessorConfig.BlockSize` when set, so the same config-field name now works across both prefix-cache plugins. The legacy wrapped-config path is preserved for backwards compatibility. Tests + presubmit green locally.

Note for once #1134 (strict plugin-config parsing) lands: today `blockSizeTokens` is silently dropped; under strict parsing it would be an outright error; this PR makes it work as expected — net win in either world.

gkneighb:
[...]

PR #1159, opened 2026-05-16
Fix precise-scorer to use tokenProcessor.BlockSize() instead of depre…

leipanhz:
…cated BlockSize field

The Scorer struct stored blockSizeTokens initialized from config.TokenProcessorConfig.BlockSize, which is deprecated in llm-d-kv-cache. Change to use the tokenProcessor.BlockSize() method to correctly resolves the block size.

**What type of PR is this?**
/kind bug

The Scorer struct in precise prefix cache stored blockSizeTokens initialized from config.TokenProcessorConfig.BlockSize, which is deprecated in llm-d-kv-cache. As a result, it doesn't read values of BlockSizeTokens by user configuration. 

https://github.com/llm-d/llm-d-router/issues/1157

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
+++ b/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
@@ -244,17 +244,18 @@ func New(ctx context.Context, config PluginConfig) (*Scorer, error) {
 	}
 
 	return &Scorer{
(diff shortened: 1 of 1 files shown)


ahg-g:
@vMaroon is this PR still relevant?

vMaroon:
The bug was still there, but considering that Lei is on vacation and the fix can be much smaller with llm-d-kv-cache patch release, we can close this in favor of #1323 (Lei is co-authored).

PR #1161, opened 2026-05-16
Rename EPP-managed headers to x-llm-d

### LINKED ISSUE #1152: [Migration] rename all headers managed by epp to have the prefix x-llm-d

Since we consolidated the EPP code under llm-d-router, we should also rename the header managed by epp. We should do that in a backward compatible manner, keep supporting the old headers for a couple of releases.

The full list is at https://github.com/llm-d/llm-d/blob/main/docs/api-reference/epp-http-headers.md

**Why is this needed**:

Alignment withe the epp's new home.

wenhug:
## Summary

Fixes #1152.

This migrates llm-d-managed EPP headers to canonical `x-llm-d-*` names while preserving old names as backward-compatible aliases for reads.

Canonical mappings:

| Old header | New header |
| --- | --- |
| `x-gateway-inference-fairness-id` | `x-llm-d-inference-fairness-id` |
| `x-gateway-inference-objective` | `x-llm-d-inference-objective` |
| `x-gateway-model-name-rewrite` | `x-llm-d-model-name-rewrite` |
| `x-slo-ttft-ms` | `x-llm-d-slo-ttft-ms` |
| `x-slo-tpot-ms` | `x-llm-d-slo-tpot-ms` |

Compatibility behavior:

- read paths accept both new and old names for migrated headers
- if both are present, the new canonical name wins deterministically
- EPP strips/ignores both new and old migrated system/control headers where system headers are already stripped
- SLO headers are treated as EPP input-control headers after this migration, so both new and old SLO names are consumed by EPP and stripped from backend-forwarded request headers

## Notes for review

Per maintainer feedback, the GAIE endpoint picker protocol headers remain unchanged and are not aliased or re-emitted under `x-llm-d-*`:

- `x-gateway-destination-endpoint-subset`
- `x-gateway-destination-endpoint`
- `x-gateway-destination-endpoint-served`

The existing dropped-reason response header also remains unchanged as `x-llm-d-request-dropped-reason` and does not emit an extra compatibility alias because it has not been released yet.

## Tests

- `git diff --check`
- `go test ./pkg/common/error ./pkg/epp/...`
- `golangci-lint run --config=./.golangci.yml` for touched packages

### DIFF
--- a/pkg/epp/handlers/request.go
+++ b/pkg/epp/handlers/request.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"maps"
 	"strconv"
+	"strings"
 	"time"
 
 	configPb "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
@@ -47,16 +48,12 @@ func (s *StreamingServer) HandleRequestHeaders(ctx context.Context, reqCtx *Requ
 	}
 
 	for _, header := range req.RequestHeaders.Headers.Headers {
-		reqCtx.Request.Headers[header.Key] = envoy.GetHeaderValue(header)
-		switch header.Key {
-		case metadata.FlowFairnessIDKey:
-			reqCtx.FairnessID = reqCtx.Request.Headers[header.Key]
-		case metadata.ObjectiveKey:
-			reqCtx.ObjectiveKey = reqCtx.Request.Headers[header.Key]
-		case metadata.ModelNameRewriteKey:
-			reqCtx.TargetModelName = reqCtx.Request.Headers[header.Key]
-		}
+		reqCtx.Request.Headers[strings.ToLower(header.Key)] = envoy.GetHeaderValue(header)
 	}
+
+	reqCtx.FairnessID, _ = metadata.GetLowerCaseHeaderValue(reqCtx.Request.Headers, metadata.FlowFairnessIDKey)
+	reqCtx.ObjectiveKey, _ = metadata.GetLowerCaseHeaderValue(reqCtx.Request.Headers, metadata.ObjectiveKey)
+	reqCtx.TargetModelName, _ = metadata.GetLowerCaseHeaderValue(reqCtx.Request.Headers, metadata.ModelNameRewriteKey)
 
 	if reqCtx.FairnessID == "" {
 		reqCtx.FairnessID = metadata.DefaultFairnessID
--- a/pkg/epp/metadata/consts.go
+++ b/pkg/epp/metadata/consts.go
@@ -16,6 +16,8 @@ limitations under the License.
 
 package metadata
 
+import "strings"
+
 const (
 	// SubsetFilterNamespace is the key for the outer namespace struct in the metadata field of the extproc request that is used to wrap the subset filter.
 	SubsetFilterNamespace = "envoy.lb.subset_hint"
@@ -29,14 +31,61 @@ const (
 	// DestinationEndpointServedKey is the metadata key used by Envoy to specify the endpoint that served the request.
 	DestinationEndpointServedKey = "x-gateway-destination-endpoint-served"
 	// FlowFairnessIDKey is the header key used to pass the fairness ID to be used in Flow Control.
-	FlowFairnessIDKey = "x-gateway-inference-fairness-id"
+	FlowFairnessIDKey = "x-llm-d-inference-fairness-id"
+	// OldFlowFairnessIDKey is the deprecated alias for FlowFairnessIDKey.
+	OldFlowFairnessIDKey = "x-gateway-inference-fairness-id"
 	// ObjectiveKey is the header key used to specify the objective of an incoming request.
-	ObjectiveKey = "x-gateway-inference-objective"
+	ObjectiveKey = "x-llm-d-inference-objective"
+	// OldObjectiveKey is the deprecated alias for ObjectiveKey.
+	OldObjectiveKey = "x-gateway-inference-objective"
 	// ModelNameRewriteKey is the header key used to specify the model name to be used when the request is forwarded to the model server.
-	ModelNameRewriteKey = "x-gateway-model-name-rewrite"
+	ModelNameRewriteKey = "x-llm-d-model-name-rewrite"
+	// OldModelNameRewriteKey is the deprecated alias for ModelNameRewriteKey.
+	OldModelNameRewriteKey = "x-gateway-model-name-rewrite"
+	// TTFTSLOHeaderKey is the header key used to specify the time-to-first-token SLO in milliseconds.
+	TTFTSLOHeaderKey = "x-llm-d-slo-ttft-ms"
+	// OldTTFTSLOHeaderKey is the deprecated alias for TTFTSLOHeaderKey.
+	OldTTFTSLOHeaderKey = "x-slo-ttft-ms"
+	// TPOTSLOHeaderKey is the header key used to specify the time-per-output-token SLO in milliseconds.
+	TPOTSLOHeaderKey = "x-llm-d-slo-tpot-ms"
+	// OldTPOTSLOHeaderKey is the deprecated alias for TPOTSLOHeaderKey.
+	OldTPOTSLOHeaderKey = "x-slo-tpot-ms"
 
 	// DefaultFairnessID is the default fairness ID used when no ID is provided in the request.
 	// This ensures that requests without explicit fairness identifiers are still grouped and managed by the Flow Control
 	// system.
 	DefaultFairnessID = "default-flow"
 )
+
+// All headerAliases keys and values must be lower case.
+var headerAliases = map[string][]string{
+	FlowFairnessIDKey:   {FlowFairnessIDKey, OldFlowFairnessIDKey},
+	ObjectiveKey:        {ObjectiveKey, OldObjectiveKey},
+	ModelNameRewriteKey: {ModelNameRewriteKey, OldModelNameRewriteKey},
+	TTFTSLOHeaderKey:    {TTFTSLOHeaderKey, OldTTFTSLOHeaderKey},
+	TPOTSLOHeaderKey:    {TPOTSLOHeaderKey, OldTPOTSLOHeaderKey},
+}
+
+// HeaderNames returns the current header name followed by deprecated aliases.
+func HeaderNames(key string) []string {
+	key = strings.ToLower(key)
+	if names, ok := headerAliases[key]; ok {
+		return names
+	}
+	return []string{key}
+}
+
+// GetLowerCaseHeaderValue returns a header value from a lower-case header map using the current name first, then deprecated aliases.
+func GetLowerCaseHeaderValue(headers map[string]string, key string) (string, bool) {
+	key = strings.ToLower(key)
+	if aliases, ok := headerAliases[key]; ok {
+		for _, alias := range aliases {
+			if value, ok := headers[alias]; ok {
+				return value, true
+			}
+		}
+		return "", false
+	}
+	value, ok := headers[key]
+	return value, ok
+}
--- a/pkg/epp/util/request/headers.go
+++ b/pkg/epp/util/request/headers.go
@@ -21,24 +21,30 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/sets"
 
+	errcommon "github.com/llm-d/llm-d-router/pkg/common/error"
 	"github.com/llm-d/llm-d-router/pkg/epp/metadata"
 )
 
 var (
 	// InputControlHeaders are sent by the Gateway/User to control EPP behavior.
 	// We must extract these, then strip them so they don't leak to the backend.
-	InputControlHeaders = sets.New(
-		strings.ToLower(metadata.FlowFairnessIDKey),
-		strings.ToLower(metadata.ObjectiveKey),
-		strings.ToLower(metadata.ModelNameRewriteKey),
-		strings.ToLower(metadata.SubsetFilterKey),
+	InputControlHeaders = lowerHeaderNames(
+		metadata.FlowFairnessIDKey,
+		metadata.ObjectiveKey,
+		metadata.ModelNameRewriteKey,
+		metadata.SubsetFilterKey,
+		metadata.TTFTSLOHeaderKey,
+		metadata.TPOTSLOHeaderKey,
 	)
 
 	// OutputInjectionHeaders are headers EPP injects for the backend.
 	// If the user sends these, they must be stripped to prevent ambiguity.
-	OutputInjectionHeaders = sets.New(
-		strings.ToLower(metadata.DestinationEndpointKey),
-		strings.ToLower(metadata.DestinationEndpointServedKey),
+	OutputInjectionHeaders = addLowerHeaders(
+		lowerHeaderNames(
+			metadata.DestinationEndpointKey,
+			metadata.DestinationEndpointServedKey,
+		),
+		errcommon.RequestDroppedReasonHeaderKey,
 	)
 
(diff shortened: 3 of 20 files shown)


pkg/common/error/error.go
```diff
@@ -46,6 +50,13 @@ const (
 	RequestDroppedReasonEvictedPriority      RequestDroppedReason = "evicted-priority"
 )
 
+func RequestDroppedReasonHeaders(reason RequestDroppedReason) map[string]string {
+	return map[string]string{
+		RequestDroppedReasonHeaderKey:    string(reason),
+		OldRequestDroppedReasonHeaderKey: string(reason),
+	}
+}
```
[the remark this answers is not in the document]
wenhug: Added a godoc comment for RequestDroppedReasonHeaders.

pkg/epp/util/request/headers.go
```diff
@@ -21,24 +21,31 @@ import (
...
-		strings.ToLower(metadata.ModelNameRewriteKey),
-		strings.ToLower(metadata.SubsetFilterKey),
+	InputControlHeaders = lowerHeaderNames(
+		metadata.FlowFairnessIDKey,
+		metadata.ObjectiveKey,
+		metadata.ModelNameRewriteKey,
+		metadata.SubsetFilterKey,
+		metadata.TTFTSLOHeaderKey,
+		metadata.TPOTSLOHeaderKey,
 	)
```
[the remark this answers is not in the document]
wenhug: Confirmed intended. SLO headers are now treated as EPP input-control headers; I updated the PR description to call out that both canonical and deprecated SLO names are consumed by EPP and stripped from backend-forwarded request headers.

config/charts/standalone/values.yaml
```diff
@@ -217,7 +217,7 @@ inferenceExtension:
                           max_requests: 40000
                     original_dst_lb_config:
                       use_http_header: true
-                      http_header_name: x-gateway-destination-endpoint
+                      http_header_name: x-llm-d-router-destination-endpoint
```
ahg-g: This is part of the protocol gateway api inference extension protocol, we shouldn't/can't change it.

pkg/common/error/error.go
```diff
@@ -25,9 +25,13 @@ import (
...
 
-// RequestDroppedReasonHeaderKey is the HTTP response header that communicates the specific
-// reason a request was dropped by flow control.
-const RequestDroppedReasonHeaderKey = "x-llm-d-request-dropped-reason"
+const (
+	// RequestDroppedReasonHeaderKey is the HTTP response header that communicates the specific
+	// reason a request was dropped by flow control.
+	RequestDroppedReasonHeaderKey = "x-llm-d-router-request-dropped-reason"
+	// OldRequestDroppedReasonHeaderKey is the deprecated alias for RequestDroppedReasonHeaderKey.
+	OldRequestDroppedReasonHeaderKey = "x-llm-d-request-dropped-reason"
```
ahg-g: this has not been released, we don't need to keep backward compatibility for

ahg-g (commented):
Thanks a lot, sorry, the issue mistakenly suggested `x-llm-d-router`, i meant `x-llm-d` only. I will update the issue.

config/charts/standalone/values.yaml
ahg-g: same goes to `x-gateway-destination-endpoint-subset` and `x-gateway-destination-endpoint-served`; these are defined as part of the gaie endpoint picker protocol (see https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/docs/proposals/004-endpoint-picker-protocol)

pkg/epp/metadata/consts.go
```diff
@@ -16,27 +16,145 @@ limitations under the License.
...
+
 const (
 	// SubsetFilterNamespace is the key for the outer namespace struct in the metadata field of the extproc request that is used to wrap the subset filter.
 	SubsetFilterNamespace = "envoy.lb.subset_hint"
 	// SubsetFilterKey is the metadata key used by Envoy to specify an array candidate pods for serving the request.
 	// If not specified, all the pods that are associated with the pool are candidates.
-	SubsetFilterKey = "x-gateway-destination-endpoint-subset"
+	SubsetFilterKey = "x-llm-d-router-destination-endpoint-subset"
+	// OldSubsetFilterKey is the deprecated alias for SubsetFilterKey.
+	OldSubsetFilterKey = "x-gateway-destination-endpoint-subset"
```
ahg-g: we shouldn't/can't change this

pkg/epp/metadata/consts.go
```diff
@@ -16,27 +16,145 @@ limitations under the License.
...
 	// DestinationEndpointNamespace is the key for the outer namespace struct in the metadata field of the extproc response that is used to wrap the target endpoint.
 	DestinationEndpointNamespace = "envoy.lb"
 	// DestinationEndpointKey is the header and response metadata key used by Envoy to route to the appropriate pod.
-	DestinationEndpointKey = "x-gateway-destination-endpoint"
+	DestinationEndpointKey = "x-llm-d-router-destination-endpoint"
+	// OldDestinationEndpointKey is the deprecated alias for DestinationEndpointKey.
+	OldDestinationEndpointKey = "x-gateway-destination-endpoint"
```
ahg-g: we shouldn't/can't change this

pkg/epp/metadata/consts.go
```diff
@@ -16,27 +16,145 @@ limitations under the License.
...
 	// DestinationEndpointServedKey is the metadata key used by Envoy to specify the endpoint that served the request.
-	DestinationEndpointServedKey = "x-gateway-destination-endpoint-served"
+	DestinationEndpointServedKey = "x-llm-d-router-destination-endpoint-served"
+	// OldDestinationEndpointServedKey is the deprecated alias for DestinationEndpointServedKey.
+	OldDestinationEndpointServedKey = "x-gateway-destination-endpoint-served"
```
ahg-g: we shouldn't/can't change this

config/charts/standalone/values.yaml
wenhug: Done — reverted the Envoy config to keep using the GAIE protocol header `x-gateway-destination-endpoint`.

pkg/common/error/error.go
wenhug: Done — kept `x-llm-d-request-dropped-reason` unchanged and removed the extra compatibility alias emission.

pkg/epp/metadata/consts.go
wenhug: Done — kept `x-gateway-destination-endpoint-subset` unchanged as a GAIE endpoint picker protocol key.

pkg/epp/metadata/consts.go
wenhug: Done — kept `x-gateway-destination-endpoint` unchanged as a GAIE endpoint picker protocol key.

pkg/epp/metadata/consts.go
wenhug: Done — kept `x-gateway-destination-endpoint-served` unchanged as a GAIE endpoint picker protocol key.

pkg/epp/metadata/consts.go
```diff
@@ -29,14 +31,123 @@ const (
...
+
+// GetHeader returns a header value using the current name first, then deprecated aliases.
+func GetHeader(headers map[string]string, key string) string {
+	value, _ := GetHeaderValue(headers, key)
+	return value
+}
+
+// GetHeaderValue returns a header value using the current name first, then deprecated aliases.
+func GetHeaderValue(headers map[string]string, key string) (string, bool) {
+	for _, name := range HeaderNames(key) {
```
ahg-g: I recommend to replace the function (which does frequent allocations) with a static map, basically the one you have above would be:

```
var headerAliases = map[string][]string{
	FlowFairnessIDKey:   {FlowFairnessIDKey, OldFlowFairnessIDKey},
	ObjectiveKey:        {ObjectiveKey, OldObjectiveKey},
	ModelNameRewriteKey: {ModelNameRewriteKey, OldModelNameRewriteKey},
	TTFTSLOHeaderKey:    {TTFTSLOHeaderKey, OldTTFTSLOHeaderKey},
	TPOTSLOHeaderKey:    {TPOTSLOHeaderKey, OldTPOTSLOHeaderKey},
}
```

and here the code becomes something like:

```
for _, name := range headerAliases[key]
```

pkg/epp/handlers/server.go
```diff
@@ -499,15 +499,17 @@ func (r *RequestContext) updateStateAndSendIfNeeded(srv extProcPb.ExternalProces
...
 		if r.RequestDroppedReason != "" {
-			ir.Headers = &extProcPb.HeaderMutation{
-				SetHeaders: []*configPb.HeaderValueOption{
-					{
-						Header: &configPb.HeaderValue{
-							Key:      errcommon.RequestDroppedReasonHeaderKey,
-							RawValue: []byte(r.RequestDroppedReason),
-						},
+			setHeaders := make([]*configPb.HeaderValueOption, 0, 2)
+			for key, value := range errcommon.RequestDroppedReasonHeaders(r.RequestDroppedReason) {
```
ahg-g: pls revert this change, the dropped reason headers have not been released yet (they were added last week), so we can change the header name without being concerned about backward compatibility.

pkg/epp/framework/plugins/requestcontrol/test/responsereceived/destination_endpoint_served_verifier.go
```diff
@@ -83,7 +83,9 @@ func (desv *DestinationEndpointServedVerifier) ResponseHeader(ctx context.Contex
 		return
 	}
 
-	actualEndpoint, ok := lbMetadata[metadata.DestinationEndpointServedKey].(string)
+	actualEndpointRaw, ok := metadata.GetValue(lbMetadata, metadata.DestinationEndpointServedKey)
+	actualEndpoint, stringOK := actualEndpointRaw.(string)
+	ok = ok && stringOK
```
ahg-g: I wasn't expecting a change to this file, this plugin doesn't access headers.

pkg/common/error/error.go
```diff
@@ -46,6 +46,13 @@ const (
 	RequestDroppedReasonEvictedPriority      RequestDroppedReason = "evicted-priority"
 )
 
+// RequestDroppedReasonHeaders returns the dropped-reason response header for reason.
+func RequestDroppedReasonHeaders(reason RequestDroppedReason) map[string]string {
```
ahg-g: Why do we need this function? This PR shouldn't be concerned with anything related to the dropped reason header.

pkg/epp/metadata/consts.go
wenhug: Done — changed the alias table to static canonical+alias slices and made GetHeader/GetValue iterate those slices directly, avoiding per-call HeaderNames allocations.

pkg/epp/handlers/server.go
wenhug: Done — reverted the dropped-reason response handling back to the direct single-header implementation from main.

pkg/epp/framework/plugins/requestcontrol/test/responsereceived/destination_endpoint_served_verifier.go
wenhug: Done — reverted this verifier to directly read the GAIE dynamic metadata key instead of using the header alias helper.

pkg/common/error/error.go
wenhug: Done — removed RequestDroppedReasonHeaders and reverted dropped-reason call sites/tests to direct single-header maps so this PR no longer adds dropped-reason compatibility plumbing.

pkg/epp/handlers/request.go
```diff
@@ -48,16 +48,12 @@ func (s *StreamingServer) HandleRequestHeaders(ctx context.Context, reqCtx *Requ
 
 	for _, header := range req.RequestHeaders.Headers.Headers {
 		reqCtx.Request.Headers[header.Key] = envoy.GetHeaderValue(header)
```
ahg-g: I think this is bug, we should store the key using lower case here: 

```suggestion
		reqCtx.Request.Headers[strings.ToLower(header.Key)] = envoy.GetHeaderValue(header)
```

pkg/epp/metadata/consts.go
```diff
@@ -29,14 +31,106 @@ const (
...
 )
+
+var headerAliases = map[string][]string{
+	FlowFairnessIDKey:   {FlowFairnessIDKey, OldFlowFairnessIDKey},
+	ObjectiveKey:        {ObjectiveKey, OldObjectiveKey},
+	ModelNameRewriteKey: {ModelNameRewriteKey, OldModelNameRewriteKey},
+	TTFTSLOHeaderKey:    {TTFTSLOHeaderKey, OldTTFTSLOHeaderKey},
+	TPOTSLOHeaderKey:    {TPOTSLOHeaderKey, OldTPOTSLOHeaderKey},
+}
+
```
ahg-g, line 68: If we assume that the header map keys are always lower case (see comment in request.go above), then wouldn't the following logic be sufficient (note that the code will always lookup the header using what this is calling the canonical key):


```
// All keys must be lower case.
var headerAliases = map[string][]string{
	FlowFairnessIDKey:   {FlowFairnessIDKey, OldFlowFairnessIDKey},
	ObjectiveKey:        {ObjectiveKey, OldObjectiveKey},
	ModelNameRewriteKey: {ModelNameRewriteKey, OldModelNameRewriteKey},
	TTFTSLOHeaderKey:    {TTFTSLOHeaderKey, OldTTFTSLOHeaderKey},
	TPOTSLOHeaderKey:    {TPOTSLOHeaderKey, OldTPOTSLOHeaderKey},
}

// GetLowerCaseHeaderValue returns a header value assuming the key is lower case using the current name first, then deprecated aliases.
func GetLowerCaseHeaderValue(headers map[string]string, key string) (string, bool) {
    key = strings.ToLower(key)
	if aliases, ok := headerAliases[key]; ok {
	  for _, a := range aliases {
        if value, ok := headers[a]; ok {
          return value, true
        }
      }
	}
   return headers[key]
}
```

pkg/epp/handlers/request.go
wenhug: Done — request headers are now normalized to lower case when they are stored in reqCtx.Request.Headers, so later EPP lookups can treat the internal map as lower-case.

pkg/epp/metadata/consts.go
wenhug: Done — simplified the metadata helper around that invariant: headerAliases is now the static lower-case canonical-to-alias table, GetLowerCaseHeaderValue does direct map lookups only, and the old case-insensitive scan/canonicalization helpers were removed.

pkg/epp/util/request/headers.go
```diff
@@ -21,24 +21,30 @@ import (
...
-		strings.ToLower(metadata.ObjectiveKey),
-		strings.ToLower(metadata.ModelNameRewriteKey),
-		strings.ToLower(metadata.SubsetFilterKey),
+	InputControlHeaders = lowerHeaderNames(
+		metadata.FlowFairnessIDKey,
+		metadata.ObjectiveKey,
+		metadata.ModelNameRewriteKey,
+		metadata.SubsetFilterKey,
+		metadata.TTFTSLOHeaderKey,
+		metadata.TPOTSLOHeaderKey,
```
ahg-g, line 37: We should treat this the same as below (i.e., call addLowerHeaders to add the aliases), in fact those are the headers that have aliases, not the output ones below.

ahg-g: we can do this as a followup

PR #1163, opened 2026-05-16
generic read attribute function

nirrozenbaum:
**What type of PR is this?**
/kind cleanup

generic read attribute function. as the attributemap will start to get used across the code, each caller will have to do the same type conversion/assertion. This new function aims to do it once and reuse it across the code.
in current PR only one place was calling AttributesMap.Get, and that place was updated.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/interface/datalayer/attributemap.go
+++ b/pkg/epp/framework/interface/datalayer/attributemap.go
@@ -89,3 +89,21 @@ func (a *Attributes) Clone() AttributeMap {
 	})
 	return clone
 }
+
+// ReadAttribute retrieves attribute with the given key from AttributeMap and asserts it to type T.
+// Second return value is 'false' if the key is not found or the type assertion fails.
+func ReadAttribute[T Cloneable](attributeMap AttributeMap, key string) (T, bool) {
+	var zero T
+
+	raw, ok := attributeMap.Get(key)
+	if !ok {
+		return zero, false
+	}
+
+	val, ok := raw.(T)
+	if !ok {
+		return zero, false
+	}
+
+	return val, true
+}
--- a/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go
+++ b/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go
@@ -108,12 +108,8 @@ func (d *detector) Consumes() map[string]any {
 }
 
 func (d *detector) getLoad(m datalayer.AttributeMap) *attrconcurrency.InFlightLoad {
-	if val, ok := m.Get(attrconcurrency.InFlightLoadKey); ok {
-		if load, ok := val.(*attrconcurrency.InFlightLoad); ok {
-			return load
-		}
-	}
-	return &attrconcurrency.InFlightLoad{}
+	val, _ := datalayer.ReadAttribute[*attrconcurrency.InFlightLoad](m, attrconcurrency.InFlightLoadKey)
+	return val
 }
 
 // Saturation calculates the saturation level of the pool.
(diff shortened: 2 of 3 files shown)


pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go
```diff
@@ -108,12 +108,8 @@ func (d *detector) Consumes() map[string]any {
...
 
 func (d *detector) getLoad(m datalayer.AttributeMap) *attrconcurrency.InFlightLoad {
-	if val, ok := m.Get(attrconcurrency.InFlightLoadKey); ok {
-		if load, ok := val.(*attrconcurrency.InFlightLoad); ok {
-			return load
-		}
-	}
-	return &attrconcurrency.InFlightLoad{}
+	val, _ := datalayer.ReadAttribute[*attrconcurrency.InFlightLoad](m, attrconcurrency.InFlightLoadKey)
+	return val
```
roytman: The method changes behavior on the failure path:
- before - it returned a safe empty struct
- now- it returns `nil`

`ReadAttribute` returns `var zero T`, which for `T = *attrconcurrency.InFlightLoad` is nil. Both callers dereference the returned pointer's fields without a `nil `check:

`Saturation()` accesses `load.Tokens` and load.Requests (lines: 130 and 133)
`Filter()` accesses `load.Tokens` and `load.Requests` (lines: 168 and 172)
If an endpoint's AttributeMap doesn't contain the key (or has the wrong type), this will panic at runtime.


```suggestion
	val, ok := datalayer.ReadAttribute[*attrconcurrency.InFlightLoad](m, attrconcurrency.InFlightLoadKey)
	if !ok {
        return &attrconcurrency.InFlightLoad{}
    }
	return val
```

pkg/epp/framework/interface/datalayer/attributemap.go
```diff
@@ -89,3 +89,21 @@ func (a *Attributes) Clone() AttributeMap {
 	})
 	return clone
 }
+
+// ReadAttribute retrieves attribute with the given key from AttributeMap and asserts it to type T.
+// Second return value is 'false' if the key is not found or the type assertion fails.
+func ReadAttribute[T Cloneable](attributeMap AttributeMap, key string) (T, bool) {
```
roytman, line 95: > "generic read attribute function. as the attributemap will start to get used across the code" 

nit: maybe make sense to add a unittest with covering: successful retrieval, missing key, and type mismatch.

roytman (commented):
nit: an edge case with `nil` `Clone()` return. 
If a `nil *InFlightLoad` is stored in the map, `Clone()` returns `nil`, which Get returns as `(nil, true)`. Then `raw.(T)` on a nil interface value fails the assertion, so `ReadAttribute` returns `(zero, false)`. 
This is technically correct, but it means that a stored `nil` value looks indistinguishable from a missing key. Worth a doc note if this distinction matters for future callers.

nirrozenbaum:
> nit: an edge case with `nil` `Clone()` return. If a `nil *InFlightLoad` is stored in the map, `Clone()` returns `nil`, which Get returns as `(nil, true)`. Then `raw.(T)` on a nil interface value fails the assertion, so `ReadAttribute` returns `(zero, false)`. This is technically correct, but it means that a stored `nil` value looks indistinguishable from a missing key. Worth a doc note if this distinction matters for future callers.

@roytman not sure how's that possible, nil is checked before storing values in the map:
https://github.com/llm-d/llm-d-router/blob/5c4fa4ce0c4b4b55d437bde4b3c7520d7c7b7780/pkg/epp/framework/interface/datalayer/attributemap.go#L49-L53

pkg/epp/framework/interface/datalayer/attributemap.go
nirrozenbaum: added

pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go
nirrozenbaum: fixed to keep existing behavior.

roytman:
> 
> @roytman not sure how's that possible, nil is checked before storing values in the map:
> 
> https://github.com/llm-d/llm-d-router/blob/5c4fa4ce0c4b4b55d437bde4b3c7520d7c7b7780/pkg/epp/framework/interface/datalayer/attributemap.go#L49-L53

Sorry, my mistake.

PR #1166, opened 2026-05-16
chore: update Helm deploy for renaming llm-d-inference-scheduler to llm-d-router

### LINKED ISSUE #1165: [Migration]: rename llm-d-inference-scheduler references in Helm charts
As per #929, the Helm chart templates, helpers scripts, and values still reference
the old `llm-d-inference-scheduler` name.

Related:
- #929 (repo rename)
- #1103 (Go module rename)
- #1085 (Docker image rename)

zdtsw:
**What type of PR is this?**

/kind cleanup

- update Helm deploy for renaming llm-d-inference-scheduler to llm-d-router
- bump GW API version to match go.mod
- remove calls to fetch GIE CRD from remote

Fixes #1165

**Test**
run `make verify-helm-charts`

```release-note
NONE
```


hexfusion (commented):
LGTM thanks @zdtsw !

hack/verify-helm.sh
```diff
@@ -15,9 +15,9 @@
 # limitations under the License.
 
 SCRIPT_ROOT=$(dirname "${BASH_SOURCE}")/..
-GATEWAY_API_VERSION="${GATEWAY_API_VERSION:-v1.3.0}"
+GATEWAY_API_VERSION="${GATEWAY_API_VERSION:-v1.5.1}"
 GKE_GATEWAY_API_VERSION="${GKE_GATEWAY_API_VERSION:-v1.4.0}"
-GIE_VERSION="${GIE_VERSION:-v1.4.0}"
+GIE_VERSION="${GIE_VERSION:-v1.5.0}"
```
nirrozenbaum: why do we still have here GIE?

hexfusion: do we still need GIE for InferencePool fetch? to your point we need to ensure the other Inference* crds are not pulling remote.

nirrozenbaum: yes,used wrong wording. 
I meant no GIE is needed for the migrated CRDs 👍🏻

zdtsw: did a bit change , removed the fucntion and the calls, 
now it only relies on 4 CRDs from  local "config/crd"
i can open another PR to follow up similar cleanup in `verify-manifests.sh` later

zdtsw: https://github.com/llm-d/llm-d-router/pull/1170

config/charts/epplib/templates/_helpers.tpl
```diff
@@ -57,7 +57,7 @@ If inferenceExtension.monitoring.provider.name is unset/empty, default to
 prometheusoperator. For backwards compatibility, provider.name=gke still maps
 to gmp when no monitoring provider is explicitly set.
 */}}
-{{- define "llm-d-inference-scheduler.inferenceExtension.monitoring.provider.name" -}}
+{{- define "llm-d-router.inferenceExtension.monitoring.provider.name" -}}
```
ahg-g, line 60: I would drop the `inferenceExtension` part, this is not consistent with how we are naming other variables.

config/charts/epplib/templates/_helpers.tpl
```diff
@@ -79,10 +79,10 @@ When inferenceExtension.monitoring.provider.name is unset/empty, use defaults.
 For backwards compatibility, provider.gke.autopilot is still honored when
 provider.name=gke and no monitoring provider is explicitly set.
 */}}
-{{- define "llm-d-inference-scheduler.inferenceExtension.monitoring.provider" -}}
+{{- define "llm-d-router.inferenceExtension.monitoring.provider" -}}
```
ahg-g, line 82: ditto

ahg-g (commented):
I think we will want to rename some of those variables to use llm-d-endpoint-picker prefix instead of llm-d-router when the coordinator service becomes part of the router.

config/charts/epplib/templates/_helpers.tpl
ahg-g: can be cleaned up as a followup.

zdtsw: follw up in https://github.com/llm-d/llm-d-router/pull/1196

PR #1168, opened 2026-05-16
feat(epp/scheduling): support fallback scheduling profiles

### LINKED ISSUE #1139: Scheduling profiles: support fallback profiles instead of only weighted linear combination
## Motivation

Today, a scheduling profile combines multiple scorers via a weighted linear combination:
`finalScore = w1 * scorer1 + w2 * scorer2 + ...`. This works well when all scorers produce
meaningful results, but it forces plugins to embed fallback logic internally when their
primary signal is unavailable.

For example, the `latency-scorer` plugin includes `compositeKVWeight`, `compositeQueueWeight`,
and `compositePrefixWeight` parameters that define a built-in fallback when predictions are
unavailable (sidecar down or not yet trained). This fallback logic does not belong inside the
scorer — it conflates two independent scoring strategies into a single plugin and makes
configuration harder to understand.

Instead, we should be able to express this as two separate profiles in the config: a primary
profile that uses latency predictions, and a fallback profile that uses KV cache utilization,
queue depth, and prefix cache scores with explicit weights. The framework would try the
primary profile first and, if it produces no results (e.g., all scorers return zero or
indicate no data), automatically run the fallback profile.

## Proposal

Add support for **fallback scheduling profiles** in the `EndpointPickerConfig`. If a
profile's scorers produce no usable results (e.g., all scores are zero, or a scorer
explicitly signals "no data available"), the framework falls through to the next profile in
the chain.

### Example config

```yaml
schedulingProfiles:
- name: latency-prediction
  plugins:
  - pluginRef: predicted-latency-producer
  - pluginRef: prefix-cache-affinity-filter
  - pluginRef: latency-scorer
    weight: 1
  - pluginRef: weighted-random-picker

- name: composite-fallback
  fallbackFor: latency-prediction
  plugins:
    weight: 1
  - pluginRef: queue-depth-scorer
    weight: 1
    weight: 1
  - pluginRef: weighted-random-picker
```

When `latency-prediction` produces no results (predictions unavailable), the framework
runs `composite-fallback` instead. Each profile is a self-contained pipeline with its own
scorers, weights, filters, and picker.

## Benefits

1. **Separation of concerns**: each profile is a clean, independent scoring strategy. No
   plugin needs to embed fallback logic for unrelated signals.
2. **User-facing clarity**: the configmap explicitly shows what runs when, rather than
   hiding fallback behavior behind composite weight parameters inside a plugin.
3. **Composability**: users can define arbitrary fallback chains without modifying plugin
   code. New scorers can be added to either profile independently.
4. **Simplifies plugin implementation**: the `latency-scorer` can drop its composite
   fallback parameters (`compositeKVWeight`, `compositeQueueWeight`,
   `compositePrefixWeight`), reducing config surface and code complexity.

## Suggested "no data" detection

A simple heuristic: if all endpoints receive a final weighted score of 0 from the profile's
scorers, treat it as "no usable data" and trigger the fallback. This avoids changes to the
scorer interface (no sentinel values or extra return types needed). A profile that scores
everything 0 is effectively a no-op — it provides no differentiation — so falling through
to the fallback is always the right behavior.

In practice this is robust: even in worst-case scenarios, endpoints differ in queue depth,
KV cache utilization, or prefix cache state, so a fallback profile using those signals will
produce non-zero differentiation.

## Design consideration: fallback as a framework feature, not a plugin

Fallback should be a **framework-level concept** built into the scheduling framework itself,
not implemented as a `ProfileHandler` plugin. While the current `ProfileHandler` interface
could technically support fallback logic (the scheduler loop already calls `Pick()`
iteratively with previous results), making fallback a plugin responsibility has drawbacks:

- It couples fallback semantics to the profile handler implementation, meaning every custom
  `ProfileHandler` would need to re-implement fallback logic.
- The `ProfileHandler` is meant to decide *which* profiles to run and *how to aggregate*
  results — fallback is a lower-level concern about individual profile failure.

Instead, fallback should be expressed as a field on `SchedulingProfile` itself (e.g.,
`fallbackProfile: <name>`), and the framework should handle the chaining automatically
before the `ProfileHandler` is involved. This keeps `ProfileHandler` focused on
multi-profile orchestration (e.g., A/B testing, P/D routing) while fallback is a
framework guarantee.

## Design consideration: prefill/decode disaggregation

In prefill/decode (P/D) disaggregated serving, prefill and decode nodes need different
scoring strategies. Combined with fallback, the config becomes combinatorial:

- Primary prefill profile + fallback prefill profile
- Primary decode profile + fallback decode profile

This needs careful design to avoid config explosion. Options include:
- Profiles declare which pod types they apply to (e.g., `podType: prefill`)
- P/D routing is handled by filters within each profile, keeping profiles pod-type-agnostic
- A hierarchical config where P/D selection happens at a layer above profile selection

The fallback mechanism should be designed with P/D in mind from the start, even if P/D
support is implemented later.

## Open questions

- Should there be a limit on fallback chain depth?
- How should fallback interact with P/D profile selection to avoid combinatorial config?

gkneighb:
**What type of PR is this?**
/kind feature

Implements #1139. Adds a **framework-level fallback mechanism** to the scheduler: when a primary scheduling profile's scorers produce no usable signal (every weighted score is zero), the scheduler chains to a configured fallback profile before `ProfileHandler` is involved.

This lets operators express the kind of strategy currently baked into the `latency-scorer` plugin (`compositeKVWeight`, `compositeQueueWeight`, `compositePrefixWeight`) as a separate, explicit fallback profile in YAML.

### Architecture

| Component | What it does |
|---|---|
| `apix/config/v1alpha1/SchedulingProfile.FallbackProfile string` (new) | Operator-facing field; names another profile to fall through to |
| `fwksched.ProfileRunResult.NoSignal bool` (new) | Set by `SchedulerProfile.Run` when all weighted scores are zero |
| `Scheduler` loop (modified) | After each profile's `Run`, checks `NoSignal`; if a fallback is registered for that primary, runs the fallback and substitutes its result under the primary's name |
| `validateFallbackReferences()` in loader (new) | Rejects dangling references, self-references, and cycles at config load time |

**ProfileHandler stays unaware** that fallback fired. The `profileRunResults` map passed to `ProcessResults` contains the actual result that should be used (fallback's, when fallback fired) stored under the primary profile's name — matching the issue's framework-level guarantee.

### Example config

```yaml
schedulingProfiles:
- name: latency-prediction
  fallbackProfile: composite-fallback
  plugins:
  - pluginRef: predicted-latency-producer
  - pluginRef: latency-scorer
    weight: 1
  - pluginRef: weighted-random-picker

- name: composite-fallback
  plugins:
    weight: 1
  - pluginRef: queue-depth-scorer
    weight: 1
    weight: 1
  - pluginRef: weighted-random-picker
```

When `latency-prediction`'s scorers produce all-zero scores (predictions unavailable), `composite-fallback` runs in its place.

### Tests

12 new test cases:

- **Profile layer (2):** `NoSignal` flag is set when all weighted scores are zero; not set when at least one is non-zero.
- **Scheduler layer (3):** fallback fires when `NoSignal=true`; doesn't fire when primary has signal; primary result passes through when no fallback configured.
- **Loader validation (7 sub-cases):** valid one- and two-level chains, dangling reference, self-reference, two- and three-node cycles.

Updates the existing `TestPDSchedule` expected results to set `NoSignal=true` on the prefill profile (which has no scorers in that fixture and correctly reports no signal under the new behavior).

Verified locally: `make presubmit` green end-to-end (lint, format, typos, govulncheck). Full unit suite: 89 packages pass; the 6 remaining failures are pre-existing infra-dependent tests (`kind`/`etcd` required), unchanged from `main`.

### Out of scope (per the issue's own open questions)

- Removing `latency-scorer`'s composite-weight params (`compositeKVWeight`, etc.) — separate cleanup PR after this lands.
- P/D-disaggregated fallback selection — needs its own design pass; current implementation works correctly for non-disaggregated profiles and doesn't preclude future P/D work.
- Multi-level chain testing: recursion is bounded by cycle detection; tests cover one- and two-level chains.

Fixes #1139

**Release note**:
```release-note
Scheduling profiles now support a `fallbackProfile` field. When a primary profile's scorers produce no usable signal (every weighted score is zero), the scheduler runs the named fallback profile in its place. Dangling references, self-references, and cycles in the fallback chain are rejected at config load time.
```

### DIFF
--- a/apix/config/v1alpha1/endpointpickerconfig_types.go
+++ b/apix/config/v1alpha1/endpointpickerconfig_types.go
@@ -139,6 +139,13 @@ type SchedulingProfile struct {
 	// Plugins is the list of plugins for this SchedulingProfile. They are assigned
 	// to the appropriate "slots" based on their type.
 	Plugins []SchedulingPlugin `json:"plugins"`
+
+	// +optional
+	// FallbackProfile names another SchedulingProfile to run when this profile's
+	// scorers produce no usable signal (all weighted scores are zero). The
+	// referenced profile must exist in the same configuration and fallback chains
+	// must not cycle; both are validated at load time. See #1139.
+	FallbackProfile string `json:"fallbackProfile,omitempty"`
 }
 
 func (sp SchedulingProfile) String() string {
--- a/pkg/epp/config/loader/configloader.go
+++ b/pkg/epp/config/loader/configloader.go
@@ -214,12 +214,59 @@ func instantiatePlugins(configuredPlugins []configapi.PluginSpec, handle fwkplug
 	return nil
 }
 
+// validateFallbackReferences validates that all FallbackProfile references on
+// the given configProfiles point to existing profile names and that no fallback
+// chain forms a cycle (including the degenerate one-node self-reference case).
+// Returns nil when the reference graph is acyclic and complete. See #1139.
+func validateFallbackReferences(configProfiles []configapi.SchedulingProfile) error {
+	nameSet := make(map[string]bool, len(configProfiles))
+	fallbackByName := make(map[string]string)
+	for _, p := range configProfiles {
+		nameSet[p.Name] = true
+		if p.FallbackProfile != "" {
+			fallbackByName[p.Name] = p.FallbackProfile
+		}
+	}
+
+	for primary, fb := range fallbackByName {
+		if primary == fb {
+			return fmt.Errorf("profile %q references itself as fallback", primary)
+		}
+		if !nameSet[fb] {
+			return fmt.Errorf("profile %q references unknown fallback profile %q", primary, fb)
+		}
+	}
+
+	// Cycle detection: from each profile, walk the fallback chain and fail if a
+	// node is revisited. Each node has at most one outgoing edge, so this is
+	// O(N) per starting profile and O(N^2) overall — fine for a startup-time
+	// validation over a handful of profiles.
+	for start := range fallbackByName {
+		visited := map[string]bool{start: true}
+		current := fallbackByName[start]
+		for current != "" {
+			if visited[current] {
+				return fmt.Errorf("fallback cycle detected involving profile %q", start)
+			}
+			visited[current] = true
+			current = fallbackByName[current]
+		}
+	}
+	return nil
+}
+
 func buildSchedulerConfig(
 	configProfiles []configapi.SchedulingProfile,
 	handle fwkplugin.Handle,
 ) (*scheduling.SchedulerConfig, error) {
 
-	profiles := make(map[string]fwksched.SchedulerProfile)
+	// Validate fallback references early so we surface configuration errors before
+	// instantiating plugins. See #1139.
+	if err := validateFallbackReferences(configProfiles); err != nil {
+		return nil, err
+	}
+
+	profiles := make(map[string]fwksched.SchedulerProfile, len(configProfiles))
 
 	for _, cfgProfile := range configProfiles {
 		fwProfile := scheduling.NewSchedulerProfile()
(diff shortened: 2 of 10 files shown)


ahg-g:
closing as per the discussion in https://github.com/llm-d/llm-d-router/issues/1139

PR #1170, opened 2026-05-16
chore: remove call to fetch GIE CRD from remote, use local ones

zdtsw:
**What type of PR is this?**

/kind cleanup

Fixes #

```release-note
NONE
```


ahg-g (commented):
Can we pls update the ones in verify-helm as well?

ahg-g:
/approve

> Can we pls update the ones in verify-helm as well?
I see this is done in the other PR

PR #1171, opened 2026-05-17
perf(epp/scheduling): cut Scheduler.Schedule allocations ~90% on large fleets

gkneighb:
**What type of PR is this?**
/kind feature

`Scheduler.Schedule` runs on the per-request hot path. A new `BenchmarkSchedule` exercises it end-to-end with a production-style plugin stack (KV-cache + queue-depth + prefix-cache + LoRA-affinity scorers behind a max-score picker). Baseline profiling surfaced two surprising allocation hotspots in the scheduler framework code — both fixed in this PR, both behavior-preserving.

### Numbers

Apple M4, Go 1.26.3, `-count=5`:

| Endpoints | ns/op before → after | Δ ns | allocs/op before → after | Δ allocs | B/op before → after |
|---|---|---|---|---|---|
| 5   | 5,580 → 4,000   | **−28%** | 159   → 73   | **−54%** | 8.0K → 4.4K |
| 25  | 16,435 → 8,900  | **−46%** | 507   → 103  | **−80%** | 26.0K → 9.3K |
| 100 | 57,284 → 28,300 | **−51%** | 1,773 → 178  | **−90%** | 93.3K → 27.0K |

Allocation counts are identical across runs at each fleet size — `benchstat` can detect 5%+ regressions reliably going forward.

### Fix 1 (~80% of the win): gate the per-endpoint debug log by `Enabled()`

In `runScorerPlugins`, this line ran inside a scorers × endpoints inner loop:

```go
logger.V(logutil.DEBUG).Info("Calculated score", "plugin", scorer.TypedName(),
    "endpoint", endpoint.GetMetadata().NamespacedName, "score", score)
```

In production (DEBUG off) the message is discarded, but Go's eager argument evaluation still boxes `scorer.TypedName()`, the `NamespacedName`, and `score` into `interface{}` and constructs the variadic slice. That's ~4 allocations per call × 4 scorers × 100 endpoints = **~1,600 allocations per request, all for output that's thrown away**. Gating by `Enabled()` skips the call entirely when DEBUG is off; behavior under `-v` is unchanged.

### Fix 2 (~10% of the win): contiguous backing storage for picker input

`runPickerPlugin` was allocating each `ScoredEndpoint` individually (`&ScoredEndpoint{...}` in a loop). Replaced with a single contiguous backing array plus a parallel pointer slice. Picker signature unchanged (still receives `[]*ScoredEndpoint`); pickers can still reorder pointers (sort/shuffle). Goes from N+1 allocations to 2, and the backing values are now cache-friendly.

### Why this PR also lands a benchmark

`BenchmarkSchedule` is the regression guard. Without it, future scheduler changes could silently reintroduce per-request allocations. The plugin stack is realistic (mirrors `TestSchedule`), the fleet sizes (5/25/100) cover constant-overhead, linear-scaling, and large-fleet regimes, and request bodies are non-trivial so the prefix scorer does real work.

### Reproduction

```bash
# baseline (on main)
git checkout main
go test -run='^$' -bench=BenchmarkSchedule -benchmem -count=10 \
  ./pkg/epp/scheduling/ | tee before.out

# this branch
git checkout scheduler-benchmark
go test -run='^$' -bench=BenchmarkSchedule -benchmem -count=10 \
  ./pkg/epp/scheduling/ | tee after.out

benchstat before.out after.out
```

### Verification

- Existing tests: `go test ./pkg/epp/scheduling/...` passes.
- `make presubmit` green (lint, format, typos, govulncheck).
- **No interface changes** — plugins, ProfileHandler implementations, and external consumers are unaffected.

**Release note**:
```release-note
EPP scheduler hot path now allocates ~90% less per request on large fleets (100+ endpoints) and runs ~50% faster. Behavior is unchanged; the optimization eliminates per-endpoint allocations that occurred even when DEBUG logging was disabled.
```


pkg/epp/scheduling/scheduler_profile.go
```diff
@@ -156,17 +156,30 @@ func (p *SchedulerProfile) runScorerPlugins(ctx context.Context, request *fwksch
...
+
 	// Iterate through each scorer in the chain and accumulate the weighted scores.
 	for _, scorer := range p.scorers {
 		logger.V(logutil.VERBOSE).Info("Running scorer plugin", "plugin", scorer.TypedName())
 		before := time.Now()
 		scores := scorer.Score(ctx, cycleState, request, endpoints)
 		metrics.RecordPluginProcessingLatency(scorerExtensionPoint, scorer.TypedName().Type, scorer.TypedName().Name, time.Since(before))
 		for endpoint, score := range scores { // weight is relative to the sum of weights
-			logger.V(logutil.DEBUG).Info("Calculated score", "plugin", scorer.TypedName(), "endpoint", endpoint.GetMetadata().NamespacedName, "score", score)
+			if debugEnabled {
```
ahg-g, line 175: no need to have an if statement here

pkg/epp/scheduling/scheduler_profile.go
```diff
@@ -156,17 +156,30 @@ func (p *SchedulerProfile) runScorerPlugins(ctx context.Context, request *fwksch
...
+				debug.Info("Calculated score", "plugin", scorer.TypedName(), "endpoint", endpoint.GetMetadata().NamespacedName, "score", score)
+			}
 			weightedScorePerEndpoint[endpoint] += enforceScoreRange(score) * scorer.Weight()
 		}
-		logger.V(logutil.DEBUG).Info("Completed running scorer plugin successfully", "plugin", scorer.TypedName())
+		if debugEnabled {
```
ahg-g: ditto

ahg-g (commented):
Thanks for doing this, a couple of nits

pkg/epp/scheduling/scheduler_profile.go
gkneighb: Agreed — dropped in 3355f5b3. That log fires 4× per request total; the gate saves 8 allocs/req in the bench (within noise).

pkg/epp/scheduling/scheduler_profile.go
gkneighb: Measured both ways with `BenchmarkSchedule/endpoints=100` (`-count=3`):

| Variant | allocs/op | ns/op |
| --- | --- | --- |
| With this gate (current PR) | 178 | 28,200 |
| Without this gate | 1,666 | 57,200 |

Removing the gate restores ~1,500 allocs/req and 2× the latency — the entire optimization in the PR table comes from this one line. It's the inner per-endpoint log on a scorers × endpoints loop, so the variadic args (`TypedName` + `NamespacedName` + `score` boxed into `interface{}`) compound across 400 calls/req at 100 endpoints.

Happy to remove the gate if you have a different mechanism in mind — dropping the per-endpoint log entirely, or restructuring the call. The gate was the smallest behavior-preserving fix I could find.

pkg/epp/scheduling/scheduler_profile.go
```diff
@@ -175,10 +186,20 @@ func (p *SchedulerProfile) runScorerPlugins(ctx context.Context, request *fwksch
 
 func (p *SchedulerProfile) runPickerPlugin(ctx context.Context, cycleState *fwksched.CycleState, weightedScorePerEndpoint map[fwksched.Endpoint]float64) *fwksched.ProfileRunResult {
 	logger := log.FromContext(ctx)
-	scoredEndpoints := make([]*fwksched.ScoredEndpoint, len(weightedScorePerEndpoint))
+
+	// Allocate the ScoredEndpoint values as a single contiguous backing array
```
nirrozenbaum, line 190: do you want to move the performance analysis documentation to some README.md and keep godoc here minimal? 

usually godoc describes current code and not previous one.

ahg-g: it is probably fine as a learning experience.

pkg/epp/scheduling/scheduler_profile.go
ahg-g: worth a comment then since this can easily get removed in a follow up refactor

ahg-g: oh, I see you have it above, lgtm then.

gkneighb:
Rebased onto latest `main` to pick up #1164 (module rename from `llm-d-inference-scheduler` → `llm-d-router`). My imports needed updating to match. Two commits replayed cleanly; behavior unchanged. The earlier `lint-and-test` failure on this PR was the rename mismatch — that's now resolved.

Note: the force-push reset @ahg-g's `/approve` (standard GitHub behavior) and the CI workflows are sitting in `action_required` state waiting for maintainer approval to run. Sorry for the churn.

Also: `pkg/sidecar/proxy/decode.go:29` on `main` still imports the old `github.com/llm-d/llm-d-inference-scheduler/pkg/telemetry` — looks like #1164 missed it. Not in this PR's scope; flagging for visibility.

PR #1173, opened 2026-05-17
feat: expose metrics registration through plugin handle

### LINKED ISSUE #1093: [Observability]: Metrics registration via framework Handle
xref: https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2785

The framework should expose via the Handle an interface. This will also allow breaking the dependency of the upstream plugins on the pkg/epp/metrics pkg.

Why is this needed:

To enable plugins to register plugin-specific metrics.

wenhug:
## Summary
- Add a lean framework `MetricsRecorder` exposed through `plugin.Handle`; it embeds `prometheus.Registerer` only.
- Wire the runner to pass the controller-runtime Prometheus registry into the handle.
- Move approximate prefix and predicted latency metric definitions/recording helpers into their plugin packages.
- Register plugin-owned collectors via `handle.Metrics().Register(...)`, tolerate repeat registration of the same collector, and fail fast when a metrics registerer is missing.
- Keep legacy subsystem names referenced from `pkg/epp/metrics` and define the future `llm_d_router_endpoint_picker` subsystem constant there for follow-up compatibility work.
- Remove the previous concrete `pkg/epp/metrics.Recorder` adapter, test-only reset helpers from production files, and orphaned prefix-indexer golden files.

## Validation
- `go build ./...`
- `go vet ./pkg/epp/framework/... ./pkg/epp/metrics/... ./cmd/epp/...`
- `go test ./pkg/epp/framework/... ./pkg/epp/metrics/... ./test/utils/...`

Fixes #1093


pkg/epp/framework/interface/plugin/metrics_recorder.go
```diff
@@ -0,0 +1,107 @@
...
+	RecordRequestTTFT(ctx context.Context, modelName, targetModelName string, ttft float64) bool
+	RecordRequestPredictedTTFT(ctx context.Context, modelName, targetModelName string, predictedTTFT float64) bool
+	RecordRequestTTFTWithSLO(ctx context.Context, modelName, targetModelName string, ttft float64, sloThreshold float64) bool
+	RecordRequestTTFTPredictionDuration(ctx context.Context, modelName, targetModelName string, duration float64) bool
+	RecordRequestTPOT(ctx context.Context, modelName, targetModelName string, tpot float64) bool
+	RecordRequestPredictedTPOT(ctx context.Context, modelName, targetModelName string, predictedTPOT float64) bool
+	RecordRequestTPOTWithSLO(ctx context.Context, modelName, targetModelName string, tpot float64, sloThreshold float64) bool
+	RecordRequestTPOTPredictionDuration(ctx context.Context, modelName, targetModelName string, duration float64) bool
+	RecordPrefixCacheSize(size int64)
+	RecordPrefixCacheMatch(matchedLength, totalLength int)
```
hexfusion: > To enable plugins to register plugin-specific metrics.

should we keep this interface lean and unopinionated and let each plugin own its own metric definitions? high level sketch

```go
// pkg/.../something/metrics.go
var prefixCacheSize = prometheus.NewGauge(prometheus.GaugeOpts{
          Name: "epp_approximate_prefix_cache_size",
          Help: "Current size of the approximate prefix cache",
})
 
 func NewPlugin(handle plugin.Handle) (*Plugin, error) {
      if err := handle.Metrics().Register(prefixCacheSize); err != nil {
         return nil, fmt.Errorf("register prefix cache size metric: %w", err)
     }
     ...
}

// collect
prefixCacheSize.Set(float64(p.indexer.Size())
```

ahg-g: +1, the main point of https://github.com/llm-d/llm-d-router/issues/1093 is to enable each plugin to register its metrics, it is not about making the metrics part of the framework interface.

The metrics definitions could be defined in a central `https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/plugins/metrics/metrics.go` pkg just so the metrics are easy to discover.

```
var prefixCacheSize = prometheus.NewGauge(prometheus.GaugeOpts{
          Name: "epp_approximate_prefix_cache_size",
          Help: "Current size of the approximate prefix cache",
})
```

wenhug: Updated in be0d940e: the framework interface is now lean and only embeds prometheus.Registerer. The approximate prefix and predicted latency plugins now own their metric definitions and register their collectors through handle.Metrics().Register(...).

wenhug: Updated in be0d940e: plugin-specific metric definitions moved into their plugin packages, and MetricsRecorder now only exposes registration. I kept duplicate registration safe by accepting repeat registration of the same collector.

pkg/epp/framework/interface/plugin/metrics_recorder.go
```diff
@@ -0,0 +1,56 @@
...
+
+import "github.com/prometheus/client_golang/prometheus"
+
+// MetricsRecorder provides plugins access to metric registration without
+// depending on the concrete EPP metrics package.
+type MetricsRecorder interface {
+	prometheus.Registerer
+}
+
+// MetricsRecorderFromHandle returns the MetricsRecorder configured on h, or a
```
ahg-g: Is this really necessary? in what case would we want to start with a noop metrics recorder? my concern is that we silently use a noop recorder, no metrics get emitted, and then it is hard to debug why.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/metrics.go
```diff
@@ -0,0 +1,99 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+	compbasemetrics "k8s.io/component-base/metrics"
+
+	metricsutil "github.com/llm-d/llm-d-router/pkg/common/observability/metrics"
+)
+
+const inferenceExtension = "inference_extension"
```
ahg-g: In https://github.com/llm-d/llm-d-router/pull/1071 we are converging on one subsystem named `llm_d_router_endpoint_picker`, but we need to keep the older metric subsystem for backward compatibility for two releases; @rlakhtakia @wenhug how do we want to handle that? should this PR handle the deprecation of the plugins metrics as well and https://github.com/llm-d/llm-d-router/pull/1071 handles only the system metrics?

Another point, lets not define the subsystem variable here and reference the one in `https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/metrics/metrics.go#L39` (and define the `llm_d_router_endpoint_picker` there as well .

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/metrics.go
```diff
@@ -0,0 +1,99 @@
...
+func recordPrefixCacheMatch(matchedLength, totalLength int) {
+	prefixCacheHitLength.WithLabelValues().Observe(float64(matchedLength))
+
+	if totalLength > 0 {
+		ratio := float64(matchedLength) / float64(totalLength)
+		prefixCacheHitRatio.WithLabelValues().Observe(ratio)
+	}
+}
+
+func resetMetrics() {
```
ahg-g: if this is only used tests, then pls define it in the test file.

pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/metrics.go
```diff
@@ -0,0 +1,282 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	compbasemetrics "k8s.io/component-base/metrics"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+	metricsutil "github.com/llm-d/llm-d-router/pkg/common/observability/metrics"
+)
+
+const (
+	inferenceObjectiveComponent = "inference_objective"
```
ahg-g: ditto on subsystem name, we are converging on llm_d_router_endpoint_picker

pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/metrics.go
```diff
@@ -0,0 +1,282 @@
...
+		log.FromContext(ctx).V(logutil.DEFAULT).Error(nil, "TTFT prediction duration must be non-negative",
+			"modelName", modelName, "targetModelName", targetModelName, "duration", duration)
+		return false
+	}
+	requestTTFTPredictionDuration.WithLabelValues(modelName, targetModelName).Observe(duration)
+	inferenceGauges.WithLabelValues(modelName, targetModelName, typeTTFTPredictionDuration).Set(duration)
+	return true
+}
+
+func resetMetrics() {
```
ahg-g: ditto, pls define in the test file if this is not used in reality

pkg/epp/metrics/testdata/prefix_indexer_hit_bytes_metric
```diff
@@ -1,19 +0,0 @@
-# HELP inference_extension_prefix_indexer_hit_bytes [ALPHA] Length of the prefix match in number of bytes in the cache lookup.
```
ahg-g, line 1: why is this and the other test files got deleted?

pkg/epp/framework/interface/plugin/metrics_recorder.go
wenhug: Updated in d18f7365: removed NoopMetricsRecorder and MetricsRecorderFromHandle. Handles no longer silently default to a noop recorder; plugins that register metrics now fail fast if handle.Metrics() is nil. The runner wires the real controller-runtime registry, and tests use explicit registries.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/metrics.go
wenhug: Updated in d18f7365: the plugin packages no longer define their own subsystem string. They reference the shared subsystem constants from pkg/epp/metrics instead. I kept the existing legacy subsystem values for these collectors to preserve compatibility; I also added the llm_d_router_endpoint_picker constant in pkg/epp/metrics for the follow-up deprecation/dual-emission work rather than changing metric names in this PR.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/metrics.go
wenhug: Updated in d18f7365: moved resetMetrics into approximateprefix/metrics_test.go so it is no longer production code.

pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/metrics.go
wenhug: Updated in d18f7365: predictedlatency now references the shared subsystem constant from pkg/epp/metrics. I kept the current legacy subsystem value here for backward compatibility and left the new llm_d_router_endpoint_picker migration as follow-up work.

pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/metrics.go
wenhug: Updated in d18f7365: moved resetMetrics into predictedlatency/metrics_test.go so it is only compiled with tests.

pkg/epp/metrics/testdata/prefix_indexer_hit_bytes_metric
wenhug: These golden files were only used by the removed pkg/epp/metrics TestPrefixCacheMetrics. The prefix metrics now live in the approximateprefix plugin package and are covered by plugin-local tests that assert the collector values directly, so leaving these golden files would make them stale and unreferenced. The metric names, subsystem, labels, and buckets are unchanged.

pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/plugin.go
```diff
@@ -152,6 +152,13 @@ func PredictedLatencyFactory(name string, rawParameters json.RawMessage, handle
 		return nil, fmt.Errorf("invalid PredictedLatency config: %w", err)
 	}
 
+	if handle == nil {
+		return nil, fmt.Errorf("plugin handle is required")
```
ahg-g:
```suggestion
		return nil, errors.New("plugin handle is required")
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
```diff
@@ -78,6 +77,12 @@ func newDataProducer(ctx context.Context, name string, config config, handle plu
 	if config.MaxPrefixTokensToMatch < 0 {
 		return nil, fmt.Errorf("invalid configuration: MaxPrefixTokensToMatch must be >= 0 (current value: %d)", config.MaxPrefixTokensToMatch)
 	}
+	if handle == nil {
+		return nil, fmt.Errorf("plugin handle is required")
```
ahg-g:
```suggestion
		return nil, errors.New("plugin handle is required")
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
ahg-g, line 23:
```suggestion
"encoding/json"
"errors"
"fmt"
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
```diff
@@ -19,7 +19,9 @@ package approximateprefix
 import (
 	"context"
 	"encoding/json"
-	"fmt"
+"encoding/json"
+"errors"
+"fmt"
```
ahg-g:
```suggestion
    "errors"
    "fmt"
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
```diff
@@ -19,7 +19,10 @@ package approximateprefix
 import (
 	"context"
 	"encoding/json"
-	"fmt"
+"encoding/json"
+"errors"
```
ahg-g:
```suggestion
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
```diff
@@ -19,7 +19,8 @@ package approximateprefix
 import (
 	"context"
 	"encoding/json"
-	"fmt"
+    "errors"
```
ahg-g: I tried to fix this, but I am giving up :), can you pls fix the formatting.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
wenhug: Done in 46d56336: switched to errors.New and added the import.

pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/plugin.go
wenhug: Done in 46d56336: switched to errors.New and added the import.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
wenhug: Done in 46d56336: switched to errors.New and added the import.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/metrics.go
```diff
@@ -0,0 +1,95 @@
...
+
+	metricsutil "github.com/llm-d/llm-d-router/pkg/common/observability/metrics"
+	eppmetrics "github.com/llm-d/llm-d-router/pkg/epp/metrics"
+)
+
+var (
+	prefixCacheSize = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: eppmetrics.InferenceExtensionSubsystem,
+			Name:      "prefix_indexer_size",
```
ahg-g, line 34: @liu-cong @vMaroon we need to align the metrics names across the approximate and precise. These metric names should clearly indicate that they are for the approximate one. Perhaps we should prefix each plugin specific metrics with the plugin's name.

pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/metrics.go
```diff
@@ -0,0 +1,273 @@
...
+		90, 120, 180, 270, 360,
+	}
+
+	predictionLatencyBuckets = []float64{
+		0.0001, 0.0005, 0.001, 0.002, 0.005, 0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1.0, 2.0, 5.0,
+	}
+)
+
+var (
+	inferenceGauges = prometheus.NewGaugeVec(
```
ahg-g, line 67: @kaushikmitr is this gauge metric really useful? aren't the histograms per type sufficient?

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/metrics.go
liu-cong: I suggest we add labels to the metrics, see https://github.com/llm-d/llm-d-router/issues/1243

PR #1179, opened 2026-05-18
Fix: Avoid build failures from stale llm-d-inference-scheduler imports

weizhoublue:
**What type of PR is this?**
 
/kind bug
 
The repo rename was applied broadly, but a few runtime Go references were missed:
  - one stale Go import path
  - several hardcoded OpenTelemetry service/tracer names
  
Fixes #

```release-note
NONE
```


elevran:
@weizhoublue please fix conflicts.
Regarding the OTel strings - is it safe to rename them? For example, dashboards depending on these specific names.
Is there a way to create backward compatibility in OTel or need a clear cut-off?

weizhoublue:
Closing — the substantive changes (OTel service/tracer name renames and stale import paths) were already resolved on main by another PR. Only a redundant import and whitespace remain.

weizhoublue:
After I fix the conflict , It appears that the issue I found at first  has been already fixed by other pr

so close this pr

Issue #1069, opened 2026-05-12
[Bug] [Regression] Data producer dependency resolution forces singleton


liu-cong:
**What happened**:

Due to the hard coded dependency key, in reality for a data producer plugin, singleton is enforced.

**What you expected to happen**:

We should allow multiple instances of the same plugin. One of such use case is the tired prefix cache where we want to configure an approximate prefix producer for HBM cache, and another one for CPU cache, with different parameters.

A possible fix for this is that the dependency resolution is based on data producer type + name. And apply defaults.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- llm-d-scheduler version (use `git describe --tags --dirty --always` if you built from source, or specify the tag if you used a tagged version or image):
- Cloud provider or hardware configuration:
- Install tools:
- Others:

liu-cong:
cc @capri-xiyue @rahulgurnani @ahg-g

rahulgurnani:
Yes. Thanks for raising. Planning to address this soon.

ahg-g:
@rahulgurnani this is a practically a bug that is preventing us from realising the benefits of tiered kv-cache, we need to have a patch for the 0.8.0 branch and also a fix for the main branch.

rahulgurnani:
Based on offline discussion with Cong, he wants to take this up.

rahulgurnani:
The change https://github.com/llm-d/llm-d-router/pull/1096 should address part of it. For supporting multiple plugins, we would need to link them with the corresponding scorer. I started working on it https://github.com/llm-d/llm-d-router/pull/930 but was on hold and didn't get to it yet.

Possible proposed example config:
```
      kind: EndpointPickerConfig
      plugins:
      - type: queue-scorer
      - type: kv-cache-utilization-scorer
      - type: approx-prefix-cache-dataproducer
         name: cpu-prefix-cache-producer
      - type: prefix-cache-scorer
        name: gpu-prefix-cache-scorer
        producerRef: gpu-prefix-cache-producer
      - type: prefix-cache-scorer
        name: cpu-prefix-cache-scorer
        producerRef: cpu-prefix-cache-producer
     -  type: approx-prefix-cache-dataproducer
        name: cpu-prefix-cache-producer
        parameters:
          autoTune: false
          lruCapacityPerServer: 41000
      requestHandling:
[...]

PR #1180, opened 2026-05-18
Add wellknown EPP config test

liu-cong:
**What type of PR is this?**

/kind test

Add EPP config that match well-lit path guides. These tests can detect regressions in EPP config API and help prevent issues such as #1069 

Fixes #

```release-note
NONE
```


cmd/epp/runner/runner.go
```diff
@@ -466,11 +473,22 @@ func (r *Runner) registerInTreePlugins() {
 	fwkplugin.Register(random.RandomPickerType, random.RandomPickerFactory)
 	fwkplugin.Register(weightedrandom.WeightedRandomPickerType, weightedrandom.WeightedRandomPickerFactory)
 	fwkplugin.Register(single.SingleProfileHandlerType, single.SingleProfileHandlerFactory)
+	fwkplugin.Register(disagg.DisaggHeadersHandlerType, disagg.HeadersHandlerFactory)
```
ahg-g: how come those were not registered before?!

ahg-g: I see them registered in https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/plugins/register.go, we probably better consolidate in one place, either do all registration here or there.

liu-cong: Merged to runner.go

test/config/prefix_cache_mode_test.go
liu-cong, line 1: This test case is already covered in the well-known test

pkg/epp/framework/plugins/register_test.go
liu-cong, line 1: There is not much value in testing that a plugin is registered. The ultimate test should be the well known test with user facing config.

ahg-g:
pls fix the lint issue, this is good to go

PR #1194, opened 2026-05-18
Fix InferenceObjective poolRef group

learner0810:
**What type of PR is this?**

/kind bug

Fix InferenceObjective poolRef group

Fixes #

```release-note
NONE
```

### DIFF
--- a/config/charts/inferencepool/templates/inferenceobjectives.yaml
+++ b/config/charts/inferencepool/templates/inferenceobjectives.yaml
@@ -10,6 +10,6 @@ metadata:
 spec:
   priority: {{ .priority }}
   poolRef:
-    group: {{ .Values.inferenceExtension.apiVersion }}
+    group: {{ index (split "/" $.Values.inferencePool.apiVersion) 0 }}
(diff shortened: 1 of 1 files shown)


config/charts/inferencepool/templates/inferenceobjectives.yaml
```diff
@@ -10,6 +10,6 @@ metadata:
 spec:
   priority: {{ .priority }}
   poolRef:
-    group: {{ .Values.inferenceExtension.apiVersion }}
+    group: {{ index (split "/" $.Values.inferencePool.apiVersion) 0 }}
```
ahg-g, line 13: actually inferenceObjective is now an `llm-d.ai` api

shmuelk: Isn't this particular group for the InferencePool referenced by the InferenceObjective?

ahg-g: ah, right

PR #1196, opened 2026-05-18
chore: more cleanup in docs + tracing for renaming to llm-d-router

zdtsw:
**What type of PR is this?**

/kind cleanup

- update tracing service name to match charts value
- but keep metrics with old value `llm_d_inference_scheduler` to avoid break upgrade
-   Helm template naming drop `inferenceExtension`

follow up comments in https://github.com/llm-d/llm-d-router/pull/1166

```
Default OTEL tracing service name changed from `llm-d-inference-scheduler` to `llm-d-router`. Deployments that set env `OTEL_SERVICE_NAME` explicitly are unaffected --override default values in code. If current filter traces by the old default name  `llm-d-inference-scheduler` , require updates queries.
```


elevran:
thanks @zdtsw 
/lgtm

A few requests:
- would it be possible to also fix `docs/create_new_filter.md`?
- can you please confirm that there are no user facing changes that should be listed under the `release-notes`?
- if we're keeping some of the old names for backward compatibility, should open an issue for v0.11 to replace with the new values
- the coverage files include the old llm-d-inference-scheduler name. Not sure if we keep them around.

zdtsw:
> thanks @zdtsw /lgtm
> 
> A few requests:
> 
> * would it be possible to also fix `docs/create_new_filter.md`?

[wen]: anything specific need update in  `docs/create_new_filter.md`?

> * can you please confirm that there are no user facing changes that should be listed under the `release-notes`?

[wen]: updated PR description

> * if we're keeping some of the old names for backward compatibility, should open an issue for v0.11 to replace with the new values

[wen]: sure, will open an issue to follow the metrics part

> * the coverage files include the old llm-d-inference-scheduler name. Not sure if we keep them around.

[wen]: i guess need to run make test-unit to flush local coverage/*.out  once all renaming are done ?

zdtsw:
follow up issue for "metrics" system name https://github.com/llm-d/llm-d-router/issues/1199

PR #1193, opened 2026-05-18
feat: add support for mooncake as new kv-connector in routing sidecar

zdtsw:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

- Add Mooncake KV transfer protocol as a new connector for the routing sidecar
- Mooncake uses concurrent prefill/decode: prefill fires async in a goroutine while decode runs synchronously, both sharing a transfer ID for KV cache coordination
-  Queries prefill pod to get the engine ID needed for KV transfer as `remote_engine_id`
- Add` --mooncake-bootstrap-port `flag or `MOONCAKE_BOOTSTRAP_PORT` env var (default: 8998, matching vLLM's default)
- Update `docs/disaggregation.md` with all support kv-connector usage
- depedency: https://github.com/llm-d/llm-d-inference-sim/pull/538 for e2e

**Which issue(s) this PR fixes**:

Fixes #1010

**Release note** _(write `NONE` if no user-facing change)_:
```release-note
NONE
```

**Example**
original request payload
```
{
  "model": "llama-3",
  "messages": [{"role": "user", "content": "PTO"}],
  "stream": true,
  "stream_options": {"include_usage": true},
  "max_tokens": 512
}
```
sidecar to prefill payload
```
{
  "model": "llama-3",
  "messages": [{"role": "user", "content": "PTO"}],
  "stream": false,
  "max_tokens": 1,
  "kv_transfer_params": {
    "do_remote_prefill": false,
    "do_remote_decode": true,
    "transfer_id": "xfer-XXXXXXXX"
  }
}
```
sidecar to decode payload
```
{
  "model": "llama-3",
  "messages": [{"role": "user", "content": "PTO"}],
  "stream": true,
  "stream_options": {"include_usage": true},
  "max_tokens": 512,
  "kv_transfer_params": {
    "do_remote_prefill": true,
    "do_remote_decode": false,
    "transfer_id": "xfer-XXXXXXXX",
    "remote_bootstrap_addr": <prefill_pod_ip>:8998,
    "remote_engine_id": "engine-abc123"
  }
}

```

PR #1200, opened 2026-05-18
refactor:  a bunch of clean up and refactor in routing sidecar

zdtsw:
**What type of PR is this?**
/kind cleanup

a bunch of chore/refactor not fix/feature change
- remove nixl case in test 
- rename function `createModelServersPDNixl` to `createModelServersPDNixlV2`
- refactor: cloneRequestWithBody replace old cloneWithBody and move to proxy_helpers.go
- refactor: move read json body unmarshal into a method to  reduce duplicated code in different kv-connectors
- refactor:  move sidecar const into a new package and refer hardcoded value by const
- refactor: move `normalizeHostPort()` in to helper , address [comments](https://github.com/llm-d/llm-d-router/pull/1193#discussion_r3259113017)

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/sidecar/constants/constants.go
+++ b/pkg/sidecar/constants/constants.go
@@ -1,0 +1,37 @@
+/*
+Copyright 2025 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package constants
+
+const (
+	// KVConnectorNIXLV2 enables the P/D KV NIXL v2 protocol
+	KVConnectorNIXLV2 = "nixlv2"
+
+	// KVConnectorSharedStorage enables the P/D KV Shared Storage protocol
+	KVConnectorSharedStorage = "shared-storage"
+
+	// KVConnectorSGLang enables SGLang the P/D KV disaggregation protocol
+	KVConnectorSGLang = "sglang"
+
+	// ECExampleConnector enables the Encoder disaggregation protocol (E/PD, E/P/D)
+	ECExampleConnector = "ec-example"
+
+	// DefaultPoolGroup is the default pool group name
+	DefaultPoolGroup = "inference.networking.k8s.io"
+
+	// LegacyPoolGroup is the legacy pool group name
+	LegacyPoolGroup = "inference.networking.x-k8s.io"
+)
--- a/pkg/sidecar/proxy/proxy.go
+++ b/pkg/sidecar/proxy/proxy.go
@@ -33,6 +33,8 @@ import (
 	lru "github.com/hashicorp/golang-lru/v2"
 	"golang.org/x/sync/errgroup"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/llm-d/llm-d-router/pkg/sidecar/constants"
 )
 
 const (
@@ -68,23 +70,12 @@ const (
 	requestFieldBootstrapPort = "bootstrap_port"
 	requestFieldBootstrapRoom = "bootstrap_room"
 
-	// KVConnectorNIXLV2 enables the P/D KV NIXL v2 protocol
-	KVConnectorNIXLV2 = "nixlv2"
-
-	// KVConnectorSharedStorage enables the P/D KV Shared Storage protocol
-	KVConnectorSharedStorage = "shared-storage"
-
-	// KVConnectorSGLang enables SGLang the P/D KV disaggregation protocol
-	KVConnectorSGLang = "sglang"
-
-	// ECExampleConnector enables the Encoder disaggregation protocol (E/PD, E/P/D)
-	ECExampleConnector = "ec-example"
-
-	// DefaultPoolGroup is the default pool group name
-	DefaultPoolGroup = "inference.networking.k8s.io"
-
-	// LegacyPoolGroup is the legacy pool group name
-	LegacyPoolGroup = "inference.networking.x-k8s.io"
+	KVConnectorNIXLV2        = constants.KVConnectorNIXLV2
+	KVConnectorSharedStorage = constants.KVConnectorSharedStorage
+	KVConnectorSGLang        = constants.KVConnectorSGLang
+	ECExampleConnector       = constants.ECExampleConnector
+	DefaultPoolGroup         = constants.DefaultPoolGroup
+	LegacyPoolGroup          = constants.LegacyPoolGroup
 )
 
 // APIType represents the type of OpenAI API being used.
--- a/pkg/sidecar/proxy/proxy_helpers.go
+++ b/pkg/sidecar/proxy/proxy_helpers.go
@@ -161,21 +161,27 @@ func (s *Server) createDecoderProxyHandler(decoderURL *url.URL, decoderInsecureS
 	return decoderProxy
 }
 
-// readJSONBody reads the request body, closes it, and unmarshals it into a
-// map. On failure it writes the appropriate HTTP error response and returns
-// ok=false so the caller can simply return.
-func (s *Server) readJSONBody(r *http.Request, w http.ResponseWriter) ([]byte, map[string]any, bool) {
+func bodyAsJSON(r *http.Request) ([]byte, map[string]any, error) {
 	defer func() { _ = r.Body.Close() }()
 	raw, err := io.ReadAll(r.Body)
 	if err != nil {
-		w.WriteHeader(http.StatusBadRequest)
-		_, _ = w.Write([]byte(err.Error()))
-		return nil, nil, false
+		return nil, nil, err
 	}
 	var parsed map[string]any
 	if err := json.Unmarshal(raw, &parsed); err != nil {
-		if err := errorJSONInvalid(err, w); err != nil {
-			s.logger.Error(err, "failed to send error response to client")
+		return nil, nil, fmt.Errorf("%w: %w", errInvalidJSON, err)
+	}
+	return raw, parsed, nil
+}
+
+func (s *Server) readJSONBody(r *http.Request, w http.ResponseWriter) ([]byte, map[string]any, bool) {
+	raw, parsed, err := bodyAsJSON(r)
+	if err != nil {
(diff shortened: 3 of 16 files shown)


pkg/sidecar/proxy/proxy_helpers.go
```diff
@@ -158,6 +161,34 @@ func (s *Server) createDecoderProxyHandler(decoderURL *url.URL, decoderInsecureS
 	return decoderProxy
 }
 
+// readJSONBody reads the request body, closes it, and unmarshals it into a
+// map. On failure it writes the appropriate HTTP error response and returns
+// ok=false so the caller can simply return.
+func (s *Server) readJSONBody(r *http.Request, w http.ResponseWriter) ([]byte, map[string]any, bool) {
+	defer r.Body.Close() //nolint:errcheck
```
roytman:
```suggestion
	defer func() { _ = r.Body.Close() }()
```
So we don't need the lint directives

pkg/sidecar/proxy/proxy_helpers.go
```diff
@@ -158,6 +161,34 @@ func (s *Server) createDecoderProxyHandler(decoderURL *url.URL, decoderInsecureS
...
+	raw, err := io.ReadAll(r.Body)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest) // TODO: check FastAPI error code when failing to read body
+		w.Write([]byte(err.Error()))         //nolint:errcheck
```
roytman:
```suggestion
				_, _ = w.Write([]byte(err.Error()))
```

test/sidecar/mock/chat_completions_handler.go
```diff
@@ -82,9 +82,6 @@ func (cc *ChatCompletionHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
 	var rawResponse string
 
 	switch cc.Connector {
-	case "nixl":
-		rawResponse = `{"remote_block_ids":[1, 2, 3], "remote_engine_id": "5b5fb28f-3f30-4bdd-9a36-958d52459200"}`
-
 	case "nixlv2":
```
roytman: We use this and similar strings in several places. Can they be replaced with the constants  from https://github.com/llm-d/llm-d-router/blob/8cffd480ad6b9b9ff19f4d6b2452317c49e5b118/pkg/sidecar/proxy/proxy.go#L72

The same for 
pkg/sidecar/proxy/connector_nixlv2.go:64:		attribute.String("llm_d.pd_proxy.connector", "nixlv2"),
pkg/sidecar/proxy/connector_nixlv2.go:200:		attribute.String("llm_d.pd_proxy.connector", "nixlv2"),
test/e2e/setup_test.go:84:	return createModelServersPDWithConnector(prefillReplicas, decodeReplicas, "nixlv2")

roytman (approved):
Thank you for your PR. It really improves the code. 
Several comments.

test/sidecar/mock/chat_completions_handler.go
zdtsw: good idea, i will get hardcoded values with const as well

zdtsw:
@elevran mind taking look at this one?

pkg/sidecar/constants/constants.go
elevran, line 1: what is the rationale for a package with just file? Can it be placed directly under `sidecar` or would that create an import loop?

pkg/sidecar/constants/constants.go
```diff
@@ -0,0 +1,37 @@
...
+	// KVConnectorSGLang enables SGLang the P/D KV disaggregation protocol
+	KVConnectorSGLang = "sglang"
+
+	// ECExampleConnector enables the Encoder disaggregation protocol (E/PD, E/P/D)
+	ECExampleConnector = "ec-example"
+
+	// DefaultPoolGroup is the default pool group name
+	DefaultPoolGroup = "inference.networking.k8s.io"
+
+	// LegacyPoolGroup is the legacy pool group name
```
elevran, line 35: Q: when (as in which GAIE release) was the pool group changed? When would it be safe to remove it completely?

pkg/sidecar/proxy/proxy.go
```diff
@@ -68,22 +70,12 @@ const (
...
-	KVConnectorSGLang = "sglang"
-
-	// ECExampleConnector enables the Encoder disaggregation protocol (E/PD, E/P/D)
-	ECExampleConnector = "ec-example"
-
-	// DefaultPoolGroup is the default pool group name
-	DefaultPoolGroup = "inference.networking.k8s.io"
-	// LegacyPoolGroup is the legacy pool group name
-	LegacyPoolGroup = "inference.networking.x-k8s.io"
+	KVConnectorNIXLV2        = constants.KVConnectorNIXLV2
```
elevran, line 73: Q: can'g the constants be used directly instead of via the alias (i.e., `constants.KVConnectorNIXLV2` instead of `KVConnectorNIXLV2`)?
Trying to understand the need for intermediaries.

pkg/sidecar/proxy/proxy_helpers.go
```diff
@@ -158,6 +161,45 @@ func (s *Server) createDecoderProxyHandler(decoderURL *url.URL, decoderInsecureS
 	return decoderProxy
 }
 
+// readJSONBody reads the request body, closes it, and unmarshals it into a
+// map. On failure it writes the appropriate HTTP error response and returns
+// ok=false so the caller can simply return.
+func (s *Server) readJSONBody(r *http.Request, w http.ResponseWriter) ([]byte, map[string]any, bool) {
```
elevran: nit: conflating reading and error return. If there's ever a need to only parse (and eg just log the error), this would need a refactor.
Consider separate `func BodyAsJSON(r http.Request) (map[string]any, error)` and `func HTTPError(w http.ResponseWriter)`. The two can be easily composed into this function.

pkg/sidecar/proxy/proxy_helpers.go
```diff
@@ -158,6 +161,45 @@ func (s *Server) createDecoderProxyHandler(decoderURL *url.URL, decoderInsecureS
...
+func cloneRequestWithBody(ctx context.Context, r *http.Request, body []byte) *http.Request {
+	cloned := r.Clone(ctx)
+	cloned.Body = io.NopCloser(bytes.NewReader(body))
+	cloned.ContentLength = int64(len(body))
+	return cloned
+}
+
+// normalizeHostPort returns the host part of a host:port string. If parsing
+// fails (e.g. no port), the input is returned as-is.
+func normalizeHostPort(hostPort string) string {
```
elevran: nit: find `normalization of host and port` confusing since it's extracting the host part only.

pkg/sidecar/constants/constants.go
zdtsw: originally, the plan is to keep all const as is (in proxy.go), and replace hardcode value (in pkg/sidecar) with them.
the part wont work with this way is the test in `chat_completions_handler.go` : either keep hardcode or creat a new package for import

i am fine with either way.

pkg/sidecar/constants/constants.go
zdtsw: it is just moved from proxy.go to constants.go

as for the API group and version change, 
we currently support both https://github.com/llm-d/llm-d-router/blob/main/pkg/sidecar/proxy/allowlist.go#L45

as in GIE https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2796 in [1.5.0 ]  (https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/tag/v1.5.0) release

pkg/sidecar/proxy/proxy.go
zdtsw: that can work.
but then everywhere ref `KVConnectorNIXLV2` will need to be replaced with  `constants.KVConnectorNIXLV2` 
to do this, is to minimize changes

pkg/sidecar/proxy/proxy_helpers.go
zdtsw: let me rename it to `extractHost` more stragith for the logic

pkg/sidecar/proxy/proxy_helpers.go
zdtsw: did some change for this  in  https://github.com/zdtsw/llm-d-router/commit/090016c78c018b90acecf93df9d2d505e6b3ffbd

test/e2e/e2e_test.go
```diff
@@ -83,7 +83,7 @@ var _ = ginkgo.Describe("Run end to end tests", ginkgo.Ordered, func() {
 
 			prefillReplicas := 1
 			decodeReplicas := 4
-			modelServers := createModelServersPDNixl(prefillReplicas, decodeReplicas)
+			modelServers := createModelServersPDNixlV2(prefillReplicas, decodeReplicas)
```
elevran, line 86: nit: is there still a NIXL v1 anywhere, or can this be renamed to just NIXL?

elevran (commented):
/lgtm
/approve
Please consider using the constants all over in a follow up PR. Also consider NIXLv2 designation comment. Not blocking merge.

pkg/sidecar/constants/constants.go
elevran: Please open an issue to remove the legacy group designation in router v0.11

test/e2e/e2e_test.go
zdtsw: i dont think we used v1 anywhere in code.  nixl was the orignal v1, then v2 was introduced with suffix.
i put v2 here to be more speicific, in case, we have a v3 someday. but we could take everything(v2) to be nixl

pkg/sidecar/constants/constants.go
zdtsw: follow up issue https://github.com/llm-d/llm-d-router/issues/1298

zdtsw:
> /lgtm /approve Please consider using the constants all over in a follow up PR. Also consider NIXLv2 designation comment. Not blocking merge.

hi,
i made the last part within `pkg/sidecar` to remove hardcoded value in https://github.com/llm-d/llm-d-router/pull/1200/changes/92342dcb0598d01c1320f998af60192d5cd1f5b5
we still have some in  `pkg/epp` but that will require move const entirely under `pkg/common` to be shared by both epp and sidecar. 
is this the correct understanding?

zdtsw:
@elevran mind having another check ?

elevran:
> hi, i made the last part within `pkg/sidecar` to remove hardcoded value in [92342dc](https://github.com/llm-d/llm-d-router/commit/92342dcb0598d01c1320f998af60192d5cd1f5b5) we still have some in `pkg/epp` but that will require move const entirely under `pkg/common` to be shared by both epp and sidecar. is this the correct understanding?

That is correct. Can be a follow up.

PR #1206, opened 2026-05-18
[sidecar] Add retry option for prefill queries

ilmarkov:
**What type of PR is this?**
/kind feature

When prefill pods are under heavy load, their Uvicorn accept queue can overflow, causing TCP connection resets that the sidecar's reverse proxy surfaces as 502 errors. Previously the sidecar treated any prefill failure as final and fell back to decode with the original request body (no kv_transfer_params), forcing the decode engine into an expensive local prefill computation.

This PR replaces the fallback-to-decode behavior with configurable retry logic on the NIXL v2 prefill path. When a prefill request fails with a transient 5xx status (502, 503, 504), the sidecar retries the same prefill host up to --prefill-max-retries times with a --prefill-retry-backoff delay between attempts. Non-transient errors (500, 501) fail immediately. If all retries are exhausted or retries are disabled (the default), the error is returned to the client.

**Breaking change**: the previous fallback-to-decode-on-prefill-failure path is removed. On prefill failure, the sidecar now returns the error to the client instead of silently dispatching an unaccelerated decode request.

```release-note
sidecar: replace fallback-to-decode on prefill failure with configurable retry logic (--prefill-max-retries, --prefill-retry-backoff). Prefill errors are now returned to the client instead of silently falling back to unaccelerated decode.
```


robertgshaw2-redhat:
I think that we should make the default behavior request failure rather than running on decode

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -143,8 +143,34 @@ func (s *Server) runNIXLProtocolV2(w http.ResponseWriter, r *http.Request, prefi
...
+	for attempt := 0; ; attempt++ {
+		pw = &bufferedResponseWriter{}
+		preq.Body = io.NopCloser(bytes.NewReader(pbody))
+		preq.ContentLength = int64(len(pbody))
+		prefillHandler.ServeHTTP(pw, preq)
+
+		if !isHTTPError(pw.statusCode) {
+			break
+		}
+		if !shouldFallbackToDecode(pw) {
```
hexfusion: just took a look at isHTTPError and shouldFallbackToDecode. There are some non-retryable scenarios where it would make sense to fail fast specifically 500 and 501.

roytman:
Should we have the same behaviour for other connectors: `sglang` and `epd`? The last case is even more interesting, because we can choose another Encoder.

elevran:
@ilmarkov can you please address conflict and DCO?

ilmarkov:
@elevran Done

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -130,8 +125,35 @@ func (s *Server) handleNIXLV2(w http.ResponseWriter, r *http.Request, prefillPod
...
+		if attempt >= s.config.PrefillMaxRetries {
+			break
+		}
+
+		s.logger.Info("retrying prefill request",
+			"attempt", attempt+1,
+			"target", prefillPodHostPort,
+			"request_id", uuidStr,
+			"previous_code", pw.statusCode)
+		time.Sleep(s.config.PrefillRetryBackoff)
```
elevran: `time.Sleep` does not check for client disconnect. If the caller cancels mid-backoff, this goroutine sleeps for the full `PrefillRetryBackoff` and then issues another call to an already-stressed prefill backend. Use a `select` with the request context:

```go
select {
case <-time.After(s.config.PrefillRetryBackoff):
case <-preq.Context().Done():
    goto afterLoop
}
```

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -140,26 +162,22 @@ func (s *Server) handleNIXLV2(w http.ResponseWriter, r *http.Request, prefillPod
 	)
 
 	if isHTTPError(pw.statusCode) {
-		s.logger.Error(err, "request failed", "code", pw.statusCode, "body", pw.buffer.String())
+		s.logger.Error(err, "prefill request failed",
```
elevran: `err` at this point is from an earlier operation that succeeded (or we would have returned). It is nil. `s.logger.Error(nil, ...)` logs at error level with no actual error value. Use `s.logger.Info` here, or pass `fmt.Errorf("prefill returned %d", pw.statusCode)` as the error.

pkg/sidecar/proxy/options.go
```diff
@@ -514,6 +523,14 @@ func (opts *Options) mergeYAMLConfiguration(cfg yamlConfiguration) {
 	if cfg.PoolGroup != "" && !opts.isFlagSet(poolGroup) {
 		opts.PoolGroup = cfg.PoolGroup
 	}
+	if cfg.PrefillMaxRetries != nil && !opts.isFlagSet(prefillMaxRetries) {
+		opts.PrefillMaxRetries = *cfg.PrefillMaxRetries
+	}
+	if cfg.PrefillRetryBackoff != "" && !opts.isFlagSet(prefillRetryBackoff) {
+		if d, err := time.ParseDuration(cfg.PrefillRetryBackoff); err == nil {
```
elevran: An invalid duration string (e.g. `prefill-retry-backoff: "200"` missing the unit) silently falls back to the 200ms default with no log or error. At minimum log the failure so operators know their setting was ignored.

pkg/sidecar/proxy/options.go
```diff
@@ -235,6 +242,8 @@ func (opts *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&opts.InferencePoolName, inferencePoolName, opts.InferencePoolName, "Deprecated: use --inference-pool instead. The specific InferencePool name (defaults to INFERENCE_POOL_NAME env var)")
 	_ = fs.MarkDeprecated(inferencePoolName, "use --inference-pool instead")
 	fs.IntVar(&opts.MaxIdleConnsPerHost, "max-idle-conns-per-host", opts.MaxIdleConnsPerHost, "max idle keep-alive connections per host for reverse proxy transports; set to at least the expected concurrency")
+	fs.IntVar(&opts.PrefillMaxRetries, prefillMaxRetries, opts.PrefillMaxRetries, "max retry attempts when a prefill request fails with a 5xx error; 0 means no retries (default)")
+	fs.DurationVar(&opts.PrefillRetryBackoff, prefillRetryBackoff, opts.PrefillRetryBackoff, "delay between prefill retry attempts")
```
elevran, line 258: nit: `Validate()` has no guard for `PrefillRetryBackoff <= 0`. With retries enabled, a zero or negative backoff produces a rapid burst of calls against an already-overloaded backend (bounded by `PrefillMaxRetries`, but undesirable). Consider adding a check consistent with the existing `DecodeChunkSize < 0` guard.

pkg/sidecar/proxy/proxy_helpers.go
```diff
@@ -210,10 +210,12 @@ func isHTTPError(statusCode int) bool {
...
-func shouldFallbackToDecode(pw *bufferedResponseWriter) bool {
-	if pw.statusCode >= http.StatusBadRequest && pw.statusCode <= http.StatusUnavailableForLegalReasons {
-		return false
-	}
-	return true
+// isRetryableStatus returns true for transient 5xx errors where retrying
+// the same host is likely to succeed (e.g. TCP connection reset, overloaded
+// accept queue). Non-transient errors like 500/501 indicate bugs or
+// unsupported operations and should fail fast.
+func isRetryableStatus(statusCode int) bool {
```
elevran, line 222: `isRetryableStatus` covers 502, 503, and 504, but the new tests only exercise 502. A short table-driven unit test here (or one additional scenario in `connector_nixlv2_test.go`) would confirm 503 and 504 are wired correctly.

Also, `options_test.go` `TestSidecarConfiguration` does not cover the two new YAML fields. Following the existing pattern, add `prefill-max-retries` and `prefill-retry-backoff` to `createConfigWithValidYAML` and the corresponding expected-value assertions.

elevran (commented):
Overall the retry mechanism is clean and the 502 paths are well-tested - thanks!

Some overall comment belong in the PR body rather than inline: 
1. the description says "the existing fallback-to-decode behavior is preserved unchanged" but the code removes it entirely. This needs to be corrected and the decision justified
1. `release-note: NONE` is wrong since the fallback removal is a user-visible contract change
1. `Fixes #` is blank. Any non-trivial work should be tracked in an issue before landing. Please follow the guideline next time.

elevran:
@ilmarkov also note the conflict

ahg-g:
@ilmarkov thanks for raising this PR, should we move forward here (resolve conflicts and address the open comments)?

elevran:
@ilmarkov 
The PR description says 'the existing fallback-to-decode behavior is preserved unchanged', but the code removes it entirely.
This removal is intentional (per @robertgshaw2-redhat request) replacing decode only processing on failure with retry-then-fail strategy.
The PR description should be fixed and a release notes section updated from `NONE` to mention this behavior change.

ilmarkov:
@elevran @ahg-g PR is updated. Please, have another look at it.

hexfusion:
LGTM

@ilmarkov can you resolve conflict please

elevran (approved):

PR #1207, opened 2026-05-19
feat: add native metrics mapping for standard triton. This allows the EPP to monitor standard NVIDIA Triton Inference Server workloads without requiring custom Data Layer configurations. 

### LINKED ISSUE #1102: Support non-LLM model servers and models
xref: https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2923

**What would you like to be added**: Support for non-LLM model servers and models such as Triton (non-LLM) and RecML for smart routing capabilities such as `queue-scorer` based metrics.

**Why is this needed**: Users are interested in smart routing capabilities for non-LLM model servers and models. The non-LLM specific capabilities such as smart routing based on `non-KV cache` metrics (e.g. `request queue depth`) should still be feasible.

**Current solutions**: For example of a Triton + RecML setup,
1. A custom `helm-values.yaml`
```
# 1. Force the EPP to scrape port 8080 instead of the inference port (8000)
model-server-metrics-port: 8080
```
```
# 2. Enable the hidden v1.4.0 dataLayer feature
      featureGates:
        - dataLayer
```
```
# 3. Explicitly declare the v1.4.0 metrics extractor
        - name: model-server-protocol-metrics
          type: model-server-protocol-metrics
          parameters:
            engineConfigs:
              - name: vllm
                # 4. OVERRIDE the vllm defaults with Triton's specific metric names
                queuedRequestsSpec: "nv_inference_pending_request_count"
                runningRequestsSpec: "nv_inference_exec_count"
                kvUsageSpec: ""
                loraSpec: ""
                cacheInfoSpec: ""
```
```
inferencePool:
  # 5. Leave this as vllm. Do NOT use "custom" or "triton-tensorrt-llm".
  modelServerType: vllm 
```
2. Payload Compatibility (The OpenAI Parser)
**The Problem:**
The EPP defaults to using an `openai-parser`. If a request arrives without a `model`, `prompt`, or `messages` field, the EPP rejects it with a `400 Bad Request`. Standard Triton API payloads use an `inputs` array.
**The Triton Configuration:**
We must wrap the Triton payload in dummy OpenAI fields. The EPP parses the dummy fields and approves the request, while Triton ignores them and reads the `inputs` array.
*Universal Payload Format:*
```json
{
  "model": "recml-model",
  "prompt": "dummy",
  "inputs": [
    {
      "name": "INPUT__0",
      "shape": [1, 4096],
      "datatype": "FP32",
      "data": [0.0]
    }
  ]
}
```
3. (Optional): Response Parsing (The GZIP Crash)
**The Problem:**
The EPP intercepts the response body from the model server to track completion metrics. Triton compresses responses using `gzip` by default. The EPP (v1.4.0) attempts to unmarshal this binary data (`\x1f`) as plain JSON and enters a crash loop, returning `503 Gateway Timeouts`.
**The Triton Configuration:**
You must configure Triton to ignore the client's `Accept-Encoding: gzip` header.
*In the Triton Deployment args (`manifests/01-triton-workloads-fast.yaml`):*
```yaml
args: 
  - "--allow-grpc=false"
  - "--http-header-forward-pattern=" # Prevents Triton from receiving the GZIP header
``` 

**Suggested solutions:**: Native support for non-LLM model servers and models
1. Enable `dataLayer` feature by default to support custom metrics.
2. Remove the requirements for `model`, `prompt`, or `messages` fields for all models or flexibly check per serving model.

PS: I am aware that there is an item official roadmap item for this request
```
6. Support for other GenAI model types (diffusion and other non-completion protocols)
7. Heterogeneous accelerators - serve workloads on multiple types of accelerator using latency and request cost-aware load balancing
```

andrewdoatgoogle:
1. Add `triton` to `defaultEngineConfigs` to natively map `nv_inference_pending_request_count` and `nv_inference_exec_count`. This allows the EPP to monitor standard NVIDIA Triton Inference Server workloads without requiring custom Data Layer configurations. 
2. Includes explicit unit test coverage and updated documentation.

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This allows the EPP to monitor standard NVIDIA Triton Inference Server workloads without requiring custom Data Layer configurations. 

Fixes #1102 

```release-note
**The Problem:**
Currently, the Endpoint Picker (EPP) auto-populates metric extraction mappings only for LLM-specific engines (`vllm`, `sglang`, `triton-tensorrt-llm`, etc.). Users running standard Machine Learning workloads on the NVIDIA Triton Inference Server must write complex custom `EndpointPickerConfig` overrides to map the queue metrics, or the `queue-scorer` breaks.

**The Proposed Change:**
This PR introduces native support for standard Triton in the Go codebase. 
 - Adds a `"triton"` configuration to `defaultEngineConfigs`.
 - Maps `QueuedRequestsSpec` to `"nv_inference_pending_request_count"`.
 - Maps `RunningRequestsSpec` to `"nv_inference_exec_count"`.
 - Explicitly leaves LLM-specific features (like KV Cache and LoRA) empty to prevent scraping errors.

**Validation:**
 - Added `TestDefaultEngineConfigsTritonValues` to explicitly assert the accuracy of the metric string mappings.
 - Confirmed `TestCoreMetricsExtractorFactoryDefaultEngine/defaultEngine_triton` passes successfully.
```


pkg/epp/framework/plugins/datalayer/extractor/metrics/factories.go
```diff
@@ -116,6 +116,16 @@ var defaultEngineConfigs = []engineConfigParams{
 		CacheBlockSizeSpec:  "nv_trt_llm_kv_cache_block_metrics{kv_cache_block_type=tokens_per}",
 		CacheNumBlocksSpec:  "nv_trt_llm_kv_cache_block_metrics{kv_cache_block_type=max}",
 	},
+	{
+		Name:                "triton",
```
ahg-g, line 128: can we pls add a comment explaining the difference between this and the two variations above? `triton-tensorrt-llm` and `trtllm-serve`?

andrewdoatgoogle:
manually validated in this project https://github.com/andrewaddo/gke-ig-ubb/blob/llmd-epp/MILESTONE_LLMD.md

pkg/epp/framework/plugins/datalayer/extractor/metrics/factories.go
andrewdoatgoogle: Added comment to explain the difference

PR #1208, opened 2026-05-19
feat(helm): auto-select passthrough-parser for triton model server. This prevents the EPP from crashing when attempting to route standard ML payloads (like KServe v2) using the default OpenAI parser.

### LINKED ISSUE #1102: Support non-LLM model servers and models

andrewdoatgoogle:
1. Update the Helm chart logic to automatically inject the `passthrough-parser` when `modelServerType` is set to `triton`. This prevents the EPP from crashing when attempting to route standard ML payloads (like KServe v2) using the default OpenAI parser. 
2. Also adds explicit integration tests to verify the parser injection.

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: This prevents the EPP from crashing when attempting to route standard ML payloads (like KServe v2) using the default OpenAI parser. 

Fixes #1102 

```release-note
**The Problem:**
By default, the Helm chart configures the EPP to use the `openai-parser`. When standard ML workloads send non-LLM payloads (e.g., KServe v2 `inputs` arrays) to a standard Triton instance, the `openai-parser` fails to unmarshal the payload, resulting in a `400 Bad Request` and routing failure.
**The Proposed Change:**
This PR updates the Helm template logic to intelligently default to the `passthrough-parser` when standard ML servers are used.
 - Modifies `_config.yaml` to automatically inject `passthrough-parser` if `modelServerType` is set to `"triton"` (unless explicitly overridden by the user).
 - Updates public documentation (`README.md` and `values.yaml`) to officially list `"triton"` as a valid `modelServerType` option.
**Validation:**
 - Added explicit assertions to `hack/verify-helm.sh` to ensure `passthrough-parser` is injected into the rendered `EndpointPickerConfig` when testing the `"triton"` model server type.
 - All Helm validation tests (`make verify-helm-charts`) pass successfully.
```


andrewdoatgoogle:
manually validated in this project https://github.com/andrewaddo/gke-ig-ubb/blob/llmd-epp/MILESTONE_LLMD.md

ahg-g:
/approve

Thanks!

Issue #910, opened 2026-04-28
Refactor the Pod watching to fit the Discovery plugin


ezrasilvera:
Refactor the current kubernetes based discovery to become a discovery plugin according to #908 
Support two plugins InferencePool plugin and StaticPodSelector plugin

ahg-g:
Is this done now with https://github.com/llm-d/llm-d-router/pull/1218 merging?

ahg-g:
@ezrasilvera can we close this?

ezrasilvera:
@ahg-g I don't think so. This is to track  moving the existing k8s discovery to align with discovery plugin.

ahg-g:
ah, right, I mixed it up with the file based one.

ahg-g:
we are not aiming to do that in this release, I will remove it from the release tracker.

ezrasilvera:
Correct, that's a longer term item.  I was referring just to the issue itself.

PR #1218, opened 2026-05-19
FileDiscovery plugin (4/4) - wiring into the runner

ezrasilvera:
## Summary
This is the last PR to enable FileDIscovery plugin in EPP, once merged this will **allow users to run EPP (and llm-d) without Kubernetes.** 

Wire the file-discovery plugin (#1135) into the runner so the EPP can run without a Kubernetes cluster. When dataLayer.discovery.pluginRef` is set, the runner skips controller-manager setup, runs discovery + servers under a `RunnableGroup`, gates request-serving on `EndpointDiscovery.Ready()`, and warns about K8s-only features that are inactive in this mode.

The K8s path is unchanged behavior-wise (one cleanup commit avoids a double-parse of the EPP config).

### Changes to the discoveryEndpoint interface 
One function   `EndpointDiscovery.Ready()` was added to the interface.

In K8s mode the controller-runtime manager waits for cache sync before the first reconcile; in file-discovery mode there is no manager. Without a synchronization point the runner would start the ext_proc server while discovery is still inside the initial file load, and early requests  would land on an empty datastore (503 / "no candidates").

`Ready()` returns a channel closed once after the plugin's first successful sync. The runner gates `epp-server` and `health` on it
(metrics is not gated, so `/metrics` is observable during a slow startup). On initial-load failure the channel stays open and the runnable group cancels via `ctx.Done()`.

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -158,6 +158,11 @@ type Runner struct {
 	parser               fwkrh.Parser
 	dlRuntime            *datalayer.Runtime
 	PluginHandle         fwkplugin.Handle
+	// rawConfig caches the result of parseConfigurationPhaseOne. Run() peeks at
+	// it to decide between K8s and file-discovery paths; setup() also parses
+	// the same config. Caching avoids re-reading the file, re-registering
+	// plugins/feature gates, and re-emitting the data-layer setup logs.
+	rawConfig *configapi.EndpointPickerConfig
 }
 
 // WithExecutableName sets the name of the executable containing the runner.
@@ -589,6 +594,10 @@ func (r *Runner) registerInTreePlugins() {
 }
 
 func (r *Runner) parseConfigurationPhaseOne(ctx context.Context, opts *runserver.Options) (*configapi.EndpointPickerConfig, error) {
+	if r.rawConfig != nil {
+		return r.rawConfig, nil
+	}
+
 	logger := log.FromContext(ctx)
 
 	var configBytes []byte
@@ -627,6 +636,7 @@ func (r *Runner) parseConfigurationPhaseOne(ctx context.Context, opts *runserver
 		setupLog.Info("Data layer: ENABLED (default)")
 	}
 
+	r.rawConfig = rawConfig
 	return rawConfig, nil
 }
 
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -158,10 +158,7 @@ type Runner struct {
 	parser               fwkrh.Parser
 	dlRuntime            *datalayer.Runtime
 	PluginHandle         fwkplugin.Handle
-	// rawConfig caches the result of parseConfigurationPhaseOne. Run() peeks at
-	// it to decide between K8s and file-discovery paths; setup() also parses
-	// the same config. Caching avoids re-reading the file, re-registering
-	// plugins/feature gates, and re-emitting the data-layer setup logs.
+	// rawConfig caches the result of parseConfigurationPhaseOne.
 	rawConfig *configapi.EndpointPickerConfig
 }
 
@@ -594,6 +591,11 @@ func (r *Runner) registerInTreePlugins() {
 }
 
 func (r *Runner) parseConfigurationPhaseOne(ctx context.Context, opts *runserver.Options) (*configapi.EndpointPickerConfig, error) {
+	// parseConfigurationPhaseOne is idempotent: Run() calls it to decide
+	// between the K8s and file-discovery paths, and the K8s path's setup()
+	// then calls it a second time. Cache the parsed config so we don't
+	// re-read the file, re-register plugins/feature gates, or re-emit the
+	// data-layer setup logs.
 	if r.rawConfig != nil {
 		return r.rawConfig, nil
 	}
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -975,10 +975,19 @@ func (r *Runner) runWithFileDiscovery(ctx context.Context, opts *runserver.Optio
 	return g.Run(ctx)
 }
 
+// metricsShutdownTimeout bounds graceful shutdown of the metrics server so a
+// scraper holding a connection at process exit cannot block termination.
+const metricsShutdownTimeout = 5 * time.Second
+
 // serveMetrics starts a standalone Prometheus metrics HTTP server.
+//
+// EPP metrics are registered with controller-runtime's registry (see
+// pkg/epp/metrics.Register), not the prometheus default registry. The handler
+// must serve ctrlmetrics.Registry directly; promhttp.Handler() would expose only
+// Go runtime/process metrics and silently omit every EPP metric.
 func serveMetrics(ctx context.Context, port int, enablePprof bool) error {
 	mux := http.NewServeMux()
-	mux.Handle("/metrics", promhttp.Handler())
+	mux.Handle("/metrics", promhttp.HandlerFor(ctrlmetrics.Registry, promhttp.HandlerOpts{EnableOpenMetrics: true}))
 	if enablePprof {
 		mux.HandleFunc("/debug/pprof/", nhpprof.Index)
 		mux.HandleFunc("/debug/pprof/cmdline", nhpprof.Cmdline)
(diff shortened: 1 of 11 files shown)


cmd/epp/runner/runner.go
```diff
@@ -797,3 +824,229 @@ func resolvePoolNamespace(poolNamespace string) string {
...
+			name = spec.Type
+		}
+		if name != ref {
+			continue
+		}
+		factory, ok := fwkplugin.Registry[spec.Type]
+		if !ok {
+			return nil, fmt.Errorf("discovery: unknown plugin type %q for plugin %q", spec.Type, ref)
+		}
+		p, err := factory(name, spec.Parameters, nil)
```
elevran: This re-instantiates the discovery plugin even though `parseConfigurationPhaseTwo` (called a few lines above) already created it via `loader.InstantiateAndConfigure` and registered it in `r.PluginHandle`. Two instances coexist, **only this second one** has `Start()` called. While no harm today (FileDiscovery's factory has no side effects), but it's confusing and breaks the moment a discovery plugin's factory does anything stateful (open a connection, claim a port, etc.). It also passes a `nil` for `Handle` here while the previous instance got the real one handle.

it would be better to look up the existing instance from `r.PluginHandle` and type-asserting to `fwkdl.EndpointDiscovery` to find the one you need.

elevran (commented):
Wiring is sound and tests pass. One design concern worth addressing before merge, see inline.

/lgtm overall. Leaving approval for later.

cmd/epp/runner/runner.go
ezrasilvera: @elevran Good catch.  I fixed in fe574571
I also run local e2e test that verified with the new resolveDiscovery:

1. Plugins instantiated once by the loader (Instantiated all plugins... log includes file-discovery).
2. EPP starting (file discovery mode) logs discoveryPlugin: file-discovery/file-discovery the loader-created instance, looked up  via r.PluginHandle.
3. watching endpoints file for changes: proves Start() fired on that same instance (the only one now).
4. Verified my test got "The capital of France is Paris." back through Envoy -> EPP -> vLLM.

cmd/epp/runner/runner.go
```diff
@@ -572,9 +592,15 @@ func (r *Runner) registerInTreePlugins() {
...
 	fwkplugin.Register(concurrency.ConcurrencyDetectorType, concurrency.ConcurrencyDetectorFactory)
 	fwkplugin.Register(utilization.UtilizationDetectorType, utilization.UtilizationDetectorFactory)
+	// register discovery plugins
+	fwkplugin.Register(discoveryfile.PluginType, discoveryfile.Factory)
 }
 
 func (r *Runner) parseConfigurationPhaseOne(ctx context.Context, opts *runserver.Options) (*configapi.EndpointPickerConfig, error) {
+	if r.rawConfig != nil {
+		return r.rawConfig, nil
+	}
```
ahg-g, line 599: pls add a comment that this can happen in the case the function is invoked multiple times.

cmd/epp/runner/runner.go
```diff
@@ -797,3 +824,219 @@ func resolvePoolNamespace(poolNamespace string) string {
...
+//
+// EPP metrics are registered with controller-runtime's registry (see
+// pkg/epp/metrics.Register), not the prometheus default registry. The handler
+// must serve ctrlmetrics.Registry directly; promhttp.Handler() would expose only
+// Go runtime/process metrics and silently omit every EPP metric.
+func serveMetrics(ctx context.Context, port int, enablePprof bool) error {
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.HandlerFor(ctrlmetrics.Registry, promhttp.HandlerOpts{EnableOpenMetrics: true}))
+	if enablePprof {
+		mux.HandleFunc("/debug/pprof/", nhpprof.Index)
```
ahg-g: lets define a function in https://github.com/llm-d/llm-d-router/blob/main/pkg/common/observability/profiling/pprof.go that returns a `map[string]http.HandlerFunc`  of handlers to register, and have this function called here and in the other place pls.

cmd/epp/runner/runner.go
```diff
@@ -797,3 +824,219 @@ func resolvePoolNamespace(poolNamespace string) string {
...
+	disc, ok := p.(fwkdl.EndpointDiscovery)
+	if !ok {
+		return nil, fmt.Errorf("discovery: plugin %q does not implement EndpointDiscovery", ref)
+	}
+	return disc, nil
+}
+
+// runWithFileDiscovery handles the execution path when a discovery plugin is configured.
+// It builds the EPP server stack without a Kubernetes cluster or controller manager.
+func (r *Runner) runWithFileDiscovery(ctx context.Context, opts *runserver.Options, rawConfig *configapi.EndpointPickerConfig) error {
```
ahg-g, line 846: Is this path covered in any integration tests?

cmd/epp/runner/runner.go
```diff
@@ -797,3 +824,219 @@ func resolvePoolNamespace(poolNamespace string) string {
...
+	}
+	disc, ok := p.(fwkdl.EndpointDiscovery)
+	if !ok {
+		return nil, fmt.Errorf("discovery: plugin %q does not implement EndpointDiscovery", ref)
+	}
+	return disc, nil
+}
+
+// runWithFileDiscovery handles the execution path when a discovery plugin is configured.
+// It builds the EPP server stack without a Kubernetes cluster or controller manager.
```
ahg-g, line 845: Are you sure we can't use controller manager to manage go routines (which seems to be the reason for this special path)? I think we can if we pass an empty `rest.Config`, and that will help us reduce all this complexity.

Note that in https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/server/controller_manager.go#L47 we already have a flag to not watch on infpool/infobj/infModelRewrite resources in standalone mode, and we can have another flag to disable that for pods as well by passing another flag, call it `nonk8s`.

cmd/epp/runner/runner.go
ezrasilvera: fixed in ca3910f4

cmd/epp/runner/runner.go
ezrasilvera: Good idea.  Done in eb060213

cmd/epp/runner/runner.go
ezrasilvera: Not really. I was testing it manually on my local system, but you are correct that integration test is needed.  I put something which I believe is good enough for the current state (at 8fa1baf7) . In the future we would probably want to add a more extensive test or even  reuse test/integration/epp envtest harness.
Fixed in 8fa1baf7

cmd/epp/runner/runner.go
ezrasilvera: > Are you sure we can't use controller manager to manage go routines (which seems to be the reason for this special path)? I think we can if we pass an empty `rest.Config`, and that will help us reduce all this complexity.
> 
> Note that in https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/server/controller_manager.go#L47 we already have a flag to not watch on infpool/infobj/infModelRewrite resources in standalone mode, and we can have another flag to disable that for pods as well by passing another flag, call it `nonk8s`.

@ahg-g Thanks for the feedback. I think that for now we should stay with the current design. 
I looked into this and technically speaking your suggestion does work: the cache isn't doing anything without informers, leader election early-returns when disabled, and the lazy recorder/REST mapper only fire on first use.

However, the savings are very small (62 lines of rungroup.go, an errgroup.Group wrapper) and I'm not sure the costs worth it. 

- First, from a design perspective - we plan eventually to move K8s itself behind a discovery plugin, the runner becomes plugin-agnostic and the K8s plugin owns its own Manager internally. The current RunnableGroup design is the right preparation; the Manager-unification proposal  actually goes in the opposite direction towards k8s.
- Basically the the statement of "This path doesn't touch K8s" moves from compile-time absence (no mgr variable) to a runtime negative-test. Future code that calls mgr.GetClient() or fires an event would fail at first request, not at build.
- We couples  non-K8s mode to controller-runtime's manager+cluster+cache+client+recorder, when today it only depends on the Runnable interface. 
- I think it will mess up code organization. Today we have two clean paths:
   - setup() -- K8s mode,
   - runWithFileDiscovery() -- file mode
For a specific mode you reads one function, top to bottom.  The proposal merges them into one setup() that decides at every step which flavor to do. To express both modes in one function you need three flags simultaneously

So currently I prefer we keep the split, but happy to revisit if a third deployment mode lands or once K8s-as-plugin happens

ahg-g:
Sounds good, do we have an issue to track moving k8s discovery to a plugin?

ezrasilvera:
> Sounds good, do we have an issue to track moving k8s discovery to a plugin?

Yes. #910 I added more info into it.

PR #1221, opened 2026-05-19
support mooncake conenctor in sidecar

### LINKED ISSUE #1220: add support for mooncake connector in sidecar
# Mooncake P/D support in the routing sidecar

The sidecar currently supports NIXL, shared-storage, and SGLang KV connectors for prefill-decode disaggregation. Mooncake uses a different protocol: the prefiller returns no kv_transfer_params, so the sidecar must generate a `transfer_id`, discover the prefiller's engine_id via its bootstrap server (/query), and construct the decode-side params itself.

Without sidecar support, requests fall through to the decoder as standalone inference — no KV transfer occurs and prefiller pods sit idle.

Implementation: `--kv-connector=mooncake` flag, bootstrap discovery with LRU caching, env-configurable port (VLLM_MOONCAKE_BOOTSTRAP_PORT, default 8998).

cc @praveingk

Gregory-Pereira:
**What type of PR is this?**

/kind feature

Enable support for the mooncake connector.

Without the sidecar doing the P/D coordination, what happens is:

  1. The EPP picks a prefiller and a decoder, but the request goes directly to the decoder (port 8200) without any prefill stage and the sidecar just passes it through because there's no `x-prefiller-host-port` header it recognizes, or the old nixlv2 connector doesn't know how to construct mooncake's `kv_transfer_params`.
  2. So the decoder runs the both prefill and decode itself as if it were a standalone vLLM instance. You get a correct response, but you're not actually doing disaggregated P/D. The prefiller pods sit idle (or handle their own standalone requests).
  3. The kv_transfer_params: null in your curl response confirms this - no KV transfer happened. The decoder did all the work.

Fixes #1220 

```release-note
Adds alternative connector support for the routing sidecar mapping `MooncakeConenctor`'s KVTransferParams
```
cc @praveingk

### DIFF
--- a/pkg/sidecar/proxy/proxy.go
+++ b/pkg/sidecar/proxy/proxy.go
@@ -68,6 +68,10 @@ const (
 	requestFieldBootstrapPort = "bootstrap_port"
 	requestFieldBootstrapRoom = "bootstrap_room"
 
+	// Mooncake fields
+	requestFieldTransferID          = "transfer_id"
+	requestFieldRemoteBootstrapAddr = "remote_bootstrap_addr"
+
 	// KVConnectorNIXLV2 enables the P/D KV NIXL v2 protocol
 	KVConnectorNIXLV2 = "nixlv2"
 
@@ -76,6 +80,9 @@ const (
 
 	// KVConnectorSGLang enables SGLang the P/D KV disaggregation protocol
 	KVConnectorSGLang = "sglang"
+
+	// KVConnectorMooncake enables the Mooncake P2P KV transfer protocol
+	KVConnectorMooncake = "mooncake"
 
 	// ECExampleConnector enables the Encoder disaggregation protocol (E/PD, E/P/D)
 	ECExampleConnector = "ec-example"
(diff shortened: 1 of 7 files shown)


pkg/sidecar/proxy/proxy.go
```diff
@@ -77,6 +81,9 @@ const (
 	// KVConnectorSGLang enables SGLang the P/D KV disaggregation protocol
 	KVConnectorSGLang = "sglang"
 
+	// KVConnectorMooncake enables the Mooncake P2P KV transfer protocol
+	KVConnectorMooncake = "mooncake"
```
robertgshaw2-redhat, line 85: we should probably think about what we want the end state naming convention to be and follow it for all new connectors

for instance:
- `sglang` is a bad name, it should be `sglang-nixl`
- `nixlv2` is a bad name, it should be `vllm-nixl`

I think we should decide what we want the naming scheme to be for end state, and add all new connectors in this way (while we migrate existing ones)

something like:
- `vllm-mooncake-v0` or something would work well to start

Gregory-Pereira:
closing in favor of #1193, will help try to review there

Issue #1182, opened 2026-05-18
[Flow Control] Tenant Fairness: Virtual Token Count (VTC) and Dispersion Metrics


LukeAVanDrie:
Flow control manages resource sharing between competing tenants (flows) within the same priority band. Our existing implementation, Round Robin (`round-robin-fairness-policy`), assumes every request consumes the exact same resources. However, in LLM serving, resource consumption is highly variable due to long context prompts and autoregressive decode loops. Naive turn sharing allows "noisy neighbors" to monopolize GPU capacity and starve other tenants.

We need a Quantity-of-Service fair scheduler to enforce resource equity based on actual compute work (tokens), along with standard metrics to measure tenant dispersion.

---

### 1. Implement Virtual Token Count (VTC) Fairness Policy

We will implement a token-level fair scheduler named `vtc-fairness-policy` implementing EPP's pluggable `FairnessPolicy` interface. Instead of round-robin turn sharing, VTC schedules requests along a continuous virtual timeline.

**Technical Requirements:**
*   Flyweight State: Maintain a map of tenant ID to accumulated virtual token count (`counters map[string]float64`) inside the scoped state of the Priority Band (`vtcBandState`), complying with EPP's Flyweight plugin pattern.
*   Scheduling Rule: The scheduler must always dispatch the active tenant queue $i$ with the minimum virtual counter:
    $$\text{Next Flow} = \arg\min_{i \in \text{active}} c_i$$
*   Counter Lift (Starvation Prevention): When a queue transitions from empty to active (rejoining the pool), lift its virtual counter to the minimum of all other currently active queues to prevent the rejoining tenant from starving the rest of the pool:
    $$c_i \leftarrow \max\left(c_i, \min_{j \in \text{active}, j \neq i} c_j\right)$$
*   Cost Function:
    $$\text{Cost}(req) = w_p \cdot \text{input tokens} + w_q \cdot \text{output tokens}$$
    Accept configurable parameters `inputTokenWeight` (default `1.0`) and `outputTokenWeight` (default `3.0` to reflect that autoregressive decode tokens consume sequentially more cache and execution time).
*   Cost Estimation: Estimate input tokens via client hints (`InputTokenCountHint()`), tokenized prompt length, or character-based estimation fallback (`PromptText() / 4` or `PayloadSize / 4`). Use a default output token length (e.g., `128`) or parse `"max_tokens"` from the raw payload map.
[...]

PR #1222, opened 2026-05-19
feat(flowcontrol): add VTC fairness policy with token-based cost and starvation prevention

gyliu513:
**What type of PR is this?**

/kind feature

Implements the `vtc-fairness-policy` plugin as described in #1182 (Task 1).

Unlike `round-robin-fairness-policy` (which treats every request equally regardless of size), VTC tracks a cumulative virtual token cost per flow. The flow with the lowest virtual counter wins the next dispatch slot, ensuring large-prompt tenants don't starve small-prompt tenants.

/cc @LukeAVanDrie 

Related to task1 for #1182

```release-note
feat(flowcontrol): add VTC fairness policy with token-based cost and starvation prevention
```


pkg/epp/framework/plugins/flowcontrol/fairness/vtc/vtc.go
```diff
@@ -0,0 +1,314 @@
...
+		return 1
+	}
+	return cost
+}
+
+// estimateInputTokens estimates the input token count from the request via a cascade of heuristics.
+//
+// Priority order:
+//  1. Client-provided hint via InputTokenCountHint() (e.g., from pre-tokenized completions prompt).
+//  2. Length of a pre-tokenized prompt in InferenceRequestBody.TokenizedPrompt.
```
roytman, line 275: There is a PR https://github.com/llm-d/llm-d-router/pull/1248, which adds support to `inference/v1/generate`. Depends on which PR will be merged first, this part should be updated.
 Can #1303 be relevant ?

gyliu513: Thanks for the heads-up, @roytman!

For #1248: good catch. The `/inference/v1/generate` endpoint uses `GenerateRequest{TokenIDs}`, which may not surface through the existing `ir.Body.TokenizedPrompt` path in `estimateInputTokens`. Depending on merge order, I'll update the cascade to handle that format when #1248 lands (or if it merges first, I'll add support before this one is merged).

For #1303: agree that centralizing token estimation is the right long-term direction. The current approach in vtc is a local heuristic to avoid a hard dependency, but if #1303 moves forward with a shared estimation backend, this function should be replaced to use that.

ahg-g:
@gyliu513 what is the status here? sorry this fell through the cracks, but if you think this is still relevant and show some evaluation to better understand its potential vs existing policies, I will be happy to review it.

LukeAVanDrie:
> @gyliu513 what is the status here? sorry this fell through the cracks, but if you think this is still relevant and show some evaluation to better understand its potential vs existing policies, I will be happy to review it.

This is on me. We talked and paused this since there was a lot of churn on the inflight tokens data producer and I wanted to wait for that to land so we can leverage it here. Will do a fresh review today and suggest some short-term pragmatic fixes here as the data architecture is still not ideal for flow control plugins.

PR #1231, opened 2026-05-20
refactor: updated v1/models to implement ProducerPlugin, moved its data definition to attribute

### LINKED ISSUE #1202: [Refactor] The v1/models attribute should follow directory pattern of other endpoint attributes
The v1/models data extractor is effectively a data producer but is not aligned with other producers.
- needs to implement `ProducerPlugin`
- the attribute definition should move to `pkg/epp/framework/plugins/datalayer/attribute/models` alongside other attributes

irar2:
**What type of PR is this?**

/kind cleanup

The v1/models data extractor needs to implement ProducerPlugin and its attribute definition should move to pkg/epp/framework/plugins/datalayer/attribute/models alongside other attributes

Fixes #1202

```release-note
NONE
```


pkg/epp/framework/plugins/datalayer/extractor/models/extractor.go
```diff
@@ -65,15 +31,17 @@ var (
 // ModelExtractor implements the models extraction.
 type ModelExtractor struct {
 	typedName fwkplugin.TypedName
+	dk        fwkplugin.DataKey
```
shmuelk, line 34: Why don't you use the global key ModelsAttributeKey

irar2: Fixed

pkg/epp/framework/plugins/datalayer/attribute/models/data_types.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// Clone returns a full copy of the object
+func (m ModelDataCollection) Clone() fwkdl.Cloneable {
+	if m == nil {
+		return nil
+	}
+	clone := make([]ModelData, len(m))
+	copy(clone, m)
+	return (*ModelDataCollection)(&clone)
```
elevran: The `(*ModelDataCollection)(&clone)` conversion is unnecessarily roundabout, and returning a pointer from a value-receiver `Clone()` is the source of the `Produces()` inconsistency flagged on `extractor.go`. Declaring `clone` as the named slice type lets you return it directly:

```suggestion
	clone := make(ModelDataCollection, len(m))
	copy(clone, m)
	return clone
```

With this, `Clone()` returns a `ModelDataCollection` value, matching what `Produces()` advertises and what consumers will receive from `Attributes.Get()`.

pkg/epp/framework/plugins/datalayer/extractor/models/extractor.go
```diff
@@ -103,6 +71,13 @@ func (me *ModelExtractor) Extract(_ context.Context, data any, ep fwkdl.Endpoint
...
-	ep.GetAttributes().Put(ModelsAttributeKey, ModelDataCollection(models.Data))
+	ep.GetAttributes().Put(me.dk.String(), attrmodels.ModelDataCollection(models.Data))
 	return nil
 }
+
+// Produces returns data produced by the producer.
+// This is a map from data key (string) produced to
+// the data type of the key (represented as data with default value casted as any field).
+func (me *ModelExtractor) Produces() map[fwkplugin.DataKey]any {
+	return map[fwkplugin.DataKey]any{me.dk: attrmodels.ModelDataCollection{}}
```
elevran, line 82: `Produces()` advertises `ModelDataCollection{}` (value), but `ModelDataCollection.Clone()` returns `*ModelDataCollection`. `Attributes.Get()` returns the result of `Clone()`, so consumers receive a pointer, not a value.

This matters because `pkg/epp/datalayer/data_graph.go:198` type-checks producer/consumer compatibility with `reflect.TypeOf` and any consumer declaring either type would mismatch one of the two. The fix is in `data_types.go:54` (return the value); no change needed here once that's done.

elevran (commented):
Two related concerns around the `Produces()` and`Clone()` type contract, see inline comments.
Also: other attribute subdirs (`concurrency/`, `latency/`, `prefix/`) ship a `README.md` — worth adding one for `attribute/models/` for symmetry, but non-blocking.

pkg/epp/framework/plugins/datalayer/attribute/models/data_types.go
irar2: Fixed

pkg/epp/framework/plugins/datalayer/extractor/models/extractor.go
irar2: Fixed in data_types

irar2:
Added README.md

elevran (approved):
/lgtm
/approve

PR #1234, opened 2026-05-20
Update simulator version to v0.9

mayabar:
**What type of PR is this?**
/kind cleanup

Use latest version of llm-d-inference-sim

```release-note
Update the default vLLM and simulator images, and remove the UDS Tokenizer and the `UDS_TOKENIZER_IMAGE` environment variable. Use `VLLM_RENDER_IMAGE` environment variable to define the render image name.
```

### DIFF
--- a/config/manifests/vllm/sim-deployment.yaml
+++ b/config/manifests/vllm/sim-deployment.yaml
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
       - name: vllm-sim
-        image: ghcr.io/llm-d/llm-d-inference-sim:latest
+        image: ghcr.io/llm-d/llm-d-inference-sim:v0.9.0
         imagePullPolicy: Always
         args:
         - --model
--- a/test/e2e/utils_test.go
+++ b/test/e2e/utils_test.go
@@ -143,7 +143,8 @@ func runKustomize(kustomizeDir string) []string {
 	session, err := gexec.Start(command, nil, ginkgo.GinkgoWriter)
 	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
 	gomega.Eventually(session).WithTimeout(600 * time.Second).Should(gexec.Exit(0))
-	return strings.Split(string(session.Out.Contents()), "\n---")
+	res := strings.Split(string(session.Out.Contents()), "\n---")
+	return res
 }
 
 // removeEmptyArgs strips YAML list items that are empty strings after variable
--- a/test/e2e/utils_test.go
+++ b/test/e2e/utils_test.go
@@ -143,8 +143,7 @@ func runKustomize(kustomizeDir string) []string {
 	session, err := gexec.Start(command, nil, ginkgo.GinkgoWriter)
 	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
 	gomega.Eventually(session).WithTimeout(600 * time.Second).Should(gexec.Exit(0))
-	res := strings.Split(string(session.Out.Contents()), "\n---")
-	return res
+	return strings.Split(string(session.Out.Contents()), "\n---")
 }
 
 // removeEmptyArgs strips YAML list items that are empty strings after variable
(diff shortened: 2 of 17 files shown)


config/manifests/vllm/sim-deployment.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
       - name: vllm-sim
-        image: ghcr.io/llm-d/llm-d-inference-sim:latest
+        image: ghcr.io/llm-d/llm-d-inference-sim:v0.9.0
```
elevran, line 18: 🙏 can you please add (not necessarily as part of this PR) a "script" (can be `go run ...`) and makefile target to ensure images are not defined to use `latest`?

test/e2e/utils_test.go
```diff
@@ -187,6 +187,98 @@ func removeEmptyLabels(inputs []string) []string {
...
 }
 
+func isModelReal(modelName string) bool {
+	url := "https://huggingface.co/api/models/" + modelName
+
+	resp, err := http.Get(url)
+	if err != nil {
+		return false
+	}
+	defer resp.Body.Close()
```
elevran, line 202: is defer `resp.Body.Close()` needed if err != nil? Will it be called?

test/e2e/utils_test.go
```diff
@@ -187,6 +187,98 @@ func removeEmptyLabels(inputs []string) []string {
...
+
+	return resp.StatusCode == http.StatusOK
+}
+
+// removeSidecarIfSimulated strips the vllm-render init container and the orphaned
+// model-cache volume from all manifests when running against a simulated (non-real) model.
+func removeSidecarIfSimulated(inputs []string) []string {
```
elevran: A meaningful simplification is to parse the manifest as YAML (e.g. `sigs.k8s.io/yaml` into `unstructured.Unstructured` or a typed `appsv1.Deployment`), filter `spec.template.spec.initContainers` and `spec.template.spec.volumes` by name, and marshal back. That handles inline-vs-spread `name:`, quoted scalars, comments, and tab-indented inputs without any of the indent edge cases the current code has to special-case. Same input/output shape (`[]string`), much smaller surface to be wrong on.

elevran (commented):
Two things worth a second look before merge:

**Release note.** Marked `NONE`, but this PR bumps the default vllm and simulator images and renames the documented env var `UDS_TOKENIZER_IMAGE` to `VLLM_RENDER_IMAGE` (with a different default image). Both are listed in `deploy/environments/dev/README.md` as user-overridable variables, so anyone scripting against these will silently break. Worth a one-liner release note flagging the var rename and the image bumps.

`removeSidecarIfSimulated` complexity - see inline comment below.

config/manifests/vllm/sim-deployment.yaml
mayabar: Will do this as a separate PR, issue for this - #1306

test/e2e/utils_test.go
mayabar: No, the defer is placed after the if err != nil { return false } early return, so it is never registered when there's an error and is only called on the success path where resp is guaranteed non-nil. This behavior matches other tests using http.Get, which only close the body on success.

mayabar:
> Two things worth a second look before merge:
> 
> **Release note.** Marked `NONE`, but this PR bumps the default vllm and simulator images and renames the documented env var `UDS_TOKENIZER_IMAGE` to `VLLM_RENDER_IMAGE` (with a different default image). Both are listed in `deploy/environments/dev/README.md` as user-overridable variables, so anyone scripting against these will silently break. Worth a one-liner release note flagging the var rename and the image bumps.
> 
Updated

test/e2e/utils_test.go
mayabar: Changed to more structured yaml update implementation

test/e2e/utils_test.go
```diff
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"time"
 
+	"gopkg.in/yaml.v3"
```
elevran: Per their [github](https://github.com/go-yaml/yaml/tree/v3.0.1) this is archived and no longer maintained. Please use the [k8s YAML package](https://github.com/kubernetes-sigs/yaml) instead,

mayabar: removed

test/e2e/utils_test.go
mayabar: @elevran Cleanup implementation updated to use `unstructured.Unstructured`

elevran (approved):
/lgtm
/approve

PR #1237, opened 2026-05-20
test(epp): add unit regression test for DynamicMetadata encode path

adelsam:
Adds TestGenerateResponseBodyResponses_DynamicMetadata to verify that generateResponseBodyResponses correctly attaches DynamicMetadata to the final ProcessingResponse chunk. This would have caught the bug where plugin-computed metadata was silently dropped before reaching Envoy.

**What type of PR is this?**

/kind test

Adding a Unit test for DynamicMetadata in ResponseBody processing.

```release-note
NONE
```


zetxqx:
For reference, this is testing this logic: https://github.com/llm-d/llm-d-router/blob/3a160b650430ab6cabbf625bde809b96e134ceba/pkg/epp/handlers/response.go#L105-L108

/lgtm

PR #1247, opened 2026-05-21
Consolidate flow control and request handler parameters in the config api

### LINKED ISSUE #1041: [Configuration] Refactor the EndpointPickerConfig to align with the high-level EPP components

Refactor the EndpointPickerConfig to group the parameters under one of the four primary components of EPP:
- RequestHandler
- FlowControl 
- Scheduler
- Datalayer

**Why is this needed**:

Currently we have some configurations under a top level component struct and some are outside. This refactor helps create a clear pattern it also makes it easier to document.

ahg-g:
**What type of PR is this?**
/kind cleanup

Refactors the config api to align with the four main components of the EPP: 
- RequestHandler
- RequestScheduler
- FlowControl
- Datalayer

Note that we don't need to keep backward compatibility since we are graduating the api to v1 and so users needs to adapt anyway.

Better UX and easier to align with docs

Part of https://github.com/llm-d/llm-d-router/issues/1041

```release-note
The EndpointPickerConfig API has been refactored to provide a more structured and logical grouping of configuration fields. These changes improve schema clarity and provide a cleaner foundation for future feature extensions.

Existing YAML/JSON configuration files must be updated to reflect the new nested structure, old fields will continue to work for two releases:

- Saturation Detector Migration: The SaturationDetector field has been moved from the top-level configuration into the FlowControl block.
- Parser Encapsulation: A new requestHandler struct has been introduced to house request-handler component configurations. The parser field has been moved into this new block.
```


shmuelk:
@ahg-g Please run `make generate` to update the generated files.

LukeAVanDrie:
I think we should keep #1041 open until we have also updated the docs in the main llm-d repo.

ahg-g:
> I think we should keep #1041 open until we have also updated the docs in the main llm-d repo.

yes, I changed to "part of" instead of "fixes"

elevran:
is it possible/worthwhile to make this a regular deprecation (ie accept both old and new formats) instead of a breaking change?
Just the config parsing that affects user facing format, internally can use the "new" Go struct.

ahg-g:
> is it possible/worthwhile to make this a regular deprecation (ie accept both old and new formats) instead of a breaking change? Just the config parsing that affects user facing format, internally can use the "new" Go struct.

I don't think it is worthwhile, the changes are to components that has just landed and are not yet used. Note that we are graduating to v1, so technically we are allowed to make breaking changes from v1alpha to v1.

ahg-g:
@elevran I added the fields back with deprecation notices.

pkg/epp/config/loader/configloader.go
```diff
@@ -90,6 +90,26 @@ func LoadRawConfig(configBytes []byte, logger logr.Logger) (*configapi.EndpointP
 				"replacement", "llm-d.ai/v1alpha1/EndpointPickerConfig")
 		}
 
+		if rawConfig.SaturationDetector != nil {
+			logger.Info("DEPRECATION: top-level saturationDetector is deprecated, use flowControl.saturationDetector instead")
+			if rawConfig.FlowControl == nil {
+				rawConfig.FlowControl = &configapi.FlowControlConfig{}
+			}
+			if rawConfig.FlowControl.SaturationDetector == nil {
```
elevran, line 100: nit: should we log that both defined and using the field under `FlowControl.SaturationDetector`?
Same for parser.

ahg-g:
Needs another approve pls after addressing the lint issues.

elevran (approved):

PR #1248, opened 2026-05-21
feat(epp): add /inference/v1/generate endpoint support

### LINKED ISSUE #1355: [Feature] Support vLLM /inference/v1/generate endpoint by the llm-d-router
To support the "tokens-in, tokens-out" initiative for disaggregated deployments, vLLM provides a raw token entry point: `/inference/v1/generate`. This endpoint accepts pre-tokenized inputs (IDs) directly rather than raw text strings.

We would like to extend the EPP to support natively and route requests to this /inference/v1/generate endpoint.

**Why is this needed**:
In disaggregated LLM serving architectures (where multi-modal embedding, prompt processing/prefill and token generation are split, or orchestration handles tokenization centrally), passing text back and forth introduces redundant tokenization and detokenization overhead.

Supporting this endpoint in EPP will:

- Eliminate Redundant Operations: Allow downstream components to consume tokens directly, saving CPU cycles.
- Exact and Consistent Token Lengths: Instead of relying on heuristics to guess the prompt token count, EPP will receive the exact and consistent token length directly. This guarantees deterministic request scheduling, and accurate KV-cache allocation.

dmitripikus:
## Summary

Adds support for routing the vLLM disaggregated Prefill/Decode `/inference/v1/generate`
endpoint through EPP. Unlike the OpenAI-compatible endpoints, this API accepts
pre-tokenized input (`token_ids`) and an optional multimodal `features` block carrying
per-modality content hashes and placeholder ranges.

### Parser

- **New `vllmhttp` parser** (`pkg/epp/framework/plugins/requesthandling/parsers/vllmhttp/`):
  embeds the OpenAI parser and handles `/inference/v1/generate` locally; all other paths
  delegate to `openai.OpenAIParser`. Plugin type: `vllmhttp-parser`. Registered in
  `cmd/epp/runner/runner.go` alongside the existing parsers.
  Explicit parser configuration is needed in the configuration map.

### Types

- New `GenerateRequest` on `InferenceRequestBody` with fields:
  - `TokenIDs []uint32` — pre-tokenized input.
  - `Features *tokenization.MultiModalFeatures` — multimodal hashes + placeholder ranges
    in the upstream kv-cache library shape.
  - `CacheSalt string` — optional prefix-cache isolation.
- Custom `UnmarshalJSON` decodes the wire `features` block (snake_case `mm_hashes` /
  `mm_placeholders` with `offset`/`length`) into `tokenization.MultiModalFeatures`, and
  validates token IDs (rejects negative, NaN, non-integer, or `> math.MaxUint32`).
- `String()` prints `{TokenIDsCount: N, MMHashes: {modality:count, ...}}` with
  alphabetically-sorted modality keys for stable log output.
- `PromptText`, `InputTokenCountHint`, and `CacheSalt` accessors handle the generate case.

### Scheduling / scoring

- **Tokenizer DataProducer**: skips the UDS/vLLM tokenizer call for generate requests
  and flattens `Generate.Features` into `TokenizedPrompt.MultiModalFeatures` via the
  existing `convertMMFeaturesToUpstream` helper, so downstream scorers see image hashes
  without any per-protocol code paths.
- **Approximate-prefix hashing**: serializes token IDs as little-endian `uint32` bytes
  for cache-key generation.
- **Precise-prefix-cache scorer** (`preciseprefixcache/precise_prefix_cache.go`):
  - New `extractTokensAndFeatures` helper unifies the `TokenizedPrompt` and
    `Body.Generate` token-source paths in both `computeBlockKeys` and `getScores`.
  - New `generateExtraFeatures` helper derives `kvblock.BlockExtraFeatures` from
    `Body.Generate.Features` so MM-hashes are folded into KV-block keys even when the
    tokenizer plugin is not configured (previously the Generate fallback dropped them,
    causing two requests with identical `token_ids` but different images to collide).

Fixes [#1355](https://github.com/llm-d/llm-d-router/issues/1355)

### E2E test:
E2E test is implemented in separate PR [#1356](https://github.com/llm-d/llm-d-router/pull/1356)

```release-note
New `/inference/v1/generate` endpoint is added, that accepts pre-tokenized prompts (`token_ids`) and optional multimodal features (image/audio/video hashes and placeholder ranges). To enable, configure the new `vllmhttp-parser` (Helm value `router.epp.parser=vllmhttp-parser`, or set `parser: vllmhttp-parser` in EPP configuration). The parser handles `/inference/v1/generate` locally and delegates all other paths to the OpenAI parser, so a single instance covers both vLLM-specific and OpenAI-compatible HTTP traffic on the same endpoint. Existing `openai-parser` deployments are unaffected and need no changes unless `/inference/v1/generate` support is desired.
```


pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -424,6 +434,40 @@ func (e *EmbeddingsRequest) String() string {
 	return fmt.Sprintf("{InputType: %T}", e.Input)
 }
 
+// GenerateRequest is a structured representation of the fields we parse out of the vLLM
+// disaggregated Prefill/Decode API at /inference/v1/generate.
+// Unlike the OpenAI-compatible endpoints, this API accepts pre-tokenized input (token IDs).
+// This struct includes fields usable for plugins and scheduling decisions.
+type GenerateRequest struct {
```
vMaroon, line 445: multi-modal features are also needed here.

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -73,6 +73,8 @@ type InferenceRequestBody struct {
 	Conversations *ConversationsRequest `json:"conversations,omitempty"`
 	// EmbeddingsRequest is the representation of the OpenAI /v1/embeddings request body.
 	Embeddings *EmbeddingsRequest `json:"embeddings,omitempty"`
+	// GenerateRequest is the representation of the vLLM native /generate request body.
```
roytman: it is `inference/v1/generate` request body

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -424,6 +434,40 @@ func (e *EmbeddingsRequest) String() string {
 	return fmt.Sprintf("{InputType: %T}", e.Input)
 }
 
+// GenerateRequest is a structured representation of the fields we parse out of the vLLM
+// disaggregated Prefill/Decode API at /inference/v1/generate.
```
roytman: I'd not explicitly mention Prefill/Decode API, just "...we parse out of the vLLM disaggregated request at /inference/v1/generate."
Even now, we use it for encoding also.

pkg/epp/framework/interface/requesthandling/types.go
roytman: @dmitripikus , the fill request structure is 
```python
request_id (str, auto-generated UUID)
token_ids (list[int]) - the token IDs to generate from
features (MultiModalFeatures | None) - multimodal hashes/placeholders
sampling_params (SamplingParams)
model (str | None)
stream (bool | None)
stream_options (StreamOptions | None)
cache_salt (str | None)
priority (int)
```
you can see its definition at https://github.com/vllm-project/vllm/blob/0fa888465e5a30b797bdf2cdcd0f57fc77541cef/vllm/entrypoints/serve/disagg/protocol.py#L62

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -424,6 +434,40 @@ func (e *EmbeddingsRequest) String() string {
...
+	TokenIDs []uint32 `json:"token_ids"`
+	// CacheSalt is an optional request parameter to isolate prefix caches for security reasons.
+	CacheSalt string `json:"cache_salt,omitempty"`
+}
+
+func (r *GenerateRequest) String() string {
+	if r == nil {
+		return nilStr
+	}
+	return fmt.Sprintf("{TokenIDsCount: %d}", len(r.TokenIDs))
```
roytman: The string representation should be extended

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -424,6 +434,40 @@ func (e *EmbeddingsRequest) String() string {
...
+		TokenIDs  []float64 `json:"token_ids"`
+		CacheSalt string    `json:"cache_salt,omitempty"`
+	}
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+	r.CacheSalt = raw.CacheSalt
+	r.TokenIDs = make([]uint32, len(raw.TokenIDs))
+	for i, v := range raw.TokenIDs {
+		r.TokenIDs[i] = uint32(v)
```
roytman, line 503: the code decodes token_ids as []float64 then truncates to uint32 with no validation. 
Standard tokenization and model vocabularies, token IDs are fundamentally conceptually non-negative, but it is better to validate the user input

```suggestion
		if v < 0 || v > math.MaxUint32 || v != math.Trunc(v) {
            return fmt.Errorf("token_ids[%d]: invalid value %v", i, v)
        }
        r.TokenIDs[i] = uint32(v)
```

pkg/epp/framework/plugins/scheduling/scorer/contextlengthaware/context_length_aware.go
roytman: Why not use early return 

```suggestion
/ /Generate requests carry pre-tokenized input — use exact count directly.
if request.Body.Generate != nil {
    return len(request.Body.Generate.TokenIDs)
}
```

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -156,6 +156,9 @@ func getUserInputLenInTokens(request *scheduling.InferenceRequest) (int, error)
 	if request.Body.Completions != nil {
 		return len(request.Body.Completions.Prompt.Raw) / AverageCharactersPerToken, nil
 	}
+	if request.Body.Generate != nil {
+		return len(request.Body.Generate.TokenIDs), nil
+	}
 	if request.Body.ChatCompletions == nil {
 		return 0, errors.New("request has neither completions nor chat completions body")
```
roytman: The error message is not correct now, It should mention generate requests too, or be generalized to "unsupported request body type".

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -850,5 +853,24 @@ func (s *Scorer) getScores(ctx context.Context, _ *scheduling.CycleState, reques
 		return scores, len(blockKeys), nil
 	}
 
+	// For generate requests, token IDs are already available — score directly.
+	if request.Body != nil && request.Body.Generate != nil {
```
roytman: Should we add unit tests for generate requests for `computeBlockKeys` and `getScores`

roytman: It is slightly wider than this specific PR, but why not to add a guard at the start of the method, something like 
```go
if request.Body = nil {
    return nil, 0, errors.New("invalid request its Body is null, ")
```

and do not check `if request.Body != nil` for all use cases.

Additionally, Tokenized and Generate prompts processing are very similar; they can be combined. The only differences are the source of the IDs, and multimodal features.

pkg/epp/framework/interface/requesthandling/types.go
dmitripikus: Hi @vMaroon  and @roytman , regarding the comment about multi-modal features.
For now I just transfer `features` field as is, without touching it.
Should I go any processing related to it when processing `/inference/v1/generate` endpoint request?

ahg-g, line 445: why do we need this if we already have the `TokenizedPrompt` field?

pkg/epp/framework/interface/requesthandling/types.go
dmitripikus: Fixed. Thanks!

pkg/epp/framework/interface/requesthandling/types.go
dmitripikus: Fixed. Thanks!

pkg/epp/framework/interface/requesthandling/types.go
roytman: The `/inference/v1/generate` entry point is vLLM's answer to tokens-in tokens-out disaggregated scenarios. The Coordinator can send tokens and multimedia features as input. 
See https://github.com/llm-d-incubation/coordinator/blob/main/docs/communication.md#option-a-inferencev1generate

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
elevran, line 1: pretty sure this is not an openai compatible endpoint so belongs in a different package path.

roytman: yes, it is vLLM proprietary. Do you suggest creating the vllm package?

pkg/epp/framework/interface/requesthandling/types.go
ahg-g: Right, but continuing down this path of adding every single protocol to the internal structure is going to cause us problems as every plugin will need to be aware of every new protocol, this is not practical.

See https://github.com/llm-d/llm-d-router/issues/1303 

/cc @vMaroon

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
elevran: already have `vllmgrpc` under parsers, this looks like `vllmhttp` (unless both grpc and http have the same formats just differnet encodings, in which case `vllm` might be more appropriate).

ahg-g: @zetxqx for suggestions as well

roytman:
@dmitripikus , please see https://github.com/llm-d/llm-d-router/pull/1222/changes#r3296945098

pkg/epp/framework/interface/requesthandling/types.go
dmitripikus: @vMaroon and @roytman , following our discussion: as multi-modal features are needed for routing in EPP, the `features` are added to GenerateRequest. Thanks!

pkg/epp/framework/interface/requesthandling/types.go
dmitripikus: Done. Thanks!

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
dmitripikus: Fixed. Thanks!

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
dmitripikus: Done. Thanks!

dmitripikus: Refactoring is done. Thanks!

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
roytman: One of the issues is EPP cannot support several parsers, in order to resolve it and use vllm specific entry points in a separate package, i suggest that vllmhttp parser extends the OpenAI parser

```goLang
// pkg/epp/framework/plugins/requesthandling/parsers/vllmhttp/vllmhttp.go
package vllmhttp

const VllmHTTPParserType = "vllmhttp-parser"

type VllmHTTPParser struct {
    typedName fwkplugin.TypedName
    openai    *openai.OpenAIParser  // delegate for standard OpenAI endpoints
}
```

`ParseRequest `checks the path first. If it matches `/inference/v1/generate`, handle it locally. Otherwise, delegate to the embedded OpenAIParser.ParseRequest.
```goLang
func (p *VllmHTTPParser) ParseRequest(ctx context.Context, body []byte, headers map[string]string) (*fwkrh.ParseResult, error) {
    path := getRequestPath(headers)
    if strings.HasSuffix(path, "/inference/v1/generate") {
        return p.parseGenerateRequest(body)
    }
    // Delegate all OpenAI-compatible paths to the embedded parser.
    return p.openai.ParseRequest(ctx, body, headers)
}
```

pkg/epp/framework/plugins/scheduling/scorer/contextlengthaware/context_length_aware.go
```diff
@@ -220,7 +220,10 @@ func estimateContextLength(request *scheduling.InferenceRequest) int {
 	if request == nil || request.Body == nil {
 		return 0
 	}
-
+	// Generate requests carry pre-tokenized input — use exact count directly.
```
roytman: nit: please replace em-dash.

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -424,6 +437,78 @@ func (e *EmbeddingsRequest) String() string {
...
+	// CacheSalt is an optional request parameter to isolate prefix caches for security reasons.
+	CacheSalt string `json:"cache_salt,omitempty"`
+}
+
+func (r *GenerateRequest) String() string {
+	if r == nil {
+		return nilStr
+	}
+	mmItems := 0
+	mmModalities := 0
```
roytman: mm_hashes is a `map[string][]string`
I suggest printing keys (e.g. the multimodality type - "image) and the length of mm_hashes.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -150,6 +150,15 @@ func getKVCacheBlocksFromRawPrompt(ctx context.Context, request *scheduling.Infe
 		rawBytes := []byte(request.Body.Completions.Prompt.PlainText())
```
roytman, line 150: @vMaroon , is it correct? for tokes prompt the `PlainText() will return `""` because:
```golang
func (p Prompt) PlainText() string {
    if p.Raw != "" { return p.Raw }         // empty
    return strings.Join(p.Strings, " ")     // Strings is nil → ""
}
```

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
zetxqx: +1 on we can use `vllmhttp` for now to move this PR forward

But going forward, I put some thoughts on expanding the parsers, there are two approaches:
1. a unified http parser, we need a router on tope of the http parser and route to each individual request handler.
2. remove the restriction of only one parser and place multiple parsers.

Here is the doc: https://docs.google.com/document/d/1KfrtXWRT9mRFfiMCw1NSOmFkwwR4mX3kFNdAkHiyfPE/edit?tab=t.0#heading=h.ty4z5utg0wk5 for upcoming plan feel free to comment

pkg/epp/framework/plugins/scheduling/scorer/contextlengthaware/context_length_aware.go
dmitripikus: Done, Thanks!

pkg/epp/framework/interface/requesthandling/types.go
dmitripikus: Done. Thanks!

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
dmitripikus: Primary parser separation we agreed here, is implemented. Thank you!

pkg/epp/framework/interface/requesthandling/types.go
roytman: @ahg-g , I absolutely agree with you, but I prefer to pass this PR and refactor the protocols process separately, e.g., as #1303.

pkg/epp/framework/plugins/requesthandling/parsers/vllmhttp/vllmhttp.go
```diff
@@ -0,0 +1,125 @@
...
+// parseGenerateRequest decodes a /inference/v1/generate body into an
+// InferenceRequestBody. Token IDs are required; everything else is optional.
+func (p *VllmHTTPParser) parseGenerateRequest(rawBody []byte) (*fwkrh.ParseResult, error) {
+	bodyMap := make(map[string]any)
+	if err := json.Unmarshal(rawBody, &bodyMap); err != nil {
+		return nil, err
+	}
+
+	var generate fwkrh.GenerateRequest
+	if err := json.Unmarshal(rawBody, &generate); err != nil || len(generate.TokenIDs) == 0 {
```
elevran: The custom `GenerateRequest.UnmarshalJSON` returns informative errors (e.g., `token_ids[2]: invalid value -1` or NaN/MaxUint32 rejections), but folding the err and `len==0` cases into one branch flattens them all into `must have non-empty token_ids field`, so a request with negative or NaN token IDs gets the wrong error.
Please split the two error paths so each returns a relevant error.

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -773,41 +800,42 @@ func (s *Scorer) scoreBlockKeys(ctx context.Context, blockKeys []kvblock.BlockHa
...
-			if s.blockSizeTokens > 0 {
-				totalBlocks = len(tp.TokenIDs) / s.blockSizeTokens
+	// Prefer pre-tokenized input (TokenizedPrompt from the tokenizer plugin or
+	// a Body.Generate carrying token IDs directly).
+	if tokens, extraFeatures, ok := s.extractTokensAndFeatures(request.Body); ok {
+		traceLogger.Info("Scoring with pre-tokenized input", "tokenCount", len(tokens))
+
+		scores, err := s.kvCacheIndexer.ScoreTokens(ctx, tokens, request.TargetModel, nil, extraFeatures)
+		if err != nil {
+			if errors.Is(err, kvcache.ErrInternalTokenizationDisabled) {
```
elevran: nit: before this PR the `TokenizedPrompt` branch wrapped any `ScoreTokens` error and returned it; the new code adds a silent swallow that previously only existed on the chat/completions fallback. Tokens are pre-supplied here, so this should not fire, but if it does it now masks the misconfiguration with an empty score map for existing TokenizedPrompt callers as well as the new Generate path. 
Please consider if to drop the swallow on this combined path.

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -424,6 +438,87 @@ func (e *EmbeddingsRequest) String() string {
...
+			MMPlaceholders map[string][]wirePlaceholder `json:"mm_placeholders"`
+		} `json:"features,omitempty"`
+	}
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+	r.CacheSalt = raw.CacheSalt
+	r.TokenIDs = make([]uint32, len(raw.TokenIDs))
+	for i, v := range raw.TokenIDs {
+		if v < 0 || v > math.MaxUint32 || v != math.Trunc(v) {
```
elevran, line 500: The four rejection branches (negative, NaN, non-integer, > MaxUint32) have no direct test. The vllmhttp parser tests round-trip through `json.Marshal` so they cannot reach these paths; a small table-driven `UnmarshalJSON` test would cover this. Can be a follow up.

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache_tokenized_test.go
```diff
@@ -218,3 +223,157 @@ func TestScorer_SkipsTokenizedPromptWhenEmpty(t *testing.T) {
...
+						"image": {{Offset: 2, Length: 4}},
+					},
+				},
+			},
+		},
+	}
+
+	scorer.Score(ctx, scheduling.NewCycleState(), request, testEndpoints)
+
+	require.NotNil(t, capturedExtraFeatures, "extraFeatures should be passed when Generate.Features is present")
```
elevran, line 305: nit: the PR description frames the multimodal-features wiring as a bug fix, but the assertion here is only `NotNil`. A regression that drops `MMHashes` while still emitting a non-nil slice would still pass. Worth running `computeBlockKeys` twice with the same `TokenIDs` and different `MMHashes`, asserting the resulting block keys differ.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -217,6 +218,13 @@ func (p *Plugin) tokenize(ctx context.Context, request *scheduling.InferenceRequ
 		renderReq := ChatCompletionsToRenderChatRequest(request.Body.ChatCompletions)
 		traceLogger.Info("Calling RenderChat for chat completions", "messageCount", len(request.Body.ChatCompletions.Messages))
 		tokenIDs, mmFeatures, err = p.tokenizer.RenderChat(ctx, renderReq)
+	case request.Body.Generate != nil:
+		// Generate requests carry pre-tokenized input — use token IDs directly, no tokenizer call needed.
```
elevran: nit: the comment paraphrases the case condition and the early return below it. Can be dropped. 
Same pattern in `context_length_aware.go:223` and the second sentence of the `GenerateRequest` doc comment in `types.go`. Focus on the non-obvious.

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
```diff
@@ -236,6 +236,7 @@ func extractRequestBody(rawBody []byte, headers map[string]string) (*fwkrh.Infer
 			return &fwkrh.InferenceRequestBody{Embeddings: &embeddings}, nil
 		}
 		return nil, errors.New("invalid embeddings request: must have input field")
+
```
elevran: nit: stray blank line, unrelated to the change.

pkg/epp/framework/plugins/scheduling/scorer/contextlengthaware/context_length_aware.go
```diff
@@ -237,6 +240,7 @@ func estimateContextLength(request *scheduling.InferenceRequest) int {
 
 	// Convert characters to approximate token count
 	estimatedTokens := int(float64(totalChars) * charToTokenMultiplier)
+
```
elevran: nit: another stray blank line.

elevran (commented):
Thanks @dmitripikus - the `generate` plumbing through the parser, types, scorers, and tokenizer is consistent, and the new tests pin the multimodal flow on the precise prefix path. 

Some process related "must fix":
- fix the merge conflict.
- Is there a tracking issue that this PR addresses? If so, should be referenced/fixed here. We request that any non-trivial work references an issue.
- there are not release-notes specified. If this creates any user visible change (e.g., user can configure this parser and it is not an internal implementation details), it should be mentioned in a `release-notes` section in the PR description. Please follow the PR template in the future.


Additional feedback not directly covered by the diff so can't be linked to specific file/line:
- `ExtractMMItems` in `pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go` (around line 191) has no Generate branch. When the tokenizer DataProducer plugin is not configured (it is optional), `Generate.Features.MMHashes` is silently dropped on the encoder-cache scoring path. Same failure mode `generateExtraFeatures` fixes for the precise prefix scorer.
- `predictedlatency` uses `request.Body.PromptText()` and computes `inputTokenCount = len(strings.Fields(promptText))`. `PromptText` returns `""` for Generate, so SLO-aware scheduling sees zero input tokens for every Generate request. Switch to `InputTokenCountHint()` as the primary source (see `predictedlatency/plugin.go` around line 291; `inflightload/token_estimator.go` already uses this pattern).
- `pkg/epp/framework/plugins/requesthandling/parsers/README.md` does not enumerate `vllmhttp-parser` and is missing a config example. Same for the parser enumeration in `config/charts/routerlib/values.yaml` and `config/charts/README.md`.
- The composite-parser pattern (vllmhttp embedding openai and exporting `GetRequestPath` so the wrapper can sniff the path) is novel here. Worth a sentence in parsers/README on when to choose vllmhttp over openai and whether they may coexist on the same route.

bedeks:
`Generate.PromptText()` returns `""`, so this makes every `/inference/v1/generate` request look like zero input tokens to predicted latency. That breaks RL scheduling for long token-in rollouts. Should this follow the pattern [here](https://github.com/llm-d/llm-d-router/blob/c3aa9569cab51eb2dabf1f8f4a0beffa0ff55b42/pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/plugin.go#L276-L296)? : prefer `request.Body.InputTokenCountHint()` and only fall back to `len(strings.Fields(promptText))`.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -150,6 +150,15 @@ func getKVCacheBlocksFromRawPrompt(ctx context.Context, request *scheduling.Infe
...
 		return getKVCacheBlocksFromRawBytes(rawBytes, blockSizeTokens), nil
 
+	case request.Body.Generate != nil:
+		// Serialize token IDs as little-endian uint32 bytes for cache key generation.
+		ids := request.Body.Generate.TokenIDs
+		buf := make([]byte, len(ids)*4)
+		for i, id := range ids {
+			binary.LittleEndian.PutUint32(buf[i*4:], id)
+		}
```
ahg-g: This is strange, not something I would expect a plugin consuming the message to handle

roytman:
> `Generate.PromptText()` returns `""`, so this makes every `/inference/v1/generate` request look like zero input tokens to predicted latency. That breaks RL scheduling for long token-in rollouts. Should this follow the pattern [here](https://github.com/llm-d/llm-d-router/blob/c3aa9569cab51eb2dabf1f8f4a0beffa0ff55b42/pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/plugin.go#L276-L296)? : prefer `request.Body.InputTokenCountHint()` and only fall back to `len(strings.Fields(promptText))`.

Hi @bedeks, it has the same behaviour as `/v1/completions`, when its prompt is token_ids. Please see the comment https://github.com/llm-d/llm-d-router/pull/1248#discussion_r3299156922.

Looks like we need a generic method to calculate inputs, based on text or on tokens.

pkg/epp/framework/plugins/requesthandling/parsers/vllmhttp/vllmhttp.go
dmitripikus: Done. Thanks!

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
dmitripikus: Fixed. Thanks!

pkg/epp/framework/interface/requesthandling/types.go
dmitripikus: Tests are added. Thanks!

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
dmitripikus: Fixed. Thanks!

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache_tokenized_test.go
dmitripikus: Done. Thanks!

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
dmitripikus: Fixed. Thanks!

pkg/epp/framework/plugins/scheduling/scorer/contextlengthaware/context_length_aware.go
dmitripikus: Fixed. Thanks!

dmitripikus:
Hi Etai @elevran , I addressed all your comments above, including "must fix" items, additional feedback not directly covered by the diff, and comments related to specific files. Thank you!

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
dmitripikus: Hi @ahg-g , I fully agree with your comment that byte serialization didn't belong in a consumer. I moved the Generate path to a new getKVCacheBlocksFromTokens helper that yields HashBlock{Tokens: chunk} directly. Hashes are unchanged since HashBlock.Hash() already covers the typed field

elevran:
/lgtm
/approve

@dmitripikus in the future please use pull request template directly for the PRs as it includes certain formatting needed for release notes extraction (I manually added in the body above).

PR #1249, opened 2026-05-21
add MayConsumePlugin interface for optional data key consumption

### LINKED ISSUE #1224: Add data consumer `MayConsume` interface

Add a `MayConsume` interface to indicate the plugin can optionally consume certain data keys. If at init time there are no producers that produce these data keys, we should log a Warning.

Together with #1223 we can make the data producer/consumer dependency system robust and easy to understand.

**Why is this needed**:

Currently if a data consumer declares  data keys in `Consumes`, the framework will make sure there are producer(s) that produce the consumed data keys, or else attempt to create the default producer(s) that produce these keys. 

There are certain cases the consumption is optional, e.g., prefix cache plugin optionally consumes tokenized input, falls back to text input if tokenized input is not available. The current solution for this case is to not declare the consumed datakey, but silently read it in the code. This behavior is not obvious.

cc @rahulgurnani

sudoalok:
Closes #1224

## What problem does this solve?

Right now, if a plugin wants to use a data key only when it's available
(like using tokenized input if present, otherwise falling back), there's
no clean way to express that. The only option is `Consumes()`, which
errors out if no producer exists — even if the plugin could work fine
without it.

## What this PR does

Adds a new `MayConsumePlugin` interface with a `MayConsume()` method.
If a plugin declares a key here and no producer exists, the framework
logs a warning at startup instead of failing. The plugin handles the
missing data case itself at runtime.

## Files changed

- `plugins.go` — new `MayConsumePlugin` interface
- `data_graph.go` — warn instead of error for unproduced optional keys
- `data_graph_test.go` — 3 test cases for the new behavior

### DIFF
--- a/pkg/epp/datalayer/data_graph.go
+++ b/pkg/epp/datalayer/data_graph.go
@@ -81,10 +81,10 @@ func CreateMissingDataProducers(ctx context.Context, defaultProducerRegistry map
 		}
 	}
 
-	// Warn about optional keys (MayConsume) with no producer — no error, just a warning.
+	// Warn about optional keys with no producer — no error, just a warning.
 	for _, p := range handle.GetAllPlugins() {
-		if mayConsumer, ok := p.(plugin.MayConsumePlugin); ok {
-			for key := range mayConsumer.MayConsume() {
+		if consumer, ok := p.(plugin.ConsumerPlugin); ok {
+			for key := range consumer.OptionalConsumes() {
 				if !producedKeys[key.String()] {
 					logger.Info("Warning: optional data key has no producer, plugin will use fallback",
 						"plugin", p.TypedName().Name, "dataKey", key.String())
--- a/pkg/epp/datalayer/data_graph.go
+++ b/pkg/epp/datalayer/data_graph.go
@@ -81,6 +81,18 @@ func CreateMissingDataProducers(ctx context.Context, defaultProducerRegistry map
 		}
 	}
 
+	// Warn about optional keys (MayConsume) with no producer — no error, just a warning.
+	for _, p := range handle.GetAllPlugins() {
+		if mayConsumer, ok := p.(plugin.MayConsumePlugin); ok {
+			for key := range mayConsumer.MayConsume() {
+				if !producedKeys[key.String()] {
+					logger.Info("Warning: optional data key has no producer, plugin will use fallback",
+						"plugin", p.TypedName().Name, "dataKey", key.String())
+				}
+			}
+		}
+	}
+
 	// Build the set of keys that are consumed but not yet produced.
 	missingKeys := make(map[string]string)
 	for _, p := range handle.GetAllPlugins() {
--- a/pkg/epp/datalayer/data_graph.go
+++ b/pkg/epp/datalayer/data_graph.go
@@ -84,8 +84,8 @@ func CreateMissingDataProducers(ctx context.Context, defaultProducerRegistry map
 	// Warn about optional keys with no producer — no error, just a warning.
 	for _, p := range handle.GetAllPlugins() {
 		if consumer, ok := p.(plugin.ConsumerPlugin); ok {
-			result := consumer.Consumes()
-			for key := range result.Optional {
+			dependencies := consumer.Consumes()
+			for key := range dependencies.Optional {
 				if !producedKeys[key.String()] {
 					logger.Info("Warning: optional data key has no producer, plugin will use fallback",
 						"plugin", p.TypedName().Name, "dataKey", key.String())
@@ -98,8 +98,8 @@ func CreateMissingDataProducers(ctx context.Context, defaultProducerRegistry map
 	missingKeys := make(map[string]string)
 	for _, p := range handle.GetAllPlugins() {
 		if consumer, ok := p.(plugin.ConsumerPlugin); ok {
-			result := consumer.Consumes()
-			for key := range result.Required {
+			dependencies := consumer.Consumes()
+			for key := range dependencies.Required {
 				if !producedKeys[key.String()] {
 					missingKeys[key.String()] = consumer.TypedName().Name
 				}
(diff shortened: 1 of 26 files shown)


LukeAVanDrie:
Hey @sudoalok, thank you for putting this PR together! Supporting optional/soft data dependencies is a highly needed improvement for EPP.

I am working to transition EPP to a [declarative, graph-compiled execution runtime](https://docs.google.com/document/d/1hl7VAEMRX44mI5FjeBNh6DRhjJfM3XUinvEcbIlDjvs/edit?usp=sharing) (where startup execution pipeline slicing and ordering is dynamically resolved using declared data dependencies) and wanted to share some feedback on your proposed disjoint `MayConsume()` interface so we can align on the cleanest path forward.

I actually originally considered something similar to what you have implemented here; however, I believe there is benefit in having a single canonical location that answers: "What data does this plugin consume?"

With a disjoint `MayConsume()` interface:
- Data dependencies are split across two different methods (`Consumes()` and `MayConsume()`), increasing cognitive load.
- Every other component in EPP that inspects plugin consumption would have to be updated to check both methods, leading to API bloat and duplication.

To maintain a single source of truth, we can declare optional dependencies as a strict subset of `Consumes()` by introducing the `OptionalConsumerPlugin` interface:

```go
type OptionalConsumerPlugin interface {
    ConsumerPlugin
    // OptionalConsumes returns the subset of Consumes() keys that are optional.
    OptionalConsumes() []DataKey
}
```

Would you be able to look over this part of the linked proposal and see if that works for your use case?

nirrozenbaum:
```go
type OptionalConsumerPlugin interface {
    ConsumerPlugin
    // OptionalConsumes returns the subset of Consumes() keys that are optional.
    OptionalConsumes() []DataKey
}
```

@LukeAVanDrie this would lead to two types of ConsumerPlugins and IMO could be very confusing to users. 
sometimes a plugin may have several fields that are mandatory and few optional, and the UX doesn't seem to be clear here, should they use `ConsumerPlugin`? `OptionalConsumerPlugin`? (I know that the answer is the latter, but the UX is very confusing). 

I think a better approach would be to have a single ConsumerPlugin interface, and and extend it to have OptionalKeys. 
it will require updating existing interface + plugins (with OptionalKeys = []), but this is the cleanest approach I can think of.

sudoalok:
Thanks for the feedback! I agree having a single interface is cleaner. @nirrozenbaum  @LukeAVanDrie  can you both agree on the exact interface signature? I'll update the implementation once there's consensus

shmuelk:
Furthermore the proposed OptionalConsumerPlugin, extends ConsumerPlugin, meaning if any Plugin only optionally consumes data, it must have a Consume function that returns an empty map.

I agree extending the existing ConsumerPlugin Consume function to return a second map is better

LukeAVanDrie:
> I agree extending the existing ConsumerPlugin Consume function to return a second map is better

Adding to the existing plugin (exactly as Nir proposed) sounds good to me. I had suggested extending the `ConsumerPlugin` to avoid a breaking change to our plugin framework in case anyone was importing this outside the repo. That was a bigger concern when EPP code was split between gateway-api-inference-extension and llm-d-router, so this is probably fine.

sudoalok:
Updated the implementation . Instead of a separate MayConsumePlugin interface, I added OptionalConsumes() directly to ConsumerPlugin alongside Consumes(). If no producer exists for an optional key, the framework logs a warning instead of erroring. Existing plugins just return nil from OptionalConsumes().

ahg-g:
> Updated the implementation . Instead of a separate MayConsumePlugin interface, I added OptionalConsumes() directly to ConsumerPlugin alongside Consumes(). If no producer exists for an optional key, the framework logs a warning instead of erroring. Existing plugins just return nil from OptionalConsumes().

This has the same issue where plugins that doesn't have optional keys will have to implement and return a nil. Why not define a struct of two maps, required and optional, that `Consumes` returns?

sudoalok:
that makes sense, changing Consumes() to return a struct with required and optional maps is cleaner 
 no need for every plugin to implement an empty OptionalConsumes().

pkg/epp/framework/interface/plugin/plugins.go
```diff
@@ -23,13 +23,24 @@ type Plugin interface {
 	TypedName() TypedName
 }
 
+// ConsumesResult holds the data keys a plugin consumes, split by whether they
+// are required (framework errors if no producer exists) or optional (framework
+// logs a warning but continues if no producer exists).
+type ConsumesResult struct {
```
ahg-g: This name is an anti-pattern, I recommend something like `DataDependencies`

pkg/epp/datalayer/data_graph.go
```diff
@@ -81,11 +81,25 @@ func CreateMissingDataProducers(ctx context.Context, defaultProducerRegistry map
 		}
 	}
 
+	// Warn about optional keys with no producer — no error, just a warning.
+	for _, p := range handle.GetAllPlugins() {
+		if consumer, ok := p.(plugin.ConsumerPlugin); ok {
+			result := consumer.Consumes()
```
ahg-g:
```suggestion
			dependencies := consumer.Consumes()
```

pkg/epp/datalayer/data_graph.go
```diff
@@ -81,11 +81,25 @@ func CreateMissingDataProducers(ctx context.Context, defaultProducerRegistry map
...
+			}
+		}
+	}
+
 	// Build the set of keys that are consumed but not yet produced.
 	missingKeys := make(map[string]string)
 	for _, p := range handle.GetAllPlugins() {
 		if consumer, ok := p.(plugin.ConsumerPlugin); ok {
-			for key := range consumer.Consumes() {
+			result := consumer.Consumes()
```
ahg-g:
```suggestion
			dependencies := consumer.Consumes()
```

pkg/epp/datalayer/data_graph.go
```diff
@@ -191,9 +205,10 @@ func buildDAG(producers map[string]plugin.ProducerPlugin, consumers map[string]p
 			if pName == cName {
 				continue
 			}
-			if producer.Produces() != nil && consumer.Consumes() != nil {
+			consumed := consumer.Consumes()
```
ahg-g:
```suggestion
			dependencies := consumer.Consumes()
```

PR #1257, opened 2026-05-22
fix(epp/flowcontrol/registry): guard propagateStatsDelta against concurrent band deletion

### LINKED ISSUE #980: EPP panic — sync.Map.Load ok-ignored in propagateStatsDelta races with GC delete
**What happened**:

Two `propagateStatsDelta` callsites in the EPP flow-control registry pull a priority-band entry out of a `sync.Map` and immediately type-assert without checking the `ok` return:

- `pkg/epp/flowcontrol/registry/shard.go:371` — `val, _ := s.priorityBands.Load(priority); band := val.(*priorityBand)`
- `pkg/epp/flowcontrol/registry/registry.go:510` — `val, _ := fr.perPriorityBandStats.Load(priority); stats := val.(*bandStats)`

Both functions are lock-free in practice — neither holds `s.mu.Lock` or `s.mu.RLock`. The in-source doc comments make this explicit (`shard.go:367` says *"atomically updates the relevant band's stats"*; `registry.go:508` says *"top-level, lock-free aggregator for all statistics"*). That design is correct for the stats hot path, but the unchecked type assertion is incompatible with it: a concurrent `Delete()` from the GC path holds `s.mu.Lock()` (e.g., `deletePriorityBand` at `shard.go:178–183`) and can land between the lock-free `Load()` and the type assertion, leaving `val == nil` and panicking the EPP goroutine.

`propagateStatsDelta` is the hot-path stats callback — the function pointer is wired into every queue's `newManagedQueue` at `shard.go:327`, so every enqueue / dequeue invokes it without holding `s.mu`. The race fires when the GC's `deletePriorityBand` lands between an enqueue's `Load` and its assertion. Default deployments are largely safe in practice (GC timeouts default to 5 minutes); an operator who tunes `FlowGCTimeout` / `PriorityBandGCTimeout` down to chase memory savings can hit this. **Not a security vulnerability** — no external trigger and no privilege boundary crossed; this is reliability/operational.

The project's own pattern in `shard.go` is "comma-ok every `priorityBands.Load`" — 5 of the 7 callsites do, only 2 don't. The 2 outliers are line 311 (`synchronizeFlow`, currently safe because it holds `s.mu.Lock()`) and line 371 (`propagateStatsDelta`, lock-free by design — this is the bug). Same pattern in `registry.go` for `perPriorityBandStats.Load`.

A 35-line patch (one-line `if !ok { return }` guard before each of the two assertions) fixes both. I'll paste it as a follow-up comment. Patch applies cleanly against current `main`, compiles cleanly, and the existing unit test suite (`go test ./pkg/epp/flowcontrol/registry/...`) passes without modification.

The two `propagateStatsDelta` callsites should check the `ok` return from `sync.Map.Load` and short-circuit on a GC race instead of dereferencing a nil interface and panicking. The propagation is best-effort and lock-free by design; missing one delta when the band has just been collected is harmless and consistent with the function's "lock-free, eventually consistent" contract.

The reproduction is a Go test that lives **in the actual `pkg/epp/flowcontrol/registry` package** — it builds a real `registryShard` via the existing `newShardTestHarness`, races a real `propagateStatsDelta` call against a real `deletePriorityBand`, and observes the panic. No mocks, no cluster, no extra dependencies. Maintainers can run it with `go test`.

```bash
git clone https://github.com/llm-d/llm-d-inference-scheduler
cd llm-d-inference-scheduler
```

Save the following as `pkg/epp/flowcontrol/registry/ld1_race_test.go`:

```go
/*
LD-1 race reproducer. Demonstrates the panic in `(*registryShard).propagateStatsDelta`
(shard.go:371) when a concurrent `deletePriorityBand` removes the band entry between
the lock-free `s.priorityBands.Load(...)` and the unchecked `val.(*priorityBand)`
type assertion.

Background: `propagateStatsDelta` is the hot-path stats callback invoked from every
queue enqueue/dequeue (see `s.propagateStatsDelta` wired into `newManagedQueue` at
shard.go:327). It is intentionally lock-free for performance — the doc comment at
shard.go:367 says "atomically updates the relevant band's stats." `deletePriorityBand`
runs from the FlowRegistry GC path under `s.mu.Lock()` (shard.go:178). The race:
GC deletes the band; the next enqueue/dequeue in flight calls propagateStatsDelta;
the lock-free Load returns nil; the assertion panics; EPP exits.

Run:
    go test -count=1 -run TestLD1_PropagateStatsDeltaRacesGCDelete \
        ./pkg/epp/flowcontrol/registry/...
*/
package registry

import (
	"fmt"
	"sync/atomic"
	"testing"
	"time"
)

func TestLD1_PropagateStatsDeltaRacesGCDelete(t *testing.T) {
	h := newShardTestHarness(t)
	shard := h.shard
	priority := highPriority

	// Goroutine A: hammer the lock-free stats hot path.
	// Mirrors what every enqueue/dequeue does via the propagateStatsDelta callback
	// wired into newManagedQueue at shard.go:327.
	stopA := make(chan struct{})
	doneA := make(chan struct{})
	var panicMsg atomic.Value // string; non-empty if the bug fired
	go func() {
		defer close(doneA)
		defer func() {
			if r := recover(); r != nil {
				panicMsg.Store(fmt.Sprintf("%v", r))
			}
		}()
		for {
			select {
			case <-stopA:
				return
			default:
			}
			shard.propagateStatsDelta(priority, 1, 1)
		}
	}()

	// Give goroutine A a brief head-start so we know it's actively in the Load+assert
	// loop, then trigger the GC path. In production this is the FlowRegistry GC
	// (registry.go:426 → shard.go:177) reclaiming an idle priority band.
	time.Sleep(10 * time.Millisecond)
	shard.deletePriorityBand(priority)

	// After deletePriorityBand, the lock-free Load in propagateStatsDelta returns
	// (nil, false). Without the patch's ok-check, the next type assertion panics.
	// Goroutine A's defer-recover captures the panic and stores the message.
	select {
	case <-doneA:
		// A panicked and exited
	case <-time.After(2 * time.Second):
		close(stopA)
		<-doneA
	}

	if v := panicMsg.Load(); v != nil {
		t.Fatalf("LD-1 race fired: propagateStatsDelta panicked when band was "+
			"concurrently GC'd: %s\n\nThis is the unpatched behavior. The fix is to "+
			"comma-ok the Load and short-circuit on !ok. See "+
			"pkg/epp/flowcontrol/registry/shard.go:371 and registry.go:510.", v)
	}
}
```

Run against unpatched `main`:

```bash
go test -count=1 -v -run TestLD1_PropagateStatsDeltaRacesGCDelete ./pkg/epp/flowcontrol/registry/...
```

Expected (test fails in ~0.01 seconds with the actual panic from the real package code):

```
=== RUN   TestLD1_PropagateStatsDeltaRacesGCDelete
    ld1_race_test.go:83: LD-1 race fired: propagateStatsDelta panicked when band was concurrently GC'd: interface conversion: interface {} is nil, not *registry.priorityBand

        This is the unpatched behavior. The fix is to comma-ok the Load and short-circuit on !ok. See pkg/epp/flowcontrol/registry/shard.go:371 and registry.go:510.
--- FAIL: TestLD1_PropagateStatsDeltaRacesGCDelete (0.01s)
FAIL
FAIL    github.com/llm-d/llm-d-inference-scheduler/pkg/epp/flowcontrol/registry  0.774s
```

Note `interface {} is nil, not *registry.priorityBand` — the real type from the real package, hit on a real `registryShard` constructed by the package's existing `newShardTestHarness`.

Apply the proposed patch (paste below as a follow-up comment, save as `patch.diff`) and re-run:

```bash
git apply /path/to/patch.diff
go test -count=1 -v -run TestLD1_PropagateStatsDeltaRacesGCDelete ./pkg/epp/flowcontrol/registry/...
```

Expected (test passes; runs the full 2-second deadline because no panic fires):

```
=== RUN   TestLD1_PropagateStatsDeltaRacesGCDelete
--- PASS: TestLD1_PropagateStatsDeltaRacesGCDelete (2.01s)
PASS
ok      github.com/llm-d/llm-d-inference-scheduler/pkg/epp/flowcontrol/registry  2.824s
```

Run the full registry test suite under the patch to confirm nothing else regressed:

```bash
go test -count=1 ./pkg/epp/flowcontrol/registry/...
```

Expected:

```
ok      github.com/llm-d/llm-d-inference-scheduler/pkg/epp/flowcontrol/registry  2.716s
```

For a live-cluster repro, configure `FlowGCTimeout=1s` and `PriorityBandGCTimeout=2s` in the EPP config, run `hey -z 5m -c 50` against two distinct priority bands, and watch `kubectl logs <epp-pod> -c epp --previous --tail 80 | grep -A 8 '^panic:'` — a successful trigger shows a stack pointing at `pkg/epp/flowcontrol/registry/shard.go:372` (or `registry.go:511`). The kubelet `RESTARTS` counter lags 15-30s behind the actual crash, so the `--previous` log is the canonical signal. The race is timing-dependent in a real cluster (minutes-to-hours); the in-package test above reproduces it deterministically in milliseconds.

```
## Verified Load callsites in shard.go (7 total):
##
##   166: if _, ok := s.priorityBands.Load(priority); ok {        — ok-checked
##   215: val, ok := s.priorityBands.Load(key.Priority)            — ok-checked
##   231: val, ok := s.priorityBands.Load(priority)                — ok-checked
##   244: val, ok := s.priorityBands.Load(priority)                — ok-checked
##   311: val, _ := s.priorityBands.Load(key.Priority)             — ok-IGNORED (synchronizeFlow, holds s.mu.Lock — currently safe)
##   336: if val, ok := s.priorityBands.Load(key.Priority); ok {   — ok-checked
##   371: val, _ := s.priorityBands.Load(priority)                 — ok-IGNORED (propagateStatsDelta, LOCK-FREE — THIS IS THE BUG)

## Same pattern in registry.go for fr.perPriorityBandStats.Load:
##
##   510: val, _ := fr.perPriorityBandStats.Load(priority)         — ok-IGNORED (FlowRegistry.propagateStatsDelta, lock-free aggregator — paired bug)
```

Validation summary:
- Reproducer panics deterministically (~0.01s) against unpatched `main` (commit `3758ae0`)
- Reproducer passes against patched code (~2.01s — runs to deadline)
- Full `pkg/epp/flowcontrol/registry/...` test suite passes under the patch (2.716s)

The patch adds an `if !ok { return }` guard before each of the two type assertions — 4 added lines per callsite, 2 callsites, +10 / −2 in the diff. Existing tests pass without modification.

**Environment**:
- Kubernetes version (use `kubectl version`): N/A (race reproduces in-package via `go test`; no cluster needed)
- llm-d-scheduler version: `3758ae0` (main, 54 commits past `gaie-1.5-sync`)
- Cloud provider or hardware configuration: N/A (in-package test; race verified via `go test`)
- Install tools: N/A (`go test` runs locally with no other dependencies)
- Others: Go 1.22+

**Note:** Bug found using the [N184 Bug and Vulnerability Scanner](https://github.com/MillaFleurs/N184)

gkneighb:
## Summary

Fixes #980. Both `propagateStatsDelta` callsites in the flow-control registry pulled a priority-band entry out of a `sync.Map` lock-free and immediately type-asserted without checking the `ok` return:

- `pkg/epp/flowcontrol/registry/shard.go:375` — `val, _ := s.priorityBands.Load(priority); band := val.(*priorityBand)`
- `pkg/epp/flowcontrol/registry/registry.go:484` — `val, _ := fr.perPriorityBandStats.Load(priority); stats := val.(*bandStats)`

The GC path that removes bands holds the shard/registry write lock, so a concurrent `Delete()` can land between the lock-free `Load` and the type assertion, leaving `val == nil` and panicking the EPP goroutine processing the stats delta. Source comments at both callsites make the lock-free design explicit, so the right fix is to make the consumer tolerate the missing entry, not to add locking.

Switched both `Load` calls to comma-ok form with an early return when the band has already been collected. Dropping a single stats delta against a band that no longer exists is acceptable for an eventually-consistent aggregator; panicking the EPP goroutine is not.

## Verified locally

Both vectors panic on current `main` with exactly the messages from the issue:

```
TestShard_PropagateStatsDelta_UnknownPriority_NoPanic
  panic: interface conversion: interface {} is nil, not *registry.priorityBand
TestFlowRegistry_PropagateStatsDelta_UnknownPriority_NoPanic
  panic: interface conversion: interface {} is nil, not *registry.bandStats
```

After the fix:
- both new regression tests pass
- `go test -race ./pkg/epp/flowcontrol/registry/...` passes (including existing concurrency tests `TestShard_Concurrency_MixedWorkload` and `TestShard_Concurrency_AllOrderedPriorityLevels_RaceSafety`)
- `make test-unit-epp` and `make presubmit` are green

## Test plan
- [x] `go test ./pkg/epp/flowcontrol/registry/...`
- [x] `go test -race ./pkg/epp/flowcontrol/registry/...`
- [x] `make test-unit-epp`
- [ ] Reviewer to confirm "drop the delta on a deleted band" is the desired semantics (alternative: log at V(4) or higher)

### DIFF
--- a/pkg/epp/flowcontrol/registry/registry.go
+++ b/pkg/epp/flowcontrol/registry/registry.go
@@ -481,7 +481,13 @@ func (fr *FlowRegistry) buildFlowComponents(key flowcontrol.FlowKey) (*flowCompo
 
 // propagateStatsDelta is the top-level, lock-free aggregator for all statistics.
 func (fr *FlowRegistry) propagateStatsDelta(priority int, lenDelta, byteSizeDelta int64) {
-	val, _ := fr.perPriorityBandStats.Load(priority)
+	val, ok := fr.perPriorityBandStats.Load(priority)
+	if !ok {
+		// The band was concurrently removed by the GC path holding fr.mu.Lock(). The
+		// stats delta is lost, which is acceptable for an eventually-consistent counter
+		// of a band that no longer exists.
+		return
+	}
 	stats := val.(*bandStats)
 	stats.len.Add(lenDelta)
 	stats.byteSize.Add(byteSizeDelta)
--- a/pkg/epp/flowcontrol/registry/registry_test.go
+++ b/pkg/epp/flowcontrol/registry/registry_test.go
@@ -255,6 +255,27 @@ func TestFlowRegistry_Stats(t *testing.T) {
 	assert.Equal(t, globalStats.TotalByteSize, totalShardBytes, "The shard byte size must equal global byte size")
 }
 
+// TestFlowRegistry_PropagateStatsDelta_UnknownPriority_NoPanic is a
+// regression test for #980. The top-level propagateStatsDelta aggregator
+// is documented as lock-free; the GC's removePriorityBand holds the
+// registry's write lock and can Delete the perPriorityBandStats entry
+// between the lock-free Load and the type assertion. Before the fix, the
+// unchecked val.(*bandStats) panics with "interface conversion:
+// interface {} is nil, not *registry.bandStats" when val is nil. Same
+// reasoning as the shard-level test: drive nil through the type
+// assertion directly via an unknown priority rather than orchestrating
+// the race.
+func TestFlowRegistry_PropagateStatsDelta_UnknownPriority_NoPanic(t *testing.T) {
+	t.Parallel()
+	h := newRegistryTestHarness(t, harnessOptions{})
+	defer func() {
+		if r := recover(); r != nil {
+			t.Fatalf("propagateStatsDelta panicked on unknown priority: %v", r)
+		}
+	}()
+	h.fr.propagateStatsDelta(nonExistentPriority, 1, 100)
+}
+
 // --- Garbage Collection Tests ---
 
 func TestFlowRegistry_GarbageCollection(t *testing.T) {
--- a/pkg/epp/flowcontrol/registry/shard.go
+++ b/pkg/epp/flowcontrol/registry/shard.go
@@ -372,7 +372,13 @@ func (s *registryShard) updateConfig(newConfig *ShardConfig) {
 // It atomically updates the relevant band's stats, the shard's total stats, and propagates the delta to the parent
 // registry.
 func (s *registryShard) propagateStatsDelta(priority int, lenDelta, byteSizeDelta int64) {
-	val, _ := s.priorityBands.Load(priority)
+	val, ok := s.priorityBands.Load(priority)
+	if !ok {
+		// The band was concurrently removed by the GC path holding s.mu.Lock(). The
+		// stats delta is lost, which is acceptable for an eventually-consistent counter
+		// of a queue that no longer exists.
+		return
+	}
 	band := val.(*priorityBand)
 	band.len.Add(lenDelta)
 	band.byteSize.Add(byteSizeDelta)
--- a/pkg/epp/flowcontrol/registry/shard_test.go
+++ b/pkg/epp/flowcontrol/registry/shard_test.go
@@ -153,6 +153,27 @@ func TestShard_Stats(t *testing.T) {
 		"Priority band byte size must reflect the items queued at that level")
 }
 
+// TestShard_PropagateStatsDelta_UnknownPriority_NoPanic is a regression
+// test for #980. propagateStatsDelta is invoked lock-free on the stats hot
+// path; the GC's deletePriorityBand holds s.mu.Lock() and can delete a
+// band's sync.Map entry between the lock-free Load and the type assertion.
+// Before the fix, the unchecked val.(*priorityBand) panics with
+// "interface conversion: interface {} is nil, not *registry.priorityBand"
+// when val is nil. The shape of the bug is the same whether nil comes
+// from a never-populated priority or from a concurrent Delete, so this
+// test exercises the never-populated case directly without needing to
+// orchestrate the race.
+func TestShard_PropagateStatsDelta_UnknownPriority_NoPanic(t *testing.T) {
+	t.Parallel()
+	h := newShardTestHarness(t)
+	defer func() {
+		if r := recover(); r != nil {
+			t.Fatalf("propagateStatsDelta panicked on unknown priority: %v", r)
+		}
+	}()
+	h.shard.propagateStatsDelta(nonExistentPriority, 1, 100)
+}
+
 func TestShard_Accessors(t *testing.T) {
 	t.Parallel()
 


elevran (commented):
/lgtm
/approve
Thanks!

elevran:
@gkneighb can you please resolve conflicts?

gkneighb:
Closing — when rebasing onto current main I discovered the structural refactor in #1128 ("Third step in refactoring the FlowControl component") deleted `pkg/epp/flowcontrol/registry/shard.go` and rewrote `propagateStatsDelta` in `registry.go` with cascading `if ok` guards on both `priorityBands.Load` and `perPriorityBandStats.Load`. The nil-load panic from #980 is no longer reachable in main, so this PR is now redundant.

Thanks @elevran for the review and the merge-conflict ping that prompted me to look — and @shmuelk for fixing it as a side-effect of the refactor.

Issue #1216, opened 2026-05-19
Add helm charts builds to ci


ahg-g:
The two charts are  
- standalone-helm-chart-push
- inferencepool-helm-chart-push

See https://github.com/llm-d/llm-d-router/blob/main/Makefile#L325-L331

They need to be added to both https://github.com/llm-d/llm-d-router/blob/main/.github/workflows/ci-dev.yaml and https://github.com/llm-d/llm-d-router/blob/main/.github/workflows/ci-release.yaml

We should also change the charts names `llm-d-router-standalone` one to `llm-d-router-gateway`

notpad:
I’ll take this.

ahg-g:
Thanks @notpad , I sent a PR to change the names: https://github.com/llm-d/llm-d-router/pull/1254

PR #1262, opened 2026-05-22
Helm refactor

ahg-g:
**What type of PR is this?**
/kind cleanup

Refactored the library chart name and namespaces from  epplib  /  inferenceExtension  to  routerlib  /  router

Part of  https://github.com/llm-d/llm-d-router/issues/1216

```release-note
NONE
```


config/charts/epplib/templates/_latency-predictor.tpl
```diff
@@ -1,112 +0,0 @@
-{{/*
```
ahg-g, line 1: this was moved, not deleted

config/charts/routerlib/templates/_latency-predictor.tpl
```diff
@@ -0,0 +1,112 @@
+{{/*
```
ahg-g, line 1: this got moved, it is not newly created

PR #1266, opened 2026-05-22
[Helm] Adding a dedicated Tokenizer option to the helm charts

### LINKED ISSUE #1246: Add support for tokenizer sidecar in the helm charts
Currently only a single sidecar is supported in the epp deployment, when using standalone mode, this sidecar is used for the envoy container. We need to enable adding more than one sidecar so that the epp deployment can be configured with the tokenizer in standalone mode.

This will simplify the guide for https://github.com/llm-d/llm-d/tree/main/guides/precise-prefix-cache-routing

ahg-g:
**What type of PR is this?**
/kind feature

Adds a tokenizer sidecar option to the epp.

Also restructures the README.md into configuration subsections with complete examples.

Fixes https://github.com/llm-d/llm-d-router/issues/1246

```release-note
NONE
```

### DIFF
--- a/config/charts/README.md
+++ b/config/charts/README.md
@@ -204,6 +204,15 @@ The following table lists all configurable parameters for the LLM-D Router chart
 | `router.proxy.volumeMounts` | Sidecar container volume mounts. | `[]` |
 | `router.proxy.volumes` | Sidecar container volumes. | `[]` |
 | `router.proxy.configMapData` | Key-value pairs to include in a ConfigMap created for the sidecar. | `{}` |
+| **EPP Tokenizer Config (`router.tokenizer.*`)** | | |
+| `router.tokenizer.enabled` | Enable a tokenizer UDS sidecar container in the EPP deployment. | `false` |
+| `router.tokenizer.image.registry` | Tokenizer container image registry. | `ghcr.io/llm-d` |
+| `router.tokenizer.image.repository` | Tokenizer container image repository. | `llm-d-uds-tokenizer` |
+| `router.tokenizer.image.tag` | Tokenizer container image tag. | `vllm-v0.19.1` |
+| `router.tokenizer.image.pullPolicy` | Tokenizer container image pull policy. | `IfNotPresent` |
+| `router.tokenizer.env` | Extra environment variables for the tokenizer container (e.g., HuggingFace token). `TOKENIZERS_DIR` and `HF_HOME` are templated automatically. | `[HF_TOKEN]` |
+| `router.tokenizer.resources` | Tokenizer container resource requests and limits. | `requests.cpu: "8"`, `requests.memory: 8Gi` |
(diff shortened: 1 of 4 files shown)


config/charts/routerlib/templates/_deployment.yaml
```diff
@@ -258,10 +258,45 @@ spec:
         {{- if .Values.router.epp.volumeMounts }}
         {{- toYaml .Values.router.epp.volumeMounts | nindent 12 }}
         {{- end }}
+        {{- if .Values.router.tokenizer.enabled }}
+            - name: tokenizer-uds
```
vMaroon, line 262: Think we should just call it `tokenizer` as we have multiple options, maybe even make the whole section explicit in values and moved as-is.

The uds-tokenizer is deprecated, this new container looks like: https://github.com/llm-d/llm-d/pull/1507/changes#diff-2eee44c186c444e64670dc8f78d8337c04b316c951387016097adcb8eee736dc

Thank you for handling this.

ahg-g: lets merge this as is and you send a PR to polish it?

ahg-g: Mostly because my focus is on getting the helm charts structure in place, the docs around it and building and publishing the new charts

config/charts/README.md
```diff
@@ -63,40 +63,73 @@ helm install vllm-qwen3-32b ./config/charts/llm-d-router-gateway \
 ```
 ---
 
-## Common Customizations
+## Configuration & Customization
 
-Since both charts use `routerlib` under the hood, most EPP customizations are shared and configured under the `router` values block.
+Since both charts use `routerlib` under the hood, all configurations and customizations are shared under the `router` values block. EPP and the LLM-D Router can be customized by grouping configuration blocks in your `values.yaml` file. Below is the complete documentation and reference for each component.
```
vMaroon: s/LLM-D/llm-d/

config/charts/README.md
```diff
@@ -63,40 +63,73 @@ helm install vllm-qwen3-32b ./config/charts/llm-d-router-gateway \
...
+#### Complete Custom EPP Core Example
 
-### Custom Environment Variables
-Define custom environment variables for EPP in your `values.yaml`:
+To fully customize the EPP core container and pod (e.g., HA scaling, custom image, debugging flags, custom environment variables, custom plugin scheduling, and resource allocations), define the `router` block in your `values.yaml` as follows:
 
 ```yaml
 router:
   epp:
+    # Run EPP in active-passive HA mode
```
vMaroon: I think we should also have active-active mentioned somewhere, which can point to the llm-d precise guide or a config in here. Not for this PR, but what do you think?

ahg-g: Good point, I added a note at the top. We need to add an operational guide for the router that includes HA. We also should offer a canned configuration that is active-active compatible.

config/charts/README.md
ahg-g: updated here and other places.

PR #1267, opened 2026-05-22
Add EPP model server protocol to docs

BenjaminBraunDev:
**What type of PR is this?**

Add EPP metric model server protocol doc from original EPP repo. Moved from https://github.com/BenjaminBraunDev/gateway-api-inference-extension/blob/main/docs/proposals/003-model-server-protocol/README.md

Only change from the original doc it removal of its "proposal" status.

```release-note
NONE
```

### DIFF
--- a/docs/model-server-protocol.md
+++ b/docs/model-server-protocol.md
@@ -1,0 +1,61 @@
+# Model Server Protocol
+
+This document describes the contract the EPP expects from model servers
+it routes traffic to. Because the EPP uses a pluggable architecture, the
+requirements below describe what is needed to use the built-in plugins;
+individual plugins may relax or extend these requirements.
+
+## Inference API Protocol
+
+The model server MUST implement OpenAI’s [Completions](https://platform.openai.com/docs/api-reference/completions)
+and [Chat](https://platform.openai.com/docs/api-reference/chat) APIs.
+
+## Metrics Reporting
+
+The inference extension scrapes metrics from the model servers to make optimal request scheduling
+decisions. The model servers MUST provide the following metrics via a Prometheus endpoint. The exact
+metric names don't necessarily need to be the same as the recommended names here, however the
+metric types and semantics MUST follow this doc.
+
+Note the requirements here are aligned with the
+[model server metrics standardization](https://docs.google.com/document/d/1SpSp1E6moa4HSrJnS4x3NpLuj88sMXr2tbofKlzTZpk)
+effort.
+
+
+| Metric | Type | Description | vLLM metric | Triton TensorRT-LLM| trtllm-serve | SGLang |
+| ----- | ---- | ------------ | ---- | ---- | ---- | ---- |
+| TotalQueuedRequests         | Gauge     | The current total number of requests in the queue.| `vllm:num_requests_waiting`| `nv_trt_llm_request_metrics{request_type=waiting}`| `trtllm_num_requests_waiting` | `sglang:num_queue_reqs`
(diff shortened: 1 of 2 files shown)


BenjaminBraunDev:
This is just migrating the original doc here. Though as gRPC features are added I'm not sure if the statement from the doc remains true:

> The model server MUST implement OpenAI’s [Completions](https://platform.openai.com/docs/api-reference/completions) and [Chat](https://platform.openai.com/docs/api-reference/chat) APIs.

ahg-g:
pls sign the commit using the -s flag when `git commit`

docs/model-server-protocol.md
```diff
@@ -0,0 +1,61 @@
...
+
+This document describes the contract the EPP expects from model servers
+it routes traffic to. Because the EPP uses a pluggable architecture, the
+requirements below describe what is needed to use the built-in plugins;
+individual plugins may relax or extend these requirements.
+
+## Inference API Protocol
+
+The model server MUST implement OpenAI’s [Completions](https://platform.openai.com/docs/api-reference/completions)
+and [Chat](https://platform.openai.com/docs/api-reference/chat) APIs.
```
ahg-g: Not true anymore, we should remove the this.

ahg-g (commented):
We don't really have an overall model server protocol right now. It is more a per-plugin protocol now and so we should organize the doc around plugins.

docs/model-server-protocol.md
BenjaminBraunDev: done

BenjaminBraunDev:
renamed doc, added a required plugins column that described what plugins require each metric

docs/plugin-metric-protocol.md
```diff
@@ -0,0 +1,63 @@
...
+
+The **Required by** column lists the plugins that need each metric to function. You only need
+to expose the metrics for plugins you actually enable.
+
+Note the requirements here are aligned with the
+[model server metrics standardization](https://docs.google.com/document/d/1SpSp1E6moa4HSrJnS4x3NpLuj88sMXr2tbofKlzTZpk)
+effort.
+
+
+| Metric | Required by | Type | Description | vLLM metric | Triton TensorRT-LLM | trtllm-serve | SGLang |
```
ahg-g: a table is ideal, but hard to read because text breaks in some cells and also one has to scroll horizontally, please propose another format that is more readable:


Image

BenjaminBraunDev: Yeah I noticed this is an issue when dashes are used instead of underscores in strings (the metrics names don't have line breaks since they have underscores). I'll reorganize the table overall to make it less wide.

BenjaminBraunDev:
@ahg-g Just updated the formatting moving each metric into its own section for readability. Let me know what you think of the formatting. (the underline text in the GH markup preview is buggy, it doesn't render like that anywhere else)

docs/plugin-metric-protocol.md
```diff
@@ -0,0 +1,125 @@
+# Plugin Metric Protocol
+
+This document describes the contract the EPP expects from model servers
+it routes traffic to. Because the EPP uses a pluggable architecture, the
+requirements below describe what is needed to use the built-in plugins;
+individual plugins may relax or extend these requirements.
+
+## Metrics Reporting
+
+The inference extension scrapes metrics from the model servers to make optimal request scheduling
```
ahg-g, line 10:
```suggestion
Some plugins require the endpoint picker to scrape metrics from the model servers to make optimal request scheduling
```

BenjaminBraunDev:
@ahg-g can I get approval one more time? The markdown link checker didn't like a link to an internal google doc that was present perviously, so I removed that link (see latest commit)

liu-cong (approved):
Approve given @ahg-g 's approve

PR #1272, opened 2026-05-23
Fix(helm): router.epp.pluginsCustomConfig not rendered in ConfigMap

weizhoublue:
**What type of PR is this?**
 
/kind bug
  

Issue : 
User configures custom EPP plugins via router.epp.pluginsCustomConfig in Helm values (as documented in  values.yaml).   ConfigMap does not render user's custom plugin configuration. EPP falls back to default plugins or fails to   find specified config file.

Root Cause: 
PR #1264 migrated EPP-specific values from router.* to router.epp.* but missed updating the template file _config.yaml
   to read from the new path.

Fixes #

```release-note
NONE
```

### DIFF
--- a/config/charts/routerlib/templates/_config.yaml
+++ b/config/charts/routerlib/templates/_config.yaml
@@ -79,8 +79,8 @@ data:
     parser:
       pluginRef: {{ $parserType }}
     {{- end }}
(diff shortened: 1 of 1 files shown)


weizhoublue:
@ahg-g  hi, could you help check whether this pr makes sense , thanks

PR #1275, opened 2026-05-23
[Helm] inferenceObjective available for both modes and graduate the httproute option

### LINKED ISSUE #1284: [HELM] Graduate HTTPRoute from experimental
Graduate HTTPRoute from experimental 

**Why is this needed**:
Used in all guides now.

### LINKED ISSUE #1285: [HELM] Make InferenceObjectives option available in both standalone and gateway modes
It is currently available to only gateway mode, but the standalone with inferencepool.create=true supports inferenceObjectives.

/assign

ahg-g:
**What type of PR is this?**
/kind feature

- Creating InferenceObjectives is now available for both gateway and standalone mode when creating inferencePool is enabled.
- Graduated HTTPRoute option from experimental

Fixes https://github.com/llm-d/llm-d-router/issues/1285
Fixes https://github.com/llm-d/llm-d-router/issues/1284

```release-note
NONE
```

### DIFF
--- a/config/charts/README.md
+++ b/config/charts/README.md
@@ -206,30 +206,37 @@ router:
 
 ---
 
-### 3. InferencePool Configuration
+### 3. InferencePool & InferenceObjectives Configuration
 
-Settings for managing the `InferencePool` Kubernetes resource which logically groups the backend model servers.
+Settings for managing the `InferencePool` Kubernetes resource and optional `InferenceObjective` priority routing rules.
 
-#### InferencePool Parameters
+> [!IMPORTANT]
+> `inferenceObjectives` can only be configured when `inferencePool.create` is `true`. Creating `InferenceObjectives` without creating an `InferencePool` is not supported and will trigger a chart validation error.
+
+#### InferencePool & InferenceObjectives Parameters
 
 | **Parameter Name** | **Description** | **Default** |
 | :--- | :--- | :--- |
 | `router.inferencePool.create` | Whether to create the `InferencePool` resource. Set to `false` in standalone mode for Service-backed routing. | `true` |
-| `router.inferencePool.apiVersion` | The API version of the `InferencePool` resource. | `inference.networking.k8s.io/v1` |
-| `router.inferencePool.group` | The API group of the `InferencePool` resource. | `inference.networking.k8s.io` |
 | `router.inferencePool.failureMode` | EPP failure mode when external processing fails (configured on the pool). Options: `[FailOpen, FailClosed]`. | `FailOpen` |
+| `router.inferenceObjectives` | List of names and priorities to create optional `InferenceObjective` resources. | `[]` |
 
-#### Complete InferencePool Example
+#### Complete InferencePool & InferenceObjectives Example
 
 ```yaml
 router:
   inferencePool:
     # Enable or disable InferencePool resource creation (false in standalone service-backed mode)
     create: true
-    apiVersion: inference.networking.k8s.io/v1
-    group: inference.networking.k8s.io
     # If EPP fails to process the request, route anyway (FailOpen) or fail the request (FailClosed)
     failureMode: FailClosed
+  
+  # Optional: Define InferenceObjective(s) for this InferencePool
+  inferenceObjectives:
+    - name: high-priority
+      priority: 100
+    - name: background-batch
+      priority: 10
 ```
 
 ---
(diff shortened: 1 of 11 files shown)


config/charts/llm-d-router-gateway/templates/inferenceextension.yaml
```diff
@@ -1,4 +1,4 @@
-{{ include "llm-d-router.validations.epp.resources" $ }}
+{{ include "llm-d-router.validations.epp" $ }}
```
ahg-g, line 1: I will replace the below definitions to use llm-d-router instead of inference-extension prefix in a followup PR

config/charts/README.md
```diff
@@ -273,101 +280,7 @@ router:
...
-        ports:
-          - 8000 # Intercept traffic on port 8000
-```
-
----
-
-### 7. Sidecar Tokenizer Configuration (`router.tokenizer.*`)
+### 5. Sidecar Tokenizer Configuration (`router.tokenizer.*`)
 
 Runs a tokenizer sidecar container to expose a Unix Domain Socket (UDS) to EPP, allowing EPP to tokenize incoming requests and enable precise, token-count-aware routing policies (e.g., precise prefix-cache matching).
```
elevran, line 285: unrelated to this PR, but the UDS sidecar is being deprecated

ahg-g: Yes, I will leave it to @vMaroon to make the adjustments across the repo, I don't think the depreciation landed yet.

config/charts/llm-d-router-gateway/values.yaml
```diff
@@ -21,10 +21,9 @@ provider:
         #   http:
         #     maxRequestsPerConnection: 256000
 
-# experimentalHttpRoute section is used to deploy httproute as part of the epp helm chart.
-# this section is temporary and should be extracted to a separate chart.
-experimentalHttpRoute:
-  enabled: false # a flag to indicate whether to create the httproute as part of the chart or not.
+# httpRoute section is used to deploy an HTTPRoute resource as part of the gateway chart.
+httpRoute:
```
elevran, line 25: Breaking rename without a deprecation alias: `experimentalHttpRoute.enabled` to `httpRoute.create`. Existing user values.yaml files keep working syntactically but the HTTPRoute will not be created (`httpRoute.create` defaults to false). Worth either keeping a `experimentalHttpRoute` alias for one release with a `helm.sh/hook`-style warning, or flagging this in the release note.

config/charts/routerlib/values.yaml
```diff
@@ -118,9 +118,14 @@ tracing:
 
 inferencePool:
```
elevran, line 119: `inferencePool.apiVersion` and `inferencePool.group` are removed here without deprecation. I don't expect many to override to the alpha API (e.g. `inference.networking.x-k8s.io/v1alpha1`, which is still referenced in `_config.yaml:19`), but anyone who does loses that override silently and ends up with the v1 group hardcoded into the rendered manifests. Either a `fail` when the deprecated keys are still set, or a release-note callout, would catch this on upgrade.

config/charts/routerlib/values.yaml
```diff
@@ -118,9 +118,14 @@ tracing:
...
   create: true
-  apiVersion: inference.networking.k8s.io/v1
   failureMode: "FailOpen"
 
+inferenceObjectives: []
```
elevran, line 123: `inferenceObjectives` moved from top-level (gateway chart) to `router.inferenceObjectives` (routerlib). Existing users with the old top-level key will silently get zero objectives created. A short note in the README's migration section, or a Helm `fail` on the legacy path, would make that explicit.

config/charts/routerlib/templates/_inferenceobjective.yaml
```diff
@@ -10,6 +11,7 @@ metadata:
 spec:
   priority: {{ .priority }}
   poolRef:
-    group: {{ index (splitList "/" $.Values.router.inferencePool.apiVersion) 0 }}
+    group: "inference.networking.k8s.io"
```
elevran, line 14: nit: `"inference.networking.k8s.io"` is now hardcoded in three places: here, `_inferencepool.yaml:2` (as part of the `apiVersion`), and `_deployment.yaml:124` (the `--pool-group` flag). Worth a small helper in `_helpers.tpl` (e.g. `llm-d-router.inferencePool.group`) so a future API-group bump touches one location instead of three.

config/charts/llm-d-router-gateway/values.yaml
```diff
@@ -34,14 +33,4 @@ experimentalHttpRoute:
...
-# created in the chart release.
-# Users reference these objectives by name in their request headers.
-# This is only used for when InferenceObjectives are known ahead of time,
-# they can still be created out of band of the chart release as needed.
-inferenceObjectives: []
-#   - name: high-priority
-#     priority: 5
-#   - name: low-priority
-#     priority: 1
+
```
elevran: nit: minor cleanup  - trailing blank lines left behind after the removed `inferenceObjectives` block.

elevran (commented):
Moving `inferenceObjectives` into `routerlib` and graduating the HTTPRoute option both make good sense and the validation that `inferenceObjectives` requires `inferencePool.create=true` is a nice touch.

Main concern: this PR makes several silent-breaking renames to the chart values surface (see comments inline for specifics), but PR description is marked `release-note: NONE`. Helm doesn't warn on unknown keys, so users with existing `values.yaml` will get a quietly different (and reduced) deployment after upgrading.

ahg-g:
> Main concern: this PR makes several silent-breaking renames to the chart values surface (see comments inline for specifics), but PR description is marked release-note: NONE. Helm doesn't warn on unknown keys, so users with existing values.yaml will get a quietly different (and reduced) deployment after upgrading.

Right, but this is practically a new chart that we are releasing (different repo, different chart names). The charts went through a significant refactor/cleanup to capitalize on this one off opportunity.

ahg-g:
Having said that, I will add a migration section to the docs.

ahg-g:
I will also add fail to all settings that changed.

config/charts/routerlib/templates/_inferenceobjective.yaml
ahg-g: `inference.networking.k8s.io` is the staple graduated group name now, it should not change in the future.

config/charts/llm-d-router-gateway/values.yaml
ahg-g: Added in https://github.com/llm-d/llm-d-router/pull/1302

config/charts/routerlib/values.yaml
ahg-g: Added checks and docs in https://github.com/llm-d/llm-d-router/pull/1302

config/charts/routerlib/values.yaml
ahg-g: Added in https://github.com/llm-d/llm-d-router/pull/1302

config/charts/llm-d-router-gateway/values.yaml
ahg-g: done.

ahg-g:
@elevran thanks for the suggestion, in a follow up PR that is based on this one (https://github.com/llm-d/llm-d-router/pull/1302), I added checks, docs and release note for the deprecation.

PR #1276, opened 2026-05-23
fix(openai): parse prompt tokens details usage

### LINKED ISSUE #1273: EPP OpenAI parser ignores cached_tokens from prompt_tokens_details
## What happened

The EPP OpenAI parser currently extracts cached token usage from the wrong field name:

```json
usage.prompt_token_details.cached_tokens
```

However OpenAI-compatible responses, vLLM responses, and the router sidecar cached-token rewriter use:

```json
usage.prompt_tokens_details.cached_tokens
```

Note the extra `s` in `tokens`.

Because of this mismatch, EPP successfully parses `prompt_tokens`, `completion_tokens`, and `total_tokens`, but silently drops `cached_tokens` from standard OpenAI/vLLM responses.

## Expected behavior

EPP should parse cached token usage from the standard field:

```json
usage.prompt_tokens_details.cached_tokens
```

Responses API usage should likewise preserve cached input tokens from:

```json
usage.input_tokens_details.cached_tokens
```

## Why this matters

`cached_tokens` is used for cache-hit observability and usage/accounting signals. Dropping this field means EPP does not call `RecordPromptCachedTokens`, so prompt cache usage can be underreported or missing even when the model server returned it correctly.

This affects:

- prompt cache hit metrics
- usage accounting
- billing / chargeback integrations
- cache effectiveness observability

## Minimal reproduction

Given this OpenAI-compatible response:

```json
{
  "object": "chat.completion",
  "usage": {
    "prompt_tokens": 100,
    "completion_tokens": 50,
    "total_tokens": 150,
    "prompt_tokens_details": {
      "cached_tokens": 40
    }
  }
}
```

Expected parsed usage:

```go
PromptTokenDetails.CachedTokens == 40
```

Actual parsed usage:

```go
PromptTokenDetails == nil
```

## Code pointers

The parser currently checks `prompt_token_details` in `pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go`.

The sidecar rewriter already uses the standard `prompt_tokens_details` field, so EPP and sidecar currently disagree on the field name.

zhouyou9505:
## Summary

Fixes cached token usage extraction in the EPP OpenAI parser.
This caused EPP to silently drop cached token usage from normal responses, so prompt cached-token metrics were not recorded.

## Changes

- Parse `cached_tokens` from `usage.prompt_tokens_details`.
- Parse Responses API cached input tokens from `usage.input_tokens_details`.
- Update the internal `Usage` JSON tag to emit `prompt_tokens_details`.
- Update parser and handler tests to cover the standard plural field.

## Why

This is a usage/accounting correctness fix. Without it, cache-hit observability can be misleading because EPP records prompt/output tokens but misses cached prompt tokens.

Fixes #1273

### DIFF
--- a/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
@@ -264,7 +264,46 @@ func extractUsage(responseBytes []byte) (*fwkrh.Usage, error) {
 	if responseErr != nil {
 		return nil, responseErr
 	}
+usg, ok := responseBody["usage"].(map[string]any)
+if !ok {
+  return nil, nil
+}
 
+usage := fwkrh.Usage{}
+
+// Chat/Completions APIs use prompt_tokens. Responses/Conversations APIs use input_tokens.
+for _, it := range []string{"prompt_tokens", "input_tokens"} {
+  if v, ok := usg[it]; ok && v != nil {
+    usage.PromptTokens = toInt(v)
+    break
+  } 
+}
+
+// Chat/Completions APIs use completion_tokens. Responses/Conversations APIs use output_tokens.
+for _, ot := range []string{"completion_tokens", "output_tokens"} {
+  if v, ok := usg[ot]; ok && v != nil {
+    usage.CompletionTokens = toInt(v)
+    break
+  } 
+}
+
+// Chat/Completions APIs use prompt_tokens_details. Responses/Conversations APIs use input_tokens_details.
+for _, details := range []string{"prompt_tokens_details", "input_tokens_details"} {
+  if detailsMap, ok := usg[details].(map[string]any); ok {
+    if cachedTokens, ok := detailsMap["cached_tokens"]; ok {
+      usage.PromptTokenDetails = &fwkrh.PromptTokenDetails{
+        CachedTokens: toInt(cachedTokens),
+      }
+    }
+  }
+}
+
+// total_tokens field name is consistent across all API types
+if v, ok := usg["total_tokens"]; ok && v != nil {
+  usage.TotalTokens = toInt(v)
+}
+
+return &usage, nil
 	if responseBody["usage"] != nil {
 		usg, ok := responseBody["usage"].(map[string]any)
 		if !ok {
--- a/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
@@ -47,6 +47,16 @@ const (
 	// The base media type for Server-Sent Events. We check for this substring
 	// to account for optional parameters like "; charset=utf-8" often appended by proxies.
 	eventStreamType = "text/event-stream"
+
+	usageField               = "usage"
+	promptTokensField        = "prompt_tokens"
+	inputTokensField         = "input_tokens"
+	completionTokensField    = "completion_tokens"
+	outputTokensField        = "output_tokens"
+	promptTokensDetailsField = "prompt_tokens_details"
+	inputTokensDetailsField  = "input_tokens_details"
+	cachedTokensField        = "cached_tokens"
+	totalTokensField         = "total_tokens"
 )
 
 // compile-time type validation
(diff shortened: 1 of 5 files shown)


pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
```diff
@@ -380,19 +383,27 @@ func extractUsageStreaming(responseText string) *fwkrh.Usage {
...
+		if streamResponse["usage"] != nil {
+			if usage, err := extractUsage(byteSlice); err == nil && usage != nil {
+				return usage
+			}
 		}
+
 		// Responses API streaming format
-		if streamResponse.Response.Usage != nil && streamResponse.Type == "response.completed" {
+		response, _ := streamResponse["response"].(map[string]any)
+		usageMap, _ := response["usage"].(map[string]any)
```
ahg-g: should we check that response is not nil and the cast above actually succeeded?

zhouyou9505: Agreed. I’ll keep this consistent with the main branch and simplify the code to make it more readable.

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
```diff
@@ -280,8 +280,16 @@ func extractUsage(responseBytes []byte) (*fwkrh.Usage, error) {
 		}
 		objectType, _ := responseBody["object"].(string)
 		usage := extractUsageByAPIType(usg, objectType)
-		if usg["prompt_token_details"] != nil {
-			if detailsMap, ok := usg["prompt_token_details"].(map[string]any); ok {
+		if usg["prompt_tokens_details"] != nil {
```
ahg-g: can you pls document the difference between `prompt_tokens_details` and `input_tokens_details`? why are we looking up both basically

zhouyou9505: I put boh the code `prompt_token_details`,`input_tokens_details`  into this `extractUsageByAPIType` function to make it more intuitive

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
ahg-g: How about we do the following and remove everything below including `extractUsageByAPIType`:

```suggestion
usg, ok := responseBody["usage"].(map[string]any)
if !ok {
  return nil, nil
}

usage := fwkrh.Usage{}

// Chat/Completions APIs use prompt_tokens. Responses/Conversations APIs use input_tokens.
for _, it := range []string{"prompt_tokens", "input_tokens"} {
  if v, ok := usg[it]; ok && v != nil {
    usage.PromptTokens = toInt(v)
    break
  } 
}

// Chat/Completions APIs use completion_tokens. Responses/Conversations APIs use output_tokens.
for _, ot := range []string{"completion_tokens", "output_tokens"} {
  if v, ok := usg[ot]; ok && v != nil {
    usage.CompletionTokens = toInt(v)
    break
  } 
}

// Chat/Completions APIs use prompt_tokens_details. Responses/Conversations APIs use input_tokens_details.
for _, details := range []string{"prompt_tokens_details", "input_tokens_details"} {
  if detailsMap, ok := usg[details].(map[string]any); ok {
    if cachedTokens, ok := detailsMap["cached_tokens"]; ok {
      usage.PromptTokenDetails = &fwkrh.PromptTokenDetails{
        CachedTokens: toInt(cachedTokens),
      }
    }
  }
}

// total_tokens field name is consistent across all API types
if v, ok := usg["total_tokens"]; ok && v != nil {
  usage.TotalTokens = toInt(v)
}

return &usage, nil
```

zhouyou9505: Thanks for the suggestion. I applied this approach and removed `extractUsageByAPIType`. that's would be more clearly

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
```diff
@@ -271,73 +264,46 @@ func extractUsage(responseBytes []byte) (*fwkrh.Usage, error) {
...
-			usage.PromptTokens = toInt(v)
-		}
-		if v, ok := usg["output_tokens"]; ok && v != nil {
-			usage.CompletionTokens = toInt(v)
-		}
-	case objectType == objectTypeChatCompletion || objectType == objectTypeChatCompletionChunk || objectType == objectTypeTextCompletion:
-		// Traditional APIs use prompt_tokens/completion_tokens
-		if v, ok := usg["prompt_tokens"]; ok && v != nil {
+	// Chat/Completions APIs use prompt_tokens. Responses/Conversations APIs use input_tokens.
+	for _, it := range []string{"prompt_tokens", "input_tokens"} {
```
shmuelk: Please use constants for the strings

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
```diff
@@ -271,73 +264,46 @@ func extractUsage(responseBytes []byte) (*fwkrh.Usage, error) {
...
+		if v, ok := usg[it]; ok && v != nil {
 			usage.PromptTokens = toInt(v)
+			break
 		}
-		if v, ok := usg["completion_tokens"]; ok && v != nil {
+	}
+
+	// Chat/Completions APIs use completion_tokens. Responses/Conversations APIs use output_tokens.
+	for _, outputTokens := range []string{"completion_tokens", "output_tokens"} {
```
shmuelk: Please use constants for the strings

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
```diff
@@ -271,73 +264,46 @@ func extractUsage(responseBytes []byte) (*fwkrh.Usage, error) {
...
+	}
 
-		if v, ok := usg["output_tokens"]; ok && v != nil {
-			usage.CompletionTokens = toInt(v)
-		} else if v, ok := usg["completion_tokens"]; ok && v != nil {
-			usage.CompletionTokens = toInt(v)
+	// Chat/Completions APIs use prompt_tokens_details. Responses/Conversations APIs use input_tokens_details.
+	for _, details := range []string{"prompt_tokens_details", "input_tokens_details"} {
+		if detailsMap, ok := usg[details].(map[string]any); ok {
+			if cachedTokens, ok := detailsMap["cached_tokens"]; ok {
```
shmuelk: Please use constants for the strings

zhouyou9505: OK, thanks for reminder.

PR #1281, opened 2026-05-23
fix: allow unstale workflow to remove labels from PRs

### LINKED ISSUE #676: Prow action fails to remove lifecycle label
For example, following a comment on #550 we get [this run](https://github.com/llm-d/llm-d-inference-scheduler/actions/runs/22616054969/job/65529103371) error:
```console
Run echo "Removing 'stale' label from issue #550"
  echo "Removing 'stale' label from issue #550"
  gh issue edit 550 --remove-label "lifecycle/stale,lifecycle/rotten"
  shell: /usr/bin/bash -e {0}
  env:
    GITHUB_TOKEN: ***
Removing 'stale' label from issue #550
failed to update https://github.com/llm-d/llm-d-inference-scheduler/pull/550: GraphQL: Resource not accessible by integration (removeLabelsFromLabelable)
failed to update 1 issue
Error: Process completed with exit code 1.
```

varad-ahirwadkar:
**What type of PR is this?**

Updates the unstale workflow - [unstale.yaml](https://github.com/llm-d/llm-d-router/blob/main/.github/workflows/unstale.yaml) so it can remove stale labels from PRs and handles `lifecycle/stale` and `lifecycle/rotten` more safely by removing them individually.

Fixes #676 

```release-note
NONE
```

### DIFF
--- a/.github/workflows/unstale.yaml
+++ b/.github/workflows/unstale.yaml
@@ -1,20 +1,32 @@
-name: 'Unstale Issue'
+name: 'Unstale Issue or PR'
 
 on:
   issues:
     types: [ reopened ]
   issue_comment:
     types: [ created ]
+  pull_request:
+    types: [ reopened ]
 
 jobs:
   remove-stale:
     runs-on: ubuntu-latest
     permissions:
       issues: write
+      pull-requests: write
     if: >-
-      github.event.issue.state == 'open' &&
-      (contains(github.event.issue.labels.*.name, 'lifecycle/stale') || 
-       contains(github.event.issue.labels.*.name, 'lifecycle/rotten'))
+      (
+        github.event_name == 'pull_request' &&
+        github.event.pull_request.state == 'open' &&
+        (contains(github.event.pull_request.labels.*.name, 'lifecycle/stale') ||
+         contains(github.event.pull_request.labels.*.name, 'lifecycle/rotten'))
+      ) ||
+      (
+        github.event_name != 'pull_request' &&
+        github.event.issue.state == 'open' &&
+        (contains(github.event.issue.labels.*.name, 'lifecycle/stale') ||
+         contains(github.event.issue.labels.*.name, 'lifecycle/rotten'))
+      )
     steps:
       - name: 'Checkout repository'
         uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


.github/workflows/unstale.yaml
```diff
@@ -11,6 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       issues: write
+      pull-requests: write
```
nirrozenbaum, line 16: why do we need PR write permission if we update issues only?

varad-ahirwadkar: We need `pull-requests: write` because the reported failure in issue #676 was for a PR (#550).

varad-ahirwadkar: Since `issue_comment` is triggered for both issues and PRs, the workflow should also have PR write permission to remove labels from PRs.

nirrozenbaum: I’m not 100% sure, but I think the issue reopened would miss PR reopen and that needs to be added explicitly. 
can you verify?

varad-ahirwadkar: Sure, I'll verify. I think you're right though, reopened PR handling is currently missing and would need to be added explicitly with `pull_request: [reopened]`

varad-ahirwadkar: I have updated the workflow to handle reopened PRs:
- Added `pull_request: [ reopened ]` so the unstale workflow also handles reopened PRs
- Updated the job condition to use the appropriate payload for each event type, `github.event.pull_request` for PR events and `github.event.issue` for issues/comments
- Made label removal work correctly for both issues and PRs by resolving the target with `github.event.issue.number || github.event.pull_request.number`

Verification -
Test PR - https://github.com/varad-ahirwadkar/llm-d-inference-scheduler/pull/7 
Test issue - https://github.com/varad-ahirwadkar/llm-d-inference-scheduler/issues/8

elevran (commented):
thanks @varad-ahirwadkar !
One small **optional** suggestion: the workflow `name:` in line 1 is `'Unstale Issue'` and no longer fully reflects scope now that it also handles PRs. Consider `'Unstale Issue or PR'`.
/lgtm
/approve
/hold if you want to change the action's `name` as indicated above

varad-ahirwadkar:
Hi @elevran ,
Thanks, I agree. I have updated the action's `name`.

PR #1283, opened 2026-05-24
feat(epp/preciseprefixcache): accept blockSizeTokens as top-level config alias

### LINKED ISSUE #1157: precise_prefix_cache fails to read the updated configuration: BlockSizeTokens

gkneighb:
## Summary

Fixes #1157. The precise-prefix-cache scorer wraps the external `kvblock.TokenProcessorConfig` whose block-size field is `blockSize`, while the sibling `approx-prefix-cache-producer` exposes the same concept as `blockSizeTokens`. Operators copying a working config from one plugin to the other (or following newer guides that document `blockSizeTokens`) would set `blockSizeTokens` at the top level of `precise-prefix-cache` and the value would be silently dropped — the scorer would fall through to the default 16-token block size.

This adds `BlockSizeTokens` as a top-level field on `PluginConfig`. When set, it overrides `TokenProcessorConfig.BlockSize`. The legacy `tokenProcessorConfig.blockSize` path remains accepted for backwards compatibility with configs that wrap the underlying `kvblock` config directly. Resolution happens in a new `PluginConfig.applyAliases` method invoked from `PluginFactory` right after `json.Unmarshal`.

This is also the right shape going forward once #1134 (strict plugin-config parsing) lands: today the orphan `blockSizeTokens` field is silently dropped; after #1134 it would be an outright error; after this PR it just works.

## Verified locally

Regression tests cover all four resolution cases:

- top-level `blockSizeTokens` sets `BlockSize`
- legacy `tokenProcessorConfig.blockSize` still works
- both set → top-level wins
- neither set → resolution untouched (handled by `New()`'s default)

`go test ./pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/...`, `make test-unit-epp`, `make test-integration-hermetic PATTERN='TestWellKnownConfigs'`, and `make presubmit` all green.

## Test plan
- [x] `go test ./pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/...`
- [x] `make test-unit-epp`
- [x] `make test-integration-hermetic PATTERN='TestWellKnownConfigs'`
- [ ] Reviewer to confirm "top-level wins over wrapped" precedence (alternative: log a warning when both are set)

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
+++ b/pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
@@ -60,6 +60,16 @@ type kvCacheIndexer interface {
 // PluginConfig holds the configuration for the
 // Scorer plugin.
 type PluginConfig struct {
+	// BlockSizeTokens is the number of tokens per KV-block. When set, it
+	// overrides TokenProcessorConfig.BlockSize. The field is a top-level
+	// alias so operators can use a single field name across both the
+	// precise- and the approximate-prefix-cache plugins (the latter has
+	// always exposed blockSizeTokens directly). The legacy
+	// tokenProcessorConfig.blockSize remains accepted for backwards
+	// compatibility with configs that wrap the underlying kvblock
+	// config directly.
+	BlockSizeTokens int `json:"blockSizeTokens,omitempty"`
+
 	// TokenProcessorConfig holds the configuration for the `kvblock.TokenProcessor` which is
 	// used to process tokens into KV-block keys.
 	TokenProcessorConfig *kvblock.TokenProcessorConfig `json:"tokenProcessorConfig"`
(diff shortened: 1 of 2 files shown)


vMaroon:
Hi @gkneighb - although there currently is a bug highlighted in the PR you referenced above, the right location for this field is in the kv-cache libraries and not a higher level config. See https://github.com/llm-d/llm-d-kv-cache/pull/564.

gkneighb:
Thanks @vMaroon, that's a fair point — the kv-cache layer is the right home and llm-d/llm-d-kv-cache#564 already uses the same deprecation pattern. The only reason I went higher up was because llm-d-router currently pins `llm-d-kv-cache v0.8.0` (tagged 2026-05-01, before #564 merged on 2026-05-07), and the rename is only in `v0.8.0-fs-v0.20` so far — but that's a release-cadence question rather than a layer question.

Closing this in favor of waiting for a stable kv-cache tag that includes #564, then a straightforward dependency bump here. I'll keep an eye on #1157 in case anyone hits the top-level-vs-wrapped placement case before the dep bump lands.

PR #1288, opened 2026-05-24
feat(epp): add conditional-decode support (Prefer: if-available) - Depends on PR #1248

### LINKED ISSUE #1366: [Feature] Extend EPP to support conditional-decode
Implement the conditional-decode as defined at https://github.com/llm-d-incubation/coordinator/blob/main/docs/communication.md#stage-3-conditional-decode, with `Prefer: if-available` HTTP header and return error code `412 Precondition Failed`.

In the meantime, it will not provide additional hint information.

**Why is this needed**:
Introduction of per-pool EPP and a coordination service to llm-d would not allow conditional decode/prefill, as supported today.

see #1116

dmitripikus:
## Summary

Adds EPP support for the coordinator's conditional-decode protocol 
(see `llm-d-incubation/coordinator/docs/communication.md`). <br>
When the request carries `Prefer: if-available`, EPP attempts a speculative early decode. <br>
The decode worker is contacted only when its KV cache already covers the prompt; otherwise EPP returns HTTP 412 so the coordinator restarts the pipeline at encode/prefill/decode.

## Changes

- **`pkg/common/routing/common.go`** — added `PreferHeader` (`prefer`) and
  `PreferIfAvailable` (`if-available`) constants. EPP receives header keys
  lowercased; the value is parsed token-aware per RFC 7240 (comma-separated,
  `;param=value` ignored, case-insensitive).
- **`pkg/common/error/error.go`** — added `PreconditionFailed` error code,
  mapped to `envoyTypePb.StatusCode_PreconditionFailed` (HTTP 412).
- **`pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go`**:
  - `isConditionalDecode(request)` — Prefer-header parse.
  - `hasCachedPrefix(endpoint)` — checks `PrefixCacheMatchInfo` on the
    chosen endpoint.
  - `Pick()` short-circuits encode + prefill stages when conditional-decode
    is signaled (decode profile still runs to score and pick the best pod).
  - `ProcessResults()` returns 412 when the chosen decode pod's prefix
    cache match is empty.
- **`pkg/epp/requestcontrol/director.go`**:
  - Preserves typed `errcommon.Error` from the scheduler instead of always
    wrapping as `ResourceExhausted` — so the 412 status reaches Envoy
    intact.
 
Fixes [#1366](https://github.com/llm-d/llm-d-router/issues/1366)

### E2E test:
E2E test is implemented in separate PR [#1356](https://github.com/llm-d/llm-d-router/pull/1356)

```release-note
- When a request carries the standard HTTP `Prefer: if-available` header (RFC 7240), the EPP routes to a decode worker only if its KV cache already covers the prompt; otherwise it returns HTTP 412 Precondition Failed so the coordinator restarts the pipeline at encode/prefill/decode. 
- The cache check reads `PrefixCacheMatchInfo` from the chosen endpoint using the default-named approximate-prefix producer's key. Deployments using the auto-created `approx-prefix-cache-producer` (the canonical decode-EPP recipe) get the optimization. Deployments using a custom-named approx producer or `precise-prefix-cache-producer` write under a different key, so the gate misses and the coordinator receives 412 on every conditional-decode request — falling back to the full pipeline (correctness preserved, optimization effectively disabled) for those configurations.
```


pkg/epp/requestcontrol/director.go
```diff
@@ -56,6 +58,35 @@ const (
...
+	}
+	primary, ok := result.ProfileResults[result.PrimaryProfileName]
+	if !ok || primary == nil || len(primary.TargetEndpoints) == 0 {
+		return false
+	}
+	endpoint := primary.TargetEndpoints[0]
+	if endpoint == nil {
+		return false
+	}
+	raw, ok := endpoint.Get(attrprefix.PrefixCacheMatchInfoDataKey.String())
```
roytman, line 92: **Suggestion**: Simplify PrefixCacheMatchInfoDataKey lookup in the director
The director in primaryEndpointHasCachedPrefix reads the attribute using:
`raw, ok := endpoint.Get(attrprefix.PrefixCacheMatchInfoDataKey.String())`

While the approximate-prefix plugin writes using:
`pod.Put(p.dk.String(), ...)`
// where p.dk = attrprefix.PrefixCacheMatchInfoDataKey.WithNonEmptyProducerName(name)

These produce the same string today because the default producer name in NewDataKey happens to equal `ApproxPrefixCachePluginType`. However, the two call sites look different, which forces a reader to trace through the DataKey implementation to confirm they match. It also means that if a second plugin instance is registered with a custom name, the director would silently miss its data.

**Suggested fix**: Export a pre-computed key string from the prefix package:


// In pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
`var PrefixCacheMatchInfoKey = PrefixCacheMatchInfoDataKey.String()`

Then both the plugin and director use `attrprefix.PrefixCacheMatchInfoKey`, making equivalence obvious at a glance.

dmitripikus: Done. Thanks!

elevran:
@dmitripikus please resolve conflicts

dmitripikus:
Hi @elevran , I resolved the conflicts in PR

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
```diff
@@ -61,7 +60,7 @@ func (p *dataProducer) TypedName() plugin.TypedName {
 
 // Produces returns the data produced by the plugin.
 func (p *dataProducer) Produces() map[plugin.DataKey]any {
-	return map[plugin.DataKey]any{p.dk: attrprefix.PrefixCacheMatchInfo{}}
+	return map[plugin.DataKey]any{attrprefix.PrefixCacheMatchInfoDataKey: attrprefix.PrefixCacheMatchInfo{}}
```
elevran: This breaks the `tiered-prefix-cache-cpu well-known` config and is the cause of the failing CI run.

That config registers a named instance `cpu-prefix-cache-producer` (type approx-prefix-cache-producer) and a `cpu-prefix-cache-scorer` that consumes `PrefixCacheMatchInfoDataKey/cpu-prefix-cache-producer` via `prefixMatchInfoProducerName`. Before this PR, `Produces()` returned the per-instance key (via `WithNonEmptyProducerName(name)`) so framework auto-resolution matched. Now `Produces()` always returns the type-default key, so the framework reports:

```
failed to create missing data producers - no default producer found for missing data key: PrefixCacheMatchInfoDataKey/cpu-prefix-cache-producer, which is consumed by: cpu-prefix-cache-scorer
```

This change also affects line 155 (`pod.Put`): the producer no longer writes a per-instance key at runtime, so any consumer that resolves through `WithNonEmptyProducerName(<custom name>)` will miss the data even at request time.

If the goal is just to make the director gate able to read the data without knowing the producer name, consider keeping the producer per-instance and resolving the read key in the director through configuration (or iterating the published keys), instead of forcing all approxprefix instances to write to one shared key.

At minimum: this is a separate concern from conditional-decode and arguably belongs in its own PR/issue.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
```diff
@@ -288,6 +288,16 @@ func (h *Handler) Pick(ctx context.Context, request *scheduling.InferenceRequest
 		return map[string]scheduling.SchedulerProfile{}
 	}
 
+	// Conditional-decode short-circuit: when the coordinator marks the request
+	// with "Prefer: if-available", the cache-availability gate in the director
+	// will already have rejected it with 412 if no decode worker has the prompt
+	// cached. By the time we reach this point the request is going through.
```
elevran: I'm not clear on the comment.
Can you please attempt to reexplain?

pkg/epp/requestcontrol/director.go
```diff
@@ -57,6 +59,35 @@ const (
...
+	}
+	primary, ok := result.ProfileResults[result.PrimaryProfileName]
+	if !ok || primary == nil || len(primary.TargetEndpoints) == 0 {
+		return false
+	}
+	endpoint := primary.TargetEndpoints[0]
+	if endpoint == nil {
+		return false
+	}
+	raw, ok := endpoint.Get(attrprefix.PrefixCacheMatchInfoKey)
```
elevran: Question: this reads the bare default key (type-default approx-prefix producer name). It will not see data written by the `precise_prefix_cache` producer, which still writes per-instance via `WithNonEmptyProducerName(name)` (see `pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go:189,207`).

So a deployment using precise prefix cache as its KV cache producer will always trip the gate and get 412 even when the cache covers the prompt.

Is conditional-decode intended to support only approximate prefix cache for now? If so, please call that out in a comment here and in the user-facing doc. If not, the gate needs a way to find the right producer's data (e.g., resolving by consumer's configured `prefixMatchInfoProducerName`).

pkg/common/routing/common.go
```diff
@@ -26,3 +39,27 @@ func StripScheme(endpoint string) string {
 	}
 	return u.Host
 }
+
+// IsConditionalDecode reports whether the request headers carry the standard
+// HTTP "Prefer: if-available" preference (RFC 7240). It is the marker the
+// coordinator uses to flag a speculative early-decode attempt: route to a
+// decode worker only if its KV cache already covers the prompt(at least partially);
```
elevran: Minor: missing space.

```suggestion
// decode worker only if its KV cache already covers the prompt (at least partially);
```

elevran (commented):
Conditional-decode mechanism reads cleanly - thanks. Two concerns block this PR:

1. The producer key change in `approximateprefix` breaks the well-known config `tiered-prefix-cache-cpu` and the test run is failing on `TestWellKnownConfigs/tiered-prefix-cache-cpu` for that reason.
2. Comment in disagg_profile_handler.go is not clear.

Also a question about whether the director gate is intended to support the precise_prefix_cache producer, since it still writes per-instance keys.

roytman:
Dima, we shouldn't change `disagg_profile_handler` at all. This handler is meant for a single EPP that chooses the best Pods for all disaggregated steps. In the scenario we discussed, we are delegating this functionality to the Coordinator, which will send requests to dedicated EPPs via the gateway. Those EPPs will then use the `single-profile-handler`.

 For example, to decode EPP
 ```yaml
apiVersion: llm-d.ai/v1alpha1
kind: EndpointPickerConfig
plugins:
- type: decode-filter           # keeps only decode pods
- type: prefix-cache-scorer     # triggers auto-creation of approx-prefix-cache-producer
- type: max-score-picker
schedulingProfiles:
- name: default
  plugins:
  - pluginRef: decode-filter
    weight: 1.0
  - pluginRef: max-score-picker
 ```
 Note: you do not need to explicitly add `approx-prefix-cache-producer` to the plugins list - it gets auto-created because `prefix-cache-scorer` declares it needs `PrefixCacheMatchInfoDataKey` and nothing produces it yet. This is consistent with the existing `[epp-estimate-prefix-cache-config.yaml](https://github.com/llm-d/llm-d-router/blob/main/deploy/config/epp-estimate-prefix-cache-config.yaml)` which works the same way.
 
 You need to revert approximateprefix/plugin.go, which @elevran mentioned before.

dmitripikus:
@roytman , I've addressed your last comment. Thanks!

roytman:
@dmitripikus, please resolve the conflicts

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
dmitripikus: Hi @elevran , Thanks!
I've reverted in commit d26af3c3 — producer is back to per-instance keys (dk = WithNonEmptyProducerName(name)) on both the Produces() declaration and the pod.Put. 
tiered-prefix-cache-cpu should be OK now.

I agree that the director-side read is a separate concern from conditional decode. 
So, as you proposed, I'll make a fix with the gate resolving the producer's key, in additional PR.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
dmitripikus: Hi @elevran , Thanks!
That whole block has been removed in d26af3c3, as 'disagg_profile_handler' is not involved in this PR.
(Alexey's comment below)

pkg/epp/requestcontrol/director.go
dmitripikus: Hi @elevran ,
For now the gate reads string  PrefixCacheMatchInfoDataKey.String(), which serializes to the default-named approx producer's entry. Configs using precise prefix cache (or any custom-named approx instance) write to a different slot, so the gate's read misses and the coordinator gets 412 on every conditional-decode request — falling back to the full pipeline rather than serving wrong results, but the optimization is disabled for those configurations.
 The user-facing docs in this repo do not reflect Coordinator-based scenarios yet. So if this is OK, I'll add clarification to Release Notes, and I'll link additional PR with  fix in director, to this PR.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
```diff
@@ -154,7 +152,7 @@ func (p *dataProducer) Produce(ctx context.Context, request *fwksched.InferenceR
 
 	for _, pod := range pods {
 		matchLen := prefixCacheServers[ServerID(pod.GetMetadata().NamespacedName)]
-		pod.Put(p.dk.String(), attrprefix.NewPrefixCacheMatchInfo(matchLen, total, blockSize))
+		pod.Put(attrprefix.PrefixCacheMatchInfoKey, attrprefix.NewPrefixCacheMatchInfo(matchLen, total, blockSize))
```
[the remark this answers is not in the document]
dmitripikus: The problem is addressed in other responses. Thanks!

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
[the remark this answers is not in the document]
dmitripikus: The whole section is removed (as specified in other comments). Thanks!

pkg/epp/requestcontrol/director.go
```diff
@@ -254,9 +285,32 @@ func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestCo
 
 	result, err := d.scheduler.Schedule(ctx, reqCtx.SchedulingRequest, snapshotOfCandidatePods)
 	if err != nil {
+		// Preserve typed errcommon.Error from the scheduler so its status code
+		// (e.g. PreconditionFailed) reaches Envoy intact. Other errors fall
+		// through to ResourceExhausted, the legacy "no endpoint" status.
+		if e, ok := err.(errcommon.Error); ok {
+			return reqCtx, e
+		}
```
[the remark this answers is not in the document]
dmitripikus: done

pkg/epp/requestcontrol/director.go
```diff
@@ -254,9 +285,32 @@ func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestCo
...
+	// surface 412 Precondition Failed so the coordinator restarts the pipeline
+	// at encode/prefill/decode. Lives in the director (not in a profile handler)
+	// so it fires regardless of which profile handler is configured.
+	if routing.IsConditionalDecode(reqCtx.Request.Headers) {
+		if !primaryEndpointHasCachedPrefix(result) {
+			logger.V(logutil.DEBUG).Info("conditional-decode: chosen decode worker has no cached prefix, returning 412")
+			return reqCtx, errcommon.Error{
+				Code: errcommon.PreconditionFailed,
+				Msg:  "no decode worker has the requested KV cache",
+			}
```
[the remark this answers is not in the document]
dmitripikus: already addressed

pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
```diff
@@ -22,7 +22,15 @@ import (
...
 
-var PrefixCacheMatchInfoDataKey = plugin.NewDataKey("PrefixCacheMatchInfoDataKey", approxprefixconstants.ApproxPrefixCachePluginType)
+var (
+	PrefixCacheMatchInfoDataKey = plugin.NewDataKey("PrefixCacheMatchInfoDataKey", approxprefixconstants.ApproxPrefixCachePluginType)
+
+	// PrefixCacheMatchInfoKey is the pre-computed serialized form of
+	// PrefixCacheMatchInfoDataKey. Both the approximate-prefix DataProducer
+	// (writer) and any consumers that need to read the default-producer entry
+	// off an endpoint should reference this constant.
+	PrefixCacheMatchInfoKey = PrefixCacheMatchInfoDataKey.String()
```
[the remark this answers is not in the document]
dmitripikus: done

dmitripikus:
@roytman, conflicts are resolved, Thanks!

pkg/common/routing/common.go
```diff
@@ -29,3 +42,27 @@ func StripScheme(endpoint string) string {
...
+// decode worker only if its KV cache already covers the prompt (at least partially);
+// otherwise EPP surfaces 412 Precondition Failed so the coordinator restarts the
+// pipeline.
+//
+// Per RFC 7240 the Prefer header value is a comma-separated list of preference
+// tokens, each with optional ";"-delimited parameters. This function matches
+// the bare "if-available" token case-insensitively, ignoring surrounding
+// whitespace, parameters, and any other tokens that may appear alongside it.
+//
+// Headers are expected to be lowercased keys (EPP receives them that way from
```
roytman: nit: do we need this comment?

pkg/epp/requestcontrol/director.go
```diff
@@ -57,6 +60,35 @@ const (
...
 )
 
+// primaryEndpointHasCachedPrefix reports whether the primary profile's chosen
+// endpoint has at least one matching prefix block in its KV cache, as observed
+// by a precise/approximate-prefix scorer during the decode profile run. It
+// returns false when the result is missing, the primary profile produced no
+// endpoint, the endpoint carries no PrefixCacheMatchInfo attribute, or the
+// recorded match has zero blocks.
+func primaryEndpointHasCachedPrefix(result *fwksched.SchedulingResult) bool {
+	if result == nil {
```
roytman, line 74: We have many cases where the method returns `false`. A message at the `V(logutil.VERBOSE)` or' V(logutil.DEBUG) ' log level can be useful for distinguishing them.

pkg/epp/requestcontrol/director_test.go
```diff
@@ -1595,3 +1596,169 @@ func newResponseBodyTestRequestContext(requestID string, completionTokens int) *
...
+	cfg := NewConfig().WithAdmissionPlugins(newMockAdmissionPlugin("admit", nil))
+	candidates := NewCachedEndpointCandidates(context.Background(), NewDatastoreEndpointCandidates(ds), time.Minute)
+	dir := NewDirectorWithConfig(ds, mockSched, &mockAdmissionController{}, candidates, cfg)
+	return dir, ctx
+}
+
+func TestDirector_HandleRequest_ConditionalDecode(t *testing.T) {
+	scheduleResultWith := func(matched, total int) *fwksched.SchedulingResult {
+		attrs := fwkdl.NewAttributes()
+		if matched >= 0 {
```
roytman, line 1711: If `matched < 0, it skips the Put entirely, testing the "attribute not present" path. 
But there's no case where the attribute key exists but holds a value of the wrong type (e.g., a plain string instead of `*attrprefix.PrefixCacheMatchInfo`), which would exercise this branch in primaryEndpointHasCachedPrefix:
```goLang
info, ok := raw.(*attrprefix.PrefixCacheMatchInfo)
if !ok {
    return false
}
```

Another place for that specific assertion test would actually be in the sibling unit test `TestPrimaryEndpointHasCachedPrefix`, since it tests the function in isolation and already covers `nil/empty/missing` cases. Adding one more case there, like {"wrong type in attribute", resultWithWrongType(), false} would cover it.

pkg/common/routing/common.go
```diff
@@ -29,3 +42,27 @@ func StripScheme(endpoint string) string {
 	}
 	return u.Host
 }
+
+// IsConditionalDecode reports whether the request headers carry the standard
+// HTTP "Prefer: if-available" preference (RFC 7240). It is the marker the
+// coordinator uses to flag a speculative early-decode attempt: route to a
+// decode worker only if its KV cache already covers the prompt (at least partially);
+// otherwise EPP surfaces 412 Precondition Failed so the coordinator restarts the
```
roytman: nit: some duplication in the comment from the line 29

roytman (commented):
several small comments

pkg/common/routing/common.go
dmitripikus: Removed redundant commit. Thanks!

pkg/epp/requestcontrol/director.go
dmitripikus: Done. Thanks!

pkg/epp/requestcontrol/director_test.go
dmitripikus: Done. Thanks!

pkg/common/routing/common.go
dmitripikus: Fixed. Thanks!

elevran:
/approve

One documented limitation (not a blocker): the gate reads only the default approx-prefix-cache-producer key. Precise-prefix or custom-named approx-prefix configurations will always get 412 on conditional-decode, falling back to the full pipeline. Correctness is preserved; the release note calls this out explicitly.
Worth a follow-up issue if broader key support is desired later.

PR #1282, opened 2026-05-24
fix(helm): align verify-helm.sh --set paths with router.* schema

weizhoublue:
**What type of PR is this?**
 
/kind bug
 
**What this PR does / why we need it**:

`hack/verify-helm.sh` still referenced outdated Helm values (`inferenceExtension.*`, `inferencePool.*`)

  | Legacy | Current |
  |--------|---------|
  | `inferencePool.modelServers.matchLabels` | `router.modelServers.matchLabels` |
  | `inferencePool.replicas` | `router.replicas` |
  | `inferenceExtension.latencyPredictor.enabled` | `router.latencyPredictor.enabled` |
  | `inferenceExtension.endpointsServer.endpointSelector` | `router.modelServers.matchLabels` |
  | `inferenceExtension.endpointsServer.createInferencePool` | `router.inferencePool.create` |
  | `inferenceExtension.endpointsServer.modelServerType` | `router.modelServers.type` |
  | `inferenceExtension.endpointsServer.targetPorts` | `router.modelServers.targetPorts` |
  | `inferenceExtension.sidecar.*` | `router.proxy.*` |
  | `inferenceExtension.flags` | `router.epp.flags` |
  | `inferenceExtension.extraServicePorts` | `router.extraServicePorts` |

**Which issue(s) this PR fixes**:

Fixes #

**Release note** _(write `NONE` if no user-facing change)_:
```release-note
NONE
```

PR #1296, opened 2026-05-24
Fix : correct targetPorts format in verify-helm-charts and add CI validation job

weizhoublue:
/kind bug

1. verify-helm.sh used --set 'router.modelServers.targetPorts[0]=8000', but the template expects the struct
  format --set 'router.modelServers.targetPorts[0].number=8000'

2. Running make verify-helm-charts  in CI

Fixes #

```release-note
NONE
```


weizhoublue:
@ahg-g  hi, this is a follow-up PR for https://github.com/llm-d/llm-d-router/pull/1282

PR #1305, opened 2026-05-25
Fix:  config drift of vllmConfig.URL after field rename 

weizhoublue:
/kind bug

PR #1079 renamed vllmConfig.HTTP → vllmConfig.URL with JSON tag http → url. The code change was made, but two config  files in deploy/config/ were missed during the rename update.

EPP fails to start with configuration loading error due to unknown field http in vllm block 

Fixes #

```release-note
NONE
```


weizhoublue:
@vMaroon   hi, could you check whether this PR makes sense,  thanks

elevran (approved):

PR #1311, opened 2026-05-25
Fourth step in refactoring of the Flow Control layer

shmuelk:
**What type of PR is this?**
/kind cleanup

This PR is the forth step in a process of simplifying the Flow Control component's code base.

In particular this PR:
1. The FlowController's processor is now started via a dedicated Start() function, usually called right after creating the FlowController. Various tests are noted exceptions to this rule.
2. The FlowController's processor is now referenced via a simple field in the FlowController.
3. The word shard has been removed in many places, in types names (ShardProcessor -> Processor), in comments, and in documentation.

**Notes** to the reviewers:
1. The removal of of the word shard in the code base is an on going effort. It will be completed in future PRs
2. Nits from previous PRs will be handled in a future PR.

Refs: #889

```release-note
NONE
```

### DIFF
--- a/pkg/epp/flowcontrol/contracts/errors.go
+++ b/pkg/epp/flowcontrol/contracts/errors.go
@@ -31,13 +31,6 @@ var (
 	// ErrPolicyQueueIncompatible indicates that a selected policy is not compatible with the capabilities of the queue.
 	ErrPolicyQueueIncompatible = errors.New("policy is not compatible with queue capabilities")
 
-	// ErrInvalidShardCount indicates that an invalid shard count was provided (e.g., zero or negative).
-	ErrInvalidShardCount = errors.New("invalid shard count")
-
-	// ErrShardDraining indicates that an operation could not be completed because the target shard is in the process of
-	// being gracefully drained. The caller should retry the operation on a different, Active shard.
-	ErrShardDraining = errors.New("shard is draining")
-
 	// ErrInvalidQueueItemHandle indicates that a QueueItemHandle provided to a SafeQueue operation (e.g.,
 	// SafeQueue.Remove()) is not valid for that queue, has been invalidated, or does not correspond to an actual item in
 	// the queue.
--- a/pkg/epp/flowcontrol/contracts/registry.go
+++ b/pkg/epp/flowcontrol/contracts/registry.go
@@ -121,8 +121,7 @@ type ActiveFlowConnection interface {
 //
 // Conformance: Implementations MUST be goroutine-safe.
 type ManagedQueue interface {
-	// Add attempts to enqueue an item, performing an atomic check on the parent shard's lifecycle state before adding
-	// the item to the underlying queue.
+	// Add attempts to enqueue an item.
 	Add(item flowcontrol.QueueItemAccessor) error
 
 	// Remove atomically finds and removes an item from the underlying queue using its handle.
--- a/pkg/epp/flowcontrol/contracts/registry.go
+++ b/pkg/epp/flowcontrol/contracts/registry.go
@@ -38,14 +38,6 @@ import (
 //     When the lease count drops to zero, the flow becomes "Idle".
 //  4. Garbage Collection: The implementation MUST automatically garbage collect a flow after it has remained
 //     continuously Idle for a configurable duration.
-//
-// # System Invariants
-//
-// Concrete implementations MUST uphold the following invariants:
-//
-//  1. Shard Consistency: All configured priority bands and registered flow instances must exist on every Active shard.
-//  2. Capacity Partitioning: Global and per-band capacity limits must be uniformly partitioned across all Active
-//     shards.
 type FlowRegistry interface {
 	FlowRegistryObserver
 	FlowRegistryDataPlane
@@ -83,23 +75,23 @@ type FlowRegistryDataPlane interface {
 	// ErrFlowInstanceNotFound if no instance exists for the given key.
 	ManagedQueue(key flowcontrol.FlowKey) (ManagedQueue, error)
 
-	// FairnessPolicy retrieves the FairnessPolicy singleton configured for the specified priority band on this shard.
+	// FairnessPolicy retrieves the FairnessPolicy singleton configured for the specified priority band.
 	// This method provides access to the immutable logic component that governs inter-flow contention.
 	// The registry guarantees that a non-nil policy is returned for any active priority band.
 	//
 	// Returns:
 	//   - FairnessPolicy: The active policy instance.
-	//   - error: A wrapped ErrPriorityBandNotFound if the priority level is not configured on this shard.
+	//   - error: A wrapped ErrPriorityBandNotFound if the priority level is not configured.
 	FairnessPolicy(priority int) (flowcontrol.FairnessPolicy, error)
 
 	// PriorityBandAccessor retrieves the read-only view of the "Flow Group" for a specific priority level.
-	// This accessor provides the state of all contending flows within the band (as seen by this shard) and serves as the
+	// This accessor provides the state of all contending flows within the band and serves as the
 	// primary input for FairnessPolicy execution.
 	//
 	// Returns an error wrapping ErrPriorityBandNotFound if the priority level is not configured.
 	PriorityBandAccessor(priority int) (flowcontrol.PriorityBandAccessor, error)
 
-	// AllOrderedPriorityLevels returns all configured priority levels that this shard is aware of, sorted in descending
+	// AllOrderedPriorityLevels returns all configured priority levels, sorted in descending
 	// numerical order. This order corresponds to highest priority (highest numeric value) to lowest priority (lowest
 	// numeric value).
 	// The returned slice provides a definitive, ordered list of priority levels for iteration, for example, by a
@@ -123,15 +115,14 @@ type ActiveFlowConnection interface {
 	FlowKey() flowcontrol.FlowKey
 }
 
-// ManagedQueue defines the interface for a flow's queue on a specific shard.
+// ManagedQueue defines the interface for a flow's queue.
 // It acts as a stateful decorator that *use an underlying SafeQueue, augmenting it with statistics tracking, and
-// lifecycle awareness (e.g., rejecting adds when a shard is draining).
+// lifecycle awareness.
 //
 // Conformance: Implementations MUST be goroutine-safe.
 type ManagedQueue interface {
 	// Add attempts to enqueue an item, performing an atomic check on the parent shard's lifecycle state before adding
 	// the item to the underlying queue.
-	// Returns ErrShardDraining if the parent shard is no longer Active.
 	Add(item flowcontrol.QueueItemAccessor) error
 
 	// Remove atomically finds and removes an item from the underlying queue using its handle.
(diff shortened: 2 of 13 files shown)


pkg/epp/flowcontrol/contracts/errors.go
```diff
@@ -31,13 +31,6 @@ var (
 	// ErrPolicyQueueIncompatible indicates that a selected policy is not compatible with the capabilities of the queue.
 	ErrPolicyQueueIncompatible = errors.New("policy is not compatible with queue capabilities")
 
-	// ErrInvalidShardCount indicates that an invalid shard count was provided (e.g., zero or negative).
-	ErrInvalidShardCount = errors.New("invalid shard count")
-
-	// ErrShardDraining indicates that an operation could not be completed because the target shard is in the process of
-	// being gracefully drained. The caller should retry the operation on a different, Active shard.
-	ErrShardDraining = errors.New("shard is draining")
```
LukeAVanDrie, line 39: This is still referenced in the registry data plane and tests which were not updated in this PR.

Specifically, the following lines still reference it:

- `pkg/epp/flowcontrol/registry/managedqueue.go:L128`

```go
if mq.isDraining() {
    return contracts.ErrShardDraining
}
```

- `pkg/epp/flowcontrol/registry/managedqueue_test.go:L144`

```go
expectErrIs: contracts.ErrShardDraining,
```

Since the shard layer has been removed, we can update `managedqueue.go` to remove the draining check in `Add()`, and delete the draining test case `"ShouldFail_AndNotChangeStats_WhenQueueIsDraining"` from `managedqueue_test.go`.

pkg/epp/flowcontrol/controller/controller.go
LukeAVanDrie, line 1: `fc.processor` is only initialized when `fc.Start()` is called. If we instantiate the controller and immediately call `EnqueueAndWait` before calling `Start()`, this will trigger a nil pointer dereference panic. While in `runner.go`, we call `Start()` sequentially, this creates a fragile lifecycle dependency.

We should consider Instantiating `fc.processor` directly inside the `NewFlowController` constructor (using the default factory) so that the field is never nil after construction, and let `Start()` only be responsible for spawning the background runner goroutine (`fc.processor.Run`).

pkg/epp/flowcontrol/controller/config.go
```diff
@@ -48,7 +48,7 @@ type Config struct {
 	// Optional: Defaults to `defaultProcessorReconciliationInterval` (5 seconds).
 	ProcessorReconciliationInterval time.Duration
```
LukeAVanDrie: Please remove this dead code related to `ProcessorReconciliationInterval`:

- `defaultProcessorReconciliationInterval` constant
- `ProcessorReconciliationInterval field from Config
- `WithProcessorReconciliationInterval` builder option
- The validation check in `validate()`
- The ignored overrides (e.g., `ProcessorReconciliationInterval: 1 * time.Hour`) in `benchmark.go` and `benchmark_test.go`
- Dead assertions in `config_test.go` and `controller_test.go`

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -109,9 +101,6 @@ type FlowController struct {
 
 	// --- Concurrent state ---
 
-	// worker is a highly concurrent pointer storing the managedWorker.
-	worker atomic.Pointer[managedWorker]
-
 	// wg waits for all worker goroutines to terminate during shutdown.
 	wg sync.WaitGroup
```
LukeAVanDrie: This is completely dead because `Wait()` is never called anywhere in the codebase (the `shutdown()` method was deleted in your previous PR). Spawning the processor via `fc.wg.Go` is identical to a raw goroutine.

LukeAVanDrie (changes requested):
The package-level `doc.go` files were missed in this refactoring and still describe the old sharded, supervisor-worker pool architecture.

Please update the following files to describe the new simplified single-worker/single-actor queuing model, or create an issue for tracking this. I am happy to pick this up once this PR is merged.

- `pkg/epp/flowcontrol/controller/doc.go`
- `pkg/epp/flowcontrol/controller/internal/doc.go`
- `pkg/epp/flowcontrol/registry/doc.go`

pkg/epp/flowcontrol/contracts/errors.go
shmuelk: I don't find the code you are referencing in the code base. It was removed in the previous PR

shmuelk:
@LukeAVanDrie said:

> The package-level doc.go files were missed in this refactoring and still describe the old sharded, supervisor-worker pool architecture.
>
> Please update the following files to describe the new simplified single-worker/single-actor queuing model, or create an issue for tracking this. I am happy to pick this up once this PR is merged.
>
> pkg/epp/flowcontrol/controller/doc.go
> pkg/epp/flowcontrol/controller/internal/doc.go
> pkg/epp/flowcontrol/registry/doc.go

As was mentioned in the PR description this PR doesn't completely update the documentation of FlowControl in an attempt to keep this PR a manageable size.

LukeAVanDrie:
> As was mentioned in the PR description this PR doesn't completely update the documentation of FlowControl in an attempt to keep this PR a manageable size.

Sounds good, I just want to make sure this is tracked as a followup. I am also happy to look into doc updates this week

pkg/epp/flowcontrol/contracts/errors.go
LukeAVanDrie: Oops, you're right. I pulled down the different branches in the chain to review them and must have gotten my local git state mixed up. Thanks for catching that, please ignore this thread!

shmuelk:
@LukeAVanDrie I believe I have corrected the issues you raised. Please review this PR again. Thanks.

LukeAVanDrie (approved):

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -16,18 +16,16 @@ limitations under the License.
 
 // Package controller contains the implementation of the FlowController engine.
 //
-// The FlowController is the central processing engine of the Flow Control layer. It is a sharded, high-throughput
+// The FlowController is the central processing engine of the Flow Control layer. It is a high-throughput
 // component responsible for managing the lifecycle of all incoming requests. It achieves this by acting as a stateless
-// supervisor that orchestrates a pool of stateful workers (ShardProcessors), distributing incoming requests among them.
+// supervisor that orchestrates a  stateful worker (SProcessor).
```
elevran: Looks like a botched rename: `(SProcessor)` should be `(Processor)`, and there's a double space before `stateful`.

```suggestion
// supervisor that orchestrates a stateful worker (Processor).
```

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -123,6 +106,7 @@ type Deps struct {
 	EndpointCandidates contracts.EndpointCandidates
 	UsageLimitPolicy   flowcontrol.UsageLimitPolicy
 	Clock              clock.WithTicker
+	ProcessorFactory   processorFactory
 }
 
 // NewFlowController creates and starts a new FlowController instance.
```
elevran, line 112: Doc is now stale: `NewFlowController` no longer starts the processor goroutine; that's `Start()`'s job. Worth updating so callers don't assume the old behavior.

```suggestion
// NewFlowController creates a new FlowController instance. The caller must invoke Start to
// begin processing requests. The provided context governs the lifecycle of the controller and its worker.
```

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -147,32 +131,56 @@ func NewFlowController(
...
+		fc.clock,
+		fc.config.ExpiryCleanupInterval,
+		fc.config.EnqueueChannelBufferSize,
+		fc.logger,
+	)
+
+	return fc
+}
+
+func (fc *FlowController) Start() {
```
elevran: Two small things on `Start`:

1. Exported method without a doc comment.
2. The contract isn't documented. As written, calling `Start` twice spawns two goroutines reading the same `enqueueChan`, which would race on items. Worth either documenting that it must be called exactly once, or guarding with `sync.Once` / an atomic flag.

Related lifecycle concern (separate but adjacent): now that startup is decoupled from `NewFlowController`, a caller who forgets `Start()` gets a silent hang. `Submit` succeeds while there's buffer space, but no goroutine drains the channel, so `EnqueueAndWait` blocks until the request context cancels with no signal that the controller never started. The previous `getOrStartWorker` made this impossible. Options:
- Call `Start()` from `NewFlowController` (the two existing call sites already do `New + Start` immediately).
- Have `Submit`/`SubmitOrBlock` return `ErrFlowControllerNotRunning` if `Start` was never called.

Not blocking, but worth a thought before this stabilizes.

pkg/epp/flowcontrol/controller/controller_test.go
```diff
@@ -841,13 +842,12 @@ func TestFlowController_EnqueueAndWait(t *testing.T) {
 }
 
 // TestFlowController_WorkerManagement covers the lifecycle of the processor (worker), including startup
-// and shutdown.
 func TestFlowController_WorkerManagement(t *testing.T) {
 	t.Parallel()
 
-	// Reconciliation validates that the controller correctly identifies and shuts down workers whose shards no longer
+	// Startup validates that the controller correctly identifies and shuts down workers whose shards no longer
```
elevran: Comment is stale relative to the renamed test: this no longer validates reconciliation or worker shutdown, just that the worker goroutine starts. Either rewrite to match the test (e.g. `Startup validates that calling Start launches the processor goroutine.`) or drop it.

pkg/epp/flowcontrol/controller/controller_test.go
```diff
@@ -856,130 +856,33 @@ func TestFlowController_WorkerManagement(t *testing.T) {
...
 				return contracts.AggregateStats{}
 			}}
-		h := newUnitHarness(t.Context(), t, &Config{}, mockRegistry)
+
+		// Initialize the processor mock with the channel needed to synchronize startup.
+		processor := &mockProcessor{runStarted: make(chan struct{})}
+
+		h := newUnitHarness(t.Context(), t, &Config{}, mockRegistry, processor)
 
 		// Pre-populate the controller with initial worker, simulating a previous state.
```
elevran: `Pre-populate the controller with initial worker, simulating a previous state.` is leftover framing from the old Reconciliation test. The new Startup test isn't simulating any prior state, so this comment can go.

pkg/epp/flowcontrol/contracts/registry.go
```diff
@@ -123,15 +115,14 @@ type ActiveFlowConnection interface {
...
 
-// ManagedQueue defines the interface for a flow's queue on a specific shard.
+// ManagedQueue defines the interface for a flow's queue.
 // It acts as a stateful decorator that *use an underlying SafeQueue, augmenting it with statistics tracking, and
-// lifecycle awareness (e.g., rejecting adds when a shard is draining).
+// lifecycle awareness.
 //
 // Conformance: Implementations MUST be goroutine-safe.
 type ManagedQueue interface {
 	// Add attempts to enqueue an item, performing an atomic check on the parent shard's lifecycle state before adding
```
elevran: Contract doc has drifted: this PR removed `ErrShardDraining` and the matching sentence below, but the `parent shard's lifecycle state` clause here still describes the removed behavior. Either trim this to just `Add attempts to enqueue an item.` or describe whatever lifecycle check (if any) `Add` is still expected to perform.

```suggestion
	// Add attempts to enqueue an item.
```

elevran (commented):
/lgtm
/approve
/hold

A few small things below, mostly around stale doc/comment text that drifted. Happy for the doc-only items to land in the follow-up cleanup PR you already mentioned in the description; flagging here so they don't get lost.

pkg/epp/flowcontrol/controller/controller.go
shmuelk: I have merged Start() into NewFlowController(), historically this was done lazily which was exploited by the tests. Changes were made to get around this which really made the separation not necessary.

Thanks.

elevran (approved):

PR #1314, opened 2026-05-25
fix(ci): go module cache never saved due to permission error

sagearc:
**Context:**
Docker named volumes are root-owned. The `actions/cache` post-step runs as the runner user, hits `EACCES` on the volume mountpoint, and silently gives up — the cache is never saved or restored.

```
Cache not found + EACCES: permission denied, lstat '/var/lib/docker/volumes/llm-d-gomodcache/_data'
```

**Before:**
Silent permission error in teardown; Go re-downloads all modules on every run.

**After:**
Cache saves on first run and restores on subsequent ones with the same `go.sum`.

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -105,13 +105,12 @@ jobs:
       - name: Verify go.mod and go.sum are tidy
         run: go mod tidy -diff
 
-      - name: Create and resolve Go cache volumes
+      - name: Create Go cache dirs
         id: go-cache
         run: |
-          docker volume create llm-d-gomodcache
-          docker volume create llm-d-gobuildcache
-          echo "mod=$(docker volume inspect llm-d-gomodcache -f '{{.Mountpoint}}')" >> $GITHUB_OUTPUT
-          echo "build=$(docker volume inspect llm-d-gobuildcache -f '{{.Mountpoint}}')" >> $GITHUB_OUTPUT
+          mkdir -p "$HOME/.cache/llm-d-gomodcache" "$HOME/.cache/llm-d-gobuildcache"
+          echo "mod=$HOME/.cache/llm-d-gomodcache" >> "$GITHUB_OUTPUT"
+          echo "build=$HOME/.cache/llm-d-gobuildcache" >> "$GITHUB_OUTPUT"
 
       - name: Cache Go modules and build cache
         uses: actions/cache@v5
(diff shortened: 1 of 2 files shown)


test/e2e/e2e_test.go
```diff
@@ -421,7 +421,7 @@ var _ = ginkgo.Describe("Run end to end tests", ginkgo.Ordered, func() {
 		})
 	})
 
-	ginkgo.When("Running an E/PD (Encode/Prefill-Decode) configuration", func() {
+	ginkgo.When("Running an E/PD (Encode/Prefill-Decode) configuration", ginkgo.Label("Extended"), func() {
 		ginkgo.It("should route multimodal requests through encode and decode pods", func() {
```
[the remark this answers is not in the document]
sagearc: Stale — `test/e2e/e2e_test.go` is not part of this PR. That change lives in #1315.

.github/workflows/ci-pr-checks.yaml
```diff
@@ -147,10 +146,16 @@ jobs:
 
       - name: Run unit tests
         shell: bash
+        env:
```
elevran: These env vars are also consumed by `make lint` (line 127), `make build` (line 136), and `make test-e2e` (line 250) via `BUILDER_RUN`. Without setting them on those steps too, they keep using the default named volumes and don't benefit from the bind-mount cache. Consider hoisting to job-level `env:` so every step inherits:

```yaml
  lint-and-test:
    needs: check-changes
    runs-on: ubuntu-latest
    timeout-minutes: 90
    env:
      GO_MOD_CACHE_VOL: ${{ steps.go-cache.outputs.mod }}
      GO_BUILD_CACHE_VOL: ${{ steps.go-cache.outputs.build }}
```

Then the per-step env blocks here and on the integration test step can go away.

elevran (commented):
The env block is only set on the two test steps. Other targets fall back to the named-volume defaults (`llm-d-gomodcache` / `llm-d-gobuildcache`) and re-create the root-owned volumes... So `lint` and `build` (which run before the tests) keep downloading modules into a throwaway named volume on every run, and the bind-mount cache only helps the test steps.

Simplest fix is to move `GO_MOD_CACHE_VOL` / `GO_BUILD_CACHE_VOL` to the `lint-and-test` job's `env:` so all steps inherit them.

.github/workflows/ci-pr-checks.yaml
sagearc: done

sagearc:
Thanks, moved the cache vars to job scope @elevran

shmuelk:
Can we please make sure the end to end test runs locally on one's laptop?

Right now it fails due to those Docker volumes that don't exist on people's laptops.

sagearc:
@shmuelk Verified locally, go cache bind-mounts correctly, no EACCES.
Full e2e gets through image builds and into the test container. There's unrelated rancher specific pre-existing issue though

PR #1319, opened 2026-05-25
Feat/add currequest tokens

kaushikmitr:
## What does this PR do?

Extends the `InFlightLoadProducer` and `TokenLoadScorer` to factor the *currently scheduling request's* projected impact into endpoint load scoring, in addition to the existing in-flight load. Without this, load-aware scorers see only past commitments — a 50-token completion and a 5000-token long-context request look identical when scoring an endpoint, and big requests can pile onto already-saturated endpoints.

The projection is per-endpoint: each candidate gets a value that accounts for that endpoint's prefix-cache hit rate, so endpoints with better cache fit for this prompt naturally appear cheaper.

## Design

A new field `UncachedRequestTokens` is added to `attrconcurrency.InFlightLoad`:

```go
type InFlightLoad struct {
    Tokens                int64  // past commitments: accumulated across in-flight requests
    Requests              int64  // past commitments: accumulated across in-flight requests
    UncachedRequestTokens int64  // speculative: this request's projected impact on this endpoint
}
```

Two semantically distinct values share one struct, transported through one data key (`InFlightLoadDataKey`):

- **`Tokens` / `Requests`** — *past commitments.* Updated by `PreRequest` (when the scheduler picks an endpoint) and decremented by `OnEvicted` (when the request's stream ends). The exact contribution per request is remembered in PluginState so the decrement is a perfect undo.
- **`UncachedRequestTokens`** — *speculative projection.* Computed fresh by `InFlightLoadProducer.Produce` each scheduling cycle from the request being scored and the endpoint's prefix-cache state. Never committed to endpoint state, never decremented — once the cycle ends, the value is discarded.

`TokenLoadScorer.Score` sums them: `tokenLoad := load.Tokens + load.UncachedRequestTokens`. Endpoints with better prefix-cache fit get lower `UncachedRequestTokens` (less uncached work) and thus score higher.

### Why both values share one struct

`Tokens` is what the endpoint *has committed to*. `UncachedRequestTokens` is what the endpoint *would commit to* if the scheduler picks it. The field doc strings document the lifecycle distinction explicitly. Carrying both on the same struct means scorers see a complete per-endpoint load picture from a single data key without needing to assemble two values.

### Why a single source of truth for the projection

`UncachedRequestTokens` could equivalently be computed inside any scorer that needs it. It's produced once, in `InFlightLoadProducer.Produce`, so every consumer reads the same value — single source of truth for the current request's per-endpoint impact, consistent across scorers, with the estimation logic owned in one place. The producer also reuses the same helper (`estimateRequestTokens`) in `PreRequest` to compute the committed value, guaranteeing the projection and the eventual commitment use identical math.

## What changed

**`pkg/epp/framework/plugins/datalayer/attribute/concurrency/data_types.go`**
- Added `UncachedRequestTokens int64` field to `InFlightLoad`
- Switched `Clone` to the value-copy idiom (`cp := *l; return &cp`) so future field additions can't be silently dropped
- Per-field doc strings making the past-commitment vs. speculative-projection distinction explicit

**`pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go`**
- `Produce` populates all three fields in a single `Put` per endpoint; `UncachedRequestTokens` is set only when a request is in scope
- New `estimateRequestTokens` helper consolidates the per-endpoint impact calculation, used by both `Produce` and `PreRequest`

**`pkg/epp/framework/plugins/scheduling/scorer/tokenload/token_load.go`**
- `Score` reads the `InFlightLoad` attribute and sums `Tokens + UncachedRequestTokens`
- Typed-nil guard on the deserialized `*InFlightLoad` so a nil-stored attribute can't panic

**Tests**
- Split `TestTokenLoadScorer` into named subtests:
  - `in-flight load only` — committed in-flight load alone produces correct scores
  - `in-flight load plus current request impact` — both contributors add to the score
  - `missing data key degrades gracefully` — absent attribute scores as zero load (1.0)
  - `typed nil attribute handles gracefully` — typed-nil `*InFlightLoad` doesn't panic, scores as zero load

## Behavior

For deployments not running any load-aware scorer that reads `UncachedRequestTokens`, this is a no-op: the field is populated but unread. For deployments running `TokenLoadScorer`, scoring now reflects request-aware load, so large requests will naturally route away from already-busy endpoints and toward endpoints with favorable prefix-cache state.

## Testing

- `go test ./pkg/epp/framework/plugins/datalayer/attribute/concurrency/... -race`
- `go test ./pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/... -race`
- `go test ./pkg/epp/framework/plugins/scheduling/scorer/tokenload/... -race`

New `TestTokenLoadScorer` subtests cover the new contribution path, the missing-key case, and the typed-nil case.

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/attribute/concurrency/data_types.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/concurrency/data_types.go
@@ -23,6 +23,7 @@ import (
 )
 
 var InFlightLoadDataKey = plugin.NewDataKey("InFlightLoadDataKey", inflightloadconstants.InFlightLoadProducerType)
+var CurrentRequestLoadDataKey = plugin.NewDataKey("CurrentRequestLoadDataKey", inflightloadconstants.InFlightLoadProducerType)
 
 // InFlightLoad captures the current real-time load of an endpoint as tracked by the EPP.
 type InFlightLoad struct {
--- a/pkg/epp/framework/plugins/datalayer/attribute/concurrency/data_types.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/concurrency/data_types.go
@@ -23,7 +23,7 @@ import (
 )
 
 var InFlightLoadDataKey = plugin.NewDataKey("InFlightLoadDataKey", inflightloadconstants.InFlightLoadProducerType)
-var CurrentRequestLoadDataKey = plugin.NewDataKey("CurrentRequestLoadDataKey", inflightloadconstants.InFlightLoadProducerType)
+var CurrentRequestEndpointImpactDataKey = plugin.NewDataKey("CurrentRequestEndpointImpactDataKey", inflightloadconstants.InFlightLoadProducerType)
 
 // InFlightLoad captures the current real-time load of an endpoint as tracked by the EPP.
 type InFlightLoad struct {
--- a/pkg/epp/framework/plugins/datalayer/attribute/concurrency/data_types.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/concurrency/data_types.go
@@ -23,6 +23,12 @@ import (
 )
 
 var InFlightLoadDataKey = plugin.NewDataKey("InFlightLoadDataKey", inflightloadconstants.InFlightLoadProducerType)
+
+// CurrentRequestEndpointImpactDataKey carries the estimated per-endpoint cost of the
+// request currently being scheduled. Set by InFlightLoadProducer.Produce
+// for each candidate endpoint based on that endpoint's prefix-cache match;
+// consumed by load-aware scorers to include the request's own impact when
+// scoring. Lifecycle: overwritten each scheduling cycle; never cleared.
 var CurrentRequestEndpointImpactDataKey = plugin.NewDataKey("CurrentRequestEndpointImpactDataKey", inflightloadconstants.InFlightLoadProducerType)
 
 // InFlightLoad captures the current real-time load of an endpoint as tracked by the EPP.
--- a/pkg/epp/framework/plugins/datalayer/attribute/concurrency/data_types.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/concurrency/data_types.go
@@ -22,27 +22,55 @@ import (
 	inflightloadconstants "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/constants"
 )
 
+// InFlightLoadDataKey carries the per-endpoint in-flight load snapshot used by
+// load-aware scorers. Populated by InFlightLoadProducer.Produce on each
+// candidate endpoint at the start of every scheduling cycle.
+//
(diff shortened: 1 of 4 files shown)


kaushikmitr:
@ahg-g this should be the final PR for now for tokenload scorer

pkg/epp/framework/plugins/datalayer/attribute/concurrency/data_types.go
```diff
@@ -24,6 +24,13 @@ import (
 
 var InFlightLoadDataKey = plugin.NewDataKey("InFlightLoadDataKey", inflightloadconstants.InFlightLoadProducerType)
 
+// CurrentRequestEndpointImpactDataKey carries the estimated per-endpoint cost of the
+// request currently being scheduled. Set by InFlightLoadProducer.Produce
+// for each candidate endpoint based on that endpoint's prefix-cache match;
+// consumed by load-aware scorers to include the request's own impact when
+// scoring. Lifecycle: overwritten each scheduling cycle; never cleared.
+var CurrentRequestEndpointImpactDataKey = plugin.NewDataKey("CurrentRequestEndpointImpactDataKey", inflightloadconstants.InFlightLoadProducerType)
```
ahg-g: why not place it in the same struct managed by the InFlightLoadDataKey instead of having it in a separate key?

pkg/epp/framework/plugins/datalayer/attribute/concurrency/data_types.go
ahg-g, line 64: Add here `uncachedRequestTokens`, we don't need to add the request count since it is always 1

pkg/epp/framework/plugins/datalayer/attribute/concurrency/data_types.go
kaushikmitr: i did this because this one does not persist, unlke InFlightLoadDataKey. We can combine them to single struct with some added comments to make sure its not confusing

ahg-g: can you pls explain what do you mean by it doesn't persist? it doesn't persist across what?

kaushikmitr: This: 
**InFlightLoadData** reflect past commitments. They're updated at PreRequest (after the scheduler picks an endpoint, so we know it's a real assignment) and decremented at EOS via OnEvicted. The tracker maintains them as the running in-flight load owned by each endpoint.
**CurrentRequestEndpointImpactData** reflects a hypothetical future. Computed in Produce, before the scheduler has chosen anything, projected onto every candidate endpoint as "this is what would happen if the request landed here." It's used to inform the scoring decision and then thrown away, nothing commits it to the endpoint's state, nothing decrements it later.
That is why I made the keys different, but agree it can be handled in the same struct. What do you think is cleaner?
In fact a scorer could compute **CurrentRequestEndpointImpactData** itself, I just centralized it in the producer so we are consistent (eg. estimated output token etc.)

Issue #1298, opened 2026-05-24
Remove legacy pool group designation (inference.networking.x-k8s.io)


zdtsw:
**What needs to happen:**

Remove the legacy API group `inference.networking.x-k8s.io` support from the router codebase. This includes:

- `LegacyPoolGroup` constant and its usage in `pkg/sidecar/proxy/proxy.go` and `pkg/sidecar/proxy/allowlist.go`
- Deprecated `inference.networking.x-k8s.io` config loader support in `pkg/epp/config/loader/configloader.go`
- Deprecation handling in `pkg/epp/controller/inferencemodelrewrite_reconciler.go` and `pkg/epp/controller/inferenceobjective_reconciler.go`
- Related test data and e2e references using the old API group

**Why:**

The legacy group `inference.networking.x-k8s.io` was superseded by the upstream GIE group change in [gateway-api-inference-extension v1.5.0](https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/tag/v1.5.0). The router currently supports both old and new groups via an allowlist, but the legacy group should be removed in v0.11.

**Context:**

- Discussed in PR #1200 review: https://github.com/llm-d/llm-d-router/pull/1200#discussion_r3278725917
- The dual-support allowlist is in `pkg/sidecar/proxy/allowlist.go`

/kind cleanup

zdtsw:
this probably should land into [v0.11 ](https://github.com/llm-d/llm-d-router/milestone/9)

PR #1321, opened 2026-05-25
chore(refactor): move shared const to pkg/common/routing

zdtsw:
- both pkg/epp and pkg/sidecar can use the same const with import

**What type of PR is this?**
/kind cleanup

follow up comments https://github.com/llm-d/llm-d-router/pull/1200#issuecomment-4534290460  move common const in to pkg/common thus epp and sidecar + tests can all use these

Fixes #

```release-note
NONE
```

**test**
`make test-unit`

### DIFF
--- a/pkg/common/routing/common.go
+++ b/pkg/common/routing/common.go
@@ -19,8 +19,7 @@ const (
 	// DefaultPoolGroup is the default InferencePool API group
 	DefaultPoolGroup = "inference.networking.k8s.io"
 
-	// LegacyPoolGroup is the legacy InferencePool API group
-	LegacyPoolGroup = "inference.networking.x-k8s.io"
+	LegacyK8SGroup = "inference.networking.x-k8s.io" // TODO: deprecated should be clean up
 
 	// KVConnectorNIXLV2 enables the P/D KV NIXL v2 protocol
 	KVConnectorNIXLV2 = "nixlv2"
--- a/pkg/common/routing/common.go
+++ b/pkg/common/routing/common.go
@@ -15,6 +15,24 @@ const (
 
 	// DataParallelEndpointHeader is the header name used to indicate the worker <ip:port> for Data Parallel
 	DataParallelEndpointHeader = "x-data-parallel-host-port"
+
+	// DefaultPoolGroup is the default InferencePool API group
+	DefaultPoolGroup = "inference.networking.k8s.io"
+
+	// LegacyPoolGroup is the legacy InferencePool API group
+	LegacyPoolGroup = "inference.networking.x-k8s.io"
+
+	// KVConnectorNIXLV2 enables the P/D KV NIXL v2 protocol
+	KVConnectorNIXLV2 = "nixlv2"
+
+	// KVConnectorSharedStorage enables the P/D KV Shared Storage protocol
+	KVConnectorSharedStorage = "shared-storage"
+
+	// KVConnectorSGLang enables SGLang the P/D KV disaggregation protocol
+	KVConnectorSGLang = "sglang"
+
+	// ECExampleConnector enables the Encoder disaggregation protocol (E/PD, E/P/D)
+	ECExampleConnector = "ec-example"
 )
 
 // StripScheme removes the scheme from an endpoint URL, returning host:port.
--- a/pkg/common/routing/common.go
+++ b/pkg/common/routing/common.go
@@ -18,8 +18,6 @@ const (
 
 	// InferencePoolAPIGroup is the default InferencePool API group
 	InferencePoolAPIGroup = "inference.networking.k8s.io"
-
-	LegacyK8SGroup = "inference.networking.x-k8s.io" // TODO: deprecated should be clean up
 )
 
 // StripScheme removes the scheme from an endpoint URL, returning host:port.
--- a/pkg/common/routing/common.go
+++ b/pkg/common/routing/common.go
@@ -16,22 +16,10 @@ const (
 	// DataParallelEndpointHeader is the header name used to indicate the worker <ip:port> for Data Parallel
 	DataParallelEndpointHeader = "x-data-parallel-host-port"
 
-	// DefaultPoolGroup is the default InferencePool API group
-	DefaultPoolGroup = "inference.networking.k8s.io"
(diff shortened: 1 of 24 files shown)


pkg/common/routing/common.go
```diff
@@ -15,6 +15,24 @@ const (
 
 	// DataParallelEndpointHeader is the header name used to indicate the worker <ip:port> for Data Parallel
 	DataParallelEndpointHeader = "x-data-parallel-host-port"
+
+	// DefaultPoolGroup is the default InferencePool API group
+	DefaultPoolGroup = "inference.networking.k8s.io"
+
+	// LegacyPoolGroup is the legacy InferencePool API group
+	LegacyPoolGroup = "inference.networking.x-k8s.io"
```
ahg-g: we should drop this, the api graduated five releases ago

pkg/epp/util/testing/wrappers.go
```diff
@@ -177,7 +178,7 @@ func MakeInferencePool(name string) *InferencePoolWrapper {
 				Name: name,
 			},
 			TypeMeta: metav1.TypeMeta{
-				APIVersion: "inference.networking.k8s.io/v1",
+				APIVersion: routing.DefaultPoolGroup + "/v1",
```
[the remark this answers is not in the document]
zdtsw: not a too big  issue for test

pkg/common/routing/common.go
zdtsw: agreed, but should we do the cleanup in https://github.com/llm-d/llm-d-router/issues/1298?

shmuelk: Its used also in the config loader code, InferenceObjective CRD, and InferenceModelRewrite CRD, where it was just deprecated this release.

Perhaps LegacyPoolGroup is a bad name. 

```suggestion
	LegacyK8SGroup = "inference.networking.x-k8s.io"
```

shmuelk: Shouldn't we have here a constant for the llmd.ai Group name?

zdtsw: updated
als for the llm-d.ai it is in the "apix" i updated code to point that one.
or we could add a `LLMDGroup = "llm-d.ai"` in common

zdtsw:
can i get this one merged?

pkg/common/routing/common.go
```diff
@@ -15,6 +15,23 @@ const (
 
 	// DataParallelEndpointHeader is the header name used to indicate the worker <ip:port> for Data Parallel
 	DataParallelEndpointHeader = "x-data-parallel-host-port"
+
+	// DefaultPoolGroup is the default InferencePool API group
+	DefaultPoolGroup = "inference.networking.k8s.io"
```
ahg-g:
```suggestion
	InferencePoolAPIGroup = "inference.networking.k8s.io"
```

pkg/sidecar/proxy/allowlist.go
```diff
@@ -43,8 +44,8 @@ const (
 
 // InferencePool API group to version mapping
 var inferencePoolGroupToVersion = map[string]string{
-	DefaultPoolGroup: "v1",
-	LegacyPoolGroup:  "v1alpha2",
+	routing.DefaultPoolGroup: "v1",
+	routing.LegacyK8SGroup:   "v1alpha2",
```
ahg-g: As I mentioned before, this api group for inferencePool is deprecated long ago, we shouldn't have this here.

pkg/common/routing/common.go
```diff
@@ -15,6 +15,23 @@ const (
...
+	KVConnectorNIXLV2 = "nixlv2"
+
+	// KVConnectorSharedStorage enables the P/D KV Shared Storage protocol
+	KVConnectorSharedStorage = "shared-storage"
+
+	// KVConnectorSGLang enables SGLang the P/D KV disaggregation protocol
+	KVConnectorSGLang = "sglang"
+
+	// ECExampleConnector enables the Encoder disaggregation protocol (E/PD, E/P/D)
+	ECExampleConnector = "ec-example"
```
ahg-g: none of the above four constants are truly shared between the epp and the proxy, it is shared between the proxy an the e2e test, and I think it is appropriate for the e2e test to import the proxy pkg to those constants definitions.

pkg/common/routing/common.go
```diff
@@ -15,6 +15,23 @@ const (
 
 	// DataParallelEndpointHeader is the header name used to indicate the worker <ip:port> for Data Parallel
 	DataParallelEndpointHeader = "x-data-parallel-host-port"
+
+	// DefaultPoolGroup is the default InferencePool API group
+	DefaultPoolGroup = "inference.networking.k8s.io"
+
+	LegacyK8SGroup = "inference.networking.x-k8s.io" // TODO: deprecated should be clean up
```
ahg-g: as mentioned below, this is not really shared, there are a few places in the epp where this is kept for backward compatibility, but for inferencePool this is long gone as an api group and is not even supported by the epp.

zdtsw: so just be clear my understanding

- we only move `intefence.networking.k8s.io` in common,
- remove `inference.networking.x-k8s.io `completely but that should be done in https://github.com/llm-d/llm-d-router/issues/1298
- keepall four connectors consts in proxy
Correct?

ahg-g: > we only move intefence.networking.k8s.io in common,

Yes.

> remove inference.networking.x-k8s.io completely but that should be done in

We can remove it from the proxy now, which means it is not shared anymore with the epp, and so we don't need to define it in the shared constants file.

> keepall four connectors consts in proxy

Yes.

zdtsw: i've updated (1) and (3) 
and dropped const on "inference.networking.x-k8s.io" , fall back to have  hardcoded value in allowlist

as for (2), the change to entirely remove "inference.networking.x-k8s.io" is much bigger than a bunch of cleanup code in sidecar/proxy and eppconfig, 
a new issue has been opened in kserve https://github.com/kserve/kserve/issues/5605 to work with https://github.com/llm-d/llm-d-router/issues/1298 to support any upgrade case in production.

PR #1330, opened 2026-05-26
ci: cache vllm-openai-cpu image across e2e run

### LINKED ISSUE #1328: ci: avoid re-pulling vllm/vllm-openai-cpu:v0.19.1 on every PR run
## Problem

Every PR run on `ci-pr-checks.yaml` re-downloads `vllm/vllm-openai-cpu:v0.19.1`
from Docker Hub before `make test-e2e`. The image is several hundred MB, so
this adds noticeable wall time and consumes Docker Hub pull bandwidth on every
push.

The pull happens via `make image-pull` → `scripts/pull_images.sh`, which is a
prerequisite of the `test-e2e` target (Makefile:292). Immediately before that,
the workflow runs `docker system prune -af` (ci-pr-checks.yaml:244-246), which
also wipes any image the runner might otherwise have retained.

No GitHub Actions cache currently covers this image. The existing
`actions/cache@v5` entries cover Go modules/build cache and the coverage
baseline only.

## Proposal

Cache the image tarball with `actions/cache@v5`, keyed on the image tag, and
load it before `make test-e2e`. Sketch:

```yaml
- name: Restore vllm image cache
  id: vllm-cache
  uses: actions/cache@v5
  with:
    path: ~/.cache/docker/vllm.tar
    key: docker-vllm-openai-cpu-v0.19.1

- name: Load cached vllm image
  if: steps.vllm-cache.outputs.cache-hit == 'true'
  run: docker load -i ~/.cache/docker/vllm.tar

- name: Pull and save vllm image
  if: steps.vllm-cache.outputs.cache-hit != 'true'
  run: |
    docker pull vllm/vllm-openai-cpu:v0.19.1
    mkdir -p ~/.cache/docker
    docker save vllm/vllm-openai-cpu:v0.19.1 -o ~/.cache/docker/vllm.tar

```

## Alternative fix

Mirror the image to GHCR to skip Docker Hub on every PR run, replacing the cache.

revit13:
**What type of PR is this?**
/kind test

## Summary

Cache the `vllm/vllm-openai-cpu` rendering image between e2e runs in the
`lint-and-test` job so PR CI does not re-pull a multi-hundred-MB image from
Docker Hub on every push.

The image is needed by `make test-e2e` via `make image-pull` →
`scripts/pull_images.sh`. Today every PR run downloads it fresh, then
`docker system prune -af` (added in #915 to free disk for kind) wipes any
copy the runner might have retained.

### Why we cache only `vllm-openai-cpu`

`make image-pull` also fetches `ghcr.io/llm-d/llm-d-inference-sim:v0.8.2`,
`ghcr.io/llm-d/llm-d-uds-tokenizer:dev`, and the EPP / sidecar images. Of
these, only `vllm-openai-cpu:v0.19.1` is worth caching:

- **`uds-tokenizer:dev` and the EPP / sidecar `dev` images** — `dev` is a
  moving tag. Caching by tag would serve stale bits; the EPP / sidecar are
  also rebuilt locally every run from the PR's source, so a cache could only
  ever be wrong.
- **`inference-sim:v0.8.2`** — pinned tag, so cacheable in principle, but
  pulled from GHCR (much higher rate limits than Docker Hub, faster from
  GitHub-hosted runners) and small (Go binary in a slim base). Wall-time
  savings would be seconds per run; not worth the extra workflow surface.
- **`vllm-openai-cpu:v0.19.1`** — pinned tag, pulled from Docker Hub
  (anonymous rate limits, slower mirror), and multi-hundred-MB. Highest
  payoff by a wide margin.

Fixes #1328

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -245,6 +245,45 @@ jobs:
         shell: bash
         run: docker system prune -af
 
+      # Cache the vllm rendering image across runs so e2e does not re-pull it
+      # from Docker Hub every time. Must run after `docker system prune -af`
+      # above, which would otherwise wipe the loaded image.
+      - name: Resolve VLLM_RENDER_IMAGE
+        id: vllm-image
+        shell: bash
+        run: |
+          set -euo pipefail
+          image=$(make -s env | awk -F= '/^VLLM_RENDER_IMAGE=/{print $2}')
+          [[ -n "$image" ]] || { echo "Could not resolve VLLM_RENDER_IMAGE from 'make env'"; exit 1; }
+          echo "image=$image" >> "$GITHUB_OUTPUT"
+
+      - name: Cache vllm rendering image
+        id: vllm-cache
+        uses: actions/cache@v5
+        with:
+          path: ${{ runner.temp }}/vllm-render-image.tar
+          key: docker-${{ steps.vllm-image.outputs.image }}
+
+      - name: Load or pull vllm rendering image
+        shell: bash
+        env:
+          VLLM_RENDER_IMAGE: ${{ steps.vllm-image.outputs.image }}
+          IMAGE_TAR: ${{ runner.temp }}/vllm-render-image.tar
+          CACHE_HIT: ${{ steps.vllm-cache.outputs.cache-hit }}
+        run: |
+          set -euo pipefail
+          if [[ "$CACHE_HIT" == "true" ]]; then
+            docker load -i "$IMAGE_TAR"
+            # Cache is already populated; drop the tarball to free disk for kind.
+            rm -f "$IMAGE_TAR"
+          else
+            docker pull "$VLLM_RENDER_IMAGE"
+            # Stage to .tmp so an interrupted save cannot publish a partial tar
+            # that would poison the cache for subsequent runs.
+            docker save "$VLLM_RENDER_IMAGE" -o "$IMAGE_TAR.tmp"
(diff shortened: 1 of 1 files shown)


elevran:
@revit13 thanks!
I think this PR is now obsolete:
  
The problem it solved was: `docker system prune -af` in the old monolithic `lint-and-test` job wiped the vllm image before it could be cached, forcing a fresh Docker Hub pull on every push. #1330 added a cache step keyed by image tag, positioned after the prune.
     
Both the job and the problem no longer exist in upstream/main:

1. `lint-and-test` is gone. #1315 and related split it into test (unit/integration) + e2e-image-* + e2e-router/e2e-gaie jobs.  so #1330 targets a job that no longer exists and would be a merge conflict.
2. `docker system prune -af` was removed in the same refactor, so there's nothing left to wipe the image between steps.
3. `vllm/vllm-openai-cpu:v0.21.0` is now pulled once in its own dedicated `e2e-image-router (renderer)` job, saved as a GHA artifact (retention-days: 1), and downloaded only by the extended e2e test shard that needs it.

The new structure achieves one Docker Hub pull per run via artifacts rather than GHA cache. The only thing #1330 still offered that the new workflow doesn't is cross-run caching (so v0.21.0 isn't re-pulled daily, only when the tag changes). That optimization would need to be re-implemented as a cache step in e2e-image-router (renderer), not in the old job. The PR as written cannot be applied.

Not sure the work is worth it to avoid the daily pull. Please **close #1328** if you do not intend to implement the additional optimization step.

Thoughts?

CC @roytman

Issue #1340, opened 2026-05-26
feat(sidecar): passthrough non-generation paths (/v1/rerank, /v1/embeddings) without completions validation


magicsong:
## Problem

The routing sidecar (`pd_sidecar` / `pkg/sidecar`, previously `llm-d-routing-sidecar:v0.8.0`) validates **every** inbound request as a completions request by requiring a `prompt` field in the request body. Non-completion endpoints — `/v1/rerank`, `/v1/embeddings`, `/v1/score` — do not carry a `prompt` field, so the sidecar rejects them with:

```
400 Bad Request
{"object":"error","message":"inference error: BadRequest - invalid completions request: must have prompt field"}
```

## Deployment context

- Sidecar args: `--port=8000 --vllm-port=8031 --connector=nixlv2 --zap-log-level=debug`
- Backend: vLLM `0.19.0` serving `Qwen3-Reranker-0.6B` on port 8031
- Endpoint used: `POST /v1/rerank`
- Kubernetes InferencePool with EPP, deployed via llm-d

## Steps to reproduce

1. Deploy any vLLM rerank or embedding model behind the llm-d sidecar.
2. Send a rerank request through the sidecar:
   ```bash
   curl -X POST http://<pod>:8000/v1/rerank \
     -H 'Content-Type: application/json' \
     -d '{"model":"my-model","query":"hello","documents":["doc1","doc2"]}'
   ```
3. Sidecar returns `400: must have prompt field`.
4. Direct call to vLLM `http://<pod>:8031/v1/rerank` returns `200 OK` correctly.

## Expected behaviour

Requests to non-generation paths (`/v1/rerank`, `/v1/embeddings`, `/v1/score`, etc.) should be **passed through** to the backend without completions-format validation. KV-cache-aware scheduling is only meaningful for prefill/decode generation requests; for rerank and embedding requests, a direct proxy is the correct and sufficient behaviour.

## Suggested fix

In the sidecar request handler, check the request path before running completions validation. If the path is not a generation endpoint, skip validation and forward directly:

```go
// pkg/sidecar or cmd/pd_sidecar handler
if !isGenerationPath(r.URL.Path) {
    // passthrough: proxy directly to vLLM, no KV-cache scheduling needed
    s.reverseProxy.ServeHTTP(w, r)
    return
}
// existing: validate prompt field, KV-cache-aware routing ...
```

[...]

PR #1341, opened 2026-05-26
proxy: document and test /v1/rerank passthrough to decoder

magicsong:
## Summary

This PR adds explicit documentation, a named constant, and regression tests for the `/v1/rerank` passthrough behaviour in the PD-sidecar.

### Background

When deploying a reranker model (e.g. `Qwen3-Reranker-0.6B`) behind an llm-d `InferencePool`, rerank requests sent to the sidecar at `/v1/rerank` received a `400 BadRequest` error with the message:

```
inference error: BadRequest - invalid completions request: must have prompt field
```

This was observed with the older `llm-d-routing-sidecar:v0.8.0` image which validated **all** incoming request bodies as completions requests requiring a `prompt` field. The current `llm-d-router` code already handles this correctly — `/v1/rerank` falls through to the `decoderProxy` catch-all in `createRoutes()`. However this behaviour was implicit and untested.

Tracked in: #1340

### Changes

- **`pkg/sidecar/proxy/chat_completions.go`**: Add `RerankPath = "/v1/rerank"` constant alongside `ChatCompletionsPath`, `CompletionsPath`, and `ResponsesPath`, making it clear that this path is a known, explicitly supported non-generation endpoint.

- **`pkg/sidecar/proxy/proxy.go`**: Add a comment in `createRoutes()` documenting that the catch-all `mux.Handle("/", s.decoderProxy)` intentionally handles non-generation paths (rerank, embeddings, etc.) as direct decoder pass-throughs without KV-cache scheduling.

- **`pkg/sidecar/proxy/proxy_test.go`**: Add two new `Entry` cases to the existing `"should forward requests to decode server"` table test, covering `POST /v1/rerank` with both plain HTTP and TLS proxy modes.

### Why this is the right fix

KV-cache-aware scheduling (disaggregated prefill) is only meaningful for generation endpoints that produce tokens. Rerankers and embedding models produce scores/vectors, not tokens, and must not be routed through the prefill/decode disaggregation path. The catch-all passthrough is correct; this PR makes the intent explicit and tested.

### DIFF
--- a/pkg/sidecar/proxy/chat_completions.go
+++ b/pkg/sidecar/proxy/chat_completions.go
@@ -44,6 +44,11 @@ var (
 
 	// ResponsesPath is the OpenAI Responses API path
 	ResponsesPath = "/v1/responses"
+
+	// RerankPath is the OpenAI-compatible rerank path used by vLLM and compatible servers.
+	// Rerank requests do not require KV-cache-aware scheduling and are proxied directly
+	// to the decoder without modification.
+	RerankPath = "/v1/rerank"
 )
 
 func openAIAPIAttr(apiType APIType) attribute.KeyValue {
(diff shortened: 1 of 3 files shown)


magicsong:
Closing this PR — it was filed against the wrong component (PD sidecar `pkg/sidecar/proxy/`). The actual bug is in the EPP framework's OpenAI parser at `pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go`.

The correct fix is in PR #1361.

PR #1350, opened 2026-05-26
Drop top-level 'tokens' field from body before forwarding request to vLLM

### LINKED ISSUE #1367: [Feature] Support extensible non-standard fields in inference requests for EPP processing
In disaggregated serving topologies involving the Coordination service, we would like to optimize EPP processing by passing additional metadata alongside standard OpenAI-compatible request payloads. This metadata is consumed by EPP during scheduling and routing, and must be stripped from the request before it is forwarded to the backend inference engine.

For example, to prevent redundant tokenization of incoming `v1/chat/completions` requests, we would like to pass pre-computed token data directly alongside standard OpenAI payload fields.

```yaml
{
  "model": "llava-v1.5-7b",
  "stream": false,
  "messages": [
    {
      "role": "user",
      "content": [
        {"type": "text", "text": "Describe these images"},
        {
          "type": "image_url",
          "image_url": {"url": "data:image/jpeg;base64,/9j/4AAQ..."}
        },
        {
          "type": "image_url",
          "image_url": {"url": "data:image/png;base64,iVBORw0K..."}
        }
      ]
    }
  ],
  "tokens": {
    "token_ids": [1, 32000, 32000, 32000, 32000, 32000, 32000, 2345, 6789],
    "features": {
      "mm_hashes": {"image": ["abc123hash", "def456hash"]},
      "mm_placeholders": {"image": [
        {"offset": 1, "length": 3},
        {"offset": 4, "length": 3}
      ]}
    }
  },
  "ec_transfer_params": {...}
  "kv_transfer_params": {...},
  "sampling_params": {"max_tokens": 1}
}
```

The tokens field on chat/completions requests is one example of this pattern. However, this is only the first such field, and we expect more to follow across multiple request types (chat/completions, completions, etc.).

We want to establish a general mechanism where:

- The Coordination service can attach extra, non-standard fields to any supported request type.
- EPP reads and acts on those fields (e.g., for KV cache affinity, prefix matching, scheduling hints).
- EPP strips the fields before forwarding the request downstream, so the backend (e.g., vLLM) never sees them.

The `tokens` field (carrying token_ids, multimodal hashes, and placeholders) is the initial concrete use case, but the design should be extensible to additional fields and request types going forward.

**Why is this needed**:
With this addition, it will be possible to optimize EPP processing; for example, by providing tokens, EPP is able to utilize them directly instead of performing a redundant prompt tokenization step itself.

Non-standard fields are not part of the OpenAI schema. If passed through, backends like vLLM print warnings and drop them, which is benign but noisy and wasteful in terms of payload size. EPP is the correct layer to consume and remove these fields, keeping the downstream request clean and conformant.

dmitripikus:
## Summary:

A top-level `tokens` block on `/v1/chat/completions` requests (`token_ids` / `mm_hashes` / `mm_placeholders`) is not part of the OpenAI API. When it presents, vLLM wars about this and proceeds. 

`tokens` key is dropped from the `PayloadMap` in the parser (after it's stored for routing purposes), so the field is gone by
the time `repackage` re-marshals the body for forwarding.

Fixes [#1367](https://github.com/llm-d/llm-d-router/issues/1367)

### DIFF
--- a/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
@@ -48,6 +48,7 @@ const (
 	// to account for optional parameters like "; charset=utf-8" often appended by proxies.
 	eventStreamType = "text/event-stream"
 
+	tokensField              = "tokens"
 	usageField               = "usage"
 	promptTokensField        = "prompt_tokens"
 	inputTokensField         = "input_tokens"
@@ -106,7 +107,7 @@ func (p *OpenAIParser) ParseRequest(ctx context.Context, body []byte, headers ma
 	if err != nil {
 		return nil, err
 	}
-	delete(bodyMap, "tokens")
+	delete(bodyMap, tokensField)
 	extractedBody.Payload = fwkrh.PayloadMap(bodyMap)
 	if stream, ok := bodyMap["stream"].(bool); ok && stream {
 		extractedBody.Stream = true
--- a/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
@@ -106,6 +106,7 @@ func (p *OpenAIParser) ParseRequest(ctx context.Context, body []byte, headers ma
 	if err != nil {
 		return nil, err
 	}
+	delete(bodyMap, "tokens")
 	extractedBody.Payload = fwkrh.PayloadMap(bodyMap)
 	if stream, ok := bodyMap["stream"].(bool); ok && stream {
 		extractedBody.Stream = true
--- a/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai_test.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai_test.go
@@ -173,7 +173,7 @@ func TestOpenAIParser_ParseRequest(t *testing.T) {
 				"tokens": map[string]any{
 					"token_ids": []any{1, 2, 3},
 					"features": map[string]any{
-						"mm_hashes":        map[string]any{"image": []any{"abc"}},
+						"mm_hashes":       map[string]any{"image": []any{"abc"}},
 						"mm_placeholders": map[string]any{"image": []any{map[string]any{"offset": 1, "length": 3}}},
 					},
 				},
--- a/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai_test.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai_test.go
@@ -163,6 +163,54 @@ func TestOpenAIParser_ParseRequest(t *testing.T) {
 			},
 		},
 		{
+			name:    "chat completions request body strips top-level tokens field",
+			headers: map[string]string{":path": "/v1/chat/completions"},
+			body: map[string]any{
+				"model": "test",
+				"messages": []any{
+					map[string]any{"role": "user", "content": "hello"},
+				},
+				"tokens": map[string]any{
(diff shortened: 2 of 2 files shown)


pkg/epp/framework/plugins/requesthandling/parsers/openai/openai_test.go
```diff
@@ -162,6 +162,54 @@ func TestOpenAIParser_ParseRequest(t *testing.T) {
...
+				Payload: fwkrh.PayloadMap{
+					"model": "test",
+					"messages": []any{
+						map[string]any{"role": "user", "content": "hello"},
+					},
+				},
+			},
+		},
+		{
+			name:    "completions request body strips top-level tokens field",
```
roytman: We don't need this test.
Completions request can contain prompt as a token_ids array, and cannot include multimedia items.

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
```diff
@@ -106,6 +106,7 @@ func (p *OpenAIParser) ParseRequest(ctx context.Context, body []byte, headers ma
 	if err != nil {
 		return nil, err
 	}
+	delete(bodyMap, "tokens")
```
roytman: It happens **before** bodyMap is assigned to extractedBody.Payload. This means the tokens are gone from the PayloadMap entirely. If a downstream routing plugin (e.g., prefix-based PD decider) wants to use pre-computed token_ids for routing decisions, it won't have access to them.
Please check with @vMaroon , @Mohamedma96 on where to store the tokens for their usage

pkg/epp/requestcontrol/director.go
```diff
@@ -314,6 +314,7 @@ func (d *Director) repackage(ctx context.Context, reqCtx *handlers.RequestContex
 	default:
 		return errcommon.Error{Code: errcommon.BadRequest, Msg: "Unsupported llmRequest parsedBody"}
 	}
+	logger.V(logutil.DEBUG).Info("Request body forwarded to vLLM", "body", string(reqCtx.Request.RawBody))
```
roytman: This will log full request bodies at debug level, which can include:

- Large prompts (potentially hundreds of KB)
- Base64-encoded images in multimodal requests
- Potentially sensitive user content
- Plus, it prints for ALL request types

let me check what can be done instead

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
roytman: nit: Minor, but the parser file already defines named constants for field keys. Consider adding:
```golang
const tokensField = "tokens"
```
for consistency with the existing pattern (usageField, promptTokensField, etc.).

pkg/epp/requestcontrol/director.go
roytman: Actually, you can remove the line, we can validate that the `tokens` field is not propagated to vllm by checking its logs.

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai_test.go
dmitripikus: Fixed. Thanks!

pkg/epp/requestcontrol/director.go
dmitripikus: Done. Thanks!

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
dmitripikus: Done. Thanks!

ahg-g:
@roytman @dmitripikus pls feel free to re-open if you think this is still relevant.

PR #1351, opened 2026-05-26
add per-request attribute store to InferenceRequest

### LINKED ISSUE #1336: [feat] Per request attributes, with DAG validation
Ability to add and retrieve per request attributes with datalayer DAG validation

**Why is this needed**:
We have three storage tiers for attributes, but only one is DAG-validated.

1. Per-endpoint `AttributeMap`. This is the only tier covered by the data-layer Produces/Consumes DAG. Type compatibility, layer ordering (FlowControl, RequestControl and thenScheduling), acyclicity, and default-producer auto-instantiation are all enforced here.
2. `scheduling.CycleState` is per cycle, flat map of key to data. ~Threaded through plugin signatures but not declared in Produces/Consumes and not validated. This is deprecated and has no production use today.~ Removed in #1335.
3. `plugin.PluginState` is per-request, keyed by requestID (with TTL and evication manaement). Also not part of the DAG.

Endpoint specific attributes which are independent of any request (e.g., metrics, models being served) are handled by `Endpoint.AttributeMap`.
Per Endpoint and Request attributes (e.g., KV cache coverage depends on request prompt and endpoint cache) can use the same (the difference being that the former can be set, by using Extractors for example, on the endpoint directly **inside the datastore**, whereas the latter are done on the `DataProducer` after the datastore snapshot is taken).
There is no natural place for request specific attributes (adding to InferenceRequest is problematic in the long run and does not allow DAG validation).

For request-only attributes (session id, agent id, parsed token bag): the most sensible mechanisms today would be a `DataProducer` that
- duplicates the same attribute for all endpoints; or
- using `PluginState`, keyed by `request.RequestID`, then read by downstream plugins via `ReadPluginStateKey[T]`. 

Both provide DAG validation but neither feels correct for request specific values used for scheduling (i.e., `PluginState` should not be used to communicate information across plugins).
Per request information belongs on the request itself.

elevran:
**What type of PR is this?**
/kind feature

Adds a request-scoped attribute map on `InferenceRequest` so producers and consumers can flow per-request data through the existing Producer/Consumer DAG, alongside the per-endpoint `AttributeMap`. Today there is no natural home for request-only attributes (session id, agent id, parsed token bag): producers have to duplicate the value across every endpoint or abuse `PluginState` as a cross-plugin handoff channel. The DAG keys off `DataKey` types regardless of where the producer stores the data, so storage location is a producer-side choice and intentionally not validated.

The store uses `any` rather than `Cloneable`. `Endpoint.AttributeMap` requires `Cloneable` because endpoints are snapshotted; `InferenceRequest` is not cloned, so the constraint would be a tax with no benefit. Type-safe access is via the generic `ReadRequestAttribute[T]` helper. Sequential per-request access is the documented contract, matching how data producers actually run.

Also tightens the `PluginState` doc comment to say it is per-plugin scratch keyed by request id and not a mechanism for sharing data across plugins; cross-plugin data flow should go through the DAG.

Fixes #1336

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/interface/plugin/plugin_state.go
+++ b/pkg/epp/framework/interface/plugin/plugin_state.go
@@ -43,10 +43,15 @@ func NewPluginState(ctx context.Context) *PluginState {
 	return pluginState
 }
 
-// PluginState provides a mechanism for plugins to store and retrieve arbitrary data by multiple extension points.
-// Data stored by the plugin in one extension point can be written, read or altered by another extension point.
-// The data stored in PluginState is always stored in the context of a given request.
-// If the data hasn't been accessed during "stalenessThreshold", it is cleaned by a cleanup internal mechanism.
+// PluginState is per-plugin scratch storage scoped to a single request. A plugin's
+// extension points (e.g. PreRequest, ResponseBody) can write, read, and alter entries
+// here to coordinate within that plugin. Entries are keyed by RequestID and reaped
+// after "stalenessThreshold" of inactivity.
+//
+// PluginState is not a cross-plugin handoff channel. Data shared between plugins must
+// flow through the Producer/Consumer DAG: write to Endpoint AttributeMap for
+// per-endpoint data, or to the InferenceRequest attribute store for per-request data.
+// The DAG validates type compatibility and execution ordering; PluginState does not.
 //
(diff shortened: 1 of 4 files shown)


ahg-g:
/approve

Nice!

PR #1354, opened 2026-05-26
flowcontrol: move priority band provisioning off request hot path

### LINKED ISSUE #1184: [Flow Control] Move Priority Band Provisioning from Hot-Path JIT to InferenceObjective Reconciler
Flow control currently uses a Just-In-Time (JIT) provisioning model. When a request arrives with an unseen priority level, the registry dynamically creates the priority band configuration, instantiates policy components, registers stats, and syncs the dispatch shards *on the hot request path* (`WithConnection` -> `ensureFlowInfrastructure` -> `ensurePriorityBand` in `pkg/epp/flowcontrol/registry/registry.go`).

This JIT model requires acquiring exclusive write locks (`fr.mu.Lock()`) on the hot data path, introducing performance latency and potential lock contention under high concurrency.

Since the EPP server and the `InferenceObjectiveReconciler` (which manages priority traffic classes) execute within the same process/binary (initialized in `cmd/epp/runner/runner.go`), we will move the responsibility of provisioning Priority Bands out of the data path and into the control plane (CRD reconciliation loop).

gyliu513:
**What type of PR is this?**

/kind bug

Fixed issue #1184 using the channel-based design agreed in the issue discussion @shmuelk @LukeAVanDrie

- **Control plane**: `InferenceObjectiveReconciler` collects desired priority levels from the datastore on each reconcile and submits them via non-blocking `SubmitDesiredPriorities` (sync fallback when the processor is not yet consuming).
- **Processor loop**: `ShardProcessor` applies topology updates and runs registry GC from its main `select`; the standalone `FlowRegistry.Run` GC goroutine is no longer started.
- **Data plane**: `ensureFlowInfrastructure` no longer JIT-provisions priority bands on the request path; missing bands return `ErrPriorityBandNotFound` (flow queues are still created on first use).
- **Fallback**: `FlowController` demotes requests to priority 0 when the requested band is not yet provisioned (availability-first).

Fixes #1184 

```release-note
NONE
```

### DIFF
--- a/pkg/epp/flowcontrol/benchmark/benchmark.go
+++ b/pkg/epp/flowcontrol/benchmark/benchmark.go
@@ -214,7 +214,7 @@ func setupRegistry(
 	}
 
 	reg := registry.NewFlowRegistry(regCfg, logr.Discard())
-	// Registry maintenance (GC, priority band sync) runs in the ShardProcessor loop started by FlowController.
+	// Registry maintenance (GC, priority band sync) runs in the Processor loop started by FlowController.
 	return reg
 }
 
--- a/pkg/epp/flowcontrol/benchmark/benchmark.go
+++ b/pkg/epp/flowcontrol/benchmark/benchmark.go
@@ -187,7 +187,6 @@ func (r *benchRequest) ReceivedTimestamp() time.Time                   { return 
 
 // setupRegistry provisions the concrete FlowRegistry.
 func setupRegistry(
-	ctx context.Context,
 	b *testing.B,
 	defaults registry.PriorityBandPolicyDefaults,
 	p priorityLevels,
@@ -215,8 +214,7 @@ func setupRegistry(
 	}
 
 	reg := registry.NewFlowRegistry(regCfg, logr.Discard())
-
-	go reg.Run(ctx)
+	// Registry maintenance (GC, priority band sync) runs in the ShardProcessor loop started by FlowController.
 	return reg
 }
 
@@ -249,7 +247,7 @@ func setupBenchmarkHarness(
 		FairnessPolicy: fPolicy.(flowcontrol.FairnessPolicy),
 	}
 
-	reg := setupRegistry(ctx, b, defaults, p)
+	reg := setupRegistry(b, defaults, p)
 
 	detector := customDetector
 	if detector == nil {
--- a/pkg/epp/flowcontrol/contracts/registry.go
+++ b/pkg/epp/flowcontrol/contracts/registry.go
@@ -55,7 +55,7 @@ type PriorityBandControlPlane interface {
 	SubmitDesiredPriorities(desired map[int]struct{})
 }
 
-// FlowRegistryBackground exposes hooks consumed by the ShardProcessor maintenance loop.
+// FlowRegistryBackground exposes hooks consumed by the Processor maintenance loop.
 type FlowRegistryBackground interface {
 	PriorityBandUpdateChannel() <-chan map[int]struct{}
 	FlowGCTimeout() time.Duration
--- a/pkg/epp/flowcontrol/contracts/registry.go
+++ b/pkg/epp/flowcontrol/contracts/registry.go
@@ -17,6 +17,8 @@ limitations under the License.
 package contracts
 
 import (
+	"time"
+
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/flowcontrol"
 )
 
@@ -31,16 +33,34 @@ import (
 // A flow instance, identified by its immutable FlowKey, has a lease-based lifecycle managed by this interface.
 // Any implementation MUST adhere to this lifecycle:
 //
-//  1. Lease Acquisition: A client calls Connect to acquire a lease. This signals that the flow is in use and protects
-//     it from garbage collection. If the flow does not exist, it is created Just-In-Time (JIT).
+//  1. Lease Acquisition: A client calls WithConnection to acquire a lease. This signals that the flow is in use and
+//     protects it from garbage collection. If the flow instance does not exist, it is created on first use.
+//     Priority bands must be provisioned by the control plane before use; they are not created on the request path.
 //  2. Active State: A flow is "Active" as long as its lease count is greater than zero.
-//  3. Lease Release: The client MUST call `Close()` on the returned `FlowConnection` to release the lease.
+//  3. Lease Release: When the WithConnection callback returns, the lease is released.
 //     When the lease count drops to zero, the flow becomes "Idle".
 //  4. Garbage Collection: The implementation MUST automatically garbage collect a flow after it has remained
 //     continuously Idle for a configurable duration.
 type FlowRegistry interface {
 	FlowRegistryObserver
 	FlowRegistryDataPlane
+	PriorityBandControlPlane
+	FlowRegistryBackground
+}
+
+// PriorityBandControlPlane submits priority band topology updates from the control plane.
+type PriorityBandControlPlane interface {
+	// SubmitDesiredPriorities queues the set of priority levels that should remain provisioned.
+	// Dynamic bands not in this set become eligible for removal once idle.
+	SubmitDesiredPriorities(desired map[int]struct{})
+}
+
+// FlowRegistryBackground exposes hooks consumed by the ShardProcessor maintenance loop.
+type FlowRegistryBackground interface {
+	PriorityBandUpdateChannel() <-chan map[int]struct{}
+	FlowGCTimeout() time.Duration
+	ApplyDesiredPriorities(desired map[int]struct{})
+	ExecuteGCCycle()
 }
 
 // FlowRegistryObserver defines the read-only, observation interface for the registry.
@@ -55,8 +75,8 @@ type FlowRegistryDataPlane interface {
 	// It is the primary and sole entry point for interacting with the data path.
 	//
 	// This method handles the entire lifecycle of a flow connection:
-	// 1. Just-In-Time (JIT) Registration: If the flow for the given FlowKey does not exist, it is created and registered
-	//    automatically.
+	// 1. Flow Registration: If the flow for the given FlowKey does not exist, it is created and registered
+	//    automatically. Priority bands must be provisioned by the control plane before use.
 	// 2. Lease Acquisition: It acquires a lifecycle lease, protecting the flow from garbage collection.
 	// 3. Callback Execution: It invokes the provided function `fn`, passing in a temporary `ActiveFlowConnection` handle.
 	// 4. Guaranteed Lease Release: It ensures the lease is safely released when the callback function returns.
--- a/pkg/epp/flowcontrol/controller/controller.go
+++ b/pkg/epp/flowcontrol/controller/controller.go
@@ -83,17 +83,17 @@ var _ processor = &internal.Processor{}
 type FlowController struct {
 	// --- Immutable dependencies (set at construction) ---
 
-	config                *Config
-	registry              registryClient
-	flowRegistry          contracts.FlowRegistry
-	registryBackground    contracts.FlowRegistryBackground
-	saturationDetector    flowcontrol.SaturationDetector
-	endpointCandidates    contracts.EndpointCandidates
-	usageLimitPolicy      flowcontrol.UsageLimitPolicy
-	clock                 clock.WithTicker
-	logger                logr.Logger
-	processorFactory      processorFactory
-	processor             processor
+	config             *Config
+	registry           registryClient
+	flowRegistry       contracts.FlowRegistry
+	registryBackground contracts.FlowRegistryBackground
+	saturationDetector flowcontrol.SaturationDetector
+	endpointCandidates contracts.EndpointCandidates
+	usageLimitPolicy   flowcontrol.UsageLimitPolicy
+	clock              clock.WithTicker
+	logger             logr.Logger
+	processorFactory   processorFactory
+	processor          processor
 
 	// --- Lifecycle state ---
 
--- a/pkg/epp/flowcontrol/controller/controller.go
+++ b/pkg/epp/flowcontrol/controller/controller.go
@@ -57,6 +57,7 @@ type processor interface {
 type processorFactory func(
 	ctx context.Context,
 	registry contracts.FlowRegistry,
+	registryBackground contracts.FlowRegistryBackground,
 	saturationDetector flowcontrol.SaturationDetector,
 	endpointCandidates contracts.EndpointCandidates,
 	usageLimitPolicy flowcontrol.UsageLimitPolicy,
@@ -82,15 +83,17 @@ var _ processor = &internal.Processor{}
 type FlowController struct {
 	// --- Immutable dependencies (set at construction) ---
 
-	config             *Config
-	registry           registryClient
-	saturationDetector flowcontrol.SaturationDetector
-	endpointCandidates contracts.EndpointCandidates
-	usageLimitPolicy   flowcontrol.UsageLimitPolicy
-	clock              clock.WithTicker
-	logger             logr.Logger
-	processorFactory   processorFactory
-	processor          processor
+	config                *Config
+	registry              registryClient
+	flowRegistry          contracts.FlowRegistry
+	registryBackground    contracts.FlowRegistryBackground
+	saturationDetector    flowcontrol.SaturationDetector
+	endpointCandidates    contracts.EndpointCandidates
+	usageLimitPolicy      flowcontrol.UsageLimitPolicy
+	clock                 clock.WithTicker
+	logger                logr.Logger
+	processorFactory      processorFactory
+	processor             processor
 
 	// --- Lifecycle state ---
 
@@ -120,9 +123,15 @@ func NewFlowController(
 	if deps.Clock == nil {
 		deps.Clock = clock.RealClock{}
 	}
+	var registryBackground contracts.FlowRegistryBackground
+	if bg, ok := deps.Registry.(contracts.FlowRegistryBackground); ok {
+		registryBackground = bg
+	}
 	fc := &FlowController{
 		config:             config,
 		registry:           deps.Registry,
+		flowRegistry:       deps.Registry,
+		registryBackground: registryBackground,
 		saturationDetector: deps.SaturationDetector,
 		endpointCandidates: deps.EndpointCandidates,
 		usageLimitPolicy:   deps.UsageLimitPolicy,
@@ -135,6 +144,7 @@ func NewFlowController(
 		fc.processorFactory = func(
 			ctx context.Context,
 			registry contracts.FlowRegistry,
+			registryBackground contracts.FlowRegistryBackground,
 			saturationDetector flowcontrol.SaturationDetector,
 			endpointCandidates contracts.EndpointCandidates,
 			usageLimitPolicy flowcontrol.UsageLimitPolicy,
@@ -147,6 +157,7 @@ func NewFlowController(
 				ctx,
 				poolName,
 				registry,
+				registryBackground,
 				saturationDetector,
 				endpointCandidates,
 				usageLimitPolicy,
@@ -163,7 +174,8 @@ func NewFlowController(
 	// Construct a new worker, but do not start its goroutine yet.
 	fc.processor = fc.processorFactory(
 		fc.parentCtx,
-		fc.registry,
+		fc.flowRegistry,
+		fc.registryBackground,
 		fc.saturationDetector,
 		fc.endpointCandidates,
 		fc.usageLimitPolicy,
@@ -227,7 +239,7 @@ func (fc *FlowController) EnqueueAndWait(
 
 	// 2. Acquire a lease for the Flow.
 	// We hold this lease for the entire duration of the request (Distribution + Queueing).
-	err := fc.registry.WithConnection(flowKey, func(conn contracts.ActiveFlowConnection) error {
+	err := fc.withConnectionWithDemotion(flowKey, func(conn contracts.ActiveFlowConnection) error {
 
 		select { // Non-blocking check on controller lifecycle.
 		case <-fc.parentCtx.Done():
@@ -264,6 +276,26 @@ func (fc *FlowController) EnqueueAndWait(
 	}
 
 	return finalOutcome, err
+}
+
+// withConnectionWithDemotion acquires a flow connection, demoting to priority 0 when the requested band is not yet provisioned.
+func (fc *FlowController) withConnectionWithDemotion(
+	key flowcontrol.FlowKey,
+	fn func(conn contracts.ActiveFlowConnection) error,
+) error {
+	err := fc.registry.WithConnection(key, fn)
+	if err == nil || !errors.Is(err, contracts.ErrPriorityBandNotFound) || key.Priority == 0 {
+		return err
+	}
+
+	fc.logger.V(logutil.DEFAULT).Info(
+		"Priority band not provisioned, demoting request to priority 0",
+		"originalPriority", key.Priority,
+		"flowID", key.ID,
+	)
+	demotedKey := key
+	demotedKey.Priority = 0
+	return fc.registry.WithConnection(demotedKey, fn)
 }
 
 // tryDistribution handles a single attempt to select a shard and submit a request.
--- a/pkg/epp/flowcontrol/registry/registry.go
+++ b/pkg/epp/flowcontrol/registry/registry.go
@@ -117,7 +117,8 @@ type FlowRegistry struct {
 	// It is updated dynamically when new bands are provisioned.
 	orderedPriorityLevels []int
 
-	// initialPriorities tracks priority bands provisioned at startup. These are never removed by control-plane sync.
+	// initialPriorities tracks priority bands provisioned at startup.
+	// These are never removed by control-plane sync or garbage collection.
 	initialPriorities map[int]struct{}
 
 	// priorityBandUpdateCh carries desired priority topology updates from the control plane to the processor loop.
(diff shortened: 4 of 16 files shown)


pkg/epp/flowcontrol/benchmark/benchmark.go
```diff
@@ -216,8 +215,7 @@ func setupRegistry(
 	}
 
 	reg := registry.NewFlowRegistry(regCfg, logr.Discard())
-
-	go reg.Run(ctx)
+	// Registry maintenance (GC, priority band sync) runs in the ShardProcessor loop started by FlowController.
```
shmuelk:
```suggestion
	// Registry maintenance (GC, priority band sync) runs in the Processor loop started by FlowController.
```

pkg/epp/flowcontrol/contracts/registry.go
```diff
@@ -49,6 +52,23 @@ import (
...
+}
+
+// PriorityBandControlPlane submits priority band topology updates from the control plane.
+type PriorityBandControlPlane interface {
+	// SubmitDesiredPriorities queues the set of priority levels that should remain provisioned.
+	// Dynamic bands not in this set become eligible for removal once idle.
+	SubmitDesiredPriorities(desired map[int]struct{})
+}
+
+// FlowRegistryBackground exposes hooks consumed by the ShardProcessor maintenance loop.
```
shmuelk:
```suggestion
// FlowRegistryBackground exposes hooks consumed by the Processor maintenance loop.
```

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -156,24 +170,101 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
...
+			fr.ExecuteGCCycle()
+		}
+	}
+}
+
+// SubmitDesiredPriorities queues a priority band topology update for the processor loop.
+// The call never blocks: stale pending updates are dropped and the latest desired state is retained.
+func (fr *FlowRegistry) SubmitDesiredPriorities(desired map[int]struct{}) {
+	if desired == nil {
+		desired = map[int]struct{}{}
```
shmuelk, line 197: If you got a nil value, don't do anything...

```suggestion
		return
```

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -156,24 +170,101 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
...
+	}
+	copy := make(map[int]struct{}, len(desired))
+	for priority := range desired {
+		copy[priority] = struct{}{}
+	}
```
shmuelk, line 202: Why are you copying the input instead of just writing to the channel?

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -156,24 +170,101 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
...
+
+	select {
+	case <-fr.priorityBandUpdateCh:
+	default:
+	}
```
shmuelk, line 208: Why are you draining the channel? Just let the old ones process.

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -156,24 +170,101 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
...
+
+	select {
+	case fr.priorityBandUpdateCh <- copy:
+	default:
+		fr.ApplyDesiredPriorities(copy)
+	}
```
shmuelk: Why not just block if the channel is full? The reconciler should already be queueing it's "changes" received from K8S.

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -156,24 +170,101 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
...
+// FlowGCTimeout returns the interval between registry garbage collection cycles.
+func (fr *FlowRegistry) FlowGCTimeout() time.Duration {
+	return fr.config.FlowGCTimeout
+}
+
+// ApplyDesiredPriorities provisions missing priority bands and removes idle bands no longer desired.
+// It is invoked by the ShardProcessor maintenance loop.
+func (fr *FlowRegistry) ApplyDesiredPriorities(desired map[int]struct{}) {
+	if desired == nil {
+		desired = map[int]struct{}{}
```
shmuelk, line 228:
```suggestion
		return
```

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -156,24 +170,101 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
...
+	}
+
+	fr.mu.Lock()
```
shmuelk, line 231: This lock is needed only because you might call this function from `SubmitDesiredPriorities`.

`SubmitDesiredPriorities` should NOT call this function, making the mutex lock unnecessary.

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -156,24 +170,101 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
...
+			continue
+		}
+		if _, wanted := desired[priority]; wanted {
+			continue
 		}
+		if fr.isPriorityBandIdle(priority) {
+			toRemove = append(toRemove, priority)
+		}
+	}
+	fr.mu.Unlock()
```
shmuelk: Ditto

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -156,24 +170,101 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
...
+
+	for _, priority := range toRemove {
+		fr.deletePriorityBand(priority)
 	}
 }
 
+// ExecuteGCCycle runs a single registry garbage collection cycle.
+func (fr *FlowRegistry) ExecuteGCCycle() {
+	fr.executeGCCycle()
+}
```
shmuelk, line 267: Why?

Just make the original `executeGCCycle` public

pkg/epp/flowcontrol/registry/registry.go
LukeAVanDrie: I think the lock is still necessary, and its scope must actually be broadened to fix a TOCTOU race.

Even if updates run solely on the processor thread, we still need the write lock. `ApplyDesiredPriorities` writes to `fr.config.PriorityBands`, which is concurrently read by requests on the hot path (`WithConnection` -> `ensureFlowInfrastructure` under RLock).

Unlocking before deletion introduces a race. If we unlock after the idle check:

1. A concurrent request can arrive, pin the band (`leaseCount` becomes 1), and proceed.
2. This thread then calls `deletePriorityBand` &rarr;  logically deletes it from `priorityBandStates` map &rarr;  calls `cleanupPriorityBandResources` which locks `fr.mu.Lock()`.
3. Since it was logically deleted, it proceeds to physically delete it from `fr.config.PriorityBands`.

The active request is then orphaned, running with a deleted configuration, leading to silent stats delta loss.

I think we should perform the deletion atomically under the same lock acquisition, re-verifying `isActive()` right before deletion:

```go
	for _, priority := range toRemove {
		if val, ok := fr.priorityBandStates.Load(priority); ok {
			if val.(*priorityBandState).isActive() {
				continue // Request pinned it; skip deletion
			}
		}
		fr.priorityBandStates.Delete(priority)
		fr.cleanupPriorityBandResourcesLocked([]int{priority}) // Locked helper
	}
	fr.mu.Unlock()
```

pkg/epp/flowcontrol/registry/registry.go
LukeAVanDrie: If we returned early on nil, a controller sync that deletes all `InferenceObjective` resources (passing nil/empty) would be ignored, leaking the existing dynamic priority bands and preventing them from being GC'd. Converting it to an empty map allows them to be safely cleaned up.

pkg/epp/flowcontrol/registry/registry.go
LukeAVanDrie: Since the channel carries the full desired state, queueing multiple stale updates in rapid succession would cause redundant dynamic provisioning/deleting cycles while draining ensures only the latest state is processed.  Updates probably will not come at a volume where this matters, but this seems like an okay optimization to me.

pkg/epp/flowcontrol/registry/registry.go
LukeAVanDrie: +1 Blocking is actually necessary here as the current synchronous fallback `ApplyDesiredPriorities` when the channel is full introduces an (unlikely) race condition where an older desired state can overwrite a newer one. E.g.,

1. Initial State: No dynamic priority bands exist.
2. Update 1 (Older - Priority 1 desired): A reconcile calls `SubmitDesiredPriorities([1])`. The channel is empty, so [1] is successfully written to the channel. (Channel contains [1]).
3. Update 2 (Newer - Priorities 1 and 2 desired): A rapid second reconcile calls `SubmitDesiredPriorities([1, 2])`. Since the channel is still full with [1], this thread hits default and runs `ApplyDesiredPriorities([1, 2])` synchronously. Registry state becomes [1, 2] (both bands provisioned).
4. The Reversion: The processor eventually wakes up, reads [1] (Update 1) from the channel, and calls `ApplyDesiredPriorities([1])`.
5. Failure: The registry identifies 2 as undesired (since it is not in [1]) and deletes it, reverting the registry state to the older [1] config and discarding the newer band 2.

In practice the race window is so small, that this is unlikely to happen though.

pkg/epp/flowcontrol/registry/registry.go
LukeAVanDrie: Although the only current production caller (`InferenceObjectiveReconciler.syncPriorityBands()`) discards its map immediately after submission (posing no active concurrent write risk today), I don't mind keeping this defensive copy as `SubmitDesiredPriorities` is a public interface that crosses goroutine boundaries.

LukeAVanDrie (changes requested):
Removing JIT provisioning of priority bands from the request path is a regression for default and demoted requests.

EPP defaults all requests without active objectives to Priority 0, and the new fallback demotes blocked requests to Priority 0.

However, EPP does not statically populate Priority 0 into `fr.config.PriorityBands` at startup unless the user explicitly lists it in their YAML config. Previously, EPP JIT-provisioned Priority 0 dynamically when the first request arrived. Because this PR removes JIT provisioning, any request defaulting/demoted to Priority 0 will now immediately fail with `ErrPriorityBandNotFound`.

We must guarantee Priority 0 exists out-of-the-box. We should statically inject Priority 0 into FlowControl.PriorityBands during configuration defaulting (e.g., in pkg/epp/config/loader/defaults.go) so that `FlowRegistry` initializes it statically on startup.

pkg/epp/flowcontrol/benchmark/benchmark.go
gyliu513: Done

pkg/epp/flowcontrol/contracts/registry.go
gyliu513: Done

pkg/epp/flowcontrol/registry/registry.go
gyliu513: +1 with @LukeAVanDrie

pkg/epp/flowcontrol/registry/registry.go
gyliu513: Yes, defensive copy for a public API that crosses goroutine boundaries. As @LukeAVanDrie points out, while the current only production caller discards its map immediately, `SubmitDesiredPriorities` is a public interface and callers may legitimately hold and mutate the map after the call. Keeping the copy here, hope it is ok.

pkg/epp/flowcontrol/registry/registry.go
gyliu513: Yes, will keep it now, hope it is ok.

pkg/epp/flowcontrol/registry/registry.go
gyliu513: @shmuelk yes, you're right, and @LukeAVanDrie found that the non-blocking fallback is actually incorrect: calling `ApplyDesiredPriorities` synchronously can cause an older desired state to overwrite a newer one already applied. After draining the stale entry we always have capacity, so the write will never block. Simplifying to a direct blocking send and removing the fallback.

pkg/epp/flowcontrol/registry/registry.go
gyliu513: Same rationale as above, nil from the processor loop means "desired state is empty", i.e., remove all dynamic bands. Converting nil to empty to trigger that cleanup.

pkg/epp/flowcontrol/registry/registry.go
gyliu513: Partially, the synchronous fallback from `SubmitDesiredPriorities` is being removed, but the lock is still necessary and @LukeAVanDrie found that its scope actually needs to be *broadened* to fix a TOCTOU race: unlocking before deletion lets a concurrent request pin a band and then have it deleted underneath it. Fixing by holding `fr.mu` through the entire deletion loop and introducing a `cleanupPriorityBandResourcesLocked` helper.

pkg/epp/flowcontrol/registry/registry.go
gyliu513: Addressed by the same TOCTOU fix, holding the lock through deletions rather than releasing it before calling `deletePriorityBand`.

pkg/epp/flowcontrol/registry/registry.go
gyliu513: Yes, good point. Will inline `executeGCCycle`'s body into `ExecuteGCCycle` and remove the private wrapper.

gyliu513 (commented):
@LukeAVanDrie @shmuelk great comments, I really learned a lot when trying to review and address those comments!

> Removing JIT provisioning of priority bands from the request path is a regression for default and demoted requests.

@LukeAVanDrie Correct, this is a regression. Fixing in `NewConfig` in registry/config.go: after all opts are applied, if priority 0 is not already in PriorityBands, it is cloned from DefaultPriorityBand (which is always set at that point) and added as a static initial band. This makes priority 0 always present and immune to GC, matching the behavior of other explicitly configured bands.

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -156,24 +170,106 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
...
 
-// Run starts the registry's background garbage collection loop.
-// It blocks until the provided context is cancelled.
+// Run blocks until the provided context is cancelled.
+// Registry maintenance (priority band sync and GC) runs in the Processor loop.
 func (fr *FlowRegistry) Run(ctx context.Context) {
-	fr.logger.Info("Starting FlowRegistry background garbage collection loop")
-	defer fr.logger.Info("FlowRegistry background garbage collection loop stopped")
+	<-ctx.Done()
+}
```
shmuelk: Why do we need this code? The call in runner.go of registry.Run was deleted.

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -156,24 +170,106 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
...
+
+// RunMaintenanceLoop applies priority band updates and runs registry GC until ctx is cancelled.
+// Production uses the Processor loop; this helper supports tests that run without a FlowController.
+func (fr *FlowRegistry) RunMaintenanceLoop(ctx context.Context) {
```
shmuelk, line 177: This is a new function that was added in this PR. It isn't referenced anywhere. Is something missing somewhere?

pkg/epp/flowcontrol/registry/registry.go
shmuelk: I understand what you are doing, I still think it would have been better to simply send deltas all of the time.

pkg/epp/flowcontrol/registry/registry.go
shmuelk: This doesn't drain the channel. You need a loop to do that. What happens if there are multiple items in the channel at once.

pkg/epp/flowcontrol/registry/registry.go
gyliu513: Yes, I will remove it.

pkg/epp/flowcontrol/registry/registry.go
gyliu513: This is mainly for test.

pkg/epp/flowcontrol/registry/registry.go
gyliu513: The channel is buffered with capacity 1, so a single non-blocking receive does fully drain it, comments?

pkg/epp/flowcontrol/registry/registry.go
gyliu513: Probably we can discuss this in another PR if you want some follow ups, this is more like a design preference, not a blocker, comments?

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -417,35 +493,34 @@ func (fr *FlowRegistry) gcPriorityBands() {
...
+		if p == priority {
+			fr.orderedPriorityLevels = append(
+				fr.orderedPriorityLevels[:i],
+				fr.orderedPriorityLevels[i+1:]...,
+			)
+			break
 		}
-
-		fr.logger.V(logging.DEFAULT).Info("Successfully deleted priority band", "priority", priority)
 	}
```
shmuelk: Replace with:
```suggestion
	fr.orderedPriorityLevels =
	       slices.DeleteFunc(fr.orderedPriorityLevels, func(p int) bool { return p == priority } )
```

gyliu513: Done

gyliu513:
@LukeAVanDrie @shmuelk can you help review again? I think all comments are now addressed.

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -417,35 +494,26 @@ func (fr *FlowRegistry) gcPriorityBands() {
...
+// The caller must hold fr.mu exclusively.
+func (fr *FlowRegistry) cleanupPriorityBandResourcesLocked(priority int) {
+	// Zombie protection: verify band was actually deleted from the state map.
+	if _, exists := fr.priorityBandStates.Load(priority); exists {
+		return
+	}
 
-		// Remove from sync.Map
-		fr.priorityBands.Delete(priority)
+	delete(fr.config.PriorityBands, priority)
```
LukeAVanDrie, line 517: This deletes statically configured bands, including the injected priority 0.

Both removal paths funnel through here, but only `ApplyDesiredPriorities` checks `initialPriorities` before calling cleanup; the time-based GC path (`gcPriorityBands` → `cleanupPriorityBandResources` → here) does not.

Static bands escape today only because `NewFlowRegistry` never creates a `priorityBandState` for them. However, as soon as a request opens a flow at a static priority, a `priorityBandState` exists, its lease drops to 0 when the flow idles, and after `PriorityBandGCTimeout` this line drops the band from config. With JIT removed nothing recreates it. For priority 0 it's fatal: `withConnectionWithDemotion` cannot demote below 0, so the next default request fails.

Repro (both assertions fail):
```go
h := newRegistryTestHarness(t, harnessOptions{manualGC: true})
key := flowcontrol.FlowKey{ID: "f", Priority: highPriority} // static band 20
h.openConnectionOnFlow(key)
h.fakeClock.Step(h.config.FlowGCTimeout + time.Second);         h.fr.ExecuteGCCycle()
h.fakeClock.Step(h.config.PriorityBandGCTimeout + time.Second); h.fr.ExecuteGCCycle()
// config.PriorityBands[highPriority] is now gone; a priority-0 variant then hard-fails.
```

We just need to honor `initialPriorities` here, mirroring the other guard you added:
```suggestion
	if _, static := fr.initialPriorities[priority]; static {
		return // static/initial bands are never collected; restores startup-equivalent state
	}

	delete(fr.config.PriorityBands, priority)
```

`collectLeasedResources` has already removed the transient `priorityBandState`, so the band lands back in its startup state. If you would rather avoid the band being re-collected every idle cycle, skip initial priorities in the `gcPriorityBands` collection loop instead. Either way, please add a regression test that ages a flow at a static priority (and at priority 0) past both timeouts and asserts the config survives and a follow-up request succeeds.

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -243,26 +332,16 @@ func (fr *FlowRegistry) WithConnection(key flowcontrol.FlowKey, fn func(conn con
 //
 // NOTE: The caller (WithConnection) must already hold a lease on the priority band to prevent GC during this operation.
 func (fr *FlowRegistry) ensureFlowInfrastructure(key flowcontrol.FlowKey) error {
-	// 1. Ensure Priority Band exists.
 	fr.mu.RLock()
 	_, exists := fr.config.PriorityBands[key.Priority]
```
LukeAVanDrie, line 337: Non-blocking nit: this existence check under its own `RLock` is redundant with `buildFlowComponents`, which re-checks the same map under a second `RLock` and returns the identical `ErrPriorityBandNotFound`.

You can drop this first `RLock`/`RUnlock` block and fold the not-found case into the single `buildFlowComponents` critical section resulting in one lock acquisition instead of two on every flow's first request.

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -156,24 +171,100 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
...
+		if _, static := fr.initialPriorities[priority]; static {
+			continue
+		}
+		if _, wanted := desired[priority]; wanted {
+			continue
+		}
+		if !fr.isPriorityBandIdle(priority) {
+			continue
+		}
+		// Re-verify under the lock: a concurrent request may have pinned the band
```
LukeAVanDrie, line 250: Non-blocking: the comment is a little misleading. `isPriorityBandIdle` and this re-check both run under the same held `fr.mu.Lock()`, so there's no window *between* them. Band pinning (`pinLeasedResource` on `priorityBandStates`) is lock-free and never takes `fr.mu`, so the write lock doesn't actually exclude a concurrent pin anyway. The removal is still safe (stale-object protection in `pinLeasedResource` + the priority-0 demotion fallback heal the race), but consider rewording to say the re-check is a cheap best-effort guard and the real safety net is the fallback + self-healing GC.

pkg/epp/flowcontrol/registry/registry.go
```diff
@@ -156,24 +171,100 @@ func NewFlowRegistry(config *Config, logger logr.Logger, opts ...RegistryOption)
...
 	}
 }
 
+// SubmitDesiredPriorities queues a priority band topology update for the processor loop.
+// Stale pending updates are dropped so only the latest desired state is retained.
+func (fr *FlowRegistry) SubmitDesiredPriorities(desired map[int]struct{}) {
+	if desired == nil {
+		desired = map[int]struct{}{}
+	}
+	copy := make(map[int]struct{}, len(desired))
```
LukeAVanDrie, line 199: Non-blocking nit: `copy` shadows the builtin. Maybe something like `predeclared`, `snapshot`, or `desiredCopy` instead?

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -266,6 +278,26 @@ func (fc *FlowController) EnqueueAndWait(
...
+		return err
+	}
+
+	fc.logger.V(logutil.DEFAULT).Info(
+		"Priority band not provisioned, demoting request to priority 0",
+		"originalPriority", key.Priority,
+		"flowID", key.ID,
+	)
+	demotedKey := key
+	demotedKey.Priority = 0
```
LukeAVanDrie, line 297: Non-blocking: for negative priorities this is a promotion, not a demotion. It's availability-first and an edge case (priorities come from provisioned objectives), so probably fine. Just flagging the semantic mismatch with the "demote" naming.

LukeAVanDrie (changes requested):
Thanks for pushing this through! One more blocker before this can land. The rest are non-blocking nits you can take or leave.

One thing worth noting: #1184's agreed end-state removed `fr.mu` entirely, and this PR keeps it (flow/queue provisioning still happens JIT on request goroutines, so `ensureFlowInfrastructure` still takes `RLock`s). This PR is still a very good incremental step toward this goal. I just wanted to check with @shmuelk whether full mutex removal is still desired as a followup or we can mark #1184 as fixed with this change.

pkg/epp/flowcontrol/registry/registry.go
gyliu513: will address in follow up PR

pkg/epp/flowcontrol/registry/registry.go
gyliu513: will address in follow up PR

pkg/epp/flowcontrol/registry/registry.go
gyliu513: will address in follow up PR

pkg/epp/flowcontrol/registry/registry.go
gyliu513: Done, thanks @LukeAVanDrie

LukeAVanDrie (approved):
Thanks, @gyliu513! LGTM

ahg-g (approved):

shmuelk:
@LukeAVanDrie said:

> One thing worth noting: https://github.com/llm-d/llm-d-router/issues/1184's agreed end-state removed fr.mu entirely, and this PR keeps it (flow/queue provisioning still happens JIT on request goroutines, so ensureFlowInfrastructure still takes RLocks). This PR is still a very good incremental step toward this goal. I just wanted to check with @shmuelk whether full mutex removal is still desired as a followup or we can mark https://github.com/llm-d/llm-d-router/issues/1184 as fixed with this change.

A full removal of the fr.mu is still the plan of record. This PR is definitely a step in that direction.

PR #1362, opened 2026-05-27
fix: resolve gRPC size parameter omission in runWithFileDiscovery runner initialization

weizhoublue:
/kind bug
 
 Issue: 
EPP running in FileDiscovery mode (non-Kubernetes environment) with custom gRPC size limits set via  --grpc-max-recv-msg-size  or  --grpc-max-send-msg-size . Inference payloads exceed default 4MiB limit.  Request fails. gRPC stream terminates with  ResourceExhausted  error ("received message larger than max").

gRPC server should processe messages up to configured size limit.
  
Root Cause:  
runWithFileDiscovery  initializes  ExtProcServerRunner  without forwarding  opts.GRPCMaxRecvMsgSize  and  opts.GRPCMaxSendMsgSize values. Size parameters default to 0, falling back to default 4MiB.

 

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -950,6 +950,8 @@ func (r *Runner) runWithFileDiscovery(ctx context.Context, opts *runserver.Optio
 		Parser:                           r.parser,
 		SaturationDetector:               eppConfig.SaturationDetector,
 		UseExperimentalDatalayerV2:       useNewMetrics,
+		GRPCMaxRecvMsgSize:               opts.GRPCMaxRecvMsgSize,
+		GRPCMaxSendMsgSize:               opts.GRPCMaxSendMsgSize,
 	}
 
 	r.customCollectors = append(r.customCollectors, collectors.NewInferencePoolMetricsCollector(ds))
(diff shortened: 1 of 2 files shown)


weizhoublue:
@ezrasilvera  hi , could you help check whether this PR makes sense ? thanks

LukeAVanDrie:
Thanks @weizhoublue! Missed this in #1155.

cmd/epp/runner/runner.go
```diff
@@ -950,6 +950,8 @@ func (r *Runner) runWithFileDiscovery(ctx context.Context, opts *runserver.Optio
 		Parser:                           r.parser,
 		SaturationDetector:               eppConfig.SaturationDetector,
 		UseExperimentalDatalayerV2:       useNewMetrics,
+		GRPCMaxRecvMsgSize:               opts.GRPCMaxRecvMsgSize,
```
ahg-g, line 953: can we define a default ExtProcServerRunner that gets used in both places and customized differently in each place if necessary? This will help avoid issues like this in the future

ezrasilvera: > can we define a default ExtProcServerRunner that gets used in both places and customized differently in each place if necessary? This will help avoid issues like this in the future

Sounds good. #1370

PR #1368, opened 2026-05-27
epp: wire FlowControl into file-discovery mode

ezrasilvera:
The runWithFileDiscovery path hardcoded NewLegacyAdmissionController, so the flowControl block in EndpointPickerConfig was parsed but silently ignored when running outside Kubernetes. Mirror the K8s mode's admission-control initialization branch

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -381,30 +381,11 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 		return nil, nil, err
 	}
 
-	// --- Admission Control Initialization ---
-	var admissionController requestcontrol.AdmissionController
-	var endpointCandidates contracts.EndpointCandidates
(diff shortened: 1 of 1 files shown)


cmd/epp/runner/runner.go
```diff
@@ -927,9 +933,41 @@ func (r *Runner) runWithFileDiscovery(ctx context.Context, opts *runserver.Optio
...
+			},
+		)
+		if err != nil {
+			return fmt.Errorf("failed to initialize Flow Controller: %w", err)
+		}
+		go registry.Run(ctx)
+		admissionController = requestcontrol.NewFlowControlAdmissionController(fc, opts.PoolName)
+	} else {
+		admissionController = requestcontrol.NewLegacyAdmissionController(eppConfig.SaturationDetector, endpointCandidates)
+	}
```
shmuelk: As this code is identical to the code in the K8S path, instead of mirroring it, why not make it a function that both paths call?

ezrasilvera: You are correct. fixed in dbeb4305

elevran:
@ezrasilvera please resolve conflict so we can merge

ezrasilvera:
> @ezrasilvera please resolve conflict so we can merge

Done

PR #1371, opened 2026-05-27
feat: add support in Helm chart for new tokenizer render

zdtsw:
**What type of PR is this?**
/kind feature

- add new schema router.tokenizer.render | router.tokenzier.uds
- the current route.tokenizer can only enable of disable for UDS the main issue is it hardcoded mount
- also introduce `router.tokenizer.modelName` which need to be used for render backend

Fixes #

```release-note
NONE
```

### DIFF
--- a/config/charts/routerlib/templates/_deployment.yaml
+++ b/config/charts/routerlib/templates/_deployment.yaml
@@ -258,33 +258,76 @@ spec:
         {{- if .Values.router.epp.volumeMounts }}
         {{- toYaml .Values.router.epp.volumeMounts | nindent 12 }}
         {{- end }}
-        {{- if .Values.router.tokenizer.enabled }}
+        {{- if .Values.router.tokenizer.uds.enabled }}
             - name: tokenizer-uds
               mountPath: /tmp/tokenizer
         {{- end }}
-        {{- if .Values.router.tokenizer.enabled }}
+        {{- if .Values.router.tokenizer.uds.enabled }}
         - name: tokenizer-uds
-          image: {{ .Values.router.tokenizer.image.registry }}/{{ .Values.router.tokenizer.image.repository }}:{{ .Values.router.tokenizer.image.tag }}
-          imagePullPolicy: {{ .Values.router.tokenizer.image.pullPolicy | default "IfNotPresent" }}
+          image: {{ .Values.router.tokenizer.uds.image.registry }}/{{ .Values.router.tokenizer.uds.image.repository }}:{{ .Values.router.tokenizer.uds.image.tag }}
+          imagePullPolicy: {{ .Values.router.tokenizer.uds.image.pullPolicy | default "IfNotPresent" }}
           env:
             - name: TOKENIZERS_DIR
               value: /tokenizers
             - name: HF_HOME
               value: /tokenizers
-            {{- if .Values.router.tokenizer.env }}
-            {{- toYaml .Values.router.tokenizer.env | nindent 12 }}
+            {{- if .Values.router.tokenizer.uds.env }}
+            {{- toYaml .Values.router.tokenizer.uds.env | nindent 12 }}
             {{- end }}
-          {{- if .Values.router.tokenizer.resources }}
+          {{- if .Values.router.tokenizer.uds.resources }}
           resources:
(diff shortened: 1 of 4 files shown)


config/charts/routerlib/values.yaml
```diff
@@ -66,25 +66,63 @@ epp:
...
-          name: llm-d-hf-token
-          key: HF_TOKEN
-  resources:
-    requests:
-      cpu: "8"
-      memory: 8Gi
-    
+  modelName: ""
+
+  # Deprecated. Use render instead.
```
hexfusion, line 76: 👍

config/charts/routerlib/values.yaml
```diff
@@ -66,25 +66,63 @@ epp:
...
+            name: llm-d-hf-token
+            key: HF_TOKEN
+    resources:
+      requests:
+        cpu: "8"
+        memory: 8Gi
+    volumeMounts: []
+
+  render:
+    enabled: false
```
shmuelk, line 97: You have disabled both the UDS renderer and the new vLLM renderer.

Shouldn't one or the other be enabled by default?

config/charts/routerlib/templates/_deployment.yaml
```diff
@@ -258,46 +258,93 @@ spec:
         {{- if .Values.router.epp.volumeMounts }}
         {{- toYaml .Values.router.epp.volumeMounts | nindent 12 }}
         {{- end }}
-        {{- if .Values.router.tokenizer.enabled }}
```
hexfusion, line 261: question: what happens if user with router.tokenizer.enabled=true already in their values.yaml?

ahg-g: the tokenizer field is new, it was merged a couple of days back, it has not been released yet

config/charts/routerlib/values.yaml
zdtsw: default values.yaml have  `pluginsCustomConfig: {} ` which need to be properly config for either uds or render to work.
mostly, anyone uses the chart to have sidecar need a custsomized values.yaml override this default one. so to have one of these two default `enabled: true` does not help much

config/charts/routerlib/templates/_deployment.yaml
zdtsw: yep, i skipped all "upgrade" cases as i hope nobody already used ths new chart (except nightly and e2e) so if we can get this change out before next release, it _should_ be ok  :P

ahg-g (commented):
this will break the nightly, do we have a PR ready in llm-d/llm-d to change the values files in the guides that uses the existing setup as a fast follow?

zdtsw:
> this will break the nightly, do we have a PR ready in llm-d/llm-d to change the values files in the guides that uses the existing setup as a fast follow?

https://github.com/llm-d/llm-d/pull/1507 is the original one adding "render" into guide, 
it has been updated to use new chart but need this PR if it is good enough.

vMaroon:
I think we can remove convenient deployment support for the uds tokenizer. We can keep a deployment example within the example configurations. There should be absolutely no reason to use it over the new path.

ahg-g:
> I think we can remove convenient deployment support for the uds tokenizer. We can keep a deployment example within the example configurations. There should be absolutely no reason to use it over the new path.

So basically we keep the chart structure as is, we just change the tokenizer sidecar to deploy render instead of uds.

zdtsw:
> > I think we can remove convenient deployment support for the uds tokenizer. We can keep a deployment example within the example configurations. There should be absolutely no reason to use it over the new path.
> 
> So basically we keep the chart structure as is, we just change the tokenizer sidecar to deploy render instead of uds.

kinda like to remove all reference here for the old UDS but only have 
```
tokenizer:
  modelName: ""
  render:
    enabled: false
    image:
....
```
or we wanna even remove the render schema to 
```
tokenizer:
  modelName: ""
  enabled: false
  image:
....

```

ahg-g:
The latter

zdtsw:
> The latter

lets follow up this in https://github.com/llm-d/llm-d-router/pull/1428 instead

Issue #1363, opened 2026-05-27
[tests] Add e2e tests to the /inference/v1/generate endpoint


roytman:
#1355 adds EPP support to the vLLm `/inference/v1/generate` endpoint.
e2e2 tests should be added.

roytman:
done in #2029

PR #1378, opened 2026-05-27
[tests][1/n] Add coordinator with e-p-d-pools dev env

revit13:
**What type of PR is this?**
/kind test

This PR implements the first phase toward adding end-to-end tests for the /inference/v1/generate endpoint, as outlined in the tracking [issue](https://github.com/llm-d/llm-d-router/issues/1363). A subsequent [PR](https://github.com/llm-d/llm-d-router/pull/1356) will introduce the actual e2e test suite.

The changes introduce a coordinator-driven E/P/D Pools topology — three independent InferencePools (one per phase), each with its own dedicated EPP, fronted by header-routed HTTPRoutes, with a coordinator pod orchestrating the multimodal pipeline.

Part of #1363

```release-note
NONE
```


test/e2e/e2e_suite_test.go
```diff
@@ -35,19 +35,19 @@ const (
 	// crdKustomizePath is the kustomize path for all CRDs (upstream GIE + local llm-d.ai).
 	crdKustomizePath = "../../config/crd"
 	// inferExtManifest is the manifest for the inference extension test resources.
-	inferExtManifest = "../../deploy/components/inference-gateway/inference-pools.yaml"
+	inferExtManifest = "../../deploy/components/inference-gateway/single-pool/inference-pools.yaml"
```
roytman: The directory is called `single-pool` but the yaml file is named `inference-pools.yaml`?  
Maybe the directory should be called `separate-pools`?

roytman (commented):
Can we reduce the number of kustomize files? 

Looks like some entries are duplicated across the files.
Please put attension that filters in EPP configurations for different pools are different:

```yaml
piVersion: llm-d.ai/v1alpha1
kind: EndpointPickerConfig
plugins:
- type: decode-filter           # keeps only decode pods
- type: prefix-cache-scorer     # triggers auto-creation of approx-prefix-cache-producer
- type: max-score-picker
schedulingProfiles:
- name: default
  plugins:
  - pluginRef: decode-filter
    weight: 1.0
  - pluginRef: max-score-picker
```

```yaml
piVersion: llm-d.ai/v1alpha1
kind: EndpointPickerConfig
plugins:
- type: prefill-filter           # keeps only prefill pods
- type: prefix-cache-scorer     # triggers auto-creation of approx-prefix-cache-producer
- type: max-score-picker
schedulingProfiles:
- name: default
  plugins:
  - pluginRef: prefill-filter
    weight: 1.0
  - pluginRef: max-score-picker
 ```

We need to check if the pools are separated do we need the filter at all.

test/e2e/e2e_suite_test.go
revit13: The directory name describes the topology: single-pool/ is the topology where one unified InferencePool covers all phases (encode, prefill, decode in one pod), and it sits alongside epd-pools/ which is the topology where the work is split across three separate per-phase pools. 
I'll rename deploy/components/inference-gateway/single-pool/inference-pools.yaml → inference-pool.yaml to indicate it contains one inference pool. Thanks

Issue #1364, opened 2026-05-27
Standardize InferenceRequestBody type across api formats for api-agnostic plugin consumption


ahg-g:
This is an overarching issue to track all sub issues related to the subject:

- [ ] https://github.com/llm-d/llm-d-router/issues/1072
- [x] https://github.com/llm-d/llm-d-router/issues/1303
- [x] https://github.com/llm-d/llm-d-router/issues/848
- [x] https://github.com/llm-d/llm-d-router/issues/1062
- [x] https://github.com/llm-d/llm-d-router/issues/1092
- [x] https://github.com/llm-d/llm-d-router/issues/943
- [x] https://github.com/llm-d/llm-d-router/issues/1355

Pls link any other related issues so we design a future proof solution for all

/assign @zetxqx

roytman:
hi @ahg-g , should [[Feature] Support vLLM /inference/v1/generate endpoint by the llm-d-router](https://github.com/llm-d/llm-d-router/issues/1355), be included?

roytman:
This issue might be relevant too: https://github.com/llm-d/llm-d-router/issues/1367

The idea is to optimize EPP processing in disaggregated deployments by providing extra fields. 
For example, providing tokens in `chat/completions` requests and preventing redundant tokenization

It was presented on the `sig-router` call on May 20.

zetxqx:
Here is the plan, PR in coming


# Execution Plan: InferenceRequestBody Standardization

Goal: Standardize InferenceRequestBody across all API formats to enable API-agnostic scheduling and request-control plugin consumption.


---

## High-Level Checklist

- [x] Stage 1: Struct Coexistence & Fallbacks
- [ ] Stage 2: OpenAI Parser Migration
- [ ] Stage 3: vLLM gRPC Parser Migration
- [ ] Stage 4: Approximate Prefix Cache Hashing Migration
- [ ] Stage 5: Context Length Scorer & Disaggregated Prefill Migration
- [ ] Stage 6: Tokenizer Plugin Migration
- [ ] Stage 7: Precise Prefix Cache Scorer Migration
- [ ] Stage 8: Legacy Field Cleanup & Deprecation

---

## Stage 1: Struct Coexistence & Fallbacks

Introduce the new unified abstractions alongside the legacy monolithic fields, and implement fallback logic inside helper methods in `types.go` to guarantee safety.

### Tasks:
- [x] Define the new UnifiedPrompt struct:
[...]

PR #1380, opened 2026-05-28
refactor(unifiedInferenceReqeustBody)[1/n]: unify request body prompt and token representations

zetxqx:
**What type of PR is this?**

/cleanup

This PR refactors the HTTP request parsers (OpenAI, vLLM HTTP, and Anthropic Messages) to use the newly introduced unified, batch-friendly slice-of-struct fields (`Prompts []UnifiedPrompt` and `TokenInputs []TokenizedInput`) in `InferenceRequestBody`.

### Proposed Go Structures

```go
type InferenceRequestBody struct {
	// Prompts contains the unified representation of the user inputs.
	// Supports batching (multiple prompts). For a single-prompt request,
	// the input is located at index 0 (Prompts[0]).
	Prompts []UnifiedPrompt

	// TokenInputs is the pre-tokenized input provided by the client.
	// Supports batching (multiple token streams). For a single-prompt request,
	// the input is located at index 0 (TokenInputs[0]).
	TokenInputs []TokenizedInput
    
	// ... other compatibility fields
}

// UnifiedPrompt is a unified representation of the user input (prompt) for a single request item.
type UnifiedPrompt struct {
	// Messages represents the structured chat history or conversation turns.
	Messages []PromptMessage
}

// PromptMessage represents a single turn or message in a conversation.
type PromptMessage struct {
	// Role specifies the role (e.g., 'user', 'assistant', 'system').
	Role string
	// Blocks contains the content blocks for this message.
	Blocks []PromptBlock
}

// PromptBlock identifies the type and carries the raw content for a media chunk (text, image, tool, doc).
type PromptBlock struct {
	Type     BlockType
	Text     string
	AssetURI string
	Data     any
}
```

---

### Detailed Mapping Guide for All 7 Request Types

Here is the side-by-side JSON comparison showing exactly how each incoming endpoint request is parsed and converted into EPP's unified request body fields (**`prompts`** and **`token_inputs`**).

#### 1a. OpenAI Chat Completions (`/v1/chat/completions`) - Text Turn with Tools
* **Flow**: Maps incoming messages turns to `messages` turns, and swallows global `tools` into the `system` turn's blocks at index 0.

<table>
<tr>
<td><b>Incoming Request JSON</b></td>
<td><b>EPP Unified JSON</b></td>
</tr>
<tr>
<td>

```json
{
  "model": "test",
  "messages": [
    {"role": "user", "content": "hello"}
  ],
  "tools": [
    {"type": "function", "function": {"name": "get_weather"}}
  ]
}
```

</td>
<td>

```json
{
  "prompts": [
    {
      "messages": [
        {
          "role": "system",
          "blocks": [
            {
              "type": "tool",
              "data": {"type": "function", "function": {"name": "get_weather"}}
            }
          ]
        },
        {
          "role": "user",
          "blocks": [
            {
              "type": "text",
              "text": "hello"
            }
          ]
        }
      ]
    }
  ]
}
```

</td>
</tr>
</table>

---

#### 1b. OpenAI Chat Completions (`/v1/chat/completions`) - Multimodal Turn
* **Flow**: Maps rich-media content block arrays directly into separate unified **`PromptBlock`** elements (e.g., a text block and an image block) within the same message turn blocks.

<table>
<tr>
<td><b>Incoming Request JSON</b></td>
<td><b>EPP Unified JSON</b></td>
</tr>
<tr>
<td>

```json
{
  "model": "test-vision",
  "messages": [
    {
      "role": "user",
      "content": [
        {"type": "text", "text": "What is in this image?"},
        {"type": "image_url", "image_url": {"url": "https://example.com/image.png"}}
      ]
    }
  ]
}
```

</td>
<td>

```json
{
  "prompts": [
    {
      "messages": [
        {
          "role": "user",
          "blocks": [
            {
              "type": "text",
              "text": "What is in this image?"
            },
            {
              "type": "image",
              "asset_uri": "https://example.com/image.png"
            }
          ]
        }
      ]
    }
  ]
}
```

</td>
</tr>
</table>

---

#### 2. Claude Messages (`/v1/messages`)
* **Flow**: Maps top-level system prompt and global `tools` into the `system` message turn at `messages[0]`, and maps conversational turns to subsequent turns. Converts structured image base64 sources into EPP's unified image blocks.

<table>
<tr>
<td><b>Incoming Request JSON</b></td>
<td><b>EPP Unified JSON</b></td>
</tr>
<tr>
<td>

```json
{
  "model": "claude-3",
  "system": "You are a helpful assistant.",
  "messages": [
    {"role": "user", "content": "Hello"}
  ],
  "tools": [
    {"name": "get_weather", "description": "Get weather"}
  ]
}
```

</td>
<td>

```json
{
  "prompts": [
    {
      "messages": [
        {
          "role": "system",
          "blocks": [
            {
              "type": "text",
              "text": "You are a helpful assistant."
            },
            {
              "type": "tool",
              "data": {"name": "get_weather", "description": "Get weather"}
            }
          ]
        },
        {
          "role": "user",
          "blocks": [
            {
              "type": "text",
              "text": "Hello"
            }
          ]
        }
      ]
    }
  ]
}
```

</td>
</tr>
</table>

---

#### 3. OpenAI Completions (`/v1/completions`) - Text Batching
* **Flow**: Maps each prompt string in the batch `prompt` array into a completely separate independent `UnifiedPrompt` structure in the `prompts` slice.
* **Note**: If the `prompt` is a batched array of multiple strings, EPP maps each string to its own separate `UnifiedPrompt` structure inside the `prompts` slice. This is structurally represented as an array of prompts, ensuring that batch items are kept independent.

<table>
<tr>
<td><b>Incoming Request JSON</b></td>
<td><b>EPP Unified JSON</b></td>
</tr>
<tr>
<td>

```json
{
  "model": "gpt-3.5-turbo-instruct",
  "prompt": ["prompt 1", "prompt 2"]
}
```

</td>
<td>

```json
{
  "prompts": [
    { // First UnifiedPrompt
      "messages": [
        {
          "role": "",
          "blocks": [{"type": "text", "text": "prompt 1"}]
        }
      ]
    },
    { // Second UnifiedPrompt
      "messages": [
        {
          "role": "",
          "blocks": [{"type": "text", "text": "prompt 2"}]
        }
      ]
    }
  ]
}
```

</td>
</tr>
</table>

---

#### 4. OpenAI Embeddings (`/v1/embeddings`) - Text Batching
* **Flow**: Maps batched prompt strings into separate `UnifiedPrompt` structures in the `prompts` slice.
* **Note**: If the `input` is a batched array of multiple strings, EPP maps each string to its own separate `UnifiedPrompt` structure inside the `prompts` slice. This is structurally represented as an array of prompts, ensuring that batch items are kept independent.

<table>
<tr>
<td><b>Incoming Request JSON</b></td>
<td><b>EPP Unified JSON</b></td>
</tr>
<tr>
<td>

```json
{
  "model": "text-embedding-3-small",
  "input": ["embedding 1", "embedding 2"]
}
```

</td>
<td>

```json
{
  "prompts": [
    { // First UnifiedPrompt
      "messages": [
        {
          "role": "",
          "blocks": [{"type": "text", "text": "embedding 1"}]
        }
      ]
    },
    { // Second UnifiedPrompt
      "messages": [
        {
          "role": "",
          "blocks": [{"type": "text", "text": "embedding 2"}]
        }
      ]
    }
  ]
}
```

</td>
</tr>
</table>

---

#### 5. OpenAI Responses (`/v1/responses`)
* **Flow**: Maps top-level `instructions` and global `tools` into the system message turn at `messages[0]`, and maps `input` into a user turn block.

<table>
<tr>
<td><b>Incoming Request JSON</b></td>
<td><b>EPP Unified JSON</b></td>
</tr>
<tr>
<td>

```json
{
  "model": "gpt-4o",
  "input": "How do I check Python instance?",
  "instructions": "Talk like a pirate.",
  "tools": [{"type": "function", "function": {"name": "get_weather"}}]
}
```

</td>
<td>

```json
{
  "prompts": [
    {
      "messages": [
        {
          "role": "system",
          "blocks": [
            {"type": "text", "text": "Talk like a pirate."},
            {
              "type": "tool",
              "data": {"type": "function", "function": {"name": "get_weather"}}
            }
          ]
        },
        {
          "role": "",
          "blocks": [{"type": "text", "text": "How do I check Python instance?"}]
        }
      ]
    }
  ]
}
```

</td>
</tr>
</table>

---

#### 6. OpenAI Conversations (`/v1/conversations`)
* **Flow**: Loops through history items and maps roles and contents directly into conversation turns.

<table>
<tr>
<td><b>Incoming Request JSON</b></td>
<td><b>EPP Unified JSON</b></td>
</tr>
<tr>
<td>

```json
{
  "model": "gpt-4o",
  "items": [
    {"role": "user", "content": "hello"},
    {"role": "assistant", "content": "hi"}
  ]
}
```

</td>
<td>

```json
{
  "prompts": [
    {
      "messages": [
        {
          "role": "user",
          "blocks": [{"type": "text", "text": "hello"}]
        },
        {
          "role": "assistant",
          "blocks": [{"type": "text", "text": "hi"}]
        }
      ]
    }
  ]
}
```

</td>
</tr>
</table>

---

#### 7. vLLM HTTP Generate (`/inference/v1/generate`)
* **Flow**: Maps raw pre-tokenized `token_ids` directly to EPP's `token_inputs` batch slice. Flattens map-shaped multimodal placeholder ranges and hashes into a flat unified list.

<table>
<tr>
<td><b>Incoming Request JSON</b></td>
<td><b>EPP Unified JSON</b></td>
</tr>
<tr>
<td>

```json
{
  "token_ids": [151644, 872, 198, 3838, 151645],
  "features": {
    "mm_hashes": {"image": ["abc123hash"]},
    "mm_placeholders": {"image": [{"offset": 1, "length": 3}]}
  }
}
```

</td>
<td>

```json
{
  "token_inputs": [
    {
      "token_ids": [151644, 872, 198, 3838, 151645],
      "multimodal_features": [
        {
          "modality": "image",
          "hash": "abc123hash",
          "offset": 1,
          "length": 3
        }
      ]
    }
  ]
}
```

</td>
</tr>
</table>

Part of # https://github.com/llm-d/llm-d-router/issues/1364#issuecomment-4560679710

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/interface/requesthandling/types.go
+++ b/pkg/epp/framework/interface/requesthandling/types.go
@@ -66,10 +66,14 @@ func (RawPayload) IsParsed() bool    { return false }
 // InferenceRequestBody contains the request-body fields that we parse out as user input,
 // to be used in forming scheduling decisions.
 type InferenceRequestBody struct {
-	// Prompt is the unified representation of the user input.
-	Prompt *UnifiedPrompt
-	// Tokens is the pre-tokenized input provided by the client.
-	Tokens *TokenizedInput
+	// Prompts contains the unified representation of the user inputs.
+	// Supports batching (multiple prompts). For a single-prompt request,
+	// the input is located at index 0 (Prompts[0]).
+	Prompts []UnifiedPrompt
+	// TokenInputs is the pre-tokenized input provided by the client.
+	// Supports batching (multiple token streams). For a single-prompt request,
+	// the input is located at index 0 (TokenInputs[0]).
+	TokenInputs []TokenizedInput
 	// ExtractedCacheSalt is an optional request parameter to isolate prefix caches for security reasons.
 	ExtractedCacheSalt string
 
@@ -87,6 +91,7 @@ type InferenceRequestBody struct {
 	Embeddings *EmbeddingsRequest `json:"embeddings,omitempty"`
 	// GenerateRequest is the representation of the vLLM /inference/v1/generate request body.
 	Generate *GenerateRequest `json:"generate,omitempty"`
+
 	// Payload contains the unmarshaled request payload or raw bytes.
 	// If the payload is unmarshaled, we can perform advanced processing (like prefix cache aware routing).
 	// If it remains as raw bytes, such processing may not be supported.
@@ -103,28 +108,23 @@ type InferenceRequestBody struct {
 // TokenizedInput represents pre-tokenized input provided by the client,
 // including multimodal features if present.
 type TokenizedInput struct {
-	// TokenIDs are the pre-tokenized input token IDs. It supports an array of token
-	// arrays to accommodate batched prompts (e.g., from /completions).
-	TokenIDs [][]uint32
+	// TokenIDs are the pre-tokenized input token IDs for a single prompt.
+	TokenIDs []uint32
 	// MultiModalFeatures holds one entry per multimodal item in prompt order.
 	MultiModalFeatures []MultiModalFeature
 }
 
-// UnifiedPrompt is a unified representation of the user input (prompt) for all request types.
+// UnifiedPrompt is a unified representation of the user input (prompt) for a single request item.
 type UnifiedPrompt struct {
-	// Items represents structured chat history or conversation items.
-	Items []PromptItem
-	// Tools field for function calling capabilities.
-	Tools []any
-	// Documents field for HuggingFace RAG documents.
-	Documents []any
+	// Messages represents the structured chat history or conversation turns.
+	Messages []PromptMessage
 }
 
-// PromptItem represents a single turn or message in a conversation.
-type PromptItem struct {
+// PromptMessage represents a single turn or message in a conversation.
+type PromptMessage struct {
 	// Role specifies the role (e.g., 'user', 'assistant', 'system').
 	Role string
-	// Blocks contains the content blocks for this item.
+	// Blocks contains the content blocks for this message.
 	Blocks []PromptBlock
 }
 
@@ -136,19 +136,23 @@ const (
 	BlockTypeImage BlockType = "image"
 	BlockTypeAudio BlockType = "audio"
 	BlockTypeVideo BlockType = "video"
+	BlockTypeTool  BlockType = "tool"
+	BlockTypeDoc   BlockType = "document"
 )
 
-// PromptBlock represents a content block (text or asset) within a prompt item.
+// PromptBlock represents a content block (text, asset, tool, or document) within a prompt item.
 type PromptBlock struct {
-	// Type specifies the block type (e.g., 'text', 'image').
+	// Type specifies the block type (e.g., 'text', 'image', 'tool', 'document').
 	Type BlockType
-	// Text contains raw text content.
+	// Text contains raw text content (e.g. text block, document content, tool description).
 	Text string
 	// AssetURI refers to a multimodal asset (e.g., image URL, audio data hash).
 	AssetURI string
+	// Data holds structured data for complex blocks (like Tool or Document).
+	Data any
 }
 
-// TokenizedPrompt contains the result of tokenizing the request prompt.
+// TokenizedPrompt contains the result of tokenizing a single request prompt.
 // It is consumed by scheduling and request-control plugins that benefit from
 // actual token data such as prefix-cache awareness.
 type TokenizedPrompt struct {
@@ -173,26 +177,15 @@ type MultiModalFeature struct {
 	Length int
 }
 
-// PromptText returns a plain-text representation of the prompt.
+// PromptText returns a plain-text representation of the prompt, aggregating all batched prompts.
 func (r *InferenceRequestBody) PromptText() string {
-	if r.Prompt != nil {
+	if len(r.Prompts) > 0 {
 		var sb strings.Builder
-		for _, item := range r.Prompt.Items {
-			for _, block := range item.Blocks {
-				if block.Type == BlockTypeText && block.Text != "" {
-					sb.WriteString(block.Text)
-					sb.WriteString(" ")
-				}
-			}
-		}
-		if len(r.Prompt.Documents) > 0 {
-			for _, doc := range r.Prompt.Documents {
-				if str, ok := doc.(string); ok {
-					sb.WriteString(str)
-					sb.WriteString(" ")
-				} else {
-					if bytes, err := json.Marshal(doc); err == nil {
-						sb.Write(bytes)
+		for _, prompt := range r.Prompts {
+			for _, msg := range prompt.Messages {
+				for _, block := range msg.Blocks {
+					if (block.Type == BlockTypeText || block.Type == BlockTypeDoc) && block.Text != "" {
+						sb.WriteString(block.Text)
 						sb.WriteString(" ")
 					}
 				}
@@ -250,17 +243,14 @@ func (r *InferenceRequestBody) PromptText() string {
 
 // InputTokenCountHint returns a best-effort input token count when the
 // caller knows it exactly (token-ID inputs), or -1 when the count has
-// to be estimated from text.
+// to be estimated from text. It aggregates counts from all batched tokens.
 func (r *InferenceRequestBody) InputTokenCountHint() int {
-	if r.Prompt != nil || r.Tokens != nil {
-		if r.Tokens != nil && len(r.Tokens.TokenIDs) > 0 {
-			total := 0
-			for _, ids := range r.Tokens.TokenIDs {
-				total += len(ids)
-			}
-			return total
+	if len(r.TokenInputs) > 0 {
+		total := 0
+		for _, t := range r.TokenInputs {
+			total += len(t.TokenIDs)
 		}
-		return -1
+		return total
 	}
 
 	// Fallback to old fields
@@ -827,4 +817,37 @@ type AnthropicImageSource struct {
 	MediaType string `json:"media_type,omitempty"`
 	Data      string `json:"data,omitempty"`
 	URL       string `json:"url,omitempty"`
+}
+
+// ConvertMMFeaturesToUpstream flattens the kv-cache map-shaped multimodal
+// metadata into the upstream flat list, sorted by placeholder offset.
+func ConvertMMFeaturesToUpstream(src *tokenization.MultiModalFeatures) []MultiModalFeature {
+	if src == nil || len(src.MMHashes) == 0 {
+		return nil
+	}
+
+	var items []MultiModalFeature
+	for modality, hashes := range src.MMHashes {
+		ranges, ok := src.MMPlaceholders[modality]
+		if !ok {
+			continue
+		}
+		n := len(hashes)
+		if len(ranges) < n {
+			n = len(ranges)
+		}
+		for i := 0; i < n; i++ {
+			items = append(items, MultiModalFeature{
+				Modality: Modality(modality),
+				Hash:     hashes[i],
+				Offset:   ranges[i].Offset,
+				Length:   ranges[i].Length,
+			})
+		}
+	}
+	if len(items) == 0 {
+		return nil
+	}
+	sort.Slice(items, func(i, j int) bool { return items[i].Offset < items[j].Offset })
+	return items
 }
--- a/pkg/epp/framework/interface/requesthandling/types.go
+++ b/pkg/epp/framework/interface/requesthandling/types.go
@@ -132,12 +132,12 @@ type PromptMessage struct {
 type BlockType string
 
 const (
-	BlockTypeText  BlockType = "text"
-	BlockTypeImage BlockType = "image"
-	BlockTypeAudio BlockType = "audio"
-	BlockTypeVideo BlockType = "video"
-	BlockTypeTool  BlockType = "tool"
-	BlockTypeDoc   BlockType = "document"
+	BlockTypeText     BlockType = "text"
+	BlockTypeImage    BlockType = "image"
+	BlockTypeAudio    BlockType = "audio"
+	BlockTypeVideo    BlockType = "video"
+	BlockTypeTool     BlockType = "tool"
+	BlockTypeDocument BlockType = "document"
 )
 
 // PromptBlock represents a content block (text, asset, tool, or document) within a prompt item.
@@ -184,7 +184,7 @@ func (r *InferenceRequestBody) PromptText() string {
 		for _, prompt := range r.Prompts {
 			for _, msg := range prompt.Messages {
 				for _, block := range msg.Blocks {
-					if (block.Type == BlockTypeText || block.Type == BlockTypeDoc) && block.Text != "" {
+					if (block.Type == BlockTypeText || block.Type == BlockTypeDocument) && block.Text != "" {
 						sb.WriteString(block.Text)
 						sb.WriteString(" ")
 					}
@@ -817,37 +817,4 @@ type AnthropicImageSource struct {
 	MediaType string `json:"media_type,omitempty"`
 	Data      string `json:"data,omitempty"`
 	URL       string `json:"url,omitempty"`
-}
-
-// ConvertMMFeaturesToUpstream flattens the kv-cache map-shaped multimodal
-// metadata into the upstream flat list, sorted by placeholder offset.
-func ConvertMMFeaturesToUpstream(src *tokenization.MultiModalFeatures) []MultiModalFeature {
-	if src == nil || len(src.MMHashes) == 0 {
-		return nil
-	}
-
-	var items []MultiModalFeature
-	for modality, hashes := range src.MMHashes {
-		ranges, ok := src.MMPlaceholders[modality]
-		if !ok {
-			continue
-		}
-		n := len(hashes)
-		if len(ranges) < n {
-			n = len(ranges)
-		}
-		for i := 0; i < n; i++ {
-			items = append(items, MultiModalFeature{
-				Modality: Modality(modality),
-				Hash:     hashes[i],
-				Offset:   ranges[i].Offset,
-				Length:   ranges[i].Length,
-			})
-		}
-	}
-	if len(items) == 0 {
-		return nil
-	}
-	sort.Slice(items, func(i, j int) bool { return items[i].Offset < items[j].Offset })
-	return items
 }
--- a/pkg/epp/framework/interface/requesthandling/types.go
+++ b/pkg/epp/framework/interface/requesthandling/types.go
@@ -65,8 +65,14 @@ func (RawPayload) IsParsed() bool    { return false }
 
 // InferenceRequestBody contains the request-body fields that we parse out as user input,
 // to be used in forming scheduling decisions.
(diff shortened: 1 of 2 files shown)


pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -86,6 +92,53 @@ type InferenceRequestBody struct {
...
+}
+
+// UnifiedPrompt is a unified representation of the user input (prompt) for all request types.
+type UnifiedPrompt struct {
+	// Items represents structured chat history or conversation items.
+	Items []PromptItem
+	// Tools field for function calling capabilities.
+	Tools []any
+	// Documents field for HuggingFace RAG documents.
+	Documents []any
```
ahg-g: 1) Which protocols support this? pls give examples.
2) What types can any be casted to here? how do downstream consumers interpret it?

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -86,6 +92,53 @@ type InferenceRequestBody struct {
...
+	// MultiModalFeatures holds one entry per multimodal item in prompt order.
+	MultiModalFeatures []MultiModalFeature
+}
+
+// UnifiedPrompt is a unified representation of the user input (prompt) for all request types.
+type UnifiedPrompt struct {
+	// Items represents structured chat history or conversation items.
+	Items []PromptItem
+	// Tools field for function calling capabilities.
+	Tools []any
```
ahg-g: What types does any take here? how do downstream consumers interpret it?

ahg-g (commented):
Looks good overall.

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -86,6 +92,53 @@ type InferenceRequestBody struct {
 	Stream bool `json:"-"`
 }
 
+// TokenizedInput represents pre-tokenized input provided by the client,
+// including multimodal features if present.
+type TokenizedInput struct {
+	// TokenIDs are the pre-tokenized input token IDs.
+	TokenIDs []uint32
```
roytman, line 112: To support an array of tokens arrays of `/completions`, should it be [][]uini32?

zetxqx: updated

pkg/epp/framework/interface/requesthandling/types.go
zetxqx: 1. currently in our internal struct tools is used as any I just keeps it as is. but we should expand it in a more structure struct. currently in our internal parsing it is supported by ChatCompletion and Response: https://github.com/llm-d/llm-d-router/blob/75ff054ce09a89e1236ed8b958c5fa98fcc4b16c/pkg/epp/framework/interface/requesthandling/types.go#L347 https://github.com/llm-d/llm-d-router/blob/75ff054ce09a89e1236ed8b958c5fa98fcc4b16c/pkg/epp/framework/interface/requesthandling/types.go#L316

2. Currently the only downstream consumers for this field is the approximate prefix plugin, I would expect only downstream approximate prefix cares about this: implemented in https://github.com/llm-d/llm-d-router/pull/1383/changes#diff-bcad99b73da2ba373d0dad540ce8a61a887b78a383cd1137d5748467199bc151

pkg/epp/framework/interface/requesthandling/types.go
zetxqx: I currently just keep our internal [any](https://github.com/llm-d/llm-d-router/blob/75ff054ce09a89e1236ed8b958c5fa98fcc4b16c/pkg/epp/framework/interface/requesthandling/types.go#L347) as is. But maybe a good step here is introduce the full struct as is instead of using any.

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -65,8 +65,14 @@ func (RawPayload) IsParsed() bool    { return false }
...
 // InferenceRequestBody contains the request-body fields that we parse out as user input,
 // to be used in forming scheduling decisions.
-// An InferenceRequestBody must contain exactly one of CompletionsRequest, ChatCompletionsRequest, ResponsesRequest, ConversationsRequest, EmbeddingsRequest, GenerateRequest, or MessagesRequest.
 type InferenceRequestBody struct {
+	// Prompt is the unified representation of the user input.
+	Prompt *UnifiedPrompt
+	// Tokens is the pre-tokenized input provided by the client.
+	Tokens *TokenizedInput
+	// ExtractedCacheSalt is an optional request parameter to isolate prefix caches for security reasons.
+	ExtractedCacheSalt string
```
ahg-g, line 78: nit: name this CacheSalt, we named the other two parameters Prompt and Tokens, not ExtractedPrompt and ExtractedTokens.

pkg/epp/framework/interface/requesthandling/types.go
zetxqx: I updated the struct to swallow the Tools and Documents, please take a look. And also instead of  making the batch array inside of the struct I made the field explicitly an array type. There are two benefits: 1. this will make the prompts field naturally be batch friendly 2. we don't need pointers any more

pkg/epp/framework/interface/requesthandling/types.go
zetxqx: I wanted to directly use CacheSalt as a field, however it has collision with existing [CacheSalt](https://github.com/llm-d/llm-d-router/blob/1f48bd868a5beb26451b278ae80a64365b50d53b/pkg/epp/framework/interface/requesthandling/types.go#L187) method.  We can unify it after the refactoring

zetxqx:
@ahg-g I basically proposed a new struct and updated the PR description. Could you take a look to see if it makes more sense?

ahg-g:
> @ahg-g I basically proposed a new struct and updated the PR description. Could you take a look to see if it makes more sense?

This is very neat! one question: is it only `/v1/completions` that has the semantics of a batch of prompts?

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -65,8 +65,18 @@ func (RawPayload) IsParsed() bool    { return false }
...
+	// Prompts contains the unified representation of the user inputs.
+	// Supports batching (multiple prompts). For a single-prompt request,
+	// the input is located at index 0 (Prompts[0]).
+	Prompts []UnifiedPrompt
+	// TokenInputs is the pre-tokenized input provided by the client.
+	// Supports batching (multiple token streams). For a single-prompt request,
+	// the input is located at index 0 (TokenInputs[0]).
+	TokenInputs []TokenizedInput
+	// ExtractedCacheSalt is an optional request parameter to isolate prefix caches for security reasons.
+	ExtractedCacheSalt string
```
elevran, line 78: nit: `Prompts`, `TokenInputs`, and `ExtractedCacheSalt` carry no JSON tag. Every existing field on InferenceRequestBody has either an explicit json tag or json:"-", so the absence here is ambiguous.
If these fields are internal and should never appear in serialized output, `json:"-"` documents that intent and protects against accidental marshaling.

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -94,7 +105,54 @@ type InferenceRequestBody struct {
...
+	BlockTypeText  BlockType = "text"
+	BlockTypeImage BlockType = "image"
+	BlockTypeAudio BlockType = "audio"
+	BlockTypeVideo BlockType = "video"
+	BlockTypeTool  BlockType = "tool"
+	BlockTypeDoc   BlockType = "document"
+)
+
+// PromptBlock represents a content block (text, asset, tool, or document) within a prompt item.
+type PromptBlock struct {
```
elevran, line 144: Anthropic's `cache_breakpoints` is a per-block field: it marks that the cumulative prefix up to and including that block should be cached. Without a `CacheBreakpoint` field on `PromptBlock`, the Anthropic parser in later stages cannot represent this and will either discard the information or need to extend this type after the fact. Adding it here is a one-field change; adding it after [2/n] parsers have compiled against this struct is a larger migration.
I expect this field will be needed by prefix matchers so needs to be present internally.

One thing we should be explicit about is the ability to reproduce the input (e.g., client request) as output (e.g., to model server). It does not have to be via the `InferenceRequest`, but I'm not clear from #1364 how this is tackled. Should allow verbatim copy as well as potential changes (e.g., ModelRewrite) to affect the output.

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -94,7 +105,54 @@ type InferenceRequestBody struct {
...
+	// Type specifies the block type (e.g., 'text', 'image', 'tool', 'document').
+	Type BlockType
+	// Text contains raw text content (e.g. text block, document content, tool description).
+	Text string
+	// AssetURI refers to a multimodal asset (e.g., image URL, audio data hash).
+	AssetURI string
+	// Data holds structured data for complex blocks (like Tool or Document).
+	Data any
```
elevran, line 152: Tool and document payloads land in Data as untyped any. Once parsers in [2/n] start populating this field and consumers read it via type assertions, typing it retroactively means updating every call site. This is the natural point in the migration to define typed structs for tool and document payloads - **before** downstream code has been written against Data `any`.

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -730,3 +818,36 @@ type AnthropicImageSource struct {
 	Data      string `json:"data,omitempty"`
 	URL       string `json:"url,omitempty"`
 }
+
+// ConvertMMFeaturesToUpstream flattens the kv-cache map-shaped multimodal
+// metadata into the upstream flat list, sorted by placeholder offset.
+func ConvertMMFeaturesToUpstream(src *tokenization.MultiModalFeatures) []MultiModalFeature {
```
elevran: This function converts a tokenization-package input type into the unified representation. Placing it here couples the core IR package to the vLLM generate parser's upstream shape (hopefully sglang has the same shape). As new parsers are added, the types file would accumulate a conversion utility per parser format, which is the wrong direction for a shared interface package. It belongs in the generate parser package. The function also has no test coverage for its non-trivial logic: length-capping when hashes and ranges are mismatched, and deterministic ordering across multiple modalities after map iteration.

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -94,7 +105,54 @@ type InferenceRequestBody struct {
...
+// BlockType identifies the type of content block in a prompt.
+type BlockType string
+
+const (
+	BlockTypeText  BlockType = "text"
+	BlockTypeImage BlockType = "image"
+	BlockTypeAudio BlockType = "audio"
+	BlockTypeVideo BlockType = "video"
+	BlockTypeTool  BlockType = "tool"
+	BlockTypeDoc   BlockType = "document"
```
elevran: nit: the constant is named BlockTypeDoc but its string value is "document". The other constants match their names (BlockTypeText = "text", BlockTypeAudio = "audio"). Consider BlockTypeDocument for consistency, or change the value to "doc".

elevran (commented):
Findings are scoped to two type shape decisions that are straightforward to address before downstream parsers start populating these fields, plus two structural concerns.

pkg/epp/framework/interface/requesthandling/types.go
ahg-g: The idea is to have this as an internal representation of the request, all other fields will be removed. 
Do we need tags when we get to that point?

pkg/epp/framework/interface/requesthandling/types.go
zetxqx: `cache_breakpoints` you mean [Cache_control](https://platform.claude.com/docs/en/build-with-claude/prompt-caching), right? [Cache_control](https://platform.claude.com/docs/en/build-with-claude/prompt-caching) is currently not supported by both vlllm and sglang now. Do you think we should support it when the model server really support it?

zetxqx: For reproduce the payload, we still have [Payload](https://github.com/llm-d/llm-d-router/blob/21fa12ea80ddaad5ee5efef79ebc1bdc87cdae76/pkg/epp/framework/interface/requesthandling/types.go#L87) to carry the either raw or parsed to `map[string]any` struct.

pkg/epp/framework/interface/requesthandling/types.go
zetxqx: +1 the json tag is deliberately omitted because we don't want people to `json.Unmarshal` or `Marshal` the InferenceRequestBody

pkg/epp/framework/interface/requesthandling/types.go
zetxqx: currently from rawbytes we only unmarshal it to any level, at [here](https://github.com/llm-d/llm-d-router/blob/21fa12ea80ddaad5ee5efef79ebc1bdc87cdae76/pkg/epp/framework/interface/requesthandling/types.go#L336). And the only consumer is at approximateprefix [producer](https://github.com/llm-d/llm-d-router/blob/21fa12ea80ddaad5ee5efef79ebc1bdc87cdae76/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go#L136-L139) it basically just Marshal it back to rawBytes. So do we really need to define a fine-grained struct here.

pkg/epp/framework/interface/requesthandling/types.go
zetxqx: Removed, this is a dead code I added from different refactoring experimental

pkg/epp/framework/interface/requesthandling/types.go
zetxqx: updated to BlockTypeDocument

zetxqx:
> > @ahg-g I basically proposed a new struct and updated the PR description. Could you take a look to see if it makes more sense?
> 
> This is very neat! one question: is it only `/v1/completions` that has the semantics of a batch of prompts?

`/v1/completions` and `v1/embeddings` both have possibility to have batch of prompts

pkg/epp/framework/interface/requesthandling/types.go
elevran: correct - `cache_control`. We can delay introducing it. The meta point is that we may need to handle protocol specific fields on the IR (which should be agnostic...). Not sure how to resolve these opposing forces.

pkg/epp/framework/interface/requesthandling/types.go
elevran: We don't need tags if it is never serialized and only the input Payload is used in serlaliztion.

ahg-g:
@zetxqx can you please sign correctly?

@elevran anything not yet addressed from your side?

pkg/epp/framework/interface/requesthandling/types.go
elevran: Q (**non blocking for this PR**): do we have a list of the different fields that are being used and the plugins that use them? Asking to understand which fields are needed on the parsed-request.

PR #1387, opened 2026-05-28
Allow skipped requests to go through scheduling profiles

zetxqx:
Currently, skipped requests (e.g. due to unsupported paths) immediately fall back to a random endpoint, bypassing the scheduling director completely. This PR routes skipped requests through the director so they can be scheduled via profiles to optimal endpoints.

### Changes
* **handlers/server**: Route skipped requests through `director.HandleRequest` instead of falling back immediately; mark `RequestState` as `RequestSkipped` at the end. And force skip==true ParseResult setting with RawPayload/
* **server_test**: Add assertions to `TestServer_Skip` to verify `HandleRequest` is called for skipped requests.

### DIFF
--- a/pkg/epp/handlers/server.go
+++ b/pkg/epp/handlers/server.go
@@ -162,10 +162,10 @@ const (
 	// RequestEvicted indicates the request was evicted by flow control.
 	// The state machine sends an ImmediateResponse(429) to the proxy.
 	RequestEvicted StreamRequestState = 8
-	// RequestSkipped indicates the request parsing was skipped.
+	// RequestResponseProcessingSkipped indicates that EPP response-phase stream interception was skipped for this request.
 	// The state machine sends a RequestHeadersResponse and RequestBodyResponse with the routing decision
 	// from the scheduling director to the proxy, and then gracefully closes the stream to stop further external processing.
-	RequestSkipped StreamRequestState = 9
+	RequestResponseProcessingSkipped StreamRequestState = 9
 )
 
 // recvResult holds the result of a srv.Recv() call from the reader goroutine.
@@ -350,7 +350,7 @@ func (s *StreamingServer) Process(srv extProcPb.ExternalProcessor_ProcessServer)
 					break
 				}
 
-				if parseResult != nil && parseResult.Skip {
+				if parseResult != nil && parseResult.SkipResponseProcessing {
 					parseResult.Body = &fwkrh.InferenceRequestBody{
 						Payload: fwkrh.RawPayload(reqCtx.Request.RawBody),
 					}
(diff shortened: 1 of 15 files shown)


liu-cong:
`Skip` means skipping EPP processing, correct? With this change, what's the difference between skipped requests vs others?

zetxqx:
> `Skip` means skipping EPP processing, correct? With this change, what's the difference between skipped requests vs others?

@liu-cong I update the code comment to reflect the current status. Short answer is now `skip=true` will skip response processing completely.

background is: 
1. currently the skip is only used by vertex ai.
2. because vertexai may have different type of gRPC request not all needed intelligent routing so perviously skip will skip whole request processing even for the EPP smart routing plugins. However, with PD setup, the request has to be directed to D node first. So we have to go through the routing system to select the D node instead of random routing to a random node. For non-PD setup, compared with random routing, this still provide load-aware routing, but needs to sacrifice the time spend in EPP processing.

liu-cong:
A few follow up questions:
1. Today we parse a list of known APIs and skip "unknown" APIs. With this change, are we sure that EPP can handle the "unknown" APIs (which means we have many nil fields in InferenceRequestBody)? I think plugins should handle them but need to double check.
2. Let's rename this to `skipResponseProcessing`, if that's what it truly means.

zetxqx:
> 2\. skipResponseProcessing

1. it was tested with vertex. and I double checked the consumption the all use switch case and handled the it correctly.
2. reanmed to skipResponseProcessing and also rename the skipRequest state to RequestResponseProcessingSkipped.

ahg-g:
rebase pls

zetxqx:
@ahg-g  @liu-cong  Rebased, can you take another look?

pkg/epp/handlers/server.go
```diff
@@ -349,13 +350,10 @@ func (s *StreamingServer) Process(srv extProcPb.ExternalProcessor_ProcessServer)
 					break
 				}
 
-				if parseResult.Skip {
-					if err = s.fallbackToRandomEndpoint(ctx, reqCtx, reqCtx.RequestSize); err != nil {
-						logger.Error(err, "Error falling back to random endpoint")
-						break
+				if parseResult != nil && parseResult.SkipResponseProcessing {
+					parseResult.Body = &fwkrh.InferenceRequestBody{
+						Payload: fwkrh.RawPayload(reqCtx.Request.RawBody),
```
liu-cong: why do we need to set payload here? isn't it already there from the ParseRequest call above?

zetxqx: Here I want to force populate the rawBody in case somebody misses populate it. It's just a safe guard.

zetxqx: updated to let parser always populate this when `SkipResponseProcessing` is true

PR #1396, opened 2026-05-29
Session Affinity: Add Hard Affinity Filter and Refactor Scorer

### LINKED ISSUE #815: A session-affinity-filter plugin

A `session-affinity-filter` plugin that returns the endpoint running the session, if no one exists, return all endpoints for the scorers to rank (edited)

**Why is this needed**:

A customer asked for a way to send the request to the endpoint that runs the session, and if one doesn't exist pick an endpoint based on all other configured scorers. The existing session-affinity-scorer doesn't guarantee that.

roytman:
**What type of PR is this?**
/kind feature

Implements stateful (hard and soft) session affinity for the EPP, fixing #815.

- New `session-affinity-filter` plugin (hard affinity): when a session is bound to a candidate endpoint, only that endpoint is returned; otherwise, all candidates pass through.
- `session-affinity-scorer` rewritten to consume the same shared binding (soft affinity): bound endpoint scored 1.0, others 0.0; other scorers still vote.
- `session-id-producer` extended to own the binding store: a TTL-expiring LRU keyed by `SessionID`, populated post-schedule via the `requestcontrol.PreRequest` hook and read pre-schedule into the new `BoundEndpoint` attribute. Configurable `lruSize` (default 1024) and `ttl` (default 30m).
- `BoundEndpoint` is the network identity (`host:port` from `net.JoinHostPort(Address, Port)`), not the K8s `NamespacedName`. Rationale: the KV cache, the binding exists to preserve dies with the inference process, so a restarted pod with a fresh IP correctly fails to match a stale binding and rebinds to a healthy candidate on the next request.
- No HTTP cookie semantics anywhere in the EPP. The `cookieName` source on `session-id-producer` only parses the inbound `Cookie` request header; no `Set-Cookie` is written.

Hash-based / consistent-hashing affinity is deferred per the issue thread.

Fixes #815

```release-note
Adds session affinity filter (hard) and reworks the session affinity scorer (soft); both share a `session-id-producer` that maintains a TTL+LRU bound-endpoint cache keyed by host:port. 
No `Set-Cookie` is written; session identifiers continue to be read from a configured request header or a named cookie.


```


pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity/session_affinity_test.go
```diff
@@ -5,147 +5,295 @@ import (
...
 	)
+}
 
-	inputEndpoints := []scheduling.Endpoint{endpointA, endpointB}
-
-	// valid session token for endpointB
-	validSessionTokenForEndpointB := base64.StdEncoding.EncodeToString([]byte(endpointB.GetMetadata().NamespacedName.String()))
+func TestSessionAffinityScoreWithSessionID(t *testing.T) {
+	s := sessionaffinity.NewSessionAffinity(0, "test-producer")
+	ctx := context.Background()
```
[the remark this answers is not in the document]
roytman: context IS imported at line 4 of the test file.

pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity/session_affinity_test.go
```diff
@@ -5,147 +5,295 @@ import (
...
-			gotScores := sessionAffinityScorer.Score(context.Background(), test.req, test.input)
-
-			if diff := cmp.Diff(test.wantScores, gotScores); diff != "" {
-				t.Errorf("Unexpected output (-want +got): %v", diff)
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			request := &scheduling.InferenceRequest{}
+			if tt.sessionID != "" {
+				// Simulate session-id-producer having set the session ID
+				request.PutAttribute("SessionIDDataKey:test-producer", attrsession.SessionID(tt.sessionID))
```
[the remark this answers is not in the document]
roytman: fixed

pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity/session_affinity.go
```diff
@@ -64,47 +102,116 @@ func (s *SessionAffinity) Category() scheduling.ScorerCategory {
...
-	response.Headers[sessionTokenHeader] = base64.StdEncoding.EncodeToString([]byte(targetPod.NamespacedName.String()))
+	cookie := &http.Cookie{
+		Name:     SessionCookieName,
+		Value:    expectedSessionToken,
+		Path:     "/",
+		HttpOnly: true,
+		Secure:   false,
+		SameSite: http.SameSiteLaxMode,
+		MaxAge:   s.maxAge, // 0 means session cookie
+	}
```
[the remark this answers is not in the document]
roytman: (Secure: false): deliberate choice; TLS terminates at the ingress in k8s, not at the EPP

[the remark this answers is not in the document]
roytman: (Secure: false): deliberate choice; TLS terminates at the ingress in k8s, not at the EPP

pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity/session_affinity.go
```diff
@@ -64,47 +102,116 @@ func (s *SessionAffinity) Category() scheduling.ScorerCategory {
...
+		"sessionToken", expectedSessionToken,
+		"targetEndpoint", targetEndpoint.NamespacedName.String())
+
+	// Append to existing Set-Cookie headers if any
+	existingSetCookie := response.Headers[SetCookieHeaderName]
+	if existingSetCookie != "" {
+		response.Headers[SetCookieHeaderName] = existingSetCookie + "\n" + cookieStr
+	} else {
+		response.Headers[SetCookieHeaderName] = cookieStr
+	}
```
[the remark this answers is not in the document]
roytman: can be a real concern, but fixing it requires changing the Response struct interface, which is out of scope for this PR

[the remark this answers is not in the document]
roytman: can be a real concern, but fixing it requires changing the Response struct interface, which is out of scope for this PR

docs/architecture.md
```diff
@@ -171,10 +193,567 @@ If the configuration is in a file, the EPP command line argument `--configFile`
...
+---
+
+#### SessionAffinity
+
+Scores the candidate pods by giving a higher score to the pods that were previously used for the same session.
+
+- **Type**: `session-affinity-scorer`
+- **Parameters**: 
+  - `maxAge`: the cookie expiration time, in seconds. The default value is 0, which denotes a session cookie (expires when the browser is closed).
+
```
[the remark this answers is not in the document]
roytman: `sessionIDProducerName` is added to `architecture.md`

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/README.md
```diff
@@ -0,0 +1,138 @@
...
+- Applications requiring strict session stickiness for correctness
+- Scenarios where routing to a different endpoint would cause session loss or errors
+
+If you only need **soft affinity** (preference but not requirement), use the `session-affinity-scorer` instead.
+
+## Difference from `session-affinity-scorer`
+
+| Feature | `session-affinity-filter` (this plugin) | `session-affinity-scorer` |
+|---------|----------------------------------------|---------------------------|
+| **Affinity type** | Hard (enforced) | Soft (weighted preference) |
```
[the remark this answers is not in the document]
roytman: False positive

docs/architecture.md
```diff
@@ -47,6 +31,42 @@ The design enables:
...
+[Envoy]:https://www.envoyproxy.io/
+[EPP]:https://gateway-api-inference-extension.sigs.k8s.io/#endpoint-picker
+[BBR]:https://gateway-api-inference-extension.sigs.k8s.io/#concepts-and-definitions
+[GIE]:https://github.com/kubernetes-sigs/gateway-api-inference-extension
+
+---
+
+### Pluggability
+
+![Pluggability Architecture](./images/plugability.png)
```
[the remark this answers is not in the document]
roytman: The real file name is `plugability.png`

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/README.md
```diff
@@ -0,0 +1,138 @@
...
+- **With session ID not matching any candidate**: Returns all endpoints (allows scoring/selection)
+- **Without session ID**: Returns all endpoints (no-op)
+- **Single endpoint in list**: Always returns that endpoint (no filtering needed)
+- **Invalid/malformed session ID**: Returns all endpoints (treats as no session)
+
+## Config
+
+| Parameter | Type | Required | Default | Description |
+|-----------|------|----------|---------|-------------|
+| `sessionIDProducerName` | `string` | No | `""` (uses default producer) | Name of the session-id-producer to consume session data from |
```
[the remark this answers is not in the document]
roytman: False positive

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/plugin_test.go
```diff
@@ -0,0 +1,135 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			request := &scheduling.InferenceRequest{}
+			if tt.sessionID != "" {
+				// Simulate session-id-producer having set the session ID
+				request.PutAttribute(attrsession.SessionIDDataKey.WithNonEmptyProducerName("session-id-producer").String(), attrsession.SessionID(tt.sessionID))
+			}
```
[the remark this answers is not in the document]
roytman: It was fixed.  The default already equals "session-id-producer"

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/plugin_test.go
```diff
@@ -0,0 +1,135 @@
...
+	)
+}
+
+func TestSessionAffinityFilter(t *testing.T) {
+	ctx := context.Background()
+
+	// Create test filter
+	plugin, err := sessionaffinity.Factory("test-filter", nil, nil)
+	assert.NoError(t, err)
+	filter := plugin.(*sessionaffinity.Plugin)
```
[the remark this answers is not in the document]
roytman: It was fixed. The default already equals "session-id-producer"

pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity/README.md
```diff
@@ -1,7 +1,70 @@
...
-**Parameters:** None.
+## What it does
+
+For each request the scorer reads the `SessionID` attribute published by a `session-id-producer`:
+
+- **Session present, endpoint in candidates** — scores that endpoint `1.0`, all others `0.0`.
+- **Session present, endpoint not in candidates** — scores all endpoints `0.0`.
+- **Session absent** — scores all endpoints `0.0`.
+
+On each response the scorer writes an `HttpOnly; SameSite=Lax` `Set-Cookie` header encoding the selected endpoint's `<namespace>/<name>` as base64. The cookie is omitted when the client already carries the correct value.
```
ahg-g: A couple of questions:

1) What is the motivation behind this behavior? why do we need the epp to manage this?
2) Why are we applying this to the scorer but not to the filter? 

note: I am not yet for or against this behavior, I am only trying to understand the "why" at this point.

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,132 @@
...
+	config := Config{}
+	if rawParameters != nil {
+		if err := rawParameters.Decode(&config); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal config: %w", err)
+		}
+	}
+
+	return &Plugin{
+		typedName:      fwkplugin.TypedName{Type: PluginType, Name: name},
+		sessionDataKey: attrsession.SessionIDDataKey.WithNonEmptyProducerName(config.SessionIDProducerName),
```
ahg-g: if parameters are empty, then `config.SessionIDProducerName` will be empty, we need to have a sensible default here

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,132 @@
...
+	// Read session ID from request attributes (populated by session-id-producer)
+	sessionID, ok := p.readSessionID(request)
+	if !ok || sessionID == "" {
+		logger.V(logutil.DEBUG).Info("SessionAffinityFilter: no session ID, keeping all endpoints",
+			"total", len(endpoints))
+		return endpoints
+	}
+
+	// Decode the session ID to get the target endpoint name
+	targetEndpoint := p.decodeSessionID(sessionID)
```
ahg-g: why do we expect the session id to be the endpoint name?

pkg/epp/framework/plugins/README.md
```diff
@@ -2,6 +2,17 @@
 
 This directory contains the available plugins. For detailed information on individual plugins, please refer to the README.md located within each plugin's respective directory. To understand how they integrate into the incoming request processing lifecycle, see the [Endpoint Picker (EPP) design](https://github.com/llm-d/llm-d/tree/main/docs/architecture/core/router/epp) document.
 
+## Session Affinity
```
ahg-g: pls revert this

pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity/README.md
roytman: 1. The EPP is the only component in the `ext_proc` pipeline with access to both the incoming request headers and the outgoing response headers. The session cookie is how a stateless client persists its endpoint binding across requests, so if the EPP does not set it, nothing does. 
On the first request, there is no session yet, so the scorer selects an endpoint normally and writes the binding into the cookie. The client then carries it on all subsequent requests, the `session-id-producer` reads it and publishes the SessionID attribute, and the filter/scorer consumes it.


2. The filter is request-only: its job is to restrict the candidate list based on what session already exists. It has no response-phase hook and does not know which endpoint was ultimately selected. The scorer runs at selection time and also implements `ResponseHeaderProcessor`, so it naturally owns the lifecycle of setting the cookie after a routing decision is made. The filter gates existing sessions; the scorer establishes new ones.

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/plugin.go
roytman: The session ID in this implementation encodes the endpoint name directly (base64(<namespace>/<name>)). This is a deliberate stateless design: no server-side session store or cross-replica coordination is needed. The scorer writes the selected endpoint's identity into the cookie after each request; the filter decodes it on the next request to recover that endpoint. 
The cookie is opaque to the client but carries the routing intent for the EPP. An alternative using an abstract token would require a distributed map from token to endpoint, adding operational complexity.

ahg-g: But the filter plugin doesn't implement this behavior, it is only the scorer that implements the response hook. Is the expectation that the scorer should always be configured when one wants to configure a session affinity filter? If so, that doesn't seem right, why not have the filter also implement the response hook and make that call idempotent in case both the filter and the scorer are configured.

roytman: Good point, I'll update the implementation.

pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity/session_affinity.go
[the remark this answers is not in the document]
roytman: Now, we don't set the HTTP "Set-Cookie" header

pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity/session_affinity.go
```diff
@@ -1,48 +1,75 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package sessionaffinity provides the soft session-affinity scorer. It reads
+// the BoundEndpoint attribute published by the session-binding-producer and
+// gives the bound endpoint a maximum score, leaving everything else at zero.
```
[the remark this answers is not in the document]
roytman: Resolved in this session: removed the duplicate package doc from `session_affinity.go` and corrected the wording in `doc.go`.

pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity/session_affinity.go
```diff
@@ -1,48 +1,75 @@
...
+// Factory builds a SessionAffinity scorer from raw plugin parameters.
+func Factory(name string, parameters *json.Decoder, _ plugin.Handle) (plugin.Plugin, error) {
+	config := Config{}
+	if parameters != nil {
+		if err := parameters.Decode(&config); err != nil {
+			return nil, fmt.Errorf("invalid session affinity config: %w", err)
+		}
+	}
+	return NewSessionAffinity(config.SessionIDProducerName).WithName(name), nil
 }
```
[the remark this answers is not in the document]
roytman: Not changed. Current scorer Factory `session_affinity.go:47` wraps any decode error. The codebase convention is to rely on StrictDecoder returning a nil decoder when raw is empty (see prefix.PrefixCachePluginFactory and activerequest.Factory); the new code follows that convention.

Adding errors.Is(err, io.EOF) would diverge from `prefix`, `activerequest`, `latency`, etc., which all just wrap Decode errors directly. The framework normalizes the empty-params case to a nil decoder upstream, so this branch is unreachable in practice.

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,131 @@
...
+		if err := rawParameters.Decode(&config); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal config: %w", err)
+		}
+	}
+
+	return &Plugin{
+		typedName: fwkplugin.TypedName{Type: PluginType, Name: name},
+		bindingDK: attrsession.BoundEndpointDataKey.WithNonEmptyProducerName(config.SessionIDProducerName),
+	}, nil
+}
```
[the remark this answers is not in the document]
roytman: Not changed. Current scorer Factory session_affinity.go:47 wraps any decode error. The codebase convention is to rely on StrictDecoder returning a nil decoder when raw is empty (see prefix.PrefixCachePluginFactory and activerequest.Factory); the new code follows that convention.

Adding errors.Is(err, io.EOF) would diverge from prefix, activerequest, latency, etc., which all just wrap Decode errors directly. The framework normalizes the empty-params case to a nil decoder upstream, so this branch is unreachable in practice.

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionid/producer.go
```diff
@@ -112,10 +168,36 @@ func (p *Producer) Produce(_ context.Context, request *fwksched.InferenceRequest
...
+		return
+	}
+	target, ok := primaryTarget(schedulingResult)
+	if !ok {
+		logger.Info("session-id-producer: no primary target endpoint to bind", "sessionID", id)
+		return
+	}
+	p.bindings.Add(attrsession.SessionID(id), target)
+	logger.Info("session-id-producer: bound session", "sessionID", id, "endpoint", target)
+}
```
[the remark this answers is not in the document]
roytman: Lines 209 and 213 of `producer.go` still log the full sessionID. Already gated by V(logutil.DEBUG), but Copilot's concern (sessionIDs can correlate to users) is reasonable. 

Judgment call.
Two reasonable options. 

- Keep as-is; sessionID is genuinely useful at DEBUG and is gated. 
- Log a short prefix only (e.g. first 8 chars) so operators can correlate without exposing the full identifier.

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionid/producer_test.go
```diff
@@ -216,7 +225,152 @@ func TestProduces_DeclaresKey(t *testing.T) {
...
+	time.Sleep(120 * time.Millisecond)
+
+	lookup := requestWithSession("session-A")
+	require.NoError(t, producer.Produce(context.Background(), lookup, nil))
+	_, ok := fwksched.ReadRequestAttribute[attrsession.BoundEndpoint](
+		lookup,
+		attrsession.BoundEndpointDataKey.WithNonEmptyProducerName("session-id-producer").String(),
+	)
+	assert.False(t, ok, "binding should have expired")
+}
```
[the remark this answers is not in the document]
roytman: `require.Eventually` does not fit here because Produce refreshes the TTL on a successful read, so a polling loop that calls Produce to observe expiration would keep the entry alive and the test would never see the binding go away.

Took the second option and widened the margins. TestBindingExpiresAfterTTL now uses ttl=100ms with a 1s sleep (~10x margin, was ~1.4x). The companion TestBindingRefreshedOnProduce uses ttl=1s with two 600ms sleeps so a single step has a 400ms margin under the TTL while the two together exceed it: that test passes only if the intermediate Produce actually refreshed the entry. Comments in both tests note why polling is not an option.

Total package wall-clock with the new tests is ~2.3s.

roytman:
**Edge case worth flagging:** 
if an operator declares two `session-id-producer` instances (different names) and points the filter at one and the scorer at the other, each producer would maintain its own private LRU and write its own binding on every PreRequest. That's wasteful (2× storage, 2× writes per request) and the two caches could diverge if someone reconfigures one. Not broken, just redundant. Same family of "stupid but possible" config as enabling both filter and scorer at once.

The framework exposes Handle.GetAllPluginsWithNames(), so we can detect it.  There are the following 3 options to handle it, ordered by intrusiveness:

1. Doc only (current). A line in each README saying "configuring both is redundant; the filter dominates the scorer." Cheapest. No code. Operators who read docs are warned; operators who don't read docs aren't.

2. Soft warning at startup. In each plugin's factory, scan handle.GetAllPluginsWithNames() and log a warning if the sibling type is also registered. Caveats:
    - Order-dependent: only one of the two will see the other (the second-instantiated). The first plugin's factory runs before the second exists. So we'd put the check in only one of them, by convention. Still works.
     - New pattern: no other in-tree plugin currently inspects Handle for sibling-detection. We'd be the first.
     - Cost: ~10 lines in one factory.
3. Hard reject at startup. Same as (2) but error instead of warn. Risks rejecting a config the operator deliberately chose for some reason we didn't anticipate. The framework lets you stack mutually-redundant plugins everywhere else in the codebase (e.g. multiple scorers can disagree); this would be the first hard cross-plugin rule.

Other plugin pairs with similar redundancies (e.g., configuring two scorers with overlapping concerns or two filters that intersect) are not policed today. Adding a special case just here is asymmetric.

In the meantime, I stay with the first option.

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionid/producer.go
```diff
@@ -82,11 +127,33 @@ func Factory(name string, rawParameters *json.Decoder, _ fwkplugin.Handle) (fwkp
 		return nil, fmt.Errorf("'%s' requires exactly one of headerName or cookieName to be set, not both", SessionIDProducerType)
 	}
 
+	size := defaultLRUSize
+	if params.LRUSize != 0 {
```
elevran: nit: `lruSize: 0` in user config is accepted and silently treated as the default (1024) rather than returning an error or logging a message.
Change the field to `*int` so nil means unconfigured and zero is a validatable explicit value, then reject zero with the same error produced for negative values?

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionid/producer.go
```diff
@@ -112,10 +183,48 @@ func (p *Producer) Produce(_ context.Context, request *fwksched.InferenceRequest
...
+	if id == "" {
+		return
+	}
+	target, ok := primaryTarget(schedulingResult)
+	if !ok {
+		logger.Info("session-id-producer: no primary target endpoint to bind", "sessionID", id)
+		return
+	}
+	p.bindings.Add(attrsession.SessionID(id), target)
+	logger.Info("session-id-producer: bound session", "sessionID", id, "endpoint", target)
```
elevran: this log line writes the raw client-supplied session ID. Sessions can hold sensitive information or be arbitrarily long. Suggest replacing session id with a fixed-length hash prefix. The `"no primary target"` log a few lines above has the same issue and should receive the same treatment.

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionid/producer.go
```diff
@@ -112,10 +183,48 @@ func (p *Producer) Produce(_ context.Context, request *fwksched.InferenceRequest
...
+}
+
 func (p *Producer) extract(request *fwksched.InferenceRequest) string {
```
elevran, line 253: nit: `extract` returns the raw trimmed header or cookie value with no length cap which can impact memory usage. Can either rely on gateway to reject large cookies, hash the cookie / session to a fixed sized string or trim the session string to a some value (e.g., 1024B)

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/README.md
```diff
@@ -0,0 +1,81 @@
...
+## Difference from session-affinity-scorer
+
+Both plugins consume the same `BoundEndpoint` attribute but differ in how strictly they enforce stickiness:
+
+| | `session-affinity-filter` | `session-affinity-scorer` |
+|---|---|---|
+| **Guarantee** | Hard, only the bound endpoint is returned when present | Soft, bound endpoint gets the highest score, others remain |
+| **On endpoint unavailability** | Falls back to the full candidate set | Other scorers determine the winner |
+
+Use the filter when strict stickiness is required for correctness. Use the scorer when a best-effort preference is sufficient and other scorers should still get a vote. Configuring both at once is redundant: the filter dominates the scorer in every case, so the extra scoring work changes no outcomes.
```
elevran: "Use the filter when strict stickiness is required for correctness" sets an expectation that the filter guarantees routing under all conditions. But when the bound endpoint is absent from the candidate list, the filter returns all candidates unchanged and any pod can serve the request. 
Probably best to remove the "correctness" framing here (reframe as "preferred / sticky routing when the bound pod is available").
Do **not** think we need to add an `onStaleBound: passthrough|error` config option so operators with strict requirements can opt into an error on miss.

pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity/README.md
```diff
@@ -1,7 +1,80 @@
...
+## Difference from session-affinity-filter
+
+Both plugins consume the same `BoundEndpoint` attribute but differ in how strictly they enforce stickiness:
+
+| | `session-affinity-scorer` | `session-affinity-filter` |
+|---|---|---|
+| **Guarantee** | Soft, bound endpoint gets the highest score, others remain candidates | Hard, only the bound endpoint is returned when present |
+| **On endpoint unavailability** | Other scorers determine the winner | Falls back to the full candidate set |
+
+Use the scorer when a best-effort preference is sufficient and other scorers should still get a vote. Use the filter when strict stickiness is required for correctness. Configuring both at once is redundant: the filter dominates the scorer in every case, so the extra scoring work changes no outcomes.
```
elevran: The two paragraphs here ("Configuring both at once is redundant..." and "Pairing the scorer and filter with different...") are near-verbatim copies of the same text in the filter README. 

Suggest removing both paragraphs from this file and replace with a single cross-reference: "See [Session Affinity Filter](../../filter/sessionaffinity/README.md#difference-from-session-affinity-scorer) for guidance on pairing both plugins."

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,145 @@
...
+
+// TypedName returns the typed name of the plugin.
```
elevran: nit: `// TypedName returns the typed name of the plugin.` restates what the function name already says. Remove it.

elevran (commented):
Good implementation overall. Will let @liu-cong, who's assigned, LGTM/approve.

**Test gaps (no specific line):**
- Neither the filter nor the scorer has a `Consumes()` unit test. That is the contract the framework uses for startup wiring; a regression there would be silent until runtime. Add a `TestConsumes` for each that verifies the returned `Required` map contains the correct `BoundEndpointDataKey`.
- `EndpointBoundForm` is a new exported function and the single shared definition of endpoint identity, but it has no direct unit tests. Add a table test covering nil endpoint, nil metadata, empty address, empty port, and the happy path.

liu-cong:
Adding the LRU makes the router stateful. Not against having more stateful features but I think we should offer a stateless option, that enables high horizontal scalability. Perhaps we can introduce another header `x-session-expiry`, the router will auto extend the expiry with a configured ttl (or even and an optional `x-session-ttl` header).

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionid/producer.go
roytman: `Parameters.LRUSize` is now *int. 
- Nil = unconfigured (default 1024). 
- Explicit non-positive values (0 or negative) are rejected with the same error. 
- New "zero lru size" rejection test in TestFactory_Validation

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionid/producer.go
roytman: We use the hash of the header/cookie value as the key to a fixed sized string and we rely on the gateway to reject large cookies

pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity/README.md
roytman: Replaced with a single cross-reference to the filter README's #difference-from-session-affinity-scorer anchor.

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/plugin.go
roytman: done

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionid/producer.go
roytman: Both `PreRequest` log lines now emit sessionIDHash (the first 8 hex chars of sha256(id)) via the new `hashSessionIDForLog` helper

roytman:
> Good implementation overall. Will let @liu-cong, who's assigned, LGTM/approve.
> 
> **Test gaps (no specific line):**
> 
>     * Neither the filter nor the scorer has a `Consumes()` unit test. That is the contract the framework uses for startup wiring; a regression there would be silent until runtime. Add a `TestConsumes` for each that verifies the returned `Required` map contains the correct `BoundEndpointDataKey`.
> 
>     * `EndpointBoundForm` is a new exported function and the single shared definition of endpoint identity, but it has no direct unit tests. Add a table test covering nil endpoint, nil metadata, empty address, empty port, and the happy path.

- TestEndpointBoundForm in `pkg/epp/framework/plugins/datalayer/attribute/session/data_types_test.go` -  table test with 6 cases: `nil` endpoint, `nil` metadata, empty address, empty port, IPv4 happy path, IPv6 wrapped-in-brackets. Plus a separate `TestEndpointBoundForm_RealEndpoint` that runs against the framework's actual `NewEndpoint` constructor, so the helper can't drift away from the real Endpoint shape.
- `TestConsumes` in the filter test - asserts `BoundEndpointDataKey.WithNonEmptyProducerName(testProducerName)` is present in Required, value type is BoundEndpoint, and Optional is empty.
- `TestConsumes` in the scorer test - same shape.

roytman:
I'm still waiting to finalize session-affinity requirements.

elevran:
@liu-cong - any progress on validating/finalizing the requirements?
@roytman - is there value in keeping this around (and fixing conflicts) or close and reopen a new PR once requirements are well understood? I would opt to close for now.

liu-cong:
I tried to capture the requirements in  https://github.com/llm-d/llm-d-router/issues/2019

PR #1397, opened 2026-05-29
Make multimodal cache affinity use optional token lengths

guygir:
## Summary

Reworks `mm-embeddings-cache-producer` so encoder-cache affinity can use tokenized multimodal placeholder lengths when tokenization is available, without requiring or auto-creating `token-producer`.

There is **no** `weightMode` toggle. Weighting is automatic:

1. If `TokenizedRequest` multimodal features are present (typically from a configured `token-producer`, including the `estimate` backend), each unique media hash uses `MultiModalFeature.Length` as `MatchItem.Size` (falling back to `1` when length is zero).
2. If tokenized features are absent, the producer falls back to unit weight `1` per unique hash from:
   - typed OpenAI chat-completions structured media blocks, or
   - `Generate.Features.MMHashes` when present.

`mm-embeddings-cache-scorer` is unchanged and still scores `matchedWeight / totalWeight`.

## Changes

- Declare `TokenizedRequest` as an **optional** data dependency (no auto-create of `token-producer`).
- Prefer `TokenizedRequest.Prompts[].MultiModalFeatures` lengths for item sizes.
- Add chat / generate unit-weight fallbacks when tokenized features are unavailable.
- Update producer and scorer READMEs.
- Add sample config: `deploy/config/epp-weighted-mm-embeddings-cache-config.yaml`.

## Validation

- `go test ./pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/...`
- `go test ./pkg/epp/framework/plugins/scheduling/scorer/mmcacheaffinity/...`
- `go test ./pkg/epp/datalayer/...`

Cluster A/B re-validation against current main is still pending (stack currently in use). Earlier VisionArena runs showed length-weighted affinity helping under higher load; unit-weight often won at low RPS where render/token overhead dominated.

Follow-up to #901. Builds on upstream `DataDependencies.Optional` (no new framework hook in this tip).

```release-note
mm-embeddings-cache-producer now optionally uses TokenizedRequest multimodal placeholder lengths as MatchItem sizes when token-producer is present; otherwise falls back to unit-weight hashes without auto-creating tokenization.
```


pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/README.md
```diff
@@ -8,8 +8,10 @@ Produces multimodal embeddings cache match data for downstream scheduling plugin
 
 For each request, the producer extracts stable multimodal item hashes from:
 
-- `TokenizedPrompt.MultiModalFeatures`, when a `token-producer` is configured
+- `TokenizedPrompt.MultiModalFeatures`, when weighted mode is enabled and
+  `token-producer` is configured
 - typed OpenAI chat-completions structured media blocks, as a lightweight fallback
+- `Generate.Features.MMHashes`, when present
 
 It keeps an in-memory LRU map from multimodal hash to the set of pods that recently
```
ahg-g, line 16: pls use the term endpoint instead of pod

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/README.md
```diff
@@ -18,15 +20,25 @@ same image, video, or audio input.
 
 Repeated references to the same multimodal hash within one request count once.
 
+## Weight Modes
```
ahg-g: it is not clear to me what this knob does from this docs, please add a short description before presenting the options.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -47,6 +48,11 @@ const (
 
 	defaultCacheSize   = 10000
 	podCleanupInterval = 2 * time.Minute
+
+	// WeightModeUnweighted counts each unique multimodal hash once with unit size.
```
ahg-g: it is not clear to me what the phrase "with unit size" means

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/README.md
guygir: Done, also aligned package/type comments in producer.go.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/README.md
guygir: Added an intro under Weight Modes.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
guygir: Reworded to "item weight 1" in the const comment and matched that terminology in the README/scorer docs to make it more clear.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -147,6 +167,15 @@ func (p *Producer) Produces() map[plugin.DataKey]any {
 	return map[plugin.DataKey]any{p.dk: attrmm.EncoderCacheMatchInfo{}}
 }
 
+// Consumes declares the token-producer dependency only for weighted mode so
+// unweighted configs are not forced to pull in tokenization.
+func (p *Producer) Consumes() map[plugin.DataKey]any {
+	if !p.useWeighted {
+		return nil
+	}
+	return map[plugin.DataKey]any{tokenproducer.TokenizedPromptDataKey: scheduling.TokenizedPrompt{}}
```
ahg-g: We can declare this as optional, when available we use the length as produced by the token-producer, when not we fall back to 1. No need to have an explicit WeightMode

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -147,6 +167,15 @@ func (p *Producer) Produces() map[plugin.DataKey]any {
 	return map[plugin.DataKey]any{p.dk: attrmm.EncoderCacheMatchInfo{}}
 }
 
+// Consumes declares the token-producer dependency only for weighted mode so
```
capri-xiyue: What will happen is token-producer is declared but with `estimate` backend? I assume the weighted model will still work as expected?

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/README.md
```diff
@@ -78,9 +99,15 @@ schedulingProfiles:
...
+
 ## Operational Notes
 
 - The cache is a best-effort routing signal, not a correctness dependency.
-- The producer remains tokenizer-free for request shapes where typed media blocks are
-  sufficient; `token-producer` is only required when relying on upstream multimodal
-  metadata.
+- `cacheSize` bounds the EPP routing LRU only; it is not a model-server encoder
+  cache capacity knob.
+- Unweighted mode remains lightweight: structured chat media blocks are enough and
```
capri-xiyue: I think token producer with `estimate` backend is also lightweight. Should the default behavior be use weighted when tokenproducer is available?

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -215,17 +253,28 @@ func itemsFromGenerateFeatures(mmHashes map[string][]string) []attrmm.MatchItem
...
 }
 
+func addTokenizedItem(itemsByHash map[string]attrmm.MatchItem, feature fwkrh.MultiModalFeature, useWeighted bool) {
+	if feature.Hash == "" {
+		return
+	}
+	if _, exists := itemsByHash[feature.Hash]; exists {
+		return
+	}
+	size := 1
```
capri-xiyue: Is `weight` a better name?

ahg-g:
@guygir are you able to address the comments? code freeze is approaching

guygir:
> @guygir are you able to address the comments? code freeze is approaching

Yes, on it. Just made some changes - will push the latest version and address all comments by tomorrow morning.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
guygir: Agreed. I removed the explicit weightMode and implemented this as an optional data dependency instead.

I added the small framework primitive because using lengths "when available" needs ordering to be deterministic. Without an optional DAG edge, the MM producer can run before token-producer and silently fall back to 1 even though tokenization is configured and weights DO exist.

With this change, if token-producer is configured, mm-embeddings-cache-producer runs after it and uses TokenizedPrompt.MultiModalFeatures.Length. If token-producer is absent, nothing is auto-created and the producer falls back to unit weight 1.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
guygir: Yes, a configured token-producer backend, including estimate, runs before the MM producer through the optional DAG edge, which uses the populated MultiModalFeature.Length values - precise or estimated.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/README.md
guygir: That makes sense, yep. Default behavior is automatic now: when token-producer is present, we use the tokenized multimodal lengths, whether estimated or precise. When it is absent, we keep the lightweight unit-weight fallback.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
guygir: Indeed, changed the local variable name to weight where it represents scoring contribution.

guygir:
Thanks @ahg-g and @capri-xiyue for the reviews, all addressed in the latest commit.

The main change is that mm-embeddings-cache-producer now treats TokenizedPrompt as an optional data dependency: if token-producer is configured, the data graph orders the MM producer after it and the producer uses MultiModalFeature.Length. if token-producer is not configured, nothing is auto-created and the producer falls back to unit weight 1 per hash.

I added a small OptionalConsumerPlugin framework hook because the alternative was not deterministic. Without a DAG edge, mm-embeddings-cache-producer can run before token-producer and silently fall back to unit weights even when tokenization is configured. A required Consumes() edge would fix ordering but would also require or auto-create token-producer, which is what we wanted to avoid in the first place...

Since this commit included a small framework change, I ran additional validations:

go test ./pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/...
go test ./pkg/epp/framework/plugins/scheduling/scorer/mmcacheaffinity/...
go test ./pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/...
go test ./pkg/epp/datalayer/...
go test ./pkg/epp/framework/interface/plugin/...

all passed on my end.

ahg-g:
Thanks @guygir , please rebase.

pkg/epp/datalayer/data_graph.go
```diff
@@ -41,15 +41,19 @@ func ValidateAndOrderDataDependencies(plugins []plugin.Plugin) ([]string, error)
 	}
 	producers := make(map[string]plugin.ProducerPlugin)
 	consumers := make(map[string]plugin.ConsumerPlugin)
+	optionalConsumers := make(map[string]plugin.OptionalConsumerPlugin)
```
ahg-g: I assume this is not intended will go aware after a rebase?

guygir: Correct

guygir:
Rebased onto current main and switched to upstream DataDependencies.Optional. kept the optional DAG ordering behavior and validated fallback,tokenized-weighted paths with tests

deploy/config/epp-weighted-mm-embeddings-cache-config.yaml
```diff
@@ -0,0 +1,47 @@
...
+dataLayer:
+  sources:
+    - pluginRef: metrics-data-source
+      extractors:
+        - pluginRef: core-metrics-extractor
+    - pluginRef: endpoint-notification-source
+      extractors:
+        - pluginRef: mm-embeddings-cache-producer
+        - pluginRef: precise-prefix-cache-scorer
+schedulingProfiles:
```
vMaroon, line 39: How were the weights tuned? Why not extend the current defaults (optimized baseline) with the `mm-embeddings-cache-scorer`?

On a different note, tuning weights is very tough, we should migrate examples to the new approach.

This can be followed up in a different PR.

guygir: They were tuned according to the mm benchmarks, but I agree it’s clearer for this sample to extend the current default/optimized baseline and add mm affinity (conservatively), at least for now.

capri-xiyue:
Hi @guygir, could you fix the CI/CD including lint and test?

ahg-g:
Is this PR still relevant?

guygir:
> Is this PR still relevant?

Yes, should still be relevant.

main still always uses item weight 1, and this PR uses tokenized multimodal placeholder lengths as item weights (when available).

Rewrote onto current main.

guygir:
Toy comparison vs main:

**Stack:** 3× `Qwen/Qwen2.5-VL-7B-Instruct-AWQ`. Both used the same sole-scorer config (`mm-embeddings-cache-scorer` + `token-producer` with the `estimate` backend). only the EPP image changed - `origin/main` for unit weight vs this PR for length-weighted sizes.

**Workload:** 240 chat requests at 10 RPS for 24s, max_tokens=1, no warmup. Each request has a short unique text prompt plus 10 images (9×32×32 tiny + 1×1536×1536 large). Images come from 15 sets (4 large + 36 tiny per set, each large image repeats 4 times), so large-item reuse should matter more under length weighting than under unit weight (and no kv cache prefix hits since the prompt is unique)

**Results:**

| Path | Actual RPS | p50 | p95 | Avg |
|------|----------:|----:|----:|----:|
| unit (`main`) | 9.19 | 1.667s | 2.730s | 1.644s |
| length (this PR) | 9.26 | 1.224s | 2.002s | 1.239s |

About 25% lower avg latency and p95 on this mixed-size workload.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -295,25 +300,104 @@ func (p *Producer) Produce(ctx context.Context, request *scheduling.InferenceReq
...
+	return itemSlice(itemsByHash)
+}
+
+func addBlockItem(itemsByHash map[string]attrmm.MatchItem, block fwkrh.ContentBlock) {
+	switch {
+	case block.ImageURL.URL != "":
+		addItem(itemsByHash, contentHash("image_url", block.ImageURL.URL), string(fwkrh.ModalityImage))
+	case block.VideoURL.URL != "":
+		addItem(itemsByHash, contentHash("video_url", block.VideoURL.URL), string(fwkrh.ModalityVideo))
+	case block.InputAudio.Data != "":
```
capri-xiyue, line 386: I think for audio, we support two format?

capri-xiyue:
@guygir Could you fix the lint? It is still failing now

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
guygir: Wired audio_url in the multimodal chat unit-weight fallback too (same URL hashing as image/video). Unit test covers both now and are green.

guygir:
> @guygir Could you fix the lint? It is still failing now

Lint is green on the current PR head (unit/e2e too).

The earlier red lint (while commits were still signed) wasn’t this change - the CI builder image failed (sum.golang.org / ghcr.io/llm-d/llm-d-builder:dev), so make build exited before golangci ran.

latest commits aren’t SSH-signed yet, I’ll re-sign and push shortly.

PR #1404, opened 2026-05-30
fix(datalayer): dispatch endpoint update notifications

### LINKED ISSUE #1403: Endpoint metadata updates do not trigger endpoint notification events
## Problem

Endpoint metadata updates are applied in the datastore, but existing endpoints do not emit an endpoint lifecycle notification.

Today, when an endpoint is first added, the datastore calls `Runtime.NewEndpoint()`, and the runtime dispatches `EventAddOrUpdate` through `endpoint-notification-source`. However, when the same endpoint already exists and only its metadata changes, the datastore only calls `Endpoint.UpdateMetadata(meta)` and returns without dispatching an event.

## Impact

Plugins that rely on `endpoint-notification-source` can miss endpoint updates. For example, precise prefix cache components use endpoint add/update events to install or refresh per-endpoint KV-events subscribers. If endpoint metadata such as address, labels, rank, or ports changes without a notification, plugin state can remain stale.

## Expected behavior

Existing endpoint metadata changes should dispatch `EventAddOrUpdate`, just like endpoint additions, so endpoint-aware plugins can refresh their per-endpoint state.

## Suggested fix

When the datastore sees an existing endpoint with changed metadata:

- update the endpoint metadata
- dispatch an endpoint update event through the runtime
- avoid dispatching when the new metadata is identical to the current metadata

Image

zhouyou9505:
## Summary

Fixes #1403.

Dispatch endpoint update notifications when metadata changes on an already known endpoint.

Previously, the datastore only dispatched endpoint lifecycle events when a new endpoint was created through `Runtime.NewEndpoint()`. If the endpoint already existed, the datastore updated the metadata in place and returned without notifying `endpoint-notification-source` consumers.

## Changes

- Add an optional `EndpointUpdater` hook for endpoint factories that can dispatch update events.
- Add `Runtime.UpdateEndpoint()` to dispatch `EventAddOrUpdate` for existing endpoints.
- Update datastore endpoint upsert logic to detect metadata changes before calling `UpdateMetadata()`.
- Avoid dispatching update events for no-op metadata reconciles.
- Add regression coverage for datastore update notifications and runtime update dispatch.

## Why this matters

Endpoint-aware plugins, such as precise prefix cache components, rely on endpoint add/update events to refresh per-endpoint state like KV-events subscribers. Without an update notification, endpoint metadata changes such as address, labels, rank, or port changes can leave plugin state stale.

## Validation

- `go mod tidy`
- `go test -count=1 ./pkg/epp/datalayer ./pkg/epp/datastore`
- `go test -count=1 ./pkg/epp/...`


pkg/epp/datalayer/factory.go
```diff
@@ -52,3 +52,9 @@ type EndpointFactory interface {
 	NewEndpoint(parent context.Context, inEndpointMetadata *fwkdl.EndpointMetadata, poolinfo PoolInfo) fwkdl.Endpoint
 	ReleaseEndpoint(ep fwkdl.Endpoint)
 }
+
+// EndpointUpdater is optionally implemented by EndpointFactory implementations
```
ahg-g: Why optionally? shouldn't we make this mandatory?

ahg-g:
/assign @elevran 

looks good to me, but will leave it to Etai to approve

pkg/epp/datalayer/factory.go
zhouyou9505: Thanks, that makes sense.

I initially made it optional to minimize the impact on existing implementations. I agree that making `UpdateEndpoint` mandatory makes the `EndpointFactory` (new,update,retire) lifecycle complete and consistent. I’ve updated the PR accordingly.

pkg/epp/datastore/datastore.go
```diff
@@ -399,7 +400,12 @@ func (ds *datastore) upsertEndpoint(meta *fwkdl.EndpointMetadata) bool {
 		ds.pods.Store(meta.NamespacedName, ep)
 		return true
 	}
-	existing.(fwkdl.Endpoint).UpdateMetadata(meta)
+	ep := existing.(fwkdl.Endpoint)
+	if reflect.DeepEqual(ep.GetMetadata(), meta) {
```
elevran: what's the downside of sending a notification anyway and letting the callee decide if to handle it or not?

elevran:
Agree that "update" completes the interface. 
The current design would miss change in labels, for example, which this PR would enable. However, Endpints store additional attributes (e.g., models hosted, metrics updated, etc.) which are not covered even after this change and not sure it's feasible to cover all paths.

Can you please provide some concrere information on which metadata updates are relevant and possible without the deletion and creation of a new endpoint?
For example, neither IP address nor port can change (AFAIK) without the endpoint removed and a new one recreated. This is the case for k8s and the file based discovery.

pkg/epp/datalayer/runtime.go
```diff
@@ -415,6 +415,19 @@ func (r *Runtime) ReleaseEndpoint(ep fwkdl.Endpoint) {
 	}
 }
 
+// UpdateEndpoint dispatches an add/update lifecycle event for an existing endpoint.
+func (r *Runtime) UpdateEndpoint(ctx context.Context, ep fwkdl.Endpoint) {
+	if ep == nil || ep.GetMetadata() == nil {
```
elevran: nit: The `if ep == nil || ep.GetMetadata() == nil { return }` and `if ctx == nil` guards do not match the existing pattern in `NewEndpoint` and `ReleaseEndpoint` in this file, which trust their callers. The only caller here is `datastore.upsertEndpoint`, which has already loaded `ep` from `ds.pods`; it cannot be nil at this point. 
Suggest dropping the guards to match the surrounding code.

pkg/epp/datastore/datastore.go
elevran: `reflect.DeepEqual` works for the common case but treats a nil `Labels` map and an empty `map[string]string{}` as unequal, which can fire spurious update events during reconciliation if upstream sometimes hands us nil and sometimes an empty map. It is also more reflection than this struct needs.

Consider a small `Equal(*EndpointMetadata) bool` method on `EndpointMetadata`, alongside `Clone()`, comparing fields explicitly and using `maps.Equal` for `Labels` (which treats nil and empty map as equal).

zhouyou9505: Thanks for the reminder, that's a good point. datastore.go should not own this decision logic.
Do you think we should still add explicit guards downstream, or is this enough for now?

elevran (commented):
Thanks for the fix. The missing update notification is a potential gap, and the test coverage matches the issue's expected behavior (no event on create, event on metadata change, no event on no-op reconcile).

A few non-blocking observations on the API shape and metadata equality. None block merge from my side; happy to approve once we discuss the points raised above and below.

pkg/epp/datastore/datastore.go
elevran: downstream shuold make the distinction - if it makes sense for its use case (e.g., which fields matter, the cost of "reacting" to change, etc.)

zhouyou9505: Got it, that makes sense. `a small Equal` reduces the impact of no-op reconciles by avoiding unnecessary update events when the endpoint metadata is semantically unchanged.

zhouyou9505:
> Thanks for the fix. The missing update notification is a potential gap, and the test coverage matches the issue's expected behavior (no event on create, event on metadata change, no event on no-op reconcile).
> 
> A few non-blocking observations on the API shape and metadata equality. None block merge from my side; happy to approve once we discuss the points raised above and below.

Thanks, done. I added EndpointMetadata.Equal next to Clone, and datastore now uses it before dispatching UpdateEndpoint.

For concrete same-key metadata updates in the current code:
- Kubernetes: Pod label updates update `Labels` in-place. InferencePool targetPorts resync updates `Port` and `MetricsHost` for an existing pod. `PodIP replacement happens through delete + add, not an in-place update`.
- File discovery: reloads with the same namespace/name update `Address`, `Port`, `MetricsHost`, and `Labels` in-place.
- NamespacedName, PodName, and RankIndex are identity/consistency fields.

elevran:
> For concrete same-key metadata updates in the current code:
> 
> * Kubernetes: Pod label updates update `Labels` in-place. InferencePool targetPorts resync updates `Port` and `MetricsHost` for an existing pod. `PodIP replacement happens through delete + add, not an in-place update`.
> * File discovery: reloads with the same namespace/name update `Address`, `Port`, `MetricsHost`, and `Labels` in-place.

Updating Ports on the InferencePool without a deployement update sounds like a foot gun to me, but the code does allow it.

PR #1405, opened 2026-05-30
disagg: add test coverage for epd test

namgyu-youn:
**What type of PR is this?**
/kind test

Adds two tests for E/PD behaviors:

1. **Multiple encoder endpoints** — E/PD deployments can have a pool of encoder pods. `PreRequest` writes all selected encoder addresses as a comma-joined value in `x-encoder-endpoints`. This test confirms the header is correctly formatted when more than one encoder pod is returned by the scheduler.

2. **Encoder pool returns no candidates** — when the encode profile runs but finds no available pods (empty `TargetEndpoints`), `ProcessResults` still includes the result in `ProfileResults`. The downstream `PreRequest` then skips writing the encode header. This test pins that contract so a future refactor doesn't accidentally drop the result or error out.

**Which issue(s) this PR fixes**: related to #608

```release-note
NONE

### CODE REVIEW

File: pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler_test.go [hunk 1]
```diff
@@ -799,6 +799,18 @@ func TestHandler_ProcessResults_EPD(t *testing.T) {
 				assert.NotContains(t, res.ProfileResults, defaultEncodeProfile)
 			},
 		},
+		{
+			name: "encode ran but returned 0 endpoints → included in results",
```
elevran: nit: test case name uses a unicode arrow. Project convention (CLAUDE.md) avoids unicode symbols. Suggest: `"encode ran but returned 0 endpoints - included in results"`.

elevran (commented):
One style nit.
/lgtm
waiting on passing tests (could be GH issue)

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler_test.go
namgyu-youn: updated, thanks for pointing it out!

elevran (approved):

PR #1408, opened 2026-05-30
feat: video token estimation formula for prefix caching

namgyu-youn:
**What type of PR is this?**:
/kind feature

Adds video token estimation formula for the approximate prefix-cache scorer, mirroring vLLM's Qwen3-VL video pipeline: frame sampling against the 32-frame cap, `smart_resize` to a per-frame pixel budget, temporal patch grouping, and per-frame timestamp token overhead.

When the video is a base64-encoded MP4 payload, duration, fps, and resolution are read directly from the container's moov box metadata. **Otherwise the estimator falls back to model-typical defaults (16s, 2fps, 640x360).** All other parameters (patch size, merge size, temporal patch, max pixels) are fixed architecture constants shared with the image estimator's TokensPerFrame-equivalent, not configuration — they describe the model, not the deployment, so there's nothing to tune per-EPP-instance.

Also fixes the image token estimation formula.

**Note**: This PR scoped to Qwen3-VL model and Video-MME dataset support because: (1) each model has its own metadata inside vLLM, and (2) each dataset has distinct parsing rules needed.

**Which issue(s) this PR fixes:** #1746

**Release note:**
```
Add video token estimation in approximate prefix cache; fix image token estimation formula.
```


pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
```diff
@@ -161,6 +163,8 @@ var defaultMultimodalConfig = multiModalTokenEstimatorConfig{
 			Factor: 1024,
 		},
 	},
+	Audio: &fixedTokenEstimatorConfig{FixedToken: 256},
+	Video: &fixedTokenEstimatorConfig{FixedToken: 1000},
```
capri-xiyue: I checked how vllm usually handle it. Basically video is a frame of images. You need to guess the number of frames and then if the model use fixed token for images, you can use the `number of frame * image fixed token`. Basically, currently your estimator config is not based on how model server calculate the placeholder tokens. Can you investigate how model server evaluate the number of audio and video placeholder tokens and then have the mimic logic in llmd? If it doesn't guess the number of placeholders based on the real model server calculation placeholder logic, it won't have better cache eviction logic when compared to  using raw bytes to the pseudo-token stream as both of them are bad estimations

capri-xiyue: For example, for qwen3 vl video frame, number of frame =min(duration * fps, max frame)

The token estimation logic for video is more complex than your current PR

namgyu-youn: Thanks so much for the detailed review and roadmap sharing. I updated the PR to focus on video only
(not audio) to keep the scope small — this PR covers Phase 2 (video token estimation) of the broader multimodal prefix-cache roadmap, not the full work.

For your question on hashing logic, here is how vLLM handles video placeholder tokens:

### 01. number of frames

vLLM downloads the video and applies:

$$\max\left(1, \min\left(N_u,\ N_f,\ \left\lfloor T \cdot f \right\rfloor\right)\right)$$

where $N_u$ = `user_num_frames`, $N_f$ = `total_frames_in_file`, $T$ = `duration`, $f$ = `fps`.

Default is 32 when no override is given. ref: https://github.com/vllm-project/vllm/blob/0b25cf441903b06811e0f75f09c303c18fe08240/vllm/multimodal/video.py#L446-L470

### 02. tokens per frame

This is model-specific — each model family uses a different formula that depends on model config
params (`patch_size`, `temporal_patch_size`, `spatial_merge_size`, etc.).

Since llm-d cannot download the video or access model config at routing time, this PR uses
`num_frames × tokens_per_frame` as an approximation, both configurable with vLLM-grounded defaults
(32 frames, 225 tokens/frame). Happy to discuss whether there's a better way!

capri-xiyue: 01. number of frames
I thought the number of frame is also per model, see https://github.com/search?q=repo%3Avllm-project%2Fvllm%20get_num_video_token&type=code

02. tokens per frame
Yep, it is based on model, I think for tokens per frame, it is the same as when I guess it for image placeholders. You can refer to the existing image placeholder guess logic.

I think we can try qwen3 vl model and gemma4 models to see how it really works. I wrote https://github.com/capri-xiyue/multi-modal-benchmark-helper/blob/main/image_token_guess.py to kind of validate whether the guessing is accurate or not. You can extend it to video and see how things goes.

namgyu-youn: Thanks, I updated the formula based on your reference and the Qwen3-VL architecture. Could you take another look?

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
```diff
@@ -143,9 +150,19 @@ type imageTokenEstimatorConfig struct {
 	FixedCfg          *fixedTokenEstimatorConfig   `json:"fixed,omitempty"`
 }
 
+// videoTokenEstimatorConfig defines the configuration for video modality.
+// Token count is estimated as NumFrames * TokensPerFrame.
+// NumFrames is model-specific in vLLM; 32 is used as a configurable starting point.
+// TokensPerFrame reuses imageTokenEstimatorConfig since a video frame is treated as an image.
+type videoTokenEstimatorConfig struct {
+	NumFrames      int                        `json:"numFrames"`
```
capri-xiyue: I think numframes should not be configured by user. I think we need to figure out how model server evaluate the number of frames and whether it varies across model. If it varies across model, user should be able to configure it based on different model. As in most cases, the. number of frame will depend on the duration of the video, the number of frame should not be a EPP config.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
capri-xiyue: Added comment

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
namgyu-youn: Agree, but as far as I know, llm-d doesn't have access to video metadata (duration, fps) at routing time. Could we separate this into a follow-up PR and keep a hardcoded default for now?

capri-xiyue: I think we can have the same mechanism as vllm, see https://docs.vllm.ai/en/stable/features/multimodal_inputs/#pre-extracted-frame-sequences-with-media_io_kwargs. When users send requests to llmd, we can get some metadata from the request?  Basically the order can be
1. llmd try to fetch video metadata(not sure whether it is feasible, for image, I was able to do it)
2. per request video metadata set by user
3. epp level user config metadata
4. epp level default video config metadata

namgyu-youn: Thanks for the suggestion, so the plan would be:
1. Keep a hardcoded default for now (EPP-level fallback)
2. Add per-request metadata support similar to vLLM
3. Build llm-d's video metadata fetching logic directly

Is it okay to land the current PR as the baseline and track 2/3 as follow-ups?

capri-xiyue: I think the hardcoded default probably should not be the number of frame. Probably EPP can set the video metadata default, like default duration, default fps and so on.

capri-xiyue: By the way, the OSS is working on the centralized token estimation in a `token-producer` plugin, eventually such video estimation code should live in the `token-producer` plugin, see https://github.com/llm-d/llm-d-router/pull/1445.

And for the video metadata default format, I think you can refer to what kind of video metadata that model server will use when calculating the number of frames.

ahg-g:
@capri-xiyue is this PR still relevant?

capri-xiyue:
> @capri-xiyue is this PR still relevant?

Yep. But I think the merge of the PR needs some evaluation regarding whether the estimation accuracy. I will follow it up with the contributor.

capri-xiyue:
@namgyu-youn Can you use qwen3-vl model and gemma4 model as an example to evaluate the estimation accuracy of  int(Duration × FPS) × tokensPerFrame  and paste the result here. You can use different fps and duration video and compare the difference about the estimated video tokens and real video tokens, I've done this for image before
```
=== Starting tests for resolution: 1920x1280 ===
[1920x1280 - Iteration 1/5] Processing: https://picsum.photos/1920/1280?random=1
  -> Calculated: 3136.69 | API: 3196 | Diff: 59.31
[1920x1280 - Iteration 2/5] Processing: https://picsum.photos/1920/1280?random=2
  -> Calculated: 3136.69 | API: 3196 | Diff: 59.31
[1920x1280 - Iteration 3/5] Processing: https://picsum.photos/1920/1280?random=3
  -> Calculated: 3136.69 | API: 3196 | Diff: 59.31
[1920x1280 - Iteration 4/5] Processing: https://picsum.photos/1920/1280?random=4
  -> Calculated: 3136.69 | API: 3196 | Diff: 59.31
[1920x1280 - Iteration 5/5] Processing: https://picsum.photos/1920/1280?random=5
  -> Calculated: 3136.69 | API: 3196 | Diff: 59.31

=== Starting tests for resolution: 1024x1024 ===
[1024x1024 - Iteration 1/5] Processing: https://picsum.photos/1024/1024?random=1
  -> Calculated: 1339.47 | API: 1391 | Diff: 51.53
[1024x1024 - Iteration 2/5] Processing: https://picsum.photos/1024/1024?random=2
  -> Calculated: 1339.47 | API: 1391 | Diff: 51.53
[1024x1024 - Iteration 3/5] Processing: https://picsum.photos/1024/1024?random=3
  -> Calculated: 1339.47 | API: 1391 | Diff: 51.53
[1024x1024 - Iteration 4/5] Processing: https://picsum.photos/1024/1024?random=4
  -> Calculated: 1339.47 | API: 1391 | Diff: 51.53
[1024x1024 - Iteration 5/5] Processing: https://picsum.photos/1024/1024?random=5
  -> Calculated: 1339.47 | API: 1391 | Diff: 51.53

=== Starting tests for resolution: 800x600 ===
[800x600 - Iteration 1/5] Processing: https://picsum.photos/800/600?random=1
  -> Calculated: 614.24 | API: 631 | Diff: 16.76
[800x600 - Iteration 2/5] Processing: https://picsum.photos/800/600?random=2
  -> Calculated: 614.24 | API: 631 | Diff: 16.76
[800x600 - Iteration 3/5] Processing: https://picsum.photos/800/600?random=3
  -> Calculated: 614.24 | API: 631 | Diff: 16.76
[800x600 - Iteration 4/5] Processing: https://picsum.photos/800/600?random=4
  -> Calculated: 614.24 | API: 631 | Diff: 16.76
[800x600 - Iteration 5/5] Processing: https://picsum.photos/800/600?random=5
  -> Calculated: 614.24 | API: 631 | Diff: 16.76

```
And for image, I use https://github.com/rahulgurnani/multimodal-microbenchmark-tool/blob/shared-multimodal/image_token_guess.py, you can extend it to video


I'd suggest try video with duration 10s, 30s,60s and 180s. For fps, I'd also suggest you try different fps, small to large.

namgyu-youn:
@capri-xiyue sorry for late reply. I noticed token estimation formula is broken with 1-2 min videos, so I extended it with Qwen3-VL-30B-A3B-Instruct (2×NVIDIA RTX 6000 Ada, Video-MME dataset).

**Simple formula (old) vs. new; https://github.com/llm-d/llm-d-router/pull/1408/changes/58abfcba1787e071b0c6b8de856dddf68fc69465)**:

Old formula breaks for 1-2 min videos because vLLM caps loaded frames at 32 for any video with >32 total frames — making duration×fps irrelevant. For a 74s/30fps video: `clamp(2220, 4, 768) × 112 = 86,016` vs actual 3675. The additional factors missed are `smart_resize`, `TEMPORAL_PATCH=2` grouping, and per-grid timestamp tokens.

New modeling correctly works with Video-MME dataset:
```
Video                             Dur   FPS        Res      Est   Actual     Diff   Diff%
----------------------------------------------------------------------------------------
fFjv93ACGo8                      74.0  30.0 640x360       3675     3675     0.00    0.0%
N1cdUjctpG8                      86.0  30.0 640x360       3675     3675     0.00    0.0%
HIjX8OPuf-w                     113.0  25.0 640x360       3677     3677     0.00    0.0%
24i4ncHuf6A                     112.0  25.0 640x360       3677     3677     0.00    0.0%
40BlVzjxu-I                     107.0  30.0 640x360       3676     3676     0.00    0.0%
0ay2Qy3wBe8                      76.0  30.0 640x360       3675     3675     0.00    0.0%
_tvmjsKXTu8                     115.0  24.0 640x360       3677     3677     0.00    0.0%
sUDY-SMREtA                      77.0  24.0 640x360       3675     3675     0.00    0.0%
PSt_op3fQck                     115.0  25.0 634x360       3677     3677     0.00    0.0%
```

**On the 10s/30s/60s/180s range**: all four durations at normal fps hit the same 32-frame cap (any video >~1s at 30fps has >32 total frames), so results would be identical to the above. The uncapped branch is a separate code path not reachable with Video-MME clips.

<details>
<summary>Benchmark script</summary>

```python
"""
Benchmark: video token estimation for Qwen3-VL using Video-MME.
Usage: python run_benchmark.py --base-url http://localhost:8000 [--num-videos 10]
Dependencies: pip install requests huggingface_hub datasets yt-dlp transformers
Reference: github.com/QwenLM/Qwen3-VL (vision_process.py)
"""

import argparse, base64, math, time
import numpy as np, requests, yt_dlp
from datasets import load_dataset
from transformers.models.qwen3_vl.video_processing_qwen3_vl import smart_resize as _smart_resize

MODEL           = "Qwen/Qwen3-VL-30B-A3B-Instruct"
PATCH_SIZE      = 16            # Qwen3-VL ViT patch size
MERGE_SIZE      = 2             # spatial merge
TEMPORAL_PATCH  = 2             # temporal grouping
VP_MIN_PIXELS   = 4_096         # video_processor.size["shortest_edge"]
VP_MAX_PIXELS   = 25_165_824    # video_processor.size["longest_edge"]
VLLM_NUM_FRAMES = 32            # VideoMediaIO default num_frames cap
SAMPLE_FPS      = 2.0
FPS_MIN_FRAMES  = 4
FPS_MAX_FRAMES  = 768
PROMPT_OVERHEAD = 13            # chat template + "Describe this video briefly."
ITERS           = 3


def estimate(total_frames, src_fps, w, h):
    # vLLM VideoBackend caps loaded frames at VLLM_NUM_FRAMES=32.
    # <=32 total -> all frames loaded, do_sample_frames=True  -> HF processor fps-samples
    # >32 total -> 32 frames pre-sampled, do_sample_frames=False -> HF uses all 32 as-is
    if total_frames <= VLLM_NUM_FRAMES:
        nframes = int(total_frames / src_fps * SAMPLE_FPS)
        nframes = max(FPS_MIN_FRAMES, min(FPS_MAX_FRAMES, nframes))
    else:
        nframes = VLLM_NUM_FRAMES

    h_bar, w_bar = _smart_resize(nframes, h, w, min_pixels=VP_MIN_PIXELS, max_pixels=VP_MAX_PIXELS)
    t_bar  = math.ceil(nframes / TEMPORAL_PATCH) * TEMPORAL_PATCH
    grid_t = t_bar  // TEMPORAL_PATCH
    grid_h = h_bar  // PATCH_SIZE
    grid_w = w_bar  // PATCH_SIZE
    visual = grid_t * grid_h * grid_w // (MERGE_SIZE ** 2)

    # Per-grid overhead: "<X.X seconds>" = 6/7/8 tokens for X <10/<100/>=100,
    # plus vision_start (1) + vision_end (1).
    # For >32-frame videos use exact linspace frame positions (matching vLLM's pre-sampling).
    if total_frames > VLLM_NUM_FRAMES:
        frame_idx  = np.linspace(0, total_frames - 1, VLLM_NUM_FRAMES, dtype=int).tolist()
        raw_ts     = [idx / src_fps for idx in frame_idx]
        timestamps = [(raw_ts[i] + raw_ts[i + 1]) / 2
                      for i in range(0, VLLM_NUM_FRAMES, TEMPORAL_PATCH)]
    else:
        duration   = total_frames / src_fps
        timestamps = [(i + 0.5) * duration / grid_t for i in range(grid_t)]
    ts_tokens = sum(6 + (t >= 10) + (t >= 100) for t in timestamps)
    return visual + ts_tokens + grid_t * 2 + PROMPT_OVERHEAD


def query(base_url, path):
    if path.startswith("http"):
        data = requests.get(path, timeout=120).content
    else:
        with open(path, "rb") as f:
            data = f.read()
    b64 = base64.b64encode(data).decode()
    r = requests.post(f"{base_url}/v1/chat/completions", timeout=180, json={
        "model": MODEL,
        "messages": [{"role": "user", "content": [
            {"type": "video_url", "video_url": {"url": f"data:video/mp4;base64,{b64}"}},
            {"type": "text", "text": "Describe this video briefly."},
        ]}],
        "max_tokens": 1,   # minimise generation; we only need prompt_tokens from usage
    })
    r.raise_for_status()
    return r.json()["usage"]["prompt_tokens"]


def resolve_stream_url(youtube_url):
    """Return (direct_cdn_url, video_id, total_frames, fps, width, height)."""
    opts = {"format": "best[ext=mp4]/best", "quiet": True, "no_warnings": True}
    with yt_dlp.YoutubeDL(opts) as ydl:
        info = ydl.extract_info(youtube_url, download=False)
    fps = info.get("fps") or 30.0
    w, h = info["width"], info["height"]
    n = int(info["duration"] * fps)
    return info["url"], info["id"], n, fps, w, h


def main():
    p = argparse.ArgumentParser()
    p.add_argument("--base-url",   default="http://localhost:8000")
    p.add_argument("--num-videos", type=int, default=10)
    args = p.parse_args()
    base_url = args.base_url.rstrip("/")

    ds = load_dataset("lmms-lab/Video-MME", split="test")
    # deduplicate: one entry per unique videoID
    seen, unique_rows = set(), []
    for row in ds:
        if row["videoID"] not in seen:
            seen.add(row["videoID"])
            unique_rows.append(row)
        if len(unique_rows) == args.num_videos:
            break

    rows = []
    for row in unique_rows:
        vid = row["videoID"]
        try:
            stream_url, _, n, fps, w, h = resolve_stream_url(row["url"])
        except Exception as e:
            print(f"SKIP {vid}: {e}")
            continue

        est = estimate(n, fps, w, h)
        print(f"\n=== {vid} | {n/fps:.1f}s {fps:.1f}fps {w}x{h} est={est} ===")
        actuals = []
        for i in range(ITERS):
            print(f"  [iter {i+1}]", end=" ", flush=True)
            try:
                actual = query(base_url, stream_url)
                print(f"api={actual} diff={abs(est-actual)}")
                actuals.append(actual)
            except Exception as e:
                print(f"ERROR: {e}")
            time.sleep(0.3)
        if actuals:
            avg = sum(actuals) / len(actuals)
            rows.append((vid, n/fps, fps, w, h, est, avg, abs(est-avg)))

    if not rows:
        return
    print(f"\n{'Video':<30} {'Dur':>6} {'FPS':>5} {'Res':>10} {'Est':>8} {'Actual':>8} {'Diff':>8} {'Diff%':>7}")
    print("-" * 88)
    for name, dur, fps, w, h, est, actual, diff in rows:
        pct = diff / actual * 100 if actual else float("nan")
        print(f"{name[:28]:<30} {dur:>6.1f} {fps:>5.1f} {w}x{h:<5} {est:>8.0f} {actual:>8.0f} {diff:>8.2f} {pct:>6.1f}%")


if __name__ == "__main__":
    main()
```

</details>

<details>
<summary>Full log</summary>

```bash
=== fFjv93ACGo8 | 74.0s 30.0fps 640x360 est=3675 ===
  [iter 1] api=3675 diff=0
  [iter 2] api=3675 diff=0
  [iter 3] api=3675 diff=0

=== N1cdUjctpG8 | 86.0s 30.0fps 640x360 est=3675 ===
  [iter 1] api=3675 diff=0
  [iter 2] api=3675 diff=0
  [iter 3] api=3675 diff=0

=== HIjX8OPuf-w | 113.0s 25.0fps 640x360 est=3677 ===
  [iter 1] api=3677 diff=0
  [iter 2] api=3677 diff=0
  [iter 3] api=3677 diff=0
ERROR: [youtube] HwnB8aCn8yE: Video unavailable. This video is no longer available because the YouTube account associated with this video has been terminated.
SKIP HwnB8aCn8yE: ERROR: [youtube] HwnB8aCn8yE: Video unavailable...

=== 24i4ncHuf6A | 112.0s 25.0fps 640x360 est=3677 ===
  [iter 1] api=3677 diff=0
  [iter 2] api=3677 diff=0
  [iter 3] api=3677 diff=0

=== 40BlVzjxu-I | 107.0s 30.0fps 640x360 est=3676 ===
  [iter 1] api=3676 diff=0
  [iter 2] api=3676 diff=0
  [iter 3] api=3676 diff=0

=== 0ay2Qy3wBe8 | 76.0s 30.0fps 640x360 est=3675 ===
  [iter 1] api=3675 diff=0
  [iter 2] api=3675 diff=0
  [iter 3] api=3675 diff=0

=== _tvmjsKXTu8 | 115.0s 24.0fps 640x360 est=3677 ===
  [iter 1] api=3677 diff=0
  [iter 2] api=3677 diff=0
  [iter 3] api=3677 diff=0

=== sUDY-SMREtA | 77.0s 24.0fps 640x360 est=3675 ===
  [iter 1] api=3675 diff=0
  [iter 2] api=3675 diff=0
  [iter 3] api=3675 diff=0

=== PSt_op3fQck | 115.0s 25.0fps 634x360 est=3677 ===
  [iter 1] api=3677 diff=0
  [iter 2] api=3677 diff=0
  [iter 3] api=3677 diff=0
```

</details>

btw, I'm not sure how to rebase this PR. Is it okay to just reland?

capri-xiyue:
> @capri-xiyue sorry for late reply. I noticed token estimation formula is broken with 1-2 min videos, so I extended it with Qwen3-VL-30B-A3B-Instruct (2×NVIDIA RTX 6000 Ada, Video-MME dataset).
> 
> **Simple formula (old) vs. new; [58abfcb](https://github.com/llm-d/llm-d-router/commit/58abfcba1787e071b0c6b8de856dddf68fc69465))**:
> 
> Old formula breaks for 1-2 min videos because vLLM caps loaded frames at 32 for any video with >32 total frames — making duration×fps irrelevant. For a 74s/30fps video: `clamp(2220, 4, 768) × 112 = 86,016` vs actual 3675. The additional factors missed are `smart_resize`, `TEMPORAL_PATCH=2` grouping, and per-grid timestamp tokens.
> 
> New modeling correctly works with Video-MME dataset:
> 
> ```
> Video                             Dur   FPS        Res      Est   Actual     Diff   Diff%
> ----------------------------------------------------------------------------------------
> fFjv93ACGo8                      74.0  30.0 640x360       3675     3675     0.00    0.0%
> N1cdUjctpG8                      86.0  30.0 640x360       3675     3675     0.00    0.0%
> HIjX8OPuf-w                     113.0  25.0 640x360       3677     3677     0.00    0.0%
> 24i4ncHuf6A                     112.0  25.0 640x360       3677     3677     0.00    0.0%
> 40BlVzjxu-I                     107.0  30.0 640x360       3676     3676     0.00    0.0%
> 0ay2Qy3wBe8                      76.0  30.0 640x360       3675     3675     0.00    0.0%
> _tvmjsKXTu8                     115.0  24.0 640x360       3677     3677     0.00    0.0%
> sUDY-SMREtA                      77.0  24.0 640x360       3675     3675     0.00    0.0%
> PSt_op3fQck                     115.0  25.0 634x360       3677     3677     0.00    0.0%
> ```
> 
> **On the 10s/30s/60s/180s range**: all four durations at normal fps hit the same 32-frame cap (any video >~1s at 30fps has >32 total frames), so results would be identical to the above. The uncapped branch is a separate code path not reachable with Video-MME clips.
> 
> Benchmark script
> ```python
> """
> Benchmark: video token estimation for Qwen3-VL using Video-MME.
> Usage: python run_benchmark.py --base-url http://localhost:8000 [--num-videos 10]
> Dependencies: pip install requests huggingface_hub datasets yt-dlp transformers
> Reference: github.com/QwenLM/Qwen3-VL (vision_process.py)
> """
> 
> import argparse, base64, math, time
> import numpy as np, requests, yt_dlp
> from datasets import load_dataset
> from transformers.models.qwen3_vl.video_processing_qwen3_vl import smart_resize as _smart_resize
> 
> MODEL           = "Qwen/Qwen3-VL-30B-A3B-Instruct"
> PATCH_SIZE      = 16            # Qwen3-VL ViT patch size
> MERGE_SIZE      = 2             # spatial merge
> TEMPORAL_PATCH  = 2             # temporal grouping
> VP_MIN_PIXELS   = 4_096         # video_processor.size["shortest_edge"]
> VP_MAX_PIXELS   = 25_165_824    # video_processor.size["longest_edge"]
> VLLM_NUM_FRAMES = 32            # VideoMediaIO default num_frames cap
> SAMPLE_FPS      = 2.0
> FPS_MIN_FRAMES  = 4
> FPS_MAX_FRAMES  = 768
> PROMPT_OVERHEAD = 13            # chat template + "Describe this video briefly."
> ITERS           = 3
> 
> 
> def estimate(total_frames, src_fps, w, h):
>     # vLLM VideoBackend caps loaded frames at VLLM_NUM_FRAMES=32.
>     # <=32 total -> all frames loaded, do_sample_frames=True  -> HF processor fps-samples
>     # >32 total -> 32 frames pre-sampled, do_sample_frames=False -> HF uses all 32 as-is
>     if total_frames <= VLLM_NUM_FRAMES:
>         nframes = int(total_frames / src_fps * SAMPLE_FPS)
>         nframes = max(FPS_MIN_FRAMES, min(FPS_MAX_FRAMES, nframes))
>     else:
>         nframes = VLLM_NUM_FRAMES
> 
>     h_bar, w_bar = _smart_resize(nframes, h, w, min_pixels=VP_MIN_PIXELS, max_pixels=VP_MAX_PIXELS)
>     t_bar  = math.ceil(nframes / TEMPORAL_PATCH) * TEMPORAL_PATCH
>     grid_t = t_bar  // TEMPORAL_PATCH
>     grid_h = h_bar  // PATCH_SIZE
>     grid_w = w_bar  // PATCH_SIZE
>     visual = grid_t * grid_h * grid_w // (MERGE_SIZE ** 2)
> 
>     # Per-grid overhead: "<X.X seconds>" = 6/7/8 tokens for X <10/<100/>=100,
>     # plus vision_start (1) + vision_end (1).
>     # For >32-frame videos use exact linspace frame positions (matching vLLM's pre-sampling).
>     if total_frames > VLLM_NUM_FRAMES:
>         frame_idx  = np.linspace(0, total_frames - 1, VLLM_NUM_FRAMES, dtype=int).tolist()
>         raw_ts     = [idx / src_fps for idx in frame_idx]
>         timestamps = [(raw_ts[i] + raw_ts[i + 1]) / 2
>                       for i in range(0, VLLM_NUM_FRAMES, TEMPORAL_PATCH)]
>     else:
>         duration   = total_frames / src_fps
>         timestamps = [(i + 0.5) * duration / grid_t for i in range(grid_t)]
>     ts_tokens = sum(6 + (t >= 10) + (t >= 100) for t in timestamps)
>     return visual + ts_tokens + grid_t * 2 + PROMPT_OVERHEAD
> 
> 
> def query(base_url, path):
>     if path.startswith("http"):
>         data = requests.get(path, timeout=120).content
>     else:
>         with open(path, "rb") as f:
>             data = f.read()
>     b64 = base64.b64encode(data).decode()
>     r = requests.post(f"{base_url}/v1/chat/completions", timeout=180, json={
>         "model": MODEL,
>         "messages": [{"role": "user", "content": [
>             {"type": "video_url", "video_url": {"url": f"data:video/mp4;base64,{b64}"}},
>             {"type": "text", "text": "Describe this video briefly."},
>         ]}],
>         "max_tokens": 1,   # minimise generation; we only need prompt_tokens from usage
>     })
>     r.raise_for_status()
>     return r.json()["usage"]["prompt_tokens"]
> 
> 
> def resolve_stream_url(youtube_url):
>     """Return (direct_cdn_url, video_id, total_frames, fps, width, height)."""
>     opts = {"format": "best[ext=mp4]/best", "quiet": True, "no_warnings": True}
>     with yt_dlp.YoutubeDL(opts) as ydl:
>         info = ydl.extract_info(youtube_url, download=False)
>     fps = info.get("fps") or 30.0
>     w, h = info["width"], info["height"]
>     n = int(info["duration"] * fps)
>     return info["url"], info["id"], n, fps, w, h
> 
> 
> def main():
>     p = argparse.ArgumentParser()
>     p.add_argument("--base-url",   default="http://localhost:8000")
>     p.add_argument("--num-videos", type=int, default=10)
>     args = p.parse_args()
>     base_url = args.base_url.rstrip("/")
> 
>     ds = load_dataset("lmms-lab/Video-MME", split="test")
>     # deduplicate: one entry per unique videoID
>     seen, unique_rows = set(), []
>     for row in ds:
>         if row["videoID"] not in seen:
>             seen.add(row["videoID"])
>             unique_rows.append(row)
>         if len(unique_rows) == args.num_videos:
>             break
> 
>     rows = []
>     for row in unique_rows:
>         vid = row["videoID"]
>         try:
>             stream_url, _, n, fps, w, h = resolve_stream_url(row["url"])
>         except Exception as e:
>             print(f"SKIP {vid}: {e}")
>             continue
> 
>         est = estimate(n, fps, w, h)
>         print(f"\n=== {vid} | {n/fps:.1f}s {fps:.1f}fps {w}x{h} est={est} ===")
>         actuals = []
>         for i in range(ITERS):
>             print(f"  [iter {i+1}]", end=" ", flush=True)
>             try:
>                 actual = query(base_url, stream_url)
>                 print(f"api={actual} diff={abs(est-actual)}")
>                 actuals.append(actual)
>             except Exception as e:
>                 print(f"ERROR: {e}")
>             time.sleep(0.3)
>         if actuals:
>             avg = sum(actuals) / len(actuals)
>             rows.append((vid, n/fps, fps, w, h, est, avg, abs(est-avg)))
> 
>     if not rows:
>         return
>     print(f"\n{'Video':<30} {'Dur':>6} {'FPS':>5} {'Res':>10} {'Est':>8} {'Actual':>8} {'Diff':>8} {'Diff%':>7}")
>     print("-" * 88)
>     for name, dur, fps, w, h, est, actual, diff in rows:
>         pct = diff / actual * 100 if actual else float("nan")
>         print(f"{name[:28]:<30} {dur:>6.1f} {fps:>5.1f} {w}x{h:<5} {est:>8.0f} {actual:>8.0f} {diff:>8.2f} {pct:>6.1f}%")
> 
> 
> if __name__ == "__main__":
>     main()
> ```
> 
> Full log
> ```shell
> === fFjv93ACGo8 | 74.0s 30.0fps 640x360 est=3675 ===
>   [iter 1] api=3675 diff=0
>   [iter 2] api=3675 diff=0
>   [iter 3] api=3675 diff=0
> 
> === N1cdUjctpG8 | 86.0s 30.0fps 640x360 est=3675 ===
>   [iter 1] api=3675 diff=0
>   [iter 2] api=3675 diff=0
>   [iter 3] api=3675 diff=0
> 
> === HIjX8OPuf-w | 113.0s 25.0fps 640x360 est=3677 ===
>   [iter 1] api=3677 diff=0
>   [iter 2] api=3677 diff=0
>   [iter 3] api=3677 diff=0
> ERROR: [youtube] HwnB8aCn8yE: Video unavailable. This video is no longer available because the YouTube account associated with this video has been terminated.
> SKIP HwnB8aCn8yE: ERROR: [youtube] HwnB8aCn8yE: Video unavailable...
> 
> === 24i4ncHuf6A | 112.0s 25.0fps 640x360 est=3677 ===
>   [iter 1] api=3677 diff=0
>   [iter 2] api=3677 diff=0
>   [iter 3] api=3677 diff=0
> 
> === 40BlVzjxu-I | 107.0s 30.0fps 640x360 est=3676 ===
>   [iter 1] api=3676 diff=0
>   [iter 2] api=3676 diff=0
>   [iter 3] api=3676 diff=0
> 
> === 0ay2Qy3wBe8 | 76.0s 30.0fps 640x360 est=3675 ===
>   [iter 1] api=3675 diff=0
>   [iter 2] api=3675 diff=0
>   [iter 3] api=3675 diff=0
> 
> === _tvmjsKXTu8 | 115.0s 24.0fps 640x360 est=3677 ===
>   [iter 1] api=3677 diff=0
>   [iter 2] api=3677 diff=0
>   [iter 3] api=3677 diff=0
> 
> === sUDY-SMREtA | 77.0s 24.0fps 640x360 est=3675 ===
>   [iter 1] api=3675 diff=0
>   [iter 2] api=3675 diff=0
>   [iter 3] api=3675 diff=0
> 
> === PSt_op3fQck | 115.0s 25.0fps 634x360 est=3677 ===
>   [iter 1] api=3677 diff=0
>   [iter 2] api=3677 diff=0
>   [iter 3] api=3677 diff=0
> ```
> 
> btw, I'm not sure how to rebase this PR. Is it okay to just reland?

If you are not sure how to rebase the PR, it is ok to just have a new PR. 

>  I noticed token estimation formula is broken with 1-2 min videos
It makes sense as I think both qwen3vl and gemma 4 has a max frame per video. I think we also need to include this in epp config. 

I think for the landing of this PR, at least we need to make sure for qwen3vl and gemma4 model, users can configure EPP to make the estimation relatively accurate.

namgyu-youn:
> I think for the landing of this PR, at least we need to make sure for qwen3vl and gemma4 model, users can configure EPP to make the estimation relatively accurate.

Sorry, I think I'm missing something. Do we actually need 2-model support with accurate token estimation at this point? Your roadmap's Phase 2 plan scoped this to one model first, so I want to check before expanding scope here.

How about (1) land this PR scoped to Qwen3-VL accuracy via the configurable frame-cap I benchmarked above, (2) track Gemma4-specific tuning as a follow-up issue? Or is there a reason both need to land together now?

capri-xiyue:
> > I think for the landing of this PR, at least we need to make sure for qwen3vl and gemma4 model, users can configure EPP to make the estimation relatively accurate.
> 
> Sorry, I think I'm missing something. Do we actually need 2-model support with accurate token estimation at this point? Your roadmap's Phase 2 plan scoped this to one model first, so I want to check before expanding scope here.
> 
> How about (1) land this PR scoped to Qwen3-VL accuracy via the configurable frame-cap I benchmarked above, (2) track Gemma4-specific tuning as a follow-up issue? Or is there a reason both need to land together now?

My docs list qwen3vl as an example, it doesn't mean we only support qwen3vl. For image, you can see we support qwen3 vl and gemma4, for video, we also want to add it.
 
It's fine for this PR we just land qwen3 vl, but my main concern about this PR is the user interface.  As I've mentioned, fps and duration video should not be part of the user interface, just the default fps and duration when epp can't grep the metadata from the video. 
 Let us discuss offline in slack to make sure we are on the same page. Are you in llmd slack group

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
```diff
@@ -143,9 +144,18 @@ type imageTokenEstimatorConfig struct {
 	FixedCfg          *fixedTokenEstimatorConfig   `json:"fixed,omitempty"`
 }
 
+// videoTokenEstimatorConfig defines the configuration for video modality.
+// Token count is estimated as clamp(int(Duration*FPS), 4, 768) * TokensPerFrame.
+type videoTokenEstimatorConfig struct {
+	Duration       float64                    `json:"duration"`
```
capri-xiyue: fps and duration should not be part of epp config as they vary across videos.

EPP can use default duration and fps when epp can't fetch it from video.

I'd be cautious when we expose such epp user interface. We need to make sure it works for at least qwen3vl and gemma4 model.

If you can't finish that in this PR, please don't add the user interface. Just write default duration and fps in the code directly and have the underlying implementation.

namgyu-youn:
> It's fine for this PR we just land qwen3 vl, but my main concern about this PR is the user interface. As I've mentioned, fps and duration video should not be part of the user interface, just the default fps and duration when epp can't grep the metadata from the video. Let us discuss offline in slack to make sure we are on the same page. Are you in llmd slack group

Not yet, please invite to <namgyu.dev@gmail.com>, thanks!

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
namgyu-youn: Oh now I see, let me fix it to use real meta-data directly.

capri-xiyue:
> > It's fine for this PR we just land qwen3 vl, but my main concern about this PR is the user interface. As I've mentioned, fps and duration video should not be part of the user interface, just the default fps and duration when epp can't grep the metadata from the video. Let us discuss offline in slack to make sure we are on the same page. Are you in llmd slack group
> 
> Not yet, please invite to [namgyu.dev@gmail.com](mailto:namgyu.dev@gmail.com), thanks!

I can't add you, can you join it directly https://llm-d.ai/docs/getting-started (see join slack).

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/token_estimator.go
```diff
@@ -85,20 +84,85 @@ func getImagePlaceholders(ctx context.Context, url string, multimodalConfig *mul
...
 			if err != nil {
 				logger.Error(err, "failed to get image dimensions from base64 content, using default image resolution")
-				numPlaceHolders = multimodalConfig.Image.DefaultResolution.Width * multimodalConfig.Image.DefaultResolution.Height / multimodalConfig.Image.DynamicCfg.Factor
+				f := multimodalConfig.Image.DynamicCfg.Factor
+				numPlaceHolders = multimodalConfig.Image.DefaultResolution.Width*multimodalConfig.Image.DefaultResolution.Height/(f*f) + 2
 			} else {
 				logger.Info(fmt.Sprintf("Using image resolution height %d width %d", resolution.Height, resolution.Width))
-				numPlaceHolders = (resolution.Width * resolution.Height) / (multimodalConfig.Image.DynamicCfg.Factor)
+				f := multimodalConfig.Image.DynamicCfg.Factor
+				numPlaceHolders = resolution.Width*resolution.Height/(f*f) + 2
```
capri-xiyue: This part is not needed. Now the f here is the already the factor * factor mentioned in roadmap docs, the roadmap docs is a little bit out-dated

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
capri-xiyue: I saw you have some video_metadata helper? Do you need to update such user interface of videoTokenEstimatorConfig? In addition, I now probably have some bandwidth to work on the video token estimation for approximate prefix cache. Are you open to grant me editor permission of your fork repo so that I can update this PR directly?

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/mm_estimate.go
```diff
@@ -148,3 +149,195 @@ func imageDimensionsFromBase64Payload(rawB64 string) (width, height int, ok bool
 	}
 	return cfg.Width, cfg.Height, true
 }
+
```
capri-xiyue, line 152: Could you add some comments explaining the logic behind your video token estimation, specifically how you are using video metadata including duration and fps and also other model specific default like sample FPS, patch size, merge size, and temporal patch?

I'm fine with this approach for the current PR. However, for future PRs, we'll need to consolidate these custom tuning parameters so the EPP interface doesn't become overly complicated. Based on my previous research, a simpler approach would be to use one factor parameter for tokens per frame, another factor for the total frame count, and likely min/max bounds for the frame count. Documenting your current estimation method in detail now will make it much easier for us to streamline these parameters later.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/video_metadata.go
```diff
@@ -0,0 +1,193 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
```
capri-xiyue, line 8: Could you add a UT to verify this works? We can use video base64 encoding content

capri-xiyue:
@namgyu-youn Thanks for your contribution! The PR looks great overall. I’ve just left a few minor comments requesting a bit more explanation on the video token estimation and some additional unit test coverage.

Also, would you be open to joining the llmd Slack community? You can find the link here: https://llm-d.ai/docs/getting-started. It would make it much easier for us to communicate! We're planning to add video support to the approximate prefix cache aware plugin for the next llmd release coming up in 3 weeks. It would be great to sync up on Slack to figure out how we can efficiently divide up those tasks.

capri-xiyue:
@liu-cong This PR looks good to me overall, and I've just left a few minor comments. Since you have approver permissions, could you please take a look as well? Thanks!

namgyu-youn:
> @namgyu-youn Thanks for your contribution! The PR looks great overall. I’ve just left a few minor comments requesting a bit more explanation on the video token estimation and some additional unit test coverage.
> 
> Also, would you be open to joining the llmd Slack community? You can find the link here: [llm-d.ai/docs/getting-started](https://llm-d.ai/docs/getting-started). It would make it much easier for us to communicate! We're planning to add video support to the approximate prefix cache aware plugin for the next llmd release coming up in 3 weeks. It would be great to sync up on Slack to figure out how we can efficiently divide up those tasks.

Thanks so much for the fast review and feedback — I'll work on applying it now.

I also recently joined the llm-d Slack as "Namgyu Youn", so feel free to reach out there. Looking forward to syncing with you and the team on how to divide up the video support work :)

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go
namgyu-youn: > Are you open to grant me editor permission of your fork repo so that I can update this PR directly?

I've invited you to my downstream fork for faster collaboration. Feel free to commit directly if I'm unavailable.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/video_metadata.go
namgyu-youn: Updated via [`368be37` (this PR)](https://github.com/llm-d/llm-d-router/pull/1408/changes/368be372bad5abb18277c7ad28e392a615021896). Could you taks a look?

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/mm_estimate.go
```diff
@@ -155,8 +155,8 @@ const (
 	defaultVideoMaxFrames     = 32         // vLLM pre-sampling cap
 	defaultVideoSampleFPS     = 2.0        // fps used for short-video resampling
 	defaultVideoPatchSize     = 16         // ViT patch size in pixels
-	defaultVideoMergeSize     = 2          // spatial merge factor
-	defaultVideoTemporalPatch = 2          // temporal grouping factor
+	defaultVideoMergeSize     = 2          // NxN patches merge into 1 visual token
```
capri-xiyue: Can we have more comments regarding how the final number of video tokens get calculated, for example, for image estimator

Estimate image tokens based on image width and height: 
Tokens=Image Width×Image Height/Factor

capri-xiyue: @namgyu-youn Friendly ping about the final formula. I know for video it will be more complex as more parameters will be needed. But I think adding it in comments will be quite useful.

namgyu-youn: updated via [`52a769f` (this PR)](https://github.com/llm-d/llm-d-router/pull/1408/changes/52a769f6a0aecb6bdd643092a8029e06cb4901cf); surely more comments are needed.

capri-xiyue:
@namgyu-youn I wrote another verification benchmark

```
func TestVideoEstimator_PlaceholderCount_Live(t *testing.T) {
	endpoint := os.Getenv("VIDEO_ESTIMATE_ENDPOINT")
	if endpoint == "" {
		endpoint = "34.57.228.181:8000"
	}
	model := os.Getenv("VIDEO_ESTIMATE_MODEL")
	if model == "" {
		model = "Qwen/Qwen3-VL-30B-A3B-Instruct"
	}

	resolutions := []string{"360p", "720p", "1080p"}
	durations := []int{1, 5, 10, 20, 30, 60, 90}

	e := newVideoEstimator()
	client := &http.Client{Timeout: 120 * time.Second}

	t.Logf("%-10s %-6s %10s %10s %10s %8s", "resolution", "dur", "estimate", "actual", "diff", "err%")
	for _, res := range resolutions {
		for _, dur := range durations {
			res, dur := res, dur
			t.Run(fmt.Sprintf("%s/%ds", res, dur), func(t *testing.T) {
				videoURL := fmt.Sprintf("https://lorem.video/%s_h264_%ds", res, dur)
				raw, err := download(context.Background(), client, videoURL)
				if err != nil {
					t.Fatalf("download %s: %v", videoURL, err)
				}
				dataURL := "data:video/mp4;base64," + base64.StdEncoding.EncodeToString(raw)

				estimate := e.placeholderCount(dataURL)
				actual, err := liveVideoTokens(context.Background(), client, endpoint, model, dataURL)
				if err != nil {
					t.Fatalf("query server: %v", err)
				}

				diff := estimate - actual
				var errPct float64
				if actual != 0 {
					errPct = float64(diff) / float64(actual) * 100
				}
				t.Logf("%-10s %-6s %10d %10d %10d %7.1f%%", res, fmt.Sprintf("%ds", dur), estimate, actual, diff, errPct)
			})
		}
	}
}

// download fetches url and returns its body.
func download(ctx context.Context, client *http.Client, url string) ([]byte, error) {
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
	if err != nil {
		return nil, err
	}
	resp, err := client.Do(req)
	if err != nil {
		return nil, err
	}
	defer resp.Body.Close()
	if resp.StatusCode != http.StatusOK {
		return nil, fmt.Errorf("status %d", resp.StatusCode)
	}
	return io.ReadAll(resp.Body)
}

// liveVideoTokens posts a single-video chat completion request and returns the
// server-reported multimodal video token count from usage.prompt_tokens_details.
func liveVideoTokens(ctx context.Context, client *http.Client, endpoint, model, videoDataURL string) (int, error) {
	reqBody := map[string]any{
		"model": model,
		"messages": []any{
			map[string]any{
				"role": "user",
				"content": []any{
					map[string]any{"type": "text", "text": "describe the video"},
					map[string]any{"type": "video_url", "video_url": map[string]any{"url": videoDataURL}},
				},
			},
		},
		"max_tokens":  1,
		"temperature": 0,
	}
	body, err := json.Marshal(reqBody)
	if err != nil {
		return 0, err
	}

	url := fmt.Sprintf("http://%s/v1/chat/completions", endpoint)
	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewReader(body))
	if err != nil {
		return 0, err
	}
	req.Header.Set("Content-Type", "application/json")

	resp, err := client.Do(req)
	if err != nil {
		return 0, err
	}
	defer resp.Body.Close()
	respBody, err := io.ReadAll(resp.Body)
	if err != nil {
		return 0, err
	}
	if resp.StatusCode != http.StatusOK {
		return 0, fmt.Errorf("status %d: %s", resp.StatusCode, respBody)
	}

	var parsed struct {
		Usage struct {
			PromptTokensDetails struct {
				MultimodalTokens struct {
					Video int `json:"video"`
				} `json:"multimodal_tokens"`
			} `json:"prompt_tokens_details"`
		} `json:"usage"`
	}
	if err := json.Unmarshal(respBody, &parsed); err != nil {
		return 0, fmt.Errorf("decode response: %w", err)
	}
	return parsed.Usage.PromptTokensDetails.MultimodalTokens.Video, nil
}
```
And I saw your token estimation has large gap when compared to actual

=== RUN   TestVideoEstimator_PlaceholderCount_Live
    mm_estimate_live_test.go:59: resolution dur      estimate     actual       diff     err%
=== RUN   TestVideoEstimator_PlaceholderCount_Live/360p/1s
    mm_estimate_live_test.go:82: 360p       1s           3654        456       3198   701.3%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/1s (0.92s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/360p/5s
    mm_estimate_live_test.go:82: 360p       5s           3654       1140       2514   220.5%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/5s (0.56s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/360p/10s
    mm_estimate_live_test.go:82: 360p       10s          3654       2280       1374    60.3%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/10s (0.75s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/360p/20s
    mm_estimate_live_test.go:82: 360p       20s          3656       4570       -914   -20.0%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/20s (0.95s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/360p/30s
    mm_estimate_live_test.go:82: 360p       30s          3654       6860      -3206   -46.7%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/30s (1.05s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/360p/60s
    mm_estimate_live_test.go:82: 360p       60s          3654      11930      -8276   -69.4%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/60s (1.77s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/360p/90s
    mm_estimate_live_test.go:82: 360p       90s          3654      11600      -7946   -68.5%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/90s (2.51s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/720p/1s
    mm_estimate_live_test.go:82: 720p       1s           3654       1776       1878   105.7%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/1s (0.30s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/720p/5s
    mm_estimate_live_test.go:82: 720p       5s           3654       4440       -786   -17.7%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/5s (0.59s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/720p/10s
    mm_estimate_live_test.go:82: 720p       10s         12560       8880       3680    41.4%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/10s (1.30s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/720p/20s
    mm_estimate_live_test.go:82: 720p       20s         12568      12050        518     4.3%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/20s (1.85s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/720p/30s
    mm_estimate_live_test.go:82: 720p       30s          3654      12410      -8756   -70.6%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/30s (2.43s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/720p/60s
    mm_estimate_live_test.go:82: 720p       60s          3654      11930      -8276   -69.4%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/60s (4.91s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/720p/90s
    mm_estimate_live_test.go:82: 720p       90s          3654      11600      -7946   -68.5%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/90s (6.90s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/1080p/1s
    mm_estimate_live_test.go:82: 1080p      1s           3654       4096       -442   -10.8%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/1080p/1s (0.77s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/1080p/5s
    mm_estimate_live_test.go:74: query server: status 500: {"error":{"message":"Internal server error","type":"InternalServerError","param":null,"code":500}}
--- FAIL: TestVideoEstimator_PlaceholderCount_Live/1080p/5s (0.81s)

=== RUN   TestVideoEstimator_PlaceholderCount_Live/1080p/10s
    mm_estimate_live_test.go:82: 1080p      10s          3654      12300      -8646   -70.3%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/1080p/10s (1.66s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/1080p/20s
    mm_estimate_live_test.go:82: 1080p      20s         12568      12050        518     4.3%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/1080p/20s (3.69s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/1080p/30s
    mm_estimate_live_test.go:82: 1080p      30s          3654      12410      -8756   -70.6%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/1080p/30s (4.32s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/1080p/60s
    mm_estimate_live_test.go:82: 1080p      60s          3654      11930      -8276   -69.4%
--- PASS: TestVideoEstimator_PlaceholderCount_Live/1080p/60s (9.14s)
=== RUN   TestVideoEstimator_PlaceholderCount_Live/1080p/90s
    mm_estimate_live_test.go:82: 1080p      90s          3654      11600      -7946   -68.5%
    
    
    
I think this would be a blocker to merge this PR. Can you see why that happens? 

And I found with 
	const sampleFPS = 2.0
	const factor = 32 * 32
	frames := int(math.Round(float64(duration) * sampleFPS))
	tokensPerFrame := int(math.Round(float64(width*height) / factor))
	raw_tokens = min(frames * tokensPerFrame, 12288) 
	
It has better estimation than your current PR with qwen3 vl model,  at the same time I will try to work on another PR to see how thing goes.

namgyu-youn:
> @namgyu-youn I wrote another verification benchmark
> 
> ```
> func TestVideoEstimator_PlaceholderCount_Live(t *testing.T) {
> 	endpoint := os.Getenv("VIDEO_ESTIMATE_ENDPOINT")
> 	if endpoint == "" {
> 		endpoint = "34.57.228.181:8000"
> 	}
> 	model := os.Getenv("VIDEO_ESTIMATE_MODEL")
> 	if model == "" {
> 		model = "Qwen/Qwen3-VL-30B-A3B-Instruct"
> 	}
> 
> 	resolutions := []string{"360p", "720p", "1080p"}
> 	durations := []int{1, 5, 10, 20, 30, 60, 90}
> 
> 	e := newVideoEstimator()
> 	client := &http.Client{Timeout: 120 * time.Second}
> 
> 	t.Logf("%-10s %-6s %10s %10s %10s %8s", "resolution", "dur", "estimate", "actual", "diff", "err%")
> 	for _, res := range resolutions {
> 		for _, dur := range durations {
> 			res, dur := res, dur
> 			t.Run(fmt.Sprintf("%s/%ds", res, dur), func(t *testing.T) {
> 				videoURL := fmt.Sprintf("https://lorem.video/%s_h264_%ds", res, dur)
> 				raw, err := download(context.Background(), client, videoURL)
> 				if err != nil {
> 					t.Fatalf("download %s: %v", videoURL, err)
> 				}
> 				dataURL := "data:video/mp4;base64," + base64.StdEncoding.EncodeToString(raw)
> 
> 				estimate := e.placeholderCount(dataURL)
> 				actual, err := liveVideoTokens(context.Background(), client, endpoint, model, dataURL)
> 				if err != nil {
> 					t.Fatalf("query server: %v", err)
> 				}
> 
> 				diff := estimate - actual
> 				var errPct float64
> 				if actual != 0 {
> 					errPct = float64(diff) / float64(actual) * 100
> 				}
> 				t.Logf("%-10s %-6s %10d %10d %10d %7.1f%%", res, fmt.Sprintf("%ds", dur), estimate, actual, diff, errPct)
> 			})
> 		}
> 	}
> }
> 
> // download fetches url and returns its body.
> func download(ctx context.Context, client *http.Client, url string) ([]byte, error) {
> 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
> 	if err != nil {
> 		return nil, err
> 	}
> 	resp, err := client.Do(req)
> 	if err != nil {
> 		return nil, err
> 	}
> 	defer resp.Body.Close()
> 	if resp.StatusCode != http.StatusOK {
> 		return nil, fmt.Errorf("status %d", resp.StatusCode)
> 	}
> 	return io.ReadAll(resp.Body)
> }
> 
> // liveVideoTokens posts a single-video chat completion request and returns the
> // server-reported multimodal video token count from usage.prompt_tokens_details.
> func liveVideoTokens(ctx context.Context, client *http.Client, endpoint, model, videoDataURL string) (int, error) {
> 	reqBody := map[string]any{
> 		"model": model,
> 		"messages": []any{
> 			map[string]any{
> 				"role": "user",
> 				"content": []any{
> 					map[string]any{"type": "text", "text": "describe the video"},
> 					map[string]any{"type": "video_url", "video_url": map[string]any{"url": videoDataURL}},
> 				},
> 			},
> 		},
> 		"max_tokens":  1,
> 		"temperature": 0,
> 	}
> 	body, err := json.Marshal(reqBody)
> 	if err != nil {
> 		return 0, err
> 	}
> 
> 	url := fmt.Sprintf("http://%s/v1/chat/completions", endpoint)
> 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewReader(body))
> 	if err != nil {
> 		return 0, err
> 	}
> 	req.Header.Set("Content-Type", "application/json")
> 
> 	resp, err := client.Do(req)
> 	if err != nil {
> 		return 0, err
> 	}
> 	defer resp.Body.Close()
> 	respBody, err := io.ReadAll(resp.Body)
> 	if err != nil {
> 		return 0, err
> 	}
> 	if resp.StatusCode != http.StatusOK {
> 		return 0, fmt.Errorf("status %d: %s", resp.StatusCode, respBody)
> 	}
> 
> 	var parsed struct {
> 		Usage struct {
> 			PromptTokensDetails struct {
> 				MultimodalTokens struct {
> 					Video int ``json:"video"``
> 				} ``json:"multimodal_tokens"``
> 			} ``json:"prompt_tokens_details"``
> 		} ``json:"usage"``
> 	}
> 	if err := json.Unmarshal(respBody, &parsed); err != nil {
> 		return 0, fmt.Errorf("decode response: %w", err)
> 	}
> 	return parsed.Usage.PromptTokensDetails.MultimodalTokens.Video, nil
> }
> ```
> 
> And I saw your token estimation has large gap when compared to actual
> 
> === RUN TestVideoEstimator_PlaceholderCount_Live mm_estimate_live_test.go:59: resolution dur estimate actual diff err% === RUN TestVideoEstimator_PlaceholderCount_Live/360p/1s mm_estimate_live_test.go:82: 360p 1s 3654 456 3198 701.3% --- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/1s (0.92s) === RUN TestVideoEstimator_PlaceholderCount_Live/360p/5s mm_estimate_live_test.go:82: 360p 5s 3654 1140 2514 220.5% --- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/5s (0.56s) === RUN TestVideoEstimator_PlaceholderCount_Live/360p/10s mm_estimate_live_test.go:82: 360p 10s 3654 2280 1374 60.3% --- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/10s (0.75s) === RUN TestVideoEstimator_PlaceholderCount_Live/360p/20s mm_estimate_live_test.go:82: 360p 20s 3656 4570 -914 -20.0% --- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/20s (0.95s) === RUN TestVideoEstimator_PlaceholderCount_Live/360p/30s mm_estimate_live_test.go:82: 360p 30s 3654 6860 -3206 -46.7% --- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/30s (1.05s) === RUN TestVideoEstimator_PlaceholderCount_Live/360p/60s mm_estimate_live_test.go:82: 360p 60s 3654 11930 -8276 -69.4% --- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/60s (1.77s) === RUN TestVideoEstimator_PlaceholderCount_Live/360p/90s mm_estimate_live_test.go:82: 360p 90s 3654 11600 -7946 -68.5% --- PASS: TestVideoEstimator_PlaceholderCount_Live/360p/90s (2.51s) === RUN TestVideoEstimator_PlaceholderCount_Live/720p/1s mm_estimate_live_test.go:82: 720p 1s 3654 1776 1878 105.7% --- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/1s (0.30s) === RUN TestVideoEstimator_PlaceholderCount_Live/720p/5s mm_estimate_live_test.go:82: 720p 5s 3654 4440 -786 -17.7% --- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/5s (0.59s) === RUN TestVideoEstimator_PlaceholderCount_Live/720p/10s mm_estimate_live_test.go:82: 720p 10s 12560 8880 3680 41.4% --- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/10s (1.30s) === RUN TestVideoEstimator_PlaceholderCount_Live/720p/20s mm_estimate_live_test.go:82: 720p 20s 12568 12050 518 4.3% --- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/20s (1.85s) === RUN TestVideoEstimator_PlaceholderCount_Live/720p/30s mm_estimate_live_test.go:82: 720p 30s 3654 12410 -8756 -70.6% --- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/30s (2.43s) === RUN TestVideoEstimator_PlaceholderCount_Live/720p/60s mm_estimate_live_test.go:82: 720p 60s 3654 11930 -8276 -69.4% --- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/60s (4.91s) === RUN TestVideoEstimator_PlaceholderCount_Live/720p/90s mm_estimate_live_test.go:82: 720p 90s 3654 11600 -7946 -68.5% --- PASS: TestVideoEstimator_PlaceholderCount_Live/720p/90s (6.90s) === RUN TestVideoEstimator_PlaceholderCount_Live/1080p/1s mm_estimate_live_test.go:82: 1080p 1s 3654 4096 -442 -10.8% --- PASS: TestVideoEstimator_PlaceholderCount_Live/1080p/1s (0.77s) === RUN TestVideoEstimator_PlaceholderCount_Live/1080p/5s mm_estimate_live_test.go:74: query server: status 500: {"error":{"message":"Internal server error","type":"InternalServerError","param":null,"code":500}} --- FAIL: TestVideoEstimator_PlaceholderCount_Live/1080p/5s (0.81s)
> 
> === RUN TestVideoEstimator_PlaceholderCount_Live/1080p/10s mm_estimate_live_test.go:82: 1080p 10s 3654 12300 -8646 -70.3% --- PASS: TestVideoEstimator_PlaceholderCount_Live/1080p/10s (1.66s) === RUN TestVideoEstimator_PlaceholderCount_Live/1080p/20s mm_estimate_live_test.go:82: 1080p 20s 12568 12050 518 4.3% --- PASS: TestVideoEstimator_PlaceholderCount_Live/1080p/20s (3.69s) === RUN TestVideoEstimator_PlaceholderCount_Live/1080p/30s mm_estimate_live_test.go:82: 1080p 30s 3654 12410 -8756 -70.6% --- PASS: TestVideoEstimator_PlaceholderCount_Live/1080p/30s (4.32s) === RUN TestVideoEstimator_PlaceholderCount_Live/1080p/60s mm_estimate_live_test.go:82: 1080p 60s 3654 11930 -8276 -69.4% --- PASS: TestVideoEstimator_PlaceholderCount_Live/1080p/60s (9.14s) === RUN TestVideoEstimator_PlaceholderCount_Live/1080p/90s mm_estimate_live_test.go:82: 1080p 90s 3654 11600 -7946 -68.5%
> 
> I think this would be a blocker to merge this PR. Can you see why that happens?
> 
> And I found with const sampleFPS = 2.0 const factor = 32 * 32 frames := int(math.Round(float64(duration) * sampleFPS)) tokensPerFrame := int(math.Round(float64(width*height) / factor)) raw_tokens = min(frames * tokensPerFrame, 12288)
> 
> It has better estimation than your current PR with qwen3 vl model, at the same time I will try to work on another PR to see how thing goes.

Thanks for sharing. I think root cause is parsing, not the formula. `parseVideoTrak()` treats mdhd.duration or stsz.sample_count == 0 as a hard failure (`"invalid mp4 timing metadata"`), and `mm_estimate.go` then falls back to constant defaults (16s / 2fps / 640x360), which plug into the formula as exactly **3654** — matching almost every row in your log regardless of actual duration.

That said, it doesn't fully explain everything: 720p/10s (12560) and 720p+1080p/20s (12568) look like parsing succeeded, yet they're still off from actual by 4-41%. So there seem to be two separate issues:

1. fMP4 clips failing mdhd/stsz parsing and silently falling back to defaults
2. some formula inaccuracy even when metadata parses correctly

For (1), I'd like to open a follow-up PR to add fMP4 support, which should fix the majority of the gap in your benchmark. For (2), how about re-run your benchmark once (1) lands? Let me know what you think.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/mm_estimate.go
```diff
@@ -148,3 +149,224 @@ func imageDimensionsFromBase64Payload(rawB64 string) (width, height int, ok bool
...
+	// minVideoFrames and maxVideoFrames clamp the resampled frame count
+	// (see sampleFrames) so extremely short or long videos still produce a
+	// sane number of temporal grid cells.
+	minVideoFrames = 4
+	maxVideoFrames = 768
+)
+
+// videoEstimator implements the Qwen3-VL-accurate token estimation:
+// vLLM frame cap, smart_resize, temporal patch grouping, and per-grid timestamp tokens.
+// It has no tunable fields: every parameter is a Qwen3-VL/vLLM architecture
```
liu-cong, line 197: what's the plan to support other models?

namgyu-youn: The next support would be Google Gemma 4. But I would like to hold off until token estimation formula is stable; there is still a gap like fMP4 parsing logic. cc @capri-xiyue to hear your thoughts.

capri-xiyue: Similar to image, first milestone is that we will support qwen vl and gemma4, make sure the video token estimation is relatively accurate for these two models as these two are the common models that customers will use. Later we can pick another 1 or 2 other models to see whether the video token estimation is still applicable for them. I will make sure the video token estimation is relatively accurate for both qwen vl and gemma4 in PR https://github.com/llm-d/llm-d-router/pull/1911

namgyu-youn:
@liu-cong gentle ping to reminder, could you take a look into this PR again? It address "video token estimation formula", which is needed for multi-modal approximate prefix cache plugin

ahg-g (approved):
approving based on @capri-xiyue lgtm and review

capri-xiyue:
Actually this PR can be reverted. I sent another PR with better interface and simple formula https://github.com/llm-d/llm-d-router/pull/1911

The code lgtm therefore initially I added /lgtm but when I tried to validated it with real video, I found it has some estimation accuracy issues for some video type, therefore I wrote another PR https://github.com/llm-d/llm-d-router/pull/1911 and make sure the estimation accuracy is good with real video data

ahg-g:
I don't see a revert button here unfortunately. @elevran any idea how to trigger a revert?

PR #1410, opened 2026-05-30
ci: cache image compile layers

sagearc:
## Summary

Split the EPP and sidecar Docker build cache changes out of #1315 so they can be reviewed independently.

Context: https://github.com/llm-d/llm-d-router/pull/1315#issuecomment-4567585374

## What changed

- Precompile the EPP binary before linking commit-specific version metadata.
- Precompile the sidecar binary before linking commit-specific version metadata.
- Keep the final image binaries built with the existing version metadata flags.

## Validation

- `git diff --check`.


Dockerfile.epp
```diff
@@ -30,9 +27,15 @@ COPY internal/ internal/
 COPY pkg/      pkg/
 COPY version/  version/
 
-RUN go mod download
```
ahg-g, line 33: is this not needed anymore?

sagearc: `go mod download` is still there (line 21), after `go.mod`/`go.sum` copy, before source copy. If modules are missing, `go build` will download them implicitly.

ahg-g:
This is the only one left in the refactor, pls fix the test issue to get it in.

PR #1413, opened 2026-05-30
Use unweighted cached-block count for prefix PD decider

### LINKED ISSUE #1047: [Feature] correct RAM-cached  usage in P/D disaggregation

jia-gao:
## Summary

Fixes #1047: the prefix-based PD decider misroutes RAM-cached requests to remote prefill because the precise prefix cache conflates a device-tier-**weighted** ranking score with a literal cached-**block** count.

`precise_prefix_cache.go` stores `int(scores[addr])` — the tier-weighted longest-prefix score from `LongestPrefixScorer` (GPU/HBM = 1.0, CPU/RAM = 0.8 by default) — into `PrefixCacheMatchInfo.matchBlocks`. The decider then computes `MatchBlocks() * BlockSizeTokens()` as cached tokens, so a mostly-RAM-cached prompt undercounts its cached length and is pushed onto remote prefill — the opposite of what a CPU-offload (LMCache) deployment wants. The undercount is worse than 20% at the low end: a short RAM-only hit (e.g. one block scoring 0.8) truncates to **zero** blocks under `int()`.

## Change

Add an unweighted `cachedBlockCount` to `PrefixCacheMatchInfo` (defaulting to `matchBlocks`, so existing producers/consumers are unchanged), populated at the two precise-prefix producer sites by counting the contiguous prefix blocks each endpoint holds — independent of device tier. The decider derives cached tokens from `cachedBlockCount`. The tier-weighted `matchBlocks` is left as-is for the ranking/affinity consumers.

- `data_types.go`: new field + `CachedBlockCount()` getter + `WithCachedBlockCount()` builder + `Clone()` update.
- `precise_prefix_cache.go` (scorer `Produce`) and `dataproducer/preciseprefixcache/producer.go`: set the unweighted count via a small `matchedBlockCount` walk over `keyToPods` (semantically equivalent to `LongestPrefixScorer` with uniform weights).
- `prefix_based_pd_decider.go`: use `CachedBlockCount()` for the cached-token computation.

## Why a separate field (and the alternatives)

The issue proposed two directions; I implemented a third that avoids changing any existing behavior, but this is a judgment call I'd like a maintainer's read on:

1. **Redefine `matchBlocks` to the unweighted count** (the issue's "minimal fix"). Simpler, but it's a **ranking behavior change**: four ratio consumers (`scorer/prefix`, `scorer/latency`, `filter/prefixcacheaffinity`, `predictedlatency`) read `MatchBlocks()/TotalBlocks()` and today prefer GPU-cached over RAM-cached pods via the tier weight; redefining collapses that signal. Not a pure bugfix.
2. **Per-tier counts** (`matchBlocksHBM`/`RAM`, the issue's "better"). Most flexible (tier-aware cost models) but largest surface; arguably scope-creep for this bug.
3. **Separate unweighted field (this PR).** Fixes the decider with zero ranking-behavior change and minimal surface; sidesteps deciding whether the ratio consumers' tier-weighting is intentional.

Happy to switch to (1) or (2) if you'd prefer.

## Out of scope (flagging, not fixing)

`dataproducer/inflightload/producer.go` has the **same misuse class** (`MatchBlocks() * blockSize` as a token count). Per "fix only the issue," I left it untouched and `matchBlocks` keeps its current value there; it would benefit from the same `CachedBlockCount()` switch in a follow-up. No `llm-d-kv-cache` change is needed — the unweighted count is recoverable from the existing `Lookup` data.

## Tests

- `prefix_based_pd_decider_test.go`: regression for the #1047 scenario (blockSize=16, 4096 input tokens, 240-block RAM hit) — the decider stays decode-only with the unweighted count, and a sanity assertion documents that the weighted score alone misroutes.
- `data_types_test.go`: field default/builder/clone.
- `dataproducer/.../utils_test.go`: `matchedBlockCount` walk edge cases (all-RAM counts fully, single block = 1 not 0, stops at first gap, pod absent from block 0 = 0, per-pod independence).

## Verification

- [x] `go build ./...`, `go vet`, `gofmt -l` clean
- [x] `go test` across `scheduling/`, `requestcontrol/`, `datalayer/` plugin trees
- [x] `golangci-lint run` (v2.8.0) — 0 issues


pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/utils.go
```diff
@@ -36,3 +37,26 @@ func extractEndpointSet(endpoints []scheduling.Endpoint) sets.Set[string] {
...
+		held := false
+		for _, entry := range keyToPods[key] {
+			if entry.PodIdentifier == podID {
+				held = true
+				break
+			}
+		}
+		if !held {
+			break
+		}
```
shmuelk: Perhaps a simpler solution:

if !slices.CompareFunc(keyToPods[key],
                                   func(entry kvblock.PodEntry) { return entry.PodIdentifier == podID }) {
    break
}

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
```diff
@@ -387,7 +387,9 @@ func (s *Scorer) Produce(ctx context.Context,
 		}
 		addr := fmt.Sprintf("%s:%s", md.Address, md.Port)
 		matchLen := int(scores[addr])
-		ep.Put(s.prefixMatchDataKey.String(), attrprefix.NewPrefixCacheMatchInfo(matchLen, len(blockKeys), blockSize))
+		cachedBlocks := matchedBlockCount(blockKeys, keyToPods, addr)
+		ep.Put(s.prefixMatchDataKey.String(), attrprefix.NewPrefixCacheMatchInfo(matchLen, len(blockKeys), blockSize).
+			WithCachedBlockCount(cachedBlocks))
```
shmuelk:
```suggestion
		ep.Put(s.prefixMatchDataKey.String(),
		        attrprefix.NewPrefixCacheMatchInfo(matchLen, len(blockKeys), blockSize).WithCachedBlockCount(cachedBlocks))
```

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/utils.go
```diff
@@ -5,12 +5,36 @@ import (
...
+		held := false
+		for _, entry := range keyToPods[key] {
+			if entry.PodIdentifier == podID {
+				held = true
+				break
+			}
+		}
+		if !held {
+			break
+		}
```
shmuelk:
```suggestion
		if !slices.CompareFunc(keyToPods[key],
func(entry kvblock.PodEntry) { return entry.PodIdentifier == podID }) {
break
}
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/utils.go
jia-gao: Done — switched to `slices.ContainsFunc` here. Went with `ContainsFunc` rather than `CompareFunc` since we only need a per-key presence check, not a slice comparison. Thanks!

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/precise_prefix_cache.go
jia-gao: After merging main, this call site is gone: the precise-prefix-cache-scorer is now a deprecated wrapper that delegates to the precise-prefix-cache-producer, so it no longer produces `PrefixCacheMatchInfo`. The `WithCachedBlockCount` call (with the formatting you suggested) now lives only in the producer at `dataproducer/preciseprefixcache/producer.go`.

pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache/utils.go
jia-gao: Same here — this scorer-side helper was removed when I merged main's prefix-cache refactor (the scorer no longer produces match info). The single remaining `matchedBlockCount`, in the producer package, now uses `slices.ContainsFunc` per your other comment.

ahg-g:
Can we please fix CI and move this forward?

jia-gao:
Thanks @ahg-g — should be fixed now. I merged latest `main` to pick up the **go1.25.11** toolchain bump from #1465/#1468, which clears the three govulncheck stdlib CVEs (GO-2026-5037/5038/5039) that were failing `lint`. I also adapted the prefix-based PD decider test to the #1445 token-handling refactor (input token count now comes from the tokenized prompt). All unit tests pass locally and the branch is current with `main`.

The CI jobs are sitting in `action_required` (workflow approval on the fork PR) — could you approve/kick off a run when you get a chance? The earlier `e2e-gaie (traffic)` failure was a `BeforeSuite` model-ready timeout (the `metrics` variant passed), so it should clear on a fresh run.

PR #1417, opened 2026-05-31
fix(apix): correct InferenceModelRewrite field metadata

### LINKED ISSUE #1015: Review and types of fix CRD fields
During the CRD migration, we identified a few areas where the field type definitions could be further refined.
For example:

- apix/v1alpha2/inferencemodelrewrite_types.go:117 
```go
Weight int32 `json:"weight"`
```
better be replaced with
```go
Weight *int32 `json:"weight,omitempty"`
```
This is the same class of bug fixed in PR https://github.com/llm-d/llm-d-inference-scheduler/pull/919.

- inferenceobjective_types.go:73 Priority should be `*int32`,  Kubernetes API conventions use fixed-width integer types (int32, int64) to ensure consistent behavior across platforms. Since this is a CRD field, it should be *int32.

- `InferenceModelRewrite` is missing `+kubebuilder:storageversion`. `InferenceObjective` has // +kubebuilder:storageversion but `InferenceModelRewrite` does not. Every CRD must have exactly one storage version. If this is the only version of InferenceModelRewrite, the annotation should be present.

richl9:
**What type of PR is this?**

/kind bug

`TargetModel.Weight` is documented as optional, but the Go API type used `int32`, which cannot distinguish an omitted weight from an explicitly configured zero value. This changes the field to `*int32` with `omitempty` so omitted weights are represented as `nil`.

This also marks `InferenceModelRewrite` v1alpha2 as the storage version, matching the CRD storage metadata and the pattern already used by `InferenceObjective`.

The weighted model rewrite selection logic and tests are updated to handle nil weights as omitted/unweighted.

Fixes #1015

```release-note
Correct InferenceModelRewrite target weight optionality and storage-version API metadata.

### CODE REVIEW

**Review by ahg-g — COMMENTED** [review]
Is this actually a bug? There is validation that the weight must be set to value between 1 and 1000000, so it should never be not set

**Comment by richl9** [comment]
> Is this actually a bug? There is validation that the weight must be set to value between 1 and 1000000, so it should never be not set

@ahg-g thanks for the comment. I think omission is allowed here. It seems min/max validation applies only when weight is present but does not make the field required. In the generated CRD, weight has minimum: 1 and maximum: 1000000, but the required list for TargetModel only contains modelRewrite, not weight. The Go comment also says weights are optional if all targetModels omit them. Would like @roytman to confirm the intended behavior here as well.

File: pkg/epp/requestcontrol/director.go
Comment (roytman) [inline pkg/epp/requestcontrol/director.go:428]: After looking over the code, I see another issue with **mixed weight**:
- All weights nil (valid per spec): totalWeight == 0, hits the uniform-random fallback. Correct.
- All weights set (valid per spec), weighted selection works normally. Correct.
- Mixed nil and non-nil (invalid per spec, but not enforced at runtime): A target with Weight == nil contributes nothing to currentWeight in the selection loop, so it can never be selected. It's effectively dead weight. The doc says: "If a weight is set for any targetModel, it must be set for all targetModels." So mixed usage violates the contract. But since there's no admission webhook enforcing this, it can happen in practice.

I suggest adding a log about the mixed weights, but the cleanest long-term fix would be validating mixed usage in the controller's reconcile loop (or a CEL validation rule on the CRD), so the warning fires at apply time rather than at request time. 

```goLang
    var totalWeight int32
    var nilCount int 
	for _, model := range models {
		totalWeight += model.Weight
		if model.Weight != nil {
			totalWeight += *model.Weight
		} else {
        nilCount++
        }
	}

	if totalWeight == 0 {
		// If total weight is 0, distribute evenly
		return models[rand.Intn(len(models))].ModelRewrite
	}
	if nilCount > 0 {
         logger.V(logutil.DEFAULT).Info("weighted model selection: mixed nil/non-nil weights, unweighted targets will never be selected",
        "weightedTargets", len(models)-nilCount, "unweightedTargets", nilCount, "totalWeight", totalWeight)
}
```

roytman:
I agree with @richl9, the validators only fire when the value is present in the JSON payload. The problem is goLang cannot tell the difference between "user sent weight: 0" (invalid) and "user omitted weight" (valid, meaning equal distribution).

With Weight int32 (no pointer, no omitempty):

1. User applies a CR without weight:
  ```yaml
  targets:
    - modelRewrite: "model-a"
    - modelRewrite: "model-b"
   ```
2. API server deserializes: Weight = 0 (Go zero value)
3. API server serializes for storage in etcd: always writes "weight": 0 (because no omitempty means zero values are included)
4. Now the stored object has "weight": 0 in it
5. On next read, kubectl shows weight: 0 -- confusing, since Minimum=1 says that's invalid
6. On update/patch, validation sees weight: 0 present in the payload and may reject it against Minimum=1

This is a well-known Kubernetes API design trap: optional numeric fields must use *int32 + omitempty so that "absent" stays absent through the serialize/store/deserialize cycle. Without the pointer, the zero value "leaks" into storage and breaks the round-trip contract.
The runtime selection logic (totalWeight == 0 fallback) happened to work correctly despite this, but the API was publishing invalid state to clients and risking validation failures on updates.

richl9:
Thanks @roytman for the feedback.

Updated the PR to log a warning for mixed omitted/explicit weights and added a unit test for that case.

roytman:
Hi @richl9, thank you for the PR and fixes. The changes look good to me, but the final call is up to @ahg-g.

richl9:
Thanks @roytman. @ahg-g could you review this PR when you get a chance? Thanks!

PR #1424, opened 2026-05-31
ci: implement Phase 2 quality gates and reliability improvements

Jwrede:
## Summary

Implements three of the four Phase 2 items from #956. The fourth item (replace `GHCR_TOKEN` PAT with `GITHUB_TOKEN`) was already completed upstream -- image push already uses `github.token`.

**Coverage regression gate enforcement:**
- `compare-coverage.sh` now exits non-zero when a coverage regression is detected (the `any_regression` variable was already tracked but unused).
- The comparison step keeps `continue-on-error: true` so that e2e tests, cache saves, and artifact uploads still run.
- A final "Enforce coverage gate" step checks the comparison outcome and fails the job on regression.

**Dependency review on PRs:**
- New `ci-dependency-review.yaml` workflow runs `actions/dependency-review-action` on pull requests targeting main, flagging high-severity vulnerabilities introduced by dependency changes.

**Builder image workflow:**
- New `ci-build-builder.yaml` workflow triggers on `Dockerfile.builder` changes.
- On PRs: validates the image builds successfully.
- On merge to main: pushes to GHCR with a content-hash tag and `latest`.

Ref: #956

## Test plan

- [ ] Coverage gate: open a PR that reduces test coverage and verify the job reports regression in the summary and fails at the "Enforce coverage gate" step
- [ ] Dependency review: open a PR that adds a dependency with a known high-severity CVE and verify the check fails
- [ ] Builder image: modify `Dockerfile.builder` on a PR and verify the build step succeeds; merge and verify the image is pushed to GHCR


scripts/compare-coverage.sh
```diff
@@ -113,6 +113,4 @@ if [[ -n "${GITHUB_STEP_SUMMARY:-}" ]]; then
     printf '%s\n' "$output" >> "$GITHUB_STEP_SUMMARY"
 fi
 
-# Exit non-zero only if we ever want hard-failure (threshold > 0 + regression)
-# Currently always 0 per project policy (report only).
-exit 0
+exit $any_regression
```
elevran, line 127: `any_regression` implies any regression? Should we treat some acceptable degradation as cost of doing business (e.g., PR changes code behavior with tests in followup PR, deleted code and tests with high coverage drop the total coverage, ...)?

.github/workflows/ci-build-builder.yaml
```diff
@@ -0,0 +1,51 @@
+name: CI - Build Builder Image
```
elevran: Q:
- does this image need its own action or can be combined with the other actions building and pushing images? In the interest of puting all image related actions into the same workflow. Obviosuly this one would benefit from being gated by changes on the builder dockerfile changes (can it be done using `if`?)
- shuold the makefile (and GH actions, if any) using the builder use the prebuilt image or build it from scratch? I suspect that at least the Makefile targets should use the prebuilt image?

elevran (commented):
@Jwrede - thanks for the quick turnaround!

generally LGTM. Can approve once you had a chance to consider the comments.

.github/workflows/ci-build-builder.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Build builder image
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: Dockerfile.builder
+          push: ${{ github.event_name == 'push' }}
+          tags: |
+            ghcr.io/llm-d/llm-d-builder:latest
+            ghcr.io/llm-d/llm-d-builder:${{ steps.tag.outputs.sha }}
+          cache-from: type=gha,scope=builder
```
elevran: Is caching here aligned with caching used in rest of CICD?

scripts/compare-coverage.sh
Jwrede: Good point. Added a configurable regression tolerance (default 2.0 percentage points). Regressions within the tolerance are reported in the summary table but do not fail the build. This covers the cases you mentioned: deleted high-coverage code, behavior changes with tests in a followup PR, etc.

The tolerance is passed as a fifth argument to the script, so it can be tuned per invocation without editing the workflow.

.github/workflows/ci-build-builder.yaml
Jwrede: Agreed on both points. Removed the standalone workflow. The builder image build can be added later as a conditional job inside ci-build-images.yaml, gated on Dockerfile.builder path changes via an if condition. That keeps all image builds in one place.

On the Makefile question: yes, make targets currently rebuild the builder from scratch every time via the image-build-builder prerequisite. Switching to pull the prebuilt GHCR image would speed up local dev, but that is a separate change since it affects the local workflow for all contributors. Happy to tackle that in a followup if you want.

.github/workflows/ci-build-builder.yaml
Jwrede: Yes, it follows the same pattern as the composite docker-build-and-push action: type=gha with a scope parameter (the composite action uses scope=${{ inputs.image-name }}, this used scope=builder). Moot now since the workflow is removed, but the pattern would carry over if the builder build is added to ci-build-images.yaml later.

elevran:
> On the Makefile question: yes, make targets currently rebuild the builder from scratch every time via the image-build-builder prerequisite. Switching to pull the prebuilt GHCR image would speed up local dev, but that is a separate change since it affects the local workflow for all contributors. Happy to tackle that in a followup if you want.

A follow up would be good. Not sure if one can compare the last modified time of the dockerfile vs the image creation time (is that image attribute even available?) to determine whether it should be rebuilt or pulled.
If we build `dev` builder images off of main branch, the only reason to not use the prebuilt tag is a local modification of the dockerfile. That would not work with Makefile timestamps but use git SHAs instead
- store builder tag using the SHA of the Dockerfile/main
- compare `git hash-object Dockerfile.builder` (current) with `git rev-parse <GIT_SHA>:Dockerfile.builder` or similar (e.g., `git diff --exit-code <GIT_SHA> -- Dockerfile.builder`)

Jwrede:
Thanks for the approval!

The SHA comparison approach makes sense. The Makefile could compute `git hash-object Dockerfile.builder`, check if a GHCR image tagged with that hash exists, and pull it instead of rebuilding. Local modifications would produce a different hash, triggering a local build as a fallback. I will open a separate issue to track this.

PR #1426, opened 2026-05-31
fix: allow model rewrite rules to fill missing model field

### LINKED ISSUE #1359: Support Rewriting missing Model Fields
The `InferenceModelRewrite` is the perfect place to put a potential default if the request completely omits the model parameter of the body. In the (old) documentation references something very similar [here](https://gateway-api-inference-extension.sigs.k8s.io/api-types/inferencemodelrewrite/#generic-wildcard-rewrites), however the current logic checks for existence before running the filters here: https://github.com/llm-d/llm-d-router/blob/93cd291f8cedd1f09f8dadf276d6703d5a767358/pkg/epp/requestcontrol/director.go#L286-L299

Moving the filters before the check would allow for default behavior to be at least configured, without modifying any existing behavior.

**Why is this needed**: Some use cases of ours are migrating from older Inference Engines (i.e. HF TGI) that do not require setting this field, and this would ease that migration. Additionally, other users of llm-D may just want to set a default model should none be specified to allow for easy migrations between models, or shifting traffic around between old/new models.

Jwrede:
**What type of PR is this?**

/kind bug

Reorders `mutateModel` so that `applyWeightedModelRewrite` runs before the model-existence check. Previously, requests without a `model` field in the body were rejected with `BadRequest` immediately, before rewrite rules had a chance to provide a default model name via generic (wildcard) `InferenceModelRewrite` rules.

This broke the [wildcard rewrite pattern](https://gateway-api-inference-extension.sigs.k8s.io/api-types/inferencemodelrewrite/#generic-wildcard-rewrites) documented in the Gateway API Inference Extension spec.

Fixes #1359

```release-note
Requests that omit the model field can now be handled by generic InferenceModelRewrite rules instead of being rejected with BadRequest.
```

**How has this been tested?**

- Existing test "model not found, expect err" still passes (no generic rewrite configured, same behavior as before)
- New test "missing model field resolved by generic rewrite" verifies that a generic rewrite rule fills in the missing model
- `go test ./pkg/epp/requestcontrol/...` passes

### DIFF
--- a/pkg/epp/requestcontrol/director.go
+++ b/pkg/epp/requestcontrol/director.go
@@ -358,16 +358,14 @@ func (d *Director) modelRewriteIfNeeded(ctx context.Context, reqCtx *handlers.Re
 }
 
 func (d *Director) mutateModel(ctx context.Context, reqCtx *handlers.RequestContext, bodyMap map[string]any) (*handlers.RequestContext, error) {
-	var ok bool
-	reqCtx.IncomingModelName, ok = bodyMap["model"].(string)
-	if !ok {
-		return reqCtx, errcommon.Error{Code: errcommon.BadRequest, Msg: "model not found in request body"}
-	}
+	reqCtx.IncomingModelName, _ = bodyMap["model"].(string)
 	if reqCtx.TargetModelName == "" {
-		// Default to incoming model name
 		reqCtx.TargetModelName = reqCtx.IncomingModelName
 	}
 	d.applyWeightedModelRewrite(ctx, reqCtx)
+	if reqCtx.TargetModelName == "" {
+		return reqCtx, errcommon.Error{Code: errcommon.BadRequest, Msg: "model not found in request body"}
+	}
 	bodyMap["model"] = reqCtx.TargetModelName
 	return reqCtx, nil
 }
--- a/pkg/epp/requestcontrol/director.go
+++ b/pkg/epp/requestcontrol/director.go
@@ -110,7 +110,9 @@ type Datastore interface {
 	PoolGet() (*datalayer.EndpointPool, error)
 	ObjectiveGet(objectiveName string) *v1alpha2.InferenceObjective
 	PodList(predicate func(fwkdl.Endpoint) bool) []fwkdl.Endpoint
-	// ModelRewriteGet returns the rewrite rule for a given model name and the name of the InferenceModelRewrite object.
+	// ModelRewriteGet returns the highest-precedence rewrite rule for a given
+	// model name (prioritizing exact matches over generic wildcard rules) and
+	// the name of the InferenceModelRewrite object.
 	ModelRewriteGet(modelName string) (*v1alpha2.InferenceModelRewriteRule, string)
 }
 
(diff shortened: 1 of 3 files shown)


ahg-g:
/assign @zetxqx 

@zetxqx can you please review?

elevran:
@Jwrede the PR description is missing a formatted release note section. Please use the PR template for the section as we rely on specific markings 🙏,

Jwrede:
Updated the PR description with the release note section. Thanks for the heads up.

pkg/epp/requestcontrol/director.go
zetxqx: This comment was not very accurate Could you help update this comment as well?

```
 // ModelRewriteGet returns the highest-precedence rewrite rule for a given model name
// (prioritizing exact matches over generic wildcard rules) and the name of the
// InferenceModelRewrite object.
```

Do you mind also adding the comment here: https://github.com/llm-d/llm-d-router/blob/2072f90aa9848813778e98064ffb33778d667cc5/pkg/epp/datastore/datastore.go#L82

pkg/epp/requestcontrol/director_test.go
```diff
@@ -797,6 +851,14 @@ func TestDirector_HandleRequest(t *testing.T) {
 					director.datastore = mockDs
 					director.endpointCandidates = NewCachedEndpointCandidates(context.Background(), NewDatastoreEndpointCandidates(mockDs), time.Minute)
 				}
```
zetxqx: I don't think this is needed, please check and feel free to remove

pkg/epp/requestcontrol/director_test.go
```diff
@@ -797,6 +851,14 @@ func TestDirector_HandleRequest(t *testing.T) {
...
+				if test.name == "missing model field resolved by generic rewrite" {
+					mockDs := &mockDatastore{
+						pods:     ds.PodList(datastore.AllPodsPredicate),
+						rewrites: []*v1alpha2.InferenceModelRewrite{genericRewrite},
+					}
+					director.datastore = mockDs
+					director.endpointCandidates = NewCachedEndpointCandidates(context.Background(), NewDatastoreEndpointCandidates(mockDs), time.Minute)
+				}
```
zetxqx: nit: can we not hard code the test name instead 

we can Add a  rewrites  field to the table-driven test struct and then in the test body
```                               
                    if len(test.rewrites) > 0 {                                                                           
                        mockDs := &mockDatastore{                                                                             
                            pods:     ds.PodList(datastore.AllPodsPredicate),
                            rewrites: test.rewrites,
                        }
                        director.datastore = mockDs
                        director.endpointCandidates = NewCachedEndpointCandidates(
                            context.Background(),
                            NewDatastoreEndpointCandidates(mockDs),
                            time.Minute,
                        )
                    }
```

zetxqx (commented):
than you, looks good, just a few nits

Jwrede:
Addressed all three points:
- Updated `ModelRewriteGet` comment in both `director.go` and `datastore.go`
- Added `rewrites` field to the table-driven test struct, replaced the hardcoded name checks with `len(test.rewrites) > 0`
- Removed the duplicate mock datastore block

Thanks for the review!

zetxqx (approved):

elevran:
@Jwrede please rebase so the Go vulnerabilities are resolved and lint/test pass

Jwrede:
Rebased onto latest main. Thanks for the heads up.

ahg-g:
sorry @Jwrede there is another conflict that requires resolving

elevran:
@Jwrede lint and tests failing.

elevran (approved):
already aproved - merging after passing tests

Jwrede:
Fixed. The Weight field on TargetModel changed to *int32 upstream; updated the test literal to use ptr.To[int32](100).

PR #1385, opened 2026-05-28
feat: update encoder cache structure and add encoder-cache query and hit metrics

rahulgurnani:
**What type of PR is this?**
/kind feature
/kind cleanup

**What this PR does / why we need it**:
This PR refactors the multimodal encoder-cache producer to use a per-endpoint cache structure and introduces metrics to track cache performance.

#### Key Changes:

1.  **Refactored Cache Architecture**: 
    - Shifted from a global `hash -> pod-set` map to individual LRU caches for each endpoint (`map[string]*lru.Cache`). 
    - This change ensures that eviction is scoped to each pod's specific affinity state, preventing a "noisy neighbor" effect where high churn on one endpoint could evict cache entries for others.

2.  **Memory-Aware Configuration**:
    - Replaced the `cacheSize` (entry count) parameter with `cacheSizeInMB`.
    - Added logic to calculate LRU capacity based on a mebibyte (MiB) budget (defaulting to 2048 MiB), assuming a fixed cost per tracked hash.

3.  **Performance Metrics**:
    - Introduced `llm_d_router_epp_encoder_cache_queries_total`: Tracks the total number of multimodal content lookups.
    - Introduced `llm_d_router_epp_encoder_cache_hits_total` (labeled by `pod`): Tracks successful matches in the LRU.
    - These metrics allow operators to calculate the item-level cache-warm rate for encoder-cache affinity routing.

4.  **Optimized Cleanup**:
    - Simplified the stale pod removal process. Instead of iterating through all hashes to prune pod sets, the producer now simply deletes the entire LRU instance for stale endpoints.

**Testing and validation**
Unit tests.
Metrics validation while running benchmarks:
image
vLLM metrics aggregated across all pods for encoder cache

vs

image

newly added metrics in EPP

As we can see, the number align.

Benchmark numbers at 30 rps:
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-b56025cf-7fff-91b0-c998-330c9c8106b1"><div dir="ltr" style="margin-left:0pt;" align="left">
Metric | Without Encoder Cache | With Encoder Cache
-- | -- | --
p50 E2E (Median) | 0.176 s | 0.161 s
p90 E2E | 0.253 s | 0.223 s
p95 E2E | 0.285 s | 0.242 s
p99 E2E | 5.924 s | 0.286 s
Max E2E | 9.528 s | 0.398 s
Avg E2E | 0.344 s | 0.169 s

</div></b>

[...]

PR #1429, opened 2026-06-01
Export cache hit metrics for encoder cache data producer

### LINKED ISSUE #1384: Export cache hit metrics for encoder cache data producer plugin
Metrics being:
1. Cache hits 
2. Total cache queries = multimodal content seen in the request

These metrics are similar to the metrics that approx prefix cache aware plugin

cc @nicolexin

Jwrede:
## Summary

- Add Prometheus metrics to the multimodal encoder-cache data producer plugin, following the pattern established by the approximate prefix cache plugin
- Three metrics exported under `llm_d_router_epp` subsystem: `encoder_cache_hit_ratio` (histogram), `encoder_cache_queries_total` (counter), `encoder_cache_hits_total` (counter)
- Metrics are recorded per endpoint in `Produce()` and registered via `handle.Metrics()` in `Factory()`

Closes #1384

## Details

The encoder cache data producer (`mm-embeddings-cache-producer`) tracks multimodal content hashes and their pod placements for cache-affinity routing. Previously, operators had no visibility into how effectively the cache was being utilized.

The three new metrics mirror what the approximate prefix cache plugin provides:

- `llm_d_router_epp_encoder_cache_hit_ratio`: Histogram (0.0 to 1.0 in 0.1 steps) of the ratio of matched multimodal items to total items per endpoint lookup. A ratio of 1.0 means all multimodal content in the request was found in that endpoint's encoder cache.
- `llm_d_router_epp_encoder_cache_queries_total`: Counter of total multimodal content items seen across all requests and endpoints.
- `llm_d_router_epp_encoder_cache_hits_total`: Counter of total cache hits (items that matched a pod's encoder cache entry).

The test handle in `producer_test.go` is updated to return a real `prometheus.Registry` instead of nil, matching the pattern used by the approximate prefix cache tests.

## Test plan

- [x] New `metrics_test.go` with tests for registration (including idempotent re-registration), recording with hits, recording with zero total
- [x] Existing `TestFactory` passes with metrics registration
- [x] All 16 tests in the multimodal package pass
- [x] `go build` clean

```release-note
Added `llm_d_router_epp_encoder_cache_hit_ratio` histogram metric.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/metrics.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/metrics.go
@@ -50,6 +50,20 @@ var (
 		[]string{"plugin_type", "plugin_name", "pod", "modality"},
 	)
 
+	// encoderCacheHitRatio records, per endpoint and per request, the fraction of the
+	// request's multimodal items already present in that endpoint's LRU. The counters
+	// give an aggregate hit rate; this histogram exposes the distribution of
+	// per-endpoint hit ratios, which surfaces uneven cache locality across pods.
+	encoderCacheHitRatio = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: eppmetrics.LLMDRouterEndpointPickerSubsystem,
+			Name:      "encoder_cache_hit_ratio",
+			Help:      metricsutil.HelpMsgWithStability("Ratio of matched multimodal items to total items per endpoint in an encoder-cache lookup.", compbasemetrics.ALPHA),
+			Buckets:   []float64{0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0},
+		},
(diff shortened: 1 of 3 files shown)


ahg-g:
pls fix the conflicts and we can move forward with this one I think.

hexfusion:
cc @Jwrede

Jwrede:
Rebased onto main, conflicts resolved. While doing so I noticed the two counters this PR originally added (`encoder_cache_queries_total` and `encoder_cache_hits_total`) have since landed via #1385/#1536/#1576, so I dropped the duplicates. The PR is now scoped down to just the new `encoder_cache_hit_ratio` histogram: the counters give an aggregate hit rate, while this exposes the distribution of per-endpoint hit ratios per request to surface uneven cache locality across pods. It is observed in `Produce` from the matched/total counts already computed per endpoint, and labelled `plugin_type`/`plugin_name` to match the existing metrics.

liu-cong (approved):
The ratio can be derived from the `query` and `hit` metrics. But having an additional `ratio` histogram adds convenience IMO.

PR #1436, opened 2026-06-02
metrics: Add plugin name and type labels to EPP metrics

### LINKED ISSUE #1243: Align on name/label for plugin scoped metrics
We need to align on a standard on how to name or label plugin metrics so that:
- avoid name collision
- ease of discoverability

Should we also prefix the metric name with the plugin type? 
Pros: 
- Name is easier to discover than labels
Cons:
- Long metric name. OpenTelemetry has a 63 char limit, prefixing with the plugin name adds risk.
- Need to deprecate existing metrics not following this pattern
- Standard prometheus metrics has the pattern `namespace_subsystem_name_unit`, further sub slicing with plugin name seems overkill.

**I propose that we label the metrics with `plugin_name` and `plugin_type`, and do NOT add plugin name to the metrics name.**

Other considerations? Once aligned, will update
https://docs.google.com/document/d/18iu0_clvv1O_nPry1x7oLKmtZyaldfGvvBGI3L_fgbY/edit?tab=t.0.

CC @ahg-g @elevran @nilig @vMaroon @rlakhtakia

liu-cong:
Introduce plugin_name and plugin_type labels for the new EPP metrics. This implements tracking of specific endpoint picker plugin details.

**What type of PR is this?**

Fixes #1243

```release-note
Metrics emitted by plugins will have `plugin_name` and `plugin_type` labels.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/metrics_test.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/metrics_test.go
@@ -39,29 +39,44 @@ func TestRecordRequestLatencyMetrics(t *testing.T) {
 	t.Cleanup(resetMetrics)
 	ctx := t.Context()
 
-	require.True(t, recordRequestTTFT(ctx, "model", "target", 0.5))
-	require.True(t, recordRequestPredictedTTFT(ctx, "model", "target", 0.4))
-	require.True(t, recordRequestTTFTPredictionDuration(ctx, "model", "target", 0.1))
-	require.True(t, recordRequestTTFTWithSLO(ctx, "model", "target", 2, 1))
-	require.True(t, recordRequestTPOT(ctx, "model", "target", 0.05))
-	require.True(t, recordRequestPredictedTPOT(ctx, "model", "target", 0.04))
-	require.True(t, recordRequestTPOTPredictionDuration(ctx, "model", "target", 0.2))
-	require.True(t, recordRequestTPOTWithSLO(ctx, "model", "target", 3, 1))
+	require.True(t, recordRequestTTFT(ctx, "test-plugin", "test-type", "model", "target", 0.5))
+	require.True(t, recordRequestPredictedTTFT(ctx, "test-plugin", "test-type", "model", "target", 0.4))
+	require.True(t, recordRequestTTFTPredictionDuration(ctx, "test-plugin", "test-type", "model", "target", 0.1))
+	require.True(t, recordRequestTTFTWithSLO(ctx, "test-plugin", "test-type", "model", "target", 2, 1))
+	require.True(t, recordRequestTPOT(ctx, "test-plugin", "test-type", "model", "target", 0.05))
+	require.True(t, recordRequestPredictedTPOT(ctx, "test-plugin", "test-type", "model", "target", 0.04))
+	require.True(t, recordRequestTPOTPredictionDuration(ctx, "test-plugin", "test-type", "model", "target", 0.2))
+	require.True(t, recordRequestTPOTWithSLO(ctx, "test-plugin", "test-type", "model", "target", 3, 1))
 
 	ttft, err := getHistogram(requestTTFT, "model", "target")
 	require.NoError(t, err)
 	require.Equal(t, uint64(1), ttft.GetSampleCount())
 	require.Equal(t, 0.5, ttft.GetSampleSum())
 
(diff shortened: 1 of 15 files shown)


elevran (commented):
Clean, mechanically consistent change that adds `plugin_name` + `plugin_type` labels to all `llm_d_router_epp_*` metrics across three plugin packages. The old/deprecated metric series are correctly left unchanged. Dependency on #1427 is now merged. One test coverage gap in the predicted-latency package.

elevran:
Q: **`release-note: NONE`.** 
Adding labels to existing ALPHA metrics changes any existing PromQL queries that rely on the label set. The metrics are brand-new (just landed in #1427), so dashboards likely do not exist yet, making this defensible -- but worth a sentence in the description acknowledging the schema change. Consider if we want to mention this PR explicitly in user facing release notes.

pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/metrics_test.go
```diff
@@ -39,44 +39,59 @@ func TestRecordRequestLatencyMetrics(t *testing.T) {
...
+
 	require.Equal(t, 0.5, testutil.ToFloat64(inferenceGauges.WithLabelValues("model", "target", typeTTFT)))
 	require.Equal(t, 0.05, testutil.ToFloat64(inferenceGauges.WithLabelValues("model", "target", typeTPOT)))
 	require.Equal(t, float64(1), testutil.ToFloat64(sloViolationCounter.WithLabelValues("model", "target", typeTTFT)))
 	require.Equal(t, float64(1), testutil.ToFloat64(sloViolationCounter.WithLabelValues("model", "target", typeTPOT)))
+
+	require.Equal(t, 0.5, testutil.ToFloat64(llmdInferenceGauges.WithLabelValues("test-plugin", "test-type", "model", "target", typeTTFT)))
+	require.Equal(t, 0.05, testutil.ToFloat64(llmdInferenceGauges.WithLabelValues("test-plugin", "test-type", "model", "target", typeTPOT)))
+	require.Equal(t, float64(1), testutil.ToFloat64(llmdSloViolationCounter.WithLabelValues("test-plugin", "test-type", "model", "target", typeTTFT)))
+	require.Equal(t, float64(1), testutil.ToFloat64(llmdSloViolationCounter.WithLabelValues("test-plugin", "test-type", "model", "target", typeTPOT)))
```
elevran, line 103: `TestRecordRequestLatencyMetrics` calls `recordRequestPredicted<XYZ>` values, all of which had their `llmd`-prefixed metric calls updated with the new plugin labels, but no assertions are added for the `llmd` variants of those metrics. The test verifies `llmdRequestTTFT` and `llmdRequestTPOT` but leaves the  `llmdRequestPredicted<XYZ>`, variants unchecked.

Add assertions for these after the existing `llmdTpot` block, following the same pattern used for the ones already added.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/metrics.go
```diff
@@ -111,24 +111,24 @@ func registerMetrics(registerer prometheus.Registerer) error {
 // RecordPDDecision increments the counter for a specific P/D routing decision.
 //
 // Deprecated: Use RecordDisaggDecision instead.
-func RecordPDDecision(modelName, decisionType string) {
+func RecordPDDecision(pluginName, pluginType, modelName, decisionType string) {
 	if modelName == "" {
```
elevran, line 115: nit: `RecordPDDecision` and `RecordDisaggDecision` already default `modelName` to `"unknown"` when empty. There is no analogous guard for `pluginName`/`pluginType`. Since values come from `h.typedName.Name`/`h.typedName.Type` that the framework sets at construction, this is not a practical risk -- just a consistency note.

elevran:
/lgtm
holding approval in case you want to fix the test.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/metrics.go
liu-cong: yes the framework already ensures name and type are valid. Not a fan of over defensive code :)

liu-cong:
EDIT: Added a release note. I guess it doesn't hurt to give more info.

RE: release note

Given these are new metrics, and adding additional labels doesn't affect existing queries, no release notes will be added.

PR #1438, opened 2026-06-02
refactor: replace PromptText with EstimatedTokenLength on InferenceRequest

zetxqx:
**What type of PR is this?**
/kind cleanup

This PR replaces the usage of `InferenceRequestBody.PromptText()` with a cleaner, more robust `InferenceRequest.EstimatedTokenLength() (length int64, tokenized bool)` method across all EPP framework plugins.

Previously, plugins had to extract the raw prompt text from the request body and estimate tokens on their own (often with duplicated logic, multiplying by a character-to-token ratio). Additionally, plugins couldn't easily determine if the prompt length was precisely known (e.g. from a tokenized prompt or user-provided hint) or just a rough fallback estimation based on request size.

The new `EstimatedTokenLength()` method encapsulates all this logic in one canonical place:
- It returns a precise count if `TokenizedPrompt` is available on the body.
- It returns a precise count if a valid `InputTokenCountHint()` is present on the body.
- It falls back to estimating token length from the `RequestSizeBytes` field (using a 4 bytes/token ratio) if precise info is missing.
- It returns a `tokenized` boolean flag representing whether the returned length is precise or estimated, allowing downstream components to log or make decisions based on the confidence of the estimate.

This change simplifies plugin code, removes the deprecated `PromptText()` helper, cleans up unused configuration and constants, and provides a unified way to query prompt sizing across all scheduling and rate limiting logic.

Another approach for #1364

Instead of unified Prompt, taking a step back consolidating the request-specific field to use `EstimatedTokenLength`. Then the rest of using request-specific field are only limited on 1. approximate producer 2. tokenizer producer 3. MultiModal producer

```release-note
NONE
```


pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -177,70 +177,6 @@ type MultiModalFeature struct {
...
-				}
-			}
-		}
-		return strings.TrimSpace(sb.String())
-	}
-
-	// Fallback to old fields
-	switch {
-	case r.Completions != nil:
-		return r.Completions.Prompt.PlainText()
```
ahg-g, line 200: As a followup, can we remove the PlainText functions from every message type as well? I only see it used in https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go#L150 for the Completions case only.

pkg/epp/framework/interface/scheduling/types.go
```diff
@@ -76,6 +80,30 @@ func (r *InferenceRequest) String() string {
...
+	if r.Body != nil {
+		if r.Body.TokenizedPrompt != nil && len(r.Body.TokenizedPrompt.TokenIDs) > 0 {
+			return int64(len(r.Body.TokenizedPrompt.TokenIDs)), true
+		}
+		if hint := r.Body.InputTokenCountHint(); hint >= 0 {
+			return int64(hint), true
+		}
+	}
+	if r.RequestSizeBytes > 0 {
+		return max(int64(r.RequestSizeBytes)/bytesPerTokenEstimate, 1), false
```
ahg-g, line 102: - Add a comment and an issue to followup with better estimations for the MM case?
- Should we calculate the estimate more accurately by iterating over the length of every prompt item in every type instead of taking the size of the whole request? will that actually make a measurable difference or not worth the complexity? @liu-cong @kaushikmitr for thoughts?

ahg-g (commented):
nice, this removed half of the direct dependencies on the prompt.

pkg/epp/framework/interface/scheduling/types.go
zetxqx: Added a comparison test in https://github.com/llm-d/llm-d-router/pull/1438/changes/c687bbd4772a2395086c09627fe1a3d31c58aa2c . It will be removed.


  To compare the two approaches  add a test case constructs actual Completions and ChatCompletions requests, marshals them to raw JSON bytes to compute a realistic  RequestSizeBytes , and prints the exact estimates:

  • Approach A (Size-based):  max(RequestSizeBytes / 4, 1)
  • Approach B (Text-based):  max(len(promptText) / 4, 1)

  Here is the exact table generated by running the test on EPP:

    =============================================================================
    Test Case                      | Req Size   | Size Est   | Text Est   | Err %
    -----------------------------------------------------------------------------
    Short Completion Request       | 72         | 18         | 4          | 350.00%
    Long Completion Request        | 1070       | 267        | 254        | 5.12%
    Short Chat Request (1 message) | 113        | 28         | 7          | 300.00%
    Medium Chat Request (2 turns)  | 172        | 43         | 14         | 207.14%
    Long Multi-turn Chat Request   | 555        | 138        | 87         | 58.62%
    =============================================================================
   
   
however please note the original [promptText](https://github.com/llm-d/llm-d-router/blob/a7f83f85d087fe0f33bf49e3ef0be5ef3c4b4ba7/pkg/epp/framework/interface/requesthandling/types.go#L201-L209) only consider the plainText but using chatTemplate([eg](https://github.com/vllm-project/vllm/blob/main/examples/template_chatml.jinja)) we have a lot of fixtures like "assistant" "system" role. So using pure text is a little under-estimate. So I think the bytes estimation is still okay.

pkg/epp/framework/interface/scheduling/types.go
```diff
@@ -76,6 +80,33 @@ func (r *InferenceRequest) String() string {
 		r.RequestID, r.TargetModel, r.Body, r.Headers)
 }
 
+// EstimatedTokenLength returns the estimated token length for the request and a boolean indicating
+// whether the request is tokenized (derived from tokenized prompt or hint).
+// Returns 0, false if the request is nil, or if there is no body and RequestSizeBytes is 0.
+func (r *InferenceRequest) EstimatedTokenLength() (length int64, tokenized bool) {
```
vMaroon, line 86: Why can't this be a producer? Better yet, everything that deals with tokens, contained within the `tokens-producer`.

ahg-g: I thought about modeling this as a producer, the issue is that some producers use this, so we first need to have a solution for producer dependencies before converting this to a producer. 

> Better yet, everything that deals with tokens, contained within the tokens-producer.

That is almost the case here, this is a thin wrapper that first returns the number of tokens from TokenizedPrompt, which is supposed to be populated by tokens-producer (actually this means we need to ensure that tokens-producer is always run first :))!

ahg-g: > Better yet, everything that deals with tokens, contained within the tokens-producer.

In some cases, we don't want to run the tokenizer at all, we just need the token count estimates.

vMaroon: I agree with the latter, but that can be an extension to the tokens-producer contract. I think we should contain all token-related state in there, and unify tokens-in / estimate to the same output so that consumers are not concerned with how the tokens are produced, and do not have to support the two cases.

ahg-g: Agreed, this is certainly much cleaner and more future proof.

PR #1444, opened 2026-06-02
Add support to the EC Nixl connector.

### LINKED ISSUE #1443: Add support to the EC Nixl connector
We need to extend the gateway/sidecar component to support the upcoming Nixl connector for vLLM Encoder-disaggregation (E-disaggregation) topologies.

**Why is this needed**:
Currently, the sidecar only supports the storage shared-based ECExampleConnector. While this works for testing, shared-storage implementations are not suitable for production deployments due to latency overhead. To support production-grade, low-latency disaggregated topologies, the sidecar must natively support the Nixl-based connector.

revit13:
We need to extend the gateway/sidecar component to support the upcoming Nixl connector for vLLM Encoder-disaggregation (E-disaggregation) topologies.

Verified end-to-end against an encoder/prefill setup using [vllm-project/vllm#42998](https://github.com/vllm-project/vllm/pull/42998) on openshift cluster.

Fixes #1443

```release-note
New EC-NIXL encoder disaggregation connector: Use `--ec-connector=ec-nixl` to the sidecar options, to route multimodal encoder requests through NIXL prior to the prefill phase.
OpenTelemetry (OTel) Span Attribute Renaming: Span attributes emitted by the encoder-disaggregation path have been updated from the llm_d.epd_proxy.* namespace to llm_d.ec_proxy.*.
```

### DIFF
--- a/pkg/sidecar/constants/constants.go
+++ b/pkg/sidecar/constants/constants.go
@@ -29,6 +29,6 @@ const (
 	// ECExampleConnector enables the Encoder disaggregation protocol (E/PD, E/P/D)
 	ECExampleConnector = "ec-example"
 
-	// ECEPDConnector enables the Encoder disaggregation Nixl protocol (E/PD, E/P/D)
-	ECEPDConnector = "ec-epd"
+	// ECConnectorNixl enables the Encoder disaggregation Nixl protocol (E/PD, E/P/D)
+	ECConnectorNixl = "ec-nixl"
 )
--- a/pkg/sidecar/constants/constants.go
+++ b/pkg/sidecar/constants/constants.go
@@ -28,4 +28,11 @@ const (
 
 	// ECExampleConnector enables the Encoder disaggregation protocol (E/PD, E/P/D)
 	ECExampleConnector = "ec-example"
+
+	// ECEPDConnector enables an Encoder disaggregation protocol that, in
+	// addition to fan-out priming, captures each encoder's ec_transfer_params
+	// response and threads them into the prefill request body as a black box.
+	// Mirrors the KV-side NIXLv2 connector's pattern of relaying transfer
+	// metadata across stages without inspecting its internal schema.
+	ECEPDConnector = "ec-epd"
 )
--- a/pkg/sidecar/constants/constants.go
+++ b/pkg/sidecar/constants/constants.go
@@ -29,10 +29,6 @@ const (
 	// ECExampleConnector enables the Encoder disaggregation protocol (E/PD, E/P/D)
 	ECExampleConnector = "ec-example"
 
-	// ECEPDConnector enables an Encoder disaggregation protocol that, in
-	// addition to fan-out priming, captures each encoder's ec_transfer_params
-	// response and threads them into the prefill request body as a black box.
-	// Mirrors the KV-side NIXLv2 connector's pattern of relaying transfer
-	// metadata across stages without inspecting its internal schema.
+	// ECEPDConnector enables the Encoder disaggregation Nixl protocol (E/PD, E/P/D)
 	ECEPDConnector = "ec-epd"
 )
--- a/pkg/sidecar/constants/constants.go
+++ b/pkg/sidecar/constants/constants.go
@@ -29,6 +29,6 @@ const (
 	// ECExampleConnector enables the Encoder disaggregation protocol (E/PD, E/P/D)
 	ECExampleConnector = "ec-example"
 
-	// ECConnectorNixl enables the Encoder disaggregation Nixl protocol (E/PD, E/P/D)
-	ECConnectorNixl = "ec-nixl"
+	// ECConnectorNIXL enables the Encoder disaggregation NIXL protocol (E/PD, E/P/D)
+	ECConnectorNIXL = "ec-nixl"
 )
--- a/pkg/sidecar/proxy/connector_ec_common.go
+++ b/pkg/sidecar/proxy/connector_ec_common.go
@@ -1,7 +1,7 @@
 // This file holds the encoder fan-out scaffolding shared by every EC
 // connector: deduplicated multimodal-item extraction and the parallel
 // per-item encoder dispatch loop. Each EC connector
-// (ec-example via fanoutEncoderPrimer, ec-epd via fanoutEncoderCollect)
+// (ec-example via fanoutEncoderPrimer, ec-nixl via fanoutEncoderCollect)
 // supplies its own per-response perItem callback and otherwise reuses
 // these helpers verbatim.
 
@@ -9,12 +9,13 @@ package proxy
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"sync"
 
 	logging "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+	"golang.org/x/sync/errgroup"
 )
 
 // Multimodal content types that need encoder processing.
@@ -139,9 +140,11 @@ func (s *Server) mmItemsForFanout(originalRequest map[string]any, requestID stri
 // callback may return an error to fail the whole fan-out, or nil to
 // accept. perItem may be nil for fire-and-forget primer-style usage.
 //
-// Encoder transport errors and non-2xx responses are hard-fail; the first
-// error from any goroutine is returned, all others discarded.
+// The first goroutine to fail cancels ctx so sibling encoder requests are
+// aborted at the transport layer. Every failure is logged before propagating;
+// grp.Wait returns the first non-nil error.
 func (s *Server) fanoutEncoder(
+	ctx context.Context,
 	originalRequest map[string]any,
 	items []map[string]any,
 	encoderHostPorts []string,
@@ -150,68 +153,56 @@ func (s *Server) fanoutEncoder(
 ) error {
 	s.logger.Info("processing multimodal items", "count", len(items), "requestID", requestID, "encoderHostPorts", encoderHostPorts)
 
-	var wg sync.WaitGroup
-	errChan := make(chan error, len(items))
-
+	grp, gctx := errgroup.WithContext(ctx)
 	for idx, mmItem := range items {
-		wg.Add(1)
-		// We will add more sophisticated Encoder pickup option later.
-		encoderHostPort := encoderHostPorts[idx%len(encoderHostPorts)]
-
-		go func(item map[string]any, hostPort string, itemIdx int) {
-			defer wg.Done()
-
-			encoderRequest := buildEncoderRequest(originalRequest, item)
+		hostPort := encoderHostPorts[idx%len(encoderHostPorts)]
+		grp.Go(func() error {
+			encoderRequest := buildEncoderRequest(originalRequest, mmItem)
 
 			body, err := json.Marshal(encoderRequest)
 			if err != nil {
-				errChan <- fmt.Errorf("failed to marshal encoder request for item %d: %w", itemIdx, err)
-				return
+				err = fmt.Errorf("failed to marshal encoder request for item %d: %w", idx, err)
+				s.logger.Error(err, "encoder fanout", "item", idx, "requestID", requestID)
+				return err
 			}
 
 			encoderHandler, err := s.encoderProxyHandler(hostPort)
 			if err != nil {
-				errChan <- fmt.Errorf("failed to get encoder proxy handler for %s: %w", hostPort, err)
-				return
+				err = fmt.Errorf("failed to get encoder proxy handler for %s: %w", hostPort, err)
+				s.logger.Error(err, "encoder fanout", "item", idx, "requestID", requestID)
+				return err
 			}
 
-			req, err := http.NewRequest("POST", ChatCompletionsPath, bytes.NewReader(body))
+			req, err := http.NewRequestWithContext(gctx, "POST", ChatCompletionsPath, bytes.NewReader(body))
 			if err != nil {
-				errChan <- fmt.Errorf("failed to create encoder request for item %d: %w", itemIdx, err)
-				return
+				err = fmt.Errorf("failed to create encoder request for item %d: %w", idx, err)
+				s.logger.Error(err, "encoder fanout", "item", idx, "requestID", requestID)
+				return err
 			}
 			req.Header.Set("Content-Type", "application/json")
-			req.Header.Set(requestHeaderRequestID, fmt.Sprintf("%s-enc-%d", requestID, itemIdx))
+			req.Header.Set(requestHeaderRequestID, fmt.Sprintf("%s-enc-%d", requestID, idx))
 
-			s.logger.V(logging.DEBUG).Info("sending encoder request", "item", itemIdx, "to", hostPort, "requestID", requestID)
+			s.logger.V(logging.DEBUG).Info("sending encoder request", "item", idx, "to", hostPort, "requestID", requestID)
 
 			pw := &bufferedResponseWriter{}
 			encoderHandler.ServeHTTP(pw, req)
 
 			if isHTTPError(pw.statusCode) {
-				errChan <- fmt.Errorf("encoder request failed for item %d with status %d: %s", itemIdx, pw.statusCode, pw.buffer.String())
-				return
+				err := fmt.Errorf("encoder request failed for item %d with status %d: %s", idx, pw.statusCode, pw.buffer.String())
+				s.logger.Error(err, "encoder fanout", "item", idx, "requestID", requestID)
+				return err
 			}
 
 			if perItem != nil {
-				if err := perItem(itemIdx, pw); err != nil {
-					errChan <- err
-					return
+				if err := perItem(idx, pw); err != nil {
+					s.logger.Error(err, "encoder fanout perItem", "item", idx, "requestID", requestID)
+					return err
 				}
 			}
 
-			s.logger.V(logging.DEBUG).Info("encoder request completed", "item", itemIdx, "requestID", requestID)
-		}(mmItem, encoderHostPort, idx)
+			s.logger.V(logging.DEBUG).Info("encoder request completed", "item", idx, "requestID", requestID)
+			return nil
+		})
 	}
-
-	wg.Wait()
-	close(errChan)
-
-	for err := range errChan {
-		if err != nil {
-			return err
-		}
-	}
-
-	return nil
+	return grp.Wait()
 }
--- a/pkg/sidecar/proxy/connector_ec_common.go
+++ b/pkg/sidecar/proxy/connector_ec_common.go
@@ -206,3 +206,60 @@ func (s *Server) fanoutEncoder(
 	}
 	return grp.Wait()
 }
+
+// runPDPipeline finalizes the post-encoder request and dispatches it to the
+// configured P/D connector or directly to the decoder. The caller has already
+// generated requestID and merged any encoder-side metadata into
+// completionRequest. On JSON-marshal failure, runPDPipeline writes the error
+// response itself (matching the existing handler pattern) and returns.
+func (s *Server) runPDPipeline(
+	w http.ResponseWriter,
+	r *http.Request,
+	completionRequest map[string]any,
+	prefillEndPoint string,
+	requestID string,
+) {
+	// Skip decode-first; the encoder has run and prefill must execute.
+	completionRequest[requestFieldCacheHitThreshold] = 0
+
+	modifiedBody, err := json.Marshal(completionRequest)
+	if err != nil {
+		if err := errorJSONInvalid(err, w); err != nil {
+			s.logger.Error(err, "failed to send error response to client")
+		}
+		return
+	}
+
+	pdRequest := cloneRequestWithBody(r.Context(), r, modifiedBody)
+	pdRequest.Header.Add(requestHeaderRequestID, requestID)
+
+	destination := "decoder"
+	if len(prefillEndPoint) > 0 {
+		destination = "prefiller"
+	}
+
+	// Don't log the full body. Inline base64 images can be MB each.
+	if v := s.logger.V(logging.DEBUG); v.Enabled() {
+		kv := []any{
+			"requestID", requestID,
+			"destination", destination,
+			"prefiller", prefillEndPoint,
+			"bodyBytes", len(modifiedBody),
+		}
+		if ec, ok := completionRequest[requestFieldECTransferParams]; ok {
+			kv = append(kv, requestFieldECTransferParams, truncateLongStrings(ec, 64))
+		}
+		v.Info("forwarding request after encoder", kv...)
+	}
+
+	if len(prefillEndPoint) > 0 {
+		s.logger.V(logging.DEBUG).Info("using P/D protocol after encoder", "prefiller", prefillEndPoint)
+		s.handlePDConnector(w, pdRequest, prefillEndPoint, APITypeChatCompletions)
+		return
+	}
+
+	s.logger.V(logging.DEBUG).Info("no prefiller configured, going directly to decoder after encoder")
+	if !s.forwardDataParallel || !s.dataParallelHandler(w, pdRequest) {
+		s.decoderProxy.ServeHTTP(w, pdRequest)
+	}
+}
--- a/pkg/sidecar/proxy/connector_ec_common.go
+++ b/pkg/sidecar/proxy/connector_ec_common.go
@@ -26,6 +26,31 @@ var mmTypes = map[string]bool{
 	"input_audio": true,
 }
 
+// truncateLongStrings recursively shortens long string values for logging.
+func truncateLongStrings(v any, maxLen int) any {
+	switch x := v.(type) {
+	case string:
+		if len(x) > maxLen {
+			return fmt.Sprintf("%s...(%d bytes)", x[:maxLen], len(x))
+		}
+		return x
+	case map[string]any:
+		out := make(map[string]any, len(x))
+		for k, vv := range x {
+			out[k] = truncateLongStrings(vv, maxLen)
+		}
+		return out
+	case []any:
+		out := make([]any, len(x))
+		for i, vv := range x {
+			out[i] = truncateLongStrings(vv, maxLen)
+		}
+		return out
+	default:
+		return v
+	}
+}
+
 // extractMMItems extracts all multimodal items from the request messages.
 func extractMMItems(requestData map[string]any) []map[string]any {
 	var items []map[string]any
--- a/pkg/sidecar/proxy/connector_ec_common.go
+++ b/pkg/sidecar/proxy/connector_ec_common.go
@@ -151,6 +151,10 @@ func (s *Server) fanoutEncoder(
 	requestID string,
 	perItem func(idx int, pw *bufferedResponseWriter) error,
 ) error {
+	if len(encoderHostPorts) == 0 {
+		return fmt.Errorf("fanoutEncoder: no encoder hostPorts provided (requestID=%s)", requestID)
+	}
+
 	s.logger.Info("processing multimodal items", "count", len(items), "requestID", requestID, "encoderHostPorts", encoderHostPorts)
 
 	grp, gctx := errgroup.WithContext(ctx)
--- a/pkg/sidecar/proxy/connector_ec_nixl.go
+++ b/pkg/sidecar/proxy/connector_ec_nixl.go
@@ -1,6 +1,7 @@
 package proxy
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -45,6 +46,7 @@ func truncateLongStrings(v any, maxLen int) any {
 // NIXLv2 tolerance for missing kv_transfer_params); transport errors,
 // non-2xx, and unparsable JSON are hard-fail.
 func (s *Server) fanoutEncoderCollect(
+	ctx context.Context,
 	originalRequest map[string]any,
 	encoderHostPorts []string,
 	requestID string,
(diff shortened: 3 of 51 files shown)


pkg/sidecar/proxy/connector_epd_ec.go
```diff
@@ -0,0 +1,184 @@
...
+	)
+	err := s.fanoutEncoder(originalRequest, items, encoderHostPorts, requestID, func(idx int, pw *bufferedResponseWriter) error {
+		var encoderResponse map[string]any
+		if err := json.Unmarshal(pw.bodyBytes(), &encoderResponse); err != nil {
+			return fmt.Errorf("failed to parse encoder response for item %d: %w", idx, err)
+		}
+		s.logger.Info("encoder response",
+			"item", idx,
+			"requestID", requestID,
+			requestFieldECTransferParams, truncateLongStrings(encoderResponse[requestFieldECTransferParams], 64))
```
shmuelk: Consider wrapping with if s.logger.Enabled() {...} as the parameters will all be calculated even if the logging is suppressed

pkg/sidecar/proxy/connector_epd_ec.go
```diff
@@ -0,0 +1,184 @@
...
+		return nil, 0, 0, err
+	}
+	return params, contributed, len(items), nil
+}
+
+// handleECEPD fans out per-image encoder requests, aggregates each
+// response's ec_transfer_params into the prefill request body, and hands
+// off to the configured P/D connector.
+func (s *Server) handleECEPD(w http.ResponseWriter, r *http.Request, prefillEndPoint string, encodeEndPoints []string) {
+	s.logger.V(4).Info("running EC-EPD protocol", "prefiller", prefillEndPoint, "encoderCount", len(encodeEndPoints))
```
shmuelk:
```suggestion
	s.logger.V(logging.DEBUG).Info("running EC-EPD protocol", "prefiller", prefillEndPoint, "encoderCount", len(encodeEndPoints))
```

pkg/sidecar/proxy/connector_epd_ec_common.go
```diff
@@ -0,0 +1,227 @@
...
+				return u
+			}
+		}
+	case "video_url":
+		if m, ok := item["video_url"].(map[string]any); ok {
+			if u, ok := m["url"].(string); ok {
+				return u
+			}
+		}
+	}
```
shmuelk: Remove repeated code.

```suggestion
	case "image_url", "audio_url", "video_url":
		if m, ok := item[itemType].(map[string]any); ok {
			if u, ok := m["url"].(string); ok {
				return u
			}
		}
```

pkg/sidecar/proxy/connector_epd_ec_common.go
```diff
@@ -0,0 +1,227 @@
...
+	raw := extractMMItems(originalRequest)
+	if len(raw) == 0 {
+		return nil
+	}
+	seenURLs := make(map[string]struct{})
+	items := make([]map[string]any, 0, len(raw))
+	for _, item := range raw {
+		if url := mmItemURL(item); url != "" {
+			if _, seen := seenURLs[url]; seen {
+				s.logger.V(4).Info("skipping duplicate multimodal URL", "url", url, "requestID", requestID)
```
shmuelk:
```suggestion
				s.logger.V(logging.DEBUG).Info("skipping duplicate multimodal URL", "url", url, "requestID", requestID)
```

pkg/sidecar/proxy/connector_epd_ec_common.go
```diff
@@ -0,0 +1,227 @@
...
+
+			req, err := http.NewRequest("POST", ChatCompletionsPath, bytes.NewReader(body))
+			if err != nil {
+				errChan <- fmt.Errorf("failed to create encoder request for item %d: %w", itemIdx, err)
+				return
+			}
+			req.Header.Set("Content-Type", "application/json")
+			req.Header.Set(requestHeaderRequestID, fmt.Sprintf("%s-enc-%d", requestID, itemIdx))
+
+			s.logger.V(4).Info("sending encoder request", "item", itemIdx, "to", hostPort, "requestID", requestID)
```
shmuelk:
```suggestion
			s.logger.V(logging.DEBUG).Info("sending encoder request", "item", itemIdx, "to", hostPort, "requestID", requestID)
```

pkg/sidecar/proxy/connector_epd_ec_common.go
```diff
@@ -0,0 +1,217 @@
...
+			}
+
+			if perItem != nil {
+				if err := perItem(itemIdx, pw); err != nil {
+					errChan <- err
+					return
+				}
+			}
+
+			s.logger.V(4).Info("encoder request completed", "item", itemIdx, "requestID", requestID)
```
shmuelk:
```suggestion
			s.logger.V(logging.DEBUG).Info("encoder request completed", "item", itemIdx, "requestID", requestID)
```

pkg/sidecar/proxy/connector_epd_shared_storage.go
```diff
@@ -17,200 +17,23 @@ limitations under the License.
...
+// fanoutEncoderPrimer sends concurrent encoder requests for each multimodal
+// item, discarding the responses (status check only). Used by the
+// `ec-example` connector to prime the encoder cache before forwarding the
+// original request to the P/D connector.
 func (s *Server) fanoutEncoderPrimer(originalRequest map[string]any, encoderHostPorts []string, requestID string) error {
-	allItems := extractMMItems(originalRequest)
-	if len(allItems) == 0 {
+	items := s.mmItemsForFanout(originalRequest, requestID)
+	if len(items) == 0 {
 		s.logger.V(4).Info("no multimodal items, skipping encoder", "requestID", requestID)
```
shmuelk: Not connected to the PR, but....


```suggestion
		s.logger.V(logging.DEBUG).Info("no multimodal items, skipping encoder", "requestID", requestID)
```

revit13:
@shmuelk I noticed other V(4) calls in the repo code that could be updated to V(logging.DEBUG) for consistency (and better LLM pattern-matching). I can fix in a follow-up PR. Thanks

pkg/sidecar/proxy/connector_epd_ec.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+		return
+	}
+
+	pdRequest := cloneRequestWithBody(r.Context(), r, modifiedBody)
+	pdRequest.Header.Add(requestHeaderRequestID, requestID)
+
+	// Don't log the full body. Inline base64 images can be MB each.
+	if v := s.logger.V(logging.DEBUG); v.Enabled() {
+		v.Info("forwarding request to prefiller",
```
shmuelk: This statement is not necessarily true, if len(prefillEndPoint) == 0....

pkg/sidecar/proxy/connector_ec_nixl.go
```diff
@@ -0,0 +1,189 @@
...
+
+	if len(prefillEndPoint) > 0 {
+		s.logger.V(logging.DEBUG).Info("using P/D protocol after encoder", "prefiller", prefillEndPoint)
+		s.handlePDConnector(w, pdRequest, prefillEndPoint, APITypeChatCompletions)
+		return
+	}
+
+	s.logger.V(logging.DEBUG).Info("no prefiller configured, going directly to decoder after encoder")
+	if !s.forwardDataParallel || !s.dataParallelHandler(w, pdRequest) {
+		s.decoderProxy.ServeHTTP(w, pdRequest)
```
shmuelk: Why are you sending the same request to both prefill and decode? Should prefill have max-tokes=1 set, which you don't want for decode.

revit13: max_tokens=1 for the prefill leg is set inside the P/D connector, not here. For example, in handleNIXLV2 ([connector_nixlv2.go:108-109](vscode-webview://1d6klgfkjg8qepkfrtk9fq823ol53h4ligmg6sm3ktd99g08r7nf/pkg/sidecar/proxy/connector_nixlv2.go#L108)):


for _, field := range tokenLimitFields {
    completionRequest[field] = 1
}
The original values are saved before that mutation and restored for the decode request. handleECEPD just passes the user's request (with ec_transfer_params and cache_hit_threshold=0 appended) to handlePDConnector — the active P/D connector owns the prefill/decode token-limit lifecycle.

pkg/sidecar/constants/constants.go
```diff
@@ -28,4 +28,7 @@ const (
 
 	// ECExampleConnector enables the Encoder disaggregation protocol (E/PD, E/P/D)
 	ECExampleConnector = "ec-example"
+
+	// ECEPDConnector enables the Encoder disaggregation Nixl protocol (E/PD, E/P/D)
+	ECEPDConnector = "ec-epd"
```
roytman: There is some redundancy in the name. What about 
```golang
ECConnectorNixl = "ec-nixl"
```

pkg/sidecar/proxy/connector_epd_ec.go
```diff
@@ -0,0 +1,194 @@
...
+		params      = make(map[string]any)
+		paramsMu    sync.Mutex
+		contributed int
+	)
+	err := s.fanoutEncoder(originalRequest, items, encoderHostPorts, requestID, func(idx int, pw *bufferedResponseWriter) error {
+		var encoderResponse map[string]any
+		if err := json.Unmarshal(pw.bodyBytes(), &encoderResponse); err != nil {
+			return fmt.Errorf("failed to parse encoder response for item %d: %w", idx, err)
+		}
+		if s.logger.Enabled() {
```
roytman: Enabled() on the base controller-runtime logger is always true at V(0), so this fires at INFO for every multimodal item per request
Suggest to use 
```golang
if v := s.logger.V(logging.DEBUG); v.Enabled()
```
like L:170 below

pkg/sidecar/proxy/connector_epd_ec.go
```diff
@@ -0,0 +1,194 @@
...
+			s.logger.Info("encoder response",
+				"item", idx,
+				"requestID", requestID,
+				requestFieldECTransferParams, truncateLongStrings(encoderResponse[requestFieldECTransferParams], 64))
+		}
+		ec, ok := encoderResponse[requestFieldECTransferParams]
+		if !ok || ec == nil {
+			s.logger.Info("warning: missing ec_transfer_params field in encoder response",
```
roytman: Here and at L:82, L:87, L:95, L:143 and L:148 
s.logger.Info("warning: ...", ...) calls. INFO level for log lines literally prefixed "warning:" reads as a level mismatch. Either demote to V(DEBUG) for the per-item ones (76, 82, 87, 95) or use a real warn logger, leaving the request-level ones (143, 148) at INFO if operator visibility is desired.

pkg/sidecar/proxy/connector_epd_ec_common.go
```diff
@@ -0,0 +1,217 @@
...
+				return
+			}
+
+			encoderHandler, err := s.encoderProxyHandler(hostPort)
+			if err != nil {
+				errChan <- fmt.Errorf("failed to get encoder proxy handler for %s: %w", hostPort, err)
+				return
+			}
+
+			req, err := http.NewRequest("POST", ChatCompletionsPath, bytes.NewReader(body))
```
roytman: it is from the original code, but it is using `http.NewRequest `rather than `http.NewRequestWithContext(r.Context(), ...)`, so client cancellation does not cancel encoder calls.
The Context can be gotten from handleECEPD(w, r, ...) has r.Context()

pkg/sidecar/proxy/connector_ec_common.go
```diff
@@ -0,0 +1,217 @@
...
+// fanoutEncoder fans out one encoder request per item, in parallel, with
+// round-robin over encoderHostPorts. perItem is invoked once per item AFTER
+// the encoder has returned a 2xx response; it receives the item's
+// positional index (post-dedup) and the buffered encoder response. The
+// callback may return an error to fail the whole fan-out, or nil to
+// accept. perItem may be nil for fire-and-forget primer-style usage.
+//
+// Encoder transport errors and non-2xx responses are hard-fail; the first
+// error from any goroutine is returned, all others discarded.
+func (s *Server) fanoutEncoder(
```
roytman, line 171: returns the first error from errChan and silently drops the rest, and waits for all goroutines before surfacing any error.
Two separate concerns to fix; both are solved by switching from sync.WaitGroup + buffered errChan to golang.org/x/sync/errgroup and logging each error before returning.

1. Fail-fast cancellation. errgroup.WithContext cancels its derived context as soon as any goroutine returns a non-nil error. Combined with the http.NewRequestWithContext change from the prior question, in-flight encoder requests on sibling goroutines get canceled at the transport layer.

2. Sibling visibility. Log each error inside the goroutine at the Error level before returning it. The first error still propagates up through `grp.Wait`, but no error is silent.

A sketch
```golang
import "golang.org/x/sync/errgroup"

func (s *Server) fanoutEncoder(
    ctx context.Context,
    originalRequest map[string]any,
    items []map[string]any,
    encoderHostPorts []string,
    requestID string,
    perItem func(idx int, pw *bufferedResponseWriter) error,
) error {
    s.logger.Info("processing multimodal items",
        "count", len(items), "requestID", requestID, "encoderHostPorts", encoderHostPorts)

    grp, gctx := errgroup.WithContext(ctx)
    for idx, mmItem := range items {
        idx, mmItem := idx, mmItem // safe to drop on Go 1.22+
        hostPort := encoderHostPorts[idx%len(encoderHostPorts)]

        grp.Go(func() error {
            encoderRequest := buildEncoderRequest(originalRequest, mmItem)

            body, err := json.Marshal(encoderRequest)
            if err != nil {
                err = fmt.Errorf("marshal encoder request item %d: %w", idx, err)
                s.logger.Error(err, "encoder fanout", "item", idx, "requestID", requestID)
                return err
            }

            encoderHandler, err := s.encoderProxyHandler(hostPort)
            if err != nil {
                err = fmt.Errorf("encoder proxy handler %s: %w", hostPort, err)
                s.logger.Error(err, "encoder fanout", "item", idx, "requestID", requestID)
                return err
            }

            req, err := http.NewRequestWithContext(gctx, "POST", ChatCompletionsPath, bytes.NewReader(body))
            if err != nil {
                err = fmt.Errorf("create encoder request item %d: %w", idx, err)
                s.logger.Error(err, "encoder fanout", "item", idx, "requestID", requestID)
                return err
            }
            req.Header.Set("Content-Type", "application/json")
            req.Header.Set(requestHeaderRequestID, fmt.Sprintf("%s-enc-%d", requestID, idx))

            s.logger.V(logging.DEBUG).Info("sending encoder request",
                "item", idx, "to", hostPort, "requestID", requestID)

            pw := &bufferedResponseWriter{}
            encoderHandler.ServeHTTP(pw, req)

                err := fmt.Errorf("encoder item %d status %d: %s", idx, pw.statusCode, pw.buffer.String())
                s.logger.Error(err, "encoder fanout", "item", idx, "requestID", requestID)
                return err
            }

            if perItem != nil {
                if err := perItem(idx, pw); err != nil {
                    s.logger.Error(err, "encoder fanout perItem", "item", idx, "requestID", requestID)
                    return err
                }
            }

            s.logger.V(logging.DEBUG).Info("encoder request completed",
                "item", idx, "requestID", requestID)
            return nil
        })
    }
    return grp.Wait()
}
```

What changed in behavior:

- The first goroutine to fail cancels gctx. Sibling encoder goroutines using req.WithContext(gctx) see context.Canceled and unwind quickly instead of running to natural completion.
- Every failure path runs s.logger.Error(...) once before returning, so the operator sees N error log lines for N failed items, not just the one that won the race to errChan.
- grp.Wait() returns the first non-nil error verbatim, preserving the existing public contract (fanoutEncoder returns one error or nil).
If preserving every error in the returned value matters (e.g. tests want to assert "all 3 encoders failed"), wrap with errors.Join:
```golang
var (
    mu   sync.Mutex
    errs []error
)
// inside grp.Go: on error, append under mu and return nil so other goroutines run to completion
// after grp.Wait: return errors.Join(errs...)
```
But that loses fail-fast cancellation, so the two goals trade off. The errgroup version above is the better default: cancel siblings, log every failure, return the first.

roytman (commented):
I suggest changing the file names; we don't need the `_epd_` part in each file name, `_ec_` should be enough.
We can have one file `connector_ec_common.go` another one `connector_ec_shared_storage.go` and teh last one `connector_ec_nixl.go` 

What do you think?

pkg/sidecar/proxy/connector_epd_ec.go
revit13: Thanks. Should we add guidance to AGENTS.md about how to express warnings, given that the logger has no native WARN level?

pkg/sidecar/constants/constants.go
```diff
@@ -28,4 +28,7 @@ const (
 
 	// ECExampleConnector enables the Encoder disaggregation protocol (E/PD, E/P/D)
 	ECExampleConnector = "ec-example"
+
+	// ECConnectorNixl enables the Encoder disaggregation Nixl protocol (E/PD, E/P/D)
+	ECConnectorNixl = "ec-nixl"
```
roytman: pkg/sidecar/proxy/options_test.go: all four parametrized rows hardcode ECExampleConnector. No test exercises `ECEPDConnector` at the configuration boundary. A small additional row would lock in that supportedECConnectors and the error-message string both surface the new value.

pkg/sidecar/proxy/proxy.go
```diff
@@ -371,7 +373,9 @@ func (s *Server) setECConnector() {
 
 	switch ecConnector {
 	case ECExampleConnector:
-		s.handleEPDConnector = s.handleEPD
+		s.handleEPDConnector = s.handleECSharedStorage
```
roytman: nit: should it be `s.handleECConnector`? up to you.

Please change the definition comment at L:216 to something like "// handles the Encode disaggregation connector request."

pkg/sidecar/proxy/proxy.go
```diff
@@ -371,7 +373,9 @@ func (s *Server) setECConnector() {
...
+	case ECConnectorNixl:
+		s.handleEPDConnector = s.handleECNIXL
 	default:
 		// Unknown EC connector value, skip encoder stage
```
roytman: Please add a warning here

pkg/sidecar/proxy/connector_ec_common.go
```diff
@@ -0,0 +1,208 @@
...
+	ctx context.Context,
+	originalRequest map[string]any,
+	items []map[string]any,
+	encoderHostPorts []string,
+	requestID string,
+	perItem func(idx int, pw *bufferedResponseWriter) error,
+) error {
+	s.logger.Info("processing multimodal items", "count", len(items), "requestID", requestID, "encoderHostPorts", encoderHostPorts)
+
+	grp, gctx := errgroup.WithContext(ctx)
```
roytman, line 185: errgroup does not bound concurrency. With N multimodal items, fanoutEncoder spawns N goroutines and N concurrent encoder HTTP requests, regardless of len(encoderHostPorts). With many images and few encoder backends, the round-robin assigns multiple in-flight requests to the same backend and can saturate it. Not a regression (prior code had the same shape with sync.WaitGroup), but worth tracking as a follow-up: `grp.SetLimit(min(len(items), maxParallel))` would let an operator cap fan-out width without changing the API.

pkg/sidecar/proxy/connector_ec_common.go
```diff
@@ -0,0 +1,208 @@
...
+	for idx, mmItem := range items {
+		hostPort := encoderHostPorts[idx%len(encoderHostPorts)]
+		grp.Go(func() error {
+			encoderRequest := buildEncoderRequest(originalRequest, mmItem)
+
+			body, err := json.Marshal(encoderRequest)
+			if err != nil {
+				err = fmt.Errorf("failed to marshal encoder request for item %d: %w", idx, err)
+				s.logger.Error(err, "encoder fanout", "item", idx, "requestID", requestID)
```
roytman, line 194: nit:
The error messages at L:165, 172, 179, 192, 198 all use the bare message `encoder fanout`. Three of them additionally include "perItem" in the message. Including a sub-stage (e.g. "encoder fanout marshal", "encoder fanout proxy-handler", "encoder fanout build-request", "encoder fanout transport", "encoder fanout perItem") gives operators a one-glance signal of which step failed without parsing the wrapped error. Optional polish.

roytman (commented):
handleECNIXL still inline-replicates handleECSharedStorage's post-encoder pipeline (connector_ec_nixl.go:157-196). Roughly 30 lines of UUID gen, cache_hit_threshold = 0, json.Marshal, cloneRequestWithBody, requestHeaderRequestID add, prefiller-or-decoder dispatch with dataParallelHandler fallback. Extracting runPDPipeline(w, r, completionRequest, prefillEndPoint, requestID) into a shared helper would collapse both handlers to (encoder step) + runPDPipeline(...). Keeps current behaviour, removes the bug-twin maintenance hazard. Cleanup, not a blocker.

fanoutEncoder still does not retry transient encoder failures. With fail-fast cancellation now wired, a single transient failure on one encoder backend cancels every other encoder request mid-flight. Where the prior implementation papered over this with "wait for everyone, return any error", the new implementation is stricter: a 503 on encoder-2 immediately cancels the work for encoder-1 and encoder-3. Probably the right policy for inference traffic, but worth confirming with the author that fail-fast (rather than per-item retry against a different encoderHostPorts entry) is the desired UX for partial encoder-pool degradation. Not a defect; behaviour change worth noting.

pkg/sidecar/constants/constants.go
elevran: `ECConnectorNixl` should be `ECConnectorNIXL`. All other NIXL identifiers in this codebase use all-caps: `KVConnectorNIXLV2`, `handleECNIXL`. Having the constant use mixed-case while the implementing function uses all-caps is inconsistent.

pkg/sidecar/proxy/connector_ec_nixl.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, 0, 0, err
+	}
+	return params, contributed, len(items), nil
+}
+
+// handleECNIXL fans out per-image encoder requests, aggregates each
+// response's ec_transfer_params into the prefill request body, and hands
+// off to the configured P/D connector.
+func (s *Server) handleECNIXL(w http.ResponseWriter, r *http.Request, prefillEndPoint string, encodeEndPoints []string) {
+	s.logger.V(logging.DEBUG).Info("running EC-EPD protocol", "prefiller", prefillEndPoint, "encoderCount", len(encodeEndPoints))
```
elevran: Stale log message: `"running EC-EPD protocol"` should be `"running EC-NIXL protocol"`.

pkg/sidecar/proxy/connector_ec_common.go
```diff
@@ -0,0 +1,265 @@
...
+	items []map[string]any,
+	encoderHostPorts []string,
+	requestID string,
+	perItem func(idx int, pw *bufferedResponseWriter) error,
+) error {
+	s.logger.Info("processing multimodal items", "count", len(items), "requestID", requestID, "encoderHostPorts", encoderHostPorts)
+
+	grp, gctx := errgroup.WithContext(ctx)
+	for idx, mmItem := range items {
+		hostPort := encoderHostPorts[idx%len(encoderHostPorts)]
```
elevran, line 187: `encoderHostPorts[idx%len(encoderHostPorts)]` panics if `encoderHostPorts` is empty. Callers guard with `if len(encodeEndPoints) > 0` today, but the function has no internal guard. Add an early return at the top of the function body, before the `errgroup` setup.

pkg/sidecar/proxy/proxy.go
```diff
@@ -210,8 +212,8 @@ type Server struct {
 	readyCh            chan struct{} // closed once addr is set and server is listening
 	handler            http.Handler  // the handler function. either a Mux or a proxy
 	allowlistValidator *AllowlistValidator
-	handlePDConnector  pdConnectorHandler  // handles the Prefiller-Decoder connector request
-	handleEPDConnector epdConnectorHandler // handles the Encoder-Prefiller-Decoder connector request
+	handlePDConnector  pdConnectorHandler // handles the Prefiller-Decoder connector request
+	handleECConnector  ecConnectorHandler // handles the Encode disaggregation connector request.
```
elevran: nit: typo should be "handles the Encoder disaggregation" (same as Prefiller, Decoder)

pkg/sidecar/proxy/connector_ec_nixl.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		return nil, 0, 0, err
+	}
+	return params, contributed, len(items), nil
+}
+
+// handleECNIXL fans out per-image encoder requests, aggregates each
+// response's ec_transfer_params into the prefill request body, and hands
+// off to the configured P/D connector.
+func (s *Server) handleECNIXL(w http.ResponseWriter, r *http.Request, prefillEndPoint string, encodeEndPoints []string) {
```
elevran, line 86: Three paths through `handleECNIXL` lack unit test coverage:
1. `encodeEndPoints` is empty and the request goes straight to `runPDPipeline` without encoder calls.
2. Text-only request body (no MM items) with non-empty `encodeEndPoints` so `fanoutEncoderCollect` returns `total=0`, the encoder block is skipped, and the request passes through with `cache_hit_threshold=0` set but no `ec_transfer_params`.
3. `runPDPipeline`'s decoder-direct branch (`prefillEndPoint == ""`) so all existing `TestHandleECEPD*` tests pass `"fake-prefiller:8000"`, so only the prefill path is ever exercised.

pkg/sidecar/proxy/connector_ec_nixl_test.go
```diff
@@ -0,0 +1,494 @@
...
+	threshold, ok := parsed[requestFieldCacheHitThreshold]
+	assert.True(t, ok, "cache_hit_threshold should still be set even when ec params are absent")
+	assert.Equal(t, float64(0), threshold)
+}
+
+// TestHandleECEPDPartiallyPopulated verifies the partial-populated branch:
+// when some encoder responses carry ec_transfer_params and others don't,
+// the connector attaches the flat map containing only the contributed
+// entries (missing items contribute no keys).
+func TestHandleECEPDPartiallyPopulated(t *testing.T) {
```
elevran, line 262: `TestHandleECEPDPartiallyPopulated` uses an atomic counter to alternate even/odd responses, but `fanoutEncoder` dispatches goroutines concurrently so request-arrival order at the backend is non-deterministic. The `assert.Len(t, ec, 1)` assertion can spuriously pass or fail. Use a request-body-keyed or request-ID-keyed backend to route responses deterministically, or relax the assertion to `assert.GreaterOrEqual(t, len(ec), 1)` and add a separate assertion that the partial-contribution warning is logged.

pkg/sidecar/proxy/options_test.go
```diff
@@ -211,7 +211,7 @@ func TestSidecarConfiguration(t *testing.T) {
 				vllmPort:                "8222",
 				dataParallelSize:        2,
 				kvConnector:             KVConnectorSGLang,
-				ecConnector:             ECExampleConnector,
+				ecConnector:             ECConnectorNixl,
```
elevran: nit: This replaces the `ECExampleConnector` value in the "flags override inline YAML" test case rather than adding a `ECConnectorNixl` case alongside it. Both connectors should retain end-to-end option-parsing coverage via the flags path.

elevran (commented):
The EC-NIXL connector is well-structured and the switch from `sync.WaitGroup`+`errChan` to `errgroup` is a clear improvement - thanks!
Left review inline.

Note:
The OTel span attributes `llm_d.epd_proxy.encoder_count`, `llm_d.epd_proxy.encoder_candidates`, and `llm_d.epd_proxy.encode_disaggregation_used` in `chat_completions.go` (lines 158-171) were not renamed alongside the `handleEPDConnector` -> `handleECConnector` rename. Any dashboards or SLOs keyed on these attribute names will silently diverge.

Request: @roytman @revit13 **if possible**. could either of you please resolve the comments after they're addressed as suggested? Please leave unresolved only for discussion. If done, mark as 👍 and then resolve. Otherwise later reviewers have to confirm that issues were actually fixed in later commits.

revit13:
> Note: The OTel span attributes `llm_d.epd_proxy.encoder_count`, `llm_d.epd_proxy.encoder_candidates`, and `llm_d.epd_proxy.encode_disaggregation_used` in `chat_completions.go` (lines 158-171) were not renamed alongside the `handleEPDConnector` -> `handleECConnector` rename. Any dashboards or SLOs keyed on these attribute names will silently diverge.
 
@elevran Renamed llm_d.epd_proxy.* span attributes to llm_d.ec_proxy.* to match the connector rename. This is a breaking change for any dashboards or SLOs keyed on the old names; noted in the release notes. I considered keeping the old names for backwards compatibility but left them out as I understood they were experimental.
Also, added the new EC-NIXL Encoder Connector to the release notes. Thanks

pkg/sidecar/proxy/connector_ec_nixl.go
```diff
@@ -0,0 +1,156 @@
...
+	"net/http"
+	"sync"
+
+	"github.com/google/uuid"
+	logging "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+)
+
+// truncateLongStrings recursively shortens long string values for logging.
+// Field-agnostic. The connector treats ec_transfer_params as opaque.
+func truncateLongStrings(v any, maxLen int) any {
```
elevran: nit: `truncateLongStrings` is defined here but called from `connector_ec_common.go` (`runPDPipeline`). Moving it to the common file would make the dependency clearer as it does not appear to be NIXL specific.

elevran (commented):
/lgtm
/approve
/hold in case you want to address remaining nit - your call.

revit13:
@elevran Thanks I addressed your last comment.

elevran (approved):

PR #1445, opened 2026-06-02
refactor: contain prompt/token handling in the token-producer

vMaroon:
## Summary

`InferenceRequestBody` is a disjoint union of protocol parsings (OpenAI
completions/chat/responses/conversations/embeddings, Anthropic messages, vLLM
generate). Many plugins reach into those protocol fields directly to derive
token counts, prompt text, cache salt, and multimodal content — each
re-implementing parsing and estimation.

This makes **`TokenizedPrompt` the single KV-cache-affecting input
abstraction**. The token-producer is the only consumer of protocol parsings;
every downstream plugin reads `TokenizedPrompt` (`TokenIDs`,
`MultiModalFeatures`, `CacheSalt`) and is unaware of the wire protocol or of
whether tokens are real or estimated.

The disjoint union is kept (this does **not** reintroduce the unified-body
shape from #1380 that #1440 reverted) — parsing stays on-demand inside
producers; only the *consumption* of derived token data is unified.

## What changed

- **Contract.** `TokenizedPrompt` is the KV-cache-affecting input. Removed the
  `PromptText()` / `InputTokenCountHint()` / `CacheSalt()` body accessors;
  `CacheSalt` now lives on `TokenizedPrompt`. Protocol-specific salt extraction
  moved into the token-producer.
- **token-producer backends.** `vllm` /render (real tokenization),
  `udsTokenizerConfig` (deprecated), and `estimate` (tokenizer-free
  byte-packing, ~4 bytes/token, key-preserving; emits MM features with
  image-placeholder weighting). A token-producer with no backend selected
  resolves to `estimate`.
- **estimate is the default producer for `TokenizedPrompt`.** When a config
  consumes `TokenizedPrompt` (prefix cache, context-length, P/D routing)
  without declaring a token-producer, the framework auto-creates one with the
  `estimate` backend. `CreateMissingDataProducers` resolves **transitively** —
  a producer auto-created for one key whose own `Required` deps are unmet
  resolves those too (`prefix-cache-scorer → approx producer → token-producer`).
- **Consumers** read `TokenizedPrompt`: approximate prefix, precise prefix (via
  #1121), context-length, P/D decider, multimodal, in-flight load,
  predicted-latency. Each that can declares `Consumes(TokenizedPrompt)`.
- **Configs.** Estimate-path configs (deploy + e2e/smoke fixtures) no longer
  declare a token-producer at all — it's auto-created. Only precise/render
  configs keep an explicit `vllm` token-producer (required for engine-correlated
  tokens).

## Behavior fixes (silent regressions surfaced and fixed)

- **Tokenizer-less prefix configs** (e.g. the `optimized-baseline` well-lit
  path) regain working prefix caching. Once the approximate producer began
  reading `TokenizedPrompt`, a config without a token-producer silently
  produced no block hashes; auto-creation restores it. Asserted in
  `TestWellKnownConfigs`.
- **E/P/D multimodal:** `always-disagg-multimodal-decider` reads
  `TokenizedPrompt.MultiModalFeatures`; auto-creation means encode-disagg fires
  without a hand-configured tokenizer.

## ⚠️ Gaps / issues / needed followups

1. **Framework change needs a tracking issue.** Transitive auto-creation in
   `CreateMissingDataProducers` (commit `e5b99c4e`) changes shared
   llm-d-router data-layer semantics; it needs its own issue + discussion
   before this is non-WIP.
2. **Precise silent degradation — accepted, not guarded.** Precise and
   approximate share the `TokenizedPrompt` key, one default producer per key. A
   precise config that omits its `vllm` token-producer now silently gets
   `estimate` tokens (no engine correlation → degraded precise hits) instead of
   the previous hard "no producer" error. Documented with a WARNING in the
   token-producer README. **Followup:** a load-time validation that precise
   requires an explicit `vllm` token-producer would restore the loud failure
   without putting estimate/real provenance back onto `TokenizedPrompt`.
3. **`inflightload` can't declare the dependency.** It reads `TokenizedPrompt`
   for token count, but its single `Consumes()` is spent on the
   metrics-attribute interface (`map[string]any`), incompatible with
   `DataDependencies`. Without a co-configured tokenizer its token-load
   estimate reads 0 — graceful degradation, not breakage. **Followup:** split
   the dual-`Consumes` interface framework-wide (also touches
   lora/queue/kvcache/runningrequests) if we want it covered. Not done here.

## Test plan

- **Unit:** token-producer (estimate key-preserving hashing, MM
  image-placeholder weighting, zero-config → estimate);
  `TestCreateMissingDataProducers_Transitive`; all migrated consumers.
- **Integration (envtest):** `TestWellKnownConfigs` (asserts the auto-created
  token-producer for tokenizer-less configs), `TestE2EConfigs_StartupParseSmoke`.
- **CI:** `make presubmit`, lint, e2e shards.

## Commits (on top of the refactor)

- `3a70e68d` feat(token-producer): select estimate backend for zero-config
- `e5b99c4e` feat(datalayer): resolve missing data producers transitively
- `ef2aefe8` feat(token-producer): auto-create as default TokenizedPrompt producer


pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -19,7 +19,7 @@ const (
 	PrefixBasedPDDeciderPluginType = "prefix-based-pd-decider"
 
 	// AverageCharactersPerToken is an estimated average characters per token,
-	// used since the request we cache is not tokenized.
+	// used by tests to derive token counts from character-length prompt fixtures.
```
ahg-g: nit: move to the test file then and don't export it.

pkg/epp/framework/plugins/scheduling/scorer/contextlengthaware/context_length_aware.go
```diff
@@ -194,52 +189,19 @@ func (p *ContextLengthAware) Category() scheduling.ScorerCategory {
...
-	if request.Body.ChatCompletions != nil {
-		for _, msg := range request.Body.ChatCompletions.Messages {
-			totalChars += len(msg.Content.Raw)
-		}
-	}
 
-	// Handle regular completions
-	if request.Body.Completions != nil {
-		totalChars += len(request.Body.Completions.Prompt.Raw)
+	if tp := request.Body.TokenizedPrompt; tp != nil {
```
ahg-g: nit: merge this if statement with the above one

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -71,10 +70,16 @@ type tokenizerPluginConfig struct {
...
 	// VLLM configures the vLLM /render backend.
 	VLLM *vllmConfig `json:"vllm,omitempty"`
+	// Estimate selects the tokenizer-free byte-packing backend; mutually
+	// exclusive with 'vllm'/'udsTokenizerConfig' and needs no 'modelName'.
+	Estimate *estimateConfig `json:"estimate,omitempty"`
 	// ModelName is the name of the model whose tokenizer should be loaded.
 	ModelName string `json:"modelName"`
 }
 
+// estimateConfig selects the estimation backend; no parameters yet.
```
ahg-g: there were a few for the multi-modal encapsulated in multiModalTokenEstimatorConfig

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -71,10 +70,16 @@ type tokenizerPluginConfig struct {
 	TokenizerConfig tokenization.UdsTokenizerConfig `json:"udsTokenizerConfig,omitempty"`
 	// VLLM configures the vLLM /render backend.
 	VLLM *vllmConfig `json:"vllm,omitempty"`
+	// Estimate selects the tokenizer-free byte-packing backend; mutually
+	// exclusive with 'vllm'/'udsTokenizerConfig' and needs no 'modelName'.
+	Estimate *estimateConfig `json:"estimate,omitempty"`
```
ahg-g, line 76: I think a better ux is to have an explicit enum field to select the backend rather than deducing it from the config being set. The reason is that each of them will likely have defaults, and so they may not be empty. This also makes it easier to control the default backend as well.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
```diff
@@ -0,0 +1,137 @@
...
+			MultiModalFeatures: convertMMFeaturesToUpstream(body.Generate.Features),
+		}, nil
+	}
+
+	raw, err := estimateBytes(body)
+	if err != nil {
+		return nil, err
+	}
+	return &fwkrh.TokenizedPrompt{TokenIDs: packBytes(raw)}, nil
+}
```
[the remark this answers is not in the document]
ahg-g: +1

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
```diff
@@ -181,92 +126,3 @@ func getKVCacheBlocksFromTokens(ids []uint32, blockSizeTokens int) iter.Seq[Hash
...
-		} else if len(msg.Content.Structured) > 0 {
-			for _, block := range msg.Content.Structured {
-				switch block.Type {
-				case "text":
-					allPseudoBytes = append(allPseudoBytes, []byte(block.Text)...)
-				case "image_url":
-					// multimodal content can't be in the same pseudo token of text.
-					allPseudoBytes = padToAlignment(allPseudoBytes, averageCharactersPerToken)
-					url := block.ImageURL.URL
-					numPlaceHolders := tokenEstimator.Estimate(fwkrh.ContentBlock{
```
ahg-g, line 229: Where did this logic for placeholder count estimates land?

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/token_estimator.go
```diff
@@ -58,27 +55,13 @@ func (e *SimpleTokenEstimator) Estimate(request *fwksched.InferenceRequest) int6
...
-		hint := request.Body.InputTokenCountHint()
-		if hint >= 0 {
-			return int64(hint)
-		}
-		// Fallback: character count from prompt text across all API types
-		// (completions, chat/completions, responses, conversations).
-		chars := len(request.Body.PromptText())
-		return int64(math.Max(1, math.Round(float64(chars)/e.CharactersPerToken)))
-	default:
+	if request == nil || request.Body == nil || request.Body.TokenizedPrompt == nil {
```
zetxqx, line 61: to enforce the `request.Body.TokenizedPrompt` is always populated before running this, we need to consume the token-producer?

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
```diff
@@ -0,0 +1,174 @@
...
+)
+
+// bytesPerToken matches the scorer's averageCharactersPerToken, so a block of N
+// pseudo-tokens covers the same input bytes as an N-token raw-byte block.
+const bytesPerToken = 4
+
+// estimateBackend packs request bytes into pseudo-tokens with no real tokenizer.
+// The IDs suit content-locality hashing only; they never match engine KV blocks,
+// so pairing this backend with the engine-correlated scorer yields misses, not bad routes.
+type estimateBackend struct{}
```
zetxqx: We also need to add this config for mm: https://github.com/vMaroon/llm-d-inference-scheduler/blob/b3f06f7b6fe9da05190849b98aa3b34ae335cf3a/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go#L147-L149

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/token_estimator.go
vMaroon: Yes. Iterating on PR, will remove WIP when it's in full coverage but more eyes on things can be helpful. We can break it into more consumable pieces later as well.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/mm_estimate.go
```diff
@@ -0,0 +1,71 @@
...
+	// Registers decoders so image.DecodeConfig can read dimensions.
+	_ "image/gif"
+	_ "image/jpeg"
+	_ "image/png"
+)
+
+const (
+	// defaultImageWidth and defaultImageHeight model a 360p image, used when an
+	// image URL is not a decodable base64 payload.
+	defaultImageWidth  = 640
```
capri-xiyue, line 38: For the estimateBackend, I think user should be able to configure it. You can check the release note in https://github.com/llm-d/llm-d-router/pull/1018

for model that the image tokens are fixed(can be set by user) like gemma4,
```
multiModalTokenEstimator:
  image:
    mode: static
    defaultResolution:
      width: 1024
      height: 768
    static:
      staticToken: 512
 ``` 
  This config means default image width and height is 1024*768 when epp can't fetch the image dimension from image url/content , and for each image, it has 512 tokens
 
``` 
multiModalTokenEstimator:
  image:
    mode: dynamic
    defaultResolution:
      width: 1024
      height: 768
    dynamic:
      factor: 1024  
```   
  This config means default image width and height is 1024*768 when epp can't fetch the image dimension from image url/content , and for each image, it has image_width*image_height/factor tokens

File: pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go (hunk 3 above)
Comment (capri-xiyue) [reply pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go]: +1

**Comment by liu-cong** [comment]
Given the size of the refactoring, can you run some micro benchmarks to compare before/after?

We have a benchmark test for approximate: https://github.com/llm-d/llm-d-router/blob/7656f17d4694c74fbb6d2a9abf931f8027823df6/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin_test.go#L881

File: pkg/epp/datalayer/data_graph.go [hunk 10]
```diff
@@ -62,82 +62,98 @@ func ValidateAndOrderDataDependencies(plugins []plugin.Plugin) ([]string, error)
 	return pluginNames, nil
 }
 
-// CreateMissingDataProducers inspects the set of already-configured plugins,
-// finds data keys that are consumed but not yet produced, and auto-instantiates
-// the default DataProducer plugin for each such key using nil parameters.
-// defaultProducerRegistry maps a data key to the plugin type that is its default producer.
+// CreateMissingDataProducers inspects the configured plugins, finds data keys
```
vMaroon, line 65: cc @elevran

vMaroon:
> Given the size of the refactoring, can you run some micro benchmarks to compare before/after?
> 
> We have a benchmark test for approximate:
> 
> https://github.com/llm-d/llm-d-router/blob/7656f17d4694c74fbb6d2a9abf931f8027823df6/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin_test.go#L881

Claude execution:

Ran fair before/after micro-benchmarks. Short version: realistic multi-consumer
configs come out neutral-to-better; a prefix-only config pays a modest cost that's
mostly down to one avoidable allocation (details + fix at the end).

**Settings:** Go 1.26.0, darwin/arm64, Apple M1 Pro · `BlockSizeTokens=16`,
`MaxPrefixBlocksToMatch=50000`, `LRUCapacityPerServer=31250` · estimate
token-producer (tokenizer-free, 4 bytes/token) · `go test -benchmem -count=6`,
benchstat means. Benchmark code: `vMaroon/llm-d-inference-scheduler@token-input-abstraction-bench`.

**1. `BenchmarkPrefixPluginStress` (the one you linked) — standalone, this branch.**
`approximate.Produce` on pre-tokenized input:

| tokens | ns/op  | B/op    | allocs |
|-------:|-------:|--------:|-------:|
| 1024   |  2,495 | 1.48 Ki |     31 |
| 4096   |  5,401 | 2.13 Ki |     33 |
| 10000  | 11,780 | 5.13 Ki |     35 |
| 50000  | 54,240 | 15.5 Ki |     33 |

**2. End-to-end vs upstream**, identical raw prompt of N chars.
before = upstream `approximate.Produce(raw)`; after = `token-producer(estimate)` + `approximate.Produce`:

| chars | before            | after total       | after: estimate / approximate |
|------:|-------------------|-------------------|-------------------------------|
| 1024  | 1.87 µs / 2.33 Ki | 2.62 µs / 3.26 Ki | 0.85 / 1.59 µs                |
| 4096  | 3.18 µs / 5.43 Ki | 5.89 µs / 9.17 Ki | 3.51 / 2.22 µs                |
| 10000 | 5.82 µs / 11.9 Ki | 11.6 µs / 21.9 Ki | 7.89 / 3.84 µs                |
| 50000 | 23.0 µs / 60.9 Ki | 51.1 µs / 117 Ki  | 38.5 / 14.1 µs                |

Single-consumer end-to-end is +40–122%. Note the consumer hot path itself got
*cheaper* (after `approximate` < before serialize+hash); the increase is the
`estimate` stage materializing a token array.

**3. Amortization** — token-derivation cost as token-consumers scale (1 hashing
consumer + N−1 count-readers, as context-length / P/D / in-flight load do).
shared = one `TokenizedPrompt`; perconsumer = each derives its own (pre-refactor):

| chars | shared (any N) | per N=1 | N=2   | N=3   | N=4   |
|------:|---------------:|--------:|------:|------:|------:|
| 1024  |          ~2.46 |    2.47 |  3.47 |  4.40 |  5.42 |
| 4096  |           ~6.2 |    5.97 | 10.19 | 13.54 | 17.71 |
| 10000 |          ~11.7 |   11.96 | 19.20 | 27.11 | 35.17 |
| 50000 |          ~50.0 |   49.66 | 86.82 | 125.6 | 161.4 |

shared stays flat in N; perconsumer grows linearly. At 4 consumers: **2.2×–3.2×
faster, 3.1×–3.9× fewer allocations**. Parity at N=1. Realistic routing configs
(any P/D, prefix + context-length, MM) are multi-consumer, so this is the net win.

**Optimization (follow-up, not this PR).** The single-consumer cost is dominated
by the `estimate` stage, which today copies the prompt string to `[]byte` and then
packs into `[]uint32` — ~2× the input in allocations, two passes. Packing the
string directly into `[]uint32` (skipping the intermediate `[]byte`) roughly halves
that stage (≈38→20 µs and ≈112→56 Ki at 50k), pulling the single-consumer overhead
from ~+122% down to ~+40%. The residual is the irreducible cost of materializing a
reusable token array — which is the point of the refactor: derived once, read by all
consumers.

pkg/epp/datalayer/data_graph.go
```diff
@@ -62,82 +62,98 @@ func ValidateAndOrderDataDependencies(plugins []plugin.Plugin) ([]string, error)
...
 func CreateMissingDataProducers(ctx context.Context, defaultProducerRegistry map[string]string, factoryRegistry map[string]plugin.FactoryFunc, handle plugin.Handle) error {
 	logger := log.FromContext(ctx)
 
-	// Collect all keys already produced by existing plugins.
-	producedKeys := make(map[string]bool)
-	for _, p := range handle.GetAllPlugins() {
-		if producer, ok := p.(plugin.ProducerPlugin); ok {
-			for key := range producer.Produces() {
-				producedKeys[key.String()] = true
+	for {
```
elevran, line 75: This loop-to-fixpoint changes shared framework semantics: a plugin auto-created to satisfy one consumer can itself pull in further plugins. The author flags this as needing a tracking issue and community discussion before the WIP is lifted. The transitive chain test (`TestCreateMissingDataProducers_Transitive`) is a good start; the tracking issue should also address cycle detection (the current "no progress" guard handles the common case but does not explicitly document the cycle behavior).

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -138,12 +178,14 @@ func NewPlugin(ctx context.Context, name string, config *tokenizerPluginConfig)
 		if err != nil {
 			return nil, fmt.Errorf("failed to initialize vLLM HTTP renderer for '%s' plugin - %w", PluginType, err)
 		}
-		tk = renderer
+		backend = renderBackend{tk: renderer}
+	default:
+		backend = estimateBackend{img: newImageEstimator(config.Estimate)}
```
elevran, line 195: This default makes the estimate backend the zero-config path, which silently satisfies a precise prefix-cache scorer's `TokenizedPrompt` dependency with non-engine tokens. The README warning is clear, but a startup-time check that rejects a precise scorer without an explicit `vllm` producer would restore the loud-failure behavior that existed before this PR. Worth tracking as a followup alongside Gap 2.

elevran (commented):
Overall, I think this is a well-structured refactoring: the `TokenizedPrompt` unification removes a complex layer of protocol-aware duplication from consumers. Three things block this from merging (i.e., agree with the WIP label):

1. **Tracking issue for `CreateMissingDataProducers` transitive resolution** (Gap 1 per author). The change to loop-to-fixpoint is a shared framework behavior change. The author explicitly calls this out as requiring its own issue + community discussion before the WIP designation can be lifted.

2. **Silent precise-prefix degradation** (Gap 2). Omitting an explicit `vllm` token-producer in a precise-prefix config previously produced a startup error; now it silently gets `estimate` tokens and degraded cache correlation. The README `[!WARNING]` documents this but a load-time guard (startup error when the precise scorer is present and only an auto-created estimate producer satisfies the dependency) would restore the loud-failure safety net.

3. **`inflightload` token-count regression** (Gap 3). When no token-producer is co-configured, `inflightload.EstimateInput` returns 0, making token-based flow control inoperative. The author marks this as graceful degradation pending a framework change; it should be tracked.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (b renderBackend) produce(ctx context.Context, body *fwkrh.InferenceRequestBody) (*fwkrh.TokenizedPrompt, error) {
+	switch {
+	case body.Completions != nil:
+		tokenIDs, _, err := b.tk.Render(ctx, body.Completions.Prompt.Raw)
+		if err != nil {
+			return nil, fmt.Errorf("tokenization failed: %w", err)
+		}
+		return &fwkrh.TokenizedPrompt{TokenIDs: tokenIDs}, nil
```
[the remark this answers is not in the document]
vMaroon: _Claude Assisted_

Fixed — the render backend passes `Prompt.TokenIDs` through when present, otherwise tokenizes `Prompt.PlainText()` (handles string-array prompts, not just `Raw`). Tests: `TestRenderBackend_CompletionsTokenIDsPassthrough` / `...ArrayUsesPlainText`.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -169,74 +178,25 @@ func (p *Plugin) Produces() map[plugin.DataKey]any {
...
-	case request.Body.Generate != nil:
-		traceLogger.Info("Using pre-tokenized token IDs from generate request", "tokenCount", len(request.Body.Generate.TokenIDs))
-		return &fwkrh.TokenizedPrompt{
-			TokenIDs:           request.Body.Generate.TokenIDs,
-			MultiModalFeatures: convertMMFeaturesToUpstream(request.Body.Generate.Features),
-		}, nil
-	default:
-		return nil, errors.New("unsupported request body type, skipping tokenization")
+		return nil
 	}
```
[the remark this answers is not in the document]
vMaroon: _Claude Assisted_

Fixed — `Produce` sets `TokenizedPrompt.CacheSalt` on the skip path when a parser pre-populated tokens without a salt, so cache-salt isolation still applies. Test: `TestProduce_SetsCacheSaltOnSkipPath`.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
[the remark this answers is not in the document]
vMaroon: _Claude Assisted_

Fixed — the estimate backend passes token-ID inputs through for `/v1/completions` and `/v1/embeddings` (alongside generate) instead of byte-estimating. Tests: `TestEstimateBackend_{Completions,Embeddings}TokenIDsPassthrough`.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		for _, block := range msg.Content.Structured {
+			switch block.Type {
+			case "text":
+				out = append(out, []byte(block.Text)...)
+			case "image_url":
+				out = align(out)
+				h := make([]byte, bytesPerToken)
+				binary.LittleEndian.PutUint32(h, uint32(xxhash.Sum64([]byte(block.ImageURL.URL))))
+				out = append(out, h...)
```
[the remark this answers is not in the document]
vMaroon: _Claude Assisted_

Outdated — the chat path was reworked; placeholder bytes go through `appendMMAsset` now.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
vMaroon: _Claude Assisted_

Done — removed from the decider; unexported to `averageCharactersPerToken` and defined in the test files that use it.

pkg/epp/framework/plugins/scheduling/scorer/contextlengthaware/context_length_aware.go
vMaroon: _Claude Assisted_

Done.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
vMaroon: _Claude Assisted_

Restored on `estimate.image`, matching the #1018 shape: `mode: static|dynamic`, `defaultResolution`, nested `static.staticToken` and `dynamic.factor`.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
vMaroon: _Claude Assisted_

Kept deduce-from-config here for backwards compatibility — existing configs have no `backend` field and a no-backend config resolves to `estimate`. An explicit `backend` selector is additive (optional, with deduction as the fallback), so a clean follow-up rather than part of this PR. Filed: #1472.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/hashing.go
vMaroon: _Claude Assisted_

Into the estimate backend — `imageEstimator.placeholderCount` in `mm_estimate.go`, emitted as multimodal features on `TokenizedPrompt`, configurable via `estimate.image`.

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/token_estimator.go
vMaroon: _Claude Assisted_

Resolved — the token-producer is registered as the default producer for `TokenizedPrompt` and auto-created (transitively) when a plugin declares `Consumes(TokenizedPromptDataKey)`; the token-consumers all declare it. `inflightload` is the exception — its single `Consumes()` is the metrics-attribute interface, so the nil guard stays defensive (returns 0); it is co-deployed with a consumer that pulls the tokenizer in. Splitting that dual-`Consumes` is a possible follow-up.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
vMaroon: _Claude Assisted_

Added on `estimate.image` (`mode`, `defaultResolution`, `static.staticToken`, `dynamic.factor`), matching the #1018 shape.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/mm_estimate.go
vMaroon: _Claude Assisted_

Done — `estimate.image` now matches the #1018 shape: `mode: static|dynamic`, `defaultResolution`, nested `static.staticToken` and `dynamic.factor`. See `estimate_test.go` for the static/dynamic/factor cases.

pkg/epp/datalayer/data_graph.go
vMaroon: _Claude Assisted_

Agreed — filed a tracking issue (#1470) covering cycle behavior. Termination today is a no-progress guard bounded by the finite default-producer registry, not explicit cycle detection; the issue will decide whether explicit detection is added.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
vMaroon: _Claude Assisted_

Right — this is the accepted Gap 2 tradeoff: estimate is the zero-config default and a precise scorer without an explicit `vllm` producer silently gets non-engine tokens. The README carries the warning; the startup-time check is the mitigation, tracked in #1471.

ahg-g:
What are the next steps for validation? benchmarks with and without this PR? to ensure we have no regressions?

capri-xiyue:
For next step validation,
can you run the 
1. https://github.com/llm-d/llm-d/tree/main/guides/optimized-baseline for text
2. optimized baseline for multimodal(image) https://github.com/llm-d/llm-d/pull/1620
to validate there is no regression?

update:
I've verified with optimized baseline for multimodal(image) https://github.com/llm-d/llm-d/pull/1620, this PR's performance looks good on the dataset

ahg-g:
@liu-cong @capri-xiyue evaluated this on optimized-baseline benchmarks for both text and MM and it looks good.

@kaushikmitr please confirm if predicted latency is also ok.
@jtechapps please confirm if scale testing shows no regressions for this PR

jtechapps:
I tested 3 different images v0.8.0, this pr, and main.

Using optimize baseline with different prefix block sizes.

## Results

For the 100k/1k configurations with large block size it looks like the PR image container and main image are similar performance and both less CPU but more memory compared to v0.8.0.

| System | Question | Output | Request Rate | Image | Prefix Blocks | CPU | Memory |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| 24k | 1k | 1k | 95.8/100 | 0.8.0 | unspecified | 23.89 | 1.64 |
| 24k | 1k | 1k | 95.8/100 | pr-1445 | unspecified | 22.83 | 3.00 |
| 24k | 1k | 1k | 95.8/100 | main | unspecified | 21.33 | 2.66 |
| 95k | 5k | 1k | 47.5/50 | 0.8.0 | unspecified | 13.51 | 3.07 |
| 95k | 5k | 1k | 47.5/50 | pr-1445 | unspecified | 10.92 | 5.43 |
| 95k | 5k | 1k | 47.6/50 | main | unspecified | 13.12 | 5.14 |
| 24k | 1k | 1k | 95.8/100 | 0.8.0 | 1563 | 31.89 | 1.9 |
| 24k | 1k | 1k | 95.8/100 | pr-1445 | 1563 | 23.07 | 3.04 |
| 24k | 1k | 1k | 95.8/100 | main | 1563 | 26.87 | 2.67 |
| 95k | 5k | 1k | 47.6/50 | 0.8.0 | 6250 | 19.22 | 3.88 |
| 95k | 5k | 1k | 47.6/50 | pr-1445 | 6250 | 16.06 | 6.15 |
| 95k | 5k | 1k | 47.7/50 | main | 6250 | 16.39 | 5.33 |

scheduler_e2e_duration_seconds p50 latency was 0.0001s for all configurations.

## Optimize baseline
```
inferenceExtension:
  extraServicePorts:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 8081
  pluginsConfigFile: optimized-baseline-plugins.yaml
  pluginsCustomConfig:
    optimized-baseline-plugins.yaml: |
      kind: EndpointPickerConfig
      plugins:
      - type: queue-scorer
      - type: kv-cache-utilization-scorer
      - type: approx-prefix-cache-producer
        parameters:
          maxPrefixBlocksToMatch: 1563
      - type: prefix-cache-scorer
      - type: no-hit-lru-scorer
      schedulingProfiles:
      - name: default
        plugins:
        - pluginRef: queue-scorer
          weight: 2
          weight: 2
          weight: 3
        - pluginRef: no-hit-lru-scorer
          weight: 2
inferencePool:
  modelServers:
    matchLabels:
      app: llm-d-sim
```

llm-d-sim
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: llm-d-sim
  namespace: llm-d-sim
  labels:
    app: llm-d-sim
spec:
  replicas: 50
  selector:
    matchLabels:
      app: llm-d-sim
  template:
    metadata:
      labels:
        app: llm-d-sim
    spec:
      initContainers:
      # vLLM render sidecar - handles tokenization via HTTP (vllm launch render)
      # restartPolicy: Always makes this a native sidecar (K8s 1.29+):
      # vllm-sim will not start until this container's readinessProbe passes
      - name: vllm-render
        image: vllm/vllm-openai-cpu:v0.19.1
        imagePullPolicy: IfNotPresent
        restartPolicy: Always
        command: ["vllm"]
        args: ["launch", "render", "Qwen/Qwen3-8B", "--port=8082", "--max-model-len=131072"]
        env:
        - name: VLLM_ALLOW_LONG_MAX_MODEL_LEN
          value: "1"
        - name: HF_TOKEN
          valueFrom:
            secretKeyRef:
              name: hf-secret
              key: token
        ports:
        - name: http-render
          containerPort: 8082
        resources:
          requests:
            cpu: "4"
            memory: "8Gi"
        livenessProbe:
          httpGet:
            path: /health
            port: 8082
          initialDelaySeconds: 300
          periodSeconds: 10
          failureThreshold: 10
        readinessProbe:
          httpGet:
            path: /health
            port: 8082
          initialDelaySeconds: 300
          periodSeconds: 5
          failureThreshold: 10
      containers:
      - name: vllm-sim
        image: ghcr.io/llm-d/llm-d-inference-sim:v0.9.0
        imagePullPolicy: IfNotPresent
        args:
        - "--model"
        - "Qwen/Qwen3-8B"
        - "--port"
        - "8000"
        - "--max-model-len"
        - "131072"
        - "--max-num-seqs"
        - "256"
        - "--max-waiting-queue-length"
        - "2048"
        - "--inter-token-latency"
        - "12ms"
        - "--prefill-time-per-token"
        - "20us"
        - "--prefill-overhead"
        - "5ms"
        - "--enable-kvcache"
        - "--kv-cache-size"
        - "1000000"
        - "--block-size"
        - "16"
        env:
        - name: POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
        - name: POD_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        - name: POD_IP
          valueFrom:
            fieldRef:
              fieldPath: status.podIP
        ports:
        - containerPort: 8000
          name: http
          protocol: TCP
        resources:
          requests:
            cpu: "4"
            memory: "8Gi"
        livenessProbe:
          httpGet:
            path: /health/ready
            port: 8000
          initialDelaySeconds: 5
          periodSeconds: 10
        readinessProbe:
          httpGet:
            path: /health/ready
            port: 8000
          initialDelaySeconds: 5
          periodSeconds: 5
```

ahg-g:
I am good with merging this PR, feel free to remove the WIP.

ahg-g:
On a separate note, we should report and monitor the epp's cpu and memory profile in nightly benchmarks.

vMaroon:
Ready to go

PR #1448, opened 2026-06-02
feat: support multiple parsers in EPP config

### LINKED ISSUE #1203: Make EPP parser configurable
## Context

The `EndpointPickerConfig` currently supports only a single `Parser` field (`*ParserConfig`), which holds a single `PluginRef` string. This means each EPP instance can only be configured with one parser at a time (e.g., `openai-parser` or `vllmgrpc-parser`), and cannot handle multiple API protocols simultaneously.

As work progresses on adding Messages API support (#943), the need to support multiple parsers per EPP has become clear – an EPP should be able to parse both OpenAI Chat Completions and Anthropic Messages API requests, for example, without requiring separate EPP deployments for each protocol.

## Problem

- `ParserConfig` in `endpointpickerconfig_types.go` is a single `PluginRef`, restricting the EPP to one parser.
- There is no mechanism to route an incoming request to the appropriate parser based on the request's protocol or content type.
- Users who need to serve multiple API formats (e.g., OpenAI + Anthropic Messages) from a single EPP cannot do so today.

## Desired Outcome

- Allow multiple parsers to be configured per EPP.
- Implement a selection mechanism (e.g., based on request path, content type, or headers) that routes each incoming request to the correct parser.
- Maintain backward compatibility – if only one parser is configured (or none), the current behavior and defaults (`openai-parser`) should be preserved.

## Scope

- Update `ParserConfig` / `EndpointPickerConfig` to support multiple parser references.
- Define and implement the request-to-parser matching logic.
- Update config validation and loader.
- Update documentation and example configs.
- Add tests covering multi-parser selection.

## Related

- #943 – [parser] add messages API support

tessapham:
**What type of PR is this?**
/kind feature

Support multiple parsers to be configured in EPP.

Fixes #1203.

The EPP configuration now supports multiple parsers via the `requestHandler.parsers` field, enabling a single EPP to handle multiple API protocols (e.g., OpenAI + Anthropic Messages) without separate deployments. The previous `requestHandler.parser` (singular) field is deprecated but remains supported for backward compatibility; if both are set, `parsers` takes precedence. The health check now validates protocol support across all configured parsers.


apix/config/v1alpha1/endpointpickerconfig_types.go
```diff
@@ -306,18 +306,22 @@ func (dle DataLayerExtractor) String() string {
 // RequestHandlerConfig contains the configuration for incoming request handling.
 type RequestHandlerConfig struct {
 	// +optional
-	// Parser specifies the parsing logic used by the EPP to process protocol messages.
+	// Parsers specifies the parsing plugins used by the EPP to process protocol messages.
 	// If unspecified, default parsing behavior will be applied.
-	Parser *ParserConfig `json:"parser,omitempty"`
+	Parsers []ParserConfig `json:"parsers,omitempty"`
```
zetxqx, line 311: for backward compatibility can we add we new Parsers field? and mark the old one as deprecate. And we should handle convert the parser to parsers. See: https://github.com/llm-d/llm-d-router/blob/ac53bff3f7649387eba7807e0a1e088ddb49e70e/DEVELOPMENT.md?plain=1#L898

cmd/epp/runner/runner.go
```diff
@@ -395,7 +395,7 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 
 	// --- Add Runnables to Manager ---
 	// Register health server.
-	if err := registerHealthServer(mgr, ctrl.Log.WithName("health"), ds, opts.GRPCHealthPort, isLeader, opts.EnableLeaderElection, r.parser); err != nil {
+	if err := registerHealthServer(mgr, ctrl.Log.WithName("health"), ds, opts.GRPCHealthPort, isLeader, opts.EnableLeaderElection, r.parsers[0]); err != nil {
```
zetxqx: instead of just using the first parser can we loop through all the parsers? if there is one protocol misMatch we should log error and fail the health check

pkg/epp/config/loader/configloader.go
```diff
@@ -104,17 +104,17 @@ func LoadRawConfig(configBytes []byte, logger logr.Logger) (*configapi.EndpointP
...
 				rawConfig.RequestHandler = &configapi.RequestHandlerConfig{}
 			}
-			if rawConfig.RequestHandler.Parser == nil {
-				//nolint:staticcheck // SA1019: rawConfig.Parser is deprecated: use requestHandler.Parser instead.
+			if len(rawConfig.RequestHandler.Parsers) == 0 {
+				//nolint:staticcheck // SA1019: rawConfig.Parser is deprecated: use requestHandler.parsers instead.
 				// If both are set, the new field is used. Tracked in https://github.com/llm-d/llm-d-router/issues/1308 (staticcheck)
-				rawConfig.RequestHandler.Parser = rawConfig.Parser
+				rawConfig.RequestHandler.Parsers = []configapi.ParserConfig{*rawConfig.Parser}
 			}
```
zetxqx, line 118: can we also check rawConfig.RequestHandler.Parser as well for backward compatibility

zetxqx (commented):
one major comment is we should keep Parser field and make sure we convert Parser to Parsers.

pkg/epp/config/loader/configloader.go
tessapham: Added check.

apix/config/v1alpha1/endpointpickerconfig_types.go
tessapham: Restored and marked `Parser` field as deprecated.

zetxqx:
/lgtm

looks good to me, thanks!

zetxqx:
cc: @elevran @ahg-g  this change is changing the config.requestHandler.parser -> config.requestHandler.parsers to support multiple parsers

apix/config/v1alpha1/endpointpickerconfig_types.go
```diff
@@ -306,9 +306,16 @@ func (dle DataLayerExtractor) String() string {
 // RequestHandlerConfig contains the configuration for incoming request handling.
 type RequestHandlerConfig struct {
 	// +optional
-	// Parser specifies the parsing logic used by the EPP to process protocol messages.
-	// If unspecified, default parsing behavior will be applied.
+	// Parser specifies a single parsing plugin used by the EPP to process protocol messages.
+	//
+	// Deprecated: use parsers instead. If both are set, parsers is used.
+	// Tracked in https://github.com/llm-d/llm-d-router/issues/1308
```
ahg-g: this has not been released yet, we can just remove it

pkg/epp/config/loader/configloader.go
```diff
@@ -104,17 +104,27 @@ func LoadRawConfig(configBytes []byte, logger logr.Logger) (*configapi.EndpointP
...
-				//nolint:staticcheck // SA1019: rawConfig.Parser is deprecated: use requestHandler.Parser instead.
+			if len(rawConfig.RequestHandler.Parsers) == 0 {
+				//nolint:staticcheck // SA1019: rawConfig.Parser is deprecated: use requestHandler.parsers instead.
 				// If both are set, the new field is used. Tracked in https://github.com/llm-d/llm-d-router/issues/1308 (staticcheck)
-				rawConfig.RequestHandler.Parser = rawConfig.Parser
+				rawConfig.RequestHandler.Parsers = []configapi.ParserConfig{*rawConfig.Parser}
+			}
+		}
+
+		//nolint:staticcheck // SA1019: rawConfig.RequestHandler.Parser is deprecated: use requestHandler.parsers instead.
```
ahg-g: pls remove, no need for this since it hasn't been released yet.

pkg/epp/handlers/response.go
```diff
@@ -48,7 +48,7 @@ func (s *StreamingServer) HandleResponseBody(ctx context.Context, reqCtx *Reques
 
 	reqCtx.ResponseSize += len(responseBytes)
 
-	parsedResp, err := s.parser.ParseResponse(ctx, responseBytes, reqCtx.Response.Headers, endOfStream)
+	parsedResp, err := s.parsers[0].ParseResponse(ctx, responseBytes, reqCtx.Response.Headers, endOfStream)
```
ahg-g, line 51: This is only using the first parser, so we are not technically supporting multiple parsers.

pkg/epp/handlers/server.go
```diff
@@ -343,7 +343,7 @@ func (s *StreamingServer) Process(srv extProcPb.ExternalProcessor_ProcessServer)
 				reqCtx.RequestSize = buf.Len()
 				buf.Reset()
 
-				parseResult, parseErr := s.parser.ParseRequest(ctx, reqCtx.Request.RawBody, reqCtx.Request.Headers)
+				parseResult, parseErr := s.parsers[0].ParseRequest(ctx, reqCtx.Request.RawBody, reqCtx.Request.Headers)
```
ahg-g, line 346: ditto, is the plan to follow up with another PR to select the actual parser to call?

pkg/epp/config/loader/configloader.go
tessapham: Thanks for pointing that out. Just saw that `Parser` wasn't pulled into the v0.8 release.

pkg/epp/handlers/server.go
tessapham: Yes, there will be a follow-up PR to implement the actual parser selection logic: iterate through all parsers and call `Match()` until one returns true.

Since there isn't any selected parser, I'm defaulting it to `parsers[0]` here.

pkg/epp/handlers/response.go
tessapham: This PR only supports processing multiple parsers as input from the EPP config, so this is hard-coded as the first parser for now.

ahg-g:
thanks, pls fix ci, otherwise this looks good to me

tessapham:
@ahg-g Thanks! CI is fixed.

PR #1457, opened 2026-06-02
Add script and Makefile target to check simulator image does not use :latest

### LINKED ISSUE #1306: Add a script and Makefile target to check that YAML files do not use the 'latest' tag as the default for the simulator

apollofps:
**What type of PR is this?**
/kind cleanup

Adds `scripts/check-latest-tags.sh` which scans YAML files across the repo for
`llm-d-inference-sim:latest` references and exits non-zero if any are found.
A `check-latest-tags` Makefile target is wired into the `presubmit` gate so
regressions are caught before merge.

All simulator image references are currently pinned to `v0.9.0`; this check
prevents future changes from accidentally introducing `:latest`.

Fixes #1306

```release-note
NONE
```


scripts/check-latest-tags.sh
```diff
@@ -0,0 +1,68 @@
...
+# Usage:
+#   ./scripts/check-latest-tags.sh [DIR ...]
+#
+# When no directories are given the entire repository is scanned.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
```
shmuelk: This is the REPO_ROOT and not the SCRIPT_ROOT

```suggestion
REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
```

scripts/check-latest-tags.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# The simulator image name to check.
+SIMULATOR_IMAGE="llm-d-inference-sim"
+
+if [[ $# -gt 0 ]]; then
+  SCAN_DIRS=("$@")
+else
+  SCAN_DIRS=("${SCRIPT_ROOT}")
```
shmuelk:
```suggestion
  SCAN_DIRS=("${REPO_ROOT}")
```

elevran:
@apollofps is it possible to adapt so that we vavlidate all images, not just the simulator, are not using latest?
Perhaps an option to run this in two mode: (1) strict - fail if using `latest`; (2) warn-only?

apollofps:
@shmuelk @elevran Pushed updates for both:

- `SCRIPT_ROOT` renamed to `REPO_ROOT`.
- Script now checks all image references, not just the simulator. Added `--warn` flag so there are two Makefile targets:
  - `make check-latest-tags` — warns only (what `presubmit` runs)
  - `make check-latest-tags-strict` — fails on any `:latest`

`presubmit` can switch to strict once the existing GPU manifest tags get pinned.

Makefile
```diff
@@ -199,9 +199,17 @@ vulncheck: image-build-builder ## Run govulncheck for known vulnerabilities
...
+	@./scripts/check-latest-tags.sh --warn
+
+.PHONY: check-latest-tags-strict
+check-latest-tags-strict: ## Check ':latest' image tags in YAML (strict; fails on any violation)
+	@./scripts/check-latest-tags.sh
+
 .PHONY: presubmit
 presubmit: LINT_NEW_ONLY=true
-presubmit: git-branch-check signed-commits-check go-mod-check format lint vulncheck
+presubmit: git-branch-check signed-commits-check go-mod-check format lint vulncheck check-latest-tags
```
elevran, line 212: The `presubmit` target uses `check-latest-tags` (warn mode), so `:latest` regressions still pass CI. Can you please create tracking issue for flipping this to `check-latest-tags-strict` once the GPU manifests are pinned? Without one, the warn-only state may persist indefinitely.

scripts/check-latest-tags.sh
```diff
@@ -0,0 +1,97 @@
...
+  # Use grep -Hn (force filenames, no recursion) since find already
+  # supplies explicit paths. Match "image:" as a YAML key (leading
+  # whitespace) to avoid substring hits in unrelated fields.
+  matches="$(find "$dir" \
+    -path '*/.git' -prune -o \
+    -path '*/vendor' -prune -o \
+    -path '*/node_modules' -prune -o \
+    -type f \( -name '*.yaml' -o -name '*.yml' \) -print0 \
+    | xargs -0 grep -Hn ':latest' \
+    | grep '[[:space:]]image:' \
```
elevran, line 68: nit: `[[:space:]]image:` requires at least one whitespace character before `image:`, so a top-level `image:` key at column 0 would be missed. Uncommon in k8s YAML, but worth noting as a known limitation.

elevran (commented):
Solid implementation - thanks!
The --warn/strict split makes sense as a stepping stone, and the script is well-hardened (prune .git/vendor, -Hn for reliable output).
Two observations inline.

elevran:
/lgtm
@apollofps some remaining nits. Will merge this - please consider opening issue to address nits.

elevran (approved):

PR #1433, opened 2026-06-01
[tests] Adds e2e tests for the /inference/v1/generate endpoint.

revit13:
**What type of PR is this?**
/kind test

**What this PR does / why we need it**:

**Overview**
Adds end-to-end tests for the `/inference/v1/generate` endpoint. 

**Which issue(s) this PR fixes**:

Part of #1363

**Release note** _(write `NONE` if no user-facing change)_:
```release-note
Update simulator image tag to v0.9.1
```

PR #1458, opened 2026-06-02
feat(sidecar): disaggregate /inference/v1/generate (token-in) requests

### LINKED ISSUE #1461: pd-sidecar does not disaggregate /inference/v1/generate (token-in) — P/D silently runs decode-only
## What happened

The pd-sidecar registers the disaggregated-prefill handler only for the OpenAI paths:

```go
// pkg/sidecar/proxy/proxy.go createRoutes()
mux.HandleFunc("POST "+ChatCompletionsPath, s.disaggregatedPrefillHandler(APITypeChatCompletions))
mux.HandleFunc("POST "+CompletionsPath,     s.disaggregatedPrefillHandler(APITypeChatCompletions))
mux.HandleFunc("POST "+ResponsesPath,       s.disaggregatedPrefillHandler(APITypeResponses))
```

vLLM's token-in `POST /inference/v1/generate` endpoint — which the EPP gained support for in #1248 — has no route here, so it falls through to the plain decoder proxy. As a result, for token-in clients running P/D, the request **silently degrades to decode-only**:

- the prefiller never receives traffic,
- no NIXL KV transfer happens,
- the decode worker does the full prefill+decode locally.

The output is still correct, so the degradation is easy to miss — it only shows up as the prefill pool sitting idle and decode TTFT/throughput not matching a working P/D setup.

## Expected

`/inference/v1/generate` should disaggregate like the OpenAI paths: the prefiller does the prefill, and the decoder pulls KV over the configured connector (e.g. NIXL).

## Notes

The handler is almost API-agnostic; the only generate-specific detail is that the token-limit fields (`max_tokens`, `min_tokens`) live nested under `sampling_params` rather than at the top level, so the prefill token-cap has to be written there.

PR #1458 implements this.

S1ro1:
Fixes #1461

## Problem

The pd-sidecar registers the disaggregated-prefill handler only for the OpenAI paths:

```go
mux.HandleFunc("POST "+ChatCompletionsPath, s.disaggregatedPrefillHandler(APITypeChatCompletions))
mux.HandleFunc("POST "+CompletionsPath,     s.disaggregatedPrefillHandler(APITypeChatCompletions))
mux.HandleFunc("POST "+ResponsesPath,       s.disaggregatedPrefillHandler(APITypeResponses))
```

vLLM's token-in `POST /inference/v1/generate` endpoint — which the EPP gained support for in #1248 — has no route, so it falls through to the plain decoder proxy. For token-in clients (e.g. tokens-in-tokens-out RL rollouts), P/D then **silently degrades to decode-only**: the prefiller never receives traffic, no NIXL KV transfer happens, and the decode worker does the full prefill+decode locally. The output is correct, so it is easy to miss.

## Change

Register `POST /inference/v1/generate` on `disaggregatedPrefillHandler` under a new `APITypeGenerate`, mirroring the existing paths (enum + `String()` + `tokenLimitFieldsForAPIType` + route).

The one way it differs from the OpenAI APIs: a generate request carries its token limits (`max_tokens`, `min_tokens`) **nested under `sampling_params`** rather than at the top level. The NIXLv2 connector caps the prefill stage to a single output token, so it has to write into the right place. A small `tokenLimitMap(req, apiType)` helper returns the map that holds those fields — `sampling_params` for generate (created if absent), the request itself otherwise — and the existing save/override/restore logic runs on it unchanged. For the OpenAI APIs `tokenLimitMap` returns the request, so their behavior is identical to before.

`min_tokens` is capped alongside `max_tokens` because vLLM rejects a request whose `min_tokens` exceeds `max_tokens`.

## Testing

- `go test ./pkg/sidecar/proxy/...` passes (66/66 specs). New generate-API specs mirror the existing chat/responses sections and assert:
  - the prefill request caps `sampling_params.max_tokens` and `sampling_params.min_tokens` to `1`, with `do_remote_decode=true` and `stream=false`;
  - the decode request restores the caller's original `sampling_params` values (and drops them again when the caller never set them);
  - `sampling_params` is created when absent;
  - a request with no prefill header passes straight through to decode.
- Verified on a multi-node SLURM P/D deployment (NixlConnector): with the route in place, prefill workers go from **0 → receiving `/inference/v1/generate` traffic**, and decode pulls KV over NIXL instead of re-prefilling locally.

Opening as draft for maintainer feedback on the approach.


pkg/sidecar/proxy/connector_nixlv2_test.go
```diff
@@ -512,4 +512,114 @@ var _ = Describe("NIXL Connector (v2)", func() {
...
+		req, err := http.NewRequest(http.MethodPost, proxyBaseAddr+GeneratePath, strings.NewReader(body))
+		Expect(err).ToNot(HaveOccurred())
+		if withPrefillHeader {
+			req.Header.Add(routing.PrefillEndpointHeader, testInfo.prefillBackend.URL[len("http://"):])
+		}
+
+		rp, err := http.DefaultClient.Do(req)
+		Expect(err).ToNot(HaveOccurred())
+		if rp.StatusCode != 200 {
+			bp, _ := io.ReadAll(rp.Body) //nolint:all
```
roytman: can you change the code to avoid linter directives?

pkg/sidecar/proxy/connector_nixlv2_test.go
```diff
@@ -512,4 +512,114 @@ var _ = Describe("NIXL Connector (v2)", func() {
...
+		prefillSP := samplingParamsOf(testInfo.prefillHandler.CompletionRequests[0])
+		Expect(prefillSP).To(HaveKeyWithValue("max_tokens", BeNumerically("==", 1)))
+		Expect(prefillSP).To(HaveKeyWithValue("min_tokens", BeNumerically("==", 1)))
+
+		decodeSP := samplingParamsOf(testInfo.decodeHandler.CompletionRequests[0])
+		Expect(decodeSP).To(HaveKeyWithValue("max_tokens", BeNumerically("==", 100)))
+		Expect(decodeSP).To(HaveKeyWithValue("min_tokens", BeNumerically("==", 5)))
+	})
+
+	It("should cap prefill and drop the caps in decode when sampling_params omits them", func() {
```
roytman, line 600: The four new specs all send a body that includes sampling_params (with or without max_tokens/min_tokens). None covers the case where the caller omits sampling_params entirely, which is exactly the scenario where the empty-map leak bites. A spec that POSTs {"model": "...", "token_ids": [...]} (no sampling_params) and asserts decodeReq does not contain sampling_params would catch the issue and lock in the contract.

roytman (commented):
Hi, @S1ro1 , thank you for the PR.

Please fix the CI failures and Copilot comments. 

https://github.com/llm-d/llm-d-router/pull/1433 adds an e2e test to the case.

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -32,6 +32,22 @@ import (
 	"github.com/llm-d/llm-d-router/pkg/telemetry"
 )
 
+// tokenLimitMap returns the map holding the token-limit fields: sampling_params
+// for the generate API (created if absent), or the request itself otherwise.
```
shmuelk, line 36:
```suggestion
// for the generate API from the request itself, if present, or created if absent.
```

PR #1459, opened 2026-06-02
Add support for length-based scoring in prefix cache plugin

### LINKED ISSUE #1095: Approximate prefix cache scorer should incorporate the absolute match length when calculating the score
xref: https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2561

Currently the approximate prefix cache scorer calculates the score based on the match percentage of the prompt itself. The absolute prompt length should be a factor in the score.

Why is this needed:
The current approach may not be ideal for short sequences, which will typically get a high match percentage (since the denominator is small) and so results in high prefix match score and could be more susceptible to hot spots.

azamikram:
**What type of PR is this?**

/kind feature

The current prefix scorer only considers the match ratio when scoring endpoints, treating all matched prefixes equally regardless of their absolute length.

This PR introduces `matchLengthWeight` and `matchLengthScaleTokens` parameters to the prefix scorer. The updated scoring function computes a weighted sum of the match ratio and the match length ratio. The match length ratio is non-linear to prioritize KV cache reuse for longer requests relative to the `matchLengthScaleTokens`.

To maintain backward compatibility, `matchLengthWeight` defaults to `0.0`, preserving the current scoring behavior.

[Benchmark results and detailed report](https://docs.google.com/document/d/1Fzu0s76uwNrJXkom4GEZhAEzJmh9RurwXd1uHU25ICg/edit?usp=sharing)

Macro Results (RatioOnlyScorer is the baseline and CostAwareScorer is the proposed function)

Use case 1: RatioOnlyScorer never migrates to a cold pod even when the load increases. CostAwareScorer migrates the short requests to a cold pod.

Use case 2: Even when RatioOnlyScorer migrates to a cold pod, it places both short and long request on the cold pod.

Use case 3: When the load increases so much that long requests also need to be migrates to a cold pod, both RatioOnlyScorer and CostAwareScorer converge.
none

Fixes #1095

```release-note
Prefix cache scorer optionally considers absolute prefix match length. 
```

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/scorer/prefix/README.md
+++ b/pkg/epp/framework/plugins/scheduling/scorer/prefix/README.md
@@ -2,26 +2,36 @@
 
 **Type:** `prefix-cache-scorer`
 
-Scores candidate endpoints using `PrefixCacheMatchInfo` prepared earlier in the request pipeline.
+Scores candidate endpoints based on how much of the request's prefix is already cached. It can consider both the *ratio* of the prefix matched and the *absolute length* of the match. By default it only considers the ratio of prefix matched.
 
 ## What it does
 
-For each candidate endpoint, the scorer reads the `PrefixCacheMatchInfo` attribute and computes:
+The scorer computes a score between `0.0` and `1.0` for each endpoint using a weighted combination of two metrics:
+
+1.  **Match Ratio**: The proportion of the request prefix that is cached.
+2.  **Match Length**: The absolute number of cached tokens, normalized against a scale factor.
+
+### Scoring Formula
 
 ```text
-score = matchBlocks / totalBlocks
+score = matchLengthWeight*matchLengthScore + (1.0-matchLengthWeight)*matchRatioScore
 ```
 
-This produces a normalized score in the range `[0, 1]`:
+Where:
+*   **`matchRatioScore`** = `matchBlocks` / `totalBlocks`
+*   **`matchLengthScore`** = `min(1.0, matchedTokens / matchLengthScaleTokens) ^ 2`
+    *   `matchedTokens` = `matchBlocks` * `blockSize`
+    *   **`matchLengthScaleTokens`** is the scaling factor.
 
-- higher score: more of the request prefix is expected to be reusable from cache
-- lower score: less prefix cache reuse is expected
+If `matchLengthWeight` is `0.0` (default), the score simplifies to just the `matchRatioScore`.
 
 If the attribute is missing, has the wrong type, or `totalBlocks` is zero, the endpoint receives score `0`.
 
-## Inputs consumed
+### Why the Quadratic Term?
 
-This scorer consumes:
+The squaring of `matchLengthScore` introduces a non-linearity motivated by the fact that attention computation grows quadratically as a function of prompt length. This ensures that the scoring function assigns a higher priority to longer requests, where the computational and latency savings of KV-cache reuse are significantly more critical.
+
+## Inputs Consumed
 
 - `PrefixCacheMatchInfo`
 
@@ -29,10 +39,22 @@ The attribute is typically produced by the approximate prefix cache data produce
 
 ## Configuration
 
-This plugin does not define any plugin-specific parameters.
+| Parameter | Type | Description | Default |
+| :--- | :--- | :--- | :--- |
+| `matchLengthWeight` | float | Weight of the absolute match length in the score. Must be between `0.0` and `1.0`. | `0.0` |
+| `matchLengthScaleTokens` | integer | The number of tokens used to normalize `matchLengthScore`. | `8192` |
 
-## Operational notes
+### Example
 
-- The scorer itself does not hash prompts or maintain cache state.
-- It only converts previously prepared prefix match data into endpoint scores.
-- To be useful, it should be used together with a data producer that populates `PrefixCacheMatchInfo`.
+```yaml
+plugins:
+- type: prefix-cache-scorer
+  parameters:
+    matchLengthWeight: 0.5
+    matchLengthScaleTokens: 16384
+```
+
+## Operational Notes
+
+*   **Tuning `matchLengthScaleTokens`**: This should be set to reflect your workload. A good rule of thumb is to set it to the **P95 prompt length** of your typical requests. Setting it too high (e.g., to the model's maximum limit when actual requests are short) will compress the score range and make the scorer less effective.
+*   The scorer is stateless; it does not manage cache state or hash prompts itself. It relies entirely on the data producer.
--- a/pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
+++ b/pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
@@ -33,12 +34,19 @@ import (
 type Config struct {
 	// The name of the data producer that produces PrefixCacheMatchInfo.
 	PrefixMatchInfoProducerName string `json:"prefixMatchInfoProducerName,omitempty"`
+	// The weight assigned to match length, between 0.0 and 1.0.
+	MatchLengthWeight float64 `json:"matchLengthWeight,omitempty"`
+	// Normalization factor for match length in terms of tokens.
+	// Used only when MatchLengthWeight > 0.
+	MatchLengthScaleTokens int `json:"matchLengthScaleTokens,omitempty"`
 }
 
 // Plugin implements the prefix cache aware scoring logic.
 type Plugin struct {
-	typedName          plugin.TypedName
-	prefixMatchDataKey plugin.DataKey
+	typedName              plugin.TypedName
+	prefixMatchDataKey     plugin.DataKey
+	matchLengthWeight      float64
+	matchLengthScaleTokens int
 }
 
 // compile-time type assertions
@@ -49,11 +57,19 @@ var (
 const (
 	// Type is the unique identifier for the prefix cache scorer plugin.
 	PrefixCacheScorerPluginType = "prefix-cache-scorer"
+	// The default weight of the absolute match length in the score.
+	// Set to 0 so by default only the match ratio is considered.
+	defaultMatchLengthWeight = 0.0
+	// Default number of tokens used as a scaling factor.
+	defaultMatchLengthScaleTokens = 8192
 )
 
 // PrefixCachePluginFactory defines the factory function for the Prefix plugin.
 func PrefixCachePluginFactory(name string, decoder *json.Decoder, handle plugin.Handle) (plugin.Plugin, error) {
-	var cfg Config
+	cfg := Config{
+		MatchLengthWeight:      defaultMatchLengthWeight,
+		MatchLengthScaleTokens: defaultMatchLengthScaleTokens,
+	}
 	if decoder != nil {
 		if err := decoder.Decode(&cfg); err != nil {
 			return nil, fmt.Errorf("failed to decode prefix cache scorer parameters: %w", err)
@@ -64,6 +80,17 @@ func PrefixCachePluginFactory(name string, decoder *json.Decoder, handle plugin.
 	if err != nil {
 		return nil, err
 	}
+
+	if cfg.MatchLengthWeight < 0.0 || cfg.MatchLengthWeight > 1.0 {
+		return nil, fmt.Errorf("matchLengthWeight must be between 0.0 and 1.0, got %f", cfg.MatchLengthWeight)
+	}
+	p.matchLengthWeight = cfg.MatchLengthWeight
+
+	if p.matchLengthWeight > 0.0 && cfg.MatchLengthScaleTokens <= 0 {
+		return nil, fmt.Errorf("matchLengthScaleTokens must be greater than 0 when matchLengthWeight is greater than 0, got %d", cfg.MatchLengthScaleTokens)
+	}
+	p.matchLengthScaleTokens = cfg.MatchLengthScaleTokens
+
 	return p, nil
 }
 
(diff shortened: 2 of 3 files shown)


pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
```diff
@@ -115,8 +138,18 @@ func (p *Plugin) Score(ctx context.Context, _ *fwksched.InferenceRequest, endpoi
...
-				scores[endpoint] = float64(prefixMatchInfo.MatchBlocks()) / float64(prefixMatchInfo.TotalBlocks())
+			totalBlocks := prefixMatchInfo.TotalBlocks()
+			if totalBlocks != 0 {
+				matchRatio := float64(prefixMatchInfo.MatchBlocks()) / float64(totalBlocks)
+				matchLengthRatio := 0.0
+				blockSize := prefixMatchInfo.BlockSizeTokens()
+				if blockSize > 0 && p.maxModelLen > 0 {
+					// (TotalBlocks * BlockSize / maxModelLen) ^ 2
+					// Capped at 1.0 as the normalized score term cannot be greater than 1.
+					matchLengthRatio = float64(totalBlocks) * float64(blockSize) / float64(p.maxModelLen)
```
zetxqx: for matchLengthRatio, should we use totalBlocks or matchedBlocks?

azamikram: Thanks for pointing it out; I fixed it. It was supposed to be `matchedBlocks`.

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
```diff
@@ -32,13 +34,17 @@ import (
 // Config defines the configuration for the prefix cache scorer plugin.
 type Config struct {
 	// The name of the data producer that produces PrefixCacheMatchInfo.
-	PrefixMatchInfoProducerName string `json:"prefixMatchInfoProducerName,omitempty"`
+	PrefixMatchInfoProducerName string   `json:"prefixMatchInfoProducerName,omitempty"`
+	PrefixLengthWeight          *float64 `json:"prefixLengthWeight,omitempty"`
```
liu-cong: can we use float and int instead of the pointers?

liu-cong (commented):
can you also add your benchmark results?

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
shmuelk: Pointers are useful for differentiating between fields not in the JSON and fields in the JSON with the default value (0 or 0.0 in this case).

pkg/epp/framework/plugins/scheduling/scorer/prefix/README.md
```diff
@@ -6,16 +6,19 @@ Scores candidate endpoints using `PrefixCacheMatchInfo` prepared earlier in the
...
 ## What it does
 
-For each candidate endpoint, the scorer reads the `PrefixCacheMatchInfo` attribute and computes:
+For each candidate endpoint, the scorer reads the `PrefixCacheMatchInfo` attribute and computes a score. By default, this is the ratio of matched blocks to total blocks. Optionally, it can also factor in the absolute length of the prefix.
+
+The score is computed as:
 
 ```text
-score = matchBlocks / totalBlocks
+score = prefixLengthWeight * matchLengthRatio + (1.0 - prefixLengthWeight) * matchRatio
```
shmuelk: I would separate out the two cases to make this clearer

pkg/epp/framework/plugins/scheduling/scorer/prefix/README.md
```diff
@@ -6,16 +6,19 @@ Scores candidate endpoints using `PrefixCacheMatchInfo` prepared earlier in the
...
 ```
 
-This produces a normalized score in the range `[0, 1]`:
+Where:
+- `matchRatio = matchBlocks / totalBlocks` (the fraction of the request prefix matched)
+- `matchLengthRatio = min(1.0, matchedBlocks * blockSize / maxModelLen) ^ 2` (square of normalized matched blocks relative to `maxModelLen`)
 
-- higher score: more of the request prefix is expected to be reusable from cache
-- lower score: less prefix cache reuse is expected
+If `prefixLengthWeight` is `0.0` (the default), the score simplifies to just the `matchRatio`.
```
shmuelk: This comment should be included in the initial description of how the plugin computes the score.

pkg/epp/framework/plugins/scheduling/scorer/prefix/README.md
```diff
@@ -29,7 +32,23 @@ The attribute is typically produced by the approximate prefix cache data produce
 
 ## Configuration
 
-This plugin does not define any plugin-specific parameters.
+This plugin supports the following configuration parameters:
```
shmuelk:
```suggestion
This plugin supports the following optional configuration parameters:
```

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
ahg-g: pls clearly document the parameters

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
```diff
@@ -32,13 +34,17 @@ import (
...
+	MaxModelLen                 *int     `json:"maxModelLen,omitempty"`
```
ahg-g: MaxModelLen is confusing (I know vllm uses this name, but it is not accurate), probably `MaxContextLength`? also, what is the unit? please prefix it with the unit, if tokens, it would be `MaxContextLengthTokens`

pkg/epp/framework/plugins/scheduling/scorer/prefix/README.md
```diff
@@ -6,16 +6,19 @@ Scores candidate endpoints using `PrefixCacheMatchInfo` prepared earlier in the
...
 
 ```text
-score = matchBlocks / totalBlocks
+score = prefixLengthWeight * matchLengthRatio + (1.0 - prefixLengthWeight) * matchRatio
 ```
 
-This produces a normalized score in the range `[0, 1]`:
+Where:
+- `matchRatio = matchBlocks / totalBlocks` (the fraction of the request prefix matched)
+- `matchLengthRatio = min(1.0, matchedBlocks * blockSize / maxModelLen) ^ 2` (square of normalized matched blocks relative to `maxModelLen`)
```
ahg-g: explain why this is applying a square

pkg/epp/framework/plugins/scheduling/scorer/prefix/README.md
```diff
@@ -29,7 +32,23 @@ The attribute is typically produced by the approximate prefix cache data produce
 
 ## Configuration
 
-This plugin does not define any plugin-specific parameters.
+This plugin supports the following configuration parameters:
+
+| Parameter | Type | Description | Default |
+| :--- | :--- | :--- | :--- |
+| `prefixLengthWeight` | float | Weight of the absolute prefix length in the score. Must be between `0.0` and `1.0`. | `0.0` |
+| `maxModelLen` | integer | The maximum context length (in tokens) supported by the model. Required if `prefixLengthWeight` > `0.0`. | `8192` |
```
ahg-g: where is this default being applied?

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
```diff
@@ -115,8 +138,19 @@ func (p *Plugin) Score(ctx context.Context, _ *fwksched.InferenceRequest, endpoi
 		}
 
 		if prefixMatchInfo, ok := info.(*attrprefix.PrefixCacheMatchInfo); ok {
```
ahg-g: to reduce indentation, pls do the following:

```
if prefixMatchInfo, ok := info.(*attrprefix.PrefixCacheMatchInfo); !ok {
  logger.V(logutil.DEFAULT).Error(nil, "PrefixCacheMatchInfo has unexpected type, assigning score 0", "endpoint", endpoint)
  return scores
}

// reset of the scores calculations
return scores

```

shmuelk:
@azamikram Please rebase with main to solve the CI issues

ahg-g (commented):
I think another approach is to set a minimum for the denominator (8k tokens for example, but we need to do some testing), this will make it simpler since it is just one parameter, not two.

hexfusion:
@azamikram we would like to get this into the release do you have cycles to address feedback?

liu-cong:
> @azamikram we would like to get this into the release do you have cycles to address feedback?

@hexfusion We need more testing on this. And there are other ideas as well like @ahg-g mentioned  `I think another approach is to set a minimum for the denominator `. I talked to @azamikram and he is trying a few different benchmark scenarios to validate. 

We may need to postpone this to the next release.

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
azamikram: I decided to go with float and int to keep things simple. Let me know what you think.

azamikram:
> can you also add your benchmark results?

Added the link to the detailed report.

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
azamikram: I agree that MaxModelLen is a bit confusing and its purpose isn't set to a limit on the maximum number of tokens. The goal is to configure the point at which the requests become most important for KV cache usage. I ended up renaming it to `prefillSaturationTokens`, let me know what you think.

azamikram:
> I think another approach is to set a minimum for the denominator (8k tokens for example, but we need to do some testing), this will make it simpler since it is just one parameter, not two.

I am now proposing to use 8192 as the default value. I also wrote in the report about how to configure this value for a specific workload. The general recommendation is to set it to P95 of the input length.

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
```diff
@@ -33,12 +34,18 @@ import (
 type Config struct {
 	// The name of the data producer that produces PrefixCacheMatchInfo.
 	PrefixMatchInfoProducerName string `json:"prefixMatchInfoProducerName,omitempty"`
+	// The weight of the absolute prefix length in the score, between 0.0 and 1.0.
+	PrefixLengthWeight float64 `json:"prefixLengthWeight,omitempty"`
+	// The number of tokens at which prefill performance saturates.
+	PrefillSaturationTokens int `json:"prefillSaturationTokens,omitempty"`
```
zetxqx: just one nit about naming, how about `PrefixSaturationTokens` ? comment could be

```
// PrefixSaturationTokens is the matched-prefix length, in tokens, at which the
  // length-based score term saturates at its maximum (1.0).
```

azamikram: Based on our discussion earlier, I updated the names to be `MatchLengthWeight` and `MatchLengthScaleTokens`.

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
```diff
@@ -33,12 +34,18 @@ import (
 type Config struct {
 	// The name of the data producer that produces PrefixCacheMatchInfo.
 	PrefixMatchInfoProducerName string `json:"prefixMatchInfoProducerName,omitempty"`
+	// The weight of the absolute prefix length in the score, between 0.0 and 1.0.
+	MatchLengthWeight float64 `json:"matchLengthWeight,omitempty"`
+	// The number of tokens at which prefill performance saturates.
```
zetxqx: can you update this comment accordingly?

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
```diff
@@ -49,11 +56,18 @@ var (
 const (
 	// Type is the unique identifier for the prefix cache scorer plugin.
 	PrefixCacheScorerPluginType = "prefix-cache-scorer"
+	// The default weight of the absolute match length in the score.
```
zetxqx, line 60: can you also add a sentence something like "the default is zero so the match length is not considered by default"

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
```diff
@@ -64,6 +78,17 @@ func PrefixCachePluginFactory(name string, decoder *json.Decoder, handle plugin.
 	if err != nil {
 		return nil, err
 	}
+
+	if cfg.MatchLengthWeight < 0.0 || cfg.MatchLengthWeight > 1.0 {
+		return nil, fmt.Errorf("matchLengthWeight must be between 0.0 and 1.0, got %f", cfg.MatchLengthWeight)
+	}
+	p.matchLengthWeight = cfg.MatchLengthWeight
+
+	if p.matchLengthWeight > 0.0 && cfg.MatchLengthScaleTokens <= 0 {
```
zetxqx, line 89: nit: no need to check `p.matchLengthWeight > 0.0` again since it is early returned on LINE#83

pkg/epp/framework/plugins/scheduling/scorer/prefix/README.md
```diff
@@ -2,37 +2,61 @@
 
 **Type:** `prefix-cache-scorer`
 
-Scores candidate endpoints using `PrefixCacheMatchInfo` prepared earlier in the request pipeline.
+Scores candidate endpoints based on how much of the request's prefix is already cached. It can consider both the *ratio* of the prefix matched and the *absolute length* of the match.
```
zetxqx: Can you add `by default it only considers the ratio of prefix matched`?

zetxqx (commented):
a couple of nits and comments refinement

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
azamikram: We use `MatchLengthScaleTokens` only when `p.matchLengthWeight > 0`. If I were to remove it, then it will break if someone uses invalid `MatchLengthScaleTokens` even without specifying `MatchLengthWeight`. In that case, we will be complaining about `MatchLengthScaleTokens` even when it is not going to be used.

But let me know what you think and I can update it if we need to.

zetxqx: got it, I missed the matchLengthWeight==0 case, this looks good

zetxqx (dismissed):

pkg/epp/framework/plugins/scheduling/scorer/prefix/plugin.go
```diff
@@ -114,13 +141,31 @@ func (p *Plugin) Score(ctx context.Context, _ *fwksched.InferenceRequest, endpoi
...
+		blockSize := prefixMatchInfo.BlockSizeTokens()
+		if blockSize > 0 && p.matchLengthScaleTokens > 0 {
+			// (matchBlocks * blockSize / matchLengthScaleTokens) ^ 2
+			// Capped at 1.0 as the normalized score term cannot be greater than 1.
+			ratio := float64(matchBlocks) * float64(blockSize) / float64(p.matchLengthScaleTokens)
+			matchLengthRatio = math.Min(1.0, ratio)
+			matchLengthRatio *= matchLengthRatio
 		}
+		scores[endpoint] += p.matchLengthWeight * matchLengthRatio
+		scores[endpoint] += (1.0 - p.matchLengthWeight) * matchRatio
```
liu-cong:
```suggestion
		matchRatioScore := float64(matchBlocks) / float64(totalBlocks)
		blockSize := prefixMatchInfo.BlockSizeTokens()
		matchLengthScore := 0.0
		// Calculate matchLengthScore when match length is considered
		if p.matchLengthWeight > 0.0 {
		   normalizedMatchLength := math.Min(1.0, float64(matchBlocks) * float64(blockSize) / float64(p.matchLengthScaleTokens))
		   matchLengthScore = normalizedMatchLength^2
		} 
		scores[endpoint] += p.matchLengthWeight * matchLengthScore + (1.0 - p.matchLengthWeight) * matchRatioScore
```

liu-cong: matchLengthRatio is confusing when we also have matchRatio, suggest renaming them to matchLengthScore and matchRatioScore.

pkg/epp/framework/plugins/scheduling/scorer/prefix/README.md
```diff
@@ -2,37 +2,61 @@
...
+
+1.  **Match Ratio**: The proportion of the request prefix that is cached.
+2.  **Match Length**: The absolute number of cached tokens, normalized against a scale factor.
+
+### Scoring Formula
 
 ```text
-score = matchBlocks / totalBlocks
+score = 0
+score += matchLengthWeight * matchLengthRatio
```
liu-cong: pls also update the doc according with the renamed matchLengthScore and matchRatioScore.

Let's also have the formula in one line, `score +=` add unnecessary cognitive overhead IMO.

azamikram: I updated the doc and the PR description as well. Also add main results in the PR from the doc.

liu-cong (approved):

PR #1462, opened 2026-06-02
feat(metrics): extract configured custom metrics

### LINKED ISSUE #1454: Support configured custom metrics in the core metrics extractor

Add support for configured custom gauge metrics in the core metrics extractor.

The extractor should be able to read user-configured gauge metric specs from model-server metrics and store the extracted numeric values as endpoint attributes under configured internal keys.

This is intended as a scoped first step toward #977.

Initial scope:

- [x] Support gauge and counter metrics.
- [x] Allow users to configure the source metric spec.
- [x] Allow users to configure the internal endpoint attribute key.
- [x] Store extracted numeric values as endpoint attributes.
- [x] Do NOT add new fixed fields to `Metrics`.

Out of scope:

- counters and histograms
- PromQL expressions
- metric scoring logic
- replacing existing well-known metrics
- broader plugin-driven metric registration

**Why is this needed**:

#977 proposes a generic metric scorer so users can route based on arbitrary model-server metrics. Before a generic scorer can consume such metrics cleanly, the data layer needs a small foundation for extracting configured custom metrics without adding a new hardcoded `Metrics` field for every signal.

Storing custom metric values as endpoint attributes keeps the existing well-known metrics unchanged while giving future scorers a stable internal key to consume.

richl9:
**What type of PR is this?**

/kind feature

This PR adds support for configured custom gauge metrics in the core metrics extractor.

Engine configs can now map a Prometheus gauge metric selector to an endpoint attribute key. The extractor reads the matching gauge value and stores it as a cloneable endpoint attribute, without adding new fixed fields to the core `Metrics` struct.

This provides the extraction/storage layer for custom numeric routing signals. Consuming those attributes in scorers or filters is left for follow-up work.

Fixes #1454

**Release note**:

```release-note
Support configured custom gauge metrics in the core metrics extractor.

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/attribute/metrics/data_types.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/metrics/data_types.go
@@ -1,0 +1,30 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+
+// ScalarMetricValue is a numeric endpoint attribute extracted from a configured scalar metric.
+type ScalarMetricValue float64
+
+func (v ScalarMetricValue) Clone() fwkdl.Cloneable {
+	return v
+}
+
+func ReadScalarMetricValue(attrs fwkdl.AttributeMap, key string) (ScalarMetricValue, bool) {
+	return fwkdl.ReadAttribute[ScalarMetricValue](attrs, key)
+}
--- a/pkg/epp/framework/plugins/datalayer/extractor/metrics/mapping.go
+++ b/pkg/epp/framework/plugins/datalayer/extractor/metrics/mapping.go
@@ -42,9 +42,10 @@ type Mapping struct {
 	// (e.g. Triton TRT-LLM).
 	CacheBlockSize *Spec
 	CacheNumBlocks *Spec
+	CustomMetrics  []CustomMetric
 }
 
-// MappingConfig holds the string-based configuration used to build a Mapping.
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by richl9** [comment]
Thanks @shmuelk. I fixed the small lint issue from the failed check and amended the commit.

File: pkg/epp/framework/plugins/datalayer/attribute/metrics/data_types.go [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+
+// GaugeValue is a numeric endpoint attribute extracted from a configured gauge metric.
+type GaugeValue float64
```
ahg-g: Should we call this `ScalarMetricValue` to represent either a Gauge or a Counter?

pkg/epp/framework/plugins/datalayer/extractor/metrics/spec.go
```diff
@@ -161,3 +161,14 @@ func extractValue(metric *dto.Metric) float64 {
 	}
 	return 0
 }
+
+func extractGaugeValue(metric *dto.Metric) (float64, bool) {
```
ahg-g: do we need this? we have extractValue above which, which actually works for both gauge or counter

pkg/epp/framework/plugins/datalayer/extractor/metrics/mapping.go
```diff
@@ -55,6 +56,17 @@ type MappingConfig struct {
 	CacheNumBlocksLabel string
 	CacheBlockSize      string
 	CacheNumBlocks      string
+	CustomMetrics       []CustomMetricConfig
+}
+
+type CustomMetric struct {
```
ahg-g, line 62: The two types, CustomMetric and CustomMetricConfig are practically identical, it should be sufficient to define one only

ahg-g (commented):
I suggest to generalize this for both gauge and counter, it should be straightforward.

richl9:
@ahg-g thanks for the comments. They make sense to me. I have addressed them and pushed again.

shmuelk:
@richl9 Thanks for the updates. Much more generic now.

/lgtm

shmuelk:
@richl9 Please update the PR title to `feat(metrics): extract configured custom metrics` as this PR is now more generic.

richl9:
> @richl9 Please update the PR title to `feat(metrics): extract configured custom metrics` as this PR is now more generic.

Thanks @shmuelk . Done.

ahg-g:
/approve

@liu-cong fyi

PR #1463, opened 2026-06-02
moriio: WRITE-mode + Wide-EP sidecar support (1P1D DP=EP=8 TP=1)

shikamd123:
**What type of PR is this?**

  feature

Adds MoRI-IO WRITE-mode and intra-node Wide-EP (DP>1) support to the
llm-d-router disaggregation sidecar's NIXLv2 connector path.  The
sidecar was previously limited to:

* MoRI-IO READ-mode wire shape only (decode pulls KV via RDMA Read).
* Single-DP deployments — `remote_dp_size` was effectively pinned to 1
  on the wire so vLLM's `MoRIIOConnector` only ever handshook with one
  remote DP rank.
* Strictly serial prefill→decode dispatch — TTFT included the full
  prefill HTTP round-trip before decode could even begin block
  allocation.

These limitations block disaggregated DeepSeek-V3 deployments on AMD
MI300X with Wide-EP (TP=1, DP=8/16) and prevent the WRITE-mode path
from being used at all.  This PR addresses each independently:

1. **WRITE-mode wire shape on the prefill leg.** Populate
   `kv_transfer_params` with `remote_host` (decode pod IP from
   `POD_IP`), `remote_handshake_port`, `remote_notify_port`,
   `remote_dp_rank` and a stable `transfer_id` so the prefill engine
   has the destination it needs to RDMA-Write KV blocks to decode.
2. **Decode-side `kv_transfer_params` backfill.** vLLM's
   `MoRIIOConnector.request_finished` does not echo back
   `transfer_id`, `remote_notify_port` or `remote_handshake_port`; the
   sidecar fills them from its static config to prevent
   `KeyError` on `send_notify_block`.
3. **Concurrent prefill/decode dispatch (`--moriio-parallel-dispatch`).**
   New `runNIXLProtocolV2WriteParallel`: build prefill+decode bodies
   up-front from the new `--moriio-prefill-*` flags and the
   EPP-supplied prefill host:port, then fire both legs in parallel
   goroutines.  Decode's `update_state_after_alloc` /
   `send_notify_block` / block allocation overlaps prefill's forward
   pass.  Strict superset of the serial path; opt-in.
4. **Configurable `--moriio-dp-size` for intra-node Wide-EP.** Emit
   `remote_dp_size` on both legs so vLLM iterates its per-DP-rank
   handshake loop over the configured DP world size instead of a
   hardcoded 1.
5. **Stable per-request DP-rank pinning.** Route both legs of a disagg
   pair to rank H = `blake2s(uuid) % dpSize` via the
   `X-Data-Parallel-Rank` HTTP header plus
   `remote_dp_rank` / `remote_dp_rank_override` `kv_transfer_params`
   fields.  Neutralises the `SO_REUSEPORT` shuffle that vLLM's
   `--api-server-count > 1` induces, which would otherwise split the
   two legs across different DP ranks → KV handshake addresses the
   wrong peer → request hangs until `VLLM_MORIIO_DEFERRED_TIMEOUT_S`
   (300s) trips.

**All additions are opt-in behind new `--moriio-*` flags.** A user who
sets none of them gets bit-identical wire shape and dispatch order to
the existing NIXLv2 path; non-MoRI-IO users (e.g. NVIDIA) and other
connectors (`sglang`, `shared_storage`, `epd`) are entirely
untouched — those connector files have zero diff in this PR.

### Files changed (9 total: 7 modified + 2 new)

| File | Δ | Purpose |
|---|---:|---|
| `go.mod` | +1 | Add `golang.org/x/crypto v0.48.0` (for blake2s) |
| `go.sum` | +2 | Hash entries for the new dep |
| `pkg/sidecar/proxy/connector_nixlv2.go` | +434 | WRITE-mode wire shape, kv_transfer_params backfill, parallel dispatch, DP-rank pinning |
| `pkg/sidecar/proxy/connector_nixlv2_test.go` | +124 | Tests for above |
| `pkg/sidecar/proxy/dp_rank.go` (new) | +92 | `pickDPRank` blake2s helper |
| `pkg/sidecar/proxy/dp_rank_test.go` (new) | +148 | Determinism + distribution tests |
| `pkg/sidecar/proxy/options.go` | +112 | `--moriio-*` CLI flag definitions + defaults |
| `pkg/sidecar/proxy/proxy.go` | +137 | `Config` MoRIIO* fields, `X-Data-Parallel-Rank` header, new `kv_transfer_params` field constants |
| `test/sidecar/mock/chat_completions_handler.go` | +48 | `MoRIIOWriteMode` test gate (loosens NIXLv2 prefill validation in WRITE-mode test mode) |

### New CLI flags (all default to OFF / pre-existing behaviour)

```
--moriio-write-mode              --moriio-parallel-dispatch
--moriio-dp-size                 --moriio-tp-size
--moriio-local-pod-ip            --moriio-decode-handshake-port
--moriio-decode-notify-port      --moriio-decode-dp-rank
--moriio-prefill-handshake-port  --moriio-prefill-notify-port
--moriio-prefill-dp-rank
```

### Companion vLLM-side PR

The `remote_dp_rank_override` sentinel handshakes with a matching
vLLM-side fix (in
[shikamd123/vllm:llmd_vllm_mori_dp8_1p1d](https://github.com/shikamd123/vllm/tree/llmd_vllm_mori_dp8_1p1d))
that, when present, uses the sidecar-supplied `remote_dp_rank`
verbatim instead of recomputing its own blake2s hash.  This avoids
cross-language hash divergence (Go `blake2s-256` vs Python
`blake2s-8`).  Each side stays independently safe: the vLLM-side
fallback engages if any caller forgets the sentinel.

### Test coverage

```
go test ./pkg/sidecar/proxy/...
```

* `pkg/sidecar/proxy/dp_rank_test.go` — 148 lines covering
  `pickDPRank` determinism, distribution, and edge cases (empty uuid,
  dpSize=0, dpSize=1, dpSize=8, multi-thousand-iteration distribution
  smoke).
* `pkg/sidecar/proxy/connector_nixlv2_test.go` — 124 added lines
  covering `--moriio-write-mode` gate, decode-side
  `kv_transfer_params` backfill, override sentinel emission,
  parallel-dispatch sequencing.
* `make test-unit-sidecar` runs the same suite under the repo's
  builder container.

### End-to-end production validation

| | |
|---|---|
| Cluster | 2× MI300X-8 nodes |
| Model | DeepSeek-V3 |
| Topology | 1P1D, single-pod intra-node Wide-EP, EP=8 / TP=1 / DP=8 |
| Sidecar image | built from this branch |
| vLLM image | built with the companion vLLM PR's commits applied |
| Workload | `vllm bench serve` C=1,2,4,8,16,32, 312 prompts |

This PR is part of (but does not by itself fully close) two existing
umbrella issues in `llm-d/llm-d`.  Cross-repo `Fixes` does not
auto-close, and neither issue is fully resolved by router-side changes
alone — both also depend on the companion vLLM PR and on
`llm-d/llm-d` Dockerfile / guide updates that are out of scope here.

* Part of llm-d/llm-d#1469 — *[P/D][AMD]: Support AMD MoRI-IO*.
  This PR delivers the disaggregation-sidecar half: WRITE-mode wire
  shape on the NIXLv2 connector path, decode-side `kv_transfer_params`
  backfill, and optional concurrent prefill/decode dispatch.

* Part of llm-d/llm-d#1470 — *[WideEP]: AMD MI355X MoRI EP Guide for
  Pollara/Thor-2 NIC*.  This PR adds the router-side enabler for
  intra-node Wide-EP: `--moriio-dp-size` and stable per-request DP-rank
  pinning (`blake2s(uuid) % dp_size` via the `X-Data-Parallel-Rank`
  header + `remote_dp_rank{,_override}` `kv_transfer_params` fields),
  which neutralises the `SO_REUSEPORT` shuffle that vLLM's
  `--api-server-count > 1` induces.  The full well-lit path
  (`Dockerfile.rocm`, MI355X guide) remains tracked in the parent
  issue.

**Release note**:

```release-note
Add MoRI-IO WRITE-mode and intra-node Wide-EP (DP>1) support to the
disaggregation sidecar.  New opt-in CLI flags: --moriio-write-mode,
--moriio-parallel-dispatch, --moriio-dp-size, --moriio-tp-size,
--moriio-local-pod-ip, --moriio-{decode,prefill}-{handshake,notify}-port,
--moriio-{decode,prefill}-dp-rank.  Default behaviour is unchanged
when no --moriio-* flags are set.
```


shikamd123:
Note: the work for dp=ep=16 is under test for internode wide-ep and will update in the same PR soon. cc @robertgshaw2-redhat

ahg-g:
Is this PR still relevant?

shikamd123:
we can close the 1p1d PR, since relevant changes already got merged with 2p2d PR a month back.

PR #1465, opened 2026-06-03
fix vulnerabilities on main

rahulgurnani:
**What type of PR is this?**

Presently, there are active vulnerabilities on main that are causing CI to break

Fixes #

```release-note
NONE
```

### DIFF
--- a/Dockerfile.builder
+++ b/Dockerfile.builder
@@ -1,5 +1,11 @@
 FROM quay.io/projectquay/golang:1.25
 
+# Upgrade Go to 1.25.11 to fix standard library vulnerabilities (CVEs)
+RUN ARCH=$(uname -m) && \
+    GOARCH=$(echo ${ARCH} | sed 's/x86_64/amd64/; s/aarch64/arm64/') && \
+    rm -rf /usr/local/go && \
+    curl -sSfL "https://go.dev/dl/go1.25.11.linux-${GOARCH}.tar.gz" | tar -C /usr/local -xzf -
+
 RUN mkdir /app
 WORKDIR /app
 
--- a/Dockerfile.epp
+++ b/Dockerfile.epp
@@ -6,7 +6,7 @@
 ARG BASE_IMAGE=gcr.io/distroless/static:nonroot
 
 # Go build stage
-FROM --platform=${BUILDPLATFORM} quay.io/projectquay/golang:1.25 AS go-builder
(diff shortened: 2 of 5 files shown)


Dockerfile.builder
```diff
@@ -1,5 +1,11 @@
 FROM quay.io/projectquay/golang:1.25
 
+# Upgrade Go to 1.25.11 to fix standard library vulnerabilities (CVEs)
+RUN ARCH=$(uname -m) && \
```
elevran, line 4: `Dockerfile.epp` and `Dockerfile.sidecar` handle this by switching to `FROM golang:1.25.11` directly. The curl + tar approach here is fragile: the arch detection via `sed` only maps `x86_64` and `aarch64`; any other host arch silently produces an incorrect `GOARCH` and downloads the wrong tarball. If `quay.io/projectquay/golang` has a `1.25.11` tag, prefer `FROM quay.io/projectquay/golang:1.25.11`; if not, switch to `FROM golang:1.25.11` like the other Dockerfiles.

elevran (commented):
Two notes not tied to a specific line: the PR description ("What this PR does") is blank, and `Fixes #` has no issue number -- for a security fix, please state which CVEs are addressed. Inline question on `Dockerfile.builder` below.

elevran:
```console 
$ docker run -it quay.io/projectquay/golang:1.25 /bin/sh
Unable to find image 'quay.io/projectquay/golang:1.25' locally
1.25: Pulling from projectquay/golang
0840e693392a: Already exists
a55d8dfc7737: Pull complete
Digest: sha256:e29cfbfdae8bb934bad258126490bfa9b10ba4c73471e734cb92d78f148fcb65
Status: Downloaded newer image for quay.io/projectquay/golang:1.25
sh-4.4# go version
go version go1.25.11 linux/amd64
sh-4.4#
```
so this has been updated online to a valid image. We need to make sure we stop caching / reusing the old image.

elevran:
Suggestion:
I think the part of the problem is the quay.io images are all marked as 1.25 - no patch in tag, so it is not automatically updated by dependabot.
The quay.io page for the image says it is tracking the docker hub version, which does have patch versions included.

Can we use the official golang instead and validate the dependabot config handles it?
If this is a more complex change, suggest using golang across all images for now (instead of project quay images) and worrying about the dependabot later.

elevran (approved):
merging with the note to unblock CICD. The builder image carries the vulnerability but it is not a production image that is run so prefer unblocking the CICD first and fixing later.

PR #1467, opened 2026-06-03
fix(epp/metrics): treat absent vllm:lora_requests_info as no LoRA loaded

### LINKED ISSUE #926: EPP report extract failed error with "vllm:lora_requests_info"
**What happened**:
When I followed https://gateway-api-inference-extension.sigs.k8s.io/guides/ using release v1.5.0, I noticed that epp will report error as follows
```
{"level":"error","ts":1777491250.9224765,"caller":"datalayer/collector.go:211","msg":"extract failed","endpoint":"10.72.7.74","extractor":"core-metrics-extractor/core-metrics-extractor","error":"metric family \"vllm:lora_requests_info\" not found","stacktrace":"sigs.k8s.io/gateway-api-inference-extension/pkg/epp/datalayer.logErrorTransition\n\t/src/pkg/epp/datalayer/collector.go:211\nsigs.k8s.io/gateway-api-inference-extension/pkg/epp/datalayer.(*Collector).startCollection.func1.1\n\t/src/pkg/epp/datalayer/collector.go:157"}
```

EPP should be able to extract related vllm lora request info 

Follow https://gateway-api-inference-extension.sigs.k8s.io/guides/ with latest release v1.5.0

**Environment**:
- Install tools:
- Others:

gkneighb:
## Summary

Fixes #926. vLLM only emits `vllm:lora_requests_info` once at least one LoRA adapter has been loaded. A vanilla deployment with no LoRA legitimately serves no such metric family, but `LoRASpec.getLatestMetric` was treating the absent family as an extraction error. Every scrape of any non-LoRA endpoint logged `"extract failed ... metric family vllm:lora_requests_info not found"` and incremented `DataLayerExtractErrorsTotal` — masking real extraction failures.

`LoRASpec.getLatestMetric` now returns `(nil, nil)` when the family is absent or has no metrics, matching the existing semantics for "family present but no matching labels". The upstream extractor already short-circuits on `metric == nil` via `else if metric != nil`, so no caller change is needed. Standard `Spec` (used for required metrics) keeps its strict not-found behavior.

## Verified locally
- Updated `TestGetLoRAMetric/missing_family_treated_as_no_adapters` expects `(nil, nil)`.
- New `TestMetricsExtractionLoRAFamilyAbsentNoError` exercises the end-to-end Extract path with no LoRA family served and asserts no error + required metrics still populated.
- `make test-unit-epp` and `make presubmit` green locally.

## Test plan
- [x] `go test ./pkg/epp/framework/plugins/datalayer/extractor/metrics/...`
- [x] `make test-unit-epp`

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/extractor/metrics/extractor.go
+++ b/pkg/epp/framework/plugins/datalayer/extractor/metrics/extractor.go
@@ -129,10 +129,7 @@ func (ext *Extractor) Extract(ctx context.Context, in fwkdl.PollInput[sourcemetr
 	}
 
 	if spec := mapping.LoraRequestInfo; spec != nil { // extract LoRA-specific metrics
-		metric, err := spec.getLatestMetric(families)
-		if err != nil {
-			errs = append(errs, err)
-		} else if metric != nil {
+		if metric := spec.getLatestMetric(families); metric != nil {
 			populateLoRAMetrics(clone, metric, &errs)
 			updated = true
 		}
--- a/pkg/epp/framework/plugins/datalayer/extractor/metrics/loraspec.go
+++ b/pkg/epp/framework/plugins/datalayer/extractor/metrics/loraspec.go
@@ -17,8 +17,6 @@ limitations under the License.
 package metrics
 
 import (
-	"fmt"
-
 	dto "github.com/prometheus/client_model/go"
 
 	sourcemetrics "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/source/metrics"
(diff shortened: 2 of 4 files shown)


gkneighb:
CI lint flagged the now-always-nil error return on `LoRASpec.getLatestMetric` (unparam). Dropped the error from the signature entirely since the function is infallible after the fix, and updated the single caller in `extractor.go` to match. `make lint` and `make test-unit-epp` green locally.

gkneighb:
Heads up — the lint failure here is the org-wide `GO-2026-5037` (`crypto/x509@go1.25.10`) vuln that `govulncheck` is now flagging on every PR, not anything in this diff. @shmuelk is fixing it in #1468 by bumping the builder image. Once that lands I'll rebase and re-ping for `/lgtm`. Not abandoned, just gated.

gkneighb:
Rebased onto current main now that #1468 has landed — `make lint` and `make test-unit-epp` green locally on the fresh builder image. Re-LGTM at your convenience @ahg-g.

elevran (approved):

PR #1468, opened 2026-06-03
Fixes for CI after latest vulnerability fix

shmuelk:
**What type of PR is this?**
/kind bug

This PR changes the builder image to use the docker provided golang image, rather than the one from quay.io. The issue with the image on quay.io is that the tags only have major and minor making it difficult to automate updates.

In addition the is PR configures dependabopt to not make major or minor semver updates of images.

```release-note
NONE
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -54,6 +54,6 @@ updates:
     commit-message:
       prefix: "deps(docker)"
     ignore:
-      # Major/minor upgrades are handled intentionally via `make upgrade-deps`; dependabot covers patches only
+      # ßdependabot covers patches only
       - dependency-name: "*"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -53,3 +53,7 @@ updates:
       - "dependencies"
     commit-message:
       prefix: "deps(docker)"
+    ignore:
+      # ßdependabot covers patches only
```
ahg-g: typo in the comment

shmuelk: Fixed

elevran (commented):
/lgtm
/approve

elevran (approved):

PR #1475, opened 2026-06-03
Allow multiple Parsers by adding match interface and building routing based on it

zetxqx:
**What type of PR is this?**
/kind feature

This PR refactors the parser routing logic to support multiple concurrent parsers:
1. Adds the `Claims` interface to parsers which returns the supported path and protocol `Claims` information.
2. Uses the `parserRouter` to register the routing table and resolve the correct parser for incoming requests.

**Parser Selection & Context Lifecycle**:
- **Selection**: When request header and body are received, the EPP extracts the request path (checking `:path`, `x-original-path`, or `x-forwarded-path` headers) and uses `ParserRouter` to match it against the paths registered by active parsers.
  - If multiple parsers support the same path, the first parser configured in the list takes precedence (first match wins).
- The resolved parser is stored in `RequestContext.ResolvedParser`.
- Response Phase: During the response body processing phase, the server retrieves the `ResolvedParser` from the `RequestContext` to ensure the same parser is used to parse response usage metrics.

**Configuration Example**:
You can now configure multiple parsers in the `EndpointPickerConfig` as follows:
```yaml
kind: EndpointPickerConfig
...
requestHandler:
  parsers:
  - pluginRef: openai-parser
  - pluginRef: anthropic-parser
```
Priority is determined by the order of the configuration (first match wins for duplicate paths).

This work is built on top of #1448.

Fixes #

```release-note
EPP now supports configuring multiple parsers under `requestHandler.parsers` in the `EndpointPickerConfig`. The router matches the request path suffix to select the appropriate parser (first match wins for duplicate parsers supporting suffix match).
```

### DIFF
--- a/pkg/epp/framework/interface/requesthandling/plugins.go
+++ b/pkg/epp/framework/interface/requesthandling/plugins.go
@@ -43,12 +43,12 @@ type Parser interface {
 	// buffered response body and 'endOfStream' set to true.
 	ParseResponse(ctx context.Context, body []byte, headers map[string]string, endofStream bool) (*ParsedResponse, error)
 
-	// Match returns the matching criteria (paths and protocols) for this parser.
-	Match() Match
+	// Claims returns the paths and protocols claimed by this parser.
+	Claims() Claims
 }
 
-// Match defines the matching criteria for a parser.
-type Match struct {
+// Claims defines the matching criteria for a parser.
+type Claims struct {
 	Paths     []string         // path patterns this parser claims (e.g., "chat/completions")
 	Protocols []v1.AppProtocol // protocols this parser supports (e.g., "h2c")
 }
--- a/pkg/epp/framework/interface/requesthandling/plugins.go
+++ b/pkg/epp/framework/interface/requesthandling/plugins.go
@@ -43,9 +43,14 @@ type Parser interface {
 	// buffered response body and 'endOfStream' set to true.
 	ParseResponse(ctx context.Context, body []byte, headers map[string]string, endofStream bool) (*ParsedResponse, error)
 
-	// SupportedAppProtocols returns the list of supported protocols.
-	// Returning an empty list means it supports all protocols.
-	SupportedAppProtocols() []v1.AppProtocol
+	// Match returns the matching criteria (paths and protocols) for this parser.
+	Match() Match
+}
+
+// Match defines the matching criteria for a parser.
+type Match struct {
+	Paths     []string         // path patterns this parser claims (e.g., "chat/completions")
+	Protocols []v1.AppProtocol // protocols this parser supports (e.g., "h2c")
 }
 
 // ParseResult contains the result of parsing the request.
--- a/pkg/epp/framework/plugins/requesthandling/parsers/vllmhttp/vllmhttp.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/vllmhttp/vllmhttp.go
@@ -81,14 +81,14 @@ func (p *VllmHTTPParser) WithName(name string) *VllmHTTPParser {
 	return p
 }
 
-func (p *VllmHTTPParser) Match() fwkrh.Match {
-	openaiMatch := p.openai.Match()
-	paths := make([]string, 0, 1+len(openaiMatch.Paths))
+func (p *VllmHTTPParser) Claims() fwkrh.Claims {
+	openaiClaims := p.openai.Claims()
+	paths := make([]string, 0, 1+len(openaiClaims.Paths))
 	paths = append(paths, generatePathSuffix)
-	paths = append(paths, openaiMatch.Paths...)
-	return fwkrh.Match{
+	paths = append(paths, openaiClaims.Paths...)
+	return fwkrh.Claims{
 		Paths:     paths,
-		Protocols: openaiMatch.Protocols,
+		Protocols: openaiClaims.Protocols,
 	}
 }
 
--- a/pkg/epp/framework/plugins/requesthandling/parsers/vllmhttp/vllmhttp.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/vllmhttp/vllmhttp.go
@@ -29,8 +29,7 @@ import (
 	"fmt"
 	"strings"
 
-	v1 "sigs.k8s.io/gateway-api-inference-extension/api/v1"
-
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/common/request"
 	fwkplugin "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
 	fwkrh "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requesthandling"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requesthandling/parsers/openai"
@@ -41,7 +40,7 @@ const (
 	VllmHTTPParserType = "vllmhttp-parser"
 
 	// generatePathSuffix is the vLLM disaggregated Prefill/Decode API path.
-	generatePathSuffix = "/inference/v1/generate"
+	generatePathSuffix = "inference/v1/generate"
 )
 
 // compile-time type validation
@@ -82,15 +81,21 @@ func (p *VllmHTTPParser) WithName(name string) *VllmHTTPParser {
 	return p
 }
 
-// SupportedAppProtocols mirrors the OpenAI parser; we accept the same HTTP-style transports.
-func (p *VllmHTTPParser) SupportedAppProtocols() []v1.AppProtocol {
-	return p.openai.SupportedAppProtocols()
+func (p *VllmHTTPParser) Match() fwkrh.Match {
+	openaiMatch := p.openai.Match()
+	paths := make([]string, 0, 1+len(openaiMatch.Paths))
+	paths = append(paths, generatePathSuffix)
+	paths = append(paths, openaiMatch.Paths...)
+	return fwkrh.Match{
+		Paths:     paths,
+		Protocols: openaiMatch.Protocols,
+	}
 }
 
 // ParseRequest handles /inference/v1/generate locally and delegates everything
 // else to the embedded OpenAI parser.
 func (p *VllmHTTPParser) ParseRequest(ctx context.Context, body []byte, headers map[string]string) (*fwkrh.ParseResult, error) {
-	if strings.HasSuffix(openai.GetRequestPath(headers), generatePathSuffix) {
+	if strings.HasSuffix(request.GetRequestPath(headers), generatePathSuffix) {
 		return p.parseGenerateRequest(body)
 	}
 	return p.openai.ParseRequest(ctx, body, headers)
--- a/pkg/epp/handlers/parsers.go
+++ b/pkg/epp/handlers/parsers.go
@@ -22,63 +22,80 @@ import (
 
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/common/request"
 	fwkrh "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requesthandling"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-// ParserDispatcher handles the routing of incoming requests to the appropriate Parser.
-type ParserDispatcher struct {
-	routes         map[string]fwkrh.Parser
+type parserEntry struct {
+	parser          fwkrh.Parser
+	normalizedPaths []string
+}
+
+// ParserRegistry handles the routing of incoming requests to the appropriate Parser.
+type ParserRegistry struct {
+	entries        []parserEntry
 	fallbackParser fwkrh.Parser
 	parsers        []fwkrh.Parser
 }
 
-// NewParserDispatcher builds a central routing table from a list of active parsers.
-// The order of the input parsers determines the priority (first match wins for identical suffixes).
-func NewParserDispatcher(parsers []fwkrh.Parser) *ParserDispatcher {
-	dispatcher := &ParserDispatcher{
-		routes: make(map[string]fwkrh.Parser),
-	}
-	seen := make(map[fwkrh.Parser]bool)
-	for _, parser := range parsers {
-		if !seen[parser] {
-			seen[parser] = true
-			dispatcher.parsers = append(dispatcher.parsers, parser)
-		}
+// NewParserRegistry builds a central routing table from a list of active parsers.
+// The order of the input parsers determines the priority (first match wins).
+func NewParserRegistry(parsers []fwkrh.Parser) *ParserRegistry {
+	registry := &ParserRegistry{}
+	seenTypes := make(map[string]bool)
+	logger := log.Log.WithName("ParserRegistry")
 
+	for i, parser := range parsers {
 		paths := parser.Claims().Paths
+		typeName := parser.TypedName().Type
 		if len(paths) == 0 {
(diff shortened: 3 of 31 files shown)


zetxqx:
# Standalone EPP Multi-Parser Verification Report

**Date**: 2026-06-04
**Namespace**: `m1`
**EPP Image**: `us-central1-docker.pkg.dev/bobzetian-gke-dev/bobinference/epp:1475`
**Backend Model**: `Qwen/Qwen3-32B` (sim)

## Executive Summary
We successfully verified the standalone EPP deployment with multiple parsers (`openai-parser`, `anthropic-parser`, `vllmhttp-parser`) enabled concurrently. EPP correctly routed various request shapes to the backend model servers and enforced parser constraints (returning 400 Bad Request for unregistered paths).

## Configuration
The EPP was configured using a custom `EndpointPickerConfig` (`custom-plugins.yaml`) mounted via ConfigMap, enabling multiple parsers:

```yaml
    requestHandler:
      parsers:
      - pluginRef: openai-parser
      - pluginRef: anthropic-parser
      - pluginRef: vllmhttp-parser
```

## Test Results Summary

| Endpoint | Parser | Request Type | Status | EPP Routing | Backend Response |
| :--- | :--- | :--- | :--- | :--- | :--- |
| `/v1/chat/completions` | `openai-parser` | Chat Completions | **SUCCESS** | Yes | 200 OK |
| `/v1/embeddings` | `openai-parser` | Embeddings | **SUCCESS** | Yes | 404 Not Found (vLLM limitation) |
| `/v1/completions` | `openai-parser` | Text Completions | **SUCCESS** | Yes | 200 OK |
| `/v1/responses` | `openai-parser` | Responses | **SUCCESS** | Yes | 200 OK |
| `/v1/conversations` | `openai-parser` | Conversations | **SUCCESS** | Yes | 404 Not Found (vLLM limitation) |
| `/v1/messages` | `anthropic-parser` | Anthropic Messages | **SUCCESS** | Yes | 200 OK |
| `/inference/v1/generate` | `vllmhttp-parser` | vLLM Generate | **SUCCESS** | Yes | 200 OK |
| `/v1/invalid-path` | None | Invalid Path | **SUCCESS** | Blocked by EPP | 400 Bad Request (EPP error) |

*Note: 404 Not Found responses from the backend (uvicorn) indicate that EPP successfully routed the request to the backend, but the backend model/server does not support that specific endpoint. This still verifies EPP routing functionality.*

## Detailed Test Logs (Responses Omitted for Brevity)

### 1. Chat Completions (`/v1/chat/completions`)
**Request:**
```bash
curl -i -X POST http://localhost:8082/v1/chat/completions \
  -d '{"model": "Qwen/Qwen3-32B", "messages": [{"role": "user", "content": "Hello!"}], "max_tokens": 50}'
```
**Response Status:** `200 OK` (Response body omitted)

### 2. Embeddings (`/v1/embeddings`)
**Request:**
```bash
curl -i -X POST http://localhost:8082/v1/embeddings \
  -d '{"model": "Qwen/Qwen3-32B", "input": "The food was delicious."}'
```
**Response Status:** `404 Not Found` (vLLM does not support embeddings for this model)

### 3. Text Completions (`/v1/completions`)
**Request:**
```bash
curl -i -X POST http://localhost:8082/v1/completions \
  -d '{"model": "Qwen/Qwen3-32B", "prompt": "Once upon a time", "max_tokens": 50}'
```
**Response Status:** `200 OK` (Response body omitted)

### 4. Responses (`/v1/responses`)
**Request:**
```bash
curl -i -X POST http://localhost:8082/v1/responses \
  -d '{"model": "Qwen/Qwen3-32B", "input": [{"type": "message", "role": "user", "content": "Tell me a joke."}]}'
```
**Response Status:** `200 OK` (Response body omitted)

### 5. Conversations (`/v1/conversations`)
**Request:**
```bash
curl -i -X POST http://localhost:8082/v1/conversations \
  -d '{"model": "Qwen/Qwen3-32B", "items": [{"type": "message", "role": "user", "content": "Start a new conversation."}]}'
```
**Response Status:** `404 Not Found` (vLLM does not support conversations for this model)

### 6. Anthropic Messages (`/v1/messages`)
**Request:**
```bash
curl -i -X POST http://localhost:8082/v1/messages \
  -d '{"model": "Qwen/Qwen3-32B", "max_tokens": 10, "messages": [{"role": "user", "content": "Hello, Claude"}]}'
```
**Response Status:** `200 OK` (Response body omitted)

### 7. vLLM Generate (`/inference/v1/generate`)
**Request:**
```bash
curl -i -X POST http://localhost:8082/inference/v1/generate \
  -d '{"model": "Qwen/Qwen3-32B", "token_ids": [1, 2, 3, 4, 5], "sampling_params": {"temperature": 0.0, "max_tokens": 10}}'
```
**Response Status:** `200 OK` (Response body omitted)

### 8. Trigger Parser Not Found Error (`/v1/invalid-path`)
**Request:**
```bash
curl -i -X POST http://localhost:8082/v1/invalid-path \
  -d '{"model": "Qwen/Qwen3-32B"}'
```
**Response Status:** `400 Bad Request`
**Response Body:**
```
inference error: BadRequest - no parser registered matching path suffix for: /v1/invalid-path
```

zetxqx:
@tessapham  @ahg-g  this should be ready for review now

I changed the match interface a litte bit based on feed back, now it returns the match criteria for path and protocol. We have a parserRouter to route to different parsers based on path suffix

pkg/epp/handlers/parsers.go
```diff
@@ -38,6 +40,61 @@ func (c *Config) String() string {
...
+			}
+		}
+	}
+
+	return router
+}
+
+// Route resolves an incoming request path to the matching Parser using suffix matching.
+func (pr *ParserRouter) Route(path string) (fwkrh.Parser, error) {
+	for suffix, parser := range pr.routes {
```
elevran: `Route()` iterates over a `map[string]fwkrh.Parser` (see earlier coment - perhaps it should iterate `Claims` instead?), which has non-deterministic order in Go. If `MatchPathSuffix` can match more than one registered suffix for a given path, the resolved parser is random. Use a slice of `(suffix, parser)` pairs in insertion order so the first-configured parser consistently wins, matching the priority contract documented in `NewParserRouter`.
Consider, for examle, `/completions` and `chat/completions` (not the real endpoints, I know, solely to make the discussion concrete).

pkg/epp/framework/interface/requesthandling/plugins.go
```diff
@@ -43,9 +43,14 @@ type Parser interface {
 	// buffered response body and 'endOfStream' set to true.
 	ParseResponse(ctx context.Context, body []byte, headers map[string]string, endofStream bool) (*ParsedResponse, error)
 
-	// SupportedAppProtocols returns the list of supported protocols.
-	// Returning an empty list means it supports all protocols.
-	SupportedAppProtocols() []v1.AppProtocol
+	// Match returns the matching criteria (paths and protocols) for this parser.
+	Match() Match
```
elevran: In `Match() Match`, both method and return type share the same identifier which reads ambiguously: callers see `parser.Match()` which suggests a predicate rather than a query for routing criteria. 
Consider changing the interface method name and return type.

**Suggestion**: 

```go
type Claim struct {
    Protocol string
    Method   string // optional, useful if we want to catch invalid calls early
    Path     string
}

type Parser interface {
    ...
    Claims() []Claim
}
```

Also, tend to think of "vllm", "OpenAI", "Anthropic" as the parsers and they support **multiple** paths and hence should have multiple claims. Would put entire parser in one package and not split into multiple, per path parsers.

pkg/epp/handlers/parsers.go
```diff
@@ -38,6 +40,61 @@ func (c *Config) String() string {
...
 
-func NewParser(config *Config) fwkrh.Parser {
-	return config.Parser
+// NewParsers returns the configured parsers.
+func NewParsers(config *Config) []fwkrh.Parser {
+	return config.Parsers
+}
+
+// ParserRouter handles the routing of incoming requests to the appropriate Parser.
+type ParserRouter struct {
```
elevran: nit: maybe Dispatcher would be appropriate?

pkg/epp/handlers/parsers.go
```diff
@@ -38,6 +40,61 @@ func (c *Config) String() string {
...
+	routes         map[string]fwkrh.Parser
```
elevran: Suggestion:
if we use `Claims`, we can consider directly dispatching on that instead of iterating on `Matches()` every time. The protcocol (grpc, http, h2) is readily available as well as the path so dispatch becomes O(1) lookup.

Also, when adding a new claim from a parser, it is easy to find and log conflicts with previously registered parsers.

pkg/epp/handlers/parsers.go
```diff
@@ -18,27 +18,73 @@ package handlers
...
+		if !seen[parser] {
+			seen[parser] = true
+			dispatcher.parsers = append(dispatcher.parsers, parser)
+		}
+
+		paths := parser.Claims().Paths
+		if len(paths) == 0 {
+			// A parser with no paths acts as a fallback. Stop processing subsequent
+			// parsers once a fallback is registered, as it will capture all remaining traffic.
+			if dispatcher.fallbackParser == nil {
```
ahg-g: nit: no need for the if statement, we set this once anyway

pkg/epp/handlers/server.go
```diff
@@ -121,6 +122,7 @@ type RequestContext struct {
 	ResponseStatusCode        string
 	RequestRunning            bool
 	Request                   *Request
+	ResolvedParser            fwkrh.Parser
```
ahg-g: nit: just `Parser`

pkg/epp/framework/plugins/requesthandling/parsers/vllmhttp/vllmhttp.go
```diff
@@ -82,15 +81,21 @@ func (p *VllmHTTPParser) WithName(name string) *VllmHTTPParser {
...
 }
 
 // ParseRequest handles /inference/v1/generate locally and delegates everything
 // else to the embedded OpenAI parser.
 func (p *VllmHTTPParser) ParseRequest(ctx context.Context, body []byte, headers map[string]string) (*fwkrh.ParseResult, error) {
-	if strings.HasSuffix(openai.GetRequestPath(headers), generatePathSuffix) {
+	if strings.HasSuffix(request.GetRequestPath(headers), generatePathSuffix) {
 		return p.parseGenerateRequest(body)
 	}
 	return p.openai.ParseRequest(ctx, body, headers)
```
ahg-g, line 101: I think they did this because we didn't have multi-parser support, with multi-parser support, we wouldn't need this and we can make this parser claim only the generate path @roytman @revit13

pkg/epp/handlers/parsers.go
```diff
@@ -18,27 +18,73 @@ package handlers
...
 
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/common/request"
 	fwkrh "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requesthandling"
 )
 
-// Config holds the configuration for the Parser.
-type Config struct {
-	Parsers []fwkrh.Parser
+// ParserDispatcher handles the routing of incoming requests to the appropriate Parser.
+type ParserDispatcher struct {
```
ahg-g: nit: how about we call it ParserRegistry? dispatcher has a strong meaning in my opinion that what this component does.

pkg/epp/handlers/parsers.go
ahg-g: oh, just saw this, I left a comment that dispatcher might be too strong for what this component does and suggested ParserRegistry

pkg/epp/handlers/parsers.go
zetxqx: Really good call, I think this might reveal a potential bigger problem using the suffix match approach we have now.

Use  `/completion` and `/chat/completion` as an example. If Parser A only implements `chat/completion` Parser B only implements `/completion` and  when a request comes in with `/completion` as suffix, the request may match the parser A. Then this request will fail parsing because it should match parserB. Currently it works, only because we have /completion and /chat/completion both supports in the openAI parser.

should we implement in this way: 1. first we honor the shortest suffix match 2. then we honor the parser order

I think currently it is fine because the endpoint we support is limited.

pkg/epp/framework/interface/requesthandling/plugins.go
zetxqx: Yes, entire parser could be in on package. but I want to make different API parser a logical group. So that people can always add the parserRequest and ParserResponse logic in one logical pacakge.

as for Claims, I would like to keep the path and protocols separated because I don't want to generate combination of different protocol and different paths.

```
// Claims defines the matching criteria for a parser.
type Claims struct {
	Paths     []string         // path patterns this parser claims (e.g., "chat/completions")
	Protocols []v1.AppProtocol // protocols this parser supports
}
```

pkg/epp/framework/plugins/requesthandling/parsers/vllmhttp/vllmhttp.go
zetxqx: agree, I want to remove this in a separate PR, in that PR I will:

1. remove openai parser from VllmHTTPParser
2. set `opeanai-parser` `anthropic-parser` `vllmhttparser` as default parser if no parser is set. Currently we only set openai-parser as default

pkg/epp/handlers/parsers.go
zetxqx: adopted ParserRegistry

pkg/epp/handlers/parsers.go
zetxqx: used the `Claims`.  currently the dispatch is based on suffix, so as you pointed out, currently I still have to loop through parsers paths, in case the suffix match multiple claimed paths. But I think using claims can still early warning a lot of info at the config loading level.

pkg/epp/handlers/parsers.go
zetxqx: Checking the history we want to drop "v1" from "v1/chat/completion" and "v1/completion" because some provider like Azure and Vertex is not using "v1" in the path: https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2813

Currently I implemented in the way that we'll always loop through the ordered parsers to find the correct parser to parse this request.

pkg/epp/framework/plugins/requesthandling/parsers/vllmhttp/vllmhttp.go
zetxqx: Had a PR based on this: #1488

ahg-g:
Thanks @zetxqx  pls fix the tests and we can move forward with this.

pkg/epp/handlers/parsers.go
elevran: Why not match on the whole path?

zetxqx: @elevran  the whole path for cloud providers is different:
1. vertex: /v1/projects/{project}/locations/{location}/endpoints/{endpoint}/chat/completions
2. azure: /openai/deployments/{deployment-id}/chat/completions

I think, if we want to do full match we can either:
1. let the user to always set the correct full path in header
2. add a configurable parameter in parser to let user configure the path they want to match
e.g.
```
  - type: openai-parser
    parameters:
        - path: "/v1/projects//locations/{location}/endpoints/{endpoint}/chat/completions"
```

Issue #1474, opened 2026-06-03
concurrency-detector's Saturation() method always returns 0


RishabhSaini:
The concurrency-detector plugin's Saturation() method always returns 0, preventing the flow control dispatch cycle from ever triggering backpressure.
Saturation() reads InFlightLoadDataKey from endpoints returned by EndpointCandidates.Locate(), which are persistent datastore ModelServer objects. However, InFlightLoadProducer.Produce() only writes this key to per-request scheduling clones created during the scheduling cycle. These clones are discarded after scheduling completes, so the persistent endpoints' AttributeMap is always empty.
The internal requestTracker/tokenTracker in the producer correctly track in-flight requests via PreRequest/OnEvicted, but this state is never surfaced to the persistent endpoints that the dispatch cycle reads from.

  Reproduction:
  1. Enable flow control with concurrency-detector (maxConcurrency: 16)
  2. Send 50+ concurrent requests
  3. Observe llm_d_router_epp_flow_control_pool_saturation stays at 0
  4. Observe llm_d_router_epp_average_running_requests shows real load

Proposed Fix:
In InFlightLoadProducer.Extract(), cache persistent endpoint references on EventAddOrUpdate. In PreRequest()/OnEvicted()/releaseTokensEarly(), sync the cumulative tracker state to the persistent endpoint's AttributeMap. This preserves the existing clone-based flow for schedulers while making the data available to the dispatch cycle.

LukeAVanDrie:
I suspect this was introduced between GAIE 0.4 and 0.5.

I tested concurrency detector when I wrote the well-lit path guide. After the guide was submitted, we bumped the chart version from 0.4 to 0.5 in https://github.com/llm-d/llm-d/pull/1353.

Which means we can bisect changes between 0.4 and 0.5 to find where this regression was introduced.

LukeAVanDrie:
IIRC there was a ton of work on token estimator data (early work for token data producer and similar). I suspect one of the refactoring PRs for that.

RishabhSaini:
I trace it down to: https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2718
That refactor replaced the direct tracker read with getLoad(e.GetAttributes()). But the AttributeMap on persistent endpoints is never populated. The tracker was moved to the InFlightLoadProducer, which only writes to scheduling clones.

LukeAVanDrie:
[...]

PR #1478, opened 2026-06-04
Add dynamic attributes for EPP AttributeMap

LukeAVanDrie:
**What type of PR is this?**

/kind bug

In the plugin framework, the flow control saturation detection mechanism did not function as intended because the `concurrency-detector` read load data from persistent `datalayer.Endpoint` objects in the `Datastore` during admission control, before request scheduling clones (which are populated with in-flight load by `InFlightLoadProducer`) were created. This resulted in `Saturation()` always returning `0` (no load).

This PR introduces dynamic attributes (getter providers) to the `AttributeMap` implementation, allowing endpoints to resolve live data directly from the source of truth (the plugin's thread-safe trackers) on-demand, without requiring complex write-backs during the request lifecycle.

 Related to #1474

**Release note**:
```release-note
NONE
```

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -664,23 +664,6 @@ func (r *Runner) parseConfigurationPhaseTwo(ctx context.Context, rawConfig *conf
 	r.parserRegistry = cfg.ParserRegistry
 	logger.Info("loaded configuration from file/text successfully")
 
-	var loadProducer *inflightload.InFlightLoadProducer
-	for _, p := range handle.GetAllPlugins() {
-		if prod, ok := p.(*inflightload.InFlightLoadProducer); ok {
-			loadProducer = prod
-			break
-		}
-	}
-
-	if loadProducer != nil {
-		key := attrconcurrency.InFlightLoadDataKey.WithNonEmptyProducerName(loadProducer.TypedName().Name)
-		r.dlRuntime.RegisterAttributeProvider(key.String(), func(eid string) fwkdl.Cloneable {
-			return &attrconcurrency.InFlightLoad{
-				Tokens:   loadProducer.GetTokens(eid),
-				Requests: loadProducer.GetRequests(eid),
-			}
-		})
-	}
 
 	return cfg, nil
 }
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -664,6 +664,24 @@ func (r *Runner) parseConfigurationPhaseTwo(ctx context.Context, rawConfig *conf
 	r.parserRegistry = cfg.ParserRegistry
 	logger.Info("loaded configuration from file/text successfully")
 
+	var loadProducer *inflightload.InFlightLoadProducer
+	for _, p := range handle.GetAllPlugins() {
+		if prod, ok := p.(*inflightload.InFlightLoadProducer); ok {
+			loadProducer = prod
+			break
+		}
+	}
+
+	if loadProducer != nil {
+		key := attrconcurrency.InFlightLoadDataKey.WithNonEmptyProducerName(loadProducer.TypedName().Name)
+		r.dlRuntime.RegisterAttributeProvider(key.String(), func(eid string) fwkdl.Cloneable {
+			return &attrconcurrency.InFlightLoad{
+				Tokens:   loadProducer.GetTokens(eid),
+				Requests: loadProducer.GetRequests(eid),
+			}
+		})
+	}
+
 	return cfg, nil
 }
 
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -664,7 +664,6 @@ func (r *Runner) parseConfigurationPhaseTwo(ctx context.Context, rawConfig *conf
 	r.parserRegistry = cfg.ParserRegistry
 	logger.Info("loaded configuration from file/text successfully")
 
-
 	return cfg, nil
 }
 
--- a/pkg/epp/datalayer/runtime.go
+++ b/pkg/epp/datalayer/runtime.go
@@ -59,6 +59,9 @@ type Runtime struct {
 
 	collectors *collectorManager // per-endpoint poller, keyed by namespaced name
 	logger     logr.Logger       // Set in Configure; used where no context is available (e.g. ReleaseEndpoint).
+
+	provMu    sync.RWMutex
+	providers map[string]func(endpointID string) fwkdl.Cloneable
 }
 
 const (
@@ -80,6 +83,7 @@ func NewRuntime(pollingInterval time.Duration) *Runtime {
 		extractors:      newExtractorMap(),
 		collectors:      newCollectorManager(),
 		logger:          logr.Discard(),
+		providers:       make(map[string]func(endpointID string) fwkdl.Cloneable),
 	}
 }
 
@@ -220,6 +224,16 @@ func (r *Runtime) Register(reg fwkdl.PendingRegistration) error {
 	r.pendingRegistrations = append(r.pendingRegistrations, reg)
 	r.pendingMu.Unlock()
 	return nil
+}
+
+// RegisterAttributeProvider registers a provider function for a dynamic attribute.
+func (r *Runtime) RegisterAttributeProvider(key string, provider func(endpointID string) fwkdl.Cloneable) {
+	r.provMu.Lock()
+	defer r.provMu.Unlock()
+	if r.providers == nil {
+		r.providers = make(map[string]func(endpointID string) fwkdl.Cloneable)
+	}
+	r.providers[key] = provider
 }
 
 // registerSource dispatches src to the matching variant manager. g enforces
@@ -371,18 +385,29 @@ func (r *Runtime) NewEndpoint(ctx context.Context, endpointMetadata *fwkdl.Endpo
 	logger, _ := logr.FromContext(ctx)
 	logger = logger.WithValues("endpoint", endpointMetadata.GetNamespacedName())
 
+	endpoint := fwkdl.NewEndpoint(endpointMetadata, nil)
+	eid := endpointMetadata.GetNamespacedName().String()
+
+	r.provMu.RLock()
+	for key, provider := range r.providers {
+		endpoint.GetAttributes().Put(key, &fwkdl.DynamicAttribute{
+			Get: func() fwkdl.Cloneable {
+				return provider(eid)
+			},
+		})
+	}
+	r.provMu.RUnlock()
+
 	dispatchers := make([]fwkdl.PollingDispatcher, 0, r.dispatchers.Count())
 	for _, d := range r.dispatchers.Dispatchers() {
 		dispatchers = append(dispatchers, d)
 	}
 	if len(dispatchers) == 0 {
 		logger.Info("No polling sources configured, creating endpoint without collector")
-		endpoint := fwkdl.NewEndpoint(endpointMetadata, nil)
 		r.dispatchEndpointEvent(ctx, logger, fwkdl.EndpointEvent{Type: fwkdl.EventAddOrUpdate, Endpoint: endpoint})
 		return endpoint
 	}
 
-	endpoint := fwkdl.NewEndpoint(endpointMetadata, nil)
 	collector := NewCollector()
 
 	key := endpointMetadata.GetNamespacedName()
--- a/pkg/epp/datalayer/runtime.go
+++ b/pkg/epp/datalayer/runtime.go
@@ -59,9 +59,6 @@ type Runtime struct {
 
 	collectors *collectorManager // per-endpoint poller, keyed by namespaced name
 	logger     logr.Logger       // Set in Configure; used where no context is available (e.g. ReleaseEndpoint).
-
-	provMu    sync.RWMutex
-	providers map[string]func(endpointID string) fwkdl.Cloneable
 }
 
 const (
@@ -83,7 +80,6 @@ func NewRuntime(pollingInterval time.Duration) *Runtime {
 		extractors:      newExtractorMap(),
 		collectors:      newCollectorManager(),
 		logger:          logr.Discard(),
-		providers:       make(map[string]func(endpointID string) fwkdl.Cloneable),
 	}
 }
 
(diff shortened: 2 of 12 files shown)


cmd/epp/runner/runner.go
```diff
@@ -660,6 +660,24 @@ func (r *Runner) parseConfigurationPhaseTwo(ctx context.Context, rawConfig *conf
...
+	if loadProducer != nil {
+		key := attrconcurrency.InFlightLoadDataKey.WithNonEmptyProducerName(loadProducer.TypedName().Name)
+		r.dlRuntime.RegisterAttributeProvider(key.String(), func(eid string) fwkdl.Cloneable {
+			return &attrconcurrency.InFlightLoad{
+				Tokens:   loadProducer.GetTokens(eid),
+				Requests: loadProducer.GetRequests(eid),
+			}
+		})
+	}
+
```
shmuelk: Is this code going to be repeated for every plugin that has the same issue?

Maybe the cloning of the endpoint data is simply a bad idea and what is needed is shallow copy (i.e. a new slice of the existing endpoint elements)

test/integration/epp/dynamic_attributes_test.go
```diff
@@ -0,0 +1,225 @@
...
+- type: round-robin-fairness-policy
+- type: fcfs-ordering-policy
+- type: concurrency-detector
+  parameters:
+    maxConcurrency: 1
+    concurrencyMode: requests
+    headroom: 0.0
+- type: mock-metrics-source
+requestHandler:
+  parser:
```
elevran: nit: heads up The test config uses `requestHandler.parser`: (singular). Once #1448 lands this will need to change to `requestHandler.parsers` to match the new schema.

pkg/epp/datalayer/runtime.go
```diff
@@ -80,6 +83,7 @@ func NewRuntime(pollingInterval time.Duration) *Runtime {
 		extractors:      newExtractorMap(),
 		collectors:      newCollectorManager(),
 		logger:          logr.Discard(),
+		providers:       make(map[string]func(endpointID string) fwkdl.Cloneable),
```
elevran: Q: What is the rationale for registering and managing dynamic attribute providers? Can it be treated directly as an attribute without the datalayer's and runner's awareness of the providers?

For example, `InFlightLoadProducer` already implements `EndpointExtractor` and receives `EventAddOrUpdate` events with the live `fwkdl.Endpoint` object. It just currently ignores that event (only handles `EventDelete` for cleanup). It could handle EventAddOrUpdate and call:

```go
  endpoint.GetAttributes().Put(key, &DynamicAttribute{Get: func() Cloneable { ... }})
```

directly on the live endpoint. When `toSchedulerEndpoints` clones the attribute map, `DynamicAttribute.Clone()` copies the function pointer and the cloned endpoint also returns live counter values. No changes to Runtime needed, no type-assertion loop in runner.go, and the producer is fully self-contained.

The `DynamicAttribute` type is still needed, just the `RegisterAttributeProvider` machinery in `Runtime` and the wiring in `runner.go` become unnecessary.

elevran (commented):
The mechanism is useful, However, by placing some constraints (I _think_) we can make the implementation simpler.
See below.

Also @shmuelk's comment raises the broader question of whether snapshotting endpoints for scheduling is **still** the right model, since it forces workarounds like this for any state that changes during a request. That is a valid architectural discussion but out of scope for this PR.
The engineering tradeoff involves (e.g.,) lock contention, GC, and value consistency/coherency. Might be worthwhile to discuss in an issue / have a discussion doc.
@ahg-g @liu-cong @vMaroon

shmuelk:
@elevran said:

> Also @shmuelk's comment raises the broader question of whether snapshotting endpoints for scheduling is still the right model, since it forces workarounds like this for any state that changes during a request. That is a valid architectural discussion but out of scope for this PR.

While it is out of scope of this PR, it is an alternate solution to the issue at hand. A much simpler one at that.

cmd/epp/runner/runner.go
ahg-g: In most cases the data stored is not only per endpoint, but also per request; for example the prefix-match length

ahg-g: > Is this code going to be repeated for every plugin that has the same issue?

I agree we better find an approach where we don't require this.

pkg/epp/datalayer/runtime.go
ahg-g: I like that! I think this is the cleanest solution.

ahg-g:
> Also @shmuelk's comment raises the broader question of whether snapshotting endpoints for scheduling is still the right model, since it forces workarounds like this for any state that changes during a request. That is a valid architectural discussion but out of scope for this PR.

Since some attributes are not only per endpoint but also per request (like prefix match length), a clone is a must.

liu-cong:
@LukeAVanDrie Isn't the root cause of this the `admissionController.Admit` runs before the data producers (which produces data to the snapshotted endpoints)? You had a proposal to run the appropriate data producers before the `admissionController.Admit`. Why don't go with that route?

BTW, I found `admissionController.Admit` and `admissionPlugins` confusing. Note we also have a `preAdmissionPlugins` which runs before the `admissionController.Admit`, not ``admissionPlugins`... Should we clarify/rename some of these?

pkg/epp/datalayer/runtime.go
elevran: If we need per request dynamic attributes we can see if that is possible using existing mechanisms. Similar infra in both

LukeAVanDrie:
> @LukeAVanDrie Isn't the root cause of this the `admissionController.Admit` runs before the data producers (which produces data to the snapshotted endpoints)? You had a proposal to run the appropriate data producers before the `admissionController.Admit`. Why don't go with that route?

That proposal ([[PUBLIC] Dynamic Plugin Execution & Shared Attributes in EPP)](https://docs.google.com/document/d/19tMcpKWfcC40xb1SGWYT7_F-J2wkAHHz5MC9mgFd2eE/edit?usp=sharing) is complementary and works well for Request-scoped attributes (which are static for the lifetime of the request, like prompt token count), but it does not solve the issue for Endpoint-scoped attributes (like in-flight load) due to staleness.

If we run `InFlightLoadProducer` at request arrival ($t_0$) to populate the endpoint attributes, and the request then waits in the flow control queue for $\Delta t$, those attributes will be stale by the time the request reaches the head of the queue. The real load of the endpoints will have changed as other requests complete. The flow control dispatch loop runs continuously and needs to evaluate the live saturation of the pool. By using dynamic Attributes, the `AttributeMap` on the endpoints always returns the fresh, real-time value from the plugin's trackers, ensuring we don't make dispatch decisions based on stale  snapshots.

For this reason, my proposal explicitly restricts which data producers can be ran before admission to `RequestScope` attributes to guard against this staleness at the framework level. Now, if we are being really precise, the target of the data (whether it belongs to a Request or an Endpoint) is likely orthogonal to its lifecycle (whether it is Static or Dynamic/lazy).

LukeAVanDrie:
> BTW, I found `admissionController.Admit` and `admissionPlugins` confusing. Note we also have a `preAdmissionPlugins` which runs before the `admissionController.Admit`, not ``admissionPlugins`... Should we clarify/rename some of these?

Yes, agreed. @elevran has also flagged this. I will create an issue / proposal for collapsing this into a smaller plugin surface.

pkg/epp/datalayer/runtime.go
LukeAVanDrie: I like this approach, thanks for the suggestion! I wasn't too familiar with the event system.

cmd/epp/runner/runner.go
LukeAVanDrie: Removed following @elevran's suggestion below.

test/integration/epp/dynamic_attributes_test.go
LukeAVanDrie: Thanks, rebased and fixed.

pkg/epp/datalayer/runtime.go
LukeAVanDrie: Applied!

liu-cong:
The term "dynamic attributes" is ambiguous because attributes should be dynamic by default. The current issue is that the scheduler loop takes a snapshot and the producers produce to that snapshot. I think data producers should output to a dynamic map, with snapshots handled exclusively by the scheduler.  @ahg-g @elevran @shmuelk thoughts?

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
```diff
@@ -175,16 +175,29 @@ func (p *InFlightLoadProducer) RegisterDependencies(r datalayer.Registrar) error
...
 	id := event.Endpoint.GetMetadata().NamespacedName.String()
 
-	p.DeleteEndpoint(id)
-	log.FromContext(ctx).V(logutil.DEFAULT).Info("Cleaned up in-flight load for deleted endpoint", "endpoint", id)
+	switch event.Type {
+	case datalayer.EventDelete:
+		p.DeleteEndpoint(id)
+		log.FromContext(ctx).V(logutil.DEFAULT).Info("Cleaned up in-flight load for deleted endpoint", "endpoint", id)
+	case datalayer.EventAddOrUpdate:
+		event.Endpoint.GetAttributes().Put(p.dk.String(), &datalayer.DynamicAttribute{
```
ahg-g, line 202: This attribute will be overwritten on the cloned endpoints that we pass to the request handling plugins. I know it will still work, but I think it will be confusing down the line, here is my recommendation:

1) Remove the request's non-cached tokens parameter into its own attribute, call it `UncachedRequestTokens`. The InFlightLoad attribute only tracks the endpoint's tokens and requests.
2) Here we populate the `InFlightLoad` while in Produce we only populate the `UncachedRequestTokens`

ahg-g:
> The term "dynamic attributes" is ambiguous because attributes should be dynamic by default. The current issue is that the scheduler loop takes a snapshot and the producers produce to that snapshot. I think data producers should output to a dynamic map, with snapshots handled exclusively by the scheduler. @ahg-g @elevran @shmuelk thoughts?

There are two types of per-endpoint attributes, cross requests and request scoped. There are two solutions to track the request scoped ones: 1) add request id as part of the attribute key 2) cloning the endpoint (current solution). I think cloning is a decent solution, and I am not sure if the former approach is necessarily better.

I agree on fining a better name for attributes that are not request-scoped if we can.

liu-cong:
> There are two types of per-endpoint attributes, cross requests and request scoped. There are two solutions to track the request scoped ones: 1) add request id as part of the attribute key 2) cloning the endpoint (current solution). I think cloning is a decent solution, and I am not sure if the former approach is necessarily better.

I think the attribute in question here is per endpoint attribute (cross request). The reason we clone the endpoint state during scheduling was primarily reducing lock contention (arguable). Because scheduling cycle is generally fast(no queuing), using a "static" cloned state is OK.

The root cause of this, I believe, is that currently data producers write to the cloned endpoint state, and therefore only scheduling plugins have access. I think data producers should write to the source endpoint (and thus dynamic).

ahg-g:
> I think the attribute in question here is per endpoint attribute (cross request). The reason we clone the endpoint state during scheduling was primarily reducing lock contention (arguable). Because scheduling cycle is generally fast(no queuing), using a "static" cloned state is OK.

Yes, but there are other attributes that are per request and we need a solution for those. What I suggested above are the two alternatives, either clone the endpoint or allow having request keyed attributes. I think the former is a reasonable solution that also as you mentioned helps avoid race conditions in some cases.

liu-cong:
> Yes, but there are other attributes that are per request and we need a solution for those. What I suggested above are the two alternatives, either clone the endpoint or allow having request keyed attributes. I think the former is a reasonable solution that also as you mentioned helps avoid race conditions in some cases.

Makes sense. For request scoped attributes should be added to the request, not to the endpoints. I was confused by that but now I understand.

Suggestions:
1. Request level attributes are almost "static" - they are produced once and consumed by scheduling plugins. They should be put in the `scheduling.InferenceRequest.attributes`. Right now some of them are written to the `endpoint.AttributeMap`.
1. Endpoint attributes are "dynamic" - they change per request. They should be written to the source endpoints as the source of truth. The endpoint state MAY be cloned (and immutable) per scheduling loop.

ahg-g:
> Request level attributes are almost "static" - they are produced once and consumed by scheduling plugins. They should be put in the scheduling.InferenceRequest.attributes. Right now some of them are written to the endpoint.AttributeMap.

Right, but the attribute is per request and per endpoint (e.g., prefix-match length). So the attribute a producer stores in scheduling.InferenceRequest.attributes will need to be map keyed by endpoints, ideally we offer a standard type that all such producers use (like the PluginState one).

In any case, this can be a followup.

liu-cong:
What do you think about this:
1. For now, pass both the original and snapshotted endpoints to the data producers. Modify the inflight load producer to write to the original endpoints.
2. Follow up 1 - let's decide where to put the request scoped data. IMO it should natually belong to the request object. The fact we use cloned endpoint is just a convenience - that it happened some request scoped data is also keyed by endpoint.
3. Follow up 2 - let's figure out if we want to reorder where some producers should run.

@ahg-g @LukeAVanDrie

ahg-g:
> For now, pass both the original and snapshotted endpoints to the data producers. Modify the inflight load producer to write to the original endpoints.

Why not continue with the approach already implemented by this PR? it is not invasive and in my opinion reasonable. Changing the data producer interface to pass in both sets of endpoints is not necessarily better.

This will fix the current bug which is the priority.

A followup to potentially improve this design is fair, but I am not sure it is urgent.

liu-cong:
I discussed with @LukeAVanDrie , the data produce/consume pattern today is not well suited for this case and we plan to take a follow up action on clarifying the different data access patterns and hopefully achieve a cleaner state.

The approach in this PR is reasonable. We will also need to follow up on whether making this the generalizable pattern moving forward.

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
LukeAVanDrie: I split the keys in my latest commit. Let me know what you think!

ahg-g:
@LukeAVanDrie I assume we verified the fix end-to-end and, add a nightly test will be added for the guide, right? Do we have an issue to track that in llm-d/llm-d?

PR #1481, opened 2026-06-04
ci: cache HuggingFace tokenizer for e2e-router extended suite

### LINKED ISSUE #1480: [CI tests] Cache HuggingFace model tokenizer for e2e-router (extended) suite in CI
The extended suite of e2e-router ([.github/workflows/ci-pr-checks.yaml:392-394](vhttps://github.com/llm-d/llm-d-router/blob/9719f955be36dffeae92ca0c3678e726f59838a7/.github/workflows/ci-pr-checks.yaml#L394) brings up the `vllm-render` sidecar, which downloads `Qwen/Qwen2.5-1.5B-Instruct` tokenizer from HuggingFace on every run.

The download happens because the model-cache volume backing /root/.cache/huggingface in the render sidecar is an emptyDir, so each kind cluster starts cold. Today's CI caches Go modules and build output (via e2e-runner-setup) but does not cache HF tokenizer and JSON config files.

### Costs:

- Wall-clock cost on the critical path (download is inside vLLM startup; failures surface as a probe timeout, not a clean CI step error).
- Exposure to HF availability and rate limits on every PR, leads to timeout errors
- Wasted egress.

### Evidence of errors 
```
ERROR 06-04 06:52:37 [repo_utils.py:47] 429 Too Many Requests for url: https://huggingface.co/api/models/Qwen/Qwen2.5-1.5B-Instruct/tree/main?recursive=true&expand=false., retrying 1 of 2
  ERROR 06-04 06:52:39 [repo_utils.py:45] Error retrieving file list: (Amz CF ID: nQXvgdviz5o_5R-a9XRTP02ijLNalQQsi9P479Q3HDIL0k9Jbqchog==)
  ERROR 06-04 06:52:39 [repo_utils.py:45]
  ERROR 06-04 06:52:39 [repo_utils.py:45] 429 Too Many Requests for url: https://huggingface.co/api/models/Qwen/Qwen2.5-1.5B-Instruct/tree/main?recursive=true&expand=false.
```

### Proposal
Cache the HF directory on the GHA runner with actions/cache, mount that path into the kind node via extraMounts, and rewrite the model-cache volume from emptyDir to hostPath at test time only. The base manifest is unchanged; production deployments stay on emptyDir.

1. Add an actions/cache step in .github/actions/e2e-runner-setup/action.yml keyed on the model name plus a manual revision suffix (hf-cache-${runner.os}-Qwen2.5-1.5B-Instruct-v1).
2. Pass the host cache path via env into make test-e2e-scheduler-run, threaded through E2E_ENV_VARS so it reaches the Go test process inside the builder container.
3. In the Go suite, when the env var is set:
      - Splice an extraMounts entry into kindClusterConfig mapping the host path to /opt/hf-cache on the kind node.
      - Rewrite the model-cache volume in render-sidecar deployments from emptyDir to hostPath: /opt/hf-cache. This follows the existing pattern in [test/e2e/utils_test.go:210-241](vscode-webview://11smrsqo00p7eoj1i4bnfqe7pldc63rltbru18rpl6n7efrdvkkl/test/e2e/utils_test.go#L210-L241) where deployments are mutated in Go rather than via a separate kustomize overlay.
4. Gate the wiring on matrix.suite.needs-renderer == 'true'. Other suites (sim only) are untouched.

Out of scope:
- No change to base deploy/components/inference-gateway/deployment.yaml.
- No new kustomize overlay.
- No pre-download with huggingface-cli. That is a separate option (tracked-or-not as a follow-up) if cold-start latency or stale-revision pain shows up after option 1 lands.

roytman:
**What type of PR is this?**

/kind test

## Summary

The `extended` suite of `e2e-router` brings up the `vllm-render` sidecar,
which downloads `Qwen/Qwen2.5-1.5B-Instruct` from HuggingFace on every run.
Today's CI caches Go modules but not HF tokenizer, so each run pays the full
download on the critical path. We have observed HF 429s causing flakes:
```
ERROR repo_utils.py:47 429 Too Many Requests for url:
https://huggingface.co/api/models/Qwen/Qwen2.5-1.5B-Instruct/tree/main
```
This change wires an `actions/cache` for the HF directory through to a
hostPath that backs the render sidecar's `/root/.cache/huggingface`, so
weights survive across runs.

The base deployment manifest stays on `emptyDir`; the override is
test-time YAML mutation gated on `HF_CACHE_HOST_PATH`, mirroring the
existing `removeRenderSidecar` pattern in `test/e2e/utils_test.go`.

## Changes

- `deploy/components/inference-gateway/deployment.yaml`: `model-cache`
  volume is `hostPath: /opt/hf-cache, type: DirectoryOrCreate`. Production
  deploys via the Helm chart at `config/charts/routerlib/templates/_deployment.yaml`
  and is unaffected.
- `.github/workflows/ci-pr-checks.yaml`: in the `e2e-router` job, three
  steps gated on `needs-renderer`:
  - resolve the model name via `make -s print-hf-model-name`,
  - create `${RUNNER_TEMP}/hf-cache`,
  - `actions/cache@v5` with key
    `hf-cache-${{ runner.os }}-${{ steps.model.outputs.name }}-v2`
    and a same-model `restore-keys` prefix.
  Only the renderer-using suite touches the cache key; sim suites bypass
  the cache entirely.
- `Makefile`: defines `export HF_MODEL_NAME ?= Qwen/Qwen2.5-1.5B-Instruct`
  next to `VLLM_RENDER_IMAGE`, plus a `print-hf-model-name` target. Forwards
  `HF_MODEL_NAME` and `HF_CACHE_HOST_PATH` into the builder container via
  `E2E_ENV_VARS`.
- `test/e2e/e2e_suite_test.go`: reads `HF_MODEL_NAME` from env (no
  hardcoded fallback). When `HF_CACHE_HOST_PATH` is set, splices
  `extraMounts` into the kind `kindClusterConfig`, mapping the runner's HF
  cache to `/opt/hf-cache` on the kind node.
- `test/e2e/requests_test.go`: `runCompletion` / `tryCompletion` /
  `runStreamingCompletion` accept `string` for the model name, matching
  the existing `runChatCompletion` signature; the cast to
  `openai.CompletionNewParamsModel` lives inside the helpers.

## Verified

End-to-end on a green run of `e2e-router (extended)`:
- `Cache HuggingFace files` step hit the **primary** cache key:
  `hf-cache-Linux-Qwen/Qwen2.5-1.5B-Instruct-v2` (~4 MB restored).
- No `huggingface.co` / `repo_utils.py` / `429` activity in the test logs.
- Post-job: `Cache hit ... not saving cache.` (idempotent).
- Sim suites (`pd`, `pd-shared-storage-*`, `pd-metrics`, `disruption`) all
  pass; none touch the HF cache key.

Fixes #1480

```release-note
NONE
```


test/e2e/setup_test.go
```diff
@@ -156,6 +156,8 @@ func createEndPointPicker(eppConfig string) []string {
 		})
 	if !usesTokenProducer(eppConfig) {
 		eppYamls = removeRenderSidecar(eppYamls)
+	} else {
+		eppYamls = rewriteModelCacheToHostPath(eppYamls, hfCacheHostPath)
```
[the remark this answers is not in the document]
roytman: Fix: pass hfCacheNodePath, gate the call on hfCacheHostPath != "" in the caller.

```golang
} else if hfCacheHostPath != "" {
    eppYamls = rewriteModelCacheToHostPath(eppYamls, hfCacheNodePath)
}
```

test/e2e/e2e_suite_test.go
```diff
@@ -370,5 +375,20 @@ nodes:
...
+
+// extraMountsBlock returns a YAML fragment to splice under the kind node when a
+// HuggingFace weights cache is provided. The host path is bind-mounted into the
+// kind node at hfCacheNodePath; setup_test.go points the render sidecar's
+// model-cache volume at the same node path.
+func extraMountsBlock(hostPath string) string {
+	if hostPath == "" {
+		return ""
+	}
+	return fmt.Sprintf("\n  extraMounts:\n  - hostPath: %s\n    containerPath: %s", hostPath, hfCacheNodePath)
```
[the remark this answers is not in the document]
roytman: Trivial fix in extraMountsBlock.

.github/actions/e2e-runner-setup/action.yml
```diff
@@ -45,3 +48,21 @@ runs:
...
+        mkdir -p "${RUNNER_TEMP}/hf-cache"
+        echo "path=${RUNNER_TEMP}/hf-cache" >> "$GITHUB_OUTPUT"
+
+    - name: Cache HuggingFace weights
+      uses: actions/cache@v5
+      with:
+        path: ${{ steps.hf-cache.outputs.path }}
+        key: hf-cache-${{ runner.os }}-Qwen2.5-1.5B-Instruct-v1
+        restore-keys: |
+          hf-cache-${{ runner.os }}-Qwen2.5-1.5B-Instruct-
```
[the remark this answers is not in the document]
roytman: The trade-off swaps one failure mode for another:

- Manual -vN: stale cache if a model bump forgets to bump the suffix. Rare event (model name changes seldom), and the comment in action.yml flags it. The cache key is co-located with the model name's only consumer, so reviewers should catch a missed bump.
- Hash the whole file: cache invalidates on every unrelated edit to e2e_suite_test.go (which is touched often: timeouts, label filters, env vars, helper additions). The common case becomes a forced re-download, which is exactly the cost this PR is trying to avoid.
A correct auto-buster would hash only the model name. That requires extracting kvModelName to a single source consumed by both the Go suite and the workflow (e.g. a .hf-model file or a build-time constant). That's a worthwhile follow-up but a larger refactor than this PR should carry; I'd rather land the cache first and treat single-source-of-truth as a separate change.

Keeping -vN for now.

sagearc (changes requested):
Thanks @roytman!

One note, the vllm render path only downloads/loads the tokenizer, not the actual model weights. Could you update the title/description to reflect that? Also, can you double-check whether the current change is still targeting the intended issue with that in mind?

roytman:
> Thanks @roytman!
> 
> One note, the vllm render path only downloads/loads the tokenizer, not the actual model weights. Could you update the title/description to reflect that? Also, can you double-check that the current change still targets the intended issue?

Hi @sagearc , you're right, the renderer downloads only the tokenizer and some JSON configs. During working on the PR I have checked that the cached size is around 4MB. 
However, the problem is not in the cached size, but in the request frequency. 
```
We got ERROR repo_utils.py:47 429 Too Many Requests for url:
https://huggingface.co/api/models/Qwen/Qwen2.5-1.5B-Instruct/tree/main
```

I updated the PR's description and its title

elevran:
@sagearc please remove the `requested change ` designation if agree with @roytman by submitting a different verdict in the UI

test/e2e/utils_test.go
sagearc, line 1: I think this entire yaml mutation block is unnecessary. We can cache `~/.cache/huggingface` on the runner and mount it into the kind node at the same path (swap `model-cache` from `emptyDir` to `hostPath` in the test yaml directly). No golang code needed.

.github/workflows/ci-pr-checks.yaml
```diff
@@ -442,6 +442,28 @@ jobs:
             docker load --input "$image"
           done
 
+      # HuggingFace weights cache. Only the renderer-using suite touches the
```
sagearc: we cache only the tokenizer

.github/workflows/ci-pr-checks.yaml
```diff
@@ -450,4 +472,5 @@ jobs:
           E2E_LABEL_FILTER: ${{ matrix.suite.label-filter }}
           LOAD_VLLM_RENDER_IMAGE: ${{ matrix.suite.needs-renderer }}
           PULL_VLLM_RENDER_IMAGE: "false"
+          HF_CACHE_HOST_PATH: ${{ steps.hf-cache.outputs.path }}
```
sagearc, line 482: The deployment already has `model-cache` mount. Isn't `HF_CACHE_HOST_PATH` pointing at the same thing?

.github/workflows/ci-pr-checks.yaml
```diff
@@ -442,6 +442,28 @@ jobs:
             docker load --input "$image"
           done
 
+      # HuggingFace weights cache. Only the renderer-using suite touches the
+      # cache key, so parallel matrix entries cannot race to save an empty
+      # directory under the same key. Bump the trailing -vN when kvModelName
+      # in test/e2e/e2e_suite_test.go changes; otherwise stale weights are
+      # served.
+      - name: Create HF cache dir
```
sagearc, line 457: Can we reuse the `model-cache` for hf caching? Why is this necessary?

.github/workflows/ci-pr-checks.yaml
```diff
@@ -442,6 +442,28 @@ jobs:
...
+        run: |
+          mkdir -p "${RUNNER_TEMP}/hf-cache"
+          echo "path=${RUNNER_TEMP}/hf-cache" >> "$GITHUB_OUTPUT"
+
+      - name: Cache HuggingFace weights
+        if: ${{ matrix.suite.needs-renderer == 'true' }}
+        uses: actions/cache@v5
+        with:
+          path: ${{ steps.hf-cache.outputs.path }}
+          key: hf-cache-${{ runner.os }}-Qwen2.5-1.5B-Instruct-v2
```
sagearc: model name hardcoded in cache key

.github/workflows/ci-pr-checks.yaml
```diff
@@ -442,6 +442,28 @@ jobs:
...
+          restore-keys: |
+            hf-cache-${{ runner.os }}-Qwen2.5-1.5B-Instruct-
```
sagearc: model name hardcoded in cache key

test/e2e/e2e_suite_test.go
```diff
@@ -76,6 +76,10 @@ var (
 	sideCarImage     = env.GetEnvString("SIDECAR_IMAGE", "ghcr.io/llm-d/llm-d-router-disagg-sidecar:dev", ginkgo.GinkgoLogr)
 	vllmRenderImage  = env.GetEnvString("VLLM_RENDER_IMAGE", "vllm/vllm-openai-cpu:v0.21.0", ginkgo.GinkgoLogr)
 	loadRenderImage  = env.GetEnvBool("LOAD_VLLM_RENDER_IMAGE", true, ginkgo.GinkgoLogr)
+	// hfCacheHostPath, when non-empty, is bind-mounted into the kind node at
+	// /opt/hf-cache and used as a hostPath for the render sidecar's
+	// /root/.cache/huggingface, so weights survive across CI runs.
```
sagearc: remove "weights"

test/e2e/e2e_suite_test.go
```diff
@@ -370,5 +375,21 @@ nodes:
     protocol: TCP
   - containerPort: 32090
     hostPort: ${METRICS_PORT}
-    protocol: TCP
+    protocol: TCP${EXTRA_MOUNTS}
```
sagearc, line 382: Can you check if changes here are still needed after all other comments are addressed?

sagearc (changes requested):

.github/workflows/ci-pr-checks.yaml
roytman: correct, changed to HuggingFace runtime cache (tokenizer + configs; render mode does not load weights).

.github/workflows/ci-pr-checks.yaml
roytman: They're the same logical cache, but as I understand, different layers, which is why both names exist:

- `model-cache` is the Pod volume mounted at `/root/.cache/huggingface` in the render sidecar. By default in `deployment.yaml`, it's an emptyDir (lives and dies with the pod, no persistence).
- `HF_CACHE_HOST_PATH` is a path on the GHA runner that we want that Pod volume to ultimately resolve to.
The full chain when the env var is set: runner host directory `${RUNNER_TEMP}/hf-cache` -> kind node extraMounts bind-mount at `/opt/hf-cache` -> Pod hostPath volume named `model-cache`. The Go suite rewrites the Pod volume from emptyDir to hostPath: `/opt/hf-cache` only when `HF_CACHE_HOST_PATH` is non-empty, so production behavior is unchanged.

So `model-cache` doesn't disappear; we make it reach the runner's filesystem instead of dying with the pod. `HF_CACHE_HOST_PATH` is the only handle we have on that runner directory from outside the kind cluster, which is why actions/cache needs it to know what to save/restore.

.github/workflows/ci-pr-checks.yaml
roytman: As described below, about the layers. The HF cache is what model-cache is for; this PR ensures it survives across runs.

test/e2e/e2e_suite_test.go
roytman: done, rephrased to "so tokenizer/config files survive across CI runs"

test/e2e/utils_test.go
roytman: Good call. Indeed, `deploy/components/inference-gateway/deployment.yaml` is consumed only by the e2e suite, so changing it directly is safe.

roytman:
Hi @sagearc, I think I've addressed all your comments. Can you review again please.

From the e2e-router (extended) we can see concrete cache-effectiveness signals:

- Primary-key hit, not partial restore. The log shows `Cache hit for: hf-cache-Linux-Qwen/Qwen2.5-1.5B-Instruct-v2` (verbatim) — exact key, not a restore-key: fallback. This means a prior run on this PR (or main) populated the cache, and this run reused it without modification.

- Zero HF network traffic. grep -iE "huggingface\\.co|repo_utils|429|Loading weights" against the full job log returned nothing. There are no 429 Too Many Requests errors.

- 3.7 MB restored, idempotent post-step. Tokenizer + configs only (consistent with vllm launch render not loading weights). Post-step says not saving cache — no churn.

- Plumbing verified end-to-end. The docker run invocation in the test step shows:
```
-e 'HF_MODEL_NAME=Qwen/Qwen2.5-1.5B-Instruct'
```
...which Makefile→workflow→builder -> Go all forwarded correctly. Confirmed by Go log line:


"Successfully loaded environment variable" key="HF_MODEL_NAME" value="Qwen/Qwen2.5-1.5B-Instruct"

roytman:
Recently, we have not observed the "429 Too Many Requests" issue so that I will close the PR.

Issue #968, opened 2026-05-03
After merging GAIE, replace remote CRDs definitions by local version


roytman:
The file `deploy/components/crds-gie/kustomization.yaml` uses 
```yaml
resources:
- https://github.com/kubernetes-sigs/gateway-api-inference-extension/config/crd?ref=v1.5.0
```
Now, when the CRD definitions are local, we can use the local version, which mitigates temporary access issues (e.g., timeouts) on external websites (GitHub).

elevran:
the above would still be needed for `InferencePool`, the `ModelRewrite` and `InferenceObjectives` would be locally sourced.

elevran:
Looking beyond the specific file mentioned, the full scope covers only the CRDs that remain upstream-owned (i.e., `InferenceModelRewrite` and `InferenceObjective` already moved to llm-d.ai and are checked in locally).
     
Remaining remote references:

**Kustomization files**:
- `deploy/components/crds-gie/kustomization.yaml` fetches v1-manifests.yaml for InferencePool and InferencePoolImport
- `deploy/components/crds-gateway-api/kustomization.yaml` fetches the entire standard gateway-api CRD set

**Verification scripts** (`hack/verify-manifests.sh`, `hack/verify-helm.sh`) curl-fetch the same two GIE CRDs plus a subset of gateway-api CRDs from raw.githubusercontent.com

Both [GIE](https://sigs.k8s.io/gateway-api-inference-extension) and [GW API](https://sigs.k8s.io/gateway-api) are pinned in `go.mod`, so the CRD YAML files in the module cache **are already the version-locked source of truth**. Fetching the same bytes from GitHub at apply/verify time adds a network dependency without adding version control benefit. Additionally: the Go module and the remote URL can silently diverge when one is bumped without the other; checked-in files make CRD schema changes visible in PRs; and the pattern is already established (`crds-istio/istio.yaml`, `config/crd/bases/`).

So I think we need two different approaches for two different consumers:
- **Hack scripts** copy from the Go module cache at runtime, no files to check in:
   ```console
   GIE_PATH="$(go env GOMODCACHE)/sigs.k8s.io/gateway-api-inference-extension@$(go list -m -f '{{.Version}}'
  sigs.k8s.io/gateway-api-inference-extension)"
   cp "${GIE_PATH}/config/crd/bases/inference.networking.k8s.io_inferencepools.yaml" "${TEMP_DIR}/"
   # etc.
   ```
[...]

PR #1482, opened 2026-06-04
update: reduce network fetch CRDs from github

zdtsw:
**What type of PR is this?**
/kind cleanup

**What this PR does / wh
- use local go mod to get CRDs and stored in deploy's components folder, avoiding CI failures caused by
  GitHub rate limits / transient fetch errors.
- Add a make target that pulls the 2 GIE CRDs + 3 Gateway API CRDs from the  module cache into deploy's components folder.

```release-note
NONE
```

### DIFF
--- a/Makefile.gen.mk
+++ b/Makefile.gen.mk
@@ -1,8 +1,11 @@
 ##@ Code generation targets
 
 # Extract upstream CRD versions from go.mod so references stay in sync with Go dependencies.
+GOMODCACHE ?= $(shell go env GOMODCACHE)
 GIE_VERSION ?= $(shell go list -m -f '{{.Version}}' sigs.k8s.io/gateway-api-inference-extension)
+GIE_MOD := $(GOMODCACHE)/sigs.k8s.io/gateway-api-inference-extension@$(GIE_VERSION)
 GATEWAY_API_VERSION ?= $(shell go list -m -f '{{.Version}}' sigs.k8s.io/gateway-api)
+GATEWAY_API_MOD := $(GOMODCACHE)/sigs.k8s.io/gateway-api@$(GATEWAY_API_VERSION)
 
 ## Location to install dependencies to
 LOCALBIN ?= $(shell pwd)/bin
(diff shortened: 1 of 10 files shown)


shmuelk:
I know we have local copies of the CRDs for Istio. Which can be a pain to update when we upgrade Istio versions.

Do we want this here for K8S Gateway CRDs and the remaining GIE CRDs?

cc @elevran @ahg-g @liu-cong

liu-cong:
I think https://github.com/llm-d/llm-d-router/issues/968 talks about CRDs defined locally (e.g., `ModelRewrite`). For CRDs defined in upstream, I think we should still consume from upstream.

elevran:
@liu-cong - you are correct. However, we have had some failures due to failing to fetch from github in action processing, so it is possible we'd want to handle the upstream as well.  
WDYT?

shmuelk:
@zdtsw Please look at the CoPilot reviews as at least some of them need attention.

@ahg-g @elevran @liu-cong @vMaroon Is something we want? Or do we want to always pull from remote sources the CRDs that are not sourced in our repo? I will point out that we already do this for Istio used in tests. Which is often a bit of a pain to maintain. @zdtsw to her credit has added automation for updating the CRDs here.

liu-cong:
Given @shmuelk 's comment on Istio CRDs, this will bring extra pain to manage. 

> However, we have had some failures due to failing to fetch from github in action processing

@elevran do we know how bad this is? can this be mitigated with some retry?

shmuelk:
@liu-cong We run Istio with slightly modified CRDs, hence the pain in upgrading.

The K8S GW and GIE CRDs we use as is. @zdtsw has added automation, which should help when versions are changed. It is something one has to remember to run....

elevran:
> @elevran do we know how bad this is? can this be mitigated with some retry?

Not sure if retries help. Github as placed rate limits and this has caused some churn and pain (rerunning failed actions), I don't have quantifiable numbers.

liu-cong:
> We run Istio with slightly modified CRDs, hence the pain in upgrading.

Thanks for the clarification! @shmuelk 

> Not sure if retries help. Github as placed rate limits and this has caused some churn and pain (rerunning failed actions), I don't have quantifiable numbers.

Thanks @elevran 

Given the above, not against this approach. Leaving to @elevran to decide.

hexfusion:
@elevran where are we here? should we merge?

elevran:
There is some overhead in maintaining remote crd copies synced locally. Conde wise I think this is ready, but given the extra work in managing, perhaps we should merge this when the next failure shows up?

hexfusion:
I am ok to defer @ahg-g? can we remove from release if this is the case.

re: https://github.com/llm-d/llm-d-router/issues/968

elevran:
Per https://github.com/llm-d/llm-d-router/pull/1482#issuecomment-4692466577 and https://github.com/llm-d/llm-d-router/pull/1482#issuecomment-4692571770 deferring for now and will revisit if/when a fetch failure is detected.

PR #1486, opened 2026-06-04
feat: add request-format agnostic EPP configuration

### LINKED ISSUE #1477: Add a request-format agnostic scheduling configuration

Currently the helm chart ships with a single default configuration. This issue recommends another configuration that is based on the passthrough parser and using request format -agnostic scorers, specifically active requests count and session affinity.

**Why is this needed**:

This is useful in cases where the request format is not known in advance or not supported.

sudoalok:
Closes #1477.

Adds a sample EPP config for cases where the request format isn't known or supported. Instead of the prefix-cache scorer (which needs to read the request body), it uses the passthrough parser plus two scorers that route purely on backend state — active-request and session-affinity.

The passthrough parser is selected explicitly via requestHandler.parsers, otherwise the EPP falls back to the openai-parser default.


ahg-g:
Thanks @sudoalok ; we need to add it as another entry to the configmap of the helm chart, not here: https://github.com/llm-d/llm-d-router/blob/main/config/charts/routerlib/templates/_config.yaml#L8

config/charts/routerlib/templates/_config.yaml
```diff
@@ -80,6 +80,35 @@ data:
...
+    # Select it by setting router.epp.pluginsConfigFile: passthrough-plugins.yaml
+    apiVersion: llm-d.ai/v1alpha1
+    kind: EndpointPickerConfig
+    plugins:
+    - type: passthrough-parser
+    - type: active-request-scorer
+    - type: session-affinity-scorer
+    - type: decode-filter
+    - type: max-score-picker
+    - type: single-profile-handler
```
ahg-g: no need to add the last three, pls remove them.

config/charts/routerlib/templates/_config.yaml
```diff
@@ -80,6 +80,35 @@ data:
       parser:
         pluginRef: {{ $parserType }}
     {{- end }}
+  passthrough-plugins.yaml: |
```
ahg-g: I recommend `payload-agnostic.yaml`

ahg-g:
sorry, one last thing, can you please add it to this test: https://github.com/llm-d/llm-d-router/blob/main/test/integration/epp/wellknown_configs_test.go#L28

sudoalok:
Added the payload-agnostic case to wellknown_configs_test.go

PR #1493, opened 2026-06-05
Fix/inflight token dependency

### LINKED ISSUE #1492: inflight-load-producer: input-token estimate silently reads zero without a token-producer
The in-flight-load producer reads request.Body.TokenizedPrompt for its input-token estimate but declared its dependency through the legacy Consumes() map[string]any, which the data-layer DAG does not read. It is therefore never treated as a ConsumerPlugin, so no token-producer is auto-created or ordered ahead of it. When no other plugin pulls in a token-producer, TokenizedPrompt is nil and EstimateInput returns 0 — the active-request/load scorers then weight on request counts only, with no token signal, silently.

Fix: declare the dependency via plugin.DataDependencies (TokenizedPrompt required). Follow-up to [#1445](https://github.com/llm-d/llm-d-router/pull/1445).

vMaroon:
**What type of PR is this?**
/kind bug

The inflight-load-producer reads request.Body.TokenizedPrompt for its input-token estimate, but declared its dependency through the legacy map-based Consumes(), which the data-layer DAG does not read. It was therefore never treated as a ConsumerPlugin: no token-producer was auto-created or ordered ahead of it, so EstimateInput silently returned 0 whenever nothing else pulled in a token-producer — the active-request/load scorers then weighted on request counts only, with no token signal.

This PR:

- Declares TokenizedPrompt as a typed required dependency, so the data-layer DAG auto-creates and orders a token-producer ahead of this producer.
- Declares PrefixCacheMatchInfo as an optional dependency — the endpoint attribute read for the cached-prefix discount. (Optional deps don't drive DAG ordering or auto-creation; this is declarative, the runtime read keeps its graceful fallback.)
- Adds a prefixMatchInfoProducerName parameter that selects which prefix-cache producer supplies that discount: empty keeps the approximate-prefix producer (default); set it to a precise-prefix-cache producer's instance name to discount against precise cache state instead. It drives both the runtime read and the optional dependency key, mirroring prefix-cache-scorer's field of the same name.

Fixes #1492

```release-note
`inflight-load-producer`: a new `prefixMatchInfoProducerName` parameter selects which prefix-cache producer supplies the cached-prefix discount - the approximate-prefix producer by default, or a precise-prefix-cache producer when set.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
@@ -48,6 +48,11 @@ type Config struct {
 	// AddEstimatedOutputTokens controls whether estimated output tokens are added to
 	// the in-flight token counter. Defaults to false.
 	AddEstimatedOutputTokens bool `json:"addEstimatedOutputTokens"`
+	// PrefixMatchInfoProducerName selects which prefix-cache producer's
+	// PrefixCacheMatchInfo to read for the cached-prefix discount. Empty defaults
+	// to the approximate-prefix producer; set it to a precise-prefix-cache
+	// producer's instance name to discount against precise cache state instead.
+	PrefixMatchInfoProducerName string `json:"prefixMatchInfoProducerName,omitempty"`
 }
 
 func defaultConfig() Config {
@@ -77,6 +82,7 @@ func InFlightLoadProducerFactory(name string, decoder *json.Decoder, handle fwkp
 		tokenEstimator:           NewSimpleTokenEstimator(),
 		addEstimatedOutputTokens: cfg.AddEstimatedOutputTokens,
 		dk:                       attrconcurrency.InFlightLoadDataKey.WithNonEmptyProducerName(name),
+		prefixMatchInfoDK:        attrprefix.PrefixCacheMatchInfoDataKey.WithNonEmptyProducerName(cfg.PrefixMatchInfoProducerName),
 		PluginState:              fwkplugin.NewPluginState(ctx),
 	}, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
```diff
@@ -440,9 +442,18 @@ func (p *InFlightLoadProducer) Produces() map[fwkplugin.DataKey]any {
...
-func (p *InFlightLoadProducer) Consumes() map[string]any {
-	return map[string]any{
-		attrprefix.PrefixCacheMatchInfoDataKey.String(): (*attrprefix.PrefixCacheMatchInfo)(nil),
+// Consumes declares TokenizedPrompt as required so the data-layer DAG orders a
+// token-producer ahead of this producer and auto-creates one when none is
+// configured; without it the input-token estimate silently reads zero.
+//
+// PrefixCacheMatchInfo is read opportunistically in uncachedInputTokens with a
+// graceful fallback and is intentionally not declared here: a missing prefix
+// producer is a valid configuration, not a degraded one.
```
ahg-g: +1; this can be declared as an optional dependency

vMaroon: Done

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
```diff
@@ -440,9 +442,19 @@ func (p *InFlightLoadProducer) Produces() map[fwkplugin.DataKey]any {
...
 }
 
-func (p *InFlightLoadProducer) Consumes() map[string]any {
-	return map[string]any{
-		attrprefix.PrefixCacheMatchInfoDataKey.String(): (*attrprefix.PrefixCacheMatchInfo)(nil),
+// Consumes declares TokenizedPrompt as required so the data-layer DAG orders a
+// token-producer ahead of this producer and auto-creates one when none is
+// configured; without it the input-token estimate silently reads zero.
+// PrefixCacheMatchInfo is optional — used to discount the already-cached prompt
+// prefix when an approximate-prefix producer is present.
```
vMaroon: This will need addressing.

vMaroon: The plugin now accepts setting match info producer.

PR #1494, opened 2026-06-05
fix(preciseprefixcache): honor cache_salt in block hashing

### LINKED ISSUE #1389: Should preciseprefixcache honor cache_salt?
/kind bug (?)

The `cache_salt` field on the OpenAI request body is documented on `InferenceRequestBody.CacheSalt` and the per-API-type sub-fields as "an optional request parameter to isolate prefix caches for security reasons." 
Only the `approximateprefix` data producer honors it and mixes the salt into the block-hash seed.

The `preciseprefixcache` paths build cache keys from `TokenIDs` (or the rendered chat prompt) plus `request.TargetModel` only. The salt is never read or used.
Two requests with identical prompts but different `cache_salt` values would produce distinct prefix-cache keys under the approximate prefix-cache but identifical cache keys under the precise prefix cache.

Is this expected / correct?

cc @vMaroon @liu-cong

vMaroon:
**What type of PR is this?**
/kind bug
/kind feature

The precise producer built block keys from tokens + model only, so cache_salt never isolated prefixes — salted requests collided and failed to correlate with the engine-side index, which already folds vLLM's block-0 cache_salt into the recomputed keys.

Fold TokenizedPrompt.CacheSalt into the first block's extra keys on both producer paths (token-producer and legacy tokenizersPoolConfig wrapper), matching vLLM's placement and the KV-event ingestion path.

Fixes #1389

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/blockkeys.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/blockkeys.go
@@ -33,8 +33,9 @@ type kvCacheIndexer interface {
 
 // computeBlockKeys hashes the request's TokenizedPrompt into KV-block keys,
 // passing any multimodal features into the block-extra-features computation
-// so MM tokens land in the right blocks. Returns (nil, nil) when the request
-// carries no tokens.
+// so MM tokens land in the right blocks. A non-empty CacheSalt is folded into
+// the first block's extra keys so salted prompts get isolated keys. Returns
+// (nil, nil) when the request carries no tokens.
 func computeBlockKeys(ctx context.Context, idx kvCacheIndexer,
 	request *scheduling.InferenceRequest, blockSizeTokens int,
 ) ([]kvblock.BlockHash, error) {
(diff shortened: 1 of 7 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/blockkeys.go
```diff
@@ -51,5 +52,25 @@ func computeBlockKeys(ctx context.Context, idx kvCacheIndexer,
 		extraFeatures = kvblock.ComputeBlockExtraFeatures(
 			mmHashes, mmPlaceholders, blockSizeTokens, len(tp.TokenIDs))
 	}
+	extraFeatures = foldCacheSalt(extraFeatures, tp.CacheSalt, len(tp.TokenIDs)/blockSizeTokens)
 	return idx.ComputeBlockKeysFromTokens(ctx, tp.TokenIDs, request.TargetModel, extraFeatures)
 }
+
+// foldCacheSalt appends the cache salt to the first block's extra keys, after
+// any multimodal hashes. vLLM puts cache_salt in block 0's extra_keys, and
```
ahg-g, line 60: Is this vllm specific? will this work for sglang?

ahg-g (commented):
The tokenizer-based approach in general is lacking documentation on its generality beyond vllm. I think we should address that. The token-producer plugin docs should clearly document its assumptions and applicability across model servers, this practically is part of the epp <-> model server protocol.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/blockkeys.go
vMaroon: router side this works for SGLang, but SGLang needs to emit the salts in the KVEvents similarly to vLLM (within `extra_keys` - they do have this internally). The latter needs a follow-up (cc @sagearc @bongwoobak).

ahg-g: Can we create an issue on the sglang side repo and I can loop in sglang folks to help address it

ahg-g: we also don't have this documented anywhere I believe (in addition to supported request apis, and I am not sure if there are other limitations)

vMaroon: added a section in the plugin's docs, will follow-up on the llm-d repo sometime soon. On SGLang, opened an issue https://github.com/sgl-project/sglang/issues/27682.

ahg-g (approved):

PR #1499, opened 2026-06-05
feat(metrics): add core TTFT and stream-aware TPOT metrics

### LINKED ISSUE #1498: Add core service-level TTFT and stream-aware TPOT metrics

Add two new Prometheus histograms to the core response handler path (pkg/epp/handlers/response.go), always-on with no plugin dependency:

- llm_d_router_epp_request_ttft_seconds: first_token_time - request_received_time
- llm_d_router_epp_request_tpot_seconds: (e2e - TTFT) / (output_tokens - 1)

This requires adding a FirstTokenTimestamp field to RequestContext, captured on the first HandleResponseBody chunk.

***Proposed Acceptance Criteria***
- TTFT histogram is recorded for every request that produces at least one response byte.
- TPOT histogram is recorded for every request with 2+ output tokens.
- Metrics are available on /metrics without enabling the predicted latency plugin.

**Why is this needed**:

***Problem***

Current core metrics only expose normalized_time_per_output_token_seconds, which computes TPOT as total_request_duration / output_token_count. This includes queue wait and prefill time in the numerator, masking actual decode-phase throughput. There is no core TTFT metric at all.

The correct formulas exist in the predicted latency plugin (predictedlatency/requestcontrol_hooks.go), but they're gated behind that plugin being loaded. Operators who don't use the latency predictor have no way to get accurate service-level SLI metrics.

tessapham:
**What type of PR is this?**

/kind feature

Adds two new always-on Prometheus histograms to the core response path:

  - `llm_d_router_epp_request_ttft_seconds` – Time to first token, measured from request received to first response byte
  - `llm_d_router_epp_request_tpot_seconds` – Average time per output token, computed as (e2e - TTFT) / (output_tokens - 1)

These formulas already exist in the predicted latency plugin (predictedlatency/requestcontrol_hooks.go), but are only available when that plugin is loaded. The existing core metric `normalized_time_per_output_token_seconds` divides total request duration (including queue wait and prefill) by raw token count – it doesn't reflect actual decode-phase throughput.

This PR promotes the plugin's stream-aware calculations to core metrics so they're available regardless of plugin configuration, supporting service-level SLI reporting.

Fixes #1498.

Add core service-level TTFT and TPOT metrics (`llm_d_router_epp_service_level_ttft_seconds`, `llm_d_router_epp_service_level_tpot_seconds`) to the response handler, available without enabling the predicted latency plugin. TTFT measures wall-clock time from request received to first response byte. TPOT measures decode-phase throughput as (e2e - TTFT) / (output_tokens - 1). For non-streaming requests, TTFT equals total request duration and TPOT is not recorded.

### DIFF
--- a/pkg/epp/handlers/response.go
+++ b/pkg/epp/handlers/response.go
@@ -79,7 +79,7 @@ func (s *StreamingServer) HandleResponseBody(ctx context.Context, reqCtx *Reques
 		metrics.RecordResponseSizes(reqCtx.IncomingModelName, reqCtx.TargetModelName, reqCtx.ResponseSize)
 		fairnessID := reqCtx.SchedulingRequest.FairnessID
 		priority := strconv.Itoa(reqCtx.Priority)
-		metrics.RecordRequestTTFT(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, fairnessID, priority, reqCtx.RequestReceivedTimestamp, reqCtx.FirstTokenTimestamp)
+		metrics.RecordRequestTTFT(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, fairnessID, priority, reqCtx.modelServerStreaming, reqCtx.RequestReceivedTimestamp, reqCtx.FirstTokenTimestamp)
 		metrics.RecordRequestTPOT(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, fairnessID, priority, reqCtx.RequestReceivedTimestamp, reqCtx.FirstTokenTimestamp, reqCtx.ResponseCompleteTimestamp, reqCtx.Usage.CompletionTokens)
 	}
 	return s.director.HandleResponseBody(ctx, reqCtx, endOfStream)
--- a/pkg/epp/handlers/response.go
+++ b/pkg/epp/handlers/response.go
@@ -18,6 +18,7 @@ package handlers
 
 import (
 	"context"
+	"time"
 
 	configPb "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	extProcPb "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
@@ -49,6 +50,10 @@ func (s *StreamingServer) HandleResponseBody(ctx context.Context, reqCtx *Reques
 
 	reqCtx.ResponseSize += len(responseBytes)
 
+	if reqCtx.FirstTokenTimestamp.IsZero() && len(responseBytes) > 0 {
+		reqCtx.FirstTokenTimestamp = time.Now()
+	}
+
 	var parsedResp *fwkrh.ParsedResponse
 	parser, err := s.getOrResolveParser(ctx, reqCtx)
 	if err != nil {
@@ -71,6 +76,8 @@ func (s *StreamingServer) HandleResponseBody(ctx context.Context, reqCtx *Reques
 		metrics.RecordNormalizedTimePerOutputToken(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, reqCtx.RequestReceivedTimestamp, reqCtx.ResponseCompleteTimestamp, reqCtx.Usage.CompletionTokens)
 		metrics.RecordRequestLatencies(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, reqCtx.RequestReceivedTimestamp, reqCtx.ResponseCompleteTimestamp)
 		metrics.RecordResponseSizes(reqCtx.IncomingModelName, reqCtx.TargetModelName, reqCtx.ResponseSize)
+		metrics.RecordRequestTTFT(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, reqCtx.RequestReceivedTimestamp, reqCtx.FirstTokenTimestamp)
+		metrics.RecordRequestTPOT(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, reqCtx.RequestReceivedTimestamp, reqCtx.FirstTokenTimestamp, reqCtx.ResponseCompleteTimestamp, reqCtx.Usage.CompletionTokens)
 	}
 	return s.director.HandleResponseBody(ctx, reqCtx, endOfStream)
 }
--- a/pkg/epp/handlers/response.go
+++ b/pkg/epp/handlers/response.go
@@ -18,6 +18,7 @@ package handlers
 
 import (
 	"context"
+	"strconv"
 	"time"
 
 	configPb "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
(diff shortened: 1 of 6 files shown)


robertgshaw2-redhat:
what happens if the request does not use streaming?

tessapham:
@robertgshaw2-redhat For non-streaming requests, TTFT is recorded but it's essentially the same as `request_duration_seconds`, and TPOT should be skipped. TPOT can't be calculated for non-streaming anyway since there's no token-by-token decode phase to measure. I can note this restriction in the release notes.

pkg/epp/handlers/response.go
```diff
@@ -71,6 +76,8 @@ func (s *StreamingServer) HandleResponseBody(ctx context.Context, reqCtx *Reques
 		metrics.RecordNormalizedTimePerOutputToken(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, reqCtx.RequestReceivedTimestamp, reqCtx.ResponseCompleteTimestamp, reqCtx.Usage.CompletionTokens)
 		metrics.RecordRequestLatencies(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, reqCtx.RequestReceivedTimestamp, reqCtx.ResponseCompleteTimestamp)
 		metrics.RecordResponseSizes(reqCtx.IncomingModelName, reqCtx.TargetModelName, reqCtx.ResponseSize)
+		metrics.RecordRequestTTFT(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, reqCtx.RequestReceivedTimestamp, reqCtx.FirstTokenTimestamp)
```
hexfusion: nit: since this PR is adding the metric anyway, worth landing fairness_id and objective as labels now? both are resolved for flow control by the time we Observe(). this would enable per-tier SLO queries like Rob raised p95 TTFT/TPOT per fairness_id × objective.

pkg/epp/metrics/llm_d_router_metrics.go
```diff
@@ -137,6 +137,34 @@ var (
...
+
+	llmdRequestTTFT = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: LLMDRouterEndpointPickerSubsystem,
+			Name:      "service_level_ttft_seconds",
+			Help:      metricsutil.HelpMsgWithStability("Service-level time to first token in seconds, measured from request received to first response byte.", compbasemetrics.ALPHA),
+			Buckets: []float64{
+				0.005, 0.025, 0.05, 0.1, 0.2, 0.4, 0.6, 0.8, 1.0, 1.25, 1.5, 2, 3, 4, 5, 6,
+				8, 10, 15, 20, 30, 45, 60, 120, 180, 240, 300, 360, 480, 600, 900, 1200,
+				1800, 2700, 3600,
```
hexfusion: nit: 35 buckets is pretty dense if we collapse the long tail above 60s it seems like it would cut a lot of series without losing alertable signal. lets just review this to make sure this is needed.  the leaves more headroom for future labels as well.

pkg/epp/metrics/llm_d_router_metrics.go
```diff
@@ -137,6 +137,34 @@ var (
 		},
 		modelLabels,
 	)
+
+	llmdRequestTTFT = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: LLMDRouterEndpointPickerSubsystem,
+			Name:      "service_level_ttft_seconds",
```
hexfusion: couple ways to harden the streaming-only assumption:

- rename to service_level_streaming_ttft_seconds / _streaming_tpot_seconds. makes the constraint discoverable, prevents PromQL footguns aggregating both modes.
- bubble up vllm source of truth via response headers RFE #36189 (https://github.com/vllm-project/vllm/issues/36189) (PR #36383). works for non-streaming too.
  
this matters more now if this is always-on (no longer gated by the latency predictor) operators might not reason about the streaming caveat unless it's in the name.
  
ref. https://github.com/vllm-project/vllm/blob/fa27d4e9cf3c8d8a5a143f38c346b27c02b2c2e3/docs/design/metrics.md?plain=1#L53

pkg/epp/handlers/response.go
tessapham: Good point. I'll wait for https://github.com/llm-d/llm-d-router/pull/1518 to land as it'll add `fairnessID` and `priority` to the response handler, which I can pull into these new metrics.

pkg/epp/metrics/llm_d_router_metrics.go
tessapham: Sounds good I can keep the buckets up to 60. Maybe keep 120 or 180 as well for any extreme outliers?

tessapham: Made 120s the ceiling for TTFT and 2s for TPOT.

pkg/epp/metrics/llm_d_router_metrics.go
tessapham: This is smart, thanks @hexfusion for the suggestions!

- I'll rename to `_streaming_tpot_seconds`. TTFT maybe we can keep as is since it can still be calculated for non-streaming requests.
- Right we can extract from the response headers once this feature lands in vLLM. Worth noting tho that this is server-side TTFT so it won't include queue/scheduling times (it can be a different TTFT metric), but definitely useful for non-streaming reqs.

pkg/epp/metrics/llm_d_router_metrics.go
```diff
@@ -137,6 +137,32 @@ var (
...
+			Help:      metricsutil.HelpMsgWithStability("Service-level time to first token in seconds, measured from request received to first response byte. For non-streaming requests, this equals total request duration.", compbasemetrics.ALPHA),
+			Buckets: []float64{
+				0.005, 0.025, 0.05, 0.1, 0.2, 0.4, 0.6, 0.8, 1.0, 1.25, 1.5, 2, 3, 4, 5, 6,
+				8, 10, 15, 20, 30, 45, 60, 120,
+			},
+		},
+		modelLabels,
```
ahg-g: Should we add a label to distinguish between streaming vs non-streaming requests? Also, add the fairness and priority labels as well.

pkg/epp/metrics/llm_d_router_metrics.go
```diff
@@ -137,6 +137,32 @@ var (
...
+		prometheus.HistogramOpts{
+			Subsystem: LLMDRouterEndpointPickerSubsystem,
+			Name:      "service_level_streaming_tpot_seconds",
+			Help:      metricsutil.HelpMsgWithStability("Service-level average time per output token in seconds for streaming requests, computed as (e2e - TTFT) / (output_tokens - 1).", compbasemetrics.ALPHA),
+			Buckets: []float64{
+				0.0005, 0.00205, 0.005, 0.01, 0.02, 0.04, 0.06, 0.08, 0.1, 0.125, 0.15, 0.2,
+				0.3, 0.4, 0.5, 0.6, 0.8, 1, 2,
+			},
+		},
+		modelLabels,
```
ahg-g: ditto

ahg-g:
@tessapham can you pls address the comments so we move forward with this PR. Please add a streaming label as well as the priority and fairness id

tessapham:
@ahg-g Addressed your comments!

pkg/epp/metrics/llm_d_router_metrics.go
tessapham: Done. `streaming` label is only added to TTFT, for TPOT it's streaming-only by default.

tessapham:
@ahg-g CI passed now.

pkg/epp/metrics/llm_d_router_metrics.go
```diff
@@ -137,6 +137,32 @@ var (
 		},
 		modelLabels,
 	)
+
+	llmdRequestTTFT = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: LLMDRouterEndpointPickerSubsystem,
+			Name:      "service_level_ttft_seconds",
```
ahg-g: sorry, just noticed this, what does service_level mean here? what does it suppose to indicate?

tessapham: It's meant to distinguish these metrics from the plugin-level TTFT/TPOT that already exist in the predicted latency plugin (`request_ttft_seconds`, `request_tpot_seconds`). `service_level` means measured from the client's perspective (request arrival to first byte), as opposed to the plugin's metrics which start after scheduling.

Lmk if you'd prefer a different prefix tho.

tessapham:
@ahg-g Good to move forward?

pkg/epp/metrics/llm_d_router_metrics.go
tessapham: Another option is not have a prefix altogether, so these new metrics will be `ttft_seconds` and `streaming_tpot_seconds`.

hexfusion: unprefixed reads as the canonical from my perspective +1

ahg-g: agree, the service_level prefix is confusing in my opinion

tessapham: Removed `service_level` prefix from metric names.

hexfusion:
LGTM!

ahg-g (approved):

PR #1513, opened 2026-06-06
pkg/common/observability/tracing: consolidate tracing initialization

### LINKED ISSUE #1503: [observability] Unify tracing init in router (merge pkg/telemetry into common/observability)
router currently has two tracing initialization paths with inconsistent behavior. `cmd/epp/main.go` calls `telemetry.InitTracing` from `pkg/telemetry` unconditionally, which defaults to an OTLP exporter, so tracing tries to connect to an OTLP collector even when the operator did not intend to enable tracing. 

Meanwhile the `runner --tracing` path calls `InitTracing` from `pkg/common/observability/tracing`, whose default exporter is console. Having two init paths and two different default exporters in the same process is confusing to configure and makes it unclear which one is in effect.

A complication: `pkg/telemetry` is not just init, it also exposes `Tracer()`, which is used across EPP scheduling plugins, pd-sidecar connectors, and the precise-prefix-cache producer, so it cannot simply be deleted but needs to be consolidate to `pkg/common/observability`.

LukeAVanDrie:
**What type of PR is this?**

/kind bug
/kind cleanup

EPP and pd-sidecar unconditionally initialized OTLP tracing, causing unwanted connection attempts and hangs when tracing was disabled. This PR consolidates tracing to `pkg/common/observability/tracing` (removing the redundant `pkg/telemetry` package) and makes initialization conditional on the `--tracing` flag in both EPP (defaults to `true`) and pd-sidecar (defaults to `false`).

_Note:_ By consolidating to `pkg/common/observability/tracing`, the default exporter changes from otlp (in `pkg/telemetry`) to console (in `pkg/common`). Since EPP defaults to --tracing=true, EPP running without `OTEL_TRACES_EXPORTER` environment variable set will now print all traces to stdout using the pretty printer, which might significantly clutter the logs. Helm deployments are unaffected as they inject `OTEL_TRACES_EXPORTER=otlp`.

Fixes #1503

**Release note**:

```release-note
Consolidated tracing initialization and tracer retrieval. Added a `--tracing` flag to `pd-sidecar` (defaulting to `false`) to allow conditionally enabling tracing and avoiding unwanted OTLP connection attempts by default.
```


pkg/common/observability/tracing/telemetry.go
```diff
@@ -138,3 +129,10 @@ func initTraceExporter(ctx context.Context, logger logr.Logger) (sdktrace.SpanEx
 
 	return traceExporter, nil
 }
+
+const instrumentationName = "llm-d-router"
+
+// Tracer returns a tracer for the llm-d router.
+func Tracer() trace.Tracer {
+	return otel.Tracer(instrumentationName)
+}
```
shmuelk, line 138: In the sidecar, which also uses this code, do we want the instrumentationName to be llm-d-router? Or do we want it to be somethings else (perhaps llm-d-epd-sidecar) ?

shmuelk:
>  Note: By consolidating to pkg/common/observability/tracing, the default exporter changes from otlp (in pkg/telemetry) to console (in pkg/common). Since EPP defaults to --tracing=true, EPP running without OTEL_TRACES_EXPORTER environment variable set will now print all traces to stdout using the pretty printer, which might significantly clutter the logs. Helm deployments are unaffected as they inject OTEL_TRACES_EXPORTER=otlp.

The various deployments of the EPP in this repo should be changed to add `OTEL_TRACES_EXPORTER=otlp` to eliminate the log clutter mentioned.

pkg/common/observability/tracing/telemetry.go
ahg-g: We have three levels, service name, instrument name and span name. We use llm-d as a prefix in all three, we also use llm-d router as both the default service name and instrument name. I think we can do better than that, I opened https://github.com/llm-d/llm-d-router/issues/1519 to discuss and define a standard.

LukeAVanDrie:
> The various deployments of the EPP in this repo should be changed to add `OTEL_TRACES_EXPORTER=otlp` to eliminate the log clutter mentioned.

Do we want to do this atomically as part of this PR?

pkg/common/observability/tracing/telemetry.go
LukeAVanDrie: Do we want to tackle #1519 in this PR? Doesn't seem to bad to implement your suggestions there, but it it needs further community discussion, maybe we get #1503 resolved first?

PR #1514, opened 2026-06-07
fix(epp): extract upstream traceparent and re-parent server span

### LINKED ISSUE #1502: [Observability] EPP: extract upstream traceparent at ext_proc ingress and re-parent the request span

- In the `RequestHeaders` branch of `Process()`, extract the upstream trace context from the Envoy header map using `propagator.Extract`.
- Move the server span start out of the top of `Process()` to after extraction, or re-parent it to the extracted context. Extracting without re-parenting has no effect because the orphan root span has already been started.
- Keep the existing downstream Inject behavior unchanged.

**Why is this needed**:

EPP runs as an Envoy `ext_proc` filter. On every request, Envoy forwards the incoming headers (including the upstream `traceparent` set by the client or the Gateway). However, EPP starts its server span at the very top of `Process()`, before the `RequestHeaders` message is received. Because the span is started before the headers are available, the upstream trace context is never extracted, so EPP begins a brand new root trace for every request. Result: the Gateway/Envoy trace and the EPP trace are two disconnected traces in Jaeger, and operators cannot follow a single end-to-end trace through the request path. Note: EPP already injects trace context downstream in `generateHeaders`, this issue only covers the missing inbound extraction.

gyliu513:
**What type of PR is this?**

/kind bug

EPP runs as an Envoy `ext_proc` filter. On every request Envoy forwards the incoming headers, including the upstream `traceparent` set by the client or the Gateway. However, EPP started its `gateway.request` server span at the very top of `Process()`, before the `RequestHeaders` message was received. Because the span was started before the headers were available, the upstream trace context was never extracted, so EPP began a brand new root trace for every request. The result was two disconnected traces in Jaeger (Gateway/Envoy and EPP), and operators could not follow a single end-to-end trace through the request path.

This PR fixed the issue as below:

- Defers the server span start out of the top of `Process()` to the `RequestHeaders` branch, where the incoming headers are first available.
- Extracts the upstream trace context from the Envoy header map using the globally configured propagator (`extractTraceContext`), then starts the span re-parented to it, so the EPP trace joins the caller's trace.
- Leaves the existing downstream `Inject` behavior in `generateHeaders` unchanged; this issue only covered the missing inbound extraction.

Fixes #1502 

```release-note
NONE
```

### DIFF
--- a/pkg/epp/handlers/server.go
+++ b/pkg/epp/handlers/server.go
@@ -31,6 +31,7 @@ import (
 	"github.com/google/uuid"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/trace"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -197,6 +198,24 @@ func (s *StreamingServer) getOrResolveParser(ctx context.Context, reqCtx *Reques
 	return parser, nil
 }
 
+// extractTraceContext returns ctx augmented with the upstream trace context
+// carried in the incoming Envoy request headers (e.g. the traceparent set by the
+// client or the Gateway), using the globally configured text map propagator.
+//
+// The header wire format is the W3C Trace Context spec:
+// https://www.w3.org/TR/trace-context/
+// Extraction uses OpenTelemetry context propagation:
+// https://opentelemetry.io/docs/concepts/context-propagation/
+func extractTraceContext(ctx context.Context, req *extProcPb.ProcessingRequest_RequestHeaders) context.Context {
+	carrier := make(propagation.MapCarrier)
+	if req != nil && req.RequestHeaders != nil && req.RequestHeaders.Headers != nil {
+		for _, header := range req.RequestHeaders.Headers.Headers {
+			carrier[strings.ToLower(header.Key)] = envoy.GetHeaderValue(header)
+		}
+	}
+	return otel.GetTextMapPropagator().Extract(ctx, carrier)
+}
+
 func (s *StreamingServer) Process(srv extProcPb.ExternalProcessor_ProcessServer) error {
 	ctx := srv.Context()
 
@@ -208,8 +227,15 @@ func (s *StreamingServer) Process(srv extProcPb.ExternalProcessor_ProcessServer)
 			attribute.String("commit-sha", version.CommitSHA),
 		),
 	)
-	ctx, span := tracer.Start(ctx, "gateway.request", trace.WithSpanKind(trace.SpanKindServer))
-	defer span.End()
+	// The server span is started in the RequestHeaders branch, once the upstream
+	// trace context carried in the incoming headers is available, so the EPP span
+	// joins the caller's trace instead of starting a disconnected root.
+	var span trace.Span
+	defer func() {
+		if span != nil {
+			span.End()
+		}
+	}()
 
 	logger := log.FromContext(ctx)
 	loggerTrace := logger.V(logutil.TRACE)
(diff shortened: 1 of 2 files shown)


hexfusion:
LGTM, thanks @gyliu513

pkg/epp/handlers/server.go
```diff
@@ -350,6 +371,13 @@ func (s *StreamingServer) Process(srv extProcPb.ExternalProcessor_ProcessServer)
 			loggerTrace = logger.V(logutil.TRACE)
 			ctx = log.IntoContext(ctx, logger)
 
+			// Re-parent the server span to the upstream trace context (e.g. the
+			// traceparent set by the client or the Gateway) carried in the incoming
+			// headers, then start it. The headers are only available here, so the span
+			// cannot be started at the top of Process without orphaning the trace.
+			ctx = extractTraceContext(ctx, v)
+			ctx, span = tracer.Start(ctx, "gateway.request", trace.WithSpanKind(trace.SpanKindServer))
```
ahg-g, line 384: Not related to this PR, but what does `gateway.request` mean? EPP is not necessarily hooked up into a gateway.

pkg/epp/handlers/server.go
```diff
@@ -197,6 +198,19 @@ func (s *StreamingServer) getOrResolveParser(ctx context.Context, reqCtx *Reques
 	return parser, nil
 }
 
+// extractTraceContext returns ctx augmented with the upstream trace context
+// carried in the incoming Envoy request headers (e.g. the traceparent set by the
+// client or the Gateway), using the globally configured text map propagator.
```
ahg-g, line 203: Can you link to any official docs on this protocol?

pkg/epp/handlers/server.go
ahg-g, line 223: I see `.Tracer`  invoked in more than one place in the epp code, is that intended (see https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go#L226)? Shouldn't we have a single tracer for each request?

pkg/epp/handlers/server.go
gyliu513: Agreed, and good catch. The `gateway.request` name predates this PR, this change only moves where the span is started so it re-parents correctly, it doesn't touch the name. You're right that it's a misnomer since EPP is an Envoy ext_proc filter and isn't necessarily behind a Gateway. I'd rather not rename it here to keep this PR scoped to the trace-context fix; I'll file a separate issue to rename it (e.g. epp.extproc.request) and update any dashboards/queries that key off it. Sound good?

pkg/epp/handlers/server.go
gyliu513: Sure. The header format is the W3C Trace Context spec (traceparent/tracestate): https://www.w3.org/TR/trace-context/. The extraction here uses OpenTelemetry's TextMapPropagator: https://opentelemetry.io/docs/concepts/context-propagation/ and the Go API https://pkg.go.dev/go.opentelemetry.io/otel/propagation. I'll add the W3C link to the extractTraceContext doc comment.

pkg/epp/handlers/server.go
gyliu513: Good catch, this is a real, pre-existing inconsistency, not something this PR introduced (the otel.Tracer(...) instantiation in server.go predates this change; this PR only moved where the span is started).

One clarification on the `single tracer per request` point: in OpenTelemetry a Tracer is keyed by instrumentation scope (name + version), not by request. A request legitimately flows through several scopes, the extproc handler, the director, the scheduling plugins, and each obtaining its own named tracer is the intended model; tracers are cheap to acquire. So the goal isn't one tracer per request, but consistent scope naming and metadata across them.

The actual problem you spotted is that we had three different ways of creating tracers: the shared tracing.Tracer() helper (scope llm-d-router, no version), server.go hand-rolling llm-d-router/epp/extproc with build ref + commit SHA, and director.go hand-rolling llm-d-router with no version. So within a single trace some spans carried build/version metadata and others didn't.

To keep this PR scoped to the trace-context fix, I split the cleanup out into #1538 / #1539. That routes all tracer creation through tracing.Tracer(): it now takes an optional scope name (defaulting to llm-d-router) and always attaches version.BuildRef and the commit-sha attribute. server.go keeps its per-component scope name via tracing.Tracer("llm-d-router/epp/extproc"), and director.go drops the duplicated literal in favor of
tracing.Tracer().

PR #1529, opened 2026-06-08
fix(epp/preciseprefixcache): reject estimate-backed token-producer

### LINKED ISSUE #1471: Precise prefix-cache: reject zero-config estimate fallback
_Claude Assisted_

Spun out of #1445 (Gap 2). The estimate token-producer backend is the zero-config default and is auto-created for any `TokenizedPrompt` consumer. A precise prefix-cache scorer/producer shares that data key, so a precise config without an explicit `vllm` token-producer is silently satisfied with non-engine tokens instead of failing loudly (the pre-#1445 behavior).

Proposal: a startup-time check that rejects a precise prefix-cache scorer/producer when no `vllm` token-producer is configured, restoring the hard failure. The token-producer README documents the risk.

cc @elevran

gkneighb:
## Summary

Fixes #1471. The precise-prefix-cache producer hashes engine-aligned token IDs into KV-block keys that must match the model server's prefix cache. The token-producer's estimate backend (the zero-config default introduced in #1445 and registered as the framework's `TokenizedPromptDataKey` default at `cmd/epp/runner/runner.go:520`) emits byte-packed pseudo-tokens that silently corrupt KV-block lookups against the real engine while leaving every health check green.

The framework's `CreateMissingDataProducers` auto-default fires *after* explicit factories, so the precise-prefix-cache factory needs only inspect `handle` for an already-loaded token-producer with a real (vllm or UDS) backend. Both "no token-producer at all" (auto-default about to install estimate) and "estimate-backed token-producer explicitly configured" fail with the same actionable message; the token-producer must come earlier in the plugins list than the precise-prefix-cache-producer.

Adds `IsRealTokenizer()` on the token-producer Plugin to expose backend type without leaking the unexported `renderBackend` struct.

## Scope notes

- All in-tree configs (`deploy/config/epp-precise-prefix-cache{,split-}-config.yaml`, e2e `kvConfig` / `kvExternalTokenizerConfig`) already declare `token-producer` before precise — zero existing-config regressions.
- The deprecated `precise-prefix-cache-scorer` self-host path (`newLegacyProducer`) bypasses this factory and is not covered by the validation. No in-tree config exercises it without an explicit token-producer; flagging for awareness rather than fixing in the same PR since the scorer factory already logs a `DEPRECATION` notice.

## Test plan
- [x] `TestPluginFactory_RejectsMissingRealTokenProducer` — `(a)` no token-producer in handle, `(b)` estimate-backed token-producer explicitly configured
- [x] `TestE2EConfigs_StartupParseSmoke/preciseConfigWithVLLMTokenProducer` — end-to-end through `NewTestHarness` with a valid `token-producer (vllm) -> precise-prefix-cache-producer -> prefix-cache-scorer` pipeline
- [x] `make test-unit-epp`
- [x] `make test-integration-hermetic` (full suite, ~75s)
- [x] `make presubmit`
- [ ] Reviewer to confirm "factory-time, config-order-dependent" check is acceptable vs. a post-`CreateMissingDataProducers` validation hook (alternative: move `requireRealTokenProducer` into a runner-level pass after auto-defaults complete; ~30-line refactor)

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
@@ -143,6 +143,9 @@ func PluginFactory(name string, rawParameters *json.Decoder, handle plugin.Handl
 	if parameters.IndexerConfig.TokenizersPoolConfig != nil {
 		return nil, errors.New("tokenizersPoolConfig is not supported; configure a token-producer plugin instead")
 	}
+	if err := requireRealTokenProducer(handle); err != nil {
+		return nil, err
+	}
 
 	p, err := New(handle.Context(), name, parameters)
 	if err != nil {
@@ -150,6 +153,35 @@ func PluginFactory(name string, rawParameters *json.Decoder, handle plugin.Handl
 	}
 
 	return p, nil
+}
+
+// requireRealTokenProducer verifies that an engine-aligned token-producer
+// (vllm/uds backend) is already loaded in handle when this factory runs. The
+// estimate backend emits byte-packed pseudo-tokens, which would silently
+// corrupt the KV-block hashes this producer feeds to the indexer. The
+// framework's auto-default for the TokenizedPrompt data key fires *after*
+// explicit factories, so absence of any token-producer here also means an
+// estimate-backed one is about to be auto-created. Both cases fail with the
+// same actionable message; the token-producer must come earlier in the YAML
+// than the precise-prefix-cache-producer. See #1471.
+func requireRealTokenProducer(handle plugin.Handle) error {
+	for _, p := range handle.GetAllPlugins() {
+		typeName := p.TypedName().Type
+		//nolint:staticcheck // SA1019: intentionally accept the deprecated 'tokenizer' alias so existing configs keep working.
+		if typeName != tokenproducer.PluginType && typeName != tokenproducer.LegacyPluginType {
+			continue
+		}
+		tp, ok := p.(*tokenproducer.Plugin)
+		if !ok {
+			continue
+		}
+		if tp.BackendKind() == tokenproducer.KindRender {
+			return nil
+		}
+	}
+	return fmt.Errorf(
+		"%s requires an engine-aligned token-producer (vllm or udsTokenizerConfig backend) "+
+			"configured earlier in the plugins list", PluginType)
 }
 
 // New constructs a precise-prefix-cache-producer. The instance name becomes
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
@@ -143,9 +143,6 @@ func PluginFactory(name string, rawParameters *json.Decoder, handle plugin.Handl
 	if parameters.IndexerConfig.TokenizersPoolConfig != nil {
 		return nil, errors.New("tokenizersPoolConfig is not supported; configure a token-producer plugin instead")
 	}
-	if err := requireRealTokenProducer(handle); err != nil {
-		return nil, err
-	}
 
 	p, err := New(handle.Context(), name, parameters)
 	if err != nil {
@@ -155,16 +152,30 @@ func PluginFactory(name string, rawParameters *json.Decoder, handle plugin.Handl
 	return p, nil
 }
 
-// requireRealTokenProducer verifies that an engine-aligned token-producer
-// (vllm/uds backend) is already loaded in handle when this factory runs. The
-// estimate backend emits byte-packed pseudo-tokens, which would silently
-// corrupt the KV-block hashes this producer feeds to the indexer. The
-// framework's auto-default for the TokenizedPrompt data key fires *after*
-// explicit factories, so absence of any token-producer here also means an
-// estimate-backed one is about to be auto-created. Both cases fail with the
-// same actionable message; the token-producer must come earlier in the YAML
-// than the precise-prefix-cache-producer. See #1471.
-func requireRealTokenProducer(handle plugin.Handle) error {
+// ValidateTokenProducer verifies that when a precise-prefix-cache-producer is
+// configured, an engine-aligned token-producer (vllm/uds backend) is present in
+// the handle. The estimate backend emits byte-packed pseudo-tokens, which would
+// silently corrupt the KV-block hashes this producer feeds to the indexer.
+//
+// This runs at the runner level after CreateMissingDataProducers, not inside the
+// factory: factory instantiation order is decided by the plugin DAG's
+// topological sort (which seeds from a map and so varies per process start), not
+// by the plugins-list order, so a factory-time check cannot rely on the
+// token-producer being constructed first. By the time this runs, every plugin --
+// explicit and auto-defaulted (an absent token-producer is auto-created as the
+// estimate backend) -- is in the handle, so the check is order-independent. See #1471.
+func ValidateTokenProducer(handle plugin.Handle) error {
+	hasPrecise := false
+	for _, p := range handle.GetAllPlugins() {
+		if p.TypedName().Type == PluginType {
+			hasPrecise = true
+			break
+		}
+	}
+	if !hasPrecise {
+		return nil
+	}
+
 	for _, p := range handle.GetAllPlugins() {
 		typeName := p.TypedName().Type
 		//nolint:staticcheck // SA1019: intentionally accept the deprecated 'tokenizer' alias so existing configs keep working.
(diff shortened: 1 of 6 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -134,6 +137,38 @@ func PluginFactory(name string, rawParameters *json.Decoder, handle plugin.Handl
...
+		}
+		tp, ok := p.(*tokenproducer.Plugin)
+		if !ok {
+			continue
+		}
+		if tp.IsRealTokenizer() {
+			return nil
+		}
+	}
+	return fmt.Errorf(
```
hexfusion, line 182: nit: this feels overly verbose could we tighten this up just a bit, perhaps something like
```go
     return fmt.Errorf(
      "%s requires an engine-aligned token-producer (vllm or udsTokenizerConfig backend) "+
          "configured earlier in the plugins list", PluginType)
```

gkneighb: Done — agreed, the "why" is already in the function docstring and #1471, no need to carry it in the operator-facing error. Tightened to your suggestion.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -226,6 +226,16 @@ func (p *Plugin) TypedName() plugin.TypedName {
 	return p.typedName
 }
 
+// IsRealTokenizer reports whether the plugin is backed by a real tokenizer
+// (vllm /render or the deprecated UDS gRPC) rather than the byte-packing
+// estimate fallback. Downstream consumers that need engine-aligned token IDs
+// (e.g. precise-prefix-cache, whose KV-block hashes must match the model
+// server) check this at startup to refuse non-engine tokens.
+func (p *Plugin) IsRealTokenizer() bool {
```
vMaroon: I think we want to shift away from the concept of "real" vs "estimate" - maybe backends can have names (in code, not settable by user) that consumers can get.

gkneighb: Refactored — `Kind` typed string (`KindRender`, `KindEstimate`) + `BackendKind() Kind` on the token-producer Plugin, with a type-switch over the existing backend structs so consumers whitelist accepted backends by name instead of a binary predicate. Updated the precise-prefix-cache check to accept `KindRender` explicitly. PTAL.

gkneighb:
@vMaroon — refactored per your suggestion: backends now expose a typed `Kind` (`KindRender`/`KindEstimate`) via `BackendKind()` rather than a "real vs estimate" boolean, so consumers whitelist accepted kinds. Let me know if the shape matches what you had in mind.

gkneighb:
@vMaroon -- following up on this. The refactor moves to a typed `Kind` (`KindRender`/`KindEstimate` via `BackendKind()`) per your "move away from real vs. estimate" point, and CI is green. Does the shape work for you, or anything you'd want changed? Happy to adjust.

sagearc (commented):
this looks good to me in general with one open question on the unticked bullet, why did you go for the "factory-time config-order-dependent" check instead of the post-validation one?

gkneighb:
@sagearc Good question -- that bullet was exactly the judgment call I wanted a reviewer's read on. Three reasons factory-time won:

1. **It keeps a plugin-local invariant in the plugin.** "My `TokenizedPrompt` must be engine-aligned" is precise-prefix-cache's requirement, not the framework's, and the factory check lives entirely in the plugin package. The post-`CreateMissingDataProducers` variant needs a home: either `runner.go` hardcodes precise-prefix's requirement (plugin knowledge leaking into the runner), or the framework grows a generic post-load validation hook -- and no such hook exists today (closest are per-plugin config `validate()` helpers), so it would be new framework API riding along on a bugfix.

2. **The order-dependence can only fail loud, never wrong.** Request-time producer order doesn't come from YAML order -- `ValidateAndOrderDataDependencies` (runner.go:666) topologically sorts producers by the produce/consume DAG after auto-defaults complete. So the only cost of checking at factory time is a *startup false-rejection* when a real token-producer is declared after precise in the YAML, and that error states the fix (reorder). The dangerous direction -- accepting an estimate-backed pipeline -- can't happen: absence at factory time implies the estimate default is exactly what `CreateMissingDataProducers` would install afterward, which is the case being rejected.

3. **Every in-tree config already satisfies the ordering** (deploy/config/*, e2e fixtures declare token-producer first), so nothing regresses.

That said, the runner-level pass is strictly more permissive -- it would accept the real-producer-after-precise orderings the factory check spuriously rejects -- and it's a contained ~30-line change. If you or the OWNERS prefer that shape I'm happy to switch in this PR; I just didn't want to grow framework surface on a fix without a maintainer asking for it.

sagearc (dismissed):
@gkneighb thanks, the locality justification makes it a good call.
LGTM

gkneighb:
Thanks @sagearc!

For the remaining CODEOWNERS gate: this PR spans two owned paths. `tokenizer/` is now covered by sagearc's approval; `preciseprefixcache/` still needs an owner. @bongwoobak could you take a look at that side? It's a single startup-time factory check (`requireRealTokenProducer`) plus tests -- sagearc's review already settled the design question of where the check lives, and CI is green with no conflicts. (@vMaroon covered both paths if you'd rather take it.)

ahg-g:
/assign @vMaroon 

Can you take a look?

gkneighb:
Friendly nudge on this one -- it's been approved by @sagearc and is green/mergeable, just needs a `preciseprefixcache/` codeowner pass to unblock. @vMaroon (or @liu-cong / @shmuelk) when you have a few minutes? It's a single startup-time factory check plus tests; happy to walk through anything.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -143,6 +143,9 @@ func PluginFactory(name string, rawParameters *json.Decoder, handle plugin.Handl
 	if parameters.IndexerConfig.TokenizersPoolConfig != nil {
 		return nil, errors.New("tokenizersPoolConfig is not supported; configure a token-producer plugin instead")
 	}
+	if err := requireRealTokenProducer(handle); err != nil {
```
bongwoobak, line 146: The plugins list order doesn't decide instantiation order, so this can't hold where it sits. `instantiatePlugins` sorts through `buildPluginDAG` (configloader.go:230), token-producer and precise have no edge between them, and `TopologicalSort` seeds its queue by ranging a map (utils.go:42), so which factory runs first varies per process start. `TestE2EConfigs_StartupParseSmoke/preciseConfigWithVLLMTokenProducer` is red on the current CI run with this exact error, and that fixture lists token-producer first.

The post-`CreateMissingDataProducers` pass you floated in the test plan is the one that works. Everything is in the handle by then (runner.go:731), so it doesn't depend on order.

gkneighb: Fixed in the latest commit. You're right that instantiation order is DAG/map-driven, not list-order -- I confirmed the `preciseConfigWithVLLMTokenProducer` smoke test was flaking exactly as you described. Moved the check to `ValidateTokenProducer`, called from the runner right after `CreateMissingDataProducers` (runner.go), so it runs once with every plugin -- explicit and auto-defaulted -- already in the handle, independent of factory order. Smoke test now passes deterministically in the hermetic suite; the factory is back to construction-only.

Credit where due: @sagearc raised this exact concern earlier in the thread (factory-time, config-order-dependent vs. a post-validation pass) and I argued for keeping it in the factory. That was wrong -- this is the post-`CreateMissingDataProducers` pass he pointed to.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -152,6 +152,49 @@ func PluginFactory(name string, rawParameters *json.Decoder, handle plugin.Handl
...
+// factory: factory instantiation order is decided by the plugin DAG's
+// topological sort (which seeds from a map and so varies per process start), not
+// by the plugins-list order, so a factory-time check cannot rely on the
+// token-producer being constructed first. By the time this runs, every plugin --
+// explicit and auto-defaulted (an absent token-producer is auto-created as the
+// estimate backend) -- is in the handle, so the check is order-independent. See #1471.
+func ValidateTokenProducer(handle plugin.Handle) error {
+	hasPrecise := false
+	for _, p := range handle.GetAllPlugins() {
+		if p.TypedName().Type == PluginType {
```
bongwoobak, line 170: This only matches `precise-prefix-cache-producer`. The self-hosted `precise-prefix-cache-scorer` builds a Producer internally but registers under its own type (precise_prefix_cache.go:136), so `hasPrecise` stays false and the check never runs for it. Reproduced on e63a90bb with the runner sequence, RegisterAsDefaultProducer then CreateMissingDataProducers then ValidateTokenProducer: a handle holding only the self-hosted scorer gets the estimate token-producer auto-created and ValidateTokenProducer returns nil. The same handle with an explicit producer errors as expected.

The scope note leaves this path out because it had its own tokenizer pool. That pool is gone since #2512, `tokenizersPoolConfig` is rejected at startup now, so a self-hosted scorer config without an explicit token-producer lands on the estimate default, and that's the group most likely to hit this.

The scorer package imports this one, so the type can't be referenced from here. A marker interface in this package avoids the cycle: `type Host interface{ PreciseProducer() *Producer }`, the scorer returns its embedded producer, and this loop checks `p.(Host)` next to the type match. Tried it on the branch, the self-hosted case flips to the error and both packages stay green otherwise. Checking the scorer type in runner.go, which already imports both, works too.

Issue #1483, opened 2026-06-04
[Observability] add fine-grained OpenTelemetry spans for scheduling/scoring pipeline


gyliu513:
**What would you like to be added**:

Fine-grained OpenTelemetry spans for the EPP scheduling/scoring pipeline so operators can see *which scorers ran, how long they took, and aggregate score signals* inside a single request trace, not only top-level `gateway.request` / `gateway.request_orchestration`.

Concrete deliverables:

1. **Parent span for scheduling**
   - Add a child span under `gateway.request_orchestration` (e.g. `llm_d.epp.schedule` or `llm_d.epp.scoring`) that wraps the scheduler path: filters → scorer chain → picker.
   - Implement in `pkg/epp/scheduling` (e.g. `Scheduler.Schedule` / `runScorerPlugins`), not only in individual plugins.

2. **Uniform per-scorer sub-spans**
   - Introduce a shared tracing wrapper (decorator or `TracedScorer`) around `scheduling.Scorer.Score`, so every scorer emits a span with consistent naming, e.g. `llm_d.epp.scorer.<type>` with attributes:
     - `llm_d.scorer.type`, `llm_d.scorer.name`, `llm_d.scorer.weight`
     - `llm_d.scorer.candidate_count`
     - `llm_d.scorer.duration_ms` (or rely on span duration)
     - `llm_d.scorer.score.max`, `llm_d.scorer.score.avg`, `llm_d.scorer.endpoints_scored`
   - **Do not** add per-pod / per-endpoint span labels (cardinality).

3. **Align existing prefix-cache tracing**
   - `preciseprefixcache` already has `llm_d.epp.scorer.prefix_cache`; migrate it to the shared wrapper (or nest it cleanly) so prefix-cache and kv-cache spans (`llm_d.kv_cache.*` from embedded llm-d-kv-cache) appear under the same scoring parent.

4. **Trace propagation (follow-up if out of scope)**
   - Document or fix that incoming W3C context should be extracted at ext_proc entry so scoring spans attach to the gateway trace (can be a separate PR; link if not done here).

5. **Acceptance criteria**
   - With tracing enabled and a sampled request, Jaeger (or equivalent) shows: `gateway.request` → `gateway.request_orchestration` → `llm_d.epp.schedule` → one span per scorer → (when configured) `llm_d.epp.scorer.prefix_cache` / `llm_d.kv_cache.*`.
[...]

Issue #1519, opened 2026-06-07
[Observability] Standardize opentelemetry service, instrument and span naming


ahg-g:
Gemini's recommendation.

  ### 1. Service Name & Namespace (Resource Attributes)

  Since each component is deployed and scaled separately, they should not share a monolithic  "llm-d-router"  service name.

  •  service.name : Set to the specific deployable binary name:
      • Endpoint Picker:  "llm-d-epp" 
      • Disaggregation Sidecar:  "llm-d-disagg-sidecar" 
      • Future Coordinator:  "llm-d-coordinator" 
  •  service.namespace  (Recommended addition): Set to  "llm-d-router"  across all components. APM tools (like Jaeger or Datadog) use this attribute to group independent microservices under the same logical platform diagram.
  ──────
  ### 2. Instrumentation Scope ( otel.Tracer(name) )

  Avoid calling  otel.Tracer("llm-d-router")  or  otel.Tracer("llm-d-epp") . Instead, name the tracer after the fully qualified Go package import path executing inside that binary:

  • In  llm-d-disagg-sidecar :  "github.com/llm-d/llm-d-router/pkg/sidecar/proxy" 
  • In  llm-d-epp :  "github.com/llm-d/llm-d-router/pkg/epp/scheduling" 
  • In  llm-d-coordinator :  "github.com/llm-d/llm-d-router/pkg/coordinator" 
  ──────
  ### 3. Span Name ( tracer.Start(ctx, name) )

  Span names should be pure action descriptions without embedding component prefixes (e.g., avoid  "llm_d.pd_proxy.request"  or  "llm_d.epp.pick" ).

  Because backend tracing tools already group every span by  service.name = "llm-d-disagg-sidecar"  and  instrumentation.name :

  • In Sidecar Proxy: Use  "POST /v1/chat/completions"  or  "forward_request" 
  • In EPP: Use  "pick_endpoint"  or  "evaluate_scores" 
  • In Coordinator: Use  "coordinate_prefill"  or  "sync_topology"

gyliu513:
For instrumentation scope, probabluy we do not need to add `github.com/llm-d`, so the tracer scope will be as below:

- In llm-d-disagg-sidecar : "llm-d-router/pkg/sidecar/proxy"
- In llm-d-epp : "llm-d-router/pkg/epp/scheduling"
- In llm-d-coordinator : "llm-d-router/pkg/coordinator"

[...]

PR #1531, opened 2026-06-08
feat: add OpenTelemetry spans to scheduling scorer pipeline

mvanhorn:
/kind feature

## Summary

Adds fine-grained OpenTelemetry spans to the EPP scheduling and scoring pipeline so operators can see per-scorer timing and the parent scheduling span in traces.

## Why this matters

Scheduling and scoring (`SchedulerProfile.Run` / `runScorerPlugins`) had no tracing, so per-scorer latency was invisible in traces (#1483). This wraps `SchedulerProfile.Run` in a parent scheduling span and adds a shared tracing wrapper (`traced_scorer.go`) that instruments each scorer plugin, reusing the existing OTel tracer that `director.go` already sets up for the `request_orchestration` span. It covers the parent scheduling span and the shared per-scorer wrapper (deliverables 1 and 2); migrating the existing prefix-cache span and ext_proc context propagation (3 and 4), plus the naming/flow-control question from the thread, are left as follow-ups.

## Testing

- Added `traced_scorer_test.go`; `go build ./pkg/epp/...` passes.

## Release note

```release-note
Add OpenTelemetry spans for the scheduling/scoring pipeline (parent scheduling span and per-scorer spans).
```

Refs #1483

### DIFF
--- a/pkg/epp/scheduling/scheduler.go
+++ b/pkg/epp/scheduling/scheduler.go
@@ -35,8 +35,6 @@ const (
 	scorerExtensionPoint                 = "Scorer"
 	pickerExtensionPoint                 = "Picker"
 	processProfilesResultsExtensionPoint = "ProcessProfilesResults"
-	schedulerInstrumentationName         = "llm-d-router"
-	scheduleSpanName                     = "llm_d.epp.schedule"
 )
 
 // NewSchedulerWithConfig returns a new scheduler with the given scheduler plugins configuration.
--- a/pkg/epp/scheduling/scheduler.go
+++ b/pkg/epp/scheduling/scheduler.go
@@ -35,6 +35,8 @@ const (
 	scorerExtensionPoint                 = "Scorer"
 	pickerExtensionPoint                 = "Picker"
 	processProfilesResultsExtensionPoint = "ProcessProfilesResults"
+	schedulerInstrumentationName         = "llm-d-router"
+	scheduleSpanName                     = "llm_d.epp.schedule"
 )
 
 // NewSchedulerWithConfig returns a new scheduler with the given scheduler plugins configuration.
--- a/pkg/epp/scheduling/scheduler_profile.go
+++ b/pkg/epp/scheduling/scheduler_profile.go
@@ -34,20 +34,35 @@ import (
 	"github.com/llm-d/llm-d-router/pkg/epp/metrics"
 )
 
+// SchedulerProfileOption configures a SchedulerProfile.
+type SchedulerProfileOption func(*SchedulerProfile)
+
+// WithScorerTracing controls whether scorer plugins are wrapped with tracing spans.
+func WithScorerTracing(enabled bool) SchedulerProfileOption {
+	return func(p *SchedulerProfile) {
+		p.traceScorers = enabled
+	}
+}
+
 // NewSchedulerProfile creates a new SchedulerProfile object and returns its pointer.
-func NewSchedulerProfile() *SchedulerProfile {
-	return &SchedulerProfile{
+func NewSchedulerProfile(options ...SchedulerProfileOption) *SchedulerProfile {
+	profile := &SchedulerProfile{
 		filters: []fwksched.Filter{},
 		scorers: []*WeightedScorer{},
 		// picker remains nil since profile doesn't support multiple pickers
 	}
+	for _, option := range options {
+		option(profile)
+	}
+	return profile
 }
 
 // SchedulerProfile provides a profile configuration for the scheduler which influence routing decisions.
 type SchedulerProfile struct {
-	filters []fwksched.Filter
-	scorers []*WeightedScorer
-	picker  fwksched.Picker
+	filters      []fwksched.Filter
+	scorers      []*WeightedScorer
+	picker       fwksched.Picker
+	traceScorers bool
 }
 
 // WithFilters sets the given filter plugins as the Filter plugins.
@@ -60,7 +75,10 @@ func (p *SchedulerProfile) WithFilters(filters ...fwksched.Filter) *SchedulerPro
 // WithScorers sets the given scorer plugins as the Scorer plugins.
 // if the SchedulerProfile has Scorer plugins, this call replaces the existing plugins with the given ones.
 func (p *SchedulerProfile) WithScorers(scorers ...*WeightedScorer) *SchedulerProfile {
-	p.scorers = scorers
+	p.scorers = make([]*WeightedScorer, 0, len(scorers))
+	for _, scorer := range scorers {
+		p.scorers = append(p.scorers, p.wrapScorer(scorer))
+	}
 	return p
 }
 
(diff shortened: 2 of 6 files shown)


pkg/epp/requestcontrol/director.go
```diff
@@ -56,8 +56,9 @@ import (
 const (
 	// dataProducerTimeout is the default per-producer execution timeout. A
 	// producer overrides it by implementing requestcontrol.TimeoutAwareProducer.
-	dataProducerTimeout       = 400 * time.Millisecond
-	responseBodyQueueCapacity = 100
+	dataProducerTimeout         = 400 * time.Millisecond
+	responseBodyQueueCapacity   = 100
+	directorInstrumentationName = "llm-d-router"
```
shmuelk: Why isn't this a common name and constant accross all components of the EPP?

pkg/epp/scheduling/scheduler.go
```diff
@@ -35,6 +35,8 @@ const (
 	scorerExtensionPoint                 = "Scorer"
 	pickerExtensionPoint                 = "Picker"
 	processProfilesResultsExtensionPoint = "ProcessProfilesResults"
+	schedulerInstrumentationName         = "llm-d-router"
+	scheduleSpanName                     = "llm_d.epp.schedule"
```
shmuelk:
```suggestion
	schedulerSpanName                     = "epp.scheduler"
```

pkg/epp/scheduling/scheduler_profile.go
```diff
@@ -169,7 +173,7 @@ func (p *SchedulerProfile) runScorerPlugins(ctx context.Context, request *fwksch
 	for _, scorer := range p.scorers {
 		logger.V(logutil.VERBOSE).Info("Running scorer plugin", "plugin", scorer.TypedName())
 		before := time.Now()
-		scores := scorer.Score(ctx, request, endpoints)
+		scores := NewTracedScorer(scorer.Scorer, scorer.Weight()).Score(ctx, request, endpoints)
```
shmuelk: Why not do this at profile build time and not at scorer run time?

This will save memory, GC, and cycles.

pkg/epp/scheduling/traced_scorer.go
```diff
@@ -0,0 +1,103 @@
...
+	scorerCandidateAttribute    = "llm_d.scorer.candidate_count"
+	scorerEndpointsAttribute    = "llm_d.scorer.endpoints_scored"
+	scorerMaxScoreAttribute     = "llm_d.scorer.score.max"
+	scorerAverageScoreAttribute = "llm_d.scorer.score.avg"
+)
+
+func schedulerTracer() trace.Tracer {
+	return otel.Tracer(schedulerInstrumentationName)
+}
+
```
shmuelk: Please add:

var _  fwksched.Scorer = &TracedScorer{}

and other functions needed to be a true scorer.

shmuelk:
Overall a great idea to wrap the scorers, which can carried forward to otherplugins as desired, in future PRs.

Just do the wrapping at profile creation time instead of scorer execution time.

Even better it can be done conditionally if tracing is enabled. Thus getting rid runtime checks.

pkg/epp/requestcontrol/director.go
gyliu513: No need to define this, as the default scope for `tracing.Tracer()` is `llm-d-router`

pkg/epp/requestcontrol/director.go
```diff
@@ -231,7 +232,7 @@ func (d *Director) getInferenceObjective(ctx context.Context, reqCtx *handlers.R
 // HandleRequest orchestrates the request lifecycle.
 // It always returns the requestContext even in the error case, as the request context is used in error handling.
 func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestContext, inferenceRequestBody *fwkrh.InferenceRequestBody) (*handlers.RequestContext, error) {
-	tracer := otel.Tracer("llm-d-router")
+	tracer := otel.Tracer(directorInstrumentationName)
```
gyliu513: I think here is already calling`tracing.Tracer()`

pkg/epp/scheduling/traced_scorer.go
```diff
@@ -0,0 +1,103 @@
...
+	scorerWeightAttribute       = "llm_d.scorer.weight"
+	scorerCandidateAttribute    = "llm_d.scorer.candidate_count"
+	scorerEndpointsAttribute    = "llm_d.scorer.endpoints_scored"
+	scorerMaxScoreAttribute     = "llm_d.scorer.score.max"
+	scorerAverageScoreAttribute = "llm_d.scorer.score.avg"
+)
+
+func schedulerTracer() trace.Tracer {
+	return otel.Tracer(schedulerInstrumentationName)
+}
```
gyliu513: Can you use `github.com/llm-d/llm-d-router/pkg/common/observability/tracing.Tracer("llm-d-router/pkg/epp/scheduling")`, I want to use `llm-d-router/pkg/epp/scheduling` as scope here.

```python
func schedulingTracer() trace.Tracer {
    return tracing.Tracer("llm-d-router/pkg/epp/scheduling")
}
```

pkg/epp/scheduling/scheduler.go
gyliu513: Based on https://github.com/llm-d/llm-d-router/issues/1519, I think @shmuelk is right, but I think we can even remove `epp`, just name it as `scheduler`, as we will have a clear scope for the span.

pkg/epp/scheduling/traced_scorer.go
gyliu513: This also align with https://github.com/llm-d/llm-d-router/issues/1519 and I have a PR at https://github.com/llm-d/llm-d-router/pull/1565

ahg-g:
can you please rebase and align with https://github.com/llm-d/llm-d-router/pull/1565? this needs to happen today to make it in the release.

mvanhorn:
Sorry this missed the release window - picking it back up. Done in a8c3f53:
- @shmuelk's suggestion implemented: scorers are now wrapped once at profile creation (WithScorerTracing option on NewSchedulerProfile, threaded from the config loader), and only when tracing is enabled - no runtime checks left in the scoring hot path.
- Rebased onto main and aligned with #1565's conventions (tracer scope, snake_case span names, llm_d.epp.* attribute keys), reusing the shared test span helper.
Build, vet, and the scheduling/config/runner test suites pass locally.

pkg/epp/scheduling/traced_scorer.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+const (
+	scorerSpanName              = "score_endpoints"
+	scorerTypeAttribute         = "llm_d.epp.scorer.type"
+	scorerNameAttribute         = "llm_d.epp.scorer.name"
+	scorerWeightAttribute       = "llm_d.epp.scorer.weight"
+	scorerCandidateAttribute    = "llm_d.epp.scorer.candidate_endpoints"
+	scorerEndpointsAttribute    = "llm_d.epp.scorer.scored_endpoints"
+	scorerMaxScoreAttribute     = "llm_d.epp.scorer.max_score"
```
ahg-g, line 37: Instead of just the max, how about logging the top five scores and the corresponding endpoints names as a string, something like `ep1:score1, ep2:score2` etc?

gyliu513:
@mvanhorn I think this PR can be closed, right?

mvanhorn:
Yes, closing. #1834 landed the same OpenTelemetry instrumentation on the scheduler scoring path, so this one is redundant. Thanks for the reviews here.

PR #1539, opened 2026-06-08
fix(tracing): unify tracer scope and version metadata

### LINKED ISSUE #1538: [Observability] Unify EPP tracer instrumentation scope and version metadata across call sites
**What happened**:

| Call site | scope name | version | attrs |
|---|---|---|---|
| `tracing.Tracer()` helper | `llm-d-router` | none | none |
| https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/handlers/server.go#L204| `llm-d-router/epp/extproc` | `BuildRef` | `commit-sha` |
| https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/requestcontrol/director.go#L234 | `llm-d-router` | none | none |

The OTel instrumentation scope (name + version + attributes) is stamped onto every span a tracer produces. Two concrete problems follow from the current state:

1. **Build/version metadata is attached to only one scope.** Within a single request, `gateway.request` (from `server.go`) carries `version` and `commit-sha`, while `gateway.request_orchestration` (from `director.go`) and spans created via the helper do not. Answering "which build produced this trace" depends on that one span surviving sampling/truncation.
2. **`director.go` bypasses the helper.** It hand-rolls `otel.Tracer("llm-d-router")`, matching the helper's name only by coincidence. Adding a version to the helper (the fix for problem 1) would leave this call site silently emitting version-less spans.

Consolidating through a single helper that always tags version + commit keeps per-component scope names while making span metadata uniform across a trace.

/cc @ahg-g 

/assign

**Environment**:
- Install tools:
- Others:

gyliu513:
**What type of PR is this?**

/kind bug

EPP created tracers in three inconsistent ways: the `tracing.Tracer()` helper used scope `llm-d-router` with no version, `server.go` hand-rolled`llm-d-router/epp/extproc` with the build ref and commit SHA, and `director.go` hand-rolled `llm-d-router` with no version. 

The OTel instrumentation scope (name + version + attributes) is stamped onto every span, so within a single
trace some spans carried build/version metadata and others did not, and `director.go` duplicated the scope name literal instead of going through the helper.

This routes all tracer creation through `tracing.Tracer()`:

- `Tracer()` takes an optional scope name (defaulting to `llm-d-router`) and always attaches `version.BuildRef` and the `commit-sha` attribute.
- `server.go` uses `Tracer("llm-d-router/epp/extproc")`, preserving the per-component scope name while getting version metadata from the helper.
- `director.go` uses `Tracer()`, dropping the duplicated literal and the direct `otel.Tracer` call.

Fixes #1538 

/cc @ahg-g 

```release-note
EPP trace spans now consistently carry the build version and commit SHA across all instrumentation scopes, so a full request trace can be attributed to a single build.
```

### DIFF
--- a/pkg/common/observability/tracing/telemetry.go
+++ b/pkg/common/observability/tracing/telemetry.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
 	"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
 	"go.opentelemetry.io/otel/propagation"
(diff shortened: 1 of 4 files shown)


ahg-g:
looks good, pls resolve the conflict.

gyliu513:
@ahg-g Conflict resolved.

PR #1543, opened 2026-06-08
test(flowcontrol): add full-path stress benchmark

RishabhSaini:
Full-path stress benchmark exercises the real producer->detector->controller pipeline under concurrent load across multiple priority bands with leak detection via counter assertions and memprofile support.


ahg-g:
@RishabhSaini can this run automatically as part of CI?

ahg-g (approved):

Issue #1495, opened 2026-06-05
concurrency-detector: New pods reports saturation as 1


asharkhan3101:
**What happened**:

When running in HA mode with active / passive setup the active pod with `concurrency-detector` plugin reports the saturation properly but the passive pod always reports saturation as `1`. 

**What you expected to happen**:

The passive pod is not actively participating in the request routing and hence it should report zero saturation.

**How to reproduce it (as minimally and precisely as possible)**:

Run EPP in active passive mode with `concurrency-detector` and take a look at the metrics emitted by each of the pods, the active pod will report proper saturation whereas the passive one will report as `1`.

**Anything else we need to know?**:

This behaviour is different from what is present in the utilization-detector, its saturation is reported as 0 for passive pod. It is due to this that autoscaling config which takes into account the pool saturation needs to change when saturation plugin changes.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.34
- llm-d-router version (use `git describe --tags --dirty --always` if you built from source, or specify the tag if you used a tagged version or image): https://github.com/llm-d/llm-d-router/pull/1478
- Cloud provider or hardware configuration: N/A
- Install tools: N/A
- Others: N/A

asharkhan3101:
This actually makes switching the saturation detector difficult because with KEDA the expression also needs to change when detector changes, let's say I want scaling to trigger at 0.8 but with concurrency detector (assuming sum aggregation), it should be 1.8 instead (1 base reported by passive pod).

Or the other option is to disable HA mode.

asharkhan3101:
A more general statement for this would be "concurrency-detector: Saturation is 1 if pod has not handled any traffic" which also includes the passive pod setup. Due to this, autoscaling event trigger immediately after the cluster has been provisioned because of saturation == 1. Once at least one traffic hits the system, its saturation numbers return normal.

LukeAVanDrie:
Tagging this for v0.9. Already signed off on the PR; thanks both of you for flagging & fixing this!

PR #1545, opened 2026-06-08
test(flowcontrol): Add flow control integration tests

RishabhSaini:
Integration tests for the flow control subsystem covering cross-component data paths, dispatch ordering, priority backpressure, fairness, capacity enforcement, eviction pipeline, error outcomes, graceful shutdown, endpoint churn, and sustained-load counter accuracy.

Bug fixes included because tests assert the corrected behavior:
- Guard nil endpoint in concurrency detector Saturation() and Filter()
- Track endpoint identity in InFlightLoadProducer to prevent stale delete events from clearing a replacement pod's tracker

### DIFF
--- a/pkg/epp/flowcontrol/integration_test.go
+++ b/pkg/epp/flowcontrol/integration_test.go
@@ -1,0 +1,1335 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package flowcontrol_test
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	reqcommon "github.com/llm-d/llm-d-router/pkg/common/request"
+	contractmocks "github.com/llm-d/llm-d-router/pkg/epp/flowcontrol/contracts/mocks"
+	"github.com/llm-d/llm-d-router/pkg/epp/flowcontrol/controller"
+	"github.com/llm-d/llm-d-router/pkg/epp/flowcontrol/eviction"
+	"github.com/llm-d/llm-d-router/pkg/epp/flowcontrol/registry"
+	fcTypes "github.com/llm-d/llm-d-router/pkg/epp/flowcontrol/types"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/flowcontrol"
+	fwkplugin "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requestcontrol"
+	fwkrh "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requesthandling"
+	fwksched "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/eviction/filtering"
+	evictionordering "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/eviction/ordering"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/fairness/globalstrict"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/fairness/roundrobin"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/ordering/fcfs"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/ordering/slodeadline"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/usagelimits"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload"
+	"github.com/llm-d/llm-d-router/pkg/epp/metadata"
+	eppmetrics "github.com/llm-d/llm-d-router/pkg/epp/metrics"
+	testutils "github.com/llm-d/llm-d-router/test/utils"
+)
+
+// ============================================================================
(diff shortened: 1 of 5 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
```diff
@@ -196,9 +197,15 @@ func (p *InFlightLoadProducer) Extract(ctx context.Context, event datalayer.Endp
 
 	switch event.Type {
 	case datalayer.EventDelete:
+		if registered, ok := p.registeredEndpoints.Load(id); ok && registered != event.Endpoint {
```
LukeAVanDrie, line 291: This relies on the `datalayer.Endpoint` instance being stable across the add and the later delete for a given pod generation. That holds today, but it's an implicit contract: if the datalayer ever reconstructs endpoint objects on delete, the guard would start leaking trackers instead of cleaning them up. Worth a comment so the assumption is visible.

pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go
```diff
@@ -127,7 +127,7 @@ func (d *detector) getLoad(m datalayer.AttributeMap) *attrconcurrency.InFlightLo
 func (d *detector) Saturation(_ context.Context, endpoints []datalayer.Endpoint) float64 {
 	var totalInflight, totalCapacity int64
 	for _, e := range endpoints {
-		if e.GetMetadata() == nil {
+		if e == nil || e.GetMetadata() == nil {
```
LukeAVanDrie: nit: There's no unit test for the nil case in this package; it's only covered from the integration file. A small test here may be helpful.

Also, you may want to look at #1495 (and #1580).

pkg/epp/flowcontrol/integration_test.go
```diff
@@ -0,0 +1,2231 @@
...
+
+// ============================================================================
+// Full-Loop Controller Tests (detector wired into dispatch cycle)
+// ============================================================================
+
+// TestSaturationFullLoop wires a real InFlightLoadProducer, real concurrency
+// detector, and real persistent endpoints into the FlowController's dispatch
+// cycle via EndpointCandidates.Locate(). Tests the exact path that broke
+// in #1474.
+func TestSaturationFullLoop(t *testing.T) {
```
LukeAVanDrie, line 133: This test, along with `TestUsageLimitThresholdGatesDispatch`, `TestHighConcurrencyFlowChurnNoDeadlock`, and `TestEndpointChurnUnderLoad`, duplicates ~40 lines of registry/controller/producer/detector setup that `newHarness` already does, and the producer+detector block recurs ~10 times across the file. A `newProducerAndConcurrencyDetector(t, handle, cfg)` helper plus a `harnessOpts` field to inject a pre-built producer/detector and `EndpointCandidates` would let these go through the harness. Fine as a follow-up if you'd rather not expand this PR.

pkg/epp/flowcontrol/integration_test.go
```diff
@@ -0,0 +1,2231 @@
...
+			}
+			outcome, err := h.fc.EnqueueAndWait(h.ctx, req)
+			results <- dispatchResult{id: s.id, outcome: outcome, err: err}
+			detector.Release()
+		}()
+		time.Sleep(5 * time.Millisecond)
+	}
+
+	time.Sleep(20 * time.Millisecond)
+	detector.Unblock(1)
```
LukeAVanDrie, line 251: The ordering assertion below assumes all three requests are queued during the preceding `time.Sleep(20 * time.Millisecond)`. It held under -cpu=1 -count=10 locally, but that fixed window is where it could flake on a loaded CI runner. Polling queue depth is safer. Generally want to avoid `time.Sleep` anywhere in CI where it can be avoided.

```
require.Eventually(t, func() bool {
	return h.fc /* registry */ .Stats().TotalLen == uint64(len(specs))
}, time.Second, time.Millisecond)
detector.Unblock(1)
```

Same applies to `TestPriorityBackpressure` and `TestFairnessRoundRobin`.

LukeAVanDrie (dismissed):
This is some really solid coverage, and the saturation-gates-dispatch tests pin down the #1474 path nicely. Thanks for taking this on!

This is obviously a massive file with a ton of new tests, but I don't think the tests are brittle. They assert on public contract (outcome enums, error sentinels, dispatch gating), and `GetTokens`/`GetRequests` are production API rather than test-only hooks, so a behavior-preserving refactor shouldn't break them. The cost is in update churn and readability. A couple non-blocking thoughts:

1. Organize by test level, not topic. The file mixes true end-to-end tests (full-loop dispatch, ordering, priority, fairness, capacity, shutdown, churn, zombie, metrics) with single-component tests (producer-only or detector-only: data path, token mode, SOS release, multi-profile, single-chunk, reregistration, nil endpoint, identity collision, plugin-state-delete, sustained-load, concurrent reads). Moving the component-level ones down to `producer_test.go` / `concurrency/detector_test.go`, where their equivalents already live, roughly halves this file and removes the redundancy at the same time. If that's too much for this PR, splitting by level within the package (`integration_test.go` plus `integration_components_test.go`) is a lighter version.

2. Trim the duplicates regardless. Several subtests re-assert single-component counter math the unit suites already cover: `TestSingleChunkResponseLifecycle` vs `TestInFlightLoadProducer_ExcludeOutputTokens_SingleChunk`, `TestPluginStateDeleteTriggersCounterCleanup` vs `TestInFlightLoadProducer_Eviction`, `TestMultiProfilePrefillDecodeTracking` vs `TestInFlightLoadProducer_MultiPodLifecycle`, `TestSustainedLoadCounterAccuracy` vs `TestInFlightLoadProducer_ConcurrencyStress`, `TestSaturationDataPath`/`TestSaturationTokenMode` vs `TestDetector_Saturation`/`TestDetector_TokenSaturation`. A behavior change means updating both copies for the same protection, so I'd drop the ones with no cross-component value.

RishabhSaini:
Addressed the actionable items (endpoint identity comment, nil unit test, require.Eventually) and split component-level tests

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
RishabhSaini: agreed added a comment mentioning this

pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go
RishabhSaini: agreed, added a test in concurrency detector test

pkg/epp/flowcontrol/integration_test.go
RishabhSaini: will followup refactor to this file after this PR merge

pkg/epp/flowcontrol/integration_test.go
RishabhSaini: agreed, noted, updated

pkg/epp/flowcontrol/integration_test.go
RishabhSaini: ok refactored

pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go
```diff
@@ -131,14 +131,16 @@ func (d *detector) Saturation(_ context.Context, endpoints []datalayer.Endpoint)
...
+			continue
+		}
+
 		if d.config.mode == modeTokens {
 			totalCapacity += d.config.maxTokenConcurrency
 		} else {
 			totalCapacity += d.config.maxConcurrency
 		}
 
-		// nil metadata = endpoint not yet populated by this EPP instance
```
LukeAVanDrie, line 140: nit: This seems like an accidental removal during rebase.

LukeAVanDrie (approved):

ahg-g (approved):

PR #1547, opened 2026-06-08
feat(prefixcacheaffinity): derive TTFT gate from peak prefill throughput

kaushikmitr:
Consolidate the TTFT-penalty and in-flight-tokens gates into a single TTFT load gate with a configurable source. useLatencyPredictor (default true) reads TTFT from the latency predictor; when false, TTFT is estimated from in-flight tokens and peakPrefillThroughput (tokens/sec). Removes MaxTokensInFlightPenalty.

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
+++ b/pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
@@ -74,9 +74,9 @@ type Config struct {
 	MaxTTFTPenaltyMs float64 `json:"maxTTFTPenaltyMs,omitempty"`
 
 	// TTFTSource selects where the load gate reads per-endpoint TTFT from.
-	// TTFTSourceLatencyPredictor (default) reads predicted TTFT from the latency
-	// predictor; TTFTSourcePrefillThroughput estimates it from in-flight tokens
-	// and PeakPrefillThroughput.
+	// TTFTSourcePrefillThroughput (default) estimates it from in-flight tokens and
+	// PeakPrefillThroughput; TTFTSourceLatencyPredictor reads predicted TTFT from
+	// the latency predictor.
 	TTFTSource TTFTSource `json:"ttftSource,omitempty"`
 
 	// PeakPrefillThroughput is the peak prefill throughput in tokens/sec, used to
@@ -94,7 +94,7 @@ var DefaultConfig = Config{
 	AffinityThreshold:      0.80,
 	ExplorationProbability: 0.01,
 	MaxTTFTPenaltyMs:       5000,
-	TTFTSource:             TTFTSourceLatencyPredictor,
+	TTFTSource:             TTFTSourcePrefillThroughput,
 
 	// Calibrated for Qwen 32B on 2x H100 80GB (TP=2), vLLM 0.19; see README.
 	PeakPrefillThroughput: 15928,
@@ -141,7 +141,7 @@ func (c *Config) validate() error {
 		return fmt.Errorf("peakPrefillThroughput must be >= 0, got %f", c.PeakPrefillThroughput)
 	}
 	switch c.TTFTSource {
-	case "", TTFTSourceLatencyPredictor, TTFTSourcePrefillThroughput:
+	case TTFTSourceLatencyPredictor, TTFTSourcePrefillThroughput:
 	default:
 		return fmt.Errorf("ttftSource must be %q or %q, got %q", TTFTSourceLatencyPredictor, TTFTSourcePrefillThroughput, c.TTFTSource)
 	}
@@ -152,10 +152,10 @@ func (c *Config) validate() error {
 }
 
 // usesLatencyPredictor reports whether the load gate sources TTFT from the
-// latency predictor. The predictor is the default; only an explicit
-// prefillThroughput selects the in-flight-token estimate.
+// latency predictor. Throughput is the default; only an explicit
+// latencyPredictor selects the predictor.
 func (c *Config) usesLatencyPredictor() bool {
-	return c.TTFTSource != TTFTSourcePrefillThroughput
+	return c.TTFTSource == TTFTSourceLatencyPredictor
 }
 
 func (p *Plugin) TypedName() fwkplugin.TypedName {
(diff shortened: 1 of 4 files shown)


shmuelk:
This PR makes a significant external change in that by default it requires the Latency Predictor.

Is that desirable?

At a minimum release notes are required.

kaushikmitr:
> This PR makes a significant external change in that by default it requires the Latency Predictor.
> 
> Is that desirable?
> 
> At a minimum release notes are required.

Currently only helm chart using this is the latency predictor so that is why made it default. We can instead turn it off and update the helm chart. But i think since no other chart is using this, we can keep it as default

pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
```diff
@@ -53,16 +54,21 @@ type Config struct {
...
 
-	// MaxTokensInFlightPenalty is the max in-flight token penalty before breaking
-	// stickiness. If the best sticky endpoint's in-flight tokens exceed the best
-	// non-sticky endpoint's in-flight tokens by more than this value, all endpoints
-	// are kept. Set to 0 to disable this gate. Default: 0 (disabled).
-	MaxTokensInFlightPenalty int64 `json:"maxTokensInFlightPenalty,omitempty"`
+	// UseLatencyPredictor selects the TTFT source for the load gate. When true
+	// (default), per-endpoint TTFT is read from the latency predictor. When
+	// false, TTFT is estimated from in-flight tokens and PeakPrefillThroughput.
+	UseLatencyPredictor bool `json:"useLatencyPredictor,omitempty"`
```
ahg-g: instead of having an explicit flag, we can check if the latency is produced and use it, if not fallback to the other approach

pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
```diff
@@ -73,6 +79,8 @@ var DefaultConfig = Config{
 	AffinityThreshold:      0.80,
 	ExplorationProbability: 0.01,
 	MaxTTFTPenaltyMs:       5000,
+	UseLatencyPredictor:    true,
+	PeakPrefillThroughput:  15928,
```
ahg-g: add a comment to justify this default

pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
ahg-g: I see we need to decide on the producer dependencies using this flag, the other option is to have this as an enum, ask antigravity (or antigravity-cli) for a suggestion for this enum's parameter and options names. I think the api will be more descriptive with an enum than a flag.

kaushikmitr: i used gemini to create an enum to determine which method to pick

pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
kaushikmitr: done

pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
```diff
@@ -73,6 +94,10 @@ var DefaultConfig = Config{
 	AffinityThreshold:      0.80,
 	ExplorationProbability: 0.01,
 	MaxTTFTPenaltyMs:       5000,
+	TTFTSource:             TTFTSourceLatencyPredictor,
```
ahg-g: I think the default should be `prefillThroughput` since it doesn't require extra set up and it just works.

pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
```diff
@@ -112,12 +137,27 @@ func (c *Config) validate() error {
 	if c.MaxTTFTPenaltyMs < 0 {
 		return fmt.Errorf("maxTTFTPenaltyMs must be >= 0, got %f", c.MaxTTFTPenaltyMs)
 	}
-	if c.MaxTokensInFlightPenalty < 0 {
-		return fmt.Errorf("maxTokensInFlightPenalty must be >= 0, got %d", c.MaxTokensInFlightPenalty)
+	if c.PeakPrefillThroughput < 0 {
+		return fmt.Errorf("peakPrefillThroughput must be >= 0, got %f", c.PeakPrefillThroughput)
+	}
+	switch c.TTFTSource {
+	case "", TTFTSourceLatencyPredictor, TTFTSourcePrefillThroughput:
```
ahg-g: we should fail on the empty case, since we have a default.

pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
```diff
@@ -112,12 +137,27 @@ func (c *Config) validate() error {
...
+		return errors.New("peakPrefillThroughput must be > 0 when ttftSource is prefillThroughput")
 	}
 	return nil
 }
 
+// usesLatencyPredictor reports whether the load gate sources TTFT from the
+// latency predictor. The predictor is the default; only an explicit
+// prefillThroughput selects the in-flight-token estimate.
+func (c *Config) usesLatencyPredictor() bool {
+	return c.TTFTSource != TTFTSourcePrefillThroughput
```
ahg-g: This will be wrong if a new case is added. I know you are doing this to cover the empty case, but as mentioned above the empty case should be an error, defaults should be set explicitly. 


```suggestion
	return c.TTFTSource == TTFTSourceLatencyPredictor
```

pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
kaushikmitr: done

ahg-g (approved):

PR #1562, opened 2026-06-09
test(e2e): add cache-affinity test for multimodal routing

### LINKED ISSUE #1541: [E2E Tests] Add cache-affinity assertion for multimodal routing

An e2e test that asserts identical multimodal content routes to the same decode pod across consecutive requests, using the `mm-embeddings-cache-producer` and `mm-embeddings-cache-scorer` profile.

The existing multimodal cases in `test/e2e/e2e_test.go` assert that the responding pod is in the candidate pool set (`BeElementOf(decodePods)`) but use only 1 decode replica, which makes cache-affinity trivially true.

**Why is this needed**:

`mm-embeddings-cache-producer` and `mm-embeddings-cache-scorer` ship on main but have no CI assertion that cache-affinity routing actually fires. A regression silently demoting same-content routing to round-robin would not be caught.

hexfusion:
**What type of PR is this?**

/kind test

Adds an e2e case under "Running multimodal cache-affinity configuration" with five assertions on the routing pipeline:

1. **Same-content same-pod** across image, audio, and video.
2. **A/B image symmetry**: distinct content keys route back to their first-served pod (catches "first-request-wins-forever" bugs).
3. **Mixed image + audio + video**: probes the empty-hash short-circuit in `ExtractMMItems`.
4. **Metric assertions** on `llm_d_router_epp_encoder_cache_{hits,queries}_total` with `modality` label (hits ≤ queries). Wrapped in `gomega.Eventually` for the async `PreRequest` LRU write.
5. **Estimate backend per-modality metric**: same config with the estimate token-producer; depends on #1618.

`swapToSimRenderSidecar` swaps the EPP's render sidecar to the inference sim (deterministic `mm_features` for all three modalities). Triggered only when the EPP config contains `mm-embeddings-cache-producer`.

Fixes #1541

```release-note
NONE
```

### DIFF
--- a/test/e2e/configs_test.go
+++ b/test/e2e/configs_test.go
@@ -269,6 +269,30 @@ schedulingProfiles:
   - pluginRef: max-score-picker
 `
 
+// Same as mmCacheAffinityConfig but the token-producer uses the estimate
+// backend, which derives mm_features in-EPP and never calls the render endpoint.
+const mmCacheAffinityEstimateConfig = `apiVersion: llm-d.ai/v1alpha1
+kind: EndpointPickerConfig
+plugins:
+- type: token-producer
+  parameters:
+    estimate: {}
+- type: mm-embeddings-cache-producer
+  parameters:
+    cacheSizeInMBPerServer: 64
+- type: mm-embeddings-cache-scorer
+- type: decode-filter
+- type: max-score-picker
+- type: single-profile-handler
+schedulingProfiles:
+- name: default
+  plugins:
+  - pluginRef: decode-filter
+  - pluginRef: mm-embeddings-cache-scorer
+    weight: 10
+  - pluginRef: max-score-picker
+`
+
 // EPP configuration for running scale model server test
 const scaleConfig = `apiVersion: llm-d.ai/v1alpha1
 kind: EndpointPickerConfig
--- a/test/e2e/configs_test.go
+++ b/test/e2e/configs_test.go
@@ -242,6 +242,33 @@ schedulingProfiles:
     weight: 10
 `
 
+// EPP configuration for multimodal cache-affinity routing. Pairs the
+// mm-embeddings-cache producer+scorer with token-producer at loopback; the sim
+// sidecar (swapped in by usesSimRender) provides mm_features.
+const mmCacheAffinityConfig = `apiVersion: llm-d.ai/v1alpha1
+kind: EndpointPickerConfig
+plugins:
+- type: token-producer
+  parameters:
+    modelName: food-review
+    vllm:
+      url: http://localhost:8000
+- type: mm-embeddings-cache-producer
+  parameters:
+    cacheSizeInMBPerServer: 64
+- type: mm-embeddings-cache-scorer
+- type: decode-filter
+- type: max-score-picker
+- type: single-profile-handler
+schedulingProfiles:
+- name: default
+  plugins:
+  - pluginRef: decode-filter
+  - pluginRef: mm-embeddings-cache-scorer
+    weight: 10
+  - pluginRef: max-score-picker
+`
+
 // EPP configuration for running scale model server test
 const scaleConfig = `apiVersion: llm-d.ai/v1alpha1
 kind: EndpointPickerConfig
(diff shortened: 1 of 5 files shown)


test/e2e/configs_test.go
```diff
@@ -242,6 +242,33 @@ schedulingProfiles:
...
+// EPP configuration for multimodal cache-affinity routing. Pairs the
+// mm-embeddings-cache producer+scorer with token-producer at loopback; the sim
+// sidecar (swapped in by usesSimRender) provides mm_features.
+const mmCacheAffinityConfig = `apiVersion: llm-d.ai/v1alpha1
+kind: EndpointPickerConfig
+plugins:
+- type: token-producer
+  parameters:
+    modelName: food-review
+    vllm:
```
capri-xiyue, line 254: Can we have two epp config, one with `vllm` backend token-producer and the other with `estimate` backend token producer

test/e2e/utils_test.go
```diff
@@ -240,6 +280,56 @@ func filterDocument(doc string) string {
 	return strings.TrimRight(string(out), "\n")
 }
 
+// swapToSimRenderSidecar rewrites the vllm-render sidecar to run the inference
+// sim, which emits mm_features that text-only vLLM does not.
+func swapToSimRenderSidecar(inputs []string, simImage, modelName string) []string {
```
capri-xiyue: nits: `inputs` here is a little bit misleading, can you change it to a better name? Seems like you are using epp config here?

test/e2e/utils_test.go
```diff
@@ -240,6 +280,56 @@ func filterDocument(doc string) string {
...
+	outputs := make([]string, len(inputs))
+	for idx, input := range inputs {
+		docs := strings.Split(input, "\n---")
```
capri-xiyue: nits: variable `docs` name is also confusing

test/e2e/e2e_test.go
```diff
@@ -698,6 +698,71 @@ var _ = ginkgo.Describe("Run end to end tests", ginkgo.Ordered, func() {
...
+			_, podA := runChatCompletionWithImages(testImageURL)
+			_, podB := runChatCompletionWithImages(testImageURL2)
+			_, podARe := runChatCompletionWithImages(testImageURL)
+			gomega.Expect(podARe).Should(gomega.Equal(podA))
+			_, podBRe := runChatCompletionWithImages(testImageURL2)
+			gomega.Expect(podBRe).Should(gomega.Equal(podB))
+
+			// Mixed image + audio in one request. Probes per-item match
+			// accounting and the empty-hash short-circuit in ExtractMMItems.
+			ginkgo.By("mixed image+audio cache affinity")
```
capri-xiyue: Do you also need video request?

test/e2e/e2e_test.go
```diff
@@ -698,6 +698,71 @@ var _ = ginkgo.Describe("Run end to end tests", ginkgo.Ordered, func() {
...
+			_, podMix := runChatCompletionWithImageAndAudio(testImageURL, testAudioData)
+			_, podMixRe := runChatCompletionWithImageAndAudio(testImageURL, testAudioData)
+			gomega.Expect(podMixRe).Should(gomega.Equal(podMix))
+
+			// Producer emits hits + queries on the shared metric registry.
+			// PreRequest is async (wg.Go), so use Eventually to avoid the race.
+			ginkgo.By("metrics: hits_total + queries_total")
```
capri-xiyue, line 751: Do you expect llm_d_router_epp_encoder_cache_hits_total==llm_d_router_epp_encoder_cache_queries_total?

test/e2e/e2e_test.go
hexfusion: two modalities was enough to exercise multi-item accounting in ExtractMMItems, but fair point all three reads as more complete.

test/e2e/e2e_test.go
hexfusion: no hits_total is the matched subset of queries_total, first-touch is always a miss, so hits < queries. added a comment to make this more clear.

test/e2e/configs_test.go
hexfusion: resolving this

hexfusion: this needed a code change to work https://github.com/llm-d/llm-d-router/pull/1618

hexfusion:
this test will fail until #1618 merges

hexfusion:
cc @capri-xiyuenits thanks for review, nits have been updated now that https://github.com/llm-d/llm-d-router/pull/1618 has merge

hexfusion:
cc @rahulgurnani

rahulgurnani:
look good to me, thanks for adding these!

ahg-g (approved):

PR #1567, opened 2026-06-09
Fix race condition in logger test

nicolexin:
**What type of PR is this?**

/kind test

This is a race condition I discovered in the GAIE cleanup: https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2959. Both TestLogger and TestRefreshPrometheusMetricsAvgValues interact with a global Prometheus metrics registry and we didn't fully wait for the metrics thread to shutdown before starting the next test.

None

```release-note
NONE
```

### DIFF
--- a/pkg/epp/datalayer/logger/logger_test.go
+++ b/pkg/epp/datalayer/logger/logger_test.go
@@ -19,6 +19,7 @@ package logger
 import (
 	"bytes"
 	"context"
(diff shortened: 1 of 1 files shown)


nicolexin:
The test failure seems unrelated to this change.

liu-cong (approved):

PR #1572, opened 2026-06-10
ci: propagate HF_TOKEN to e2e tests to avoid HF Hub rate limits

### LINKED ISSUE #1527: CI: Hugging Face Hub 429 rate-limit e2e flake
**What's happening**:

`ci-pr-checks` runs are failing on transient `huggingface.co` 429s during model file-listing API calls from the `vllm-render` container. Across the last ~36 hours, **8 of the last 30 failed runs** carry this signature:

```
ERROR ... 429 Too Many Requests for url:
  https://huggingface.co/api/models/Qwen/Qwen2.5-1.5B-Instruct/tree/main?recursive=true&expand=false.
  retrying 1 of 2
ERROR ... 429 Too Many Requests (final retry, gives up)
ERROR ... Error retrieving file list. Returning an empty list.

→ pydantic_core._pydantic_core.ValidationError: 1 validation error for ModelConfig
  Value error, Invalid repository ID or local directory specified: 'Qwen/Qwen2.5-1.5B-Instruct'.
```

The pydantic error is misleading: the model repo exists; HF Hub is just rate-limited and vLLM falls back to "no config found." Root cause is GitHub Actions runners sharing public IPs that hit the HF Hub anonymous quota. Same shape as the docker.io flake tracked in #1525, different registry.

The `e2e-router` matrix (suites: `extended`, `disruption`, `pd-shared-storage-disagg`, `pd-metrics`) and `e2e-gaie` are the visible victims; any job that spins up the real `vllm-render` container is at risk.

**Reproduce the list** (re-runnable, so the report doesn't go stale):

```bash
gh run list -R llm-d/llm-d-router -w ci-pr-checks.yaml -s failure -L 30 \
  --json databaseId,headBranch -q '.[] | "\(.databaseId)\t\(.headBranch)"' \
| while read RUN BRANCH; do
    for JOB in $(gh api "repos/llm-d/llm-d-router/actions/runs/$RUN/jobs" \
        --jq '.jobs[] | select(.conclusion=="failure") | .id'); do
      gh api "repos/llm-d/llm-d-router/actions/jobs/$JOB/logs" \
        -H 'Accept: application/vnd.github.raw' 2>/dev/null \
        | grep -q '429 Too Many Requests.*huggingface.co' \
        && { echo "https://github.com/llm-d/llm-d-router/actions/runs/$RUN ($BRANCH)"; break; }
    done
  done
```

Affected branches at the time of filing: `session`, `epd_connector`, `fix/model-rewrite-missing-field`, `refine-ttft-tpot-metrics`, `implement-on-top-of-1448`, `fix/tokenizer-protocol-coverage-and-subpath-parser`, `main`.

**Why this matters**:

Combined with #1525 (Docker Hub anonymous rate limit), anonymous public-registry quotas now account for a large fraction of `ci-pr-checks` failures. Re-runs + reviewer time burned per PR.

**Proposed mitigations**:

- **Authenticated HF Hub access.** Add an `HF_TOKEN` repo secret and propagate `HF_TOKEN: ${{ secrets.HF_TOKEN }}` into the e2e job env. HF Hub's anonymous quota is substantially stricter than even a free authenticated account's.
- **Pre-cached model artifacts in the e2e image** with `HF_HUB_OFFLINE=1` set on the `vllm-render` container. Each test starts from a local cache; no `hf.co` API calls at all.
- **Retry + backoff layer.** Wrap the `vllm-render` startup in retry-with-backoff so a single 429 doesn't fail the whole job.

Jwrede:
## Summary
- Propagate `HF_TOKEN` from a GitHub repo secret into e2e test containers so vllm-render and model-existence checks authenticate against HF Hub
- Authenticated requests have a substantially higher rate limit than anonymous ones, addressing the 429 flakes described in #1527
- If the secret is not configured, behavior is unchanged (empty token, anonymous requests)

### Changes
- **Makefile**: add `HF_TOKEN` to `E2E_ENV_VARS` forwarding list
- **test/e2e/setup_test.go**: read `HF_TOKEN` from env instead of hardcoding empty string
- **test/e2e/utils_test.go**: attach Bearer token header to the `isModelReal` HF API call
- **ci-pr-checks.yaml**: pass `secrets.HF_TOKEN` in both `e2e-gaie` and `e2e-router` job steps

### Prerequisites
A repo admin must create the `HF_TOKEN` repository secret with a valid HuggingFace access token (a free account token is sufficient for the models used in CI).

Closes #1527

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -363,6 +363,7 @@ jobs:
           E2E_LABEL_FILTER: ${{ matrix.suite.label-filter }}
           PULL_SIDECAR_IMAGE: "false"
           PULL_VLLM_RENDER_IMAGE: "false"
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: make test-e2e-gaie-run
 
   e2e-router:
@@ -450,4 +451,5 @@ jobs:
           E2E_LABEL_FILTER: ${{ matrix.suite.label-filter }}
           LOAD_VLLM_RENDER_IMAGE: ${{ matrix.suite.needs-renderer }}
           PULL_VLLM_RENDER_IMAGE: "false"
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: make test-e2e-router-run
--- a/Makefile
+++ b/Makefile
@@ -122,7 +122,7 @@ endif
 # Should we pass ALL env vars here?
 E2E_ENV_VARS = EPP_IMAGE VLLM_IMAGE SIDECAR_IMAGE VLLM_RENDER_IMAGE \
                E2E_KEEP_CLUSTER_ON_FAILURE E2E_PORT E2E_METRICS_PORT K8S_CONTEXT READY_TIMEOUT \
-               E2E_LABEL_FILTER LOAD_VLLM_RENDER_IMAGE
+               E2E_LABEL_FILTER LOAD_VLLM_RENDER_IMAGE HF_TOKEN
 BUILDER_E2E_ENV_FLAGS = $(foreach v,$(E2E_ENV_VARS),$(if $($(v)),-e '$(v)=$($(v))'))
 ifneq ($(filter command line environment,$(origin NAMESPACE)),)
 BUILDER_E2E_ENV_FLAGS += -e NAMESPACE=$(NAMESPACE)
--- a/test/e2e/setup_test.go
+++ b/test/e2e/setup_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"strconv"
 	"time"
 
@@ -31,7 +32,7 @@ func createModelServersFromKustomize(kustomizeDir string, extra map[string]strin
 		"${DECODE_ROLE}":             "",
 		"${EPP_NAME}":                "e2e-epp",
 		"${NAMESPACE}":               nsName,
-		"${HF_TOKEN}":                "",
+		"${HF_TOKEN}":                os.Getenv("HF_TOKEN"),
 		"${VLLM_EXTRA_ARGS_E}":       "",
 		"${VLLM_EXTRA_ARGS_P}":       "",
 		"${VLLM_EXTRA_ARGS_D}":       "",
--- a/test/e2e/utils_test.go
+++ b/test/e2e/utils_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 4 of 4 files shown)


elevran (commented):
Where is the original HF_TOKEN coming from (e.g., who sets the repo secret, is it a personal or org owned token)? 
Do we know when and how to update it before expiration?

Jwrede:
The PR only wires up the plumbing for a `HF_TOKEN` repo secret. It does not create or manage the token itself.

HuggingFace supports two token types for this:
- **Fine-grained tokens** with read-only scope on public models, which can have an expiration policy set at creation time.
- **User access tokens** (non-expiring), scoped to a user account.

For a CI context, a fine-grained token with read-only public repo access would be the best fit. It can be created under an org-owned HF account (or a dedicated service account) and stored as a GitHub repo secret by anyone with repo admin access.

If no secret is configured, the behavior is unchanged: all HF API calls remain anonymous. The token only raises the rate limit ceiling.

ahg-g:
> The PR only wires up the plumbing for a `HF_TOKEN` repo secret. It does not create or manage the token itself.
> 
> HuggingFace supports two token types for this:
> 
> * **Fine-grained tokens** with read-only scope on public models, which can have an expiration policy set at creation time.
> * **User access tokens** (non-expiring), scoped to a user account.
> 
> For a CI context, a fine-grained token with read-only public repo access would be the best fit. It can be created under an org-owned HF account (or a dedicated service account) and stored as a GitHub repo secret by anyone with repo admin access.
> 
> If no secret is configured, the behavior is unchanged: all HF API calls remain anonymous. The token only raises the rate limit ceiling.

we need to do this before merging this PR, otherwise ci will fail now that it looks up for the HF token, right?

Jwrede:
No, CI will not fail without the secret. The lookup is guarded: in `isModelReal` the `Authorization` header is only set when `HF_TOKEN` is non-empty, so when the secret is unconfigured the value is an empty string and the request stays anonymous, which is identical to the previous behavior (the kustomize substitution was hardcoded to `""` before this PR). The token only raises the anonymous rate-limit ceiling when present. So this is safe to merge as-is, and adding the repo secret can be done at any time afterward as a pure improvement, with no migration or expiry coupling to the merge.

ahg-g:
thanks @Jwrede pls address the conflict

Jwrede:
Rebased onto `main` and resolved the conflict in `.github/workflows/ci-pr-checks.yaml`: kept the `HF_TOKEN` env var and took main's renamed `make test-e2e-router-run` target. No other changes.

elevran (approved):

PR #1579, opened 2026-06-10
Pin :latest image tags and enable strict presubmit check

### LINKED ISSUE #1568: check-latest-tags: flip presubmit to strict mode and fix column-0 edge case
/kind cleanup

**What this tracks:**

Follow-up from #1457 (per [elevran's review](https://github.com/llm-d/llm-d-router/pull/1457#pullrequestreview-2973498531)):

1. **Switch presubmit to strict mode.** The `presubmit` target currently runs `check-latest-tags` in warn-only mode, so `:latest` regressions still pass CI. Once the existing GPU manifests (`config/manifests/vllm/gpu-deployment.yaml`, `gpu-grpc-deployment.yaml`) are pinned to specific versions, flip presubmit to use `check-latest-tags-strict`.

2. **Handle `image:` at column 0.** The grep pattern `[[:space:]]image:` requires leading whitespace, so a top-level `image:` key at column 0 would be missed. Uncommon in Kubernetes YAML but worth fixing — e.g. change the pattern to `(^|[[:space:]])image:` or equivalent.

**Related PR:** #1457

Jwrede:
**What type of PR is this?**
/kind cleanup

Follow-up from #1457 (per elevran's review).

1. Pins GPU manifest images to specific versions:
   - `vllm/vllm-openai:latest` to `vllm/vllm-openai:v0.21.0` (4 manifests)
   - `lmsysorg/sglang:latest` to `lmsysorg/sglang:v0.5.12` (1 manifest)
2. Flips the `presubmit` target from `check-latest-tags` (warn-only) to `check-latest-tags-strict` so `:latest` regressions block merge.
3. Fixes the grep pattern in `check-latest-tags.sh` to detect `image:` at column 0 (the previous pattern required leading whitespace).

Resolves #1568

No.

**How has this been tested?**

Ran `./scripts/check-latest-tags.sh` in strict mode and confirmed zero violations. Verified column 0 detection with a test YAML file containing `image: foo/bar:latest` at column 0.

### DIFF
--- a/scripts/check-latest-tags.sh
+++ b/scripts/check-latest-tags.sh
@@ -14,9 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Check that YAML files do not use the ':latest' image tag.
-# Container images should be pinned to a specific version so that builds
-# and tests are reproducible.
+# Check that external container images in YAML files do not use the ':latest'
+# tag. Images for components outside the llm-d project are pinned to a specific
+# version so that builds and tests are reproducible and not broken by upstream
+# API or behavior changes. Images owned by the llm-d project are allowed to
+# track ':latest' on main so cross-component regressions surface early.
 #
 # Usage:
 #   ./scripts/check-latest-tags.sh [--warn] [DIR ...]
@@ -31,6 +33,9 @@ set -o nounset
 set -o pipefail
 
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+
+# Images under these llm-d-owned registries may use ':latest'.
+OWNED_IMAGE_RE='image:[[:space:]]*"?(ghcr\.io|quay\.io)/llm-d/'
 
 WARN_ONLY=false
 
@@ -70,6 +75,7 @@ for dir in "${SCAN_DIRS[@]}"; do
     | grep -v ':latest-' \
     | grep -v 'description:' \
     | grep -v '<your-registry>' \
+    | grep -vE "$OWNED_IMAGE_RE" \
     | awk -F: '{content = substr($0, index($0,$3)); if (content !~ /^[[:space:]]*#/) print}' \
     || true)"
 
@@ -79,16 +85,16 @@ for dir in "${SCAN_DIRS[@]}"; do
 done
 
 if [[ -z "$violations" ]]; then
-  echo "No ':latest' image tags found in YAML files."
+  echo "No ':latest' image tags found for external images in YAML files."
   exit 0
 fi
 
 if [[ "$WARN_ONLY" == true ]]; then
-  echo "WARNING: The following YAML files use the ':latest' image tag."
+  echo "WARNING: The following YAML files use the ':latest' tag for an external image."
 else
-  echo "ERROR: The following YAML files use the ':latest' image tag."
+  echo "ERROR: The following YAML files use the ':latest' tag for an external image."
 fi
-echo "Pin images to a specific version for reproducible builds."
+echo "Pin external images to a specific version for reproducible builds."
 echo ""
 echo "$violations"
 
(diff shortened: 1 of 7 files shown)


scripts/check-latest-tags.sh
```diff
@@ -58,14 +58,15 @@ for dir in "${SCAN_DIRS[@]}"; do
...
+  # whitespace, or column 0 after the grep line-number prefix) to
+  # avoid substring hits in unrelated fields.
   matches="$(find "$dir" \
     -path '*/.git' -prune -o \
     -path '*/vendor' -prune -o \
     -path '*/node_modules' -prune -o \
     -type f \( -name '*.yaml' -o -name '*.yml' \) -print0 \
     | xargs -0 grep -Hn ':latest' \
-    | grep '[[:space:]]image:' \
+    | grep -E '([[:space:]]image:|:[0-9]+:image:)' \
```
ahg-g, line 74: so one thing I think we should do is use latest image on main but released ones on the release branches. This helps us catch regressions early in integrations with other components, especially other llm-d components like the latency predictor.

elevran: I'm split on that. We've had multiple intgeration and e2e test failures when a dependency changed APIs or behaviors. Each caused 1-2d to stabilize again.
In order to help early integration and regressiongs, we may want to have a separate test run (e.g., nighly) with latest versions. I would prefer to keep the PR CICD fixed on well known tags.

ahg-g: But isn't the purpose of CI/CD to catch such regressions and flag them?

ahg-g: Fixing the tags of llm-d external components (like vllm or sglang) may be ok since we don't have control over them, but for llm-d dependencies, I prefer we use latest on the main branch to catch regressions early

elevran: > But isn't the purpose of CI/CD to catch such regressions and flag them?

The downside of not fixing llm-d tags is that someone pushes a PR that fails due to a reason unrelated to the code change. It is often up to maintainers to detect these and chase them down.

Agree it is a delicate balance, hence the suggestion to run regressions as part of a special action, not as part of normal PR checks.
Open to implement either way.

Jwrede:
Thanks both. A few observations from the PR as it stands:

The only tags this PR pins are the external ones we do not control: `vllm/vllm-openai` and `lmsysorg/sglang`. The llm-d-owned images in the repo (`llm-d-inference-sim`, `llm-d-router-endpoint-picker`) are already pinned to specific versions on main today, so nothing here forces llm-d components onto fixed tags.

That lines up with @ahg-g's split: pin what we do not control, keep flexibility for llm-d components. The one place the PR overreaches is the strict presubmit gate, which forbids `:latest` for any image, including future llm-d ones.

I can make the check enforce exactly that distinction instead: fail on `:latest` for external images, allow it for llm-d-owned images (`ghcr.io/llm-d/*`). External deps stay pinned in PR CI, which addresses the unrelated-failure problem @elevran raised since those breakages came from external API changes, while llm-d components stay free to track latest on main to catch regressions early. The nightly-latest run for llm-d components sounds worthwhile too, but probably as its own follow-up PR.

Happy to update the check to the allowlist approach, or keep it strict and pin everything. Let me know which direction you prefer and I will adjust.

elevran:
The unrelated failures were also triggered by internal llm-d components (tokenizer, simulator), not only the external components.
We can keep the check lax on internal components for now, and once the nightly regression action is done in a separate PR, flip that as well.

Jwrede:
Done in df76c955. The check now allows `:latest` for llm-d-owned images (`ghcr.io/llm-d/*`, `quay.io/llm-d/*`) and fails only on external `:latest`, so external deps stay pinned in PR CI while internal components can track latest on main. Flipping internal to strict once the nightly regression action lands sounds like the right follow-up. The external pins in this PR (vllm, sglang) are unchanged.

elevran (approved):

PR #1580, opened 2026-06-10
fix(concurrency-detector): report zero saturation for endpoints with nil metadata

### LINKED ISSUE #1495: concurrency-detector: New pods reports saturation as 1

Jwrede:
**What type of PR is this?**
/kind bug

The concurrency-detector's `Saturation()` method skipped endpoints with nil metadata, excluding them from both the inflight and capacity totals. When all endpoints lacked metadata (freshly provisioned pods or passive HA pods), `totalCapacity` stayed zero and the method returned 1.0, incorrectly signalling full saturation.

This caused autoscaling events to trigger immediately after cluster provisioning, and the passive pod in an HA setup permanently reported saturation as 1 instead of 0.

The fix moves capacity accumulation before the nil-metadata guard. Every endpoint in the candidate list contributes its configured max concurrency to the denominator. Endpoints with nil metadata contribute zero inflight load to the numerator. An empty candidate list still returns 1.0 (fail-closed).

The utilization-detector does not have this problem because it reads metrics from the model servers directly (which always report their resource usage), while the concurrency-detector relies on in-flight load data that only gets populated after the EPP processes traffic.

Resolves #1495

Yes. Freshly provisioned pods and passive HA pods now report saturation as 0 instead of 1 when the concurrency-detector plugin is active.

**How has this been tested?**

Added `TestDetector_NilMetadataEndpoint` covering:
- All endpoints with nil metadata returns saturation 0.0
- Mixed nil metadata and loaded endpoints returns correct ratio (e.g., 10 inflight / 20 capacity = 0.5)

All existing tests continue to pass.

### DIFF
--- a/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go
+++ b/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go
@@ -137,6 +137,8 @@ func (d *detector) Saturation(_ context.Context, endpoints []datalayer.Endpoint)
 			totalCapacity += d.config.maxConcurrency
 		}
 
+		// nil metadata = endpoint not yet populated by this EPP instance
+		// (fresh or passive); counts as idle capacity with zero in-flight load.
 		if e.GetMetadata() == nil {
 			continue
 		}
@@ -148,6 +150,10 @@ func (d *detector) Saturation(_ context.Context, endpoints []datalayer.Endpoint)
 		} else {
 			totalInflight += load.Requests
 		}
+	}
+
+	if totalCapacity == 0 {
+		return 1.0
 	}
 
 	return float64(totalInflight) / float64(totalCapacity)
--- a/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go
+++ b/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go
@@ -125,8 +125,18 @@ func (d *detector) getLoad(m datalayer.AttributeMap) *attrconcurrency.InFlightLo
 //
 //	Saturation = Total Inflight Requests / Total MaxConcurrency Capacity.
 func (d *detector) Saturation(_ context.Context, endpoints []datalayer.Endpoint) float64 {
+	if len(endpoints) == 0 {
+		return 1.0
+	}
+
 	var totalInflight, totalCapacity int64
 	for _, e := range endpoints {
+		if d.config.mode == modeTokens {
+			totalCapacity += d.config.maxTokenConcurrency
+		} else {
+			totalCapacity += d.config.maxConcurrency
+		}
+
 		if e.GetMetadata() == nil {
 			continue
 		}
@@ -135,15 +145,9 @@ func (d *detector) Saturation(_ context.Context, endpoints []datalayer.Endpoint)
 
 		if d.config.mode == modeTokens {
 			totalInflight += load.Tokens
-			totalCapacity += d.config.maxTokenConcurrency
 		} else {
 			totalInflight += load.Requests
-			totalCapacity += d.config.maxConcurrency
 		}
-	}
-
-	if totalCapacity == 0 {
-		return 1.0
 	}
 
 	return float64(totalInflight) / float64(totalCapacity)
--- a/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector_test.go
+++ b/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector_test.go
@@ -534,6 +534,34 @@ func TestDetector_ConcurrencyStress(t *testing.T) {
 	require.Equal(t, int64(0), reg.get(fullID).Requests)
 }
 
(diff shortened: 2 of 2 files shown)


pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go
```diff
@@ -135,17 +145,11 @@ func (d *detector) Saturation(_ context.Context, endpoints []datalayer.Endpoint)
...
 			totalInflight += load.Requests
-			totalCapacity += d.config.maxConcurrency
 		}
 	}
 
-	if totalCapacity == 0 {
-		return 1.0
-	}
-
 	return float64(totalInflight) / float64(totalCapacity)
```
LukeAVanDrie, line 159: Non-blocking nit: the `totalCapacity == 0` guard you removed was covering two cases: the empty candidate list (now handled by the new `len(endpoints) == 0` check at the top) and a zero `maxConcurrency/maxTokenConcurrency`.

The latter is prevented in practice by `validateConfig`, so via the factory path this genuinely cannot divide by zero. The only residual path is direct `newDetector(config{maxConcurrency: 0}, …)` construction (test-only), which would produce `0/0 = NaN`.

If you would like to be overly defensive for internal callers, a `if totalCapacity == 0 { return 1.0 }` just before the return restores the original fail-closed contract cheaply. Equally fine to leave it; your call.

pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go
```diff
@@ -125,8 +125,18 @@ func (d *detector) getLoad(m datalayer.AttributeMap) *attrconcurrency.InFlightLo
...
+
 	var totalInflight, totalCapacity int64
 	for _, e := range endpoints {
+		if d.config.mode == modeTokens {
+			totalCapacity += d.config.maxTokenConcurrency
+		} else {
+			totalCapacity += d.config.maxConcurrency
+		}
+
 		if e.GetMetadata() == nil {
```
LukeAVanDrie, line 142: Could you add a one-liner here on the intended semantics? The fix correctly reads nil metadata as "endpoint exists with zero locally-tracked in-flight load" (counts toward capacity, contributes 0 to load) rather than "unknown implies skip". This matches how `Filter` already treats absent load as available. But it's the inverse of the `utilization-detector`'s missing-data policy (which fail-closes to 1.0), so it's non-obvious. Something like:

```
// nil metadata = endpoint not yet populated by this EPP instance (fresh/passive endpoint);
// counts as idle capacity with zero in-flight load, not skipped.
```

would save the next reader the archaeology.

LukeAVanDrie (dismissed):
Nice fix! Thanks for the clear writeup, it made this easy to follow.

I traced it back through #1474 → #1478 → #1495 and checked it against the `utilization-detector` to make sure it converges rather than just patching locally. It holds up:

- Passive/fresh pods now report 0 instead of 1. In active/passive HA the active EPP dispatches all traffic, so its local in-flight view is the true pool total, and the passive contributes a clean 0. So, KEDA sum-aggregation works without the 1.8-style threshold fudge @asharkhan3101 was forced into.
- It converges with the `utilization-detector` in the case the issue calls out (idle passive endpoint implies 0 from both). The aggregation models even line up: `Σinflight / (N·maxConcurrency)` is the mean of per-endpoint ratios which is the same shape as `utilization-detector`'s `Σscore_i / N`.
- It also repairs two pre-existing internal contradictions. A nil-metadata endpoint now behaves identically to a known-but-idle one (`single_endpoint_empty/unknown_endpoint_assumed_empty` already returned 0.0), and `Saturation` now agrees with `Filter`, which already treated absent load as available via `getLoad`'s zero default. Previously `Saturation` excluded those endpoints while `Filter` kept them.

As an optional followup (separate PR), it's worth a note in the concurrency README / KEDA guide that concurrency saturation reflects this EPP's locally-dispatched load. It's exactly right for active/passive, but in active/active the in-flight view is fragmented across replicas, so no single EPP sees the true pool total. Operators should aggregate across endpoints. This is the deliberate inverse of the `utilization-detector`, which reads server-scraped metrics and therefore fail-closes on missing data rather than treating it as idle.

ahg-g:
This is critical for this release, pls prioritize merging by end of day Friday.

ahg-g (approved):

PR #1582, opened 2026-06-10
Add explicit token-producer backend selector

### LINKED ISSUE #1472: token-producer: explicit backend selector field
_Claude Assisted_

Spun out of #1445. The token-producer infers its backend from which config block is set (`estimate`/`vllm`/`udsTokenizerConfig`), with `estimate` as the zero-config default. An explicit `backend` enum is better UX since each backend may have non-empty defaults and it makes the default backend explicit.

Backwards-compatible: an optional `backend: estimate|vllm|uds` field that takes precedence when set, with the current deduction kept as the fallback for existing configs (which have no `backend` field).

cc @ahg-g

zhouyou9505:
## Summary

- add an optional `backend` selector for the token-producer plugin
- support explicit `estimate`, `vllm`, and `uds` backend selection while preserving legacy inference when omitted
- update token-producer docs, chart docs, sample configs, and e2e config fixtures to show explicit backend selection

Fixes #1472

## Validation

Local curl smoke testing with a temporary HTTP harness that loads token-producer through `PluginFactory`:

- `backend: estimate`: `POST /v1/completions` returned estimated token IDs from the in-process backend
- `backend: vllm`: `POST /v1/completions` with a local fake vLLM `/render` server returned `token_ids: [101,102,103]`
- `backend: uds`: `POST /v1/completions` with a local fake gRPC Unix socket tokenizer returned `token_ids: [201,202,203]`


pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
sagearc: Since `{"backend":"estimate","vllm":{}}` is now accepted this comment is not longer true

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -58,13 +58,23 @@ const (
...
-// backend; `udsTokenizerConfig` selects the deprecated gRPC-over-UDS backend;
-// `estimate` selects the tokenizer-free byte-packing backend, which is also the
-// zero-config default when no backend is set.
+// Backend selection: `backend` explicitly selects the backend when set. When it
+// is omitted, existing configs continue to infer the backend from `estimate`,
+// `vllm`, `udsTokenizerConfig`, or `modelName`.
 type tokenizerPluginConfig struct {
+	// Backend explicitly selects "estimate", "vllm", or "uds". When omitted,
+	// the plugin infers the backend from legacy config fields.
+	Backend tokenizerBackend `json:"backend,omitempty"`
```
sagearc: nit: duplicated explanation of the inferred backend path

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -58,13 +58,23 @@ const (
...
-// Backend selection: `vllm` or `modelName` selects the vLLM HTTP /render
-// backend; `udsTokenizerConfig` selects the deprecated gRPC-over-UDS backend;
-// `estimate` selects the tokenizer-free byte-packing backend, which is also the
-// zero-config default when no backend is set.
+// Backend selection: `backend` explicitly selects the backend when set. When it
+// is omitted, existing configs continue to infer the backend from `estimate`,
+// `vllm`, `udsTokenizerConfig`, or `modelName`.
 type tokenizerPluginConfig struct {
+	// Backend explicitly selects "estimate", "vllm", or "uds". When omitted,
+	// the plugin infers the backend from legacy config fields.
```
sagearc: The config fields themselves are not legacy, only the implicit inference logic, right?

sagearc (changes requested):
Thanks @zhouyou9505! Looks good to me overall, just left a few minor comments.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
zhouyou9505: Correct. Got it. The fields themselves are not legacy. Only the implicit backend inference behavior is legacy and kept for backward compatibility. I will refine the description.

zhouyou9505:
@sagearc Hi, I’ve addressed the review feedback and pushed the changes. Could you please help review again? Thanks!

zhouyou9505:
@sagearc @vMaroon Hi, I’ve addressed the review feedback and pushed the changes. Could you please help review again? Thanks!

sagearc:
Thanks @zhouyou9505

sagearc (dismissed):

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -128,29 +136,67 @@ func PluginFactory(name string, rawParameters *json.Decoder, handle plugin.Handl
...
+		switch config.Backend {
+		case tokenizerBackendEstimate, tokenizerBackendVLLM, tokenizerBackendUDS:
+			return config.Backend, nil
+		default:
+			return "", fmt.Errorf("invalid configuration for '%s' plugin: backend must be one of %q, %q, or %q", PluginType, tokenizerBackendEstimate, tokenizerBackendVLLM, tokenizerBackendUDS)
+		}
 	}
 
-	return p, nil
+	estimate := config.Estimate != nil
```
ahg-g, line 186: The point of having an explicit parameter for selecting the backend is to remove this guessing logic. it is hard to do this now without breaking existing deployments. I think the best next step is to remove TokenizerConfig (it is deprecated already) and make make VLLM and Estimate a oneof during validation.

zhouyou9505: Thanks, understood. I’ll remove the deprecated UDS configuration and replace the fallback inference with validation requiring exactly one of vllm or estimate. This means empty configurations will no longer implicitly select estimate. I’ll update the tests and documentation accordingly.

zhouyou9505:
Thanks for the recommendation. I removed the deprecated `TokenizerConfig`/UDS backend and the separate `backend` selector, and changed `vllm` and `estimate` to an exactly-one validation rule.

One compatibility concern is that this change may cause existing deployments using previously accepted configurations to fail during startup.

The resulting compatibility behavior is:

| Previous configuration | Behavior now |
| --- | --- |
| `modelName` + `vllm: {}` | Valid |
| `estimate: {}` | Valid |
| `parameters` omitted entirely | Valid; defaults to `estimate` |
| `parameters: {}` | Invalid; startup fails because an explicit backend variant is required |
| `modelName` only | Invalid; startup fails because it no longer implicitly selects vLLM |
| `backend: vllm` or `backend: estimate` | Invalid; startup fails because `backend` has been removed |
| `udsTokenizerConfig` | Invalid; startup fails because the deprecated UDS backend has been removed |
| Both `vllm` and `estimate` | Invalid |

The omitted-parameters case remains valid for the framework's automatically created token producer, which still uses the estimate backend. Explicitly configured parameter objects must select exactly one variant.

Does this compatibility break match what you had in mind, or should we preserve a transition path for previously accepted configurations?  @ahg-g 

Ultimately, only the following two explicit forms are recommended
```
# Estimate
- type: token-producer
  parameters:
    estimate: {}

# vLLM
- type: token-producer
  parameters:
    modelName: Qwen/Qwen3-8B
    vllm:
      url: http://localhost:8000
```

PR #1585, opened 2026-06-10
fix: strip query parameters from request path before parser resolution

Gregory-Pereira:
## Summary

- Strip query parameters from the `:path` HTTP/2 pseudo-header at header ingestion in `HandleRequestHeaders`, so all downstream consumers (parser registry, per-parser endpoint validation) see a clean path
- Add defense-in-depth stripping in `GetRequestPath` for any code path that reads path headers directly
- Fixes parser lookup failure and anthropic-parser endpoint validation rejection when clients append query strings (e.g. Claude Code sends `POST /v1/messages?beta=true`)

Closes #TBD

## Test plan

- [ ] New unit tests for `HandleRequestHeaders` verify `:path` query stripping
- [ ] New unit tests for `GetRequestPath` verify stripping across all header sources (`:path`, `x-original-path`, `x-forwarded-path`)
- [ ] Existing parser registry and handler tests continue to pass
- [ ] Validated end-to-end with Claude Code against `quay.io/grpereir/llm-d-router-endpoint-picker:fix-query-string-v3` — requests to `/v1/messages?beta=true` now route correctly through the anthropic-parser

### DIFF
--- a/pkg/epp/handlers/request.go
+++ b/pkg/epp/handlers/request.go
@@ -50,6 +50,9 @@ func (s *StreamingServer) HandleRequestHeaders(ctx context.Context, reqCtx *Requ
 	for _, header := range req.RequestHeaders.Headers.Headers {
 		key := strings.ToLower(header.Key)
 		value := envoy.GetHeaderValue(header)
+		// The HTTP/2 :path pseudo-header includes query parameters (e.g.
+		// /v1/messages?beta=true). Strip them so parser suffix matching and
+		// per-parser endpoint validation see only the path component.
 		if key == ":path" {
 			value, _, _ = strings.Cut(value, "?")
 		}
--- a/pkg/epp/handlers/request.go
+++ b/pkg/epp/handlers/request.go
@@ -48,15 +48,7 @@ func (s *StreamingServer) HandleRequestHeaders(ctx context.Context, reqCtx *Requ
 	}
 
 	for _, header := range req.RequestHeaders.Headers.Headers {
-		key := strings.ToLower(header.Key)
-		value := envoy.GetHeaderValue(header)
-		// The HTTP/2 :path pseudo-header includes query parameters (e.g.
-		// /v1/messages?beta=true). Strip them so parser suffix matching and
-		// per-parser endpoint validation see only the path component.
-		if key == ":path" {
-			value, _, _ = strings.Cut(value, "?")
-		}
-		reqCtx.Request.Headers[key] = value
+		reqCtx.Request.Headers[strings.ToLower(header.Key)] = envoy.GetHeaderValue(header)
 	}
 
 	reqCtx.ObjectiveKey, _ = metadata.GetLowerCaseHeaderValue(reqCtx.Request.Headers, metadata.ObjectiveKey)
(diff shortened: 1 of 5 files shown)


Gregory-Pereira:
Without this:

```console
❯ is this working?
  ⎿  API Error: 400 inference error: unsupported API endpoint: /v1/messages?beta=true
```

With this:

```console
ok hope this works

  Thought for 5s

I'm not sure what you'd like me to do — your message doesn't specify a task. Could you clarify what you'd like me to help with?

For example:
- Run tests in this directory?
- Review the current changes?
- Something else related to the agentic guide?
```

pkg/epp/handlers/request.go
```diff
@@ -48,7 +48,12 @@ func (s *StreamingServer) HandleRequestHeaders(ctx context.Context, reqCtx *Requ
 	}
 
 	for _, header := range req.RequestHeaders.Headers.Headers {
-		reqCtx.Request.Headers[strings.ToLower(header.Key)] = envoy.GetHeaderValue(header)
+		key := strings.ToLower(header.Key)
```
tessapham: Can we add a comment to explain?

pkg/epp/handlers/request.go
```diff
@@ -48,7 +48,12 @@ func (s *StreamingServer) HandleRequestHeaders(ctx context.Context, reqCtx *Requ
...
+		value := envoy.GetHeaderValue(header)
+		if key == ":path" {
```
tessapham: Is `:path` the only header query params need to be stripped from?

tessapham (approved):
Just a clarification question. Otherwise /lgtm.

pkg/epp/handlers/request.go
```diff
@@ -48,7 +48,15 @@ func (s *StreamingServer) HandleRequestHeaders(ctx context.Context, reqCtx *Requ
...
-		reqCtx.Request.Headers[strings.ToLower(header.Key)] = envoy.GetHeaderValue(header)
+		key := strings.ToLower(header.Key)
+		value := envoy.GetHeaderValue(header)
+		// The HTTP/2 :path pseudo-header includes query parameters (e.g.
+		// /v1/messages?beta=true). Strip them so parser suffix matching and
+		// per-parser endpoint validation see only the path component.
+		if key == ":path" {
+			value, _, _ = strings.Cut(value, "?")
+		}
+		reqCtx.Request.Headers[key] = value
```
zetxqx: Change it here means we'll mutate the original headers, and finally we'll send the mutated header to the envoy proxy and then to the model server. https://github.com/llm-d/llm-d-router/blob/dcb52be80d5c5cac3e299b162b11886e48792174/pkg/epp/handlers/request.go#L93

Do we really want this? or just changing `GetRequestPath`  the parser called is enough?

zetxqx: I think what we need to do is for anthropic parser we also use the `reqeust.GetRequestPath` instead of the following:

https://github.com/llm-d/llm-d-router/blob/dcb52be80d5c5cac3e299b162b11886e48792174/pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go#L82

can you help change this?

zetxqx (approved):
looks good thanks for catching this

pkg/epp/handlers/request.go
Gregory-Pereira: My understanding is that this was the case — :path is the only HTTP/2 pseudo-header that carries query parameters based on RFC 9113: https://www.rfc-editor.org/info/rfc9113/. However this is no longer applicable here

ahg-g (approved):

PR #1088, opened 2026-05-13
feat(parsers): support Anthropic `v1/messages` API

tessapham:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #943.

**Release note** _(write `NONE` if no user-facing change)_:
```release-note
NONE
```

PR #1587, opened 2026-06-10
routing sidecar support /v1/messages for pd

Gregory-Pereira:
**What type of PR is this?**
/kind bug
/kind feature

Sits somewhere between the two, follows on after #1088 

This will register the /v1/messages API in the sidecar for pd-support.

cc @tessapham @hexfusion @vMaroon

### DIFF
--- a/pkg/sidecar/proxy/proxy.go
+++ b/pkg/sidecar/proxy/proxy.go
@@ -424,6 +424,7 @@ func (s *Server) createRoutes() *http.ServeMux {
 	})
 	mux.HandleFunc("POST "+ChatCompletionsPath, s.disaggregatedPrefillHandler(APITypeChatCompletions))
 	mux.HandleFunc("POST "+CompletionsPath, s.disaggregatedPrefillHandler(APITypeChatCompletions))
+	mux.HandleFunc("POST "+MessagesPath, s.disaggregatedPrefillHandler(APITypeChatCompletions))
 	mux.HandleFunc("POST "+ResponsesPath, s.disaggregatedPrefillHandler(APITypeResponses))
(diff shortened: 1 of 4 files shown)


tessapham (approved):
LGTM. Good catch!

pkg/sidecar/proxy/proxy.go
```diff
@@ -424,6 +424,7 @@ func (s *Server) createRoutes() *http.ServeMux {
 	})
 	mux.HandleFunc("POST "+ChatCompletionsPath, s.disaggregatedPrefillHandler(APITypeChatCompletions))
 	mux.HandleFunc("POST "+CompletionsPath, s.disaggregatedPrefillHandler(APITypeChatCompletions))
+	mux.HandleFunc("POST "+MessagesPath, s.disaggregatedPrefillHandler(APITypeChatCompletions))
```
ahg-g, line 427: I think it is better to define a new APIType that represents messages, otherwise logs and traces will be misleading/confusing for message api requests

shmuelk: +1 An investigation of the code shows that this will be a local change  here in proxy.go

ahg-g: @tessapham would you like this merged and you followup with a PR to fix this?

tessapham: Yes I'll open a separate PR to add a new APIType for Messages. This can go in to unblock agentic testing.

ahg-g (approved):

pkg/sidecar/proxy/proxy.go
ahg-g: thanks, merged.

PR #1589, opened 2026-06-11
fix(sidecar): take effect mooncake bootstrap port configured in yaml

weizhoublue:
**What type of PR is this?**
 
/kind bug
 

Struct options.go defines JSON tag for  mooncake-bootstrap-port , 

``` yaml
type yamlConfiguration struct {
	Port                           int      `json:"port,omitempty"`
	VLLMPort                       int      `json:"vllm-port,omitempty"`
	MooncakeBootstrapPort          int      `json:"mooncake-bootstrap-port,omitempty"`
```

but options.go  lacks assignment logic to copy field into final options.go struct.

This may lead that  the mooncake-bootstrap-port  in YAML config file (via  --configuration-file  or  --configuration  flag) will not take effect as expected. 
  

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/sidecar/proxy/options.go
+++ b/pkg/sidecar/proxy/options.go
@@ -371,6 +371,11 @@ func (opts *Options) Validate() error {
 		return fmt.Errorf("--decode-chunk-size must be a non-negative integer (0 disables chunked decode), got %d", opts.DecodeChunkSize)
 	}
 
+	// Validate mooncake bootstrap port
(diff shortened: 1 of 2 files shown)


weizhoublue:
@zdtsw  could you help check whether this follow-up for PR #1193 makes sense, thanks

zdtsw:
> @zdtsw could you help check whether this follow-up for PR #1193 makes sense, thanks

Thanks for the fix! I missed the one for yamlConfiguration

ahg-g (approved):

PR #1591, opened 2026-06-11
feat(metrics): add inter-token latency (ITL) metric

### LINKED ISSUE #1590: Add core inter-token latency (ITL) metric for streaming requests
A new always-on Prometheus histogram `llm_d_router_epp_streaming_inter_token_latency_seconds` that records the time between consecutive response body chunks in streaming responses. This measures per-chunk decode latency as experienced by the client, complementing the existing TTFT and TPOT metrics being added in #1499.

**Why is this needed**:

Current metrics report either total request duration or averaged per-token time (NTPOT/TPOT), which mask decode-phase lags. ITL records every inter-token interval individually, enabling P99 ITL alerting and identifying GPU contention or scheduling hiccups that don't show up in averaged metrics.

tessapham:
**What type of PR is this?**

/kind feature

Adds a new always-on Prometheus histogram `lm_d_router_epp_streaming_inter_token_latency_seconds` that records the time between consecutive response body chunks in streaming responses. This complements the TTFT and TPOT metrics added in #1499.

TPOT averages per-token time across the full request, which masks decode lags. ITL records each inter-chunk (i.e., inter-token) interval individually, enabling P99 alerting on tail latency spikes that TPOT's averaging hides.
  
Fixes #1590.

Add streaming inter-token latency metric (llm_d_router_epp_streaming_inter_token_latency_seconds) to the response handler. Records the time between consecutive response body chunks for streaming requests, enabling per-interval decode latency monitoring. Not emitted for non-streaming requests.

### DIFF
--- a/pkg/epp/handlers/response.go
+++ b/pkg/epp/handlers/response.go
@@ -56,6 +56,15 @@ func (s *StreamingServer) HandleResponseBody(ctx context.Context, reqCtx *Reques
 		reqCtx.FirstTokenTimestamp = time.Now()
 	}
 
+	if reqCtx.modelServerStreaming && len(responseBytes) > 0 {
+		now := time.Now()
+		if !reqCtx.LastChunkReceivedTimestamp.IsZero() {
+			itl := now.Sub(reqCtx.LastChunkReceivedTimestamp).Seconds()
+			metrics.RecordInterTokenLatency(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, itl)
+		}
+		reqCtx.LastChunkReceivedTimestamp = now
+	}
+
 	var parsedResp *fwkrh.ParsedResponse
 	parser, err := s.getOrResolveParser(ctx, reqCtx)
 	if err != nil {
--- a/pkg/epp/handlers/response.go
+++ b/pkg/epp/handlers/response.go
@@ -59,8 +59,12 @@ func (s *StreamingServer) HandleResponseBody(ctx context.Context, reqCtx *Reques
 	if reqCtx.modelServerStreaming && len(responseBytes) > 0 {
 		now := time.Now()
 		if !reqCtx.LastChunkReceivedTimestamp.IsZero() {
+			fairnessID := metadata.DefaultFairnessID
+			if reqCtx.SchedulingRequest != nil {
+				fairnessID = reqCtx.SchedulingRequest.FairnessID
+			}
 			itl := now.Sub(reqCtx.LastChunkReceivedTimestamp).Seconds()
-			metrics.RecordInterTokenLatency(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, itl)
+			metrics.RecordInterTokenLatency(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, fairnessID, strconv.Itoa(reqCtx.Priority), itl)
 		}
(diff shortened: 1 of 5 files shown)


tessapham:
cc @hexfusion

pkg/epp/handlers/response.go
```diff
@@ -49,6 +50,15 @@ func (s *StreamingServer) HandleResponseBody(ctx context.Context, reqCtx *Reques
 
 	reqCtx.ResponseSize += len(responseBytes)
 
+	if reqCtx.modelServerStreaming && len(responseBytes) > 0 {
+		now := time.Now()
+		if !reqCtx.LastChunkReceivedTimestamp.IsZero() {
+			itl := now.Sub(reqCtx.LastChunkReceivedTimestamp).Seconds()
+			metrics.RecordInterTokenLatency(ctx, reqCtx.IncomingModelName, reqCtx.TargetModelName, itl)
```
hexfusion: question: is this inter-chunk or inter-token?

tessapham: It's inter-chunk, but for streaming responses each chunk typically is one token. Let me check again if true inter-token latency can be calculated.

hexfusion: seems reasonable maybe we can add a note in the help text that this measures inter-chunk.

tessapham: It's already been noted in the help text: `"Inter-token latency in seconds for streaming requests, measured as the time between consecutive response body chunks."`

hexfusion:
LGTM

tessapham:
@ahg-g @elevran Appreciate a review on this so it goes in with the TTFT and TPOT metrics for v0.8. Thank you!

pkg/epp/metrics/llm_d_router_metrics.go
```diff
@@ -163,6 +163,18 @@ var (
...
+	llmdInterTokenLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: LLMDRouterEndpointPickerSubsystem,
+			Name:      "streaming_inter_token_latency_seconds",
+			Help:      metricsutil.HelpMsgWithStability("Inter-token latency in seconds for streaming requests, measured as the time between consecutive response body chunks.", compbasemetrics.ALPHA),
+			Buckets: []float64{
+				0.001, 0.005, 0.01, 0.02, 0.04, 0.06, 0.08, 0.1, 0.15, 0.2, 0.3, 0.5, 0.75, 1, 2,
+			},
+		},
+		modelLabels,
```
ahg-g: what about priority and fairness id?

tessapham: ITL measures decode-phase chunk timing on the GPU, so it's the same regardless of which tenant sent the request. `fairness_id` and `priority` affect queue wait (captured by TTFT) and overall request duration (captured by TPOT), but not the inter-token gap during decoding.

tessapham: I can add them to be consistent with other metrics if you'd prefer.

ahg-g: I think we need to add them, it is useful to monitor ITL across QoS classes (modeled as a combination of priority and flow id)

tessapham: Makes sense, added.

ahg-g (approved):

PR #1592, opened 2026-06-11
feat: add priority based holdback strategies in flow control

### LINKED ISSUE #1185: [Flow Control] Priority-Based Progressive Holdback (Adaptive Capacity Reserves)
Flow control currently enforces binary backpressure: when the backend pool reaches 100% saturation, EPP stops dispatching and queues incoming requests. 

To protect critical real-time latency SLOs under heavy load, we need a **progressive holdback** mechanism. This allows EPP to proactively gate/buffer lower-priority traffic (e.g., best-effort batch at 75% load, standard at 90% load) as pool saturation rises, thereby maintaining a reserve of available compute capacity to absorb sudden high-priority bursts.

Since priority levels are dynamically derived from `InferenceObjective` CRDs at runtime, hardcoded mappings are brittle. We will implement a parameter-driven `priority-holdback-policy* plugin. This plugin will implement EPP's pluggable `flowcontrol.UsageLimitPolicy` interface and support multiple gating algorithms to dynamically distribute admission ceilings across the $[0, 1]$ saturation range.

---

### 1. Pluggable Progressive Gating Strategies

The plugin will compute an admission ceiling $c_i \in [0.0, 1.0]$ for each active priority. In the EPP `dispatchCycle`, traffic is gated if the pool's current saturation exceeds the priority's ceiling ($S > c_i$).

The plugin may support the following gating strategies (seeking feedback here):

**A. `static`**
*   Algorithm: Fixed, explicit mapping.
*   Formula:
    $$c_i = \begin{cases} 
    \text{StaticMap}[p_i] & \text{if } p_i \in \text{StaticMap} \\ 
    \text{DefaultStaticCeiling} & \text{otherwise}
    \end{cases}$$

**B. `stepwise-spread`**
*   Algorithm: Automatically divides the range between a configured `minCeiling` ($c_{min}$) and `maxCeiling` ($c_{max}$) into equal steps based strictly on the number $N$ of currently active priorities, ignoring their absolute numerical difference.
*   Formula (for index $i$ in sorted descending active priorities):
    $$c_i = c_{max} - i \times \frac{c_{max} - c_{min}}{N - 1}$$

**C. `linear-proportional`**
*   Algorithm: Scales ceilings linearly between $c_{min}$ and $c_{max}$ based on the **numerical value** of active priorities relative to the observed active range $[p_{min}, p_{max}]$.
*   Formula:
    $$r_i = \frac{p_i - p_{min}}{p_{max} - p_{min}}$$
    $$c_i = c_{min} + r_i \times (c_{max} - c_{min})$$

**D. `sigmoid-smoothed`**
*   **Algorithm**: Maps priorities to ceilings along a normalized logistic sigmoid curve to create "soft floors" and "soft ceilings", grouping similar priorities together.
*   **Formula**:
    $$\sigma(x) = \frac{1}{1 + e^{-k(x - x_0)}}$$
    $$S_{norm}(x) = \frac{\sigma(x) - \sigma(0)}{\sigma(1) - \sigma(0)}$$
    $$c_i = c_{min} + S_{norm}(x_i) \times (c_{max} - c_{min})$$
    *(Support mapping over either the absolute priority `value` domain or the relative index `rank` domain)*

---

### 2. Safety Guarantees & Corner Cases

*   Single Active Priority ($N = 1$): To respect flow control's work-conserving nature, if only a single priority level is active, EPP will bypass progressive holdback and assign it $c_{max}$ (usually `1.0`), preventing artificial limits when the pool has spare capacity.
*   Dynamic Priority Inversion Prevention: To guarantee that a higher priority is never gated before a lower priority (regardless of mathematical edge cases), the plugin will run a right-to-left monotonizing sweep over the calculated ceilings:
    ```go
    // Ensure ceilings[i] >= ceilings[i+1] (highest first)
    for i := n - 2; i >= 0; i-- {
        if ceilings[i] < ceilings[i+1] {
            ceilings[i] = ceilings[i+1]
        }
    }
    ```
*   Boot-Time Validation: The plugin factory will perform local priority-inversion validation on startup when using the `static` mapping. EPP will fail to boot on invalid configurations.

---

### 3. Configuration Schema

The policy is configured inside the existing EPP `plugins` block and referenced inside `flowControl`:

```yaml
kind: EndpointPickerConfig
metadata:
  name: default-config
featureGates:
- flowControl
plugins:
- type: priority-holdback-policy
  name: dynamic-holdback
  parameters:
    strategy: "sigmoid-smoothed"
    minCeiling: 0.6
    maxCeiling: 1.0
    sigmoidSteepness: 8.0
    sigmoidMidpoint: 0.5
    sigmoidDomain: "value"
flowControl:
  usageLimitPolicyPluginRef: dynamic-holdback
```

loicmarchal:
**What type of PR is this?**
/kind feature

Add progressive holdback based on priorities.

It implements the priority-holdback-policy plugin (flowcontrol.UsageLimitPolicy), which computes differentiated admission ceilings per priority level. Lower-priority traffic is gated at lower saturation thresholds, reserving capacity for higher-priority work as the pool approaches saturation.

It replaces the default static usage limit policy's binary backpressure approach with priority-aware stepped gating.

Two gating strategies are supported:
- stepwise-spread: equal ceiling spacing by rank, ignoring numerical priority values.
$$c_i = c_{max} - i \times \frac{c_{max} - c_{min}}{N - 1}$$
- linear-proportional: ceiling spacing proportional to numerical priority differences.
$$c_i = c_{min} + r_i \times (c_{max} - c_{min})$$
$$r_i = \frac{p_i - p_{min}}{p_{max} - p_{min}}$$

Both are the same linear mapping applied to different input domains (rank vs value). With evenly distributed priorities, they converge.

It's a first implementation for a v1. More strategies could be added to offer more control of the ceilings distribution to the user (e.g. Sigmoid curve instead of linear).

Fixes #1185 

```release-note
Add priority-aware stepped gating policy as an alternative to the default binary backpressure approach
```

### DIFF
--- a/pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/README.md
+++ b/pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/README.md
@@ -4,7 +4,7 @@
 
 A usage limit policy that computes differentiated admission ceilings per priority level. As pool saturation rises, lower-priority traffic is gated first, reserving capacity for higher-priority work.
 
-This replaces the default static usage limit policy (which applies a single ceiling to all priorities) with priority-aware stepped gating.
+This can be used in place the default static usage limit policy (which applies a single ceiling to all priorities) with priority-aware stepped gating.
 
 ## What It Does
 
@@ -39,7 +39,7 @@ Use when the numerical spacing between priority values carries meaning and prior
 
 - `strategy` (string, required, no default): Gating algorithm: `"stepwise-spread"` or `"linear-proportional"`.
 - `minCeiling` (float64, required, no default): Ceiling for the lowest priority. Must be in `[0.0, 1.0)`.
-- `maxCeiling` (float64, required, default: `1.0`): Ceiling for the highest priority. Must be in `(0.0, 1.0]`.
+- `maxCeiling` (float64, optional, default: `1.0`): Ceiling for the highest priority. Must be in `(0.0, 1.0]`.
 
 `minCeiling` is required because it determines how aggressively low-priority traffic is gated and there is no universally correct default.
 
--- a/pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/README.md
+++ b/pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/README.md
@@ -1,0 +1,63 @@
+# Priority Holdback Policy Plugin
+
+**Type:** `priority-holdback-policy`
+
+A usage limit policy that computes differentiated admission ceilings per priority level. As pool saturation rises, lower-priority traffic is gated first, reserving capacity for higher-priority work.
+
+This replaces the default static usage limit policy (which applies a single ceiling to all priorities) with priority-aware stepped gating.
+
+## What It Does
+
+Each active priority level receives an admission ceiling in [0.0, 1.0]. During each dispatch cycle, the Flow Controller compares current pool saturation against each priority's ceiling. When saturation exceeds a priority's ceiling, that priority's traffic is held back.
+
+Higher-priority traffic continues to flow until saturation reaches its own (higher) ceiling, providing quality-of-service differentiation under load.
+
+## Strategies
+
+Two strategies are available for computing per-priority ceilings. Both produce `maxCeiling` for the highest priority and `minCeiling` for the lowest.
+
+### `stepwise-spread`
+
+Distributes ceilings in equal steps by rank, ignoring numerical priority values.
+
+    c_i = maxCeiling - i * (maxCeiling - minCeiling) / (N - 1)
+
+Where `i` is the index in descending priority order (0 = highest) and `N` is the count of active priorities.
+
+Use when priorities represent ordinal categories (e.g., "critical", "normal", "batch") where the numerical values are arbitrary labels.
+
+### `linear-proportional`
+
+Scales ceilings proportionally to the numerical priority value within the observed active range.
+
+    r_i = (p_i - pMin) / (pMax - pMin)
+    c_i = minCeiling + r_i * (maxCeiling - minCeiling)
+
+Use when the numerical spacing between priority values carries meaning and priorities that are numerically close should behave similarly under pressure.
+
+**Parameters:**
+
+- `strategy` (string, required, no default): Gating algorithm: `"stepwise-spread"` or `"linear-proportional"`.
+- `minCeiling` (float64, required, no default): Ceiling for the lowest priority. Must be in `[0.0, 1.0)`.
+- `maxCeiling` (float64, required, default: `1.0`): Ceiling for the highest priority. Must be in `(0.0, 1.0]`.
+
+`minCeiling` is required because it determines how aggressively low-priority traffic is gated and there is no universally correct default.
+
+**Configuration Example:**
+```yaml
+plugins:
+  - type: priority-holdback-policy
+    name: my-holdback-policy
+    parameters:
+      strategy: stepwise-spread
+      minCeiling: 0.4
+      maxCeiling: 0.9
+flowControl:
+  usageLimitPolicyPluginRef: my-holdback-policy
+```
+
+---
+
+## Related Documentation
+- [Static Usage Limit Policy](../README.md)
+- [Flow Control Overview](../../fairness/README.md)
--- a/pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/README.md
+++ b/pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/README.md
@@ -12,13 +12,21 @@ Each active priority level receives an admission ceiling in [0.0, 1.0]. During e
 
 Higher-priority traffic continues to flow until saturation reaches its own (higher) ceiling, providing quality-of-service differentiation under load.
 
-## Strategies
+## Configuration
 
-Two strategies are available for computing per-priority ceilings. Both produce `maxCeiling` for the highest priority and `minCeiling` for the lowest.
+Behavior is configured via two independent parameters: `shape` and `domain`. The resulting ceilings always range from `maxCeiling` for the highest priority to `minCeiling` for the lowest.
 
-### `stepwise-spread`
+### `shape`
 
-Distributes ceilings in equal steps by rank, ignoring numerical priority values.
+The interpolation curve used to distribute ceilings across the range. Currently only `"linear"` is supported.
+
+### `domain`
+
+How priority levels are mapped to positions in the ceiling range.
+
+#### `rank` (default)
+
+Distributes ceilings in equal steps by ordinal rank, ignoring numerical priority values. Ceilings are evenly spaced across `[minCeiling, maxCeiling]`.
 
     c_i = maxCeiling - i * (maxCeiling - minCeiling) / (N - 1)
 
@@ -26,7 +34,7 @@ Where `i` is the index in descending priority order (0 = highest) and `N` is the
 
 Use when priorities represent ordinal categories (e.g., "critical", "normal", "batch") where the numerical values are arbitrary labels.
 
-### `linear-proportional`
+#### `value`
 
 Scales ceilings proportionally to the numerical priority value within the observed active range.
 
@@ -37,7 +45,8 @@ Use when the numerical spacing between priority values carries meaning and prior
 
 **Parameters:**
 
-- `strategy` (string, required, no default): Gating algorithm: `"stepwise-spread"` or `"linear-proportional"`.
+- `shape` (string, optional, default: `"linear"`): Interpolation curve. Currently only `"linear"` is supported.
+- `domain` (string, optional, default: `"rank"`): Priority mapping: `"rank"` or `"value"`.
 - `minCeiling` (float64, required, no default): Ceiling for the lowest priority. Must be in `[0.0, 1.0)`.
 - `maxCeiling` (float64, optional, default: `1.0`): Ceiling for the highest priority. Must be in `(0.0, 1.0]`.
 
@@ -49,7 +58,8 @@ plugins:
   - type: priority-holdback-policy
     name: my-holdback-policy
     parameters:
-      strategy: stepwise-spread
+      shape: linear
+      domain: rank
       minCeiling: 0.4
       maxCeiling: 0.9
 flowControl:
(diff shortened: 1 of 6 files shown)


LukeAVanDrie:
Didn't get around to reviewing this today; sorry @loicmarchal! Will take a look when I am back in office Tuesday. Let's target this for v0.10 if possible and try to land this alongside eviction (which @RishabhSaini has a draft for).

loicmarchal:
> Didn't get around to reviewing this today; sorry @loicmarchal! Will take a look when I am back in office Tuesday. Let's target this for v0.10 if possible and try to land this alongside eviction (which @RishabhSaini has a draft for).

Thank you @LukeAVanDrie. No worries. Sounds good for v0.10

loicmarchal:
@LukeAVanDrie I implemented a new plugin in its own package. I hesitated between this approach and making use of the existing usage limit. The way the usage limit policy was implemented uses a function that is simply called in the `ComputeLimit` function. We could look into making the usage limit more generic and simply return the function associated with the strategy defined in the config. That might require some refactoring, and maybe adapting the doc a little bit. I'll take a look at it.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/README.md
```diff
@@ -0,0 +1,63 @@
+# Priority Holdback Policy Plugin
+
+**Type:** `priority-holdback-policy`
+
+A usage limit policy that computes differentiated admission ceilings per priority level. As pool saturation rises, lower-priority traffic is gated first, reserving capacity for higher-priority work.
+
+This replaces the default static usage limit policy (which applies a single ceiling to all priorities) with priority-aware stepped gating.
```
LukeAVanDrie: nit: "replaces the default" reads as automatic but the plugin only takes effect when `flowControl.usageLimitPolicyPluginRef` points at it. Suggest softening to "can be used in place of the default static usage limit policy" so users know it's opt-in.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/priority_holdback.go
```diff
@@ -0,0 +1,128 @@
...
+// ComputeLimit returns an admission ceiling for each priority. With a single active priority,
+// holdback is bypassed (ceiling = cMax) to preserve work-conserving behavior.
+func (p *priorityHoldbackPolicy) ComputeLimit(_ context.Context, _ float64, priorities []int) (ceilings []float64) {
+	if len(priorities) == 0 {
+		return []float64{}
+	}
+	if len(priorities) == 1 {
+		return []float64{p.cMax}
+	}
+	return p.computeFn(p.cMin, p.cMax, priorities)
```
LukeAVanDrie, line 109: Non-blocking, your call: the monotonicity of both strategies relies on the contract's "highest first" ordering of priorities, and there's no monotonizing sweep (the one mentioned in #1185). That's correct for these two linear strategies, but a one-liner here (or on the strategy funcs) noting "ceilings are monotone only because priorities arrive sorted descending per the `UsageLimitPolicy` contract; sigmoid/static strategies will need the explicit monotonizing sweep" would benefit future contributors.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/config.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+}
+
+// validateConfig checks the constraints of the fully defaulted configuration.
+// It aggregates all validation failures rather than failing on the first error.
+func validateConfig(cfg *apiConfig) error {
+	var errs []error
+
+	if cfg.Strategy != nil {
+		switch *cfg.Strategy {
```
LukeAVanDrie: Non-blocking nit: this switch has no default, so an unrecognized strategy leaves `computeFn` nil and `ComputeLimit` nil-panics for ≥ 2 priorities. It's unreachable through the factory (strategy is validated in `buildConfig`), so not a real bug today.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/README.md
```diff
@@ -0,0 +1,63 @@
...
+    r_i = (p_i - pMin) / (pMax - pMin)
+    c_i = minCeiling + r_i * (maxCeiling - minCeiling)
+
+Use when the numerical spacing between priority values carries meaning and priorities that are numerically close should behave similarly under pressure.
+
+**Parameters:**
+
+- `strategy` (string, required, no default): Gating algorithm: `"stepwise-spread"` or `"linear-proportional"`.
+- `minCeiling` (float64, required, no default): Ceiling for the lowest priority. Must be in `[0.0, 1.0)`.
+- `maxCeiling` (float64, required, default: `1.0`): Ceiling for the highest priority. Must be in `(0.0, 1.0]`.
```
LukeAVanDrie: `maxCeiling` is actually optional; it's omitempty and defaults to 1.0 in `applyDefaults`, so "required" contradicts "default".

```suggestion
- `maxCeiling` (float64, optional, default: `1.0`): Ceiling for the highest priority. Must be in `(0.0, 1.0]`.
```

LukeAVanDrie (commented):
Right now `stepwise-spread`/`linear-proportional` encode a 2-D space (curve × domain) as a flat enum: both are the *same linear map* over different input domains (rank vs value), which the PR description already says.

I'd recommend two orthogonal fields instead:

```yaml
parameters:
    curve: linear      # optional, default: linear   (future: sigmoid)
    domain: rank       # optional, default: rank     (rank | value)
    minCeiling: 0.4    # required
    maxCeiling: 0.9    # optional, default: 1.0
```
Exact mapping:

- `stepwise-spread` → `curve: linear, domain: rank`
- `linear-proportional` → `curve: linear,  domain: value`.

Sigmoid is a speculative strategy we should consider. This refactoring makes it purely additive (one new `curve` value, no rename), and it's near-free to support now vs. a breaking config migration later. Two notes for the implementation

1. default `domain: rank` (most people use arbitrary priority numbers and expect ordinal behavior
2. `value` only matters when the numeric spacing is deliberate), and you don't need to design the sigmoid-specific params (steepness/midpoint) now; the field split is just forward-compatible with them.

If we'd rather formally rule sigmoid out, the alternative is to drop `curve` entirely and keep only `domain: rank|value`. Either way, I prefer this over the current flat enum, which is neither minimal nor extensible.

Also happy to do this as a followup.

LukeAVanDrie:
@loicmarchal really nice work on this. Sorry for the review delay. I have one suggestion regarding the config; else this looks great to me. Also fine merging this in its current state and addressing the config as a followup

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/priority_holdback.go
loicmarchal: That's correct, I considered the monotonizing sweep was not needed with the current strategies so we should only introduce it when we add strategies that could affect the decreasing order. I agree it might not be obvious for future development though. I added the comment.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/config.go
loicmarchal: Indeed. The reason why I left it like this is because there is no obvious default for strategy. So it should be guarded by the `buildConfig` (in the `checkRequired`). A non nil strategy should be a precondition to the `validateConfig`

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/README.md
loicmarchal: good catch, thank you!

loicmarchal:
> I'd recommend two orthogonal fields instead:

I agree. That was a main argument in my [comment](https://github.com/llm-d/llm-d-router/issues/1185#issuecomment-4714789029)

My proposal:
```yaml
parameters:
    shape: linear      # optional, default: linear   (future: sigmoid) => clarify in the doc we currently only support linear
    domain: rank       # optional, default: rank     (rank | value)
    minCeiling: 0.4    # required
    maxCeiling: 0.9    # optional, default: 1.0
```
This way, we are future-proof for sigmoid, exponential, step, etc. At the same time, if we stay as it is, nothing will break.

If ok with you, I'll implement that change, then maybe we can merge and add sigmoid (or others) as a future implementation?

loicmarchal:
Thank you for your review @LukeAVanDrie. I changed the config.

LukeAVanDrie (dismissed):
/lgtm

Thanks @loicmarchal!

loicmarchal:
@ahg-g @liu-cong does it look ok for you? Can it be merged?

lioraron:
@ahg-g @liu-cong @elevran @shmuelk This PR is critical for batch-gateway integration with flow-control - https://github.com/llm-d/llm-d-router/issues/1185#issuecomment-4928992959.

Can you please look into merging this?

Thanks.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/config.go
```diff
@@ -0,0 +1,167 @@
+/*
+Copyright 2025 The Kubernetes Authors.
```
liu-cong:
```suggestion
Copyright 2026 The Kubernetes Authors.
```

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/priority_holdback_test.go
```diff
@@ -0,0 +1,518 @@
+/*
+Copyright 2025 The Kubernetes Authors.
```
liu-cong:
```suggestion
Copyright 2026 The Kubernetes Authors.
```

liu-cong (dismissed):

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/config.go
```diff
@@ -0,0 +1,167 @@
...
+		domain:     *safeCfg.Domain,
+		minCeiling: *safeCfg.MinCeiling,
+		maxCeiling: *safeCfg.MaxCeiling,
+	}, nil
+}
+
+// checkRequired verifies that mandatory fields are present before defaulting.
+func checkRequired(cfg *apiConfig) error {
+	if cfg.MinCeiling == nil {
+		return fmt.Errorf("minCeiling is required")
```
liu-cong:
```suggestion
		return errors.New("minCeiling is required")
```

liu-cong (approved):

PR #1597, opened 2026-06-11
Set session affinity token in response headers

learner0810:
**What type of PR is this?**

/kind bug

The session affinity scorer sets x-session-token so clients can send it on later requests and be routed back to the same pod. Response headers are already sent to Envoy before response body processing runs, so setting the token from ResponseBody does not reliably include it in the client response. This change registers the scorer as a ResponseHeaderProcessor and sets the token while response headers are still mutable.

The tests are updated to cover the response header hook, nil header maps, nil target pods, and nil responses.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity/session_affinity.go
+++ b/pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity/session_affinity.go
@@ -23,7 +23,7 @@ const (
 
 // compile-time type assertion
 var _ scheduling.Scorer = &SessionAffinity{}
-var _ requestcontrol.ResponseBodyProcessor = &SessionAffinity{}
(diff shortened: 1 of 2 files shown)


elevran:
@learner0810 - thanks for submitting this.

While correcting an issue, this change alone will not cause the client to send back the session identify header - it's just not part of the HTTP standard. See https://github.com/llm-d/llm-d-router/pull/600 for discussion of using standard `cookie` header mechanism to fix (which was ultimately **not merged**).
There is ongoing work around this: #815 #1396. Please review prior and existing work and let us know if anything would be amiss once they land.

elevran (approved):
The fix is correct -- `ResponseHeaderProcessor` is the right hook for setting response headers while they're still mutable, and the revised nil-check ordering is safer than the original. 
Approving - waiting for checks to complete

PR #1599, opened 2026-06-11
feat(epp): add load-aware picker to mitigate thundering-herd concentration

### LINKED ISSUE #1598: Imrpove request disribution for bursty workloads (e.g., RL)
Improve request scheduling during bursty traffic spikes, such as RL or system coming out of saturation.
Given that multiple requests are scheduled concurrently and likely **observing the same system state, they'll tend concentrate on a small** number (possibly one) of nodes as optimal. When request arrival is more staggered, requests are more likely to see different system states.

This scheduling pattern was observed in RL step cycles (cc @ezrasilvera) and batch inference runs, causing an imbalaned request distribution.

The goal is to experiment with algorithms that provide better load balancing with minimal impact on overall performance to counter thundering herd effects.

**Why is this needed**:
When multiple requests are run through scheduling, each runs in its own Go routine, concurrently with others.
The majority of these will see identical or very similar inputs (e.g., system state is snapshotted from the datastore, metrics are updated only peridically, etc.). Consequently, filtering, scoring and linear combination result in the same small subset being preferred (either by the max score picker or weighted variant). 

- Active request counting scorer help this somewhat as it does introduce some additional variability. However, scoring runs before picking, so there is a potential read/write timing (i.e., many requests will see the same Scorer state until Pick runs - we are at the mercy of the Go scheduler...)
- some scorers can make assumption on final scheduling (eg., assume the best prefix will be selected in Pick() so treat the request as if it was scheduled. This can contradict the actual choice (corrected after Pick()) which is affected by multiple scorers and their linear combination.

elevran:
**What type of PR is this?**
/kind feature

Adds `LoadAwarePicker`, a scheduling picker that mitigates the thundering-herd endpoint concentration problem. When a burst of concurrent requests arrives, stateless pickers (max-score, weighted-random) all snapshot the same endpoint metrics and converge on the same endpoint. This picker serializes the final selection step behind a mutex so each goroutine in a burst observes picks made by all preceding goroutines before it chooses.

Two multiplicative score adjustments are applied under the mutex using live per-endpoint state:
- **Concentration factor** `min(1.0, expectedShare/actual)`: penalizes endpoints that have received more than their fair share of picks in the last 10 seconds.
- **Capacity factor** `max(0, 1 - load/maxConcurrency)`: derived via Little's Law from a per-endpoint EMA of throughput, latency, and tokens-per-request. A pending-request counter (incremented at `Pick`, decremented at `PreRequest`) bridges the pick-to-dispatch gap so the next goroutine in the same burst sees it immediately.

The plugin implements `scheduling.Picker`, `requestcontrol.PreRequest`, and `requestcontrol.ResponseBodyProcessor`, and is wired automatically via the runner's `AddPlugins` handle flow.

See `pkg/epp/framework/plugins/scheduling/picker/loadaware/README.md` for usage guidance, and the linked issue for full analysis and ADR.

Fixes https://github.com/llm-d/llm-d-router/issues/1598

```release-note
Add load-aware-picker plugin: serializes endpoint selection and applies concentration and capacity score adjustments to mitigate thundering-herd scheduling concentration.
```


pkg/epp/framework/plugins/scheduling/picker/loadaware/README.md
```diff
@@ -0,0 +1,66 @@
+# Load-Aware Picker
+
+**Type:** `load-aware-picker`
+
+Serializing picker that aims to mitigate thundering-herd endpoint concentration by making each pick
+visible to the next before it chooses.
+
+## Why this exists
```
vMaroon, line 8: Can we attach a benchmark to this doc

shmuelk: Also take care with how the mutex in the plugin affects throughput and parallelism

pkg/epp/framework/plugins/scheduling/picker/loadaware/picker.go
```diff
@@ -0,0 +1,366 @@
...
+var _ fwkrc.PreRequest = &LoadAwarePicker{}
+var _ fwkrc.ResponseBodyProcessor = &LoadAwarePicker{}
+
+// LoadAwarePickerFactory is the factory function for LoadAwarePicker.
+func LoadAwarePickerFactory(name string, rawParameters *json.Decoder, handle fwkplugin.Handle) (fwkplugin.Plugin, error) {
+	if rawParameters != nil {
+		if err := rawParameters.Decode(&struct{}{}); err != nil {
+			return nil, fmt.Errorf("failed to parse parameters for %s: %w", LoadAwarePickerType, err)
+		}
+	}
```
shmuelk: If the plugin takes no parameters, why parse?




```suggestion
func LoadAwarePickerFactory(name string, _ *json.Decoder, handle fwkplugin.Handle) (fwkplugin.Plugin, error) {
```

shmuelk:
Just curious wouldn't the existing "LoadAwareScorer/RequestCountingScorer" with.an appropriate weight hdo the trick here?

I understand that cache aware scorers of any kind tend to cluster requests on a few model servers. Couldn't "LoadAwareScorer/RequestCountingScorer" with an appropriate weight lower the scorer of heavily loaded model servers?

pkg/epp/framework/plugins/scheduling/picker/loadaware/README.md
elevran: definitely - should not merge without showing benefits and not negative impact in the common case

pkg/epp/framework/plugins/scheduling/picker/loadaware/picker.go
elevran: Place holder for future parameters, but you're correct - not needed now, removing.

elevran:
> Just curious wouldn't the existing "LoadAwareScorer/RequestCountingScorer" with.an appropriate weight hdo the trick here?
> 
> I understand that cache aware scorers of any kind tend to cluster requests on a few model servers. Couldn't "LoadAwareScorer/RequestCountingScorer" with an appropriate weight lower the scorer of heavily loaded model servers?

have done some early testing and the load aware/request counting were showing biased selection.
Will provide more details.
The goal is to keep some kv cache affinity without th thundering herd, not replacing it.

elevran:
/hold need to provide benchmark results before moving forward.
Currently this is speculative, not empirical.

pkg/epp/framework/plugins/scheduling/picker/loadaware/picker.go
elevran: Fixed in commit 9ef3d805. Removed the no-op parameter parsing block and blanked the unused parameter.

pkg/epp/framework/plugins/scheduling/picker/loadaware/picker.go
```diff
@@ -0,0 +1,361 @@
...
+	p.requestCache.Delete(req.RequestID)
+
+	duration := time.Since(rec.dispatchTime).Seconds()
+	if duration <= 0 {
+		return
+	}
+
+	inputTokens := 0
+	if req.Body != nil && req.Body.TokenizedPrompt != nil {
+		inputTokens = len(req.Body.TokenizedPrompt.TokenIDs)
```
Mohamedma96, line 250: inputTokens = len(req.Body.PerPromptTokens.TokenIDs)

Mohamedma96 (commented):
Hi,

I have tested the issue with main epp and sent a burst of 50 requests , and I saw that all the requests are handled in a single pod.
I built a new image for this PR and it seems that when getting a burst, the load is distributed among the pods as expected. 
The testing was done on openshift cluster.
the image is pushed to quay.io if you want to test it.


Behavior before the fix:

Image

After deploying the image with the fix:

Image


So it seems this is doing the job, thanks

elevran:
closing as the current design has negative impact on normal operations. Would need a different heuristics/algorithm to make this more useful across the board.

PR #1600, opened 2026-06-11
feat(standalone): align agentgateway configuration with upstream GAIE PR 2905

abhay1999:
This PR aligns the standalone agentgateway configuration in the llm-d-router Helm chart with the upstream [GAIE PR #2905](https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2905)

We are making this change to address Daneyon's request for parity:

- Deleted agentgateway-service.yaml so that no physical K8s Service is created for agentgateway.
- Added a helper llm-d-router.agentgateway.logicalBackendName to derive the pseudo-service name from .Values.router.modelServers.matchLabels.app (falls back to .Release.Name).
- Updated llm-d-router.proxy.agentgatewayConfig to read ports directly from targetPorts, removing the old router.proxy.agentgateway.service configuration.
- Added a validation check to fail if anyone uses the deprecated router.proxy.agentgateway configuration.
- Updated the verification script hack/verify-helm.sh and the documentation in config/charts/README.md.

Note: This is a breaking change because the old service configuration fields are removed, which means it will require a minor version bump for the next release.


config/charts/llm-d-router-standalone/templates/_validations.tpl
```diff
@@ -46,32 +46,17 @@ standalone validations
     {{- fail (printf ".Values.router.proxy.proxyType must be one of [envoy, agentgateway], got %q" $proxyType) -}}
   {{- end -}}
   {{- if eq $proxyType "agentgateway" -}}
+    {{- if hasKey $proxy "agentgateway" -}}
+      {{- fail ".Values.router.proxy.agentgateway is no longer supported; standalone agentgateway derives its logical backend from router.modelServers settings" -}}
+    {{- end -}}
     {{- if ne .Values.router.inferencePool.create false -}}
       {{- fail ".Values.router.inferencePool.create=false is required when proxyType=agentgateway; standalone agentgateway currently supports only service-backed routing" -}}
```
danehans, line 53: @abhay1999 can you update this validation error message to reflect the changes in this PR, i.e. "standalone agentgateway uses EPP endpoint discovery with a logical service backend"

hack/verify-helm.sh
```diff
@@ -155,35 +155,28 @@ if eval "${invalid_proxy_command}"; then
   exit 1
 fi
 
-missing_agentgateway_service_command="${HELM} template ${SCRIPT_ROOT}/config/charts/llm-d-router-standalone --set router.modelServers.matchLabels.app=llm-instance-gateway --set router.inferencePool.create=false --set router.proxy.proxyType=agentgateway >/dev/null"
-echo "Executing: ${missing_agentgateway_service_command}"
-if eval "${missing_agentgateway_service_command}"; then
-  echo "Helm template unexpectedly succeeded for missing agentgateway service.name"
+deprecated_agentgateway_service_command="${HELM} template ${SCRIPT_ROOT}/config/charts/llm-d-router-standalone --set router.modelServers.matchLabels.app=llm-instance-gateway --set router.inferencePool.create=false --set router.proxy.proxyType=agentgateway --set router.proxy.agentgateway.service.name=foo >/dev/null"
```
danehans, line 158: Add the upstream GAIE parity assertions for rendered logical backend name and hostname. I confirmed they render correctly but these tests only check passthrough mode and absence of the old Service.

danehans:
A couple nits, otherwise /lgtm

config/charts/llm-d-router-standalone/templates/_validations.tpl
abhay1999: Done. Updated the validation error message to reflect that standalone agentgateway uses EPP endpoint discovery with a logical service backend.

hack/verify-helm.sh
abhay1999: Done. Added the assertions to verify that both name: "default/llm-instance-gateway" and hostname: "llm-instance-gateway" are correctly rendered in standalone mode.

danehans:
/lgtm

cc: @ahg-g

ahg-g (approved):

Issue #1595, opened 2026-06-11
feat: separate render deployment from EPP/simulator pods; use estimate tokenizer in e2e


mayabar:
### Problem
Two related bottlenecks in the current render (tokenizer) setup:

1. **Render is a single-replica deployment with no path to scale**
The render pod is deployed as a single instance used by the EPP token-producer plugin. Under load, all render requests funnel through this one pod, making it a throughput bottleneck with no way to scale horizontally.

2. **Per-test-case render startup multiplies e2e latency**
E2e tests tear down and recreate the full stack — router (EPP), simulators, and render — for each test case. Each recreation includes spinning up a render container and waiting for the tokenizer to load. This overhead compounds across the test suite.

3. **Real render triggers HuggingFace API calls**
When a real render instance starts, it queries the HuggingFace API to check if model is real HF model. Tests mix real and fake (dummy) models, so every test-case startup hits this external dependency — adding latency, requiring network access, and creating a failure mode unrelated to what the test is actually validating. This is the primary motivation for switching to a simulated render in e2e.

### Proposed Change
**Deployment**: Extract render into its own Deployment + Service definition, independent of the EPP deployment. This makes replicas a first-class configuration knob — operators can scale the render service independently under load.

**E2e tests**: Replace the real render deployment in tests with the estimate tokenizer backend (byte-packing fallback, no external service required). Pair this with simulators running in a mode that bypasses real tokenization. This removes the render startup wait from the critical path of every test run without changing what the tests actually validate at the routing layer.

PR #1601, opened 2026-06-11
feat: deploy vllm-render as standalone service; use simulated render in e2e

mayabar:
**What type of PR is this?**
/kind feature
/kind test

Extracts the vllm-render container from the EPP Deployment sidecar into its own standalone Deployment + Service. This makes the render service independently scalable — operators can tune replica count without touching the EPP deployment. It also removes the tight coupling between EPP and render lifecycle.

For e2e tests, replaces the real vllm-render instance with the inference simulator image running with --force-dummy-tokenizer. 

This eliminates:
- the HuggingFace API call on every test-case startup (which required network access and hit an external dependency unrelated to what routing tests validate)
- the per-test render container startup wait
- the runtime YAML mutation logic (removeRenderSidecar, isModelReal, removePodSpecListItem) that patched out the sidecar for simulated-model tests

The EPP config is updated to point at `http://vllm-render:8082` (the new Service) instead of `http://localhost:8000`, and simulators receive -`-force-dummy-tokenizer` to match.

Ref #1595

**Release note (write NONE if no user-facing change):**

`vllm-render` is now deployed as an independent Deployment and Service rather than a sidecar in the EPP pod, allowing it to be scaled independently.


deploy/environments/dev/base-kind-istio/vllm-render.yaml
```diff
@@ -0,0 +1,69 @@
...
+    metadata:
+      labels:
+        app: vllm-render
+    spec:
+      containers:
+      - name: vllm-render
+        image: ${VLLM_RENDER_IMAGE}
+        imagePullPolicy: IfNotPresent
+        command: ["vllm", "launch", "render"]
+        args: ["${MODEL_NAME}", "--port=8082"]
```
elevran: The previous tokenization sidecar scheme injected `${HF_TOKEN}` for gated models. This standalone Deployment has no env section, so operators using a real **gated** model in `base-kind-istio` could hit a silent auth failure with no obvious cause. A comment or placeholder env entry here pointing to where to inject the token would help. Or mention this in the PR body (I don't think this is user facing - it is for dev's - so likely no need for mention in release-note.

deploy/components/vllm-decode/deployment.yaml
```diff
@@ -68,6 +68,7 @@ spec:
         - "--port=8200"
         - "--model=${MODEL_NAME}"
         - "--data-parallel-size=${VLLM_DATA_PARALLEL_SIZE}"
+        - "--render-url=http://vllm-render:8082"
```
elevran: nit: `http://vllm-render:8082` is hardcoded in base component YAML alongside substitution variables like `${MODEL_NAME}`. Using `${VLLM_RENDER_URL}` here (and in the encode/prefill components) would be consistent with the project pattern and let overlays override it. Not blocking.

deploy/environments/dev/e2e-infra/vllm-render.yaml
```diff
@@ -0,0 +1,67 @@
...
+          periodSeconds: 10
+          failureThreshold: 5
+          timeoutSeconds: 1
+        readinessProbe:
+          httpGet:
+            path: /health
+            port: 8082
+          periodSeconds: 5
+          failureThreshold: 1
+          timeoutSeconds: 1
```
elevran, line 50: nit: `volumes: []` is an explicit empty list; omit the field. Kubernetes treats absent and empty identically.

deploy/environments/dev/base-kind-istio/vllm-render.yaml
```diff
@@ -0,0 +1,69 @@
...
+          httpGet:
+            path: /health
+            port: 8082
+          periodSeconds: 5
+          failureThreshold: 1
+          timeoutSeconds: 1
+        volumeMounts:
+        - name: model-cache
+          mountPath: /root/.cache/huggingface
+      volumes:
```
elevran: nit: trailing whitespace.

elevran (commented):
Good PR - thanks!
It clean removal of external test dependencies and YAML mutation machinery. A few suggestions/nits before merging are noted inlnie.
Also - please resolve conflicts,

deploy/environments/dev/base-kind-istio/vllm-render.yaml
mayabar: a placeholder for ${HF_TOKEN} was added to yaml files

deploy/components/vllm-decode/deployment.yaml
mayabar: added

deploy/environments/dev/e2e-infra/vllm-render.yaml
mayabar: removed

deploy/environments/dev/base-kind-istio/vllm-render.yaml
mayabar: removed

test/e2e/setup_test.go
```diff
@@ -33,9 +30,9 @@ func createModelServersFromKustomize(kustomizeDir string, extra map[string]strin
 		"${EPP_NAME}":                "e2e-epp",
 		"${NAMESPACE}":               nsName,
 		"${HF_TOKEN}":                os.Getenv("HF_TOKEN"),
-		"${VLLM_EXTRA_ARGS_E}":       "",
-		"${VLLM_EXTRA_ARGS_P}":       "",
-		"${VLLM_EXTRA_ARGS_D}":       "",
+		"${VLLM_EXTRA_ARGS_E}":       "--force-dummy-tokenizer",
+		"${VLLM_EXTRA_ARGS_P}":       "--force-dummy-tokenizer",
+		"${VLLM_EXTRA_ARGS_D}":       "--force-dummy-tokenizer",
```
elevran, line 37: ${VLLM_RENDER_URL} is not in this substitution map. After kustomize build and substituteMany, the decode/encode/prefill containers will receive the literal string --render-url=${VLLM_RENDER_URL} as a command argument. removeEmptyArgs will not catch it because the value is not empty. Add "${VLLM_RENDER_URL}": "http://vllm-render:" + vllmRenderPort here, consistent with how createRender already handles ${VLLM_RENDER_PORT}.

test/e2e/configs_test.go
```diff
@@ -187,7 +187,7 @@ plugins:
   parameters:
     modelName: Qwen/Qwen2.5-1.5B-Instruct
     vllm:
-      url: http://localhost:8000
+      url: http://vllm-render:8082
```
elevran: Nit: URL is hardcoded; if VLLM_RENDER_PORT is changed the EPP config and the render deployment will diverge silently. Building from vllmRenderPort (same as createRender) keeps them in sync.

test/e2e/configs_test.go
```diff
@@ -219,7 +219,7 @@ plugins:
...
```
elevran: Same nit as line 190: hardcoded port.

ahg-g:
Not that users use the helm charts for deployment, so this PR doesn't practically enable vllm-render as a standalone for end users.

mayabar:
> Not that users use the helm charts for deployment, so this PR doesn't practically enable vllm-render as a standalone for end users.

@ahg-g That's correct. The scope of this specific PR is strictly limited to updating the e2e tests and the kind deployment. Is this scope ok from your point of view?

elevran (approved):

PR #1607, opened 2026-06-11
make EPP gRPC healthCheck configurable in helm

### LINKED ISSUE #1604: gRPC healthcheck port is not configurable because the helm template does not expose it
Update config/charts/inferencepool/templates/gke.yaml to make the port configurable.

**Why is this needed**:

The gRPC healthcheck port of gaie-pd-epp is not exposed by config/charts/inferencepool/templates/gke.yaml. The hardcoded port can be blocked by the firewall or security policy of GCP cloud projects (especially Google internal projects), causing GCP health check traffic unable to reach  gaie-pd-epp. Switching to an allowlisted port can fix this issue. But the port is not exposed by config/charts/inferencepool/templates/gke.yaml. Thus, it should be exposed and configurable.

Porting from https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2884

zetxqx:
**What type of PR is this?**
/kind feature

Makes the EPP gRPC health check port configurable via a single chart value, `router.epp.grpcHealthPort` (optional, defaults to 9003).

The port was previously hardcoded in five places in the EPP deployment template (the `grpc-health` container port and the liveness/readiness probe ports) and the `--grpc-health-port` flag was never passed to the EPP binary, so the port could not be changed coherently. The value now drives the flag, the container port, and the probe ports together, so the binary, pod spec, and probes can never drift apart.

On GKE, when the port is set to anything other than 9003, the chart renders a `HealthCheckPolicy` targeting the EPP Service with a gRPC check on that port. With the default 9003 no policy is rendered, since the GKE gateway controller already health-checks the EPP backend on gRPC 9003, keeping the rendered output identical to before this change.

**How it was verified**:

`make verify-helm-charts` passes. With `grpcHealthPort` unset the rendered output is byte-identical to before this change.

End to end on a GKE cluster with a `gke-l7-regional-external-managed` Gateway and `llm-d-inference-sim` as the model server:

```
helm upgrade --install vllm-sim ./config/charts/llm-d-router-gateway \
  --set provider.name=gke \
  --set httpRoute.create=true \
  --set router.modelServers.matchLabels.app=vllm-sim \
  --set router.epp.grpcHealthPort=9113
```

After the upgrade, the EPP pod restarts and becomes Ready, confirming the binary and the probes moved to 9113 together, and the `vllm-sim-epp` `HealthCheckPolicy` is created targeting the EPP Service. The load balancer health check can be observed with `gcloud compute health-checks list --filter="type=GRPC"`.

image

Fixes #1604

**Release note**:
```release-note
The EPP gRPC health check port is configurable via `router.epp.grpcHealthPort` (default 9003). 
```

### DIFF
--- a/config/charts/llm-d-router-gateway/templates/gke.yaml
+++ b/config/charts/llm-d-router-gateway/templates/gke.yaml
@@ -35,6 +35,30 @@ spec:
         port: {{ $port }}
         grpcServiceName: "/vllm.grpc.engine.VllmEngine/HealthCheck"
     {{- end }}
+{{- $eppHealthPort := int (.Values.router.epp.grpcHealthPort | default 9003) }}
+{{- /* The GKE gateway controller health-checks the EPP backend on gRPC port
+9003 by default, so an explicit HealthCheckPolicy is only needed for other ports. */}}
+{{- if ne $eppHealthPort 9003 }}
+---
+kind: HealthCheckPolicy
+apiVersion: networking.gke.io/v1
+metadata:
+  name: {{ .Release.Name }}-epp
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "llm-d-router.labels" . | nindent 4 }}
+spec:
+  targetRef:
+    group: ""
+    kind: Service
+    name: {{ include "llm-d-router.name" . }}
+  default:
+    config:
+      type: GRPC
+      grpcHealthCheck:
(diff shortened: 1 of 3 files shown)


ahg-g:
looks good, please resolve the conflicts and this can be merged

zetxqx:
> looks good, please resolve the conflicts and this can be merged

@ahg-g  conflicts resolved

ahg-g (approved):

Issue #1606, opened 2026-06-11
[Flow Control]: control-plane priority bands are garbage-collected while still desired (regression from #1354)


LukeAVanDrie:
**What happened**:

After #1354, a priority band defined **only** via an `InferenceObjective` (i.e. not present in the static EPP config) is garbage-collected after a period of inactivity. Nothing re-provisions it until the next reconcile event, so every subsequent request at that priority is demoted to priority 0 and then **rejected** with `priority band not found`.

Observed with a band at priority `-1`:

```
setup.flow-registry  registry/registry.go:523  Successfully deleted priority band  priority=-1
flow-controller      controller/controller.go  Priority band not provisioned, demoting request to priority 0  originalPriority=-1
flow-controller      internal/processor.go     Rejecting item.  flowKey=batch-A:18446744073709551615  error="...: priority band not found"
```

(`18446744073709551615` == `uint64(-1)`; note the rejected item is still at priority **-1**, not 0 — see the demotion note under "Anything else".)

**What you expected to happen**:

A priority band that the control plane still desires — i.e. backed by a live `InferenceObjective` — should remain provisioned regardless of how long it has been idle, and requests at that priority should continue to be served. Inactivity should not silently disable a configured priority.

**How to reproduce it (as minimally and precisely as possible)**:

End-to-end:
1. Enable the Flow Control feature gate.
2. Create an `InferenceObjective` with a priority **not** present in the static EPP config (e.g. `-1`), and send traffic to confirm it works.
3. Leave that objective idle (no traffic, no reconcile events) for longer than `PriorityBandGCTimeout`.
4. Send a request again → it is demoted to priority 0 and rejected with `priority band not found`.

Unit (deterministic): provision a band via the control plane, idle it past `PriorityBandGCTimeout`, run a GC cycle, and observe the band is deleted. A failing regression test (`TestFlowRegistry_PriorityBandGarbageCollection/ShouldNotCollectControlPlaneDesiredBand_AfterInactivity`) will be included with the fix PR.

**Anything else we need to know?**:

_Root cause:_

Band removal has two authorities that disagree:

- `ApplyDesiredPriorities` (control plane) removes bands that are `(not static) ∧ (not desired) ∧ idle` — correct.
- `gcPriorityBands` (timer) removes bands that are `(not static) ∧ idle-past-timeout` — **with no notion of "desired."**

[...]

PR #1608, opened 2026-06-11
Fix/flowcontrol band gc desired

LukeAVanDrie:
**What type of PR is this?**

/kind bug

Fixes a regression from #1354 where a priority band defined only via an `InferenceObjective` (not present in the static EPP config) was garbage-collected after a period of inactivity. Nothing re-provisioned it until the next reconcile, so subsequent requests at that priority were rejected with `priority band not found`.

#1354 moved band provisioning to the control plane but left band *lifecycle* under the old idleness-GC model: the periodic GC only exempted startup bands and had no knowledge of the control plane's desired set, so it reaped still-desired bands once their flows drained.

Three commits:

1. **Retain control-plane-desired bands across GC.** The registry now tracks the last-applied desired set, and a band is protected from collection when it is either a startup band or currently desired (`initialPriorities ∪ desiredPriorities`). `ApplyDesiredPriorities` and the GC path share this predicate at the single physical-deletion chokepoint, so a band is removed only when it is unprotected and idle. Includes a regression test.

2. **Enqueue fallback requests at the fallback priority.** When the requested band is not provisioned, the request was leased at priority 0 but the `FlowItem` still carried the original priority, so the processor looked it up at the missing band and rejected it anyway. The effective (priority-0) request is now threaded through to item creation so the request is leased, distributed, and enqueued consistently at priority 0. Includes a controller regression test.

3. **Address non-blocking review feedback from #1354** (no functional change): collapse a redundant lock acquisition in `ensureFlowInfrastructure`, rename the priority-0 "demotion" to the more accurate "fallback" (it is a promotion for negative priorities), un-shadow the builtin `copy`, and reword a misleading concurrency comment.

Note for reviewers: commit 2 keeps the existing availability-first fallback behavior from #1354 — during the brief window before the control plane provisions a band, a request is served at priority 0 rather than its requested priority. This is only reachable in that warm-up window (priorities are sourced from provisioned objectives). The follow-up issue (#1606) tracks the residual creation race and the broader band-lifecycle simplification.

Part of #1606 

```release-note
Fixed a regression where a priority band defined only via an InferenceObjective (not in the static EPP config) could be garbage-collected after a period of inactivity, causing subsequent requests at that priority to be rejected with "priority band not found".
```

### DIFF
--- a/pkg/epp/flowcontrol/controller/controller.go
+++ b/pkg/epp/flowcontrol/controller/controller.go
@@ -282,6 +282,12 @@ func (fc *FlowController) EnqueueAndWait(
 
 // fallbackRequest wraps a FlowControlRequest to override its flow key, so a request that falls back to a different
 // priority is enqueued under the band that was actually leased rather than its original (unprovisioned) band.
+//
+// Trade-off: downstream consumers see this wrapper, so item.OriginalRequest().FlowKey() reports the fallback
+// priority rather than the requested one — despite the method name. This is intentional, since the item must be
+// leased, distributed, and enqueued consistently at the fallback priority. The originally requested priority
+// therefore survives only in the withConnectionWithFallback log; surfacing it in metrics is left as a follow-up
+// (a dedicated fallback counter labeled with the original priority).
 type fallbackRequest struct {
 	flowcontrol.FlowControlRequest
 	key flowcontrol.FlowKey
--- a/pkg/epp/flowcontrol/controller/controller.go
+++ b/pkg/epp/flowcontrol/controller/controller.go
@@ -239,7 +239,7 @@ func (fc *FlowController) EnqueueAndWait(
 
 	// 2. Acquire a lease for the Flow.
 	// We hold this lease for the entire duration of the request (Distribution + Queueing).
-	err := fc.withConnectionWithDemotion(flowKey, func(conn contracts.ActiveFlowConnection) error {
+	err := fc.withConnectionWithDemotion(req, func(conn contracts.ActiveFlowConnection, effectiveReq flowcontrol.FlowControlRequest) error {
 
 		select { // Non-blocking check on controller lifecycle.
 		case <-fc.parentCtx.Done():
@@ -249,7 +249,9 @@ func (fc *FlowController) EnqueueAndWait(
 		}
 
 		// Attempt to distribute the request once, passing the active connection.
-		item, err := fc.tryDistribution(reqCtx, req, enqueueTime, conn)
+		// effectiveReq carries the demoted flow key when the requested band was not provisioned, so the
+		// item is enqueued under the band that was actually leased.
+		item, err := fc.tryDistribution(reqCtx, effectiveReq, enqueueTime, conn)
 		if err != nil {
 			// Distribution failed terminally (e.g., context cancelled during blocking submit).
 			// The item has already been finalized by tryDistribution.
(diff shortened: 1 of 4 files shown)


LukeAVanDrie:
FYI @RishabhSaini and @gyliu513

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -278,24 +280,44 @@ func (fc *FlowController) EnqueueAndWait(
 	return finalOutcome, err
 }
 
-// withConnectionWithDemotion acquires a flow connection, demoting to priority 0 when the requested band is not yet provisioned.
-func (fc *FlowController) withConnectionWithDemotion(
-	key flowcontrol.FlowKey,
-	fn func(conn contracts.ActiveFlowConnection) error,
+// fallbackRequest wraps a FlowControlRequest to override its flow key, so a request that falls back to a different
+// priority is enqueued under the band that was actually leased rather than its original (unprovisioned) band.
+type fallbackRequest struct {
```
gyliu513, line 291: Once the key is rewritten, the original priority becomes invisible downstream: RecordFlowControlRequestEnqueueDuration in internal/processor.go labels fallback requests as priority "0", indistinguishable from genuine priority-0 traffic, and the only remaining signal is the V(DEFAULT) log below. Since "requests not served at their requested priority" is exactly the degraded mode this PR is about, consider a dedicated fallback counter (labeled with the original priority) so operators can alert on it, fine as a follow-up under #1606 if you prefer.

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -278,24 +280,44 @@ func (fc *FlowController) EnqueueAndWait(
...
+	flowcontrol.FlowControlRequest
+	key flowcontrol.FlowKey
+}
+
+func (r fallbackRequest) FlowKey() flowcontrol.FlowKey { return r.key }
```
gyliu513, line 296: With this override, item.OriginalRequest() no longer returns the original request on the fallback path, its FlowKey reports the rewritten priority. The new controller test relies on exactly that, so it's clearly intended, but worth a comment here noting the trade-off, since the method name suggests otherwise.

pkg/epp/flowcontrol/registry/registry_test.go
```diff
@@ -926,6 +929,41 @@ func TestFlowRegistry_PriorityBandGarbageCollection(t *testing.T) {
...
+		_, exists := h.fr.config.PriorityBands[desiredPrio]
+		h.fr.mu.RUnlock()
+		require.True(t, exists,
+			"Control-plane-desired band %d must survive GC after inactivity", desiredPrio)
+
+		// A follow-up request to the still-desired band must not be rejected/demoted.
+		err := h.fr.WithConnection(key, func(conn contracts.ActiveFlowConnection) error { return nil })
+		require.NoError(t, err,
+			"Request to a still-desired band must succeed after inactivity")
+		assert.NotErrorIs(t, err, contracts.ErrPriorityBandNotFound)
```
gyliu513: require.NoError above already fails the test on any error, does this still needed?

pkg/epp/flowcontrol/controller/controller.go
LukeAVanDrie: I think we can do this as a followup. The primary issue this PR addresses is the premature GC of controller-desired bands (commit 1). The controller changes here (commit 2) are for a very unlikely race condition, and I think the impact of that is secondary.

ahg-g:
I am not following why we have GC for "Priority Bands". A priority band is defined in a deterministic manner: either via static configuration or via infObj. In the former, the priority never gets deleted; in the latter we create and delete the band in tandem with creating or deleting the infObj. Are we not designing the creation and deletion of bands in this way?

LukeAVanDrie:
> I am not following why we have GC for "Priority Bands". A priority band is defined in a deterministic manner: either via static configuration or via infObj. In the former, the priority never gets deleted; in the latter we create and delete the band in tandem with creating or deleting the infObj. Are we not designing the creation and deletion of bands in this way?

Agreed, band GC shouldn't be its own thing. It's a leftover from when bands were created JIT on the request path alongside flows; #1354 moved provisioning to the control plane but left the GC behind.

However, there is a catch with deleting strictly in tandem with the infObj. The band can still have in-flight requests when its infObj is deleted, so we can't drop it synchronously, it has to drain to idle first. Reconciles are edge-triggered, so if it's still active at delete time, nothing re-checks it once it goes idle except the periodic GC. That's all GC is doing now.

This PR doesn't remove GC, it just makes it respect the desired set so it stops reaping bands we still want. Removing it entirely is the followup (e.g., re-evaluate a band when its last flow drains instead of on a timer). I have more details on a longer-term, cleaner solution in #1606.

pkg/epp/flowcontrol/controller/controller.go
LukeAVanDrie: Expanded on this in the `fallbackRequest` go-doc.

pkg/epp/flowcontrol/registry/registry_test.go
LukeAVanDrie: Removed, good catch!

ahg-g (approved):

PR #1609, opened 2026-06-11
refactor(flowcontrol): improve observability logging

RishabhSaini:
## Summary
- Rename `reqID` to `requestID` across processor logs for consistency
- Bump capacity rejection from DEBUG to DEFAULT with utilization stats (`totalLen`, `totalCapacityRequests`, `totalByteSize`, `totalCapacityBytes`)
- Add non-dispatch outcome log in `EnqueueAndWait` so TTL expirations and context cancellations are visible at DEFAULT
- Add saturation value to HoL blocking log
- Clarify vague rejection messages
- Demote periodic sweep log to TRACE, bump shutdown eviction to DEFAULT
- Include evicted `requestIDs` in eviction complete log

## Test plan
- [x] Existing unit tests pass (`go test -race ./pkg/epp/flowcontrol/controller/... ./pkg/epp/flowcontrol/eviction/`)
- [x] No behavior change — logging only

### DIFF
--- a/pkg/epp/flowcontrol/controller/controller.go
+++ b/pkg/epp/flowcontrol/controller/controller.go
@@ -272,7 +272,13 @@ func (fc *FlowController) EnqueueAndWait(
 	// return a valid rejection outcome.
 	// In the success case (where the closure ran), finalOutcome is set inside the closure.
 	if err != nil && finalOutcome == types.QueueOutcomeNotYetFinalized {
-		return types.QueueOutcomeRejectedOther, fmt.Errorf("%w: %w", types.ErrRejected, err)
+		finalOutcome = types.QueueOutcomeRejectedOther
+		err = fmt.Errorf("%w: %w", types.ErrRejected, err)
+	}
+
+	if finalOutcome != types.QueueOutcomeDispatched {
+		fc.logger.V(logutil.VERBOSE).Info("Request dropped",
+			"requestID", req.ID(), "flowKey", flowKey, "outcome", finalOutcome, "err", err)
 	}
 
 	return finalOutcome, err
@@ -415,7 +421,7 @@ func (fc *FlowController) distributeRequest(
 	}
 
 	// processor is busy. Attempt a single blocking submission to the candidate.
-	fc.logger.V(logutil.TRACE).Info("Processor is busy, attempting blocking submit", "requestID", reqID)
+	fc.logger.V(logutil.DEBUG).Info("Processor is busy, attempting blocking submit", "requestID", reqID)
 	err := fc.processor.SubmitOrBlock(ctx, item)
 	if err != nil {
 		return types.QueueOutcomeRejectedOther, fmt.Errorf("%w: request not accepted: %w", types.ErrRejected, err)
--- a/pkg/epp/flowcontrol/controller/internal/processor.go
+++ b/pkg/epp/flowcontrol/controller/internal/processor.go
@@ -256,7 +256,7 @@ func (sp *Processor) enqueue(item *FlowItem) {
 	// The ultimate guarantee of cleanup for any races is the runCleanupSweep mechanism.
 	if finalState := outcome; finalState != nil {
 		sp.logger.V(logutil.TRACE).Info("Item finalized externally before processing, discarding.",
-			"outcome", finalState.Outcome, "err", finalState.Err, "flowKey", key, "reqID", req.ID())
+			"outcome", finalState.Outcome, "err", finalState.Err, "flowKey", key, "requestID", req.ID())
 		return
 	}
 
@@ -264,7 +264,7 @@ func (sp *Processor) enqueue(item *FlowItem) {
 	managedQ, err := sp.registry.ManagedQueue(key)
 	if err != nil {
 		finalErr := fmt.Errorf("configuration error: failed to get queue for flow key %s: %w", key, err)
-		sp.logger.Error(finalErr, "Rejecting item.", "flowKey", key, "reqID", req.ID())
+		sp.logger.Error(finalErr, "Rejecting request, queue lookup failed", "flowKey", key, "requestID", req.ID())
 		item.FinalizeWithOutcome(types.QueueOutcomeRejectedOther, fmt.Errorf("%w: %w", types.ErrRejected, finalErr))
 		return
 	}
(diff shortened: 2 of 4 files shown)


pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -256,32 +256,35 @@ func (sp *Processor) enqueue(item *FlowItem) {
...
 		return
 	}
 
 	// --- Capacity Check ---
 	// This check is safe because it is performed by the single-writer Run goroutine.
 	if !sp.hasCapacity(key.Priority, req.ByteSize()) {
-		sp.logger.V(logutil.DEBUG).Info("Rejecting request, queue at capacity",
-			"flowKey", key, "reqID", req.ID(), "reqByteSize", req.ByteSize())
+		stats := sp.registry.Stats()
+		sp.logger.V(logutil.DEFAULT).Info("Rejecting request, queue at capacity",
```
LukeAVanDrie: Hmm, actually, thoughts on verbose logging here?

This will trigger whenever we have backpressure not absorbed into the queues. When this log triggers, it likely won't be one-off, but instead for a flood of requests.

RishabhSaini: agreed!

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -256,32 +256,35 @@ func (sp *Processor) enqueue(item *FlowItem) {
...
-		sp.logger.Error(finalErr, "Rejecting item.", "flowKey", key, "reqID", req.ID())
+		sp.logger.Error(finalErr, "Rejecting request, priority band lookup failed", "flowKey", key, "requestID", req.ID())
 		item.FinalizeWithOutcome(types.QueueOutcomeRejectedOther, fmt.Errorf("%w: %w", types.ErrRejected, finalErr))
 		return
 	}
 
 	// --- Capacity Check ---
 	// This check is safe because it is performed by the single-writer Run goroutine.
 	if !sp.hasCapacity(key.Priority, req.ByteSize()) {
+		stats := sp.registry.Stats()
```
LukeAVanDrie: `hasCapacity` already calls `sp.registry.Stats()` internally, so this adds a second full snapshot back-to-back on every capacity rejection, and `Stats()` isn't cheap for what we need. It takes the global registry `RLock` and allocates a `map[int]PriorityBandStats` by ranging every band when the check only needs the totals + the one band for this priority. All on the single Run goroutine that does all dispatching.

Rather than gate the call, I'd suggest we delete `Stats()` entirely and fold the capacity check into the registry. It's a reporting-shaped API (whole-registry map snapshot) doing a single-band O(1) decision, and these two hot-path sites are its only production callers; nothing else reads it. Replace it with an allocation-free method that returns the decision and the numbers for the log:

```go
// contracts/registry.go — replaces `Stats() AggregateStats`
HasCapacity(priority int, itemByteSize uint64) (bool, BandCapacity)

type BandCapacity struct {  // flat, single-band — no map
    TotalByteSize, TotalCapacityBytes uint64
    TotalLen, TotalCapacityRequests   uint64
    BandByteSize, BandCapacityBytes   uint64
    BandLen, BandCapacityRequests     uint64
}
```

Impl reads the two total atomics + a single `perPriorityBandStats.Load(priority)` (O(1), no map, no Range); brief `RLock` only for config limits. Call site collapses to:

```go
ok, capStats := sp.registry.HasCapacity(key.Priority, req.ByteSize())
if !ok {
    if logger := sp.logger.V(logutil.DEBUG); logger.Enabled() {
        logger.Info("Rejecting request, queue at capacity",
            "flowKey", key, "requestID", req.ID(), "reqByteSize", req.ByteSize(),
            "totalLen", capStats.TotalLen, "totalCapacityRequests", capStats.TotalCapacityRequests,
            "totalByteSize", capStats.TotalByteSize, "totalCapacityBytes", capStats.TotalCapacityBytes)
    }
    item.FinalizeWithOutcome(types.QueueOutcomeRejectedCapacity,
        fmt.Errorf("%w: %w", types.ErrRejected, types.ErrQueueAtCapacity))
    return
}
```

`processor.hasCapacity` goes away; its arithmetic moves into the registry method.

This does involve updating the following contract method + `AggregateStats` type, the mock, and a few `Stats()`-based test assertions.

If you'd rather keep this PR strictly logging, the minimal version is: have `hasCapacity` return its snapshot so we don't call `Stats()` twice, and wrap the log in `if logger := sp.logger.V(logutil.DEBUG); logger.Enabled()`. But, I was intending on doing this refactoring anyways for some mem alloc perf concerns I found while microbenchmarking.

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -272,7 +272,13 @@ func (fc *FlowController) EnqueueAndWait(
 	// return a valid rejection outcome.
 	// In the success case (where the closure ran), finalOutcome is set inside the closure.
 	if err != nil && finalOutcome == types.QueueOutcomeNotYetFinalized {
-		return types.QueueOutcomeRejectedOther, fmt.Errorf("%w: %w", types.ErrRejected, err)
+		finalOutcome = types.QueueOutcomeRejectedOther
+		err = fmt.Errorf("%w: %w", types.ErrRejected, err)
+	}
+
+	if finalOutcome != types.QueueOutcomeDispatched {
+		fc.logger.V(logutil.DEFAULT).Info("Request dropped",
```
LukeAVanDrie: I would also move this from DEFAULT to VERBOSE for similar reasons I mentioned in the other thread.

LukeAVanDrie: Also, consider adding `"err", err`. For `RejectedOther`, the reason lives in `err`, and without it this log can't explain why an "other" rejection happened.

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -518,8 +523,8 @@ func (sp *Processor) evictAll() {
 
 			// Finalization is idempotent; safe to call even if already finalized externally.
 			item.FinalizeWithOutcome(outcome, errShutdown)
-			logger.V(logutil.TRACE).Info("Item evicted during shutdown.",
-				"reqID", item.OriginalRequest().ID())
+			logger.V(logutil.DEFAULT).Info("Request evicted during shutdown.",
```
LukeAVanDrie: Non-blocking nit: each item evicted here also has a blocked `EnqueueAndWait` goroutine that will log `"Request dropped"` once it unblocks . So every queued request gets two log lines on shutdown. I think we should centralize the per-outcome log in `finalizeInternal` (in `item.go`). It would make this line redundant and give exactly-once logging across all paths.

LukeAVanDrie (changes requested):
Nice change! Outside of a couple nits, my main concern is the perf overhead of calling `Stats` a second time. My suggestion below expands the scope of this PR, so I am happy to defer it and tackle as a follow up. If I recall correctly, `Stats` was a top 5 issue when I was doing some microbenchmarking. It's not _terrible_, but the method really doesn't need to exist as  a reporting API as we already capture all this in our prometheus metrics now.

pkg/epp/flowcontrol/controller/controller.go
RishabhSaini: yes agreed, updated

pkg/epp/flowcontrol/controller/internal/processor.go
RishabhSaini: agreed, updated

pkg/epp/flowcontrol/controller/internal/processor.go
RishabhSaini: Agreed on the full refactor Stats() is a reporting API doing a hot-path decision. I've taken the minimal version for this PR: hasCapacity now returns (bool, AggregateStats) so the snapshot is reused, no double call.

LukeAVanDrie (approved):
/lgtm

thanks!

PR #1612, opened 2026-06-11
fix(flowcontrol): map shutdown-drained requests to 503

RishabhSaini:
## Summary
- Return ServiceUnavailable (503) instead of Internal (500) when requests are evicted or rejected due to `ErrFlowControllerNotRunning` during shutdown
- Add table test covering every `QueueOutcome` → HTTP status mapping

## Context
During a rolling update, `evictAll` finalizes queued requests as `EvictedOther` with `ErrFlowControllerNotRunning`. The previous mapping treated all `EvictedOther`/`RejectedOther` as `Internal` (500), which triggers false alerts. Shutdown-drained requests are transient unavailability, not internal errors.

## Test plan
- [x] Table test: dispatched→nil, capacity→429, TTL→503, cancel→503, shutdown→503, internal→500
- [x] Existing admission tests pass

### DIFF
--- a/pkg/epp/requestcontrol/admission.go
+++ b/pkg/epp/requestcontrol/admission.go
@@ -18,6 +18,7 @@ package requestcontrol
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -231,7 +232,9 @@ func translateFlowControlOutcome(outcome types.QueueOutcome, err error) error {
 	case types.QueueOutcomeEvictedContextCancelled:
 		return errcommon.Error{Code: errcommon.ServiceUnavailable, Msg: "client disconnected: " + msg, Headers: map[string]string{errcommon.RequestDroppedReasonHeaderKey: string(errcommon.RequestDroppedReasonContextCancelled)}}
 	case types.QueueOutcomeRejectedOther, types.QueueOutcomeEvictedOther:
-		// No x-removal-reason header: these are internal/unexpected failures, not a specific removal policy.
+		if errors.Is(err, types.ErrFlowControllerNotRunning) {
+			return errcommon.Error{Code: errcommon.ServiceUnavailable, Msg: "flow controller shutting down: " + msg, Headers: map[string]string{errcommon.RequestDroppedReasonHeaderKey: string(errcommon.RequestDroppedReasonShuttingDown)}}
+		}
 		return errcommon.Error{Code: errcommon.Internal, Msg: "internal flow control error: " + msg}
 	default:
 		return errcommon.Error{Code: errcommon.Internal, Msg: "unhandled flow control outcome: " + msg}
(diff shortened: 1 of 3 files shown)


LukeAVanDrie:
Let's also make sure we update this section of the technical docs in llm-d

https://github.com/llm-d/llm-d/blob/main/docs/architecture/core/router/epp/flow-control.md#failure-modes--error-mapping

pkg/epp/requestcontrol/admission.go
```diff
@@ -231,7 +232,9 @@ func translateFlowControlOutcome(outcome types.QueueOutcome, err error) error {
 	case types.QueueOutcomeEvictedContextCancelled:
 		return errcommon.Error{Code: errcommon.ServiceUnavailable, Msg: "client disconnected: " + msg, Headers: map[string]string{errcommon.RequestDroppedReasonHeaderKey: string(errcommon.RequestDroppedReasonContextCancelled)}}
 	case types.QueueOutcomeRejectedOther, types.QueueOutcomeEvictedOther:
-		// No x-removal-reason header: these are internal/unexpected failures, not a specific removal policy.
+		if errors.Is(err, types.ErrFlowControllerNotRunning) {
+			return errcommon.Error{Code: errcommon.ServiceUnavailable, Msg: "flow controller shutting down: " + msg, Headers: map[string]string{errcommon.RequestDroppedReasonHeaderKey: string(errcommon.RequestDroppedReasonEvicted)}}
+		}
```
LukeAVanDrie, line 237: The 503 mapping is right, but `RequestDroppedReasonEvicted` is documented as *"request was dispatched to an inference server and then killed."* Shutdown drops never reach a server:
- `EvictedOther` here is `evictAll` draining the queue
- `RejectedOther`+`ErrFlowControllerNotRunning` is Submit-during-shutdown (never queued)
 
So `evicted` miscategorizes these for anything keying on `x-llm-d-request-dropped-reason`.

Suggest adding a Rejected-family reason and using it:

```go
// pkg/common/error/error.go, in the "Rejected" group
RequestDroppedReasonShuttingDown RequestDroppedReason = "rejected-shutting-down"
```

```suggestion
		if errors.Is(err, types.ErrFlowControllerNotRunning) {
			return errcommon.Error{Code: errcommon.ServiceUnavailable, Msg: "flow controller shutting down: " + msg, Headers: map[string]string{errcommon.RequestDroppedReasonHeaderKey: string(errcommon.RequestDroppedReasonShuttingDown)}}
		}
```

LukeAVanDrie (changes requested):

pkg/epp/requestcontrol/admission.go
RishabhSaini: good catch. agreed, updated

LukeAVanDrie (approved):

ahg-g (approved):
Needs a top level approver because the PR touches a code path outside flowcontrol

LukeAVanDrie:
> Let's also make sure we update this section of the technical docs in llm-d
> 
> https://github.com/llm-d/llm-d/blob/main/docs/architecture/core/router/epp/flow-control.md#failure-modes--error-mapping

Created https://github.com/llm-d/llm-d/pull/1801

PR #1618, opened 2026-06-12
fix(tokenizer): label estimate multimodal features by modality

### LINKED ISSUE #1617: estimate token-producer flattens multimodal modality to "image"

The `token-producer` estimate backend should label multimodal features with their actual modality (`image`/`audio`/`video`), matching the HTTP `/render` backend, instead of hardcoding `ModalityImage`.

**Why is this needed**:

The estimate backend sets `Modality: ModalityImage` on every multimodal item, so audio and video are counted as `image` in `encoder_cache_queries_total` / `encoder_cache_hits_total`. Per-modality cache observability is wrong and diverges from the `/render` backend, which carries the real modality. (The vLLM gRPC engine parser has the same symptom but is blocked on a proto field with no per-item modality — out of scope here.)

hexfusion:
**What type of PR is this?**

/kind bug

The token-producer `estimate` backend hardcoded `ModalityImage` for every multimodal item, so audio and video were reported as `image` in `encoder_cache_queries_total` / `encoder_cache_hits_total` diverging from the vLLM `/render` backend, which carries the real modality. This threads the actual modality through `appendMMAsset` and adds `ModalityAudio`/`ModalityVideo` consts. Routing is unaffected (scoring keys on the content hash, not modality); this corrects per-modality cache observability.

Fixes #1617

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/interface/requesthandling/types.go
+++ b/pkg/epp/framework/interface/requesthandling/types.go
@@ -34,8 +34,13 @@ const nilStr = "<nil>"
 // Modality identifies the type of multimodal content in a prompt.
 type Modality string
 
-// ModalityImage is the only currently supported modality.
-const ModalityImage Modality = "image"
+// Modality values match the model-server's multimodal hash keys so labels agree
+// across backends.
+const (
+	ModalityImage Modality = "image"
+	ModalityAudio Modality = "audio"
+	ModalityVideo Modality = "video"
+)
 
 // RequestPayload represents a strongly-typed unmarshaled request payload or raw bytes.
 type RequestPayload interface {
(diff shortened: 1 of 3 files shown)


ahg-g:
nice, pls rebase.

hexfusion:
> nice, pls rebase.

done

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -34,8 +34,12 @@ const nilStr = "<nil>"
 // Modality identifies the type of multimodal content in a prompt.
 type Modality string
 
-// ModalityImage is the only currently supported modality.
-const ModalityImage Modality = "image"
+// Modality values match vLLM's mm_hashes keys so labels agree across backends.
```
vMaroon: nit: this is technically right, but as we work towards standardization across vLLM and SGLang, perhaps can say "model-server" or have a standalone description that does not refer to the latter.

hexfusion: thats fair

hexfusion: updated

ahg-g (approved):

PR #1619, opened 2026-06-12
fix(sidecar): add startup validation for CLI port and replica options

weizhoublue:
**What type of PR is this?**

/kind bug

add startup validation for CLI port and replica options

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/sidecar/proxy/options.go
+++ b/pkg/sidecar/proxy/options.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"math"
 	"net/url"
 	"os"
 	"slices"
@@ -510,6 +511,27 @@ func validateWideEPHosts(flag string, hosts []string, dpSize, dpLocal int) error
 // Validate checks the Options for invalid or conflicting values.
 // Complete must be called before Validate.
 func (opts *Options) Validate() error {
+	// Validate data parallel size
+	if opts.DataParallelSize < 1 {
+		return fmt.Errorf("--data-parallel-size must be a positive integer, got %d", opts.DataParallelSize)
+	}
+
+	port, err := strconv.Atoi(opts.Port)
+	if err != nil {
+		return fmt.Errorf("--port must be a valid integer, got %q", opts.Port)
+	}
+	if err := validatePortRange(port, opts.DataParallelSize); err != nil {
+		return fmt.Errorf("--port %w", err)
+	}
+
+	vllmPort, err := strconv.Atoi(opts.vllmPort)
+	if err != nil {
+		return fmt.Errorf("--vllm-port must be a valid integer, got %q", opts.vllmPort)
+	}
+	if err := validatePortRange(vllmPort, opts.DataParallelSize); err != nil {
+		return fmt.Errorf("--vllm-port %w", err)
+	}
+
 	// Validate KV connector
 	if _, ok := supportedKVConnectors[opts.KVConnector]; !ok {
 		return fmt.Errorf("--kv-connector must be one of: %s", supportedKVConnectorNamesStr)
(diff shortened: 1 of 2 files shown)


hexfusion:
@weizhoublue thanks for doing this I believe there was an issue we could close related to this could you link as fixes :pray:

weizhoublue:
> @weizhoublue thanks for doing this I believe there was an issue we could close related to this could you link as fixes 🙏

Sorry,  I browsed through all existing issue titles but couldn't find one about validation for the CLI flags.
Could you kindly give me some keywords or issue link ? thanks

hexfusion:
I thinkk its somewhat related to https://github.com/llm-d/llm-d-router/issues/979

hexfusion:
or maybe you would like to take that one over? thank you

weizhoublue:
> or maybe you would like to take that one over? thank you

Thanks, I got it.

I think #979 is a different issue. This PR validates startup CLI options, while #979 is about runtime SSRF allowlist validation in allowlist.go, specifically enforcing host:port tuples for X-Prefiller-Host-Port. 

I submitted a separate PR https://github.com/llm-d/llm-d-router/pull/1647  for #979.

pkg/sidecar/proxy/options.go
```diff
@@ -326,6 +326,29 @@ func (opts *Options) Complete() error {
...
+
+	// Validate vllmPort
+	vllmPortNum, err := strconv.Atoi(opts.vllmPort)
+	if err != nil {
+		return fmt.Errorf("--vllm-port must be a valid integer, got %q", opts.vllmPort)
+	}
+	if vllmPortNum < 1 || vllmPortNum > 65535 {
+		return fmt.Errorf("--vllm-port must be between 1 and 65535, got %d", vllmPortNum)
+	}
+
```
elevran: suggest extracting to a shared 
```go
func isValidPort(value, flagName string) (int, error)
```

elevran:
@weizhoublue - working on this?

weizhoublue:
> @weizhoublue - working on this?

Apologize for my late response.
I have updated the code according to your good suggestion.
Thanks

pkg/sidecar/proxy/options.go
weizhoublue: That is good. 
Done.

pkg/sidecar/proxy/options.go
```diff
@@ -326,6 +326,26 @@ func (opts *Options) Complete() error {
...
+	// Validate data parallel size
+	if opts.DataParallelSize < 1 {
+		return fmt.Errorf("--data-parallel-size must be a positive integer, got %d", opts.DataParallelSize)
+	}
+	if opts.DataParallelSize-1 > 65535-portNum {
+		return errors.New("--port plus --data-parallel-size must not exceed 65535")
+	}
+	if opts.DataParallelSize-1 > 65535-vllmPortNum {
+		return errors.New("--vllm-port plus --data-parallel-size must not exceed 65535")
+	}
```
elevran: ni: this check can also be factored.

Or combined to check both the lowest and highest ports in the range.
```go
func isValidPortAndRange(opts.Port,  opts.DataParallelSize, "--port") (int, error) {
   // calls isValidPort(port) and isValidPort(port+dataparallelsize -1) - if base and top ports are valid, ports in the range are valid too
   // returns the base port and an error
}
```

elevran (dismissed):
/hold
Thanks for fixing.
Approved from my point of view, let a nit and hold in case you want to address. Otherwise just cancel the hold

weizhoublue:
> /hold Thanks for fixing. Approved from my point of view, let a nit and hold in case you want to address. Otherwise just cancel the hold

Updated as suggested. 
Please take another look, thanks

pkg/sidecar/proxy/options.go
```diff
@@ -593,6 +605,33 @@ func (opts *Options) Validate() error {
...
+		offset := dataParallelSize - 1
+		if offset > int(^uint(0)>>1)-port {
+			return 0, fmt.Errorf("%s plus --data-parallel-size must not exceed 65535", flagName)
+		}
+		if err := isValidPort(port + offset); err != nil {
+			return 0, fmt.Errorf("%s %w", flagName, err)
+		}
+	}
+	return port, nil
+}
```
elevran: There are a few improvements that can be done for readability:
- define a constant for MaxPort value instead of hardcoding 65535
- the `offset > int(^uint(0)>>1)-port` check is a safeguard against integer overflow during the addition, but since `MaxUint16` is much smaller than the max value of an `int` (likely 64b), the failure is caught by the subsequent isValidPort call - not the overflow.

```go
func validatePortRange(startPort int, rangeSize int) error {
	const maxPort = math.MaxUint16
    
    if rangeSize <= 0 || rangeSize >= MaxPort {
		return fmt.Errorf("invalid port range")
	}

	if startPort < 1 || startPort > MaxPort {
		return fmt.Errorf("start port %d is out of valid range [1, %d]", startPort, MaxPort)
	}

	endPort := startPort + rangeSize - 1
	if endPort > MaxPort {
		return fmt.Errorf("port range [%d, %d] exceeds maximum port value", startPort, endPort)
	}

	return nil
}
```
note that the error strings above are missing the flag names, etc. - focused on the logic

weizhoublue: Thanks for detailed suggestion. 
Updated code as suggested.

elevran (approved):

Issue #977, opened 2026-05-03
Create a generic endpoint attributes scorer


liu-cong:
This is porting over this feature request in the legacy repo:  https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2201

@LukeAVanDrie Had a POC earlier https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2237

**What would you like to be added**:

A scorer that's configurable to score based on an arbitrary metric from the model server. To do this we need to configure: the metric spec (name, label, type, etc.) and the scoring algo. To begin with, we will only support "gauge" metric, and  a "linear normalization" algo which produces a normalized [0-1] score based on either `lower_is_better` or `higher_is_better`.

An example:

```
plugins:
- type: endpoint-attribute-scorer
  parameters:
      attribute: "num_requests_running"
      attributeType: "scalar" 
      algo: 
        type: "linear_lower_is_better" # initially support linear_lower_is_better and linear_higher_is_better
        normalization:
           fixedRange: 
              min: 0
              max: 100
          # another option is adaptiveRange
          #  the score is normalized against an adaptive [min, max] range between all the available endpoints.
          # adaptiveRange:
```

**Why is this needed**:

While we implemented several in-tree scorers based on well-known metrics such as kv cache utilization, some use cases may desire a different metric. Having a generic metric scorer avoids the friction of writing a new scorer each time.

One potential use case is that running requests is desired than queue depth for latency sensitive applications. 

--EDIT 06/29--
Metrics are just a type of attributes of an endpoint. To make this more general, we

liu-cong:
We need a way for plugins to register metrics they consume, captured in  https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2785

We can then build the scorer logic on top of it.

richl9:
Hi @liu-cong, I would like to work on a scoped first step for #977.

Based on the original feature request and the earlier POC feedback, I’m thinking of splitting the work into the following pieces:

- [x] Add support for configured custom metrics in the core metrics extractor.
- [x] Store custom metric values as endpoint attributes under configured internal keys, instead of adding new fixed fields to Metrics.
- [ ] Add a metric-scorer that reads one configured numeric endpoint attribute.
- [ ] Support linear normalization across candidate endpoints.
- [ ] Support lower_is_better and higher_is_better directions.

I opened a scoped issue #1454 for the first item so boundaries can be reviewed first before implementation.

[...]

PR #1620, opened 2026-06-12
Add generic metric scorer for configured endpoint attributes

### LINKED ISSUE #1594: Add generic metric scorer for configured endpoint attributes

Add a generic `metric-scorer` scheduling plugin that scores endpoints using one configured numeric endpoint attribute.

This is the next scoped step toward #977 after #1454. The custom metric extraction layer already stores configured model-server metric values as endpoint attributes. This issue covers only the scorer that consumes one of those attributes.

Scope:

- Read one configured endpoint attribute by `attributeKey`.
- Support numeric custom metric attributes produced by the core metrics extractor.
- Support linear normalization across candidate endpoints.
- Support both scoring directions (lower_is_better and higher_is_better)
- Return a neutral(same) score when all candidate endpoints have the same metric value.
- Add tests for config validation, missing attributes, equal values, and both scoring directions.
- Add README/config docs for the new scorer.

Example configuration:

```yaml
plugins:
- type: metric-scorer
  parameters:
    attributeKey: "custom.queue_depth"
    normalization:
      type: "linear"
      direction: "lower_is_better"

### PR DESCRIPTION
**Pull request by sudoalok**
Adds a `metric-scorer` scheduling plugin that scores endpoints based on one configured endpoint attribute (produced by the custom metrics extraction layer). Normalization is linear across the candidates and both lower_is_better and higher_is_better directions are supported.

A few decisions I made along the way, open to changing any of them:
- if all endpoints have the same value they all get 1.0, same as queue-scorer does
- normalization.type defaults to linear since that's the only supported type for now, anything else fails validation
- direction has no default and must be set explicitly, felt safer than guessing for a generic scorer
- endpoints that don't have the attribute score 0.0 and are left out of the min/max calculation

Fixes #1594

### CODE REVIEW

File: pkg/epp/framework/plugins/scheduling/scorer/metric/metric.go [hunk 1]
```diff
@@ -0,0 +1,170 @@
+/*
+Copyright 2025 The Kubernetes Authors.
```
liu-cong:
```suggestion
Copyright 2026 The Kubernetes Authors.
```

pkg/epp/framework/plugins/scheduling/scorer/metric/metric.go
```diff
@@ -0,0 +1,170 @@
...
+
+// Consumes returns the list of data that is consumed by the plugin.
+func (s *MetricScorer) Consumes() map[string]any {
+	return map[string]any{
+		s.attributeKey: attrmetrics.ScalarMetricValue(0),
+	}
+}
+
+// WithName sets the name of the scorer.
+func (s *MetricScorer) WithName(name string) *MetricScorer {
```
liu-cong: let's remove this, name should be initialized in the NewMetricScorer method direclty

pkg/epp/framework/plugins/scheduling/scorer/metric/metric.go
liu-cong, line 1: I updated description in https://github.com/llm-d/llm-d-router/issues/977

 Metric is just a special type of attributes. Let's call this `endpoint_attribute_scorer` to make it more generic.

pkg/epp/framework/plugins/scheduling/scorer/metric/metric.go
```diff
@@ -0,0 +1,170 @@
...
+	directionLowerIsBetter  = "lower_is_better"
+	directionHigherIsBetter = "higher_is_better"
+)
+
+// compile-time type assertion
+var _ fwksched.Scorer = &MetricScorer{}
+
+type normalizationParameters struct {
+	Type      string `json:"type"`
+	Direction string `json:"direction"`
```
liu-cong: Suggest something like this: 

```
// the score  is normalized against a fixed [min, max] range. 
type fixedRange struct {
   min, max float64
}

// // the score is normalized against an adaptive [min, max] range between all the available endpoints.
type adaptiveRange struct {
}


type normalization struct {
   // Only of the following normalization strategies can be chosen
   fixedRange *fixedRange
   adaptiveRange *adaptiveRange
}

type algo struct {
    Type      string `json:"type"` # the algorithm type, we will initially support `linear_lower_is_better` and `linera_higher_is_better`
    Normalization normalization `json:"normalization"`
}
```

**Rationale**
* Merge the `linear` and `direction` into a single algo type to allow more flexible algorithms in the future. For example, an algo that may not have a "direction"
* Support existing normalization strategies, fixed range (e.g., kv cache unitiozation) and adaptive (queue depth).

pkg/epp/framework/plugins/scheduling/scorer/metric/metric.go
sudoalok: bumping to 2026 here and in metric_test.go.

pkg/epp/framework/plugins/scheduling/scorer/metric/metric.go
sudoalok: make sense removing WithName and setting the name directly in the constructor (NewEndpointAttributeScorer takes name now).

pkg/epp/framework/plugins/scheduling/scorer/metric/metric.go
sudoalok: sounds good. [#977](https://github.com/sudoalok/llm-d/issues/977)'s example config uses endpoint-attribute-scorer (hyphen, matches the existing type-string style), so i'll go with that.

pkg/epp/framework/plugins/scheduling/scorer/metric/metric.go
sudoalok: i like folding direction into the algo type, leaves room for direction-less algos later. two quick things before i restructure:

type string: your comment uses linear_lower_is_better / linear_higher_is_better, but the #977 example has type: linear_normalization with a separate direction. which one is canonical?
normalization: do you want fixedRange in this PR, or land the algo structure + adaptiveRange (the current behavior) now and add fixedRange as a follow-up? happy either way,

liu-cong:
@sudoalok a gentle ping on this :)

liu-cong (approved):

PR #1633, opened 2026-06-12
fix: Make TokenizedPrompt.MultiModalFeatures per-prompt

### LINKED ISSUE #1584: Make `MultiModalFeatures` per-prompt in `TokenizedPrompt`
## Summary

`TokenizedPrompt.MultiModalFeatures` is a single flat list, but `PerPromptTokens` supports multiple prompts. The MM feature offsets are only valid relative to one token array, forcing consumers like `computeBlockKeys` to make assumptions about which prompt the features belong to. MultiModalFeatures should be per-prompt to match the shape of `PerPromptTokens`.

## Context

**Depends on**: PR #860, which introduces `PerPromptTokens [][]uint32` as the sole token field in `TokenizedPrompt`.

After PR #860, `TokenizedPrompt` has:

```
type TokenizedPrompt struct {
    PerPromptTokens    [][]uint32
    MultiModalFeatures []MultiModalFeature  // singleton, offsets relative to PerPromptTokens[0]
    CacheSalt          string
}
```

Consumers that need to associate MM features with token arrays must use positional assumptions (e.g. `if i == 0 { mmf = tp.MultiModalFeatures }`). This couples the block-key computation in `preciseprefixcache/blockkeys.go` to upstream tokenizer behavior;  it has to know that MM features always belong to the first prompt.

If a future protocol supports both multi-prompt and multimodal content, the current structure cannot represent per-prompt MM features, and the consumer assumption would silently drop features for prompts beyond the first.

## Proposed change

Replace the singleton list with a per-prompt structure:

```go
type TokenizedPrompt struct {
    PerPromptTokens    [][]uint32
    MultiModalFeatures [][]MultiModalFeature  // one list per prompt, indexed like PerPromptTokens
    CacheSalt          string
}
```

Single-prompt requests (chat, generate) would use a length-1 outer slice. Consumers iterate `PerPromptTokens` and `MultiModalFeatures` in lockstep without needing to know which protocols support multimodal.

albertoperdomo2:
**What type of PR is this?**

/kind bug
/kind feature

## What this PR does

Introduces a `PromptTokens` struct that bundles token IDs and multimodal features per prompt, replacing the two parallel slices on `TokenizedPrompt` (`PerPromptTokens [][]uint32` and `MultiModalFeatures []MultiModalFeature`). Renames `TokenizedPrompt` to `TokenizedRequest` to distinguish the request-level container from the new per-prompt type.

**Before:**
```go
type TokenizedPrompt struct {
    PerPromptTokens    [][]uint32
    MultiModalFeatures []MultiModalFeature  // flat, wrong
    CacheSalt          string
}
```

**After:**
```go
type TokenizedRequest struct {
    Prompts   []PromptTokens
    CacheSalt string
}

type PromptTokens struct {
    TokenIDs           []uint32
    MultiModalFeatures []MultiModalFeature
}
```

The association between tokens and their multimodal features is now structural, not positional. Consumers no longer need bounds checks or guards like `if len(tp.PerPromptTokens) == 1` to pair features with the correct token array.

The `InferenceRequestBody` field is renamed from `TokenizedPrompt` to `TokenizedRequest` to match the type. The data-key identifiers (`TokenizedPromptDataKey`, `tokenizedPromptKeyID`) are intentionally unchanged since they are plugin registry keys, not type names.

Consumers updated:

- **blockkeys.go**: Iterates `tp.Prompts` directly; the `len(tp.PerPromptTokens) == 1` guard disappears because each `PromptTokens` carries its own features.
- **multimodal/producer.go** (`ExtractMMItems`): Iterates `tp.Prompts` accessing `p.MultiModalFeatures`.
- **disagg/multimodal_helpers.go** (`hasMultimodalContent`): Checks `len(p.MultiModalFeatures) > 0` per prompt.
- **approximateprefix/hashing.go**: Iterates `tp.Prompts` using `p.TokenIDs`.
- **vllmgrpc.go**: Constructs `TokenizedRequest` with `Prompts: []PromptTokens{{...}}` for pre-tokenized gRPC inputs.

## Why we need it

`PerPromptTokens` was `[][]uint32` (per-prompt) but `MultiModalFeatures` was `[]MultiModalFeature` (flat). MM feature offsets are only valid relative to one token array, but the flat shape forced consumers to assume which prompt the features belonged to. The `blockkeys.go` consumer had an explicit `len(tp.PerPromptTokens) == 1` guard to avoid misapplying features to multi-prompt requests.

Bundling tokens and features into `PromptTokens` eliminates this assumption and the guard, and makes the per-prompt association structural rather than requiring lockstep indexing of parallel slices.

### Known follow-up

`flattenMMFeatures` in `scorer/preciseprefixcache/legacy_producer.go` and `convertMMFeaturesToUpstream` in `tokenizer/tokenizer.go` are now identical (same input type, same output type, same logic). Deduplicating them is out of scope here since it crosses package boundaries and the legacy producer is deprecated.

Fixes #1584

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/interface/requesthandling/types.go
+++ b/pkg/epp/framework/interface/requesthandling/types.go
@@ -112,9 +112,9 @@ type InferenceRequestBody struct {
 	// If the payload is unmarshaled, we can perform advanced processing (like prefix cache aware routing).
 	// If it remains as raw bytes, such processing may not be supported.
 	Payload RequestPayload `json:"-"`
-	// TokenizedPrompt contains parser-derived tokenization results when available.
+	// TokenizedRequest contains parser-derived tokenization results when available.
 	// It is nil when the request was not already tokenized.
-	TokenizedPrompt *TokenizedPrompt `json:"-"`
+	TokenizedRequest *TokenizedRequest `json:"-"`
 
 	// Stream indicates whether the request specifies a streaming response (e.g., via a stream field).
 	// This typically implies the model server's response will be streamed.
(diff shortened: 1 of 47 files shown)


albertoperdomo2:
cc: @ahg-g @liu-cong

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -113,11 +113,10 @@ type TokenizedPrompt struct {
...
 	PerPromptTokens [][]uint32
-	// MultiModalFeatures holds one entry per multimodal item in prompt order.
-	// Nil if the prompt contains no multimodal content. Offsets are relative
-	// to PerPromptTokens[0] (always single-prompt when multimodal content is
-	// present).
-	MultiModalFeatures []MultiModalFeature
+	// MultiModalFeatures holds multimodal items per prompt, indexed in
+	// lockstep with PerPromptTokens. Single-prompt requests use a length-1
+	// outer slice. Nil if the prompt contains no multimodal content.
+	MultiModalFeatures [][]MultiModalFeature
```
ahg-g: it is a bigger change, but creating a `PromptTokens` struct that encapsulates the tokens and the mm features is more idiomatic
```
type TokenizedPrompt struct {
... 
  PromptTokens []PromptToken
...
}

type  struct {
  Tokens []uint32
   MultiModalFeatures []MultiModalFeature
}
```

but we probably need better names, TokenizedPrompt and PromptToken are too close, perhaps ask Claude / Antigravity to suggest better names?

albertoperdomo2:
@ahg-g anything blocking us on this one?

pkg/epp/framework/plugins/requestcontrol/dataproducer/prefixhash/hashing.go
bongwoobak: Missed rename. tp.PerPromptTokens should be 

`range tp.Prompts { ... p.TokenIDs ... }`

bongwoobak:
burstprefix (not in this PR, pulled in by today's main-merge #1727) still uses the old TokenizedPrompt

albertoperdomo2:
@bongwoobak @ahg-g is there anything preventing us from merging this one?

ahg-g:
sorry for the late reply, this looks great! please rebase when you get a chance.

ahg-g (approved):

PR #1636, opened 2026-06-12
feat(flowcontrol): add request outcome counter metric

RishabhSaini:
## Summary
- Add `flow_control_requests_total` counter labeled by outcome, priority,
  and inference pool
- Enables direct monitoring of rejection/eviction rates without log parsing
- Incremented once per request in `EnqueueAndWait` after outcome is determined

## Test plan
- [x] Existing unit tests pass
- [ ] Verify metric appears in Prometheus after enabling flow control


LukeAVanDrie:
cc @gyliu513

pkg/epp/metrics/llm_d_router_metrics.go
```diff
@@ -328,6 +328,15 @@ var (
 		},
 		[]string{"inference_pool"},
 	)
+
+	llmdFlowControlRequestsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: LLMDRouterEndpointPickerSubsystem,
+			Name:      "flow_control_requests_total",
+			Help:      metricsutil.HelpMsgWithStability("Total number of requests processed by the Flow Control layer, labeled by outcome.", compbasemetrics.ALPHA),
```
LukeAVanDrie: Non-blocking nit: the Help says "labeled by outcome", but the metric is labeled by `outcome`, `priority`, and `inference_pool`. The sibling flow-control metrics in this file (`flow_control_queue_size`, `flow_control_queue_bytes`, `flow_control_pool_saturation`) don't enumerate labels in their Help at all, so I'd just drop the trailing clause for consistency:

```suggestion
			Help:      metricsutil.HelpMsgWithStability("Total number of requests processed by the Flow Control layer.", compbasemetrics.ALPHA),
```

Your call though; completing the list ("labeled by outcome, priority, and inference pool") works just as well.

LukeAVanDrie (dismissed):
/lgtm

Note for other reviewers:  this counter is *not* redundant with `flow_control_request_queue_duration_seconds_count`. That histogram is only recorded once an item is finalized (`item.finalizeInternal`), whereas this counter also captures the controller-level early-rejection path where no item is ever created. So it genuinely fills a coverage gap (in addition to being a more convenient to consume metric).

pkg/epp/metrics/llm_d_router_metrics.go
LukeAVanDrie: Oh, we should also probably update `pkg/epp/metrics/metrics_test.go`

RishabhSaini: agreed, updated

gyliu513 (commented):
please also update docs/metrics.md when new metric is added

LukeAVanDrie:
> please also update docs/metrics.md when new metric is added

And https://github.com/llm-d/llm-d/blob/main/docs/architecture/core/router/epp/flow-control.md#metrics--observability

@gyliu513 there's a lot of x-repo docs to keep synced. Observability is one common source of this. Is there a practical solution here, or do we just need to be vigilant?

gyliu513:
@LukeAVanDrie I think we need to have a single source of truth, llm-d can refer to llm-d-router repo, I think this will helps. Keeping the docs in multiple places will be difficult to sync. comments?

LukeAVanDrie:
> @LukeAVanDrie I think we need to have a single source of truth, llm-d can refer to llm-d-router repo, I think this will helps. Keeping the docs in multiple places will be difficult to sync. comments?

This is what we do for per-plugin config info (we link to llm-d-router READMEs. As long as we are consistent this seems reasonable to me. Maybe we can tack this to the observability proposal and then sweep the llm-d docs?

RishabhSaini:
The existing flow control metrics are also undocumented in metrics.md. Will follow up with a separate PR documenting all of them:
   - flow_control_request_queue_duration_seconds (histogram)
 - flow_control_dispatch_cycle_duration_seconds (histogram)
 - flow_control_request_enqueue_duration_seconds (histogram)
 - flow_control_queue_size (gauge)
  - flow_control_queue_bytes (gauge)
 - flow_control_pool_saturation (gauge)

gyliu513:
> > @LukeAVanDrie I think we need to have a single source of truth, llm-d can refer to llm-d-router repo, I think this will helps. Keeping the docs in multiple places will be difficult to sync. comments?
> 
> This is what we do for per-plugin config info (we link to llm-d-router READMEs. As long as we are consistent this seems reasonable to me. Maybe we can tack this to the observability proposal and then sweep the llm-d docs?

Yes, let me follow up

gyliu513:
@LukeAVanDrie updated https://github.com/llm-d/llm-d/pull/1685

ahg-g:
@gyliu513 @LukeAVanDrie we need to keep updating the metrics in llm-d/llm-d because it is the source of llm-d.ai; the main user-facing docs. We shouldn't expect users to dig into the individual repos for end user docs. The same thing goes for the grafana dashboards in my opinion, a user shouldn't need to care about more than llm-d/llmd when deploying the llm-d stack and looking for operational guidance.

The one thing we are linking from llm-d/llm-d to llm-d/llm-d-router is the individual plugin docs because it is very detailed.

ahg-g:
Thanks @RishabhSaini, pls rebase.

ahg-g (approved):

PR #1638, opened 2026-06-12
test(flowcontrol): harden and reorganize the flow-control benchmark suite

LukeAVanDrie:
**What type of PR is this?**

/kind test
/kind cleanup

Hardens and clarifies the flow-control benchmark suite (`pkg/epp/flowcontrol/benchmark`). Test-only; no production code changes.

- Re-scope the full-path benchmark (`FullPathStress` → `FullPath`) to what it uniquely measures: dispatch throughput and allocation overhead through the *real* producer → concurrency detector → controller path. Drops an off-layer producer leak-counter assertion (covered by the inflightload producer's own unit tests), reports `d/s` like the other benchmarks, and fixes a `b.Fatalf` called from inside `RunParallel` (invalid off the benchmark goroutine).
- Fix the mock `benchDetector` so the `PerformanceMatrix` L sweep is meaningful: it accumulates in-flight up to the limit and reports saturated once exceeded (real W>L backpressure), while self-healing the empty-queue idle grants the processor's 1 ms dispatch ticker would otherwise leak. Previously this latched and hung at L=1. Adds detector unit tests for saturation-at-limit and self-heal.
- Replace fixed bootstrap `time.Sleep`s with a warmup request that blocks until dispatch for deterministic readiness instead of arbitrary timing.
- Reorganize into `doc.go` (two-family overview) + `helpers_test.go` + `benchmark_test.go`, and dedup the shared registry/policy setup.
- Add a `bench-smoke` make target + CI step (`go test -bench=. -benchtime=1x`) so benchmark bodies execute on every PR; `go test` only compiles them otherwise, letting runtime breakage (panics, deadlocks) go undetected.

N/A -- expanding upon the work in #1543

```release-note
NONE
```

### DIFF
--- a/pkg/epp/flowcontrol/benchmark/detector_test.go
+++ b/pkg/epp/flowcontrol/benchmark/detector_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2026 The Kubernetes Authors.
+Copyright 2026 The llm-d Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/epp/flowcontrol/benchmark/detector_test.go
+++ b/pkg/epp/flowcontrol/benchmark/detector_test.go
@@ -1,0 +1,125 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package benchmark
+
+import (
+	"context"
+	"testing"
+)
+
+// TestBenchDetector_NoLeakOnNonDispatchingCycles verifies the detector self-heals empty-queue idle
+// grants: the processor's dispatchCycle evaluates Saturation on a 1ms ticker even when the queue is
+// empty, so those grants never dispatch and are never Released. The detector may report saturated
+// briefly but must never latch -- grants keep recurring and no saturated run exceeds the stuck
+// threshold.
+func TestBenchDetector_NoLeakOnNonDispatchingCycles(t *testing.T) {
+	d := &benchDetector{}
+	d.concurrencyLimit.Store(1)
+
+	grants, run, maxRun := 0, 0, 0
+	for i := 0; i < 100; i++ {
+		if d.Saturation(context.Background(), nil) < 1.0 {
+			grants++
+			run = 0
+		} else {
+			run++
+			if run > maxRun {
+				maxRun = run
+			}
+		}
+	}
+	if grants == 0 {
+		t.Fatal("detector latched saturated across all idle cycles (permanent leak)")
+	}
+	if maxRun > stuckReadThreshold {
+		t.Fatalf("idle saturated run = %d, want <= %d (self-heal should reclaim)", maxRun, stuckReadThreshold)
+	}
+}
+
+// TestBenchDetector_SaturatesAtLimit asserts the W>L backpressure the PerformanceMatrix sweep exists
+// to measure: with the limit reached by in-flight (un-Released) grants, the next Saturation must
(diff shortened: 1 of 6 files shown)


LukeAVanDrie:
cc: @RishabhSaini 

doing some re-organizing as our bench setup is getting complex

LukeAVanDrie:
Note to reviewers: This is a very low-priority PR in a low-churn path. We can wait a bit before merging this

pkg/epp/flowcontrol/benchmark/doc.go
```diff
@@ -0,0 +1,66 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
ahg-g: llm-d authors

pkg/epp/flowcontrol/benchmark/detector_test.go
```diff
@@ -0,0 +1,125 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
shmuelk: llm-d Authors

pkg/epp/flowcontrol/benchmark/helpers_test.go
```diff
@@ -120,37 +85,84 @@ type testDetector interface {
...
+// The complication: the processor also runs dispatchCycle on a 1ms ticker when the queue is EMPTY
+// (notably at startup). Those idle grants never dispatch and so are never Released; a naive detector
+// would latch saturated forever and deadlock every request until its TTL. We distinguish leaked
+// idle grants from genuine load saturation with releaseCount: while releases are
+// flowing the saturation is real; if two consecutive saturated reads see zero releases between them
+// the outstanding grants are leaked idle permits, so we reclaim one. The 1.0/0.99 split sits below
+// usagelimits.DefaultPolicy's 1.0 ceiling, so a grant never trips head-of-line gating.
+//
+// stuckReads and lastReleaseCount are touched only by the single-threaded dispatch loop (Saturation)
+// and so need no synchronization; inFlight and releaseCount are atomic (Release runs on workers).
```
shmuelk, line 106: +1  Currently the benchmark fails to run precisely due to the above complication.

My question is, what happens in the benchmark when there is true queueing, even if momentary queueing? Shouldn't there be true saturation (i.e. Inflight > level) ?

ahg-g:
@LukeAVanDrie would you like to address the comments and move this one forward?

elevran:
@LukeAVanDrie is this still relevant or can be closed?

LukeAVanDrie:
> @LukeAVanDrie is this still relevant or can be closed?

yes, but low priority and fell through the cracks

this is just for maintainability

will rebase today, resolve conflicts and address feedback; this isn't critical for the release though

pkg/epp/flowcontrol/benchmark/doc.go
LukeAVanDrie: Done; switched to `Copyright 2026 The llm-d Authors.`

pkg/epp/flowcontrol/benchmark/detector_test.go
LukeAVanDrie: Done; switched to `Copyright 2026 The llm-d Authors.`

pkg/epp/flowcontrol/benchmark/helpers_test.go
LukeAVanDrie: Yes, under true queueing the detector does report true saturation: once in-flight (un-`Release`d) grants reach L, every subsequent `Saturation` read returns 1.0 (covered by `TestBenchDetector_SaturatesAtLimit`). The reclaim path never fires while that saturation is genuine, because genuine load means dispatched workers are calling `Release`, and any release observed between two saturated reads resets the stuck counter and keeps returning 1.0.

The reclaim only triggers on consecutive saturated reads with zero releases in between which the signifies the empty-queue idle-tick leak (a grant that never led to a dispatch, so no `Release` can ever arrive for it). The one edge case is a genuine holder whose worker is descheduled long enough to stall its `Release` past the threshold; then we reclaim spuriously and transiently admit past L. For a CPU-cost benchmark, briefly loosening L just shifts the coordinate's operating point, while the opposite failure mode (latching saturated) deadlocks the run until request TTL, which is exactly the L=1 hang this PR fixes. 

I've expanded the comment on `stuckReadThreshold` in the latest commit to spell this out.

LukeAVanDrie:
> @LukeAVanDrie is this still relevant or can be closed?

rebased and feedback addressed

ahg-g (approved):

Issue #1287, opened 2026-05-24
Long context support and evaluation


ahg-g:
**What would you like to be added**:

Long context support and evaluation, ensuring nothing breaks for 100k+ context 

**Why is this needed**:
Agentic workloads have a ~100k+ ISL and ~500 OSL, we need to evaluate that the router supports this type of workload at high concurrency without breaking.

jtechapps:
I tested the Optimized Baseline configuration with (tracing disabled, verbosity=1) with different `maxPrefixBlocksToMatch` sizes and request rates below. I am also running in gateway mode. I am using 50 llm-d-simulator pods as my backend. I am also requesting 100 CPUs and 500 GiB of memory and using c3 machine type for running EPP.

## Results:

I tested 50 and 100 request rates for 25k and 100k input sizes. I found that the default resource requests (4 CPU, 8 GiB memory) were too low for EPP so I gave it 100/500GiB and recorded the usage. The maxPrefixBlocksToMatch had impact on CPU usage and small impact on memory. Memory usage remained low except for longer decode requests (lots of inflight requests).

The errors that were seen were not from EPP failing but from model servers being overloaded/restarting.

| input_type | output_len | maxPrefixBlocksToMatch | requested_rate | peak_cpu_cores | peak_mem_gib | error_rate_percent |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| 100k | 500 | 256 | 50 | 15.13 | 2.27 | 0.00% |
| 100k | 500 | 512 | 50 | 15.97 | 2.39 | 0.00% |
| 100k | 500 | 1024 | 50 | 16.96 | 2.87 | 0.00% |
| 100k | 500 | 2048 | 50 | 17.14 | 3.76 | 1.87% |
| 100k | 1000 | 256 | 50 | 17.51 | 3.66 | 0.00% |
| 100k | 1000 | 512 | 50 | 18.00 | 3.88 | 0.00% |
| 100k | 1000 | 1024 | 50 | 19.40 | 4.26 | 0.00% |
| 100k | 1000 | 2048 | 50 | 20.28 | 5.23 | 2.00% |
| 100k | 5000 | 256 | 50 | 29.44 | 12.33 | 0.00% |
| 100k | 5000 | 512 | 50 | 28.88 | 12.35 | 0.00% |
| 100k | 5000 | 1024 | 50 | 30.95 | 12.54 | 0.37% |
[...]

PR #1641, opened 2026-06-13
Bump default epp cpu 4->8

liu-cong:
This change is triggered by:
1. Increase of the maxPrefixTokens in approximate prefix producer leading to potentially more compute cost for long inputs.
2. Recent refactor(multi-modal support, aligning on token format) added some overhead for better code readability
3. Based on test result in https://github.com/llm-d/llm-d-router/issues/1287#issuecomment-4666058475, this config can support ISL/OSL 100k/1k at 25 QPS with 25 model servers.

**What type of PR is this?**

Fixes #

```release-note
Default EPP CPU request is increased from 4 to 8 to accommodate longer prefix matching and other enhancements. 
```

### DIFF
--- a/config/charts/routerlib/values.yaml
+++ b/config/charts/routerlib/values.yaml
@@ -57,7 +57,7 @@ epp:
   # EPP container resources: CPU limits unset for burst capacity, memory capped at 16Gi
   resources:
     requests:
-      cpu: "4"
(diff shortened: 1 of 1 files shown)


ahg-g:
Addressed by https://github.com/llm-d/llm-d-router/pull/1642

PR #1644, opened 2026-06-13
Refactor request metrics

### LINKED ISSUE #1646: Bring consistency to request metrics
The aggregated request metrics names, labels and message description are inconsistent, we should bring consistency:

1) All metrics are prefixed with `request`
2) All metrics have the labels model, target model, fairness id and priority
3) Latency metrics either use abbreviations and full name (tpot, itl and ttft); the latter should be fine.

Note that the metrics in https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/metrics/llm_d_router_metrics.go have not been released yet, so it is safe to change them.

ahg-g:
**What type of PR is this?**
/kind feature

Refactors all aggregate request metrics in two ways:
1) all consistently prefixed with request_
2) all have model, target_model, priority and fairness id labels
3) removing duplicate metrics emitted by the predicted latency plugin (ttft and tpot)

Fixes https://github.com/llm-d/llm-d-router/issues/1646

```release-note
NONE
```


ahg-g:
@gyliu513 pls take a look. The aggregated request metrics names, labels and message description were inconsistent, this PR aims to bring consistency. Note that the metrics this PR adjusts have not been released yet, so it is safe to change them.

Once this merges, I will send a follow up PR to update https://github.com/llm-d/llm-d/blob/main/docs/resources/observability/metrics.md

pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/metrics.go
```diff
@@ -92,16 +92,6 @@ var (
 		modelLabels,
 	)
 
-	llmdRequestTTFT = prometheus.NewHistogramVec(
```
ahg-g, line 95: @kaushikmitr fyi; I remove the ttft and tpot metrics emitted by the predicted latency producer since they are reported by epp core.

gyliu513:
Thanks @ahg-g , looks good to me, I think IPP may also need to follow same pattern, let me check IPP as well, thanks!

pkg/epp/handlers/server.go
```diff
@@ -310,7 +326,8 @@ func (s *StreamingServer) Process(srv extProcPb.ExternalProcessor_ProcessServer)
 			}
 		}
 		if reqCtx.RequestRunning {
-			metrics.DecRunningRequests(reqCtx.IncomingModelName)
+			fairnessID, priority := extractFairnessAndPriority(reqCtx)
```
vMaroon:
```suggestion
```

Redundant call since L314 always executes before on the same `reqCtx`.

liu-cong (approved):

PR #1645, opened 2026-06-13
fix: don't abort kind-dev-env.sh on macOS when PROM_ENABLED=true

sudoalok:
## What this does

Stops `PROM_ENABLED=true make env-dev-kind` from aborting on macOS.

The Prometheus pre-flight check in `scripts/kind-dev-env.sh` reads
`/proc/sys/fs/inotify/max_user_instances`, a Linux-only path. On macOS that
read fails and, under `set -eo pipefail`, the failed `cat` kills the whole
script before the cluster is ever created.

This falls back to a passing value when the path is absent, so the host-side
check degrades gracefully off-Linux. On Linux the behavior is unchanged.

```diff
 if [ "${PROM_ENABLED}" == "true" ]; then
-  INOTIFY_INSTANCES=$(cat /proc/sys/fs/inotify/max_user_instances)
+  INOTIFY_INSTANCES=$(cat /proc/sys/fs/inotify/max_user_instances 2>/dev/null || echo 512)

### DIFF
--- a/scripts/kind-dev-env.sh
+++ b/scripts/kind-dev-env.sh
@@ -200,7 +200,10 @@ done
 
 # Prometheus config-reloader needs sufficient inotify resources
 if [ "${PROM_ENABLED}" == "true" ]; then
-  INOTIFY_INSTANCES=$(cat /proc/sys/fs/inotify/max_user_instances)
+  # On non-Linux hosts (e.g. macOS) /proc does not exist and the kind pods run in
+  # the container-runtime VM, not the host, so fall back to a passing value rather
+  # than letting the failed read abort the script under `set -e`.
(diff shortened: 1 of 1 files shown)


shmuelk (approved):

elevran:
@sudoalok please fix the conflict

sudoalok:
thanks @elevran - looks like this was superseded. [#1654](https://github.com/sudoalok/llm-d/issues/1654) already merged the same macOS fix on main (guards the /proc read with -r plus a node-side check for the VM case), which covers what this PR did and more. nothing left for this one to add, so closing as superseded by [#1654](https://github.com/sudoalok/llm-d/issues/1654)

elevran:
closing as superseded by [#1654](https://github.com/sudoalok/llm-d/issues/1654)

PR #1649, opened 2026-06-14
docs: update precise-prefix-cache plugin and default plugins

zdtsw:
**What type of PR is this?**
/kind documentation
/kind deprecation

- use precise-prefix-cache-producer instead of deprecated precise-prefix-cache-scorer in example
- document default plugins
- rename epp-precise-prefix-cache-split-config.yaml to epp-precise-prefix-cache-config.yam

Fixes #

```release-note
NONE
```


vMaroon (approved):
clean

PR #1653, opened 2026-06-15
Session affinity filter/scorer can optionally pick the scheduling profile to inject the routed endpoint from. This enables P/D disaggregation support.

liu-cong:
**What type of PR is this?**
/kind feature

Fixes #

```release-note
Session affinity filter/scorer can optionally pick the scheduling profile to inject the routed endpoint from. This enables P/D disaggregation support.
```

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/session_affinity.go
+++ b/pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/session_affinity.go
@@ -107,13 +107,6 @@ func (s *SessionAffinity) Filter(ctx context.Context, request *scheduling.Infere
 
 // ResponseHeader sets the session header on the response sent to the client.
 func (s *SessionAffinity) ResponseHeader(ctx context.Context, request *scheduling.InferenceRequest, response *requestcontrol.Response, targetPod *datalayer.EndpointMetadata) {
-	podToWrite := targetPod
-	if s.profileName != "" && request != nil && request.SchedulingResult != nil {
-		if result := request.SchedulingResult.ProfileResults[s.profileName]; result != nil && len(result.TargetEndpoints) > 0 && result.TargetEndpoints[0] != nil {
-			if md := result.TargetEndpoints[0].GetMetadata(); md != nil {
-				podToWrite = md
-			}
-		}
-	}
+	podToWrite := sessionutil.ResolvePodToWrite(request, s.profileName, targetPod)
 	sessionutil.WriteResponseHeader(ctx, SessionAffinityType, s.sessionHeader, response, podToWrite)
 }
--- a/pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/session_affinity.go
+++ b/pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/session_affinity.go
@@ -38,7 +38,8 @@ type parameters struct {
 	// HeaderName overrides the default x-session-token header used to read and
 	// write the session token. When empty the default is used.
 	HeaderName string `json:"headerName"`
-	// ProfileName is the name of the profile this instance is associated with.
+	// ProfileName is the name of the profile this instance is associated with (optional).
+	// When empty, the plugin defaults to the primary (decode) pod.
 	// Used in ResponseHeader to look up the correct target pod from SchedulingResult.
 	ProfileName string `json:"profileName"`
 }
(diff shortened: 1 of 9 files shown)


liu-cong:
cc @roytman

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/session_affinity.go
```diff
@@ -99,6 +106,12 @@ func (s *SessionAffinity) Filter(ctx context.Context, request *scheduling.Infere
 }
 
 // ResponseHeader sets the session header on the response sent to the client.
-func (s *SessionAffinity) ResponseHeader(ctx context.Context, _ *scheduling.InferenceRequest, response *requestcontrol.Response, targetPod *datalayer.EndpointMetadata) {
-	sessionutil.WriteResponseHeader(ctx, SessionAffinityType, s.sessionHeader, response, targetPod)
+func (s *SessionAffinity) ResponseHeader(ctx context.Context, request *scheduling.InferenceRequest, response *requestcontrol.Response, targetPod *datalayer.EndpointMetadata) {
```
roytman, line 109: identical resolution logic in filter and scorer. This implementation can be moved to pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/session_affinity.go

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/session_affinity.go
```diff
@@ -99,6 +106,12 @@ func (s *SessionAffinity) Filter(ctx context.Context, request *scheduling.Infere
...
+	podToWrite := targetPod
```
roytman: profile-scoped instance writes the wrong pod on decode-only requests
`ResponseHeader` initializes `podToWrite` to the decode `targetPod` and only overwrites it when the named profile is present in `SchedulingResult`

The decision to run prefill is dynamic, per request. When the decider rejects prefill, the handler explicitly sets `profileResults[h.prefillProfile] = nil`. Then `ProcessResults` filters that back out [disagg_profile_handler.go](https://github.com/llm-d/llm-d-router/blob/745bae5e8ca50da7c9adcc207ea7d20dbd54700b/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go#L360
```go
if prefillRes, ok := profileResults[h.prefillProfile]; ok && prefillRes != nil {
    updatedResults[h.prefillProfile] = prefillRes
}
```
The key exists in the working map, but the value is `nil,` so `prefillRes != nil` is `false` and the prefill profile is not carried into the final SchedulingResult.ProfileResults. 

So the PR's guard fails and the prefill instance's ResponseHeader writes the `decode` pod (the global TargetPod) into `x-session-token-prefill`. Next request, the prefill Filter decodes a decode-pod name, finds no match among prefill candidates, and silently falls through to no pinning.

The PR's profileName is generic, so an encode instance (profileName: encode) hits the identical failure when encode is skipped. The fix must be profile-agnostic, which is another reason to centralize it. 
Suggested fix (for `plugins/scheduling/filter/sessionaffinity/session_affinity.go`)

```goLang
podToWrite := targetPod
if s.profileName != "" {
    // profile-scoped: never fall back to the decode pod; the profile may have been skipped.
    podToWrite = nil
    if request != nil && request.SchedulingResult != nil {
        if result, ok := request.SchedulingResult.ProfileResults[s.profileName]; ok && len(result.TargetEndpoints) > 0 {
            podToWrite = result.TargetEndpoints[0].GetMetadata()
        }
    }
}
sessionutil.WriteResponseHeader(ctx, SessionAffinityType, s.sessionHeader, response, podToWrite)
```

roytman (commented):
There are missing test cases:
Both unit test files add a single prefill profile lookup case. The new branches that are untested are exactly where the bug lives:
- profileName set, profile absent from ProfileResults (the decode-only case)
- profileName set, TargetEndpoints empty
- profileName set, request/SchedulingResult nil

It's better to extend the integration [test](test/integration/epp/session_affinity_test.go), which can validate the cross-request PD flow to prove the feature works and guard against decode-only regressions.


Even with the existing session affinity filter and scorer, users can still create a configuration that includes both; while harmless, this is misleading. Furthermore, adding prefill filters and scorers allows for genuinely invalid configurations where one could overwrite the other. 
Should we investigate the possibility of validating cross-plugin configurations to prevent these kinds of user mistakes?

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/session_affinity.go
liu-cong: great catch! I refactored it so that when profileName is specified, it only returns the target pod associated with that profile (and if not set, return nil).

vMaroon (approved):

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/session_affinity.go
vMaroon: nit: exercise decode-only path

liu-cong: also added decode-only path in the integ test

PR #1657, opened 2026-06-15
Tag container images with commit SHA

pierDipi:
**What type of PR is this?**

/kind feature

Dev images are currently tagged only with the branch name, which is mutable: every push to main overwrites the previous image. This makes it impossible to pin a deployment or test run to the exact code that produced the image.

Add the commit SHA as an additional image tag so each build is individually addressable and immutable.

```release-note
Add the commit SHA as an additional image tag so each build is individually addressable and immutable.
```


shmuelk:
@pierDipi Please update the PR description to add a release note. Users should know about these extra tags.

pierDipi:
> @pierDipi Please update the PR description to add a release note. Users should know about these extra tags.

Done

liu-cong (approved):

PR #1658, opened 2026-06-15
Add EPP request/response processing latency metrics

### LINKED ISSUE #1553: Add critical EPP latency metrics
Add the following metrics:

* [x] `llm_d_router_epp_plugin_duration_seconds` Add this to **all** plugins. Currently this metric is only added to scheduler plugins.
* [ ] `llm_d_router_epp_request_processing_duration_seconds` The time it takes from EPP receives the request to it picking an endpoint. This is the "EPP Overhead" for request handling.
* [ ] `llm_d_router_epp_response_processing_duration_seconds` The time it takes from EPP receives the response to it sending response back to the proxy. This is the "EPP Overhead" for response handling.

**Why is this needed**:
Helps understand EPP overhead and pin point performance issues at e2e level, and per plugin level.

Reference: existing metrics
https://docs.google.com/document/d/18iu0_clvv1O_nPry1x7oLKmtZyaldfGvvBGI3L_fgbY/edit?tab=t.0

satyamg1620:
**What type of PR is this?**
/kind feature

Adds two EPP latency metrics from #1553 to expose EPP processing overhead:

`llm_d_epp_request_processing_duration_seconds` — from request receipt until the request body has been handled; includes admission control.
`llm_d_epp_response_processing_duration_seconds` — per-chunk handler sum for streaming, header-to-completion span for non-streaming; excludes model-server generation time.

Fixes #1553

```release-note
Add llm_d_epp_request_processing_duration_seconds and llm_d_epp_response_processing_duration_seconds metrics measuring EPP request and response processing overhead.
```

### DIFF
--- a/pkg/epp/handlers/server.go
+++ b/pkg/epp/handlers/server.go
@@ -136,6 +136,10 @@ type RequestContext struct {
 	RequestDroppedReason errcommon.RequestDroppedReason
 	modelServerStreaming bool
 
+	// responseProcessingDuration accumulates the time EPP spends in its response
+	// handlers across all response events, excluding model server generation time.
+	responseProcessingDuration time.Duration
+
 	Response *Response
 
 	reqHeaderResp  *extProcPb.ProcessingResponse
@@ -465,6 +469,7 @@ func (s *StreamingServer) Process(srv extProcPb.ExternalProcessor_ProcessServer)
 		case *extProcPb.ProcessingRequest_RequestTrailers:
 			// This is currently unused.
 		case *extProcPb.ProcessingRequest_ResponseHeaders:
+			respHeaderStart := time.Now()
 			for _, header := range v.ResponseHeaders.Headers.GetHeaders() {
 				value := string(header.RawValue)
 				loggerTrace.Info("header", "key", header.Key, "value", value)
@@ -478,12 +483,14 @@ func (s *StreamingServer) Process(srv extProcPb.ExternalProcessor_ProcessServer)
 			reqCtx.RequestState = ResponseReceived
 			reqCtx = s.HandleResponseHeaders(ctx, reqCtx, v)
 			reqCtx.respHeaderResp = s.generateResponseHeaderResponse(reqCtx)
+			reqCtx.responseProcessingDuration += time.Since(respHeaderStart)
 
 		case *extProcPb.ProcessingRequest_ResponseBody:
 			endOfStream := v.ResponseBody.EndOfStream
 			chunk := v.ResponseBody.Body
 
 			if reqCtx.modelServerStreaming {
+				respBodyStart := time.Now()
 				if endOfStream {
 					reqCtx.ResponseComplete = true
 					reqCtx.ResponseCompleteTimestamp = time.Now()
@@ -493,6 +500,10 @@ func (s *StreamingServer) Process(srv extProcPb.ExternalProcessor_ProcessServer)
 				chunk = rewriteModelName(chunk, reqCtx.TargetModelName, reqCtx.IncomingModelName)
 				// For streaming response, we send response chunk back to envoy every time we received it.
 				reqCtx.respBodyResp = generateResponseBodyResponses(chunk, endOfStream, reqCtx.Response.DynamicMetadata)
+				reqCtx.responseProcessingDuration += time.Since(respBodyStart)
+				if endOfStream {
+					metrics.RecordResponseProcessingLatency(reqCtx.responseProcessingDuration)
+				}
 			} else {
 				respBody = append(respBody, chunk...)
 				if endOfStream {
@@ -553,6 +564,7 @@ func (s *StreamingServer) finishResponse(ctx context.Context, reqCtx *RequestCon
 		return
 	}
 
+	start := time.Now()
 	reqCtx.ResponseComplete = true
 	reqCtx.ResponseCompleteTimestamp = time.Now()
 	reqCtx = s.HandleResponseBody(ctx, reqCtx, body, true)
@@ -562,6 +574,8 @@ func (s *StreamingServer) finishResponse(ctx context.Context, reqCtx *RequestCon
 		// For non-streaming response, we send response back to envoy after receiving all the response body.
 		reqCtx.respBodyResp = generateResponseBodyResponses(body, setEos, reqCtx.Response.DynamicMetadata)
 	}
+	reqCtx.responseProcessingDuration += time.Since(start)
+	metrics.RecordResponseProcessingLatency(reqCtx.responseProcessingDuration)
 }
 
 // rewriteModelName replaces occurrences of the target (internal) model name with the
--- a/pkg/epp/handlers/server.go
+++ b/pkg/epp/handlers/server.go
@@ -136,9 +136,13 @@ type RequestContext struct {
 	RequestDroppedReason errcommon.RequestDroppedReason
 	modelServerStreaming bool
 
-	// responseProcessingDuration accumulates the time EPP spends in its response
-	// handlers across all response events, excluding model server generation time.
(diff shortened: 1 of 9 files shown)


satyamg1620:
@liu-cong can you please review this PR

liu-cong:
> @liu-cong can you please review this PR

Sorry for the delay. I am focusing on the release right now. Will try to get to this later this week if everything goes well on the release.

satyamg1620:
> Sorry for the delay. I am focusing on the release right now. Will try to get to this later this week if everything goes well on the release.

Sure. Thanks @liu-cong

pkg/epp/metrics/metrics_test.go
```diff
@@ -845,6 +845,60 @@ func TestSchedulerE2ELatency(t *testing.T) {
...
+	for _, duration := range durations {
+		RecordRequestProcessingLatency(duration)
+	}
+
+	want, err := os.Open("testdata/llm_d_request_processing_duration_seconds_metric")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer want.Close()
+	if err := promtestutil.GatherAndCompare(metrics.Registry, want, "llm_d_router_epp_request_processing_duration_seconds"); err != nil {
```
elevran: This assertion looks for the wrong metric name. `LLMDRouterEndpointPickerSubsystem` is `"llm_d_epp"` (line 28 of `llm_d_router_metrics.go`), so the histogram is emitted as `llm_d_epp_request_processing_duration_seconds`, matching the rest of the package (`llm_d_epp_request_total`, `llm_d_epp_request_error_total`, etc.). Drop the `_router_` segment here and in the testdata filename + fixture `# HELP`/`# TYPE`/metric lines.

```suggestion
		if err := promtestutil.GatherAndCompare(metrics.Registry, want, "llm_d_epp_request_processing_duration_seconds"); err != nil {
```

pkg/epp/metrics/metrics_test.go
```diff
@@ -845,6 +845,60 @@ func TestSchedulerE2ELatency(t *testing.T) {
...
+	for _, duration := range durations {
+		RecordResponseProcessingLatency(duration)
+	}
+
+	want, err := os.Open("testdata/llm_d_response_processing_duration_seconds_metric")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer want.Close()
+	if err := promtestutil.GatherAndCompare(metrics.Registry, want, "llm_d_router_epp_response_processing_duration_seconds"); err != nil {
```
elevran: Same off-by-one as the previous comment: drop `_router_` so the assertion matches the emitted name `llm_d_epp_response_processing_duration_seconds`. Also rename the testdata file and update its contents.

```suggestion
		if err := promtestutil.GatherAndCompare(metrics.Registry, want, "llm_d_epp_response_processing_duration_seconds"); err != nil {
```

elevran (commented):
Two CI failures on this PR. The `test` job fails because the new unit tests assert the wrong subsystem constant. 

The PR title and the release-note block both also advertise the wrong metric name and must be fixed.

Release-note should read:
```
Add `llm_d_epp_request_processing_duration_seconds` and `llm_d_epp_response_processing_duration_seconds` metrics measuring EPP request and response processing overhead.
```

Can be done here or in a follow up PR (please open a tracking issue if deferring):
- `docs/metrics.md` does not list the new metrics. Since the PR description treats them as user-facing, a one-line entry under the existing `llm_d_epp_*` section would keep the catalog accurate.
- The e2e-gaie preset in `test/e2e/epp/utils_test.go:260-289` does not assert these metrics either, so they are not exercised by the e2e-gaie suite. Worth a follow-up to add them to the preset.

elevran:
Please be sure to fix the all relevant fields of the new metrics (names, Help text, unit tests, and the testdata fixtures) - they all use the wrong prefix,

pkg/epp/handlers/server.go
```diff
@@ -553,6 +572,7 @@ func (s *StreamingServer) finishResponse(ctx context.Context, reqCtx *RequestCon
 		return
 	}
 
+	start := time.Now()
```
liu-cong, line 591: the start time should be when response headers are received

pkg/epp/handlers/server.go
```diff
@@ -478,12 +494,14 @@ func (s *StreamingServer) Process(srv extProcPb.ExternalProcessor_ProcessServer)
...
 			reqCtx = s.HandleResponseHeaders(ctx, reqCtx, v)
 			reqCtx.respHeaderResp = s.generateResponseHeaderResponse(reqCtx)
+			reqCtx.responseProcessingDuration += time.Since(respHeaderStart)
 
 		case *extProcPb.ProcessingRequest_ResponseBody:
 			endOfStream := v.ResponseBody.EndOfStream
 			chunk := v.ResponseBody.Body
 
 			if reqCtx.modelServerStreaming {
+				respBodyStart := time.Now()
```
liu-cong, line 519: The start should strictly be measured when headers are received because body should come after.

pkg/epp/handlers/server.go
```diff
@@ -478,12 +494,14 @@ func (s *StreamingServer) Process(srv extProcPb.ExternalProcessor_ProcessServer)
 			reqCtx.RequestState = ResponseReceived
 			reqCtx = s.HandleResponseHeaders(ctx, reqCtx, v)
 			reqCtx.respHeaderResp = s.generateResponseHeaderResponse(reqCtx)
+			reqCtx.responseProcessingDuration += time.Since(respHeaderStart)
```
liu-cong: the endtime should be after `finishResponse`

pkg/epp/requestcontrol/director.go
liu-cong, line 1: The `request_processing_duration_seconds` should also be implemented in server.go, start from request header is received (we already record `reqCtx.RequestReceivedTimestamp`, and end after request body is handled, likely [here](https://github.com/llm-d/llm-d-router/blob/4ed07d6de9c6aec8c8a2c9fcd510b81177dba042/pkg/epp/handlers/server.go#L463)

satyamg1620:
Thanks @liu-cong. Two questions before I rewrite this.

**Request metric**. Happy to move it to server.go, RequestReceivedTimestamp through the end of the RequestBody case. But that window swallows director.HandleRequest, so it swallows admission control too. Right now I subtract that ([director.go#L304-L306](https://github.com/llm-d/llm-d-router/blob/f9bf0a8a97353e226e2581f8abd8ea63f3349e0d/pkg/epp/requestcontrol/director.go#L304-L306)) because under flow control Admit is EnqueueAndWait and the queue wait would dominate the histogram. Do you want it still excluded after the move, or just the raw span?

If it stays excluded, I'd like to fix what Copilot caught while I'm in there. I subtract the whole Admit call, but LegacyAdmissionController.Admit doesn't queue at all, it just runs saturation detection. Flow control is off by default, so today we subtract real EPP work out of an EPP-overhead metric. I think the cleanest fix is Admit returning its own queue wait (zero for legacy). Can you please suggest

**Response metric**. On #1553 you said accumulated sum makes sense for streaming, but the inline comments here describe a single elapsed span. I read that as: non-streaming becomes one span headers-to-finishResponse, streaming keeps the sum, since an elapsed span there would just re-measure generation time. Right?

liu-cong:
> the queue wait would dominate the histogram

This is expected by the definition of this metric. The queue wait time only applies when queuing happens.

>  I read that as: non-streaming becomes one span headers-to-finishResponse, streaming keeps the sum, since an elapsed span there would just re-measure generation time. 

Yeah for streaming it's a bit complex, we will need to measure the sum of per-chunk EPP response processing latency.

satyamg1620:
@liu-cong @elevran can you please review this PR

liu-cong (approved):

Issue #1485, opened 2026-06-04
[Observability] P/D sidecar observability: add Prometheus /metrics, wire trace propagation, document enablement


gyliu513:
**What would you like to be added**

The P/D sidecar (`cmd/pd-sidecar`), deployed next to decode pods on the data path, is an observability blind spot. Three gaps:

- **1. Prometheus metrics**
  - EPP already exposes `/metrics` (default port 9090). The sidecar is a separate process and has **no scrape endpoint** — `cmd/pd-sidecar/main.go` starts no metrics server and the sidecar code imports no Prometheus client. P/D timing exists **only** as OTel span attributes (`llm_d.pd_proxy.true_ttft_ms` etc., `pkg/sidecar/proxy/connector_sglang.go:172`, `connector_nixlv2.go:281`), not as Prometheus series.
  - Add `/metrics` with low-cardinality metrics: request / disagg counts, prefill & decode latency histograms, error counters.

- **2. Trace propagation**
  - The sidecar already emits spans (`llm_d.pd_proxy.request|prefill|decode`) and extracts inbound context via otelhttp. But outbound prefill/decode calls go through a plain `http.Transport` cloned from the default and never inject W3C trace context, there is no `otelhttp.NewTransport` wrapping and no `propagation.Inject` anywhere in the sidecar. As a result vLLM's `llm_request` spans attach as unrelated traces in Jaeger instead of under the same trace.
  - Wrap the outbound transport with `otelhttp.NewTransport` (or inject the propagator explicitly) so EPP → routing-proxy → vLLM share one trace.

- **3. Enablement & docs**
  - The `OTEL_*` env vars in the chart (`config/charts/routerlib/templates/_deployment.yaml`) apply to the **router/EPP** container, gated by `.Values.router.tracing.enabled`. The pd-sidecar deployment is not in this repo's charts, so **sidecar-side tracing enablement is undocumented**. `docs/metrics.md` does not mention the sidecar at all, and there is no observability capability matrix.
  - Document `OTEL_*` for the sidecar and add a capability matrix recording current state (spans exist) vs. gaps (metrics + propagation + sane defaults).

**Why is this needed**

P/D is a production blind spot. Operators can scrape EPP but not the routing proxy on the data path, and traces from EPP, routing-proxy, and vLLM appear as separate services unless propagation is wired end-to-end.

@nicolexin ^^

gyliu513:
[...]

PR #1677, opened 2026-06-17
feat(sidecar): expose Prometheus /metrics on the P/D sidecar

gyliu513:
**What type of PR is this?**

/kind feature

The P/D sidecar (`cmd/pd-sidecar`) is an observability blind spot: it starts no metrics server and imports no Prometheus client. P/D timing exists only as OTel span attributes (`llm_d.pd_proxy.*`), not as scrapeable Prometheus series, so operators can scrape EPP but not the routing proxy on the data path.

This adds a Prometheus `/metrics` endpoint to the sidecar with low-cardinality metrics under the `llm_d_disagg_sidecar` subsystem:

- `llm_d_disagg_sidecar_requests_total{api_type}` - all requests handled
- `llm_d_disagg_sidecar_disagg_requests_total{disagg_type}` - requests routed through disaggregation
- `llm_d_disagg_sidecar_encode_duration_seconds` - encode stage latency histogram
- `llm_d_disagg_sidecar_prefill_duration_seconds` - prefill stage latency histogram
- `llm_d_disagg_sidecar_decode_duration_seconds` - decode stage latency histogram
- `llm_d_disagg_sidecar_request_errors_total{stage}` - encode/prefill/decode stage errors

Addresses gap 1 (Prometheus metrics) of #1485.

```release-note
Add a Prometheus /metrics endpoint to the P/D sidecar on a dedicated --metrics-port (0, the default, disables it) exposing request, disaggregation, encode/prefill/decode latency, and error metrics under the llm_d_disagg_sidecar subsystem
```


shmuelk:
@gyliu513 As the sidecar isn't just for P/D I think the metrics should be llm_d_disagg_sidecar.....

pkg/sidecar/metrics/metrics.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	compbasemetrics "k8s.io/component-base/metrics"
+	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	metricsutil "github.com/llm-d/llm-d-router/pkg/common/observability/metrics"
+)
+
+// subsystem is the Prometheus subsystem for sidecar metrics, aligned with the
+// llm_d.pd_proxy.* span namespace.
+const subsystem = "llm_d_pd_proxy"
```
shmuelk:
```suggestion
const subsystem = "llm_d_disagg_sidecar"
```

pkg/sidecar/metrics/metrics.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Stage labels for prefill/decode error attribution.
+const (
+	StagePrefill = "prefill"
+	StageDecode  = "decode"
+)
```
shmuelk: What about encode?

pkg/sidecar/metrics/metrics.go
```diff
@@ -0,0 +1,136 @@
...
+
+	decodeDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: subsystem,
+			Name:      "decode_duration_seconds",
+			Help:      metricsutil.HelpMsgWithStability("Decode stage latency in seconds, by connector.", compbasemetrics.ALPHA),
+			Buckets:   latencyBuckets,
+		},
+		[]string{"connector"},
+	)
```
shmuelk: What about encode?

pkg/sidecar/proxy/chat_completions.go
```diff
@@ -165,6 +167,7 @@ func (s *Server) disaggregatedPrefillHandler(apiType APIType) http.HandlerFunc {
 				attribute.Int("llm_d.ec_proxy.encoder_count", len(allowedEncoders)),
 				attribute.Int("llm_d.ec_proxy.encoder_candidates", len(encoderHostPorts)),
 			)
+			metrics.RecordDisagg(s.config.ECConnector)
```
shmuelk: I will point out that this is for encode.

shmuelk:
Another general comment.

An additional HTTP serving port was added, presumably in the style of the EPP. While this is needed in the EPP, since the base port is gRPC and not plain HTTP. In the case of the disagg siodecar, there already is an HTTP listener. Why not simply add a path to its mux?

pkg/sidecar/metrics/metrics.go
gyliu513: Done

pkg/sidecar/metrics/metrics.go
gyliu513: yes, encode is missing here, adding a StageEncode constant and an encode_duration_seconds histogram, and recording it on the EC connector path alongside prefill/decode.

pkg/sidecar/metrics/metrics.go
gyliu513: yes, added encode.

pkg/sidecar/proxy/chat_completions.go
gyliu513: Yes, this is the encode branch. It'll stay as the disagg counter for the EC connector, and with the new encode_duration_seconds / StageEncode metrics the encode path will be fully covered.

gyliu513:
> An additional HTTP serving port was added, presumably in the style of the EPP. While this is needed in the EPP, since the base port is gRPC and not plain HTTP. In the case of the disagg siodecar, there already is an HTTP listener. Why not simply add a path to its mux?

My original thinking was to mirror the EPP layout and keep metrics on a dedicated port (default 9090, the conventional Prometheus port) so that scraping and firewalling stay independent of the data-plane serving port — operators could expose 9090 to Prometheus without touching the request path, and disable it cleanly with --metrics-port=0.

That said, you're right that the EPP needs a second listener only because its base port is gRPC, and that reasoning doesn't carry over here — the sidecar's serving port is already plain HTTP. Adding a /metrics path to the existing mux is the simpler design and lets me drop the standalone server, the --metrics-port flag, and its validation/tests. I'll switch to that.

pkg/sidecar/metrics/metrics.go
```diff
@@ -0,0 +1,153 @@
...
+		[]string{"api_type"},
+	)
+
+	disaggRequestsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: subsystem,
+			Name:      "disagg_requests_total",
+			Help:      metricsutil.HelpMsgWithStability("Total requests routed through disaggregation, by connector.", compbasemetrics.ALPHA),
+		},
+		[]string{"connector"},
```
shmuelk: Since the connectors used in an execution of the sidecar are fixed (one for KV and another for EC), counting requests by connector isn't that usefull. I think it would be better to to track by disaggregation type (p/d, e/pd, etc., if possible)

pkg/sidecar/metrics/metrics.go
```diff
@@ -0,0 +1,153 @@
...
+	)
+
+	encodeDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: subsystem,
+			Name:      "encode_duration_seconds",
+			Help:      metricsutil.HelpMsgWithStability("Encode stage latency in seconds, by connector.", compbasemetrics.ALPHA),
+			Buckets:   latencyBuckets,
+		},
+		[]string{"connector"},
```
shmuelk: Again here the label doesn't add anything

pkg/sidecar/metrics/metrics.go
```diff
@@ -0,0 +1,153 @@
...
+	)
+
+	prefillDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: subsystem,
+			Name:      "prefill_duration_seconds",
+			Help:      metricsutil.HelpMsgWithStability("Prefill stage latency in seconds, by connector.", compbasemetrics.ALPHA),
+			Buckets:   latencyBuckets,
+		},
+		[]string{"connector"},
```
shmuelk: Again here the label doesn't add anything

pkg/sidecar/metrics/metrics.go
```diff
@@ -0,0 +1,153 @@
...
+	)
+
+	decodeDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: subsystem,
+			Name:      "decode_duration_seconds",
+			Help:      metricsutil.HelpMsgWithStability("Decode stage latency in seconds, by connector.", compbasemetrics.ALPHA),
+			Buckets:   latencyBuckets,
+		},
+		[]string{"connector"},
```
shmuelk: Again here the label doesn't add anything

pkg/sidecar/metrics/metrics.go
```diff
@@ -0,0 +1,153 @@
...
+	)
+
+	errorsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: subsystem,
+			Name:      "request_errors_total",
+			Help:      metricsutil.HelpMsgWithStability("Total prefill/decode stage errors, by connector and stage.", compbasemetrics.ALPHA),
+		},
+		[]string{"connector", "stage"},
```
shmuelk: Again here the connector label doesn't add anything

elevran:
@gyliu513 if still working on PR, please resolve conflicts and review comments. Otherwise, please kindly close.

gyliu513:
@elevran Will update this soon when I'm back.

pkg/sidecar/metrics/metrics.go
gyliu513: Done

pkg/sidecar/metrics/metrics.go
gyliu513: Done

pkg/sidecar/metrics/metrics.go
gyliu513: Done

pkg/sidecar/metrics/metrics.go
gyliu513: Done

pkg/sidecar/metrics/metrics.go
gyliu513: Done

gyliu513:
@shmuelk The e2e-router failure on the last two pushes turns out to be caused by moving /metrics onto the sidecar's serving mux (the change we discussed in the earlier thread), not by the disaggregation-type commit.

What happens: the sidecar's HTTP listener is a transparent reverse proxy — every unregistered path falls through to the decoder via mux.Handle("/", s.decoderProxy). In the DP e2e configuration the InferencePool targetPorts point at the sidecar's ports (8000/8001), so the EPP scrapes the model server's /metrics through the sidecar. Registering GET /metrics on the mux shadows that: the EPP now receives the sidecar's own metrics instead of vLLM's, never sees fresh vllm:* families, and inference_pool_ready_pods stays 0 — which is exactly the DP spec timeout in CI ("Expected 0 to be > 0" in waitForEPPToDiscoverPods). The non-DP e2e configs delete the sidecar and run vLLM directly on 8000, which is why only the DP spec fails.

Why this didn't show up when the mux change first went in: the June CI run passed against a main that still used fixed sleeps in e2e. #1988 (merged July 15) replaced the sleeps with polling inference_pool_ready_pods, and rebasing onto current main picked that gate up. So the green run in June was a false negative — even then, the EPP in DP configs was silently scraping sidecar metrics instead of vLLM metrics; it just wasn't asserted.

So the "there already is an HTTP listener" premise doesn't quite hold here: the listener's path namespace belongs to the model server, and /metrics is a path the EPP already depends on reaching through it. Two ways to fix:

1. Go back to a dedicated --metrics-port for the sidecar's own metrics (the first version of this PR), keeping the serving port fully transparent.
2. Keep the single listener but serve sidecar metrics on a non-colliding path such as /sidecar/metrics.

I lean toward option 1 — /metrics on its own port is what Prometheus tooling expects, and it avoids ever competing with the model server's path namespace again. WDYT? Thanks

gyliu513:
@shmuelk any comments for https://github.com/llm-d/llm-d-router/pull/1677#issuecomment-5049521706 ? Thanks

roytman:
> @shmuelk any comments for [#1677 (comment)](https://github.com/llm-d/llm-d-router/pull/1677#issuecomment-5049521706) ? Thanks

Hi @gyliu513 , unfortunately, @shmuelk has retired. Per the comment mentioned, I agree with your suggestion to use a dedicated --metrics-port.

gyliu513:
@roytman @elevran updated, can you help review?

pkg/sidecar/metrics/metrics.go
```diff
@@ -0,0 +1,158 @@
...
+	StagePrefill = "prefill"
+	StageDecode  = "decode"
+)
+
+// Disaggregation type labels: which stages of the request are split across pods.
+const (
+	DisaggTypePD  = "p/d"
+	DisaggTypeEPD = "e/p/d"
+	DisaggTypeED  = "e/d"
+)
```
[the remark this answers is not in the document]
roytman: Please update the PR description.

pkg/sidecar/proxy/connector_sglang.go
```diff
@@ -149,6 +152,7 @@ func (s *Server) handleSGLangConcurrentRequests(w http.ResponseWriter, r *http.R
 	s.decoderProxy.ServeHTTP(w, decodeReq)
 
 	decodeDuration := time.Since(decodeStart)
+	metrics.RecordDecodeDuration(decodeDuration)
```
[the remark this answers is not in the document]
roytman: Correct, `SGLang` decode path never increments `request_errors_total{stage="decode"}` on a failed decode

pkg/sidecar/proxy/connector_mooncake.go
```diff
@@ -248,6 +251,7 @@ func (s *Server) handleMooncakeConcurrentRequests(w http.ResponseWriter, r *http
 	s.decoderProxy.ServeHTTP(w, decodeReq)
 
 	decodeDuration := time.Since(decodeStart)
+	metrics.RecordDecodeDuration(decodeDuration)
```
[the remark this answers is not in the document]
roytman: as above

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -409,6 +412,7 @@ retryLoop:
 	}
 
 	decodeDuration := time.Since(decodeStart)
+	metrics.RecordDecodeDuration(decodeDuration)
 	decodeSpan.SetAttributes(attribute.Float64("llm_d.pd_proxy.decode.duration_ms", float64(decodeDuration.Milliseconds())))
```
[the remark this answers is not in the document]
roytman: decode duration is recorded (`connector_nixlv2.go:445`) but no status/error path exists; `finalizeDecodeWriter()` erroring returns early without recording anything at all

pkg/sidecar/metrics/metrics.go
```diff
@@ -0,0 +1,160 @@
...
+
+// subsystem is the Prometheus subsystem for the disaggregation sidecar metrics.
+const subsystem = "llm_d_disagg_sidecar"
+
+// Stage labels for encode/prefill/decode error attribution.
+const (
+	StageEncode  = "encode"
+	StagePrefill = "prefill"
+	StageDecode  = "decode"
```
roytman, line 43: `StageDecode` has no producer anywhere in this PR, grep confirms zero `RecordError(StageDecode)` call sites outside the test file. `request_errors_total{stage="decode"}` will never appear in a scrape, so an alert built on it never fires, and the help text at line 108 (`Total encode/prefill/decode stage errors`) is inaccurate about what the metric actually carries. 
This is the same gap Copilot flagged per-connector below (sglang/mooncake/nixlv2 decode paths), fixing those closes this too, but as shipped, an exported constant exists with no code path that produces it.

pkg/sidecar/metrics/metrics_test.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Register must be idempotent so repeated calls do not panic on duplicate
+// registration with controller-runtime's registry.
+func TestRegisterIdempotent(t *testing.T) {
+	require.NotPanics(t, func() {
+		Register()
+		Register()
+	})
+}
```
roytman, line 96: Nothing in this suite pins the actual exported metric names (`llm_d_disagg_sidecar_requests_total`, etc.) that four doc comments elsewhere promise operators. A rename of the subsystem or any `Name:` field would pass this whole suite while silently breaking every scrape config/dashboard built on it. 
Suggest one `promtestutil.GatherAndCount/CollectAndCompare` assertion against the registered name after Register().

pkg/sidecar/proxy/connector_ec_nixl.go
```diff
@@ -102,14 +104,17 @@ func (s *Server) handleECNIXL(w http.ResponseWriter, r *http.Request, prefillEnd
 
 	// Step 1: fan out to encoders, collect per-image ec_transfer_params.
 	if len(encodeEndPoints) > 0 {
+		encodeStart := time.Now()
```
roytman, line 107: Two asymmetries here: 

1. `fanoutEncoderCollect` returns `total == 0` immediately when there's no multimodal input (confirmed by the total > 0 guard two lines below), so a text-only request still gets a sub-microsecond encode_duration_seconds sample, skewing the lowest histogram bucket. 
2. The error branch at line 110 returns before line 117's Observe call, so this histogram is success-only, while `prefill_duration_seconds` in the P/D connectors is recorded before the status check and covers both outcomes - two "stage latency" histograms with different inclusion rules aren't comparable. 

Suggest only observing when an encoder was actually invoked, and recording on both outcomes for consistency with the prefill histograms.

pkg/sidecar/proxy/connector_ec_nixl.go
```diff
@@ -102,14 +104,17 @@ func (s *Server) handleECNIXL(w http.ResponseWriter, r *http.Request, prefillEnd
...
 		params, contributed, total, err := s.fanoutEncoderCollect(r.Context(), completionRequest, encodeEndPoints, requestID)
```
roytman: `pkg/sidecar/proxy/connector_ec_shared_storage.go:61` (on the fanoutEncoderPrimer error branch)

`handleECNIXL`'s structurally identical encode-fanout block (`connector_ec_nixl.go:107-117`) got `metrics.RecordEncodeDuration/metrics.RecordError(StageEncode)`; this one didn't. Both connectors are reachable from the same handleECConnector dispatch and both drive `disagg_requests_total{disagg_type="e/p/d"|"e/d"}`, so with --ec-connector=ec-example the encode-stage histograms/counters stay at zero regardless of actual encoder failures. 

Suggest mirroring the ec-nixl instrumentation there.

pkg/sidecar/proxy/connector_nixlv2.go
roytman, line 489: This whole `--moriio-parallel-dispatch` path (lines 479-804) records no metric at all, not prefill duration (line 703 sets the span attribute but never calls `RecordPrefillDuration`), not decode duration (line 726, same gap vs. `RecordDecodeDuration)`, not prefill error (line 705's `isHTTPError` branch never calls `RecordError(StagePrefill)`, unlike the serial path's equivalent at line 252), and not the KV-wait-timeout abort at line 772 (arguably a prefill error by any reading). 
Meanwhile, `chat_completions.go` still counts these as `disagg_type="p/d"` requests. With this flag enabled, the dashboard shows P/D traffic with zero stage-latency samples. 

This can be resolved in a separate PR.

pkg/sidecar/proxy/proxy.go
```diff
@@ -624,6 +626,7 @@ func (s *Server) createRoutes() *http.ServeMux {
 	mux.HandleFunc("GET /health", func(w http.ResponseWriter, _ *http.Request) {
 		w.WriteHeader(http.StatusOK)
 	})
+
```
roytman: **nit**: do we need this blank line?

pkg/sidecar/proxy/options.go
```diff
@@ -282,7 +282,7 @@ func (opts *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&opts.PoolGroup, poolGroup, opts.PoolGroup, "group of the InferencePool this Endpoint Picker is associated with.")
 	fs.IntVar(&opts.DecodeChunkSize, decodeChunkSize, opts.DecodeChunkSize, "enables chunked decode mode when > 0; value is the token budget per chunk. For best performance should be a multiple of the block size.")
 	fs.BoolVar(&opts.Tracing, tracingFlag, opts.Tracing, "Enable OpenTelemetry tracing")
-	fs.IntVar(&opts.MetricsPort, metricsPort, opts.MetricsPort, "Port for the Prometheus /metrics endpoint (exposes the moriio_dns_* counters). 0 (the default) disables it. Takes precedence over the MORIIO_METRICS_ADDR env var.")
+	fs.IntVar(&opts.MetricsPort, metricsPort, opts.MetricsPort, "Port for the Prometheus /metrics endpoint (exposes the moriio_dns_* and llm_d_disagg_sidecar_* counters). 0 (the default) disables it. Takes precedence over the MORIIO_METRICS_ADDR env var.")
```
roytman, line 283: The release note says metrics default to port 9090, with `--metrics-port=0` to disable, but this line (and the merge logic further down) confirms the default is `0` -  i.e. disabled by default, and operators must explicitly pass `--metrics-port=<port>` to turn metrics on at all. Worth synchronizing the release note's port-default claim with the implementation.

roytman (commented):
Thank you @gyliu513 for adding metrics to the sidecar, and sorry for the long review process.

Several comments:

`pkg/sidecar/proxy/connector_shared_storage.go` and  `/ pkg/sidecar/proxy/connector_p2p.go`

Neither of these two connectors records any stage metric, while both still drive `disagg_requests_total{disagg_type="p/d"}` via chat_completions.go. That leaves 2 of 5 P/D connectors producing a disagg count with zero corresponding latency/error data, an operator has to read all five files to know which deployments actually populate the new histograms.
This can be resolved in a separate PR, but worth mentioning a relevant issue number.

pkg/sidecar/metrics/metrics.go
gyliu513: Confirmed. StageDecode has no producer yet. This is the same gap as the per-connector decode-path issues noted elsewhere in this review; I'll track it under the same follow-up issue as you mentioned rather than a separate one, since fixing the connector instrumentation closes this too

pkg/sidecar/metrics/metrics_test.go
gyliu513: Good catch, will add. Planning to add a CollectAndCompare (or GatherAndCount) assertion in TestRegisterIdempotent that pins the registered metric names after Register(), so a subsystem/Name rename fails the test instead of silently drifting from the doc comments.

pkg/sidecar/proxy/connector_ec_nixl.go
gyliu513: Both points confirmed. Will fix by moving the RecordEncodeDuration call inside the `total > 0` branch (so text-only requests don't skew the histogram) and adding a RecordError(StageEncode)/RecordEncodeDuration call on the error path too, matching how prefill_duration_seconds records both outcomes in the P/D connectors.

pkg/sidecar/proxy/connector_ec_nixl.go
gyliu513: Yes, will mirror the ec-nixl instrumentation here. RecordEncodeDuration on success, RecordError(StageEncode) on the error branch, once the fix from the connector_ec_nixl.go comment above lands, since I'd rather fix the pattern once and copy the corrected version.

pkg/sidecar/proxy/connector_nixlv2.go
gyliu513: Yes, the moriio-parallel-dispatch path (lines 479-804) has zero metric coverage: no RecordPrefillDuration, RecordDecodeDuration, RecordError(StagePrefill), and the KV-wait-timeout abort at line 772 isn't recorded as an error either. Agreed this is separate-PR scope given the size; I'll include it in the same follow-up issue your comments above.

pkg/sidecar/proxy/proxy.go
gyliu513: Will remove.

pkg/sidecar/proxy/options.go
gyliu513: Good catch, will fix. The default is actually 0 (disabled), operators must pass `--metrics-port` explicitly to enable it. Updating the release note to match.

pkg/sidecar/metrics/metrics.go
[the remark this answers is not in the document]
gyliu513: Agreed, will update the PR description/release note to match the implemented subsystem (llm_d_disagg_sidecar) and label schema rather than changing the metric names, a connector label would require converting the plain Histograms to HistogramVecs, which is a bigger change than this PR's scope.

pkg/sidecar/proxy/connector_sglang.go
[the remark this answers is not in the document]
gyliu513: Yes, will wrap `w` with a status-capturing ResponseWriter around ServeHTTP and call RecordError(StageDecode) on a non-2xx final status, matching the prefill error path above it.

pkg/sidecar/proxy/connector_nixlv2.go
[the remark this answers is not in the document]
gyliu513: Good catch on the finalizeDecodeWriter() early-return too, right now a flush failure records neither duration nor an error, so that path is invisible in metrics entirely. Will capture the decode status via the cachedTokensResponseWriter (as suggested) and call RecordError(StageDecode) both on that early-return and on a non-2xx final status.

pkg/sidecar/proxy/connector_mooncake.go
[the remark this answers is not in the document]
gyliu513: Same fix as the sglang decode path above, will wrap the response writer and call RecordError(StageDecode) here too.

gyliu513 (commented):
Thanks for the thorough review, @roytman!

gyliu513:
@roytman follow up issue https://github.com/llm-d/llm-d-router/issues/2755

gyliu513:
@roytman can you help review again? I think all comments are now addressed.

gyliu513:
@roytman @elevran @ahg-g I think this is ready to merge, can you help check? Thanks!

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -378,12 +382,19 @@ retryLoop:
...
 		decodeSpan.SetStatus(codes.Error, "failed to flush cached token response writer")
 		return
 	}
 
 	decodeDuration := time.Since(decodeStart)
+	metrics.RecordDecodeDuration(decodeDuration)
+	if statusWriter.statusCode < 200 || statusWriter.statusCode >= 300 {
+		metrics.RecordError(metrics.StageDecode)
+		decodeSpan.SetStatus(codes.Error, "decode request failed")
+	}
```
roytman, line 397: This file's serial decode path is now fully instrumented above, but `runNIXLProtocolV2WriteParallel` further down in this same file (~line 430, the `MoRIIOParallelDispatch && MoRIIOWriteMode` config) has no `metrics.Record*` calls at all,  not on prefill failure, KV-wait timeout, or decode-abort replay. It's a real, selectable production path, not dead code, and it already imports this same `metrics` package for the path above.

pkg/sidecar/metrics/metrics_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+// TestMetricNames pins the fully-qualified names this package exports, so a
+// subsystem or Name field rename fails this test instead of silently
+// breaking every scrape config/dashboard built on the doc comments' promises.
+func TestMetricNames(t *testing.T) {
```
roytman, line 101: Everything above tests the `metrics` package in isolation (counters, histogram deltas, label values, `Register()` idempotency, this name-pinning test). None of the ~15 call sites in `pkg/sidecar/proxy/` are verified to actually invoke the right `Record*` with the right label from the right branch, e.g. the three-way `disagg_type` selection in `dispatch.go`, or the encode-error branch in `connector_ec_nixl.go`/`connector_ec_shared_storage.go` (existing handler tests only reach a 200 from the encoder; the failure-path tests call `fanoutEncoder*` directly, bypassing the handler). A label swap at any call site would compile and pass every existing test untouched.

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -378,12 +382,19 @@ retryLoop:
...
 	if err := finalizeDecodeWriter(); err != nil {
+		metrics.RecordDecodeDuration(time.Since(decodeStart))
+		metrics.RecordError(metrics.StageDecode)
 		s.logger.Error(err, "failed to flush cached token response writer")
 		decodeSpan.SetStatus(codes.Error, "failed to flush cached token response writer")
 		return
 	}
 
 	decodeDuration := time.Since(decodeStart)
+	metrics.RecordDecodeDuration(decodeDuration)
```
roytman, line 393: Once headers are written and the stream breaks mid-flight (client disconnect, LB timeout), `httputil.ReverseProxy`'s copy loop panics with `http.ErrAbortHandler`, which `net/http` deliberately swallows without logging. That panic unwinds past this line and line 395 - `RecordDecodeDuration`/`RecordError(StageDecode)` never run. The parallel MoRI-IO decode path elsewhere in this file already has a `recover()` for this exact panic; this serial path doesn't. Net effect: `request_errors_total{stage="decode"}` undercounts exactly the failure mode (mid-stream disconnect) an operator most needs visibility into.

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -378,12 +382,19 @@ retryLoop:
...
+	if statusWriter.statusCode < 200 || statusWriter.statusCode >= 300 {
```
roytman, line 394: nit: This reinvents `isHTTPError()` (`statusCode < 200 || statusCode >= 300`) instead of calling it — `connector_pd_common.go:121`, added in this same PR, uses the shared helper for the identical check. Worth using it here too for consistency.

pkg/sidecar/proxy/connector_pd_common.go
```diff
@@ -110,9 +113,15 @@ func (s *Server) runConcurrentPD(
 	decodeStart := time.Now()
 
 	decodeReq = decodeReq.WithContext(ctx)
-	s.decoderProxy.ServeHTTP(w, decodeReq)
+	decodeWriter := &statusCapturingResponseWriter{ResponseWriter: w}
+	s.decoderProxy.ServeHTTP(decodeWriter, decodeReq)
```
roytman, line 117: Same panic-swallows-metrics concern as the analogous decode path in `connector_nixlv2.go`: if `s.decoderProxy.ServeHTTP` panics with `http.ErrAbortHandler` after writing headers (client disconnect mid-stream), execution never reaches the `metrics.RecordDecodeDuration`/`RecordError` calls below.

pkg/sidecar/proxy/connector_ec_nixl.go
```diff
@@ -120,15 +122,24 @@ func (s *Server) handleECNIXL(w http.ResponseWriter, r *http.Request, prefillEnd
 
 	// Step 1: fan out to encoders, collect per-image ec_transfer_params.
 	if len(encodeEndPoints) > 0 {
+		encodeStart := time.Now()
 		params, contributed, total, err := s.fanoutEncoderCollect(r.Context(), body, encodeEndPoints, requestID)
 		if err != nil {
+			// fanoutEncoderCollect only errors once it has actually dispatched
```
roytman, line 128: **Factual drift:** "only errors once it has actually dispatched to an encoder" doesn't hold in every case, and the case where it doesn't is an operationally important one.

`fanoutEncoderCollect` delegates to `fanoutEncoder` (`connector_ec_common.go:189`), which launches one goroutine per multimodal item and blocks in `errgroup.Wait()`. Two situations return an error with zero encoder round-trips made:

1. `len(encoderHostPorts) == 0` (`connector_ec_common.go:197`) — checked before any goroutine is spawned, so the return is near-instant every time.
2. Every item fails at the same pre-dispatch step (`json.Marshal`, `encoderProxyHandler` construction, or `http.NewRequestWithContext`) at roughly the same time. This isn't rare: `encoderHostPorts[idx%len(encoderHostPorts)]` round-robins across configured hosts, so with a single configured encoder host — a common deployment — a handler-construction failure (bad URL, broken TLS config, host unreachable in a way that fails before `ServeHTTP`) hits every item, every request, concurrently and near-instantly.

(With multiple items and a genuine mixed outcome, `errgroup.Wait()` blocks until *all* goroutines finish, so a slow real dispatch elsewhere in the same fanout keeps the overall duration meaningful — that's the case where the comment's claim happens to hold.)

Case 2 matters because it's exactly an encoder-outage/misconfiguration scenario: right when you want `encode_duration_seconds` to give a clean latency read, this floods it with near-zero samples, pulling p50/p90 down instead of showing the outage as elevated latency or as simply absent (like the `total == 0` case at line 139 already handles by skipping the sample).

**Suggested fix:** drop `metrics.RecordEncodeDuration(time.Since(encodeStart))` from this `err != nil` branch (same change in `connector_ec_shared_storage.go`'s identical branch) and keep only `metrics.RecordError(metrics.StageEncode)`. That makes the error path consistent with the `total == 0` policy; both agree "no representative signal" means "don't sample the histogram,"  and the comment can just be deleted along with the line rather than rewritten. If failed-after-a-while dispatch latency is actually a signal worth keeping, that's better done by threading a real "did we reach `ServeHTTP`" count through `fanoutEncoder` (mirroring the `total` value this PR already added to `fanoutEncoderPrimer`), **but that's a bigger change than this PR comment should try to resolve inline.**

pkg/sidecar/proxy/connector_ec_shared_storage.go
```diff
@@ -60,13 +64,25 @@ func (s *Server) handleECSharedStorage(w http.ResponseWriter, r *http.Request, p
...
+			metrics.RecordEncodeDuration(time.Since(encodeStart))
+			metrics.RecordError(metrics.StageEncode)
 			s.logger.Error(err, "encoder processing failed", "requestID", requestID)
 			if err := errorBadGateway(err, w); err != nil {
 				s.logger.Error(err, "failed to send error response to client")
 			}
 			return
 		}
+		if total > 0 {
+			// total == 0 means there was no multimodal input to encode, so no
```
roytman, line 81: nit: This 3-line comment is byte-identical to `connector_ec_nixl.go:139`  and sits inside the `if total > 0` branch explaining the branch that *isn't* taken (`total == 0`). Since `connector_ec_nixl.go` implements the NIXL connector (`ECConnectorNIXL`, the more production-oriented path per `proxy.go:591`) and this file implements `ECExampleConnector`, consider treating the NIXL copy as canonical: move the comment above the `if total > 0 {` guard there so it reads as the guard's rationale, and either drop this copy in favor of a short reference back to it, or keep both but make sure they stay in sync — right now a future edit to one has no reason to touch the other, and they'll drift.

pkg/sidecar/proxy/connector_ec_shared_storage.go
```diff
@@ -60,13 +64,25 @@ func (s *Server) handleECSharedStorage(w http.ResponseWriter, r *http.Request, p
 
 	// Step 1: Process through Encoder cluster (if has MM input)
 	if len(encodeEndPoints) > 0 {
-		if err := s.fanoutEncoderPrimer(r.Context(), body, encodeEndPoints, requestID); err != nil {
+		encodeStart := time.Now()
+		total, err := s.fanoutEncoderPrimer(r.Context(), body, encodeEndPoints, requestID)
+		if err != nil {
+			// fanoutEncoderPrimer only errors once it has actually dispatched
```
roytman, line 70: Same factual drift as the identical comment in `connector_ec_nixl.go:128` — see that comment for detail (`fanoutEncoderPrimer` delegates to `fanoutEncoder`, which can error pre-dispatch).

pkg/sidecar/proxy/status_response_writer.go
```diff
@@ -55,6 +55,37 @@ func (w *bufferedResponseWriter) bodyBytes() []byte {
...
+}
+
+func (w *statusCapturingResponseWriter) WriteHeader(statusCode int) {
+	if w.statusCode == 0 {
+		w.statusCode = statusCode
+	}
+	w.ResponseWriter.WriteHeader(statusCode)
+}
+
+func (w *statusCapturingResponseWriter) Write(b []byte) (int, error) {
```
roytman, line 74: Sets `statusCode = 200` before checking whether the delegated `Write` actually succeeds, and never inspects the returned error to revise that classification. A failed delivery (broken pipe, TLS error) gets fed into `isHTTPError(decodeWriter.statusCode)` as a false success. Separately, the doc comment above the struct (line 59) says this records the "final" status code, but the `if w.statusCode == 0` guard in both `WriteHeader` and `Write` actually captures the *first* one written — worth fixing the wording too.

pkg/sidecar/metrics/metrics.go
```diff
@@ -0,0 +1,160 @@
...
+const (
+	DisaggTypePD  = "p/d"
+	DisaggTypeEPD = "e/p/d"
+	DisaggTypeED  = "e/d"
+)
+
+// latencyBuckets covers encode, prefill, and decode wall-clock latency in
+// seconds, from a few milliseconds to several minutes (decode can stream for a
+// long time).
+var latencyBuckets = []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 30, 60, 120, 300}
```
roytman, line 56: `pkg/common/observability/metrics` already defines `GeneralLatencyBuckets` with the explicit contract that every llm-d component reuses it so PromQL translates cleanly across components (see `pkg/coordinator/metrics` for the analogous `step_duration_seconds` histogram using it). Worth reusing here instead of a bespoke bucket list, it also reaches further out (3600s vs this 300s ceiling), which fits the "decode can stream for a long time" rationale better.

pkg/sidecar/metrics/metrics.go
```diff
@@ -0,0 +1,160 @@
...
+// Stage labels for encode/prefill/decode error attribution.
+const (
+	StageEncode  = "encode"
+	StagePrefill = "prefill"
+	StageDecode  = "decode"
+)
+
+// Disaggregation type labels: which stages of the request are split across pods.
+const (
+	DisaggTypePD  = "p/d"
```
roytman, line 48: `pkg/coordinator/metrics` already names this exact concept with `PathDecodeOnly = "decode-only"`, `PathPrefillDecode = "prefill-decode"`, `PathEncodePrefillDecode = "encode-prefill-decode"`. This introduces a second vocabulary (`"p/d"`, `"e/p/d"`, `"e/d"`) for the same dimension, which makes a cross-component dashboard joining router and sidecar stage paths need a translation table. Worth matching the coordinator's existing spelling.

roytman (commented):
Hi @gyliu513 , sorry for the late review. 

I added several comments, some of which may be out of scope for this PR and can be addressed separately. Feel free to open an issue(s)

Additionally, several dispatch protocols lack instrumentation entirely: connector_shared_storage.go (maybe this one isn't so important), connector_p2p.go, and the chunked-decode path in decode.go compute isHTTPError(...); they handle errors themselves but never call into the new metrics package.

Issue #545, opened 2026-01-08
Topology Aware Routing to lower KV Cache transfer latency


RishabhSaini:
### Problem
 Prefill and decode pods are currently selected independently without network locality awareness. KV cache transfer between pods on different nodes incurs inter-node network latency (infiniband upto 400 GB/s)
 vs. intra-node transfer (NVLink upto 900GB/s), directly impacting TTFT SLO compliance.

 ### Proposal
 Add topology-aware prefill pod selection:
 - After selecting decode pod, prefer prefill pods on same node (giving them a higher score)
 - Fallback to global selection if non-colocated pods still have a higher score

RishabhSaini:
> You can use existing k8s labels for [hostname](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetesiohostname), [zone](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone), and [region](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion). We already collect labels for pods.

Could you point to me in code where are these node, zone, regions labels are set in modelService pods? I can only seem to find nodeName in the spec of the modelService?

elevran:
A scorer should not make runtime calls to the k8s API. It is expected to operate on existing data only to keep scheduling overheads minimal.

The node, zone and region are Pod labels. We already store labels for the Pod so retrieving those (**if** defined) should be trivial.
These are (presumably) standard k8s labels, independent of any ModelServer abstraction we may have.
The scorer shuold set higher P/D scores based on location affinity. If co-location of P and D nodes can not be determined the scorer should given a zero score to all Pods equally.

Therefore, my suggestion is to implement the scorer based on existing labels when available and assign zero score when not available.
If multiple labels exist (node, zone and region) give higher marks to closer entities (e.g., 1.0 for same node, 0.5 for same zone and 0.25 for same region).

elevran:
We currently do not track node objects s in llm-d or GAIE.
Let me think if we can enable this in GAIE first (as a custom data source) and then can add relevant extractor for both Node and Pod objects. Alternative would be to create a Pod data source +custom extractor and use the k8s Client to retrieve/cache node labels.

ahg-g:
One drawback of this proposal is that the decode endpoint is not picked in a topology aware manner, it is picked in silo.

[...]

PR #1678, opened 2026-06-17
feat(epp): add topology-extractor datalayer plugin

elevran:
**What type of PR is this?**
/kind feature

Adds a `topology-extractor` endpoint extractor that stamps each endpoint with a `Topology` attribute (hostname, rack, zone, region) for use by topology-aware scoring.

The plugin registers for both endpoint lifecycle events and Pod k8s notification events. Label names for each topology field are configured via `hostname`, `rack`, `zone`, and `region` parameters; when omitted, the standard Kubernetes topology labels are used as defaults:
- hostname: `kubernetes.io/hostname` (corev1.LabelHostname)
- rack: `topology.kubernetes.io/rack` (does not exist in corev1 but follows the same pattern)
- zone: `topology.kubernetes.io/zone` (corev1.LabelTopologyZone)
- region: `topology.kubernetes.io/region` (corev1.LabelTopologyRegion)

Labels are read from the endpoint's pod metadata at endpoint event time. When the hostname label is absent on a pod, the plugin falls back to `spec.hostname` from the Pod notification event. Rack, zone and region have no fallback.

Note: zone and region values are not read from k8s Node objects as that would require additional RBAC to allow GET/LIST on Nodes. Rack is not a standard k8s label defined on nodes, so read from endpoint labels alone. The topology fields are populated when the pod itself carries the corresponding labels (e.g. propagated via the Downward API or admission webhook from the node's labels).

This is groundwork for topology-aware scoring in disaggregated inference: a planned `topology-locality-scorer` will score endpoints whose `Topology.Hostname` (or other topology attributes) matches a key on the request attribute store, enabling locality-aware routing that reduces KV cache transfer latency across nodes.

Refs #545

**Test plan**:

```release-note
Add `topology-extractor` datalayer plugin: stamps endpoints with a `Topology` attribute (hostname, rack, zone, region). Label names are configurable; defaults are the standard Kubernetes topology labels when they exist. Hostname falls back to `spec.hostname` when the label is absent.
```


pkg/epp/framework/plugins/datalayer/attribute/topology/README.md
```diff
@@ -0,0 +1,20 @@
...
+- **Fields**:
+  - `Hostname`: The host name of the endpoint. Sourced from `spec.hostname`
+    on the Pod object, or from a user-configured pod label.
+
+## Producers
+
+The following plugins produce this attribute:
+
+- **`topology-extractor`** (Data Layer): Sets the `Topology` attribute using
+  `spec.hostname` from the Pod object, or the value of a configured pod label.
```
elevran:
```suggestion
  `spec.hostname` from the Pod object, or the value of a configured endpoint label.
```

pkg/epp/framework/plugins/datalayer/extractor/topology/extractor.go
```diff
@@ -0,0 +1,273 @@
...
+
+// podGVK is the core/v1 Pod GVK watched by the pod notification handler.
+var podGVK = schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Pod"}
+
+// params holds the user-facing configuration for the topology extractor.
+type params struct {
+	// HostnameLabel is the pod label whose value is used as the topology hostname.
+	// When empty (default), spec.hostname from the Pod object is used instead.
+	// When set but absent on the pod, the attribute is not added.
+	HostnameLabel string `json:"hostnameLabel,omitempty"`
```
elevran: Am considering using attribute (e.g., Host, AZ, DC) and corresponding label name. When defined, the labels value is used. If omitted, standard k8s topology labels are used.

vMaroon: I think this will be useful

pkg/epp/framework/plugins/datalayer/extractor/topology/extractor.go
```diff
@@ -0,0 +1,273 @@
...
+			return nil
+		}
+		event.Endpoint.GetAttributes().Put(h.ext.dk.String(), &attrtopology.Topology{Hostname: val})
+		return nil
+	}
+
+	// No-label path: track endpoint; stamp immediately if hostname already cached.
+	key := podKey(meta)
+	h.ext.mu.Lock()
+	h.ext.endpoints[key] = append(h.ext.endpoints[key], event.Endpoint)
```
vMaroon: This action is non-idempotent: `EventAddOrUpdate` re-fires on update with the same endpoint, so this append can duplicate entries. The inner slice can be keyed by `NamespacedName` as well which does cover the distinction for multiple-ranks, although might be a (currently exercised) fragile representation.

elevran: switched to using a map so the same entry is reused instead of appending new entries.

elevran:
@vMaroon anything else needed on this?

johnpanos:
Thanks for all the work on this, all. This would be huge for us.

pkg/epp/framework/plugins/datalayer/attribute/topology/data_types.go
```diff
@@ -0,0 +1,53 @@
...
+	TopologyExtractorType = "topology-extractor"
+)
+
+// TopologyAttributeKey identifies the Topology attribute stored on each endpoint.
+var TopologyAttributeKey = plugin.NewDataKey("Topology", TopologyExtractorType)
+
+// Topology carries the locality information for an endpoint.
+// Fields are populated from pod labels at endpoint creation time.
+// Hostname falls back to spec.hostname when the configured label is absent.
+type Topology struct {
```
ahg-g, line 37: The rack (or block) is also important. I know it is not a standard label in k8s, but we can make the label key a config parameter.

elevran: added, Used a default value that matches the default k8s labels format (despite not having one defined)

ahg-g (approved):

PR #1684, opened 2026-06-18
Delegate conditional-decode 412 decision to prefix-based-pd-decider

dmitripikus:
**What type of PR is this?**

/kind feature

Move the RFC 7240 `Prefer: if-available` 412 gate out of the director and into
the `prefix-based-pd-decider` plugin. One `nonCachedTokens` threshold now drives
both decisions:
disaggregated P/D and conditional-decode rejection.

The director had simplistic and non-configurable implementation of decode condition.

Behavior change: the 412 gate now rejects when the non-cached suffix is at or above `nonCachedTokens` (rather than rejecting only on a complete cache miss), and `nonCachedTokens: 0` disables the gate entirely. Deployments wanting the 412 gate must include `- type: prefix-based-pd-decider` with a non-zero `nonCachedTokens` in their EPP config; deployments using the canonical `pd-epp-config.yaml` already declare the plugin and pick up the gate automatically. See `docs/disaggregation.md` for details.

[1686](https://github.com/llm-d/llm-d-router/issues/1686)

### Changes

- New `requestcontrol.ConditionalDecodeDecider` interface; auto-registered
  through `Config.AddPlugins`.
- `*PrefixBasedPDDecider` implements it via `ShouldRejectConditionalDecode`,
  sharing `computeNonCachedTokens` with the existing `disaggregate` path
  (different fail-mode semantics: 412 fail-closed, disaggregation fail-open).
- Director consults the configured decider

**Release note**
```release-note
    plugins:
     ...
     - type: prefix-based-pd-decider
      parameters:
        nonCachedTokens: <enter tokens number here>
    ...
    
 should be added to decoder configuration map to support conditional decode   
```

### DIFF
--- a/docs/disaggregation.md
+++ b/docs/disaggregation.md
@@ -425,7 +425,7 @@ The gate uses the same `nonCachedTokens` threshold as the disaggregation decisio
 - Plugin not declared in the config → gate disabled (the EPP forwards conditional-decode requests unconditionally).
 - Prefix-cache state unreadable on the chosen endpoint (e.g. no approximate-prefix producer wired up) → fail closed (412).
 
-This behavior differs from earlier releases, where the gate was a hard-coded check inside the director that forwarded on any cache hit and rejected only on a complete cache miss. The threshold is now in tokens (not blocks), is operator-configurable, and is opt-in via the plugin.
+The threshold is configured in tokens, drives both the disaggregation decision and the 412 gate, and is opt-in: omitting the plugin disables the gate.
 
 #### Always-Disagg PD Decider
 The `always-disagg-pd-decider` is a simpler alternative used mainly for testing or benchmarking.
--- a/docs/disaggregation.md
+++ b/docs/disaggregation.md
@@ -413,6 +413,20 @@ The `prefix-based-pd-decider` plugin makes the disaggregation decision according
 - `nonCachedTokens`: Number of non-cached tokens that trigger disaggregation
   - If set to 0, disaggregation never occurs for any request
 
+**Conditional-decode 412 gate**
+
+When this plugin is configured, it also drives the EPP's RFC 7240 `Prefer: if-available` (conditional-decode) gate. The coordinator uses that header to mark a speculative early-decode attempt: forward to a decode worker only if its KV cache already covers the prompt, otherwise return HTTP 412 Precondition Failed so the coordinator restarts the pipeline at encode/prefill/decode.
+
+The gate uses the same `nonCachedTokens` threshold as the disaggregation decision:
+
+- Non-cached suffix `< nonCachedTokens` → forward to the chosen decode worker.
+- Non-cached suffix `≥ nonCachedTokens` → return 412.
+- `nonCachedTokens: 0` → gate disabled (forwards every conditional-decode request).
+- Plugin not declared in the config → gate disabled (the EPP forwards conditional-decode requests unconditionally).
+- Prefix-cache state unreadable on the chosen endpoint (e.g. no approximate-prefix producer wired up) → fail closed (412).
+
+This behavior differs from earlier releases, where the gate was a hard-coded check inside the director that forwarded on any cache hit and rejected only on a complete cache miss. The threshold is now in tokens (not blocks), is operator-configurable, and is opt-in via the plugin.
+
 #### Always-Disagg PD Decider
 The `always-disagg-pd-decider` is a simpler alternative used mainly for testing or benchmarking.
 It always triggers disaggregation, regardless of prefix cache state or prompt characteristics.
--- a/docs/disaggregation.md
+++ b/docs/disaggregation.md
@@ -415,7 +415,10 @@ The `prefix-based-pd-decider` plugin makes the disaggregation decision according
 
 **Conditional-decode 412 gate**
 
-When this plugin is configured, it also drives the EPP's RFC 7240 `Prefer: if-available` (conditional-decode) gate. The coordinator uses that header to mark a speculative early-decode attempt: forward to a decode worker only if its KV cache already covers the prompt, otherwise return HTTP 412 Precondition Failed so the coordinator restarts the pipeline at encode/prefill/decode.
+When this plugin is configured, it also gates requests carrying the RFC 7240 `Prefer: if-available` header: a request is forwarded to the chosen decode worker only when the worker's KV cache already covers the prompt; otherwise the EPP returns HTTP 412 Precondition Failed.
+
+> **Note**
+> The llm-d coordinator uses this header to mark speculative early-decode attempts. A 412 response signals that the local cache was insufficient, prompting the coordinator to restart the pipeline at encode/prefill/decode.
 
 The gate uses the same `nonCachedTokens` threshold as the disaggregation decision:
 
--- a/pkg/epp/framework/interface/requestcontrol/plugins.go
+++ b/pkg/epp/framework/interface/requestcontrol/plugins.go
@@ -98,3 +98,12 @@ type PreAdmitter interface {
 	plugin.Plugin
 	PreAdmit(ctx context.Context, request *fwksched.InferenceRequest) error
 }
+
+// ConditionalDecodeDecider answers the RFC 7240 "Prefer: if-available" gate:
+// given the chosen decode endpoint, should the request be rejected with HTTP
+// 412 (because the local KV cache does not cover enough of the prompt) or
+// forwarded? At most one such plugin is consulted per director.
+type ConditionalDecodeDecider interface {
+	plugin.Plugin
+	ShouldRejectConditionalDecode(ctx context.Context, request *fwksched.InferenceRequest, endpoint fwksched.Endpoint) bool
+}
--- a/pkg/epp/framework/interface/requestcontrol/plugins.go
+++ b/pkg/epp/framework/interface/requestcontrol/plugins.go
@@ -99,10 +99,9 @@ type PreAdmitter interface {
 	PreAdmit(ctx context.Context, request *fwksched.InferenceRequest) error
 }
 
-// ConditionalDecodeDecider answers the RFC 7240 "Prefer: if-available" gate:
-// given the chosen decode endpoint, should the request be rejected with HTTP
-// 412 (because the local KV cache does not cover enough of the prompt) or
-// forwarded? At most one such plugin is consulted per director.
+// ConditionalDecodeDecider decides whether a request carrying the
+// RFC 7240 "Prefer: if-available" header should be rejected with HTTP 412
+// Precondition Failed or forwarded to the chosen decode endpoint.
 type ConditionalDecodeDecider interface {
 	plugin.Plugin
 	ShouldRejectConditionalDecode(ctx context.Context, request *fwksched.InferenceRequest, endpoint fwksched.Endpoint) bool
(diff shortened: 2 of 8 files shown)


pkg/epp/requestcontrol/director.go
```diff
@@ -61,49 +60,26 @@ const (
...
-// endpoint has at least one matching prefix block in its KV cache, as observed
-// by a precise/approximate-prefix scorer during the decode profile run. It
-// returns false when the result is missing, the primary profile produced no
-// endpoint, the endpoint carries no PrefixCacheMatchInfo attribute, or the
-// recorded match has zero blocks. False-return reasons are logged at
-// V(logutil.DEBUG) to disambiguate misconfiguration (no scorer attached) from
-// a real cache miss.
-func primaryEndpointHasCachedPrefix(logger logr.Logger, result *fwksched.SchedulingResult) bool {
+// primaryDecodeEndpoint returns the first endpoint chosen by the primary
+// profile, or nil when the result is empty or malformed. False-return reasons
```
roytman: I think the comment is stale.
The function no longer inspects the cache, it only returns the endpoint. So, the "cache miss" framing is now inaccurate.

pkg/epp/requestcontrol/director.go
```diff
@@ -330,14 +306,21 @@ func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestCo
 	// at encode/prefill/decode. Lives in the director (not in a profile handler)
 	// so it fires regardless of which profile handler is configured.
 	if routing.IsConditionalDecode(reqCtx.Request.Headers) {
-		if !primaryEndpointHasCachedPrefix(logger, result) {
-			logger.V(logutil.DEBUG).Info("conditional-decode: chosen decode worker has no cached prefix, returning 412")
-			return reqCtx, errcommon.Error{
-				Code: errcommon.PreconditionFailed,
-				Msg:  "no decode worker has the requested KV cache",
+		decider := d.requestControlPlugins.ConditionalDecodeDecider()
+		switch {
```
roytman:
```go
if decider == nil {
    logger.V(logutil.DEBUG).Info("conditional-decode: no decider configured, forwarding")
} else {
    endpoint := primaryDecodeEndpoint(logger, result)
    if endpoint == nil || decider.ShouldRejectConditionalDecode(ctx, reqCtx.SchedulingRequest, endpoint) {
        ...
    }
    ...
}
```

roytman:
Please open an issue and add it to the PR,

Behavior change in the 412 gate needs to be called out explicitly. This is the substantive change, and it is easy to miss:

- Old gate (primaryEndpointHasCachedPrefix): forwarded if the endpoint had any matching prefix block (MatchBlocks() > 0); rejected only on a complete cache miss.
- New gate: rejects when the non-cached suffix is >= nonCachedTokens, and when nonCachedTokens == 0 the gate is disabled entirely (always forwards).

We need to update the [README](https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/README.md#prefixbasedpddecider) file about the conditional decode behaviour.

pkg/epp/requestcontrol/director.go
dmitripikus: I've resolved this when taking care of lint problem.
Thanks!

dmitripikus:
@roytman , addressing your fist set of comments:
- issue is created and associated with PR
- behavior change is clarified in "README.md" you mentioned, and also in "disaggregation.md" and in PR description.
Thanks!

docs/disaggregation.md
```diff
@@ -413,6 +413,20 @@ The `prefix-based-pd-decider` plugin makes the disaggregation decision according
...
+
+The gate uses the same `nonCachedTokens` threshold as the disaggregation decision:
+
+- Non-cached suffix `< nonCachedTokens` → forward to the chosen decode worker.
+- Non-cached suffix `≥ nonCachedTokens` → return 412.
+- `nonCachedTokens: 0` → gate disabled (forwards every conditional-decode request).
+- Plugin not declared in the config → gate disabled (the EPP forwards conditional-decode requests unconditionally).
+- Prefix-cache state unreadable on the chosen endpoint (e.g. no approximate-prefix producer wired up) → fail closed (412).
+
+This behavior differs from earlier releases, where the gate was a hard-coded check inside the director that forwarded on any cache hit and rejected only on a complete cache miss. The threshold is now in tokens (not blocks), is operator-configurable, and is opt-in via the plugin.
```
vMaroon: We should avoid temporal language as instructed in AGENTS.md.

Would be interesting to understand if your agent "willingly" disobeyed that instruction or it just lost it in context.

docs/disaggregation.md
```diff
@@ -413,6 +413,20 @@ The `prefix-based-pd-decider` plugin makes the disaggregation decision according
 - `nonCachedTokens`: Number of non-cached tokens that trigger disaggregation
   - If set to 0, disaggregation never occurs for any request
 
+**Conditional-decode 412 gate**
+
+When this plugin is configured, it also drives the EPP's RFC 7240 `Prefer: if-available` (conditional-decode) gate. The coordinator uses that header to mark a speculative early-decode attempt: forward to a decode worker only if its KV cache already covers the prompt, otherwise return HTTP 412 Precondition Failed so the coordinator restarts the pipeline at encode/prefill/decode.
```
vMaroon: Wonder if this gate can be described independently, with coordinator use as a note.

pkg/epp/framework/interface/requestcontrol/plugins.go
```diff
@@ -98,3 +98,12 @@ type PreAdmitter interface {
 	plugin.Plugin
 	PreAdmit(ctx context.Context, request *fwksched.InferenceRequest) error
 }
+
+// ConditionalDecodeDecider answers the RFC 7240 "Prefer: if-available" gate:
+// given the chosen decode endpoint, should the request be rejected with HTTP
+// 412 (because the local KV cache does not cover enough of the prompt) or
+// forwarded? At most one such plugin is consulted per director.
+type ConditionalDecodeDecider interface {
```
vMaroon, line 105: Description should be self-contained, and descriptive of what the module does.

pkg/epp/requestcontrol/director.go
dmitripikus: Fixed, Thank you!

docs/disaggregation.md
dmitripikus: @vMaroon , this is an interesting point.
In this case the agent didn't work according to the instruction.
The agent states that "it didn't deliberately set the rule aside,  but it also didn't consciously check it against the wording I was producing..."
We probably need to explicitly re-check/re-run against AGENTS.md

In any case, I re-wrote the comment not to contain temporal language

docs/disaggregation.md
dmitripikus: I re-wrote the comments to describe the gate independently and pull the coordinator behavior into a note. Thank you!

pkg/epp/framework/interface/requestcontrol/plugins.go
dmitripikus: I re-phrased the description.

Thank you!

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/README.md
```diff
@@ -170,20 +172,35 @@ plugins:
...
 2. Read `PrefixCacheMatchInfo` from the decode endpoint attributes.
 3. Compute uncached suffix length.
-4. Return true (disaggregate) if uncached tokens ≥ `nonCachedTokens`.
+4. **Disaggregation:** return true (disaggregate) if uncached tokens ≥ `nonCachedTokens`.
+5. **Conditional-decode gate:** return true (reject with 412) if uncached tokens ≥ `nonCachedTokens`.
 
 #### How It Works
 
-The prompt token count is `len(request.Body.TokenizedPrompt.TokenIDs)`, populated by a `token-producer` — auto-created with the tokenizer-free `estimate` backend when none is configured. Prefix cache state is read from the `PrefixCacheMatchInfo` attribute on the decode endpoint, populated by `approx-prefix-cache-producer`. If the attribute is absent or malformed, disaggregation is skipped. Setting `nonCachedTokens: 0` disables the decider entirely (always returns false).
+The prompt token count is `len(request.Body.TokenizedPrompt.TokenIDs)`, populated by a `token-producer` — auto-created with the tokenizer-free `estimate` backend when none is configured. Prefix cache state is read from the `PrefixCacheMatchInfo` attribute on the decode endpoint, populated by `approx-prefix-cache-producer`.
```
roytman, line 194: nit: there is some duplication here and in the `What it does`

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -93,58 +97,91 @@ func (d *PrefixBasedPDDecider) WithName(name string) *PrefixBasedPDDecider {
...
 func (d *PrefixBasedPDDecider) disaggregate(ctx context.Context, request *scheduling.InferenceRequest, endpoint scheduling.Endpoint) bool {
-	logger := log.FromContext(ctx)
 	debugLogger := log.FromContext(ctx).V(logging.DEBUG)
 
 	// NonCachedTokens defines the minimum number of non-cached tokens required
 	// to trigger disaggregated PD. A value of 0 disables disaggregation.
 	if d.config.NonCachedTokens == 0 {
 		return false
 	}
+	nonCachedTokens, ok := d.computeNonCachedTokens(ctx, request, endpoint)
```
roytman, line 107: We cannot see it here, due to conflicts, but this PR removes the short-circuit and now spams Error logs on the fail-closed/fail-open path.
In the original disaggregate, there was an early return:
```go
if d.config.PromptTokens > 0 && inputTokens < d.config.PromptTokens {
		debugLogger.Info("Input is shorter than the promptTokens, no disaggregated PD")
		return false
	}
```
before reading the endpoint's `PrefixCacheMatchInfo` attribute. The refactor moved the attribute read into `computeNonCachedTokens`, called before that comparison. Mathematically, the final decision is unchanged (`nonCachedTokens <= inputTokens`, so `if inputTokens < threshold` the result is still `false/no-reject`), but the side effect isn't: every request now triggers `endpoint.Get()` and, if the attribute is missing, `logger.Error(nil, "unable to read prefix cache state")` — regardless of whether the input was even long enough for the decision to matter.

This bites hardest on the new conditional-decode gate path: previously, a missing `PrefixCacheMatchInfo` attribute during the 412 check only produced a `V(DEBUG)` log (primaryEndpointHasCachedPrefix's debug-only logging). Now it produces an Error-level log for every single conditional-decode request in any deployment that declares prefix-based-pd-decider for the gate without wiring approx-prefix-cache-producer into the decode profile — a configuration the PR's own docs describe as normal, documented, fail-closed behavior (not an error). That's log-volume/alerting noise for an expected code path, and it's untested: no test combines a short input with a missing attribute, so nothing catches this.

Fix is small: keep the length-based short-circuit inside computeNonCachedTokens (or in each caller before calling it), so the endpoint read and its Error log only fire when the input is actually long enough for the threshold comparison to matter.

roytman (commented):
@dmitripikus , please resolve the conflicts.

nit: same fact stated three times
The `fail-open/fail-closed` rationale (disaggregation fails open, 412 gate fails closed) is written out in full in three places: `docs/disaggregation.md`, the new table in `disagg/README.md` and the doc comment on `ShouldRejectConditionalDecode` `prefix_based_pd_decider.go. Not a blocker, but worth trimming two of the three down to a pointer at review/merge time.

dmitripikus:
This PR is replaced by PR https://github.com/llm-d/llm-d-router/pull/2335 that contains new implementation of the feature (new impl is based on PR https://github.com/llm-d/llm-d-router/pull/2215)
Therefore this PR is closed.

PR #1689, opened 2026-06-18
fix(flowcontrol): call up() after down() in maxMinHeap.Remove

RishabhSaini:
## Summary

- `maxMinHeap.Remove` only calls `down()` after swapping the removed element
  with the last array element. When the replacement comes from a different
  subtree, ancestor-level violations go unfixed. Add the missing `up()` call.
- Add a white-box test with a 6-element heap that proves the corruption.

## Details

In a max-min heap, levels alternate between max and min. The subtree property
(each node is the max or min of its entire subtree) is NOT transitive across
alternating levels, so `down()` alone cannot fix violations with ancestors.

**Current production impact: none.** The only `Remove` caller (`dispatchItem`)
always removes the root (index 0) via `PeekHead`, where `down()` alone is
correct regardless of heap layout. However, `Remove` accepts an arbitrary
handle and its contract promises O(log n) arbitrary removal. Any future caller
(request cancellation, queue eviction) would hit silent heap corruption.

## Test plan

- [ ] New test `TestMaxMinHeap_Remove_CrossSubtreeSwap` fails without the fix,
      passes with it
- [ ] All existing queue conformance and property tests pass

### DIFF
--- a/pkg/epp/flowcontrol/framework/plugins/queue/maxminheap.go
+++ b/pkg/epp/flowcontrol/framework/plugins/queue/maxminheap.go
@@ -285,9 +285,12 @@ func (h *maxMinHeap) Remove(handle flowcontrol.QueueItemHandle) (flowcontrol.Que
 		delete(h.handles, handle)
 		h.byteSize.Add(^item.OriginalRequest().ByteSize() + 1) // Atomic subtraction
 
-		// The swapped item at index i might violate the heap property, so we must restore it
-		// by bubbling the item down the heap.
+		// The swapped item at index i might violate the heap property in either direction:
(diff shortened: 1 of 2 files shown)


LukeAVanDrie:
@RishabhSaini and I discussed this and are going ahead with #1690 which resolves the issue and moves the code to a more maintainable state.

shmuelk:
> @RishabhSaini and I discussed this and are going ahead with https://github.com/llm-d/llm-d-router/pull/1690 which resolves the issue and moves the code to a more maintainable state.

Can this PR be closed?

RishabhSaini:
in favor of #1690

PR #1690, opened 2026-06-18
refactor(flowcontrol): single-ended SafeQueue backed by container/heap

LukeAVanDrie:
**What type of PR is this?**

/kind cleanup

The flow-control `SafeQueue` was double-ended (`PeekHead` + `PeekTail`), backed by a hand-rolled max-min heap. @RishabhSaini identified a latent bug in that bespoke data structure (#1689). In practice nothing consumes `PeekTail`, the only reference was the `flowQueueAccessor` passthrough itself. Dispatch (EDF/SLO ordering) needs only the head,
and eviction uses an independent ordering, so a double-ended dispatch queue buys nothing.

This makes the queue single-ended and swaps the custom heap for the standard library:

- Remove `PeekTail` from `QueueInspectionMethods` and every implementer, mock, and test.
- Rename `PeekHead` -> `Peek` now that there is a single end.
- Replace `maxminheap.go` with `priorityqueue.go`, a `container/heap`-backed priority queue (`PriorityQueue`). This deletes the custom sift logic and the latent bug class with it, while keeping O(log n) targeted removal via per-item index tracking.
- Drop the internal handle map; `Remove` validates membership by index identity, preserving the existing error contract with less state.
- Rename the registered queue `MaxMinHeap` -> `PriorityQueue`.

`container/heap` with index tracking is already the established pattern in `flowcontrol/eviction/queue.go`, so this aligns the dispatch queue with existing style. FCFS / `ListQueue` behavior is unchanged. 

N/A, but a more aggressive refactoring that resolves the issue surfaced in #1689.

```release-note
flowcontrol: the priority-ordered dispatch queue is now backed by the standard library container/heap and registered as `PriorityQueue` (previously `MaxMinHeap`). Any configuration that explicitly set the queue to `MaxMinHeap` must be updated to `PriorityQueue`.
```


RishabhSaini (approved):
The stdlib swap eliminates the bug class entirely and improves maintainability. Code is correct, all invariants preserved.
Nit: the PR description says "N/A" for the issue link but references #1689. Consider linking it as "Fixes #1689" so it auto-closes.
/lgtm

pkg/epp/flowcontrol/framework/plugins/queue/priorityqueue.go
```diff
@@ -0,0 +1,246 @@
+/*
+Copyright 2025 The Kubernetes Authors.
```
shmuelk:
```suggestion
Copyright 2026 The llm-d Authors.
```

pkg/epp/flowcontrol/framework/plugins/queue/priorityqueue.go
```diff
@@ -0,0 +1,246 @@
...
+	"sync/atomic"
+
+	"github.com/llm-d/llm-d-router/pkg/epp/flowcontrol/contracts"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/flowcontrol"
+)
+
+// PriorityQueueName is the name of the priority queue implementation.
+//
+// This queue provides a concurrent-safe priority queue backed by the standard library's
+// container/heap. Items are ordered by the configured OrderingPolicy, with the highest-priority
```
shmuelk:
```suggestion
// This queue provides a concurrent-safe priority queue that implements the standard library's
// container/heap interface. Items are ordered by the configured OrderingPolicy, with the highest-priority
```

pkg/epp/flowcontrol/framework/plugins/queue/priorityqueue_test.go
```diff
@@ -0,0 +1,82 @@
+/*
+Copyright 2025 The Kubernetes Authors.
```
shmuelk:
```suggestion
Copyright 2026 The llm-d Authors.
```

pkg/epp/flowcontrol/framework/plugins/queue/priorityqueue.go
```diff
@@ -0,0 +1,246 @@
...
+		return nil
+	}
+	return pq.heap.items[0].item
+}
+
+// Add adds an item to the queue.
+// Time complexity: O(log n).
+func (pq *priorityQueue) Add(item flowcontrol.QueueItemAccessor) {
+	pq.mu.Lock()
+	defer pq.mu.Unlock()
```
shmuelk: The only stament that needs locking is 165, wouldn't it be better to surround that statement with the Lock and Unlock calls?

pkg/epp/flowcontrol/framework/plugins/queue/priorityqueue.go
```diff
@@ -0,0 +1,246 @@
...
+	item.SetHandle(hi)
+	heap.Push(pq.heap, hi)
+	pq.byteSize.Add(item.OriginalRequest().ByteSize())
+}
+
+// Remove removes an item from the queue.
+// Time complexity: O(log n).
+func (pq *priorityQueue) Remove(handle flowcontrol.QueueItemHandle) (flowcontrol.QueueItemAccessor, error) {
+	pq.mu.Lock()
+	defer pq.mu.Unlock()
```
shmuelk: The statements that need locking are 189 through 193, perhaps the Lock and Unlock calls should wrap those statements

pkg/epp/flowcontrol/framework/plugins/queue/priorityqueue.go
```diff
@@ -0,0 +1,246 @@
...
+	itemsToKeep := make([]*heapItem, 0, len(pq.heap.items))
+
+	for _, hi := range pq.heap.items {
+		if predicate(hi.item) {
+			removedItems = append(removedItems, hi.item)
+			hi.Invalidate()
+			hi.index = -1
+			pq.byteSize.Add(^hi.item.OriginalRequest().ByteSize() + 1) // Atomic subtraction.
+		} else {
+			itemsToKeep = append(itemsToKeep, hi)
```
shmuelk: As itemsToKeep is pre-allocated to be the size of the current heap, which will be more than sufficient even if no items are removed by the predicate. Why not have a counter and do simple assignments instead of calling append here?

pkg/epp/flowcontrol/framework/plugins/queue/priorityqueue.go
```diff
@@ -0,0 +1,246 @@
...
+
+	drainedItems := make([]flowcontrol.QueueItemAccessor, len(pq.heap.items))
+	for i, hi := range pq.heap.items {
+		drainedItems[i] = hi.item
+		hi.Invalidate()
+		hi.index = -1
+	}
+
+	pq.heap.items = make([]*heapItem, 0)
+	pq.byteSize.Store(0)
```
shmuelk, line 253:
```suggestion
	pq.mu.Lock()
	old := pq.heap.items
	pq.heap.items = make([]*heapItem, 0)
	pq.byteSize.Store(0)
	pq.mu.Unlock()

	drainedItems := make([]flowcontrol.QueueItemAccessor, len(old))
	for i, hi := range old {
		drainedItems[i] = hi.item
		hi.Invalidate()
		hi.index = -1
	}
```

pkg/epp/flowcontrol/framework/plugins/queue/priorityqueue.go
LukeAVanDrie: Done. These were carried over verbatim from the old maxminheap.go (originally from gateway-api-inference-extension); fixed both new files to the current header. I actually used `git mv` to rename the files, but the delta is large enough for it do reflect as an addition and deletion rather than modification.

pkg/epp/flowcontrol/framework/plugins/queue/priorityqueue.go
LukeAVanDrie: Done. The allocation + `SetHandle` happen before the item is shared, and `byteSize` is atomic, so the lock now covers only `heap.Push`.

pkg/epp/flowcontrol/framework/plugins/queue/priorityqueue.go
LukeAVanDrie: Partially applied. I hoisted the `nil` check and the type assertion out (they only inspect the caller's handle). But the `IsInvalidated()` check has to stay inside the lock: `isInvalidated` (and `index`) are written under this same mutex by `Remove`/`Cleanup`/`Drain`, so reading them lock-free would be a data race.

So the locked region is now: `IsInvalidated` → identity check → `heap.Remove` → invalidate.

pkg/epp/flowcontrol/framework/plugins/queue/priorityqueue.go
LukeAVanDrie: Nice call; I switched to in-place compaction which drops the second slice allocation entirely. Added a small tail-nil loop so the removed `*heapItem`s aren't retained by the vacated capacity.

pkg/epp/flowcontrol/framework/plugins/queue/priorityqueue.go
LukeAVanDrie: I'd like to push back on this one. Moving the loop outside the lock makes `hi.Invalidate()` / `hi.index = -1` race with a concurrent `Remove(handle)`. Even after the slice is swapped out, a caller can still hold a handle to one of the drained items and call `Remove`, which reads `hi.isInvalidated` and `hi.index` under the lock. Drain mutating those same fields lock-free is a data race. 

If we ever want to shorten this O(n) section, the clean way would be making `isInvalidated`/`index` atomic so invalidation can go lock-free, but that's a broader change I'd rather not fold into this PR.

LukeAVanDrie: Also, `Drain` is only ran on the graceful shutdown path. I don't think we need to be concerned with optimizing this method.

shmuelk: ok

shmuelk (approved):

PR #1696, opened 2026-06-19
fix(sidecar): allowlist selector parsing for GA InferencePool API

pierDipi:
The GA InferencePool API (inference.networking.k8s.io/v1) nests pod labels under spec.selector.matchLabels, while the deprecated alpha API (inference.networking.x-k8s.io/v1alpha2) uses a flat spec.selector map.

The allowlist validator read spec.selector unconditionally and fmt.Sprintf'd nested map values into label strings, producing invalid selectors like "map[app.kubernetes.io/name:...]" that Kubernetes rejected for exceeding the 63-character label value limit.

Branch the selector path based on the detected API group and use NestedStringMap for type-safe extraction.


zdtsw:
Do we want to take this as backport to next release or get it fixed in 0.9.0? @pierDipi thinking for midstream we can get it in immediately

pierDipi:
Next release is fine

shmuelk (approved):

Issue #955, opened 2026-05-01
Migrate e2e tests from GAIE


roytman:
After merging GAIE and this repository, we need to migrate GAIE tests.

ahg-g:
This is also to track merging the scripts, we have `hack/test-e2e.sh` and `test/scripts/run_e2e.sh`, right?

roytman:
Yes for `hack/test-e2e.sh`, it was migrated to `hack/test-e2e.sh`. There was no `test/scripts/run_e2e.sh`

ahg-g:
we have one now: https://github.com/llm-d/llm-d-router/blob/main/test/scripts/run_e2e.sh, should we merge them into one?

roytman:
yes, the scripts and the tests as well, this is the point of the issue 

> better e2e tests integration (e.g. directory structure), reuse the configuration, remove redundant/duplicated env variables and files(?)

I plan to return to that later, so I haven't closed the issue.

hexfusion:
@shmuelk made progress here can we close this @roytman

roytman:
Hi, @hexfusion , I think we can do more for integration, e.g., the test scripts that @ahg-g mentioned above

PR #1698, opened 2026-06-20
test: consolidate and rename e2e runner scripts

roytman:
**What type of PR is this?**
/kind test

## What

Consolidate shared logic between the two e2e runner scripts and rename them
(plus their Makefile targets) so the name states which suite each runs.

- Extract the duplicated interrupt-cleanup trap and label-filtered Ginkgo
  runner into `test/scripts/e2e-common.sh`, sourced by both runners.
- Move `hack/test-e2e.sh` to `test/scripts/`, co-located with the other runner.
- Rename for clarity:
  - `hack/test-e2e.sh`        -> `test/scripts/test-e2e-gaie.sh`   (GAIE suite)
  - `test/scripts/run_e2e.sh` -> `test/scripts/test-e2e-router.sh` (router suite)
- Rename the Makefile targets `test-e2e-scheduler{,-run}` to
  `test-e2e-router{,-run}`, aligning them with the existing `e2e-router` CI job
  and the "router" product name.
- Update the CI workflow to call `make test-e2e-router-run`.

## Why

The two scripts shared near-identical cleanup and test-invocation blocks, and
their names (`test-e2e.sh`, `run_e2e.sh`) gave no hint which suite each drives.
The Makefile also still calls the second suite "scheduler" while CI already
called it "router"; this unifies on "router".

## Behavior

No change to what the suites test or how CI shards them. The GAIE and router
e2e jobs run exactly as before; only file names, the shared helper, and the target
names changed.

Part of #955

```release-note
NONE
```


test/scripts/test-e2e-router.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+# Copyright 2026 The Kubernetes Authors.
```
shmuelk:
```suggestion
# Copyright 2026 The llm-d Authors.
```

test/scripts/e2e-common.sh
```diff
@@ -0,0 +1,42 @@
+# Copyright 2026 The Kubernetes Authors.
```
shmuelk:
```suggestion
# Copyright 2026 The llm-d Authors.
```

shmuelk:
Looks great. Just two minor changes to the copyright statements

roytman:
> Looks great. Just two minor changes to the copyright statements

@shmuelk , thank you for your review. I accepted your suggestions.

shmuelk (approved):

ahg-g:
The end goal is to merge into one e2e runner, right?

ahg-g (approved):

roytman:
> The end goal is to merge into one e2e runner, right?

Indeed, that was my original plan. I initially combined them into a single file using `if [ "$SUITE" = gaie ]` branching, but found a few challenges:

- The cluster-ownership logic, image sets, and env contracts genuinely differ. As a result, the merged script ended up being mostly two separate branches side-by-side, plus a few shared helpers.
- The Makefile uses two separate targets: `test-e2e-gaie-run` and `test-e2e-router-run`.
- In `ci-pr-checks.yaml`, two separate CI jobs (`e2e-gaie` and `e2e-router`) bind to those target names by string. They run as separate, parallel jobs with different inputs. Given how long these tests take to run, it is actually beneficial to keep them as separate jobs and Makefile targets.
- Anyone running the merged script manually would still have to explicitly choose which part (or both) to execute.

shmuelk:
> The end goal is to merge into one e2e runner, right?

See PR #1699 which merges the IGW tests into the llm-d-router tests. After that PR merges I will be removing the old IGW tests and combining the test utilities.

PR #1704, opened 2026-06-22
fix(sidecar): strip ec_transfer_params from decode request in EPD mode

### LINKED ISSUE #1703: `Got ec_transfer_params` warning in decode logs
[new_nixlv2_test.txt](https://github.com/user-attachments/files/29199897/new_nixlv2_test.txt)

**What happened**:

In EPD (Encoder-Prefill-Decode) mode, `ec_transfer_params` injected by the
encoder stage is forwarded to the decode pod in the NIXL v2 connector. The
decode pod has no EC connector, so vLLM logs a warning for every request:

    (EngineCore pid=357) WARNING [core.py:369] Got ec_transfer_params,
    but no ECConnector found. Disabling ECTransfer for this request.

Requests still succeed (200 OK), but the warning is noisy and incorrect --
the decode pod does not need ec_transfer_params.

The decoder should not get a requests with `ec_transfer_params`

Place the attached unit test in pkg/sidecar/proxy/connector_nixlv2_test.go and run it using:

```bash
make test-filter PATTERN=TestProxy TYPE=sidecar
```

**Environment**:
- llm-d-router version (use `git describe --tags --dirty --always` if you built from source, or specify the tag if you used a tagged version or image): gaie-1.5-sync-1518-gf61ab917-dirty
- Install tools:
- Others:

revit13:
**What type of PR is this?**
/kind bug

In EPD (Encoder-Prefill-Decode) mode, every decode request arrived at the
decode pod carrying `ec_transfer_params`, causing vLLM to log a warning per
request:

    (EngineCore) WARNING Got ec_transfer_params, but no ECConnector found.
    Disabling ECTransfer for this request.

Requests succeeded (200 OK), but the warning is incorrect — the decode pod
has no EC connector and does not use these params.

## Root cause

`handleNIXLV2` receives a request body that `runPDPipeline` already
serialized with `ec_transfer_params` present (injected by the encoder
stage). The same `completionRequest` map is mutated in place for both the
prefill and decode stages. Encoder-only fields were never removed before
marshaling the decode body.

## Fix

Delete `ec_transfer_params` from `completionRequest` when preparing the
decode stage, before setting `kv_transfer_params` and marshaling the decode
body (`connector_nixlv2.go`).

## Test

Added a Ginkgo integration test in `connector_nixlv2_test.go` that drives
the full EPD path (encoder backend + prefill mock + decode mock) and
asserts:
- prefill request carries `ec_transfer_params` (correct — prefiller needs it)
- decode request does not carry `ec_transfer_params` (the fix)

The test fails before the fix and passes after.

Fixes #1703

```release-note
NONE
```

### DIFF
--- a/pkg/sidecar/proxy/connector_nixlv2.go
+++ b/pkg/sidecar/proxy/connector_nixlv2.go
@@ -375,6 +375,7 @@ retryLoop:
 			}
 		}
 	}
+	delete(completionRequest, requestFieldECTransferParams)
 	completionRequest[requestFieldKVTransferParams] = pKVTransferParams
 
 	dbody, err := json.Marshal(completionRequest)
--- a/pkg/sidecar/proxy/connector_nixlv2_test.go
+++ b/pkg/sidecar/proxy/connector_nixlv2_test.go
@@ -1058,6 +1058,81 @@ var _ = Describe("NIXL Connector (v2)", func() {
 		Expect(testInfo.prefillHandler.GetCompletionHeaders()[0].Get(requestHeaderDataParallelRank)).To(BeEmpty())
 		Expect(testInfo.decodeHandler.GetCompletionHeaders()[0].Get(requestHeaderDataParallelRank)).To(BeEmpty())
 	})
+
+	// EC+PD regression: ec_transfer_params is injected into the prefill request
+	// by the encoder stage but must not appear in the decode request. The decode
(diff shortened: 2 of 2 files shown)


elevran:
@revit13 is this ongoing?

revit13:
This PR needs more testing. closing it for now. Thanks

PR #1721, opened 2026-06-23
feat(prefix): auto-tune prefix match cap from model max_model_len

### LINKED ISSUE #1505: Auto-tune maxPrefixBlocksToMatch

Change `approx-prefix-cache-producer` plugin to auto-tune maxPrefixBlocksToMatch when it is not explicitly set in EPP config.

The value it should be auto-tuned to should be max_model_len/(block size tokens)

**Why is this needed**:
https://github.com/llm-d/llm-d-router/issues/1217#issuecomment-4596100068

This will ensure that EPP checks upto the max context length of the model in the prefix matching logic. This will be one less configuration needed for large models. 

Required steps
1. resource usage of EPP with very large context models and prefix matching
2. runtime checks for memory pressure in EPP
3. Fetch the max_model_len from curl http://$ENDPOINT/v1/models

mayur-tolexo:
Closes #1505.

The prefix-cache producer caps matching at a fixed `maxPrefixTokensToMatch` (default 128k), which under-matches long prefixes on larger-context models and is one more thing to tune per model. Per the discussion on #1505, this auto-tunes it from the model's `max_model_len` when `autoTune` is on and the cap is left at its default. An explicitly set cap is still respected.

`max_model_len` isn't a metric, so it's read from `/v1/models` — fetched once per pod off the request path and cached, with concurrent misses collapsed to one fetch. Until it resolves, or if the pod can't be reached, we fall back to the current default, so nothing changes in the common case. The fetch reuses the `model-server-metrics` scheme/TLS flags, and cached entries are dropped when a pod leaves the active set.

Note: like the metrics scraper, the `/v1/models` request isn't authenticated — a follow-up if secured model servers are in scope.

```
$ go test -race -cover ./pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/...
ok  	github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix	1.917s	coverage: 85.0% of statements
```

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -499,6 +499,9 @@ func (r *Runner) registerInTreePlugins() {
 	// data layer models source/extractor
 	fwkplugin.Register(srcmodels.ModelsDataSourceType, srcmodels.ModelDataSourceFactory)
 	fwkplugin.Register(attrmodels.ModelsExtractorType, extmodels.ModelServerExtractorFactory)
+	// Default producer of the /v1/models attribute: fetches once per endpoint on
+	// add rather than polling, and self-wires its endpoint-notification source.
+	fwkplugin.RegisterAsDefaultProducer(extmodels.ModelsEndpointExtractorType, extmodels.ModelEndpointExtractorFactory, attrmodels.ModelsAttributeKey)
 
 	fwkplugin.Register(prefix.PrefixCacheScorerPluginType, prefix.PrefixCachePluginFactory)
 	fwkplugin.Register(maxscore.MaxScorePickerType, maxscore.MaxScorePickerFactory)
--- a/pkg/epp/framework/plugins/datalayer/attribute/models/data_types.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/models/data_types.go
@@ -37,6 +37,9 @@ type ModelDataCollection []ModelData
 type ModelData struct {
 	ID     string `json:"id"`
 	Parent string `json:"parent,omitempty"`
+	// MaxModelLen is the model's context window in tokens as reported by the
+	// model server; zero when the server does not report it.
+	MaxModelLen int `json:"max_model_len,omitempty"`
 }
 
 // String returns a string representation of the model info
--- a/pkg/epp/framework/plugins/datalayer/attribute/models/data_types_test.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/models/data_types_test.go
@@ -1,0 +1,38 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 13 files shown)


vMaroon:
Thank you for the contribution! I think the goal is very aligned with where we wanna take the project.

`max_model_len` is constant per-endpoint runtime info, so I think it fits better as an endpoint attribute than a fetch+cache living inside this plugin.

The `models-data-extractor` already parses `/v1/models` into a per-endpoint `ModelDataCollection` attribute - it just doesn't keep `max_model_len`, though it's right there in the same payload. We could surface the field on that attribute and read it where the cap is derived, falling back to the default when it's absent.

That makes the models source a dependency of the cap derivation, so it should be declared as one. `inflightload` is the precedent.

mayur-tolexo:
> Thank you for the contribution! I think the goal is very aligned with where we wanna take the project.
> 
> `max_model_len` is constant per-endpoint runtime info, so I think it fits better as an endpoint attribute than a fetch+cache living inside this plugin.
> 
> The `models-data-extractor` already parses `/v1/models` into a per-endpoint `ModelDataCollection` attribute - it just doesn't keep `max_model_len`, though it's right there in the same payload. We could surface the field on that attribute and read it where the cap is derived, falling back to the default when it's absent.
> 
> That makes the models source a dependency of the cap derivation, so it should be declared as one. `inflightload` is the precedent.

Thanks for the guidance - reworked it. max_model_len now rides the existing models attribute: added `MaxModelLen` to `ModelData`, and the prefix producer declares `ModelsAttributeKey` as a required dependency and reads it where the cap is derived, falling back to the default when absent. Dropped the standalone `fetch+cache`; PTAL.

vMaroon:
Ok after further thinking - I think we can just go right away for the proper solution of creating an `EndpointExtractor` that runs once on endpoint-add instead of polling constant information. This would be aligned with this effort:
- https://github.com/vllm-project/vllm/issues/38147

So it would be an `EndpointExtractor` on the `EndpointNotificationSource` fetching `/v1/models` once on endpoint-add and writing the same `ModelsAttributeKey`. Should also make it a dependency that self-wires as a default source.

The consumer stays as-is, only the source flips from the per-tick poll to a once-on-add fetch.

What do you think?

mayur-tolexo:
> Ok after further thinking - I think we can just go right away for the proper solution of creating an `EndpointExtractor` that runs once on endpoint-add instead of polling constant information. This would be aligned with this effort:
> 
> * [[RFC]: Endpoints to Retrieve User Configuration and Runtime Data vllm-project/vllm#38147](https://github.com/vllm-project/vllm/issues/38147)
> 
> So it would be an `EndpointExtractor` on the `EndpointNotificationSource` fetching `/v1/models` once on endpoint-add and writing the same `ModelsAttributeKey`. Should also make it a dependency that self-wires as a default source.
> 
> The consumer stays as-is, only the source flips from the per-tick poll to a once-on-add fetch.
> 
> What do you think?

+1, let's go with the extractor approach. Reading through #38147, will update the PR.

vMaroon:
To clarify, the attached vLLM issue would be where we're heading, for the scope of this PR we'd still be only looking at `/v1/models`.

elevran:
@mayur-tolexo @vMaroon - please close if not actively pursuing

PR #1727, opened 2026-06-23
burst prefix-cache affinity for prompt-sharing requests

### LINKED ISSUE #1726: Burst prefix-cache affinity
## The problem

Prefix-cache-aware routing scores each request against the cache state *left by past requests*. Within one batch window that signal is blind to the most relevant information available: the other prompts arriving in the same window.

Concretely: a burst of prompts that share a prefix all arrive before any of themhas warmed a cache. The prefix-cache scorer therefore returns the **same empty (zero) score on every replica for every prompt in the burst** - the score
reflects the past, not the cache the sibling prompts in this window are about to create. With identical scores the scorer cannot distinguish replicas, load balancing scatters the burst, and the shared prefix is prefilled redundantly on several replicas instead of once. The cache only becomes useful *after* the burst has scattered - exactly when it is too late to help it.

The root cause is that scoring is per-request against a stale snapshot. The fix is to consider the window jointly: look at the prompts arriving together and co-locate the ones that share a prefix, before routing, so the shared prefix is prefilled once and the rest reuse it.

This is a property of prefix-sharing bursts in general, spanning a spectrum of overlap. At the partial-overlap end: requests sharing a system prompt, RAG queries over a shared context, one prompt contained in a longer one, batched
fan-out, evaluation sweeps. The full-overlap special case is a burst of **identical** prompts - the shared prefix is the entire prompt - as produced by RL rollout / GRPO sampling (`G` samples per prompt) or any multi-sample decoding submitted as separate requests. The identical case is the easiest to see and to measure, but it is one point on the spectrum, not the whole problem.

It is **not** the same as duplicate-completion handling (e.g. vLLM's `n`): that only helps identical completions resolved inside a single engine, and does nothing for *distinct* prompts that merely share a prefix, or for samples a client submits as separate requests. The problem here is a routing/scoring gap across independent requests in a window, which `n` does not address.

## GRPO example 

Any prefix-sharing burst exhibits this; RL rollout is simply the instance we measured because the sharing is easy to quantify. Each training step dispatches a batch of prompts, each as several separate requests that share a prefix, all submitted at once - so the window sees many prompts that should co-locate and the scorer scores them all zero. The same scatter would occur for distinct prompts
sharing a system prompt or context; rollout just makes the effect clean to count.

Measured on the baseline (4 vLLM replicas, TP=1, Qwen3-4B, 81,936 rollout requests over 40 steps, routed via EPP):
- **Group co-location is poor.** Of the full 8-sample groups, ~86% scatter across 2-4 of the 4 replicas; only ~14% land fully on one replica.   Mean distinct replicas per group ~= 2.83.
- **The 8 samples of a group arrive almost simultaneously.** Reconstructing submission time as `ts - pick_s - gen_s`, the intra-group arrival spread is  p50 ~0.5 ms, p90 ~0.8 ms, p99 ~12.5 ms. verl fires all 8 as one batch.
- **Routing latency exceeds the arrival spread.** EPP `pick_s` is p50 ~72 ms,   so by the time the first sample of a group has finished routing, all 8 have  already been dispatched and scored against a cold cache. Any signal derived  from a completed routing decision is structurally too late for that group.
- **Consequence on cache efficiency.** Prefix-cache hit rate falls as replicas  grow (around 61% at 2 replicas down to around 50% at 4 replicas) - more  replicas means more scatter and more redundant prefill of the shared prompt.

So the shared prefix is prefilled on ~3 of 4 replicas per group instead of once, and the prefix-routing benefit is largely lost under bursty traffic.

Problem graphs (baseline, current prefix-cache scorer):

Group co-location - the cause: 8-sample groups scatter across replicas (only 13.6% land on one replica, mean 2.83 distinct replicas per group):

Image

Prefix-cache hit rate - the cost: the hit rate falls as replicas grow, ~51% at 4 replicas:

Image

## The idea

Within a short time window, treat the requests that arrive together as a batch and assign them **jointly, before routing**, using the request prompts themselves (not a post-routing cache index). Requests that share a prompt prefix are steered onto the same replica (or a bounded set of replicas), so the shared prefix is prefilled once and the rest of the group reuses it.

Three parameters control it:

- `windowDurationMs` (T): how long to batch arrivals. The data shows a group's samples arrive within ~1 ms (p90) to ~12.5 ms (p99), so a small window captures a group; T = 100 ms captures ~99.7% of groups with negligible added latency relative to multi-second generation.
- `maxPerReplica` (k): cap on how many samples of one group go to a single replica. `k = -1` (unlimited) sends a whole group to one replica; because there are far more groups than replicas (256 groups vs 4 replicas per step), whole-group placement still balances load across replicas at the group level while prefilling each prompt once. A finite `k` (e.g. 2) spreads a group across a bounded set of replicas when decode parallelism matters more.
- `minColocateBlocks`: minimum shared leading blocks for two *distinct* prompts to co-locate, and the minimum a single (non-plicated) request must share with another to gain an affinity at all. 0 (default) disables both - placement is purely load-balanced and only identical prompts co-locate, the behavior measured below. A value `> 0` co-locates prompts that share a long preamble (RAG context, system prompt, multi-turn history) or one prompt contained in another, even when each such prompt is unique. It is only a floor (don't bother co-locating on a trivial 1-block match); over-clustering is prevented structurally by the fair-share cap, not by this threshold, so co-location is balance-safe at any value.

Co-locating identical samples does not serialize decode (vLLM continuous-batches the concurrent sequences) and does not raise per-replica KV pressure (the same total samples per replica, now sharing prefix blocks instead of recomputingthem).

ezrasilvera:
### Summary

A burst of prefix-sharing prompts arriving in one window all score zero on the prefix-cache scorer (no cache is warm yet), so load balancing scatters them and the shared prefix is prefilled redundantly. This adds a request-level data producer that, within a batch window, assigns prompt-sharing requests jointly and before routing - co-locating those that share a prefix so it is prefilled once. It covers identical-prompt bursts (e.g. RL rollout samples) and distinct prompts that share a prefix (shared system prompt, RAG context, one prompt contained in another), bounded by a per-replica fair-share cap so co-location never unbalances the cluster.

Fixes: #1726 

### What changed

Two commits:

1. **refactor(prefix): extract block-hashing into shared prefixhash package** - behavior-preserving move of the prompt block-hashing (`GetBlockHashes`, `HashBlock`, chained block hash) out of `approximateprefix` into a new `dataproducer/prefixhash` package, so multiple prefix producers derive identical block hashes. `approximateprefix` keeps a `type blockHash = prefixhash.BlockHash` alias; no behavior change.

2. **feat(prefix): add burst prefix cache producer for grouped requests** - new `dataproducer/burstprefix` package implementing a `DataProducer`:
   - **Batch-wait**: requests arriving within `windowDurationMs` are collected; when the window seals they are assigned jointly. Processing the batch under a lock removes the per-request race that makes near-simultaneous arrivals unschedulable together.
   - **Grouping**: requests are grouped by identical prompt prefix (the chained block-hash sequence) - this decides which requests are samples of one prompt, not where they land. An identical group of more than one member is always a placement unit. A single request becomes a unit only when `minColocateBlocks > 0` and it shares at least that many leading blocks with another request in the batch (so one prompt contained in another, or many prompts sharing a system preamble, co-locate); a request overlapping no other, and any prefix-less request, is scored zero everywhere so other scorers decide.
   - **Assignment**: within a group, samples fill one replica up to `maxPerReplica` (k) before spilling to the next least-loaded replica; `k = -1` places the whole group on one replica. Identical groups are placed first and kept whole - same-prompt co-location is the firm structure and is never broken by an attaching singleton - then prefix-sharing units attach to it. A unit prefers a replica that already holds a unit sharing at least `minColocateBlocks` leading blocks *and* is still under its fair share of the batch (total placed samples / replicas); otherwise placement is load-balanced. Units are placed longest-prefix first so shorter units match against the richest index. The fair-share cap is what makes co-location balance-safe: many units sharing one prefix spread across replicas up to their share rather than stampeding onto the seed replica. `0` keeps placement to identical groups, purely load-balanced.
   - **Scoring**: emits `PrefixCacheMatchInfo` (full match on the assigned replica, zero elsewhere) and reuses the existing `prefix-cache-scorer` unchanged - point its `prefixMatchInfoProducerName` at this producer. This is the same producer/scorer split already used by the precise and approximate prefix paths.

Also: registration in `cmd/epp/runner/runner.go`, a sample config `deploy/config/epp-burst-prefix-cache-config.yaml`, package READMEs, and the producer index table entry.

### How to enable

Wire `token-producer -> burst-prefix-cache-producer -> prefix-cache-scorer` (see the sample config). Key parameters:

| parameter | default | meaning |
|---|---|---|
| `windowDurationMs` | 100 | batch window T in ms |
| `maxPerReplica` | -1 | per-replica cap k; -1 = whole group to one replica |
| `blockSizeTokens` | 64 | token block size for prefix hashing |
| `maxPrefixTokensToMatch` | 0 | cap on matched prefix tokens; 0 uses the default block cap |
| `minColocateBlocks` | 0 | min shared leading blocks for inter-prompt co-location and for a single request to gain an affinity; 0 = identical groups only (load-balanced) |

The scorer reads from the producer by name:

```yaml
- type: prefix-cache-scorer
  parameters:
    prefixMatchInfoProducerName: burst-prefix-cache-producer
```

### Testing

- Unit tests for the assignment core: an unlimited group co-locates whole; `k` spreads evenly and fills before spilling; a lone group of one and empty-prefix requests get no affinity; distinct groups spread across replicas. With `minColocateBlocks` set: prefix-sharing families co-locate within their fair share; many groups sharing one prefix spread across replicas (no stampede); a shared prefix below the threshold does not co-locate; a single request overlapping a group attaches to it while the identical group stays whole; a lone request overlapping nothing gets no affinity; and overlapping singletons co-locate with no identical groups present (the shared-context case).
- Integration test for batch-wait `Produce`: a concurrent burst of identical prompts all co-locate on one replica; a lone request gets no affinity.
- Race detector clean. `make presubmit` passes (format, lint, typos, govulncheck, tags).

### Results (measured)

Measured on a real verl + llm-d integration: llm-d running in no-kube (local) mode, with EPP serving as the vLLM endpoint picker for verl's rollout requests. The only variable between the two arms is the burst producer; same GRPO/GSM8K rollout workload (4 vLLM replicas, TP=1, Qwen3-4B, group sampling G=8, 40 steps).

- **baseline** - EPP with the current prefix-cache scorer.
- **Burst-producer** - same, plus the burst prefix-cache producer (`windowDurationMs=100`, `maxPerReplica=-1`).

| metric | baseline | Burst-producer |
|---|---|---|
| 8-sample groups fully co-located | 13.6% | 100.0% |
| mean distinct replicas per group | 2.83 | 1.00 |
| prefix-cache hit rate | 51.2% | 68.3% |
| mean rollout gen time / step (s) | 21.0 | 16.4 |

Every full group co-locates onto one replica (mean 1.00 distinct replicas, down from 2.83), the prefix-cache hit rate rises 51.2% -> 68.3%, and mean rollout gen time per step drops 21.0 s -> 16.4 s (~22%). Latency does **not** regress - it improves, and most of the gain is in the tail: the per-step series flattens from spikes near 45 s to a steady ~13-24 s band. Co-locating a group does not serialize decode (vLLM continuous-batches the concurrent sequences) and removing redundant prefill frees headroom.

These timing numbers are a floor, not a ceiling. This workload is decode-dominant with very short prompts (~55-88 tokens, about one block), so prefill is a small fraction of end-to-end time and the latency win here mostly reflects freed prefill headroom rather than the prefill saving itself. As prompts grow - reasoning traces, long system prompts, RAG context - prefill becomes a far larger share of the work, so prefilling a shared prefix once instead of on every replica should yield substantially larger timing wins. The co-location and hit-rate gains shown here are the same mechanism; their latency payoff scales with prompt length.

Group co-location - groups go from 13.6% on a single replica (mean 2.83) to 100% (mean 1.00):

result_gen_time_per_step

Prefix-cache hit rate - rises 51.2% -> 68.3%:

result_group_colocation

Mean rollout gen time per step - drops 21.0 s -> 16.4 s, tail spikes flattened:

result_prefix_cache_hitrate

### DIFF
--- a/deploy/config/epp-burst-prefix-cache-config.yaml
+++ b/deploy/config/epp-burst-prefix-cache-config.yaml
@@ -18,6 +18,7 @@ plugins:
     parameters:
       windowDurationMs: 100       # batch window T
       maxPerReplica: -1           # k; -1 sends all samples of a group to one replica
+      minColocateBlocks: 0        # >0 co-locates distinct groups sharing >= N leading blocks; 0 = load-balanced
   - type: prefix-cache-scorer
     parameters:
       prefixMatchInfoProducerName: burst-prefix-cache-producer
--- a/deploy/config/epp-burst-prefix-cache-config.yaml
+++ b/deploy/config/epp-burst-prefix-cache-config.yaml
@@ -11,7 +11,7 @@ plugins:
     parameters:
       modelName: hf-repo/model-name           # set to the model used in the vLLM deployment
       vllm:
-        url: http://localhost:8000
+        url: http://<vllm-service>.<namespace>.svc.cluster.local:8000
   - type: single-profile-handler
   - type: decode-filter
   - type: burst-prefix-cache-producer
--- a/deploy/config/epp-burst-prefix-cache-config.yaml
+++ b/deploy/config/epp-burst-prefix-cache-config.yaml
@@ -1,0 +1,34 @@
+# Sample EPP config: burst prefix cache pipeline
+#   token-producer -> burst-prefix-cache-producer -> prefix-cache-scorer
+#
+# Co-locates bursts of prompt-sharing requests (e.g. RL rollout group samples)
+# onto shared replicas so a shared prefix is prefilled once instead of scattered
+# across replicas on a cold cache.
+apiVersion: llm-d.ai/v1alpha1
+kind: EndpointPickerConfig
+plugins:
+  - type: token-producer
+    parameters:
+      modelName: hf-repo/model-name           # set to the model used in the vLLM deployment
+      vllm:
+        url: http://localhost:8000
+  - type: single-profile-handler
+  - type: decode-filter
+  - type: burst-prefix-cache-producer
+    parameters:
+      windowDurationMs: 100       # batch window T
+      maxPerReplica: -1           # k; -1 sends all samples of a group to one replica
+  - type: prefix-cache-scorer
+    parameters:
+      prefixMatchInfoProducerName: burst-prefix-cache-producer
+  - type: load-aware-scorer
+  - type: max-score-picker
+schedulingProfiles:
+  - name: default
+    plugins:
+      - pluginRef: decode-filter
+      - pluginRef: prefix-cache-scorer
+        weight: 2.0
+      - pluginRef: load-aware-scorer
+        weight: 1.0
+      - pluginRef: max-score-picker
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
@@ -17,7 +17,8 @@ limitations under the License.
 package burstprefix
 
 import (
-	"sort"
+	"cmp"
+	"slices"
 	"strings"
 
 	fwksched "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
@@ -167,14 +168,17 @@ func assign(entries []*entry, k, minColocateBlocks int) {
 
 	shared := sharedPrefixKeys(groups, order, minColocateBlocks)
 	var groupUnits, singleUnits []string
+	counts := map[string]int{} // prefix-block count per placed unit, computed once
 	total := 0
 	for _, key := range order {
 		switch {
 		case len(groups[key]) >= 2:
 			groupUnits = append(groupUnits, key)
+			counts[key] = totalBlocks(groups[key][0].hashes)
 			total += len(groups[key])
 		case shared[key]:
 			singleUnits = append(singleUnits, key)
+			counts[key] = totalBlocks(groups[key][0].hashes)
 			total += len(groups[key])
 		}
 	}
@@ -184,8 +188,8 @@ func assign(entries []*entry, k, minColocateBlocks int) {
 
 	// Longest-prefix first within each tier: a longer prefix seeds more blocks in
 	// the index, so shorter units match against the richest set of placed blocks.
-	sortByPrefixLen(groupUnits, groups)
-	sortByPrefixLen(singleUnits, groups)
+	sortByPrefixLen(groupUnits, counts)
+	sortByPrefixLen(singleUnits, counts)
 
 	maxShare := (total + len(replicas) - 1) / len(replicas) // ceil: equal samples per replica
 
@@ -197,9 +201,9 @@ func assign(entries []*entry, k, minColocateBlocks int) {
 }
 
 // sortByPrefixLen orders keys by descending prefix-block count (stable).
-func sortByPrefixLen(keys []string, groups map[string][]*entry) {
-	sort.SliceStable(keys, func(i, j int) bool {
-		return totalBlocks(groups[keys[i]][0].hashes) > totalBlocks(groups[keys[j]][0].hashes)
+func sortByPrefixLen(keys []string, counts map[string]int) {
+	slices.SortStableFunc(keys, func(a, b string) int {
+		return cmp.Compare(counts[b], counts[a])
 	})
 }
 
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
@@ -18,6 +18,7 @@ package burstprefix
 
 import (
 	"encoding/binary"
+	"sort"
 	"strings"
 
 	fwksched "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
@@ -53,14 +54,68 @@ func groupKey(hashes [][]prefixhash.BlockHash) string {
 	return b.String()
 }
 
-// assign steers each batched request to a replica so samples sharing a prompt
-// co-locate. Requests are grouped by identical prefix; only groups with more
-// than one member receive an affinity (singletons have no reuse to exploit and
-// are left for other scorers). Within a group, samples fill one replica up to
-// maxPerReplica (k) before spilling to the next least-loaded replica; k == -1
-// places the whole group on one replica. Group placement is balanced by the
-// batch-wide assignment count so distinct groups spread across replicas.
-func assign(entries []*entry, k int) {
+// batchIndex records which replicas hold each prefix block, for matching later
+// groups against groups already placed in this batch. Because a whole prompt
+// prefix is added at once, holding block i implies holding blocks 0..i, so a
+// greedy walk that stops at the first unheld block yields the longest contiguous
+// shared prefix per replica (the approximateprefix matchLongestPrefix property).
+type batchIndex struct {
+	holders map[prefixhash.BlockHash]map[string]struct{}
+}
+
+func newBatchIndex() *batchIndex {
+	return &batchIndex{holders: map[prefixhash.BlockHash]map[string]struct{}{}}
+}
+
+// add records every block of hashes as held by replica.
+func (b *batchIndex) add(hashes [][]prefixhash.BlockHash, replica string) {
+	for _, ph := range hashes {
+		for _, h := range ph {
+			s := b.holders[h]
+			if s == nil {
+				s = map[string]struct{}{}
+				b.holders[h] = s
+			}
+			s[replica] = struct{}{}
+		}
+	}
+}
+
+// longestPrefix returns, per replica, the number of leading prefix blocks that
+// replica already holds (summed across prompts) - the shared-prefix length.
+func (b *batchIndex) longestPrefix(hashes [][]prefixhash.BlockHash) map[string]int {
+	res := map[string]int{}
+	for _, ph := range hashes {
+		for _, h := range ph {
+			holders := b.holders[h]
+			if len(holders) == 0 {
+				break
+			}
+			for name := range holders {
+				res[name]++
+			}
+		}
+	}
+	return res
+}
+
+// assign steers each batched request to a replica so prompt-sharing samples
+// co-locate. Pass 1 groups requests by identical prefix (samples of one prompt);
+// only groups with more than one member receive an affinity (singletons have no
+// reuse and are left for other scorers).
+//
+// Pass 2 places the groups jointly over the whole batch. Groups are placed
+// longest-prefix first so they seed the batch index for shorter groups to match
+// against. Each group prefers a replica that already holds a group sharing at
+// least minColocateBlocks leading blocks AND is still below its fair share of the
+// batch (total placed samples / replicas); this prefills a long shared prefix
+// once across groups without letting many prefix-sharing groups stampede onto one
+// replica. With no eligible match a group seeds the least-loaded replica. Within
+// a group, samples fill one replica up to maxPerReplica (k) before spilling to
+// the next least-loaded replica; k == -1 places the whole group on one replica.
+// minColocateBlocks == 0 disables inter-group co-location (placement is purely
+// load-balanced).
+func assign(entries []*entry, k, minColocateBlocks int) {
 	groups := map[string][]*entry{}
 	order := []string{}
 	for _, e := range entries {
@@ -74,33 +129,59 @@ func assign(entries []*entry, k int) {
 		groups[key] = append(groups[key], e)
 	}
 
-	// load is the batch-wide count of samples assigned per replica, used to
-	// spread groups across replicas.
-	load := map[string]int{}
+	// Placeable groups (more than one member) and the total samples they
+	// contribute, which sets the per-replica fair-share cap.
+	var placeKeys []string
+	total := 0
+	var replicas []fwksched.Endpoint
 	for _, key := range order {
 		members := groups[key]
 		if len(members) < 2 {
 			continue
 		}
-		assignGroup(members, members[0].pods, k, load)
+		placeKeys = append(placeKeys, key)
+		total += len(members)
+		if replicas == nil {
+			replicas = members[0].pods
+		}
+	}
+	if len(placeKeys) == 0 || len(replicas) == 0 {
+		return
+	}
+
+	// Longest-prefix first: a longer group seeds more blocks in the index, so
+	// shorter groups match against the richest set of already-placed prefixes.
+	sort.SliceStable(placeKeys, func(i, j int) bool {
+		return totalBlocks(groups[placeKeys[i]][0].hashes) > totalBlocks(groups[placeKeys[j]][0].hashes)
+	})
+
+	maxShare := (total + len(replicas) - 1) / len(replicas) // ceil: equal samples per replica
+
+	idx := newBatchIndex()
+	load := map[string]int{} // batch-wide samples assigned per replica
+	for _, key := range placeKeys {
+		placeGroup(groups[key], replicas, k, minColocateBlocks, maxShare, idx, load)
 	}
 }
 
-// assignGroup places the members of one group across replicas honoring the
-// per-replica cap k and updating the batch-wide load counter.
-func assignGroup(members []*entry, replicas []fwksched.Endpoint, k int, load map[string]int) {
+// placeGroup places one identical-prompt group. The first (primary) replica is
+// chosen with the inter-group prefix preference bounded by the fair-share cap;
+// remaining samples (when k caps the group) spill to least-loaded replicas. The
+// group's blocks are then recorded so later groups can match against it.
+func placeGroup(members []*entry, replicas []fwksched.Endpoint, k, minColocateBlocks, maxShare int, idx *batchIndex, load map[string]int) {
 	if len(replicas) == 0 {
 		return
 	}
-	perReplica := map[string]int{} // samples of THIS group already on a replica
+	hashes := members[0].hashes // identical group: any member represents it
+	matches := idx.longestPrefix(hashes)
 
+	perReplica := map[string]int{} // samples of THIS group already on a replica
+	preferMatch := minColocateBlocks > 0
 	i := 0
 	for i < len(members) {
-		target := pickReplica(replicas, perReplica, k, load)
+		target := pickReplica(replicas, perReplica, k, load, matches, minColocateBlocks, maxShare, preferMatch)
 		name := target.GetMetadata().NamespacedName.String()
 
-		// Fill this replica up to its remaining capacity before moving on, so a
-		// group concentrates rather than scatters.
 		run := len(members) - i
 		if k != unlimitedPerReplica {
 			capLeft := k - perReplica[name]
@@ -117,14 +198,54 @@ func assignGroup(members []*entry, replicas []fwksched.Endpoint, k int, load map
 		perReplica[name] += run
 		load[name] += run
 		i += run
+		preferMatch = false // only the primary replica gets the prefix preference
+	}
+
+	for _, m := range members {
+		if m.assigned != nil {
+			idx.add(hashes, m.assigned.GetMetadata().NamespacedName.String())
+		}
 	}
 }
 
-// pickReplica returns the least batch-loaded replica that still has capacity for
-// this group. When every replica is at the cap (more members than k*replicas)
-// it falls back to the least-loaded replica ignoring the cap. Ties break by
-// running requests then by name for deterministic placement.
-func pickReplica(replicas []fwksched.Endpoint, perReplica map[string]int, k int, load map[string]int) fwksched.Endpoint {
+// pickReplica chooses the target replica for the next run of a group. When
+// preferMatch is set it first tries the replica sharing the longest prefix (at
+// least minColocateBlocks blocks) that still has per-group capacity and is below
+// its fair share of the batch; otherwise, or when no such replica exists, it
+// falls back to the least batch-loaded replica. The fair-share bound is what
+// keeps many prefix-sharing groups from stampeding onto a single replica.
+func pickReplica(replicas []fwksched.Endpoint, perReplica map[string]int, k int, load, matches map[string]int, minColocateBlocks, maxShare int, preferMatch bool) fwksched.Endpoint {
+	if preferMatch {
+		var best fwksched.Endpoint
+		var bestName string
+		bestMatch := 0
+		for _, r := range replicas {
+			name := r.GetMetadata().NamespacedName.String()
+			if k != unlimitedPerReplica && perReplica[name] >= k {
+				continue
+			}
+			if load[name] >= maxShare {
+				continue // at fair share: co-locating here would unbalance the batch
+			}
+			m := matches[name]
+			if m < minColocateBlocks {
+				continue
+			}
+			if best == nil || m > bestMatch || (m == bestMatch && less(r, name, load, best, bestName)) {
+				best, bestName, bestMatch = r, name, m
+			}
+		}
+		if best != nil {
+			return best
+		}
+	}
+	return pickLeastLoaded(replicas, perReplica, k, load)
+}
+
+// pickLeastLoaded returns the least batch-loaded replica with capacity for this
+// group, falling back to the overall least-loaded replica when all are at the
+// cap. Ties break by running requests then by name for deterministic placement.
+func pickLeastLoaded(replicas []fwksched.Endpoint, perReplica map[string]int, k int, load map[string]int) fwksched.Endpoint {
 	var best fwksched.Endpoint
 	var bestName string
 	for _, r := range replicas {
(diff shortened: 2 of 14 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
```diff
@@ -0,0 +1,337 @@
...
+
+// longestPrefix returns, per replica, the number of leading prefix blocks that
+// replica already holds (summed across prompts) - the shared-prefix length.
+func (b *batchIndex) longestPrefix(hashes [][]prefixhash.BlockHash) map[string]int {
+	res := map[string]int{}
+	for _, ph := range hashes {
+		for _, h := range ph {
+			holders := b.holders[h]
+			if len(holders) == 0 {
+				break
```
elevran: The `break` exits only the **inner** `for _, h := range ph` loop, not the **outer** loop over prompts. After a block miss in prompt N, the outer loop continues into prompt N+1 and keeps crediting those blocks -- a replica missing block X mid-prompt will have its score inflated by blocks it does not hold. 
I suspect this is not what you intended. If so, Use a labelled break.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
```diff
@@ -0,0 +1,337 @@
...
+	for _, ph := range hashes {
+		total += len(ph)
+	}
+	return total
+}
+
+// encodePrefix serializes the first n prefix blocks. Equal results imply an
+// equal n-block leading prefix: each block hash chains its predecessor, so an
+// equal leading hash sequence means an equal prompt prefix.
+func encodePrefix(hashes [][]prefixhash.BlockHash, n int) string {
```
elevran, line 40: The `|` separator after each prompt's blocks makes keys for the same leading N blocks differ when prompts are split differently across `PerPromptTokens` entries. For `minColocateBlocks > 0`, two requests sharing the same first N blocks but with different prompt splits produce different signatures and fail to co-locate.

Block hashes already encode prompt boundaries (each prompt starts a fresh chain from the model seed), so the separator is not needed for correctness. Removing it makes the key stable across different prompt splits. Which behavior makes more sense in your use case?

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
```diff
@@ -0,0 +1,337 @@
...
+	for _, e := range entries {
+		key := groupKey(e.hashes)
+		if key == "" {
+			continue
+		}
+		if _, ok := groups[key]; !ok {
+			order = append(order, key)
+		}
+		groups[key] = append(groups[key], e)
+		if replicas == nil {
```
elevran, line 161: Q: Are there acses where the endpoint set can change mid-window or are you guaranteed working on a defined snapshot until window is done?

Asking since `replicas` is captured from the first entry with a non-empty group key; all other entries' `pods` fields are ignored. If the endpoint set changes mid-window (rolling update, scale event), entries may have `e.assigned` pointing to a pod absent from their own `pods` slice. The match loop in `Produce` then silently finds no match and produces zero affinity with no error, no log - probably not what you intended.

If this is a possibility, suggest logging when `e.pods` differs from the captured `replicas`.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/plugin.go
```diff
@@ -0,0 +1,188 @@
...
+		b := p.batch
+		time.AfterFunc(p.window, func() { p.seal(b) })
+	}
+	b := p.batch
+	b.entries = append(b.entries, e)
+	p.mu.Unlock()
+
+	select {
+	case <-b.sealed:
+	case <-ctx.Done():
```
elevran, line 153: `b.entries` has no size limit. Under a sustained burst or a long window one batch can accumulate unboundedly many entries, each holding a `hashes` slice and a `pods` slice. Consider a `maxBatchSize` config parameter (e.g. default 1000) that causes `Produce` to return an error immediately once the cap is reached.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/plugin.go
```diff
@@ -0,0 +1,188 @@
...
+
+// Produce collects the request into the current batch window, waits for the
+// window to seal, then attaches PrefixCacheMatchInfo reflecting the joint
+// assignment: a full match on the assigned replica and zero elsewhere.
+func (p *dataProducer) Produce(ctx context.Context, request *fwksched.InferenceRequest, pods []fwksched.Endpoint) error {
+	hashes := prefixhash.GetBlockHashes(ctx, request, p.config.BlockSizeTokens, p.maxBlocks)
+	e := &entry{hashes: hashes, pods: pods}
+
+	p.mu.Lock()
+	if p.batch == nil {
```
elevran, line 138: nit: There is no per-request visibility into batch-window wait time. A request that opens the window waits ~`windowDurationMs`; one that arrives just before seal waits nearly nothing. A `logger.V(logging.DEBUG).Info("batch sealed", "wait_ms", ...)` or a histogram metric would let operators observe actual wait times and tune `windowDurationMs`. This is especially needed if window duration can be set reasonably high compared with other processing times in the system.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/plugin.go
```diff
@@ -0,0 +1,188 @@
...
+
+	maxBlocks := defaultMaxPrefixBlocks
+	if cfg.MaxPrefixTokensToMatch > 0 {
+		maxBlocks = cfg.MaxPrefixTokensToMatch / cfg.BlockSizeTokens
+	}
+
+	return &dataProducer{
+		typedName: plugin.TypedName{Type: PluginType, Name: name},
+		config:    cfg,
+		window:    time.Duration(cfg.WindowDurationMs) * time.Millisecond,
```
elevran, line 124: nit: When `MaxPrefixTokensToMatch > 0` but less than `BlockSizeTokens`, integer division yields `maxBlocks = 0`, causing every `groupKey` to return `""` and the plugin to silently become a no-op. A check at construction time (`if cfg.MaxPrefixTokensToMatch > 0 && cfg.MaxPrefixTokensToMatch < cfg.BlockSizeTokens`) would catch this early.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/plugin_test.go
```diff
@@ -0,0 +1,105 @@
...
+	p, err := newDataProducer(context.Background(), "burst", config{WindowDurationMs: 50, MaxPerReplica: unlimitedPerReplica, BlockSizeTokens: 4})
+	require.NoError(t, err)
+
+	endpoints := []fwksched.Endpoint{testEndpoint("pod1"), testEndpoint("pod2")}
+	err = p.Produce(context.Background(), tokenizedRequest([]uint32{1, 2, 3, 4}), endpoints)
+	require.NoError(t, err)
+
+	assert.Empty(t, p.assignedReplica(t, endpoints),
+		"a lone request in a window shares no prompt and must not be steered")
+}
```
elevran, line 105: Two batch-lifecycle paths are not covered:

1. Context cancelled before the batch seals -- the `ctx.Done()` arm in `Produce` is untested. Add a test that passes an already-cancelled context and asserts `Produce` returns a non-nil error and no endpoint has a non-zero match.

2. Second batch window after the first seals -- the transition where `seal` sets `p.batch = nil` and the next request starts a fresh batch is not exercised.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign_test.go
```diff
@@ -0,0 +1,250 @@
...
+	entries := concat(x1, x2, y1, y2)
+
+	assign(entries, unlimitedPerReplica, 3)
+
+	// The shared prefix falls short of the threshold, so groups are placed purely
+	// by load and a family is not kept together.
+	assert.NotEqual(t, assignedName(x1[0]), assignedName(x2[0]),
+		"a shared prefix below minColocateBlocks must not co-locate")
+	assert.Equal(t, map[string]int{"pod1": 4, "pod2": 4}, counts(entries))
+}
```
elevran, line 250: Two edge cases in the assignment logic are not covered:

1. Single replica -- all entries must land on it regardless of fair-share or load logic. A bug in the `len(replicas) == 1` path would be invisible.

2. All replicas at `maxShare` before the group is fully placed -- the `capLeft = 1` overflow guard at `placeGroup` line 82 (within the file) is never exercised.

deploy/config/epp-burst-prefix-cache-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+# Co-locates bursts of prompt-sharing requests (e.g. RL rollout group samples)
+# onto shared replicas so a shared prefix is prefilled once instead of scattered
+# across replicas on a cold cache.
+apiVersion: llm-d.ai/v1alpha1
+kind: EndpointPickerConfig
+plugins:
+  - type: token-producer
+    parameters:
+      modelName: hf-repo/model-name           # set to the model used in the vLLM deployment
+      vllm:
```
elevran, line 13: `http://localhost:8000` in a published deployment sample will silently route tokenization calls to a local port when copied verbatim. Use a DNS placeholder instead:

```suggestion
        url: http://<vllm-service>.<namespace>.svc.cluster.local:8000
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
```diff
@@ -0,0 +1,337 @@
...
+	idx := newBatchIndex()
+	load := map[string]int{} // batch-wide samples assigned per replica
+	for _, key := range append(groupUnits, singleUnits...) {
+		placeGroup(groups[key], replicas, k, minColocateBlocks, maxShare, idx, load)
+	}
+}
+
+// sortByPrefixLen orders keys by descending prefix-block count (stable).
+func sortByPrefixLen(keys []string, groups map[string][]*entry) {
+	sort.SliceStable(keys, func(i, j int) bool {
```
elevran: nit: Two issues with `sortByPrefixLen`:

1. `sort.SliceStable` is the older API; the rest of the codebase uses `slices.SortStableFunc` (e.g. `picker/maxscore/picker.go`).

2. `totalBlocks(groups[keys[i]][0].hashes)` is called on every comparator invocation -- O(N log N) redundant sums for N groups. Precompute a `map[string]int` of block counts before both sort calls.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
```diff
@@ -0,0 +1,337 @@
...
+func encodePrefix(hashes [][]prefixhash.BlockHash, n int) string {
+	var b strings.Builder
+	var buf [8]byte
+	count := 0
+	for _, ph := range hashes {
+		for _, h := range ph {
+			if count >= n {
+				return b.String()
+			}
+			binary.LittleEndian.PutUint64(buf[:], uint64(h))
```
elevran: nit: `binary.LittleEndian.PutUint64(buf[:], uint64(h))` duplicates the unexported `toBytes` in the `prefixhash` package introduced by this PR (`prefixhash/hashing.go`). Exporting it as `ToBytes` and calling it here removes the duplication.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
```diff
@@ -0,0 +1,337 @@
...
+	// Longest-prefix first within each tier: a longer prefix seeds more blocks in
+	// the index, so shorter units match against the richest set of placed blocks.
+	sortByPrefixLen(groupUnits, groups)
+	sortByPrefixLen(singleUnits, groups)
+
+	maxShare := (total + len(replicas) - 1) / len(replicas) // ceil: equal samples per replica
+
+	idx := newBatchIndex()
+	load := map[string]int{} // batch-wide samples assigned per replica
+	for _, key := range append(groupUnits, singleUnits...) {
```
elevran: nit: `append(groupUnits, singleUnits...)` allocates a new backing array solely to range over it once. Two sequential `for _, key := range` loops avoid the allocation.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
```diff
@@ -0,0 +1,337 @@
...
+		if best == nil || less(r, name, load, best, bestName) {
+			best, bestName = r, name
+		}
+	}
+	return best
+}
+
+// less reports whether replica a should be preferred over replica b: fewer
+// assigned samples first, then fewer running requests, then lower name.
+func less(a fwksched.Endpoint, aName string, load map[string]int, b fwksched.Endpoint, bName string) bool {
```
elevran, line 331: nit: `less` breaks ties on `RunningRequestsSize`, but the `load-aware-scorer` that runs after `Produce` scores on `WaitingQueueSize`. The two signals can pull in opposite directions. Either align the tiebreaker with the scorer's signal, add a comment explaining why a different metric is intentional here, or document their intended use together.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/plugin.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// ProduceTimeout extends the producer timeout to cover the batch window.
+func (p *dataProducer) ProduceTimeout() time.Duration {
+	return p.window + produceTimeoutMargin
+}
+
+// newDataProducer initializes a new burst prefix cache producer.
+func newDataProducer(_ context.Context, name string, cfg config) (*dataProducer, error) {
+	if cfg.WindowDurationMs <= 0 {
```
elevran, line 91: `windowDurationMs` is only validated to be `> 0`. A misconfigured large value can block every request for an arbitrary duration. Add a reasonable upper bound (milliseconds?) so misconfiguration fails at startup rather than silently degrading every request. You can cap to a value and produce a log if exceeding it. Consider adding proper documentation on the README to inform users on reasonable values and the existance of an upper limit.

elevran (commented):
Thanks - the burst producer addresses a real problem in RL workoads and the measured results are compelling. 
Some correctness issues need addressing before merge, plus several gaps in the batch lifecycle tests. All comments inlined.

A;so - there is no linked GitHub issue. Per project conventions, non-trivial work must be tracked in an issue before merge. Please follow project guidelines on future PRs.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
ezrasilvera: Fixed in 2496eed2.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
ezrasilvera: Fixed in 771bca01. Removed the separator; note that per-prompt hash re-seeding (each prompt starts a fresh chain from the model seed), not the separator, is what governs cross-split matching, so keys are now stable across different prompt splits.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
ezrasilvera: Fixed in 771bca01. Added prefixhash.PutBlockHash, a shared helper that writes into a caller-provided buffer, so the duplication is removed without the per-block allocation that exporting toBytes would have introduced.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
ezrasilvera: Fixed in cc34a617. Produce now emits a DEBUG log when the assigned replica is not among the request's own pods (endpoint set changed mid-window), instead of degrading to zero affinity silently.

ezrasilvera:
> Thanks - the burst producer addresses a real problem in RL workoads and the measured results are compelling. Some correctness issues need addressing before merge, plus several gaps in the batch lifecycle tests. All comments inlined.
> 
> A;so - there is no linked GitHub issue. Per project conventions, non-trivial work must be tracked in an issue before merge. Please follow project guidelines on future PRs.

There is a git issue but it wasn't linked. I just linked it #1726

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/plugin.go
ezrasilvera: Fixed in 1d1956fb. Added a maxBatchSize config (default 1000, -1 = unlimited), validated at construction; Produce now returns an error once a window reaches the cap instead of accumulating entries unbounded.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/plugin.go
ezrasilvera: Fixed in 60222a58. Each entry now records its enqueue time and Produce logs wait_ms at DEBUG once the window seals, so operators can observe actual wait times and tune windowDurationMs.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/plugin.go
ezrasilvera: Fixed in f01d0e6a. newDataProducer now rejects maxPrefixTokensToMatch > 0 && < blockSizeTokens at construction (it would otherwise yield maxBlocks=0 and a silent no-op).

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/plugin.go
ezrasilvera: Fixed in f01d0e6a. Added an upper bound (maxWindowDurationMs = 10000): windowDurationMs above it is rejected at construction rather than stalling every request. README now documents the 1..10000 range and advises keeping the window small relative to other processing times.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
ezrasilvera: Fixed in 4ead43b6. Block counts are now computed once per placed unit (in the classification loop) and stored in a map; sortByPrefixLen switched to slices.SortStableFunc, matching the codebase idiom.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
ezrasilvera: Fixed in aef2bcd9. Replaced the append(...) with two sequential range loops (groupUnits then singleUnits), preserving order with no throwaway allocation.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
ezrasilvera: Fixed in d6f8f3bd. Added a comment on less clarifying that batch-local assigned load is the primary signal and running requests only break ties; the difference from the load-aware-scorer's WaitingQueueSize is intentional (placement balances work committed in this window, not downstream queue depth).

deploy/config/epp-burst-prefix-cache-config.yaml
ezrasilvera: Fixed in 0b01d36c. Replaced http://localhost:8000 with the cluster-DNS placeholder http://<vllm-service>.<namespace>.svc.cluster.local:8000.

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/plugin_test.go
ezrasilvera: Fixed in 29398dfd. Added TestProduce_CancelledContextBeforeSeal (already-cancelled context returns an error and yields no non-zero match) and TestProduce_SecondWindowAfterSeal (asserts seal resets p.batch to nil and a fresh window co-locates independently).

pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign_test.go
ezrasilvera: Fixed in 75a1f418. Added TestAssign_SingleReplicaPlacesEverything (distinct groups all land on the sole replica) and TestAssign_OverflowGuardBalancesBeyondCap (group of 8, k=2, 2 replicas: exercises the capLeft=1 overflow guard and asserts even rebalancing).

ezrasilvera:
@elevran Thanks for the detailed review. I address all issues you raised

elevran (approved):

PR #1734, opened 2026-06-24
Added an example of how to extend EPP with a plugin and build a custom image that runs it

ahg-g:
**What type of PR is this?**
/kind documentation

Adds a working example of how to extend EPP with a plugin and build a custom image.

Fixes #

```release-note
NONE
```

### DIFF
--- a/examples/customscorer/Dockerfile
+++ b/examples/customscorer/Dockerfile
@@ -1,0 +1,23 @@
+# Dockerfile demonstrating how to build a custom Endpoint Picker (EPP)
+# container image with an out-of-tree custom scorer plugin compiled in.
+
+FROM golang:1.25 as builder
+
+WORKDIR /workspace
+
+# Copy go.mod and go.sum
+COPY go.mod go.sum ./
+RUN go mod download
+
+# Copy the entire codebase
+COPY . .
+
+# Build the custom EPP binary containing the out-of-tree scorer plugin
+RUN CGO_ENABLED=0 GOOS=linux go build -a -o custom-epp ./examples/customscorer/cmd/main.go
+
+FROM gcr.io/distroless/static:nonroot
+WORKDIR /
+COPY --from=builder /workspace/custom-epp .
+USER 65532:65532
+
+ENTRYPOINT ["/custom-epp"]
--- a/examples/customscorer/cmd/main.go
+++ b/examples/customscorer/cmd/main.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package main demonstrates how to build an out-of-tree custom scorer plugin
+// and compile it into a custom Endpoint Picker binary and container image.
+package main
+
+import (
+	"os"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+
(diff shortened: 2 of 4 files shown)


vMaroon:
Overall looks good, can you add a short README describing the example / walking the user through with pointers?

tessapham:
cc @evacchi Your example Wasm plugin could be added as another way to extend EPP?

examples/customscorer/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Dockerfile demonstrating how to build a custom Endpoint Picker (EPP)
+# container image with an out-of-tree custom scorer plugin compiled in.
+
+FROM golang:1.25 as builder
```
elevran, line 4: nit: The base image is `golang:1.25` but go.mod declares go 1.26.5. With the default `GOTOOLCHAIN=auto` this silently downloads the 1.26.5 toolchain at build time. Suggest matching go.mod:

```suggestion
FROM golang:1.26 AS builder
```
(Uppercasing `AS`, as done in other docker files, also silences the buildkit FromAsCasing warning)

examples/customscorer/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# Copy go.mod and go.sum
+COPY go.mod go.sum ./
+RUN go mod download
+
+# Copy the entire codebase
+COPY . .
+
+# Build the custom EPP binary containing the out-of-tree scorer plugin
+RUN CGO_ENABLED=0 GOOS=linux go build -a -o custom-epp ./examples/customscorer/cmd/main.go
```
elevran, line 16: nit: build the package directory rather than the single file, so the example keeps building if the package later gains additional files:

```suggestion
RUN CGO_ENABLED=0 GOOS=linux go build -a -o custom-epp ./examples/customscorer/cmd
```

examples/customscorer/cmd/main.go
```diff
@@ -0,0 +1,46 @@
...
+	"os"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/llm-d/llm-d-router/cmd/epp/runner"
+	"github.com/llm-d/llm-d-router/examples/customscorer/plugins"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+)
+
+func init() {
```
elevran, line 31: nit: might be worth a comment (here or in a README for the example), to explain plugin registration via init() instead of using `fwkplugin.Register()` in main.

examples/customscorer/plugins/scorer.go
```diff
@@ -0,0 +1,101 @@
...
+	return s.typedName
+}
+
+// Category returns the scorer category.
+func (s *FixedScorer) Category() scheduling.ScorerCategory {
+	return scheduling.Distribution
+}
+
+// Consumes returns data dependencies. FixedScorer requires no endpoint attributes.
+func (s *FixedScorer) Consumes() plugin.DataDependencies {
```
elevran, line 86: Q: is this the idiomatic way for declaring no dependencies? 
If the intent is to show DAG processing, it would make sense to declare a real dependency that consumes data. 
If not, just do not implement `Consumes()` at all.
See #836 and #1039

examples/customscorer/plugins/scorer_test.go
```diff
@@ -0,0 +1,75 @@
...
+	require.NoError(t, err)
+	require.NotNil(t, created)
+
+	scorer, ok := created.(*FixedScorer)
+	require.True(t, ok)
+	assert.Equal(t, 0.75, scorer.score)
+}
+
+func TestFactoryInvalidParameters(t *testing.T) {
+	dec := plugin.StrictDecoder([]byte(`{"score": 1.5}`))
```
elevran, line 52: nit: may want to show a couple of validation use cases organized as table driven tests to steer future implementations towards table driven tests?

elevran (commented):
One gap for code whose purpose is teaching: there is no `README` or usage doc.
A short `examples/customscorer/README.md` showing the docker build invocation and a sample `EndpointPickerConfig` that references the plugin would improve self-serve for users.

elevran:
@ahg-g not needed anymore?

PR #1738, opened 2026-06-24
fix: clear KV block index entries on endpoint removal

### LINKED ISSUE #1736: Clear KV block index entries when endpoint is removed
When an endpoint is removed (pod restart, scaling down, container crash), `RemoveSubscriber` cancels the ZMQ connection but does not clear the pod's entries from the KV block index. This leaves stale entries that reference blocks no longer present on the pod.

The `Extract` method in the precise-prefix-cache producer already receives `EventDelete` for endpoint removal. Adding an `Index.Clear` call on this path would clean up stale entries immediately.

Related: https://github.com/llm-d/llm-d-kv-cache/issues/372

RishabhSaini:
## Summary
- Call `Index.Clear` on `EventDelete` to remove stale entries for pods that are no longer running
- Uses meta.Address (pod IP) to match the PodIdentifier stored in the index by the ZMQ event pipeline
- Updates test helper to provide a fake indexer instead of nil

## Context
When a vLLM pod restarts or is removed, RemoveSubscriber cancels the ZMQ connection but leaves stale index entries. These cause the router to preferentially route to a pod that no longer has the expected KV cache, degrading routing quality.

## Known limitation
There is a small race window where Pool workers may still be processing events queued before RemoveSubscriber cancelled the subscriber. These in-flight events can re-add a few entries after Index.Clear. 

Fixes #1736

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
@@ -52,6 +52,12 @@ func (p *Producer) Extract(ctx context.Context, event fwkdl.EndpointEvent) error
 		logger.V(logging.DEBUG).Info("Adding subscriber", "endpoint", endpointKey)
 	case fwkdl.EventDelete:
 		p.subscribersManager.RemoveSubscriber(ctx, endpointKey)
+		if meta.Address != "" {
+			if err := p.kvCacheIndexer.KVBlockIndex().Clear(ctx, meta.Address); err != nil {
+				logger.Error(err, "Failed to clear index entries for removed endpoint",
+					"endpoint", endpointKey, "address", meta.Address)
+			}
+		}
 		logger.V(logging.DEBUG).Info("Removed KV-events subscriber", "endpoint", endpointKey)
 	}
 	return nil
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
@@ -53,7 +53,8 @@ func (p *Producer) Extract(ctx context.Context, event fwkdl.EndpointEvent) error
 	case fwkdl.EventDelete:
 		p.subscribersManager.RemoveSubscriber(ctx, endpointKey)
 		if meta.Address != "" {
-			if err := p.kvCacheIndexer.KVBlockIndex().Clear(ctx, meta.Address); err != nil {
+			if err := p.kvCacheIndexer.KVBlockIndex().Clear(ctx, fmt.Sprintf("%s:%s", meta.Address, meta.Port)); err != nil {
+
 				logger.Error(err, "Failed to clear index entries for removed endpoint",
 					"endpoint", endpointKey, "address", meta.Address)
 			}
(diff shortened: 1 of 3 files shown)


shmuelk:
What about cases of vLLM's Data Parallel feature in which there are multiple model servers on one pod?

If the pod goes down, all of the endpoints go away, your algorithm takes care of that.

What happens if somehow one endpoint on the pod goes away and not the whole pod?

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
```diff
@@ -52,6 +52,12 @@ func (p *Producer) Extract(ctx context.Context, event fwkdl.EndpointEvent) error
 		logger.V(logging.DEBUG).Info("Adding subscriber", "endpoint", endpointKey)
 	case fwkdl.EventDelete:
 		p.subscribersManager.RemoveSubscriber(ctx, endpointKey)
+		if meta.Address != "" {
+			if err := p.kvCacheIndexer.KVBlockIndex().Clear(ctx, meta.Address); err != nil {
```
bongwoobak: Clear matches PodIdentifier by exact equality, but the index is keyed as "<Address>:<Port>" everywhere else. meta.Address alone matches nothing, so this clears no entries.

```suggestion
			if err := p.kvCacheIndexer.KVBlockIndex().Clear(ctx, fmt.Sprintf("%s:%s", meta.Address, meta.Port)); err != nil {

```

bongwoobak: @shmuelk this also covers your DP question

Keyed by `address:port` removing one rank clears only that rank's entries and a full-pod teardown fires one `EventDelete` per rank .

So all of them get cleared.

RishabhSaini: agreed, updated. thanks for the review

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
```diff
@@ -52,6 +52,13 @@ func (p *Producer) Extract(ctx context.Context, event fwkdl.EndpointEvent) error
 		logger.V(logging.DEBUG).Info("Adding subscriber", "endpoint", endpointKey)
 	case fwkdl.EventDelete:
 		p.subscribersManager.RemoveSubscriber(ctx, endpointKey)
+		if meta.Address != "" {
+			if err := p.kvCacheIndexer.KVBlockIndex().Clear(ctx, fmt.Sprintf("%s:%s", meta.Address, meta.Port)); err != nil {
+
```
vMaroon:
```suggestion
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
```diff
@@ -52,6 +52,13 @@ func (p *Producer) Extract(ctx context.Context, event fwkdl.EndpointEvent) error
...
+				logger.Error(err, "Failed to clear index entries for removed endpoint",
+					"endpoint", endpointKey, "address", meta.Address)
```
vMaroon:
```suggestion
				logger.Error(err, "Failed to clear index entries for removed endpoint",
					"endpoint", endpointKey, "address", meta.Address, "port", meta.Port)
```

vMaroon (approved):

PR #1739, opened 2026-06-24
fix(inflightload): prevent in-flight counter underflow on endpoint flap

### LINKED ISSUE #1950: [Bug :bug:]:  inflight-load-producer: active request counter can go negative on endpoint flap
### Contact Details

_No response_

## Summary

The `inflight-load-producer`'s per-endpoint active **request** counter can go
negative when an endpoint leaves and rejoins the endpoint set while requests are
in flight.

## Details

The counter is a pure +1 / -1 ledger:

- `+1` in `PreRequest` (`requestTracker.inc`) — creates the counter if absent
- `-1` in `OnEvicted` (`decIfPresent`) — only if present, fires at most once per
  request (atomic `Swap` guard + `LoadAndDelete`)
- counter dropped entirely in `DeleteEndpoint`, on an endpoint `EventDelete`

On a single counter instance the `+1` always precedes its matching `-1` and each
request decrements at most once, so the counter stays >= 0. It goes negative only
when the counter instance that received the `+1` is replaced before the `-1`
lands:

1. Request A routed to endpoint E -> `requests[E] = 1`.
2. E leaves the endpoint set (readiness flap NotReady->Ready, rollout, scale) ->
   `EventDelete` drops `requests[E]`.
3. E rejoins with the same `NamespacedName` and new traffic arrives -> request B's
   `PreRequest` recreates `requests[E] = 1`.
4. A completes -> its `-1` lands on the new counter (releases are keyed only by
   endpoint-ID string, so the pre-flap and post-flap E are indistinguishable) ->
   `requests[E] = 0`; B completes -> `-1` -> **-1**.

If E is deleted and not re-added, the stale `-1` is skipped (no negative); the
delete-then-re-add is required.

## Impact

A negative active-request count for an endpoint feeds downstream load/TTFT
estimation with an invalid value.

## Reproduction

Driving a single endpoint through add -> request A -> `EventDelete` ->
`EventAddOrUpdate` (same `NamespacedName`) -> request B -> drain A, B lands the
counter at -1.

## Affected code

`pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go`
(`requestTracker.inc`, `OnEvicted`/`decIfPresent`, `DeleteEndpoint`)

### Version

v0.6.0

### Relevant log output

```shell

```

LukeAVanDrie:

Per-endpoint in-flight request and token counters in the inflight-load producer are a +1/-1 ledger keyed by `NamespacedName`. The increment happens in `PreRequest`; the matching decrement on release looked the counter up again by endpoint ID. When an endpoint flaps (goes `NotReady` so its `EventDelete` drops the counter, then rejoins under the same `NamespacedName` and new traffic recreates it), a still-in-flight request's decrement landed on the recreated counter instead of the one it incremented, so the counter could fall below zero. This showed up as negative pool saturation in the concurrency detector, and also makes the affected endpoint look permanently unsaturated in the scheduling filter. The effect was larger when a pod crashed while holding many in-flight requests.

The existing `registeredEndpoints` pointer-identity guard only drops out-of-order delete events for an already-replaced endpoint object, so it did not cover the normal flap sequence.

Thanks @asharkhan3101, @kaushikmitr, and @RishabhSaini for the root cause analysis!

Changes:
- inc/add return the exact `*atomic.Int64` they mutate. `PreRequest` stores those pointers on the per-request entry, and release decrements the captured instance directly, so a release after a flap lands on the orphaned counter and leaves the live counter correct.
- Remove the `addIfPresent`/`decIfPresent` lookup-by-ID guard, which is no longer needed.
- Add `decrementClamped`, a CAS helper that floors at zero, following the existing `predictedlatency.decrementEndpointCounter` pattern, as defense in depth.
- Add tests for the flap and crash-with-load cases, plus coverage for the stale-delete guard.

Fixes [llm-d/issues/1950](https://github.com/llm-d/llm-d/issues/1950)

```release-note
Fix per-endpoint in-flight load counters going negative when an endpoint became NotReady and rejoined, which produced negative pool saturation and could skew scheduling toward the affected endpoint.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer_test.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer_test.go
@@ -312,64 +312,93 @@ func TestInFlightLoadProducer_DumpStateCapsEndpoints(t *testing.T) {
 	require.Equal(t, int64(104), state.Endpoints[0].Requests)
 }
 
-// TestInFlightLoadProducer_FlapDoesNotUnderflow asserts that an in-flight request's release lands
-// on the exact counter instance it incremented, even after the endpoint is deleted and recreated
-// under the same NamespacedName. A release from a request that predates the flap hits the orphaned
-// counter, leaving the live counter accurate and never negative.
+// TestInFlightLoadProducer_FlapDoesNotUnderflow asserts that a request's release lands on the exact
+// counter instance it incremented, even after the endpoint is deleted and recreated under the same
+// NamespacedName. A release that predates the flap hits the orphaned counter, leaving the live
+// counter accurate and never negative. Cases cover the request and token counters across both the
+// EndOfStream eviction path (OnEvicted) and the StartOfStream early-release path (releaseTokensEarly).
 func TestInFlightLoadProducer_FlapDoesNotUnderflow(t *testing.T) {
 	t.Parallel()
 
-	producer := newTestProducer(t)
-	ctx := context.Background()
-	endpointName := "flap-endpoint"
-	endpointID := fullEndpointName(endpointName)
+	requests := func(p *InFlightLoadProducer, id string) int64 { return p.requestTracker.get(id) }
+	tokens := func(p *InFlightLoadProducer, id string) int64 { return p.tokenTracker.get(id) }
 
-	// E joins the endpoint set. The same Endpoint object is reused for the
-	// matching delete so the registeredEndpoints stale-delete guard allows
-	// cleanup (mirrors the datalayer's same-pointer add/delete contract).
-	epA := newStubSchedulingEndpoint(endpointName)
-	require.NoError(t, producer.Extract(ctx, datalayer.EndpointEvent{
-		Type:     datalayer.EventAddOrUpdate,
-		Endpoint: epA,
-	}))
+	tests := []struct {
+		name                     string
+		addEstimatedOutputTokens bool
+		release                  requestcontrol.Response // chunk that triggers the release under test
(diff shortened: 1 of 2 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer_test.go
```diff
@@ -311,6 +311,168 @@ func TestInFlightLoadProducer_DumpStateCapsEndpoints(t *testing.T) {
 	require.Equal(t, int64(104), state.Endpoints[0].Requests)
 }
 
+// TestInFlightLoadProducer_FlapDoesNotUnderflow asserts that an in-flight request's release lands
+// on the exact counter instance it incremented, even after the endpoint is deleted and recreated
+// under the same NamespacedName. A release from a request that predates the flap hits the orphaned
+// counter, leaving the live counter accurate and never negative.
+func TestInFlightLoadProducer_FlapDoesNotUnderflow(t *testing.T) {
```
kaushikmitr, line 320: Code-wise the token counter gets the identical fix (captured tokenCounter + decrementClamped in both OnEvicted and the early-release path releaseTokensEarly). But the new tests only assert requestTracker, there's no flap coverage for the token counter, including the token-specific early-release path. Can you add a test for token counters as well

LukeAVanDrie: Done; also converted to a table-driven test.

ahg-g (approved):

PR #1747, opened 2026-06-25
fix(flowcontrol): resolve ManagedQueue eagerly and fix FlowKey.String

RishabhSaini:
## Summary

- **Resolve ManagedQueue in Phase 1 of `processAllQueuesConcurrently`** instead of deferring to Phase 3 workers. This eliminates a race where flow GC deletes a flow between collection and the worker's `ManagedQueue(key)` lookup, producing spurious error-level logs with full stack traces every GC cycle.
- **Fix `FlowKey.String()` for negative priorities**: `strconv.FormatUint(uint64(Priority))` rendered `-1` as `18446744073709551615` in logs. Changed to `strconv.Itoa`.

Both bugs were discovered and reproduced on a live cluster with flow control enabled (utilization-detector, `max-num-seqs=2`, multiple InferenceObjectives).

## Test plan

- [x] Regression test: `should resolve ManagedQueue eagerly so late GC does not skip work` — uses a phase-aware mock that succeeds during `IterateQueues` and fails after, so the test passes only with eager resolution
- [x] Verified test FAILs without the fix (detached HEAD at test-only commit)
- [x] Verified test PASSes with the fix
- [x] All 33 existing processor tests pass
- [x] All flow control package tests pass (`go test ./pkg/epp/flowcontrol/... -count=1`)
- [x] `go build ./...` and `go vet` clean

### DIFF
--- a/pkg/epp/flowcontrol/controller/internal/processor.go
+++ b/pkg/epp/flowcontrol/controller/internal/processor.go
@@ -536,9 +536,15 @@ func (sp *Processor) processAllQueuesConcurrently(
 ) {
 	logger := sp.logger.WithName(ctxName)
 
-	// Phase 1: Collect all queues to be processed into a single slice.
-	// This avoids holding locks on the shard while processing, and allows us to determine the optimal number of workers.
-	var queuesToProcess []flowcontrol.FlowQueueAccessor
+	type queueTask struct {
+		mq     contracts.ManagedQueue
+		logger logr.Logger
+	}
+
+	// Phase 1: Collect all queues and resolve ManagedQueue handles in one pass.
+	// Resolving here (instead of in the worker) eliminates a race where flow GC
+	// can delete a flow between collection and the worker's ManagedQueue lookup.
+	var work []queueTask
 	for _, priority := range sp.registry.AllOrderedPriorityLevels() {
 		band, err := sp.registry.PriorityBandAccessor(priority)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -536,56 +536,60 @@ func (sp *Processor) processAllQueuesConcurrently(
...
 		if err != nil {
 			logger.Error(err, "Failed to get PriorityBandAccessor", "priority", priority)
 			continue
 		}
 		band.IterateQueues(func(queue flowcontrol.FlowQueueAccessor) bool {
-			queuesToProcess = append(queuesToProcess, queue)
-			return true // Continue iterating.
+			key := queue.FlowKey()
+			mq, err := sp.registry.ManagedQueue(key)
+			if err != nil {
```
LukeAVanDrie, line 556: Non-blocking, but worth considering: this now swallows the error completely.

Dropping the error severity is the right move as the usual cause is a benign flow-GC race and the stack trace was just noise. But `ManagedQueue` can also fail with `ErrPriorityBandNotFound`, which would signal a real inconsistency (band map vs. `AllOrderedPriorityLevels`) rather than a routine race.

A debug-level line keeps that diagnosable without bringing back the spam:

```suggestion
			if err != nil {
				// Benign during normal operation: flow GC can delete a flow between the
				// IterateQueues snapshot and this lookup. Log at debug for diagnosability.
				logger.V(logutil.DEBUG).Info("Skipping queue; ManagedQueue no longer resolvable",
					"flowKey", key, "err", err)
				return true
			}
```

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -536,56 +536,60 @@ func (sp *Processor) processAllQueuesConcurrently(
...
-	// Phase 1: Collect all queues to be processed into a single slice.
-	// This avoids holding locks on the shard while processing, and allows us to determine the optimal number of workers.
-	var queuesToProcess []flowcontrol.FlowQueueAccessor
+	type queueTask struct {
+		mq     contracts.ManagedQueue
+		logger logr.Logger
+	}
+
+	// Phase 1: Collect all queues and resolve ManagedQueue handles in one pass.
+	// Resolving here (instead of in the worker) eliminates a race where flow GC
```
LukeAVanDrie: Minor wording nit: `IterateQueues` releases the registry `RLock` before invoking this callback, so the `ManagedQueue` lookup still runs outside the lock and a small window remains where GC can delete the flow between the snapshot and this call. The fix greatly narrows that window and makes the failure benign rather than eliminating it.

It might be worth saying "greatly narrows" so a future reader doesn't assume the lookup is race-free.

LukeAVanDrie (dismissed):

pkg/epp/flowcontrol/controller/internal/processor.go
RishabhSaini: agreed, added

LukeAVanDrie (approved):

PR #1751, opened 2026-06-26
fix(flowcontrol): return 503 (not 429) when no endpoints exist

### LINKED ISSUE #1735: No active endpoint should not be considered as full saturation
**What happened**: Pool Saturation returns 1 if there is no active endpoint, which makes it enqueue requests and drop them as 429 after the queue is full, and TTL requests are dropped as 503.

**What you expected to happen**: If there is no active endpoints, the upstream sees more 429s than 503 which can be confusing since it paints a picture of some availability instead of none.

**How to reproduce it (as minimally and precisely as possible)**: Use saturation detector and flow control with very low queue size, send request and you will observe a lot of 429s if all engine is scaled to 0. It should be instead 503 for all of the requests.

**Environment**:
- llm-d-router version (use `git describe --tags --dirty --always` if you built from source, or specify the tag if you used a tagged version or image): 0.9.1
- Install tools: N/A
- Others: N/A

LukeAVanDrie:
**What type of PR is this?**

/kind bug

When the inference pool has no active endpoints (e.g. all model-server replicas scaled to zero), the saturation detector intentionally reports full saturation (`1.0`) so flow control buffers requests as a scale-from-zero waiting room. The bug was in the terminal error code when a buffered request had to be dropped:

- A request that overflowed the queue while the pool was empty was finalized as `QueueOutcomeRejectedCapacity` → HTTP 429.
- A request that expired its TTL in the queue was finalized as `QueueOutcomeEvictedTTL` → HTTP 500.

So with zero endpoints, upstreams saw a confusing mix of 429s and 503s for the same "no backends" condition. The 429s are misleading. `429 Too Many Requests` implies the service is up and rate-limiting, inviting retries against a pool that has nothing to serve them.

The fix makes the error code encode availability rather than which buffer limit tripped:

- 429 (ResourceExhausted) — capacity exists but is contended (backpressure).
- 503 (ServiceUnavailable) — no serving capacity exists right now.

A new `QueueOutcomeRejectedNoEndpoints` outcome (wrapping `ErrNoEndpoints`) is emitted when the queue rejects at capacity and the pool is empty; the admission layer maps it to 503 with an `x-llm-d-request-dropped-reason: rejected-no-endpoints` header. The processor caches pool emptiness from the dispatch cycle (single Run goroutine, no locking) and consults it in the capacity-rejection path. The discriminator is endpoint count `== 0`, not `saturation >= 1.0`, since a non-empty but fully saturated pool also reports `1.0` and is legitimately 429.

Resulting behavior:

|                         | Capacity rejection | TTL eviction |
| ----------------------- | ------------------ | ------------ |
| Empty pool          | 503 (was 429)  | 503          |
| Non-empty saturated | 429                | 503          |

Scale-from-zero buffering is preserved; only the terminal error code changes, and only when the pool is empty. TTL eviction already returned 503 and is unchanged. Unit tests cover both the producer side (empty-pool rejection → `RejectedNoEndpoints`; non-empty → `RejectedCapacity`) and the translation side (new outcome → 503 + header).

Fixes #1735

```release-note
Flow control now returns 503 (Service Unavailable) instead of 429 (Too Many Requests) when a request is rejected at queue capacity while the inference pool has no endpoints (e.g. scaled to zero), no longer signalling false partial availability. Requests dropped this way carry the header `x-llm-d-request-dropped-reason: rejected-no-endpoints`.
```


mtparet:
Thanks!

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -280,6 +286,15 @@ func (sp *Processor) enqueue(item *FlowItem) {
 	// --- Capacity Check ---
 	// This check is safe because it is performed by the single-writer Run goroutine.
 	if ok, stats := sp.hasCapacity(key.Priority, req.ByteSize()); !ok {
+		// When the pool has no endpoints, the queue is acting as a scale-from-zero waiting room. A capacity rejection in
+		// that state reflects genuine unavailability (surfaced as 503), not backpressure against a contended pool (429).
+		if sp.poolEmpty {
+			sp.logger.V(logutil.DEBUG).Info("Rejecting request, queue at capacity with no endpoints",
+				"flowKey", key, "reqID", req.ID(), "reqByteSize", req.ByteSize())
+			item.FinalizeWithOutcome(types.QueueOutcomeRejectedNoEndpoints, fmt.Errorf("%w: %w",
```
ahg-g, line 294: Does this mean we shed all requests if the pool is empty? if so, this should not be the desired outcome, we need to queue requests of priorities above 0 as long as there is capacity to queue them

LukeAVanDrie: This is only for the case where:

1. no endpoints are ready
2. the buffers have filled up to their respective len / byte caps
3. new traffic comes in

Before, we were returning 429 for this specific case.

LukeAVanDrie: There's a matrix in the PR description. Capacity rejection only happens once the buffers are full.

ahg-g:
pls fix the unit test

LukeAVanDrie:
> pls fix the unit test

updated, I also added a new integration test which should directly illustrate your other question

ahg-g (approved):

Issue #1450, opened 2026-06-02
[RFC] Introduce request metadata parse point and explicit `agent-identity` attribute


elevran:
**What would you like to be added**:

1. A plugin point that allows request level attributes to be set as early as possible and used by the differnet subsystems. These should be limited to request metadata only (e.g., path, method, headers), not request body,
1. An explicit `agent-identity` attribute

**Why is this needed**:

#1244 added `agent-identity` based on request headers, This was added as a PreAddmission plugin. 
This conflates independent concepts
- PreAdmission was selected only so it runs before flow control (to allow all agent requests to be assigned to the same flow control identity). There is nothing in the code which is FlowControl or Admission related.
- a `agent-identity` might be usefull in other subsystems (e.g., program level optimizations in scheduling, kv cache control). Placing that into FairnessID makes it impossible to use as a reliable signal (the FairnessID can be set from unrelated values, such as JWT claims indicating the organizational gruup).

Not advocating for adding yet another Plugin extension point, might be a replacement of the PreAddmission introduced in #1244 (though, arguably, PreAdmission could be useful for active queue management...).
A metadata parse could also be used by for Session Affinity attribute extraction.

The flow (if something like this is implemented):
1. Parse headers to get agent identity, attach to request as request attribute (#1351)
2. Other plugins can extract session identity attribute for their use (eg., set intoFairness ID)

CC: @praveingk @D-Sai-Venkatesh @ahg-g @liu-cong @vMaroon

praveingk:
@elevran I think we wanted to implement something like this from the start.

The problem was FairnessID could not be written into by the program-aware plugin, once it is set in the InferenceRequest, and the queues already created.
Hence, we had to set the fairnessID directly based on the agent identity; instead of writing to a request attribute. I agree that, we will need agent-identity as a request metadata for identity-based affinity scoring.

Is it ok to have both ?
1) FarinessID set by the pre-admitter already..
2) Agent Identity request attribute additionally populated by the pre-admitter [agent identity](https://github.com/D-Sai-Venkatesh/llm-d-inference-scheduler/blob/b4a5c21d9d80eccdc8a12955d87c1babd092fb98/pkg/epp/framework/plugins/requestcontrol/preadmitter/agentidentity/agent_identity.go) plugin as well.

That would mean, we can add agent identity to the existing plugin.
Please let me know your thoughts. 

@D-Sai-Venkatesh Correct me if I am wrong here.

[...]

PR #1754, opened 2026-06-26
Extract agent-identity as a request metadata

praveingk:
Addresses https://github.com/llm-d/llm-d-router/issues/1450


pkg/epp/requestcontrol/director.go
```diff
@@ -59,6 +59,9 @@ const (
 	// producer overrides it by implementing requestcontrol.TimeoutAwareProducer.
 	dataProducerTimeout       = 400 * time.Millisecond
 	responseBodyQueueCapacity = 100
+	// AgentIdentityKey is the request attribute key used by the agent-identity
+	// header processor and read by the Director to derive FairnessID.
+	AgentIdentityKey = "agent-identity"
```
elevran: AgentIdentityKey is set to "agent-identity" here and AgentIdentityAttributeKey is set to "agent-identity" in the agentidentity plugin. These are two independent definitions of the same string. If they diverge, the Director silently falls through to DefaultFairnessID for every request. Define once and import. Since the Director already has implicit coupling to this concept, the cleanest fix is to import agentidentity.AgentIdentityAttributeKey directly, or move the constant to the framework interface package (pkg/epp/framework/interface/requestcontrol/) where it is neutral.
Also, AgentIdentityKey is exported but has no consumers outside the requestcontrol package (director.go and director_test.go are both in the same package). An unexported agentIdentityKey is sufficient.

praveingk: Thanks @elevran . Will address it.

praveingk: Addressed in https://github.com/llm-d/llm-d-router/pull/1754/commits/b3c59f3e013dc5db95c9158a2802a00518bdcb4d

praveingk:
@elevran I have addressed the comments. Can you please review.

pkg/epp/framework/interface/requestcontrol/plugins.go
```diff
@@ -92,9 +92,18 @@ type Admitter interface {
 	Admit(ctx context.Context, request *fwksched.InferenceRequest, pods []fwksched.Endpoint) error
 }
 
-// PreAdmitter runs after InferenceRequest creation but before admission control.
-// It can mutate InferenceRequest fields such as FairnessID and Headers.
-type PreAdmitter interface {
+// RequestHeaderProcessor runs after InferenceRequest creation but before admission control.
+// It processes request metadata (headers, path, method) to attach attributes to the request
```
elevran, line 96: Q: is comment accurate with respect to what is processed? I think only headers are available. Are _path_ and _method_ also availble on `InferenceReuqest`?

pkg/epp/framework/interface/requestcontrol/plugins.go
```diff
@@ -92,9 +92,18 @@ type Admitter interface {
...
-	PreAdmit(ctx context.Context, request *fwksched.InferenceRequest) error
+	RequestHeader(ctx context.Context, request *fwksched.InferenceRequest) error
 }
+
+const (
+	// AgentIdentityKey is the request-attribute key under which a
+	// RequestHeaderProcessor publishes the resolved agent identity.
+	// Downstream consumers such as the Director read it
+	// via scheduling.ReadRequestAttribute to derive the FairnessID.
+	AgentIdentityKey = "agent-identity"
```
elevran: nit: this key is specific to the agent-identity plugin but lives in the generic `requestcontrol` interface package. Consider moving it into the plugin package itself and having the Director import it from there. Keeps the framework package free of plugin-specific naming, and any future `RequestHeaderProcessor` plugin can colocate its own key with the producer.

pkg/epp/requestcontrol/director.go
```diff
@@ -276,9 +276,15 @@ func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestCo
 	ctx = log.IntoContext(ctx, logger)
 	logger.V(logutil.DEBUG).Info("LLM request assembled")
 
-	if err := d.runPreAdmissionPlugins(ctx, reqCtx.SchedulingRequest); err != nil {
+	if err := d.runRequestHeaderProcessors(ctx, reqCtx.SchedulingRequest); err != nil {
 		return reqCtx, err
 	}
+	// Derive FairnessID from agent-identity attribute if not already set by explicit header.
+	if reqCtx.SchedulingRequest.FairnessID == "" {
```
elevran, line 284: nit: with this addition there are two consecutive `if FairnessID == ""` checks (here and again at line 288). Fold into one block so the precedence reads top-to-bottom in a single `if`.

elevran (commented):
The refactor cleanly executes what #1450 asked for - agent identity is no longer buried inside `FairnessID`, but exposed as a first-class request attribute the Director reads. Left a couple of nits and a question - none are blockers. Let me know if you want merged as is or with fixes.

pkg/epp/framework/interface/requestcontrol/plugins.go
shmuelk: Envoy adds a set of "pseudo headers" to the request. They include:
 - `:method` - The method of the request
 - `:path` - the path of the request

praveingk:
Thanks @elevran . I have addressed your comments in https://github.com/llm-d/llm-d-router/pull/1754/commits/41a2329351ae19fbdb4fe8044c9fca70389b26d0

elevran (approved):

PR #1827, opened 2026-06-26
fix: base release-notes assembly branch on main to exclude unrelated diffs

liu-cong:
## Summary
- Fixes the release-notes assembly workflow including unrelated files in the auto-generated PR (e.g. `hack/push-chart.sh` in #1688).
- Root cause: the tag lives on a release branch that carries release-specific changes (e.g. pinned image versions) that must never be merged back to main. The workflow branched from the tagged commit, so those release-branch-only changes leaked into the PR diff against main.
- Fix: split the script into three phases:
  1. **Read fragments from the tagged commit** — preserves the existing guard against post-tag fragments.
  2. **Branch from `origin/main`** — the PR now only contains release-notes file changes, not release-branch-specific diffs.
  3. **Apply changes on main** — update `RELEASE-NOTES.md` and remove only the fragments that exist on main.
- Additionally, skip the assembly job entirely for RC tags (e.g. `v0.9.0-rc.1`) since release notes should only be assembled for final releases.

## Test plan
Simulated the fixed workflow locally against the v0.9.0 tag (which lives on a release branch that diverges from main):

- [x] Fragments are read from the tagged commit: found 40 fragments on v0.9.0.
- [x] Branch is created from `origin/main`, not the tagged commit.
- [x] Only `RELEASE-NOTES.md` is staged — no unrelated files (e.g. `hack/push-chart.sh`) appear in the diff.
- [x] Fragments already removed from main (by the original #1688 merge) are safely skipped via the `-f` guard.
- [x] No fragments merged after the tag are picked up (glob runs on the tagged commit checkout, not main).
- [x] RC tags are excluded via `if: ${{ !contains(github.ref_name, '-rc') }}` condition on the job.

### DIFF
--- a/.github/workflows/release-notes-assemble.yaml
+++ b/.github/workflows/release-notes-assemble.yaml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   assemble:
+    if: ${{ !contains(github.ref_name, '-rc') }}
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/.github/workflows/release-notes-assemble.yaml
+++ b/.github/workflows/release-notes-assemble.yaml
@@ -45,6 +45,9 @@ jobs:
         run: |
           set -euo pipefail
 
+          # ── Phase 1: read fragments from the tagged commit ──────────
+          # The checkout step landed us on the tagged commit so that
+          # fragments merged into main *after* the tag are excluded.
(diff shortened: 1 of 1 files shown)


liu-cong:
cc @elevran @ahg-g @vMaroon

ahg-g (approved):

PR #1834, opened 2026-06-27
feat(epp): add OpenTelemetry spans for the scheduler scoring path

### LINKED ISSUE #1692: Add span for scheduler score
Sub task for #1483

mvanhorn:
**What type of PR is this?**

/kind feature

## Summary

Adds fine-grained OpenTelemetry spans to the EPP scheduler scoring path so a single request trace shows which scorers ran, how long each took, and the aggregate score signals each produced.

`SchedulerProfile.runScorerPlugins` now opens a parent `llm_d.epp.scoring` span over the scorer chain, and each scorer invocation runs inside an `llm_d.epp.scorer.<type>` child span. The child spans nest under the parent, and the existing plugin-internal span (e.g. `score_prefix_cache`) nests under its scorer span. This mirrors the span-emission pattern already used by the precise-prefix-cache scorer.

Attributes are request- and chain-level only:

- parent: `llm_d.epp.scorer.count`, `llm_d.epp.scoring.candidate_endpoints`, plus `gen_ai.request.model` / `gen_ai.request.id` when present.
- per-scorer: `llm_d.epp.scorer.type`, `llm_d.epp.scorer.name`, `llm_d.epp.scorer.weight`, `llm_d.epp.scorer.candidate_endpoints`, and aggregate `llm_d.epp.scorer.score.max` / `.avg` / `endpoints_scored` derived from the returned score map.

No per-pod / per-endpoint attribute keys are emitted, keeping span cardinality bounded.

## Why this matters

Issue #1692 ("Add span for scheduler score") is the scoped sub-task of umbrella #1483, which asks for spans across the scheduling/scoring pipeline so operators can see scorer-level behavior inside a request trace instead of only the top-level `gateway.request` / `gateway.request_orchestration` spans. Today the scorer chain is invisible in traces: there is no way to tell which scorer dominated a routing decision or which one was slow.

This package is allocation-sensitive (the surrounding code documents per-request allocation work in `runScorerPlugins` and `runPickerPlugin`), so the scoring path stays allocation-free when tracing is disabled: the parent span's `IsRecording()` is checked once and all attribute and child-span construction is skipped on the default no-op path, with the tracer and span-kind option resolved once rather than per scorer. `BenchmarkSchedule` confirms the disabled path holds at the prior baseline.

## Testing

- [x] Parent `llm_d.epp.scoring` span recorded with one `llm_d.epp.scorer.<type>` child per scorer, children nested under the parent, with correct weight, candidate-count, and aggregate max/avg attributes.
- [x] No per-pod / per-endpoint attribute keys present on any emitted span.
- [x] Scoring results and scorer call counts are unchanged with a no-op tracer; no spans recorded on the disabled path.
- [x] A scorer returning an empty score map still emits its span and omits aggregates without dividing by zero.

Fixes #1692

**Release note**:
```release-note
Add OpenTelemetry spans for the EPP scheduler scoring path: a parent `llm_d.epp.scoring` span over the scorer chain and per-scorer `llm_d.epp.scorer.<type>` child spans carrying scorer identity, weight, candidate count, and aggregate score signals. No per-endpoint attributes are emitted, and spans are skipped entirely when tracing is disabled.
```


pkg/epp/scheduling/scheduler_profile.go
```diff
@@ -184,6 +213,69 @@ func (p *SchedulerProfile) runScorerPlugins(ctx context.Context, request *fwksch
 	return weightedScorePerEndpoint
 }
 
+// runScorer invokes a single weighted scorer and records its latency metric.
```
elevran, line 221: Is there a need to have a span per plugin or we can live with a single span for `scheduling`?
Any reason to suspect that a scorer would take noticeable time to warrant creating such fine grained spans?
It would be good to have some rationale for the production of detailed OTel traces at the granularity of internal function calls.

mvanhorn:
Fair question — the rationale, and I'm happy to drop to a single span if it doesn't hold up.

Why per-scorer rather than one `scheduling` span: the scorer chain is the part of the EPP path where the plugin set is operator-configured, so the interesting number is *which* scorer is slow on a given deployment, not that scoring took N ms. With one span you get the aggregate and then have to bisect by disabling scorers; with a child span per scorer the trace answers it directly. It also gives the existing plugin-internal spans (e.g. `score_prefix_cache`) a natural parent instead of hanging them off the request span.

On cost: I don't assume scorers are individually slow — the child spans are there for the tail, not the median. The hot path is guarded: `tracer.Start` on the parent returns a non-recording span when tracing is disabled, and `span.IsRecording()` gates every attribute and child-span construction after it, so with tracing off the added work is one `Start` call and a boolean check. Attributes are request- and chain-level only, no per-endpoint keys, so span cardinality stays bounded by the configured scorer count.

If you'd rather keep the trace coarse, I'll collapse to the single `llm_d.epp.scoring` span and keep the per-scorer latency metric, which covers the "which scorer is slow" question at lower trace cost. Say which you prefer and I'll push that — the branch needs a rebase anyway.

elevran:
Prefer not to add data and visualization until we have some initial indication of its need / value.
@ahg-g @gyliu513 a question on #1483 and #1692 - is there any reason to collect span on every filter/scorer? I'm concerned with information and network load if we collect and report too much. I'm not aware of any reason to expect a filter/scorer to take time.
The following might be a plausible order (in terms of increasing logging and code)
- single span for scheduling
- span for each scheduling profile and **phase** (all filters, all scores, picker)
- collect, but not log spans, for all plugins (map[string]time.Duration), and create a span (with the collect info) only when the overall scheduling is at least 3x the moving average (and/or over some absolute value).
- collect and report on all plugins, always

If the collection and distribution/reporting is super cheap than we can just so the last. Do you have some information/expectations?

gyliu513:
@elevran 

Fair concern on trace volume — let me separate cost from value.

**On cost:** Router tracing uses head-based `ParentBased(TraceIDRatioBased)` sampling (default 0.1 via `OTEL_TRACES_SAMPLER` / `OTEL_TRACES_SAMPLER_ARG`) with a batched OTLP exporter. The keep/drop decision happens at the root span, so on unsampled requests these scorer spans are not recorded or exported — this PR already guards that path with `IsRecording()` and stays allocation-free at baseline. On sampled traces, per-scorer spans add a small, bounded number of spans we've already decided to keep. The volume knob is the sample rate (and upstream sampling via `ParentBased`), not scorer granularity — "report on all plugins, always" effectively means "always on sampled traces," which is affordable.

That's why I'd avoid "span only when scheduling > 3x the moving average." Head-based sampling has already committed to the trace before a scorer runs, so a per-plugin "only if slow" gate yields partial, non-deterministic traces (some plugins present, some not) that break the parent/child structure — plus it needs stateful moving-average tracking on a hot path we keep allocation-free.

**On value:** Most scorers are sub-millisecond, agreed — but the exception is the point. Precise-prefix-cache already emits its own span because it does real lookup work, and cache/KV-aware scorers vary with pod state. Per-scorer spans show which scorer dominated or stalled a decision on a sampled trace; phase-level spans can't. For scoring I lean toward keeping the per-scorer spans as in this PR.

**Scope note:** This PR is scorers only (#1692). Filters (#1483) are separate and more numerous/trivial — happy to use a conservative default there (phase-level; per-plugin behind a flag if needed) when we get to it.

elevran:
@gyliu513 thanks for clarifying.

> On cost: Router tracing uses head-based ParentBased(TraceIDRatioBased) sampling (default 0.1 via OTEL_TRACES_SAMPLER / OTEL_TRACES_SAMPLER_ARG)...

Am I understanding correctly that this collects a probalistic sample of traces? Won't that mean that the chance of capturing the exception is also based on the sample ratio, making a rare event ("the exception") 10x more rare (assuming 0.1 SAMPLER means 10% of samples are picked)?
Just confirming my understanding, I'm ok with progressing on all scorers (though I'd argue for the same logic in filters if they do similar work, e.g., on prefix caching)

gyliu513:
@elevran Yes, your understanding is correct.

`ParentBased(TraceIDRatioBased(0.1))` is probabilistic, head-based sampling: roughly 10% of root traces are kept, and child spans (including per-scorer) are only recorded/exported on those kept traces. So visibility of a rare tail event in traces is scaled by the sample rate — a 1-in-1000 event shows up in traces roughly 1-in-10,000 at 0.1 sampling.

That's the intentional tradeoff: traces are for deep inspection on a sampled subset, not complete tail capture. For "which scorer is slow" on aggregate (including rare tails), we still rely on `llm_d_epp_plugin_duration_seconds`, which is unsampled. Spans complement that by tying a sampled request's full scorer chain (and nested prefix-cache / kv-cache spans) together in one trace. For targeted debugging, operators can also raise `OTEL_TRACES_SAMPLER_ARG` or rely on upstream `ParentBased` if the gateway force-samples a trace.

On filters: agreed, when we get to #1483, I'd use the same pattern for heavy plugins (e.g. prefix-cache-like filters): per-plugin spans where the work is non-trivial, phase-level for the rest. Happy to proceed with all scorers in this PR as-is.

gyliu513:
@elevran @ahg-g can you help review?

pkg/epp/scheduling/scheduler_profile.go
```diff
@@ -35,6 +35,14 @@ import (
 	"github.com/llm-d/llm-d-router/pkg/epp/metrics"
 )
 
+// tracerScope is the OTel instrumentation scope for scheduler-level scoring
+// spans emitted from this package.
+const tracerScope = "llm-d-router/pkg/epp/scheduling"
```
elevran: constant duplicates the existing exported `TracerScope` in `constants.go`

pkg/epp/scheduling/scheduler_profile.go
```diff
@@ -203,6 +229,69 @@ func (p *SchedulerProfile) runScorerPlugins(ctx context.Context, request *fwksch
...
+			attribute.Float64("llm_d.epp.scorer.score.max", maxScore),
+			attribute.Float64("llm_d.epp.scorer.score.avg", totalScore/float64(len(scores))),
+			attribute.Int("llm_d.epp.scorer.endpoints_scored", len(scores)),
+		)
+	}
+
+	return scores
+}
+
+// setRequestSpanAttributes mirrors the request-identity attribute schema used
```
elevran: duplicates the inline request-attribute block already used in `runFilter/PickerPlugins` in this file. Makes sense to unify all three call sites on one helper. Optional in this PR (i.e. easily done now or can be a follow-up).

elevran (commented):
two minor nits before this can merge

mvanhorn:
Both nits addressed in 0a1a00b.

Dropped the local tracerScope constant and pointed the tracer at the exported TracerScope in constants.go. Also pulled the request-attribute block into one requestSpanAttributes helper, so runFilterPlugins, runScorerPlugins and runPickerPlugin all go through it instead of building the same two attributes inline three times. Emitted attributes are unchanged, and there's a small test covering nil / empty / populated requests.

Still owe you an answer on the per-plugin span granularity question from July 1. I'd rather write that one properly than bolt it onto this comment.

elevran:
@mvanhorn please address CICD failures.

mvanhorn:
Addressed. The failures were not from the tracing change: upstream renamed `EndpointMetadata.NamespacedName` to `ID`, and this branch was 74 commits behind, so `scheduler_profile_test.go` stopped compiling. That one package failing to typecheck is what took out both `lint` and `test`.

Merged main and updated the test to the new field. The scoring spans are unchanged.

Locally: `go vet ./pkg/epp/scheduling/...` clean, `go test ./pkg/epp/scheduling/...` ok, `go build ./...` clean, `gofmt` clean.

elevran:
DCO bot acting up. Can override manually.

> Still owe you an answer on the per-plugin span granularity question from July 1. I'd rather write that one properly than bolt it onto this comment.

@mvanhorn should hold PR merging on that or can be later?

elevran (approved):

mvanhorn:
Appreciate the merge, @elevran. The scheduler spans should make EPP latency much easier to trace.

Issue #1755, opened 2026-06-26
[Operations] Tracking: implement DumpState for all in-tree plugins


liu-cong:
Follow-up to #1148: add `plugin.StateDumper` coverage across all in-tree plugins that are registered in `cmd/epp/runner/runner.go`.

Scope:
- Implement `DumpState() (json.RawMessage, error)` for plugins that maintain mutable runtime state.
- Keep payloads sanitized and bounded per `docs/plugin_debug.md`.
- Add/extend plugin-level tests for dump behavior.

Deep-dive classification (production plugins only):
- Evaluated plugin packages: 56
- Already implemented: 1 (`requestcontrol/dataproducer/inflightload`)
- Requires DumpState follow-up: 8
- No mutable runtime state (no DumpState needed): 47
- Test-only plugin packages were excluded from scope.

Child issues:
- [ ] #1776 - [Operations] Implement DumpState for file-discovery plugin
- [ ] #1782 - [Operations] Implement DumpState for predicted-latency-producer plugin
- [ ] #1789 - [Operations] Implement DumpState for mm-embeddings-cache-producer plugin
- [ ] #1790 - [Operations] Implement DumpState for no-hit-lru-scorer plugin
- [ ] #1794 - [Operations] Implement DumpState for precise-prefix-cache-producer plugin
- [ ] #1798 - [Operations] Implement DumpState for program-aware-fairness plugin
- [ ] #1801 - [Operations] Implement DumpState for approx-prefix-cache-producer plugin
- [ ] #1803 - [Operations] Implement DumpState for round-robin-fairness-policy plugin

PR #1835, opened 2026-06-27
feat: implement DumpState for approx-prefix-cache-producer

### LINKED ISSUE #1801: [Operations] Implement DumpState for approx-prefix-cache-producer plugin
Parent tracking issue: #1755

Plugin registration constant: `reqdataprodprefix.ApproxPrefixCachePluginType`
Plugin package: `github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix`

Acceptance criteria:

thc1006:
**What type of PR is this?**

/kind feature

Adds `plugin.StateDumper` to the `approx-prefix-cache-producer` so its prefix index
can be inspected through `GET /debug/plugins/state`, the same way the

`DumpState` returns the number of indexed prefix-cache blocks per pod, sorted by
count and capped at 100 pods (with a `truncated` flag) so the response stays small
when a pool has many pods. It includes pod names and counts only; the block hashes
are derived from prompt content, so they are left out on purpose.

Reading the per-pod sizes needs a small `PodBlockCounts()` accessor on the indexer,
which takes the existing read lock.

```json
{"pods":[{"pod":"default/vllm-a","blocks":1240},{"pod":"default/vllm-b","blocks":980}],"totalPods":2,"maxPods":100,"truncated":false}
```

Part of #1755.

Fixes #1801

```release-note
NONE
```

**Testing**:

New unit tests, all passing:
- [x] Per-pod block counts are reported, ordered by count, with no block hash in the output
- [x] The pod list is capped and the truncated flag is set, busiest pod first
- [x] Nil and empty indexers return valid JSON without panicking

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sort"
 	"sync"
 	"time"
 
@@ -52,6 +53,7 @@ var minBlockSizeTokens = 64
 var (
 	_ requestcontrol.DataProducer = &dataProducer{}
 	_ requestcontrol.PreRequest   = &dataProducer{}
+	_ plugin.StateDumper          = &dataProducer{}
 )
 
 // dataProducer is a plugin that produces data consumed by approx prefix cache aware scheduling.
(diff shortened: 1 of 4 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
```diff
@@ -69,6 +71,56 @@ func (p *dataProducer) TypedName() plugin.TypedName {
...
 
+const maxDebugDumpPods = 100
+
+// prefixIndexState is the sanitized snapshot returned by DumpState. It carries
+// per-pod block counts only; the prompt-derived block hashes are never exposed.
+type prefixIndexState struct {
+	Pods      []podBlockCount `json:"pods"`
+	TotalPods int             `json:"totalPods"`
+	MaxPods   int             `json:"maxPods"`
+	Truncated bool            `json:"truncated"`
```
shmuelk: @thc1006 Isn't this field extraneous? A simple compare by the user of TotalPods and MaxPods will indicate whether or not the data outputed has been truncated.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
```diff
@@ -69,6 +71,56 @@ func (p *dataProducer) TypedName() plugin.TypedName {
...
+
+	sort.SliceStable(state.Pods, func(a, b int) bool {
+		if state.Pods[a].Blocks != state.Pods[b].Blocks {
+			return state.Pods[a].Blocks > state.Pods[b].Blocks
+		}
+		return state.Pods[a].Pod < state.Pods[b].Pod
+	})
+
+	if len(state.Pods) > maxDebugDumpPods {
+		state.Pods = state.Pods[:maxDebugDumpPods]
```
shmuelk, line 118: Rather than simply truncating wouldn't it be better to sort by blocks per pod to show as much information as possible?

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
thc1006: You're right, this is derivable from totalPods > maxPods, so I've dropped the field and documented that on the struct. The cap stays in place so the payload is still bounded. I made the same change to the other list-style dumps in this series (#1836, #1837, #1838) for consistency.

The same reasoning applies to inflight-load-producer and the example in docs/plugin_debug.md. Those are out of scope here, but I'm fine opening a follow-up to align them if that's useful.

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
thc1006: It already sorts by blocks per pod before capping (descending, with the pod name as a tiebreaker), so the cap keeps the busiest pods. If you had a different ordering in mind, I can change it.

shmuelk (approved):

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/plugin.go
```diff
@@ -69,6 +71,55 @@ func (p *dataProducer) TypedName() plugin.TypedName {
 	return p.typedName
 }
 
+const maxDebugDumpPods = 100
+
+// prefixIndexState is the sanitized snapshot returned by DumpState. It carries
+// per-pod block counts only; the prompt-derived block hashes are never exposed.
+// The dump is partial when TotalPods exceeds MaxPods.
+type prefixIndexState struct {
```
liu-cong, line 79: let's also add the pluginState

thc1006: This needs the same snapshot accessor on the shared PluginState that I raised on #1838 and #1840. The wrinkle specific to this plugin is that the state here (SchedulingContextState) is cleared at the end of PreRequest, so the dump is empty outside an in-flight request. If the aim is to inspect in-flight requests I can list their request ids and state keys; if you'd also want the values, those are prompt-derived block hashes, so they carry the same sanitization consideration as the #1838 cache keys. I'd add the accessor once and reuse it across the three PRs.

liu-cong (approved):

thc1006:
Thanks very much for the reviews, and for approving #1840 and #1835.

As far as I can tell these are all just waiting on CI at this point:

- #1835 and #1839 are only red on `e2e-router (pd)`, which is the same flaky check I flagged earlier on #1839. #1839's run is actually from June 27 and hasn't re-run since, and that job has a history of flipping between pass and fail on unchanged code, so a re-run should sort it out.
- #1840 passed everything that ran, but the e2e jobs are still behind the first-time-contributor "Approve and run" gate, so they haven't started.
- For #1838 I pushed the podList change we talked about, which dropped the earlier approval, so it needs another look whenever you have a moment.

Really appreciate the time on these. No urgency on my end.

PR #1836, opened 2026-06-27
feat: implement DumpState for file-discovery

### LINKED ISSUE #1776: [Operations] Implement DumpState for file-discovery plugin
Parent tracking issue: #1755

Plugin registration constant: `discoveryfile.PluginType`
Plugin package: `github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/discovery/file`

Acceptance criteria:

thc1006:
**What type of PR is this?**

/kind feature

Adds `plugin.StateDumper` to the file-discovery plugin so the endpoints it
currently has loaded can be inspected through `GET /debug/plugins/state`, the
same way the `inflight-load-producer` already exposes its state.

`DumpState` reports the endpoint identities (namespace/name) only, sorted and
capped at 100 with a `truncated` flag so the payload stays bounded. Addresses
and labels are deliberately left out.

The endpoint set is read concurrently with file reloads, so this adds a
read-write mutex to guard it; `load` now swaps the set under the lock. Behaviour
is otherwise unchanged.

```json
{"endpoints":["default/pod-a","default/pod-b"],"totalEndpoints":2,"maxEndpoints":100,"truncated":false}
```

Part of #1755.

Fixes #1776

```release-note
Implement DumpState for file-discovery plugin
```

**Testing**:

New unit tests, all passing (race detector enabled):
- [x] Endpoint identities are reported and sorted
- [x] The list is capped with the truncated flag set, keeping the sorted-first endpoints
- [x] Concurrent reloads and dumps run cleanly under -race
- [x] An empty set returns valid JSON

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
+++ b/pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
@@ -26,6 +26,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"sort"
 	"strconv"
 	"sync"
 
@@ -71,6 +72,8 @@ type FileDiscovery struct {
 	typedName fwkplugin.TypedName
 	path      string
 	watchFile bool
+	// mu guards endpoints, which DumpState reads concurrently with load.
+	mu sync.RWMutex
 	// endpoints is the set of endpoint identities applied to the datastore
 	// from the last successful load. Used as a key set only -- values are
 	// zero-byte structs. Compared against the entries parsed during a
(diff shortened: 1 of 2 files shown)


pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
```diff
@@ -108,6 +114,41 @@ func Factory(name string, parameters *json.Decoder, _ fwkplugin.Handle) (fwkplug
...
 
+const maxDebugDumpEndpoints = 100
+
+// discoveryState is the sanitized snapshot returned by DumpState: discovered
+// endpoint identities only, never their addresses or labels.
+type discoveryState struct {
+	Endpoints      []string `json:"endpoints"`
+	TotalEndpoints int      `json:"totalEndpoints"`
+	MaxEndpoints   int      `json:"maxEndpoints"`
+	Truncated      bool     `json:"truncated"`
```
shmuelk: Why is the field Truncated needed. A quick manual comparison of TotalEndpoints and MaxEndpoints will tell a user whether or not the output was truncated.

thc1006: Right, totalEndpoints > maxEndpoints carries the same information, so I've removed the field and noted the derivation on the struct. The cap stays so the dump is still bounded.

shmuelk (approved):

liu-cong (approved):

liu-cong:
Thanks!

PR #1837, opened 2026-06-27
feat: implement DumpState for no-hit-lru-scorer

### LINKED ISSUE #1790: [Operations] Implement DumpState for no-hit-lru-scorer plugin
Parent tracking issue: #1755

Plugin registration constant: `nohitlru.NoHitLRUType`
Plugin package: `github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/scorer/nohitlru`

Acceptance criteria:

thc1006:
**What type of PR is this?**

/kind feature

Adds `plugin.StateDumper` to the no-hit-lru-scorer so the endpoints it tracks in
its cold-request LRU can be inspected through `GET /debug/plugins/state`, the
same way the `inflight-load-producer` already exposes its state.

`DumpState` lists the tracked endpoints in LRU order (least-recently-used first,
the order the scorer favors for the next cold request), capped at 100 with a
`truncated` flag so the payload stays bounded. Only endpoint identities are
reported. The cache is already concurrency-safe, so no extra locking is needed.

```json
{"endpoints":["default/pod-a","default/pod-b"],"totalEndpoints":2,"maxEndpoints":100,"truncated":false}
```

Part of #1755.

Fixes #1790

```release-note
NONE
```

**Testing**:

New unit tests, all passing:
- [x] Tracked endpoints are reported in LRU order
- [x] The list is capped with the truncated flag set, keeping the least-recently-used endpoints
- [x] An empty LRU and a zero-value scorer both return valid JSON without panicking

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/scorer/nohitlru/no_hit_lru.go
+++ b/pkg/epp/framework/plugins/scheduling/scorer/nohitlru/no_hit_lru.go
@@ -33,6 +33,7 @@ const (
 var _ scheduling.Scorer = &NoHitLRU{}
 var _ requestcontrol.PreRequest = &NoHitLRU{}
 var _ plugin.ConsumerPlugin = &NoHitLRU{}
+var _ plugin.StateDumper = &NoHitLRU{}
 
 // Parameters defines the parameters for the NoHitLRU scorer.
 type Parameters struct {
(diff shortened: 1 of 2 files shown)


pkg/epp/framework/plugins/scheduling/scorer/nohitlru/no_hit_lru.go
```diff
@@ -122,6 +123,41 @@ func (s *NoHitLRU) Category() scheduling.ScorerCategory {
...
 
+const maxDebugDumpEndpoints = 100
+
+// noHitLRUState is the sanitized snapshot returned by DumpState: the endpoint
+// identities tracked in the cold-request LRU, nothing else.
+type noHitLRUState struct {
+	Endpoints      []string `json:"endpoints"`
+	TotalEndpoints int      `json:"totalEndpoints"`
+	MaxEndpoints   int      `json:"maxEndpoints"`
+	Truncated      bool     `json:"truncated"`
```
shmuelk: Why is the field Truncated needed? A simple manual comparison of Totalendpoints and MaxEndpoints will tell the user that the data has been truncated.

pkg/epp/framework/plugins/scheduling/scorer/nohitlru/no_hit_lru.go
```diff
@@ -122,6 +123,41 @@ func (s *NoHitLRU) Category() scheduling.ScorerCategory {
...
+	if s.lruCache == nil {
+		return json.Marshal(state)
+	}
+
+	// Keys returns a fresh slice, least-recently-used first, safe to truncate.
+	// https://pkg.go.dev/github.com/hashicorp/golang-lru/v2#Cache.Keys
+	keys := s.lruCache.Keys()
+	state.TotalEndpoints = len(keys)
+	if len(keys) > maxDebugDumpEndpoints {
+		keys = keys[:maxDebugDumpEndpoints]
```
shmuelk, line 154: Rather than doing a simple truncation some sort perhaps based on use counts might be better to show as much interesting information as possible.

pkg/epp/framework/plugins/scheduling/scorer/nohitlru/no_hit_lru.go
thc1006: Agreed, it's redundant with totalEndpoints > maxEndpoints. I've removed it and documented that on the struct, keeping the cap for bounded output.

pkg/epp/framework/plugins/scheduling/scorer/nohitlru/no_hit_lru.go
thc1006: There's no per-endpoint use count to sort by here. The cold-request LRU stores endpoint identity only (the value is an empty struct). What the dump preserves is the LRU order itself, least-recently-used first, which is the ranking the scorer acts on: those are the endpoints most likely to be selected for the next cold request, so they're the ones kept when the list is capped. If you'd prefer a different ordering, I can look into it.

shmuelk (approved):

liu-cong (approved):
Thanks!

PR #1838, opened 2026-06-27
feat: implement DumpState for mm-embeddings-cache-producer

### LINKED ISSUE #1789: [Operations] Implement DumpState for mm-embeddings-cache-producer plugin
Parent tracking issue: #1755

Plugin registration constant: `mmproducer.ProducerType`
Plugin package: `github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal`

Acceptance criteria:

thc1006:
**What type of PR is this?**

/kind feature

Adds `plugin.StateDumper` to the mm-embeddings-cache-producer so its state can be
inspected through `GET /debug/plugins/state`, the same way the

`DumpState` reports the known pods from the datalayer and the per-pod encoder-cache
occupancy (how many items each pod has cached), ordered by count and capped at 100
so the payload stays bounded; the totals report the full counts so a consumer can
tell when a list is partial. It emits pod identities and counts only; the cached
multimodal content hashes (the cache keys) are deliberately left out.

```json
{"podList":["default/vllm-a","default/vllm-b"],"totalKnownPods":2,"pods":[{"pod":"default/vllm-a","items":12},{"pod":"default/vllm-b","items":8}],"totalPods":2,"maxPods":100}
```

Part of #1755.

Fixes #1789

```release-note
NONE
```

**Testing**:

New unit tests, all passing:
- [x] Known pods and per-pod item counts are reported (ordered by count), with no content hash in the output
- [x] Both lists are capped while the totals report the full counts, busiest pod first
- [x] An empty cache returns valid JSON with empty lists, not null

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
@@ -23,6 +23,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sort"
 	"sync"
 	"time"
 
@@ -59,6 +60,7 @@ var (
 	_ requestcontrol.DataProducer = &Producer{}
 	_ requestcontrol.PreRequest   = &Producer{}
 	_ fwkdl.EndpointExtractor     = &Producer{}
+	_ plugin.StateDumper          = &Producer{}
 )
 
 // Parameters configures the multimodal encoder-cache data producer.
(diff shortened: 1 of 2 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -172,6 +174,48 @@ func (p *Producer) TypedName() plugin.TypedName {
...
+	p.mutex.RUnlock()
+
+	sort.SliceStable(pods, func(a, b int) bool {
+		if pods[a].Items != pods[b].Items {
+			return pods[a].Items > pods[b].Items
+		}
+		return pods[a].Pod < pods[b].Pod
+	})
+	if len(pods) > maxDebugDumpPods {
+		pods = pods[:maxDebugDumpPods]
```
shmuelk, line 242: Rather than a simple truncation, perhaps a sort by items per pod would be better at showing as much information as possible.

thc1006: It already sorts by items per pod before the cap (descending, with the pod name as a tiebreaker), so the busiest pods are kept. I also dropped the truncated field here to match the rest of the series, since it's derivable from totalPods > maxPods.

shmuelk (dismissed):

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -172,6 +174,47 @@ func (p *Producer) TypedName() plugin.TypedName {
 	return p.typedName
 }
 
+const maxDebugDumpPods = 100
+
+// encoderCacheState is the sanitized snapshot returned by DumpState: per-pod
+// cached-item counts only. The multimodal content hashes are never exposed.
+// The dump is partial when TotalPods exceeds MaxPods.
+type encoderCacheState struct {
```
liu-cong, line 187: can we also dump the pluginState, the entire `podList`, and the `caches` (dump for the first 100  pods, and the first 100 cache entries)

thc1006: Thanks. I've added the podList part: the dump now lists the datalayer's known pods (sorted, capped at 100 with the total reported), separate from the per-pod cache view.

Before I add the cache entries I'd like to check one thing with you. The cache keys are the multimodal content hashes, and plugin_debug.md (and this plugin's own doc) currently state that content-derived values are kept out of the dump, which is why it started as counts only. I have no objection to exposing them for debugging, given that the endpoint is auth-gated and the hashes aren't reversible, but it does reverse that documented position, so I'd prefer a quick ok from you and shmuelk before I make the switch. I'll update the doc as part of the same change.

pluginState is the same question as on #1840: it needs a new enumeration accessor on the shared PluginState, and the entries are cleared at the end of PreRequest, so the dump is empty outside an in-flight request. I'll add it once the approach is settled.

gkneighb:
Same context as my review on #1840: I implemented DumpState for this plugin in #1850 before yielding to this PR, so I know the state surface. Verified beyond reading:

- The branch applies cleanly on current main, builds, and the package tests pass locally with `-race` (worth noting since this PR adds a concurrent-writes test, which is a nice touch the others in the series don't have).
- The locking claims hold against the existing code: `podList()` is indeed called outside the cache lock in `removeStalePods` (`producer.go:292` vs the `Lock` at `:301`), so `DumpState` sampling it the same way is consistent, and the documented best-effort inconsistency between `PodList` and the per-pod view is the honest trade. Sorting and capping happen on a copied slice outside the lock, and `lru.Cache.Len()` is itself thread-safe.

On the open question to @liu-cong about dumping the cache keys: I'd argue for keeping the dump counts-only, slightly more strongly than the PR does. The content hashes aren't reversible, but they are **correlatable** -- the same hash appearing across pods or over time links identical multimodal content between requests, which is exactly the class of signal `plugin_debug.md`'s no-content-derived-values rule exists to keep out of an operational endpoint. Occupancy counts answer the debugging questions (is the cache filling, is affinity working, which pods are hot) without that. If enumeration is genuinely needed later, it deserves the deliberate doc change thc1006 describes rather than riding along here. I reached the same counts-only conclusion implementing #1850.

pluginState is the same cross-plugin question as on #1840 (request-scoped, cleared at PreRequest end, needs a shared enumeration accessor) -- best settled once for the whole #1755 series.

liu-cong (approved):

liu-cong:
the failed checks are unrelated

Issue #1625, opened 2026-06-12
[Observaibility] epp/metrics: unbounded label cardinality from user-controlled inputs (metrics cardinality DoS)


gyliu513:
**What happened**:

Several EPP metrics use label values that come directly from untrusted request input (request body and headers) with no validation, allowlist, hashing, or cap. Because the Prometheus client *Vec types never evict label combinations they have seen, a client can create an unbounded number of time series by varying these values, exhausting EPP memory and degrading /metrics scrape (and the upstream Prometheus). This is a metrics-cardinality denial-of-service vector that requires only small, low-volume traffic, not high request load.

Affected labels and where the values originate

1. `model_name` / `target_model_name`. `IncomingModelName` is taken verbatim from the request body `model` field, and an unknown model is not rejected — `getInferenceObjective` falls back to a default objective and the request proceeds. The value is then recorded as a label by `RecordRequestCounter` / `RecordRequestSizes` etc. These labels are attached to most business metrics: `request_total`, `request_error_total`, `request_duration_seconds`, `request_sizes`, `response_sizes`, `input_tokens`, `output_tokens`, `prompt_cached_tokens`, `running_requests`. This is the highest-impact case because it spans many metrics and needs only a request body, no headers.

2. `fairness_id`. The value comes from the `x-llm-d-inference-fairness-id` request header (with deprecated alias `x-gateway-inference-fairness-id`). The only processing is defaulting an empty value to `default-flow`. It flows into `FlowKey.ID` and is recorded by the flow-control metrics: `flow_control_queue_size`, `flow_control_queue_bytes`, `flow_control_request_queue_duration_seconds`, `flow_control_request_enqueue_duration_seconds`.

3. `pod_name` / `namespace` / `port`, attached to `scheduler_attempts_total`. Not attacker-controlled, but unbounded over time: pod churn (rolling updates, spot reclaim, autoscaling) keeps minting new label sets while old ones are never evicted, producing a slow time-series leak rather than a sudden spike.

Impact
- EPP memory exhaustion (OOMKill) from retained time series, taking down request routing for legitimate traffic.
- /metrics scrape latency growth and timeouts; upstream Prometheus ingestion pressure.
- Achievable with low request volume, the cost is in label cardinality, not throughput.

/cc @ahg-g 

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
[...]

PR #1839, opened 2026-06-27
feat: implement DumpState for program-aware-fairness

### LINKED ISSUE #1798: [Operations] Implement DumpState for program-aware-fairness plugin
Parent tracking issue: #1755

Plugin registration constant: `programaware.ProgramAwarePluginType`
Plugin package: `github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware`

Acceptance criteria:

thc1006:
**What type of PR is this?**

/kind feature

Adds `plugin.StateDumper` to the program-aware-fairness plugin so its health can
be inspected through `GET /debug/plugins/state`, the same way the

Unlike the other plugins, this one keys its state by program ID, which comes from
a user-controlled request header (`x-llm-d-inference-fairness-id`) and is
high-cardinality. Dumping those IDs would expose user-controlled, tenant-identifying
values, which `docs/plugin_debug.md` says to omit. So `DumpState` reports only
bounded aggregates and no IDs: the number of tracked programs, total in-flight
requests, and Jain's fairness index (which the plugin already computes). That is
the right operational signal for a fairness policy without leaking identities.

```json
{"totalPrograms":12,"totalInFlight":34,"fairnessIndex":0.92}
```

A shared `jainFairnessIndex` helper keeps the index formula in one place so the
dump and the existing metric come from a single snapshot of the program map.

Part of #1755.

Fixes #1798

```release-note
NONE
```

**Testing**:

New unit tests, all passing:
- [x] Aggregates (program count, in-flight, fairness index) are reported
- [x] User-controlled program IDs never appear in the dump
- [x] An empty policy returns valid JSON (trivially fair)

### DIFF
--- a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
+++ b/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
@@ -72,6 +72,7 @@ var (
 	_ flowcontrol.FairnessPolicy  = &ProgramAwarePlugin{}
 	_ fwkrc.PreRequest            = &ProgramAwarePlugin{}
 	_ fwkrc.ResponseBodyProcessor = &ProgramAwarePlugin{}
+	_ plugin.StateDumper          = &ProgramAwarePlugin{}
 )
 
 //nolint:revive // factory name matches sibling fairness plugins.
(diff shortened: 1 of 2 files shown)


shmuelk (approved):

thc1006:
The failing e2e-router (pd) check is the "should report metrics" spec, which is flaky on main at the moment. It fails on the base commit (1fa3803d) with the same spec and nothing changed, and this PR only touches the program-aware fairness plugin. Could you re-run that job when you get a chance?

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
```diff
@@ -120,6 +121,45 @@ func (p *ProgramAwarePlugin) TypedName() plugin.TypedName {
 	return plugin.TypedName{Type: ProgramAwarePluginType, Name: p.name}
 }
 
+// fairnessDumpState is the sanitized snapshot returned by DumpState. Program IDs
+// come from a user-controlled request header, so they are omitted; only these
+// bounded aggregates are reported.
+type fairnessDumpState struct {
```
liu-cong, line 127: can we dump the entire programMetrics (can configure an upper limit of, say 1000).

thc1006: That makes sense. A per-program list (program id, in-flight, dispatched and wait counts, average wait) capped at 1000 and sorted by in-flight would cover it, and because the program id is already exposed as the program_id metric label it isn't new exposure. The one thing I want to get right is where to put it: this PR is already approved and only waiting on CI, so adding it here would re-trigger review. Would you prefer that, or to land this one as it stands and let me follow up with programMetrics shortly afterwards? I'm fine either way.

gkneighb:
Completing my pass over this series (#1840, #1838): I implemented DumpState for this plugin in #1851 before yielding, so same context. Verified locally: applies cleanly on current main, builds, package tests pass. The `jainFairnessIndex` extraction is behavior-preserving (same `n <= 1 || sumSq == 0` guard), the single-pass claim over `programMetrics` holds, and the `TestDumpStateOmitsProgramIDs` regression test is a nice guardrail for the sanitization position.

On the open question with @liu-cong about dumping per-program entries, one consideration from adjacent work: the justification "program id is already exposed as the `program_id` metric label so it isn't new exposure" is accurate today (`prometheus.go:26`, `strategy_las.go:72`) -- but that existing exposure is itself flagged work, not a settled baseline. Those GaugeVecs are keyed by a user-controlled header with no cardinality bound, which is Vector 2 of #1625 (the metrics-cardinality issue; #1909 fixes the model-name vector and fairness/program IDs are the deferred follow-up). Two implications:

- A per-program dump **capped at 1000 and sorted by in-flight** is bounded by construction -- strictly safer than the uncapped labels -- so no objection to the list itself.
- But if the metrics side later folds or caps program IDs, the debug dump would become the only surface emitting raw tenant-correlatable IDs. Worth deciding that deliberately and reconciling `docs/plugin_debug.md`'s tenant-identifying language when the follow-up lands, same shape as the content-hash question on #1838.

On where to put it: landing this as-is (approved, aggregates-only) and doing the per-program list as a follow-up seems right -- it keeps this PR at its reviewed scope and gives the sanitization question the separate discussion it deserves.

liu-cong (approved):

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/plugin.go
liu-cong: approved, feel free to open a PR to follow up, thanks

PR #1840, opened 2026-06-27
feat: implement DumpState for precise-prefix-cache-producer

### LINKED ISSUE #1794: [Operations] Implement DumpState for precise-prefix-cache-producer plugin
Parent tracking issue: #1755

Plugin registration constant: `preciseproducer.PluginType`
Plugin package: `github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache`

Acceptance criteria:

thc1006:
**What type of PR is this?**

/kind feature

Adds `plugin.StateDumper` to the precise-prefix-cache-producer so its operational
state can be inspected through `GET /debug/plugins/state`, the same way the

This producer's core state is the external `llm-d-kv-cache` KV-block index, whose
interface has no enumerate or count method and whose keys are prompt-derived
hashes, so that index is not exposed. `DumpState` reports the bounded state it can:
the active KV-event subscriber pod identities and the live speculative-cache
request ids (both sorted and capped, with totals so a consumer can tell when a list
is partial), whether speculative indexing is enabled, and the block size in tokens.

To make the subscribers and speculative entries testable, the `subscribersManager`
field is narrowed from `*kvevents.SubscriberManager` to a small local interface,
mirroring the existing `kvCacheIndexer` interface in this package.

```json
{"subscribers":["default/vllm-a","default/vllm-b"],"totalSubscribers":2,"maxSubscribers":100,"speculativeIndexing":true,"speculativeEntries":["req-1","req-2"],"totalSpeculativeEntries":2,"maxSpeculativeEntries":100,"blockSizeTokens":64}
```

Part of #1755.

Fixes #1794

```release-note
Implemented DumpState for precise-prefix-cache-producer
```

**Testing**:

New unit tests, all passing:
- [x] Active subscriber pod identities and live speculative-cache request ids are listed (sorted), with speculative state and block size reported
- [x] Both lists are capped while the totals report the full counts
- [x] A zero-value producer (nil subscriber manager and nil speculative cache) returns valid JSON with empty lists, not null

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sort"
 	"time"
 
 	"github.com/jellydator/ttlcache/v3"
@@ -64,7 +65,19 @@ type PluginConfig struct {
 	SpeculativeTTL string `json:"speculativeTTL"`
 }
 
-var _ requestcontrol.DataProducer = &Producer{}
+var (
+	_ requestcontrol.DataProducer = &Producer{}
+	_ plugin.StateDumper          = &Producer{}
+)
+
+// subscriberManager is the subset of kvevents.SubscriberManager the producer
+// relies on, narrowed so tests can substitute a fake.
+type subscriberManager interface {
+	EnsureSubscriber(ctx context.Context, podIdentifier, endpoint, topicFilter string, remoteSocket bool) error
+	RemoveSubscriber(ctx context.Context, podIdentifier string)
+	GetActiveSubscribers() ([]string, []string)
+	Shutdown(ctx context.Context)
+}
 
 // Producer is a DataProducer plugin that maintains a KV-block prefix-cache
 // index by subscribing to vLLM KV-events and writes per-endpoint
@@ -78,7 +91,7 @@ type Producer struct {
 	typedName      plugin.TypedName
 	kvCacheIndexer kvCacheIndexer
 
-	subscribersManager *kvevents.SubscriberManager
+	subscribersManager subscriberManager
 	kvEventsConfig     *kvevents.Config
 
 	kvBlockScorer kvcache.KVBlockScorer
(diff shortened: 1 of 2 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -202,6 +214,38 @@ func (p *Producer) TypedName() plugin.TypedName {
...
+}
+
+// DumpState reports the producer's bounded operational state: how many KV-event
+// subscribers are active, whether speculative indexing is on and how many
+// speculative entries are live, and the block size in tokens. Subscriber
+// identities and the prompt-derived block index are not exposed.
+func (p *Producer) DumpState() (json.RawMessage, error) {
+	activeSubscribers := 0
+	if p.subscribersManager != nil {
+		ids, _ := p.subscribersManager.GetActiveSubscribers()
```
liu-cong, line 248: let's print the actual subscribers instead of just the count. The purpose of this is for debugging

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -202,6 +214,38 @@ func (p *Producer) TypedName() plugin.TypedName {
...
+		activeSubscribers = len(ids)
+	}
+	speculativeEntries := 0
+	if p.speculativeCache != nil {
+		speculativeEntries = p.speculativeCache.Len()
```
liu-cong: let's print the first N entries (N=100 can be a good default, no need to make it configurable for now).

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -202,6 +214,38 @@ func (p *Producer) TypedName() plugin.TypedName {
 	return p.typedName
 }
 
+// precisePrefixState is the sanitized snapshot returned by DumpState. The
+// KV-block index is keyed by prompt-derived block hashes and is not enumerable,
+// so only these bounded, non-sensitive aggregates are reported.
+type precisePrefixState struct {
```
liu-cong, line 229: we should also add the pluginState.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
thc1006: Done. It now lists the active subscriber pod identities (sorted), with the total count alongside. I did keep a cap of 100 so the payload stays bounded, though the total still reflects the real number, and in practice the subscriber set is small enough that the cap rarely comes into play. If you'd rather have it uncapped, that's an easy change.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
thc1006: Done as well. The dump now lists up to 100 speculative-cache entries keyed by request id, sorted for stable output, with the total alongside.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
thc1006: For this one I'd like to settle the approach with you first. PluginState has no way to enumerate its entries today (only Read/Write/Delete), so exposing it means adding a snapshot method to the shared plugin_state.go. Its entries are also request-scoped and cleared at the end of PreRequest, so the dump would be empty except while a request is in flight. If the goal is to inspect in-flight requests, would listing their request ids and state keys (without the values) be enough? I'd add the accessor once and reuse it for #1835 and #1838 rather than writing it three times.

elevran:
@thc1006 @liu-cong @shmuelk - there are a number of similar DumpState PRs with a cap on entries set to 100. I suspect that was (at least partially) to cap memory.

- is it possible to "stream" the data directly to the socket/http.ResponseWriter so the memory in flight is limited?
- might it make sense to allow query params on the request? Could be useful to: list which state dumps are needed, put a cap on the output along with a filter or sorting order so the "important" data is returned?

thc1006:
@elevran Thanks. The cap is mainly to keep each dump bounded and readable per docs/plugin_debug.md; limiting memory is more of a side effect than the goal. Since each plugin's dump is already capped, the payload stays bounded even though the endpoint marshals the whole response in memory today.

Both streaming and query params sound worthwhile, but they'd land at the StateDumper interface / endpoint level rather than in any single plugin, so I'll leave the shape to you and the other maintainers and adjust these PRs to match. They implement the current interface as-is, so they can move independently or be picked up as a follow-up, whichever you prefer.

gkneighb:
Reviewing with some context: I implemented DumpState for this same plugin in #1848 before finding this PR and yielding, so I know this state surface well. Beyond reading the diff, I verified:

- The branch applies cleanly on current main, builds, and the package tests pass locally. (The CI run here is from before main's recent churn, so a rebase would be worth it just to re-run the full current check set.)
- `GetActiveSubscribers()` matches the `SubscriberManager` method set in the pinned llm-d-kv-cache version.

On the design choices, agreeing specifically:

- Not enumerating the KV-block index is right -- the indexer interface has no enumeration, and its keys are prompt-derived block hashes, which `docs/plugin_debug.md` excludes. Same conclusion I reached in #1848.
- The speculative-cache keys are `request.RequestID`s (`prerequest.go:190`), not prompt-derived values, so listing them is consistent with the no-payload rule. Reporting only pod identities and dropping the endpoints half of `GetActiveSubscribers` (internal socket addresses) is also the more sanitized choice.
- On the open pluginState question (@liu-cong): the pushback checks out in the code -- entries are deleted at the end of `PreRequest` (`prerequest.go:130`), so a dump would only ever show requests in flight at that instant, and enumerating would need a new snapshot method on the shared pluginState. That reads like a cross-plugin follow-up rather than something to hold this PR on.

One take-it-or-leave-it nit: the `inflightload` reference signals partial dumps with a `truncated` bool while this uses `totalX` + `maxX`. The totals are strictly more informative, but if dumps should stay uniform across plugins for tooling, one convention should win -- fine as a follow-up across the DumpState PRs rather than a change here.

liu-cong (dismissed):

liu-cong:
> there are a number of similar DumpState PRs with a cap on entries set to 100. I suspect that was (at least partially) to cap memory.

@elevran The use case of this is mostly likely calling this DumpState endpoint in a terminal. The main purpose of limiting the response size is to not overwhelming the terminal output. 

I am OK with the current implementation . If we find we need more info or need filtering/querying, we can follow up on that.

thc1006:
Rebased onto main to clear the conflict. Only the kv-cache import paths clashed, since those packages moved into this repo since I opened this, so the code is the same one you approved. Build, `go test -race` and lint are green on the new base.

Sorry about the dismissed approval, @liu-cong, that was the force-push rather than anything new to look at.

liu-cong (approved):

Issue #1843, opened 2026-06-29
Return per-endpoint scores in EPP dynamic metadata


adelsam:
### What
Extend the EPP -> Envoy ext-proc dynamic metadata contract (request path) so the
Endpoint Picker can optionally emit the scheduler's score alongside each returned
endpoint, without changing the existing `x-gateway-destination-endpoint` field that
downstream consumers already depend on.

### Why
The scheduler computes a weighted score per candidate endpoint, but the score is
discarded inside the picker (`ProfileRunResult` carries only `TargetEndpoints []Endpoint`) and never leaves the router. Downstream only sees _which_ endpoints
were chosen, not _how strongly_ they were preferred. Exposing the score enables:
- **Observability** — record why an endpoint was selected and how close the runner-up was, without re-deriving scorer state.
- **Downstream re-ranking / fallback** — gateway providers integrating via the metadata path can make load-aware fallback or hedging decisions over a ranked, scored set.
- **Debuggability** — validate scorer/picker behavior in a live fleet from outside the router process.

### Scope
This is an interface change to a contract shared with Envoy and gateway providers,
so it warrants discussion before code lands. The change spans two layers:

1. Preserve the score out of the picker (additive `ScoredEndpoints` on `ProfileRunResult`; `TargetEndpoints` unchanged).
2. Emit it on the wire as an additive key under the existing `envoy.lb` namespace; `x-gateway-destination-endpoint` stays byte-identical and emission is opt-in.

Non-goals: changing how scores are computed, the request header format, or gateway-side behavior. The response-path metadata from [#1237](https://github.com/llm-d/llm-d-router/pull/1237) (`x-gateway-inference-request-cost`) is unchanged.

### Open questions for discussion
1. Emission control — per-pool config flag (default off), per-request header opt-in, or always on?
1. Which score — post-weight aggregate (already range-enforced to [0,1]) vs. raw per-scorer breakdown?
1. Wire shape — parallel `endpoint -> score` map vs. structured ranked list of `{endpoint, score}`?
1. Number format — any precision/normalization guarantees consumers can rely on?
1. Profiles — surface scores only from the primary profile, or non-primary too?
1. A draft design doc covering current state, data flow, and the proposed options is ready to accompany this (docs/proposals/endpoint-scores-in-dynamic-metadata.md).

Related: [#1237](https://github.com/llm-d/llm-d-router/pull/1237)

elevran:
[...]

PR #1844, opened 2026-06-29
docs: propose per-endpoint scores in EPP dynamic metadata

adelsam:
Draft RFC for extending the EPP dynamic metadata contract so the picker can optionally emit per-endpoint scores alongside the destination endpoint. Tracks llm-d/llm-d-router#1843.

**What type of PR is this?**
/kind documentation

Proposal for expanded EPP interface sharing scores.

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/proposals/endpoint-scores-in-dynamic-metadata.md
+++ b/docs/proposals/endpoint-scores-in-dynamic-metadata.md
@@ -1,0 +1,227 @@
+# Proposal: Return per-endpoint scores in EPP dynamic metadata
+
+| Field | Value |
+| --- | --- |
+| Status | Draft / Request for comments |
+| Tracking issue | [#1843](https://github.com/llm-d/llm-d-router/issues/1843) |
+| Related PR | [#1237](https://github.com/llm-d/llm-d-router/pull/1237) (DynamicMetadata encode-path regression test) |
+| Affected interface | EPP -> Envoy ext-proc dynamic metadata on the request path |
+
+## Summary
+
+The Endpoint Picker (EPP) already communicates the chosen destination endpoint(s)
+to Envoy two ways: as a request header and as an unstructured ext-proc dynamic
+metadata key under the `envoy.lb` namespace. Today that metadata carries only the
+endpoint address(es); the score the scheduler computed for each endpoint is
+discarded inside the picker and never leaves the router.
+
+This proposal adds an opt-in extension to the dynamic metadata contract so EPP can
+emit the picker's score alongside each returned endpoint, without changing the
+existing endpoint-address fields that downstream consumers already depend on.
+
+## Motivation
+
+The scheduler computes a weighted score per candidate endpoint, the picker ranks
+on it, and then collapses the result to an ordered list of endpoints. The score is
+lost at `runPickerPlugin` / `Pick`, so anything downstream of EPP only sees *which*
+endpoints were chosen, not *how strongly* they were preferred.
+
+Exposing the score enables:
+
+- **Observability.** Operators and request tracing can record why an endpoint was
+  selected and how close the runner-up was, without re-deriving scorer state.
+- **Downstream re-ranking and fallback.** Gateway providers that integrate via the
+  metadata path (the integration point that already motivates the dual
+  header/metadata contract) can make load-aware fallback or hedging decisions over
+  a ranked, scored set rather than a flat ordered list.
+- **Debuggability of scoring changes.** Score visibility at the gateway boundary
+  makes it possible to validate scorer/picker behavior in a live fleet from outside
+  the router process.
+
+This is an interface question for the community precisely because dynamic metadata
+is a contract with Envoy and with gateway providers, not an internal detail.
+
+## Current state
+
+### Data flow
+
+1. `SchedulerProfile.runScorerPlugins` produces
+   `weightedScorePerEndpoint map[Endpoint]float64`
+   (`pkg/epp/scheduling/scheduler_profile.go`).
+2. `runPickerPlugin` wraps each entry in a `ScoredEndpoint{Endpoint, Score}` and
+   calls the picker.
+3. The picker (e.g. `MaxScorePicker.Pick`,
+   `pkg/epp/framework/plugins/scheduling/picker/maxscore/picker.go`) ranks the
+   `ScoredEndpoint`s but returns
+   `ProfileRunResult{TargetEndpoints []Endpoint}` -- **the score is dropped here.**
+4. `Director.prepareRequest` (`pkg/epp/requestcontrol/director.go`) reads
+   `result.ProfileResults[primary].TargetEndpoints`, joins the addresses into a
+   comma-separated string, and stores it on `reqCtx.TargetEndpoint`.
+5. `StreamingServer.generateRequestHeaderResponse` /  `generateMetadata`
+   (`pkg/epp/handlers/request.go`) emit the string as both a header and dynamic
+   metadata.
+
+### Wire format today
+
+Request-header response dynamic metadata:
+
+```
+envoy.lb:                                  # metadata.DestinationEndpointNamespace
+  x-gateway-destination-endpoint: "10.0.0.1:8200,10.0.0.2:8200"   # metadata.DestinationEndpointKey (string, comma-joined)
+```
+
+Relevant constants live in `pkg/epp/metadata/consts.go`.
+
+The first structural gap: the score never survives step 3, so **no plumbing change
+at the picker boundary, no metadata change downstream**. Both halves are needed.
+
+## Goals
+
+- Let EPP optionally emit a numeric score per returned endpoint in dynamic metadata.
+- Preserve the existing `x-gateway-destination-endpoint` field and header exactly as
+  they are today (backward compatible; no consumer is forced to change).
+- Keep the scoring path pluggable: a custom picker decides whether/what scores to
+  surface.
+
+## Non-goals
+
+- Changing how scores are *computed* (scorers, weights, `enforceScoreRange`).
+- Changing the request header (`x-gateway-destination-endpoint`) format. Headers
+  are a flat string surface; scores belong in structured metadata only.
+- Defining gateway-side behavior. How a consumer uses the scores is out of scope;
+  this proposal only defines what EPP emits.
+- The response-path metadata (`reqCtx.Response.DynamicMetadata`, e.g.
+  `x-gateway-inference-request-cost`) covered by #1237 is unchanged.
+
+## Design
+
+Two layers change: the in-process result carrying scores out of the picker, and the
+wire format emitted to Envoy.
+
+### 1. Carry scores out of the picker
+
+`ProfileRunResult` currently exposes only `TargetEndpoints []Endpoint`. Add scores
+without breaking existing readers. Preferred option:
+
+```go
+// pkg/epp/framework/interface/scheduling/types.go
+type ProfileRunResult struct {
+    TargetEndpoints []Endpoint
+    // ScoredEndpoints, when set, is parallel to TargetEndpoints and carries the
+    // picker's score for each. Pickers that do not score MAY leave it nil.
+    ScoredEndpoints []ScoredEndpoint
+}
+```
+
+`ScoredEndpoint{Endpoint, Score}` already exists, so pickers like `MaxScorePicker`
+have the scored, sorted slice in hand and only need to stop discarding it. Existing
+consumers that read `TargetEndpoints` keep working unchanged.
+
+Alternatives considered for this layer:
+
+- **Replace `TargetEndpoints` with `[]ScoredEndpoint`.** Cleaner long term but
+  touches every picker, profile handler, and director read site at once. Heavier
+  than the interface question warrants for a first step.
+- **Side channel on the endpoint metadata.** Stuffing the score into
+  `EndpointMetadata` conflates "what the endpoint is" with "what this request's
+  picker thought of it." Rejected.
+
+### 2. Wire format
+
+Emit scores under the existing `envoy.lb` namespace as a new, additive key so the
+endpoint address field is untouched. Two shapes are on the table:
+
+**Option A -- parallel scores map (recommended).** Add a sibling key whose value is
+a struct mapping endpoint address to score:
+
+```
+envoy.lb:
+  x-gateway-destination-endpoint: "10.0.0.1:8200,10.0.0.2:8200"   # unchanged
+  x-gateway-destination-endpoint-scores:                          # new, struct
+    "10.0.0.1:8200": 0.91
+    "10.0.0.2:8200": 0.74
+```
+
+- Pro: existing field byte-identical; consumers opt in by reading the new key.
+- Pro: lookup by address is direct.
+- Con: ordering/ranking is implied by the address list, not the scores struct
+  (structpb maps are unordered).
+
+**Option B -- structured ranked list.** Replace the implied parallelism with an
+explicit ordered list of `{endpoint, score}` entries under a new key, leaving the
+old string field in place for compatibility:
+
+```
+envoy.lb:
+  x-gateway-destination-endpoint: "10.0.0.1:8200,10.0.0.2:8200"   # unchanged
+  x-gateway-destination-endpoints:                                # new, list of structs
+    - endpoint: "10.0.0.1:8200"
+      score: 0.91
+    - endpoint: "10.0.0.2:8200"
+      score: 0.74
+```
+
+- Pro: rank order is explicit and self-contained.
+- Pro: extensible -- future per-endpoint fields (profile name, reason) slot in.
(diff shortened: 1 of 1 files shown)


docs/proposals/endpoint-scores-in-dynamic-metadata.md
```diff
@@ -0,0 +1,227 @@
...
+## Summary
+
+The Endpoint Picker (EPP) already communicates the chosen destination endpoint(s)
+to Envoy two ways: as a request header and as an unstructured ext-proc dynamic
+metadata key under the `envoy.lb` namespace. Today that metadata carries only the
+endpoint address(es); the score the scheduler computed for each endpoint is
+discarded inside the picker and never leaves the router.
+
+This proposal adds an opt-in extension to the dynamic metadata contract so EPP can
+emit the picker's score alongside each returned endpoint, without changing the
```
elevran: nit: the picker does not score endpoints, it picks them based on the scores.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
```diff
@@ -0,0 +1,227 @@
...
+- **Downstream re-ranking and fallback.** Gateway providers that integrate via the
+  metadata path (the integration point that already motivates the dual
+  header/metadata contract) can make load-aware fallback or hedging decisions over
+  a ranked, scored set rather than a flat ordered list.
+- **Debuggability of scoring changes.** Score visibility at the gateway boundary
+  makes it possible to validate scorer/picker behavior in a live fleet from outside
+  the router process.
+
+This is an interface question for the community precisely because dynamic metadata
+is a contract with Envoy and with gateway providers, not an internal detail.
```
elevran: xref the GAIE issue?

docs/proposals/endpoint-scores-in-dynamic-metadata.md
```diff
@@ -0,0 +1,227 @@
...
+2. `runPickerPlugin` wraps each entry in a `ScoredEndpoint{Endpoint, Score}` and
+   calls the picker.
+3. The picker (e.g. `MaxScorePicker.Pick`,
+   `pkg/epp/framework/plugins/scheduling/picker/maxscore/picker.go`) ranks the
+   `ScoredEndpoint`s but returns
+   `ProfileRunResult{TargetEndpoints []Endpoint}` -- **the score is dropped here.**
+4. `Director.prepareRequest` (`pkg/epp/requestcontrol/director.go`) reads
+   `result.ProfileResults[primary].TargetEndpoints`, joins the addresses into a
+   comma-separated string, and stores it on `reqCtx.TargetEndpoint`.
+5. `StreamingServer.generateRequestHeaderResponse` /  `generateMetadata`
```
elevran, line 68: should scored endpoints also propagate back as a header or metadata only?
What are the pros and cons of each approach (e.g., is one or the other more compact, less visible on the wire, ...)?

docs/proposals/endpoint-scores-in-dynamic-metadata.md
```diff
@@ -0,0 +1,227 @@
...
+
+The first structural gap: the score never survives step 3, so **no plumbing change
+at the picker boundary, no metadata change downstream**. Both halves are needed.
+
+## Goals
+
+- Let EPP optionally emit a numeric score per returned endpoint in dynamic metadata.
+- Preserve the existing `x-gateway-destination-endpoint` field and header exactly as
+  they are today (backward compatible; no consumer is forced to change).
+- Keep the scoring path pluggable: a custom picker decides whether/what scores to
```
elevran: what is the rationale for making this a picker option as opposed to having the scheduler always use the ScoredEndpoints map/array?

docs/proposals/endpoint-scores-in-dynamic-metadata.md
```diff
@@ -0,0 +1,227 @@
...
+`ProfileRunResult` currently exposes only `TargetEndpoints []Endpoint`. Add scores
+without breaking existing readers. Preferred option:
+
+```go
+// pkg/epp/framework/interface/scheduling/types.go
+type ProfileRunResult struct {
+    TargetEndpoints []Endpoint
+    // ScoredEndpoints, when set, is parallel to TargetEndpoints and carries the
+    // picker's score for each. Pickers that do not score MAY leave it nil.
+    ScoredEndpoints []ScoredEndpoint
```
elevran: I think TargetEndpoints is a subset of ScoredEndpoints, but is **not** required to include all. For example, a maximum score picker can return a single target.
Pickers perform a linear combination (using scores and weights) on the computed scores. 
Not against extending the struct (might have additional uses), but could benefit from explaining why this is a inidividual picker selection and not a subsystem decision based on the value. Why should the picker be involved?

docs/proposals/endpoint-scores-in-dynamic-metadata.md
```diff
@@ -0,0 +1,227 @@
...
+envoy.lb:
+  x-gateway-destination-endpoint: "10.0.0.1:8200,10.0.0.2:8200"   # unchanged
+  x-gateway-destination-endpoint-scores:                          # new, struct
+    "10.0.0.1:8200": 0.91
+    "10.0.0.2:8200": 0.74
+```
+
+- Pro: existing field byte-identical; consumers opt in by reading the new key.
+- Pro: lookup by address is direct.
+- Con: ordering/ranking is implied by the address list, not the scores struct
```
elevran: Not sure I would classify as a con - this is direct implication of the design decision to support backward compatibility. Gateways are expected to use the `x-gateway-destination-endpoint` metadata/header for decisions and the scores are only used as fallback/reporting (e.g., only use a fallback if it scored considerably better than others).

docs/proposals/endpoint-scores-in-dynamic-metadata.md
elevran, line 1: We currently do not have a `proposals` section in the docs.
@ahg-g @liu-cong @vMaroon - do we want to take this route or use issues/google docs for proposals?

docs/proposals/endpoint-scores-in-dynamic-metadata.md
```diff
@@ -0,0 +1,227 @@
...
+  x-gateway-destination-endpoints:                                # new, list of structs
+    - endpoint: "10.0.0.1:8200"
+      score: 0.91
+    - endpoint: "10.0.0.2:8200"
+      score: 0.74
+```
+
+- Pro: rank order is explicit and self-contained.
+- Pro: extensible -- future per-endpoint fields (profile name, reason) slot in.
+- Con: duplicates the address list across two keys.
```
elevran: nit: is that not also the case in option A?

docs/proposals/endpoint-scores-in-dynamic-metadata.md
```diff
@@ -0,0 +1,227 @@
...
+compatibility), with Option B noted as the path if richer per-endpoint payloads are
+wanted later. Both keep `x-gateway-destination-endpoint` as the source of truth for
+the address list.
+
+New constant in `pkg/epp/metadata/consts.go`, e.g.
+`DestinationEndpointScoresKey = "x-gateway-destination-endpoint-scores"`.
+
+### 3. Plumbing
+
+- `Director.prepareRequest`: when `ScoredEndpoints` is populated, carry the
```
elevran: Unless I'm misreading earlier content (L83, L110-111), this is not aligned with the earlier options of making this a Picker decision. 
I prefer this option.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
```diff
@@ -0,0 +1,227 @@
...
+  `ScoredEndpoints` consistently with `TargetEndpoints`.
+- Unit, metadata layer: extend the `generateMetadata` / encode-path coverage (the
+  surface #1237 added tests for) to assert the new key is present with correct
+  values when scores exist, and absent when they do not.
+- Director: scores propagate from `ProfileRunResult` to `reqCtx` and into the
+  request-header response.
+
+## Open questions
+
+1. **Emission control.** Per-pool config flag, per-request header opt-in, or always
```
elevran: clarification: are you proposing some sort of InferencePool config change (can be a label or attribute) or an EPP configuration? Since this is currently scoped to llm-d (and not GAIE), an EPP config / command line might be more appropriate.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
```diff
@@ -0,0 +1,227 @@
...
+   on? Leaning config flag, default off.
+2. **Which score?** The post-weight aggregate (`weightedScorePerEndpoint`,
+   already range-enforced to [0,1]) vs. raw per-scorer scores. Aggregate is the
+   natural first cut; per-scorer breakdown is a larger contract.
```
elevran: agree - the `weightedScorePerEndpoint` seems aligned with the stated goals.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
```diff
@@ -0,0 +1,227 @@
...
+3. **Wire shape.** Option A vs. Option B above.
+4. **Number format.** structpb only has `double`. Any precision/normalization
+   guarantees consumers should be able to rely on?
```
elevran: what is the potential impact for the stated use cases?

docs/proposals/endpoint-scores-in-dynamic-metadata.md
```diff
@@ -0,0 +1,227 @@
...
+5. **Profiles.** Should scores from non-primary profiles ever be surfaced, or only
+   the primary profile that sets the destination?
```
elevran: While there can be use cases where this makes sense (e.g., selecting P/D pairs for fallback), this would change the wire formats and "bind" gateways to EPP internal and specific configurations that are meaningless - the gateway is only expected to act on the primary endpoint.

elevran:
@adelsam - still working on this?

adelsam:
Yes, sorry for the delay. Revision pushed in `beea1b3f` addressing your review. The main change: scores are now sourced from the full `weightedScorePerEndpoint` map in `SchedulerProfile.Run` rather than from the picker's selection, per your L83/L112/L178 comments. [#2004](https://github.com/llm-d/llm-d-router/pull/2004) will need the same rework, and I'll update GAIE [#2985](https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2985) to match.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
adelsam: You're right. `MaxScorePicker` truncates to `maxNumOfEndpoints` before returning, so "parallel to TargetEndpoints" would have emitted only the selected endpoints' scores — exactly one under the default configuration. That makes the runner-up and fallback use cases motivating this proposal impossible to satisfy.

Reworked as you suggest: `ProfileRunResult` gains `ScoredCandidates`, populated in `SchedulerProfile.Run` from the complete `weightedScorePerEndpoint` map before the picker runs. Pickers need no changes, availability no longer depends on picker configuration, and the unselected candidates survive. The picker-parallel approach is retained as a rejected alternative with that reasoning. This also resolves the inconsistency you flagged at L178.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
adelsam: Thanks, fixed throughout. The doc now says scorers compute the weighted score and the picker ranks and selects on it, in the summary, motivation, and data flow.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
adelsam: Added — [#2985](https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2985) is now in the header table and References. I also split the framing: the metadata key name and value shape belong upstream since the namespace is shared across EPP implementations, while this doc covers the llm-d-router side (score sourcing and the emission knob). As you mention on L208 that the config knob is llm-d-scoped. I'll update [#2985](https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2985) as well, since it still describes scores on the selected endpoints rather than the full candidate set.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
adelsam: Should be Metadata only, added a "Metadata, not header" subsection: dynamic metadata stays inside Envoy while request headers are forwarded to the model-server pod, so a header would leak EPP scoring internals to backends and add bytes to every upstream request; metadata is structured while a header would need scores flattened into a string, introducing a parsing contract on a surface that has none today; and gateway providers already read the endpoint decision from metadata, so the scores land next to the field they qualify.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
adelsam: No good rationale, now dropped. Scores are captured in `SchedulerProfile.Run` from `weightedScorePerEndpoint` for every candidate, unconditionally and for every picker. See the L112 thread for detail. The "keep the scoring path pluggable" goal is gone.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
adelsam: You read it correctly, the sections contradicted each other. It should be: the scheduler profile always records scores, and a config knob controls emission. Detail in the L112 thread.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
adelsam: It should be EPP configuration (flag or config file), default off — not an InferencePool field, as you suggested. Stated explicitly in the Enablement section and removed from the open questions.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
adelsam: Thanks — updated.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
adelsam: Probably not a big deal for the stated use cases. Written up under "Number format" instead of promising a normalization guarantee.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
adelsam: Agreed and closed as a non-goal: per-profile scores would bind gateways to EPP's internal profile configuration, which is meaningless to a consumer acting only on the primary endpoint. The prefill/decode pair fallback case is called out as warranting its own proposal and wire format rather than being folded in here.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
adelsam: Happy either way — the content is also in [#1843](https://github.com/adelsam/llm-d-router/issues/1843) and mirrored upstream in GAIE [#2985](https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2985), so if the project prefers issues or Google Docs I'll close this PR and move the discussion. Just let me know what you settle on with @ahg-g @liu-cong @vMaroon.

docs/proposals/endpoint-scores-in-dynamic-metadata.md
```diff
@@ -0,0 +1,278 @@
...
+
+New constant in `pkg/epp/metadata/consts.go`, e.g.
+`DestinationEndpointScoresKey = "x-gateway-destination-endpoint-scores"`. The key
+name and shape are the shared half of this proposal and are settled in GAIE #2985;
+llm-d-router follows that outcome rather than establishing the format on its own.
+
+#### Number format
+
+`structpb` numbers are doubles and the weighted aggregate is a `float64` already
+range-enforced to [0,1], so EPP introduces no precision loss. Observability and
```
elevran, line 218: nit: inaccurate claim...
The weighted aggregate is not range-enforced to [0,1]. Each individual scorer's score is clamped to [0,1] before weighting, but the sum across scorers is not bounded
The precision argument for structpb doubles still holds regardless of magnitude...

elevran:
@adelsam the design is approved. 
I had discussed it with other maintainers and we will not currently adopt a `docs/proposal` directory as it requires ongoing work to keep it up to date.
Using a PR with the markdown proposal, as you did here, seems to be a good middle ground for allowing feedback (line by line, unlike an issue) and maintaining some historical references for the final code PR that implemenets the proposal. Once implemented, the code becomes the upt to date source of truth.

Thus, the **design is approved** despite **closing this PR**. 
You may proceed with aligning #2004 with it - I had removed the correspnding `hold` label from #2004.
Note that PR #2004 has a failing test that also needs fixing.

elevran:
Design approved. Work will be done in #2004 or followups.

PR #1856, opened 2026-07-01
feat: consume InFlightLoadProducer in predicted-latency producer

kaushikmitr:
The predicted-latency producer maintained its own prefillTokensInFlight counter keyed by endpoint, accumulating the full prompt token count with no discount for the already-cached prefix. The InFlightLoadProducer already tracks per-endpoint in-flight load, discounts the cached prompt prefix, and exposes an active request count, with robust release and rollback handling.

Consume that producer's InFlightLoad attribute instead: prefill_tokens_in_flight now reflects the uncached prefill work in flight, and num_request_running is sourced from the same producer (falling back to vLLM metrics when the attribute is absent). This removes the bespoke counter along with its increment, decrement, and TTL-eviction bookkeeping.
With the agentic workload (in the predicted latency guide on GPU) this gives better performance than earlier
image


pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/requestcontrol_hooks.go
```diff
@@ -96,14 +99,15 @@ func (pl *PredictedLatency) PreRequest(ctx context.Context, request *fwksched.In
...
-		pl.endpointCounter(&pl.prefillTokensInFlight, prefillPodKey).Add(int64(predictedLatencyCtx.inputTokenCount))
-		predictedLatencyCtx.prefillTokensAtDispatchOnPrefill = pl.endpointCounter(&pl.prefillTokensInFlight, prefillPodKey).Load()
+	// Snapshot the in-flight load (cache-discounted prefill tokens and active
+	// request count) the request saw at dispatch, sourced from the
+	// InFlightLoadProducer. These become the prefill-tokens-in-flight and
+	// num-request-running training features. The request count falls back to
+	// vLLM metrics when the attribute is absent.
+	pl.snapshotInFlightLoad(decodeEndpoint, &predictedLatencyCtx.prefillTokensAtDispatch, &predictedLatencyCtx.requestsAtDispatch)
+	if prefillEndpoint != nil {
+		pl.snapshotInFlightLoad(prefillEndpoint, &predictedLatencyCtx.prefillTokensAtDispatchOnPrefill, &predictedLatencyCtx.requestsAtDispatchOnPrefill)
```
BenjaminBraunDev: I think there's an ordering problem here. The `InFlightLoad` attribute is not a stored value, reading it returns whatever is in the producer's tracker at that moment. The producer adds this request's tokens to those trackers in its own PreRequest hook. So the value we snapshot here depends on which plugin's PreRequest runs first: if the producer ran first, the snapshot includes this request's own tokens, if not, it doesn't.

This is an issue since `plugin.DataDependencies` only orders the Produce hooks (`OrderDataProducerPlugins` in `request_control_config.go`). The PreRequest list is built from `GetAllPlugins()`, which iterates a map, so the order is random. The old counter code didn't have this problem because it incremented and read its own atomic in one place.

I suggest taking the snapshot during Produce instead of here. generatePredictions already reads the load for every candidate endpoint there, and Produce is DAG-ordered and always runs before the commit, so the value is well defined. If PreRequest just reuses the value we already read for the winning endpoint, the training features stop depending on hook order and also match the prediction features exactly.

It might be worth addressing the random ordering of non-Produce hooks in a separate issue, but for this PR just make the snapshot not depend on it.

pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/requestcontrol_hooks.go
```diff
@@ -195,25 +190,13 @@ func (pl *PredictedLatency) ResponseBody(ctx context.Context, request *fwksched.
 				)
 				entry.PrefillTokensInFlight = predictedLatencyCtx.prefillTokensAtDispatch
 				entry.DecodeTokensInFlight = predictedLatencyCtx.decodeTokensAtDispatch
+				entry.NumRequestRunning = predictedLatencyCtx.requestsAtDispatch
```
BenjaminBraunDev, line 200: This gives us two definitions of NumRequestRunning. Training entries now use the router's in-flight tracker, but mid-stream TPOT predictions in buildPredictionRequest still use the scraped vLLM gauge. The tracker counts requests this EPP has committed (including queued ones, and only its own traffic), the gauge is the engine's actual running batch. The model trains on one and gets queried with the other. 

Should we consolidate this? If we use the tracker, the mid-stream path can read the live attribute off the endpoints already stored in the ctx, same way refreshLastSeenMetrics gets fresh metrics, should be a small change. If you think we should predict with the gauge, then training should probably use it too.

pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/requestcontrol_hooks.go
kaushikmitr: this was a subtle bug, thanks for catching this. I fixed it.

pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/requestcontrol_hooks.go
kaushikmitr: I think mid request tpot predictions (only used for observility) has many inconsistencies and noisy and not very useful. Also turned off by default. Let me create a follow up issue to remove it alteogher.

kaushikmitr: https://github.com/llm-d/llm-d-router/issues/2016

BenjaminBraunDev: Sounds good, thanks for making that issue

pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/requestcontrol_hooks.go
BenjaminBraunDev: Made an issue for plugin ordering generally #2040

ahg-g (approved):

Issue #1706, opened 2026-06-22
feat: GPU utilization-aware routing via DCGM metrics, utilization filter and scorer


szedan-rh:
## Problem

The EPP scores endpoints based on KV cache utilization, queue depth, running requests, and prefix cache affinity. These are application-level metrics from the model server. There's no direct GPU hardware utilization awareness — a pod can have moderate KV cache usage but high GPU compute/memory pressure. This can lead to routing requests to pods near GPU saturation, causing latency spikes.

Model servers (vLLM, SGLang) do not expose GPU compute utilization (SM occupancy) or total VRAM usage — only KV cache block occupancy within pre-allocated memory.

## Proposal

Add GPU utilization awareness to the scheduling pipeline via three components:

1. **DCGM data source** — a new data source plugin that scrapes NVIDIA DCGM Exporter metrics from a per-pod sidecar or node-level DaemonSet. Collects `DCGM_FI_DEV_GPU_UTIL` (compute utilization) and maps it to a new `GPUUtilizationPercent` field on the `Metrics` struct. This is a separate scrape target from the model server's `/metrics` endpoint.

2. **`gpu-utilization-filter`** — hard cutoff that excludes pods above a configurable GPU utilization threshold (default: 90%). Falls back to passing all pods through if every candidate exceeds the threshold. Runs early in the filter chain, before affinity-based filters.

3. **`gpu-utilization-scorer`** — Distribution-category scorer that prefers pods with more GPU headroom (`score = 1 - utilization`). Complements existing KV cache and queue-based scorers.

## Scope

- New DCGM metrics data source plugin (scrapes DCGM Exporter Prometheus endpoint)
- Extend `Metrics` struct with `GPUUtilizationPercent`
- Implement filter and scorer plugins following existing patterns
- Requires DCGM Exporter deployed alongside model-serving pods
- NVIDIA-specific; other accelerators (AMD, Intel) would need equivalent metric sources

szedan-rh:
@elevran - Could you please look into this issue and confirm is making sense to have this feature ?

elevran:
This seems like useful new information.
Would be good to have some empirical support in favor of including this.

A few specific items to consider:
- is the DGCM network endpoint per node or per endpoint? How discovered from Endpoints/Pods?
- how frequently should DGCM be scarped? Every 50ms like nedpoint metrics or something else?
- Cosnider resuing exisitng EPP datalayer mechanisms:
[...]

PR #1857, opened 2026-07-01
file-discovery: add per-endpoint metricsAddress/metricsPort

hexfusion:
**What type of PR is this?**

/kind feature

Lets a `file-discovery` endpoint declare a metrics scrape target (`metricsAddress` / `metricsPort`) distinct from its serving `address`/`port`. The scrape host may be a hostname, unlike the serving address (validated as IPv4).

This supports cross-cluster routing: traffic forwards to an endpoint IP while aggregate metrics are scraped from a separate host. Both fields default to the serving `address`/`port` when unset; `metricsAddress` rejects URL metacharacters and `metricsPort` is range-checked (`ErrInvalidMetricsPort`).

N/A

**Release note**:
```release-note
file-discovery: endpoints can set `metricsAddress`/`metricsPort` to scrape metrics from a host distinct from the serving address.
```

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
+++ b/pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"sort"
 	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/fsnotify/fsnotify"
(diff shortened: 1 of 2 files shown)


pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
```diff
@@ -246,6 +256,22 @@ func (f *FileDiscovery) load(notifier fwkdl.DiscoveryNotifier) error {
 			errs = append(errs, fmt.Errorf("endpoint %q: invalid port %q", e.Name, e.Port))
 			continue
 		}
+		metricsAddr, metricsPort := e.Address, e.Port
+		if e.MetricsAddress != "" {
+			// MetricsAddress lands in the scrape URL host; reject URL metacharacters.
+			if strings.ContainsAny(e.MetricsAddress, "/@#? \t") {
```
vMaroon: nit: the blocklist rejects `\t` and space but not `\r/\n`.

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
```diff
@@ -246,6 +256,22 @@ func (f *FileDiscovery) load(notifier fwkdl.DiscoveryNotifier) error {
...
+				errs = append(errs, fmt.Errorf("endpoint %q: %w %q", e.Name, ErrInvalidMetricsAddress, e.MetricsAddress))
+				continue
+			}
+			metricsAddr = e.MetricsAddress
+		}
+		if e.MetricsPort != "" {
+			if portNum, err := strconv.Atoi(e.MetricsPort); err != nil || portNum < 1 || portNum > 65535 {
```
vMaroon: nit: this port range check is now identical to the serving-port one above (L255). Could pull out a small validatePort(string) error and call it in both spots.

elevran:
@hexfusion 
Understand the need, would like to suggest abother approach:

- having a solution that works across all endpoints, whether file generated or discovered is required.
- it might be cleaner to
   - define a new DataSource for your use case that uses labels for port identification
   - can reuse the HTTPDataSource and metrics extractor if that helps (reusable, same metrics names)
   - see https://github.com/llm-d/llm-d-router/issues/1706#issuecomment-4853793124 which identifies similar requirements (configurable port in HTTP and longer scarping period)
   - may also want to use the disable metrics flag in the config if the built-in metrics collector/extractors are not appropriate in your use case.

hexfusion:
closing in favor of https://github.com/llm-d/llm-d-router/issues/1903

PR #1858, opened 2026-07-01
datalayer: TLS options (CA verify + mTLS client cert) for metrics scrape

hexfusion:
**What type of PR is this?**

/kind feature

Adds optional TLS options to the metrics and models data sources (via a `TLSOptions` struct) so an operator can secure the scrape over https:

- `caCertPath` verifies the scrape target's server certificate against a supplied PEM CA bundle, instead of relying on `insecureSkipVerify`.
- `clientCertPath` / `clientKeyPath` present a client certificate for mTLS, so the target authenticates the scraper by certificate rather than a bearer token.

Together these enable a mutually-authenticated scrape (e.g. scraping a target that fronts its metrics with mTLS and a private/service CA) with no shared credential. Empty values preserve existing behavior (system CA pool, honoring `insecureSkipVerify`, no client certificate).

**Usage**

Set the options under the `metrics-data-source` plugin's `parameters` (same fields are available on `models-data-source`):

```yaml
plugins:
  - type: metrics-data-source
    parameters:
      scheme: https
      path: /metrics
      caCertPath: /etc/epp/tls/ca.crt        # verify target server cert (bundle OK)
      clientCertPath: /etc/epp/tls/tls.crt   # present client cert -> mTLS
      clientKeyPath: /etc/epp/tls/tls.key
```

Mount the cert/key/CA (e.g. a cert-manager-issued Secret) into the EPP pod at those paths.

**Note on certificate rotation:** the certificate is loaded once when the data source is constructed, so a rotated cert is not picked up until the EPP restarts. For short-TTL rotation, bounce the EPP on renewal (e.g. a reloader annotation) or use longer-lived certs. Reload-on-change would be a reasonable follow-up.

N/A

**Release note**:
```release-note
metrics/models data sources: add optional `caCertPath`, `clientCertPath`, and `clientKeyPath` to verify the scrape target's server certificate against a CA bundle and present a client certificate for mTLS.
```

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/source/http/datasource.go
+++ b/pkg/epp/framework/plugins/datalayer/source/http/datasource.go
@@ -19,11 +19,13 @@ package http
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"io"
 	"net/http"
 	"net/url"
+	"os"
 	"reflect"
 	"runtime/debug"
 	"slices"
@@ -60,8 +62,23 @@ type HTTPDataSource[T any] struct {
 	exts []fwkdl.PollingExtractor[T]
 }
 
-// NewHTTPDataSource constructs a typed polling dispatcher.
-func NewHTTPDataSource[T any](scheme, path string, skipCertVerification bool,
+// TLSOptions configures the https transport. The zero value verifies the target
+// against the system CA pool with no client certificate.
+type TLSOptions struct {
+	// SkipVerify disables verification of the target's server certificate.
+	SkipVerify bool
+	// CACertPath is a PEM CA bundle used to verify the target instead of the
+	// system pool. Ignored when SkipVerify is set.
+	CACertPath string
+	// ClientCertPath and ClientKeyPath present a client certificate for mTLS.
+	// Both must be set together.
+	ClientCertPath string
+	ClientKeyPath  string
+}
+
+// NewHTTPDataSource constructs a typed polling dispatcher. For https, tlsOpts configures
+// server verification (CACertPath) and optional mTLS (ClientCertPath/ClientKeyPath).
+func NewHTTPDataSource[T any](scheme, path string, tlsOpts TLSOptions,
 	pluginType, pluginName string, parser func(io.Reader) (T, error)) (*HTTPDataSource[T], error) {
 	if scheme != "http" && scheme != "https" {
 		return nil, fmt.Errorf("unsupported scheme: %s", scheme)
@@ -73,8 +90,12 @@ func NewHTTPDataSource[T any](scheme, path string, skipCertVerification bool,
 		},
 	}
 	if scheme == "https" {
+		tlsCfg, err := tlsClientConfig(tlsOpts)
+		if err != nil {
+			return nil, err
+		}
 		httpsTransport := baseTransport.Clone()
-		httpsTransport.TLSClientConfig = &tls.Config{InsecureSkipVerify: skipCertVerification}
+		httpsTransport.TLSClientConfig = tlsCfg
 		cl.Transport = httpsTransport
 	}
 	return &HTTPDataSource[T]{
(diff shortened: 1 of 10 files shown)


pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -60,8 +62,9 @@ type HTTPDataSource[T any] struct {
 	exts []fwkdl.PollingExtractor[T]
 }
 
-// NewHTTPDataSource constructs a typed polling dispatcher.
-func NewHTTPDataSource[T any](scheme, path string, skipCertVerification bool,
+// NewHTTPDataSource constructs a typed polling dispatcher. For https, an
+// optional caCertPath PEM CA bundle verifies the scrape target's cert.
```
shmuelk: If you are adding documentation on the caCertPath parameter also add documentation on the skipCertVerification parameter

hexfusion: updated @shmuelk thanks

hexfusion:
related to #1903

ahg-g:
Is this ready to merge?

pkg/epp/framework/plugins/datalayer/source/models/datasource.go
```diff
@@ -28,13 +28,15 @@ type modelsDatasourceParams struct {
 	Path string `json:"path"`
 	// InsecureSkipVerify defines whether model server certificate should be verified or not.
 	InsecureSkipVerify bool `json:"insecureSkipVerify"`
+	// CACertPath is an optional PEM CA bundle to verify the scrape target cert.
+	CACertPath string `json:"caCertPath"`
```
elevran, line 32: PR body and YAML example say `clientCertPath` / `clientKeyPath` are available on `models-data-source`, but this struct only adds `CACertPath`. Add `ClientCertPath` / `ClientKeyPath` to `modelsDatasourceParams` and wire them through `ModelDataSourceFactory`

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -60,8 +62,23 @@ type HTTPDataSource[T any] struct {
...
 
-// NewHTTPDataSource constructs a typed polling dispatcher.
-func NewHTTPDataSource[T any](scheme, path string, skipCertVerification bool,
+// TLSOptions configures the https transport. The zero value verifies the target
+// against the system CA pool with no client certificate.
+type TLSOptions struct {
+	// SkipVerify disables verification of the target's server certificate.
+	SkipVerify bool
+	// CACertPath is a PEM CA bundle used to verify the target instead of the
+	// system pool. Ignored when SkipVerify is set.
```
elevran, line 71: nit: the comment says `CACertPath` is "Ignored when SkipVerify is set", but the implementation still reads and parses the file, builds the pool, and assigns it to `cfg.RootCAs`. Behavior is correct, but the file load is wasted work and contradicts the comment. Suggest short-circuit when constructing the config

```go
if !opts.SkipVerify && opts.CACertPath != "" {
    pool, err := caCertPool(opts.CACertPath)
    ...
    cfg.RootCAs = pool
}
```

elevran (commented):
The new `TLSOptions` struct is a clean replacement for the `skipCertVerification bool` flag. Cert rotation is explicitly out of scope and called out in the description.

Left two inline comments above. The asymmetry between metrics and models is the main one to settle before merge. The other is a nit.

`TestMetricsDataSourceFactory_TLS` is the only factory-level coverage on this PR. A symmetric `TestModelDataSourceFactory_TLS` would catch plumbing regressions on the models side. Optional - fine wth filing a follow-up rather than blocking on it, but if you change the parameters for models in this might be more prudent to include in this PR.

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
hexfusion: updated, thanks

elevran (approved):

PR #1871, opened 2026-07-01
Adding standard deviation metrics for monitoring epp pool load variance

### LINKED ISSUE #1870: Additional variance metrics to track InferencePool health

Currently, the EPP scheduler in LLM-d-router project exposes some [metrics to track the pool health](https://llm-d.ai/docs/architecture/core/router/epp/scheduling#pool-health-metrics). These include average kv cache utilization, average number of pending requests and average number of running requests. This is a feature request to add a new set of variance metrics on the inference pool that would help track the differences in load between the replicas within an inference pool.

**Why is this needed**:
Within the EPP, utilizing scheduling profiles, users are able to set weights to various scorer plugins. But the efficiency of intelligent routing and weights applied to different plugins may directly depends on the variance in load between replicas in the same inference pool. Infact, in an extreme case, where there is absolutely no variance, random/round robin routing might be sufficient in comparison to a very intelligent scorer weighting that may be an overkill. Hence, it is vital to understand the variance in load on replicas within the various InferencePools deployed for production workloads.

Along with the average values for kv cache utilization, waiting and active requests, it would be useful to see a standard deviation metric within an inference pool that would help observe the variance within an inference pool over time. And hence the following additional metrics are requested:

```
llm_d_epp_std_dev_kv_cache_utilization
llm_d_epp_std_dev_queue_size
llm_d_epp_std_dev_running_requests
```

SachinVarghese:
**What type of PR is this?**

/kind feature

Currently, the EPP scheduler in LLM-d-router project exposes some [metrics to track the pool health](https://llm-d.ai/docs/architecture/core/router/epp/scheduling#pool-health-metrics). These include average kv cache utilization, average number of pending requests and average number of running requests across replicas within a pool.  Within the EPP, utilizing scheduling profiles, users are able to set weights to various scorer plugins. But the efficiency of intelligent routing and weights applied to different plugins may directly depends on the variance in load between replicas in the same inference pool. Infact, in an extreme case, where there is absolutely no variance, random/round robin routing might be sufficient and a very intelligent scorer weighting may be an overkill. Hence, it is vital to understand the variance in load on replicas within the various InferencePools deployed.

Fixes #1870

```release-note
Adding standard deviation metrics for monitoring epp pool variance
```

### DIFF
--- a/pkg/epp/datalayer/logger/logger.go
+++ b/pkg/epp/datalayer/logger/logger.go
@@ -123,7 +123,6 @@ func refreshPrometheusMetrics(logger logr.Logger, datastore datalayer.PoolInfo, 
 	metrics.RecordInferencePoolAvgKVCache(pool.Name, summary.kvCache.mean)
 	metrics.RecordInferencePoolAvgQueueSize(pool.Name, summary.queueSize.mean)
 	metrics.RecordInferencePoolAvgRunningRequests(pool.Name, summary.runningRequests.mean)
-
 	metrics.RecordInferencePoolStdDevKVCache(pool.Name, summary.kvCache.stdv)
 	metrics.RecordInferencePoolStdDevQueueSize(pool.Name, summary.queueSize.stdv)
 	metrics.RecordInferencePoolStdDevRunningRequests(pool.Name, summary.runningRequests.stdv)
@@ -165,9 +164,11 @@ func calculateSummary(endpoints []fwkdl.Endpoint) summary {
 	size := float64(len(endpoints))
 	totals := calculateTotals(endpoints)
 
-	result.kvCache.mean = totals.kvCache / size
-	result.queueSize.mean = float64(totals.queueSize) / size
-	result.runningRequests.mean = float64(totals.runningRequests) / size
+	if size > 0 {
+		result.kvCache.mean = math.Round((totals.kvCache/size)*100) / 100
+		result.queueSize.mean = math.Round((float64(totals.queueSize)/size)*100) / 100
+		result.runningRequests.mean = math.Round((float64(totals.runningRequests)/size)*100) / 100
+	}
 
 	for _, pod := range endpoints {
 		metrics := pod.GetMetrics()
@@ -180,13 +181,15 @@ func calculateSummary(endpoints []fwkdl.Endpoint) summary {
 	if size > 2 {
 		num = size - 1
 	}
-	result.kvCache.vrce /= num
-	result.queueSize.vrce /= num
-	result.runningRequests.vrce /= num
+	if size > 0 {
+		result.kvCache.vrce = math.Round((result.kvCache.vrce/num)*100) / 100
+		result.queueSize.vrce = math.Round((result.queueSize.vrce/num)*100) / 100
+		result.runningRequests.vrce = math.Round((result.runningRequests.vrce/num)*100) / 100
 
-	result.kvCache.stdv = math.Sqrt(result.kvCache.vrce)
-	result.queueSize.stdv = math.Sqrt(result.queueSize.vrce)
-	result.runningRequests.stdv = math.Sqrt(result.runningRequests.vrce)
+		result.kvCache.stdv = math.Round(math.Sqrt(result.kvCache.vrce)*100) / 100
+		result.queueSize.stdv = math.Round(math.Sqrt(result.queueSize.vrce)*100) / 100
+		result.runningRequests.stdv = math.Round(math.Sqrt(result.runningRequests.vrce)*100) / 100
+	}
 
 	return result
 }
--- a/pkg/epp/datalayer/logger/logger.go
+++ b/pkg/epp/datalayer/logger/logger.go
@@ -150,7 +150,6 @@ func calculateTotals(endpoints []fwkdl.Endpoint) totals {
 type stats struct {
 	mean float64 // average
 	stdv float64 // standard deviation
-	vrce float64 // variance
 }
 
 type summary struct {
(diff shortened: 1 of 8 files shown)


pkg/epp/metrics/testdata/llm_d_kv_cache_std_dev_metrics
```diff
@@ -0,0 +1,3 @@
+# HELP llm_d_epp_std_dev_kv_cache_utilization [ALPHA] The standard deviation kv cache utilization for an inference server pool.
+# TYPE llm_d_epp_std_dev_kv_cache_utilization gauge
+llm_d_epp_std_dev_kv_cache_utilization{name="p1} 0.1
```
albertoperdomo2: This is missing a `"` here: `{name="p1"}`.

pkg/epp/datalayer/logger/logger.go
```diff
@@ -141,3 +145,61 @@ func calculateTotals(endpoints []fwkdl.Endpoint) totals {
...
+	if size > 2 {
+		num = size - 1
+	}
+
+	result.kvCache.vrce /= num
+	result.queueSize.vrce /= num
+	result.runningRequests.vrce /= num
+
+	// Round stats to two decimal places
+	result.kvCache.mean = math.Round(result.kvCache.mean*100) / 100
```
albertoperdomo2: Should we have something like this:

```
func round2(x float64) float64 { return math.Round(x*100) / 100 }
```

Instead of the `math.Round(x*100)/100` repeated pattern?

pkg/epp/datalayer/logger/logger.go
```diff
@@ -141,3 +145,61 @@ func calculateTotals(endpoints []fwkdl.Endpoint) totals {
...
+	result.queueSize.mean = math.Round(result.queueSize.mean*100) / 100
+	result.runningRequests.mean = math.Round(result.runningRequests.mean*100) / 100
+
+	result.kvCache.vrce = math.Round((result.kvCache.vrce)*100) / 100
```
albertoperdomo2: I think that rounding the variance before taking `sqrt` compounds error. Variance is not an output, only `stdv` is, so rounding `800.004` to `800.0` before `math.Sqrt` changes the `stdv` result in the lower decimal places. I would only round the final outputs (`mean` and `stdv`).

pkg/epp/datalayer/logger/logger.go
```diff
@@ -141,3 +145,61 @@ func calculateTotals(endpoints []fwkdl.Endpoint) totals {
 	}
 	return result
 }
+
+// stats holds aggregated metric values
+type stats struct {
+	mean float64 // average
+	stdv float64 // standard deviation
+	vrce float64 // variance
```
albertoperdomo2: Variance is only used to compute `stdv`,  it's not reported to Prometheus and no caller outside the test reads it. Do we want to expose it in the struct? And do we want to check it in the tests or only check the reported `stdv` and `mean` fields?

pkg/epp/datalayer/logger/logger.go
```diff
@@ -141,3 +145,61 @@ func calculateTotals(endpoints []fwkdl.Endpoint) totals {
...
+	result.runningRequests.mean = float64(totals.runningRequests) / size
+
+	for _, pod := range endpoints {
+		metrics := pod.GetMetrics()
+		result.kvCache.vrce += (metrics.KVCacheUsagePercent - result.kvCache.mean) * (metrics.KVCacheUsagePercent - result.kvCache.mean)
+		result.queueSize.vrce += (float64(metrics.WaitingQueueSize) - result.queueSize.mean) * (float64(metrics.WaitingQueueSize) - result.queueSize.mean)
+		result.runningRequests.vrce += (float64(metrics.RunningRequestsSize) - result.runningRequests.mean) * (float64(metrics.RunningRequestsSize) - result.runningRequests.mean)
+	}
+
+	var num = 1.0
```
albertoperdomo2: I would rewrite this as:

```
denominator := math.Max(1, size-1)
```

Which I think is clearer.

SachinVarghese:
@albertoperdomo2 thanks for the review. I have now made the requested changes in the PR.

pkg/epp/datalayer/logger/logger.go
```diff
@@ -141,3 +145,51 @@ func calculateTotals(endpoints []fwkdl.Endpoint) totals {
...
+
+	if size < 1 {
+		return result
+	}
+
+	result.kvCache.mean = totals.kvCache / size
+	result.queueSize.mean = float64(totals.queueSize) / size
+	result.runningRequests.mean = float64(totals.runningRequests) / size
+
+	for _, pod := range endpoints {
```
albertoperdomo2, line 150: Here, I think the `result.kvCache.stdv` is misleading, because at this point, within the loop, it is just summing squares and it does not contain the stddev itself. I would use local variables for the Square Sums within the loop:

```
var kvSS, queueSS, reqSS float64
```

and use that to compute the exported variables:

```
result.kvCache.stdv = round2(math.Sqrt(kvSS / sampleSize))
result.queueSize.stdv = round2(math.Sqrt(queueSS / sampleSize))
result.runningRequests.stdv = round2(math.Sqrt(reqSS / sampleSize))
```

pkg/epp/datalayer/logger/logger.go
```diff
@@ -141,3 +145,51 @@ func calculateTotals(endpoints []fwkdl.Endpoint) totals {
...
+	queueSize       stats
+	runningRequests stats
+}
+
+func calculateSummary(endpoints []fwkdl.Endpoint) summary {
+	var result summary
+	size := float64(len(endpoints))
+	totals := calculateTotals(endpoints)
+
+	if size < 1 {
```
albertoperdomo2: The guard `size < 1` on the `size` is indirect, I would guard before the conversion to **float64**:

```
if len(endpoints) == 0 {
  return summary{}
}
```

pkg/epp/datalayer/logger/logger.go
```diff
@@ -141,3 +145,51 @@ func calculateTotals(endpoints []fwkdl.Endpoint) totals {
 	}
 	return result
 }
```
albertoperdomo2: `calculateSummary` calls `calculateTotals` (first pass for sums), then loops again for variance. At EPP scale, I do not think this matters for performance, but the two passes exist only because `calculateTotals` is a separate function called from another place. If `calculateSummary` folded the mean computation into its own single pass, the previous issue of using the stdv as an accumulator and the two pass could be resolved together, so I would approach it that way.

SachinVarghese: Removed `calculateTotals` function in dfa533a

pkg/epp/datalayer/logger/logger.go
SachinVarghese: Fixed in [5a3a9cd](https://github.com/llm-d/llm-d-router/pull/1871/commits/5a3a9cdfcdf9a656f875545df61746c1c1c85320)

pkg/epp/datalayer/logger/logger.go
SachinVarghese: Fixed in [5a3a9cd](https://github.com/llm-d/llm-d-router/pull/1871/commits/5a3a9cdfcdf9a656f875545df61746c1c1c85320)

albertoperdomo2:
Thanks for your contribution! LGTM 👍🏽

ahg-g (approved):

PR #1865, opened 2026-07-01
migrate: github.com/llm-d/coordinator core @ 876a357

roytman:
Migrates the following paths from github.com/llm-d/coordinator (@ 876a357) with full git history:

```
-  pkg/common/  → pkg/coordinator/common 
-  pkg/config → pkg/coordinator/config 
-  pkg/connectors → pkg/coordinator/connectors 
-  pkg/gateway → pkg/coordinator/gateway 
-  pkg/pipeline → pkg/coordinator/pipeline 
-  pkg/server → pkg/coordinator/server 
-  Makefile → Makefile.coord.mk 
-  README.md → README.coord.md 
-  Dockerfile.coordinator → Dockerfile.coordinator 
-  cmd/coordinator → cmd/coordinator
```

**Merge via "Create a merge commit"** — squash or rebase discards the migrated history.

PR #1872, opened 2026-07-01
migrate: github.com/llm-d/coordinator @ 876a357

### LINKED ISSUE #1875: Migrate llm-d-incubation/coordinator  into llm-d-router
## Summary

Consolidate the [github.com/llm-d-incubation/coordinator](https://github.com/llm-d-incubation) repository into `llm-d-router`,
preserving full git history, so the coordinator and the router live and evolve
in one repo.

## Motivation

The coordinator and the router are developed and released together.
Keeping them in separate repositories forces cross-repo dependency bumps and
splits review and CI. A single repo removes that coordination overhead.

## Scope

Import from `github.com/llm-d/coordinator` (@876a357) with full history, under
namespaced paths to avoid collision with existing router code:

| Source (coordinator) | Destination (llm-d-router) |
| --- | --- |
| `pkg` | `pkg/coordinator` |
| `cmd/coordinator` | `cmd/coordinator` |
| `config` | `config/coordinator` |
| `deploy` | `deploy/coordinator` |
| `test` | `test/coordinator` |
| `Makefile` | `Makefile.coord.mk` |
| `README.md` | `README.coord.md` |
| `Dockerfile.coordinator` | `Dockerfile.coordinator` |
| `docs/architecture.md` | `docs/coordinator_architecture.md` |
| `docs/communication.md` | `docs/communication.md` |

Go import paths are rewritten from `github.com/llm-d/coordinator/pkg/...` to
`github.com/llm-d/llm-d-router/pkg/coordinator/...`. New dependencies pulled in
with the migrated code: `go-chi/chi/v5`, `spf13/viper` (plus their indirect
deps).

## Out of scope (follow-up issues)

- Wiring `Makefile.coord.mk` into the root `Makefile`.
- Adding CI coverage (build/test) for the coordinator packages.
- Any behavioral changes to the migrated code.

## Acceptance criteria

- [ ] All listed paths present under their new locations with history intact.
- [ ] `lint` and `test` pass on the migrated Go packages.
- [ ] Import paths reference `pkg/coordinator/...`, including in docs/README examples.
- [ ] Merge preserves history (merge commit, not squash/rebase).

## Notes

- Merge via "Create a merge commit"; squash or rebase discards migrated history.
- The signed-commits CI gate will fail because upstream history is unsigned;
  this needs a documented exception or admin merge for history-migration PRs.

roytman:
**What type of PR is this?**

/kind feature

Consolidate the github.com/llm-d/coordinator repository into llm-d-router,
preserving full git history, so the coordinator and the router live and evolve
in one repo.

The coordinator and the router are developed and released together.
Keeping them in separate repositories forces cross-repo dependency bumps and
splits review and CI. A single repo removes that coordination overhead.

Migrates the following paths from `github.com/llm-d/coordinator` (@ 876a357) with full git history:

- `pkg` → `pkg/coordinator`
- `Makefile` → `Makefile.coord.mk`
- `README.md` → `README.coord.md`
- `Dockerfile.coordinator` → `Dockerfile.coordinator`
- `cmd/coordinator` → `cmd/coordinator`
- `config` → `config/coordinator`
- `docs/architecture.md` → `docs/coordinator_architecture.md`
- `docs/communication.md` → `docs/communication.md`
- `deploy` → `deploy/coordinator`
- `test` → `test/coordinator`

To pick up future upstream changes: `migrate-gaie-paths.sh --since 876a357 pkg pkg/coordinator Makefile Makefile.coord.mk README.md README.coord.md Dockerfile.coordinator Dockerfile.coordinator cmd/coordinator cmd/coordinator docs/architecture.md docs/coordinator_architecture.md docs/communication.md docs/communication.md deploy deploy/coordinator test test/coordinator`

**Merge via "Create a merge commit"** — squash or rebase discards the migrated history.

Which issue(s) this PR fixes:
fixes #1875


elevran:
DCO and CI should be fixed. We can ignore the signed commits.
upstream/main merged a fix for the markdown link checks which you may want to pick up by rebasing on top.

elevran:
Does this replace #1865 ?

roytman:
> DCO and CI should be fixed. We can ignore the signed commits. upstream/main merged a fix for the markdown link checks which you may want to pick up by rebasing on top.

Yes, I'm working on that, and it will replace #1865 
Set it as a draft for now

Dockerfile.coordinator
```diff
@@ -0,0 +1,47 @@
...
+RUN go mod download
+
+# Copy the go source
+COPY cmd/ cmd/
+COPY pkg/ pkg/
+COPY configs/ configs/
+
+RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build \
+    -ldflags="${LDFLAGS} -X github.com/llm-d/coordinator/pkg/version.CommitSHA=${COMMIT_SHA} -X github.com/llm-d/coordinator/pkg/version.BuildRef=${BUILD_REF}" \
+    -o bin/coordinator ./cmd/coordinator/...
```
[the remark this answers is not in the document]
roytman: Fixed in `0a0828d0`. 
Updated the build-stage copy to COPY `config/coordinator/ config/coordinator/` and dropped the -X ldflags entirely: the coordinator binary has no version package, so `github.com/llm-d/coordinator/pkg/version`.{CommitSHA,BuildRef} never existed in this module. The linker silently ignored them, but they were dead, so they're gone rather than repointed.

Makefile.coord.mk
```diff
@@ -0,0 +1,193 @@
...
+.PHONY: test-e2e-coordinator
+test-e2e-coordinator: image-build-coordinator image-build-epp image-build-builder image-pull ## Run coordinator e2e tests against a new kind cluster
+	test/scripts/run_e2e_coordinator.sh
+
+.PHONY: build
+build: image-build-builder ## Build the coordinator binary
+	@printf "\033[33;1m==== Building coordinator ====\033[0m\n"
+	$(BUILDER_RUN) 'go build -ldflags "$(LDFLAGS)" -o bin/coordinator ./cmd/coordinator/...'
+
+COORDINATOR_CONFIG ?= configs/coordinator.yaml
```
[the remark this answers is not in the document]
roytman: Fixed in `0a0828d0`. `COORDINATOR_CONFIG` now defaults to `config/coordinator/coordinator.yaml`, matching the migrated location, so make run works out of the box.

test/coordinator/e2e/internal/e2eutil/util.go
```diff
@@ -0,0 +1,130 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package e2eutil holds helpers shared across coordinator e2e test suites.
+// The internal/ path prevents import from outside test/e2e/.
+package e2eutil
```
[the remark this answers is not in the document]
roytman: Fixed in `0a0828d0`. Comment now reads `test/coordinator/e2e/`, matching the actual internal/ boundary.

test/coordinator/real-vllm-e2e/README.md
```diff
@@ -0,0 +1,122 @@
...
+go test -tags=e2e -v -run TestE2E_Completions ./test/real-vllm-e2e/...
+
+# Only encode tests
+go test -tags=e2e -v -run TestE2E_Encode ./test/real-vllm-e2e/...
+
+# Only prefill tests
+go test -tags=e2e -v -run TestE2E_Prefill ./test/real-vllm-e2e/...
+
+# Single test
+go test -tags=e2e -v -run TestE2E_Encode_RenderThenEncode ./test/real-vllm-e2e/...
```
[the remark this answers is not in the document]
roytman: Resolved by removing the `real-vllm-e2e` suite from this PR

test/coordinator/real-vllm-e2e/README.md
```diff
@@ -0,0 +1,122 @@
...
+  `RENDER_E2E_URL` is wrong.
+- **`connection refused` on encode/prefill** — the vLLM+EPP endpoint is not
+  running, or `ENCODE_E2E_GATEWAY` is wrong.
+- **HTTP 404 from render** — the service is reachable but does not expose
+  `/render`-suffixed routes. Confirm `RENDER_E2E_URL` targets a vLLM instance
+  with the render extension loaded, not a plain vLLM server.
+- **HTTP 400 on image tests** — the model is not vision-capable. Set
+  `ENCODE_E2E_MODEL` / `RENDER_E2E_MODEL` to a vision model.
+- **No tests run with `go test ./test/real-vllm-e2e/...`** — you forgot
+  `-tags=e2e`. Without the tag the package contains no buildable Go source.
```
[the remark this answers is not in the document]
roytman: Fixed in `0a0828d0` as part of the repo-wide path correction in this file.

test/coordinator/real-vllm-e2e/README.md
```diff
@@ -0,0 +1,122 @@
...
+go test -tags=e2e -v -run TestE2E_Prefill ./test/real-vllm-e2e/...
+
+# Single test
+go test -tags=e2e -v -run TestE2E_Encode_RenderThenEncode ./test/real-vllm-e2e/...
+
+# Override endpoints
+RENDER_E2E_URL=http://vllm-render:8000 \
+ENCODE_E2E_GATEWAY=http://vllm-epp:8080 \
+ENCODE_E2E_MODEL=Qwen/Qwen3-VL-2B-Instruct \
+  go test -tags=e2e -v ./test/real-vllm-e2e/...
```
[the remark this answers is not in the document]
roytman: Resolved by removing the `real-vllm-e2e` suite from this PR

Dockerfile.coordinator
```diff
@@ -0,0 +1,47 @@
...
+
+# Runtime stage
+FROM ${BASE_IMAGE}
+
+WORKDIR /
+COPY --from=builder /workspace/bin/coordinator /coordinator
+COPY --from=builder /workspace/configs/ /configs/
```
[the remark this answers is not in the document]
roytman: fixed

pkg/coordinator/pipeline/context.go
```diff
@@ -0,0 +1,106 @@
...
+var hopByHopHeaders = map[string]bool{
+	"connection":          true,
+	"keep-alive":          true,
+	"proxy-authenticate":  true,
+	"proxy-authorization": true,
+	"te":                  true,
+	"trailers":            true,
+	"transfer-encoding":   true,
+	"upgrade":             true,
+}
```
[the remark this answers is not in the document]
roytman: Fixed in 9fd15d22. Corrected to `trailer` (singular), matching the standard hop-by-hop header per RFC errata 4522 and Go's own net/http/httputil list. The plural was letting a client-supplied Trailer header through.

test/coordinator/scripts/run_e2e_coordinator.sh
```diff
@@ -0,0 +1,21 @@
...
+    fi
+    exit 130
+}
+
+trap cleanup INT TERM
+
+echo "Running coordinator end-to-end tests"
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+NAMESPACE="${NAMESPACE:-default}" go test -v -timeout 120m ${DIR}/../e2e/coordinator/ -ginkgo.v -ginkgo.fail-fast
```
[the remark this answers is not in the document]
roytman: This script is invoked only via make (Makefile.coord.mk:135), which runs from the module root, so `go test ${DIR}/../e2e/coordinator/` resolves correctly. It also follows the existing convention in this repo's other e2e runners (`test-e2e-router.sh`, `test-e2e-gaie.sh` use the same "${DIR}/../e2e/..." form). Leaving it consistent with those; happy to switch to a repo-relative path if you'd prefer that style repo-wide.

We will have an additional e2e test integration process. In the meantime, the Coordinator e2e tests are not part of the standard CI process.

pkg/coordinator/connectors/kv/sglang.go
```diff
@@ -0,0 +1,111 @@
...
+func parseSGLangBootstrapPort(raw string) (port int, rejected bool) {
+	if raw == "" {
+		return defaultSGLangBootstrapPort, false
+	}
+	p, err := strconv.Atoi(raw)
+	if err != nil {
+		return defaultSGLangBootstrapPort, true
+	}
+	return p, false
+}
```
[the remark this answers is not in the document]
roytman: Fixed in c044ce3a. parseSGLangBootstrapPort now rejects values <1 or >65535 and falls back to the default (logged as rejected), same as the non-numeric path. 
Added test cases for zero, negative, 65536, and 65535.

Makefile.coord.mk
```diff
@@ -0,0 +1,193 @@
...
+go-mod-check: image-build-builder
+	@echo "Checking go.mod/go.sum are clean..."
+	$(BUILDER_RUN) 'go mod tidy'
+	@git diff --exit-code go.mod go.sum || \
+	( echo "ERROR: go.mod/go.sum are not tidy. Run 'go mod tidy' and commit."; exit 1 )
+
+.PHONY: tidy
+tidy: ## Tidy go modules
+	go mod tidy
+
```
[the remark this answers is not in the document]
roytman: Fixed in c044ce3a. tidy now runs go mod tidy inside the builder via $(BUILDER_RUN) with image-build-builder as a prerequisite, matching go-mod-check, build, test-unit, etc.

README.coord.md
```diff
@@ -0,0 +1,246 @@
...
+5. **decode** - Forwards the final request to decode worker, streams response back
+
+## Quick Start
+
+```bash
+# Build
+make build
+
+# Run with default config
+make run
```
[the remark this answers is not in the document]
roytman: Fixed in c044ce3a. 
The Quick Start now uses `make -f Makefile.coord.mk build|run|test`, with a note that the root Makefile does not include the coordinator targets.

pkg/coordinator/common/httplog/redact.go
```diff
@@ -0,0 +1,81 @@
...
+
+// RedactedHeaders returns a flattened copy of h with values of sensitive
+// headers replaced by a redaction sentinel. It accepts either http.Header
+// (map[string][]string) or map[string]string and always returns the flat
+// form. Keys are normalized to lowercase so log output is consistent
+// regardless of how the input was canonicalized. A multi-valued header keeps
+// only its first value. A header with no value (an empty slice or an empty
+// string) retains its key with an empty-string value, so both input forms
+// behave the same. Values longer than maxValueLen are truncated.
+func RedactedHeaders[V string | []string](h map[string]V) map[string]string {
```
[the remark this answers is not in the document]
roytman: This one doesn't hold, the code compiles and go vet passes on all call sites. 
`RedactedHeaders[V string | []string](h map[string]V)` accepts `RedactedHeaders(req.Header)` because Go infers `V = []string` and `http.Header` (underlying type `map[string][]string`) is assignable to the parameter type. A ~map constraint isn't needed. Added `TestRedactedHeaders_AcceptsRequestHeaderField `in 6fe469ec, which passes `req.Header` directly, to lock this in.

Makefile.coord.mk
```diff
@@ -0,0 +1,193 @@
...
+test: test-unit ## Run all tests
+
+.PHONY: test-unit
+test-unit: image-build-builder
+	@printf "\033[33;1m==== Running Unit Tests ====\033[0m\n"
+	$(BUILDER_RUN) "go test -v -race $(TEST_PACKAGES)"
+
+.PHONY: test-e2e-coordinator
+test-e2e-coordinator: image-build-coordinator image-build-epp image-build-builder image-pull ## Run coordinator e2e tests against a new kind cluster
+	test/scripts/run_e2e_coordinator.sh
```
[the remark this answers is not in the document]
roytman: Fixed in d7465d90. `test-e2e-coordinator` now invokes `test/coordinator/scripts/run_e2e_coordinator.sh`, matching where the migration placed the runner.

test/coordinator/e2e/coordinator/e2e_suite_test.go
```diff
@@ -0,0 +1,249 @@
...
+	epdPoolsKustomizeDir    = "../../../deploy/environments/dev/epd-pools"
+	coordinatorComponentDir = "../../../deploy/components/coordinator"
+	rendererComponentDir    = "../../../deploy/components/vllm-render"
+
+	envoyManifest = "testdata/envoy.yaml"
+
+	crdGatewayAPIPath = "../../../deploy/components/crds-gateway-api"
+	crdGIEPath        = "../../../deploy/components/crds-gie"
+
+	baseRbacManifest = "../../../deploy/components/inference-gateway/base/rbac.yaml"
```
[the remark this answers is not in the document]
roytman: Fixed in d7465d90. All manifest and kustomize fixture paths now point at `deploy/coordinator/... `where the migration placed them. 
Verified all 12 resolve on disk.

cmd/coordinator/main.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+		step, err := pipeline.Build(stepCfg.Type, gwClient, params)
+		if err != nil {
+			return nil, err
+		}
+
+		steps = append(steps, step)
+	}
+	return steps, nil
+}
```
shmuelk, line 180: Perhaps move this function to the pipeline package and rename it Build?

cmd/coordinator/main.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	if _, ok := out[steps.ParamKVConnector]; !ok && kvConnector != "" {
+		out[steps.ParamKVConnector] = kvConnector
+	}
+	if _, ok := out[steps.ParamECConnector]; !ok && ecConnector != "" {
+		out[steps.ParamECConnector] = ecConnector
+	}
+	return out
+}
+
```
shmuelk, line 160: Perhaps move this function to the pipeline package?

cmd/coordinator/main.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil
+	}
+	for _, s := range p.Steps {
+		if s.Type == steps.RenderStepName {
+			return nil
+		}
+	}
+	return fmt.Errorf("pipeline.use_openai_format=false requires a %q step (the tokens-in format sends token IDs that render produces)", steps.RenderStepName)
+}
+
```
shmuelk, line 146: Perhaps move this function to the pipeline package?

cmd/coordinator/main.go
```diff
@@ -0,0 +1,176 @@
...
+	// (http://user:pass@host) and must not reach startup logs. NO_PROXY is a
+	// plain host list, so it is safe to log verbatim.
+	log.Info("proxy environment",
+		"http_proxy_set", os.Getenv("HTTP_PROXY") != "",
+		"https_proxy_set", os.Getenv("HTTPS_PROXY") != "",
+		"NO_PROXY", os.Getenv("NO_PROXY"))
+
+	gwClient := gateway.New(cfg.Gateway)
+
+	steps, err := buildPipeline(cfg, gwClient)
```
shmuelk, line 85:
```suggestion
	steps, err := pipeline.Build(cfg, gwClient)
```

cmd/coordinator/main_test.go
```diff
@@ -0,0 +1,60 @@
...
+package main
+
+import (
+	"testing"
+
+	"github.com/llm-d/llm-d-router/pkg/coordinator/config"
+	"github.com/llm-d/llm-d-router/pkg/coordinator/steps"
+)
+
+func TestValidatePipeline(t *testing.T) {
```
shmuelk, line 26: Perhaps move this to the pipeline package

config/coordinator/coordinator.yaml
```diff
@@ -0,0 +1,224 @@
...
+      #   ec_connector: ec-nixl     # overrides pipeline.ec_connector
+      #   kv_connector: kv-nixl     # overrides pipeline.kv_connector
+
+    # -------------------------------------------------------------------
+    # decode: streams the completion back to the client. Last step.
+    # -------------------------------------------------------------------
+    - type: decode
+      # params:
+      #   use_openai_format: true   # see notes above
+      #   kv_connector: kv-nixl     # overrides pipeline.kv_connector
```
shmuelk, line 224: I noticed that only render has all of the extended TCP connection parameters.

encode, decode, and prefill don't need them?

deploy/coordinator/components/coordinator/service-account.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: llm-d-coordinator
```
shmuelk, line 4: Why do you need a separate service account?

Do you need a service account altogether?

deploy/coordinator/components/crds-gateway-api/gateway-api.yaml
```diff
@@ -0,0 +1,17527 @@
+# Copyright The Kubernetes Authors.
```
shmuelk, line 1: Why is this needed? deploy/components/crds-gateway-api already exists

deploy/coordinator/components/crds-gateway-api/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+# ------------------------------------------------------------------------------
```
shmuelk, line 1: Why is this needed? deploy/components/crds-gateway-api already exists

deploy/coordinator/components/crds-gie/gie.yaml
```diff
@@ -0,0 +1,1131 @@
+apiVersion: apiextensions.k8s.io/v1
```
shmuelk, line 1: Why is this needed? deploy/components/crds-gie already exists

deploy/coordinator/components/crds-gie/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
+# ------------------------------------------------------------------------------
```
shmuelk, line 1: Why is this needed? deploy/components/crds-gie already exists

cmd/coordinator/main.go
roytman: Agreed this reads better in the pipeline package, but it can't move as-is: steps already imports pipeline (decode.go, encode.go, etc.), and these functions depend on steps constants (RenderStepName, ParamKVConnector, ParamECConnector), so relocating them into pipeline creates an import cycle, which is why they currently live in main (the only package that can depend on both). 
There's also a name clash: `pipeline.Build` already exists as the per-step registry builder. 
Doing this properly means either a new pipeline/builder subpackage or relocating those constants + renaming the existing Build. 
I'd like to do it as a focused follow-up PR rather than widen this migration PR, see https://github.com/llm-d/llm-d-router/issues/1879 
Does that work for you?

cmd/coordinator/main.go
roytman: see https://github.com/llm-d/llm-d-router/pull/1872#discussion_r3512329612

cmd/coordinator/main.go
roytman: https://github.com/llm-d/llm-d-router/pull/1872#discussion_r3512329612

config/coordinator/coordinator.yaml
roytman: encode, decode, prefill and conditional_decode have a gateway client with all its TCP parameters

deploy/coordinator/components/coordinator/service-account.yaml
roytman: it was from a separate deployment; I have to do an additional integration path after code merging see, https://github.com/llm-d/llm-d-router/issues/1877

deploy/coordinator/components/crds-gateway-api/gateway-api.yaml
roytman: as above

deploy/coordinator/components/crds-gateway-api/kustomization.yaml
roytman: as above

deploy/coordinator/components/crds-gie/gie.yaml
roytman: as above

deploy/coordinator/components/crds-gie/kustomization.yaml
roytman: as above

pkg/coordinator/pipeline/pipeline.go
```diff
@@ -0,0 +1,101 @@
...
+// sent to the client.
+type UpstreamError struct {
+	Step       string
+	StatusCode int
+	Body       string
+}
+
+func (e *UpstreamError) Error() string {
+	return fmt.Sprintf("%s: upstream returned HTTP %d: %s", e.Step, e.StatusCode, e.Body)
+}
```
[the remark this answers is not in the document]
roytman: Fixed in 153db842. UpstreamError.Error() no longer includes Body, so the upstream response (which can carry prompt/user data) is no longer logged via the error-level log in handleInference. The Body field stays on the struct for programmatic inspection, and it was already kept off the client response.

README.coord.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+## Development
+
+```bash
+make build    # Build binary to bin/coordinator
+make test     # Run all tests
+make lint     # Run golangci-lint
+make tidy     # Run go mod tidy
+make clean    # Remove build artifacts
```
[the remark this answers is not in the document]
roytman: Fixed in 29aab373. The Development section now uses make -f Makefile.coord.mk build|test|lint|tidy|clean, matching the Quick Start section, since the root Makefile doesn't include the coordinator targets.

cmd/coordinator/main.go
shmuelk: ok

cmd/coordinator/main_test.go
roytman: This test moves together with validatePipeline, so it's part of the same refactor. As noted on the main.go thread, moving these functions into pipeline currently creates an import cycle (steps already imports pipeline, and validatePipeline/buildPipeline depend on steps constants like RenderStepName), plus pipeline.Build already exists as the per-step registry builder. I'd like to do the move properly, either a pipeline/builder subpackage or relocating those constants and renaming the existing Build, as a focused follow-up PR rather than expanding this migration. Does that work?

docs/coordinator_architecture.md
```diff
@@ -0,0 +1,702 @@
...
+that aborts startup.
+
+### Steps over five points
+
+1. Pick a type name and declare it as a constant.
+2. Register the factory in `init()`.
+3. Implement the constructor (`StepFactory`): parse `params`, validate, build the step.
+4. Implement `Name()` and `Execute()`.
+5. If the step needs the gateway client or other injected dependencies, expose a setter
+   (see [Dependency injection](#dependency-injection)).
```
shmuelk: I thought this changed?

docs/coordinator_architecture.md
```diff
@@ -0,0 +1,702 @@
...
+  `FlushInterval: -1`, which forwards each write immediately (SSE chunks stream through; a
+  non-streaming JSON response passes through as one body). A terminal step like `decode`
+  returns `nil`, since it is the last step; a non-terminal step that has fully served the
+  response returns `ErrPipelineDone` to skip the remaining steps (the `conditional-decode`
+  cache hit). `decode` and `conditional-decode` are the reference implementations.
+
+A step is constructed by a `StepFactory`:
+
+```go
+type StepFactory func(params map[string]any) (Step, error)
```
shmuelk: I thought this changed?

docs/coordinator_architecture.md
```diff
@@ -0,0 +1,702 @@
...
+func init() {
+    pipeline.Register(ExampleStepName, NewExampleStep)
+}
+
+type ExampleStep struct {
+    threshold int
+    gwClient  *gateway.Client
+}
+
+func NewExampleStep(params map[string]any) (pipeline.Step, error) {
```
shmuelk: Is this still the API?

docs/coordinator_architecture.md
```diff
@@ -0,0 +1,702 @@
...
+
+func (s *ExampleStep) Execute(ctx context.Context, reqCtx *pipeline.RequestContext) error {
+    // read/mutate reqCtx; call s.gwClient as needed.
+    return nil
+}
+```
+
+Because the file lives in `pkg/steps`, its `init()` runs whenever the binary imports the
+package (the entrypoint already does, via `cmd/coordinator/main.go`). No central list to
+edit: registration is the only wiring step.
```
shmuelk, line 466: It was originally this way with plugins in the EPP. I was asked to change it to centralized registration, because it was believed to be easier to see the big picture.

docs/coordinator_architecture.md
```diff
@@ -0,0 +1,702 @@
...
+
+The gateway client is injected through the `gateway.ClientAware` interface
+([pkg/coordinator/gateway/client.go](../pkg/coordinator/gateway/client.go)):
+
+```go
+// in pkg/gateway
+type ClientAware interface {
+    SetGatewayClient(*Client)
+}
+```
```
shmuelk: Does this still exist?

docs/coordinator_architecture.md
```diff
@@ -0,0 +1,702 @@
...
+
+[cmd/coordinator/main.go](../cmd/coordinator/main.go) applies it after building each step:
+
+```go
+if ga, ok := step.(gateway.ClientAware); ok {
+    ga.SetGatewayClient(gwClient)
+}
+```
```
shmuelk: Ditto

docs/coordinator_architecture.md
```diff
@@ -0,0 +1,702 @@
...
+
+If a step needs the gateway client, implement `gateway.ClientAware` (a
+`SetGatewayClient(*gateway.Client)` method); the four Gateway-facing steps (`encode`,
+`prefill`, `decode`, `conditional-decode`) do. To inject a different dependency, define an
+analogous setter interface next to that dependency and add a type-assert block in
+`buildPipeline`. Keep injection optional: a step that does not implement the interface is
+left untouched.
```
shmuelk: Ditto

docs/coordinator_architecture.md
roytman: No, it doesn't, updated

docs/coordinator_architecture.md
roytman: Good catches, the doc had drifted from the code. 
Fixed in d1051bca: StepFactory now shows the gwClient parameter, the example step uses constructor injection, and the Dependency-injection section describes passing the client through the factory. 
Removed the `ClientAware`/`SetGatewayClient` description since that interface no longer exists in the code.

docs/coordinator_architecture.md
roytman: yes, as above

docs/coordinator_architecture.md
roytman: now it's 
```go
func NewExampleStep(gwClient *gateway.Client, params map[string]any) (pipeline.Step, error) {
```.
The document was updated.

File: docs/coordinator_architecture.md (hunk 32 above)
Comment (roytman) [reply docs/coordinator_architecture.md:466]: we can return to this later

File: docs/coordinator_architecture.md (hunk 34 above)
Comment (roytman) [reply docs/coordinator_architecture.md]: fixed.

File: deploy/coordinator/components/coordinator/deployment.yaml [hunk 36]
```diff
@@ -0,0 +1,64 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: llm-d-coordinator
+  labels:
+    app: llm-d-coordinator
+spec:
+  selector:
+    app: llm-d-coordinator
+  type: NodePort
```
capri-xiyue, line 58: why it should be a type `NodePort` here?

deploy/coordinator/components/crds-gateway-api/gateway-api.yaml
```diff
@@ -0,0 +1,17527 @@
...
+metadata:
+  annotations:
+    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/gateway-api/pull/4530
+    gateway.networking.k8s.io/bundle-version: v1.5.1
+    gateway.networking.k8s.io/channel: standard
+  labels:
+    gateway.networking.k8s.io/policy: Direct
+  name: backendtlspolicies.gateway.networking.k8s.io
+spec:
+  group: gateway.networking.k8s.io
```
capri-xiyue, line 33: Why do we need to provide the standard Gateway API CRDs here? I thought users were supposed to get them by following the standard setup guide.

pkg/coordinator/connectors/ec/ec.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/llm-d/llm-d-router/pkg/coordinator/pipeline"
+)
+
+// DefaultECConnectorName is the EC connector selected when an empty string is
+// passed to Build. Defaults to ec-shared-storage (no-op on the wire).
+const DefaultECConnectorName = SharedStorage
```
capri-xiyue, line 31: later should we change the default ec connector to nixl?

pkg/coordinator/connectors/kv/kv.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/llm-d/llm-d-router/pkg/coordinator/pipeline"
+)
+
+// DefaultKVConnectorName is the KV connector selected when an empty string is
+// passed to Build. Defaults to kv-shared-storage (no-op on the wire).
+const DefaultKVConnectorName = SharedStorage
```
capri-xiyue, line 33: same here

pkg/coordinator/connectors/kv/names.go
```diff
@@ -0,0 +1,23 @@
...
+limitations under the License.
+*/
+
+package kv
+
+const (
+	NIXL          = "kv-nixl"
+	SharedStorage = "kv-shared-storage"
+	SGLang        = "kv-sglang"
+)
```
shmuelk, line 23: nit: As this is part of the interface, I would merge this file with kv.go. This can be a followup PR.

pkg/coordinator/connectors/kv/sglang.go
```diff
@@ -0,0 +1,112 @@
...
+// default is returned.
+func parseSGLangBootstrapPort(raw string) (port int, rejected bool) {
+	if raw == "" {
+		return defaultSGLangBootstrapPort, false
+	}
+	p, err := strconv.Atoi(raw)
+	if err != nil || p < 1 || p > 65535 {
+		return defaultSGLangBootstrapPort, true
+	}
+	return p, false
```
shmuelk, line 64: nit: Tgis function is called from only one place, a small function. Why not merge them?

pkg/coordinator/connectors/kv/sglang.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// resolveSGLangBootstrapPort reads SGLANG_BOOTSTRAP_PORT once on first use,
+// where a configured context logger is available to report a rejected value.
+func resolveSGLangBootstrapPort(ctx context.Context) int {
+	sglangBootstrapPortOnce.Do(func() {
+		raw := os.Getenv(envSGLangBootstrapPort)
```
shmuelk, line 71: nit: I think it would be better not to use an environment variable here and instead use a field in the connector's parameters.

pkg/coordinator/gateway/client.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("sending request to gateway: %w", err)
+	}
+
+	// Buffer the body only to log it at TRACE; otherwise hand the caller the
+	// unread stream so large prefill/encode responses are not held in memory.
+	if v := logger.V(logutil.TRACE); v.Enabled() {
```
shmuelk, line 106: 👍

pkg/coordinator/gateway/client.go
```diff
@@ -0,0 +1,195 @@
...
+func (c *Client) BaseURL() string {
+	return c.baseURL
+}
+
+// Transport returns the underlying HTTP transport for reuse by reverse proxies.
+func (c *Client) Transport() http.RoundTripper {
+	return c.httpClient.Transport
+}
+
+// redactBody parses JSON and replaces string values longer than 50 chars with
```
shmuelk: The comment says 50 characters, the code says 200.

Please use a constant

deploy/coordinator/components/crds-gateway-api/gateway-api.yaml
vMaroon: Agreed, pls minimize deployment artifacts.

pkg/coordinator/gateway/client.go
```diff
@@ -0,0 +1,195 @@
...
+	return redactStringsDepth(v, 0)
+}
+
+func redactStringsDepth(v any, depth int) any {
+	if depth >= maxRedactDepth {
+		return "[truncated]"
+	}
+	switch val := v.(type) {
+	case string:
+		if len(val) > 50 {
```
shmuelk: Is this suppose to be the same length as above? Again use a constant.

pkg/coordinator/gateway/client.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+		return val
+	case map[string]any:
+		out := make(map[string]any, len(val))
+		for k, item := range val {
+			out[k] = redactStringsDepth(item, depth+1)
+		}
+		return out
+	case []any:
+		const maxElems = 10
```
shmuelk: Please move this constant outside the function. Keep all of these truncation length constants together

pkg/coordinator/gateway/client_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+// redactStrings is the log-redaction safety net: it keeps base64 blobs,
+// credential-bearing URLs, and oversized arrays out of trace logs. These cases
+// pin each branch.
+func TestRedactStrings(t *testing.T) {
+	const shortStr = "keep me"
+	longBase64 := "data:image/png;base64," + strings.Repeat("A", 60)
+	longURL := "https://example.com/" + strings.Repeat("a", 60)
+	longHTTP := "http://example.com/" + strings.Repeat("a", 60)
+	longPlain := strings.Repeat("x", 60)
```
shmuelk, line 65: Please double check these tests against the code that actually redacts text.

pkg/coordinator/gateway/paths.go
```diff
@@ -0,0 +1,80 @@
...
+// arbitrary path. There is no error return because an unrecognized path is not a
+// failure: it maps to the generate format by design.
+func DetectFormat(path string) RequestFormat {
+	if strings.Contains(path, PathChatCompletions) {
+		return FormatChatCompletions
+	}
+	if strings.Contains(path, PathCompletions) {
+		return FormatCompletions
+	}
+	return FormatGenerate
```
shmuelk, line 68: What about responses and messages?

roytman: they are in the plan

pkg/coordinator/connectors/ec/ec.go
roytman: Possible. Actually, I would like to add `None` connectors as well (see issue #1881 ). This will allow the coordinator to be used for deployments that lack certain disaggregations (such as E-PD, P-D, and even D, at least for testing purposes). The default value can then depend on the coordinator pipeline.

However, the issue is that the configured connectors depend on the specific deployment, which means disaggregation could fail **silently** in some cases."

deploy/coordinator/components/coordinator/deployment.yaml
roytman: Please do not review the coordinator deployment. It is done for e2e tests, and I'll return to it after the merge. See https://github.com/llm-d/llm-d-router/issues/1877

deploy/coordinator/components/crds-gateway-api/gateway-api.yaml
roytman: Please do not review the coordinator deployment. It is done for e2e tests, and I'll return to it after the merge. See https://github.com/llm-d/llm-d-router/issues/1877

pkg/coordinator/pipeline/context.go
```diff
@@ -0,0 +1,106 @@
...
+	// KVTransferParams carries the prefill pod's KV-cache transfer hints to the
+	// decode step. Populated by PrefillStep from the prefill response; consumed
+	// by the KV connector when building the decode request.
+	KVTransferParams map[string]any
+
+	// ResponseWriter is used by decode steps to stream the final response to the client.
+	ResponseWriter http.ResponseWriter
+}
+
+type MultimodalEntry struct {
```
shmuelk, line 99: Missing a godoc comment

pkg/coordinator/pipeline/context.go
```diff
@@ -0,0 +1,106 @@
...
+	Index       int
+	Hash        string
+	Base64Data  string
+	ContentType string
+	KwargsData  string
+	Placeholder PlaceholderRange
+}
+
+type PlaceholderRange struct {
```
shmuelk, line 111: Missing a godoc comment

pkg/coordinator/pipeline/pipeline.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	defer func() {
+		stats := make([]any, 0, len(timings)*2)
+		for _, t := range timings {
+			stats = append(stats, t.name, t.duration.String())
+		}
+		logger.V(logutil.DEFAULT).Info("pipeline step timings", stats...)
+	}()
+
+	for _, step := range p.steps {
```
shmuelk: Eliminate use of append:


```suggestion
	for idx, step := range p.steps {
```

pkg/coordinator/pipeline/pipeline.go
```diff
@@ -0,0 +1,101 @@
...
+	type stepTiming struct {
+		name     string
+		duration time.Duration
+	}
+	timings := make([]stepTiming, 0, len(p.steps)+1)
+	if reqCtx.ParseDuration > 0 {
+		timings = append(timings, stepTiming{name: "parse", duration: reqCtx.ParseDuration})
+	}
+	defer func() {
+		stats := make([]any, 0, len(timings)*2)
```
shmuelk: Eliminate use of append:


```suggestion
	timings := make([]stepTiming, len(p.steps))
	defer func() {
		stats := make([]any, 0, (len(timings)+1)*2)
		if reqCtx.ParseDuration > 0 {
		      stats = append(stats, "parse", reqCtx.ParseDuration.String())
	}
```

pkg/coordinator/pipeline/pipeline.go
```diff
@@ -0,0 +1,101 @@
...
+	}()
+
+	for _, step := range p.steps {
+		if err := ctx.Err(); err != nil {
+			return fmt.Errorf("pipeline cancelled: %w", err)
+		}
+		logger.V(logutil.TRACE).Info("step starting", "step", step.Name())
+		start := time.Now()
+		err := step.Execute(ctx, reqCtx)
+		timings = append(timings, stepTiming{name: step.Name(), duration: time.Since(start)})
```
shmuelk: Eliminate use of append:


```suggestion
		timings[idx] = stepTiming{name: step.Name(), duration: time.Since(start)}
```

pkg/coordinator/pipeline/pipeline_test.go
```diff
@@ -0,0 +1,130 @@
...
+	fn   func(ctx context.Context, rc *RequestContext) error
+}
+
+func (m *mockStep) Name() string { return m.name }
+func (m *mockStep) Execute(ctx context.Context, rc *RequestContext) error {
+	return m.fn(ctx, rc)
+}
+
+func TestPipeline_ExecutesStepsInOrder(t *testing.T) {
+	var order []string
```
shmuelk:
```suggestion
	order := []string{}
```

pkg/coordinator/server/handlers.go
```diff
@@ -0,0 +1,139 @@
...
+
+	"github.com/llm-d/llm-d-router/pkg/coordinator/config"
+	"github.com/llm-d/llm-d-router/pkg/coordinator/pipeline"
+)
+
+// validRequestID bounds a client-supplied x-request-id to alphanumerics and
+// dashes, at most 128 characters. handleInference replaces a header that fails
+// the match with a generated UUID before it reaches error responses; the
+// request-logging middleware redacts it. This keeps attacker-controlled content
+// out of both surfaces.
```
shmuelk, line 43: Doesn't redacting the request id make it harder for providers to debug issues?

pkg/coordinator/gateway/client.go
roytman: Fixed in `b74200fa`. 
The truncation limits are now named constants grouped together (`maxRedactStringLen`, `maxRedactRawBodyLen`, `maxRedactElems`, `maxRedactDepth`), `maxElems` was moved out of the function, and the redactBody comment was corrected. 
Note the per-field JSON limit (50) and the non-JSON whole-body window (200) are intentionally different: 200 leaves enough of a malformed body to diagnose why it didn't parse.

pkg/coordinator/gateway/client_test.go
roytman: Double-checked, the expectations matched the code (50-char boundary kept, >50 redacted, base64/url markers, non-JSON >200 truncated). 
Updated the tests in `b74200fa` to reference the new constants instead of literals, so they track the code if a limit changes.
`maxRedactRawBodyLen` (200 characters) is tested in `TestRedactBody`

pkg/coordinator/connectors/kv/sglang.go
roytman: `parseSGLangBootstrapPort` is also called directly by the unit test (`TestParseSGLangBootstrapPort`), which is why it's split out: it's a pure (raw) -> (port, rejected) function that's easy to table-test, while resolveSGLangBootstrapPort wraps it with `os.Getenv`, `sync.Once`, and logging that would be awkward to exercise. 
I'd keep them separate for that testability, but will merge if you feel strongly.

pkg/coordinator/connectors/kv/sglang.go
roytman: Agreed a connector param would be more consistent than an env var. It's a bit more than a local change though: `kv.Build` currently takes only the connector name, and connectors carry no config, so making this a param means threading it from the step params: through kv.Build into a stateful connector, touching the connector abstraction for all KV/EC types. 

I'd prefer to do that as a focused follow-up (env var → connector params) rather than expand this migration PR. Does that work?

https://github.com/llm-d/llm-d-router/issues/1893

pkg/coordinator/server/server.go
```diff
@@ -0,0 +1,137 @@
...
+func New(cfg config.ServerConfig, p *pipeline.Pipeline) *Server {
+	maxBodySize := cfg.MaxRequestBodySize
+	if maxBodySize == 0 {
+		// Zero means unset; Viper fills this from the config default in
+		// production. Direct callers (e.g. tests) that leave it unset get
+		// the same default.
+		maxBodySize = config.DefaultMaxRequestBodySize
+	}
+	if maxBodySize < 0 {
+		panic(fmt.Sprintf("server: MaxRequestBodySize must be positive, got %d", maxBodySize))
```
shmuelk: It might be better to have the function return an error rather than calling panic

pkg/coordinator/server/server.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	if maxBodySize > (math.MaxInt64-1)/config.BytesPerMB {
+		// maxRequestBodySize*1024*1024+1 is used as the io.LimitReader sentinel;
+		// an MB value that overflows int64 when converted to bytes would cause
+		// LimitReader to receive a negative limit and return immediate EOF.
+		panic(fmt.Sprintf("server: MaxRequestBodySize must be at most %d MB, got %d", int64((math.MaxInt64-1)/config.BytesPerMB), maxBodySize))
```
shmuelk: ditto

pkg/coordinator/server/server.go
```diff
@@ -0,0 +1,137 @@
...
+	s := &Server{pipeline: p, maxRequestBodySize: maxBodySize}
+
+	r := chi.NewRouter()
+	r.Use(middleware.RequestID)
+	r.Use(middleware.RealIP) //nolint:staticcheck // coordinator runs behind a trusted gateway that sets the forwarded-IP headers
+	r.Use(middleware.Recoverer)
+	r.Use(logRequestResponse)
+
+	r.Post(gateway.PathChatCompletions, s.handleInference)
+	r.Post(gateway.PathCompletions, s.handleInference)
```
shmuelk, line 113: What about generate?

pkg/coordinator/server/handlers.go
roytman: Only invalid request IDs are replaced, not all of them. 
A client `x-request-id` matching `^[a-zA-Z0-9\-]{1,128}$` (UUIDs, ULIDs, alphanumeric trace IDs, i.e. essentially all real ones) passes through verbatim and is used in both the log context and error responses, so normal provider debugging is unaffected.

The replacement exists as a security fix: it's finding `S11 (Medium)` in the coordinator's `TECH_DEBT_AUDIT.md`, `x-request-id` was previously accepted verbatim and concatenated into error-response bodies and logs, allowing `newline/CRLF log-injection` and `response injection`. Validating and replacing non-conforming IDs closes that.

Debuggability is still preserved on the replacement path: we log that a replacement happened plus the rejected length ("replaced invalid client request ID", "rejectedLength", N), and the fresh UUID is used consistently across logs and the error response, so the request stays traceable; we just don't reflect the attacker-controlled raw value.

pkg/coordinator/server/server.go
roytman: It is for the next steps

BTW, we have to check how `generate` works in disaggregated scenarios

pkg/coordinator/server/server.go
roytman: Done in `d58857d4`. `server.New` now returns `(*Server, error)` for both the negative and overflow `MaxRequestBodySize` guards instead of panicking, and `main` handles it with a logged error + exit, matching how buildPipeline failures are already handled right above the call. 
So a bad `max_request_body_size` now gives a clean startup error instead of a stack trace. Tests updated to assert the error.

pkg/coordinator/server/server.go
roytman: https://github.com/llm-d/llm-d-router/issues/1894

pkg/coordinator/steps/encode.go
```diff
@@ -0,0 +1,259 @@
...
+	maxParallel := 8
+	if v, ok, err := paramInt(params, "max_parallel"); err != nil {
+		return nil, err
+	} else if ok {
+		if v <= 0 {
+			return nil, fmt.Errorf("max_parallel must be positive, got %d", v)
+		}
+		maxParallel = v
+	}
+	ecName, _ := params[ParamECConnector].(string)
```
shmuelk: Is this gauranteed not to panic, if some placed for example an integer value in the ECConnector field in the config?

pkg/coordinator/steps/encode.go
```diff
@@ -0,0 +1,259 @@
...
+
+			resp, err := s.gwClient.Post(gCtx, path, bodyBytes, headers)
+			if err != nil {
+				err = fmt.Errorf("encode[%d]: request: %w", i, err)
+				logger.Error(err, "encode fanout request", "index", i, "path", path)
+				return err
+			}
+			defer resp.Body.Close()
+
+			if resp.StatusCode/100 != 2 {
```
shmuelk: Please add a comment that you are checking for 2XX status codes

pkg/coordinator/steps/prefill.go
```diff
@@ -0,0 +1,194 @@
...
+	kv              kv.Connector
+	ec              ec.Connector
+}
+
+func NewPrefillStep(gwClient *gateway.Client, params map[string]any) (pipeline.Step, error) {
+	if gwClient == nil {
+		return nil, errors.New("prefill: gateway client is required")
+	}
+	useOpenAI := parseUseOpenAIFormat(params)
+	kvName, _ := params[ParamKVConnector].(string)
```
shmuelk: Is this guaranteed not to panic if an integer value was entered as the connector name?

pkg/coordinator/steps/prefill.go
```diff
@@ -0,0 +1,194 @@
...
+		v.Info("request body", "method", "POST", "path", path, "bodyLen", len(bodyBytes), "headers", httplog.RedactedHeaders(headers))
+	}
+
+	resp, err := s.gwClient.Post(ctx, path, bodyBytes, headers)
+	if err != nil {
+		return fmt.Errorf("prefill: request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode/100 != 2 {
```
shmuelk: Add a comment that you are checking for 2XX status codes

pkg/coordinator/steps/render.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	idleConnTimeout := 90 * time.Second
+	if v, ok, err := paramDuration(params, "idle_conn_timeout"); err != nil {
+		return nil, err
+	} else if ok {
+		idleConnTimeout = v
+	}
+
+	address, _ := params["address"].(string)
```
shmuelk: Are you guaranteed that this won't panic if the address field contains an integer?

pkg/coordinator/steps/prefill.go
roytman: yes, but see https://github.com/llm-d/llm-d-router/pull/1872#discussion_r3528805629

pkg/coordinator/steps/render.go
roytman: Yes, but see https://github.com/llm-d/llm-d-router/pull/1872#discussion_r3528805629

pkg/coordinator/steps/encode.go
roytman: Rather than just annotate the `/100 != 2` check, I made encode and prefill require exactly `http.StatusOK`, matching render and replace-media-urls (1908ba5f). 
So the status handling is now consistent and self-documenting across all steps, no special 2xx comment needed.

pkg/coordinator/steps/prefill.go
roytman: see https://github.com/llm-d/llm-d-router/pull/1872#discussion_r3525040074

pkg/coordinator/steps/encode.go
roytman: Yes, it used the `comma-ok` form so it never panicked, but a non-string value was silently discarded (defaulting the connector or emptying the address below). 
Fixed in af527d57: added a `paramString` helper that returns a config error when the key is present but not a string, and used it for the connector names (encode/prefill/decode) and render's address. So a mistyped value like ec_connector: 5 now fails fast with ec_connector: expected string, got int. Added `TestParamString` for the missing/string/wrong-type cases.

shmuelk (approved):

Gregory-Pereira (approved):
/lgtm

hexfusion:
very exciting LGTM :tada:

elevran (approved):

PR #1873, opened 2026-07-02
Add accelerator capability aware scheduler plugin

### LINKED ISSUE #1868: feat: accelerator capability-aware routing for heterogeneous GPU pools
## Problem

Some deployments serve multimodal/VLM traffic across a heterogeneous accelerator pool, for example NVIDIA MIG slices such as `1g.10gb`, `2g.20gb`, and full-GPU pods. The request cost can vary significantly with input image size or estimated vision tokens, but the scheduler currently has no narrow, label-based way to match larger requests to larger static accelerator profiles.

This is distinct from runtime GPU load awareness in #1706. That issue focuses on dynamic DCGM utilization. This issue is intentionally scoped to static capability known from endpoint labels.

## Proposed narrow scope

Add a scheduler plugin that reuses the `context-length-aware` pattern from #550:

- read request size from `TokenizedPrompt.TokenCount()`, produced by `token-producer`
- read static accelerator capability ranges from endpoint labels
- support scorer mode by default
- optionally support filtering when `enableFiltering` is true
- pass through unlabeled endpoints and assign them a neutral score for incremental rollout

Suggested plugin type:

```yaml
type: accelerator-capability-aware
```

Suggested default label:

```yaml
llm-d.ai/accelerator-capability-range: "0-2048"
```

The label values are intentionally generic estimated-token ranges rather than vendor-specific MIG names. Operators can map their local profiles to ranges, for example small MIG slices for smaller multimodal requests and full GPUs for larger image-heavy requests.

## Out of scope

- DCGM scraping or runtime GPU utilization
- pod-to-device mapping
- DaemonSet or sidecar DCGM exporter topology
- vendor-specific accelerator discovery
- changing model server metrics

Those remain better covered by #1706 or future runtime-metrics work.

## Initial PR shape

I plan to open a small draft PR with:

- `accelerator-capability-aware` scorer/filter plugin
- default label and parser for `min-max` ranges
- token-producer dependency through `TokenizedPrompt`
- unit tests for factory, scoring, filtering, invalid ranges, and unlabeled endpoints
- README documentation

The draft is meant to validate the narrow API and naming before expanding behavior.

anencore94:
## What

Adds a draft `accelerator-capability-aware` scheduling plugin for heterogeneous accelerator pools.

- Matches estimated request size from `TokenizedPrompt.TokenCount()` against endpoint label ranges.
- Uses `llm-d.ai/accelerator-capability-range` as the default static capability label.
- Supports scorer mode by default and optional filtering via `enableFiltering`.
- Treats unlabeled endpoints neutrally so the label can be rolled out incrementally.
- Registers the plugin in the EPP runner and documents configuration.

## Why

This keeps static MIG / accelerator-profile routing separate from the runtime DCGM utilization work in #1706. It follows the narrower design proposed in #1868 and reuses the same request-size dependency pattern as `context-length-aware`.

Closes #1868.

## Validation

- `go test ./pkg/epp/framework/plugins/scheduling/scorer/acceleratorcapabilityaware`
- `go test ./cmd/epp/runner`

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -108,6 +108,7 @@ import (
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/profilehandler/dataparallel"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/profilehandler/disagg"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/profilehandler/single"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/scorer/acceleratorcapabilityaware"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/scorer/activerequest"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/scorer/contextlengthaware"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/scorer/kvcacheutilization"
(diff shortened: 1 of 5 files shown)


ahg-g:
sounds like a reasonable direction, not sure if there is a way to generalize this more, but no objectives to merging it as is. @elevran any thoughts on this?

nicole-lihui:
using the request token count to choose between accelerator tiers is useful.

I may be missing a distinction, but the range semantics and scorer/filter behavior seem to overlap with `context-length-aware`. If both plugins are enabled, would they filter or score the same request-size signal twice? If the ranges are not semantically different, could we extend or document `context-length-aware` instead?

PR #1876, opened 2026-07-02
Ensure that plugins that depend on other plugins are instantiated after their dependencies

### LINKED ISSUE #1421: Order plugin instantiation via a dependency graph
Plugin instantiation based on a dependency graph. All plugins with no dependencies are instantiated up front. Those with dependencies on other plugins will be instantiated after any plugins they depend on. It should be noted that dependencies are configuration dependent, that is they are not static compile time dependencies, but rather run-time dependencies.

Plugins will declare their own plugin dependencies by:

  1. Having a public function that parses their configuration parameters, including the setting of default values.
  2. Be registered via the plugin.RegisterWithPluginDependencies function.
  3. In the plugin's configuration struct fields must be marked as references to plugin names by adding a Go struct field tag of pluginRef. Please note that the value of the struct tag can be ignored, but seems to be needed to pass lint.

The configuration parser function must be of type:

```go
// ConfigParserFunc is the definition of the factory functions that are used to parse the
// parameters of plugins that have been registered as having dependencies on other plugins
type ConfigParserFunc func(parameters *json.Decoder, handle Handle) (any, error)
```

The plugin.RegisterWithPluginDependencies function is as follows:

```go
// RegisterWithPluginDependencies registers a factory for the given plugin type and records it as dependent on
// other plugins referenced in the configuration struct returned by the plugin's configuration parser function.
func RegisterWithPluginDependencies(pluginType string, factory FactoryFunc, parser ConfigParserFunc) {
	Register(pluginType, factory)
	PluginsWithPluginDependencies[pluginType] = parser
}
```

The configuration loader's plugin instantiation code will use the registered plugin configuration parsers
to get instances of the registered plugins configuration structs. Go reflection interfaces will then be used
to determine what plugins are referenced by name. Although this adds an extra parse of the configuration
of these plugins and the Go reflection analysis of the structs involved, potentially recursing through
the structs, it should be noted that this will happen once per such plugin at EPP startup time.

This will feed code that will build the dependency graph and order the plugin instantiations as appropriate.

Why is this needed:
Today plugins are instantiated in the order they appear in the textual configuration's plugins section.

In most cases this is perfectly fine, however there are plugins that reference other plugins. The noted example is the DisaggProfileHandler. The deprecated PdProfileHandler is similar in nature. For these plugins to work correctly they must be placed in the configuration after any plugins they reference.

This behavior is error prone.

shmuelk:
**What type of PR is this?**
/kind feature

Today plugins are instantiated in the order they appear in the textual configuration's plugins section.

In most cases this is perfectly fine, however there are plugins that reference other plugins. The noted example is the DisaggProfileHandler. The deprecated PdProfileHandler is similar in nature. For these plugins to work correctly they must be placed in the configuration after any plugins they reference.

This behavior is error prone.

This PR enables plugins to declare the plugins they are dependent on. These dependencies are by name in the dependent plugin's configuration.

See issue #1421 for more details.

Fixes #1421 

```release-note
The need to place the Disaggregation profile handler after its deciders in the plugins list of the configuration profile has been removed.
```

### DIFF
--- a/pkg/epp/config/loader/configloader.go
+++ b/pkg/epp/config/loader/configloader.go
@@ -229,7 +229,7 @@ func decodeRawConfig(configBytes []byte) (*configapi.EndpointPickerConfig, error
 func instantiatePlugins(configuredPlugins []configapi.PluginSpec, handle fwkplugin.Handle) error {
 	orderedPlugins, err := buildPluginDAG(configuredPlugins, handle)
 	if err != nil {
-		return fmt.Errorf("cycle in dependencies between plugins: %w", err)
+		return fmt.Errorf("failed to build plugin dependency graph: %w", err)
 	}
 
 	for _, spec := range orderedPlugins {
@@ -252,8 +252,8 @@ func buildPluginDAG(configuredPlugins []configapi.PluginSpec, handle fwkplugin.H
 
 	for _, spec := range configuredPlugins {
 		if parserFunc, ok := fwkplugin.PluginsWithPluginDependencies[spec.Type]; !ok {
-			// All plugins that don't have dependencies should be instantiated first
-			orderedPlugins = append(orderedPlugins, spec)
+			// Add plugins that don't have dependencies to the graph
+			graph[spec.Name] = []string{}
 		} else {
 			// Ignore extra fields.
 			configStruct, err := parserFunc(fwkplugin.StrictDecoder(spec.Parameters), handle)
@@ -271,7 +271,11 @@ func buildPluginDAG(configuredPlugins []configapi.PluginSpec, handle fwkplugin.H
 	}
 
 	for _, pluginName := range sortedGraph {
-		orderedPlugins = append(orderedPlugins, pluginMap[pluginName])
+		if spec, ok := pluginMap[pluginName]; ok {
+			orderedPlugins = append(orderedPlugins, spec)
+		} else {
+			return nil, fmt.Errorf("a plugin has a dependency on the unregistered plugin %s", pluginName)
+		}
 	}
 
 	return orderedPlugins, nil
--- a/pkg/epp/config/loader/configloader.go
+++ b/pkg/epp/config/loader/configloader.go
@@ -19,6 +19,7 @@ package loader
 import (
 	"errors"
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
@@ -29,7 +30,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/apimachinery/pkg/util/sets"
 
 	configapi "github.com/llm-d/llm-d-router/apix/config/v1alpha1"
 	"github.com/llm-d/llm-d-router/pkg/epp/config"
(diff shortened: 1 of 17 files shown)


pkg/epp/config/loader/configloader.go
```diff
@@ -248,6 +245,77 @@ func instantiatePlugins(configuredPlugins []configapi.PluginSpec, handle fwkplug
...
+	}
+
+	sortedGraph, err := util.TopologicalSort(graph)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, pluginName := range sortedGraph {
+		orderedPlugins = append(orderedPlugins, pluginMap[pluginName])
+	}
```
elevran, line 279: `buildPluginDAG` appends non-dependent plugins to `orderedPlugins` first, then passes only the dependent plugins to `TopologicalSort`. However `TopologicalSort` tracks in-degrees for all nodes mentioned as neighbors, so non-dependent plugins that are referenced as dependencies also appear in its output and get appended a second time, causing double instantiation.

Concretely: with `disagg-profile-handler` depending on `prefix-based-pd-decider` (not in `PluginsWithPluginDependencies`), `TopologicalSort` returns `[prefix-based-pd-decider, disagg-profile-handler]` and `prefix-based-pd-decider` lands in the ordered list twice.

This also risks a nil-factory panic when a referenced name is absent from the config (zero-value `PluginSpec` has an empty `Type`, so `Registry[""]` is nil).

Fix: only append plugins that are keys in `graph` (i.e., registered as having dependencies):

```suggestion
	for _, pluginName := range sortedGraph {
		if _, ok := graph[pluginName]; ok {
			orderedPlugins = append(orderedPlugins, pluginMap[pluginName])
		}
	}
```

pkg/epp/config/loader/configloader.go
```diff
@@ -223,20 +227,13 @@ func decodeRawConfig(configBytes []byte) (*configapi.EndpointPickerConfig, error
...
-		if spec.Type == "" {
-			return fmt.Errorf("plugin '%s' is missing a type", spec.Name)
-		}
-		if pluginNames.Has(spec.Name) {
-			return fmt.Errorf("duplicate plugin name '%s'", spec.Name)
-		}
-		pluginNames.Insert(spec.Name)
+	orderedPlugins, err := buildPluginDAG(configuredPlugins, handle)
+	if err != nil {
+		return fmt.Errorf("cycle in dependencies between plugins: %w", err)
```
elevran: The prefix "cycle in dependencies between plugins" wraps all errors from `buildPluginDAG`, including parse failures unrelated to cycles. A neutral message (e.g., `failed to build plugin dependency graph: %w`) is more accurate.

pkg/epp/config/loader/testdata_test.go
```diff
@@ -533,74 +671,23 @@ schedulingProfiles:
...
-dataLayer:
-  injectDefaults: false
-`
-
-// successDataLayerExplicitConfigText has the datalayer enabled with an explicit non-metrics source.
-// The loader should inject the default metrics source in addition to the user's source (additive).
-const successDataLayerExplicitConfigText = `
+// pluginsRefedByPointerText represents a valid config with a plugin that is dependent
+// on another plugin. In this case the dependent plugin is before the dependency and is
+// referenced via a pointer
```
elevran: nit: comment is copied from `pluginsRefedByPointerText` and still describes the pointer case. This constant is the cycle-detection error case.

pkg/epp/config/loader/configloader.go
shmuelk: Fixed

pkg/epp/config/loader/configloader.go
shmuelk: Fixed

pkg/epp/config/loader/testdata_test.go
shmuelk: fixed

elevran (approved):

PR #1884, opened 2026-07-03
epp: add --ha-populate-non-leader-datastore to serve from non-leaders

mathias-kende:
**What type of PR is this?**

/kind feature

When leader election is enabled, the InferencePool/Pod/CRD reconcilers default to running only on the leader (controller-runtime marks controllers as needing leader election), so non-leader replicas keep an empty datastore and answer every ext_proc request with a 503 "failed to find endpoint candidates". If Envoy ever routes to a standby (e.g. a persistent H2 connection that has not been re-resolved yet during a leadership handoff) the request fails hard.

Add an opt-in flag that registers those reconcilers with NeedLeaderElection=false so they run on every replica and keep the datastore populated. Readiness and the ext_proc health check stay leader-gated, so non-leaders remain NotServing and are not advertised; this only lets a request that does reach a standby be routed (with suboptimal scoring) instead of returning 503.

Default is off (unchanged upstream behavior).

none

```release-note
When leader election is enabled (--ha-enable-leader-election), the InferencePool, Pod, and CRD reconcilers now also run on non-leader EPP replicas so their datastore stays populated. Non-leaders remain NotServing but this lets a request that reaches a standby EPP be routed instead of returning a 503. This is controlled by the haPopulateNonLeaderDatastore feature gate, which is enabled by default. To restore the previous leader-only behavior, disable it in the featureGates section of the EPP configuration file: `haPopulateNonLeaderDatastore=false`.
```

### DIFF
--- a/cmd/epp/runner/feature_gate_test.go
+++ b/cmd/epp/runner/feature_gate_test.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package runner
+
+import (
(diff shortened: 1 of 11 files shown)


ahg-g:
I suggest to make this the default behavior, I don't see a reason why would one want to disable it. 

We can do this by implementing the flag as a feature gate that is enabled by default to give users the ability to disable it if it causes unforeseen issues, and then remove the feature gate after one release if no issues were reported.

mathias-kende:
Thanks for the review. I converted this to a feature gate.

ahg-g (approved):
Thanks!

PR #1885, opened 2026-07-03
epp: add --drain-timeout for graceful shutdown

mathias-kende:
**What type of PR is this?**

/kind feature

By default the ext_proc and health gRPC servers are manager runnables, so on SIGTERM they are stopped at the same moment the manager releases the leader lease. Any request Envoy still sends during that window (in-flight streams, or a persistent connection not yet re-resolved by DNS) is rejected.

With --drain-timeout > 0 the ext_proc and health servers are instead run on a context that outlives the manager. On SIGTERM the EPP:
  - flips a `draining` flag so readiness / ext_proc health report NOT_SERVING (Kubernetes drains the pod from the Service endpoints), while liveness stays SERVING to avoid a restart mid-drain;
  - lets the manager stop, releasing the leader lease immediately so a standby is promoted;
  - keeps the ext_proc server accepting requests for the drain window, then GracefulStops (draining in-flight streams).

The datastore's pod list survives manager shutdown, so endpoint resolution keeps working during the drain (with stale metrics).

Default (0) is unchanged upstream behavior.

none

```release-note
Add the --drain-timeout flag to allow graceful shutdown of the EPP. Which defaults to 30 seconds.
```


pkg/epp/server/options.go
```diff
@@ -153,6 +154,10 @@ func (opts *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.IntVar(&opts.GRPCPort, "grpc-port", opts.GRPCPort, "gRPC port used for communicating with Envoy proxy.")
 	fs.BoolVar(&opts.EnableLeaderElection, "ha-enable-leader-election", opts.EnableLeaderElection,
 		"Enables leader election for high availability. When enabled, readiness probes will only pass on the leader.")
+	fs.DurationVar(&opts.DrainTimeout, "drain-timeout", opts.DrainTimeout,
```
ahg-g, line 159: I think we better set a default and not disable it, what defaults did you test with? would 5min be a good one since in some cases requests can take a long time?

cmd/epp/runner/health.go
```diff
@@ -52,11 +53,16 @@ const (
 func (s *healthServer) Check(ctx context.Context, in *healthPb.HealthCheckRequest) (*healthPb.HealthCheckResponse, error) {
 	isLive := s.datastore.PoolHasSynced()
 	protocolMatches := s.checkProtocolSupport(isLive)
+	// While draining (graceful shutdown), every non-liveness check reports
+	// NOT_SERVING so Kubernetes removes the pod from the Service endpoints, while
```
ahg-g:
```suggestion
	// NOT_SERVING so Kubernetes removes the EPP pod from the Service endpoints, while
```

cmd/epp/runner/runner.go
```diff
@@ -417,7 +487,22 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
...
+		// flipped by Run() so readiness/ext_proc health go NOT_SERVING and the
+		// pod is drained from the Service endpoints.
+		r.draining = &atomic.Bool{}
+		r.serverRunner = serverRunner
+		r.healthGRPCPort = opts.GRPCHealthPort
+		r.healthGRPCServer = newHealthGRPCServer(ctrl.Log.WithName("health"), ds, isLeader, r.draining, opts.EnableLeaderElection, supporters)
+		return mgr, ds, nil
+	}
+
+	if err := registerHealthServer(mgr, ctrl.Log.WithName("health"), ds, opts.GRPCHealthPort, isLeader, nil, opts.EnableLeaderElection, supporters); err != nil {
```
ahg-g: If we are to keep the existing approach for setting up the servers as manager runnables, then I would wrap lines 505 to 512 in an else clause and remove the return statement at line 502 to clearly distinguish the different execution paths.

cmd/epp/runner/runner.go
```diff
@@ -417,7 +487,22 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
...
+		// cancelled after the drain window, so on SIGTERM the manager stops
+		// (releasing the leader lease) while these keep serving. `draining` is
+		// flipped by Run() so readiness/ext_proc health go NOT_SERVING and the
+		// pod is drained from the Service endpoints.
+		r.draining = &atomic.Bool{}
+		r.serverRunner = serverRunner
+		r.healthGRPCPort = opts.GRPCHealthPort
+		r.healthGRPCServer = newHealthGRPCServer(ctrl.Log.WithName("health"), ds, isLeader, r.draining, opts.EnableLeaderElection, supporters)
+		return mgr, ds, nil
+	}
```
ahg-g: Instead of maintaining two setups depending on whether or not draining is set, I suggest we drop setting up the servers as manager runnables and just adopt this approach for both cases, wdyt?

ahg-g (commented):
Thanks, really nice! a few suggestions to try and unify the logic and avoid creating separate approaches depending on whether draining is set or not.

cmd/epp/runner/runner.go
mathias-kende: Yes, this is a good simplification if we’re confident in this new setup.

Done.

cmd/epp/runner/health.go
mathias-kende: Done.

pkg/epp/server/options.go
mathias-kende: I have tested it with only 20 seconds. I think that it’s enough for all the propagation because the EPP is actively killed only by Kubernetes. So the long multi-minute timeout is set as the termination grace period of the Pod.

In that sense, maybe "drain" is not the right word and I should rename this whole thing to "lame-duck-mode" as the actual drain happens afterward.

cmd/epp/runner/runner.go
mathias-kende: Ack.

mathias-kende (commented):
Thanks, please take another look.

pkg/epp/server/options.go
mathias-kende: One caveat is that I had to revert the default drain duration in the e2e test and I did not look into fixing them for the new behavior (although they pass with the new code, with the 0s drain). Let me know if you think that they should be fixed.

cmd/epp/runner/runner.go
```diff
@@ -256,15 +260,66 @@ func (r *Runner) Run(ctx context.Context) error {
...
+	select {
+	case <-time.After(drainTimeout):
+		setupLog.Info("Drain window elapsed, stopping ext_proc server")
+	case err := <-serveErr:
+		// The servers exited on their own (e.g. listener error) during the drain.
 		return err
 	}
-	setupLog.Info("Controller manager terminated")
-	return nil
+	serveCancel()
```
ahg-g, line 322: we call this via defer above, is calling it here again intentional?

ahg-g (commented):
Thanks, looks great! One last comment, CI is failing as well.

cmd/epp/runner/runner.go
mathias-kende: Yes, I think we do need it (but I’m far from a go concurrency expert so I can be wrong), because serveErr will deadlock otherwise as serveCancel() would run only after we return.

Another option would be to remove the defer entirely and instead call serveCancel on line 319 too. It’s maybe slightly more error prone if someone forget to call it in a new path. Let me know if you’d prefer that.

mathias-kende (commented):
Thanks. The CI is passing now.

ahg-g (approved):

Issue #1737, opened 2026-06-24
Detect missed KV events via sequence gap and request replay


RishabhSaini:
The ZMQ subscriber extracts the sequence number from each message (`zmq_subscriber.go:134`) but never checks for gaps. Events missed during a network blip or ZMQ high-water-mark overflow are silently lost, leaving the index incomplete.

vLLM publishers assign monotonically increasing sequence numbers to each event batch. Detecting gaps is a single comparison per message: 
if `seqReceived > lastSeq + 1`, events were missed.

Once a gap is detected, the subscriber can request the missing events from vLLM's replay buffer via its ROUTER socket configurable per-pod at `replay_endpoint`). vLLM buffers the last 10,000 event batches and serves them on demand by starting sequence number.

This also covers EPP restart: the first message received will have `seq > 0`, triggering a full replay to rebuild the index.

Depends on: https://github.com/vllm-project/vllm/pull/45177
Related: https://github.com/llm-d/llm-d-kv-cache/issues/372

PR #1887, opened 2026-07-03
feat: replay buffer consumer for KV event index recovery

RishabhSaini:
## Summary

Add replay buffer consumer to rebuild the KV block index after EPP
restart or network blips. The subscriber connects to vLLM's ROUTER
socket via a DEALER and replays missed events.

- **Proactive replay on connect** — rebuilds index immediately, no
  wait for live traffic
- **Gap detection** — `seq > lastSeq + 1` triggers replay from
  `lastSeq + 1`
- **Dual watermarks** — `lastSeq` (index) vs `lastLiveSeq` (wire)
  so replay overlap cannot inflate dedup reference counts
- **Strict event gate** — after replay, live events with
  `seq <= lastSeq` are dropped before reaching AddTask
- **Sequence reset detection** — `seq < lastLiveSeq` clears index
  and dedup via pool queue (ordered with pending events)
- **Cooldown on failure only** — successful replay clears failure
  state; failed replay backs off 30s
- **DEALER over REQ** — REQ can only receive one reply per request

`EnsureSubscriber` gains `sourceEndpoint` and `replayEndpoint`
parameters. Replay endpoint computed from
`PodDiscoveryConfig.ReplaySocketPort` in the extractor.

Depends on: vllm-project/vllm#45177 (merged)

## Test plan

- [x] Proactive replay rebuilds serving endpoint identity on connect
- [x] Gap replay does not duplicate the triggering event
- [x] Events dropped during replay cooldown
- [x] Sequence reset clears and rebuilds pod index
- [x] Replayed live events do not trigger another replay

```release-note
Add replay buffer consumer for KV event index recovery after EPP
restart or network blips.
```

Relates to #1737

### DIFF
--- a/pkg/kvevents/zmq_subscriber.go
+++ b/pkg/kvevents/zmq_subscriber.go
@@ -27,8 +27,10 @@ import (
 )
 
 const (
-	// How long to wait before retrying to connect.
-	retryInterval = 5 * time.Second
+	retryInterval       = 5 * time.Second
+	replayTimeout       = 30 * time.Second
+	replaySocketTimeout = 10 * time.Second
+	replayCooldown      = 30 * time.Second
 )
 
 // zmqSubscriber connects to a ZMQ publisher and forwards messages to a pool.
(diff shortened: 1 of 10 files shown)


pkg/kvevents/zmq_subscriber.go
```diff
@@ -119,19 +151,29 @@ func (z *zmqSubscriber) runSubscriber(ctx context.Context) {
...
-			debugLogger.Error(nil, "Sequence frame too short", "got", len(seqBytes), "want", 8, "topic", topic, "endpoint", z.endpoint)
-			continue
+		// Gap detection: request replay for missed events.
+		if z.replayEndpoint != "" && z.canAttemptReplay() {
+			if z.hasLastSeq && seq > z.lastSeq+1 {
+				missed := seq - z.lastSeq - 1
+				logger.Info("Detected gap in event sequence, requesting replay",
+					"lastSeq", z.lastSeq, "currentSeq", seq, "missed", missed,
+					"endpoint", z.endpoint)
+				if !z.requestReplay(ctx, z.lastSeq+1) {
```
bongwoobak: Gap replay asks for `lastSeq+1`, and the triggering event is already buffered on the vLLM side (`_service_replay` sends everything with `seq >= start_seq`), so it comes back and AddTask below stores it a second time. `trackStore` counts both while the engine only ever emits one `BlockRemoved`, so `filterRemove` suppresses the eviction and the block stays in the index. `TestPool_DuplicateStoreSurvivesFirstRemove` is the same mechanism.

Could you skip already-covered seqs before AddTask? Ties into the cooldown and restart cases below.

pkg/kvevents/zmq_subscriber.go
```diff
@@ -143,5 +185,112 @@ func (z *zmqSubscriber) runSubscriber(ctx context.Context) {
 			Sequence: seq,
 			Payload:  payload,
 		})
+
+		// Freeze lastSeq during cooldown so the gap is preserved for
+		// retry — blocks past a gap are dropped (broken parent chain)
+		// and can only be recovered via replay.
+		if z.replayEndpoint == "" || z.canAttemptReplay() {
```
bongwoobak: Cooldown freezes `lastSeq` but the AddTask above still runs, so events land in the index without the watermark moving. When cooldown lifts, replay starts from the frozen `lastSeq+1` and re-sends that whole window, so the duplicate-store problem above hits every block in it instead of just one. The description justifies the freeze by post-gap blocks being dropped on a broken parent chain, but nothing here drops them, they all go through AddTask.

A seq guard alone will not cover this since `lastSeq` is frozen at that point. How about splitting it into a replay start point and a processed watermark?

pkg/kvevents/zmq_subscriber.go
```diff
@@ -143,5 +185,112 @@ func (z *zmqSubscriber) runSubscriber(ctx context.Context) {
...
+			if !z.hasLastSeq || seq > z.lastSeq {
```
bongwoobak: A container restart keeps the pod IP, so `EnsureSubscriber` sees the same endpoint, no-ops, and this subscriber survives with its `lastSeq` while vLLM's seq restarts at 0. After that `seq > lastSeq+1` never fires and this line never advances `lastSeq` either, so gap detection stays dead for the rest of the subscriber's life. It also makes the seq guard from the other two comments risky, every event after a restart sits below the watermark and would get skipped.

`seq < lastSeq` needs to reset the state and pull a full replay, otherwise that guard cannot go in safely.

vMaroon (approved):
Verified and tested, iterated async for speed.

PR #1895, opened 2026-07-05
Add DCGM data source and extractor for GPU utilization

noalimoy:
**What type of PR is this?**
/kind feature

The EPP routes requests using application-level metrics (KV cache, queue
depth) but has no GPU hardware utilization signal. This adds the
data-layer infrastructure to collect GPU metrics from a DCGM Exporter endpoints:

- `HTTPDataSource` port override to scrape a different port on the same
  pod IP (backward-compatible, default 0 preserves existing behavior)
- `dcgm-data-source` plugin that polls and parses Prometheus metrics
- `dcgm-extractor` plugin that reads `DCGM_FI_DEV_GPU_UTIL`, aggregates
  across GPUs (max), and stores normalized utilization in the endpoint's
  `AttributeMap`
- `GPUUtilization` attribute type with `Produces`/`Consumes` data key

A follow-up PR adds a filter and scorer that consume this attribute.

Part of #1706

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/attribute/gpu/data_types.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/gpu/data_types.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package gpu
+
+import (
+	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+)
+
+const (
+	// DCGMExtractorType is the plugin type for the DCGM metrics extractor.
+	DCGMExtractorType = "dcgm-extractor"
+)
+
+// GPUUtilizationDataKey identifies per-endpoint GPU utilization published
+// by the DCGM extractor and consumed by GPU-aware filters and scorers.
+var GPUUtilizationDataKey = plugin.NewDataKey("GPUUtilization", DCGMExtractorType)
+
+// GPUUtilization is a normalized GPU compute utilization value in [0.0, 1.0],
+// derived from DCGM_FI_DEV_GPU_UTIL (which reports 0-100).
+// For multi-GPU pods the extractor aggregates across visible devices.
+type GPUUtilization float64
+
+func (v GPUUtilization) Clone() fwkdl.Cloneable {
+	return v
+}
+
+// ReadGPUUtilization retrieves GPU utilization from an endpoint's AttributeMap.
+func ReadGPUUtilization(attrs fwkdl.AttributeMap, key string) (GPUUtilization, bool) {
+	return fwkdl.ReadAttribute[GPUUtilization](attrs, key)
+}
--- a/pkg/epp/framework/plugins/datalayer/extractor/dcgm/extractor.go
+++ b/pkg/epp/framework/plugins/datalayer/extractor/dcgm/extractor.go
@@ -29,7 +29,10 @@ import (
 	sourcemetrics "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/source/metrics"
 )
 
-const gpuUtilMetricName = "DCGM_FI_DEV_GPU_UTIL"
+const (
+	gpuUtilMetricName = "DCGM_FI_DEV_GPU_UTIL"
+	podLabelName      = "pod"
+)
 
 var _ fwkplugin.ProducerPlugin = &Extractor{}
 
@@ -65,17 +68,26 @@ func DCGMExtractorFactory(name string, _ *json.Decoder, _ fwkplugin.Handle) (fwk
 }
 
 // Extract parses DCGM_FI_DEV_GPU_UTIL from the Prometheus families,
-// aggregates across GPUs (max), normalizes to [0.0, 1.0], and stores
-// the result in the endpoint's AttributeMap.
+// keeps samples for this endpoint's pod (when the pod label is present),
+// aggregates across matching GPUs (max), normalizes to [0.0, 1.0], and
+// stores the result in the endpoint's AttributeMap.
 func (e *Extractor) Extract(_ context.Context, in fwkdl.PollInput[sourcemetrics.PrometheusMetricMap]) error {
 	family, ok := in.Payload[gpuUtilMetricName]
 	if !ok {
 		return fmt.Errorf("metric %q not found in DCGM response", gpuUtilMetricName)
 	}
 
+	podName := ""
+	if meta := in.Endpoint.GetMetadata(); meta != nil {
+		podName = meta.PodName
+	}
+
 	maxUtil := 0.0
 	found := false
 	for _, m := range family.GetMetric() {
+		if !sampleBelongsToPod(m, podName) {
+			continue
+		}
 		val := gaugeValue(m)
 		if val > maxUtil {
 			maxUtil = val
@@ -83,7 +95,7 @@ func (e *Extractor) Extract(_ context.Context, in fwkdl.PollInput[sourcemetrics.
 		found = true
 	}
 	if !found {
-		return fmt.Errorf("metric %q present but has no samples", gpuUtilMetricName)
+		return fmt.Errorf("metric %q present but has no samples for pod %q", gpuUtilMetricName, podName)
 	}
 
 	normalized := attrgpu.GPUUtilization(maxUtil / 100.0)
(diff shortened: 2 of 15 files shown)


pkg/epp/framework/plugins/datalayer/attribute/gpu/data_types.go
```diff
@@ -0,0 +1,45 @@
...
+// GPUUtilization is a normalized GPU compute utilization value in [0.0, 1.0],
+// derived from DCGM_FI_DEV_GPU_UTIL (which reports 0-100).
+// For multi-GPU pods the extractor aggregates across visible devices.
+type GPUUtilization float64
+
+func (v GPUUtilization) Clone() fwkdl.Cloneable {
+	return v
+}
+
+// ReadGPUUtilization retrieves GPU utilization from an endpoint's AttributeMap.
```
liu-cong: Nit: callers always pass `GPUUtilizationDataKey.WithNonEmptyProducerName(...).String()`. Consider making this a no-arg helper that encapsulates the key construction, since there's only one producer. Not blocking.

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -61,7 +68,9 @@ type HTTPDataSource[T any] struct {
 }
 
 // NewHTTPDataSource constructs a typed polling dispatcher.
-func NewHTTPDataSource[T any](scheme, path string, skipCertVerification bool,
+// portOverride, when non-zero, makes the source scrape podIP:portOverride
+// instead of the endpoint's MetricsHost.
+func NewHTTPDataSource[T any](scheme, path string, portOverride int, skipCertVerification bool,
```
liu-cong: Can we create an `Option` struct for optional args such as the port override, so `NewHTTPDataSource(existingargs, opts... Option)`

noalimoy: Fixed.

pkg/epp/framework/plugins/datalayer/source/dcgm/datasource.go
```diff
@@ -0,0 +1,89 @@
...
+// NewHTTPDCGMDataSource constructs a DCGMDataSource with the given parameters.
+// Use this function directly in tests to bypass JSON parameter marshaling.
+func NewHTTPDCGMDataSource(scheme, path string, port int, name string) (*http.HTTPDataSource[metrics.PrometheusMetricMap], error) {
+	return http.NewHTTPDataSource(scheme, path, defaultDCGMInsecureSkipVerify,
+		DCGMDataSourceType, name, parsePrometheus,
+		http.WithPortOverride(port))
+}
+
+// DCGMDataSourceFactory instantiates a dcgm-data-source plugin from configuration.
+func DCGMDataSourceFactory(name string, parameters *json.Decoder, _ plugin.Handle) (plugin.Plugin, error) {
```
elevran, line 59: should-fix: this package has no test files. DCGMDataSourceFactory's default params and scheme validation aren't covered - worth mirroring source/metrics/datasource_config_test.go's pattern for the defaults and invalid-scheme case.

elevran:
@noalimoy PR has been review and needs rebase/conflict resolution to progress.

elevran (commented):
Clean, well-tested addition - the extractor logic (missing metric, empty samples, multi-GPU max aggregation, normalization) is covered thoroughly, and the port-override mechanism on HTTPDataSource is backward-compatible for existing callers (variadic option, no signature break for source/metrics or source/models).

noalimoy:
Thanks @elevran for the review!
I've been out sick since last Tuesday (fell on my way to work:(  ) - and just got back today. I already have a local patch ready from last week for the DaemonSet support (`useNodeAddress` + pod-label filtering in the extractor), will push it soon along with `source/dcgm` factory tests.

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -231,5 +272,15 @@ func (s *HTTPDataSource[T]) AppendExtractor(ext fwkplugin.Plugin) error {
 }
 
 func (s *HTTPDataSource[T]) getEndpoint(ep Addressable) *url.URL {
-	return &url.URL{Scheme: s.scheme, Host: ep.GetMetricsHost(), Path: s.path}
+	host := ep.GetMetricsHost()
+	if s.portOverride > 0 {
```
elevran, line 279: `useNodeAddress` is only honored inside the `portOverride > 0` branch. 
Consider validating `useNodeAddress` requires a non-zero `portOverride` at construction time.

pkg/epp/framework/plugins/datalayer/extractor/dcgm/extractor.go
```diff
@@ -0,0 +1,140 @@
...
+			continue
+		}
+		val := gaugeValue(m)
+		if val > maxUtil {
+			maxUtil = val
+		}
+		found = true
+	}
+	if !found {
+		return fmt.Errorf("metric %q present but has no samples for pod %q", gpuUtilMetricName, podName)
```
elevran, line 101: Both this and the "metric not found" branch above return hard errors, incrementing the `DataLayerExtractErrorsTotal` on every poll cycle. 
For DaemonSet mode, this fires continuously whenever a pod isn't yet visible to the exporter (exporter restart/lag, pod just scheduled) - not necessarily a real error each time. 
While this matches and is consistent with the existing `models`/`metrics` extractor convention, which also hard-errors on a miss, worth confirming the resulting error-counter noise is acceptable for DaemonSet steady state with slow-starting pods.

pkg/epp/framework/plugins/datalayer/source/dcgm/datasource.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/source/metrics"
+)
+
+const DCGMDataSourceType = "dcgm-data-source"
+
+const (
+	defaultDCGMScheme             = "http"
+	defaultDCGMPath               = "/metrics"
+	defaultDCGMPort               = 9400
+	defaultDCGMInsecureSkipVerify = true
```
elevran, line 38: `insecureSkipVerify` defaults to `true` and the default `scheme` is `http` anyway, making the TLS default moot in the common case. This matches the other HTTP data sources so just flagging.
No action needed unless the DCGM uses TLS by default.

pkg/epp/framework/plugins/datalayer/source/dcgm/datasource.go
```diff
@@ -0,0 +1,95 @@
...
+	Path               string `json:"path"`
+	Port               int    `json:"port"`
+	InsecureSkipVerify bool   `json:"insecureSkipVerify"`
+	// UseNodeAddress scrapes the node IP (DaemonSet) instead of the pod IP (sidecar).
+	UseNodeAddress bool `json:"useNodeAddress"`
+}
+
+// NewHTTPDCGMDataSource constructs a DCGMDataSource with the given parameters.
+// Use this function directly in tests to bypass JSON parameter marshaling.
+func NewHTTPDCGMDataSource(scheme, path string, port int, name string, opts ...http.Option) (*http.HTTPDataSource[metrics.PrometheusMetricMap], error) {
```
elevran, line 52: nit: `NewHTTPDCGMDataSource` isn't called anywhere in this PR (including its own tests). Expect it's waiting on the addition of DCGM based filter/scorer.
May want to cover in tests in the meantime, otherwise it's unused exported symbol.

pkg/epp/framework/plugins/datalayer/extractor/dcgm/extractor.go
```diff
@@ -0,0 +1,140 @@
...
+// Produces declares the data key this extractor publishes.
+func (e *Extractor) Produces() map[fwkplugin.DataKey]any {
+	return map[fwkplugin.DataKey]any{e.dk: attrgpu.GPUUtilization(0)}
+}
+
+// sampleBelongsToPod reports whether m should contribute to this endpoint's
+// utilization. When podName is empty, or the sample has no pod label, the
+// sample is kept (sidecar / single-tenant payloads). When both are set, only
+// matching samples are kept (DaemonSet multi-pod payloads).
+func sampleBelongsToPod(m *dto.Metric, podName string) bool {
```
elevran, line 118: Unlabeled samples are kept regardless of `podName` (L121). Documented and intentional for the sidecar case, but for DaemonSet mode, if the exporter ever emits a mix of labeled and unlabeled samples for the same metric family (e.g., partial misconfiguration, mixed exporter versions during a rollout), unlabeled samples get silently folded into every pod's `max` aggregation. 
Edge case, not a bug given the explicit comment/tests - just flagging to confirm this is desired behavior.

pkg/epp/framework/plugins/datalayer/extractor/dcgm/extractor.go
```diff
@@ -0,0 +1,140 @@
...
+	for _, l := range m.GetLabel() {
+		if l.GetName() == name {
+			return l.GetValue(), true
+		}
+	}
+	return "", false
+}
+
+func gaugeValue(m *dto.Metric) float64 {
+	if g := m.GetGauge(); g != nil {
```
elevran, line 139: nit: returns `0` for a non-gauge metric rather than erroring, so if `DCGM_FI_DEV_GPU_UTIL` were ever exposed as a non-gauge type this would silently report 0% utilization instead of failing loudly. Low risk of DCGM changes, hence might want to err on the more strict side to surface.

elevran (commented):
PR is well scoped and follows repo's conventions. 
A few points below worth a look before merge; none are hard blockers.

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
noalimoy: added construction-time validation: `WithUseNodeAddress` now returns an error if `portOverride` is zero.

pkg/epp/framework/plugins/datalayer/extractor/dcgm/extractor.go
noalimoy: Yes, you're right - matches existing extractor convention. Noise is bounded to pod churn and self-heals once the exporter sees the pod.

pkg/epp/framework/plugins/datalayer/source/dcgm/datasource.go
noalimoy: added direct tests for `NewHTTPDCGMDataSource` (with and without `WithUseNodeAddress`).

pkg/epp/framework/plugins/datalayer/extractor/dcgm/extractor.go
noalimoy: intentional. DaemonSet mode requires `DCGM_EXPORTER_KUBERNETES=true` which labels all samples. Mixed labeled/unlabeled is a deployment misconfiguration, not a runtime scenario we need to guard against in code.

pkg/epp/framework/plugins/datalayer/extractor/dcgm/extractor.go
noalimoy: made `gaugeValue` strict - returns an error for non-gauge metrics instead of silent 0.

elevran (dismissed):

elevran (approved):

PR #1904, opened 2026-07-06
Add GPU utilization filter and scorer plugins

noalimoy:
**What type of PR is this?**
/kind feature

Adds GPU utilization-aware routing by wiring the DCGM extractor (#1895)
to the generic `endpoint-attribute-filter` and `endpoint-attribute-scorer`.

- The DCGM extractor emits `ScalarMetricValue` so the generic plugins
  can consume it directly — no dedicated filter/scorer code needed.
- Sample config (`sim-epp-gpu-config.yaml`) demonstrates the full
  pipeline: DCGM scrape → extract → filter overloaded endpoints
  (threshold 0.90, graceful fallback) → score by lowest utilization.
- Contract test validates the attribute key string stability.
- Smoke test entry validates config parse and plugin instantiation.

**Depends on**: #1895 (DCGM extractor and data source)

Part of #1706

**Release note**:
```release-note
GPU utilization-aware routing via the generic endpoint-attribute-filter and endpoint-attribute-scorer, configured against the DCGM extractor's ScalarMetricValue output.
```


liu-cong:
Please consider using the generic endpoint attribute scorer in https://github.com/llm-d/llm-d-router/issues/977, we can add a filter implementation as well.

noalimoy:
> Please consider using the generic endpoint attribute scorer in #977, we can add a filter implementation as well.

thanks @liu-cong. read through #977 and the review on
#1620 -- will adapt the scorer to use the generic `endpoint-attribute-scorer`
once it lands.

Is a generic attribute filter also planned? the `gpu-utilization-filter`
here has `threshold` + `graceful-fallback` logic that doesn't map to scoring,
so I want to know whether to keep it as-is or wait for a generic filter
framework.

liu-cong:
Thanks @noalimoy. We can create a generic filter too, see https://github.com/llm-d/llm-d-router/issues/1922 . Would you like to repurpose this PR for the generic filter implementation?

liu-cong:
Now we have the generic scorer and filter, would you like to update this PR? @noalimoy

noalimoy:
> Now we have the generic scorer and filter, would you like to update this PR? @noalimoy

Thanks @liu-cong !
@elevran - would you also prefer this direction? The DCGM extractor (now merged in #1895) currently stores a custom `GPUUtilization` type. To use the generic plugins, it would need to emit `ScalarMetricValue` instead. I can make that change and replace this PR's filter/scorer with a config example.

deploy/config/sim-epp-gpu-config.yaml
elevran, line 1: Is this intended as an example? If so, suggest folding relevant parts into the README and not as a separate deploy file.

pkg/epp/framework/plugins/datalayer/attribute/gpu/README.md
```diff
@@ -1,15 +1,18 @@
...
 
 ## Producers
 
-The following plugins produce this attribute:
-
 - **`dcgm-extractor`** (Data Layer): Extracts GPU utilization from the DCGM Exporter Prometheus endpoint.
+
+## Consumers
+
+- **`endpoint-attribute-filter`** (Scheduling): Filters endpoints exceeding a configurable GPU utilization threshold.
```
elevran: These are the generic ones, right? It could be read to imply that these are fixed consumers.
Would be clearer if you provide some additional description of the data collection (e.g., what needs to be enabled on the cluster and the EPP config) and how to use the generic filter/scorer with the plugin's produced data.
Include a sample config of the relevant plugins in the EPP config section (portions of the deploy/... file from above)

noalimoy: Good point — these are the generic plugins, not dedicated consumers. I'll slim down this README to just reference the `DataKey` constant (which is all this package exports), and expand the [`extractor/dcgm/README.md`](https://github.com/noalimoy/llm-d-inference-scheduler/blob/dcgm/gpu-filter-scorer/pkg/epp/framework/plugins/datalayer/extractor/dcgm/README.md) instead with:

- Clarification that consumers are the generic `endpoint-attribute-filter`/`scorer` (configurable, not hardwired)
- Cluster prerequisites (DCGM Exporter, `DCGM_EXPORTER_KUBERNETES=true` for DaemonSet)
- Full EPP config example showing the end-to-end pipeline

The extractor README is the natural home for this because it documents what the plugin produces and how to consume its output — keeping documentation co-located with the code that owns the contract. The `attribute/gpu/` package only defines a key constant and shouldn't duplicate usage instructions.

deploy/config/sim-epp-gpu-config.yaml
noalimoy: Yes, it's a sample config — same pattern as the other files in `deploy/config/` (e.g., `sim-epp-kvcache-config.yaml`, `pd-epp-config.yaml`).

I'll fold the relevant portions into the `extractor/dcgm/README.md` as a documented usage example either way (see next comment).

Would you prefer I also remove the standalone file, or keep it alongside the other samples for discoverability?

elevran: > Would you prefer I also remove the standalone file, or keep it alongside the other samples for discoverability?

You can keep it alongside. Not sure if/how useful these are.

elevran (approved):
Nice reuse of the existing generic endpoint-attribute-filter/scorer instead of adding dedicated GPU plugin code — keeps this small and avoids duplicating the threshold/scoring logic.
Thanks @noalimoy!

PR #1909, opened 2026-07-07
fix(epp/metrics): bound model-name label cardinality from user input

### LINKED ISSUE #1625: [Observaibility] epp/metrics: unbounded label cardinality from user-controlled inputs (metrics cardinality DoS)

gkneighb:
**What type of PR is this?**
/kind bug

Model-name labels on EPP request metrics come from the request body, which is not validated against a closed set of served models, and `client_golang` `*Vec` types never evict label combinations. A client sending distinct `model` values therefore grows the time series set without bound (metrics cardinality DoS).

Implements the shape converged on in the issue discussion (allowlist the enumerable set, bounded cap as backstop for the by-design passthrough path):

- A shared limiter in `pkg/epp/metrics` bounds every request-derived model label. Unconfigured names are admitted up to a cap of 1000 distinct values; beyond that they fold into a fixed `"other"` label.
- Names enumerated by `InferenceModelRewrite` rules (exact-match sources and rewrite targets) are **pinned** via `datastore.ModelRewriteSet` on reconciliation: they always emit their real label and do not consume cap slots, so a junk-name flood cannot displace a configured model into the overflow bucket.
- Admission is sticky (a value admitted once always returns itself), keeping paired gauge calls (running-requests inc/dec, flow-control queue add/sub) balanced.

Scope notes:
- Covers the model-labeled record functions in `pkg/epp/metrics` (request/latency/token/TTFT/TPOT/ITL/running-requests, flow-control queue gauges, and the rewrite-decision counter's request-derived source label). Vector 2 (`fairness_id`, header-derived) and pod-name churn from the issue are deferred follow-ups per the issue discussion.
- Plugin-local metrics with model labels (`profilehandler/disagg`, `dataproducer/predictedlatency`) have the same exposure but live outside the central metrics package; flagging for a follow-up rather than folding into this PR.
- One documented edge: a name that folded to `"other"` and is later pinned by a new rewrite rule can leave a small, bounded residue on the overflow series for gauge pairs in flight at that instant.

Fixes #1625

**Test plan**
- [x] Bounded-label unit tests: flood of 1000 distinct names stays capped; pinned (configured) names survive a full cap without consuming slots; empty names pass through
- [x] `configuredModelNames` extraction from InferenceModelRewrite rules (exact sources + targets, duplicates collapse)
- [x] `go test ./pkg/...`
- [x] `make test-integration-hermetic` (130 tests)
- [x] `make presubmit`

```release-note
EPP model-name metric labels are now bounded: names configured via InferenceModelRewrite always emit their real label, while unconfigured request-supplied names are limited to 1000 distinct values and collapse into an "other" label beyond that, preventing unbounded metric cardinality growth from arbitrary request bodies.
```

### DIFF
--- a/pkg/epp/datastore/datastore.go
+++ b/pkg/epp/datastore/datastore.go
@@ -38,6 +38,7 @@ import (
 	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
 	"github.com/llm-d/llm-d-router/pkg/epp/datalayer"
 	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+	"github.com/llm-d/llm-d-router/pkg/epp/metrics"
 	podutil "github.com/llm-d/llm-d-router/pkg/epp/util/pod"
 )
 
@@ -243,6 +244,9 @@ func (ds *datastore) ObjectiveGetAll() []*v1alpha2.InferenceObjective {
 }
 
 func (ds *datastore) ModelRewriteSet(infModelRewrite *v1alpha2.InferenceModelRewrite) {
+	// Configured model names always emit their real metric label; only
+	// unconfigured request-supplied names are subject to the cardinality cap.
+	metrics.PreAdmitModelLabels(configuredModelNames(infModelRewrite)...)
 	ds.mu.Lock()
 	defer ds.mu.Unlock()
 	ds.modelRewrites.set(infModelRewrite)
--- a/pkg/epp/datastore/modelrewritestore.go
+++ b/pkg/epp/datastore/modelrewritestore.go
@@ -145,6 +145,25 @@ func (ms *modelRewriteStore) getAll() []*v1alpha2.InferenceModelRewrite {
 	return rewrites
 }
 
+// configuredModelNames returns every model name an InferenceModelRewrite
+// enumerates: exact-match sources and rewrite targets. These form the closed
+// set of operator-configured names that metric labels must never fold into the
+// cardinality overflow bucket.
+func configuredModelNames(infModelRewrite *v1alpha2.InferenceModelRewrite) []string {
+	names := sets.New[string]()
+	for _, rule := range infModelRewrite.Spec.Rules {
+		for _, match := range rule.Matches {
+			if match.Model != nil {
+				names.Insert(match.Model.Value)
+			}
+		}
+		for _, target := range rule.Targets {
+			names.Insert(target.ModelRewrite)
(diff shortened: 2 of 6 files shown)


elevran (approved):
thanks - approve.
This handles a common case where one EPP instance reports on all. The real cardinality issue is in the Prometheus server, and the EPP code might not be able to handle that well - a restarted EPP or multiple concurrent EPP instances run with their own provate "seen" caches.

PR #1910, opened 2026-07-07
feat(epp/parsers/openai): make claimed path suffixes configurable

### LINKED ISSUE #1524: OpenAIParser: make claimed path suffixes configurable

Expose `OpenAIParser`'s claimed path suffixes as plugin configuration. Ship the current set (canonical endpoints + `chat/completions/render` + `completions/render` after #1515) as defaults; let users add further sub-paths via config without a code change.

Sketch:

```yaml
- type: openai-parser
  parameters:
    additionalPathClaims:
      - chat/completions/render
      - completions/render
```

`determineAPITypeFromPath` becomes prefix-driven — longest canonical endpoint prefix the claim sits under wins — so the API-type dispatch stays pure while the claim list becomes data-driven.

**Why is this needed**:

Surfaced by @ahg-g in review on #1515. The current hardcoded list requires a code change every time vLLM ships a new sub-path, and `/render` likely won't be the last. Worth aligning the approach with @zetxqx's broader `InferenceRequestBody` standardization in #1364 before implementation lands.

gkneighb:
**What type of PR is this?**
/kind feature

Makes the OpenAI parser's claimed path suffixes configurable, per the shape discussed on the issue:

- **Config**: `additionalPathClaims` on `openai-parser` parameters (mirrors `preadmitter/agentidentity`'s configurable-list-plus-defaults pattern; the factory now consumes its previously-discarded decoder). Extras merge into `Claims()` so the parser registry routes them to this parser.
- **Dispatch**: each claim is resolved at config-decode time to the deepest canonical endpoint it sits under, segment-aware -- `chat/completions/pipeline` parses as chat/completions, not bare completions -- via small unexported helpers in the openai package. Request-time dispatch consults the configured claims (deepest first) before the existing canonical suffix chain, which is untouched.
- **Validation**: a claim that sits under no canonical endpoint fails configuration loading with an error naming the claim and the valid endpoints, rather than silently dispatching as completions.

Fixes #1524

**Test plan**
- [x] Factory config table: extras merged into claims; strict decode rejects unknown fields; claims under no canonical endpoint, without a segment boundary (`fancycompletions`), or empty are rejected
- [x] Claim-resolution table incl. prefixed (`myservice/completions`), nested (`v2/chat/completions/stream`), and slash-wrapped forms
- [x] End-to-end regression: chat body on a configured custom sub-path parses as chat/completions (previously fell through to the completions default and misparsed)
- [x] `go test ./pkg/...`
- [x] `make test-integration-hermetic` (130 tests)
- [x] `make presubmit`

```release-note
The openai-parser now accepts an `additionalPathClaims` parameter listing extra path suffixes to route to it. Each claim must sit under a canonical OpenAI endpoint (segment-aware) and is parsed as that endpoint's API type; claims under no canonical endpoint fail configuration loading.
```

### DIFF
--- a/pkg/epp/framework/plugins/requesthandling/parsers/README.md
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/README.md
@@ -86,3 +86,17 @@ requestHandler:
   - pluginRef: openaiParser
   - pluginRef: anthropicParser
 ```
+
+### Custom path claims for `openai-parser`
+
+`openai-parser` accepts an `additionalPathClaims` parameter listing extra path suffixes to route to it, for OpenAI-compatible providers that expose sub-paths or prefixed variants of the canonical endpoints. Each claim must sit under a canonical OpenAI endpoint (`chat/completions`, `completions`, `embeddings`, `responses`, `conversations`) as a complete run of path segments; the request body is then parsed as that endpoint's API type. Matching is segment-aware, so `chat/completions/pipeline` is parsed as `chat/completions` rather than bare `completions`. A claim that sits under no canonical endpoint fails configuration loading.
+
+```yaml
+plugins:
+- name: openaiParser
+  type: openai-parser
+  parameters:
+    additionalPathClaims:
+    - chat/completions/pipeline
+    - myservice/embeddings
+```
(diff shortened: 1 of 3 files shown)


elevran:
@gkneighb please resolve conflict

PR #1917, opened 2026-07-07
fix(flowcontrol): clamp overflowing SLO TTFT header in ordering

anxkhn:
**What type of PR is this?**

/kind bug

The `slo-deadline` flow-control ordering policy derives each request's deadline
from the client-supplied `x-llm-d-slo-ttft-ms` header. Scaling an out-of-range
millisecond value to `time.Duration` can overflow and produce a deadline before
the request arrived. Because the policy dispatches the earliest deadline first,
that request can incorrectly jump ahead of SLO-bound requests.

This checks the arithmetic bound before scaling and assigns
`sloMaxDeadlineTime` to an out-of-range value, matching the existing handling
for invalid TTFT values. The bound is inline rather than defined as another
policy maximum.

None.

**Release note**:
```release-note
Fix a priority inversion in the slo-deadline flow-control ordering policy where an out-of-range `x-llm-d-slo-ttft-ms` request header could overflow the computed deadline and sort first.
```

**Test plan**:
- [x] An out-of-range TTFT receives the far-future deadline.
- [x] An out-of-range TTFT sorts after a valid SLO-bound request.


pkg/epp/framework/plugins/flowcontrol/ordering/slodeadline/slo_deadline.go
```diff
@@ -83,6 +84,11 @@ func (p *sloDeadlinePolicy) TypedName() plugin.TypedName {
 
 var sloMaxDeadlineTime = time.Unix(0, 1<<63-1)
 
+// sloMaxTTFTMilliseconds is the largest TTFT value (in milliseconds) that fits in
+// a time.Duration once scaled to nanoseconds. Larger values would overflow int64
+// and wrap negative, so they are rejected as invalid.
```
ahg-g: it doesn't get rejected, the sloMaxDeadlineTime gets applied in all cases; so my question is do we really need this (it defines another max which is confusing)?

anxkhn:
Pushed a revision addressing the latest review/CI feedback.

pkg/epp/framework/plugins/flowcontrol/ordering/slodeadline/slo_deadline_test.go
```diff
@@ -145,6 +151,16 @@ func TestCalculateSLODeadline(t *testing.T) {
 	accInvalid := &mocks.MockQueueItemAccessor{OriginalRequestV: reqInvalid}
 	assert.Equal(t, sloMaxDeadlineTime, calculateSLODeadline(accInvalid))
 
+	// Out-of-range value: a TTFT large enough to overflow time.Duration must be
+	// treated as invalid (far-future deadline), not wrapped into a past deadline.
+	reqOverflow := mocks.NewMockFlowControlRequest(5, "overflow", testFlowKey)
+	reqOverflow.ReceivedTimestampV = now
+	reqOverflow.InferenceRequestV = &scheduling.InferenceRequest{Headers: map[string]string{
+		sloTtftHeader: strconv.FormatInt(math.MaxInt64, 10),
```
LukeAVanDrie, line 159: Non-blocking nit: both new tests use `math.MaxInt64` ms, which is ~10⁶× past the cutoff, so they'd still pass if the guard's comparator or divisor were slightly wrong. Consider pinning the exact boundary instead so it turns these into off-by-one regression tests:

```go
// Exactly at the bound: still valid, no overflow.
bound := math.MaxInt64 / int64(time.Millisecond)
reqAtBound := makeSLOReq(now, strconv.FormatInt(bound, 10))
assert.Equal(t, now.Add(time.Duration(bound)*time.Millisecond),
	calculateSLODeadline(&mocks.MockQueueItemAccessor{OriginalRequestV: reqAtBound}))

// One past the bound: invalid, far-future deadline.
reqPastBound := makeSLOReq(now, strconv.FormatInt(bound+1, 10))
assert.Equal(t, sloMaxDeadlineTime,
	calculateSLODeadline(&mocks.MockQueueItemAccessor{OriginalRequestV: reqPastBound}))
```

(I ran this locally against your branch and both assertions hold.) Your call; the current tests already cover the reported bug.

LukeAVanDrie (dismissed):
Nice fix!

LukeAVanDrie:
@anxkhn this is ready to merge as soon as the lint / typos checks are resolved (and this is rebased onto main)

pkg/epp/framework/plugins/flowcontrol/ordering/slodeadline/slo_deadline_test.go
anxkhn: thanks, i kept the current regression coverage since it directly exercises the reported max input, as noted. the typo checks are fixed and the branch is rebased onto current main. @lukeavandrie, could you take another look when you have a chance?

LukeAVanDrie (approved):

PR #1919, opened 2026-07-07
epp: serve the metrics endpoint over mTLS

### LINKED ISSUE #1918: epp: serve the metrics endpoint over mTLS

An option to serve the EPP metrics endpoint over mTLS: `--metrics-cert-dir` (server tls.crt/tls.key) to serve TLS, and `--metrics-client-ca-file` (CA bundle) to require and verify a client cert.

**Why is this needed**:

The metrics endpoint only supports bearer-token auth or none. Scraping a spoke EPP aggregate from outside the cluster needs mutual authentication and integrity without minting and rotating bearer tokens.

hexfusion:
**What type of PR is this?**
/kind feature

Serves the EPP metrics endpoint over mTLS. `--metrics-cert-dir` supplies the server cert (`tls.crt`/`tls.key`) and enables TLS serving; `--metrics-client-ca-file` supplies a CA bundle and makes the endpoint require and verify a client certificate, so it can be scraped with mutual authentication instead of bearer-token auth.

Fixes #1918

```release-note
EPP: the metrics endpoint can now be served over mTLS via `--metrics-cert-dir` (server cert) and `--metrics-client-ca-file` (client CA).
```

### DIFF
--- a/pkg/epp/server/options.go
+++ b/pkg/epp/server/options.go
@@ -17,9 +17,13 @@ limitations under the License.
 package server
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"math"
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -27,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	"github.com/llm-d/llm-d-router/pkg/common/observability/logging"
 	"github.com/llm-d/llm-d-router/pkg/common/routing"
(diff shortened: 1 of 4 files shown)


elevran:
@hexfusion can you please resolve conflict?

hexfusion:
> @hexfusion can you please resolve conflict?

updated thanks

pkg/epp/server/options.go
```diff
@@ -102,6 +102,8 @@ type Options struct {
 	EnableCertReload        bool   // Enables certificate reloading of the certificates specified in --cert-path.
 	SecureServing           bool   // Enables secure serving.
 	MetricsEndpointAuth     bool   // Enables authentication and authorization of the metrics endpoint.
+	MetricsClientCAFile     string // PEM CA that requires a verified client cert on the metrics endpoint (mTLS).
+	MetricsCertDir          string // Directory with the metrics server cert (tls.crt/tls.key) that enables metrics TLS.
```
elevran: can't have client mTLS without the server side certificate and key also being configurable.
The `options.Validate()` checks should be extended to require `MetricsCertDir` when `MetricsClientCAFile` is set (one way dependency). Might also want `options.Complete()` to check existence and readability of all paths.

elevran (dismissed):
Overall - approve.
Leaving apporval, hold and three non-blocking suggestions. Remove the hold to merge or fix first - your call.

1. Sentinel naming: `errReadMetricsClientCA` and `errNoValidMetricsCA` live in `cmd/epp/runner/runner.go` while the rest of the mTLS-related errors live in `pkg/epp/server/options.go`. Worth colocating in case they ever escape the package boundary.
2. `checkReadable` opens-and-closes to verify access; `os.Stat` is one syscall lighter but loses the read-permission check. Either is fine.
3. minor test coverage gap (minor): `TestCompleteMetricsCertFiles` covers missing-dir and full-dir but not the partial case where `tls.crt` exists and `tls.key` is missing. That would pass `Complete()` and fail at server start, which is acceptable. Don't gate on this.

hexfusion:
these are reasonable changes Etai thanks applied

elevran (approved):

Issue #1574, opened 2026-06-10
P2P support


liranschour:
# P2P Connector — Feature Design

P2P Connector generalizes the [PD Connector](https://github.com/llm-d/llm-d-router/issues/1197) into a fully symmetric peer-to-peer mode. It reuses the same secondary-tier architecture (CPU KV cache in canonical layout, NIXL data path, ZMQ control path, unified bi-directional `P2PSession`) but drops the hard Prefiller / Decoder role distinction.

Every vLLM instance is a **peer**. For any given request, a peer plays one of two roles:

- **Consumer** (a.k.a. client): pulls KV blocks for the request from a remote peer's CPU cache instead of computing them locally.
- **Producer** (a.k.a. server): serves KV blocks from its CPU cache when a remote consumer asks for them.

A single peer can act as consumer for some requests and producer for others at the same time over the same session.

The orchestration layer indicates the request mode through `kv_transfer_params`. The same dict shape covers prefill/decode disaggregation and the symmetric P2P case.

## Orchestration-Level Protocol

`kv_transfer_params` is a dict carrying one or more of the following keys. Each key's value is itself a dict with the fields required for that mode.

Each key names the **remote counterpart** this peer transfers with (not this peer's own role), so the name reads as "the remote ___ I transfer with".

| Key | Set on | Value | Meaning |
|---|---|---|---|
| `remote_decoder` | Prefiller request | `{kv_request_id: str}` | This peer is the prefill producer for `kv_request_id`. Computed KV blocks must remain available for the remote decoder to pull. |
| `remote_prefiller` | Decoder request | `{kv_request_id: str, remote_host: str, remote_port: int}` | This peer is the decode consumer. Pull KV for this request from the remote prefiller at `remote_host:remote_port`, correlating via `kv_request_id`. |
| `remote_kv_source` | Consumer request | `{kv_request_id: str, remote_host: str, remote_port: int}` | This peer is a P2P consumer. Look up cached blocks at the remote source at `remote_host:remote_port` and pull whatever it has, correlating via `kv_request_id`. |

### Allowed combinations

The only legal multi-key combination is `remote_decoder` + `remote_kv_source`: a prefill producer is allowed to additionally act as a P2P consumer for the same request, looking up cached blocks at another peer to skip prefix prefill while still keeping its computed blocks available for a remote decoder to pull.

All other combinations are forbidden and must be rejected at the orchestration layer:

[...]

PR #1924, opened 2026-07-07
feat(epp): add p2p-source-producer for P2P prefix pulls

### LINKED ISSUE #1923: EPP: add p2p-source-producer to emit the KV cache source header
Part of #1929.

## Motivation

The prefix-cache producers already score every candidate pod's cached prefix blocks per request, but the result is only used for endpoint ranking. With the P2P Connector (#1574), a pod can pull cached prefix KV blocks from a peer's CPU cache instead of recomputing them — the EPP just needs to tell the sidecar which peer holds the blocks.

## Proposal

A new standalone plugin, `p2p-source-producer`, sets a header on the request to the decode sidecar:

```
x-kv-cache-source-host-port: <ip:port>
```

carrying the candidate pod holding the most cached prefix tokens for the request's prompt, across all candidate pods (prefill and decode roles).

The plugin consumes `PrefixCacheMatchInfo` the same way the prefix-cache-scorer and the PD decider do, so it is producer-agnostic: it works with the approximate producer out of the box and with the precise producer via `prefixMatchInfoProducerName` — the same mechanism that already gates the P/D disaggregation decision.

Behavior:
- `minCachedTokenDelta` (default 1): the header is set only when the best peer's cached prefix tokens exceed the computing pod's own by at least this delta. The computing pod is the selected prefill pod (P/D disaggregation) or the selected decode pod (no disaggregation). A tie or self-match never sets the header — pulling adds no blocks.
- The plugin is opt-in: deployments that don't list it are unaffected.
- The header is always addressed to the decode sidecar; the sidecar decides where to apply it (decode leg without disaggregation, prefill leg with — the `decode + p2p` combination allowed by #1574).

## Out of scope

Sidecar consumption of the header (injecting `kv_transfer_params.p2p`) — separate issue/PR on the sidecar side.

nilig:
**What type of PR is this?**

/kind feature

The prefix-cache producers score every candidate pod's cached prefix blocks per request, but the result is only used for endpoint ranking. With the P2P Connector (#1574) a pod can pull cached prefix KV blocks from a peer's CPU cache instead of recomputing them — the EPP just needs to tell the sidecar which peer holds the blocks.

Adds a standalone `p2p-source-producer` plugin: `Produce` stashes the candidate holding the most cached prefix tokens, and `PreRequest` sets `x-kv-cache-source-host-port` to that peer when it out-caches the pod computing the prefix (the prefill endpoint under P/D disaggregation, the decode endpoint otherwise) by at least `minCachedTokenDelta` tokens. The plugin consumes `PrefixCacheMatchInfo` the same way the prefix-cache-scorer and the PD decider do, so it is producer-agnostic — the approximate producer works out of the box, the precise producer via `prefixMatchInfoProducerName`. Opt-in: deployments that don't list the plugin are unaffected. Sidecar consumption of the header is out of scope (tracked under #1574).

Fixes #1923

**Test plan**:
- [x] Header set when the best peer out-caches the selected decode pod by at least the configured token delta
- [x] Header not set below the delta, on a self-match/tie, or when no candidate holds cached blocks
- [x] With a prefill profile, the comparison is against the prefill pod; a third pod that out-caches it sets the header
- [x] Spoofed inbound header is always removed
- [x] Factory defaults, config wiring, rejection of delta < 1, and the name-bound `PrefixCacheMatchInfo` dependency
- [x] Cluster (CoreWeave, 2 decode pods, approx producer, random picker): tcpdump on the decode pod shows `x-kv-cache-source-host-port: <peer>:8000` arriving on the first request scheduled to a pod that a peer out-cached; EPP TRACE shows token-based decisions (128 vs 0 sets the header; cold start, self-match, and ties skip)

**Release note**:
```release-note
New p2p-source-producer plugin sets the x-kv-cache-source-host-port header to the pod holding the most cached prefix tokens, enabling P2P prefix pulls; works with both the approximate and precise prefix-cache producers.
```

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -85,8 +85,8 @@ import (
 	reqdataprodprefix "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload"
-	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/kvcachesource"
 	mmproducer "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource"
 	preciseproducer "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache"
 	latencyproducer "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionid"
@@ -520,7 +520,7 @@ func (r *Runner) registerInTreePlugins() {
 	fwkplugin.Register(burstprefix.PluginType, burstprefix.Factory)
 	fwkplugin.Register(mmcacheaffinity.Type, mmcacheaffinity.Factory)
 	fwkplugin.Register(preciseproducer.PluginType, preciseproducer.PluginFactory)
-	fwkplugin.Register(kvcachesource.PluginType, kvcachesource.PluginFactory)
+	fwkplugin.Register(p2psource.PluginType, p2psource.PluginFactory)
 
 	// Flow Control plugins
 	fwkplugin.Register(globalstrict.GlobalStrictFairnessPolicyType, globalstrict.GlobalStrictFairnessPolicyFactory)
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -85,6 +85,7 @@ import (
 	reqdataprodprefix "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/kvcachesource"
 	mmproducer "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal"
 	preciseproducer "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache"
 	latencyproducer "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency"
@@ -519,6 +520,7 @@ func (r *Runner) registerInTreePlugins() {
 	fwkplugin.Register(burstprefix.PluginType, burstprefix.Factory)
 	fwkplugin.Register(mmcacheaffinity.Type, mmcacheaffinity.Factory)
 	fwkplugin.Register(preciseproducer.PluginType, preciseproducer.PluginFactory)
+	fwkplugin.Register(kvcachesource.PluginType, kvcachesource.PluginFactory)
 
 	// Flow Control plugins
 	fwkplugin.Register(globalstrict.GlobalStrictFairnessPolicyType, globalstrict.GlobalStrictFairnessPolicyFactory)
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/kvcachesource/producer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/kvcachesource/producer.go
@@ -1,0 +1,212 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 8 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/kvcachesource/producer.go
```diff
@@ -0,0 +1,212 @@
...
+
+// Config configures the kv-cache-source-producer.
+type Config struct {
+	// PrefixMatchInfoProducerName is the name of the data producer that
+	// produces PrefixCacheMatchInfo. Empty selects the default producer.
+	PrefixMatchInfoProducerName string `json:"prefixMatchInfoProducerName,omitempty"`
+	// MinCachedBlockDelta is the minimum number of cached blocks the best
+	// peer must hold beyond the pod computing the prefix for the header to
+	// be set. Must be >= 1; defaults to 1.
+	MinCachedBlockDelta int `json:"minCachedBlockDelta,omitempty"`
```
vMaroon: I think this should be in tokens and not blocks.

cmd/epp/runner/runner.go
```diff
@@ -518,6 +519,7 @@ func (r *Runner) registerInTreePlugins() {
 	fwkplugin.Register(burstprefix.PluginType, burstprefix.Factory)
 	fwkplugin.Register(mmcacheaffinity.Type, mmcacheaffinity.Factory)
 	fwkplugin.Register(preciseproducer.PluginType, preciseproducer.PluginFactory)
+	fwkplugin.Register(kvcachesource.PluginType, kvcachesource.PluginFactory)
```
vMaroon: nit: the new `kv-cache-source-producer` is registered but missing from `dataproducer/README.md`.

pkg/epp/framework/plugins/requestcontrol/dataproducer/kvcachesource/producer.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/llm-d/llm-d-router/pkg/common/routing"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requestcontrol"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
+	attrprefix "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/attribute/prefix"
+)
+
+const (
+	// PluginType is the registered type name of the kv-cache-source-producer.
+	PluginType = "kv-cache-source-producer"
```
vMaroon: The name `kv-cache-source` reads ambiguously from the EPP side: `DataSource`/`EndpointSource` already mean "raw data feed" in the datalayer, so `kv-cache-source-producer` parses as "a producer backed by the KV-cache data source" rather than "emits the peer to pull from". 

Perhaps can go for `p2p-source-producer` or something like that?

pkg/epp/framework/plugins/requestcontrol/dataproducer/kvcachesource/producer.go
nilig: Done — `minCachedTokenDelta`, compared in tokens via the endpoint's `PrefixCacheMatchInfo` block size, matching `nonCachedTokens`.

cmd/epp/runner/runner.go
nilig: Added.

pkg/epp/framework/plugins/requestcontrol/dataproducer/kvcachesource/producer.go
nilig: Renamed to `p2p-source-producer` (package `p2psource`). Kept the header as `x-kv-cache-source-host-port` — it names the semantics for the sidecar, not the plugin.

pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+For each request the plugin consumes the per-endpoint `PrefixCacheMatchInfo` produced by a prefix-cache producer (`approx-prefix-cache-producer` or `precise-prefix-cache-producer`), finds the candidate endpoint caching the most prompt tokens, and stashes it on the request. After scheduling it compares that best-match peer against the pod that will compute the prefix — the `prefill` profile target under P/D disaggregation, otherwise the primary target — and sets the header only when the peer out-caches the computing pod by at least `minCachedTokenDelta` tokens. Any inbound value of the header is removed. When no peer out-caches the computing pod, the request proceeds unchanged.
+
+**Parameters:**
+
+- `prefixMatchInfoProducerName` (string, optional): Name of the prefix-cache producer whose `PrefixCacheMatchInfo` to consume. Defaults to the default (unnamed) prefix-cache producer; set it to a named producer such as `precise-prefix-cache-producer` when using a non-default one.
```
vMaroon:
```suggestion
- `prefixMatchInfoProducerName` (string, optional): Name of the prefix-cache producer instance to consume `PrefixCacheMatchInfo` from, e.g. `precise-prefix-cache-producer`. Empty selects the default (unnamed) producer.
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/README.md
```diff
@@ -0,0 +1,46 @@
...
+- `minCachedTokenDelta` (int, optional, default: `1`): Minimum number of cached prompt tokens the best peer must hold beyond the computing pod for the header to be set. Must be `>= 1`. Higher values suppress pulls of short prefixes that are cheap to recompute.
```
vMaroon, line 12: IMO we should follow-up on a long-term effort of automatic profiling and tuning here. Profiling keeps coming up in tuning of several plugins, perhaps we can start pushing on this more actively (cc @hickeyma @kaushikmitr).

pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/README.md
nilig: Applied.

pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/README.md
nilig: Agree — `minCachedTokenDelta` is hand-tuned today, so a profiling/auto-tune pass would subsume it (and the similar thresholds across plugins). Out of scope for this PR; happy to open a follow-up issue to track the broader effort.

nilig: Opened #1943 to track this.

vMaroon (approved):

PR #1926, opened 2026-07-08
feat(standalone): make helm --set proxyType=envoy|agentgateway work without overlay files

abhay1999:
## What this does

Before this PR, `helm install --set router.proxy.proxyType=agentgateway` would silently ignore the built-in preset args for agentgateway and deploy the Envoy args instead (or deploy with empty args). You had to layer a separate `-f agentgateway-proxy.values.yaml` file to get the right behaviour. This PR removes that requirement — both proxy types now work with just `--set router.proxy.proxyType=...`.

## Root causes fixed

**1. `mergeOverwrite` in-place mutation (`_helpers.tpl`)**

`mergeOverwrite $preset $proxy` mutates `$preset` in place before returning. The preset args (from `presets.agentgateway.args` or `presets.envoy.args`) were being read back from `$preset` AFTER the merge, at which point they had already been overwritten by the top-level `proxy.args: []`. Fix: capture `$presetArgs` before the merge and restore them when the user has not supplied explicit args.

**2. Envoy ext_proc cluster always used TLS (`values.yaml`)**

The `ext_proc` cluster config hardcoded `transport_socket` (TLS) and `tls_options` in the health-check. When EPP runs with `--secure-serving=false` (required for standalone agentgateway mode), Envoy tried to TLS-handshake with a plaintext EPP, the health-check failed, `failure_mode_allow: true` let traffic through without endpoint selection, and requests returned 503 from the ORIGINAL_DST cluster. Fix: two new helpers (`envoyExtProcTLSOptions`, `envoyExtProcTransportSocket`) emit those blocks only when `router.epp.flags.secure-serving` is not set to `false`.

## Changes

- `config/charts/routerlib/templates/_helpers.tpl`
  - `llm-d-router.proxy`: capture preset args before `mergeOverwrite`, restore when user supplies none
  - `llm-d-router.proxy.eppSecureServing`: returns `"true"`/`"false"` based on `router.epp.flags.secure-serving`
  - `llm-d-router.proxy.envoyExtProcTLSOptions`: emits `tls_options` block only when EPP uses TLS
  - `llm-d-router.proxy.envoyExtProcTransportSocket`: emits `transport_socket` block only when EPP uses TLS

- `config/charts/llm-d-router-standalone/values.yaml`
  - Replace hardcoded TLS blocks with `{{ include ... | indent N }}` calls that respect the EPP TLS flag

## Testing

Tested end-to-end on Kind (Apple Silicon, no GPU):

```bash
# proxyType=envoy — no overlay file
helm install envoy-test $CHART \
  -f base.values.yaml \
  --set router.proxy.proxyType=envoy \
  --set router.epp.flags.secure-serving=false \
  ...

curl http://localhost:18092/v1/completions ...
# → {"choices":[{"text":"Give a man a fish..."}]} ✓

# proxyType=agentgateway — no overlay file  
helm install ag-test $CHART \
  -f base.values.yaml \
  --set router.proxy.proxyType=agentgateway \
  ...

curl http://localhost:18094/v1/completions ...
# → {"choices":[{"text":"Alas, poor Yorick..."}]} ✓
```

Both return valid responses. `helm lint` passes for both charts.

## Related

- PR #1600 (merged): pseudo-service model for agentgateway, removed `agentgateway-service.yaml`
- llm-d/llm-d#1719: docs PR that simplifies `base.values.yaml` (`args: []`) so the preset takes effect

/cc @danehans

Closes: N/A — this is a follow-up to the standalone agentgateway work started in #1600


danehans:
@ahg-g PTAL as this PR resolves your https://github.com/llm-d/llm-d/pull/1719#discussion_r3382194394 comment. This PR will allow users to simply pick an inference gateway implementation by setting `proxyType: envoy|agentgateway`. I tested PR head 9e71eac3834555184f3d731e69e7106239d40a19 in kind with both standalone proxy modes:

- `proxyType=envoy `with `router.epp.flags.secure-serving=false`: curl returned `HTTP_STATUS:200` with text `envoy route ok`.
- `proxyType=agentgateway`: curl returned `HTTP_STATUS:200` with text `agentgateway route ok`.
- Applied config confirmed Envoy restored its preset args and omitted `tls_options` / `transport_socket` for plaintext EPP.
- Applied config confirmed Agentgateway restored `["-f","/config/config.yaml"]` and EPP ran with `--secure-serving=false`.

ahg-g (approved):

PR #1927, opened 2026-07-08
feat(datalayer): extend file-discovery to support cluster endpoints

### LINKED ISSUE #1903: feat(datalayer): extend `file-discovery` to support cluster endpoints
### Problem

The `file-discovery` plugin's address validation only accepts IPv4 addresses, rejecting hostnames like `spoke-us-east.example.com`. For multi-cluster hub mode, cluster endpoints are addressed by hostname (the spoke gateway), not pod IPs.

### Proposed solution

Change the address validation to accept hostnames. No new struct fields needed — `MetricsHost` stays as `Address:Port` because the hub EPP scrapes metrics from the same spoke gateway address it routes traffic to. Set `PodName = ""` for cluster endpoints.

Hub-mode YAML uses the exact same format — no new fields:

```yaml
endpoints:
  - name: cluster-us-east
    address: spoke-us-east.example.com   # hostname → treated as cluster
    port: "443"
    labels:
      region: us-east
      model: llama-7b
  - name: cluster-eu-west
    address: spoke-eu-west.example.com
    port: "443"
    labels:
      region: eu-west
      model: llama-7b
```

### Acceptance Criteria

- [ ] Replace IPv4-only validation with non-empty address check
- [ ] Set `PodName = ""` when address is a hostname (not a valid IPv4)
- [ ] Unit tests — existing tests still pass (backward compatibility)
- [ ] Unit tests — hostname address (not IPv4): verify it passes validation and `PodName` is empty
- [ ] Unit tests — IPv4 address: verify `PodName` is still set (existing behavior)
- [ ] Unit tests — empty address: verify it fails validation
- [ ] Unit tests — `MetricsHost` is `Address:Port` for both hostname and IPv4 entries

### Context
[Design doc](https://docs.google.com/document/d/1FweyMiLXF5li3jTzfgd-mSyCvMLdfvnV9PNSXJs_Aqo/edit?usp=sharing)

liavweiss:
**What type of PR is this?**

/kind feature

Extends the `file-discovery` plugin to accept hostname addresses (not just IPv4), enabling multi-cluster hub mode where entire clusters are treated as endpoints.

Changes:
- Relax address validation to accept RFC 1123 hostnames using `validation.IsDNS1123Subdomain` from `k8s.io/apimachinery`
- Set `PodName = ""` for hostname-based entries (cluster endpoints), preserving `PodName = e.Name` for IPv4 entries (pod endpoints)
- Add optional `metricsPort` field to `EndpointEntry` so the metrics scrape port can differ from the traffic port (needed when the spoke EPP metrics endpoint is on a separate port from the spoke gateway)
- IPv6 addresses remain rejected (same as before)
- Fully backward compatible - existing pod YAML files with IPv4 addresses work identically

Hub-mode YAML example:
```yaml
endpoints:
  - name: cluster-us-east
    address: spoke-us-east.example.com
    port: "443"
    metricsPort: "9090"
```

Fixes #1903

**Note**:

The `metricsPort` field is optional and backward compatible. When omitted, `MetricsHost` defaults to `Address:Port` (existing behavior). When set, `MetricsHost` uses `Address:metricsPort`, allowing the hub EPP to scrape spoke EPP metrics on a different port than the traffic gateway port. This complements #1858 (mTLS scraper) and #1919 (mTLS metrics server).

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/discovery/file/README.md
+++ b/pkg/epp/framework/plugins/datalayer/discovery/file/README.md
@@ -18,10 +18,10 @@ datastore on each change.
 ## How It Works
 
 - **Initial load.** On `Start`, the file is read once. Each entry is
-  validated (address must be a parseable IP, port must be in `[1, 65535]`)
-  and applied via `notifier.Upsert`. Per-entry validation errors are logged
-  and the entry is skipped; file-level problems (open, parse, size > 1 MiB)
-  abort startup.
+  validated (address must be a valid IPv4 or RFC 1123 hostname, port must be
+  in `[1, 65535]`) and applied via `notifier.Upsert`. Per-entry validation
+  errors are logged and the entry is skipped; file-level problems (open,
+  parse, size > 1 MiB) abort startup.
 - **Reload (optional).** When `watchFile: true`, fsnotify Write / Create /
   Remove events trigger a reload. After an atomic rename or ConfigMap-style
   symlink swap (which destroys the inode being watched), the watcher is
@@ -42,11 +42,26 @@ plugin's `path` parameter.
 endpoints:
   - name: <string>              # required -- unique within the file
     namespace: <string>         # optional -- defaults to "default"
-    address: <IPv4>             # required -- must be a valid IPv4 address
+    address: <string>           # required -- IPv4 address or RFC 1123 hostname
     port: <string>              # required -- integer 1-65535 as a string
+    metricsAddress: <string>    # optional -- metrics scrape hostname (defaults to address)
+    metricsPort: <string>       # optional -- metrics scrape port (defaults to port)
     labels:                     # optional -- arbitrary key/value labels
       <key>: <value>
 ```
+
+Both IPv4 addresses and hostnames are accepted. `Name` is always set to the
+entry name regardless of address type.
+
+### Endpoint Type Label
+
+Each endpoint is tagged with the label `llm-d.ai/endpoint-type` so downstream
+plugins can distinguish pod endpoints from cluster endpoints explicitly. The
+label can be set in the endpoints file; when omitted, the plugin auto-detects
+it from the address format:
+
+- IPv4 address → `llm-d.ai/endpoint-type: pod`
+- Hostname → `llm-d.ai/endpoint-type: cluster`
 
 ## Configuration
 
@@ -75,7 +90,7 @@ dataLayer:
     pluginRef: file-discovery
 ```
 
-A two-endpoint file referenced by the config above:
+A two-endpoint file with pod IPs:
 
 ```yaml
 endpoints:
@@ -87,13 +102,38 @@ endpoints:
     port: "8000"
 ```
 
+A cluster endpoints file with host names instead of IP addresses.
+In multi-cluster deployments, `address` is typically a gateway that routes
+inference traffic, while metrics are scraped directly from the spoke EPP at
+a different host — hence `metricsAddress`:
+
+```yaml
+endpoints:
+  - name: cluster-us-east
+    address: gateway.apps.spoke-us-east.example.com
+    port: "443"
+    metricsAddress: epp.apps.spoke-us-east.example.com
+    metricsPort: "9090"
+    labels:
+      region: us-east
+  - name: cluster-eu-west
+    address: gateway.apps.spoke-eu-west.example.com
+    port: "443"
+    metricsAddress: epp.apps.spoke-eu-west.example.com
+    metricsPort: "9090"
+    labels:
+      region: eu-west
+```
+
 ## Limitations
 
 - The endpoints file is capped at 1 MiB.
-- `address` must be a literal IPv4 address. Hostnames are not resolved;
-  IPv6 is not supported.
-- Metrics are scraped from `address:port` (same host and port that serves
-  inference); separate metrics endpoints are not supported.
+- `address` must be a valid IPv4 address or an RFC 1123 hostname. IPv6 is
+  not supported. Hostnames are not resolved by the plugin; DNS resolution
+  happens at scrape/connect time.
+- Metrics are scraped from `metricsAddress:metricsPort`. When `metricsAddress`
+  is not set, it defaults to `address`. When `metricsPort` is not set, it
+  defaults to `port`.
 - File-discovery mode runs the EPP without a Kubernetes controller manager,
   so several K8s-only features are inactive: the `InferenceModelRewrite`
   and `InferenceObjective` reconcilers do not run, and any
--- a/pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
+++ b/pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
@@ -28,10 +28,12 @@ import (
 	"os"
 	"sort"
 	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/fsnotify/fsnotify"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/yaml"
 
@@ -39,15 +41,24 @@ import (
 	fwkplugin "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
 )
 
-const PluginType = "file-discovery"
+const (
+	PluginType = "file-discovery"
+
+	// EndpointTypeLabel distinguishes pod from cluster endpoints.
+	EndpointTypeLabel   = "llm-d.ai/endpoint-type"
+	EndpointTypePod     = "pod"
+	EndpointTypeCluster = "cluster"
+)
 
 // EndpointEntry is the YAML/JSON representation of a single endpoint.
 type EndpointEntry struct {
-	Name      string            `json:"name"                yaml:"name"`
-	Namespace string            `json:"namespace,omitempty" yaml:"namespace,omitempty"`
-	Address   string            `json:"address"             yaml:"address"`
-	Port      string            `json:"port"                yaml:"port"`
-	Labels    map[string]string `json:"labels,omitempty"    yaml:"labels,omitempty"`
+	Name           string            `json:"name"                      yaml:"name"`
+	Namespace      string            `json:"namespace,omitempty"       yaml:"namespace,omitempty"`
+	Address        string            `json:"address"                   yaml:"address"`
+	Port           string            `json:"port"                      yaml:"port"`
+	MetricsAddress string            `json:"metricsAddress,omitempty"  yaml:"metricsAddress,omitempty"`
+	MetricsPort    string            `json:"metricsPort,omitempty"     yaml:"metricsPort,omitempty"`
+	Labels         map[string]string `json:"labels,omitempty"          yaml:"labels,omitempty"`
 }
 
 // EndpointsFile is the top-level structure of the endpoints YAML/JSON file.
@@ -238,9 +249,12 @@ func (f *FileDiscovery) load(notifier fwkdl.DiscoveryNotifier) error {
 	incoming := make(map[types.NamespacedName]struct{}, len(ef.Endpoints))
 	var errs []error
 	for _, e := range ef.Endpoints {
-		if ip := net.ParseIP(e.Address); ip == nil || ip.To4() == nil {
-			errs = append(errs, fmt.Errorf("endpoint %q: invalid IPv4 address %q", e.Name, e.Address))
-			continue
+		ip := net.ParseIP(e.Address)
+		if ip == nil || ip.To4() == nil {
+			if len(validation.IsDNS1123Subdomain(strings.ToLower(e.Address))) > 0 {
+				errs = append(errs, fmt.Errorf("endpoint %q: address %q is not a valid IPv4 or hostname", e.Name, e.Address))
+				continue
+			}
 		}
 		if portNum, err := strconv.Atoi(e.Port); err != nil || portNum < 1 || portNum > 65535 {
 			errs = append(errs, fmt.Errorf("endpoint %q: invalid port %q", e.Name, e.Port))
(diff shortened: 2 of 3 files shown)


ezrasilvera:
Thank you for the PR, overall, it looks good. I added a few comments inside.

ezrasilvera:
Please go over the README and update it according to your proposed changes. Few examples:
`pkg/epp/framework/plugins/datalayer/discovery/file/README.md:45`

    address: <IPv4>             # required -- must be a valid IPv4 address

This is not true anymore, hostnames are allowed now. Two more lines in the same file are also out of date:

`pkg/epp/framework/plugins/datalayer/discovery/file/README.md:93`

    - `address` must be a literal IPv4 address. Hostnames are not resolved;

`pkg/epp/framework/plugins/datalayer/discovery/file/README.md:95`

    - Metrics are scraped from `address:port` (same host and port ...);
      separate metrics endpoints are not supported.

ezrasilvera:
nit:  (for the future) The address/port checks are inline in the file plugin. When we add more discovery plugins, it may be better to move this into a shared helper so all plugins use the same rules (including the metricsPort check from comment inside).

liavweiss:
Thank you @ezrasilvera, I made the necessary changes.

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
```diff
@@ -42,12 +42,17 @@ plugin's `path` parameter.
...
+    address: <string>           # required -- IPv4 address or RFC 1123 hostname
     port: <string>              # required -- integer 1-65535 as a string
+    metricsPort: <string>       # optional -- metrics scrape port (defaults to port)
     labels:                     # optional -- arbitrary key/value labels
       <key>: <value>
 ```
 
+When `address` is an IPv4, the endpoint is treated as a pod (`PodName` is set
+to the entry name). When `address` is a hostname, the endpoint is treated as a
+cluster (`PodName` is empty), enabling multi-cluster hub mode.
```
elevran: Q: how is the podName fields used? What's the downside of setting it non-empty?

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
```diff
@@ -87,13 +92,32 @@ endpoints:
     port: "8000"
 ```
 
+A cluster endpoints file for multi-cluster hub mode:
```
elevran: nit: is multicluster the only mode where hostnames are used?
```suggestion
A cluster endpoints file with host names instead of IP addresses:
```

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
```diff
@@ -42,12 +42,17 @@ plugin's `path` parameter.
 endpoints:
   - name: <string>              # required -- unique within the file
     namespace: <string>         # optional -- defaults to "default"
-    address: <IPv4>             # required -- must be a valid IPv4 address
+    address: <string>           # required -- IPv4 address or RFC 1123 hostname
     port: <string>              # required -- integer 1-65535 as a string
+    metricsPort: <string>       # optional -- metrics scrape port (defaults to port)
```
elevran, line 48: is this value reflected into the datalayer's metrics collection?
An alternative path might be to define a new data source to scrape remote endpoints by providing a label/tag with the hostname:port.
I think that the current endpoints assume an inference server and you are scraping remote clusters (EPPs, I presume?), so the use of the current metrics data source and extractors seem at odds with your design/goal.

elevran (commented):
Trying to get a high leve read on the goal first.
I believe this PR lays the ground work for treating remote cluster EPPs as endpoints for scheduling. 
Not sure the end-to-end flow for this has been considered. The current file plugin populates "Pods" from a file instead of a k8s reconciler. They both feed into the same data store Endpoint abstraction. From the point of ingestion and onwards, file-discovery endpoints and pod-reconciler endpoints land in the exact same `ds.pods` map/fwkdl.Endpoint abstraction, and nothing downstream (scheduler, scorers, metrics collector) distinguishes "came from file" vs "came from a real Pod" except via the fields on EndpointMetadata itself (PodName, Address, MetricsHost).
My concern is that the use of hostname and metrics ports could potentially break that assumption.

elevran:
Additional notes

1. `mmcacheaffinity` scorer and metrics counters use `PodName` purely for log/metric labels, so an empty PodName for cluster entries means their scheduler-attempt metrics and trace logs will show a blank pod label. Cosmetic, not a correctness bug, but worth knowing metrics dashboards keyed on pod will show blanks for hub-mode cluster targets.

1. There are no integration, e2e and coverage tests, only unit tests + one wiring smoke test. Since the existing file discovery changed only the ingest and not the actual endpoint properties, this was less of an issue. I think an integration and e2e tests are needed to confirm the changes in this PR. No need to spin up different clusters, but maybe using file based plugin with hostnames and ports for vllm-simulator would be sufficient to exercise the metrics-scrape path end-to-end with a real HTTP server behind a hostname+metricsPort.

1. A few downstream address usage that assumes IP-like semantics. A few places build raw host:port strings from Address for non-metrics purposes (disagg_profile_handler.go, p2psource/producer.go, preciseprefixcache). These will pass a hostname through fine syntactically (`net.JoinHostPort` doesn't care), but if any consumer assumes Address is dialable/resolvable in the same network namespace as the EPP (e.g., prefix-cache indexer keying, disaggregated P/D routing), a hostname pointing to a remote cluster's gateway may behave differently than a pod IP. That's inherent to the hub-mode feature intent, not a bug, but it's the kind of thing that would only surface once other plugins (P/D disaggregation, precise-prefix-cache) are exercised against cluster-type endpoints, which the PR doesn't add tests for. 

Overall, I would recommend considering which data sources and plugins would be useful for the cross cluster case and validate if they carry no IP address or other assumptions that are violated by the use case's input. Either conform to the assumptions or create cross-cluster variants of the relevant plugins to prevent unexpected failures from broken assumptions,
For example, I suspect the metrics collected are EPP level and not inference-engine level. The current extractor will not work well with that.

cc: @vMaroon @hexfusion

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
liavweiss: Yes, `metricsPort` feeds into `MetricsHost` which the metrics scraper uses to connect. You're right that the current core-metrics-extractor assumes vLLM-format metrics - that's why a spoke-epp engine type is being added in [PR #1928](https://github.com/llm-d/llm-d-router/pull/1928) to handle the different metric format from spoke EPPs. This PR only adds discovery support; it intentionally does not modify the metrics extraction logic.

elevran: Thanks for confirming.
I would prefer we don't overload the engine type for this but instead define a new Extractor for EPP metrics (which are **not** the same as inference engine metrics). You are essentially extracting different metrics from different objects for slightly different use cases and the overload can be confusing (e.g., introduce logic into existing components to separate server and cluster instead of just creating a different extractor for cluster).
Left comments on #1928 with more detail.

liavweiss:
> Additional notes
> 
> 1. `mmcacheaffinity` scorer and metrics counters use `PodName` purely for log/metric labels, so an empty PodName for cluster entries means their scheduler-attempt metrics and trace logs will show a blank pod label. Cosmetic, not a correctness bug, but worth knowing metrics dashboards keyed on pod will show blanks for hub-mode cluster targets.
> 2. There are no integration, e2e and coverage tests, only unit tests + one wiring smoke test. Since the existing file discovery changed only the ingest and not the actual endpoint properties, this was less of an issue. I think an integration and e2e tests are needed to confirm the changes in this PR. No need to spin up different clusters, but maybe using file based plugin with hostnames and ports for vllm-simulator would be sufficient to exercise the metrics-scrape path end-to-end with a real HTTP server behind a hostname+metricsPort.
> 3. A few downstream address usage that assumes IP-like semantics. A few places build raw host:port strings from Address for non-metrics purposes (disagg_profile_handler.go, p2psource/producer.go, preciseprefixcache). These will pass a hostname through fine syntactically (`net.JoinHostPort` doesn't care), but if any consumer assumes Address is dialable/resolvable in the same network namespace as the EPP (e.g., prefix-cache indexer keying, disaggregated P/D routing), a hostname pointing to a remote cluster's gateway may behave differently than a pod IP. That's inherent to the hub-mode feature intent, not a bug, but it's the kind of thing that would only surface once other plugins (P/D disaggregation, precise-prefix-cache) are exercised against cluster-type endpoints, which the PR doesn't add tests for.
> 
> Overall, I would recommend considering which data sources and plugins would be useful for the cross cluster case and validate if they carry no IP address or other assumptions that are violated by the use case's input. Either conform to the assumptions or create cross-cluster variants of the relevant plugins to prevent unexpected failures from broken assumptions, For example, I suspect the metrics collected are EPP level and not inference-engine level. The current extractor will not work well with that.
> 
> cc: @vMaroon @hexfusion

Correct - plugins like `disagg` and `precise-prefix-cache` are pod-level and wouldn't be configured in hub mode.
We can use for example `load-aware scorers` and `session-affinity` and add new hub-specific scorers as needed.

Agreed that integration testing is needed. This PR only changes what gets stored in the datastore at ingest - the actual scraping behavior is unchanged. Full integration testing of the hub-mode scrape flow will be done in a follow-up PR after this one and [#1928](https://github.com/llm-d/llm-d-router/pull/1928) are merged, since #1928 is the first PR that actually uses hostname+metricsPort for scraping.

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
liavweiss: PodName is used in three places:

1. `datastore.PodDelete()` - matches endpoints by PodName to delete them on pod lifecycle events. If we set PodName non-empty for cluster entries, a pod deletion event with a coincidentally matching name would accidentally remove a cluster endpoint from the datastore.
2. `metrics.go` - used as a Prometheus label on schedulerAttemptsTotal. Blank means the label is empty for cluster entries, which is cosmetic.
3. `mmcacheaffinity` scorer - used for debug logging only.

Setting it empty is intentional: it prevents `PodDelete` from accidentally matching cluster endpoints, and it's semantically correct since these entries represent clusters, not pods.

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
hexfusion: I agree with @elevran , les put the extra effort in here to make this feel more native and idiomatic.

liavweiss:
Hi @elevran , to address your concern about relying on empty `PodName` as a signal for cluster endpoints [https://github.com/llm-d/llm-d-router/pull/1928#discussion_r3620392199](url) - I've added an explicit `llm-d.ai/endpoint-type` label in the `file-discovery` plugin. The label is auto-detected from the address format (IPv4 → pod, hostname → cluster) and can also be set explicitly in the endpoints file. Yehudit's PR can use this label instead of checking `PodName == ""`.

pkg/epp/framework/plugins/datalayer/discovery/file/README.md
elevran: IIUC, PodDelete is called from the reconciler and the EPP enables only one endpoint discovery mechanism at a time - either file based or k8s reconcilation, but not both.

elevran:
> Hi @elevran , to address your concern about relying on empty `PodName` as a signal for cluster endpoints [https://github.com/llm-d/llm-d-router/pull/1928#discussion_r3620392199](url) - I've added an explicit `llm-d.ai/endpoint-type` label in the `file-discovery` plugin. The label is auto-detected from the address format (IPv4 → pod, hostname → cluster) and can also be set explicitly in the endpoints file. Yehudit's PR can use this label instead of checking `PodName == ""`.

what about an endpint "type" attribute that is set, instead of a label? The file plugin can set it on creation. I would prefer we give cluster endpoints a name (even if not a pod name).
Consider a different PR that replaces PodName (which is k8s centric) with a more generic endpoint name//identity?

hexfusion:
> what about an endpint "type" attribute that is set, instead of a label? The file plugin can set it on creation. I would prefer we give cluster endpoints a name (even if not a pod name).
Consider a different PR that replaces PodName (which is k8s centric) with a more generic endpoint name//identity?

I am picking this up, hope this helps

#2147

liavweiss:
Ok, I'll wait for #2147 to merge, then rebase and adapt my PR to focus on the hostname acceptance and metricsPort support

liavweiss:
Added `metricsAddress` field to support multi-cluster deployments where the routing target and the metrics scraping target are different services.
The hub routes inference requests to the spoke gateway, but scrapes pool-average metrics from the spoke EPP - which may have different endpoints.

elevran:
#2232 adds some changes file based discovery. 
@hexfusion @liavweiss - is there overlap or ordering between these (and #2147) as well?

pkg/epp/framework/plugins/datalayer/discovery/file/plugin.go
```diff
@@ -250,13 +264,36 @@ func (f *FileDiscovery) load(notifier fwkdl.DiscoveryNotifier) error {
...
+		}
+		mPort := e.Port
+		if e.MetricsPort != "" {
+			mPort = e.MetricsPort
+		}
+		labels := e.Labels
+		if labels == nil {
+			labels = make(map[string]string)
+		}
+		if v := labels[EndpointTypeLabel]; v != EndpointTypePod && v != EndpointTypeCluster {
```
elevran, line 279: nit:
when a user sets `llm-d.ai/endpoint-type` to an invalid value (e.g., typo `clustr`), the code silently overwrites it based on address-type auto-detection with no log line. A typo in a user's own config gets silently discarded rather than surfaced. IMO, worth a debug/info log when auto-correcting, so operators aren't confused why their explicit label didn't take.

liavweiss:
Should we keep this pr, or should that only live in a separate multicluster-file-discovery plugin (as in https://github.com/llm-d/llm-d-router/pull/2232)? Either way, the `metricsAddress/metricsPort` fields are still needed.
@elevran

hexfusion:
yeah I really appreciate your effort here,  this is mostly replaced by  #2232 the issue is generally how can we add this without affecting existing plugins. if there are outstanding items probably makes sense for a separate PR. :pray:

elevran:
> ... Either way, the metricsAddress/metricsPort fields are still needed.

I would prefer not to add endpoint type specific fields just yet so that we do not change anything relevant to intra-cluster serving.. For now this can sit as an attribute on the endpoint that's used by the multicluster metrics data source only.
Once the dust settles on this we can figure out how to refactor the endpoint type if needed (e.g., promote type to the interace, add metrics endpoint, ...). 

@hexfusion if this is fully replaced by #2232, can please close here to avoid redundant reviews? 
@liavweiss - thanks for the hard work! 🙏

hexfusion:
@liavweiss please close this in favor of #2232 i don't have the permissions :pray:

PR #1937, opened 2026-07-09
feat(sidecar): inject kv_transfer_params.p2p from the KV cache source header

### LINKED ISSUE #1930: Sidecar: inject kv_transfer_params.p2p from the KV cache source header
Part of #1929. Consumes the header emitted by the EPP's `p2p-source-producer` (#1923).

## Proposal

When a request arrives with `x-kv-cache-source-host-port: <ip:port>`, the sidecar tells vLLM to pull cached prefix blocks from that peer, per the #1574 protocol:

- **Without P/D disaggregation** (no `x-prefiller-host-port`): inject on the local vLLM request
  ```json
  "kv_transfer_params": {"p2p": {"kv_request_id": "<uuid>", "remote_host": "<ip>", "remote_port": <p2p-connector-port>}}
  ```
- **With P/D disaggregation** (both headers): add the same `p2p` block to the **prefill leg** alongside its existing `decode` block — the `decode + p2p` combination #1574 allows (the prefill producer skips prefix recompute by pulling from the peer, while still serving the decoder). The decode leg is unchanged.

`remote_host` comes from the header; `remote_port` is the sidecar's `--p2p-connector-port` (the header's port is the peer's serving port, not its P2P tier port). The `p2p.kv_request_id` is its own fresh UUID, independent of the PD correlation id — in P2P mode it is consumer-side only (#1574). The forbidden combinations (`prefill + p2p`) cannot arise by construction since the `p2p` block is only attached to legs that don't carry `prefill`.

## Tests

Unit (`pkg/sidecar/proxy`, following `connector_p2p_test.go`):
- Header present, no `x-prefiller-host-port`: the local vLLM request carries `kv_transfer_params.p2p` with `remote_host` from the header, `remote_port` from `--p2p-connector-port`, and a fresh `kv_request_id`
- Both headers: the prefill leg carries `decode` + `p2p` (the `p2p` block with its own `kv_request_id`, independent of `decode.kv_request_id`); the decode leg carries only `prefill` — never `prefill + p2p`
- No header: no `p2p` block injected (existing behavior unchanged)
- Malformed header value (not host:port): request proceeds without `p2p`, header not forwarded
- Two requests: distinct `p2p.kv_request_id` values

Cluster validation: two pods with pinned `PYTHONHASHSEED`, warm one pod's cache, route a same-prefix request to the other with the header set — the source pod logs `NIXL write_blocks`, the pulling pod reports `kv_offload_load_bytes > 0`.

## Out of scope

EPP-side selection of the peer (done in #1923); vLLM-side connector behavior (vllm-project/vllm#47636).

nilig:
**What type of PR is this?**
/kind feature

The EPP's `p2p-source-producer` sets `x-kv-cache-source-host-port` to the peer
holding the most cached prefix tokens. This PR consumes that header in the
routing sidecar and injects `kv_transfer_params.p2p` so vLLM pulls those blocks
over the OffloadingConnector P2P tier instead of recomputing them:

- Without P/D disaggregation, the decoder-only path injects `p2p` on the local
  request. It rebuilds `kv_transfer_params` from scratch (the sidecar owns that
  field, so client-supplied keys are dropped), routes through `dispatchDecode`
  so chunked decode still applies, and skips injection when the source resolves
  to the local pod.
- With P/D disaggregation, the prefill leg carries `decode` + `p2p` (the only
  legal multi-key combination).

Malformed, empty-host, disallowed, or non-offloading requests proceed unchanged.

Builds on the `p2p-source-producer` from #1924 (merged). The `offloading` connector's header handling and `p2p` injection are documented in `docs/disaggregation.md`.

Fixes #1930

**Release note**:
```release-note
The routing sidecar honors the `x-kv-cache-source-host-port` header, injecting `kv_transfer_params.p2p` so vLLM pulls cached prefix KV from the named peer over the OffloadingConnector P2P tier instead of recomputing it.
```

**Test plan** (new tests, all passing):
- [x] decoder-only path injects `p2p` and replaces any client-supplied `kv_transfer_params`
- [x] empty-host / malformed source header is ignored
- [x] source equal to the local pod IP skips injection (no self-pull)
- [x] disaggregated prefill leg carries `decode` + `p2p` with distinct request IDs

Cluster-validated on a 1P/1D OffloadingConnector deployment: the decoder-only path pulled a 128-token prefix cross-pod (`LookupRespMsg hits=2`, `kv_offload_load_bytes` on the puller, `write_blocks -> transfer_done` on the source).


pkg/sidecar/proxy/chat_completions.go
```diff
@@ -131,6 +136,34 @@ func (s *Server) disaggregatedPrefillHandler(apiType APIType) http.HandlerFunc {
...
+				kvCacheSource = ""
+			}
+		}
+		if kvCacheSource != "" {
+			span.SetAttributes(attribute.String("llm_d.pd_proxy.kv_cache_source", kvCacheSource))
+			// The disagg prefill leg reads the source from context (it is built
+			// deep under handlePDConnector). The decoder-only path uses the
+			// local value directly, so context is set only when disaggregating.
+			if len(prefillHostPort) > 0 {
+				ctx = context.WithValue(r.Context(), kvCacheSourceKey, kvCacheSource)
```
shmuelk: Is there a good reason why you are passing the kvCacheSource via the context and not as a parameter or a field in some struct?

`decodeWithP2PSource` gets kvCacheSource as a parameter...

pkg/sidecar/proxy/connector_p2p.go
```diff
@@ -66,11 +67,18 @@ func (s *Server) handleP2P(w http.ResponseWriter, r *http.Request, prefillPodHos
...
-	prefillData[requestFieldKVTransferParams] = map[string]any{
+	prefillKVParams := map[string]any{
 		requestFieldP2PDecodeParams: map[string]any{
 			requestFieldKVRequestID: kvRequestID,
 		},
 	}
+	// decode + p2p is the only legal multi-key combination: the prefill
+	// producer additionally pulls cached prefix blocks from the peer while
+	// keeping its computed blocks available for the decoder.
+	if src := kvCacheSourceFromContext(r.Context()); src != "" {
```
shmuelk: This function seems to get the original request, it could get the kvCacheSource from the header...

pkg/sidecar/proxy/connector_p2p_source_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+		decodeReqs := testInfo.decodeHandler.GetCompletionRequests()
+		Expect(decodeReqs).To(HaveLen(1))
+		Expect(decodeReqs[0]).ToNot(HaveKey(requestFieldKVTransferParams))
+
+		testInfo.cancelFn()
+		<-testInfo.stoppedCh
+	})
+
+	It("should not inject p2p params when the source is the local pod", func() {
```
shmuelk, line 224: The code above makes several checks of the provided header. It does not seem to check for the address of the local pod.

pkg/sidecar/proxy/connector_p2p_source_test.go
```diff
@@ -0,0 +1,229 @@
...
+		GinkgoT().Setenv("POD_IP", "10.9.9.9")
```
shmuelk, line 225: Its also not clear that the address of 10.9.9.9 will pass the SRF test. I would check for an address that passes the SSRF check.

pkg/sidecar/proxy/chat_completions.go
nilig: No good reason — switched to an explicit parameter. `kvCacheSource` now threads through the `pdConnectorHandler` signature into `handleP2P`/`handleNIXLV2`, matching `decodeWithP2PSource`. Dropped `kvCacheSourceKey` and the `context.WithValue`.

pkg/sidecar/proxy/connector_p2p.go
nilig: It can't re-read it: the handler validates the header and strips it (`r.Header.Del(routing.KVCacheSourceHeader)`) before dispatch, so a client-supplied value never reaches vLLM. `handleP2P` gets the already-validated value as a parameter instead.

pkg/sidecar/proxy/connector_p2p_source_test.go
nilig: The local-pod guard runs at injection, not in the header-validation switch: `decodeWithP2PSource` skips when `extractHost(source) == POD_IP` (decoder-only) and `addP2PPullToPrefill` skips when the source resolves to the prefiller (disagg). Both are covered — `should not inject p2p params when the source is the local pod` and `should not add p2p to the prefill leg when the source is the prefiller itself`.

pkg/sidecar/proxy/connector_p2p_source_test.go
nilig: `10.9.9.9` is a private pod-range address (the kind an InferencePool allows), but it passes here because the test explicitly disables the allowlist — `AllowlistValidator{enabled: false}`. An enabled validator requires a live InferencePool, which a unit test can't stand up, so a real SSRF-pass path isn't unit-testable; I added a comment on the `BeforeEach` saying exactly that. The malformed and empty-host cases still exercise header validation.

nilig:
**Update — the PR now also handles the P2P prefix pull when PD runs over NIXL, not just the offloading PD connector.**

The p2p prefix pull is always the OffloadingConnector's job; the PD transfer is either NIXL or the offloading connector. When PD is NIXL, the two compose via vLLM `MultiConnector(NixlConnector + OffloadingConnector-with-p2p-tier)` — NIXL moves KV prefill→decode, the OffloadingConnector pulls the cached prefix — so the sidecar injects the same `kv_transfer_params.p2p` block onto the NIXL prefill leg.

Changes:
- `--enable-p2p-pull` declares the OffloadingConnector p2p tier is present when the PD connector is NIXL: `p2pPullAvailable() = KVConnector==offloading || EnableP2PPull`.
- The header gate moves from `KVConnector != offloading` to `!p2pPullAvailable()` — behavior-preserving for every existing config (offloading still honors the header, plain NIXL still ignores it).
- `addP2PPullToPrefill` centralizes the `source != prefiller` self-guard and the `p2p` block build, shared by `handleP2P` and `handleNIXLV2`. The `p2p` key coexists with the flat NIXL fields; MultiConnector routes each to its connector.
- Injected at all three NIXL prefill-leg sites (WRITE-serial, READ-serial, parallel-dispatch).

Tests: new `connector_nixlv2_p2p_test.go` (composes-with-flag / ignores-without-flag / self-guard); full sidecar suite green; presubmit clean.

Validated on a 1P/1D `MultiConnector(NIXL+Offloading-p2p)` cluster: the prefill leg carries `do_remote_decode` + `p2p` together, PD transfers over NIXL/RDMA, and a cold consumer pulls the warmed prefix over the p2p tier (`kv_offload_load_bytes` > 0) while a control miss stays at 0. End-to-end pull requires the deployment to run the OffloadingConnector p2p tier with `offload_prompt_only: false` (the guide overlay covers this).

shmuelk (dismissed):

pkg/sidecar/proxy/options.go
```diff
@@ -269,7 +270,9 @@ func (opts *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.IntVar(&opts.MooncakeBootstrapPort, mooncakeBootstrapPortFlag, opts.MooncakeBootstrapPort,
 		"the port used to query the Mooncake bootstrap endpoint on prefill pods (only used with --kv-connector=mooncake)")
 	fs.IntVar(&opts.P2PConnectorPort, p2pConnectorPortFlag, opts.P2PConnectorPort,
-		"the prefiller's OffloadingConnector P2P tier listening port, injected as remote_port on the decode leg (only used with --kv-connector=offloading)")
+		"the prefiller's OffloadingConnector P2P tier listening port, injected as remote_port on the decode leg (used with --kv-connector=offloading or --enable-p2p-pull)")
+	fs.BoolVar(&opts.EnableP2PPull, enableP2PPull, opts.EnableP2PPull,
```
vMaroon, line 275: nit: `--enable-p2p-pull` is CLI-only, unlike its sibling `p2p-connector-port` and the other bool flags, so deployments configured through `--configuration`/`--configuration-file` can't enable it. I suggest adding `EnableP2PPull *bool `json:"enable-p2p-pull,omitempty"`` to `yamlConfiguration` and a `cfg.EnableP2PPull != nil && !opts.isFlagSet(enableP2PPull)` clause in `mergeYAMLConfiguration`.

nilig: Good catch — fixed in `1771e370`. Added `EnableP2PPull *bool` to `yamlConfiguration` and the mirrored clause in `mergeYAMLConfiguration` (`cfg.EnableP2PPull != nil && !opts.isFlagSet(enableP2PPull)`), so it now matches `--p2p-connector-port` and the other bool flags — config-file deployments can enable it, with the CLI flag still taking precedence. Added `enable-p2p-pull` to the config-file test fixtures + assertion too.

docs/disaggregation.md
roytman, line 579: --enable-p2p-pull has no row in either flags table and no prose explaining the NIXL+MultiConnector composition it requires

pkg/sidecar/proxy/options_test.go
roytman, line 231: Here and at 304
Both "flags override YAML" cases set enable-p2p-pull: true in the YAML fixture and assert `o.EnableP2PPull = true` in expected, but neither adds `enableP2PPull` to inputFlags. Since the CLI flag is never actually set, these assertions pass whether or not the merge logic's !`opts.isFlagSet(enableP2PPull)` guard exists — they prove "YAML sets the field," not "flag wins over YAML," which is what the third commit is actually about. Consider setting the CLI flag to the opposite of the YAML value in one of these cases.

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -168,6 +168,11 @@ func (s *Server) handleNIXLV2(w http.ResponseWriter, r *http.Request, prefillPod
 		}
 	}
 
+	// Compose the OffloadingConnector p2p pull onto the prefill leg when a source
```
roytman: nit: Same 3-line rationale comment is duplicated verbatim at this call site and at line 500, and the same fact is already stated in addP2PPullToPrefill's own doc comment. Consider trimming one copy.

pkg/sidecar/proxy/connector_ec_common.go
```diff
@@ -283,7 +283,10 @@ func (s *Server) runPDPipeline(
 
 	if len(prefillEndPoint) > 0 {
 		s.logger.V(logging.DEBUG).Info("using P/D protocol after encoder", "prefiller", prefillEndPoint)
-		s.handlePDConnector(w, pdRequest, prefillEndPoint, APITypeChatCompletions)
+		// The encoder path does not carry a KV cache source: the P2P prefix pull
+		// is not wired through encoder disaggregation. The empty source skips the
+		// p2p injection regardless of --enable-p2p-pull.
+		s.handlePDConnector(w, pdRequest, prefillEndPoint, "", APITypeChatCompletions)
```
roytman, line 289: I'm not sure that it is true. The encode phase is done, and now the sidecar can start the PD phase, and use the KV cache sources.

pkg/sidecar/proxy/connector_p2p.go
```diff
@@ -208,3 +211,79 @@ func (s *Server) handleP2PConcurrentRequests(w http.ResponseWriter, r *http.Requ
...
 	}
 }
+
+// p2pPullAvailable reports whether this deployment can pull cached prefix over
+// the OffloadingConnector P2P tier. That tier is the PD connector itself when
+// KVConnector is offloading, or is composed alongside NIXL via MultiConnector
+// (declared with --enable-p2p-pull) when the PD connector is NIXL.
+func (s *Server) p2pPullAvailable() bool {
+	return s.config.KVConnector == KVConnectorOffloading || s.config.EnableP2PPull
+}
```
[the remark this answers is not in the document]
nilig: Fixed. `p2pPullAvailable()` now gates on the connector: `offloading || (EnableP2PPull && KVConnector == nixlv2)`, so `--enable-p2p-pull` on any other connector no longer injects unroutable `kv_transfer_params.p2p`.

Added a fail-fast guard in `Validate()` too - `--enable-p2p-pull` is rejected unless `--kv-connector=nixlv2` (mirrors the existing `offloading` + `data-parallel-size>1` check), so the misconfig errors at startup rather than silently degrading routing.

docs/disaggregation.md
```diff
@@ -533,6 +533,8 @@ Specifies which KV transfer protocol the sidecar uses to coordinate prefill/deco
...
 
+When the request also carries the `x-kv-cache-source-host-port` header (set by the EPP `p2p-source-producer` to a peer holding more cached prefix than the pod computing the prefix), the sidecar injects an additional `p2p` key so vLLM pulls that cached prefix over the P2P tier instead of recomputing it. Under disaggregation the prefiller leg carries `{"decode": {...}, "p2p": {"kv_request_id": <own id>, "remote_host": <source host>, "remote_port": <p2p-connector-port>}}` (the only supported multi-key combination); without a prefiller the decoder-only request carries `{"p2p": {...}}` alone. A malformed or disallowed source header (or any source header when the connector is not `offloading`) is ignored and the request proceeds unchanged. For the pulled blocks to be servable, the source pod must offload its generated (decode-phase) KV: set `offload_prompt_only: false` in its `kv_connector_extra_config` (the default `true` offloads only prefill blocks).
```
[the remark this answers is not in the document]
nilig: Updated. The doc now states the source header is honored only by `offloading`, or `nixlv2` with `--enable-p2p-pull`, matching the actual gating.

pkg/sidecar/proxy/options_test.go
nilig: Fixed. The `flags override inline YAML` case now sets `--enable-p2p-pull=false` against the YAML's `true` and asserts `EnableP2PPull=false`, so it exercises the `!opts.isFlagSet(enableP2PPull)` merge guard - drop the guard and YAML's `true` wins and the test fails.

docs/disaggregation.md
nilig: Added. `--enable-p2p-pull` now has a row in the connector-specific flags table (keyed `nixlv2`) plus prose on the `MultiConnector(NixlConnector + OffloadingConnector)` composition it requires and the startup rejection on other connectors.

pkg/sidecar/proxy/connector_nixlv2.go
nilig: Trimmed. Both call sites now carry a one-line locator; the `MultiConnector` routing rationale lives only in `addP2PPullToPrefill`'s doc comment.

pkg/sidecar/proxy/connector_ec_common.go
nilig: Agreed the source could be used once encode is done and the PD phase starts. Keeping it out of this PR though - validating the full EPD path needs its own testing, so I've tracked it in #2007 rather than widen scope here.

vMaroon (approved):

PR #1949, opened 2026-07-09
add toolcalling telemetry to EPP

Gregory-Pereira:
**What type of PR is this?**
/kind feature

Adds tool-calling telemetry to the EPP so operators can verify that tool-calling parameters (tools, tool_choice, parallel_tool_calls) survive intact through the request processing pipeline.

  Problem: There is no visibility into whether tool-calling parameters are preserved, modified, or dropped as requests flow through the EPP. If a plugin or code path inadvertently mutates tool fields, there is no signal to detect it.

  Solution: A new `pkg/epp/toolcalling `package extracts a normalized ToolCallingSnapshot from the parsed PayloadMap before and after director.HandleRequest(). The snapshots are compared to produce a preservation status, which is emitted as Prometheus metrics and OTel trace span events. Snapshot metadata is propagated downstream via low-cardinality HTTP headers for sidecar and vLLM follow-up work.

  New Prometheus metrics (all ALPHA, subsystem llm_d_epp):
  - tool_calling_requests_total - requests by tool-calling presence
  - tool_choice_total - requests by tool_choice kind (unset/none/auto/required/named/unknown)
  - tool_definitions_count - distribution of tool definition count per request
  - parallel_tool_calls_total - requests by parallel_tool_calls value
  - tool_calling_preserved_total - requests by preservation status across the EPP boundary

  Propagation headers (for downstream components):
  - x-llm-d-tool-snapshot-hash
  - x-llm-d-tool-present
  - x-llm-d-tool-choice-kind
  - x-llm-d-tool-definitions-count
  - x-llm-d-tool-parallel-calls

### DIFF
--- a/pkg/epp/toolcalling/snapshot.go
+++ b/pkg/epp/toolcalling/snapshot.go
@@ -1,0 +1,259 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package toolcalling
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"sort"
+	"strconv"
+
+	"go.opentelemetry.io/otel/attribute"
+)
+
+const (
+	HeaderSnapshotHash      = "x-llm-d-tool-snapshot-hash"
+	HeaderToolPresent       = "x-llm-d-tool-present"
+	HeaderToolChoiceKind    = "x-llm-d-tool-choice-kind"
+	HeaderToolDefsCount     = "x-llm-d-tool-definitions-count"
+	HeaderParallelToolCalls = "x-llm-d-tool-parallel-calls"
+
+	ChoiceUnset    = "unset"
+	ChoiceNone     = "none"
+	ChoiceAuto     = "auto"
+	ChoiceRequired = "required"
+	ChoiceNamed    = "named"
+	ChoiceUnknown  = "unknown"
+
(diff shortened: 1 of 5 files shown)


ahg-g:
What is the status of this PR? @Gregory-Pereira @gyliu513 ?

pkg/epp/toolcalling/snapshot.go
```diff
@@ -0,0 +1,259 @@
...
+// body map and returns a normalized snapshot. Returns nil when none of
+// the tool-calling keys are present.
+func ExtractFromPayloadMap(m map[string]any) *ToolCallingSnapshot {
+	_, hasTools := m["tools"]
+	_, hasToolChoice := m["tool_choice"]
+	_, hasParallel := m["parallel_tool_calls"]
+
+	if !hasTools && !hasToolChoice && !hasParallel {
+		return nil
+	}
```
gyliu513, line 78: If none of `tools`, `tool_choice`, or `parallel_tool_calls` are present, the function returns `nil`. Then in `PreservationStatus`:

```go
if inbound == nil || outbound == nil {
    return PreservedUnknown
}
```

This means the worst-case scenario this PR is meant to catch, a plugin stripping the tool-calling fields entirely between inbound and outbound, produces the exact same `PreservedUnknown` result as a normal request that never used tool calling at all. `tool_calling_preserved_total{status="false"}` will never fire for a total field drop; it only fires when both snapshots are non-nil but their hashes differ.

This is compounded in server.go: both the outbound trace span event and the header propagation are gated on `outboundToolSnapshot != nil`, so a total drop is invisible in metrics, traces, and headers alike.

gyliu513 (commented):
Hi @Gregory-Pereira , any plan to update this PR?

pkg/epp/toolcalling/snapshot.go
```diff
@@ -0,0 +1,259 @@
...
+	ToolDefinitionsCount int
+	ToolDefinitionsHash  string
+	ParallelToolCalls    string
+	SnapshotHash         string
+}
+
+// ExtractFromPayloadMap reads tool-calling fields from a parsed request
+// body map and returns a normalized snapshot. Returns nil when none of
+// the tool-calling keys are present.
+func ExtractFromPayloadMap(m map[string]any) *ToolCallingSnapshot {
```
ahg-g, line 71: if this is protocol specific (i.e., OAI), then we need to place this logic in the corresponding parser. @zetxqx pls advise.

PR #1953, opened 2026-07-09
Add tool-calling preservation telemetry to routing sidecar

Gregory-Pereira:
**What type of PR is this?**
  /kind feature

  Adds the sidecar's half of the cross-component tool-calling preservation contract established in #1949. The EPP extracts a tool-calling snapshot from each request and propagates it via `x-llm-d-tool-*` headers. The sidecar now reads those headers, emits a preservation metric, adds OTel span events, and forwards the headers downstream to vLLM.

  This is the sidecar's first Prometheus metric. Changes:

  - New `pkg/sidecar/proxy/metrics.go`: dedicated Prometheus registry with `llm_d_sidecar_tool_calling_preserved_total` counter
  - `/metrics` endpoint on the existing HTTP mux (alongside `/health`)
  - Header bridging helper to convert `http.Header` to the flat `map[string]string` the toolcalling package expects
  - Telemetry block in `disaggregatedPrefillHandler()`: reads upstream snapshot, records preservation, emits span events
  - Unit tests for header bridging, metrics endpoint, and header forwarding through the handler

  The sidecar does not modify tool-calling fields, so preservation is reported as `"true"` when EPP headers are present. This detects regressions if future code mutates the body, and surfaces `"unknown"` when the EPP is absent.

  Depends on #1949


ahg-g:
@Gregory-Pereira is this PR still relevant?

PR #1956, opened 2026-07-09
feat(epp): add soft-reflective-ceiling usage limit policy

### LINKED ISSUE #1954: Priority-aware graduated backpressure flow-control plugin
**What would you like to be added**: A new flow-control usage-limit policy that gates lower-priority bands proportionally as pool saturation rises, keeping the highest-priority band open at all times. Where the existing `static-usage-limit-policy` caps dispatch uniformly across all priorities, the proposed `soft-reflective-ceiling-policy` graduates the ceiling by band rank so that the flow controller reserves headroom for critical traffic under load. Gated items remain queued rather than being rejected — this is dispatch gating, not admission control.

For each dispatch decision the policy computes `ceiling[i] = 1 - i * saturation / (N - 1)`, where band 0 is the highest priority. Once a lower band hits its reflective ceiling, its ceiling alternates between fully open and fully closed with period `round(saturation / (1 - saturation))`, so its effective dispatch rate degrades in proportion to saturation rather than falling off a cliff at a fixed threshold. The algorithm takes no parameters — behavior is derived entirely from the number of active priority bands and the observed saturation.

More details [here](https://docs.google.com/document/d/1yJuBOBV7fGEZX25YcikDjk8eRBdUYnDthudzc5FTtF0/edit?tab=t.0#heading=h.ijjvy5t0q55r).

**Why is this needed**:

1. **Graceful degradation**: lower-priority traffic is gated progressively rather than all at once, so the pool absorbs saturation without a step-function drop in throughput for non-critical classes.
2. **Latency protection for critical traffic**: the highest-priority band is never gated, and each lower band is gated before higher bands as saturation rises, preserving headroom without pre-reserving capacity.
3. **Tuning-free**: no thresholds to configure. Signal conditioning (smoothing, hysteresis) is delegated to the `SaturationDetector` layer per the `UsageLimitPolicy` contract, so operators tune the detector once and the policy adapts.
4. **Complements existing mechanisms**: sits alongside the sibling `static-usage-limit-policy` (uniform ceiling, framework-injected by default) and the probabilistic admission approach proposed in #925 (rejection-based shedding for negative-priority classes). Operators can pick the tradeoff that matches their workload.

Infrastructure impact is minimal: the plugin implements the existing `UsageLimitPolicy` interface and is opt-in via `flowControl.usageLimitPolicyPluginRef`. A sample deployment configuration and a policy README documenting when to prefer it over the static default are included.

Co-Authors: @jgchn @toslali-ibm / @mtoslalibu

kalantar:
**What type of PR is this?**
/kind feature

Adds a `soft-reflective-ceiling-policy` usage-limit plugin that implements graduated, priority-aware dispatch gating. The highest-priority band is never gated; for each lower band the policy computes a reflective ceiling `1 - i * saturation / (N - 1)` where `i` is the band's rank (0 = highest) and `N` is the number of active bands. Once a band reaches its ceiling, its state alternates fully open and fully closed with period `round(saturation / (1 - saturation))`, so the effective dispatch rate degrades in proportion to saturation rather than falling off a cliff at a fixed threshold. Gated items remain queued; the ceiling controls dispatch, not admission.

Contrast with the existing `static-usage-limit-policy` (uniform ceiling across all priorities, framework-injected by default): this policy graduates the ceiling by band rank so critical traffic keeps flowing while lower priorities are throttled first. It has no tunable parameters — behavior is derived entirely from the priority count and observed saturation, with signal conditioning delegated to the `SaturationDetector` per the `UsageLimitPolicy` contract.

Includes plugin factory registration in the EPP runner, a sample `deploy/config` scenario configuration, and a policy README documenting when to prefer it over the static default.

Fixes #1954

```release-note
Add soft-reflective-ceiling-policy: an opt-in usage-limit plugin that gates lower-priority traffic proportionally as pool saturation rises, while never gating the highest-priority band.
```


pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/policy.go
```diff
@@ -0,0 +1,141 @@
...
+	return newPolicy(name, logger), nil
+}
+
+type policy struct {
+	name string
+
+	mu sync.Mutex
+	// counters holds one tick counter per band. Pointers (rather than values)
+	// ensure slice growth never copies an atomic value.
+	counters []*atomic.Int64
```
shmuelk: As you are not storing the actual priority associated with each entry, what happens if a priority band is added dynamically (via a new InferenceObjective) ?

I think this should be a map[int]*atomic.Int64. The locking gets more intersting.

kalantar: Fixed. `sync.Map` keyed by priority; `LoadOrStore` handles first-write contention, so we can drop the explicit mutex. Added a test case.

pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/policy.go
```diff
@@ -0,0 +1,142 @@
...
+
+		switch {
+		case saturation < reflectiveCeiling:
+			ceilings[i] = 1.0
+		case saturation >= 1.0:
+			ceilings[i] = 0.0
+		default:
+			// 1e-9 guards against round-off when saturation is very near 1.0.
+			period := int64(math.Max(1, math.Round(saturation/(1.0-saturation+1e-9))))
+			tick := p.counterFor(priority).Add(1)
```
LukeAVanDrie: Per-priority counters desynchronize and will starve lower bands entirely.

The counter for a band only advances while that band is in this gated branch, and bands enter gating at different saturation levels (band N-1 first). So under any gradual saturation ramp the counters end up phase-shifted. Because `period = round(sat/(1-sat))` depends only on saturation, all gated bands share the same period P and their offsets stay constant mod P. If the offsets differ, a lower band's open ticks land permanently on a higher band's closed ticks.

That matters because `dispatchCycle` stops the entire cycle at the first band where `saturation >= ceilings[i]`; it never reaches lower bands on that call. Repro I ran locally (simulating the processor's gate loop): priorities `[100, 0, -50]`, 3 calls at saturation 0.6 (band 2 gated alone), then 100 calls at 0.7 (both gated, period 2):

```
band 2 open ticks: 50/100 (expected ~50 per README)
band 2 open ticks actually reachable through processor: 0/100
```

Band 2 gets zero dispatch for as long as saturation holds which is strictly worse than the static policy at the same saturation, and the desync phase is hit with probability (P-1)/P per band pair on any monotonic ramp. Lockstep only survives a step change that gates all bands at once.

Suggested fix: one policy-wide tick, shared by all gated bands. Since ceilings strictly decrease with rank, the gated set is always a suffix, so a shared tick makes the per-call output monotone (all gated bands open together) and no open tick is wasted behind a closed higher band:

```go
type policy struct {
	name string
	// tick advances once per ComputeLimit call in which at least one band is
	// gated. All gated bands share it, so their open ticks coincide and the
	// per-call gating decision is monotone across ranks — required because
	// the dispatch loop stops at the first gated band.
	tick atomic.Int64
}
```

with `tick.Add(1)` hoisted out of the per-band loop (once per call when any band alternates). This also removes the `sync.Map`, the unbounded counter growth under dynamic-priority churn, and most of the statefulness that the contract discussion below is about. `TestComputeLimit_Alternation*` need phase updates and `TestComputeLimit_StatePersistsAcrossPrioritySetChanges` becomes obsolete; a new test covering the rising-saturation scenario above would lock the fix in.

pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/policy.go
```diff
@@ -0,0 +1,142 @@
...
+	return newPolicy(name, logger), nil
+}
+
+type policy struct {
+	name string
+
+	// counters maps priority value to a tick counter. Keyed by priority (not
+	// rank) so that adding or removing a priority band at runtime does not
+	// reassign an existing band's counter to a different priority.
+	counters sync.Map
```
LukeAVanDrie: The `UsageLimitPolicy` interface doc (`interface/flowcontrol/plugins.go`) says implementations "MUST be stateless: it is a pure function". You document the deviation honestly here and in the README, but post-merge the contract statement would simply be false. Since the alternation mechanism is this policy's differentiator (and it can't be expressed statelessly), I think the right move is to amend the interface comment in this PR -- e.g. "SHOULD be stateless; small bounded dispatch-spreading state (tick counters) is permitted, signal conditioning is not" -- rather than leave the two contradicting each other.

Relatedly, the README's Trade-offs section says the contract "prefers" stateless implementations; today it says MUST, so whichever way this lands, the two texts should agree.

pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/README.md
```diff
@@ -0,0 +1,69 @@
...
+  - type: soft-reflective-ceiling-policy
+flowControl:
+  usageLimitPolicyPluginRef: soft-reflective-ceiling-policy
+```
+
+Unlike the static usage limit policy, this policy is **not** framework-injected. Declare it explicitly to activate it.
+
+## Trade-offs
+
+- **Not Stateless**: The `UsageLimitPolicy` contract prefers stateless implementations. The alternation pattern is only observable across calls, so the policy keeps a small, bounded, per-band atomic counter. Concurrency safety is preserved via pointer-atomic counters with a mutex guarding slice growth.
```
LukeAVanDrie: Non-blocking nit: this sentence likely describes an older implementation. The code uses a `sync.Map`, and there's no mutex or slice. (If you adopt the shared-tick fix, this bullet gets simpler still: a single atomic counter, no map.)

pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/README.md
```diff
@@ -0,0 +1,69 @@
...
+**Reflective ceiling per band:**
+
+    ceiling[i] = 1 - i * saturation / (N - 1)
+
+where `N = len(priorities)`. Band 0 is never gated (`ceiling[0] = 1.0`).
+
+**Per-band decision:**
+
+- If `saturation < ceiling[i]`: the band is fully open (`1.0`); items dispatch normally.
+- If `saturation >= 1.0`: non-critical bands are fully gated (`0.0`); their items remain queued until saturation drops.
```
LukeAVanDrie, line 30: Non-blocking nit: this (together with "Band 0 is never gated" above) reads as if the critical band keeps dispatching at full saturation. It won't: the flow controller gates on `saturation >= ceiling`, so at saturation ≥ 1.0 even ceiling 1.0 blocks. The controller halts all dispatch, per the `SaturationDetector` contract. 

Suggest one clarifying sentence, e.g.: "At saturation ≥ 1.0 the Flow Controller itself stops dispatching all bands, including the highest; this policy's ceiling of 1.0 for band 0 matters only while saturation is below 1.0."

pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/policy.go
LukeAVanDrie, line 1: missing license header

pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/policy_test.go
LukeAVanDrie, line 1: missing license header

LukeAVanDrie (changes requested):
Thanks for driving this! Two things before this can merge:

1. Confirmed starvation bug from per-priority counter phase desync: Under any gradual saturation rise, bands enter gating at different times, so their counters end up phase-shifted. Since `dispatchCycle` aborts the whole cycle at the first closed band, a lower band whose open ticks land on a higher band's closed ticks dispatches never. The fix is small (single shared tick counter) and detailed inline.
2. The statelessness contract needs an explicit decision. The interface says implementations MUST be stateless; this PR knowingly deviates. I'd rather we amend the contract wording in this PR than merge code that contradicts it. Flagging for discussion rather than asking you to redesign.

Plus some small doc/convention nits inline. Happy to re-review quickly once the counter fix is in.

pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/policy_test.go
kalantar: addressed

pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/policy.go
kalantar: addressed

pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/README.md
kalantar: attempted to clarify

pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/README.md
kalantar: Included change in commit https://github.com/llm-d/llm-d-router/pull/1956/commits/173f7609542e0d68d06051289d3341f3d200dc4b along with change in policy statement.

pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/policy.go
kalantar: Replaced the `sync.Map` of per-priority counters with a single policy-wide `atomic.Int64` tick that all gated bands share. Ceilings are now monotone across ranks per call — every gated band opens together or every one closes together — so a lower gated band's open ticks are always reachable through the dispatch loop. Regression test `TestComputeLimit_NoStarvationUnderRisingSaturation` reproduces your scenario (warm up at sat=0.6, then run at sat=0.7 with three bands) and asserts monotonicity plus equal open counts across bands 1 and 2. `TestComputeLimit_StatePersistsAcrossPrioritySetChanges` retired as you noted; the two `TestComputeLimit_Alternation*` tests pass unchanged (at N=2 there's no phase-desync possible, so shared-tick and per-band counters produce identical output).

pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/policy.go
kalantar: Suggested change in commit https://github.com/llm-d/llm-d-router/pull/1956/commits/173f7609542e0d68d06051289d3341f3d200dc4b

LukeAVanDrie (approved):
Thanks for driving this! LGTM

pkg/epp/framework/plugins/flowcontrol/usagelimits/softreflectiveceiling/README.md
LukeAVanDrie: thanks, looks good now!

LukeAVanDrie:
@ahg-g This new plugin LGTM. Because it registers the plugin in runner.go, I don't have perms to approve. Can you endorse just that portion?

liu-cong (approved):

PR #1959, opened 2026-07-10
epp: let parsers own model-name extraction and rewrite

### LINKED ISSUE #1900: Ability to directly set the model name in plugins

Would it be possible to pass a RequestContext object to the plugins so that they can write directly the model name to it? Alternatively, the director could read the model from the InferenceRequest object that plugins are producing.

**Why is this needed**:

Our request parsing plugin is returning a `RawPayload` not a `PayloadMap` because we use multipart HTTP requests (and a `PayloadMap` would be serialized back to JSON). This has the impact that the router director can’t read the request and no longer extracts the model name from it to fill the prometheus metric labels.

zetxqx:
**What type of PR is this?**

/kind feature

The director only read the model name from `PayloadMap` request bodies, so a request whose parser returns a `RawPayload` or `PayloadProto` (for example a multipart HTTP body that must be forwarded unchanged) had an empty `IncomingModelName` — leaving its prometheus metric labels blank.

This moves model-name extraction into the parser:

- Add `InferenceRequestBody.Model`, populated by each parser during `ParseRequest`. The director reads it into `reqCtx.IncomingModelName` for **every** request, including those forwarded unchanged, so metric labels are populated regardless of payload type.
- Add the `ModelNameRewriter` parser interface for writing the resolved target model back into the payload. It takes and returns a `MarshalablePayload`, so only serializable (repackageable) payloads are mutated and the result is guaranteed to be re-forwarded.
- `repackage` now serializes any `Marshaler` payload instead of switching on the concrete payload type.

Fixes #1900

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/epp/requestcontrol/director.go
+++ b/pkg/epp/requestcontrol/director.go
@@ -20,7 +20,6 @@ package requestcontrol
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"math/rand"
@@ -244,6 +243,9 @@ func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestCo
 
 	logger := log.FromContext(ctx)
 
+	// Record the client-facing model for every request, including forwarded-unchanged ones.
+	reqCtx.IncomingModelName = inferenceRequestBody.Model
+
 	err = d.modelRewriteIfNeeded(ctx, reqCtx, inferenceRequestBody)
 	if err != nil {
 		return reqCtx, err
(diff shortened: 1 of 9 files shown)


pkg/epp/requestcontrol/director.go
```diff
@@ -356,45 +358,44 @@ func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestCo
...
 func (d *Director) modelRewriteIfNeeded(ctx context.Context, reqCtx *handlers.RequestContext, inferenceRequestBody *fwkrh.InferenceRequestBody) error {
-	if v, ok := inferenceRequestBody.Payload.(fwkrh.PayloadMap); ok {
-		// Mutate the model name inside the map, this is currently only supported if the payload is a PayloadMap.
-		_, err := d.mutateModel(ctx, reqCtx, v)
-		if err != nil {
-			return err
-		}
+	rewriter, ok := reqCtx.Parser.(fwkrh.ModelNameRewriter)
+	if !ok {
+		return nil
```
ahg-g, line 365: I would log a warning here

ahg-g (commented):
one nit

pkg/epp/requestcontrol/director.go
zetxqx: added

ahg-g (approved):

Issue #1952, opened 2026-07-09
expose KV cache location (device tier) in prefix-cache match data


nilig:
The kvblock index already tracks where each cached block lives: vLLM offloading events carry `Medium`, `llm-d-kv-cache` stores it per pod entry (`kvblock.PodEntry.DeviceTier`), and `Lookup` returns it to the caller. The router drops it — the precise producer flattens the lookup result to a single `matchLen` per pod, and `PrefixCacheMatchInfo` has no location field. Downstream consumers (scorers, source selection) see "N cached tokens" with no way to distinguish GPU-resident blocks from offloaded ones.

Proposal to discuss: expose cache location in the produced data — e.g. per-tier match counts in `PrefixCacheMatchInfo`, or a tier filter on the lookup — so consumers can count or weight tiers per their semantics. The kv-cache library's scorer already supports per-medium weights, so scoring integration has precedent.

Motivating usage — P2P source selection: the P2P source-side lookup resolves block hashes against the offloading manager's tiers (CPU offload region + p2p secondary), not the engine's GPU KV cache. `p2p-source-producer` (#1923), however, selects the source from the location-blind match counts, so it can emit `x-kv-cache-source-host-port` for blocks the source cannot serve. Observed on a 1P/1D cluster (multi-turn, precise producer, always-disagg): on turn 2 the EPP saw 128 cached tokens on the decode pod and fired the header, but the prefiller's pull got `hits=0 misses=1` because decode's turn-1 blocks were still GPU-only; the identical pull hits once the source has offloaded them. `offload_prompt_only: false` narrows the window by offloading during generation, but the views remain structurally divergent: the EPP routes on the GPU-cache view while P2P serves the offload-tier view. With per-tier data, `p2p-source-producer` would compute its delta over p2p-servable tiers only.

Related: #1923 (p2p-source-producer), #1950 (approx/estimate response indexing — the same selection logic, different producer gap).

cc @vMaroon

PR #1962, opened 2026-07-10
feat(epp): expose per-tier cached-block counts in prefix match info

nilig:
**What type of PR is this?**
/kind feature

The precise prefix-cache producer flattens KV-block lookups into a single cached-block count, so consumers of `PrefixCacheMatchInfo` cannot distinguish GPU-resident blocks from offloaded ones. The kvblock index already tracks the device tier per pod entry and returns it from `Lookup`; this PR carries it through: `PrefixCacheMatchInfo` gains a per-device-tier breakdown of the contiguous cached prefix, populated by the precise producer and recorded as found in the index (entries without a confirmed tier, e.g. speculative, count under the empty string). Producers without tier data (approx, burst) leave it nil.

Purely additive: no existing field, accessor, or consumer behavior changes.

A follow-up PR makes `p2p-source-producer` select sources by p2p-servable tiers using this data, and will close the issue.

Part of #1952

**Release note**:
```release-note
PrefixCacheMatchInfo produced by the precise prefix-cache producer now carries per-device-tier cached-block counts, letting consumers distinguish GPU-resident from offloaded prefix blocks.
```

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
@@ -44,9 +44,8 @@ type PrefixCacheMatchInfo struct {
 	// per device tier, the unweighted count of contiguous cached prefix blocks
 	// the endpoint holds in that tier, from the first block until the first
 	// block missing from that tier. A block held in several tiers counts once
-	// per tier. Entries without a confirmed device tier (e.g. speculative
-	// index entries) count under the empty string. Nil when the producer
-	// supplies no tier data.
+	// per tier. Speculative index entries count under the synthetic
+	// "speculative" tier. Nil when the producer supplies no tier data.
 	cachedBlocksByTier map[string]int
 }
 
@@ -85,7 +84,8 @@ func (p *PrefixCacheMatchInfo) CachedBlockCount() int {
 }
 
 // WithCachedBlocksByTier sets the per-device-tier contiguous cached-block
-// counts and returns the receiver for chaining.
+// counts and returns the receiver for chaining. Takes ownership of the map;
+// the caller must not mutate it after the call.
 func (p *PrefixCacheMatchInfo) WithCachedBlocksByTier(cachedBlocksByTier map[string]int) *PrefixCacheMatchInfo {
 	p.cachedBlocksByTier = cachedBlocksByTier
 	return p
--- a/pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
@@ -17,6 +17,8 @@ limitations under the License.
 package prefix
 
 import (
+	"maps"
+
 	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
 	approxprefixconstants "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/constants"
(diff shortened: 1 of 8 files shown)


pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
```diff
@@ -39,6 +41,13 @@ type PrefixCacheMatchInfo struct {
 	// the prefix-based PD decider) get an accurate cached-token figure rather
 	// than a tier-attenuated one. Defaults to matchBlocks when not set.
 	cachedBlockCount int
+	// per device tier, the unweighted count of contiguous cached prefix blocks
+	// the endpoint holds in that tier, from the first block until the first
+	// block missing from that tier. A block held in several tiers counts once
+	// per tier. Entries without a confirmed device tier (e.g. speculative
+	// index entries) count under the empty string. Nil when the producer
+	// supplies no tier data.
+	cachedBlocksByTier map[string]int
```
vMaroon, line 53: nit: I think we should eject unnamed tiers.

nilig: The `""` bucket is exactly the speculative entries: event-sourced entries always carry a tier (the events pool defaults missing ones to `gpu`), and `PreRequest` inserts speculative pods with `Speculative: true` and no tier, since vLLM hasn't reported placement yet. Ejecting them means a just-selected pod shows `CachedBlockCount > 0` with an empty tier map, so a tier-aware consumer would score it as cold during the exact window speculative indexing exists to cover.

Wonder if we could name them instead of ejecting. I'd have `PreRequest` set the tier at insert (e.g. `DeviceTier: "speculative"`), so nothing unnamed ever enters the index and the counting stays a plain group-by. Alternatively `matchedBlockCountByTier` could key on `Speculative` and fold those under a `"speculative"` key in the map, leaving the index untouched. WDYT?

pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
```diff
@@ -75,11 +84,26 @@ func (p *PrefixCacheMatchInfo) CachedBlockCount() int {
 	return p.cachedBlockCount
 }
 
+// WithCachedBlocksByTier sets the per-device-tier contiguous cached-block
+// counts and returns the receiver for chaining.
+func (p *PrefixCacheMatchInfo) WithCachedBlocksByTier(cachedBlocksByTier map[string]int) *PrefixCacheMatchInfo {
+	p.cachedBlocksByTier = cachedBlocksByTier
```
roytman, line 94: nit: stores the map by reference without cloning. Fine today since the only caller builds a fresh map per call, but worth a doc note ("takes ownership of the map") if a future caller ever reuses one.

nilig: Done, added the ownership note to `WithCachedBlocksByTier`. Cloning in the setter would be an allocation for a caller that does not exist, so the doc note it is.

pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
nilig: Went with the counting-side option: `matchedBlockCountByTier` folds `Speculative` entries under a `"speculative"` key, so nothing unnamed surfaces in `CachedBlocksByTier` and the index is untouched.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/utils.go
```diff
@@ -54,3 +54,44 @@ func matchedBlockCount(keys []kvblock.BlockHash, keyToPods map[kvblock.BlockHash
 	}
 	return count
 }
+
+// speculativeTierKey is the CachedBlocksByTier key for speculative index
+// entries: PreRequest inserts them before vLLM has reported placement, so
+// they carry no device tier.
+const speculativeTierKey = "speculative"
```
bongwoobak: This key is part of the `PrefixCacheMatchInfo` contract (its doc comment in data_types.go mentions the "speculative" tier) but the constant is private to this package. 

The follow-up p2p-source-producer will need to filter this key out of `CachedBlocksByTier`, so it would have to hardcode the string. 

How about exporting it from attrprefix like `attrprefix.SpeculativeTierKey` and referencing it here?

nilig: Done - moved it to `attrprefix.SpeculativeTierKey`, declared next to the `CachedBlocksByTier` doc that names it, so the p2p-source producer can reference it instead of hardcoding the string.

vMaroon (approved):

PR #1965, opened 2026-07-10
Adding precise prefix cache indexer observability metrics

### LINKED ISSUE #1964: Precise prefix cache indexer observability metrics

Currently, the precise prefix cache plugin within the EPP scheduler in llm-d-router doesn't expose any observability metrics about the kv cache matches. This is inconsistent with the default approximate prefix cache plugin which exposes high level kv cache hit/blocks metrics. This is a feature request to add a new set of precise prefix cache indexer metrics that would track the precise cache hit ratios.

**Why is this needed**:

The efficiency of intelligent routing and weights applied to precise prefix cache plugin directly depends on the hit ratios statistics of the replicas in the same inference pool. Infact, it is vital to understand the increase and variance in kv cache hit ratios on replicas within the various InferencePools deployed for production workloads.

The following additional metrics are requested:
```
precise_prefix_indexer_max_hit_ratio
precise_prefix_indexer_avg_hit_ratio
precise_prefix_indexer_std_dev_hit_ratio
```

SachinVarghese:
**What type of PR is this?**

/kind feature

Currently, the precise prefix cache plugin within the EPP scheduler in llm-d-router doesn't expose any observability metrics about the kv cache matches. This is inconsistent with the default approximate prefix cache plugin which exposes high level kv cache hit/blocks metrics. This pull request adds a new set of precise prefix cache indexer metrics that would track the precise cache hit ratios.

Fixes #1964 

```release-note
Adding precise prefix cache indexer observability metrics
```


pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/metrics.go
```diff
@@ -0,0 +1,85 @@
...
+
+	metricsutil "github.com/llm-d/llm-d-router/pkg/common/observability/metrics"
+	eppmetrics "github.com/llm-d/llm-d-router/pkg/epp/metrics"
+)
+
+var (
+	llmdPrefixCacheMaxHitRatio = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: eppmetrics.LLMDRouterEndpointPickerSubsystem,
+			Name:      "precise_prefix_indexer_max_hit_ratio",
```
ahg-g: we aligned in the past on not having the plugin name as part of the metric name. @vMaroon @liu-cong fyi

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -304,6 +313,9 @@ func (p *Producer) produceFromBlockKeys(ctx context.Context, span trace.Span,
 			attrprefix.NewPrefixCacheMatchInfo(matchLen, totalBlocks, p.blockSizeTokens).WithCachedBlockCount(cachedBlocks))
 	}
 
+	max, avg, std := calculateHitRatioStats(endpointHitRatios)
+	recordPrefixCacheHitRatio(p.typedName.Name, p.typedName.Type, max, avg, std)
```
ahg-g: In the approximate case, we record the metric for the picked endpoint, I think we should do the same here, the metric gets recorded in PreRequest. 

Reporting the max across all endpoints is useful since we may not necessarily pick that endpoint, but mean and stddev seems less useful, but I am not totally against them. Should we add them also to the approximate producer? @liu-cong @kaushikmitr wdyt?

SachinVarghese: In theory, with intelligent precise prefix cache based routing, the cache hit ratios should improve across replicas and if not, the weighting of this plugin on the scheduling profile should be adjusted accordingly - this has been our experience. So, it is important to monitor the average hit ratios across replicas in an prefix cache indexer. Similarly, the variance in hit ratios help in modelling the difference in kv cache accumulated in each replica. Continous monitoring of these signals can help decide the direction of weight changes for the precise prefix cache plugin in the scheduling profile. 

The reason this metric recording was not implemented in the prerequest method was due to the need for index lookup across an endpoint set to calculate these stats mentioned above. Another index lookup could get intensive at scale and it was simpler to retrive the metrics from an existing lookup computation in produce method.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/metrics.go
SachinVarghese: kindly suggest the correct naming convention for these set of metrics, will update as needed

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
ahg-g: > In theory, with intelligent precise prefix cache based routing, the cache hit ratios should improve across replicas and if not, the weighting of this plugin on the scheduling profile should be adjusted accordingly - this has been our experience. So, it is important to monitor the average hit ratios across replicas in an prefix cache indexer. Similarly, the variance in hit ratios help in modelling the difference in kv cache accumulated in each replica. Continous monitoring of these signals can help decide the direction of weight changes for the precise prefix cache plugin in the scheduling profile.

Sounds good, can we add similar summary metrics to the approximate producer as well pls?

> The reason this metric recording was not implemented in the prerequest method was due to the need for index lookup across an endpoint set to calculate these stats mentioned above. Another index lookup could get intensive at scale and it was simpler to retrive the metrics from an existing lookup computation in produce method. 

Tracking the hit ratio of the picked endpoint is useful, it is a strong indication of how good the actual pick was. The summary stat metric can stay here, but the hit ratio of the picked endpoint can be recorded in PreRequest.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/metrics.go
ahg-g: This and the approximate producer should share these metrics, the plugin name and type label will differentiate the source, so they would be:

- `prefix_indexer_hit_ratio` for the hit ratio of the picked endpoint
- `prefix_indexer_max_hit_ratio` for the endpoint with the max hit ratio
- `prefix_indexer_avg_hit_ratio` for the avg
- `prefix_indexer_std_dev_hit_ratio` for the std dev

As suggested in the other other comment, can we please record all these metrics for both the approximate and the precise.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/metrics.go
```diff
@@ -0,0 +1,85 @@
...
+		return errors.New("precise prefix cache metrics registerer is required")
+	}
+	for _, collector := range []prometheus.Collector{
+		llmdPrefixCacheMaxHitRatio,
+		llmdPrefixCacheAvgHitRatio,
+		llmdPrefixCacheStdDevHitRatio,
+	} {
+		if err := registerer.Register(collector); err != nil {
+			var alreadyRegistered prometheus.AlreadyRegisteredError
+			if errors.As(err, &alreadyRegistered) && alreadyRegistered.ExistingCollector == collector {
```
ahg-g: What does `alreadyRegistered.ExistingCollector == collector` evaluate? will this check pass if a similar metric was defined and registered by some other plugin? I am asking since I am not sure if this is a deep equal check or just a check if the existingCollector is the exact same instance as the collector.

SachinVarghese: Good catch. I looked into this and based on current impl here, if a collector is already registerd by another plugin, this condition will fail (even with the same name, subsystem combination -> it is a deep equal check)  and this `registerMetrics` function will error.  

Now given that, it may be ideal to list the various metric collectors across all plugins into a separate pkg labelled by plugin name and type but that may be out of scope for this PR. Instead here I can make the precise prefix cache plugin utlize the existing collectors and recording functions from the approximate prefix plugin which solves the immediate inconsistency issue the PR aimed to solve.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/metrics.go
SachinVarghese: Updated in latest change

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
SachinVarghese: Updated in latest change

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/metrics.go
SachinVarghese: Updated in latest change

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/metrics.go
```diff
@@ -111,14 +144,14 @@ func registerMetrics(registerer prometheus.Registerer) error {
...
-func recordPrefixCacheSize(pluginName, pluginType string, size int64) {
+func RecordPrefixCacheSize(pluginName, pluginType string, size int64) {
 	prefixCacheSize.WithLabelValues().Set(float64(size))
 	llmdPrefixCacheSize.WithLabelValues(pluginName, pluginType).Set(float64(size))
 }
 
 // recordPrefixCacheMatch records both the hit ratio and hit length for a prefix indexer match.
 // matchedLength is the number of characters that matched, and totalLength is the total prefix length.
-func recordPrefixCacheMatch(pluginName, pluginType string, matchedLength, totalLength int) {
+func RecordPrefixCacheMatch(pluginName, pluginType string, matchedLength, totalLength int) {
```
bongwoobak, line 154: `prefix_indexer_hit_bytes` is declared as bytes with byte-scale buckets (metrics.go:102-104) and the approximate producer feeds it bytes (plugin.go:305). The precise producer passes block counts (preciseprefixcache/prerequest.go:180), so its observations land in the 0-16 bucket nearly every time. The deprecated `inference_extension_` copy takes no labels at all (metrics.go:106), so both producers share one series with no way to tell them apart.

Scaling by `blockSize*averageCharactersPerToken` on the precise side would fix it, or drop the shared length metric and keep the ratios. The ratios are fine either way since numerator and denominator scale together.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
```diff
@@ -158,17 +165,19 @@ func (p *Producer) PreRequest(ctx context.Context,
...
+		if err := index.Add(ctx, nil, promptKeys, allPodEntries); err != nil {
 			logger.Error(err, "Failed to add speculative entries to index",
 				"pod", speculativePod.PodIdentifier)
 		}
+		totalBlocks += len(promptKeys)
 	}
-
-	allPodEntries := []kvblock.PodEntry{speculativePod}
+	matchLen := state.prefixCacheMatchLengths[speculativePod.PodIdentifier]
+	approximateprefix.RecordPrefixCacheMatch(p.typedName.Name, p.typedName.Type, int(matchLen), totalBlocks)
```
bongwoobak, line 180: This metric is about what the engine actually has cached. That doesn't change depending on whether we're also seeding predictions, so it shouldn't be gated on `speculativeIndexing`. Right now it only records when the flag is on, so it vanishes in the default config. The max/avg/stddev stats you added in Produce record on every request. Same feature, two different conditions.
Pulling this call out from under the flag, up to where the target endpoint gets resolved, lines them up.

Separately, I'd keep confirmed and speculative measurements apart rather than blend them.

A speculative entry is a guess that a pod will hold a block, and the count feeding this metric doesn't distinguish it from a block the engine confirmed, so once speculation is on the hit ratio climbs partly on our own guesses. The flag is already on the entries `Lookup` returns and `matchedBlockCountByTier` reads it right below, so skipping speculative entries in the count stays local to this file.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -363,13 +371,15 @@ func (p *Producer) produceFromBlockKeys(ctx context.Context, span trace.Span,
...
 	maxMatch := 0
+	endpointHitRatios := make([]float64, 0, len(endpoints))
 	for _, ep := range endpoints {
 		md := ep.GetMetadata()
 		if md == nil {
 			continue
 		}
 		addr := fmt.Sprintf("%s:%s", md.Address, md.Port)
 		matchLen := int(aggregatedScores[addr])
+		endpointHitRatios = append(endpointHitRatios, float64(matchLen)/float64(totalBlocks))
```
bongwoobak, line 382: `aggregatedScores` is the scorer's weighted sum, not a block count. Blocks get weighted by device tier and the default config is gpu 1.0 / cpu 0.8, so a block that lives in CPU and not GPU contributes 0.8 and this ratio lands below the block ratio the help text promises. The approximate producer puts a real count into the same metric name, so `plugin_type` ends up splitting two different units instead of two implementations of the same measurement.

`cachedBlocks` just below is the literal match count, and it stops at the same block the scorer's chain does, so on a GPU-only deployment the two are identical and nothing changes. Could you move this append under that loop and use it?

```go
endpointHitRatios = append(endpointHitRatios, float64(cachedBlocks)/float64(totalBlocks))
```

`prefixCacheMatchLengths` carries the same value into `PreRequest`, so that one needs the block count too.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
gyliu513, line 180: `RecordPrefixCacheMatch` expects character/byte-scale lengths (see its doc comment and the only other caller in `approximateprefix/plugin.go:305`, which converts via `matchLen*blockSize*averageCharactersPerToken` before calling). Here `matchLen`/`totalBlocks` are passed straight through as raw KV-block counts, so `prefix_indexer_hit_bytes` (bucketed 16..65536) ends up recording block-count-scale values instead of byte-scale values for this plugin. Can we apply the same block→byte conversion before calling `RecordPrefixCacheMatch`?

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/metrics.go
```diff
@@ -87,11 +117,14 @@ var (
 	)
 )
 
-func registerMetrics(registerer prometheus.Registerer) error {
+func RegisterMetrics(registerer prometheus.Registerer) error {
```
gyliu513, line 120: `registerMetrics`/`recordPrefixCacheSize`/`recordPrefixCacheMatch` were exported here solely so `preciseprefixcache` could call into them cross-package. Per the repo's PR-scope convention (no new abstractions/pattern changes beyond the issue), could `preciseprefixcache` get its own metrics registration instead of reaching into `approximateprefix`'s internals? Coupling the "precise" plugin's metrics through the "approximate" plugin's package makes both harder to reason about independently.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
```diff
@@ -50,18 +51,24 @@ type speculativeEntries struct {
 // via PluginState, avoiding a second hash on the same request.
 // perPromptKeys holds one slice of block keys per prompt; single-prompt
 // requests use a length-1 outer slice.
+// prefixCacheMatches is a map of pod identifier to its longest prefix cache match length.
```
gyliu513, line 54: This struct doc comment says "prefixCacheMatches is a map of pod identifier to its longest prefix cache match length", but the field is named `prefixCacheMatchLengths`, and its values come from `producer.go`'s `aggregatedScores`, a scorer-weighted sum accumulated across every prompt in a multi-prompt request, not a single "longest match length". Could the comment be updated to describe what the map actually holds?

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/utils.go
```diff
@@ -0,0 +1,54 @@
...
+	if len(hitRatios) > 1 {
+		stdDevRatio = varianceSum / float64(len(hitRatios)-1)
+		stdDevRatio = math.Sqrt(stdDevRatio)
+	}
+
+	// Round to two decimal places for consistency in metrics reporting
+	avgRatio = math.Round(avgRatio*100) / 100
+	stdDevRatio = math.Round(stdDevRatio*100) / 100
+
+	return maxRatio, avgRatio, stdDevRatio
```
gyliu513, line 53: Does `maxRatio` need rounding too?

pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/utils.go
```diff
@@ -0,0 +1,54 @@
...
+limitations under the License.
+*/
+
+package approximateprefix
+
+import (
+	"math"
+)
+
+func CalculateHitRatioStats(hitRatios []float64) (maxRatio float64, avgRatio float64, stdDevRatio float64) {
```
gyliu513, line 23: `CalculateHitRatioStats` (this function) re-implements mean/sample-variance/round-to-2-decimals inline, but `pkg/epp/datalayer/logger/logger.go:143,186` already has this pattern (`calculateSummary` + `round2`, used for KV-cache/queue/running-request stats). Worth extracting a shared helper instead of duplicating the math here?

Issue #1877, opened 2026-07-02
[Coordinator] Integrate Coordinator e2e tests, deployments  and its Makefile into the llm-d-router infrastructure


roytman:
After migrating llm-d-incubation/coordinator into llm-d-router, we have to integrate its e2e tests and deployments into llm-d-router CI tests

roytman:
Done by #1975

PR #1975, opened 2026-07-12
[coordinator] e2e tests: containerize and share deploy components

revit13:
**What type of PR is this?**
/kind test

Reworks how the coordinator's e2e tests run and removes duplication between the coordinator's deploy manifests and the main router's, fixing several bugs found along the way.

Containerize test-e2e-coordinator: the target now runs the whole suite (including kind create cluster) inside the builder container via docker run, matching the main router's test-e2e-run pattern, instead of executing directly on the host.
Share deploy components with the router: crds-gateway-api, crds-gie, vllm-encode, and vllm-prefill now reference the router's shared deploy/components/* instead of coordinator-local vendored/duplicated copies. vllm-decode intentionally stays coordinator-local (the router's decode carries a routing-sidecar the coordinator's pipeline doesn't use) with a comment explaining why.
Drop the unused Gateway API CRD install from the coordinator e2e suite — the coordinator never creates a Gateway/HTTPRoute object (it routes on an EPP-Phase header via a hand-rolled Envoy config), so only GIE CRDs are installed now.
Parameterize vllm-render: port and image are now ${VLLM_RENDER_PORT}/${VLLM_RENDER_IMAGE} (previously hardcoded 8082 and reusing ${VLLM_IMAGE}), and dropped unused PORT/POD_NAME/POD_NAMESPACE env vars from the container.
Bug fixes:
--force-dummy-tokenizer was being passed twice to the encode/prefill/decode simulator containers (baked into the base manifest and re-added by the simulator overlay patch).
Relative paths in the coordinator's Go e2e test package were off by one directory level, so kustomize/manifest lookups resolved to the wrong location.
image-build-epp called a script that doesn't exist (scripts/build-epp-image.sh); it now falls through to the existing generic image-build-% pattern.
Docs: README.coord.md gains a Table of Contents, documents test-e2e-coordinator, E2E_KEEP_CLUSTER_ON_FAILURE, accessing the kind cluster's kubeconfig after a preserved failure, and the full e2e environment-variable table.

Depends on https://github.com/llm-d/llm-d-router/pull/1601 — once that merges, further cleanup can be applied here (e.g. sharing the vllm-render component with the router too, the same way vllm-encode/vllm-prefill already are)

#1877

```release-note
NONE
```


deploy/components/vllm-encode/kustomization.yaml
```diff
@@ -3,6 +3,7 @@
 #
 # Encoder deployment for multimodal disaggregation scenarios (E/PD, E/P/D).
 # Includes a PVC for shared encoder embeddings cache.
```
roytman: can you remove this line? And rephrase the comments (we need PVC only for shared storage EC connector)

deploy/coordinator/environments/dev/epd-pools/kustomization.yaml
```diff
@@ -2,8 +2,11 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-- ../../../components/vllm-encode/
-- ../../../components/vllm-prefill/
+- ../../../../components/vllm-encode/
+- ../../../../components/vllm-prefill/
+# vllm-decode is coordinator-local, not the shared deploy/components/vllm-decode: the
+# router's decode carries a routing-sidecar proxying 8000->8200 for connector-based P/D
```
roytman: Why can't we reuse the shared deploy/components/vllm-decode? it is similar as the `epd` scenario without the sidecar

Makefile.coord.mk
```diff
@@ -1,23 +1,35 @@
 SHELL := /usr/bin/env bash
 
-# Image registry + dev-environment image tags (single source of truth).
```
roytman, line 3: should we return versions.mk?

roytman:
CI tests do not include the coordinator

roytman:
We need to add rank-port topology (for DP)

Makefile.coord.mk
revit13: Checked, it's not actually shared here, the original pull_images.sh read versions.mk, but this repo's version (reused from the router) doesn't. So it'd only have one consumer, same as keeping it inline. Can restore it if you'd rather match upstream's layout for easier syncing.

revit13:
> We need to add rank-port topology (for DP)

I will address it in a separate PR if thats OK.

revit13:
> CI tests do not include the coordinator

I added a new workflow for the coordinator.

test/utils/parallel.go
```diff
@@ -0,0 +1,98 @@
...
+// processPortOffset is the gap between consecutive parallel processes' ports,
+// shared by ProcessPort (NodePorts) and BuildExtraPortMappings (Kind's
+// containerPort/hostPort mappings) so a process's actual exposed port always
+// matches the port it computes and dials.
+const processPortOffset = 100
+
+// ProcessPort returns the NodePort for the current Ginkgo parallel process for
+// a service whose first process uses basePort. When tests run in parallel,
+// each process gets its own NodePort of the form basePort + processPortOffset
+// * (the process number minus one), so process one gets exactly basePort.
```
roytman, line 36: The file has zero unit tests

test/coordinator/e2e/coordinator/e2e_suite_test.go
```diff
@@ -87,17 +84,19 @@ var (
 	containerRuntime = env.GetEnvString("CONTAINER_RUNTIME", "docker", ginkgo.GinkgoLogr)
 	eppImage         = env.GetEnvString("EPP_IMAGE", "ghcr.io/llm-d/llm-d-router-endpoint-picker:dev", ginkgo.GinkgoLogr)
 	vllmSimImage     = env.GetEnvString("VLLM_IMAGE", "ghcr.io/llm-d/llm-d-inference-sim:v0.10.0", ginkgo.GinkgoLogr)
+	vllmRenderImage  = env.GetEnvString("VLLM_RENDER_IMAGE", vllmSimImage, ginkgo.GinkgoLogr)
+	vllmRenderPort   = env.GetEnvString("VLLM_RENDER_PORT", "8082", ginkgo.GinkgoLogr)
 	coordinatorImage = env.GetEnvString("COORDINATOR_IMAGE", "", ginkgo.GinkgoLogr)
 	modelName        = env.GetEnvString("MODEL_NAME", "Qwen/Qwen3-VL-2B-Instruct", ginkgo.GinkgoLogr)
 
-	nsName     = env.GetEnvString("NAMESPACE", "default", ginkgo.GinkgoLogr)
+	numProcesses = env.GetEnvInt("E2E_NUM_PROCS", 1, ginkgo.GinkgoLogr)
```
roytman, line 92: `E2E_NUM_PROCS` is never set by any workflow, and neither run script passes -ginkgo.procs/-p to go test. So numProcesses is always 1 in every CI run — the baseName-N / port+100*idx branches have never run, before or after this PR. This PR generalizes that already-dormant router code into a shared package and hands the coordinator suite the identical, still-unexercised capability. Pre-existing pattern, not introduced by this PR, but worth flagging since the PR explicitly reframes it as new "parallel test infrastructure" — infrastructure that's currently only proven at n=1.

Probably it is a part of #1976 

```
grep -r E2E_NUM_PROCS *
test/coordinator/e2e/coordinator/e2e_suite_test.go:92:	numProcesses = env.GetEnvInt("E2E_NUM_PROCS", 1, ginkgo.GinkgoLogr)
test/utils/parallel.go:71:// E2E_NUM_PROCS environment variable) matches the actual number of Ginkgo
test/utils/parallel.go:76:		ginkgo.Fail(fmt.Sprintf("The value of the environment variable `E2E_NUM_PROCS` (%d) is not equal to the number of ginkgo processes being run (%d)",
test/e2e/e2e_suite_test.go:87:	numProcesses     = env.GetEnvInt("E2E_NUM_PROCS", 1, ginkgo.GinkgoLogr)
```

revit13: Right, as far as I know @shmuelk is planning to have a follow up PR.

test/utils/parallel.go
revit13: No other file in `test/utils/` has a unit test file. The unit test target (make test-unit) explicitly excludes the test/ subtree, so any *_test.go added here would not be picked up by it.
  The functions that call ginkgo.GinkgoParallelProcess() (ProcessPort, Namespace) also can't be meaningfully tested outside a running Ginkgo suite. The pure helpers could have tests, but  they'd need a separate make target to run.

roytman:
- patch-decode.yaml uses $patch: replace on the vllm container's ports list — correct today, but if the router ever adds a port to the shared vllm-decode base, this silently drops it with no test failure. A one-line comment noting the list is fully replaced (not merged) would save someone a confusing debug session later.

- Neither ci-coordinator.yaml's nor ci-pr-checks.yaml's path filters include deploy/**, so a PR touching only a shared deploy component won't trigger either e2e suite. Pre-existing gap, but this PR is the point where a second consumer starts depending on that tree.

- LocalhostURL in `parallel.go` is missing a doc comment that its sibling exported functions all have.
The new `ci-coordinator.yaml` builds builder+coordinator+epp images serially in one job, unlike the router's matrixed/cached image-build jobs in ci-pr-checks.yaml. Probably fine as a simpler first pass, flagging for awareness only.
This can be a part of a separate PR

- `crds-gie` and `vllm-decode` shared kustomizations lack the "also used by the coordinator" heads-up comment that vllm-prefill got. A router contributor touching either has no signal that the coordinator now depends on them.

roytman (dismissed):
/lgtm

test/coordinator/e2e/coordinator/utils_test.go
```diff
@@ -80,3 +83,60 @@ func podsInDeploymentsReady(objects []string) {
 		}
 	}
 }
+
+// dumpPodsAndLogs prints pod statuses and container logs for the given namespace
+// to the Ginkgo writer. Call this before cleanup to ensure the information is
+// available when CI tests fail.
+func dumpPodsAndLogs(nsName string) {
```
elevran: This introducs a duplication of `dumpPodsAndLogs` and the helper below from `test/e2e/utils_test.go:487`, which he PR otherwise works to eliminate.
Moving both helpers into `test/utils` (alongside `parallel.go`) keeps the dedupe theme consistent across the diff. Note that the router's copy has slightly richer signatures (`previous bool`, explicit `context`) but the minimum useful shared shape is `func DumpPodsAndLogs(nsName string)` matching this version.

test/coordinator/e2e/coordinator/utils_test.go
```diff
@@ -80,3 +83,60 @@ func podsInDeploymentsReady(objects []string) {
...
+	tailLines := int64(200)
+	limitBytes := int64(1 << 20) // 1MiB
+	req := testConfig.KubeCli.CoreV1().Pods(nsName).GetLogs(podName, &corev1.PodLogOptions{
+		Container:  containerName,
+		TailLines:  &tailLines,
+		LimitBytes: &limitBytes,
+	})
+	stream, err := req.Stream(testConfig.Context)
+	if err != nil {
+		ginkgo.GinkgoWriter.Printf("(failed to fetch logs: %v)\n", err)
```
elevran: nit: 
when `GetLogs`/`Stream` fails (often because cleanup races and the pod is gone), the dump prints `(failed to fetch logs: %v)` without the pod/container context. Wrap with the pod/container names so an operator scanning a failure dump can tell which container failed. Same suggestion applies to the existing router copy in `test/e2e/utils_test.go:586`.

elevran (commented):
Containerization and component-sharing are well-scoped - thanks.
See inline comments on the new copy of `dumpPodsAndLogs`/`dumpContainerLogs`.

elevran:
@roytman @revit13 see #2048 which settles on `main` instead of `dev` tags. Left note there.

test/coordinator/e2e/coordinator/setup_test.go
```diff
@@ -183,30 +183,17 @@ func createCoordinator(config string) []string {
 	objects = append(objects, testutils.CreateObjsFromYaml(testConfig, docs, nsName)...)
 
 	podsInDeploymentsReady(objects)
-	if k8sContext != "" {
```
capri-xiyue, line 186: why do we remove this?

test/utils/dump.go
```diff
@@ -0,0 +1,171 @@
...
+	// dumpTailLines is the number of trailing log lines fetched per container.
+	dumpTailLines = int64(200)
+	// dumpLimitBytes caps the log bytes fetched per container.
+	dumpLimitBytes = int64(1 << 20) // 1MiB
+)
+
+// DumpPodsAndLogs prints pod statuses and container logs for the given namespace
+// to the Ginkgo writer. Call this before cleanup to ensure the information is
+// available when CI tests fail.
+func DumpPodsAndLogs(cfg *TestConfig, nsName string) {
```
capri-xiyue, line 43: do you need UT for this?

README.coord.md
```diff
@@ -2,10 +2,35 @@
...
+  - [Development](#development)
+    - [Running Tests](#running-tests)
+      - [Unit Tests](#unit-tests)
+      - [End-to-End Tests](#end-to-end-tests)
+
 ## Architecture
 
 ```
-Client -> Coordinator -> Inference Gateway -> EPP -> vLLM Workers
+Client -> Inference Gateway -> Coordinator -> Inference Gateway -> EPP -> vLLM Workers
```
capri-xiyue: I think the diagram would be clearer if we used a single double-ended arrow between the Inference Gateway and the Coordinator, rather than writing it out as Inference Gateway -> Coordinator -> Inference Gateway which is quite confusing.

test/coordinator/e2e/coordinator/setup_test.go
revit13: This was the direct-NodePort path (the port-forward plus a second /readyz poll against coordinatorBaseURL()). Nothing exercises it anymore, all traffic goes through the Envoy gateway, so it was dead code. Readiness is still covered: podsInDeploymentsReady confirms the pod is up, and the gateway /readyz poll still guards against Envoy's STRICT_DNS resolution lagging the per-test Service. Removed coordinatorBaseURL/getCoordinatorPort too since they had no other callers.

test/utils/dump.go
revit13: Not sure about that. This is best-effort diagnostic output that runs on E2E failure, it has no return value, swallows all errors, and only affects how readable CI logs are, never a test result. It's exercised by the E2E suites themselves. Nothing else in test/utils (context, network, parallel, server, ...) is unit-tested either, so adding a UT just here would be inconsistent. A UT would mostly end up asserting on formatted column output, which is brittle and low-value. Happy to add one if you'd still like coverage, e.g. a fake.NewSimpleClientset case, but I'd lean against it.

elevran (approved):

.github/workflows/ci-coordinator.yaml
```diff
@@ -0,0 +1,106 @@
+name: CI - Coordinator Test
```
elevran, line 1: should CI for coordinator share the existing CICD files instead of dedicated actions?
Just wonder how much is duplicated and whether the can be combined instead of introducing new actions and merge gates.

.github/workflows/ci-pr-checks.yaml
```diff
@@ -45,6 +45,11 @@ jobs:
               - .github/actions/docker-build-and-push/action.yml
               - .github/actions/e2e-runner-setup/action.yml
               - .github/workflows/ci-pr-checks.yaml
+              - '!Dockerfile.coordinator'
```
elevran, line 48: are these consistent with the previous items? None contain `!` prefixes.

Makefile.coord.mk
```diff
@@ -1,23 +1,35 @@
 SHELL := /usr/bin/env bash
```
elevran, line 1: can you please confirm the need for maintaining a coordinator makefile vs inlining targets into existing makefiles?

elevran (commented):
@revit13 please open a follow up issue and PR to determine if and what shared build and CICD code can be leveraged. I have concerns about the additional variables and targets that might be duplicated across GH actions and Makefiles.

Issue #1931, opened 2026-07-08
test: Exploit Ginkgo's ability to run tests in parallel in the End to End tests


shmuelk:
**What would you like to be added**:
Today's processors, including user laptops and desktops, are all multi-processors and are capable of running many processes in parallel. Ginkgo, the test framework used in the llm-d-router End to End tests supports running  tests in parallel. These tests run in separate processes. This capability should be exploited by the llm-d-router End to End tests to shorten the time it takes to run the tests.

**Why is this needed**:
The llm-d-router End to End tests run for about fifteen minutes on a high end Mac (M4 Max) and probably slower on slower machines. This affects developer productivity, as the End to End tests should be run before creating and updating PRs.

In addition the split done in the CI pipeline is in-efficient as a Kind cluster is spun up for each test "fragment".

Instead the tests should be run using Ginkgo's CLI's ability to run tests in parallel in separate processes. The tests should be broken up into groups (perhaps via Ginkgo.When) that are run in parallel. Each such group is run using it's own K8S namespace (InferencePool, EPP, Envoy, Model Servers, etc), providing isolation between the tests.

PR #1976, opened 2026-07-12
test: Parallelize the End to End tests - Part one infrastructure

shmuelk:
**What type of PR is this?**
/kind test

Currently the llm-d-router End to End tests run for quite a long time (about sixteen minutes on my Mac M4 Max).

This long time has affected the CI pipeline in the past. A work around was done to run the tests in parts in parallel in separate Kind clusters. This splitting up of the tests needs to be managed manually and has overhead as prepetory steps are done over and over again for each test "slice'

A better solution is to run tests in parallel on the same Kind cluster using K8S Namespaces for separation as needed.

This PR modifies the End to End test infrastructure to enable the running of different tests in different namespaces. A follow on PR will exploit this ability and Gonkgo's ability to run tests in parallel.

Refs #1931

```release-note
NONE
```

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -79,8 +79,9 @@ var (
 	sideCarImage     = env.GetEnvString("SIDECAR_IMAGE", "ghcr.io/llm-d/llm-d-router-disagg-sidecar:dev", ginkgo.GinkgoLogr)
 	vllmRenderImage  = env.GetEnvString("VLLM_RENDER_IMAGE", "vllm/vllm-openai-cpu:v0.21.0", ginkgo.GinkgoLogr)
 	loadRenderImage  = env.GetEnvBool("LOAD_VLLM_RENDER_IMAGE", true, ginkgo.GinkgoLogr)
-	// nsName is the namespace in which the K8S objects will be created
-	nsName = env.GetEnvString("NAMESPACE", "default", ginkgo.GinkgoLogr)
+	numProcesses     = env.GetEnvInt("E2E_NUM_PROCS", 1, ginkgo.GinkgoLogr)
+	// baseNsName is the base of the namespace in which the K8S objects will be created
+	baseNsName = env.GetEnvString("NAMESPACE", getDefaultNsName(), ginkgo.GinkgoLogr)
 
 	// k8sContext is the Kubernetes context to work with
 	k8sContext = env.GetEnvString("K8S_CONTEXT", "", ginkgo.GinkgoLogr)
@@ -115,15 +116,16 @@ var _ = ginkgo.BeforeSuite(func() {
 	setupK8sClient()
 	setupNameSpace()
 	createCRDs()
-	createEnvoy()
+	nsName := getNamespace()
+	createEnvoy(nsName)
 	infraSubs := map[string]string{
 		"${EPP_NAME}": "e2e-epp",
 	}
 	rbacYamls := substituteMany(testutils.ReadYaml(rbacManifest), infraSubs)
-	rbacObjects = testutils.CreateObjsFromYaml(testConfig, rbacYamls)
+	rbacObjects = testutils.CreateObjsFromYaml(testConfig, rbacYamls, nsName)
 	saYamls := substituteMany(testutils.ReadYaml(serviceAccountManifest), infraSubs)
-	serviceAccountObjects = testutils.CreateObjsFromYaml(testConfig, saYamls)
-	serviceObjects = testutils.ApplyYAMLFile(testConfig, servicesManifest)
+	serviceAccountObjects = testutils.CreateObjsFromYaml(testConfig, saYamls, nsName)
+	serviceObjects = testutils.ApplyYAMLFile(testConfig, servicesManifest, nsName)
 
 	// Prevent failure in tests due to InferencePool not existing before the test
 	infPoolObjects = createInferencePool(1, false)
@@ -148,7 +150,9 @@ var _ = ginkgo.AfterSuite(func() {
 // ReportAfterEach only fires for individual specs and would miss setup/teardown failures.
 var _ = ginkgo.ReportAfterSuite("cleanup", func(report ginkgo.Report) {
 	if !report.SuiteSucceeded {
-		dumpPodsAndLogs()
+		for idx := range numProcesses {
+			dumpPodsAndLogs(fmt.Sprintf("%s-%d", baseNsName, idx+1))
+		}
 	}
 
 	shouldKeep := keepClusterOnFailure && !report.SuiteSucceeded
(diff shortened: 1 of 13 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -148,7 +151,9 @@ var _ = ginkgo.AfterSuite(func() {
 // ReportAfterEach only fires for individual specs and would miss setup/teardown failures.
 var _ = ginkgo.ReportAfterSuite("cleanup", func(report ginkgo.Report) {
 	if !report.SuiteSucceeded {
-		dumpPodsAndLogs()
+		for idx := range numProcesses {
+			dumpPodsAndLogs(fmt.Sprintf("%s-%d", baseNsName, idx+1))
+		}
```
roytman: The dump loop always suffixes the namespace (baseNsName-<n>), but getNamespace() returns baseNsName unsuffixed when numProcesses == 1. In the default single-process config, objects live in namespace "default" while this dumps "default-1", so on a suite failure no pod statuses or logs are captured, which is exactly when they are needed. Mirror the getNamespace() semantics:


```suggestion
		for idx := range numProcesses {
			ns := baseNsName
			if numProcesses > 1 {
				ns = fmt.Sprintf("%s-%d", baseNsName, idx+1)
			}
			dumpPodsAndLogs(ns)
		}
```

test/e2e/e2e_suite_test.go
```diff
@@ -190,6 +196,21 @@ var _ = ginkgo.ReportAfterSuite("cleanup", func(report ginkgo.Report) {
 
 // Create the Kubernetes cluster for the E2E tests and load the local images
 func setupK8sCluster() {
+	// extraPortMappings is substituted into `extraPortMappings: ${EXTRA_PORT_MAPPINGS}` in the Kind
+	// cluster configuration. Each item must use 2-space indentation to match that field's level in the
+	// YAML. If the field is ever reindented in inference-pools.yaml, update the format string here too.
```
roytman: nit: this comment references `inference-pools.yaml`, but the substitution target here is the inline `kindClusterConfig` template in this file. Reads like a copy of the targetPorts comment below. Point it at `kindClusterConfig` instead.

test/e2e/e2e_suite_test.go
```diff
@@ -190,6 +196,21 @@ var _ = ginkgo.ReportAfterSuite("cleanup", func(report ginkgo.Report) {
...
+	var extraPortMappingsBuilder strings.Builder
+	for idx := range numProcesses {
+		inc := idx * 100
+		fmt.Fprintf(&extraPortMappingsBuilder, "\n  - containerPort: %d", 30080+inc)
+		fmt.Fprintf(&extraPortMappingsBuilder, "\n    hostPort: %d", basePort+inc)
+		fmt.Fprintf(&extraPortMappingsBuilder, "\n    protocol: TCP")
```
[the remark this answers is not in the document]
shmuelk: This is setup for a future PR in which tests will actually be run in parallel. In that PR that service definition will be updated on a test by test basis to map to the correct nodePort.

test/e2e/e2e_suite_test.go
```diff
@@ -108,22 +110,29 @@ func TestEndToEnd(t *testing.T) {
 }
 
 var _ = ginkgo.BeforeSuite(func() {
```
elevran, line 112: A plain `BeforeSuite` means every parallel process will independently run `kind create cluster` against the same cluster name once `-p` is enabled - not concurrency-safe. Worth switching to `SynchronizedBeforeSuite` so cluster creation runs once and is broadcast.

test/e2e/e2e_suite_test.go
```diff
@@ -148,7 +157,13 @@ var _ = ginkgo.AfterSuite(func() {
 // ReportAfterEach only fires for individual specs and would miss setup/teardown failures.
 var _ = ginkgo.ReportAfterSuite("cleanup", func(report ginkgo.Report) {
```
elevran, line 158: The `k8sContext != ""` cleanup branch further down only tears down the reporting process's own namespace (`getNamespace()`), not all `numProcesses` namespaces - resources from other processes under an externally supplied `K8S_CONTEXT` would never be cleaned up once parallelism is enabled.

test/e2e/e2e_suite_test.go
```diff
@@ -346,38 +375,51 @@ func createInferencePool(numTargetPorts int, toDelete bool) []string {
...
-		"--context="+k8sContext, "--namespace="+nsName)
+	command := exec.Command("kubectl", "port-forward", "pod/"+eppPodName, strconv.Itoa(getMetricsPort())+":9090",
+		"--context="+k8sContext, "--namespace="+getNamespace())
 	eppPortForwardSession, err = gexec.Start(command, ginkgo.GinkgoWriter, ginkgo.GinkgoWriter)
 	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
 	// Give it a moment to establish
 	time.Sleep(3 * time.Second)
 }
 
+func getPort() int {
```
elevran, line 403: nit: `getPort`, `getMetricsPort`, `getNamespace`, `getDefaultNsName` are the crux of this PR's scheme but have no comments explaining the stride (100) or why ports use a `-1` offset while namespace doesn't. Worth a short comment given this is the mechanism the whole PR introduces.

test/e2e/utils_test.go
```diff
@@ -424,7 +424,7 @@ func extractFinishReasonFromStreaming(sseData string) string {
 }
 
 // getPodRequestCount gets the total vLLM request count from a pod's metrics endpoint.
-func getPodRequestCount(podName string) int {
+func getPodRequestCount(nsName string, podName string) int {
```
elevran: nit 
```suggestion
func getPodRequestCount(nsName, podName string) int {
```

test/e2e/e2e_suite_test.go
```diff
@@ -266,43 +297,40 @@ func setupK8sClient() {
...
+	ginkgo.By("Ensuring namespace exists: " + getNamespace())
 	testutils.EventuallyExists(testConfig, func() error {
 		return testConfig.K8sClient.Get(testConfig.Context,
-			types.NamespacedName{Name: nsName}, &corev1.Namespace{})
+			types.NamespacedName{Name: getNamespace()}, &corev1.Namespace{})
 	})
 }
 
 // createCRDs creates the Inference Extension CRDs used for testing.
 func createCRDs() {
```
elevran, line 324: suggestion: cluster-scoped `crdObjects` bookkeeping lives only on process 1; if process 1 races another process for CRD creation, its `ReportAfterSuite` deletion could remove CRDs other processes still need. Not exercised today since parallelism isn't enabled, but worth fixing (or a follow up with processes > 1).

elevran (commented):
Solid infrastructure groundwork for parallel E2E - namespace threading and the port-per-process scheme are consistent and match issue #1931's ask. One thing to flag before the follow-on PR turns on `-p`: the `e2e-epp-metrics` Service (deploy/environments/dev/e2e-infra/services.yaml, not touched by this diff) still hardcodes `nodePort: 32090` while `getMetricsPort()` here computes a per-process value - once parallelism is enabled that Service will collide across processes. Not a defect in this PR since it doesn't touch that block, but worth a tracking note or follow-up issue since it means the NodePort path isn't actually parallel-safe yet.

shmuelk:
@elevran wrote:

>One thing to flag before the follow-on PR turns on -p: the e2e-epp-metrics Service (deploy/environments/dev/e2e-infra/services.yaml, not touched by this diff) still hardcodes nodePort: 32090 while getMetricsPort() here computes a per-process value - once parallelism is enabled that Service will collide across processes. Not a defect in this PR since it doesn't touch that block, but worth a tracking note or follow-up issue since it means the NodePort path isn't actually parallel-safe yet.

Indeed in this PR the nodePorts in the services are not parallel safe. In the next PR the setup and teardown changes and will modify those nodePorts are appropriate.

test/e2e/e2e_suite_test.go
shmuelk: In the next PR the use of BeforeSuite is replaced with SynchronizedBeforeSuite, which invokes two functions, one before process one (the first one) and another one before each and every other one.

The cluster is created and the CRDs are registered before process one runs.

test/e2e/e2e_suite_test.go
shmuelk: In the next PR which exploits Ginkgo's parallelism feature, as there can be more tests to run in parallel than processes, each test sets itself up (namespace, envoy, inference pool, Service Account, RBAC, etc) runs the test(s) and then tears down what was setup.

This actually not done per test, but down per group of tests, grouped by a Ginkgo "When block".

test/e2e/e2e_suite_test.go
shmuelk: In the next PR the use of BeforeSuite is replaced with SynchronizedBeforeSuite, which invokes two functions, one before process one (the first one) and another one before each and every other one.

The cluster is created and the CRDs are registered before process one runs.

elevran (approved):

PR #1978, opened 2026-07-12
feat: add balanceBy tokens option to burst prefix cache producer

### LINKED ISSUE #1977: Burst prefix-cache producer: balance by tokens
## Description
The burst prefix-cache producer (`burst-prefix-cache-producer`) balances co-located prompt-sharing units across replicas by request count. On high prompt-length-variance workloads an even request count does not imply an even token/KV load, so a replica drawing several long prompts KV-saturates and straggles the window makespan.

Concrete example (RL rollout burst): a rollout step emits 8 GRPO groups in one window, each an identical-prompt unit of G=8 samples (64 requests total), spread over 4 replicas. Request-count balancing places 2 groups (16 samples) per replica, perfectly even by count. But the prompts vary - say 2 long-context groups at ~32k prompt tokens and 6 short groups at ~4k. With prefix caching each group's prompt is prefilled once on its replica, so a replica's prefill/KV load is the sum of its groups' prompt lengths, not its sample count. If the two long groups happen to land on the same replica it holds ~64k distinct prefix tokens while another holds ~8k: an ~8x token/KV imbalance at identical request counts (16 samples each). The heavy replica saturates KV and prefills ~64k tokens while the light one idles at ~8k, and since the generation phase ends only when the slowest replica finishes, that one replica bounds the whole step.

## Proposed solution
Add a `balanceBy` option (`requests` default, or `tokens`) that balances by prefix-block (token) load instead, charging each unit's prefix once per replica and discounting blocks the replica already holds. In the example above, token balancing would place one long group per replica (or spread them) so no replica stacks both 32k prefixes. Placement-only, opt-in, back-compatible. 

**Possible follow-ups**: a lexicographic tie-break or a weighted mode combining tokens and request count (see PR for detail).

ezrasilvera:
## What this PR does

Adds a `balanceBy` option to the burst prefix-cache producer (`burst-prefix-cache-producer`): `requests` (default) or `tokens`. In `requests` mode the producer balances co-located units across replicas by request count, the existing behavior. In `tokens` mode it balances by prefix-block (token) load: each unit's prefix is charged once per replica it lands on, discounting the leading blocks that replica already holds (prefilled by an earlier request sharing the prefix). On workloads with high prompt-length variance, request-count balancing leaves the token/KV load uneven, so a replica drawing several long prompts KV-saturates and straggles the window makespan; token balancing spreads long-prompt units so no replica saturates on stacked prefixes while prefix-sharing units still pack cheaply. 
**The change is placement-only and fully back-compatible: the default preserves current behavior and token balancing is opt-in.** Implementation reuses the per-unit prefix-block count already computed in the producer; only the balancing unit (and a shared-prefix discount) changes, so the co-location and fair-share logic is unchanged. Configured as `balanceBy: tokens` under the producer parameters; an empty or omitted value defaults to `requests`, and any other value is rejected at construction.

Fixes #1977 

## Observed motivation (example RL workload)

On an example RL rollout workload with high prompt-length variance (prompts from a few thousand up to tens of thousands of tokens), request-count balancing kept request counts near-even but left the per-replica prefix-token load roughly 6x imbalanced, so one replica saturated KV (100%) and became a straggler while others idled. Token balancing spread the long prompts across replicas (no KV saturation), collapsing the straggler while keeping co-location's low mean per-request time. Native (per-request) routing is shown as a reference.

| metric (example RL workload)  | `balanceBy: requests` | `balanceBy: tokens` |
|---|---|---|
| rollout generation time / step | 99.1s | 52.8s |
| slowest replica (straggler) | 96.1s | 49.2s |
| per-replica peak KV | 31%-100% (saturated) | 48%-93% (no saturation) |

On this workload token balancing beats request balancing by ~47% and native by ~24% on rollout generation time. Preliminary single-workload measurement; this is a routing-only change, convergence unchanged.

## Future work / alternatives (not in this PR)**

Balancing is currently `requests | tokens` (either/or). Two ways to combine both quantities, deferred for discussion: 

1. a near-free lexicographic tie-break inside token mode - balance by prefix blocks primarily, break ties by batch-assigned request count - which needs no tuning knob and degenerates to sensible behavior in both the long-prompt and short-prompt regimes
2. A full `balanceBy: weighted` third mode - a weighted sum of prefix blocks and request count - which is the most expressive but needs scale normalization between the two quantities and a weight knob to tune. The string enum (rather than a boolean) was chosen specifically so either can be added as a new value without a breaking rename. Neither is built here because the measured pathology is a pure token imbalance (request count was already balanced), so a blend would not beat pure token mode on that workload; a weighted mode is worth revisiting if a decode-bound or mixed workload shows request count also mattering.

## Test plan:

- [x] token balancing spreads a long-prefix unit off the replica holding short-prefix units, and packs the short units together (contrasted against request-count balancing, which co-locates a short unit with the long one)
- [x] the shared-prefix discount keeps units that share a leading prefix co-located under token balancing, rather than spilling once the naive full-prefix charge would exceed the fair share
- [x] invalid `balanceBy` value is rejected at producer construction
- [x] all existing burst-producer placement tests pass unchanged under the default (`requests`) mode
- [x] `make presubmit` passes (format, lint, govulncheck, mod, tag checks)

## Release note:
```release-note
burst-prefix-cache-producer: add a `balanceBy` parameter (`requests` (default) or `tokens`). With `tokens`, placement is balanced by prefix-block (input token) load instead of request count, charging each unit's prefix once per replica and discounting blocks it already holds, to avoid KV-saturating a single replica on high prompt-length-variance workloads. Default behavior is unchanged.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/README.md
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/README.md
@@ -61,6 +61,7 @@ This producer emits `PrefixCacheMatchInfo`; it does not score. Reuse the
 | `maxPrefixTokensToMatch` | 0 | cap on matched prefix tokens; 0 uses the default block cap. A positive value must be >= `blockSizeTokens`, otherwise it yields zero prefix blocks. |
 | `minColocateBlocks` | 0 | min shared leading blocks for inter-unit co-location and for a single request to gain an affinity; 0 disables both (only identical groups are placed, purely load-balanced) |
 | `maxBatchSize` | 1000 | max requests one window may accumulate; Produce returns an error once reached. -1 = unlimited. |
+| `balanceBy` | `requests` | quantity balanced across replicas within a window: `requests` (by request count) or `tokens` (by prefix-block load, charging each unit's prefix once per replica and discounting the leading blocks already held there). Token balancing spreads long-prompt units so no replica saturates on stacked prefixes while request counts stay even. |
 
 ## Operational notes
 
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/burstprefix/assign.go
@@ -145,7 +145,7 @@ outer:
 // over the batch (longest-prefix first) so a shared prefix is prefilled once per
 // replica rather than scattered. Identical groups are placed before prefix-sharing
 // singletons so proven same-prompt co-location anchors the layout.
-func assign(entries []*entry, k, minColocateBlocks int) {
+func assign(entries []*entry, k, minColocateBlocks int, byTokens bool) {
 	groups := map[string][]*entry{}
 	order := []string{}
 	var replicas []fwksched.Endpoint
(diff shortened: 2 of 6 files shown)


elevran (commented):
lgtm
please add a release-note section to the PR description body.

ezrasilvera:
> please add a release-note section to the PR description body.

Done

elevran (approved):

PR #1986, opened 2026-07-13
fix(tokenizer): convert Anthropic tools to OpenAI schema for /render

### LINKED ISSUE #1861: Anthropic /v1/messages not tokenized by the render (vLLM) token-producer backend — precise prefix cache never hits
Hi maintainers, thanks for your work on llm-d-router.

**What happened**:

Requests parsed from the Anthropic `/v1/messages` API are not tokenized by the render (vLLM HTTP `/render`) token-producer backend, so they are excluded from precise prefix-cache routing.

The render backend's `produce()` switch in `pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go` only handles `body.Completions`, `body.ChatCompletions`, and `body.Generate`. The Anthropic parser populates `body.Messages` (`*MessagesRequest`), which falls through to the `default` case and returns `unsupported request body type, skipping tokenization`.

With no `TokenizedPrompt`, `precise-prefix-cache-producer` has nothing to look up, `kvcache_index_lookup_hits_total` never increments, and the `prefix-cache-scorer` logs:

```
PrefixCacheMatchInfo not found for endpoint, assigning score 0
```

This mirrors #1507, which added the `Messages` branch to the **estimate** backend (`estimate.go`) but left the render backend unchanged.

Anthropic `/v1/messages` requests should be tokenized via the vLLM `/render` path (same as `/v1/chat/completions`), so precise prefix-cache routing works and cache-hit metrics increment.

1. Deploy EPP with `precise-prefix-cache-routing-plugins.yaml`, configuring `token-producer` with a `vllm` backend (`vllm.url: http://localhost:8000`).
2. Send traffic to `/v1/chat/completions` or `/v1/completions` -> `kvcache_index_lookup_hits_total` increases, no error logs.
3. Send equivalent traffic to `/v1/messages` (e.g. via an Anthropic Messages client) -> `kvcache_index_lookup_hits_total` does not increase, and the EPP logs `PrefixCacheMatchInfo not found for endpoint, assigning score 0` on every request.

**Environment**:
- llm-d-router version / image: `ghcr.io/llm-d/llm-d-router-endpoint-picker-dev:main` (confirm exact tag/digest)
- Others: model `Qwen/Qwen3.6-27B-FP8`, render sidecar `vllm/vllm-openai-cpu:v0.19.1`

dkhachyan:
**What type of PR is this?**

  /kind bug

   Follow-up to https://github.com/llm-d/llm-d-router/pull/1896.

  Anthropic Messages requests (`/v1/messages`) failed tokenization against the
  vLLM `/render` endpoint when the request carried `tools`. The router forwarded
  Anthropic-shaped tool definitions (`{name, description, input_schema}`)
  verbatim, but vLLM's OpenAI-compatible `/render` expects the function schema
  (`{type: "function", function: {name, description, parameters}}`), so it
  rejected the request with HTTP 400 (`Field required: body.tools[0].function`).
  This broke the token-producer path and, in turn, precise-prefix-cache routing
  for Claude Code traffic.

  `MessagesToRenderChatRequest` now maps Anthropic custom tools into the OpenAI
  function schema (`input_schema` -> `parameters`, nested under `function`).
  Tools that are already OpenAI-shaped (carry a `function` field) or are not
  object-shaped pass through unchanged.

  Fixes https://github.com/llm-d/llm-d-router/issues/1861


sagearc (commented):
Thanks @dkhachyan, please fix your commits sign-off
https://github.com/llm-d/llm-d/blob/main/PR_SIGNOFF.md

PR #1988, opened 2026-07-13
test(e2e): wait for EPP pod discovery instead of sleeping

### LINKED ISSUE #764: fix(e2e): replace time.Sleep with deterministic pod-discovery wait

elevran:
/kind test
/kind flake

The EPP reports gRPC health as SERVING as soon as the InferencePool object is set, even before the pod-discovery controller has finished its initial list. Several e2e tests raced against that discovery by using a fixed sleep, or by issuing requests immediately after EPP readiness with no wait at all.

This PR adds `waitForEPPToDiscoverPods`, which polls the EPP metrics for readiness and number of discovered endpoints > 0, and calls it synchronously from `createEndPointPicker`, the chokepoint nearly all e2e tests already funnel through. Reorders the two KV tests to create model servers before the EPP so the wait has pods to find, and drops the now-redundant sleeps.

Also fixes `startEPPMetricsPortForward`, which was called from six sites and unconditionally started a new `kubectl port-forward` each time, leaking orphaned processes. It now guards on an existing session, and `AfterSuite` resets the session variable to nil after terminating it. The existing-cluster (`k8sContext`) check moved into the function itself instead of being duplicated at every call site.

Fixes #764

```release-note
NONE
```

### DIFF
--- a/test/e2e/utils_test.go
+++ b/test/e2e/utils_test.go
@@ -118,6 +118,26 @@ func getPodNames(labels map[string]string) []string {
 	return names
 }
 
+// waitForEPPToDiscoverPods blocks until the EPP's inference_pool_ready_pods
+// gauge for poolName reports at least one pod, indicating the InferencePool
+// controller has finished its initial pod discovery. The EPP reports gRPC
+// health as SERVING as soon as the pool is set, even if pod discovery found
+// zero endpoints, so readiness alone does not guarantee the datastore is
+// populated. Polling the gauge avoids routing a real request through the
+// EPP, which would otherwise be recorded as a routing decision and skew
+// tests that assert exact decision-type counts.
+func waitForEPPToDiscoverPods(poolName string) {
+	ginkgo.By("Waiting for EPP to discover pool members")
+	metricsURL := fmt.Sprintf("http://localhost:%d/metrics", getMetricsPort())
+	if k8sContext != "" {
+		startEPPMetricsPortForward()
+	}
+	labelMatch := fmt.Sprintf(`name="%s"`, poolName)
+	gomega.Eventually(func() int {
+		return getCounterMetric(metricsURL, "inference_pool_ready_pods", labelMatch)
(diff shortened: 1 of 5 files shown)


test/e2e/utils_test.go
```diff
@@ -118,6 +118,26 @@ func getPodNames(labels map[string]string) []string {
...
+// gauge for poolName reports at least one pod, indicating the InferencePool
+// controller has finished its initial pod discovery. The EPP reports gRPC
+// health as SERVING as soon as the pool is set, even if pod discovery found
+// zero endpoints, so readiness alone does not guarantee the datastore is
+// populated. Polling the gauge avoids routing a real request through the
+// EPP, which would otherwise be recorded as a routing decision and skew
+// tests that assert exact decision-type counts.
+func waitForEPPToDiscoverPods(poolName string) {
+	ginkgo.By("Waiting for EPP to discover pool members")
+	metricsURL := fmt.Sprintf("http://localhost:%s/metrics", metricsPort)
```
shmuelk: Be ready for parallel tests:


```suggestion
	metricsURL := fmt.Sprintf("http://localhost:%d/metrics", getMetricsPort())
```

elevran: Fixed in commit 93f42516. Replaced the removed static metricsPort variable with getMetricsPort() as suggested.

shmuelk (approved):

ahg-g (approved):

PR #1999, opened 2026-07-14
Add generic endpoint attribute filter

### LINKED ISSUE #1922: Generic endpoint attribute filter
Similar to #977 , we create a filter implementation.

A generic filter that applies to a configurable (scalar) attribute of an endpoint.

Example config:
```
plugins:
  - type: endpoint-attribute-filter
    parameters:
      attribute: "num_requests_running"
      
      # Common attribute lifecycle policies
      onMissing: "Pass"              # Pass | Fail (Default: Pass)
      fallbackOnEmpty: true          # Default: false
      
      # Algorithm specification
      algorithm:
        type: "threshold"            # threshold | percentile | topK | range
        threshold:
          operator: "LessThan"       # LessThan | LessThanOrEqual | GreaterThan | GreaterThanOrEqual | Equal | NotEqual
          value: 10
```

Other algorithm choices:

```
# Algorithm: Relative Percentile (e.g., drop endpoints in top 10% of load)
algorithm:
  type: "percentile"
  percentile:
    cutoffPercentile: 90
    keep: "below"                    # below | above

---
# Algorithm: Top-K (e.g., keep the K least-loaded endpoints)
algorithm:
  type: "topK"
  topK:
    k: 5
    order: "Ascending"              # Ascending (lowest first) | Descending

---
# Algorithm: Range / Interval (e.g., keep endpoints with CPU between 10% and 80%)
algorithm:
  type: "range"
  range:
    min: 0.10
    max: 0.80
    inclusive: true

```

sudoalok:
Adds an endpoint-attribute-filter scheduling plugin that filters candidate endpoints on one configured endpoint attribute, following the config from #1922. This covers the threshold algorithm plus the onMissing / fallbackOnEmpty policies. percentile, topK and range can come as follow-ups like discussed on the issue.

A few decisions I made along the way, open to changing any of them:

- onMissing defaults to Pass and fallbackOnEmpty to false, same defaults as the issue
- only "threshold" is accepted for algorithm.type for now, anything else fails validation. felt safer that a config meant for percentile/topK/range fails loudly instead of quietly passing every endpoint through
- the threshold block itself is required, and the operator has to be one of the six values from the issue
- fallbackOnEmpty only kicks in when the filter dropped every endpoint. if even one survives, no fallback
- endpoints missing the attribute never reach the comparison, Pass keeps them and Fail drops them
- Consumes() has the same shape as the endpoint-attribute-scorer (#1620) so the two stay consistent. happy to move both to the newer DataDependencies form in a follow-up once the metrics extractor declares its produced keys

Added table tests for the factory validation, all six operators, and the policy/fallback cases. fmt/vet/tests clean, config loader and runner suites pass too.

Fixes #1922

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -101,6 +101,7 @@ import (
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requesthandling/parsers/vllmgrpc"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requesthandling/parsers/vllmhttp"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/filter/bylabel"
+	endpointattributefilter "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/filter/endpointattribute"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity"
 	sessionaffinityfilter "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/filter/sessionaffinity"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/filter/sloheadroomtier"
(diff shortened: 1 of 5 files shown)


liu-cong (approved):
thanks!

PR #2004, opened 2026-07-14
feat: emit per-endpoint scores in EPP dynamic metadata

### LINKED ISSUE #1843: Return per-endpoint scores in EPP dynamic metadata

adelsam:
**What type of PR is this?**
/kind feature

The scheduler computes a weighted score per candidate endpoint, but that score is
dropped at the `ProfileRunResult` boundary, so downstream consumers of the ext-proc
dynamic metadata only see which endpoints were chosen, not how strongly they were
preferred or how the alternatives compared.

This PR implements the proposal from #1843 (design doc in #1844) in two layers:

1. Record the scores in the scheduler profile: additive, nil-safe
   `ScoredCandidates` on `ProfileRunResult`, covering every endpoint the profile
   scored, including candidates the picker did not select. It is populated in
   `runPickerPlugin` from the weighted score map, so pickers are unchanged and
   score availability does not depend on which picker is configured. The complete
   set is already materialized there, so this costs no extra allocation.
2. Emit them on the wire, opt-in: with the new `--emit-endpoint-scores` flag
   (off by default), the request-path dynamic metadata carries an
   `x-gateway-destination-endpoint-scores` struct under the existing `envoy.lb`
   namespace, mapping each scored endpoint to its score. Only the primary
   profile's scores are emitted.

`x-gateway-destination-endpoint` (header and metadata) is unchanged and remains the
sole source of truth for endpoint selection and ordering; with the flag off the
dynamic metadata is byte-identical to today. Scores are the post-weight aggregate
emitted as raw doubles; no range or precision guarantee is part of the contract.
`ScoredCandidates` ordering is unspecified, so consumers key by endpoint.

Sourcing the scores from the scheduler profile rather than from the picker's
selection is a change made in review. Pickers truncate to `maxNumOfEndpoints`, so
picker-sourced scores describe only the winners, which is a single score under the
default configuration. That cannot express how the alternatives compared, which is
the runner-up visibility motivating the proposal. The first two commits are kept
as pushed so the rework is readable as a delta; they are ready to squash on merge.

**Test plan**:
- [x] A profile run records the score of every endpoint it scored, including
      candidates the picker did not select (verified to fail when the recording
      is removed)
- [x] Director propagates the primary profile's scores onto the request context,
      including endpoints absent from the destination list, and leaves them nil
      for unscored results
- [x] Request-header response metadata carries the scores struct when the flag is
      on and scores exist; the key is absent when scores are missing or the flag
      is off
- [x] Hermetic ext-proc tests: with `--emit-endpoint-scores`, `envoy.lb` metadata
      maps every scored candidate to its score, including one the picker did not
      select; by default the key is absent and the metadata is unchanged

Fixes #1843

**Release note**:
```release-note
The Endpoint Picker can optionally emit the scheduler's per-endpoint scores in the request-path dynamic metadata under the `x-gateway-destination-endpoint-scores` key of the `envoy.lb` namespace, enabled with the new `--emit-endpoint-scores` flag (off by default). Known gap: no scores are emitted under the deprecated `data-parallel-profile-handler`, or under the deprecated `pd-profile-handler` when `primaryPort` is set, because those handlers rebuild the scheduling result without carrying the scores through.
```


elevran:
suggest /hold until we close implementation details in #1844

elevran:
> suggest /hold until we close implementation details in #1844

@adelsam did you have a chance to review comments on #1844?

adelsam:
Resolved discussion on https://github.com/llm-d/llm-d-router/pull/1844 and rebased this branch. Waiting for CI to run.

pkg/epp/framework/interface/scheduling/types.go
```diff
@@ -164,6 +164,11 @@ type ScoredEndpoint struct {
 // ProfileRunResult captures the profile run result.
 type ProfileRunResult struct {
 	TargetEndpoints []Endpoint
+	// ScoredCandidates carries the weighted score of every endpoint the profile
+	// scored, including candidates the picker did not select. TargetEndpoints is
+	// a subset of these, so the two are not parallel. Ordering is unspecified;
+	// consumers key by endpoint. Nil when the profile ran no scorers.
```
elevran: nit: 
"Nil when the profile ran no scorers" doesn't match the implementation. `runScorerPlugins` pre-populates every filtered endpoint with a score of 0 before running scorers, so a profile with zero scorers still produces a non-nil ScoredCandidates (with all zeros). 
The only way to get nil appears to be a picker returning a nil result. Consider rewording, e.g. "Nil only if the picker returns a nil result."
Not sure I would mention it.

elevran (commented):
Clean implementation of the endpoint-scores proposal.

One gap worth flagging (not introduced by this PR's diff, but relevant to its "score availability does not depend on which picker is configured" claim): the legacy dataparallel.ProfileHandler.ProcessResults and disagg.PdProfileHandler.ProcessResults with primaryPort rewriting active both rebuild a fresh ProfileRunResult without copying ScoredCandidates through. Deployments still using these deprecated handlers will get an empty scores map with `--emit-endpoint-scores` on even though the profile did score candidates.
Since both handlers are deprecated this is likely an acceptable known gap, but worth a comment noting it for awareness (e.g., in release-notes).
Other then that - small comment nit inline.

adelsam:
Thanks @elevran! I incorporated your feedback.

elevran (approved):

PR #2008, opened 2026-07-14
feat(flowcontrol): expose program-aware fairness state via DumpState

### LINKED ISSUE #1798: [Operations] Implement DumpState for program-aware-fairness plugin

chethanuk:
**What type of PR is this?**

/kind feature

Implements `plugin.StateDumper` for the `program-aware-fairness` flow-control plugin so it participates in the `/debug/plugins/state` endpoint.

`DumpState` reports the active strategy, Jain's fairness index, and per-program fairness accounting: dispatched count, in-flight count, average queue wait, wait observations, LAS attained service, and last completion time. Programs are sorted busiest first (in-flight, then dispatched count) and the list is capped at 1000 entries with `totalPrograms`, `maxPrograms`, and `truncated` fields so operators can tell when the dump is partial.

Per-program values are read under their own per-entry locks, so the snapshot is best-effort rather than a single instant; this avoids a global lock contending the scheduling hot path. `LastCompletionTime` is emitted only after at least one request has completed, so the eviction-seed timestamp is never reported as a completion.

```mermaid
sequenceDiagram
    participant H as /debug/plugins/state handler
    participant P as ProgramAwarePlugin.DumpState
    participant M as programMetrics (sync.Map)
    participant S as LASStrategy.SnapshotService

    H->>P: DumpState()
    P->>S: snapshot attained service per program
    P->>M: range per-program metrics
    P->>P: sort busiest first, cap at 1000
    P-->>H: bounded JSON state
```

Example response state:

```json
{
  "strategy": "las",
  "jainsFairnessIndex": 0.94,
  "programs": [
    {
      "programID": "team-a",
      "dispatchedCount": 42,
      "inFlight": 3,
      "averageWaitTimeMs": 12.5,
      "waitCount": 40,
      "attainedService": 8200,
      "lastCompletionTime": "2026-07-14T10:00:00.123Z"
    }
  ],
  "totalPrograms": 1,
  "maxPrograms": 1000,
  "truncated": false
}
```

**Testing**:

- [x] Empty and zero-value plugin returns valid JSON with an empty (`[]`) program list and no panic
- [x] Completed request reports dispatched/wait counts, attained service, and completion time
- [x] In-flight request reports zero attained service and no completion time
- [x] Multiple programs are ordered busiest first
- [x] Program list truncates at the cap with `truncated` and `totalPrograms` set
- [x] Payload contains only the documented keys
- [x] `SnapshotService` reflects accumulated per-program service

Fixes #1798

```release-note
NONE
```


LukeAVanDrie (changes requested):
Thanks for this, the implementation itself is well done! Unfortunately the PR has been overtaken by #1839, which merged earlier today and closed #1798: main now has a `DumpState` for this plugin, so this branch is in hard merge conflict. More importantly, the merged version is deliberately aggregates-only. The code comment says program
IDs "come from a user-controlled request header, so they are omitted", and it landed with a `TestDumpStateOmitsProgramIDs` regression test. A per-program dump was discussed on #1839: @liu-cong asked for exactly this shape (full `programMetrics`, capped at ~1000), @thc1006 offered o follow up with it after landing, and the thread flagged that program-ID exposure interacts with the metrics-cardinality work in #1625, so it deserves its own discussion.

So rather than closing outright, I'd suggest:
1. Rebase onto current main, building on top of #1839's merged implementation.
2. Reframe this PR as the per-program follow-up from the #1839 thread, and sync with @thc1006, who had claimed it, to avoid duplicate work.
3. Address the sanitization question explicitly in the PR description: why per-program IDs are acceptable in the debug dump given the merged exclusion comment, the regression test this would remove, and the #1625 concern (or propose capping/hashing/omitting the IDs). That call is @liu-cong's to make.

thc1006:
@chethanuk Sorry for the collision, and thanks for doing the work. I did say on #1839 I'd follow up with the per-program dump, but you've already written it, so please carry it rather than let that go to waste. I can review or help if that's useful.

Two things from the #1839 thread that might save you time. The merged version is aggregates-only on purpose and ships a `TestDumpStateOmitsProgramIDs` test, so going per-program means deliberately removing it and explaining why. And @liu-cong asked for the full `programMetrics` capped around 1000, sorted by in-flight. The sanitization call is his: program IDs already appear as the `program_id` metric label, so it isn't new exposure, but it does touch the cardinality work in #1625.

chethanuk:
> so please carry it rather than let that go to waste. I can review or help if that's useful.

Thank you @thc1006, carrying it as you suggested -- rebuilt on #1839 as the per-program follow-up. Please review the PR :) 

- On "isn't new exposure": agree the raw ID already exists as the program_id label, but the dump
  is a one-shot census of all programs on the metrics server (auth is on by default but
  disableable via --metrics-endpoint-auth), which is a different surface than per-series metric
  reads. So default is sha256-hashed; raw stays liu-cong's call as you said.
- #1625 interaction: cap at 1000, hash + service lookup only for retained rows after the cap,
  per-ID read-only lookup instead of a full map copy -- the dump doesn't add a new amplification
  lever on the fairness_id map.

chethanuk:
> PR has been overtaken by https://github.com/llm-d/llm-d-router/pull/1839, which merged earlier today and closed https://github.com/llm-d/llm-d-router/issues/1798: main now has a DumpState for this plugin

Thank for the detailed review @LukeAVanDrie, 
Done, all three:
- Rebuilt on top of merged #1839 (not a rebase of the stale branch). Purely additive: the totalPrograms / totalInFlight / fairnessIndex keys are unchanged.
- Reframed as the per-program follow-up liu-cong asked for on #1839 ("dump the entire programMetrics (can configure an upper limit of, say 1000)"); 
- Sanitization, in the description and code: IDs are sha256-hashed by default, so #1839's omission survives (guard test is now TestDumpStateHashesProgramIDs: raw ID absent, hash present). Hashing is surface hygiene / a stable pseudonym, not confidentiality -- low-entropy IDs are dictionary-reversible and already a program_id metric label. Raw is a one-line seam flip, but it would override #1839's omission and #1798's "without exposing sensitive data" criterion -- so that stays @liu-cong's call, not a default.
- On #1625: the program ID here is exactly the fairness_id header that issue names, so the dump caps at 1000 rows (truncated flag) and defers hashing + per-ID service lookup until after the cap; the full-map service snapshot is replaced by a read-only per-ID lookup. Payload and per-row work are bounded; the one cheap row scan is the same O(N) class as #1839's existing aggregate pass.
- Tests: cap/truncation, sort order, hash guard, completion-time semantics, orphaned-program fallback; package tests and targeted -race pass.

LukeAVanDrie (approved):
Thanks for the quick turnaround @chethanuk! 

FYI, the PR description still shows the previous design.

@liu-cong, there is one open decision still. The dump defaults to sha256-hashed program IDs to preserve #1839's raw-ID omission while giving you the per-program list you asked for. If you'd rather have raw IDs for operability (they do already appear as the `program_id` metric label), that's a one-line change. What are your thoughts here?

liu-cong:
The purpose of this is for debugging, so the meta question is: is the state we dump here helpful for debugging? For example, what can I get from the sanitized program ID? 

@LukeAVanDrie You have more context than I do. So deferring to Luke for approval.

PR #1853, opened 2026-06-30
server: add plugin name to default plugin state debug entry

MOHITPARMAR007:
## What type of PR is this?

/kind feature

## What this PR does / why we need it

Adds a `Plugin` field to `pluginStateDebugEntry` so the `/debug/plugins/state` endpoint includes the plugin's instance name explicitly in each entry, not just as the map key.

This repurposes #1803 per maintainer feedback (see discussion in #1833): the `round-robin-fairness-policy` plugin has no centralized state to dump, since its cursor is stored per priority band rather than on the plugin itself. Instead of implementing a plugin-specific `DumpState` for round-robin, the plugin name is now included in the default debug entry for **all** plugins (both `StateDumper`-implementing and non-implementing plugins).

## Which issue(s) this PR fixes

Fixes #1803

## Testing

Updated existing unit tests (`TestPluginStateDebugHandlerIncludesPlugins`, `TestPluginStateDebugHandlerReportsPluginStateErrors`) to assert the new `plugin` field is present in the JSON response.

## Release note

```release-note
NONE
```

PR #1833, opened 2026-06-27
flowcontrol/roundrobin: implement DumpState for round-robin-fairness-…

MOHITPARMAR007:
## What type of PR is this?

/kind feature

## What this PR does / why we need it

Implements `plugin.StateDumper` on the `round-robin-fairness-policy` plugin as part of the tracking issue #1755.

The round-robin cursor is stored per priority band on the band's `PolicyState`, not on the plugin itself, so there is no centralized mutable state to snapshot. `DumpState` returns plugin identity information (name and type) only.

## Which issue(s) this PR fixes

Fixes #1803

## Testing

Added unit tests covering:
- Default plugin name
- Custom plugin name  
- Valid JSON output

## Release note

```release-note
NONE
```

Issue #1803, opened 2026-06-26
[Operations] Add pluginName to the default DumpState


liu-cong:
Parent tracking issue: #1755

Plugin registration constant: `roundrobin.RoundRobinFairnessPolicyType`
Plugin package: `github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/fairness/roundrobin`

Acceptance criteria:
- Plugin implements `plugin.StateDumper` with bounded, sanitized output.
- Add/update tests that cover `DumpState` behavior and error handling.
- Debug output remains operationally useful without exposing sensitive data.

MOHITPARMAR007:
hi @liu-cong , can i work on this ?

liu-cong:
Thanks @MOHITPARMAR007 , assigned it to you

MOHITPARMAR007:
Hi @liu-cong , could you help me understand why my PR is not passing all the checks? I'm new to open source, so pardon me if this is a basic question.

liu-cong:
@MOHITPARMAR007 To troubleshoot the check failures, click on the failed tests and it should give you some error log. For this case, you failed the signed commit test `  For instructions on how to set up GPG/SSH signing and verify your commits, please see [GitHub Documentation](https://docs.github.com/en/authentication/managing-commit-signature-verification).
`

BTW, it looks like this particular plugin doesn't really have any state to be dumped. Let's repurpose this issue to add the plugin name to the default `DumpState`, see comment in https://github.com/llm-d/llm-d-router/pull/1833#issuecomment-4826224983

PR #2011, opened 2026-07-14
feat(flowcontrol): expose round-robin fairness identity via DumpState

chethanuk:
**What type of PR is this?**

/kind feature

Implements `StateDumper` on the round-robin fairness flow-control policy so `/debug/plugins/state` reports its identity and state-ownership model instead of the generic "plugin does not support state collection" message.

The plugin instance is stateless by design: per-band rotation cursors are created by `NewState` and owned by the flow-control registry, one per priority band, so no cursor position is reachable from the plugin. The dump makes that explicit, so operators inspecting the endpoint know the policy is wired in and know where its runtime state actually lives, rather than reading the unsupported message as a gap.

```mermaid
flowchart LR
    A["/debug/plugins/state handler"] --> B["StateDumper.DumpState"]
    B --> C["static snapshot JSON: policy, stateful, cursorOwner"]
```

Example entry:

```json
{
  "name": "round-robin-fairness-policy",
  "type": "round-robin-fairness-policy",
  "state": {
    "policy": "round-robin-fairness-policy",
    "stateful": false,
    "cursorOwner": "flow-control registry, per priority band"
  }
}
```

The payload is fixed and bounded regardless of band or flow cardinality, and carries no flow IDs or request data.

Related to #1755

N/A

```release-note
NONE
```

**Testing**:

- [x] Dump returns valid JSON with exactly the policy identity, stateless flag, and cursor-owner fields, for default and custom plugin names and a zero-value receiver
- [x] Dump is byte-identical before and after `Pick` advances a band cursor and never contains flow IDs


shmuelk (approved):

LukeAVanDrie (commented):
Thanks for the contribution! The implementation is clean, the GoDoc is clear, and the tests are thorough.

Before this merges, though, I want to flag some history this PR doesn't reference, because itvappears to reverse a prior decision. This plugin was covered by #1803 (child of the #1755 tracking issue linked here). Two earlier PRs implemented identity-only `DumpState` on it — #1828 and #1833 — and both were closed after review: @shmuelk suggested adding the plugin name to the *default* debug entry instead, and @liu-cong agreed on #1803 ("it looks like this particular plugin doesn't really have any state to be dumped"). That handler-level fix merged as #1853, so `/debug/plugins/state` already reports `name`, `type`, and an explanatory message for every non-`StateDumper` plugin — which covers most of what this PR's description is after.

`docs/plugin_debug.md` also documents the unsupported message as intended behavior, and #1755 scoped `DumpState` to plugins with *mutable runtime state* (this one was classified among the dozens without any).

@shmuelk since you've approved: is the intent to revisit that position? The one thing this dump adds over #1833's version is the `stateful`/`cursorOwner` fields — a "no state here by design, cursors live in the registry" signal the default entry can't express. If that's the argument for merging, it seems worth deciding on #1755 whether that becomes a convention (and if so, whether it should be handler-level rather than a constant payload re-implemented in each of the dozens of stateless plugins). If not, this seems superseded by #1853.

liu-cong:
The purpose of the dumpstate is to get some internal runtime state for debugging purpose, that cannot be obtained/inferred from other sources like plugin config. The info added in this PR is "static". I don't find it useful.

Please reopen if there is a disagreement. Thanks!

PR #2015, opened 2026-07-14
Revert "[tests] Adds e2e tests for the /inference/v1/generate endpoint."

ahg-g:
Reverts llm-d/llm-d-router#1433

As per @roytman , one of the tests are failing, which is strange since it should be blocking the merge.


elevran (approved):
reverting to allow further analysis

Issue #1593, opened 2026-06-11
[RFC] Shared Scheduling State for Horizontally Scalable EPP Deployments


nicole-lihui:
## Summary

Today, EPP supports high availability through leader election,
but only a single EPP instance actively handles ext-proc traffic.

While this pattern provides failover capability, it does not increase request processing capacity,
and EPP becomes a potential bottleneck under high QPS and long-context inference workloads.

This RFC proposes introducing optional shared scheduling state. With default behavior unchanged, it adds an optional active-active operating mode:

- `classic` preserves the current default behavior; existing deployments require zero changes.
- `stateful` EPP holds shared scheduling state and exposes read/write access through an internal gRPC State API.
- `stateless` EPP scales horizontally to handle ext-proc traffic, accessing global shared state from stateful EPP via gRPC.

This allows multiple EPP instances to process requests concurrently while preserving existing scheduling semantics:

- Scheduling computation logic such as scorer, filter, and picker continues to run locally, without changing Gateway API, InferencePool API, or existing plugin semantics.
- The local shadow always runs. If remote state is unavailable, the system gracefully degrades to local-only selection—EPP stability takes priority over global accuracy.

We welcome community feedback on the design and tradeoffs described in this RFC.

## Production Context

**This RFC is grounded in real production challenges.** Our team, serving large-scale production inference traffic, has identified critical risks in EPP's current active-passive HA model:

1. **Single-point throughput ceiling**: the leader EPP cannot scale throughput by adding replicas, making it the performance bottleneck for the entire inference cluster.
2. **Long-context overload**: under high concurrency, the ability to handle long prompts is constrained, impacting SLAs for latency-sensitive workloads.
3. **Multi-replica state divergence**: shared scheduling state cannot be coordinated across replicas, causing significant degradation in routing quality that is unacceptable in production.

These issues have materially affected system availability and efficiency in our production environment. We are proposing this RFC in collaboration with the community and committing to contribute the implementation, with the goal of addressing this critical scalability gap for llm-d-router users.

## Motivation

In production, EPP sits on the critical path of inference requests. Every request flows through:

```
Gateway
  -> Envoy ext-proc
      -> EPP
          -> Model Server
```

As model service throughput grows, the request volume seen by EPP grows in tandem. However, the current HA mode is active-passive: multiple replicas may be deployed, but only the leader passes readiness checks and handles ext-proc traffic. This pattern provides failover, but:

[...]

PR #2024, opened 2026-07-15
[Draft]Shared Scheduling State for Horizontally Scalable EPP

nicole-lihui:
**What type of PR is this?**

/kind feature

Exploratory/draft PR for RFC #1593 — **not meant to be merged as-is**. This is the full
implementation from a feasibility spike, put up for the community to discuss and use as
a reference. If the direction gets a thumbs up, the plan is to split it into smaller PRs
that can be reviewed independently (roughly following the P0/P1/P2 progression already
in the commit history), with targeted optimization in each.

The core idea:
- Split "who serves traffic" (`stateless` EPP replicas, no leader election) from "who
  holds shared scheduling state" (a single `stateful` replica), talking to each other
  over an internal gRPC State API.
- Graceful degradation (FailOpen/LocalFallback) back to the local view whenever the
  stateful replica isn't reachable.

`--epp-mode` defaults to `classic`, so existing deployments that don't set the flag are
completely unaffected. Architecture write-up and steps to run it locally on kind are in
[docs/shared-scheduling-state.md](https://github.com/nicole-lihui/llm-d-router/blob/bb6e51bb4cd7a76f91f737a1c5e900982e01a14c/docs/shared-scheduling-state.md).

This round came with real load testing, not just design reasoning — numbers below are
from a single-node kind cluster with the vLLM simulator, **and will keep getting
updated**.

**Throughput / p99** (`classic-1repl` vs. 3 stateless + 1 stateful replica):

| Concurrency | Metric | classic-1repl | stateless+stateful | Δ |
|---|---|---|---|---|
| 25 | throughput | 324.2 req/s | 557.7 req/s | **+72.0%** |
| 25 | p99 | 137.3 ms | 96.0 ms | **-30.1%** |
| 100 | throughput | 388.1 req/s | 564.1 req/s | **+45.4%** |
| 100 | p99 | 442.6 ms | 467.9 ms | +5.7% (worse) |

Error rate remained 0.00% throughout.

The latest run shows a clear throughput improvement at both concurrency levels. 
The c=100 p99 difference is small (+5.7%) and within the expected benchmark variance.

An earlier run showed a larger c=100 p99 regression (+22.3%). Investigation found this
was caused by a bug in the concurrency-lease store sharding: sharding by requestID broke
the atomicity of the shared admission counter under load, allowing requests to exceed the
configured concurrency limit.

The issue was fixed with a lock-free CAS loop. After the fix, the p99 overhead directly
attributable to remote State API calls decreased from +28.8% to +3.7%.

Note: these results use shortened benchmark parameters (20s/2 repeats instead of the
default 60s/3). A full benchmark sweep is still needed for final numbers.
**Fault injection** (killed the stateful replica's Pod mid-load, 4 independent runs):

| Run | Pod delete → new Pod Ready | Error rate during the outage |
|---|---|---|
| 1 | 10.7s | 0.00% |
| 2 | 10.8s | 0.00% |
| 3 | 10.8s | 0.00% |
| 4 | 10.8s | 0.00% |

**Not measured yet**: how much shared prefix-cache benefit gets lost while the stateful
replica is restarting, and how long it takes to recover. Two dedicated experiments so far
didn't produce a clean signal (the prompt pool was too small relative to a single
replica's own request rate, so each replica covered the whole pool on its own traffic
alone and never actually needed the shared state) — redesigning that experiment is next.

**Questions where we'd want the community's input**:

- Is a single active-passive stateful replica (no sharding path) okay long-term, or is
  sharding a hard requirement before this can move past spike status?
- What eviction policy should the in-memory prefix/inflight store use (LRU? TTL?
  size-based?) — needs to be settled before a Redis/PVC-backed `Store`, and it'll shape
  the interface.
- Is plaintext gRPC (no TLS/mTLS) okay for this internal-only State API, or should mTLS
  be a prerequisite?
  
 
Relates to #1593

```release-note
NONE
```


nicole-lihui:
**This PR is big (44 files, ~6.2k lines) — most of that is generated code and tests.**
Here's where it's actually worth spending review time:

| Priority | File(s) | Why |
|---|---|---|
| Start here | `pkg/epp/statestore/store.go` | The core abstraction — `StateStore`/`InflightState`/`PrefixState`/`FlowControlState` interfaces and their contracts. Everything else implements or decorates this. |
| High | `pkg/epp/statestore/failopen.go`, `concurrency_lease.go` | The actual degradation logic: FailOpen/LocalFallback, async best-effort remote writes. Most of the interesting design decisions live here. |
| High | `pkg/epp/statestore/stateapi/memory_store.go` | The stateful replica's storage — sharded locks, concurrency-cap enforcement. Worth a close read: a lock-ordering bug got found and fixed here this round. |
| High | `cmd/epp/runner/runner.go` | Where `--epp-mode` actually changes behavior — the leader-election on/off switch, gRPC server/client wiring (including the reconnect backoff/keepalive config), admission controller assembly. |
| Medium | `pkg/epp/requestcontrol/admission.go`, `director.go` | Concurrency lease hooked into the existing admission and response-completion flow. |
| Medium | `pkg/epp/server/options.go` | New flags (`--epp-mode`, `--state-access-mode-*`, ...) and their validation. |
| Skip | `pkg/epp/statestore/stateapi/proto/gen/*.pb.go` (~1.9k lines) | Generated from `state_api.proto` — check the `.proto` instead if you care about the wire contract. |
| Just skim | `*_test.go` (most of the rest of the diff) | Big but mechanical; only worth reading closely if you want to double-check a specific claim below. |

cc @albertoperdomo2 @vMaroon @Lucas-Fernandes-Martins

elevran:
@nicole-lihui a few clarifying questions
- do you have an estimate on the load a single EPP can handle and the total load you expect for a single pool in your environment?
- are there similar run numbers for a set up with three active EPPs - without any state synchronization or coordination? I think this would be 3 statelss, 0 stateful using the terminology of this PR.
- any real HW benchmarks (IIUC, the test is using the simulator)?

nicole-lihui:
Thanks for the questions @elevran 

1. Single-EPP capacity: classic-1repl hits ~324 req/s 
 c=25 and ~388 req/s 
 c=100 in this simulator setup (see PR description). 
Our production RPS is comfortably under that, so raw EPP throughput isn't actually our driver — it's the restart/upgrade traffic-gap on a single active replica. (Token volume is a separate, model-serving-capacity question, not an EPP one.)

2. Yes — all state-access-mode forced to Local (zero coordination, "3 stateless, 0 stateful" in this PR's terms):

Concurrency | classic-1repl | stateless-local (no coordination) | stateless+stateful (coordinated)
-- | -- | -- | --
25 throughput | 324.2 req/s | 619.6 req/s | 557.7 req/s
25 p99 | 137.3 ms | 89.3 ms | 96.0 ms
100 throughput | 388.1 req/s | 574.6 req/s | 564.1 req/s
100 p99 | 442.6 ms | 451.2 ms | 467.9 ms

zero coordination performs on par with or slightly better than coordinated here. 

Caveat: 
this only measures request-handling speed, not routing quality. 
Zero-coordination replicas each see only their own routed requests, so prefix-affinity and load-balancing decisions run on a fragmented view — that cost would show up downstream (extra prefill, uneven backend load), not in these numbers.
 Haven't measured it yet, and I'd weigh it at least as heavily as the throughput numbers.


3. No real HW yet — resource constraint on our end, not by design. 
All numbers above are single-node kind + vLLM simulator.

One more caveat: 
EPP's CPU/memory cost scales with input token length too, not just request rate — prefix-matching work grows with prompt length. 
So these numbers reflect the simulator's default prompt length; if production inputs run longer, the real ceiling is probably more conservative. Read these as relative comparisons between configs, not an absolute capacity bound.

PR #2026, opened 2026-07-15
flowcontrol: add explicit domain to priority-holdback-policy

### LINKED ISSUE #2023: [Flow Control] Explicit ceilings policy for priority holdback

Add a new `domain` value -- `"explicit"` -- to the existing `priority-holdback-policy` plugin. When `domain` is set to `"explicit"`, the operator provides a map of priority levels to ceiling values directly, bypassing algorithmic interpolation.

#### Configuration

```yaml
plugins:
  - type: priority-holdback-policy
    name: my-explicit-holdback
    parameters:
      domain: explicit
      ceilings:
        100: 0.95
        50: 0.70
        10: 0.30
```

- `ceilings` (map[int]float64, required when `domain: explicit`): maps each priority level to its admission ceiling in `[0.0, 1.0]`.
- `shape`, `minCeiling`, `maxCeiling` are unused (and rejected) when `domain: explicit`.

#### Behavior

- On each dispatch cycle, `ComputeLimit` receives the current set of active priorities. For each priority in the input, the policy looks up its ceiling from the configured map.
- If an active priority is **not present** in the map, the request is rejected at config load time or a default/fallback strategy applies.
- Ceilings must be monotonically non-increasing when priorities are sorted highest-first (the `UsageLimitPolicy` contract). Validation enforces this at config load time.

**Why is this needed**:

The `priority-holdback-policy` plugin computes admission ceilings algorithmically from a `shape`/`domain` pair and a `[minCeiling, maxCeiling]` range. The existing domains (`rank` and `value`)  works well when the operator wants evenly distributed or proportionally spaced ceilings, but it does not cover scenarios where specific, hand-tuned ceiling values are required for each priority band.

ishwar170695:
## What type of PR is this?

/kind feature

## What this PR does / why we need it

This PR implements the new `explicit` domain for the `priority-holdback-policy` plugin, allowing operators to configure admission ceilings directly instead of relying on algorithmically generated values.

### Changes

* Add `domainExplicit` and a `ceilings map[int]float64` configuration field.
* Validate `explicit` configurations by:

  * requiring `ceilings`,
  * rejecting `shape`, `minCeiling`, and `maxCeiling`,
  * ensuring ceiling values are within `[0.0, 1.0]`,
  * enforcing monotonically non-increasing ceilings when priorities are sorted highest-first.
* Add `computeLimitExplicit`, which returns configured ceilings for each priority. Missing priorities default to `0.0`, consistent with the `UsageLimitPolicy` contract.
* Preserve the existing single-priority optimization for algorithmic domains while disabling it for `explicit` via an `enableSinglePriorityBypass` flag so configured ceiling values are always respected.
* Add unit and integration tests covering configuration validation, explicit ceiling computation, and `ComputeLimit` behavior.
* Document the new `explicit` domain in the plugin README with configuration examples and validation rules.

## Which issue(s) this PR fixes

Fixes #2023

## Release note

```release-note
Adds support for the `explicit` domain in the priority-holdback-policy plugin, allowing operators to configure priority admission ceilings directly.
```


loicmarchal:
Thanks @ishwar170695, this looks good.
Just one discussion about the default value, see https://github.com/llm-d/llm-d-router/issues/2023#issuecomment-4996890869 (cc @LukeAVanDrie)

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/config.go
```diff
@@ -49,26 +52,36 @@ type apiConfig struct {
...
 	// A value of 1.0 means the highest priority is only gated at full saturation.
 	//
 	// Defaults to 1.0 if unset. Must be in (0.0, 1.0] and strictly greater than MinCeiling.
+	// Unused and rejected when domain is "explicit".
 	MaxCeiling *float64 `json:"maxCeiling,omitempty"`
+
+	// Ceilings maps each priority level to its admission ceiling in [0.0, 1.0].
+	//
+	// Required when domain is "explicit". Ceilings must be monotonically non-increasing
+	// when priorities are sorted highest-first. Unused when domain is not "explicit".
```
loicmarchal: should also be rejected when domain is not "explicit"

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/config.go
```diff
@@ -98,16 +112,29 @@ func buildConfig(apiCfg *apiConfig) (*config, error) {
...
-		shape:      *safeCfg.Shape,
-		domain:     *safeCfg.Domain,
-		minCeiling: *safeCfg.MinCeiling,
-		maxCeiling: *safeCfg.MaxCeiling,
-	}, nil
+	cfg := &config{
+		domain: *safeCfg.Domain,
+	}
+	if *safeCfg.Domain == domainExplicit {
+		cfg.shape = defaultShape
```
loicmarchal: we don't expect a shape if domain is `explicit`, this will set it to `linear`

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/priority_holdback.go
```diff
@@ -144,3 +161,15 @@ func computeLimitLinearProportional(cMin, cMax float64, priorities []int) (ceili
 	}
 	return ceilings
 }
+
+// computeLimitExplicit looks up each priority in the configured map and returns the ceiling slice
+// in the same order as the input. Missing entries return 0.0, consistent with the UsageLimitPolicy
```
loicmarchal: I'm still not sure missing entries should return 0.0 (see discussion in https://github.com/llm-d/llm-d-router/issues/2023#issuecomment-4988058254)

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/README.md
```diff
@@ -43,16 +43,31 @@ Scales ceilings proportionally to the numerical priority value within the observ
...
+Use when hand-tuned, per-band ceiling values are required and the algorithmic domains do not produce the desired distribution.
+
+**Constraints:**
+- `ceilings` (map of integer priority to float64) is required.
+- Each ceiling must be in `[0.0, 1.0]`.
+- Ceilings must be monotonically non-increasing when priorities are sorted highest-first.
+- `shape`, `minCeiling`, and `maxCeiling` must not be set; they are rejected at config load time.
+
+When the policy receives an active priority with no entry in the map (priorities are dynamic at runtime), it returns `0.0` for that priority, consistent with the `UsageLimitPolicy` contract that `0.0` represents a fully gated priority.
+
```
loicmarchal, line 59: same comment as above about missing ceilings

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/priority_holdback.go
LukeAVanDrie: +1

Since the registry always provisions a `DefaultPriorityBand`, an unlisted priority isn't a theoretical case: it gets a dynamically provisioned band that this line then silently and permanently gates, with no signal to the operator. Whatever we decide about *where* the explicit-vs-dynamic-provisioning validation lives, I don't think we should ship the silent `0.0` default (or document it as "consistent with the contract"; the contract defines what `0.0` means, it doesn't make it a safe fallback).

The doc comment, `TestComputeLimitExplicit_MissingEntryReturnsZero`, and the README paragraph would all need to change together.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/config.go
LukeAVanDrie: +1

`ceilings` set alongside `domain: rank`/`value` is currently accepted and silently ignored. Since the PR strictly rejects `shape`/`minCeiling`/`maxCeiling` under `explicit`, the reverse should hold too.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/config.go
LukeAVanDrie: +1 

Suggest leaving the zero value:

```go
	if *safeCfg.Domain == domainExplicit {
		cfg.ceilings = safeCfg.Ceilings
	} else { ... }
```

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/priority_holdback.go
```diff
@@ -56,25 +57,40 @@ type priorityHoldbackPolicy struct {
 	name      string
 	cMax      float64
 	cMin      float64
```
LukeAVanDrie: Non-blocking nit: after moving to closure-captured config, `cMin` is assigned in the constructor but never read (only `cMax` survives, for the single-priority bypass). The field can be dropped.

LukeAVanDrie (changes requested):
Thanks @ishwar170695! Agree with all @loicmarchal's feedback here. Outside of that, I just had a couple nits.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/config.go
ishwar170695: Done. Added 
```
if len(cfg.Ceilings) > 0 {
		errs = append(errs, errors.New("ceilings must not be set when domain is not \"explicit\""))
	}
```
in validateConfig function.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/config.go
ishwar170695: Removed `cfg.shape = defaultShape`

ishwar170695:
Thanks for the reviews! I've addressed the other comments (rejecting ceilings for non-explicit domains, removing the default shape, removing the unused cMin, and updating the docs/tests).

I have one question about the remaining discussion around missing explicit ceilings.

From tracing the current flow, it looks like the registry always provisions `DefaultPriorityBand`/`DefaultNegativePriorityBand`, so an explicit ceilings map may not cover every active priority.

Would you prefer this PR to also enforce that explicit cannot be used with dynamically provisioned priority bands (i.e. validate at startup that every active priority has an explicit ceiling), or would you rather keep that as a follow-up?

I wanted to clarify the intended architecture before introducing additional validation or changing the runtime behavior for missing entries.

loicmarchal:
> Would you prefer this PR to also enforce that explicit cannot be used with dynamically provisioned priority bands (i.e. validate at startup that every active priority has an explicit ceiling), or would you rather keep that as a follow-up?

_TL;DR: I think we should enforce both that `explicit` holdback policy and dynamic band provisioning cannot be used together, and that every active priority band has a defined ceiling, at validation as part of this PR._

I think we should enforce that `explicit` holdback policy and dynamic band provisioning cannot be used together at validation as part of this PR. Otherwise any band created dynamically would immediately be gated with the default `0.0` (and no particular default value can ensure monotonicity). This is like silently enforcing that both can't be used together. I think it's better to be explicit about it.

For the validation that every active priority band has a defined ceiling, I would also lean towards rejecting at validation if active bands are missing ceilings. Otherwise, the config semantic would be "we are defining priority bands that we can choose to fully gate from the beginning by not defining any associated ceilings". It feels a bit more like a misconfiguration.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/priority_holdback.go
```diff
@@ -144,3 +159,12 @@ func computeLimitLinearProportional(cMin, cMax float64, priorities []int) (ceili
 	}
 	return ceilings
 }
+
+// computeLimitExplicit looks up each configured priority ceiling.
```
loicmarchal, line 190: nit: `computeLimitExplicit` would still return `0.0` for priority-ceiling pairs not defined in the ceilings map. I think that's ok if we ensure this case can't happen at the validation stage. I would still add in the function doc as a precondition that the set of keys of `ceilings` must be equal to the set of priorities in `priorities` (but still returns `0.0` if it's not).

ishwar170695:
> TL;DR: I think we should enforce both that `explicit` holdback policy and dynamic band provisioning cannot be used together, and that every active priority band has a defined ceiling, at validation as part of this PR.

I'll implement the validation in this PR so that `explicit `cannot be used with dynamically provisioned priority bands and every active priority band must have a configured ceiling. I'll also update the `computeLimitExplicit `documentation to state that completeness is guaranteed by validation.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/config.go
loicmarchal: Thank you @ishwar170695.
nit: also update the doc: Unused **and rejected** when domain is not "explicit".

ishwar170695: @loicmarchal done.

ishwar170695:
Implemented the requested validation and documentation updates:

* Rejects `explicit` domain when dynamic priority band provisioning (`DefaultPriorityBand` / `DefaultNegativePriorityBand`) is configured.
* Validates that every active priority band (including priority `0`) has a configured ceiling during config loading.
* Updated `computeLimitExplicit` documentation to state that complete ceiling coverage is guaranteed by validation, while preserving the existing `0.0` fallback only as a defensive behavior.
* Added loader validation tests covering the new validation paths.

pkg/epp/config/loader/flowcontrol.go
```diff
@@ -59,6 +66,27 @@ func buildFlowControlConfig(
 		return nil, fmt.Errorf("failed to resolve usage limit policy: %w", err)
 	}
 
+	if phPolicy, ok := usageLimitPolicy.(ExplicitPolicyInfo); ok && phPolicy.Domain() == priorityholdback.DomainExplicit {
```
loicmarchal: I'm hesitant with this approach. I think it might complicate the validation and it kind of leaks internal implementation details from the usage limit policy (especially the holdback policy in this case, such as ceilings map)

loicmarchal: Maybe we could look at a solution with a validator in the `UsageLimitPolicy`? For example have a `ConfigValidator` interface with a `ValidateConfig()` function that we can implement in `priorityHoldbackPolicy`.

Then have here something like:
```golang
if err := usageLimitPolicy.ValidateConfig(info); err != nil {
    return nil, fmt.Errorf("usage limit policy config validation failed: %w", err)
}
```

For the `info` parameter, we could have some struct generic enough to be reusable with any usage limit policy if needed. So something like this in the `flowcontrol` package (`/pkg/epp/framework/interface/flowcontrol/flow.go`):

```golang
type ConfigValidator interface {
    ValidateConfig(info ConfigInfo) error
}

type ConfigInfo struct {
    // StaticPriorities lists the priority values from explicitly configured bands.
    StaticPriorities []int
    // DynamicProvisioningEnabled is true when default band templates are configured.
    DynamicProvisioningEnabled bool
}
```

And add `ConfigValidator` in the `UsageLimitPolicy` interface.

The implementation of `ValidateConfig` then becomes:

```golang
func (p *priorityHoldbackPolicy) ValidateConfig(info flowcontrol.ConfigInfo) error {
    if p.cfg.domain != domainExplicit {
        return nil
    }

    if info.DynamicProvisioningEnabled {
        return fmt.Errorf(
            "priority holdback policy with explicit domain cannot be used with dynamic priority band provisioning",
        )
    }

    for _, prio := range info.StaticPriorities {
        if _, ok := p.cfg.ceilings[prio]; !ok {
            return fmt.Errorf("priority band %d has no configured ceiling in explicit domain", prio)
        }
    }

    return nil
}
```

loicmarchal: If we don't want to modify the `UsageLimitPolicy` interface, we could not add `ConfigValidator` in the `UsageLimitPolicy` interface, and have something like:
```golang
f v, ok := usageLimitPolicy.(ConfigValidator); ok {
    ...
    v.ValidateConfig(info)
}
```

loicmarchal: @LukeAVanDrie any thoughts?

LukeAVanDrie: Agree with the `ConfigValidator` direction. 

+1 to the type assertion approach. If it generalizes, we can always promote it into the `UsageLimitPolicy` interface later. II think we can define `ConfigValidator`/`ConfigInfo` in `pkg/epp/framework/interface/flowcontrol` next to `UsageLimitPolicy`.

Just make sure to derive `ConfigInfo` from the resolved `registryConfig` rather than the raw `apiConfig`. The
 registry config already contains the final static band set, including the always-provisioned band 0. `StaticPriorities` becomes the keys of `registryConfig.PriorityBands`.  This removes the hardcoded `ceilings[0]` checks, the separate `apiConfig == nil` branch, and any possibility of the validation drifting from what the registry actually provisions.

LukeAVanDrie: @loicmarchal One complication for `DynamicProvisioningEnabled`. Currently, that boolean can't be `false`. `apiConfig.DefaultPriorityBand`/`DefaultNegativePriorityBand` only customize the template. `registry.NewConfig` auto-initializes `DefaultPriorityBand` when unset, and provisioning is driven unconditionally by the `InferenceObjective` reconciler (`SubmitDesiredPriorities` -> `ApplyDesiredPriorities` -> `provisionPriorityBandLocked`).

I think the current mutual-exclusion check validates the wrong thing. Rejecting the config when `apiConfig.DefaultPriorityBand != nil` doesn't prevent dynamic provisioning.

E.g.,

With `domain: explicit`, no templates configured, and ceilings for bands {0, 100}, this validation passes. If we then apply an `InferenceObjective` with priority 50, it provisions band 50 from the auto-initialized template, and it is silently gated at 0.0.

What do you think of these options?

1. Enforce it at the provisioning seam: when the usage-limit policy constrains the priority domain, `ApplyDesiredPriorities` (or the processor) refuses to provision an unlisted band and logs an error. That's the only place the invariant can actually be held today.
2. Keep config-time validation of the static bands only (which this PR does well), drop the template check, and make the runtime gating loud: `logger.Error` + metric when `computeLimitExplicit` hits a missing priority, plus a README paragraph.
3. Add a real "disable dynamic band provisioning" config fact honored by the registry/reconciler path . This is probably the cleanest semantics, but a bigger change that probably deserves its own issue.

LukeAVanDrie: Maybe we can look into option 3 after the next release? I don't think it's worth the time investment upfront.

loicmarchal: @LukeAVanDrie Good catch, you're right!

+1 for options 1 and 2 and to look into option 3 later

loicmarchal:
@ishwar170695 have you had a chance to look at the last changes? We would like to ship this policy as its needed to use flow control with batch-gateway. Happy to help look at the last changes if you don't have free cycles at the moment.

ishwar170695:
Hi @loicmarchal, thanks for the ping, and sorry for the delay.

I've addressed the latest review feedback and pushed an update.

The validation now uses a `ConfigValidator` with a generic `ConfigInfo`, so the loader no longer contains policy-specific validation logic. I also refactored the tests so the loader only verifies delegation/error propagation, while the priority-holdback policy owns the validation-specific test cases.

I also dropped the dynamic provisioning validation, following the discussion around keeping the larger runtime semantics for a follow-up, so this PR focuses on the static priority validation and the `ConfigValidator` refactor.

I'd appreciate another look when you have a chance. Thanks again for the design suggestions, they helped make the validation flow much cleaner.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/priority_holdback.go
```diff
@@ -144,3 +194,21 @@ func computeLimitLinearProportional(cMin, cMax float64, priorities []int) (ceili
...
+// configured priority band has a corresponding ceiling entry. Dynamically
+// provisioned bands might fall back to 0.0 with an error log.
+func computeLimitExplicit(ctx context.Context, ceilings map[int]float64, priorities []int) []float64 {
+	result := make([]float64, len(priorities))
+	logger := log.FromContext(ctx)
+	for i, p := range priorities {
+		if c, ok := ceilings[p]; ok {
+			result[i] = c
+		} else {
+			logger.Error(nil, "Missing explicit ceiling for dynamically provisioned priority band", "priority", p)
```
loicmarchal: nit: I think we should just be generic and log `missing explicit ceiling for priority band %d, ceiling has been set to 0.0`. Although we know from config validation that a missing ceiling should only come from dynamically provisioned priority bands, the contract (pre-condition) of this function is that it will return `0.0` and log an error if a priority has no ceiling. Also this log could give the impression that we can actually define an explicit ceiling for a dynamically provisioned priority band, which is not possible.

pkg/epp/framework/plugins/flowcontrol/usagelimits/priorityholdback/priority_holdback.go
```diff
@@ -144,3 +194,21 @@ func computeLimitLinearProportional(cMin, cMax float64, priorities []int) (ceili
 	}
 	return ceilings
 }
+
+// computeLimitExplicit looks up each configured priority ceiling.
+// Precondition: configuration validation guarantees that every statically
+// configured priority band has a corresponding ceiling entry. Dynamically
+// provisioned bands might fall back to 0.0 with an error log.
```
loicmarchal: I think the precondition should not mention anything about config validation. It just states what we expect as input (meaning we don't guarantee consistent result if the input doesn't respect the precondition, regardless of how the input was created). For example, changing how the config is validated in the future should have no impact on the function precondition.

In this case we could just state something like `Every priority in priority is expected to have a corresponding ceiling in ceilings. Priorities with missing ceilings will receive a ceiling value of 0.0`

loicmarchal:
Thank you @ishwar170695! The new validation looks good. Just some nit.

Should we also look into refusing to provision an unlisted band and logs an error in `ApplyDesiredPriorities`?
We should also add a metric and some paragraph in the README to mention the incompatibility between explicit holdback policy and dynamic band provisioning.

ishwar170695:
Thanks for the feedback! I've addressed the remaining comments in the latest update:

* Made the missing-ceiling log message generic.
* Updated the `computeLimitExplicit` doc comment to describe the function precondition directly, without referencing configuration validation.
* Updated the README to document the explicit domain limitation and fallback behavior, and clarified the configuration validation wording.

I'd appreciate another look when you have a chance. Thanks!

ishwar170695:
> Should we also look into refusing to provision an unlisted band and logs an error in `ApplyDesiredPriorities`?

I looked into `ApplyDesiredPriorities`, and in the current implementation it provisions bands unconditionally without an error path, while the registry remains unaware of the configured usage limit policy. That made it feel like a broader behavioral change than the validation refactor. 
I was also trying to keep the validation within the policy, following the `ConfigValidator `direction. If you think refusing provisioning in the registry is the preferred design, I'm happy to extend this PR accordingly.

loicmarchal:
Thank you @ishwar170695! The last changes look good.

>I looked into ApplyDesiredPriorities, and in the current implementation it provisions bands unconditionally without an error path, while the registry remains unaware of the configured usage limit policy. 

One option would be to add some information in the registry `Config`. Something like 
```golang
type Config struct {
  // ...
  dynamicPrioritiesAllowed bool
  // ...
}
```
If it doesn't find the priority (`if _, ok := fr.config.PriorityBands[priority]; !ok `) it checks `dynamicPrioritiesAllowed` first and if `false`, it logs an error and return, otherwise it provisions the priority. @LukeAVanDrie what do you think?

ishwar170695:
> One option would be to add some information in the registry `Config`. Something like
> 
> ```go
> type Config struct {
>   // ...
>   dynamicPrioritiesAllowed bool
>   // ...
> }
> ```
> 
> If it doesn't find the priority (`if _, ok := fr.config.PriorityBands[priority]; !ok `) it checks `dynamicPrioritiesAllowed` first and if `false`, it logs an error and return, otherwise it provisions the priority

Thanks! I've implemented this approach and pushed an update. The registry now refuses to provision unknown priority bands when dynamic priority provisioning is disabled, while preserving the default behavior for existing users. I'd appreciate another look when you have a chance.

ishwar170695:
@LukeAVanDrie @loicmarchal the `lint` and `test` failures were caused by a stale merge conflict in `registry_test.go` — a `/*` comment was left from an earlier merge that broke the Go parser. The conflict has been resolved: the branch now contains both the `TestApplyDesiredPriorities_*` tests and the upstream `countSeriesWithFairnessID` / `TestFlowRegistry_GarbageCollection_PrunesMetricSeries` additions with their required imports. 

Could you approve the ci workflow runs? Thanks!

apix/config/v1alpha1/endpointpickerconfig_types.go
```diff
@@ -420,6 +420,12 @@ type FlowControlConfig struct {
 	// SaturationDetector specifies which saturation detector plugin to use for both Admission and
 	// Flow Control. If omitted, "utilization-detector" is used by default.
 	SaturationDetector *SaturationDetectorConfig `json:"saturationDetector,omitempty"`
+
+	// +optional
+	// AllowDynamicPriorityProvisioning controls whether the registry creates new priority bands
+	// dynamically when requested by the control plane during reconciliation.
+	// Defaults to true when omitted.
+	AllowDynamicPriorityProvisioning *bool `json:"allowDynamicPriorityProvisioning,omitempty"`
```
ahg-g, line 428: I don't support adding this configuration, it is just not compatible with the fact that we have an object api to add priorities at runtime. The new feature being added in this PR should adapt to this fact.

ishwar170695:
@ahg-g @loicmarchal @LukeAVanDrie Thanks for the feedback. I want to make sure we're aligned on the intended direction before reworking this again.

The `AllowDynamicPriorityProvisioning` gate was added to address the earlier concern in this thread about explicit ceilings silently gating dynamically provisioned bands. I'm happy to rework it if there's a better fit with the object API.

@ahg-g, thanks for pointing out the interaction with the object API. I hadn't considered that implication. If I'm understanding correctly, the concern is that a static configuration gate isn't compatible with adding priorities at runtime, and that the explicit-domain feature should instead adapt to that model.

Could you align on which direction you'd prefer?

* keep the registry config gate,
* enforce the constraint at the provisioning/object API path instead,
* or another approach.

Once there's consensus, I'll update the implementation accordingly.

apix/config/v1alpha1/endpointpickerconfig_types.go
loicmarchal: @ahg-g I understand your point of view. I think the best solution in this case should be to not add this configuration and get a ceiling interpolated linearly from the two nearest configured neighbors for any dynamically provisioned priority.

We could define ceilings explicitly in the configuration for any chosen priority. Any missing ceilings (either because it wasn't define in the configuration for a statically configured priority or because it's a dynamically provisioned priority) gets calculated as a linear interpolation between the configured neighbors ceilings. What do you think? cc @LukeAVanDrie

ishwar170695:
@ahg-g @loicmarchal @LukeAVanDrie

To keep this moving, I'll remove `AllowDynamicPriorityProvisioning` and implement linear neighbor interpolation for missing priorities.

* Between two configured ceilings: linearly interpolate.
* Outside the configured range: clamp to the nearest boundary.
* Empty `ceilings`: already rejected by validation.

One remaining edge case: if only one ceiling is configured, should all unlisted/dynamic priorities use that same ceiling?

loicmarchal:
Thanks @ishwar170695, that plan matches what I had in mind for an anchored approach.

On the single-ceiling case: yes, every unlisted priority should get that same ceiling. It falls out of the clamp rule you already described — with one anchor, every other priority is outside the configured range on one side or the other. No special case needed, and it's the only option that keeps the ceiling sequence monotonically non-increasing: a single point defines no slope, and picking an implicit second anchor would bring back `minCeiling`/`maxCeiling` semantics that `explicit` rejects. The configuration then means a uniform admission ceiling with no holdback differentiation, which seems like a legitimate thing for an operator to ask for.

Two points to settle with it:

- Interpolate over priority value, not rank. Rank makes a band's ceiling depend on which other bands happen to be provisioned. If a band is dynamically provisioned, it will change the ceiling for all the other dynamic bands between two anchors. Once a dynamic band is garbage collected, it will change the ceiling values again between the same two anchors. `AllOrderedPriorityLevels()` is recomputed every dispatch cycle and `PriorityBandGCTimeout` removes idle dynamic bands, so that situation could happen. Interpolating on value makes the config a total function from priority to ceiling: a dynamically provisioned band of a specific vlaue will always get the same ceiling, determinable from the YAML alone. Worth stating in the doc comment, since this plugin already uses both words as domain names.
- With interpolation defined for missing entries, I think the `ValidateConfig` check that rejects static priorities without a ceiling should go too. Otherwise a missing static priority is a startup error while a missing dynamic one is interpolated, and there's no principled difference between them. That leaves `ValidateConfig` with nothing to do for this policy, so the `ConfigValidator`/`ConfigInfo` interface and the loader plumbing can come out as well. Sorry this reverses what I asked for earlier in the thread; the interpolation semantics make the coverage requirement unnecessary. Validation would keep: non-empty ceilings, values in [0.0, 1.0], monotonically non-increasing by descending priority.

We could also consider using `minCeiling`/`maxCeiling` to add out-of-range bounds for priorities above the highest and below the lowest configured entry, but that should probably be a separate future concern if we ever need it. For now, we should keep rejecting `shape`/`minCeiling`/`maxCeiling` under `explicit`. Using those fields in the future would be possible without a breaking change.

Also, `computeLimitExplicit` can no longer hit the `0.0` fallback, so the precondition doc, the error log, the test, and the README paragraph should be updated together.

@ahg-g @LukeAVanDrie does this work for you?

PR #2032, opened 2026-07-15
feat(epp): load-aware p2p source selection

### LINKED ISSUE #2031: p2p-source-producer: source selection is load-blind and concentrates pull traffic on one peer
The producer picks the pull source by strict argmax over cached tokens. Two consequences:

1. Equal counts lose to iteration order, so every consumer of a widely-replicated prefix pulls from the first matching endpoint in list order - the normal state for any hot prefix after fleet acquisition - while equally-cached peers sit idle.
2. Load never factors in, although a source serves pulls from its engine-step loop, so the busiest peer is also the slowest to serve a pull.

A hard minimum-queue tie-break is not enough: endpoint metrics refresh about once per scrape interval (~1s), so a stale one-request queue difference would herd a whole window of requests onto whichever peer last read the shortest queue.

Proposal: among the max-cached peers, sample the source with probability proportional to `1/(1+waiting queue)`, using a request-ID hash as the deterministic sampling coordinate. Exact ties spread uniformly, a one-request queue difference shifts share about 2:1 instead of starving anyone, and deeply-queued sources are effectively avoided. A strictly better-cached peer still wins regardless of load, so `minCachedTokenDelta` semantics are unchanged. No new configuration.

Same spirit as `max-score-picker`'s shuffle-before-sort (ties are the common case; iteration order is the wrong tie-break) and `weighted-random-picker`'s proportional sampling, adapted to the data-producer phase where the choice must be deterministic per request.

nilig:
**What type of PR is this?**
/kind feature

The `p2p-source-producer` picks the pull source by strict argmax over cached tokens, which is load-blind in two ways: equal counts lose to iteration order, so every consumer of a widely-replicated prefix pulls from the same peer while equally-cached peers sit idle; and a deeply-queued source is chosen over an idle one, although sources serve pulls from their engine-step loop and respond slower under load.

Sample the source among max-cached peers with probability proportional to `1/(1+waiting queue)`, using a request-ID hash as the deterministic sampling coordinate. Exact ties spread uniformly, a one-request queue difference (noise at metrics-scrape granularity) shifts share about 2:1 instead of herding the window, and deeply-queued sources are effectively avoided. A strictly better-cached peer still wins regardless of load, so `minCachedTokenDelta` semantics are unchanged. No new configuration. Same spirit as `max-score-picker`'s shuffle-before-sort and `weighted-random-picker`'s proportional sampling, adapted to the data-producer phase where the choice must be deterministic per request.

Cloud validation (16x gpt-oss-120b, H200, hot 16K prefix replicated fleet-wide, 200 requests at 8 req/s through the gateway, source distribution from routing-sidecar `running P2P source protocol` logs):

| | before (argmax) | after (this PR) |
|---|---|---|
| header-driven pulls | 154 | 162 |
| distinct sources used | 8 | 12 |
| top-1 source share | 82% | 29% |
| client TTFT p50 | 0.071s | 0.069s |

**Test plan**
- [x] Unit: tie spread by request hash is uniform across tied peers and deterministic per request
- [x] Unit: share ordering follows inverse queue weights; a one-request queue difference does not starve the deeper peer
- [x] Unit: a strictly better-cached peer wins regardless of load; nil metrics preserve delta-only behavior
- [x] Cloud: pull-source concentration on a replicated hot prefix drops from 82% top-1 to 29% with no client latency change

Fixes #2031

```release-note
The p2p-source-producer spreads KV pull traffic across equally-cached source peers proportionally to their load instead of always selecting the first, avoiding pull hot-spots on widely-replicated prefixes.
```


pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/producer.go
```diff
@@ -142,20 +143,54 @@ func (p *Producer) attrKey() string {
...
+		total := 0.0
+		for _, ep := range endpoints {
+			if ep.GetMetadata() == nil || p.cachedTokenCount(ep) != maxCached {
+				continue
+			}
+			w := 1.0 / (1.0 + float64(waitingQueueSize(ep)))
+			candidates = append(candidates, ep)
+			weights = append(weights, w)
+			total += w
+		}
```
shmuelk: Waiting queue size is only being used as a tie breaker. I don't see that prevents all of the requests from going to a single endpoint.

nilig: You're right - the queue weight only arbitrated exact ties, so a peer one block ahead took every pull regardless of load. Widened the sampling to a one-block band in 2aec3558: a peer within one block of the maximum now competes on `1/(1+waiting queue)`, since the one-block shortfall costs the destination a single block of recompute, noise next to a queue-depth difference on the source. Beyond one block the better-cached peer still wins - spreading there trades real pull benefit for recompute. Concentration on a sole holder of a prefix remains, and I think that's correct: it's the only source, and it's self-limiting - pods that pull become equal-count sources once their blocks are indexed, so the spread takes over after the first wave.

pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/producer.go
```diff
@@ -142,20 +145,60 @@ func (p *Producer) attrKey() string {
...
+// the moment one replica drifts a block ahead: a peer one block short costs
+// the destination a single block of recompute, noise next to a queue-depth
+// difference on the source. Proportional weights spread ties uniformly,
+// mildly prefer a one-request-shorter queue, and starve deeply-queued
+// sources. No-op when no candidate holds any cached block.
 func (p *Producer) Produce(ctx context.Context, request *scheduling.InferenceRequest, endpoints []scheduling.Endpoint) error {
-	best := bestMatchPeer{}
+	maxCached := 0
 	for _, ep := range endpoints {
-		md := ep.GetMetadata()
```
roytman, line 151: This loop doesn't check `ep.GetMetadata() == nil`, but the candidate-filtering loop right after it does (`ep.GetMetadata() == nil || cached == 0 || cached+blockSize < maxCached`). Concretely: if an endpoint with `nil` metadata reports the pool's highest cached-token count, `maxCached` gets pinned to that value, that endpoint itself is then excluded from candidates for lacking metadata, and every real endpoint whose count falls more than one block short of that inflated `maxCached `is excluded too. If no real endpoint survives, candidates is empty, `best.cachedTokens` stays 0, and the header is silently never set — even though a real out-caching peer existed.

The old code avoided this because the metadata check and the max-tracking happened in the same loop, so a metadata-less endpoint could never influence best. The repo already treats metadata-nil endpoints as a case worth defending against (see the existing TestProduce_MissingMatchInfo_TreatedAsZero test with a bare endpoint), so this isn't purely theoretical, though in production the endpoint list comes from the real pod datastore, so it's unlikely to actually lack metadata today.

Suggestion: skip metadata-nil endpoints in the `maxCached` loop too, mirroring the second loop.

pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/producer.go
```diff
@@ -142,20 +145,60 @@ func (p *Producer) attrKey() string {
...
-		if md == nil {
-			continue
+		if cached := p.cachedTokenCount(ep); cached > maxCached {
```
roytman: `cachedTokens` is computed twice per endpoint. The first call, is here via `cachedTokenCount`, and the second one is in the L:178.
Cheap, and not worth restructuring on its own — flagged only because fixing the metadata-nil gap from before would naturally unify the two loops and remove this too.

pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/producer.go
nilig: Fixed in 9eafb141. Metadata-less endpoints are excluded before computing `maxCached`, and the usable endpoints' cached counts and block sizes are retained so `cachedTokens()` is evaluated once per endpoint. Added regression coverage for a metadata-less endpoint pinning the maximum and suppressing a valid source.

pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/producer.go
nilig: The unification in 9eafb141 removes this too: `cachedTokens()` is evaluated once per endpoint in the collection pass.

elevran (approved):

PR #2036, opened 2026-07-15
refactor(flowcontrol): remove the queue plugin registry

### LINKED ISSUE #2021: flowcontrol: remove the vestigial per-band queue selection

Remove the queue-selection surface from flow control, left over from when multiple queue implementations existed:

- The `PriorityBandConfig.Queue` field, the `WithQueue` config option, and the `defaultQueue` default.
- The string-keyed queue registry/factory (`MustRegisterQueue`, `NewQueueFromName`, `RegisteredQueueName`).

The priority queue would be constructed directly at flow provisioning, making the queue a pure internal detail.

**Why is this needed**:

After unifying flow control on a single priority queue (#1831), there is only one queue implementation, so a configurable per-band queue name selects nothing and is misleading.  This was surfaced in review of #1831.

LukeAVanDrie:
**What type of PR is this?**

/kind cleanup

With flow control unified on a single priority queue (#1831), the queue-selection machinery no longer selects anything. This removes it:

- Delete the string-keyed queue registry/factory (`MustRegisterQueue`, `NewQueueFromName`, `RegisteredQueueName`, the `init()` self-registration) and construct the priority queue directly via `queue.New`.
- Remove the per-band `Queue` config field, the `WithQueue` option, and the `defaultQueue` default. Queues are now a pure internal detail, not user-selectable.
- Remove the vestigial `Name()` from the queue contract; its only consumer was a `queueType` log field that is constant given one implementation.
- Consolidate the queue tests into `priorityqueue_test.go` (table-driven ordering, focused `TestPriorityQueue_*` functions); delete `functional_test.go`.
- Drop an unreachable nil-`OrderingPolicy` guard in `buildFlowComponents`; the provisioning-failure tests now trigger the real `ErrPriorityBandNotFound`, which exercises the same lease-rollback and error-scoping paths.

  Fixes #2021

**Release note**:

```release-note
NONE
```


shmuelk:
Nice straight forward cleanup.

Two nits:
  1. As the PriorityQueue is no longer a plugin, it should probably be moved to be with the rest of the flow control code under pkg/epp/flowcontrol.
  2. The name PriorityQueue is somewhat confusing as flow control has priorityBands. The PriorityQueue queues based on the OrderingPolicy. Perhaps naming it OrderedQueue or RankedQueue (then I would change the OrderingPolicy to RankingPolicy as well) ?
 
I am happy for these nits to be dealt with in a follow up PR. @LukeAVanDrie your choice.

LukeAVanDrie:
> Two nits:
> 
> 1. As the PriorityQueue is no longer a plugin, it should probably be moved to be with the rest of the flow control code under pkg/epp/flowcontrol.
> 2. The name PriorityQueue is somewhat confusing as flow control has priorityBands. The PriorityQueue queues based on the OrderingPolicy. Perhaps naming it OrderedQueue or RankedQueue (then I would change the OrderingPolicy to RankingPolicy as well) ?
> 
> I am happy for these nits to be dealt with in a follow up PR. @LukeAVanDrie your choice.

Thanks for the review!

Regarding the first point, I agree that moving the file makes sense, but I will defer that to a follow-up PR to keep this specific cleanup focused as you suggested.

On the second point about naming, I am honestly a bit torn. I definitely understand your point about the potential confusion since flow control already uses the term `priorityBands`. On the other hand, `PriorityQueue` is the standard, idiomatic name for this specific data structure, so moving away from it might also cause confusion for those expecting standard nomenclature. 

I am not completely sure which tradeoff is better. Maybe we can ask in the Slack to get a few more opinions on whether to keep it as `PriorityQueue` or change it to `OrderedQueue` or `RankedQueue`? Let me know what you think! I would hold on the rename regardless until after the next release though as it is not particularly urgent.

shmuelk (approved):

PR #2041, opened 2026-07-16
fix(sidecar): cap min_tokens on the prefill leg for chat completions

vMaroon:
**What type of PR is this?**

/kind bug

P/D disaggregation rejects any chat completions request that sets `min_tokens` above 1.

The prefill leg caps output at one token, and `proxy.go` stages that cap through per-API token-limit field lists. `generateStyleTokenLimitFields` includes `min_tokens`, but `chatCompletionTokenLimitFields` did not, so on the routes that map to `APITypeChatCompletions` the client's `min_tokens` reached the prefiller unchanged next to `max_tokens=1`. vLLM's `CompletionRequest` and `ChatCompletionRequest` pass `min_tokens` straight into `SamplingParams`, whose validation rejects `min_tokens > max_tokens`:

```
ValueError: min_tokens must be less than or equal to max_tokens=1, got 5.
```

vLLM maps that to a 400, and 400 is not in the sidecar's retryable set (`isRetryableStatus` covers 502/503/504), so the status propagates and the request dies before decode ever runs. Any `min_tokens > 1` fails unconditionally — this is not load- or timing-dependent.

Repro against a P/D deployment:

```bash
curl -X POST "$GATEWAY/v1/chat/completions" -H 'Content-Type: application/json' -d '{
  "model": "Qwen/Qwen3-32B",
  "messages": [{"role": "user", "content": "Hello"}],
  "max_tokens": 100,
  "min_tokens": 5
}'
```

Adding `min_tokens` to `chatCompletionTokenLimitFields` stages it exactly like the existing caps: pinned to 1 on prefill and restored to the client's value on decode. Pinning to 1 is valid against `max_tokens=1` and semantically inert on that leg, since the prefiller's single sampled token is discarded once decode pulls the KV; `min_tokens` constrains generation, which only decode does. This matches what the generate API already does in the same connector.

Two consequences of that one-line change:

- `savedTokenValues` was a fixed `[2]savedField` that a third field would have overflowed, so it now sizes off the field list.
- `runNIXLProtocolV2WriteParallel` (MoRI-IO WRITE-mode concurrent dispatch) stages its own prefill body instead of going through the shared token-limit handling, so it gets the same save/restore.

Out of scope, worth their own issues: `connector_mooncake.go` and `connector_shared_storage.go` build prefill bodies independently and have the same gap. The Responses API is unaffected — vLLM's `ResponsesRequest` has no `min_tokens` field.

Fixes #

**Release note**:
```release-note
NONE
```

**Test plan**:

- [x] `min_tokens` capped on the prefill leg and restored to the caller's value on decode
- [x] caps applied on prefill and dropped on decode when the request omits them
- [x] same capping and restore on the MoRI-IO concurrent-dispatch path
- [x] full sidecar suite green (112 specs, `./pkg/sidecar/...`)

### DIFF
--- a/pkg/sidecar/proxy/connector_nixlv2.go
+++ b/pkg/sidecar/proxy/connector_nixlv2.go
@@ -115,7 +115,7 @@ func (s *Server) handleNIXLV2(w http.ResponseWriter, r *http.Request, prefillPod
 		present bool
 	}
 	tokenMap, createdSamplingParams := tokenLimitMap(completionRequest, apiType)
-	var savedTokenValues [2]savedField
+	savedTokenValues := make([]savedField, len(tokenLimitFields))
 	for i, field := range tokenLimitFields {
 		if v, ok := tokenMap[field]; ok {
 			savedTokenValues[i] = savedField{field: field, val: v, present: true}
@@ -324,7 +324,7 @@ retryLoop:
 		completionRequest[requestFieldStreamOptions] = streamOptionsValue
 	}
 
-	for i := range savedTokenValues[:len(tokenLimitFields)] {
+	for i := range savedTokenValues {
 		sv := &savedTokenValues[i]
 		delete(tokenMap, sv.field)
 		if sv.present {
@@ -459,6 +459,7 @@ func (s *Server) runNIXLProtocolV2WriteParallel(
 	maxTokensValue, maxTokensOk := completionRequest[requestFieldMaxTokens]
 	maxCompletionTokensValue, maxCompletionTokensOk := completionRequest[requestFieldMaxCompletionTokens]
 	maxOutputTokensValue, maxOutputTokensOk := completionRequest[requestFieldMaxOutputTokens]
+	minTokensValue, minTokensOk := completionRequest[requestFieldMinTokens]
 
 	// Pin both legs to the same DP rank (kv_transfer_params + HTTP header).
 	dpRank := pickDPRank(uuidStr, s.config.MoRIIODPSize)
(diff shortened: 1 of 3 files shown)


roytman:
Hi @vMaroon , there is a conflict between the PR title and the actual description/implementation. While the title is generic, the implementation focuses strictly on the Nixlv2 connector.
The PR description even mentions: "Out of scope, worth their own issues: connector_mooncake.go and connector_shared_storage.go"
Since the fixes are straightforward, I suggest including`connector_mooncake.go`, `connector_shared_storage.go` and `connector_p2p.go` in this PR.

We will update the coordinator kv connectors accordingly.

revit13:
This PR fixes min_tokens capping on the prefill leg for chat/completions requests via nixlv2, but there is a similar gap in pkg/sidecar/proxy/connector_ec_common.go in buildEncoderRequest (line ~114):

```bash
 for k, v := range originalRequest {
     encoderRequest[k] = v   // copies min_tokens from client
 }
 encoderRequest["max_tokens"] = 1
 // min_tokens is never zeroed here
```
 The encoder leg also sets max_tokens=1 but does not clear min_tokens, so a client request with min_tokens > 0 will reach the encoder pod with min_tokens > max_tokens=1, triggering the same vLLM SamplingParams validation error. Reproduced on a live cluster with min_tokens=50.

 Fix would be to add encoderRequest["min_tokens"] = 0 alongside the max_tokens override.

roytman:
@vMaroon, looks like we need a helper utility that can be reused among different connectors.
Do you want to create it, or we can merge this PR and we'll take it from there?

shmuelk:
@vMaroon Please fix the grammar issue raised by CoPilot...

After that we can merge this PR. The refcatoring mentioned by @roytman and @revit13 can be done in a follow up PR

roytman (approved):
/lgtm

PR #2047, opened 2026-07-16
fix(flowcontrol): return 429 on TTL eviction when pool has endpoints

### LINKED ISSUE #2046: [Flow Control] Queue-wait TTL eviction should return 429, not 503, when the pool has endpoints
## What happened

A queued request evicted because its queue-wait TTL expired (`flowControl.defaultRequestTTL`) gets HTTP 503: `translateFlowControlOutcome` maps `QueueOutcomeEvictedTTL` to `ServiceUnavailable` ([`pkg/epp/requestcontrol/admission.go`](https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/requestcontrol/admission.go#L233-L234)), with `x-llm-d-request-dropped-reason: rejected-ttl-expired`.

## Why this looks wrong

#1751 defined the split:

> - 429 (ResourceExhausted) — capacity exists but is contended (backpressure).
> - 503 (ServiceUnavailable) — no serving capacity exists right now.

A TTL eviction in a **non-empty pool** is the first case: endpoints are serving, the request was shed after waiting out contention. The service is exhausted, not unavailable. It is the same overload class as `QueueOutcomeRejectedCapacity` (429) — the shed just happens after waiting instead of at enqueue. Today the same saturated pool returns 429 or 503 depending only on whether the request got a queue slot, and the 503 misleads clients/load balancers into treating a live-but-backlogged backend as down.

On retry semantics: #1022 noted a TTL-expired request shouldn't be blindly retried — but that is what the `rejected-ttl-expired` drop-reason header conveys. The status code should still truthfully signal overload of a live service; 429 + header preserves both.

## Proposed change

- `QueueOutcomeEvictedTTL` → `ResourceExhausted` (429), keeping the `rejected-ttl-expired` header.
- Preserve #1751's empty-pool discrimination: TTL eviction with **zero endpoints** keeps signalling 503, reusing the pool-emptiness signal #1751 introduced for the capacity-rejection path.

|                     | Capacity rejection | TTL eviction |
|---------------------|--------------------|--------------|
| Empty pool          | 503                | 503          |
| Non-empty saturated | 429                | **429** (was 503) |

`EvictedContextCancelled` and shutdown-drained (#1612) stay 503, unchanged.

Happy to send a PR (mapping change + unit tests for the translation and the empty-pool discriminator).

mtparet:
**What type of PR is this?**

/kind bug

A queued request evicted because its queue-wait TTL (`flowControl.defaultRequestTTL`) expired was mapped to `ServiceUnavailable` (503). Per the availability doctrine established in #1751 — 429 = capacity exists but is contended (backpressure); 503 = no serving capacity exists right now — a TTL eviction in a non-empty pool is backpressure: endpoints are serving and the request was shed after waiting out contention. It is the same overload class as a queue-capacity rejection (429); the shed just happens after waiting instead of at enqueue. The 503 also misleads clients and load balancers into treating a live-but-backlogged backend as down.

This PR maps `QueueOutcomeEvictedTTL` to `ResourceExhausted` (429), keeping the `x-llm-d-request-dropped-reason: rejected-ttl-expired` header so consumers can still apply TTL-specific retry policy. It preserves #1751's empty-pool discrimination: a TTL eviction while the pool has zero endpoints returns 503 with reason `rejected-no-endpoints`, using the same whole-pool `Locate` discriminator (via the cached `EndpointCandidates` already constructed in the runner).

|                     | Capacity rejection | TTL eviction |
|---------------------|--------------------|--------------|
| Empty pool          | 503                | 503          |
| Non-empty saturated | 429                | **429** (was 503) |

`EvictedContextCancelled` and shutdown-drained (#1612) are unchanged (503).

**Test plan**:
- [x] Translation table test: TTL-with-endpoints → 429 + `rejected-ttl-expired`; TTL-empty-pool → 503 + `rejected-no-endpoints`; all other outcomes unchanged
- [x] Admission controller test covering both TTL branches via mocked `EndpointCandidates`

Fixes #2046

```release-note
Flow control now returns 429 (Too Many Requests) instead of 503 (Service Unavailable) when a queued request is evicted because its queue-wait TTL expired while the inference pool has endpoints, signalling retryable backpressure; the `x-llm-d-request-dropped-reason: rejected-ttl-expired` header is unchanged. A TTL eviction while the pool has no endpoints returns 503 with reason `rejected-no-endpoints`.
```


shmuelk (dismissed):

mtparet:
@shmuelk what do you think about the e2e-tests failure, should I investigate ? (seems not related)

LukeAVanDrie:
> @shmuelk what do you think about the e2e-tests failure, should I investigate ? (seems not related)

It was a flake that should be fixed on main now. I just rebased onto main for you and will try to get this merged.

Edit: never mind, I don't have code owners perms for request control dir, so @shmuelk will need to approve it again. If you do so, @shmuelk, can you make sure to remove the automatic co-author credit it will give me from the "Merge branch 'main' into fix/ttl-eviction-429" in the final squash commit?

LukeAVanDrie (dismissed):

LukeAVanDrie:
looks like one lint error too

shmuelk:
@mtparet Please sign your commits

ahg-g (approved):

mtparet:
Sorry @shmuelk I am off this week and did the commit fix not from my usual laptop. Thanks for the merge 🙏

PR #2048, opened 2026-07-16
Consolidating images and artifacts

ahg-g:
**What type of PR is this?**
/kind feature

1. Published a `main` tag for both `llm-d-router-endpoint-picker` and `llm-d-router-disagg-sidecar` images. These two images replace the use of  `llm-d-router-endpoint-picker-dev:main` and `llm-d-router-disagg-sidecar-dev:main`
2. `llm-d-router-endpoint-picker-dev:main` and `llm-d-router-disagg-sidecar-dev` continues to be published on every commit with the commit hash as a tag. These will be used for debugging and will not be referenced in any charts explicilty.
3. Stopped publishing `charts/llm-d-router-gateway-dev` and `charts/llm-d-router-standalone-dev` charts, the v0 chart (i.e., the one that builds from main) will be published under `charts/llm-d-router-gateway` and `charts/llm-d-router-standalone`

Fixes #

```release-note
Stopped publishing `charts/llm-d-router-gateway-dev` and `charts/llm-d-router-standalone-dev` charts, use `charts/llm-d-router-gateway` and `charts/llm-d-router-standalone` with --version v0 instead. Also, now one can use `llm-d-router-endpoint-picker:main`/`llm-d-router-disagg-sidecar:main` instead of `llm-d-router-endpoint-picker-dev:main`/`llm-d-router-disagg-sidecar-dev:main` to reference the images built from the head of main.
```


elevran:
Looks good, thanks.

@roytman @revit13 since I was reviewing #1975 one thing to note:

test/coordinator/e2e/coordinator/e2e_suite_test.go now defaults `EPP_IMAGE` to `:main`, but the coordinator's Makefile.coord.mk in 1975 sets `EPP_TAG ?= dev` and exports `EPP_IMAGE=...:dev`, which shadows the defaults set in this PR (e.g., whenever the coordinator e2e suite runs through the Makefile target). Direct `go test` invocations from the coordinator test dir would pick up `:main`. I think this is inconsistent and you should confirming which is the intended behavior.

elevran (dismissed):

shmuelk:
@ahg-g and @elevran PLEASE Do NOT merge this PR

It breaks running the end to end tests locally. The llm-d-router end to end tests in CI build an image before the running the tests. Using a pre-built image that was created after the last PR was merged is a BAD idea.

ahg-g:
> @ahg-g and @elevran PLEASE Do NOT merge this PR
> 
> It breaks running the end to end tests locally. The llm-d-router end to end tests in CI build an image before the running the tests. Using a pre-built image that was created after the last PR was merged is a BAD idea.

Which specific line in this PR changes the existing semantics? What do you suggest as a fix to the issue you noted?

test/coordinator/e2e/coordinator/e2e_suite_test.go
```diff
@@ -85,7 +85,7 @@ var (
 	printCoordinatorLogs = env.GetEnvBool("E2E_PRINT_COORDINATOR_LOGS", false, ginkgo.GinkgoLogr)
 
 	containerRuntime = env.GetEnvString("CONTAINER_RUNTIME", "docker", ginkgo.GinkgoLogr)
-	eppImage         = env.GetEnvString("EPP_IMAGE", "ghcr.io/llm-d/llm-d-router-endpoint-picker:dev", ginkgo.GinkgoLogr)
```
ahg-g, line 88: @shmuelk  I guess this is the one we need to revert (and all similar instances of it in other files); where do we build an image with a dev tag?

elevran (commented):
holding until @shmuelk points are reviewed

elevran:
we should look at ensuring e2e runs with the latest code from main + PR changes. I think right now it runs with whatever state the PR is udpated to.
Merge queue might be the native approach - need to investigate

test/coordinator/e2e/coordinator/e2e_suite_test.go
ahg-g: See latest commit, I reverted these changes

shmuelk: First of all `make image-build` builds the images (router + sidecar) with a tag of dev.

Second of all in the CI in the `ci-pr-checks.yaml` file builds images with a tag of dev for use in the end to end tests. In the end to end tests we want to use freshly built images with all of the code from the PR under test and not an image that was built with the merge of a prior PR.

Please don't misunderstand me, for other test purposes, such as the nightly performance test using a pre-built image is great. Naming it without the `-dev` in the name, simply with a tag of main, is a very good thing.

shmuelk:
/hold cancel

@ahg-g Thank you very much for the changes

ahg-g:
> Looks good, thanks.
> 
> @roytman @revit13 since I was reviewing #1975 one thing to note:
> 
> test/coordinator/e2e/coordinator/e2e_suite_test.go now defaults `EPP_IMAGE` to `:main`, but the coordinator's Makefile.coord.mk in 1975 sets `EPP_TAG ?= dev` and exports `EPP_IMAGE=...:dev`, which shadows the defaults set in this PR (e.g., whenever the coordinator e2e suite runs through the Makefile target). Direct `go test` invocations from the coordinator test dir would pick up `:main`. I think this is inconsistent and you should confirming which is the intended behavior.

I think the last commit addresses this as well.

revit13:
> Looks good, thanks.
> 
> @roytman @revit13 since I was reviewing #1975 one thing to note:
> 
> test/coordinator/e2e/coordinator/e2e_suite_test.go now defaults `EPP_IMAGE` to `:main`, but the coordinator's Makefile.coord.mk in 1975 sets `EPP_TAG ?= dev` and exports `EPP_IMAGE=...:dev`, which shadows the defaults set in this PR (e.g., whenever the coordinator e2e suite runs through the Makefile target). Direct `go test` invocations from the coordinator test dir would pick up `:main`. I think this is inconsistent and you should confirming which is the intended behavior.

@elevran Thanks for flagging this. I dug into the attribution and I don't think the e2e suites actually diverge here:

- Both e2e suites default EPP_IMAGE to ...llm-d-router-endpoint-picker:dev (test/coordinator/e2e/coordinator/e2e_suite_test.go and test/e2e/e2e_suite_test.go), matching their respective Makefiles (Makefile.coord.mk and top-level Makefile, both EPP_TAG ?= dev). Both export EPP_IMAGE into the suite, so the Makefile-driven and direct go test paths resolve to the same :dev tag.
- This PR (#2048) doesn't touch either e2e suite or any Makefile. main shows up in two places here: the EPP_TAG default in test/perf/run_nightly_perf.py, and a published image tag (epp-additional-tags: ...:main in ci-dev.yaml). So post-#2048 the canonical published dev EPP image is llm-d-router-endpoint-picker:main, while the e2e suites still default to :dev.
- This is fine today because the e2e paths build EPP locally and tag it :dev rather than pulling, so they don't depend on the published tag.

elevran (approved):

Issue #2039, opened 2026-07-15
Make pd-sidecar flags modelserver agnostic


rahulgurnani:
Presently, several of the flags that pd-sidecar takes as input are vLLM specific, for example, `vllm-port` (should be named `model-server-port`), `kv-connector=sglang` (kv_connector is part of vLLM, we can perhaps rename this as well)

This issue involves re-visiting some of these user facing flags to ensure they look model server agnostic. 

**Change strategy**

The changes in the flags can happen by first introducing new flags and then deprecating the old flags over the course of 2 releases in a phased manner, similar to https://github.com/llm-d/llm-d/pull/1419

cc @elevran

zdtsw:
open https://github.com/llm-d/llm-d-router/pull/2054 to first get deprecated flags removed

zdtsw:
Some thoughts:
1.
rename `--vllm-port` to  `--model-server-port`
Add `--model-server-port`; keep `--vllm-port` as a deprecated alias + log deprecation warning. Do the final cleanup of  `--vllm-port ` in release later.

2.
`--data-parallel-size` keep the flag name but do some twek in the help text (mianly remove vllm )

3.
split `sglang` out of `--kv-connector` into its own new flag `--sglang-connector`
`--kv-connector ` values `nixlv2/mooncake/offloading/shared-storage` all map to vLLM's internal KVConnector implementations. `sglang` is different in kind,  it's a peer model server with its own PD bootstrap protocol (separate handleSGLang path), not a vLLM connector. So it should not belong as a `--kv-connector ` as a value [ref](https://github.com/llm-d/llm-d-router/blob/c82561bfd05902c93b55998e9ee9a410977c0d07/pkg/sidecar/proxy/proxy.go#L464)
e.g
-  Add `--sglang-connector` (boolean). It refers internally to the existing SGLang code path -- no behavior change for now
-  `--kv-connector=sglang` keeps working through the deprecation window with a warning.
-  Add a gate: reject if both `--sglang-connector` and an explicit` --kv-connector` are set.
-  After deprecation window: drop `sglang` from the `--kv-connector `

rahulgurnani:
I agree with 1. and 2.

For 3. , it may be better to do something that generalizes across model servers.

PR #2054, opened 2026-07-16
chore: remove deprecated sidecar flags

zdtsw:
**What type of PR is this?**
/kind cleanup
/kind deprecation

flag deprecation was added in 0.7.0 (2026.02)
- `connector` is replaced by `kv-connector`
- `prefiller-use-tls` and  `decoder-use-tls` are both replaced by `enable-tls`
- `prefiller-tls-insecure-skip-verify` and `decoder-tls-insecure-skip-verify` are both replaced by `tls-insecure-skip-verify`
- yamlconfig is parsed strictly, a removed key or CLI flag now aborts sidecar startup with  error

Ref #2039

```release-note
The following pd-sidecar flags, deprecated since v0.7.0, have been removed in 0.10
Use their replacements instead:
- `--connector` -> `--kv-connector`
- `--prefiller-use-tls` -> `--enable-tls=prefiller`
- `--decoder-use-tls` -> `--enable-tls=decoder`
- `--prefiller-tls-insecure-skip-verify` -> `--tls-insecure-skip-verify=prefiller`
- `--decoder-tls-insecure-skip-verify` -> `--tls-insecure-skip-verify=decoder`
The same names apply to the sidecar YAML configuration keys.
```

### DIFF
--- a/pkg/sidecar/proxy/options.go
+++ b/pkg/sidecar/proxy/options.go
@@ -76,14 +76,6 @@ const (
 	configurationFile         = "configuration-file"
 	tracingFlag               = "tracing"
 
-	// Deprecated flags
-	connector                      = "connector"
-	prefillerUseTLS                = "prefiller-use-tls"
-	decoderUseTLS                  = "decoder-use-tls"
-	encoderUseTLS                  = "UseTLSForEncoder"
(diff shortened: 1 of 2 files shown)


liu-cong (approved):
thanks!

Issue #1906, opened 2026-07-06
[Performance] run pprof and analyze the results


liu-cong:
Once we run pprof as part of the load testing https://github.com/llm-d/llm-d-router/issues/1346, we should analyze the results, and suggest improvements if applicable.

jtechapps:
I ran a few performance using the automation introduced in https://github.com/llm-d/llm-d-router/pull/2022.

# EPP Control Plane Performance & Profiling Evaluation Report

## Executive Summary

This report synthesizes performance benchmarks, resource utilization metrics, and `pprof` CPU/memory profiling captured for the **Endpoint Picker (EPP)** router service under stress testing workloads (10 vLLM simulator replicas, high QPS, initial 10k/1k prompt baselines, and large 100,000-token prompt inputs).

---

## 1. High-Level CPU Usage Breakdown (100% Accounting)

Under peak 100k-token prompt workload at 10 QPS (**~3.33 active CPU cores** / 100.02 CPU sample seconds), CPU execution time is distributed across the following core subsystems:

| Subsystem / Category | Flat CPU Time | % of Total CPU | Operational Description |
|---|---|---|---|
| 🌐 **1. Network I/O & Linux Kernel Syscalls** | **19.31s** | **19.31%** | Kernel socket operations (`Syscall6`, `netpoll`, `crypto/tls`) servicing streaming HTTP/gRPC data packets. |
| 📄 **2. JSON Request Payload & Stream Decoding** | **18.60s** | **18.60%** | Unmarshalling 100k-token prompt JSON request bodies (`encoding/json.checkValid`, `unquoteBytes`, `OpenAIParser.ParseRequest`). |
| 🧹 **3. Go GC & Runtime Heap Memory Management** | **13.11s** | **13.11%** | Go allocator & Garbage Collection routines (`mallocgc`, `nextFreeFast`, `memclr`, `memmove`) servicing short-lived JSON & buffer allocations. |
| ⚙️ **4. Goroutine Scheduling & Concurrency Sync** | **12.77s** | **12.77%** | Multithreaded OS thread coordination (`futex`, `selectgo`, `stealWork`, `schedule`) managing goroutines across active CPU cores. |
| 🛠️ **5. Go Runtime Low-level Internal Helpers** | **11.85s** | **11.85%** | Stack management, Go map operations (`internal/runtime/maps`), and assembly byte routines (`indexbytebody`, `aeshashbody`). |
[...]

PR #2058, opened 2026-07-16
feat(epp): match whole prompt when both prefix caps are zero

kaushikmitr:
**What type of PR is this?**
/kind feature

Setting both `maxPrefixTokensToMatch` and `maxPrefixBlocksToMatch` to `0` now
matches the whole prompt. Previously it hashed nothing at all, leaving the
producer, its indexer, and its cleanup goroutine running while contributing
nothing to scoring, with nothing in the config surface signalling it.

When this plugin was written, the prefix match existed only to inform prefix
*scoring*. For that purpose a capped match is fine: the cap is uniform across
pods, so it biases every candidate equally and the ranking still holds. Matching
"enough" prefix was sufficient, and the cap bought bounded EPP CPU.

The match info is now also consumed by the inflight-load producer to compute
`uncachedInputTokens` -- the prompt tokens an endpoint must actually compute,
discounting the prefix it already has cached. That consumer needs an absolute
quantity rather than a comparable one, so the cap stops being neutral. Beyond
the cap, `uncachedInputTokens` adds the un-indexed tail back from the estimated
`inputTokens` and charges it as fully uncached, because the index cannot say
otherwise. Where the server actually holds that tail in its prefix cache, the
endpoint's in-flight load is over-counted, and the error scales with how far the
prompt runs past the cap rather than with the pod's real cache state. Two pods
with genuinely different cached tails look identical to the estimate.

The 131072-token default was sized against 128K-context models, where it never
binds. GLM-5.2 serves a 1M-token context on Vertex AI
(`vertex_ai/zai-org/glm-5-maas`); at full context ~87% of the prompt falls past
the cap, so the estimate is dominated by the unindexed tail.

Zeroing both caps removes the cap as a source of that error. Prompt length is
already bounded by the model server's context window, so an absent cap matches
at most one context window of tokens -- operators get context-window-sized
matching without hand-maintaining a token count that tracks `max_model_len`.

This does not make the accounting exact. The routing-side index is an
approximation of server-side cache state, so evictions on the model server still
produce a delta between what the index believes is cached and what the server
actually holds. That residual is expected to be small relative to the cap-induced
error, and unlike the cap it is not systematic in prompt length.

Uncapped matching raises EPP CPU on long prompts (see docs/operations.md), so
this stays opt-in: defaults are unchanged, and the behavior only applies when
both caps are explicitly `0`.

Notes on scope:
- Precedence is unchanged -- `maxPrefixTokensToMatch` still wins when set (> 0).
- A token cap smaller than the block size still resolves to zero blocks and
  hashes nothing. That is a misconfiguration (`burstprefix` rejects it outright),
  not a request for unlimited matching, so it is left as-is.
- To disable prefix matching, remove the producer from the configuration. Zeroing
  the caps is no longer a way to do that; README is updated accordingly.

Fixes #

**Release note**:
```release-note
The approximate prefix-cache producer (approx-prefix-cache-producer) now matches the entire prompt when both maxPrefixTokensToMatch and maxPrefixBlocksToMatch are set to 0, which previously disabled prefix matching entirely. This makes in-flight token accounting reflect the full prompt on deployments with context windows larger than the 131072-token default cap. Defaults are unchanged. Uncapped matching increases EPP CPU utilization on long prompts. To disable prefix matching, remove the producer from the configuration rather than zeroing the caps.

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/README.md
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/README.md
@@ -12,8 +12,10 @@ For each request, the plugin consumes `request.Body.TokenizedPrompt` (token IDs)
 
 - `autoTune` (bool, optional, default: `true`): Infer `blockSizeTokens` and `lruCapacityPerServer` from endpoint metrics when available.
 - `blockSizeTokens` (int, optional, default: `16`): Prefix block size in tokens. Used when `autoTune` is false or endpoint metrics are unavailable; ignored when auto-tuned from metrics. Values below the minimum of `64` are clamped up at request time (see #1158), so the `16` default is effectively `64` absent a larger metric/configured value.
-- `maxPrefixBlocksToMatch` (int, optional, default: `2048`): Maximum number of prefix blocks hashed and matched per request. Not auto-tuned. `0` disables matching (zero blocks hashed).
+- `maxPrefixBlocksToMatch` (int, optional, default: `2048`): Maximum number of prefix blocks hashed and matched per request. Not auto-tuned. Applies only when `maxPrefixTokensToMatch` is `0`.
 - `maxPrefixTokensToMatch` (int, optional, default: `131072`): Cap expressed in tokens instead of blocks (`maxBlocks = maxPrefixTokensToMatch / blockSizeTokens`). Not auto-tuned. Takes precedence over `maxPrefixBlocksToMatch` when set (> 0); set to `0` to fall back to the block-based cap. The `131072` default (128K, the context window of large production models such as gpt-oss 120b) is a reasonable upper bound that covers the long-prompt use cases seen in production.
+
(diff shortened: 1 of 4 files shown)


kaushikmitr:
@liu-cong @ahg-g @jtechapps

ahg-g (approved):

jtechapps:
From performance perspective this is fine:

Even with matching 1M input tokens matching 50 tokens vs 1M is < 9% difference in CPU.

[fixed_50_prefix_vs_full_prefix_analysis.md](https://github.com/jtechapps/llm-d-router/blob/performanceanalysisresults/llm-d-router-analysis/fixed_50_prefix_vs_full_prefix_analysis.md)

I also did pprof analysis at 100k input tokens and the approximate prefix matching computation is efficient. 

https://github.com/llm-d/llm-d-router/issues/1906#issuecomment-4985530888

jtechapps:
We can probably close https://github.com/llm-d/llm-d-router/issues/1505 when this is merged.

kaushikmitr:
> We can probably close #1505 when this is merged.

i think (maybe in a separate PR as it will change default behavior) make defaults 0

PR #2067, opened 2026-07-17
Add support for ZMQ metrics updates

azamikram:
**Update (2026-09-16):** The ZMQ extractor is now engine-based, selected by a required `engine` parameter. The first supported engine is SGLang, using its native ZMQ LoadStat stream (sgl-project/sglang#34608). The vLLM decoder will return as `engine: vllm` once vllm-project/vllm#48969 settles, so this PR no longer waits on the vLLM side. New benchmark results (EPP CPU, freshness, model server load) are in this comment: https://github.com/llm-d/llm-d-router/pull/2067#issuecomment-5693345709

**What type of PR is this?**

/kind feature

Detailed Report: [link](https://docs.google.com/document/d/10Juqz_a-LhrOdpezV--NesW1KRHHIIEGV0of8xR7OX0/edit?tab=t.0)

## Summary & Proposal
This PR introduces an asynchronous, event-driven ZeroMQ (ZMQ) metrics streaming architecture over persistent TCP sockets (`tcp://<pod_ip>:5556`) to push state updates directly from model servers (specifically vLLM) to the EPP router.

### Problem Statement
The existing pull-based HTTP polling approach periodically issues GET requests to each model server's `/metrics` endpoint (at a default 50ms interval). Because model server state mutates exclusively at GPU step boundaries, HTTP polling creates two critical failure modes:
1. **Redundant Duplication:** Polling during long prefills fetches identical unmutated state repeatedly (95.4% duplicate messages in our experiments).
2. **Telemetry Blind Spots & Stalls:** Polling cannot align with fast decode iterations or sudden traffic bursts, delaying state visibility. Attempting to lower the polling interval to 6ms actually triggers large number of state update messages contented the network bandwidth and CPU thread contention, spiking P95 update latency to **131.8 ms** (and peak stalls to **798.9 ms**).

### Proposed Solution
With this push-based technique, the model server streams a lightweight binary update (`msgspec` Msgpack) to EPP immediately after each GPU execution step when engine state actually changes. In addition to improving state freshness, this push-based mechanism drastically reduces CPU utilization on both EPP and model servers by eliminating continuous HTTP status check parsing and string formatting during idle and active periods alike.

---

## Key Performance Results
Empirical evaluation on a 3-pod vLLM cluster running `Qwen/Qwen3-32B` under a 300 QPS burst workload demonstrates significant throughput, latency, and resource improvements:
- **Increased Throughput:** Cluster token throughput increased by **+18.44%** (from 31,177 to 36,925 tokens/s).
- **Lower Tail Latency:** P99 TTFT improved by **-16.20%** (reducing tail prompt wait time from 54.92s to 46.02s). P99 End-to-End latency improved by **-15.18%**.
- **Drastic CPU Savings:** Active EPP router CPU usage dropped by **-64.51%** (from 324.87 to 115.30 mcores). Zero-traffic idle CPU dropped by **-95.96%** on EPP (from 108.71 to 4.39 mcores/endpoint) and **-81.02%** on vLLM engines (from 91.48 to 17.36 mcores/pod).

**Macro results:**

Throughput and CPU utilization

image

Latency CDF

image

Idle CPU utilization

image

### Architecture & Design Changes

Overall Design

image

#### 1. Data Layer Runtime & Collector Orchestration
- Updated `Collector.Start()` to manage concurrent goroutines for interval-based polling (`runPolling`) and persistent socket streaming (`runStreaming`).
- Integrated `streamingDispatchers` registry into source registration, configuration binding, and endpoint provisioning (`NewEndpoint`).
- Added `validateMetricSourceConflicts()` to prevent configuring both HTTP polling (`metrics-data-source`) and ZMQ streaming (`zmq-metrics-data-source`) metrics sources concurrently.

#### 2. ZeroMQ Data Source Plugin (`pkg/epp/framework/plugins/datalayer/source/zmqmetrics/datasource.go`)
- Implements `StreamingDispatcher`, connecting ZMQ SUB sockets to endpoint TCP addresses (`tcp://<ip>:<port>`, default port 5555).
- Includes automatic exponential backoff reconnection logic (1s up to 30s max backoff) on stream interruptions, and per-extractor panic recovery to prevent extractor failures from disrupting the stream worker loop.

#### 3. Msgpack Metrics Extractor (`pkg/epp/framework/plugins/datalayer/extractor/zmqmetrics/extractor.go`)
- Decodes Msgpack payloads via unmarshaling into `ZmqMetricsStats`.
- Updates endpoint state atomically (`RunningRequestsSize`, `WaitingQueueSize`, `KVCacheUsagePercent`, `CacheBlockSize`, `CacheNumBlocks`, `KvCacheMaxTokenCapacity`).

#### 4. Plugin Registration & Default Ingestion (`runner.go` and `defaults.go`)
- Registered `zmq-metrics-data-source` and `zmq-state-extractor` into the in-tree plugin registry.
- Updated default configuration loader (`ensureDataLayer`) to bypass default HTTP polling metrics injection when `zmq-metrics-data-source` is present.

Fixes #

vLLM side issue: https://github.com/vllm-project/vllm/issues/43643
vLLM side PR: https://github.com/vllm-project/vllm/pull/48969

```release-note
Add zmq-metrics-data-source and zmq-state-extractor plugins for push-based endpoint metrics (SGLang LoadStat)
```

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -71,10 +71,12 @@ import (
 	extractormetrics "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/extractor/metrics"
 	extmodels "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/extractor/models"
 	exttopology "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/extractor/topology"
+	extractorzmq "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/extractor/zmqmetrics"
 	srcdcgm "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/source/dcgm"
 	sourcemetrics "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/source/metrics"
 	srcmodels "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/source/models"
 	sourcenotifications "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/source/notifications"
+	sourcezmq "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/source/zmqmetrics"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/fairness/globalstrict"
 	programaware "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/fairness/roundrobin"
@@ -640,6 +642,8 @@ func (r *Runner) registerInTreePlugins() {
 	// register datalayer metrics collection plugins
 	fwkplugin.Register(sourcemetrics.MetricsDataSourceType, sourcemetrics.MetricsDataSourceFactory)
 	fwkplugin.Register(extractormetrics.MetricsExtractorType, extractormetrics.CoreMetricsExtractorFactory)
+	fwkplugin.Register(sourcezmq.ZMQDataSourceType, sourcezmq.ZMQDataSourceFactory)
+	fwkplugin.Register(extractorzmq.ZMQExtractorType, extractorzmq.ZMQExtractorFactory)
 	// register datalayer notification source plugins
 	fwkplugin.Register(sourcenotifications.NotificationSourceType, sourcenotifications.NotificationSourceFactory)
 	fwkplugin.Register(sourcenotifications.EndpointNotificationSourceType, sourcenotifications.EndpointSourceFactory)
--- a/pkg/epp/config/loader/defaults.go
+++ b/pkg/epp/config/loader/defaults.go
@@ -27,6 +27,7 @@ import (
 	fwksched "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
 	extractormetrics "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/extractor/metrics"
 	sourcemetrics "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/source/metrics"
+	sourcezmq "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/source/zmqmetrics"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requesthandling/parsers/anthropic"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requesthandling/parsers/openai"
@@ -302,7 +303,7 @@ func ensureDataLayer(cfg *configapi.EndpointPickerConfig, handle fwkplugin.Handl
 	if cfg.DataLayer != nil && cfg.DataLayer.InjectDefaults != nil && !*cfg.DataLayer.InjectDefaults {
 		return nil
 	}
-	if cfg.DataLayer != nil && hasSourceOfType(cfg.DataLayer, sourcemetrics.MetricsDataSourceType) {
+	if cfg.DataLayer != nil && (hasSourceOfType(cfg.DataLayer, sourcemetrics.MetricsDataSourceType) || hasSourceOfType(cfg.DataLayer, sourcezmq.ZMQDataSourceType)) {
 		return nil
 	}
 
--- a/pkg/epp/config/loader/defaults_test.go
+++ b/pkg/epp/config/loader/defaults_test.go
@@ -105,6 +105,23 @@ func TestEnsureDataLayer(t *testing.T) {
 		require.Len(t, cfg.DataLayer.Sources, 1, "no duplicate metrics source")
 	})
 
+	t.Run("existing zmq-metrics-data-source suppresses default metrics-data-source injection", func(t *testing.T) {
+		cfg := &configapi.EndpointPickerConfig{
+			DataLayer: &configapi.DataLayerConfig{
+				Sources: []configapi.DataLayerSource{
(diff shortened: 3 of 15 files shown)


elevran:
@azamikram thanks for driving this and sharing results - looks promising!
A few additional bits of information that could be useful:
- any measurement of the impact on GC? Not needing to allocate Prom text parsers and parse entire sets just for the few metrics of interest.
- this could be integrated with the vllm API extension framework?
- (and consequetly to the above of integrating with vllm FE API) would it be possible to use SSE or grpc streaming response for the metrics updates, instead of zmq? Might be an easier integration path if all remains HTTP

azamikram:
@elevran, thanks for the excellent pointers.

> any measurement of the impact on GC?

I ran a microbenchmark and your intuition was correct that by not parsing the Prom text, we can save a large number of memory alloc calls. Compared to EPP's default 50ms HTTP polling, ZMQ Streaming generates 4.0x less heap garbage (10.58 MB/s vs. 42.05 MB/s). Compared to extreme 6ms polling, ZMQ reduces heap allocation rate by 20x (10.58 MB/s vs. 211.20 MB/s).

I have added more detailed results in the main report. Here's the link and checkout the `effect on GC` tab: https://docs.google.com/document/d/10Juqz_a-LhrOdpezV--NesW1KRHHIIEGV0of8xR7OX0/edit?tab=t.pslyev1kpomj

> this could be integrated with the vllm API extension framework?

In my current PR on vLLM side, I simply added the ZMQ PUB in their FE API. However, [it is possible to avoid changing their frontend and possibly (kind of) monkey patch the PUB into API extension framework.](https://docs.vllm.ai/en/latest/design/endpoint_plugins/) However, I'm not sure if we will be able collect all the required state from the engine that way or not. Some investigative work needs to be done on this.

> would it be possible to use SSE or grpc streaming

Absolutely! The key idea we are proposing is that the model server should be responsible for communicating the state to EPP rather than EPP periodically probing the model server. Any protocol that allows for a persistent streaming connection between model server and EPP would work. Another property that we look for is the ability for the model server to broadcast its state to enable this feature in active-active. But ZMQ, gRPC, and SSE all support broadcasting.

We didn't try SSE but between gRPC and ZMQ, we picked ZMQ because it avoids parsing the full network stack which results in slightly lower latency at higher frequency. We have detailed results comparing ZMQ and gRPC in our main report.

zetxqx:
## Update: SGLang is the first engine on the ZMQ path

The latest push reworks `zmq-state-extractor` to decode passed by a required `engine` parameter. The first supported engine is SGLang: the extractor subscribes to SGLang's native ZMQ LoadStat stream (enabled with `--load-publish-endpoint`, available since SGLang v0.5.19, added in sgl-project/sglang#34608). The vLLM decoder is removed for now and will come back as `engine: vllm` once its wire format settles.

Each LoadStat message is a small msgpack array (msgspec `array_like` with a tag):

```text
["LoadStat", num_running_reqs, num_waiting_reqs, num_tokens, max_total_num_tokens, attn_dp_rank]
```

SGLang publishes on every scheduler step, and re-sends the last snapshot about once per second when idle. 

Config:

```yaml
dataLayer:
  plugins:
  - type: zmq-metrics-data-source
    parameters:
      port: 5557
  - type: zmq-state-extractor
    parameters:
      engine: sglang
  sources:
  - pluginRef: zmq-metrics-data-source
    extractors:
    - pluginRef: zmq-state-extractor
```

## Measured benefits: ZMQ push vs HTTP /metrics polling

Setup: this PR's EPP + SGLang v0.5.19 (Qwen3-8B, 1x H100 per pod) on GKE. Polling mode scrapes each pod's `/metrics` every 50 ms. ZMQ mode subscribes to SGLang's LoadStat push stream. 

### 1. Idle EPP CPU

| Endpoints | polling | zmq | saved |
|---|---|---|---|
| 1 | 146 mcores | 3.4 | 142 |
| 2 | 276 mcores (3 runs) | 4.7 | 271 |
| 4 | 552 mcores (2 runs) | 5.1 | 547 |

EPP CPU vs endpoint count, idle and under load

### 2. EPP CPU under load

Load: closed loop, 32 concurrent streaming completions, 200 output tokens each, 180 s, about 20 req/s, about 4000-4200 output tokens/s total.

| Endpoints | polling | zmq | saved |
|---|---|---|---|
| 1 | 1392 mcores | 1372 | ~0 (within noise) |
| 2 | 1703 mcores (3 runs) | 1325 | 378 |
| 4 | 1909 mcores (2 runs) | 1366 | 543 |

### 3. Freshness

Measured inside the pod with one clock: the ZMQ stream and a 50 ms `/metrics` poll captured at the same time. Workload: 45 s capture on 1 SGLang pod; at t=12 s, 6 completion requests sent through the EPP, one every 0.4 s (short prompt, 120 output tokens, about 1 s each). Requests overlap partially, so the running count ramps up and down through many short-lived states: peak 3 concurrent in run 1, 2 in run 2. The rest of the capture is idle except the 10 s liveness-probe blips. Two runs.

| | ZMQ | 50 ms polling |
|---|---|---|
| State transitions captured | 18/18 and 14/14 | 11/18 and 9/14 |
| Detection lag, median | ~0 (it is the source) | 22 / 42 ms |
| Detection lag, max | ~0 | 283 / 202 ms |
| Cost per observation | ~0 | 7.6 ms GET+parse |

- Polling missed every short-lived state in both runs, including the peak concurrency of the burst.
- Max polling lag was 4-6x the poll interval, because SGLang updates its Prometheus gauges lazily (about 200 ms cadence). A faster poll interval cannot fix this.
- End to end, a ZMQ update reaches the EPP's endpoint state in under 10 ms; new-request admissions are published immediately.

True running count (ZMQ) vs the 50 ms polled view during the burst, run 1

### 4. Model server also beneftis

In Active-Active EPP setup, Every EPP replica polls independently. 

Idle SGLang pod, 3 EPP replicas:
| | polling | zmq |
|---|---|---|
| SGLang CPU | 1520 mcores | 1015 (= idle sglang baseline) |
| Extra CPU to be observed | +505 mcores (+50%) | 0 |
| Each EPP replica | ~135 mcores | ~5 |

The scrape tax is almost linear: baseline 1015, +165 mcores per polling EPP replica (1 replica: 1180, 3 replicas: 1520).

SGLang pod idle CPU vs number of EPP replicas observing it

PR #2073, opened 2026-07-18
feat: implement EPP plugin stability lifecycle and --allow-experimental-plugins flag (#1912)

### LINKED ISSUE #1912: [FEAT] Introduce Plugin Stability Lifecycle (Alpha, Beta, Stable) and Dependency Validation
### Problem Statement & Motivation

Currently, plugins in `llm-d-router` (`pkg/epp/framework/plugins/`) lack formal stability categorization (Alpha, Beta, Stable). For backwards-incompatible changes or deprecations, the project currently relies on ad-hoc logging and manual README updates across minor releases. 

To provide clear compatibility guarantees to users and maintainers, we propose introducing a **Plugin Stability Lifecycle (Alpha, Beta, Stable)** 

- **Alpha**: Rapid experimentation, breaking changes anytime, no deprecation period.
- **Beta**:  [Current policy] Feature-complete; breaking changes follow a formal **+2 minor version deprecation policy** (warn in release $N$ and $N+1$, remove in $N+2$).
- **Stable**: Enterprise-grade; breaking parameter or behavioral changes only occur on major EPP version bumps.

liu-cong:
This PR implements the proposal for EPP plugin stability lifecycle and CLI flag gating discussed in [#1912](https://github.com/llm-d/llm-d-router/issues/1912).

FIXES #1912

1. Generally new plugins should start with Alpha, which allows agile iterations without worrying about deprecation. Alpha plugins can only be used when `--allow-experimental-plugins` is set on the EPP runner.
2. All plugins added after the 0.9 release are marked as Alpha. Others are marked as Beta (backwards compatible).
3. In addition, the session_affinity filter/scorer, `sheddable` and `priority_time` plugins are marked Alpha despite pre-dating 0.9. These are never really used and still under active iteration.

**Release note**:

```release-note
Implemented plugin stability levels (Alpha, Beta, Stable) in EPP registry. Alpha plugins are opt-in via the `--allow-experimental-plugins=true` CLI flag.
```


shmuelk:
@liu-cong Overall the PR seems quite good.

One question.

You dealt with Consumers depending on Producers of a lower level of stability. This is good.

You didn't deal with other cases of dependencies betwen plugins. Such as the latest Disaggregated plugin which is dependent on decider plugins.

Shouldn't this PR deal with that as well?

pkg/epp/datalayer/data_graph.go
```diff
@@ -139,6 +140,43 @@ func CreateMissingDataProducers(ctx context.Context, defaultProducerRegistry map
 		}
 	}
 
+	return ValidatePluginStabilityDependencies(handle.GetAllPlugins(), logger)
+}
+
+// ValidatePluginStabilityDependencies checks if a consumer plugin depends on a producer plugin of lower stability
+// and returns an error if a stability degradation is detected.
+func ValidatePluginStabilityDependencies(plugins []plugin.Plugin, logger logr.Logger) error {
```
LukeAVanDrie: This validation runs unconditionally and never checks the `experimentalPlugins` gate. That brings back the concern @shmuelk raised in #1912 ("it will prevent new ProviderPlugins that provide the same information from new sources"), which the feature gate was supposed to answer.

For example: `burstprefix` (Alpha) exists to produce `PrefixCacheMatchInfoDataKey` for the `prefix` scorer (Beta), which lists that key under `Consumes().Required`. A user who enables `experimentalPlugins` and points the prefix scorer at burstprefix gets a fatal "stability degradation detected" at startup with no way around it, so the gate doesn't actually unlock the main Alpha-producer use case.

What if we plumb the feature gates in and log a warning instead when the gate is enabled? Enabling the gate is the user's consent to experimental behavior. With the gate off, Alpha plugins can't instantiate at all, so the hard error there is mostly redundant anyway.

Separate nit: the `logger` parameter is unused

pkg/epp/config/loader/configloader.go
```diff
@@ -260,9 +273,20 @@ func buildPluginDAG(configuredPlugins []configapi.PluginSpec, handle fwkplugin.H
...
 			}
-			graph[spec.Name] = findPluginDependencies(configStruct)
+			deps := findPluginDependencies(configStruct)
+			graph[spec.Name] = deps
+
+			parentStability := fwkplugin.GetPluginStability(spec.Type)
+			for _, depName := range deps {
+				if depSpec, ok := pluginMap[depName]; ok {
+					childStability := fwkplugin.GetPluginStability(depSpec.Type)
+					if childStability.Rank() < parentStability.Rank() {
```
LukeAVanDrie: Same gate question as the data-layer check: a Beta profile handler referencing a future Alpha decider via pluginRef will hard-fail even with `experimentalPlugins` enabled. Whatever is decided there should apply here too so the two checks stay consistent.

cmd/epp/runner/runner.go
```diff
@@ -543,109 +543,134 @@ func setupDatastore(ctx context.Context, epFactory datalayer.EndpointFactory,
...
+	// Beta
+	fwkplugin.Register(globalstrict.GlobalStrictFairnessPolicyType, fwkplugin.StabilityBeta, globalstrict.GlobalStrictFairnessPolicyFactory)
+	fwkplugin.Register(roundrobin.RoundRobinFairnessPolicyType, fwkplugin.StabilityBeta, roundrobin.RoundRobinFairnessPolicyFactory)
+	fwkplugin.Register(programaware.ProgramAwarePluginType, fwkplugin.StabilityBeta, programaware.ProgramAwarePluginFactory)
+	fwkplugin.Register(fcfs.FCFSOrderingPolicyType, fwkplugin.StabilityBeta, fcfs.FCFSOrderingPolicyFactory)
+	fwkplugin.Register(edf.EDFOrderingPolicyType, fwkplugin.StabilityBeta, edf.EDFOrderingPolicyFactory)
+	fwkplugin.Register(slodeadline.SLODeadlineOrderingPolicyType, fwkplugin.StabilityBeta, slodeadline.SLODeadlineOrderingPolicyFactory)
+	fwkplugin.Register(usagelimits.StaticUsageLimitPolicyType, fwkplugin.StabilityBeta, usagelimits.StaticPolicyFactory)
+	// Alpha
+	fwkplugin.Register(priorityholdback.PolicyType, fwkplugin.StabilityAlpha, priorityholdback.PolicyFactory)
```
LukeAVanDrie, line 634: You removed the `init()` registrations in `flowcontrol/eviction/filtering/sheddable.go` and `flowcontrol/eviction/ordering/priority_time.go` but didn't add them here, so `sheddable-eviction-filter` and `priority-then-time-eviction-order-policy` now have no registration path at all.

To be fair this isn't a regression, since on main nothing outside the flowcontrol integration test imports those packages and the `init()`s never ran in the EPP binary anyway. But if runner.go is the single source of truth, they should be listed:

```suggestion
	fwkplugin.Register(priorityholdback.PolicyType, fwkplugin.StabilityAlpha, priorityholdback.PolicyFactory)
	// eviction policies
	fwkplugin.Register(evictionfiltering.SheddableFilterType, fwkplugin.StabilityAlpha, evictionfiltering.SheddableFilterFactory)
	fwkplugin.Register(evictionordering.PriorityThenTimeOrderingType, fwkplugin.StabilityAlpha, evictionordering.PriorityThenTimeOrderingFactory)
```

The package predates v0.9.0, so the letter of the rule says Beta, but I'd register them Alpha: they were never reachable in a released binary, so there are no existing users to stay compatible with, and the eviction runtime they belong to isn't wired up yet (see #2061).

cmd/epp/runner/runner.go
```diff
@@ -543,109 +543,134 @@ func setupDatastore(ctx context.Context, epFactory datalayer.EndpointFactory,
...
+
 	// register discovery plugins
-	fwkplugin.Register(discoveryfile.PluginType, discoveryfile.Factory)
+	// Beta
+	fwkplugin.Register(discoveryfile.PluginType, fwkplugin.StabilityBeta, discoveryfile.Factory)
+
 	// register request header processor plugins
-	fwkplugin.Register(agentidentity.PluginType, agentidentity.PluginFactory)
+	// Beta
+	fwkplugin.Register(agentidentity.PluginType, fwkplugin.StabilityBeta, agentidentity.PluginFactory)
```
LukeAVanDrie: Non-blocking: `agentidentity` was added after the v0.9.0 tag, but is registered Beta, while the PR description says all post-0.9 plugins start Alpha. Should this be `StabilityAlpha`? Everything else I spot-checked follows the rule.

pkg/epp/framework/interface/plugin/stability.go
```diff
@@ -0,0 +1,42 @@
...
+	// StabilityAlpha indicates an experimental plugin with no backwards-compatibility guarantees.
+	StabilityAlpha StabilityLevel = "Alpha"
+	// StabilityBeta indicates a feature-complete plugin following the +2 minor release deprecation policy.
+	StabilityBeta StabilityLevel = "Beta"
+	// StabilityStable indicates a production-grade plugin guaranteed to be backwards compatible.
+	StabilityStable StabilityLevel = "Stable"
+)
+
+// ExperimentalPluginsFeatureGate is the feature gate name used to enable Alpha and Beta plugins.
+const ExperimentalPluginsFeatureGate = "experimentalPlugins"
```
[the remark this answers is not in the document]
LukeAVanDrie: +1

pkg/epp/framework/interface/plugin/stability.go
```diff
@@ -0,0 +1,56 @@
...
+
+// PluginMetadata holds stability and lifecycle metadata for a registered plugin type.
+type PluginMetadata struct {
+	Type               string
+	Stability          StabilityLevel
+	Deprecated         bool
+	DeprecatedIn       string
+	ScheduledRemovalIn string
```
LukeAVanDrie, line 37: Non-blocking: `Deprecated`, `DeprecatedIn`, `ScheduledRemovalIn`, and `ReplacementType` can't be set through any registration API (`Register` only takes stability), so the deprecation warning in `instantiatePlugins` is unreachable today. 

Either trim these fields until there's a way to set them, or add something like a `RegisterDeprecated` variant in this PR so the warning is live.

pkg/epp/datalayer/data_graph.go
```diff
@@ -139,6 +140,43 @@ func CreateMissingDataProducers(ctx context.Context, defaultProducerRegistry map
...
+	producerKeyMap := make(map[string]plugin.ProducerPlugin)
+	for _, p := range plugins {
+		if producer, ok := p.(plugin.ProducerPlugin); ok {
+			for key := range producer.Produces() {
+				producerKeyMap[key.String()] = producer
```
LukeAVanDrie: With multiple producers of the same key, last writer wins, so a violation through an earlier producer is missed or attributed to the wrong producer. `map[string][]plugin.ProducerPlugin` and checking each producer keeps it exhaustive.

LukeAVanDrie (dismissed):
One follow-up... the Alpha gate is only enforced in `instantiatePlugins`, so plugins auto-created via `registerDefaultPlugin` or `CreateMissingDataProducers` bypass it. No Alpha default producers exist today, so probably just a TODO.

pkg/epp/datalayer/data_graph.go
liu-cong: good call, maybe we should not validate this, but purely rely on the feature gate.

liu-cong:
@LukeAVanDrie @shmuelk I removed the dependency validation  on lower stability plugins, and use the feature gate as the single gating mechanism. Let me know what you think.

pkg/epp/framework/plugins/flowcontrol/eviction/filtering/sheddable.go
```diff
@@ -28,10 +28,6 @@ import (
 // This uses the project-wide definition of sheddable from requtil.IsSheddable.
 const SheddableFilterType = "sheddable-eviction-filter"
 
-func init() {
-	plugin.Register(SheddableFilterType, SheddableFilterFactory)
```
liu-cong, line 32: moved to runner.go

pkg/epp/framework/plugins/flowcontrol/eviction/ordering/priority_time.go
```diff
@@ -27,10 +27,6 @@ import (
 // breaking ties by newest dispatch time (least KV-cache investment).
 const PriorityThenTimeOrderingType = "priority-then-time-eviction-order-policy"
 
-func init() {
```
liu-cong, line 30: moved to runner.go

cmd/epp/runner/runner.go
liu-cong: Added them in runner.go and marked them alpha

pkg/epp/framework/interface/plugin/stability.go
liu-cong: Used these for deprecated plugins.

shmuelk:
@liu-cong The hermetic integration tests are failing. Partly to the use of an Alpha level plugin without your new feature gate.

cmd/epp/runner/runner.go
```diff
@@ -543,117 +550,148 @@ func setupDatastore(ctx context.Context, epFactory datalayer.EndpointFactory,
...
+	fwkplugin.Register(attrgpu.DCGMExtractorType, fwkplugin.StabilityAlpha, extdcgm.DCGMExtractorFactory)
+
+	// scheduling & profile handler plugins
+	// Beta
+	fwkplugin.Register(prefix.PrefixCacheScorerPluginType, fwkplugin.StabilityBeta, prefix.PrefixCachePluginFactory)
+	fwkplugin.Register(maxscore.MaxScorePickerType, fwkplugin.StabilityBeta, maxscore.MaxScorePickerFactory)
+	fwkplugin.Register(random.RandomPickerType, fwkplugin.StabilityBeta, random.RandomPickerFactory)
+	fwkplugin.Register(weightedrandom.WeightedRandomPickerType, fwkplugin.StabilityBeta, weightedrandom.WeightedRandomPickerFactory)
+	fwkplugin.Register(single.SingleProfileHandlerType, fwkplugin.StabilityBeta, single.SingleProfileHandlerFactory)
+	fwkplugin.Register(headerprofile.HeaderProfileHandlerType, fwkplugin.StabilityAlpha, headerprofile.HeaderProfileHandlerFactory)
```
elevran: `headerprofile.HeaderProfileHandlerType` is registered here under the "Beta" heading but tagged `StabilityAlpha`, and registered again below under "Alpha" (~L615). While the second call silently overwrites the first , suggest deleting the stray line here and keeping only the one under the Alpha section.

pkg/epp/framework/plugins/README.md
```diff
@@ -2,7 +2,28 @@
...
 
+## Plugin Stability Levels
+
+Every plugin in `llm-d-router` is assigned a **Stability Level** upon registration (`cmd/epp/runner/runner.go` is the single source of truth for in-tree plugin stability):
+
+| Stability Level | Lifecycle & Backwards Compatibility Guarantees | Command Line Flag Requirement |
+|---|---|---|
+| **Alpha** | Experimental features under active development. No backwards-compatibility guarantees (parameters or behavior may change anytime). | Requires `--allow-experimental-plugins` CLI flag. |
+| **Beta** | Feature-complete and enabled by default. Backwards-compatible within current version; subject to a +2 minor version deprecation policy before removal. | Allowed by default (no CLI flag required). |
+| **Stable** | Production-grade and fully backwards-compatible across minor releases. Breaking changes only on major version bumps. | Allowed by default (no CLI flag required). |
```
elevran, line 13: nit: this table documents Stable as a real tier with its own guarantees, but no in-tree plugin in `runner.go` is currently registered as `StabilityStable`. I think we can promote a few long-lived plugins (e.g. single, maxscore) or note that Stable isn't populated yet.

cmd/epp/runner/runner.go
```diff
@@ -441,6 +443,11 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 
 	scheduler := scheduling.NewSchedulerWithConfig(r.schedulerConfig)
 
+	if err := fwkplugin.ValidatePluginStability(r.PluginHandle, opts.AllowExperimentalPlugins); err != nil {
+		setupLog.Error(err, "Plugin stability validation failed")
```
elevran, line 447: No test drives the real startup path (`setup`/`runWithFileDiscovery`) with an Alpha plugin configured and `--allow-experimental-plugins` left at its default `false`, asserting startup fails. Existing tests call `ValidatePluginStability` directly in isolation. A test through the actual CLI-flag-to-runner path would catch a future regression where the flag stops reaching the check.

elevran (commented):
Solid implementation of the plugin stability lifecycle from #1912 - the Alpha/Beta gating logic, CLI flag wiring, and validation placement (after all plugin init phases, covering configured, default, and auto-created producer plugins) are all correct and well tested at the unit level. A few small things to clean up before merge.

Two docs outside this diff are now stale and worth a follow-up:
- docs/create_new_filter.md:172 - `plugin.Register(bylabel.LabelSelectorFilterType, bylabel.SelectorFactory)` uses the old 2-arg signature. Following this guide to add a new plugin will fail to compile after this PR. Worth updating to `plugin.Register(type, stability, factory)` plus a note on choosing Alpha/Beta/Stable.
- docs/plugins-vs-builtins.md:20 - same stale 2-arg `plugin.Register(type, factory)` reference.

pkg/epp/config/loader/configloader.go
```diff
@@ -172,11 +177,6 @@ func InstantiateAndConfigure(
 		return nil, fmt.Errorf("scheduler config build failed: %w", err)
 	}
 
-	featureGates, err := loadFeatureConfig(rawConfig.FeatureGates)
```
roytman, line 175: nit: This code block was moved to the beginning of the method, but `featureGates` is only used at line 189, so it can remain as it was.

pkg/epp/framework/plugins/README.md
liu-cong: I will note that currently no stable plugins yet. I think we can promote a few when we actually declare 1.0 launch.

elevran (dismissed):

elevran (dismissed):

ahg-g (approved):

PR #2075, opened 2026-07-18
feat(sidecar): support data parallelism with the offloading P2P tier

### LINKED ISSUE #1889: feat(sidecar): DP-aware support for offloading connector (wide-EP P2P)
## Background

PR #1888 added the `offloading` connector (`--kv-connector=offloading`), which
uses vLLM's `OffloadingConnector` and a P2P (ZMQ/TCP) secondary tier for KV
transfer without RDMA. That PR explicitly rejects `--data-parallel-size > 1`
at startup because wide-EP pods would have every DP rank bind the same
`POD_IP:<p2p-connector-port>`, causing a port conflict.

## Problem

With DP=N, each rank needs a distinct port: `base_port + dp_rank_local`.
The decode leg must receive the exact rank's port as `remote_port` so it pulls
KV from the correct rank. The sidecar is the only entity that knows which rank
was selected — the EPP encodes the DP rank in the routing header as part of
its DP-aware pick.

## Proposed design

Following the convention already used by the MoRI-IO parallel-dispatch path
(sidecar names the rank, vLLM owns port layout):

- vLLM: `TieringOffloadingSpec` P2P tier binds `base_port + dp_rank_local`
  instead of a fixed port. Optionally: expose base port as a vLLM env var
  (analogous to NIXL's port constant) so neither side needs to negotiate it.
- Sidecar: read the selected DP rank from the EPP routing header, compute
  `remote_port = base_port + rank`, inject into the decode leg's
  `kv_transfer_params.prefill.remote_port`.

The sidecar change is straightforward once the vLLM-side port layout is
defined. The vLLM change is the prerequisite.

## Dependency

The vLLM-side change (per-rank port binding in `TieringOffloadingSpec`) needs
to land first. cc @liranschour

### LINKED ISSUE #2074: Lift the offloading/DP>1 sidecar restriction now that vLLM binds per-DP-rank P2P ports
The sidecar rejects `--kv-connector=offloading` with `--data-parallel-size > 1` at startup (`pkg/sidecar/proxy/options.go`), and `docs/disaggregation.md` documents the restriction as "every DP rank would bind the same `POD_IP:<p2p-connector-port>`. DP-aware support is not yet implemented." The engine-side rationale is no longer true on vLLM main:

- vllm-project/vllm#47636 (merged 2026-07-15) binds the P2P control socket per DP rank (`base_port + data_parallel_index`) and adds `VLLM_P2P_SIDE_CHANNEL_HOST`/`VLLM_P2P_SIDE_CHANNEL_PORT`.
- vllm-project/vllm#47987 (merged 2026-07-12) gives each DP replica a distinct shared-memory offload region (`instance_id_dp{index}`).

So a wide-EP pod no longer collides on the bind, and the validation blocks a configuration the engine now supports.

Lifting the restriction is more than removing the check:

1. The relaxed validation applies only when the deployed engine includes the per-rank port scheme (>= merge commit `615834ee`), so the support floor needs to be stated or detected.
2. The sidecar injects `remote_port` for P2P pulls. Against a DP > 1 source pod, the serving port is now rank-dependent; single-endpoint serving across ranks is vllm-project/vllm#48021 (open). Pulls targeting DP > 1 sources should be gated on that, or the port selection needs to be rank-aware.
3. `docs/disaggregation.md` needs the restriction paragraph rewritten to the new support matrix.

Suggested sequencing: relax the validation and update the docs together with the first supported engine bump that includes vllm#47636/#47987, and note the DP > 1 source-pull dependency on vllm#48021 until it merges.

nilig:
**What type of PR is this?**
/kind feature

vLLM binds DP rank `r`'s P2P tier on `<p2p-connector-port>+r` and gives each replica a distinct shared-memory offload region (vllm-project/vllm#47636, vllm-project/vllm#47987), so the sidecar's startup rejection of `--kv-connector=offloading` with `--data-parallel-size > 1` guards a configuration the engine now supports. This lifts the rejection and makes the injected `remote_port` rank-aware: rank `r` is served on `<base port>+r`, so the matching P2P port is `--p2p-connector-port` + `r`. Ports outside the rank range fall back to rank 0, so a stale or foreign header degrades to a lookup miss instead of an error. Assumes pods in the pool share the sidecar port layout and DP size; requires an engine with the per-rank port scheme (documented in `docs/disaggregation.md`).

Fixes #2074
Fixes #1889

**Test plan**:
- [x] Rank derivation: offsets, range fallbacks, unparsable ports, and propagation to per-rank server clones
- [x] Validation accepts `offloading` with `--data-parallel-size > 1`
- [x] Cluster (2-pod P/D rig, sidecar built from this branch): DP=1 header-driven pull unchanged - full 8K-token pull through the sidecar; DP=4 injects `remote_port` 7779 for a rank-2 source header (trace-verified), no base-port attempts

```release-note
The routing sidecar supports data parallelism with the offloading P2P tier: the injected P2P port is offset by the target endpoint's DP rank, and --kv-connector=offloading no longer rejects --data-parallel-size > 1 (requires vLLM with per-DP-rank P2P ports).
```


roytman:
@nilig , is this PR related to #1889

nilig:
Yes - #1889 is the design issue for this; it was blocked on the vLLM side binding per-rank P2P ports. That landed in vllm-project/vllm#47636 / vllm-project/vllm#47987, tracked in #2074, and this PR is the sidecar half: `remote_port` is offset by the selected endpoint's DP rank. Added `Fixes #1889` to the description so both close.

pkg/sidecar/proxy/connector_p2p.go
```diff
@@ -243,10 +247,41 @@ func (s *Server) p2pSourceParams(sourceHostPort string) map[string]any {
...
+	if err != nil {
+		return base
+	}
+	port, err := strconv.Atoi(portStr)
+	if err != nil {
+		return base
+	}
+	rank := port - s.dpBasePort
+	if rank < 0 || rank >= s.config.DataParallelSize {
+		return base
```
roytman, line 287: nit: add at least debug messages here and above where it checks `targetHostPort` and port. 
```go
if rank < 0 || rank >= s.config.DataParallelSize {
		s.logger.V(logging.DEBUG).Info("target port outside DP rank range, using rank-0 P2P port",
			"target", targetHostPort, "base_port", s.dpBasePort, "dp_size", s.config.DataParallelSize)
		return base
	}
```

pkg/sidecar/proxy/connector_p2p.go
```diff
@@ -243,10 +247,41 @@ func (s *Server) p2pSourceParams(sourceHostPort string) map[string]any {
...
+	}
+	return base + rank
```
roytman, line 289: as Copilot mentioned, it can be outside the legal port number.

pkg/sidecar/proxy/connector_p2p.go
nilig: Added `DEBUG` logging for each fallback path in a6810714.

pkg/sidecar/proxy/connector_p2p.go
nilig: This is validated in ecc31ebf: `Validate()` rejects configurations where `P2PConnectorPort + DataParallelSize - 1` exceeds 65535, with boundary coverage for the highest valid and first invalid rank ports.

roytman (dismissed):

vMaroon (approved):

PR #2076, opened 2026-07-19
fix: sync deprecated flow control env var into feature gate map

### LINKED ISSUE #2065: `ENABLE_EXPERIMENTAL_FLOW_CONTROL_LAYER` doesn't actually enable Flow Control
**What happened**:
Setting `ENABLE_EXPERIMENTAL_FLOW_CONTROL_LAYER=true` looks like it works (you get the "this is deprecated" log line) but it doesn't actually turn on the Flow Control admission layer. `initAdmissionControl` still picks `LegacyAdmissionController` no matter what, and there's no `flow_control_*` series showing up in `/metrics` either.

Since this env var isn't removed yet, setting it to `true` should enable Flow Control the same way `featureGates: [flowControl]` in the config file already does.

1. Start the EPP with just `ENABLE_EXPERIMENTAL_FLOW_CONTROL_LAYER=true` set, nothing in the config's `featureGates`.
2. Send some traffic, check `/metrics` for `flow_control_*`, or grep the logs for "Initializing experimental Flow Control layer".
3. Neither ever shows up — it's silently running Legacy the whole time.

**Environment**:
- llm-d-router version (use `git describe --tags --dirty --always` if you built from source, or specify the tag if you used a tagged version or image): main
- Install tools:
- Others:

elinacse:
initAdmissionControl reads r.featureGates, which is populated once during parseConfigurationPhaseOne and never revisited. The deprecated `ENABLE_EXPERIMENTAL_FLOW_CONTROL_LAYER` env var was only applied later, in `parseConfigurationPhaseTwo`, and only patched `rawConfig.FeatureGates` — never `r.featureGates`. So setting the env var logged the deprecation warning but left the EPP silently running the legacy admission controller instead of Flow Control.

As flagged in review, `ENABLE_EXPERIMENTAL_FLOW_CONTROL_LAYER` was deprecated when GIE 1.5.0 introduced `featureGates: [flowControl]`, and that replacement has been in since router 0.8.0. Rather than fix the env var to work again, this PR removes it entirely ahead of the 0.10.0 release. Use `featureGates: [flowControl]` in the EndpointPickerConfig instead.

Fixes #2065


shmuelk:
The environment variable ENABLE_EXPERIMENTAL_FLOW_CONTROL_LAYER was deprecated in GIE 1.5.0 in favor of the mentioned featureGate.

GIE 1.50. was merged into the llm-d-router in release 0.8.0. I would think that with the upcoming 0.10.0 release the deprecated environment variable ENABLE_EXPERIMENTAL_FLOW_CONTROL_LAYER can be simply deleted.

elinacse:
@shmuelk removed ENABLE_EXPERIMENTAL_FLOW_CONTROL_LAYER. please review it .

LukeAVanDrie (approved):
Thanks @elinacse!

Heads up: this conflicts with #2180 (flip the `flowControl` gate on by default). #2180 rewrites the deprecation log message inside `applyDeprecatedEnvFeatureGate`, which this PR deletes, so` git merge-tree` reports a content conflict in `cmd/epp/runner/runner.go`. `feature_gate_test.go` merges cleanly.

No coordination needed beyond that. Whichever lands first, the other rebases, and if this one goes first I will just drop my change to that function since it goes away with the variable.

elinacse:
@LukeAVanDrie Thanks for the review . can you help in merging this PR?

LukeAVanDrie:
> @LukeAVanDrie Thanks for the review . can you help in merging this PR?

I cannot, you will need an approval from someone in llm-d/router-maintainers. I only have perms for review `pkg/epp/flowcontrol/...` and its plugin impls and fraemwork types under `https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/plugins/flowcontrol/...` and `https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/interface/flowcontrol/...`

ahg-g (approved):

PR #2077, opened 2026-07-19
feat(Coordinator) Support /inference/v1/generate user messages

### LINKED ISSUE #1894: feat(Coordinator) Support /inference/v1/generate user messages
Coordinator uses `/inference/v1/generate` vLLM interface but doesn't expect to receive these messages.
It should be extended.

revit13:
**What type of PR is this?**
/kind feature

Adds Coordinator support for a new /inference/v1/generate endpoint that accepts pre-tokenized "generate-format" requests (raw token_ids plus multimodal features), alongside the existing /completions and /chat/completions chat-style endpoints.

Changes by file

pkg/coordinator/server/server.go — Registers the new route:
r.Post(gateway.DefaultGeneratePath, s.handleInference)  // "/inference/v1/generate"
It reuses the existing handleInference handler, so the request flows through the same pipeline.

pkg/coordinator/steps/render.go — Adds a third branch to RenderStep.Execute. When OriginalPath is the generate path, it dispatches to a new executeGenerate, which:
- Extracts and validates token_ids into reqCtx.TokenIDs (with the existing token-limit check).
- Parses the features map into reqCtx.MultimodalEntries (with the placeholder-limit check).
- toIntSlice is moved from render.go into utils.go so both executeCompletions (token-array prompt path) and the new extractTokenIDs can share it.

pkg/coordinator/steps/utils.go — New parsing helpers for the generate wire format:
- toIntSlice (relocated) — []any → []int, non-negative integers only.
- anyToNonNegativeInt — single numeric JSON value → non-negative int.
- extractTokenIDs — validates token_ids is present, an array, non-empty; returns ErrBadRequest otherwise.
- extractMultimodalEntries — builds []MultimodalEntry from the parallel mm_hashes / mm_placeholders / kwargs_data slices under features. Notable contract:
  - mm_hashes and mm_placeholders are required and must be equal length.
  - kwargs_data is optional (second commit, "Make kwargs_data optional"): absent means every item resolves from the encoder cache by hash (cache-hit request → KwargsData == ""). When present it must be parallel-length, but individual items may be null (mixed cache-hit batch → "").
  - Text-only requests (no image hashes) return nil entries.

Tests added

- handlers_test.go — TestHandleInference_GenerateRouteRegistered: the new route is wired up (not 404/405).
- render_test.go — text-only, multimodal, multiple-images, and missing-token_ids (expects ErrBadRequest) cases.
- utils_test.go — TestExtractTokenIDs and TestExtractMultimodalEntries covering the new extraction/coercion helpers (empty/missing/non-array token_ids, mismatched slice lengths, null kwargs items, text-only).
-

Fixes #1894 

```release-note
Coordinator now supports the `/inference/v1/generate` endpoint for user-issued generate requests.
```


pkg/coordinator/server/handlers_test.go
```diff
@@ -233,3 +235,14 @@ func TestHandleInference_OverlongRequestIDIsRejected(t *testing.T) {
 		t.Fatalf("overlong request_id must not leak to the client: %q", rec.Body.String())
 	}
 }
+
+func TestHandleInference_GenerateRouteRegistered(t *testing.T) {
```
roytman: Worth turning it into a table-driven test over all five registered routes (PathChatCompletions, PathCompletions, DefaultGeneratePath POST; /healthz, /readyz GET) instead of just the one.

pkg/coordinator/steps/encode.go
```diff
@@ -169,6 +169,8 @@ func (s *EncodeStep) buildEncodeTokenIDs(fullTokenIDs []int, entry pipeline.Mult
 	placeholderTokenID := 0
 	if len(fullTokenIDs) > 0 {
 		bos = fullTokenIDs[0]
+		// Only the upper bound is checked here; offset >= 0 is guaranteed by
```
roytman: nit: `buildEncodeTokenIDs` is shared by all three paths (`generate`, `completions`, `chat completions`), but `extractMultimodalEntries` only populates `MultimodalEntries` for the generate path. For `chat/completions`, Placeholder comes from the internal render service's JSON response (render.go:264), not from `extractMultimodalEntries`. 
The guarantee holds in practice because that response is server-generated and trusted, but the comment as written overclaims a specific source that isn't universally true. 
Consider rewording to something like: "offset >= 0 is guaranteed for all paths, either by extractMultimodalEntries or by the trusted render-service response" so the comment doesn't mislead a future reader auditing the other callers.

revit13:
btw, when tested against a real vLLM encoder. A token-in /inference/v1/generate with a cache-hit kwargs_data (null) for an mm_hash not already in the encoder's mm cache makes the encoder assert in preprocessing and drop the request, though the API still returns 200 OK:

[v1/engine/core.py:807] preprocess_add_request -> mm_receiver_cache.get_and_update_features
[multimodal/cache.py:660] assert mm_item is not None
AssertionError: Expected a cached item for mm_hash='deadbeef...deadbeef'
INFO: "POST /inference/v1/generate HTTP/1.1" 200 OK

reported in https://github.com/vllm-project/vllm/issues/49181

pkg/coordinator/steps/encode.go
```diff
@@ -169,6 +169,10 @@ func (s *EncodeStep) buildEncodeTokenIDs(fullTokenIDs []int, entry pipeline.Mult
 	placeholderTokenID := 0
 	if len(fullTokenIDs) > 0 {
 		bos = fullTokenIDs[0]
+		// Only the upper bound is checked here; offset >= 0 is guaranteed for all
+		// paths, either by extractMultimodalEntries (generate) or by the trusted
+		// render-service response (chat/completions). A negative offset would
+		// index out of range.
 		if entry.Placeholder.Offset < len(fullTokenIDs) {
```
roytman, line 185: unbounded `entry.Placeholder.Length` can crash the whole coordinator process, not just fail one request.

`anyToNonNegativeInt` (utils.go) only checks non-negative + int-representable, no upper bound. `max_total_placeholder_tokens` is the only gate and ships disabled by default (`config/coordinator/coordinator.yaml:167-170`). This `make([]int, 1+entry.Placeholder.Length)` runs inside the `g.Go(...)` closure at line 105 - `errgroup` deliberately does not recover panics from `f()`, and a panic in a non-request goroutine isn't caught by net/http's per-connection recovery. A single request with `length: 9007199254740992` panics or OOMs the whole process, taking down every in-flight request on the instance.

Suggest: enforce a hard ceiling on `offset`/`length` in `anyToNonNegativeInt`/`extractMultimodalEntries` independent of the optional config knob, and/or wrap this goroutine body with `recover()`.

pkg/coordinator/steps/render.go
```diff
@@ -134,6 +136,32 @@ func (s *RenderStep) Execute(ctx context.Context, reqCtx *pipeline.RequestContex
...
+	tokenIDs, err := extractTokenIDs(reqCtx.Body["token_ids"])
+	if err != nil {
+		return fmt.Errorf("render: %w", err)
+	}
+	if err := s.checkTokenLimit(len(tokenIDs)); err != nil {
+		return err
+	}
+	reqCtx.TokenIDs = tokenIDs
+
+	features, _ := reqCtx.Body["features"].(map[string]any)
```
roytman: **Important: malformed `features` is silently coerced to "text-only" instead of erroring.**

If `features` is present but not an object (a string, array, bool), this type assertion fails, `ok` is discarded, and `features` becomes `nil` - `extractMultimodalEntries(nil)` returns `nil, nil`, so the request silently proceeds as text-only. This directly contradicts the "fail loudly, don't silently coerce to absent" contract that `mmImageArray` documents and enforces one level down (utils.go). A client sending malformed `features` gets a 200 with their images silently dropped, no log, no error.

Suggest:
```go
rawFeatures, ok := reqCtx.Body["features"]
var features map[string]any
if ok && rawFeatures != nil {
    features, ok = rawFeatures.(map[string]any)
    if !ok {
        return fmt.Errorf("render: features must be an object, got %T: %w", rawFeatures, pipeline.ErrBadRequest)
    }
}
```

pkg/coordinator/steps/decode.go
```diff
@@ -88,17 +88,35 @@ func (s *DecodeStep) Execute(ctx context.Context, reqCtx *pipeline.RequestContex
...
+		reqCtx.Body[reqcommon.FieldKVTransferParams] = kvParams
 		if len(reqCtx.TokenIDs) > 0 {
 			reqCtx.Body["prompt"] = reqCtx.TokenIDs
 		}
+	default:
+		// The /inference/v1/generate engine reads transfer params only from
+		// sampling_params.extra_args; a top-level kv_transfer_params is ignored,
+		// so the decode worker never pulls the prefill KV over NIXL. Merge into
+		// the client's sampling_params to preserve max_tokens and other fields.
+		sampling, ok := reqCtx.Body[reqcommon.FieldSamplingParams].(map[string]any)
```
roytman, line 109: `sampling_params`/`extra_args` type mismatches are silently overwritten, discarding client-requested generation parameters.

Nothing upstream (render/prefill) validates the type of `sampling_params` for the generate format. If a client sends `"sampling_params": [1,2,3]` (client bug or malicious input), this falls into the `!ok` branch and silently replaces it with `{}` here at decode time - discarding `max_tokens`/`temperature`/etc. with no `logger.Error`, no debug log, no `ErrBadRequest`. The caller gets a 200 with backend-default sampling and no indication their parameters were dropped.

Suggest: validate `sampling_params`'s type once in `render.go`'s `executeGenerate` (return `ErrBadRequest` on mismatch, consistent with how `token_ids`/`features` are already validated there), or at minimum log when falling back to an empty map here.

pkg/coordinator/steps/utils.go
```diff
@@ -116,3 +134,194 @@ func coerceParamsMap(logger logr.Logger, v any, label string) map[string]any {
...
+		pMap, ok := rawPlaceholders[i].(map[string]any)
+		if !ok {
+			return nil, fmt.Errorf("mm_placeholders[%d] must be an object: %w", i, pipeline.ErrBadRequest)
+		}
+		// The non-negative guarantee here is load-bearing, not just input
+		// hygiene. EncodeStep.buildEncodeTokenIDs indexes fullTokenIDs[offset]
+		// (guarded only on the upper bound) and allocates make([]int, 1+length);
+		// a negative offset or length panics there. vLLM's own schema declares
+		// these as plain ints and accepts negatives, so this stays stricter
+		// deliberately. Do not relax it to a plain int parse.
```
roytman, line 313: **Test gap matching the encode.go finding above:** this comment correctly identifies that a negative `offset`/`length` would panic downstream, but no test in `utils_test.go` or `render_test.go` ever sends a negative `offset`/`length` through `extractMultimodalEntries`/`executeGenerate` to exercise this guard. Since the comment itself flags this as the safety-critical property ("Do not relax it to a plain int parse"), it's the highest-value missing test - if this guard is ever weakened, nothing catches it.

pkg/coordinator/steps/utils.go
```diff
@@ -116,3 +134,194 @@ func coerceParamsMap(logger logr.Logger, v any, label string) map[string]any {
...
+		if err != nil {
+			return nil, fmt.Errorf("invalid token at index %d: %v: %w", i, err, pipeline.ErrBadRequest)
+		}
+		out = append(out, n)
+	}
+	return out, nil
+}
+
+// anyToNonNegativeInt converts a single JSON-unmarshalled numeric value to a non-negative int.
+func anyToNonNegativeInt(v any) (int, error) {
```
roytman, line 174: `TestExtractTokenIDs` only covers `float64` variants (negative, non-integer, overflow) and a top-level non-array. No case exercises the `default:` branch here with a non-numeric array element, e.g. `token_ids: ["abc"]` or a non-string `mm_hashes[i]` / non-object `mm_placeholders[i]` element in `extractMultimodalEntries`. These are the malformed-JSON-type cases most likely to be hit by a genuinely malicious or buggy client.

pkg/coordinator/steps/render.go
```diff
@@ -326,28 +354,3 @@ func (s *RenderStep) checkPlaceholderLimit(entries []pipeline.MultimodalEntry) e
 	}
 	return nil
 }
```
roytman, line 386: nit: pre-existing but now more exploitable: two entries with adversarially large `Length` values summing past `math.MaxInt` wrap `total` negative, making `total > s.maxTotalPlaceholderTokens` false and silently bypassing this limit. This bug predates the PR, but `Length` previously came from an internal render-service response - now it's a raw integer supplied directly by the end user via `/inference/v1/generate`, so the one mitigation an operator can enable is itself bypassable by the same untrusted input it's meant to police. Worth a follow-up given the encode.go fix above will likely touch adjacent validation anyway.

pkg/coordinator/steps/utils.go
```diff
@@ -116,3 +134,194 @@ func coerceParamsMap(logger logr.Logger, v any, label string) map[string]any {
 		return nil
 	}
 }
+
+// toIntSlice converts a JSON-unmarshalled []any of numeric elements to []int.
+// Each element must be a non-negative integer represented as float64 or json.Number.
+// Callers wrap the returned error with their own field context.
```
roytman: nit: the comment says "Callers wrap the returned error with their own field context," but neither caller does: `extractTokenIDs` returns this unwrapped, and `executeCompletions` only prepends the generic `"render: %w"` step prefix - no field name (`token_ids` vs `prompt`) is ever added. An end user can't tell from the error which field was invalid.

Suggest either correcting the comment, or having callers actually add field context, e.g. `fmt.Errorf("token_ids[%d]: %w", ...)`.

docs/communication.md
```diff
@@ -604,7 +607,7 @@ EPP-Phase: prefill
 `kwargs_data` carries the same per-image base64 tensors from the render step (same values sent to the encode stage). Each blob is a msgpack-serialized `MultiModalKwargsItem` containing both `pixel_values` and `image_grid_thw` (and any other model-specific keys). The prefill worker needs `image_grid_thw` to compute mRoPE (multimodal Rotary Position Embedding) positional encodings for the visual tokens.
 
 > [!NOTE]
-> Due to a bug in the `/inference/v1/generate` implementation, the `kv_transfer_params` are not propagated as expected, so we will use a workaround:
+> Due to a bug in the `/inference/v1/generate` implementation, top-level `kv_transfer_params` and `ec_transfer_params` are not propagated to the engine: the endpoint reads transfer parameters only from `sampling_params.extra_args` (the top-level `ec_transfer_params` is used solely to echo back in the response). The coordinator nests both under `extra_args`:
```
roytman: This claims "the top-level `ec_transfer_params` is used solely to echo back in the response," but no example in this doc shows a response echoing `ec_transfer_params` back, and this is a claim about the external vLLM engine's behavior, not something the coordinator's own code demonstrates or tests. If this was confirmed against the real engine, worth a pointer to how; otherwise consider softening the phrasing so it doesn't read as a settled, verified fact.

docs/communication.md
```diff
@@ -626,11 +629,16 @@ EPP-Profile: prefill
     ]},
     "kwargs_data": {"image": ["<base64-encoded-pixel-tensor-1>", "<base64-encoded-pixel-tensor-2>"]}
   },
-  "ec_transfer_params": {
-    "abc123hash": {"peer_host": "10.0.0.1", "peer_port": 5501, "size_bytes": 2359296, "nixl_agent_metadata_b64": "TklYTA..."},
-    "def456hash": {"peer_host": "10.0.0.2", "peer_port": 5502, "size_bytes": 2359296, "nixl_agent_metadata_b64": "QWdlbnQ..."}
-  },
-  "sampling_params": {"max_tokens": 1, "extra_args": {"kv_transfer_params":{"do_remote_decode": true, "do_remote_prefill": false}}}
+  "sampling_params": {
```
roytman, line 666: I think we have to update the document on how the "generate" input request is processed. Currently, the document covers only `/v1/complitions` and `/v1/chat/complitions`

pkg/coordinator/steps/render.go
```diff
@@ -124,6 +123,9 @@ func (s *RenderStep) SetServiceAddress(addr string) {
 func (s *RenderStep) Name() string { return RenderStepName }
 
 func (s *RenderStep) Execute(ctx context.Context, reqCtx *pipeline.RequestContext) error {
+	if reqCtx.OriginalPath == gateway.DefaultGeneratePath {
```
roytman, line 126: Do we need it? The render provides tokenization, but `generate` gets tokens-in, so I doubt that we need this step

test/coordinator/e2e/coordinator/coordinator_test.go
```diff
@@ -47,35 +48,35 @@ var (
...
-		runCoordinatorPipeline([]byte(fmt.Sprintf(
+		runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
 			`{"model":%q,"messages":[{"role":"user","content":[{"type":"image_url","image_url":{"url":%q},"uuid":"image-0"},{"type":"text","text":"Describe what you see."}]}],"max_tokens":150}`,
 			modelName, inlineImageDataURI,
 		)), allSteps, 1)
 	})
 
 	ginkgo.It("routes a multimodal chat completion with one inline and one remote image end-to-end", func() {
-		runCoordinatorPipeline([]byte(fmt.Sprintf(
+		runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
```
roytman, line 79: do we have only chat/completions tests?

test/coordinator/e2e/coordinator/generate_endpoint_test.go
```diff
@@ -0,0 +1,142 @@
...
+func generateTokenIDs(images []genImage) []int {
+	if len(images) == 0 {
+		ids := make([]int, 20)
+		for i := range ids {
+			ids[i] = 1000 + i
+		}
+		return ids
+	}
+
+	const placeholderTokenID = 32000
```
roytman, line 112: please add a comment, about this value. (and that it's model; dependent)

test/coordinator/e2e/coordinator/generate_endpoint_test.go
```diff
@@ -0,0 +1,142 @@
...
+			// A cached entry leaves kwargs[i] nil (JSON null); the encoder
+			// resolves it from its own cache by hash.
+			if !img.Cached {
+				kwargs[i] = generateTestKwargs
+			}
+		}
+		body["features"] = map[string]any{
+			"mm_hashes":       map[string]any{"image": hashes},
+			"mm_placeholders": map[string]any{"image": placeholders},
+			"kwargs_data":     map[string]any{"image": kwargs},
```
roytman, line 86: so we don't have a test with real mm data?

pkg/coordinator/steps/render.go
revit13: The naming is misleading. executeGenerate doesn't tokenize; on the generate path the client sends token_ids and features directly, so there's no render-service call.

It's still needed, though, because it's the step that normalizes the client body into reqCtx.TokenIDs and reqCtx.MultimodalEntries - the typed fields the downstream steps read. Encode/prefill/decode build fresh outgoing bodies from those fields, not from the raw request body: e.g. encode fans out one sub-request per image using the parsed per-entry Hash/Offset/Length/KwargsData and indexes token_ids[offset] + allocates 1+length. It's also where validatePlaceholderBounds runs, guarding against an unbounded client-supplied length (memory exhaustion) that vLLM doesn't enforce on the generate endpoint. Removing the step would mean duplicating that parse + bounds check into every consumer.

I've added a doc comment on the function spelling this out. If the concern is that this ingestion doesn't belong in a step called render, I'm happy to split it into a dedicated parse/validate step and keep render for tokenization only - let me know if you'd prefer that.

test/coordinator/e2e/coordinator/coordinator_test.go
revit13: Not only chat/completions. The generate endpoint has its own end-to-end coverage in generate_endpoint_test.go (text-only, single-image, and a two-image case with one cached/null-kwargs_data entry), all going through the same runCoordinatorPipeline helper with gateway.DefaultGeneratePath. That's why the helper was parametrized on path.

test/coordinator/e2e/coordinator/generate_endpoint_test.go
revit13: Correct, the generate e2e uses synthetic stand-ins (kwargs_data is base64 "tensor", arbitrary token IDs/hashes) because /inference/v1/generate is tokens-in (render parses locally, no renderer call) and the sim runs in echo mode, so tensor content isn't validated regardless. The chat/completions e2e does use a real image through vllm-render. If we want real-mm coverage for generate, I can add a two-step test: render the real image via vllm-render, then feed the resulting token_ids+features into /inference/v1/generate.

revit13:
@roytman as we discussed, I changed the code to skip the encode step on the generate path.

roytman (dismissed):

roytman (commented):
@revit13 can you update `docs/communication.md` the encoder part that due to https://github.com/vllm-project/vllm/issues/46722, we do not support e-disaggregation for `/inference/v1/generate`  and skip encoding for the calls

docs/communication.md
```diff
@@ -952,8 +1002,8 @@ The coordinator uses the `EPP-Profile` HTTP header to identify the pipeline stag
 |-------------------|----------------------|---------------------------|
 | Encode            | `encode`             | `/v1/chat/completions` or `/inference/v1/generate` |
```
roytman: please remove `/inference/v1/generate`

roytman (approved):
Thank you, Revital.

elevran (approved):

PR #2081, opened 2026-07-19
test: harden port-bind race in EPP integration harness

### LINKED ISSUE #1066: [Flaky Test] hermetic integration tests fail with "Server failed to bind port"
**What happened**:

The hermetic integration tests under `test/integration/epp/` flake intermittently on CI with `Server failed to bind port 127.0.0.1:<port>` errors, causing `lint-and-test` to fail on unrelated PRs.

Observed in at least:

- PR #1058: `TestFullDuplexStreamed_KubeInferenceObjectiveRequest/Standalone-NoCRD/no_backend_pods_available#01`
  https://github.com/llm-d/llm-d-inference-scheduler/actions/runs/25718109453/job/75516805509
- PR #1043: `TestFullDuplexStreamed_GRPC_KubeInferenceObjectiveRequest/do_not_shed_requests_by_default`
  https://github.com/llm-d/llm-d-inference-scheduler/actions/runs/25633770069/job/75241972222

Failure signature (different sub-tests, same root cause):

```
Error:    Condition never satisfied
Messages: Server failed to bind port 127.0.0.1:<random-port>
```

Hermetic integration tests should pass deterministically on CI without flaking due to port allocation races.

Flaky — run `lint-and-test` repeatedly on a PR; intermittently a sub-test of `TestFullDuplexStreamed_*` fails with the bind-port error within its 10s budget.

Likely cause: the random-port allocation in the hermetic test fixture races with another process / parallel sub-test on the runner; the 10s bind retry budget expires before the port frees up.

Suggested follow-up:

- Audit the port-allocation helper in the hermetic harness; prefer a listen-then-pass-fd pattern over allocate-then-bind to avoid the race.
- Or: increase the retry/timeout budget on bind failures.

**Environment**:

- CI: GitHub Actions, `ubuntu-latest` runner (default `lint-and-test` workflow).
- Not environment-specific to a user setup; reproduces on CI only.

chethanuk:
**What type of PR is this?**

/kind test
/kind bug

The hermetic EPP integration tests intermittently fail with `Server failed to bind
port <addr>` (#1066). The harness allocates ports with `GetFreePort`, which closes
its listener before the server binds, leaving a TOCTOU window: another process (or a
concurrently torn-down test still releasing the port) can take the port between the
close and the server's `mgr.Start` bind. The old harness waited passively with
`require.Eventually` on the gRPC port and failed the whole test on any bind error,
so the race surfaced as a hard, hard-to-diagnose failure instead of being retried.

This hardens the harness (test-only; `GetFreePort`'s signature is untouched):

- `isAddrInUse(err)` classifies a bind failure as address-in-use, unwrapping to
  `syscall.EADDRINUSE` and falling back to the stringified message for errors that
  crossed a boundary and no longer unwrap.
- `RetryOnAddrInUse` runs the harness startup up to `portBindMaxAttempts` (5) times,
  retrying **only** on an address-in-use error and running a `reset` between attempts
  (`metrics.Reset`, to avoid controller-runtime's global-registry double-register
  panic). Any other error, or success, returns immediately; a genuine bind bug that
  never clears still fails loud after the budget.
- `WaitExtProcReady` replaces the passive wait: it returns ready when the ext-proc
  gRPC port accepts a connection, surfaces an early manager error so the caller can
  retry, and times out with a non-`EADDRINUSE` error so a server that never binds
  fails loud instead of looping.

```mermaid
flowchart TD
    A[NewTestHarness] --> B[RetryOnAddrInUse: attempt]
    B --> C[re-roll grpc/metrics/health ports]
    C --> D[start manager goroutine -> mgrErr]
    D --> E[WaitExtProcReady grpc port]
    E -->|TCP accept| F[ready: return nil]
    E -->|mgrErr early| G{isAddrInUse?}
    E -->|timeout| H[fail loud: non-EADDRINUSE]
    G -->|yes| I{attempts left?}
    G -->|no| H
    I -->|yes| J[reset: metrics.Reset] --> B
    I -->|no| K[fail: port bind failed after N attempts]
```

Scope is the test harness only (`test/integration/util.go`,
`test/integration/epp/harness.go`), plus table-driven unit tests for the new
helpers. Because it is a flake mitigation, elimination can only be confirmed by CI
over time rather than proven here; if maintainers prefer the listen-then-pass-fd
approach the issue also mentions over the retry-budget path taken here, happy to
discuss.

Fixes #1066

```release-note
NONE
```


elevran:
thanks @chethanuk - PR needs conflict resolution.

I see a few options to resolve the issue:
- retry on EADDRINUSE in the test harness only. Cheap, test-only, no production change. This is path chosen by this PR. Downside: mgr.Start() doesn't return a clean typed error for a bind failure per-runnable, so needs pattern-matching on error text ("address already in use") from the aggregated manager error, and to rebuild the whole manager with a fresh port set, since the port is baked into eppOptions before ctrl.NewManager is even constructed. 
- inject a live `net.Listener` instead of an int (port number). Removes the race for real, everywhere (prod and test), because nothing is ever closed and reopened. But it changes `runnable.GRPCServer's` signature and the option-passing chain across `options.go`, `runserver.go`, `runner.go`, all of which are production code, all currently plumbing int ports from CLI flags down. Doable, but a bigger diff for what is fundamentally a test-flakiness problem.
- FD passing - removes the race for real, but FD passing is non-trivial and would need to propagate through options, runnables's etc. Similar drawbacks as previous option passing a live listener.
- use `SO_REUSEPORT` : on Linux (need to confirm for Mac users...), `SO_REUSEPORT` lets multiple sockets bind the identical addr:port concurrently and load-balances between them. To enable, every socket sharing that port must set the flag, not just the first one. So this doesn't stay test-only either: the `net.Listen` in internal/runnable/grpc.go:38's would need to opt in via a `net.ListenConfig{Control: ...}` with `SO_REUSEPORT`/`SO_REUSEADDR`. It's a smaller production change than full listener-injection (one ListenConfig.Control callback in grpc.go, reusable for the health server too). The listener binding to the initial "0" port is only closed after the real server is confirmed up (or at test cleanup), eliminiating.

The last three options all eliminate the window instead of working around it. Each uses a different mechanism.
I think the `SO_REUSEPORT` option is a small, localized change. It closes the race for GRPC, health, and metrics ports without touching the CLI-flag/options plumbing that other options might need to rewire.

1. `net.ListenConfig` with SO_REUSEPORT+SO_REUSEADDR in internal/runnable/grpc.go (and wherever the metrics server binds, if that's also just `net.Listen`)
1. `GetFreePort` returns `(*net.Listener, error)` and callers hold the listener open until after harness startup completes, then close it in `t.Cleanup`.

chethanuk:
Thank you. Agreed on eliminating the window rather than working around it, and your critique of the retry is fair: it string-matches the aggregated manager error and rebuilds the manager per attempt. I dropped it, rebased onto main, and pushed the replacement.

Your item 2 is the fix, and it turns out it doesn't need the sockopt. While a listener holds `127.0.0.1:P`, an explicit rebind of P gets `EADDRINUSE`, and 2000 subsequent `bind(127.0.0.1:0)` draws never return P. That's a test in the PR now (`TestReserveListenerDefendsPort`) rather than a number you have to take on trust. So the reservation already defends itself. What was missing is only that the server re-derived the port instead of accepting the listener. That's what I implemented.

Two reasons I didn't go with `SO_REUSEPORT`:

1. Readiness stops being observable. With both sockets in the group and the reservation socket listening but never accepting, I ran 100 dials: all 100 completed the handshake, 54 reached the real server, and 46 sat in the reservation's accept queue. So a dial-based probe reports ready before the server is, which is the failure we're trying to remove. controller-runtime has no non-dial "has bound" signal either. Runnables in the `Others` group get a nil ready check that defaults to `return true` (`runnable_group.go:298`). It was a throwaway program, happy to attach it.
2. It changes production bind semantics. Two EPP processes in one netns would silently split ext-proc traffic instead of failing loudly. Scoping the flag to tests only avoids that, but then the test and production bind paths diverge.

One correction to myself: it's specifically a plain `net.Listen` reservation plus a `SO_REUSEPORT` server that can't coexist. Both with the flag does work, it's just the configuration with the two problems above.

On diff size, the listener rides on `ExtProcServerRunner`, a struct literal, not `Options`, so there's no CLI flag, no new dependency, and no build-tag split. `GRPCServerOnListener` sits beside `GRPCServer`, plus an optional field that is nil at every production construction site. That is +48/-23 over two files, mostly a shared helper. Same shape as `manager.Server.Listener` ("port reservation needs to happen in advance of this runnable starting", `pkg/manager/server.go:46`) and apiserver's `SecureServingOptions.Listener`.

Metrics and health turned out not to need a port at all. Nothing scrapes the metrics port in the hermetic tests (they read the in-process registry) and nothing dials the health port, so both are kernel-assigned now. Together with the ext-proc listener that leaves zero pre-selected ports. It also killed a bug I had missed: `GetFreePort` was called three times back to back, each closing before the next, so two calls could return the same port.

The soak was 10x clean on the branch, but 10x clean on `main` too, so it proves nothing and I won't present it as evidence. The argument here is structural.

`cmd/epp/runner/run_with_file_discovery_test.go:196` keeps its own port-based helper on purpose, so something still covers the production `net.Listen` path. That does make "Fixes #1066" a slight overclaim until it has its own issue. Happy to file one.

The commits are droppable individually. If you'd still rather see the `SO_REUSEPORT` version I'll build it, I'd just want to settle point 2 first.

elevran (approved):

elevran:
@chethanuk needs a rebase on main to pick up the Go vulnerability fix.

chethanuk:
Thanks @elevran I have rebased

Issue #2060, opened 2026-07-16
Flow control dispatch cycle receives empty endpoint list at r≥2, causing saturation=1.0 fallback


lioraron:
### Description

When running flow control with `concurrency-detector` and **2+ replicas (r≥2)**, the `flow_control_pool_saturation` metric is stuck at **1.0**, causing the dispatch cycle to stop dispatching requests entirely. Interactive TTFT p95/p99 degrades to 1.7–3.9 seconds (vs ~100ms at r=1).

This is a **different bug** from #1474 (which causes saturation to always return **0.0** when endpoints are present). At r≥2, the endpoint list passed to `Saturation()` appears to be **empty**, hitting the zero-capacity fallback:

```go
// detector.go:136-138
if totalCapacity == 0 {
    return 1.0
}
```

### Evidence

- **r=1**: `flow_control_pool_saturation` = 0.0 (consistent with #1474 — endpoints exist but InFlightLoad is always zero)
- **r=2**: `flow_control_pool_saturation` avg = 1.0 (endpoints missing from the list entirely)
- Same config, same GIE v1.5.0, same `maxConcurrency=1000` — only replica count differs

With `maxConcurrency=1000` and 2 endpoints, `totalCapacity` should be 2000 and saturation should be ~0.04 (or 0.0 per #1474). The 1.0 value is only possible when the endpoint list from `EndpointCandidates.Locate()` is empty or all endpoints have nil metadata.

### Impact

This is a **blocker for batch-gateway benchmarking** ([llm-d/llm-d-batch-gateway#491](https://github.com/llm-d/llm-d-batch-gateway/issues/491)). Flow control works well at r=1 (dispatch ordering gives −51% TTFT p99), but is completely broken at r≥2. We cannot extend benchmarks to multi-replica configurations until this is fixed.

### Possible causes

1. **Race condition in endpoint registration**: `Runtime.NewEndpoint()` returns `nil` if `collectors.LoadOrStore` finds an existing collector. With 2 pods reconciling rapidly, there may be a window where neither pod is stored in the datastore.
2. **CachedEndpointCandidates caching an empty result**: The 50ms TTL cache could cache an initial empty result if the dispatch cycle starts before pods are registered, and the cache never refreshes.
3. **Data layer collector failure**: If `collector.Start()` fails for a pod, `NewEndpoint()` returns nil and the pod is never stored.

### Reproduction

```yaml
# EPP config
plugins:
  - type: concurrency-detector
[...]

PR #2092, opened 2026-07-20
fix(datastore): retry dropped endpoint registrations instead of silently untracking pods

LukeAVanDrie:
**What type of PR is this?**

/kind bug

An endpoint upsert that overlaps an in-flight delete observes `Runtime.NewEndpoint` returning nil (the previous collector is still registered) while the pods-map entry is already gone. The datastore skipped storing the pod and the reconciler reported success, so the pod stayed untracked until the next pod event; for a stable Ready pod that event may never come. An empty datastore pods map makes the flow control saturation detectors fail closed, pinning `flow_control_pool_saturation` at 1.0 and halting dispatch (#2060).

```mermaid
sequenceDiagram
    participant Del as Delete path (PodDelete / resync)
    participant DS as Datastore
    participant R as Runtime (collector registry)
    participant Rec as PodReconciler (upsert)

    Del->>DS: PodDelete(pod)
    DS->>DS: pods.Delete(key)
    DS->>R: ReleaseEndpoint(ep)
    Note over R: dispatches EventDelete to extractors<br/>(collector still registered)
    Rec->>DS: PodUpdateOrAddIfNotExist(pod)
    DS->>DS: pods.Load(key): miss
    DS->>R: NewEndpoint(meta)
    R-->>DS: nil (collector still registered)
    DS->>DS: pods.Load(key): still miss
    DS-->>Rec: errRegistrationDropped
    Note over Rec: before this PR: nil swallowed, reconcile "succeeds",<br/>pod untracked until the next pod event.<br/>Now: error returned, controller-runtime requeues with backoff.
    R->>R: collectors.Remove(key), Stop()
    Rec->>DS: retry: PodUpdateOrAddIfNotExist(pod)
    DS->>R: NewEndpoint(meta)
    R-->>DS: endpoint
    DS->>DS: pods.Store(key, ep)
    DS-->>Rec: nil (tracked)
```

This PR makes the drop detectable and retryable:

- `upsertEndpoint` distinguishes the benign duplicate-start race (a concurrent upsert stored the entry; this call's metadata is applied through the update path) from a dropped registration, which is returned as an error wrapping a sentinel.
- `PodReconciler` propagates the error so controller-runtime requeues with backoff.
- `podResyncAll` propagates the error through `PoolSet`, and a `needsResync` flag makes the retried `PoolSet` re-run the resync (the pool is stored before resyncing, so the pool comparison alone would skip the retry).
- Discovery-source upserts log the drop; no requeue mechanism exists on that path.
- Tests: dropped-registration unit and reconciler tests, a concurrent add/remove completeness test, and a resync-retry regression test (fails without the `needsResync` flag).

Follow-up, deliberately not in this PR: have `Runtime.NewEndpoint` return a typed error distinguishing "collector already registered" from "collector failed to start", and evaluate closing the window in `ReleaseEndpoint` itself. The current release ordering (dispatch the delete event, then deregister the collector) is critical for extractor state, so a naive reorder is not safe.

Related to #2060. Closure is gated on a multi-replica re-test against main; see the issue thread.

**Release note**:

```release-note
Fixed a race where an endpoint added while its previous registration was still being deleted could be silently dropped from the EPP datastore, leaving the endpoint unroutable and pinning flow_control_pool_saturation at 1.0 until the next pod event. Dropped registrations are now retried.
```

### DIFF
--- a/pkg/epp/controller/pod_reconciler.go
+++ b/pkg/epp/controller/pod_reconciler.go
@@ -59,7 +59,12 @@ func (c *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		return ctrl.Result{}, fmt.Errorf("unable to get pod - %w", err)
 	}
 
-	c.updateDatastore(ctx, pod)
+	// Returning the error makes controller-runtime requeue with backoff, so an endpoint
+	// registration dropped by an upsert racing an in-flight delete is retried instead of
+	// leaving the pod untracked until the next pod event.
+	if err := c.updateDatastore(ctx, pod); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to update datastore for pod %s - %w", req.NamespacedName, err)
+	}
 	return ctrl.Result{}, nil
 }
 
@@ -91,16 +96,12 @@ func (c *PodReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(c)
 }
 
-func (c *PodReconciler) updateDatastore(ctx context.Context, pod *corev1.Pod) {
+func (c *PodReconciler) updateDatastore(ctx context.Context, pod *corev1.Pod) error {
 	logger := log.FromContext(ctx)
 	if !podutil.IsPodReady(pod) || !c.Datastore.PoolLabelsMatch(pod.Labels) {
 		logger.V(logutil.DEBUG).Info("Pod removed or not added")
 		c.Datastore.PodDelete(pod.Name)
-	} else {
-		if c.Datastore.PodUpdateOrAddIfNotExist(ctx, pod) {
-			logger.V(logutil.DEFAULT).Info("Pod added")
-		} else {
-			logger.V(logutil.DEFAULT).Info("Pod already exists")
-		}
+		return nil
 	}
+	return c.Datastore.PodUpdateOrAddIfNotExist(ctx, pod)
 }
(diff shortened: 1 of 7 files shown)


ahg-g (commented):
Why not address the race condition itself? I am surprised that we don't do that for all things related to pod and endpoint updates to the datastore.

LukeAVanDrie:
> Why not address the race condition itself? I am surprised that we don't do that for all things related to pod and endpoint updates to the datastore.

Mostly because the nil from `NewEndpoint` isn't only the race. `collector.Start` failing produces the same thing with no concurrency at all, so the caller needs a retry path either way.

Fixing the race at the source is also less simple than it looks. `ReleaseEndpoint` intentionally dispatches `EventDelete` before deregistering the collector. If we flip that, a racing `NewEndpoint` can register a fresh collector whose state the stale `EventDelete` then tears down. Locking in the datastore instead means holding a lock across plugin dispatch and a blocking collector `Stop`.

We probably need per-key serialization in the collector registry plus a typed error from `NewEndpoint`. This is a datalayer change I'd rather do separately (it's the follow-up listed in the description; happy to file the issue).

With the retry, losing the race costs one requeue instead of a permanently missing endpoint.

If you have any other ideas here that are less complex, please let me know. To my best knowledge this race window is very small and unlikely to be hit in practice.

ahg-g (approved):

PR #1360, opened 2026-05-26
feat(Flow Control): demand-driven in-flight eviction for flow control

RishabhSaini:
resolves: #1119 

  - Adds SLO-aware eviction of in-flight requests when higher-priority requests are blocked by pool saturation
  - Opt-in via `enableEviction: true` in flow control config, disabled by default

  ## When eviction fires

  The dispatch cycle already detects HoL blocking (`saturation >= usageLimit`). When this occurs, we check the tightest SLO/TTL deadline across queued flows using:

  remaining <= elapsed × (saturation / usageLimit)

  - **remaining**: time until deadline (`x-llm-d-slo-ttft-ms` header, or `enqueueTime + TTL` fallback)
  - **elapsed**: how long the head-of-queue request has waited
  - **saturation / usageLimit**: overload ratio, heavier overload triggers eviction sooner

  Tight SLOs evict faster. High saturation evicts faster. There are no static margins. It self-tunes from observable state. Debounced per priority band (default 100ms).

  ## How many requests evicted

  `min(queuedCount, evictable)`: proportional to demand, capped at available sheddable requests. Only evicts when evictable priority is strictly lower than the blocked band (no same-priority churn). 

Uses existing `PriorityThenTimeOrdering` (lowest priority first, newest dispatch first to minimize KV-cache waste) and `SheddableFilter` (priority < 0 only).

Issue #1119, opened 2026-05-14
[Flow Control] Eviction of in-flight dispatched requests


ahg-g:
xref: https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2632

## Summary

Add the ability to evict in-flight requests already dispatched to model servers, freeing KV-cache blocks and compute for higher-priority or SLO-urgent requests waiting in flow control queues.

## Motivation

Today flow control can only reject requests *before* dispatch. Once a sheddable request is consuming KV blocks on vLLM, there is no way to reclaim those resources even if high-priority requests are queued behind it. Eviction closes this gap.

### Tracking in-flight requests

Track dispatched requests using existing RequestControl hooks in a queue with pluggable policies to filter for evictable requests from in-flight and order them.

### Eviction

Signal for the Eviction Signal is a combination of:
- PreDispatched Flow Control Queues: Signals if there is a sufficient amount of high priority with urgent SLO request waiting  
- SaturationDetection: Signals existing load on the InferencePool

These signals will also help determine the total number of in-flight requests to be evicted in order to free up capacity for urgent high-priority requests

Levels of Eviction:
- `CompleteEvictionPolicy`: Evicted Requests have their KV Cache on the Pool freed
- `PartialEvictionPolicy`: KV Cache of evicted requests are offloaded to be resumed later when retried  

### Post-Eviction Request Handling

Retry Mechanism:
- `DumpEvictedRequestPolicy`: Evicted requests are dumped with no retry, leaving it to the user to enqueue again
- `RetryEvictedRequestPolicy`: Evicted requests are added to the back of the flow queues with exponential backoffs and eventual hard limit

shmuelk:
@ahg-g If I understand this issue, you are looking for the ability to "evict/cancel" requests that have been sent to a model server to help with saturation events.

Once a request is sent to a model server by Envoy, how can the EPP evict/cancel an inference request? I don't think you can even "evict/cancel" a request that is streaming its ouput and the streaming has started.

If you are talking about another check after the EPP's scheduling has run before a result is sent to Envoy, that is a different story. With the noted exception of the tokenization producer plugin, most plugins run very quickly and it's hard to see the benefit here.

LukeAVanDrie:
Design proposal is up in #2061; implementation PRs (simulation harness, controller and wiring) follow once the design settles, but I will also share my drafts shortly.

[...]

PR #2093, opened 2026-07-20
feat(flowcontrol): demand-driven in-flight eviction

LukeAVanDrie:
**What type of PR is this?**

/kind feature

Implements the reclamation controller from the in-flight eviction design (#2061, `docs/flow-control-eviction.md`): flow control can now evict in-flight sheddable requests (priority < 0) when higher-priority requests are queued behind pool saturation, converting the priority-inversion window from unbounded natural churn time to a bounded revocation round-trip.

The merged victim filter/ordering policies and the ext_proc eviction channel are used as-is; the merged RequestEvictor is reused with small tracker extensions (commit 1), and the wiring is adapted from draft PR #1360; the wiring commit carries a Co-authored-by trailer for @RishabhSaini. Maintainers, please retain it in the squash-merge message. This PR adds the piece #1119 left open, deciding when and how much to evict:

- Trigger: on Head-of-Line blocking, scan blocked bands for queued demand strictly above the victim head's priority; bands with unattainable (zero) ceilings are excluded. The demand band's priority bounds victim priority for the whole decision, so multi-revocation decisions cannot churn the demand band itself.
- Sizing: a saturation-gauge deficit against a mean-footprint credit, with pending-reclaim debits so capacity in the revoking/reclaiming stages is never counted twice.
- Pacing: confirmation-gated. New decisions wait until every evicted request's stream has terminated plus a grace covering the sensor's confirmation-to-visibility lag, with a timeout so a hung stream cannot wedge reclamation.
- Configuration: one public field, `flowControl.enableEviction` (default false). The grace is derived at wiring time from the selected saturation detector (dispatch-tick scale for the concurrency detector; refresh interval + staleness threshold + an engine reclaim budget for scraped sensors); remaining parameters are internal. The design doc's Configuration section carries the rationale.
- Observability: revocations issued and terminal outcomes (confirmed, timed_out), reclaim target, pending reclaim, and issue-to-confirmation latency.

Review guide: the diff is +1365/-38 across 22 files, but the decision logic is one new file, `pkg/epp/flowcontrol/controller/internal/reclamation.go` (~330 lines, half of it the invariants as comments); review it against the design doc's Trigger/Sizing/Pacing sections. The rest splits into tracker extensions on the merged `RequestEvictor` (commit 1), metrics (mechanical), and wiring (commit 3, adapted from #1360). Tests are ~45% of the diff. The three commits are ordered to be read in sequence.

A closed-loop benchmark reproducing the design doc's Validation results is published alongside this PR as a companion draft (necessarily based on `main`, since GitHub cannot base a PR on a fork branch; its reviewable diff is its final commit). Per the design doc, an end-to-end measurement of issue-to-abort latency is part of the acceptance criteria; eviction remains experimental (within the experimental flow control layer) until that measurement exists.

Part of #1119 (partial eviction and retry orchestration are deferred per the design doc's non-goals; close the issue here if we consider the core ask delivered)

**Release note**:
```release-note
Flow control can evict in-flight sheddable requests (priority < 0) to reclaim pool capacity when higher-priority requests are blocked by saturation. Disabled by default; enable with flowControl.enableEviction. When enabled, sheddable requests may be terminated mid-stream with HTTP 429.
```

**Test plan** (new tests; functionality verified):

- [x] Reclamation sizing across the decision table, including the exact-boundary (zero-deficit) case and zero-ceiling demand exclusion
- [x] Pacing gate lifecycle: closed while outstanding, cooling through the grace, timeout retirement with late-confirmation ignored
- [x] Trigger through the real dispatch cycle: HoL break fires a decision, empty bands and non-dominated victims do not, repeat cycles against a stale gauge are suppressed
- [x] Tracker confirmation listener fires at most once, only for evicted requests; failed evictions re-track without notifying; victim priority bound stops a decision at the demand band
- [x] Config translation and validation
- [x] Race detector clean on the eviction and controller packages

Not covered here (tracked as acceptance criteria in the design doc): live e2e measurement of the ImmediateResponse-to-engine-abort chain.

Known edge, tolerated with a comment at the site: re-tracking a victim (bound hit or failed eviction) races a concurrent completion in a sub-microsecond window; the re-inserted dead entry costs at most one spurious ConfirmationTimeout stall on a later decision.


pkg/epp/flowcontrol/eviction/request_evictor.go
```diff
@@ -157,8 +193,23 @@ func (p *RequestEvictor) EvictN(ctx context.Context, n int) ([]string, error) {
 		}
 		item := items[0]
 
+		if item.Priority >= priorityBound {
+			// The heap orders victims lowest-priority first, so no remaining victim can be below the
+			// bound either. Re-tracking races a concurrent completion: if cleanupRequest ran between
+			// the pop and this Track, the entry is re-inserted dead and survives until a future
+			// decision evicts it, which then stalls the pacing gate for one ConfirmationTimeout. The
+			// window is sub-microsecond and the cost is bounded, so it is tolerated.
```
RishabhSaini, line 202: EvictN on a dead entry again means being bounded on a defaultEvictionConfirmationTimeout = 10 sec

alexagriffith:
great work, this is a crucial PR for ensuring flow control can work with mixed workloads and not affect higher priorities. I built the flow control image from this PR and ran some benchmarks https://github.com/alexagriffith/flow-control-benchmarks/tree/e177aa50b986370da3b6b3de04e5579796da3707/benchmark-data/batch-eviction

pkg/epp/flowcontrol/eviction/request_evictor.go
RishabhSaini: Since the queue already has Peek() and the heap guarantees the root is the lowest-priority victim, this could be a PopIfBelow(priorityBound) that checks the root's priority before popping preventing a race. 
Not asking to change it in this PR, but worth a follow-up since this path fires on every multi-victim decision with mixed priorities.

LukeAVanDrie: `Evict` on a dead entry closes a channel with no reader and returns nil, so the controller books the revocation as outstanding and `GateOpen` stays shut until `sweepTimeoutsLocked` retires it. Worth noting 10s is the ceiling, not the norm as the `ctx.Done()` cleanup goroutine from `PreRequest` fires `cleanupRequest` a second time when the request context cancels, and that delivers a late confirmation. The ordering just isn't guaranteed.

There's also a second-order effect: a re-tracked dead entry also inflates `inFlight`, and sizing uses `credit = saturation / inFlight`, so the leak shrinks the credit and biases `n` upward. `MaxRevocationsPerDecision` caps that at 2 and the error is 1/N, so it only matters at very low in-flight counts.

The `PopIfBelow` approach sounds reasonable to me. I will file a tracking issue for that and we can do this as a followup.

pkg/epp/flowcontrol/controller/internal/reclamation.go
```diff
@@ -0,0 +1,337 @@
...
+//	credit        = saturation / inFlight     (mean lease footprint estimate)
+//	n             = max(1, ceil((reclaimTarget - pendingReclaim) / credit)),
+//	                capped at MaxRevocationsPerDecision and at the evictable lease count
+//
+// demandPriority is the eligible demand band's priority, passed to the actuator as a strict upper
+// bound on victim priority so that no victim in the decision sits at or above the demand band
+// (the victim-head check in the trigger does not cover later victims of a multi-revocation
+// decision).
+func (c *ReclamationController) Reclaim(ctx context.Context, saturation, ceiling float64, demandPriority int) {
+	inFlight, evictable := c.evictor.Stats()
```
RishabhSaini, line 162: Stats() reads InFlightLen() and EvictableLen() under two separate lock acquisitions. A concurrent Untrack between them can make the snapshot inconsistent e.g. inFlight=10 then a completion drops it to 9, then evictable=5. Reclaim uses inFlight to compute credit = saturation / inFlight, so a stale value skews the sizing. This is bounded by MaxRevocationsPerDecision but an easy to fix: a single Stats() method on EvictionQueue returning both values under one lock.

RishabhSaini (approved):
LGTM. Thanks for the thorough write-up made this much easier to follow. Sizing and pacing logic is sound: deficit against mean-footprint credit, pending-reclaim debits to avoid double-evicting, and demand-priority bound to prevent same-band churn. Tests cover the key edge cases well. Two non-blocking comments inline.

ahg-g (approved):

ahg-g:
I will continue to review after the merge, this is safe to merge since it is gated.

PR #2094, opened 2026-07-20
Llmd leader worker clean 2p2d enable

shikamd123:
**What type of PR is this?**

feature

Enables AMD **MoRI-IO WRITE-mode** and **Wide-EP** disaggregation in the llm-d
sidecar, so a single DP world can be split across pods for prefill/decode
serving on ROCm (MI300X).

- Turns on the MoRI-IO feature gate and adds the Wide-EP dispatch options
  (`--moriio-write-mode`, `--moriio-parallel-dispatch`, `--moriio-dp-size`,
  `--moriio-dp-size-local`, `--moriio-remote-hosts`, `--moriio-decode-hosts`).
- Supports both **serial WRITE** (default: prefill → await → decode) and
  **parallel WRITE** (concurrent prefill/decode) dispatch.
- Makes DP-rank routing **router-authoritative**: in serial mode the decode leg
  is pinned to the `remote_dp_rank` the prefill leg actually returned (falling
  back to a stable hash only if omitted); in parallel mode the rank is pinned up
  front with `remote_dp_rank_override=true`. Both legs agree without each side
  independently hashing the request id.
- Adds **DNS hostname resolution** for the per-pod fan-out hosts (resolved to IPs
  at startup, LWS-friendly); raw IPs remain accepted for backward compatibility.
- Documents the modes/flags/examples in `MORIIO_README.md` and adds option tests.

**Release note**:
Enable AMD **MoRI-IO WRITE-mode** and **Wide-EP** disaggregation in the llm-d router sidecar, so a single data-parallel world can be split across pods for prefill/decode serving on ROCm (MI300X). Supports **1P1D** (DP=EP=8, intra-node) and **2P2D** (DP=EP=16, inter-node via LeaderWorkerSet).

**Highlights**
- **Router-authoritative DP-rank routing:** in serial WRITE the decode leg is pinned to the `remote_dp_rank` the prefill leg returned; the returned rank is validated (in-range integer) and applied consistently to both the routing header and the request body, with a deterministic hash fallback when it is missing/invalid.
- **Peer DNS re-resolution on the request path:** peer hostnames are re-resolved under a short TTL (default 30s, `MORIIO_DNS_RESOLVE_TTL_SECONDS`) so peer pod restarts are picked up without a router restart. Lookups are context-bounded (default 5s, `MORIIO_DNS_RESOLVE_TIMEOUT_SECONDS`), de-duplicated via singleflight, and use serve-stale-while-revalidate; the resolver is seeded from startup-resolved IPs so the first request skips a lookup and a request-time DNS outage serves last-known-good.
- **Safe parallel dispatch (opt-in):** serial WRITE dispatch remains the SLA-safe default. Parallel dispatch (`--moriio-parallel-dispatch`) now cancels the decode leg when prefill fails and buffers the decode response until the prefill outcome is known, so a failed prefill returns the prefill error instead of a false 200, with a bounded KV-wait backstop (`--moriio-parallel-decode-wait-timeout`, default 30s).
- **Observability:** optional Prometheus `/metrics` endpoint for the sidecar, enabled via the `--metrics-port` flag (default `0` = disabled), exposing `moriio_dns_reresolve_total`, `moriio_dns_ip_changed_total`, and `moriio_dns_lookup_failures_total`.
- Documents modes/flags/tunables in `MORIIO_README.md`

Companion to vLLM PR vllm-project/vllm#https://github.com/vllm-project/vllm/pull/45043

### DIFF
--- a/pkg/sidecar/proxy/connector_nixlv2.go
+++ b/pkg/sidecar/proxy/connector_nixlv2.go
@@ -303,12 +303,23 @@ retryLoop:
 
 	dreq.Header.Add(requestHeaderRequestID, uuidStr)
 
-	// Stamp the same DP-rank pin on the decode leg.
+	// Decode's DP rank is PROPAGATED from the prefill leg's returned
+	// kv_transfer_params (remote_dp_rank = the rank prefill actually ran on),
+	// not independently re-derived here. This is the router-applies-the-
+	// connector-returned-rank model (PR #45043 review, njhill): the prefill
+	// connector returns the rank, the router pins the decode leg to it, so both
+	// legs agree without each hashing the request id. Fall back to the hash only
+	// if the prefill response omitted it (older vLLM without the returnable rank).
+	decodeDPRank := pickDPRank(uuidStr, s.config.MoRIIODPSize)
+	if pkv, ok := pKVTransferParams.(map[string]any); ok {
+		if rv, present := pkv[requestFieldRemoteDPRank]; present {
+			if ri, ok := toInt(rv); ok {
+				decodeDPRank = ri
+			}
+		}
+	}
 	if s.config.MoRIIODPSize > 1 {
-		dreq.Header.Set(
-			requestHeaderDataParallelRank,
-			strconv.Itoa(pickDPRank(uuidStr, s.config.MoRIIODPSize)),
-		)
+		dreq.Header.Set(requestHeaderDataParallelRank, strconv.Itoa(decodeDPRank))
 	}
 
 	delete(completionRequest, requestFieldStream)
--- a/pkg/sidecar/proxy/connector_nixlv2.go
+++ b/pkg/sidecar/proxy/connector_nixlv2.go
@@ -18,13 +18,13 @@ package proxy
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net"
 	"net/http"
 	"strconv"
-	"sync"
 	"time"
 
 	"github.com/google/uuid"
@@ -309,18 +309,29 @@ retryLoop:
 	// not independently re-derived here. This is the router-applies-the-
 	// connector-returned-rank model (PR #45043 review, njhill): the prefill
 	// connector returns the rank, the router pins the decode leg to it, so both
-	// legs agree without each hashing the request id. Fall back to the hash only
-	// if the prefill response omitted it (older vLLM without the returnable rank).
-	decodeDPRank := pickDPRank(uuidStr, s.config.MoRIIODPSize)
-	if pkv, ok := pKVTransferParams.(map[string]any); ok {
-		if rv, present := pkv[requestFieldRemoteDPRank]; present {
-			if ri, ok := toInt(rv); ok {
-				decodeDPRank = ri
+	// legs agree without each hashing the request id. The returned rank is
+	// validated to be in [0, dp_size); an omitted, non-numeric, or out-of-range
+	// value falls back to the deterministic hash. The header AND the decode
+	// body's remote_dp_rank are then pinned to the SAME validated value so they
+	// cannot target different ranks and hang the transfer.
+	// DP-rank propagation is a MoRI-IO WRITE-mode concern only. In standard
+	// NIXLv2 READ mode the decode body's remote_dp_rank / remote_dp_rank_override
+	// and the x-data-parallel-rank header are left untouched, matching the legacy
+	// wire shape.
+	if s.config.MoRIIOWriteMode {
+		decodeDPRank, usedReturned := resolveDecodeDPRank(pKVTransferParams, uuidStr, s.config.MoRIIODPSize)
+		if pkv, ok := pKVTransferParams.(map[string]any); ok {
+			if rv, present := pkv[requestFieldRemoteDPRank]; present && !usedReturned && s.config.MoRIIODPSize > 1 {
+				s.logger.V(1).Info("prefill returned invalid/out-of-range remote_dp_rank; using hash fallback",
+					"request_id", uuidStr, "returned", rv,
+					"dp_size", s.config.MoRIIODPSize, "rank", decodeDPRank)
 			}
+			pkv[requestFieldRemoteDPRank] = decodeDPRank
+			pkv[requestFieldRemoteDPRankOverride] = true
 		}
-	}
-	if s.config.MoRIIODPSize > 1 {
-		dreq.Header.Set(requestHeaderDataParallelRank, strconv.Itoa(decodeDPRank))
+		if s.config.MoRIIODPSize > 1 {
+			dreq.Header.Set(requestHeaderDataParallelRank, strconv.Itoa(decodeDPRank))
+		}
 	}
 
 	delete(completionRequest, requestFieldStream)
@@ -610,10 +621,16 @@ func (s *Server) runNIXLProtocolV2WriteParallel(
 		return
 	}
 
-	// Build cloned requests under separate contexts so each carries its own
-	// span lineage and either side can be observed/cancelled independently
-	// without affecting the other.
-	pCtx, prefillSpan := tracer.Start(parentCtx, "llm_d.pd_proxy.prefill",
+	// Fire prefill and decode concurrently, but DEFER committing decode's
+	// response to the client until prefill's outcome is known (the commit
+	// point). Both legs share a cancelable context so a failed or hung prefill
+	// aborts decode's in-flight request / KV wait immediately instead of
+	// letting it hang, and decode's output is buffered so a failed prefill can
+	// never surface a bogus 200 to the client.
+	dispatchCtx, cancel := context.WithCancel(parentCtx)
+	defer cancel()
+
+	pCtx, prefillSpan := tracer.Start(dispatchCtx, "llm_d.pd_proxy.prefill",
 		trace.WithSpanKind(trace.SpanKindInternal),
 	)
 	prefillSpan.SetAttributes(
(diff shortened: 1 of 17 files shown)


shmuelk:
@shikamd123 Please remove the file PR_DESCRIPTION.md and add any missing text to the PR description.

pkg/sidecar/proxy/options_test.go
```diff
@@ -1231,3 +1231,234 @@ func TestCompleteTLSConfiguration(t *testing.T) {
...
+		result, err := resolveHostsToIPs(nil)
+		require.NoError(t, err)
+		require.Empty(t, result)
+
+		result, err = resolveHostsToIPs([]string{})
+		require.NoError(t, err)
+		require.Empty(t, result)
+	})
+
+	t.Run("raw IPv4 addresses are passed through", func(t *testing.T) {
```
shikamd123, line 1323: remove the IP addresses, there should be DNS resolution, test it and update it.

pkg/sidecar/proxy/options_test.go
```diff
@@ -1231,3 +1231,234 @@ func TestCompleteTLSConfiguration(t *testing.T) {
...
+		t.Skip("MoRI-IO feature is dormant; skipping DNS resolution tests")
+	}
+
+	t.Run("raw IP addresses are passed through", func(t *testing.T) {
+		opts := NewOptions()
+		opts.MoRIIOWriteMode = true
+		opts.MoRIIODecodePodIP = "10.0.1.1"
+		opts.MoRIIODPSize = 16
+		opts.MoRIIODPSizeLocal = 8
+		opts.MoRIIORemoteHosts = []string{"10.0.0.1", "10.0.0.2"}
```
shikamd123: do we need IP addresses?

vcave:
@vMaroon Could we merge this PR and include it in the upcoming release? This would enable moriio-based guides support in the main llm-d repo

shikamd123:
@robertgshaw2-redhat  @vcave  can we please get a review on the router PR and target it to get merged so that we can have the feature enablement in August's LLMD release?

ahg-g:
can you pls address the lint error

release-notes.d/unreleased/2094.md
```diff
@@ -0,0 +1,7 @@
+---
+pr: 2094
+url: https://github.com/llm-d/llm-d-router/pull/2094
+author: shikamd123
+date: 2026-08-04
+---
+Enable AMD MoRI-IO WRITE-mode and Wide-EP disaggregation in the sidecar, with request-path TTL re-resolution of peer DNS names (serve-stale-while-revalidate, per-lookup timeout bounding, and singleflight de-duplication) so peer pod restarts are picked up without a router restart. Tunable via `MORIIO_DNS_RESOLVE_TTL_SECONDS` (default 30) and `MORIIO_DNS_RESOLVE_TIMEOUT_SECONDS` (default 5); opt-in Prometheus counters (`moriio_dns_reresolve_total`, `moriio_dns_ip_changed_total`, `moriio_dns_lookup_failures_total`) are exposed at `/metrics` when `MORIIO_METRICS_ADDR` is set.
```
ahg-g: pls add this in the description of the PR as a release note (see https://github.com/llm-d/llm-d-router/pull/2299#issue-5079423850 as an example). This fragment then gets automatically created via a github action.

pkg/sidecar/proxy/options.go
```diff
@@ -48,9 +50,9 @@ const (
 	// fail with a clear error message directing users to wait for the feature
 	// to be officially released.
 	//
-	// TODO(AMD-MoRI-IO): Set to true once CI tests and production validation
-	// are complete in a future release candidate.
-	MoRIIOFeatureEnabled = false
+	// AMD-MoRI-IO: Feature enabled for 1P1D (DP=EP=8) intra-node Wide-EP
+	// and 2P2D (DP=EP=16) inter-node Wide-EP with LeaderWorkerSet support.
+	MoRIIOFeatureEnabled = true
```
nilig, line 55: This also enables parallel dispatch, but that path does not handle a failed prefill. A focused test with prefill returning 500 and decode returning 200 returns 200 to the client. In a real WRITE request, decode can instead wait for KV that will never arrive and hang. We should cancel decode as soon as prefill fails, return the prefill error when possible, and add a test for this case before enabling the feature.

pkg/sidecar/proxy/proxy.go
```diff
@@ -340,9 +361,55 @@ type Server struct {
...
+
 	config Config
 }
 
+// resolver lazily initializes and returns the request-path host resolver.
+// Initialization is deferred so it can use s.logger (populated in Start) and
+// so Clone'd servers each get their own instance.
+func (s *Server) resolver() *hostResolver {
+	s.resolverOnce.Do(func() {
+		s.hostResolver = newHostResolver(s.logger, resolveTTLFromEnv())
```
nilig, line 383: `Complete()` already resolves each DNS name, but this creates an empty resolver and the request path ignores those resolved values whenever the original specs are present. The first request therefore repeats the lookups. If DNS is unavailable then, we send the raw hostname even though startup had a working IP, which can leave the MoRI-IO handshake hanging. Please seed the resolver with the startup-resolved spec-to-IP values and add a test for a request-time DNS failure.

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -303,12 +304,23 @@ retryLoop:
...
+	// kv_transfer_params (remote_dp_rank = the rank prefill actually ran on),
+	// not independently re-derived here. This is the router-applies-the-
+	// connector-returned-rank model (PR #45043 review, njhill): the prefill
+	// connector returns the rank, the router pins the decode leg to it, so both
+	// legs agree without each hashing the request id. Fall back to the hash only
+	// if the prefill response omitted it (older vLLM without the returnable rank).
+	decodeDPRank := pickDPRank(uuidStr, s.config.MoRIIODPSize)
+	if pkv, ok := pKVTransferParams.(map[string]any); ok {
+		if rv, present := pkv[requestFieldRemoteDPRank]; present {
+			if ri, ok := toInt(rv); ok {
```
nilig: If `remote_dp_rank` is present but is not a valid in-range integer, the header falls back to the hash while the decode request body keeps the invalid returned value. The header and body can then target different ranks and hang the transfer. Please validate the returned rank and either use it consistently in both places or fail the request. Please also test returned, omitted, invalid, and out-of-range rank values.

nilig (changes requested):
These issues can produce false success responses or hung requests, so I think they need to be fixed before merge.

pkg/sidecar/proxy/options.go
shikamd123: agree, let me add the necessary changes.

pkg/sidecar/proxy/proxy.go
shikamd123: will address the changes.

pkg/sidecar/proxy/connector_nixlv2.go
shikamd123: got it, let me address it.

release-notes.d/unreleased/2094.md
shikamd123: done

pkg/sidecar/proxy/options.go
shikamd123: Prefill failure now cancels decode via a shared context and returns the prefill error (never a false 200) through a buffered commit point, with a 30s KV-wait backstop → 504; parallel dispatch stays opt-in with serial as default.
Covered by prefill-500→500, both-block→504-no-hang, and success→200-SSE tests in `connector_nixlv2_parallel_commit_test.go`.

pkg/sidecar/proxy/connector_nixlv2.go
shikamd123: `resolveDecodeDPRank` now validates the returned rank as an in-range int and pins header + body to the same value, with a deterministic hash fallback.
Test: `TestResolveDecodeDPRank` (returned/omitted/invalid/out-of-range/single-DP).

pkg/sidecar/proxy/proxy.go
shikamd123: `seedResolver` primes the request-path resolver with startup-resolved spec→IP, so the first request skips the lookup and a DNS outage serves last-known-good.
Tests: `TestServer_ResolverSeededFromStartup`, `TestHostResolver_SeedServesStartupIPOnDNSFailure`.

nilig (dismissed):
Approving to unblock the feature. One non-blocking follow-up: `remote_dp_rank` still accepts fractional JSON numbers because `toInt` truncates them. For example, `3.5` becomes rank `3`. Please reject non-integer values and add a fractional-rank test in a follow-up.

vMaroon (dismissed):

nilig:
Closing/reopening to re-trigger CI — the force-push left the current head with no workflow runs.

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -303,12 +304,28 @@ retryLoop:
...
+	// not independently re-derived here. This is the router-applies-the-
+	// connector-returned-rank model (PR #45043 review, njhill): the prefill
+	// connector returns the rank, the router pins the decode leg to it, so both
+	// legs agree without each hashing the request id. The returned rank is
+	// validated to be in [0, dp_size); an omitted, non-numeric, or out-of-range
+	// value falls back to the deterministic hash. The header AND the decode
+	// body's remote_dp_rank are then pinned to the SAME validated value so they
+	// cannot target different ranks and hang the transfer.
+	decodeDPRank, usedReturned := resolveDecodeDPRank(pKVTransferParams, uuidStr, s.config.MoRIIODPSize)
+	if pkv, ok := pKVTransferParams.(map[string]any); ok {
```
nilig: `remote_dp_rank` and `remote_dp_rank_override` are also written when `MoRIIOWriteMode` is false, so the default NIXLv2 decode wire changes and a connector that honors the override can pin transfers to rank 0. Please gate this body mutation on `MoRIIOWriteMode` and add a standard-mode test asserting that these fields are absent.

shikamd123: let me add the guard, thanks for the review

shikamd123: addressed

nilig (approved):

ahg-g (approved):

PR #2111, opened 2026-07-21
test(coordinator): Fix request timeout for multimodal e2e specs

### LINKED ISSUE #2112: Coordinator e2e multimodal specs time out
reported by @roytman -
The coordinator e2e multimodal image specs (`test/coordinator/e2e/coordinator/coordinator_test.go`) intermittently fail while posting to the gateway:

    Post "http://localhost:30080/v1/chat/completions": context deadline
    exceeded (Client.Timeout exceeded while awaiting headers)

After the request aborts, the spec's DeferCleanup deletes the workload, producing "graceful shutdown" log lines from the coordinator and model servers. Those are a consequence of the failed request tearing the deployment down, not the cause.

For example: https://github.com/llm-d/llm-d-router/actions/runs/29765682769/job/88432002857?pr=2090

revit13:
**What type of PR is this?**
/kind test
/kind bug

Fixes the intermittent coordinator e2e failure where the client
`POST /v1/chat/completions` fails with `context deadline exceeded` while
the coordinator only logs health probes (the request never reaches it).

[e2e_failure_log.txt](https://github.com/user-attachments/files/30258973/e2e_failure_log.txt)

Root cause (confirmed from the Envoy access log): the client
POST reaches Envoy and matches the `coordinator` STRICT_DNS cluster, but
Envoy momentarily cannot open a new upstream connection to the coordinator
Service ClusterIP (a transient kube-proxy/conntrack settling window in the
kind cluster). Because that cluster used `connect_timeout: 1000s` and the
route had no retry, the single connect attempt was held far past the
client's 60s timeout, turning a few-second transient into a hung request.
The gateway view proves it: the POST logs `cluster=coordinator upstream=-`
for the full ~55s, while a `/readyz` to the same ClusterIP succeeded 2ms
earlier and the transient cleared ~7s later (the readiness poll rides it
out by retrying; the POST had no such resilience).

Fix: on the coordinator cluster/route, fail a stuck connect fast and retry
it on a fresh connection.
- coordinator cluster `connect_timeout: 1000s -> 5s`
- default route `retry_policy: { retry_on: "connect-failure", num_retries: 3 }`
`connect-failure` is safe for the non-idempotent POST: the connection never
established, so nothing reached the coordinator.

Also included (hardening, same failure surface):
- Create the coordinator and per-phase EPP Services, ServiceAccounts, and
  RoleBindings once in BeforeSuite (`createStableInfra`) and recreate only
  the Deployments per spec, mirroring the non-coordinator e2e. This removes
  per-spec Service ClusterIP churn (a real problem, though not the root
  cause of this flake).
- `shutdown_timeout: 25s -> 1s` so a torn-down coordinator stops serving
  promptly instead of lingering during the brief pod-drain overlap between
  specs.
- Add failure-only diagnostics (Envoy stdout access log, per-deployment
  log dumps, and Envoy `/clusters`) so a future failure records the
  gateway's view of the hung request. These print only on spec failure
  (or with `E2E_PRINT_COORDINATOR_LOGS=true`).

### Here is a streamlined breakdown of the logs proving the exact point of failure in the request flow:

Fact 1 — the POST reached Envoy. Envoy's access log contains it (Envoy only logs requests it received):
[envoy] 1784703989.516 POST /v1/chat/completions code=0 flags=- cluster=coordinator upstream=- epp-phase=- id=431f407b… dur=5000
epp-phase=- = no EPP-Phase header → this is the external client POST, and cluster=coordinator = Envoy matched it to the default route → coordinator cluster. Proven: the POST arrived at Envoy and was routed to the coordinator cluster.

Fact 2 — Envoy never obtained an upstream connection to the coordinator. The same request is flushed every 5s (because of access_log_flush_interval), 11 snapshots, and in every one the upstream field is empty:
… upstream=- … dur=5000
… upstream=- … dur=9999
… upstream=- … dur=15000
… (20000, 25000, 30001, 35000, 40000, 45001, 50001) …
… upstream=- … dur=55001
upstream=- from dur=5000 to dur=55001 means Envoy never bound the request to an upstream host/connection for ~55 seconds. Contrast with a working request, whose field is populated: upstream=10.96.23.10:8080. Proven: no upstream connection was ever established for the POST.

Fact 3 — it never reached the coordinator. The coordinator logs every request on entry (server.go:67). Across the whole failing spec its log contains only:
14 × "method":"GET","path":"/healthz"
18 × "method":"GET","path":"/readyz"
Zero POST. Proven: the POST did not reach the coordinator process — consistent with Fact 2.

Fact 4 — the coordinator was reachable at that instant (so this is not "coordinator down"). 0.5 ms before the POST started, a /readyz connected to the same ClusterIP and succeeded:
- coordinator log: ts=1784703989.5154674 … "method":"GET","path":"/readyz" … x-request-id:"b90897a8…" (the coordinator received it)
- Envoy access log for that id: GET /readyz code=200 upstream=10.96.23.10:8080 dur=0

POST START_TIME is 1784703989.516. So: at …989.5155 a request connected to 10.96.23.10 and got 200; at …989.516 the POST began and never got an upstream for 55s. Proven: the endpoint was connectable immediately before the POST.

Proven from the logs: the POST reached Envoy, was routed to the coordinator cluster, and Envoy never obtained an upstream connection to the coordinator for the full ~55 s (upstream=-), so nothing reached the coordinator, even though the coordinator answered a request on the same ClusterIP 0.5 ms earlier. It hung in the pending/connecting state (code=0, flags=- — no terminal UF/UH/DC) because connect_timeout: 1000s meant the attempt never failed; the client's 60 s deadline fired first.

Fixes #2112

```release-note
NONE
```

### DIFF
--- a/test/coordinator/e2e/coordinator/coordinator_test.go
+++ b/test/coordinator/e2e/coordinator/coordinator_test.go
@@ -121,10 +121,11 @@ func runCoordinatorPipeline(body []byte, expectedSteps []string, expectedImages 
 		testutils.DeleteObjects(testConfig, encodePool, nsName)
 	})
 
-	// Dump coordinator logs on failure, or always when E2E_PRINT_COORDINATOR_LOGS is
-	// set. Registered second → runs first (LIFO), so the deployment still exists.
+	// Dump all pod logs (coordinator, EPPs, Envoy, workers) on failure, or always
+	// when E2E_PRINT_LOGS is set. Registered second → runs first (LIFO), so the
+	// pods still exist.
 	ginkgo.DeferCleanup(func() {
-		if !ginkgo.CurrentSpecReport().Failed() && !printCoordinatorLogs {
+		if !ginkgo.CurrentSpecReport().Failed() && !printLogs {
 			return
 		}
 		testutils.DumpPodsAndLogs(testConfig, nsName, testutils.WithFullLogs())
--- a/test/coordinator/e2e/coordinator/coordinator_test.go
+++ b/test/coordinator/e2e/coordinator/coordinator_test.go
@@ -31,15 +31,11 @@ import (
 	testutils "github.com/llm-d/llm-d-router/test/utils"
 )
 
-const (
-	// requestTimeout bounds text-only chat completions.
-	requestTimeout = 60 * time.Second
-	// multimodalRequestTimeout bounds multimodal chat completions, which run the
-	// extra replace-media-urls/render/encode stages; 60s times out before they
-	// finish. Long budget so the request completes and the coordinator logs
-	// (dumped at the end of the spec) show where the time actually went.
-	multimodalRequestTimeout = 600 * time.Second
-)
+// requestTimeout is intentionally long (applies to all specs, text-only and
+// multimodal) so the request completes even on the intermittent slow/black-holed
+// path, letting the end-of-spec coordinator log dump show whether and when the
+// request actually reached the coordinator.
+const requestTimeout = 600 * time.Second
 
 // testImageURL and testImageURL2 are publicly accessible images used to
 // exercise multimodal requests that trigger the encode stage.
@@ -135,7 +131,12 @@ func runCoordinatorPipeline(body []byte, expectedSteps []string, expectedImages 
 		if !ginkgo.CurrentSpecReport().Failed() && !printCoordinatorLogs {
 			return
 		}
-		dumpCoordinatorLogs(nsName)
+		dumpDeploymentLogs(nsName, "llm-d-coordinator", "coordinator")
+		dumpDeploymentLogs(nsName, eppName+"-encode", "epp")
+		dumpDeploymentLogs(nsName, eppName+"-prefill", "epp")
+		dumpDeploymentLogs(nsName, eppName+"-decode", "epp")
+		dumpDeploymentLogs(nsName, "envoy", "envoy")
+		dumpEnvoyClusters(nsName)
 	})
 
 	// Pools first so each EPP can resolve its --pool-name.
@@ -166,11 +167,7 @@ func runCoordinatorPipeline(body []byte, expectedSteps []string, expectedImages 
 	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
 	req.Header.Set("Content-Type", "application/json")
 
-	timeout := requestTimeout
-	if expectedImages > 0 {
-		timeout = multimodalRequestTimeout
-	}
-	client := &http.Client{Timeout: timeout}
+	client := &http.Client{Timeout: requestTimeout}
 	resp, err := client.Do(req)
 	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
 	defer resp.Body.Close()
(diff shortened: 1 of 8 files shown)


elevran:
What is the major source of latency in the tests and can it be mitigated instead of extending the deadline?
Prefer shorter tests with predictable inputs and not relying on external downloads. For example, you could use a "fixed" local image from `testdata` directory, or generate a random one on the fly.

shmuelk (dismissed):

shmuelk:
One could stand up a local http server that serves local files...

revit13:
/hold
Thanks for the review! Holding this for now, I'm still investigating the root cause and want to confirm it before we merge.

elevran:
> /hold Thanks for the review! Holding this for now, I'm still investigating the root cause and want to confirm it before we merge.

thanks. the fact the [e2e test fails](https://github.com/llm-d/llm-d-router/actions/runs/29828859735/job/88628567711?pr=2111) on this PR suggests that the longer timeout is not the right fix.

revit13:
> > /hold Thanks for the review! Holding this for now, I'm still investigating the root cause and want to confirm it before we merge.
> 
> thanks. the fact the [e2e test fails](https://github.com/llm-d/llm-d-router/actions/runs/29828859735/job/88628567711?pr=2111) on this PR suggests that the longer timeout is not the right fix.

Thanks. The first commit's timeout bump did get the e2e passing, but I agree a longer timeout isn't the real fix, it only masks the issue. I'm switching to a root-cause fix that doesn't touch the timeout and will update the PR.

test/coordinator/e2e/coordinator/coordinator_test.go
```diff
@@ -127,17 +127,12 @@ func runCoordinatorPipeline(body []byte, expectedSteps []string, expectedImages
...
-		if k8sContext != "" {
-			args = append(args, "--context="+k8sContext)
-		}
-		out, err := exec.Command("kubectl", args...).CombinedOutput()
-		if err != nil {
-			fmt.Fprintf(ginkgo.GinkgoWriter, "\n--- coordinator logs (kubectl error: %v) ---\n%s\n---\n", err, string(out))
-		} else {
-			fmt.Fprintf(ginkgo.GinkgoWriter, "\n--- coordinator logs ---\n%s\n---\n", string(out))
-		}
+		dumpDeploymentLogs(nsName, "llm-d-coordinator", "coordinator")
```
roytman: why do you want to dump logs for ALL, even successful tests?

test/coordinator/e2e/coordinator/coordinator_test.go
```diff
@@ -180,6 +175,47 @@ func runCoordinatorPipeline(body []byte, expectedSteps []string, expectedImages
...
 	gomega.Expect(raw).NotTo(gomega.BeEmpty(), "coordinator returned empty body")
 	verifyCoordinatorSteps(nsName, expectedSteps, expectedImages, true, true)
+
+	// Diagnostic: dump the coordinator logs once the request has completed. The
+	// per-step timestamps show when the request reached the coordinator (relative
+	// to when it was sent) and how long each pipeline step took, so we can tell
+	// late-arrival (gateway) from slow processing (pipeline). Also dump Envoy's
+	// access log so each request's gateway view (code/flags/cluster/upstream) is
+	// visible even on green runs, as a baseline to compare against a failure.
+	dumpDeploymentLogs(nsName, "llm-d-coordinator", "coordinator")
```
roytman: "dump the coordinator logs once the request has completed" can create a lot of logs

test/coordinator/e2e/coordinator/testdata/envoy.yaml
```diff
@@ -80,6 +80,13 @@ data:
                     "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
                     stat_prefix: http-8081
                     preserve_external_request_id: true
+                    access_log:
```
roytman, line 83: Access log for each request

roytman (commented):
@revit13 thank you for your PR.

Please update the PR description; it is wrong now. 
Please explain your intention about the logs; do you plan to add them permanently? (it will create huge log outputs), Or temporarily, for the fix validation?

revit13:
> @revit13 thank you for your PR.
> 
> Please update the PR description; it is wrong now. Please explain your intention about the logs; do you plan to add them permanently? (it will create huge log outputs), Or temporarily, for the fix validation?

Its for the fix validation. Thanks

test/coordinator/e2e/coordinator/e2e_suite_test.go
```diff
@@ -131,6 +132,10 @@ var _ = ginkgo.BeforeSuite(func() {
 	}
 
 	rendererObjects = createRenderer()
+
+	// Coordinator and EPP Services/RBAC are created once and kept stable across
+	// specs (see createStableInfra).
+	stableInfraObjects = createStableInfra()
```
roytman: Partial-failure leak:
This single assignment which runs after the whole function returns. If it fails partway (e.g. the coordinator Service creates, then an EPP's applyManifest call fails a `gomega.Expect`), `stableInfraObjects` stays nil and `ReportAfterSuite's` cleanup is skipped even though some objects were already created cluster-side. 

Harmless on the default disposable kind-cluster CI flow, but leaks permanently if K8S_CONTEXT points at a shared/persistent cluster with a pre-existing namespace. 

Fix: append to a suite-level slice incrementally inside `createStableInfra` rather than returning a fresh slice at the end.

test/coordinator/e2e/coordinator/setup_test.go
```diff
@@ -222,13 +229,39 @@ func createEPPConfigMap(name, content string) {
...
+// churn per spec. Mirrors the non-coordinator e2e, which creates the EPP
+// Services and RBAC once in its setup and recreates only the Deployment per test.
+func createStableInfra() []string {
+	eppManifests := []string{encodeEPPManifest, prefillEPPManifest, decodeEPPManifest}
+	// Coordinator Service + ServiceAccount, plus Service + ServiceAccount +
+	// RoleBinding per EPP.
+	objects := make([]string, 0, 2+len(eppManifests)*3)
+
+	docs := e2eutil.RunKustomize(coordinatorComponentDir)
+	docs = e2eutil.FilterKinds(docs, "ConfigMap", "Deployment")
```
roytman, line 251: This exclude list and the ones in `createCoordinator` and `createEndPointPicker` must stay complementary — together they need to create every resource kind in the manifests exactly once. 
Nothing enforces that today; a manifest gaining a new kind without updating both sides silently drops or duplicates a resource. 
A small unit test asserting the three FilterKinds calls partition the kind set with no gap or overlap (no cluster needed) would catch drift here.

revit13: I understand the concern, but I'm leaning against adding that assertion here:
It’s test scaffolding, so validating these setup helpers essentially means writing a test for a test. Plus, these manifests rarely change.
Any drift is already caught by the e2e test. If a kind is excluded entirely, the suite fails to come up. If it's duplicated, it fails with an AlreadyExists error.
Both functions already clearly reference createStableInfra for the split definition.

I'd rather keep this PR tightly scoped to fixing the flake. That said, I'm happy to open a small follow-up PR to add it. Thanks

roytman (approved):
/lgtm

revit13:
/hold cancel

Updated the PR description after finding the root cause. Captured the failing request directly (Envoy access log with START_TIME + in-flight flush): the client POST reaches Envoy and matches the coordinator cluster but never gets an upstream connection (upstream=- for ~55s), while a /readyz to the same ClusterIP succeeded ~0.5ms earlier. With connect_timeout: 1000s and no retry, that brief connect hiccup became a 60s hang. Fix: connect_timeout: 1000s → 5s + a default-route retry_policy (connect-failure, 3 retries). Churn was ruled out, single stable ClusterIP throughout.

elevran:
Q: the PR bundles the retry fix with a second, independent change: creating coordinator/EPP Services once (in `createStableInfra` instead of per-spec, to stop ClusterIP churn). 
The Cluster IP churn is plausibly the actual root cause driver (repeated Service recreation is a more Kind/kube-proxy-specific trigger for this than mere randomness). It is a separate mechanism from the Envoy retry, so those are independent fixes.

Do you know whether the retry policy alone (without the Service-stabilization change) was verified to fix the flake, or whether both changes together were needed?

/approve

elevran (approved):
approving - please see question on root cause above.

revit13:
> Q: the PR bundles the retry fix with a second, independent change: creating coordinator/EPP Services once (in `createStableInfra` instead of per-spec, to stop ClusterIP churn). The Cluster IP churn is plausibly the actual root cause driver (repeated Service recreation is a more Kind/kube-proxy-specific trigger for this than mere randomness). It is a separate mechanism from the Envoy retry, so those are independent fixes.
> 
> Do you know whether the retry policy alone (without the Service-stabilization change) was verified to fix the flake, or whether both changes together were needed?
> 
> /approve

Thanks. I did not test the retry policy on its own. What I can confirm is that the "create Services once" change alone did not fix the flake: the failing run had that change in place (the coordinator ClusterIP was stable across specs, no churn), and it still hung. The Envoy access log I added for diagnosis showed why: /readyz returned 200 on the coordinator ClusterIP immediately before the request that hung, so the ClusterIP was reachable and Service/ClusterIP churn was not the cause.

PR #2116, opened 2026-07-21
epp: add turn-priority strategy to program-aware fairness

### LINKED ISSUE #2088: epp/fairness: add turn-priority scheduling strategy to program-aware fairness
Labels: kind/feature, area/epp

Add a second scoring strategy, turn-priority, to the program-aware-fairness flow-control policy, alongside the existing las (Least Attained Service) strategy.

turn-priority orders waiting flows by their head request's priority:

prio = turn_number + turnPriorityTimeWeight * elapsed_seconds

- Turn number is the program's dispatched-request count (per fairness ID), so deeper sessions score higher and are favored.
- The elapsed term ages waiting flows upward so lower-turn flows are not starved.
- Scoring engages only under contention — with no waiting flow nothing dispatches; with exactly one, it dispatches directly.
- The turn counter resets on eviction — a program idle past the eviction TTL has likely lost its KV cache, so on return it restarts from turn one and re-earns priority from a cold state.

Adds one config field, turnPriorityTimeWeight (default 0.5).

**Why is this needed**:

This is a step toward KV-cache-aware scheduling for multi-turn agent workloads, in the spirit of the [Continuum](https://arxiv.org/pdf/2511.02230) paper.

Continuum's central observation is that multi-turn agent workloads interleave LLM calls with tool calls, and evicting a session's KV cache during those pauses destroys reuse across turns — the paper addresses this with TTL-based KV-cache pinning plus program-level priority to preserve multi-turn continuity.

turn-priority targets the same objective from the scheduling side rather than the cache-pinning side: favor the flows whose KV cache is warmest and deepest, so their sessions complete sooner and their cached prefix is reused before it can be evicted. The eviction-driven counter reset makes the connection explicit — turn number is used as a proxy for accumulated KV-cache warmth, and it is discarded precisely when the cache is assumed cold.

**Future work**

1. Some more tweaks, strategies and signals can be used to further refine the priority score.
2. The priority score can be sent to the inference-server to allow better eviction/serving when the requests leave llm-d
3. Full KV-cache TTL retention (Continuum's pinning mechanism) can be an additional, separated feature.

ruocco:
**What type of PR is this?**
/kind feature

Adds a second scoring strategy, `turn-priority`, to the `program-aware-fairness` flow-control policy, alongside the existing `las` strategy. It orders waiting flows by their head request's `prio = turn_number + turnPriorityTimeWeight * elapsed_seconds`, favoring deeper sessions while the elapsed term ages waiting flows up to avoid starvation. Scoring engages only under contention (a lone waiting flow dispatches directly). The per-program turn counter resets on eviction to match cold KV-cache state.

Adds one config field, `turnPriorityTimeWeight` (default `0.5`), with validation.

Fixes #2088

```release-note
Add a `turn-priority` scoring strategy to the program-aware fairness policy, configurable via `strategy: turn-priority` with a new `turnPriorityTimeWeight` field (default 0.5).
```

**Test plan** _(new tests, all passing locally)_:
- [x] Turn-priority scoring prefers higher turn number and longer wait
- [x] Single waiting flow bypasses scoring; empty queues skipped; zero-weight ties resolve
- [x] Config factory accepts `turn-priority`; rejects negative `turnPriorityTimeWeight`

### DIFF
--- a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
+++ b/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
@@ -17,9 +17,33 @@ Choose this policy when:
 
 * **Identifies programs** via the fairness ID header carried on each request.
 * **Tracks per-program metrics** through the request lifecycle: queue wait time, dispatched count, in-flight count, last completion time, and (LAS) attained service in weighted tokens.
-* **Selects a queue to dispatch from** using the configured strategy. Currently `las` (Least Attained Service) is supported; programs with the lowest accumulated service score highest.
+* **Selects a queue to dispatch from** using the configured strategy: `las` (Least Attained Service), where programs with the lowest accumulated service score highest, or `turn-priority`, which favors deeper sessions.
 * **Decays attained service** in wall-clock time so a long-idle program is not penalized indefinitely. Decay is applied lazily whenever a program's service is read or accumulated, so it does not depend on the program being visited by the dispatch loop. The rate is set by an explicit half-life.
 * **Evicts idle program state** on a periodic sweep so per-program memory and Prometheus label series do not accumulate forever.
+
+## Strategies
+
+### `las`
+
+Orders programs by least attained service, so a program that has consumed fewer weighted tokens is dispatched first.
+
+### `turn-priority`
+
+Orders programs by how deep their session is. A session that has already taken many turns most likely still has its prefix in the KV cache, so serving it again costs less prefill and finishing it frees the slot sooner.
+
+Depth alone would starve shallow sessions, so head wait is scored alongside it under `turnPriorityTimeWeight`:
+
+```
+score = turnNumber + turnPriorityTimeWeight * headWaitSeconds
+```
+
+Head wait is unbounded, so a shallow program overtakes any deeper rival once it has waited long enough. The two terms are a raw count and raw seconds, so `turnPriorityTimeWeight` converts one second of waiting into a number of turns. A value tuned for one traffic profile does not carry over to a workload with a different depth or wait distribution.
+
+A program that has been idle beyond `turnPriorityInactivitySeconds` counts as turn one and re-earns depth, on the grounds that a session dormant that long has stopped competing for its own prefix. That threshold is separate from `evictionTtlSeconds`, which governs only when per-program bookkeeping is reclaimed.
+
+The default of `120` leaves a margin over the longest activity observed within a session, so a program still taking turns keeps its depth. On an agentic trace replay averaging 48 turns per session, mean think time between turns was about 2.5s and the slowest turn-to-turn cycle reached 96s under the heaviest load tested. A cycle includes service and queueing, so the post-completion gap this threshold measures is smaller. Workloads whose sessions pause for longer need a higher value.
+
+Under contention a prefix is evicted by capacity pressure while its program is still active, and ordering by turn depth is what keeps those prefixes resident. A low prefix-cache hit rate is therefore weak evidence for lowering this value.
 
 ## Unit of Fairness
 
@@ -51,10 +75,12 @@ flowControl:
 
 | Field | Default | Description |
 |---|---|---|
-| `strategy` | `las` | Scoring strategy. Only `las` is supported. |
+| `strategy` | `las` | Scoring strategy: `las` or `turn-priority`. |
 | `lasWeightService` | `0.8` | Weight on the inverted attained-service signal. Higher values prioritize underserved programs more aggressively. |
 | `lasWeightHeadWait` | `0.2` | Weight on the head-of-queue age. Acts as a tiebreaker on cold start when programs have equal attained service. |
 | `lasHalfLifeSeconds` | `60` | Wall-clock half-life of attained service. `0` disables decay, making attained service cumulative for the program's lifetime. |
+| `turnPriorityTimeWeight` | `0.05` | (`turn-priority`) Score contribution per second of head wait, against one point per turn of depth. `0` orders on depth alone. |
+| `turnPriorityInactivitySeconds` | `120` | (`turn-priority`) A program idle for longer than this counts as turn one. `0` disables the reset. |
 | `evictionTtlSeconds` | `3600` | A program with no completion in this window is evicted from the metrics map. |
 | `evictionSweepSeconds` | `300` | How often the eviction sweep runs. Must be `> 0`. |
 
@@ -79,6 +105,7 @@ mean the scheduler is scoring the program on stale service.
 
 * **Abandoned requests block eviction**: Requests abandoned after dispatch leave `inFlight` non-zero, and the eviction sweep skips any program with non-zero `inFlight`, so its `ProgramMetrics` entry and Prometheus series persist indefinitely.
 * **Memory and label-series growth**: A new program ID adds a `ProgramMetrics` entry plus per-program Prometheus label series. The eviction sweep bounds growth, but a workload with rapidly churning program IDs (e.g. a fresh ID per request) will see TTL-bounded accumulation. Choose a TTL that matches your churn rate.
+* **Turn-priority weighting is scale-dependent**: the score sums a turn count and a wait in seconds, so `turnPriorityTimeWeight` converts between the two rather than balancing them proportionally. A value tuned for one traffic profile does not carry over to another with a different depth or wait distribution.
 * **Decay accrues during generation**: Decay is continuous wall-clock time, including while a program's requests are in flight; the service consumed by an in-flight request is added back, post-decay, when it completes. With half-lives well above typical generation times this effect is negligible.
 
 ## Related Documentation
--- a/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
+++ b/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
@@ -9,9 +9,17 @@ The program-aware fairness policy schedules per-program queues using aggregated 
 
 Choose this policy when:
 
-* **Workflow-level fairness matters**: Distinct agentic workflows or tenants share the same pool, and you want fair allocation between them rather than between individual requests.
+* **Workflow-level fairness matters**: Distinct agentic workflows or tenants share the same pool, and you want allocation decided between them rather than between individual requests.
+
+Under `las`:
+
 * **Token cost varies widely between flows**: A program issuing a few large requests should not dominate a program issuing many small ones, or vice versa.
 * **Programs may go idle and return**: Idle programs should not retain accumulated service indefinitely; on return they should compete on near-equal footing with persistently active ones.
+
+Under `turn-priority`:
+
+* **Sessions are long and prefix reuse dominates cost**: Multi-turn sessions share a growing prefix, and dispatching the deepest ones keeps that prefix resident instead of re-prefilling it.
+* **Fairness IDs identify sessions**: Traffic carries a session header, so one fairness ID is one session. This strategy is not the right choice for a pool whose requests share a single ID.
 
 ## What it does
 
@@ -31,29 +39,46 @@ Orders programs by least attained service, so a program that has consumed fewer 
 
 Orders programs by how deep their session is. A session that has already taken many turns most likely still has its prefix in the KV cache, so serving it again costs less prefill and finishing it frees the slot sooner.
 
-Depth alone would starve shallow sessions, so head wait is scored alongside it under `turnPriorityTimeWeight`:
+Head wait is scored alongside depth under `turnPriorityTimeWeight`, so a shallower flow ages up while it waits:
 
 ```
 score = turnNumber + turnPriorityTimeWeight * headWaitSeconds
 ```
 
-Head wait is unbounded, so a shallow program overtakes any deeper rival once it has waited long enough. The two terms are a raw count and raw seconds, so `turnPriorityTimeWeight` converts one second of waiting into a number of turns. A value tuned for one traffic profile does not carry over to a workload with a different depth or wait distribution.
+The two terms are a raw count and raw seconds, so `turnPriorityTimeWeight` converts one second of waiting into a number of turns. A value tuned for one traffic profile does not carry over to a workload with a different depth or wait distribution.
 
-A program that has been idle beyond `turnPriorityInactivitySeconds` counts as turn one and re-earns depth, on the grounds that a session dormant that long has stopped competing for its own prefix. That threshold is separate from `evictionTtlSeconds`, which governs only when per-program bookkeeping is reclaimed.
+Head wait is bounded by the flow controller's request TTL, which defaults to 60s and is measured from the same enqueue time this score reads. A waiting request therefore accrues at most
 
-The default of `120` leaves a margin over the longest activity observed within a session, so a program still taking turns keeps its depth. On an agentic trace replay averaging 48 turns per session, mean think time between turns was about 2.5s and the slowest turn-to-turn cycle reached 96s under the heaviest load tested. A cycle includes service and queueing, so the post-completion gap this threshold measures is smaller. Workloads whose sessions pause for longer need a higher value.
+```
+turnPriorityTimeWeight * requestTTL
+```
+
+turn-equivalents before it is shed, and that product is the maximum depth a newcomer can overcome. The default weight of `0.05` against a 60s TTL gives 3, so a turn-one request tops out at a score of 4 and never overtakes a session at turn 5 or deeper. Under sustained contention a new session is shed rather than dispatched ahead of established ones.
+
+That is the intended trade-off at the default: capacity goes to sessions whose prefix is already resident, and new sessions are admitted as the depth of completing ones frees up. It also means the wait term is not a general starvation guard at this weight. Raising the weight so that `turnPriorityTimeWeight * requestTTL` exceeds the depth of the workload lets a newcomer overtake inside the TTL, at the cost of prefix retention: covering 30 turns at a 60s TTL takes `0.5`. Only the product matters, so a deployment that changes `requestTTL` has to revisit the weight.
+
+A program that has been idle beyond `turnPriorityInactivitySeconds` has its next request scored as turn one, on the grounds that a session dormant that long has stopped competing for its own prefix. Once that request is dispatched it re-prefills the whole conversation, and the program competes at its full depth again. The idle gap is measured from the program's last completion to the head request's arrival, so a request that has since been queuing does not itself age the program into a reset.
+
+The default of `120` leaves a margin over the longest gap observed within an active session, so a program still taking turns keeps its depth. On an agentic trace replay averaging 48 turns per session, mean think time between turns was about 2.5s and the slowest turn-to-turn cycle reached 96s under the heaviest load tested. That cycle figure includes service and queueing, while the gap this threshold measures spans only last completion to next arrival, so the effective margin is wider than the 24s difference suggests. Workloads whose sessions pause for longer between turns need a higher value.
 
 Under contention a prefix is evicted by capacity pressure while its program is still active, and ordering by turn depth is what keeps those prefixes resident. A low prefix-cache hit rate is therefore weak evidence for lowering this value.
 
+This strategy requires one fairness ID per session. It reads a program's dispatched count as the depth of a single conversation, which holds only when the ID identifies one session: the fairness ID header, or the `agent-identity` attribute that the request-header plugin derives from a session header.
+
+Requests carrying neither share `default-flow`. Its dispatched count is the sum over unrelated clients, so it is not a session depth, it grows with total unlabeled traffic, and it rarely stays idle long enough for the inactivity reset to fire. Ordering it against real sessions by that count is not meaningful, and a pool that mixes labeled and unlabeled traffic gives `default-flow` a depth no single session can match. Route unlabeled traffic to a priority band on a different fairness policy.
+
 ## Unit of Fairness
 
-**Attained service**, measured as a weighted sum of input and output tokens consumed by the program. Output tokens are weighted twice as much as input tokens to reflect their relative compute cost.
+Under `turn-priority`, **session depth**: the count of requests already dispatched for the program, plus the waiting request.
+
+Under `las`, **attained service**, measured as a weighted sum of input and output tokens consumed by the program. Output tokens are weighted twice as much as input tokens to reflect their relative compute cost.
 
 ## Inputs consumed
 
 * **Program identity**: From the request's `FairnessID` field, which the framework populates from the fairness ID header.
 * **Queue state**: Reads queue length and the head item's enqueue time from the `FlowQueueAccessor`.
-* **Token usage**: From the `Response.Usage` field on stream completion.
+* **Token usage** (`las`): From the `Response.Usage` field on stream completion.
+* **Dispatched count, in-flight count, last completion time** (`turn-priority`): From the per-program metrics maintained across the request lifecycle.
 
 ## Configuration
 
@@ -79,8 +104,8 @@ flowControl:
 | `lasWeightService` | `0.8` | Weight on the inverted attained-service signal. Higher values prioritize underserved programs more aggressively. |
 | `lasWeightHeadWait` | `0.2` | Weight on the head-of-queue age. Acts as a tiebreaker on cold start when programs have equal attained service. |
 | `lasHalfLifeSeconds` | `60` | Wall-clock half-life of attained service. `0` disables decay, making attained service cumulative for the program's lifetime. |
-| `turnPriorityTimeWeight` | `0.05` | (`turn-priority`) Score contribution per second of head wait, against one point per turn of depth. `0` orders on depth alone. |
-| `turnPriorityInactivitySeconds` | `120` | (`turn-priority`) A program idle for longer than this counts as turn one. `0` disables the reset. |
+| `turnPriorityTimeWeight` | `0.05` | (`turn-priority`) Score contribution per second of head wait, against one point per turn of depth. `turnPriorityTimeWeight * requestTTL` bounds the depth a waiting request can overcome, which at the default is 3 turns. `0` orders on depth alone, with longer head wait breaking ties. |
+| `turnPriorityInactivitySeconds` | `120` | (`turn-priority`) A program idle for longer than this has its next request scored as turn one. `0` disables the inactivity reset. Depth is still lost when the program is evicted after `evictionTtlSeconds`. |
 | `evictionTtlSeconds` | `3600` | A program with no completion in this window is evicted from the metrics map. |
 | `evictionSweepSeconds` | `300` | How often the eviction sweep runs. Must be `> 0`. |
 
@@ -88,7 +113,7 @@ A complete sample is shipped at [`deploy/config/sim-program-aware-config.yaml`](
 
 ## Observability
 
-The plugin exports two shared collectors and one strategy-owned collector under the `llm_d_epp` Prometheus subsystem:
+The plugin exports two shared collectors under the `llm_d_epp` Prometheus subsystem, plus one owned by the `las` strategy. Under `turn-priority` the strategy registers no collectors, so `program_aware_attained_service_tokens` is absent:
 
 | Metric | Type | Labels | Description |
 |---|---|---|---|
@@ -106,6 +131,8 @@ mean the scheduler is scoring the program on stale service.
 * **Abandoned requests block eviction**: Requests abandoned after dispatch leave `inFlight` non-zero, and the eviction sweep skips any program with non-zero `inFlight`, so its `ProgramMetrics` entry and Prometheus series persist indefinitely.
 * **Memory and label-series growth**: A new program ID adds a `ProgramMetrics` entry plus per-program Prometheus label series. The eviction sweep bounds growth, but a workload with rapidly churning program IDs (e.g. a fresh ID per request) will see TTL-bounded accumulation. Choose a TTL that matches your churn rate.
 * **Turn-priority weighting is scale-dependent**: the score sums a turn count and a wait in seconds, so `turnPriorityTimeWeight` converts between the two rather than balancing them proportionally. A value tuned for one traffic profile does not carry over to another with a different depth or wait distribution.
+* **Turn-priority sheds new sessions under contention**: a waiting request accrues at most `turnPriorityTimeWeight * requestTTL` turn-equivalents before the flow controller sheds it, 3 at the defaults, so a session deeper than that is never overtaken by a newcomer. Raising the weight to cover the depth of the workload trades prefix retention for admission latency.
+* **Turn depth is read one dispatch behind**: the dispatched count is incremented after the pick, off the dispatch path, so consecutive picks for one program can score against a count that has not yet caught up.
 * **Decay accrues during generation**: Decay is continuous wall-clock time, including while a program's requests are in flight; the service consumed by an in-flight request is added back, post-decay, when it completes. With half-lives well above typical generation times this effect is negligible.
 
 ## Related Documentation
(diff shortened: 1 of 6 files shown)


ruocco:
CC: @praveingk @vMaroon

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
```diff
@@ -17,7 +17,9 @@ Choose this policy when:
...
-* **Selects a queue to dispatch from** using the configured strategy. Currently `las` (Least Attained Service) is supported; programs with the lowest accumulated service score highest.
+* **Selects a queue to dispatch from** using the configured strategy. Two are supported:
+  * `las` (Least Attained Service): programs with the lowest accumulated service score highest.
+  * `turn-priority`: scoring engages only under contention. With no flow waiting nothing is dispatched; with exactly one waiting flow it is dispatched directly. When two or more flows are waiting, each scores by its head request's `prio = turn_number + turnPriorityTimeWeight * time_elapsed_in_seconds`, highest first. The elapsed term ages waiting flows up so lower-turn flows are not starved. Turn number is the program's dispatched-request count plus the waiting request itself: each program (fairness ID) has its own counter, advanced once per dispatch. The counter resets when idle program state is evicted: a program idle past the eviction TTL has likely lost its KV cache, so on return it restarts from turn one and re-earns priority from a cold state.
```
shmuelk, line 22: I'm fairly sure that individual instances of an agentic session will NOT get their own fairnessID. There is a queue in flowcontrol per fairnessID (per priorityBand). In the use case you are describibg there would be a single element in each queue, being very wastefull.

The original fairness policy was, I think, looking at each agentic workload as a whole, that is all of the requests for the same workload, vs requests for a different workload.

This PR if I understand is trying to look at the individual sessions and prioritize them. This sounds more like an in queue prioritization policy, rather than a fairness policy.

ruocco: Thanks @shmuelk . 

You are correct, this PR is about scheduling sessions, and prioritizing some (the more advanced) in place of others.

Is your concern the fact that each session has its own queue, with (most of the times) only one element inside being the active request, and you are suggesting that it would be more efficient to have a single queue, with sessions as elements?

What would you recommend? Moving away from Fairness and into a more generic OrderingPolicy?

shmuelk: The main point I was trying to raise is that the fairnessID is NOT a sessionID. You need to track by some sort of sessionID (there are other PRs that have dealt with sessions and sessionIDs in the past).

You are tracking by program or agenticc workload level and not what you wanted.

praveingk: @shmuelk I guess you are referring to this PR https://github.com/llm-d/llm-d-router/pull/1754/  
This assigns fairnessID based on the agent's session ID parsed  from the header of a request.

shmuelk: I would claim that an agentID as described in the PR is not a sessionID. It identifies the agent, not the session with the agent.

ruocco: @shmuelk , maybe I hooked into the wrong handle, but at the moment the fairnessID is tied to each session, and this commit achieves the intended behaviour, at least for the OTEL traces replay via inference-perf.

If there's a better way on getting a session hook, or a better place in llm-d where to implement this, I'm happy to refactor.

ruocco:
Adding comparison plots between this PR vs a normal deployment:

comparison_request_latency
comparison_session_duration
comparison_throughput
comparison_vllm_cache_prefix_cache_hit_rate_running

- **GPU**: H100 80G
- **Model**: Qwen3-30B-A3B-Instruct-2507 (256K context len)
- **Tensor parallelism**: 4
- **Replica**: 1
- **Traces**: Exgentic/agent-llm-traces-v2

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
praveingk: @shmuelk Actually the agentID is the sessionID derived from `session-id` in the request header. Maybe, we should rename it to `agent-session-id` to remove confusion in future.

LukeAVanDrie:
@ruocco sorry for the delay on this PR. Will block off time to review this week so we can land this in v0.11

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
praveingk: We can omit "scoring engages only under contention", since its the same for all strategies.
Also, I feel, it would be better to simplify the description to just provide the intuition.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
```diff
@@ -0,0 +1,94 @@
...
+	fwksched "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
+)
+
+const turnPriorityStrategyName = "turn-priority"
+
+var _ Strategy = &turnPriorityStrategy{}
+
+// turnPriorityStrategy scores each flow by its head request's priority
+//
+//	prio = turnNumber + timeWeight * elapsedSeconds
```
praveingk: should `elapsedSeconds` be called `waitTime` or  `headWait` for simplicity?

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
```diff
@@ -0,0 +1,94 @@
...
+		return waiting[0].Queue
+	}
+
+	var best flowcontrol.FlowQueueAccessor
+	bestScore := math.Inf(-1)
+	now := time.Now()
+
+	for _, qi := range waiting {
+		head := qi.Queue.Peek()
+		elapsed := now.Sub(head.EnqueueTime()).Seconds()
```
praveingk: This can be simplified to use `time.Since()`

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (s *turnPriorityStrategy) Collectors() []prometheus.Collector { return nil }
+
+// turnNumberFor returns the turn number of a program's head request: the count
+// of requests already dispatched for that program plus the waiting request
+// itself (dispatchedCount+1).
+//
+// The counter is program-scoped: each program (fairness ID) has its own count,
+// and every dispatch advances it by one. The count resets to zero when idle
+// program state is evicted: a program that goes idle past the eviction TTL has
```
praveingk: I think here we are mixing two things: 
1) Garbage collection/eviction time. The metrics are evicted based on a configured time, default to `3600`
2) Program inactive time. 

May be the latter should be configurable separately in this plugin so that it's not tied to the eviction cycle?

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
```diff
@@ -0,0 +1,94 @@
...
+	bestScore := math.Inf(-1)
+	now := time.Now()
+
+	for _, qi := range waiting {
+		head := qi.Queue.Peek()
+		elapsed := now.Sub(head.EnqueueTime()).Seconds()
+		if elapsed < 0 {
+			elapsed = 0
+		}
+		score := float64(turnNumberFor(qi.Metrics)) + s.timeWeight*elapsed
```
praveingk: Since both are different metrics (turn count, and seconds), I think they should be normalized using https://github.com/llm-d/llm-d-router/blob/229b8ea5f56a99c732b02506cbd733cbc3d13cea/pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy.go#L43  to arrive at a score.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
```diff
@@ -0,0 +1,94 @@
...
+	case 1:
+		return waiting[0].Queue
+	}
+
+	var best flowcontrol.FlowQueueAccessor
+	bestScore := math.Inf(-1)
+	now := time.Now()
+
+	for _, qi := range waiting {
+		head := qi.Queue.Peek()
```
praveingk: better to do a `nil` check.

praveingk (changes requested):

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
ruocco: Renamed to headWait.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
ruocco: Done

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
ruocco: Added an explicit head == nil check after Peek(). This also removes the second Peek() call that was in the scoring loop.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
ruocco: I tried rangeNormalize and it doesn't work here, so this PR keeps the formula unnormalized and a redesign follows separately.
The problem is that min-max normalization is relative to whoever is contending in that pick. It preserves ordering but discards magnitude, and the magnitude is the part that matters for starvation.

The fix isn't normalization but a ratio-based design: have the timeWeight (to be renamed) be a way of skewer the pick so that N elements are taken from the longest waiting, and (1-N) from the deepest session. That's a large enough change that I'd rather land it as a follow-up PR than expand this one. I'll be testing this soon.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
ruocco: Introduced `inactivitySeconds`, as a way of tagging a request that has been idle too long

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
ruocco: Rephrased

ruocco:
@praveingk Thank you for the review. I've addressed most of your concerns and requests for changes (except notably the normalization one). And rebased. Let me know how it's looking now.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
praveingk: makes sense 👍🏽

praveingk (approved):
@ruocco Thanks for addressing the comments. Looks good to me.

praveingk:
@LukeAVanDrie @RishabhSaini Can you please help reviewing the PR?

ahg-g:
This is nice! Did we try to evaluate this algorithm using agentx traces?

ruocco:
> This is nice! Did we try to evaluate this algorithm using agentx traces?

@ahg-g not yet, but planning to, as they are a better fit for this, with deeper turns.

ruocco:
Some new results using the Weka traces (2.5h timeout) and bumping the context length of the model to 1M. Here's the numbers:

weka_legend
weka_aggregated

KV Cache visualization:

weka_kvcache_visual

Takeaways:

- At low concurrency, or with high weight, the policy does not hurt, it falls back to FCFS behaviours
- With contention and proper weight, it helps with several metrics, and doesn't show any regression signs

animeshtrivedi:
@LukeAVanDrie and @RishabhSaini -- can you also please have a look when you get time? Any input on this would be appreciated. Thanks!

ruocco:
**To Recap:**

The intuition behind this strategy implementation for the program-aware-fairness scheduler is to prioritize the agentic sessions that have made more progress, i.e. ordering waiting flows by session depth rather than arrival order. An advanced session is worth dispatching ahead of a fresh one for two reasons:

- its prefix is most likely still resident in the KV cache, so serving it again costs less prefill;
- it is closer to completion, so finishing it frees the slot sooner.


This ideas has been explored in scientific literature, for example: https://arxiv.org/pdf/2511.02230

Empirical results show that under KV Cache pressure, the turn-priority scheduler performs significantly better than "First come first served" (FCFS, default in llm-d) in every metrics, while providing no downsides when no cache pressure exists. In particular, the latest experiments, replaying the [weka traces](https://huggingface.co/datasets/semianalysisai/cc-traces-weka-with-subagents-051926), Qwen3-30B-A3B-Instruct-2507, TP4 on H200, 1M context, 512 GiB CPU offload, show:

extensive

**The KV Cache hit rate behind the improvements:**

When looking at the hit rate at high concurrency levels, the turn-based priority scheduler manages to retain a very high hit rate, while FCFS collapses and has to recompute most requests.

weka_kvcache_visual

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
```diff
@@ -0,0 +1,110 @@
...
+// re-earns depth from a cold state: a session dormant that long has stopped
+// competing for its own prefix. The threshold is independent of the metrics
+// eviction TTL, which governs only when per-program bookkeeping is reclaimed.
+func (s *turnPriorityStrategy) turnNumberFor(metrics *ProgramMetrics) int64 {
+	if metrics == nil {
+		return 1
+	}
+	if s.inactivitySeconds > 0 && metrics.InFlight() == 0 {
+		last := metrics.LastCompletionTime()
+		if !last.IsZero() && time.Since(last).Seconds() > s.inactivitySeconds {
```
LukeAVanDrie: `time.Since(last)` is think time plus the head request's queue wait. For a sequential session the next request is enqueued right after the previous completion, so once queue wait passes `inactivitySeconds` this returns 1 for every deep session that is not running a parallel request, and ordering degenerates to `timeWeight * headWait` (FCFS). Sessions with `InFlight > 0` are exempt, so fan-out sessions keep their depth and sequential ones lose it.

Reproduction: `seedTurns(40, now.Add(-130*time.Second))` with the head enqueued at `now.Add(-129*time.Second)` and `inactivitySeconds: 120` returns `turnNumberFor== 1`.

The gap you likely want is last completion to head arrival. `Pick` already has the head, so pass its enqueue time down and sample `now` once, as LAS does:

```
func (s *turnPriorityStrategy) turnNumberFor(metrics *ProgramMetrics, headEnqueue time.Time) int64 {
	if metrics == nil {
		return 1
	}
	if s.inactivitySeconds > 0 && metrics.InFlight() == 0 {
		last := metrics.LastCompletionTime()
		if !last.IsZero() && headEnqueue.Sub(last).Seconds() > s.inactivitySeconds {
			return 1
		}
	}
	return metrics.DispatchedCount() + 1
}
```

and in `Pick`, keep `head.EnqueueTime()` on the candidate and score with `s.turnNumberFor(c.metrics, c.headEnqueue)`.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
```diff
@@ -17,10 +17,34 @@ Choose this policy when:
...
+
+Orders programs by how deep their session is. A session that has already taken many turns most likely still has its prefix in the KV cache, so serving it again costs less prefill and finishing it frees the slot sooner.
+
+Depth alone would starve shallow sessions, so head wait is scored alongside it under `turnPriorityTimeWeight`:
+
+```
+score = turnNumber + turnPriorityTimeWeight * headWaitSeconds
+```
+
+Head wait is unbounded, so a shallow program overtakes any deeper rival once it has waited long enough. The two terms are a raw count and raw seconds, so `turnPriorityTimeWeight` converts one second of waiting into a number of turns. A value tuned for one traffic profile does not carry over to a workload with a different depth or wait distribution.
```
LukeAVanDrie: Head wait is bounded by the flow controller's request TTL. `DefaultRequestTTL` is 60s and expiry is measured from the same enqueue time this score reads. At the default weight of 0.05 a request accrues at most `0.05 * 60 = 3` turn-equivalents before it is shed, so a turn-1 request tops out at 4 and never overtakes a session at turn 5 or deeper. Under contention every request of a new session expires undispatched. 

The bound `turnPriorityTimeWeight * requestTTL` is the maximum depth a newcomer can overcome. The doc has to state it and the default weight has to be chosen with it in mind. The 0.5 the issue originally proposed gives 30 turns at a 60s TTL, which covers the 48-turn average you cite. If shedding newcomers under contention is the intended trade-off, that is also a valid answer, but it should be written down here and the test adjusted to a wait that fits inside the TTL.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
```diff
@@ -0,0 +1,110 @@
...
+			return 1
+		}
+	}
+	return metrics.DispatchedCount() + 1
```
LukeAVanDrie, line 113: Needs a decision before merge. This is a lifetime count per fairness ID with no cap, and not every fairness ID is a session. Requests with no fairness header and no agent-identity attribute all land in `default-flow`. In a pool that mixes labeled agent sessions with unlabeled clients, `default-flow` gains one turn per request and is rarely idle for 120s, so the reset never fires. After a few thousand requests no real session can tie it inside the TTL. Tenant-style IDs and agents that fan out N parallel calls per turn have the same shape. Under LAS such an ID is penalized. 
Under turn-priority it is rewarded.

Either of these would close it:

- pin `metadata.DefaultFairnessID` to turn 1 here, or
- state in the README that turn-priority assumes per-session fairness IDs and that unlabeled traffic belongs in a band with a different policy.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
```diff
@@ -17,10 +17,34 @@ Choose this policy when:
...
+
+A program that has been idle beyond `turnPriorityInactivitySeconds` counts as turn one and re-earns depth, on the grounds that a session dormant that long has stopped competing for its own prefix. That threshold is separate from `evictionTtlSeconds`, which governs only when per-program bookkeeping is reclaimed.
```
LukeAVanDrie: "Re-earns depth" is not exactly what the code does. `DispatchedCount()` is never reset, so only the first request after the pause scores as turn one. Once it is dispatched the program is back at its full historical count (30-turn session, idle 10 min: 1, then 32, then 32). I think the code behaviour is the right one, since that first request re-prefills the entire conversation, but the doc should say so. Suggested wording if you keep the code as is:

```suggestion
A program that has been idle beyond `turnPriorityInactivitySeconds` has its next request scored as turn one, on the grounds that a session dormant that long has stopped competing for its own prefix. Once that request is dispatched it re-prefills the whole conversation, and the program competes at its full depth again. That threshold is separate from `evictionTtlSeconds`, which governs only when per-program bookkeeping is reclaimed.
```

The `turnNumberFor` doc comment ("re-earns depth from a cold state") needs the same correction. If you would rather implement the gradual re-earn, that needs a per-program baseline in the strategy. Either works. The two have to match.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
```diff
@@ -17,10 +17,34 @@ Choose this policy when:
...
+
+The default of `120` leaves a margin over the longest activity observed within a session, so a program still taking turns keeps its depth. On an agentic trace replay averaging 48 turns per session, mean think time between turns was about 2.5s and the slowest turn-to-turn cycle reached 96s under the heaviest load tested. A cycle includes service and queueing, so the post-completion gap this threshold measures is smaller. Workloads whose sessions pause for longer need a higher value.
```
LukeAVanDrie: "A cycle includes service and queueing, so the post-completion gap this threshold measures is smaller" is the intended semantics, and it becomes correct once the enqueue-time fix above lands. As written today the measured gap includes queueing, so the 96s cycle figure is the one to compare against and the margin is 24s. This paragraph needs a second pass after the code change.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
```diff
@@ -51,10 +75,12 @@ flowControl:
 
 | Field | Default | Description |
 |---|---|---|
-| `strategy` | `las` | Scoring strategy. Only `las` is supported. |
+| `strategy` | `las` | Scoring strategy: `las` or `turn-priority`. |
 | `lasWeightService` | `0.8` | Weight on the inverted attained-service signal. Higher values prioritize underserved programs more aggressively. |
 | `lasWeightHeadWait` | `0.2` | Weight on the head-of-queue age. Acts as a tiebreaker on cold start when programs have equal attained service. |
 | `lasHalfLifeSeconds` | `60` | Wall-clock half-life of attained service. `0` disables decay, making attained service cumulative for the program's lifetime. |
+| `turnPriorityTimeWeight` | `0.05` | (`turn-priority`) Score contribution per second of head wait, against one point per turn of depth. `0` orders on depth alone. |
+| `turnPriorityInactivitySeconds` | `120` | (`turn-priority`) A program idle for longer than this counts as turn one. `0` disables the reset. |
```
LukeAVanDrie: Non-blocking nit. "`0` disables the reset" is not quite true, and neither is line 42's "evictionTtlSeconds governs only when per-program bookkeeping is reclaimed". Eviction deletes the `ProgramMetrics` entry and `getOrCreateMetrics` recreates it at `DispatchedCount = 0`, so depth resets after `evictionTtlSeconds` regardless of this field. Under LAS the bookkeeping was a decayed-to-zero value, so the claim held. Under turn-priority the bookkeeping is the depth.

```suggestion
| `turnPriorityInactivitySeconds` | `120` | (`turn-priority`) A program idle for longer than this counts as turn one. `0` disables the inactivity reset. Depth is still lost when the program is evicted after `evictionTtlSeconds`. |
```

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
LukeAVanDrie, line 17: Non-blocking nit, but this bullet is wrong for turn-priority. It and the one above it ("a program issuing many small requests should not dominate") describe LAS. 

Under turn-priority a returning program scores at turn 1 against active programs at turn N, and the many-small-requests program is the one whose depth grows fastest. The "Unit of Fairness" section (line 50), "Token usage" under inputs (line 56), and "one strategy-owned collector" (line 91) are also LAS-only. Turn-priority's `OnCompleted` and `Collectors` are no-ops, so `program_aware_attained_service_tokens` is not registered with this strategy.

A "Under `las`" qualifier on each, or a short per-strategy split of the "Why choose" list, would fix all four.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
```diff
@@ -0,0 +1,110 @@
...
+func (s *turnPriorityStrategy) Pick(_ int, queues map[string]QueueInfo) flowcontrol.FlowQueueAccessor {
+	// Collect the flows with pending work: with no waiting flow there is nothing
+	// to dispatch, and with exactly one the choice is forced.
+	type candidate struct {
+		queue    flowcontrol.FlowQueueAccessor
+		metrics  *ProgramMetrics
+		headWait float64
+	}
+
+	waiting := make([]candidate, 0, len(queues))
```
LukeAVanDrie: Non-blocking nit. Nothing in this score depends on the other candidates (LAS needs the min/max pass, this does not), so the slice, the `switch` on its length, and the second loop can collapse into one best-tracking loop. The result is the same in every case: all empty gives nil, and one waiting gives that queue since `bestScore` starts at `-Inf`. That removes the per-Pick allocation and reduces up to two clock reads per candidate to one `now := time.Now()` sample. The enqueue-time fix above already changes `turnNumberFor`'s signature, so folding this in there avoids a second signature change.

Related, and also non-blocking: `DispatchedCount` is bumped in `PreRequest` on the request goroutine, well after `Pick`, and the processor re-runs the dispatch cycle on every arrival and a 1ms tick without waiting for it. A program with several queued items is scored against a stale count for consecutive picks. `plugin.Pick` already sees the dispatch synchronously, so the count could move there. At minimum a comment noting the lag.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
```diff
@@ -0,0 +1,110 @@
...
+	case 1:
+		return waiting[0].queue
+	}
+
+	var best flowcontrol.FlowQueueAccessor
+	bestScore := math.Inf(-1)
+
+	for _, c := range waiting {
+		score := float64(s.turnNumberFor(c.metrics)) + s.timeWeight*c.headWait
+		if score > bestScore {
```
LukeAVanDrie: Non-blocking nit. Equal scores are resolved by map iteration order, so with `timeWeight: 0` (documented as "orders on depth alone") two same-depth programs are a coin flip rather than FCFS. LAS has the same property, so this is not a regression, but a tie-break on longer head wait is one line and gives the zero-weight mode FIFO order within a depth:

```suggestion
		if score > bestScore || (score == bestScore && c.headWait > bestWait) {
```

with `bestWait` tracked alongside `bestScore`.

LukeAVanDrie (dismissed):

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
LukeAVanDrie: Closing this thread with what the code does today.

The Director sets `FairnessID` from the explicit fairness header when present. Otherwise it reads the `agent-identity` request attribute. That attribute is produced by the `agent-identity` request-header plugin from session headers: `x-claude-code-session-id`, `x-session-affinity`, and Codex's `session-id`. The key is named for the plugin, but the value is a per-session ID. So for traffic that goes through that plugin, one fairness ID is one session, and the strategy tracks what it intends to.

The cost raised here, one flow-control queue per session with usually one item in it, is real but predates this PR. It comes from #1754 choosing fairness ID as the session handle. #2120 made dispatch scan only active queues, which removes most of the per-idle-queue cost. Whether session prioritization belongs in a fairness policy or in an ordering policy is a design question for flow control as a whole, and it should be taken in an issue rather than block this PR. We are still working on the primitives alignment here, so I wouldn't block this PR on it.

The case where a fairness ID is not a session (no header, no agent-identity, so everything lands in `default-flow`) does matter for this strategy. That is tracked in another thread, so resolving this one.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
LukeAVanDrie: Agree that min-max normalization is the wrong tool here. One thing this thread did not account for is that the magnitude argument only holds while a request is still queued. The flow controller sheds at `DefaultRequestTTL`, 60s by default, so the wait term is capped at `timeWeight * 60`. I wrote that up in another thread.

The ratio-based design you describe (a fixed share of picks from the longest-waiting flow, the rest from the deepest) would sidestep both the unit mismatch and the TTL cap, since the share does not depend on how long anyone has waited. Could you file an issue for that follow-up so it is tracked? Then this PR only needs to state the bound and pick a default consistent with it.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
praveingk: Agree 👍🏽 I am in favor of the second option right now.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
ruocco: Fixed. `turnNumberFor` takes the head's enqueue time and measures `headEnqueue.Sub(last)`.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
ruocco: It's a trade-off, and the "optimum" depends on the workload profile and the intended behaviour, that's why it's kept, at least for now, as a tunable. We found the best performance are achieved with low weights, and that was the reasoning behind the default. I've added some comments and adjusted the test.

Note that with deep traces, with flowControl and defaultRequestTTL to 60s, requests _will_ be dropped _regardless_, probably even more so. In our analysis, when the system is fully saturated with parallel replay of weka_traces, e.g. 128 concurrency, the average wait time for a request with is ~180s with w0.05 and ~360s with FCFS.


Related: we'll be also opening an issue soon with some proposals to improve the formula, and minimise the shedding.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
ruocco: The README now also states the per-session assumption.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
ruocco: Fixed.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
ruocco: Rewritten.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
ruocco: Fixed.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/README.md
ruocco: Changed to be "Under las".

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
ruocco: Done together with the enqueue-time change.

Also, I added a comment for `DispatchedCount` noting the lag, since changing pick would affect LAS as well. If we find this to be a bigger issue than thought, we should implement a proper counter for the turn-priority strategy.

pkg/epp/framework/plugins/flowcontrol/fairness/program-aware/strategy_turnpriority.go
ruocco: Fixed.

ruocco:
@LukeAVanDrie thanks for the review. I have addressed all your comments and pushed a commit with the relevant fixes.

@praveingk thanks for the additional feedback!

ahg-g (approved):

Issue #2106, opened 2026-07-21
[Flow Control] Decide a cardinality policy for caller-controlled flow control labels


LukeAVanDrie:
**What would you like to be added**:

A recorded cardinality policy for caller-influenced metric labels, plus implementation. Options (can differ per label):

1. Drop `fairness_id` from metric labels (keep it in logs/traces).
2. Hash/bucket it (bounded, loses readability).
3. Allow-list via config; overflow to `other`.
4. Keep as-is, `DeleteLabelValues` on flow GC, documented operator guidance (weakest).

Also in scope: bound the `priority` label or document the exposure; set a deprecation timeline for the duplicated `inference_extension_*` metric twins (every flow control metric except `requests_total` is emitted twice, doubling series).

**Why is this needed**:

With flow control default-on, `fairness_id` becomes an unauthenticated, caller-controlled, unbounded label: it is the raw client header `x-llm-d-inference-fairness-id` (`pkg/epp/metadata/consts.go:34`, read at `director.go:265`). `model_name`/`target_model_name` come from the request body. No limiting or allow-listing exists for `fairness_id`, and series are never deleted on flow GC. Net: a metrics-cardinality DoS against EPP memory and the metrics pipeline. Part of #1187; the flip bar is the recorded decision — implementation may fast-follow.

LukeAVanDrie:
@liu-cong , @gyliu513 , @ahg-g 

Here is my proposal for v0.10:
 
- `fairness_id`: bounded via the `boundedLabel` mechanism that already caps model labels (1000 distinct values, overflow to `other`) applied at every record site, both metric families. Flow control series are also pruned (`DeletePartialMatch`) when the registry GCs a flow. This is option 3 minus the config allow-list, plus option 4. Memory and scrape size are capped no matter what callers send.
- `model_name`/`target_model_name`: already bounded; the one gap (`RecordFlowControlRequestQueueDuration` skipped `boundModels`) must be fixed.
- `priority`: Since it resolves from `InferenceObjective.Spec.Priority` and flow control rejects unprovisioned bands, cardinality is bounded by configured objectives. So, no cardinality cap needed here.
- `inference_extension_*` twins: already tracked in #1070, so no action under this issue.

Two accepted limitations, consistent with the existing model-label tradeoffs:

[...]

Issue #1187, opened 2026-05-18
[Flow Control] Production readiness: graduate flow control to GA (opt-in)


LukeAVanDrie:
Tracks graduating the flow control layer (#1270) to GA. Code references pinned to `6cc1dac7`.

**Posture.** Flow control is GA but explicitly opt-in: saturation detection needs per-deployment tuning, so enabling the layer is a deliberate operator decision (decision: https://github.com/llm-d/llm-d-router/pull/2180#issuecomment-5119639333). The default flip (#2104) is deferred, with a concrete revisit condition: saturation detection no longer requiring per-deployment tuning, which is the direction that work is headed. The revisit matters long term because every release with the gate off grows the behavioral divergence between the enabled and disabled paths.

**Scope.** This issue owns the router: hardening, router perf numbers, and router-side observability and reliability for the opt-in path. Guide posture, deployment docs, and guide-level validation are llm-d/llm-d#2114.

### Remaining work

Boxes are checked on merge, not on PR open. Each item is a small, independently revertible PR; they are independent and can be in flight concurrently. All of it applies to the opt-in path; none of it is gated on a default flip.

- [x] **A** #2095 -- `DefaultRequestTTL` and per-band `MaxRequests` have no effect
- [x] **B** #2097 -- pre-admission TTL/cancel returns 500; registry lookup mislabeled as 429 (PR #2162, in review)
- [x] **C** #2101 [@ishwar170695] -- log quality pass; request drops visible at default verbosity
- [x] **E** #2100 -- stale-metrics saturation silently halts all dispatch; `pool_saturation` semantics (PR #2165, in review)
- [ ] **F** #2102 [@sudoalok] -- per-band and global queue utilization gauges (PR #2184, in review)
- [x] **G** #2103 -- feature-gate wiring tested in both directions
- [x] **H** #2106 -- cardinality policy for caller-controlled flow control labels
- [x] **I** #2092 -- datastore retries dropped endpoint registrations
- [ ] **J** -- router numbers: `pkg/epp/flowcontrol` benchmarks and the e2e suite with the gate on. Not started. Still wanted: this is the sizing evidence an operator needs before opting in, and the baseline for any future default-flip discussion. The guide-level A/B at saturating load is llm-d/llm-d#2114 section 2.
- [ ] **K** -- graduation (PR #2180): de-experimentalized naming, a loader warning when a `flowControl:` section is set while the gate is off, enablement/tuning/sizing docs.

[...]

PR #2119, opened 2026-07-21
fix(metrics): bound fairness_id label cardinality and prune GC'd flow series

LukeAVanDrie:
**What type of PR is this?**

/kind bug

The `fairness_id` label on request and flow-control metrics is populated from a client request header (`x-llm-d-inference-fairness-id`) or an agent-identity attribute, so its cardinality is not operator-bounded. Prometheus vectors never evict label combinations: every distinct fairness ID ever observed permanently allocated histogram series across both the deprecated and `llm_d_epp` metric families, growing pod RSS and scrape payloads without bound. This is a metrics-cardinality DoS against EPP memory and the metrics pipeline. Implements the cardinality policy proposed in #2106 (my proposal, awaiting maintainer sign-off). Sending the implementation alongside to make the proposal concrete and reviewable; if the discussion lands on a different option, I'll rework this PR to match.

Three changes, following the existing model-label precedent:

- Bound `fairness_id` through a `boundedLabel` limiter (cap 1000, overflow to `other`) in every record function that takes it, exactly as model names are bounded via `boundModels`.
- Bound model labels in `RecordFlowControlRequestQueueDuration`, which was missing the `boundModels` call its sibling functions have (model names come from the request body).
- Prune a flow's flow-control series from both metric families when the registry garbage-collects the flow, so the vectors track live flows. Pruning runs after `cleanupFlowResources` and outside the registry lock; `DeletePartialMatch` scans whole vectors and must not run under it.

Accepted limitations (consistent with the model-label tradeoffs; see #2106 for the rationale): cap admission is first-come rather than allow-listed, and cap slots are never recycled. Memory stays bounded either way; per-tenant observability can degrade under a deliberate ID flood.

FullPath benchmark (registry/flow churn, M4 Pro, `-benchtime=3s`): 14.5KB -> 4.0KB per op, 264 -> 94 allocs per op.

Per #2106; part of #1187.

**Release note**:

```release-note
The fairness_id metric label is now capped at 1000 distinct values; additional fairness IDs are reported under the `other` label value. Flow control metric series for a fairness ID are removed when its flow is garbage collected. This bounds EPP memory and metrics-pipeline load, which were previously unbounded because the label derives from a client-controlled header.
```

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -63,6 +63,12 @@ Names below omit the subsystem prefix. Unless a section states otherwise, the pr
 and the release stage is ALPHA. Request and latency metrics share the label set
 `{model_name, target_model_name, fairness_id, priority}`.
 
+Client-derived label values are cardinality-bounded: `model_name` and `target_model_name` (from the
+request body) and `fairness_id` (from the `x-llm-d-inference-fairness-id` header) are each capped at
+1000 distinct values, after which new values are reported as `other`. Model names configured through
+InferenceModelRewrite rules never fold to `other`. Flow control series for a `fairness_id` are
+removed when its flow is garbage collected.
+
 ### Request and latency
 
 | Name | Type | Notes |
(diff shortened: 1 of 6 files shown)


LukeAVanDrie:
cc: @liu-cong

liu-cong (approved):

PR #2120, opened 2026-07-21
perf(flowcontrol): index active queues so dispatch scans O(active) flows

LukeAVanDrie:
**What type of PR is this?**

/kind cleanup

Every dispatch cycle, the fairness policy's `Pick` calls `PriorityBandAccessor.IterateQueues`, which snapshotted every registered queue in the band (empty ones included) into a freshly allocated slice under the registry read lock. Per-dispatch cost was O(registered flows) in CPU and allocation (94-98% of all allocation in the flow-control benchmarks) and throughput decayed superlinearly under flow churn as registered-but-idle flows accumulated between GC cycles. The 1 ms dispatch ticker pays this cost even at zero traffic.

Each priority band now maintains a lock-free index of its non-empty queues, updated on empty<->non-empty transitions inside the queue's existing stats critical section (transitions are serialized per queue by the queue mutex; the index is a `sync.Map`, so the update never touches the registry mutex). `IterateQueues` ranges the index directly: no registry lock, no allocation, cost proportional to flows that currently hold items.

This narrows `IterateQueues` to active (non-empty) queues, matching its documented contract ("each active Flow"). There are four production callers. `globalstrict.Pick` and reclamation's `bandHasQueuedItems` already skip empty queues, and the processor's cleanup/drain sweeps are no-ops on them. The `program-aware` LAS policy, however, used empty-queue visits to drive attained-service decay, so the second commit makes decay lazy: it is folded in from the decay anchor whenever a program's service is read or accumulated, independent of visits.

Wall-clock decay reparameterizes the knob, so the third commit removes `lasDecayFactor`. It was a per-`Pick` factor, and the only reading available under wall-clock decay is per-second, which is a different setting carrying the same name. The `lasHalfLifeSeconds: 60` default would then override it and discard the operator's tuned value with no signal. Decay is parameterized by `lasHalfLifeSeconds` alone, where `0` disables it. Plugin parameters decode with `DisallowUnknownFields`, so a config still carrying `lasDecayFactor` fails at startup with the key named. @D-Sai-Venkatesh @praveingk, flagging for LAS review.

The view is eventually consistent; callbacks must tolerate a visited queue reporting `Len() == 0`, which they already did under the snapshot implementation (a snapshotted queue could drain before the callback ran).

Index deactivation uses `CompareAndDelete` rather than a plain delete: a cleanup-sweep worker can drain a queue through a handle resolved before `deleteFlow` removed it, and if a successor queue was registered under the same flow ID in the interim, an unconditional delete would hide the live successor from dispatch and future sweeps. A regression test covers this interleaving.

Note for reviewers: iterating under the registry read lock instead was considered and rejected. The cleanup sweep calls `registry.ManagedQueue()` (which takes the read lock) from inside the iteration callback, so yielding under the lock is a recursive `RLock` that deadlocks when a writer is queued.

Benchmarks (`PerformanceMatrix`, M4 Pro):

| Config | Before | After |
|---|---|---|
| L=0/P=1/F=5000/W=10 | 25.2k d/s, 56.1KB/op | 370k d/s, 1.7KB/op |

Every F=5000 configuration improves 6-15x. At F=10 the before and after runs are statistically indistinguishable (`benchstat`, `-count=6`, all p > 0.05).

Part of #1187

**Release note**:

```release-note
Flow control dispatch cost no longer scales with the number of registered flows: fairness policies scan only non empty queues, removing a per-dispatch allocation proportional to flow count. Action required for the `program-aware-fairness` plugin: the `lasDecayFactor` parameter is removed and a config that still sets it will fail startup. Attained-service decay is now wall-clock based and parameterized solely by `lasHalfLifeSeconds` (default 60, `0` disables decay); decay no longer pauses while requests are in flight. Action required for out-of-tree fairness policies: `PriorityBandAccessor.IterateQueues` now visits only non-empty queues. A policy that relied on a callback for every registered queue, for example to age per-flow state, must drive that work from elapsed time.
```


LukeAVanDrie:
/assign @shmuelk 

After this, it's probably not worth making any more targeted perf improvements as we are definitely not a bottleneck anymore. My main goal was to reduce unnecessary temp allocs which create GC pressure which can bleed out and affect perf of other parts of EPP.

LukeAVanDrie:
Found some issues, converting this back to draft while I rework this.

edit: Done and moved back out of DRAFT

LukeAVanDrie:
cc: @D-Sai-Venkatesh and @praveingk

This perf improvement necessitates changing an aspect of the program aware fairness plugin. Would like your feedback on this to make sure my changes are aligned with your intentions for how decay works.

D-Sai-Venkatesh:
Hey @LukeAVanDrie, sure will take a look and post our comments here.

LukeAVanDrie:
> Hey @LukeAVanDrie, sure will take a look and post our comments here.

Thanks, no urgency here as this is not a release blocker for v0.10. I just fixed up the go-bench recently, got fresh results, and sent a fix out for the worst cpu and mem bottlenecks.

shmuelk:
@LukeAVanDrie Please rebase. It will fix the CI failure

LukeAVanDrie:
cc: @jtechapps 

I have my micro-bench framework for the flow control runtime itself, but I was wondering if I am able to use your new perf tooling to investigate overall EPP perf impact _before_ merging (I know it's a nightly).

ahg-g:
All related PRs merged, can you pls rebase

LukeAVanDrie:
@ahg-g rebased and ready for review; I would like sign-off from @D-Sai-Venkatesh first as this is a breaking change to the program aware fairness plugin.

D-Sai-Venkatesh:
hey @LukeAVanDrie. I have looked at the changes to program aware plugin and they make sense to me. We were also planning to remove the decay based on factor, thanks for that. I have also run an experiment to compare with previous version and the results are as expected did not see any discrepancies. Thanks for updating the plugin to keep it in sync with flow control changes.

ahg-g (approved):

Issue #1188, opened 2026-05-18
[Testing] Consolidate Hermetic Integration Test Helpers into Structured test/framework


LukeAVanDrie:
Integration test helpers are scattered across `test/integration/util.go`, `test/integration/epp/harness.go`, and `pkg/epp/util/testing/wrappers.go`. This causes two problems:

1. Poor discoverability: builders live in leaf packages (fluent mock builders inside EPP packages, ext_proc gRPC builders in `test/integration`), so contributors re-implement them.
2. Cyclic dependency risk: hermetic tests under `test/integration/epp` import the root `test/integration` package for gRPC mock builders. If utilities in `test/integration` ever import domain-level structures, they trigger Go import cycles.

The fix is a structured `test/framework/` tree that separates generic K8s/net utilities from EPP/Envoy domain-specific builders.

### Design

The layout and import-layering rules (R1 through R5) live in [`test/framework/doc.go`](https://github.com/llm-d/llm-d-router/blob/main/test/framework/doc.go). Two placements need explanation:

- GAIE CRD builders get their own package, `test/framework/gaie`. The in-package test in `pkg/epp/datastore` calls `MakeInferencePool`, and the envtest harness imports `cmd/epp/runner`, which imports `pkg/epp/datastore`. Housing the builders and the harness in one package would pull the server into that test and risk a cycle.
- The envtest harness goes in the subpackage `test/framework/epp/harness`, whose layer (R4) may import any package.

### Tasks

- [x] Create the `test/framework` root with layering rules in `doc.go`, wired into unit tests, coverage, and lint (#2121)
- [x] Unify the duplicated `GetFreePort` helpers into a single IPv4-safe implementation in `test/framework/net` (#2083, #2122)
- [ ] Move the fluent Pod builders from `pkg/epp/util/testing/wrappers.go` to `test/framework/k8s`
- [ ] Move the GAIE CRD builders (`MakeInferencePool`, `MakeInferenceObjective`) to `test/framework/gaie`; delete `pkg/epp/util/testing/wrappers.go`
- [ ] Move the ext_proc builders and stream executors from `test/integration/util.go` to `test/framework/epp`, after which nothing imports `test/integration`
- [ ] Export the EPP harness API, then move `test/integration/epp/harness.go` to `test/framework/epp/harness`

Each task is a single-commit PR containing only a mechanical move, opened after the previous one merges. The harness goes last because it is the heaviest and touches the hermetic suite.

### Out of scope

[...]

PR #2121, opened 2026-07-21
test: wire test/framework packages into unit tests and coverage

chethanuk:
**What type of PR is this?**

/kind cleanup
/kind test

First of eight in the #1188 split. It creates the `test/framework` package root and
wires it into the build before any helper moves there.

The wiring has to come first. `epp_TEST_PACKAGES` filters with `grep -v /test/` and
`sidecar_TEST_PACKAGES` is `./pkg/sidecar/...`, so nothing under `test/` runs in
`make test-unit` today. A package added there in a later PR would compile, pass review
and never execute.

The `Makefile` gets a `framework_TEST_PACKAGES = ./test/framework/...` entry and a
`test-unit-framework` target on `test-unit`, and `test-coverage` depends on `test-unit`
rather than repeating the suite list.
`framework_test.go` is one empty test so the package reports `ok` rather than
`? [no test files]`; without it the wiring cannot be told apart from a no-op.
`.golangci.yml` gets `importas` aliases for the five planned packages up front.
`no-unaliased` is false and `no-extra-aliases` is unset, so entries for packages that
do not exist yet are inert.

`test/framework/doc.go` states the layering rules: the subpackages are ordered by what
they are allowed to import, from the generic leaves at the bottom that take only the
standard library and Kubernetes API packages up to the harness at the top, which may
import anything.

**Test plan**:

- [x] `go build` and `go vet ./test/...` clean at this commit
- [x] `go test ./test/framework/...` reports `ok`, not `[no test files]`

Part of #1188

```release-note
NONE
```


.golangci.yml
```diff
@@ -81,6 +81,16 @@ linters:
           alias: errcommon
         - pkg: github.com/llm-d/llm-d-router/test/utils/igw
           alias: igwtestutils
```
shmuelk, line 83: Not connected to this PR, but the path `test/utils/igw` no longer exists

shmuelk (approved):

elevran:
Lint failing. Please also rebase 🙏

chethanuk:
@elevran or @shmuelk please merge if its good?

liu-cong (approved):

liu-cong:
approving given @elevran's comments are addressed and @shmuelk 's approval

PR #2122, opened 2026-07-21
test: move GetFreePort into test/framework/net

chethanuk:
**What type of PR is this?**

/kind cleanup
/kind test

Second of eight in the #1188 split. It follows the Makefile wiring PR that creates
`test/framework` and puts it under `make test-unit`.

This branch is based on #2121, so the diff currently includes that PR's commit as
well. Once #2121 merges this becomes a single commit touching only the
`test/framework/net` files.

`GetFreePort` is a network helper. It sat in `test/utils`, which is how a package
becomes a grab bag: nothing in the name says what belongs there, so everything does.
`test/framework/net` is where a port helper belongs.

The move is verbatim. The function body is unchanged; the deltas are the package
clause, the llm-d license header and a package comment, none of which
`test/utils/network.go` carried. Callers switch to the `fwknet` alias that
`.golangci.yml` already declares.

`test/framework/net` imports nothing from `pkg/`, so it stays an R1 leaf under the
layering rules in `test/framework/doc.go`. The chain moves one leaf package per PR so
each is revertable on its own, with the harness last because it is the heaviest.

**Test plan**:

- [x] `go build` and `go vet ./test/...` clean at this commit
- [x] `go test ./test/framework/...` passes, including the moved test

Part of #1188

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -81,6 +81,16 @@ linters:
           alias: errcommon
         - pkg: github.com/llm-d/llm-d-router/test/utils/igw
           alias: igwtestutils
+        - pkg: github.com/llm-d/llm-d-router/test/framework/net
+          alias: fwknet
+        - pkg: github.com/llm-d/llm-d-router/test/framework/context
+          alias: fwkcontext
+        - pkg: github.com/llm-d/llm-d-router/test/framework/k8s
+          alias: fwkk8s
+        - pkg: github.com/llm-d/llm-d-router/test/framework/gaie
+          alias: fwkgaie
+        - pkg: github.com/llm-d/llm-d-router/test/framework/epp
+          alias: fwkepp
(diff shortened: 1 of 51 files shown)


.golangci.yml
```diff
@@ -81,6 +81,16 @@ linters:
           alias: errcommon
         - pkg: github.com/llm-d/llm-d-router/test/utils/igw
```
shmuelk, line 82: Not related to this PR, but the path `test/utils/igw` no longer exists

chethanuk: I didn't touch it since it wasn't part of this PR. I'll probably handle that separately in another PR?

shmuelk:
@chethanuk Please rebase and resolve the conflicts

chethanuk:
@shmuelk Please review

ahg-g (approved):

Issue #1090, opened 2026-05-13
[Flow Control] Allow setting different default maxBytes configs for priorities >=0 and for < 0


ahg-g:
xref: https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2452

What would you like to be added:

Allow setting two different default priority configs, one for priorities above or equal to zero and one for priorities strictly below zero.

Why is this needed:
This allows setting lower maxBytes limits on queues with negative priorities; setting to zero can designate them as sheddable and offer better protection for higher priority requests.

LukeAVanDrie:
PR #1099 implemented the request eight days after the issue was filed, but its body linked only the upstream xref (kubernetes-sigs/gateway-api-inference-extension#2452), so #1090 never auto-closed.

flowControl.defaultNegativePriorityBand` provides a separate default band template for priorities strictly below zero and falls back to `defaultPriorityBand` when unset. Setting a small `maxRequests` or `maxBytes` on it makes negative-priority traffic sheddable.

Closing as completed.

Issue #2104, opened 2026-07-21
[Flow Control] [Graduation Blocker] Enable the flowControl feature gate by default


LukeAVanDrie:
**What would you like to be added**:

- `loader.RegisterFeatureGate(flowcontrol.FeatureGate, false)` -> `true` (`cmd/epp/runner/runner.go:674`); update the gate-off log message (`runner.go:880-885`) -- no longer "Experimental".
- WARN when a `flowControl:` config section is present while the gate is explicitly disabled (silently ignored today, `configloader.go:188-195`).
- Resolve `ENABLE_EXPERIMENTAL_FLOW_CONTROL_LAYER`: the env var is currently broken on main (#2065, fix in flight as PR #2076 -- phase-one/phase-two ordering means it never reaches `initAdmissionControl`). With the gate default-on it becomes a no-op; decide with #2065's author whether to land #2076 first or fast-track removal, and add the deprecation note either way.
- Docs: enable-by-default announcement, opt-out via `featureGates: ["flowControl=false"]`, the operational dependency on metrics scrape health, and tuning pointers (per-band `maxRequests`/`maxBytes`, `defaultRequestTTL`, priority-holdback policy).
- Flip the parameterized loader test added for the gate-wiring test issue.

Draft release note: Flow control is now enabled by default. Requests queue under saturation and are admitted by priority instead of being rejected; sheddable (negative-priority) requests now buffer until TTL or capacity limits instead of shedding immediately on saturation -- lower per-band `maxRequests`/`maxBytes` or configure the priority-holdback usage-limit policy to shed earlier. Opt out with `featureGates: ["flowControl=false"]`.

**Why is this needed**:

The goal of #1187 (and #1270): flow control becomes the default admission path in the end-of-July llm-d-router release. Depends on the config-defaults, status-mapping, recover, and saturation-observability issues landing first, plus e2e + benchmark numbers from a gate-flipped branch (to be attached here). Lands ~Jul 28 to soak in nightly CI before the cut; go/no-go criteria live on #1187. One-line revert.

LukeAVanDrie:
Closing: the default flip was decided against. Flow control is GA but explicitly enabled, since saturation detection needs per-deployment tuning; the gate keeps its false default. The decision and docs landed in #2180.

PR #2123, opened 2026-07-21
fix(flowcontrol): apply effective defaults for DefaultRequestTTL and per-band MaxRequests

### LINKED ISSUE #2095: [Flow Control] [Graduation Blocker] DefaultRequestTTL and per-band MaxRequests have no effective defaults, allowing unbounded queueing
**What happened**:

Two flow control config defaults do not exist despite being documented or expected:

1. `DefaultRequestTTL` accepts 0 with no default applied. `validate()` rejects only negative values (`pkg/epp/flowcontrol/controller/config.go:115-118`) and `NewConfig` never sets the field. The admission adapter always passes 0 (`pkg/epp/requestcontrol/admission.go:197`), so the controller config is the only TTL source; with TTL 0, `createRequestContext` builds a plain cancel context (`controller/controller.go:405-422`) and queued requests die only on client disconnect.
2. The `PriorityBandConfig.MaxRequests` doc comment says "Defaults to `defaultPriorityBandMaxRequests` (5000)" (`pkg/epp/flowcontrol/registry/config.go:149-152`), but that constant does not exist and `applyDefaults` (`config.go:397-417`) never touches `MaxRequests`. The effective per-band request-count limit is unlimited. Only bytes are bounded (1 GB/band); global limits default to unlimited (`config.go:312-313`).

Together: under sustained saturation, requests accumulate silently with no TTL and no count bound until the 1 GB/band byte cap.

A sane default TTL and the documented per-band count default to be applied when the config omits them, so that a default configuration cannot queue unboundedly.

Enable the `flowControl` feature gate with a minimal config (no `defaultRequestTTL`, no band limits). Saturate the pool so dispatch is gated. Observe queue length growing without TTL evictions; confirm from code that no count limit or TTL applies.

Part of #1187; blocks enabling the gate by default.

Proposed fix:
- Apply a `defaultRequestTTL` when the API value is unset (proposal: 60s -- discuss here). Decide the semantics of an explicit 0 (reject in `validate()`, or document as "no TTL, disconnect-only") -- anything but the current silent behavior.
- Add `defaultPriorityBandMaxRequests = 5000` applied in `applyDefaults` when `MaxRequests == 0`, mirroring the existing `MaxBytes` handling (including any explicit "unlimited" opt-out).
- Fix the adjacent stale comment: `FlowGCTimeout` doc claims 1 hour, constant is 5 minutes (`registry/config.go:98` vs `:50`).

Related: #1090 proposes split defaults for priorities >= 0 vs < 0 (the fuller design; the `DefaultNegativePriorityBand` template at `registry/config.go:90-94` is a partial step). This issue is the minimal safe default and does not preclude it.

Acceptance:
- Config unit tests assert both defaults from an empty API config, plus the explicit-zero TTL behavior.
- Release note documents both defaults (behavior change for current opt-in users).

**Environment**:
- Install tools: n/a
- Others: n/a

LukeAVanDrie:
**What type of PR is this?**

/kind bug

`DefaultRequestTTL` and per-band `MaxRequests` had no effective defaults: `validate()` only rejects negative TTLs, and the band `MaxRequests` doc comment references a `defaultPriorityBandMaxRequests` constant that does not exist. A default configuration could queue requests with no time or count bound until the 1 GB per-band byte limit.

Changes:

- `defaultRequestTTL` defaults to 60s when unset. The API field is a `*metav1.Duration`, so unset and explicit zero are distinguishable: an explicit `"0s"` keeps disconnect-only behavior and is documented as such. The 60s value is a queue-wait budget. The well-lit guides configure no gateway request timeout, so this is the only bound on queue wait in those deployments. Where a shorter gateway or client deadline exists, context cancellation evicts the request first. This repo's chart defaults the route timeout to 300s, so with default config the TTL fires first; a request queued for 60s has already lost most of its time-to-first-token budget, and operators who prefer longer waits can raise `defaultRequestTTL`.
- Per-band `maxRequests` defaults to 5000 when unset, applied in `applyDefaults` alongside the existing `MaxBytes` handling. Band-level 0 is treated as unset, matching how `maxBytes: 0` is handled today. The two defaults are consistent: under a dispatch halt, occupancy self-limits at arrival rate times TTL, so the count cap only binds above roughly 83 req/s per band.
- Global limits stay unlimited by default. Band count is bounded by the distinct priorities defined in InferenceObjectives, so per-band caps bound total memory exposure.
- Doc fixes: the `FlowGCTimeout` default (5 minutes; the comment claimed 1 hour), the `DefaultNegativePriorityBand` comment claiming `maxBytes: "0"` drops traffic immediately (band-level 0 has never meant zero capacity), explicit zero-semantics notes on all four limit fields, and the `InitialEffectiveTTL` adapter docstring with a TODO for per-request and per-band TTL scopes (#1090).

Tests: defaults asserted at the controller, registry (including the auto-provisioned priority-0 band and both dynamic templates), and loader layers, with nil vs explicit-zero round-trips for both fields. The flowcontrol tree passes under `-race`.

Design discussion and the recorded 60s rationale are in the plan comment on #2095.

Fixes #2095

Part of #1187.

**Release note**:

```release-note
Flow control queueing is now bounded by default: `defaultRequestTTL` defaults to 60s and per-band `maxRequests` defaults to 5000 when unset. For existing flow control users this is a behavior change: an explicit `defaultRequestTTL: "0s"` disables the TTL (requests then wait until client disconnect), and a band `maxRequests`/`maxBytes` of 0 is treated as unset and receives the default. To approximate the legacy fast-fail behavior for sheddable traffic, configure a small `defaultNegativePriorityBand.maxRequests`. Scale-to-zero deployments should raise `defaultRequestTTL` above the expected cold-start time.
```


LukeAVanDrie:
There is one consequence of the 60s TLL default worth deciding. The queue also serves as the scale-from-zero waiting room (with an empty pool, requests queue instead of failing). The default TTL cuts that room to 60s. Cold starts may take O(minutes), and the chart's route timeout would have carried them, so a scale-from-zero deployment on pure defaults now sheds everything during cold start (after 60s).

Options:

1. Ship 60s and document it (this PR). The shed is visible (503 with a `TTLExpired` reason header) and the fix is one config line; sizing guidance goes in #2104.
2. Default to ~300s to match the chart. Under sustained saturation a FIFO queue with a TTL converges to a standing queue at the TTL (every served request also waits that long), so 300s means holding five minutes of doomed work, and the TTL stops being a useful queue-wait budget.
3. Pause or extend the TTL while the pool is empty, so it measures "the pool is up and you still couldn't dispatch." 

[3] is probably the right end state, but it's new mechanism.

This is the biggest user-facing decision for enablement by default, so we want to make sure we get it right.

LukeAVanDrie:
cc: @ahg-g, @RishabhSaini

pkg/epp/flowcontrol/controller/config.go
```diff
@@ -24,6 +24,13 @@ import (
 )
 
 const (
+	// defaultRequestTTL is the default Time-To-Live applied to queued requests when the configuration
+	// does not specify one. It is a queue-wait budget: a request still undispatched after this long is
+	// shed with a retryable backpressure signal instead of being served with severely degraded
+	// time-to-first-token. It is also the only bound on queue wait when neither the client nor the
+	// gateway enforces a request deadline (the well-lit guides configure no gateway request timeout);
+	// where such deadlines exist and fire sooner, context cancellation evicts the request first.
+	defaultRequestTTL = 60 * time.Second
```
ahg-g, line 33: Requests can come with a timeout in their context, right? if so, what happens if this is lower than the timeout set by the client?

LukeAVanDrie: The two deadlines compose with minimum-wins semantics.

We derive the queue context via `context.WithDeadlineCause(ctx, enqueueTime.Add(TTL), ErrTTLExpired)`, where ctx is the inbound request context.

- On client deadline sooner than the TTL, the parent's deadline wins and fires first. Crucially, the ErrTTLExpired cause only attaches if the TTL's own deadline is the one that expires; when the parent fires, the cause is the parent's cancellation. So it flows down the QueueOutcomeEvictedContextCancelled path and is reported as client disconnect/timeout, not as a TTL shed. Attribution stays honest in metrics and status mapping.
- On client deadline later (or absent), the 60s TTL fires first with the ErrTTLExpired cause → EvictedTTL → the TTLExpired dropped-reason response. The client's larger budget is deliberately cut at 60s of queue wait. The remaining budget is preserved for generation. I.e., TTL is "queue-wait budget" not "client-deadline proxy".

In short, if the client sets a higher timeout, we still enforce maximum of 60s queue-wait budget by default.

LukeAVanDrie: Long-term I would like to be able to express TTL at the per-priority (InfObj) and request level, but that is not in scope for this change as it would be a new transport mechanism.

ahg-g (approved):

PR #2125, opened 2026-07-21
feat(epp): add encoder cache features to the latency predictor

kaushikmitr:
What type of PR is this?

/kind feature

What this PR does / why we need it:

Adds two multimodal encoder-cache features to the latency predictor: encoder_input_size (total size of the request's unique multimodal items) and encoder_matched_size (portion likely present in the target pod's encoder cache, from the multimodal producer's per-pod match data).

For multimodal requests, encoder compute dominates TTFT when the target pod's encoder cache misses, and this cost differs per pod. The predictor cannot see this today, so TTFT predictions carry large unexplained variance on multimodal workloads, degrading SLO-aware routing and admission. input - matched approximates the encoder work a pod would actually perform.

The features are opt-in via a new useEncoderCacheFeatures plugin parameter (default false; no behavior change for existing deployments). When enabled, the plugin consumes EncoderCacheMatchInfo as a required dependency so the DAG orders the multimodal producer first, auto-creating it if absent. encoderCacheMatchInfoProducerName selects a specific producer instance. Sizes flow into bulk predictions per endpoint and into TTFT training entries for the serving pod (prefill pod in disaggregated mode); TPOT paths send zeros, following the prefix_cache_score convention. The Bayesian Ridge path applies matching TTFT coefficients when present; models trained without them are unaffected.

Note: the Python predictor server needs matching schema support (encoder_input_size, encoder_matched_size, both non-negative ints, matched <= input) before tree models learn from these fields; until then they are sent and ignored.

Which issue(s) this PR fixes:

Fixes #

Release note:

The latency predictor supports optional multimodal encoder-cache features (encoder_input_size, encoder_matched_size), enabled with the useEncoderCacheFeatures parameter, to improve TTFT predictions for multimodal requests.

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/README.md
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/README.md
@@ -16,6 +16,8 @@ DataProducer, PreRequest, ResponseHeader, ResponseBody, ProducerPlugin, Consumer
 - TPOT training data collection at EOS (streaming mode)
 - Per-endpoint running request queue tracking (TPOT SLO priority queue)
 - Prefix cache score forwarding from `PrefixCacheMatchInfo` attributes
+- Multimodal encoder-cache size forwarding from `EncoderCacheMatchInfo` attributes
+  (opt-in via `useEncoderCacheFeatures`)
 - TPOT neutralization for prefill endpoints in disaggregated serving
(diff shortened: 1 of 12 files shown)


ahg-g:
Did we do any evaluation?

kaushikmitr:
> Did we do any evaluation?

yes, https://github.com/llm-d/llm-d/pull/2078/changes

ahg-g (approved):

PR #2131, opened 2026-07-22
test(flowcontrol): deflake integration tests that raced on fixed sleeps

### LINKED ISSUE #2117: [Flake] `TestFlowControlMetricsEmitted` asserts queue_size > 0 is timing dependent
/kind flake

`TestFlowControlMetricsEmitted` in `pkg/epp/flowcontrol/integration_test.go` intermittently fails with:

    integration_test.go:1190:
        Error:      "0" is not greater than "0"
        Test:       TestFlowControlMetricsEmitted
        Messages:   queue_size should be > 0 while a request is actively queued

The test enqueues a blocked request, sleeps 50ms, then gathers the `llm_d_epp_flow_control_queue_size` gauge and asserts it is > 0. Under CI load, 50ms is not always enough for the gauge update to be observed, so the assertion can race.

Observed in: https://github.com/llm-d/llm-d-router/actions/runs/29835941941/job/88652219957 (PR #1678, unrelated change - only touched `cmd/epp/runner/runner.go` import ordering).

Suggest replacing the fixed sleep with a poll/require.Eventually on the gauge value.

LukeAVanDrie:
**What type of PR is this?**

/kind test

`TestFlowControlMetricsEmitted` slept a fixed 50ms and then asserted the `queue_size` gauge was > 0, racing against the scheduling of the `EnqueueAndWait` goroutine (and against the request's 200ms TTL) under CI load.

The test now:
- waits for queue admission via `require.Eventually` on registry stats before reading the gauge (the gauge increments before admission, so it is deterministically > 0 at that point),
- finalizes the request deterministically by unblocking the detector and asserting dispatch, then asserts the gauge returns to exactly 0 (new coverage as the old test never asserted the decrement),
- filters the gauge by the test's `fairness_id` label and runs in the sequential test phase, so its exact-value assertions are deterministic even though the metrics registry is process-global (sequential execution also guarantees the label is admitted under the fairness-ID cardinality cap before the parallel stress tests can exhaust it).

A scan of the file for the same flake class turned up the three capacity tests (`TestGlobalAndBandCapacityInteraction`, `TestByteCapacityEnforcement`, `TestEmptyPoolRejectsAsNoEndpoints`), which staggered all enqueues on 5ms sleeps with 500ms TTLs.  A CI stall could let capacity-filling requests expire before the overflow requests arrived. They now share a `fillQueue` harness helper that fills capacity, waits until the fill is observable in registry stats, and only then lets the test send overflow traffic; this also tightened the assertions from `>=`/`<=` margins to exact outcomes.

Verified with `-race` across `GOMAXPROCS` 1/2/4 (105 runs), 50 runs under 12-way CPU oversubscription (via busy loops) to approximate loaded CI, and shuffled full-package runs — no failures.

  Fixes #2117

  ```release-note
  NONE
  ```


LukeAVanDrie:
cc: @elevran

shmuelk:
@LukeAVanDrie Please rebase this PR, it will fix the CI failure

pkg/epp/flowcontrol/integration_test.go
```diff
@@ -497,37 +497,27 @@ func TestGlobalAndBandCapacityInteraction(t *testing.T) {
...
-			reqCtx, reqCancel := context.WithTimeout(h.ctx, 500*time.Millisecond)
-			defer reqCancel()
-			req := &testRequest{id: id, key: key, byteSize: 100, ttl: 500 * time.Millisecond}
-			outcome, err := h.fc.EnqueueAndWait(reqCtx, req)
-			results <- dispatchResult{id: id, outcome: outcome, err: err}
-		}()
-		time.Sleep(5 * time.Millisecond)
-	}
+	release := h.fillQueue(t, key, 3, 100, func() bool {
+		return h.reg.Stats().TotalLen == 3
```
shmuelk: Better would be:


```suggestion
    const count = 3
	release := h.fillQueue(t, key, count, 100, func() bool {
		return h.reg.Stats().TotalLen == count
```

liu-cong (approved):

PR #2132, opened 2026-07-22
fix(flowcontrol): add bounded request drop summaries

### LINKED ISSUE #2101: [Flow Control] [Graduation Blocker] Log quality pass; request drops visible at default verbosity

A quality pass over `pkg/epp/flowcontrol/` log messages and levels:
- Bounded drop visibility at default verbosity: keep the per-request drop log (`controller/controller.go:276-279`) at `V(VERBOSE)`, and add a time-bounded aggregate at default -- e.g. a periodic summary from the processor ("N requests dropped in the last interval, by outcome") and/or first-occurrence state-transition logs ("band X began/stopped rejecting on capacity"), both bounded by time or config cardinality rather than request rate. Rationale: at the shipped default verbosity (-v=2) controller-runtime's zap sampler is DISABLED (crlog adds its 100/sec sampler only when the level does not enable V(2)), so a per-request Info drop log would emit one unsampled JSON line per drop at exactly the moment the system is saturated. Rates are already covered by `flow_control_requests_total{outcome}`; the aggregate log covers "which flow/band and why" without the volume; per-request forensics remain at -v=3.
- Unify the mixed `err`/`error` log keys (`controller.go:278`, `processor.go:265,574` vs `processor.go:449`).
- Add request IDs (or ID lists at DEBUG) to the sweep/drain logs, which today carry only counts (`processor.go:489-490`, `registry/managedqueue.go:153,166`).
- Add used-vs-limit stats to the pool-empty capacity rejection branch (`processor.go:292-293`).
- General message wording and level fit against the documented conventions (DEFAULT once-per-request, DEBUG per-item, TRACE state transitions).

**Why is this needed**:

Request drops are currently invisible at default verbosity -- the only per-request outcome log sits above the default level, so operators of a default-on deployment cannot see why requests were rejected or evicted without raising verbosity. Part of #1187; the drop-log item blocks enabling the gate by default.

Coordination: overlaps the silent-expiration logging PR claimed in #1187's comments -- align before starting. Note the original #1187 text's `requestID` -> `reqID` rename is stale: the codebase standard is `requestID`.

ishwar170695:
**What type of PR is this?**
/kind cleanup

This PR addresses the logging improvements tracked in #2101 by:

- adding bounded periodic summaries of non-dispatched request outcomes at `V(DEFAULT)` during processor cleanup sweeps while preserving per-request drop logs at `V(VERBOSE)`
- adding `requestIDs` to `ManagedQueue` cleanup and drain debug logs
- unifying structured error keys to `err`
- extending pool-empty rejection logs with queue capacity context
- adding unit tests covering drop accounting and periodic summary behavior

Fixes #2101

**Release note**:
```release-note
Added periodic aggregate summary logging for flow control request drops.

### CODE REVIEW

File: pkg/epp/flowcontrol/controller/internal/processor.go [hunk 1]
```diff
@@ -510,6 +528,7 @@ func (p *Processor) shutdown() {
 				// Finalize buffered items.
 				item.FinalizeWithOutcome(types.QueueOutcomeRejectedOther,
 					fmt.Errorf("%w: %w", types.ErrRejected, types.ErrFlowControllerNotRunning))
+				p.recordDrop(types.QueueOutcomeRejectedOther)
```
LukeAVanDrie, line 532: Everything recorded here and in `evictAll()` is never flushed. The only `flushDropSummary()` call site is the sweep ticker, and `runCleanupSweep` exits on the same `ctx.Done()` that drives us into `shutdown()`. The shutdown evictions, plus anything accumulated since the last tick, are counted and then silently discarded.

We can add a final flush at the end of `shutdownOnce.Do`, after `p.evictAll()` returns (as the worker pool's `wg.Wait()` inside `processAllQueuesConcurrently` guarantees all `recordDrop` calls have landed by then):

```go
p.evictAll()
p.flushDropSummary()
```

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -539,11 +558,36 @@ func (p *Processor) evictAll() {
 			// Finalization is idempotent; safe to call even if already finalized externally.
 			// The per-request log is emitted by EnqueueAndWait when it unblocks.
 			item.FinalizeWithOutcome(outcome, errShutdown)
+			p.recordDrop(outcome)
```
LukeAVanDrie: `FinalizeWithOutcome` is a `sync.Once` no-op when the item was already finalized externally, but this line unconditionally records the hardcoded `EvictedOther`. An item that actually finished as `EvictedTTL` or
`EvictedContextCancelled` (finalized by the controller but not yet swept when shutdown drains the queue) gets counted under the wrong bucket.

We can record the authoritative outcome instead (guaranteed non-nil after the finalize call):

```suggestion
			p.recordDrop(item.FinalState().Outcome)
```

pkg/epp/flowcontrol/controller/internal/processor.go
LukeAVanDrie, line 272: Every terminal branch in `enqueue()` now records its drop except this early return for items finalized externally while buffered in `enqueueChan`. Those items never enter a queue, so the sweep can't count them either; they're absent from the summary entirely.

Maybe unlikely to trigger in practice, but we can harden against this by adding `p.recordDrop(finalState.Outcome)` to that branch.

You already guard against `Dispatched`/`NotYetFinalized` in `recordDrop` so this should be safe.

LukeAVanDrie (changes requested):
Thanks! This is almost ready, I just flagged some small edge cases in outcome accounting that should all be small fixes.

Non-blocking, but there is also no test for the sweep path (`sweepFinalizedItems` -> `recordDrop`), which is the main channel for TTL/cancellation drops.

We could also consider labeling by priority in the log summary (since this is cardinality bound). Per-flow is probably not worth it. Up to your discretion as I think it is already okay in its current state.

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -473,6 +485,7 @@ func (p *Processor) runCleanupSweep(ctx context.Context) {
 			return
 		case <-ticker.C():
 			p.sweepFinalizedItems()
+			p.flushDropSummary()
```
LukeAVanDrie, line 489: note: this binds log frequency to the expiry cleanup interval (1s); that coupling seems reasonable to me and probably doesn't warrant an independent ticker

pkg/epp/flowcontrol/controller/internal/processor.go
ishwar170695: Done. Added a final `flushDropSummary()` after `evictAll()` in `shutdown()` so any outcomes accumulated since the last sweep (including shutdown evictions) are emitted before exit.

pkg/epp/flowcontrol/controller/internal/processor.go
ishwar170695: Updated this to record `item.FinalState().Outcome` so externally finalized items retain their authoritative outcome instead of always being counted as `EvictedOther`.

pkg/epp/flowcontrol/controller/internal/processor.go
ishwar170695: Added `p.recordDrop(finalState.Outcome)` to this early-return path so externally finalized items contribute to the aggregate summary as well.

ishwar170695:
> Non-blocking, but there is also no test for the sweep path (`sweepFinalizedItems` -> `recordDrop`), which is the main channel for TTL/cancellation drops.

Added a test covering the sweep path by verifying that an externally finalized item increments the corresponding drop counter when swept.

LukeAVanDrie (dismissed):
lgtm

LukeAVanDrie:
Once the linter issue is resolved, I can merge this

ishwar170695:
> Once the linter issue is resolved, I can merge this

Fixed.

LukeAVanDrie (dismissed):

LukeAVanDrie:
> > Once the linter issue is resolved, I can merge this
> 
> Fixed.

Still tripping; can you run go fmt? That seems to be the issue

ishwar170695:
> > > Once the linter issue is resolved, I can merge this
> > 
> > 
> > Fixed.
> 
> Still tripping; can you run go fmt? That seems to be the issue

Ran `gofmt` and pushed the fix.

LukeAVanDrie (approved):

PR #2142, opened 2026-07-22
feat: add new --model-server-port to replace --vllm-port

zdtsw:
**What type of PR is this?**
/kind feature

- disagg-sidecar expose vllm specific flag name which should be model server neutral. add --model-server-port to replace --vllm-port which stays as a deprecated alias until final removal.
- update usage text on --data-parallel-size

ref #https://github.com/llm-d/llm-d-router/issues/2039

```release-note
The pd-sidecar `--vllm-port` flag is deprecated in favor of the vendor-neutral `--model-server-port` and will be removed in a future release. `--vllm-port` continues to work as an alias meanwhile. The same rename applies to the sidecar YAML configuration key: `vllm-port` -> `model-server-port`.
```

cc @rahulgurnani

### DIFF
--- a/pkg/sidecar/proxy/options.go
+++ b/pkg/sidecar/proxy/options.go
@@ -132,7 +132,7 @@ type Options struct {
 	// Fields with direct CLI flags are bound here via embedding; derived fields are set in Complete().
 	Config
 
-	// modelServerPort is the port the model server(vLLM, SGLang etc) is listening on; used to compute Config.DecoderURL in Complete().
+	// modelServerPort is the port the model server (vLLM, SGLang, etc.) is listening on; used to compute Config.DecoderURL in Complete().
 	modelServerPort string
 	// vllmPort is the deprecated alias for modelServerPort; migrated in Complete().
 	vllmPort string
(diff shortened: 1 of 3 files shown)


pkg/sidecar/proxy/options.go
```diff
@@ -243,8 +247,10 @@ func (opts *Options) AddFlags(fs *pflag.FlagSet) {
 	// Add Go flags to pflag (for zap options compatibility)
 	fs.AddGoFlagSet(goFlagSet)
 	fs.StringVar(&opts.Port, port, opts.Port, "the port the sidecar is listening on")
-	fs.StringVar(&opts.vllmPort, vllmPort, opts.vllmPort, "the port vLLM is listening on")
-	fs.IntVar(&opts.DataParallelSize, dataParallelSize, opts.DataParallelSize, "the vLLM DATA-PARALLEL-SIZE value")
+	fs.StringVar(&opts.modelServerPort, modelServerPort, opts.modelServerPort, "the port the model server is listening on")
+	fs.StringVar(&opts.vllmPort, vllmPort, opts.vllmPort, "the port the model server is listening on")
+	_ = fs.MarkDeprecated(vllmPort, "use --model-server-port instead; --vllm-port will be removed after the deprecation period")
```
elevran, line 253: please grep for use of the deprecated CLI string in the code base and replace with the new name.
For example `test/sidecar/config/base/pod-qwen.yaml:11` still passes `--vllm-port=8001` (that's fine since it's the alias and still works but will break once we deprecate). Also, CLI tests in `pkg/sidecar/proxy/options_test.go` still use the now deprecated value.

pkg/sidecar/proxy/options.go
```diff
@@ -333,6 +339,13 @@ func (opts *Options) Complete() error {
 		return err
 	}
 
+	// Migrate deprecated --vllm-port to --model-server-port.
```
elevran: Please create a follow up issue assigned to the v0.12 miletone to fully remove the deprecated value

zdtsw: created https://github.com/llm-d/llm-d-router/issues/2172 to follow removal

pkg/sidecar/proxy/options.go
zdtsw: updated both test and config example

pkg/sidecar/proxy/options_test.go
```diff
@@ -1130,6 +1130,44 @@ func TestCompleteMoRIIOWriteModeGuards(t *testing.T) {
...
+func TestModelServerPortYAML(t *testing.T) {
+	opts, testPFlagSet := newTestOptions(t)
+	yaml := "{model-server-port: 8203}"
+	setFlag(t, testPFlagSet, inlineConfiguration, &yaml)
+	require.NoError(t, testPFlagSet.Parse(nil))
+
+	require.NoError(t, opts.Complete())
+	require.NoError(t, opts.Validate())
+	require.Equal(t, "http://localhost:8203", opts.DecoderURL.String())
+}
```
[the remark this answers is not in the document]
zdtsw: updated

shmuelk (approved):

elevran (approved):

PR #2147, opened 2026-07-22
epp: generalize endpoint identity

hexfusion:
**What type of PR is this?**
/kind cleanup

Rename `EndpointMetadata.PodName` to `Name`. The endpoint model is not
Kubernetes-only: file-discovered endpoints have no pod, and `NamespacedName` is the
identity. `Name` is always set, so an endpoint has a name regardless of discovery
source.

followup from https://github.com/llm-d/llm-d-router/pull/1927

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/interface/datalayer/endpoint_metadata.go
+++ b/pkg/epp/framework/interface/datalayer/endpoint_metadata.go
@@ -23,10 +23,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-// EndpointMetadata represents the relevant Kubernetes Pod state of an inference server.
+// EndpointMetadata describes an inference endpoint.
 type EndpointMetadata struct {
 	NamespacedName types.NamespacedName
-	PodName        string
+	Name           string
 	Address        string
 	// NodeAddress is the node IP hosting this pod (pod.Status.HostIP).
 	// Empty for non-Kubernetes discovery sources (e.g. file discovery).
@@ -60,7 +60,7 @@ func (epm *EndpointMetadata) Clone() *EndpointMetadata {
 			Name:      epm.NamespacedName.Name,
 			Namespace: epm.NamespacedName.Namespace,
 		},
-		PodName:     epm.PodName,
+		Name:        epm.Name,
 		Address:     epm.Address,
 		NodeAddress: epm.NodeAddress,
 		Port:        epm.Port,
@@ -77,7 +77,7 @@ func (epm *EndpointMetadata) Equal(other *EndpointMetadata) bool {
 		return epm == other
 	}
 	return epm.NamespacedName == other.NamespacedName &&
-		epm.PodName == other.PodName &&
+		epm.Name == other.Name &&
 		epm.Address == other.Address &&
 		epm.NodeAddress == other.NodeAddress &&
 		epm.Port == other.Port &&
--- a/pkg/epp/framework/interface/datalayer/endpoint_metadata.go
+++ b/pkg/epp/framework/interface/datalayer/endpoint_metadata.go
@@ -23,11 +23,17 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+// ID uniquely identifies an endpoint. It aliases types.NamespacedName so the identity
+// can later move off the Kubernetes-specific type without churning datastore keys.
+// Every discovery source must set it uniquely.
+type ID = types.NamespacedName
+
 // EndpointMetadata describes an inference endpoint.
 type EndpointMetadata struct {
-	NamespacedName types.NamespacedName
-	Name           string
-	Address        string
+	// ID is the endpoint's unique identity, and the datastore key.
+	ID      ID
+	Name    string
+	Address string
 	// NodeAddress is the node IP hosting this pod (pod.Status.HostIP).
 	// Empty for non-Kubernetes discovery sources (e.g. file discovery).
 	NodeAddress string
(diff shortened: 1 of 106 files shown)


elevran:
For the k8s case, there are existing labels for engine type (vllm, sglang, custom) that you may want to use and always assign an endpoint type.

However, this assumes we'd want to make endpoint's type a "first class concept". 

An alternative route is to introduce a type attribte solely for the multicluster use case for now, alogn with multicluster specific plugins in the different EPP components (e.g., request control, data layer, scheduling filters and scorers, etc.). Those would operate only on cluster typed endpoints and serve as no-op in other cases (in fact they should not be configured at all if not running in multicluster mode...). 
- If the plugin does not care for the endpoint type, no change is needed and it can be reused.
- if it operates differently on engine and cluster type endpoints, then trying to overload existing plugins for that can lead to missed type checks and the wrong actions. I think initially it may be cleaner to not introduce that differntiation into existing plugins.

pkg/epp/framework/interface/datalayer/endpoint_metadata.go
```diff
@@ -17,23 +17,48 @@ limitations under the License.
...
+func ParseEndpointType(s string) (EndpointType, error) {
+	switch t := EndpointType(s); t {
+	case "":
+		return EndpointTypeEngine, nil
+	case EndpointTypeEngine, EndpointTypeEPP:
+		return t, nil
+	default:
+		return "", fmt.Errorf("%w %q", ErrUnknownEndpointType, s)
+	}
+}
```
elevran: nit: the zero value of EndpointType ("") is distinct from EndpointTypeEngine, so an EndpointMetadata built by direct struct literal without going through ParseEndpointType will have Type == "" rather than the documented default.
See note below on option of making this an opetioan attribute instead of a struct field.

pkg/epp/framework/interface/datalayer/endpoint_metadata.go
```diff
@@ -17,23 +17,48 @@ limitations under the License.
...
-	PodName        string
+	Name           string
 	Address        string
 	// NodeAddress is the node IP hosting this pod (pod.Status.HostIP).
 	// Empty for non-Kubernetes discovery sources (e.g. file discovery).
 	NodeAddress string
 	Port        string
 	MetricsHost string
 	Labels      map[string]string
+	Type        EndpointType
```
elevran: Have you considered perhaps introducing an attribute instead of extending the endpoint struct itself?
The attribute can be set as an optional field when needed/relevant, unlike a built-in value that has more change impact.

elevran (commented):
Clean rename plus a well-tested additive field, matches the discussion from #1927. 
See question and nit inlined. 
Please weigh pros and cons of a new struct field (with defaults when not set) vs use of an optional Endpoint Attribute.

hexfusion:
> Clean rename plus a well-tested additive field, matches the discussion from #1927. See question and nit inlined. Please weigh pros and cons of a new struct field (with defaults when not set) vs use of an optional Endpoint Attribute.

good points I simplified the scope of this PR and added the attribute handling to a followup https://github.com/llm-d/llm-d-router/pull/2191

elevran (commented):
suggestions (follow up PRs if accepted), it would be worthwhile to 
- clearly indicate what the name is used for (e.g., identity that requires uniqueness, just meaningful to users to correlate with the runtime environment, etc.). If this is an identity, consider hiding behind an `Identity/ID()` function to emphasize.
- consider whether namespace still makes sense (esp. when not k8s centric)

Approved from my point of view.
/cc @ahg-g @liu-cong @vMaroon @nilig for awareness

liu-cong:
> suggestions (follow up PRs if accepted), it would be worthwhile to
> 
> * clearly indicate what the name is used for (e.g., identity that requires uniqueness, just meaningful to users to correlate with the runtime environment, etc.). If this is an identity, consider hiding behind an `Identity/ID()` function to emphasize.
> * consider whether namespace still makes sense (esp. when not k8s centric)
> 
> Approved from my point of view. /cc @ahg-g @liu-cong @vMaroon @nilig for awareness

NamespacedName is unique, renaming it to ID() is reasonable. Given this PR tries to "generalize endpoint identity", can we do the NamespacedName rename as well?

pkg/epp/framework/interface/datalayer/endpoint_metadata.go
```diff
@@ -95,9 +101,9 @@ func (epm *EndpointMetadata) GetRankIndex() int {
 	return epm.RankIndex
 }
 
-// GetNamespacedName gets the namespace name of the Endpoint.
-func (epm *EndpointMetadata) GetNamespacedName() types.NamespacedName {
-	return epm.NamespacedName
+// GetID returns the endpoint's unique identity.
+func (epm *EndpointMetadata) GetID() ID {
```
hexfusion, line 105: dropped GetNamespacedName() for GetID(), clean like the PodName rename in this PR. sny downstream callers worth keeping? If so I'll leave a // Deprecated: use GetID shim. cc @elevran

elevran: I'm not aware of any downstream callers.

elevran (approved):
clean, well-scoped mechanical rename PR: `PodName` → `Name`, `NamespacedName` → `ID` (with `GetID()` accessor), plus a public `ID` type alias. This is a pure rename touching many files, with no behavior change.

PR #2149, opened 2026-07-22
docs: require monotone ceilings in UsageLimitPolicy contract

LukeAVanDrie:
**What type of PR is this?**

/kind documentation

Documents a conformance requirement on `UsageLimitPolicy` implementations: returned ceilings must be monotonically non-increasing in the given priority order. The flow controller's dispatch loop visits bands from highest to lowest priority and stops at the first gated band, so a lower band whose ceilin g exceeds a higher band's can report itself open on calls where the dispatch loop never reaches it, starving that band. The requirement previously existed only implicitly in the dispatch loop's behavior; this states it in the interface contract where implementers will see it.

N/A (doc-comment-only change). Related to #1956.

 ```release-note
 NONE
```

### DIFF
--- a/pkg/epp/framework/interface/flowcontrol/plugins.go
+++ b/pkg/epp/framework/interface/flowcontrol/plugins.go
@@ -148,9 +148,9 @@ type SaturationDetector interface {
 // considered on that call.
 //
 // Conformance: Implementations MUST ensure all methods are goroutine-safe. Returned ceilings MUST be
-// monotonically non-increasing in the given priority order: because the dispatch loop stops at the first
-// gated band, a lower band whose ceiling exceeds that of a higher band can be marked open on calls where
-// it is unreachable, starving it.
+// monotonically non-increasing in the given priority order (highest priority first): because the
+// dispatch loop stops at the first gated band, a lower band whose ceiling exceeds that of a higher band
+// can be marked open on calls where it is unreachable, starving it.
 type UsageLimitPolicy interface {
 	plugin.Plugin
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by LukeAVanDrie** [comment]
@loicmarchal can I get your opinion here?

cc: @kalantar

File: pkg/epp/framework/interface/flowcontrol/plugins.go [hunk 1]
```diff
@@ -143,9 +143,14 @@ type SaturationDetector interface {
...
 // priority band, the returned ceiling is compared against current saturation. If saturation exceeds the
-// ceiling for a given priority, requests at that priority are gated (not dispatched).
-//
-// Conformance: Implementations MUST ensure all methods are goroutine-safe.
+// ceiling for a given priority, requests at that priority are gated (not dispatched). The dispatch loop
+// visits bands from highest to lowest priority and stops at the first gated band; lower bands are not
+// considered on that call.
+//
+// Conformance: Implementations MUST ensure all methods are goroutine-safe. Returned ceilings MUST be
+// monotonically non-increasing in the given priority order: because the dispatch loop stops at the first
```
loicmarchal: nit: "in the given priority descending order"? Although it is mentioned below the priorities are ordered highest first

loicmarchal:
> @loicmarchal can I get your opinion here?
> 
Thanks @LukeAVanDrie, lgtm (just a very minor comment/question). It's a good addition of doc to cover the new priority holdback policy

pkg/epp/framework/interface/flowcontrol/plugins.go
LukeAVanDrie: updated the phrasing to make this clearer

ahg-g (approved):

PR #2151, opened 2026-07-23
datalayer: hot-reload scrape client certificate on rotation

### LINKED ISSUE #2045: [Feature] Hot-reload the metrics/models scrape client cert on rotation

The metrics/models scrape client loads its TLS client cert (and CA) once, at data-source construction, and never re-reads them. Add hot-reload so a rotated cert is picked up without restarting the EPP.

- wire `tls.Config.GetClientCertificate` to a controller-runtime `certwatcher` (`sigs.k8s.io/controller-runtime/pkg/certwatcher`)  already a dependency, and already how the metrics *server* hot-reloads started/stopped with the data-source lifecycle.
-  cycle the scrape connections or swap the transport so warm keep-alive connections re-handshake once and pick up the new cert per rotation vs per scrape.

The CA pool (`RootCAs`) can stay load-once for now the root/intermediate rotates rarely, so a restart on CA rotation is acceptable; hot-reloading the CA can be a second increment.

Follow-up to #1858 (client mTLS); complements #1919 (server mTLS).

**Why is this needed**:

When certs are auto-rotated, a client that loads once presents a stale cert at renewal and mTLS breaks until the EPP is restarted. The metrics server already hot-reloads via `certwatcher`; this makes the client/scrape side symmetric, and is the piece that makes provider-managed cert rotation actually land in the EPP without a pod bounce.

hexfusion:
**What type of PR is this?**
/kind feature

The scrape client loaded its TLS client cert once, at construction, so a rotated
cert broke mTLS until the EPP restarted. Rotation already worked in the serving
direction (ext-proc via `common.CertReloader`, metrics server via the `certwatcher`
controller-runtime builds for `CertDir`), nothing covered the outbound scrape.

- `GetClientCertificate` serves the current cert from a controller-runtime
  `certwatcher`, so new handshakes present whatever is on disk now.
- The manager runs the watcher (`Runtime.Start` adds it), so its fsnotify watch and
  periodic re-read are active and the source needs no reload logic of its own.
- On rotation the watcher closes idle connections, so pooled keep-alives
  re-handshake with the new cert instead of lingering on the old one until they
  idle out.

Fixes #2045

**Release note**:
```release-note
The metrics and models scrape client now reloads its mTLS client certificate on rotation without an EPP restart. The CA bundle is still read once at startup.
```

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/source/http/datasource.go
+++ b/pkg/epp/framework/plugins/datalayer/source/http/datasource.go
@@ -18,6 +18,7 @@ package http
 
 import (
 	"context"
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
@@ -32,6 +33,7 @@ import (
 	"slices"
 	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"sigs.k8s.io/controller-runtime/pkg/log"
@@ -132,13 +134,11 @@ func NewHTTPDataSource[T any](scheme, path string, tlsOpts TLSOptions,
 		},
 	}
 	if scheme == "https" {
-		tlsCfg, err := tlsClientConfig(tlsOpts)
+		rt, err := newTLSReloader(tlsOpts)
 		if err != nil {
 			return nil, err
 		}
-		httpsTransport := baseTransport.Clone()
-		httpsTransport.TLSClientConfig = tlsCfg
-		cl.Transport = httpsTransport
+		cl.Transport = rt
 	}
 	return &HTTPDataSource[T]{
 		typedName:      fwkplugin.TypedName{Type: pluginType, Name: pluginName},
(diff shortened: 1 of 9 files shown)


pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -170,25 +182,82 @@ func caCertPool(path string) (*x509.CertPool, error) {
...
 		cfg.RootCAs = pool
 	}
-	if opts.ClientCertPath != "" || opts.ClientKeyPath != "" {
-		cert, err := tls.LoadX509KeyPair(opts.ClientCertPath, opts.ClientKeyPath)
-		if err != nil {
-			return nil, fmt.Errorf("%w: %w", ErrLoadClientCert, err)
+	if opts.ClientCertPath == "" && opts.ClientKeyPath == "" {
+		return cfg, noopCertRefresh, nil
+	}
+	watcher, err := certwatcher.New(opts.ClientCertPath, opts.ClientKeyPath)
```
elevran: certwatcher.New(...) is constructed here but Start(ctx) is never called anywhere, so the library's fsnotify-driven instant reload and its own periodic re-check never activate. The only reload path is throttledReload, invoked from Dispatch and gated to once per minute - meaning cert freshness is coupled to scrape traffic continuing to occur. If an endpoint stops being scraped (unhealthy, removed from rotation) before its next tick, its cached client cert never reloads regardless of how long it has rotated. If this coupling is intentional (comment says rotations are rarer than scrapes), consider calling out this tradeoff explicitly, since it's a meaningful semantic gap versus typical certwatcher usage (webhook/metrics servers rely on the passive Start()-driven watch).

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -170,25 +182,82 @@ func caCertPool(path string) (*x509.CertPool, error) {
 	return pool, nil
 }
 
+// noopCertRefresh is the refresher used when there is no client cert to reload.
+func noopCertRefresh(context.Context) {}
+
+// reloadingTransport swaps in a fresh transport on rotation, closing idle
```
elevran: The comment justifies swapping the entire Transport on rotation by "closing idle connections would leave a busy one on its old-cert handshake indefinitely" - but a TLS connection's negotiated client cert is fixed at handshake time regardless of which Transport object dialed it. Since GetClientCertificate is already re-invoked fresh on every new handshake, a single never-swapped *http.Transport with CloseIdleConnections() on rotation should produce identical observable behavior (idle conns close, busy conns finish on their already-negotiated cert, new conns pick up the new cert). The extra atomic.Pointer[http.Transport] indirection doesn't appear to buy anything the simpler form wouldn't. Worth reconsidering unless there's a behavior difference I'm missing.

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -170,25 +182,82 @@ func caCertPool(path string) (*x509.CertPool, error) {
...
-		if err != nil {
-			return nil, fmt.Errorf("%w: %w", ErrLoadClientCert, err)
+	if opts.ClientCertPath == "" && opts.ClientKeyPath == "" {
+		return cfg, noopCertRefresh, nil
+	}
+	watcher, err := certwatcher.New(opts.ClientCertPath, opts.ClientKeyPath)
+	if err != nil {
+		return nil, nil, fmt.Errorf("%w: %w", ErrLoadClientCert, err)
+	}
+	// Cached so handshakes do not take certwatcher's lock, RegisterCallback seeds it.
```
elevran: This local atomic.Pointer[tls.Certificate] cache duplicates certwatcher.CertWatcher's own cached cert, which is already exposed via GetCertificate(*tls.ClientHelloInfo) (*tls.Certificate, error) - a lock-protected read of the cache, not a disk re-read, and it ignores its argument. GetClientCertificate here could be a one-line adapter over watcher.GetCertificate(nil), dropping this cache and the RegisterCallback-based population of it (RegisterCallback would then only be needed to trigger onCertReload). The comment's stated rationale ("handshakes do not take certwatcher's lock") is a negligible optimization over the library's own RLock and adds a second source of truth to keep in sync.

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -131,14 +139,17 @@ func NewHTTPDataSource[T any](scheme, path string, tlsOpts TLSOptions,
 			Transport: baseTransport,
 		},
 	}
+	refreshCert := noopCertRefresh
 	if scheme == "https" {
-		tlsCfg, err := tlsClientConfig(tlsOpts)
+		rt := newReloadingTransport()
+		tlsCfg, refresh, err := tlsClientConfig(tlsOpts, rt.reload)
```
elevran: RegisterCallback fires its callback synchronously with the already-cached cert during setup (certwatcher.New does an initial ReadCertificate before returning), which is before rt.tlsCfg is assigned two lines below. That first rt.reload() call installs a transport with TLSClientConfig: nil, only overwritten by the explicit rt.reload() call at line 150. Not a live bug today, but fragile: this ordering depends on the second call being present. A future cleanup pass that removes it as "redundant" would silently reintroduce a window with no TLS client config. Consider assigning rt.tlsCfg before calling tlsClientConfig (e.g. via a setter closure) so a single reload() is provably correct on its own.

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -170,25 +182,82 @@ func caCertPool(path string) (*x509.CertPool, error) {
...
+	}
+	return cfg, throttledReload(watcher), nil
+}
+
+// throttledReload re-reads the cert at most once per certCheckInterval.
+func throttledReload(watcher *certwatcher.CertWatcher) func(context.Context) {
+	var lastCheck atomic.Int64
+	return func(ctx context.Context) {
+		last, now := lastCheck.Load(), time.Now().UnixNano()
+		if now-last < int64(certCheckInterval) || !lastCheck.CompareAndSwap(last, now) {
```
elevran: Two things here worth a look:

1. lastCheck.CompareAndSwap(last, now) advances the throttle timestamp before watcher.ReadCertificate() is known to succeed. A transient read failure (e.g. racing an atomic rename during rotation) locks out retry for a full certCheckInterval even though the condition might clear within seconds - and the failure is only logged, never surfaced as a metric, so a persistent reload failure is invisible until the old cert actually expires.

2. time.Now().UnixNano() is wall-clock, not monotonic. If the host clock steps backward (VM live-migration, checkpoint/restore, NTP correction), now-last goes negative and stays below certCheckInterval indefinitely, silently stalling reload detection until wall time catches back up. Low likelihood in a typical NTP-managed pod, but worth a monotonic-safe comparison (e.g. store/compare via time.Time and Sub(), or only CAS after a successful read).

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -221,6 +290,7 @@ func (s *HTTPDataSource[T]) Poll(ctx context.Context, ep fwkdl.Endpoint) (T, err
 // in DataLayerExtractErrorsTotal and do NOT surface as a returned error.
 // This keeps the collector's poll/extract counters cleanly separated.
 func (s *HTTPDataSource[T]) Dispatch(ctx context.Context, ep fwkdl.Endpoint) error {
+	s.refreshCert(ctx)
```
elevran: s.refreshCert(ctx) runs before Dispatch creates its own bounded pollCtx. On the one tick per minute where the throttle clears and ReadCertificate() actually runs, blocking disk I/O consumes the caller's shared defaultCollectionTimeout budget (1s, set in Collector.run) before Poll's own timeout is even created. A slow or contended read on that tick could push Poll past its deadline and fail the scrape for an otherwise-healthy endpoint. Likely rare in practice, but consider giving refreshCert its own short timeout independent of the poll budget.

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -170,25 +182,82 @@ func caCertPool(path string) (*x509.CertPool, error) {
...
+}
+
 // tlsClientConfig builds a tls.Config: server verification via CACertPath (or the
 // system pool), plus an mTLS client certificate when ClientCertPath is set.
-func tlsClientConfig(opts TLSOptions) (*tls.Config, error) {
-	cfg := &tls.Config{InsecureSkipVerify: opts.SkipVerify}
+// The returned func reloads that certificate and calls onCertReload, the CA does not.
+// onCertReload fires once during setup, before the caller has finished wiring.
+func tlsClientConfig(opts TLSOptions, onCertReload func()) (*tls.Config, func(context.Context), error) {
+	cfg := &tls.Config{InsecureSkipVerify: opts.SkipVerify, MinVersion: tls.VersionTLS12}
```
elevran: nit: MinVersion: tls.VersionTLS12 looks unrelated to this PR's stated purpose (cert hot-reload). Go has defaulted MinVersion to TLS 1.2 since 1.14, so this is likely a no-op restating the existing default. If it's an intentional hardening change, consider noting the reason inline.

elevran (commented):
Thanks for tackling scrape-client-cert hot reload. The overall direction (certwatcher + a refresh hook driven from Dispatch) is reasonable, but there are a few things worth addressing before merge.
Comments below at specific lines.
Also, have you consider if/how this affects traffic coming from the gatway if it is served over the same port as the metrics?

hexfusion:
reworked to let the manager run the certwatcher and dropped the dispatch driven throttle entirely. much cleaner, appreciate the review.

> Also, have you consider if/how this affects traffic coming from the gatway if it is served over the same port as the metrics?

i looked more deeply and don't think there's an interaction. this is on the EPP's outbound metrics-scrape client only. even if a target served gateway traffic and metrics on one port, the reload's `CloseIdleConnections` closes only our scrape client's own idle connections to that target, so our next scrape just re-handshakes. from the gateway connection pools are per-process, so we can't touch the gateway's connections from here. gateway traffic also reaches the epp as inbound ext-proc on a separate listener, untouched by this.

we only reload when the scrape client itself is configured with a client cert. if mtls is handled by a mesh/sidecar below the app, no cert is configured here and we do nothing.  if the cert file is managed by cert-manager or service-ca and mounted for us, rotating it on disk is exactly what this reload picks up. so it's keyed on whether a cert is configured, not on who manages it.

elevran:
> i looked more deeply and don't think there's an interaction. this is on the EPP's outbound metrics-scrape client only. even if a target served gateway traffic and metrics on one port, the reload's `CloseIdleConnections` closes only our scrape client's own idle connections to that target, so our next scrape just re-ha

I should have been clearer.
If metrics are scraped over TLS and metrics and inference share a port, that implies that the Gateway needs to initiate a TLS connection to the model server as well.
If metrics cert (or CA cert) are reloaded, then the gateway should undergo a similar change (or work with insecure-skip-verify since the initial cert/CA has expired and is no longer valid). 
The ask was to consider the feature from the point of view of additional users of the same endpoint.

pkg/epp/framework/plugins/datalayer/source/metrics/README.md
```diff
@@ -26,6 +26,7 @@ The plugin config supports:
 -   `insecureSkipVerify` (default true): Whether to skip TLS certificate verification when using the "https" scheme.
 -   `caCertPath`: PEM CA bundle to verify the target's server cert.
 -   `clientCertPath` / `clientKeyPath`: client certificate for mTLS. Set both together.
+    Reloaded on rotation without a restart. `caCertPath` is read once at startup.
```
elevran, line 29: This says the CA is read once at startup, but reloadCA in datasource.go actively re-reads the CA file and rebuilds RootCAs on change (also covered by TestTLSReloader_CAReload). The doc contradicts the code - please update to describe CA hot-reload, or confirm CA reload is not intended and fix the implementation instead.

pkg/epp/framework/plugins/datalayer/source/models/README.md
```diff
@@ -22,6 +22,7 @@ The Models Data Source polls inference server pods for model information and pas
 - `insecureSkipVerify` (bool, optional, default: `true`): Skip TLS certificate verification.
 - `caCertPath` (string, optional): PEM CA bundle to verify the target's server cert.
 - `clientCertPath` / `clientKeyPath` (string, optional): client certificate for mTLS. Set both together.
+  Reloaded on rotation without a restart. `caCertPath` is read once at startup.
```
elevran, line 25: Same issue as the metrics README: this states the CA is read once at startup, but the implementation hot-reloads it. Please align docs with actual behavior.

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -151,46 +151,200 @@ func NewHTTPDataSource[T any](scheme, path string, tlsOpts TLSOptions,
...
+		return false
+	}
+	t.current.Load().CloseIdleConnections() // re-handshake so keep-alives present the new cert
+	logger.Info("client certificate reloaded")
+	return true
+}
+
+// reloadCA reports whether the bundle was read successfully. A false return drives
+// a retry on the next scrape.
+func (t *tlsReloader) reloadCA(ctx context.Context) bool {
```
elevran, line 317: This function hot-reloads the CA bundle and swaps the transport on change, which contradicts the release note ("The CA bundle is still read once at startup") and the two README updates in this PR. Please reconcile: either the release note/docs are wrong, or CA reload here is unintended scope beyond issue #2045 (which explicitly deferred CA hot-reload to a follow-up).

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -151,46 +151,200 @@ func NewHTTPDataSource[T any](scheme, path string, tlsOpts TLSOptions,
...
+		ok = t.reloadCA(ctx) && ok
+	}
+	if !ok {
+		t.nextCheck.Store(t.since() + tlsRetryInterval.Nanoseconds()) // read failed, retry soon
+	}
+}
+
+// reloadCert reports whether the files were read successfully (a no-op re-read
+// counts as success). A false return drives a retry on the next scrape.
+func (t *tlsReloader) reloadCert(ctx context.Context) bool {
```
elevran, line 294: nit: maybeReload's nextCheck CAS already ensures only one goroutine passes the throttle gate per interval, so reloadMu is redundant here (harmless, just an extra lock with no concurrent contention to serialize). Not blocking, just noting for simplification.

elevran (commented):
Solid implementation overall: atomic pointers, throttled reload, fail-closed on read/parse errors, and good test coverage for rotation, retry, and CA-reject paths. 


1. The PR description and commit message say the client cert is served via a controller-runtime certwatcher with the manager running the watcher (Runtime.Start), with an active fsnotify watch and periodic re-read. The actual code has no certwatcher and no manager registration at all - it's a hand-rolled reloader gated on scrape traffic (RoundTrip calls maybeReload on a 10s throttle), so an unscraped endpoint never reloads. That's a materially different design from what's described. Worth correcting the PR/commit description so the history is accurate.

2. The repo already has common.CertReloader (pkg/common/certs.go) - fsnotify + debounce, atomic.Pointer[tls.Certificate] - used by ext-proc and the sidecar proxy for the same class of problem. This PR introduces a third, differently-designed reload mechanism instead of reusing or extending that one. There may be a good reason (e.g. wanting reload gated by actual scrape usage rather than fsnotify), but it'd help to state it.

3. Issue #2045 explicitly scoped the CA to load-once, deferring CA hot-reload to a follow-up ("hot-reloading the CA can be a second increment"). This PR implements CA hot-reload anyway. Might be fine, but worth confirming the scope expansion was deliberate.

elevran:
@hexfusion is this still expecte? If so:
- Conflicts should be resolved 
- Please address review comments
- Would appreciate your thoughts on [this comment](https://github.com/llm-d/llm-d-router/pull/2151#issuecomment-5088109359 )

ahg-g:
@hexfusion friendly ping, is this PR still needed?

PR #2153, opened 2026-07-23
fix(kvevents): honor configured engine adapter in precise prefix cache

wukehu:
What type of PR is this?

/kind bug

Enable precise prefix-cache routing for SGLang KV events.

The repository already provides an SGLang KV-event adapter and the
`kvEventsConfig.engineType` option, but the precise-prefix-cache producer
always instantiated the vLLM adapter. As a result, `engineType: sglang`
was not applied by the producer.

SGLang may also emit KV events with a page size smaller than the EPP
canonical block size. Processing each page independently produces no
canonical request key, which leaves subsequent `parent_block_hash`
references unresolved.

This change:

- Selects the KV-event adapter using `kvEventsConfig.engineType`.
- Accumulates sub-canonical SGLang pages before indexing.
- Preserves immutable parent-page chains for radix-tree branches.
- Cleans pending page state on block removal and cache reset.
- Documents the SGLang engine configuration.

**Validation**:

- Added a producer wiring test for unsupported engine types.
- Added cross-batch SGLang page accumulation coverage.
- Added radix-tree branch coverage.
- Passed unit tests for `pkg/kvevents`, `pkg/kvevents/engineadapter`, and
  `preciseprefixcache`.
- Cross-compiled and deployed the EPP as a Linux/amd64 binary.
- Validated a 2-prefill/2-decode SGLang deployment.
- First request produced 7 precise-index admissions.
- Repeated prefix produced 7 lookup hits and a max pod hit count of 7.
- No tokenizer or parent-chain errors were observed.

EPP restart replay of pre-existing SGLang KV events is outside the scope
of this PR.

N/A

**Release note**:

```release-note
Enable precise prefix-cache routing for SGLang KV events.

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/README.md
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/README.md
@@ -36,6 +36,9 @@ upstream. No-op otherwise.
 | `speculativeIndexing` | bool | `false` | Seed predicted entries on routing decisions. |
 | `speculativeTTL` | duration | `2s` | TTL for speculative entries. |
 
+Set `kvEventsConfig.engineType` to `sglang` for SGLang KV-events. It defaults
+to `vllm` when omitted.
+
 See [llm-d-kv-cache/docs/configuration.md](https://github.com/llm-d/llm-d-kv-cache/blob/main/docs/configuration.md)
 for nested parameter details.
 
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
@@ -80,7 +80,7 @@ type subscriberManager interface {
 }
 
 // Producer is a DataProducer plugin that maintains a KV-block prefix-cache
-// index by subscribing to vLLM KV-events and writes per-endpoint
+// index by subscribing to engine KV-events and writes per-endpoint
 // PrefixCacheMatchInfo for each request. Operators pair it with the
 // generic prefix-cache-scorer (set prefixMatchInfoProducerName to this
 // producer's instance name) to route requests by precise cache locality.
@@ -178,7 +178,11 @@ func New(ctx context.Context, name string, config PluginConfig) (*Producer, erro
 		return nil, fmt.Errorf("failed to create KVBlockScorer: %w", err)
 	}
 
-	pool := kvevents.NewPool(config.KVEventsConfig, indexer.KVBlockIndex(), tokenProcessor, engineadapter.NewVLLMAdapter())
+	adapter, err := engineadapter.NewAdapter(config.KVEventsConfig.EngineType)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create KV-events engine adapter: %w", err)
+	}
+	pool := kvevents.NewPool(config.KVEventsConfig, indexer.KVBlockIndex(), tokenProcessor, adapter)
 	pool.Start(ctx)
 
 	subscribersManager := kvevents.NewSubscriberManager(pool)
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer_test.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer_test.go
@@ -740,6 +740,22 @@ func TestNew_BlockSizeFlowsViaTokenProcessor(t *testing.T) {
 	}
 }
 
+func TestNewRejectsUnsupportedKVEventEngineType(t *testing.T) {
+	ctx, cancel := context.WithCancel(utils.NewTestContext(t))
+	t.Cleanup(cancel)
+
+	idxCfg, err := kvcache.NewDefaultConfig()
+	require.NoError(t, err)
+	kvEventsCfg := kvevents.DefaultConfig()
+	kvEventsCfg.EngineType = "unsupported"
+
+	_, err = New(ctx, "unsupported-engine", PluginConfig{
+		IndexerConfig:  idxCfg,
+		KVEventsConfig: kvEventsCfg,
+	})
+	require.ErrorContains(t, err, `unsupported engine type: "unsupported"`)
+}
+
 type fakeSubscriberManager struct {
 	ids       []string
 	endpoints []string

### CODE REVIEW

File: pkg/kvevents/pool.go [hunk 1]
```diff
@@ -143,6 +158,118 @@ func NewPool(cfg *Config, index kvblock.Index, tokenProcessor kvblock.TokenProce
...
+		return event, true
+	}
+
+	combined := cloneBlockStoredEvent(pages[len(pages)-1])
+	combined.ParentHash = parentHash
+	combined.BlockHashes = combined.BlockHashes[:0]
+	combined.Tokens = combined.Tokens[:0]
+	combined.ExtraKeys = combined.ExtraKeys[:0]
+	for i := len(pages) - 1; i >= 0; i-- {
+		combined.BlockHashes = append(combined.BlockHashes, pages[i].BlockHashes...)
```
bongwoobak: The combined event re-includes parent pages that an earlier combined block already indexed — in the branch case of `TestSGLangSubcanonicalPagesAccumulateAcrossBatches`, hashes 100..109 end up in both combineds. `trackStore` (pool.go:592) then counts them twice, but SGLang announced each of those pages once on the wire, so the node's eventual `BlockRemoved` gets suppressed by the dedup filter and the count drifts up by one per branch for shared prefixes. That breaks the wire-mirroring invariant documented on `eventDedupFilter` (event_dedup_filter.go:88).

How about passing only the triggering event's hashes to `trackStore` while still giving `Index.Add` the full combined list?

pkg/kvevents/pool.go
```diff
@@ -143,6 +158,118 @@ func NewPool(cfg *Config, index kvblock.Index, tokenProcessor kvblock.TokenProce
...
+// accumulateSGLangStore returns an event only after sub-canonical SGLang pages
+// form a full canonical block. Messages for one pod are processed in order by
+// the sharded Pool, while the mutex protects state belonging to different pods.
+func (p *Pool) accumulateSGLangStore(
+	event *BlockStoredEvent, podIdentifier, modelName string,
+) (*BlockStoredEvent, bool) {
+	canonicalBlockSize := p.tokenProcessor.BlockSize()
+	if p.engineType != "sglang" || canonicalBlockSize <= 0 ||
+		len(event.Tokens) == 0 || len(event.Tokens) >= canonicalBlockSize ||
+		canonicalBlockSize%len(event.Tokens) != 0 || len(event.BlockHashes) == 0 {
```
bongwoobak: If `page_size` doesn't divide the canonical block size (e.g. page_size 24, blockSize 64), every event falls through this gate, then produces zero request keys downstream and is dropped — no log on either path. So a config mismatch silently disables the feature this PR adds, and there's nothing to grep for in production. Same for the `tokenCount != canonicalBlockSize` branch below.

Could you add a rate-limited warning on these paths so the mismatch is diagnosable?

pkg/kvevents/pool.go
wukehu: Good catch. You're right: the combined event is an index representation,
not a new wire-level store announcement. Passing all combined hashes to
trackStore() violates the wire-mirroring invariant when a branch reuses
already indexed parent pages.

I'll preserve the triggering event's original block hashes before
accumulation, use the full combined hash list for Index.Add(), and pass
only the triggering wire event's hashes to trackStore().

I'll also add a regression test that removes a shared-prefix branch and
verifies that no extra dedup reference suppresses the final eviction.
Thanks for pointing this out.

pkg/kvevents/pool.go
wukehu: Agreed. The current fallback makes an incompatible SGLang page size look
like a normal event, while it eventually produces no canonical request
keys and silently disables precise indexing.

I'll add rate-limited warnings for:

- a page size that does not divide the canonical block size; and
- an accumulated chain whose token count exceeds or otherwise does not
  match the canonical block size.

The warning will include the pod, model, page size, token count, and
canonical block size so the configuration mismatch is diagnosable
without flooding logs. I'll add unit coverage for both paths.

pkg/kvevents/pool.go
```diff
@@ -143,6 +165,137 @@ func NewPool(cfg *Config, index kvblock.Index, tokenProcessor kvblock.TokenProce
...
+		len(event.BlockHashes) == 0 {
+		return event, true
+	}
+	if canonicalBlockSize%len(event.Tokens) != 0 {
+		if p.pageSizeWarningLimiter.Allow() {
+			log.FromContext(ctx).Info("SGLang KV page size does not divide the canonical block size",
+				"podIdentifier", podIdentifier,
+				"modelName", modelName,
+				"pageSize", len(event.Tokens),
+				"tokenCount", len(event.Tokens),
```
bongwoobak: nit: duplicates `pageSize` on this path — both log `len(event.Tokens)`.
```suggestion
```

pkg/kvevents/pool.go
```diff
@@ -113,7 +116,22 @@ type Pool struct {
...
-	wg    sync.WaitGroup
+	// SGLang can emit one KV event per page even when its page size is smaller
+	// than the canonical index block size. Keep those immutable page links until
+	// they form a complete canonical block; otherwise the first partial page is
+	// discarded and every following parent hash becomes unresolvable.
+	engineType               string
+	pendingStoreMu           sync.Mutex
+	pendingSGLangStore       map[sglangPendingStoreKey]*BlockStoredEvent
+	pageSizeWarningLimiter   *rate.Limiter
+	tokenCountWarningLimiter *rate.Limiter
```
bongwoobak: nit: these limiters are global, so one misconfigured pod's warning can starve another pod's for the whole interval. Fine at 1/min if intentional — a per-pod key would make multi-pod mismatches visible, but I don't feel strongly.

pkg/kvevents/pool.go
```diff
@@ -336,6 +489,12 @@ func (p *Pool) processEventBatch(ctx context.Context, batch *EventBatch, podIden
 	for _, genericEvent := range batch.Events {
 		switch ev := genericEvent.(type) {
 		case *BlockStoredEvent:
+			wireBlockHashes := append([]uint64(nil), ev.BlockHashes...)
```
bongwoobak: nit: this copies on every BlockStored, including the vLLM path where `ev` is never replaced. Copying inside `accumulateSGLangStore` (or only when `engineType == "sglang"`) keeps the hot path allocation-free.

pkg/kvevents/pool.go
wukehu: This is intentional for now. The page-size mismatch is a deployment/model configuration error, so pods selected by one EPP are expected to share the same page-size configuration; one warning per failure class per minute is sufficient to surface that condition without adding keyed limiter state. A per-pod limiter would also need lifecycle cleanup to avoid unbounded cardinality as pods churn. If the EPP later supports heterogeneous page sizes within one pool, we should revisit this as a pod-keyed limiter and clear it on pod removal/reset

pkg/kvevents/pool.go
wukehu: Addressed in 1c2d0790. Removed the duplicate tokenCount log field.

pkg/kvevents/pool.go
wukehu: Addressed in 1c2d0790. The block-hash copy is now limited to the SGLang path, keeping the vLLM hot path allocation-free.

ahg-g:
Is this ready to merge? @rahulgurnani have we not tested precise kv-cache routing with sglang before?

sagearc (changes requested):
Thanks for catching the adapter mismatch. I think the current title may overstate the scope and combines two separate changes:
1. Fix the adapter mismatch.
2. Accumulate sglang pages across events to form a complete canonical block.

The first is a straightforward correctness fix, although the documented setup already works because the vllm adapter can parse sglang's compatible msgpack arrays.

Regarding (2), **this makes the kvevent pool stateful**. It is only needed when the sglang page size is smaller than the router’s canonical block size. In the sglang's precise guide, both are 64.

Would you consider moving this to a follow-up PR so the stateful design can be discussed separately?

rahulgurnani:
can you share more details about the validations you did? (mentioned in the PR description)

wukehu:
Thanks. I reran the validation and also revisited the earlier "no hit" result, because the original test method could produce a false conclusion.

The earlier check used the CLI's short default prompt. The resulting request had only 36 prompt tokens while the EPP canonical block size was 64. All three repeated requests returned HTTP 200 and the PD routing conformance checks passed, but the prompt could not form one complete canonical block. The EPP counters were unchanged before and after those replays:

```
admissions / lookup_requests / lookup_hits = 2 / 1 / 0
```

So request success, or even backend cache activity, was not evidence that the EPP precise index had admitted or matched a key. A valid precise-routing check needs a prompt longer than one canonical block and must compare EPP index counters before and after replay.

I then ran a strict A/B with the same 226-token prompt, request ID, 2P2D deployment, SGLang `page_size=64`, and EPP `blockSizeTokens=64`:

| Revision | Before | After | Delta |
| --- | --- | --- | --- |
| Parent `aced3336` (hard-coded vLLM adapter) | 2 / 1 / 0 | 5 / 4 / 6 | +3 / +3 / +6 |
| PR `b268d324` (configured SGLang adapter) | 2 / 1 / 0 | 5 / 4 / 6 | +3 / +3 / +6 |

For both revisions:
- all three requests returned HTTP 200;
- the repeated requests selected the same Decode pod;
- replays 2 and 3 reported `cached_tokens=192`;
- all routing conformance summaries passed.

Therefore, the documented 64/64 setup already works on the parent because the vLLM adapter accepts SGLang's currently compatible msgpack arrays. This PR should not claim that it newly enables SGLang precise routing. The remaining value of the current adapter-only change is to honor the existing `engineType` configuration and fail fast for unsupported values. Cross-event accumulation for sub-canonical SGLang pages is not part of the current change and should be discussed separately if needed.

sagearc:
Thanks, since the pr now contains only the adapter mismatch fix, it looks good to me @ahg-g @bongwoobak @rahulgurnani 

> Accumulate sglang pages across events to form a complete canonical block.

@wukehu can you please open an issue on that?

sagearc (approved):

wukehu:
Opened #2235 to track the cross-event accumulation design for sub-canonical SGLang pages, including wire-level store/removal accounting, pending-state lifecycle, diagnostics, and runtime acceptance criteria.

liu-cong (approved):

rahulgurnani:
Thanks for submitting this fix!

PR #2155, opened 2026-07-23
ci(coordinator): Align coordinator CI image build with the router's m…

### LINKED ISSUE #2154: ci(coordinator): Align coordinator CI image build with the router's matrixed, cached pattern
Following comment from https://github.com/llm-d/llm-d-router/pull/1975#issuecomment-4979754959

Restructure the image-build stage of ci-coordinator.yaml to match the pattern already used by the router in ci-pr-checks.yaml:

- Replace the single serial job that builds builder + coordinator + epp with a strategy.matrix job (e2e-images), one leg per image, each exporting a .tar artifact.
- Add an e2e-images-status aggregator job so a single required check can gate the matrix (individual matrix legs cannot be marked required in branch protection).
- Have the e2e-tests job download and docker load the artifacts, then run the non-rebuilding target test-e2e-coordinator-run instead of rebuilding images inline.
- Use the shared .github/actions/docker-build-and-push action with push: 'false' and type=gha layer caching (scope=<image-name>), matching the router.
- Unify the builder image name to llm-d-builder (shared Dockerfile.builder), so the coordinator and router share one builder image and one build cache scope instead of duplicating an identical image under llm-d-coordinator-builder. Requires moving BUILDER_IMAGE_NAME in Makefile.coord.mk in lockstep so the e2e run container resolves.

Why is this needed:

This addresses a review comment from https://github.com/llm-d/llm-d-router/pull/1975#issuecomment-4979754959, which flagged that the coordinator CI builds builder + coordinator + epp images serially in one job, unlike the router's matrixed/cached image-build jobs in ci-pr-checks.yaml, and suggested aligning them in a separate PR.

Aligning them:

- Runs image builds in parallel instead of serially, reducing e2e wall-clock time.
- Reuses BuildKit layers across runs via the GHA cache, and lets the coordinator and router share the llm-d-builder and llm-d-router-endpoint-picker cache scopes since they build identical Dockerfiles.
- Removes a duplicate builder image (llm-d-coordinator-builder) that was byte-identical to llm-d-builder.
- Keeps the two CI workflows consistent, so future changes to the build/cache mechanism apply uniformly.

revit13:
**What type of PR is this?**
/kind cleanup
/kind test

Aligns the coordinator CI image build (`ci-coordinator.yaml`) with the router's
matrixed, cached pattern in `ci-pr-checks.yaml`, instead of building the images
serially in one job.

- Replaces the single serial build job with a `strategy.matrix` job
  (`e2e-images`), one leg per image (builder, coordinator, epp built;
  simulator pulled). Each leg exports its image as a `.tar` artifact.
- Adds an `e2e-images-status` aggregator job so a single required status
  check can gate the matrix (individual matrix legs cannot be marked required
  in branch protection).
- The `e2e-tests` job now downloads and `docker load`s the artifacts, then
  runs the non-rebuilding target `test-e2e-coordinator-run` instead of
  rebuilding images inline.
- Builds go through the shared `.github/actions/docker-build-and-push` action
  with `push: 'false'` and `type=gha` layer caching (`scope=<image-name>`),
  matching the router.
- Unifies the builder image name to `llm-d-builder` (the coordinator and
  router share a single `Dockerfile.builder`), so both reuse one builder image
  and one build cache scope instead of duplicating an identical image under
  `llm-d-coordinator-builder`. `BUILDER_IMAGE_NAME` in `Makefile.coord.mk`
  moves in lockstep so the e2e run container resolves.

This addresses a review comment from
https://github.com/llm-d/llm-d-router/pull/1975#issuecomment-4979754959,
which flagged the serial build as a follow-up for a separate PR.

Fixes #2154

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci-coordinator.yaml
+++ b/.github/workflows/ci-coordinator.yaml
@@ -1,7 +1,9 @@
 name: CI - Coordinator Test
 
 # Runs coordinator unit tests and e2e tests. Path filter excludes paths not
-# exercised by the coordinator.
+# exercised by the coordinator. The e2e-images matrix job is followed by an
+# e2e-images-status job that aggregates its result, since GitHub required
+# status checks cannot reference individual matrix legs.
 on:
   push:
     branches:
@@ -40,6 +42,7 @@ jobs:
               - scripts/**
               - hack/**
               - test/**
+              - .github/actions/docker-build-and-push/action.yml
               - .github/actions/e2e-runner-setup/action.yml
               - .github/workflows/ci-coordinator.yaml
               - '!Dockerfile.sidecar'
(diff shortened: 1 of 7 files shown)


roytman (dismissed):

.github/workflows/ci-coordinator.yaml
```diff
@@ -40,6 +42,7 @@ jobs:
               - scripts/**
               - hack/**
               - test/**
+              - .github/actions/docker-build-and-push/action.yml
```
elevran, line 46: The `src` filter has no entry for `config/coordinator/**`, but `Dockerfile.coordinator` copies that directory into the image and it is the coordinator's default runtime config. A PR that only edits `config/coordinator/coordinator.yaml` matches none of these globs, so `unit-tests`/`e2e-images`/`e2e-tests` all report skipped rather than failed, and a broken config can ship without ever being exercised by e2e.

.github/workflows/ci-coordinator.yaml
```diff
@@ -98,9 +176,40 @@ jobs:
...
+            docker load --input "$image"
+          done
+
       - name: Run coordinator e2e tests
         shell: bash
         env:
           GO_MOD_CACHE_VOL: ${{ steps.e2e-setup.outputs.go-mod-cache }}
           GO_BUILD_CACHE_VOL: ${{ steps.e2e-setup.outputs.go-build-cache }}
-        run: make -f Makefile.coord.mk test-e2e-coordinator
+        run: make -f Makefile.coord.mk test-e2e-coordinator-run
```
elevran, line 217: This switches the CI-invoked target from `test-e2e-coordinator` (self-building) to `test-e2e-coordinator-run` (requires pre-loaded images via `image-pull`, which does not pull builder/coordinator images). A developer copying this exact command from a CI log to reproduce locally will hit stale or missing images with no indication that `test-e2e-coordinator` is the right local entry point. Worth a note in README.coord.md or the Makefile help text.

.github/workflows/ci-coordinator.yaml
```diff
@@ -85,10 +88,85 @@ jobs:
...
+        if: ${{ matrix.image.source == 'build' }}
+        uses: actions/checkout@v7
+
+      - name: Build ${{ matrix.image.name }} image
+        if: ${{ matrix.image.source == 'build' }}
+        uses: ./.github/actions/docker-build-and-push
+        with:
+          docker-file: ${{ matrix.image.docker-file }}
+          image-name: ${{ matrix.image.image-name }}
+          tag: dev
```
elevran, line 132: This bakes `BUILD_REF=dev` into the EPP image via `docker-build-and-push`, while local builds through `Makefile.coord.mk`'s `image-build-%` rule pass `BUILD_REF` from `git describe --tags`. No current e2e test asserts on this, and `Dockerfile.coordinator` never wires `COMMIT_SHA`/`BUILD_REF` into its ldflags at all, so it is inert today - but the EPP image's build-ref in logs/metrics/tracing will always read "dev" from this path.

Dn't think that is the intended behavior.

.github/workflows/ci-coordinator.yaml
```diff
@@ -85,10 +88,85 @@ jobs:
           GO_BUILD_CACHE_VOL: ${{ steps.go-cache.outputs.build }}
         run: make -f Makefile.coord.mk test-unit
 
-  e2e-tests:
+  e2e-images:
```
elevran, line 92: nit: this is a third near-identical copy of the build + upload-artifact + download-artifact + load scaffold (the router's ci-pr-checks.yaml already has two). Not asking to change it here since issue #2154 scopes this PR as aligning with the existing pattern rather than extracting a shared workflow, but flagging as a candidate for a future `workflow_call` (the repo already has one in `ci-build-images.yaml`).

elevran (commented):
Clean, faithful port of the router's matrix-build-images pattern to the coordinator workflow. One real CI coverage gap below, plus two lower-severity notes and a non-blocking follow-up suggestion.

elevran (approved):

PR #2156, opened 2026-07-23
fix: flowcontrol - finalize requests during shutdown

### LINKED ISSUE #2099: [Flow Control] Request submitted during shutdown can strand until TTL with no finalization
**What happened**:

A `Submit` racing processor shutdown can succeed after the drain: the shutdown-flag check (`controller/internal/processor.go:137`) passes before `isShuttingDown.Store` (`:500`); the one-shot drain loop (`:503-516`) empties the channel and exits; the racing send (`:140-147`) then lands in the buffer and Submit returns nil. The item is stranded on `enqueueChan` with no reader, and the caller blocks in `awaitFinalization` until its TTL fires or the client disconnects -- indefinitely if no TTL is configured. Bounded by `EnqueueChannelBufferSize` (default 100) items.

Requests arriving during shutdown are finalized promptly with the shutdown outcome (503 ShuttingDown), like every other request the drain path handles.

Race `Submit` against processor shutdown; see `TestConcurrentEnqueueDuringShutdown` (`pkg/epp/flowcontrol/integration_test.go:805`), which exercises the window but does not currently assert bounded unblock time without a TTL.

Part of #1187.

Proposed fix (tactical): in `awaitFinalization` (`controller/controller.go:383-403`), also select on the controller's shutdown signal and finalize with `QueueOutcomeRejectedOther` + `ErrFlowControllerNotRunning` (translates to 503 ShuttingDown). Finalization is idempotent (`TestFlowItem_Finalize_Idempotency`) and the sweep/evictAll paths tolerate double-finalization. Stranded items remain unread on the buffered channel until process exit, which is acceptable during shutdown.

The structural fix is the post-flip engine merge tracked on #1187 (the controller/processor handoff seam exists to route across N workers; there is now exactly one); this issue only closes the user-visible gap. Acceptance: strengthen the test above to assert every caller unblocks within a bound WITHOUT relying on request TTL.

Release note: shutdown no longer strands late-arriving requests until TTL.

**Environment**:
- Install tools: n/a
- Others: n/a

nicole-lihui:
**What type of PR is this?**

/kind bug

Finalizes requests that race with Flow Controller shutdown instead of leaving them stranded until request TTL or client cancellation.
Shutdown outcomes preserve the distinction between rejected requests and requests evicted from the queue.

Fixes #2099

```release-note
Requests submitted during Flow Controller shutdown are finalized promptly with a shutdown response instead of waiting for request TTL.
```


pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -388,13 +395,20 @@ func (fc *FlowController) awaitFinalization(
 	case <-reqCtx.Done():
 		// Asynchronous Finalization (Controller-initiated):
 		// The request Context expired (Cancellation/TTL) while the item was being processed.
+		if fc.parentCtx.Err() != nil {
```
LukeAVanDrie, line 398: Thanks for making this deterministic; good catch!

LukeAVanDrie (approved):
lgtm, thanks @nicole-lihui!

PR #2162, opened 2026-07-24
fix(flowcontrol): map pre-admission TTL/cancel to eviction statuses i…

### LINKED ISSUE #2097: [Flow Control] [Graduation Blocker] Pre-admission TTL/cancel returns 500; registry lookup failure mislabeled as saturation (429)
**What happened**:

1. A request whose TTL expires or whose client disconnects BEFORE the item is admitted to a queue gets HTTP 500. `inferOutcome` (`pkg/epp/flowcontrol/controller/internal/item.go:174-198`) maps TTL/cancel on a not-yet-queued item to `QueueOutcomeRejectedOther`, and `translateFlowControlOutcome` maps non-shutdown `RejectedOther` to 500 Internal (`pkg/epp/requestcontrol/admission.go:238-243`). "Not yet queued" includes items buffered in the enqueue channel and callers blocked in `SubmitOrBlock` -- exactly the overload conditions where TTL expiries peak. The same client behavior an instant later (item queued) yields `EvictedTTL`/`EvictedContextCancelled` and
503. There is also a narrow race with the same result: `Finalize` infers queued-ness from the item handle (`item.go:132`), which `managedQueue.Add` sets with no lock shared with the finalizing goroutine, so a TTL firing mid-Add lands in the same 500 path.
2. `tryDistribution` finalizes a ManagedQueue lookup failure for a leased flow (logged as "Invariant violation") as `QueueOutcomeRejectedCapacity` (`controller/controller.go:353-359`), surfacing an internal error as 429 with the `Saturated` dropped-reason header.

Client-caused terminations (TTL, disconnect) map to the same status pre-admission as post-admission -- never 500. Note #2046 (PR #2047) proposes changing the post-admission queued-TTL mapping itself to 429 when the pool has endpoints; whichever way that lands, the pre-admission path must follow it. Internal invariant violations return 500 without a saturation header.

Unit level: `translateFlowControlOutcome(QueueOutcomeRejectedOther, fmt.Errorf("%w: %w", ErrRejected, ErrTTLExpired))` returns Internal (500). End to end: fill the enqueue channel (default buffer 100) so submissions block, then cancel a waiting request; the response is 500.

Part of #1187; blocks enabling the gate by default. Under a saturation stall plus TTL storm, operators see a wall of 500s (and 500-based alerts) for what is actually backpressure.

Proposed fix: in the `RejectedOther`/`EvictedOther` case of the translator, branch on `errors.Is(err, ErrTTLExpired)` and `errors.Is(err, ErrContextCancelled)` before the Internal fallback, mapping each to the SAME status + dropped-reason header as its post-admission equivalent (`EvictedTTL` / `EvictedContextCancelled`) -- i.e. delegate to that mapping rather than hardcoding, so this stays correct whichever way #2046 / PR #2047 resolves the queued-TTL 429-vs-503 question. Change the `tryDistribution` lookup-failure finalization to `QueueOutcomeRejectedOther` with a non-sentinel error. The existing `TestTranslateFlowControlOutcome` table covers only RejectedOther+shutdown and +generic; add rows for the new branches first. Coordinate with #2046/#2047 (same function, overlapping tests).

Release note: pre-admission TTL/cancel no longer returns 500.

**Environment**:
- Install tools: n/a
- Others: n/a

LukeAVanDrie:
**What type of PR is this?**

/kind bug

Two outcome-mapping fixes in flow control admission.

1. A TTL expiry or client disconnect that fires before an item reaches a queue (buffered in the enqueue channel, blocked in submission, or racing `managedQueue.Add`) finalizes as `QueueOutcomeRejectedOther`, which `translateFlowControlOutcome` mapped to 500. Once the item is queued, the same event maps to an eviction status with a dropped-reason header. These pre-admission windows are widest under overload, so a saturation stall with TTL expiries showed up as a spike of 500s that was really backpressure.

   The translator now recognizes `ErrTTLExpired` and `ErrContextCancelled` in the `RejectedOther`/`EvictedOther` arm (after the shutdown check, which keeps precedence) and delegates to the matching eviction mapping, so the two paths cannot drift apart. The `ttlPoolEmpty` discriminator from #2047 threads through the delegation, and the pool-emptiness probe in `Admit` also fires for these outcomes when they wrap `ErrTTLExpired`, so a pre-admission TTL expiry gets the same split as a queued TTL eviction: 429 with the ttl-expired reason when the pool has endpoints, 503 with the no-endpoints reason when it is empty. The same branches cover `EvictedOther` for symmetry; nothing currently emits it with these sentinels.

2. `tryDistribution` finalized a ManagedQueue lookup failure for a leased flow as `QueueOutcomeRejectedCapacity`, reporting an internal invariant violation as 429 with the `Saturated` header. It now finalizes as `RejectedOther`: a 500 with no saturation header. This also keeps the `RejectedCapacity` label in `flow_control_requests_total` to genuine capacity rejections. The registry error is flattened with `%v` because wrapping it would let `ErrPriorityBandNotFound` flow back through the connection closure and trip the priority-0 fallback in `withConnectionWithFallback`.

Tests: new `TestTranslateFlowControlOutcome` rows for pre-admission TTL (both pool states) and cancel on both outcomes, plus a shutdown-precedence row. New `TestFlowControlAdmissionController_Admit` rows cover the pre-admission TTL 429/503 split through the `Admit` probe. The controller lookup-failure test now also asserts that registry sentinels do not leak into the finalized error and that no fallback retry happens. Both packages pass with `-race`.

Fixes #2097

Part of #1187.

**Release note**:

```release-note
NONE
```


LukeAVanDrie:
I would like to hold this until #2047 is merged upon which I will make the updates mentioned in the PR description.

ahg-g:
> I would like to hold this until #2047 is merged upon which I will make the updates mentioned in the PR description.

merged now

LukeAVanDrie:
Rebased and ready for review

liu-cong (approved):

PR #2180, opened 2026-07-25
docs: reflect flow controls GA posture (formerly experimental)

LukeAVanDrie:
**What type of PR is this?**

/kind feature

Repurposed from the default flip, per the decision below: flow control is a GA feature that is enabled explicitly, because saturation detection still needs per-deployment tuning and enabling the layer should be a deliberate choice by an operator who has read the tuning guidance (https://github.com/llm-d/llm-d-router/pull/2180#issuecomment-5119639333). The `flowControl` feature gate keeps its `false` default; this PR carries the graduation around that posture:

- The gate-off and init log lines no longer call the layer "Experimental".
- The loader warns when a `flowControl:` config section is present while the gate is off. Everything in the section except `saturationDetector` (which the legacy admission path also uses) is silently ignored, and with an opt-in gate that misconfiguration is easy to hit: write the section, forget the gate.
- A test pins the registered default, so a future flip is a deliberate edit rather than a side effect.
- Docs: the flow control README covers enablement, the behavior change when the gate is on, the tuning knobs, and the dependency on the EPP's model-server metrics scrape staying fresh. operations.md accounts for queued request bodies in EPP memory sizing and notes that queues, fairness, and the saturation view are per replica in Active-Active. architecture.md and the `FeatureGates` API docs cover per-gate defaults and the `name=false` syntax.

For reviewers of the earlier revision: the flip, the `ENABLE_EXPERIMENTAL_FLOW_CONTROL_LAYER` no-op change (the variable has since been removed on main), and the test changes that only existed to absorb a default-on world are gone. The ignored-config warning, the log lines, and the docs remain, rewritten for the opt-in default.

Part of #1187. The flip itself (#2104) is deferred, not fixed; it becomes the revisit item if the tuning requirement goes away.

```release-note
NONE
```

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -912,13 +912,13 @@ func (r *Runner) initAdmissionControl(
 	endpointCandidates contracts.EndpointCandidates,
 ) (contracts.EndpointCandidates, requestcontrol.AdmissionController, contracts.PriorityBandControlPlane) {
 	if !r.featureGates[flowcontrol.FeatureGate] {
-		setupLog.Info("Experimental Flow Control layer is disabled, using legacy admission control")
+		setupLog.Info("Flow Control layer is disabled via the flowControl feature gate, using legacy admission control")
 		return endpointCandidates,
 			requestcontrol.NewLegacyAdmissionController(eppConfig.SaturationDetector, endpointCandidates),
 			nil
 	}
 	endpointCandidates = requestcontrol.NewCachedEndpointCandidates(ctx, endpointCandidates, 50*time.Millisecond)
-	setupLog.Info("Initializing experimental Flow Control layer")
+	setupLog.Info("Initializing Flow Control layer")
 	registry := fcregistry.NewFlowRegistry(eppConfig.FlowControlConfig.Registry, setupLog)
 	fc := fccontroller.NewFlowController(
 		ctx,
--- a/pkg/epp/flowcontrol/README.md
+++ b/pkg/epp/flowcontrol/README.md
@@ -30,6 +30,47 @@ between the Routing and Scheduling layers. It decides *if* and *when* a request,
 mechanism for managing diverse SLOs, ensuring fairness among competing workloads, and maintaining system stability under
 high load.
 
+### Enabling, disabling, and tuning
+
+Flow control is disabled by default and is enabled explicitly, via the `flowControl` feature gate
+in the EPP config:
+
+```yaml
+featureGates: ["flowControl"]
+```
+
+With the gate off, the legacy saturation-only admission path is used. With it on, saturation no
+longer pushes excess requests into each model server's local queue: they wait in the EPP and
+dispatch by priority as capacity frees, with queue wait bounded by a TTL. Sheddable
+(negative-priority) requests buffer under the same TTL and capacity rules instead of being
+rejected immediately on saturation. Enabling the layer changes queueing behavior under load, so
+review the tuning knobs below as part of turning it on.
+
+Setting a `flowControl:` config section while the gate is disabled logs a warning: everything in it
+except `saturationDetector` (which the legacy admission path also uses) is ignored.
+
+Operationally, dispatch decisions depend on fresh endpoint metrics: saturation detection reads the
+model-server metrics that the EPP's own data layer scrapes from pods (this is independent of
+cluster monitoring such as Prometheus). Endpoints whose metrics are older than the detector's
+`metricsStalenessThreshold` count as fully saturated, so if the EPP loses its scrape path to all
+pods (network policy, metrics port change, a starved refresh loop), dispatch halts and queued
+requests eventually shed at their TTL. Keep the refresh interval comfortably inside the staleness
+threshold and monitor scrape health when running with flow control on.
+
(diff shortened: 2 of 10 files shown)


loicmarchal:
@LukeAVanDrie I looked at it and everything looks good to me. Since `ENABLE_EXPERIMENTAL_FLOW_CONTROL_LAYER` is now a no-op (and the depreciation log makes it clear), #2065 can probably be closed once this PR is merged.

ahg-g:
As discussed on slack, please keep the gate off by default and continue to have this an opt-in feature for now.

LukeAVanDrie:
Updated per discussion and ready for review. I kept the docs and tests, but reverted the flip.

pkg/epp/flowcontrol/README.md
```diff
@@ -30,6 +30,47 @@ between the Routing and Scheduling layers. It decides *if* and *when* a request,
...
+dispatch by priority as capacity frees, with queue wait bounded by a TTL. Sheddable
+(negative-priority) requests buffer under the same TTL and capacity rules instead of being
+rejected immediately on saturation. Enabling the layer changes queueing behavior under load, so
+review the tuning knobs below as part of turning it on.
+
+Setting a `flowControl:` config section while the gate is disabled logs a warning: everything in it
+except `saturationDetector` (which the legacy admission path also uses) is ignored.
+
+Operationally, dispatch decisions depend on fresh endpoint metrics: saturation detection reads the
+model-server metrics that the EPP's own data layer scrapes from pods (this is independent of
```
liu-cong: >this is independent of
cluster monitoring such as Prometheus

I don't quite follow.... Consider just deleting it

liu-cong:
```suggestion
model-server metrics that the EPP's own data layer scrapes from endpoints (this is independent of
```

pkg/epp/flowcontrol/README.md
```diff
@@ -30,6 +30,47 @@ between the Routing and Scheduling layers. It decides *if* and *when* a request,
...
+cluster monitoring such as Prometheus). Endpoints whose metrics are older than the detector's
+`metricsStalenessThreshold` count as fully saturated, so if the EPP loses its scrape path to all
+pods (network policy, metrics port change, a starved refresh loop), dispatch halts and queued
```
liu-cong:
```suggestion
endpoints (network policy, metrics port change, a starved refresh loop), dispatch halts and queued
```

cmd/epp/runner/runner.go
```diff
@@ -874,13 +874,13 @@ func (r *Runner) initAdmissionControl(
 	endpointCandidates contracts.EndpointCandidates,
 ) (contracts.EndpointCandidates, requestcontrol.AdmissionController, contracts.PriorityBandControlPlane) {
 	if !r.featureGates[flowcontrol.FeatureGate] {
```
liu-cong, line 914: A feature gate is supposed to be removed eventually once the feature goes to GA. we should consider making flow control a command line flag for Optin instead of a feature gate.

@ahg-g thoughts?

LukeAVanDrie: The gate here is still mid-lifecycle rather than a permanent GA switch. This PR "graduates" the quality posture, but the enablement lifecycle is unchanged. #2104 (default-on) is the deferred revisit item, and removing the gate follows that flip once the saturation tuning requirement goes away (or at least that's what I am suggesting as the enabled by default bar).

A CLI flag would entrench the opt-in as permanent surface we'd later have to deprecate, would break existing configs that already set featureGates: ["flowControl"], and would split enablement from the `flowControl:` section it activates.

If we do land on a flag, I'd rather do it as a follow-up than re-plumb the opt-in here. WDYT @ahg-g?

This is largely a doc / test change for polish before the release cut.

pkg/epp/flowcontrol/README.md
LukeAVanDrie: done, thanks

pkg/epp/flowcontrol/README.md
LukeAVanDrie: done, thanks

pkg/epp/flowcontrol/README.md
LukeAVanDrie: Removed, this is already captured in data layer docs. Was trying to explain that flow control with default detector (utilization) has a dependency on fresh endpoint metrics. Originally this PR was enabling FC by default, so users would be subject to this automatically. Thought, pretty much all of EPP relies on this, so this note really isn't needed.

cmd/epp/runner/runner.go
liu-cong: I am OK deferring that, but please update the PR title to focus on doc/test change. "graduating" a feature usually means making the feature gate ON by default.

liu-cong: Also consider removing the the release note since there is really no user facing change except the warning log?

LukeAVanDrie: done and done

LukeAVanDrie:
Fixed commit signing and rebased

liu-cong (approved):

PR #2184, opened 2026-07-25
feat(flowcontrol): add per-band and global capacity utilization gauges

### LINKED ISSUE #2102: [Flow Control] [Graduation Blocker] Per-band and global queue utilization ratio gauges

Precomputed utilization ratio gauges (not raw capacity values):
- `flow_control_queue_utilization_requests{priority, inference_pool}` and `..._utilization_bytes{...}` in [0,1] = occupancy / effective limit, per priority band, plus global series when global limits are configured. For an unlimited dimension: omit the series (preferred) or emit 0 -- decide in review.
- Instrumented where the counters already live (registry stats propagation or the dispatch cycle). Deliberately NOT built on registry `Stats()`, which is slated for removal with the post-flip engine refactor.

**Why is this needed**:

Only occupancy gauges exist (`flow_control_queue_size`/`_bytes`) while limits live in config, so operators cannot alert on "queue at N% of its limit" without joining config into PromQL. This is also the demand-pressure half of a composite autoscaling signal (e.g. KEDA): demand = queue utilization, supply = `pool_saturation`. Part of #1187; blocks enabling the gate by default (alerting bar).

Acceptance: extend `TestFlowControlMetricsEmitted` to assert the ratio reflects occupancy/limit and returns to 0 after drain; help text documents the range and the unlimited-dimension behavior. Release note: new metrics.

sudoalok:
adds the capacity utilization gauges from #2102 so operators can alert on "queue is at N% of its limit" without joining config values into the PromQL.

flow_control_capacity_utilization_requests and _bytes, labels {priority, inference_pool} to match flow_control_requests_total, llm_d_epp prefix only. computed in the dispatch cycle right next to the pool_saturation emit since utilization is basically the demand side twin of it. per band series always get emitted, global aggregate only when a global limit is set, and an unlimited dimension gets skipped instead of showing up as 0.

few notes:

started out as queue_utilization like the issue says, renamed it after shmuelk's review. "queue" already means the per flow queue in this code (flow_control_queue_size and _bytes both carry fairness_id) so the old name read like a per flow id metric. these are per priority band, aggregated over all the flows in the band. capacity is what the contract already calls the limit so the name is just occupancy / configured capacity now.

the global aggregate uses priority="" while the per band ones carry the actual band number. can switch to a named sentinel if you'd rather have that.

it reads registry.Stats() once per dispatch cycle like we discussed on #1187, its periodic so not on the enqueue hot path, and the metric names/labels stay the same when the engine merge moves the data source later.

extended TestFlowControlMetricsEmitted with a capacityUtilizationGauge helper, same shape as queueSizeGaugeSum. the harness now sets maxRequests alongside bandMaxRequests so it checks the actual ratios (0.1 for the band, 0.25 for the aggregate with one request queued) and that both go back to 0 after drain, instead of just asserting > 0 like i had it before.

also added both metrics to docs/metrics.md, every other flow control metric is documented there and these werent.

build/vet/fmt clean, flowcontrol and metrics suites pass, race clean too.

Fixes #2102

**Release note**:
```release-note
Added flow control capacity utilization gauges. `llm_d_epp_flow_control_capacity_utilization_requests` and `llm_d_epp_flow_control_capacity_utilization_bytes` report occupancy as a fraction of a priority band's effective capacity (0.0-1.0), labelled by `priority` and `inference_pool`. `llm_d_epp_flow_control_global_capacity_utilization_requests` and `llm_d_epp_flow_control_global_capacity_utilization_bytes` report the all-bands rollup, emitted only when a global capacity is configured. Operators can alert on "band at N% of its limit" without joining config values into PromQL.
```


shmuelk:
Please remove the file release-notes.d/unreleased/2184.md and add the release notes to the proper (and missing) release notes section of the PR description.

shmuelk:
The PR's description and code talks about queue utilization. I thought the metrics would be per queue. That would actually be quite bad, as it would be priorityBand per flow ID. There are those that want to assign every user session its own flow ID.

Instead the metrics are per priorityBand, which is much more contained. I think better names are needed here. Unfortunately as a true Software Engineer, I'm having trouble coming up with a better name/term.

sudoalok:
yeah good catch on both

 i didnt realize the release-notes bot generates that from the PR description, i had hand written it (and numbered it 2190 by mistake on top of that). the note is in the description now under the release-note block.

on the naming you're right that the name was misleading. these are per priority band, aggregated over all the flows in the band, never per flow id. the problem is "queue" already means the per flow queue in this code, flow_control_queue_size and flow_control_queue_bytes both carry fairness_id, so a queue_utilization with no fairness_id reads exactly like the per flow id explosion you were worried about.

renamed them to flow_control_capacity_utilization_requests and flow_control_capacity_utilization_bytes. capacity is what the contract already calls the limit (CapacityRequests, CapacityBytes, TotalCapacityRequests in PriorityBandStats/AggregateStats), so the name is literally occupancy / configured capacity, and the priority label carries the level, with priority="" for the aggregate across all bands. i thought about a band_ prefix but that would be wrong for the "" row.

@LukeAVanDrie flagging this since #2102 spells out flow_control_queue_utilization_* by name. happy to go back to the issue's name if you'd rather keep that contract.

pkg/epp/flowcontrol/integration_test.go
```diff
@@ -1163,7 +1164,10 @@ func TestFlowControlMetricsEmitted(t *testing.T) {
 	eppmetrics.Register()
```
LukeAVanDrie, line 1164: Non-blocking. `queueSizeGaugeSum` is safe against cross-test contamination because `fairness_id` is unique per test. The capacity series is keyed only by `(priority, inference_pool)` and every harness here uses `inference_pool="integration-test"`, so an earlier sequential test leaving a nonzero band-0 value would satisfy your `Eventually(> 0)` on stale data and fail the `InDelta(0.1, ...)` below it.

Both series are absent at test entry today, so this is latent. One line makes it order-independent:

```suggestion
	eppmetrics.Register()
	// The capacity gauges are keyed only by (priority, inference_pool), which every harness in this
	// package shares, so the assertions below need a clean slate.
	eppmetrics.Reset()
```

pkg/epp/flowcontrol/integration_test.go
```diff
@@ -1201,6 +1219,37 @@ func TestFlowControlMetricsEmitted(t *testing.T) {
...
+	}, time.Second, time.Millisecond,
+		"band and aggregate utilization ratios should return to 0 after the queue drains")
+}
+
+// globalPriorityLabel is the priority label value carried by the all-bands aggregate series.
+const globalPriorityLabel = ""
+
+// capacityUtilizationGauge returns the request-count capacity utilization ratio recorded for the given
+// priority band, or -1 if no series exists for that band.
+func capacityUtilizationGauge(t *testing.T, priority string) float64 {
```
LukeAVanDrie: Non-blocking nit: I noticed two gaps in the test assertions that are worth closing.

**(1) The `bytes` metric is not asserted.**

The `capacityUtilizationGauge` helper hardcodes the `requests` metric family, meaning the `_bytes` metric is never actually tested. Consider parameterizing the helper by family name.

```suggestion
func capacityUtilizationGauge(t *testing.T, family, priority string) float64 {
```

**(2) The "omit-vs-zero" behavior is not enforced.**

Issue #2102 left open whether unbounded capacities should emit `0` or omit the metric entirely. This PR correctly implements omission, but it lacks a test to enforce it. A future refactor that accidentally starts emitting `0` for an unlimited global capacity would pass CI, but it would silently break downstream alerts that may rely on series absence.

Your helper already returns `-1` if a series is missing, which is the perfect hook. You can lock in the omission behavior by adding a second test harness where `maxRequests` is unset, and asserting `require.Equal(t, -1.0, capacityUtilizationGauge(t, reqFamily, ""))`.

docs/metrics.md
```diff
@@ -271,6 +271,29 @@ Exposed when the `flowControl` feature gate is enabled.
 *   **Usage:** A nonzero value during a dispatch stall indicates a model-server metrics collection
     problem (scrape path, port, TLS, auth) rather than genuine overload.
 
+#### `flow_control_capacity_utilization_requests`
+
+*   **Type:** Gauge
+*   **Labels:** `priority`, `inference_pool`
+*   **Description:** Fraction of the configured request-count capacity currently occupied (0.0-1.0),
+    aggregated over every flow in a priority band. This is not a per-flow-queue metric: `priority`
```
LukeAVanDrie: Same two corrections as the help text.

One more sentence worth adding under **Usage**: when a global cap sits below the sum of the band caps, the per-band ratio understates pressure.

We can keep the band cap as the per-band denominator rather than `min(band, global)` as making a band's ratio depend on other bands' config would be worse IMO.

Let's just make sure this is documented.

pkg/epp/metrics/llm_d_router_metrics.go
```diff
@@ -399,6 +399,24 @@ var (
 		[]string{"detector"},
 	)
 
+	llmdFlowControlCapacityUtilizationRequests = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: LLMDRouterEndpointPickerSubsystem,
+			Name:      "flow_control_capacity_utilization_requests",
+			Help:      metricsutil.HelpMsgWithStability("Fraction of the configured request-count capacity currently occupied, aggregated over every flow in a priority band (0.0-1.0). The aggregate across all bands is reported with priority=\"\" and only when a global request-count capacity is configured; a dimension with no configured capacity is omitted rather than reported as 0.", compbasemetrics.ALPHA),
```
LukeAVanDrie: Two corrections here, which should also be applied to the `_bytes` help text and `docs/metrics.md`:

1. Per-band capacity is never omitted: Because `applyDefaults` always sets a fallback (5000 requests or 1 GB), a band is never unconfigured. Only the global series can be omitted. (The `> 0` guard is fine to keep, but the help text shouldn't imply per-band omission).
2. "Configured capacity" is misleading: Since unconfigured bands fall back to defaults, use "effective capacity" instead. Otherwise, operators might be confused when an unconfigured band emits values against a 5000-request denominator (e.g., emitting `0.0002` for 1 queued request).

```suggestion
			Help:      metricsutil.HelpMsgWithStability("Fraction of a priority band's effective request-count capacity currently occupied, aggregated over every flow in the band (0.0-1.0). The denominator falls back to a default when the band does not configure one. Global capacity is optional and unset by default; the all-bands series is emitted only when it is configured.", compbasemetrics.ALPHA),
```

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -356,6 +356,37 @@ func (p *Processor) hasCapacity(priority int, itemByteSize uint64) (bool, contra
...
+			metrics.RecordFlowControlCapacityUtilizationRequests(priorityStr, p.poolName,
+				float64(band.Len)/float64(band.CapacityRequests))
+		}
+		if band.CapacityBytes > 0 {
+			metrics.RecordFlowControlCapacityUtilizationBytes(priorityStr, p.poolName,
+				float64(band.ByteSize)/float64(band.CapacityBytes))
+		}
+	}
+
+	// Aggregate across all bands, only when a global capacity is configured.
```
LukeAVanDrie: Consider separating the global aggregate into its own metric family.

Currently, this implementation emits the global rollup alongside its per-band components in the same metric family, using a blank label (`priority=""`). Doing this breaks standard PromQL aggregations over the family:

*   Double counting: Aggregations like `sum by (inference_pool) (...)` will double-count the capacity (summing both the individual bands and the global total)
*   Skewed rankings: Functions like `max(...)` and `topk(3, ...)` will incorrectly rank the global rollup against the individual bands
*   UI quirks: Grafana legends keyed by `priority` will render an awkward blank entry for the global series
*   Ambiguous selectors: The query `{priority=""}` is ambiguous because it also matches series where the label is entirely absent, failing to clearly show user intent

Additionally, using `WithLabelValues("", ...)` introduces a new convention; it isn't currently used anywhere in the codebase outside of tests.

To fix this, give the global aggregate its own dedicated metric family with no `priority` label at all.

LukeAVanDrie (changes requested):

LukeAVanDrie:
> @LukeAVanDrie flagging this since #2102 spells out flow_control_queue_utilization_* by name. happy to go back to the issue's name if you'd rather keep that contract.

No, your new naming is much better! Only blocker in my review is  using `priority=""` for the aggregate. We should move it to its own metric family. That also dissolves your objection to a `band_` prefix, since there is no `""` row left once it moves. I would still skip the prefix, because the `priority` label already carries it.

sudoalok:
done, all five.

moved the aggregate to its own family, flow_control_global_capacity_utilization_requests/_bytes, just {inference_pool}. help text and docs say effective capacity now and i took out the per band omission bit since applyDefaults always gives the band a denominator. added the note about a global cap sitting under the sum of the band caps.

tests have Reset() now, helper takes the family so bytes is covered too, and theres a new test for the absence vs zero case. the bytes assertion failed first time actually, harness only bounded maxRequests so the bytes rollup genuinely wasnt there and my assertion was wrong. bounded both now.

LukeAVanDrie (approved):

ahg-g (approved):

Issue #1880, opened 2026-07-02
feat: configurable per-source scraping interval in the data layer


noalimoy:
### Problem

All datalayer polling sources share a single ~50ms `Collector` tick
(`defaultRefreshInterval` in `runtime.go`). Every registered
`PollingDispatcher` is invoked on every tick, regardless of how
frequently its underlying data actually changes.

This is appropriate for latency-sensitive model-server metrics (KV cache,
queue depth) but excessive for sources whose data changes on a coarser
timescale. For example, `/v1/models` returns loaded model names — data
that changes infrequently — yet is scraped 20 times per second per
endpoint.

As new data sources are added (e.g. hardware-level metrics from DCGM
Exporter in #1706, where the driver updates values every ~1s), the
mismatch grows: scraping faster than the source's own update rate
returns duplicate values and adds unnecessary HTTP overhead. At 100
endpoints, a 50ms poll against a 1s source means ~1900 wasted requests
per second.

Ref: discussion on #1706.

### Proposed approaches

**A. Source-level throttle (`minInterval` in `HTTPDataSource`)**

Each source maintains a `lastPollTime`. `Dispatch()` returns early if
`minInterval` has not elapsed since the last successful poll. The
`Collector` tick and overall architecture remain unchanged.

```yaml
plugins:
  - type: metrics-data-source
    parameters:
      minInterval: 50ms   # explicit, same as current default
  - type: dcgm-data-source
    parameters:
      minInterval: 1s
```

- Minimal change: ~15 lines in `HTTPDataSource`, new config parameter
- Fully backward-compatible (default `0` = current behavior)
- Downside: `Collector` still wakes every 50ms and calls `Dispatch()`
  on every source, even when most will no-op

**B. Per-source scheduling in the data layer**

Each source declares its own interval. The `Collector` (or `Runtime`)
maintains independent timers and only invokes a source when its interval
fires.

```yaml
dataLayer:
  sources:
    - pluginRef: metrics-data-source
      interval: 50ms
    - pluginRef: dcgm-data-source
      interval: 1s
```

- Cleaner architecture: no wasted dispatch calls
- Larger change: touches `Collector`, `Runtime`, config schema
- More flexible long-term (each source defines its own cadence)

### Out of scope

- Changing the default 50ms interval for `metrics-data-source` (separate
  discussion if needed)
- Adaptive/dynamic intervals based on rate of change

elevran:
some additional constraints set for the design
[...]

PR #2195, opened 2026-07-27
Increase default refresh-metrics-interval from 50ms to 60s

Lucas-Fernandes-Martins:
**What type of PR is this?**

The 50ms default causes excessive memory allocations from the Prometheus TextParser on every scrape cycle, leading to OOM under production workloads. By default, none of the plugins recommended in the optimized baseline use the scraped metrics, with the expection of `KVCacheBlockSize`, which is constant. Therefore, it makes sense to increase the polling interval to 60s, to prevent memory strains.

This is coupled with [this](https://github.com/llm-d/llm-d/pull/2118) PR in llm-d to update documentation.

TBD

```release-note
NONE
```

### DIFF
--- a/pkg/epp/datalayer/runtime.go
+++ b/pkg/epp/datalayer/runtime.go
@@ -62,7 +62,7 @@ type Runtime struct {
 }
 
 const (
-	defaultRefreshInterval = 50 * time.Millisecond
+	defaultRefreshInterval = 60 * time.Second
 )
 
 // NewRuntime creates a new Runtime with the given polling interval.
--- a/pkg/epp/server/options.go
+++ b/pkg/epp/server/options.go
@@ -129,7 +129,7 @@ func NewOptions() *Options {
 		PoolGroup:                        routing.InferencePoolAPIGroup,
 		EndpointTargetPorts:              []int{},
 		DisableEndpointSubsetFilter:      false,
-		RefreshMetricsInterval:           50 * time.Millisecond,
+		RefreshMetricsInterval:           60 * time.Second,
 		RefreshPrometheusMetricsInterval: 5 * time.Second,
 		MetricsStalenessThreshold:        2 * time.Second,
 		TotalQueuedRequestsMetric:        "vllm:num_requests_waiting",


elevran:
@Lucas-Fernandes-Martins understand the issue and motivation. Exploring some alternatives below:

The simplest workaround is to use the `--refresh-metrics-interval` CLI flag, which you can set to 60s to reduce the scraping interval. A few things to note:
- it is a global flag so affects all scrapers (e.g., v1/models, DCGM, ...)
- there is a separate `--metrics-staleness-threshold` (default 2s) for when scraped data is treated as stale. Might want to change that as well

If that does not work there are some other options we can consider , each as a different PR scope:
1. Investigate reducing allocations and OOM in the Prometheus TextParser (e.g., by reusing the parsers, IIRC there's a `Clear()` method somewhere to allow clean reuse). This couold potentially solve OOM but is still excessive work for no value.
1. Introduce per source polling interval in parameters (in addition to scheme, path, tls config, ...). The `PollingDispatcher` interface would need an `Interval() time.Duration` method, and the main scraping loop in the data layer runtime would need to either create one ticker per dispatcher or use some form of a timing wheel (e.g., high enough base frequency of ticks and call into specific collectors on multiple of the base frequency according to their interval).
1. Introduce a (new) class of data sources that only need fire once successfully to collect static information. IIRC, vllm introduced a `/config endpoint that can be read once (not sure of sglang has the same option). We can then remove the static KVBlockSize from the metrics scraping (it does not really belong there since it is fixed configuration, not a runtime metric).

Also note:
- failing tests on PR
- missing release note section in PR description (inclreasing the default scraping is user facing change)

elevran:
@Lucas-Fernandes-Martins xref #2214 (addressing #1880)

elevran:
@Lucas-Fernandes-Martins assuming #2214 merges soon, do you still see a need for this PR?
You should be able to set the metrics source scarping interval as high as desired in the configuration file

Lucas-Fernandes-Martins:
Thank you @elevran that makes complete sense #2214 does seem to address it perfectly! I'll close this PR and #2118 in llm-d as well

PR #2197, opened 2026-07-27
fix(EPP): stream base64 image decoding to reduce memory

### LINKED ISSUE #2196: EPP OOM under high multimodal load
**What happened**:

Under high multimodal load, EPP can face OOM issues. 

I believe there's a low hanging fruit optimization which might help: currently, `imageDimensionsFromBase64Payload` decodes the entire image just to retrieve its dimensions. This can be avoided by 'lazily` decoding the image using `base64.NewDecoder(base64.StdEncoding, strings.NewReader(rawB64))` in this two-line [PR](https://github.com/llm-d/llm-d-router/pull/2197).

No OOM (under reasonable constraints, of course given infinite load EPP would OOM).

Fire a few hundred large concurrent multimodal requests at the EPP. For ease of reproduction, one can decrease the default memory of the EPP (to 16 GiB for instance) and thus require less load to reach OOM.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.36.1
- llm-d-router version (use `git describe --tags --dirty --always` if you built from source, or specify the tag if you used a tagged version or image): gaie-1.5-sync-2031-g364c2415
- Cloud provider or hardware configuration: not applicable
- Install tools: not applicable
- Others:

Lucas-Fernandes-Martins:
**What type of PR is this?**
/kind bug

Under high multimodal load, EPP can face OOM, due to holding multiple decoded image buffers for concurrent requests. For this function, we only care about the image sizes, so we can safely 'lazily' decode the image, get only the header we're interested in, and stop right there, avoiding decoding the entire image. 
Use base64.NewDecoder streaming instead of DecodeString, so image.DecodeConfig only reads the header bytes. Avoids allocating the full decoded image payload just to read dimensions.

Fixes #2196 

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/mm_estimate.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/mm_estimate.go
@@ -157,7 +157,7 @@ func imageDimensionsFromBase64(url string) (width, height int, ok bool) {
 // returns its pixel dimensions.
 func imageDimensionsFromBase64Payload(rawB64 string) (width, height int, ok bool) {
 	// Image decoding is streamed to reduce memory overhead, since
-	// only headers are necessary
+	// only headers are necessary.
 	r := base64.NewDecoder(base64.StdEncoding, strings.NewReader(rawB64))
 	cfg, _, err := image.DecodeConfig(r)
 	if err != nil || cfg.Width <= 0 || cfg.Height <= 0 {
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/mm_estimate.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/mm_estimate.go
@@ -17,7 +17,6 @@ limitations under the License.
 package tokenizer
 
 import (
-	"bytes"
 	"encoding/base64"
 	"image"
 	"strings"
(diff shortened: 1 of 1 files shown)


capri-xiyue:
/lgtm

@Lucas-Fernandes-Martins Thank you for the contribution! Could you fix the CI/CD in this PR so that we can merge it?

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/mm_estimate.go
```diff
@@ -157,11 +156,8 @@ func imageDimensionsFromBase64(url string) (width, height int, ok bool) {
 // imageDimensionsFromBase64Payload decodes a bare base64 image payload and
 // returns its pixel dimensions.
 func imageDimensionsFromBase64Payload(rawB64 string) (width, height int, ok bool) {
-	decoded, err := base64.StdEncoding.DecodeString(rawB64)
-	if err != nil {
-		return 0, 0, false
-	}
-	cfg, _, err := image.DecodeConfig(bytes.NewReader(decoded))
+	r := base64.NewDecoder(base64.StdEncoding, strings.NewReader(rawB64))
```
capri-xiyue, line 161: nits: Could you add a comment indicating we need to stream base64 image decoding here to reduce memory ?

Lucas-Fernandes-Martins:
Thank you @capri-xiyue for the review! I've added a comment and fixed the CI!

ahg-g (approved):

PR #2203, opened 2026-07-27
[CI][Security]: Run gosec via golangci-lint in the lint job

hickeyma:
**What type of PR is this?**

/kind feature

There are no CI scans for llm-d-router code today. `govulncheck` covers dependency CVEs and Trivy covers image layers but both look at code we pull in rather than code written. 

This PR adds `gosec` which finds insecure Go patterns: weak crypto, permissive file modes, unchecked integer conversions, SQL built by concatenation.

It runs through `golangci-lint` rather than as a standalone job or the upstream Docker action for the following reasons:

- Toolchain correctness. The `securego/gosec` container ships `GOTOOLCHAIN=local` with a Go older than the project `go.mod` requires. This means every package load fails and the scan silently reports nothing. Running inside the builder container (`golang:1.26.5`) removes that class of bug
- Existing suppressions. The repo already carries 6 `//nolint:gosec` directives. `golangci-lint` honours them. The standalone gosec ignores them and re-reports sites the team has already reviewed
- Fork PRs. Nearly all contributions to the project are from forks which get a read only token that cannot upload SARIF. A lint step gates the PR regardless because blocking a merge needs no write token
- Single toolchain version. The builder image pins `golangci-lint`, so there is no second copy to drift

The trade-off of doing it this way is that `golangci-lint `sets the SARIF `ruleId` to the linter name. This means all findings group under `gosec` rather than per rule (G115, G402, …) and rule metadata such as CWE mappings is not emitted. The G-code is still present in each finding's message text and alerts remain per location. It also bundles `gosec` `v2.23.0,` so rules newer than that are unavailable until `golangci-lint` is upgraded.

*Baseline and gating:*

The `gosec` scan is advisory to start with `SECURITY_LINT_EXIT_CODE` defaults to `0`. This means the findings are reported without failing the job. The current baseline is 50 findings. Once those are triaged and fixed or suppressed, setting the variable to 1 turns this into a merge gate.

*Secret scanning and CodeQL are handled natively:*

An earlier revision of this PR added a separate `ci-security.yaml` workflow carrying Gitleaks and CodeQL jobs. Both are dropped in favour of GitHub's repo level features which provide similar capability without a workflow to maintain:

- **Secret scanning.** This is a public repo, so provider patterns, generic patterns (private keys, connection strings) and push protection are available at no cost. Push protection is the stronger control because it rejects the secret at push time instead of reporting it once it is already in history. A full history `gitleaks` run over 2571 commits found nothing, so there is no backlog a CI job would surface
- **CodeQL.** To be enabled through the repository's own code scanning settings which needs admin access rather than a workflow file. Thanks @elevran for offering to set this up

Part of #956 (Phase 3, SAST and secret scanning).

/cc @elevran 

**Verification**:

- [x] `gosec`only `golangci-lint` config produces valid SARIF with repo-relative paths
- [x] `make lint-security` expands correctly and appears in `make help`
- [x] `actionlint` clean on `ci-lint.yaml`
- [ ] `make presubmit`

```release-note
NONE
```


.github/workflows/ci-security.yaml
```diff
@@ -0,0 +1,109 @@
...
+        uses: securego/gosec@9e75c0576c9878035d4221392108d458abe10fc3 # v2.28.0
+        with:
+          args: -no-fail -fmt sarif -out gosec.sarif ./...
+
+      # Fork PRs run with a read only GITHUB_TOKEN that cannot upload SARIF.
+      # The scan above still surfaces findings in the job log for those runs.
+      - name: Upload Gosec SARIF
+        if: ${{ !cancelled() && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) }}
+        uses: github/codeql-action/upload-sarif@v4.35.4 # v4.35.4 68bde559dea0fdcac2102bfdf6230c5f70eb485e
+        with:
```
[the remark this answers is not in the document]
hickeyma: Agree done in a0aeedb7de67984586965cd083b8bad5ead90091. Both `upload-sarif` refs now pin @68bde559dea0fdcac2102bfdf6230c5f70eb485e # v4.35.4.

One gotcha if you're verifying it: `refs/tags/v4.35.4` is an annotated tag, so the object ID it returns (3ce22a6e...) isn't a commit. You have to dereference it first.

.github/workflows/ci-security.yaml
```diff
@@ -0,0 +1,109 @@
...
+permissions: read-all
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  gosec:
+    if: github.event_name != 'schedule'
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
hickeyma: Good catch on the mismatch but it's the description that's wrong and not the condition. I've fixed the wording instead.

Tightening these to PR only would break the workflow. This PR is itself from a fork and you can see both SARIF uploads skipped in the run because fork PRs get a read only GITHUB_TOKEN that can't write to the Security tab. Most contributions  here come from forks, so pushes to main are the only thing that ever populates code scanning.

.github/workflows/ci-security.yaml
```diff
@@ -0,0 +1,109 @@
...
+        uses: actions/setup-go@v7
+        with:
+          go-version-file: go.mod
+
+      # Go has no `none` build mode, so the database is only populated if the
+      # source is compiled between init and analyze.
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v4.35.4 # v4.35.4 68bde559dea0fdcac2102bfdf6230c5f70eb485e
+        with:
+          languages: go
```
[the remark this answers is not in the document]
hickeyma: Done in a0aeedb7de67984586965cd083b8bad5ead90091, now pinned to @68bde559dea0fdcac2102bfdf6230c5f70eb485e # v4.35.4.

.github/workflows/ci-security.yaml
```diff
@@ -0,0 +1,109 @@
...
+          build-mode: autobuild
+
+      - name: Perform CodeQL analysis
+        uses: github/codeql-action/analyze@v4.35.4 # v4.35.4 68bde559dea0fdcac2102bfdf6230c5f70eb485e
+        with:
+          category: /language:go
```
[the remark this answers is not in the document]
hickeyma: Same here. That's all four codeql action refs on @68bde559dea0fdcac2102bfdf6230c5f70eb485e # v4.35.4 now.

.github/workflows/ci-security.yaml
```diff
@@ -0,0 +1,128 @@
...
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      security-events: write
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v7
+
+      - name: Run Gosec
+        uses: securego/gosec@9e75c0576c9878035d4221392108d458abe10fc3 # v2.28.0
```
elevran: Q: would it make sense to have this run as part of golangci linting by adding `gosec` to the list of linter?
Instead of an explicit step. That would fail merging of insecure code (but without the sarif report, which IIUC is used to collect findings and would be empty?)

.github/workflows/ci-security.yaml
```diff
@@ -0,0 +1,128 @@
...
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@v7
+        with:
+          name: gosec-sarif
+          path: gosec.sarif
+          retention-days: 7
+          if-no-files-found: warn
+
+  gitleaks:
+    if: github.event_name != 'schedule'
```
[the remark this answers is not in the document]
hickeyma: Fair point and it makes the file internally consistent. The codeql job already uses an explicit allowlist while these two used the negation. Switched both gosec and gitleaks to be similar.

Behavior is unchanged as that's the exact set the negation resolved to. No ref guard on push because the `on:` block already restricts it to `main`.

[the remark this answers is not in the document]
hickeyma: Changed in f9a4c1558f3975ccf3d71747e5662bbac1f693f6

.github/workflows/ci-security.yaml
hickeyma: Good question @elevran. Digging into it turned up something interesting. The gosec job isn't currently finding anything because it isn't successfully analysing anything. The `securego/gosec` image ships `GOLANG_VERSION=1.26.3` with `GOTOOLCHAIN=local` baked in and the project `go.mod` requires `1.26.5`. Therefore, every package load fails instantly. The run enumerates all 203 package dirs in very quickly and writes an empty SARIF while `-no-fail` keeps it green. The codeql job avoids this because it runs `setup-go` with `go-version-file: go.mod` first.

So on your specific point. The SARIF is already empty, just not for the reason either of us thought.

Yes, moving `gosec` into `golangci-lint` has real advantages. It'd run in the builder container on the right toolchain and it would gate the PR rather than reporting after merge. Another good reason is that contributions to the project come from forks and fork PRs get a read only token that can't upload SARIF. This means that today the Security tab only ever gets written on push to main. A lint gate works on forks.

The trade off is SARIF granularity. `golangci-lint` `v2` can emit SARIF but as one blob for all linters rather than a separate `gosec` category in the Security tab which was the reason for the standalone job.

Suggest we split it up. First, I'll fix the toolchain so we can see the actual baseline (using the same `setup-go` + `go install` pattern this workflow already uses for gitleaks). We can then decide on `golangci-lint` with real numbers.  What do you think?

elevran: would it be possible to create a second golangci-lint config file (e.g., with `-security` suffix) with just the relevant security linters (possibly only gosec) and a SARIF output?
We can run a golangci-lint GH action with the secondary configuration as a separate step in the lint check to get just the relevant linter output.

elevran:
> The gitleaks CLI is used rather than the action which needs a paid licence for org-owned repos and is no longer MIT

- the action requires a license from [gitleaks.io](https://gitleaks.io/). That site says
    ```
    Free Organization License Key
    Fill out Google Form to recieve an organization license.
    ```
    AFAICT, it's free to use on organization repositories, but requires a free license key from the Gitleaks Official Site when using the official automated GitHub Action.
- the `gitleaks` author shifted focus to [betterleaks](https://github.com/betterleaks/betterleaks). There is an [action for betterleaks](https://github.com/dortort/betterleaks-action) , though not heavily used like the `gitleaks` action.
- is the native GH secret scanning (repo level setting) insufficient in your view?

elevran:
PS - can run the repo's native GH tooling to enable a separate codeql workflow with advanced settings, if you think that would be useful.

.github/workflows/ci-security.yaml
hickeyma: Done in 4c39b691b4d40e08309613b12b4da20b487c35e6 and it seems better than the standalone job. `.golangci-security.yml` is now just gosec and SARIF output, like you suggested.

One difference to your suggestion is that it is run through a `make` `lint-security` target in the builder container rather than the GH action. The action unfortunately pulls its own `golangci-lint`, so we'd end up with two versions free to drift from the Go pin. That drift is what  caused the empty scans to begin with and would want to avoid this. Contributors get the same scan locally this way too.

Your idea fixed something I'd missed. There is already 6 `//nolint:gosec` directives in the tree. `golangci-lint` respects them while standalone `gosec` ignores them and keeps flagging things already signed off.

There are however 2 thigs to be aware of:

- SARIF `ruleId` becomes the linter name, so findings group under `gosec` instead of per rule and CWE data is dropped. The G-code is still in the message and alerts are still per location, so more a downgrade than a loss
- It also bundles `gosec` `v2.23.0` against the `v2.28.0` I'd pinned, so we lose the newer rules. One of those issues found is the metrics server in `runner.go` has no `ReadHeaderTimeout`. This can be fixed separately since this scan won't flag it

Note that `gosec` is advisory for now with `SECURITY_LINT_EXIT_CODE is 0` so we can triage the baseline before it blocks merges.

hickeyma:
Thanks @elevran , both good catches in https://github.com/llm-d/llm-d-router/pull/2203#issuecomment-5104223820. 

Re the licence., you're right and it's free rather than paid. The action's README says you "will need to obtain a free license key" for org-owned repos, so my description is wrong. The part that does hold is that the action has been NOASSERTION since `v2.0.0` while the CLI stayed MIT which is why the CLI installed.  Thanks for the heads up on betterleaks.

I agree with your suggestion on using native secret scanning instead. It seems sufficient and I am ok with dropping the gitleaks job. The one real gap with gitleaks is custom patterns which need Secret Protection. If there are llm-d specific credential formats worth matching that is the argument for revisiting. Otherwise native seems to cover it.

Could you confirm secret scanning and push protection are enabled at the repo or org level?

I agree also with your suggestion for CodeQL. I think that it would be useful. The job in this PR has never actually executed. `workflow_dispatch` is not reachable from a fork branch, so `build-mode: autobuild` is unverified and its first real run would be on merge to `main` which is late in the day to fins issues. I am dropping it in favour of whatever you enable.

One suggestion is to use default setup rather than advanced, unless you want the config visible in-repo. It is a settings toggle with no workflow file to maintain and the query pack updates itself and you need neither custom queries nor a custom build. Advanced is the better pick if in-repo visibility matters more. Either way the two conflict for the same language, so which one goes on is your call.

That means then that this PR becomes just the `gosec` change only. 

What do you think?

elevran:
I think we can scope the PR to `gosec` linting and use GH native CodeQL and secret scanning.
Waiting for confirmation from other maintainers before enabling on the repo.

hickeyma:
>  I think we can scope the PR to gosec linting and use GH native CodeQL and secret scanning.

Ok, done in 240a74fbe30d476d0ac5f03717edf922b6e52e84.

> Waiting for confirmation from other maintainers before enabling on the repo.

Thats ok. It can be done separately to the PR.

.github/workflows/ci-lint.yaml
```diff
@@ -103,3 +108,28 @@ jobs:
...
+      - name: Run make lint-security
+        env:
+          GO_MOD_CACHE_VOL: ${{ steps.go-cache.outputs.mod }}
+          GO_BUILD_CACHE_VOL: ${{ steps.go-cache.outputs.build }}
+        run: make lint-security
+
+      # Fork PRs run with a read only GITHUB_TOKEN that cannot upload SARIF.
+      - name: Upload gosec SARIF
+        if: ${{ !cancelled() && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) }}
+        uses: github/codeql-action/upload-sarif@68bde559dea0fdcac2102bfdf6230c5f70eb485e # v4.35.4
```
elevran, line 121: upload-sarif is pinned by full SHA but all other gitub actions in this job use floating tags. 
This aligns ith GitHub's guidance to pin security-sensitive actions by SHA, but the mix is worth a one-line note so it doesn't read as an oversight during future review and any future changes pin to a new SHA and not a version.

Makefile
```diff
@@ -255,6 +255,15 @@ lint: image-build-builder ## Run lint (use LINT_NEW_ONLY=true to only check new
 	@printf "\033[33;1m==== Running linting ====\033[0m\n"
 	$(BUILDER_RUN) 'GOFLAGS=-buildvcs=false golangci-lint run $(LINT_ARGS) && typos'
 
+# Reports findings without failing while the initial baseline is triaged.
+# Set to 1 to block merges on new findings.
+SECURITY_LINT_EXIT_CODE ?= 0
```
elevran, line 260: because SECURITY_LINT_EXIT_CODE defaults to 0, the new job can never fail CI, so nothing currently forces the finding baseline to actually get triaged. 
Can you please open a follow up issue for triaging, fixing and ultimately flipping the default value of SECURITY_LINT_EXIT_CODE ?

elevran (approved):
approve - no blockers.
Can merge as is or with clarification of SHA pinning on action.
/hold in case you would like to address.
Would appreciate a follow up issue for triaging and fixing findings.

elevran:
Secret scanning and CodeQL (common setup) eanbled at repo level.

Note: while it's possible to post-process `gosec.sarif` after `make lint-security` runs and rewrite ruleId before upload using a short script step (gosec's messages start with the G code, e.g. "G115: ...") , this adds a moving part to maintain. Deferred.

elevran:
/hold cancel
merge as-is, add note on SHA use in follow up

Makefile
hickeyma: Thanks @elevran for covering this in #2267

.github/workflows/ci-lint.yaml
hickeyma: Note added in #2269

hickeyma:
Thanks @elevran for the reviews and merging. I responded to the open issues in https://github.com/llm-d/llm-d-router/pull/2203#discussion_r3682518935 and https://github.com/llm-d/llm-d-router/pull/2203#discussion_r3711579870

PR #2204, opened 2026-07-27
feat(scheduling): add endpoint attribute weight scorer

### LINKED ISSUE #2201: feat: add a generic endpoint-label-weight-scorer for static routing preferences

A generic `endpoint-label-weight-scorer` that maps a configured endpoint label's value to a static, configurable score.

**Why is this needed**:

Production InferencePools commonly mix GPU types (e.g. H100/A100/L40S) with different throughput. There's currently no way to score endpoints by a discrete label value, so operators either exclude slower GPUs entirely via `bylabel`, or rely on dynamic running-load scorers alone — which under load tends to concentrate traffic on faster endpoints and can starve slower-but-still-useful endpoints of traffic.

This scorer is meant to compose with existing dynamic scorers (e.g. running-request/queue-depth via `endpoint-attribute-scorer`) through SchedulingProfile weights, so overall throughput can be optimized while keeping weaker endpoints in rotation instead of idle.

Related:
- #1868 / #1873: static accelerator capability — matches request size against a numeric label *range*, different mechanism from mapping a discrete label *value* to a weight.
- #2059: generic numeric endpoint-attribute scorer/filter — this composes with it (dynamic side), doesn't overlap (static side).

nicole-lihui:
**What type of PR is this?**

/kind feature

Add `label-producer` and `endpoint-attribute-weight-scorer`.

The producer exposes Pod labels as endpoint attributes, and the scorer maps discrete attribute values to normalized routing scores. This supports static preferences for heterogeneous GPUs, zones, racks, and similar endpoint groups while keeping missing or unknown values eligible.

Fixes #2201

**Test plan**:

- [x] Unit tests cover configuration validation, score normalization, fallback behavior, and scorer category.
- [x] Hermetic integration tests verify weighted GPU routing and eligibility of missing or unknown attributes.

```release-note
Add endpoint attribute weight scoring for configurable static routing preferences.
```

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/extractor/label/producer.go
+++ b/pkg/epp/framework/plugins/datalayer/extractor/label/producer.go
@@ -1,0 +1,148 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package label provides an endpoint extractor that copies a Pod label into
+// an endpoint attribute.
+package label
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+	fwkplugin "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+	attrmetrics "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/attribute/metrics"
+	attrstring "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/attribute/string"
+	sourcenotifications "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/source/notifications"
+)
+
+const (
+	// LabelProducerType identifies the label producer plugin.
+	LabelProducerType = "label-producer"
+
+	valueTypeString = "string"
+	valueTypeNumber = "number"
+)
+
+var (
+	_ fwkplugin.Plugin        = (*Producer)(nil)
+	_ fwkdl.Registrant        = (*Producer)(nil)
+	_ fwkdl.EndpointExtractor = (*Producer)(nil)
+)
+
+type parameters struct {
+	// Label is the Pod label to copy.
+	Label string `json:"label"`
+	// AttributeKey is the endpoint attribute key to populate.
+	AttributeKey string `json:"attributeKey"`
+	// ValueType controls whether the value is stored as a string or number.
+	ValueType string `json:"valueType"`
+}
+
+// Factory creates a label producer.
+func Factory(name string, decoder *json.Decoder, _ fwkplugin.Handle) (fwkplugin.Plugin, error) {
+	var params parameters
+	if decoder != nil {
+		if err := decoder.Decode(&params); err != nil {
+			return nil, fmt.Errorf("failed to parse the parameters of the '%s' plugin: %w", LabelProducerType, err)
+		}
+	}
+	return NewProducer(name, params)
+}
+
+// NewProducer validates parameters and creates a label producer.
+func NewProducer(name string, params parameters) (*Producer, error) {
+	if name == "" {
+		name = LabelProducerType
+	}
+	if params.Label == "" {
+		return nil, fmt.Errorf("'%s' requires a non-empty 'label'", LabelProducerType)
+	}
+	if params.AttributeKey == "" {
+		return nil, fmt.Errorf("'%s' requires a non-empty 'attributeKey'", LabelProducerType)
+	}
+	if params.ValueType == "" {
+		params.ValueType = valueTypeString
+	}
+	if params.ValueType != valueTypeString && params.ValueType != valueTypeNumber {
+		return nil, fmt.Errorf("'%s' valueType must be %q or %q, got %q", LabelProducerType, valueTypeString, valueTypeNumber, params.ValueType)
+	}
+
+	return &Producer{
+		typedName:    fwkplugin.TypedName{Type: LabelProducerType, Name: name},
+		label:        params.Label,
+		attributeKey: params.AttributeKey,
+		valueType:    params.ValueType,
+	}, nil
+}
+
+// Producer copies a Pod label into an endpoint attribute.
+type Producer struct {
+	typedName    fwkplugin.TypedName
+	label        string
+	attributeKey string
+	valueType    string
+}
+
+// TypedName returns the type and name of the producer.
+func (p *Producer) TypedName() fwkplugin.TypedName {
+	return p.typedName
+}
+
+// RegisterDependencies wires the producer to endpoint lifecycle events.
+func (p *Producer) RegisterDependencies(r fwkdl.Registrar) error {
+	return r.Register(fwkdl.PendingRegistration{
+		Owner:      p.typedName,
+		SourceType: sourcenotifications.EndpointNotificationSourceType,
+		Extractor:  p,
+		DefaultSource: sourcenotifications.NewEndpointDataSource(
+			sourcenotifications.EndpointNotificationSourceType,
+			sourcenotifications.EndpointNotificationSourceType,
+		),
+	})
+}
+
+// Extract copies the configured label into the endpoint attribute.
+func (p *Producer) Extract(_ context.Context, event fwkdl.EndpointEvent) error {
+	if event.Type == fwkdl.EventDelete || event.Endpoint == nil {
+		return nil
+	}
+
+	meta := event.Endpoint.GetMetadata()
+	if meta == nil {
+		return nil
+	}
+	value := meta.Labels[p.label]
+
+	if p.valueType == valueTypeString {
+		event.Endpoint.GetAttributes().Put(p.attributeKey, attrstring.Value(value))
+		return nil
+	}
+
+	if value == "" {
+		event.Endpoint.GetAttributes().Put(p.attributeKey, attrmetrics.ScalarMetricValue(0))
+		return nil
+	}
+	number, err := strconv.ParseFloat(value, 64)
+	if err != nil {
+		event.Endpoint.GetAttributes().Put(p.attributeKey, attrmetrics.ScalarMetricValue(0))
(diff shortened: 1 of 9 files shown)


elevran:
thanks @nicole-lihui !
Q: given that labels are already part of the Endpoint metadata state, what is the added value of making labels into Endpoint attributes? It can probably result in more focused/granular plugins, but I wonder what other reasons might be.

nicole-lihui:
> thanks @nicole-lihui ! Q: given that labels are already part of the Endpoint metadata state, what is the added value of making labels into Endpoint attributes? It can probably result in more focused/granular plugins, but I wonder what other reasons might be.

@elevran Good question. This actually traces back to #2201 — @shmuelk   suggested decoupling the scorer from `labels` directly, wrapping them as `attributes` instead, I think it's very reasonable.

And You're right that today everything's coming from Pod `labels` (GPU type, node/zone, topology tier), so yeah, it's already sitting on Endpoint metadata. 

But I've been in a Slack thread(https://llm-d.slack.com/archives/C08T1E128PK/p1784565094833379) lately about network-aware P/D scoring, and it made the attribute layer click for a different reason — down the line these could come from something dynamic (topology probing, runtime metrics) instead of just a static label producer. 
If the scorer reads "attributes" instead of "labels" directly, that swap doesn't touch the scorer at all.

So right now it's just labels under the hood, but the indirection is there so we're not boxed into static-only later.

elevran:
@nicole-lihui can you please rebase/resolve conflicts?

elevran:
1. Note: the PR predates a later refactor (attribute keys moved from string to a DataKey type in PR #2190) and was never rebased. It likely doesn't compile against current main :-(. Several uses of string instead of DataKey are in the code.
2.  Neither Producer nor EndpointAttributeWeightScorer implements ProducerPlugin.Produces() / ConsumerPlugin.Consumes(). Without these, the data-dependency DAG ordering added in #2359 has no guarantee this producer runs before this scorer, and the scorer can't be auto-provisioned as a default producer dependency.

Haven't reviewed the code in detail given (1) and (2) above which should be resolved first (along with the conflict)

nicole-lihui:
@elevran  I've resolved the conflicts. 

The README includes a [configuration example combining region and GPU preferences](https://github.com/llm-d/llm-d-router/blob/04fc442aa9d551534907cabdead56a8b97af8f4a/pkg/epp/framework/plugins/scheduling/scorer/attributeweight/README.md#region-and-gpu-preference). I tested the configuration locally with one simulated endpoint per region/GPU combination. 
The combined score is `3 * region_score + gpu_score`, with each dimension normalized to `1` or `0.5`.

Across 20,000 selections, `weighted-random-picker` produced the following distribution:

| Endpoint labels | Combined score | Expected share | Selections | Observed share |
| --- | ---: | ---: | ---: | ---: |
| region-1 / H100 | 4 | 33.33% | 6,688 | 33.44% |
| region-1 / A100 | 3.5 | 29.17% | 5,830 | 29.15% |
| region-2 / H100 | 2.5 | 20.83% | 4,130 | 20.65% |
| region-2 / A100 | 2 | 16.67% | 3,352 | 16.76% |

Using the same scores, `max-score-picker` selected the highest-scoring endpoint, `region-1/H100`, in **100/100** runs.

PTAL, thanks!

pkg/epp/framework/plugins/scheduling/scorer/attributeweight/attributeweight.go
```diff
@@ -0,0 +1,127 @@
...
+const EndpointAttributeWeightScorerType = "endpoint-attribute-weight-scorer"
+
+var (
+	_ fwksched.Scorer          = (*EndpointAttributeWeightScorer)(nil)
+	_ fwkplugin.ConsumerPlugin = (*EndpointAttributeWeightScorer)(nil)
+)
+
+type parameters struct {
+	AttributeKey string             `json:"attributeKey"`
+	Producer     *string            `json:"producer"`
```
elevran: Suggestion/request/nit:
The `Producer *string` value is a pointer even though it's mandatory (e.g., nil is rejected below). The reason is that a plain `string` can't distinguish "omitted" from an explicit `""`, which selects the empty producer namespace. `endpointattribute.go`'s `Producer` field uses the identical trick with a comment explaining why - worth copying that comment here for completeness.

pkg/epp/framework/plugins/scheduling/scorer/attributeweight/attributeweight.go
```diff
@@ -0,0 +1,127 @@
...
+			return nil, fmt.Errorf("%q weight for value %q underflows to zero when normalized against max weight %v",
+				EndpointAttributeWeightScorerType, value, maxWeight)
+		}
+		scores[value] = score
+		fallbackScore = min(fallbackScore, score)
+	}
+
+	return &EndpointAttributeWeightScorer{
+		typedName:     fwkplugin.TypedName{Type: EndpointAttributeWeightScorerType, Name: name},
+		dataKey:       fwkplugin.NewDataKey(params.AttributeKey, *params.Producer),
```
elevran: This builds the `DataKey` directly (`NewDataKey(params.AttributeKey, *params.Producer)`) instead of reusing `attrmetrics.ResolveConfiguredKey`, which the sibling `endpoint-attribute-scorer` uses to let `producer` be omitted (defaulting) and to reject an attribute name containing `/` with no producer set. Since this scorer's whole purpose is composing with `endpoint-attribute-scorer` on matching `attributeKey`/`producer`, the two now have different config ergonomics and failure modes for the same parameter shape. Worth reusing the helper, or a deliberate note on why not.

pkg/epp/framework/plugins/scheduling/scorer/attributeweight/attributeweight_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+	assert.Equal(t, 1.0, scores[endpoints[0]])
+	assert.Equal(t, 0.5, scores[endpoints[1]])
+	assert.Equal(t, 0.25, scores[endpoints[2]])
+	assert.Equal(t, 0.25, scores[endpoints[3]])
+	assert.Equal(t, 0.25, scores[otherProducer])
+	assert.Equal(t, 0.25, scores[wrongType])
+}
+
+func TestLabelProducerDependencies(t *testing.T) {
```
elevran, line 156: `TestLabelProducerDependencies` only exercises non-empty producer names. The README documents `producer: ""` as selecting the empty namespace, but that path has no test coverage - worth adding a case with `Producer: pointer("")`.

pkg/epp/framework/plugins/datalayer/extractor/label/producer.go
```diff
@@ -0,0 +1,145 @@
...
+	producer, err := NewProducer(name, params)
+	if err != nil {
+		return nil, err
+	}
+	if handle != nil {
+		for _, existing := range handle.GetAllPlugins() {
+			if existing.TypedName().Type == LabelProducerType {
+				return nil, fmt.Errorf("multiple %s instances configured (%s, %s); only one instance is supported",
+					LabelProducerType, existing.TypedName(), producer.TypedName())
+			}
```
elevran, line 64: nit:
`LabelProducerType` is formatted with `%s` here but `%q` at lines 75, 80, and 84 in the same file. Pick one for consistency.

pkg/epp/framework/plugins/scheduling/scorer/attributeweight/attributeweight_test.go
```diff
@@ -0,0 +1,192 @@
...
+func TestValidation(t *testing.T) {
+	for _, test := range []struct {
+		name    string
+		mutate  func(*parameters)
+		wantErr string
+	}{
+		{name: "missing attribute", mutate: func(p *parameters) { p.AttributeKey = "" }, wantErr: "attributeKey"},
+		{name: "missing producer", mutate: func(p *parameters) { p.Producer = nil }, wantErr: "producer"},
+		{name: "missing weights", mutate: func(p *parameters) { p.Weights = nil }, wantErr: "weights"},
+		{name: "zero weight", mutate: func(p *parameters) { p.Weights = map[string]float64{"H100": 0} }, wantErr: "positive"},
```
elevran, line 72: nit:
`TestValidation` only exercises `weight: 0` for the `weight <= 0` branch. A negative weight is a plausible operator sign-error - worth a dedicated case asserting the same "must be finite and positive" error.

pkg/epp/framework/plugins/datalayer/extractor/label/README.md
```diff
@@ -0,0 +1,25 @@
+# Label Producer
+
+**Type:** `label-producer`
+
+Publishes Pod labels as string attributes that follow label updates.
+Enable this Alpha plugin with `--allow-experimental-plugins=true`.
```
elevran, line 6: nit: 
worth a line noting Pod labels are assumed operator/cluster-admin controlled - in a multi-tenant cluster, a less-trusted party could set labels on their own serving pods to bias affinity scoring toward themselves.
If everything is namespace scoped - likely less of a concern.

elevran (commented):
Nice addition, and it follows the existing plugin conventions closely.
Looks good to me from a correctness point of view.

A few consistency and coverage gaps below, plus one open question: 
`label-producer`'s factory hard-enforces a single instance per EPP (scanning `handle.GetAllPlugins()`), which prevents configuring multiple disjoint label sets under different producer names. Is that scoping intentional, or should it allow multiple named instances the way most other producers do?
If intentional - worth a release note, README mention, etc.

nicole-lihui:
@elevran 
This is intentional for (`label-producer`'s factory hard-enforces a single instance per EPP) . 
because `Data layer` deduplicates extractor bindings by source and extractor type. Removing the Factory check alone would allow multiple instances in the configuration, but only one would actually be bound and run. Keeping the explicit validation **avoids this silent misconfiguration**.

Changing the `Data layer` binding semantics is outside the scope of this PR(maybe new issue to fix it).

And I documented the single-instance restriction and clarified that all label mappings should be configured in the same instance.

pkg/epp/framework/plugins/scheduling/scorer/attributeweight/attributeweight.go
nicole-lihui: Both scorers ultimately use `NewDataKey`, but their default-producer semantics are different. `endpoint-attribute-scorer` consumes numeric attributes, so it can default to `core-metrics-extractor`. `endpoint-attribute-weight-scorer` consumes string attributes and has no applicable default producer.

I chose not to adopt `ResolveConfiguredKey` because its omitted-producer behavior would point this scorer to `core-metrics-extractor`, creating an invalid producer/value-type pairing and a misleading dependency that this scorer does not otherwise have. 
When `producer` is explicitly configured, the helper provides no additional behavior over `NewDataKey`, so I kept the direct construction and documented the reason.

elevran: Makes sense, the numeric-vs-string default-producer asymmetry is a real distinction. Thanks for adding the code comment too.

elevran (approved):
Thanks for addressing all the feedback. The empty-producer test, negative-weight test, %q consistency fix, and README updates all look correct. The ResolveConfiguredKey question has a solid rationale, both in your reply and now in a code comment. LGTM.

PR #2209, opened 2026-07-27
feat(epp): add active request filter

kaushikmitr:
Add an active-request-filter scheduling plugin that drops endpoints whose EPP-tracked in-flight request count exceeds a configured maximum. Consumes the same inflight-load-producer attribute as the active-request-scorer.

Note: 
What happens when all endpoint filtered out?

Default (fallbackOnEmpty: false) — the request fails with a 500. The filter returns an empty list, and the scheduler aborts the profile run at [scheduler_profile.go:123-125](vscode-webview://1khdmomatp9kcfcasr0i3djq6duf8absgc5h493qlqbk8va54jkd/pkg/epp/scheduling/scheduler_profile.go#L123-L125) with "no endpoints available for the given request" (code Internal). That maps to HTTP 500 InternalServerError back to the client ([error.go:104-105](vscode-webview://1khdmomatp9kcfcasr0i3djq6duf8absgc5h493qlqbk8va54jkd/pkg/common/error/error.go#L104-L105)). In a multi-profile setup, only that profile fails — its entry in the results map is nil and the profile handler decides what to do; if it's the only/primary profile, the request errors out.

With fallbackOnEmpty: true — the filter steps aside. It returns the original unfiltered candidate list, so the request still routes: downstream scorers (e.g. active-request-scorer, queue-depth) rank the overloaded endpoints and the picker chooses the least-bad one. The filter is effectively a no-op for that cycle.

So the choice is between two intents:

Hard admission cap (false): "if everyone is too busy, reject." Caveat: it surfaces as a 500, not a 429, so clients can't distinguish "saturated, retry later" from a real server error. If backpressure under saturation is the actual goal, the flow-control/saturation-detector layer is the purpose-built mechanism — it queues or sheds with proper semantics rather than failing the scheduling cycle.
Soft preference (true): "avoid busy endpoints when possible, but never fail a request because of it." This is the safer default for most routing setups, which is why the README example sets it.

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/filter/activerequest/filter.go
+++ b/pkg/epp/framework/plugins/scheduling/filter/activerequest/filter.go
@@ -1,0 +1,154 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package activerequest
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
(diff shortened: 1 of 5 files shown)


kaushikmitr:
fixes: https://github.com/llm-d/llm-d-router/issues/2210

ahg-g:
This is great, thanks!

ahg-g (approved):

pkg/epp/framework/plugins/scheduling/filter/activerequest/filter.go
```diff
@@ -0,0 +1,154 @@
...
+// parameters defines the parameters for ActiveRequestFilter.
+type parameters struct {
+	// MaxRequests is the maximum number of in-flight requests an endpoint may
+	// have and still pass the filter. Endpoints with a request count greater
+	// than MaxRequests are dropped.
+	MaxRequests int64 `json:"maxRequests"`
+
+	// FallbackOnEmpty returns the unfiltered candidates when every endpoint
+	// was filtered out, so the request can still be routed somewhere.
+	FallbackOnEmpty bool `json:"fallbackOnEmpty"`
```
ahg-g, line 46: Actually we should default this to true. Use a reference here to distinguish between the case when this is deliberately set to false and not set.

ahg-g: perhaps this wont matter since one needs to set MaxRequests anyway, right?

PR #2213, opened 2026-07-28
[docs]: Define area/* label taxonomy

hickeyma:
**What type of PR is this?**
/kind documentation

Adds `docs/area_taxonomy.md` which lays out ten `area/*` labels and maps each to the directories it covers (for example, `area/scheduling` → `pkg/epp/scheduling`). The goal is to give `/area` directives and a future path based auto labeler
the same set of boundaries to work from instead of each inventing its own.

**Note:**
`area/epp` and `area/dev` already exist. The other eight still need to be created and that needs repo write access I don't have. Could someone with maintainer access run through the list in the doc?

Part of #956 (Part 4, Labeling)

Discussed in https://github.com/llm-d/llm-d-router/pull/2113#issuecomment-5045869989

/cc @elevran

```release-note
NONE
```


docs/area_taxonomy.md
```diff
@@ -0,0 +1,23 @@
...
+| `area/telemetry` | `pkg/telemetry/**`, `pkg/common/observability/**` | Telemetry and observability |
+| `area/dev` | `.github/**`, `hack/**`, `scripts/**`, `build/**`, `Makefile` | Dev tooling and CI/CD |
+| `area/docs` | `docs/**`, top-level `*.md` | Documentation |
+
+A change can span multiple areas. Apply every label that fits rather than picking one.
+
+## Applying a label
+
+- Manually: add an `/area <name>` line to a PR or issue body (see `.github/workflows/pr-kind-label.yaml` and `.github/workflows/issue-kind-label.yaml`).
+- Automatically: path based labeling from this table is tracked separately against [#956](https://github.com/llm-d/llm-d-router/issues/956).
```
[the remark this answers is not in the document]
hickeyma: No, its ok.

docs/area_taxonomy.md
```diff
@@ -0,0 +1,23 @@
+# Area taxonomy
```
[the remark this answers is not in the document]
hickeyma: Fixed.

docs/area_taxonomy.md
```diff
@@ -0,0 +1,23 @@
...
+|---|---|---|
+| `area/epp` | `pkg/epp/**`, `cmd/epp/**` | Endpoint Picker core |
+| `area/scheduling` | `pkg/epp/scheduling/**`, `pkg/epp/requestcontrol/**` | Scheduling and request routing decisions |
+| `area/flowcontrol` | `pkg/epp/flowcontrol/**` | Admission, queuing and flow control |
+| `area/kvcache` | `pkg/kvcache/**`, `pkg/kvevents/**` | KV cache indexing and events |
+| `area/coordinator` | `pkg/coordinator/**`, `cmd/coordinator/**` | Disaggregated inference coordinator |
+| `area/sidecar` | `pkg/sidecar/**`, `cmd/pd-sidecar/**` | PD sidecar proxy |
+| `area/datalayer` | `pkg/epp/datalayer/**`, `pkg/epp/datastore/**` | EPP data layer and datastore |
+| `area/telemetry` | `pkg/telemetry/**`, `pkg/common/observability/**` | Telemetry and observability |
+| `area/dev` | `.github/**`, `hack/**`, `scripts/**`, `build/**`, `Makefile` | Dev tooling and CI/CD |
```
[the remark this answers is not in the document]
hickeyma: Good catch

docs/area_taxonomy.md
```diff
@@ -0,0 +1,23 @@
+# Area taxonomy
+
+`area/*` labels route issues and pull requests to the right reviewers by component. Each area maps to one or more directories, so it can be applied either manually (via an `/area` directive on a PR or issue body) or automatically from the paths a change touches.
```
elevran, line 3: Q: when doing path based matching, would something under pkg/epp/scheduling pick up both `area/epp` and `area/scheduling`?

docs/area_taxonomy.md
```diff
@@ -0,0 +1,23 @@
...
+
+| Label | Paths | Scope |
+|---|---|---|
+| `area/epp` | `pkg/epp/**`, `cmd/epp/**` | Endpoint Picker core |
+| `area/scheduling` | `pkg/epp/scheduling/**`, `pkg/epp/requestcontrol/**` | Scheduling and request routing decisions |
+| `area/flowcontrol` | `pkg/epp/flowcontrol/**` | Admission, queuing and flow control |
+| `area/kvcache` | `pkg/kvcache/**`, `pkg/kvevents/**` | KV cache indexing and events |
+| `area/coordinator` | `pkg/coordinator/**`, `cmd/coordinator/**` | Disaggregated inference coordinator |
```
elevran: consider adding `test/` paths to relevant areas (e.g., test/coordinator should map to `area/coordinator` as well)

elevran (commented):
thanks @hickeyma this is a good initial breakdown.
Once this mapping is automated via `.github/labeler.yml`, do you think we should keep both this doc and the labeler yaml? Prefer to have a single source of truth and not keeping the two in sync.

hickeyma:
> Once this mapping is automated via .github/labeler.yml, do you think we should keep both this doc and the labeler yaml? Prefer to have a single source of truth and not keeping the two in sync.

@elevran I think the single source of truth is a good idea. When `.github/labeler.yml` lands it can becomes the authority for the path globs with the scope descriptions moved into comments there and this doc drops the table. What stays in the doc is what the YAML cannot express. Things like the `/area` directive path handled by `pr-kind-label.yaml`and the overlap rule below. I would like to land the table now since 8 of these labels still need creating and the boundaries are the thing to agree on first. The labeler PR will do the consolidation.

docs/area_taxonomy.md
hickeyma: Yes, both. Labeler evaluates each label's globs independently, so `pkg/epp/scheduling/**` matches `area/epp` and `area/scheduling`. Nested areas are additive, not exclusive. Making `area/epp` exclusive needs negated globs and all-globs-to-all-files semantics would then drop `area/epp` from any PR that touches both a sub-area and EPP core. I think this would be worse than the overlap. I stated the additive rule explicitly in the doc in f8b4716a4c6b73b0e9308917243235fe83d04588.

docs/area_taxonomy.md
hickeyma: Good point, I didn't cover `test/`. Added `test/coordinator/**` -> `area/coordinator`, `test/sidecar/**` -> `area/sidecar`, `test/integration/epp/**` -> `area/epp`in f8b4716a4c6b73b0e9308917243235fe83d04588. For the shared harness (`test/e2e`, `test/framework`, `test/utils`, `test/scripts`, `test/perf`, `test/profiling`) there is no single component owner, so I mapped those to 
`area/dev` alongside `hack/` and `scripts/`. I let the existing `kind/testing` carry the "this is a test change" signal rather than adding an `area/testing` label.

elevran (approved):

PR #2214, opened 2026-07-28
feat(datalayer): configurable per-source scrape interval

### LINKED ISSUE #1880: feat: configurable per-source scraping interval in the data layer

noalimoy:
**What type of PR is this?**
/kind feature

All polling data-layer sources share one Collector goroutine per endpoint and one base tick (`--refresh-metrics-interval`, default 50ms). That cadence fits latency-sensitive model-server metrics, but wastes HTTP traffic for sources that update more slowly (e.g. DCGM ~1s, `/v1/models`).

Each polling source plugin now accepts an optional `interval` parameter (alongside `scheme`, `path`, `port`, etc.). The `PollingDispatcher` interface exposes `Interval() time.Duration`; the Runtime converts it to a period in base ticks (must be a positive multiple of the base tick) and the Collector dispatches a source only when due. Omitting `interval` keeps every-tick behavior.

Scheduling uses a tick-based next-due check in the existing Collector loop. A full timing wheel was not needed for the small number of polling sources per endpoint; we can revisit if that grows.

Key design choices:
- `Interval() time.Duration` added to `PollingDispatcher` — each source declares its own cadence; zero means every base tick
- `intervalDispatcher` decorator — Collector and manager are untouched.
- Base tick derived from the smallest source interval (option 1) — no source is forced to run slower than configured
- Each endpoint gets its own decorator clone .
- `interval` lives in plugin `parameters` — source-specific configuration, same as port or scheme
- Non-exact multiples are rounded to the nearest base-tick multiple (minimum 1 tick); a warning is logged
- `NewHTTPDataSource` rejects negative intervals at construction time

Example:

```yaml
plugins:
- type: dcgm-data-source
  name: dcgm-source
  parameters:
    port: 9400
    interval: "1s"   # must be a multiple of --refresh-metrics-interval
- type: dcgm-extractor
  name: dcgm-extractor
dataLayer:
  sources:
  - pluginRef: dcgm-source
    extractors:
    - pluginRef: dcgm-extractor
```

Fixes #1880

## Test plan
- [x] intervalDispatcher fires only on period boundaries (period 1, 2, 3)
- [x] intervalDispatcher propagates inner Dispatch errors
- [x] Independent counters per decorator clone (no shared state)
- [x] periodTicks rounds non-exact multiples to nearest base-tick multiple
- [x] periodTicks clamps intervals smaller than base to 1 tick
- [x] Configure wraps dispatcher with correct period (75ms -> 2 ticks, 1s -> 20 ticks)
- [x] Configure lowers base tick to smallest source interval (5s flag + 1s source -> 1s base, slow source gets period 5)
- [x] NewRuntime clamps polling interval to 50ms floor
- [x] Negative interval rejected by NewHTTPDataSource constructor
- [x] `make presubmit` passes

```release-note
Polling data-layer sources accept an optional `interval` plugin parameter (e.g. `interval: "1s"`). It must be a positive multiple of `--refresh-metrics-interval` (default 50ms); when omitted, the source scrapes every base tick.
```


docs/architecture.md
```diff
@@ -224,6 +224,12 @@ The data layer follows a Source -> Extract -> Attribute lifecycle:
 - Extractors populate per-endpoint attributes in the shared datastore for scorers
 - Scoring can rely on numerical metrics or metadata (model ID, adapter tags)
 
+Polling sources share one Collector goroutine per endpoint. The base tick is
+`--refresh-metrics-interval` (default 50ms). Each polling source plugin accepts an
+`interval` parameter (e.g. `"1s"`) that must be a positive multiple of the base tick;
```
elevran: nit: 
- does not have to be exact multiple. we can do the math of `round-to-nearest(duration/base)` in the data layer runtime. In the doc just say that the interval will be rounded to the nearest multiple of base.
- I'm fine with keeping it as metrics refresh, but it isn't really tied to metrics anymore. Just something to consider. Rename and deprecation of the flag is not worth it, IMHO.

Q: 
how do you propose to treat cases where `interval < base tick`? Right now user needs to be aware and configure correctly. Two options (at least): (1) pick smallest interval as base period (not necessarily `--refresh-metrics-interval`) and define all others as multiple (2) use  `--refresh-metrics-interval` as base frequency, everything else runs off of it - log a warning if an interval is below and operate the lower values at the base frequency.
I prefer the first option.

side note: lets define a minimal base interval (50ms?) and log a warning if user sets lower.

pkg/epp/framework/interface/datalayer/plugin.go
```diff
@@ -74,10 +75,12 @@ type NotificationExtractor interface {
 //     DataLayerExtractErrorsTotal and do NOT surface as the return error.
 //   - AppendExtractor is a pure append; duplicate-Type detection is the caller's
 //     responsibility (see runtime.Configure).
+//   - Interval returns the desired scrape period. Zero means every base tick.
 type PollingDispatcher interface {
 	plugin.Plugin
 	Dispatch(ctx context.Context, ep Endpoint) error
 	AppendExtractor(ext plugin.Plugin) error
+	Interval() time.Duration
```
elevran, line 83: Q/nit:
another option would have been to add a (derived) interface with an additional `Interval()` method so that only interested collectors need to implement it and not need to add `Interval() time.Duration { return 0 }` where it is not needed. The runtime can assert and query the interval of only needed implementation.
WDYT?
Not pushing for it - looking for feedback.

pkg/epp/framework/plugins/datalayer/source/dcgm/datasource.go
```diff
@@ -71,6 +74,11 @@ func DCGMDataSourceFactory(name string, parameters *json.Decoder, _ plugin.Handl
 	if cfg.UseNodeAddress {
 		opts = append(opts, http.WithUseNodeAddress())
 	}
+	if intervalOpt, err := http.ParseIntervalOption(cfg.Interval); err != nil {
```
elevran: is there a parse duration in the standard lib that would work?

pkg/epp/framework/plugins/datalayer/source/models/README.md
```diff
@@ -22,6 +22,9 @@ The Models Data Source polls inference server pods for model information and pas
 - `insecureSkipVerify` (bool, optional, default: `true`): Skip TLS certificate verification.
 - `caCertPath` (string, optional): PEM CA bundle to verify the target's server cert.
 - `clientCertPath` / `clientKeyPath` (string, optional): client certificate for mTLS. Set both together.
+- `interval` (string, optional): Scrape period (e.g. `"5s"`). Must be a positive
```
elevran: nit: the same help string appears in multiple locations. Would it be possible to have it specified once (e.g., might need a data layer README)?

elevran (commented):
Thanks @noalimoy - generally can merge.
Left a few questions/nits for your conisderations, especially the note on base frequency defined higher than an interval.

elevran:
@noalimoy left feedback for your consideration; please also address failing lint

docs/architecture.md
noalimoy: Looked at both options in depth.

With the default base tick (50ms) both options produce identical behavior — no current source needs a shorter interval. The difference surfaces when a source sets `interval` below the base tick.

Option 1 has a subtle side effect in that case: lowering the base tick speeds up *all* sources (including ones with no explicit interval), not just the one that requested a faster rate. Option 2 keeps the base stable - `interval < base` runs at the base frequency with a warning log.

I'll implement option 2 (round to nearest multiple + warn + 50ms floor) in this PR. If a real use case for dynamic base derivation comes up we can add the two-pass logic as a follow-up — it's additive on top of option 2.

pkg/epp/framework/interface/datalayer/plugin.go
noalimoy: Considered a derived `IntervalDispatcher` interface. Since all three polling sources (dcgm, metrics, models) support `interval` and the zero value already means "every base tick", keeping it on `PollingDispatcher` avoids a type assertion in the runtime and makes the contract explicit - every polling source declares its cadence, even if it's the default.

If out-of-tree polling plugins become common and `Interval()` is truly never needed for most of them, splitting the interface would make more sense. for now I think the single interface is simpler.

pkg/epp/framework/plugins/datalayer/source/dcgm/datasource.go
noalimoy: Yes - `ParseIntervalOption` wraps `time.ParseDuration` from the standard lib. The wrapper just deduplicates the empty-check + parse + wrap-as-Option logic shared by all three factories.

pkg/epp/framework/plugins/datalayer/source/models/README.md
noalimoy: Agree the duplication isn't ideal. but its same pattern as `scheme`, `path`, etc. each plugin README is self-contained. `architecture.md` already covers the mechanism centrally.

pkg/epp/framework/plugins/datalayer/source/http/datasource.go
```diff
@@ -108,6 +111,27 @@ func WithUseNodeAddress() Option {
...
+// every base tick (the default).
+func WithInterval(d time.Duration) Option {
+	return func(o *options) { o.interval = d }
+}
+
+// ParseIntervalOption parses a duration string from plugin parameters and
+// returns a WithInterval Option. Returns (nil, nil) when raw is empty so
+// callers can simply append the non-nil result to their option slice.
+func ParseIntervalOption(raw string) (Option, error) {
+	if raw == "" {
```
elevran, line 125: restructure callers to check `raw == ""` themselves before calling, or have (better) `ParseIntervalOption` return a zero-value Option (i.e., no-op func with no side effects) instead of nil.

pkg/epp/datalayer/schedule.go
```diff
@@ -0,0 +1,48 @@
...
+type ScheduledDispatcher struct {
+	Dispatcher  fwkdl.PollingDispatcher
+	PeriodTicks int // must be >= 1; 1 = every base tick
+}
+
+// PeriodTicks converts a configured scrape interval into a positive number of
+// base ticks. interval <= 0 means every base tick (backward-compatible default).
+// A non-zero interval must be a positive multiple of base so cadences stay on
+// the Collector's single ticker (no arbitrary durations such as 53ms).
+func PeriodTicks(interval, base time.Duration) (int, error) {
```
elevran: the implementation does not seem to align with the response in the [comment](https://github.com/llm-d/llm-d-router/pull/2214/files#r3681500091). `PeriodTicks` still does the original behavior: it errors on interval < base or non-multiples, with no rounding, no warning log, and no minimum-base floor enforcement

pkg/epp/datalayer/schedule.go
elevran, line 1: Having thought about it some more, a generic `intervalDispatcher` that wraps any `fwkdl.PollingDispatcher`, decrements counter internally, and no-ops until counter is at zero would be simpler overall:
`Collector.Start` keeps its original `[]fwkdl.PollingDispatcher` signature, and `pollingDispatchers`/`Register`/`Scheduled()` in manager.go don't need to change at all. That's a reduction in touched files (no schedule.go at all, no manager.go and collector.go untouched).

docs/architecture.md
elevran: I could have been clearer.
What is the expected behavior when user sets `--refresh-metrics-interval` to 5s as the default and the config has a Plugin with an explicit interval of 1s?
Under option 1 it runs every 1s but under option 2 it runs every 5s. Both values are much larger than the 50ms floor.
Under option 2, the user must set `--refresh-metrics-interval` lower than every config interval and runs the risk of overriding the collectors with explicit intervals in the config

pkg/epp/datalayer/schedule.go
noalimoy: Oops, I missed the `git push` - the changes were ready locally but not pushed, which is why you still saw the original behavior

docs/architecture.md
noalimoy: Agreed, option 1 is cleaner for the user. One thing to be aware of: lowering the base tick also speeds up sources that don't set an explicit `interval` (they run every base tick by default). With the current 50ms default that's a non-issue, but worth noting.

I'll add the pre-scan as well.

pkg/epp/datalayer/schedule.go
noalimoy: Done.

docs/architecture.md
elevran: since we floor it on 50ms, Sources without interval would not be running faster than it runs today, so I think that's fine.
The 50ms floor should probably go into Option.go parsing and validation. Can be a follow up if not already done

elevran (approved):
Thanks for the rework - the intervalDispatcher decorator is clean and the per-endpoint clone with independent counters is exactly right. 

Can you please ensure the 50ms floor is in Option.go parsing/validation as a follow-up if not already done?

Issue #1686, opened 2026-06-18
Conditional-decode 412 gate is hard-coded; make it configurable via a plugin


dmitripikus:
**What would you like to be added**:

The RFC 7240 `Prefer: if-available` (conditional-decode) gate in the EPP director should be moved out of the director and into a configurable plugin. The threshold should be expressed in non-cached tokens, consistent with the `nonCachedTokens` parameter the `prefix-based-pd-decider` plugin already uses for P/D disaggregation decisions.

In details:

- A `ConditionalDecodeDecider` extension-point interface in `pkg/epp/framework/interface/requestcontrol/`.
- `prefix-based-pd-decider` implements it (and other deciders can be added later) without director changes.
- The director consults the configured decider when the request carries `Prefer: if-available`, instead of running its own  logic.
- Deployments that do not declare a decider get a no-op gate (forward unconditionally) — the gate becomes opt-in.

**Why is this needed**:

The current implementation is simplistic and non-configurable:

- The director has a hard-coded "at least one cached prefix block" rule (`primaryEndpointHasCachedPrefix` in `pkg/epp/requestcontrol/director.go`). Operators currently have no way to tune how aggressive the 412 gate is for their workload.
- A single fixed threshold expressed in *blocks* is too eager for long prompts (a tiny prefix hit forwards anyway) and too conservative for short ones.
- The director duplicates cache-state extraction (`PrefixCacheMatchInfo` lookup, unweighted-cached-block accounting) that `prefix-based-pd-decider` already does for the same `PrefixCacheMatchInfo`.

dmitripikus:
Hi @elevran , @roytman , @ahg-g , @liu-cong ,

I plan to configure and implement decision for Conditional-decode using approach similar to that in previous PD implementation.

Configuration:

```bash
  epd-plugins.yaml: |
    ...    
    plugins:
    ...
    - type: prefix-based-pd-decider
      parameters:
        nonCachedTokens: 300
```
I plan to use decision logics - based on number of nonCachedTokens - that sits in `prefix-based-pd-decider` plugin, and invoke it  from 'director.go', after scheduling.
Here is some primary example (not a final version):
https://github.com/dmitripikus/llm-d-router/blob/4035835f2f855f57ca7f1a801994a4da9049382f/pkg/epp/requestcontrol/director.go#L310

## My question is:

Looking at:
```bash
type Config struct {
	preAdmissionPlugins      []fwkrc.PreAdmitter
[...]

PR #2215, opened 2026-07-28
feat(epp): add error return to PreRequest plugin interface

dmitripikus:
**What type of PR is this?**

/kind feature

Adds an `error` return to the `PreRequest` plugin interface so plugins can signal failure to the director. The director's runner now invokes every registered `PreRequest` plugin regardless of individual failures and returns their aggregated errors via `errors.Join`, so a caller sees all failures at once rather than only the first.

This is groundwork for #1686 (conditional-decode routing), where a new plugin at the `PreRequest` extension point needs to be able to fail a request. Per @elevran's [comment on the issue](https://github.com/llm-d/llm-d-router/issues/1686#issuecomment-3626229587), this preliminary PR lands the interface change on its own so the conditional-decode work can build on it.

Every existing implementation of `PreRequest` returns `nil` today; behavior is unchanged. Individual plugins that currently log-and-swallow errors (e.g. `preciseprefixcache` calling `index.Add`) are left as-is here — surfacing those errors is a follow-up.

Note on aggregation strategy: this diverges from the first-error-wins pattern used by `runAdmissionPlugins` and `runRequestHeaderProcessors`. Every `PreRequest` plugin runs its side effects even if a peer plugin has already failed, and the joined error covers all of them.

Issue #2216

**Release note**:
The `PreRequest` plugin interface now returns an `error`. Existing built-in plugins return `nil` and are unaffected; out-of-tree plugin authors implementing `PreRequest` must update their method signatures.

### DIFF
--- a/pkg/epp/flowcontrol/eviction/request_evictor_test.go
+++ b/pkg/epp/flowcontrol/eviction/request_evictor_test.go
@@ -67,7 +67,7 @@ func TestRequestEvictor_PreRequest_CreatesEvictChannel(t *testing.T) {
 	re := NewRequestEvictor(&testOrdering{}, &acceptAllFilter{}, &NoOpEvictor{})
 
 	ctx := context.Background()
-	_ = re.PreRequest(ctx, makeInferenceRequest("req-1", -1), makeSchedulingResult())
+	require.NoError(t, re.PreRequest(ctx, makeInferenceRequest("req-1", -1), makeSchedulingResult()))
 
 	evictCh := re.EvictionRegistry().Get("req-1")
 	require.NotNil(t, evictCh, "EvictCh should be registered after PreRequest")
@@ -82,7 +82,7 @@ func TestRequestEvictor_ResponseBody_DeregistersEvictChannel(t *testing.T) {
 
 	ctx := context.Background()
 	request := makeInferenceRequest("req-1", -1)
-	_ = re.PreRequest(ctx, request, makeSchedulingResult())
+	require.NoError(t, re.PreRequest(ctx, request, makeSchedulingResult()))
 	require.NotNil(t, re.EvictionRegistry().Get("req-1"))
 
 	re.ResponseBody(ctx, request, &requestcontrol.Response{EndOfStream: true}, nil)
@@ -97,7 +97,7 @@ func TestRequestEvictor_EvictN_ClosesEvictChannel(t *testing.T) {
 	re := NewRequestEvictor(&testOrdering{}, &acceptAllFilter{}, evictor)
 
 	ctx := context.Background()
-	_ = re.PreRequest(ctx, makeInferenceRequest("req-1", -1), makeSchedulingResult())
+	require.NoError(t, re.PreRequest(ctx, makeInferenceRequest("req-1", -1), makeSchedulingResult()))
 
 	evictCh := re.EvictionRegistry().Get("req-1")
 	require.NotNil(t, evictCh)
@@ -121,7 +121,7 @@ func TestRequestEvictor_EvictN_ReTracksOnFailure(t *testing.T) {
 	re := NewRequestEvictor(&testOrdering{}, &acceptAllFilter{}, &failingEvictor{})
 
 	ctx := context.Background()
-	_ = re.PreRequest(ctx, makeInferenceRequest("req-1", -1), makeSchedulingResult())
+	require.NoError(t, re.PreRequest(ctx, makeInferenceRequest("req-1", -1), makeSchedulingResult()))
 
 	evicted, err := re.EvictN(ctx, 1)
 	require.NoError(t, err)
@@ -140,7 +140,7 @@ func TestRequestEvictor_RaceBetweenEvictAndCompletion(t *testing.T) {
 	requests := make([]*scheduling.InferenceRequest, 10)
 	for i := range requests {
 		requests[i] = makeInferenceRequest("req-"+string(rune('a'+i)), -1)
-		_ = re.PreRequest(ctx, requests[i], makeSchedulingResult())
+		require.NoError(t, re.PreRequest(ctx, requests[i], makeSchedulingResult()))
 	}
 
 	var wg sync.WaitGroup
@@ -176,7 +176,7 @@ func TestRequestEvictor_CtxCancellationTriggersCleanup(t *testing.T) {
 	re := NewRequestEvictor(&testOrdering{}, &acceptAllFilter{}, evictor)
 
 	ctx, cancel := context.WithCancel(context.Background())
-	_ = re.PreRequest(ctx, makeInferenceRequest("req-1", -1), makeSchedulingResult())
+	require.NoError(t, re.PreRequest(ctx, makeInferenceRequest("req-1", -1), makeSchedulingResult()))
 
 	// Verify tracked.
 	assert.Equal(t, 1, re.queue.InFlightLen())
@@ -201,7 +201,7 @@ func TestRequestEvictor_CleanupCallsEvictorCleanup(t *testing.T) {
 	re := NewRequestEvictor(&testOrdering{}, &acceptAllFilter{}, evictor)
 
 	ctx := context.Background()
-	_ = re.PreRequest(ctx, makeInferenceRequest("req-1", -1), makeSchedulingResult())
+	require.NoError(t, re.PreRequest(ctx, makeInferenceRequest("req-1", -1), makeSchedulingResult()))
 
 	// Evict to create a sync.Once entry in the evictor.
 	_, _ = re.EvictN(ctx, 1)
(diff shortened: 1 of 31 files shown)


pkg/epp/requestcontrol/director_test.go
```diff
@@ -185,16 +185,18 @@ func (m *mockRequestHeaderPlugin) RequestHeader(_ context.Context, request *fwks
 type mockPreRequestPlugin struct {
 	name     string
 	modifyFn func(request *fwksched.InferenceRequest)
+	err      error
```
elevran, line 188: this new err field is exercised by `TestRunPreRequestPlugins_AggregatesErrors` but not by the existing `preRequestPlugin` case in `TestDirector_HandleRequest` (around line 512), which never sets err. 
That leaves the new error-propagation branch in `prepareRequest/HandleRequest` (director.go:484-486) untested at the **integration** level. Add a case there where the plugin fails and assert on the resulting response/status - this would have caught the classification gap noted in the review body (see error.go:78).

pkg/epp/requestcontrol/director.go
```diff
@@ -588,16 +590,27 @@ func (d *Director) GetRandomEndpoint() *fwkdl.EndpointMetadata {
...
+// runPreRequestPlugins invokes every registered PreRequest plugin, wraps each non-nil
+// error with the plugin's typed name, and returns the joined result. Plugins are not
+// short-circuited: a failure in one does not skip the rest, so every plugin still runs
+// its side effects and the caller sees all failures.
 func (d *Director) runPreRequestPlugins(ctx context.Context, request *fwksched.InferenceRequest,
-	schedulingResult *fwksched.SchedulingResult) {
+	schedulingResult *fwksched.SchedulingResult) error {
 	loggerDebug := log.FromContext(ctx).V(logutil.DEBUG)
+	var errs []error
 	for _, plugin := range d.requestControlPlugins.preRequestPlugins {
```
elevran, line 631: since this loop does not short-circuit, if an earlier plugin applies an irreversible side effect and a later plugin then fails, the earlier side effect is not rolled back. 
Today this is safe because server.go's check on `TargetPod != nil && !ResponseComplete` cleanup fires `HandleResponseBody` for any aborted request with a picked pod, covering the existing `RequestEvictor` tracking. However, worth documenting as an explicit invariant so future `PreRequest` plugins with side effects (tracked alongside #1686) don't quietly violate it.

pkg/epp/requestcontrol/director_test.go
```diff
@@ -1974,3 +1976,36 @@ func TestDirector_HandleRequest_ConditionalDecode(t *testing.T) {
 		})
 	}
 }
+
+// TestRunPreRequestPlugins_AggregatesErrors verifies that runPreRequestPlugins
+// invokes every registered PreRequest plugin regardless of individual failures
+// and returns the joined error covering all failed plugins.
+func TestRunPreRequestPlugins_AggregatesErrors(t *testing.T) {
```
elevran, line 2191: nit: no explicit case here for zero-plugins / all-succeed on runPreRequestPlugins. Low risk given integration coverage elsewhere, but cheap to add.

pkg/epp/requestcontrol/director.go
elevran, line 1: The aggregated error from `runPreRequestPlugins` (in director.go:601-613, not changed in this PR) is never classified once it reaches the response-building layer. It wraps each plugin error with fmt.Errorf and joins them with errors.Join, so CanonicalCode's type assertion (in pkg/common/error/error.go:78 (`err.(Error)`)) can never match it - it always falls through to Unknown. `BuildErrResponse`'s default branch then returns `status.Errorf(status.Code(err), ...)` with `codes.Unknown`, and that raw error is returned directly from the stream handler instead of a proper Envoy `ImmediateResponse`, unlike every other error path in director.go. The full joined error text is also embedded verbatim in the gRPC status and OTel span, leaking internal detail.

This is dormant today because every existing PreRequest implementation returns nil, but it will break as soon as a real erroring plugin lands (tracked in #1686), which is exactly what this groundwork is meant to enable. Either use `errors.As` in `CanonicalCode` to find an `errcommon.Error` inside a wrapped/joined chain, or having `runPreRequestPlugins`/`prepareRequest` translate the joined error into a single `errcommon.Error` before it reaches `BuildErrResponse`.

Additional note: the `PreRequest` interface doc comment should mention that only an `errcommon.Error` return is fully classified downstream, so future plugin authors return the right type.

elevran (commented):
Thanks @dmitripikus - this lays the groundwork by threading an error return through the PreRequest plugin extension point, aggregates failures without short-circuiting, and is covered by a solid new unit test.

pkg/epp/requestcontrol/director.go
```diff
@@ -588,16 +590,27 @@ func (d *Director) GetRandomEndpoint() *fwkdl.EndpointMetadata {
...
+		err := plugin.PreRequest(ctx, request, schedulingResult)
 		metrics.RecordPluginProcessingLatency(fwkrc.PreRequestExtensionPoint, plugin.TypedName().Type, plugin.TypedName().Name, time.Since(before))
+		if err != nil {
+			loggerDebug.Info("PreRequest plugin failed", "plugin", plugin.TypedName(), "error", err.Error())
+			errs = append(errs, fmt.Errorf("PreRequest %q failed: %w", plugin.TypedName().String(), err))
+			continue
+		}
 		loggerDebug.Info("Completed running PreRequest plugin successfully", "plugin", plugin.TypedName())
 	}
+	return errors.Join(errs...)
```
bongwoobak, line 643: A typed `errcommon.Error` does not survive this. `CanonicalCode` does a bare type assertion `err.(Error)` instead of `errors.As` (pkg/common/error/error.go:78), so anything wrapped with `%w` and joined reads as `Unknown`, and `BuildErrResponse` falls into the default branch which returns no ImmediateResponse at all. server.go:556 then just ends the ext_proc stream with a gRPC error, so the client gets neither a status code nor a body.

I ran it with one mock plugin returning `errcommon.Error{Code: PreconditionFailed}`. The joined error gives `CanonicalCode = "Unknown"` and `BuildErrResponse` returns `(nil, gRPC Unknown)`. Passing the same error unwrapped gives `PreconditionFailed` and a real response. `errors.Join` wraps even a single error, so the one-plugin-fails case hits this too.

Every other error path out of HandleRequest avoids it, either by returning a typed error directly (admission, scheduler, repackage, the conditional-decode gate at director.go:340) or by returning the plugin error unwrapped so a typed one survives (runRequestHeaderProcessors, line 624). This is the only one that wraps, which means a plugin has no way to choose its status code.

That is the part that matters for what this is groundwork for. #1686 wants the decider to answer 412, and the comment there says the plan is to collect the errors and return them in a body, but the body is what never gets built. `errors.As` does find the typed error inside the joined tree, so unwrapping when there is a single typed failure and collapsing to `Internal` otherwise would cover it. director.go:320-331 already does that same unwrap for the scheduler.

pkg/epp/requestcontrol/director_test.go
dmitripikus: I've added code to preserve typed PreRequest plugin error codes, and I added relevant tests. Thank you!

pkg/epp/requestcontrol/director.go
dmitripikus: The item is documented, Thank you!

pkg/epp/requestcontrol/director_test.go
dmitripikus: Done. Thanks!

pkg/epp/flowcontrol/eviction/request_evictor_test.go
```diff
@@ -67,7 +67,7 @@ func TestRequestEvictor_PreRequest_CreatesEvictChannel(t *testing.T) {
 	re := NewRequestEvictor(&testOrdering{}, &acceptAllFilter{}, &NoOpEvictor{})
 
 	ctx := context.Background()
-	re.PreRequest(ctx, makeInferenceRequest("req-1", -1), makeSchedulingResult())
+	_ = re.PreRequest(ctx, makeInferenceRequest("req-1", -1), makeSchedulingResult())
```
capri-xiyue: nits: For the test, should we check the return is nil?

pkg/epp/flowcontrol/integration_test.go
```diff
@@ -90,7 +90,7 @@ func TestConcurrentSaturationReads(t *testing.T) {
 					"decode": {TargetEndpoints: []fwksched.Endpoint{schedEp}},
 				},
 			}
-			pd.producer.PreRequest(ctx, req, result)
+			_ = pd.producer.PreRequest(ctx, req, result)
```
capri-xiyue: same here, nits: For the test, should we check the return is nil?

pkg/epp/requestcontrol/director.go
dmitripikus: The classification issue was addressed in commit c7f4d98a
The PreRequest interface doc comment is updated
Thank you!

pkg/epp/requestcontrol/director.go
dmitripikus: Fixed in commit c7f4d98a and tightened in commit 9a496de8. prepareRequest now unwraps when there is a single typed failure, collapses to Internal otherwise so the joined text still reaches the client instead of an arbitrarily picked typed code.

Relevant tests are added.

Thanks!

pkg/epp/requestcontrol/director.go
```diff
@@ -499,7 +499,18 @@ func (d *Director) prepareRequest(ctx context.Context, reqCtx *handlers.RequestC
...
+		// (e.g. PreconditionFailed) reaches Envoy intact, even after the
+		// wrapping applied by runPreRequestPlugins (fmt.Errorf + errors.Join).
+		// Fall through to Internal for untyped/multiple errors so response
+		// building does not fall through to Unknown in BuildErrResponse.
+		var e errcommon.Error
+		if errors.As(err, &e) {
+			return reqCtx, e
+		}
+		return reqCtx, errcommon.Error{Code: errcommon.Internal, Msg: err.Error()}
+	}
```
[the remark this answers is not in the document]
dmitripikus: Commit 9a496de8 ([fix(epp): collapse multi-plugin PreRequest failures to Internal](vscode-webview://10jnhlbkpeqplhv757jvbc4o9bjra45ojsua8quvcvp9pr2shb6p/pkg/epp/requestcontrol/director.go#L502-L517)) addresses this comment

pkg/epp/flowcontrol/eviction/request_evictor_test.go
dmitripikus: Tests are updated, so that a non-nil return actually fails the test.
Thank you!

pkg/epp/flowcontrol/integration_test.go
dmitripikus: Tests are updated, so that a non-nil return actually fails the test.
Thank you!

ahg-g (approved):

PR #2221, opened 2026-07-29
[Flow Control] Stage-Aware Saturation Gating for disaggregated P/D deployments

### LINKED ISSUE #1186: [Flow Control / Disaggregation] Stage-Aware Saturation Gating for Prefill/Decode Pools
Flow control currently treats all model servers as a single uniform pool when evaluating saturation. In a disaggregated Prefill/Decode (P/D) architecture, this aggregate average obscures whether the specific bottleneck lies in the Prefill tier or the Decode stage. A surge in Prefill requests could overwhelm Prefill workers while the global pool average appears healthy because Decode workers are underutilized.

To protect disaggregated serving pools from dynamic bottlenecks, flow control must become stage-aware. We will dynamically partition the candidate endpoints in EPP's dispatch loop based on standard pod role labels, evaluate saturation on each tier independently, and enforce progressive backpressure based on a combined max-saturation rule ($\max(S_{prefill}, S_{decode})$).

---

### 1. Update Prometheus Saturation Metrics

We will extend EPP's `flow_control_pool_saturation` Prometheus gauge in `pkg/epp/metrics/metrics.go` to accept a `stage` label, allowing operators to monitor saturation metrics per disaggregated stage.

**Technical Requirements:**
*   Gauge Vector Update: Add a `"stage"` string to EPP's `flowControlPoolSaturation` labels array.

---

### 2. Implement Dynamic Endpoint Partitioning

We will group candidate endpoints dynamically based on EPP's established role label conventions.

**Technical Requirements:**
*   Role Constants: Define standard role key and value constants (location TBD):
    ```go
    const (
    	PodRoleLabelKey         = "llm-d.ai/role"
    	PodRolePrefill          = "prefill"
    	PodRoleDecode           = "decode"
    	PodRolePrefillDecode    = "prefill-decode"
    	PodRoleBoth             = "both" // Legacy alias
    	PodRoleAllStage         = "encode-prefill-decode"
    )
    ```
*   Partitioning Helper: Implement the dynamic partitioning helper inside `pkg/epp/flowcontrol/controller/internal/processor.go`:
    ```go
    func partitionEndpoints(endpoints []fwkdl.Endpoint, labelKey string) (prefill, decode, interleaved []fwkdl.Endpoint) {
    		meta := ep.GetMetadata()
    		if meta == nil || meta.Labels == nil {
    			interleaved = append(interleaved, ep)
    			continue
    		}
    		role := meta.Labels[labelKey]
    		switch role {
    		case PodRolePrefill, "encode-prefill":
    			prefill = append(prefill, ep)
    		case PodRoleDecode:
    			decode = append(decode, ep)
    		case PodRolePrefillDecode, PodRoleBoth, PodRoleAllStage:
    			interleaved = append(interleaved, ep)
    		case "":
    			// Endpoints completely lacking the label default to decode for monolithic safety
    			decode = append(decode, ep)
    		default:
    			interleaved = append(interleaved, ep)
    		}
    	}
    	return
    }
    ```

---

### 3. Implement Saturation Gating Max-Gate

We will update `ShardProcessor.dispatchCycle` in `pkg/epp/flowcontrol/controller/internal/processor.go` to evaluate saturation independently across stages and apply backpressure if *either* stage is overloaded.

**Technical Requirements:**
*   Partition Candidates: Partition the candidate pool inside `ShardProcessor.dispatchCycle` using the new helper.
*   Independent Evaluation & Stage Metrics: Compute saturation independently on each partition and emit stage-specific metrics:
    ```go
    prefillSat = sp.saturationDetector.Saturation(ctx, prefillPool)
    metrics.RecordFlowControlPoolSaturation(sp.poolName, "prefill", prefillSat)

    decodeSat = sp.saturationDetector.Saturation(ctx, decodePool)
    metrics.RecordFlowControlPoolSaturation(sp.poolName, "decode", decodeSat)

    interleavedSat = sp.saturationDetector.Saturation(ctx, interleavedPool)
    metrics.RecordFlowControlPoolSaturation(sp.poolName, "interleaved", interleavedSat)
    ```
*   Tandem Queue Gating: Compute the overall gating saturation as:
    $$\text{saturation} = \max(\text{prefillSat}, \text{decodeSat})$$. If interleaved workers are active, evaluate $$\max(\text{saturation}, \text{interleavedSat})$$.
*  Integration: Record the global metric and assign this disaggregation-aware saturation directly to the local `saturation` variable:
    ```go
    metrics.RecordFlowControlPoolSaturation(sp.poolName, "global", saturation)
    ```
    The computed priority ceilings and HoL blocking checks (`saturation >= usageLimit`) will automatically consume this ceiling without further data-path changes.

loicmarchal:
**What type of PR is this?**

/kind feature

The flow control system currently computes a single saturation value across all endpoints. In disaggregated prefill/decode (P/D) deployments, this masks stage-specific bottlenecks: a prefill surge can overwhelm prefill workers while the global average stays low because idle decode workers dilute the signal, so backpressure never kicks in when it should.

This PR makes the dispatch cycle stage-aware by:

1. Adding a `partitionEndpoints` helper that classifies endpoints into prefill, decode, and interleaved buckets using the `llm-d.ai/role` pod label (reusing role constants from `bylabel/roles.go`). Endpoints without a role label or without metadata default to the decode bucket, matching the existing `NewDecodeRole(allowsNoLabel: true)` convention for monolithic deployment safety. Encode-only pods and unrecognized role values are excluded from all buckets because they are rejected by every role filter and receive no traffic.
2. Interleaved endpoints (`prefill-decode`, `both`, `encode-prefill-decode`) are included in both the prefill and decode pools, mirroring how the scheduling filters route requests. Saturation is computed independently per non-empty stage, then `max(S_prefill, S_decode)` is used as the effective saturation fed into priority ceilings and HoL blocking. Empty partitions delete their gauge series to prevent stale readings. When all partitions are empty, the detector is called on the full pool to preserve existing empty-pool behavior.
3. Adding a `"stage"` label to both `flow_control_pool_saturation` Prometheus gauges, emitting per-stage (`"prefill"`, `"decode"`) and `"effective"` values each dispatch cycle. The gauge help text and `docs/metrics.md` are updated to document the new label.

The existing gating logic (priority ceilings, HoL blocking) is unchanged -- it uses the effective saturation value as before. Monolithic deployments (no role labels) behave identically to before: all endpoints land in the decode bucket, producing a single-stage saturation equal to the previous flat-pool value.

Fixes #1186

**Release note**:

```release-note
Flow control saturation gating is now stage-aware in disaggregated prefill/decode deployments. Endpoints are partitioned by the `llm-d.ai/role` pod label, with interleaved pods included in both stage pools. Saturation is computed per stage, and backpressure triggers on the worst stage (`max(S_prefill, S_decode)`). A new `"stage"` label on the `flow_control_pool_saturation` metric enables per-stage observability (`"prefill"`, `"decode"`, `"effective"`). Monolithic deployments are unaffected.
```

### DIFF
--- a/pkg/epp/flowcontrol/controller/internal/processor.go
+++ b/pkg/epp/flowcontrol/controller/internal/processor.go
@@ -463,6 +463,7 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
 		{"decode", decodePool},
 	} {
 		if len(part.endpoints) == 0 {
+			metrics.DeleteFlowControlPoolSaturation(p.poolName, part.name)
 			continue
 		}
 		stageSat := p.saturationDetector.Saturation(ctx, part.endpoints)
--- a/pkg/epp/flowcontrol/controller/internal/processor.go
+++ b/pkg/epp/flowcontrol/controller/internal/processor.go
@@ -447,7 +447,12 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
 		p.regime.Store(&regimeSample{empty: empty, since: p.clock.Now()})
 	}
 
-	prefillPool, decodePool, interleavedPool := partitionEndpoints(pool)
+	prefillOnly, decodeOnly, interleaved := partitionEndpoints(pool)
+
+	// Interleaved pods serve both stages, so they contribute capacity to both pools,
+	// mirroring how the scheduling filters route requests.
+	prefillPool := append(prefillOnly, interleaved...)
+	decodePool := append(decodeOnly, interleaved...)
 
 	saturation := -1.0
 	for _, part := range []struct {
@@ -456,7 +461,6 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
 	}{
 		{"prefill", prefillPool},
 		{"decode", decodePool},
-		{"interleaved", interleavedPool},
 	} {
 		if len(part.endpoints) == 0 {
 			continue
--- a/pkg/epp/flowcontrol/controller/internal/processor.go
+++ b/pkg/epp/flowcontrol/controller/internal/processor.go
@@ -542,8 +542,10 @@ func (p *Processor) ceilingsBuffer(n int) []float64 {
 }
 
 // partitionEndpoints classifies endpoints into prefill, decode, and interleaved buckets
-// based on the llm-d.ai/role pod label. Endpoints without a role label, without metadata,
-// or with an unrecognized role value default to the decode bucket for monolithic deployment safety.
+// based on the llm-d.ai/role pod label. Endpoints without a role label or without metadata
+// default to the decode bucket, matching the decode-filter's allowsNoLabel convention for
+// monolithic deployment safety. Encode-only pods and unrecognized role values are excluded
+// from all buckets because they are rejected by every role filter and receive no traffic.
 func partitionEndpoints(endpoints []fwkdl.Endpoint) (prefill, decode, interleaved []fwkdl.Endpoint) {
 	for _, ep := range endpoints {
 		if ep == nil {
@@ -558,12 +560,16 @@ func partitionEndpoints(endpoints []fwkdl.Endpoint) (prefill, decode, interleave
 		switch role {
 		case bylabel.RolePrefill, bylabel.RoleEncodePrefill:
 			prefill = append(prefill, ep)
-		case bylabel.RoleDecode:
+		case bylabel.RoleDecode, "":
 			decode = append(decode, ep)
 		case bylabel.RolePrefillDecode, bylabel.RoleBoth, bylabel.RoleEncodePrefillDecode:
 			interleaved = append(interleaved, ep)
+		case bylabel.RoleEncode:
+			// Encode-only pods receive no prefill or decode traffic; excluding them
+			// keeps both stage signals clean.
 		default:
-			decode = append(decode, ep)
+			// Unrecognized role values are rejected by every role filter and receive
+			// no traffic; counting them anywhere dilutes the stage signal.
 		}
 	}
 	return
--- a/pkg/epp/flowcontrol/controller/internal/processor.go
+++ b/pkg/epp/flowcontrol/controller/internal/processor.go
@@ -471,7 +471,7 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
 		saturation = p.saturationDetector.Saturation(ctx, pool)
 	}
 
-	metrics.RecordFlowControlPoolSaturation(p.poolName, "global", saturation)
+	metrics.RecordFlowControlPoolSaturation(p.poolName, "effective", saturation)
 
 	// Record capacity utilization ratios (the demand-side twin of saturation) from the same periodic sample.
 	p.recordCapacityUtilization()
--- a/pkg/epp/flowcontrol/controller/internal/processor.go
+++ b/pkg/epp/flowcontrol/controller/internal/processor.go
@@ -446,7 +446,7 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
 	saturation := p.saturationDetector.Saturation(ctx, pool)
 
 	// Record pool saturation metric
-	metrics.RecordFlowControlPoolSaturation(p.poolName, saturation)
+	metrics.RecordFlowControlPoolSaturation(p.poolName, "global", saturation)
 
 	// Record capacity utilization ratios (the demand-side twin of saturation) from the same periodic sample.
 	p.recordCapacityUtilization()
--- a/pkg/epp/flowcontrol/controller/internal/processor.go
+++ b/pkg/epp/flowcontrol/controller/internal/processor.go
@@ -447,20 +447,20 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
 		p.regime.Store(&regimeSample{empty: empty, since: p.clock.Now()})
 	}
 
-	prefillOnly, decodeOnly, interleaved := partitionEndpoints(pool)
+	prefill, decode, interleaved := partitionEndpoints(pool)
 
 	// Interleaved pods serve both stages, so they contribute capacity to both pools,
 	// mirroring how the scheduling filters route requests.
-	prefillPool := append(prefillOnly, interleaved...)
-	decodePool := append(decodeOnly, interleaved...)
+	prefill = append(prefill, interleaved...)
+	decode = append(decode, interleaved...)
 
 	saturation := -1.0
 	for _, part := range []struct {
 		name      string
 		endpoints []fwkdl.Endpoint
 	}{
-		{"prefill", prefillPool},
-		{"decode", decodePool},
+		{"prefill", prefill},
+		{"decode", decode},
 	} {
 		if len(part.endpoints) == 0 {
 			metrics.DeleteFlowControlPoolSaturation(p.poolName, part.name)
@@ -563,7 +563,7 @@ func partitionEndpoints(endpoints []fwkdl.Endpoint) (prefill, decode, interleave
 			prefill = append(prefill, ep)
 		case bylabel.RoleDecode, "":
 			decode = append(decode, ep)
-		case bylabel.RolePrefillDecode, bylabel.RoleBoth, bylabel.RoleEncodePrefillDecode:
+		case bylabel.RolePrefillDecode, bylabel.RoleBoth, bylabel.RoleEncodePrefillDecode: //nolint:staticcheck // RoleBoth is deprecated but must be matched for backward compatibility
 			interleaved = append(interleaved, ep)
 		case bylabel.RoleEncode:
 			// Encode-only pods receive no prefill or decode traffic; excluding them
--- a/pkg/epp/flowcontrol/controller/internal/processor.go
+++ b/pkg/epp/flowcontrol/controller/internal/processor.go
@@ -32,7 +32,9 @@ import (
 	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
 	"github.com/llm-d/llm-d-router/pkg/epp/flowcontrol/contracts"
 	"github.com/llm-d/llm-d-router/pkg/epp/flowcontrol/types"
+	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/flowcontrol"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/filter/bylabel"
 	"github.com/llm-d/llm-d-router/pkg/epp/metrics"
 )
 
@@ -439,13 +441,36 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
 	}()
 
 	pool := p.endpointCandidates.Locate(ctx, nil)
+
 	// Run is the sole writer, so the load and the store cannot interleave with another write.
 	if empty := len(pool) == 0; empty != p.regime.Load().empty {
 		p.regime.Store(&regimeSample{empty: empty, since: p.clock.Now()})
 	}
-	saturation := p.saturationDetector.Saturation(ctx, pool)
 
-	// Record pool saturation metric
+	prefillPool, decodePool, interleavedPool := partitionEndpoints(pool)
+
+	saturation := -1.0
+	for _, part := range []struct {
+		name      string
+		endpoints []fwkdl.Endpoint
+	}{
+		{"prefill", prefillPool},
+		{"decode", decodePool},
+		{"interleaved", interleavedPool},
+	} {
+		if len(part.endpoints) == 0 {
+			continue
+		}
+		stageSat := p.saturationDetector.Saturation(ctx, part.endpoints)
+		metrics.RecordFlowControlPoolSaturation(p.poolName, part.name, stageSat)
+		if stageSat > saturation {
+			saturation = stageSat
+		}
+	}
+	if saturation < 0 {
+		saturation = p.saturationDetector.Saturation(ctx, pool)
+	}
+
 	metrics.RecordFlowControlPoolSaturation(p.poolName, "global", saturation)
 
 	// Record capacity utilization ratios (the demand-side twin of saturation) from the same periodic sample.
@@ -510,6 +535,34 @@ func (p *Processor) ceilingsBuffer(n int) []float64 {
 		buf[i] = 1.0
 	}
 	return buf
+}
+
+// partitionEndpoints classifies endpoints into prefill, decode, and interleaved buckets
+// based on the llm-d.ai/role pod label. Endpoints without a role label, without metadata,
+// or with an unrecognized role value default to the decode bucket for monolithic deployment safety.
+func partitionEndpoints(endpoints []fwkdl.Endpoint) (prefill, decode, interleaved []fwkdl.Endpoint) {
+	for _, ep := range endpoints {
+		if ep == nil {
+			continue
+		}
+		meta := ep.GetMetadata()
+		if meta == nil || meta.Labels == nil {
+			decode = append(decode, ep)
+			continue
+		}
+		role := meta.Labels[bylabel.RoleLabel]
+		switch role {
+		case bylabel.RolePrefill, bylabel.RoleEncodePrefill:
+			prefill = append(prefill, ep)
+		case bylabel.RoleDecode:
+			decode = append(decode, ep)
+		case bylabel.RolePrefillDecode, bylabel.RoleBoth, bylabel.RoleEncodePrefillDecode:
+			interleaved = append(interleaved, ep)
+		default:
+			decode = append(decode, ep)
+		}
+	}
+	return
 }
 
 // selectItem applies the configured fairness and ordering policies to select a single item.
--- a/pkg/epp/flowcontrol/controller/internal/processor_test.go
+++ b/pkg/epp/flowcontrol/controller/internal/processor_test.go
@@ -1261,6 +1261,39 @@ func TestProcessor(t *testing.T) {
 				assert.True(t, dispatched, "should dispatch when only non-empty partitions are healthy")
 			})
 
+			t.Run("should include interleaved endpoints in both stage pools", func(t *testing.T) {
+				t.Parallel()
+				h := newTestHarness(t, testCleanupTick)
+
+				q := h.addQueue(testFlow)
+				require.NoError(t, q.Add(h.newTestItem("item-1", testFlow, testTTL)))
+
+				h.endpointCandidates.Candidates = []fwkdl.Endpoint{
+					makeEndpoint(bylabel.RolePrefill),
+					makeEndpoint(bylabel.RoleDecode),
+					makeEndpoint(bylabel.RolePrefillDecode), // interleaved
+				}
+
+				// Track which endpoints each Saturation call receives.
+				var calls [][]string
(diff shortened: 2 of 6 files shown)


pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -375,10 +377,32 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
...
+		{"prefill", prefillPool},
+		{"decode", decodePool},
+		{"interleaved", interleavedPool},
+	} {
+		if len(part.endpoints) == 0 {
+			continue
+		}
+		stageSat := p.saturationDetector.Saturation(ctx, part.endpoints)
+		metrics.RecordFlowControlPoolSaturation(p.poolName, part.name, stageSat)
+		if stageSat > saturation {
```
loicmarchal, line 397: this is currently implementing the simple approach to calculate `saturation = max(S_prefill, S_decode, S_interleaved)`.

I'm looking into a solution to count inflight prefill request into the decode saturation to correctly count for projected decode load and not gate too late (as discussed in https://github.com/llm-d/llm-d-router/issues/1186#issuecomment-4645218840).

LukeAVanDrie: I think we land simple first and defer this. I am also interested in investigating some approaches here if you're open to collaborating.

loicmarchal: @LukeAVanDrie so you suggest we go with this solution first and look into the projected decode saturation as a followup PR? That's ok with me. Sure, of course I'll be happy to collaborate on that one.

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -375,10 +377,32 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
...
+		endpoints []fwkdl.Endpoint
+	}{
+		{"prefill", prefillPool},
+		{"decode", decodePool},
+		{"interleaved", interleavedPool},
+	} {
+		if len(part.endpoints) == 0 {
+			continue
+		}
+		stageSat := p.saturationDetector.Saturation(ctx, part.endpoints)
```
LukeAVanDrie, line 395: The utilization detector records `flow_control_stale_endpoints` as a side effect of every `Saturation()` call, keyed only by detector name (as of #2165). Calling it once per partition overwrites the gauge per stage each cycle, so it ends up reflecting only the last non-empty partition (decode, in typical P/D).

A scrape outage on the prefill fleet scores every prefill pod as 1.0 (fail-closed), so `max()` pins effective saturation at exactly 1.0 and halts all dispatch, while the stale gauge reads 0 from the decode partition. The utilization README tells operators to use this gauge to tell a scrape outage from genuine overload, so the runbook now misdiagnoses.

There's not really a simple fix here since the recording lives inside the stage-blind detector. Options are a `stage` label on the stale gauge (requires plumbing stage identity into the call) or moving partitioning into a stage-aware detector that owns per-stage stale accounting.

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -423,6 +447,34 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
...
+		}
+		role := meta.Labels[bylabel.RoleLabel]
+		switch role {
+		case bylabel.RolePrefill, bylabel.RoleEncodePrefill:
+			prefill = append(prefill, ep)
+		case bylabel.RoleDecode:
+			decode = append(decode, ep)
+		case bylabel.RolePrefillDecode, bylabel.RoleBoth, bylabel.RoleEncodePrefillDecode:
+			interleaved = append(interleaved, ep)
+		default:
```
LukeAVanDrie, line 471: Two boundary cases land in decode that shouldn't, judging by the `bylabel` filter acceptance sets, which define where traffic can actually go:

- `RoleEncode` is a defined constant in the package this PR imports, and encode pods are not decode candidates. Here they pollute the decode signal with encode-stage load and inflate the decode denominator.
- A pod with a defined but unrecognized role value is rejected by every role filter and receives no traffic. Counting it as decode means an idle, unroutable pod dilutes the decode average. (`NewDecodeRole`'s `allowsNoLabel` covers missing labels, not unrecognized values, so it doesn't cover this case.)

Suggest an explicit `case bylabel.RoleEncode:` and excluding unknown roles from every bucket:

```go
case bylabel.RoleEncode:
	// Encode-only workers receive no prefill or decode traffic; excluding them
	// keeps both stage signals clean. (Or: their own bucket, if encode-stage
	// gating is in scope.)
default:
	// Unrecognized role values are rejected by every role filter and receive
	// no traffic; counting them anywhere dilutes the stage signal.
```

Longer term, deriving the partition from the acceptance sets of `NewPrefillRole`/`NewDecodeRole` would keep this switch from drifting from routing.

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -423,6 +447,34 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
 	return false
 }
 
+// partitionEndpoints classifies endpoints into prefill, decode, and interleaved buckets
+// based on the llm-d.ai/role pod label. Endpoints without a role label, without metadata,
+// or with an unrecognized role value default to the decode bucket for monolithic deployment safety.
+func partitionEndpoints(endpoints []fwkdl.Endpoint) (prefill, decode, interleaved []fwkdl.Endpoint) {
```
LukeAVanDrie, line 453: Non-blocking design question... The filters treat interleaved pods as candidates for both prefill and decode traffic. Partitioning them into their own bucket means:

- (a) hot dedicated-prefill pods gate all dispatch even when idle interleaved capacity exists that the scheduler would use for prefill work, and
- (b) a small hot interleaved partition gates a large healthy P/D fleet.

A partition mirroring routing (`S_prefill` over prefill + interleaved, `S_decode` over decode + interleaved) avoids both and reduces to identical behavior for pure P/D and monolithic fleets.

This is something I overlooked in the original issue spec, sorry

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -375,10 +377,32 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
...
+	saturation := -1.0
+	for _, part := range []struct {
+		name      string
+		endpoints []fwkdl.Endpoint
+	}{
+		{"prefill", prefillPool},
+		{"decode", decodePool},
+		{"interleaved", interleavedPool},
+	} {
+		if len(part.endpoints) == 0 {
```
LukeAVanDrie, line 392: nit: On the metric side, once a partition goes empty (prefill scaled to zero, pods relabeled), its `stage` series is never written again and freezes at the last value, so a dashboard would show stale prefill saturation for a fleet that no longer exists.

Non-blocking, but consider deleting the label values for empty partitions (a new `DeleteFlowControlPoolSaturation` helper in the metrics package).

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -375,10 +377,32 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
 
 	pool := p.endpointCandidates.Locate(ctx, nil)
 	p.poolEmpty = len(pool) == 0
-	saturation := p.saturationDetector.Saturation(ctx, pool)
 
-	// Record pool saturation metric
-	metrics.RecordFlowControlPoolSaturation(p.poolName, saturation)
+	prefillPool, decodePool, interleavedPool := partitionEndpoints(pool)
```
LukeAVanDrie, line 381: nit: This runs every ~1ms dispatch cycle and allocates up to three growing slices per call. Probably minor next to the existing per-cycle costs, but the FC benchmarks exist and are worth one run. If it shows up, scratch buffers on `Processor` should be safe since the dispatch loop is single-goroutine.

This can also be deferred into a separate perf PR if it's relevant at all.

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -375,10 +377,32 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
...
+		metrics.RecordFlowControlPoolSaturation(p.poolName, part.name, stageSat)
+		if stageSat > saturation {
+			saturation = stageSat
+		}
+	}
+	if saturation < 0 {
+		saturation = p.saturationDetector.Saturation(ctx, pool)
+	}
+
+	metrics.RecordFlowControlPoolSaturation(p.poolName, "effective", saturation)
```
LukeAVanDrie, line 405: nit: I think `effective` is the right label here, but the PR description and release note still say `"global"`. Can we update the release note before merge?

pkg/epp/flowcontrol/controller/internal/processor_test.go
```diff
@@ -1102,6 +1103,181 @@ func TestProcessor(t *testing.T) {
...
+				q := h.addQueue(testFlow)
+				require.NoError(t, q.Add(h.newTestItem("item-1", testFlow, testTTL)))
+
+				h.endpointCandidates.Candidates = []fwkdl.Endpoint{
+					makeEndpoint(bylabel.RolePrefill),
+					makeEndpoint(bylabel.RoleDecode),
+				}
+
+				// Prefill is healthy but decode is saturated.
+				h.saturationDetector.SaturationFunc = func(_ context.Context, endpoints []fwkdl.Endpoint) float64 {
```
LukeAVanDrie, line 1201: nit: This mock returns 1.0 whenever any endpoint in the call is decode, so the pre-change code (one `Saturation()` call on the full mixed pool) also returns 1.0 and the test also passes. As written it can't catch a regression to flat-pool gating. 

Suggest making the mock encode the motivating scenario: saturated only for a homogeneous prefill subset, healthy for any mixed set:

```go
h.saturationDetector.SaturationFunc = func(_ context.Context, endpoints []fwkdl.Endpoint) float64 {
		if ep.GetMetadata().Labels[bylabel.RoleLabel] != bylabel.RolePrefill {
			// Any mixed or decode set reads healthy: the flat average is
			// diluted by idle decode workers.
			return 0.3
		}
	}
	return 1.5 // homogeneous prefill subset: tier saturated
}
```

With this mock the old code dispatches and the new code gates.

pkg/epp/metrics/metrics.go
```diff
@@ -376,7 +376,7 @@ var (
 					"llm_d_epp_flow_control_stale_endpoints).",
 				compbasemetrics.ALPHA),
 		},
-		[]string{"inference_pool"},
+		[]string{"inference_pool", "stage"},
```
LukeAVanDrie, line 379: The gauge Help text (here and in llm_d_router_metrics.go) doesn't mention the new label. Worth a sentence enumerating the stage values (prefill, decode, interleaved, effective) and noting that effective is the gating signal. docs/metrics.md also still lists the labels as inference_pool only.

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -31,7 +31,9 @@ import (
 	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
 	"github.com/llm-d/llm-d-router/pkg/epp/flowcontrol/contracts"
 	"github.com/llm-d/llm-d-router/pkg/epp/flowcontrol/types"
+	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/flowcontrol"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/filter/bylabel"
```
LukeAVanDrie, line 36: This should be moved out to a neutral location IMO.

LukeAVanDrie (changes requested):
Thanks @loicmarchal! Have a few bits of feedback and some nits, but this overall looks correct.

Two design thoughts. Happy to take both to #1186 instead of blocking here:

- **Interleaved pods serve both stages.** Per the role filters they are prefill and decode candidates, so a partition that mirrors routing would be `S_prefill = detector(prefill + interleaved)` and `S_decode = detector(decode + interleaved)`. I missed this when I wrote the issue spec.
- **Where this lives long-term.** The inline shape in `dispatchCycle` is fine to land. For the agreed follow-up (projected decode load), I'd aim at a stage-aware detector implementation rather than more processor logic as the projection will need pod-level metrics (in-flight prefill, decode thresholds) that only a detector sees, and a detector that partitions internally can also record per-stage stale counts.

loicmarchal:
Thank you @LukeAVanDrie and sorry for the delay! 

> **Interleaved pods serve both stages.** Per the role filters they are prefill and decode candidates, so a partition that mirrors routing would be `S_prefill = detector(prefill + interleaved)` and `S_decode = detector(decode + interleaved)`. I missed this when I wrote the issue spec.

That's right! I updated how the endpoints are grouped in prefill/decode for the saturation level calculation.
 
> **Where this lives long-term.** The inline shape in `dispatchCycle` is fine to land. For the agreed follow-up (projected decode load), I'd aim at a stage-aware detector implementation rather than more processor logic as the projection will need pod-level metrics (in-flight prefill, decode thresholds) that only a detector sees, and a detector that partitions internally can also record per-stage stale counts.

Yes I agree, I started looking into possible solutions for the projected decode load. We can discuss it in #1186. We can go on with this implementation for now but we will need move this once we implement the projected decode saturation.

pkg/epp/flowcontrol/controller/internal/processor.go
loicmarchal: the call for the decode partition will indeed overwrite the gauge from the (prefill) stage call to `Saturation()`. We could start with a `stage` label on the stale gauge (as a quick fix?), but probably moving the partitioning into a stage-aware detector that owns per-stage stale accounting would be the way to go (also to solve the projected decode saturation)?

pkg/epp/flowcontrol/controller/internal/processor.go
loicmarchal: Good catch, I fixed that!

pkg/epp/flowcontrol/controller/internal/processor.go
loicmarchal: I added a `DeleteFlowControlPoolSaturation` helper in the metrics package and the processor calls it for empty partitions

pkg/epp/flowcontrol/controller/internal/processor.go
loicmarchal: I updated the PR description.

pkg/epp/metrics/metrics.go
loicmarchal: Good point! I added a mention of the stage values and "effective" being the gating signal.

pkg/epp/flowcontrol/controller/internal/processor.go
loicmarchal: Agreed, the flow-control -> scheduling import isn't great. The constants need a neutral location that both packages can import from. An option could be to use a small standalone package (e.g. `pkg/epp/common/roles`) with just the constants. Both `bylabel` and the processor could import from it.

Another option could be to move them to `framework/interface/datalayer` alongside `EndpointMetadata`. That would require some more refactoring.

What do you think?

loicmarchal:
Thank you @LukeAVanDrie, I've addressed your comments. When you have free cycles, could take another look?

pkg/epp/flowcontrol/controller/internal/processor.go
LukeAVanDrie: I'd defer to the stage-aware detector rather than take the stage-label quick fix. The recording happens inside `Detector.Saturation`, keyed only by detector name, and the detector has no stage identity. Adding a stage label means either changing the `SaturationDetector` interface or inferring the stage from call order, and both cost more than the detector follow-up we already agreed on for projected decode load. A detector that partitions internally can own per-stage stale accounting as part of the same work.

For the interim, two small things in this PR:

1. One sentence in `docs/metrics.md` and the gauge help text: `flow_control_stale_endpoints` carries no stage label and is written on every detector call, so it reflects the most recently evaluated stage; a reading of 0 does not rule out stale metrics in another stage.
2. A tracking issue for per-stage stale accounting, referenced from that sentence.

With those in place I'm happy to resolve the thread.

LukeAVanDrie: E.g.,

> `flow_control_stale_endpoints` carries no stage label and is written on every detector call, so it reflects the most recently evaluated stage. A reading of 0 does not rule out stale metrics in another stage; per-stage stale accounting is tracked in #NNNN.

pkg/epp/flowcontrol/controller/internal/processor.go
LukeAVanDrie: I'd go with the standalone constants package (`pkg/epp/common/roles` or similar). The move is mechanical and it survives the planned detector follow-up: a stage-aware detector plugin that imported `bylabel` would keep the cross-plugin dependency, while a neutral constants package serves both. The datalayer option pulls `EndpointMetadata` refactoring into scope for no gain here.

Want to create a followup issue so we don't forget this?

LukeAVanDrie: Or just open a PR (small enough that it may not warrant an issue)

LukeAVanDrie (dismissed):

LukeAVanDrie:
Thanks @loicmarchal, this LGTM. Only have nits which are fine deferring as we are going to eventually refactor this a bit as discussed.

You're PR body and release note both still describe an `interleaved` metric stage and `max(S_prefill, S_decode, S_interleaved)`. The implementation emits `prefill`, `decode`, and `effective`, with `effective = max(prefill, decode)`. The body also says unrecognized role values default to the decode bucket, and the code excludes them. The release note is user-facing generated content, so the text needs to match before merge.

LukeAVanDrie:
@ahg-g Handing this off to you. I have reviewed from a flow control perspective, but don't have code owners approval for every path. Generally could use another perspective on the PR strategy here too.

LukeAVanDrie:
@loicmarchal please sign the commits; I may have broken this when rebasing your changes onto main. If so, apologies!

loicmarchal:
> Only have nits which are fine deferring as we are going to eventually refactor this a bit as discussed.

We will need to refactor for the predicted decode saturation indeed. If it's ok to look at those with the refactoring, we can do it in a follow-up PR?

> You're PR body and release note both still describe an `interleaved` metric stage and `max(S_prefill, S_decode, S_interleaved)`. The implementation emits `prefill`, `decode`, and `effective`, with `effective = max(prefill, decode)`. The body also says unrecognized role values default to the decode bucket, and the code excludes them. The release note is user-facing generated content, so the text needs to match before merge.

I'm sorry, I forgot to adapt the PR description and release note completely after all the changes. Thank you for pointing it out! I updated it. Let me know if the new description is better.

pkg/epp/flowcontrol/controller/internal/processor.go
loicmarchal: Thank you, sounds good. I will look into it and open a PR. It should not create too much conflict (regardless of which PR is merged first)

pkg/epp/flowcontrol/controller/internal/processor.go
loicmarchal: I agree it's better to defer. I added the sentence in the doc and gauge helper and created an issue: https://github.com/llm-d/llm-d-router/issues/2475

LukeAVanDrie (approved):

ahg-g (approved):

PR #2230, opened 2026-07-29
feat(epp): add --tls-min-version and --tls-cipher-suites flags

### LINKED ISSUE #2239: Support TLS min-version and cipher-suites configuration on EPP gRPC server
## Problem

The EPP's ext_proc gRPC server currently has no way to configure the minimum TLS version or allowed cipher suites. Cluster administrators who need to enforce a TLS security profile (e.g., OpenShift's cluster-wide TLS profile) cannot restrict the EPP to specific TLS versions or cipher suites.

## Proposed Solution

Add two new CLI flags to the EPP:

- `--tls-min-version` — accepts Go `crypto/tls` version constant names (e.g., `VersionTLS12`, `VersionTLS13`)
- `--tls-cipher-suites` — accepts a comma-separated list of Go `crypto/tls` cipher suite names (e.g., `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256`)

The parsed values should be applied to the `tls.Config` used by the ext_proc gRPC server in both TLS config paths (static cert and cert-reload).

## Context

Companion issue in kserve: https://github.com/kserve/kserve/pull/5924 adds template variables to pass these flags from the KServe config system to the EPP container args.

andresllh:
**What type of PR is this?**
/kind feature

Adds two new CLI flags to the EPP (Endpoint Picker) for configuring TLS
minimum version and cipher suites on the ext_proc gRPC server:

- `--tls-min-version` — accepts Go `crypto/tls` version constant names
  (e.g., `VersionTLS12`, `VersionTLS13`)
- `--tls-cipher-suites` — accepts a comma-separated list of Go `crypto/tls`
  cipher suite names (e.g., `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256`).
  Only effective for TLS 1.2 and below; TLS 1.3 cipher suites are fixed
  by the Go standard library and not configurable via this flag.

These flags allow cluster administrators to enforce a TLS security profile
on the llm-d-router's gRPC server, aligning with platform-level TLS policies
(e.g., OpenShift's cluster-wide TLS profile).

The parsed values are applied to both TLS config paths (static cert and
cert-reload) via an extracted `applyTLSOverrides` method, with dedicated
unit tests in `tls_test.go`.

Companion PR: https://github.com/kserve/kserve/pull/5924 (adds template
variables to pass these flags from the KServe config system).

Fixes #2239

**Release note**:
```release-note
Added --tls-min-version and --tls-cipher-suites CLI flags to llm-d-router for configuring TLS security profile on the ext_proc gRPC server. Note: --tls-cipher-suites only applies to TLS 1.2 and below; TLS 1.3 cipher suites are fixed by Go's crypto/tls and not configurable.
```


pkg/epp/server/options.go
```diff
@@ -213,6 +217,10 @@ func (opts *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&opts.EnableGRPCStreamMetrics, "enable-grpc-stream-metrics", opts.EnableGRPCStreamMetrics,
 		"Enables ext_proc gRPC stream metrics (in-flight gauge, hold-duration histogram, completions counter by code).")
 	fs.BoolVar(&opts.SecureServing, "secure-serving", opts.SecureServing, "Enables secure serving.")
+	fs.StringVar(&opts.TLSMinVersion, "tls-min-version", opts.TLSMinVersion,
+		"Minimum TLS version for secure serving (e.g., VersionTLS12, VersionTLS13).")
+	fs.StringSliceVar(&opts.TLSCipherSuites, "tls-cipher-suites", opts.TLSCipherSuites,
+		"Comma-separated list of TLS cipher suites for secure serving (Go crypto/tls names, e.g., TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256).")
```
elevran: Go's crypto/tls only honors CipherSuites for TLS 1.2 and below; TLS 1.3 suites are fixed and not configurable here. If an admin sets --tls-min-version=VersionTLS12 alongside a restrictive --tls-cipher-suites list, a client can still negotiate TLS 1.3 with suites outside that list, so the cipher restriction is silently not enforced on 1.3 connections. 
- Worth a one-line callout in the help text given the stated goal of enforcing a platform TLS profile.
- Update to release-note in PR description to make this more explicit on release

zdtsw:
i think in general this will work, basically the idea is to put the responsiblity on the one who  creates router deployment to choose values that are both valid and compatible.

just food for thought:
- wrong TLS version fails fast (pod crash on startup), so that case is safe. 
- incompatible cipher is fuzzier, depends on whether the gateway is fail open or closed, and in the worst
case the request bypasses the router entirely and falls back to round-robin.

elevran (commented):
Clean, well-tested addition - thanks!
Minor doc update before this merges.

pkg/epp/server/options.go
andresllh: Done — updated both:

- **Flag help text**: `--tls-cipher-suites` now reads: "Only effective for TLS 1.2 and below; TLS 1.3 cipher suites are not configurable."
- **Release note**: Added the same caveat: "Note: --tls-cipher-suites only applies to TLS 1.2 and below; TLS 1.3 cipher suites are fixed by Go's crypto/tls and not configurable."

zdtsw (commented):
LGTM, thanks

elevran (approved):

PR #2233, opened 2026-07-30
fix(kvevents): attribute KV events to the subscriber serving endpoint

### LINKED ISSUE #2229: Precise index attributes KV events to the topic identity, not the publishing rank endpoint
Per-rank ZMQ subscribers know exactly which serving endpoint they represent (`ensureSubscriber` registers Address, Port, RankIndex per rank endpoint), but that identity is not propagated with received events:

- the subscriber forwards only `{Topic, Sequence, Payload}` ([zmq_subscriber.go:141-145](https://github.com/llm-d/llm-d-router/blob/main/pkg/kvevents/zmq_subscriber.go#L141-L145)),
- the vLLM adapter derives `podID` from the topic and drops the payload's `DataParallelRank` ([vllm_adapter.go:81-95](https://github.com/llm-d/llm-d-router/blob/main/pkg/kvevents/engineadapter/vllm_adapter.go#L81-L95)),
- the topic-derived identity is written into the index ([pool.go:327-358](https://github.com/llm-d/llm-d-router/blob/main/pkg/kvevents/pool.go#L327-L358)).

When multiple rank publishers share one topic (e.g. `kv@podIP:8000@model` for all eight ranks of a wide-EP pod), every rank's blocks are indexed under one endpoint. Precise scoring looks up per-rank candidate identities (`podIP:8003`), so the ranks that actually hold the blocks score zero, and `p2p-source-producer` cannot select the engine that holds the prefix.

The pool carries a related TODO: `// DP rank is the sentinel until PR #370 makes the index DP-aware` ([pool.go:341-343](https://github.com/llm-d/llm-d-router/blob/main/pkg/kvevents/pool.go#L341-L343)). The closed, unmerged [llm-d-kv-cache#370](https://github.com/llm-d/llm-d-kv-cache/pull/370) explored propagating `DataParallelRank` through `EventBatch`, `PodEntry`, and scoring. That broader numeric-rank propagation is not required to resolve this defect. Each per-rank subscriber already represents exactly one serving endpoint, so carrying its registered identity with each message provides the direct index attribution and isolates the event deduplication scope for that rank endpoint. Numeric `DataParallelRank` propagation remains separate future index metadata and diagnostics work.

## Acceptance

- [ ] A rank-3 event received via its registered subscriber is indexed under `podIP:8003`, not `podIP:8000`.
- [ ] Two same-pod publishers sharing a topic are not merged under one serving endpoint.
- [ ] Removing one rank endpoint clears only that endpoint's confirmed entries.
- [ ] A prefix present only on worker local rank 3 produces a precise score only for `workerIP:8003`.
- [ ] Event deduplication is isolated per registered rank endpoint; same-pod ranks do not share reference counts.

## Scope

Distinct from the P2P listener-addressing defect (#2227): that issue's deployment-side port compensation makes worker-pod publishers *reachable*, but does not resolve which endpoint their events are *attributed* to.

nilig:
**What type of PR is this?**
/kind bug

A pod running N DP ranks publishes KV events on N rank-specific ZMQ sockets that share one topic (`kv@podIP:8000@model`), and the index derives the holder identity from the topic, so all ranks' blocks collapse onto one endpoint entry. Precise scoring looks up per-rank identities (`podIP:8003`), so every rank except the topic one scores zero and `p2p-source-producer` cannot select the engine that holds the prefix.

Each per-rank subscriber already represents exactly one serving endpoint (`ensureSubscriber` registers `Address`, `Port`, `RankIndex`), so the fix binds that identity to every message it forwards: `kvevents.RawMessage.SourceEndpoint` carries the subscriber's serving endpoint, `Pool` uses it as the KV-index identity and the ordered-sharding key, and `SubscriberManager` recreates a subscriber when either its transport or source endpoint changes. The topic-derived identity remains the fallback for global-socket subscribers (`ZMQEndpoint` mode), which pass an empty source endpoint.

The identity string is `fmt.Sprintf("%s:%s", Address, Port)`, matching the read-side convention in `utils.go` and `prerequest.go` on any address family.

Fixes #2229

**Test plan**:
- [x] Two subscribers sharing one publisher topic index their blocks under two distinct serving endpoints (real-ZMQ regression, `-race`)
- [x] Source-endpoint propagation through subscriber, pool sharding, and index identity; topic fallback when no source endpoint is set
- [x] Live on a GLM-5.2 wide-EP cell (2 pods x 8 ranks per role): a prefix seeded only on worker global rank 11 produced `scores:{"<workerIP>:8003": 45}` (45/45 block keys), `p2p-source-producer` emitted that endpoint, and the routed request completed with a verified P2P pull (consumer `kv_offload` load = tokens x 92.6 KB/token) in 2.9 s

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
@@ -73,12 +73,13 @@ func (p *Producer) ensureSubscriber(ctx context.Context, meta *fwkdl.EndpointMet
 	endpointKey := meta.NamespacedName.String()
 	port := p.kvEventsConfig.PodDiscoveryConfig.SocketPort + meta.GetRankIndex()
 	zmqEndpoint := fmt.Sprintf("tcp://%s:%d", meta.Address, port)
+	sourceEndpoint := fmt.Sprintf("%s:%s", meta.Address, meta.Port)
 
 	logger := log.FromContext(ctx).WithName(p.typedName.String())
 	// subscriberCtx is plugin-lifetime; caller ctx would tear subscribers
 	// down on request completion.
 	if err := p.subscribersManager.EnsureSubscriber(p.subscriberCtx, endpointKey,
-		zmqEndpoint, p.kvEventsConfig.TopicFilter, true); err != nil {
+		sourceEndpoint, zmqEndpoint, p.kvEventsConfig.TopicFilter, true); err != nil {
 		logger.Error(err, "Failed to ensure KV-events subscriber for endpoint",
 			"endpoint", endpointKey, "address", meta.Address)
 		return fmt.Errorf("ensure subscriber for %s: %w", endpointKey, err)
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor_test.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor_test.go
@@ -196,6 +196,32 @@ func TestProducer_ExtractEndpoint_OffsetsZMQPortByRankIndex(t *testing.T) {
 	}
 }
 
+func TestProducer_EnsureSubscriber_PassesServingEndpoint(t *testing.T) {
+	cfg := kvevents.DefaultConfig()
+	cfg.DiscoverPods = true
+	cfg.PodDiscoveryConfig = kvevents.DefaultPodReconcilerConfig()
+	cfg.PodDiscoveryConfig.SocketPort = 5557
+
+	subscribers := &fakeSubscriberManager{}
+	p := &Producer{
+		typedName:          plugin.TypedName{Type: PluginType, Name: PluginType},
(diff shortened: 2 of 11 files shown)


roytman:
nit: "dedup scope carries the real rank instead of the sentinel" from #2229 is not addressed. pool.go:381 and 513, still hardcode `dataParallelRank: noDataParallelRank`, and event_dedup_filter.go still carries the pre-existing TODO(#370) for this. 
Practically, this stopped mattering as a correctness issue, because `blockScope.podIdentifier `is now the rank-specific SourceEndpoint — two ranks on the same pod no longer share a dedup scope regardless of the sentinel.

nilig:
You're right that the `dataParallelRank` field still contains the sentinel value. This PR separates ranks in a different way: each rank's `SourceEndpoint` is used as the deduplication identity. Therefore, two ranks on the same pod no longer share deduplication state.

I updated #2229 to describe that actual requirement. I'll also clarify the stale code comments. I don't think we should add the broader rank propagation from #370 to this PR, since it would not change the behavior of this path.

vMaroon (approved):

PR #2234, opened 2026-07-30
fix(sidecar): rank-endpoint P2P pulls and wide-EP port-base compensation

### LINKED ISSUE #2227: Wide-EP deployments must compensate vLLM's global DP-rank P2P port offset
vLLM binds each `OffloadingConnector` P2P listener at:

```text
configured base port + global data-parallel rank
```

The routing sidecar derives the source rank from its serving-port offset, which is pod-local, and addresses:

```text
--p2p-connector-port + pod-local rank
```

These conventions are equivalent when each pod is an independent DP group. They diverge on multi-pod DP groups such as LWS wide-EP when every pod receives the same configured P2P base.

vLLM's current behavior is explicit in [`P2PSecondaryTierManager`](https://github.com/vllm-project/vllm/blob/5fa01544480b60275edf913e1a620d229e40553d/vllm/v1/kv_offload/tiering/p2p/manager.py#L268-L272):

```python
dp_index = offloading_spec.config.parallel.data_parallel_index
port = int(port) + dp_index
```

The sidecar derives a pod-local rank and returns `base + rank` in [`p2pPortFor`](https://github.com/llm-d/llm-d-router/blob/aced3336c1db2ea446bc48f57e488e7ec25bd0dc/pkg/sidecar/proxy/connector_p2p.go#L251-L286).

## Reproduction

Measured on a 2-pod x DP8 LWS group with `data_parallel_size=16`:

| Pod | Local ranks | Global ranks | Flat base | vLLM listeners |
|---|---:|---:|---:|---:|
| Leader | 0-7 | 0-7 | 7777 | 7777-7784 |
| Worker | 0-7 | 8-15 | 7777 | 7785-7792 |

For worker local rank 3/global rank 11:

```text
Sidecar address: 7777 + 3  = 7780
vLLM listener:   7777 + 11 = 7788
```

In the observed configuration with `kv_load_failure_policy: fail`, the connection to the non-listening port stalled until the 600-second client timeout.

Test configuration:

- Model: `zai-org/GLM-5.2-FP8`
- One prefill and one decode LWS group
- One leader and one worker pod per group
- Eight local DP ranks per pod
- Prefix seeded on worker local rank 3/global rank 11
- Prefix length: 24,576 tokens
- Expected transfer: 2,276.4 MB
- Prefix fresh to the consumer
- Counter: `kv_offload_total_bytes_total{transfer_type="CPU_to_GPU"}`

| Deployment | Worker configured base | vLLM listener | Sidecar address | Bytes | Result |
|---|---:|---:|---:|---:|---|
| Flat base | 7777 | 7788 | 7780 | 0.0 MB | Client timeout |
| Compensated base | 7769 | 7780 | 7780 | 2,276.4 MB | Success |

## Resolution

Normalize the actual P2P listener range on every pod by compensating the configured vLLM base with that pod's global starting rank:

```bash
START_RANK=$(( ${LWS_WORKER_INDEX:-0} * DP_SIZE_LOCAL ))
P2P_BASE=$(( 7777 - START_RANK ))
```

Supply `P2P_BASE` as `secondary_tiers[].port` in the vLLM `OffloadingConnector` configuration.

vLLM then calculates:

```text
actual port
= configured base + global rank
= 7777 - START_RANK + START_RANK + local rank
= 7777 + local rank
```

For 2 pods x DP8:

| Pod | `START_RANK` | Configured base | Actual listeners |
|---|---:|---:|---:|
| Leader | 0 | 7777 | 7777-7784 |
| Worker | 8 | 7769 | 7777-7784 |

The sidecar remains configured with:

```text
--p2p-connector-port=7777
```

`--data-parallel-start-rank` is unchanged, so the DP16 topology and wide-EP collectives are unaffected.

The compensated deployment transferred the expected 2,276.4 MB through the stock EPP and sidecar. No global-rank header or router code change is required for this deployment.

## Acceptance

- [x] A worker source succeeds through the stock routing path with a compensated base.
- [x] The consumer receives 2,276.4 MB for the 24,576-token prefix.
- [x] The wide-EP manifests calculate `P2P_BASE` for every role serving P2P pulls.
- [x] Every pod exposes the same actual local listener range.
- [x] The wide-EP guide documents the compensation and required NetworkPolicy range.

## Scope

This issue covers P2P listener addressing only.

KV-event listener compensation uses the same arithmetic, but precise-index endpoint attribution is separate: the current subscriber path can index events using the shared topic endpoint rather than the exact per-rank serving endpoint. That behavior is not resolved by the P2P deployment validation here.

Separately, vLLM should fail a connection to a non-listening P2P port within a bounded interval and apply the configured failure policy instead of leaving the request stalled until the client timeout.

nilig:
**What type of PR is this?**
/kind bug
/kind documentation

Two changes that make the P2P source pull work on multi-rank pods:

1. The self-pull guards compared only hosts, so two DP ranks on the same pod (`10.0.6.107:8003` vs `10.0.6.107:8007`) were treated as one engine and the pull was skipped exactly where it is needed - `addP2PPullToPrefill` on the prefill leg, and `decodeWithP2PSource` on the decoder-only path (which collapsed identity to `POD_IP`). Both guards now compare full serving endpoints through a scheme-agnostic `normalizeEndpoint`: a same-pod cross-rank source is a valid peer and injects, while the selected rank's own endpoint still skips. The injected `remote_kv_source` params derive from the normalized endpoint, so scheme-qualified sources produce a bare `remote_host` and the rank-correct `remote_port`.

2. `docs/disaggregation.md` documented multi-pod DP groups (LWS wide-EP) as not covered. vLLM binds the P2P and KV-events listeners at `configured base + global data_parallel_index`, while the router addresses `pod IP + pod-local rank`; the docs now state the compensation contract that reconciles the two - each pod subtracts its global start rank from both configured bases (`P2P_BASE=$((7777 - START_RANK))`, `KV_EVENTS_BASE=$((5557 - START_RANK))`) - with the LWS formulas and the resulting per-pod port ranges. The base-port formulas themselves belong in deployment sources (guides), not router code.

Fixes #2227

**Test plan**:
- [x] Prefill and decoder-only guards skip same-endpoint pulls and inject `remote_kv_source` for same-pod different-rank sources; scheme-qualified sources produce the correct `remote_host` and rank-specific `remote_port` (ginkgo, `-race`)
- [x] Live on a GLM-5.2 wide-EP cell with compensated bases: every pod binds P2P `7777-7784` / KV events `5557-5564` / serving `8000-8007`; a same-pod cross-rank pull (global rank 11 -> 14) completed end to end - source accept on the rank-correct port, one consumer load of tokens x 92.6 KB/token, HTTP 200 in 2.9 s

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/sidecar/proxy/connector_p2p.go
+++ b/pkg/sidecar/proxy/connector_p2p.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
 	"strconv"
 	"strings"
@@ -232,11 +233,33 @@ func (s *Server) p2pPullAvailable() bool {
 // routes it to the OffloadingConnector and the NIXL fields to the
 // NixlConnector.
 func (s *Server) addP2PPullToPrefill(prefillKVParams map[string]any, kvCacheSource, prefillPodHostPort string) {
-	source, _ := strings.CutPrefix(kvCacheSource, "http://")
-	prefiller, _ := strings.CutPrefix(prefillPodHostPort, "http://")
+	source := normalizeEndpoint(kvCacheSource)
+	prefiller := normalizeEndpoint(prefillPodHostPort)
 	if source != "" && source != prefiller {
 		prefillKVParams[requestFieldRemoteKVSource] = s.p2pSourceParams(kvCacheSource)
 	}
+}
+
+// normalizeEndpoint canonicalizes an endpoint for equality comparison.
+// Scheme-qualified values (any scheme, not only http) are parsed as URLs
+// and reduced to their host:port; bare values are treated as host:port
+// directly. Both forms are re-joined through the net package so IPv6
+// literals compare consistently; values without a parseable port are
+// returned scheme-stripped, matching extractHost's fallback.
+func normalizeEndpoint(s string) string {
+	hostport := s
+	if scheme, rest, ok := strings.Cut(s, "://"); ok && scheme != "" {
+		if u, err := url.Parse(s); err == nil && u.Host != "" {
+			hostport = u.Host
+		} else {
+			hostport = rest
+		}
+	}
+	host, port, err := net.SplitHostPort(hostport)
+	if err != nil {
+		return hostport
+	}
+	return net.JoinHostPort(host, port)
 }
 
 // p2pSourceParams builds the kv_transfer_params.remote_kv_source block for a
--- a/pkg/sidecar/proxy/connector_p2p.go
+++ b/pkg/sidecar/proxy/connector_p2p.go
@@ -226,13 +226,15 @@ func (s *Server) p2pPullAvailable() bool {
 // addP2PPullToPrefill adds the OffloadingConnector P2P pull block to a prefill
 // leg's kv_transfer_params so the prefiller pulls cached prefix from
 // kvCacheSource while keeping its own computed blocks available for the
-// decoder. It is a no-op when no source is set or the source resolves to the
-// prefiller itself, since there is nothing to pull from oneself. The
+// decoder. It is a no-op when no source is set or the source is the selected
+// prefill endpoint, since there is nothing to pull from oneself. The
 // remote_kv_source key composes with NIXL params: vLLM's MultiConnector
 // routes it to the OffloadingConnector and the NIXL fields to the
 // NixlConnector.
 func (s *Server) addP2PPullToPrefill(prefillKVParams map[string]any, kvCacheSource, prefillPodHostPort string) {
-	if kvCacheSource != "" && extractHost(kvCacheSource) != extractHost(prefillPodHostPort) {
+	source, _ := strings.CutPrefix(kvCacheSource, "http://")
+	prefiller, _ := strings.CutPrefix(prefillPodHostPort, "http://")
+	if source != "" && source != prefiller {
 		prefillKVParams[requestFieldRemoteKVSource] = s.p2pSourceParams(kvCacheSource)
 	}
 }
--- a/pkg/sidecar/proxy/connector_p2p.go
+++ b/pkg/sidecar/proxy/connector_p2p.go
@@ -236,7 +236,7 @@ func (s *Server) addP2PPullToPrefill(prefillKVParams map[string]any, kvCacheSour
 	source := normalizeEndpoint(kvCacheSource)
 	prefiller := normalizeEndpoint(prefillPodHostPort)
 	if source != "" && source != prefiller {
-		prefillKVParams[requestFieldRemoteKVSource] = s.p2pSourceParams(kvCacheSource)
+		prefillKVParams[requestFieldRemoteKVSource] = s.p2pSourceParams(source)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


roytman (dismissed):

pkg/sidecar/proxy/connector_p2p.go
```diff
@@ -226,13 +226,15 @@ func (s *Server) p2pPullAvailable() bool {
...
-// prefiller itself, since there is nothing to pull from oneself. The
+// decoder. It is a no-op when no source is set or the source is the selected
+// prefill endpoint, since there is nothing to pull from oneself. The
 // remote_kv_source key composes with NIXL params: vLLM's MultiConnector
 // routes it to the OffloadingConnector and the NIXL fields to the
 // NixlConnector.
 func (s *Server) addP2PPullToPrefill(prefillKVParams map[string]any, kvCacheSource, prefillPodHostPort string) {
-	if kvCacheSource != "" && extractHost(kvCacheSource) != extractHost(prefillPodHostPort) {
+	source, _ := strings.CutPrefix(kvCacheSource, "http://")
+	prefiller, _ := strings.CutPrefix(prefillPodHostPort, "http://")
```
elevran: nit: 
only `http://` is stripped, Not an issue today since header producers (disagg_headers_handler.go, disagg_profile_handler.go, p2psource/producer.go) always emit plain `ip:port` via net.JoinHostPort, but if that ever changes the same-endpoint guard would silently stop matching. Consider reusing `extractHost` for the comparison instead of a local CutPrefix, for consistency with the rest of the file.

pkg/sidecar/proxy/connector_p2p_test.go
```diff
@@ -199,3 +199,20 @@ var _ = Describe("p2pSourceParams", func() {
 		Expect(params[requestFieldRemotePort]).To(Equal(7779))
 	})
 })
+
+var _ = Describe("addP2PPullToPrefill", func() {
+	It("injects a pull when source and prefiller are different ranks on the same pod", func() {
```
elevran, line 204: nit: this covers the **cross-rank** injection case, but there is no explicit unit test asserting the **same-rank**/same-endpoint case still skips. The test plan describes it as covered, but the only place it is implicitly exercised is the pre-existing full-flow test, which does **not** set a KV cache source at all. A short second It block mirroring this one but with equal host:port would close that gap.

elevran (commented):
Verified the fix locally - checked out the branch and ran the full pkg/sidecar/proxy suite plus a couple of scratch tests for edge cases; all pass. Good, minimal fix that matches the issue. Two minor notes below, neither blocking.

pkg/sidecar/proxy/connector_p2p.go
nilig: `extractHost` drops the port, so reusing it for this comparison would make `10.0.6.107:8003` and `10.0.6.107:8007` compare equal again - the exact bug this PR fixes. The guard needs the full serving endpoint. Every producer of both headers emits plain `host:port` via `net.JoinHostPort` (`disagg_headers_handler.go`, `p2psource/producer.go`), so the `http://` trim keeps the sidecar's existing tolerance without discarding the rank-bearing port. I'd keep the comparison as is.

pkg/sidecar/proxy/connector_p2p_test.go
nilig: Agreed - the cross-rank test proves injection but nothing pins the same-endpoint skip directly. Added an explicit equal `host:port` case asserting `remote_kv_source` is absent in bba473bb.

pkg/sidecar/proxy/connector_p2p.go
elevran: I see, thanks.
My concern was that stripping just `http` would fail if using TLS on the communication or any another protocol. Right now either works since everything is just ip:port without a protocol spec but this is error prone if/when we add a protocol specification to the values.
If you really care just for the host and port, you can use `url.Parse()` to handle the protocol, then `net.SplitHostPort()` to isolate the server and the port.

nilig: Makes sense. The comparison has to keep the full `host:port` - the port identifies the rank - so I'll add a small normalization helper: parse a scheme-qualified URL when one is present (`url.Parse` handles `https://` and anything else), fall back to treating the value as plain `host:port`, and compare the normalized endpoints. The same-host/different-port regression case stays in the tests, plus a scheme-variant case. Would that address your concern?

elevran: that works, thanks

elevran (approved):

pkg/sidecar/proxy/connector_p2p.go
nilig: Done - `normalizeEndpoint` parses a scheme-qualified URL when one is present and falls back to the raw `host:port`; both self-pull guards (prefill leg and decoder-only path) compare through it, and the injected params derive from the normalized endpoint as of `d7e63d29`.

PR #2237, opened 2026-07-30
feat(kvcache): add opt-in xxhash hashing to the KV-block token processor

vMaroon:
**What type of PR is this?**
/kind feature

The CBOR+FNV-64a block-hash chain costs ~193us and 3142 allocs per 10k-token prompt on the request path. Request-side block keys are internal-only: engine hashes are opaque to the index and the kvevents pool re-derives request keys through the same `chunkedTokenDatabase`, so the algorithm only has to agree between the request path and event ingestion within one EPP. `tokenProcessorConfig.hashAlgorithm: xxhash` selects an xxhash chain at ~23.7us and 12 allocs (~8x cheaper), with token IDs encoded little-endian so keys are architecture-independent for Redis-shared indexes. The default (`cbor-fnv`) stays byte-identical, so this PR changes no behavior on its own.

Like `hashSeed`, the algorithm must stay fixed across the index's lifetime: changing it while a persisted/shared index holds keys, or while engines hold blocks indexed under the old keys, mixes algorithms in parent-chain resolution and produces request keys lookups never match. The field doc states the flush-and-reset requirement. Flipping the default is deliberately left to a follow-up with a release note.

**Release note**:
```release-note
The precise-prefix-cache token processor accepts tokenProcessorConfig.hashAlgorithm: "xxhash" for ~8x cheaper request-path block hashing; the default "cbor-fnv" behavior is unchanged.
```

**Test plan**:
- [x] xxhash chain determinism, per-model and per-seed differentiation, MMHash taint, parent-key continuation, partial-block exclusion, unknown-algorithm rejection
- [x] benchmarks covering both algorithms


sagearc (approved):
lgtm

pkg/kvcache/kvblock/token_processor.go
```diff
@@ -157,19 +198,60 @@ func (db *chunkedTokenDatabase) hash(parent uint64, tokens []uint32, extra inter
...
-		if extraFeatures[i] != nil {
-			extra = extraFeatures[i].MMHashes
+		if digest != nil {
+			var extras []MMHash
+			if extraFeatures[i] != nil {
+				extras = extraFeatures[i].MMHashes
+			}
+			prefix = hashXX(digest, prefix, chunk, extras)
+		} else {
+			var extra interface{}
```
sagearc: why the drift? can we use `var extras []MMHash` here too?

sagearc (dismissed):

liu-cong (dismissed):

pkg/kvcache/kvblock/token_processor.go
```diff
@@ -31,6 +33,17 @@ import (
...
 
+// Accepted TokenProcessorConfig.HashAlgorithm values.
+const (
+	// HashAlgorithmCBORFNV hashes each block with FNV-64a over the canonical
+	// CBOR encoding of [parent, tokens, extra]. This is the default.
+	HashAlgorithmCBORFNV = "cbor-fnv"
+	// HashAlgorithmXXHash hashes each block with a single xxhash digest over
+	// the parent hash (8 bytes little-endian), the raw token-ID bytes, and
+	// each extra-feature string prefixed with its 8-byte little-endian length.
+	HashAlgorithmXXHash = "xxhash"
```
yankay: Just a small thought: since `xxHash` includes several algorithms, would `xxh64` be a more precise name here?

hexfusion:
cc @liu-cong @sagearc PTAL

sagearc (dismissed):

liu-cong (dismissed):

ahg-g (dismissed):

pkg/kvcache/kvblock/token_processor.go
```diff
@@ -201,16 +276,21 @@ func (db *chunkedTokenDatabase) TokensToKVBlockKeys(
 	parentKey BlockHash, tokens []uint32, modelName string,
 	extraFeatures []*BlockExtraFeatures,
 ) ([]BlockHash, error) {
+	chunks := db.chunkTokens(tokens)
+	if len(chunks) == 0 {
+		return nil, nil
+	}
+
+	var digest *xxhash.Digest // one per call: the processor is shared across goroutines
```
vMaroon:
```suggestion
	var digest *xxhash.Digest // one per call
```

vMaroon: done to trigger stuck tests - couldnt see UI retest options. Pls approve again (cc @ahg-g @sagearc)

ahg-g:
is this ready?

ahg-g (approved):

PR #2238, opened 2026-07-30
New session_id_header for session affinity scorer plugin

### LINKED ISSUE #2019: Evolving session affinity
## Background

We currently have session affinity filter/scorer implemented as a `encoded_endpoint_header` algo: the router encodes the picked endpoint to the client for the first request, and then the client can encode the same endpoint in subsequent requests in a header to request the affinity. This implementation is stateless but requires client side orchestration.  Another useful implementation is client simply provides a `session_id` in the header or cookie, which also aligns with agentic applications.

We also have an [agentidentity](https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/plugins/requestcontrol/preadmitter/agentidentity) plugin which parses session id info from well-known agent harnesses. Currently it's used to set the flow control fairness id, but it's important that we can feed it to the session affinity filter/scorer. 
 
## Requirement

1. In addition to the existing `encoded_endpoint_header`  algo, as well as `session_id_header` algo. Configurable by the user.
1. We should be able to bridge the agent identity plugin and the session affinity. If session id header is not provided, the session id can be derived from the well known agent harnesses. I recommend keeping them separate plugins, but have the agent identity plugin also produce a session id attribute.

cc @ahg-g @roytman @elevran

D-Sai-Venkatesh:
**What type of PR is this?**

/kind feature

Adds a `session_id_header` algorithm to the `session-affinity-scorer` and `session-affinity-filter`, picked with a new `strategy` param next to the existing default `encoded_endpoint_header`. Both algorithms now sit behind a small internal `strategy` interface (`score`/`preRequest`/`responseHeader`), so the plugin just delegates.

The session ID comes from a request header, or from the attribute set by the agent-identity plugin. The scorer remembers which pod served that session and gives highest score for the same, dropping the mapping after a TTL if the session goes idle.

- New session lands on the candidate pod with the fewest bound sessions.
- Bound session stays on its pod while that pod is still a candidate.
- Bound session whose pod dropped out of the candidate set migrates to the present pod with the fewest bound sessions, on the next request.
- Bindings unused past the TTL are swept in the background.

Fixes #2019

```release-note
Adds a new `session_id` session affinity strategy that pins a client-supplied session id (from a header or request attribute) to a pod, alongside the existing `encoded_endpoint_header`. Session affinity plugins (`session-affinity-scorer`, `session-affinity-filter`) now take per-strategy config blocks: `encodedEndpointHeaderConfig` (with `header`) and `sessionIdConfig` (with `sources` and eviction settings).
```


D-Sai-Venkatesh:
Here is an initial result from our early evaluation:
image

Setup:
- **Model**: `nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-FP8`
- **Decode deployment**: 3 replicas (decode only), TP 4, H100s
- **Traces**: `Exgentic/agent-llm-traces`
- **Static EPP scorers**: queue-scorer (weight: 2), kv-cache-utilization-scorer (weight: 2), prefix-cache-scorer(weight: 3, only in precise-prefix scenario), session-affinity-scorer (weight: 3, only in session-affinity scenario)

D-Sai-Venkatesh:
Will push one more commit which will move the `session_id_header` algorithm to util folder and it will be consumed by both session affinity scorer and filter.

D-Sai-Venkatesh:
Hey @liu-cong, just pushed a new commit to move the core session_id_header algorithm into a shared type in util/sessionaffinity so the scorer and filter both use it instead of each keeping a copy. Also added session_id_header to the filter. Let me know if any changes are required.

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/session_affinity.go
```diff
@@ -31,56 +31,122 @@ import (
...
+	// which resends it on subsequent requests.
+	StrategyEncodedEndpointHeader = "encoded_endpoint_header"
+
+	// StrategySessionIDHeader maps an opaque client-supplied session identifier
+	// to a pod.
+	StrategySessionIDHeader = "session_id_header"
 )
 
 // parameters configures the SessionAffinity filter.
 type parameters struct {
```
liu-cong, line 45: I suggest the following refactor. 
1. Separate config struct for different strategy instead of a flat structure. This is clean and easy to extend.
2. For the SessionIDStrategy, we should not by default look at the agent identity attribute. Session affinity plugins should be a self contained plugin and should not be tied to other plugins. We can let the user configure it explicitly.
3. I propose we support multiple "session id sources" in priority order. This is very flexible - it allows the user to specify a header, and fallback to the agent id attribute for agent traffic.

```
type parameters struct {
	// Strategy is StrategyEncodedEndpointHeader (the default) or
	// StrategySessionID.
	Strategy string `json:"strategy"`
	// Only one of the strategies should be configured
    EncordedEndpointHeaderConfig EncordedEndpointHeaderConfig
    SessionIDConfig SessionIDConfig
}

// Shared between filter and corer, in util/sessionaffinity
type EncordedEndpointHeaderConfig struct {
    // Name of the http header
    // defaults to `x-session-token`
  	Header string `json:"header"`
}

type SessionIDConfig struct {
    // If no sources are configured, defaults to http header `x-session-id`
    // If multiple sources are configured, they are evaluated in priority order
    Sources []SessionIDSource `json:"sources"`
   	// EvictionTTLSeconds is how long a session binding survives unused.
	// session_id_header only.
	EvictionTTLSeconds float64 `json:"evictionTtlSeconds"`
	// EvictionSweepSeconds is how often expired bindings are swept.
	// session_id_header only.
	EvictionSweepSeconds float64 `json:"evictionSweepSeconds"`
}

type SessionIDSource struct {
   // Only one of the following sources can be configured.
   // Name of the http header
   Header string `json:"header"`
   // Name of the request attribute populated by upstream plugins
   Attribute string `json:"attribute"`
}
```

D-Sai-Venkatesh: Thanks, implemented the above (per-strategy config, ordered sources, no implicit agent-identity fallback). The config types and resolution logic live in util/sessionaffinity for consumption by both the scorer and filter.

pkg/epp/framework/plugins/scheduling/util/sessionaffinity/session_id.go
```diff
@@ -0,0 +1,426 @@
...
+			target = endpoint
+		}
+	}
+
+	if podName != "" {
+		s.pluginState.Write(request.RequestID, s.pluginKey, &boundPodPresence{present: target != nil})
+	}
+
+	if target == nil {
+		target = s.leastLoadedPod(endpoints)
```
liu-cong: I don't think we should make any additional decision here if we cannot find a pod with bounded session. We should simply return error/false and let the consumer decide what to do.

pkg/epp/framework/plugins/scheduling/filter/sessionaffinity/session_affinity.go
```diff
@@ -31,56 +31,113 @@ import (
...
 	// SessionAffinityType is the type of the SessionAffinity filter.
 	SessionAffinityType = "session-affinity-filter"
+
+	// StrategyEncodedEndpointHeader echoes the picked pod back to the client,
+	// which resends it on subsequent requests.
+	StrategyEncodedEndpointHeader = "encoded_endpoint_header"
+
+	// StrategySessionIDHeader maps an opaque client-supplied session identifier
+	// to a pod.
+	StrategySessionIDHeader = "session_id_header"
```
liu-cong: Let's rename to StrategySessionID as it can also consume the request session id attribute.

D-Sai-Venkatesh: Renamed

pkg/epp/framework/plugins/scheduling/util/sessionaffinity/session_id.go
D-Sai-Venkatesh: Hey @liu-cong, removed the `leastLoadedPod`. Choose now returns (nil, false) when there's no bound pod present (unbound or bound-but-absent), so pod selection in those cases depends entirely on the other consumers.

liu-cong (approved):

D-Sai-Venkatesh:
Hey @liu-cong, thank you very much for your comments and the review.

Issue #2174, opened 2026-07-24
inflight-load-producer: release prefill request counter at start-of-stream


ruocco:
With the default producer config (`addEstimatedOutputTokens=false`), the `inflight-load-producer` releases only token counters at `StartOfStream`; the **request** counter for every profile, including prefill, is held until `EndOfStream`.

Under P/D disaggregation this misrepresents prefill pods. A prefill endpoint finishes its physical work at TTFT and its scraped Waiting/Running drop, but its `InFlightLoad.Requests` stays elevated for the whole decode duration of every request whose prefill it served.

Consequences:

- **Utilization detector scrape-lag credit (#2066):** the credit `max(0, InFlightRequests - (WaitingQueueSize + RunningRequestsSize))` is inflated on prefill endpoints, roughly by the number of requests currently decoding whose prefill they served. This bias scales with pool throughput rather than being a small constant, and can saturate `qRatio` under sustained decode load. #2066 documents this as an assumption (credit assumes aggregated / non-P/D pools) rather than fixing it.
- **Concurrency detector:** it also sees phantom in-flight requests on prefill pods for the same reason.

Proposal: release the prefill profile's request counter at `StartOfStream` in both config paths, since prefill is physically complete at first token. The `addEstimatedOutputTokens=true` path already does this via `release()`.

Context: #2066 (see the discussion on `pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/detector.go`), and the capacity-ledger proposal in #2061.

LukeAVanDrie:
@kaushikmitr this can be closed now, right?

PR #2243, opened 2026-07-30
fix(epp): release prefill request count at StartOfStream in inflight load producer

kaushikmitr:
With addEstimatedOutputTokens=false, the prefill profile's request counter was held until EndOfStream, overstating prefill endpoint concurrency for the entire decode phase. The first response chunk means prefill has completed and handed off, so release the prefill entry in full at StartOfStream, matching the addEstimatedOutputTokens=true path.

No current guide or well-lit path is affected: none of the shipped P/D configurations filter or score on the active request count for the prefill profile. The fix matters for deployments that add a utilization filter based on active requests on prefill endpoints, without it, the filter would act on request counts that stay inflated for the full decode duration.

**What type of PR is this?**

Fixes #
https://github.com/llm-d/llm-d-router/issues/2174
```release-note
`InFlightLoad.Requests` on prefill endpoints now drops at first token instead of staying elevated through decode. No shipped well-lit path filters on it but any deployment with a custom config that scores or filters on active request count for the prefill profile sees different behavior, and the debug state endpoint reports different numbers.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
@@ -423,7 +423,10 @@ func (p *InFlightLoadProducer) ResponseBody(
 	// the prompt cost, which is consumed by prefill. As soon as the first chunk
 	// arrives (StartOfStream), prefill is done across all profiles, so free the
 	// token counters for every targeted endpoint regardless of profile name.
-	// Request counters are still released on EndOfStream below via PluginState.Delete.
+	// The prefill profile's entry is released in full (request counter included):
+	// the first chunk means the prefill worker has finished and handed off, so
+	// the request is no longer in flight on that endpoint. Other profiles'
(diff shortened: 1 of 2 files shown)


LukeAVanDrie (approved):
Let's add a short release note to the PR description since `InFlightLoad.Requests` on prefill endpoints now drops at first token instead of staying elevated through decode. No shipped well-lit path filters on it but any deployment with a custom config that scores or filters on active request count for the prefill profile sees different behavior, and the debug state endpoint reports different numbers.

Otherwise, this LGTM!

kaushikmitr:
> Let's add a short release note to the PR description since `InFlightLoad.Requests` on prefill endpoints now drops at first token instead of staying elevated through decode. No shipped well-lit path filters on it but any deployment with a custom config that scores or filters on active request count for the prefill profile sees different behavior, and the debug state endpoint reports different numbers.
> 
> Otherwise, this LGTM!

done

ahg-g (approved):

PR #2248, opened 2026-07-31
fix(observability): inject trace_id/span_id into request loggers

### LINKED ISSUE #1630: Logs are not correlated with traces: trace_id/span_id never injected into the logger
/kind feature
/label observability

## Description

The router emits OpenTelemetry spans and structured zap (logr) logs, but the two are never correlated: no code path injects `trace_id`/`span_id` into the logger, and the zap setup has no automatic correlation hook. When debugging a request, there is no way to jump from a log line to the corresponding trace, or to find the logs belonging to a given trace.

## Current behavior

In the EPP entry point (`pkg/epp/handlers/server.go`), the logger is enriched with the request ID and stored via `log.IntoContext` before the `gateway.request` span is started; after `tracer.Start`, the logger in the context is never updated, so all downstream logs carry only the request ID. In the sidecar entry point (`pkg/sidecar/proxy/chat_completions.go`), the `forward_request` span only sets span attributes and the proxy logger is unaware of the span context. The global zap configuration (`pkg/common/observability/logging/logger.go`) is a plain JSON encoder with no otel correlation, so there is no fallback either. The existing `SpanContext()` calls in the connectors are used for span timing attributes, not log correlation.

umakantv:
**What type of PR is this?**
/kind bug

The router emits OpenTelemetry spans and structured zap/logr logs, but the two were never correlated: no code path injected `trace_id`/`span_id` into the logger. Debugging a request required guessing which log lines belonged to a given trace.

This adds a small `logging.WithTrace` helper and wires it at the request entry points:

- **EPP** (`pkg/epp/handlers/server.go`): after starting the `gateway.request` span, enrich the request logger and store it via `log.IntoContext` so all downstream `log.FromContext` call sites pick up `trace_id`/`span_id`.
- **Sidecar** (`pkg/sidecar/proxy/chat_completions.go`): after starting `forward_request`, derive a request-scoped logger with trace fields, put it in the request context, and use it for handler logs.

Field names follow the OpenTelemetry log/trace correlation convention (`trace_id`, `span_id` as hex strings). When no valid span is present, the logger is left unchanged.

Connector call sites that still use the process-level `s.logger` are unchanged; call sites with a request context can use `log.FromContext` to pick up correlation.

Fixes #1630

**Special notes for your reviewer**:
- Nested spans intentionally keep the request-root `span_id` injected at the entry point; filtering by `trace_id` covers the full request.
- Host Go unit tests for the changed packages pass (`logging`, `sidecar/proxy`, `epp/handlers`). Full `make presubmit` requires the builder container/Docker, which was not available in this environment.

```release-note
Structured logs from the EPP and sidecar request entry points now include OpenTelemetry `trace_id` and `span_id` fields for log-trace correlation.
```

## Test plan
- [x] Unit tests for `logging.WithTrace` (valid span, no span, invalid span)
- [x] `go test -race` for `./pkg/common/observability/logging/`, `./pkg/sidecar/proxy/`, `./pkg/epp/handlers/`


gyliu513:
@umakantv can you help fix the CI issue?

umakantv:
Alright, I will check

umakantv:
@gyliu513 tests are taking forever to run

umakantv:
@gyliu513 I think some workflows need approval

gyliu513:
@umakantv https://github.com/llm-d/llm-d-router/pull/2271#issuecomment-5281598182

PR #2252, opened 2026-08-01
perf: Reduce allocations on the response-body streaming path

albertoperdomo2:
**What type of PR is this?**

/kind bug

The EPP response-body hot path allocates on every streaming chunk for operations that are either unnecessary at that phase or can be simply avoided. Under sustained load (~40 req/s, streaming, `gpt-oss-120b`) GC dominates the CPU profile (`lfstack.pop`, `gcDrainN`, and `getempty` together consume over 50% of total CPU time). 

This PR targets the three cheapest wins on that path:

1. Per-phase metadata extraction instead of every-request: `ExtractMetadataValues` was called unconditionally before the phase switch, allocating a map on every message including intermediate response-body chunks that never use metadata. Moved the call into the three cases that consume it (request headers, end-of-stream body, response headers) so intermediate chunks pay nothing.

2. Guarded TRACE logging with `.Enabled()` checks: `HandleResponseBody` and `runResponseBodyPlugins` are called per chunk. Unconditional `logger.V(TRACE).Info(...)` calls allocate zap encoder clones even when TRACE is disabled. Added `.Enabled()` guards on the per-chunk path.

3. Cached `plugin.TypedName()` into a local variable: Each `run*Plugins` loop called `plugin.TypedName()` 2-4 times per plugin per request (for logging and metrics). `TypedName()` is an interface method returning a two-string struct; caching it in a local avoids repeated virtual dispatch and copies.

**Test plan**: All existing and new added tests pass. 

Profiles (30s window, ~1,090-1,170 requests, openai/gpt-oss-120b) collected on the same node under the same load (~40 req/s steady-state). PRE uses the upstream v0.9.0 image; POST uses the `perf/reduce-response-streaming-allocs` build.

*Runtime metrics (go_memstats, 30s delta):*

| Metric | PRE (v0.9.0) | POST (`perf/reduce-response-streaming-allocs`) | Delta |
|--------|-------------|---------------|-------|
| Bytes allocated | 8.02 GB | 5.24 GB | **-34.7%** |
| Malloc count | 91.5M | 72.8M | **-20.5%** |
| GC cycles | 43 | 23 | **-46.5%** |
| GC wall time | 43.8 ms | 21.8 ms | **-50.1%** |
| Total CPU (all cores) | 243.1 s | 144.2 s | **-40.7%** |

*Response-body path allocations (pprof alloc_objects, focused on HandleResponseBody + related):*

| Function | PRE (v0.9.0) | POST (`perf/reduce-response-streaming-allocs`) | Delta |
|----------|-----|------|-------|
| `ExtractMetadataValues` (flat) | 1,518,320 | 0 | **-100%** |
| `Director.HandleResponseBody` (cum) | 17,110,843 | 0 | **-100%** |
| `HandleResponseBody` full path (cum) | 26,797,938 | 8,598,881 | **-67.9%** |
| `zapLogger.WithValues` (flat) | 4,718,807 | 1,693,089 | **-64.1%** |

*CPU breakdown: GC + memory management (`lfstack.pop`, `gcDrainN`, `getempty`, `gcDrain`):*

| | PRE (v0.9.0) | POST (`perf/reduce-response-streaming-allocs`) | Delta |
|---|-----|------|-------|
| GC-related CPU | 103.1 s | 17.3 s | **-83.2%** |

The response-body path fires per chunk (high frequency), so small per-call savings compound across millions of chunks per window. The net effect is an 83% reduction in GC CPU and 46.5% fewer GC cycles under comparable load.

`NONE`

### DIFF
--- a/pkg/common/envoy/metadata.go
+++ b/pkg/common/envoy/metadata.go
@@ -21,11 +21,12 @@ import (
 )
 
 func ExtractMetadataValues(req *extProcPb.ProcessingRequest) map[string]any {
-	metadata := make(map[string]any)
-	if req != nil && req.MetadataContext != nil && req.MetadataContext.FilterMetadata != nil {
-		for key, val := range req.MetadataContext.FilterMetadata {
-			metadata[key] = val.AsMap()
-		}
+	if req == nil || req.MetadataContext == nil || len(req.MetadataContext.FilterMetadata) == 0 {
+		return nil
+	}
+	metadata := make(map[string]any, len(req.MetadataContext.FilterMetadata))
+	for key, val := range req.MetadataContext.FilterMetadata {
+		metadata[key] = val.AsMap()
 	}
 	return metadata
 }
--- a/pkg/common/envoy/metadata_test.go
+++ b/pkg/common/envoy/metadata_test.go
@@ -39,12 +39,35 @@ func TestExtractMetadataValues(t *testing.T) {
 
 	tests := []struct {
 		name     string
-		metadata map[string]*structpb.Struct
+		req      *extProcPb.ProcessingRequest
 		expected map[string]any
 	}{
 		{
-			name:     "Exact match",
-			metadata: makeFilterMetadata(),
+			name:     "Nil request",
+			req:      nil,
+			expected: nil,
+		},
+		{
+			name:     "Nil MetadataContext",
+			req:      &extProcPb.ProcessingRequest{},
+			expected: nil,
+		},
+		{
+			name: "Empty FilterMetadata",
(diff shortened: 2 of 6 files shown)


elevran:
can you please fix failing CICD 🙏
pkg/epp/requestcontrol/director_bench_test.go:59:
`EndpointMetadata` no longer has a `NamespacedName` field after the merge with main (it was renamed to `ID`, which the rest of this branch already uses). This is the cause of the failing lint/typecheck job. Same fix needed at line 108.

liu-cong (approved):

PR #2253, opened 2026-08-01
test(coordinator): rewire e2e onto shared inference-gateway and add 3-EPP topology

revit13:
**What type of PR is this?**
/kind cleanup
/kind test

Reworks the coordinator e2e suite to exercise the disaggregation topologies the deployment guide (llm-d/llm-d#2119) actually ships (single-EPP and 3-EPP deployments), and drops the coordinator's private copy of the gateway manifests.

- **Use the shared inference-gateway component.** The suite loads EPP, InferencePool, RBAC, ServiceAccounts, and Services from `deploy/components/inference-gateway/` instead of the coordinator-local `epd-pool/` manifests. As in the router e2e, the coordinator fronts the EPP with a hand-rolled Envoy, so that component's `gateways.yaml`/`httproutes.yaml` stay unused.

- **Consolidate the coordinator manifests.** Flattens `deploy/coordinator/components/coordinator/` up to `deploy/coordinator/` and deletes the now-redundant manifests it replaced.

- **Add an optional 3-EPP topology (`E2E_EPP_TOPOLOGY=3epp`).** Runs one role-scoped EPP and InferencePool per phase (encode / prefill / decode), with an Envoy route that dispatches each `EPP-Profile` to the matching role. Each EPP sees only its own role's pods, so it uses `single-profile-handler` and no role filter, with per-role scorers mirroring the guide. The default single-EPP topology is unchanged.

- **Extract common resources.** `shared-envoy-resources.yaml` holds the Envoy Deployment and Service; per-topology manifests carry only the routing ConfigMap. Topology branching lives in `eppsToCreate()` / `poolNames()`.

- **Deploy the renderer like the router e2e.** `vllm-render` runs as a standalone Deployment/Service, reached through the substituted `${VLLM_RENDER_URL}`.

- **CI.** `e2e-tests` becomes a `topology: [single, 3epp]` matrix, aggregated by an `e2e-tests-status` job (required checks can't reference individual matrix legs).

Follow-up: a later PR will add parallelization to coordinator e2e tests, as in the router.

Fixes #

```release-note
NONE
```


test/coordinator/e2e/coordinator/e2e_suite_test.go
```diff
@@ -78,10 +103,15 @@ var (
...
+	// threeEPP selects the 3-EPP topology (one role-scoped EPP + InferencePool per
+	// phase) instead of the default single-EPP topology. See envoy3EPPManifest and
+	// the eppConfig{Encode,Prefill,Decode} configs.
+	threeEPP = env.GetEnvString("E2E_EPP_TOPOLOGY", "single", ginkgo.GinkgoLogr) == "3epp"
+
 	containerRuntime = env.GetEnvString("CONTAINER_RUNTIME", "docker", ginkgo.GinkgoLogr)
 	eppImage         = env.GetEnvString("EPP_IMAGE", "ghcr.io/llm-d/llm-d-router-endpoint-picker:dev", ginkgo.GinkgoLogr)
 	vllmSimImage     = env.GetEnvString("VLLM_IMAGE", "ghcr.io/llm-d/llm-d-inference-sim:v0.10.2", ginkgo.GinkgoLogr)
-	vllmRenderImage  = env.GetEnvString("VLLM_RENDER_IMAGE", vllmSimImage, ginkgo.GinkgoLogr)
+	vllmRenderImage  = env.GetEnvString("VLLM_RENDER_IMAGE", "vllm/vllm-openai-cpu:v0.21.0", ginkgo.GinkgoLogr)
```
roytman, line 114: Why did you change the simulator image to the real vLLM one? Was it part of the test, or was it done intentionally? 
Contradicts README.coord.md:139 and the Makefile comment, and the manifest still passes simulator-only flags (--mode=echo, --force-dummy-tokenizer). CI masks it because make always overrides the env var; a bare go test would break.

roytman (commented):
The new 3-EPP topology has zero assertions that verify per-role routing actually works. Workers run in echo mode and are role-agnostic, so if envoy-3-epp.yaml's EPP-Profile dispatch sent an encode request to the decode EPP, the pool's own role-scoped selector would still return a plausible 200 — nothing would catch it. getPodNames is only used for replica-count assertions, never to confirm which pod served which leg. This is exactly the scenario the PR's stated goal needs covered.

.github/workflows/ci-coordinator.yaml
roytman, line 37: The filter, doesn't include `deploy/**`, so the e2e matrix legs silently skip when the shared manifests the suite now depends on (deploy/components/inference-gateway/*) change. Pre-existing gap, but this PR makes it materially worse since the suite now reads that shared directory.

test/coordinator/e2e/coordinator/setup_test.go
```diff
@@ -80,35 +92,89 @@ func createCRDs() {
...
 
-// createInferencePool creates the InferencePool covering all three worker
-// roles. When toDelete is set, the existing pool is removed first so the test
-// starts clean.
+// renameEPPConfigVolume retargets the EPP Deployment's config volume, volume
+// mount, and ConfigMap reference from the shared "epp-config" name to cmName so
+// each 3-EPP role mounts its own ConfigMap. It matches only the "name: epp-config"
+// line endings, leaving the "/etc/epp/epp-config.yaml" mount path and ConfigMap
+// key (still keyed "epp-config.yaml") untouched.
+func renameEPPConfigVolume(docs []string, cmName string) []string {
```
roytman, line 142: The method does a string-replace with no count assertion, if the shared manifest's formatting ever changes, it silently no-ops and all three 3-EPP EPPs would mount the wrong config, with the suite likely still green.

deploy/environments/dev/coordinator-e2e-infra/inference-pools-3-epp.yaml
```diff
@@ -0,0 +1,55 @@
+# Role-scoped InferencePools for the coordinator 3-EPP e2e topology. Each pool
+# selects a single worker role via the llm-d.ai/role label (all three roles share
+# the same app label, so role is what distinguishes them) and is backed by that
+# role's own EPP Service. Mirrors the guide's "3 separate EPPs" option, where each
```
roytman: which guide?

test/coordinator/e2e/coordinator/configs_test.go
```diff
@@ -117,3 +117,64 @@ schedulingProfiles:
     weight: 1
   - pluginRef: max-score-picker
 `
+
+// The 3-EPP topology (E2E_EPP_TOPOLOGY=3epp) runs one role-scoped EPP per phase,
+// each backing its own role-scoped InferencePool. Because a role-scoped EPP only
+// ever sees its own role's pods, it needs no role filter and no profile-selection
+// plugin: single-profile-handler picks the one "default" profile, and the Envoy
+// route (envoy-3-epp.yaml) is what dispatches each EPP-Profile value to the right
+// EPP. The plugin/scorer choices mirror the guide's "3 separate EPPs" values
```
roytman: which guide, `router/coord-disaggregation-{encode,prefill,decode}.values.yaml)` ?

test/coordinator/e2e/coordinator/configs_test.go
```diff
@@ -117,3 +117,64 @@ schedulingProfiles:
...
+schedulingProfiles:
+- name: default
+  plugins:
+  - pluginRef: active-request-scorer
+  - pluginRef: max-score-picker
+`
+
+// eppConfigPrefill keeps prefix groups on cache-warm pods
+// (prefix-cache-affinity-filter), then picks by queued prefill token load
+// (token-load-scorer), matching the guide's prefill values file.
```
roytman: guide?

test/coordinator/e2e/coordinator/e2e_suite_test.go
```diff
@@ -78,10 +103,15 @@ var (
 	keepClusterOnFailure = env.GetEnvBool("E2E_KEEP_CLUSTER_ON_FAILURE", false, ginkgo.GinkgoLogr)
 	printLogs            = env.GetEnvBool("E2E_PRINT_LOGS", false, ginkgo.GinkgoLogr)
 
+	// threeEPP selects the 3-EPP topology (one role-scoped EPP + InferencePool per
+	// phase) instead of the default single-EPP topology. See envoy3EPPManifest and
+	// the eppConfig{Encode,Prefill,Decode} configs.
```
roytman: the real constant name for Encode and Decode is different

test/coordinator/e2e/coordinator/e2e_suite_test.go
revit13: It aligns with the router e2e tests, which use the same default value ([test/e2e/e2e_suite_test.go:79](vscode-webview://1d6klgfkjg8qepkfrtk9fq823ol53h4ligmg6sm3ktd99g08r7nf/test/e2e/e2e_suite_test.go#L79)). It's just the Go fallback default and has no effect in practice: the simulator render image overrides it (Makefile.coord.mk exports VLLM_RENDER_IMAGE=$(VLLM_IMAGE), and the manifest passes the sim-only flags --mode=echo --force-dummy-tokenizer), so the render pod always runs the simulator.  Happy to change it if you'd prefer.

roytman (approved):

elevran (approved):

PR #2259, opened 2026-08-03
Plugin: add request-control Screener extension point

hasB4K:
**What type of PR is this?**

/kind feature

Adds a `Screener` request-control extension point for preliminary filtering of located endpoint candidates before data producers, admission plugins, and scheduling profiles.

Most endpoint-selection plugins should implement a scheduling `Filter`. A developer should implement a `Screener` only for a mandatory constraint that must apply globally to every scheduling profile.

Each screener receives an independent copy of the same located endpoint pool. The framework intersects their returned subsets, so composition is independent of plugin registration order.

```text
Locate
  -> Screeners
  -> DataProducers
  -> admission plugins
  -> scheduling profiles
```

This extracts only the generic plugin mechanism from #2141. It does not include the disaggregation implementation or rollout fixtures.

Related to #2141.

**Testing**:

- [x] `go test ./pkg/epp/requestcontrol -count=1`
- [x] `make lint`

```release-note
NONE
```

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -60,16 +60,21 @@ The design enables:
 
 See the upstream [Request Scheduler](https://github.com/llm-d/llm-d/blob/main/docs/architecture/core/router/epp/scheduling.md) doc for the canonical scheduling model.
 
-1. **Filtering**
+1. **Pre-scheduling candidate filtering**
+   - Global `PreSchedulingCandidateFilter` plugins receive the same located endpoint set
+   - Their returned subsets are intersected before data producers and scheduling profiles run
+   - Use this stage for mandatory routing constraints that every profile must observe
+
+2. **Filtering**
    - Pods in an `InferencePool` go through a sequential chain of filters
    - Pods may be excluded based on criteria like model compatibility, resource usage, or custom logic
 
-2. **Scoring**
+3. **Scoring**
    - Filtered pods are scored using a weighted set of scorers
    - Scorers currently run sequentially (future: parallel execution)
    - Scorers access a shared datastore populated by the data layer
 
-3. **Pod Selection**
+4. **Pod Selection**
    - The highest-scored pod is selected
    - If multiple pods share the same score, one is selected at random
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -66,9 +66,10 @@ Request control runs once per request before any scheduling profiles:
 
 1. Request headers are processed and flow-control admission completes
 2. Endpoint candidates are located
-3. Global `PreSchedulingCandidateFilter` plugins receive independent copies of the same located endpoint set
-   - Their returned subsets are intersected
-   - Use this stage for mandatory routing constraints that every profile must observe
+3. Global `Screener` plugins perform preliminary filtering of located endpoints
+   - Each screener receives an independent copy of the same endpoint set, and their returned subsets are intersected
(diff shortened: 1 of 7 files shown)


docs/architecture.md
```diff
@@ -60,16 +60,21 @@ The design enables:
 
 See the upstream [Request Scheduler](https://github.com/llm-d/llm-d/blob/main/docs/architecture/core/router/epp/scheduling.md) doc for the canonical scheduling model.
 
-1. **Filtering**
+1. **Pre-scheduling candidate filtering**
```
elevran: nit: 
The new `PreSchedulingCandidateFilter` step is a global, once-per-request, director-level extension point that runs before `DataProducer` and `Admission` plugins. Steps 2–4 (Filtering/Scoring/Pod Selection) describe stages inside a single `Scheduling` profile, which can run multiple times per request (e.g., two profiles for disaggregated P/D). 
Numbering the new global filter as "1" in the same list implies it's the same kind of step as "2. Filtering," but it isn't, and the doc still says nothing about where Admission/DataProducer fit in the overall flow relative to it. As written, a reader could think step 1 flows directly into step 2 with nothing else in between, which is inaccurate.
Perhaps add explicit "Request Control" and "Scheduling" sub-sections?

pkg/epp/requestcontrol/director.go
```diff
@@ -305,6 +305,13 @@ func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestCo
 	}
 
 	snapshotOfCandidatePods := d.toSchedulerEndpoints(endpointCandidates)
+	snapshotOfCandidatePods = d.runPreSchedulingCandidateFilters(ctx, reqCtx.SchedulingRequest, snapshotOfCandidatePods)
+	if len(snapshotOfCandidatePods) == 0 {
```
elevran, line 310: The new `ServiceUnavailable` path (i.e. filters eliminate all candidates) is only exercised via the internal `runPreSchedulingCandidateFilters` helper test, never through `Director.HandleRequest` itself (director_test.go's TestDirector_HandleRequest table has no case for this). Please add an end-to-end test case for this path.

docs/architecture.md
hasB4K: Agreed, it might create some confusion. Changing!

ahg-g: Copying from https://github.com/llm-d/llm-d-router/pull/2141#discussion_r3711375473


Using filter is going to collide with the filter extension point during scheduling. To clearly distinguish between the two, I asked gemini (see below) for suggestions and it recommended to use the term Screener/Screen ( `Screener` for the interface and `Screen` for the function name). I agree with this suggestion and I think it makes it easier to document and name associated plugins (`disagg-rollout-screener`).

Prompt:
`
in llm-d-router we have an interface named filter that filters out candidates during request scheduling, we are adding another interface that does filtering before request scheduling, suggest a name for the interface such that it doesn't collide with the existing filter interface, as a suggestion, use a synonym to the word filter
`
Image

hasB4K: Renamed!

ahg-g:
/lgtm

Beautiful, thanks a lot!

ahg-g:
Leaving the approve to Etai in case he has further comments.

elevran (approved):
thanks @hasB4K, looks good.
Can now rebase #2141 on top of this.

Issue #2264, opened 2026-08-04
Propagate EPP-resolved priority to native backend schedulers (SGLang/vLLM)


sywangyi:
### What
Directly inject the EPP-resolved request `priority` into the outbound request body's `"priority"` field, so the backend's own native priority scheduler (SGLang's `--enable-priority-scheduling` or vLLM's `--scheduling-policy priority`) can use it — no pd-sidecar involvement required.

### Why
SGLang and vLLM disagree on the direction of the `priority` field:
- SGLang: **higher** integer = more urgent.
- vLLM: **lower** integer = more urgent.

Our `InferenceObjective` convention (e.g. `app-critical=100`, `app-batch=-10`) follows SGLang's "higher is more urgent" direction. `director.go` detects the target backend's engine type via the `llm-d.ai/engine-type` Pod label and negates the priority value when routing to a vLLM backend, so relative ordering (critical < normal < batch) comes out correct on both engines.

### Changes
- `pkg/epp/requestcontrol/director.go`: adds `injectPriorityIntoBody()` (writes `reqCtx.Priority` into the request payload map) and `isVLLMBackend()` (reads the scheduling result's primary target endpoint's `llm-d.ai/engine-type` label), invoked from `HandleRequest()` after `prepareRequest()`.
- `pkg/epp/requestcontrol/director_test.go`: updated `wantMutatedBody` expectations to include the injected `priority` field; adds coverage for the new injection path.

### Validation
Benchmarked end-to-end on a 14B model (Qwen2.5-14B-Instruct) in A100 with a mixed-priority workload (critical/normal/batch traffic classes, equal-length prompts to isolate the priority effect from prompt-length confounds) against both backends. Each backend was A/B tested using a clean single-commit diff (feature commit vs its immediate parent) to isolate priority injection as the sole variable, with 2 independent trials per leg to confirm reproducibility.

**SGLang** (with-priority vs baseline):

| Class | Metric | Baseline | With-Priority | Δ |
|---|---|---|---|---|
| critical | ttft p99 | 8910.6 ms | 2738.3 ms | **-69.3%** |
| normal | ttft p99 | 8773.4 ms | 9105.0 ms | +3.8% |
| batch | ttft p99 | 9674.8 ms | 14503.2 ms | +49.9% |

**vLLM** (with-priority vs baseline):

| Class | Metric | Baseline | With-Priority  | Δ |
|---|---|---|---|---|
[...]

PR #2265, opened 2026-08-04
Propagate EPP-resolved priority to native backend schedulers (SGLang/…

sywangyi:
…vLLM)

What
Directly injects the EPP-resolved request priority into the outbound request body's "priority" field, so the backend's own native priority scheduler (SGLang's --enable-priority-scheduling or vLLM's --scheduling-policy priority) can use it — no pd-sidecar involvement required.

Why
SGLang and vLLM disagree on the direction of the priority field:

SGLang: higher integer = more urgent.
vLLM: lower integer = more urgent.
Our InferenceObjective convention (e.g. app-critical=100, app-batch=-10) follows SGLang's "higher is more urgent" direction. director.go now detects the target backend's engine type via the llm-d.ai/engine-type Pod label and negates the priority value when routing to a vLLM backend, so relative ordering (critical < normal < batch) comes out correct on both engines.

**What type of PR is this?**

Fixes #

```release-note
NONE
```


sywangyi:
this pr could also be used to work with sglang Priority-Based KV Cache Eviction

pkg/epp/requestcontrol/director.go
```diff
@@ -357,12 +357,77 @@ func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestCo
...
 		return reqCtx, err
 	}
+	d.injectPriorityIntoBody(reqCtx, inferenceRequestBody, result)
 	if err := d.repackage(ctx, reqCtx, inferenceRequestBody); err != nil {
 		return reqCtx, err
 	}
 	return reqCtx, nil
 }
 
+// engineTypeLabelKey is the Pod label used to identify the inference engine
```
ahg-g: We should implement this as part of the parser plugin, not hardcode it here, see how we did this for the model name re-write.

sywangyi: hi @ahg-g thanks for the review，I have updated the PR according to your suggestion

xiaojun-zhang:
This PR is to resolve issue #2264

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
```diff
@@ -144,6 +145,40 @@ func (p *OpenAIParser) RewriteModelName(payload fwkrh.MarshalablePayload, model
 	return m, nil
 }
 
+// Engine type labels used by endpoint metadata to identify backend priority semantics.
+const (
+	engineTypeLabelKey       = "llm-d.ai/engine-type"
+	legacyEngineTypeLabelKey = "inference.networking.k8s.io/engine-type"
```
ahg-g: No need to define this legacy engine type, it is deprecated.

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
```diff
@@ -144,6 +145,40 @@ func (p *OpenAIParser) RewriteModelName(payload fwkrh.MarshalablePayload, model
 	return m, nil
 }
 
+// Engine type labels used by endpoint metadata to identify backend priority semantics.
+const (
+	engineTypeLabelKey       = "llm-d.ai/engine-type"
```
ahg-g: as a followup, can we please move this definition to a common place under framework/plugins/constants.go

pkg/epp/requestcontrol/director.go
```diff
@@ -399,6 +394,27 @@ func (d *Director) modelRewriteIfNeeded(ctx context.Context, reqCtx *handlers.Re
...
+	if !ok {
+		logger.V(logutil.DEBUG).Info("parser does not implement PriorityRewriter, skipping priority rewrite")
+		return nil
+	}
+	payload, ok := inferenceRequestBody.Payload.(fwkrh.MarshalablePayload)
+	if !ok {
+		logger.V(logutil.DEBUG).Info("payload does not implement MarshalablePayload, skipping priority rewrite")
+		return nil
+	}
+	mutated, err := rewriter.RewritePriority(payload, reqCtx.Priority, fwkrh.PriorityRewriteContext{TargetEndpoint: reqCtx.TargetPod})
```
ahg-g: Here we always override the priority with the one we read from InferenceObjective, shouldn't we make this an opt-in choice?

pkg/epp/framework/interface/requesthandling/plugins.go
```diff
@@ -59,6 +60,20 @@ type ModelNameRewriter interface {
...
+type PriorityRewriteContext struct {
+	TargetEndpoint *fwkdl.EndpointMetadata
+}
+
+// PriorityRewriter is implemented by parsers whose forwarded body can carry a
+// backend-native priority field.
+type PriorityRewriter interface {
+	// RewritePriority writes priority into the payload and returns it. Taking and
+	// returning a MarshalablePayload guarantees the result is repackageable.
+	RewritePriority(payload MarshalablePayload, priority int, ctx PriorityRewriteContext) (MarshalablePayload, error)
```
zetxqx: Super nit: we should make ctx the first parameter making this go idiomatic.

pkg/epp/requestcontrol/director.go
zetxqx: @ahg-g @sywangyi  where do you think is the better place to configure this as an opt-in feature?

I can think of several options:

Option 1: add a parameter to the parser plugin 
```
  plugins:
  - type: openai-parser
    parameters:
      propagatePriority: true   # default false
```
* Pros: parser is closest place to the original request.
* Cons: I checked almost all post requests vllm and sglang accepts has a priority field. then that means we need to set propagatePriority to true to all kinds of parsers.

Option2:  Add a field in the InferenceObjective CRD
```
  spec:
    priority: 100
    propagatePriority: true   # default false
```
* Pros: we are putting the priority mutation in one place for both flow control and the model server's own priority policy. When a request is not matching an inferenceObjective or infenceObjective is simply not set we can `strip` the priority field in the original request.
* Cons: CRD changes, may need more coordination.

Option3: add command line arg to EPP config (worst)

--propagatePriority: true # default false


I think a bigger problem the PR revealed is that people can actually set the priority in their request and currently llm-d(epp) just **passthrough** it to the model server and the priority field may take effect when the model server arg is set with `--scheduling-policy priority`. I think we can use this opportunity to always mutate to delete the `priority field` in the original request when `propagatePriority` is set to false. WDYT?

sywangyi: I agree the passthrough behavior is a real problem. If EPP is the authority for request priority, client-supplied `priority` should not be allowed to directly influence the backend scheduler.

My only concern is tying that behavior implicitly to `propagatePriority=false`. That flag sounds like “do not inject EPP priority”, while stripping client priority is a separate policy decision: passthrough vs strip vs overwrite.

Maybe we should model this explicitly as a priority field policy, for example:
- passthrough: preserve the original request
- strip: remove client-supplied priority
- propagate: overwrite with the EPP-resolved priority

Then the default can be `strip` if we want the safer behavior, and users must explicitly opt into `passthrough` or `propagate`.

zetxqx: I'd actually go one step simpler: client-supplied `priority` should never reach the backend scheduler, unconditionally. We can always use the `InferenceObjective` is the single source of truth for priority,
  so stripping isn't a policy choice users pick, it should EPP enforces. That means we don't need a three-value policy: EPP always removes the client-supplied field, and `propagatePriority` just controls whether EPP injects the InferenceObjective-resolved value or not. Any thoughts? @ahg-g

sywangyi:
@ahg-g and @zetxqx I have updated the PR according to your review comment, thanks

pkg/epp/framework/plugins/requesthandling/parsers/openai/README.md
```diff
@@ -9,7 +9,9 @@ Parses HTTP/H2C requests and responses in the OpenAI API format.
 
 Supports all standard OpenAI-compatible endpoints: completions, chat/completions, conversations, responses, embeddings, and images/generations. The fields parsed out vary by endpoint: the request's input content (prompt, messages, or input), the streaming mode, and token usage from responses that report it.
 
-**Parameters:** None.
+**Parameters:**
+
+- `propagatePriority` (bool, default: `false`): When enabled, injects the EPP-resolved request priority into the outgoing request body's `priority` field. Client-supplied `priority` is always removed first so backend-native priority scheduling is governed by EPP.
```
ahg-g: I don't follow why we would remove the priority always. This is not a backward compatible change btw.

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
```diff
@@ -146,6 +162,41 @@ func (p *OpenAIParser) RewriteModelName(payload fwkrh.MarshalablePayload, model
...
+	m, ok := payload.(fwkrh.PayloadMap)
+	if !ok {
+		return payload, nil
+	}
+	delete(m, "priority")
+	if !p.propagatePriority {
+		return m, nil
+	}
+	if isVLLMTarget(ctx) {
+		priority = -priority
```
ahg-g: pls document why we are doing this

ahg-g:
What about the other parsers? messages api? tokenized requests?

pkg/epp/framework/plugins/requesthandling/parsers/openai/README.md
zetxqx: We have a previous discussion: https://github.com/llm-d/llm-d-router/pull/2265#discussion_r3817687963

Basically, do we want the end user who sends the request know what priority they should set based on the model engine type? If we don't want the end user know that level information we need to always strip the priority otherwise the request can be misconfigured and if EPP blindly forward it to model server.

In the implementation of this PR the priority is always derived from the inferenceObjective configuration.

pkg/epp/requestcontrol/director.go
```diff
@@ -347,6 +350,34 @@ func (d *Director) modelRewriteIfNeeded(ctx context.Context, reqCtx *handlers.Re
...
+		return nil
+	}
+	priorPayloadMap, _ := payload.AsMap()
+	_, hadClientPriority := priorPayloadMap["priority"]
+	mutated, err := rewriter.RewritePriority(fwkrh.PriorityRewriteContext{TargetEndpoint: reqCtx.TargetPod}, payload, reqCtx.Priority)
+	if err != nil {
+		return err
+	}
+	mutatedPayloadMap, _ := mutated.AsMap()
+	_, hasBackendPriority := mutatedPayloadMap["priority"]
```
zetxqx: I think we can add a output parameter `mutated bool` in rewritePriority

```
type PriorityRewriter interface {
	RewritePriority(ctx PriorityRewriteContext, payload MarshalablePayload, priority int) (MarshalablePayload, bool, error)
}
```

In this way, we don't need to check asMap and priority here in the director

zetxqx (commented):
The CI is failing as well

sywangyi:
> What about the other parsers? messages api? tokenized requests?

Good point — I went through all the parsers. Priority propagation writes a top-level JSON priority field, so it only applies to parsers whose wire body is open JSON. I update PR to add support for sglanghttp and anthropic as well. for those use protobuf (vertexai, vllmgrpc, passthrough) can't carry a priority field without an upstream proto change,=>intentionally untouched

sywangyi:
> The CI is failing as well

fixed the failure

pkg/epp/framework/plugins/requesthandling/parsers/openai/README.md
ahg-g: > Basically, do we want the end user who sends the request know what priority they should set based on the model engine type? If we don't want the end user know that level information we need to always strip the priority otherwise the request can be misconfigured and if EPP blindly forward it to model server.

A couple of points:

1) I strongly recommend to avoid unnecessary and frequent request body mutations, we should avoid that wherever possible.

2) Missing up with the request is a surprising behavior that the user can't even turn off, I think a higher level component should enforce policies like this and we need to avoid mutating the request body without the user explicitly opting into this behavior.

Also, this flag should be a general switch, we don't need each parser to have its own kill switch for the feature.

pkg/epp/requestcontrol/director.go
```diff
@@ -347,6 +350,35 @@ func (d *Director) modelRewriteIfNeeded(ctx context.Context, reqCtx *handlers.Re
...
+	if !ok {
+		logger.V(logutil.DEBUG).Info("payload does not implement MarshalablePayload, skipping priority rewrite")
+		return nil
+	}
+	mutatedPayload, mutated, err := rewriter.RewritePriority(fwkrh.PriorityRewriteContext{TargetEndpoint: reqCtx.TargetPod}, payload, reqCtx.Priority)
+	if err != nil {
+		return err
+	}
+	// Store the result back so repackage serializes the mutated payload.
+	inferenceRequestBody.Payload = mutatedPayload
```
ahg-g: why do we assign here in all cases? should we limit assignment to the case where mutated=true?

pkg/epp/framework/plugins/requesthandling/parsers/common.go
```diff
@@ -16,7 +16,49 @@ limitations under the License.
...
+func RewritePriority(ctx fwkrh.PriorityRewriteContext, payload fwkrh.MarshalablePayload, priority int) (fwkrh.MarshalablePayload, bool, error) {
+	m, ok := payload.(fwkrh.PayloadMap)
+	if !ok {
+		return payload, false, nil
+	}
+	delete(m, "priority")
+	if isVLLMTarget(ctx) {
+		priority = -priority
+	}
+	m["priority"] = priority
```
ahg-g, line 50: is this a field the same for both OAI and Message (Anthropic) protocols? if so, please document this assumption here.

sywangyi: update

pkg/epp/requestcontrol/director.go
sywangyi: updated

ahg-g (approved):

PR #2271, opened 2026-08-04
tracing: correlate logs with traces at request entry points

### LINKED ISSUE #1630: Logs are not correlated with traces: trace_id/span_id never injected into the logger

satyamg1620:
**What type of PR is this?**
/kind bug

The router emits OpenTelemetry spans and structured logr/zap logs, but nothing links them: no code path injects `trace_id`/`span_id` into the logger. Debugging a request means matching timestamps between the trace backend and the log backend.

This adds `tracing.LoggerWithSpanContext(ctx, span)`, which enriches the logger stored in the context with the active span's `trace_id`/`span_id`, and calls it at the two request entry points:

- **EPP** (`pkg/epp/handlers/server.go`): after `gateway.request` starts. Downstream code reads the logger via `log.FromContext`, so enriching here covers the whole request - the director, the scheduler, the plugins, and the kv-cache index.
- **Sidecar** (`pkg/sidecar/proxy/chat_completions.go`): after `forward_request` starts. The proxy logger is process-scoped, so it is seeded into the context first and the handler's own call sites use the request-scoped logger.

Correlation is injected once per context chain. logr appends values rather than replacing them, so injecting at nested span boundaries would repeat `trace_id` and `span_id` on every log line - the EPP nests up to eight spans - and log backends do not agree on which of a duplicated JSON key to keep. The logged `span_id` is therefore the outermost instrumented span, which names the entry span of the request; `trace_id` is the field to query by and selects the whole request across components.

The helper is a no-op when the span context is invalid, so nothing changes when tracing is disabled.

**Scope of the sidecar change**

The sidecar entry point correlates its own log lines and puts the correlated logger on the request context. The connectors reached from it (`connector_nixlv2.go`, `connector_shared_storage.go`, `connector_p2p.go`, `decode.go`, and the EC connectors) log through the process-scoped `s.logger` and are **not** correlated by this PR. Converting them is a mechanical but broad change across roughly 130 call sites and belongs in its own PR; the request context already carries what they need. Tracked as a follow-up.

**Why not a zap correlation hook**

Issue #1630 notes that the zap setup has no automatic correlation hook. A hook cannot work here: correlation lives on the context, and logr call sites do not hand the context to the sink. Injecting where the context is in scope is the only place the trace is knowable, and it keeps the correlation independent of which logr backend is configured.

Fixes #1630

Also covers llm-d/llm-d-kv-cache#665, which tracks the same gap for the kv-cache packages: `pkg/kvcache/indexer.go` keeps its own call so those packages self-correlate when driven directly. Reached through an EPP request the context is already correlated and the call is a no-op.

This PR supersedes #2248; @umakantv is credited as co-author.

**Release note**:
```release-note
Structured logs from the EPP and the disaggregation sidecar now carry the OpenTelemetry `trace_id` and `span_id` of the request, so logs and traces can be correlated.
```

## Test plan

Helper unit tests (`pkg/common/observability/tracing`):
- [x] Correlation fields are injected from a sampled span
- [x] A nested span boundary does not append a second pair of fields
- [x] Values already on the logger, and values added between boundaries, survive enrichment
- [x] No fields are added when the span context is invalid

Entry-point regression tests, which fail if the wiring is removed rather than only exercising the helper:
- [x] EPP: driving `StreamingServer.Process` with an upstream `traceparent` produces request logs carrying that trace ID exactly once, alongside the request ID
- [x] EPP: no correlation fields when tracing is disabled
- [x] Sidecar: `disaggregatedPrefillHandler` log lines carry the `forward_request` trace, and a connector adopting the request context gets the same trace
- [x] Sidecar: no correlation fields when tracing is disabled

- [x] `go build ./...`, `go vet`, `gofmt`, and `go test -race` on the touched packages
- [x] `make presubmit`

### DIFF
--- a/pkg/common/observability/tracing/logging.go
+++ b/pkg/common/observability/tracing/logging.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2025 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tracing
+
+import (
+	"context"
+
+	"go.opentelemetry.io/otel/trace"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	// LogKeyTraceID is the structured-log field carrying the active trace ID.
+	// It matches the W3C trace-context identifier used across the llm-d stack
+	// so logs from every component can be queried uniformly by trace.
+	LogKeyTraceID = "trace_id"
+
+	// LogKeySpanID is the structured-log field carrying the active span ID.
+	LogKeySpanID = "span_id"
+)
+
+// correlatedKey marks a context whose logger already carries the correlation
+// fields, so nested span boundaries do not append a second pair.
+type correlatedKey struct{}
+
+// LoggerWithSpanContext enriches the logger stored in ctx with the trace_id and
+// span_id of span. Call it immediately after starting a span at a request entry
+// point; downstream code reading the logger via log.FromContext then emits log
+// lines tagged with the active trace.
+//
+// Enrichment happens once per context chain. logr accumulates values rather than
+// replacing them, so calling this at a nested span boundary would repeat trace_id
+// and span_id on every log line, and which pair a backend keeps for duplicate keys
+// is undefined. The logged span_id is therefore the entry span of the request.
+//
+// The marker lives on the context while the fields live on the logger, so the two
+// desync if a caller installs a logger not derived from log.FromContext of an
+// already-correlated context. Derive from the context logger rather than building
+// one from scratch.
+func LoggerWithSpanContext(ctx context.Context, span trace.Span) context.Context {
+	if ctx.Value(correlatedKey{}) != nil {
+		return ctx
+	}
+
+	// An invalid span context means tracing is off or the span is a no-op.
+	sc := span.SpanContext()
+	if !sc.IsValid() {
+		return ctx
+	}
+
+	logger := log.FromContext(ctx).WithValues(
+		LogKeyTraceID, sc.TraceID().String(),
+		LogKeySpanID, sc.SpanID().String(),
+	)
+	return log.IntoContext(context.WithValue(ctx, correlatedKey{}, struct{}{}), logger)
+}
--- a/pkg/common/observability/tracing/logging_test.go
+++ b/pkg/common/observability/tracing/logging_test.go
@@ -1,0 +1,172 @@
+/*
+Copyright 2025 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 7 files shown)


satyamg1620:
@gyliu513 @liu-cong Can you please review this PR

gyliu513 (commented):
@satyamg1620 , I think this PR should fix https://github.com/llm-d/llm-d-router/issues/1630, can you make this PR as a general solution for router but not just kv-cache, comments?

satyamg1620:
> @satyamg1620 , I think this PR should fix #1630, can you make this PR as a general solution for router but not just kv-cache, comments?

Sure @gyliu513 , I will make the changes accordingly

ahg-g:
@gyliu513 is this ready?

satyamg1620:
> @gyliu513 is this ready?

@ahg-g I am pushing few changes in sometime what @gyliu513 asked for

satyamg1620:
@gyliu513 @ahg-g Can you please review this PR

gyliu513:
@satyamg1620 my bad, I found there is a PR created two weeks ago at https://github.com/llm-d/llm-d-router/pull/2248 by @umakantv, but I think this PR also covered kv cache as well.

@umakantv, considering this one has more coverage compared with your PR, do you mind close your PR and go with this one, and can you work with @satyamg1620 to consolidate this one (propose some comments?) and add you as co-author for this PR, make sense?

satyamg1620:
> @satyamg1620 my bad, I found there is a PR created two weeks ago at #2248 by @umakantv, but I think this PR also covered kv cache as well.
> 
> @umakantv, considering this one has more coverage compared with your PR, do you mind close your PR and go with this one, and can you work with @satyamg1620 to consolidate this one (propose some comments?) and add you as co-author for this PR, make sense?

@umakantv Let me know, I can add you as a co-author in this PR

umakantv:
> > @satyamg1620 my bad, I found there is a PR created two weeks ago at #2248 by @umakantv, but I think this PR also covered kv cache as well.
> > @umakantv, considering this one has more coverage compared with your PR, do you mind close your PR and go with this one, and can you work with @satyamg1620 to consolidate this one (propose some comments?) and add you as co-author for this PR, make sense?
> 
> @umakantv Let me know, I can add you as a co-author in this PR

@satyamg1620 sure, please add me, thanks

umakantv:
@satyamg1620 this PR looks complete to me related to the changes that were in my branch.
I will close my PR, we can continue with this branch as-is.

satyamg1620:
@gyliu513 Can you please review this PR

satyamg1620:
@gyliu513 @ahg-g Can you please review the changes

gyliu513:
Thanks @satyamg1620 , looks good to me.

@ahg-g can you help review to see if this can be approved? Thanks

ahg-g (approved):

PR #2164, opened 2026-07-24
feat(telemetry): align kvcache OTel scope and span names with router standard

Alex-ai-future:
## Summary

Align OpenTelemetry naming in the router-internalized kvcache packages with the conventions established in llm-d-kv-cache#654. This allows operators to filter traces across kv-cache and router in a unified dashboard.

## Changes

| Dimension | Before | After |
|-----------|--------|-------|
| **Instrumentation scope** | `"llm-d-kv-cache/pkg/kvcache"` | `"llm-d-router/pkg/kvcache"` |
| **Span names** | `llm_d.kv_cache.score_tokens`, `llm_d.kv_cache.scorer.compute`, `llm_d.kv_cache.index.add`, `llm_d.kv_cache.index.evict`, `llm_d.kv_cache.index` | `score_tokens`, `compute_scores`, `index_add`, `index_evict`, `index_lookup` |
| **Tracer helper** | `telemetry.Tracer()` from `pkg/telemetry` | `tracing.Tracer()` from `pkg/common/observability/tracing` |
| **Attribute keys** | `llm_d.kv_cache.*` | **Unchanged** |

## Impact

- **Breaking for trace consumers**: dashboards/queries filtering on old span names or scope names need updating.
- `service.name` remains `llm-d-router` (no change).
- `pkg/telemetry` is left untouched, pending the fast follow cleanup requested by ahg-g in PR #1886.

## Test plan

- [x] `go test ./pkg/kvcache/...`
- [x] `go test ./pkg/telemetry/...`

## Release note

```release-note
Standardized OpenTelemetry instrumentation scope names and span names for the internalized kvcache packages.
```

PR #2272, opened 2026-08-04
feat: add OpenTelemetry spans for tokenization and KV-cache event processing

### LINKED ISSUE #644: [Observability] Extend OTel Tracing Coverage (kvevents + tokenization)
Sub task of #617 

### Background
Index Add, Evict, and Lookup are already traced via the tracedIndex wrapper (pkg/kvcache/kvblock/traced_index.go, completed in #637), and score_tokens / scorer.compute spans exist. The remaining gaps are the ZMQ event-processing path and tokenization span propagation.

### Scope
Add spans for the ZMQ event-processing path in pkg/kvevents (message receive → decode → pool dispatch) and strengthen context propagation so tokenization spans link to their parent request span.

Instrument at the interface boundary following the existing tracedIndex wrapper pattern, do not inline spans into business logic. Reuse telemetry.InstrumentationName and the llm_d.kv_cache.* span-name prefix.

satyamg1620:
**What type of PR is this?**

/kind feature

Closes the two remaining OTel coverage gaps carried over from llm-d-kv-cache#644, whose subject code now lives in this repo:

1. **Tokenization**: a `tokenize` span in the token-producer plugin, attributed with backend (`uds`/`vllm`/`estimate`), token count, and the shared `mm.*` attributes. The span covers the backend call only, so the already-tokenized skip path stays untraced.

2. **KV-cache events**: `events_receive` -> `events_process` -> `events_decode` across the ZMQ pipeline. The receive span's identity is carried on `RawMessage` across the worker queue, so index writes nest under the event that caused them instead of appearing as detached root spans.

**Event tracing is opt-in** via `kvEventsConfig.tracing` (default `false`). This addresses the trace-volume concern raised in review: KV events arrive at many times the inference request rate, so with a shared head sampler, always-on event traces would crowd request traces out of the exported volume. When off, no event spans are created at all and request tracing is unaffected.

```yaml
- type: precise-prefix-cache-producer
  parameters:
    kvEventsConfig:
      zmqEndpoint: tcp://0.0.0.0:5557
      tracing: true
```

**Notes**

- *Context handoff.* llm-d-kv-cache#679 put a `context.Context` on the queued task. That is unsafe here: `runSubscriber` returns and retries on any recv error, so a queued message would carry a cancelled context into the worker and fail its index lookups. Only `trace.SpanContext` crosses the queue; the worker re-parents onto its own live context.

- *No interface change.* An earlier revision added `ctx` to `EngineAdapter.ParseMessage` so a decorator could open `events_decode`. Per review that breaking change is dropped: the span now wraps the adapter call in an unexported `Pool.decode` helper, and `EngineAdapter` is untouched, so out-of-tree adapters need no migration.

- *Local parent across the queue.* The receive span's identity crosses the worker queue in memory, so `processRawMessage` re-parents with `trace.ContextWithSpanContext` rather than `ContextWithRemoteSpanContext`. The remote form sets the span context's remote flag, which selects the `remoteParent*` branch of a `ParentBased` sampler and reports an in-process handoff as cross-process in exported spans.

- *Hot-path cost with tracing on.* The tracer is resolved once per `Pool` at construction rather than per message (`tracing.Tracer` rebuilds its instrumentation options on every call), span start options are package-level slices rather than variadic arguments rebuilt per call, and attribute construction sits behind `span.IsRecording()`. Measured on the ingest path: 953 ns/op and 14 allocs/op before, 750 ns/op and 8 allocs/op after.

- *No cost with tracing off.* `Pool.startSpan` returns the context untouched and a shared no-op span when `Config.Tracing` is unset, so `Start` is never reached on the default path. The pipeline's three spans cost 11.16 ns/op, 0 B/op and 0 allocs/op per message, against 100.6 ns/op, 144 B/op and 3 allocs/op when the same path ran through a no-op tracer. A no-op `Start` still builds its variadic option slice and a context per call, which at KV-event rates is not negligible.

**On #644's instrumentation constraints**

#644 asks to "instrument at the interface boundary following the existing tracedIndex wrapper pattern, do not inline spans into business logic. Reuse telemetry.InstrumentationName and the llm_d.kv_cache.* span-name prefix." None of the three is followed here; the reasoning for each:

1. **Interface boundary.** Superseded by review on this PR: "`processRawMessage` already holds a live ctx and a span at the call site, so `events_decode` could be opened inline there with no interface change and no migration cost for out-of-tree adapters." `events_receive` and `events_process` have no interface to wrap at all, and `tokenize` cannot be a decorator without losing `gen_ai.request.model` / `gen_ai.request.id`, which the backend interface never receives.

2. **`telemetry.InstrumentationName`.** No such identifier exists in llm-d-router; it belonged to llm-d-kv-cache's `pkg/telemetry`. The local equivalent is unexported (`tracing.instrumentationName`), reachable only as `tracing.Tracer()` with no scope. The pattern #644 points at does not use it either: `traced_index.go` passes a per-package scope. `main` carries those scopes as per-package `TracerScope` constants (`pkg/kvcache/constants.go`, `pkg/kvcache/kvblock/constants.go`, `pkg/epp/framework/plugins/requestcontrol/constants.go`) since #2164; the new `pkg/kvevents/constants.go` is the same pattern, and the `tokenize` span uses the existing requestcontrol constant.

3. **`llm_d.kv_cache.*` span-name prefix.** #2164 renamed exactly these five spans -- `llm_d.kv_cache.index.add` to `index_add`, `.index.evict` to `index_evict`, `.index` to `index_lookup`, `.score_tokens` to `score_tokens`, `.scorer.compute` to `compute_scores` -- and left the `llm_d.kv_cache.*` **attribute** keys in place. This PR matches `main` on both counts: bare `events_receive` / `events_process` / `events_decode` span names, `llm_d.kv_cache.events.*` attributes. No prefixed span names remain in the repo.

**Test plan**

Unit tests use an in-memory span recorder. Deployed behaviour was checked separately on a kind stack (vLLM simulator publishing KV events over ZMQ), with the exporter set to console at 100% sampling.

- [x] Tokenize span is emitted with backend, model, request id, token count and `mm.*`, and parents to the caller's span rather than rooting
- [x] Tokenize span records an error status when the backend fails, is absent entirely on the already-tokenized skip path, and reports `skipped_no_tokens` when the backend returns no tokens
- [x] `addTask` emits the receive span and stamps its identity onto the queued message
- [x] The receive span is the parent of the process span across the worker-queue boundary, and processing still traces as its own root when no identity was carried
- [x] Decode failure marks both the decode and process spans
- [x] A pool built the production way emits the decode span
- [x] Index writes nest under the process span instead of appearing as detached roots
- [x] No event spans at all when `kvEventsConfig.tracing` is unset, and the default is off
- [x] The process span records its parent as local, not remote, across the queue handoff
- [x] The disabled path allocates nothing, asserted with `testing.AllocsPerRun`
- [x] `make presubmit` and `go test ./pkg/kvevents/... ./pkg/epp/framework/plugins/requestcontrol/... ./pkg/kvcache/...`

On the kind stack with `tracing: true`, two traces, as intended, since KV events arrive asynchronously from vLLM and are not part of any router request:

```
request                        (root)    events_receive              (root)
  request_orchestration                    events_process
    tokenize                                 events_decode
    produce_precise_prefix_cache             index_add
    run_scheduler_profile
```

`events_process` records its parent as local (`Parent.Remote = false`) and `index_add` nests under it in the same trace.

With `tracing` unset (the default), no `events_*` spans are emitted, `index_add` returns to being a detached root exactly as before this PR, and the request trace is unchanged. KV events still flow through the pipeline in that configuration, so the disabled path is exercised rather than skipped.

Fixes https://github.com/llm-d/llm-d-kv-cache/issues/644

```release-note
EPP can now emit OpenTelemetry spans for prompt tokenization and for the KV-cache event pipeline (receive, decode, dispatch), linking KV-cache index operations to the event that triggered them. Tokenization spans are emitted whenever tracing is enabled; KV-cache event spans are opt-in via the new `tracing` field on `kvEventsConfig` and are off by default.
```

### DIFF
--- a/pkg/kvevents/events.go
+++ b/pkg/kvevents/events.go
@@ -76,7 +76,10 @@ type RawMessage struct {
 	// bridging the worker-queue boundary. Only the span identity crosses, never
 	// the subscriber's context: a subscriber reconnect cancels that context, and
 	// a queued task must not inherit the cancellation.
-	SpanContext trace.SpanContext
+	//
+	// A pointer keeps the 64-byte span context off messages the default
+	// configuration never traces. Nil when Config.Tracing is unset.
+	SpanContext *trace.SpanContext
 }
 
 // EngineAdapter defines the interface for engine-specific message parsers.
--- a/pkg/kvevents/events.go
+++ b/pkg/kvevents/events.go
@@ -14,6 +14,10 @@
 
 package kvevents
 
+import (
+	"go.opentelemetry.io/otel/trace"
+)
+
 // KVCacheSpecKind identifies vLLM KV cache group semantics.
 type KVCacheSpecKind string
 
@@ -68,6 +72,11 @@ type RawMessage struct {
 	SourceEndpoint string
 	// reset clears the message's pod before later messages on the same queue.
 	reset bool
+	// SpanContext links processing back to the span that received the message,
+	// bridging the worker-queue boundary. Only the span identity crosses, never
+	// the subscriber's context: a subscriber reconnect cancels that context, and
+	// a queued task must not inherit the cancellation.
+	SpanContext trace.SpanContext
 }
 
 // EngineAdapter defines the interface for engine-specific message parsers.
--- a/pkg/kvevents/pool.go
+++ b/pkg/kvevents/pool.go
@@ -242,9 +242,9 @@ func newEventTracer(enabled bool) trace.Tracer {
 }
 
 // startSpan opens a pipeline span, or leaves ctx untouched and returns a span
-// that records nothing when event tracing is disabled. A no-op span is cheap
-// but not free, and KV events arrive at many times the request rate, so the
-// default path skips Start rather than relying on a no-op tracer.
+// that records nothing when event tracing is disabled. The default path skips
+// Start rather than relying on a no-op tracer, which still builds an option
+// slice and a context per call. See Config.Tracing.
 func (p *Pool) startSpan(ctx context.Context, name string, opts []trace.SpanStartOption) (context.Context, trace.Span) {
 	if p.tracer == nil {
 		return ctx, disabledSpan
@@ -368,14 +368,15 @@ func (p *Pool) processRawMessage(ctx context.Context, msg *RawMessage) {
 	// cancellation, so index operations below land in the message's trace.
 	// The queue handoff stays inside one process, so the parent is local: a
 	// remote parent selects a different ParentBased sampler branch.
-	if msg.SpanContext.IsValid() {
-		ctx = trace.ContextWithSpanContext(ctx, msg.SpanContext)
+	if msg.SpanContext != nil {
+		ctx = trace.ContextWithSpanContext(ctx, *msg.SpanContext)
 	}
 
 	ctx, span := p.startSpan(ctx, "events_process", consumerSpanOptions)
 	defer span.End()
-	// A span that tracing sampled out still reports IsRecording false; skip
-	// attribute construction so those messages cost nothing either.
+	// Guards every attribute block in this package: a span the sampler dropped
+	// reports IsRecording false, so those messages skip attribute construction
+	// too, not just the ones with tracing off.
 	tracingActive := span.IsRecording()
 	if tracingActive {
 		span.SetAttributes(
--- a/pkg/kvevents/pool.go
+++ b/pkg/kvevents/pool.go
@@ -22,10 +22,15 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
+	tracenoop "go.opentelemetry.io/otel/trace/noop"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+	"github.com/llm-d/llm-d-router/pkg/common/observability/tracing"
 	"github.com/llm-d/llm-d-router/pkg/kvcache/kvblock"
 	"github.com/llm-d/llm-d-router/pkg/kvcache/metrics"
 )
@@ -92,6 +97,13 @@ type Config struct {
 	// EngineType selects the inference engine adapter ("vllm" or "sglang").
 	// Default: "vllm".
 	EngineType string `json:"engineType,omitempty"`
+	// Tracing enables the receive, process and decode spans this package emits.
+	// KV events arrive at many times the inference request rate, so those spans
+	// are opt-in: with a shared head sampler, always-on event spans would crowd
+	// request traces out of the exported volume. Index spans reached from the
+	// event path are emitted by the traced Index wrapper and are not gated
+	// here, so an exporter still receives those with this unset.
+	Tracing bool `json:"tracing,omitempty"`
 	// DiscoverPods enables the Kubernetes pod reconciler for automatic
 	// per-pod subscriber management. When enabled, the reconciler watches
 	// Kubernetes pods and creates/removes ZMQ subscribers dynamically.
@@ -161,7 +173,12 @@ type Pool struct {
 	// tier, KV-cache group, DP rank) and a store must be counted only after
 	// Index.Add succeeds — both of which only the Pool observes.
 	dedup *eventDedupFilter
-	wg    sync.WaitGroup
+	// tracer is resolved once: tracing.Tracer rebuilds its instrumentation
+	// options on every call, which is not free on the per-message event path.
+	// Nil when Config.Tracing is unset, which is what startSpan tests to skip
+	// span construction on the default path.
+	tracer trace.Tracer
+	wg     sync.WaitGroup
 	// queueDepth mirrors the number of tasks queued across all shards. It is
 	// tracked incrementally rather than by summing queue.Len() so that the
 	// depth gauge stays O(1) on the enqueue/dequeue hot path.
@@ -190,6 +207,7 @@ func NewPool(cfg *Config, index kvblock.Index, tokenProcessor kvblock.TokenProce
 		adapter:        adapter,
 		groupCatalog:   kvblock.NewGroupCatalog(),
 		dedup:          newEventDedupFilter(),
+		tracer:         newEventTracer(cfg.Tracing),
 	}
 
 	for i := 0; i < p.concurrency; i++ {
@@ -199,6 +217,39 @@ func NewPool(cfg *Config, index kvblock.Index, tokenProcessor kvblock.TokenProce
 	metrics.Register()
 
 	return p
+}
+
+// Span start options are built once. Passing them variadically at each call
+// site allocates a fresh slice per message.
+var (
+	consumerSpanOptions = []trace.SpanStartOption{trace.WithSpanKind(trace.SpanKindConsumer)}
+	internalSpanOptions = []trace.SpanStartOption{trace.WithSpanKind(trace.SpanKindInternal)}
+)
+
+// disabledSpan stands in for a real span when event tracing is off. A single
+// package-level value keeps the default path free of allocation while call
+// sites use the span unconditionally.
+var disabledSpan trace.Span = tracenoop.Span{}
+
+// newEventTracer returns the event-pipeline tracer, or nil when event tracing
+// is disabled. Resolving it once keeps the per-message path clear of
+// tracing.Tracer's per-call option building.
+func newEventTracer(enabled bool) trace.Tracer {
+	if !enabled {
+		return nil
+	}
+	return tracing.Tracer(TracerScope)
+}
+
+// startSpan opens a pipeline span, or leaves ctx untouched and returns a span
+// that records nothing when event tracing is disabled. A no-op span is cheap
+// but not free, and KV events arrive at many times the request rate, so the
+// default path skips Start rather than relying on a no-op tracer.
+func (p *Pool) startSpan(ctx context.Context, name string, opts []trace.SpanStartOption) (context.Context, trace.Span) {
+	if p.tracer == nil {
+		return ctx, disabledSpan
+	}
+	return p.tracer.Start(ctx, name, opts...)
 }
 
 // addQueueDepth adjusts the tracked queue depth by delta and publishes the new
@@ -313,16 +364,70 @@ func (p *Pool) processRawMessage(ctx context.Context, msg *RawMessage) {
 		return
 	}
 
-	podID, modelName, batch, err := p.adapter.ParseMessage(msg)
+	// Parent to the receive span while keeping the worker's context for
+	// cancellation, so index operations below land in the message's trace.
+	// The queue handoff stays inside one process, so the parent is local: a
+	// remote parent selects a different ParentBased sampler branch.
+	if msg.SpanContext.IsValid() {
+		ctx = trace.ContextWithSpanContext(ctx, msg.SpanContext)
+	}
+
+	ctx, span := p.startSpan(ctx, "events_process", consumerSpanOptions)
+	defer span.End()
+	// A span that tracing sampled out still reports IsRecording false; skip
+	// attribute construction so those messages cost nothing either.
+	tracingActive := span.IsRecording()
+	if tracingActive {
+		span.SetAttributes(
+			attribute.String("llm_d.kv_cache.events.topic", msg.Topic),
+			attribute.Int("llm_d.kv_cache.events.payload_size_bytes", len(msg.Payload)),
+		)
+	}
+
+	podID, modelName, batch, err := p.decode(ctx, msg)
 	if err != nil {
+		if tracingActive {
+			span.SetStatus(codes.Error, err.Error())
+		}
 		logger.Error(err, "Failed to parse message")
 		return
 	}
+
 	if msg.SourceEndpoint != "" {
 		podID = msg.SourceEndpoint
 	}
+	if tracingActive {
+		span.SetAttributes(
+			attribute.String("llm_d.kv_cache.events.pod_id", podID),
+			attribute.Int("llm_d.kv_cache.events.event_count", len(batch.Events)),
+		)
+	}
 
 	p.processEventBatch(ctx, &batch, podID, modelName)
+}
+
+// decode spans the adapter's payload decode. It wraps the call rather than the
+// EngineAdapter itself so out-of-tree adapters need no signature change.
+//
+//nolint:gocritic // unnamedResult: named returns conflict with nonamedreturns linter
+func (p *Pool) decode(ctx context.Context, msg *RawMessage) (string, string, EventBatch, error) {
+	_, span := p.startSpan(ctx, "events_decode", internalSpanOptions)
+	defer span.End()
+
+	podID, modelName, batch, err := p.adapter.ParseMessage(msg)
+	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
+		return podID, modelName, batch, err
+	}
+
+	// pod_id and event_count live on events_process, which holds the effective
+	// pod after the SourceEndpoint override. Repeating the pre-override pod
+	// under the same key would give one attribute two meanings in one trace.
+	if span.IsRecording() {
+		span.SetAttributes(attribute.String("gen_ai.request.model", modelName))
+	}
+
+	return podID, modelName, batch, nil
 }
 
 func (p *Pool) clearPod(ctx context.Context, podIdentifier string) {
--- a/pkg/kvevents/zmq_subscriber.go
+++ b/pkg/kvevents/zmq_subscriber.go
@@ -254,8 +254,6 @@ func (z *zmqSubscriber) addTask(ctx context.Context, topic string, seq uint64, p
 	// every stage of the pipeline.
 	_, span := z.pool.startSpan(ctx, "events_receive", consumerSpanOptions)
 	defer span.End()
-	// A span that tracing sampled out still reports IsRecording false; skip
-	// attribute construction so those messages cost nothing either.
 	if span.IsRecording() {
 		attrs := []attribute.KeyValue{
 			attribute.String("llm_d.kv_cache.events.topic", topic),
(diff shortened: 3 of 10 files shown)


satyamg1620:
@gyliu513 
Does "instrument at the interface boundary, do not inline" still hold now the code is outside llm-d-kv-cache? `events_receive` and `events_process` have no interface, and inventing ones for tracing runs against the repo's no-new-abstractions rule.

Also do you want me to handle this issue as 2 subtask https://github.com/llm-d/llm-d-kv-cache/issues/644? Let me know I can accordingly revert the changes to just tokenization part as done in https://github.com/llm-d/llm-d-kv-cache/issues/644

satyamg1620:
@vMaroon Can you please review this PR

pkg/kvevents/zmq_subscriber.go
```diff
@@ -231,18 +235,34 @@ func (z *zmqSubscriber) runSubscriber(ctx context.Context) {
...
 		z.hasLastSeq = true
 	}
 }
 
-func (z *zmqSubscriber) addTask(topic string, seq uint64, payload []byte) {
+// addTask hands a received message to the pool, carrying the receive span's
+// identity so processing rejoins this trace across the worker queue. The span
+// starts after Recv returns so it measures handoff work rather than the idle
+// wait for the next message.
+func (z *zmqSubscriber) addTask(ctx context.Context, topic string, seq uint64, payload []byte) {
```
gyliu513, line 252: This is the KV-event ingest hot path: one call per message from every vLLM pod and the instrumentation is not free even when tracing is disabled. `tracing.Tracer(TracerScope)` is resolved per call, and it builds a `TracerConfig` with `WithInstrumentationVersion` plus an `attribute.Set` for the commit SHA every time; the `SetAttributes` argument slice (including `strconv.FormatUint`) is built unconditionally even against a no-op span.

One caveat if you take this: cache the tracer on `zmqSubscriber`/`Pool` at construction, not in a package-level `var`. A tracer resolved before the first `otel.SetTracerProvider` is permanently pinned to the first provider ever installed, so a package-level var silently stops recording for every test after the first one that swaps in a `tracetest.SpanRecorder`. Resolving it in the constructor works with the ordering your new tests already use.

pkg/kvevents/events.go
```diff
@@ -76,7 +87,8 @@ type EngineAdapter interface {
 	// ParseMessage parses a raw transport message into domain data.
 	// It extracts pod identity and model name from the topic,
 	// and decodes the payload into an EventBatch.
-	ParseMessage(msg *RawMessage) (podID, modelName string, batch EventBatch, err error)
+	// ctx carries the caller's trace context so decorators can span the decode.
+	ParseMessage(ctx context.Context, msg *RawMessage) (podID, modelName string, batch EventBatch, err error)
```
gyliu513: Is the breaking `ParseMessage(ctx, ...)` change earning its keep? Both in-tree adapters discard the ctx (`_ context.Context`), and the only consumer is the decorator, which passes it to a callee that ignores it. `processRawMessage` already holds a live ctx and a span at the call site, so `events_decode` could be opened inline there with no interface change and no migration cost for out-of-tree adapters.  Comments?

gyliu513 (commented):
A general question: Every KV-event message now starts a new root trace. The default sampler is `parentbased_traceidratio` at 0.1, and KV events arrive at a far higher rate than inference requests, so after this change KV-event traces will dominate the exported trace volume by a wide margin and will crowd out request traces in most backends. Is there an intended way for an operator to keep request tracing on while turning this pipeline off, a config flag, or a separate sampler for this scope?

pkg/kvevents/events.go
satyamg1620: Agreed. I will drop the interface changes

pkg/kvevents/zmq_subscriber.go
satyamg1620: Taking both. Tracer cached on Pool and zmqSubscriber at construction, not package-level, for the provider-pinning reason you gave, and attribute construction moved behind `span.IsRecording()`, matching what scheduler_profile.go already does. Adding an addTask test since events_receive is currently uncovered.

satyamg1620:
> A general question: Every KV-event message now starts a new root trace. The default sampler is `parentbased_traceidratio` at 0.1, and KV events arrive at a far higher rate than inference requests, so after this change KV-event traces will dominate the exported trace volume by a wide margin and will crowd out request traces in most backends. Is there an intended way for an operator to keep request tracing on while turning this pipeline off, a config flag, or a separate sampler for this scope?

Fair point, and it's three spans per message, not one. A per-scope sampler isn't expressible, SamplingParameters carries span name, kind and attributes but not the instrumentation scope, so it would have to be name-prefix matching in InitTracing, which changes global policy for every scope; that feels like its own issue.

Proposal here: a tracing bool on kvevents.Config (already plugin-configurable via kvEventsConfig), read once at construction, default false. Request tracing stays unaffected unless an operator opts in, and when off no spans are created at all. Works for you, or keep this PR to instrumentation and create a follow-up PR?

pkg/kvevents/zmq_subscriber.go
gyliu513: Good, tracer is cached on Pool/zmqSubscriber at construction, and attribute building is gated behind IsRecording(). LGTM, thanks for the fix.

gyliu513:
Hi @satyamg1620 sorry for late.

> Proposal here: a tracing bool on kvevents.Config (already plugin-configurable via kvEventsConfig), read once at construction, default false. Request tracing stays unaffected unless an operator opts in, and when off no spans are created at all. Works for you, or keep this PR to instrumentation and create a follow-up PR?

Yes, that works for me, good to see it landed in this PR (kvEventsConfig.tracing, default false) rather than deferred. No need for a follow-up PR on this point.

lgtm. @ahg-g @vMaroon can you do another review to see if we are good to land this?

pkg/kvevents/pool.go
```diff
@@ -313,18 +337,75 @@ func (p *Pool) processRawMessage(ctx context.Context, msg *RawMessage) {
 		return
 	}
 
-	podID, modelName, batch, err := p.adapter.ParseMessage(msg)
+	// Parent to the receive span while keeping the worker's context for
+	// cancellation, so index operations below land in the message's trace.
+	if msg.SpanContext.IsValid() {
+		ctx = trace.ContextWithRemoteSpanContext(ctx, msg.SpanContext)
```
yankay: `events_receive` is created in this process, but `ContextWithRemoteSpanContext` marks it as a remote parent and may select the remote-parent sampler branch. Could we use `ContextWithSpanContext` and assert that `process.Parent().IsRemote()` is `false`?

yankay (commented):
Thanks for adding this tracing coverage. 
With event tracing disabled, three no-op `Start/End` pairs still run per message. A real project-path benchmark measured about 304 B and 9 extra allocations per message versus `main`. Could we bypass these calls before `Start` when tracing is disabled?

satyamg1620:
> Thanks for adding this tracing coverage. With event tracing disabled, three no-op `Start/End` pairs still run per message. A real project-path benchmark measured about 304 B and 9 extra allocations per message versus `main`. Could we bypass these calls before `Start` when tracing is disabled?

You're right, and thanks for measuring it rather than just flagging it. I'd been treating the no-op tracer as free and it isn't: Start still builds its variadic option slice and a fresh context every call, three times per message. I get 100.6 ns, 144 B and 3 allocs for the three pairs on their own. The other half is RawMessage, which went from 72 to 136 bytes once it started carrying a 64-byte trace.SpanContext, and that struct is heap allocated per message. Between the two I think we're looking at the same thing your 304 B picked up.

Spans now go through a small Pool.startSpan helper that returns the context untouched and a shared no-op span when Config.Tracing is off, so Start is never reached on the default path. I also moved the span options into package-level slices, which stops the enabled path rebuilding them per message.

`BenchmarkPipelineSpans_Disabled-12    11.16 ns/op    0 B/op    0 allocs/op`

TestStartSpan_DisabledPathAllocatesNothing uses AllocsPerRun to keep it there. I also ran the disabled path on a kind stack against live KV events, mostly to be sure the nil-tracer branch doesn't fall over in a real deployment: no event spans, no panic, and index_add back to being a detached root.

One thing I left alone. Those 64 bytes on RawMessage are unavoidable while the SpanContext is stored by value. A *trace.SpanContext would get the struct back to 80 bytes and push the cost onto the enabled path only, but it changes an exported field that out-of-tree adapters see, so I didn't want to make that call on my own. Happy to do it if you think it's worth the pointer.

pkg/kvevents/pool.go
satyamg1620: Taken. ContextWithRemoteSpanContext is just ContextWithSpanContext(..., rsc.WithRemote(true)), so an in-process queue handoff was being recorded as remote. No behaviour change under the default ParentBased config — remote and local branches are both AlwaysSample/NeverSample — but wrong in the exported span, and divergent under a custom sampler.

Switched to ContextWithSpanContext. Console exporter on a kind stack, same request shape:

`
before:  events_process   Parent.Remote = true`
`
after:   events_process   Parent.Remote = false
`

events_decode and index_add both still nest under events_process in the same trace. Added your assertion to TestProcessRawMessage_ParentsToReceiveSpan; it fails against the old call.

satyamg1620:
@gyliu513 @ahg-g @yankay Can you please review this PR

yankay:
HI @satyamg1620 

`events.go:79`
> **P1** Thanks for removing the no-op `Start/End` calls. Disabled tracing still costs 64 B per message and about 12-13% runtime because `trace.SpanContext` is stored by value. Could we avoid carrying this metadata when tracing is disabled?

`tokenizer_tracing_test.go:75`
> **P1** This introduces two `goconst` failures: extract `"user"` here and reuse `blockTypeImage` for `"image"` at line 125. Could we fix both so lint passes?

satyamg1620:
> HI @satyamg1620
> 
> `events.go:79`
> 
> > **P1** Thanks for removing the no-op `Start/End` calls. Disabled tracing still costs 64 B per message and about 12-13% runtime because `trace.SpanContext` is stored by value. Could we avoid carrying this metadata when tracing is disabled?
> 
> `tokenizer_tracing_test.go:75`
> 
> > **P1** This introduces two `goconst` failures: extract `"user"` here and reuse `blockTypeImage` for `"image"` at line 125. Could we fix both so lint passes?

Thanks @yankay.

**`events.go:79` — taken.** `unsafe.Sizeof(RawMessage{})` was 72 -> 136. `SpanContext` is now `*trace.SpanContext`, nil when tracing is off.

Enqueue path (`addTask` plus a drain), darwin/arm64, `-count=5`:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| main | 129.6 | 88 | 2 |
| before | 155.0 | 152 | 2 |
| after | 144.3 | 88 | 2 |

Full ingest path: 920 B/op on main and after, 984 before. Allocation is back to main.

One trap worth flagging. The obvious form measures 152 B / 3 allocs:

```go
if sc := span.SpanContext(); sc.IsValid() {
	msg.SpanContext = &sc
}
```

`&sc` makes `sc` escape, so it heap-allocates on every message including the ones that are never traced. Binding the copy inside the branch body avoids that. There is a comment on it so it does not get simplified back.

No behaviour change: a sampled-out span still returns a valid context, so the parent link still crosses the queue. Only the disabled path skips it. The field is added by this PR, so nothing out-of-tree depends on it yet.

**`tokenizer_tracing_test.go:75` — I don't think this one holds.**

`lint` is green on this SHA: https://github.com/llm-d/llm-d-router/actions/runs/33313814390/job/99584377549. CI runs plain `make lint`, not `--new`.

`"user"` already has 67 other occurrences in this package and `"image"` has 18, all pre-existing on `main`, so extracting the single occurrence this PR adds does not change the result. The only string this PR pushes over the threshold is `"render failed"` (`:142`, `:150`, plus the existing `tokenizer_test.go:491`). I can extract that one if you want it.

Which command are you seeing the failures from? If `make lint` fails on this branch for you I'd like to reproduce it.

ahg-g:
@gyliu513 should this PR apply the convention implemented in https://github.com/llm-d/llm-d-router/pull/2503 ?

gyliu513:
@ahg-g Yes, but I prefer we merge this first and do the semantic convention in another PR, @PlateauGao has another PR here https://github.com/llm-d/llm-d-router/pull/2516 addressing this issue.

yankay:
> > HI   嗨@satyamg1620
> > `events.go:79`
> > > **P1** Thanks for removing the no-op `Start/End` calls. Disabled tracing still costs 64 B per message and about 12-13% runtime because `trace.SpanContext` is stored by value. Could we avoid carrying this metadata when tracing is disabled?P1 感谢你们取消了那些无用的空调用。即使禁用跟踪功能，每次消息传输仍然需要消耗 64 字节的存储空间，而且由于 `trace.SpanContext` 是以值的形式存储的，因此运行时间也会增加大约 12-13%。我们能否在禁用跟踪功能时避免携带这些元数据呢？
> > 
> > 
> > `tokenizer_tracing_test.go:75`
> > > **P1** This introduces two `goconst` failures: extract `"user"` here and reuse `blockTypeImage` for `"image"` at line 125. Could we fix both so lint passes?P1 这里出现了两个错误：在行 125 中，变量 `"user"` 被提取了，而变量 `blockTypeImage` 又被重复使用了。我们能否修复这两个错误，让代码检查能够通过呢？
> 
> Thanks @yankay.  谢谢。
> 
> **`events.go:79` — taken.** `unsafe.Sizeof(RawMessage{})` was 72 -> 136. `SpanContext` is now `*trace.SpanContext`, nil when tracing is off. `events.go:79` ——已被获取。 `unsafe.Sizeof(RawMessage{})` 的值从 72 变为 136。 `SpanContext` 现在等于 `*trace.SpanContext` ，当禁用追踪功能时，该变量的值为空。
> 
> Enqueue path (`addTask` plus a drain), darwin/arm64, `-count=5`:入站路径（ `addTask` 加上一个消耗节点），达尔文/ARM64 架构， `-count=5` ：
> 
> ns/op	B/op	allocs/op  分配资源/操作
> main  主要的	129.6	88	2
> before  在……之前	155.0	152	2
> after  之后	144.3	88	2
> Full ingest path: 920 B/op on main and after, 984 before. Allocation is back to main.完整处理路径：在主要节点上，每次操作处理 920 字节的数据；而在其他节点上，每次操作处理 984 字节的数据。资源分配已恢复至主要节点。
> 
> One trap worth flagging. The obvious form measures 152 B / 3 allocs:有一个陷阱值得注意。显然，这个陷阱的规模为 152 B / 3 单位：
> 
> ```go
> if sc := span.SpanContext(); sc.IsValid() {
> 	msg.SpanContext = &sc
> }
> ```
> 
> `&sc` makes `sc` escape, so it heap-allocates on every message including the ones that are never traced. Binding the copy inside the branch body avoids that. There is a comment on it so it does not get simplified back. `&sc` 使得 `sc` 得以逃脱限制，因此每次消息发送时都会进行内存分配，包括那些根本不会被追踪的消息。将副本绑定到分支内部可以避免这种情况。关于这一点还有一段注释，所以相关内容不会被进一步简化处理。
> 
> No behaviour change: a sampled-out span still returns a valid context, so the parent link still crosses the queue. Only the disabled path skips it. The field is added by this PR, so nothing out-of-tree depends on it yet.没有行为变化：被排除在外的那段路径仍然能返回一个有效的上下文，因此父链接仍然会穿过队列。只有被禁用的路径会跳过它。这个字段是通过这个补丁添加的，所以目前还没有任何与树结构无关的功能依赖于它。
> 
> **`tokenizer_tracing_test.go:75` — I don't think this one holds. `tokenizer_tracing_test.go:75` ——我认为这个观点站不住脚。**
> 
> `lint` is green on this SHA: https://github.com/llm-d/llm-d-router/actions/runs/33313814390/job/99584377549. CI runs plain `make lint`, not `--new`. `lint` 在这个 SHA 代码中为绿色：https://github.com/llm-d/llm-d-router/actions/runs/33313814390/job/99584377549。CI 运行的是普通的 `make lint` ，而不是 `--new` 。
> 
> `"user"` already has 67 other occurrences in this package and `"image"` has 18, all pre-existing on `main`, so extracting the single occurrence this PR adds does not change the result. The only string this PR pushes over the threshold is `"render failed"` (`:142`, `:150`, plus the existing `tokenizer_test.go:491`). I can extract that one if you want it. `"user"` 在这个包中已经出现了 67 次了，而 `"image"` 则出现了 18 次，这些出现次数都存在于 `main` 中。因此，提取这个补丁新增的单个出现次数并不会改变最终结果。唯一超过阈值的字符串就是 `"render failed"` （还有 `:142` 、 `:150` ，以及已经存在的 `tokenizer_test.go:491` ）。如果你需要的话，我可以提取出那个字符串。
> 
> Which command are you seeing the failures from? If `make lint` fails on this branch for you I'd like to reproduce it.你们是在哪条命令上遇到故障的呢？如果 `make lint` 这条命令在这个分支上会失败，我希望你能帮我重现这种情况。

Thanks for checking. The pointer change resolves the allocation concern. You are also right about goconst; I cannot reproduce those findings with the project builder, so I’m retracting that comment.

4.

> > HI @satyamg1620
> > `events.go:79`
> > > **P1** Thanks for removing the no-op `Start/End` calls. Disabled tracing still costs 64 B per message and about 12-13% runtime because `trace.SpanContext` is stored by value. Could we avoid carrying this metadata when tracing is disabled?
> > 
> > 
> > `tokenizer_tracing_test.go:75`
> > > **P1** This introduces two `goconst` failures: extract `"user"` here and reuse `blockTypeImage` for `"image"` at line 125. Could we fix both so lint passes?
> 
> Thanks @yankay.
> 
> **`events.go:79` — taken.** `unsafe.Sizeof(RawMessage{})` was 72 -> 136. `SpanContext` is now `*trace.SpanContext`, nil when tracing is off.
> 
> Enqueue path (`addTask` plus a drain), darwin/arm64, `-count=5`:
> 
> ns/op	B/op	allocs/op
> main	129.6	88	2
> before	155.0	152	2
> after	144.3	88	2
> Full ingest path: 920 B/op on main and after, 984 before. Allocation is back to main.
> 
> One trap worth flagging. The obvious form measures 152 B / 3 allocs:
> 
> ```go
> if sc := span.SpanContext(); sc.IsValid() {
> 	msg.SpanContext = &sc
> }
> ```
> 
> `&sc` makes `sc` escape, so it heap-allocates on every message including the ones that are never traced. Binding the copy inside the branch body avoids that. There is a comment on it so it does not get simplified back.
> 
> No behaviour change: a sampled-out span still returns a valid context, so the parent link still crosses the queue. Only the disabled path skips it. The field is added by this PR, so nothing out-of-tree depends on it yet.
> 
> **`tokenizer_tracing_test.go:75` — I don't think this one holds.**
> 
> `lint` is green on this SHA: https://github.com/llm-d/llm-d-router/actions/runs/33313814390/job/99584377549. CI runs plain `make lint`, not `--new`.
> 
> `"user"` already has 67 other occurrences in this package and `"image"` has 18, all pre-existing on `main`, so extracting the single occurrence this PR adds does not change the result. The only string this PR pushes over the threshold is `"render failed"` (`:142`, `:150`, plus the existing `tokenizer_test.go:491`). I can extract that one if you want it.
> 
> Which command are you seeing the failures from? If `make lint` fails on this branch for you I'd like to reproduce it.

Thanks for checking. The pointer change resolves the allocation concern. You are also right about goconst; I cannot reproduce those findings with the project builder :-)

yankay:
HI @satyamg1620

Thanks for checking. The pointer change resolves the allocation concern. You are also right about goconst: the findings came from my local golangci-lint 2.13.1, while the project builder uses 2.10.0 and reports no issues. Sorry for the confusion :-)

ahg-g (approved):

Issue #2276, opened 2026-08-05
[Coordinator] Add metrics


revit13:
Add metrics to the coordinator to provide missing visibility into request volume, latency, pipeline steps, and decode-cache performance.

roytman:
Details are in [docs: add coordinator metrics specification ](https://github.com/llm-d/llm-d-router/pull/2277)

ahg-g:
What is the status of adding observability to the coordinator?

roytman:
PRs #2277 and #2435 resolved the issue. So I'm closing this issue. 

There is an additional issue, #2482, that extends this one and has its own PR #2616.

Issue #74, opened 2025-05-13
Q: Range of scores?


elevran:
Thinking about the valid output range for Scorers.
The final pod score is a linear combination of scores (i.e., Sum of `Wi`*`Si` for all scorers `i`). 
Currently scores are in range [0.0 .. 1.0] and weights are non-negative integers.

For the sake of discussion, suppose I would like to schedule to the Pod with the best _ratio_ of prefix and load (e.g., more prefix and less load. A Pod with 0.7 prefix and 0.1 load should be preferred to one with 0.75 prefix and 0.9 load).
Intuitively (or naively, YMMV), I would express the pod score as: `Prefix` - `Load`.

This runs into two issues:
- weights can't be negative
- scorers assign higher numbers to "better" pods. This means that it needs to invert the actual load (i.e., assign as `1.0 - load`). This means that using negative weights on "good" Pods does not do the right thing, so the scorer needs to assume something about the way the score might be weighted..

An alternative could be to define scores in the range [-1.0 .. +1.0] (with `0.0` being a neutral score) and use only positive weights.

Opening this issue for discussion.

mayabar:
Lets assume that the scorer result as a numerical value that ranks a pod’s ability to run a task based on a specific aspect. In load scorer example, it's result does not represent the pod’s current load, it indicates how available the pod is to handle an additional task. Does it make sense? Given this definition of the score, positive weights are enough.

elevran:
@mayabar it does make sense at a high level. I'm still trying to understand how (i.e., what weights) we approximate the original motivating example: pick the best pod based on prefix coverage and load.
Consider the following table (note: Load represents the actual load, not `1-load`, so 0.3 means 70% free as shown in the `1-L` column)

| Pod | Prefix | Load | P-L | 1-L |
|---|---|---|---|---|
| 1 | 0.8 | 0.3 | 0.5 | 0.7 |
| 2 | 0.7 | 0.1 | 0.6 | 0.4 |
| 3 | 0.4 | 1.0 | -0.6 | 0.0 |
| 4 | 0.5 | 0.0 | 0.5 | 1.0 |

Pod 2 should be selected as having the best `P`-`L` score.
What weights should I use for Prefix and (inverted) load?

nirrozenbaum:
your table has a typo - pod2 `1-L=0.9`
any equal weight to both scorers will give you what you're requesting.
example:
prefix_weight = 1, load_weight = 1 (total 2).
[...]

PR #2277, opened 2026-08-05
docs: add coordinator metrics specification

### LINKED ISSUE #123: add OWNERS file
typically in an open source project the repo contains OWNERS file, which specifies the reviewers/approvers of the repo (additional owners files can be put in sub directories). 

This is very useful for a newcomer and helps in understanding who they should reach out to when they open a new issue/PR. 

usually there’s also a protection rule that only the defined approvers can approve PRs for merge.

revit13:
**What type of PR is this?**
    /kind documentation
    /kind feature

### PR Description

    This PR adds `docs/metrics.coord.md` as the first step towards instrumenting the coordinator. It defines the metrics that the coordinator will expose, including request latency, pipeline step durations, and upstream sub-requests.

By establishing the metric naming conventions, label sets, and cardinality bounds up front, we can align the design with EPP's existing telemetry before opening implementation PRs.

    Part of #2276

### DIFF
--- a/docs/metrics.coord.md
+++ b/docs/metrics.coord.md
@@ -23,7 +23,7 @@ translate between the two components. Where a name matches an EPP metric, the
 Every metric on this page is exposed on a single `/metrics` endpoint served by the coordinator
 process, alongside the inference paths and `/healthz` and `/readyz` on the coordinator's chi router
 (`pkg/coordinator/server/server.go`). The registry it serves, the port or flag that selects it, and
-whether it is authenticated are settled by the phase 1 implementation. EPP's Prometheus wiring is
+whether it is authenticated are open. EPP's Prometheus wiring is
 separate: it lives in `cmd/epp/runner` and serves the controller-runtime registry rather than the
 default one.
 
@@ -65,7 +65,7 @@ A step may issue zero or many backend calls and includes local work; a phase is 
 `conditional-decode` issues a decode-phase call: `steps/decode_proxy.go` `newDecodeProxyRequest`
 stamps `EPPProfileHeader=PhaseDecode` unconditionally, and both the probe and the decode step go
 through it, so a cache-miss request sends two decode-phase calls. See the
-[open decision](#upstream-phase-family-phase-2) on the phase family.
+[open decision](#upstream-phase-family) on the phase family.
 
 ## Error classes
 
@@ -91,7 +91,7 @@ failure on the decode leg reaches the client but is absent from both error metri
 
 ## Metrics catalog
 
-### Request family (phase 1)
+### Request family
 
 Recorded in `server/handlers.go` `handleInference`. Requests that exit early because they are
 malformed (body-read error, 413, invalid JSON) count as `bad_request` with `model_name=unknown`.
@@ -110,7 +110,7 @@ Unless otherwise noted, metrics in this family share the label `{model_name}`.
 dashboard panel that groups the family by model has no series for it. EPP's
 `llm_d_epp_request_running` is labelled by model, fairness ID, and priority.
 
-### Pipeline step family (phase 1)
+### Pipeline step family
 
 Recorded in `pipeline.go` `Execute`, one observation per step per request. This family measures each
 stage's whole wall time, local work plus orchestration plus any backend calls the stage makes, not
@@ -151,7 +151,7 @@ histogram_quantile(0.95, sum by (le) (rate(
   llm_d_coordinator_pipeline_step_duration_seconds_bucket{step="decode"}[5m])))
 ```
 
-### Upstream phase family (phase 2)
+### Upstream phase family
 
 Recorded in `steps/encode.go`, `prefill.go` and `decode_proxy.go`. This family counts and times the
 outbound sub-requests the coordinator sends to the gateway carrying the epp-profile header. Only
@@ -201,7 +201,7 @@ upstream_request_total{phase="prefill"} += 1
 upstream_request_total{phase="decode"}  += 1
 ```
 
-### Disaggregation decision and decode cache (phase 3)
+### Disaggregation decision and decode cache
 
 #### `disagg_decision_total`
 
--- a/docs/metrics.coord.md
+++ b/docs/metrics.coord.md
@@ -173,12 +173,13 @@ This is reachable by any client, and it is a mitigation the implementation has t
 
 Capping the distinct values is the approach that fits: `model_name` is capped at 1000 over the
 process lifetime and further values are reported as `other`. The bounded-label helper lives in
-`pkg/common/observability/metrics/cardinality.go` (`BoundedLabel`, `OverflowValue = "other"`) and is
-instantiated with the same cap by both `pkg/coordinator/metrics/cardinality.go` and
-`pkg/epp/metrics/cardinality.go`, so the two components share one guard. An allowlist has no source
-of truth here, since the coordinator's config carries no model list and the coordinator is otherwise
-model-agnostic. The overflow value must not be `unknown`, which already means the request carried no
-model at all.
+[`pkg/common/observability/metrics/cardinality.go`](../pkg/common/observability/metrics/cardinality.go)
+(`BoundedLabel`, `OverflowValue = "other"`) and is instantiated with the same cap by both
+[`pkg/coordinator/metrics/cardinality.go`](../pkg/coordinator/metrics/cardinality.go) and
+[`pkg/epp/metrics/cardinality.go`](../pkg/epp/metrics/cardinality.go), so the two components share
+one guard. An allowlist has no source of truth here, since the coordinator's config carries no model
+list and the coordinator is otherwise model-agnostic. The overflow value must not be `unknown`,
+which already means the request carried no model at all.
 
 ## Related documentation
 
--- a/docs/metrics.coord.md
+++ b/docs/metrics.coord.md
@@ -90,9 +90,13 @@ probe has its own metric,
 the EPP side, as `request_error_total{error_code="PreconditionFailed"}`.
 
 `upstream_4xx` and `upstream_5xx` therefore describe the render, prefill, and encode steps, the only
-ones that turn a non-2xx response into an error. The conditional-decode and decode steps
-proxy the upstream response straight to the client without constructing an error, so an upstream
-failure on the decode leg reaches the client but is absent from both error metrics.
+ones that turn a non-2xx response into an error. The conditional-decode and decode steps proxy the
+upstream response straight to the client without constructing an error, so an upstream failure on the
+decode leg reaches the client but is absent from both error metrics. This covers the worst case: a
+worker that dies mid-generation truncates the client's response while the request still counts as a
+success. Both failure modes are already detected by the decode proxy, which writes 502 when the call
+fails before the response starts and logs a truncation when the copy fails after it, so closing the
+gap is a matter of recording those two points rather than adding detection.
 
 ## Metrics catalog
 
@@ -214,10 +218,13 @@ decode one step is one gateway call:
   one concurrent call per image, so `pipeline_step_duration_seconds{step="encode"}` observes the
   whole stage once per request (the fan-out envelope: slowest call plus the EC merge) and hides what
   a single image cost. A step p95 well above the call p95 means the width of the fan-out, not any one
-  image, is the expense. For prefill and decode the call latency already equals the step duration. Two loose ends remain: the `phase` label then carries exactly one value, which
-  invites empty `phase="prefill"` queries, and `generalLatencyBuckets` runs to 1h, far past any
-  single encode call. Either drop the label and name the metric for what it measures, or keep it
-  explicitly for forward compatibility, and consider `stepLatencyBuckets`.
+  image, is the expense. For prefill and decode the call latency already equals the step duration. Two
+  loose ends remain. The `phase` label carries exactly one value, which invites empty
+  `phase="prefill"` queries: either drop it and name the metric for what it measures, or keep it
+  explicitly for forward compatibility. And `generalLatencyBuckets` runs to 1h, far past any single
+  encode call, so `stepLatencyBuckets` fits the range better; note that `generalLatencyBuckets` is
+  EPP's own set, so switching gives up bucket-for-bucket comparison with
+  `llm_d_epp_request_duration_seconds`.
 - There is no `upstream_request_error_total`. A phase-call failure aborts its step and is already
   counted by `pipeline_step_errors_total` under the same `error_code`. Per-call encode fan-out
   failure granularity is the one gap, but the errgroup context cancels sibling calls on the first
@@ -320,11 +327,16 @@ transport-independent measure of image size, and are already computed in render 
 ## Cardinality
 
 `model_name` labels every request-family metric, three of them histograms, and the handler takes it
-straight from the request body without validation. An arbitrary set of client-supplied strings
-multiplied by the latency buckets is
-a cardinality-explosion vector against the coordinator's own memory, reachable by any client. The
-mitigation has to be chosen before implementing: allowlist against configured models, cap the
-distinct values as EPP does, or accept the risk explicitly.
+straight from the request body without validation. Each distinct value creates its own set of series,
+and a histogram multiplies that by its bucket count, so a client looping over invented model names
+grows the coordinator's memory without bound. This is reachable by any client, and it is a mitigation
+the implementation has to carry.
+
+Capping the distinct values is the approach that fits: EPP caps `model_name` at 1000 over the process
+lifetime and reports further values as `other`, and matching that keeps the two components
+consistent. An allowlist has no source of truth here, since the coordinator's config carries no model
+list and the coordinator is otherwise model-agnostic. The overflow value must not be `unknown`, which
+already means the request carried no model at all.
 
 ## Related documentation
 
--- a/docs/metrics.coord.md
+++ b/docs/metrics.coord.md
@@ -46,7 +46,9 @@ three to follow a request end to end.
 
 ## Labels
 
-Two labels use the words `encode`, `prefill`, and `decode`. They are not interchangeable.
+The `step` and `phase` labels share the values `encode`, `prefill`, `decode`, and
+`conditional-decode`, but are not interchangeable: a step is a pipeline stage, a phase is a single
+backend call.
 
 The **`step` label** names a stage of the coordinator's internal pipeline, observed once per request
 per stage. A step covers local work, orchestration, and any backend calls that stage makes. The
@@ -71,6 +73,10 @@ A step may issue zero or many backend calls and includes local work; a phase is 
 The first three values match the epp-profile header the coordinator sends. `conditional-decode` does
 not: the probe carries the decode profile header, like the decode step, and is distinguished by the
 `Prefer: if-available` header it adds. See the [Upstream phase family](#upstream-phase-family).
+
+The **`decision_type` label** names which phases a request ended up running, one value per request
+rather than per stage or per call: `decode-only`, `prefill-decode`, or `encode-prefill-decode`. See
+[`disagg_decision_total`](#disagg_decision_total).
 
 ## Error classes
 
--- a/docs/metrics.coord.md
+++ b/docs/metrics.coord.md
@@ -5,12 +5,15 @@ runs to serve them (see [Coordinator Architecture](coordinator_architecture.md))
 from the Endpoint Picker (EPP) metrics documented in [Metrics](metrics.md): the two components
 measure different points in the same request path.
 
-## Subsystems and naming
+## Subsystem and naming
 
-Every series carries the subsystem prefix `llm_d_coordinator_`, so a metric's full Prometheus name is
-`llm_d_coordinator_<name>`. Names below omit the prefix. All metrics are ALPHA stage.
+A metric's full Prometheus name is `<subsystem>_<name>`. The coordinator uses a single subsystem:
 
-Naming mirrors the EPP request family (`pkg/epp/metrics`) so PromQL expressions and dashboard panels
+| Prefix | Scope |
+|---|---|
+| `llm_d_coordinator_` | Every coordinator metric: request, pipeline step, upstream phase, disaggregation decision, and decode cache. |
+
+Naming mirrors the EPP request family so PromQL expressions and dashboard panels
 translate between the two components. Where a name matches an EPP metric, the
 [Relationship to EPP metrics](#relationship-to-epp-metrics) section states what differs.
 
@@ -21,13 +24,14 @@ translate between the two components. Where a name matches an EPP metric, the
 ### Coordinator metrics endpoint
 
 Every metric on this page is exposed on a single `/metrics` endpoint served by the coordinator
-process, alongside the inference paths and `/healthz` and `/readyz` on the coordinator's chi router
-(`pkg/coordinator/server/server.go`). The registry it serves, the port or flag that selects it, and
-whether it is authenticated are open. EPP's Prometheus wiring is
-separate: it lives in `cmd/epp/runner` and serves the controller-runtime registry rather than the
-default one.
+process, alongside the inference paths and `/healthz` and `/readyz`. EPP's metrics endpoint is
+separate, served from its own process on the controller-runtime registry.
 
-### Metrics served elsewhere
+TBD: the registry the endpoint serves, its address, and whether it is authenticated. EPP authenticates
+its endpoint by default (`--metrics-endpoint-auth`), which requires RBAC for TokenReview and
+SubjectAccessReview; matching that would extend the coordinator's ServiceAccount permissions.
+
+### Other scrape targets
 
 All coordinator metrics are self-instrumented: the coordinator counts and times its own work
 in-process, and scrapes nothing. Two series record a signal that originates upstream but is still
@@ -40,14 +44,16 @@ the other two endpoints: per-request token counts and pool aggregates such as KV
 and queue depth on EPP's (see [Metrics](metrics.md)), multimodal cache counters on vLLM's. Scrape all
 three to follow a request end to end.
 
-## Label vocabularies
+## Labels
 
-Two vocabularies use the words `encode`, `prefill`, and `decode`. They are not interchangeable.
+Two labels use the words `encode`, `prefill`, and `decode`. They are not interchangeable.
 
-**`step`** is a stage of the coordinator's internal pipeline, observed once per request per stage. A
-step covers local work, orchestration, and any backend calls that stage makes.
+The **`step` label** names a stage of the coordinator's internal pipeline, observed once per request
+per stage. A step covers local work, orchestration, and any backend calls that stage makes. The
+values are the built-in step names; [Coordinator Architecture](coordinator_architecture.md) describes
+what each one does and how a pipeline is assembled from them.
 
-| Value | Stage |
+| `step` value | Stage |
 |---|---|
 | `render` | Call the render service to tokenize and apply the chat template. |
 | `replace-media-urls` | Rewrite media URLs in the request body (local work only). |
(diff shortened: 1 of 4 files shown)


roytman:
Hi Guangya Liu @guygir, can you take a look from an observability point of view?

docs/metrics.coord.md
```diff
@@ -0,0 +1,345 @@
...
+
+One client request passes through the coordinator, the EPP behind the gateway, and the vLLM workers,
+and each of the three serves its own `/metrics`. Measurements this page does not list are on one of
+the other two endpoints: per-request token counts and pool aggregates such as KV-cache utilization
+and queue depth on EPP's (see [Metrics](metrics.md)), multimodal cache counters on vLLM's. Scrape all
+three to follow a request end to end.
+
+## Labels
+
+Two labels use the words `encode`, `prefill`, and `decode`. They are not interchangeable.
```
capri-xiyue: nits: `Three labels`

revit13: Reworded to name the pair instead of counting: "The `step` and `phase` labels share
the values `encode`, `prefill`, `decode`, and `conditional-decode`, but are not
interchangeable: a step is a pipeline stage, a phase is a single backend call." Thanks

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+from the Endpoint Picker (EPP) metrics documented in [Metrics](metrics.md): the two components
+measure different points in the same request path.
+
+## Subsystem and naming
+
+A metric's full Prometheus name is `<subsystem>_<name>`. The coordinator uses a single subsystem:
+
+| Prefix | Scope |
+|---|---|
+| `llm_d_coordinator_` | Every coordinator metric: request, pipeline step, upstream phase, disaggregation decision, and decode cache. |
```
mayabar: Maybe use `llm-d-coordinator_` to keep a single `_`  as the separator between the subsystem and the name?

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+Recorded by the request handler. Requests that exit early because they are malformed (body-read
+error, 413, invalid JSON) count as `bad_request` with `model_name=unknown`.
+
+| Name | Type | Notes |
+|---|---|---|
+| `request_total` | Counter | Every inbound client request, including malformed ones. |
+| `request_error_total` | Counter | Failed requests; adds label `error_code`. |
+| `request_duration_seconds` | Histogram | End-to-end request latency; `generalLatencyBuckets` (5ms to 1h). |
+| `request_size_bytes` | Histogram | Request body length; powers-of-2 buckets. |
+| `response_size_bytes` | Histogram | Bytes streamed to the client, measured by a counting `ResponseWriter` wrapper in the handler rather than by parsing the body. |
```
mayabar, line 85: nit: maybe define buckets here?

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+### Upstream phase family
+
+Label set `{phase}` (one outbound backend call, not a pipeline stage, see [Labels](#labels)).
+
+Recorded by the conditional-decode, encode, prefill, and decode steps. This family counts and times the outbound sub-requests to the gateway. (Render and media fetches are not included, their latency is tracked in `pipeline_step_duration_seconds`).
+
+| Name | Type | Notes |
+|---|---|---|
+| `upstream_request_total` | Counter | Gateway sub-requests, one per call. (e.g. encode contributes one per image). |
+| `upstream_request_duration_seconds` | Histogram | Latency of one encode call (encode only). |
```
mayabar: nit: maybe define buckets here?

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+    *   `miss`: The worker returned HTTP 412. The pipeline continues to encode/prefill/decode.
+*   **Note:** Not redundant with `disagg_decision_total` since a cache miss does not indicate whether the subsequent path was `prefill-decode` or `encode-prefill-decode`.
+
+## Relationship to EPP metrics
+
+| Coordinator metric | EPP counterpart | Difference |
+|---|---|---|
+| Request family (`request_total`, etc.) | `llm_d_epp_*` (same names) | Coordinator counts single client requests at entry; EPP counts every sub-request reaching the gateway. EPP adds flow-control labels (`fairness_id`, `priority`). |
+| `disagg_decision_total` | `llm_d_epp_disagg_decision_total` | Coordinator counts phases *executed*; EPP counts routing decisions *made* and adds plugin labels. |
+| `pipeline_step_*`, `upstream_request_*`, `decode_cache_lookups_total` | None | Unique to coordinator. |
```
mayabar: nit: maybe add a prefix to the coordinator’s metrics in the first column to match the second column?

docs/metrics.coord.md
roytman: we should be in line with `llm_d_epp_` and `llm_d_router_epp_` metrics:

roytman: maybe instead of 
"Every coordinator metric: request, pipeline step, upstream phase, disaggregation decision, and decode cache. "
"Canonical, coordinator-wide: " request, response, pipeline step/latency, and other metrics"

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+translate between the two components. Where a name matches an EPP metric, the
+[Relationship to EPP metrics](#relationship-to-epp-metrics) section states what differs.
+
+`model_name` is taken from the request body; an empty or absent model is recorded as `unknown`.
+
+## Scrape topology
+
+### Coordinator metrics endpoint
+
+Every metric on this page is exposed on a single `/metrics` endpoint served by the coordinator
```
roytman, line 26: on the metrics port  (default 9090, configurable with --metrics-port)

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+process, alongside the inference paths and `/healthz` and `/readyz`. EPP's metrics endpoint is
+separate, served from its own process on the controller-runtime registry.
+
+TBD: the registry the endpoint serves, its address, and whether it is authenticated. EPP authenticates
```
roytman: EPP ??

docs/metrics.coord.md
revit13: Hyphens aren't valid in classic Prometheus metric names. Even with Subsystem: "llm-d-coordinator", the text exposition escapes them back, so the scraped name is llm_d_coordinator_request_total either way. Keeping it also matches the EPP subsystem llm_d_epp in this repo, which the coordinator naming mirrors. Thanks

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+its endpoint by default (`--metrics-endpoint-auth`), which requires RBAC for TokenReview and
+SubjectAccessReview; matching that would extend the coordinator's ServiceAccount permissions.
+
+### Other scrape targets
+
+All coordinator metrics are self-instrumented: the coordinator counts and times its own work
+in-process, and scrapes nothing. Two series record a signal that originates upstream but is still
```
roytman: "It is not clear about two series record" 
I don't think that `decode_cache_lookups_total` is a good name.

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+measured by the coordinator: `upstream_request_duration_seconds` times each encode sub-request, and
+`decode_cache_lookups_total` records the decode server's response to the conditional-decode probe.
+
+One client request passes through the coordinator, the EPP behind the gateway, and the vLLM workers,
```
roytman, line 41: Why do we need this paragraph?

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+and each of the three serves its own `/metrics`. Measurements this page does not list are on one of
+the other two endpoints: per-request token counts and pool aggregates such as KV-cache utilization
+and queue depth on EPP's (see [Metrics](metrics.md)), multimodal cache counters on vLLM's. Scrape all
+three to follow a request end to end.
+
+## Labels
+
+The `step` and `phase` labels share most of their values, but measure different boundaries: a step is a pipeline stage (which may make multiple calls), while a phase is a single backend call.
```
roytman: Do we need this separation?

docs/metrics.coord.md
revit13: docs/metrics.md doesn't document buckets for EPP histograms, so I dropped the bucket details here to keep the two docs aligned. Can add exact boundaries to both if you prefer. Thanks

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+The `error_code` label on `request_error_total` and `pipeline_step_errors_total` uses four values:
+
+- **`bad_request`**: Client-side errors (e.g., malformed body).
+- **`upstream_4xx`**: 4xx errors from upstream (e.g., render, prefill, encode).
+- **`upstream_5xx`**: 5xx errors from upstream.
+- **`internal`**: All other coordinator-internal faults.
+
+**Key behaviors:**
+- The conditional-decode probe's HTTP 412 (cache miss) is handled internally and is **not** an error. It is tracked separately by [`decode_cache_lookups_total`](#decode_cache_lookups_total).
+- `upstream_4xx` and `upstream_5xx` only apply to the `render`, `prefill`, and `encode` steps. The `decode` and `conditional-decode` steps proxy responses directly to the client, so upstream failures during decode currently reach the client without incrementing these error metrics.
```
roytman: I think we have the response code from `decodea and `conditional-decode`

docs/metrics.coord.md
revit13: Done, the comparison table now uses full names in both columns. The catalog above still omits the prefix, since it states it once at the top. Thanks

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+
+Recorded by the request handler. Requests that exit early because they are malformed (body-read
+error, 413, invalid JSON) count as `bad_request` with `model_name=unknown`.
+
+| Name | Type | Notes |
+|---|---|---|
+| `request_total` | Counter | Every inbound client request, including malformed ones. |
+| `request_error_total` | Counter | Failed requests; adds label `error_code`. |
+| `request_duration_seconds` | Histogram | End-to-end request latency; `generalLatencyBuckets` (5ms to 1h). |
+| `request_size_bytes` | Histogram | Request body length; powers-of-2 buckets. |
```
roytman: why it is "powers-of-2 buckets." ? The EPP metrics don't have it

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+| `response_size_bytes` | Histogram | Bytes streamed to the client, measured by a counting `ResponseWriter` wrapper in the handler rather than by parsing the body. |
+| `request_running` | Gauge | Requests in flight. |
```
roytman, line 86: Should we add 
request_input_tokens
request_ttft_seconds 	Histogram 	Time to first token; adds label streaming.
request_streaming_tpot_seconds 	Histogram 	Time per output token (streaming).
request_streaming_itl_seconds 	Histogram 	Inter-token latency (streaming).

(I'm not sure that we can measure the last two)

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+
+EPP's `llm_d_epp_request_running` adds `fairness_id` and `priority`. Both come from flow control,
```
roytman: we can remove it

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+### Pipeline step family
+
+Label set `{step}` (a stage of the coordinator's internal pipeline, see [Labels](#labels)).
+
+Recorded by the pipeline executor, one observation per step per request. This family measures each stage's total wall time (local work, orchestration, and all backend calls made by the stage). It answers where time goes inside the coordinator and which internal stage failed.
+
+| Name | Type | Notes |
+|---|---|---|
+| `pipeline_step_duration_seconds` | Histogram | Per-step latency. Uses fine-grained `stepLatencyBuckets` (100us to 60s) to capture both microsecond steps (render) and seconds-long steps (decode). |
+| `pipeline_step_errors_total` | Counter | Step failures; adds label `error_code`. |
```
roytman: `step_request_running` | Gauge | Requests in flight. |

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Pipeline step family
+
+Label set `{step}` (a stage of the coordinator's internal pipeline, see [Labels](#labels)).
+
+Recorded by the pipeline executor, one observation per step per request. This family measures each stage's total wall time (local work, orchestration, and all backend calls made by the stage). It answers where time goes inside the coordinator and which internal stage failed.
+
+| Name | Type | Notes |
+|---|---|---|
+| `pipeline_step_duration_seconds` | Histogram | Per-step latency. Uses fine-grained `stepLatencyBuckets` (100us to 60s) to capture both microsecond steps (render) and seconds-long steps (decode). |
```
roytman: maybe we can remove the `pipeline_` prefix

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+| `pipeline_step_errors_total` | Counter | Step failures; adds label `error_code`. |
+
+**Key details:**
+*   Only steps that actually executed are recorded. Trailing steps from early exits (e.g., conditional-decode hit) or failures are not emitted.
+
+### Upstream phase family
+
```
roytman, line 104: Again, I don't think (maybe I'm wrong) we don't need this label vs phase separation. It can mislead, plus we have other steps that call external services - `render`, `url-downloader` and probably others in the future. 
For all these steps the metrics can be relevant.

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+*   **Fan-out:** A single client request can result in multiple backend calls (e.g., one conditional-decode probe, two encode calls for two images, one prefill, one decode).
+*   **Narrower scope than steps:** 
+    *   `upstream_request_duration_seconds` is encode-only because it tracks per-image latency. Prefill and decode already have a 1:1 mapping between call latency and step duration, so their timings are tracked solely in `pipeline_step_duration_seconds`.
+    *   There is no `upstream_request_error_total`. A phase-call failure aborts its step and is already counted by `pipeline_step_errors_total`.
+*   **Conditional-decode:** The probe gets its own phase rather than counting as `decode`, keeping fan-out ratios accurate.
+
+### Disaggregation decision and decode cache
+
+#### `disagg_decision_total` (Counter)
+*   **Labels:** `model_name`, `decision_type` (`decode-only`, `prefill-decode`, `encode-prefill-decode`)
```
roytman, line 122: Coordinator's decision is only one type - disaggregate or not; it is the response from conditional-decode`
Maybe we should redefine this counter

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+*   **Description:** Records which phases actually ran for a request. Unlike EPP, it lacks `encode-decode` because encode always implies prefill in the coordinator.
+
+#### `decode_cache_lookups_total` (Counter)
+*   **Labels:** `result` (`hit` or `miss`)
+*   **Description:** Records the outcome of the conditional-decode probe.
```
roytman: Conditional decode can forward the request to a Decode node (doesn't return 412) in two cases: cache hit, or the input prompt is too short. Coordinator cannot know why the request was sent to a Decode node.

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Relationship to EPP metrics
+
+| Coordinator metric | EPP counterpart | Difference |
+|---|---|---|
+| Request family (`request_total`, etc.) | `llm_d_epp_*` (same names) | Coordinator counts single client requests at entry; EPP counts every sub-request reaching the gateway. EPP adds flow-control labels (`fairness_id`, `priority`). |
+| `disagg_decision_total` | `llm_d_epp_disagg_decision_total` | Coordinator counts phases *executed*; EPP counts routing decisions *made* and adds plugin labels. |
+| `pipeline_step_*`, `upstream_request_*`, `decode_cache_lookups_total` | None | Unique to coordinator. |
+
+**EPP-only metrics:** EPP exposes token counts, latencies (TTFT/TPOT/ITL), and metrics for scheduling, flow control, and pool aggregates. The coordinator does not measure these as they are outside its scope.
```
roytman: I think Coordinator can measure it, furthermore, in a deployment with Coordinator, these metrics from EPP are not correct.

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Deliberate omissions
+
+### Token counts
+
+The coordinator emits no token-count metrics. EPP already parses the vLLM `usage` block to emit `request_input_tokens`, `request_output_tokens`, and `request_cached_tokens`, and the coordinator avoids duplicating this effort. Furthermore, extracting output tokens would require the coordinator to parse the streamed SSE response, defeating the purpose of a pure byte proxy (which is why we use `response_size_bytes`).
```
roytman: request_input_tokens are part of the Renderer response

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+| `upstream_request_total` | Counter | Gateway sub-requests, one per call. (e.g. encode contributes one per image). |
+| `upstream_request_duration_seconds` | Histogram | Latency of one encode call (encode only). |
+
+**Key details:**
+*   **Fan-out:** A single client request can result in multiple backend calls (e.g., one conditional-decode probe, two encode calls for two images, one prefill, one decode).
+*   **Narrower scope than steps:** 
+    *   `upstream_request_duration_seconds` is encode-only because it tracks per-image latency. Prefill and decode already have a 1:1 mapping between call latency and step duration, so their timings are tracked solely in `pipeline_step_duration_seconds`.
+    *   There is no `upstream_request_error_total`. A phase-call failure aborts its step and is already counted by `pipeline_step_errors_total`.
+*   **Conditional-decode:** The probe gets its own phase rather than counting as `decode`, keeping fan-out ratios accurate.
+
```
roytman, line 118: if we work with the "inference/v1/generate" interface, we can add more metrics, especially about multimedia tensors.

docs/metrics.coord.md
```diff
@@ -0,0 +1,182 @@
...
+### Token counts
+
+The coordinator emits no token-count metrics. EPP already parses the vLLM `usage` block to emit `request_input_tokens`, `request_output_tokens`, and `request_cached_tokens`, and the coordinator avoids duplicating this effort. Furthermore, extracting output tokens would require the coordinator to parse the streamed SSE response, defeating the purpose of a pure byte proxy (which is why we use `response_size_bytes`).
+
+*Future candidate:* `request_prompt_tokens` (Histogram, after render) could be added to correlate prompt size with encode fan-out without needing a cross-component join.
+
+### Per-request image visibility
+
+Neither the coordinator nor EPP tracks per-request image count or size. Currently, image count is only visible in aggregate via `upstream_request_total{phase="encode"}`. Image byte size is unreliable because URL images are fetched by `replace-media-urls` and do not traverse the coordinator.
```
roytman: yes, but we have base64 size

roytman:
Thank you @revit13 for organising everything together; eventually, this document can be merged with `docs/metrics.md`

docs/metrics.coord.md
revit13: docs/metrics.md doesn't document buckets for EPP histograms, so I dropped the bucket details here to keep the two docs aligned. Can add exact boundaries to both if you prefer. Thanks

docs/metrics.coord.md
revit13: Reworded, "two series record a signal that originates upstream" was doing too much work. It now names them directly.
Agreed on the name too. The coordinator doesn't look anything up in a cache; it sends a conditional-decode probe and records the answer. Renamed decode_cache_lookups_total to conditional_decode_probes_total, labels unchanged (result=hit|miss).

docs/metrics.coord.md
revit13: It answers "I have this page's metrics, why can't I see my whole request?" The coordinator only sees its own hop, so the section says where the rest lives and that you need all three endpoints to follow a request end to end.

The enumeration of what's on EPP was duplicating the "Relationship to EPP metrics" section, so I cut that and kept only the part that isn't stated elsewhere:


One client request passes through the coordinator, the EPP behind the gateway, and the vLLM workers,
and each of the three serves its own `/metrics`. Following a request end to end means scraping all
three: what this page does not list is on EPP's endpoint (see [Metrics](metrics.md)) or vLLM's.
If you think even that is redundant with the intro, I'm fine dropping the paragraph entirely.

docs/metrics.coord.md
revit13: Yes, the two aren't interchangeable. The encode step fans out one concurrent sub-request per multimodal entry, so a six-image request is one step="encode" observation and six phase="encode" observations. With only one label you either lose the per-sub-request latency distribution (is one image slow, or are they all slow?) or lose the wall-clock cost of the stage, which is what shows up in client latency since the fan-out is parallel. render and replace-media-urls also make no backend call, so they exist as steps but have no phase.

The old wording buried that in a parenthetical. Rewrote it to state the divergence up front with the fan-out example.

docs/metrics.coord.md
revit13: Split verdict on these four.

request_input_tokens: yes, feasible. The coordinator already holds reqCtx.TokenIDs after render, so it's a len() on a field it computed itself, no response parsing needed. The page already lists it as a future candidate under the name request_prompt_tokens; I can rename it to request_input_tokens to match EPP, though the sources differ (coordinator counts the rendered prompt, EPP reads the vLLM usage block).

request_ttft_seconds: feasible as first-byte, and my stated reason for omitting it is too strong. The distinction is what the first bytes on the wire actually are. A streaming response arrives as SSE frames, data: {...}\n\n, and the first frame carries no generated text: it's the role-delta chunk with an empty content. To know a frame carries a real token you have to split the stream on frame boundaries, JSON-decode the chunk, and check choices[0].delta.content is non-empty. That is parsing the response body, which is exactly what the decode step avoids: it hands upstream bytes to io.Copy and never looks inside them. So a true first-token timestamp does need parsing.

Time to first byte needs none of that. The response-writer wrapper we already use for response_size_bytes just records when the first Write lands, no knowledge of the payload. It's off by one frame from true TTFT, which is negligible next to the render/encode/prefill wait it captures, and it's the one measurement nothing else can produce: EPP's decode-leg TTFT starts after those phases have finished, so only the coordinator sees the client's full wait.

TPOT and ITL: your instinct is right, not from here. Both need per-token boundaries for every token, not just the first, so the parsing problem above applies to the whole stream. Worse, the decode leg is a reverse proxy with FlushInterval: -1 and writes come out of io.Copy's buffer, so a write isn't reliably one SSE frame; counting writes would mean inferring token boundaries from buffer boundaries. Same reason we skip output-token counts, and EPP already emits both per leg.

How do you want to handle it? I'd rather not grow this PR, so my preference is to correct the TTFT rationale on this page and open a separate issue for adding input tokens plus first-byte TTFT. Say the word if you'd rather they land here.

docs/metrics.coord.md
revit13: You're right that the current split is arbitrary. phase really means "gateway call": render calls the renderer service and replace-media-urls downloads one URL per image, and neither is counted.

I'd widen it rather than drop it. replace-media-urls fetching eight images is one step but eight calls, so a step-only metric can't tell one slow download from eight. Same for encode's per-image fan-out.

Proposal: upstream_request_total{upstream} and upstream_request_duration_seconds{upstream} covering every outbound call (render, media-fetch, encode, prefill, conditional-decode, decode). Steps answer "where did time go", upstreams answer "how many calls and how slow was each", and future steps that call out just get a value.

Does that address it, or do you still see the two as redundant?

docs/metrics.coord.md
revit13: Right, the labels asserted a reason the coordinator can't see. Changed hit/miss to served/deferred, with served noted as "cached or too short to disaggregate, we can't tell which". Also fixed the error-class bullet that called the 412 a cache miss.

The code has the same assumption (errCacheMiss in decode_proxy.go); renaming it is a code change, so I left it out unless you want it here.

docs/metrics.coord.md
revit13: Please see my response above on the same topic. Thanks

docs/metrics.coord.md
revit13: Please see my response above on the same topic. Thanks

docs/metrics.coord.md
mayabar: I added this comment because in other place the buckets boundaries were explained, not critical(`request_duration_seconds` | Histogram | End-to-end request latency; `generalLatencyBuckets` (5ms to 1h).)

docs/metrics.coord.md
```diff
@@ -0,0 +1,187 @@
...
+[Relationship to EPP metrics](#relationship-to-epp-metrics) section states what differs.
+
+`model_name` is taken from the request body; an empty or absent model is recorded as `unknown`.
+
+## Scrape topology
+
+### Coordinator metrics endpoint
+
+Every metric on this page is exposed on a single `/metrics` endpoint served by the coordinator
+process on the metrics port (default 9090, configurable with `--metrics-port`), separate from the listener
```
gyliu513, line 27: Does the coordinator has `--metrics-port` flag? Seems it registers only `--config` plus the logging flags, and everything else (including `server.listen_addr`) comes from the YAML in `config/coordinator/coordinator.yaml`.

docs/metrics.coord.md
```diff
@@ -0,0 +1,187 @@
...
+
+- **`bad_request`**: Client-side errors (e.g., malformed body).
+- **`upstream_4xx`**: 4xx errors from upstream (e.g., render, prefill, encode).
+- **`upstream_5xx`**: 5xx errors from upstream.
+- **`internal`**: All other coordinator-internal faults.
+
+**Key behaviors:**
+- The conditional-decode probe's HTTP 412 (the worker declined to serve it) is handled internally and is **not** an error. It is tracked separately by [`conditional_decode_probes_total`](#conditional_decode_probes_total-counter).
+- `upstream_4xx` and `upstream_5xx` are recorded for the `render`, `prefill`, and `encode` steps, which read the upstream status before continuing. The `decode` and `conditional-decode` steps stream the upstream response straight to the client, and today only the conditional-decode probe inspects its status (to detect the 412 miss), so a decode-phase status error reaches the client uncounted. The status is available to both: it can be read in the reverse proxy's `ModifyResponse` hook before any bytes are forwarded. What stays uncountable is a failure that occurs after streaming has begun, since the 200 and a partial body are already on the wire.
```
gyliu513: This bullet says only `decode` streams without inspecting status, but per the point above `conditional-decode` inspects the status *only* for 412, so its non-412 failures are equally uncounted today. Worth folding into the same sentence so the gap is not understated.

docs/metrics.coord.md
```diff
@@ -0,0 +1,187 @@
...
+straight from the request body without validation. Each distinct value creates its own set of series,
+and a histogram multiplies that by its bucket count, so a client looping over invented model names
+grows the coordinator's memory without bound. This is reachable by any client, and it is a mitigation
+the implementation has to carry.
+
+Capping the distinct values is the approach that fits: EPP caps `model_name` at 1000 over the process
+lifetime and reports further values as `other`, and matching that keeps the two components
+consistent. An allowlist has no source of truth here, since the coordinator's config carries no model
+list and the coordinator is otherwise model-agnostic. The overflow value must not be `unknown`, which
+already means the request carried no model at all.
```
gyliu513: The cardinality analysis is right, but its scope is too narrow: `disagg_decision_total` also carries `model_name`and is not part of the request family, so the bound has to cover it too. Suggest phrasing the rule as "every metric carrying `model_name`".

Also worth naming the existing implementation so the coordinator reuses rather than reinvents it: `pkg/epp/metrics/cardinality.go` already has the limiter (`maxModelLabelValues = 1000`, `overflowValue = "other"`). Pointing at it makes "matching EPP" a checkable instruction.

docs/metrics.coord.md
```diff
@@ -0,0 +1,187 @@
...
+
+One client request passes through the coordinator, the EPP behind the gateway, and the vLLM workers,
+and each of the three serves its own `/metrics`. Following a request end to end means scraping all
+three: what this page does not list is on EPP's endpoint (see [Metrics](metrics.md)) or vLLM's.
+
+## Labels
+
+The `step` and `upstream` labels share most of their values, but measure different boundaries: a step is a pipeline stage, while an upstream is a single outbound call. They diverge where a stage is not one call: the `encode` step fans out one concurrent sub-request per multimodal entry, so a request with six images records one `step="encode"` observation and six `upstream="encode"` observations.
+
+- **`step`**: A stage of the internal pipeline, observed once per request per stage. Covers local work and all outbound calls made by that stage. Values: `render`, `replace-media-urls`, `encode`, `prefill`, `conditional-decode`, `decode`. See [Coordinator Architecture](coordinator_architecture.md).
```
gyliu513: The `step` values are presented as a closed set, but steps are registry plugins (`pipeline.Register`), so an out-of-tree step contributes its own label value. Please state that the label value is the step's `Name()` and that cardinality is bounded by the registered step types, otherwise the Cardinality section reads as if `model_name` is the only unbounded-looking label.

docs/metrics.coord.md
roytman: yes, it is added in #2435

docs/metrics.coord.md
roytman: in #2435 both steps now capture `decodeOutcome{Status, TransportErr}` from the reverse proxy and return `UpstreamStreamedError` on any 4xx/5xx or transport failure, verified by `TestConditionalDecodeStep_ServerError` and `TestDecodeStep_TransportError`.

docs/metrics.coord.md
dmitripikus: Done in PR [#74](https://github.com/revit13/llm-d-inference-scheduler/pull/74). Each histogram row now names its bucket variable inline: request_duration_seconds, step_duration_seconds, and upstream_request_duration_seconds reference GeneralLatencyBuckets (5 ms to 1 h); request_size_bytes references RequestSizeBuckets (64 B to 1 GiB, powers of two); request_input_tokens references TokenCountBuckets (1 to ~1 M).

docs/metrics.coord.md
dmitripikus: Doc bullet rewritten in PR [#74](https://github.com/revit13/llm-d-inference-scheduler/pull/74). upstream_4xx and upstream_5xx are now recorded for every step that calls out, including decode and conditional-decode

docs/metrics.coord.md
dmitripikus: Added in PR [#74](https://github.com/revit13/llm-d-inference-scheduler/pull/74). request_input_tokens is now listed as a real metric in the Request table; before it was only mentioned as a possible future addition. The row also shows the range of token counts we cover. The three streaming timings — time to first token, per-output-token time, and inter-token latency — stay out for now.

docs/metrics.coord.md
dmitripikus: Removed in PR[ #74](https://github.com/revit13/llm-d-inference-scheduler/pull/74).

docs/metrics.coord.md
dmitripikus: Renamed in PR [#74](https://github.com/revit13/llm-d-inference-scheduler/pull/74). The metric is now execution_path_total with the path label (same three values), the catalog subsection is retitled "Execution path and conditional decode", and the Relationship-to-EPP row reads: "coordinator observes the phases that actually ran on the client request; EPP records the routing decision that was made and adds plugin labels." Matches the code in #2435

docs/metrics.coord.md
dmitripikus: Follow-up in PR [#74](https://github.com/revit13/llm-d-inference-scheduler/pull/74). The result label on conditional_decode_probes_total now has three values instead of two: served, deferred, and a new error. The error value covers any other failure on the probe — the worker returned an error, or the call couldn't reach it at all. The earlier rename from hit/miss to served/deferred — the change that stopped the label from claiming a reason the coordinator can't see — is already in. The new error value comes with #2435

docs/metrics.coord.md
dmitripikus: Fixed in PR [#74](https://github.com/revit13/llm-d-inference-scheduler/pull/74). Both decode and conditional-decode now catch every failure from the worker (any 4xx/5xx response, plus network failures) before the response streams back. A non-412 failure on conditional-decode counts in step_errors_total and shows up as conditional_decode_probes_total.result="error".

docs/metrics.coord.md
dmitripikus: Rewritten in PR [#74](https://github.com/revit13/llm-d-inference-scheduler/pull/74). The Cardinality section now covers every metric with model_name (so execution_path_total is included), and it names the shared code that enforces the cap: pkg/common/observability/metrics/cardinality.go (BoundedLabel, cap 1000, overflow "other"), used by both the coordinator and the EPP. #2435 moved it out of pkg/epp/metrics so both sides share one copy.

docs/metrics.coord.md
dmitripikus: Added in PR [#74](https://github.com/revit13/llm-d-inference-scheduler/pull/74). The step bullet now says the label value is each step's registered name (from pipeline.Register), and the list of names is labeled "Values in the built-in registry" — so a step added later just adds its own name and the doc stays correct.

revit13:
Hi @gyliu513 , thank you for your comments on this PR! Apologies for my delayed response. @Dmitri Pikus has addressed your feedback, and the PR is now ready for re-review. Thanks again!

gyliu513:
@revit13 thanks for the update, lgtm!

roytman (approved):

elevran (approved):
approving on LGTM from @gyliu513 and @roytman

PR #2284, opened 2026-08-05
feat(flowcontrol): split queue-wait budget by unavailability regime

### LINKED ISSUE #2183: [Flow Control] Queue-wait TTL should split by unavailability regime, like the status code in #2046

Split the flow control queue-wait budget by unavailability regime, the same way #2046 / #2047 split the TTL-eviction status code.

That issue established the discriminator and the table:

|                     | Capacity rejection | TTL eviction |
|---------------------|--------------------|--------------|
| Empty pool          | 503                | 503          |
| Non-empty saturated | 429                | 429          |

The budget should follow the same seam:

|                     | Queue-wait budget          | Expiry means        |
|---------------------|----------------------------|---------------------|
| Empty pool          | long, sized to cold start  | 503, genuine outage |
| Non-empty saturated | short, sized to a TTFB SLO | 429, backpressure   |

The signal already exists: `pkg/epp/flowcontrol/controller/internal/processor.go` carries `poolEmpty` and uses it to pick `QueueOutcomeRejectedNoEndpoints` over `QueueOutcomeRejectedCapacity` at the capacity boundary (#1751).

This was option [3] in the design discussion on #2123:

> 3. Pause or extend the TTL while the pool is empty, so it measures "the pool is up and you still
>    couldn't dispatch."
>
> [3] is probably the right end state, but it's new mechanism.

#2123 shipped option [1] (60s plus documentation) to unblock the default-on flip (#2104 / #2180). This issue is the follow-through.

Open questions, roughly in order of how much they shape the design:

1. **What defines the empty-pool regime.** `poolEmpty` is the obvious predicate, but the utilization detector scores stale-metric endpoints as 1.0 saturated (`pkg/epp/saturationdetector/utilization/detector.go:124-125`). A total scrape failure therefore presents as the saturated regime (short budget, 429) when the honest reading is unavailability. Decide whether the predicate is "no endpoints in the pool" or "no endpoint with fresh metrics".

2. **Regime is not fixed at enqueue.** The pool going empty to non-empty is the entire point of scale-from-zero, so the effective deadline has to be re-evaluated while queued, not frozen at admission. Likely shape: the item carries two budgets and the expiry scan charges elapsed time to whichever regime is current.

3. **Transition accounting.** A pod appears at t=90s, past the short saturation budget but inside the long cold-start budget. Charging against an already-exhausted saturation budget would shed the request the instant it becomes servable, which is pathological. Probably grant a fresh saturation budget on the transition, since the request only just became dispatchable.

4. **Config surface.** Keep `defaultRequestTTL` as the saturation budget and add a second field for the no-endpoint budget, or generalize per band. Explicit `0s` must keep meaning "disabled". Related: the per-priority and per-request TTL scopes discussed on #2123 are still unbuilt.

5. **Pairs with a gateway deadline.** A long cold-start budget only helps if the caller waits. The well-lit guides configure no gateway request timeout, which is why the TTL currently acts as the primary deadline rather than a backstop. Guides shipping a timeout is arguably the more important fix; the two should be documented together.

6. **Metrics.** A distinct outcome for no-endpoint expiry keeps the regimes separable on dashboards. The queue-duration metric already labels by outcome.

**Why is this needed**:

`defaultRequestTTL` is a single budget (60s since #2123). One number is serving two regimes that want opposite values.

*No viable endpoint* (scale-from-zero, total outage). Waiting is the only path to success, so the budget wants to be long: a cold start is image pull plus weight load, minutes not seconds. 60s is too short, which is why #2123's own release note already tells scale-to-zero users to raise it.

*Endpoints exist but are saturated.* Waiting past the time-to-first-token budget yields a degraded response anyway; the useful answer is to shed so the caller can retry or load-shed elsewhere. Here 60s is far too long. Nothing interactive has a 60s TTFB budget, and the legacy admission path rejected sheddable traffic immediately.

A single default cannot serve both, so it protects neither. It also costs observability: any client or gateway deadline under 60s wins the race, and `inferOutcome` (`pkg/epp/flowcontrol/controller/internal/item.go:178-184`) then records `EvictedContextCancelled` rather than `EvictedTTL`. The shed-under-saturation signal disappears and the symptom reads as "clients went away".

The default already bites in practice. Flipping the gate on by default (#2180) broke two e2e disruption specs: with the gate on, an empty pool is a scale-from-zero waiting room, and `QueueOutcomeRejectedNoEndpoints` only fires at queue capacity (`processor.go:294-306`), which the 5000/band default never reaches for a probe request. So the request holds for the full 60s TTL while the test's 10s client timeout and 30s probe window both expire first. #2180 works around it with a per-test `defaultRequestTTL: 5s`. Under the split that override becomes an honest statement ("this deployment does not do scale-from-zero") rather than a fudged timeout.

One sizing note worth keeping regardless of the design chosen: under a dispatch halt, queue occupancy is arrival rate times TTL. #2123's own math puts the 5000/band cap binding at roughly 83 req/s with a 60s TTL, versus roughly 330-500 req/s at 10-15s. A short saturation budget makes the capacity limit bind before the time limit, which is the ordering you want, because a capacity rejection is immediate and attributable while a TTL eviction is neither.

Related: #2046, #2047, #2097, #2123, #2104, #1751, #1187.

satyamg1620:
**What type of PR is this?**
/kind feature

`defaultRequestTTL` is one queue-wait budget serving two regimes that want opposite values: an empty pool is a cold start where waiting is the only path to success, a saturated pool should shed early to protect TTFB. One number protects neither.

- Adds `noEndpointRequestTTL`, in force while the candidate pool has no endpoints, following the seam #2046 used for the TTL-eviction status code.
- Moves expiry from the request context to the processor sweep, where pool emptiness is observed, so the budget is re-evaluated while queued rather than fixed at admission.
- Charges elapsed time from the later of enqueue and the last regime change, so a pool scaling up grants queued requests a fresh budget instead of shedding them the moment they become dispatchable.
- Relaxes the request context to the longer of the two budgets; it is now a backstop, not the budget.
- Adds an `EvictedNoEndpoints` outcome that maps to 503 without a pool probe and keeps the regimes separable on dashboards.
- Leaves `noEndpointRequestTTL` following `defaultRequestTTL` when unset, so splitting the regimes is opt-in: a deployment that stated one bound on queue wait keeps it in both regimes, and an explicit `0s` still disables eviction outright.

Behavior at the defaults is unchanged in kind. Expiry now lands at the first sweep after the budget elapses rather than on an exact timer (under a second later at the default sweep interval), and while the pool is empty it carries the `EvictedNoEndpoints` outcome label instead of `EvictedTTL`. The HTTP status and the `x-llm-d-request-dropped-reason` header are unchanged in both regimes; the empty-pool error text gains a `: no endpoints available` suffix.

Reviewer notes:

- The inheritance in `NewConfigFromAPI` is load-bearing, not cosmetic. Defaulting the two budgets independently narrows the documented meaning of `defaultRequestTTL: "0s"`, which #2123 shipped as "disables the TTL", to the regime where the pool has endpoints. A deployment that had disabled queue-wait eviction would then start shedding 60s into a scale-from-zero, and one that set a short bound would pick up a cold-start budget it never asked for. Both are the scale-to-zero deployments this split exists to serve. Inheriting keeps `0s` whole and keeps the split opt-in.
- `EffectiveTTL` on `QueueItemAccessor` is one regime's budget, not the point at which the item is evicted. Its doc now says so, since ordering policies read it and EDF derives a deadline from it.
- Two test fixtures were corrected, not just adapted: `TestTTLExpiryEvictsQueuedRequest` claimed to queue "under saturation" but ran against an empty pool, and `TestZombieCapacityStarvation` has to reach its zombie state through a caller deadline now that the sweep finalizes and removes an expired item in the same pass.
- Out of scope, tracked on the issue: gateway-timeout guidance, and whether the predicate should be "no endpoints" or "no endpoint with fresh metrics".
- Deferred to the sweep-only enforcement follow-up: finalization runs inside the `ManagedQueue` write lock, so a mass expiry does its metric work under the lock that `Add` and `Remove` contend on; `isExpired` reads the processor-global budget in the empty-pool regime, so a per-request TTL from #1090 would be ignored there; and EDF's deadline no longer matches the real expiry deadline for the same reason.

Test plan:

- [x] Budget selection per regime, including that an explicit zero disables eviction there
- [x] A queued request holds past the saturation budget on an empty pool and sheds as unavailability when the cold-start budget runs out
- [x] A pool scaling from zero dispatches a request whose saturation budget had nominally elapsed
- [x] Regime changes are timestamped once per transition; an unchanged regime does not restart the budget
- [x] Empty-pool expiry surfaces as 503 with the no-endpoints drop reason
- [x] An unset `noEndpointRequestTTL` follows `defaultRequestTTL`, an explicit one overrides it, and an explicit `0s` on either budget does not inherit
- [x] `noEndpointRequestTTL` round-trips from YAML to the controller config, and a config naming only `defaultRequestTTL: 0s` disables eviction in both regimes
- [x] `make presubmit` and `make test-unit-epp`

Manually verified on kind (ad hoc, not added to the e2e suite, which does not enable the `flowControl` gate). With `defaultRequestTTL: 2s`: shed as 503 at 8.011s and 45.005s against 8s and 45s cold-start budgets, and a request queued against an empty pool was served (200) after 15.3s across a real scale-from-zero. A kind scale-up takes ~11s to yield a discoverable endpoint even with vllm-sim, so this budget must be sized well above pod startup.

Fixes #2183

**Release note**:
```release-note
Flow control queue wait can now be budgeted separately for an empty pool. `noEndpointRequestTTL` bounds how long a request waits while the candidate pool has no endpoints, leaving `defaultRequestTTL` to bound waiting against a saturated pool. Size it above expected pod startup: a request shed while the pool is still coming up would have been servable moments later. An explicit `0s` waits indefinitely for an endpoint. The budget in force is re-evaluated while a request is queued, so a pool that scales up moves its queued requests onto `defaultRequestTTL` with a fresh budget, though total queue wait stays bounded by the longer of the two budgets plus a short expiry-sweep margin, and a regime change does not extend that bound. Empty-pool expiry is reported as `EvictedNoEndpoints` and returns 503. `noEndpointRequestTTL` follows `defaultRequestTTL` when unset, and both default to 60s when neither is set, so an existing `defaultRequestTTL` continues to govern both regimes until you split them, including an explicit `0s`, which disables eviction in both. At the defaults an empty-pool expiry lands at the first expiry sweep after the budget elapses (under a second later at the default sweep interval) and carries the `EvictedNoEndpoints` outcome label instead of `EvictedTTL`, so dashboards or alerts keyed on `EvictedTTL` for scale-from-zero need to account for the new label. The HTTP status and the `x-llm-d-request-dropped-reason` header are unchanged.
```

### DIFF
--- a/apix/config/v1alpha1/endpointpickerconfig_types.go
+++ b/apix/config/v1alpha1/endpointpickerconfig_types.go
@@ -403,8 +403,9 @@ type FlowControlConfig struct {
 	// within it is shed with a retryable backpressure error rather than served late, and with no
 	// client or gateway deadline it is the only bound on waiting against a pool that has endpoints.
 	// Where such deadlines exist and fire sooner, they evict the request first (client disconnect).
-	// An explicit "0s" disables eviction while the pool has endpoints: such requests then wait until
-	// client disconnect or controller shutdown.
+	// An explicit "0s" disables eviction while the pool has endpoints, and, unless NoEndpointRequestTTL
+	// overrides it, while the pool is empty as well: such requests then wait until client disconnect or
+	// controller shutdown.
 	DefaultRequestTTL *metav1.Duration `json:"defaultRequestTTL,omitempty"`
 
 	// +optional
@@ -419,9 +420,10 @@ type FlowControlConfig struct {
 	// Total queue wait stays bounded by the longer of the two budgets, plus a short expiry-sweep margin;
 	// a regime change grants a fresh budget but does not extend that bound, so a request that changes
 	// regime near the bound may be evicted before the fresh budget elapses.
-	// If omitted, it defaults to 60s. An explicit "0s" disables eviction while the pool is empty:
-	// requests then wait until an endpoint appears, the client disconnects, or the controller shuts
-	// down.
+	// If omitted, it follows DefaultRequestTTL, so splitting the regimes is opt-in and a configuration that
+	// sets only DefaultRequestTTL keeps that bound in both; it defaults to 60s when neither is set. An
+	// explicit "0s" disables eviction while the pool is empty: requests then wait until an endpoint
+	// appears, the client disconnects, or the controller shuts down.
 	NoEndpointRequestTTL *metav1.Duration `json:"noEndpointRequestTTL,omitempty"`
 
 	// +optional
--- a/apix/config/v1alpha1/endpointpickerconfig_types.go
+++ b/apix/config/v1alpha1/endpointpickerconfig_types.go
@@ -415,6 +415,9 @@ type FlowControlConfig struct {
 	// unavailability rather than as backpressure.
 	// The budget in force is re-evaluated while the request is queued, so a pool that scales from zero
 	// moves its queued requests onto DefaultRequestTTL, and each regime change starts a fresh budget.
+	// Total queue wait stays bounded by the longer of the two budgets, plus a short expiry-sweep margin;
+	// a regime change grants a fresh budget but does not extend that bound, so a request that changes
+	// regime near the bound may be evicted before the fresh budget elapses.
 	// If omitted, it defaults to 60s. An explicit "0s" disables eviction while the pool is empty:
 	// requests then wait until an endpoint appears, the client disconnects, or the controller shuts
 	// down.
--- a/apix/config/v1alpha1/endpointpickerconfig_types.go
+++ b/apix/config/v1alpha1/endpointpickerconfig_types.go
@@ -407,6 +407,20 @@ type FlowControlConfig struct {
 	DefaultRequestTTL *metav1.Duration `json:"defaultRequestTTL,omitempty"`
 
 	// +optional
+	// NoEndpointRequestTTL bounds queue wait while the candidate pool has no endpoints, replacing
+	// DefaultRequestTTL for as long as that holds. The two regimes want opposite budgets: with no
+	// endpoint to dispatch to, waiting is the only path to success and the budget should cover a cold
+	// start (image pull plus weight load), whereas a saturated pool should shed early enough to keep
+	// time-to-first-token within an SLO. A request that exhausts this budget is evicted as genuine
+	// unavailability rather than as backpressure.
+	// The budget in force is re-evaluated while the request is queued, so a pool that scales from zero
+	// moves its queued requests onto DefaultRequestTTL, and each regime change starts a fresh budget.
+	// If omitted, it defaults to 60s. An explicit "0s" disables eviction while the pool is empty:
+	// requests then wait until an endpoint appears, the client disconnects, or the controller shuts
+	// down.
+	NoEndpointRequestTTL *metav1.Duration `json:"noEndpointRequestTTL,omitempty"`
+
+	// +optional
 	// DefaultPriorityBand allows you to define a template for handling traffic with priority levels
 	// that are not explicitly configured in `PriorityBands`.
 	// This ensures that unforeseen traffic classes are still managed (e.g., given a default capacity
@@ -470,6 +484,10 @@ func (fcc *FlowControlConfig) String() string {
 
 	if fcc.DefaultRequestTTL != nil {
 		parts = append(parts, fmt.Sprintf("DefaultRequestTTL: %s", fcc.DefaultRequestTTL.Duration))
+	}
+
+	if fcc.NoEndpointRequestTTL != nil {
+		parts = append(parts, fmt.Sprintf("NoEndpointRequestTTL: %s", fcc.NoEndpointRequestTTL.Duration))
 	}
 
 	if fcc.DefaultPriorityBand != nil {
--- a/apix/config/v1alpha1/endpointpickerconfig_types.go
+++ b/apix/config/v1alpha1/endpointpickerconfig_types.go
@@ -397,13 +397,14 @@ type FlowControlConfig struct {
 	MaxRequests *resource.Quantity `json:"maxRequests,omitempty"`
 
 	// +optional
-	// DefaultRequestTTL bounds how long a request may wait in the queue before it is evicted.
+	// DefaultRequestTTL bounds how long a request may wait in the queue while the candidate pool has
+	// endpoints; NoEndpointRequestTTL bounds that wait while it has none.
 	// If omitted, it defaults to 60s. This is a queue-wait budget: a request that cannot dispatch
-	// within it is shed with a retryable backpressure error rather than served late, and it is the
-	// only bound on queue wait when neither the client nor the gateway enforces a request deadline.
+	// within it is shed with a retryable backpressure error rather than served late, and with no
+	// client or gateway deadline it is the only bound on waiting against a pool that has endpoints.
 	// Where such deadlines exist and fire sooner, they evict the request first (client disconnect).
-	// An explicit "0s" disables the TTL: requests then wait until client disconnect or controller
-	// shutdown.
+	// An explicit "0s" disables eviction while the pool has endpoints: such requests then wait until
+	// client disconnect or controller shutdown.
 	DefaultRequestTTL *metav1.Duration `json:"defaultRequestTTL,omitempty"`
 
 	// +optional
--- a/pkg/epp/flowcontrol/controller/controller.go
+++ b/pkg/epp/flowcontrol/controller/controller.go
@@ -57,6 +57,7 @@ type processorFactory func(
 	endpointCandidates contracts.EndpointCandidates,
 	usageLimitPolicy flowcontrol.UsageLimitPolicy,
 	clock clock.WithTicker,
+	noEndpointRequestTTL time.Duration,
 	cleanupSweepInterval time.Duration,
 	enqueueChannelBufferSize int,
 	logger logr.Logger,
@@ -150,6 +151,7 @@ func NewFlowController(
 			endpointCandidates contracts.EndpointCandidates,
 			usageLimitPolicy flowcontrol.UsageLimitPolicy,
 			clock clock.WithTicker,
+			noEndpointRequestTTL time.Duration,
 			cleanupSweepInterval time.Duration,
 			enqueueChannelBufferSize int,
 			logger logr.Logger,
@@ -164,6 +166,7 @@ func NewFlowController(
 				endpointCandidates,
 				usageLimitPolicy,
 				clock,
+				noEndpointRequestTTL,
 				cleanupSweepInterval,
 				enqueueChannelBufferSize,
 				logger,
@@ -206,6 +209,7 @@ func NewFlowController(
 		fc.endpointCandidates,
 		fc.usageLimitPolicy,
 		fc.clock,
+		fc.config.NoEndpointRequestTTL,
 		fc.config.ExpiryCleanupInterval,
 		fc.config.EnqueueChannelBufferSize,
 		fc.logger,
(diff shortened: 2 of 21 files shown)


satyamg1620:
@LukeAVanDrie Can you please review this PR

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -422,23 +438,32 @@ func (fc *FlowController) awaitFinalization(
...
+	saturationTTL := req.InitialEffectiveTTL()
+	if saturationTTL <= 0 {
+		saturationTTL = fc.config.DefaultRequestTTL
 	}
 
-	if effectiveTTL > 0 {
-		reqCtx, cancel := context.WithDeadlineCause(ctx, enqueueTime.Add(effectiveTTL), types.ErrTTLExpired)
-		return reqCtx, cancel, enqueueTime
+	// A zero budget in either regime disables eviction there, so no backstop can be derived.
+	backstop := max(saturationTTL, fc.config.NoEndpointRequestTTL)
```
LukeAVanDrie: A queued request's wait is governed by two competing eviction mechanisms:
1. The context deadline (an exact timer).
2. The sweep (polls every `ExpiryCleanupInterval`).

The problem is that only the sweep is aware of the unavailability regime, meaning only the sweep can emit `EvictedNoEndpoints`. 

In the target configuration for this feature (`noEndpointRequestTTL >= saturationTTL`), the `backstop` computed here equals `noEndpointRequestTTL`. This means the exact timer fires at the exact same moment the sweep would ordinarily evict the request for an empty pool. 

Because the exact timer always beats the polling sweep, the following sequence occurs:
* `awaitFinalization` fires first with `ErrTTLExpired`.
* `inferOutcome` inaccurately records `EvictedTTL`.
* The system relies on a secondary pool probe in `admission.go` to return a 503, which can misread the state if the pool comes up between the eviction and the probe.
* As a result, `EvictedNoEndpoints` never actually fires.

**Preferred Fix: Sweep-only enforcement:**

Make the sweep the exclusive enforcer of queue-wait budgets, leaving the request context to carry caller intent only (using `context.WithCancel(ctx)` here).

* Pros: We can delete the `max()` and backstop reasoning, outcomes/labels become correct by construction, the queued-path pool probe becomes unnecessary, and the pre-queue handoff stays bounded by the `distributeCtx` deadline you already added.
* Cons:*
  * Expiry becomes quantized to the sweep tick.
  * Total wait loses its hard cap since a flapping pool keeps granting fresh budgets (though this can be mitigated by relying on gateway-timeouts as guided in #2183).
  * `TestZombieCapacityStarvation` and the `EnqueueAndWait` unit harness lean on the context TTL and would require rework.

**Minimal-Diff Alternative: Pad the backstop:**\

If we want to keep a per-request safety net, we can pad the backstop past the sweep tick to guarantee the sweep evaluates first:

```go
backstop := max(saturationTTL, fc.config.NoEndpointRequestTTL) + 2*fc.config.ExpiryCleanupInterval
```

**Impact on Defaults:**

Either approach slightly alters the shipped defaults: an empty-pool expiry at 60s/60s becomes `EvictedNoEndpoints` at the first tick *after* 60s, rather than `EvictedTTL` + probe at exactly 60s. However, this aligns nicely with the `RejectedNoEndpoints` precedent (#1751/#2046), leaves the HTTP code and drop-reason headers unchanged, and only requires a minor update to the PR description and release notes.

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -374,7 +383,9 @@ func (p *Processor) dispatchCycle(ctx context.Context) bool {
 	}()
 
 	pool := p.endpointCandidates.Locate(ctx, nil)
-	p.poolEmpty = len(pool) == 0
+	if empty := len(pool) == 0; p.poolEmpty.Swap(empty) != empty {
```
LukeAVanDrie: The regime is published as two atomics written back to back, and the sweep reads them with two loads in `sweepFinalizedItems`. Even though the window for torn reads is small, we can close it entirely since `Run` is the only writer. A single pointer closes the window and also removes the zero-timestamp handling in the sweep:

```go
type regimeSample struct {
	empty bool
	since time.Time // zero if the regime has never changed
}

// Written only by the Run goroutine; the sweep loads one consistent sample.
regime atomic.Pointer[regimeSample]
```

pkg/epp/flowcontrol/integration_test.go
```diff
@@ -1395,3 +1400,103 @@ func TestEndpointChurnUnderLoad(t *testing.T) {
...
+	t.Parallel()
+
+	// Pool intentionally empty (no endpointCandidates set): the queue is a scale-from-zero waiting room.
+	detector := newBlockedDetector()
+	h := newHarness(t, harnessOpts{
+		detector: detector,
+		controllerCfg: &controller.Config{
+			// The saturation budget is long enough that only the no-endpoint budget can shed this request.
+			DefaultRequestTTL:        5 * time.Minute,
+			NoEndpointRequestTTL:     100 * time.Millisecond,
```
LukeAVanDrie, line 1429: This is why the suite misses the backstop issue above: the budgets here are inverted relative to the target configuration, so the backstop sits at 5m and the sweep wins comfortably.

Please add a case with `noEndpointRequestTTL >> defaultRequestTTL` asserting `EvictedNoEndpoints`. It fails on the current code, so it doubles as the regression test for the fix.

apix/config/v1alpha1/endpointpickerconfig_types.go
```diff
@@ -386,6 +386,20 @@ type FlowControlConfig struct {
...
 
+	// +optional
+	// NoEndpointRequestTTL bounds queue wait while the candidate pool has no endpoints, replacing
+	// DefaultRequestTTL for as long as that holds. The two regimes want opposite budgets: with no
+	// endpoint to dispatch to, waiting is the only path to success and the budget should cover a cold
+	// start (image pull plus weight load), whereas a saturated pool should shed early enough to keep
+	// time-to-first-token within an SLO. A request that exhausts this budget is evicted as genuine
+	// unavailability rather than as backpressure.
+	// The budget in force is re-evaluated while the request is queued, so a pool that scales from zero
+	// moves its queued requests onto DefaultRequestTTL, and each regime change starts a fresh budget.
```
LukeAVanDrie, line 397: Non-blocking: The current docstring and the release note imply that a regime change unconditionally grants a full, fresh budget. However, because the total wait is capped by the context backstop (which is set to the longer of the two budgets), this fresh budget can be truncated.

E.g., if an empty pool comes up just before `noEndpointRequestTTL` expires, the request nominally receives a fresh saturation budget. But because the overall backstop timer is almost up, the request will be shed with a 429 (backpressure) shortly after becoming dispatchable.

I recommend adding a single sentence to explicitly cover this cap. For example:

> Total queue wait remains bounded by the longer of the two budgets; a regime change does not extend that overall bound.

LukeAVanDrie (changes requested):

pkg/epp/flowcontrol/controller/controller.go
LukeAVanDrie: @ahg-g I am comfortable merging this with this one-line fix mentioned above (plus regression test + updating the release note):

`backstop := max(saturationTTL, fc.config.NoEndpointRequestTTL) + 2*fc.config.ExpiryCleanupInterval
`

I would like to land a cleaner pattern here though, so this would be an intermediary state for #2183, and I would want a followup issue / PR (can land in v0.11). 

The other torn-read fix is small, but is so is the race window, so that can also reasonably be deferred.

LukeAVanDrie (dismissed):
Thanks for the thorough follow-up! This addresses all my feedback. I still want to land the cleaner sweep-only enforcement pattern (drop the backstop/max() reasoning and the queued-path probe) as a follow-up; I'll file a dedicated issue for it targeting v0.11, since #2183 closes with this merge.

LukeAVanDrie:
@vMaroon; this LGTM, just need a maintainer to approve for the non-flow control paths (config loader + APIX change)

LukeAVanDrie:
@satyamg1620 whenever you're ready, can you rebase this onto upstream/main?

satyamg1620:
@vMaroon Can you please review and merge the PR

ahg-g (approved):

PR #2289, opened 2026-08-05
fix: add dedicated APITypeMessages for Anthropic Messages API in routing sidecar

Gregory-Pereira:
**What type of PR is this?**
/kind bug

The `/v1/messages` route was aliased to `APITypeChatCompletions`, causing unused and ignored OpenAI-specific fields (`max_completion_tokens`, `max_output_tokens`, `min_tokens`) to be included prefill request bodies during disaggregated inference. The Anthropic Messages API only uses `max_tokens`.

Adds `APITypeMessages` with its own token limit field set and threads `apiType` through the parallel dispatch and EC connector paths so each API type only touches its own fields.

cc @tessapham

### DIFF
--- a/pkg/sidecar/proxy/connector_nixlv2.go
+++ b/pkg/sidecar/proxy/connector_nixlv2.go
@@ -75,7 +75,7 @@ func (s *Server) handleNIXLV2(w http.ResponseWriter, r *http.Request, prefillPod
 	if s.config.MoRIIOParallelDispatch && s.config.MoRIIOWriteMode {
 		// MoRI-IO requires transfer_id to carry the "tx" prefix for message routing.
 		transferID := "tx" + uuidStr
-		s.runNIXLProtocolV2WriteParallel(w, r, original, completionRequest, uuidStr, transferID, prefillPodHostPort, kvCacheSource)
+		s.runNIXLProtocolV2WriteParallel(w, r, original, completionRequest, uuidStr, transferID, prefillPodHostPort, kvCacheSource, apiType)
 		return
 	}
 
@@ -444,6 +444,7 @@ retryLoop:
 func (s *Server) runNIXLProtocolV2WriteParallel(
 	w http.ResponseWriter, r *http.Request, original []byte,
 	completionRequest map[string]any, uuidStr, transferID, prefillPodHostPort, kvCacheSource string,
+	apiType APIType,
 ) {
 	s.logger.V(4).Info("running NIXL protocol V2 (concurrent dispatch)",
 		"url", prefillPodHostPort, "request_id", uuidStr)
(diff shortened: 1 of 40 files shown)


pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -444,6 +444,7 @@ retryLoop:
 func (s *Server) runNIXLProtocolV2WriteParallel(
 	w http.ResponseWriter, r *http.Request, original []byte,
 	completionRequest map[string]any, uuidStr, transferID, prefillPodHostPort, kvCacheSource string,
```
tessapham, line 446: Maybe rename `completionRequest` can be renamed to just `request` or `requestData` to be generic.

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -456,10 +457,21 @@ func (s *Server) runNIXLProtocolV2WriteParallel(
...
+
+	type savedField struct {
+		field   string
+		val     any
+		present bool
+	}
+	tokenLimitFields := tokenLimitFieldsForAPIType(apiType)
+	savedTokenValues := make([]savedField, len(tokenLimitFields))
+	for i, field := range tokenLimitFields {
+		if v, ok := completionRequest[field]; ok {
```
tessapham, line 469: Do we want to use `tokenMap[field]` instead of `completionRequest[field]`, similar to L117-L123 in the serial path?

```
// tokenLimitMap returns the map holding the token-limit fields: sampling_params
// for the generate API (created if absent), or the request itself otherwise.
```

Not sure if if Generate API needs to be considered in the parallel path.

pkg/sidecar/proxy/connector_nixlv2.go
```diff
@@ -501,10 +513,9 @@ func (s *Server) runNIXLProtocolV2WriteParallel(
 
 	completionRequest[requestFieldStream] = false
 	delete(completionRequest, requestFieldStreamOptions)
-	completionRequest[requestFieldMaxTokens] = 1
-	completionRequest[requestFieldMaxCompletionTokens] = 1
-	completionRequest[requestFieldMaxOutputTokens] = 1
-	completionRequest[requestFieldMinTokens] = 1
+	for _, field := range tokenLimitFields {
+		completionRequest[field] = 1
```
tessapham, line 517: Ditto - either rename or use `tokenLimitMap` instead.

tessapham (commented):
A couple comments on first review.

I think it's logical to make max tokens in `PrimeSingleTokenRequest()` a variable in this PR as well, as you mentioned offline, but up to you to include it or not.

ahg-g:
@Gregory-Pereira is this still relevant? if this is a bug, can we please move it forward?

ahg-g:
what is the status of this PR?

Gregory-Pereira:
> what is the status of this PR?

Just sitting down to take another pass and address comments now. Sorry its taken so long to get back to it

PR #2290, opened 2026-08-05
feat(epp): add port parameter to metrics-data-source

kaushikmitr:
**What type of PR is this?**
/kind feature

Adds an optional `port` parameter to the `metrics-data-source` plugin that overrides the scrape port per data source, mirroring the existing `dcgm-data-source` port option.

The scrape port always follows the InferencePool `targetPorts`. When a sidecar fronts the model server on the pool target port (as on Vertex AI managed serving, where an access sidecar terminates traffic on :15080 and the engine serves `/metrics` only on :7080), the engine metrics endpoint cannot be scraped: the sidecar returns 404 for `/metrics`, and every metrics-backed plugin (utilization-filter server conditions, kv/queue scorers) reads zeros. With this parameter:

```yaml
- type: metrics-data-source
  parameters:
    port: 7080
```

routing continues to target the pool port while metrics are scraped from the engine port.

None

**Release note**:
```release-note
The metrics-data-source plugin accepts an optional `port` parameter that overrides the metrics scrape port when it differs from the InferencePool target port (for example when a sidecar fronts the model server).
```

**Test plan**:
- [x] Factory accepts the port parameter (with strict decoding) and constructs the data source; zero/omitted port keeps default behavior
- [x] Existing metrics data source unit tests pass

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/source/metrics/datasource.go
+++ b/pkg/epp/framework/plugins/datalayer/source/metrics/datasource.go
@@ -18,6 +18,7 @@ package metrics
 
 import (
 	"encoding/json"
+	"fmt"
 	"io"
 
 	"github.com/prometheus/common/expfmt"
@@ -43,7 +44,7 @@ type metricsDatasourceParams struct {
 	Scheme string `json:"scheme"`
 	// Path defines the URL path used in metrics retrieval (e.g., "/metrics").
 	Path string `json:"path"`
-	// Port, when non-zero, overrides the endpoint's inference port for metrics retrieval.
+	// Port, when positive, overrides the endpoint's inference port for metrics retrieval.
 	// Use when the model server exposes metrics on a port other than the InferencePool target port.
 	Port int `json:"port"`
 	// InsecureSkipVerify defines whether model server certificate should be verified or not.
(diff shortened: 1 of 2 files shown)


pkg/epp/framework/plugins/datalayer/source/metrics/datasource.go
```diff
@@ -43,6 +44,12 @@ type metricsDatasourceParams struct {
 	Scheme string `json:"scheme"`
 	// Path defines the URL path used in metrics retrieval (e.g., "/metrics").
 	Path string `json:"path"`
+	// Port, when positive, overrides the endpoint's inference port for metrics retrieval.
+	// Use when the model server exposes metrics on a port other than the InferencePool target port.
+	// The override applies to every endpoint of the source. Do not set it on pools with
+	// multiple target ports (data-parallel ranks): each rank must be scraped on its own
+	// port, and a single override would point all ranks at the same one.
+	Port int `json:"port"`
```
ahg-g: pls use a pointer instead, when set validate and use, when not set, default to the existing behavior.

kaushikmitr: Done. Port is now *int: nil keeps the endpoint's inference port, a set value is validated to 1-65535 and applied (explicit 0 is rejected)

ahg-g (approved):

PR #2292, opened 2026-08-05
feat(scheduler): size approx-prefix-cache LRU to offload-extended KV cache capacity

### LINKED ISSUE #1650: `approx-prefix-cache-producer` sizes its LRU to HBM only, ignoring KV‑cache offload → prefix‑match estimates are wrong for every consumer
## Summary

On deployments with vLLM **KV‑cache offloading** enabled, the `approx-prefix-cache-producer`'s autotune sizes its per‑endpoint LRU from `num_gpu_blocks` (HBM only). The real effective cache is HBM **+ host/CPU offload**, which is much larger. The producer's LRU therefore evicts prefix entries that the actual cache still holds, so its **per‑endpoint prefix‑match estimates (`matchedBlocks / totalBlocks`) are systematically under‑reported.**

This is a **producer‑side data correctness bug**, not a routing‑policy bug. Any plugin that consumes the prefix‑match data is fed wrong values and silently degrades — e.g. the **`prefix-cache-scorer`** (returns near‑zero scores → no routing preference) and the **`prefix-cache-affinity-filter`** (finds no endpoint above `affinityThreshold` → "no sticky endpoints"). The deployment looks configured and healthy; prefix‑aware routing just doesn't do anything.

## Environment

- **llm-d-router** EPP (main latest)
- **vLLM** with KV offloading via the native offload connector (`kv_offloading_backend="native"`)
- Model: `Qwen/Qwen3-Coder-480B-A35B-Instruct-FP8`
- Producer: `approx-prefix-cache-producer`; observed consumers: `prefix-cache-scorer`, `prefix-cache-affinity-filter`
- Workload: agentic multi‑turn `conversation_replay` (large per‑conversation system prompts ~160K tokens)

## Root cause

1. The producer autotune derives LRU capacity from the endpoint's `num_gpu_blocks × block_size` — **HBM only**.
   - Measured: `num_gpu_blocks=1570`, `block_size=256` → **~402K tokens/endpoint**.
2. The working set exceeds that: ~5 conversations/endpoint × ~160K‑token prefixes ≈ **~800K tokens** → the producer's LRU evicts a conversation's prefix **before its next turn**.
3. But the **real** cache (HBM + host offload) still holds those prefixes — the offload tier is doing most of the caching.
4. So the producer reports `matchedBlocks / totalBlocks` far lower than the true cached fraction. **Every consumer of that score is now wrong** — independent of which routing plugin is configured.

## Evidence

- **Producer output is under‑counted:** its LRU models ~402K tokens/endpoint while the real (offload‑extended) cache retains much more — vLLM external (offload) prefix‑cache hit rate ≈ **45%** vs on‑device ≈ **4.7%**. The prefixes the producer assumes are evicted are in fact still cached.
- **Symptom via one consumer (affinity filter, `-v=4`):** essentially **every** decision is `PrefixCacheAffinityFilter: no sticky endpoints` (all endpoints score < 0.80); `narrowed to sticky` ≈ 0. A `prefix-cache-scorer` fed the same under‑counted matches would likewise produce near‑zero, undifferentiated scores → no routing preference. Same root cause, different consumer.
- vLLM metrics: `vllm:cache_config_info{ num_gpu_blocks="1570", block_size="256", kv_offloading_backend="native", num_cpu_blocks="None", kv_offloading_size="None" }`.
- The EPP endpoint model already carries a `KvCacheMaxTokenCapacity` field, but it is **`0` (unpopulated)** — the slot for "true capacity" exists; the data isn't flowing.
- **Manual fix validated:** setting on the producer
  ```yaml
  autoTune: false
  blockSizeTokens: 256
  lruCapacityPerServer: 1570000   # = num_gpu_blocks × ~1000, sized to cover the offload-extended cache
  ```
  restores correct prefix matches. Same workload/seed, vs the HBM‑only config: **conc=10 → TTFT −34%, TPOT −34%, throughput +10%** (gains grow where prefix reuse matters).

## Impact

- The producer's prefix‑match estimates are wrong on **any** offloaded deployment, so **every prefix‑aware plugin is silently degraded**:
  - `prefix-cache-scorer`: low/undifferentiated scores → routing ignores cache locality.
  - `prefix-cache-affinity-filter`: no endpoint clears the threshold → filter is a no‑op.
- Knobs on those consumers (scorer weight, `affinityThreshold`, `maxTTFTPenaltyMs`) have **no observable effect**, because their *input* is wrong. (We swept `maxTTFTPenaltyMs` 5k–250k at conc=40 — completely flat — since the gate it controls never had a sticky endpoint to act on.)
- Failure is **silent**: config looks correct, system is healthy; only producer/consumer debug logs reveal the no‑op.

## Proposed fix (llm-d-router)

Fix it once, at the producer — that corrects all consumers simultaneously. In the `approx-prefix-cache-producer` autotune:

1. **Detect offload** from `cache_config_info.kv_offloading_backend` (anything other than `none`/empty), or the presence of `vllm:external_prefix_cache_*` counters. (The producer already scrapes `cache_config_info` for `num_gpu_blocks`/`block_size`.)
2. **Size the LRU to total physical cache when offload is detected**, not HBM:
   ```
   if offloadDetected:
       if num_cpu_blocks (or offload size) known:
           lruCapacity = num_gpu_blocks + num_cpu_blocks      # exact
       else:                                                  # size unreported today
           lruCapacity = num_gpu_blocks * offloadLruMultiplier   # configurable; default e.g. 10–20×
           log.Warn("offload detected but capacity unreported; using multiplier — set lruCapacityPerServer to override")
   else:
       lruCapacity = num_gpu_blocks                           # current behavior (HBM only)
   ```
3. Keep the explicit `lruCapacityPerServer` override (the current escape hatch).
4. Log the detected offload backend and the chosen LRU capacity + source (exact vs multiplier vs override).

## Upstream dependency (vLLM)

For *exact* automatic sizing, vLLM should expose the offload capacity in metrics — populate `num_cpu_blocks` / `kv_offloading_size` in `cache_config_info`, or populate the `KvCacheMaxTokenCapacity` the EPP already reads. With native offload these are currently `None`/`0`. (Worth a paired vLLM issue.)

## Workaround (today)

Disable autotune and size the LRU to the offload‑extended cache explicitly:
```yaml
- type: approx-prefix-cache-producer
  parameters:
    autoTune: false
    blockSizeTokens: 256          # match vLLM block_size
    lruCapacityPerServer: 1570000 # ~ num_gpu_blocks × 1000; size to cover the working set under offload
    maxPrefixTokensToMatch: 262144
```

## Acceptance criteria

- With offloading enabled and **default** producer config, the producer's prefix‑match estimates reflect the offload‑extended cache (not HBM only), so downstream prefix‑aware plugins (`prefix-cache-scorer`, `prefix-cache-affinity-filter`) see correct scores **without** a manual `lruCapacityPerServer` override.
- A startup/INFO log states: offload detected (backend), chosen LRU capacity, and whether it came from exact capacity, a multiplier, or an explicit override.

BenjaminBraunDev:
**What type of PR is this?**
/kind bug

The `approx-prefix-cache-producer` autotune sizes its per-endpoint LRU from `num_gpu_blocks` (HBM only). With KV cache offloading the actual cache is HBM plus a host tier, so the LRU evicts prefixes the server still holds and consumers (`prefix-cache-scorer`, `prefix-cache-affinity-filter`) get under-counted matches. See the issue for root cause and benchmarks.

Approach: when the model server reports its total cache size (GPU plus CPU), we size each pod's LRU to that total. When it reports only the GPU cache size but we can see offloading is enabled, we keep the GPU-only size and log a warning to set `lruCapacityPerServer` manually. If a pod's reported size changes later, its LRU is resized in place to match.

Per-engine behavior:

- **SGLang**: v0.5.10 and later report the CPU cache size as `sglang:hicache_host_total_tokens`, so we size to `max(hicache_host_total_tokens, max_total_num_tokens)`. Max rather than sum, because the CPU cache also holds a copy of everything in the GPU cache. Earlier versions only report the GPU cache size (`sglang:max_total_num_tokens`), so we size to that, and `lruCapacityPerServer` covers the rest.
- **vLLM**: no version reports the CPU cache size. We can see that offloading is enabled (the connector's `vllm:kv_offload_*` metrics appear at startup) but not how big it is, so we size to the GPU cache (`num_gpu_blocks`) and log the warning. Exact sizing needs a new `OffloadingConnector` metric: vllm-project/vllm#29605 tried populating the legacy `num_cpu_blocks` field and was rejected because offloading is now a connector concern, the maintainer wants the offload connector to export its own capacity metric instead.
- **trtllm-serve**: `trtllm_kv_cache_max_blocks` already counts GPU and CPU cache blocks together, so sizing was already correct and nothing changes. The exception is the experimental V2 cache manager, where this metric is GPU-only, so those deployments need `lruCapacityPerServer`.

Unrelated fix picked up along the way: newer SGLang (v0.5.16) renamed its block size and block count metrics, so the default SGLang mapping now reads both the old and new names.

Fixes #1650

**Testing**:

- [x] Unit tests: extractor (max semantics, detection via series and label) and producer (capacity order, LRU resize)
- [x] Fixture tests: default engine mappings run against real `/metrics` scrapes, checked in as `testdata/*.prom`
- [x] Live verification on 8 deployments (2x8xH200), summarized below

All vLLM rows are v0.26.0 (latest release at time of testing). SGLang versions tested were v0.5.16 and v0.5.9 respectively. Both trtllm rows are v1.3.0rc23.

| Deployment | Observed | Outcome |
| --- | --- | --- |
| vLLM, no offload | no `kv_offload_*` series, label `None` | correct |
| vLLM, offload enabled | `kv_offload_*` series present via both `--kv-offloading-size` and `--kv-transfer-config`, label set or `None` respectively | undercounts + warning |
| SGLang v0.5.10+, hicache on | `hicache_host_total_tokens = 400001`, matching `--hicache-ratio 2` x 200000 GPU tokens | correct |
| SGLang pre-v0.5.10, hicache on | CPU metric absent, only `max_total_num_tokens = 200000` | undercounts |
| trtllm legacy manager, 4GiB host cache | `max_blocks = 21129`, host blocks included | correct |
| trtllm V2 manager, 4GiB host cache | `max_blocks = 19958`, GPU blocks only | undercounts |

The 4GiB host cache is ~1170 blocks at ~3.6MiB per 32-token block, matching the 21129 vs 19958 gap between the two managers.

In the undercount cases, user must turn `autotune: false` and manually set `lruCapacityPerServer` to override the metric manually. There is unfortunately no Prometheus visibility into the SGLang and trtllm undercounts so we can't log a warning.

**PR size breakdown** (most of the diff is tests):

- ~295 lines of code and docs (extractor specs, producer sizing, LRU resize, README and metric protocol docs)
- ~557 lines of test code (unit tables plus fixture tests)
- ~307 lines of `testdata/*.prom`, real `/metrics` scrapes from the deployments above, so the default engine mappings stay verified against actual engine output

**Release note**:
```release-note
The approx-prefix-cache-producer now sizes its per-endpoint LRU to the offload-extended KV cache capacity: SGLang `hicache_host_total_tokens` is used when present (v0.5.10+), vLLM KV uses GPU blocks + undercount warning, trtllm is already the combined amount unless v2 kv-cache manager is used which undercounts.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/README.md
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/README.md
@@ -16,8 +16,10 @@ For each request, the plugin consumes `request.Body.TokenizedPrompt` (token IDs)
 - `maxPrefixTokensToMatch` (int, optional, default: `131072`): Cap expressed in tokens instead of blocks (`maxBlocks = maxPrefixTokensToMatch / blockSizeTokens`). Not auto-tuned. Takes precedence over `maxPrefixBlocksToMatch` when set (> 0); set to `0` to fall back to the block-based cap. The `131072` default (128K, the context window of large production models such as gpt-oss 120b) is a reasonable upper bound that covers the long-prompt use cases seen in production.
 
 Setting both caps to `0` matches the whole prompt. Prompt length is bounded by the model server's context window, so this caps matching at one context window without needing to name a token count. It raises EPP CPU on long prompts (see [operations](../../../../../../../docs/operations.md)); to turn prefix matching off, remove this producer from the configuration rather than zeroing the caps.
-- `lruCapacityPerServer` (int, optional, default: `31250`): Per-pod LRU index capacity. Used when `autoTune` is false or endpoint metrics are unavailable; ignored when auto-tuned from metrics.
+- `lruCapacityPerServer` (int, optional, default: `31250`): Per-pod LRU index capacity. Used when `autoTune` is false or endpoint metrics are unavailable; ignored when auto-tuned from metrics. Also the escape hatch for deployments with KV cache offloading whose model servers report only GPU capacity (e.g. vLLM with an offload connector, SGLang older than v0.5.10 with hicache, trtllm-serve's V2 KV cache manager): disable `autoTune` and size this to the offload-extended cache. A warning is logged when offloading is detected but unaccounted for (vLLM OffloadingConnector metrics) so the undercount is not silent.
 - `blockSize` (int, optional): Deprecated — character-based block size. Use `blockSizeTokens` instead.
+
+**Auto-tuned capacity resolution order** (per pod, when `autoTune` is true): total token capacity metric across tiers divided by the server block size (e.g. SGLang `max(hicache_host_total_tokens, max_total_num_tokens)`), then GPU blocks, then `lruCapacityPerServer`, then the built-in default. The chosen capacity and its source are logged when a pod's LRU is created or resized.
 
 **Configuration Examples:**
 
--- a/docs/plugin-metric-protocol.md
+++ b/docs/plugin-metric-protocol.md
@@ -86,6 +86,40 @@ If absent, the value is taken from the `approximate-prefix` plugin's `LRUCapacit
 | Triton TensorRT-LLM | `nv_trt_llm_kv_cache_block_metrics{kv_cache_block_type=max}` | — |
 | trtllm-serve | `trtllm_kv_cache_max_blocks` | — |
 
+Note on trtllm-serve with host offloading (`kv_cache_config.host_cache_size`): under the default
+(legacy) KV cache manager, `trtllm_kv_cache_max_blocks` already counts GPU plus host blocks, so
+the auto-tuned capacity covers the offload-extended cache with no further configuration. Under
+the opt-in `use_kv_cache_manager_v2` (also the default for some hybrid-Mamba model families),
+the same gauge is GPU-only and the host tier is not observable via Prometheus; on such
+deployments disable the `approx-prefix-cache-producer`'s `autoTune` and set
+`lruCapacityPerServer` explicitly to avoid under-reporting prefix matches.
+
+### TotalKVCacheTokens (optional)
+
+The total effective KV cache capacity in tokens across all memory tiers (GPU HBM plus host/CPU
+offload). When reported, it takes precedence over NumGPUBlocks for auto-tuning the approximate
+prefix cache, so prefix-match estimates cover the offload-extended cache.
+
+- **Type:** Gauge (the maximum across the configured per-tier gauges is used)
+- **Required by:** `prefix-cache-scorer`, `prefix-cache-affinity-filter` (via `approximate-prefix` when `AutoTune` is enabled)
+
+| Model server | Metric | Notes |
+| --- | --- | --- |
+| SGLang | `sglang:hicache_host_total_tokens`, `sglang:max_total_num_tokens` | The hicache host pool (SGLang v0.5.10+, hierarchical cache enabled) is an inclusive superset of the device pool under the default `write_through` policy, so the effective total is the max of the two gauges, not their sum. On older SGLang versions only the device gauge exists; disable `autoTune` and set `lruCapacityPerServer` to `hicache-ratio x device tokens / page size` to account for the host tier. |
+| vLLM | — | Not reported. vLLM exposes no offload-tier capacity metric (`num_cpu_blocks` is always `None` on V1). See KVCacheOffloadDetection below. |
+| trtllm-serve | — | Not needed under the legacy KV cache manager (NumGPUBlocks already includes host blocks); not available under the V2 manager. |
+
+### KVCacheOffloadDetection (optional)
+
+Metrics whose presence indicates the model server offloads KV cache to a tier whose capacity is
+not reported. Detection does not change auto-tuned capacity by itself; it drives an operator
+warning that prefix matches will be under-reported unless `autoTune` is disabled and
+`lruCapacityPerServer` is set.
+
+| Model server | Signal |
+| --- | --- |
+| vLLM | Presence of any `vllm:kv_offload_*` OffloadingConnector series (registered at startup, before traffic), or a numeric `kv_offloading_size` label (GiB) on `vllm:cache_config_info` (only populated by the built-in `--kv-offloading-size` flag; hand-written `--kv-transfer-config` deployments leave it `None`). `vllm:external_prefix_cache_*` is deliberately not used: it is registered even without a connector. |
+
 ## LoRA Adapter Serving
 
 **Required by:** `lora-affinity-scorer`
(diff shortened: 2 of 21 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/README.md
```diff
@@ -16,9 +16,11 @@ For each request, the plugin consumes `request.Body.TokenizedPrompt` (token IDs)
...
+- `lruCapacityPerServer` (int, optional, default: `31250`): Per-pod LRU index capacity. Used when `autoTune` is false or endpoint metrics are unavailable; ignored when auto-tuned from metrics. Also the escape hatch for deployments with KV cache offloading whose model servers report only GPU capacity (e.g. vLLM with an offload connector, SGLang older than v0.5.10 with hicache, trtllm-serve's V2 KV cache manager): disable `autoTune` and size this to the offload-extended cache. A warning is logged when offloading is detected but unaccounted for (vLLM OffloadingConnector metrics) so the undercount is not silent.
 - `blockSize` (int, optional): Deprecated — character-based block size. Use `blockSizeTokens` instead.
 
+**Auto-tuned capacity resolution order** (per pod, when `autoTune` is true): total token capacity metric across tiers divided by the server block size (e.g. SGLang `max(hicache_host_total_tokens, max_total_num_tokens)`), then GPU blocks, then `lruCapacityPerServer`, then the built-in default. The chosen capacity and its source are logged when a pod's LRU is created or resized.
```
kaushikmitr, line 22: maybe allow users to configure such then they want to size lru based on gpu, cpu, or sum total (default sum total)

BenjaminBraunDev: I don't think engines give us enough for a gpu/cpu/sum knob today. For instance SGLang is the only engine that exports both numbers and even then summing is wrong: the CPU pool holds a copy of everything in the GPU pool, which is why autotune does max(cpu, gpu). trtllm reports one already-combined number we can't split (without hitting the json metrics endpoint separate form epp podmetrics), and vLLM reports GPU only, so there is nothing to sum anywhere.

Issue #1632, opened 2026-06-12
Tracing initialization is not production ready (TLS/auth, default exporter, sampler support, env mutation)


gyliu513:
**What would you like to be added**:

Make the OpenTelemetry tracing initialization usable against a real production backend, not just a local plaintext collector. Four concrete improvements:

1. OTLP transport security and protocol. The OTLP exporter is currently created with `WithInsecure` hardcoded:

```go
traceExporter, err = otlptracegrpc.New(ctx, otlptracegrpc.WithInsecure())
```

Because `WithInsecure` is passed as an explicit option, it overrides the standard insecure environment variable and forces a plaintext connection, and there is no OTLP/HTTP code path (only gRPC). Allow TLS and authentication to be configured (respect the standard insecure, headers, and certificate environment variables instead of forcing insecure), and consider supporting OTLP/HTTP as an alternative transport.

2. Default exporter. When the exporter type is not set, it defaults to `console`, which uses a stdout pretty printer:

```go
stdouttrace.New(stdouttrace.WithPrettyPrint())
```

Change the default away from console pretty print (to `otlp`, or require the exporter to be configured explicitly), so that enabling tracing without configuration does not flood stdout with multi line JSON.

3. Sampler support and validation. Only `parentbased_traceidratio` is handled; the standard values `always_on`, `always_off`, and `traceidratio` all fall through and quietly revert to a 0.1 ratio. Separately, an unparseable sampler argument is silently replaced with 0.1:

```go
if samplerType == "parentbased_traceidratio" {
    fraction, err := strconv.ParseFloat(samplerARG, 64)
    if err != nil {
        fraction = 0.1
    }
    sampler = sdktrace.ParentBased(sdktrace.TraceIDRatioBased(fraction))
} else {
    // unsupported sampler type, fall back to 0.1
}
```

Support the standard sampler values, and at minimum log a warning (or fail fast) when the sampler argument is invalid instead of silently substituting 0.1.

4. Configuration without mutating process environment. Defaults are currently applied by writing back into the process environment so the SDK picks them up later:

```go
os.Setenv("OTEL_SERVICE_NAME", defaultServiceName)
os.Setenv("OTEL_EXPORTER_OTLP_ENDPOINT", "http://localhost:4317")
```

Build the resource attributes and exporter options explicitly and hand them to the SDK, rather than writing values into the environment.

**Why is this needed**:

[...]

PR #2298, opened 2026-08-06
tracing: configure SDK explicitly instead of mutating process env

### LINKED ISSUE #2297: tracing: configure the SDK explicitly instead of mutating process environment

Remove the `os.Setenv` calls in `InitTracing` (`pkg/common/observability/tracing/telemetry.go`) and configure the SDK directly.

Build the resource with `resource.New`, applying the built-in defaults (`service.name` from the caller-supplied name, `service.version` from `version.BuildRef`) before `resource.WithFromEnv()`. Detectors are merged in order with later ones winning, so placing the environment detector last keeps `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES` overriding the built-in service name. That precedence is the main thing to get right and to test.

Two constraints on the implementation:

- **Do not merge in `resource.Default()`.** It carries semantic conventions v1.41.0 (from `go.opentelemetry.io/otel/sdk` v1.44.0), while this package uses v1.37.0. `resource.Merge` on two differing non-empty schema URLs returns `ErrSchemaURLConflict` and drops the schema URL from the result, which would silently strip the schema URL from every span's resource. Aligning the semconv version is a separate change.

- **Remove the default OTLP endpoint rather than passing it as an exporter option.** The SDK already defaults the gRPC exporter to `localhost:4317`, so the value is redundant. Passing it explicitly would be worse than removing it: `NewGRPCConfig` applies explicit options after environment configuration, so an explicit endpoint option would override an operator's `OTEL_EXPORTER_OTLP_ENDPOINT`.

**Why is this needed**:

Writing into the process environment has side effects on any other code in the same process that reads those variables, is not reentrant, and makes tests interfere with each other. `InitTracing` has no test coverage at all; the environment mutation is what makes it untestable, so this is a prerequisite for meaningfully testing the rest of #1632.

**Acceptance criteria**:

- No `os.Setenv` in the `tracing` package.
- `service.name` defaults to the caller-supplied name (`llm-d-epp`, `llm-d-disagg-sidecar`).
- `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES` override that default, with `OTEL_SERVICE_NAME` taking precedence between the two.
- Attributes injected by the Helm chart (`k8s.pod.name`, `k8s.node.name`) land on the resource alongside the defaults.
- The resource retains its schema URL.
- Tests covering the above, plus a guard asserting `InitTracing` leaves the `OTEL_*` variables untouched in the process environment.

Part of #1632.

satyamg1620:
**What type of PR is this?**
/kind bug

`InitTracing` applied its defaults by calling `os.Setenv` for `OTEL_SERVICE_NAME` and `OTEL_EXPORTER_OTLP_ENDPOINT`, relying on the SDK to read them back. Mutating the process environment affects any other code in the process that reads those variables, is not reentrant, and leaves tests interfering with each other — `InitTracing` had no test coverage at all as a result.

The resource is now built directly via `resource.New`. Detectors are applied in order and later ones win, so placing `WithFromEnv` last keeps `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES` overriding the built-in service name.

This also fixes trace attribution. The previous resource was built with `resource.NewWithAttributes`, which never consults the environment, so neither `service.name` nor the chart's `OTEL_RESOURCE_ATTRIBUTES` (`k8s.namespace.name`, `k8s.node.name`, `k8s.pod.name`) ever reached a span — the `os.Setenv` calls had no effect on the exported resource. Spans now carry them.

Three implementation notes:

- **The plaintext loopback default is pinned, conditionally.** The SDK defaults the gRPC endpoint to `localhost:4317` but negotiates TLS for it, and #2312 removed the hardcoded `WithInsecure`. Simply dropping the old default therefore pointed the local development path at TLS against a plaintext collector. `localCollectorOptions` supplies the endpoint and the insecure transport only when the environment sets none of the endpoint, insecure or certificate variables; as soon as one is set the options are dropped, so an operator's configuration is never overridden.

- **A partial resource no longer aborts startup.** `resource.New` reports a malformed `OTEL_RESOURCE_ATTRIBUTES` entry alongside a resource holding everything it could parse. Returning that error made a single stray comma in a chart value fail `InitTracing`, which the EPP runner turns into a process exit. The error is logged and the degraded resource is used.

- **`resource.Default()` is not merged in.** It carries semconv v1.41.0 while this package uses v1.37.0, and merging conflicting schema URLs drops the schema URL from the result. Aligning the semconv version is a separate change; the reason is recorded in a comment so it is not re-added by accident.

Fixes #2297

Part of #1632.

**Test plan**:

- [x] `service.name` defaults to the caller-supplied name when the environment is empty
- [x] `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES` override that default, with `OTEL_SERVICE_NAME` winning between the two
- [x] Chart-injected attributes land alongside the defaults, and the schema URL is preserved
- [x] `InitTracing` leaves every `OTEL_*` variable untouched in the environment
- [x] A malformed `OTEL_RESOURCE_ATTRIBUTES` entry keeps the parseable attributes and lets initialization succeed
- [x] With no transport variables set, the exporter reaches a plaintext collector on the loopback default
- [x] An operator's endpoint, insecure flag and certificate still decide the transport (existing OTLP coverage from #2312 unchanged)
- [x] Tests pass with `OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES` and `OTEL_EXPORTER_OTLP_ENDPOINT` already set in the environment
- [x] Verified on a kind dev stack with an OpenTelemetry collector sidecar on the pod loopback: spans arrive carrying `service.name=llm-d-epp` with no `OTEL_SERVICE_NAME` set, plus `k8s.namespace.name`, `k8s.node.name` and `k8s.pod.name`, at schema URL `1.37.0`

```release-note
Spans exported by EPP and the disaggregated sidecar now carry `service.name` and the attributes from `OTEL_RESOURCE_ATTRIBUTES`, including the `k8s.namespace.name`, `k8s.node.name` and `k8s.pod.name` values set by the Helm chart. These were previously never applied to the exported resource, so spans reached the collector without service identity.
```


pkg/common/observability/tracing/telemetry.go
```diff
@@ -49,14 +49,10 @@ func InitTracing(ctx context.Context, logger logr.Logger, defaultServiceName str
...
 	loggerWrap := &errorHandler{logger: logger}
 
-	_, ok := os.LookupEnv("OTEL_SERVICE_NAME")
-	if !ok {
-		os.Setenv("OTEL_SERVICE_NAME", defaultServiceName)
-	}
-
-	_, ok = os.LookupEnv("OTEL_EXPORTER_OTLP_ENDPOINT")
-	if !ok {
-		os.Setenv("OTEL_EXPORTER_OTLP_ENDPOINT", "http://localhost:4317")
```
ahg-g, line 59: How is this default value applied now?

satyamg1620: Thanks @ahg-g for the review.

The SDK supplies the endpoint: NewGRPCConfig seeds localhost:4317 before it applies env config or explicit options ([options.go#L115-L131](https://github.com/open-telemetry/opentelemetry-go/blob/v1.44.0/exporters/otlp/otlptrace/otlptracegrpc/internal/otlpconfig/options.go#L115-L131)). But it negotiates TLS for that default, and #2312 removed the hardcoded WithInsecure — so the http:// scheme in the old os.Setenv value was what selected plaintext. Dropping it outright broke the local collector path: tls: first record does not look like a TLS handshake, 0 spans delivered.

Pushed a fix that pins the endpoint and the insecure transport, but only when the environment sets none of the endpoint, insecure or certificate variables. As soon as one is set the options are dropped, so an operator's OTEL_EXPORTER_OTLP_ENDPOINT and its scheme still decide where spans go and how the connection is secured.

satyamg1620:
@ahg-g @elevran Can you please review this PR

ahg-g (approved):

Issue #2282, opened 2026-08-05
[Coordinator] Relay Prefill endpoint metadata to Decode endpoint


elevran:
**What would you like to be added**:
Specify headers copied by the coordinator from the prefill response into the decide request.
These carry metadata regarding the routing decisions made in one phase into the other.

**Why is this needed**:
Alogrithms that require cross phase synchronization communicate using HTTP headers. Example include
- Preferring P and D nodes based on topology (e.g., select D on the same node as P)
- roll out version subsets (during rollout of new P/D versions, enforce consistent version across P and D phases)

elevran:
xref #2301

roytman:
We probably need broader design decisions here.

- The Coordinator now supports multiple fan-out steps, meaning it acts as both the encoder and the renderer ([#issuecomment-5207653711](https://github.com/llm-d/llm-d-router/pull/2301#issuecomment-5207653711)); we need to define the merging policy.
- How should we handle a situation where a header from the original user request conflicts with the same header (with a different value) returned in one of the step responses??
- Should a returned header be forwarded only to the immediate next step, or to all subsequent steps?
   - For example, assume step A returns header R.
   - The Coordinator copies R into the request to step B.
   - Step B does not return header R
   - Should the Coordinator drop header R when calling step C?

PR #2299, opened 2026-08-06
Add topology-aware filter and scorer for P/D routing

### LINKED ISSUE #545: Topology Aware Routing to lower KV Cache transfer latency

elevran:
**What type of PR is this?**
/kind feature

Adds the two consumers of the `Topology` attribute (`topology-extractor`, #1678) for disaggregated prefill/decode routing:

- `topology-affinity-filter`: keeps candidate endpoints whose topology is co-located with the peer endpoint at a configured minimum affinity (default `host`). Fails open (returns all candidates unchanged) when no peer topology is available, or when no candidate meets the threshold, since locality is a preference and must never make a request unroutable.
- `topology-affinity-scorer`: grades candidates on a bandwidth-shaped proximity curve (host=1.00, rack=0.20, zone=0.05, region=0.02, none=0.00), reflecting the KV-cache transfer path (NVLink vs NIC/switch hops vs inter-datacenter).

Both compare a candidate against a peer endpoint selected in an earlier scheduling phase. In single-EPP disaggregated deployments, `disagg-profile-handler` selects decode first; before running the `prefill` profile (only when prefill is actually needed), it now publishes the selected decode endpoint as a `peer-endpoint` request attribute. The
new plugins should run in the `prefill` profile and read that attribute. The handler itself stays topology-agnostic: it publishes an `Endpoint`, not a `Topology`, so it gains no dependency on the new packages.

A missing topology value never matches, including empty against empty: an endpoint with no hostname never counts as co-located with a peer that also has no hostname.

Coordinator support (separate prefill/decode EPPs, topology carried over a response/request header) is out of scope for this PR and will follow in a separate PR referencing #2282.

Fixes #545

**Release note**:
```release-note
Add `topology-affinity-filter` and `topology-affinity-scorer` plugins for topology-aware prefill/decode routing, preferring co-located prefill and decode pods to shorten the KV-cache transfer path.
```

**Test plan**:
- [x] `pkg/epp/framework/plugins/scheduling/util/topology/...` - Compare across all levels, empty-never-matches, nil inputs, ParseLevel valid/invalid
- [x] `pkg/epp/framework/plugins/scheduling/filter/topologyaffinity/...` - minAffinity keeps correct set, fail-open on no-match and on absent peer, missing attribute dropped
- [x] `pkg/epp/framework/plugins/scheduling/scorer/topologyaffinity/...` - level curve values, all-zero when peer absent, missing attribute safe
- [x] `pkg/epp/framework/plugins/scheduling/profilehandler/disagg/...` - peer-endpoint attribute published before prefill profile runs, absent when prefill is skipped
- [x] `make test-integration-hermetic PATTERN="TestE2EConfigs_StartupParseSmoke"` - new sample config parses and all referenced plugins instantiate
- [x] `make presubmit`

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
+++ b/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
@@ -331,6 +331,9 @@ func (h *Handler) Pick(ctx context.Context, request *scheduling.InferenceRequest
 	if _, hasPrefillProfile := profiles[h.prefillProfile]; hasPrefillProfile {
 		if _, executed := profileResults[h.prefillProfile]; !executed {
 			if h.pdDecider != nil && h.pdDecider.disaggregate(ctx, request, decodeRes.TargetEndpoints[0]) {
+				// Publish the decode pick so plugins in the prefill profile (e.g.
+				// topology affinity) can compare candidates against it.
+				request.PutAttribute(scheduling.PeerEndpointAttributeKey, decodeRes.TargetEndpoints[0])
 				span.SetAttributes(attribute.String("llm_d.epp.profile_handler.decision", "run_prefill"))
 				return map[string]scheduling.SchedulerProfile{h.prefillProfile: profiles[h.prefillProfile]}
 			}
--- a/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
+++ b/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
@@ -36,6 +36,13 @@ const (
 	defaultEncodeProfile  = "encode"
 )
 
+// PeerEndpointAttributeKey is the request-attribute key under which this
+// handler publishes the endpoint selected in an earlier scheduling phase
+// (the decode pick, before running the prefill profile), for plugins in a
+// later profile to compare against (e.g. topology affinity between a
+// disaggregated prefill and decode pick). The value is an Endpoint.
+const PeerEndpointAttributeKey = "peer-endpoint"
+
 // ── Factory & constructor ────────────────────────────────────────────────────
 
 type disaggProfilesParameters struct {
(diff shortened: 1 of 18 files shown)


pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
```diff
@@ -331,6 +338,9 @@ func (h *Handler) Pick(ctx context.Context, request *scheduling.InferenceRequest
 	if _, hasPrefillProfile := profiles[h.prefillProfile]; hasPrefillProfile {
 		if _, executed := profileResults[h.prefillProfile]; !executed {
 			if h.pdDecider != nil && h.pdDecider.disaggregate(ctx, request, decodeRes.TargetEndpoints[0]) {
+				// Publish the decode pick so plugins in the prefill profile (e.g.
+				// topology affinity) can compare candidates against it.
+				request.PutAttribute(PeerEndpointAttributeKey, decodeRes.TargetEndpoints[0])
```
ahg-g, line 343: I know this is existed before and should be resolved in a followup PR, but picking the prefill last is suboptimal. The priority should be to pick the endpoint with the highest kv-cache match across domains/racks and then constrain the decoder to that rack.

ahg-g (approved):
Very nice! but I think the existing order of running through the phases is suboptimal, we should do prefill first.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
elevran: agree it's something to explore. We can revisit the single disagg profile to try out P+D togther vs the current D first (since if D has sufficient coverage, there's no need for P at all).
Le's track that as a separate issue than topology for the Coordinator case (#2300)

elevran: created #2315

Issue #2143, opened 2026-07-22
[Feature Request] A native DisaggregatedSet support routing


hasB4K:
**What would you like to be added**:
A native DisaggregatedSet support routing.

**Why strict revision routing matters (not just a nice-to-have)**:
* disaggregated inference technically routes fine without any of this. DisaggregatedSet's placement policy runs at a layer below routing, and you could argue routing across revisions during a rolling update is harmless. In production it isn't. When rolling from revision A to B, you run into two failure classes:
  * Revision B not backward-compatible with A — either new-revision pods die outright when they receive a request meant for the old shape, or they fail gracefully and the retry also fails, dropping requests and tanking QoS.
  * Corrupted KVCache on revision A — vLLM produced a bad KV cache on some requests on a bad version that had a bug, and you do not want that cache carried into the newer revision's pods.

**Another approach we also tried**
We also tried using plugins, but you end up with many pitfalls in the config at the end, I think it would be easier that way.

PR #2301, opened 2026-08-06
feat(coordinator): forward response headers across phases [Safe DisaggSet 3/4]

hasB4K:
**What type of PR is this?**
/kind feature
/kind documentation
/kind test

Adds a pipeline-level `forward_response_headers` allowlist. The coordinator captures configured headers from render, encode, and prefill responses, then sends them with later upstream requests.

Configured names are reserved for upstream responses, so client-provided values are excluded until a phase returns a replacement value. Invalid, duplicate, and non-forwardable names are rejected at startup.

When forwarding is configured, the first encode request completes before the remaining encode fan-out. Its selected headers are therefore available to every remaining encode request and to later phases.

The mechanism is phase-agnostic and does not depend on a particular EPP plugin.

Supports #2143.

**Testing**:

- [x] `make presubmit`

```release-note
Coordinator pipelines can forward configured response headers to later phases.
```


elevran (commented):
Clean, well scoped change. The forward_response_headers allowlist, the override-on-decode semantics, and the test coverage (param parsing, override behavior, end to end prefill-to-decode flow) all look correct. CI is green and I did not find any correctness issues in the PR.

Two design questions before merge, not blockers, but worth answering so we know if a follow up is needed or we do it here:

1. **Relaxed prefix based forwarding mode**: today an operator must enumerate exact header names (`x-disagg-revision`, `x-disagg-slice`) in `forward_response_headers`. These are internal EPP/coordinator plugin headers that a coordinator operator has no organic way to know, so this couples the coordinator config to whatever the EPP plugin happens to stamp. Would it make sense to add a relaxed mode that auto-forwards all response headers under a default prefix (e.g. `x-llm-d-*`), alongside the explicit allowlist for strict cases? Note this interacts with an existing property of `ForwardedHeaders()`: any client-supplied header not on the hop-by-hop/internal blocklist already passes through untouched today. So a client can currently set `x-disagg-revision` on the original request and it reaches decode unmodified unless the prefill step happens to override it. A prefix based relaxed mode would need the coordinator to strip client-supplied headers under that prefix on first ingress, otherwise it widens the spoofable surface instead of narrowing it. That is a bigger change than this PR (ingress-side stripping), so probably belongs in a separate issue/PR, but wanted to flag it so we agree on scope.

2. **Phase scoping**: the storage mechanism (`SetDownstreamHeader` / `downstreamHeaders` on `RequestContext`, consumed by `ForwardedHeaders()`) is already phase-agnostic and used from encode, prefill to decode_proxy, and render. But the forward_response_headers parameter is only exposed on `PrefillStep`. That matches the current use case (prefill to decode) and keeps the PR minimal, but is it deliberately prefill-only for now, or should the parsing/validation move into a shared helper so a future step that also stamps response headers does not need to reimplement it? Do we need different headers for different phases (side note: don't expect an operator to know that - my prefernce is that we make this configuration and least overhead for operators - default to copy by prefix across all stages - and ensure plugins that need communicating use the right header names)

hasB4K:
One design concern remains for E/P/D.

`forward_response_headers` is straightforward between sequential phases, but Encode is a fan-out stage: one user request can contain several images, and we want to encode them in parallel.

When Encode is part of the DisaggregatedSet compatibility boundary, all Encode requests and the later Prefill/Decode requests must use the same revision. Otherwise we could get:

```text
image 1 -> Encode revision A
image 2 -> Encode revision B
Prefill -> revision A
```

That is not safe to assume compatible, especially when Prefill consumes the encoder-transfer descriptors through EC-NIXL.

The implementation in this PR currently sends one Encode request first, captures its revision, and then starts the remaining Encode requests in parallel with that revision. It is correct, but adds a real latency cost:

```text
before: max(all image encode times)
after:  first image encode time + max(remaining image encode times)
```

This can nearly double Encode-stage latency when image costs are similar. There is no extra cost for a single-image request, and this is not a problem when Encode is outside the DisaggregatedSet.

A routing key alone does not fully solve this in active-active EPP. Different EPP replicas can observe different Pod snapshots during a rollout, so the same key could resolve to different weighted revisions. Avoiding that race would require sticky routing or shared strongly consistent state, which feels like heavy machinery.

I considered two ways to make the revision decision before the Encode fan-out:

1. Reuse `conditional-decode`: let the 412 response carry the selected revision, then run all Encode requests in parallel with it.
2. Add a lightweight scheduling request dedicated to selecting a compatible revision.

Reusing `conditional-decode` could work, but it couples revision selection to an optional cache-probe mechanism and gives its 412 response an unrelated scheduling meaning. A lightweight request feels cleaner and more explicit:

```text
Coordinator -> Gateway/EPP: select a compatible revision
EPP -> Coordinator: selected revision or opaque routing token
Coordinator -> Gateway/EPP: parallel Encode requests carrying that decision
Coordinator -> Gateway/EPP: Prefill and Decode carrying the same decision
```

It adds one small scheduling round trip, but it does not wait for image encoding or GPU work. More importantly, it creates one unambiguous decision before the fan-out without shared state or an active-active routing-key race.

The selection logic should remain in EPP. EPP owns the Pod snapshot, DisaggregatedSet labels, rollout shares, and revision-gating policy. The coordinator may run in another namespace, so it should not need Kubernetes watches, Pod/DisaggregatedSet RBAC, or Kubernetes-specific knowledge. Keeping this in EPP also avoids duplicating revision-selection logic in two components that could later drift apart.

My preference is therefore to keep this PR scoped to generic header forwarding for sequential stages, and follow it with the lightweight EPP scheduling request for E/P/D.

pkg/epp/config/loader/configloader.go
```diff
@@ -229,6 +232,23 @@ func InstantiateAndConfigure(
...
 }
 
+func injectCrossReplicaSyncer(handle fwkplugin.Handle, syncer fwkdl.CrossReplicaSyncer, logger logr.Logger) error {
+	for _, candidate := range handle.GetAllPlugins() {
+		consumer, ok := candidate.(fwkdl.CrossReplicaSyncerConsumer)
+		if !ok {
+			continue
+		}
+		if syncer == nil {
+			logger.Info("WARNING: plugin cross-replica coordination is using its single-replica fallback", "plugin", candidate.TypedName())
```
Lucas-Fernandes-Martins: nit but i wonder if we need a log there. IMO we should log once on startup saying plugin is not syncing cross replica if it's eligible for crossReplicaSyncing once it's added and never again.

Lucas-Fernandes-Martins: Ah ok, the log is for the consumer to know the plugin is in single-replicas fallback. Makes sense, but since syncer is shared by all plugins, I think this should maybe be just one log, not one per plugin.

hasB4K: Thanks for the review! I changed the log (it's not a Warning per se), but I think this log only happen at startup already

EDIT: okay, you meant only log once, for all plugins. I changed this behavior, thanks.

pkg/coordinator/pipeline/context.go
```diff
@@ -37,28 +38,80 @@ var internalForwardingHeaders = map[string]bool{
...
+		out[key] = value
+	}
 	return out
 }
 
+// CaptureResponseHeaders records configured response headers for subsequent
+// pipeline steps. When a step has multiple responses, each response contributes
+// its first value and the most frequent value is recorded. Ties are resolved by
+// the order of the responses. Unconfigured headers are ignored.
+func (rc *RequestContext) CaptureResponseHeaders(responses ...http.Header) {
```
hasB4K, line 77: Why aggregate here? Encode turns one request with many images into one upstream call per image. Each call can land on a different disaggregatedset slice and return a different value. For example, if 30 responses return slice-01 and 20 return slice-02, Prefill should receive slice-01. Using the first or last response would make routing depend on completion timing and could influence sibling calls. Encode therefore waits for the entire fan-out, passes responses here in image order, and this generic code votes independently for every allowlisted header.

For revision, this won't be an issue, since it will be the same value anyway for all images thanks to #2571 and #2572.

pkg/coordinator/pipeline/context.go
```diff
@@ -37,28 +37,74 @@ var internalForwardingHeaders = map[string]bool{
...
```
elevran, line 77: Now that the mutex is gone, this function is only safe when called from a single goroutine. Today that holds at every call site (encode.go after g.Wait, prefill.go and render.go on the main step goroutine), but nothing in the type enforces it. A future step that fans out and calls CaptureResponseHeaders inside the fan-out would silently race, and `go test` without -race would not catch it. Worth a one-line note on the function: "must not be called concurrently."

pkg/coordinator/pipeline/context.go
```diff
@@ -37,28 +37,74 @@ var internalForwardingHeaders = map[string]bool{
...
+	for name := range rc.forwardResponseHeaders {
+		counts := make(map[string]int)
+		order := make([]string, 0)
+		for _, headers := range responses {
+			values := headers.Values(name)
+			if len(values) == 0 {
```
elevran, line 83: `values[0]` can be "" if a worker returns the configured header with an empty value. `""` counts and can win the majority count if more workers return it than any non-empty value. Filter empty strings here, or document that workers should not return empty configured headers, to make the behavior explicit.

pkg/coordinator/steps/encode.go
```diff
@@ -171,11 +128,65 @@ func (s *EncodeStep) Execute(ctx context.Context, reqCtx *pipeline.RequestContex
 	for _, r := range results {
 		s.ec.MergeEncodeResponse(ctx, reqCtx, r)
 	}
+	reqCtx.CaptureResponseHeaders(responseHeaders...)
```
elevran, line 133: On the `g.Wait` error path, `responseHeaders` from successful siblings is intentionally discarded (no CaptureResponseHeaders call). Worth a comment here so a future reader does not add a partial-aggregation branch that leaks some siblings' headers across a failed step.

elevran (commented):
Re-review since 4b08309d, covering 2aba7373, c0633d3a, b5713810. `go test -race` on pipeline/steps/builder/config: clean. Three comments inline.

hasB4K:
Thanks @elevran! I addressed your comment in https://github.com/llm-d/llm-d-router/pull/2301/changes/de50f106730940fd6f121edce19a59bfdba6c832.

Regarding my comment on EPD https://github.com/llm-d/llm-d-router/pull/2301#issuecomment-5207653711 - just to let you know that it's now solved at the EPP level thanks to #2570 and #2571. The PR #2572 is also a follow up PR of this one (to properly handle the EPD case in the coordinator since we can't always rely on request-id)

elevran (commented):
Well-scoped and well-tested. The concurrency contract for header aggregation
after encode fan-out (aggregate only once all parallel requests complete) is
documented and honored by every call site; the dedicated race test confirms
sibling encode requests never see each other's captured headers mid-fan-out.
No correctness or security issues found. All CI is green.

One thing worth raising before merge, or before PR 4/4: pipeline validation
now has two separate entry points. builder.validatePipeline is the existing
single place that validates PipelineConfig at startup; this PR adds
pipeline.NewWithForwardResponseHeaders as a second, independent validation
path called directly from main.go, which validates header names after
builder.Build has already returned. Functionally fine, but worth confirming
whether this split is intentional or whether header-name validation should
live inside builder.Build instead.

Two optional, non-blocking notes: a configured forward header that no phase
ever captures is dropped silently with no diagnostic trail (matches the
documented contract, so not a bug, but a debug-level log could help); and the
new header-forwarding state on RequestContext is private with accessor
methods while similar existing fields (KVTransferParams, ECTransferParams)
are public, a minor style inconsistency worth being aware of.

hasB4K:
Thanks @elevran ! Following up on the pipeline validation point: the split was not necessary, so I consolidated the construction path in https://github.com/llm-d/llm-d-router/pull/2301/changes/0b5e459a7c5b5bf1e7bbc6e583f30ef238f096fe.

`builder.Build()` now validates the pipeline topology, builds the configured steps, calls `pipeline.NewWithForwardResponseHeaders()` to validate and configure the pipeline-wide header allowlist, and returns the completed `*pipeline.Pipeline`. So... `main.go` now has a single pipeline construction call. and the header validation rules remain owned by the pipeline package.

hasB4K:
Closing as already integrated by #2572.

#2572 was stacked on this branch and was merged first. Its squash commit, `557d1974`, contains the complete response-header forwarding change from this PR, including the follow-up that centralizes pipeline construction in `builder.Build()`.

Because #2572 was squash-merged, the original commits from #2301 are not ancestors of `main`, which made GitHub report conflicts even though the functionality is already present.

ahg-g:
oh, was it?! sorry I didn't notice that, my bad!

hasB4K:
> oh, was it?! sorry I didn't notice that, my bad!

No worries, I think we are good anyway! I'll do a follow up PR if needed.

PR #2033, opened 2026-07-15
test: Parallelize the End to End tests - Part two enablement

shmuelk:
**What type of PR is this?**
/kind test

**What this PR does / why we need it**:
Currently the llm-d-router End to End tests run for quite a long time (about sixteen minutes on my Mac M4 Max).

This long time has affected the CI pipeline in the past. A work around was done to run the tests in parts in parallel in separate Kind clusters. This splitting up of the tests needs to be managed manually and has overhead as prepetory steps are done over and over again for each test "slice'

A better solution is to run tests in parallel on the same Kind cluster using K8S Namespaces for separation as needed.

This PR enables the End to End tests to run in parallel. The tests are broken up into groups, mostly by Ginkgo When blocks. By default five such groups run in parallel, all on the same Kind cluster, each using its own namespace with an envoy instance and its own nodePort.

On my laptop a Mac M4 (Max) this PR has reduced the time to run the tests from about sixteen minutes, to between five and six minutes, a significant improvement.

A future PR will update CI to exploit this PR and remove the "test labels"

**Which issue(s) this PR fixes**:
Refs #1931

**Release note** _(write `NONE` if no user-facing change)_:
```release-note
NONE
```

PR #2302, opened 2026-08-06
test(coordinator): parallelize the e2e suite

### LINKED ISSUE #2932: test(coordinator): run the coordinator e2e suite in parallel

  Run the coordinator e2e suite (`test/coordinator/e2e/coordinator`) in parallel Ginkgo processes. Use the same mechanism the router e2e suite uses (#1931, PR #2033): `ginkgo run
  --procs=N` over one Kind cluster, with a namespace and a NodePort for each process.

  - Add `E2E_NUM_PROCS` to `Makefile.coord.mk` (default `5`). `E2E_NUM_PROCS=1` runs the suite serially.
  - Run the suite with `ginkgo --procs` through the shared `run_ginkgo_suite` in `test/scripts/e2e-common.sh`, not `go test`.
  - Put each spec in its own `Ordered` group. Each group creates and deletes its own namespace (`<base>-N`, base `e2e-coordinator`), Envoy, and Services/RBAC.
  - Give each process its own gateway NodePort, so the Envoy Services do not collide on one cluster.
  - Process 1 creates the resources that all processes share: the kind cluster, the CRDs, and one vllm-render deployment in the base namespace. The coordinators reach vllm-render across
  namespaces.
  - Move the namespace create and delete helpers that both suites use into `test/utils`.
  - Document `E2E_NUM_PROCS` and the namespace layout in `README.coord.md`.

  **Why is this needed**:

  The coordinator e2e suite runs its specs one at a time (`go test -timeout 120m ... -ginkgo.fail-fast`). Each spec deploys its own encode, prefill, and decode workers and a coordinator,
  and waits for them to be ready. The suite run time is the sum of all spec times.

  The router e2e suite runs in parallel since #1931 / PR #2033, with `E2E_NUM_PROCS` (default `5`). The coordinator suite has part of the same setup (`numProcesses`,
  `testutils.NamespaceForProcess`, `testutils.RequireParallelProcessesMatch`), but it does not run in parallel.

  With parallel runs, the suite takes less time locally and in the `ci-coordinator.yaml` e2e job. It also uses the same runner and conventions as the router suite.

revit13:
**What type of PR is this?**
/kind test

Parallelizes the coordinator e2e suite, applying the same mechanism the router
e2e already uses (#1931 / PR #2033): `ginkgo run --procs=N` over one Kind
cluster, with per-process namespaces and NodePorts.

Measured on a local dev machine (22 cores, 31 GB, WSL2), not CI:
`make -f Makefile.coord.mk test-e2e-coordinator-run` drops from 8m41s to 5m05s
at the default `E2E_NUM_PROCS=5`, a 1.7x speedup. CI runners are smaller, so
expect a different absolute number there

- **Suite lifecycle.** `SynchronizedBeforeSuite` provisions the cluster, CRDs,
  and a single `vllm-render` once on process 1; every other process only builds
  its own client. `ReportAfterSuite` tears down what process 1 created.

- **Per-group infra.** A `testWrapper` mirroring the router's creates the
  namespace, Envoy, and the Services/ServiceAccounts/RBAC each group binds to in
  `BeforeAll`, and removes them in `AfterAll`. Each of the 9 scenarios becomes
  its own `When(..., ginkgo.Ordered, testWrapper(...))` group so Ginkgo can
  distribute them; `Ordered` keeps a group on the process that built its Envoy.

- **Shared renderer.** One `vllm-render` in the base namespace serves every
  process. The pipeline config gains `${RENDER_NAMESPACE}` for the render step
  so it resolves cross-namespace while the gateway stays on `${NAMESPACE}`.

- **Per-process NodePort.** `shared-envoy-resources.yaml` takes
  `${ENVOY_NODE_PORT}` instead of a hardcoded `30080`, matching the router's
  `envoy.yaml`; without it processes 2..N collide on one port.

- **Shared helpers.** `SetupNamespace`/`DeleteNamespace` move to `test/utils`,
  replacing near-identical copies in both suites. `DeleteNamespace` waits
  `ReadyTimeout` rather than `ExistsTimeout`: finalizing a namespace means
  reaping every pod in it, and the EPP's drain alone exceeds 30s.

- **Runner.** `run_ginkgo_suite` takes an optional timeout; the coordinator
  passes 90m. `E2E_NUM_PROCS` (default 5) is forwarded through
  `Makefile.coord.mk` so `ginkgo --procs` and the suite's `numProcesses` agree.

  ### Behavior change: namespace delete wait

  The shared `testutils.DeleteNamespace` waits until the namespace is gone for up to `ReadyTimeout` (default 3m, set with `READY_TIMEOUT`). This changes both suites:

  | Suite | Before | After |
  |---|---|---|
  | Router | Waits up to `ExistsTimeout` (default 30s) | Waits up to `ReadyTimeout` (default 3m) |
  | Coordinator | Does not wait | Waits up to `ReadyTimeout` (default 3m) |

  Kubernetes deletes every pod in a namespace before it removes the namespace. The EPP pod alone takes more than 30s to stop, and parallel runs on one node take longer. A 30s limit can
  fail a delete that is still in progress.

  A longer timeout cannot cause a false failure. A stuck delete fails after 3m, not 30s.

Fixes #2932 

```release-note
NONE
```

### DIFF
--- a/test/coordinator/e2e/coordinator/coordinator_test.go
+++ b/test/coordinator/e2e/coordinator/coordinator_test.go
@@ -49,55 +49,68 @@ var (
 )
 
 var _ = ginkgo.Describe("Coordinator pipeline", func() {
-	ginkgo.It("routes a text only chat completion end-to-end", func() {
-		runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
-			`{"model":%q,"messages":[{"role":"user","content":"hello"}]}`,
-			modelName,
-		)), textOnlySteps, 0, 0, 0)
-	})
+	ginkgo.When("the request is text only", ginkgo.Ordered, testWrapper(func() {
+		ginkgo.It("routes a text only chat completion end-to-end", func() {
+			runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
+				`{"model":%q,"messages":[{"role":"user","content":"hello"}]}`,
+				modelName,
+			)), textOnlySteps, 0, 0, 0)
+		})
+	}))
 
-	ginkgo.It("forwards the client min_tokens/max_tokens to decode and caps them on prefill and encode", func() {
-		runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
-			`{"model":%q,"messages":[{"role":"user","content":[{"type":"image_url","image_url":{"url":%q},"uuid":"image-0"},{"type":"text","text":"Describe what you see."}]}],"min_tokens":3,"max_tokens":5}`,
-			modelName, inlineImageDataURI,
-		)), allSteps, 1, 3, 5)
-	})
+	ginkgo.When("the request carries token limits", ginkgo.Ordered, testWrapper(func() {
+		ginkgo.It("forwards the client min_tokens/max_tokens to decode and caps them on prefill and encode", func() {
+			runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
+				`{"model":%q,"messages":[{"role":"user","content":[{"type":"image_url","image_url":{"url":%q},"uuid":"image-0"},{"type":"text","text":"Describe what you see."}]}],"min_tokens":3,"max_tokens":5}`,
+				modelName, inlineImageDataURI,
+			)), allSteps, 1, 3, 5)
+		})
+	}))
 
-	ginkgo.It("forwards min_tokens/max_tokens to decode and caps them on prefill and encode with OpenAI passthrough disabled", func() {
-		runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
-			`{"model":%q,"messages":[{"role":"user","content":[{"type":"image_url","image_url":{"url":%q},"uuid":"image-0"},{"type":"text","text":"Describe what you see."}]}],"min_tokens":3,"max_tokens":5}`,
-			modelName, inlineImageDataURI,
-		)), allSteps, 1, 3, 5, coordinatorConfigNIXLGenerate)
-	})
+	ginkgo.When("the request carries token limits and OpenAI passthrough is disabled", ginkgo.Ordered, testWrapper(func() {
+		ginkgo.It("forwards min_tokens/max_tokens to decode and caps them on prefill and encode", func() {
+			runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
+				`{"model":%q,"messages":[{"role":"user","content":[{"type":"image_url","image_url":{"url":%q},"uuid":"image-0"},{"type":"text","text":"Describe what you see."}]}],"min_tokens":3,"max_tokens":5}`,
+				modelName, inlineImageDataURI,
+			)), allSteps, 1, 3, 5, coordinatorConfigNIXLGenerate)
+		})
+	}))
 
-	ginkgo.It("routes a multimodal image chat completion end-to-end", func() {
-		runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
-			`{"model":%q,"messages":[{"role":"user","content":[{"type":"image_url","image_url":{"url":%q},"uuid":"image-0"},{"type":"text","text":"Describe what you see."}]}],"max_tokens":150}`,
-			modelName, testImageURL,
-		)), allSteps, 1, 0, 0)
-	})
+	ginkgo.When("the request carries one remote image", ginkgo.Ordered, testWrapper(func() {
+		ginkgo.It("routes a multimodal image chat completion end-to-end", func() {
+			runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
+				`{"model":%q,"messages":[{"role":"user","content":[{"type":"image_url","image_url":{"url":%q},"uuid":"image-0"},{"type":"text","text":"Describe what you see."}]}],"max_tokens":150}`,
+				modelName, testImageURL,
+			)), allSteps, 1, 0, 0)
+		})
+	}))
 
-	ginkgo.It("routes a multimodal chat completion with two images end-to-end", func() {
-		runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
-			`{"model":%q,"messages":[{"role":"user","content":[{"type":"image_url","image_url":{"url":%q},"uuid":"image-0"},{"type":"image_url","image_url":{"url":%q},"uuid":"image-1"},{"type":"text","text":"What is in these two images?"}]}],"max_tokens":150}`,
-			modelName, testImageURL, testImageURL2,
-		)), allSteps, 2, 0, 0)
-	})
+	ginkgo.When("the request carries two remote images", ginkgo.Ordered, testWrapper(func() {
+		ginkgo.It("routes a multimodal chat completion with two images end-to-end", func() {
+			runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
+				`{"model":%q,"messages":[{"role":"user","content":[{"type":"image_url","image_url":{"url":%q},"uuid":"image-0"},{"type":"image_url","image_url":{"url":%q},"uuid":"image-1"},{"type":"text","text":"What is in these two images?"}]}],"max_tokens":150}`,
+				modelName, testImageURL, testImageURL2,
+			)), allSteps, 2, 0, 0)
+		})
+	}))
 
-	ginkgo.It("routes a multimodal chat completion with an inline base64 image end-to-end", func() {
-		runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
-			`{"model":%q,"messages":[{"role":"user","content":[{"type":"image_url","image_url":{"url":%q},"uuid":"image-0"},{"type":"text","text":"Describe what you see."}]}],"max_tokens":150}`,
-			modelName, inlineImageDataURI,
-		)), allSteps, 1, 0, 0)
-	})
+	ginkgo.When("the request carries one inline base64 image", ginkgo.Ordered, testWrapper(func() {
+		ginkgo.It("routes a multimodal chat completion with an inline base64 image end-to-end", func() {
+			runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
+				`{"model":%q,"messages":[{"role":"user","content":[{"type":"image_url","image_url":{"url":%q},"uuid":"image-0"},{"type":"text","text":"Describe what you see."}]}],"max_tokens":150}`,
+				modelName, inlineImageDataURI,
+			)), allSteps, 1, 0, 0)
+		})
+	}))
 
-	ginkgo.It("routes a multimodal chat completion with one inline and one remote image end-to-end", func() {
-		runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
-			`{"model":%q,"messages":[{"role":"user","content":[{"type":"image_url","image_url":{"url":%q},"uuid":"image-0"},{"type":"image_url","image_url":{"url":%q},"uuid":"image-1"},{"type":"text","text":"Describe what you see in both images."}]}],"max_tokens":150}`,
-			modelName, inlineImageDataURI, testImageURL,
-		)), allSteps, 2, 0, 0)
-	})
-
+	ginkgo.When("the request mixes an inline and a remote image", ginkgo.Ordered, testWrapper(func() {
+		ginkgo.It("routes a multimodal chat completion with one inline and one remote image end-to-end", func() {
+			runCoordinatorPipeline(gateway.PathChatCompletions, []byte(fmt.Sprintf(
+				`{"model":%q,"messages":[{"role":"user","content":[{"type":"image_url","image_url":{"url":%q},"uuid":"image-0"},{"type":"image_url","image_url":{"url":%q},"uuid":"image-1"},{"type":"text","text":"Describe what you see in both images."}]}],"max_tokens":150}`,
+				modelName, inlineImageDataURI, testImageURL,
+			)), allSteps, 2, 0, 0)
+		})
+	}))
 })
 
 // inlineImageDataURI is a 64x64 solid-color PNG encoded as a base64 data URI.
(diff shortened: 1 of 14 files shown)


test/coordinator/e2e/coordinator/coordinator_test.go
```diff
@@ -207,6 +196,21 @@ func runCoordinatorPipeline(path string, body []byte, expectedSteps []string, ex
...
+	}
+
+	// Torn down here rather than via ginkgo.DeferCleanup: Ginkgo runs DeferCleanup
+	// after the group's AfterAll, which has already deleted the namespace by then,
+	// so the objects would be gone. A failing assertion above aborts before these
+	// deletes, leaving the workload in place for testWrapper's AfterAll to dump.
+	testutils.DeleteObjects(testConfig, coordinator, nsName)
+	testutils.DeleteObjects(testConfig, modelServers, nsName)
+	testutils.DeleteObjects(testConfig, epp, nsName)
+	testutils.DeleteObjects(testConfig, pool, nsName)
```
roytman, line 211: Moving these deletes out of `ginkgo.DeferCleanup` means they only run when every assertion above passes. `testWrapper`'s `AfterAll` only backstops this by deleting the whole namespace, and only `if createdNameSpace`. Under `K8S_CONTEXT` against a pre-existing namespace (or the serial `E2E_NUM_PROCS=1` default, where `NAMESPACE` is unset -> `default`), a failing spec now leaks its coordinator/EPP/model-servers/pool permanently.

Worse, `CreateObjsWithVerifier` treats `AlreadyExists` as success, so the next run silently adopts the stale coordinator ConfigMap/Deployment instead of a fresh one - a later, unrelated-looking failure.

Could these deletes move into `testWrapper`'s `AfterAll` (after the failure dump) instead, so cleanup is unconditional again the way `DeferCleanup` guaranteed?

test/coordinator/scripts/run_e2e_coordinator.sh
```diff
@@ -1,21 +1,34 @@
...
+# unconditional and only meaningful when the suite created that cluster
+# (K8S_CONTEXT unset); when running against an existing context it is a no-op
+# unless a cluster of that name happens to exist.
+trap 'e2e_handle_interrupt "e2e-coordinator-tests"' INT TERM
 
 echo "Running coordinator end-to-end tests"
 
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-NAMESPACE="${NAMESPACE:-default}" go test -v -timeout 120m ${DIR}/../e2e/coordinator/ -ginkgo.v -ginkgo.fail-fast
+run_ginkgo_suite "${DIR}/../e2e/coordinator/"
```
roytman, line 34: The PR description says `run_ginkgo_suite` takes an optional timeout and the coordinator passes 90m, but `e2e-common.sh` (unchanged by this PR) still hardcodes `--timeout 45m` — down from the old script's `-timeout 120m`.

Probably fine for the parallel default, but the documented serial escape hatch (`E2E_NUM_PROCS=1`) runs all 9 groups sequentially - cluster/CRD setup, 4 image loads, then 9x Envoy+EPP+vLLM+coordinator provisioning - and 45m seems tight for that. Is the optional-timeout parameter meant to land in this PR, or has 45m been confirmed sufficient for the serial path?

test/coordinator/e2e/coordinator/setup_test.go
```diff
@@ -35,57 +38,102 @@ import (
...
+				// A failing spec aborts before its inline teardown, so its workload
+				// is still up; dump it before the deletes below remove it.
+				testutils.DumpPodsAndLogs(testConfig, nsName, testutils.WithFullLogs())
+				if keepClusterOnFailure {
+					return
+				}
+			}
+			testutils.DeleteObjects(testConfig, stableInfraObjects, nsName)
+			if portForwardSession != nil {
+				portForwardSession.Terminate()
```
roytman, line 121: `portForwardSession.Terminate()` doesn't wait for the process to actually exit. Since `ginkgo.Ordered` groups run sequentially on the same process and reuse `getGatewayPort()`, a slow-to-release bind could make the next group's `port-forward` silently fail to bind, surfacing only as a 10m `waitForCoordinatorReady` timeout rather than a clear error. Worth a `gomega.Eventually(portForwardSession).Should(gexec.Exit())` before returning.

roytman (commented):
A general comment:
`test/scripts/e2e-common.sh`'s `run_ginkgo_suite` uses `${E2E_NUM_PROCS}` without a fallback, and the script runs under `set -euo pipefail`. Fine via `make` (which sets the default), but invoking the script directly will hit an `unbound variable`. Suggest `--procs="${E2E_NUM_PROCS:-1}"`.

test/coordinator/scripts/run_e2e_coordinator.sh
revit13: Good point, the description was ahead of the code. I restored the optional timeout argument to `run_ginkgo_suite` (default 45m), and the coordinator script passes `120m`, the same as the
  old script, which leaves room for the serial `E2E_NUM_PROCS=1` path. The router suite keeps 45m.

test/coordinator/scripts/run_e2e_coordinator.sh
```diff
@@ -1,21 +1,37 @@
...
+trap 'e2e_handle_interrupt "e2e-coordinator-tests"' INT TERM
 
 echo "Running coordinator end-to-end tests"
 
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-NAMESPACE="${NAMESPACE:-default}" go test -v -timeout 120m ${DIR}/../e2e/coordinator/ -ginkgo.v -ginkgo.fail-fast
+# Every spec deploys its own encode/prefill/decode workers and coordinator, so
+# each process carries several full workload cycles, and all nine with
+# E2E_NUM_PROCS=1; 120m leaves headroom over the shared 45m default.
+run_ginkgo_suite "${DIR}/../e2e/coordinator/" 120m
```
roytman: `run_ginkgo_suite` passes `120m` here, but `.github/workflows/ci-coordinator.yaml` caps the `e2e-tests` job at `timeout-minutes: 60`. On a slow CI run the job gets hard-killed before ginkgo's own timeout fires, so we lose `DumpPodsAndLogs` and the kind-cluster teardown that `ReportAfterSuite` would otherwise give us. Can we pick a ginkgo timeout that fits under the job cap (e.g. `50m`), or raise `timeout-minutes` past 120m? The two should be chosen together.

Related: `E2E_NUM_PROCS ?= 5` in `Makefile.coord.mk` was benchmarked on a 22-core/31GB box; `ubuntu-latest` is 4 vCPU/16GB, which is part of why 60m may be tight at the default proc count.

test/coordinator/e2e/coordinator/setup_test.go
```diff
@@ -19,73 +19,161 @@ package coordinate2e
...
+const portForwardExitTimeout = 30 * time.Second
+
+// createEnvoy applies the active topology's Envoy routing ConfigMap plus the
+// shared Envoy Deployment and Service in nsName. Against an existing cluster
+// (K8S_CONTEXT set) the kind nodePort mapping is unavailable, so it also
+// forwards the gateway port and returns the session for the caller to terminate.
+// It appends to objects as it goes rather than returning them at the end, so a
+// failure partway through still leaves the already-created ids tracked for
+// teardown; the namespace delete does not cover this when the suite did not
+// create the namespace.
```
roytman: teardown," but `applyManifest`/`CreateObjsFromYaml` only return `objNames` after every doc in the file is created *and* verified (`CreateObjsWithVerifier` blocks on `VerifyObj` inside the loop). `sharedEnvoyManifest` holds both the Envoy Deployment and Service, so a readiness timeout on the Deployment tracks neither for cleanup, exactly the `K8S_CONTEXT` + pre-existing-namespace case the comment says is covered.

`createTracked` (added in this PR) does this correctly. Could `createEnvoy`/`createStableInfra` route through it instead, or should the comment be corrected to not claim partial-failure tracking?

and the matching comment at :403-405 on createStableInfra

Root-cause fix: `make applyManifest` itself tracks partially, instead of tracking around it.

`applyManifest` only returns after `CreateObjsFromYaml` has verified every object in the file (`CreateAndVerifyObjs` → `CreateObjsWithVerifier` returns post-loop). If the 2nd doc in a file (e.g. the Envoy Deployment) fails `DeploymentAvailable`, this call never returns, so nothing from the file — not even the already-created 1st doc — reaches `*objects`. That's the opposite of what the doc comment on :58-61 says.

`createTracked` already solves this: its verifier closure appends each id before calling `VerifyObj`, so a failure partway through still tracks everything created so far. Could `applyManifest` take `objects *[]string` and call `createTracked` internally instead of `CreateObjsFromYaml`? All 3 call sites (here, :80, :418) are `*objects = append(*objects, applyManifest(...)...)` today, so this collapses to one fix instead of three.

```go
// applyManifest reads path, substitutes subs, and creates the resulting objects
// in nsName, appending each id to objects as it is created so a verify failure
// on one object still leaves the earlier ones in the file tracked for teardown.
func applyManifest(nsName, path string, subs map[string]string, objects *[]string) {
	docs := testutils.ReadYaml(path)
	docs = e2eutil.SubstituteMany(docs, subs)
	createTracked(nsName, docs, objects)
}
```
Then each call site drops the `append(*objects, ...)` wrapper:
```go
applyManifest(nsName, manifest, infraSubs, objects)
```

test/coordinator/e2e/coordinator/setup_test.go
```diff
@@ -19,73 +19,161 @@ package coordinate2e
...
+				// Dump the workload before the deletes below remove it.
+				testutils.DumpPodsAndLogs(testConfig, nsName, testutils.WithFullLogs())
+				if keepClusterOnFailure {
+					return
+				}
+			}
+			deleteSpecWorkload(nsName)
+		})
+
+		ginkgo.AfterAll(func() {
```
roytman, line 149: `portForwardSession.Terminate().Wait(portForwardExitTimeout)` asserts via gomega. If the port-forward process doesn't exit within 30s, this fails the `AfterAll` node and every delete below it (`stableInfraObjects`, `envoyObjects`, `DeleteNamespace`) is skipped, leaking the group's namespace/infra for the rest of the suite, and the next group can then hit `AlreadyExists` adoption on it.

Could the wait move to the end of `AfterAll` (the deletes don't need the port released first), or follow `Terminate()` with a `Kill()` after the wait window so teardown proceeds regardless of the port-forward outcome?

roytman (commented):
@revit13 , pls resolve the conflicts 
and several small comments:
The "ids are tracked as created so a partial failure leaks nothing, namespace delete is no backstop" rationale is stated at `createEnvoy` (:58-61), `createStableInfra` (:403-405), `specWorkload` (:154-157), and `createTracked` (:388-390). Could this live only on `createTracked` and the other three just reference it? The same applies to the "Service and ServiceAccount come from createStableInfra" note repeated at :187, :208, :305, :327.

test/coordinator/e2e/coordinator/setup_test.go
revit13: Agreed, thanks. AfterAll now calls a stopPortForward helper: SIGTERM, a wait of up to portForwardExitTimeout with no assertion, then SIGKILL and a second wait of the same length. It never fails the node, so the deletes and DeleteNamespace always run, and the port is still released for the next group in the normal case.

test/coordinator/scripts/run_e2e_coordinator.sh
revit13: Good catch. I lowered the ginkgo timeout to 50m so it fires before the 60m job cap, which keeps the log dump and kind teardown in ReportAfterSuite. For reference, main runs the same nine specs serially under a 45m ginkgo timeout on the same runner.
On E2E_NUM_PROCS: Agreed that the benchmark box is not representative. I'll check this PR's e2e-tests duration on ubuntu-latest at the default of 5 processes and lower it for CI if the run is tight.

roytman (dismissed):

test/coordinator/scripts/run_e2e_coordinator.sh
```diff
@@ -1,4 +1,18 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
+# Copyright 2026 The llm-d Authors.
```
elevran, line 3: Duplicate license header: this hunk inserts a full Apache-2.0 header block, but the file already has the identical block a few lines below (unchanged context in the diff). Also unrelated to "parallelize the e2e suite" - likely a copy-paste artifact from porting the router script. Drop the new block, keep only the shebang change.

test/utils/utils.go
```diff
@@ -240,22 +240,71 @@ func EventuallyExists(testConfig *TestConfig, getResource func() error) {
...
+
+	return true
+}
+
+// DeleteNamespace deletes nsName and waits until the API server reports it gone,
+// so a caller that recreates it does not race the terminating namespace. The
+// wait is ReadyTimeout, not ExistsTimeout: finalizing a namespace means reaping
+// every workload still in it, which under parallel runs sharing one node takes
+// far longer than an object needs to appear in the API server.
+func DeleteNamespace(testConfig *TestConfig, nsName string) {
```
elevran, line 275: nit: 
this consolidation changes `DeleteNamespace`'s wait from `ExistsTimeout` (30s) to `ReadyTimeout` (3m default) **for both suites**, including the router suite which this PR doesn't otherwise touch. The comment here explains why (namespace finalization reaps every pod, and EPP drain alone exceeds 30s), and it's a safe direction (only makes a stuck delete take longer to fail, never a false failure) - but worth calling out explicitly in the PR description since it's a real behavior change bundled into what reads as a pure extraction.

test/coordinator/e2e/coordinator/setup_test.go
```diff
@@ -19,73 +19,175 @@ package coordinate2e
...
-	ns := &corev1.Namespace{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: nsName,
-		},
+	ginkgo.By("Port-forward did not exit after SIGTERM, sending SIGKILL")
+	session.Kill()
+	select {
+	case <-session.Exited:
+	case <-time.After(portForwardExitTimeout):
+		ginkgo.By("Port-forward did not exit after SIGKILL")
```
elevran, line 71: nit: when a port-forward doesn't exit after SIGKILL, this just logs via ginkgo.By and returns. Per the doc comment above, a failed bind on the next group then "surfaces only as a waitForCoordinatorReady timeout" - a real leak here turns into a confusing, unrelated-looking failure in the next group after the full 10m readiness timeout. Consider failing fast here instead (ginkgo.Fail) so the failure is attributed to the group that caused it and next group does spend 10m waiting to fail.

test/coordinator/e2e/coordinator/setup_test.go
```diff
@@ -19,73 +19,175 @@ package coordinate2e
...
+				}
+			}
+			deleteSpecWorkload(nsName)
+		})
+
+		ginkgo.AfterAll(func() {
+			if portForwardSession != nil {
+				stopPortForward(portForwardSession)
+			}
+			if ginkgo.CurrentSpecReport().Failed() && keepClusterOnFailure {
```
elevran, line 153: nit: 
gates teardown/keep-cluster-on-failure on ginkgo.CurrentSpecReport().Failed(), which reflects only the last-run spec in the Ordered group. Correct today since every group wraps exactly one It, but worth a one-line comment noting that assumption so a future group with multiple specs doesn't silently lose failure tracking for earlier specs in the group.

elevran (commented):
Solid, faithful port of the router e2e suite's parallelization pattern (`SynchronizedBeforeSuite`, per-process namespaces/NodePorts, testWrapper-based Ordered groups) to the coordinator suite. All 9 scenarios are preserved 1:1 in the conversion, CI runs the parallel path green, and the shared vllm-render / namespace isolation mechanics check out. Two things worth fixing before merge, plus one process note:

No issue is linked (Fixes # is empty, and there's no closing issue reference). The PR body cites #1931 as the origin of the pattern being ported, but there's no coordinator-specific tracking issue for this work (which is substantial).

test/coordinator/e2e/coordinator/setup_test.go
revit13: I checked this against Ginkgo (v2.28.3, internal/group.go). Without ContinueOnFailure, an Ordered group skips its remaining specs after the first failure and runs AfterAll within that failed spec, so CurrentSpecReport().Failed() is true there even when the group has several specs. The router suite relies on the same behavior with groups of 2 to 4 specs. The real assumption is "no ContinueOnFailure", so I added a comment that says that.

test/utils/utils.go
revit13: Agreed. Added a "Behavior change in DeleteNamespace" section to the PR description that covers both suites (router: 30s to 3m; coordinator: no wait to wait-until-gone). Thanks

PR #2308, opened 2026-08-06
[WIP] feat(epp): serve aggregated /v1/models from router

### LINKED ISSUE #1878: Expose an OpenAI-compatible /v1/models endpoint on the router for InferencePool model discovery

An OpenAI-compatible `GET /v1/models` endpoint on the llm-d Router that returns the
set of models currently served by the endpoints of the InferencePool it fronts,
in the standard OpenAI list format. The model set should be the deduplicated union
of models reported by ready endpoints, and ideally include LoRA adapters as vLLM
does.

The data layer already collects each endpoint's loaded models #1231 .
This feature is a read-only aggregation of state the router already maintains.

**Why is this needed**:

Today there is no way for a client to ask the router "which models can I send you?"
without out-of-band knowledge (Kubernetes API access to InferencePool/HTTPRoute
resources, or scraping individual model server pods).

zdtsw:
**What type of PR is this?**

/kind feature

- `GET /v1/models` previously fell through to a single random backend, so LoRA adapters loaded on  other pods were missing from the response.
 - The EPP now answers `/v1/models` itself, building the list from the datalayer's per-endpoint model data instead of hitting a model server directly. The result is the de-duplicated,  alphabetically sorted models across all pods, with a deterministic winner when the same model ID is reported by multiple pods.
 - Extended `ModelData` with `Object`, `Created` and `OwnedBy` so the response carries the standard OpenAI fields. Vendor-specific fields (e.g. `max_model_len`, `root`, `permission`) are not returned — the endpoint exposes the OpenAI-compatible schema by design.
 - Returns 503 while no endpoint has reported its model list yet (empty pool, or still being scraped after startup/scale-up, since scraping is interval-based), so clients retry instead of seeing a misleading empty list.
 - Added `models-data-source `and `models-data-extractor` plugins to all deploy config YAMLs

Fixes #1878

```release-note
 The router now answers `GET /v1/models` itself, returning the combined, de-duplicated, alphabetically sorted list of models and LoRA adapters across all pods in the pool, instead of forwarding to a single pod (which could omit adapters loaded elsewhere). The response uses the standard OpenAI fields (id, object, created, owned_by, and parent for adapters); vendor-specific fields a single backend previously returned (e.g. max_model_len, root, permission) are no longer included.  While no endpoint has reported its models yet (empty pool, or still being scraped after startup or scale-up), the endpoint returns 503 so clients retry. Requires `models-data-source `and `models-data-extractor `to be configured in dataLayer.sources; deployments without them will always return 503 when hit `/v1/models`.
```


pkg/epp/handlers/models.go
```diff
@@ -0,0 +1,126 @@
...
+	extProcPb "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+	envoyTypePb "github.com/envoyproxy/go-control-plane/envoy/type/v3"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	envoy "github.com/llm-d/llm-d-router/pkg/common/envoy"
+	errcommon "github.com/llm-d/llm-d-router/pkg/common/error"
+	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+	attrmodels "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/attribute/models"
+	extmodels "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/extractor/models"
```
ahg-g: As a design principle, the runtime (the logic that handles the requests and executes the framework and the plugins) should not reference the individual plugins explicitly, they should interact via the framework interfaces only.

zdtsw: thanks!
moved logic from handler to datastore now

pkg/epp/datastore/datastore.go
```diff
@@ -38,6 +39,8 @@ import (
 	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
 	"github.com/llm-d/llm-d-router/pkg/epp/datalayer"
 	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+	attrmodels "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/attribute/models"
+	extmodels "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/extractor/models"
```
elevran, line 43: `datastore` is core code and previously depended only on attribute packages (see director.go + attrprefix). This adds a dependency on `extractor/models` just to get the `ModelResponse{Object, Data}` envelope. How about moving that envelope type into the attribute package so core datastore code doesn't reach into extractor internals?

pkg/epp/datastore/datastore_test.go
```diff
@@ -1626,3 +1629,130 @@ func TestDiscoveryNotifier_WorksAlongsideDirectUpsert(t *testing.T) {
...
+}
+
+func unmarshalModelResponse(t *testing.T, body json.RawMessage) extmodels.ModelResponse {
+	t.Helper()
+	var resp extmodels.ModelResponse
+	require.NoError(t, json.Unmarshal(body, &resp))
+	return resp
+}
+
+func TestAggregateModels(t *testing.T) {
```
elevran, line 1663: nit: no test covers a single endpoint reporting the same model ID twice, or AggregateModels running concurrently with pod add/remove. Both are realistic given the extractor scrapes external data and the datastore is shared across goroutines.

pkg/epp/handlers/models_test.go
```diff
@@ -0,0 +1,228 @@
...
+	tests := []struct {
+		name        string
+		method      string
+		path        string
+		wantHandled bool
+	}{
+		{name: "GET /v1/models is served", method: http.MethodGet, path: "/v1/models", wantHandled: true},
+		{name: "trailing slash is served", method: http.MethodGet, path: "/v1/models/", wantHandled: true},
+		{name: "query string is ignored", method: http.MethodGet, path: "/v1/models?foo=bar", wantHandled: true},
+		{name: "POST is not served", method: http.MethodPost, path: "/v1/models", wantHandled: false},
```
elevran, line 71: nit: the mock has no call counter, so nothing asserts AggregateModels() is skipped for these non-GET/non-/v1/models cases - a bug that called it unconditionally before the path/method check wouldn't be caught.

elevran (commented):
Solid implementation overall - matches issue #1878's ask. Two things worth addressing:

1. `deploy/config/pd-topology-epp-config.yaml` is the only EPP config under `deploy/config/` that was not wired with `models-data-source`/`models-data-extractor`. `/v1/models` will permanently return 503 for any deployment using it.
2. The 18-file deploy-config sweep is broader than issue #1878 asks for. The existing `ensureDataLayer` default-injection pattern (used for `metrics-data-source`) would let new/omitted configs pick this up automatically instead of requiring every YAML to be hand-edited, and would also close point 1 for free. I think it is worth following that pattern instead. @ahg-g WDYT?

A few smaller items called out inline.

pkg/epp/datastore/datastore.go
zdtsw: agreed

zdtsw:
> Solid implementation overall - matches issue #1878's ask. Two things worth addressing:
> 
> 1. `deploy/config/pd-topology-epp-config.yaml` is the only EPP config under `deploy/config/` that was not wired with `models-data-source`/`models-data-extractor`. `/v1/models` will permanently return 503 for any deployment using it.
> 2. The 18-file deploy-config sweep is broader than issue #1878 asks for. The existing `ensureDataLayer` default-injection pattern (used for `metrics-data-source`) would let new/omitted configs pick this up automatically instead of requiring every YAML to be hand-edited, and would also close point 1 for free. I think it is worth following that pattern instead. @ahg-g WDYT?
> 
> A few smaller items called out inline.

good idea!
will it be better to have another/followup PR handle this.
here if I can get `models-data-source`/`models-data-extractor` added into  `deploy/config/pd-topology-epp-config.yaml` 
in the followup, then to add support for model source  into `ensureDataLayer` along with existing metric source  + revert the changes in all deploy yaml files?

pkg/epp/datastore/datastore.go
```diff
@@ -534,3 +538,45 @@ func selectorEqual(a, b labels.Selector) bool {
 	}
 	return a.String() == b.String()
 }
+
+func (ds *datastore) AggregateModels() (json.RawMessage, int) {
```
ahg-g: I am not sure this is the right place for this function. As I suggested in the above comment, this creates a dependency for core code / runtime on a plugin-specific configuration and implementation. 

I think the behavior that this PR is trying to add should be modeled as a plugin that takes a dependency on the producer of the model data. We don't currently have an extension point for such a plugin, but we should discuss one.

zdtsw: think the easier way is to split the work into seveal pieces:
- start with added these missing fileds in to ModelData which can be indepdenctly worked with main branch
- we then can have something for a plugin for the rest part, but something might be good to check : is it preferable to have empty body request go thro director then all existing plugins can see it or a simple call before fallback in `HandleRequestHeaders` is good enough (close to what the current PR does)

ahg-g: > is it preferable to have empty body request go thro director then all existing plugins can see it or a simple call before fallback in HandleRequestHeaders is good enough (close to what the current PR does)

I think the latter makes more sense, but lets have a concrete proposal for the new extension point.

zdtsw: created https://github.com/llm-d/llm-d-router/issues/2500 to track the porposal

elevran:
@zdtsw still pursuing this? If not, please close. Otherwise, ping when no longer in draft

zdtsw:
> @zdtsw still pursuing this? If not, please close. Otherwise, ping when no longer in draft

Thanks, @elevran ! I will revisit this draft shortly
btw if you have time, would you mind taking a look at this issue https://github.com/llm-d/llm-d-router/issues/1878 to assess if this is the right direction? 
originally this PR 2308 is to fix https://github.com/llm-d/llm-d-router/issues/1878 , during code review it seems like intorducing a new plugin is required, thus the scope extended and i did not fully follow it thro in the past weeks.

PR #2309, opened 2026-08-06
Output length aware TokenLoadScorer

### LINKED ISSUE #2305: Explore methods of approximating output token length
A lightweight estimate of output sequence length, fit for common agentic patterns and done inline with EPP code. This provides another tradeoff point of accuracy/effort (e.g., simplest and least accurate using `TokenLoadScorer` estimates load of 1.5 × ISL, vs a likely more accurate prediction via an external sidecar/service).

Produce an estimate as a request attribute that can be consumed by various plugins.

**Why is this needed**:
The EPP has limited awareness of output sequence length (OSL). Scheduling decisions treat a short 50-token tool-call the same as an 8,000-token reasoning chain, causing potential scheduling issues:

1. **Wrong load balancing**. `TokenLoadScorer` estimates load as 1.5 × ISL. In agentic workloads input and output are nearly independent, the [GLM-5.2 production trace](https://llm-d.ai/blog/serving-glm-5-2-agentic-workloads-on-llm-d) shows median input 195K tokens, median output 317 tokens. Thus, the EPP can overload pods running short requests and underloads pods running reasoning chains.
1. **Head-of-line blocking**. A single reasoning chain (p99 = 7,551 tokens) at the front of the Flow Control queue stalls an entire burst of short tool-calls (p50 = 317 tokens) behind it. Sub-agent bursts arrive 51-deep at p90.
1. **No KV-pressure awareness by request type**. Under memory pressure, LONG and SHORT requests are throttled equally, even though admitting a SHORT request is far less risky.

This is a research spike at this stage and depends on finding reasonably accurate and lightweight estimator.

DolevAdas:
**What type of PR is this?**

/kind feature

Add an `outlen-bucket` `RequestHeaderProcessor` plugin (same pattern as`agent-identity`) that classifies each request into an OSL bin (LONG / SHORT / UNKNOWN) from request-time signals (`enable_thinking`, `thinking_budget`, `has_tools`) and publishes it as a request attribute. Signals are read from the already-parsed`ChatTemplateKWArgs` and `Tools`.

`TokenLoadScorer` (`SimpleTokenEstimator.EstimateOutputFromRequest`) reads the attribute via`ReadRequestAttribute[oslbucket.OSLBucket]` and replaces`1.5 × ISL`: LONG → 4,096, UNKNOWN→ 1000,  SHORT → 100
Storing the *bin* (not a token count) keeps the attribute reusable by the follow-on queue-reordering and KV-gating mechanisms.
Add two log lines in `InFlightLoadProducer`: estimated bucket at `PreRequest`, actual `CompletionTokens` at `EndOfStream` for drift tracking.

OSL estimation formula validated across 22,575 samples (5 datasets):
 KIMI-K2.5 (reasoning/math), rStar-Coder (competitive programming with reasoning traces), Salesforce xlam-function-calling-60k (tool-call responses), WildChat-4.8M (general real-user traffic), and Nvidia Nemotron-SFT-ARC-AGI (reasoning + tool-use); the results were strong with 90.1% precision and 79.7% recall.

`Fixes #2305`.

```release-note
The `inflight-load-producer` scheduler plugin now estimates a request's output tokens from an output-length class (LONG/UNKNOWN/SHORT) published by the new opt-in `outlen-bucket` plugin, replacing the fixed input-length ratio; enable `outlen-bucket` in the `plugins` list to use it.
```


pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/token_estimator.go
```diff
@@ -110,3 +134,44 @@ func (e *SimpleTokenEstimator) EstimateOutput(inputTokens int64, maxOutputTokens
...
+// 100, and UNKNOWN (or missing attribute) use 1000 — preserving the ranking
+// invariant SHORT < UNKNOWN < LONG. All values are bounded by the client-requested
+// cap.
+func (e *SimpleTokenEstimator) EstimateOutputFromRequest(request *fwksched.InferenceRequest) int64 {
+	if request == nil || request.Body == nil {
+		return 0
+	}
+	body := request.Body
+
+	bucket, _ := fwksched.ReadRequestAttribute[oslbucket.OSLBucket](request, oslbucket.OSLBucketKey)
```
elevran, line 150: This read depends on the osl-bucket plugin's RequestHeader() having already run for this request. That ordering holds today only because director.go's HandleRequest happens to run all RequestHeaderProcessor hooks before PreRequest, not because of any declared dependency (RequestHeaderProcessor has no Consumes/Produces equivalent, and the request-attribute API is untyped/string-keyed with no ordering check). If osl-bucket isn't configured, or a future refactor reorders those calls, every request silently reads as UNKNOWN with no error or warning.

This matches the existing agentidentity precedent, so I'm not blocking on it, but could you add a short comment here (or in the README) making the dependency on RequestHeader-before-PreRequest ordering explicit? That saves the next person from having to trace director.go to discover it.
Another option is to require Produce/Consume relations.

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/token_estimator.go
```diff
@@ -110,3 +134,44 @@ func (e *SimpleTokenEstimator) EstimateOutput(inputTokens int64, maxOutputTokens
...
+	switch bucket {
+	case oslbucket.OSLBucketLong:
+		est := longOutputEstimateTokens
+		if body.MaxOutputTokens != nil && *body.MaxOutputTokens > 0 && *body.MaxOutputTokens < est {
```
elevran, line 154: nit: the three branches each repeat the same "clamp to MaxOutputTokens if smaller" pattern. Could pull that into a small helper applied after picking the flat bucket value, e.g. `clamp(est, body.MaxOutputTokens)`, to avoid the duplication.

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
```diff
@@ -395,6 +396,15 @@ func (p *InFlightLoadProducer) PreRequest(ctx context.Context, request *fwksched
 
 	inputTokens := p.tokenEstimator.EstimateInput(request)
 
+	if request.Body != nil {
+		bucket, _ := fwksched.ReadRequestAttribute[oslbucket.OSLBucket](request, oslbucket.OSLBucketKey)
```
elevran, line 400: nit: 
Is `oslbucket.OSLBucket` attribute always present? If not, consider skipping the log of a nil bucket.

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
```diff
@@ -446,18 +456,16 @@ func (p *InFlightLoadProducer) estimateRequestTokens(endpoint fwksched.Endpoint,
...
-		var maxOutputTokens *int64
-		if request != nil && request.Body != nil {
-			maxOutputTokens = request.Body.MaxOutputTokens
-		}
-		// Output tokens are based on the full input, not the cached portion.
-		tokens += p.tokenEstimator.EstimateOutput(inputTokens, maxOutputTokens)
+		// Output tokens are independent of the prefix-cache discount that shapes
+		// adjustedInput. EstimateOutputFromRequest derives them from the OSL bucket
+		// published as a request attribute by the osl-bucket plugin.
+		tokens += p.tokenEstimator.EstimateOutputFromRequest(request)
```
elevran, line 462:
```suggestion
		// Adjust output tokens estimate based on the output bucket request attribute.
		tokens += p.tokenEstimator.EstimateOutputFromRequest(request)
```
The comment indicated what was being replaced more than what remains after the change.

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer_test.go
```diff
@@ -118,15 +118,15 @@ func TestInFlightLoadProducer_Produce(t *testing.T) {
 	require.False(t, ok)
 
 	// 2. Produce with request -> should put UncachedRequestTokens
-	req := makeTokenRequest("req1", 4) // 4 input tokens -> 10 total (with output)
+	req := makeTokenRequest("req1", 4) // 4 input + 1000 output (UNKNOWN) = 1004 total
```
elevran, line 121: nit:
the use of fixed numbers (like `+ 1000 output (UNKNOWN) = 1004 total`) makes tests relatively fragile (e.g. change the unknown bucket estimate to 750 requires a test change as well). Reconsider reliance on "magic" values in test.
Possible resolution: return estimate as class (e.g., "uknown") and size (e.g., 1000). Assert on the more stable aspects and behavior (e.g., if there are no tool calls or thinking budgets then can't classify request --> "unknown").
Or provide two methods, one to return the class (an enum) and the other to map a class to a token size.

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/token_estimator.go
```diff
@@ -32,12 +33,35 @@ type TokenEstimator interface {
...
 // DefaultOutputRatio is the estimated output-to-input token ratio used when no
 // ratio is configured.
 const DefaultOutputRatio = 1.5
 
+const (
+	// longOutputEstimateTokens is the flat output-token estimate for a LONG
+	// (reasoning) request. It is deliberately a fixed value rather than the
+	// client's thinking_budget: the estimate exists to rank requests by load,
+	// where the LONG-vs-SHORT separation dominates, not to predict exact length.
+	longOutputEstimateTokens int64 = 4096
```
elevran, line 52: Q: any reason to prefer int64 over int?

pkg/epp/framework/plugins/requestcontrol/requestheader/oslbucket/README.md
```diff
@@ -0,0 +1,89 @@
+# OSL Bucket
+
+**Type:** `osl-bucket`
+**Interfaces:** `requestcontrol.RequestHeaderProcessor`
+
+Predicts the **output sequence length (OSL)** bin of a request from request-time
```
elevran, line 6: Q: how common are the terms "output sequence length" and "osl"? 
Using "output length" and "outlen" might be more immediate.

pkg/epp/framework/plugins/requestcontrol/requestheader/oslbucket/README.md
```diff
@@ -0,0 +1,89 @@
...
+| `SHORT` | tool-call / short response | < 500 tokens |
+| `UNKNOWN` | no reliable signal | consumers fall back to their own estimate |
+
+The plugin only classifies and publishes — it does not itself change routing.
+Separating production from consumption means the bin is a reliable shared signal
+for any subsystem, exactly like `agent-identity`.
+
+## How It Works
+
+Classification (validated on 22,575 samples across 5 real-world LLM datasets;
```
elevran, line 30: nit: This does not belong in the readme. Users should not care for these details.

pkg/epp/framework/plugins/requestcontrol/requestheader/oslbucket/README.md
```diff
@@ -0,0 +1,89 @@
...
+see `research-directions/osl-aware-scheduling/README.md`):
```
elevran, line 31: the `research-directions/osl-aware-scheduling/README.md` file is not (and should not be) part of the PR. Remove

pkg/epp/framework/plugins/requestcontrol/requestheader/oslbucket/oslbucket.go
```diff
@@ -0,0 +1,210 @@
...
+
+// OSLBucket is the predicted output-sequence-length category for a request,
+// derived from request-time signals before any tokens are generated.
+type OSLBucket int8
+
+const (
+	// OSLBucketUnknown means no reliable signal was found; consumers use their
+	// own fallback (e.g. the ratio-based token estimate). It is the zero value,
+	// so a missing attribute reads as UNKNOWN.
+	OSLBucketUnknown OSLBucket = iota
```
elevran, line 60: `oslbucket.OSLBucketUnkown` ["stutters"](https://michaelwhatcott.com/go-code-that-stutters/), as do other functions and exported symbols in this package.
From a consumer point of view, the symbol name is always prefixed by the package name so avoid repetition.

pkg/epp/framework/plugins/requestcontrol/requestheader/oslbucket/oslbucket.go
```diff
@@ -0,0 +1,210 @@
...
+//
+// Validated on 22,575 samples across 5 real-world LLM datasets
+// (KIMI-K2.5, rStar-Coder, xlam-function-calling-60k, WildChat-4.8M,
+// Nemotron-SFT-ARC-AGI-v1):
+//
+//   - enable_thinking=true  -> LONG:  90.1% precision, 79.7% recall
+//   - enable_thinking=false AND has_tools=true -> SHORT: 100.0% precision, 56.9% recall
+//
+// ISL is intentionally excluded: no correlation with OSL, adds noise.
+// See research-directions/osl-aware-scheduling/README.md for full analysis.
```
elevran, line 88: not needed.

pkg/epp/framework/plugins/requestcontrol/requestheader/oslbucket/oslbucket.go
```diff
@@ -0,0 +1,210 @@
...
+	if enableThinking != nil && *enableThinking {
+		return OSLBucketLong
+	}
+
+	// Large thinking budget without explicit enable_thinking -> treat as LONG.
+	if thinkingBudget != nil && *thinkingBudget > longBudgetThresholdTokens {
+		return OSLBucketLong
+	}
+
+	// Tools without thinking -> short tool-call JSON (measured p50 = 41 tokens, 100% precision).
```
elevran, line 118: TMI on the length and precision.

pkg/epp/framework/plugins/requestcontrol/requestheader/oslbucket/oslbucket.go
```diff
@@ -0,0 +1,210 @@
...
+// the result as a request attribute.
+func (p *Plugin) RequestHeader(_ context.Context, request *scheduling.InferenceRequest) error {
+	if request == nil || request.Body == nil {
+		return nil
+	}
+	request.PutAttribute(OSLBucketKey, EstimateOSLBucket(request.Body))
+	return nil
+}
+
+// boolPtrFromAny coerces a JSON-decoded value into a *bool. It accepts a native
```
elevran, line 164: Consider replacing the hand rolled boolPtrFromAny and int64PtrFromAny:
1. The two functions differ only in the native type and how they parse a string/float, so factor that out.
2. Drop the float64 case entirely (the map came from json.Unmarshal so you get float64 for numbers. If instead you decode with a json.Decoder and call dec.UseNumber(), all numbers come through as json.Number uniformly, with one less case to handle in every coercion function, and no float64 to int64 precision worries.
3. Just use a library, such as `mitchellh/mapstructure` with `WeaklyTypedInput: true`

elevran (commented):
Nice addition, and the OSL bucket classification is well tested (boundaries, coercion helpers, precision/recall documented). 
Feedback inline.

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/token_estimator.go
DolevAdas: int64 matches the InferenceRequestBody.MaxOutputTokens and the in-flight token counters it feeds (atomic.Int64, via concurrencyTracker.add(delta int64))

DolevAdas:
**Benchmarks:**
Two benchmarks were run on **gpt-oss-120b** to evaluate the llm-d-router OSL estimator introduced in this PR.

**Summary:** The new output-length-aware estimator causes no regressions on a production-like Nemotron workload and provides meaningful latency improvements (p95 E2E: −9%, median TTFT: −26%) when decode KV capacity is the constraining resource, primarily by reducing load imbalance across decode pods.

**Benchmark 1: Real-World Validation (Nemotron-SFT-ARC-AGI):**

The first benchmark replayed a sample of 1,000 requests drawn uniformly at random  across the Nemotron-SFT-ARC-AGI dataset's eight configurations. Both arms were identical except for the router estimator:
- OLD: 1.5 × input_tokens
- NEW: Per-request output bucket (SHORT / UNKNOWN / LONG)

The workload was heavily decode dominated and consisted almost entirely of long OSL requests, the buckets are: 983 LONG, 15 SHORT, and 2 UNKNOWN. The actual input lengths ranged from 179 tokens to 100,265 tokens, with a median (p50) of 7,901 tokens and a p95 of 44,172 tokens.

Results ( P/D disaggregation with 4 decode and 2 prefill on a H200 node,6 GPUs, TP=1, full KV: decode --gpu-memory-utilization=0.90, --max-model-len=131072 , 30 req/s, ≤256 in flight, N=1000, seed 42)

| Metric  | OLD (ratio) | NEW (osl-bucket) | Change |
| --- | --- | --- | --- |
| **Load imbalance across pods** (`kv_cov`; lower = more even) | 0.172 | **0.143** | −17% (NEW more even) |
| **Preemptions** (total / max on any pod) | 0 / 0 | 0 / 0 | none (full-KV control) |
| Avg KV-cache fill across the pool (≈ equal total work) | 49.6% | 49.5% | parity |
| **End-to-end latency, p95 — all requests** | 153,217 ms | 156,981 ms | +2.4% (within noise) |
| **Median time-to-first-token — all requests** | 1,982 ms | 2,027 ms | +2.3% (within noise) |
| Worst-case queue on a single pod (peak waiting requests) | 2 | 3 | +1 (both ≈ 0) |
| Short-request tail — E2E latency p95 (n=15) | 132,062 ms | 85,625 ms | −35% (n=15, noisy) |

Verdict: no harm. NEW places load ~17% more evenly (kv_cov 0.143 vs 0.172), but because KV is not the binding resource at 0.90 the aggregate latencies sit at parity (all-request E2E p95 within 2.4%; median TTFT within ~45 ms). On a natural workload that is 98% LONG with tens-of-thousands-of-token prompts, service time is dominated by prompt processing rather than decode scheduling, leaving little opportunity for output-length-aware balancing to improve latency.

**Benchmark 2: Decode-Bound Stress Test:**
We ran a second benchmark to evaluate the estimator under conditions where decode-load balancing is expected to have a measurable impact. The goal was to evaluate whether the osl-bucket estimator provides better decode-load balancing than the existing ratio-based estimator (1.5 × ISL).

 To do this, we created a mixed output-length workload with the following ratio: 600 Short OSL / 100 Unknown / 300 Long OSL.
Since the Nemotron workload is predominantly composed of long requests, we supplemented the short-request bucket with additional samples from the xLAM Function Calling workload. Furthermore, we only sampled requests with up to 1,024 input tokens to create a light-prefill workload and keep most of the processing load on the decoders. This helps better highlight the performance differences between the OSL estimators.

We ran the same setup as before P/D disaggregation with 4 decode and 2 prefill on a H200 node,6 GPUs, TP=1,30 req/s, ≤256 in flight, N=1000, seed 42 but with gpu-memory-utilization=0.50 to intentionally make KV capacity the constraining resource. This creates a scenario where decode-load placement decisions have a meaningful impact on queueing behavior and latency.

Results:

| Metric  | OLD (ratio) | NEW (osl-bucket) | Change |
| --- | --- | --- | --- |
| **Load imbalance across pods** (`kv_cov`; lower = more even) | 0.147 | **0.074** | **−50%** |
| **Worst-case queue on a single pod** (peak waiting requests) | 71 | **37** | **−48%** |
| Avg KV-cache fill across the pool (≈ equal total work) | 59.0% | 57.2% | ~same |
| **End-to-end latency, p95 — all requests** | 84.3 s | **76.7 s** | **−9%** |
| End-to-end latency, p95 — long-output requests | 88.0 s | 78.4 s | −11% |
| End-to-end latency, p95 — unknown-length requests | 84.0 s | 76.6 s | −9% |
| **Median time-to-first-token — all requests** | 2.33 s | **1.73 s** | **−26%** |
| Median time-to-first-token — long-output requests | 2.05 s | 1.39 s | −32% |
| Short-request tail — time-to-first-token p95 *(small regression)* | 4.45 s | 4.67 s | +5% |


The new osl-bucket estimator distributes decode load much more evenly than the old ratio estimator. At essentially the same average KV-cache occupancy across the pool (59.0% for OLD vs 57.2% for NEW) the osl-bucket estimator spread that work far more evenly across the pods (kv_cov 0.147 → 0.074, −50%) and cuts the peak decode-queue depth nearly in half (wait_max 71 → 37, −48%). This is the mechanism the PR targets: OLD lets long-output requests pile onto a single pod, driving it toward the KV wall while its neighbors sit half-idle, whereas NEW osl-bucket estimator spreads those requests out.
The better balance converts directly into lower latency where it matters most, aggregate end-to-end p95 drops 9% (84.3s → 76.7s), the long-output classes improve most (LONG p95 −11%, UNKNOWN p95 −9%), and median time-to-first-token falls 26% overall (2333ms → 1729ms, and −32% for LONG).The one small downside is on the slowest short requests — their 95th-percentile time-to-first-token rises ~5% — because spreading long requests more evenly means a short request occasionally waits behind a long one; even so, the typical short request is faster and the short-request SLO stays ~98% in both versions.

Across both benchmarks, the new estimator consistently improved decode-load balance. On production-like workloads this translated to neutral performance, while in decode-bound scenarios it produced measurable latency improvements with no increase in overall resource consumption.

DolevAdas:
YAMLs and results for the benchmarks

[benchmark2.zip](https://github.com/user-attachments/files/31032303/benchmark2.zip)
[benchmark1.zip](https://github.com/user-attachments/files/31032304/benchmark1.zip)

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/token_estimator.go
```diff
@@ -93,17 +73,48 @@ func (e *SimpleTokenEstimator) EstimateInput(request *fwksched.InferenceRequest)
...
+	}
+	return e.clampOutput(est, request.Body.MaxOutputTokens)
+}
+
+// clampOutput bounds est by the client-requested cap and the operator cap. The
+// client cap applies only when positive (a zero or unset cap is treated as no
+// cap, consistent with the osl-bucket classifier's handling of max_output_tokens);
+// the operator cap applies whenever non-negative.
+func (e *SimpleTokenEstimator) clampOutput(est int64, clientCap *int64) int64 {
+	if clientCap != nil && *clientCap > 0 && *clientCap < est {
```
elevran, line 117: Previously a client cap of exactly 0 clamped the estimate to 0 (there was a test for it). Now 0 is treated the same as "no cap", matching the osl-bucket classifier's own convention but diverging from InferenceRequestBody.MaxOutputTokens's documented contract elsewhere (0 is a valid explicit cap). Likely low-impact, but worth a one-line comment confirming this is intentional since it's a silent behavior change.

elevran (commented):
Nice, well-tested addition - the osl-bucket classification logic and the scorer wiring both look solid, and this round clearly addressed the naming/magic-number/README feedback from the previous review. 

A couple of things worth closing out before merge, neither in this diff's changed files so noting here instead of inline:

- pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/README.md:31 and :35 still document the `outputRatio` parameter and the old ratio-based formula `(round(inputTokens * outputRatio))` that this PR's `token_estimator.go` removes. Anyone configuring inflight-load-producer from that doc will pass a parameter the factory silently ignores.
- Also: EstimateOutputFromRequest depends on the osl-bucket plugin's RequestHeader() having already run. I traced this through director.go and it does hold as a phase-level guarantee today, but it isn't declared anywhere (no Consumes/Produces), so a future plugin-order change could silently break it. Not blocking, might be worth a follow-up issue.

elevran:
thanks, @DolevAdas - this should be ready to merge. 
Still open: whether "OSL"/"osl-bucket" terminology is clear enough vs. "output length"/"outlen". It's a naming preference, not a code issue. Current code/README still use OSL throughout. WDYT? I don't feel strongly either way but was not familiar with the OSL acronym coming into the review and `output length` would have been clearer.

elevran (dismissed):

elevran:
@DolevAdas can you please add a `release note` sectionby editing the PR description, replacing `NONE` with the user facing change?

elevran (approved):

PR #2313, opened 2026-08-06
Skip janitor reaping for requests bound to a live context

### LINKED ISSUE #2295: Flow Control: 5-minute PluginState staleness reaper destroys in-flight accounting for live requests, collapsing the saturation gate under sustained load
## Summary

When Flow Control is enabled with the `concurrency-detector` saturation gate, the in-flight accounting that drives the dispatch gate is corrupted for any request that produces no response chunk for 5 minutes: the `PluginState` staleness reaper (hardcoded `stalenessThreshold = 5min`) evicts the request's `inflight-load-producer` entry and decrements the in-flight counters while the request still occupies the model server. The gate then sees phantom capacity and dispatches more work, which deepens engine queues, which pushes more requests past the 5-minute mark. The loop is self-accelerating and does not recover when load drops.

We measured this end to end on an 8-replica vLLM pool with an aggregate dispatch limit of 96 requests (`maxConcurrency: 12` x 8 endpoints). A single 10-minute 2x traffic burst was enough to trigger the loop; engine occupancy (running + waiting) went from a healthy steady state of ~28 to ~1000 within 55 minutes of the burst and ~1700 by the end of the run, while the EPP's own in-flight count read 40-60% of reality and decayed ~90/minute without completions. 24% of all requests ended with TTFT above 300 s; a 15-minute near-idle period (0.1 qps) failed to drain the backlog.

A predictable objection is that requests waiting 5 minutes without a token are unrealistic outside synthetic tests. Our data supports the premise and refutes the conclusion: in 51 minutes of healthy-plus-burst operation, 0 of 8545 requests waited that long. The 5-minute waits were then manufactured by the gate's own burst-overshoot behavior within 15 minutes, at nominal load, while the accounting was still accurate (verified by live comparison). The chronological reconstruction below shows each step with data; the code walkthrough that follows maps each step to the responsible code path.

## Environment

- llm-d-router: `main` @ `cc715b1b8ef62028950ddae38dfe75b292f2d58a` (image built 2026-08-03)
- Model pool: Qwen3-Coder-480B-A35B-Instruct-FP8, vLLM, TP=4 on GB200, 8 replicas, 1 EPP replica
- Workload: agentic multi-turn traces (streaming chat completions, contexts up to 262k tokens; each completed turn immediately issues the session's next turn, so turn throughput is ~3/s at 1 qps session arrival), client per-request timeout 1200 s
- Session arrival schedule (t = seconds since benchmark start): 1.0 qps for 0-2700, 2.0 qps for 2700-3300, 1.0 qps for 3300-6000, 0.1 qps for 6000-6900, 1.0 qps for 6900-9600
- EPP config (relevant part):

```yaml
featureGates:
- flowControl
plugins:
- type: concurrency-detector
  parameters:
    maxConcurrency: 12   # per endpoint; 8 endpoints -> 96 aggregate
    concurrencyMode: requests
    headroom: 0.0
# plus inflight-load-producer, token-load-scorer,
# prefix-cache-affinity-filter, precise-prefix-cache-producer, token-producer
flowControl:
  maxRequests: "1k"
  maxBytes: "4Gi"
  defaultRequestTTL: "1200s"
  saturationDetector:
    pluginRef: concurrency-detector
```

Data sources: per-request client records (18699 requests) and per-pod vLLM metrics scraped every ~5 s for the whole run, plus two live EPP metric snapshots taken during the run.

## Measured timeline: how a healthy system reached collapse

Per-phase request statistics (client-side):

| Phase | n | e2e p50/p95/p99 (s) | TTFT p50/p95/p99 (s) | requests with TTFT > 300 s |
|---|---|---|---|---|
| 1 qps steady (0-2700) | 6080 | 5.8 / 46 / 93 | 0.8 / 3.0 / 8.8 | **0** |
| 2 qps burst (2700-3300) | 2465 | 17 / 120 / 243 | 1.3 / 58 / 71 | **0** |
| 1 qps (3300-6000) | 4995 | 266 / 1090 / 1194 | 204 / 1022 / 1185 | 1678 |
| 0.1 qps (6000-6900) | 986 | 338 / 860 / 1095 | 284 / 736 / 930 | 417 |
| 1 qps (6900-9600) | 4173 | 435 / 1065 / 1167 | 377 / 996 / 1112 | 2390 |

Pool occupancy (sum of vLLM running + waiting across the 8 replicas, from scraped engine metrics):

| t (s) | 0-2700 | 3120 | 3300 | 3960 | 4560 | 4920 | 6000 | 6900 | 9600 |
|---|---|---|---|---|---|---|---|---|---|
| running + waiting | 23-32 | 103 | 128 | 310 | 406 | 370 | ~1000 | ~700 | ~1700 |

### Stage 0 (t < 2700): healthy steady state

45 minutes at 1 qps: 23-32 requests in-system, TTFT p99 of 8.8 s, zero requests anywhere near a 5-minute wait. This is the regime a skeptic has in mind when arguing 5-minute waits cannot happen, and the data agrees: they do not happen here.

### Stage 1 (2700-3300): burst overshoots the gate; still zero 5-minute waits

The gate should hold aggregate in-flight at 96. Occupancy crosses 103 by t=3120 and reaches 128 at burst end. The gate is open-loop: its counter increments in `PreRequest`, which fires only after the dispatched request finishes the scheduling pipeline (including a remote `/tokenize` call for prompts up to 262k tokens, hundreds of ms to seconds each), while dispatch decisions are made in microsecond cycles against the not-yet-incremented count. Engine waiting depth at burst end is only ~10 per the pool; all 2465 burst-phase requests still complete with TTFT far below 300 s.

### Stage 2 (3300-4200): the ratchet - occupancy triples at nominal load, with accounting still accurate

This is the step that manufactures the 5-minute condition. After the burst ended and arrivals returned to 1 qps (a rate this pool had sustained for 45 minutes), occupancy climbed 128 -> 310 in 11 minutes. No reaping is involved in this stage, and the accounting is demonstrably accurate when quiescent: a live snapshot at t~3890 read EPP counted in-flight 148 vs 140 actually in the engines (the difference is responses in transit). Still zero requests over 300 s.

**How more than 96 requests get admitted although the gate hard-stops at 96.** The gate bounds the *counted* in-flight at dispatch time, but the true in-flight is:

    true_inflight = counted + in_pipeline

where `in_pipeline` is every request already dispatched whose `PreRequest` increment has not landed yet. The increment fires only after the dispatched request finishes the scheduling pipeline - including `token-producer`'s remote `/tokenize` HTTP call against the serving pods for prompts up to 262k tokens - a latency L of hundreds of milliseconds to seconds per request, and longer when the pods are loaded. The dispatch loop meanwhile re-reads the counter every ~14 us cycle. The gate cannot see `in_pipeline`, and it has no mechanism to recall work it already released.

The ratchet cycle, per completion wave:

1. Turn completions arrive in bursts of 100-200/minute (turns batched together finish together); each decrements the counter immediately at `EndOfStream`. Counted falls from ~148 through 96 and below.
2. The instant counted dips under 96, the gate opens and drains the EPP queue (~211 deep at this point) at microsecond cadence. Every released request is invisible for the next L seconds, so the gate keeps reading a low count and keeps dispatching. The queue does not run dry: each completed turn immediately mints the session's next turn into it.
3. Dispatch stops only when the late increments from the earliest releases push counted back over 96. By then the gate has released roughly (completions in the wave) + (dispatch rate x L) requests - the first part legitimate replacement, the second part pure overshoot, bounded only by queue depth.
4. The in-pipeline increments then drain into the counter; counted settles at 96 + overshoot (which is why the t~3890 snapshot could read an *accurate* 148 with the gate closed). The overshoot is now real running/waiting work in the engines and stays until completed. The next wave repeats from this higher floor.

The invariant actually enforced is therefore not "in-flight <= 96" but "no new dispatch while *counted* >= 96", which permits growth of one dispatch-burst per completion wave: 128 -> 160 -> 310 -> 406. The cleanest specimen is t=3900 -> 3960: ~150 completions in that minute, occupancy 160 -> 310 - about 300 dispatches went out in wave-open windows, ~150 replacing completions plus ~150 net overshoot, all before a single accounting entry was reaped.

### Stage 2b (3300-4560): prefix-cache affinity concentrates the backlog onto one pod

The Stage 2 excess is not spread evenly. Per-pod waiting depth (running+waiting per pod, from engine metrics):

| t (s) | per-pod occupancy (run+wait) | max single-pod waiting |
|---|---|---|
| 3360 | 16+2, 19+5, 21+3, 13+3, 10+4, 12+4, 23+4, 14+2 | 5 |
| 3960 | 16+11, **19+64**, 13+16, 22+13, 13+41, 18+6, 13+18, 14+13 | 64 |
| 4560 | 14+20, **16+159**, 20+10, 12+12, 17+13, 15+39, 20+12, 16+13 | 159 |
| 4920 | 3+0, **16+124**, 3+0, **16+201**, 1+0, 3+0, 2+0, 1+0 | 201 |

How the hotspot forms: every turn of a session carries the full conversation as its prompt prefix, and the pod that served the session's earlier turns holds those KV blocks. The prefix-cache affinity scheduling stage therefore steers each session's next turn back to its home pod - reusing tens of thousands of cached prefix tokens normally beats a cold prefill elsewhere, and under healthy load this is the correct and profitable decision. During a backlog it self-reinforces: the sessions homed on the backlogged pod keep appending their next turns to its queue (session turn arrival is closed-loop, but new sessions keep arriving at 1 qps and being distributed), and the affinity decision weighs prefill savings against the target pod's load using the same `InFlightLoad` data that (a) lags reality during Stage 2 drain bursts and (b) is outright falsified once Stage 3 reaping begins - a reaped hot pod looks lightly loaded, so both the affinity trade-off and the load scorer keep selecting it. By t=4920 two pods hold 124 and 201 waiting while six pods are nearly empty.

The hotspot matters for the timeline because per-request wait is queue-position x per-slot service time on that pod, not the pool average: a request at position ~60 on a pod draining ~10-15 turns/minute waits ~5 minutes. Pool-level occupancy of ~300 understates what requests on the hot pod experience.

### Stage 3 (t = 4207): first reaps - the loop closes

The first requests to exceed 300 s without a first token were admitted at t=3907-3915 (by a Stage 2 drain burst, into the hot pod's queue) and crossed the 5-minute mark at t=4207-4215: 54 of them inside a single one-minute window, matching the janitor's 1/minute sweep cadence. From that point the reaper deletes their accounting entries, the counters drop without completions, and the gate refills the "freed" capacity. In terms of the Stage 2 ratchet, the reaper changes the trigger of the counter dips: gate-opening windows no longer require genuine completions - janitor sweeps now delete 50-100 entries per minute for requests still stuck in the engines, so the gate keeps opening (with the same lag overshoot on every drain) even as real throughput collapses. The crossing rate and occupancy growth then track each other:

| window (s) | new TTFT>300s crossings | pool occupancy |
|---|---|---|
| 4200-4500 | 65 | ~170-280 |
| 4500-4800 | 110 | ~180-400 |
| 4800-5100 | 309 | 155 -> 537 |
| 5400-6000 | 673 | 540 -> ~1000 |

Simultaneously, running collapses from ~130 to 21-44 while waiting explodes (KV pressure; cumulative preemptions climb from 5 to 16): each reaped slot is refilled with a new dispatch into pods that can no longer make progress. Turn throughput falls from ~200/minute to ~20/minute.

A second live snapshot late in the run (t~8440) quantifies the corruption directly: EPP counted 588 in-flight (saturation 6.125 x 96) versus 1029 actually in the engines, with the counted value decaying ~90/minute with no matching completions - the janitor sweep visible in real time. Cumulative EPP counters at that moment: 13643 enqueued, 13484 dispatched, 159 queued, zero rejections - nothing bypassed the flow-control queue; the gate itself kept reopening.

### Stage 4: absorbing state

Fifteen minutes at 0.1 qps (t=6000-6900, near-zero offered load) drained occupancy only from ~1000 to ~700, because reaped slots keep being refilled from the still-deep EPP queue and surviving sessions keep issuing turns. Resuming 1 qps drove occupancy to ~1700 by run end. Final tally: 4485 of 18699 requests (24%) exceeded 300 s TTFT, 1924 errors/cancellations, final-phase e2e p50 of 435 s against the healthy-phase 5.8 s.

## Mechanism (code walkthrough)

All references at commit `cc715b1b`.

1. `concurrency-detector` does not count requests itself. It consumes `InFlightLoad` produced by `inflight-load-producer` (`pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go`, `Consumes()`; pool capacity is `maxConcurrency x numEndpoints`, `detector.go:146`).
2. `inflight-load-producer` increments per-endpoint atomic counters in its `PreRequest` hook (`pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go:376`) and stores a per-request entry in `PluginState`. The decrement fires when the entry is released: normally at `EndOfStream` via `PluginState.Delete` (`producer.go:446-507`), or via `OnEvicted` if the janitor reaps it (comment at `producer.go:150-152`: "whether triggered by explicit release at end-of-stream or by the janitor's TTL reaper"). The `PreRequest`-after-scheduling placement is what drives Stages 1-2: dispatch decisions read a count that trails the dispatches themselves by the scheduling-pipeline latency.
3. Entry liveness between increment and release is maintained exclusively by `PluginState.Touch(requestID)`, called once per response body chunk (`producer.go`, the `EndOfStream == false` branch). A request that has not produced its first token has never been touched.
4. `PluginState` hardcodes the liveness bound (`pkg/epp/framework/interface/plugin/plugin_state.go:31-37`):

```go
stalenessThreshold = time.Minute * 5
cleanupInterval    = time.Minute
```

5. A request sitting in the model server's waiting queue (or mid-prefill, or a non-streaming request mid-generation) produces zero chunks. Five minutes after `PreRequest`, the janitor evicts its entry; `OnEvicted` decrements the counters; the EPP permanently forgets the request while the engine still holds it. This is Stage 3.
6. The dispatch loop re-reads saturation every cycle and hard-stops at `saturation >= 1.0` (`pkg/epp/flowcontrol/controller/internal/processor.go:378-397`). The logic is correct; its input is what fails. With reaped counters reading low, the gate reopens and drains the EPP queue into the already-backlogged engines, closing the loop.
7. The same `InFlightLoad` data feeds `token-load-scorer` and the prefix-cache affinity cost model, so routing under corruption prefers exactly the most-backlogged endpoints (Stage 2b's second amplifier; observed end state in a follow-up run: 430 waiting on one pod, 0 on others).

## Why the "requests never wait 5 minutes in production" objection does not hold

- It assumes the 5-minute waits are an input. They are an output: the measured run had zero such waits through 51 minutes of healthy and burst operation, and 54 of them fifteen minutes after a routine 10-minute 2x burst ended - produced by the gate's own overshoot ratchet (Stage 2) plus ordinary prefix-affinity queueing (Stage 2b), both at nominal load with accounting verified accurate. Any deployment that ever transiently exceeds drain capacity is one burst away from the same sequence.
- Non-streaming requests never emit intermediate chunks, so `Touch` never fires for them: any non-streaming request whose total latency exceeds 5 minutes (6k-18k output tokens at 20-60 tok/s - routine for reasoning and code-generation workloads) is reaped mid-flight on a perfectly healthy system. Streaming requests are likewise exposed by any single inter-chunk gap over 5 minutes.
- The failure mode inverts incident handling: an engine that stalls for more than 5 minutes has all its requests reaped, so the EPP concludes it is empty and (via the shared `InFlightLoad` data) steers new traffic toward it. Failing open onto a degraded endpoint is the opposite of the posture the project's own `utilization-detector` documentation argues for on missing data.

## Suggested fixes (any one of the first three prevents the collapse)

1. Do not reap accounting entries of requests whose ext_proc stream is still alive. Tie entry lifetime to request-context cancellation / stream termination instead of wall-clock chunk gaps. The reaper then only covers genuinely leaked entries (its apparent purpose).
2. Make `stalenessThreshold` configurable per consumer, and document that it must exceed the maximum expected time-in-system (TTFT tail plus full generation time), not typical time.
3. Periodically reconcile the in-flight counters against scraped engine telemetry (`num_requests_running + num_requests_waiting` are already collected by the data layer); log and correct divergence. In our run the scraped metrics tracked reality correctly the entire time - the accurate signal existed in the EPP while the gate consumed the corrupted one.
4. Independently: dispatch-time reservation for the counter increment (increment optimistically at dispatch, reconcile after scheduling). This closes the Stage 1-2 overshoot ratchet, which is what builds the backlog that feeds the reaper; it is worth fixing even with 1-3 in place, since it tripled occupancy at nominal load before any reaping occurred.

## Reproduction

Reproduced as described above on an 8x vLLM pool; the shape generalizes to any pool where per-turn service time is seconds-to-minutes and requests share cacheable prefixes:

1. Enable `flowControl` with `concurrency-detector` (`concurrencyMode: requests`, `maxConcurrency = target / numEndpoints`, `headroom: 0`).
2. Run a steady multi-turn streaming workload comfortably within capacity, then apply a 2x rate step for ~10 minutes, then return to the steady rate.
3. Watch `flow_control_pool_saturation x capacity` versus `sum(vllm:num_requests_running + vllm:num_requests_waiting)`. Occupancy ratchets above the gate during and after the step (Stage 2); ~5 minutes after per-pod waiting queues form, the counted value starts decaying in 1/minute steps without completions (Stage 3) and occupancy grows without bound while the dispatched counter keeps rising.

Unit-level demonstration without a cluster: create an `inflight-load-producer` entry via `PreRequest`, send no response chunks for >5 minutes, observe `OnEvicted` decrement while the (simulated) request is still active, then observe the saturation gate admit new work.

LukeAVanDrie:
**What type of PR is this?**

/kind bug

The `PluginState` janitor reaps entries for any request with no read/write/Touch activity for five minutes. The `inflight-load-producer` only touches entries on response chunks, so a request that spends more than five minutes before its first token (deep engine queue, long prefill), a non-streaming request with a long generation, or any single inter-chunk gap over five minutes is reaped while still running on the model server. The eviction rolls back the request's in-flight counters, the `concurrency-detector` saturation gate sees phantom capacity and dispatches more work into the already-backlogged pool, which pushes more requests past the five-minute mark.

#2295 measures this loop end to end: a single 10-minute 2x burst collapsed an 8-replica pool, and the system did not recover at near-zero offered load.

Every dispatched request already releases its accounting when its ext_proc stream ends: normally at `EndOfStream`, and on error/disconnect/panic via the forced `HandleResponseBody(..., endOfStream=true)` in the stream handler's deferred cleanup. The janitor's only legitimate target is entries whose stream teardown never ran.

This PR ties janitor reaping to request liveness instead of chunk cadence:

- `PluginState` gains `BindLiveness(ctx, requestID)`. The janitor skips (and refreshes) stale requests whose bound ctx is still alive; once the ctx dies, they reap on the normal schedule. Unbound requests behave exactly as before (the session-affinity plugin relies on the janitor for GC of never-deleted entries).
- `inflight-load-producer` binds the ext_proc stream context in `PreRequest`, which is canceled exactly when the request's stream terminates.
- `NewPluginState` accepts `WithStalenessThreshold` / `WithCleanupInterval` options (defaults unchanged) so tests can drive the real janitor goroutine.

The regression test reproduces the corruption unit-level: without the fix, a live request's counters roll back to zero after the staleness threshold with no response chunks.

Fixes #2295

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/interface/plugin/plugin_state.go
+++ b/pkg/epp/framework/interface/plugin/plugin_state.go
@@ -176,6 +176,10 @@ func (s *PluginState) Touch(requestID string) {
 // janitor never reaps a bound request whose ctx is still alive; it refreshes
 // the request's last access time instead. Once ctx is done, the entries are
 // reaped on the normal staleness schedule. Delete removes the binding.
+//
+// The caller must have written state for requestID before binding: the janitor
+// reclaims bindings only for request IDs it can reach through the access-time
+// index, so a binding with no state behind it is never reclaimed.
 func (s *PluginState) BindLiveness(ctx context.Context, requestID string) {
 	if requestID == "" || ctx == nil {
 		return
@@ -217,6 +221,7 @@ func (s *PluginState) cleanStaleRequests() {
 		lastAccessTime := v.(time.Time)
 		if time.Since(lastAccessTime) > s.stalenessThreshold {
 			if bound, ok := s.requestLiveness.Load(requestID); ok && bound.(context.Context).Err() == nil {
+				log.Log.V(logutil.TRACE).Info("Holding stale request with a live bound context", "requestID", requestID, "lastAccessTime", lastAccessTime)
 				s.Touch(requestID)
 				return true
 			}
--- a/pkg/epp/framework/interface/plugin/plugin_state.go
+++ b/pkg/epp/framework/interface/plugin/plugin_state.go
@@ -36,9 +36,32 @@ const (
 	cleanupInterval = time.Minute
 )
 
+// PluginStateOption configures a PluginState created by NewPluginState.
+type PluginStateOption func(*PluginState)
+
+// WithStalenessThreshold overrides the default staleness threshold.
+func WithStalenessThreshold(threshold time.Duration) PluginStateOption {
+	return func(s *PluginState) {
+		s.stalenessThreshold = threshold
+	}
+}
+
+// WithCleanupInterval overrides the default cleanup interval.
+func WithCleanupInterval(interval time.Duration) PluginStateOption {
+	return func(s *PluginState) {
+		s.cleanupInterval = interval
+	}
+}
+
 // NewPluginState initializes a new PluginState and returns its pointer.
-func NewPluginState(ctx context.Context) *PluginState {
-	pluginState := &PluginState{}
+func NewPluginState(ctx context.Context, opts ...PluginStateOption) *PluginState {
+	pluginState := &PluginState{
+		stalenessThreshold: stalenessThreshold,
+		cleanupInterval:    cleanupInterval,
+	}
+	for _, opt := range opts {
+		opt(pluginState)
+	}
 	go pluginState.cleanup(ctx)
 	return pluginState
 }
@@ -46,7 +69,8 @@ func NewPluginState(ctx context.Context) *PluginState {
 // PluginState is per-plugin scratch storage scoped to a single request. A plugin's
 // extension points (e.g. PreRequest, ResponseBody) can write, read, and alter entries
 // here to coordinate within that plugin. Entries are keyed by RequestID and reaped
-// after "stalenessThreshold" of inactivity.
+// after "stalenessThreshold" of inactivity, unless the request is bound to a live
+// ctx via BindLiveness.
 //
 // PluginState is not a cross-plugin handoff channel. Data shared between plugins must
 // flow through the Producer/Consumer DAG: write to Endpoint AttributeMap for
@@ -60,6 +84,11 @@ type PluginState struct {
 	storage sync.Map
 	// key: RequestID, value: time.Time
 	requestToLastAccessTime sync.Map
+	// key: RequestID, value: context.Context bound via BindLiveness
+	requestLiveness sync.Map
+
+	stalenessThreshold time.Duration
+	cleanupInterval    time.Duration
 }
 
 // Read retrieves data with the given "key" in the context of "requestID" from PluginState.
@@ -104,6 +133,7 @@ func (s *PluginState) Write(requestID string, key StateKey, val StateData) {
 // removed by a racing DeleteKey (or another Delete) on the same requestID.
 func (s *PluginState) Delete(requestID string) {
 	s.requestToLastAccessTime.Delete(requestID)
+	s.requestLiveness.Delete(requestID)
 	val, ok := s.storage.LoadAndDelete(requestID)
 	if !ok {
 		return
@@ -142,6 +172,17 @@ func (s *PluginState) Touch(requestID string) {
 	s.requestToLastAccessTime.Store(requestID, time.Now())
 }
 
+// BindLiveness marks the request as alive for as long as ctx is not done. The
+// janitor never reaps a bound request whose ctx is still alive; it refreshes
+// the request's last access time instead. Once ctx is done, the entries are
+// reaped on the normal staleness schedule. Delete removes the binding.
+func (s *PluginState) BindLiveness(ctx context.Context, requestID string) {
+	if requestID == "" || ctx == nil {
+		return
+	}
+	s.requestLiveness.Store(requestID, ctx)
+}
+
 // LastAccessTime returns the last access time for the given requestID and a
 // boolean indicating if the requestID was found.
 func (s *PluginState) LastAccessTime(requestID string) (time.Time, bool) {
@@ -153,7 +194,7 @@ func (s *PluginState) LastAccessTime(requestID string) (time.Time, bool) {
 
 // cleanup periodically deletes data associated with the given requestID.
 func (s *PluginState) cleanup(ctx context.Context) {
-	ticker := time.NewTicker(cleanupInterval)
+	ticker := time.NewTicker(s.cleanupInterval)
 	defer ticker.Stop()
 	for {
 		select {
@@ -167,13 +208,18 @@ func (s *PluginState) cleanup(ctx context.Context) {
 }
 
 // cleanStaleRequests iterates through all requests and removes those that haven't been
-// accessed for longer than stalenessThreshold. This operation is safe to run concurrently
+// accessed for longer than stalenessThreshold, except requests whose bound liveness ctx
+// is still alive (see BindLiveness). This operation is safe to run concurrently
 // with other operations on the PluginState.
 func (s *PluginState) cleanStaleRequests() {
 	s.requestToLastAccessTime.Range(func(k, v any) bool {
 		requestID := k.(string)
 		lastAccessTime := v.(time.Time)
-		if time.Since(lastAccessTime) > stalenessThreshold {
+		if time.Since(lastAccessTime) > s.stalenessThreshold {
+			if bound, ok := s.requestLiveness.Load(requestID); ok && bound.(context.Context).Err() == nil {
+				s.Touch(requestID)
+				return true
+			}
 			log.Log.V(logutil.DEBUG).Info("Cleaning up stale request from PluginState", "requestID", requestID, "lastAccessTime", lastAccessTime)
 			s.Delete(requestID) // cleanup stale requests (this is safe in sync.Map)
 		}
--- a/pkg/epp/framework/interface/plugin/plugin_state.go
+++ b/pkg/epp/framework/interface/plugin/plugin_state.go
@@ -28,12 +28,12 @@ import (
 )
 
 const (
-	// stalenessThreshold defines the threshold for considering data as stale.
+	// defaultStalenessThreshold defines the threshold for considering data as stale.
 	// if data of a request hasn't been read/write in the last "stalenessThreshold", it is considered as stale data
 	// and will be cleaned in the next cleanup cycle.
-	stalenessThreshold = time.Minute * 5
-	// cleanupInterval defines the periodic interval that the cleanup go routine uses to check for stale data.
-	cleanupInterval = time.Minute
+	defaultStalenessThreshold = time.Minute * 5
+	// defaultCleanupInterval defines the periodic interval that the cleanup go routine uses to check for stale data.
+	defaultCleanupInterval = time.Minute
 )
 
 // PluginStateOption configures a PluginState created by NewPluginState.
@@ -56,8 +56,8 @@ func WithCleanupInterval(interval time.Duration) PluginStateOption {
 // NewPluginState initializes a new PluginState and returns its pointer.
 func NewPluginState(ctx context.Context, opts ...PluginStateOption) *PluginState {
 	pluginState := &PluginState{
-		stalenessThreshold: stalenessThreshold,
-		cleanupInterval:    cleanupInterval,
+		stalenessThreshold: defaultStalenessThreshold,
+		cleanupInterval:    defaultCleanupInterval,
 	}
 	for _, opt := range opts {
 		opt(pluginState)
--- a/pkg/epp/framework/interface/plugin/plugin_state_test.go
+++ b/pkg/epp/framework/interface/plugin/plugin_state_test.go
@@ -109,7 +109,7 @@ func TestPluginState_EvictionCallback(t *testing.T) {
 	data.evictedID = ""
 	data.evictedKey = ""
 	state.Write(requestID, key, data)
-	state.requestToLastAccessTime.Store(requestID, time.Now().Add(-2*stalenessThreshold))
+	state.requestToLastAccessTime.Store(requestID, time.Now().Add(-2*defaultStalenessThreshold))
 	state.cleanStaleRequests()
 	assert.Equal(t, requestID, data.evictedID)
 	assert.Equal(t, key, data.evictedKey)
@@ -128,7 +128,7 @@ func TestPluginState_Touch(t *testing.T) {
 	state.Write(requestID, key, data)
 
 	// Set last access time to near-stale
-	nearStale := time.Now().Add(-stalenessThreshold + time.Second*10)
+	nearStale := time.Now().Add(-defaultStalenessThreshold + time.Second*10)
 	state.requestToLastAccessTime.Store(requestID, nearStale)
 
 	// Touch it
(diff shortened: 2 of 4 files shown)


LukeAVanDrie:
cc: @yangligt2

LukeAVanDrie:
/assign @ahg-g 

this is ready for review now

pkg/epp/framework/interface/plugin/plugin_state.go
```diff
@@ -36,17 +36,41 @@ const (
 	cleanupInterval = time.Minute
```
ahg-g: lets add a default prefix to the name pls

pkg/epp/framework/interface/plugin/plugin_state.go
```diff
@@ -142,6 +172,21 @@ func (s *PluginState) Touch(requestID string) {
...
 
+// BindLiveness marks the request as alive for as long as ctx is not done. The
+// janitor never reaps a bound request whose ctx is still alive; it refreshes
+// the request's last access time instead. Once ctx is done, the entries are
+// reaped on the normal staleness schedule. Delete removes the binding.
+//
+// The caller must have written state for requestID before binding: the janitor
+// reclaims bindings only for request IDs it can reach through the access-time
+// index, so a binding with no state behind it is never reclaimed.
+func (s *PluginState) BindLiveness(ctx context.Context, requestID string) {
```
ahg-g, line 183: Why do we need a separate interface instead of passing the context as part of the `Write` function call?

pkg/epp/framework/interface/plugin/plugin_state.go
LukeAVanDrie: Done, renamed to defaultStalenessThreshold / defaultCleanupInterval.

pkg/epp/framework/interface/plugin/plugin_state.go
LukeAVanDrie: Mostly because `Write` runs with many different contexts. The producer writes per-chunk from `ResponseBody`, other plugins write from background or scoped contexts, and none of those track the request's lifetime. If `Write` took a ctx, every write would rebind to whatever the caller happened to hold, and a long-lived one would make the entry unreapable. The stream ctx in `PreRequest` is the only one that dies when the request does to my knowledge, so binding it once there seemed safer than making it a side effect of every write.

It also keeps the behavior opt-in with unbound requests reaping as before, which session-affinity relies on. And changing `Write`'s signature touches every plugin for something only one call site needs.

Happy to change it as a follow up refactoring effort if you feel strongly. Adding the new interface also smells, but I couldn't think of a super clean approach here.

LukeAVanDrie: I can cut an issue to at least investigate this more post-merge. If you have any alternative ideas on how to do this cleanly, we can discuss there.

ahg-g (approved):

PR #2319, opened 2026-08-07
Add coordinator support for topology-aware P/D routing

### LINKED ISSUE #2282: [Coordinator] Relay Prefill endpoint metadata to Decode endpoint

### LINKED ISSUE #2300: Topology-aware P/D routing for coordinator (separate prefill/decode EPP) deployments
#545 added `topology-affinity-filter` and `topology-affinity-scorer` for single-EPP disaggregated deployments, where one EPP process runs both the `decode` and `prefill` scheduling profiles and can compare a prefill candidate against the already-selected decode endpoint through an in-process request attribute (`peer-endpoint`, published by `disagg-profile-handler`).

In coordinator deployments, prefill and decode are picked by separate EPP processes, so the peer's topology has to cross the wire instead. This is out of scope for #545 and needs its own PR.

**Scope**

- Wire format: `key=value` comma-separated pairs on a header (e.g. `x-peer-topology: host=node12,rack=r7,zone=us-east1-a,region=us-east1`).
- A prefill-side response stamper plugin (`topology-stamp-handler`) that reads the selected prefill endpoint's `Topology` attribute and sets it on the response header.
- Coordinator plumbing to carry the header from the prefill response to the decode request. See #2282 
    - Header copying (`pkg/coordinator/pipeline/context.go`, `pkg/coordinator/steps/prefill.go`,  `pkg/coordinator/steps/decode_proxy.go`), mirroring the existing `KVTransferParams` channel. 
    - Hardening so a client-supplied `x-peer-topology` header is stripped rather than trusted (`internalForwardingHeaders`, EPP `OutputInjectionHeaders`).
- Introducing `peerTopologyHeader` parameter on `topology-affinity-filter` and `topology-affinity-scorer` once header decoding is real, plus README/config-sample updates for both plugins and a new README for the stamper.

Depends on / relates to #2282 for the general prefill-to-decode metadata relay mechanism.

elevran:
**What type of PR is this?**
/kind feature

Extends `topology-affinity-filter` and `topology-affinity-scorer` (added in #2299 for single-EPP deployments) to support coordinator deployments, where prefill and decode are picked by separate EPP processes and the peer endpoint cannot be resolved from a shared request attribute.

- `topology-affinity-filter`/`topology-affinity-scorer`: resolve the peer topology from a configurable request header (`peerTopologyHeader`) when the peer-endpoint attribute is absent.
- New `topology-stamp-handler` plugin: encodes a scheduling profile's selected endpoint topology onto a response header (`key=value` wire format, e.g. `host=node12,rack=r7,zone=us-east1-a,region=us-east1`), for the prefill EPP to hand off to the coordinator.
- Coordinator: `PrefillStep` captures the header from the prefill response; `newDecodeProxyRequest` forwards it onto the decode request.
- Hardening: the header is denylisted from `ForwardedHeaders()`'s client-header copy and stripped from inbound EPP requests via `OutputInjectionHeaders`, so a client cannot spoof the peer topology.

Fixes #2282 
Fixes #2300

**Release note**:
```release-note
topology-affinity-filter and topology-affinity-scorer now support coordinator deployments
(separate prefill/decode EPPs) via a new topology-stamp-handler plugin and a
peerTopologyHeader parameter that carries the peer endpoint's topology across the wire.
```

**Test plan**:
- [x] `util/topology`: `Encode`/`Decode` round-trip, separator-rejection, unknown-key
  tolerance, malformed-input handling
- [x] `topology-stamp-handler`: header set from the configured/primary profile's endpoint;
  no-op when the profile didn't run, the endpoint lacks the attribute, or
  `SchedulingResult`/response headers are nil
- [x] `PeerTopology` resolver: attribute preferred over header when both present; header
  fallback when attribute absent
- [x] `topology-affinity-filter`/`-scorer`: match/score from `peerTopologyHeader` when no
  peer attribute is present
- [x] Coordinator: `PrefillStep` captures the response header into `RequestContext`;
  `newDecodeProxyRequest` forwards a non-empty value and omits it when empty;
  `ForwardedHeaders()` drops a client-supplied `x-peer-topology`
- [x] `OutputInjectionHeaders` strips a client-supplied `x-peer-topology` on EPP ingress
- [x] `make presubmit` (gofmt, lint, typos, govulncheck, DCO/signed-commits)

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -103,6 +103,7 @@ import (
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/requestattributereporter"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/requestheader/agentidentity"
+	topologystamp "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp"
 	disaggregatedsetrollout "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout"
 	testresponsereceived "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/test/responsereceived"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requesthandling/parsers/anthropic"
@@ -567,6 +568,10 @@ func (r *Runner) registerInTreePlugins() {
 	// request control screeners
 	// Alpha
 	fwkplugin.Register(disaggregatedsetrollout.PluginType, fwkplugin.StabilityAlpha, disaggregatedsetrollout.Factory)
+
+	// request control response-received handlers
+	// Alpha
+	fwkplugin.Register(topologystamp.PluginType, fwkplugin.StabilityAlpha, topologystamp.Factory)
 
 	// bylabel role filters
 	// Beta
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -103,7 +103,7 @@ import (
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/requestattributereporter"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/requestheader/agentidentity"
-	topologystamp "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp"
 	disaggregatedsetrollout "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout"
 	testresponsereceived "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/test/responsereceived"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requesthandling/parsers/anthropic"
--- a/pkg/coordinator/gateway/paths.go
+++ b/pkg/coordinator/gateway/paths.go
@@ -16,7 +16,11 @@ limitations under the License.
 
 package gateway
 
-import "strings"
+import (
+	"strings"
+
+	reqcommon "github.com/llm-d/llm-d-router/pkg/common/request"
+)
 
 const (
 	PathChatCompletions = "/v1/chat/completions"
@@ -24,10 +28,9 @@ const (
 	DefaultGeneratePath = "/inference/v1/generate"
 
 	EPPProfileHeader = "EPP-Profile"
-	// PeerTopologyHeader must match the default headerName in
-	// pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp
-	// and peerTopologyHeaderName in pkg/epp/util/request/headers.go.
-	PeerTopologyHeader = "x-peer-topology"
+	// PeerTopologyHeader is reqcommon.PeerTopologyHeaderKey, the single
+	// definition topology-stamp-handler's default headerName must match.
+	PeerTopologyHeader = reqcommon.PeerTopologyHeaderKey
 	ContentTypeHeader  = "Content-Type"
 	ContentTypeJSON    = "application/json"
 
--- a/pkg/coordinator/gateway/paths.go
+++ b/pkg/coordinator/gateway/paths.go
@@ -23,9 +23,10 @@ const (
 	PathCompletions     = "/v1/completions"
 	DefaultGeneratePath = "/inference/v1/generate"
 
-	EPPProfileHeader  = "EPP-Profile"
-	ContentTypeHeader = "Content-Type"
-	ContentTypeJSON   = "application/json"
+	EPPProfileHeader   = "EPP-Profile"
+	PeerTopologyHeader = "x-peer-topology"
+	ContentTypeHeader  = "Content-Type"
+	ContentTypeJSON    = "application/json"
 
 	PhaseEncode  = "encode"
 	PhasePrefill = "prefill"
--- a/pkg/coordinator/gateway/paths.go
+++ b/pkg/coordinator/gateway/paths.go
@@ -23,7 +23,10 @@ const (
 	PathCompletions     = "/v1/completions"
 	DefaultGeneratePath = "/inference/v1/generate"
 
-	EPPProfileHeader   = "EPP-Profile"
+	EPPProfileHeader = "EPP-Profile"
+	// PeerTopologyHeader must match the default headerName in
+	// pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp
+	// and peerTopologyHeaderName in pkg/epp/util/request/headers.go.
 	PeerTopologyHeader = "x-peer-topology"
 	ContentTypeHeader  = "Content-Type"
 	ContentTypeJSON    = "application/json"
--- a/pkg/coordinator/pipeline/context.go
+++ b/pkg/coordinator/pipeline/context.go
@@ -34,7 +34,8 @@ var hopByHopHeaders = map[string]bool{
 }
 
 var internalForwardingHeaders = map[string]bool{
-	"epp-profile": true,
+	"epp-profile":     true,
+	"x-peer-topology": true,
 }
 
 // ForwardedHeaders returns original request headers suitable for forwarding
@@ -86,6 +87,13 @@ type RequestContext struct {
 	// decode step. Populated by PrefillStep from the prefill response; consumed
 	// by the KV connector when building the decode request.
 	KVTransferParams map[string]any
+	// PeerTopology carries the prefill endpoint's encoded topology from the
+	// prefill response to the decode request, for topology-affinity-filter
+	// and topology-affinity-scorer running in the decode EPP's profile.
+	// Populated by PrefillStep from the x-peer-topology response header;
+	// forwarded onto the decode request by newDecodeProxyRequest. Empty when
+	// the prefill EPP's config has no topology-stamp-handler.
+	PeerTopology string
 
 	// ResponseWriter is used by decode steps to stream the final response to the client.
 	ResponseWriter http.ResponseWriter
--- a/pkg/coordinator/pipeline/context.go
+++ b/pkg/coordinator/pipeline/context.go
@@ -20,6 +20,8 @@ import (
 	"net/http"
 	"strings"
 	"time"
+
+	reqcommon "github.com/llm-d/llm-d-router/pkg/common/request"
 )
 
 var hopByHopHeaders = map[string]bool{
@@ -34,8 +36,19 @@ var hopByHopHeaders = map[string]bool{
 }
 
 var internalForwardingHeaders = map[string]bool{
-	"epp-profile":     true,
-	"x-peer-topology": true,
+	"epp-profile":                   true,
+	reqcommon.PeerTopologyHeaderKey: true,
+}
+
+// StripInternalHeaders removes headers the coordinator alone injects into
+// requests it forwards (e.g. x-peer-topology, copied from the prefill
+// response onto the decode request). Call this on a client request's headers
+// as soon as they are received, so a client-supplied value is never mistaken
+// for one the coordinator generated itself.
+func StripInternalHeaders(headers http.Header) {
+	for key := range internalForwardingHeaders {
+		headers.Del(key)
+	}
 }
 
 // ForwardedHeaders returns original request headers suitable for forwarding
--- a/pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp/stamp.go
+++ b/pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp/stamp.go
@@ -23,7 +23,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
+	reqcommon "github.com/llm-d/llm-d-router/pkg/common/request"
 	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
 	fwkplugin "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
 	fwkrc "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requestcontrol"
@@ -37,10 +39,9 @@ const PluginType = "topology-stamp-handler"
 
 // defaultHeaderName is the response header the encoded topology is written
 // to, and the header topology-affinity-filter/-scorer read the peer topology
-// from in coordinator deployments. Must match gateway.PeerTopologyHeader in
-// pkg/coordinator/gateway and peerTopologyHeaderName in
-// pkg/epp/util/request/headers.go.
-const defaultHeaderName = "x-peer-topology"
+// from in coordinator deployments. Defaults to reqcommon.PeerTopologyHeaderKey,
+// the single definition gateway.PeerTopologyHeader must match.
+const defaultHeaderName = reqcommon.PeerTopologyHeaderKey
 
 type parameters struct {
 	// HeaderName is the response header the encoded topology is written to.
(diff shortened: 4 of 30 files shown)


pkg/coordinator/pipeline/context.go
```diff
@@ -34,7 +34,8 @@ var hopByHopHeaders = map[string]bool{
 }
 
 var internalForwardingHeaders = map[string]bool{
-	"epp-profile": true,
+	"epp-profile":     true,
```
roytman: we have a constant for that
`x-peer-topology` can be another constant

pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp/stamp.go
```diff
@@ -0,0 +1,115 @@
...
+
+// PluginType is the plugin type for the topology stamp handler.
+const PluginType = "topology-stamp-handler"
+
+// defaultHeaderName is the response header the encoded topology is written
+// to, and the header topology-affinity-filter/-scorer read the peer topology
+// from in coordinator deployments. Must match gateway.PeerTopologyHeader in
+// pkg/coordinator/gateway and peerTopologyHeaderName in
+// pkg/epp/util/request/headers.go.
+const defaultHeaderName = "x-peer-topology"
```
roytman: Should it be the same constant, at least for pkg/epp/util/request/headers.go and pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp/stamp.go ?

pkg/epp/util/request/headers.go
```diff
@@ -25,6 +25,15 @@ import (
...
 )
 
+// peerTopologyHeaderName is the coordinator's cross-EPP header carrying the
+// prefill endpoint's topology to the decode EPP. A client-supplied value must
+// not reach the backend, since topology-affinity-filter and
+// topology-affinity-scorer trust it as the peer to compare against. Must
+// match gateway.PeerTopologyHeader in pkg/coordinator/gateway and
+// defaultHeaderName in
+// pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp.
+const peerTopologyHeaderName = "x-peer-topology"
```
roytman: The same comment about one constant.

pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp/stamp.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	result := request.SchedulingResult.ProfileResults[profileName]
+	if result == nil || len(result.TargetEndpoints) == 0 || result.TargetEndpoints[0] == nil {
+		return
+	}
+	topo, ok := fwkdl.ReadAttribute[*attrtopology.Topology](result.TargetEndpoints[0], h.dataKey.String())
+	if !ok {
+		return
+	}
+	response.Headers[h.headerName] = topoutil.Encode(topo)
```
roytman, line 130: The stamped header never leaves the prefill EPP. 
Here, the header is added to the response headers, But that same header was also added to `OutputInjectionHeaders`, and `generateResponseHeaders` (response.go:154) strips anything matching `IsSystemOwnedHeader` before building the Envoy SetHeaders mutation.

pkg/epp/framework/plugins/scheduling/util/topology/peer.go
```diff
@@ -26,18 +26,30 @@ import (
...
-	if !ok || peer == nil {
+	if peer, ok := fwksched.ReadRequestAttribute[fwksched.Endpoint](request, disagg.PeerEndpointAttributeKey); ok && peer != nil {
+		if topo, ok := fwkdl.ReadAttribute[*attrtopology.Topology](peer, dataKey); ok {
+			return topo, true
+		}
+	}
+	if headerName == "" {
+		return nil, false
+	}
+	header := request.Headers[headerName]
```
roytman: Case-sensitive header lookup - should we change the operator-configured header name to lowercase? Like `metadata.GetLowerCaseHeaderValue does.` does.

pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp/stamp.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	if params.HeaderName == "" {
+		params.HeaderName = defaultHeaderName
+	}
+	if name == "" {
+		name = PluginType
+	}
+	return &Handler{
+		typedName:   fwkplugin.TypedName{Type: PluginType, Name: name},
+		headerName:  params.HeaderName,
```
roytman: Case-sensitive header lookup - should we change the operator-configured header name to lowercase? Like `metadata.GetLowerCaseHeaderValue does.` does.

pkg/coordinator/gateway/paths.go
```diff
@@ -23,9 +23,13 @@ const (
 	PathCompletions     = "/v1/completions"
 	DefaultGeneratePath = "/inference/v1/generate"
 
-	EPPProfileHeader  = "EPP-Profile"
-	ContentTypeHeader = "Content-Type"
-	ContentTypeJSON   = "application/json"
+	EPPProfileHeader = "EPP-Profile"
+	// PeerTopologyHeader must match the default headerName in
```
roytman: nit: should be "must match the **PeerTopologyHeader** in ...  (it breaks grep-based rename discovery )

cmd/epp/runner/runner.go
```diff
@@ -101,6 +101,7 @@ import (
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/requestattributereporter"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/requestheader/agentidentity"
+	topologystamp "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp"
```
roytman: Redundant import alias

pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp/stamp.go
```diff
@@ -0,0 +1,115 @@
...
+	// ProfileName selects the scheduling profile whose selected endpoint's
+	// topology is stamped. Empty selects the primary profile's endpoint.
+	ProfileName string `json:"profileName,omitempty"`
+	// TopologyProducerName selects the topology-extractor instance to read
+	// endpoint topology from. Defaults to the extractor's default producer.
+	TopologyProducerName string `json:"topologyProducerName,omitempty"`
+}
+
+var _ fwkrc.ResponseHeaderProcessor = &Handler{}
+
```
roytman, line 62: Should we have?
```go
var _ fwkplugin.ConsumerPlugin = &Handler{}
func (h *Handler) Consumes() fwkplugin.DataDependencies{ 
...
}
```

roytman (commented):
Thank you, Etai, 
Several comments:

- "Stripped from inbound EPP requests via OutputInjectionHeaders" is not accurate - that set is only ever consulted on egress (backend-bound in generateHeaders, client/coordinator-bound in generateResponseHeaders). `director.go` passes the raw inbound header map straight into the scheduling framework with no filtering. The actual trust boundary is operational (don't set peerTopologyHeader on a decode EPP a client can reach directly), which the two plugin READMEs state correctly, but the PR description and test-plan checkbox for this claim do not match the code. Compounding this: the header name is configurable per-plugin, but every strip/denylist site (`headers.go`, `context.go`) hardcodes the literal `x-peer-topology` default, so a renamed header bypasses even the coordinator-side protection silently.

- Missing test: TestGenerateResponseHeaders_Sanitization (pre-existing, untouched by this PR) is exactly where we should validate if the `x-peer-topology` header leaves or not EPP.

- Q. Should we add some DEBUG messages in the no-op/empty branch in `stamp.go`/`decode_proxy.go`?

pkg/coordinator/pipeline/context.go
elevran: Fixed in aae87664d99eaae1e756890d79658fc7ab2bdd07. Added reqcommon.PeerTopologyHeaderKey in pkg/common/request/constants.go and switched context.go to reference it.

pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp/stamp.go
elevran: Fixed in aae87664d99eaae1e756890d79658fc7ab2bdd07. paths.go, stamp.go, and context.go now all reference reqcommon.PeerTopologyHeaderKey.

pkg/epp/util/request/headers.go
elevran: Fixed in aae87664d99eaae1e756890d79658fc7ab2bdd07, but not by pointing peerTopologyHeaderName at the new shared constant - I removed the const entirely.

Digging into why headers.go had it at all: this same OutputInjectionHeaders set is consulted by both generateHeaders (request.go, backend-bound) and generateResponseHeaders (response.go:154, response-to-client/coordinator-bound) via IsSystemOwnedHeader. Having "x-peer-topology" in that shared denylist meant generateResponseHeaders stripped topology-stamp-handler's own response header before it ever reached the coordinator - see my reply on your other comment below, that's the actual bug you flagged there. So the fix for both threads is the same change: delete the entry rather than give it a second home.

pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp/stamp.go
elevran: Fixed in aae87664d99eaae1e756890d79658fc7ab2bdd07. Removed x-peer-topology from OutputInjectionHeaders in headers.go, which is what generateResponseHeaders consulted to strip it. Added a test case to the pre-existing TestGenerateResponseHeaders_Sanitization confirming the header now reaches the coordinator. This was the root-cause functional bug - the feature was non-functional with the default header name before this fix.

pkg/epp/framework/plugins/scheduling/util/topology/peer.go
elevran: Fixed in aae87664d99eaae1e756890d79658fc7ab2bdd07. peer.go lowercases the configured header name before the map lookup, matching the lowercase-key convention used elsewhere.

pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp/stamp.go
elevran: Fixed in aae87664d99eaae1e756890d79658fc7ab2bdd07. stamp.go's Factory now lowercases params.HeaderName once, at construction time.

pkg/coordinator/gateway/paths.go
elevran: Fixed in aae87664d99eaae1e756890d79658fc7ab2bdd07. paths.go now defines PeerTopologyHeader = reqcommon.PeerTopologyHeaderKey, so the comment no longer needs the exact phrase - the identifier itself is grep-discoverable.

cmd/epp/runner/runner.go
elevran: Fixed in aae87664d99eaae1e756890d79658fc7ab2bdd07. Removed the redundant alias in runner.go.

pkg/epp/framework/plugins/requestcontrol/responsereceived/topologystamp/stamp.go
elevran: Fixed in aae87664d99eaae1e756890d79658fc7ab2bdd07. Added var _ fwkplugin.ConsumerPlugin = &Handler{} and Consumes(), declaring the Topology attribute as Optional since ResponseHeader already no-ops when it's absent.

elevran:
Thanks for the thorough review. Responding to each point (all fixes are in aae87664d99eaae1e756890d79658fc7ab2bdd07, also pushed to this branch):

**Trust boundary claim in the PR description** - accepted, and now fixed in code rather than left as documentation. `pkg/coordinator/pipeline/context.go` adds `StripInternalHeaders`, called from `handlers.go` right after the client's request headers are cloned, before they ever land in `OriginalHeaders`. It strips `epp-profile` and `x-peer-topology` from a client's request at the coordinator's ingestion point, using the same `internalForwardingHeaders` denylist `ForwardedHeaders()` already relies on for outbound requests. Worth noting: `reqCtx.PeerTopology` itself was already never populated from a client value (`PrefillStep` sets it only from the prefill EPP's own response header), so the live attack surface was narrower than the PR description implied - but stripping at ingestion closes the gap in code instead of leaving it to an operational assumption. I'll fix the PR description/test-plan checkbox to describe this accurately.

**Compounding concern: hardcoded header name vs. the configurable `headerName` param** - still valid, and NOT addressed by the fix above. `topology-stamp-handler`'s `headerName` param lets an operator rename the header `topology-affinity-filter`/`-scorer` read, but `PrefillStep`, `decode_proxy.go`, `ForwardedHeaders()`, and the new `StripInternalHeaders` all key off the single `reqcommon.PeerTopologyHeaderKey` default. A renamed header silently breaks the coordinator's own capture-and-forward (a functional regression, not just a security gap) and bypasses the ingestion-time strip. Fixing this properly means threading the configured header name through coordinator config rather than assuming a fixed default - a larger, separate change I don't want to fold into this PR's diff. Could you file an issue for it (or I will), so it's tracked rather than silently dropped?

**Missing test** - accepted, fixed. Added a case to the pre-existing `TestGenerateResponseHeaders_Sanitization` asserting `x-peer-topology` reaches the coordinator.

**DEBUG messages in `stamp.go`/`decode_proxy.go` no-op branches** - declined for now: `decode_proxy.go` already surfaces this at DEBUG/TRACE via its existing header-dump logging, and `stamp.go` has no sibling logging convention among the other `requestcontrol/responsereceived` plugins to match. Happy to add one if you still think it's worth it.

pkg/coordinator/gateway/paths.go
```diff
@@ -16,16 +16,23 @@ limitations under the License.
...
 
 const (
 	PathChatCompletions = "/v1/chat/completions"
 	PathCompletions     = "/v1/completions"
 	DefaultGeneratePath = "/inference/v1/generate"
 
-	EPPProfileHeader  = "EPP-Profile"
-	ContentTypeHeader = "Content-Type"
-	ContentTypeJSON   = "application/json"
+	EPPProfileHeader = "EPP-Profile"
```
roytman, line 30: nit: It's out of scope of the PR, but can you move the headers to "github.com/llm-d/llm-d-router/pkg/common/request"

roytman (commented):
LGTM

Three follow-ups worth filing as separate issues rather than blocking this PR: 

- the coordinator hardcodes the peer-topology header name rather than deriving it from headerName/peerTopologyHeader, so a non-default config silently reopens the spoofing gap this PR closes for the default;
- topology-stamp-handler configured on a decode profile leaks the header to the client (have a fix #2365 ). 
- Consumes()'s doc comment promises a startup warning on a missing producer, but TopologyExtractor doesn't implement `ProducerPlugin` (no `Produces()` method), so that warning actually fires unconditionally — it can't currently distinguish a correct config from a typo'd `topologyProducerName`. The same gap exists for topology-affinity-filter/-scorer already, so not something to fix in this PR, but worth its own follow-up issue (adding Produces() to TopologyExtractor is a ~4-line fix that would make all three plugins' checks meaningful).
something like:
```go
// pkg/epp/framework/plugins/datalayer/extractor/topology/extractor.go

var (
	_ fwkplugin.Plugin            = (*TopologyExtractor)(nil)
	_ fwkplugin.ProducerPlugin    = (*TopologyExtractor)(nil)
	_ fwkdl.Registrant            = (*TopologyExtractor)(nil)
	_ fwkdl.EndpointExtractor     = (*endpointHandler)(nil)
	_ fwkdl.NotificationExtractor = (*podNotificationHandler)(nil)
)

// Produces declares the data key this extractor publishes.
func (e *TopologyExtractor) Produces() map[fwkplugin.DataKey]any {
	return map[fwkplugin.DataKey]any{e.dk: attrtopology.Topology{}}
}
```

roytman (dismissed):

elevran:
thanks @roytman - all feedback addressed (fixed or issues created).
Would appreciate another look over.

roytman (dismissed):

elevran:
```go
_ fwkplugin.ProducerPlugin    = (*TopologyExtractor)(nil)
```
was already merged on main from a differnt PR (see L142). Moved into the expected block and pushed again so it's visible.

roytman (dismissed):

ahg-g:
@elevran needs a rebase unfortunately

PR #2325, opened 2026-08-07
feat(coordinator): async-broker step serving OpenAI compatible enqueue, wait, and passthrough modes

BenjaminBraunDev:
**What type of PR is this?**
/kind feature

Adds an optional async-broker pipeline step to the coordinator that gives llm-d-async an OpenAI compatible door through the existing gateway. Clients opt in per request with an `X-AP-Mode` header, and requests without the header pass through the step untouched.

- **enqueue**: writes the request onto the configured llm-d-async queue and answers 202 with an id. The result is collected later with `GET /v1/requests/{id}`, tenant scoped, with `DELETE` for early reclaim.
- **wait**: same enqueue, but the connection is held and completed by a Redis notification when the result lands, returns 504 at deadline.
- **passthrough**: forwards live traffic while classifying it reserved or overflow against the same Redis quota counters the AP's gates use, so one quota account covers a tenant's live and queued traffic.

Queued modes stamp nothing on the message that would re-trigger the step: AP dispatches carry no mode header, so they re-enter the pipeline as ordinary requests and get phased to the EPP like any synchronous call. Objective and fairness headers are always stamped server side, so clients cannot self-assign priority.

The first commit is a small standalone extension: pipeline steps can register auxiliary HTTP routes on the coordinator server, which the broker step uses for fetch, delete, and the models list.

The AP-side prerequisites (result TTLs, per-lane objective stamping, deadline classification) are in llm-d/llm-d-async#394. This step is the coordinator-based alternative to the standalone frontend drafted in llm-d/llm-d-async#392, reusing the coordinator as the single entry point instead of adding a deployable.

Tested with unit tests (config validation, all three modes, quota fail open, wait wake-up and fallback, tenant scoped fetch) and end to end on a GKE cluster with the full stack (Envoy, coordinator, EPP, llm-d-async, vLLM on H200s): all modes served through the gateway, priority labels resolving at the EPP on queued and live paths, deadline aborts surfacing as 504, streaming passthrough, and cross tenant fetch denial. Requests that do not opt in measured about 1.4 microseconds of added step time.

Related: llm-d/llm-d-async#308, llm-d/llm-d-async#394

**Release note**:
```release-note
Add an optional async-broker coordinator step serving OpenAI compatible enqueue, wait, and passthrough modes against llm-d-async queues, with tenant scoped result fetch and delete by request id.
```


jtechapps:
We should document how `streaming: true` requests are handled. Are Streaming requests are only supported during passthrough? and what happens for streamed requests for enqueue/wait are they downgraded to non streamed request automatically or does it cause the request to fail?

pkg/coordinator/steps/asyncbroker/broker.go
```diff
@@ -0,0 +1,384 @@
...
+	default:
+		writeOpenAIError(w, http.StatusBadRequest, "invalid_request_error", api.ErrCodeInvalidRequest,
+			fmt.Sprintf("unknown %s value", s.cfg.ModeHeader))
+		return pipeline.ErrPipelineDone
+	}
+	if reqCtx.Stream && mode != asyncModePassthrough {
+		writeOpenAIError(w, http.StatusBadRequest, "invalid_request_error", api.ErrCodeInvalidRequest,
+			"stream is only supported in passthrough mode")
+		return pipeline.ErrPipelineDone
+	}
```
BenjaminBraunDev, line 157: > We should document how `streaming: true` requests are handled. Are Streaming requests are only supported during passthrough? and what happens for streamed requests for enqueue/wait are they downgraded to non streamed request automatically or does it cause the request to fail?

@jtechapps We fail the request for enqueue/wait, passthrough is supported

pkg/coordinator/server/server_routes_test.go
```diff
@@ -0,0 +1,72 @@
...
+type auxRouteStep struct {
+	stubStep
+	registered bool
+}
+
+func (s *auxRouteStep) RegisterRoutes(r chi.Router) {
+	s.registered = true
+	r.Get("/v1/requests/{id}", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte(chi.URLParam(r, "id")))
```
jtechapps: >G705: XSS via taint analysis (gosec)

can you fix this?

BenjaminBraunDev: done

jtechapps (approved):

docs/coordinator_async_broker.md
```diff
@@ -0,0 +1,139 @@
...
+
+**Wait** returns the model's response on the original connection with the upstream status mirrored, exactly as if the model server had answered directly, and the delivered result is deleted eagerly. Wake-up is a Redis keyspace notification on the result key, with a polling fallback when notifications are unavailable. The hold runs to the request deadline and answers 504 there, or ends early at `wait_cap_seconds` with the 202 response, leaving the request fetchable. If the client disconnects, the step cancels the request pre-dispatch.
+
+**Passthrough** classifies and stamps, then lets the pipeline continue, so streaming and upstream errors behave exactly as they do without the step.
+
+## Endpoints
+
+The step registers three routes on the coordinator listener:
+
+- `GET /v1/requests/{id}` fetches a queued result, tenant scoped and non-destructive
```
roytman, line 60: This will require updating the gateway routing; currently, it forwards all requests except `/v1/completions`, /v1/chat/complitions` and /inference/v1/generate1 to decode nodes. (`/v1/responses` will be added soon)

pkg/coordinator/pipeline/pipeline.go
```diff
@@ -62,6 +62,12 @@ func New(steps []Step) *Pipeline {
 	return &Pipeline{steps: steps}
```
roytman, line 86: if the `async_broker` should be the first step, I'd add a validation check before execution of the pipeline/server.

docs/coordinator_async_broker.md
```diff
@@ -0,0 +1,139 @@
...
+| `X-AP-Mode: wait` | Written to the broker queue, connection held until the result lands | request and response semantics over the queue |
+
+## Request contract
+
+Everything is communicated through headers on a standard OpenAI request, and payloads are not parsed. The step resolves the tenant from a header, classifies the request reserved or overflow against Redis quota counters using the same key scheme as the AP's redis-quota gate (one quota account per tenant across all modes when both sides use the same attribute), and expresses priority as InferenceObjective names the EPP understands. Objective and fairness headers are always stamped server side, so clients cannot self-assign priority.
+
+```
+POST http://gateway:8081/v1/chat/completions
+Content-Type: application/json
+X-Team: premium                    # tenant (quota account, fairness id)
```
roytman, line 23: Tenant identity is entirely client-asserted; no auth boundary exists. `X-Team` is read straight off the request header (`async_broker.go:168`, `async_broker_routes.go:50-55`) and used as the sole isolation key for fetch/delete/quota. 
Any caller reaching the gateway can read/delete another tenant's result or draw on another tenant's quota tier — exactly the priority self-assignment the step otherwise guards against. Needs an explicit deployment invariant ("gateway must overwrite this header") or the security claims in the code/docs need to be walked back.

pkg/coordinator/server/server.go
```diff
@@ -78,6 +78,17 @@ func logRequestResponse(next http.Handler) http.Handler {
 	})
 }
 
+// RouteRegistrar is implemented by pipeline steps that serve auxiliary HTTP
+// endpoints from the coordinator listener, beyond the built-in inference
+// routes (for example, result retrieval for a queueing step). RegisterRoutes
+// is called once per implementing step at server construction, after the
+// built-in routes are registered. Registering a path the server already owns
```
roytman: Duplicate registration silently overwrites; no panic, ever. "safety net" doesn't exist for a future step's route collision.
see https://github.com/go-chi/chi/blob/8b258c7bb28f97a5f2a856ff7ef962578fec9215/tree.go#L352-L380
chi's own test acknowledging this ("duplicate will have no effect"): [tree_test.go#L46](https://github.com/go-chi/chi/blob/8b258c7bb28f97a5f2a856ff7ef962578fec9215/tree_test.go#L46)

pkg/coordinator/steps/asyncbroker/config.go
```diff
@@ -0,0 +1,302 @@
...
+	// Objectives maps tier -> objective names by classification, stamped on
+	// passthrough requests as ObjectiveHeader. Tiers without an entry are not
+	// stamped. Queued dispatches are stamped by the AP's merge policy
+	// (lane_objectives), not here.
+	Objectives      map[string]asyncObjectivePair `json:"objectives,omitempty"`
+	ObjectiveHeader string                        `json:"objective_header,omitempty"`
+	FairnessHeader  string                        `json:"fairness_header,omitempty"`
+}
+
+func (c *asyncBrokerConfig) applyDefaults() {
```
roytman, line 166: Wait-mode connection holds are unbounded by default. No default max_seconds/wait_cap_seconds is applied and a test confirms 1000000 seconds is accepted verbatim. A default deployment lets a client hold a goroutine/socket/Redis subscription for ~11.5 days — also blocks graceful shutdown.

pkg/coordinator/steps/asyncbroker/wait.go
```diff
@@ -0,0 +1,119 @@
...
+			w.mu.Unlock()
+		}
+	}
+}
+
+// register subscribes to the key's notification channel and returns a wake
+// channel plus a cleanup func. Callers must check the key AFTER register
+// returns: a result that landed before the subscription is only found by
+// that check.
+func (w *asyncResultWaiter) register(ctx context.Context, key string) (<-chan struct{}, func(), error) {
```
roytman, line 100: two concurrent wait requests sharing a tenant+id (which the docs explicitly support via retry-by-id) overwrite each other's map entry; whichever finishes first unsubscribes the channel the other is still parked on. Degrades to slow polling silently rather than failing loudly.

pkg/coordinator/steps/asyncbroker/quota.go
```diff
@@ -0,0 +1,94 @@
...
+	key := fmt.Sprintf("%s%s:%s", q.cfg.Prefix, q.cfg.Attribute, tenant)
+	res, err := quotaAcquireScript.Run(ctx, q.rdb, []string{key}, limit, q.cfg.WindowSeconds).Result()
+	if err != nil {
+		return asyncClassificationReserved, nil, fmt.Errorf("quota check failed (failing open): %w", err)
+	}
+	if v, ok := res.(int64); !ok || v == 0 {
+		return asyncClassificationOverflow, nil, nil
+	}
+
+	release := func() {
```
roytman, line 90: Fail-open quota design is undermined by silent failure paths. 
Quota release errors are discarded with no logger even present on the classifier; repeated Redis errors permanently skew a tenant toward overflow with zero diagnostic trail, the opposite of "fail open." Separately, the wait-loop's lookupResult error is checked then ignored twice (async_broker.go:329-330, :364), so a corrupted result or Redis error looks identical to "not ready yet" and burns the full wait window silently.

pkg/coordinator/steps/asyncbroker/broker.go
```diff
@@ -0,0 +1,384 @@
...
+			Deadline: now.Add(timeout).Unix(),
+			Payload:  reqCtx.Body,
+			Metadata: metadata,
+			Headers:  fwd,
+			Endpoint: reqCtx.OriginalPath,
+		},
+		RequestQueueName: queue,
+		ResultQueueName:  resultKey(tenant, reqCtx.RequestID),
+	}
+	if err := s.sub.SubmitRequest(ctx, msg); err != nil {
```
roytman, line 279: SubmitRequest failure and fetch/delete Redis failures (`async_broker_routes.go:79-83, 109-114`) never call the step's own logger (which exists specifically for this), while the raw Redis error string is written straight into the client-facing JSON body - an internal leak with no matching server-side signal.

pkg/coordinator/steps/asyncbroker/routes.go
```diff
@@ -0,0 +1,225 @@
...
+// writeResult maps a stored ResultMessage onto the HTTP response. StatusCode
+// > 0 mirrors the upstream status and body verbatim. Error codes map:
+// GATE_DROPPED -> 429, DEADLINE_EXCEEDED -> 504, INVALID_REQUEST -> 400,
+// CANCELLED -> 499, everything else -> 502, wrapped in the OpenAI error
+// envelope. Returns the body write error so callers that confirm delivery
+// can act on it; the error-envelope branches always return nil.
+func writeResult(w http.ResponseWriter, res *api.ResultMessage) error {
+	if res.StatusCode > 0 {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(res.StatusCode)
```
roytman, line 258: res is deserialized straight from whatever JSON is sitting in the Redis mailbox. Nothing guarantees StatusCode is a valid HTTP status. 
Go's `net/http` panics if you call `WriteHeader` with a code outside 100-999. So a malformed or corrupted stored result (bad producer write, bit flip, a future AP bug) turns a fetch into a panic. `middleware.Recoverer` (registered in server.go) catches it and turns it into a 500, so it's not a crash — but it's an ugly way to get there, and the client gets a generic 500 instead of a clear "malformed stored result" signal. 
A one-line guard (`if res.StatusCode < 100 || res.StatusCode > 599 { ...502... }`) avoids relying on the recovery middleware as a safety net for untrusted data.

pkg/coordinator/steps/asyncbroker/routes.go
```diff
@@ -0,0 +1,225 @@
...
+	asyncStateReady asyncRequestState = iota
+	asyncStatePending
+	asyncStateUnknown
+)
+
+// lookupResult reads a request's result without consuming it. Pending is
+// detected via the producer-maintained active-token key, which exists from
+// submit until the result is flushed. Both reads are tenant-scoped, so a
+// caller with the wrong tenant sees an unknown id.
+func lookupResult(ctx context.Context, rdb *redis.Client, tenant, id string) (asyncRequestState, *api.ResultMessage, error) {
```
roytman, line 173: lookupResult's race can return 410 for a result that just landed
There are two separate, non-atomic Redis reads. Picture the timing: the AP delivers a result and clears the active-token marker between these two reads. 
Read #1 (line 152) happens just before delivery - empty. Read #2 (line 164) happens just after it's cleared - also empty. 
`lookupResult` concludes `asyncStateUnknown`, and `handleFetch` returns `410 Gone` ("unknown, expired, or already deleted") for a request whose result is sitting right there in Redis, one instant too late to see. In wait mode, this is harmless — the polling loop just tries next tick again. On the plain fetch route, a client can get a spurious 410 for a request that actually succeeded. A second mailbox re-read after a marker-miss (before returning Unknown) closes the window.

pkg/coordinator/steps/async_broker_routes.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+	case asyncStatePending:
+		writePending(w, id)
+	default:
+		writeOpenAIError(w, http.StatusGone, "invalid_request_error", "UNKNOWN_REQUEST",
+			"request id is unknown, expired, or already deleted")
+	}
+}
+
+func (s *AsyncBrokerStep) handleDelete(w http.ResponseWriter, r *http.Request) {
```
roytman: That's the entire handler - a Del on the result mailbox key, nothing else. There's no call to `s.sub.CancelRequests` (the method that exists and is used for exactly this purpose in the wait-mode disconnect path, `async_broker.go:353`). So: if a request is still queued/in-flight and a client calls `DELETE /v1/requests/{id}`, the mailbox key is deleted, the client gets `204 No Content`, and the request keeps running server-side — the AP will eventually deliver the result and recreate the same mailbox key, silently undoing the "delete."

The doc (`docs/coordinator_async_broker.md:58`) frames this accurately as "reclaims a result early" — i.e., it's an early garbage-collect of a result you no longer want, not a cancel — so the behavior matches what's documented. The suggestion is purely about the HTTP semantics: `204 No Content` on a `DELETE /v1/requests/{id}` reads to most API consumers as "this request is cancelled/gone," which isn't quite true if it's still queued. Either call `CancelRequests` too (making the name match the effect), or make the doc's "reclaims early" framing land right at the endpoint description so it's unambiguous that `DELETE` only clears the mailbox.

pkg/coordinator/steps/async_broker.go
```diff
@@ -0,0 +1,384 @@
...
+				}
+			}
+			return
+		}
+		select {
+		case <-waitCtx.Done():
+			if ctx.Err() != nil {
+				// Client disconnected: nobody will fetch this result, so
+				// cancel pre-dispatch (best effort, per the producer's
+				// cancellation contract). A client retry with the same id
```
roytman:
```go
// llm-d-async/producer@v0.9.0/redis_sortedset_producer.go:206-221
// Clear any stale cancellation marker for this request ID before enqueue...
pipe.Del(ctx, api.RequestCancellationKey(r.ReqID()))
pipe.Set(ctx, api.RequestActiveTokenKey(r.ReqID()), ir.RequestToken, activeTTL)
pipe.ZAdd(ctx, targetQueue, redis.Z{Score: score, Member: string(msgBytes)})
```
`SubmitRequest` clears the cancellation marker and resets the active-token marker — that part of the comment is accurate. But it does not remove any prior entry for the same id from the sorted-set queue. So if a client retries an id while the original submission is still sitting unpopped in the queue, you now have two queue entries for the same id. Whether that causes a double-dispatch depends entirely on whatever token check the consumer does at pop time, and that consumer-side code isn't in the vendored producer package, so it can't be verified from what's available in this repo's dependency tree. "Re-submits cleanly" asserts a guarantee (no double-dispatch) that the cited function doesn't itself provide - it's a claim about a different, unverified part of the system. Worth either narrowing the comment to what `SubmitRequest` actually does, or explicitly citing the consumer-side guarantee it's relying on.

roytman (commented):
Thank you for the PR.
Several additional comments:
- Test coverage gaps (all tests force wakeup_mode: poll): the Redis pub/sub notify wake-up path — the actual mechanism the PR advertises — is completely untested. Same for the quota fail-open branch (Redis error case) — only the happy path is exercised. Cross-tenant `DELETE` is untested (only cross-tenant GET is covered).
- In addition to the unit tests we have to check posibility to add CI e2e tests
- Validation is inconsistent. `FetchGraceSeconds < 0` is explicitly rejected, but `WaitCapSeconds, Timeouts[*].DefaultSeconds/MaxSeconds`, and `Quota.WindowSeconds/Limits` accept negative values and silently reinterpret them as "unset"/"unlimited" rather than erroring — same shape as the field that is checked, just missed.
- Read-path validation gap on tenant/id. The enqueue path rejects a `:` in the tenant (key-aliasing risk) and the inference path bounds `{id}` to a regex; the aux fetch/delete routes do neither, relying on incidental protection from the write-side check.
- 6 sites log recoverable Redis errors via Info + "error" key instead of `logger.Error(err, ...)`
- Fact duplication across doc/config/struct comments: timeout defaults, fetch-grace default, and "no header = untouched" are each stated 4-7 times across `docs/coordinator_async_broker.md`, `async_broker_config.go`, and inline comments.

docs/coordinator_async_broker.md
BenjaminBraunDev: The tenant is not the sole key for fetch/delete. Results live at results:req:\<tenant\>:\<id\> and reading another tenant's result requires knowing the request UUID, which is effectively unguessable for minted ids.

Agreed the `X-Team` header itself is trusted, so I dropped the "cannot probe" wording and added a deployment note saying the request id is the only secret protecting a stored result. I don't think we need to claim gateway must enforce this, since it's not really the proper threat model for `llm-d-router`.

pkg/coordinator/steps/asyncbroker/config.go
BenjaminBraunDev: This was intentional, since the user would have directly ask for a timeout of that length, but I see that it comes across as unpolished, so I will set the wait deadline to 1 hour by default.

docs/coordinator_async_broker.md
BenjaminBraunDev: true, I added a deployment note that fetch, delete, and /v1/models need adding to the coordinator's HTTPRoute. That actual example change probably belongs in a llm-d guide itself, especially if this is something that will be added soon

pkg/coordinator/pipeline/pipeline.go
BenjaminBraunDev: Added, the builder only allows async-broker to be first if it's present

pkg/coordinator/server/server.go
BenjaminBraunDev: Agreed, corrected comment and also clarified steps must use paths of their own

pkg/coordinator/steps/asyncbroker/wait.go
BenjaminBraunDev: fixed this by keeping a refcounted list of channels per key, so publish wakes all waiters and the last one out is unsubscribed. Added tests for two waiters on one key + a notify-mode wait test.

pkg/coordinator/steps/asyncbroker/quota.go
BenjaminBraunDev: Release failures are now logged in step logger, wait loop logs the first lookup failure per hold, classification failures are already logged.

pkg/coordinator/steps/asyncbroker/broker.go
BenjaminBraunDev: done, enqueue and fetch/delete failures now log the error server side and send message to client

pkg/coordinator/steps/asyncbroker/routes.go
BenjaminBraunDev: Done, status outside [100, 599] now returns 502 MALFORMED_RESULT instead of going to WriteHeader

pkg/coordinator/steps/asyncbroker/routes.go
BenjaminBraunDev: Fixed. After an active-marker miss, lookup re-reads the mailbox once, so a result landing between the two reads answers 200 instead of 410.

pkg/coordinator/steps/async_broker_routes.go
BenjaminBraunDev: Yeah agree we should call `CancelRequests`, made DELETE call `CancelRequests` before removing the mailbox + updated doc

pkg/coordinator/steps/async_broker.go
BenjaminBraunDev: Reworked this: a retry with a client-chosen id now first tries to reattach instead of enqueueing a duplicate. A dispatched request is reused, a still queued cancelled copy is revived by clearing the cancel marker, and a cancelled tombstone routes the retry to a fresh enqueue, so nothing should double dispatch.

BenjaminBraunDev:
@roytman Unit testing and the small items should all be addressed. I'm running benchmarks again with these new changes. For CI e2e, we would want an HTTPRoute change and a deployment guide on the llm-d side, so I say we hold off on that until we have a guide, and in a different PR. Thanks for the in-depth review!

pkg/coordinator/steps/asyncbroker/routes.go
BenjaminBraunDev: Reran the benchmark with those conditions and got zero 410s in 3600 requests.

BenjaminBraunDev:
Okay benchmarks finished, I verified each of the changes and updates to make sure prior issues are now fixed.

config/coordinator/coordinator.yaml
```diff
@@ -94,6 +94,49 @@ pipeline:
...
+    #     # wait_cap_seconds bounds held wait connections, ending the hold
+    #     # with the 202 response. Holds otherwise run to the request
+    #     # deadline.
+    #     # wait_cap_seconds: 55
+    #     # fetch_grace_seconds is the mailbox TTL applied after a delivered
+    #     # fetch. Zero deletes the result on delivery.
+    #     # fetch_grace_seconds: 60
+
     # -------------------------------------------------------------------
     # replace-media-urls: first step. Downloads any http(s) image_url
```
roytman: If `async-broke` must be the first step, this comment should be rephrased to something like "replace-media-urls: an optional step executed prior to request content processing. Downloads any http(s) image_url ...."

pkg/coordinator/steps/async_broker_config.go
```diff
@@ -0,0 +1,325 @@
...
+	// stamped. Queued dispatches are stamped by the AP's merge policy
+	// (lane_objectives), not here.
+	Objectives      map[string]asyncObjectivePair `json:"objectives,omitempty"`
+	ObjectiveHeader string                        `json:"objective_header,omitempty"`
+	FairnessHeader  string                        `json:"fairness_header,omitempty"`
+}
+
+func (c *asyncBrokerConfig) applyDefaults() {
+	if c.ModeHeader == "" {
+		c.ModeHeader = "X-AP-Mode"
```
roytman: Should the header be defined as a constant (the same as the others)? `async_broker_test.go` uses it several times.

pkg/coordinator/steps/async_broker_config.go
```diff
@@ -0,0 +1,325 @@
...
+		c.Quota.Attribute = "userid"
+	}
+	if c.Quota.WindowSeconds == 0 {
+		c.Quota.WindowSeconds = 300
+	}
+	if c.WakeupMode == "" {
+		c.WakeupMode = "auto"
+	}
+	if c.ForwardHeaders == nil {
+		c.ForwardHeaders = []string{"x-llm-d-slo-ttft-ms", "x-llm-d-slo-tpot-ms"}
```
roytman: `x-llm-d-slo-ttft-ms`, `x-llm-d-slo-tpot-ms`, `x-llm-d-inference-objective` and `"x-llm-d-inference-fairness-id"` are redefined in `pkg/epp/metadata/consts.go`

pkg/coordinator/steps/async_broker_config.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+	if c.ObjectiveHeader == "" {
+		c.ObjectiveHeader = "x-llm-d-inference-objective"
+	}
+	if c.FairnessHeader == "" {
+		// Matches api.FairnessIDHeader in llm-d-async; the constant is not in
```
roytman: The comment even flags that the author checked for a matching constant- just in the wrong repo

roytman (commented):
Hi @BenjaminBraunDev, thanks for the fixes. Please resolve the CI failure and the conflict.

`async-broker` responds on `/v1/models` according to its configuration. llm-d delegates this question to vLLM; there may be a misconfiguration here. 

Given that the `async-broker` step contains several files, it makes sense to place them into a dedicated package.
Furthermore, because it relies on Redis and may require changes to gateway routing rules, clear documentation will be necessary.

config/coordinator/coordinator.yaml
BenjaminBraunDev: done

pkg/coordinator/steps/async_broker_config.go
BenjaminBraunDev: done

pkg/coordinator/steps/async_broker_config.go
BenjaminBraunDev: fixed

pkg/coordinator/steps/async_broker_config.go
BenjaminBraunDev: fixed

BenjaminBraunDev:
@roytman Thanks, addressed each comment.

Also removed the /v1/models endpoint (did this initially for parity with the OpenAI API, but looks like it was already forwarded to model servers) and moved step files to their own `asyncbroker` package. For docs, docs/coordinator_async_broker.md covers the Redis requirements and routing changes in the deployment notes section.

docs/coordinator_async_broker.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Deployment notes
+
+- The gateway must route `GET/DELETE /v1/requests/*` to the coordinator. Stock llm-d routing forwards only the inference paths, so these need adding to the coordinator's HTTPRoute.
+- The tenant header is trusted as asserted, the same as everywhere else on the llm-d serving path. Request id is the only secret protecting a stored result, so clients that need an unguessable handle should omit `X-Request-Id` and use the minted UUID.
+- Set `result_ttl_seconds` on every AP queue the step feeds, or unfetched results never expire.
+- Redis needs keyspace notifications enabled for the wait wake-up (`notify-keyspace-events Kl`). The step detects their absence and falls back to polling.
```
roytman, line 140: This step relies on two more Redis preconditions that aren't stated here, and neither `llm-d-async`.

**Topology**: `broker.go:96` uses `redis.NewClient(opts)`, which is the `go-redis`'s standalone client, with no Cluster or Sentinel support. 
`llm-d-async`'s own connection code (`pkg/redis/redis_conn.go`) does the identical thing, and its reference install script deploys a single bitnami/redis release with no cluster/replication mode set. So if the AP's Redis is ever fronted by Sentinel or run as a Cluster, this client can't follow a Sentinel failover or a Cluster MOVED redirect. Worth stating explicitly: `redis_url` must be a standalone endpoint (directly, or via a cluster-aware proxy presenting a standalone interface), and Redis should be deployed as a standalone server.

**Eviction policy**: Correctness depends on several TTL'd keys (active-token markers, cancellation markers, quota counters, result mailboxes) surviving to their natural expiry - `lookupResult`, `reattachRetry`, and the fail-open quota logic all key off "does this Redis key still exist." `maxmemory-policy` alone doesn't fix this: `noeviction` only takes effect once `maxmemory` is actually set, and by default `maxmemory` is `0` (unlimited). The reference [install script](https://github.com/llm-d/llm-d-async/blob/68155ce4d10ab67d937600758d7e3096672c00d5/deploy/install.sh#L427) from `llm-d-async` deploys plain bitnami/redis with no memory limit configured at all, so today there's nothing to trigger eviction either way, by accident rather than by design. The real requirement is both together: an actual `maxmemory` value and `maxmemory-policy noeviction`, so a write past the limit fails loudly (surfaced through this step's existing error handling) instead of silently evicting a live key. And `maxmemory` should sit with real headroom below the container's memory limit; fork/replication buffer overhead during a BGSAVE can transiently exceed it, and too little headroom trades a graceful Redis-level OOM error for a hard kernel/cgroup OOM-kill.

Might be worth raising against `llm-d-async` too, since that's where the constraint actually originates.

roytman (commented):
Thank you for the updates.

A couple of last Redis-related things, but I'm not a Redis expert, so please take them with a grain of salt:

- I think some Redis installation instructions could be useful. Beyond the two gaps in the inline comment (topology, eviction policy), `llm-d-async`'s own docs don't go much further than "you need a running Redis instance" and  the most specific Redis guide ([guides/batch-serving/asynchronous-processing/redis/README.md](https://github.com/llm-d/llm-d/blob/main/guides/batch-serving/asynchronous-processing/redis/README.md)) is just a bare helm install redis bitnami/redis command, with nothing on sizing, persistence, or hardening. Might be worth a pointer here, or raising it against llm-d-async directly since that's the more natural home for general Redis deployment guidance.

- Related: nothing states what Redis ACL permissions this needs for a non-default-user connection. At minimum: @scripting (the three EVAL-based scripts — quota acquire/release, un-cancel), @pubsub (wait-mode wake-up), and ideally CONFIG access for the keyspace-notification auto-detect. The CONFIG one already degrades gracefully to polling if blocked, so it's not load-bearing — but worth a line for anyone running a locked-down ACL user.

Otherwise, the code /lgtm

docs/coordinator_async_broker.md
BenjaminBraunDev: I added this to the deployment notes and opened llm-d/llm-d-async#421 for the broader discussion since the constraints should be handled there.

BenjaminBraunDev:
Added a line in deployment notes documenting what permissions Redis needs, differing further setup docs to llm-d/llm-d-async#421

BenjaminBraunDev:
@ahg-g
Both @jtechapps and @roytman have given this a good review, do you think we can get this merged?

elevran:
@BenjaminBraunDev - unbderstand and agree that we want to expose async APIs to clients. I see several options for implementing this in llm-d (as a service behind the Gateway API ingress point):

- standalone dedicated API server
- coordinator based extensions
- existing batch API server extension

Can you please highlight the pros and cons of these options and the reasons for preferring the Coordinator? 

PS - need to resolve conflicts on the PR...

ahg-g:
Agree on documenting the pros and cons. My preference is to have this as a coordinator extension, standalone service adds yet another component to the ecosystem that the coordinator can fill, while the batch API is not aligned with the gap we are trying to address (seamlessly handle sync traffic via a message queue, as in hiding the fact that these requests are not batch in principle)

release-notes.d/unreleased/2325.md
```diff
@@ -0,0 +1,13 @@
+---
+pr: 2325
```
ahg-g: pls remove this file and add it as a release note on the issue description

BenjaminBraunDev: done

BenjaminBraunDev:
@elevran @ahg-g 
We actually built the standalone server first, assuming that's what you meant by a dedicated API server. llm-d/llm-d-async#392 was a complete standalone frontend, and the review on llm-d/llm-d-async#376 is where we decided to move it into the coordinator. So with that in mind:

Standalone API server (llm-d-async#392)
- Pro: Is isolated, releases on its own cadence and can't touch the sync serving path.
- Pro: scales independently, which is a pro in the context of wait mode holding one open connection per waiting client.
- Con: another component to build, deploy, and monitor. Around a third of the code was rebuilding server plumbing the coordinator already has (bootstrap, health endpoints, reverse proxying, request parsing).
- Con: live traffic tied to async quota has to be classified and stamped the same way as queued traffic, so the frontend and the AP had to mirror each other's Redis bookkeeping. Its live mode was also a reverse proxy in front of the same gateway the AP dispatches through, so a second proxy.

A slimmer version that only exposes enqueue and fetch would avoid the proxy duplication, but it can't tie live traffic to the same quota and priority account, which is why we had the frontend combined with the sync path.

Coordinator step (this PR)
- Pro: one entry point and one code path. Dispatched requests re-enter the same pipeline, so queued traffic takes the same steps, EPP phasing, and PD path as sync traffic, with nothing to keep in sync by hand.
- Pro: passthrough is the same as skipping async queues while still taking on priority and affecting quotas. The step classifies, stamps objective and fairness server side, and lets the rest of the pipeline serve the request. Whereas the old frontend did this by proxying the whole request.
- Pro: no new component + the step is optional. Requests without the mode header only cost one header lookup. No Redis connection is formed unless the step is actually configured.
- Con: the broker ships on the coordinator's release cadence, and held wait connections sit in the coordinator.

A batch solution seems out of line with the initial design goal of having a single endpoint to deliver sync and async requests to. Batch solves a different problem and can sit parallel to this system, feeding into the same async queues the plugin does.

Coordinator allows easy composition with existing setups since you can use the same gateway for async and sync, using headers to differentiate.

ahg-g:
can you pls rebase just so we get the markdown test to pass?

ahg-g:
Actually this is passing now, so it is good to go from my side.

@elevran does the above provide a satisfactory answer?

ahg-g (approved):

PR #2329, opened 2026-08-08
feat(epp): align prefix-cache tokens to UTF-8 character boundaries

### LINKED ISSUE #1711: [Optimization] Support configurable bytesPerToken or language presets to prevent cache boundary shift
Hi team,

I briefly mentioned this idea in Slack earlier. I noticed that the estimate backend in token-producer (`pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go`) hardcodes `bytesPerToken = 4` to pack raw input bytes into pseudo-tokens for prefix hashing.

For ASCII-dominant languages this is reasonable (~4 chars ≈ 1 token), but for Chinese (UTF-8, 3 bytes/char), each pseudo-token spans only ~1.33 characters. This **undercounts tokens by ~25%**, and more critically, causes 3-byte characters to frequently **misalign with the 4-byte boundary**. As a result, minor prompt edits can shift the hash-block boundaries and severely degrade cache hit rates.

I’d love to work on this by introducing a configurable `bytesPerToken` or language-prefixed presets in the estimate backend. I'll use `llm-d-inference-sim` locally to prototype and validate the cache hit rate improvement.

Could you please **assign this issue to me**? I'd be glad to submit a PR once the local simulation tests pass. Thanks!

NineSue:
**What type of PR is this?**

/kind feature

Two things here:

First — replace the fixed 4-byte slicing in `packBytes` with UTF-8 character-aware
packing. Before, 3-byte CJK characters (or 4-byte emoji) could land across two
pseudo-token slots, so any small edit to a non-ASCII prompt shifted every downstream
block hash and trashed the prefix cache. Now `utf8.DecodeRune` reads the character
boundary from the bytes themselves and packs whole characters per slot.

The original issue suggested making `bytesPerToken` configurable or adding language
presets — doesn't really work. One number can't handle mixed-language prompts (set
it to 3 and English wastes slots; set it to 4 and CJK breaks). UTF-8 already encodes
width in the leading byte. Just read it. No knob needed.

Second — @hhk7734 spotted that the original code mixed text bytes and multimodal
placeholder hashes into a single `[]byte` before feeding it to `packBytes`. When
`packBytes` inserted UTF-8 boundary padding, it shifted the placeholder offsets so
feature records pointed at wrong tokens. Fixed by splitting them into separate paths
via `estimatedTokenStream`: text is buffered and flushed through `packBytes` first,
then placeholder hashes are inserted directly as `uint32`. Offsets stay accurate
regardless of padding.

Pure ASCII output is byte-identical to the legacy loop. All existing tests pass
unchanged. New tests cover CJK, emoji, invalid UTF-8 byte packing, and multimodal
offset consistency after UTF-8 text.

Credit to @hhk7734 for catching the offset-shift bug.

Fixes #1711

```release-note
Prefix-cache pseudo-token packing now respects UTF-8 character boundaries instead of
a fixed 4-byte stride. Pure ASCII prompts are unchanged. Non-ASCII prompts produce
character-aligned slots that prevent wholesale cache invalidation on small edits.
Multimodal placeholder offsets no longer drift from UTF-8 boundary padding.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"strconv"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/cespare/xxhash/v2"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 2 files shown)


NineSue:
@hhk7734 caught a real bug — when text bytes and multimodal placeholder hashes share one byte stream, UTF-8 boundary padding from packBytes can shift the feature offsets. Fixed now: text and placeholder hashes go through separate paths. Also switched to utf8.DecodeRune instead of the hand-rolled bit-masking. Thanks!

NineSue:
@vMaroon Heads up on #2329 — the UTF-8 issue (#1711) still reproduces on current main: packBytes still packs fixed 4-byte strides, so non-ASCII prompts still invalidate the prefix cache on small edits. Branch is 141 commits behind and needs a port onto the new TokenizedRequest / per-prompt-features API (estimatedTokenStream was removed). Before I invest in that port, is this direction still wanted upstream, or has the boundary-alignment idea been superseded?

PR #2330, opened 2026-08-08
Add Prometheus alerting rules for KV-cache and EPP health

satyamg1620:
**What type of PR is this?**

/kind feature

The EPP ships Prometheus metrics but no alerting rules, so operators have to write their own from scratch. This adds defaults for the KV-cache index, the KV-event stream, and the EPP request path.

Rules target the current `llm_d_router_epp_kv_cache_*` and `llm_d_epp_*` names, not the deprecated `kvcache_*` aliases.

Two expressions aggregate before comparing, which is load-bearing rather than stylistic:

- `request_error_total` carries an `error_code` label that `request_total` does not, and requests rejected before body parsing increment only the error counter. An unaggregated ratio matches nothing at all, so both sides are summed. The resulting figure is a fleet-wide signal and can exceed the true share of failed requests; this is documented in `docs/metrics.md`.
- TTFT series carry `fairness_id`, bounded at 1000 distinct values. Grouping by `(le, model_name, streaming)` keeps one alert per model instead of one per label combination.

The `PrometheusRule` manifest is the single source. Plain-Prometheus users extract the groups with `yq '.spec'`, so the two forms cannot drift.

Fixes https://github.com/llm-d/llm-d-kv-cache/pull/676

**Release note**:
```release-note
Add default Prometheus alerting rules for the KV-cache index, KV-event stream, and EPP request path. Apply `deploy/components/monitoring/prometheus-rules.yaml` with prometheus-operator, or extract the rule groups with `yq '.spec'` for a plain Prometheus.
```

**Test plan**

- [x] `make verify-alert-rules` — 7 rules check clean; 13 promtool cases cover firing and quiet paths for every alert, using the label sets the EPP  actually emits
- [x] Suite fails against unaggregated expressions, confirming it catches the  label-match and cardinality faults rather than passing vacuously
- [x] `kubectl kustomize deploy/components/monitoring` builds; manifest passes server-side validation against the prometheus-operator CRD
- [x] Rules load and evaluate healthy in a real Prometheus scraping a live EPP on kind; `EPPHighRequestErrorRate` reaches `firing` under driven traffic
- [x] `make presubmit`


gyliu513:
@satyamg1620 based on issue https://github.com/llm-d/llm-d/issues/1983, can you move this PR for llm-d/llm-d repo? Our target is component user-facing assets for observability should go with llm-d/llm-d.

satyamg1620:
Moving this PR to llm-d/llm-d repo, so closing this PR

PR #2331, opened 2026-08-09
fix: lint in testcase

zdtsw:
**What type of PR is this?**
/kind test

probably caused by different PRs touched the same tests in race condition
fix error:
```
docker run --rm -u $(id -u):$(id -g) -v $(pwd):/app:Z -w /app -v /home/runner/.cache/llm-d-gomodcache:/go/pkg/mod:z -v /home/runner/.cache/llm-d-gobuildcache:/go/cache:z ghcr.io/llm-d/llm-d-builder:dev sh -c 'GOFLAGS=-buildvcs=false golangci-lint run --config=./.golangci.yml && typos'
Error: pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer_test.go:959:21: Error return value of `producer.PreRequest` is not checked (errcheck)
	producer.PreRequest(reqCtx, req, res)
```

Fixes #

```release-note
NONE
```

cc @dmitripikus

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer_test.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer_test.go
@@ -956,7 +956,7 @@ func TestInFlightLoadProducer_JanitorSkipsLiveRequest(t *testing.T) {
 
 	req := makeTokenRequest("req-janitor", 4) // 4 input + 6 output = 10 tokens
 	res := makeSchedulingResult(endpointName)
-	producer.PreRequest(reqCtx, req, res)
+	_ = producer.PreRequest(reqCtx, req, res)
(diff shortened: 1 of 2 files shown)


dmitripikus:
The above looks right as a lint-only fix — both files already use _ = ...PreRequest(...) at every other call site, and these two were the outliers that caused linter issue. Bringing them into line with the surrounding pattern is the smallest correct change.

One note for follow-up (not for this PR): on https://github.com/llm-d/llm-d-router/pull/2215 I switched the flowcontrol/eviction and flowcontrol/integration tests from _ = ...PreRequest(...) to require.NoError(t, ...PreRequest(...)) so a non-nil return on the happy path actually fails the test rather than being silently discarded. The same argument applies to the dataproducer/inflightload and dataproducer/p2psource tests, but converting all ~30+ call sites is a separate sweep and doesn't belong in a lint fix.

vMaroon (approved):

Issue #2332, opened 2026-08-09
[Feature] Add model-affinity-filter for multi-cluster model-aware routing


liavweiss:
A new scheduling filter plugin (`model-affinity-filter`) that enables model-aware routing in multi-cluster hub EPP deployments.

**Why is this needed**:

In a multi-cluster hub topology, the Hub EPP discovers spoke clusters via `multicluster-file-discovery`. Each spoke serves a specific model (e.g., Spoke1 serves TinyLlama, Spoke3 serves Qwen). Currently there is no built-in way for the Hub EPP to route a request only to spokes that serve the requested model.

The existing `bylabel` / `label-selector-filter` uses static values configured at deploy time and does not inspect the request. We need a dynamic, per-request filter that:

1. Reads the target model from the request (header or body `model` field)
2. Matches it against an endpoint label (e.g., `model: TinyLlama/TinyLlama-1.1B-Chat-v1.0`)
3. Retains only matching endpoints as candidates for scoring/picking

**Detailed description**:

- Plugin type: `model-affinity-filter`
- Interface: `scheduling.Filter`
- Model resolution order: configurable request header (`x-gateway-model-name`) -> request body `model` field
- Configurable parameters:
  - `labelKey` (default: `model`) - the endpoint label key to match against
  - `modelHeader` (default: `x-gateway-model-name`) - the request header to read model name from
- Behavior:
  - If no model is resolved - all endpoints pass through unfiltered
  - Endpoints without the label are filtered out
  - If no endpoints match - returns empty list (scheduler reports routing error)

**Use case**:

Multi-cluster Hub EPP: route requests to the correct spoke cluster based on which model it serves. For example, with 3 spokes where Spoke1 and Spoke2 serve TinyLlama and Spoke3 serves Qwen, a request for TinyLlama will only be routed to Spoke1 or Spoke2.

elevran:
> Each spoke serves a specific model (e.g., Spoke1 serves TinyLlama, Spoke3 serves Qwen).

I don't think we make that assumption today. The hub multicluster EPP selects between (remote) endpoints of the **same** model.
Different models should have different hub EPPs for now. We may wish to relax this constraint in the future but its impact needs to be analyzed.

liavweiss:
Thanks for the feedback @elevran

I believe the single-model assumption is preserved in this architecture:

1. Spoke EPPs each serve exactly one model - their metrics (KV-cache, queue depth) reflect only that model's capacity.
[...]

PR #2333, opened 2026-08-09
feat(epp): add model-affinity-filter for multi-cluster model-aware routing

liavweiss:
**What type of PR is this?**
/kind feature

Adds a new `model-affinity-filter` scheduling plugin for multi-cluster hub EPP deployments.

In a hub topology, the Hub EPP routes inference requests to downstream spoke clusters discovered via `multicluster-file-discovery`. Each spoke serves a specific model. This filter reads the target model name from the request (via the `x-gateway-model-name` header set by IPP, or falling back to the `model` field from the request body) and retains only endpoints whose `model` label matches - enabling model-aware routing without requiring per-model scheduler profiles.

**Fixes**
#2332 

```release-note
Add model-affinity-filter scheduling plugin for multi-cluster hub EPP deployments that filters endpoints by model label.

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -111,6 +111,7 @@ import (
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requesthandling/parsers/vllmhttp"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/filter/bylabel"
 	endpointattributefilter "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/filter/endpointattribute"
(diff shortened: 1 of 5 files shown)


elevran:
@liavweiss thanks. Having a single EPP with multiple models may break some assumption in the code base. For example, flow control and saturation assume that all endpoint contribute to capacity. With multiple models in the same EPP, flow control provides wrong signals (it should have admission, capacity and saturation per model - not the aggregate). There might be other cases.
There is always an option to relax this in the future, but warrants a discussion outside of a PR

szedan-rh:
@liavweiss - we need to have a proposal doc how we want to implement that, while keeping in mind the epp main target.

PR #2334, opened 2026-08-09
refactor: move envoy subset filter to a screener plugin

elevran:
**What type of PR is this?**

/kind cleanup

The envoy subset filter (envoy.lb.subset_hint / x-gateway-destination-endpoint-subset) was hardcoded in `DatastoreEndpointCandidates.Locate`. This refactor moves it into a request-control Screener plugin (`pkg/epp/framework/plugins/requestcontrol/screener/envoysubset`) and registers it as a system built-in from the runner, gated only by the existing `--disable-endpoint-subset-filter` CLI flag. No config-file knob is added, so the user cannot enable it in two places at once.

DatastoreEndpointCandidates and CachedEndpointCandidates now always return the full pod set; the cache key is no longer metadata-dependent. `InferenceRequest` gains a `Metadata` field that carries the Envoy dynamic-metadata map from `envoy.ExtractMetadataValues` to the screener, mirroring how `Headers` already flows.

The `--disable-endpoint-subset-filter` CLI flag is the single switch. Behavior is unchanged.

NA. Cleanup following the introduction of endpoint screening plugins.

```release-note
NONE
```

New tests in `pkg/epp/framework/plugins/requestcontrol/screener/envoysubset/plugin_test.go` cover the screener
Updated tests in `pkg/epp/requestcontrol/candidates_test.go` cover the simplified Locate and cache-key behavior

### DIFF
--- a/pkg/epp/framework/interface/scheduling/types.go
+++ b/pkg/epp/framework/interface/scheduling/types.go
@@ -52,6 +52,9 @@ type InferenceRequest struct {
 	Body *fwkrh.InferenceRequestBody
 	// Headers is a map of the request headers.
 	Headers map[string]string
+	// Metadata is the raw Envoy dynamic-metadata map carried by the ext_proc request.
+	// Used by request-control plugins that screen on Envoy-supplied hints (e.g. subset load balancing).
+	Metadata map[string]any
 	// Request Objective
 	Objectives RequestObjectives
 	// FairnessID is the identity used by the flow control system to group requests into fairness queues.
--- a/pkg/epp/framework/interface/scheduling/types.go
+++ b/pkg/epp/framework/interface/scheduling/types.go
@@ -52,8 +52,7 @@ type InferenceRequest struct {
 	Body *fwkrh.InferenceRequestBody
 	// Headers is a map of the request headers.
 	Headers map[string]string
-	// Metadata is the raw Envoy dynamic-metadata map carried by the ext_proc request.
-	// Used by request-control plugins that screen on Envoy-supplied hints (e.g. subset load balancing).
+	// Metadata is the raw dynamic-metadata map carried by the ext_proc request.
 	Metadata map[string]any
 	// Request Objective
 	Objectives RequestObjectives
--- a/pkg/epp/framework/plugins/requestcontrol/screener/envoysubset/plugin.go
+++ b/pkg/epp/framework/plugins/requestcontrol/screener/envoysubset/plugin.go
@@ -1,0 +1,115 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package envoysubset implements a request-control Screener that narrows the
+// candidate endpoint set to the subset indicated by Envoy's dynamic metadata
+// under envoy.lb.subset_hint / x-gateway-destination-endpoint-subset.
+//
+// The plugin is a system built-in: the runner registers it directly on
+// startup, gated only by the --disable-endpoint-subset-filter CLI flag. It is
+// not exposed via the plugin registry and is not user-configurable.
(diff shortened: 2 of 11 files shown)


pkg/epp/framework/plugins/requestcontrol/screener/envoysubset/plugin.go
elevran, line 1: unsure if this is an envoy specific feature or any GAIE compliant gateway supports it. 
Can rename path from `envoysubset` to `gwsubset` (or similar)

pkg/epp/framework/plugins/requestcontrol/screener/gwsubset/plugin.go
ahg-g: It is not an envoy specific feature, this is part of inference gateway protocol.

pkg/epp/framework/interface/scheduling/types.go
```diff
@@ -52,6 +52,9 @@ type InferenceRequest struct {
 	Body *fwkrh.InferenceRequestBody
 	// Headers is a map of the request headers.
 	Headers map[string]string
+	// Metadata is the raw Envoy dynamic-metadata map carried by the ext_proc request.
```
ahg-g:
```suggestion
	// Metadata is the raw dynamic-metadata map carried by the ext_proc request.
```

pkg/epp/requestcontrol/director.go
```diff
@@ -302,7 +303,7 @@ func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestCo
 		return reqCtx, err
 	}
 
-	endpointCandidates := d.endpointCandidates.Locate(ctx, reqCtx.Request.Metadata)
+	endpointCandidates := d.endpointCandidates.Locate(ctx, nil)
```
ahg-g: should we remove the second parameter from the `Locate` interface?

ahg-g (commented):
By moving this to the screener, the logic in requestcontrol/candidates.go becomes a cached list of all endpoints. I think this is better placed in pkg/epp/datastore/datastore.go; basically we maintain a slice that mirrors the pods map and is returned by PodList(); wdyt?

elevran:
> By moving this to the screener, the logic in requestcontrol/candidates.go becomes a cached list of all endpoints. I think this is better placed in pkg/epp/datastore/datastore.go; basically we maintain a slice that mirrors the pods map and is returned by PodList(); wdyt?

Agree the candidates layer is thin now. Confirming my understanding of your feedback:

The mirror-slice idea drops the decorator, but the cache does two things: 
1. lower contention on the datastore; and 
1. a stable snapshot for the flow-control dispatch loop.

Having a mirror slice addresses the first but not the second.
Replacing the cache with a live slice changes dispatch-loop semantics that this PR isn't trying to touch.

If snapshot stability isn't actually that useful, happy to take this as a follow-up. But it feels like a separate refactor to me.
Is that what you had in mind?

ahg-g:
> > By moving this to the screener, the logic in requestcontrol/candidates.go becomes a cached list of all endpoints. I think this is better placed in pkg/epp/datastore/datastore.go; basically we maintain a slice that mirrors the pods map and is returned by PodList(); wdyt?
> 
> Agree the candidates layer is thin now. Confirming my understanding of your feedback:
> 
> The mirror-slice idea drops the decorator, but the cache does two things:
> 
> 1. lower contention on the datastore; and
> 2. a stable snapshot for the flow-control dispatch loop.
> 
> Having a mirror slice addresses the first but not the second. Replacing the cache with a live slice changes dispatch-loop semantics that this PR isn't trying to touch.
> 
> If snapshot stability isn't actually that useful, happy to take this as a follow-up. But it feels like a separate refactor to me. Is that what you had in mind?

May be I am missing something, but which snapshot are you referring to? The code in pkg/epp/requestcontrol/candidates.go creates a single cache entry with key `__all_endpoints__`, this entry is basically a list of all pods we get from `d.datastore.PodList(datastore.AllPodsPredicate)` (should be renamed to EndpointList) that we then take a snapshot of at line 313 in director.go. 

In https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/flowcontrol/controller/internal/processor.go#L441 Locate is invoked during the dispatch loop (which doesn't respect subsetting) to check for saturation again, but that can be done on whatever datastor.PodList returns. 

I actually just noticed that this PR breaks subsetting during saturation detection at admission: https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/requestcontrol/admission.go#L75.


The best course of action in my opinion is:
1) Drop support for subsetting, it is just not compatible with flow control. 
2) Remove requestcontrol/candidates.go, change datastore.PodList to return a list that is tracked internally in datastore as a mirror of the map.

PR #2335, opened 2026-08-09
feat(epp): move conditional-decode gate into prefix-based-pd-decider

### LINKED ISSUE #1686: Conditional-decode 412 gate is hard-coded; make it configurable via a plugin

dmitripikus:
**What type of PR is this?**

/kind feature

Moves the RFC 7240 `Prefer: if-available` conditional-decode 412 gate out of the director's hard-coded check and into a configurable plugin behavior on the existing `prefix-based-pd-decider`.

Today the director hard-codes "reject when the primary decode endpoint has zero matched prefix blocks", which is expressed in tier-weighted blocks and cannot be tuned. Operators have no way to reject speculative decodes based on the actual token count that would need to be prefilled locally.

This PR:

- Adds `PreRequest` to `PrefixBasedPDDecider`. When a request carries `Prefer: if-available` and the non-cached suffix is at least `nonCachedTokens` (unweighted, so a RAM-cached prefix contributes its full length), the plugin returns `errcommon.Error{Code: PreconditionFailed}` — the director's runner preserves the typed code end-to-end so the 412 response is byte-identical to what the director used to emit.
- Honors `promptTokens` in the gate the same way `disaggregate()` does: prompts shorter than `promptTokens` skip both remote prefill and the 412 gate, so the operator's short-prompt shortcut applies uniformly.
- Removes `primaryEndpointHasCachedPrefix` and the inline gate block from `pkg/epp/requestcontrol/director.go`.
- Adds a director-side default-deny: after PreRequest plugins run, any `Prefer: if-available` request that no plugin marked as handled is rejected with 412 (`conditional-decode request received but no gate plugin is configured`). Deployments without a gate plugin fail fast rather than silently forwarding a speculative decode to a decoder that may not hold the KV cache.
- `nonCachedTokens: 0` disables the gate (consistent with disaggregation being off).

Groundwork was landed separately in #2215 (adds `error` return to the `PreRequest` interface).

Fixes #1686

**Release note**:
```release-note
The conditional-decode 412 gate (RFC 7240 `Prefer: if-available`) is now enforced by the `prefix-based-pd-decider` plugin instead of a hard-coded director check. Deployments that declare the plugin enforce the gate at the configured `nonCachedTokens` threshold (unweighted cached-block count) and honor its `promptTokens` short-prompt bypass. Deployments without a gate plugin fall back to a director-side default-deny: any `Prefer: if-available` request that no PreRequest plugin claimed is rejected with 412 (`conditional-decode request received but no gate plugin is configured`), surfacing a missing gate plugin as a fast fallback rather than a silent forward. To restore the previous "any weighted cache hit lets it through" behavior, set `nonCachedTokens: 1`.
```


docs/communication.md
```diff
@@ -361,6 +361,8 @@ The coordinator attempts an early decode immediately after rendering. This allow
 
 The coordinator adds the `Prefer: if-available` HTTP header to signal that the decode worker should only proceed if the KV cache is already available. If it responds with 412 Precondition Failed, the pipeline continues as normal.
 
+The 412 is emitted on the router (EPP) side by the `prefix-based-pd-decider` plugin when declared in the deployment's plugin config, using its `nonCachedTokens` threshold (see [disaggregation.md](disaggregation.md)). Deployments without the plugin do not enforce the gate and always forward the speculative decode.
```
roytman: Actually, `prefix-based=pd-decider` has 2 metrics: `nonCachedTokens` and a separate minimum prompt-length gate - `promptTokens`

docs/coordinator_architecture.md
```diff
@@ -343,6 +343,11 @@ decode serve this directly" question. The step sends the decode request with the
 The 412 response may carry scheduling or cache-locality hints in headers or body; these
 are not yet consumed by the coordinator.
 
+The 412 is emitted on the router (EPP) side by the `prefix-based-pd-decider` plugin when
+declared in the deployment's plugin config; see [disaggregation.md](disaggregation.md) for
+the plugin's `nonCachedTokens` threshold. Deployments without the plugin never emit 412
+and always forward speculative decode requests.
```
roytman: see the comment in the communication.md. 
Why do we need the same explanation twice?

docs/disaggregation.md
```diff
@@ -417,11 +417,16 @@ The `prefix-based-pd-decider` plugin makes the disaggregation decision according
...
   - If set to 0, the prompt-length gate is disabled
   - If set to a positive value, requests with fewer prompt tokens run locally on the decode worker without remote prefill
+  - Applies to the disaggregation decision only; the conditional-decode 412 gate below is unaffected
+
+**Conditional-decode 412 gate**
+
+When a request carries the RFC 7240 `Prefer: if-available` header (used by the coordinator's speculative early-decode step, see [coordinator_architecture.md](coordinator_architecture.md)), the plugin also acts as a request-time gate: if the non-cached suffix of the prompt on the chosen decode endpoint is at least `nonCachedTokens`, the request is rejected with HTTP 412 Precondition Failed so the coordinator restarts the pipeline at encode/prefill/decode. `nonCachedTokens: 0` disables the gate; the same non-cached-suffix threshold governs both the P/D routing decision and the 412 decision. Deployments that do not declare this plugin do not enforce the gate — conditional-decode requests are always forwarded. The gate reads unweighted cached-block state, so a RAM-cached prefix contributes its full token count.
```
roytman: The plugin works in the same way, with `promptTokens`. We have the same explanation for the third time. Can we reduce it?

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -42,8 +45,21 @@ func (p PrefixBasedPDDeciderConfig) validate() error {
...
-// compile-time type assertion
-var _ deciderPlugin = &PrefixBasedPDDecider{}
+// compile-time type assertions
+var (
+	_ deciderPlugin    = &PrefixBasedPDDecider{}
+	_ fwkrc.PreRequest = &PrefixBasedPDDecider{}
+)
+
+// errCondDecodeCacheMiss is the fixed 412 returned by the conditional-decode
+// gate when a decoder cannot satisfy the request from its KV cache. Byte-for-
```
roytman: We don't need the explanation about the previous implementation

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -102,6 +118,76 @@ func (d *PrefixBasedPDDecider) WithName(name string) *PrefixBasedPDDecider {
...
+// NonCachedTokens, the request is rejected with HTTP 412 so the coordinator
+// falls back to the full encode/prefill/decode pipeline. NonCachedTokens == 0
+// disables the gate. Requests without the header are a no-op.
+//
+// promptTokens is deliberately not honored here: it is a P/D routing shortcut
+// for short prompts, not a cache-adequacy predicate.
+//
+// Fails closed (412) when cache state cannot be read from the endpoint, since
+// a coordinator restart is cheaper than forwarding to a decoder that may not
+// hold the cache.
```
roytman: can we reduce the comment?

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -102,6 +118,76 @@ func (d *PrefixBasedPDDecider) WithName(name string) *PrefixBasedPDDecider {
...
 }
 
+// PreRequest gates requests carrying the RFC 7240 "Prefer: if-available"
+// header used by the coordinator's speculative early-decode step. When the
+// non-cached suffix of the prompt on the chosen decode endpoint reaches
+// NonCachedTokens, the request is rejected with HTTP 412 so the coordinator
+// falls back to the full encode/prefill/decode pipeline. NonCachedTokens == 0
+// disables the gate. Requests without the header are a no-op.
+//
+// promptTokens is deliberately not honored here: it is a P/D routing shortcut
```
roytman: Why, if the prompt is short, we can go forward to Decoder, and the user, if he wants, can configure `promptTokens = 0`

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -102,6 +118,76 @@ func (d *PrefixBasedPDDecider) WithName(name string) *PrefixBasedPDDecider {
...
+// for short prompts, not a cache-adequacy predicate.
+//
+// Fails closed (412) when cache state cannot be read from the endpoint, since
+// a coordinator restart is cheaper than forwarding to a decoder that may not
+// hold the cache.
+func (d *PrefixBasedPDDecider) PreRequest(ctx context.Context, request *scheduling.InferenceRequest,
```
roytman, line 131: the function looks very similar to `disaggregate`, can they be combined?

docs/disaggregation.md
roytman: **Suggestion**: Keep docs/disaggregation.md:428-430 as the single canonical explanation (it already lives next to the nonCachedTokens/promptTokens parameter docs, which is the right home). Shrink the other two to one-line cross-references with no restated mechanism, e.g.:

- communication.md:364-> "The 412 is enforced by the prefix-based-pd-decider plugin's conditional-decode gate; see `disaggregation.md` (add a correct link to the explanation)

- coordinator_architecture.md:346-349 -> same one-liner.

docs/communication.md
dmitripikus: I initially framed promptTokens as a routing-only knob, but you're right that disaggregate() already treats "prompt shorter than promptTokens" as "skip remote prefill", and having PreRequest reject those same requests with 412 forces the coordinator to do remote prefill anyway, which is the opposite of what the operator asked for. I'll honor promptTokens in PreRequest so both paths short-circuit on the same shortcut, and update the docs to describe both knobs together in one place.

dmitripikus: Done. Thank you!

docs/coordinator_architecture.md
dmitripikus: I've addressed it in commit c600063b.
Thank you!

docs/disaggregation.md
dmitripikus: Both comments above are addressed by commit c600063b

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
dmitripikus: Done in c600063b — comment now describes only what the variable is.
Thanks!

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
dmitripikus: PreRequest now honors promptTokens the same way disaggregate() does: prompts shorter than promptTokens forward to decode; operators who don't want that bypass set promptTokens = 0. Doc/tests/PR body updated to match.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
dmitripikus: Extracted needsRemotePrefill() helper in a393ddc5. Both disaggregate and PreRequest now thin-wrap it. The helper answers the decision.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
dmitripikus: Done in a393ddc5 — doc string is 3 lines now; the mechanism details moved to needsRemotePrefill's doc string where both callers can share them.

docs/disaggregation.md
```diff
@@ -417,11 +417,12 @@ The `prefix-based-pd-decider` plugin makes the disaggregation decision according
 
 **Parameter:**
 
-- `nonCachedTokens`: Number of non-cached tokens that trigger disaggregation
-  - If set to 0, disaggregation never occurs for any request
-- `promptTokens`: Minimum prompt length in tokens before prefix-cache-based disaggregation logic is applied
-  - If set to 0, the prompt-length gate is disabled
-  - If set to a positive value, requests with fewer prompt tokens run locally on the decode worker without remote prefill
+- `nonCachedTokens`: Number of non-cached tokens that trigger disaggregation, and the threshold for the conditional-decode 412 gate. `0` disables both behaviors.
```
roytman: it is not only conditional-decode 412 gate, but a regular disaggregation gate

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -102,65 +115,110 @@ func (d *PrefixBasedPDDecider) WithName(name string) *PrefixBasedPDDecider {
...
 	if err != nil {
-		logger.Error(err, "prefix decider: failed to get user input length in tokens")
-		return false
+		return false, fmt.Errorf("failed to get user input length in tokens: %w", err)
 	}
-
 	if d.config.PromptTokens > 0 && inputTokens < d.config.PromptTokens {
-		debugLogger.Info("Input is shorter than the promptTokens, no disaggregated PD")
-		return false
+		debugLogger.Info("Input shorter than promptTokens",
```
roytman: I'd add something like "disaggregation is not required"

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -102,65 +115,110 @@ func (d *PrefixBasedPDDecider) WithName(name string) *PrefixBasedPDDecider {
...
+	}
+	return primary.TargetEndpoints[0]
+}
+
+// disaggregate reports whether remote prefill should run for this request.
+// Fails soft: any read failure logs at ERROR and returns false so scheduling
+// falls back to the decode-only path.
 func (d *PrefixBasedPDDecider) disaggregate(ctx context.Context, request *scheduling.InferenceRequest, endpoint scheduling.Endpoint) bool {
-	logger := log.FromContext(ctx)
+	needs, err := d.needsRemotePrefill(ctx, request, endpoint)
```
roytman, line 176: In some cases, we can call this function twice; why not to store its results in the context?

docs/disaggregation.md
```diff
@@ -417,11 +417,12 @@ The `prefix-based-pd-decider` plugin makes the disaggregation decision according
...
+- `promptTokens`: Minimum prompt length in tokens before the plugin's routing and gating logic applies. Prompts shorter than this run locally on the decode worker without remote prefill; the 412 gate honors the same shortcut. `0` disables it.
+
+**Conditional-decode 412 gate**
+
+Requests carrying `Prefer: if-available` (used by the coordinator's speculative early-decode step, see [coordinator_architecture.md](coordinator_architecture.md)) are gated by the plugin using the same `promptTokens` / `nonCachedTokens` thresholds as the disaggregation decision: when the chosen decode endpoint's non-cached suffix would trigger remote prefill, the plugin returns HTTP 412 Precondition Failed so the coordinator restarts the pipeline at encode/prefill/decode. Deployments that do not declare this plugin do not enforce the gate — conditional-decode requests are always forwarded. Cache state is read as unweighted contiguous blocks, so a RAM-cached prefix contributes its full token count.
```
roytman, line 425: originally `prefix-based-pd-decider` was (is) a decider for `disagg-profile-handler` and did not work as a standalone plugin. I think it will be good to add a separate paragraph in the document where we can explain about conditional-decode and provide an example of yaml file configuration with `prefix-based-pd-decider`

roytman (commented):
There is another inherited issue:
The gate fails open, not closed, when the tokenizer hasn't run — getUserInputLenInTokens returns (0, nil) for a nil TokenizedPrompt (not an error). That makes nonCachedTokens = 0 - hitPrefixTokens, always below threshold, so the request is forwarded — the opposite of the "fails closed" claim in the PreRequest doc comment. This is reachable in production: director.go explicitly swallows DataProducer (tokenizer) failures, so a tokenizer backend hiccup silently disables the whole gate. The new test "no TokenizedPrompt: zero-token prompt trivially covered" (wantReject: false) locks this behavior in rather than catching it.

We should validate if we want to leave this behaviour or now. 
The possible fix is:
```go
func getUserInputLenInTokens(request *scheduling.InferenceRequest) (int, error) {
	if request == nil || request.Body == nil {
		return 0, errors.New("request or request body is nil")
	}
	if request.Body.TokenizedPrompt == nil {
		return 0, errors.New("prompt not tokenized")
	}
	return request.Body.TokenizedPrompt.TokenCount(), nil
}
```
One test needs to flip with it - `prefix_based_pd_decider_test.go`, "no TokenizedPrompt: zero-token prompt trivially covered":

```go
{
    name:            "no TokenizedPrompt: cache state unknown, fails closed",
    nonCachedTokens: 5,
    headers:         preferIfAvailable,
    result:          resultWithEndpoint(makeTestEndpoint(0)),
    request:         completionsRequestWithPrompt(fwkrh.Prompt{}),
    wantReject:      true,
},
```

docs/disaggregation.md
dmitripikus: Done. 
nonCachedTokens is now described as the threshold at which the plugin's gate fires, applied symmetrically to disaggregation and the 412 response, rather than only naming the 412 case as a "gate." 
Thanks!

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
dmitripikus: Done. Both short-input branches now spell out "disaggregation not required" in the debug log. Thanks!

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
dmitripikus: Done. needsRemotePrefill now memoizes (needs, err) on the request via PutAttribute (same pattern as PeerEndpointAttributeKey), so disaggregate populates it during scheduling and PreRequest reads without recomputing. Two existing tests exercised the same request with two endpoints — a test-only shape (production always pairs one request with one endpoint at PrimaryProfileName: decodeProfile) — I split them into fresh requests per scenario and added TestNeedsRemotePrefill_MemoizesOnRequest to show the reuse.
Thanks!

docs/disaggregation.md
dmitripikus: Done. The Prefix-Based PD Decider section intro now names both roles (decider inside disagg-profile-handler, standalone PreRequest gate) and their shared parameters, and the 412 gate subsection carries a gate-only EndpointPickerConfig example so someone landing there sees exactly what a standalone declaration looks like. Cross-links point at the profile handler config and the full-deployment examples for the combined-role setups. Thanks!

dmitripikus:
Regarding getUserInputLenInTokens(): fixed in 2598fff9 — it returns an error when TokenizedPrompt == nil. To keep the two callers of needsRemotePrefill consistent, 4fa11709 makes PreRequest log-and-forward on any gate-evaluation error too, matching disaggregate's soft-fail. Both log at Info. 412 is still returned when the gate legitimately fires (non-cached suffix meets the threshold). Tests updated. Thanks!

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -102,75 +125,143 @@ func (d *PrefixBasedPDDecider) WithName(name string) *PrefixBasedPDDecider {
...
+// zero-length input.
 func getUserInputLenInTokens(request *scheduling.InferenceRequest) (int, error) {
 	if request == nil || request.Body == nil {
 		return 0, errors.New("request or request body is nil")
 	}
-
-	if tp := request.Body.TokenizedPrompt; tp != nil {
-		return tp.TokenCount(), nil
+	if request.Body.TokenizedPrompt == nil {
+		return 0, errors.New("prompt not tokenized")
```
roytman, line 283: So we do return an Error?

dmitripikus: getUserInputLenInTokens() returns error now, to identify the fact that request.Body.TokenizedPrompt  length is unknown.
But the functionality of disaggregate did not change (only with INFO message about the error is added)
PreRequest()  functionality is aligned with disaggregate

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -102,75 +125,143 @@ func (d *PrefixBasedPDDecider) WithName(name string) *PrefixBasedPDDecider {
...
+// chosen endpoint meets NonCachedTokens. Returns (false, nil) when the plugin
+// is disabled, the prompt is shorter than PromptTokens, or the prompt is
+// shorter than NonCachedTokens. A non-nil error means the endpoint's cache
+// state or the request's input length could not be read; both callers
+// soft-fail on error — disaggregate returns false, PreRequest forwards.
+//
+// The outcome is memoized on the request: disaggregate populates it during
+// scheduling, and PreRequest reuses it without recomputing.
+func (d *PrefixBasedPDDecider) needsRemotePrefill(ctx context.Context, request *scheduling.InferenceRequest, endpoint scheduling.Endpoint) (bool, error) {
+	if request != nil {
```
roytman: We can check at the start of the method if `request == nil` and return (false, error),
Anyway `ReadRequestAttribute` returns the same; it will simplify the method

pkg/epp/requestcontrol/director.go
roytman, line 641: The old gate returned from `HandleRequest` immediately, before any `PreRequest` plugin ran. Now it's itself a `PreRequest` plugin, and `runPreRequestPlugins` never short-circuits — it runs every plugin and joins the errors. So other `PreRequest` plugins with side effects now execute for requests that end up rejected with 412. 
If a second plugin also errors, the two joined errors no longer unwrap to a single typed error in `prepareRequest`, and the typed `PreconditionFailed` collapses to Internal (500), which breaks the coordinator's 412 fallback path.

roytman (commented):
I'm not sure that it is a desirable behaviour:

If `prefix-based-pd-decider` isn't configured at all, a `Prefer: if-available` request is simply forwarded - no 412 check happens anywhere, silently and by design:

1. Nothing in the top-level plugins: list produces a `*PrefixBasedPDDecider` instance, so AddPlugins's type-assertion (request_control_config.go:105) never adds anything to preRequestPlugins for this purpose.
2. `runPreRequestPlugins` iterates over a plugin list that simply doesn't contain any conditional-decode check.
3. No code anywhere calls `routing.IsConditionalDecode` for a rejecting purpose — that's the only call site left, and it's inside the plugin that doesn't exist in this deployment.
4. The request proceeds through `HandleRequest` normally and gets forwarded to whichever decode endpoint scheduling picked, regardless of whether that endpoint actually has the prompt cached.

This re is another problematic scenario:
We have a deployment that does declare `prefix-based-pd-decider` - opting into the gate, but where the gate can't actually evaluate cache state because a dependency it needs (the `PrefixCacheMatchInfo` attribute, normally produced by `approx-prefix-cache-producer/prefix-cache-scorer`) is missing, e.g. because that scorer plugin wasn't also wired in. In that case:

- `computeNeedsRemotePrefill` (prefix_based_pd_decider line 239-241) hits `endpoint.Get(attrprefix.PrefixCacheMatchInfoDataKey.String()`), finds nothing, returns an error.

- `PreRequest` (line 151-153) catches that error and forwards the request anyway, same outcome (forward, no 412) as the "not configured at all" case, but reached via a completely different path, and not documented anywhere as expected behavior.
The problem that now (computeNeedsRemotePrefill, lines 222-246) all of those return an error, and PreRequest (lines 151-153) and disaggregate (lines 182-184) both forward on **any** error.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -102,75 +125,143 @@ func (d *PrefixBasedPDDecider) WithName(name string) *PrefixBasedPDDecider {
...
-	return true
+	debugLogger.Info("Computed non-cached suffix",
+		"hitPrefixTokens", hitPrefixTokens, "inputTokens", inputTokens,
+		"nonCachedTokens", nonCachedTokens, "threshold", d.config.NonCachedTokens)
+	return nonCachedTokens >= d.config.NonCachedTokens, nil
 }
 
-// getUserInputLenInTokens returns an estimated token count for the user input.
+// getUserInputLenInTokens returns the token count carried on the request.
+// A missing TokenizedPrompt is reported as an error so the conditional-decode
```
roytman: Actually, it is the opposite

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go
```diff
@@ -102,75 +125,143 @@ func (d *PrefixBasedPDDecider) WithName(name string) *PrefixBasedPDDecider {
...
 	logger := log.FromContext(ctx)
+	debugLogger := logger.V(logging.DEBUG)
+	endpoint := primaryDecodeEndpoint(schedulingResult)
+	if endpoint == nil {
+		debugLogger.Info("conditional-decode: no primary decode endpoint, rejecting")
+		return errCondDecodeCacheMiss
+	}
+	needs, err := d.needsRemotePrefill(ctx, request, endpoint)
+	if err != nil {
+		logger.Info("conditional-decode: error evaluating gate, forwarding", "error", err.Error())
```
roytman: Here, at line 183, I think the repository convention is to use logger.Error(err, ...)

roytman:
we can prevent open gateway when the plugin is not installed by something like:
```go
# PrefixBasedPDDecider.PreRequest marks that it evaluated the header, before its early NonCachedTokens == 0 return:
func (d *PrefixBasedPDDecider) PreRequest(ctx context.Context, request *scheduling.InferenceRequest,
	schedulingResult *scheduling.SchedulingResult) error {
	if request == nil || !routing.IsConditionalDecode(request.Headers) {
		return nil
	}
	request.PutAttribute(routing.ConditionalDecodeHandledAttributeKey, true)
		return nil
	}
	... // unchanged
}
# Director.prepareRequest (director.go), right after the existing runPreRequestPlugins block, adds a default-deny check:
if routing.IsConditionalDecode(reqCtx.Request.Headers) {
	if _, handled := reqCtx.SchedulingRequest.GetAttribute(routing.ConditionalDecodeHandledAttributeKey); !handled {
		return reqCtx, errcommon.Error{
			Code: errcommon.PreconditionFailed,
			Msg:  "conditional-decode request received but no gate plugin is configured",
		}
	}
}

File: pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go (hunk 13 above)
Comment (dmitripikus) [reply pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go]: One request == nil check at the top of needsRemotePrefill returning (false, error), and the two guards around the cache read/write are gone. ReadRequestAttribute/PutAttribute are unreachable with nil now. Thanks!

**Comment by dmitripikus** [comment]
Regarding the cases when:
-  prefix-based-pd-decider isn't configured at all in the configuration;
- plugin declared, but it's dependencies are missing.

Both scenarios addressed:

Scenario A (plugin absent): PrefixBasedPDDecider.PreRequest marks routing.ConditionalDecodeHandledAttributeKey = true as soon as it sees the header, and director.prepareRequest rejects any conditional-decode request whose attribute stayed unset with 412. A forgotten gate plugin now surfaces as the coordinator's cache-miss fallback rather than a silent forward.

Scenario B (plugin declared, producer missing): PrefixBasedPDDecider declares its Consumes() with PrefixCacheMatchInfoDataKey and TokenizedPromptDataKey as Required. The framework's dependency graph resolves them at init — either auto-creates the default producers (approx-prefix-cache-producer, tokenizer) or fails startup with ErrNoDefaultProducer. No more silent per-request forward on misconfiguration.

Docs and tests updated for both scenarios. 

Thanks!

File: pkg/epp/requestcontrol/director.go
Comment (dmitripikus) [reply pkg/epp/requestcontrol/director.go:641]: Both items mentioned are intentional:

Side effects on 412'd requests: runPreRequestPlugins doesn't short-circuit by design — every plugin runs. PreRequest plugins with side effects must be safe to run twice regardless of our gate.

Two failures collapse to 500: prepareRequest only preserves a typed code when there's exactly one unwrapped error. Two simultaneous PreRequest failures are rare and usually indicate a bigger problem.

File: pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go (hunk 14 above)
Comment (dmitripikus) [reply pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go]: Yes, you're right — that doc comment was stale from before we aligned PreRequest with disaggregate. 
Fixed. Thanks!

File: pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go (hunk 15 above)
Comment (dmitripikus) [reply pkg/epp/framework/plugins/scheduling/profilehandler/disagg/prefix_based_pd_decider.go]: Switched both disaggregate and PreRequest's soft-fail branches to logger.Error(err, msg).
Thanks!

File: docs/disaggregation.md [hunk 16]
```diff
@@ -396,7 +396,12 @@ PD deciders determine whether prefill should be offloaded to a separate worker,
 
 #### Prefix-Based PD Decider
 
-The `prefix-based-pd-decider` plugin makes the disaggregation decision according to the length of the non-cached suffix of the prompt relative to tokens already cached on the selected decode pod.
+The `prefix-based-pd-decider` plugin compares the request's non-cached suffix on the selected decode endpoint against a threshold. It fills two roles, sharing the same `nonCachedTokens` / `promptTokens` parameters:
+
+- Disaggregation decider inside `disagg-profile-handler` — routes prefill remotely when the threshold is met (see [Profile Handler Configuration](#profile-handler-configuration)).
+- Conditional-decode gate — rejects `Prefer: if-available` requests with HTTP 412 when the same threshold would trigger remote prefill.
+
+A deployment can declare the plugin for either role independently, or both.
```
roytman: Theoretically, this statement is right, but EPP configuration is different when we work with Sidecar or Coordinator. Sidecar uses the `disagg-profile-handler` and never sends conditional-decode, Coordinator uses default handlers, so it doesn't use `disagg-profile-handler` with `prefix-based-pd-decider`. 
Therefore, the statement can be something like: depending on the existing deployment "Sidecar or Coordinator" the plugin can be used in one way or another.
Furthermore, if the user defines them both with probably different parameters, the caching decision will prevent the second evaluation.  It should be mentiond

docs/disaggregation.md
```diff
@@ -417,11 +422,36 @@ The `prefix-based-pd-decider` plugin makes the disaggregation decision according
...
+- `promptTokens`: Minimum prompt length in tokens before the plugin's routing and gating logic applies. Prompts shorter than this run locally on the decode worker without remote prefill; the 412 gate honors the same shortcut. `0` disables it.
+
+**Conditional-decode 412 gate**
+
+Requests carrying `Prefer: if-available` (used by the coordinator's speculative early-decode step, see [coordinator_architecture.md](coordinator_architecture.md)) are gated by the plugin using the same `promptTokens` / `nonCachedTokens` thresholds as the disaggregation decision: when the chosen decode endpoint's non-cached suffix would trigger remote prefill, the plugin returns HTTP 412 Precondition Failed so the coordinator restarts the pipeline at encode/prefill/decode. Cache state is read as unweighted contiguous blocks, so a RAM-cached prefix contributes its full token count.
+
+Deployments that do not declare any conditional-decode gate plugin still reject `Prefer: if-available` requests: after the PreRequest phase, the director rejects unclaimed conditional-decode requests with 412 by default so a missing gate plugin surfaces as the coordinator's cache-miss fallback rather than a silent forward.
```
roytman: Too many details, I'd remove "the PreRequest phase"

pkg/epp/requestcontrol/director.go
```diff
@@ -522,6 +458,18 @@ func (d *Director) prepareRequest(ctx context.Context, reqCtx *handlers.RequestC
 		return reqCtx, errcommon.Error{Code: errcommon.Internal, Msg: err.Error()}
 	}
 
+	// Default-deny for "Prefer: if-available" when no PreRequest plugin
+	// claimed the header. Ensures a missing gate plugin surfaces as a 412 so
+	// the coordinator's cache-miss fallback runs, instead of a silent forward.
+	if routing.IsConditionalDecode(reqCtx.SchedulingRequest.Headers) {
+		if _, handled := reqCtx.SchedulingRequest.GetAttribute(fwkrc.ConditionalDecodeHandledAttributeKey); !handled {
```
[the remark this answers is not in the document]
dmitripikus: The description had drifted from the code. Updated both the "What this PR does" bullet and the release note to describe the two-layer behavior: the plugin runs the threshold check when declared, and the director-side default-deny rejects any unclaimed Prefer: if-available request with 412 when no gate plugin is configured. Both now explicitly mention the 412 response body message so operators know what to look for. 
Thanks!

docs/disaggregation.md
```diff
@@ -417,11 +422,36 @@ The `prefix-based-pd-decider` plugin makes the disaggregation decision according
...
+
+The gate runs from the plugin's `PreRequest` hook, so it is independent of any `disagg-profile-handler` wiring. A deployment that wants the coordinator's cache-miss fallback but no disaggregation can declare the plugin at the top level and skip the profile handler. The plugin declares its data dependencies (`PrefixCacheMatchInfo` and `TokenizedPrompt`) as required, so a missing producer surfaces as a startup error rather than a silent per-request forward:
```
roytman: Sorry, I don't understand what you want to say here.

docs/communication.md
```diff
@@ -361,6 +361,8 @@ The coordinator attempts an early decode immediately after rendering. This allow
 
 The coordinator adds the `Prefer: if-available` HTTP header to signal that the decode worker should only proceed if the KV cache is already available. If it responds with 412 Precondition Failed, the pipeline continues as normal.
 
+The 412 is enforced by the `prefix-based-pd-decider` plugin's conditional-decode gate; see [disaggregation.md](disaggregation.md#prefix-based-pd-decider) for configuration.
```
roytman: you can use something like 
"The prefix-based-pd-decider plugin enforces the 412 when configured; deployments without a claiming gate plugin still get 412 by default (see [disaggregation.md](disaggregation.md))."

roytman (commented):
Thank you for all the fixes. 
Several small comments.
Please update the PR description + release-note block — stale "opt-in, always forward" claim, now false; will ship a wrong changelog entry as-is.

docs/communication.md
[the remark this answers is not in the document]
dmitripikus: coordinator_architecture.md had already been updated to describe both paths; communication.md was missed. Fixed in 711ae34b with matching wording so operators looking at either doc get the full picture (plugin-configured gate first, director-side default-deny as the fallback). Thanks!

dmitripikus:
@roytman , regarding your comment:
"Please update the PR description + release-note block — stale "opt-in, always forward" claim, now false; will ship a wrong changelog entry as-is."

I've updated - see details in my answer to relevant Copilot comment

docs/disaggregation.md
dmitripikus: Rewrote the paragraph in c9f0abe0 to frame the two roles by deployment topology (sidecar-based → decider inside disagg-profile-handler; coordinator-based → standalone PreRequest gate) rather than as arbitrary options. 
Thanks!

docs/disaggregation.md
dmitripikus: Trimmed the "after the PreRequest phase" clause.
Thanks!

docs/disaggregation.md
dmitripikus: Trimmed the confusing preamble. The first two sentences were mechanism detail and a repeat of the topology framing just above. The one sentence I kept out of the old paragraph — the one about dependency validation — moved to a one-liner after the YAML block.

Thanks!

docs/communication.md
dmitripikus: The semantic content was already there (from an earlier Copilot feedback), but your phrasing is much tighter, so I updated. 

Thanks!

roytman (dismissed):
Thank you @dmitripikus

pkg/epp/requestcontrol/director.go
```diff
@@ -536,6 +472,18 @@ func (d *Director) prepareRequest(ctx context.Context, reqCtx *handlers.RequestC
 		return reqCtx, errcommon.Error{Code: errcommon.Internal, Msg: err.Error()}
 	}
 
+	// Default-deny for "Prefer: if-available" when no PreRequest plugin
```
elevran, line 475: #1686 explicitly asked for a no-op gate when no decider is configured ("forward unconditionally... the gate becomes opt-in"). This PR instead implements a default-deny (412 when no plugin claims the header), which is a deliberate reversal of that stated design, justified in the PR body as safer (avoids silently forwarding a speculative decode to a decoder that may not hold the cache). I think default-deny is the right call.

elevran (approved):

Issue #2315, opened 2026-08-06
Evaluate picking P and D nodes in single profile


elevran:
**What would you like to be added**:

Currently, when doing disaggregated P/D, the D node is selected first and if prefill is needed, a P node is added.
With #2299, we attempt to co-locate the two nodes, but selection still operates on P and D is separate phases.
Explore potential benefits of selecting P+D pair based on the combination. See https://github.com/llm-d/llm-d-router/pull/2299#discussion_r3731458268

**Why is this needed**:

Could potentially lead to more optimized P+D selection (e.g., D with lower coverage on the same node as best P might be better than a D node with higher coverage).

Results may be applicable to the topology aware Coordinator based P/D selecion as well (#2300)

ahg-g:
I think simply flipping the order such that we pick prefill first is going to be the most beneficial, and at the same time a simple change. I suspect that https://github.com/llm-d/llm-d-router/pull/2299 will not be useful without that as one would not want to restrict prefill picking at all, kv-cache hit ratio is the most important to optimize for.

nilig:
## Evaluation update

The topology-aware routing plugins work functionally, but we did not measure a
performance benefit on the tested deployments. Hard host-level topology
filtering increased mean latency because it narrowed the eligible endpoint
pool, while same-node and cross-node KV transfers had effectively similar
measured performance.

This evaluation covers the currently implemented sequential mechanisms: the
topology filter/scorer from #2299, which evaluates candidates against an
endpoint selected in the preceding phase, and the configurable prefill-first
order from #2342. We ran both stage orders across the baseline, filter, and
scorer policies.

### Test deployment

- Model: `openai/gpt-oss-120b`
- Model server: vLLM `v0.23.0`
- Router and sidecar: llm-d Router `v0.10.0`
- Two nodes with 8 H200 GPUs per node
- Per node: four TP1 prefill pods and one TP4 decode pod
- Total: eight prefill GPUs and eight decode GPUs
- NIXL configured with CUDA KV buffers and the UCX backend
- UCX transport diagnostics enabled

We evaluated:

- Baseline routing without topology affinity
- Strict `topology-affinity-filter`
- Soft `topology-affinity-scorer`
- Forced-local and forced-remote P/D selection
- 8K, 16K, 32K, and 64K input lengths
- Loads from 1 to 40 offered QPS
- Five randomized repetitions at 8K/256 and 40 QPS
[...]

PR #2340, opened 2026-08-09
re-ordered the scheduling stages in the disagg handler

ahg-g:
re-ordered the scheduling stages in the disagg handler such that prefiller is picked first, then encoder, then decoder

**What type of PR is this?**
/kind feature

With https://github.com/llm-d/llm-d-router/pull/2299 merging, picking the prefill last becomes suboptimal. The priority should be to pick the endpoint with the highest kv-cache match across domains/racks and then constrain the decoder to that rack.

Related to https://github.com/llm-d/llm-d-router/issues/2315

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/README.md
+++ b/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/README.md
@@ -30,9 +30,9 @@ Orchestrates up to three scheduling stages per request — decode (always), and 
 
 Runs each scheduling stage in sequence and assembles the final result from all stages that ran.
 
-1. Run the decode profile (always).
+1. If a prefill decider is configured and approves the request, run the prefill profile.
 2. If an encode decider is configured and approves the request, run the encode profile.
-3. If a prefill decider is configured and approves the request, run the prefill profile.
+3. Run the decode profile (always).
 4. Return the assembled scheduling result with decode as the primary profile.
 
 #### How It Works
--- a/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
+++ b/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
@@ -38,7 +38,7 @@ const (
 
 // PeerEndpointAttributeKey is the request-attribute key under which this
 // handler publishes the endpoint selected in an earlier scheduling phase
-// (the decode pick, before running the prefill profile), for plugins in a
+// (the prefill pick, before running the decode profile), for plugins in a
 // later profile to compare against (e.g. topology affinity between a
 // disaggregated prefill and decode pick). The value is an Endpoint.
 const PeerEndpointAttributeKey = "peer-endpoint"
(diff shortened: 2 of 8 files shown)


ahg-g:
This PR changes existing semantics, which is not good. To maintain backward compatibility https://github.com/llm-d/llm-d-router/pull/2342 makes this an opt-in choice.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
```diff
@@ -280,7 +280,7 @@ func newDisaggProfileHandler(handlerType, decodeProfile, prefillProfile, encodeP
 }
 
 // Pick implements scheduling.ProfileHandler.
-// Stages run in order: decode → encode (optional) → prefill (optional).
+// Stages run in order: prefill (optional) → encode (optional) → decode.
 // Returns the next profile to execute, or an empty map when all stages are done.
 func (h *Handler) Pick(ctx context.Context, request *scheduling.InferenceRequest, profiles map[string]scheduling.SchedulerProfile,
```
elevran, line 285: A few comments
- IIRC, you need to run encoding to get the object's token for prefill stage,
- running prefill before decode is changed user behavior and warrants a releae note
- for backward compat, worth considering making the order configurable via DisaggProfileHandler parameters so that decode first can still be selected (might need to be the default for compatibility or at least add a deprecation note after validating the performance gain from prefill first). The parameter can be simple bool (run prefill first), or an array of steps/profiles to avoid another full reorder in the code next time the trade-off changes.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/README.md
```diff
@@ -30,9 +30,9 @@ Orchestrates up to three scheduling stages per request — decode (always), and
 
 Runs each scheduling stage in sequence and assembles the final result from all stages that ran.
 
-1. Run the decode profile (always).
+1. If a prefill decider is configured and approves the request, run the prefill profile.
```
elevran, line 33: nit: the lead-in sentence above this list (line 27) still says "decode (always), and optionally encode and prefill," but this numbered list now runs prefill -> encode -> decode. Update the lead-in to match.

elevran (commented):
Reordering prefill-before-decode makes sense given #2299 (rack-affinity constraint needs the prefill pick before decode runs). Good catch fixing the nil-endpoint case in PrefixBasedPDDecider in the latest push - that was a real bug in the first version of this PR.

Two things before merge: the release note is marked NONE but this changes default routing/metrics behavior for every disagg deployment; and worth considering whether stage order should be configurable rather than hardcoded, since different deployments may want either (decode vs prefill heavy?).

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
ahg-g: I sent out https://github.com/llm-d/llm-d-router/pull/2342 which adds a parameter and preserves backward compatibility and adds a release note, I will close this PR

ahg-g: > IIRC, you need to run encoding to get the object's token for prefill stage,

This is only changing the picking order within the scheduler, not the execution order.

elevran: what I'm not sure of is if the prefill plugins need token information that is produced as part of the encoding profile's plugins. If that's the case (might not be... could be that they all run earlier to produce request attributes), then encode profile must run before the prefill profile.
It was meant as an open question - I don't know if that's the case.

Issue #1892, opened 2026-07-04
Cross-replica state synchronization for active-active EPP


vMaroon:
Active-active EPP partitions request-path-learned routing state (`approx-prefix-cache-producer` index, `inflight-load-producer` counters, `no-hit-lru-scorer` ordering): each replica only learns from requests it routed, so `docs/operations.md` warns against active-active with approximate prefix routing, and active-passive failover promotes an empty-state standby.

Proposal: a `state-sync` component replicating per-plugin state deltas between replicas over the KV-events ZMQ transport, with plugins opting in via a `ReplicatedState` interface and a snapshot bootstrap for replicas standing up. Reads stay in local memory. Design doc with mechanism, payloads, and alternatives: [[PUBLIC] EPP Cross-Replica State Synchronization](https://docs.google.com/document/d/1V8deqcHxOfqLyfOx9hVWdY1dOZotW-T8wKKZTs6exDE/edit).

Stages, each gated on the #1290 benchmark methodology:

- [ ] Generalize the KV-events transport (`pkg/events` extraction, publisher) — behavior-neutral, on top of #1862
- [ ] `state-sync` component + `approx-prefix-cache-producer` add replication
- [ ] `inflight-load-producer` per-origin gauges
- [ ] Snapshot bootstrap, `no-hit-lru-scorer` touches, speculative seeds

Depends on #1862.

elevran:
design aspects/concerns that need to be addressed:
- group membership (how do you know which peers to connect to and when they go away/replaced) - as far as I know ZMQ discovery is manual (or learnt from the InferencePool peers in our case - for k8s can use the EPP service EndpointSlice, unsure how to handle without k8s)
- publishing frequency and lag (notify on every event can be expensive on high load, do we send snapshot or delta?)
- handling of notification spikes / thundering herd (all replicas still receive pod XYZ is free at once and direct to it)
- stream establishment on startup has lag (and that's ok, I think)
- crash recovery behavior (do not treat state as persistent, clean up the snapshot values of failed EPPs).

The use of ZMQ and events feels like the wrong abstraction and layer to me. 
I would use bidirectional grpc (or TCP) to carry information between EPPs directly with EWMA on the current proxy and external observations (initialize 100% to local EPP observations until we get any external observations). I would still use some form of power-of-two choice on routing to avoid thundering herd.

ahg-g:
[...]

PR #2341, opened 2026-08-10
feat: EPP peer discovery infrastructure

evacchi:
**What type of PR is this?**

/kind feature

Introduces the base infrastructure for peer discovery: the plugin interface,
data types, in-memory store, config wiring, and documentation. No concrete
discovery mechanism is included; a Kubernetes-based implementation (EndpointSlice
reconciler, RBAC, Helm/kustomize) will follow in a separate PR once the
enablement and transport design questions from #1892 are settled.

- `PeerDiscovery` interface mirroring `EndpointDiscovery`: `Start`, `Ready`,
  selected via `dataLayer.peerDiscovery.pluginRef` in the EPP config.
- `PeerMetadata`, `PeerStore`, `PeerNotifier` types and the `NewPeerNotifier`
  adapter, following the same pattern as endpoint discovery.
- `MemoryPeerStore`: goroutine-safe, in-memory peer set with deterministic
  ordering.
- `PeerDiscoveryConfig` added to `DataLayerConfig` with `pluginRef`.
- `docs/peer-discovery.md` covering the interface, config, store, and how to
  write a plugin.

`PeerStore` is intentionally separate from `Datastore`: they share the
Upsert/Delete/notifier shape but serve different consumers (scheduler vs.
cross-replica syncer) and have different lifecycles (InferencePool vs. EPP
Deployment).

**Test plan**:

- [x] Interface contract: fake plugin drives PeerNotifier into a recording store
- [x] Full stack: fake plugin -> PeerNotifier -> MemoryPeerStore upsert and delete
- [x] MemoryPeerStore unit: upsert, replace, delete, no-op delete, deterministic ordering

Part of #1892.

```release-note
NONE
```


pkg/epp/server/options.go
```diff
@@ -78,6 +78,11 @@ type Options struct {
 	PoolNamespace string // Namespace of the InferencePool this Endpoint Picker is associated with.
 	PoolName      string // Name of the InferencePool this Endpoint Picker is associated with.
 	//
+	// Peer discovery (active-active state synchronization).
+	//
+	EnablePeerDiscovery bool   // Enables discovery of peer EPP replicas. Requires PeerServiceName.
```
zdtsw, line 83: do we need this flag, i mean isnt `PeerServiceName` enough to enable the feature

pkg/epp/controller/epp_peer_reconciler.go
```diff
@@ -0,0 +1,153 @@
...
+		For(&discoveryv1.EndpointSlice{}).
+		WithEventFilter(predicate.NewPredicateFuncs(forThisService)).
+		WithOptions(controller.Options{NeedLeaderElection: &needLeaderElection}).
+		Complete(r)
+}
+
+// endpointReady reports whether an endpoint is serving. A nil Ready condition is
+// treated as not ready.
+func endpointReady(ep *discoveryv1.Endpoint) bool {
+	return ep.Conditions.Ready != nil && *ep.Conditions.Ready
```
zdtsw, line 131: should not check `serving` instead?

pkg/epp/server/options.go
evacchi: yeah debatable, I added that because implicitly enabling when PeerServiceName is present seemed brittle, but I am not strongly opinionated

pkg/epp/controller/epp_peer_reconciler.go
evacchi: incorrect doc string, `Ready` is when it is serving _and not_ terminating, so I think `Ready` should be the appropriate one (you don't use a peer that's serving but about to terminate) -- updated

zdtsw: right, the one in termination should be skipped

cmd/epp/runner/runner.go
```diff
@@ -489,6 +497,10 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 		GRPCMaxSendMsgSize:               opts.GRPCMaxSendMsgSize,
 		EnableGRPCStreamMetrics:          opts.EnableGRPCStreamMetrics,
 		EmitEndpointScores:               opts.EmitEndpointScores,
+		PeerDiscovery: runserver.PeerDiscovery{
+			ServiceName: peerServiceName,
+			SelfAddress: os.Getenv("POD_IP"),
```
elevran: `SelfAddress` comes from `POD_IP`, but nothing in this PR's Helm chart or kustomize overlay sets a `POD_IP` env var on the EPP container. With it unset, `SelfAddress` is `""`, which never matches a real endpoint address, so a replica never excludes itself from its own peer set - silently, since nothing consumes the store yet to surface the bug. Needs a downward-API `POD_IP` env var added to the EPP pod spec (or a `Validate()` check that fails fast when peer discovery is enabled but `POD_IP` is empty).

pkg/epp/server/options.go
```diff
@@ -175,6 +180,11 @@ func (opts *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&opts.PoolNamespace, "pool-namespace", opts.PoolNamespace,
 		"Namespace of the InferencePool this Endpoint Picker is associated with.")
 	fs.StringVar(&opts.PoolName, "pool-name", opts.PoolName, "Name of the InferencePool this Endpoint Picker is associated with.")
+	fs.BoolVar(&opts.EnablePeerDiscovery, "enable-peer-discovery", opts.EnablePeerDiscovery,
```
elevran: See review body point 1 - this flag pair introduces a second enablement idiom alongside the existing plugin-config pattern for the same kind of feature.

pkg/epp/framework/interface/datalayer/peer.go
```diff
@@ -0,0 +1,74 @@
...
+	// Address is the IP address of the peer replica.
+	Address string
+	// Port is the port the peer listens on for state sync, as a string. Empty
+	// when the discovery source does not expose a port.
+	Port string
+}
+
+// PeerStore is the narrow interface required by NewPeerNotifier. Any store that
+// implements PeerUpsert and PeerDelete satisfies it.
+type PeerStore interface {
```
elevran, line 45: See review body point 3 - `PeerStore` duplicates `Datastore`'s existing Upsert/Delete/notifier shape for a different kind of object. Worth stating why this is separate rather than folding into `Datastore`.

pkg/epp/controller/epp_peer_reconciler_test.go
```diff
@@ -0,0 +1,192 @@
...
+		Endpoints:   eps,
+	}
+}
+
+func nn(name string) types.NamespacedName {
+	return types.NamespacedName{Namespace: testPeerNS, Name: name}
+}
+
+func TestEPPPeerReconciler(t *testing.T) {
+	scheme := runtime.NewScheme()
```
elevran: nit: `TestEPPPeerReconciler` covers add, no-op, and delete, but never exercises an in-place metadata change for an existing peer (the `existing != peer` branch that re-emits an `Upsert`) or a `nil` `Ready` condition (vs. explicit `false`). Worth adding both cases.

pkg/epp/server/options.go
elevran: Prefer seeing this a config file option, not CLI flags. See feedback on L183 (`fs.BoolVar(&opts.EnablePeerDiscovery, "enable-peer-discovery", ...)`)

elevran (commented):
Solid implementation of the mechanics (reconciler diffing, self-exclusion, RBAC gating, tests, envtest coverage) but I have design questions to discuss and resolve, plus one concrete bug.

1. **Enablement mechanism**: this mirrors `EndpointDiscovery`/`DiscoveryNotifier` in shape, but `docs/discovery.md` establishes that this class of feature (a discovery source feeding a datastore) is selected via a plugin entry in the EPP config (`dataLayer.discovery.pluginRef`), not a CLI flag. Presence of the plugin ref is what enables it - no separate `--enable-*` flag exists for endpoint discovery. This PR introduces `--enable-peer-discovery`/`--epp-peer-service` instead, a second, competing enablement idiom for the same shape of problem. Recommend a `PeerDiscovery` plugin interface (mirroring `EndpointDiscovery`) selected via config, both for consistency and because it would let peer discovery work under file-discovery mode too (today it's structurally unavailable there since it's wired directly into the controller-runtime manager setup).

2. **RBAC widening**: the EndpointSlice watch requires a new `discovery.k8s.io/endpointslices` get/list/watch grant plus an informer cache. Is there a lighter alternative? The EPP Service is currently a plain `ClusterIP` Service (`_service.yaml`), so a DNS SRV query against it today would only resolve to the single VIP, not individual replica addresses - getting per-pod records via DNS would require making the Service **headless**. That's a real option, but it changes how the same Service is resolved on the ext_proc traffic path too (`_helpers.tpl` shows Envoy uses `STRICT_DNS` against this Service in service mode; headless would shift load-balancing from kube-proxy to Envoy directly) - a second behavior change riding along with peer discovery, not a like-for-like swap. Worth deciding deliberately rather than defaulting to the EndpointSlice watch by inertia, and if EndpointSlice is kept, worth stating why explicitly (e.g., per-endpoint `Ready`/`Terminating` conditions and push vs. poll latency are real advantages this PR could call out).

3. **Peer storage**: `MemoryPeerStore` is structurally identical to the endpoint tracking the existing `Datastore` already does (same key type, same Upsert/Delete/notifier shape). What's the reasoning for a standalone store rather than extending `Datastore` with a second collection?

4. **Scope**: nothing in this PR consumes `PeerStore`/`PeerNotifier` - the `PeerDiscovery.Store` field comment itself calls it "the handoff point for a cross-replica syncer" that doesn't exist yet. Given (1) and (2) are still open design questions, splitting this into (a) the base interfaces/store with no watcher, and (b) the first concrete discovery mechanism once the enablement/transport questions are settled, would let each land on its own merits. It would also naturally produce a smaller diff (no RBAC/Helm/kustomize changes in the base-infra PR).

5. **Docs**: the analogous endpoint-discovery feature has a full doc (`docs/discovery.md`) covering the interface, config wiring, and a runnable example. This PR adds two CLI flags, an RBAC permission, a Helm value, and a Kustomize component with no equivalent documentation - worth adding before merge, once the enablement mechanism from point 1 is settled.

One concrete bug alongside the design questions - see inline comment on runner.go.

pkg/epp/server/options.go
evacchi: good point thanks

pkg/epp/framework/interface/datalayer/peer.go
evacchi: I'll extend the doc comment to explain it; essentially the lifecycle of `Datastore` vs `PeerStore` should be different. The `Datastore` is bound to the InferencePool while the `PeerStore` is used by the EPP deployment, adding Peer data in the Datastore felt out of place. The pattern was kept the same for design consistency.

pkg/epp/controller/epp_peer_reconciler_test.go
evacchi: dropping, PR will be rescoped to interface-only as suggested above

cmd/epp/runner/runner.go
evacchi: dropping this, rescoping to interface-only

evacchi:
> 1. **Enablement mechanism**: this mirrors `EndpointDiscovery`/`DiscoveryNotifier` in shape, but `docs/discovery.md` establishes that this class of feature (a discovery source feeding a datastore) is selected via a plugin entry in the EPP config (`dataLayer.discovery.pluginRef`), not a CLI flag. 

totally fair, apologies for the oversight -- I think I actually did this in the past already :D


> 3. **Peer storage**: `MemoryPeerStore` is structurally identical to the endpoint tracking the existing `Datastore` already does (same key type, same Upsert/Delete/notifier shape). What's the reasoning for a standalone store rather than extending `Datastore` with a second collection?

answered inline, essentially although they are similar in shape, the data store has a different lifecycle than the peerstore and it felt a bit out of place -- besides datastore contains a few fields already. Not particularly attached to this design decision and I'm happy to collapse them if that's better.

> 2. **RBAC widening** and **Scope**

As suggested, I'll reduce scope here to mostly interfaces.

> 5. **Docs**: the analogous endpoint-discovery feature has a full doc (`docs/discovery.md`)

adding docs

config/charts/routerlib/templates/_rbac.yaml
```diff
@@ -49,6 +49,9 @@ rules:
 - apiGroups: [""]
   resources: ["pods"]
   verbs: ["get", "watch", "list"]
+- apiGroups: ["discovery.k8s.io"]
+  resources: ["endpointslices"]
+  verbs: ["get", "watch", "list"]
```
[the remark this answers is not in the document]
evacchi: no longer applies

evacchi:
rescoped: only interfaces. Comments addressed. Rebased.

elevran (approved):
rescoped diff directly and completely addresses all prior review feedback; small, clean, matches existing discovery.go patterns closely.
Thanks!

elevran:
one thing worth discussing:
does it make sense to group the two discovery plugins in the config togther?
```yaml
discovery:
    - endoints:
       pluginRef: ....
    - peers:
       pluginRef: ....
```
Would require a backward compatibility shim for two releases but after would be consolidated.
Thoughts?
Can be a separate PR as there are no users for peers yet.

elevran:
@evacchi - merging, but would appreciate your thoughts on the config schema change.

evacchi:
oh yes that sounds good!

PR #2342, opened 2026-08-10
Added an option to disagg handler to allow running prefill first.

ahg-g:
**What type of PR is this?**
/kind feature

Added an option to disagg handler to allow running prefill first.

With https://github.com/llm-d/llm-d-router/pull/2299 merging, picking the prefill last becomes suboptimal. The priority should be to pick the endpoint with the highest kv-cache match across domains/racks and then constrain the decoder to that rack.

Related to https://github.com/llm-d/llm-d-router/issues/2315

```release-note
A new parameter named StageOrder is added to disagg-profile-handler plugin to allow selecting the order of picking the prefill and decode endpoints. Existing behavior was always picking decoder first, this parameter allows picking the prefill first.
```

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/always_disagg_mm_decider.go
+++ b/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/always_disagg_mm_decider.go
@@ -27,10 +27,15 @@ func AlwaysDisaggMulimodalDeciderPluginFactory(name string, _ *json.Decoder, _ p
 	return newAlwaysDisaggEncodeDecider().WithName(name), nil
 }
 
-func newAlwaysDisaggEncodeDecider() *AlwaysDisaggMultimodalDecider {
+// NewAlwaysDisaggEncodeDecider creates a new AlwaysDisaggMultimodalDecider.
+func NewAlwaysDisaggEncodeDecider() *AlwaysDisaggMultimodalDecider {
 	return &AlwaysDisaggMultimodalDecider{
 		typedName: plugin.TypedName{Type: AlwaysDisaggMulimodalPluginType},
 	}
+}
+
+func newAlwaysDisaggEncodeDecider() *AlwaysDisaggMultimodalDecider {
+	return NewAlwaysDisaggEncodeDecider()
 }
 
 // TypedName returns the typed name of the plugin.
--- a/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/always_disagg_mm_decider.go
+++ b/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/always_disagg_mm_decider.go
@@ -24,7 +24,7 @@ type AlwaysDisaggMultimodalDecider struct {
 // AlwaysDisaggMulimodalDeciderPluginFactory defines the factory function for creating
 // a new instance of the AlwaysDisaggEncodeDecider.
 func AlwaysDisaggMulimodalDeciderPluginFactory(name string, _ *json.Decoder, _ plugin.Handle) (plugin.Plugin, error) {
-	return newAlwaysDisaggEncodeDecider().WithName(name), nil
+	return NewAlwaysDisaggEncodeDecider().WithName(name), nil
 }
 
 // NewAlwaysDisaggEncodeDecider creates a new AlwaysDisaggMultimodalDecider.
@@ -32,10 +32,6 @@ func NewAlwaysDisaggEncodeDecider() *AlwaysDisaggMultimodalDecider {
 	return &AlwaysDisaggMultimodalDecider{
 		typedName: plugin.TypedName{Type: AlwaysDisaggMulimodalPluginType},
 	}
-}
-
-func newAlwaysDisaggEncodeDecider() *AlwaysDisaggMultimodalDecider {
-	return NewAlwaysDisaggEncodeDecider()
 }
 
 // TypedName returns the typed name of the plugin.
--- a/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/always_disagg_pd_decider.go
+++ b/pkg/epp/framework/plugins/scheduling/profilehandler/disagg/always_disagg_pd_decider.go
@@ -25,11 +25,10 @@ type AlwaysDisaggPDDecider struct {
 // a new instance of the AlwaysDisaggPDDecider.
 func AlwaysDisaggPDDeciderPluginFactory(name string, _ *json.Decoder,
 	_ plugin.Handle) (plugin.Plugin, error) {
-	return NewAlwaysDisaggPDDecider().WithName(name), nil
+	return newAlwaysDisaggPDDecider().WithName(name), nil
 }
 
-// NewAlwaysDisaggPDDecider creates a new AlwaysDisaggPDDecider.
-func NewAlwaysDisaggPDDecider() *AlwaysDisaggPDDecider {
+func newAlwaysDisaggPDDecider() *AlwaysDisaggPDDecider {
 	return &AlwaysDisaggPDDecider{
 		typedName: plugin.TypedName{Type: AlwaysDisaggPDDeciderPluginType},
 	}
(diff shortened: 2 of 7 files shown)


ahg-g:
See feedback at https://github.com/llm-d/llm-d-router/pull/2340#discussion_r3750119201

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
```diff
@@ -361,6 +419,63 @@ func (h *Handler) Pick(ctx context.Context, request *scheduling.InferenceRequest
...
+		decision := DisaggDecisionType(encodeUsed, prefillUsed)
+		RecordDisaggDecision(h.typedName.Name, h.typedName.Type, request.TargetModel, decision)
+		span.SetAttributes(attribute.String("llm_d.epp.profile_handler.decision", "complete_"+decision))
+		return map[string]scheduling.SchedulerProfile{}
+	}
+
+	// ── Stage 1: Prefill (optional) ────────────────────────────────────────
+	if _, hasPrefillProfile := profiles[h.prefillProfile]; hasPrefillProfile {
+		if _, executed := profileResults[h.prefillProfile]; !executed {
+			if h.pdDecider != nil && h.pdDecider.disaggregate(ctx, request, nil) {
```
elevran: I think passing `nil` to the decider breaks prefill-first with `PrefixBasedPDDecider`, which returns false (and logs an error) when the endpoint is nil. End result: prefill never runs in prefill-first mode for the most common decider. Either drop this gate in prefill-first mode (run prefill whenever the profile exists), or validate at config time that the chosen decider supports prefill-first.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
```diff
@@ -361,6 +419,63 @@ func (h *Handler) Pick(ctx context.Context, request *scheduling.InferenceRequest
 	return map[string]scheduling.SchedulerProfile{}
 }
 
+func (h *Handler) pickPrefillFirst(ctx context.Context, span trace.Span, request *scheduling.InferenceRequest,
+	profiles map[string]scheduling.SchedulerProfile, profileResults map[string]*scheduling.ProfileRunResult,
+) map[string]scheduling.SchedulerProfile {
+	// If decode has already run, we are done.
```
elevran, line 429: this completion path doesn't set `llm_d.epp.profile_handler.decode_failed=true` when decode ran and returned no targets. `pickDecodeFirst` sets it on the equivalent path (~L319). Add the same attribute here for telemetry parity.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/always_disagg_pd_decider.go
```diff
@@ -28,12 +28,17 @@ func AlwaysDisaggPDDeciderPluginFactory(name string, _ *json.Decoder,
 	return newAlwaysDisaggPDDecider().WithName(name), nil
 }
 
-func newAlwaysDisaggPDDecider() *AlwaysDisaggPDDecider {
+// NewAlwaysDisaggPDDecider creates a new AlwaysDisaggPDDecider.
+func NewAlwaysDisaggPDDecider() *AlwaysDisaggPDDecider {
```
elevran: nit: 
Drop the lowercase wrapper at L38, and change call sites to use `NewAlwaysDisaggPDDecider`.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/always_disagg_mm_decider.go
```diff
@@ -27,12 +27,17 @@ func AlwaysDisaggMulimodalDeciderPluginFactory(name string, _ *json.Decoder, _ p
 	return newAlwaysDisaggEncodeDecider().WithName(name), nil
 }
 
-func newAlwaysDisaggEncodeDecider() *AlwaysDisaggMultimodalDecider {
+// NewAlwaysDisaggEncodeDecider creates a new AlwaysDisaggMultimodalDecider.
```
elevran, line 30: nit: 
consolidate `newAlwaysDisaggEncodeDecider` code to `NewAlwaysDisaggEncodeDecider` and drop the lcase version?

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/scheduler_test.go
```diff
@@ -283,3 +283,78 @@ func TestPDSchedule(t *testing.T) {
 		})
 	}
 }
+
+func TestPDSchedule_PrefillFirst(t *testing.T) {
```
elevran, line 287: this test uses `NewAlwaysDisaggPDDecider` which always returns true and so hides the bug in `pickPrefillFirst`. Add a sibling test using `PrefixBasedPDDecider` in prefill-first mode so the broken path is exercised.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
```diff
@@ -36,11 +36,36 @@ const (
 	defaultEncodeProfile  = "encode"
 )
 
+// StageOrder defines the execution order of stages in the disaggregation profile handler.
+type StageOrder string
+
+const (
+	// StageOrderDecodeFirst runs Decode -> Encode -> Prefill (default).
+	// Decode runs first; PD decider inspects the picked decode endpoint to determine if prefill is needed.
```
elevran, line 44: nit: 
The order is relative (X before Y), but X is not necessarily the absolute **first**. Can be a little confusing for some future reader with less context. Maybe something along the lines of `AnchorDisaggregationOnPrefill`/`Decode`? Or is that too long, anchoring is also not immediately clear, etc.?

elevran (commented):
1. The PR description says prefill-first is for "picking the endpoint with the highest kv-cache match across domains/racks and then constrain the decoder to that rack." That intent can't be realized while the prefill gate is held by a decider that requires a decode endpoint to make its decision. Options:
- In prefill-first mode, always run prefill when the profile exists (drop the `pdDecider.disaggregate` call from Stage 1).
- Add a prefill-first-specific decider that scores prefill candidates directly.
- Document **explicitly** that prefill-first only works with deciders that don't depend on the decode endpoint (essentially just `always-disagg-pd-decider`), and reject `PrefixBasedPDDecider` at config validation time.

2. The `docs/disaggregation.md` file describes the disagg-profile-handler config but doesn't mention `stageOrder`. Add it next to the existing config examples.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
ahg-g: Great catch, I will drop the decider completely in the prefill first case.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
ahg-g: hmm, I think "first" is representative and short.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go
ahg-g: done.

pkg/epp/framework/plugins/scheduling/profilehandler/disagg/scheduler_test.go
ahg-g: Removed the decider for the prefill-first order.

ahg-g (commented):
### Summary of Changes

  1. Dropped PD Decider in Prefill-First Mode:
      • In disagg_profile_handler.go, Stage 1 runs prefill unconditionally whenever profiles[h.prefillProfile] is configured and not yet executed (removing the h.pdDecider.disaggregate(...) check).
  2. Added Decode-Failure Telemetry Parity:
      • In disagg_profile_handler.go, when decode was executed but failed or returned no target endpoints, the span attribute llm_d.epp.profile_handler.decode_failed=true is now set, matching pickDecodeFirst.
  3. Cleaned up Decider Constructors:
      • In always_disagg_pd_decider.go and always_disagg_mm_decider.go, consolidated constructors to always_disagg_pd_decider.go and always_disagg_mm_decider.go, dropping the redundant lowercase functions.
  4. Updated Documentation:
      • Documented the stageOrder parameter ("decode-first" / "prefill-first") in disaggregation.md along with configuration examples.
  5. Updated Unit Tests:
      • Updated tests in disagg_profile_handler_test.go, pd_profile_handler_test.go, and scheduler_test.go to reflect that prefill-first execution operates without a PD decider.

elevran (approved):

PR #2350, opened 2026-08-11
test(kvcache): release CostAware test resources

### LINKED ISSUE #2349: CostAware KV-block tests retain Ristretto resources across subtests
# CostAware KV-block tests retain Ristretto resources across subtests

/kind bug

**What happened**:

`TestCostAwareIndexBehavior` creates a fresh `CostAwareMemoryIndex` for each
common index-behavior subtest. The default CostAware configuration creates a
Ristretto cache with `NumCounters=1e8`, which allocates substantial admission
metadata and starts background processing. The test process did not close each
instance after its subtest completed, so unused cache resources accumulated.

The complete `pkg/kvcache/...` suite can exceed the memory limit in constrained
Docker or CI environments and terminate with an OOM kill.

Each CostAware test instance should release its Ristretto resources after the
subtest finishes. The complete KV-cache test suite should run without memory
growth proportional to the number of CostAware subtests.

1. Run the KV-cache tests with the CostAware behavior suite using the default
   CostAware configuration.
2. Observe that each behavior subtest creates another Ristretto instance with
   `NumCounters=1e8`.
3. Run the complete suite in a memory-constrained Docker or CI environment:

   ```bash
   go test -count=1 -v ./pkg/kvcache/...
   ```

4. The process can be OOM-killed after CostAware resources accumulate.

The fix is limited to test resource lifecycle management:

- expose a close method on `CostAwareMemoryIndex` that releases its local
  Ristretto, LRU, and auxiliary key-index resources;
- register `t.Cleanup` in CostAware test construction;
- use a small Ristretto counter count in tests while preserving the production
  default of `1e8`.

The fix does not change the `Index` interface, production index lifecycle,
Redis behavior, or Redis/miniredis timeout handling.

**Environment**:

- llm-d-router version: current main-based development checkout
- Install tools: Go test, Docker or CI runner with an enforced memory limit
- Others: failure reproduces when the test process retains multiple default
  CostAware Ristretto instances

Alex-ai-future:
**What type of PR is this?**

/kind test

CostAware KV-block tests create a fresh Ristretto-backed index for each common
behavior subtest. Default-sized Ristretto instances retain their background
resources until they are closed, causing memory to accumulate across the test
process in memory-constrained Docker and CI environments.

This PR adds deterministic cleanup for CostAware test instances, uses a small
Ristretto counter count in tests, and verifies repeated close behavior. The
production `NumCounters` default remains unchanged.

Fixes #2349

```release-note
NONE
```

**Test plan**:

- [x] Verify CostAware lifecycle cleanup, behavior, size, and key-index
  eviction tests.
- [x] Run `go test -count=1 ./pkg/kvcache/...`.
- [x] Run `go test -count=5 ./pkg/kvcache/kvblock`.


pkg/kvcache/kvblock/cost_aware_memory.go
```diff
@@ -156,6 +164,16 @@ func (m *CostAwareMemoryIndex) MaxCost() int64 {
 	return m.data.MaxCost()
 }
 
+// Close releases the index's cache resources. It is safe to call multiple times.
```
vMaroon: `Close()` lands on the concrete `*CostAwareMemoryIndex`, but nothing can reach it in production: `Indexer` holds a `kvblock.Index` (wrapped again by `NewTracedIndex`), and its `Run` just blocks on `ctx.Done()` with no shutdown that calls `Close`. So the Ristretto release only runs in tests that hold the concrete type. `RedisIndex` has the same problem from the other direction — it owns a `redis.NewClient` (`redis.go:62`) that also leaks with no `Close`. 

I think the clean fix is `Close() error` on the `Index` interface: close the client in Redis, no-op in `InMemoryIndex`, and delegate in the `NewTracedIndex`/`NewInstrumentedIndex` wrappers, then call it from an `Indexer` shutdown hook. If lifecycle is out of scope for this PR, let's drop the concrete `Close` or mark it as a follow-up.

Alex-ai-future: Thanks, agreed. I narrowed this PR to the test-only fix: it keeps the bounded NumCounters configuration for CostAwareMemoryIndex tests and removes the concrete Close() API. 

Production index lifecycle management needs a coordinated change across the Index interface, its implementations and wrappers, and Indexer shutdown. I plan to follow up with that separately.

ahg-g (approved):

PR #2352, opened 2026-08-11
feat(datalayer): typed attribute slots and value-safety registry

### LINKED ISSUE #2338: Enforce declared types when writing to request/endpoint attribute stores
**What would you like to be added**

Runtime validation that the value passed to `AttributeMap.Put` and `InferenceRequest.PutAttribute` matches the type its producer declared via `Produces()`. Build a `producerName -> DataKey -> reflect.Type` registry once at startup from registered `Produces()` calls and consult it at write time. Mismatched types fail fast instead of silently misreading at the consumer.

**Why is this needed**

The endpoint attribute store (`pkg/epp/framework/interface/datalayer/attributemap.go`) is `map[string]Cloneable` and the request store (`pkg/epp/framework/interface/scheduling/attributes.go`) is `sync.Map` of `any`. Both keys are `dk.String()` (`"<dataType>/<producerName>"`) so producer namespaces already separate, but nothing checks the value's type at write time. Today the only type safety is declared-vs-declared in the DAG build (`pkg/epp/datalayer/data_graph.go:229`), which does not catch a producer that lies at `Put` time.

Real-world signal: `pkg/epp/framework/plugins/scheduling/util/sessionaffinity/session_id.go:196` reads attributes via `reflect.ValueOf(v).Kind() == reflect.String` because plain `string` assertions miss the named `SessionID` type that producers actually store. The workaround exists because there is nowhere else for the consumer to find out the declared type.

Secondary problem in scope: `pkg/epp/framework/plugins/scheduling/filter/endpointattribute/filter.go:154` still uses the legacy `Consumes() map[string]any` signature, which the DAG validator (`data_graph.go:81-87`) does not walk. That filter is invisible to init-time type checking. Migrating it to `DataDependencies` is part of the same fix.

**Proposed approach (minimal)**

1. Walk registered `ProducerPlugin.Produces()` once at startup and stash `producerName -> DataKey -> reflect.Type` on the framework handle.
2. Add `Put(dk DataKey, val Cloneable)` on `AttributeMap` and `PutAttribute(dk DataKey, val any)` on `InferenceRequest`. Both look up the declared type and reject mismatch.
3. Keep the string-keyed methods as deprecated back-compat shims that call the typed version with `reflect.TypeOf(nil)` (no registry entry, accepted) or warn at write time.
4. Migrate the 8-10 known call sites that already have `dk` in scope at the Put site (e.g. `topology/extractor.go:221`, `inflightload/producer.go:343`, `models/extractor.go:56`, `dcgm/extractor.go:106`, `sessionid/producer.go`).
5. Migrate `endpointattribute/filter.go` to `DataDependencies`.

**Performance note**

If `Put` lands on a hot path (e.g. endpoint attributes re-written on each metric scrape), consider per-key typed slots (a generic `Slot[T]` constructed once per producer) so the type check at write time is an interface assertion rather than `reflect.TypeOf` per call. The declared type still comes from `Produces()` for DAG validation; reflect lives there once at startup.

**Out of scope (add when someone asks)**

- Generated typed accessors per key.
- A central schema registry separate from `Produces()`.
- Read-only accessors or compile-time enforcement (Go reflection is the ceiling here).

elevran:
**What type of PR is this?**
/kind cleanup
/kind feature

Adds **typed attribute slots** and a startup attribute registry to catch value-type mismatches at write time and at framework init, complementing PR #2190's key-safety check.

- `Slot[T Cloneable]` pins the value type at compile time. A producer that builds `Slot[*Foo]` but holds a `*Bar` value fails at the assignment boundary, not at some downstream reader.
- A `Registry` built from `ProducerPlugin.Produces()` at framework startup. `ValidateConsumer` rejects a plugin whose `Consumes()` references a key the registry does not know, or names a value type that does not match the producer's declaration. Both halves of the value-safety check.

Note that some consumers accept a polymorhic string like object from any source so need to test for string like objects and can't use a Slot[string] (eg. `session.SessionID` must validate `reflect.ValueOf(v).Kind() == reflect.String` since it is a new type, not a Go string).

Fixes #2338
Complements #2190.
| | PR 2190 | This PR |
|---|---|---|
| Constrains | WHO may claim a key | WHAT name + value type the producer/consumer agree on |
| Failure surface | startup, registry walk | startup registry walk + per-write type assertion |

Neither replaces the other. Both are needed: 2190 keeps a plugin from claiming a key it does not own; this keeps it from writing a value of the wrong type to the key it does own.

```release-note
NONE
```
This is **not** a user facing change

**Additional notes for reviewer:**
Migrated seven Put sites to slots (topology, models, dcgm, inflightload uncached tokens, sessionid request side) and switched the endpoint-attribute filter's legacy `Consumes() map[string]any` to `plugin.DataDependencies` so
the DAG and registry actually validate it.

Five legacy `Consumes() map[string]any` scorers (kvcacheutilization, queuedepth, runningrequests, loraaffinity, endpointattribute scorer) are out of scope here; they do not implement `plugin.ConsumerPlugin`, so the registry does not
see them today. They will be migrated in a follow-up PR #2355.

**Test plan:**
- [x] `go test ./pkg/epp/framework/interface/datalayer/...` (slot and registry unit tests)
- [x] `go test ./pkg/epp/datalayer/...` (registry integration through `ValidateAndOrderDataDependencies`, including unknown-key and type-mismatch cases)
- [x] `go test ./pkg/epp/framework/plugins/scheduling/util/sessionaffinity/...` (the consumer test now publishes `SessionID` rather than plain string, asserting the typed accessor works)

### DIFF
--- a/pkg/epp/framework/interface/datalayer/slot.go
+++ b/pkg/epp/framework/interface/datalayer/slot.go
@@ -1,0 +1,116 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package datalayer
+
+import (
+	"fmt"
+	"reflect"
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+)
+
+// attributeWriter is the minimal contract Slot.Put needs: anything that can
+// store a Cloneable value under a DataKey. Both datalayer.AttributeMap
+// and scheduling.Endpoint satisfy it, so Slot works on both stores without
+// a GetAttributes indirection.
+type attributeWriter interface {
+	Put(plugin.DataKey, Cloneable)
+}
+
+// Slot is a typed write handle for a single DataKey. The declared value type
+// is captured once at construction, so Put writes through a single interface
+// assertion. No reflect on the hot path.
+//
+// One producer owns one Slot per DataKey it writes. NewSlot[T] pins the type
+// at compile time; a producer that builds Slot[*Foo] but holds a *Bar value
+// fails at the assignment boundary, not at some downstream reader.
+//
+// Slot is the value-safety complement to PR 2190's key-safety check: that PR
+// keeps a producer from claiming a key it does not own, this keeps it from
+// writing a value of the wrong type to the key it does own.
+type Slot[T Cloneable] struct {
+	dk        plugin.DataKey
+	valueType reflect.Type
+}
+
+// NewSlot returns a typed Slot bound to dk. The reflect.Type is captured here
+// (once, at construction) for the framework startup validator; Put uses a
+// direct type assertion and never reflects the value.
+func NewSlot[T Cloneable](dk plugin.DataKey) *Slot[T] {
+	return &Slot[T]{dk: dk, valueType: reflect.TypeOf((*T)(nil)).Elem()}
+}
+
+// DataKey returns the key this slot writes to. Useful for Produces()
+// declarations and log messages.
+func (s *Slot[T]) DataKey() plugin.DataKey {
+	return s.dk
+}
+
+// Type returns the declared reflect.Type for this slot. Used by the framework
+// startup validator when building the attribute registry from a producer's
+// Produces() map.
+func (s *Slot[T]) Type() reflect.Type {
+	return s.valueType
+}
+
+// Put writes val to w under this slot's key. A typed-nil pointer, interface,
+// map, slice, channel, or func val is dropped (the underlying store's nil
+// check does not catch typed nils because the interface still carries the
+// type); for value-type Ts there is no nil state and the value is stored
+// unconditionally.
+//
+// The kind switch keeps reflect off the hot path for the common value-type
+// case; the reflect call only fires for Ts where nil is possible, and the
+// cost is a single IsNil probe.
+func (s *Slot[T]) Put(w attributeWriter, val T) bool {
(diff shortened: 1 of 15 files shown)


pkg/epp/framework/interface/datalayer/slot.go
```diff
@@ -0,0 +1,116 @@
...
+
+// attributeWriter is the minimal contract Slot.Put needs: anything that can
+// store a Cloneable value under a DataKey. Both datalayer.AttributeMap
+// and scheduling.Endpoint satisfy it, so Slot works on both stores without
+// a GetAttributes indirection.
+type attributeWriter interface {
+	Put(plugin.DataKey, Cloneable)
+}
+
+// Slot is a typed write handle for a single DataKey. The declared value type
```
ahg-g, line 36: how do we force all places in the code to use slot instead of the existing approach to putting attribute values?

elevran: > how do we force all places in the code to use slot instead of the existing approach to putting attribute values?

Today, `Put(dk, val)` on `AttributeMap` is still public, so the type check on `Slot[T]` is indeed advisory. 
Due to the polymorphic cases the PR already documents (e.g., `SessionID` is `reflect.String`-kind but is not assignable to `string` and can't be expressed as `Slot[string]`) we can't force it. The other option is to disallow the polymorphic attributes.

We can make the following changes to **streer** callers to go through slots:

- attributemap.go: rename the `Put(dk, val)` body to unexported `putRaw`; remove the public method. `ReadAttribute[T]` stays as-is.
- datalayer/slot.go: add `RawSlot` whose `Put(m, val any)` calls `putRaw` and doc it: "use only when the value type cannot be pinned at compile time (polymorphic string-kind values, request-side any-backed store). Prefer Slot[T]."
- Migrate the sessionid request-side producer to `RawSlot` with a godoc line pointing at the reflect-string-kind reader.

If you think it's not worth it without forcing the typed slots, I'm ok with dropping it. I think it is a step in the right direction but leaves a loop hole still open.

ahg-g: it is a step in the right direction, I think it is worth following up on to tighten the api to force all consumers to use the new interface.

elevran: Filed #2779 to track tightening the API to force writes through Slot[T]. Resolving this thread as agreed - the current PR stands as the value-safety complement to #2190, and the API tightening is scoped out to the follow-up issue since it touches every existing Put call site, not just sessionid.

ahg-g (approved):

PR #2357, opened 2026-08-11
feat(coordinator): proxy unregistered paths to the decode profile

dmitripikus:
**What type of PR is this?**
/kind feature

Callers hitting the coordinator on a path it does not register today (e.g. `/v1/models`, `/v1/messages`, `/v1/responses`, `/v1/embeddings`, `/tokenize`) get a stock chi 404 and have to bypass the coordinator to reach the equivalent vLLM endpoint.

This PR adds a chi `NotFound` handler that reverse-proxies any unregistered path to the Inference Gateway with `EPP-Profile: decode`, so EPP dispatches the request to a decode pod (a full vLLM instance). Method, body, query, and forwarded headers pass through unchanged; `X-Request-Id` is validated and replaced with a UUID if malformed, matching `handleInference`'s sanitization. Registered paths still hit the pipeline; method-mismatches on registered paths still return 405.

Mirrors the sidecar's existing `mux.Handle("/", decoderProxy)` fallback pattern.

Fixes #

**Release note**:
```release-note
Coordinator now forwards requests for unregistered paths (e.g. /v1/models, /v1/messages, /v1/responses, /v1/embeddings) directly to a decode pod via EPP-Profile: decode, so clients no longer need to bypass the coordinator to reach these vLLM endpoints.
Enforce server.max_request_body_size on passthrough (unregistered) paths; oversize requests now return 413.
```

### DIFF
--- a/pkg/coordinator/server/handlers_test.go
+++ b/pkg/coordinator/server/handlers_test.go
@@ -44,9 +44,20 @@ func (s stubStep) Name() string { return s.name }
 
 func (s stubStep) Execute(context.Context, *pipeline.RequestContext) error { return s.err }
 
+// stubGatewayURL is a placeholder used by tests that never actually issue a
+// passthrough request. A real value only matters in passthrough_test.go.
+const stubGatewayURL = "http://gateway-stub.invalid"
+
 func newTestServer(stepErr error) *Server {
+	return newTestServerWithGateway(stepErr, stubGatewayURL)
+}
+
+func newTestServerWithGateway(stepErr error, gatewayURL string) *Server {
 	p := pipeline.New([]pipeline.Step{stubStep{name: "stub", err: stepErr}})
-	srv, err := New(config.ServerConfig{}, p)
+	// A default *http.Transport{} is safe here: the passthrough handler needs a
+	// non-typed-nil RoundTripper so httputil.ReverseProxy dials the test server.
+	gw := gateway.NewWithTransport(&http.Transport{}, gatewayURL)
+	srv, err := New(config.ServerConfig{}, p, gw)
 	if err != nil {
 		panic(err) // default config is always valid
 	}
@@ -134,7 +145,7 @@ func TestHandleInference_BodyOverConfiguredCapMapsTo413(t *testing.T) {
 	// A body larger than server.max_request_body_size (in MB) is rejected before parsing.
 	// Use a 1 MB cap and send 1 MB + 1 byte to trigger the limit.
 	p := pipeline.New([]pipeline.Step{stubStep{name: "stub"}})
-	srv, err := New(config.ServerConfig{MaxRequestBodySize: 1}, p)
+	srv, err := New(config.ServerConfig{MaxRequestBodySize: 1}, p, gateway.NewWithTransport(nil, stubGatewayURL))
 	if err != nil {
 		t.Fatalf("New: %v", err)
 	}
@@ -149,7 +160,7 @@ func TestHandleInference_BodyOverConfiguredCapMapsTo413(t *testing.T) {
 
 func TestNew_RejectsNegativeMaxRequestBodySize(t *testing.T) {
 	p := pipeline.New([]pipeline.Step{stubStep{name: "stub"}})
-	if _, err := New(config.ServerConfig{MaxRequestBodySize: -1}, p); err == nil {
+	if _, err := New(config.ServerConfig{MaxRequestBodySize: -1}, p, gateway.NewWithTransport(nil, stubGatewayURL)); err == nil {
 		t.Fatal("expected error for negative MaxRequestBodySize")
 	}
 }
@@ -158,8 +169,22 @@ func TestNew_RejectsOverflowMaxRequestBodySize(t *testing.T) {
 	// MaxInt64 would cause maxRequestBodySize+1 to overflow to a negative
 	// io.LimitReader limit, making it return immediate EOF.
 	p := pipeline.New([]pipeline.Step{stubStep{name: "stub"}})
-	if _, err := New(config.ServerConfig{MaxRequestBodySize: math.MaxInt64}, p); err == nil {
+	if _, err := New(config.ServerConfig{MaxRequestBodySize: math.MaxInt64}, p, gateway.NewWithTransport(nil, stubGatewayURL)); err == nil {
 		t.Fatal("expected error for MaxRequestBodySize > MaxInt64-1")
+	}
+}
+
+func TestNew_RejectsNilGatewayClient(t *testing.T) {
+	p := pipeline.New([]pipeline.Step{stubStep{name: "stub"}})
+	if _, err := New(config.ServerConfig{}, p, nil); err == nil {
+		t.Fatal("expected error for nil gateway client")
+	}
+}
+
+func TestNew_RejectsGatewayURLWithoutHost(t *testing.T) {
+	p := pipeline.New([]pipeline.Step{stubStep{name: "stub"}})
+	if _, err := New(config.ServerConfig{}, p, gateway.NewWithTransport(nil, "not-a-url")); err == nil {
+		t.Fatal("expected error for gateway URL without a host")
 	}
 }
 
@@ -237,6 +262,10 @@ func TestHandleInference_OverlongRequestIDIsRejected(t *testing.T) {
 }
 
 func TestRoutesRegistered(t *testing.T) {
+	// With the NotFound passthrough in place, an unregistered path no longer
+	// returns 404 — it reverse-proxies to the stub gateway and returns 502.
+	// Assert 200 to prove the path reaches its handler (stub step + handleHealth
+	// both leave the recorder at its default 200).
 	const inferenceBody = `{"model":"m","token_ids":[1,2,3]}`
 	tests := []struct {
 		name   string
@@ -256,8 +285,8 @@ func TestRoutesRegistered(t *testing.T) {
 			req := httptest.NewRequest(tt.method, tt.path, strings.NewReader(tt.body))
 			rec := httptest.NewRecorder()
 			srv.httpServer.Handler.ServeHTTP(rec, req)
-			if rec.Code == http.StatusNotFound || rec.Code == http.StatusMethodNotAllowed {
-				t.Fatalf("route %s %s not registered: got HTTP %d", tt.method, tt.path, rec.Code)
+			if rec.Code != http.StatusOK {
+				t.Fatalf("route %s %s expected 200, got HTTP %d", tt.method, tt.path, rec.Code)
 			}
 		})
 	}
--- a/pkg/coordinator/server/handlers_test.go
+++ b/pkg/coordinator/server/handlers_test.go
@@ -298,3 +298,16 @@ func TestRoutesRegistered(t *testing.T) {
 		})
 	}
 }
+
+func TestRoutesRegistered_MethodMismatchReturns405(t *testing.T) {
+	// A registered route with the wrong method must return 405, not fall
+	// through to the passthrough. Chi's default MethodNotAllowed handler
+	// produces this; the coordinator does not override it.
+	srv := newTestServer(nil)
+	req := httptest.NewRequest(http.MethodGet, gateway.PathChatCompletions, nil)
+	rec := httptest.NewRecorder()
+	srv.httpServer.Handler.ServeHTTP(rec, req)
+	if rec.Code != http.StatusMethodNotAllowed {
+		t.Fatalf("expected 405 for GET on POST-only %s, got %d", gateway.PathChatCompletions, rec.Code)
+	}
+}
--- a/pkg/coordinator/server/handlers_test.go
+++ b/pkg/coordinator/server/handlers_test.go
@@ -188,6 +188,13 @@ func TestNew_RejectsGatewayURLWithoutHost(t *testing.T) {
 	}
 }
 
+func TestNew_RejectsUnparsableGatewayURL(t *testing.T) {
+	p := pipeline.New([]pipeline.Step{stubStep{name: "stub"}})
+	if _, err := New(config.ServerConfig{}, p, gateway.NewWithTransport(nil, "http://gw:notaport")); err == nil {
+		t.Fatal("expected error for unparsable gateway URL")
+	}
+}
+
 // deadlineRecorder wraps httptest.ResponseRecorder with a SetWriteDeadline
 // method so http.NewResponseController can reach it; it records the deadline
 // the handler sets.
--- a/pkg/coordinator/server/passthrough.go
+++ b/pkg/coordinator/server/passthrough.go
@@ -1,0 +1,98 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package server
+
+import (
+	"errors"
+	"log"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/google/uuid"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+	reqcommon "github.com/llm-d/llm-d-router/pkg/common/request"
+
+	"github.com/llm-d/llm-d-router/pkg/coordinator/gateway"
+)
+
+// handlePassthrough is the chi NotFound catch-all: any path the coordinator
+// does not register (e.g. /v1/models, /v1/messages, /v1/responses, /v1/embeddings)
+// is reverse-proxied to the gateway with EPP-Profile: decode, so EPP dispatches
+// it to a decode pod. Method, body, query, and forwarded headers are preserved;
+// X-Request-Id is validated and replaced with a UUID if malformed, matching
+// handleInference's sanitization.
+func (s *Server) handlePassthrough(w http.ResponseWriter, r *http.Request) {
+	requestID := r.Header.Get(reqcommon.RequestIDHeaderKey)
+	if !validRequestID.MatchString(requestID) {
+		requestID = uuid.New().String()
+	}
+
+	logger := ctrl.Log.WithName("passthrough").WithValues(reqcommon.RequestIDHeaderKey, requestID)
+	logger.V(logutil.DEFAULT).Info("received request", "method", r.Method, "path", r.URL.Path)
+
+	// A passthrough response may stream (e.g. /v1/messages with stream:true).
+	// Clear the write deadline like handleInference does for streaming so a
+	// long response is not cut by WriteTimeout mid-stream.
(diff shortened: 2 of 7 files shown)


pkg/coordinator/server/passthrough.go
```diff
@@ -0,0 +1,98 @@
...
+func newPassthroughProxy(logger logr.Logger, gatewayURL *url.URL, transport http.RoundTripper, requestID string) *httputil.ReverseProxy {
+	return &httputil.ReverseProxy{
+		Director: func(r *http.Request) {
+			r.URL.Scheme = gatewayURL.Scheme
+			r.URL.Host = gatewayURL.Host
+			r.Host = gatewayURL.Host
+			r.Header.Set(reqcommon.RequestIDHeaderKey, requestID)
+			r.Header.Set(gateway.EPPProfileHeader, gateway.PhaseDecode)
+		},
+		FlushInterval: -1,
```
roytman, line 122: Nothing verifies the response is actually streamed, not buffered. 
FlushInterval: -1 specifically so SSE/chunked responses (the PR's own stated motivation: /v1/messages with stream:true) reach the client incrementally. 
Every test in `passthrough_test.go` uses an upstream that writes the whole body in one w.Write before returning, so a regression that broke streaming (e.g. FlushInterval changed to 0, transport swapped for one that buffers, a wrapping middleware inserted before r.NotFound) would pass every existing test yet silently degrade to buffered delivery in production. This is cheap to catch: httptest.ResponseRecorder already implements http.Flusher and tracks a Flushed bool field. Have the upstream handler write two chunks with an explicit Flush() between them, then assert rec.Flushed == true (or, more precisely, assert the two chunks were visible to the recorder as separate writes). Priority: add before merge — this is the one gap that could plausibly hide a bug in the feature's core reason for existing.

pkg/coordinator/server/server.go
```diff
@@ -82,9 +84,21 @@ type Server struct {
...
+	gwClient           *gateway.Client
+	gatewayURL         *url.URL
 }
 
-func New(cfg config.ServerConfig, p *pipeline.Pipeline) (*Server, error) {
+func New(cfg config.ServerConfig, p *pipeline.Pipeline, gwClient *gateway.Client) (*Server, error) {
+	if gwClient == nil {
+		return nil, errors.New("server: gateway client is required")
+	}
+	gatewayURL, err := url.Parse(gwClient.BaseURL())
```
roytman: nit: url.Parse error branch is not tested
It has two failure branches: err != nil (parse failure) and gatewayURL.Host == "" (parsed OK, no host). 
TestNew_RejectsGatewayURLWithoutHost (handlers_test.go:184) only exercises the second — "not-a-url" parses without error and falls through to the host check, it never reaches the err != nil path. A value like gateway.NewWithTransport(nil, "http://gw:notaport") (invalid port) fails url.Parse directly and would hit the untested branch. Not critical (malformed gateway URLs are a config-time concern normally caught by integration/startup), but it's a genuinely uncovered branch,

pkg/coordinator/server/passthrough.go
```diff
@@ -0,0 +1,98 @@
...
+		Transport:     transport,
+		ErrorLog:      log.New(&passthroughErrorLogWriter{logger: logger}, "", 0),
+		ErrorHandler: func(w http.ResponseWriter, _ *http.Request, err error) {
```
roytman: Client-cancellation is logged as an Error, indistinguishable from a real backend fault. ErrorHandler calls logger.Error for every RoundTrip failure, including `context.Canceled` from a client disconnecting mid-request — a normal, frequent event on a public catch-all route. The stdlib's own copyBuffer explicitly excludes context. Canceled from its log call a few lines away in the same file the author clearly read, and this repo's own flowcontrol code treats cancellation specially. This will pollute error dashboards/alerts on exactly the route most likely to see abusive traffic.

pkg/coordinator/server/passthrough.go
```diff
@@ -0,0 +1,98 @@
...
+// handlePassthrough is the chi NotFound catch-all: any path the coordinator
+// does not register (e.g. /v1/models, /v1/messages, /v1/responses, /v1/embeddings)
+// is reverse-proxied to the gateway with EPP-Profile: decode, so EPP dispatches
+// it to a decode pod. Method, body, query, and forwarded headers are preserved;
+// X-Request-Id is validated and replaced with a UUID if malformed, matching
+// handleInference's sanitization.
+func (s *Server) handlePassthrough(w http.ResponseWriter, r *http.Request) {
+	requestID := r.Header.Get(reqcommon.RequestIDHeaderKey)
+	if !validRequestID.MatchString(requestID) {
+		requestID = uuid.New().String()
```
roytman, line 80: Invalid/malicious X-Request-Id replacement isn't logged, unlike handleInference which logs rejectedLength when it substitutes a UUID handlers.go:95-99  vs passthrough.go:45-48
 This is the most attacker-exposed handler in the server (catch-all for anything unregistered) and currently leaves zero trace of header-injection probing.

roytman (commented):
Additional comments:

- Missing test: nothing proves streaming actually streams. FlushInterval: -1 exists specifically so SSE/chunked responses aren't buffered — the PR's own stated motivation — but every test writes the full body in one shot before returning. A regression that broke streaming would pass every test. httptest.ResponseRecorder tracks Flushed bool; a two-chunk-with-explicit-Flush() test would catch this cheaply.

dmitripikus:
Regarding 'missing test: nothing proves streaming actually streams'. 
The relevant test is added

pkg/coordinator/server/passthrough.go
dmitripikus: Addressed in TestPassthrough_StreamsChunksAsTheyArrive.

The upstream in the test writes chunk1, calls Flusher.Flush(), writes chunk2, flushes again. The test then asserts httptest.ResponseRecorder.Flushed == true. I sanity-checked it discriminates by temporarily flipping FlushInterval: -1 to 0 — the test fails as expected.

One subtlety worth noting: the upstream sets an explicit Content-Length and a non-SSE Content-Type. Without those, httputil.ReverseProxy auto-flushes anyway — it forces immediate flushing for text/event-stream responses and for responses with ContentLength == -1 (chunked upstream) — and the assertion would pass even if our FlushInterval: -1 were removed. Pinning both takes the stdlib's auto-flush out of the picture so the test really does gate on our proxy config.

pkg/coordinator/server/server.go
dmitripikus: TestNew_RejectsUnparsableGatewayURL is added, using http://gw:notaport (invalid port) ,so url.Parse itself fails and the err != nil branch is exercised.

pkg/coordinator/server/passthrough.go
dmitripikus: Both ErrorHandlers now log context.Canceled / DeadlineExceeded at V(logutil.VERBOSE).Info, keeping Error for real faults. Follows stdlib copyBuffer and pkg/epp/flowcontrol. Tests cover both cancel kinds plus a normal error.

pkg/coordinator/server/passthrough.go
dmitripikus: handlePassthrough() now logs "replaced invalid client request ID" at logutil.DEFAULT with rejectedLength only, matching handleInference. Header-injection probing leaves a trace; raw attacker content still never touches the log.

dmitripikus:
@roytman , regarding your comment about code duplication between main flow and passthrough flow,
There is duplication (request-id sanitize + replacement log, and now the two proxy ErrorHandlers / errorLogWriters). Under  "smallest correct change" rule (specified in AGENTS.md) may be it's better not to refactor onto this PR, but to  open a follow-up issue for it — start with the two request-id helpers in pkg/coordinator/server/, and consider a pkg/coordinator/common/proxyutil/ for the shared reverse-proxy scaffolding separately.

What do you think?

pkg/coordinator/server/passthrough.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/google/uuid"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+	reqcommon "github.com/llm-d/llm-d-router/pkg/common/request"
+
+	"github.com/llm-d/llm-d-router/pkg/coordinator/gateway"
+)
+
+// handlePassthrough is the chi NotFound catch-all: any path the coordinator
```
roytman: I think it will be better to store all passthrough-related values in a separate struct:
in server.go
```diff
type Server struct {
        httpServer         *http.Server
        pipeline           *pipeline.Pipeline
        maxRequestBodySize int64
-       gwClient           *gateway.Client
-       gatewayURL         *url.URL
+       passthrough        *passthroughHandler
 }

func New(cfg config.ServerConfig, p *pipeline.Pipeline, gwClient *gateway.Client) (*Server, error) {
-       if gwClient == nil {
-               return nil, errors.New("server: gateway client is required")
-       }
-       gatewayURL, err := url.Parse(gwClient.BaseURL())
+       passthrough, err := newPassthroughHandler(gwClient)
        if err != nil {
-               return nil, fmt.Errorf("server: parse gateway URL %q: %w", gwClient.BaseURL(), err)
-       }
-       if gatewayURL.Host == "" {
-               return nil, fmt.Errorf("server: gateway URL %q is missing a host", gwClient.BaseURL())
        }
   . . . 

-               gwClient:           gwClient,
-               gatewayURL:         gatewayURL,
+               passthrough:        passthrough,

-       r.NotFound(s.handlePassthrough)
+       r.NotFound(s.passthrough.ServeHTTP)
``` 

and in the passthrough.go
```diff
-// handlePassthrough is the chi NotFound catch-all: any path the coordinator
+// passthroughHandler is the chi NotFound catch-all: any path the coordinator

-func (s *Server) handlePassthrough(w http.ResponseWriter, r *http.Request) {
+type passthroughHandler struct {
+       gatewayURL *url.URL
+       transport  http.RoundTripper
+}
+
+// newPassthroughHandler resolves and validates the gateway target once, at
+// server construction, so a misconfigured gateway URL fails startup rather
+// than the first passthrough request.
+func newPassthroughHandler(gwClient *gateway.Client) (*passthroughHandler, error) {
+       if gwClient == nil {
+               return nil, errors.New("server: gateway client is required")
+       }
+       gatewayURL, err := url.Parse(gwClient.BaseURL())
+       if err != nil {
+               return nil, fmt.Errorf("server: parse gateway URL %q: %w", gwClient.BaseURL(), err)
+       }
+       if gatewayURL.Host == "" {
+               return nil, fmt.Errorf("server: gateway URL %q is missing a host", gwClient.BaseURL())
+       }
+       return &passthroughHandler{gatewayURL: gatewayURL, transport: gwClient.Transport()}, nil
+}
+
+func (h *passthroughHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 ....
-       proxy := newPassthroughProxy(logger, s.gatewayURL, s.gwClient.Transport(), requestID)
+       proxy := newPassthroughProxy(logger, h.gatewayURL, h.transport, requestID)
```

pkg/coordinator/server/passthrough.go
```diff
@@ -0,0 +1,112 @@
...
+
+	// A passthrough response may stream (e.g. /v1/messages with stream:true).
+	// Clear the write deadline like handleInference does for streaming so a
+	// long response is not cut by WriteTimeout mid-stream.
+	if err := http.NewResponseController(w).SetWriteDeadline(time.Time{}); err != nil && !errors.Is(err, http.ErrNotSupported) {
+		logger.V(logutil.DEFAULT).Info("could not clear write deadline", "error", err)
+	}
+
+	proxy := newPassthroughProxy(logger, s.gatewayURL, s.gwClient.Transport(), requestID)
+	proxy.ServeHTTP(w, r)
```
roytman, line 105: handlePassthrough never applies `s.maxRequestBodySize` — it hands r.Body straight to ReverseProxy. server.max_request_body_size silently stops applying to every unregistered path. Wrap in http.MaxBytesReader, or document the exemption as deliberate.

pkg/coordinator/server/handlers_test.go
```diff
@@ -237,6 +269,10 @@ func TestHandleInference_OverlongRequestIDIsRejected(t *testing.T) {
 }
 
 func TestRoutesRegistered(t *testing.T) {
+	// With the NotFound passthrough in place, an unregistered path no longer
```
roytman, line 272: Nothing asserts the 405-on-method-mismatch claim from the PR description. Add one case, e.g. GET on gateway.PathChatCompletions expecting 405.

pkg/coordinator/server/passthrough.go
```diff
@@ -0,0 +1,112 @@
...
+	logger := ctrl.Log.WithName("passthrough").WithValues(reqcommon.RequestIDHeaderKey, requestID)
+	if requestIDReplaced && clientRequestID != "" {
+		// Log the rejected length, never the raw value, to avoid reflecting
+		// attacker-controlled content into the log.
+		logger.V(logutil.DEFAULT).Info("replaced invalid client request ID", "rejectedLength", len(clientRequestID))
+	}
+	logger.V(logutil.DEFAULT).Info("received request", "method", r.Method, "path", r.URL.Path)
+
+	// A passthrough response may stream (e.g. /v1/messages with stream:true).
+	// Clear the write deadline like handleInference does for streaming so a
```
roytman: Claims parity with handleInference, which only clears the write deadline when stream == true. Here it's unconditional, dropping that guard for every request. Fix the comment, or gate it.

pkg/coordinator/server/passthrough.go
```diff
@@ -0,0 +1,112 @@
...
+			// public routes, not backend faults; keep them out of error dashboards.
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+				logger.V(logutil.VERBOSE).Info("passthrough proxy: client cancelled", "error", err)
+			} else {
+				logger.Error(err, "passthrough proxy error")
+			}
+			w.WriteHeader(http.StatusBadGateway)
+		},
+	}
+}
```
roytman: `errors.Is(err, context.DeadlineExceeded)` here is true whenever the gateway client's `ResponseHeaderTimeout` fires - Go's `net/http timeoutError` (returned by `Transport.RoundTrip` on that timeout) has an `Is` method that deliberately matches `context.DeadlineExceeded` (`net/http/transport.go`). That's a backend-side timeout (the gateway/decode pod didn't answer in time), not client cancellation, so this branch demotes exactly the signal this Error log exists to catch.

There's no legitimate client-side source of `context.DeadlineExceeded` in this flow either — nothing in `pkg/coordinator` puts a deadline on the inbound request's context (`cfg.ReadTimeout/WriteTimeout` are connection-level, not context cancellation), so in practice this branch never actually observes a client-imposed cutoff, only backend timeouts.


```suggestion
		ErrorHandler: func(w http.ResponseWriter, _ *http.Request, err error) {
			// A request whose own context already ended (client disconnected) is a
            // routine event, not a backend fault; keep it out of error dashboards.
            // Checking req.Context().Err() rather than err's identity matters:
            // Transport.RoundTrip's ResponseHeaderTimeout error satisfies
            // errors.Is(err, context.DeadlineExceeded) by stdlib design (see
            // net/http's timeoutError), which would otherwise misclassify a hung
            // backend as a client cancellation.
			if req.Context().Err() != nil {
                logger.V(logutil.VERBOSE).Info("passthrough proxy: client cancelled", "error", req.Context().Err())
            } else {
                // RoundTrip failed for a reason unrelated to the client's own request
                // lifecycle (e.g. ResponseHeaderTimeout) — a real backend fault.
                logger.Error(err, "passthrough proxy error")
           }
            w.WriteHeader(http.StatusBadGateway)
		},
	}
}
```

This works because `httputil.ReverseProxy` passes the Director-rewritten request into `ErrorHandler`, and that request's context is derived from the original inbound request's context (`reverseproxy.go,` `ctx := req.Context()` near the top of `ServeHTTP`) — so r`eq.Context().Err()` still faithfully reflects whether the original client request was canceled, regardless of what `RoundTrip` itself returned.

pkg/coordinator/steps/decode_proxy.go
```diff
@@ -110,7 +110,13 @@ func newDecodeProxy(logger logr.Logger, transport http.RoundTripper, modifyRespo
...
 				return
 			}
-			logger.Error(proxyErr, "proxy error")
+			// Client cancellation and request-context deadlines are routine events on
+			// public routes, not backend faults; keep them out of error dashboards.
+			if errors.Is(proxyErr, context.Canceled) || errors.Is(proxyErr, context.DeadlineExceeded) {
+				logger.V(logutil.VERBOSE).Info("decode proxy: client cancelled", "error", proxyErr)
+			} else {
+				logger.Error(proxyErr, "proxy error")
+			}
```
roytman: Same root cause as the `passthrough.go` comment on this PR: `ResponseHeaderTimeout` on the gateway transport produces an error that satisfies errors.Is(_, context.DeadlineExceeded) by stdlib design, so this also demotes a real backend timeout to V(VERBOSE).

The same fix applies here, keeping the existing errCacheMiss short-circuit ahead of it:

```suggestion
ErrorHandler: func(w http.ResponseWriter, req *http.Request, proxyErr error) {
                        if errors.Is(proxyErr, errCacheMiss) {
                                return
                        }
                       // A request whose own context already ended (client disconnected) is a
                       // routine event, not a backend fault; keep it out of error dashboards.
                       // Checking req.Context().Err() rather than proxyErr's identity matters:
                       // Transport.RoundTrip's ResponseHeaderTimeout error satisfies
                       // errors.Is(err, context.DeadlineExceeded) by stdlib design (see
                       // net/http's timeoutError), which would otherwise misclassify a hung
                       // backend as a client cancellation.
                       if req.Context().Err() != nil {
                               logger.V(logutil.VERBOSE).Info("decode proxy: client cancelled", "error", req.Context().Err())
                        } else {
                                logger.Error(proxyErr, "proxy error")
                        }
```

roytman (commented):
Several more comments.

pkg/coordinator/server/passthrough.go
dmitripikus: Done — applied your suggestion. The gateway URL and transport are moved to a passthroughHandler struct, and newPassthroughHandler runs the three startup checks (nil client, URL parse, missing host) up front. The NotFound route now points at s.passthrough.ServeHTTP.

pkg/coordinator/server/passthrough.go
dmitripikus: Done. passthroughHandler.ServeHTTP now wraps r.Body with http.MaxBytesReader using the configured cap, and the reverse proxy's ErrorHandler maps *http.MaxBytesError to 413.

Thanks!

pkg/coordinator/server/handlers_test.go
dmitripikus: Added TestRoutesRegistered_MethodMismatchReturns405 in handlers_test.go — sends GET on gateway.PathChatCompletions through the chi router and asserts 405. Locks the boundary against a future refactor that could route method mismatches to the passthrough.

Thanks!

pkg/coordinator/server/passthrough.go
dmitripikus: The passthrough never reads the request body, so it can't tell if a request is streaming. Since most catch-all paths here can stream (/v1/messages and others), I went with "fix the comment" rather than "gate it".

Thanks!

pkg/coordinator/server/passthrough.go
dmitripikus: The old check tried to spot a cancelled client by looking at the error, but Go's HTTP client returns the same kind of error when the backend is slow to respond — so a hung gateway was being logged as if the client had walked away. Fixed by looking at the request itself instead: if the request is still alive, the error is a real backend problem. Same fix applied to the decode proxy, plus a test that pins the "slow backend, live client" case.

pkg/coordinator/steps/decode_proxy.go
dmitripikus: Fixed in the same commit as the passthrough.go
Thanks!

roytman (approved):
Thank you.

PR #2359, opened 2026-08-11
feat(epp): order all requestcontrol hooks by data-dependency DAG

### LINKED ISSUE #2040: [feat] Order all plugin hook lists by the data-dependency DAG, not just Produce
[`ValidateAndOrderDataDependencies`](https://github.com/llm-d/llm-d-router/blob/7b7342343136d0a53b6f2a9e268300057cf0dbc6/pkg/epp/datalayer/data_graph.go#L37) topologically sorts plugins by their `Produces()`/`Consumes()` declarations, but the result is only applied to the Produce extension point via [`OrderDataProducerPlugins`](https://github.com/llm-d/llm-d-router/blob/7b7342343136d0a53b6f2a9e268300057cf0dbc6/pkg/epp/requestcontrol/request_control_config.go#L112) ([called here in runner.go](https://github.com/llm-d/llm-d-router/blob/7b7342343136d0a53b6f2a9e268300057cf0dbc6/cmd/epp/runner/runner.go#L737)). Every other hook list (preRequest, preAdmission, responseReceived, responseStreaming, admission) is populated by [`AddPlugins`](https://github.com/llm-d/llm-d-router/blob/7b7342343136d0a53b6f2a9e268300057cf0dbc6/pkg/epp/requestcontrol/request_control_config.go#L88) from [`GetAllPlugins`](https://github.com/llm-d/llm-d-router/blob/7b7342343136d0a53b6f2a9e268300057cf0dbc6/pkg/epp/framework/interface/plugin/handle.go#L88), which iterates a map. So those hooks run in an order that is arbitrary and changes across EPP restarts (e.g. [`runPreRequestPlugins`](https://github.com/llm-d/llm-d-router/blob/7b7342343136d0a53b6f2a9e268300057cf0dbc6/pkg/epp/requestcontrol/director.go#L581) just walks the list).

This bit us in #1856: `PredictedLatency.PreRequest` read the live `InFlightLoad` attribute that `InFlightLoadProducer.PreRequest` mutates, so a training feature silently changed in meaning depending on which plugin happened to run first that boot. The fix there was to move the read into Produce, but that's a per-plugin workaround. Any future cross-plugin interaction in a non-Produce hook hits the same trap, and nothing in the API warns about it: `Consumes()` reads like a plugin-level ordering contract when it's actually Produce-only.

Proposal:

1. Apply the already-computed topological order to all hook lists at startup (the sort exists, this is a few lines at the `OrderDataProducerPlugins` call site).
2. Decide per extension point what the order should mean. Producer-before-consumer is clearly right for PreRequest. For response hooks the reverse may be correct, since producers release state there.
3. Document on `DataDependencies` which extension points the ordering covers.

Even if 2 needs discussion, doing 1 alone makes hook order deterministic across restarts.

chethanuk:
**What type of PR is this?**

/kind feature

`ValidateAndOrderDataDependencies` already computes a topological order over the plugins' `Produces()`/`Consumes()` declarations, but `OrderDataProducerPlugins` applied it to the DataProducer list only. Every other hook list is populated by `AddPlugins` from `GetAllPlugins`, which ranges a map, so those hooks ran in an order that changed between EPP restarts. #1856 was one instance: `PredictedLatency.PreRequest` read an `InFlightLoad` attribute that `InFlightLoadProducer.PreRequest` mutates, and which ran first depended on the boot.

`OrderDataProducerPlugins` becomes `OrderPlugins` (`pkg/epp/requestcontrol/request_control_config.go:130`) and applies the same order to all seven requestcontrol extension points: RequestHeader, Screener, Admission, DataProducer, PreRequest, ResponseReceived, ResponseStreaming. `Screener` is the newest of the seven — #2259 added it after this branch was first written — and it is included (`request_control_config.go:137`); screener results are set-intersected, so ordering does not change which endpoints survive, but it makes execution and the per-plugin latency metrics deterministic and keeps the "every extension point" contract from having a hole the next time one is added.

Plugins the DAG does not rank (neither producer nor consumer) are kept and appended in name order rather than in arrival order — arrival order is map order, so leaving it alone would leave the nondeterminism this PR removes (`orderByName`, `request_control_config.go:149`).

`TopologicalSort` itself was nondeterministic: Kahn's algorithm may dequeue any ready node, and the queue was built from a map. It now dequeues the largest ready node, which the existing final reverse turns into ascending name order (`pkg/epp/util/utils.go:59`).

Direction: producer-before-consumer is right for RequestHeader, Screen, Admit, Produce and PreRequest. For ResponseHeader and ResponseBody a producer may release the state a consumer reads, so the correct direction is not settled; those lists are ordered the same way for determinism only, and the `DataDependencies` doc comment says so rather than asserting producer-first is canonical there.

On @shmuelk's note that this duplicates #1421/#1876: #1876 orders plugin *instantiation* from `pluginRef` names in the configuration, which is a different axis — the instantiated plugins still land in the handle's map, and the per-request hook lists are still built from ranging it. This PR does not change instantiation order, and #1876's `pluginRef` dependencies are untouched.

Fixes #2040

**Test plan**:

`go test ./pkg/epp/requestcontrol/ ./pkg/epp/util -run 'TestConfig_OrderPlugins|TestTopologicalSort' -count=1 -v`

```
--- PASS: TestConfig_OrderPlugins_AllHookLists (0.00s)
    --- PASS: TestConfig_OrderPlugins_AllHookLists/preRequest_orders_producer_before_consumer_and_keeps_unranked_plugins
    --- PASS: TestConfig_OrderPlugins_AllHookLists/requestHeader_orders_producer_before_consumer
    --- PASS: TestConfig_OrderPlugins_AllHookLists/screeners_order_producer_before_consumer
    --- PASS: TestConfig_OrderPlugins_AllHookLists/responseReceived_orders_producer_before_consumer
    --- PASS: TestConfig_OrderPlugins_AllHookLists/responseStreaming_orders_producer_before_consumer
    --- PASS: TestConfig_OrderPlugins_AllHookLists/admission_orders_producer_before_consumer
    --- PASS: TestConfig_OrderPlugins_AllHookLists/dataProducer_orders_producer_before_consumer
--- PASS: TestConfig_OrderPlugins_EdgeCases (0.00s)
    --- PASS: TestConfig_OrderPlugins_EdgeCases/empty_config_with_empty_order
    --- PASS: TestConfig_OrderPlugins_EdgeCases/no_plugin_is_ranked:_unranked_plugins_are_kept,_sorted_by_name
    --- PASS: TestConfig_OrderPlugins_EdgeCases/order_contains_names_absent_from_the_config
    --- PASS: TestConfig_OrderPlugins_EdgeCases/ranked_and_unranked_mixed:_ranked_first,_unranked_appended_by_name
--- PASS: TestConfig_OrderPlugins_InputOrderIndependent (0.00s)
    --- PASS: TestConfig_OrderPlugins_InputOrderIndependent/input_order_[0_1_2_3]
    --- PASS: TestConfig_OrderPlugins_InputOrderIndependent/input_order_[3_2_1_0]
    --- PASS: TestConfig_OrderPlugins_InputOrderIndependent/input_order_[1_3_0_2]
    --- PASS: TestConfig_OrderPlugins_InputOrderIndependent/input_order_[2_0_3_1]
ok  	github.com/llm-d/llm-d-router/pkg/epp/requestcontrol	0.064s
--- PASS: TestTopologicalSort_Deterministic (0.00s)
    (8 subtests: empty graph, single node, independent nodes ascending, dependency
     respected, linear chain, diamond, topological order wins over name order,
     cycle is rejected)
ok  	github.com/llm-d/llm-d-router/pkg/epp/util	0.008s
```

- [x] Red/green checked: deleting `c.screeners = orderByName(...)` fails `screeners_order_producer_before_consumer` and two of the four `InputOrderIndependent` permutations; reducing `OrderPlugins` to the old DataProducer-only behaviour fails 11 subtests across the three tests.
- [x] `go test -race -count=1 ./pkg/epp/requestcontrol/... ./pkg/epp/util ./pkg/epp/framework/interface/plugin ./cmd/epp/runner` — all ok
- [x] `go vet` clean on the same packages
- [x] The mocks carry compile-time `var _ fwkrc.X = &hookPlugin{}` assertions for all seven interfaces, so a future signature change cannot silently drop a mock off a hook list and leave the matrix asserting on empty slices

```release-note
All requestcontrol extension points (RequestHeader, Screener, Admission, DataProducer, PreRequest, ResponseReceived, ResponseStreaming) now execute in the order computed from the plugins' Produces/Consumes declarations, so hook execution order no longer varies between EPP restarts. Plugins that declare neither run after the ordered ones, in name order.
```


chethanuk:
ignore

chethanuk:
Hi @ahg-g is issue https://github.com/llm-d/llm-d-router/issues/2040#issue-4896968530 closed? or do I need to change something in PR? What was the reason we closed it?

ahg-g:
sorry, by mistake

ahg-g (approved):

PR #2363, opened 2026-08-12
Filter prefix cache events by cache group kind

### LINKED ISSUE #2362: bug(kvevents): sparse hybrid-cache groups create false prefix residency
**What happened**:

The precise-prefix index treats sparse hybrid KV-cache events as contiguous prefix state.

DeepSeek-V4-Flash emits group-aware `BlockStored` events with heterogeneous layouts. In the captured workload, group 0 is dense `mla_attention` with block size 256. Groups 1–4 are `sliding_window_mla` with block sizes 64, 64, 4, and 8.

The sliding-window events contain the logical token span but only hashes for the blocks resident in that cache group. `Pool.processEventBatch` generates canonical request keys from the full token span, then `Index.Add` infers the engine-to-request-key mapping from the number of hashes and canonical keys. That ratio does not preserve logical block positions for a sparse group, so the router records false prefix residency and can resolve parent hashes against the wrong lineage.

This presents as low `max_match_blocks` and poor prompt reuse even when the reusable prefix is resident. The unpatched 60 RPM, TP2 × 4-replica run reached 24.7% cumulative prompt-token reuse.

Only KV-cache events whose group semantics and block span preserve canonical prefix positions should contribute to precise-prefix indexing.

Group-aware stores should be accepted for:

- `full_attention`
- `mla_attention`
- `sink_full_attention`

These events should also represent a dense span:

```text
len(tokens) % block_size == 0
len(tokens) / block_size == len(block_hashes)
```

Events from sparse or unsupported cache groups should not add or remove prefix residency. Events without group metadata should remain supported for compatibility with legacy producers.

1. Deploy llm-d precise prefix-cache routing with DeepSeek-V4-Flash and speculative decoding.
2. Use the nightly scheduler image and the DP event-port configuration from [[llm-d#2203](https://github.com/llm-d/llm-d/pull/2203)](https://github.com/llm-d/llm-d/pull/2203).
3. Configure the router token processor with `blockSize: 16`.
4. Run the attached 60 RPM workload against four TP2 replicas using the [[Oreo harness](https://github.com/flexwang/oreo)](https://github.com/flexwang/oreo).
5. Inspect group-aware `BlockStored` events and routing traces.
6. Observe sparse `sliding_window_mla` events entering the prefix index, parent lookup failures, low `max_match_blocks`, and low cumulative prompt reuse.

The deployment manifests are available in the [[reproduction gist](https://gist.github.com/flexwang/b5dd7d3e77da1885694a1d5535d27444)](https://gist.github.com/flexwang/b5dd7d3e77da1885694a1d5535d27444).

Changing the router block size does not recover the missing logical positions. A smaller canonical block size increases the number of request keys, but a sparse hash list still cannot be mapped safely without per-hash positions.

A bounded run that filtered unsupported and non-dense groups completed 28 requests without observed errors and reached 75.9% cumulative prompt-token reuse. The deployment changed from three to four ready replicas during that run, so this is validation of the filtering behavior rather than a matched performance comparison.

The filtered run skipped 660 stores and 38,647 removals. It still recorded 29 parent-key lookup failures for admitted groups. Those failures need separate lineage handling; the group filter prevents sparse events from corrupting the index but does not reconstruct missing parent mappings.

**Environment**:

- Kubernetes: CKS, H200s
- llm-d-router: precise-prefix-cache producer with `tokenProcessorConfig.blockSize: 16`
- Model: DeepSeek-V4-Flash
- Serving topology: TP2 × 4 replicas
- Scheduler: nightly image with DP event ports configured according to llm-d#2203
- Workload: `prod_60rpm.jsonl`, driven at 60 RPM through Oreo (DM for info)

vMaroon:
**What type of PR is this?**

/kind bug

Group-aware KV-cache events can describe cache groups whose hashes do not cover a contiguous logical token span. The router currently generates canonical keys from the full token span and lets `Index.Add` infer the engine-to-request-key mapping from the two list lengths. That creates false prefix residency for sparse hybrid-cache groups.

This change admits group-aware stores only for `full_attention`, `mla_attention`, and `sink_full_attention`. It also requires each admitted event to describe a dense span:

```text
len(tokens) % block_size == 0
len(tokens) / block_size == len(block_hashes)
```

The filter runs before parent-key lookup or index mutation. Group metadata is still learned from rejected stores so matching removals can be skipped consistently. Events without group metadata retain the existing behavior.

The change exposes rejected events through:

- `llm_d_router_epp_kv_cache_events_stores_skipped_total{cache_kind,reason}`
- `llm_d_router_epp_kv_cache_events_removals_skipped_total{cache_kind,reason}`

Parent-key lookup errors for admitted events include the group, cache kind, token count, hash count, and block size.

On the DeepSeek-V4-Flash reproduction, a bounded run completed 28 requests without observed errors and reached 75.9% cumulative prompt-token reuse. The run changed from three to four ready replicas, so this validates the event-filtering behavior rather than providing a matched performance comparison. The corresponding unpatched TP2 × 4 run reached 24.7% reuse.

The admitted full-attention group still recorded 29 parent-key lookup failures. This PR reports the event context for those failures but does not reconstruct missing lineage.

Fixes #2362

**Test plan**:

- [x] `go test ./pkg/kvcache/metrics ./pkg/kvevents -count=1`
- [x] `make build`
- [x] Run the DeepSeek-V4-Flash workload against the patched router

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/kvcache/metrics/collector.go
+++ b/pkg/kvcache/metrics/collector.go
@@ -37,6 +37,11 @@ const routerSubsystem = "llm_d_router_epp"
 // Keeping it as a constant localizes label-schema changes to this package.
 const podIdentifierLabel = "pod_identifier"
 
+const (
+	cacheKindLabel = "cache_kind"
+	reasonLabel    = "reason"
+)
+
 // dualCounter emits a value to both the deprecated kvcache_* counter and the
 // current llm_d_router_epp_* counter. It satisfies prometheus.Collector so both
 // register, and exposes the recording/read methods the call sites use.
@@ -134,6 +139,22 @@ var (
 	DedupRemovedHashesForwarded = newDualCounter("kvevents", "dedup_removed_hashes_forwarded_total",
 		"kv_cache_events_dedup_removed_hashes_forwarded_total",
 		"Block hashes forwarded for eviction after the KV-event dedup filter (block hashes, not BlockRemoved events)")
+	// KVEventStoresSkipped counts group-aware stores that cannot safely
+	// contribute to prefix indexing.
+	KVEventStoresSkipped = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Subsystem: routerSubsystem, Name: "kv_cache_events_stores_skipped_total",
+		Help: metricsutil.HelpMsgWithStability(
+			"Total number of KV-cache store events skipped before prefix indexing",
+			compbasemetrics.ALPHA),
+	}, []string{cacheKindLabel, reasonLabel})
+	// KVEventRemovalsSkipped counts removals for groups excluded from prefix
+	// indexing.
+	KVEventRemovalsSkipped = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Subsystem: routerSubsystem, Name: "kv_cache_events_removals_skipped_total",
+		Help: metricsutil.HelpMsgWithStability(
+			"Total number of KV-cache removal events skipped for groups excluded from prefix indexing",
+			compbasemetrics.ALPHA),
+	}, []string{cacheKindLabel, reasonLabel})
 
 	// SubscriberActive tracks the number of ZMQ subscribers currently managed by
 	// the SubscriberManager. A subscriber is counted from creation until it is
@@ -189,6 +210,7 @@ func Collectors() []prometheus.Collector {
 		Admissions, Evictions,
 		LookupRequests, LookupHits, LookupLatency, MaxPodHitCount,
 		DedupRemovedHashesSuppressed, DedupRemovedHashesForwarded,
+		KVEventStoresSkipped, KVEventRemovalsSkipped,
 		SubscriberActive, SubscriberReconnections, MessagesReceived, ZMQErrors,
 		PoolQueueDepth, PoolCapacity,
 	}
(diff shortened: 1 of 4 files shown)


bongwoobak:
I tested this too, hope it helps accelerate.

Single MI308X node, gemma-4-31B-it with gemma-4-31B-it-assistant as the MTP drafter, vLLM 0.25.1, `--block-size 64`, router `blockSize: 64`. Built both EPP binaries from source, `main@4a0edbb0` vs this branch, same engine. Fresh EPP index per run, prompt never cached, 5120 tokens, two requests, `max_tokens: 256`.

| | baseline | this PR |
|---|---|---|
| `engine key not found` | 30 | 0 |
| distinct parent hashes missed | 3 | 0 |
| `kvcache_index_admissions_total` | 498 | 88 |
| `stores_skipped_total{sliding_window,unsupported_cache_kind}` | - | 80 |
| `removals_skipped_total` | - | 0 |
| vLLM prefix cache hit on the 2nd request | 4992 / 5120 | 4992 / 5120 |

The 498 to 88 drop is not a coverage loss. All six groups carry the same `token_ids` and derive the same canonical request keys, so the baseline was writing the same 80 keys six times. The distinct key set is identical and vLLM side reuse is unchanged.

On the 29 admitted-group failures you mentioned, in a run where the engine had cached the prefixes before EPP started I saw the same thing, all of them `full_attention` with a dense span. They go to zero once the lineage is published from block 0, so it looks like EPP joining mid-stream rather than a filter gap.

flexwang:
Thanks a lot for the fix, we also tested on our side and it provides meaningful boost on prefix cache hit rate in a scaled system.

ahg-g (approved):

PR #2371, opened 2026-08-13
feat(epp): add latency-observation scorer and producer

### LINKED ISSUE #2366: latency-aware routing from observed TTFT

Endpoint scoring by observed time-to-first-token, using statistics the EPP measures itself rather than metrics scraped from the endpoint.

- `latency-observer-producer`: a **black-box observer**. It measures each request's actual time-to-first-token, pairs it with the in-flight load the endpoint was carrying at dispatch, and publishes three anchor points per endpoint: a **floor**, a **low-load** and a **typical-load** latency.
- `latency-observation-scorer`: interpolates between those anchors at the endpoint's **live** in-flight load to predict this request's TTFT, and ranks candidates on it.

`inflight-load-producer` is auto-created from the observer's required data key. The observer itself is listed under `dataLayer.sources`, which is how the datalayer picks up the periodic driver described below.

This targets **EPP hub deployments**, which route across peer cluster gateways. Unlike the metric scorers it needs no separate multi-cluster variant, because it never reads a per-pod metric field but only the response it already sees.

**Why is this needed**:

Every existing load-aware scorer reads something the endpoint reports about itself: queue depth, KV-cache utilisation, running requests. That works when the endpoint is a pod the EPP scrapes directly.

It stops working at a trust boundary. A multi-cluster hub reaches each peer over its gateway, and today's only cross-cluster signals (`llm-d.ai/multicluster-kv-cache-utilization` and `llm-d.ai/multicluster-queue-size`) are scraped from that peer, so the peer must be a cooperating llm-d EPP exposing pool metrics across the boundary. That rules out any cluster the operator does not run, cannot instrument, or can only reach through a gateway exposing nothing but the inference API.

Scraped metrics are also only ever as fresh as the scrape interval: they describe the endpoint as it was up to one poll ago, and across a gateway that lag compounds with the peer's own collection delay. An observation is produced by the request itself, so there is nothing to poll and nothing to go stale. The EPP already sees every request it dispatched and every response that came back, and currently discards them.

### The periodic recompute

EPP has no data source that watches traffic, so measurement happens in the request-control hooks, while the summary is republished every `intervalDuration`.

The summary reaches the scorer as an endpoint attribute, published through a `DynamicAttribute` attached once per endpoint. Each republish therefore swaps a pointer behind that attribute rather than writing the attribute map, which is how `inflight-load-producer` already publishes.

The recompute is driven by the datalayer's **polling dispatcher**, which already visits every endpoint on a tick, so the plugin owns no goroutine of its own. It needs no framework change: `PollingDispatcher` is an existing datalayer interface, and `Dispatch(ctx, endpoint)` means only "your turn, for this endpoint, now" — nothing obliges it to scrape. `cross-replica-publisher` is already precisely this, a `PollingDispatcher` with no extractors that polls nothing.

---

Happy to open a PR to work on this.

Mohammad-nassar10:
**What type of PR is this?**
/kind feature

Adds two EPP plugins that score endpoints by predicted time-to-first-token. `latency-observer-producer` times each request's TTFT, pairs it with the in-flight load the endpoint was carrying at dispatch, and publishes three anchor points: a floor, a low-load and a typical-load latency. `latency-observation-scorer` interpolates between those anchors at the endpoint's live in-flight load to predict this request's TTFT, and ranks candidates on it.

The data producer adopts an observer pattern where each endpoint's curve is measured from its own traffic, endpoints on different hardware or model sizes compare on equal terms with no capacity model to maintain. The inputs are fresher too: the anchors come from responses the EPP already received, and the in-flight load is read at scoring time, so nothing is stale by a scrape interval or by the endpoint's own reporting lag. Needing nothing from the endpoint but a response also means it works where scraping cannot reach: a hub routing across peer cluster gateways, or an endpoint that exposes no metrics at all.

Mechanically it reuses what is already there: request-control hooks for measurement, a `DynamicAttribute` for publication (as `inflight-load-producer` does), and a `PollingDispatcher` for the periodic recompute (as `cross-replica-publisher` does).

We did runs with two different models and see that the predicted ttft of the scorer follows/tracks the actual ttft.
Also, comparing against `multicluster-queue-scorer`, the existing multi-cluster load-aware option. Setup: two namespaces serving the same model on unequal GPU counts (3 and 2), a summarization workload with RPS starting from 12 to 60 and back. This scorer matches queue-scorer at the low stages, where nothing is queuing, and comes in slightly ahead at the loaded stages on both the median and the tail.

Fixes #2366 

```release-note
Add `latency-observation-scorer` and `latency-observer-producer`: Score endpoints by predicted time-to-first-token, measured from observed responses rather than scraped metrics, using each endpoint's own latency curve at its current load.
```


pkg/epp/framework/plugins/datalayer/attribute/latency/ttft_percentiles.go
```diff
@@ -0,0 +1,83 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
vMaroon:
```suggestion
Copyright 2026 The llm-d Authors.
```

This and all other boilerplates.

pkg/epp/framework/plugins/requestcontrol/dataproducer/latencyobserver/README.md
```diff
@@ -0,0 +1,167 @@
...
+## Load confidence
+
+A percentile over a handful of requests is noise, so `minRequests` gates the snapshot twice: the
+floor on `Observations` (cumulative), the loaded anchors on `RecentRequestCount` (short window). Below
+the threshold the floor reads as zero, marking the endpoint cold for the scorer exploration logic;
+with a floor but too few recent requests the scorer predicts at the floor alone. The split lets a
+calibrated endpoint keep its floor through an idle spell while it re-earns the loaded anchors. Both
+counts and the threshold travel in the snapshot, so the scorer needs no parameter of its own.
+
+## Published fields
```
vMaroon, line 106: How should the knobs for this plugins tuned? Plain english, intuitive directions are helpful.

pkg/epp/framework/plugins/scheduling/scorer/latencyobservation/README.md
```diff
@@ -0,0 +1,178 @@
...
+a probe never distorts the trusted endpoints' normalisation.
+
+Together with the observer's load confidence gate this reproduces what a manual warmup would do: a
+new endpoint reads as cold → receives probes → crosses the calibration threshold → competes on its
+true latency.
+
+Note the default is non-zero. At `0` an uncalibrated endpoint is seeded at the best observed TTFT
+and ties for the win, taking a share of all traffic before anything is known about it.
+
+## Parameters
```
vMaroon, line 134: How should the knobs for this plugins tuned? Plain english, intuitive directions are helpful.

pkg/epp/framework/plugins/scheduling/scorer/latencyobservation/README.md
```diff
@@ -0,0 +1,178 @@
+# Latency Observation Scorer
```
vMaroon, line 1: Do you have benchmarks that show this in action? Standalone is nice, but comparing with llm-d's closest/most relevant offering is great.

cc @kaushikmitr

pkg/epp/framework/plugins/scheduling/scorer/latencyobservation/README.md
Mohammad-nassar10: Added a tuning column. Thanks!

pkg/epp/framework/plugins/requestcontrol/dataproducer/latencyobserver/README.md
Mohammad-nassar10: These published fields are not tunable, maybe you meant the "Parameters" section, I added a tuning column there.

pkg/epp/framework/plugins/datalayer/attribute/latency/ttft_percentiles.go
Mohammad-nassar10: Fixed this in the new added files, thanks!

pkg/epp/framework/plugins/scheduling/scorer/latencyobservation/README.md
Mohammad-nassar10: We did runs with two different models and see that the predicted ttft of the scorer follows/tracks the actual ttft.
Also, comparing against `multicluster-queue-scorer`, the existing multi-cluster load-aware option. Setup: two namespaces serving the same model on unequal GPU counts (3 and 2), a summarization workload with RPS starting from 12 to 60 and back. This scorer matches `queue-scorer` at the low stages, where nothing is queuing, and comes in slightly ahead at the loaded stages on both the median and the tail.

pkg/epp/framework/plugins/requestcontrol/dataproducer/README.md
```diff
@@ -34,6 +35,7 @@ The framework resolves a DAG from each plugin's `Produces` and `Consumes` declar
 - `inflight-load-producer` **optionally** consumes `PrefixCacheMatchInfo` from an approx or precise prefix producer; prefix-discounting is applied automatically when the attribute is present.
 - `p2p-source-producer` **requires** `PrefixCacheMatchInfo` from a prefix producer; set `prefixMatchInfoProducerName` to select a non-default producer instance. Omitting it binds the default key, which auto-wires the approximate producer (no error) — set it explicitly for precise-only deployments. Set `prefillProfileName` to match a renamed `disagg-profile-handler` prefill profile.
 - `predicted-latency-producer` **optionally** consumes `PrefixCacheMatchInfo`; set `prefixMatchInfoProducerName` in its config to the name of the prefix producer instance.
+- `latency-observer-producer` **requires** `InFlightLoad`, so `inflight-load-producer` is ordered ahead of it and auto-created when absent. It must itself be listed under `dataLayer.sources`, which is what drives its recompute; auto-creation alone leaves every endpoint reading cold.
```
elevran: This line is misleading and contradicts latencyobserver/README.md's own "Required" callout. `RegisterAsDefaultProducer` only wires the attribute into the DAG - it does not add the producer to `dataLayer.sources`, so a scorer-only config never ticks it and every endpoint reads as cold. `latency-observer-producer` must be listed under `dataLayer.sources` explicitly.

```suggestion
- `latency-observer-producer` **requires** `InFlightLoad`, so `inflight-load-producer` is ordered ahead of it and auto-created when absent. It must itself be listed under `dataLayer.sources`; auto-creation from `latency-observation-scorer`'s required data key only wires the attribute, not the periodic tick that publishes it. See the [producer README](latencyobserver/README.md#configuration).
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/latencyobserver/hooks.go
```diff
@@ -0,0 +1,175 @@
...
+		endpoint.Put(p.inflightAtDispatchDataKey, &attrconcurrency.InFlightLoad{
+			Requests: readInFlightRequests(endpoint, p.inFlightLoadDataKey),
+		})
+	}
+	return nil
+}
+
+// readInFlightRequests returns the endpoint's request count under key, or zero
+// when the attribute is absent or malformed.
+func readInFlightRequests(endpoint fwksched.Endpoint, key fwkplugin.DataKey) int64 {
```
elevran, line 103: `readInFlightRequests` defaults a missing/malformed attribute to 0 silently. Under `DataParallelProfileHandler.ProcessResults`, each target endpoint is rebuilt with a fresh, empty attribute map before `PreRequest` runs, so `Produce`'s dispatch-time in-flight write never survives to the read in `PreRequest`. Every DP-routed request records dispatch-time load as 0, silently corrupting calibration with no error or log.

Consider carrying the dispatch-time load through `PluginState` keyed by `request.RequestID`, the same mechanism already used for the `PreRequest` -> `ResponseBody` handoff, instead of relying on endpoint-object identity surviving the profile handler. 
Or we can take a look at `DataParallelProfileHandler.ProcessResults` and align its beahvior. Worth an issue to determine need there, regardless of how you decide to handle it here.

pkg/epp/framework/plugins/requestcontrol/dataproducer/latencyobserver/tracker.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+	}
+
+	sort.Slice(out, func(i, j int) bool { return out[i].value < out[j].value })
+	return out
+}
+
+// percentileOf returns the p-th percentile of a value-sorted slice by linear
+// interpolation. p is a fraction, e.g. 0.25 for P25.
+func percentileOf(sorted []observation, p float64) float64 {
```
elevran, line 82: No direct unit tests for this package's pure math (percentile computation, sliding window, bucket-history floor in flush.go). It's trivially testable in isolation and the sibling `inflightload` package tests its own math this way.

pkg/epp/framework/plugins/requestcontrol/dataproducer/latencyobserver/plugin_test.go
```diff
@@ -0,0 +1,229 @@
...
+	t.Run("tolerates events without an endpoint", func(t *testing.T) {
+		p := newObserver(t)
+		require.NoError(t, p.Extract(ctx, fwkdl.EndpointEvent{Type: fwkdl.EventAddOrUpdate}))
+		require.NoError(t, p.Extract(ctx, fwkdl.EndpointEvent{Type: fwkdl.EventDelete}))
+	})
+}
+
+// Concurrent callers must share one state instance and lose no observations.
+// Run under -race to also catch unsynchronised access.
+func TestConcurrentRecording(t *testing.T) {
```
elevran, line 178: `TestConcurrentRecording` only races `record()` against itself, never against `publish`/`flush` - the exact race the atomic-swap design in flush.go exists to handle. Also `Config.resolve()`'s six validation branches (duration parsing, window/max/min bounds, percentile ordering) have no test coverage.

pkg/epp/framework/plugins/scheduling/scorer/latencyobservation/plugin.go
```diff
@@ -0,0 +1,330 @@
...
+
+	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+	fwkplugin "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+	fwksched "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
+	attrconcurrency "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/attribute/concurrency"
+	attrlatency "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/attribute/latency"
+)
+
+// ScorerType is the plugin type of the latency-observation scorer.
+const ScorerType = "latency-observation-scorer"
```
elevran: nit: the README notes this pair is intended for EPP hub deployments routing across peer clusters, not for selecting model-serving endpoints directly. If that's the only intended use, consider reflecting that in the plugin type name (e.g., add `-hub` or `-multicluster`) so it's clear from configuration alone rather than only from the README.

pkg/epp/framework/plugins/requestcontrol/dataproducer/latencyobserver/README.md
```diff
@@ -0,0 +1,168 @@
...
+## Architecture
+
+The data producer adopts an observer pattern, passively collecting metrics about the request/response
+traffic. It captures latency in a sliding window and computes window-based percentiles for the
+scorer.
+It builds on the request-control hooks and publishes statistics as a datalayer attribute:
+
+| hook | what it does |
+|---|---|
+| `Produce` | Update inflight load at the time of current request |
```
elevran: nit: the hook table describes `Produce` as "Update inflight load", but it only reads and pins the existing in-flight load at dispatch time - it doesn't update it.

elevran (commented):
Thanks for contributing - this could be useful where latency and queueing signals from metrics can be unreliblae (e.g., lag).

Several aspects need attention:
The observer's own README already documents the dataLayer.sources requirement correctly (see its "Required" callout and Configuration section). The one place that contradicts it is the top-level dataproducer README, flagged inline below and the contradiction must be resolved.
Other feedback inlined.

pkg/epp/framework/plugins/datalayer/attribute/latency/ttft_percentiles.go
elevran: @vMaroon - have seen these on several PRs. New code should only have llm-d. Code migrated from GIE should have both.
Worthwhile fixing this post release (can be scripted). WDYT?

pkg/epp/framework/plugins/requestcontrol/dataproducer/latencyobserver/hooks.go
Mohammad-nassar10: Went with `PluginState` keyed by request ID so the reading no longer depends on endpoint identity surviving the handler. Also, I can open an issue for `ProcessResults` passing nil attributes.

pkg/epp/framework/plugins/requestcontrol/dataproducer/latencyobserver/tracker.go
Mohammad-nassar10: Added `tracker_test.go` covering this package's functionality.

pkg/epp/framework/plugins/requestcontrol/dataproducer/latencyobserver/plugin_test.go
Mohammad-nassar10: Added `TestConcurrentRecordAndPublish` which races `record` against `publish`, and `resolve()`'s validation branches are now covered in `flush_test.go`.

pkg/epp/framework/plugins/requestcontrol/dataproducer/latencyobserver/README.md
Mohammad-nassar10: Fixed, thanks!

pkg/epp/framework/plugins/requestcontrol/dataproducer/README.md
Mohammad-nassar10: Fixed, thanks!

pkg/epp/framework/plugins/scheduling/scorer/latencyobservation/plugin.go
Mohammad-nassar10: It is intended for EPP hub but it also works wherever an endpoint answers requests, so it can be generalized to other uses when added to the router, so I'd rather keep the name general.

pkg/epp/framework/plugins/requestcontrol/dataproducer/latencyobserver/hooks.go
elevran: thanks - please open an issue to determine if `DataParallelProfileHandler.ProcessResults` passing a target endpoint is rebuilt with a fresh, empty attribute map before `PreRequest` runs is reasonable.

pkg/epp/framework/plugins/scheduling/scorer/latencyobservation/plugin.go
elevran: that's fine. If meant as a generic latency predictor than @vMaroon request above (comparison with current latencypredictor sidecar) should be addressed. @kaushikmitr any thoughts on this PR?

Mohammad-nassar10: I meant to keep our scope here for multicluster, hub deployments Then in the future to expand it to other use cases. So we can't compare with pod's scorers but to compare it with the existing scorers intended to multicluster such as `multicluster-queue-scorer`. Will add `-hub` to make it clearer.

elevran:
approved from my point of view.
@vMaroon @hexfusion ny feedback needed before we merge?

Mohammad-nassar10:
Thanks @elevran.
@vMaroon @hexfusion, all earlier comments are addressed, is there anything else you'd like changed, or is this good to merge? Thanks.

ahg-g:
This is interesting, but I think a comprehensive evaluation is important here, how this compares to current llm-d default that uses inflight load with prefix affinity, also how it compares to predicted latency. This is a fairly complex plugin and I am looking for evidence that it is worth maintaining upstream.

Mohammad-nassar10:
> This is interesting, but I think a comprehensive evaluation is important here, how this compares to current llm-d default that uses inflight load with prefix affinity, also how it compares to predicted latency. This is a fairly complex plugin and I am looking for evidence that it is worth maintaining upstream.

Here's a comparison [analysis document](https://docs.google.com/document/d/1p2dpOegfvoyBEGiF9h1v3dGdJdJE-HNEku1XNJdKBII/edit?tab=t.0) with every latency-relevant EPP scorer, with evaluation results and possible complementary scorers.

ahg-g (approved):

PR #2372, opened 2026-08-13
[disagg-sidecar] Fix cached token usage for non-streaming P/D responses

### LINKED ISSUE #2622: Cached token usage is not rewritten for split non-streaming responses
**What happened**:

When disaggregated prefill/decode is enabled, the sidecar proxy rewrites `usage.prompt_tokens_details.cached_tokens` using the cached token count reported by the prefiller.

For a non-streaming request, `httputil.ReverseProxy` may split the decoder's JSON response across multiple `Write` calls. The cached-token response writer attempts to parse and rewrite each incomplete chunk separately. JSON parsing fails, and the response may retain the decoder's original `cached_tokens` value instead of reporting the prefiller's cached token count.

The proxy should buffer the complete non-streaming JSON response and consistently set `usage.prompt_tokens_details.cached_tokens` to the cached token count reported by the prefiller.

Streaming responses should continue to be processed incrementally as SSE.

1. Enable disaggregated prefill/decode with the NIXL v2 connector.
2. Send an OpenAI-compatible chat completion request with `"stream": false`.
3. Use a request where the prefiller reports cached tokens.
4. Have the decoder's JSON response delivered across multiple `Write` calls. For example:

   First write:

   ```json
   {"usage":{"prompt_tokens":64,
   ```

   Second write:

   ```json
   "prompt_tokens_details":{"cached_tokens":49}}}
   ```

5. Assume the prefiller reported `7` cached tokens.
6. Inspect the final response.

The response may report `49` cached tokens because neither incomplete JSON chunk can be rewritten independently.

A deterministic unit reproducer can wrap an `httptest.ResponseRecorder`, write the two chunks separately, finalize the response writer, and inspect the resulting JSON.

The response handling mode should be selected from the request's `stream` setting. Inferring the mode from the response `Content-Type` or the first response chunk can misclassify non-streaming JSON responses.

The non-streaming writer needs to buffer data until it has a complete JSON document. Any remaining incomplete body should be forwarded unchanged during finalization.

**Environment**:
- Install tools:
- Others:

```

### PR DESCRIPTION
**Pull request by learner0810**
**What type of PR is this?**
/kind bug

Fixes incorrect `cached_tokens` reporting for non-streaming P/D requests.

A non-streaming JSON response may be delivered through multiple `ResponseWriter.Write` calls. Previously, each chunk was parsed independently, so partial JSON could not be rewritten and the decoder's `cached_tokens` value was returned unchanged.

This change:

- Buffers non-streaming response chunks until a complete JSON response is available.
- Rewrites `usage.prompt_tokens_details.cached_tokens` using the prefiller's value.
- Uses the request's explicit streaming mode instead of inferring it from the response.
- Preserves the existing streaming response behavior.

Fixes  https://github.com/llm-d/llm-d-router/issues/2622

```release-note
Fix cached token usage reporting for non-streaming P/D inference requests.
```

### DIFF
--- a/pkg/sidecar/proxy/cached_tokens_usage_rewriter.go
+++ b/pkg/sidecar/proxy/cached_tokens_usage_rewriter.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"io"
 	"net/http"
-	"strings"
 
 	"github.com/felixge/httpsnoop"
 )
@@ -32,6 +31,7 @@ type cachedTokensUsageRewriter struct {
 	wroteHeader  bool
 	streaming    bool
 	streamBuffer []byte
+	jsonBuffer   []byte
 }
 
 // OpenAI-compatible chat usage reports prompt cache hits at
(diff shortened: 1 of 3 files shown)


roytman:
One process nit: could you link the issue this fixes (Fixes #)?  
/lgtm

learner0810:
> One process nit: could you link the issue this fixes (Fixes #)? /lgtm


Done, I’ve added Fixes https://github.com/llm-d/llm-d-router/issues/2622   to the PR description. Thanks for the reminder.

roytman (approved):

PR #2374, opened 2026-08-13
fix(chart): avoid inferring GMP from GKE gateway provider

Alex-ai-future:
/kind bug

The chart makes two independent decisions: which Gateway implementation to use and which monitoring backend should scrape EPP metrics. However, when `provider.name=gke` is selected and no monitoring provider is configured, the chart silently infers GMP.

This causes a silent monitoring failure for GKE clusters using Prometheus Operator or unauthenticated scraping. With `prometheus.enabled=true` and `auth.enabled=false`, the inferred GMP provider skips `ServiceMonitor`, while the GMP template skips `PodMonitoring` because authentication is disabled. The chart renders successfully, but no monitoring resource is created.

This change keeps the monitoring default as `prometheusoperator` independently of the Gateway provider. It provides the following user-facing benefits:

- GKE Gateway users can use Prometheus Operator monitoring without an extra provider override.
- Enabling Prometheus metrics no longer silently produces no scrape resource in the default configuration.
- Selecting a GKE Gateway no longer implicitly creates GMP-specific monitoring resources or RBAC.
- GMP remains supported through the explicit `router.monitoring.provider.name=gmp` setting.

Users relying on the previous implicit GMP behavior must set the monitoring provider explicitly.

No linked issue.

**Test plan**:

- [x] Verified GKE Gateway with an unset monitoring provider renders a `ServiceMonitor` and no `PodMonitoring`.
- [x] Verified explicit GMP monitoring renders `PodMonitoring` and no `ServiceMonitor`.
- [x] Ran Helm dependency builds, chart lint, shell syntax validation, and the focused Helm rendering assertions.
- [ ] `make presubmit` could not run because the current environment does not provide `make`, Go, or a container runtime.

**Release note**:
```release-note
Router Helm charts now default to Prometheus Operator monitoring independently of the selected Gateway provider. Users relying on implicit GMP monitoring must configure the GMP provider explicitly.
```


ahg-g:
Thanks, was this tested on gke?

Alex-ai-future:
> Thanks, was this tested on gke?

This PR fixes a deterministic Helm rendering issue. The GKE gateway provider and monitoring provider are independent settings, so the chart-level behavior can be verified with Helm rendering tests. I have not run a live GKE scrape test; the remaining validation would be an end-to-end monitoring smoke test on GKE.

ahg-g (approved):

PR #2378, opened 2026-08-13
test(e2e): warm up before disruption in streaming-request spec

### LINKED ISSUE #2377: [flake] e2e: streaming-disruption spec flakes on Envoy active-health-check race
**Symptom**

`test/e2e/disruption_test.go`, spec "A decode pod is killed while a streaming request is in-flight / should not hang and should recover routing" intermittently fails with:

```
[FAILED] streaming request failed to connect
Expected
    <string>:
not to be empty
In [It] at: /app/test/e2e/disruption_test.go:179
```

Failure latency in the observed CI run was ~1ms after the request was sent - consistent with a local Envoy reply, not a real network timeout. A rerun of the same job passed with no code changes, confirming a timing-dependent flake rather than a deterministic failure.

**Root cause**

Every other spec in `disruption_test.go` sends a warm-up request via `runCompletion`/`tryCompletion` (`test/e2e/requests_test.go`) before its disruptive action. Those helpers go through the `openai-go` SDK client, which retries on connection errors/5xx by default.

This specific spec skips that warm-up and sends its first request via `sendStreamingCompletion`, which does a single `disruptionClient.Do(req)` on a plain `http.Client` with no retry (`test/e2e/disruption_test.go:370-389`).

The fix is to add the same pre-disruption warm-up (`runCompletion`) used by the sibling "mid-request" spec and by "All pods are gone" / "The EPP is killed" specs, before starting the streaming goroutine in the "streaming request is in-flight" spec. This closes the race by construction instead of adding ad hoc retry logic to the raw streaming client.

elevran:
**What type of PR is this?**

/kind test

The "A decode pod is killed while a streaming request is in-flight" e2e spec sent its first request through a raw, non-retrying client, unlike its sibling disruption specs, which all warm up via `runCompletion` first. That first request can race Envoy's initial active health check on the EPP's ext_proc cluster (10s interval, see #1961), producing a local "no healthy upstream" reply that the spec misreports as "streaming request failed to connect."

Adds the same pre-disruption warm-up used by the other disruption specs, closing the race by construction instead of adding retry logic to the raw streaming client.

Fixes #2377

```release-note
NONE
```

### DIFF
--- a/test/e2e/disruption_test.go
+++ b/test/e2e/disruption_test.go
@@ -167,6 +167,10 @@ var _ = ginkgo.Describe("Disruption tests", func() {
 			gomega.Expect(prefillPods).Should(gomega.BeEmpty())
 			gomega.Expect(decodePods).Should(gomega.HaveLen(2))
 
+			ginkgo.By("Verifying requests route successfully before disruption")
+			nsHdr, _, _ := runCompletion(simplePrompt, simModelName)
+			gomega.Expect(nsHdr).Should(gomega.Equal(nsName))
+
 			ginkgo.By("Starting a streaming request")
 			connected := make(chan string, 1)
 			errCh := make(chan error, 1)


elevran:
@ahg-g can you please take a look? 4 line fix for an e2e flake.

ahg-g (approved):

PR #2382, opened 2026-08-13
fix(flowcontrol): prevent aggregate stats underflow on flow GC

### LINKED ISSUE #2381: Flow control aggregate stats underflow when a non-empty queue is deregistered while a stale handle drains it
**What happened**:

`FlowRegistry.Stats()` can report `TotalLen: 18446744073709551615` after a flow is garbage-collected while a cleanup-sweep worker still holds a `ManagedQueue` handle to the flow's queue.

`deleteFlow` deducts a non-empty queue's stats from the aggregates before dropping the queue from the band (`pkg/epp/flowcontrol/registry/registry_helpers.go:178-185`):

```go
if mq, ok := band.queues[key.ID]; ok && mq != nil {
    // Safe-guard: Deduct any unswept capacity before destroying the queue
    if mqLen := int64(mq.Len()); mqLen > 0 {
        fr.propagateStatsDelta(key.Priority, -mqLen, -int64(mq.ByteSize()))
    }
}
delete(band.queues, key.ID)
```

The deduction leaves the underlying `SafeQueue` full, and the `managedQueue` object survives as long as anything holds a reference to it. The cleanup sweep holds such a reference: `processAllQueuesConcurrently` resolves every `ManagedQueue` handle in phase 1 and processes the handles in phase 3, with no registry lock held in between (`pkg/epp/flowcontrol/controller/internal/processor.go:645`). If `deleteFlow` runs between the two phases, the worker later drains a queue whose contents have already been deducted.

When the worker calls `Cleanup`, `Drain`, or `Remove` on the stale handle, `applyAndPropagateLocked` measures a negative delta from the queue's own before/after state and propagates it. The aggregates are decremented twice for the same items. `Stats()` reads the aggregates as `uint64`, so the second decrement wraps.

The `managedQueue` doc comment states that aggregates are sums of deltas measured across queue mutations inside one critical section and therefore cannot go negative. `deleteFlow`'s deduction is not measured across a mutation, so the guarantee does not hold: the queue's remaining history stays pending in the `SafeQueue` and is applied on top of the deduction.

Introduced by commit a0c8e6d7 (#1128), where `deleteFlow` first gained the stats deduction.

Aggregate stats stay non-negative and equal to the sum of live queue contents. Each item is deducted from the aggregates once, whether it leaves through a sweep or through `deleteFlow`.

The interleaving is sequential; no goroutine timing is required.

1. Register a flow and `Add` one item (100 bytes).
2. Resolve a `ManagedQueue` handle for the flow, as the sweep's phase 1 does.
3. Call `fr.deleteFlow(key)` under the registry write lock, as GC does.
4. Call `Cleanup`, `Drain`, or `Remove` on the handle from step 2, as the sweep's phase 3 or a
   request finalization does.
5. Read `fr.Stats()`.

Observed against `main` @ 800ec0e4:

```
after add:        totalLen=1                    totalBytes=100
after deleteFlow: totalLen=0                    totalBytes=0
after staledrain: totalLen=18446744073709551615 totalBytes=18446744073709551516
```

In production the trigger is GC of a flow that still holds finalized-but-unswept items, concurrent
with a sweep pass. The sweep and the GC both run on timers, so the probability accumulates over
process lifetime.

Consequences of one occurrence:

- The corruption is permanent for the process lifetime. The aggregates end up offset below true usage by the deleted queue's contents, and nothing restores them.
- `hasCapacity` consumes the wrapped values. While the underlying `int64` aggregate is negative, admission is erratic: `TotalByteSize+itemByteSize` either stays near `MaxUint64` and rejects the request or wraps past zero and admits it, depending on the item size. Once real load exceeds the deficit, the aggregates turn positive but understate usage, and capacity checks overcommit.
- The `flow_control_*` gauges report values near `MaxUint64`.
- The only signal is a DEBUG-level log line in `deleteFlow`.

Fix: have `deleteFlow` drain the queue through the `managedQueue` wrapper instead of hand-rolling the delta. The deduction is then measured across a real mutation, and a stale handle observes an empty queue and propagates nothing. A PR with the fix and a regression test covering the `Drain`, `Cleanup`, and `Remove` interleavings follows.

**Environment**:
- Kubernetes version (use `kubectl version`): n/a, unit-level, no cluster required
- llm-d-router version: `main` @ 800ec0e4
- Install tools: n/a
- Others: reproduced with `go test -race ./pkg/epp/flowcontrol/registry/...`

LukeAVanDrie:
**What type of PR is this?**

/kind bug

`deleteFlow` deducts a non-empty queue's stats from the registry aggregates without emptying the underlying `SafeQueue`. The cleanup sweep resolves `ManagedQueue` handles before processing them, with no registry lock held in between, so a handle resolved before the GC can drain the queue afterward and propagate the same deduction a second time. The `int64` aggregates go negative and the `uint64` casts in `Stats()` wrap to values near `MaxUint64`, corrupting capacity checks and the `flow_control_*` gauges for the process lifetime.

The fix drains the queue through the `managedQueue` wrapper instead of hand-rolling the delta. The deduction is then measured across the actual mutation inside the queue's critical section, and a stale handle observes an empty queue and propagates nothing.

Fixes #2381 

```release-note
NONE
```

### DIFF
--- a/pkg/epp/flowcontrol/registry/registry_helpers.go
+++ b/pkg/epp/flowcontrol/registry/registry_helpers.go
@@ -170,17 +170,18 @@ func (fr *FlowRegistry) deleteFlow(key flowcontrol.FlowKey) {
 	fr.logger.V(logging.DEBUG).Info("Deleting queue instance.", "flowKey", key)
 	if val, ok := fr.priorityBands.Load(key.Priority); ok {
 		band := val.(*priorityBand)
-		// Requests in a queue that are asynchronously finalized (e.g., due to client
-		// stream cancellation or context timeout), they are left in the queue for the
-		// GC process to clean them up, including updating the capacity. Here we remove
-		// a flow queue, potentially with such requests waiting for GC, therefor the
-		// capacity stats are updated here before removing the queue.
+		// Requests that are asynchronously finalized (e.g., due to client stream
+		// cancellation or context timeout) are left in the queue for the cleanup sweep.
+		// A queue deleted here may still hold such items, and the sweep may still hold a
+		// ManagedQueue handle to it (handles are resolved before processing, without
+		// registry locks). Draining through the wrapper both empties the queue and
+		// deducts the stats in one critical section, so a later mutation through a stale
+		// handle observes an empty queue and propagates nothing.
 		if mq, ok := band.queues[key.ID]; ok && mq != nil {
-			// Safe-guard: Deduct any unswept capacity before destroying the queue
-			if mqLen := int64(mq.Len()); mqLen > 0 {
-				fr.logger.V(logging.DEBUG).Info("Deregistering non-empty queue during GC, flushing stats",
-					"flowKey", key, "unsweptCount", mqLen)
-				fr.propagateStatsDelta(key.Priority, -mqLen, -int64(mq.ByteSize()))
+			if mq.Len() > 0 {
+				fr.logger.V(logging.DEBUG).Info("Deregistering non-empty queue during GC, draining unswept items",
+					"flowKey", key, "unsweptCount", mq.Len())
+				mq.Drain()
 			}
 		}
 		delete(band.queues, key.ID)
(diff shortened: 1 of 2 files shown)


RishabhSaini (approved):
LGTM, thanks for fixing this quickly. The fix is clean draining through the wrapper instead of hand-rolling the delta closes the double-deduction race. Test covers all three stale-handle paths.

Issue #79, opened 2025-05-13
CI: skip integration/e2e tests when no code has changed


shaneutt:
Right now if you change documentation like the `README.md`, or anything like that it will still trigger the full test suite including integration tests. To reduce waste, we should have those tests run in CI only when relevant changes have occurred.

elevran:
@shaneutt - can GH actions be configured based on file types in change set?

shaneutt:
In Kubernetes SIGs we occasionally use [dorny/paths-filter](https://github.com/dorny/paths-filter), example [here](https://github.com/kubernetes-sigs/blixt/blob/07586866a65e37742dcc4f3cfca7a5cd937d731b/.github/workflows/build.yaml#L58).

elevran:
Fixed in GH action (and changed in #570 from negative filter to positive - run when source has been changed)

PR #2383, opened 2026-08-13
feat(predictedlatency): include selected prediction in training entries

### LINKED ISSUE #2380: [Latency Predictor] Preserve selected prediction with matching actual TTFT/TPOT observation

Include the selected latency prediction together with the matching actual latency in the router's training data.

The router already has the prediction used for the selected endpoint and records the actual latency after the request completes. The missing part is sending the selected prediction with the actual value so the latency predictor can calculate prediction error and violation rate.

For P/D:

- selected prefill prediction with actual TTFT;
- selected decode prediction with actual TPOT.

For monolithic serving:

- selected endpoint prediction with the actual request latency.

Only the selected endpoint prediction should be sent, not predictions for all candidates.

**Why is this needed**:

[llm-d-latency-predictor#40](https://github.com/llm-d/llm-d-latency-predictor/issues/40) and [PR #79](https://github.com/llm-d/llm-d-latency-predictor/pull/79) use predicted and actual latency to detect model drift and trigger retraining. Today, the router sends the actual latency but does not include the matching prediction in the training entry, so the predictor cannot calculate the live prediction error.

The change should keep existing training behavior working when a prediction is unavailable.

**Acceptance criteria**:

- Add the selected prediction to the training entry using the field names expected by the latency predictor: `predicted_ttft_ms` and `predicted_tpot_ms`.
- Pair the prediction with the actual latency from the same endpoint and serving stage.
- Do not send all candidate predictions with one actual value.
- Preserve the current TPOT aggregation semantics.
- Add tests for monolithic, P/D prefill, P/D decode, and missing-prediction cases.
- Verify the serialized training entry contains the matching predicted and actual values.

madhugoutham:
**What type of PR is this?**
/kind feature

The latency predictor needs each actual TTFT/TPOT paired with the prediction used for the selected endpoint so it can compute live prediction error and trigger drift-based retraining ([PR #79](https://github.com/llm-d/llm-d-latency-predictor/pull/79)).

Today the router already stores the selected prediction in request context and measures actuals, but training entries only sent actuals. This PR adds optional `predicted_ttft_ms` / `predicted_tpot_ms` on `TrainingEntry` and populates them from the selected endpoint/stage:

- monolithic / decode: selected prediction with matching actual
- P/D: prefill prediction with actual TTFT; decode prediction with actual TPOT
- missing or non-finite prediction: field omitted (not sent as `0`)
- legacy entries without predicted fields remain valid for supervised training

No new prediction or training HTTP call is added on the request hot path; values are copied into the existing buffered training flush.

Fixes #2380

**Test plan**:
- [x] Selected-endpoint lookup then TTFT training includes `predicted_ttft_ms`
- [x] P/D prefill TTFT uses prefill prediction (not decode)
- [x] P/D decode TPOT uses selected decode prediction (not prefill / other candidate)
- [x] Missing prediction omits predicted fields; actuals still train
- [x] Serialized JSON asserts exact field names and omitempty behavior
- [x] SLO-invalid selected prediction still included for live drift
- [x] Non-finite predicted values rejected / omitted
- [x] `go test ./pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/... -count=1`
- [x] Local `/add_training_data_bulk` dump showed `predicted_ttft_ms` with matching `actual_ttft_ms` (HTTP 202)

**Release note**:
```release-note
Include selected predicted TTFT/TPOT in latency-predictor training samples for live drift detection.

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/requestcontrol_hooks.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/requestcontrol_hooks.go
@@ -201,6 +201,9 @@ func (pl *PredictedLatency) ResponseBody(ctx context.Context, request *fwksched.
 				entry.PrefillTokensInFlight = predictedLatencyCtx.prefillTokensAtDispatch
 				entry.DecodeTokensInFlight = predictedLatencyCtx.decodeTokensAtDispatch
 				entry.NumRequestRunning = predictedLatencyCtx.requestsAtDispatch
+				// Request-average actual TPOT is paired with the scheduling-time
+				// selected decode/primary prediction (avgPredictedTPOT).
+				entry.PredictedTPOT = predictedLatencyMSPtr(predictedLatencyCtx.avgPredictedTPOT)
 				if err := pl.latencypredictor.AddTrainingDataBulk([]latencypredictor.TrainingEntry{entry}); err != nil {
 					logger.V(logutil.DEBUG).Error(err, "record TPOT training failed")
 				}
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/requestcontrol_hooks.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/requestcontrol_hooks.go
@@ -197,13 +197,13 @@ func (pl *PredictedLatency) ResponseBody(ctx context.Context, request *fwksched.
 					0,
 					0,
 					0,
+					trainingEntryOptions{PredictedTPOT: predictedLatencyMSPtr(predictedLatencyCtx.avgPredictedTPOT)},
 				)
 				entry.PrefillTokensInFlight = predictedLatencyCtx.prefillTokensAtDispatch
 				entry.DecodeTokensInFlight = predictedLatencyCtx.decodeTokensAtDispatch
 				entry.NumRequestRunning = predictedLatencyCtx.requestsAtDispatch
 				// Request-average actual TPOT is paired with the scheduling-time
 				// selected decode/primary prediction (avgPredictedTPOT).
-				entry.PredictedTPOT = predictedLatencyMSPtr(predictedLatencyCtx.avgPredictedTPOT)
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by madhugoutham** [comment]
PTAL @kaushikmitr

File: pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/requestcontrol_hooks.go [hunk 1]
```diff
@@ -201,6 +201,9 @@ func (pl *PredictedLatency) ResponseBody(ctx context.Context, request *fwksched.
 				entry.PrefillTokensInFlight = predictedLatencyCtx.prefillTokensAtDispatch
 				entry.DecodeTokensInFlight = predictedLatencyCtx.decodeTokensAtDispatch
 				entry.NumRequestRunning = predictedLatencyCtx.requestsAtDispatch
+				// Request-average actual TPOT is paired with the scheduling-time
+				// selected decode/primary prediction (avgPredictedTPOT).
```
kaushikmitr, line 206: did you test this? I am wondering why buildTrainingEntry is not updated

madhugoutham: Yes, I tested it.

```json
{
  "predicted_tpot_ms": 18.0,
  "actual_tpot_ms": 25.0
}
```

`buildTrainingEntry()` currently creates the common measured-latency fields, while the selected prediction is added afterward from the scheduling context. I agree that moving the optional prediction fields into `buildTrainingEntry()` would make the pairing clearer and safer, so I’ll refactor it accordingly.

madhugoutham: @kaushikmitr moved PredictedTTFT and PredictedTPOT into buildTrainingEntry through trainingEntryOptions. This keeps the selected prediction paired with the matching actual TTFT/TPOT and preserves omission when no prediction is available. Added coverage for selected endpoint/P-D pairing, missing predictions, JSON serialization and non-finite values.

ahg-g:
/assign @kaushikmitr 

@kaushikmitr pls take a look

Issue #2358, opened 2026-08-11
Support enabling EPP feature gates without duplicating the full plugin config


LukeAVanDrie:
**What would you like to be added**:

A way to enable an EPP feature gate (for example `flowControl`) on top of an existing plugin configuration without copying the entire `EndpointPickerConfig` document. This issue proposes (a), a `--feature-gates` flag, as the primary direction, and (b), structured plugin config in the chart, as a complementary chart-side option.

**(a) A `--feature-gates` EPP flag**

Kubelet-style: `--feature-gates=flowControl=true`.

- `loadFeatureConfig` (`pkg/epp/config/loader/configloader.go`) parses kubelet-style `name=bool` entries and defaults bare names to `true`. The flag can feed the same parser.
- Later `featureGates` entries override earlier ones. Appending flag-supplied entries to `rawConfig.FeatureGates` in `Runner.parseConfigurationPhaseOne` (`cmd/epp/runner/runner.go`), after the config file is read, makes flag entries win over file entries.
- The flag definition belongs in `pkg/epp/server/options.go`, alongside `--config-file`.
- No chart change is needed for Helm delivery: `routerlib` renders arbitrary EPP flags from `router.epp.flags`.

**(b) Structured plugin config in the chart**

Add `router.epp.pluginsConfig` (a structured map) to `routerlib`, rendered with `toYaml` into the ConfigMap (`config/charts/routerlib/templates/_config.yaml`) alongside the existing `pluginsCustomConfig`. Helm deep-merges maps across `-f` files, so an overlay can set `featureGates: [flowControl]` and inherit `plugins` and `schedulingProfiles` from an earlier values file. Limits: Helm replaces lists as a unit, so plugins cannot be merged per-entry; YAML comments are lost in the `toYaml` round trip; non-Helm deployments are unaffected.

**Why is this needed**:

Enabling a feature gate requires owning the whole config document via `pluginsCustomConfig`, which the chart splices into the ConfigMap as a single string. Every llm-d guide that enables `flowControl` carries a full copy of the optimized-baseline plugin list, and the copies drift: the keda-epp-queue guide's copy no longer matches optimized-baseline's plugin set, and llm-d/llm-d#2248 tracks a guide that broke while working around this gap. Reviewers on llm-d/llm-d#2229 asked for an overlay that adds only the gate. Helm values layering cannot produce one, because layering replaces string values and the whole config is one string.

LukeAVanDrie:
[...]

PR #2388, opened 2026-08-14
feat(epp): add --feature-gates flag

abhishekkarki:
**What type of PR is this?**
/kind feature

Enabling an EPP feature gate currently requires owning the entire `EndpointPickerConfig` document. The chart splices it into the ConfigMap as a single string, and Helm replaces strings rather than merging them, so an overlay that wants to add only a gate has to restate the whole plugin list. Those copies drift — llm-d/llm-d#2252 worked around this with a `kubectl patch` that rewrites the whole `plugins` key, and its own reviewer note warns that anything extra in a deployment gets silently dropped.

This adds `--feature-gates`, taking kubelet-style `name=bool` entries (a bare name enables the gate). Entries are appended to `rawConfig.FeatureGates` in `LoadRawConfig` after the config is decoded and before `validateFeatureGates`, so `loadFeatureConfig`'s last-wins ordering makes flag entries override file entries, and a typo'd gate from the flag is rejected exactly like one from the file.

Appending to the struct rather than the map `LoadRawConfig` returns is load-bearing: `InstantiateAndConfigure` and `validateConfig` each re-derive gates from `rawConfig.FeatureGates`. Patching only the returned map would give `r.featureGates[flowControl] == true` alongside a nil `FlowControlConfig`, and `initAdmissionControl` would then nil-deref `.Registry`.

Verified on a kind cluster by patching only the EPP args:

| | before | after |
|---|---|---|
| ConfigMap `featureGates` | absent | absent |
| EPP args | — | `--feature-gates=flowControl=true` |
| Startup log | `Flow Control layer is disabled via the flowControl feature gate` | `Initializing Flow Control layer` |

Two notes for reviewers: slice flags render as `{}` in the "Flags processed" log line, which is pre-existing and matches `tls-cipher-suites` and `endpoint-target-ports`. And `make vulncheck` fails identically on clean `main` (Go 1.26.5 stdlib, fixed in 1.26.6).

Refs llm-d/llm-d-router#2358 — part (a) only; @hugosmoreira is taking (b).

```release-note
Adds a `--feature-gates` EPP flag taking kubelet-style `name=bool` entries (a bare name enables the gate). Entries are applied after the configuration file's `featureGates` list and override it, so a deployment can enable a gate such as `flowControl` without restating the whole `EndpointPickerConfig`.
```

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -193,6 +193,10 @@ If the configuration is in a file, the EPP command line argument `--config-file`
  to specify the full path of the file in question. If the configuration is passed as in-line
  text the EPP command line argument `--config-text` should be used.
 
+Feature gates can also be set with the `--feature-gates` command line argument, which takes a 
+ comma-separated list of kubelet-style `name=bool` entries; a bare name enables the gate. These
+ entries are applied after the configuration's own `featureGates` list, so they override it.
+
 ### Default plugins
 
 The EPP injects these plugins when they are absent, so a configuration does not need to list them. Some
@@ -211,7 +215,8 @@ RequestHandler:
 
 FlowControl:
 - The flow control admission layer itself is off by default; enable it with
-  `featureGates: ["flowControl"]`.
+  `featureGates: ["flowControl"]` in the configuration, or with
+  `--feature-gates=flowControl=true` on the command line.
 - `fcfs-ordering-policy`, `global-strict-fairness-policy`, and `static-usage-limit-policy` are configured when absent.
 - `utilization-detector` is configured as the saturation detector when none is set.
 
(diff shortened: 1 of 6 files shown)


docs/architecture.md
```diff
@@ -211,7 +215,8 @@ RequestHandler:
 
 FlowControl:
 - The flow control admission layer itself is off by default; enable it with
-  `featureGates: ["flowControl"]`.
+  `featureGates: ["flowControl"]` in the configuration, or with
+  `--feature-gate=flowControl=true` on the command line.
```
GolfRider: Could we change it to --feature-gates=flowControl=true so the documented command works?

GolfRider:
LGTM.

Tested locally on macOS arm64 with Go 1.26.6.  Focused unit and race tests passed. 
I also verified config override behavior, explicit enable/disable, comma-separated gates, rejection of unknown gates, and successful EPP startup in file-discovery mode. 
The implementation behaved correctly in the tested paths.

docs/architecture.md
abhishekkarki: Good catch, fixed in the latest push. Thanks for testing.

LukeAVanDrie (approved):
Thanks, this LGTM!

ahg-g (approved):

PR #2389, opened 2026-08-15
fix(tokenizer): translate Anthropic tool calls and thinking for render

mschulist:
**What type of PR is this?**
/kind bug

Precise prefix routing was broken for the Anthropic Messages API whenever a conversation used tools or extended thinking. The Messages-to-render translation dropped `tool_use`, `tool_result`, and `thinking` content blocks and forwarded Anthropic-shaped tool definitions (`input_schema`) verbatim, so the tokens computed by the vLLM `/render` backend diverged from what vLLM computes when it serves `/v1/messages` (it converts the request to OpenAI chat shape internally in `vllm/entrypoints/anthropic/serving.py`). The result was prefix-cache misses - or render failures - on tool-using traffic such as Claude Code.

The conversion now mirrors vLLM's own translation so both paths feed the identical chat-template pipeline:

- `tool_use` blocks become `tool_calls` with CPython `json.dumps`-formatted arguments (separators, `ensure_ascii` escaping, preserved wire key order), matching the exact string vLLM renders into the prompt.
- `tool_result` blocks become role-`tool` messages, emitted before the user message that carried them (vLLM's ordering); images in results become their own user message.
- `thinking` blocks become the assistant `reasoning` field; `redacted_thinking` is accepted but contributes no tokens.
- Tool definitions become OpenAI function tools, with `input_schema` passed through as raw JSON so the template's re-serialization keeps the wire key order (a Go map round-trip would alphabetize keys and desynchronize prefix hashes).
- The system prompt joins text blocks with no separator and strips Claude Code's per-request billing header, as vLLM does.
- The estimate backend folds tool/thinking text into its pseudo-token stream so tool traffic no longer collides on one prefix key.
- Outbound render spans are named after the route (`tokenize_render /v1/chat/completions/render`) instead of the generic `HTTP POST`.

Fixes #

```release-note
Fixed prefix-cache-aware routing for the Anthropic Messages API with tool use and extended thinking: the tokenizer now translates tool calls, tool results, and thinking blocks the same way vLLM does when serving /v1/messages, so rendered tokens match the server's and prefix routing hits.
```

**Test plan**:
- [x] Unit tests: Anthropic conversion of tool_use (incl. CPython-style argument formatting), tool_result (ordering, tool_call_id, text joining, image follow-up messages), thinking/reasoning, content omission for tool-only assistant messages, system joining and billing-header stripping, and tool-schema conversion.
- [x] Unit tests: render payload preserves tool-schema key order through RenderChat; multimodal timeout detection for rebuilt payloads; outbound render span naming.
- [x] Unit tests: parser round-trips tool_use/tool_result/thinking blocks with wire bytes intact; estimate backend distinguishes tool traffic in the prefix stream.
- [x] `make presubmit` passes; full EPP unit suite green with `-race`.

### DIFF
--- a/pkg/epp/framework/interface/requesthandling/types.go
+++ b/pkg/epp/framework/interface/requesthandling/types.go
@@ -676,9 +676,19 @@ type MessagesRequest struct {
 	// not a message with role "system".
 	System AnthropicContent `json:"system,omitempty"`
 	// Tools field for tool use capabilities.
-	Tools []any `json:"tools,omitempty"`
+	Tools []AnthropicTool `json:"tools,omitempty"`
 	// CacheSalt isolates prefix caches for security.
 	CacheSalt string `json:"cache_salt,omitempty"`
+}
+
+// AnthropicTool is a tool definition in the Anthropic schema. InputSchema keeps
+// the raw JSON bytes so the wire key order survives downstream re-serialization.
+type AnthropicTool struct {
+	Name         string          `json:"name"`
+	Description  string          `json:"description,omitempty"`
+	InputSchema  json.RawMessage `json:"input_schema,omitempty"`
+	Strict       *bool           `json:"strict,omitempty"`
+	DeferLoading *bool           `json:"defer_loading,omitempty"`
 }
 
 func (r *MessagesRequest) String() string {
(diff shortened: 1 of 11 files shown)


vMaroon:
Thank you for fixing this @mschulist - actually just faced. We should have this in the release.

cc @hexfusion @zdtsw

vMaroon (approved):

PR #2391, opened 2026-08-15
feat(epp): add sglanghttp parser for /generate

NaomiEisen:
**What type of PR is this?**
/kind feature

Adds request and response parsing for SGLang's native `/generate` endpoint, specifically for pre-tokenized (token-in/token-out) SGLang requests.

**Details:**

- Currently supports only pre-tokenized requests. Multimodal/text inputs are not supported.

- Reuses `GenerateRequest` rather than introducing an SGLang-specific type. This is a light initial implementation and might not be ideal in the long run - `GenerateRequest` was designed for vLLM's wire format and cannot be used to unmarshal SGLang payloads directly. Batched-token handling is also not uniform (a single prompt goes into `GenerateRequest.TokenIDs`; a batch goes into `TokenizedPrompt.PerPromptTokens`). A more generic native-generate structure, or separate per-backend types may make sense later, after the SGLang Rust migration.

- Updates the parser registry's Resolve to use longest-suffix matching instead of relying solely on parser order. This avoids a path collision where SGLang's `/generate` suffix could incorrectly match vLLM's `/inference/v1/generate`. 

- Batch response usage is summed across per-prompt `meta_info` objects to produce a single request-level Usage. This is symmetric with `TokenizedPrompt.TokenCount()`, which sums input tokens across all prompts in a batch (the batch is treated as one request).

_Validated against against a cluster running SGLang engines (token-in/token-out `/generate`)._

**Release note**:
```release-note
Add SGLang /generate endpoint parser for pre-tokenized requests.
```

### DIFF
--- a/pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2026 The Kubernetes Authors.
+Copyright 2026 The llm-d Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp.go
@@ -145,7 +145,6 @@ func hasJSONValue(data json.RawMessage) bool {
 	return len(data) > 0 && strings.TrimSpace(string(data)) != "null"
 }
 
-
 func parseCacheSalt(data json.RawMessage) (string, error) {
 	if !hasJSONValue(data) {
 		return "", nil
--- a/pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp.go
@@ -94,9 +94,6 @@ func (p *SGLangHTTPParser) Claims() fwkrh.Claims {
 type sgLangGenerateWire struct {
 	InputIDs       json.RawMessage `json:"input_ids"`
 	ExtraKey       json.RawMessage `json:"extra_key"`
-	ImageData      json.RawMessage `json:"image_data"`
-	VideoData      json.RawMessage `json:"video_data"`
-	AudioData      json.RawMessage `json:"audio_data"`
 	SamplingParams json.RawMessage `json:"sampling_params"`
 	Stream         bool            `json:"stream"`
 }
@@ -123,9 +120,6 @@ func (p *SGLangHTTPParser) parseGenerateRequest(rawBody []byte) (*fwkrh.ParseRes
 		return nil, fmt.Errorf("invalid generate request: %w", err)
 	}
 
-	if hasMultimodalData(wire.ImageData) || hasMultimodalData(wire.VideoData) || hasMultimodalData(wire.AudioData) {
-		return nil, errors.New("unsupported generate request: multimodal inputs are not supported; only pre-tokenized prompts are supported")
-	}
 	if !hasJSONValue(wire.InputIDs) {
 		return nil, errors.New("invalid generate request: input_ids must be provided")
 	}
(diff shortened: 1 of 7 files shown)


pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp.go
```diff
@@ -0,0 +1,361 @@
...
+		Stream:          wire.Stream,
+	}
+	if len(batches) == 1 {
+		result.Generate.TokenIDs = batches[0]
+	} else {
+		result.TokenizedPrompt = &fwkrh.TokenizedPrompt{
+			PerPromptTokens: batches,
+			CacheSalt:       cacheSalt,
+		}
+	}
```
sagearc: let's scope batch support out of the initial PR?

pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp.go
```diff
@@ -0,0 +1,361 @@
...
+		return nil, fmt.Errorf("unsupported generate request: %w", err)
+	}
+	batches, err := parseInputIDs(wire.InputIDs)
+	if err != nil {
+		return nil, fmt.Errorf("invalid generate request: %w", err)
+	}
+	// Single prompt: TokenIDs for the producer to copy. Batch: TokenizedPrompt,
+	// because TokenIDs cannot hold [][]uint32 (producer skips if already set).
+	result := &fwkrh.InferenceRequestBody{
+		Generate:        &fwkrh.GenerateRequest{CacheSalt: cacheSalt},
```
sagearc: Since `GenerateRequest` is now the shared for generate endpoints, can sglang decode into it directly? The `sgLangGenerateWire` temporary wire type introduced here is immediately discarded and creates a different normalization path from vllm (also it might give the impression we support mm handling)

pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp_test.go
sagearc, line 1: can you consolidate tests after removing batch cases?

pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp.go
```diff
@@ -0,0 +1,361 @@
...
+	} `json:"meta_info"`
+}
+
+// ParseResponse extracts token usage from a /generate response.
+func (p *SGLangHTTPParser) ParseResponse(_ context.Context, body []byte, headers map[string]string, _ bool) (*fwkrh.ParsedResponse, error) {
+	if len(body) == 0 {
+		return nil, nil //nolint:nilnil
+	}
+	if isEventStream(headers) {
+		return &fwkrh.ParsedResponse{Usage: extractStreamingUsage(body)}, nil
```
sagearc, line 202: sglang reports total token usage in every streaming chunk. Please record usage only from the final chunk so each request is counted once.

sagearc (changes requested):
Thanks @NaomiEisen!

pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp.go
NaomiEisen: Done :)

pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp.go
NaomiEisen: Good catch! I'm now using `finish_reason` field to detect the final chunk. It's null on intermediate chunks, and populated only on the final one.

sagearc (dismissed):

sagearc (dismissed):

sagearc:
cc @ahg-g @vMaroon @zetxqx

vMaroon (dismissed):

pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp.go
```diff
@@ -0,0 +1,274 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
vMaroon:
```suggestion
Copyright 2026 The llm-d Authors.
```

pkg/epp/framework/plugins/requesthandling/parsers/sglanghttp/sglanghttp_test.go
```diff
@@ -0,0 +1,350 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
vMaroon:
```suggestion
Copyright 2026 The llm-d Authors.
```

vMaroon (approved):

sagearc:
@vMaroon @NaomiEisen please fix CI

Issue #2059, opened 2026-07-16
Migrate to generic endpoint attribute filter/scorer and deprecate legacy  metrics code


liu-cong:
We now support filter/scorer for arbitrary endpoint attributes: 
[filter](https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/plugins/scheduling/filter/endpointattribute)
[scorer](https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/plugins/scheduling/scorer/endpointattribute)

They can replace legacy `runningrequest`, `queuedepth` and `kvcaheutilization` scorers.

To graduate them, we should:
* [ ] Benchmark using the generic scorer vs. legacy ones and confirm performance parity 
* [ ] Remove special metrics handling code - now all metrics consumption are via the flexible data layer framework
* [ ] Deprecate legacy filter/scorers

MicheleCampi:
I'd like to work on this graduation, starting with item 1 (benchmark).

Plan for the benchmark:
- Use the `test/perf` nightly harness (`run_nightly_perf.py`) with the existing shared-prefix configs, adding a router config variant that swaps `runningrequest`/`queuedepth`/`kvcacheutilization` for their `endpointattribute` equivalents.
- Compare TTFT/TPOT/throughput and scheduler-side overhead across repeated runs, reporting mean ± std dev per config, results posted here for review.

If parity is confirmed, I'm happy to follow through with items 2 and 3 (metrics handling removal, legacy deprecation) as separate PRs, sequenced however you prefer. Does this work?

MicheleCampi:
Following up on my earlier offer for item 1. I re-checked the parity mapping against current `main` (`f56f3bd9`, 55 commits on from when I first looked) — the three legacy scorers and the `endpointattribute` scorer/filter are structurally unchanged, and `runningrequests` is still absent from the default profile in `config/loader/defaults.go`, so it needs adding to both variants for a like-for-like comparison.

One semantic difference is worth deciding on before any benchmark numbers get produced, because it changes what the comparison measures. `QueueScorer.Score` reads `WaitingQueueSize` as a plain `int`, so an endpoint that hasn't reported metrics yet is indistinguishable from one with an empty queue: it enters as the minimum, and `(max - v) / (max - min)` gives it `1.0` — the best possible score. The endpoint-attribute path treats a missing attribute as absent rather than zero, so it lands at the opposite end.

[...]

PR #2393, opened 2026-08-16
Add endpoint-attribute parity coverage and a benchmark config (#2059 item 1)

MicheleCampi:
**What type of PR is this?**

/kind cleanup

Item 1 of #2059 asks for a benchmark comparing the legacy metric scorers against the generic endpoint-attribute scorer. This adds the router config that benchmark would run, and the test that makes its result mean something.

**The test first, because it changes what the config has to prove.**

`EndpointAttributeScorer.Consumes()` declares its data dependency as `Optional`, and `datalayer.Scope` only builds the set of `DataKey`s a plugin is permitted to touch — nothing verifies that a producer exists for what a consumer reads. A variant whose `customMetrics` block is wrong, or missing entirely, therefore loads cleanly through both phases of the runner (I checked), runs, and scores every endpoint `0.0`. That is indistinguishable from a scorer that discriminates badly, and a benchmark would report it as a result.

So parity is asserted behaviourally instead. Both implementations receive the same endpoints, each value carried both on the `Metrics` struct the legacy scorers read and on the attribute map the endpoint-attribute scorer reads, and the score maps must agree. Removing either `Put` makes the test fail with `legacy=1` against `ea=0`.

That failure is itself worth recording: **a missing value reads as best under the legacy scorers and worst under the endpoint-attribute scorer.** The legacy path declares a zero default in `Consumes()`, so an endpoint with no scrape yet looks like an empty queue; the endpoint-attribute path sees the attribute as absent and assigns `0.0`. A benchmark should discard its warm-up window for that reason, and any migration should be aware the two disagree about endpoints that have not reported.

**The config.**

`endpoint-attribute-parity.yaml` replaces `queue-scorer` and `kv-cache-utilization-scorer` with two `endpoint-attribute-scorer` instances and changes nothing else — same plugins, same weights — so a difference in the benchmark is attributable to the swap.

- queue → `adaptiveRange` with `linear_lower_is_better`, matching the min-max normalisation the legacy scorer performs across the candidate set, including its neutral `1.0` when all endpoints are equal.
- kv-cache → `fixedRange [0,1]`, since that scorer is absolute (`1 - usage`); adaptive normalisation would not reproduce it.

Two scope notes:

- `running-requests-size-scorer` is deliberately in neither arm. It is not in `optimized-baseline`, and adding it to the baseline in order to have something to compare would change the configuration the project actually runs.
- The `vllm` `engineConfig` is restated in full because the built-in merge is per engine, not per list: declaring `vllm` to attach `customMetrics` suppresses the `vllm` built-in alone, and dropping one of its five specs would break the legacy arm rather than the one under test.

Part of #2059 (item 1). Items 2 and 3 are not addressed here.

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/scorer/endpointattribute/legacy_parity_test.go
+++ b/pkg/epp/framework/plugins/scheduling/scorer/endpointattribute/legacy_parity_test.go
@@ -1,0 +1,108 @@
+package endpointattribute
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+	fwksched "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
+	attrmetrics "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/attribute/metrics"
+	extmetrics "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/datalayer/extractor/metrics"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/scorer/kvcacheutilization"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/scheduling/scorer/queuedepth"
+)
+
+// Behavioural parity for llm-d-router#2059 item 1.
+//
+// Loading a config proves nothing here. The endpoint-attribute scorer
+// declares its dependency as Optional (Consumes(), this file's package), and
+// datalayer.Scope only builds the set of keys a plugin is allowed to touch —
+// it never checks that something produces what something else consumes. A
+// variant whose customMetrics block is wrong, or absent, therefore loads
+// cleanly, runs, and scores every endpoint 0.0. That is indistinguishable
+// from a scorer that discriminates badly, and a benchmark would report it as
+// a result.
+//
+// So the check is behavioural: both implementations get the same endpoints —
+// the value on the Metrics struct the legacy scorers read AND on the
+// attribute map the endpoint-attribute scorer reads — and the score maps
+// must agree.
+
+func endpointWith(queue int, kvUsage float64) fwksched.Endpoint {
+	attrs := fwkdl.NewAttributes()
+	attrs.Put(attrmetrics.ScalarMetricDataKey(extmetrics.WaitingQueueSizeKey),
+		attrmetrics.ScalarMetricValue(float64(queue)))
+	attrs.Put(attrmetrics.ScalarMetricDataKey(extmetrics.KVCacheUsagePercentKey),
+		attrmetrics.ScalarMetricValue(kvUsage))
+	return fwksched.NewEndpoint(
+		&fwkdl.EndpointMetadata{},
+		&fwkdl.Metrics{WaitingQueueSize: queue, KVCacheUsagePercent: kvUsage},
+		attrs,
+	)
+}
+
+func adaptiveParams(attributeKey string) parameters {
+	return parameters{
+		AttributeKey: attributeKey,
+		Algorithm: algorithmParameters{
+			Type:          algorithmLinearLowerIsBetter,
+			Normalization: normalizationParameters{AdaptiveRange: &adaptiveRangeParameters{}},
+		},
+	}
+}
+
+func fixedParams(attributeKey string, min, max float64) parameters {
+	return parameters{
+		AttributeKey: attributeKey,
+		Algorithm: algorithmParameters{
+			Type:          algorithmLinearLowerIsBetter,
+			Normalization: normalizationParameters{FixedRange: &fixedRangeParameters{Min: min, Max: max}},
+		},
+	}
+}
+
+func TestParityQueueScorer(t *testing.T) {
+	cases := [][]fwksched.Endpoint{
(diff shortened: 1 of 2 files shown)


MicheleCampi:
@hexfusion @elevran — gentle ping, this has been quiet since the 16th.

Item 1 of #2059: a benchmark config putting the endpoint-attribute scorer at parity with the legacy ones, plus `TestParityQueueScorer` and `TestParityKVCacheScorer`. Additive only — **+200/−0**, two new files, no existing behaviour touched.

I have just rebased onto `main` at 227f6cad, 31 commits on: merges clean, builds, package tests pass. Checks are green.

**Why a behavioural parity test rather than a loader test.** `EndpointAttributeScorer.Consumes()` declares its key as `Optional`, so a config where the `customMetrics` block is missing or wrong does not fail at init — `data_graph.go` logs *"optional data key has no producer, plugin will use fallback"* and carries on. The scorer then returns 0.0 for every endpoint, which is indistinguishable from a scorer that discriminates badly. A benchmark would report that as a result, and one warning line in startup logs is thin protection against it.

The test asserts the behaviour instead: the same endpoint values are placed on both the `Metrics` struct and the attribute map, and both scorers must agree. Remove one `attrs.Put` and it fails with `legacy=1` against `ea=0` — which is also the semantic gap worth knowing about, since an absent value reads as *best* under the legacy default and *worst* under the attribute path.

Happy to split the config and the test into separate PRs if that is easier to review.

MicheleCampi:
@elevran — #2355 changed the picture for this PR, in a way that I think makes it more useful rather than less.

After that migration, of the ten scorers under `scheduling/scorer/` **eight declare `Required`** and only two remain `Optional` — `topologyaffinity` and `endpointattribute`. So `QueueScorer` now fails at init when its producer is missing, while `EndpointAttributeScorer` still logs `optional data key has no producer, plugin will use fallback` and scores 0.0 on every endpoint.

That is exactly the pair this PR benchmarks side by side. The parity test asserts they agree on identical endpoint values, so a config where `customMetrics` is missing or misnamed shows up as `legacy=1` against `ea=0` rather than as a plausible benchmark number. The asymmetry is sharper now than when I opened it.

I am not proposing to change the `Optional` declaration — the plugin genuinely handles absence, and #2355 left it deliberately. The point is only that a benchmark comparing the two paths should be able to tell "no signal" from "poor signal", and right now nothing does.

Status: rebased on `main` at c99864c5, 26 commits on. Still merges clean, package tests pass, checks green, **+200/−0** in two new files.

Since you wrote much of this scorer and the migration, your read on whether the parity assertion matches the intended semantics is worth more than a generic review. Happy to split the config and the test if that is easier.

pkg/epp/framework/plugins/scheduling/scorer/endpointattribute/legacy_parity_test.go
```diff
@@ -0,0 +1,108 @@
...
+//
+// So the check is behavioural: both implementations get the same endpoints —
+// the value on the Metrics struct the legacy scorers read AND on the
+// attribute map the endpoint-attribute scorer reads — and the score maps
+// must agree.
+
+func endpointWith(queue int, kvUsage float64) fwksched.Endpoint {
+	attrs := fwkdl.NewAttributes()
+	attrs.Put(attrmetrics.ScalarMetricDataKey(extmetrics.WaitingQueueSizeKey),
+		attrmetrics.ScalarMetricValue(float64(queue)))
```
elevran, line 37: The asymmetric behaviour the comment block above describes (legacy=1 vs ea=0 when the value is absent) is real but not exercised by any case here. Every `attrs.Put` has a matching `Metrics` field.

A small follow-up subtest would lock the property in:

```go
// Endpoint with the metric on Metrics but no attribute: legacy treats 0 as
// empty queue (best); endpoint-attribute returns 0.0 for the absent attr.
noAttr := fwksched.NewEndpoint(
    &fwkdl.EndpointMetadata{},
    &fwkdl.Metrics{WaitingQueueSize: 0, KVCacheUsagePercent: 0.0},
    fwkdl.NewAttributes(),
)
```

Pair it with the existing legacy/ea scorers and assert `legacy[noAttr] == 1.0`, `ea[noAttr] == 0.0`. Skippable for this PR if you'd rather keep it tight. The property is real, just absent from the test code.

elevran (commented):
Thanks - the parity assertion catches a real semantic gap with its effect increased given #2355. One observation: the comment block at the top of the test file documents the asymmetric runtime behaviour (legacy=1 vs ea=0 when the value is absent), but this is not tested: every test case here puts the value on both `Metrics` and `attrs`. Please add a test to assert this behavior (see below).
Please also fix failing lint.

MicheleCampi:
Both addressed in `3e331774` — thanks, the first one was a fair catch: the file documented the behaviour and then never took that path.

`TestDivergenceWhenAttributeAbsent` does what you suggested. One endpoint with `WaitingQueueSize: 0` on `Metrics` and an empty attribute map, and it asserts `legacy[noAttr] == 1.0` against `ea[noAttr] == 0.0`. I kept it in this PR rather than deferring — a comment block claiming behaviour that no case exercises is the kind of thing this PR exists to stop.

I checked it fails for the right reason before pushing: flipping the expected `ea` score to 1.0 turns it red.

Lint was the `importas` rule — `extractormetrics`, not `extmetrics`. It only started running after last week's rebase brought the file into the path filter, which is why it was green when I first asked for review and red when you looked.

Tests pass, gofmt and vet clean, no failing checks on the head commit. Ready for another look whenever suits — happy to take any further changes.

elevran (approved):

PR #2394, opened 2026-08-16
fix(options): validate refresh-metrics-interval 50ms floor at startup

### LINKED ISSUE #1880: feat: configurable per-source scraping interval in the data layer

noalimoy:
**What type of PR is this?**
/kind cleanup

The `--refresh-metrics-interval` flag accepted arbitrarily small values, which
were silently clamped to 50ms at runtime. This moves the 50ms floor into
`options.Validate()` so the process fails fast at startup with a clear error
instead of running with a silently different value.

Follow-up to #2214 per reviewer feedback.

- Adds `RefreshMetricsInterval < 50ms` check in `Validate()`
- Removes the now-redundant `requestedPollingInterval` field and clamp warning from `Runtime`

Fixes #1880

## Test plan
- [x] Flag value below 50ms rejected at validation
- [x] Flag value at 50ms passes validation
- [x] `make test` passes

```release-note
NONE
```

### DIFF
--- a/pkg/epp/server/options.go
+++ b/pkg/epp/server/options.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
+	ctrl "sigs.k8s.io/controller-runtime"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	"github.com/llm-d/llm-d-router/pkg/common/observability/logging"
@@ -38,9 +39,10 @@ import (
 )
 
 const (
-	DefaultGrpcPort      = 9002
-	DefaultPoolNamespace = "default"        // default when pool namespace is empty (CLI flag default is empty)
-	DefaultDrainTimeout  = 30 * time.Second // graceful shutdown drain window
+	DefaultGrpcPort           = 9002
+	DefaultPoolNamespace      = "default"        // default when pool namespace is empty (CLI flag default is empty)
+	DefaultDrainTimeout       = 30 * time.Second // graceful shutdown drain window
+	MinRefreshMetricsInterval = 50 * time.Millisecond
 )
 
 // deprecatedMetricFlags lists metric flags that are superseded by engineConfigs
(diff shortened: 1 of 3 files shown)


pkg/epp/server/options.go
```diff
@@ -385,6 +385,9 @@ func (opts *Options) Validate() error {
 		return errMetricsTLSWithoutAuth
 	}
 
+	if opts.RefreshMetricsInterval < 50*time.Millisecond {
```
elevran: I would prefer not to fail fast here, but instead clamp the floor at 50ms and issue a log warning. 
 Also, consider making `50*time.Millisecond` a constant used both here and as the default value when the CLI flag is missing.

elevran:
@noalimoy lint failing. Will need to restamp after you fix it.

noalimoy:
> @noalimoy lint failing. Will need to restamp after you fix it.

Fixed

elevran (approved):

elevran (approved):

PR #2396, opened 2026-08-16
role-aware load split for P/D disaggregation

### LINKED ISSUE #2395: Load on P/D disaggregated
The in-flight load producer currently assigns both input and estimated output tokens to every endpoint that handles a request, regardless of whether that endpoint is a prefill-only, decode-only, or both.

### Root cause
`estimateRequestTokens` in `inflightload/producer.go` unconditionally adds`ISL + OSL_est`, with no awareness of the `llm-d.ai/role` pod label that already exists in the system to distinguish prefill-only (`prefill`, `encode-prefill`) from decode-only (`decode`) endpoints.

/assign-me

DolevAdas:
**What type of PR is this?**
/kind feature

In a P/D disaggregated deployment, `inflight-load-producer` currently charges **ISL + OSL_est** to every endpoint that handles a request, regardless of whether that endpoint is a prefill-only or decode-only worker. This over-counts token load on both sides and degrades scheduling accuracy.
This PR makes `estimateRequestTokens` role-aware by reading the `llm-d.ai/role` pod label that the EPP already uses for endpoint selection

The split is backward-compatible: pods without a role label fall through to the original `ISL + OSL_est` path, so monolithic deployments are unaffected.

**Notes**
This PR is best merged after PR #2309 `OSL-AwareDecodeScheduling`.

Fixes #2395

```release-note
Role-aware load split in P/D deployments - input tokens to prefill endpoints, estimated output tokens to decode endpoints.
```


LukeAVanDrie:
fyi @loicmarchal as we are doing something similar for saturation gauge

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
```diff
@@ -44,6 +44,13 @@ const (
 	InFlightLoadProducerType = inflightloadconstants.InFlightLoadProducerType
 	profilePrefill           = "prefill"
 	maxDebugDumpEndpoints    = 100
+
+	// Pod-role label and values -- canonical definitions in
+	// pkg/epp/framework/plugins/scheduling/filter/bylabel/roles.go.
+	podRoleLabel         = "llm-d.ai/role"
```
elevran: These constants duplicate the label name and role values already exported from `bylabel` (`RoleLabel`, `RolePrefill`, `RoleEncodePrefill`, `RoleDecode` in pkg/epp/framework/plugins/scheduling/filter/bylabel/roles.go). The comment here even points at that file as canonical instead of importing it.

There's also an existing role-classification implementation to model after: `partitionEndpoints` in pkg/epp/flowcontrol/controller/internal/processor.go imports `bylabel` directly and switches on all role values (including `RolePrefillDecode`, `RoleBoth`, `RoleEncodePrefillDecode`, `RoleEncode`). That package is already used outside its own subtree, so there's no import-cycle concern.

Importing `bylabel` and using its constants here would avoid the risk of the two definitions silently drifting if role values ever change.

elevran (commented):
Solid fix for the over-counting issue, and the test coverage (unit cases plus the end-to-end PreRequest/lifecycle test) is thorough. One should-fix on constant duplication below.

ahg-g:
@elevran is this good to go?

ahg-g (approved):

PR #2397, opened 2026-08-16
feat(metrics): extract vLLM kv_offload_tiering_chunk_* metrics and add storage tier weight

zdtsw:
**What type of PR is this?**
/kind feature

Add native support for vLLM `kv_offload_tiering_chunk_*` Prometheus metrics in the metrics extractor, and add a `"storage"` tier weight to the precise prefix cache producer so storage-tier blocks score lower than GPU/CPU.

**Extraction:**
- Extract `kv_offload_tiering_chunk_*` metrics as new endpoint attributes.
- Skip missing tiering series without erroring: vLLM creates a tier's counter only after that tier's first event, so an absent series cannot be distinguished from a misconfigured tier label. Attributes stay unset until the series appears, keeping  non-tiering endpoint distinguishable from an idle tiering one.
- Default vLLM config targets tier `1:fs`. Deployments with multiple  external media override via `engineConfigs`.
- Rename `CustomMetric` to `AttributeMetric` for reuse; retain `CustomMetric` as a type alias.

**Scoring:**
- Add `"storage"` (weight 0.3) to `DefaultKVCacheBackendConfig`:   gpu=1.0, cpu=0.8, storage=0.3.
- default suits varied media. Deployments with fast storage  (e.g. NVMe) should raise it via precise-prefix-cache-producer's `parameters.indexerConfig.kvCacheBackendConfigs`.

```release-note
Add extraction of vLLM kv_offload_tiering_chunk_* metrics. Storage-tier KV cache blocks (from TieringOffloadingSpec offloading) now score at 0.3 by default. Adjust it via kvCacheBackendConfigs for faster or slower media.
```

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/extractor/metrics/README.md
+++ b/pkg/epp/framework/plugins/datalayer/extractor/metrics/README.md
@@ -18,6 +18,7 @@ The Core Metrics Extractor is a data layer plugin responsible for extracting mod
     -   **KV Cache Usage**: Percentage of KV cache currently utilized.
     -   **LoRA Adapters**: Information about active and waiting LoRA adapters.
     -   **Cache Configuration**: Block size and total number of GPU blocks.
+    -   **Tiered Offloading** (optional): KV cache tiering metrics (`kv_offload_tiering_*`) when vLLM is configured with `TieringOffloadingSpec`. Silently skipped on non-tiering deployments.
 5.  Stores these values as attributes on the endpoint, making them available to scheduling plugins.
 
 ## Attributes produced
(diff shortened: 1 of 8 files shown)


yankay:
HI @zdtsw 
Thanks for adding this! vLLM renamed these metrics from `block_{hits,queries}` to `chunk_{hits,queries}` in vllm-project/vllm#52812. I tested the PR against vLLM v0.28.0: the current selectors report both `block_*` families as missing, while `chunk_*` works. Could we update the selectors and tests, or support both names if compatibility is needed? :)

zdtsw:
> HI @zdtsw Thanks for adding this! vLLM renamed these metrics from `block_{hits,queries}` to `chunk_{hits,queries}` in [vllm-project/vllm#52812](https://github.com/vllm-project/vllm/pull/52812). I tested the PR against vLLM v0.28.0: the current selectors report both `block_*` families as missing, while `chunk_*` works. Could we update the selectors and tests, or support both names if compatibility is needed? :)

Thanks for catching this @yankay! You're right 
Think it is good enough to rename metrics in this PR, as it is not used anywhere in llm-d yet. I've updated the selectors and attribute keys to use the new chunk_* names only.

PR #2400, opened 2026-08-16
fix(kvcache): report index hit metrics as contiguous prefix chains

### LINKED ISSUE #2398: kv-block index hit metrics count non-contiguous blocks and race the caller's result map
`instrumentedIndex.Lookup` folds its hit metrics through `recordHitMetrics`, which has two problems.

**The counts are not contiguous.**

`recordHitMetrics` counts every block a pod holds anywhere in the request:

```go
podCount := make(map[string]int)
for _, pods := range keyToPods {
    for _, p := range pods {
        podCount[p.PodIdentifier]++
    }
}
```

A pod holding blocks 0, 7 and 42 reports a count of 3, the same as a pod holding blocks 0, 1 and 2. Only the second pod offers a prefix the scheduler can act on, since a prefix-cache hit has to start at the first block and run contiguously. So `kv_cache_index_lookup_hits_total` and `kv_cache_index_max_pod_hit_count_total` overstate the cache hit a request can actually use, and the degree of overstatement depends on how fragmented the index is.

**The fold races the caller.**

It is dispatched as `go recordHitMetrics(pods)` while `Lookup` returns `pods` to its caller. The caller owns that map once `Lookup` returns and is free to mutate or reuse it, so the goroutine can read it concurrently.

**Proposed fix**

Count the longest contiguous chain any single pod holds starting from the first requested block, and run the fold synchronously. Duplicate device tiers for one pod at the same block should count once.

This changes the meaning of both counters, so it is a breaking change for anyone with dashboards or alerts on them and needs a release note saying so.

nilig:
**What type of PR is this?**

/kind bug

`recordHitMetrics` counted every block a pod held anywhere in the request, so a
pod holding scattered blocks reported the same hit count as a pod holding a
usable prefix. The scheduler can only act on a contiguous prefix from the first
block, so the metric overstated the cache hit a request could actually use.
Count the longest contiguous chain from the first request key instead.

The fold also ran in a goroutine over the result map, which the caller owns once
`Lookup` returns, so it could race the caller's reuse of that map. It now runs
synchronously.

This changes the meaning of two existing metrics. Dashboards reading them will
report different values for identical traffic:

- `kv_cache_index_lookup_hits_total`
- `kv_cache_index_max_pod_hit_count_total`

**Test plan**:

- [x] `TestMaxContiguousPodHits` - covers a chain broken by a gap, a pod that
      first appears after the leading key, duplicate device tiers at one key
      counting once, and a leading key held by nobody
- [x] `make presubmit`
- [x] `make test-unit`

Fixes #2398

```release-note
BREAKING: `kv_cache_index_lookup_hits_total` and
`kv_cache_index_max_pod_hit_count_total` now report the longest contiguous prefix
chain held by a single pod, counted from the first requested block. They
previously counted every matching block regardless of contiguity, which
overstated the cache hit a request could actually use. Dashboards and alerts on
these metrics need to be re-baselined.
```

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -374,8 +374,8 @@ Prefix `llm_d_epp_`. Registered only when the embedded llm-d-kv-cache metrics ar
 | `kv_cache_index_admissions_total` | Counter | Blocks admitted to the index. |
 | `kv_cache_index_evictions_total` | Counter | Blocks evicted from the index. |
 | `kv_cache_index_lookup_requests_total` | Counter | Index lookups performed. |
-| `kv_cache_index_lookup_hits_total` | Counter | Lookups that matched at least one block. |
-| `kv_cache_index_max_pod_hit_count_total` | Counter | Best per-pod hit count observed per lookup. |
+| `kv_cache_index_lookup_hits_total` | Counter | Contiguous prefix blocks matched by the best pod per lookup. |
+| `kv_cache_index_max_pod_hit_count_total` | Counter | Longest contiguous per-pod prefix chain observed per lookup. |
 | `kv_cache_index_lookup_latency_seconds` | Histogram | Index lookup latency. |
 | `kv_cache_events_dedup_removed_hashes_suppressed_total` | Counter | Deduplicated removal hashes suppressed. |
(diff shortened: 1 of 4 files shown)


sagearc (approved):
Thanks @nilig

vMaroon (approved):

PR #2401, opened 2026-08-16
perf(kvcache): make index lookup recency read-only

### LINKED ISSUE #2399: InMemoryIndex.Lookup promotes LRU recency, taking the cache write lock on a read path
`InMemoryIndex.Lookup` reads the routing data with `Get`:

```go
pods, found := m.data.Get(requestKey)
```

`Get` on the underlying LRU promotes the entry to most-recently-used, which has two consequences on what is otherwise a pure read path.

**It takes the cache write lock.** Every key in the request acquires the LRU's write lock to reorder the list, so concurrent lookups serialize against each other and against admissions, on a path that only needs to read.

**Read traffic reorders eviction.** A block that is looked up often but never re-admitted is held in the index by lookup traffic alone. Eviction then reflects which blocks the scheduler asked about rather than which blocks the engines actually hold, which is the signal the index is supposed to carry.

**Proposed fix**

Use `Peek`, which reads the entry without promoting it, so eviction order reflects admission and update traffic only.

This is a deliberate behavior change to eviction and should carry a release note. Worth noting separately, and not addressed here: `Evict` has a sibling-scan that promotes recency through its `m.data.Get(rk)` emptiness check.

nilig:
**What type of PR is this?**

/kind cleanup

`InMemoryIndex.Lookup` promoted LRU recency for every key it touched, taking the
cache write lock on what is otherwise a read path and letting read traffic
reorder eviction. `Peek` reads the entry without promoting it, so eviction order
now reflects admission and update traffic only.

This is a deliberate behavior change: a block that is looked up often but never
re-admitted is no longer held in the index by lookups alone.

**Test plan**:

- [x] `TestLookupDoesNotRefreshIndexLRU` - verified failing before the change and
      passing after, so it gates the behavior rather than describing it
- [x] `make presubmit`
- [x] `make test-unit`

Fixes #2399

```release-note
KV-block index lookups no longer refresh an entry's LRU recency. Index eviction
order now reflects block admission and updates rather than lookup traffic.
```

### DIFF
--- a/pkg/kvcache/kvblock/in_memory.go
+++ b/pkg/kvcache/kvblock/in_memory.go
@@ -120,7 +120,7 @@ func (m *InMemoryIndex) Lookup(ctx context.Context, requestKeys []BlockHash,
 	highestHitIdx := 0
 
 	for idx, requestKey := range requestKeys {
-		if pods, found := m.data.Get(requestKey); found { //nolint:nestif // TODO: can this be optimized?
+		if pods, found := m.data.Peek(requestKey); found { //nolint:nestif // TODO: can this be optimized?
 			if pods == nil || pods.cache.Len() == 0 {
 				traceLogger.Info("no pods found for key, cutting search", "key", requestKey)
 				return podsPerKey, nil // early stop since prefix-chain breaks here
--- a/pkg/kvcache/kvblock/in_memory_bench_test.go
+++ b/pkg/kvcache/kvblock/in_memory_bench_test.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kvblock_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	. "github.com/llm-d/llm-d-router/pkg/kvcache/kvblock"
+)
+
+func BenchmarkInMemoryIndexLookup(b *testing.B) {
+	for _, tc := range []struct {
+		blocks  int
+		entries int
+	}{
+		{blocks: 225, entries: 2},
+		{blocks: 225, entries: 10},
+		{blocks: 1034, entries: 2},
+	} {
+		b.Run(fmt.Sprintf("blocks=%d/entries=%d", tc.blocks, tc.entries), func(b *testing.B) {
+			ctx := log.IntoContext(b.Context(), logr.Discard())
+			index, err := NewInMemoryIndex(&InMemoryIndexConfig{
+				Size:         tc.blocks + 1,
+				PodCacheSize: tc.entries,
+			})
+			if err != nil {
+				b.Fatal(err)
+			}
+
+			keys := make([]BlockHash, tc.blocks)
+			for i := range keys {
+				keys[i] = BlockHash(i + 1)
+			}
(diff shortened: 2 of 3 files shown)


GolfRider:
Tested locally—the KV-block unit and race tests passed. I also confirmed that Get updates LRU recency under an exclusive lock, while Peek performs a read-only lookup. 

The change matches the issue’s intended behavior, and the regression test covers the eviction-order change. Looks good to me.

GolfRider (commented):
Tested locally—the KV-block unit and race tests passed. I also confirmed that Get updates LRU recency under an exclusive lock, while Peek performs a read-only lookup.

The change matches the issue’s intended behavior, and the regression test covers the eviction-order change. Looks good to me.

sagearc (commented):
After this change, what refreshes the corresponding blocks in the router's LRU for the chosen endpoint?

nilig:
Short answer: `Add` refreshes it, and there was never an endpoint-specific refresh to begin with.

The LRU is keyed by block hash only — there is no per-endpoint recency. Which endpoints hold a block lives inside the entry, not in the LRU ordering. `Lookup` also runs while candidates are being scored, before the picker has chosen a destination, so the old `Get` promotion never meant "refresh the chosen endpoint's blocks" — it promoted every block consulted during scoring, for winners and losers alike.

After this change, recency is refreshed whenever a block is written to the index: `Add` resolves the entry through `m.data.Get`, so every `BlockStored` event from an engine promotes it. With `speculativeIndexing` enabled (off by default), routing a request also refreshes that request's blocks, since `PreRequest` inserts a speculative entry for the selected endpoint over those keys.

In practice recency barely matters here: the only thing it controls is capacity eviction. Blocks leave the index because engines report `BlockRemoved`, or a pod resets or goes away (`Clear`) — not because they aged. Capacity eviction only kicks in past the configured `size`, default 1e8 keys, which realistic fleets never reach.

The one scenario this PR changes: an undersized index plus a block an engine keeps serving from cache but never re-stores. That block sends no new events, so nothing refreshes it, and it could get capacity-evicted while still resident on the engine. vLLM has a feature for exactly this case: `kv_cache_report_mode: "full"` (vllm-project/vllm#45261) makes the engine emit `BlockStored` for reused prefix blocks too, and `Add` then refreshes them. I'd rather cover that case with engine events than restore promotion in `Lookup` — the engine knows which blocks are actually alive, while lookups also touch blocks for candidates that lost the pick, so they're a noisy signal to drive eviction with.

vMaroon:
can you profile the effects of this change to determine whether it is worth it

nilig:
I profiled the Get to Peek change in isolation against the PR parent. Both revisions used the exact same benchmark fixture; the only production-code difference was m.data.Get(requestKey) versus m.data.Peek(requestKey).

Apple M4 Max, Go 1.26.6, five one-second samples, -cpu=16:

| Shape | Get | Peek | Change |
|---|---:|---:|---:|
| 225 blocks, 2 entries, serial | 101.9 us | 100.9 us | -1.0% |
| 225 blocks, 2 entries, 16 readers | 85.9 us | 56.3 us | -34.4% |
| 225 blocks, 10 entries, 16 readers | 137.7 us | 127.2 us | -7.6% |
| 1,034 blocks, 2 entries, 16 readers | 352.0 us | 163.4 us | -53.6% |

Allocations were materially unchanged. The serial effect is negligible, as expected. The benefit appears under concurrent scoring because Get takes the LRU write lock and changes recency for every requested block, while Peek permits concurrent readers.

The gain narrows as per-block result materialization becomes dominant. This PR removes top-level LRU write contention, while #2402 addresses the larger lookup cost by fusing lookup and scoring without materializing the per-block result map. The two changes are complementary.

I also pushed the reproducible benchmark and a complementary regression test confirming that Add still refreshes recency. The KV-block race suite passes.

Given the one-line production change, the intended eviction semantics, and the measured concurrent improvement, I think it is worth keeping.

nilig:
@vMaroon Givign it another thought, you’re right. I missed the multi-turn retention case. A reused prefix may be written once and then only accessed through Lookup; replacing Get with Peek makes those accesses invisible to the LRU. Under capacity pressure, the router could therefore evict a hot prefix while the engine still holds it.
The benchmark demonstrates that the LRU write lock is costly, but it does not justify changing the eviction semantics. I don’t think #2401 should merge as-is. I’ll preserve read-driven recency in #2402 as well and treat reducing that contention as a separate optimization with the multi-turn retention behavior covered by a regression test.

PR #2410, opened 2026-08-17
Bound and resume KV-event replay

vMaroon:
## What changed

- cap process-wide concurrent replay requests at eight subscribers
- resume an interrupted replay response from the next sequence instead of restarting the stream
- use a two-second per-attempt socket deadline within a two-minute end-to-end replay deadline
- add tests for replay admission, cancellation, and partial-response recovery

## Why

Each active-active EPP owns the full subscriber set. Starting all rank replays concurrently creates avoidable pressure on the publishers and EPP, while an interrupted response causes repeated work and can prevent the index from converging before readiness.

This bounds startup fanout and preserves progress across a dropped replay response.

## Scope

This PR is limited to `pkg/kvevents`. It does not include the precise scored-lookup or routing-path changes from the performance branch.

## Validation

```text
go test ./pkg/kvevents
```


pkg/kvevents/replay_limiter_test.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2025 The llm-d Authors.
```
vMaroon:
```suggestion
// Copyright 2026 The llm-d Authors.
```

sagearc (dismissed):
It is better than current state, but we restore a lossy representation of the actual kv cache. The bottleneck needs to be profiled (both memory allocations & perf) as we might not need to parse all the information provided in the replayed kv events, some of the high pressure we're seeing is avoidable without sacrificing information.

ahg-g (dismissed):

sagearc (approved):

PR #2414, opened 2026-08-17
perf(tokenizer): avoid Messages render map round trip

vMaroon:
## What changed

Anthropic Messages requests now stay in the typed render representation from protocol conversion through the vLLM HTTP renderer. The generic `RequestPayload` path remains the fallback for renderers that do not implement the typed capability.

The HTTP path still preserves:

- renderer-side model stamping
- multimodal timeout selection
- tool-only assistant messages with omitted `content`
- the existing OpenAI-shaped Messages and tool conversion from #2389

## Why

The Messages path converted the typed request into JSON-backed generic maps before the vLLM HTTP renderer marshaled it again. Large shared system prompts therefore paid an avoidable marshal, unmarshal, and second marshal on every routing request.

The typed path removes that map round trip without changing the wire body sent to `/v1/chat/completions/render`.

## Benchmark

Apple M4 Pro, Go 1.26.6, 240K-character system prompt, median of five runs:

| Path | Time | Bytes | Allocations |
|---|---:|---:|---:|
| Generic map round trip | 1.69 ms/op | 845 KB/op | 50 allocs/op |
| Typed single marshal | 0.94 ms/op | 563 KB/op | 20 allocs/op |

This is approximately 44% less preparation time, 33% fewer allocated bytes, and 60% fewer allocations.

## Validation

```text
go test ./pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer
go test -race ./pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer
go test ./pkg/epp/...
go test ./pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer \
  -run '^$' -bench '^BenchmarkMessagesRenderPreparation240K$' -benchmem -count=5
```

All pass locally.

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
@@ -27,6 +27,8 @@ import (
 
 	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
 	fwkrh "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requesthandling"
+	"github.com/llm-d/llm-d-router/pkg/kvcache/tokenization"
+	tokenizerTypes "github.com/llm-d/llm-d-router/pkg/kvcache/tokenization/types"
 )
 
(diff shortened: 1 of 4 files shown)


vMaroon:
handled in #2707

PR #2420, opened 2026-08-17
docs: add prose rules for model-drafted text to AGENTS.md

LukeAVanDrie:
**What type of PR is this?**

/kind documentation

Adds a "Constructions to delete" table to the code style section of AGENTS.md, and extends the existing framing rule to cover deictic references ("this PR", "see above").

Many contributions now arrive with model-drafted prose in PR descriptions, code comments, and commit bodies. The drafts share a recognizable set of rhetorical constructions: coined aphorisms, "X, not Y" framings, triads, self-congratulatory adjectives ("comprehensive", "production-ready"). Any one of them is harmless. At the density models produce them they displace the facts a reviewer needs. AGENTS.md is the file contributor tooling reads, so naming the constructions there, each with its fix, moves the rewrite to before review and gives a reviewer a row to link rather than a taste argument to make from scratch.

This is my very opinionated take on house style, so pushback is welcome here. It's in the same spirit as the file's existing bans on temporal framing and Unicode symbols. None of these constructions is wrong in general prose, and a deliberate human writer can use any of them well. The table targets the failure mode where they appear reflexively and in bulk. 

None; guidance-only change to AGENTS.md.

**Release note:**

```release-note
NONE
```

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -41,9 +41,25 @@ llm-d Router. Go service that routes inference requests to model-serving pods vi
 
 - Standard Go. `make format` and `make lint` are authoritative.
 - Comments are terse and only present when the WHY is non-obvious. Never paraphrase the code.
-- Docs and comments describe the current state on its own terms. No "previously", "now", "recently", "renamed from", "added to fix", or other temporal or conversational framing. A reader with no context for the change must still understand the text.
+- Docs and comments describe the current state on its own terms. No "previously", "now", "recently", "renamed from", "added to fix", "this PR", "see above", or other temporal, deictic, or conversational framing. A reader with no context for the change must still understand the text.
 - State each fact once, in its canonical location. Do not duplicate across struct docs, prose, tables, inline comments, and examples.
 - Do not use Unicode symbols or special characters in general, unless explicitly requested.
+
+### Constructions to delete
+
+Model-drafted prose drifts toward compressed, rhetorical phrasing. Before shipping any prose (comments, docs, commit bodies, PR descriptions), decompress it: rewrite each such sentence as a plain statement of the fact it carries, judged against what the change is actually for. The test: if a sentence sounds quotable, delete it.
+
+| Tic | Example | Fix |
+|---|---|---|
+| X, not Y | "the queue is not a buffer, it is a fairness mechanism" | say what it is, once |
+| Coined aphorism | "a shed request is not a failure, it is the contract" | delete the sentence |
+| Closing zinger | a paragraph that ends on a beat instead of a fact | end on information |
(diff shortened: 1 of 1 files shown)


elevran:
@LukeAVanDrie - thanks.
In general, LGTM. Did you find the table at the end useful? Unsure if more rules help the model extrapolate or implies that examples not in table are valid usage.

how do you feel about instructing agents to use https://www.asd-ste100.org/? Would you mind testing to see if this improves readability?

LukeAVanDrie:
> @LukeAVanDrie - thanks. In general, LGTM. Did you find the table at the end useful? Unsure if more rules help the model extrapolate or implies that examples not in table are valid usage.
> 
> how do you feel about instructing agents to use https://www.asd-ste100.org/? Would you mind testing to see if this improves readability?

@elevran, thanks!

The table was lifted from a personal project of mine (with examples switched to be llm-d relevant), and I have also been using it as a skill to test drive. I find it very helpful. Regarding extrapolation, the table is meant to act more as a set of anti-patterns that establishes a stylistic boundary, rather than an exhaustive list of the only valid rules.

I am familiar with the ASD-STE100 approach, but I haven't experimented with it. That would likely be the stronger, more deterministic approach, but it also completely dries up anything interesting about the writing. That may be great for docs/proposals, but maybe not the best fit for every type of writing artifact (PR descriptions, code comments, commit bodies, etc.).

Thoughts on test driving this for a week or so and then discussing if anyone has noticed or felt a difference?

elevran:
Makes sense. Standard Technical English is a different scope than this PR.

The table names specific, recognizable failure modes which makes it actually actionable for both a contributor rewriting text and a reviewer citing a row instead of arguing taste from scratch. It fits the file's existing terse, prescriptive style (it already bans temporal/conversational framing), so this reads as a natural extension rather than scope creep.

Two things worth considering:
- "Dash pileup" row has some overlap with the pre-existing "no Unicode symbols" rule. Not a conflict, just slightly redundant guidance
- Several of the "fixes" (e.g. "delete the sentence") are somewhat blunt for cases where the underlying fact is genuinely worth keeping, just not phrased well. The intro line already covers that ("rewrite it as a plain statement of the fact it carries"), so it's not actually a gap just unsure which resolution the agent would adopt (rewrite or eliminate).

elevran (approved):
LGTM. 
@LukeAVanDrie - some notes for discussion in comments. Leaving to your discretion if to merge or revise.

LukeAVanDrie:
> LGTM. @LukeAVanDrie - some notes for discussion in comments. Leaving to your discretion if to merge or revise.

Thanks, I'm going to merge this now so we can test drive it. I suspect we will be revisiting this file at which point we can evaluate your two points against some actual data and make adjustments.

PR #2425, opened 2026-08-17
perf(epp): reduce streaming parser allocations

### LINKED ISSUE #2423: perf(epp): eliminate per-chunk byte/string copies in streaming usage parsers

This issue proposes a small hot-path optimization that improves the measured streaming parser paths by 10% to 97% in microbenchmarks while reducing allocations by up to 100%.

Make the OpenAI and Anthropic streaming usage parsers operate directly on the `[]byte` response chunks received from Envoy instead of converting each chunk to a `string` and, on parsing paths, converting the SSE payload back to `[]byte`.

The proposed implementation should preserve the existing OpenAI Chat Completions, Responses API, vLLM HTTP, and Anthropic behavior while parsing SSE lines directly as byte slices, skipping unnecessary Anthropic JSON decoding, and adding focused correctness tests and `-benchmem` benchmarks.

**Why is this needed**:

Envoy sends response bodies to EPP through `ext_proc` one chunk at a time. The repository's performance analysis reports approximately 100,000 response chunk callbacks per second for a 10,000-output-token workload at 10 requests/second:

- [Output-token performance analysis: callback rate and EPP CPU scaling](https://github.com/llm-d/llm-d-router/blob/dd65342ee538ab76d4b9f7e1efb5d8e95385d06e/test/perf/config/studyanalysis/output_token_length_impact_analysis.md#L10-L14)
- [The same analysis attributes 0.81 CPU cores to response parsing overhead](https://github.com/llm-d/llm-d-router/blob/dd65342ee538ab76d4b9f7e1efb5d8e95385d06e/test/perf/config/studyanalysis/output_token_length_impact_analysis.md#L48-L50)

Small per-chunk allocations therefore compound across long streaming responses.

### OpenAI parser

The OpenAI parser currently converts every incoming byte chunk to a string:

```go
func (p *OpenAIParser) parseStreamResponse(chunk []byte) (*fwkrh.ParsedResponse, error) {
    usage := extractUsageStreaming(string(chunk))
    return &fwkrh.ParsedResponse{
        Usage: usage,
    }, nil
}
```

Source: [openai.go lines 190-194](https://github.com/llm-d/llm-d-router/blob/dd65342ee538ab76d4b9f7e1efb5d8e95385d06e/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go#L190-L194)

When a line may contain usage data, the parser converts the string subslice back to bytes before JSON decoding:

```go
lines := strings.SplitSeq(responseText, "\n")
for line := range lines {
    content, ok := strings.CutPrefix(line, streamingRespPrefix)
    if !ok {
        continue
    }
    if content == "[DONE]" || !strings.Contains(content, "usage") {
        continue
    }
    byteSlice := []byte(content)
    if err := json.Unmarshal(byteSlice, &streamResponse); err != nil {
        continue
    }
}
```

Source: [openai.go lines 372-394](https://github.com/llm-d/llm-d-router/blob/dd65342ee538ab76d4b9f7e1efb5d8e95385d06e/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go#L372-L394)

This leaves an avoidable full-chunk `[]byte -> string` allocation on every streaming callback and another `string -> []byte` copy on usage-bearing lines. The vLLM HTTP parser delegates response parsing to the OpenAI parser, so it is affected by the same path.

### Anthropic parser

The Anthropic parser has the same full-chunk conversion:

```go
func (p *AnthropicParser) parseStreamResponse(chunk []byte) (*fwkrh.ParsedResponse, error) {
    usage := extractUsageStreaming(string(chunk))
    return &fwkrh.ParsedResponse{Usage: usage}, nil
}
```

It also allocates a slice through `strings.Split`, converts each SSE data line back to `[]byte`, and runs `json.Unmarshal` for ordinary content events that do not contain usage:

```go
lines := strings.Split(responseText, "\n")
for _, line := range lines {
    if !strings.HasPrefix(line, streamingRespPrefix) {
        continue
    }
    content := strings.TrimPrefix(line, streamingRespPrefix)

    var event struct {
        // ...
    }
    if err := json.Unmarshal([]byte(content), &event); err != nil {
        continue
    }
}
```

Source: [anthropic.go lines 200-223](https://github.com/llm-d/llm-d-router/blob/dd65342ee538ab76d4b9f7e1efb5d8e95385d06e/pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go#L200-L223)

Most Anthropic streaming events are `content_block_delta` events without usage data. Decoding each of those events only to discover that no usage is present creates unnecessary CPU and garbage-collection pressure.

### Allocation evidence

Go escape analysis on the current implementation reports that both `string(chunk)` conversions escape to the heap:

```text
openai/openai.go:191:40: string(chunk) escapes to heap
anthropic/anthropic.go:201:40: string(chunk) escapes to heap
anthropic/anthropic.go:222:35: ([]byte)(content) escapes to heap
```

This can be reproduced with:

```bash
go test -run '^$' -gcflags='all=-m=2' \
  ./pkg/epp/framework/plugins/requesthandling/parsers/openai \
  ./pkg/epp/framework/plugins/requesthandling/parsers/anthropic 2>&1 |
  grep -E 'string\(chunk\)|\(\[\]byte\)\(content\)'
```

### Prototype microbenchmark

A local prototype using byte-native SSE parsing produced the following `benchmem` results on an Apple M4 Pro with Go 1.26.6. These are parser microbenchmarks, not end-to-end EPP throughput claims.

| Parser path | Before | Prototype | Change |
|---|---:|---:|---:|
| OpenAI, no usage | 43.0 ns/op, 128 B/op, 2 allocs/op | 24.2 ns/op, 48 B/op, 1 alloc/op | 43% faster, 62% fewer bytes |
| OpenAI, with usage | 649.7 ns/op, 504 B/op, 10 allocs/op | 582.1 ns/op, 344 B/op, 8 allocs/op | 10% faster, 32% fewer bytes |
| Anthropic, no usage | 565.8 ns/op, 512 B/op, 10 allocs/op | 15.3 ns/op, 0 B/op, 0 allocs/op | 97% faster, allocations eliminated |
| Anthropic, with usage | 717.0 ns/op, 880 B/op, 18 allocs/op | 625.6 ns/op, 704 B/op, 15 allocs/op | 13% faster, 20% fewer bytes |

The large relative improvement in the Anthropic no-usage benchmark comes from replacing JSON decoding with a byte scan. It does not imply that the complete EPP or scheduling system becomes 37 times faster. The expected system-level benefit is lower allocation and GC pressure for high-output streaming workloads.

### Prior art

- [#1583](https://github.com/llm-d/llm-d-router/pull/1583) reduced allocations in OpenAI response usage parsing but retained the full-chunk byte/string conversion.
- [#2252](https://github.com/llm-d/llm-d-router/pull/2252) demonstrated that removing allocations from the per-response-chunk path can materially reduce GC activity and CPU usage.

### Acceptance criteria

- Existing OpenAI, vLLM HTTP, and Anthropic streaming behavior remains unchanged.
- Full-chunk byte/string conversions and unnecessary Anthropic JSON decoding are removed.
- Unit tests and benchmarks demonstrate reduced allocations.

muchengl:
**What type of PR is this?**

/kind cleanup

This PR reduces allocations in the OpenAI and Anthropic streaming response parsers, which run once for every Envoy `ext_proc` response chunk.

It parses SSE data directly from `[]byte`, removing avoidable byte/string conversions. Anthropic events without usage data now also skip JSON decoding. Existing parsing behavior remains unchanged, with additional correctness tests and benchmarks.

Microbenchmarks show:

| Path | Before | After |
|---|---:|---:|
| OpenAI, no usage | 43.0 ns, 128 B, 2 allocs | 24.2 ns, 48 B, 1 alloc |
| Anthropic, no usage | 565.8 ns, 512 B, 10 allocs | 15.3 ns, 0 B, 0 allocs |

These are parser microbenchmarks; the expected system-level benefit is reduced allocation and GC pressure for high-output streaming workloads.

Fixes #2423

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go
@@ -209,6 +209,7 @@ func extractUsageStreaming(responseBytes []byte) *fwkrh.Usage {
 	lines := bytes.SplitSeq(responseBytes, []byte("\n"))
 	for line := range lines {
 		content, ok := bytes.CutPrefix(line, []byte(streamingRespPrefix))
+		// Safe because only message_start/message_delta carry usage, both with a literal "usage" key.
 		if !ok || !bytes.Contains(content, []byte("usage")) {
 			continue
 		}
--- a/pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go
@@ -17,6 +17,7 @@ limitations under the License.
 package anthropic
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -198,19 +199,19 @@ func extractUsage(responseBytes []byte) (*fwkrh.Usage, error) {
 //	event: message_stop
 //	data: {"type":"message_stop"}
 func (p *AnthropicParser) parseStreamResponse(chunk []byte) (*fwkrh.ParsedResponse, error) {
-	usage := extractUsageStreaming(string(chunk))
+	usage := extractUsageStreaming(chunk)
 	return &fwkrh.ParsedResponse{Usage: usage}, nil
 }
 
-func extractUsageStreaming(responseText string) *fwkrh.Usage {
+func extractUsageStreaming(responseBytes []byte) *fwkrh.Usage {
 	var result *fwkrh.Usage
 
-	lines := strings.Split(responseText, "\n")
-	for _, line := range lines {
-		if !strings.HasPrefix(line, streamingRespPrefix) {
+	lines := bytes.SplitSeq(responseBytes, []byte("\n"))
+	for line := range lines {
+		content, ok := bytes.CutPrefix(line, []byte(streamingRespPrefix))
+		if !ok || !bytes.Contains(content, []byte("usage")) {
 			continue
 		}
-		content := strings.TrimPrefix(line, streamingRespPrefix)
 
 		var event struct {
 			Type    string `json:"type"`
@@ -219,7 +220,7 @@ func extractUsageStreaming(responseText string) *fwkrh.Usage {
 			} `json:"message"`
 			Usage map[string]any `json:"usage"`
 		}
-		if err := json.Unmarshal([]byte(content), &event); err != nil {
+		if err := json.Unmarshal(content, &event); err != nil {
 			continue
 		}
 
--- a/pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic_test.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic_test.go
@@ -605,6 +605,13 @@ func TestAnthropicParser_ParseResponse_Streaming(t *testing.T) {
 			},
 		},
 		{
+			name:  "content delta with usage text but no usage object",
+			chunk: []byte("event: content_block_delta\ndata: {\"type\":\"content_block_delta\",\"delta\":{\"type\":\"text_delta\",\"text\":\"usage\"}}"),
+			want: &fwkrh.ParsedResponse{
+				Usage: nil,
+			},
+		},
+		{
 			name:  "message_stop without usage",
 			chunk: []byte("event: message_stop\ndata: {\"type\":\"message_stop\"}"),
 			want: &fwkrh.ParsedResponse{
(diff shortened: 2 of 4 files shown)


pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go
```diff
@@ -198,19 +199,19 @@ func extractUsage(responseBytes []byte) (*fwkrh.Usage, error) {
...
+func extractUsageStreaming(responseBytes []byte) *fwkrh.Usage {
 	var result *fwkrh.Usage
 
-	lines := strings.Split(responseText, "\n")
-	for _, line := range lines {
-		if !strings.HasPrefix(line, streamingRespPrefix) {
+	lines := bytes.SplitSeq(responseBytes, []byte("\n"))
+	for line := range lines {
+		content, ok := bytes.CutPrefix(line, []byte(streamingRespPrefix))
+		if !ok || !bytes.Contains(content, []byte("usage")) {
```
elevran, line 213: nit:
worth a one-line comment noting this short-circuit is safe only because usage data is exclusively carried by message_start/message_delta events, both of which always contain a literal "usage" JSON key.
That invariant isn't obvious from the code alone and would help a future editor who adds a new event-type case without re-verifying it.

elevran (dismissed):
thanks - LGTM.
Minor nit inline.

pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go
muchengl: @elevran  Rebased onto latest main and added the one-line comment explaining why the `"usage"` short-circuit is safe for `message_start`/`message_delta` events. 
Thanks a lot for the careful review!

elevran (approved):

PR #2426, opened 2026-08-17
Point sim-epp token-producer at the render Service

tonibofarull:
**What type of PR is this?**

/kind bug

The sim-epp-kvcache EPP config set `token-producer.url` to
http://localhost:8000, but the kind dev env runs the vLLM render server
as a separate Service (vllm-render:8082), not a sidecar. The EPP pod has
no listener on localhost:8000, so every request failed tokenization
before the scorer ran and cache-aware routing never engaged.

Use `VLLM_RENDER_URL` (already exported, defaults to
http://vllm-render:8082) and add it to `kind-dev-env.sh`'s envsubst var
list so it is substituted into the config.

Fixes #

```release-note
NONE
```

### DIFF
--- a/deploy/config/sim-epp-kvcache-config.yaml
+++ b/deploy/config/sim-epp-kvcache-config.yaml
@@ -7,7 +7,7 @@ plugins:
   parameters:
     modelName: TinyLlama/TinyLlama-1.1B-Chat-v1.0  # replace to use a different model
     vllm:
-      url: http://localhost:8000
+      url: ${VLLM_RENDER_URL}  # substituted by kind-dev-env.sh envsubst (render Service)
 - type: precise-prefix-cache-producer
   parameters:
     tokenProcessorConfig:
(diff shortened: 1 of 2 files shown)


yamt (approved):
lgtm

yankay:
Thanks @tonibofarull 
lgtm

tonibofarull:
@yamt @yankay who can I ask for a review and merge?

tonibofarull:
Hi @ahg-g, may I get a review? :bow:

elevran (approved):
thanks - the new URL resolves to a real backend. The envsubst edit mirrors the whitelist already used at `scripts/kind-dev-env.sh:414` for the kustomize apply.
@tonibofarull CICD tests are broken and should be fixed before merge.

tonibofarull:
I did rebase as it seems to be a flaky test. Let's see if it passes now.

PR #2432, opened 2026-08-18
D to P KV transfer

kapiljain1989:
**What type of PR is this?**

feature

vLLM/NIXL supports bidirectional KV cache transfer via bidirectional_kv_xfer: true on the NixlConnector. In multi-turn agentic serving, a decode pod holds KV cache from prior conversation turns. When the next turn arrives, a prefill pod can pull that existing KV from the decode pod via NIXL RDMA instead of recomputing the full conversation history, significantly reducing TTFT for long multi-turn conversations.

The routing sidecar currently assumes a unidirectional flow: prefill produces KV, decode consumes. This design extends the nixlv2 connector to support the bidirectional pattern so that the sidecar can carry kv_transfer_params from decode responses across conversation turns and inject them into subsequent prefill requests, enabling the prefill engine to pull existing KV from the decode pod.

**Design Doc**
https://docs.google.com/document/d/1I_8zABJjs5BUxtIZbFuqO0YG7pdOByczZDCizBjQ2xI/edit?usp=sharing

Used Approach C from the design Doc

 
Fixes #

```release-note
NONE
```


Gregory-Pereira:
cc @vMaroon @nilig @vitabortnikov

kapiljain1989:
# Bidirectional KV Transfer Test Results

**Test Date:** 2026-08-18 09:48:29 EDT  
**Model:** meta-llama/Meta-Llama-3.1-8B-Instruct  
**Configuration:** 1 prefill pod + 1 decode pod (2 GPUs total)

---

## Executive Summary

✅ **Bidirectional KV transfer is working correctly**

Both test scenarios demonstrate successful Decode→Prefill (D→P) KV cache transfer:
- **Scenario 1**: Normal multi-turn flow with cache injection on Turn 2
- **Scenario 2**: Prefill pod restart between turns - cache resilience confirmed (zero impact)

The sidecar-based cache architecture proves robust: cache state persists on the decode pod regardless of prefill pod lifecycle.

---

## Scenario 1: Normal Bidirectional Flow

### Turn 1 - Establish KV Cache
**Session:** `scenario1-normal-1787060909`  
**Request:** "Write a function to calculate factorial"

**Response Analysis:**
```json
{
  "kv_transfer_params": {
    "do_remote_decode": true,
    "do_remote_prefill": false,
    "remote_block_ids": [[1,2,3,4,5,6]],
    "remote_num_tokens": 90,
    "remote_engine_id": "058e2f21-b108-4cd9-899b-19f0280b9082",
    "remote_host": "10.0.1.233",
    "remote_port": 5600
  }
}
```

**Sidecar Log:**
```
"cached kv_transfer_params from decode response"
```

**What happened:**
1. Prefill computed KV for 41 prompt tokens
2. Sent KV to decode via NIXL (P→D transfer)
3. Decode generated 50 tokens
4. **Decode sidecar intercepted response and cached `kv_transfer_params`** ✓

---

### Turn 2 - Use Cached KV (D→P Transfer)
**Session:** `scenario1-normal-1787060909` (same as Turn 1)  
**Request:** "Make it recursive"

**Response Analysis:**
```json
{
  "kv_transfer_params": {
    "do_remote_decode": true,
    "do_remote_prefill": false,
    "remote_block_ids": [[1,7,8,9,10,11]],
    "remote_num_tokens": 87,
    "remote_engine_id": "058e2f21-b108-4cd9-899b-19f0280b9082",
    "remote_host": "10.0.1.233",
    "remote_port": 5600
  }
}
```

**Sidecar Logs:**
```
"injected cached kv_transfer_params"  ← D→P transfer!
"cached kv_transfer_params from decode response"
```

**What happened:**
1. EPP routed request to same decode pod (session affinity)
2. **Decode sidecar retrieved cached params from Turn 1** ✓
3. **Sidecar injected params into prefill request** ✓
4. Prefill pulled existing KV from decode (D→P transfer)
5. Prefill computed only NEW tokens (38 prompt tokens vs 41 in Turn 1)
6. Decode sidecar cached Turn 2 response

**Evidence of D→P transfer:**
- Turn 2 shows **"injected cached kv_transfer_params"** log entry
- Prompt tokens reduced from 41 → 38 (conversation context reused)

---

## Scenario 2: Prefill Restart Between Turns

### Turn 1 - Establish KV Cache
**Session:** `scenario2-restart-1787060916`  
**Request:** "Write a function to sort an array"

**Response Analysis:**
```json
{
  "kv_transfer_params": {
    "remote_block_ids": [[1,2,6,12,13,14]],
    "remote_num_tokens": 91
  }
}
```

**Sidecar Log:**
```
"cached kv_transfer_params from decode response"
```

---

### Prefill Pod Restart
**Old pod:** `kv-test-vllm-prefill-7df9748479-vvmk2`  
**New pod:** `kv-test-vllm-prefill-7df9748479-9dljd`

**Restart completed successfully** - new pod ready in ~130 seconds

---

### Turn 2 - After Prefill Restart (Cache Resilience Test)
**Session:** `scenario2-restart-1787060916` (same as Turn 1)  
**Request:** "Add error handling"

**Response Analysis:**
```json
{
  "kv_transfer_params": {
    "remote_block_ids": [[1,15,16,17,18,19]],
    "remote_num_tokens": 87
  }
}
```

**Sidecar Logs:**
```
"injected cached kv_transfer_params"  ← D→P transfer STILL WORKS!
"cached kv_transfer_params from decode response"
```

**What happened:**
1. New prefill pod has no state (fresh vLLM engine)
2. EPP routed to same decode pod (session affinity intact)
3. **Decode sidecar still had cached params from Turn 1** ✓
4. **Sidecar injected cached params into new prefill pod's request** ✓
5. New prefill pod pulled KV from decode (D→P transfer)
6. Turn 2 completed successfully

**Critical finding:**
- **Prefill restart has ZERO impact on cache** ✓
- Cache persists on decode sidecar, not prefill
- D→P transfer works with brand new prefill pods

---

kapiljain1989:
Test 3: Below Threshold (55 tokens < 64)

  Session: threshold-test-123

  Turn 1:
  - Prompt: "Say hello"
  - Remote tokens: 55 (below threshold)
  - Log: "cached kv_transfer_params from decode response"

  Turn 2:
  - Prompt: "Now say goodbye"
  - Expected: Skip D→P transfer
  - Actual: ✅ "skipped D→P transfer (below threshold)"  with remote_tokens: 55, threshold: 64
  - Prefill recomputed locally instead of pulling from decode

nilig (commented):
The mechanism matches the contract in vLLM's NixlConnector guide (a stateful proxy carrying `kv_transfer_params` across turns), and the sidecar is the right home for it. Two blockers, then smaller things.

**1. Streaming responses break, and the feature never engages on them.** `newKVTransferParamsCaptureWriter` routes every write through a `bufio.Writer`, but its `Flush()` only flushes the underlying `ResponseWriter`, never the bufio — so on a streaming decode the client receives tokens in 4096-byte batches instead of per-chunk. Reproducible with a small test: one 57-byte SSE write plus `Flush()` delivers 0 bytes to the client; bytes first appear once 4096 accumulate. And `finalize` runs a single `json.Unmarshal` over the whole body, which fails on an SSE stream, so for streaming sessions nothing is ever cached — the feature silently no-ops for exactly the multi-turn chat traffic it targets, while still paying the buffering cost. `cachedTokensUsageRewriter` in the same package is the pattern to follow: `httpsnoop` wrapping that preserves `http.Flusher`, per-write passthrough, and an SSE-aware line parser. A side effect of the current shape: the bufio fragments >4KB non-streaming bodies into partial writes into the rewriter underneath, so the `cached_tokens` usage rewrite can silently no-op on large responses whenever a session header is present.

**2. The session header needs an owner before the flag is usable.** The sidecar reads `x-session-token` straight from the incoming request; nothing injects it or strips client-supplied values, so a client can name another session and have that session's blocks pulled into its own prefill. The engine will not catch this: the P side pulls `remote_block_ids` positionally with no token comparison, and the vLLM guide explicitly assumes "the router is able to detect such mismatch across turns" (vllm-project/vllm#43094) — that detection duty is precisely the EPP half of this feature. I understand the EPP work is planned as a follow-up; that makes it a declared dependency of enabling `--enable-bidirectional-kv-xfer`, so it needs a tracking issue covering header injection, stripping client-supplied values, and the cross-turn mismatch detection upstream expects from the router. Same-session mismatches are real too (reasoning models strip thinking traces from history, so turn N+1 is not a token-extension of turn N). Related: repo convention wants this PR linked to an issue as well.

**3. Wonder if we could replace the blind merge with an allowlist.** `for k, v := range cached { kvParams[k] = v }` copies whatever the decode engine returned. Today that includes `do_remote_prefill`/`do_remote_decode` with values that happen to match what the sidecar just set, and vLLM main has already added a `transfer_mode` key that would flow through unreviewed. Copying exactly the keys the P side consumes (`remote_engine_id`, `remote_request_id`, `remote_host`, `remote_port`, `remote_block_ids`, `remote_num_tokens`, `tp_size`, `remote_blocks_expiry_time`) keeps the protocol fields under sidecar control. `remote_blocks_expiry_time` must stay on the list — it is what lets the P-side worker decline expired reads instead of pulling freed blocks. The decoder-only path already rebuilds `kv_transfer_params` from scratch rather than passing foreign keys through, so there is precedent in the package.

**4. Composition with the OffloadingConnector P2P pull is undefined.** The injection runs immediately before `addP2PPullToPrefill` on the same map, and `Validate()` allows `--enable-p2p-pull` together with `--enable-bidirectional-kv-xfer`, so a request carrying both a cached session and `x-kv-cache-source-host-port` hands the prefiller two sources for the same history: `remote_kv_source` (CPU-tier pull) and the D-node `remote_*` params. `MultiConnector` will run both, so at minimum we transfer the overlap twice. I think the sidecar should pick one — e.g. skip the p2p source when the cached params cover at least as many tokens — and the flag help should say so. The two mechanisms are complementary rather than competing: D-side GPU blocks live only for `decoder_kv_blocks_ttl`, while the offload pull serves from CPU DRAM well past that and from any pod holding the prefix, so coverage-based precedence seems natural.

**5.** The feature exists only on the serial dispatch path; `runNIXLProtocolV2WriteParallel` neither captures nor injects, so with MoRI-IO parallel dispatch the flag silently does nothing. Fine to scope out, but the flag help should say it.

**6.** The sidecar threshold duplicates the engine's `kv_recompute_threshold` (both default 64, enforced on P in `pull_scheduler`). Dropping the sidecar knob might be simpler; if we keep it, the help text should note the engine enforces its own floor so the two stay in sync.

**7.** The defaults imply capacity the decode pool cannot hold: at a 70K-token agentic context, one session is roughly 23 GB of KV on a Llama-3.3-70B-class model, so an 8-GPU decode replica holds on the order of 15-25 such sessions in HBM — while the LRU defaults to 4096 sessions pinned for 480s. Sizing guidance in the flag docs would save operators from configuring a cache two orders beyond what HBM serves.

**8. Tests:** `startBidiKVProxy` switches the prefill mock to `KVConnectorSharedStorage` to get past the nixlv2 mock's `expected remote_engine_id:null` check. That check models the engine contract, so I think the mock should learn the bidirectional turn-2 shape instead of being bypassed. A streaming-response case would have caught point 1.

nits:
- `bidirectional_kv_capture.go` says "Copyright 2026 The Kubernetes Authors"; llm-d-original files use "The llm-d Authors" (#2450 is cleaning this up).
- The Unicode arrows in the flag help and the DEBUG log line ("D→P") — repo style is ASCII, "D to P".
- `mergeYAMLConfiguration` treats `bidirectional-recompute-threshold: 0` as unset, so "always pull" is only expressible via the CLI flag.
- Five new flags with no README or docs update.

kapiljain1989:
Hi @nilig , Thank you for the review.
I think I should work first on first two blocker.

kapiljain1989:
HI @nilig , for the blocker i have open https://github.com/llm-d/llm-d-router/pull/2831 could you please review?

if you are ok with these changes, I will commit the updated code here for review

PR #2434, opened 2026-08-18
config: consolidate discovery plugins under a single discovery section (follow up #2341)

pazqo:
Group endpoint discovery and peer discovery under dataLayer.discovery with endpoints and peers sub-fields instead of two separate top-level fields (dataLayer.discovery and dataLayer.peerDiscovery).

A migration shim in the config loader preserves backward compatibility:
- dataLayer.discovery.pluginRef migrates to discovery.endpoints.pluginRef
- dataLayer.peerDiscovery migrates to discovery.peers

An example can be seen in `pkg/epp/config/loader/configloader_test.go`: `Deprecated_discovery`, `TestLoadRawConfig`, `Deprecated_dataLayer`.

Follows the existing deprecation pattern used for parser and saturationDetector fields.

**What type of PR is this?**
/kind cleanup
/kind deprecation

Consolidates the two discovery plugin config fields (dataLayer.discovery and dataLayer.peerDiscovery) into a single dataLayer.discovery section with endpoints and peers sub-fields. A migration shim preserves backward compatibility for the old config format, following the existing deprecation pattern used for parser and saturationDetector.

Follow-up to https://github.com/llm-d/llm-d-router/pull/2341, as discussed with @evacchi and @elevran 

```release-note
The `dataLayer.discovery.pluginRef` config field is deprecated. Use `dataLayer.discovery.endpoints.pluginRef` instead. The old field is still accepted and automatically migrated. Peer discovery is now configured exclusively via `dataLayer.discovery.peers`.
```


apix/config/v1alpha1/endpointpickerconfig_types.go
```diff
@@ -234,15 +234,15 @@ type DataLayerConfig struct {
...
+	// Discovery groups configuration for all discovery plugins (endpoint discovery
+	// and peer discovery). If omitted, the EPP uses Kubernetes-based discovery for
+	// endpoints and disables peer discovery.
 	Discovery *DiscoveryConfig `json:"discovery,omitempty"`
 	// +optional
 	// PeerDiscovery specifies which PeerDiscovery plugin to use for discovering
 	// peer EPP replicas. If omitted, peer discovery is disabled.
+	//
+	// Deprecated: use discovery.peers instead. If both are set, the new field is used.
 	PeerDiscovery *PeerDiscoveryConfig `json:"peerDiscovery,omitempty"`
```
elevran: Peer discovery was introduced in #2341 and hasn't shipped in a release, so nobody depends on `dataLayer.peerDiscovery` yet. Rather than deprecating it (keeping the field + doc comment), consider just removing it now that `discovery.peers` exists - there's no real back-compat need.

pkg/epp/config/loader/configloader.go
```diff
@@ -143,6 +145,37 @@ func LoadRawConfig(configBytes []byte, logger logr.Logger) (*configapi.EndpointP
...
+	//nolint:staticcheck // SA1019: dl.Discovery.PluginRef is deprecated: use discovery.endpoints instead.
+	if dl.Discovery != nil && dl.Discovery.PluginRef != "" {
+		logger.Info("DEPRECATION: dataLayer.discovery.pluginRef is deprecated, use dataLayer.discovery.endpoints.pluginRef instead.")
+		if dl.Discovery.Endpoints == nil {
+			dl.Discovery.Endpoints = &configapi.EndpointDiscoveryConfig{PluginRef: dl.Discovery.PluginRef}
+		}
+	}
+
+	//nolint:staticcheck // SA1019: dl.PeerDiscovery is deprecated: use discovery.peers instead.
+	if dl.PeerDiscovery != nil {
```
elevran: Since `dataLayer.peerDiscovery` never released, this migration branch (and its "DEPRECATION:" log) isn't needed - only the `dl.Discovery.PluginRef` branch above corresponds to a real, in-use deprecated form.

elevran (commented):
Thanks for following-up on #2341 and consolidating the two discovery fields under one section. The endpoint-discovery migration path (old bare `pluginRef` -> `endpoints.pluginRef`) correctly preserves back-compat with a deprecation log, matching the existing parser/saturationDetector pattern.

Since peer discovery hasn't shipped in any release yet and has no real users, treating its rename as a "deprecation" (kept field, migration shim, log line, release note) seems like more machinery than needed for a pre-release rename - see inline comments. Suggest just renaming it outright and dropping the compat path for that one field, keeping the shim scoped to the endpoint-discovery field that's actually released.

This affects the PR body's release-note section as well and needs correcting to scope backward compatibility for `endpoints` field only.


Outside this PR diff: `pkg/epp/framework/interface/datalayer/peer.go`:69 still documents the old dataLayer.peerDiscovery.pluginRef path - worth updating for consistency once the field is renamed.

pazqo:
> Thanks for following-up on #2341 and consolidating the two discovery fields under one section. The endpoint-discovery migration path (old bare `pluginRef` -> `endpoints.pluginRef`) correctly preserves back-compat with a deprecation log, matching the existing parser/saturationDetector pattern.
> 
> Since peer discovery hasn't shipped in any release yet and has no real users, treating its rename as a "deprecation" (kept field, migration shim, log line, release note) seems like more machinery than needed for a pre-release rename - see inline comments. Suggest just renaming it outright and dropping the compat path for that one field, keeping the shim scoped to the endpoint-discovery field that's actually released.
> 
> This affects the PR body's release-note section as well and needs correcting to scope backward compatibility for `endpoints` field only.
> 
> Outside this PR diff: `pkg/epp/framework/interface/datalayer/peer.go`:69 still documents the old dataLayer.peerDiscovery.pluginRef path - worth updating for consistency once the field is renamed.

Yeah, that makes sense. I'll remove the un-necessary deprecation things and keep them on already shipped features. Thanks for the review. I pushed an updated version addressing your comments.

pkg/epp/config/loader/configloader.go
```diff
@@ -143,6 +145,25 @@ func LoadRawConfig(configBytes []byte, logger logr.Logger) (*configapi.EndpointP
...
+//	dataLayer.discovery.pluginRef -> dataLayer.discovery.endpoints.pluginRef
+func migrateDiscoveryConfig(logger logr.Logger, rawConfig *configapi.EndpointPickerConfig) {
+	if rawConfig.DataLayer == nil {
+		return
+	}
+	dl := rawConfig.DataLayer
+
+	//nolint:staticcheck // SA1019: dl.Discovery.PluginRef is deprecated: use discovery.endpoints instead.
+	if dl.Discovery != nil && dl.Discovery.PluginRef != "" {
+		logger.Info("DEPRECATION: dataLayer.discovery.pluginRef is deprecated, use dataLayer.discovery.endpoints.pluginRef instead.")
```
elevran: the existing saturationDetector and parser migrations both end their log message with "If both are set, the new field is used.". This migration should match. The code silently prefers `endpoints.pluginRef` when both are populated, so a user with both fields set gets no signal that one of them was ignored.

cmd/epp/runner/runner.go
```diff
@@ -259,7 +259,7 @@ func (r *Runner) Run(ctx context.Context) error {
 		setupLog.Error(err, "Failed to parse configuration")
 		return err
 	}
-	if rawConfig.DataLayer != nil && rawConfig.DataLayer.Discovery != nil {
+	if rawConfig.DataLayer != nil && rawConfig.DataLayer.Discovery != nil && rawConfig.DataLayer.Discovery.Endpoints != nil {
```
elevran: nit: the 3 nil checks repeat the long field path three times. A local could makes it read more cleanly (`dlCfg := rawConfig.DataLayer`). Pure sugar, not blocking.

elevran (commented):
One consistency gap in the new migration's log message; otherwise looks good.

elevran:
LGTM can merge after fixing deprecation message.

pkg/epp/config/loader/configloader.go
pazqo: Should be fixed now, thanks!

elevran (approved):

PR #2435, opened 2026-08-18
feat(coordinator): expose Prometheus metrics

dmitripikus:
**What type of PR is this?**
/kind feature

Adds Prometheus metrics to the coordinator. With EPP metrics alone, operators see per-leg sub-requests, not client requests, so they cannot answer request rate, latency, error mix, per-step wall time, upstream fan-out, or disaggregation path from the coordinator process itself.

Three commits map to the three families in the spec (#2277):

1. Endpoint + request family — new `pkg/coordinator/metrics` package, `/metrics` on its own port (default 9090, `--metrics-port`), and `request_total` / `request_error_total` / `request_duration_seconds` / `request_size_bytes` / `request_running`.
2. Step + upstream families — the pipeline executor brackets every step with `step_running` / `step_duration_seconds` / `step_errors_total` and classifies failures (`bad_request`, `upstream_4xx`, `upstream_5xx`, `internal`); each outbound call site emits `upstream_request_total` / `upstream_request_duration_seconds`.
3. Per-request outcomes — `execution_path_total`, `conditional_decode_probes_total`, `request_input_tokens`.

Every `model_name` label is capped at 1000 distinct values with an `other` overflow, matching EPP's cardinality guard.

Deviations from the WIP doc in #2277, per review feedback that has not yet landed in the doc:
- Drops `response_size_bytes`
- Renames `disagg_decision_total` to `execution_path_total` and its `decision_type` label to `path` (the coordinator observes rather than decides).
- Adds `request_input_tokens` (token IDs are available on `RequestContext.TokenIDs` after render).

Part of #2276

**Release note**:
```release-note
Coordinator now exposes Prometheus metrics on a dedicated /metrics endpoint (default port 9090, configurable via --metrics-port or server.metrics_port). Metrics cover request rate/latency/error mix, per-pipeline-step wall time and error classification, per-upstream call counts and latencies, execution path (decode-only / prefill-decode / encode-prefill-decode), conditional-decode probe outcomes, and render-derived input token count. Every model_name label is capped at 1000 distinct values with an "other" overflow.
```


pkg/coordinator/metrics/cardinality.go
```diff
@@ -0,0 +1,85 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import "sync"
+
+// The model_name label is populated from the request body, which no
```
roytman, line 23: The file is a near line-for-line reimplementation of `pkg/epp/metrics/cardinality.go`'s boundedLabel (same mutex-guarded map-with-cap-and-overflow design, ~85 duplicated lines) rather than reusing it. `pkg/common/observability/metrics` already exists as a shared home (both packages import it for `HelpMsgWithStability`) but the bounded-label logic wasn't moved there.

Since `pkg/epp/metrics` is unexported, direct reuse needs a small refactor — moving boundedLabel into the shared pkg/common/observability/metrics package that both sides already import. 

Another option is to leave it unchanged, but to open a refactoring issue and comment on it here. The refactoring can be done in a separate PR.

cmd/coordinator/main.go
```diff
@@ -97,36 +116,83 @@ func main() {
...
+}
+
+// serveMetrics stands up a Prometheus /metrics HTTP server on port and blocks
+// until ctx is cancelled or the underlying ListenAndServe returns
+// unexpectedly. On ctx cancellation the server is drained via Shutdown
+// bounded by metricsShutdownTimeout. Uses the shared controller-runtime
+// registry so every package that registers against it (this coordinator's
+// metrics, controller-runtime's process collectors) is exposed on the same
+// endpoint.
+func serveMetrics(ctx context.Context, port int) error {
```
roytman, line 179: No test coverage for new `run()/serveMetrics()`. This PR rewrites process shutdown from a single-server select-loop to an errgroup-based dual-server drain. There is no main_test.go anywhere in cmd/coordinator/, so this shutdown-safety-critical logic (the exact class of bug this PR just fixed) has zero automated regression protection. Every other package in this PR is thoroughly tested; this one file is the outlier.

Suggestion: add a minimal `main_test.go` covering: `MetricsPort <= 0` skips the metrics goroutine, and a failure in one server triggers a graceful drain of the other (this is directly testable by injecting a port collision or cancelling the context).

roytman (commented):
Thank you, @dmitripikus. I'm still working on the review, but I have 2 initial comments.

pkg/coordinator/config/config.go
```diff
@@ -40,7 +40,11 @@ const BytesPerMB = 1024 * 1024
 const DefaultMaxRequestBodySize = 64 // 64 MB
 
 type ServerConfig struct {
-	ListenAddr         string        `mapstructure:"listen_addr"`
+	ListenAddr string `mapstructure:"listen_addr"`
```
roytman: looks like a wrong formatting

pkg/coordinator/config/config.go
```diff
@@ -40,7 +40,11 @@ const BytesPerMB = 1024 * 1024
...
+	// MetricsPort is the port for the Prometheus /metrics endpoint. A
+	// non-positive value disables the endpoint, matching pkg/sidecar/proxy
```
roytman: we have some inconsistency here (can be resolved in a separate PR) 

- pkg/sidecar/proxy: `MetricsPort > 0` → enabled; 0 (default) or unset → disabled, falls back to MORIIO_METRICS_ADDR env var. Documented in the flag help text and README.
- pkg/coordinator (this PR): same rule, explicitly copied: cfg.MetricsPort > 0 starts the goroutine, non-positive logs "metrics endpoint disabled" and skips it. Its own comment states it's "matching pkg/sidecar/proxy semantics."
- cmd/epp/runner: No disable path at all. serveMetrics is unconditionally wired into the run group; --metrics-port defaults to 9090 and has no "0 disables" semantics in its flag help, its Options.Validate(), or the server code. Setting --metrics-port=0 binds ":0" (OS picks a random ephemeral port) rather than disabling anything; a negative value would fail ListenAndServe and — since EPP's run-group cancels all tasks when one fails — take down the whole process.

cmd/coordinator/main.go
```diff
@@ -97,36 +116,83 @@ func main() {
...
+	if err := run(srv, cfg.Server); err != nil {
 		log.Error(err, "server error")
 		os.Exit(1)
 	}
 }
 
-// serveUntilSignal starts srv and blocks until it exits or a signal is received.
-// On SIGTERM/SIGINT it initiates a graceful drain bounded by shutdownTimeout.
-func serveUntilSignal(srv *server.Server, shutdownTimeout time.Duration) error {
+// run starts the inference server and, when cfg.MetricsPort > 0, the
```
roytman: Calling the coordinator server an "inference server"  is a real naming problem. 

- Established meaning conflict. docs/discovery.md:338 already uses "inference server" to mean vLLM itself: "vLLM (or a compatible inference server) running on the target addresses." The coordinator's own architecture doc is careful never to call the coordinator "the inference server" — it says the coordinator "accepts inference requests" (`coordinator_architecture.md:3`) and drives an "inference request (OpenAI-compatible API)" through the pipeline. So "inference server" already has an established referent in this repo, and it's the vLLM backend, not the coordinator's frontend.
- Internal inconsistency. Nowhere else in the coordinator code is this server called "the inference server",  the type is just `server.Server` (`server.go:81`), `main()` logs "starting coordinator" and "failed to create server", and the handler is handleInference (named for the request type it handles, `/v1/chat/completions` etc., not for what the server is). Only run()'s doc comment and its three wrapped-error strings ("inference server: %w", "inference server shutdown: %w") introduce the term — purely to contrast it with "the metrics server" in that one function.

Fix is mechanical - in the file, swap "inference server" → "coordinator server" (or just "server")

pkg/coordinator/config/config.go
```diff
@@ -75,6 +79,7 @@ func Load(path string) (*Config, error) {
 
 	v.SetDefault("log_level", 2)
 	v.SetDefault("server.listen_addr", ":8080")
+	v.SetDefault("server.metrics_port", 9090)
```
roytman, line 79: we need to update the config example `config/coordinator/coordinator.yaml`

pkg/coordinator/metrics/metrics.go
```diff
@@ -0,0 +1,176 @@
...
+func withLabel(base []string, extra string) []string {
+	out := make([]string, 0, len(base)+1)
+	out = append(out, base...)
+	return append(out, extra)
+}
+
+var (
+
+	// generalLatencyBuckets covers durations from 5ms to 1 hour; identical to
+	// the EPP request-duration ladder so PromQL translates cleanly between the
```
roytman: yes, it is identical to `pkg/epp/metrics/metrics.go:60-64`

pkg/coordinator/metrics/metrics.go
```diff
@@ -0,0 +1,176 @@
...
+	// two components.
+	generalLatencyBuckets = []float64{
+		0.005, 0.025, 0.05, 0.1, 0.2, 0.4, 0.6, 0.8, 1.0, 1.25, 1.5, 2, 3, 4, 5, 6,
+		8, 10, 15, 20, 30, 45, 60, 120, 180, 240, 300, 360, 480, 600, 900, 1200,
+		1800, 2700, 3600,
+	}
+
+	// requestSizeBuckets ranges from 64 bytes to 1 GiB, matching the EPP
```
roytman: identical to `pkg/epp/metrics/llm_d_router_metrics.go:73-77`

pkg/coordinator/metrics/metrics.go
```diff
@@ -0,0 +1,176 @@
...
+	// request-size ladder. Wide enough for multimodal bodies with inlined
+	// image data.
+	requestSizeBuckets = []float64{
+		64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536,
+		131072, 262144, 524288, 1048576, 2097152, 4194304, 8388608,
+		16777216, 33554432, 67108864, 134217728, 268435456, 536870912, 1073741824,
+	}
+
+	// inputTokensBuckets is a power-of-two ladder from 1 to 1M input tokens;
```
roytman: Almost identical to EPP, fixes the typo `32778`→`32768`

pkg/coordinator/metrics/metrics.go
```diff
@@ -0,0 +1,176 @@
...
+		executionPathTotal,
+		conditionalDecodeProbesTotal,
+	}
+}
+
+// Register wires every coordinator metric onto reg. A collector already
+// present on reg is treated as success, so calling Register more than once
+// (e.g. across tests using a fresh prometheus.NewRegistry() each time) is
+// safe.
+func Register(reg prometheus.Registerer) error {
```
roytman, line 135: `Register/Reset`: are not a duplicate of EPP's `Register/Reset` - actually structurally here is better. EPP hand-repeats its ~60-metric list once in Register and again in Reset (`llm_d_router_metrics.go:424-503`, `506-578`), and its label composition uses raw append(modelLabels, "error_code") with no aliasing guard - exactly the hazard withLabel was built to prevent. 
The coordinator's single allCollectors() list feeding both functions, plus withLabel, is arguably a pattern EPP should adopt, not the reverse.

pkg/coordinator/pipeline/pipeline.go
```diff
@@ -52,6 +53,29 @@ func (e *UpstreamError) Error() string {
 	return fmt.Sprintf("%s: upstream returned HTTP %d", e.Step, e.StatusCode)
 }
 
+// UpstreamStreamedError signals that a streaming step (decode,
+// conditional-decode) saw an upstream failure after the response was already
+// committed to the client. The reverse proxy either streamed a 4xx/5xx body
```
roytman, line 58: Should we mention here that we exceed the 412 error for conditional-decode

pkg/coordinator/pipeline/pipeline.go
```diff
@@ -80,23 +105,96 @@ func (p *Pipeline) Execute(ctx context.Context, reqCtx *RequestContext) error {
...
 		}
-		logger.V(logutil.TRACE).Info("step starting", "step", step.Name())
+		name := step.Name()
+		logger.V(logutil.TRACE).Info("step starting", "step", name)
+		coordmetrics.IncStepRunning(name)
 		start := time.Now()
-		err := step.Execute(ctx, reqCtx)
-		timings[idx] = stepTiming{name: step.Name(), duration: time.Since(start)}
+		var err error
+		func() {
```
roytman: The current code is correct, but the idiomatic Go fix for "I need a defer scoped to one loop iteration" is to extract a named function rather than an inline IIFE — it's the same defer-scoping trick, just without the closure-capturing-an-outer-variable smell:
```go
func (p *Pipeline) runStep(ctx context.Context, reqCtx *RequestContext, step Step, idx int, timings []stepTiming) error {
	name := step.Name()
	coordmetrics.IncStepRunning(name)
	start := time.Now()
	defer func() {
		d := time.Since(start)
		coordmetrics.RecordStepDuration(name, d)
		coordmetrics.DecStepRunning(name)
		timings[idx] = stepTiming{name: name, duration: d}
		if r := recover(); r != nil {
			coordmetrics.IncStepErrorTotal(name, coordmetrics.ErrorCodeInternal)
			panic(r)
		}
	}()
	return step.Execute(ctx, reqCtx)
}
```
and the loop becomes:
```go
for idx, step := range p.steps {
	if err := ctx.Err(); err != nil {
		return fmt.Errorf("pipeline cancelled: %w", err)
	}
	logger.V(logutil.TRACE).Info("step starting", "step", step.Name())
	if err := p.runStep(ctx, reqCtx, step, idx, timings); err != nil {
		if errors.Is(err, ErrPipelineDone) {
			executed[step.Name()] = true
			return nil
		}
		coordmetrics.IncStepErrorTotal(step.Name(), coordmetrics.ClassifyErrorCode(err, ClassifyOpts))
		return fmt.Errorf("step %q failed: %w", step.Name(), err)
	}
	executed[step.Name()] = true
	logger.V(logutil.TRACE).Info("step complete", "step", step.Name())
}
```
Why this is better, not just different:
- No outer-variable mutation through a closure. The current code does `var err error; func() { ...; err = step.Execute(...) }()` — assigning into a variable declared outside the closure. A real function that just returns is the more direct idiom and is easier to read at a glance.
- Independently testable. `runStep` can be unit-tested directly (call it, panic a stub step, assert the gauge/duration/error recording) without going through the whole Execute loop.
- Same defer-scoping guarantee. A method's own body is exactly as valid a function boundary for defer as an anonymous IIFE — nothing is lost.

This is a pure readability refactor, not a correctness fix — the anonymous-function version in the PR is not buggy, this is just the more idiomatic shape for "defer must run once per loop iteration."

pkg/coordinator/metrics/cardinality.go
dmitripikus: Done. Moved boundedLabel up to pkg/common/observability/metrics.
The coordinator side is now a thin wrapper.
Thanks!

cmd/coordinator/main.go
dmitripikus: Done. Two tests in cmd/coordinator/main_test.go: MetricsPort=0 + cancel → run returns nil; metrics port pre-bound via test/framework/net.ReserveListener → run returns "metrics server: ..." and the inference address stops accepting connections. Small refactor: signal.NotifyContext moved into main() so tests can cancel run without sending SIGTERM.
Thanks!

pkg/coordinator/config/config.go
dmitripikus: Fixed. Moved the note to a trailing comment, so the block is one aligned run again and ListenAddr is untouched — the hunk is now purely additive.
Thanks!

pkg/coordinator/config/config.go
dmitripikus: For now I replaced wrong statement "matching pkg/sidecar/proxy semantics": coordinator matches the > 0 gate but not the MORIIO_METRICS_ADDR env-var fallback. Now the comment states the rule directly.

The broader inconsistency with cmd/epp/runner (no disable gate; --metrics-port=0 binds a random ephemeral port, negative crashes the whole errgroup) is a real bug worth its own issue + PR. I have not filed either yet — I'll open the issue if you agree with the semantics, i.e. align EPP with coordinator: <= 0 disables, log "metrics endpoint disabled", update --metrics-port help text; leave sidecar's env-var fallback alone as legacy.

cmd/coordinator/main.go
dmitripikus: I did renaming as proposed.

Thanks!

pkg/coordinator/config/config.go
dmitripikus: Added a commented metrics_port: 9090 block right after listen_addr in config/coordinator/coordinator.yaml, with the non-positive-disables rule, the --metrics-port flag override, and the COORDINATOR_SERVER_METRICS_PORT env tag — matching the yaml's own convention for optional fields with defaults.

pkg/coordinator/metrics/metrics.go
dmitripikus: Exported new GeneralLatencyBuckets in pkg/common/observability/metrics, and both pkg/coordinator/metrics and pkg/epp/metrics now reference it.

Thanks!

pkg/coordinator/metrics/metrics.go
dmitripikus: Fixed. The request-size numbers (64 bytes up to 1 GiB, 25 entries) now live in pkg/common/observability/metrics as RequestSizeBuckets, and all three callers point at it.

Thanks!

pkg/coordinator/metrics/metrics.go
dmitripikus: Fixed. The 19-number list for input-token and cached-token histograms is now metricsutil.TokenCountBuckets in pkg/common/observability/metrics, and five histograms point at it (coordinator's request_input_tokens, EPP's live request_input_tokens and request_cached_tokens, plus the two deprecated counterparts in metrics.go). The shared value uses 32768, so EPP adopts the coordinator's typo fix.

pkg/coordinator/metrics/metrics.go
dmitripikus: I agree. The coordinator's Register/Reset was written in a way that:

* one allCollectors() list drives both functions, so they can't drift
* the resettableCollector interface catches a missing Reset() at compile time
* withLabel allocates a fresh slice on every use, so sibling append calls can't share backing memory
I propose not to do changes here, but to open a separate issue proposing pkg/epp/metrics adopt the same pattern.

@roytman , What do you think?

pkg/coordinator/pipeline/pipeline.go
dmitripikus: Done. The conditional-decode block in the yaml now says both cases: on a cache hit the worker answers and encode/prefill are skipped; on a miss the worker returns HTTP 412 and the pipeline runs through encode and prefill as normal.

pkg/coordinator/pipeline/pipeline.go
dmitripikus: Done. Extracted the per-step body into a runStep method with the exact signature you proposed. The var err error write from inside a closure is gone, stepTiming moved to package scope

cmd/coordinator/main_test.go
```diff
@@ -0,0 +1,135 @@
...
+	case <-time.After(5 * time.Second):
+		t.Fatal("run did not return within 5s after cancel")
+	}
+}
+
+// A bind failure in the metrics goroutine must fault the errgroup and drain
+// the coordinator server; run's returned error surfaces the metrics-server
+// failure, and the coordinator socket must no longer accept connections.
+func TestRun_MetricsPortCollision_DrainsCoordinatorServer(t *testing.T) {
+	blocker, err := fwknet.ReserveListener()
```
roytman: This test hangs (times out at the 5s select) on macOS, and I'd guess on any environment where IPv6 dual-stack binds don't shadow IPv4. Reproduced directly:
```go
l1, _ := net.Listen("tcp", "127.0.0.1:0")
port := l1.Addr().(*net.TCPAddr).Port     // OS-assigned, e.g. 55044

l2, _ := net.Listen("tcp", fmt.Sprintf(":%d", port))
// l2 succeeds -> [::]:55044, no collision with l1's 127.0.0.1:55044
```

`fwknet.ReserveListener()` deliberately binds IPv4-loopback only ("Force IPv4 to prevent flakes on dual-stack CI"), but serveMetrics's `http.Server{Addr: ":<port>"}` binds the wildcard address. The test assumes these two collide on the same port number, but that's Linux-specific dual-stack behavior (IPV6_V6ONLY=0 by default), not something guaranteed by the net package. On a platform where the wildcard bind resolves to a separate IPv6-only socket, serveMetrics's ListenAndServe just succeeds quietly, gctx never gets cancelled by an error, and run() blocks forever — which is exactly the observed hang.

This passes in CI only because the runners are Linux. It'll hang for any contributor running the suite on macOS (and possibly in IPv6-disabled containers, depending on which way that cuts).

Fix: reserve the blocking port the same way serveMetrics actually binds it, so the collision is guaranteed regardless of platform:

```go
blocker, err := net.Listen("tcp", fmt.Sprintf(":%d", inferencePort)) // or however the free port is chosen
```
rather than going through `fwknet.ReserveListener()`'s IPv4-loopback bind.

dmitripikus: Fixed. The blocker is now bound with net.Listen("tcp", ":0") so it occupies the same wildcard socket serveMetrics will try to bind, making the collision guaranteed regardless of the platform's dual-stack behavior.
Thanks!

dmitripikus:
Regarding Suppressed comment #1 — no endpoint test for /metrics:
The /metrics wiring mirrors [cmd/epp/runner/runner.go:1224](https://github.com/llm-d/llm-d-router/blob/main/cmd/epp/runner/runner.go#L1224), which also has no endpoint test; metric registration is covered in pkg/coordinator/metrics/.
So no actions taken in this PR

Regarding Suppressed #2 - duplication of @roytman 's comment. Fixed.

Regarding Suppressed #3 - PR description is updated

Regarding Suppressed #4 - Leaving the reference to 'docs/metrics.coord.md' in place so it resolves once that doc lands.

Regarding Suppressed #5 - Fixed. Added "error" to the outcome list so the contract matches ProbeResult* constants and the metric's Help string. 

Regarding Suppressed #6 + #7 - Fixed. The step now returns *pipeline.UpstreamError via the shared helper, so image-fetch 4xx/5xx classify as upstream_4xx/upstream_5xx (not internal)

cmd/coordinator/main.go
```diff
@@ -97,36 +116,82 @@ func main() {
...
+		ReadHeaderTimeout: 5 * time.Second,
+	}
+	go func() {
+		<-ctx.Done()
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), metricsShutdownTimeout)
 		defer cancel()
-		if err := srv.Shutdown(shutdownCtx); err != nil {
-			return err
-		}
+		_ = srv.Shutdown(shutdownCtx)
```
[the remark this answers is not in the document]
dmitripikus: Fixed. serveMetrics now tracks the shutdown goroutine with a done channel and blocks on it before returning, so an active /metrics scrape completes (or times out at metricsShutdownTimeout) before the process can exit. 

Thanks!

pkg/coordinator/config/config.go
```diff
@@ -75,6 +76,7 @@ func Load(path string) (*Config, error) {
...
```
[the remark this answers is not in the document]
dmitripikus: Added a named metrics/9090 container port and Service port on the coordinator, plus a coordinator-metrics ServiceMonitor under deploy/components/monitoring/, mirroring the EPP shape.
Thanks!

dmitripikus:
Coordinator metrics are open by default (specified both in code and documentation)
Default port is 9090.

Thanks!

pkg/coordinator/pipeline/pipeline.go
```diff
@@ -80,23 +108,103 @@ func (p *Pipeline) Execute(ctx context.Context, reqCtx *RequestContext) error {
...
+
+		if path, ok := classifyExecutionPath(executed); ok {
+			coordmetrics.IncExecutionPath(reqCtx.Model, path)
+		}
+		// Render populates TokenIDs on success (chat, completions, and generate
+		// paths); recording only when non-empty keeps failed-before-render
+		// requests out of the histogram.
+		if n := len(reqCtx.TokenIDs); n > 0 {
+			coordmetrics.RecordRequestInputTokens(reqCtx.Model, n)
+		}
```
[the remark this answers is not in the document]
dmitripikus: The gate now checks whether the render step ran, not whether TokenIDs is non-empty, so a valid empty-prompt completions request is observed at 0. Extended TestExecute_InputTokensRecordedOnlyWhenPopulated with a zero-length case.

Thanks!

pkg/coordinator/pipeline/pipeline.go
```diff
@@ -80,23 +108,103 @@ func (p *Pipeline) Execute(ctx context.Context, reqCtx *RequestContext) error {
...
+				// error, and the step's work counts toward the execution path.
+				executed[name] = true
 				return nil
 			}
-			return fmt.Errorf("step %q failed: %w", step.Name(), err)
+			coordmetrics.IncStepErrorTotal(name, coordmetrics.ClassifyErrorCode(err, ClassifyOpts))
+			return fmt.Errorf("step %q failed: %w", name, err)
 		}
-		logger.V(logutil.TRACE).Info("step complete", "step", step.Name())
+		executed[name] = true
```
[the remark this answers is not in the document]
dmitripikus: The pipeline now tracks started (any step that entered runStep) separately from executed (succeeded / clean-exit) and feeds classifyExecutionPath the started map, so a request that reaches decode contributes to execution_path_total even when decode returns an upstream error.

pkg/coordinator/server/handlers.go
```diff
@@ -44,30 +45,71 @@ import (
...
 		return
 	}
 	parseDuration := time.Since(parseStart)
 
 	stream, _ := parsed[reqcommon.FieldStream].(bool)
-	model, _ := parsed["model"].(string)
+	if m, ok := parsed["model"].(string); ok && m != "" {
+		model = m
+	}
+	coordmetrics.RecordRequestSize(model, len(body))
```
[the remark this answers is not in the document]
dmitripikus: request_size_bytes is now recorded in the entry-time defer, so read failures, 413, invalid JSON, and null bodies each emit a sample

pkg/coordinator/steps/conditional_decode.go
```diff
@@ -68,10 +70,25 @@ func (s *ConditionalDecodeStep) Execute(ctx context.Context, reqCtx *pipeline.Re
 	}
 
 	var cacheMiss bool
-	proxy := newDecodeProxy(logger, s.gwClient.Transport(), func(resp *http.Response) error {
-		if resp.StatusCode == http.StatusPreconditionFailed {
+	coordmetrics.IncUpstreamRequestTotal(coordmetrics.UpstreamConditionalDecode)
+	transport := &timedRoundTripper{
+		inner: s.gwClient.Transport(),
```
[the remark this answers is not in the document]
dmitripikus: Fixed. The call-site guard alone would not have caught the panic. gateway.NewWithTransport(nil, ...) was passing a nil *http.Transport into a RoundTripper interface field, and Go stores that as a non-nil interface holding a nil pointer. So Transport() did not return nil — it returned an interface that skipped every == nil check and then panicked on the nil pointer inside RoundTrip. That is also why http.Client's own default-transport fallback was already broken for a nil-transport client. The fix has two parts: NewWithTransport now stores a plain nil interface when the caller passes nil, and the call site falls back to http.DefaultTransport. Regression test: TestConditionalDecodeStep_NilClientTransport.

pkg/coordinator/steps/decode.go
```diff
@@ -76,8 +79,21 @@ func (s *DecodeStep) Execute(ctx context.Context, reqCtx *pipeline.RequestContex
 		return err
 	}
 
-	proxy := newDecodeProxy(logger, s.gwClient.Transport(), nil)
+	coordmetrics.IncUpstreamRequestTotal(coordmetrics.UpstreamDecode)
+	transport := &timedRoundTripper{
+		inner: s.gwClient.Transport(),
```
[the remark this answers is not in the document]
dmitripikus: Fixed. Same two-part fix as on the conditional-decode thread: NewWithTransport normalizes a nil *http.Transport to a plain nil interface, and the call site falls back to http.DefaultTransport. Without the NewWithTransport part, the call-site == nil check would still miss the typed-nil interface and panic in RoundTrip. Regression test: TestDecodeStep_NilClientTransport.

roytman (dismissed):
Thank you, @dmitripikus .

pkg/coordinator/metrics/metrics.go
```diff
@@ -0,0 +1,148 @@
...
+	ErrorCodeUpstream4xx = "upstream_4xx"
+	ErrorCodeUpstream5xx = "upstream_5xx"
+	ErrorCodeInternal    = "internal"
+)
+
+// Upstream label values for the upstream_request_* metrics. Step names come
+// from each step file's own StepName constant (pkg/coordinator/steps/*.go).
+const (
+	UpstreamRender            = "render"
+	UpstreamMediaFetch        = "media-fetch"
```
gyliu513: label mismatch: `UpstreamMediaFetch = "media-fetch"` but its step is `replace-media-urls`, breaking the step/upstream label-join convention every other step follows. Do you want to keep step and label to be consistent?

pkg/coordinator/steps/conditional_decode.go
```diff
@@ -68,10 +70,32 @@ func (s *ConditionalDecodeStep) Execute(ctx context.Context, reqCtx *pipeline.Re
...
+	if inner == nil {
+		inner = http.DefaultTransport
+	}
+	transport := &timedRoundTripper{
+		inner: inner,
+		record: func(d time.Duration) {
+			coordmetrics.RecordUpstreamRequestDuration(coordmetrics.UpstreamConditionalDecode, d)
+		},
+	}
+	proxy, out := newDecodeProxy(logger, transport, func(resp *http.Response) error {
```
gyliu513, line 73: the probe metric only increments in `ModifyResponse`, so a transport-level failure (no response received) is counted in `upstream_request_total` but silently missing from `conditional_decode_probes_total`.

pkg/coordinator/metrics/classify.go
```diff
@@ -0,0 +1,59 @@
...
+	// upstream error"; classification falls through.
+	IsUpstream func(err error) (status int, ok bool)
+}
+
+// ClassifyErrorCode maps a pipeline error to the error_code label emitted on
+// request_error_total and step_errors_total. Both metric families use the
+// same label vocabulary, so the two call sites share this function to stay
+// consistent — a new bucket or a shifted status band is one edit, both
+// metrics pick it up.
+func ClassifyErrorCode(err error, opts ClassifyOptions) string {
```
gyliu513, line 44: no case for upstream status `0` (transport failure), so it falls into `ErrorCodeInternal`, the same bucket as real app bugs/panics. This breaks the `internal = our code is broken` alerting semantics.

pkg/coordinator/metrics/metrics.go
dmitripikus: Renamed to UpstreamReplaceMediaURLs = "replace-media-urls" so the label matches ReplaceMediaURLsStepName, keeping the step/upstream label-join consistent with every other step.

Thanks!

pkg/coordinator/steps/conditional_decode.go
dmitripikus: Fixed. I introduced ProbeResultTransportError = "transport_error" and incremented it in the out.TransportErr != nil branch. I also added TestConditionalDecodeStep_TransportError or testing this.

Thanks!

pkg/coordinator/metrics/classify.go
dmitripikus: Fixed. Added ErrorCodeUpstreamTransport = "upstream_transport" and routed (status=0, ok=true) into it, so transport failures no longer land under error_code="internal". 

Thanks!

dmitripikus:
@roytman commented on this PR:

**Duplicated upstream-transport instrumentation between decode.go and conditional_decode.go**
pkg/coordinator/steps/decode.go:82-96 and pkg/coordinator/steps/conditional_decode.go:73-87 build the same instrumented transport nearly verbatim — same nil-fallback comment, same IncUpstreamRequestTotal + timedRoundTripper construction, differing only in the coordmetrics.Upstream* constant:

```
coordmetrics.IncUpstreamRequestTotal(coordmetrics.UpstreamDecode) // or UpstreamConditionalDecode
// The wrapper below calls inner.RoundTrip unconditionally; fall back to
// http.DefaultTransport when the client was built without a transport,
// matching http.Client.Do's own behavior.
inner := s.gwClient.Transport()
if inner == nil {
    inner = http.DefaultTransport
}
transport := &timedRoundTripper{
    inner: inner,
    record: func(d time.Duration) {
        coordmetrics.RecordUpstreamRequestDuration(coordmetrics.UpstreamDecode /* or ...ConditionalDecode */, d)
    },
}

```

Worth pulling into a small helper next to timedRoundTripper in decode_proxy.go, e.g.:

```
// instrumentedTransport wraps base in a timedRoundTripper that records
// upstream_request_total/upstream_request_duration_seconds for upstream,
// falling back to http.DefaultTransport when base is nil (matching
// http.Client.Do's own behavior).
func instrumentedTransport(base http.RoundTripper, upstream string) http.RoundTripper {
    if base == nil {
        base = http.DefaultTransport
    }
    coordmetrics.IncUpstreamRequestTotal(upstream)
    return &timedRoundTripper{
        inner: base,
        record: func(d time.Duration) {
            coordmetrics.RecordUpstreamRequestDuration(upstream, d)
        },
    }
}
```

Then both call sites collapse to one line each: transport := instrumentedTransport(s.gwClient.Transport(), coordmetrics.UpstreamDecode) (and ...UpstreamConditionalDecode in the other file).
Non-blocking — the rest of the PR's instrumentation (the UpstreamCall helper for the non-streaming steps, the shared BoundedLabel cardinality extraction, the centralized Pipeline.runStep) is well-factored, this is just the one spot where the streaming-step case duplicated itself.

dmitripikus:
Regarding duplicated upstream-transport instrumentation between decode.go and conditional_decode.go:

Extracted as instrumentedTransport in decode_proxy.go. Both call sites collapse to one line, time import dropped from decode.go and conditional_decode.go

Thanks!

gyliu513:
lgtm. thanks @dmitripikus and @roytman

roytman (dismissed):

roytman (approved):

vMaroon (approved):
None-coordinator changes look good

PR #2437, opened 2026-08-18
fix(epp): merge streaming usage per field instead of replacing it

### LINKED ISSUE #2431: Anthropic parser: message_delta chunk overwrites the prompt tokens captured from message_start
**What happened**:

The Anthropic streaming format splits usage across two SSE events: `message_start` carries `input_tokens` and `cache_read_input_tokens`, and `message_delta` carries `output_tokens` at the end of the stream (format comment at `pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go:190-199`).

`extractUsageStreaming` (anthropic.go:205-258) merges the two events only when they appear in the same parse call; its `result` is local to the call. Envoy delivers the response body to the EPP in many ext_proc chunks, and the message content streams between the two events, so under normal streaming they arrive in different calls.

The accumulation site replaces the whole struct: `reqCtx.Usage = *parsedResp.Usage` (`pkg/epp/handlers/response.go:81`). The chunk carrying `message_start` sets `Usage{PromptTokens: N, PromptTokenDetails: ...}`. The later chunk carrying `message_delta` parses to `Usage{CompletionTokens: M}` with zero prompt fields, and the assignment discards `N` and the cached-token detail. The final recorded usage for an Anthropic stream is `{PromptTokens: 0, CompletionTokens: M}`.

The final usage for an Anthropic stream retains the prompt tokens and cached-token detail from `message_start` alongside the completion tokens from `message_delta`, so every ResponseStreaming plugin invocation after the `message_delta` chunk, including the end-of-stream invocation, sees the full usage.

Drive an Anthropic streamed response through `StreamingServer.HandleResponseBody` as two chunks: the first containing the `message_start` event, the second containing the `message_delta` event. After the second chunk, `reqCtx.Usage.PromptTokens` is zero.

The parser table tests (anthropic_test.go, streaming cases) feed both events to a single `ParseResponse` call, where the call-local merge works, so they pass. No test drives two chunks through the handler and asserts that prompt tokens survive the second chunk. That test is the reproduction: it fails on the current code.

Impact:

- Every ResponseStreaming plugin invocation after the `message_delta` chunk sees `Response.Usage.PromptTokens == 0`, including the end-of-stream invocation. The program-aware fairness plugin (LAS strategy) computes attained service from those fields, so Anthropic streams are charged for output only.
- `requestattributereporter` evaluates its CEL expression over the final usage, so terminal records for Anthropic streams report zero prompt tokens and zero cached tokens.
- Prometheus token metrics are unaffected: `RecordInputTokens` fires with `N` on the `message_start` chunk, and the zero-valued call on the `message_delta` chunk is dropped by the `size > 0` guard (`pkg/epp/metrics/metrics.go:631-638`).

Proposed fix: merge per field at the accumulation site instead of replacing the struct. Copy `PromptTokens`, `CompletionTokens`, and `PromptTokenDetails` from the parsed usage only when the incoming value is non-zero, then recompute `TotalTokens` from the merged fields unless the parsed block carried its own total.

Parsers are stateless and shared across requests, so cross-chunk state belongs on `RequestContext`; fixing it inside the Anthropic parser would require adding per-request state to the parser interface. The OpenAI-family parsers emit usage once per stream with all fields populated, and non-streamed responses parse in one call, so a non-zero-field merge leaves their behavior unchanged.

**Environment**:

- Kubernetes version (use `kubectl version`): N/A, code-level bug
- llm-d-router version: `main` @ `7987423a`
- Install tools: N/A
- Others: N/A

himanshu1573:
**What type of PR is this?**
/kind bug

The Anthropic streaming format splits usage across two SSE events: `message_start` carries `input_tokens` and `cache_read_input_tokens`, and `message_delta` carries `output_tokens` at the end of the stream. Envoy delivers the response body to the EPP in many ext_proc chunks and the message content streams between the two events, so they reach `ParseResponse` in different calls and the parser's call-local merge does not apply.

`reqCtx.Usage = *parsedResp.Usage` then replaced the whole struct, so the chunk carrying `message_delta` discarded the prompt tokens and the cached-token detail recorded by `message_start`. The final usage for an Anthropic stream was `{PromptTokens: 0, CompletionTokens: M}`.

This merges per field at the accumulation site, as proposed in the issue. Cross-chunk state stays on `RequestContext`: parsers are stateless and shared across requests, so keeping it there would mean adding per-request state to the parser interface.

Two details worth review:

- **`TotalTokens`.** A non-zero-wins merge is wrong for this field on its own, because the Anthropic `message_delta` parses to a non-zero but partial total of `0 + output_tokens`. A block reporting both prompt and completion tokens keeps the total it came with (a passthrough for the OpenAI-family parsers, which report all fields at once); a partial block has the total derived from the merged fields.
- **Token metrics.** `RecordInputTokens` and friends are histograms and were called with the post-assignment `reqCtx.Usage`. Once the merged struct carries the prompt tokens on the `message_delta` chunk, recording from it would `Observe` them a second time, and `RecordPromptCachedTokens` has no `size > 0` guard to absorb it. The three calls now read the usage parsed from the current chunk, which keeps one observation per request. The added test asserts the sample counts, and fails with `expected: 0x1, actual: 0x2` if the calls are pointed back at the merged struct.

The OpenAI-family parsers emit usage once per stream with every field populated, and non-streamed responses parse in one call, so their behavior is unchanged.

Fixes #2431

**Test plan**:

- [x] `TestHandleResponseBodyModelStreaming_AnthropicUsageAccumulation` drives a four-chunk Anthropic stream through `HandleResponseBody` and asserts the final usage keeps the prompt tokens and cached-token detail from `message_start` alongside the completion tokens from `message_delta`, and that each token histogram observes exactly one sample. It fails on the unfixed code with `PromptTokens: 0`, matching the report.
- [x] `make presubmit`
- [x] `go test ./pkg/epp/handlers/... ./pkg/epp/framework/plugins/requesthandling/...`

```release-note
Fixed prompt and cached token counts being dropped from the recorded usage of streamed Anthropic responses, which reported zero prompt tokens to response plugins and terminal request records.
```

### DIFF
--- a/pkg/epp/handlers/response.go
+++ b/pkg/epp/handlers/response.go
@@ -78,11 +78,13 @@ func (s *StreamingServer) HandleResponseBody(ctx context.Context, reqCtx *Reques
 		}
 	}
 	if parsedResp != nil && parsedResp.Usage != nil {
-		reqCtx.Usage = *parsedResp.Usage
-		metrics.RecordInputTokens(reqCtx.IncomingModelName, reqCtx.TargetModelName, fairnessID, priority, reqCtx.Usage.PromptTokens)
-		metrics.RecordOutputTokens(reqCtx.IncomingModelName, reqCtx.TargetModelName, fairnessID, priority, reqCtx.Usage.CompletionTokens)
-		if reqCtx.Usage.PromptTokenDetails != nil {
-			metrics.RecordPromptCachedTokens(reqCtx.IncomingModelName, reqCtx.TargetModelName, fairnessID, priority, reqCtx.Usage.PromptTokenDetails.CachedTokens)
+		mergeUsage(&reqCtx.Usage, *parsedResp.Usage)
+		// Metrics observe the values this chunk carried, not the accumulated ones: a field
+		// already reported by an earlier chunk would otherwise be observed a second time.
+		metrics.RecordInputTokens(reqCtx.IncomingModelName, reqCtx.TargetModelName, fairnessID, priority, parsedResp.Usage.PromptTokens)
+		metrics.RecordOutputTokens(reqCtx.IncomingModelName, reqCtx.TargetModelName, fairnessID, priority, parsedResp.Usage.CompletionTokens)
+		if parsedResp.Usage.PromptTokenDetails != nil {
+			metrics.RecordPromptCachedTokens(reqCtx.IncomingModelName, reqCtx.TargetModelName, fairnessID, priority, parsedResp.Usage.PromptTokenDetails.CachedTokens)
 		}
 	}
 	if endOfStream {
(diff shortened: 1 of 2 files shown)


elevran (approved):
LGTM.

PR #2442, opened 2026-08-18
perf(epp): parse only metric families consumed by extractors

### LINKED ISSUE #2441: perf(epp): reduce scrape overhead by filtering unused metric families
/kind performance

**TL;DR**
EPP parses every model server's full /metrics response 20 times a second and reads five families out of it. Filtering the text before it reaches the parser frees ~1.4 cores at 100 endpoints. End-to-end at 8,000 req/s: CPU -50%, resident heap -32%, allocation rate -48%, p99.99 latency -88%. p50 unchanged. Four files, no API or config change. PR #2442.

## What happened

The metrics data source parses every family in a model server's `/metrics` response before extractors select the values they need.

**The full response is materialized**

The parser stores its output as a map of generated protobuf objects:

```go
type PrometheusMetricMap = map[string]*dto.MetricFamily
```

[`parseMetrics`](https://github.com/llm-d/llm-d-router/blob/620045bb19a95b088dc53f046c78a2d712b90a76/pkg/epp/framework/plugins/datalayer/source/metrics/datasource.go#L117) calls `TextToMetricFamilies`, which unconditionally builds this pointer-heavy object graph for the complete response.

**Only five families are consumed**

The [vLLM mapping](https://github.com/llm-d/llm-d-router/blob/620045bb19a95b088dc53f046c78a2d712b90a76/pkg/epp/framework/plugins/datalayer/extractor/metrics/factories.go#L93-L97) reads `vllm:num_requests_waiting`, `vllm:num_requests_running`, `vllm:kv_cache_usage_perc`, `vllm:lora_requests_info`, and `vllm:cache_config_info`. Selection happens only after the full map exists.

For the representative 42 KB fixture, the parser materializes 39 families, 40 metrics, 72 labels, 300 histogram buckets, and 1,368 struct pointers. **Extractors retain 5 families, 5 metrics, 11 labels, no buckets, and 63 pointers: only 4.6% of the pointer graph.**

**Tokenizer work dominates**

A full parse makes 12,581 allocations per scrape. Roughly 79% come from the tokenizer turning metric names, label names, and label values into strings. The cost therefore tracks response size: unused histograms, counters, and `python_*` or `process_*` families are tokenized and allocated before being discarded. Filtering before `expfmt` avoids that work without changing the parsed data model.

## Why it matters

Each endpoint is polled at the default `--refresh-metrics-interval=50ms` ([source](https://github.com/llm-d/llm-d-router/blob/620045bb19a95b088dc53f046c78a2d712b90a76/pkg/epp/server/options.go#L140)), or 20 scrapes per second. Shipped configurations do not override it, so the cost scales with fleet size and is paid even while the endpoint picker is idle.

**Synthetic scrape cost**

On a 42 KB vLLM scrape (Apple M4 Pro, Go 1.26.6), 100 endpoints consume 1.43 cores and generate 868 MB/s of garbage with the full parse, versus 0.056 cores and 20 MB/s after filtering. At 1,000 endpoints, the corresponding figures are 14.3 versus 0.56 cores and 8.7 GB/s versus 198 MB/s. Whole-process CPU includes concurrent GC; full methodology and raw results accompany the PR.

**End-to-end effect**

With 100 endpoints at 20 Hz, `GOMAXPROCS=4`, and 8,000 requests/s through the real data-source, extractor, scheduler, and scorer paths:

| | current full parse | with filtering | change |
|---|---:|---:|---:|
| CPU | 2.66 cores | 1.32 cores | **-50.3%** |
| resident heap, mean | 688 MB | 470 MB | **-31.7%** |
| allocation rate | 1,716 MB/s | 896 MB/s | **-47.8%** |
| latency, mean | 114.0 us | 86.5 us | -24.1% |
| latency, p90 | 171.7 us | 92.3 us | **-46.3%** |
| latency, p99.99 | 22.6 ms | 2.8 ms | **-87.5%** |

This frees 1.34 cores from the 4-core budget, or 668 us per scrape, within 2.8% of the microbenchmark result. **p50 remains unchanged** at 73 us because scraping is off the request path, and p99 is noisy across repetitions, so no p99 claim is made. Under saturation, served throughput rises from 18,336 to 23,837 requests/s (**+30%**) and CPU per request drops 33%.

## Proposal

Let extractors declare the metric families they read, then filter each scrape to the union of those declarations before parsing. The names already exist through `Mapping.MetricNames()` and `MappingRegistry.Mappings()` and can be exposed through a small optional interface:

```go
type FamilyNamer interface {
	MetricNames() []string
}
```

The data source filters the raw exposition text before passing it to `expfmt`. If a bound extractor does not implement `FamilyNamer`, the source keeps everything, so existing extractors retain the current behavior. The filter compares bytes and uses a pooled buffer, preventing discarded families from reaching the tokenizer.

**Alternatives considered**

Pooling `expfmt.TextParser` does not remove its dominant per-token allocations. Protobuf exposition requires model-server support and still decodes unused families. Filtering is vendor-independent and remains useful with either approach.

**Correctness**

The filter handles optional headers, bare samples, and histogram or summary suffixes. `TestFilteredParseMatchesFullParse` verifies that every retained family produces the same `dto.MetricFamily` as a full parse.

## Prior art

This is not a new observation. The project's own profiling found it and quantified it.

Issue #1906 (`[Performance] run pprof and analyze the results`) carries a profiling report of EPP under a ten-replica simulator workload. Its CPU accounting table attributes 12.5% cumulative CPU to `prometheus/common/expfmt`. Its improvement section is more specific:

> **Opportunity 2: Binary OpenMetrics Scraping / Buffer Recycling**
>
> **Problem**: Polling plain-text `/metrics` from backend pods consumes **14.5% of cumulative CPU time** and drives **~51% of total heap memory allocations** (~6.08 GB allocated over 30s).
>
> **Recommendation**: Transition backend metrics polling to protobuf/binary OpenMetrics format or reuse `expfmt.TextParser` instances via a `sync.Pool` to eliminate repetitive heap allocation.

That report's 51% of heap allocations and the 47.8% reduction in allocation rate measured above are independent measurements of the same effect, and they agree.

The copy committed at [`test/perf/config/studyanalysis/pprof_analysis_baseline_vs_precise.md`](https://github.com/llm-d/llm-d-router/blob/620045bb19a95b088dc53f046c78a2d712b90a76/test/perf/config/studyanalysis/pprof_analysis_baseline_vs_precise.md) ends after the "What Is Optimized" section, so the improvement opportunities exist only in the issue thread. #1906 is closed. Its JSON-parsing opportunity was discussed and declined as not worth the architectural cost; this one drew no follow-up issue or PR, and neither recommendation is implemented in the tree.

Two related threads:

- #1109 lists "Reduce or cache datalayer metrics parsing work" as the third of three improvement areas. Its own follow-up measurement re-attributes that issue's latency to request body streaming, so the metrics item was not pursued there.
- #1566 reports 7126m (7.1 cores) of idle EPP CPU with 100 simulator pods. The measurement above accounts for roughly 1.4 of those cores. The remainder is elsewhere in the collection path, so this change reduces that number without closing the issue on its own.

## Scope

Four files change:

- `source/metrics/filter.go` (new): the family selector and the text filter.
- [`source/metrics/datasource.go`](https://github.com/llm-d/llm-d-router/blob/620045bb19a95b088dc53f046c78a2d712b90a76/pkg/epp/framework/plugins/datalayer/source/metrics/datasource.go): parse through the filter, register the hook.
- [`source/http/datasource.go`](https://github.com/llm-d/llm-d-router/blob/620045bb19a95b088dc53f046c78a2d712b90a76/pkg/epp/framework/plugins/datalayer/source/http/datasource.go): `WithExtractorHook`, so a source can observe the extractors that bind to it.
- [`extractor/metrics/extractor.go`](https://github.com/llm-d/llm-d-router/blob/620045bb19a95b088dc53f046c78a2d712b90a76/pkg/epp/framework/plugins/datalayer/extractor/metrics/extractor.go): implement `MetricNames`.

No configuration change, no API change, no change to what extractors observe.

## Follow-ups, out of scope here

All three are separate issues; listing them so the boundary is explicit.

1. Skip parsing entirely when a scrape is byte-identical to the previous one. Most gauges do not change within 50ms.
2. Jitter the per-endpoint tickers. [`pkg/epp/datalayer/runtime.go:450`](https://github.com/llm-d/llm-d-router/blob/620045bb19a95b088dc53f046c78a2d712b90a76/pkg/epp/datalayer/runtime.go#L450) calls `NewTimeTicker`, which is a thin wrapper over a bare `time.NewTicker` ([`collector.go:50`](https://github.com/llm-d/llm-d-router/blob/620045bb19a95b088dc53f046c78a2d712b90a76/pkg/epp/datalayer/collector.go#L50)), and collectors are created together at startup, so scrapes arrive in phase.
3. Apply the same filter to the DCGM source. [`source/dcgm/datasource.go:100`](https://github.com/llm-d/llm-d-router/blob/620045bb19a95b088dc53f046c78a2d712b90a76/pkg/epp/framework/plugins/datalayer/source/dcgm/datasource.go#L100) parses the exporter's whole response into the same `PrometheusMetricMap`, and its extractor reads one family, `DCGM_FI_DEV_GPU_UTIL`. The mechanism proposed here drops in once the interface exists.

muchengl:
**What type of PR is this?**
/kind feature

This PR reduces model-server metrics parsing overhead in the EPP data layer.

`core-metrics-extractor` now declares the Prometheus metric families it consumes based on the existing engine metric mappings. `metrics-data-source` uses those declarations to filter model-server `/metrics` responses before passing them to the Prometheus text parser.

This avoids parsing metric families that EPP does not use for scheduling, such as extra runtime/process metrics exposed by vLLM.

The behavior is conservative: if no bound extractor declares metric families, the data source keeps the existing full-parse behavior.

Fixes #2441

```release-note
NONE

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/extractor/metrics/extractor.go
+++ b/pkg/epp/framework/plugins/datalayer/extractor/metrics/extractor.go
@@ -101,6 +101,27 @@ func (ext *Extractor) Produces() map[fwkplugin.DataKey]any {
 	return produced
 }
 
+var _ sourcemetrics.FamilyNamer = &Extractor{}
+
+// MetricNames declares every family this extractor can read, letting the
+// metrics data source discard the rest of a scrape before parsing it. The
+// engine of an endpoint is only known at extraction time, so the declaration
+// is the union over all registered mappings.
+func (ext *Extractor) MetricNames() []string {
+	seen := map[string]struct{}{}
+	var names []string
+	for _, mapping := range ext.registry.Mappings() {
+		for _, name := range mapping.MetricNames() {
+			if _, dup := seen[name]; dup {
+				continue
+			}
+			seen[name] = struct{}{}
+			names = append(names, name)
+		}
+	}
+	return names
+}
+
 // Extract transforms the typed metrics payload into endpoint attributes.
 func (ext *Extractor) Extract(ctx context.Context, in fwkdl.PollInput[sourcemetrics.PrometheusMetricMap]) error {
 	families := in.Payload
--- a/pkg/epp/framework/plugins/datalayer/source/metrics/filter.go
+++ b/pkg/epp/framework/plugins/datalayer/source/metrics/filter.go
@@ -1,0 +1,296 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"bufio"
+	"bytes"
+	"io"
+	"sync"
+	"sync/atomic"
+
+	fwkplugin "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+)
+
+// FamilyNamer is the optional contract an extractor implements to declare the
+// Prometheus metric families it reads. The metrics data source unions the
+// declarations of its bound extractors and discards every other family before
+// parsing. An extractor that does not implement it forces a full parse, so the
+// interface is safe to adopt incrementally.
+type FamilyNamer interface {
+	MetricNames() []string
+}
+
+// suffixes a scrape may append to a family name for histogram and summary series.
+var seriesSuffixes = [...]string{"_bucket", "_sum", "_count"}
+
+// familySelector holds the union of family names the bound extractors declared.
+// Extractors bind during configuration and the set is read on every scrape, so
+// the resolved set is published through an atomic pointer rather than a mutex:
+// scrapes never contend with each other.
+type familySelector struct {
+	mu    sync.Mutex
+	names map[string]struct{}
+	// resolved is nil until at least one extractor declares a family. A nil
+	// value means "keep everything", which is the behaviour of a data source
+	// whose extractors do not implement FamilyNamer.
+	resolved atomic.Pointer[map[string]struct{}]
+}
+
+// observe records the families ext declares, if it declares any.
+func (s *familySelector) observe(ext fwkplugin.Plugin) {
+	namer, ok := ext.(FamilyNamer)
+	if !ok {
+		return
+	}
+	names := namer.MetricNames()
+	if len(names) == 0 {
+		return
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.names == nil {
+		s.names = make(map[string]struct{}, len(names))
+	}
+	for _, name := range names {
+		if name != "" {
+			s.names[name] = struct{}{}
+		}
+	}
+	// Publish a copy so readers never observe a map being written.
+	published := make(map[string]struct{}, len(s.names))
+	for name := range s.names {
+		published[name] = struct{}{}
+	}
+	s.resolved.Store(&published)
+}
+
+// wanted returns the published set, or nil when no extractor declared anything.
+func (s *familySelector) wanted() map[string]struct{} {
+	if p := s.resolved.Load(); p != nil {
+		return *p
+	}
+	return nil
+}
+
+// bufferPool recycles the scratch buffers filterFamilies writes into. A scrape
+// keeps only the families an extractor reads, so the buffer stays small and
+// steady-state scrapes allocate nothing.
+var bufferPool = sync.Pool{
+	New: func() any { return new(bytes.Buffer) },
+}
+
+// lineReaderPool recycles the readers used to split a scrape into lines.
+var lineReaderPool = sync.Pool{
+	New: func() any { return bufio.NewReaderSize(nil, defaultReadBufferSize) },
+}
+
+// defaultReadBufferSize is the initial line-splitting buffer. Lines longer than
+// this are handled by accumulating fragments, so the value bounds memory rather
+// than the accepted line length.
+const defaultReadBufferSize = 16 << 10
+
+// maxFamilyNameSize preallocates the buffer holding the family name in force.
+// Longer names still work; the buffer grows.
+const maxFamilyNameSize = 128
+
+// filterFamilies copies the text-exposition lines belonging to want from src
+// into dst, dropping everything else.
+//
+// A scrape groups a family's "# HELP"/"# TYPE" headers immediately before its
+// samples, so the decision made for a header usually carries to the samples
+// that follow it. Carrying it blindly would be wrong, because headers are
+// optional and a scrape may start a new family with a bare sample line, so a
+// sample is only granted the standing decision when its own name belongs to
+// the family that header named. Any other sample is matched on its own name.
+func filterFamilies(dst *bytes.Buffer, src io.Reader, want map[string]struct{}) error {
+	reader, _ := lineReaderPool.Get().(*bufio.Reader)
+	reader.Reset(src)
+	defer func() {
+		reader.Reset(nil)
+		lineReaderPool.Put(reader)
+	}()
+
+	var state filterState
+	state.family = make([]byte, 0, maxFamilyNameSize)
+	for {
+		line, err := readLine(reader)
+		if len(line) > 0 && state.keepLine(line, want) {
+			dst.Write(line)
+		}
+		if err != nil {
+			if err == io.EOF {
+				return nil
+			}
+			return err
+		}
+	}
+}
+
+// filterState carries the decision made for the family a header named, so the
+// samples that follow it are admitted without repeating the lookup.
+type filterState struct {
+	// family is the name from the most recent header, empty when no header is
+	// in force. It is copied because the line it came from is reused.
+	family []byte
+	// keep is the decision made for family.
+	keep bool
+}
+
+// keepLine reports whether line survives filtering, updating the state.
+func (s *filterState) keepLine(line []byte, want map[string]struct{}) bool {
+	trimmed := trimLeadingSpace(line)
+	if isBlank(trimmed) {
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: pkg/epp/framework/plugins/datalayer/extractor/metrics/extractor.go [hunk 1]
```diff
@@ -101,6 +101,27 @@ func (ext *Extractor) Produces() map[fwkplugin.DataKey]any {
 	return produced
 }
 
+var _ sourcemetrics.FamilyNamer = &Extractor{}
```
muchengl, line 104: https://github.com/llm-d/llm-d-router/issues/2441 
> Let extractors declare the metric families they read, then filter each scrape to the union of those declarations before parsing.

pkg/epp/framework/plugins/datalayer/source/metrics/filter.go
```diff
@@ -0,0 +1,296 @@
...
+// filterFamilies copies the text-exposition lines belonging to want from src
+// into dst, dropping everything else.
+//
+// A scrape groups a family's "# HELP"/"# TYPE" headers immediately before its
+// samples, so the decision made for a header usually carries to the samples
+// that follow it. Carrying it blindly would be wrong, because headers are
+// optional and a scrape may start a new family with a bare sample line, so a
+// sample is only granted the standing decision when its own name belongs to
+// the family that header named. Any other sample is matched on its own name.
+func filterFamilies(dst *bytes.Buffer, src io.Reader, want map[string]struct{}) error {
```
elevran, line 121: This reimplements a Prometheus text-format line scanner from scratch. 
Unfortunately, `expfmt.TextParser` (used below) has no filtering hook, so there's no drop-in standard-package swap - but `github.com/prometheus/prometheus/model/textparse` is the tokenizer Prometheus's own scrape loop uses. It's spec-compliant by construction. It would remove this hand-written scanner entirely, at the cost of a much heavier transitive dependency (pulls in a large slice of `prometheus/prometheus`). Flagging this as a design question rather than a request to change it here, since dependency changes need discussion first.
Would you mind looking into that or for an option to reuse a subset of that package?

I think this is a useful change, just have concerns about handrolled scanner compliance over time.

pkg/epp/framework/plugins/datalayer/source/metrics/filter.go
```diff
@@ -0,0 +1,296 @@
...
+		switch line[i] {
+		case '{', ' ', '\t', '\n', '\r':
+			return line[:i]
+		}
+	}
+	return line
+}
+
+// matches reports whether name, or the family it is a series of, is wanted.
+func matches(name []byte, want map[string]struct{}) bool {
```
elevran, line 240: Structural assumption might casuse issues.
Stripping `_bucket`/`_sum`/`_count` and checking the base name against `want` means a standalone metric literally named e.g. `x_sum` gets treated as part of a wanted histogram/summary family `x`, even though it has no relation to it - the declared TYPE is only consulted for the family's own header, never to disambiguate this suffix collision. A tokenizer that tracks each family's declared type (like `model/textparse`) doesn't have this ambiguity.

pkg/epp/framework/plugins/datalayer/extractor/metrics/extractor.go
```diff
@@ -101,6 +101,27 @@ func (ext *Extractor) Produces() map[fwkplugin.DataKey]any {
...
+
+// MetricNames declares every family this extractor can read, letting the
+// metrics data source discard the rest of a scrape before parsing it. The
+// engine of an endpoint is only known at extraction time, so the declaration
+// is the union over all registered mappings.
+func (ext *Extractor) MetricNames() []string {
```
elevran, line 110: nit: no dedicated test for this method - the union/dedup logic across `registry.Mappings()` isn't exercised directly, only transitively via other extractor tests.

pkg/epp/framework/plugins/datalayer/source/metrics/filter.go
```diff
@@ -0,0 +1,296 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
elevran, line 2: These files are created in llm-d, not k8s, so missing "Copyright 2026 The llm-d Authors." line. Also other new files in this PR.
See #2450 for context.

elevran (commented):
Nice perf win and good test coverage on the filter itself (long lines, brace-in-label-value, name-prefix collisions all covered). Noting this is a draft per the author's own flag - posting for early feedback rather than a merge gate.

Main open question: `filter.go` hand-rolls a Prometheus text-exposition scanner (HELP/TYPE parsing, family grouping, comment/blank-line handling). `expfmt.TextParser` (already a dependency) can't filter before parsing, so there is no drop-in standard-package replacement, but `github.com/prometheus/prometheus/model/textparse` is the actual spec-compliant tokenizer Prometheus's own scrape loop uses for this - see the inline comment on `filterFamilies` for tradeoffs. Worth a discussion before this lands, since it would be a new (and heavier) dependency.

pkg/epp/framework/plugins/datalayer/source/metrics/filter.go
muchengl: I agree. A hand-rolled scanner may not be ideal long term. I’ll look into textparse and whether we can reuse a smaller subset.

ahg-g:
@muchengl this is marked as a draft, is this indeed the case?

muchengl:
> @muchengl this is marked as a draft, is this indeed the case?

Hi @ahg-g , Thanks for bringing this up!

As @elevran pointed out, the approach in this PR is still not ideal and feels somewhat hacky.

The issue itself is valid (#2441), and filtering unused Prometheus metrics would clearly save resources. Based on my research, ideally, SGLang and vLLM should first fix their Prometheus metric query/filtering support, and llm-d could then query only the metrics it needs.

However, the related upstream issues/PRs have not been accepted yet, so there is currently no clean way to do this filtering.

I've opened PRs for this in both upstream projects:
SGLang: https://github.com/sgl-project/sglang/pull/35768
vLLM: https://github.com/vllm-project/vllm/pull/53160

The fixes on the vLLM and SGLang sides are both just a few lines of code. If you think this issue is worth addressing, I can prioritize pushing those upstream changes forward.

PR #2444, opened 2026-08-18
feat(epp): record a termination cause on end-of-stream response records

### LINKED ISSUE #2429: Record a termination cause on the end-of-stream response record

A `TerminationCause` string enum in `framework/interface/requestcontrol`, set on every end-of-stream `Response` record and empty on earlier invocations:

- `natural`: the model server ended the stream. Used when no other cause was recorded. Includes error responses delivered in full.
- `client-disconnect`: the ext_proc stream was torn down while the response was in flight. A client disconnect is the most common reason. Any teardown the EPP observes only as a cancelled stream (Envoy drain or restart, an upstream reset) produces the same value, because the EPP cannot distinguish the reasons from its side of the stream.
- `evicted`: the EPP terminated the request to reclaim capacity.
- `error`: the stream ended without completing for any other reason.

`StreamingServer` classifies the cause because both inputs are in scope in its forced-completion defer: `ctx.Err()` and `RequestState`. The director runs its final pass on a fresh context and cannot observe the cancellation. At end-of-stream, the director defaults an unset cause to `natural` and carries the cause on every end-of-stream record, including the one delivered to the eviction tracker.

An evicted request classifies as `evicted` even though its teardown also cancels the context; the eviction is the more specific cause.

The boundary between `client-disconnect` and `error` is approximate. A teardown that reaches `Recv` as a half-close before the context cancellation propagates classifies as `error`. Requests whose response processing is skipped never observe completion, so their records always carry a non-natural cause. Unit tests pin the mapping.

The field is additive. No existing plugin reads it, no behavior changes, and the enum has no config surface.

Exporting the cause as a metric is out of scope. The field lives on the in-process record consumed by plugins; whether it should also appear as a label on an exported counter is a separate question.

**Why is this needed**:

`ResponseBody` plugins receive a `requestcontrol.Response` on every chunk, with `EndOfStream` set on the last one. The record carries no field that says whether the model server finished the stream or the stream was cut short.

The EPP can tell the outcomes apart at the moment the stream ends. `StreamingServer` runs a deferred forced-completion pass when the ext_proc stream tears down before the response completes, and `RequestContext.RequestState` holds `RequestEvicted` when the EPP terminated the request to reclaim capacity. Neither signal reaches the record.

A stream that never completes carries no `Usage`, so its record reports zero output tokens. A partial token count, where one exists, understates how long the request would have run.

- Long requests are the most likely to be truncated, so fitting a per-request footprint estimator on unmarked truncated observations biases its estimates downward.
- Operators cannot attribute stream terminations, because a client disconnect, an EPP eviction, and a model server error all produce the same record.

A termination signal would also serve request fail-over and migration (#2345), which needs to know whether a stream ended because the request completed.

LukeAVanDrie:
**What type of PR is this?**

/kind feature

`ResponseBody` plugins receive a `requestcontrol.Response` on every chunk, with `EndOfStream` set on the last one. The record does not say whether the model server ended the stream or the stream was cut short, so a consumer that learns from response records cannot exclude truncated observations, and an operator cannot attribute stream terminations.

Adds a `TerminationCause` string enum to `framework/interface/requestcontrol`, set on every end-of-stream `Response` and empty on earlier invocations:

- `natural`: the model server ended the stream. Includes error responses delivered in full.
- `client-disconnect`: the ext_proc stream was torn down while the response was in flight. A client disconnect is the most common reason. Any teardown the EPP observes only as a cancelled stream (Envoy drain or restart, an upstream reset) produces the same value, because the EPP cannot distinguish the reasons from its side of the stream.
- `evicted`: the EPP terminated the request to reclaim capacity.
- `error`: the stream ended without completing for any reason not otherwise classified.

`StreamingServer` classifies the cause in its forced-completion defer, where `ctx.Err()` and `RequestState` are both in scope. The director resolves the cause once per end-of-stream pass, defaults an unset cause to `natural`, and carries it on every end-of-stream record, including the one delivered to the eviction tracker.

The boundary between `client-disconnect` and `error` is approximate. A teardown that reaches `Recv` as a half-close before the context cancellation propagates classifies as `error`. Requests whose response processing is skipped never observe completion, so their records always carry a non-natural cause. Unit tests pin the mapping.

Known limitation, to be tracked in a followup issue: a gRPC upstream abort surfaces as response trailers with a non-OK `grpc-status`, which the trailers path does not inspect, so the truncated stream classifies as `natural`.

The field is additive. No existing plugin reads it, and the enum has no config surface. Exporting the cause as a metric label is out of scope.

**Test plan**:

- [x] Classifier maps eviction, cancelled context, skipped response processing, and the remaining teardowns to the documented causes
- [x] End-of-stream records default an unset cause to `natural`; a recorded cause survives to the record
- [x] Mid-stream records carry no cause

Fixes #2429

```release-note
Response records delivered to ResponseBody plugins carry a TerminationCause on the end-of-stream invocation, distinguishing natural completion from client disconnects, flow-control evictions, and errors.
```

### DIFF
--- a/pkg/epp/framework/interface/requestcontrol/plugins.go
+++ b/pkg/epp/framework/interface/requestcontrol/plugins.go
@@ -80,10 +80,12 @@ type ResponseHeaderProcessor interface {
 //   - For non-streaming: Invoked once with response.EndOfStream set to true.
 //   - Plugins must treat the call where response.EndOfStream == true as the final lifecycle hook
 //     to perform cleanup or final logging.
+//   - The end-of-stream call carries response.TerminationCause; earlier calls leave it empty.
 //
 // TODO(https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2079):
-// Update signature to pass error/termination state. This is a breaking change required for plugins to distinguish
-// between success, errors, and disconnects.
+// Upstream proposes passing error/termination state through the signature. Response.TerminationCause
+// carries the termination state without a signature change; it does not distinguish success from
+// an error response.
 type ResponseBodyProcessor interface {
 	plugin.Plugin
 	ResponseBody(ctx context.Context, request *fwksched.InferenceRequest, response *Response, targetEndpoint *datalayer.EndpointMetadata)
--- a/pkg/epp/framework/interface/requestcontrol/types.go
+++ b/pkg/epp/framework/interface/requestcontrol/types.go
@@ -22,6 +22,31 @@ import (
 	requesthandling "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requesthandling"
 )
 
+// TerminationCause labels how a dispatched request's response stream ended. Observers that learn
+// from response records need it to tell a complete observation from a truncated one: a stream cut
+// short by a disconnect or an eviction reports only the tokens emitted before the cut. Requests
+// whose response processing is skipped never observe completion, so their records always carry a
+// non-natural cause.
+type TerminationCause string
+
+const (
+	// TerminationCauseNatural is a stream the model server ended on its own, including responses
+	// that deliver an error status in full.
+	TerminationCauseNatural TerminationCause = "natural"
+	// TerminationCauseClientDisconnect is a stream torn down under the EPP while the response was
+	// in flight. A client disconnect is the most common reason; any teardown the EPP observes only
+	// as a cancelled stream (Envoy drain or restart, an upstream reset, the EPP's own drain)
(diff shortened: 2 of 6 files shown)


elevran (commented):
This PR is additive (new field, no signature break), the classification logic in terminationCause() is correct for every path I traced (normal completion, eviction, disconnect, skipped-response-processing), and tests exercise all four causes plus the "mid-stream carries no cause" invariant. The known limitation (gRPC trailers not inspected) is called out explicitly in the PR body as a documented gap for a follow-up issue

elevran (approved):

PR #2445, opened 2026-08-18
Honor podLabelSelector for KV-event subscriptions

vMaroon:
**What type of PR is this?**

/kind bug

Apply `kvEventsConfig.podDiscoveryConfig.podLabelSelector` to endpoints delivered by `endpoint-notification-source`. Pool members that publish no KV events, such as decode-only pods in a P/D pool, otherwise get dialed and retried indefinitely. On a 32-prefill/12-decode GLM-5.2 pool, each EPP replica logged about 2.5 subscriber connection-refused errors per second (48k in six hours), burying real errors.

Parse the selector at startup when pod discovery is enabled and subscribe only to matching endpoints. The default selector is empty, so omitted or empty selectors match every endpoint; an invalid explicit selector fails startup. When an endpoint stops matching, remove its subscriber and cached index entries. Excluded add/update events clear cached entries only when a subscriber was removed. Deletion performs cleanup even without a subscriber, since speculative entries may exist.

The original in-cluster validation of the selector filter reduced subscriber connection errors to zero on that pool, with index admissions and lookups unchanged. The lifecycle cleanup changes were verified locally.

None; localized bug fix.

**Test plan**:

- [x] Configuration controls subscriptions for equality, set-based, and empty selectors; invalid selectors fail startup only when discovery is enabled.
- [x] Label changes and deletions remove subscribers and cached entries; endpoints can subscribe again when their labels match.
- [x] Omitted parameters and nested discovery configs preserve subscriptions for unlabeled endpoints.
- [x] Excluded add/update events skip index cleanup when no subscriber exists; deletion still clears cached entries.

**Release note**:

```release-note
With `kvEventsConfig.discoverPods: true`, `precise-prefix-cache-producer` applies `kvEventsConfig.podDiscoveryConfig.podLabelSelector` to endpoint notifications. Only matching endpoints receive KV-event subscriptions. The selector defaults to empty: omitted or empty selectors include all endpoints, preserving subscriptions for existing configs that leave it unset. Invalid selectors fail EPP startup when pod discovery is enabled. Endpoints that stop matching have their subscriptions and cached index entries removed.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
@@ -29,12 +29,9 @@ import (
 
 var _ fwkdl.EndpointExtractor = &Producer{}
 
-// Extract processes endpoint lifecycle events emitted by the
-// endpoint-notification-source: add/update installs a per-pod ZMQ KV-events
-// subscriber, delete tears one down. No-op unless per-pod discovery is
-// enabled, and endpoints outside PodDiscoveryConfig.PodLabelSelector are
-// ignored so that pool members that publish no KV events (e.g. decode-only
-// pods) are never dialed.
+// Extract subscribes matching endpoints to per-pod KV events when discovery is
+// enabled. Deleted endpoints and endpoints that stop matching lose their
+// subscriber and cached index entries.
 func (p *Producer) Extract(ctx context.Context, event fwkdl.EndpointEvent) error {
 	if !p.kvEventsConfig.DiscoverPods || p.kvEventsConfig.PodDiscoveryConfig == nil {
 		return nil
@@ -47,19 +44,14 @@ func (p *Producer) Extract(ctx context.Context, event fwkdl.EndpointEvent) error
 	logger := log.FromContext(ctx).WithName(p.typedName.String())
 	endpointKey := meta.ID.String()
 
-	if p.podSelector != nil && !p.podSelector.Matches(labels.Set(meta.Labels)) {
-		logger.V(logging.TRACE).Info("Skipping endpoint outside podLabelSelector",
-			"endpoint", endpointKey, "selector", p.podSelector.String())
-		return nil
-	}
-
-	switch event.Type {
-	case fwkdl.EventAddOrUpdate:
+	matchesSelector := p.podSelector == nil || p.podSelector.Matches(labels.Set(meta.Labels))
+	switch {
+	case event.Type == fwkdl.EventAddOrUpdate && matchesSelector:
 		if err := p.ensureSubscriber(ctx, meta); err != nil {
 			return err
 		}
 		logger.V(logging.DEBUG).Info("Adding subscriber", "endpoint", endpointKey)
-	case fwkdl.EventDelete:
+	case event.Type == fwkdl.EventAddOrUpdate || event.Type == fwkdl.EventDelete:
 		p.subscribersManager.RemoveSubscriber(ctx, endpointKey)
 		if meta.Address != "" {
 			if err := p.kvCacheIndexer.KVBlockIndex().Clear(ctx, fmt.Sprintf("%s:%s", meta.Address, meta.Port)); err != nil {
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/llm-d/llm-d-router/pkg/common/observability/logging"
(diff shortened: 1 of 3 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -158,6 +160,16 @@ func PluginFactory(name string, rawParameters *json.Decoder, handle plugin.Handl
 // The kvcache indexer, KV-events pool, and any local ZMQ subscriber start
 // in background goroutines bound to ctx.
 func New(ctx context.Context, name string, config PluginConfig) (*Producer, error) {
+	var podSelector labels.Selector
+	if kc := config.KVEventsConfig; kc != nil && kc.DiscoverPods && kc.PodDiscoveryConfig != nil && kc.PodDiscoveryConfig.PodLabelSelector != "" {
```
bongwoobak, line 164: Nothing on main actually reads `podLabelSelector` today, this PR is the first consumer, and the default in `kvevents.DefaultConfig()` is `llm-d.ai/inference-serving=true`, not empty. `PluginFactory` starts from that default and merges the user's JSON on top, so anyone who never set the field starts filtering on upgrade and pods without that label quietly lose their subscriptions. The test sets the field explicitly in every case so the default path isn't covered.

Can the default become `""` here? Or at least a release note line saying what happens to configs that leave it unset.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
```diff
@@ -44,13 +44,14 @@ func (p *Producer) Extract(ctx context.Context, event fwkdl.EndpointEvent) error
...
+	case event.Type == fwkdl.EventAddOrUpdate && matchesSelector:
 		if err := p.ensureSubscriber(ctx, meta); err != nil {
 			return err
 		}
 		logger.V(logging.DEBUG).Info("Adding subscriber", "endpoint", endpointKey)
-	case fwkdl.EventDelete:
+	case event.Type == fwkdl.EventAddOrUpdate || event.Type == fwkdl.EventDelete:
 		p.subscribersManager.RemoveSubscriber(ctx, endpointKey)
 		if meta.Address != "" {
 			if err := p.kvCacheIndexer.KVBlockIndex().Clear(ctx, fmt.Sprintf("%s:%s", meta.Address, meta.Port)); err != nil {
```
bongwoobak, line 57: This also runs on every update of an endpoint the selector excludes, and on Redis `Clear` is a full keyspace SCAN. Those endpoints never had a subscriber, so there is nothing to clear there. If `RemoveSubscriber` returned whether it actually removed something, the clear could be gated on that and the excluded pods stay free.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
elevran: +1
Should enable backward compatibility or at least have a release note to indicate the new behavior.

elevran (commented):
The test additions cover both the startup validation and the lifecycle-cleanup edge cases well.
Backward compat should be considered per @bongwoobak feedback.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
vMaroon: Changed the default selector to `""` so configs that omit the field keep subscribing to all discovered endpoints. Added factory-path regression coverage for omitted parameters and nested configs, while explicit selectors still filter.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
vMaroon: `RemoveSubscriber` now reports whether it removed a subscriber. Excluded add/update events skip `Clear` when it returns false. Deletions still clear entries because speculative indexing can populate them without a subscriber. The regression test verifies repeated excluded updates cause no extra clears.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
```diff
@@ -44,13 +44,14 @@ func (p *Producer) Extract(ctx context.Context, event fwkdl.EndpointEvent) error
...
 
-	switch event.Type {
-	case fwkdl.EventAddOrUpdate:
+	matchesSelector := p.podSelector == nil || p.podSelector.Matches(labels.Set(meta.Labels))
+	switch {
+	case event.Type == fwkdl.EventAddOrUpdate && matchesSelector:
 		if err := p.ensureSubscriber(ctx, meta); err != nil {
 			return err
 		}
 		logger.V(logging.DEBUG).Info("Adding subscriber", "endpoint", endpointKey)
```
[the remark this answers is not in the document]
vMaroon: Removed the unconditional log. `ensureSubscriber` already logs after ensuring a subscriber and returns before that log when the address is missing.

elevran (approved):
Verified the backward-compat fix - thanks.
All three review threads look addressed - approving.

Issue #1724, opened 2026-06-23
Improve InFlightLoad Token Accounting During Response Streaming


Mohamedma96:
### What?

/kind feature

### why we need it

This issue discusses and improves the accuracy of endpoint token load accounting used by the EPP scheduler.

Today, InFlightLoad.Tokens is incremented when a request is assigned to an endpoint, but the token contribution is only released at a lifecycle boundary (start-of-stream or end-of-stream depending on configuration). As a result, an endpoint may continue to appear heavily loaded even after it has already generated a significant portion of its response.

This behavior affects scheduling decisions that rely on token-based load measurements.

This solution introduces incremental token draw-down during response streaming so that the tracked token load more accurately reflects the remaining work being processed by an endpoint.

Instead of:

```
+1000 tokens at request start
...
...
...
-1000 tokens at completion
```

the token load will be gradually reduced as response chunks are produced:

```
+1000
-100
-150
-200
...
```

This allows endpoint load information to converge toward actual runtime utilization throughout the generation lifecycle.

### Relationship to Existing Functionality

This work is a natural extension of the token load-aware scheduling introduced in #655.

Issue #655 added the token-load-scorer, enabling routing decisions based on InFlightLoad.Tokens.

This proposal improves the quality of the underlying load signal consumed by that scorer by ensuring token accounting reflects ongoing generation progress rather than only request start and completion events.

As a result, the scheduler, token-load scorer, prefix-cache-affinity filter, and saturation detection mechanisms can make decisions based on more realistic endpoint utilization.

### Current Behavior
Scheduling
Request arrives.
Scheduler reads InFlightLoad.Tokens.
Endpoint is selected.
PreRequest() increments:
requestTracker
tokenTracker
Response Generation

During generation, token load remains unchanged.

Example:

Request estimate: 1000 tokens

Pod load:
```
1000
1000
1000
1000
1000
0

```
The scheduler continues to observe the full token contribution until the request lifecycle completes.

### Proposed Behavior

Introduce incremental token release during response streaming.

Example:

Request estimate: 1000 tokens

Pod load:
```
1000
850
700
500
300
100
0
```

This allows future scheduling cycles to observe a more realistic representation of endpoint load.

### Expected Benefits
More accurate token-based scheduling decisions.
Improved effectiveness of the token-load scorer introduced in #655.
Better endpoint utilization balancing.
More responsive saturation detection.
More accurate inputs for token-based filters and scorers.
[...]

PR #2446, opened 2026-08-18
feat(epp): expose a streamed-event count on the response record

### LINKED ISSUE #2430: Expose a streamed-token count on the response record

A `StreamedEvents` count on the response record: the number of SSE data events observed while parsing a streamed response, carried as a progress signal that correlates with output length and survives stream truncation.

- `requesthandling.ParsedResponse` gains `StreamedEvents int`, the events observed in one parse call.
- The OpenAI parser sets it, excluding the `[DONE]` terminator. Non-streamed responses return zero.
- `handlers.RequestContext` accumulates it across chunks.
- `requestcontrol.Response` carries the running total on every ResponseStreaming plugin invocation, including the deferred end-of-stream invocation on the disconnect path.

An event split at a chunk boundary is counted with the half carrying the `data: ` prefix. A boundary landing inside the prefix drops that event and one landing inside the terminator counts it; each is a one-event error, within the accuracy stated below.

**Scope across parsers.** The OpenAI parser sets the count, and the vLLM HTTP parser inherits it by delegation. The Vertex AI parser also delegates, but its streamed responses arrive as gRPC-framed JSON bodies rather than SSE, so it reports zero, as do the Anthropic parser (separate SSE handling) and the vLLM gRPC parser (chunked gRPC); all three can be extended in follow-up work. Because unsupported parsers report zero, zero is not evidence that nothing was generated, and consumers must not use the count to lower an estimate.

**Accuracy.** `StreamedEvents` counts protocol frames, not tokens. For bare completions-API streams that emit one content delta per token, the count tracks the token count closely. Chat streams add bookend events (role delta, finish reason, optional usage-only final event), and the Responses API interleaves lifecycle events (`response.created`, `output_item.added`, per-item events), so the count runs high by a small per-request constant; a server that batches tokens into one event makes it run low. Counting only content deltas would require decoding every event's JSON on the response hot path to remove a constant-sized bias. The counting as implemented is a prefix scan over lines the parser already splits and adds no per-chunk allocations. If #2423 moves the streaming parsers to operate on `[]byte`, the counter moves with them.

**Non-goals.** Replacing `usage`. When a stream completes with a usage block, `usage` is authoritative. `StreamedEvents` covers the streams where `usage` never arrives. Consumer wiring is also out of scope: this issue lands the signal only, and the plugins that should read it (below) are follow-up work.

**Why is this needed**:

Output-token counts reach the EPP only through the `usage` block. For streamed responses, OpenAI-compatible servers emit the block on the final event before `[DONE]`, and only when the client set `stream_options.include_usage`. A stream that ends early never carries the block. A stream whose client did not opt in never carries it either.

Client disconnects follow a separate code path. When the ext_proc stream aborts, a deferred cleanup in `pkg/epp/handlers/server.go` invokes the director's end-of-stream hook directly and the response parser never runs. Usage cannot arrive on that path. A length signal for an aborted stream has to be accumulated on the request context while the stream is flowing.

The token metrics (`RecordOutputTokens` and the related recorders) are gated on a parsed usage block and never fire for a truncated stream, and every ResponseStreaming plugin sees a zero `Response.Usage`.

The requests whose output length matters most for capacity and fairness accounting are long requests that were evicted or abandoned, and those produce no usage block. In-tree consumers that can pick the signal up in follow-up work:

- The program-aware fairness plugin (LAS strategy) computes attained service from `Response.Usage`. A client that disconnects before the final event is charged zero cost.
- `requestattributereporter` skips zero values when writing to dynamic metadata, so a truncated stream leaves no terminal record in access logs.
- #2305 (approximating output token length) needs observed output lengths as ground truth, and #1724 (incremental InFlightLoad drawdown during streaming) needs a per-chunk generation signal.

The event count is a monotone progress signal that survives truncation because it accumulates while the stream flows. An evicted request that emitted 900 events was a long request even though its exact token count is unknown. The predicted-latency plugin already keeps a private chunk counter (`generatedTokenCount`) for its TPOT estimate; a shared field exposes a finer-grained version of the same signal to other plugins.

#2429 proposes a termination cause on the same end-of-stream record. The two fields pair naturally: the cause marks a record as truncated, and `StreamedEvents` bounds how much the truncated request generated.

Related: #2429 (termination cause on the response record), #2305 (output-length approximation), #1724 (incremental token accounting during streaming), #2423 (byte-based streaming parsers; the counter follows if that refactor lands first).

LukeAVanDrie:
**What type of PR is this?**

/kind feature

Adds `StreamedEvents`, a count of SSE data events observed while parsing a streamed response:

- `requesthandling.ParsedResponse.StreamedEvents`: events observed in one parse call, set by the OpenAI parser (`[DONE]` and CRLF terminators excluded). The vLLM HTTP parser inherits it by delegation. The Vertex AI parser also delegates, but its streamed responses arrive as gRPC-framed JSON bodies rather than SSE, so it reports zero, as do the Anthropic and vLLM gRPC parsers; all three can be extended in follow-up work.
- `handlers.RequestContext.StreamedEvents`: accumulated across chunks.
- `requestcontrol.Response.StreamedEvents`: the running total, on every ResponseStreaming invocation including the deferred end-of-stream dispatch after a client disconnect.

Streams that truncate, or whose clients never set `stream_options.include_usage`, deliver no usage block, so the response record carries no length signal for the requests whose output length matters most for capacity and fairness accounting. The event count accumulates while the stream flows, so it survives truncation. It counts protocol frames, not tokens; the field contract documents the accuracy bounds, and zero is not evidence of zero output (unsupported parsers report zero).

Scope is the signal only. Consumers (LAS attained-service fallback, `requestattributereporter`, #2305, #1724) are follow-up work per the issue.

Fixes #2430

**Release note**:

```release-note
The response record passed to ResponseStreaming plugins carries `StreamedEvents`, a running count of SSE data events that survives stream truncation and client disconnects, giving plugins a length signal for streams that never deliver a usage block.
```

**Test plan** (functionality verified by new tests; all ran and passed locally):

- [x] Per-chunk counting in the OpenAI parser, including terminator and CRLF-terminator exclusion and chunk-boundary cuts (after the prefix: counted once; inside the prefix: dropped; inside the terminator: counted).
- [x] Accumulation across chunks on the request context, including an event cut at a chunk boundary counting once and a truncated stream keeping the count it reached.
- [x] Each ResponseStreaming invocation carries the accumulator value at dispatch time, through the async chunk queue and the synchronous end-of-stream dispatch.
- [x] `make presubmit` green locally.

### DIFF
--- a/pkg/epp/framework/interface/requestcontrol/types.go
+++ b/pkg/epp/framework/interface/requestcontrol/types.go
@@ -38,6 +38,12 @@ type Response struct {
 	ReqMetadata map[string]any
 	// Token usage counts parsed from the response body.
 	Usage requesthandling.Usage
+	// StreamedEvents is the running count of stream data events observed so far. It is the only
+	// length signal this record carries for a truncated stream, since a stream that never
+	// completes carries no usage block. Consumers must not treat zero as evidence that nothing
+	// was generated; requesthandling.ParsedResponse documents which parsers count and how the
+	// count deviates from the token count.
+	StreamedEvents int
 	// DynamicMetadata is a map of metadata that can be passed to the Envoy. It is populated into the dynamic
 	// metadata when processing ProcessingResponse_RequestHeaders.
 	DynamicMetadata *structpb.Struct
--- a/pkg/epp/framework/interface/requesthandling/plugins.go
+++ b/pkg/epp/framework/interface/requesthandling/plugins.go
@@ -83,4 +83,12 @@ type ParseResult struct {
 type ParsedResponse struct {
 	// Usage is only populate when the raw response has usage.
 	Usage *Usage
+	// StreamedEvents is how many stream data events this parse observed, zero for a non-streamed
+	// response or a parser that does not count. For servers that stream one content delta per
+	// token the count approximates the tokens in the parsed chunk, but protocol events (role
+	// deltas, finish reasons, usage-only events, Responses API lifecycle events) inflate it, a
+	// server that batches tokens into one event deflates it, and the scan recognizes only the
+	// "data: " framing this parser already assumes (no-space prefixes, multi-line data fields,
+	// and keep-alive data lines skew it further).
+	StreamedEvents int
 }
--- a/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
@@ -203,11 +203,17 @@ func countStreamEvents(chunk []byte) int {
 	count := 0
 	for line := range bytes.SplitSeq(chunk, []byte("\n")) {
 		content, ok := bytes.CutPrefix(line, []byte(streamingRespPrefix))
-		if ok && !bytes.Equal(bytes.TrimSuffix(content, []byte("\r")), []byte("[DONE]")) {
+		if ok && !isStreamTerminator(content) {
 			count++
 		}
 	}
 	return count
+}
+
+// isStreamTerminator reports whether an SSE data payload is the [DONE] terminator, tolerating a
+// trailing \r left by CRLF line splitting.
+func isStreamTerminator(content []byte) bool {
+	return bytes.Equal(bytes.TrimSuffix(content, []byte("\r")), []byte("[DONE]"))
 }
 
 // determineAPITypeFromPath determines the API type based on the request path.
@@ -401,7 +407,7 @@ func extractUsageStreaming(responseBytes []byte) *fwkrh.Usage {
 			continue
 		}
 		// When the stream is terminated with [DONE] or there's not any usage data, skip the line
-		if bytes.Equal(content, []byte("[DONE]")) || !bytes.Contains(content, []byte("usage")) {
+		if isStreamTerminator(content) || !bytes.Contains(content, []byte("usage")) {
 			continue
 		}
 		if err := json.Unmarshal(content, &streamResponse); err != nil {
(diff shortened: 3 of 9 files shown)


LukeAVanDrie:
A few notes on deliberate non-changes:

- The evictor's `Response` literal in `director.go` stays minimal: the eviction tracker consumes only `RequestID` and `EndOfStream`, and padding it has no consumer until #1724.
- The pre-existing "Malformed JSON in stream (skipped)" fixture in `openai_test.go` uses a backquoted `\n`, so it is one line rather than the two its name implies; left untouched here, worth a separate test fix.
- If #2425 (byte-based `extractUsageStreaming`) lands first, `countStreamEvents` converts to `bytes.SplitSeq`/`bytes.CutPrefix` in the rebase; either order costs a few lines.

elevran (dismissed):
`extractUsageStreaming` (existing code, untouched by this PR) checks content == "[DONE]" without trimming a trailing `\r`, while the new `countStreamEvents` trims `\r` before comparing.
Harmless today since a `[DONE]\r` line also fails the usage-substring check in `extractUsageStreaming` and gets skipped either way, but the two functions now diverge in how they detect the terminator.
Mentioning for your awareness only, not proposing to block on it.

I found nothing else worth commenting - this is clean and have no concerns.
Approving and NOT merging in case you'd like to align termination checks in this PR.
Also - please fix conflict

LukeAVanDrie:
> `extractUsageStreaming` (existing code, untouched by this PR) checks content == "[DONE]" without trimming a trailing `\r`, while the new `countStreamEvents` trims `\r` before comparing. Harmless today since a `[DONE]\r` line also fails the usage-substring check in `extractUsageStreaming` and gets skipped either way, but the two functions now diverge in how they detect the terminator. Mentioning for your awareness only, not proposing to block on it.

Fixed in [1c1ca2d1d65fbcb35a3c37fd0b610a0f0af554fd](https://github.com/llm-d/llm-d-router/pull/2446/changes/1c1ca2d1d65fbcb35a3c37fd0b610a0f0af554fd). Good catch!

elevran (approved):

PR #2451, opened 2026-08-19
fix(sidecar): await prefill before dispatching PD Multi Tier decode leg

### LINKED ISSUE #2438: sidecar: PD Multi Tier decode leg is dispatched concurrently with prefill, charging producer latency to the connector's load deadline
### What happened

`handleP2P` dispatches the prefill and decode legs **concurrently**. For the
PD Multi Tier path this means the decoder's `FetchMsg` reaches the
prefiller before the prefiller has stored any KV for that `kv_request_id`. The
prefiller parks the demand awaiting supply, and the decoder spends its fixed
`_LOAD_TIMEOUT_S` (30 s, hard-coded in vLLM's
`kv_offload/tiering/p2p/session/client.py`) waiting for KV that has not been
produced yet. On expiry it aborts the load and, under
`kv_load_failure_policy: recompute`, recomputes the prompt locally — the work
disaggregation exists to avoid.

The current comment on `handleP2P` states the ordering is safe because the
connector parks blocks stored before the fetch binds the session. Parking is
indeed safe for correctness, but it silently charges the producer's queue and
compute time against the consumer's load deadline.

### Why this differs from the NIXL path

The default `nixlv2` path is sequential — it forwards to the prefiller and waits
for it to return before dispatching decode. Concurrent dispatch there is an
opt-in gated on `MoRIIOParallelDispatch && MoRIIOWriteMode`, and even then it
only defers *committing* decode's response. The PD Multi Tier path is the outlier.

### Evidence

Measured on 8 prefill + 8 decode (gpt-oss-120b, TP=1, vLLM v0.27.1, 16x H200),
instrumenting the vLLM Multi Tier session:

- `fetch_hit_immediate = 0` in every sample — the decoder's fetch never found
  blocks already staged.
- Supply delay (fetch parked -> matched by a store): mean **15.83 s**, max
  **29.92 s**, i.e. truncated by the 30 s deadline rather than bounded by
  transfer cost. The transfers themselves run at ~45 GiB/s and account for ~14 s
  of work across a 26-minute run.
- 6,083 load timeouts logged against 10,800 requests.

With the legs sequenced (prefill awaited, then decode dispatched):

| | concurrent | sequential |
| --- | ---: | ---: |
| supply delay mean | 15.830 s | **0.002 s** |
| supply delay max | 29.920 s | **0.234 s** |
| consumer load latency mean | 14.387 s | **0.291 s** |

### Proposed fix

Run the prefill leg to completion, then dispatch decode, returning the prefill
error verbatim if it fails (decode is then never dispatched, so it cannot fetch
KV that was never stored). This matches the NIXL connector's ordering.

A change is prepared on `fix/p2p-sequential-dispatch`
(`pkg/sidecar/proxy/connector_p2p.go`, +57/-54); `make test-unit-sidecar` passes.
I will open the PR against this issue.

### Note

This does not fix everything. There is a separate upstream vLLM defect where a
producer that already holds the requested blocks supplies nothing at all, so the
consumer still times out regardless of ordering — filed as
https://github.com/vllm-project/vllm/issues/52808. The sequencing change above is
independently correct and removes the ordering-induced share of the timeouts.

nilig:
**What type of PR is this?**

/kind bug

The PD Multi Tier decode leg pulls KV from the prefiller's secondary tier.
`handleP2P` dispatched both legs concurrently, so the decoder's fetch reached the
prefiller before any blocks were stored for that `kv_request_id`. The prefiller
parks that demand awaiting supply, and the decoder spends its fixed load deadline
(`_LOAD_TIMEOUT_S`, 30s in vLLM) waiting for KV that has not been produced yet.
On expiry it aborts the load and, under `kv_load_failure_policy: recompute`,
recomputes the prompt locally — the work disaggregation exists to avoid.

This runs the prefill leg to completion first and returns the prefill error
verbatim if it fails, so decode is never dispatched against KV that was never
stored. It matches the `nixlv2` path, which already forwards to the prefiller and
waits for it to return; concurrent dispatch there is opt-in behind
`MoRIIOParallelDispatch && MoRIIOWriteMode`.

Measured on 8 prefill + 8 decode (gpt-oss-120b, TP=1, vLLM v0.27.1, 16x H200),
instrumenting the vLLM Multi Tier session:

| | concurrent | sequential |
| --- | ---: | ---: |
| supply delay mean | 15.830 s | 0.002 s |
| supply delay max | 29.920 s | 0.234 s |
| consumer load latency mean | 14.387 s | 0.291 s |

This does not remove every timeout: a producer that already holds the requested
blocks supplies nothing regardless of ordering, tracked upstream as
https://github.com/vllm-project/vllm/issues/52808.

Fixes #2438

**Test plan**:

- [x] Decode is not dispatched until the prefill leg returns (prefill held open;
      decode request count stays zero, then fires exactly once on release).
- [x] A failing prefill is returned to the caller verbatim and decode is never
      dispatched.
- [x] Existing PD Multi Tier `kv_transfer_params` coverage still passes.

Both new specs were confirmed to fail against the previous concurrent
implementation and pass with this change.

```release-note
Fixed the PD Multi Tier decode leg being dispatched before the prefiller had stored its KV, which charged prefill latency against the KV load deadline and caused requests to time out and recompute under load.
```


pkg/sidecar/proxy/connector_p2p.go
```diff
@@ -193,32 +208,21 @@ func (s *Server) handleP2PConcurrentRequests(w http.ResponseWriter, r *http.Requ
...
-		)
 	}
+	decodeSpan.SetAttributes(
+		attribute.Float64("llm_d.pd_proxy.total_duration_ms", float64(totalDuration.Milliseconds())),
+		attribute.Float64("llm_d.pd_proxy.true_ttft_ms", float64(trueTTFT.Milliseconds())),
+		attribute.Float64("llm_d.pd_proxy.decode_duration_ms", float64(decodeDuration.Milliseconds())),
+		attribute.Bool("llm_d.pd_proxy.concurrent_pd", false),
+	)
 }
-
```
roytman, line 215: Q. This doc comment got dropped in this PR. p2pPullAvailable itself isn't otherwise touched by this fix, so this looks like accidental collateral damage rather than an intended edit or a reduction of the comments. Mind restoring it?

nilig: Good catch, that was accidental - the comment and its preceding blank line got swallowed when the concurrent helper was replaced. Restored verbatim in `e756fc43`.

I also re-checked the rest of the diff for the same problem; everything else removed is the old concurrent path (its doc comment described concurrent dispatch, so it was replaced rather than dropped) and the now-unused `context` import.

roytman (dismissed):

vMaroon (approved):

PR #2455, opened 2026-08-19
feat: auto-tune approx prefix cache block size from prefix_match_unit

RishabhSaini:
/kind feature

Auto-tunes the approx-prefix-cache-producer's `blockSizeTokens` from vLLM's `prefix_match_unit` metric when available.

For Mamba-style models (Kimi, Nemotron), the physical KV cache block size is very large (e.g. 4096 tokens). The approx producer previously auto-tuned from `CacheBlockSize` (physical), producing zero routing signal for prompts shorter than 4096 tokens. vLLM already exposes `prefix_match_unit` as a label on `vllm:cache_config_info` — this is the granularity at which vLLM internally matches prefix cache hits (e.g. 32 tokens for hybrid models).

When `prefix_match_unit` is present and positive, auto-tune now prefers it over `block_size`. The existing `minBlockSizeTokens` floor (64) still applies, so a `prefix_match_unit` of 32 is clamped to 64 — still 64x better than 4096 for routing signal. The floor protects the LRU indexer from memory growth at very small block sizes (#1158). For standard attention models where `prefix_match_unit` is not set, behavior is unchanged.

**Changes:**
- `Metrics` struct: added `CachePrefixMatchUnit` field
- Metrics extractor: reads `prefix_match_unit` label from `vllm:cache_config_info` gauge
- `GetBlockSize()`: prefers `CachePrefixMatchUnit` over `CacheBlockSize` when auto-tuning

Fixes #

**Release note**:
```release-note
The approx-prefix-cache-producer now auto-tunes block size from vLLM's prefix_match_unit metric when available, improving cache-aware routing for Mamba-style models with large physical block sizes.
```

## Test plan
- [x] Existing approx prefix cache tests pass
- [x] Metrics extractor tests pass
- [x] Lint passes (only pre-existing unparam issue in unrelated file)
- [ ] Verify with a Mamba-style model that `prefix_match_unit` label appears in `vllm:cache_config_info` and auto-tune picks it up

### DIFF
--- a/pkg/epp/framework/interface/datalayer/metrics.go
+++ b/pkg/epp/framework/interface/datalayer/metrics.go
@@ -34,6 +34,7 @@ type Metrics struct {
 	KVCacheUsagePercent     float64
 	KvCacheMaxTokenCapacity int
 	CacheBlockSize          int
+	CachePrefixMatchUnit    int
 	// Number of GPU blocks in the model server for KV Cache.
 	CacheNumBlocks int
 
@@ -76,6 +77,7 @@ func (m *Metrics) Clone() *Metrics {
 		KVCacheUsagePercent:     m.KVCacheUsagePercent,
 		KvCacheMaxTokenCapacity: m.KvCacheMaxTokenCapacity,
 		CacheBlockSize:          m.CacheBlockSize,
+		CachePrefixMatchUnit:    m.CachePrefixMatchUnit,
 		CacheNumBlocks:          m.CacheNumBlocks,
 		UpdateTime:              m.UpdateTime,
 	}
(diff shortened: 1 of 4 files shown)


pkg/epp/framework/interface/datalayer/metrics.go
```diff
@@ -34,6 +34,7 @@ type Metrics struct {
 	KVCacheUsagePercent     float64
 	KvCacheMaxTokenCapacity int
 	CacheBlockSize          int
+	CachePrefixMatchUnit    int
```
ahg-g, line 37: pls document the semantics of this parameter, hard to deduce that from the name.

ahg-g (approved):

PR #2458, opened 2026-08-19
refactor(flowcontrol): derive QueueOutcome from the finalization error

### LINKED ISSUE #2168: [Flow Control] Derive QueueOutcome from the finalization error instead of storing both

Make `QueueOutcome` a pure function of the finalization error instead of a second, independently chosen value.

Every finalization site today picks an (outcome, error) pair by hand, and the two encode the same fact twice. The error's outer sentinel (`ErrRejected` vs `ErrEvicted`) already carries the pre/post-admission split, and the inner sentinel (`ErrQueueAtCapacity`, `ErrNoEndpoints`, `ErrTTLExpired`, `ErrContextCancelled`, `ErrFlowControllerNotRunning`) carries the cause. #2097 was the pairing drifting out of agreement, and #2162 re-aligned it by hand in the translator.

The mapping also exists twice in helper form, in opposite directions. The queue-wait sweep picks an outcome in `isExpired` and reconstitutes the error with `expiryError(outcome)`, while `Finalize(cause)` derives the outcome from the error in `inferOutcome`. #2284 added `QueueOutcomeRejectedNoEndpoints`, `QueueOutcomeEvictedNoEndpoints`, and an eviction error that wraps two inner sentinels (`ErrEvicted: ErrTTLExpired: ErrNoEndpoints`), bringing the hand-paired surface to roughly fourteen sites across `processor.go` and `controller.go`, plus two synthesized pairs inside `EnqueueAndWait`.

Proposed shape:

- Add `types.OutcomeFromError(err) (QueueOutcome, bool)`. nil maps to `Dispatched`. The `ErrEvicted` family maps to `EvictedNoEndpoints`/`EvictedTTL`/`EvictedContextCancelled`/`EvictedOther` by inner sentinel, checking `ErrNoEndpoints` before `ErrTTLExpired` because the no-endpoint expiry wraps both. The `ErrRejected` family maps to `RejectedNoEndpoints`/`RejectedCapacity`/`RejectedOther` the same way. An error missing a family sentinel is an invariant violation: return `ok=false`, log it, and classify as `RejectedOther`.
- `finalizeInternal` computes the outcome from the error. The item API becomes `Finalize(cause)`, which wraps by queued-ness as it does today, plus `FinalizeWithError(err)` taking a pre-wrapped error, nil for dispatch. `inferOutcome` and the sweep's outcome-keyed `expiryError` collapse into the one mapping: the sweep builds the error and the outcome is derived.
- `translateFlowControlOutcome` becomes `translateFlowControlError(err, ttlPoolEmpty)` and switches on sentinels alone, removing the recursion that re-entered `RejectedOther` wrapping `ErrTTLExpired` as `EvictedTTL`. The delegation semantics from #2162 stay: pre-admission TTL/cancel still finalizes in the `ErrRejected` family on purpose, since rejected items never consumed queue capacity and the metrics split relies on that. `ttlPoolEmpty` stays a parameter because pool emptiness is a live probe whose answer the error cannot carry.
- Add a table test asserting the derived outcome for every producer's current error expression, with a coverage check that every `QueueOutcome` value has a producer row.

Every current producer already wraps the correct sentinels, so the change is mechanical: `item.go`, `processor.go`, `controller.go`, and their tests. `EnqueueAndWait`'s signature and the metrics label vocabulary do not change.

**Why is this needed**:

Two representations of one fact stay in agreement only by convention. #2097 is what the drift looks like in practice: a producer and the translator disagreed about what `RejectedOther` wrapping `ErrTTLExpired` means, and requests that should have been 503s became 500s. After this change, the only way to get the outcome wrong is to wrap the wrong sentinel, which is the same single decision the translator already depends on.

Part of #1187.

LukeAVanDrie:
**What type of PR is this?**

/kind cleanup

Makes `QueueOutcome` a pure function of the finalization error instead of a second, independently chosen value.

Every finalization site picked an (outcome, error) pair by hand, and the pair encodes one fact twice: the outer sentinel (`ErrRejected` vs `ErrEvicted`) carries the pre/post-admission split and the inner sentinel carries the cause. #2097 came from the pairing drifting out of agreement. The mapping also existed twice in helper form, in opposite directions: the queue-wait sweep derived the error from an outcome (`expiryError(outcome)`) while `inferOutcome` derived an outcome from an error.

- `types.OutcomeFromError(err)` is the single mapping. `finalizeInternal` derives the outcome from the error; an error missing a family sentinel is logged as an invariant violation and classified `RejectedOther`.
- `FlowItem.FinalizeWithOutcome(outcome, err)` becomes `FinalizeWithError(err)`, nil meaning dispatched. `inferOutcome` splits into `wrapCause` plus the shared mapping. The sweep builds the error and the outcome is derived, so the inverse helper is gone.
- `translateFlowControlOutcome` becomes `translateFlowControlError(err, poolEmpty)` and switches on error sentinels, removing the recursion that re-entered `RejectedOther` wrapping `ErrTTLExpired` as `EvictedTTL`. Pre-admission TTL/cancel still finalizes in the `ErrRejected` family (#2162), since the metrics split relies on rejected items never having consumed queue capacity; the single `ErrTTLExpired` arm covers both families. Pool emptiness is a live state the error cannot carry, so it is passed as a `poolEmpty func() bool` and called only in the TTL arm.
- `EnqueueAndWait`'s signature and the metrics label vocabulary are unchanged. `TestOutcomeFromError` pins the derived outcome for every producer error expression and fails when a `QueueOutcome` value has no producer row. The integration suite is untouched and asserts observed (outcome, error) pairs end to end.

Behavior changes:

- The blocking-submit failure path wraps `ErrRejected` once: `distributeRequest` wraps, and `tryDistribution` returns the finalized error unchanged.
- Processor drop counts record the outcome stored on the item (`finalizeAndRecordDrop`) instead of a literal at the call site. Under a finalization race the summary counts the winner's outcome, matching `flow_control_requests_total`, which already recorded the winner.
- The enqueue path's two registry-lookup rejections flatten the registry error with `%v`, matching the existing convention in `tryDistribution`: a `%w`-preserved `ErrPriorityBandNotFound` returned through the connection closure would be misread by `withConnectionWithFallback` as a lease-acquisition failure and retried at priority 0.
- Flow control errors that map to HTTP 500 carry `x-llm-d-request-dropped-reason: rejected-internal`, so every flow control drop response now carries the reason header.
- A bare `context.DeadlineExceeded` finalization cause is wrapped as `ErrTTLExpired: <cause>`, preserving the original error in the chain like the cancellation arm.

Fixes #2168

```release-note
Flow control drop responses that map to HTTP 500 now carry the header `x-llm-d-request-dropped-reason: rejected-internal`.
```

**Test plan**:

- [x] Derived outcome matches the stored outcome for every producer error expression, and every `QueueOutcome` value has a producer row
- [x] Expiry errors carry the regime sentinels for the saturation and no-endpoint regimes
- [x] Pre- and post-admission TTL and cancellation map to the same HTTP statuses from the error alone, and errors without a family sentinel map to 500 with the `rejected-internal` reason header
- [x] A bare deadline-exceeded cause classifies as a TTL outcome with the original error preserved in the chain

### DIFF
--- a/pkg/epp/flowcontrol/controller/internal/item.go
+++ b/pkg/epp/flowcontrol/controller/internal/item.go
@@ -25,6 +25,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/go-logr/logr"
+
 	"github.com/llm-d/llm-d-router/pkg/epp/flowcontrol/types"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/flowcontrol"
 	"github.com/llm-d/llm-d-router/pkg/epp/metrics"
@@ -54,12 +56,13 @@ type FlowItem struct {
 	enqueueTime     time.Time
 	effectiveTTL    time.Duration
 	originalRequest flowcontrol.FlowControlRequest
+	logger          logr.Logger
 
 	// --- Synchronized State ---
 
 	// handle stores the types.QueueItemHandle atomically.
 	// Written by the Processor (SetHandle) when admitted.
-	// Read by inferOutcome (called by Finalize) to infer the outcome (Rejected vs. Evicted).
+	// Read by Finalize to select the error family (Rejected vs. Evicted).
 	// Distinguishing between pre-admission (Rejection) and post-admission (Eviction) during asynchronous finalization
 	// relies on whether this handle is nil or non-nil.
 	handle atomic.Pointer[flowcontrol.QueueItemHandle]
@@ -81,11 +84,17 @@ type FlowItem struct {
 var _ flowcontrol.QueueItemAccessor = &FlowItem{}
 
 // NewItem allocates and initializes a new FlowItem for a request lifecycle.
-func NewItem(req flowcontrol.FlowControlRequest, effectiveTTL time.Duration, enqueueTime time.Time) *FlowItem {
+func NewItem(
+	req flowcontrol.FlowControlRequest,
+	effectiveTTL time.Duration,
+	enqueueTime time.Time,
+	logger logr.Logger,
+) *FlowItem {
 	return &FlowItem{
 		enqueueTime:     enqueueTime,
 		effectiveTTL:    effectiveTTL,
 		originalRequest: req,
+		logger:          logger,
 		done:            make(chan *FinalState, 1),
 	}
 }
@@ -128,27 +137,33 @@ func (fi *FlowItem) SetHandle(handle flowcontrol.QueueItemHandle) { fi.handle.St
 func (fi *FlowItem) Finalize(cause error) {
 	fi.onceFinalize.Do(func() {
 		// Atomically load the handle to determine if the item was admitted to a queue.
-		// This synchronization is critical for correctly inferring the outcome across goroutines.
+		// This synchronization is critical for correctly classifying the outcome across goroutines.
 		isQueued := fi.Handle() != nil
-		outcome, finalErr := inferOutcome(cause, isQueued)
-		fi.finalizeInternal(outcome, finalErr)
+		fi.finalizeInternal(wrapCause(cause, isQueued))
 	})
 }
 
-// FinalizeWithOutcome sets the item's terminal state explicitly.
+// FinalizeWithError sets the item's terminal state from a pre-wrapped finalization error: nil for dispatch, otherwise
+// an error wrapping `types.ErrRejected` or `types.ErrEvicted`. The outcome is derived from the error.
 //
 // This method is intended for synchronous finalization by the Processor (Dispatch, Reject) or the Controller
 // (Distribution failure).
 // It is idempotent.
-func (fi *FlowItem) FinalizeWithOutcome(outcome types.QueueOutcome, err error) {
+func (fi *FlowItem) FinalizeWithError(err error) {
 	fi.onceFinalize.Do(func() {
-		fi.finalizeInternal(outcome, err)
+		fi.finalizeInternal(err)
 	})
 }
 
 // finalizeInternal is the core finalization logic. It must be called within the sync.Once.Do block.
-// It captures the state, stores it atomically, and signals the Done channel.
-func (fi *FlowItem) finalizeInternal(outcome types.QueueOutcome, err error) {
+// It derives the outcome from the error, captures the state, stores it atomically, and signals the Done channel.
+func (fi *FlowItem) finalizeInternal(err error) {
+	outcome, ok := types.OutcomeFromError(err)
+	if !ok {
+		fi.logger.Error(err, "Invariant violation: finalization error wraps neither ErrRejected nor ErrEvicted",
+			"requestID", fi.originalRequest.ID(), "flowKey", fi.originalRequest.FlowKey())
+	}
+
 	finalState := &FinalState{
 		Outcome: outcome,
 		Err:     err,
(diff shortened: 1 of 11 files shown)


LukeAVanDrie:
@RishabhSaini

pkg/epp/flowcontrol/controller/internal/item.go
```diff
@@ -169,30 +184,26 @@ func (fi *FlowItem) finalizeInternal(outcome types.QueueOutcome, err error) {
...
-// was already admitted to a queue.
-func inferOutcome(cause error, isQueued bool) (types.QueueOutcome, error) {
+// wrapCause normalizes a finalization cause (e.g., a Context error) into the canonical sentinel chain and wraps it
+// with the family sentinel that matches the item's admission status.
+func wrapCause(cause error, isQueued bool) error {
 	var specificErr error
-	var outcomeIfEvicted types.QueueOutcome
 	switch {
 	case errors.Is(cause, types.ErrTTLExpired) || errors.Is(cause, context.DeadlineExceeded):
 		specificErr = types.ErrTTLExpired
```
RishabhSaini:
```suggestion
		specificErr = fmt.Errorf("%w: %w", types.ErrTTLExpired, cause)
```
preserve the original cause for deadline expiry like in the case for deadline to maintain consistency

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -776,17 +770,22 @@ func (p *Processor) evictAll() {
...
-			item.FinalizeWithOutcome(outcome, errShutdown)
-			p.recordDrop(item.FinalState().Outcome)
+			p.finalizeAndRecordDrop(item, errShutdown)
 		}
 	}
 	p.processAllQueuesConcurrently("evictAll", processFn)
 }
 
+// finalizeAndRecordDrop finalizes item with err and counts the outcome actually stored. Finalization is idempotent,
+// so under a race with an external finalization the winner's outcome is the one counted.
```
RishabhSaini: worth mentioning that the contention is b/w the controller's thread resulting in `TTLExpiry` and processoer's thread resulting in `QueueAtCapacity` which the sync.Once resolved

pkg/epp/requestcontrol/admission.go
```diff
@@ -237,53 +230,41 @@ func (r *flowControlRequest) FlowKey() flowcontrol.FlowKey {
...
-		case errors.Is(err, types.ErrTTLExpired):
-			return translateFlowControlOutcome(types.QueueOutcomeEvictedTTL, err, ttlPoolEmpty)
-		case errors.Is(err, types.ErrContextCancelled):
-			return translateFlowControlOutcome(types.QueueOutcomeEvictedContextCancelled, err, ttlPoolEmpty)
-		default:
-			return errcommon.Error{Code: errcommon.Internal, Msg: "internal flow control error: " + msg}
-		}
 	default:
-		return errcommon.Error{Code: errcommon.Internal, Msg: "unhandled flow control outcome: " + msg}
+		return errcommon.Error{Code: errcommon.Internal, Msg: "internal flow control error: " + msg}
```
RishabhSaini:
```suggestion
		return errcommon.Error{Code: errcommon.Internal, Msg: "internal flow control error: " + msg, Headers: map[string]string{errcommon.RequestDroppedReasonHeaderKey: "rejected-internal"}}
```
it should almost never fire but being consistent here if filtering by error key

RishabhSaini (dismissed):
/lgtm
Thanks @LukeAVanDrie for the much needed cleanup here. The single OutcomeFromError mapping eliminates the dual-encoding drift that caused #2097. The exhaustive test with the coverage loop is the right guard against future regressions.
Few nits mentioned inline to improve consitency and error propogation and please rebase

pkg/epp/flowcontrol/controller/internal/item.go
LukeAVanDrie: Done, with one adjustment. The cause reaching this arm is usually `types.ErrTTLExpired` itself (set via `context.WithDeadlineCause`), and wrapping that again would repeat the sentinel in the message. So the arm is split: a cause already carrying `ErrTTLExpired` is kept as-is, and a bare `context.DeadlineExceeded` is wrapped as `ErrTTLExpired: <cause>`, preserving the original error like the cancellation arm. Added a classification test case for the bare-deadline cause.

pkg/epp/flowcontrol/controller/internal/processor.go
LukeAVanDrie: Done. The comment now names the contending goroutines: the controller finalizing on TTL expiry against the processor's capacity rejection, with the winner picked by the item's sync.Once.

pkg/epp/requestcontrol/admission.go
LukeAVanDrie: Done, via a typed constant (`RequestDroppedReasonInternal = "rejected-internal"`) to match the existing header vocabulary rather than a string literal. The translate-level and Admit-level tests pin the header, and the release note is no longer NONE since this adds a value to a documented header.

RishabhSaini:
all comments resolved, needs one more rebase

LukeAVanDrie:
@RishabhSaini remaining test failure appears to be a flake

RishabhSaini (approved):

ahg-g (approved):

PR #2460, opened 2026-08-19
fix(charts): roll the EPP pod when its ConfigMaps change

### LINKED ISSUE #2459: EPP pod template has no ConfigMap checksum, so a plugin-config-only helm upgrade never restarts the EPP
The EPP pod template carries no annotation derived from the ConfigMaps it mounts, so a
`helm upgrade` that changes only plugin configuration leaves the pod template byte-identical.
Kubernetes correctly declines to roll, the running EPP keeps the configuration it parsed at
startup, and every signal a user would check reports success.

Verified against `main` at `38add279` and the published `llm-d-router-standalone:v0`
(`sha256:3efb0608…`), 2026-08-19.

## Reproduction

Render the same release twice, changing nothing but `router.epp.pluginsCustomConfig`:

```
helm template test oci://ghcr.io/llm-d/charts/llm-d-router-standalone --version v0 \
  -f base.values.yaml -f plugins-<variant>.values.yaml \
  --set router.modelServers.matchLabels.app=demo -n testns
```

| | `featureGates: []` | `featureGates: [flowControl]` |
|---|---|---|
| ConfigMap `data` | differs | differs |
| pod template sha1 | `64ac885bafbacacc4bb779ae07e3d39b4e290270` | `64ac885bafbacacc4bb779ae07e3d39b4e290270` |

The proxy ConfigMap behaves the same way — changing `router.proxy.configMap.data` leaves the
pod template at `4d96427eaecca2802581897246ccafb908ce1dc2` either way.

`config/charts/routerlib/templates/_deployment.yaml` has one `annotations:` block on each of
the `StatefulSet` and `Deployment` branches, and it is only:

```yaml
      {{- with .Values.router.epp.podAnnotations }}
      annotations:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

## Why the config never takes effect

`cmd/epp/runner/runner.go` reads the config file once and caches it, with a comment saying so:

```go
// parseConfigurationPhaseOne is idempotent: Run() calls it to decide
// between the K8s and file-discovery paths, and the K8s path's setup()
// then calls it a second time. Cache the parsed config so we don't
// re-read the file, ...
if r.rawConfig != nil {
	return r.rawConfig, nil
}
```

There is no watcher on that path; the repo's only `fsnotify` use is the file-based
endpoint-discovery plugin. The plugins volume is a whole-directory `configMap` mount at
`/config` with no `subPath`, so the kubelet does keep the file on disk current — nothing
re-reads it.

## Why this is worth an S1 rather than an inconvenience

Every available success signal is true and irrelevant:

- `helm upgrade` prints `Release "..." has been upgraded.` and bumps the revision.
- `kubectl get cm <release>-epp -o yaml` shows the new configuration, because the kubelet
  syncs the mounted file regardless of whether anything re-reads it.
- `kubectl rollout status deploy/<release>-epp` prints `successfully rolled out`
  **immediately**, because the Deployment genuinely is at its desired state. This is the
  worst of the three: it is the command you would reach for to check exactly this, and on an
  unchanged pod template it cannot distinguish "rolled out with the new config" from "never
  rolled, because there was nothing to roll".

The only signal that told the truth was the EPP's own startup log still saying
`Flow Control layer is disabled via the flowControl feature gate, using legacy admission
control` — and by then that line was three days old, so even it looked like it might be stale.

We caught this because a preflight greps that log line. Had it checked the ConfigMap instead —
the obvious thing to check, and what "verify the config landed" would ordinarily mean — a
fifteen-cell benchmark would have run entirely under legacy admission control, produced
plausible numbers, and been published as "EPP flow control makes no measurable difference".

## Suggested fix

Add the conventional checksum annotation to both pod templates. Hashing the whole
`llm-d-epp.config` partial rather than one ConfigMap covers all four it renders (EPP plugins,
proxy config, and the two latency-predictor ones), every one of which is mounted into this
same pod:

```yaml
      annotations:
        checksum/config: {{ include "llm-d-epp.config" . | sha256sum }}
        {{- with .Values.router.epp.podAnnotations }}
        {{- toYaml . | nindent 8 }}
        {{- end }}
```

Happy to open the PR.

shimib:

The EPP pod template carries no annotation derived from the ConfigMaps it mounts, so a
`helm upgrade` that changes only plugin configuration leaves the pod template byte-identical
and Kubernetes correctly declines to create a new ReplicaSet. `cmd/epp/runner/runner.go`
reads the config file once and caches it in `r.rawConfig`, and nothing watches that path, so
the running pod keeps the configuration it parsed at startup.

What makes this worth fixing rather than documenting is that every signal reports success:
`helm upgrade` bumps the revision, `kubectl get cm` shows the new config (the kubelet syncs
the mounted file regardless of whether anything re-reads it), and `kubectl rollout status`
returns `successfully rolled out` immediately, because the Deployment genuinely is at its
desired state. That last one is the command you would reach for to check exactly this, and on
an unchanged pod template it cannot distinguish "rolled out with the new config" from "never
rolled, because there was nothing to roll".

This adds the conventional checksum annotation to both the `Deployment` and `StatefulSet` pod
templates, through a shared `llm-d-epp.podAnnotations` helper so the two branches cannot drift.

The hash is taken over the whole `llm-d-epp.config` partial rather than a single ConfigMap.
That partial renders four ConfigMaps — EPP plugins, the proxy config, and the two
latency-predictor ones — and all of them are mounted into this same pod, so hashing the
partial covers them in one annotation. The proxy ConfigMap has the same defect today:
changing `router.proxy.configMap.data` currently leaves the pod template untouched.

User-supplied `router.epp.podAnnotations` are still emitted, after the checksum.

**Verification**

Rendered `llm-d-router-standalone` and `llm-d-router-gateway` from this branch, comparing the
sha1 of the pod template:

| change | before this PR | after |
|---|---|---|
| `pluginsCustomConfig` `featureGates: []` → `[flowControl]` | identical | differs |
| `proxy.configMap.data.extra.txt` `one` → `two` | identical | differs |
| nothing (same values rendered twice) | identical | identical |

Also swept the 32-combination cross product of `monitoring.prometheus.enabled` ×
`monitoring.provider.name` (`PrometheusOperator`/`GMP`) × `monitoring.prometheus.auth.enabled`
× `inferencePool.create` × `epp.replicas` (1/2) on both charts:

| | combos | pod templates carrying `checksum/config` | fused YAML documents |
|---|---|---|---|
| `main` | 32 | 0 | 0 |
| this branch | 32 | 32 | 0 |

`helm lint` passes on both charts. The `StatefulSet` branch
(`gkePreferredBackends.enabled=true`) and `podAnnotations` merging were checked separately.

Fixes #2459

**Release note**:
```release-note
The EPP pod now restarts when its ConfigMaps change. A `helm upgrade` that modifies only plugin, proxy or latency-predictor configuration previously left the running EPP on its old configuration while reporting success.
```

### DIFF
--- a/config/charts/routerlib/templates/_helpers.tpl
+++ b/config/charts/routerlib/templates/_helpers.tpl
@@ -516,3 +516,19 @@ Deprecation validations
 {{- fail "Top-level 'inferenceObjectives' is deprecated. Please migrate your values to 'router.inferenceObjectives'." }}
 {{- end }}
 {{- end -}}
+
+{{/*
+Annotations for the EPP pod template.
+
+The EPP parses its plugin configuration once at startup and never re-reads the
+mounted file, so a helm upgrade that touches only a ConfigMap has to change the
+pod template or the running pod keeps its old configuration. Hash the whole
+config partial rather than one ConfigMap: it renders the plugins, proxy and
+latency-predictor ConfigMaps, and all of them are mounted into this pod.
+*/}}
+{{- define "llm-d-epp.podAnnotations" -}}
+checksum/config: {{ include "llm-d-epp.config" . | sha256sum }}
+{{- with .Values.router.epp.podAnnotations }}
+{{- toYaml . | nindent 0 }}
+{{- end }}
+{{- end -}}
--- a/config/charts/routerlib/templates/_deployment.yaml
+++ b/config/charts/routerlib/templates/_deployment.yaml
@@ -317,10 +317,8 @@ spec:
       labels:
         {{- include "llm-d-router.selectorLabels" . | nindent 8 }}
         {{- include "llm-d-router.modeLabels" . | nindent 8 }}
-      {{- with .Values.router.epp.podAnnotations }}
       annotations:
-        {{- toYaml . | nindent 8 }}
-      {{- end }}
+        {{- include "llm-d-epp.podAnnotations" . | nindent 8 }}
     spec:
 {{ include "llm-d-epp.deployment-pod-spec" . }}
 ---
@@ -351,10 +349,8 @@ spec:
       labels:
         {{- include "llm-d-router.selectorLabels" . | nindent 8 }}
         {{- include "llm-d-router.modeLabels" . | nindent 8 }}
-      {{- with .Values.router.epp.podAnnotations }}
       annotations:
-        {{- toYaml . | nindent 8 }}
-      {{- end }}
+        {{- include "llm-d-epp.podAnnotations" . | nindent 8 }}
     spec:
 {{ include "llm-d-epp.deployment-pod-spec" . }}
 ---


ZimmerHao:
When will this PR be merged? I’ve run into the same issue.

config/charts/routerlib/templates/_helpers.tpl
```diff
@@ -516,3 +516,19 @@ Deprecation validations
...
+{{/*
+Annotations for the EPP pod template.
+
+The EPP parses its plugin configuration once at startup and never re-reads the
+mounted file, so a helm upgrade that touches only a ConfigMap has to change the
+pod template or the running pod keeps its old configuration. Hash the whole
+config partial rather than one ConfigMap: it renders the plugins, proxy and
+latency-predictor ConfigMaps, and all of them are mounted into this pod.
+*/}}
+{{- define "llm-d-epp.podAnnotations" -}}
```
elevran, line 529: One (**unlikely**) edge case worth a look: if `router.epp.podAnnotations` sets a key named `checksum/config`, the rendered pod template ends up with two `checksum/config` keys in the same map. Not a blocker, but worth either merging so the computed hash always wins, or a deliberate decision to leave it as a reserved key.

Suggest building one merged dict and rendering it once, using `merge` so first argument's keys are never overwritten by later arguments:

```go
{{- define "llm-d-epp.podAnnotations" -}}
{{- $annotations := merge (dict "checksum/config" (include "llm-d-epp.config" . | sha256sum)) (.Values.router.epp.podAnnotations | default dict) }}
{{- toYaml $annotations }}
{{- end -}}
```

the `default dict` guards against nils which merge does not like.

elevran (approved):

elevran:
@shimib leaving https://github.com/llm-d/llm-d-router/pull/2460#discussion_r3947664085 for a follow up. Would be good if you can address.

PR #2462, opened 2026-08-19
fix(scheduling): return typed capacity rejection when filters drain candidates

### LINKED ISSUE #2428: Scheduling drain returns an untyped 429 with no drop-reason header
**What happened**:

When scheduling filters remove every candidate endpoint, the client receives `429 Too Many Requests` with no `x-llm-d-request-dropped-reason` header. No code path selects that response. It is produced by the following sequence:

1. `pkg/epp/scheduling/scheduler_profile.go:129` returns `errcommon.Error{Code: Internal}` for "no endpoints available for the given request". `Internal` maps to 500, which misclassifies a capacity condition as a server fault.
2. `pkg/epp/scheduling/scheduler.go:81-88` discards the error. It is logged at VERBOSE and the map entry is set to `nil`; the error value is never retained. Profile handlers cannot see it, and every handler that checks for a nil result raises a fresh untyped error (`single_profile_handler.go:91`, `dp_profile_handler.go:112`, `header_profile_handler.go:208`, and `pd_profile_handler.go:244` for a failed decode profile).
3. `pkg/epp/requestcontrol/director.go:340` runs `errors.As` for a typed `errcommon.Error`, misses, and falls through to `ResourceExhausted` on line 344, a branch whose comment describes it as the legacy "no endpoint" status.

A scheduling drain classified the same way flow control classifies its rejections (`pkg/epp/requestcontrol/admission.go:256-259`): `ResourceExhausted` with `x-llm-d-request-dropped-reason: rejected-saturated` for a capacity rejection, `ServiceUnavailable` with `rejected-no-endpoints` when there is no serving capacity. The scheduling path emits neither the code nor the header, so operators cannot distinguish a scheduling drain from any other 429, and dashboards keyed on the drop-reason header miss these requests.

Configure a scheduling profile with any filter that can eliminate all endpoints, send a request while the filter's condition holds against a non-empty pool, and inspect the response: 429 with no `x-llm-d-request-dropped-reason` header. At the code level: `Scheduler.Schedule` with a profile whose filter returns an empty endpoint list yields an untyped error (`failed to run scheduler profile '<name>'`) from which `errors.As` cannot extract an `errcommon.Error`, which triggers the director fallback.

The header profile handler's README (`pkg/epp/framework/plugins/scheduling/profilehandler/headerprofile/README.md`) documents the same fallback as misleading for its no-profile-selected case and ties the missing typed-error plumbing to #1686. The scope of the fix is the profile-drain case: a profile ran and its filters emptied the candidate set. The no-profile-selected case (untyped error at `scheduler.go:92-94`) remains with #1686.

The missing status also blocks making the saturation detectors' fail-open behavior configurable: a `failOpen: false` setting requires the resulting rejection to carry a status the client can act on.

Proposed fix:

- `scheduler_profile.go:129`: return `ResourceExhausted` with the `rejected-saturated` drop reason, matching `admission.go:256`. The empty-pool case never reaches the profile run. The director returns `ServiceUnavailable` when the pool has no candidates (`director.go:305-320`, covering an empty Locate result and screeners eliminating every endpoint), so a drain during a profile run means the pool was non-empty and filters emptied it.
- `scheduler.go`: retain the per-profile errors, keyed by profile name so a profile that fails in one iteration and succeeds in a later one leaves no stale entry, and join them into the error returned from `Schedule` so a typed code survives to the director's `errors.As`. The change covers all profile handlers without modifying them. The P/D handler treats a failed prefill profile with a successful decode as a success; the join applies only when `ProcessResults` itself errors, so that behavior is unchanged.
- `director.go:305-320`: the pre-scheduling `ServiceUnavailable` rejections carry the `rejected-no-endpoints` drop reason. The status codes are already correct; the header makes every pre-dispatch rejection visible to dashboards keyed on `x-llm-d-request-dropped-reason`, including a rejection produced by a saturation detector configured to fail closed.

A non-load filter that empties the candidate set also reports `rejected-saturated`. Attributing a drain to a specific filter would require carrying a reason through the `Filter` interface, a larger change than the problem warrants.

Clients continue to receive 429; there is no user-visible status change. The response gains the drop-reason header, and the status is set at the rejection site instead of by the director fallback.

**Environment**:
- Kubernetes version (use `kubectl version`): n/a, code-level defect
- llm-d-router version (use `git describe --tags --dirty --always` if you built from source, or specify the tag if you used a tagged version or image): main at 7987423a
- Install tools: n/a
- Others: n/a

LukeAVanDrie:
**What type of PR is this?**

/kind bug

When scheduling filters remove every candidate endpoint, the profile run returns an `Internal` error that `Scheduler.Schedule` drops after logging. Profile handlers see a nil result and raise fresh untyped errors, the director's `errors.As` finds no typed error, and the legacy `ResourceExhausted` fallback sets the response: a 429 with no `x-llm-d-request-dropped-reason` header. Operators cannot distinguish a scheduling drain from any other 429, and dashboards keyed on the drop-reason header miss the requests.

The fix classifies each rejection at the site that detects it, with the status code and drop-reason header flow control uses for the matching condition:

- `SchedulerProfile.Run` returns `ResourceExhausted` with the `rejected-saturated` header when filters drain a non-empty candidate set. An empty pool is rejected in the director before any profile runs.
- `Scheduler.Schedule` retains per-profile errors, keyed by profile name so a profile that fails in one iteration and succeeds in a later one leaves no stale entry, and joins them into a `ProcessResults` error in sorted-name order. The typed code survives to the director's `errors.As` without modifying any profile handler. The P/D handler's treatment of a failed prefill profile with a successful decode as a success is unchanged because the join applies only when `ProcessResults` itself errors.
- The director's pre-scheduling `ServiceUnavailable` rejections (no endpoint candidates located, screeners eliminating every candidate) carry the `rejected-no-endpoints` header.

Clients receive the same status codes as before; the response gains the header.

Known trade-offs, each requiring interface changes that #2428 rules out:

- A filter drain caused by permanent misconfiguration, for example a label filter no pod matches, also reports `rejected-saturated`. Attributing a drain to a specific filter requires carrying a reason through the `Filter` interface.
- When `ProcessResults` fails for a reason of its own while a profile also drained, for example the disagg handler in prefill-first mode with the decode profile missing from the configuration, the joined typed error decides the response: 429 with `rejected-saturated` where 500 would name the configuration fault. Causal attribution requires handlers to see profile errors (#1686).
- When custom profiles fail with different typed codes, the sorted join selects the code by profile-name order. In-tree profiles emit only `ResourceExhausted`, so no stock configuration observes the ordering.

Out of scope: the no-profile-selected untyped error and other untyped 429 sources such as a headerprofile no-match (#1686), and the admission-plugin denial classification.

Fixes #2428

**Test plan**:

- [x] A profile whose filter drains the candidate set surfaces a typed `ResourceExhausted` error with the `rejected-saturated` header from `Schedule`.
- [x] The director returns the typed code and drop-reason header from a joined scheduler error instead of the untyped-error fallback.
- [x] The empty-candidates and screener-drain rejections carry the `rejected-no-endpoints` header on their 503 responses.

**Release note**:

```release-note
Requests rejected before dispatch always carry the x-llm-d-request-dropped-reason response header: a 429 with reason rejected-saturated when scheduling filters drain all candidate endpoints, and a 503 with reason rejected-no-endpoints when no candidates exist or screeners eliminate them all.
```

### DIFF
--- a/pkg/common/error/error.go
+++ b/pkg/common/error/error.go
@@ -26,7 +26,7 @@ import (
 )
 
 // RequestDroppedReasonHeaderKey is the HTTP response header that communicates the specific
-// reason a request was dropped by flow control.
+// reason the EPP dropped a request.
 const RequestDroppedReasonHeaderKey = "x-llm-d-request-dropped-reason"
 
 // RequestDroppedReason is the reason a request was rejected before dispatch or evicted after dispatch.
--- a/pkg/epp/requestcontrol/director.go
+++ b/pkg/epp/requestcontrol/director.go
@@ -304,8 +304,9 @@ func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestCo
 	endpointCandidates := d.endpointCandidates.Locate(ctx, reqCtx.Request.Metadata)
 	if len(endpointCandidates) == 0 {
 		return reqCtx, errcommon.Error{
-			Code: errcommon.ServiceUnavailable,
-			Msg:  "failed to find endpoint candidates for serving the request",
+			Code:    errcommon.ServiceUnavailable,
+			Msg:     "failed to find endpoint candidates for serving the request",
+			Headers: map[string]string{errcommon.RequestDroppedReasonHeaderKey: string(errcommon.RequestDroppedReasonNoEndpoints)},
 		}
 	}
 
@@ -313,8 +314,9 @@ func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestCo
 	snapshotOfCandidatePods = d.runScreeners(ctx, reqCtx.SchedulingRequest, snapshotOfCandidatePods)
 	if len(snapshotOfCandidatePods) == 0 {
 		return reqCtx, errcommon.Error{
-			Code: errcommon.ServiceUnavailable,
-			Msg:  "screeners eliminated all endpoint candidates",
+			Code:    errcommon.ServiceUnavailable,
+			Msg:     "screeners eliminated all endpoint candidates",
+			Headers: map[string]string{errcommon.RequestDroppedReasonHeaderKey: string(errcommon.RequestDroppedReasonNoEndpoints)},
 		}
 	}
 	// Prepare per request data by running DataProducer plugins.
--- a/pkg/epp/requestcontrol/director_test.go
+++ b/pkg/epp/requestcontrol/director_test.go
@@ -414,11 +414,13 @@ func TestDirector_HandleRequest(t *testing.T) {
 		initialTargetModelName  string // Initial target model in the reqCtx.
 		parser                  fwkrh.Parser
 		wantErrCode             string                   // Expected errcommon code string
+		wantDroppedReason       string                   // If non-empty, expected x-llm-d-request-dropped-reason header on the error
 		wantReqCtx              *handlers.RequestContext // Fields to check in the returned RequestContext
 		targetModelName         string                   // Expected model name after target model resolution
 		admitRequestDenialError error                    // Expected denial error from admission plugin
 		dataProducerPlugin      *mockDataProducerPlugin
 		screener                *mockScreener
+		emptyEndpoints          bool // If true, the director locates no endpoint candidates.
 		preRequestPlugins       []*mockPreRequestPlugin
 		requestHeaderPlugin     *mockRequestHeaderPlugin
 		wantMutatedBody         map[string]any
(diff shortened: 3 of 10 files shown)


ahg-g:
looks good, pls fix the failed tests

LukeAVanDrie:
> looks good, pls fix the failed tests

done! that test actually shows that this change works properly e2e now

ahg-g (approved):

PR #2463, opened 2026-08-19
(WIP) feat: add compatibility-screener plugin and string metric label extraction

dagrayvid:
Add a new `compatibility-screener` plugin that groups endpoints by a metrics-derived compatibility value (e.g. NIXL config hash) and filters to one compatible group per request, preventing KV transfer failures during rolling upgrades of disaggregated P/D deployments. Also extends the custom metrics extractor to support extracting string label values from info-style Prometheus gauges (via new `labelName` config field) to enable this.

**What type of PR is this?**

/kind feature

Adds a `compatibility-screener` plugin that prevents KV transfer failures during rolling upgrades of disaggregated P/D deployments. The screener groups endpoints by a compatibility value (e.g. NIXL config hash scraped from a vLLM Prometheus metric), drops groups missing required roles, and picks one compatible group per request — so prefill and decode always land on pods that can exchange KV cache data.

To support this, the `core-metrics-extractor` is extended with a `labelName` field on `customMetrics` that extracts string label values from info-style gauges (stored as a new `StringMetricValue` attribute type). 

**Example configuration:**
```yaml
plugins:
- parameters:
    engineConfigs:
    - name: vllm
      customMetrics:
      - attributeKey: "vllm.io/nixl-compat-hash"
        metricSpec: "vllm:nixl_config_info"
        labelName: "compatibility_hash"
  type: core-metrics-extractor
- parameters:
    attributeKey: "vllm.io/nixl-compat-hash"
    headerName: "x-compat-hash"
    roleLabelKey: llm-d.ai/role
    requireRoles: [prefill, decode]
  type: compatibility-screener
```

**Test results:**

16,000-request benchmark during a rolling dtype change on a live P/D cluster (3 prefill + 3 decode, Qwen3-0.6B, NIXL `kv_both`):
- **Without screener:** 14,593 errors (91%)
- **With screener:** 0 errors

Depends on vLLM PR exposing `vllm:nixl_config_info` metric: https://github.com/vllm-project/vllm/pull/52999

Fixes #

```release-note
Add compatibility-screener plugin for filtering endpoints by NIXL config hash during rolling upgrades of disaggregated P/D deployments.
```

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -105,6 +105,7 @@ import (
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/requestattributereporter"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/requestheader/agentidentity"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/requestheader/outlenbucket"
+	compatibilityfilter "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/screener/compatibilityfilter"
 	disaggregatedsetrollout "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout"
 	testresponsereceived "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/test/responsereceived"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requesthandling/parsers/anthropic"
@@ -575,6 +576,7 @@ func (r *Runner) registerInTreePlugins() {
 	// request control screeners
 	// Alpha
 	fwkplugin.Register(disaggregatedsetrollout.PluginType, fwkplugin.StabilityAlpha, disaggregatedsetrollout.Factory)
+	fwkplugin.Register(compatibilityfilter.PluginType, fwkplugin.StabilityAlpha, compatibilityfilter.Factory)
 
 	// bylabel role filters
 	// Beta
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -103,6 +103,7 @@ import (
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/requestattributereporter"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/requestheader/agentidentity"
+	compatibilityfilter "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/screener/compatibilityfilter"
 	disaggregatedsetrollout "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout"
 	testresponsereceived "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/test/responsereceived"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requesthandling/parsers/anthropic"
@@ -567,6 +568,7 @@ func (r *Runner) registerInTreePlugins() {
 	// request control screeners
 	// Alpha
 	fwkplugin.Register(disaggregatedsetrollout.PluginType, fwkplugin.StabilityAlpha, disaggregatedsetrollout.Factory)
+	fwkplugin.Register(compatibilityfilter.PluginType, fwkplugin.StabilityAlpha, compatibilityfilter.Factory)
 
 	// bylabel role filters
 	// Beta
--- a/pkg/epp/framework/plugins/datalayer/attribute/metrics/data_types.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/metrics/data_types.go
@@ -37,6 +37,14 @@ func (v ScalarMetricValue) Clone() fwkdl.Cloneable {
 	return v
 }
 
+// StringMetricValue is a string endpoint attribute extracted from a Prometheus
+// metric label on an info-style gauge (e.g. a compatibility hash).
+type StringMetricValue string
+
+func (v StringMetricValue) Clone() fwkdl.Cloneable {
+	return v
+}
+
 // ScalarMetricDataKey returns the key under which the core metrics extractor
 // publishes the custom scalar metric configured as attributeKey. The data type
 // of a custom metric is chosen in configuration rather than in code, so the key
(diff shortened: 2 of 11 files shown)


ahg-g:
@hasB4K pls see this, related to the work you are doing with disaggSet rollout aware screening.

hasB4K:
Hello @dagrayvid,

Thanks for sharing this. We are already working on the same end-to-end problem across https://github.com/llm-d/llm-d-router/pull/2141 and its follow-up PRs #2570 (epp), #2571 (epp), #2301 (coordinator), and #2572 (coordinator).

The main component is the DisaggregatedSet rollout screener ([README.md on main](https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout#readme)), which largely covers this use case.

The core idea is:
- We use the `DisaggregatedSet` revision (https://github.com/kubernetes-sigs/lws/tree/main/keps/766-DisaggregatedSet) as a compatibility boundary shared by all roles. Prefill and decode Pods created from the same desired state receive the same revision. The revision is an aggregated hash on all the roles (prefill/decode) exposed at the pod level.
- A revision receives no traffic until it has at least one Ready Pod for every required role.
- During a rollout, `max-role` uses the most numerous role as the rollout-progress signal and routes requests proportionally between complete revisions.
- Once a revision is selected, every role serving that logical request is restricted to that revision.

For example:
```
                  Prefill   Decode   Traffic
Old revision A        2        6       60%
New revision B        1        4       40%
```

in that example, decode is the globally largest role, so its 6:4 distribution determines the 60:40 traffic split. A request assigned to revision B uses both prefill B and decode B. If B does not yet have both roles, it receives 0% traffic.

This also handles E/P/D. A logical request can create several encode requests (ie. one per image, in parallel for one request). Those "sub-requests" can reach different EPP replicas before any response header exists. As currently implemented in your PR, the compatibility screener makes an independent local random choice for each request, so different encode requests could select different hashes.

The rollout screener instead gives every parallel sub=request the same `x-llm-d-revision-decision-id`. During a rollout, the EPP replicas use an atomic GetOrSet through the shared syncer so that the first selected revision becomes authoritative for all those subrequests. Once a response contains x-llm-d-disagg-revision, the coordinator forwards it and subsequent phases use strict filtering. 

(Note: `x-llm-d-revision-decision-id` is 100% optional in a standard P/D approach, this issue only arise in E/P/D)

This limitation is not inherently caused by obtaining the key from metrics; it comes from making independent decisions without shared coordination. This could be achieved with metrics, but I'm pretty certain that it will be more complex.

We deliberately do not rely only on a NIXL compatibility hash because a rollout boundary can represent more than transport compatibility. For example:
- An older vLLM version may corrupt KV state because of a bug. Its NIXL hash could remain compatible with the patched version, but we still do not want KV state crossing between those revisions.
- A deployment may switch from NIXL to LMCache, Mooncake, or another disaggregated stack where a NIXL-specific hash is absent or irrelevant.
- Other behavioral or configuration changes may require isolation even when the transport format itself remains compatible (block_size change, adding spec dec, etc.)

The implementation is also not technically tied to DisaggregatedSet. Its configuration (https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout#configuration) accepts
arbitrary scope, revision/group, and role label keys. It can therefore use any stable Pod label representing a compatibility group:

```
revisionGating:
  revisionHeaderName: x-nixl-compat-hash
  revisionLabelKey: llm-d.ai/nixl-compat-hash
  roleLabelKey: llm-d.ai/role
  mode: max-role
  requiredRoles: [prefill, decode]
```

The advantage of DisaggregatedSet is that it already creates one coordinated revision across all roles and rolls those roles as one logical deployment.

My recommendation would therefore be one of these:

1. Use the DisaggregatedSet revision directly as the broader compatibility boundary. It will also solve the "we need to have a good P:D ratio", otherwise, if you use separate LWS, one prefill could rollout quicker than decode for the same "compatibility hash" - then you will have a spike of errors.
2. Expose the NIXL compatibility hash as a Pod label and configure the existing rollout screener to use it.

As a side note: The metric-derived hash could still be useful as a more broad compatibility signal and could allow compatible revisions to mix deliberately. I just think that the revision is more precise, and more safe, since it covers more use cases. If really using metrics would be the only way, and the revision podLabel (or a custom nixl config only one) can't be used (I don't see what could be the reason here however), we should think about re-using the DisaggregatedSet Rollout Screener logic, and I would advise to not create a new Screener that tries to solve exactly the same use case.

markmc:
> We deliberately do not rely only on a NIXL compatibility hash because a rollout boundary can represent more than transport compatibility. For example:
>
> * An older vLLM version may corrupt KV state because of a bug. Its NIXL hash could remain compatible with the patched version, but we still do not want KV state crossing between those revisions.

Agree. This is an important point. The internal vLLM compatibility hash should be seen as a safety net only, not a guarantee that a given pairing is desirable

dagrayvid:
Thank you @hasB4K for the detailed response. I agree that DisaggregatedSet is the right long-term solution. This PR was intended as a solution for platforms where DisaggregatedSet is not yet supported, where two separate Deployments are being used for P/D.

I will explore an alternative solution of using a higher level orchestrator (kserve) to add a Pod label representing compatibility / revision so that we can re-use the existing rollout screener and then follow-up here.

hasB4K:
Thanks @dagrayvid. One clarification: KServe can be the component that propagates this label, but the rollout screener does not require KServe specifically. It only needs a stable compatibility/revision label shared by the Prefill and Decode Pods.

For two Deployments managed by the same Helm chart, we could compute one aggregate revision from both role configurations and put the same value on both Pod templates, for example, it would probably look something like that:

```yaml
{{- $compat := dict "prefill" .Values.prefill "decode" .Values.decode "model" .Values.model -}}
{{- $revision := (toJson $compat | sha256sum | trunc 32) -}}

spec:
  template:
    metadata:
      labels:
        llm-d.ai/compatibility-revision: {{ $revision | quote }}
```

The label block would be used in both Deployments. The hashed inputs should include compatibility-relevant settings such as image digests, model revision, connector/backend, dtype, and arguments, while excluding things such as replica counts.

The one case Helm cannot solve by itself is if we specifically need vLLM's exact runtime-generated hash, because that value only exists after the process starts. That would need a small controller to read the value, patch the live Pod label, and keep the Pod NotReady until the label has been set... This is not so easy, and looks custom/brittle, but it's possible. I prefer the Helm approach especially since you are mentioning that the goal is to solve this (ie. a temp thing):

> This PR was intended as a solution for platforms where DisaggregatedSet is not yet supported, where two separate Deployments are being used for P/D.

Final but not least, DisaggregatedSet is in GA, and is used in prod already at Mistral. And it's quite stable, so I would recommend for you to try it 😉

dagrayvid:
Thanks @hasB4K, this makes sense and I appreciate the example for how this looks with Helm. Agreed kserve is not required for this. I am working on a PoC of how a label like this can be added to the KServe LLMInferenceService controller which is using Deployments for P/D with llm-d.

Issue #2470, opened 2026-08-19
perf(epp): hot-path allocation and logging hygiene in the core framework


LukeAVanDrie:
**What would you like to be added**:

A set of small, independent fixes for allocation and logging inefficiencies in the EPP core request path and datalayer. Each is local and behavior-preserving. Perf-motivated changes carry benchmark evidence (`go test -bench` + benchstat against the parent commit; the existing `BenchmarkSchedule`, `BenchmarkScope`, and `BenchmarkHandleResponseBody` cover the affected paths):

- [ ] Guard hot-path log and span construction in scheduler, director, and handlers (findings 1-7 below)
- [ ] Cache `Scope` allowed-key sets per plugin (finding 8)
- [ ] Guard scrape-path TRACE logging and remove the `labelsMatch` map build (findings 9-10)
- [ ] Logging and tooling consistency: dedupe the zap level encoder copied between `pkg/common/observability/logging` and `pkg/sidecar/proxy/options.go`; convert sidecar bare `V(4)`/`V(5)` calls to the named constants and guard the full-request-body `V(5)` logs; drop dead exported constants in the metrics extractor; bump the envtest pin in `Dockerfile.builder` (control plane 1.31 is EOL while kubectl is 1.35)

**Why is this needed**:

The findings are of the same kind: log and tracing work that allocates on every request or scrape even when the verbosity level or tracing is disabled, and static per-plugin data rebuilt on every invocation. The pattern is already documented in the code: `runScorerPlugins` guards its per-endpoint debug log with a comment noting the unguarded version "accounted for ~80% of total allocations per Scheduler.Schedule call" (`pkg/epp/scheduling/scheduler_profile.go`). The remaining call sites were never converted.

Findings, by area:

Scheduling framework (per request):

- `pkg/epp/scheduling/scheduler_profile.go` — unguarded `V(DEBUG)` logs box the full endpoint slice before and after the filter chain, before scoring, and around picking; unguarded `V(VERBOSE)` logs and repeated `TypedName()` calls run per plugin in each chain.
- `pkg/epp/scheduling/scheduler_profile.go` (`runPickerPlugin`) — `topScoredEndpoints` sorts the whole candidate set, with `ID.String()` in the comparator, only to set span attributes, and runs even when the span is non-recording (the default). `runScorerPlugins` in the same file already has the `span.IsRecording()` guard.
[...]

PR #2471, opened 2026-08-19
perf(epp): derive Scope allowed-key sets once at datalayer init

LukeAVanDrie:
**What type of PR is this?**

/kind cleanup

`datalayer.Scope` rebuilt the `allowedPut`/`allowedGet` maps from `Produces()`/`Consumes()` on every invocation. Scope runs for every filter, every scorer, and every DataProducer on every request, while the declarations are fixed at plugin construction.

The sets are now derived once at startup: the runner calls `datalayer.RegisterScopeSpecs(handle.GetAllPlugins())` next to `ValidateAndOrderDataDependencies`, after auto-created producers exist. `Scope` looks the spec up by the plugin's typed name, the same key `ValidateAndOrderDataDependencies` uses. A plugin that never went through registration is confined to nothing, with a one-time error log, so a wiring bug fails closed.

Benchmarks (`benchstat`, 10 interleaved runs per side, Apple M4 Pro):

```
                      │     main      │  this PR      vs base                │
BenchmarkScope sec/op
Scope/endpoints=10      533.4n ± 3%    318.2n ± 4%   -40.34% (p=0.000 n=10)
Scope/endpoints=100     1.733µ ± 4%    1.640µ ± 6%    -5.40% (p=0.018 n=10)

BenchmarkScope allocs/op
Scope/endpoints=10      10.000 ± 0%    7.000 ± 0%    -30.00% (p=0.000 n=10)
Scope/endpoints=100     10.000 ± 0%    7.000 ± 0%    -30.00% (p=0.000 n=10)
```

Part of #2470

```release-note
NONE
```

## Test plan

- [x] New unit tests: an unregistered plugin is confined to nothing (declared reads and writes both denied, without panicking); re-registering a typed name replaces its spec
- [x] Existing `./pkg/epp/datalayer/...`, `./pkg/epp/scheduling/...`, `./pkg/epp/requestcontrol/...`, disagg profile-handler, and `./cmd/epp/...` unit tests (tests that assert scoped data flow now register their plugins)
- [x] `make presubmit` stages (signed-commits-check trips on the pre-existing GitHub merge commit `843d2271`, which the DCO and signed-commit CI checks exempt)
- [x] `BenchmarkScope` before/after with benchstat

### DIFF
--- a/pkg/epp/datalayer/endpoint_scope.go
+++ b/pkg/epp/datalayer/endpoint_scope.go
@@ -189,6 +189,53 @@ func (e *ScopedEndpoint) Clone() fwkdl.AttributeMap {
 	return clone
 }
 
+// scopeSpec holds the allowed-key sets derived from a plugin's declarations.
+// The maps are built once per plugin and shared read-only by every ScopedEndpoint
+// wrapper across all subsequent invocations.
+type scopeSpec struct {
+	allowedPut map[fwkplugin.DataKey]struct{}
+	allowedGet map[fwkplugin.DataKey]struct{}
+}
+
+// scopeSpecs caches one scopeSpec per plugin instance, keyed by identity.
+// Produces() and Consumes() are source-level declarations fixed at plugin
+// construction, so the derived sets never change, while Scope runs for every
+// filter, scorer, and DataProducer on every request. Plugin instances live for
+// the process lifetime, so entries are never evicted.
+var scopeSpecs sync.Map // fwkplugin.Plugin -> *scopeSpec
+
+func scopeSpecFor(plugin fwkplugin.Plugin) *scopeSpec {
+	if cached, ok := scopeSpecs.Load(plugin); ok {
+		return cached.(*scopeSpec)
+	}
+
+	produces := map[fwkplugin.DataKey]any{}
+	if producer, ok := plugin.(fwkplugin.ProducerPlugin); ok {
+		produces = producer.Produces()
+	}
+	spec := &scopeSpec{
+		allowedPut: make(map[fwkplugin.DataKey]struct{}, len(produces)),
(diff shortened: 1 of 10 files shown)


pkg/epp/datalayer/endpoint_scope.go
```diff
@@ -189,36 +189,68 @@ func (e *ScopedEndpoint) Clone() fwkdl.AttributeMap {
...
+}
+
+// scopeSpecs caches one scopeSpec per plugin instance, keyed by identity.
+// Produces() and Consumes() are source-level declarations fixed at plugin
+// construction, so the derived sets never change, while Scope runs for every
+// filter, scorer, and DataProducer on every request. Plugin instances live for
+// the process lifetime, so entries are never evicted.
+var scopeSpecs sync.Map // fwkplugin.Plugin -> *scopeSpec
+
+func scopeSpecFor(plugin fwkplugin.Plugin) *scopeSpec {
```
ahg-g: This is an init logic, I recommend to call this explicitly from the datalayer init code instead of doing it on the fly

LukeAVanDrie: Done. The sets are now derived once at startup: the runner calls `datalayer.RegisterScopeSpecs(handle.GetAllPlugins())` next to `ValidateAndOrderDataDependencies`, after auto-created producers exist, and `Scope` looks the spec up by the plugin's typed name (the same key `ValidateAndOrderDataDependencies` uses).

A plugin that never went through registration is confined to nothing, with a one-time error log, so a wiring bug fails closed instead of silently widening access. This also drops the lazy `sync.Map` and the plugin-as-map-key requirement.

LukeAVanDrie: updated bench data in PR description too

ahg-g (approved):

PR #2472, opened 2026-08-19
perf(epp): guard hot-path log and span construction in scheduler, director, and handlers

LukeAVanDrie:
**What type of PR is this?**

/kind cleanup

Passing arguments to a disabled logger boxes them into a heap-allocated slice, and attribute construction for a non-recording span is wasted work. `runScorerPlugins` and `runResponseBodyPlugins` already guard against both, with comments noting the unguarded per-endpoint log line alone accounted for ~80% of allocations per `Scheduler.Schedule` call. The same pattern is applied to the remaining call sites:

- `scheduling/scheduler_profile.go`: guard the DEBUG logs that box the full endpoint slice in the filter and picker runners, cache `TypedName()` per plugin, and skip span attribute construction when the span is non-recording. `topScoredEndpoints` sorted the entire candidate set on every request to compute attributes for a span that is non-recording by default; the sort now runs only when the span is recording.
- `scheduling/scheduler.go`: guard the five VERBOSE logs per scheduling round (one boxes the whole profile-results map) and cache the profile handler's `TypedName()`.
- `requestcontrol/director.go`: convert the five unguarded plugin runners (`runPreRequestPlugins`, `runRequestHeaderProcessors`, `runScreeners`, `runAdmissionPlugins`, `runResponseHeaderPlugins`) to the guarded pattern `runResponseBodyPlugins` documents.
- `handlers/response.go`: guard the per-chunk DEBUG log in `HandleResponseBody`.
- `handlers/server.go`: guard the per-header TRACE log in the response-header loop and stop materializing every header value as a string when only two keys are inspected.

Log output at every verbosity level is unchanged. The change removes construction cost when the level or tracing is disabled.

Benchmarks (`benchstat`, count=10, builder container, logging and tracing disabled, which is the production default):

```
                         │    main     │  this PR    vs base                 │
BenchmarkSchedule sec/op
Schedule/endpoints=5-8     11.511µ ± 4%   9.130µ ± 3%  -20.69% (p=0.000 n=10)
Schedule/endpoints=25-8     25.24µ ± 1%   22.21µ ± 7%  -12.03% (p=0.000 n=9)
Schedule/endpoints=100-8    81.67µ ± 2%   75.07µ ± 2%   -8.08% (p=0.000 n=10)

BenchmarkSchedule allocs/op
Schedule/endpoints=5-8       205.0 ± 0%   142.0 ± 0%   -30.73% (p=0.000 n=10)
Schedule/endpoints=25-8      323.0 ± 6%   260.0 ± 8%   -19.50% (p=0.000 n=9)
Schedule/endpoints=100-8     678.0 ± 0%   615.0 ± 0%    -9.29% (p=0.000 n=10)

```

B/op for `BenchmarkSchedule` drops 18%/7%/3% at 5/25/100 endpoints. `BenchmarkHandleResponseBody` shows no significant change (sec/op p=0.118, B/op and allocs/op equal); its per-chunk runner was already guarded.

Part of #2470

```release-note
None
```

## Test plan

- [x] Existing unit tests: `./pkg/epp/scheduling/...`, `./pkg/epp/requestcontrol/...`, `./pkg/epp/handlers/...` (behavior-preserving change; no new tests)
- [x] `make presubmit`
- [x] `BenchmarkSchedule` before/after with benchstat; `BenchmarkHandleResponseBody` re-run on the rebased base to confirm no regression


pkg/epp/handlers/response.go
```diff
@@ -46,7 +46,9 @@ import (
 // body as a single "stream" event.
 func (s *StreamingServer) HandleResponseBody(ctx context.Context, reqCtx *RequestContext, responseBytes []byte, endOfStream bool) *RequestContext {
 	logger := log.FromContext(ctx)
-	logger.V(logutil.DEBUG).Info("HandleResponseBody is triggered", "len(responseBytes)", len(responseBytes), "endOfStream", endOfStream)
+	if debug := logger.V(logutil.DEBUG); debug.Enabled() {
```
ahg-g, line 51: pls add a comment why we do this so that in the future this doesn't get reverted.

ahg-g (commented):
it is really unfortunate that we have to do this, are we sure there is no flag in the logging library that prevents this overhead?

We should document somewhere in the code, perhaps on a couple of places that this PR changes, that this is intentional for performance reasons just so it doesn't get reverted.

LukeAVanDrie:
> it is really unfortunate that we have to do this, are we sure there is no flag in the logging library that prevents this overhead?

My understanding is that this is not an issue with the logging library itself but rather how Go evaluates function arguments. Since arguments are eagerly evaluated before the function is called, `logger.Info("msg", "len", len(responseBytes))` means Go will first execute `len(responseBytes)` as part of preparing the arguments for the function call.

It gets worse if the logging function is variadic and accepts arbitrary key-value pairs as it triggers interface boxing and slice allocations.

E.g., for `func (l *Logger) Info(msg string, keysAndValues ...any)` the Go compiler may not know that the `keysAndValues` slice isn't being stored anywhere so it will allocate the slice and the boxed primitives on the heap. To my knowledge, this is what is causing the perf degradation. 

The options at hand are:
1. guard pattern (this PR); enforced only in the most latency critical areas on the hot path
2. loggers with strongly typed field constructors (e.g., `zap`'s `zap.Int("len", len(responseBytes))`) (saves the interface boxing, not arg evaluation)

We also cannot create a generic helper function for the guard pattern as it would run into the exact same problems. E.g.,

```
func LogDebug(logger logr.Logger, msg string, args ...any) {
    if debug := logger.V(logutil.DEBUG); debug.Enabled() {
        debug.Info(msg, args...)
    }
}
```

still triggers the eager eval and boxing.


Possibly avoidable if we force lazy evaluation via a closure. This would save us the arg eval, but would still (somtimes) trigger a heap allocation. E.g.

```
func LogDebugLazy(logger logr.Logger, msg string, argsFn func() []any) {
    if debug := logger.V(logutil.DEBUG); debug.Enabled() {
        debug.Info(msg, argsFn()...)
    }
}
...
LogDebugLazy(logger, "HandleResponseBody is triggered", func() []any {
    return []any{"len", len(responseBytes)}
})
```

pkg/epp/handlers/response.go
LukeAVanDrie: Maybe something in AGENTS.md or similar as well? I don't think there's a good way to enforce this pattern through static analysis as it's pretty subjective when it should be applied and often requires some empirical evidence.

LukeAVanDrie: Added comments at both guard sites in handlers noting the pattern is intentional (a disabled logger still heap-allocates its variadic args). The scheduler and director runners carry the fuller rationale on `runScorerPlugins` and `runResponseBodyPlugins`. On documenting it more broadly: `AGENTS.md` already covers this under "Guard expensive log construction" in the Logging section, so no change needed there.

ahg-g (approved):

PR #2480, opened 2026-08-20
feat(coordinator) Add DP-rank pinning parity to the coordinator's kv-nixl connector

### LINKED ISSUE #2478: feat(coordinator) Add DP-rank pinning parity to the coordinator's kv-nixl connector
**Problem**
The sidecar's `nixlv2` KV connector (`pkg/sidecar/proxy/connector_nixlv2.go`) pins a disaggregated request's prefill and decode legs to the same vLLM data-parallel rank: it hashes the request ID into a rank and sets it as the `x-data-parallel-rank` header on both HTTP calls (`pkg/sidecar/proxy/dp_rank.go`). This guards against SO_REUSEPORT-sharing DP backends routing the two legs to different ranks.

The coordinator's equivalent connector, `kv-nixl` (`pkg/coordinator/connectors/kv/nixl.go`), has no DP awareness at all — no rank hashing, no header, and no dp_size configuration anywhere in the coordinator. There's no known failure from this today (plain NIXLv2 READ mode addresses the KV pull via the explicit remote_engine_id/remote_host/remote_port the prefill response returns, so it isn't exposed to the same rank-mismatch problem in practice), but the coordinator should carry the same defensive behavior as the sidecar so a DP>1 deployment behaves identically under either orchestration model.

**Proposal**
Add a `pipeline.dp_size` config key (default 1, disabled), merged into step params the same way `kv_connector/ec_connector` already are.
Add a small `dp_rank.go` to `pkg/coordinator/connectors/kv/` with the same hash-based rank picking as the sidecar (duplicated, not shared — scoped to the coordinator only); leave it to a separate PR for combining duplicated functionality.
`kv-nixl` gains an optional RankHeaderer interface (PrefillHeaders/DecodeHeaders); PrefillStep/DecodeStep set the x-data-parallel-rank header on the outgoing gateway calls when implemented. kv-sglang/kv-shared-storage are unaffected.
No `kv_transfer_params` body changes — this only affects the HTTP header, matching plain NIXLv2 READ-mode behavior in the sidecar.

**Scope**
`kv-nixl` only. `kv-sglang` and `kv-shared-storage` do not need this — their sidecar equivalents (sglang, shared-storage) never reference DP rank either.

roytman:
**What type of PR is this?**
/kind feature

The sidecar's nixlv2 KV connector pins a disaggregated request's prefill and decode legs to the same vLLM data-parallel rank via the x-data-parallel-rank header, guarding against SO_REUSEPORT-sharing DP backends routing the two legs to different ranks. The coordinator's equivalent connector, kv-nixl, had no DP awareness at all.

This adds the same behavior to the coordinator:

New pipeline.dp_size config option (default 1, disabled), merged into step params the same way kv_connector/ec_connector already are.
kv-nixl hashes the request ID into a deterministic rank and sets it as the x-data-parallel-rank header on both the prefill and decode requests, preferring a remote_dp_rank the prefill response echoes back when present. kv-sglang and kv-shared-storage are unaffected — their sidecar equivalents never reference DP rank either.
No kv_transfer_params body field changes; this is header-only, matching the sidecar's plain (non-MoRI-IO) NIXLv2 behavior.

Fixes #2478

```release-note
Add `pipeline.dp_size` coordinator config option for the `kv-nixl` connector: pins a request's prefill and decode calls to the same vLLM data-parallel rank via the `x-data-parallel-rank` header, matching the llm-d-router sidecar's `nixlv2` connector behavior.

```


pkg/coordinator/pipeline/builder/builder.go
```diff
@@ -31,7 +31,13 @@ import (
 // validatePipeline rejects configurations that cannot work before any step runs.
 // The tokens-in format (use_openai_format=false) sends token IDs that only the
 // render step produces, so it requires a render step in the pipeline.
+// pipeline.dp_size must be a positive integer: mergeConnectorDefaults treats
```
elevran, line 34: The "reject non-positive pipeline.dp_size before building steps" commit adds test cases expecting this to reject DPSize <= 0, but the function body wasn't updated - CI confirms both new subtests fail (`validatePipeline() error = <nil>, wantErr true`). Needs a check like:
```go
if p.DPSize <= 0 {
    return fmt.Errorf("pipeline.dp_size must be a positive integer, got %d", p.DPSize)
}
```

elevran (commented):
Nice parity work with the sidecar's nixlv2 connector, and good test coverage overall. CI's unit-tests job is currently red though - the third commit's validatePipeline change was never actually added to the function, only to its tests, so the two new dp_size validation subtests fail. Once that's fixed and CI is green this looks good to me.

pkg/coordinator/pipeline/builder/builder.go
roytman: Thank you @elevran. I think you checked the previous version; that was fixed. 
All CI tests are "green," and the required test is there.

roytman:
`x-data-parallel-rank header` is used by `data-parallel-profile-handler` as a workaround for a bug in istio. It is not relevant for Istio >= 1.28.1. Therefore, `data-parallel-profile-handler` was deprecated.
I have validated that the coordinator supports `data-parallel` with `simple-profile-handler` and DP-aware routing.

Issue #2447, opened 2026-08-18
Accurate Cached Token Usage Metrics in P/D Setup


robertgshaw2-redhat:
**What would you like to be added**:
- when vllm is launched with  `--enable-prompt-tokens-details`, it will return how many tokens are cached in the openai responses (https://github.com/vllm-project/vllm/blob/8f4a7f45c53ab52b17023d3ca804e477daa36a23/vllm/entrypoints/openai/chat_completion/serving.py#L88)
- in a p/d setup, since the sidecar sends a fresh request to the decode worker and all tokens are cached in the decode worker, it will report 100% cached rate
- what we need to do is instead report how many tokens were cached on the prefiller

**Why is this needed**:
- other systems (e.g. someone building a maas on llm-d) can use this accurate info for billing

**Implementation Ideas**:
- P instance returns cached tokens in kv_xfer_params, then vllm decode instance uses this in the reporting rather than how many kvs were cached locally
- sidecar modifies the streamed responses from D with this info

elevran:
I think we need to work out the mechanics of where this belongs - in the model server layer, mediated via kv_xfer_params by sidecar or coordinator, or in the upper layer such as the sidecar.
There are efficiency and correctness concerns. Whatever we end up doing needs to work for other inference engines, such as sglang.
My preference is for each inference server to report status of what it did (ie P reports what was cached, consider p2p as well, and D reports 100% cached tokens). The higher layers are responsible for aggregagting and presenting the the information. This should be weighed against the complexity (e.g., needed for every API endpoint) and efficiency (e.g., sidecar needs to process all of D's output to rewrite cached usage)

PR #2481, opened 2026-08-20
perf(sidecar): skip usage rewrite parsing for frames without usage

sudoalok:
**What type of PR is this?**

/kind cleanup

replaceCachedTokensSSELine unmarshals every sse frame looking for usage, but only the last frame in a stream has usage in it. so a 500 token response does ~500 full json.Unmarshal calls to end up rewriting one field once.

put a bytes.Contains check for "usage" in front of the parse. frames without it already came out unchanged, they just paid for the unmarshal first, so the output is the same. if the model writes the word usage in its own text the check passes and the parse runs and finds nothing, same as now.

numbers from the benchmarks in this PR:

content frame: 3689 ns/op 1816 B/op 40 allocs/op -> 224 ns/op 0 B/op 0 allocs
usage frame: unchanged

added benchmarks for both frame shapes to the existing test file. existing tests pass, race clean, vet and gofmt clean.

the openai streaming parser already does the same strings.Contains(content, "usage") check before unmarshalling, and its the same idea as the guarded log sites in #2470. came across this while looking at #2447 but its unrelated to that discussion, this path already runs today on the nixl connector.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/sidecar/proxy/cached_tokens_usage_rewriter_test.go
+++ b/pkg/sidecar/proxy/cached_tokens_usage_rewriter_test.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strings"
+	"testing"
 
 	. "github.com/onsi/ginkgo/v2" // nolint:revive
 	. "github.com/onsi/gomega"    // nolint:revive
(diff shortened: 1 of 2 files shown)


pkg/sidecar/proxy/cached_tokens_usage_rewriter_test.go
```diff
@@ -215,3 +216,24 @@ var _ = Describe("Cached token usage rewriter", func() {
 		Expect(recorder.Body.String()).To(ContainSubstring(`"cached_tokens":7`))
 	})
 })
```
roytman, line 230: The two new benchmarks exercise the no-usage guard's performance but don't assert correctness. Every existing `It` case in this file that reaches `replaceCachedTokensJSON` has `usage` in its body, and the one case that doesn't ("should preserve non-JSON streamed data lines") uses non-JSON, which isn't valid JSON either. 
So there's no test covering the actual hot path this PR optimizes: a valid-JSON streaming delta with no usage field. If the guard's key or quoting were ever wrong, no spec would catch it, only a benchmark number silently changing.

Could you add something like:

```go
It("should preserve streamed content chunks without usage", func() {
    body := []byte(`data: {"choices":[{"delta":{"content":" the"}}]}` + "\n\ndata: [DONE]\n")
    updated := replaceCachedTokens(body, 7)
    Expect(updated).To(Equal(body))
})
```

sudoalok: added, plus one for a chunk with the word usage inside the content text.

small correction though, a wrong key is already covered. i changed usageKey to "usaje" and several existing specs fail, since a too narrow guard skips real usage frames. whats uncovered is the guard being removed or too wide, everything passes there and only the benchmark moves.

so these two pin the no-usage path as byte for byte unchanged, which nothing did before.

elevran:
@sudoalok please resolve conflict

pkg/sidecar/proxy/cached_tokens_usage_rewriter_test.go
```diff
@@ -262,4 +263,37 @@ var _ = Describe("Cached token usage rewriter", func() {
...
 		Expect(recorder.Body.String()).To(ContainSubstring(`"cached_tokens":7`))
 	})
+
+	It("should preserve streamed content chunks without usage", func() {
+		body := []byte(`data: {"choices":[{"delta":{"content":" the"}}]}` + "\n\ndata: [DONE]\n")
+		Expect(replaceCachedTokens(body, 7)).To(Equal(body))
+	})
+
+	It("should preserve streamed content chunks that mention usage in the content", func() {
+		// The guard matches on the JSON key, but content is free text: a chunk that
```
elevran, line 273: The test was supposed to prove that when a content frame's text happens to mention the word "usage" (so the cheap Contains check matches and parsing proceeds), the rewriter still produces correct output and leaves the frame untouched.

The test builds its SSE frame from a Go raw string containing `\"usage\"` with escaped quotes (`\"`), but `bytes.Contains` is checking for the exact 7-byte sequence "usage", with unescaped quotes.

So the guard's `Contains` check returns false for this test's input, same as the "frame without usage" test right above it. The test passes, but for the wrong reason: it's not exercising "guard matches but parse finds nothing," it's exercising "guard doesn't match" again, just with different-looking input.

To actually hit the intended branch, the content value needs to produce an unescaped "usage" substring in the serialized frame.

elevran (commented):
Nice, focused perf fix, and the benchmark numbers back up the claim. The approach aligned with the existing `bytes.Contains` guard used in the openai and anthropic streaming parsers.

One test issue: the "mention usage in the content" test doesn't actually exercise the case it's meant to (see inlined comment) below.

elevran (approved):

Issue #2485, opened 2026-08-20
[Flow Control] Response-feedback contract for external dispatchers (tracking)


LukeAVanDrie:
**What would you like to be added**:

Router-side emission of the response-feedback contract consumed by external dispatchers, proposed in llm-d/llm-d-async#379 with a reference consumer in llm-d/llm-d-async#382. Each field is independently useful and ships without coordination with the consumer. The normative contract text for each field lives in its own issue; each field is ratified by accepting that issue.

- [ ] `Retry-After` on capacity rejections: #2487
- [ ] `x-llm-d-flow-queue-duration-ms` response header: #2486
- [ ] `x-llm-d-flow-band-headroom-requests` response header: #2488 
- [ ] Deadline header to per-request queue-wait budget: #2316 (owned separately)

Joint work with `llm-d-async`, to be filed when ratification starts:

- [ ] Priority taxonomy: the (tier, classification) to priority-integer mapping and its carrier
- [ ] Conformance test asserting the objective-to-priority mapping preserves the dispatcher's lane order

**Why is this needed**:

The contract spans two repositories and several weeks of independently mergeable changes. A single tracking issue records completion state for maintainers of both projects; the design rationale stays in llm-d/llm-d-async#379 and the per-field contracts stay in the child issues.

LukeAVanDrie:
cc: @jtechapps, @shimib, @RishabhSaini, @BenjaminBraunDev, @rahulgurnani

PR #2456, opened 2026-08-19
refactor(flowcontrol): read capacity counters where they live

LukeAVanDrie:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

First of three PRs for #2169, which merges the flow controller and its processor into a single engine. It moves the capacity counters onto the structures that own them so the admission path reads them directly instead of building a registry-wide snapshot per request.

Costs removed. `hasCapacity` called `registry.Stats()` once per request, which takes the registry read lock, ranges a `sync.Map`, and allocates a fresh `AggregateStats` map. `enqueue` paid a second registry read lock per request for a `PriorityBandAccessor` existence check that the preceding `ManagedQueue` lookup already covers. Every queue mutation paid two `sync.Map` lookups to route its stats delta through `propagateStatsDelta`.

Changes:

- `FlowRegistryDataPlane` gains `CapacitySnapshot(priority)`: a lock-free, allocation-free read of the band and registry occupancy counters against their capacity limits. Capacities can be read without a lock because band configuration and the global limits are immutable once published. The comment on `priorityBand.config` claimed it is "updated during dynamic scaling events (updateConfig)"; no such mechanism exists in the package. This was vestigial from when we had sharding / data parallelism.
- Band occupancy counters move onto `priorityBand`, and the registry totals use the same `occupancyStats` type. `managedQueue` applies its measured deltas through pointers captured at construction, replacing the `propagateStatsDeltaFunc` callback and the parallel `perPriorityBandStats` map. The `onActiveTransition` callback that maintains the active-queue index (#2120) is untouched.
- `enqueue` drops the redundant `PriorityBandAccessor` check. `hasCapacity` returns the snapshot lookup error and `enqueue` maps it to `RejectedOther`; a band-missing race used to surface as a capacity rejection with all-zero stats. Capacity-rejection logs now include the band occupancy fields. Under the default config the band limit is the one that binds, since the global limits default to unlimited.
- `contracts.FlowRegistryObserver` is removed together with `contracts.AggregateStats` and `contracts.PriorityBandStats`; the only production consumer was `hasCapacity`, and nothing outside the flow control packages references the removed API. A concrete `Stats()` remains on `*FlowRegistry` for tests and debugging, returning a `Global` dimension and a per-band map built on `contracts.CapacityDimension`.

Admission still runs on the single processor goroutine, so the check-then-add sequence remains atomic. Admission semantics are unchanged, and the PR is independently revertible ahead of the engine merge.

[...]

PR #2489, opened 2026-08-20
feat(flowcontrol): attach Retry-After hint to
  capacity rejections

### LINKED ISSUE #2487: Emit Retry-After on capacity rejections

Flow control rejects a request on queue capacity with HTTP 429 and `x-llm-d-request-dropped-reason: rejected-saturated`. Add a `Retry-After` header to that response, carrying an advisory projection of when the rejecting scope next has a free slot.

Proposed contract:

- Emitted only on capacity rejections (429, `rejected-saturated`). Queue-TTL 429s carry no hint: the request's latency budget expired, and a drain estimate carries no information about whether a retry would fit the next budget. 503s and post-dispatch eviction 429s carry no hint.
- The value is RFC 9110 delta-seconds: a whole number of seconds, rounded up from the projection, capped at 30. The cap bounds the cost of a wrong estimate; an overestimate keeps clients away from capacity that has already recovered, while an underestimate costs one re-rejection.
- The drain rate is measured per scope as dispatch completions per second: the dispatch loop counts completions per priority band and registry-wide, and folds each count into an exponentially weighted moving average about once per second.
- The projection is per priority band. A rejection projects from the scope whose capacity check failed; when the band and the global limit both failed, admission needs room in both, so the longer projection is the binding one. A band's measured rate reflects strict-priority dispatch on its own: a band starved by higher-priority traffic drains slowly and projects a long wait, and a higher band's projection is unaffected by backlog below it.
- Fairness policies do not enter the projection. Capacity limits are per band and global, with no per-flow capacity, so a dispatch from any flow in the band frees a slot that a retry from any flow can take. Fairness selects which queued flow dispatches next, which shapes time in queue after admission; Retry-After projects time to admission.
- The header is omitted when the projection is under one second, because a whole-second floor would over-throttle a band that drains in milliseconds, and omitted while the estimator has no measurement. A client that receives no header applies its own retry policy.
- The value is advisory, and each endpoint-picker replica projects from its own queues. Clients are expected to bound how long they honor the hint; the `aimd` gate clamps it with its `max_retry_after` configuration.

Model:

A capacity rejection means a queue is at its limit: occupancy Q equals capacity C for the band, the registry-wide total, or both. Queued requests leave for the pool at the scope's drain rate r, so occupancy falls toward a watermark W and the projected wait is:

```
Retry-After = (Q - W) / r
```

The emission sets W = C - 1, one free slot, because the rejected client needs one admission opportunity and pacing beyond one request is the client's decision. With Q = C at rejection, the projection reduces to 1 / r.

```
occupancy
    C |##########  full: request rejected
  C-1 |- - - - -\- - - - -  watermark W: one free slot
      |          \
      |           \  drains at slope -r
      '-----------+------------------ time
                  ^
                  Retry-After = 1/r
```

llm-d-async#379 ("Stock, flow, and the integral between them") derives the same picture from queue dynamics, with occupancy as the accumulated difference between arrivals and service. The emission reads occupancy and capacity directly from the registry's counters and estimates r from observed dispatches; no integration is involved.

The projection ignores the arrival rate. While demand exceeds drain, slots open at r regardless of how fast competing arrivals re-take them; a projection based on net queue decline would answer when the queue becomes empty, and the client only needs to know when it can next be admitted. Competition for the opening slot is why the value is advisory rather than a reservation.

Open question for review: the emission policy for a measured drain rate of zero when the band is non-empty. After an idle period the rate estimate decays to zero, so a burst that fills the queue draws rejections whose hint sits at the cap even when the pool drains quickly. Options: emit the cap immediately, emit the cap only after several consecutive zero-drain estimator windows, or omit the header on zero drain.

**Why is this needed**:

The drop-reason header identifies the rejection cause and carries no estimate of when capacity returns, so each client falls back to its own backoff heuristic. The router can project the wait from dispatch rates it already observes. The async processor's feedback-based dispatch gate (proposal llm-d/llm-d-async#379, implementation llm-d/llm-d-async#382) consumes `Retry-After` on capacity rejections as a bounded window-close instruction with slow-start reopen; the proposal's contract table records the field as not yet emitted by the router. Because `Retry-After` is standard HTTP, clients other than the processor can use it without llm-d-specific parsing.

Related: #2486 and  #2488 cover the advisory response headers from the same contract. #2316 covers the deadline-to-TTL mapping, a separate part of the same contract.

LukeAVanDrie:
**What type of PR is this?**

/kind feature

Flow control returns a capacity rejection as HTTP 429 with `x-llm-d-request-dropped-reason: rejected-saturated`. The response identifies the rejection cause and carries no estimate of when capacity returns. This PR adds a `Retry-After` header: an advisory projection of when the rejecting scope next has a free slot.

- The processor tracks dispatch completion rates per priority band and globally: counters incremented at dispatch finalization, folded into an EWMA about once per second by the dispatch cycle. Both sites run on the single-writer Run goroutine, so the estimator needs no synchronization.
- On a capacity rejection, the projection is `1 / rate` for the scope whose capacity check tripped (the band, the global limit, or the longer of the two). The current window's instantaneous rate raises a decayed EWMA, so a burst that resumes dispatching after an idle stretch corrects the estimate without waiting for the next fold.
- Emission policy: RFC 9110 delta-seconds, rounded up, capped at 30 seconds. The header is omitted when the projection is under one second, because a whole-second floor would over-throttle a fast-draining scope, and when no tripped scope has a measured nonzero rate (startup, or measured-zero drain). An absent header means no estimate; clients apply their own retry policy.
- The hint travels inside the rejection error as `types.QueueCapacityError`, which unwraps to `ErrQueueAtCapacity`; the `EnqueueAndWait` signature and existing sentinel matching are unchanged. The admission translation reads it with `errors.As` and adds the header beside the drop reason.
- Scope: capacity rejections only. A queue-TTL 429 reports an exhausted latency budget, and a drain estimate carries no information about whether a retry would fit the next budget; TTL 429s, 503s, and post-dispatch eviction 429s carry no Retry-After.

The value is advisory and per-replica. The async processor's feedback-based dispatch gate (llm-d/llm-d-async#379, llm-d/llm-d-async#382) consumes it as a bounded window-close instruction and clamps it independently.

The contract, including the model behind the projection and the open question on zero-drain emission, is specified in #2487. This PR rebases onto #2456 when that lands (`hasCapacity` returns `contracts.CapacitySnapshot` there; the adaptation is mechanical).

Fixes #2487

Part of #2485.

**Release note**

```release-note
Flow control capacity rejections (HTTP 429 with reason `rejected-saturated`) now carry a `Retry-After` header projecting when the rejecting scope next has a free slot, derived from observed dispatch rates. The header is omitted when no estimate is available or the projected wait is under one second, and is capped at 30 seconds.
```

Test plan (functionality verified by new tests):
- [x] Drain estimator: warmup and measured-zero omission, one-slot projection with round-up, sub-second omission, 30s cap, burst correction of a decayed rate, global-scope and both-scopes-tripped projections
- [x] Processor integration: cold-start capacity rejection carries the typed cause with no hint; rejection after observed drain carries the projected hint
- [x] Outcome translation: hint renders as `retry-after` delta-seconds beside the drop reason; absent on zero hint and on TTL-expiry 429s
- [x] Typed error: `errors.Is` compatibility with `ErrQueueAtCapacity` and `errors.As` round-trip

### DIFF
--- a/pkg/common/error/error.go
+++ b/pkg/common/error/error.go
@@ -29,6 +29,10 @@ import (
 // reason a request was dropped by flow control.
 const RequestDroppedReasonHeaderKey = "x-llm-d-request-dropped-reason"
 
+// RetryAfterHeaderKey is the HTTP response header carrying the advisory retry delay on capacity
+// rejections, in RFC 9110 delta-seconds.
+const RetryAfterHeaderKey = "retry-after"
+
 // RequestDroppedReason is the reason a request was rejected before dispatch or evicted after dispatch.
 type RequestDroppedReason string
 
--- a/pkg/epp/flowcontrol/controller/internal/drain.go
+++ b/pkg/epp/flowcontrol/controller/internal/drain.go
@@ -1,0 +1,150 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package internal
+
+import (
+	"math"
+	"time"
+
+	"github.com/llm-d/llm-d-router/pkg/epp/flowcontrol/contracts"
+)
+
+const (
+	// drainFoldInterval is how often accumulated dispatch counts are folded into the rate EWMAs.
+	drainFoldInterval = time.Second
+
+	// drainEWMAAlpha weights the most recent fold window against the accumulated rate.
+	drainEWMAAlpha = 0.3
+
+	// retryAfterHintCap bounds the emitted hint. The value is advisory (RFC 9110) and consumers bound it independently;
+	// the cap keeps a decayed estimate from projecting arbitrarily far.
+	retryAfterHintCap = 30 * time.Second
+)
+
+// drainEstimator tracks dispatch completion rates per priority band and globally, and projects the wait until a slot
+// frees in a rejecting scope. It is owned by the Processor and accessed only from the Run goroutine, so it needs no
+// synchronization.
+type drainEstimator struct {
+	perBand  map[int]*drainRate
+	global   drainRate
+	lastFold time.Time
+}
+
+// drainRate is one scope's dispatch-rate state: the EWMA over completed fold windows and the count accumulated in the
+// current window.
+type drainRate struct {
+	count  uint64
+	rate   float64 // dispatches per second
+	folded bool
+}
+
+func newDrainEstimator(now time.Time) *drainEstimator {
+	return &drainEstimator{perBand: make(map[int]*drainRate), lastFold: now}
+}
+
+// recordDispatch counts one dispatch completion against the band and the global scope.
+func (e *drainEstimator) recordDispatch(priority int) {
+	b := e.perBand[priority]
+	if b == nil {
+		b = &drainRate{}
+		e.perBand[priority] = b
+	}
+	b.count++
+	e.global.count++
+}
+
+// maybeFold folds accumulated counts into the rate EWMAs once per drainFoldInterval.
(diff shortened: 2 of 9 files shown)


ahg-g:
@LukeAVanDrie do we still need this?

LukeAVanDrie:
@ahg-g going to close the PR but leave the issue open. I am handing this off, so there are some design discussions to be had still.

PR #2490, opened 2026-08-20
feat(epp): add x-llm-d-flow-queue-duration-ms response header

### LINKED ISSUE #2486: Advisory queue-duration response header

A response header, `x-llm-d-flow-queue-duration-ms`, on responses whose requests passed through flow control admission: the time the request spent in flow control between submission and dispatch, as integer milliseconds.

Proposed contract:

- Stamped at the ext-proc response-headers phase, before any streamed body, so it reaches the client ahead of the first token.
- A value of `0` is emitted and means the request dispatched without measurable queueing. An absent header means flow control did not process the request.
- Measured at the admission boundary (wall time around the flow controller's enqueue-and-wait call), so it covers queue residency plus dispatch handoff.
- The measurement is per-replica.

The header name is provisional in the llm-d-async proposal; maintainer acceptance ratifies it for the cross-project contract.

**Why is this needed**:

Queue time rises before rejections begin, so it is the earliest congestion signal available to a client. In the async processor's feedback-based dispatch gate (proposal llm-d/llm-d-async#379, implementation llm-d/llm-d-async#382), an accepted response that queued past a configured target holds the dispatch window rather than growing it, so the gate regulates before the router starts rejecting. The realized queue duration also tightens the send-time guard the gate applies when ordering evidence. The proposal's contract table records the name as provisional pending ratification in llm-d-router.

Related: #2487  and #2488. #2316 covers the deadline-to-TTL mapping, a separate part of the same contract.

jtechapps:
**What type of PR is this?**

/kind feature

Emits `x-llm-d-flow-queue-duration-ms` on responses whose requests passed through flow control admission, carrying the wall-clock time the request spent in `EnqueueAndWait` as integer milliseconds.

Queue time rises before rejections begin, so it is the earliest congestion signal available to a client. An external dispatcher can use it to regulate its dispatch window before the router starts shedding load.

Contract (per #2486):
- Measured at the admission boundary (wall time around the flow controller's enqueue-and-wait call), covering queue residency plus dispatch handoff. Per-replica.
- Stamped at the ext-proc response-headers phase, ahead of the first token.
- A value of `0` means the request dispatched without measurable queueing. An absent header means flow control did not process the request (e.g. the legacy admission path).

Fixes #2486

**Test plan**:

- [x] `x-llm-d-flow-queue-duration-ms` is emitted with the elapsed milliseconds when a request is admitted through flow control, `0` for a sub-millisecond dispatch, and omitted when flow control did not process the request.
- [x] The admission boundary stamps the queue duration on both dispatch and rejection; the legacy admission path leaves it unstamped.

**Release note**:
```release-note
Responses admitted through flow control now carry an x-llm-d-flow-queue-duration-ms header reporting the time the request spent in flow control admission, as integer milliseconds. A value of 0 means the request dispatched without measurable queueing; the header is absent when flow control did not process the request.
```

### DIFF
--- a/pkg/epp/requestcontrol/admission.go
+++ b/pkg/epp/requestcontrol/admission.go
@@ -175,11 +175,16 @@ func (fcac *FlowControlAdmissionController) Admit(
 	}
 
 	// Measure at the admission boundary: wall time around enqueue-and-wait covers queue residency plus
-	// dispatch handoff. Stamped on reqCtx for emission as the FlowQueueDuration response header.
+	// dispatch handoff. Stamped on reqCtx for emission as the FlowQueueDuration response header, only on
+	// dispatch: rejections short-circuit into an immediate error response that never reads these fields,
+	// and their durations carry no signal (a capacity rejection is ~0, a TTL eviction is the configured
+	// TTL, a cancellation is the client's disconnect time).
 	start := time.Now()
 	outcome, err := fcac.flowController.EnqueueAndWait(ctx, fcReq)
-	reqCtx.FlowControlQueueDuration = time.Since(start)
-	reqCtx.FlowControlAdmitted = true
+	if outcome == types.QueueOutcomeDispatched {
+		reqCtx.FlowControlQueueDuration = time.Since(start)
+		reqCtx.FlowControlAdmitted = true
+	}
 	logger.V(logutil.DEBUG).Info("Flow control outcome",
 		"requestID", reqCtx.SchedulingRequest.RequestID, "outcome", outcome, "error", err)
 	// A TTL expiry signals backpressure (429) when serving capacity exists, but genuine unavailability (503) when
--- a/pkg/epp/requestcontrol/admission.go
+++ b/pkg/epp/requestcontrol/admission.go
@@ -174,7 +174,12 @@ func (fcac *FlowControlAdmissionController) Admit(
 		modelName:         reqCtx.IncomingModelName,
 	}
 
+	// Measure at the admission boundary: wall time around enqueue-and-wait covers queue residency plus
+	// dispatch handoff. Stamped on reqCtx for emission as the FlowQueueDuration response header.
+	start := time.Now()
 	outcome, err := fcac.flowController.EnqueueAndWait(ctx, fcReq)
+	reqCtx.FlowControlQueueDuration = time.Since(start)
+	reqCtx.FlowControlAdmitted = true
 	logger.V(logutil.DEBUG).Info("Flow control outcome",
 		"requestID", reqCtx.SchedulingRequest.RequestID, "outcome", outcome, "error", err)
 	// A TTL expiry signals backpressure (429) when serving capacity exists, but genuine unavailability (503) when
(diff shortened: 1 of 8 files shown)


LukeAVanDrie (commented):
`pkg/epp/util/request/headers.go:44` maintains the set of EPP-injected headers that must be stripped from user and backend traffic to prevent ambiguity. The fix is one line (add `metadata.FlowQueueDurationHeaderKey` via `lowerHeaderNames`) plus a `headers_test.go` case.

pkg/epp/requestcontrol/admission_test.go
```diff
@@ -348,6 +353,49 @@ func TestFlowControlAdmissionController_Admit(t *testing.T) {
...
+
+	t.Run("flow control stamps duration on rejection", func(t *testing.T) {
+		t.Parallel()
+		reqCtx := newReqCtx()
+		fc := &mockFlowController{outcome: fctypes.QueueOutcomeRejectedCapacity}
+		ac := NewFlowControlAdmissionController(fc, "pool", &mocks.MockEndpointCandidates{})
+
+		require.Error(t, ac.Admit(ctx, reqCtx, 0))
+		assert.True(t, reqCtx.FlowControlAdmitted)
+		assert.GreaterOrEqual(t, reqCtx.FlowControlQueueDuration, time.Duration(0))
```
LukeAVanDrie: Nit: this is always true since `time.Since` can't go negative. Either give the rejection mock a delay too and assert against it, or drop the assertion and let `FlowControlAdmitted` carry that subtest.

pkg/epp/requestcontrol/admission.go
```diff
@@ -174,7 +174,12 @@ func (fcac *FlowControlAdmissionController) Admit(
 		modelName:         reqCtx.IncomingModelName,
 	}
 
+	// Measure at the admission boundary: wall time around enqueue-and-wait covers queue residency plus
+	// dispatch handoff. Stamped on reqCtx for emission as the FlowQueueDuration response header.
+	start := time.Now()
 	outcome, err := fcac.flowController.EnqueueAndWait(ctx, fcReq)
+	reqCtx.FlowControlQueueDuration = time.Since(start)
```
LukeAVanDrie: nit: The unconditional stamp after EnqueueAndWait writes state no reader can observe on rejection paths. Rejections short-circuit into an immediate error  response and never reach `generateResponseHeaders`.

The values would also carry no information there as a capacity rejection happens at the submission boundary (duration ~0), a TTL eviction's duration is the configured TTL, and a cancellation's duration is the client's own disconnect time. 

The consumer contract ([llm-d/llm-d-async#379](https://github.com/llm-d/llm-d-async/issues/379)) reads queue duration only on accepted responses; rejections are handled by the dropped-reason header and Retry-After (#2487).

Consider stamping FlowControlAdmitted and FlowControlQueueDuration only on the dispatch (nil-error) path. That also makes FlowControlAdmitted mean what its name says (today it is set true on rejection too) and the rejection subtest flips to asserting the fields stay unset.

Setting the value redundantly is benign though, so not blocking on this.

LukeAVanDrie (changes requested):
Thanks @jtechapps! This implementation LGTM, just blocking on the header sanitization piece; the other comments are non-blocking.

pkg/epp/requestcontrol/admission_test.go
jtechapps: Done

pkg/epp/requestcontrol/admission.go
jtechapps: Fixed. Note if we support in-queue evictions from other requests in the future we may want to add the header to rejection path.

LukeAVanDrie (approved):
Thanks @jtechapps; lgtm!

ahg-g (approved):

Issue #2475, opened 2026-08-20
[Flow Control] Per-stage stale endpoint accounting for stage-aware saturation


loicmarchal:
`flow_control_stale_endpoints` carries no `stage` label and is written on every `Saturation()` call. With stage-aware saturation gating (#1186), the detector is called once per non-empty stage partition per dispatch cycle, so the gauge reflects only the most recently evaluated stage. A reading of 0 does not rule out stale metrics in another stage.

This matters for diagnosing scrape outages: if the prefill fleet loses metrics collection, the detector scores those endpoints as 1.0 (fail-closed) and the effective saturation pins at 1.0, halting all dispatch. But the stale endpoints gauge may read 0 from the subsequent decode evaluation, causing operators to misdiagnose the stall as genuine overload.

The fix belongs in a stage-aware detector implementation rather than in the processor, since the detector owns pod-level metrics and staleness tracking. The detector should partition endpoints internally and emit per-stage stale counts.

PR #2492, opened 2026-08-21
feat(flowcontrol): make stale-metrics scoring configurable

### LINKED ISSUE #2408: flowcontrol: discuss configurable handling for stale endpoint metrics
## What happened

The utilization saturation detector treats an endpoint with missing or stale metrics as fully saturated (`1.0`). This is an intentional fail-closed behavior: when the scheduler cannot observe an endpoint reliably, it stops dispatching rather than risk sending traffic to an overloaded backend.

This makes sense as the default safety behavior. However, it can have a large availability impact for small or single-endpoint InferencePools.

In a single-endpoint pool, a stale metrics sample effectively makes the whole pool look saturated:

```text
stale metrics
  -> endpoint saturation = 1.0
  -> pool saturation = 1.0
  -> saturation >= usage limit
  -> flow-control applies HoL blocking
```

In our deployment we observed the following:

- Direct requests to the runtime service completed in about 40ms.
- Requests through the EPP/IGW path consistently took about 9.7s.
- `llm_d_epp_flow_control_pool_saturation` was observed at `1.0`.
- EPP logs showed repeated HoL blocking decisions.
- Restarting the EPP restored the normal request latency.

The 9.7s latency may also include a timeout or retry in the gateway/ext-proc path, so we have not established that the utilization detector alone explains the full latency. The observation is that stale metrics can turn a small pool into an unavailable pool even while the runtime may still be serving requests.

## Current behavior

The relevant behavior is:

1. Missing or stale metrics contribute `1.0` to pool saturation.
2. The default usage limit is `1.0`.
3. Dispatch is blocked when `saturation >= usageLimit`.

The utilization detector scheduling filter has a different fallback: if all candidates are filtered out, it returns the original candidate list. The flow-control saturation path therefore takes a more conservative stance than the scheduling filter when metrics are stale.

## What would be useful

Would it make sense to make stale-metrics handling configurable while keeping `fail-closed` as the default?

For example, deployments could choose between:

- `fail-closed`: preserve the current behavior.
- A bounded grace period or last-known value for short metrics interruptions.
- An explicit `fail-open` mode for deployments that prefer availability when metrics collection is unavailable.

Any non-default mode would need clear documentation and operational metrics, since stale metrics can also mean that the model server is overloaded or unreachable.

We would be interested in feedback on which behavior is the best fit for the router. We can prepare a focused implementation after the preferred behavior is agreed.

todayim:
**What type of PR is this?**
/kind feature

Adds a configurable `stalenessPolicy` to the utilization saturation detector:

- `saturated` (default) preserves the current fail-closed behavior and scores missing/stale endpoint metrics as `1.0`.
- `ignore` excludes missing/stale endpoints from the pool saturation average; an all-stale pool evaluates to `0.0` and continues dispatching.

The stale endpoint gauge and rate-limited detector log remain active under both policies. Documentation and metric help text describe the configured posture and the operational trade-off. The scheduling `Filter` behavior is unchanged. Tests cover defaults, validation, mixed/all-stale pools, empty pools, and stale-endpoint observability.

Fixes #2408

```release-note
The utilization saturation detector now supports `stalenessPolicy: saturated|ignore`, defaulting to `saturated`.
```

**Test plan**:
- `go test ./pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/... -count=3`
- `git show --check`

### DIFF
--- a/pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/README.md
+++ b/pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/README.md
@@ -23,18 +23,26 @@ The global pool saturation is then evaluated across all candidate endpoints as a
     PoolSaturation = Average(EndpointScore)
 
 **Heterogeneous Deployments:** Because this detector calculates saturation as an unweighted average of individual endpoint scores, it treats all endpoints equally regardless of their physical capacity. In deployments with heterogeneous compute (e.g., mixing H100 and L4 nodes), a small, saturated endpoint has the exact same impact on global backpressure as a massive, saturated endpoint. Contrast this with the Concurrency Detector, which evaluates saturation as a single aggregate fraction, biasing toward larger endpoints.
-*Note: Endpoints with missing or stale metrics are aggressively scored as 100% saturated (fail-closed).*
+*Note: Endpoints with missing or stale metrics are scored according to `stalenessPolicy`: as 100% saturated under `saturated` (default), or excluded from the pool average under `ignore`. An empty candidate list reads 1.0 under both policies.*
 
-**Operational dependency:** because stale endpoints score as saturated, Flow Control dispatch depends on the
-health of model-server metrics collection (scrape path, port, TLS, auth). A fleet-wide scrape outage pins
+**Operational dependency:** under `saturated`, a fleet-wide scrape outage pins
 `flow_control_pool_saturation` at 1.0 and halts dispatch entirely. The `flow_control_stale_endpoints` gauge and a
 rate-limited detector log distinguish this from genuine overload: stale endpoints read exactly 1.0, while genuine
-oversubscription typically reads above 1.0. This fail-closed posture is deliberate — admitting blind on missing
+oversubscription typically reads above 1.0. `saturated` is the default because admitting blind on missing
 data risks overloading model servers with no backpressure signal at all. Staleness also tends to correlate with
-overload: a server too busy to serve its metrics endpoint is often the one that is saturated, so failing open
-would hide exactly the wrong endpoints. The posture is not configurable; if field evidence shows a need, a
-staleness policy (for example, holding the last known score for a bounded window) can be added later without
-changing this default.
+overload: a server too busy to serve its metrics endpoint is often the one that is saturated, so `ignore`
+discounts the endpoints most likely to be overloaded from the backpressure signal. Deployments whose availability budget cannot absorb a dispatch halt
+during a metrics collection outage can select `ignore`, which excludes stale endpoints from the saturation
+average and scores an all-stale pool at 0.0, consistent with the Filter's fail-open fallback. Under `ignore`,
+a fleet-wide scrape outage means dispatching with no backpressure signal at all; the stale-endpoints gauge and
+detector log are the primary operational signal in that posture.
+
+`stalenessPolicy` governs the `SaturationDetector` role only. The `Filter` role always drops stale candidates
+and falls back to the original list when every candidate is stale, independent of this setting. Slow scrape
+cadences should be tuned with `metricsStalenessThreshold` rather than `ignore`: if the poll interval is
+longer than the staleness threshold, every sample reads stale between polls and `saturated` will block
+dispatch almost continuously. The policy field is an enum so additional degradations (for example, holding
+the last known score for a bounded window) can be added later.
 
 ### Role in Scheduling (The Traffic Shaper)
 The detector implements the `Filter` interface to protect individual endpoints. It removes endpoints from candidate lists if their telemetry is stale, or if they exceed specific safety limits:
@@ -59,7 +67,8 @@ The plugin accepts JSON parameters decoding to the following fields:
 
 - `queueDepthThreshold` (`int`): Target waiting queue depth limit. Serves as the "ideal" queue capacity for a single endpoint. Must be > 0. (Default: `5`)
 - `kvCacheUtilThreshold` (`float64`): Target KV cache memory utilization limit, expressed as a fraction. Must be in `(0.0, 1.0]`. (Default: `0.8`)
-- `metricsStalenessThreshold` (`string` / duration): Maximum age of metrics before an endpoint is considered stale. Stale endpoints are treated as 100% saturated. Must be > 0. (Default: `"200ms"`)
+- `metricsStalenessThreshold` (`string` / duration): Maximum age of metrics before an endpoint is considered stale. How stale endpoints are scored is controlled by `stalenessPolicy`. Must be > 0. (Default: `"200ms"`)
+- `stalenessPolicy` (`string`): How endpoints with missing or stale metrics contribute to pool saturation: `"saturated"` scores them as fully saturated; `"ignore"` excludes them from the saturation average. (Default: `"saturated"`)
 - `headroom` (`float64`): Allowed burst capacity above the ideal thresholds, expressed as a fraction (e.g., `0.2` for 20%). Must be >= 0.0. (Default: `0.0`)
 
 ## Trade-offs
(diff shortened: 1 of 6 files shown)


pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/detector_test.go
```diff
@@ -284,6 +308,80 @@ func TestDetector_Saturation(t *testing.T) {
...
+		},
+		{
+			name: "Single pod with good capacity",
+			pods: []fwkdl.Endpoint{
+				makePodMetric("pod1", 2, 0.5, baseTime),
+			},
+			wantSaturation: 0.5 / 0.9,
+		},
+		{
+			name: "Single stale pod scores fresh average",
```
LukeAVanDrie: nit: the name doesn't describe the assertion; the case covers the all-stale path for a one-pod pool. Something like "Single stale pod excluded; all-stale pool scores 0.0".

pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/README.md
```diff
@@ -23,18 +23,26 @@ The global pool saturation is then evaluated across all candidate endpoints as a
...
 rate-limited detector log distinguish this from genuine overload: stale endpoints read exactly 1.0, while genuine
-oversubscription typically reads above 1.0. This fail-closed posture is deliberate — admitting blind on missing
+oversubscription typically reads above 1.0. `saturated` is the default because admitting blind on missing
 data risks overloading model servers with no backpressure signal at all. Staleness also tends to correlate with
-overload: a server too busy to serve its metrics endpoint is often the one that is saturated, so failing open
-would hide exactly the wrong endpoints. The posture is not configurable; if field evidence shows a need, a
-staleness policy (for example, holding the last known score for a bounded window) can be added later without
-changing this default.
+overload: a server too busy to serve its metrics endpoint is often the one that is saturated, so ignoring
+would pass traffic to exactly the wrong endpoints. Deployments whose availability budget cannot absorb a dispatch halt
```
LukeAVanDrie: nit: a bit overstated. Under `ignore` the Filter still drops stale endpoints in a mixed pool, so traffic only reaches them through the all-stale fallback. The added risk is that the backpressure signal discounts the endpoints most likely to be overloaded. 

Consider: "so `ignore` discounts the endpoints most likely to be overloaded from the backpressure signal." instead.

pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/detector_test.go
```diff
@@ -439,3 +537,37 @@ func TestDetector_StaleEndpointObservability(t *testing.T) {
...
+		KVCacheUtilThreshold:      0.90,
+		MetricsStalenessThreshold: time.Hour,
+		StalenessPolicy:           StalenessIgnore,
+	}, logr.Discard())
+
+	detector.Saturation(context.Background(), []fwkdl.Endpoint{
+		makePodMetric("stale", 1, 0.1, time.Now().Add(-2*time.Hour)),
+	})
+
+	families, err := ctrlmetrics.Registry.Gather()
```
LukeAVanDrie: nit (optional): this duplicates the `staleGaugeValue` closure in `TestDetector_StaleEndpointObservability`. A shared `staleGaugeValue(t, detectorName)` helper would let both tests use it.

pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/README.md
```diff
@@ -23,18 +23,26 @@ The global pool saturation is then evaluated across all candidate endpoints as a
     PoolSaturation = Average(EndpointScore)
 
 **Heterogeneous Deployments:** Because this detector calculates saturation as an unweighted average of individual endpoint scores, it treats all endpoints equally regardless of their physical capacity. In deployments with heterogeneous compute (e.g., mixing H100 and L4 nodes), a small, saturated endpoint has the exact same impact on global backpressure as a massive, saturated endpoint. Contrast this with the Concurrency Detector, which evaluates saturation as a single aggregate fraction, biasing toward larger endpoints.
-*Note: Endpoints with missing or stale metrics are aggressively scored as 100% saturated (fail-closed).*
+*Note: Endpoints with missing or stale metrics are scored according to `stalenessPolicy`: as 100% saturated under `saturated` (default), or excluded from the pool average under `ignore`.*
```
LukeAVanDrie: nit: the note could state that an empty candidate list reads 1.0 under both policies; only the metric help text documents that.

LukeAVanDrie (dismissed):
Thanks, this matches the shape we agreed on in #2408. One item before merge: #2221 landed on main and rewrote both saturation metric help strings (`stage` label on `flow_control_pool_saturation`, a per-stage staleness caveat on `flow_control_stale_endpoints` pointing at #2475), so the branch now conflicts in `pkg/epp/metrics/llm_d_router_metrics.go`. When you rebase, fold the `stalenessPolicy` phrasing into the stage-aware help text from #2221; a resolution that keeps the pre-#2221 strings would drop the stage semantics. `metrics.go` auto-merges, but re-read the combined paragraph.

On the refresh-interval observation: that explains the incident. The router's default `refresh-metrics-interval` is 50ms with a 50ms clamp floor, so the 15s cadence came from your deployment's controller; against the 200ms staleness default that leaves endpoints stale for ~98.7% of every cycle. A startup warning when the refresh interval exceeds a detector's staleness threshold is worth doing; please file it as its own issue. Since this PR resolves the configurable-posture ask, I'd also change `Ref #2408` to `Fixes #2408` and let the warning track separately.

Nits inline; none block beyond the rebase.

todayim:
Thanks for the review.

I rebased onto main and resolved the conflict with #2221 by preserving the stage-aware metric help text while folding in the `stalenessPolicy` wording. I also addressed the inline comments: renamed the all-stale test case, clarified the README wording and empty-pool behavior, and shared the stale-gauge helper between the observability tests.

The refresh/staleness mismatch warning is tracked separately in #2514, and the PR now uses `Fixes #2408`.

`go test ./pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/... -count=3` passes; `git diff --check` clean.

pkg/epp/metrics/llm_d_router_metrics.go
```diff
@@ -406,8 +406,8 @@ var (
 			Help: metricsutil.HelpMsgWithStability(
 				"Number of candidate endpoints whose metrics are missing or older than the staleness threshold, as of "+
 					"the most recent saturation evaluation. Recorded by the utilization saturation detector, which scores "+
-					"these endpoints as fully saturated in flow_control_pool_saturation (fail-closed): a nonzero value "+
-					"during a dispatch stall indicates a metrics collection problem rather than genuine overload. "+
+					"these endpoints according to stalenessPolicy: saturated by default or excluded under ignore. A nonzero "+
+					"value during a dispatch stall indicates a metrics collection problem rather than genuine overload. "+
```
LukeAVanDrie, line 410: non-blocking nit: this diagnosis only holds under `saturated`. Under `ignore`, stale endpoints are excluded from the average, so a dispatch stall means the fresh endpoints are genuinely oversubscribed. An operator reading this during a stall could misread real overload as a scrape problem.

LukeAVanDrie (approved):

pazqo:
Hi, what's the status here? Do we have any update? Thanks!

todayim:
Hi @pazqo — the review feedback from @LukeAVanDrie is addressed and he approved the current head (`1d5b320`); full CI (unit, e2e, lint) is green and the branch remains mergeable with main. We're waiting on a final approval from one of the requested reviewers (@RishabhSaini, @hexfusion, @elevran). Happy to make any further changes if needed.

ahg-g (approved):

PR #2495, opened 2026-08-21
feat(epp): make the fairness_id metric label cap configurable

mathias-kende:
**What type of PR is this?**

/kind feature

The fairness_id metric label is capped at a fixed 1000 distinct values to bound cardinality. In large deployments (several active EPP + many different models) this can lead to a huge cardinality. This PR exposes the cap as a new `--fairness-id-metric-label-limit` flag to allow lowering the cardinality (down to retain only the `other` label).

None

```release-note
Add the `--fairness-id-metric-label-limit` flag to lower the cardinality of fairness related metrics.
```

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -385,6 +385,7 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 	setupLog.Info("EPP config after phase two", "config", eppConfig)
 
 	// --- Setup Metrics Server ---
+	metrics.SetFairnessIDLabelLimit(opts.FairnessIDMetricLabelLimit)
 	r.customCollectors = append(r.customCollectors, collectors.NewInferencePoolMetricsCollector(ds))
 	metrics.Register(r.customCollectors...)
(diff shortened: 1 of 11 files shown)


LukeAVanDrie (commented):
docs/metrics.md is now stale. It still states fairness_id "is capped at 1000 distinct values" as a fixed fact. Can we change this to say the cap defaults to 1000 and is configurable via --fairness-id-metric-label-limit, including the 0-collapses-everything behavior? Else, this LGTM

ahg-g (commented):
I see this enforcement is limited to metrics defined in epp/metrics, but not if the fairness_id was used in other places, like plugin emitted metrics: https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go#L203

I think we may want to make the cardinality part of the epp/framework/observability pkg and export it so that both plugins and core code can use it? I can approve for now if you wish for this to be addressed in a followup PR.

mathias-kende:
Thanks for the review. I fixed the metrics.md doc and moved the cardinality to the observability package (I also preserved the previous version in the commit history if you prefer this in two different commits). Please take another look.

ahg-g (approved):

PR #2501, opened 2026-08-21
feat: pod watch peer discovery plugin

evacchi:
**What type of PR is this?**

/kind feature

Adds a Kubernetes peer-discovery mechanism for cross-replica EPP state sync:
a `PeerDiscovery` plugin that watches this EPP deployment's own Pods through
the datalayer notification runtime, instead of its Service's EndpointSlices.

- `k8s-peer-discovery` plugin
  (`pkg/epp/framework/plugins/datalayer/discovery/k8speer`), Alpha stability,
  with `selector`, `port`, and `namespace` parameters, selected via
  `dataLayer.discovery.peers.pluginRef`.
- The plugin registers a Pod notification extractor. The extractor filters
  Pods by namespace, selector, and readiness, requires `POD_IP` to exclude
  self, and hands upsert/delete operations to `Start`, which applies them to
  the `PeerNotifier` from a single goroutine. The configured namespace must
  match the EPP deployment namespace because the Pod cache is namespace scoped.
- Datalayer notification controllers run on every replica, not only the
  leader, so each replica maintains its own peer set.
- Runner wiring resolves the configured `PeerDiscovery` plugin, owns the
  `MemoryPeerStore`, and runs `Start` as a manager runnable. Nothing reads
  the store yet; the `CrossReplicaSyncer` consumer lands separately.
- `PeerDiscovery.Ready` no longer promises a complete initial peer set.
  Unlike endpoints, where an empty set fails every request, a partial peer
  set only lowers sync accuracy, and the set changes with replica churn
  anyway. The notification runtime delivers Pods one at a time and has no
  end-of-initial-list signal, so a strict guarantee would need new shared
  datalayer machinery with no consumer.
- `docs/peer-discovery.md` documents the plugin, its parameters, and an
  example config.

- Part of #1892 
- Follow up to #2341

```release-note
NONE
```


pkg/epp/controller/epp_peer_reconciler.go
```diff
@@ -0,0 +1,164 @@
...
+		id := types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}
+		desired[id] = fwkdl.PeerMetadata{ID: id, Address: pod.Status.PodIP, Port: r.Port}
+	}
+	return desired
+}
+
+func (r *EPPPeerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	matches := func(obj client.Object) bool {
+		return obj.GetNamespace() == r.Namespace && r.Selector.Matches(labels.Set(obj.GetLabels()))
+	}
```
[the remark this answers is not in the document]
evacchi: this is following pre-existing convention

pkg/epp/controller/epp_peer_reconciler.go
```diff
@@ -0,0 +1,135 @@
...
+	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+	podutil "github.com/llm-d/llm-d-router/pkg/epp/util/pod"
+)
+
+// EPPPeerReconciler discovers peer EPP replicas by watching Pods matching this
+// EPP deployment's own selector and drives their membership through a
+// PeerNotifier. Peer discovery must run on every replica (not just the
+// leader), so leader election is disabled for its controller.
+type EPPPeerReconciler struct {
+	client.Reader
```
elevran, line 43: This reconciler duplicates machinery the datalayer framework already provides generically. See `pkg/epp/framework/plugins/datalayer/extractor/topology/extractor.go`'s `podNotificationHandler` for the existing pattern: a plugin implementing Registrant registers a NotificationExtractor for Pod GVK, and Runtime.Start binds it to the manager without any per-plugin wiring in runner.go. Filtering (readiness there, readiness+namespace+selector+self-IP here) happens inside `Extract()` against the deep-copied unstructured object.

Worth reconsidering this whole file in favor of that path - see review body for the full rationale.

cmd/epp/runner/runner.go
```diff
@@ -507,6 +518,13 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 		return nil, nil, err
 	}
 
+	if kp, ok := peerDiscPlugin.(*k8speer.Plugin); ok {
```
elevran, line 521: resolvePeerDiscovery resolves any fwkdl.PeerDiscovery generically, but only *k8speer.Plugin is ever actually started here. A future PeerDiscovery implementation that isn't *k8speer.Plugin would resolve fine and pass config validation, then silently never run - no error, no log. Compare to EndpointDiscovery, which is started generically (disc.Start(ctx, fwkdl.NewDiscoveryNotifier(ds))) regardless of concrete type.

This goes away if the plugin is reworked to use Registrant/NotificationExtractor instead (see review body), since Runtime.Start binds every registered NotificationSource generically.

test/integration/epp/epp_peer_discovery_integration_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+	peerDisc, ok := plugin.(*k8speer.Plugin)
+	require.True(t, ok, "factory returned %T, want *k8speer.Plugin", plugin)
+
+	mgr, mgrClient := setupTestManager(t, testEnv.Config, nsName)
+	require.NoError(t, peerDisc.SetupWithManager(mgr))
+
+	startManagerAndWaitForSync(ctx, t, mgr)
+
+	// Create pods. The manager-registered Start runnable binds the buffer to
```
elevran, line 175: Stale comment: "binds the buffer to the store" refers to a buffering mechanism that was removed (see the "refactor: simpler design, no mutexes or buffers" commit). The reconciler now writes directly into the store via a notifier bound at construction, there's no buffer to bind.

pkg/epp/framework/interface/plugin/plugins.go
```diff
@@ -16,7 +16,9 @@ limitations under the License.
 
 package plugin
 
-import "encoding/json"
+import (
```
elevran, line 19: nit: unrelated import-style reformat (single import to grouped form), nothing else in this file is touched by this PR. Suggest reverting to keep the diff scoped to peer discovery.

docs/peer-discovery.md
```diff
@@ -105,6 +106,49 @@ returning a deterministically ordered snapshot of the current peer set.
 
 ---
 
+## In-tree plugin: k8s-peer-discovery
+
+The `k8s-peer-discovery` plugin
+(`pkg/epp/framework/plugins/datalayer/discovery/k8speer`) watches Pods
```
elevran, line 112: This section should be modified if the built-in k8s notification extractor mechanism is used.

elevran (commented):
This implementation follows the existing controller-runtime reconciler pattern closely, but I'd like to raise an architectural question: this adds a bespoke controller-runtime reconciler (EPPPeerReconciler) wired directly against the manager, plus concrete-type special-casing in runner.go. 
The code in pkg/epp/datalayer already generalizes exactly this need ("watch a k8s GVK, react to events") via NotificationSource/NotificationExtractor/Registrant - see how TopologyExtractor's podNotificationHandler watches Pods and filters by readiness inside Extract(), auto-creating its own Pod notification source. Runtime.Start(ctx, mgr) then binds every registered NotificationSource to the manager generically, with no per-plugin-type wiring needed in runner.go.

Implementing k8s-peer-discovery as a NotificationExtractor for the Pod GVK (plus filtering by namespace/selector/readiness/self-IP, etc) would drop epp_peer_reconciler.go and its test entirely, drop the concrete-type assertion in runner.go, and also means any future non-k8speer PeerDiscovery implementation gets started for free instead of silently never running (see inline comment).

One detail to work out under that design: Ready() currently closes after the reconciler's first List() completes; the notification-extractor path has no equivalent single "initial sync done" callback, so Ready() would need a different signal (e.g. waiting on the manager cache sync for the Pod GVK directly).

Also confirming: Plugin.Store() isn't consumed anywhere yet (no CrossReplicaSyncer wiring) - assuming that's expected as staged follow-up work toward #1892.

evacchi:
thanks for the review, will follow-up soon!

pkg/epp/framework/plugins/datalayer/discovery/k8speer/plugin.go
```diff
@@ -0,0 +1,180 @@
...
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+func (p *Plugin) acceptsPod(pod *corev1.Pod) bool {
+	return p.selector.Matches(labels.Set(pod.Labels)) &&
+		pod.Status.PodIP != "" &&
+		pod.Status.PodIP != p.selfAddress &&
```
elevran, line 182: `selfAddress` comes from `os.Getenv("POD_IP")` (likely via the k8s Downward API) with no validation that it is set. If POD_IP is not injected, `selfAddress` is "". A real pod IP is never equal to "", so this EPP replica passes the self-exclusion check and discovers itself as a peer.

Factory already treats selector/port/namespace as required; `POD_IP` should get the same treatment (fail fast rather than degrade silently into a self-referential peer set).

As a fallback or alternative: in a normal (i.e., non-hostNetwork) pod, `os.Hostname()` returns the pod's own name. Since the plugin already requires namespace and is already watching every Pod in it, you could exclude `self` by matching `os.Hostname()` against `id.Name` (namespace-scoped) instead of, or as a fallback to, the IP comparison. This does not require extra API calls or DNS resolution.

pkg/epp/framework/plugins/datalayer/discovery/k8speer/extractor.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (e *podExtractor) Extract(ctx context.Context, event fwkdl.NotificationEvent) error {
+	if event.Object == nil {
+		return nil
+	}
+
+	// A Pod's namespace never changes, so Pods outside it are never peers.
+	// Skipping them keeps unrelated Pod churn off the handoff to Start.
+	id := types.NamespacedName{Name: event.Object.GetName(), Namespace: event.Object.GetNamespace()}
+	if id.Namespace != e.plugin.namespace {
```
elevran, line 55: The plugin's `namespace` parameter is independent of the Pod cache's actual scope (`gknn.Namespace` in pkg/epp/server/controller_manager.go). If a user sets `namespace` to anything other than the EPP's own deployment namespace, this filter silently drops every event and the peer set stays empty forever: no error, no log. Worth a debug/warn log when the peer set stays empty past Ready, or documenting plainly that this must match the EPP's own namespace.

cmd/epp/runner/runner.go
```diff
@@ -934,6 +943,34 @@ func (r *Runner) resolveDiscovery(rawConfig *configapi.EndpointPickerConfig) (fw
...
+	p := r.PluginHandle.Plugin(ref)
+	if p == nil {
+		return fmt.Errorf("peerDiscovery: no plugin found with name %q", ref)
+	}
+	disc, ok := p.(fwkdl.PeerDiscovery)
+	if !ok {
+		return fmt.Errorf("peerDiscovery: plugin %q does not implement PeerDiscovery", ref)
+	}
+
+	peerStore := statesync.NewMemoryPeerStore()
```
elevran, line 974: peerStore is local to setupPeerDiscovery and never exposed. Nothing in this PR or on main reads Peers(). MemoryPeerStore's doc comment says "a CrossReplicaSyncer reads it via Peers", so a consumer lands in a follow-up PR. 
Consider if to introduce it in the follow up or here. Leaving to your discretion.

elevran (commented):
Design is clean overall: reusing the existing namespace-scoped Pod watch avoids the RBAC/Service changes the earlier approaches needed, the single-goroutine handoff via the unbuffered channel is straightforward, and the doc updates to `PeerDiscovery.Ready()` semantics are clear. Test coverage (unit + integration) looks solid. CI is green except lint which needs addressing.

pkg/epp/framework/plugins/datalayer/discovery/k8speer/extractor_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (r *testRegistrar) Register(reg fwkdl.PendingRegistration) error {
+	r.registration = reg
+	return nil
+}
+
+func TestFactoryRequiresPodIP(t *testing.T) {
+	t.Setenv("POD_IP", "")
+
+	_, err := Factory("peer-disc", json.NewDecoder(
```
elevran, line 48: Use `fwkplugin.StrictDecoder(json.RawMessage(...))` instead of `json.NewDecoder(strings.NewReader(...))`. The framework wraps runtime parameters in `StrictDecoder` (with `DisallowUnknownFields`), so this test would let a typo'd field pass that production would reject. See pkg/epp/framework/plugins/datalayer/discovery/file/plugin_test.go for the canonical pattern in this package.

pkg/epp/framework/plugins/datalayer/discovery/k8speer/plugin.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+const PluginType = "k8s-peer-discovery"
+
+type params struct {
+	// Selector matches this EPP deployment's own pods (its peers), in
+	// Kubernetes label selector syntax (e.g. "app=my-epp"). Required.
+	Selector string `json:"selector"`
+	// Port is the port peer replicas listen on for state sync. Required.
+	Port string `json:"port"`
```
elevran, line 46: The sibling discovery/file plugin validates port with `strconv.Atoi` plus a 1..65535 range check. `PeerMetadata.Port` is a string here too, and a peer consumer will dial it at runtime, so a non-numeric value would surface as a dial failure far from the config. Worth a matching check at Factory time so misconfiguration is rejected early and loudly.

test/integration/epp/epp_peer_discovery_integration_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+	params := fmt.Sprintf(`{"selector":"app=epp-peer-test","port":"9002","namespace":%q}`, nsName)
+	plugin, err := k8speer.Factory("peer-disc", json.NewDecoder(strings.NewReader(params)), nil)
+	require.NoError(t, err)
+
+	peerDisc, ok := plugin.(*k8speer.Plugin)
+	require.True(t, ok, "factory returned %T, want *k8speer.Plugin", plugin)
+
+	mgr, mgrClient := setupTestManager(t, testEnv.Config, nsName)
+	runtime := datalayer.NewRuntime(0)
```
elevran: nit: 0 is opaque without looking at the constructor's doc. A one-line comment (for example: 0 = synchronous handoff, no buffered queue) or a link to the constructor's doc would keep the next reader from second-guessing the literal.

pkg/epp/framework/plugins/datalayer/discovery/k8speer/plugin.go
```diff
@@ -0,0 +1,184 @@
...
+		return nil, errors.New(PluginType + ": 'selector' parameter is required")
+	}
+	if p.Port == "" {
+		return nil, errors.New(PluginType + ": 'port' parameter is required")
+	}
+	if p.Namespace == "" {
+		return nil, errors.New(PluginType + ": 'namespace' parameter is required")
+	}
+	selfAddress := os.Getenv("POD_IP")
+	if selfAddress == "" {
```
elevran, line 98: have you considered using the `os.Hostname()` fallback, noted below, when POD_IP is not set? 
Neither port nor hostname are stable anyway, but the host name might be preferred since does not require setting additional Downward API parameters in the Pod Spec.

elevran (commented):
Release note is NONE though the plugin is opted into via config and changes Alpha surface. 
This is a new user facing feature and PR description requires some description for awareness.

pkg/epp/framework/plugins/datalayer/discovery/k8speer/plugin.go
evacchi: Ah right forgot about that! Will do

pkg/epp/framework/plugins/datalayer/discovery/k8speer/extractor_test.go
evacchi: tests now use the framework’s StrictDecoder.

pkg/epp/framework/plugins/datalayer/discovery/k8speer/plugin.go
evacchi: plugin configuration now rejects ports outside 1..65535.

pkg/epp/framework/plugins/datalayer/discovery/k8speer/plugin.go
evacchi: should work now

pkg/epp/framework/plugins/datalayer/discovery/k8speer/extractor.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil
+	}
+
+	switch event.Type {
+	case fwkdl.EventDelete:
+		return e.plugin.emit(ctx, peerEvent{kind: fwkdl.EventDelete, id: id})
```
elevran, line 61: nit: 
`EventDelete` is forwarded for any Pod in the namespace, without the selector/readiness check that `EventAddOrUpdate` goes through. Harmless since `PeerNotifier.Delete` is a no-op for an unknown ID, but every Pod deletion in the namespace now round-trips through the handoff channel to `Start`, not just ones that were ever peers.
Flagging since I don't think it is intended.

pkg/epp/framework/plugins/datalayer/discovery/k8speer/plugin.go
```diff
@@ -0,0 +1,196 @@
...
+const PluginType = "k8s-peer-discovery"
+
+type params struct {
+	// Selector matches this EPP deployment's own pods (its peers), in
+	// Kubernetes label selector syntax (e.g. "app=my-epp"). Required.
+	Selector string `json:"selector"`
+	// Port is the port peer replicas listen on for state sync. Required.
+	Port string `json:"port"`
+	// Namespace is the namespace of this EPP deployment's pods. Required.
+	Namespace string `json:"namespace"`
```
elevran, line 49: nit: 
namespace must match the EPP deployment's own namespace, and nothing at startup verifies that. A misconfigured value silently discovers zero peers instead of failing fast.
Might be worth a follow-up to read the pod's actual namespace (using downward API) and warn (if not defined via Downward API) or fail (if defined) on mismatch, but that can be a separate enhancement and PR.

elevran (dismissed):
Two non-blocking suggestions below, nothing blocking left.
Placing /hold in case you'd like to address the Selector filtering on delete. Feel free to cancel the hold.

pkg/epp/framework/plugins/datalayer/discovery/k8speer/extractor.go
evacchi: I think this is slightly more involved than the other suggestions so I'd keep this in the follow-up

pkg/epp/framework/plugins/datalayer/discovery/k8speer/plugin.go
evacchi: this seems a relatively low-hanging fruit if we assume that we can read it from the NAMESPACE env var, so I'm doing it (if that's what you mean)

elevran:
@evacchi lint is failing, likely causing other checks to fail as well. Please fix

pkg/epp/framework/plugins/datalayer/discovery/k8speer/extractor.go
elevran: Please open a follow up issue to track.

elevran (approved):
approved. 
@evacchi please file the follow up issue

PR #2509, opened 2026-08-22
epp: remove code deprecated since v0.7.0

### LINKED ISSUE #2493: Remove code deprecated as of v0.7.0
## Summary

Several plugins, parameters, and metrics were marked `Deprecated` as of the
v0.7.0 release (see `release-0.7` branch). The project is now past release-0.10.0 -
several majors past any reasonable deprecation window. This issue tracks removing them.

Scope is strictly limited to items already deprecated by v0.7.0. Later
deprecations (by-label filters, legacy tokenizer plugin, `nohitlru`/
`predictedlatency` fields, `--vllm-port`, the `llm_d_inference_scheduler_*`
metric aliases in `pkg/epp/metrics/metrics.go`) were deprecated at v0.9.0+ and
are out of scope here.

## Candidates

1. **`pd-profile-handler` plugin** (whole plugin)
   - `pkg/epp/framework/plugins/scheduling/profilehandler/disagg/pd_profile_handler.go`
   - `PdProfileHandlerType`, `PdProfileHandlerParameters`, `PdProfileHandlerFactory`,
     `PdProfileHandlerConfigParser`, `NewPdProfileHandler`, `PdProfileHandler` struct + methods
   - Superseded by `disagg-profile-handler`. Registered via
     `RegisterDeprecatedWithPluginDependencies` in `cmd/epp/runner/runner.go:619`.

   - **Breaking** for any EndpointPickerConfig still using `pd-profile-handler`.

2. **`SchedulerPDDecisionCount` metric + `RecordPDDecision` func**
   - `pkg/epp/framework/plugins/scheduling/profilehandler/disagg/metrics.go`
   - Only consumer is (1); becomes dead code once (1) is removed.
   - Drops the `llm_d_inference_scheduler_pd_decision_total` metric (superseded
     by `llm_d_epp_pd_decision_total`). 
   - **Breaking** for dashboards/alerts still scraping the old metric name.

3. **`RoleBoth = "both"` label value**
   - `pkg/epp/framework/plugins/scheduling/filter/bylabel/roles.go`
   - Deprecated alias for `RolePrefillDecode`, but still wired into
     `NewDecodeRole`/`NewPrefillRole` as an accepted label value.
   - **Runtime behavior change, not just cleanup**: pods labeled
     `llm-d.ai/role: both` would stop matching decode/prefill filters.

4. **Legacy flat parameters on `disagg-profile-handler`**
   - `pkg/epp/framework/plugins/scheduling/profilehandler/disagg/disagg_profile_handler.go`
   - `decodeProfile`, `prefillProfile`, `encodeProfile`, `prefillDeciderPluginName`,
     `deciderPluginName`, `encodeDeciderPluginName` (`legacyDisaggProfileHandlerParameters`,
     `toDisaggParams`)
   - Superseded by the nested `profiles`/`deciders` config format.
   - **Breaking** for any config still using the flat parameter format.

## Notes

- Items 1, 2 and 4 remove config surface that active deployments may still use
- Item 3 changes routing behavior for live traffic. This likely needs a
  release note and probably a major/minor version bump per semver, not a
  silent patch.
- `data-parallel-profile-handler` plugin logs "Deprecated: Use simple-profile-handler with Istio >= 1.28.1" but is registered via plain fwkplugin.Register (not RegisterDeprecated) in runner.go:588 and doesn't have a formal removal schedule tracked by the plugin registry.

roytman:
**What type of PR is this?**
/kind cleanup
/kind deprecation

Five plugins/parameters/metrics were marked `Deprecated` as of the v0.7.0
release (verified against the `release-0.7` branch). The project is now
several majors past that, well beyond any reasonable grace period. This
removes them:

- `pd-profile-handler` plugin — superseded by `disagg-profile-handler`.
- `SchedulerPDDecisionCount`/`RecordPDDecision`/`LlmdPDDecisionCount` — the
  `pd_decision_total` metric pair, dead code once `pd-profile-handler` is
  gone (it was their only caller).
- `RoleBoth` (`"both"`) label alias — use `prefill-decode` instead.
  **Behavior change**: pods labeled `llm-d.ai/role: both` no longer match
  the decode/prefill filters.
- `disagg-profile-handler`'s legacy flat parameter format
  (`decodeProfile`/`prefillProfile`/`encodeProfile`/`deciderPluginName`/etc.)
  — use the nested `profiles`/`deciders` format.

`data-parallel-profile-handler` is intentionally **not** removed here: unlike
the others it has no formal removal schedule in the plugin registry, and the
e2e "vLLM Data Parallel" test depends on its `primaryPort` header-rewrite
behavior, which `single-profile-handler` doesn't replicate. Tracked
separately if still wanted.

Also updates e2e/integration test fixtures that referenced the removed
plugin types as YAML string literals (missed by an initial symbol-only
grep): drops the `deprecatedPdConfig` fixture and its dedicated e2e test,
adds an equivalent nixlv2 e2e test against `disagg-profile-handler` so that
connector isn't left untested, and migrates the
`wide-ep-lws-experimental-dp-aware` integration case onto
`disagg-profile-handler`/`disagg-headers-handler`.

Fixes #2493

**Test plan**:
- [x] `make format` / `make lint` — clean
- [x] `make test-unit-epp`, `make test-unit-sidecar`, `make test-unit-framework` — pass
- [x] `make test-integration-hermetic PATTERN=TestWellKnownConfigs` — migrated `wide-ep-lws-experimental-dp-aware` config parses and instantiates correctly
- [x] `make test-integration-hermetic PATTERN=TestE2EConfigs_StartupParseSmoke` — remaining e2e config fixtures still parse after fixture removal

```release-note
- Removed the deprecated `pd-profile-handler` plugin; use `disagg-profile-handler`.
- Removed the deprecated `llm_d_inference_scheduler_pd_decision_total` metric; use `llm_d_epp_pd_decision_total`.
- Removed the deprecated `"both"` value for the `llm-d.ai/role` pod label; use `"prefill-decode"` instead. Pods labeled `both` no longer match the decode/prefill filters.
- Removed `disagg-profile-handler`'s deprecated flat parameter format (`decodeProfile`, `prefillProfile`, `encodeProfile`, `deciderPluginName`, `prefillDeciderPluginName`, `encodeDeciderPluginName`); use the nested `profiles`/`deciders` format.
```


yankay:
Hi @roytman  , thanks for the cleanup. Could we also remove `"both"` from the coordinator e2e decode selector in `test/coordinator/e2e/coordinator/configs_test.go`? This fixture manually mirrors `decode-filter`, so it would be helpful to keep their supported values aligned. Thanks! :)

roytman:
> Hi @roytman , thanks for the cleanup. Could we also remove `"both"` from the coordinator e2e decode selector in `test/coordinator/e2e/coordinator/configs_test.go`? This fixture manually mirrors `decode-filter`, so it would be helpful to keep their supported values aligned. Thanks! :)

Hi @yankay , thank you for the good catch. Fixed in [65693ea](https://github.com/llm-d/llm-d-router/pull/2509/changes/65693eab495be11512ede7a91539ac6738af01da)

ן

elevran (dismissed):
Thanks - well-scoped removal that matches issue #2493 exactly. Code, docs unit/integration/e2e fixtures all updated consistently, including a metrics.md line that was actually wrong before this PR (both metrics were described as sharing labels, but only the llm_d_epp_ variant carries plugin_name/plugin_type).

One minor, non-blocking note: pkg/epp/framework/interface/plugin/registry.go's `RegisterDeprecatedWithPluginDependencies` loses its only caller once pd-profile-handler's registration goes away in this PR (that file isn't touched here, so I can't leave an inline comment on it). Not blocking - it's exported, no linter catches it - but worth deciding whether to drop it now or keep it as a template for a future deprecated plugin that also has dependencies.

Otherwise looks good to merge once required reviews are in.

cmd/epp/runner/runner.go
```diff
@@ -616,8 +616,6 @@ func (r *Runner) registerInTreePlugins() {
 	fwkplugin.Register(single.SingleProfileHandlerType, fwkplugin.StabilityBeta, single.SingleProfileHandlerFactory)
 	fwkplugin.RegisterDeprecated(disagg.DisaggHeadersHandlerType, fwkplugin.StabilityBeta, disagg.HeadersHandlerFactory, "v0.9.0", "v0.11.0", disagg.DisaggProfileHandlerType) //nolint:staticcheck // deprecated in v0.9.0 (#905)
 	fwkplugin.RegisterDeprecated(disagg.PrefillHeaderHandlerType, fwkplugin.StabilityBeta, disagg.HeadersHandlerFactory, "v0.9.0", "v0.11.0", disagg.DisaggProfileHandlerType) //nolint:staticcheck // deprecated in v0.9.0 (#905)
-	fwkplugin.RegisterDeprecatedWithPluginDependencies(disagg.PdProfileHandlerType, fwkplugin.StabilityBeta, disagg.PdProfileHandlerFactory,                                   //nolint:staticcheck // deprecated in v0.7.0 (#732/#756)
```
gyliu513, line 620: please also remove the dead function `RegisterDeprecatedWithPluginDependencies` at `registry.go`

pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -563,7 +563,7 @@ func partitionEndpoints(endpoints []fwkdl.Endpoint) (prefill, decode, interleave
 			prefill = append(prefill, ep)
 		case bylabel.RoleDecode, "":
 			decode = append(decode, ep)
-		case bylabel.RolePrefillDecode, bylabel.RoleBoth, bylabel.RoleEncodePrefillDecode: //nolint:staticcheck // RoleBoth is deprecated but must be matched for backward compatibility
```
gyliu513, line 566: Instead of removing `RoleBoth,`, can you add another case for `RoleBoth` by logging some info message like 

```
endpoint uses deprecated role label \"both\"; excluded from all prefill/decode pools, relabel to \"prefill-decode\
```

roytman:
@elevran , @gyliu513 , I intentionally left `RegisterDeprecatedWithPluginDependencies` for possible future usage. But will be happy to remove it; your call.

pkg/epp/flowcontrol/controller/internal/processor.go
roytman: `RoleBoth` label was deprecated in v0.7.0. Why do you want to leave it?

gyliu513: In case there are some upgrade cases, for those cases, the pod may still have old labels, thoughts?

gyliu513:
@roytman I think leaving some dead code is weird, can you either remove or comment it out?

elevran:
> @roytman I think leaving some dead code is weird, can you either remove or comment it out?

Remove please. Git remembers if we ever need to resurrect.

pkg/epp/flowcontrol/controller/internal/processor.go
roytman: It is the same as when somebody uses the deprecated and now removed objects/methods

roytman:
`RegisterDeprecatedWithPluginDependencies` is removed

gyliu513:
lgtm, thanks @roytman

ahg-g (approved):

PR #2510, opened 2026-08-22
Repair precise cache indexes with bounded full reports

### LINKED ISSUE #1581: Precise prefix caching KV block match error
**What happened**:

When EPP container is newer than vLLM containers, the vLLM containers have received traffic before EPP container startup, and the EPP is configured to use precise prefix caching, the EPP logs the following error for each request where it can't find the parent KV block:

{"level":"error","ts":1781099812.4359953,"caller":"kvevents/pool.go:245","msg":"Failed to get request key for parent block","parentEngineKey":"9908210600003700225","effectiveModelName":"openai/gpt-oss-120b","error":"engine key not found: 9908210600003700225","stacktrace":"github.com/llm-d/llm-d-kv-cache/pkg/kvevents.(*Pool).processEventBatch\n\t/go/pkg/mod/github.com/llm-d/llm-d-kv-cache@v0.7.1/pkg/kvevents/pool.go:245\ngithub.com/llm-d/llm-d-kv-cache/pkg/kvevents.(*Pool).processRawMessage\n\t/go/pkg/mod/github.com/llm-d/llm-d-kv-cache@v0.7.1/pkg/kvevents/pool.go:205\ngithub.com/llm-d/llm-d-kv-cache/pkg/kvevents.(*Pool).worker.func1\n\t/go/pkg/mod/github.com/llm-d/llm-d-kv-cache@v0.7.1/pkg/kvevents/pool.go:181\ngithub.com/llm-d/llm-d-kv-cache/pkg/kvevents.(*Pool).worker\n\t/go/pkg/mod/github.com/llm-d/llm-d-kv-cache@v0.7.1/pkg/kvevents/pool.go:184"}

EPP is able to handle prefix matching without errors for all requests.

Run a single-replica EPP with a vLLM model server with precise prefix caching enabled.  Send some requests to vLLM via the EPP.  Restart the EPP.  Send more requests to vLLM via EPP and check the EPP logs.

This may be expected behavior but was looking for clarification on whether this pointed to a configuration mistake on my side. Eventually the KV cache should clear out and prefix cache matches should be successful but with low traffic that could take some time.  I also recall there was some discussion of persisting the prefix cache beyond the lifecycle of a single EPP container run, but I'm just looking for general guidance with this issue and if there are any known remediations.

**Environment**:
- llm-d-router version (use `git describe --tags --dirty --always` if you built from source, or specify the tag if you used a tagged version or image): 0.8.0
- Cloud provider or hardware configuration: AWS EKS + Istio
- Install tools:  Helm
- Others: vLLM 0.19.1 (although this issue has been spotted on newer versions)

vMaroon:
Recover event-derived prefix state after subscriber attachment or a missing-parent fault without making every request authoritative. Base repair decisions on the selected endpoint's confirmed, non-speculative residency and keep the mechanism opt-in.

Fixes #1581 

```release-note
Set `fullReportRepair: {}` on `precise-prefix-cache-producer` to request bounded full cache reports and recover prefix-cache state after missed KV events. Repair is disabled by default. Configure `fullReportThreshold`, `minMissingBlocks`, and `cooldown` under `fullReportRepair` to control report frequency. For disaggregated deployments, `fullReportRepair.prefillProfile` must match the handler's prefill profile. This mode requires vLLM pod discovery without replay and vLLM support for full report requests and `BlockStored.origin` values `NEW` and `REUSED`. Global ZMQ and replay configurations are rejected.
```


pkg/kvevents/event_dedup_filter.go
```diff
@@ -101,10 +101,15 @@ func (s blockScope) key(blockHash uint64) dedupKey {
 // pops on eviction and emits BlockRemoved), so the map tracks the engine's
 // offload-pool capacity and drains in steady state; an entry is also reclaimed
 // the moment its count returns to zero, and clear() drops a whole pod on
-// AllBlocksCleared. The only growth paths are lost ZMQ removes or an
+// AllBlocksCleared. The growth paths are lost ZMQ removes or an
 // index-internal eviction with no matching wire remove, both of which leave a
-// harmless over-estimate (see above) rather than a correctness error. This
-// mirrors the bounded-by-the-wire behavior of Dynamo's EventDedupFilter.
+// harmless over-estimate (see above) rather than a correctness error, and a
+// full KV-cache report (the preciseprefixcache producer's fullReportRepair)
```
nilig: Full-report re-announcements cannot increment this physical-residency refcount. [vLLM's full-report contract](https://github.com/vllm-project/vllm/pull/45261) emits `BlockStored` for an already-resident reused block without changing block state, so `store(H) -> report store(H) -> sole remove(H)` leaves the count at 1 and suppresses `Index.Evict`. The precise index then advertises a hit after vLLM has evicted the block; cooldown only slows the inflation, and index eviction does not clear the poisoned dedup entry. We need an origin discriminator such as [vLLM #51699](https://github.com/vllm-project/vllm/pull/51699) and a set/reconciliation path that updates the index without `trackStore`, plus a store/report/remove regression test.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
```diff
@@ -199,3 +215,65 @@ func (p *Producer) PreRequest(ctx context.Context,
...
+	requestFull, reason := p.fullReportRepair.shouldRequest(endpoint, match)
+	if !requestFull {
+		return
+	}
+	// Use the body mutator so every protocol serializer sees the new argument.
+	request.Body.MutatePayloadMap(func(payload fwkrh.PayloadMap) {
+		if setXArgs {
+			payload["vllm_xargs"] = xargs
+		}
+		xargs["kv_cache_report_mode"] = "full"
```
nilig, line 271: `/inference/v1/generate` never sees this top-level field. This producer also handles `Body.Generate`, but [current vLLM](https://github.com/vllm-project/vllm/blob/da329cc303a5233e17fa3d553ce0a3d6ceea87a8/rust/src/server/src/routes/inference/generate/types.rs#L33-L50) expects custom sampling arguments under `sampling_params.vllm_xargs` and flattens the top-level key into ignored `other`. We still consume `shouldRequest`'s cooldown and increment `FullReportRequests`, so native Generate traffic records a repair attempt without emitting a report. Branch on the parsed request shape and mutate the nested native field, with a serialized Generate wire-shape test.

pkg/kvevents/pool.go
```diff
@@ -506,6 +528,8 @@ func (p *Pool) processEventBatch(ctx context.Context, batch *EventBatch, podIden
 						"numTokens", len(ev.Tokens),
 						"numBlockHashes", len(ev.BlockHashes),
 						"blockSize", ev.BlockSize)
+					// The child is dropped, so the endpoint needs a repair report.
+					p.notifyStreamEvent(podIdentifier, StreamEventMissingParent)
```
nilig: Keep enough fault identity to prevent unrelated traffic from consuming this signal. `MissingParent` sets one endpoint-wide `force`, and `shouldRequest` clears it on whichever eligible request arrives next. If prefix/LoRA A loses a child at 168/200 blocks and disjoint B arrives first at the same endpoint, B's report consumes the flag but cannot restore A; A remains above the 0.80 threshold and is never repaired. Carry the affected model/prefix through the stream event and consume it only for an overlapping request, or keep the fault armed until there is evidence that the affected lineage was repaired.

nilig (changes requested):
Full-report repair still has three correctness gaps that can leave stale cache residency in the index or consume a repair without repairing the affected prefix. I think these need to be resolved before merge.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -396,11 +420,19 @@ func (p *Producer) produceFromBlockKeys(ctx context.Context, span trace.Span,
 			info.WithMM(attrprefix.MMMatchInfo{MatchBlocks: countMMMatchedBlocks(mmBlockIndices, cachedBlocks)})
 		}
 		ep.Put(p.dk, info)
+		// Repair ignores speculative rows so predictions cannot hide a stale index.
+		if repairMatches != nil {
```
nilig, line 395: non-blocking: Enabling repair adds a confirmed-prefix scan for every candidate even when coverage is complete and no report is requested. On this head, with the real in-memory index and default scorer over a warm 3,750-block prefix, I measured `22.25 ms -> 22.99 ms` at 8 pods (+3.3%) and `439.5 ms -> 513.8 ms` at 96 pods (+16.9%, `p < .001`, 12 samples per arm) on an Apple M4 Max with `GOMAXPROCS=1`. The control performs the same `PluginState.Write`, so the delta isolates the added confirmed scan. This is a high-end fully replicated prefix, but it exposes another `O(blocks * pods^2)` traversal at large candidate counts. Could we derive confirmed counts during the existing block traversal so repair does not rescan the pod list for every block and candidate?

yankay:
HI @vMaroon 

Thanks for the discussion. Two follow-ups:

1. The large-integer corruption appears to be a broader request-body mutation issue, though this PR exposes it through repair. I opened #2518 to track it separately.
2. vLLM's [`AnthropicMessagesRequest`](https://github.com/vllm-project/vllm/blob/0a21947d710f5aedb1865038ebef20e141b29c58/vllm/entrypoints/anthropic/protocol.py#L121-L169) does not accept `vllm_xargs`, so this repair path appears ineffective for Anthropic requests. [#2511](https://github.com/llm-d/llm-d-router/pull/2511) does not change this schema. Could we confirm and exclude Anthropic if needed?

Please correct me if I have missed anything. :)

vMaroon:
> HI @vMaroon
> 
> Thanks for the discussion. Two follow-ups:
> 
> 1. The large-integer corruption appears to be a broader request-body mutation issue, though this PR exposes it through repair. I opened [Request payload repackaging silently loses precision for large JSON integers #2518](https://github.com/llm-d/llm-d-router/issues/2518) to track it separately.
> 2. vLLM's [`AnthropicMessagesRequest`](https://github.com/vllm-project/vllm/blob/0a21947d710f5aedb1865038ebef20e141b29c58/vllm/entrypoints/anthropic/protocol.py#L121-L169) does not accept `vllm_xargs`, so this repair path appears ineffective for Anthropic requests. [#2511](https://github.com/llm-d/llm-d-router/pull/2511) does not change this schema. Could we confirm and exclude Anthropic if needed?
> 
> Please correct me if I have missed anything. :)

Hey, (1) sounds good, for (2), I PR'd vLLM before opening this one. 
- https://github.com/vllm-project/vllm/pull/53308

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/repair.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	state, eligible := r.endpoints[endpoint]
+	if !eligible {
+		return false, ""
+	}
+	// Ignore small gaps before considering either repair path.
+	missing := match.total - match.confirmed
+	if missing < r.minMissing || match.total <= 0 {
```
bongwoobak: This floor gates the integrity path too, not just the ratio one, since it sits before the force check. `missing` can never exceed `total`, so with `minMissingBlocks` 32 and the default 16-token block any request whose prompt is under 512 tokens fails it, and a missing_parent fault on an endpoint whose traffic stays under that stays armed in `force` for the subscriber lifetime without ever producing a report. For the ratio path a block-count floor makes sense, but a report is worth more than the blocks it carries, `emit_cached_block_events` emits the whole reused prefix as one root-anchored event carrying its token span, so even a sub-floor request rebuilds the chain that was making ordinary incremental events drop their children. Could force use a smaller floor, or skip it?

nilig:
The issues from my previous review are addressed, including the extra confirmed-prefix scan.

Non-blocking clarification: after EPP attaches to a warm engine, `reportSupported` remains false until an origin-bearing event arrives. If requests reuse cached blocks without completing new blocks, repair can remain inactive despite missing index entries. Do we accept that startup limitation, or should we establish capability independently of new cache activity?

We also need to coordinate with #2707: its parser represents existing `vllm_xargs` objects as `json.RawMessage`, which `repairPayloadMap` rejects, so those requests silently skip repair. The parser-to-repair test passes on this PR alone and fails with both PRs combined. We should support that representation in one of the PRs before both land.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
```diff
@@ -199,3 +215,81 @@ func (p *Producer) PreRequest(ctx context.Context,
...
+		selected = prefill
+	}
+	if selected == nil || len(selected.TargetEndpoints) == 0 {
+		return
+	}
+	metadata := selected.TargetEndpoints[0].GetMetadata()
+	if metadata == nil {
+		return
+	}
+	endpoint := fmt.Sprintf("%s:%s", metadata.Address, metadata.Port)
```
ahg-g: Recommend to consolidate endpoint address formation behind a utility func, I am sure there are other places that does the same.

pkg/kvevents/pool.go
```diff
@@ -354,6 +385,13 @@ func (p *Pool) worker(ctx context.Context, workerIndex int) {
 
 // processRawMessage decodes the raw message payload using the adapter and processes the resulting event batch.
 func (p *Pool) processRawMessage(ctx context.Context, msg *RawMessage) {
+	if msg.stream != nil {
+		msg.stream.mu.RLock()
+		defer msg.stream.mu.RUnlock()
+		if msg.stream.detached {
```
ahg-g: Was this a bug? what was the side effect of processing messages of from a broken stream?

pkg/kvevents/pool.go
```diff
@@ -545,6 +585,9 @@ func (p *Pool) processEventBatch(ctx context.Context, batch *EventBatch, podIden
 		switch ev := genericEvent.(type) {
 		case *BlockStoredEvent:
 			deviceTier := normalizeDeviceTier(ev.DeviceTier)
+			if ev.Origin == "NEW" || ev.Origin == "REUSED" {
```
ahg-g: Should we define the values in an enum? are these values defined by vllm? if so, should we have a layer of abstraction so we can translate it from sglang kv events if sglang communicates them differently?

pkg/kvevents/pool.go
```diff
@@ -545,6 +585,9 @@ func (p *Pool) processEventBatch(ctx context.Context, batch *EventBatch, podIden
...
+				p.notifyStreamEvent(podIdentifier, StreamEventReportSupported)
```
ahg-g, line 588: it seems we are tracking support for report based on block events. This seems a bit odd, I would expect that we decide that on a "stream established" event, rather than a "block event"

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/repair.go
```diff
@@ -0,0 +1,183 @@
...
+			delete(state.missing, block)
+		}
+	case kvevents.StreamEventCleared:
+		state.missing = nil
+		state.reportSupported = false
+	}
+	r.endpoints[endpoint] = state
+}
+
+func (r *fullReportRepair) shouldRequest(endpoint string, match repairMatch) (bool, string) {
```
ahg-g: pls document when a report is requested

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
```diff
@@ -52,6 +54,12 @@ type speculativeEntries struct {
 // requests use a length-1 outer slice.
 type blockKeysState struct {
 	perPromptKeys [][]kvblock.BlockHash
+	repairMatches map[string]repairMatch
+}
+
+type repairMatch struct {
```
ahg-g, line 64: document what this type tracks and per what

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -368,15 +392,19 @@ func (p *Producer) produceFromBlockKeys(ctx context.Context, span trace.Span,
 		if len(mmBlockIndices) > 0 {
 			info.WithMM(attrprefix.MMMatchInfo{MatchBlocks: countMMMatchedBlocks(mmBlockIndices, match.MatchedBlocks)})
 		}
+		if repairMatches != nil {
+			addr := fmt.Sprintf("%s:%s", md.Address, md.Port)
```
ahg-g: we already create the endpoint address earlier in the loop, dedup pls

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
```diff
@@ -199,3 +215,81 @@ func (p *Producer) PreRequest(ctx context.Context,
...
+	requestFull, reason := p.fullReportRepair.shouldRequest(endpoint, match)
+	if !requestFull {
+		return
+	}
+	// Use the body mutator so every protocol serializer sees the new argument.
+	request.Body.MutatePayloadMap(func(payload fwkrh.PayloadMap) {
+		if request.Body.Generate != nil {
+			payload["sampling_params"] = argsParent
+		}
+		argsParent["vllm_xargs"] = xargs
```
ahg-g: This is vllm specific, can we abstract this?

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/prerequest.go
```diff
@@ -199,3 +215,81 @@ func (p *Producer) PreRequest(ctx context.Context,
...
+			return
+		}
+	}
+	xargs, ok := repairPayloadMap(argsParent["vllm_xargs"])
+	if !ok {
+		log.FromContext(ctx).V(logging.DEBUG).Info("Skipping full report repair for malformed vllm_xargs",
+			"requestID", request.RequestID, "endpoint", endpoint)
+		return
+	}
+	requestFull, reason := p.fullReportRepair.shouldRequest(endpoint, match)
```
ahg-g: should we do this check earlier in the func?

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/repair.go
```diff
@@ -0,0 +1,183 @@
...
+	state, eligible := r.endpoints[endpoint]
+	if !eligible || !state.reportSupported || match.total <= 0 {
+		return false, ""
+	}
+	if r.cooldown > 0 && !state.lastRequest.IsZero() && r.clock.Since(state.lastRequest) < r.cooldown {
+		return false, ""
+	}
+	reason := "integrity"
+	if len(state.missing) == 0 {
+		missing := match.total - match.confirmed
```
ahg-g: pls document this heuristic, why do we trigger a report in this case.

ahg-g (commented):
Thanks Maroon. My main observation from an architectural standpoint is around consistency: the KV-cache repair and stream observation logic introduces what looks like a parallel extension mechanism rather than leveraging EPP’s core plugin framework.

I don't want to block this PR on a redesign, but I think it highlights an important architectural gap. Let's schedule a wider follow-up discussion on how KV-cache events, lifecycle hooks, and indexing should cleanly fit into EPP's plugin architecture moving forward.

ahg-g:
@vMaroon  What is the status of this PR?

pkg/kvevents/event_dedup_filter.go
```diff
@@ -192,3 +193,23 @@ func (f *eventDedupFilter) clear(podIdentifier string) {
 	defer f.mu.Unlock()
 	delete(f.refs, podIdentifier)
 }
+
+// trackReport restores an untracked residency without adding a physical copy.
+func (f *eventDedupFilter) trackReport(scope blockScope, hashes []uint64) {
```
hyeongyun0916: If the router misses both NEW stores for the same hash (for example, two physical copies computed in the same step before a router restart) and then receives that hash through a full report, does trackReport set the count to 1, so the first BlockRemoved evicts the entry while a copy is still resident?

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/repair.go
```diff
@@ -0,0 +1,224 @@
...
+//     report and an under-indexed one restores its prefix. The floor and ratio
+//     keep small gaps and mostly indexed prompts from spending the cooldown.
+func (r *fullReportRepair) shouldRequest(endpoint string, match repairMatch) (string, bool) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	state := r.endpoints[endpoint]
+	if state == nil || !state.reportSupported || match.total <= 0 || r.coolingDown(state) {
+		return "", false
+	}
+	if len(state.missing) > 0 {
```
hyeongyun0916: suppose request X, routed to endpoint A, has its new blocks dropped for a missing parent. Those blocks stay in A's fault list, so do later requests Y and Z to A keep carrying kv_cache_report_mode: full once per cooldown, unless one of them reuses X's blocks (for example, a follow-up turn of X's conversation) or the engine evicts them?

PR #2511, opened 2026-08-22
fix(tokenizer): match Anthropic prefix tokens to vLLM rendering

vMaroon:
**What type of PR is this?**

/kind bug

Precise prefix-cache routing can miss reusable blocks for Anthropic `/v1/messages` requests when the router and vLLM derive different prompt tokens from the same request. Because several Anthropic controls affect the beginning of the rendered prompt, these differences can make the block identity diverge from block zero even while KV-event ingestion is healthy.

The Anthropic render path added by [#1896](https://github.com/llm-d/llm-d-router/pull/1896) still had three sources of divergence:

- vLLM reparses nested JSON in `tools[].input_schema` and assistant `tool_use.input`, which canonicalizes object-key order. The router retained the incoming wire order.
- vLLM's handling of inline system messages depends on its Anthropic conversion and model chat template. The router could therefore merge or preserve those messages differently from vLLM. Claude Code billing headers introduce another request-varying prefix unless both sides remove them.
- The parsed Messages request dropped `chat_template_kwargs` and `output_config.effort`. vLLM passes the former to the chat template and maps the latter to `reasoning_effort`, enabling thinking only when the caller did not explicitly set `enable_thinking`.

This PR aligns the router with the request vLLM renders:

- Canonicalizes nested Anthropic JSON after request reparsing.
- Matches vLLM's missing `input_schema.type` insertion order.
- Removes Claude Code billing headers.
- Routes Anthropic requests through the typed vLLM HTTP renderer.
- Automatically detects whether vLLM merges or preserves inline system messages by comparing `/v1/messages/count_tokens` with merged and unmerged render candidates.
- Caches detection once per renderer and fails fast when detection is unavailable or ambiguous instead of producing incorrect cache keys.
- Preserves caller-provided `chat_template_kwargs` without hardcoding model-specific values.
- Maps `output_config.effort` with vLLM's precedence: it overwrites `reasoning_effort`, while an explicit caller-provided `enable_thinking` wins over the derived default.
- Retains `vllm.mergeAnthropicInlineSystem` only as a compatibility override.

The inline-system detector is isolated and marked for deletion once vLLM provides a native Anthropic render endpoint.

Related work:

- [llm-d-router #1896](https://github.com/llm-d/llm-d-router/pull/1896) - merged; added Anthropic support to the render tokenizer.
- [vLLM #44602](https://github.com/vllm-project/vllm/pull/44602) - merged; preserved inline system-message position for prefix caching.
- [vLLM #46025](https://github.com/vllm-project/vllm/pull/46025) - merged; introduced chat-template-based inline-system detection.
- [vLLM #46196](https://github.com/vllm-project/vllm/pull/46196) - open; resolves the model chat template before applying that detection.
- [vLLM #47681](https://github.com/vllm-project/vllm/pull/47681) - open; proposes a shared inline-system policy for OpenAI and Anthropic requests.
- [vLLM #52978](https://github.com/vllm-project/vllm/pull/52978) - open; handles trailing Anthropic system messages without undoing mid-conversation prefix reuse.
- [vLLM #45803](https://github.com/vllm-project/vllm/pull/45803) - open; adds native `/v1/messages/render`, which would replace the router-side conversion and detector.

Verification:

- Parser-to-forwarded-payload parity for nested and reverse-ordered Anthropic JSON.
- Missing `input_schema.type` insertion-order parity.
- Tool-use input, inline-system, billing-header, `chat_template_kwargs`, and `output_config.effort` conversion.
- Merged and preserved automatic detection.
- Explicit override, cached detection, strict-template rejection, and ambiguous-result failure.
- Focused package tests, race tests for the tokenizer and Anthropic parser, and `go vet`.
- A controlled integration reproducer showed zero matches for prompt-affecting Anthropic controls before the fix and full reusable-prefix matches after the fix.
- The patched build restored routing matches in live validation without transport failures, restarts, or request-path errors.

Follow-up to [#1861](https://github.com/llm-d/llm-d-router/issues/1861) and [#1896](https://github.com/llm-d/llm-d-router/pull/1896).

```release-note
Anthropic `/v1/messages` prefix-cache routing preserves prompt-affecting template controls and matches vLLM's request rendering.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
@@ -569,7 +569,8 @@ func appendImageBlock(blocks []tokenizerTypes.ContentBlock, src *fwkrh.Anthropic
 
 // anthropicToolCall converts a tool_use block into an OpenAI function tool
 // call. Arguments are CPython json.dumps formatted (separators, ASCII
-// escaping, wire key order) - the exact string vLLM renders into the prompt.
+// escaping, forwarded key order) - the exact string vLLM renders into the
+// prompt after EPP repackage.
 func anthropicToolCall(b fwkrh.AnthropicContentBlock) map[string]any {
 	id := b.ID
 	if id == "" {
@@ -594,6 +595,7 @@ func pythonArguments(raw json.RawMessage) string {
 	case "", "null", "{}":
 		return "{}"
 	}
+	raw = canonicalizeRepackagedJSON(raw)
 	if out, err := pythonDumps(raw); err == nil {
 		return out
 	}
@@ -638,8 +640,7 @@ func anthropicToolResultContent(b fwkrh.AnthropicContentBlock) (string, []tokeni
 }
 
 // convertAnthropicTools rewrites Anthropic tool definitions into OpenAI
-// function tools. input_schema passes through as raw JSON so the template's
-// re-serialization keeps the wire key order.
+// function tools.
 func convertAnthropicTools(tools []fwkrh.AnthropicTool) []any {
 	if len(tools) == 0 {
 		return nil
@@ -650,6 +651,11 @@ func convertAnthropicTools(tools []fwkrh.AnthropicTool) []any {
 		if len(schema) == 0 || bytes.Equal(schema, []byte("null")) {
 			schema = json.RawMessage(`{"type":"object"}`)
 		}
+		// Anthropic requests are forwarded from PayloadMap, whose Marshal uses
+		// encoding/json and therefore sorts every object key. Canonicalize the
+		// schema the same way before tokenization so the lookup blocks match the
+		// prompt vLLM actually receives after EPP repackage.
+		schema = canonicalizeAnthropicInputSchema(schema)
 		fn := map[string]any{
 			"name":       t.Name,
 			"parameters": schema,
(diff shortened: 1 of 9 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/tokenizer.go
```diff
@@ -565,6 +592,7 @@ func pythonArguments(raw json.RawMessage) string {
 	case "", "null", "{}":
 		return "{}"
 	}
+	raw = canonicalizeRepackagedJSON(raw)
```
nilig: [P1] Use the same JSON order that is sent upstream

This always sorts keys inside `tool_use.input`. But the director only re-serializes the body when `Mutated` is true. In the normal same-model path, it forwards the original bytes. For example, vLLM can receive `{"z":0,"a":1}` while the router tokenizes `{"a":1,"z":0}`. The router then gets different token IDs from vLLM, so cache-aware routing can choose the wrong backend. The same problem applies to `input_schema`. Could we either always repackage Anthropic requests or tokenize the exact bytes that will be forwarded? The test should also go through `Director.repackage` with both mutated and unmutated requests.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http_anthropic_mode.go
```diff
@@ -0,0 +1,94 @@
...
+			{Role: "user", Content: fwkrh.AnthropicContent{Raw: "router-user-probe-two"}},
+		},
+	}
+	countRequest := anthropicCountTokensRequest{
+		Model:    r.modelName,
+		System:   probe.System,
+		Messages: probe.Messages,
+	}
+	var countResponse anthropicCountTokensResponse
+	if err := r.postJSON(ctx, anthropicCountPath, countRequest, r.timeout, &countResponse); err != nil {
```
nilig, line 64: [P1] Do not require `count_tokens` from a render-only server

`vllm launch render` supports the render routes, but it does not serve `/v1/messages/count_tokens`, so this probe returns 404 in supported render-only setups. Detection also runs when the request has no inline `system` message, even though both modes produce the same result in that case. A failed detection is not cached, so every Anthropic request tries again while holding the mutex, and precise routing gets no token IDs. Could we skip detection when there is no inline system message and provide a configured mode that works with render-only endpoints?

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
```diff
@@ -172,6 +181,29 @@ func (r *vllmHTTPRenderer) RenderChat(ctx context.Context, payload fwkrh.Request
 	return r.postChatRender(ctx, body, r.chatTimeout(pm))
 }
 
+// RenderAnthropic converts an Anthropic request using the same inline-system
+// mode as the serving vLLM process and renders the resulting chat request.
+func (r *vllmHTTPRenderer) RenderAnthropic(ctx context.Context, request *fwkrh.MessagesRequest) ([]uint32, *tokenization.MultiModalFeatures, error) {
+	mergeInlineSystem, err := r.anthropicInlineSystemMode(ctx)
```
nilig, line 187: [P2] Include detection in the producer timeout

The outer producer timeout is only `max(timeout, mmTimeout)`. The first automatically detected request makes three calls before the real render: one count call and two probe renders. With the defaults, a multimodal request can need `3 x 5s + 30s = 45s`, but the outer limit is 30 seconds. That leaves the real render only about 15 seconds. Could we run and cache detection earlier, or include these extra calls in `ProduceTimeout` until the mode is known?

nilig (commented):
I found three cases where this can produce the wrong tokens or stop producing tokens. Details are inline.

nilig:
One late-mutation case is still mismatched. `RenderAnthropic` reads `body.Mutated` during DataProducer execution, but `PreRequest` runs later. In #2510, `fullReportRepair` then adds `vllm_xargs.kv_cache_report_mode` through `MutatePayloadMap`, changing `Mutated` from false to true.

The router therefore tokenizes the original order, such as `{"z":0,"a":1}`, but `repackage` later sorts it and vLLM receives `{"a":1,"z":0}`. I reproduced this transition with a focused test.

Could we make the serialization decision stable across the whole request path, either by preserving nested JSON order during late mutations or by canonicalizing and repackaging Anthropic requests before tokenization? A regression test should cover `Produce(Mutated=false) -> PreRequest mutation -> Director.repackage`.

vMaroon:
handled in #2707

PR #2512, opened 2026-08-22
Remove deprecated UDS/legacy tokenizer code and drop llm-d-kv-cache

### LINKED ISSUE #2361: Remove deprecated UDS/legacy tokenizer code and drop `llm-d-kv-cache` dependency

Remove the deprecated UDS and legacy tokenizer code paths, and drop the `github.com/llm-d/llm-d-kv-cache`
After cleanup, the `llm-d-kv-cache` module should no longer required in `go.mod`.

**Why is this needed**:

The UDS tokenizer backend was deprecated in #1079 (2026-05-24) and the embedded tokenizers were deprecated in llm-d/llm-d-kv-cache#473 (2026-04-18). The vLLM HTTP `/render` backend (#890) has been the default replacement since 0.9.0 release

The `llm-d-kv-cache` module is now only imported to support these deprecated code paths.

satyamg1620:
**What type of PR is this?**

/kind cleanup
/kind deprecation

Removes the deprecated UDS tokenizer backend and the in-process tokenization pool, then drops the `github.com/llm-d/llm-d-kv-cache` module from `go.mod`.

The vLLM HTTP `/render` backend has been the default replacement since 0.9.0. The UDS backend and the tokenization pool were the only remaining importers of `llm-d-kv-cache`, so removing them removes the dependency.

Removed:

- `udsTokenizerConfig` on the `token-producer` plugin, along with the UDS tokenizer adapter and the type-bridging helpers it needed.
- `indexerConfig.tokenizersPoolConfig` on `kvcache.Config`, and the precise-prefix-cache scorer's prompt-tokenizing wrapper that consumed it. The scorer now delegates to the producer directly and takes its tokens from an upstream `token-producer`.

Plugin parameters are decoded with `DisallowUnknownFields`, so a configuration that still sets either field is rejected at startup with an error naming the field, rather than being ignored.

`LegacyPluginType` (`tokenizer`) and the `precise-prefix-cache-scorer` deprecation are separate and are left in place.

Fixes #2361

**Verification**:

- [x] `make presubmit`
- [x] `make test-unit`
- [x] Live kind stack, EPP built from this branch: prefix-cache routing over the vLLM `/render` token-producer returns real token counts and a KV cache hit on a repeated prefix.
- [x] Live kind stack: a config setting `udsTokenizerConfig`, and one setting `indexerConfig.tokenizersPoolConfig`, each stop EPP at startup with an unknown-field error naming that parameter.

**Release note**:
```release-note
The deprecated UDS tokenizer backend and the indexer's in-process tokenization pool are removed. A configuration that sets `udsTokenizerConfig` on `token-producer`, or `indexerConfig.tokenizersPoolConfig` on `precise-prefix-cache-producer` or `precise-prefix-cache-scorer`, fails at startup with an unknown-field error naming the parameter. Remove the parameter from every plugin that sets it: `precise-prefix-cache-scorer` ignores its own parameters when a `precise-prefix-cache-producer` is also configured, so a stale value there is not always reported. For tokenization, set `vllm.url` on a `token-producer` to select the vLLM HTTP /render backend; the precise prefix cache takes its tokens from that plugin.

```


pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http_test.go
```diff
@@ -352,7 +352,7 @@ func TestVLLMHTTPRenderer_HTTPError(t *testing.T) {
 func TestPluginFactory_RejectsBothBackends(t *testing.T) {
 	params := `{
 		"modelName": "m",
-		"udsTokenizerConfig": {"socketFile": "/tmp/foo.sock"},
+		"estimate": {},
```
vMaroon, line 441: Could we add a factory test that passes `udsTokenizerConfig` through `plugin.StrictDecoder` and asserts the unknown-field error? The release note declares this startup behavior, but the unit tests only lock the corresponding `tokenizersPoolConfig` rejection.

vMaroon (commented):
Thank you for picking this up - overall looks good.

The branch still conflicts with `main` in the tokenizer README. The resolution should retain `main`'s HTTP/HTTPS, TLS, and mTLS documentation while removing the UDS bullet and migration section.

satyamg1620:
> Thank you for picking this up - overall looks good.
> 
> The branch still conflicts with `main` in the tokenizer README. The resolution should retain `main`'s HTTP/HTTPS, TLS, and mTLS documentation while removing the UDS bullet and migration section.

Rebased. Updated  tokenizer README

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http_test.go
satyamg1620: Added TestPluginFactory_RejectsUDSTokenizerConfig alongside the existing factory tests: it passes udsTokenizerConfig through plugin.StrictDecoder and asserts unknown field "udsTokenizerConfig", so both parameters named in the release note are now locked by tests.

vMaroon (approved):

vMaroon:
Thanks

vMaroon:
cc @zdtsw

PR #2503, opened 2026-08-21
llm-d observability: align OpenTelemetry GenAI semconv and refactor telemetry usages

PlateauGao:
**What type of PR is this?**
/kind feature
/kind cleanup

**What this PR does / why we need it**:
This PR implements subtasks 1&2 of https://github.com/llm-d/llm-d-router/issues/708 to align OpenTelemetry semantic conventions and eliminate hardcoded tracing attribute strings across the llmd router:

1. **Centralized Telemetry Package (`pkg/telemetry`)**:
   - Introduces `pkg/telemetry/genai.go` wrapping OpenTelemetry (otlp) offical  `semconv/v1.41.0` (lastest release v1.44.0) GenAI attribute keys and provides type-safe helper constructors
   - Introduces `pkg/telemetry/llmd.go` for all internal `llm_d.*` namespace attributes. 
   - Adds unit tests

2. **Refactoring Across EPP & KV Cache**:
   - Replaced raw string literals 
   - Updated unit test assertions 

Glad to start contributing to llm-d, please let me know if any contributing steps I am missing. 
**Which issue(s) this PR fixes**:
Fixes #708

**Release note** _(write `NONE` if no user-facing change)_:
```release-note
telemetry: align OpenTelemetry GenAI semantic conventions and refactor telemetry usages

Issue #708, opened 2026-03-11
[Observability] Embrace otel GenAI Semantic Convention for inference observability


gyliu513:
**What would you like to be added**:

The inference scheduler is using otel for observability but it is not using otel genai semantic convention yet, but just hard-coding the attributes like `gen_ai.request.model`, it should use genai semantic convention from otel.

**Why is this needed**:

GenAI Semantic Convention is the standard for GenAI Observability.

gyliu513:
@wenhug do you want to create a PR for this? Seems you already have a PR in your local branch?

gyliu513:
@PlateauGao I think we can take this one as parent issue and you can open some sub tasks if needed, hope this is ok, thanks!

PlateauGao:
Thanks @gyliu513! 

I propose breaking down this parent issue into the following sub-tasks:

- **Subtask 1**: Bump `otel/semconv` dependency, create centralized helper package in `pkg/telemetry/genai.go` and any other potential otlp semantic namespace (wrapping official semconv constants + typed custom helpers), and add unit tests.
- **Subtask 2**: Refactor scheduler and plugin files to replace hard-coded string literals with centralized helpers.
- **Subtask 3**: Align router metrics with GenAI semantic conventions. (this may be discussed later as potential breaking changes in downstream users/dashboards/alerts)

For attributes, we can use official `semconv` for standard GenAI&k8s fields, namespace internal implementation details under `llm_d.*`, and later explore upstreaming general inference attributes (like KV cache metrics) to the OTel GenAI namespace if they are commonly used in all inference infra.

Please let me know if this sounds good to you? I will start subtask1&2 once we are aligned on the plan.

gyliu513:
@PlateauGao sounds great, thanks!

PR #2516, opened 2026-08-23
llm-d observability: finish remaining OpenTelemetry GenAI semconv and telemetry refactoring

PlateauGao:
**What type of PR is this?**
/kind cleanup

This is a follow-up PR to #2503 (implemented subtasks 1&2 of https://github.com/llm-d/llm-d-router/issues/708), completing the migration of all remaining hardcoded `llm_d.*` and `gen_ai.*` OpenTelemetry attributes across the codebase to centralized & type-safe semantic conventions in `pkg/common/observability/semconv`

### Summary of Changes:
1. **Semantic Conventions (`pkg/common/observability/semconv`)**:
   - Expanded `llm_d.go` with typed `attribute.Key` definitions and constructors for remaining legacy usages
   - Added unit test 

### Verification:
- local go test
- local make lint

https://github.com/llm-d/llm-d-router/issues/708

```release-note
NONE

### DIFF
--- a/pkg/common/observability/semconv/llm_d.go
+++ b/pkg/common/observability/semconv/llm_d.go
@@ -21,8 +21,18 @@ import (
 )
 
 // Internal llm-d specific attribute keys.
-// All custom router, scheduler, scorer, and cache attributes are namespaced under "llm_d.*".
+// All custom router, scheduler, scorer, cache, and sidecar attributes are namespaced under "llm_d.*".
 const (
+	// EPP Scheduling, Filtering & Picking attributes
+	LLMDEPPProfileNameKey               = attribute.Key("llm_d.epp.scheduling.profile.name")
+	LLMDEPPFilterCandidateEndpointsKey  = attribute.Key("llm_d.epp.filter.candidate_endpoints")
+	LLMDEPPFilterFilteredEndpointsKey   = attribute.Key("llm_d.epp.filter.filtered_endpoints")
+	LLMDEPPScorerCountKey               = attribute.Key("llm_d.epp.scorer.count")
+	LLMDEPPScoringCandidateEndpointsKey = attribute.Key("llm_d.epp.scoring.candidate_endpoints")
+	LLMDEPPPickerCandidateEndpointsKey  = attribute.Key("llm_d.epp.picker.candidate_endpoints")
+	LLMDEPPPickerTopEndpointsKey        = attribute.Key("llm_d.epp.picker.top_endpoints")
+	LLMDEPPPickerTopScoresKey           = attribute.Key("llm_d.epp.picker.top_scores")
+
 	// EPP Scorer attributes
 	LLMDEPPScorerTypeKey               = attribute.Key("llm_d.epp.scorer.type")
 	LLMDEPPScorerNameKey               = attribute.Key("llm_d.epp.scorer.name")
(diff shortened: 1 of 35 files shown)

### CODE REVIEW

File: pkg/kvevents/pool.go [hunk 1]
```diff
@@ -22,7 +22,7 @@ import (
 	"sync"
 	"sync/atomic"
 
-	"go.opentelemetry.io/otel/attribute"
+	"github.com/llm-d/llm-d-router/pkg/common/observability/semconv"
```
gyliu513: The semconv import landed in the third-party group here, worth moving it into the existing llm-d-router group, please also check if other files following this pattern as well.

pkg/common/observability/semconv/llm_d.go
```diff
@@ -21,76 +21,206 @@ import (
...
+	LLMDPDProxyChunkedDecodeChunkSizeKey      = attribute.Key("llm_d.pd_proxy.chunked_decode.chunk_size")
+	LLMDPDProxyChunkedDecodeStreamingKey      = attribute.Key("llm_d.pd_proxy.chunked_decode.streaming")
+	LLMDPDProxyChunkedDecodeChunksKey         = attribute.Key("llm_d.pd_proxy.chunked_decode.chunks")
+	LLMDPDProxyChunkedDecodeTotalTokensKey    = attribute.Key("llm_d.pd_proxy.chunked_decode.total_tokens")
+	LLMDPDProxyChunkedDecodeDurationMsKey     = attribute.Key("llm_d.pd_proxy.chunked_decode.duration_ms")
+
+	// EC Proxy attributes
+	LLMDECProxyEncodeDisaggregationUsedKey = attribute.Key("llm_d.ec_proxy.encode_disaggregation_used")
+	LLMDECProxyEncoderCountKey             = attribute.Key("llm_d.ec_proxy.encoder_count")
+	LLMDECProxyEncoderAllowedKey           = attribute.Key("llm_d.ec_proxy.encoder_allowed")
```
gyliu513: LLMDECProxyEncoderCount and LLMDECProxyEncoderAllowed define two distinct keys, llm_d.ec_proxy.encoder_count and llm_d.ec_proxy.encoder_allowed, but they carry the same quantity: both call sites in chat_completions.go pass len(allowedEncoders), one on the success branch and one on the SSRF-fallback branch. The doc comments are identical too ("returns an attribute for allowed encoder count in EC proxy"), which is the tell.

This came in from the old hardcoded strings, so it isn't introduced here, but centralizing these keys is the moment it becomes visible, and shipping both as named API makes the split permanent. Two keys for one quantity means a dashboard has to know which branch it's querying to pick the right one. Could we collapse to a single key?

pkg/epp/scheduling/scheduler_profile_picker_tracing_test.go
```diff
@@ -34,6 +34,8 @@ import (
 	fwksched "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
 )
 
+const pod3Endpoint = "/pod3"
```
gyliu513: This pod3Endpoint const looks unrelated to the semconv migration, could it be dropped from this PR?

pkg/common/observability/semconv/llm_d.go
PlateauGao: Kept count only. I can address some semantic definition in this PR but I may prefer to do it separately and keep this PR focus on no-op migration only. Intuitively, I expect to see more semantic/metrics changes like this, especially after having SIG to introduce standard inference otlp semconv.

pkg/kvevents/pool.go
PlateauGao: Done for all components

pkg/epp/scheduling/scheduler_profile_picker_tracing_test.go
PlateauGao: dropped

gyliu513:
lgtm, thanks @PlateauGao !

@ahg-g  can you help approve? Thanks

ahg-g (approved):

PR #2517, opened 2026-08-24
fix(epp): classify non-OK gRPC trailers as errors

### LINKED ISSUE #2468: gRPC stream aborts delivered via trailers record TerminationCause natural
**What happened**:

A gRPC model server that aborts mid-generation ends the response with trailers carrying a non-OK `grpc-status`. The `ResponseTrailers` case in `StreamingServer.Process` calls `finishResponse` without reading the trailers, so `ResponseComplete` is set and the end-of-stream record carries `TerminationCause: natural` (#2444) with whatever usage was parsed before the abort. The `:status` header is 200 for gRPC errors, so `ResponseStatusCode` does not flag the request either.

A consumer that filters on `TerminationCause` to exclude truncated observations keeps these records with their partial token counts.

A stream aborted by the model server classifies as something other than `natural`.

Not yet reproduced end to end. The description comes from reading `pkg/epp/handlers/server.go` (the `ResponseTrailers` case and `finishResponse`) and the vllmgrpc parser. The shipped configurations set `response_trailer_mode: SEND` (`config/charts/llm-d-router-standalone/values.yaml` and the dev Envoy configs), so the trailers case fires for gRPC backends. A reproduction needs a gRPC backend that aborts mid-stream.

Trailers end every gRPC response, successful ones included, so a fix branches on the `grpc-status` value rather than on trailer arrival: it reads `grpc-status` in the trailers path and records a cause before `finishResponse` marks the response complete. It also needs a taxonomy decision: `natural` includes complete error responses, and a gRPC abort delivers a stream end without the full response, so it fits `error` unless consumers need to separate server aborts from EPP-side failures with a distinct value.

Related: #2429 (the `TerminationCause` field), #2444 (its implementation).

**Environment**:
- Kubernetes version (use `kubectl version`): not applicable, found by code reading
- llm-d-router version: main at c51b75a1
- Cloud provider or hardware configuration: not applicable
- Install tools: not applicable
- Others:

yuanhuizang-cloud:
**What type of PR is this?**

/kind bug

A gRPC response can have HTTP status 200 while still terminating with a non-OK `grpc-status` in response trailers.

The EPP `ResponseTrailers` path previously called `finishResponse` without inspecting `grpc-status`. As a result, aborted gRPC streams could reach end-of-stream handling without a termination cause and be recorded as `natural`.

This change classifies non-zero `grpc-status` trailers as `TerminationCauseError` before the response is finalized.

It also adds unit coverage for gRPC trailer termination classification.

Validation:
- `go test ./pkg/epp/handlers -count=1`
- `go test ./pkg/epp/... -count=1`
- `git diff --check`

Fixes #2468

```release-note
NONE
```

### DIFF
--- a/pkg/epp/handlers/server.go
+++ b/pkg/epp/handlers/server.go
@@ -266,7 +266,7 @@ func terminationCauseFromGRPCTrailers(trailers *extProcPb.HttpTrailers) fwkrc.Te
 
 	for _, header := range trailers.GetTrailers().GetHeaders() {
 		if header.Key == "grpc-status" {
-			if grpcStatus := string(header.RawValue); grpcStatus != "" && grpcStatus != "0" {
+			if grpcStatus := envoy.GetHeaderValue(header); grpcStatus != "" && grpcStatus != "0" {
 				return fwkrc.TerminationCauseError
 			}
 			return ""
--- a/pkg/epp/handlers/server.go
+++ b/pkg/epp/handlers/server.go
@@ -259,6 +259,22 @@ func terminationCause(reqCtx *RequestContext, ctxErr error) fwkrc.TerminationCau
 	}
 }
 
+func terminationCauseFromGRPCTrailers(trailers *extProcPb.HttpTrailers) fwkrc.TerminationCause {
+	if trailers == nil || trailers.GetTrailers() == nil {
+		return ""
+	}
+
+	for _, header := range trailers.GetTrailers().GetHeaders() {
+		if header.Key == "grpc-status" {
+			if grpcStatus := string(header.RawValue); grpcStatus != "" && grpcStatus != "0" {
+				return fwkrc.TerminationCauseError
+			}
+			return ""
+		}
+	}
+	return ""
+}
+
 func extractFairnessAndPriority(reqCtx *RequestContext) (string, string) {
 	if reqCtx == nil {
 		return metadata.DefaultFairnessID, "0"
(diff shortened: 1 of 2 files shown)


pkg/epp/handlers/server.go
```diff
@@ -259,6 +259,22 @@ func terminationCause(reqCtx *RequestContext, ctxErr error) fwkrc.TerminationCau
...
 }
 
+func terminationCauseFromGRPCTrailers(trailers *extProcPb.HttpTrailers) fwkrc.TerminationCause {
+	if trailers == nil || trailers.GetTrailers() == nil {
+		return ""
+	}
+
+	for _, header := range trailers.GetTrailers().GetHeaders() {
+		if header.Key == "grpc-status" {
+			if grpcStatus := string(header.RawValue); grpcStatus != "" && grpcStatus != "0" {
```
LukeAVanDrie: Envoy's HeaderValue can populate either value or raw_value depending on version/config, which is why pkg/common/envoy/headers.go has GetHeaderValue handling both. Please use that util here as well. Else, this LGTM!

LukeAVanDrie (changes requested):

pkg/epp/handlers/server.go
yuanhuizang-cloud: Updated to use envoy.GetHeaderValue and added coverage for the HeaderValue.Value representation. Thanks for catching this.

LukeAVanDrie (approved):

ahg-g (approved):

PR #2519, opened 2026-08-24
Preserve JSON number precision during request repackaging

### LINKED ISSUE #2518: Request payload repackaging silently loses precision for large JSON integers
**What happened**:

Request parsers decode payload maps with `json.Unmarshal`, which converts JSON numbers to `float64`. If the payload is later mutated and repackaged, large integers can be silently rounded.

Input:

```json
{"model":"test","prompt":"hello","seed":9007199254740993}
```

After parsing, mutation, and marshaling:

```json
{"model":"test","prompt":"hello","seed":9007199254740992,"vllm_xargs":{"kv_cache_report_mode":"full"}}
```

This can already occur during model-name rewriting. It was also exposed while reviewing #2510 because full-report repair calls `MutatePayloadMap`.

Repackaging must preserve the exact values of untouched JSON numbers.

1. Parse the input above with `OpenAIParser.ParseRequest`.
2. Mutate the returned payload or trigger model-name rewriting.
3. Marshal the `PayloadMap`.
4. Observe that `seed` changes from `9007199254740993` to `9007199254740992`.

The OpenAI, Anthropic, and vLLM HTTP parsers use the same decoding pattern. Using `json.Decoder.UseNumber()` for payload maps and adding a parse -> mutate -> marshal regression test should preserve integer precision.

**Environment**:

- Kubernetes version: N/A
- llm-d-router version: `main` at `ebab687927f5331c1b39defcb579a07951bce56c`
- Install tools: Source checkout
- Others: N/A

yankay:
**What type of PR is this?**

/kind bug

Decode HTTP request payload maps in the OpenAI, Anthropic, and vLLM HTTP
parsers with a shared parser utility based on `json.Decoder.UseNumber()`.
Rename the parser utility package from `grpcutil` to `parserutil` so it can
contain both gRPC and JSON helpers. The utility accepts exactly one JSON value
and rejects trailing data, so payload mutations and repackaging preserve
untouched JSON number values. Typed request parsing remains unchanged.

**Performance impact**:

Decode-only performance can regress for very large numeric arrays. With one
million `token_ids`, payload-map decoding was about 11% slower and used about
21% more memory. The full parse, mutate, and marshal path was about 8% faster,
but used about 13% more memory. This PR prioritizes correctness; preserving
large `token_ids` arrays as raw JSON can be evaluated separately because it
changes their runtime type in `PayloadMap`.

Fixes #2518

**Test plan**:

- [x] Preserve a large integer through OpenAI request parsing, payload mutation, and repackaging.
- [x] Preserve `json.Number` payload values in OpenAI, Anthropic, and vLLM HTTP parsing.
- [x] Reject second JSON values and malformed trailing data in the shared decoder.

```release-note
Preserve JSON number precision when HTTP request payloads are mutated and repackaged.
```

### DIFF
--- a/pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"strings"
 
 	v1 "sigs.k8s.io/gateway-api-inference-extension/api/v1"
@@ -99,8 +100,13 @@ func (p *AnthropicParser) ParseRequest(_ context.Context, body []byte, headers m
 	}
 
 	bodyMap := make(map[string]any)
-	if err := json.Unmarshal(body, &bodyMap); err != nil {
+	decoder := json.NewDecoder(bytes.NewReader(body))
+	decoder.UseNumber()
+	if err := decoder.Decode(&bodyMap); err != nil {
 		return nil, fmt.Errorf("error unmarshaling request body: %w", err)
+	}
+	if err := decoder.Decode(&struct{}{}); err != io.EOF {
+		return nil, errors.New("error unmarshaling request body: unexpected trailing data")
 	}
 
 	var messagesReq fwkrh.MessagesRequest
--- a/pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go
+++ b/pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go
@@ -22,7 +22,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"strings"
 
 	v1 "sigs.k8s.io/gateway-api-inference-extension/api/v1"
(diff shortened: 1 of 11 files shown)


pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go
```diff
@@ -99,9 +100,14 @@ func (p *AnthropicParser) ParseRequest(_ context.Context, body []byte, headers m
 	}
 
 	bodyMap := make(map[string]any)
-	if err := json.Unmarshal(body, &bodyMap); err != nil {
+	decoder := json.NewDecoder(bytes.NewReader(body))
+	decoder.UseNumber()
+	if err := decoder.Decode(&bodyMap); err != nil {
```
zetxqx: Logic looks good. nitpick: it is duplicated in three parsers. Do you mind extracting this to a util function under pkg/epp/framework/plugins/requesthandling/parsers/util in this or a followup PR?

we can extract a common util function and use it in the three parsers

```go
func Unmarshal(data []byte, v any) error {
        dec := json.NewDecoder(bytes.NewReader(data))
        dec.UseNumber()
        if err := dec.Decode(v); err != nil {
                return err
        }
        switch err := dec.Decode(&struct{}{}); {
        case errors.Is(err, io.EOF):
                return nil
        case err == nil:
                return ErrTrailingData
        default:
                return fmt.Errorf("%w: %v", errors.New("unexpected trailing data after JSON value"), err)
        }
  }
```

ahg-g: +1

yankay: HI @zetxqx

Thanks for the suggestion. I moved the shared decoder to `parsers/util/json.go` and updated all three parsers to use `parserutil.Unmarshal`, with trailing-data coverage centralized there. Could you please take another look? :)

ahg-g (approved):

PR #2523, opened 2026-08-24
feat: Support array<array<integer>> prompt in /v1/completions

### LINKED ISSUE #1573: Support `array<array<integer>>` prompt in `/v1/completions`
## Summary

The `/v1/completions` endpoint does not accept `array<array<integer>>` as a prompt format, despite the OpenAI spec defining it as a valid form. Requests using this format currently fail with `"unsupported array element type"`.

## Context

Per the [OpenAI completions spec](https://platform.openai.com/docs/api-reference/completions/create#completions-create-prompt), the `prompt` field accepts four forms:

| Format | Status |
|---|---|
| `string` | Supported |
| `array<string>` | Supported (PR #860) |
| `array<integer>` (token IDs) | Supported |
| `array<array<integer>>` (token ID arrays) | **Not supported** |

**Depends on**: PR #860 (`fix/handle-string-array-completions-prompt`), which introduces `TokenizedPrompt.PerPromptTokens [][]uint32` and the multi-prompt plumbing. Once that lands, the gap is only in parsing and the tokenizer passthrough: token arrays don't need tokenization, they just need to be wrapped into `PerPromptTokens` directly.

## Acceptance criteria

- `prompt: [[1,2,3],[4,5,6]]` parses into `Prompt.TokenIDArrays` and flows through to `PerPromptTokens`
- Mixed nested types and invalid token IDs are rejected with clear errors
- `EmbeddingsInput` continues to reject nested arrays
- All existing tests continue to pass

albertoperdomo2:
**What type of PR is this?**

/kind feature

Adds support for the fourth and final OpenAI `/v1/completions` prompt form: `array<array<integer>>` (multiple pre-tokenized prompts as nested token ID arrays, e.g. `[[1,2,3],[4,5,6]]`).

The parser previously rejected this form with `"unsupported array element type"`. The multi-prompt plumbing (`TokenizedRequest.Prompts`) was already in place; only the parsing and backend passthrough for the nested shape were missing.

Changes:
- Unify `Prompt.TokenIDs` and `EmbeddingsInput.TokenIDs` from `[]uint32` to `[][]uint32`. A flat `[1,2,3]` parses to `[][]uint32{{1,2,3}}`; nested `[[1,2],[3,4]]` parses to two entries.
- Extract `parseTokenIDs` helper and add a `[]any` case to `parseArrayInput` so both `Prompt` and `EmbeddingsInput` share the same three-case type switch (string, float64, []any).
- Add `NewTokenizedRequest([][]uint32)` constructor, replacing four identical loops across render and estimate backends.
- Empty sub-arrays are rejected at parse time.
- Both `renderBackend` and `estimateBackend` pass through pre-tokenized prompts without calling the tokenizer, unchanged from the existing flat path.

Fixes #1573

```release-note
The `/v1/completions` endpoint now accepts `array<array<integer>>` prompts (multiple pre-tokenized token ID arrays), completing support for all four OpenAI prompt forms.
```

### DIFF
--- a/pkg/epp/framework/interface/requesthandling/types.go
+++ b/pkg/epp/framework/interface/requesthandling/types.go
@@ -211,6 +211,15 @@ type PromptTokens struct {
 	MultiModalFeatures []MultiModalFeature
 }
 
+// NewTokenizedRequest builds one PromptTokens entry per token-ID array.
+func NewTokenizedRequest(arrays [][]uint32) *TokenizedRequest {
+	prompts := make([]PromptTokens, len(arrays))
+	for i, ids := range arrays {
+		prompts[i] = PromptTokens{TokenIDs: ids}
+	}
+	return &TokenizedRequest{Prompts: prompts}
+}
+
 // TokenCount returns the total number of tokens across all prompts.
 func (tp *TokenizedRequest) TokenCount() int {
 	if tp == nil {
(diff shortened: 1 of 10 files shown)


pkg/epp/framework/interface/requesthandling/types.go
ahg-g, line 358: What calls this?

ahg-g (approved):

PR #2531, opened 2026-08-24
feat(epp): add session-state-producer plugin

### LINKED ISSUE #2524: [Agentic] Add Session-state-producer

Scheduling plugins can see a session's *identity* but not its *history*.

`agent-identity` resolves a session ID from client headers and publishes it as a request attribute.

However, there is no plugin actually keeping track of an agentic session's state (turns taken, last request seen, etc) and producing it to the scheduling plugins to use that.

Hence, proposing to add a `session-state-producer` that can track a session's history

Initially, we can start with tracking the following states for sessions :

| Field | Purpose |
|---|---|
| `TurnsTaken` | turns this session has dispatched |
| `Duration` | Overall duration of this session |
| `LastSeenAt` | eviction and inactivity decay |

and expand this in future. 
**Why is this needed**:

There are several proposals to bring session-centric routing to llm-d-router. 
Session's states are the basic primitive that can be useful to different strategies of session-centric-routing.

Following proposals are complementary or related to this issue : 

1) [[Agentic] Session-Aware Routing for llm-d](https://docs.google.com/document/d/1QhERAOXrkrChhetQSGIYjPTk_e8X3GLGm8IDF25FmQ8/edit?tab=t.0#heading=h.z9uhi7djn41c) 

2) [llm-d integration: Session-Addressable KVEvents Integration for Agentic Inference](https://docs.google.com/document/d/19m0n7obHJeCV_WkKDISSzOCY4rrHk7G9OqDG1YlPTCI/edit?tab=t.0#heading=h.wwv9k4fifaxs) 

@vMaroon @D-Sai-Venkatesh

paulzhn:
Add session history tracking for agentic routing.

**What type of PR is this?**

/kind feature

Adds a `session-state-producer` plugin that maintains in-memory state for agentic sessions identified by the `agent-identity` request attribute.

The plugin publishes the following session history for scheduling plugins:

- `TurnsTaken`: number of turns previously dispatched for the session
- `Duration`: time elapsed since the session was first observed
- `LastSeenAt`: time the session was previously observed
- `InFlightRequests`: dispatched requests whose response lifecycle has not ended
- `CompletedRequests`: requests whose responses ended naturally
- `TotalInputTokens`: prompt tokens reported by naturally completed responses
- `TotalOutputTokens`: completion tokens reported by naturally completed responses

Every response termination removes the request from `InFlightRequests`. Only natural completions, including fully returned model error responses, increment `CompletedRequests` and the token totals. Client disconnects, evictions, and internal errors do not contribute because their usage may be incomplete.

Session state is isolated per EPP instance. Inactive entries are removed using configurable eviction TTL and sweep intervals, while sessions with in-flight requests are retained until those requests end.

Fixes #2524

```release-note
Add the session-state-producer plugin, which publishes per-session turn, timing, in-flight, completion, and token-usage history for agentic routing.
```


D-Sai-Venkatesh:
Hey @paulzhn, thanks for the PR. Few comments:

We could track additional metrics such as:

1. In-flight requests: requests dispatched from EPP that have not yet completed.
2. Completed requests: requests that have completed.
3. Total input tokens: the sum of input tokens consumed by completed requests.
4. Total output tokens: likewise, the sum of output tokens consumed by completed requests.

pkg/epp/framework/plugins/datalayer/attribute/session/data_types.go
```diff
@@ -20,18 +20,33 @@ limitations under the License.
...
+// SessionStateDataKey identifies session history published on the request
+// attribute store. The default producer is the session-state-producer.
+var SessionStateDataKey = plugin.NewDataKey("SessionStateDataKey", sessionstateconstants.SessionStateProducerType)
+
 // SessionID is the session identifier extracted from a request.
 type SessionID string
 
+// SessionState is the session history available before the current request is
+// dispatched.
+type SessionState struct {
```
D-Sai-Venkatesh: These definitions could move into the session-state-producer package itself, keeping all of its types under the plugin folder.

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionstate/producer.go
```diff
@@ -0,0 +1,206 @@
...
+	lastSeenAt  time.Time
+}
+
+// Producer tracks session history within one EPP instance.
+type Producer struct {
+	typedName fwkplugin.TypedName
+	dk        fwkplugin.DataKey
+
+	mu       sync.Mutex
+	sessions map[string]*sessionRecord
```
D-Sai-Venkatesh: Can we encapsulate this into a SessionStateRegistry with appropriate methods, so the mutex and the map are not exposed on the producer struct directly? Something like RecordDispatch on the write side, and getters such as GetState that return a snapshot. Right now the lock/unlock and the "create record if missing" block are duplicated across Produce, PreRequest, and evictIdle, so having the registry own the concurrency rules removes that duplication and gives us one place to reason about them. It also makes the session bookkeeping testable on its own, independent of the producer.

paulzhn:
> Hey @paulzhn, thanks for the PR. Few comments:
> 
> We could track additional metrics such as:
> 
> 1. In-flight requests: requests dispatched from EPP that have not yet completed.
> 2. Completed requests: requests that have completed.
> 3. Total input tokens: the sum of input tokens consumed by completed requests.
> 4. Total output tokens: likewise, the sum of output tokens consumed by completed requests.

I have added 4 metrics: `InFlightRequests` `CompletedRequests` `TotalInputTokens` `TotalOutputTokens`, and the last three only count if the response finishes naturally. If you have any questions please let me know.

paulzhn:
Hi, if available would you please review my PR, thanks for your time :) @D-Sai-Venkatesh @ahg-g @hexfusion

pkg/epp/framework/plugins/datalayer/attribute/session/README.md
```diff
@@ -1,6 +1,6 @@
 # Session Attributes
 
-Per-request session identity used by affinity-aware scorers and filters.
+Per-request session identity used by session-aware scheduling plugins.
```
praveingk: I think the changes to this `session/README.md` are not relevant to the `session-state-producer` . Can be reverted.

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionstate/README.md
```diff
@@ -0,0 +1,58 @@
...
+`agent-identity` request-header plugin and publishes per-request session state
+for scheduling plugins through the request attribute store.
+
+The producer uses only the `agent-identity` attribute. `SessionIDDataKey` is a
+separate attribute produced by `session-id-producer` from one configured header
+or cookie for general session-affinity use.
+
+## Configuration
+
+Both plugins must be enabled, and Alpha plugins must be allowed by the EPP
```
praveingk, line 13: Can we flag a warning if agent-identity is not enabled? 
or `data_graph.go` enforces such hard dependencies.

paulzhn: Hard dependencies might be a better idea, for the new plugin wouldn't do anything if there's no `agent-identity` configured.

paulzhn:
Hi, I’ve added data dependency declarations for producing and consuming `AgentIdentityKey`. This requires a small change to the `agent-identity` plugin so that it explicitly declares the key it produces. I think this makes the data flow and dependency between the plugins much clearer, so the additional change is worthwhile. @praveingk

praveingk (approved):
Thanks for addressing the comments. Looks good to me 👍🏽

paulzhn:
@vMaroon Could you please take a look at this PR and approve the pending workflow runs?

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionstate/producer.go
```diff
@@ -0,0 +1,194 @@
...
+		return fmt.Errorf("evictionSweepSeconds must be > 0, got %v", p.EvictionSweepSeconds)
+	}
+	return nil
+}
+
+var (
+	_ requestcontrol.DataProducer          = &Producer{}
+	_ requestcontrol.PreRequest            = &Producer{}
+	_ requestcontrol.ResponseBodyProcessor = &Producer{}
+	_ fwkplugin.ConsumerPlugin             = &Producer{}
```
ahg-g, line 62: what about  fwkplugin.ProducerPlugin ?

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionstate/registry.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+type sessionRecord struct {
+	state       SessionState
+	firstSeenAt time.Time
+}
+
+// SessionStateRegistry stores session history and owns its concurrency rules.
+// Its zero value is ready for use.
+type SessionStateRegistry struct {
```
ahg-g, line 31: can't we use PluginState instead of creating a custom registry?

paulzhn: PluginState is designed for per-request state, so we'd need extra logic to manage session lifetimes and eviction across requests. I think keeping the registry is simpler here.

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionstate/producer.go
paulzhn: DataProducer embeds ProducerPlugin, so this should cover both.

praveingk:
@ahg-g Can you please check if this can be merged?

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionstate/registry.go
ahg-g: PluginState is not necessarily per-request, it is a generic state management api, one can use any key, while in most cases plugins use a per request key, one can use a per session key as well.

paulzhn: Using PluginState would change two behaviors:
1. evictionTtlSeconds: 0 currently disables eviction. PluginState has no option to turn off cleanup, so we would have to drop that behavior.
2. The registry keeps sessions with in-flight requests, even after the TTL. PluginState cleans up based on last access, so a long-running request could lose its session state before it finishes.

I’ve pushed the implementation to [this branch](https://github.com/paulzhn/llm-d-router/tree/issue-2524-pluginstate-variant). If you’re okay with these changes, let me know and I’ll push them to this PR branch and update the PR description.

ahg-g (approved):

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionstate/registry.go
ahg-g: sounds good, if it requires all these changes then having a separate registry could be a better option for now

PR #2547, opened 2026-08-25
fix: saturation detector fails open when all endpoints have stale metrics

pazqo:
**What type of PR is this?**
/kind bug

When every candidate endpoint has stale or missing metrics, `Saturation()` returned 1.0 (fail-closed), which blocked all dispatch via HoL blocking. For single-endpoint InferencePools, a brief metrics scrape delay caused the EPP to return HTTP 500 to the gateway, even though the vLLM backend was healthy and serving all forwarded requests successfully.

Two changes:

1. **Fail-open when all endpoints are stale:** `Saturation()` now returns 0.0 when `staleCount == len(candidates)`, consistent with `Filter()` which already fails open in the same situation. Mixed pools (some stale, some fresh) still score stale endpoints as saturated to correctly bias traffic toward fresh endpoints.

2. **Raise default staleness threshold from 200ms to 5s:** The previous 200ms default was shorter than typical metrics scrape intervals, virtually guaranteeing periodic staleness even under normal operation. 5s provides a comfortable margin above the default scrape interval while still detecting genuinely stale metrics.

**Root cause chain:**
1. Utilization detector's `Saturation()` scores stale endpoints as 1.0 (fail-closed)
2. Flow control dispatch cycle sees saturation >= usageLimit, enforces HoL blocking on all priority bands
3. Queued requests expire via TTL (HTTP 429) or the scheduler sees zero endpoints after filtering (HTTP 500)

Change 1 addresses the fail-closed behavior (step 1). Change 2 reduces how often the stale condition is triggered in the first place.

Fixes the root cause of RHOAIENG-87650 (llm-d EPP returns HTTP 500 when metrics-staleness-threshold leaves zero pickable endpoints in single-endpoint InferencePools).

**Release note**:
```release-note
Fixed a bug where the saturation detector would block all request dispatch when every endpoint in the pool had stale metrics, causing HTTP 500 errors for single-endpoint InferencePools during brief metrics scrape delays. Also raised the default metrics staleness threshold from 200ms to 5s to prevent spurious staleness under normal scrape intervals.
```

## Test plan
- [x] Regression test `TestDetector_SingleEndpointStaleMetrics_SaturationFilterInconsistency` verifies Saturation fails open for a single stale endpoint
- [x] Updated existing test cases for all-stale scenarios to expect 0.0 instead of 1.0
- [x] Verified mixed-pool behavior (some stale, some fresh) is unchanged
- [x] All tests in `pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/` pass
- [x] All tests in `pkg/epp/flowcontrol/...` and `pkg/epp/requestcontrol/...` pass
- [ ] Follow-up: integration test wiring real Detector into Processor dispatch cycle

### DIFF
--- a/pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/config.go
+++ b/pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/config.go
@@ -32,8 +32,9 @@ const (
 	// DefaultKVCacheUtilThreshold is the default KV cache utilization (0.0 to 1.0) threshold.
 	DefaultKVCacheUtilThreshold float64 = 0.8
 	// DefaultMetricsStalenessThreshold defines how old metrics can be before they are considered
-	// stale.
-	DefaultMetricsStalenessThreshold time.Duration = 200 * time.Millisecond
+	// stale. The value must comfortably exceed the metrics scrape interval so that a single
+	// delayed scrape does not cause the detector to treat healthy endpoints as saturated.
+	DefaultMetricsStalenessThreshold time.Duration = 5 * time.Second
 	// defaultHeadroom is the default burst allowance (0%).
 	defaultHeadroom float64 = 0.0
 )
--- a/pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/detector.go
+++ b/pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/detector.go
@@ -161,6 +161,16 @@ func (d *Detector) Saturation(_ context.Context, candidates []datalayer.Endpoint
 		d.maybeLogStaleEndpoints(staleCount, len(candidates))
 	}
 
+	// Fail-open when every candidate has stale or missing metrics. A stale timestamp means the
+	// detector has not received fresh telemetry; it does not mean the endpoint is unhealthy or
+	// overloaded. When some endpoints are fresh, scoring stale ones as saturated correctly biases
+	// traffic toward the fresh endpoints. But when ALL are stale there is no fresh alternative,
+	// and hard-blocking dispatch (saturation=1.0) rejects requests against healthy backends.
+	// This is consistent with Filter, which already fails open in the same situation.
+	if staleCount == len(candidates) {
+		return 0.0
+	}
+
 	return totalScore / float64(len(candidates))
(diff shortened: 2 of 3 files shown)


LukeAVanDrie:
@pazqo Thanks for the PR. The root-cause chain in the description matches what we diagnosed in #2408, and this looks like the same incident class (the deployment behind #2408 was also RHOAI, with the modelpp controller setting `--refresh-metrics-interval=15s`). I'm closing this in favor of the work already in flight there, since it covers both changes; details below.

**Change 1 (fail-open when all endpoints are stale):** #2408 landed on making the posture configurable rather than changing the default: a `stalenessPolicy` parameter on `utilization-detector`, with `saturated` (current behavior) as the default and `ignore`, under which stale endpoints are excluded from the pool average and an all-stale pool reads 0.0 and dispatches. That is implemented in #2492, which is through review, and it gives you exactly the behavior you are after here, opt-in. We kept fail-closed as the default because stale metrics can also mean the model server is overloaded or unreachable, and silently dropping the backpressure signal is not a safe default. On the `Filter` comparison: the divergence is deliberate; `Filter` chooses among candidates for a request that has already been admitted, while `Saturation` gates admission.

**Change 2 (200ms -> 5s default):**

> The previous 200ms default was shorter than typical metrics scrape intervals, virtually guaranteeing periodic staleness even under normal operation.

The interval in question here is how often the EPP polls the model servers. The default is 50ms, and 50ms is also the enforced floor (`pkg/epp/server/options.go`), so the 200ms threshold carries a 4x margin at the default cadence and staleness does not occur under normal operation. It appears when a deployment raises the polling interval -- 15s in the #2408 case, which a 5s threshold would not cover either. The fix on that side is restoring `refresh-metrics-interval` toward the default; raising the router-wide threshold instead opens a blind spot for deployments at the default cadence (with continuous batching, an endpoint's KV cache can grow unobserved for the full threshold window).

If your deployment cannot run at a 50ms cadence, `metricsStalenessThreshold` is already configurable per deployment today, and #2492 adds the policy knob on top. One ask for future PRs: link a GitHub issue rather than an internal tracker so the discussion is visible here. In this case #2408 already had the agreed design and an implementation in review.

Issue #2427, opened 2026-08-17
[P/D] Always Disagg Decider Runs Prefill work on Decode if no Prefillers are READY


robertgshaw2-redhat:
**What happened**:
- I configured EPP with always disagg 
- My prefill worker did not become READY [I messed up the readiness probe)
- Requests were processed all on decode

**What you expected to happen**:
- I would expect the request to fail if there are no prefill workers avialble

**How to reproduce it (as minimally and precisely as possible)**:

- router yaml
```yaml

# Deploy:
#   helm upgrade --install kimik3 oci://ghcr.io/llm-d/charts/llm-d-router-standalone --version v0.9.0 -n default -f kimik3/disaggregated/router.values.yaml
router:
  epp:
    replicas: 1
    flags:
      ha-enable-leader-election: false

    # The default pool is too small to fit the EPP, so let it land on a GPU
    # node instead of going Pending.
    tolerations:
      - key: nvidia.com/gpu
        operator: Equal
        value: present
        effect: NoSchedule
    env:
      - name: HF_TOKEN
        valueFrom:
          secretKeyRef:
            name: llm-d-hf-token
            key: HF_TOKEN

    pluginsConfigFile: "plugins.yaml"
    pluginsCustomConfig:
      plugins.yaml: |
        apiVersion: llm-d.ai/v1alpha1
        kind: EndpointPickerConfig
        plugins:
          - type: disagg-headers-handler
          - type: always-disagg-pd-decider
          - type: disagg-profile-handler
            parameters:
              deciderPluginName: always-disagg-pd-decider
          - type: prefill-filter
          - type: decode-filter
          - type: approx-prefix-cache-producer
          - type: inflight-load-producer
          - type: prefix-cache-affinity-filter
            parameters:
              peakPrefillThroughput: 10000
          - type: token-load-scorer
          - type: active-request-scorer
          - type: max-score-picker
        schedulingProfiles:
        # Prefill does the prompt pass: keep prefix groups on cache-warm pods
        # (affinity filter) and pick by queued prefill token load.
        - name: prefill
          plugins:
          - pluginRef: prefill-filter
          - pluginRef: prefix-cache-affinity-filter
          - pluginRef: token-load-scorer
          - pluginRef: max-score-picker
        # Decode is bound by concurrent in-flight requests, not prefill
        # throughput: pick the least-busy endpoint.
        - name: decode
          plugins:
          - pluginRef: decode-filter
[...]

PR #2549, opened 2026-08-25
test(requestcontrol): give each Admit sub-test its own reqCtx

### LINKED ISSUE #2543: Data race in TestFlowControlAdmissionController_Admit under -race (parallel sub-tests share reqCtx)
**What happened**:

`TestFlowControlAdmissionController_Admit` in `pkg/epp/requestcontrol/admission_test.go` fails intermittently under `go test -race` on `main` since #2490 (`7e5bdc184`, merged 2026-08-25). Observed on CI run https://github.com/llm-d/llm-d-router/actions/runs/32826039629 (after PR #2435 was rebased onto main).

Race detector output points at `pkg/epp/requestcontrol/admission.go:185-186`, both goroutines running inside the sub-test closure at `admission_test.go:337`:

```
WARNING: DATA RACE
Write at 0x00c000a4c0e8 by goroutine 353:
  pkg/epp/requestcontrol.(*FlowControlAdmissionController).Admit()
      /app/pkg/epp/requestcontrol/admission.go:185
  pkg/epp/requestcontrol.TestFlowControlAdmissionController_Admit.func1()
      /app/pkg/epp/requestcontrol/admission_test.go:337

Previous write at 0x00c000a4c0e8 by goroutine 355:
  (same stack)
```

Root cause: the test declares a single `reqCtx` before the sub-test loop and shares it across every `t.Run(..., func(t){ t.Parallel(); ... })`. #2490 added two writes to `reqCtx.FlowControlQueueDuration` / `reqCtx.FlowControlAdmitted` on the dispatched path (`admission.go:184-187`). The `sheddable_dispatched` and `non_sheddable_dispatched` sub-cases both hit that path in parallel, so they race on the shared struct.

`go test -race` on `pkg/epp/requestcontrol/...` passes deterministically.

```
git checkout main
go test -race -count=10 -run 'TestFlowControlAdmissionController_Admit$' ./pkg/epp/requestcontrol/...
```

The race is timing-dependent, so a single run can pass. `-count=10` (or varying GOMAXPROCS) reliably surfaces it.

Suggested fix (test-side, no production change): give each sub-test its own `reqCtx`, matching the pattern already used by the sibling test `TestFlowControlAdmissionController_StampsQueueDuration` at `admission_test.go:356-376` (a `newReqCtx()` helper called inside each `t.Run` closure).

An alternative would be to have `Admit` avoid mutating a caller-owned struct on the dispatched path, but the new response-header wiring needs those fields stamped on `reqCtx`, so the test-side fix is the smaller correct change.

Blast radius: this fires on any PR whose branch rebases onto current main under `-race`, i.e. every PR opened against the coordinator, sidecar, or router codebases right now.

**Environment**:
- Go 1.26.6 (builder image `ghcr.io/llm-d/llm-d-builder:dev`)
- CI job: `test` in workflow `.github/workflows/ci-pr-checks.yaml` (`make test-unit-epp`)
- Head of `main` at the time of report: `ddce73f9` (contains offending commit `7e5bdc184`)

elinacse:
**What type of PR is this?**
/kind bug

TestFlowControlAdmissionController_Admit declared a single reqCtx before
its parallel sub-test loop and shared it across every t.Run(..., func(t)
{ t.Parallel(); ... }). FlowControlAdmissionController.Admit stamps
reqCtx.FlowControlQueueDuration and reqCtx.FlowControlAdmitted on the
dispatched path (added in #2490), so the sheddable_dispatched and
non_sheddable_dispatched sub-cases raced on those writes under
`go test -race`.

Each sub-test now constructs its own reqCtx inside the t.Run closure,
matching how a real request has its own RequestContext. No production
code changes.

Verified by reproducing the race on the pre-fix code with
`go test -race -count=10 -run 'TestFlowControlAdmissionController_Admit$' ./pkg/epp/requestcontrol/...`,
then confirming it's gone after the fix (also ran -race -count=30, and
-race across all of ./pkg/...).

Fixes #2543

**Release note**:
```release-note
NONE


**Test plan** (if you want to include one in the PR body):
```markdown
**Test plan**:
- [x] `go test -race -count=30 ./pkg/epp/requestcontrol/...` passes deterministically
- [x] `go test -race ./pkg/...` passes across the full package tree


LukeAVanDrie:
The diff here is empty as this branch's only commit (`d1fa975`) is the #2427 prefill fix, which already merged via #2521.

The reqCtx change described in the PR body was never pushed, so it looks like the PR was opened from the old #2427 branch by mistake. Could you commit the test fix on a fresh branch off main and either force-push it here or open a new PR? Thanks for de-flaking this test suite!

elevran:
Closing as superseded by #2579
@elinacse please note that the head ref points at a stale #2427 branch, you probably meant to open against fix/2543-flowcontrol-admit-test-race?
At any rate, #2543 was fixed by #2579 and is already resolved on main.

PR #2550, opened 2026-08-25
tracing: support standard sampler values and validate the argument

### LINKED ISSUE #2530: tracing: support the standard OTEL_TRACES_SAMPLER values and validate the sampler argument

Handle the standard `OTEL_TRACES_SAMPLER` values in `InitTracing` (`pkg/common/observability/tracing/telemetry.go`), and stop substituting `0.1` for an unparseable `OTEL_TRACES_SAMPLER_ARG` without a signal.

Only `parentbased_traceidratio` is matched today. `always_on`, `always_off`, `traceidratio`, `parentbased_always_on` and `parentbased_always_off` all fall through to the else branch and are served a `ParentBased(TraceIDRatioBased(0.1))` sampler. Map each
to its SDK constructor: `AlwaysSample`, `NeverSample`, `TraceIDRatioBased`, and the `ParentBased` wrappers around the first two.

Separately, a `strconv.ParseFloat` failure replaces the operator's value with `0.1` silently, and a ratio outside `[0, 1]` is not checked at all. Both should reach the error handler the way an unsupported sampler type already does.

Selection belongs in a `newSampler` helper, matching `newResource` and `localCollectorOptions`, so the mapping is testable without starting a tracer provider.

Two things to settle before implementing:

- **`jaeger_remote` and `parentbased_jaeger_remote` are out of scope.** They require `go.opentelemetry.io/contrib/samplers/jaegerremote`, which is not in the build graph. Either value should get the unsupported-type warning.

- **The default stays `parentbased_traceidratio` with `0.1`.** The specification default is `parentbased_always_on`; adopting it would take every deployment that does not set the variable from 10% to 100% sampling. Keeping the current default is deliberate and worth a comment rather than a silent divergence from the specification.

**Why is this needed**:

An operator who selects a standard sampler value gets 10% ratio sampling instead, and an operator who sets `OTEL_TRACES_SAMPLER_ARG` with a typo gets 10% instead of the value they typed. Neither is reported where it is decided, so the sampling rate in effect does not match the configuration under review. `always_off` has the clearest consequence: it is how an operator keeps tracing configured but dormant, and it currently samples one request in ten. The chart surfaces both variables under `router.tracing.sampling`, so these are values operators are expected to set.

**Acceptance criteria**:

- `always_on`, `always_off`, `traceidratio`, `parentbased_always_on`, `parentbased_always_off` and `parentbased_traceidratio` each produce the corresponding sampler.
- An unsupported sampler value falls back to the default and is reported through the error handler.
- An unparseable `OTEL_TRACES_SAMPLER_ARG` is reported rather than silently replaced.
- A ratio outside `[0, 1]` is rejected the same way.
- `OTEL_TRACES_SAMPLER_ARG` is ignored, without a warning, for samplers that take no argument.
- Unset variables still yield `parentbased_traceidratio` at `0.1`.
- Table-driven tests over the value set above, asserting on the sampler's `Description()`.

Part of #1632.

satyamg1620:
**What type of PR is this?**
/kind bug

`InitTracing` matched only `parentbased_traceidratio`, so `always_on`, `always_off`, `traceidratio` and both `parentbased_always_*` values fell through to a 10% ratio sampler. `always_off` is the sharpest case: it is how tracing is kept configured but dormant, and it sampled one request in ten.

Invalid arguments were also absorbed silently. An unparsable `OTEL_TRACES_SAMPLER_ARG` became `0.1` with no diagnostic, and a ratio outside `[0, 1]` reached the SDK, which clamps it - `10` meant as ten percent sampled every request. Both are now reported through the error handler and fall back to the documented default.

Selection moves into `newSampler`, alongside the existing `newResource` and `localCollectorOptions` helpers. A rejected value returns an error *with* a usable sampler, mirroring `newResource`: sampling is a runtime knob, so a typo in it should
not stop process startup.

The default stays `parentbased_traceidratio` at `0.1`. The specification default is `parentbased_always_on`, which would move every deployment that sets neither variable from 10% to 100% sampling; that belongs in its own change. `jaeger_remote` and `parentbased_jaeger_remote` report as unsupported, since the sampler they need is not in the build graph.

Fixes #2530

**Test plan**:

Unit (`TestNewSampler`, table-driven over 15 cases):
- [x] All six supported sampler values produce their corresponding sampler
- [x] An unsupported or empty value is reported and falls back to the default
- [x] An unparsable argument, and one outside `[0, 1]`, are reported rather than substituted
- [x] The argument is ignored, without a warning, by samplers that take none
- [x] Bounds `0` and `1` are accepted; unset variables still yield `parentbased_traceidratio` at `0.1`
- [x] A rejected configuration does not fail startup (`TestInitTracingSurvivesRejectedSampler`)

Verified on a kind dev stack, 20 requests per case, counting entry spans in the EPP log:
- [x] `always_on` samples 20/20 (1/20 before this change), `always_off` samples 0/20
- [x] `OTEL_TRACES_SAMPLER_ARG=10` warns and falls back to 0.1 (sampled 20/20 silently before)
- [x] `parentbased_traceidratio` at `1.0` is unchanged, confirming no regression for the value the chart ships

```release-note
Tracing now supports the standard OTEL_TRACES_SAMPLER values (always_on, always_off, traceidratio, parentbased_always_on, parentbased_always_off) in addition to parentbased_traceidratio. An unsupported sampler, an unparsable OTEL_TRACES_SAMPLER_ARG, or a ratio outside [0, 1] is now logged and falls back to parentbased_traceidratio at 0.1 instead of being applied silently.

### DIFF
--- a/pkg/common/observability/tracing/telemetry.go
+++ b/pkg/common/observability/tracing/telemetry.go
@@ -63,26 +63,9 @@ func InitTracing(ctx context.Context, logger logr.Logger, defaultServiceName str
 		return nil, err
 	}
 
-	// Go SDK doesn't have an automatic sampler, handle manually
-	samplerType, ok := os.LookupEnv("OTEL_TRACES_SAMPLER")
-	if !ok {
-		samplerType = "parentbased_traceidratio"
-	}
-	samplerARG, ok := os.LookupEnv("OTEL_TRACES_SAMPLER_ARG")
-	if !ok {
-		samplerARG = "0.1"
-	}
-
-	sampler := sdktrace.ParentBased(sdktrace.TraceIDRatioBased(0.1))
-	if samplerType == "parentbased_traceidratio" {
-		fraction, err := strconv.ParseFloat(samplerARG, 64)
-		if err != nil {
-			fraction = 0.1
-		}
-
-		sampler = sdktrace.ParentBased(sdktrace.TraceIDRatioBased(fraction))
-	} else {
-		loggerWrap.Handle(fmt.Errorf("unsupported sampler type: %s, fallback to parentbased_traceidratio with 0.1 Ratio", samplerType))
+	sampler, err := newSampler()
+	if err != nil {
+		loggerWrap.Handle(fmt.Errorf("trace sampler configuration degraded: %w", err))
 	}
 
 	opt := []sdktrace.TracerProviderOption{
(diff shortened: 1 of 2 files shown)


satyamg1620:
@ahg-g Can you please review this PR

ahg-g (approved):
nice

PR #2559, opened 2026-08-26
Route cached follow-ups by remaining prefill work

### LINKED ISSUE #2540: Route cached follow-ups by remaining prefill work, not total context length

A way for `context-length-aware` to classify a request by its remaining prefill work instead of its total context length, using the transferable-prefix information `p2p-source-producer` already computes before scheduling.

## Problem

Deployments that split prefill workers into short-work and long-work pools (to avoid head-of-line blocking and to enable prefill-specific tuning such as prefill context parallel on the long pool) route with `context-length-aware` on total context length. Long-context follow-ups are classified by total length even when most of the context is available through P2P KV-cache sharing:

- Initial 100K request runs on P-long.
- Follow-up contains 101K total tokens.
- P-long has a transferable 100K-token prefix.
- Only 1K tokens require new prefill work.

Today `context-length-aware` classifies the follow-up as 101K, so it remains on P-long and can wait behind cold long-prefill requests. It should be classified as approximately 1K and move to P-short, with P-short pulling the cached 100K KV from P-long over the P2P tier and prefilling only the new tokens.

## Proposal

Expose a reusable-prefix token floor from `p2p-source-producer` during its existing pre-scheduling `Produce` phase.

1. `p2p-source-producer.Produce` selects a transferable cache source and publishes `ReusablePrefixTokens`, a floor `G` on the tokens every valid destination can reuse either locally or by pulling from the sampled source.
2. `context-length-aware`, opted in via `reusableTokensProducerName`, routes a known positive prompt on `routingLength = max(1, totalTokens - G)`.
3. The scheduler selects P-short based on the remaining prefill work.
4. The existing `p2p-source-producer.PreRequest` phase adds the source header after P-short is selected.

This does not change when or how the P2P transfer happens. It exposes information from the existing pre-scheduling phase so the prefiller choice can account for the planned transfer.

```yaml
plugins:
  - type: precise-prefix-cache-producer
    name: precise-cache
    parameters:
      tokenProcessorConfig:
        blockSizeTokens: 64
      speculativeIndexing: false
  - type: p2p-source-producer
    name: p2p-cache-source
    parameters:
      prefixMatchInfoProducerName: precise-cache
      minCachedTokenDelta: 1
      prefillProfileName: prefill
  - type: context-length-aware
    name: prefill-work-router
    parameters:
      label: llm-d.ai/prefill-work-range
      enableFiltering: true
      reusableTokensProducerName: p2p-cache-source
```

## Design notes

**Floor and delta.** The floor must hold for every candidate the scheduler may pick, so the positive published floor is `G = S - minCachedTokenDelta + 1`, where `S` is the sampled source's pull-servable (CPU-tier) prefix. When that value is not positive, no floor is published. The producer publishes a scheduling floor only when the sampled source provides confirmed per-tier cache data. A tierless approximate source can still be used by the existing `PreRequest` header-selection path, but it cannot drive hard P-short routing. The effective classification when a floor is published is therefore `newTokens + minCachedTokenDelta - 1`, so the P-short work-range ceiling has to absorb the delta: a deployment that tunes `minCachedTokenDelta` for pull economics (order 10K tokens on measured tiers) needs a P-short range like `0-16384`, or the follow-up classifies back into P-long. This relationship needs to be stated in the plugin docs.

**Baseline, not the end state.** Static ranges migrate every eligible cached follow-up unconditionally, so at low load this pays a pull where staying on an idle holder would be a local hit. That is acceptable for the target workload (sustained long-prefill contention) and keeps the baseline simple. Follow-up work, tracked separately: a load-gated variant that keeps follow-ups on the cache holder until its prefill queue saturates (in the spirit of `prefix-cache-affinity-filter`'s `peakPrefillThroughput`), and a per-endpoint routing length that is exact instead of a request-wide floor.

**P-short becomes the session home.** After the first migration the freshest full-prefix copy lives on P-short, so P-short pods need the CPU offload tier and KV-events publishing like any pull source; otherwise every subsequent turn re-pulls the full context from P-long. P-short also needs full logical `max-model-len` and KV capacity sized for full-length sequences - its prefill work is short, its memory footprint is not.

**Degradation properties.** Requests with no confirmed tier-aware pullable source, or a follow-up arriving before offload events are indexed, fall back to total-length routing, i.e. current behavior. A stale positive cache entry is different: it can route the request to P-short and then miss during transfer, so the serving stack still needs a recompute fallback and monitoring for transfer failures. Two paths need documenting and monitoring: with `enableFiltering: true` an arbitrary work-range label gap can leave an empty prefill candidate set and degrade to a decode-only prefill under `disagg-profile-handler`, and a failed pull leaves the full prefill on P-short (`--long-prefill-token-threshold` on the engine remains the backstop).

**Why is this needed**:

Head-of-line blocking in split prefill pools defeats the QoS the split exists to provide: requests with a large reusable prefix wait behind cold long prefills because the router only sees total context length. Agentic workloads are the concrete case - large dynamic prefixes with short per-turn additions, where nearly all of each follow-up's context is transferable. Routing on remaining prefill work moves those follow-ups to the short pool and lets P2P KV-cache sharing supply the prefix, keeping the long pool for genuinely long cold work (where prefill context parallel applies).

Implementation: #2559. Measured results are in the comments below.

nilig:
**What type of PR is this?**
/kind feature

Routes long-context follow-ups by their remaining prefill work instead of
their total context length, so a cached follow-up moves to a short-work
prefiller and pulls its prefix over the P2P tier instead of queueing behind
cold long prefills.

Mechanism: `p2p-source-producer.Produce` publishes a name-bound
`ReusablePrefixTokens` request attribute - a floor `G = S - D + 1` on the
tokens every valid destination can reuse either locally or by pulling from
the sampled source (`S` = the source's confirmed CPU-tier prefix, `D` =
`minCachedTokenDelta`). `context-length-aware` gains an opt-in
`reusableTokensProducerName` parameter and routes on
`max(1, totalTokens - G)` against a dedicated work-range label. The P2P
source header name and `host:port` value format are unchanged. Its emission
gate now ignores speculative cache entries on the computing endpoint, so
unconfirmed blocks cannot suppress a required pull. This gate correction
applies whenever `p2p-source-producer` uses tier-aware cache data,
independently of `reusableTokensProducerName`. The remaining-work scheduling
behavior is opt-in: without `reusableTokensProducerName`,
`context-length-aware` behaves exactly as before.

Soundness of the floor against the header gate:

- The computing-side cached count excludes speculative index entries
  (`ConfirmedCachedBlockCount` on `PrefixCacheMatchInfo`, populated by the
  precise producer), so a phantom speculative overlap cannot suppress the
  pull after the floor routed the request.
- The floor publishes only from sources with per-tier data; tier-less
  producers keep their header behavior but never publish a floor.
- Routing length is clamped to a minimum of 1 so work-range labels with a
  nonzero minimum cannot empty the candidate pool.
- The factory rejects `reusableTokensProducerName` combined with the default
  context-length label, which describes logical capacity, not prefill work.

Also included, delineated for review: the request attribute store is
eagerly initialized before producer goroutines launch
(`InitializeAttributeStore` in `pkg/epp/framework/interface/scheduling`,
called from the producer executor), removing a lazy-init data race with a
timeout-abandoned producer writing concurrently with scheduling reads; and
`getContextLength` memoizes its routing length per request so Filter and
Score cannot disagree if a floor lands mid-cycle. Both carry regression
tests (the race test fails under `-race` without the fix).

Validated on two clusters; full data, figures, exact manifests, and a
deploy guide:
https://github.com/nilig/llm-d/blob/findings/p2p-context-migration-glm-c64/findings/p2p-context-migration-glm/RESULTS.md

- Llama-3.1-8B (single-GPU workers): exact mechanism counters end to end;
  under a real capacity queue, follow-up TTFT 23.1s -> 3.0s; idle parity
  when migration is gated by the existing `prefix-cache-affinity-filter`
  (config-only composition, no new policy code).
- GLM-5.2-FP8 (three DP8 H200 engines, AgentX agentic traces, C64,
  counterbalanced ABBA blocks): cached-follow-up TTFT p50 -53%, fleet p90
  -31%, +7% turns completed. With `--long-prefill-token-threshold` live on
  all prefill workers as a stronger baseline, the engine mitigation recovers
  ~8% and remaining-work routing adds -33% on top - the measured queueing is
  KV-capacity-bound, so the mechanisms compose.

Fixes #2540

**Test plan**:

- [x] Unit tests for the floor publication (per-tier gating, delta
  arithmetic, no-source absence, instance-name key binding) and for the
  scorer's subtraction (filter and score modes, missing/wrong-typed
  attribute fallback, clamping, factory validation)
- [x] Lifecycle test: cold request stays on the long-work class, cached
  follow-up routes to the short-work class and the source header points at
  the holder
- [x] Regression tests: speculative-only destination still receives the
  header; cancelled producer context stores no attributes; late attribute
  write is race-free under `-race`; routing-length snapshot ignores a
  late-arriving floor
- [x] `go test -race` on all touched packages; `make presubmit` and the full
  unit suite green at the branch tip and at the merge point with current
  `main`
- [x] End-to-end on two clusters (counters exact to the token; see the
  results link above)

**Release note**:
```release-note
The `context-length-aware` scorer can route requests by remaining prefill
work: the new optional `reusableTokensProducerName` parameter subtracts the
P2P-transferable prefix published by a named `p2p-source-producer`, so
cached long-context follow-ups can be routed to short-work prefill workers
and pull their prefix over the P2P tier.

The `p2p-source-producer` header gate now considers only confirmed
destination cache blocks when tier data is available. With speculative
indexing enabled, unconfirmed blocks no longer suppress a required P2P pull.
```

### DIFF
--- a/pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
+++ b/pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
@@ -24,9 +24,19 @@ import (
 	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
 	approxprefixconstants "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/constants"
+	p2psourceconstants "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/constants"
 )
 
 var PrefixCacheMatchInfoDataKey = plugin.NewDataKey("PrefixCacheMatchInfoDataKey", approxprefixconstants.ApproxPrefixCachePluginType)
+
+// ReusablePrefixTokensDataKey identifies the request-wide reusable prefix
+// token floor published by a p2p-source-producer instance.
+var ReusablePrefixTokensDataKey = plugin.NewDataKey("ReusablePrefixTokensDataKey", p2psourceconstants.P2PSourcePluginType)
+
+// ReusablePrefixTokens is a floor on the prompt tokens every valid destination
+// can reuse either from confirmed local cache or by pulling confirmed CPU-tier
+// blocks from the sampled source.
+type ReusablePrefixTokens int
 
 // SpeculativeTierKey is the CachedBlocksByTier key for speculative index
 // entries, which carry no engine-reported device tier.
@@ -47,6 +57,10 @@ type PrefixCacheMatchInfo struct {
 	// the prefix-based PD decider) get an accurate cached-token figure rather
 	// than a tier-attenuated one. Defaults to matchBlocks when not set.
 	cachedBlockCount int
+	// unweighted count of contiguous cached prefix blocks confirmed by the
+	// engine. Speculative index entries are excluded, while confirmed blocks
+	// can move between device tiers across the prefix.
+	confirmedCachedBlockCount int
 	// per device tier, the unweighted count of contiguous cached prefix blocks
 	// the endpoint holds in that tier, from the first block until the first
 	// block missing from that tier. A block held in several tiers counts once
@@ -96,6 +110,20 @@ func (p *PrefixCacheMatchInfo) CachedBlockCount() int {
 	return p.cachedBlockCount
 }
 
+// WithConfirmedCachedBlockCount sets the unweighted contiguous count of
+// confirmed, non-speculative cached blocks and returns the receiver for
+// chaining.
+func (p *PrefixCacheMatchInfo) WithConfirmedCachedBlockCount(confirmedCachedBlockCount int) *PrefixCacheMatchInfo {
+	p.confirmedCachedBlockCount = confirmedCachedBlockCount
+	return p
+}
+
+// ConfirmedCachedBlockCount returns the unweighted contiguous count of
+// confirmed, non-speculative cached blocks on the endpoint.
+func (p *PrefixCacheMatchInfo) ConfirmedCachedBlockCount() int {
+	return p.confirmedCachedBlockCount
+}
+
 // WithCachedBlocksByTier sets the per-device-tier contiguous cached-block
 // counts and returns the receiver for chaining. Takes ownership of the map;
 // the caller must not mutate it after the call.
@@ -113,11 +141,12 @@ func (p *PrefixCacheMatchInfo) CachedBlocksByTier() map[string]int {
 
 func (p *PrefixCacheMatchInfo) Clone() fwkdl.Cloneable {
 	clone := &PrefixCacheMatchInfo{
-		matchBlocks:        p.matchBlocks,
-		totalBlocks:        p.totalBlocks,
-		blockSizeTokens:    p.blockSizeTokens,
-		cachedBlockCount:   p.cachedBlockCount,
-		cachedBlocksByTier: maps.Clone(p.cachedBlocksByTier),
+		matchBlocks:               p.matchBlocks,
+		totalBlocks:               p.totalBlocks,
+		blockSizeTokens:           p.blockSizeTokens,
+		cachedBlockCount:          p.cachedBlockCount,
+		confirmedCachedBlockCount: p.confirmedCachedBlockCount,
+		cachedBlocksByTier:        maps.Clone(p.cachedBlocksByTier),
 	}
 	if p.mm != nil {
 		clone.mm = ptr.To(*p.mm)
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/producer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/producer.go
@@ -14,10 +14,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package p2psource emits the KV cache source header: a candidate pod
-// within one block of the most cached prefix KV blocks for the request, for
-// the routing sidecar to pull from over the P2P connector instead of
-// recomputing them.
+// Package p2psource selects a peer within one block of the most cached prompt
+// prefix, publishes a request-wide reusable-prefix floor for scheduling, and
+// emits the selected peer in the KV cache source header after scheduling.
 package p2psource
 
 import (
@@ -50,6 +49,14 @@ const (
 	defaultMinCachedTokenDelta = 1
 )
 
+// ReusablePrefixTokensDataKey identifies the request-wide reusable prefix
+// token floor published by a p2p-source-producer instance.
+var ReusablePrefixTokensDataKey = plugin.NewDataKey("ReusablePrefixTokensDataKey", PluginType)
+
+// ReusablePrefixTokens is a floor on the prompt tokens every valid destination
+// can reuse either locally or by pulling from the sampled source.
+type ReusablePrefixTokens int
+
 // Config configures the p2p-source-producer.
 type Config struct {
 	// PrefixMatchInfoProducerName is the name of the data producer that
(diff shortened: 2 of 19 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/producer.go
```diff
@@ -303,25 +325,28 @@ const cpuDeviceTier = "cpu"
...
-// pull whatever their tier, so the computing side stays tier-blind.
+// cachedTokenCount returns the endpoint's confirmed cached prompt tokens
+// across all tiers, or 0 when its PrefixCacheMatchInfo is absent. Producers
+// without tier data keep their trusted tierless count.
 func (p *Producer) cachedTokenCount(ep scheduling.Endpoint) int {
 	info := p.matchInfo(ep)
 	if info == nil {
 		return 0
 	}
+	if info.CachedBlocksByTier() != nil {
```
bongwoobak: This changes the header gate for existing deployments, not only the new opt-in path. With speculativeIndexing on, speculative entries on the computing pod used to count toward computingCached (main reads CachedBlockCount here, and matchedBlockCount has no Speculative filter) and suppress the header. Now they don't, so a pod that just had a same-prefix request routed to it gets a pull pointed at it while those blocks are still unconfirmed. TestPreRequest_SpeculativeComputingCacheDoesNotSuppressHeader pins exactly that, it fails against the old cachedTokenCount.

I think the new behavior is the right one, the soundness bullet in the description already depends on it. But the description also says PreRequest header emission is unchanged, and the release note only mentions the scorer parameter. Could you add a line to the release note? The gate moves for anyone running p2p-source-producer with speculative indexing today, regardless of reusableTokensProducerName.

pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
```diff
@@ -96,6 +110,20 @@ func (p *PrefixCacheMatchInfo) CachedBlockCount() int {
 	return p.cachedBlockCount
 }
 
+// WithConfirmedCachedBlockCount sets the unweighted contiguous count of
+// confirmed, non-speculative cached blocks and returns the receiver for
+// chaining.
+func (p *PrefixCacheMatchInfo) WithConfirmedCachedBlockCount(confirmedCachedBlockCount int) *PrefixCacheMatchInfo {
```
bongwoobak, line 119: confirmedCachedBlockCount stays 0 for a producer that never calls this, while cachedBlockCount defaults to matchBlocks in the constructor. approx and burst don't set it, so the only thing keeping the computing side from reading 0 for them is the CachedBlocksByTier() != nil branch in p2psource (producer.go:347), and a producer that publishes tier data without knowing about this setter would fire the header on every request.

How about defaulting it to the cached count, with WithCachedBlockCount setting both and this one overriding afterwards? Then cachedTokenCount can just read ConfirmedCachedBlockCount and drop the byTier branch.

pkg/epp/framework/plugins/requestcontrol/dataproducer/p2psource/producer.go
nilig: Good catch. The header name and `host:port` value format are unchanged, but ignoring speculative destination entries changes the emission gate independently of `reusableTokensProducerName`. I updated the description and release note to state that behavior.

pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
nilig: Agreed. `ConfirmedCachedBlockCount` now defaults to `CachedBlockCount`, `WithCachedBlockCount` updates both, and `WithConfirmedCachedBlockCount` remains the explicit override. `cachedTokenCount` now reads the confirmed count directly. The new tests cover the defaults and a tier-aware producer that does not set a separate confirmed count.

vMaroon (approved):
great

PR #2563, opened 2026-08-26
fix(epp): preserve endpoint attributes in data-parallel profile

### LINKED ISSUE #2411: Profile handlers rebuild target endpoints with an empty attribute map before `PreRequest`
`DataParallelProfileHandler.ProcessResults` and `PDProfileHandler.ProcessResults` rebuild their
targets to override the port:

```go
targetEndpoint := scheduling.NewEndpoint(newMetadata, target.GetMetrics().Clone(), nil)
```

Metadata and metrics are cloned; attributes are passed as `nil`, and `NewEndpoint` turns nil into
an empty map. `ProcessResults` runs inside `Schedule`, before the director calls `PreRequest` plugins. So, under
these two handlers every `PreRequest` plugin sees endpoints with no attributes, while under `single` it sees them populated. The map is empty rather than absent, so there is no error and no way to tell the two cases apart at runtime.

Is that intended? If so, it's worth documenting that PreRequest can't rely on endpoint attributes.
If not, passing `target.Clone()` instead of `nil` carries them through.

Hit in #2371, worked around there by moving the value into `PluginState`.
Ready to open a PR for either the fix or the doc change, whichever you prefer.

yuanhuizang-cloud:
**What type of PR is this?**

/kind bug

Preserves endpoint attributes when `DataParallelProfileHandler.ProcessResults`
rebuilds target endpoints to override the primary port.

Previously, the rebuilt endpoint passed `nil` attributes to `NewEndpoint`,
which caused `PreRequest` plugins to receive an empty attribute map.

This change carries the original endpoint attributes forward while keeping the
existing metadata port rewrite behavior unchanged.

Fixes #2411 

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/profilehandler/dataparallel/dp_profile_handler.go
+++ b/pkg/epp/framework/plugins/scheduling/profilehandler/dataparallel/dp_profile_handler.go
@@ -124,7 +124,7 @@ func (h *ProfileHandler) ProcessResults(_ context.Context, request *scheduling.I
 	for _, target := range profileResult.TargetEndpoints {
 		newMetadata := target.GetMetadata().Clone()
 		newMetadata.Port = h.primaryPort
-		targetEndpoint := scheduling.NewEndpoint(newMetadata, target.GetMetrics().Clone(), nil)
+		targetEndpoint := scheduling.NewEndpoint(newMetadata, target.GetMetrics(), target.Clone())
 		newResult.TargetEndpoints = append(newResult.TargetEndpoints, targetEndpoint)
 	}
 	modifiedResults := map[string]*scheduling.ProfileRunResult{singleProfileName: &newResult}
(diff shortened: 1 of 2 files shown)


pkg/epp/framework/plugins/scheduling/profilehandler/dataparallel/dp_profile_handler.go
```diff
@@ -124,7 +124,7 @@ func (h *ProfileHandler) ProcessResults(_ context.Context, request *scheduling.I
 	for _, target := range profileResult.TargetEndpoints {
 		newMetadata := target.GetMetadata().Clone()
 		newMetadata.Port = h.primaryPort
-		targetEndpoint := scheduling.NewEndpoint(newMetadata, target.GetMetrics().Clone(), nil)
+		targetEndpoint := scheduling.NewEndpoint(newMetadata, target.GetMetrics().Clone(), target.Clone())
```
ahg-g: `NewEndpoint` already calls clone: https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/interface/scheduling/types.go#L120

yuanhuizang-cloud: Thanks, updated to rely on NewEndpoint for cloning the metrics.

yankay:
Tested in a simulated environment. Everything worked as expected.

/lgtm

ahg-g (approved):

PR #2565, opened 2026-08-26
feat(flowcontrol): add program-score ordering policy

### LINKED ISSUE #2548: [Flow Control] Add Program-aware ScoringOrderingPolicy

Issue #1634 adds `ScoringOrderingPolicy`, which enables ordering policies where the score for a queue item changes with time.

Proposing the first `ScoringOrderingPolicy` which makes decisions based on program/agent state. Requests are scored based on number of turns taken by the program and the number of tokens the agent has consumed, similar to the current program-aware plugin (`pkg/epp/framework/plugins/flowcontrol/fairness/program-aware`). #2524's session state will be exported for this plugin to consume.

The plugin is designed to be extensible to support more scoring signals. For example, a future signal could score requests by longest-prefix-match so cache-warm requests are dispatched ahead of cold ones.

Related: #2524. Will attach a PR soon.

**Why is this needed**:

Today's program-aware fairness plugin maps each agent identity 1:1 to a `FairnessID`, one queue per agent. A tenant running 5 agents gets 5 queues and 5x the dispatch share of a tenant running 1 agent. Any program-aware scoring strategy added to this ordering policy inherits that same granularity question and needs it resolved
first, not re-litigated per strategy. Scoring within a single queue avoids this, dispatch share no longer depends on how many queues a tenant's agents happen to occupy.

elinacse:
Adds ScoringOrderingPolicy, an OrderingPolicy extension for policies whose Less result changes purely with elapsed time. The FlowController periodically re-heapifies queues using such a policy (reusing the existing cleanup-sweep goroutine and worker pool) so a time-decayed score surfaces at dispatch time even for an item that is never itself added or removed; plain OrderingPolicy implementations are unaffected.

Adds program-score-ordering-policy, the first ScoringOrderingPolicy: it orders requests within a single shared queue by the originating program/agent's decayed turns taken and tokens consumed, dispatching the least-served program first. This avoids the queue-per-agent granularity problem of program-aware-fairness, where a tenant running more agents gets a proportionally larger dispatch share purely by occupying more queues. Session state (turns, tokens) is tracked internally by the plugin via the existing PreRequest/ResponseBody hooks, independent of #2524.

**What type of PR is this?**

Fixes #2548

### DIFF
--- a/cmd/epp/runner/runner.go
+++ b/cmd/epp/runner/runner.go
@@ -84,6 +84,7 @@ import (
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/fairness/roundrobin"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/ordering/edf"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/ordering/fcfs"
+	programscore "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/ordering/program-score"
 	slodeadline "github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/ordering/slodeadline"
 	"github.com/llm-d/llm-d-router/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency"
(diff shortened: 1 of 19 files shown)


praveingk:
@elinacse We are still evolving this proposal, and have to do a lot of benchmarks before changing the program-aware plugin to ordering yet. 
@D-Sai-Venkatesh is working on this. Please reach out to him. Happy to collaborate on this.

PR #2566, opened 2026-08-26
tracing: default the trace exporter to otlp instead of console

### LINKED ISSUE #2564: tracing: stop defaulting the trace exporter to console pretty print

Change the default of `OTEL_TRACES_EXPORTER` in `initTraceExporter` (`pkg/common/observability/tracing/telemetry.go`) away from `console`, and stop building the stdout exporter unconditionally.

`stdouttrace.New(stdouttrace.WithPrettyPrint())` is constructed before the exporter type is read, the type then defaults to `console`, and the exporter is replaced only when the value is `otlp`. The stdout exporter is created even when it is discarded, and any unrecognised value selects console.

Default to `otlp` and construct exactly one exporter, selected by the value. Handle `none`, the specification's value for no exporter, and report an unrecognised value through the error handler instead of falling into console.

This interacts with the loopback fallback added in #2311: with `otlp` as the default, `InitTracing` with no environment set will try to reach a plaintext collector on `localhost:4317` through `localCollectorOptions`. Export is batched and non-blocking, so the failure stays in the background, but it is a change in behavior and belongs in the release note.

**Why is this needed**:

`--tracing` with no further configuration is the first path a developer takes, and console pretty print writes a multi-line JSON block to stdout for every sampled span. In the EPP that is up to eight spans per sampled request, interleaved with the structured logs on the same stream, which makes both unreadable and defeats log collectors that parse a line at a time. `console` remains available as an explicit value.

The chart already sets `OTEL_TRACES_EXPORTER: "otlp"` (`config/charts/routerlib/templates/_deployment.yaml`), so chart-based deployments are unaffected and the change reaches only direct `--tracing` invocations.

**Acceptance criteria**:

- With no `OTEL_TRACES_EXPORTER` set, the OTLP exporter is used.
- `console` still selects the stdout exporter.
- `none` disables span export.
- An unrecognised value is reported through the error handler.
- The stdout exporter is not constructed when it is not selected.
- Release note covering the changed default.

Part of #1632.

satyamg1620:
**What type of PR is this?**
/kind bug

`initTraceExporter` built the stdout exporter before reading the environment, and an unset `OTEL_TRACES_EXPORTER` selected it. Enabling tracing without further configuration pretty printed a multi-line JSON block per span onto the same stream as the structured logs; on the kind dev stack, 10 requests produced 11 such blocks interleaved with the EPP's own output. Any value other than `otlp` - a typo, or a value the SDK supports and this code does not - also landed on console with no diagnostic.

The type is resolved before anything is constructed, exactly one exporter is built, and an unrecognised value is reported through the error handler. `traceExporterType` and `newTraceExporter` split resolution from construction, matching the
`newSampler` / `samplerFraction` pair alongside them: resolution can degrade to the default, construction stays fatal.

`none` registers no span processor at all rather than installing a no-op exporter, so no batch processor, goroutine or queue exists for spans that are never exported. Span creation, sampling and context propagation are untouched, so `none` means "do not export", not "do not trace".

One behaviour change to call out: with `otlp` as the default, an unconfigured process reaches for a plaintext collector on `localhost:4317` through `localCollectorOptions`. Export is batched, so the failure stays in the background - verified on a stack with no collector, where every request still returned 200. The chart sets `OTEL_TRACES_EXPORTER: "otlp"` explicitly, so chart-based deployments are unaffected.

Fixes #2564

**Test plan**:

Unit:
- [x] Unset resolves to `otlp`; `otlp`, `console` and `none` resolve to themselves
- [x] An unrecognised or empty value is reported and falls back to `otlp`
- [x] The stdout exporter is not constructed for the `otlp` type
- [x] `none` leaves spans valid, sampled and propagating
- [x] `InitTracing` with nothing configured delivers a span to a collector bound to the
      loopback default, exercising the default and the loopback fallback together

Verified on a kind dev stack, 10 requests per case, with the exporter variable removed from the deployment rather than changed:
- [x] Unset resolves to `otlp` and writes no span blocks to stdout (`console` and 11 blocks before this change)
- [x] `console` still pretty prints; `none` writes nothing and makes no export attempt
- [x] An unrecognised value warns and falls back to `otlp`
- [x] With an OTLP collector sidecar on the pod's loopback and no exporter configured, spans arrive carrying `service.name: llm-d-epp` while stdout stays clean
- [x] With no collector listening, export fails in the background and all requests return 200

```release-note
The default trace exporter is now otlp instead of console. Enabling tracing without setting OTEL_TRACES_EXPORTER exports spans to a collector rather than pretty printing them to stdout, and with no collector configured the export fails in the background without affecting request handling. OTEL_TRACES_EXPORTER=none disables span export while leaving span creation and context propagation intact, and an unrecognised value is logged and falls back to otlp.

### DIFF
--- a/pkg/common/observability/tracing/telemetry.go
+++ b/pkg/common/observability/tracing/telemetry.go
@@ -57,11 +57,12 @@ func InitTracing(ctx context.Context, logger logr.Logger, defaultServiceName str
 		loggerWrap.Handle(fmt.Errorf("%s: %v", "build trace resource degraded", err))
 	}
 
-	traceExporter, err := initTraceExporter(ctx, logger)
+	exporterType, err := traceExporterType()
 	if err != nil {
-		loggerWrap.Handle(fmt.Errorf("%s: %v", "init trace exporter failed", err))
-		return nil, err
+		loggerWrap.Handle(fmt.Errorf("trace exporter configuration degraded: %w", err))
 	}
+
+	logger.Info("init OTel trace exporter", "type", exporterType)
 
 	sampler, err := newSampler()
 	if err != nil {
@@ -69,9 +70,19 @@ func InitTracing(ctx context.Context, logger logr.Logger, defaultServiceName str
 	}
 
 	opt := []sdktrace.TracerProviderOption{
-		sdktrace.WithBatcher(traceExporter),
 		sdktrace.WithSampler(sampler),
 		sdktrace.WithResource(res),
+	}
+
+	// "none" registers no span processor at all. Spans are still created and
+	// propagated, so instrumented code and context propagation are unaffected.
+	if exporterType != exporterTypeNone {
+		traceExporter, err := newTraceExporter(ctx, exporterType)
+		if err != nil {
+			loggerWrap.Handle(fmt.Errorf("%s: %v", "init trace exporter failed", err))
+			return nil, err
+		}
+		opt = append(opt, sdktrace.WithBatcher(traceExporter))
 	}
 
 	tracerProvider := sdktrace.NewTracerProvider(opt...)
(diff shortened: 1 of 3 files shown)


ahg-g (approved):
great

PR #2570, opened 2026-08-26
feat(epp): add atomic cross-replica coordination API [Safe DisaggSet 1/4]

hasB4K:
**What type of PR is this?**
/kind feature

Adds an atomic `GetOrSet` operation to `CrossReplicaSyncer` for plugins that need one request-level decision across EPP replicas.

Adds `CrossReplicaSyncerConsumer` and loader injection, so those plugins receive the configured syncer. `LocalSyncer` remains the single-replica fallback and owns a fixed expiration for atomic values. Its startup message is emitted once even when several consumers use the fallback.

Most of the plugin don't need to use a `CrossReplicaConsumer`. This only useful when a specific plugin requires some precise syncing mechanism. I added some documentation when plugin authors should use a periodic `CrossReplicaContributor` instead of a precise consumer.

Follow-up to #1921.
Prelude of #2571 and #2301
```release-note
Add atomic GetOrSet support for EPP CrossReplicaSyncer plugins.
```


pkg/epp/framework/interface/datalayer/cross_replica_syncer.go
```diff
@@ -42,16 +43,33 @@ type CrossReplicaSyncer interface {
...
 	CrossReplicaState() CrossReplicaSpec
 }
 
+// CrossReplicaSyncerConsumer receives the configured CrossReplicaSyncer for
+// exact request-level coordination with atomic operations such as GetOrSet.
+// Use it only when a CrossReplicaContributor cannot tolerate periodic
+// synchronization.
+type CrossReplicaSyncerConsumer interface {
+	SetCrossReplicaSyncer(CrossReplicaSyncer) error
+}
```
ahg-g: instead of doing this, make the syncer part of the handle: https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/interface/plugin/handle.go#L60

ahg-g (approved):

PR #2571, opened 2026-08-26
feat(epp): coordinate DisaggregatedSet revision decisions [Safe DisaggSet 2/4]

hasB4K:
**What type of PR is this?**
/kind feature

Adds EPP-side DisaggregatedSet revision coordination. The rollout Screener stores its selected revision with `CrossReplicaSyncer.GetOrSet`, keyed by `x-llm-d-revision-decision-id` and falling back to `x-request-id` when the decision ID is absent.

Strict revision selection remains authoritative. A strict slice selection still coordinates the revision. The Screener fails closed when coordination fails or returns an invalid revision.

**Dependency:** this intentionally depends on #2570, which adds `CrossReplicaSyncer.GetOrSet`, `CrossReplicaSyncerConsumer`, and the local syncer fallback. This PR is a single EPP-only commit directly on `main`, so its CI is expected to fail until #2570 merges.

#2572 adds the coordinator integration that generates one decision ID and sends it to E/P/D requests. This PR contains no coordinator implementation.

This is a follow-up to #2141 and implements the EPP revision-coordination part of #2143.

Part of #2143.

**Testing**:

- [x] `make presubmit` on the equivalent stacked change with #2570 present

```release-note
Coordinate DisaggregatedSet revision decisions across EPP replicas.
```


pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/plugin.go
```diff
@@ -54,6 +55,11 @@ type Screener struct {
 	scope            labels.Selector
 	revisionLabelKey string
 	roleLabelKey     string
+	decisionStateKey fwkdl.StateKey
```
hasB4K: wrong variable name?

pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/plugin.go
```diff
@@ -54,6 +55,11 @@ type Screener struct {
...
+	syncer           fwkdl.CrossReplicaSyncer
+	localDecisionMu  sync.Mutex
+	localDecisions   map[string]localRevisionDecision
+	lastLocalSweep   time.Time
```
hasB4K: This should be in a small GetOrSet utils class I believe (and no, not a LocalSyncer).

pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/screener.go
```diff
@@ -116,12 +157,12 @@ func (c *Screener) screenStrictSelectors(_ context.Context, request *fwksched.In
...
 
-func (c *Screener) hasStrictHeader(request *fwksched.InferenceRequest) bool {
+func (c *Screener) hasStrictRevisionHeader(request *fwksched.InferenceRequest) bool {
 	if request == nil {
 		return false
 	}
 	for _, selector := range c.config.HeaderSelectors {
-		if selector.Mode == ModeStrict && request.Headers[selector.HeaderName] != "" {
+		if selector.Mode == ModeStrict && selector.LabelKey == c.revisionLabelKey && request.Headers[selector.HeaderName] != "" {
 			return true
```
hasB4K: I don't get this change, all mode strict will be tied to revision only now?

pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/local_get_or_set.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (s *localGetOrSet) GetOrSet(key, candidate string) (string, bool) {
+	now := time.Now()
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.values == nil {
+		s.values = make(map[string]localGetOrSetValue)
+	}
+	if s.lastSweep.IsZero() || now.Sub(s.lastSweep) >= localGetOrSetSweepInterval {
```
ahg-g: A little strange to do this here on the get path instead of an independent gc go routine.

pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/screener.go
```diff
@@ -61,6 +76,42 @@ func (c *Screener) Screen(ctx context.Context, request *fwksched.InferenceReques
...
+	if decisionID := request.Headers[reqcommon.RevisionDecisionIDHeaderKey]; decisionID != "" {
+		return decisionID
+	}
+	return request.Headers[reqcommon.RequestIDHeaderKey]
+}
+
+func (c *Screener) getOrSetRevision(ctx context.Context, id, candidate string) (string, error) {
+	syncer := c.crossReplicaSyncer()
+	if syncer == nil {
+		revision, _ := c.localRevisionDecisions.GetOrSet(id, candidate)
```
ahg-g: pls document why a local syncer is needed.

pkg/common/request/constants.go
```diff
@@ -17,7 +17,8 @@ limitations under the License.
 package request
 
 const (
-	RequestIDHeaderKey = "x-request-id"
+	RequestIDHeaderKey          = "x-request-id"
+	RevisionDecisionIDHeaderKey = "x-llm-d-revision-decision-id"
```
ahg-g, line 26: pls document why this is needed, is it because x-request-id may not be set?

ahg-g (commented):
a couple of nits, nothing is blocking

pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/README.md
```diff
@@ -318,6 +312,13 @@ within one NVL72 NVLink domain. Preferring the same slice can avoid a slower
 cross-domain KV-cache transfer while retaining fallback capacity when that
 slice is unavailable or overloaded.
 
+Strict revision selection is supported for both P/D and E/P/D because the
+shared revision decision is coordinated independently of phase response order.
+Strict slice selection is supported for P/D, where a forwarded prefill slice
+can constrain decode, but it does not provide an end-to-end slice guarantee for
+E/P/D. Slice selection should normally use `prefer` so KV-cache and load-aware
+scoring can select another slice when it is a better candidate.
```
hasB4K: Basically, it works for `disaggregatedset.x-k8s.io/revision` but not for `disaggregatedset.x-k8s.io/slice` in EPD. Do you think we will ever need strict filtering for slice @BenjaminBraunDev? I think we don't need it at all since slices are for NVL72 racks, so we can still have comms between prefill and decode cross NVL72 racks, we just want to favor comms within one rack, but not enforce it. In my head I added list of several strict filter because it was free, but now I realize that it's not for EPD support, and I want to simplify the plugin to only support one headerSelector @ahg-g.

it will look like this (config-wise), I think we can do the change since the plugin is in Alpha anyway.

```
plugins:
- type: disaggregatedset-rollout-screener
  name: rollout-screener
  parameters:
    scope:
      labelSelector: disaggregatedset.x-k8s.io/name=my-set

    revisionGating:
      revisionHeaderName: x-disagg-revision  # default
      revisionLabelKey: disaggregatedset.x-k8s.io/revision  # default
      roleLabelKey: disaggregatedset.x-k8s.io/role  # default
        
      mode: max-role
      requireRoles: [prefill, decode]

- type: header-label-affinity-scorer
  name: slice-affinity
  parameters:
    headerName: x-disagg-slice
    labelKey: disaggregatedset.x-k8s.io/slice
    stampResponseHeader: true

- type: weighted-random-picker
  name: picker

schedulingProfiles:
- name: decode
  plugins:
  - pluginRef: slice-affinity
    weight: 3
  - pluginRef: picker
```

INSTEAD OF

```
plugins:
- type: disaggregatedset-rollout-screener
  name: rollout-screener
  parameters:
    scope:
      labelSelector: "disaggregatedset.x-k8s.io/name=my-set"
    headerSelectors:
    - name: revision
      headerName: x-disagg-revision
      labelKey: disaggregatedset.x-k8s.io/revision
      mode: strict
    - name: slice
      headerName: x-disagg-slice
      labelKey: disaggregatedset.x-k8s.io/slice
      mode: prefer
    revisionGating:
      mode: max-role # sum | max-role | disabled
      requireRoles:
        values: [prefill, decode]
- type: header-label-affinity-scorer
  name: slice-affinity
  parameters:
    headerName: x-disagg-slice
    labelKey: disaggregatedset.x-k8s.io/slice
- type: weighted-random-picker
  name: picker

schedulingProfiles:
- name: decode
  plugins:
  - pluginRef: slice-affinity
    weight: 3
  - pluginRef: picker
```

pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/README.md
```diff
@@ -127,22 +131,20 @@ roles and keeps only endpoints with that revision.
...
+`x-llm-d-revision-decision-id` for the request and sends it to every encode,
+prefill, and decode EPP request.
 
-With that forwarding in place, the prefill EPP chooses a covered revision and
-stamps it when the selected prefill begins responding. The decode EPP then
-applies the forwarded revision strictly:
+Each Screener can propose a covered revision, but atomic `GetOrSet` makes the
+first stored revision authoritative. In E/P/D, parallel encode requests can
+therefore begin together without waiting for one encode response. Prefill and
+decode use the same decision ID and receive the same revision.
```
hasB4K: No. The Decode will simply use the X-Disagg-Revision header to select its revision since prefill llm-d and decode llm-d might not use the same redis.

The GetOrSet mechanism is only useful for EPD, yes we also use it for PD if we don't have the X-Disagg-Revision. But that's okay.

pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/README.md
```diff
@@ -257,24 +247,18 @@ plugins:
...
       mode: max-role
-      requireRoles:
-        values: [prefill, decode]
+      requiredRoles: [prefill, decode]
 - type: header-label-affinity-scorer
   name: slice-affinity
   parameters:
     headerName: x-disagg-slice
     labelKey: disaggregatedset.x-k8s.io/slice
+    stampResponseHeader: true
```
hasB4K: Is this really needed? It seems to me that this could be done by default? Explain.

pkg/epp/framework/plugins/scheduling/scorer/headerlabelaffinity/README.md
```diff
@@ -44,8 +46,8 @@ schedulingProfiles:
   - pluginRef: picker
 ```
 
-The scorer does not write response headers. Protocols that return the selected
-label to a later request must configure response-header stamping separately.
+Response-header stamping is optional. Enable it when a coordinator must pass
+the selected label from one request to a later request.
```
hasB4K: So on first role, let's say prefill, if the header is not present then we need to stamp it based on the later request, otherwise we need to not stamp it? Explain this, and stampResponseHeader should now be by default at true.

pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/local_get_or_set.go
```diff
@@ -0,0 +1,84 @@
...
+	"sync"
+	"time"
+)
+
+const (
+	localGetOrSetTTL           = 10 * time.Minute
+	localGetOrSetSweepInterval = time.Minute
+)
+
+type localGetOrSet struct {
```
ahg-g: instead of having a custom store, we can use https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/interface/plugin/plugin_state.go; you may need to add a `ReadOrWrite` (aka GetOrSet) function to `PluginState`.

ahg-g: basically completely drop localGetOrSet and just create a PluginState instance for the screener plugin.

pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/config.go
```diff
@@ -44,85 +30,57 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
-// DisaggregatedSet operator label conventions. Used as defaults when the
-// operator omits RevisionGating.RevisionLabelKey / RoleLabelKey.
+// DisaggregatedSet rollout protocol and operator label defaults.
 const (
-	DefaultRevisionLabel = "disaggregatedset.x-k8s.io/revision"
-	DefaultRoleLabel     = "disaggregatedset.x-k8s.io/role"
+	DefaultRevisionHeader = "x-disagg-revision"
```
ahg-g: can we use an `x-llm-d` prefix pls here?

pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/screener.go
```diff
@@ -18,116 +18,149 @@ package disaggregatedsetrollout
...
+	// A single downstream request needs no identifier beyond its request ID.
+	return request.Headers[reqcommon.RequestIDHeaderKey]
+}
+
+func (c *Screener) getOrSetRevision(ctx context.Context, id, candidate string) (string, error) {
+	syncer := c.crossReplicaSyncer()
+	if syncer == nil {
+		// A coordinator issues a separate request for each phase. When those
+		// requests reach one EPP process, this store makes their shared decision
+		// ID reuse the first revision without requiring cross-replica storage.
```
ahg-g: pls clarify that this is needed specifically to handle parallel encoder requests, I believe this is the only case that require a local store to coordinate revision selection of parallel requests that share the same request id.

pkg/epp/framework/interface/plugin/plugin_state.go
```diff
@@ -125,17 +125,19 @@ func (s *PluginState) Read(requestID string, key StateKey) (StateData, error) {
...
-	if ok {
-		stateData = stateMap.(*sync.Map)
-	} else {
-		stateData = &sync.Map{}
-	}
-
+	stateMap, _ := s.storage.LoadOrStore(requestID, &sync.Map{})
+	stateData := stateMap.(*sync.Map)
 	stateData.Store(key, val)
+}
```
hasB4K: For Write - Why are you using a LoadOrStore if LoadOrStore won't write stuff if already exists?

pkg/epp/framework/interface/plugin/plugin_state.go
```diff
@@ -125,17 +125,19 @@ func (s *PluginState) Read(requestID string, key StateKey) (StateData, error) {
...
-	s.storage.Store(requestID, stateData)
+// ReadOrWrite atomically returns the data already stored for key and requestID,
+// or stores and returns val when no data exists. The boolean reports whether
+// the returned data was already present.
+func (s *PluginState) ReadOrWrite(requestID string, key StateKey, val StateData) (actual StateData, existed bool) {
+	s.requestToLastAccessTime.Store(requestID, time.Now())
+	stateMap, _ := s.storage.LoadOrStore(requestID, &sync.Map{})
+	actualValue, existed := stateMap.(*sync.Map).LoadOrStore(key, val)
+	return actualValue.(StateData), existed
 }
```
hasB4K: How is the atomic mechanism done?

hasB4K:
## Why `needCoordination` exists

Short version: **most users should omit this setting and keep the default, `true`.**

Coordination is only active during a rollout, when the screener observes Pods from more than one revision. A NotReady Pod from the new revision is enough to enable it before that revision receives traffic. Why? Because if a NotReady pod of another revision exists, it means that in a few seconds, or minutes, we will need to handle the coordination. Pods doesn't start immediately, so except if pods start in less than 5s, I think we should be fine.

This behavior avoid the synchronization with redis or something. With one observed revision, there is no `GetOrSet` call. A request that already carries `x-llm-d-disagg-revision` also bypasses `GetOrSet` and uses that revision as a strict constraint.

### P/D versus E/P/D

| Topology | Request ordering | How the revision stays consistent |
|---|---|---|
| P/D | Prefill completes before decode starts. | Prefill chooses a revision and returns `x-llm-d-disagg-revision`; the coordinator copies it into the decode request, which filters strictly. |
| E/P/D | Encode and prefill requests can start in parallel. | No response header exists yet, so the parallel requests share `x-llm-d-revision-decision-id`. Atomic `GetOrSet` makes the first selected revision authoritative for all of them, including requests reaching different EPP replicas. |

A deployment that is guaranteed to remain sequential P/D may set:

```yaml
revisionGating:
  needCoordination: false
```

That is only an optional rollout-time optimization. It should not normally be set: leaving the default enabled adds no syncer round trip with one observed revision, and disabling it would make a later move to parallel E/P/D unsafe. E/P/D must keep it enabled and use a shared cross-replica syncer when multiple EPP replicas are running.

### Redis `GetOrSet` benchmark

I benchmarked the intended Redis 7+ operation, `SET <unique-request-id> revision-1 PX 300000 NX GET`, using Redis 8.2.2 and go-redis v9.20.1. Redis ran in a local Docker container. Every request ID was unique, so this measured the write path. Each rate/worker combination ran for one minute; 229,200 operations completed with zero errors.

| Requests/s | 10 workers p50 / p99 | 50 workers p50 / p99 |
|---:|---:|---:|
| 10 | 0.498 / 1.411 ms | 0.497 / 2.821 ms |
| 50 | 0.445 / 2.200 ms | 0.442 / 2.216 ms |
| 100 | 0.388 / 2.268 ms | 0.398 / 1.844 ms |
| 250 | 0.297 / 1.722 ms | 0.308 / 2.232 ms |
| 500 | 0.241 / 2.797 ms | 0.230 / 1.755 ms |
| 1,000 | **0.195 / 1.295 ms** | **0.184 / 1.531 ms** |

At 1,000 requests/s, the observed median was therefore about **~0.2 ms**. Redis reported about 5.49 microseconds of server execution per command; most observed latency came from the Go client, Docker networking, and host scheduling. Production network distance, TLS, and authentication can increase it.

This cost applies only to unpinned requests during the short interval in which multiple rollout revisions are observed. For that reason, the safe default is also the practical default: leave `needCoordination: true`.

pkg/epp/framework/plugins/scheduling/scorer/headerlabelaffinity/plugin.go
```diff
@@ -92,3 +104,16 @@ func (s *Scorer) Score(_ context.Context, request *fwksched.InferenceRequest, en
 	}
 	return scores
 }
+
+// ResponseHeader reports the selected endpoint's actual label, which a
+// coordinator can copy into a later request. This deliberately does not depend
+// on whether the current request carried the header: affinity is soft, so the
+// endpoint that won can have a different label from the requested preference.
```
ahg-g, line 111: if there is no guarantee that this was indeed the selected endpoint, why is it is useful to copy it to the response header for later requests? what use case do you have in mind? how will the coordinator potentially use it?

pkg/common/request/constants.go
```diff
@@ -18,6 +18,10 @@ package request
 
 const (
 	RequestIDHeaderKey = "x-request-id"
+	// RevisionDecisionIDHeaderKey groups downstream requests into one rollout
+	// decision. A coordinator can assign distinct request IDs to the prefill and
```
ahg-g: the coordinator sets the `x-llm-d-revision-decision-id` header, right? if so, pls clarify that.

pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/config.go
```diff
@@ -44,85 +30,58 @@ import (
...
+//  2. Load shaping: when no revision header is present, weighted-random-pick
+//     one surviving revision and keep only its pods.
+//
+// A request carrying RevisionHeaderName always selects that revision strictly.
+// The selected endpoint's revision is stamped into the same response header.
+type RevisionGating struct {
+	RevisionHeaderName string     `json:"revisionHeaderName,omitempty"`
+	RevisionLabelKey   string     `json:"revisionLabelKey,omitempty"`
+	RoleLabelKey       string     `json:"roleLabelKey,omitempty"`
+	NeedCoordination   *bool      `json:"needCoordination,omitempty"`
```
ahg-g: if this is intended to be an opt-out feature, I would call it `disableCoordination`

ahg-g (commented):
nice, a few nits.

pkg/epp/framework/plugins/scheduling/scorer/headerlabelaffinity/plugin.go
hasB4K: The main use case is slice locality. In a normal P/D flow, the prefill response tells the coordinator which slice EPP scheduled it on. The coordinator then copies x-disagg-slice into the decode request so decode prefers the same slice and avoids an unnecessary cross-slice KV-cache transfer (if it can). Basically, if you have a lot of images that went on a specific slice, you want the next role to use this slice as a marker to favor this slice to select the next decode. But in any case, it's a best effort, since the transfer between slice (ie. cross NVL72 racks) is still quite fast.

The same mechanism also covers E/P/D. Each parallel image encode response reports its scheduled slice; the coordinator selects the most frequent value and forwards it to the later phase which I implemented here: https://github.com/llm-d/llm-d-router/pull/2301/changes/c0633d3af94c689dbe868caf179d32d8b230ae60 (other PR)

ahg-g:
Looks great, thanks for upgrading to 1.36 and using the newer AsyncTracker library. pls resolve the conflict, sorry about that.

ahg-g (approved):

yankay:
Hi @hasB4K and @ahg-g 

Thanks for the work on this. Would it be helpful to mention the breaking change from `requireRoles.values` to `requiredRoles` in the release note, with a brief migration note? This version difference came up https://github.com/llm-d/llm-d/pull/2379#discussion_r4014188184

ahg-g:
> Hi @hasB4K and @ahg-g
> 
> Thanks for the work on this. Would it be helpful to mention the breaking change from `requireRoles.values` to `requiredRoles` in the release note, with a brief migration note? This version difference came up [llm-d/llm-d#2379 (comment)](https://github.com/llm-d/llm-d/pull/2379#discussion_r4014188184)

I don't think this was released to warrant a migration.

PR #2572, opened 2026-08-26
feat(coordinator): share revision decisions across phases [Safe DisaggSet 4/4]

hasB4K:
**What type of PR is this?**
/kind feature
/kind documentation
/kind test

Generates one coordinator-owned `x-llm-d-revision-decision-id` for each incoming request and sends it to every Encode, Prefill, and Decode request. A revision-aware EPP can use the same identifier with an atomic cross-replica operation, so parallel encode requests and later phases use one revision decision.

Client-provided values for this header are discarded. The header is independent of `x-request-id`, so a client cannot select or accidentally reuse a cached revision decision.

The coordinator also forwards `x-llm-d-disagg-revision` between phases by default. Declaring `pipeline.forward_response_headers` replaces that default, and an explicit empty list disables response-header forwarding. This carries the EPP-selected revision to later phases as a strict constraint and discards a client-provided value for that header.

**Dependencies:** this depends on #2571, which defines `RevisionDecisionIDHeaderKey` and the DisaggregatedSet Screener, and #2301, which adds pipeline-level response-header forwarding. The branch includes #2301 until that PR merges.

Part of #2143.

**Testing**:

- [x] `go test -race ./pkg/coordinator/pipeline/... ./pkg/coordinator/steps/... ./pkg/coordinator/config/... ./pkg/coordinator/server/... ./cmd/coordinator/... ./pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout`

```release-note
Coordinator requests share a revision decision ID across Encode, Prefill, and Decode phases and forward the selected rollout revision between phases by default.
```


ahg-g:
@hasB4K can you rebase to so the test picks the changes to https://github.com/llm-d/llm-d-router/pull/2571 pls?

ahg-g:
looks good otherwise

ahg-g (approved):

PR #2573, opened 2026-08-26
tracing: support OTLP/HTTP as an alternative to the gRPC exporter

### LINKED ISSUE #2567: tracing: support OTLP/HTTP as an alternative to the gRPC exporter

Select the OTLP transport from `OTEL_EXPORTER_OTLP_PROTOCOL` / `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` in `initTraceExporter` (`pkg/common/observability/tracing/telemetry.go`), and add an `otlptracehttp` path alongside the existing `otlptracegrpc` one.

`otlptracegrpc.New` is the only exporter constructed today. No OTLP exporter package reads the protocol variable - protocol selection is the caller's job - so `http/protobuf` currently has no effect and spans still go over gRPC to port 4317, against a backend that may only serve 4318.

Three implementation notes:

- **This adds a module.** `go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp` is not in `go.sum`. It is a new direct dependency at the same version as the gRPC exporter (v1.44.0) and from the same release train as the modules already required, so it adds no new transitive versions, but it needs agreement before the work starts.

- **`http/json` is out of scope.** The Go SDK ships no JSON encoder for OTLP; only `grpc` and `http/protobuf` are implementable. `http/json` should reach the error handler as unsupported rather than falling back silently.

- **The loopback fallback becomes protocol-dependent.** `localCollectorOptions` pins `localhost:4317` and plaintext when no transport variable is set. The HTTP exporter defaults to 4318, so the fallback has to select per protocol, and the option types (`otlptracegrpc.Option` and `otlptracehttp.Option`) are distinct, so one helper cannot serve both as currently written.

**Why is this needed**:

OTLP/HTTP is the transport many managed backends and ingress paths expose, and some environments terminate gRPC at the edge or disallow HTTP/2 to the collector. With gRPC as the only path, those deployments cannot export at all. The protocol variable is read by nothing: an operator sets `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf`, sees no error, and gets no spans.

**Acceptance criteria**:

- `http/protobuf` exports over OTLP/HTTP; `grpc` and unset both export over gRPC.
- `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` takes precedence over `OTEL_EXPORTER_OTLP_PROTOCOL`.
- An unsupported protocol value is reported through the error handler.
- The TLS, header and certificate behavior established in #2311 holds on the HTTP path.
- With no transport variables set, the fallback targets 4317 for gRPC and 4318 for HTTP, plaintext in both cases.
- Tests against an in-process HTTP collector, mirroring the gRPC collector in `telemetry_otlp_test.go`.

Part of #1632.

satyamg1620:
**What type of PR is this?**
/kind feature

No OTLP exporter package reads `OTEL_EXPORTER_OTLP_PROTOCOL` - protocol selection is the caller's job - so setting `http/protobuf` had no effect. Spans still went over gRPC to port 4317, against a backend that may serve only 4318, and nothing reported the ignored setting. Since #2566 made `otlp` the default exporter, that now means no traces at all rather than unexpected console output.

The protocol is resolved from `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` then `OTEL_EXPORTER_OTLP_PROTOCOL`, and `newTraceExporter` builds `otlptracehttp` or `otlptracegrpc` accordingly. An unsupported value is reported through the error handler and falls back to gRPC. `http/json` is among those: the Go SDK ships no JSON encoder for OTLP, so the alternative would be sending protobuf to an endpoint configured for JSON.

Two choices worth review:

- **Resolution stops at the first variable that is set, not the first that is valid.** A rejected per-signal value does not fall through to the generic one, which would otherwise apply a transport that was not chosen for traces.
- **`newTraceExporter` takes the `*errorHandler`.** The protocol is resolved inside the OTLP branch so that `console` and `none` neither validate nor log a protocol that has no bearing on them; reporting at the point of resolution needs both the logger and the handler, and `errorHandler` carries both.

The loopback fallback splits into `localGRPCCollectorOptions` (4317) and `localHTTPCollectorOptions` (4318), sharing `transportConfigured()`. The option types are distinct, so one helper cannot serve both.

`go.mod` gains one line and `go.sum` two hash lines. `otlptracehttp` is pinned to v1.44.0, matching the other otel modules: resolving it by `go get` picks v1.46.0 and drags `otel`, `sdk`, `trace` and `stdouttrace` up with it. No other module moves and no new transitive versions enter the build graph.

Fixes #2567

**Test plan**:

Unit:
- [x] Unset, `grpc` and `http/protobuf` resolve as expected; the per-signal variable wins over the generic one
- [x] `http/json`, an unrecognised value and an empty value are reported and fall back to gRPC
- [x] A rejected per-signal value does not fall through to the generic one
- [x] OTLP/HTTP reaches an in-process collector, defaults to loopback 4318, sends `OTEL_EXPORTER_OTLP_HEADERS`, and reaches a TLS collector via `OTEL_EXPORTER_OTLP_CERTIFICATE`

Verified on a kind dev stack against a real `otel/opentelemetry-collector` sidecar, run
with one receiver enabled at a time so that spans arriving identifies the transport used:
- [x] `http/protobuf` with an HTTP-only collector on 4318 delivers spans; the same config with `grpc` delivers none and logs connection-refused on 4317
- [x] Unset still reaches a gRPC-only collector on 4317, unchanged
- [x] A per-signal `http/protobuf` overrides a generic `grpc`
- [x] `http/json` warns and falls back to gRPC
- [x] Spans delivered over HTTP decode correctly, carrying `service.name: llm-d-epp` and the expected EPP span names

```release-note
Tracing can now export over OTLP/HTTP. Setting OTEL_EXPORTER_OTLP_PROTOCOL or OTEL_EXPORTER_OTLP_TRACES_PROTOCOL to http/protobuf sends spans over HTTP to port 4318 instead of gRPC to 4317; grpc and an unset variable keep the existing behaviour. An unsupported protocol, including http/json which the Go SDK cannot encode, is logged and falls back to gRPC.

### DIFF
--- a/pkg/common/observability/tracing/telemetry.go
+++ b/pkg/common/observability/tracing/telemetry.go
@@ -18,9 +18,11 @@ package tracing
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"go.opentelemetry.io/otel"
@@ -75,7 +77,9 @@ func InitTracing(ctx context.Context, logger logr.Logger, defaultServiceName str
 
 	// "none" registers no span processor at all. Spans are still created and
 	// propagated, so instrumented code and context propagation are unaffected.
-	if exporterType != exporterTypeNone {
+	if exporterType == exporterTypeNone {
+		logger.Info("init OTel trace exporter", "type", exporterType)
+	} else {
 		traceExporter, err := newTraceExporter(ctx, loggerWrap, exporterType)
 		if err != nil {
 			loggerWrap.Handle(fmt.Errorf("%s: %v", "init trace exporter failed", err))
@@ -217,7 +221,8 @@ func traceExporterType() (string, error) {
 }
 
 // The OTLP transports OTEL_EXPORTER_OTLP_PROTOCOL selects between. http/json is not
-// listed: the Go SDK ships no JSON encoder for OTLP, so it cannot be served.
+// among them: the Go SDK ships no JSON encoder for OTLP, so it is unsupported like any
+// value outside this set.
 const (
 	protocolGRPC = "grpc"
 	protocolHTTP = "http/protobuf"
@@ -231,28 +236,38 @@ var otlpProtocolEnv = []string{
 	"OTEL_EXPORTER_OTLP_PROTOCOL",
 }
 
-// otlpProtocol resolves the OTLP transport. The per-signal variable wins when both are
-// set, as the specification requires.
+// otlpProtocol resolves the OTLP transport. The per-signal variable wins when both name
+// a supported one.
 //
-// An unsupported value is returned as an error alongside the default, and resolution
-// stops at the first variable that is set: a rejected per-signal value is a
-// misconfiguration to report, not a reason to consult the generic one.
+// The specification governs the rest: an empty value reads the same as unset, enum
+// values match case-insensitively, and a value the implementation does not recognise is
+// reported and then ignored, leaving the next variable to apply. gRPC applies when none
+// of them resolves.
+//
+// Ignored values are returned alongside the protocol that won, so a misconfiguration is
+// still reported even when a later variable supplies a usable transport.
 func otlpProtocol() (string, error) {
+	var ignored []error
+
 	for _, key := range otlpProtocolEnv {
-		protocol, ok := os.LookupEnv(key)
-		if !ok {
+		protocol := strings.ToLower(strings.TrimSpace(os.Getenv(key)))
+		if protocol == "" {
 			continue
 		}
 
 		switch protocol {
 		case protocolGRPC, protocolHTTP:
-			return protocol, nil
+			return protocol, errors.Join(ignored...)
 		default:
-			return defaultProtocol, fmt.Errorf("unsupported %s %q, falling back to %s", key, protocol, defaultProtocol)
+			ignored = append(ignored, fmt.Errorf("unsupported %s %q, ignored", key, protocol))
 		}
 	}
 
-	return defaultProtocol, nil
+	if len(ignored) > 0 {
+		ignored = append(ignored, fmt.Errorf("no supported OTLP protocol configured, falling back to %s", defaultProtocol))
+	}
+
+	return defaultProtocol, errors.Join(ignored...)
 }
 
 // newTraceExporter builds the exporter named by exporterType, which traceExporterType
@@ -341,7 +356,7 @@ func localGRPCCollectorOptions() []otlptracegrpc.Option {
 }
 
 // localHTTPCollectorOptions is localGRPCCollectorOptions for the HTTP transport, which
-// the SDK defaults to a different port.
+// serves on its own port and takes an unrelated option type.
 func localHTTPCollectorOptions() []otlptracehttp.Option {
 	if transportConfigured() {
 		return nil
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by satyamg1620** [comment]
@ahg-g Can you please review this PR

File: pkg/common/observability/tracing/telemetry.go [hunk 1]
```diff
@@ -217,6 +216,45 @@ func traceExporterType() (string, error) {
...
+	"OTEL_EXPORTER_OTLP_PROTOCOL",
+}
+
+// otlpProtocol resolves the OTLP transport. The per-signal variable wins when both are
+// set, as the specification requires.
+//
+// An unsupported value is returned as an error alongside the default, and resolution
+// stops at the first variable that is set: a rejected per-signal value is a
+// misconfiguration to report, not a reason to consult the generic one.
+func otlpProtocol() (string, error) {
```
JaredTan95, line 249: Invalid `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` values should be ignored rather than immediately falling back to gRPC. For example, if `TRACES_PROTOCOL=invalid` and `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf`, the valid generic setting should still be used.

Per the [[OTel SDK environment variable specification](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/)](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/), unrecognized enum values should be ignored with a warning.

Please `continue` here and only fall back to gRPC after no valid protocol configuration is found.

pkg/common/observability/tracing/telemetry_otlp_test.go
```diff
@@ -350,3 +360,246 @@ func TestInitTracingDefaultsToOTLPCollector(t *testing.T) {
...
+			wantErr: true,
+		},
+		{
+			name:    "an unrecognised protocol is reported",
+			env:     map[string]string{"OTEL_EXPORTER_OTLP_PROTOCOL": "thrift"},
+			want:    protocolGRPC,
+			wantErr: true,
+		},
+		{
+			name:    "an empty value is reported rather than treated as unset",
```
JaredTan95: An empty `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` should be treated as unset, so the valid generic `OTEL_EXPORTER_OTLP_PROTOCOL` can still be used.

The [[OTel environment variable specification](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/)](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/) explicitly requires empty values to be interpreted the same as unset values. This test should therefore expect `http/protobuf` rather than falling back to gRPC.

pkg/common/observability/tracing/telemetry.go
```diff
@@ -217,6 +216,45 @@ func traceExporterType() (string, error) {
...
+	for _, key := range otlpProtocolEnv {
+		protocol, ok := os.LookupEnv(key)
+		if !ok {
+			continue
+		}
+
+		switch protocol {
+		case protocolGRPC, protocolHTTP:
```
JaredTan95, line 259: OTLP protocol values are enums and should be parsed case-insensitively. The [[OTel environment variable specification](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/)](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/) states that enum values SHOULD be interpreted in a case-insensitive manner.

Consider applying `strings.ToLower(protocol)` before matching so `GRPC` and `HTTP/PROTOBUF` are accepted.

gyliu513:
I think all the comments from @JaredTan95 are now addressed, lgtm.

@ahg-g can you help approve？

ahg-g (approved):

Issue #2283, opened 2026-08-05
Allow plugins to contribute to EPP readiness


hasB4K:
**What would you like to be added**:

Hey there! I would like plugins to have an optional way to contribute a condition to EPP readiness.

For example, a plugin could implement a small readiness contract saying whether it has completed its initial synchronization or can reach a required backend. EPP would aggregate those conditions into its readiness response. A plugin that is not ready should make EPP report `NOT_SERVING` for readiness, without affecting liveness.

I see at least three concrete uses already:

- Precise KV-cache routing in a multiple-EPP setup may need an initial snapshot from the vLLMs before the EPP should receive traffic.
- The DisaggregatedSet rollout Screener in #2141 needs an initial Pod notification snapshot before it can determine which cross-role revisions are safe. It currently fails closed with HTTP 503 during that short window.
- A Redis-backed syncer, such as the work discussed around #1921, may need to confirm that it can connect to its Redis backend and complete its initial synchronization before advertising readiness.

The exact interface is open for discussion. Something like an optional plugin readiness checker could work, as long as it remains generic and is evaluated only for readiness and not liveness.

**Why is this needed**:

Today EPP readiness covers datastore synchronization, leadership, protocol support, and draining, but plugins cannot contribute their own asynchronous initialization state.

Without a generic hook, EPP can become ready while a configured plugin still cannot safely process traffic. Each plugin would otherwise need bespoke health wiring. A shared readiness contract would keep this behavior consistent and reusable across snapshot-based routing and external state synchronizers.

PR #2576, opened 2026-08-26
feat(epp): gate DisaggregatedSet readiness on initial Pod snapshot

hasB4K:
**What type of PR is this?**

/kind feature

**NOTE FOR THE REVIEWERS**
FYI, this PR is totally unrelated and independent of #2570, #2571, #2570 and #2301. It touch the Disagg screener, but it's not the same subject. This PR is more a "nice to have" when doing a EPP rolling update to be sure that the new EPP won't serve 503 requests during the init.

Builds on the plugin readiness hook from #2307, and on the Disagg screener in #2141. A DisaggregatedSet rollout Screener with revision gating stays `NOT_SERVING` until it has applied one complete initial Pod snapshot, including an empty snapshot. This removes the startup window in which EPP is ready but revision gating has no coverage data and therefore fails requests closed with HTTP 503.

EPP receives the initial Pod cache as individual notifications, not as one atomic event. Marking the Screener ready after the first notification could expose it while its Pod map is only partially populated. For example, it may have observed a decode Pod for a revision but not its prefill Pod yet; revision gating would treat that revision as missing a required role and return an incorrect HTTP 503. A synchronized snapshot is the point at which the Screener knows it has processed every Pod currently in the cache, including the valid empty case.

The notification data layer adds an optional `NotificationSnapshotExtractor` contract. It delivers the cached initial list to opt-in extractors and holds normal notifications behind that snapshot. The Screener keeps no Kubernetes client; it receives the snapshot through the framework-owned source.

Follow-up to #2283.

**How was this tested?**:

- `make lint`
- `make test-unit`

**Release note**:
```release-note
DisaggregatedSet rollout EPPs wait for an initial Pod snapshot before receiving traffic.
```


ahg-g:
@hasB4K pls sign the commit

pkg/epp/datalayer/k8s_bind.go
```diff
@@ -68,15 +81,33 @@ func BindNotificationSource(src fwkdl.NotificationSource, extractors []fwkdl.Not
...
+
+type notificationInitialSnapshotRunnable struct {
+	reconciler *notificationReconciler
+}
+
+var _ manager.LeaderElectionRunnable = (*notificationInitialSnapshotRunnable)(nil)
+
+func (*notificationInitialSnapshotRunnable) NeedLeaderElection() bool { return false }
+
+func (r *notificationInitialSnapshotRunnable) Start(ctx context.Context) error {
```
ahg-g: pls document when this is expected to be called by the manager, is it after the initial watch sync?

ahg-g:
Can we not use WaitForCacheSync on every registered kind on startup instead of manually listing and taking a snapshot?

hasB4K:
> Can we not use WaitForCacheSync on every registered kind on startup instead of manually listing and taking a snapshot?

Thanks, I dug into this a bit. I do not think `WaitForCacheSync()` alone gives us the guarantee we need.

  The flow today is:

```
  Kubernetes API
        |
        v
  local controller-runtime cache
        |
        v
  controller queue
        |
        v
  Reconcile()
        |
        v
  Extract()
        |
        v
  Screener's internal Pod map
  ```

`WaitForCacheSync()` only means that the local cache has completed its initial List. The Screener does not read that cache directly. It is updated later through Reconcile() and Extract().

Controller-runtime itself makes this distinction: it waits for the cache first, then separately waits for the event handler to receive the initial events ([here](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.23.3/pkg/internal/source/kind.go#L101-L112)). But this handler only enqueues reconciliation requests ([here](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.23.3/pkg/internal/source/event_handler.go#L62-L87)). The workers that run those requests start only afterwards ([here](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.23.3/pkg/internal/controller/controller.go#L296-L316)), and our Extract() call happens inside the reconciler ([here](https://github.com/hasB4K/llm-d-router/blob/mathis/disaggregatedset-screener-readiness/pkg/epp/datalayer/k8s_bind.go#L184-L202)).

So this state is possible:
```
  local cache:       [prefill rev-A, decode rev-A]
  Screener Pod map:  []
```

Even though `WaitForCacheSync()` has completed. Revision gating would see no complete revision and return the startup 503 we are trying to avoid.

**Comment by ahg-g** [comment]
I see, so the issue is not about waiting for the sync with the api-server; the issue is that the epp declares ready before the controller-runtime gets a chance to send the events to our reconciler. I wonder if there is a way to block startup until the events are processed rather than doing all this snapshotting logic? 

What I don't like is that the information we get from the initial snapshot is not really missing currently, it will eventually be sent to the plugin as events, and so we are adding all this snapshotting complexity and increasing the plugin interface surface area not because we are missing data, but because we are not blocking startup on receiving and processing these events.

**Comment by hasB4K** [comment]
> I see, so the issue is not about waiting for the sync with the api-server; the issue is that the epp declares ready before the controller-runtime gets a chance to send the events to our reconciler.

Yes exactly!

> I wonder if there is a way to block startup until the events are processed rather than doing all this snapshotting logic?
> What I don't like is that the information we get from the initial snapshot is not really missing currently, it will eventually be sent to the plugin as events, and so we are adding all this snapshotting complexity and increasing the plugin interface surface area not because we are missing data, but because we are not blocking startup on receiving and processing these events.

yeah, I don't like this implementation as well, I will try to find an alternative.

File: cmd/epp/runner/runner.go [hunk 2]
```diff
@@ -522,7 +522,9 @@ func (r *Runner) setup(ctx context.Context, cfg *rest.Config, opts *runserver.Op
 	r.draining = &atomic.Bool{}
 	r.serverRunner = serverRunner
 	r.healthGRPCPort = opts.GRPCHealthPort
-	r.healthGRPCServer = newHealthGRPCServer(ctrl.Log.WithName("health"), ds, isLeader, r.draining, opts.EnableLeaderElection, supporters, pluginReadinessCheckers(r.PluginHandle.GetAllPlugins()))
+	readinessCheckers := pluginReadinessCheckers(r.PluginHandle.GetAllPlugins())
+	readinessCheckers = append(readinessCheckers, r.dlRuntime)
```
hasB4K: **ELI5:** the health server already asks a list of readiness checkers, "is it safe to receive traffic yet?" This adds the whole data-layer runtime to that list. As a result, Kubernetes does not mark EPP Ready until both the existing checks *and* initial notification processing are done. Keeping this at the framework level also avoids teaching the rollout screener about Kubernetes startup details.

pkg/epp/datalayer/k8s_bind.go
```diff
@@ -36,43 +46,104 @@ import (
 // The framework core owns the cache and reconciliation; the source only receives
 // deep-copied events via Notify.
 func BindNotificationSource(src fwkdl.NotificationSource, extractors []fwkdl.NotificationExtractor, mgr ctrl.Manager) error {
+	_, err := bindNotificationSource(src, extractors, mgr)
```
hasB4K: This small wrapper preserves the existing public API: callers that only need success/failure still get an `error`. The internal helper additionally returns a progress tracker, which the runtime needs for readiness.

pkg/epp/datalayer/k8s_bind.go
```diff
@@ -36,43 +46,104 @@ import (
...
+	return err
+}
+
+func bindNotificationSource(src fwkdl.NotificationSource, extractors []fwkdl.NotificationExtractor, mgr ctrl.Manager) (*notificationInitialSync, error) {
 	gvk := src.GVK()
 	log := mgr.GetLogger().WithName("notification-controller").WithValues("gvk", gvk.Kind)
+	initialSync := newNotificationInitialSync(src.TypedName().String())
```
hasB4K: **ELI5:** every notification source gets its own startup checklist. For this PR the important source watches Pods, but the mechanism stays generic and will also work for another Kubernetes kind without screener-specific code.

pkg/epp/datalayer/k8s_bind.go
```diff
@@ -36,43 +46,104 @@ import (
...
 
 	obj := &unstructured.Unstructured{}
 	obj.SetGroupVersionKind(gvk)
+	// Kind.WaitForSync completes after its handler has enqueued the initial
+	// list. Track those keys through successful notification dispatch.
+	enqueue := &handler.TypedEnqueueRequestForObject[*unstructured.Unstructured]{}
+	eventHandler := handler.TypedFuncs[*unstructured.Unstructured, reconcile.Request]{
+		CreateFunc: func(ctx context.Context, event event.TypedCreateEvent[*unstructured.Unstructured], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+			if event.IsInInitialList && event.Object != nil {
+				initialSync.processed.Start(client.ObjectKeyFromObject(event.Object))
```
hasB4K: `IsInInitialList` is controller-runtime's label for objects discovered during the first API-server list. We put each such object key on the checklist *before* forwarding it to the normal work queue. That closes the bug's gap: "the object was queued" can no longer be mistaken for "the plugin processed the object."

pkg/epp/datalayer/k8s_bind.go
```diff
@@ -36,43 +46,104 @@ import (
...
 
-	return ctrl.NewControllerManagedBy(mgr).
+	err := ctrl.NewControllerManagedBy(mgr).
 		// Naming the controller allows you to see specific metrics/logs for this watch
 		Named(controllerName).
-		For(obj).
-		// ResourceVersionChanged is safer for generic notifications than GenerationChanged,
-		// as it catches metadata and status updates that the consumer might need.
-		WithEventFilter(predicate.ResourceVersionChangedPredicate{}).
+		WatchesRawSource(&notificationInitialSyncSource{
```
hasB4K: Previously `.For(obj)` constructed this event source internally, where we could not observe initial-list progress. Building the same source explicitly lets us wrap its `WaitForSync` and event handler. It is still one informer/watch and the same resource-version predicate—there is no second list or duplicate snapshot.

pkg/epp/datalayer/k8s_bind.go
```diff
@@ -36,43 +46,104 @@ import (
...
+	src         fwkdl.NotificationSource
+	extractors  []fwkdl.NotificationExtractor
+	initialSync *notificationInitialSync
+	gvk         schema.GroupVersionKind
+	log         logr.Logger
+}
+
+type notificationInitialSync struct {
+	upstreamSynced atomic.Bool
+	processed      synctrack.AsyncTracker[types.NamespacedName]
```
hasB4K: Think of `AsyncTracker` as a two-part latch: (1) the source says the initial list has been fully delivered, and (2) the set of initial object keys waiting for successful processing is empty. It is thread-safe, so this remains correct if reconciliation later uses multiple workers.

pkg/epp/datalayer/k8s_bind.go
```diff
@@ -36,43 +46,104 @@ import (
...
+type notificationInitialSyncSource struct {
+	source.SyncingSource
+	initialSync *notificationInitialSync
+}
+
+func (s *notificationInitialSyncSource) WaitForSync(ctx context.Context) error {
+	if err := s.SyncingSource.WaitForSync(ctx); err != nil {
+		return err
+	}
+	s.initialSync.upstreamSynced.Store(true)
```
hasB4K: This flips only the first half of that latch: controller-runtime has finished delivering/enqueuing the initial list. Any keys still waiting in `AsyncTracker` keep readiness false until their reconciliations finish.

pkg/epp/datalayer/k8s_bind.go
```diff
@@ -99,26 +170,32 @@ func (rn *notificationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
-	return rn.dispatch(ctx, log, event)
+	err = rn.dispatch(ctx, log, event)
+	if err == nil {
+		rn.initialSync.processed.Finished(req.NamespacedName)
```
hasB4K: A key is crossed off the startup checklist only after notification dispatch succeeds. If either the source or an extractor fails, the error goes back to controller-runtime, the item is retried, and EPP remains unready. Calling `Finished` for a later non-initial event is intentionally harmless because `AsyncTracker` ignores keys it was not waiting for.

pkg/epp/datalayer/k8s_bind.go
```diff
@@ -99,26 +170,32 @@ func (rn *notificationReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 	if processed == nil {
-		return ctrl.Result{}, nil
+		return nil
 	}
 
+	var extractErrors []error
 	for _, ext := range rn.extractors {
 		if err := ext.Extract(ctx, *processed); err != nil {
 			log.Error(err, "extractor failed", "extractor", ext.TypedName())
+			extractErrors = append(extractErrors, fmt.Errorf("extractor %s failed: %w", ext.TypedName(), err))
```
hasB4K: Extractor failures used to be logged and swallowed. That is no longer safe because swallowed errors would let us declare an initial item "finished" even though plugin state was not updated. We still call every extractor, collect their errors, and return the combined error so controller-runtime retries the event.

pkg/epp/datalayer/k8s_bind_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (*testSyncingSource) Start(context.Context, workqueue.TypedRateLimitingInterface[reconcile.Request]) error {
+	return nil
+}
+
+func (s *testSyncingSource) WaitForSync(context.Context) error {
+	return s.waitErr
+}
+
+func TestNotificationInitialSync(t *testing.T) {
```
hasB4K: This focused unit test exercises the checklist without running Kubernetes: even after the upstream list reports synced, readiness stays false until *both* queued keys are marked finished. It is the smallest model of the race we are fixing.

pkg/epp/datalayer/k8s_bind_test.go
```diff
@@ -0,0 +1,134 @@
...
+		t.Fatalf("CheckReady() error = %v, want pending source name", err)
+	}
+
+	initialSync.upstreamSynced.Store(true)
+	if err := runtime.CheckReady(); err != nil {
+		t.Fatalf("CheckReady() after an empty initial list: %v", err)
+	}
+}
+
+func TestNotificationDispatchReturnsExtractorErrors(t *testing.T) {
```
hasB4K: This protects an important side effect of the readiness fix: extractor errors must now reach the reconciler, but one failing extractor should not prevent the other extractors from seeing the event. Returning the error then causes a retry.

pkg/epp/datalayer/runtime.go
```diff
@@ -397,19 +400,48 @@ func (r *Runtime) findSourceByType(sourceType string, gvkFilter *schema.GroupVer
 // Start is called to enable the Runtime to start processing data collection. It wires
 // Kubernetes notifications into the manager.
 func (r *Runtime) Start(ctx context.Context, mgr ctrl.Manager) error {
-	return r.notification.ForEach(func(srcName string, src fwkdl.NotificationSource) error {
+	notificationSyncs := make([]*notificationInitialSync, 0, r.notification.Count())
```
hasB4K: During startup we bind every configured notification source and retain the checklist returned for each one. We publish the new list of trackers only after all bindings succeed, so health checks never observe a half-configured set.

pkg/epp/datalayer/runtime.go
```diff
@@ -397,19 +400,48 @@ func (r *Runtime) findSourceByType(sourceType string, gvkFilter *schema.GroupVer
...
+
+	r.notificationSyncMu.Lock()
+	r.notificationSyncs = notificationSyncs
+	r.notificationSyncMu.Unlock()
+	return nil
+}
+
+// CheckReady reports whether every notification source has processed the
+// events from its initial Kubernetes list.
+func (r *Runtime) CheckReady() error {
```
hasB4K: **ELI5:** this is the runtime's single answer to "are all startup inboxes empty?" No notification sources means there is nothing to wait for. Otherwise, one unfinished source is enough to keep EPP unready, and its name is included in the error for useful probe logs. The read lock protects concurrent health checks.

pkg/epp/datalayer/runtime.go
```diff
@@ -397,19 +400,48 @@ func (r *Runtime) findSourceByType(sourceType string, gvkFilter *schema.GroupVer
...
+
+	for _, initialSync := range r.notificationSyncs {
+		if !initialSync.hasSynced() {
+			return fmt.Errorf("notification source %s has not processed its initial events", initialSync.sourceName)
+		}
+	}
+	return nil
+}
+
+func (*Runtime) TypedName() fwkplugin.TypedName {
```
hasB4K: `TypedName` is the small adapter that lets `Runtime` satisfy the framework's existing `ReadinessChecker` interface. This is why we do not need to enlarge the screener/plugin API just for startup synchronization.

pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/README.md
```diff
@@ -338,11 +338,11 @@ With `sum` or `max-role`, a revision is ineligible until every required role
 has at least one Ready Pod. If no revision survives, request control returns
 HTTP 503.
 
-After EPP startup or restart, the in-memory distribution is empty until Pod
-notifications deliver the initial snapshot. During that interval, `sum` and
-`max-role` return HTTP 503 rather than route without cross-role coverage
-information. A follow-up change is required to make receipt of the initial Pod
-snapshot a condition of EPP readiness.
+When revision gating is enabled, EPP does not report ready until the data layer
```
hasB4K: This documents the user-visible split: during *startup*, Kubernetes withholds traffic until the initial state is known; after startup, losing a required role is a real availability condition and the screener still deliberately returns 503.

pkg/epp/framework/plugins/requestcontrol/screener/disaggregatedsetrollout/screener.go
```diff
@@ -34,8 +34,6 @@ func (c *Screener) Screen(ctx context.Context, request *fwksched.InferenceReques
 		}
 		seenRevisions := uniqueRevisions(current, c.revisionLabelKey)
 		distribution := c.distributionSnapshot()
-		// TODO: Make the initial Pod snapshot a condition of EPP readiness.
```
hasB4K: This TODO can now be removed because the generic data-layer readiness gate provides the missing guarantee. The screener still fails closed if its distribution is empty, which remains a useful safety net for genuine runtime state changes.

test/integration/epp/runtime_notification_test.go
```diff
@@ -38,6 +42,132 @@ var (
 	errTest = errors.New("test error")
 )
 
+type controlledNotificationExtractor struct {
```
hasB4K: The controlled extractor is a test-only traffic light. It tells the test "I received this object" and then waits for permission to succeed or fail. That turns a sub-second timing race into deterministic steps we can assert.

test/integration/epp/runtime_notification_test.go
```diff
@@ -38,6 +42,132 @@ var (
...
+	extractor := newControlledNotificationExtractor("initial-sync")
+	gvk := schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Pod"}
+	src := notifications.NewK8sNotificationSource(notifications.NotificationSourceType, "pod-watcher", gvk)
+	require.NoError(t, runtime.Configure(&datalayer.Config{
+		Sources: []datalayer.DataSourceConfig{
+			{Plugin: src, Extractors: []fwkplugin.Plugin{extractor}},
+		},
+	}, logger))
+	require.NoError(t, runtime.Start(ctx, mgr))
+	require.Error(t, runtime.CheckReady(), "notification runtime reported ready before manager startup")
```
hasB4K: This assertion covers the very beginning of startup: merely registering the source must not make it ready before the manager/informer has even started.

test/integration/epp/runtime_notification_test.go
```diff
@@ -38,6 +42,132 @@ var (
...
+
+func TestRuntimeNotificationInitialSyncReadiness(t *testing.T) {
+	t.Run("waits for every initial event", func(t *testing.T) {
+		ctx, mgr, apiClient, namespace, runtime, extractor := setupInitialNotificationSyncTest(t)
+		require.NoError(t, apiClient.Create(ctx, newTestPod("pod-a", namespace)))
+		require.NoError(t, apiClient.Create(ctx, newTestPod("pod-b", namespace)))
+
+		startManagerAndWaitForSync(ctx, t, mgr)
+
+		first := receiveInitialEvent(ctx, t, extractor)
```
hasB4K: Here two existing Pods enter the extractor one at a time. The test releases the first and verifies readiness is still false; only after the second completes may readiness become true. This proves cache sync alone is not what opens the gate.

test/integration/epp/runtime_notification_test.go
```diff
@@ -38,6 +42,132 @@ var (
...
+		require.Error(t, runtime.CheckReady())
+		extractor.results <- nil
+
+		require.Eventually(t, func() bool {
+			return runtime.CheckReady() == nil
+		}, eventWaitTimeout, eventPollInterval)
+		require.Len(t, extractor.GetEvents(), 2)
+	})
+
+	t.Run("accepts an empty initial list", func(t *testing.T) {
```
hasB4K: The empty-list case prevents the opposite bug: an EPP watching a namespace with zero matching objects must become ready once Kubernetes confirms the initial list is complete—it must not wait forever for an event that will never exist.

test/integration/epp/runtime_notification_test.go
```diff
@@ -38,6 +42,132 @@ var (
...
+		}, eventWaitTimeout, eventPollInterval)
+	})
+
+	t.Run("waits for a failed initial event retry", func(t *testing.T) {
+		ctx, mgr, apiClient, namespace, runtime, extractor := setupInitialNotificationSyncTest(t)
+		require.NoError(t, apiClient.Create(ctx, newTestPod("pod-a", namespace)))
+
+		startManagerAndWaitForSync(ctx, t, mgr)
+
+		first := receiveInitialEvent(ctx, t, extractor)
```
hasB4K: This simulates a real processing failure. The first attempt fails, controller-runtime delivers the same Pod again, and readiness stays false throughout. Only the successful retry removes that Pod from the startup checklist.

pkg/epp/datalayer/runtime.go
```diff
@@ -389,19 +392,48 @@ func (r *Runtime) findSourceByType(sourceType string, gvkFilter *schema.GroupVer
...
+	return nil
+}
+
+// CheckReady reports whether every notification source has processed the
+// events from its initial Kubernetes list.
+func (r *Runtime) CheckReady() error {
+	r.notificationSyncMu.RLock()
+	defer r.notificationSyncMu.RUnlock()
+
+	for _, initialSync := range r.notificationSyncs {
```
ahg-g, line 435: Is there a chance we get here before runtime.Start finishes execution? if so, then we may return ready before we get a chance to initialize notificationSyncs

pkg/epp/datalayer/k8s_bind.go
```diff
@@ -36,43 +46,104 @@ import (
...
+
+type notificationInitialSync struct {
+	upstreamSynced atomic.Bool
+	processed      synctrack.AsyncTracker[types.NamespacedName]
+	sourceName     string
+}
+
+func newNotificationInitialSync(sourceName string) *notificationInitialSync {
+	initialSync := &notificationInitialSync{sourceName: sourceName}
+	initialSync.processed.UpstreamHasSynced = initialSync.upstreamSynced.Load
```
ahg-g: pls document what this does exactly, is this overriding the `UpstreamHasSynced` func?

pkg/epp/datalayer/k8s_bind.go
ahg-g: pls document what this "starts" exactly; basically tracking this key until finish is invoked.

pkg/epp/datalayer/k8s_bind.go
ahg-g: also, what calls UpstreamHasSynced?

ahg-g: it seems the semantics of this has changed in 1.36: https://github.com/kubernetes/client-go/blob/release-1.36/tools/cache/synctrack/synctrack.go#L95

pkg/epp/datalayer/k8s_bind.go
```diff
@@ -36,43 +46,104 @@ import (
...
+	return s.processed.HasSynced()
+}
+
+type notificationInitialSyncSource struct {
+	source.SyncingSource
+	initialSync *notificationInitialSync
+}
+
+func (s *notificationInitialSyncSource) WaitForSync(ctx context.Context) error {
+	if err := s.SyncingSource.WaitForSync(ctx); err != nil {
```
ahg-g, line 149: what does this guarantee? that Reconcile was invoked on all events marked with IsInInitialList? if so, please document that.

ahg-g (commented):
I would drop AsyncTracker and would just use a `sets.Set[string]` directly, AsyncTracker is a little confusing with all the UpstreamHasSynced and HasSynced functions. It will anyway break if we update client-go to 1.36. 

If you prefer to continue to use it, please update client-go to 1.36 and adjust the code according to the new library interfaces.

ahg-g (approved):
beautiful!

PR #2579, opened 2026-08-26
Fix Data race in TestFlowControlAdmissionController_Admit

### LINKED ISSUE #2543: Data race in TestFlowControlAdmissionController_Admit under -race (parallel sub-tests share reqCtx)

ahg-g:
**What type of PR is this?**
/kind test

Fixes a data race in TestFlowControlAdmissionController_Admit test

Fixes https://github.com/llm-d/llm-d-router/issues/2543

```release-note
NONE
```


LukeAVanDrie (approved):
lgtm! thanks

vMaroon (approved):

PR #2589, opened 2026-08-27
refactor(sidecar): extract shared concurrent prefill/decode dispatch helper

### LINKED ISSUE #2587: Sidecar: Mooncake/SGLang connectors duplicate concurrent prefill/decode dispatch
`handleMooncakeConcurrentRequests`,  and `handleSGLangConcurrentRequests` in `pkg/sidecar/proxy` each reimplement the same pattern nearly line-for-line: an async prefill goroutine with panic recovery, a synchronous decode on the calling goroutine, and identical OTel span/timing bookkeeping around both. 
Extract into a shared helper so the three connectors differ only in their protocol-specific request bodies.

roytman:
**What type of PR is this?**
/kind cleanup

Mooncake and SGLang each reimplemented the same concurrent-dispatch pattern
nearly line for line: an async prefill goroutine with panic recovery, a
synchronous decode on the calling goroutine, and identical OTel span/timing
bookkeeping around both. Extracts it into a single `runConcurrentPD` helper
in `pkg/sidecar/proxy/connector_pd_common.go`; each connector now supplies
only its protocol-specific request bodies and connector name.

Fixes SGLang's inline status-code check to use the existing
`isHTTPError` helper instead of a hand-rolled range comparison.

The per-request prefill-completion log line, previously at TRACE for
Mooncake and SGLang, is now DEBUG for both, matching the level P2P's
sequential path already logs at.

Fixes #2587

```release-note
Sidecar: prefill-completion log lines for the Mooncake and SGLang P/D connectors move from TRACE to DEBUG verbosity, matching the OffloadingConnector (P2P) connector's existing level.
```
Test plan:
- [x] Full sidecar test suite (make test-filter PATTERN=. TYPE=sidecar), 141/141 specs pass
- [x]  make format / make lint clean


pkg/sidecar/proxy/connector_pd_common.go
```diff
@@ -0,0 +1,140 @@
...
+		prefillDuration := time.Since(prefillStart)
+		prefillSpan.SetAttributes(
+			attribute.Int("llm_d.pd_proxy.prefill.status_code", pw.statusCode),
+			attribute.Float64("llm_d.pd_proxy.prefill.duration_ms", float64(prefillDuration.Milliseconds())),
+		)
+		if isHTTPError(pw.statusCode) {
+			prefillSpan.SetStatus(codes.Error, "prefill request failed")
+		}
+		s.logger.V(logging.DEBUG).Info("concurrent-dispatch prefill request completed", "connector", connector, "status", pw.statusCode)
+	}()
```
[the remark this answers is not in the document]
roytman: I think a unified DEBUG level is preferred.

pkg/sidecar/proxy/connector_pd_common.go
```diff
@@ -0,0 +1,140 @@
...
+	if prepareRequests != nil {
+		prepareRequests(prefillReq, decodeReq)
+	}
+
+	// Prefill runs in a goroutine: response is discarded (status/duration only).
+	// Decode runs on the calling goroutine: writes the actual response back via w.
+	ctx, prefillSpan := tracer.Start(ctx, "llm_d.pd_proxy.prefill",
+		trace.WithSpanKind(trace.SpanKindInternal),
+	)
+	prefillSpan.SetAttributes(
```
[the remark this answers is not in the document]
roytman: set to "prefill" and "decode".

pkg/sidecar/proxy/connector_pd_common.go
```diff
@@ -0,0 +1,145 @@
...
+// runConcurrentPD fires the prefill and decode legs of a concurrent-dispatch
+// P/D protocol (Mooncake, OffloadingConnector P2P, SGLang) in parallel:
+// prefill runs in a goroutine and its response is discarded (only status and
+// duration are recorded on its span), while decode runs on the calling
+// goroutine and streams its response to w. connector is recorded on both
+// spans to identify the protocol. The per-request "prefill request
+// completed" log line is unified at DEBUG for all three connectors,
+// deliberately dropping the prior per-connector split (Mooncake/SGLang at
+// TRACE, P2P at DEBUG): three verbosity levels for structurally identical
+// code was the inconsistency worth fixing, not a behavior worth preserving.
```
elevran: This reads as changelog/rationale rather than a description of current behavior - the "X, not Y" construction and "prior per-connector split" framing are better avoided in comments - focus on the current behvaior not historical changes. Suggest something like: "The prefill-completion log line is emitted at DEBUG for all three connectors."

pkg/sidecar/proxy/connector_pd_common.go
```diff
@@ -0,0 +1,145 @@
...
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
+
+	logging "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
```
elevran: redundant alias, `logging` already matches the package name.

pkg/sidecar/proxy/connector_p2p.go
```diff
@@ -32,7 +32,7 @@ import (
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
 
-	"github.com/llm-d/llm-d-router/pkg/common/observability/logging"
+	logging "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
```
elevran: redundant alias - not needed.

elevran (commented):
Clean extraction, matches the linked issue and preserves the existing tracing/timing/panic-recovery behavior for Mooncake and SGLang. A couple of small things worth a look before merge: the new helper's doc comment reads more like a changelog than a description of current behavior (see inline), and the log level for prefill completion moved from TRACE to DEBUG for Mooncake/SGLang, which is a small but real behavior change despite the PR description saying "No behavior change" - so confirming that's intentional. Also, P2P is mentioned in the description as part of the duplicated pattern, but it was already moved to sequential dispatch in #2451 and isn't touched here - might be worth updating the description so reviewers don't look for P2P changes in the diff?

pkg/sidecar/proxy/connector_pd_common.go
roytman: Sorry, I missed that. Strange, I've seen this redundant alias pattern in other PRs. Opened https://github.com/llm-d/llm-d-router/issues/2757 to prevent it automatically.

pkg/sidecar/proxy/connector_pd_common.go
```diff
@@ -0,0 +1,138 @@
...
+		prefillHandler.ServeHTTP(pw, prefillReq)
+		prefillDuration := time.Since(prefillStart)
+		prefillSpan.SetAttributes(
+			attribute.Int("llm_d.pd_proxy.prefill.status_code", pw.statusCode),
+			attribute.Float64("llm_d.pd_proxy.prefill.duration_ms", float64(prefillDuration.Milliseconds())),
+		)
+		if isHTTPError(pw.statusCode) {
+			prefillSpan.SetStatus(codes.Error, "prefill request failed")
+		}
+		s.logger.V(logging.DEBUG).Info("concurrent-dispatch prefill request completed", "connector", connector, "status", pw.statusCode)
```
elevran, line 97: The earlier review asked whether moving this log from TRACE to DEBUG was intentional. The commit removes the explanatory paragraph but keeps the DEBUG level. Either revert to TRACE to match the prior behavior, or acknowledge here that the new level is intentional.

pkg/sidecar/proxy/connector_pd_common.go
```diff
@@ -0,0 +1,138 @@
...
+// and its response is discarded (only status and duration are recorded on
+// its span), while decode runs on the calling goroutine and streams its
+// response to w. connector is recorded on both spans to identify the
+// protocol.
+func (s *Server) runConcurrentPD(
+	w http.ResponseWriter,
+	r *http.Request,
+	prefillBody, decodeBody []byte,
+	prefillHost, connector string,
+	prepareRequests func(prefillReq, decodeReq *http.Request),
```
elevran, line 46: The trimmed function comment is fine as a high-level summary, but `prepareRequests` no longer carries its prior one-line note (e.g., "lets Mooncake inject the DP-rank header per request"). A future caller benefits from a one-line WHY at the parameter declaration.

elevran (commented):
PR description still mentions P2P as part of the duplicated pattern, but the in-code doc comment was corrected: P2P is sequential after #2451 and does not call runConcurrentPD, so there is no P2P diff in this PR. Worth trimming the description so reviewers landing from it do not look for a P2P change that is not there.

Three comments inline.

pkg/sidecar/proxy/connector_pd_common.go
roytman: Yes it is intentional to be in line with other connectors

elevran (approved):

PR #2591, opened 2026-08-27
fix(sidecar): forward untouched request fields as raw JSON to keep key order

### LINKED ISSUE #2590: sidecar: re-marshaling the request body as map[string]any reorders nested JSON keys (tools schema)
## What happened

The P/D sidecar decodes the request body into `map[string]any` and marshals it again for the prefill and decode legs (`pkg/sidecar/proxy/connector_nixlv2.go`, and the same pattern in `connector_shared_storage.go`, `connector_p2p.go`, `connector_mooncake.go`, `connector_sglang.go`, `decode.go`). `encoding/json` sorts map keys at every depth when marshaling a map, so the key order of every nested object the sidecar does not touch is lost.

`tools[].function.parameters.properties` is the visible case. Chat templates render `tools` into the prompt verbatim (`{{ tools | tojson }}`), so the model sees the parameter schema in alphabetical order instead of the order the client sent. Schema order matches the training distribution; a reordered schema can degrade tool-calling accuracy (missing or mis-mapped parameters).

Providers are tested for this. The [MiniMax-Provider-Verifier](https://github.com/MiniMax-AI/MiniMax-Provider-Verifier) `Scenario-Check-Pass-Rate` metric checks whether the provider preserves the original JSON key order in tool definitions and is expected to be 100%; reordering `parameters.properties` (alphabetical sorting) fails it. A P/D deployment fronted by the sidecar fails this check.

## Reproduction

Pure `encoding/json`, no sidecar needed:

```go
original := []byte(`{"tools":[{"type":"function","function":{"name":"example","parameters":{"properties":{"some-parameter":{"type":"string"},"xyz":{"type":"string"},"123":{"type":"string"},"another-parameter":{"type":"string"}}}}}]}`)
var req map[string]any
json.Unmarshal(original, &req)
req["kv_transfer_params"] = map[string]any{"do_remote_decode": true}
out, _ := json.Marshal(req)
fmt.Println(string(out))
```

Output orders `properties` as `123, another-parameter, some-parameter, xyz`.

## Expected

Fields the sidecar does not modify (`tools`, `messages` content, `response_format`, `chat_template_kwargs`, ...) are forwarded byte-for-byte to prefill and decode.

## Proposed fix

Two approaches, each in its own PR. They fix the same issue; only one should merge.

### A. Decode only inspected fields, keep the rest as `json.RawMessage` (#2591)

Decode the top level into `map[string]json.RawMessage`. Fields the sidecar reads or rewrites (`stream`, `max_tokens`, `kv_transfer_params`, ...) are decoded into Go values; every other field stays a `json.RawMessage`, so `json.Marshal` emits its original bytes. `messages` is decoded lazily by the two code paths that read it (EC item extraction, chunked decode).

- Keeps the existing `map[string]any` model; connectors change one call each.
- Requires an explicit allowlist of inspected fields. A field read by a connector but missing from the list decodes as `json.RawMessage`, so a `req[field].(bool)` type assertion silently reports the field as absent.

### B. Edit the raw body in place with `sjson` (#2595)

Keep the validated body as `[]byte`. Connectors apply their edits with `sjson.SetBytes` / `DeleteBytes` through a small `requestBody` helper and read values with `gjson`. Every field the sidecar does not set is forwarded byte-for-byte, including number formatting and escaping.

- No allowlist to maintain.
- NIXLv2 no longer saves and restores `stream` and token-limit fields between legs; each leg is built from the original bytes.
- Larger diff: every connector's body handling changes from map access to path edits. `gjson` and `sjson` were already indirect dependencies and become direct ones.

Both PRs include a NIXLv2 test asserting that `tools` and `messages` reach prefill and decode byte-for-byte.

rebel-jinhwan:
**What type of PR is this?**
/kind bug

The sidecar decodes the request body into `map[string]any` and marshals it again for the prefill and decode legs. `encoding/json` sorts map keys at every depth on Marshal, so nested key order in fields the sidecar never touches is lost. `tools[].function.parameters.properties` is the visible case: chat templates render `tools` into the prompt verbatim, so the model sees the schema in alphabetical order instead of the order the client sent.

Providers are tested for this. The [MiniMax-Provider-Verifier](https://github.com/MiniMax-AI/MiniMax-Provider-Verifier) `Scenario-Check-Pass-Rate` metric checks whether the provider preserves the original JSON key order in tool definitions and is expected to be 100%; reordering `parameters.properties` (alphabetical sorting) fails it. A P/D deployment fronted by the sidecar fails this check today.

`decodeRequestBody` decodes only the top-level fields the sidecar reads or rewrites (`stream`, `max_tokens`, `kv_transfer_params`, ...). Every other field stays a `json.RawMessage`, so its bytes are forwarded unchanged. `messages` is also kept raw and decoded element-by-element by `requestMessages` only when a connector needs it. All connectors that parse the body (NIXLv2, shared storage, P2P, Mooncake, SGLang, EC, chunked decode) go through it.

Fixes #2590

**Test plan**:
- [x] Fields not inspected by the sidecar are kept as raw bytes and marshal back unchanged; inspected fields are decoded
- [x] NIXLv2 prefill and decode request bodies carry `tools` byte-for-byte, including nested key order
- [x] Chunked decode keeps the client's messages byte-for-byte across multiple chunks
- [x] A `null` request body is rejected; a `null` messages field reads as absent

**Release note**:
```release-note
The P/D sidecar forwards request fields it does not modify byte-for-byte, so nested JSON key order (for example in `tools[].function.parameters`) is preserved when proxying to prefill and decode.
```

### DIFF
--- a/pkg/sidecar/proxy/decode.go
+++ b/pkg/sidecar/proxy/decode.go
@@ -423,7 +423,7 @@ func appendChunkToRequest(req map[string]any, text string) {
 	if text == "" {
 		return
 	}
-	messages, _ := req[requestFieldMessages].([]any)
+	messages, _ := requestMessages(req)
 	req[requestFieldMessages] = append(messages, map[string]any{
 		requestFieldRole:    roleAssistant,
 		requestFieldContent: text,
--- a/pkg/sidecar/proxy/decode.go
+++ b/pkg/sidecar/proxy/decode.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
@@ -234,7 +235,7 @@ func (s *Server) runChunkedDecodeFromMap(w http.ResponseWriter, r *http.Request,
 		// Append the generated text to the request so the next chunk continues
 		// from where this one left off.
 		s.logger.V(logging.TRACE).Info("chunked decode: appending chunk text to request", "chunkText", chunkText)
-		appendChunkToRequest(completionRequest, chunkText)
+		appendChunkToRequest(s.logger, completionRequest, chunkText)
 	}
 
 	span.SetAttributes(
@@ -418,16 +419,26 @@ func extractChoiceText(choice map[string]any) string {
 }
 
 // appendChunkToRequest appends the generated text from a chunk to the request
-// so the next chunk continues from where this one left off.
-func appendChunkToRequest(req map[string]any, text string) {
+// so the next chunk continues from where this one left off. The messages the
+// client sent are appended to as raw bytes, keeping their key order intact
+// across chunks.
+func appendChunkToRequest(logger logr.Logger, req map[string]any, text string) {
 	if text == "" {
 		return
 	}
-	messages, _ := requestMessages(req)
-	req[requestFieldMessages] = append(messages, map[string]any{
+	messages, err := requestMessages(req)
+	if err != nil {
+		logger.V(logging.DEBUG).Info("chunked decode: cannot read request messages", "error", err)
+	}
+	chunk, err := json.Marshal(map[string]any{
 		requestFieldRole:    roleAssistant,
 		requestFieldContent: text,
 	})
+	if err != nil {
+		logger.V(logging.DEBUG).Info("chunked decode: cannot encode chunk text", "error", err)
+		return
+	}
+	req[requestFieldMessages] = append(messages, chunk)
 }
 
 // toInt converts a JSON number value (float64, int, or json.Number) to int.
--- a/pkg/sidecar/proxy/proxy_helpers.go
+++ b/pkg/sidecar/proxy/proxy_helpers.go
@@ -197,21 +197,24 @@ var inspectedRequestFields = map[string]struct{}{
 	requestFieldAddGenerationPrompt:  {},
 }
 
-// requestMessages returns the request's messages array, decoding it from raw
-// bytes on first use. The second return value is false when the field is
-// absent or not an array.
-func requestMessages(req map[string]any) ([]any, bool) {
+// requestMessages returns the request's messages, decoding the array on first
+// use and keeping each element as raw bytes so that re-marshaling the request
+// preserves the key order inside every message. An absent field yields a nil
+// slice and no error.
+func requestMessages(req map[string]any) ([]json.RawMessage, error) {
 	switch v := req[requestFieldMessages].(type) {
-	case []any:
-		return v, true
+	case nil:
+		return nil, nil
+	case []json.RawMessage:
+		return v, nil
 	case json.RawMessage:
-		var messages []any
+		var messages []json.RawMessage
 		if err := json.Unmarshal(v, &messages); err != nil {
-			return nil, false
+			return nil, err
 		}
-		return messages, true
+		return messages, nil
 	default:
-		return nil, false
+		return nil, fmt.Errorf("messages is %T, want a JSON array", v)
 	}
 }
 
--- a/pkg/sidecar/proxy/proxy_helpers.go
+++ b/pkg/sidecar/proxy/proxy_helpers.go
@@ -195,7 +195,24 @@ var inspectedRequestFields = map[string]struct{}{
 	requestFieldCacheHitThreshold:    {},
 	requestFieldContinueFinalMessage: {},
 	requestFieldAddGenerationPrompt:  {},
-	requestFieldMessages:             {},
+}
+
+// requestMessages returns the request's messages array, decoding it from raw
+// bytes on first use. The second return value is false when the field is
+// absent or not an array.
+func requestMessages(req map[string]any) ([]any, bool) {
+	switch v := req[requestFieldMessages].(type) {
+	case []any:
+		return v, true
+	case json.RawMessage:
+		var messages []any
+		if err := json.Unmarshal(v, &messages); err != nil {
+			return nil, false
+		}
+		return messages, true
+	default:
+		return nil, false
+	}
 }
 
 // decodeRequestBody parses a JSON object body. Fields in inspectedRequestFields
--- a/pkg/sidecar/proxy/proxy_helpers.go
+++ b/pkg/sidecar/proxy/proxy_helpers.go
@@ -218,12 +218,14 @@ func requestMessages(req map[string]any) ([]json.RawMessage, error) {
 	}
 }
 
-// decodeRequestBody parses a JSON object body. Fields in inspectedRequestFields
-// are decoded to Go values; all others are kept as json.RawMessage.
+// decodeRequestBody parses a JSON object body, applying inspectedRequestFields.
 func decodeRequestBody(raw []byte) (map[string]any, error) {
 	var fields map[string]json.RawMessage
 	if err := json.Unmarshal(raw, &fields); err != nil {
 		return nil, err
+	}
+	if fields == nil {
+		return nil, errors.New("request body is not a JSON object")
 	}
 	parsed := make(map[string]any, len(fields))
 	for k, v := range fields {
--- a/pkg/sidecar/proxy/proxy_helpers.go
+++ b/pkg/sidecar/proxy/proxy_helpers.go
@@ -169,11 +169,55 @@ func bodyAsJSON(r *http.Request) ([]byte, map[string]any, error) {
 	if err != nil {
 		return nil, nil, err
 	}
-	var parsed map[string]any
-	if err := json.Unmarshal(raw, &parsed); err != nil {
+	parsed, err := decodeRequestBody(raw)
+	if err != nil {
 		return nil, nil, fmt.Errorf("%w: %w", errInvalidJSON, err)
 	}
 	return raw, parsed, nil
+}
+
+// inspectedRequestFields lists the top-level request fields the sidecar reads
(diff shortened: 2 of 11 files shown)


test/sidecar/mock/chat_completions_handler.go
```diff
@@ -53,6 +53,7 @@ type ChatCompletionHandler struct {
 	MoRIIOWriteMode     bool
 	RequestCount        atomic.Int32
 	CompletionRequests  []map[string]any
+	CompletionRawBodies [][]byte
 	CompletionHeaders   []http.Header
 	CompletionResponses []map[string]any
```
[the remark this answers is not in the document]
rebel-jinhwan: Added `GetCompletionRawBodies()` mirroring the other captured fields; the test uses it.

pkg/sidecar/proxy/proxy_helpers.go
```diff
@@ -169,13 +169,57 @@ func bodyAsJSON(r *http.Request) ([]byte, map[string]any, error) {
...
+	requestFieldMaxOutputTokens:      {},
+	requestFieldMinTokens:            {},
+	requestFieldSamplingParams:       {},
+	requestFieldStream:               {},
+	requestFieldStreamOptions:        {},
+	requestFieldCacheHitThreshold:    {},
+	requestFieldContinueFinalMessage: {},
+	requestFieldAddGenerationPrompt:  {},
+	requestFieldMessages:             {},
+}
```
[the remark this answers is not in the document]
rebel-jinhwan: Good catch. `messages` is now kept as `json.RawMessage` and decoded on first use via `requestMessages()` in the two consumers (chunked decode, EC fanout). The integration test also asserts `messages` is forwarded byte-for-byte.

pkg/sidecar/proxy/decode.go
```diff
@@ -423,7 +423,7 @@ func appendChunkToRequest(req map[string]any, text string) {
 	if text == "" {
 		return
 	}
-	messages, _ := req[requestFieldMessages].([]any)
+	messages, _ := requestMessages(req)
```
roytman: This fixes the single-chunk case, but from the second chunk onward the byte-for-byte guarantee still breaks: `requestMessages` decodes `messages` out of `json.RawMessage` into `[]any` and writes the decoded-and-appended slice back into `req`. Every later chunk's `maps.Clone` + `json.Marshal` then re-encodes the entire messages array generically, re-sorting keys inside every prior message object — the same bug class #2590 is filed against, just relocated to chunk 2+. tools stays safe since it's never touched here, so the reported symptom is fixed, but the PR's test name and release note ("messages... byte-for-byte") overclaim for `DecodeChunkSize > 0` with multi-chunk generations. 
Worth either scoping the claim in the release note/test name, or adding a multi-chunk test that pins the known limitation.

pkg/sidecar/proxy/proxy_helpers_test.go
```diff
@@ -0,0 +1,45 @@
...
+		Expect(parsed[requestFieldMaxTokens]).To(BeNumerically("==", 5))
+		Expect(parsed["tools"]).To(Equal(json.RawMessage(tools)))
+
+		out, err := json.Marshal(parsed)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(string(out)).To(ContainSubstring(`"tools":` + tools))
+	})
+
+	It("rejects non-object bodies", func() {
+		_, err := decodeRequestBody([]byte(`[1,2]`))
```
roytman, line 42: This only proves an array is rejected. `decodeRequestBody([]byte("null"))` returns `(map[string]any{}, nil)` with no error, `json.Unmarshal` into a map target treats top-level `null` as a successful no-op - so a `null` body currently sails through as an empty, valid request. That's arguably safer than HEAD (which panics on the first nil-map write), but it contradicts what this test's name implies it covers. Suggest adding an explicit case for `null` so the behavior is pinned intentionally rather than incidentally.

pkg/sidecar/proxy/proxy_helpers.go
```diff
@@ -169,13 +169,74 @@ func bodyAsJSON(r *http.Request) ([]byte, map[string]any, error) {
...
+	requestFieldMinTokens:            {},
+	requestFieldSamplingParams:       {},
+	requestFieldStream:               {},
+	requestFieldStreamOptions:        {},
+	requestFieldCacheHitThreshold:    {},
+	requestFieldContinueFinalMessage: {},
+	requestFieldAddGenerationPrompt:  {},
+}
+
+// requestMessages returns the request's messages array, decoding it from raw
```
roytman: **nit**: for `"messages": null, json.Unmarshal([]byte("null"), &messages)` succeeds with `messages == nil`, so this returns `(nil, true)`, but `null` is neither "absent" nor "an array," so the comment's own stated rule doesn't hold for this input. No observed downstream harm (both callers handle a `nil` slice fine), just worth tightening the wording, e.g. "false when the field is absent or cannot be decoded as a slice."

pkg/sidecar/proxy/proxy_helpers.go
```diff
@@ -169,13 +169,74 @@ func bodyAsJSON(r *http.Request) ([]byte, map[string]any, error) {
...
+		if err := json.Unmarshal(v, &messages); err != nil {
+			return nil, false
+		}
+		return messages, true
+	default:
+		return nil, false
+	}
+}
+
+// decodeRequestBody parses a JSON object body. Fields in inspectedRequestFields
```
roytman: **nit**: This restates the same fact that the `inspectedRequestFields` comment two blocks above already states ("decodeRequestBody decodes only these; every other field stays a json.RawMessage"). 
Suggestion: trim this to something that doesn't duplicate, e.g. 
```go
// decodeRequestBody parses a JSON object body, applying inspectedRequestFields.
```

pkg/sidecar/proxy/proxy_helpers.go
```diff
@@ -169,13 +169,74 @@ func bodyAsJSON(r *http.Request) ([]byte, map[string]any, error) {
...
+
+// requestMessages returns the request's messages array, decoding it from raw
+// bytes on first use. The second return value is false when the field is
+// absent or not an array.
+func requestMessages(req map[string]any) ([]any, bool) {
+	switch v := req[requestFieldMessages].(type) {
+	case []any:
+		return v, true
+	case json.RawMessage:
+		var messages []any
```
roytman: Before this PR, a bad `messages` shape was just a failed type assertion with no error value to lose. Now a malformed-but-syntactically-valid shape (e.g. `"messages": {}`) produces a real, descriptive `json.Unmarshal`error here that's discarded on the spot. Both callers (`appendChunkToRequest`, `extractMMItems`) have a logger available, so worth adding a `logger.V(logging.DEBUG)` line at the call site so a client-side bad-shape bug doesn't vanish without a trace.

roytman (commented):
Thanks for finding this and putting together a clear fix, nice catch on the `encoding/json` key-sorting behavior.

I left several inline comments, mostly around the chunked-decode path and a couple of test/comment precision points, nothing blocking the overall approach.

One more thing worth flagging: `pkg/coordinator` has the same underlying issue. `handleInference` decodes the request body into `map[string]any` and it gets re-marshaled a few times through the pipeline (encode, prefill, decode, render steps), so `tools[].function.parameters` would reorder there too. Not asking you to take that on here, it's a separate component and probably deserves its own issue. If you'd like to extend the fix there too, in this PR or a follow-up, happy to see it, but it's not required.

pkg/sidecar/proxy/decode.go
rebel-jinhwan: Fixed rather than scoped. `requestMessages` now returns `[]json.RawMessage` and this appends the assistant message as raw bytes, so chunk 2 onward no longer re-encodes the earlier messages. Added a test that appends two chunks and asserts the client's message survives byte-for-byte; it fails against the previous implementation.

https://github.com/llm-d/llm-d-router/pull/2591/commits/422c62b3

pkg/sidecar/proxy/proxy_helpers_test.go
rebel-jinhwan: `decodeRequestBody` now rejects a top-level `null` (`json.Unmarshal` leaves the map nil), so the test name matches what it covers, and the `null` case is asserted explicitly.

https://github.com/llm-d/llm-d-router/pull/2591/commits/5adbd6ea

pkg/sidecar/proxy/proxy_helpers.go
rebel-jinhwan: Reworded. The bool is gone (it returns an error now), and a `null` messages field is pinned by a test as equivalent to an absent one.

https://github.com/llm-d/llm-d-router/pull/2591/commits/422c62b3

pkg/sidecar/proxy/proxy_helpers.go
rebel-jinhwan: Trimmed to `// decodeRequestBody parses a JSON object body, applying inspectedRequestFields.`

https://github.com/llm-d/llm-d-router/pull/2591/commits/5adbd6ea

pkg/sidecar/proxy/proxy_helpers.go
rebel-jinhwan: `requestMessages` returns the error now, and both callers log it at `V(logging.DEBUG)`; `appendChunkToRequest` and `extractMMItems` take a `logr.Logger` from their callers.

https://github.com/llm-d/llm-d-router/pull/2591/commits/422c62b3

roytman (approved):

elevran (approved):
/approve on @roytman LGTM

PR #2600, opened 2026-08-27
Add audio content block support to tokenizer for multimodal routing

RishabhSaini:
## Summary

Enable prefix-cache and token-load scorers to handle audio content in chat completions requests. Previously audio blocks were dropped during tokenization, making them invisible to routing decisions.

## Changes

- Add AudioBlock, AudioURLBlock types to kvcache tokenization types
- Add AudioURL field to ContentBlock in requesthandling types
- Update tokenizer to pass audio blocks through to kvcache types
- Add audioEstimator with dynamic/static modes for token estimation
- Update estimate backend to hash audio URL/data for prefix caching
- Update PlainText() to include audio URLs for text-based matching
- Add audio metadata headers (x-llm-d-audio-duration-seconds, etc.)

## Audio estimation

Uses mid-range compressed audio (~8000 bytes/sec) for base64 data, defaulting to 10-second duration for URL-based audio.

## Testing

- All existing tests pass
- Build compiles successfully

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
@@ -339,15 +339,6 @@ func appendMMAsset(out []byte, features []fwkrh.MultiModalFeature, modality fwkr
 	return out, features
 }
 
-// assetPlaceholderCount derives a deterministic placeholder count (>= 1) from an
-// asset's byte length for modalities without a dedicated estimator.
-func assetPlaceholderCount(dataLen int) int {
-	if n := (dataLen + bytesPerToken - 1) / bytesPerToken; n > 0 {
-		return n
-	}
-	return 1
-}
-
 // packBytes packs bytes into little-endian uint32 tokens (zero-padded tail).
 // Reinterpreting them reproduces the input, so locality keys are unchanged.
 func packBytes(raw []byte) []uint32 {
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
@@ -42,12 +42,17 @@ const bytesPerToken = 4
 type estimateBackend struct {
 	img imageEstimator
 	vid videoEstimator
+	aud audioEstimator
 }
 
 // parseMMMetadataHeaders reads the x-llm-d-* request headers into an mmMetadata.
-// Only video is populated today; image and audio parsing slot in here later.
+// Video and audio are populated from headers; image parsing follows the same
+// pattern when its headers are added.
 func parseMMMetadataHeaders(headers map[string]string) mmMetadata {
-	return mmMetadata{video: parseVideoMetadataHeaders(headers)}
+	return mmMetadata{
+		video: parseVideoMetadataHeaders(headers),
+		audio: parseAudioMetadataHeaders(headers),
+	}
 }
 
 // mmMetadataCtxKey keys the request-scoped mmMetadata on the context, carrying it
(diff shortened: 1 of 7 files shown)


pkg/kvcache/tokenization/types/types.go
zetxqx, line 101: no need to add update the plainText here, this is dead code no callers. Feel free to delete the whole function in this PR or separate PR/issue

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
```diff
@@ -229,9 +258,11 @@ func (b estimateBackend) appendChatMessage(out []byte, features []fwkrh.MultiMod
 			out, features = appendMMAsset(out, features, fwkrh.ModalityImage, block.ImageURL.URL, b.img.placeholderCount(block.ImageURL.URL))
 		case "video_url":
 			out, features = appendMMAsset(out, features, fwkrh.ModalityVideo, block.VideoURL.URL, b.vid.placeholderCount(meta.video))
-		case "input_audio", "audio_url":
-			data := block.InputAudio.Data + block.InputAudio.Format
-			out, features = appendMMAsset(out, features, fwkrh.ModalityAudio, data, assetPlaceholderCount(len(data)))
+		case "audio_url":
+			out, features = appendMMAsset(out, features, fwkrh.ModalityAudio, block.AudioURL.URL, b.aud.placeholderCount(block.AudioURL.URL, false, meta.audio))
+		case "input_audio", "audio":
```
zetxqx, line 263: I think we can remove "audio" as openAI API only supports https://developers.openai.com/api/docs/guides/audio?example=audio-in "input_audio" as the input audio field.

For audio_url I know vllm supports it.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
```diff
@@ -229,9 +258,11 @@ func (b estimateBackend) appendChatMessage(out []byte, features []fwkrh.MultiMod
...
+			data := block.InputAudio.Data
```
zetxqx, line 264: this logic is changed a little bit, originally we have `data := block.InputAudio.Data + block.InputAudio.Format` . @capri-xiyue  is this okay?

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/mm_estimate.go
```diff
@@ -183,6 +184,15 @@ type videoMetadata struct {
 	fps           float64 // source frames per second
 }
 
+// audioMetadata carries per-request audio properties parsed from the
+// x-llm-d-audio- request headers. A zero field means "not provided"; the
+// estimator falls back per field to configuration and then built-in defaults.
+type audioMetadata struct {
+	duration   float64 // seconds
+	sampleRate int     // Hz
+	channels   int
```
zetxqx, line 193: currently these two fields are not used any where, are we planning to add more logic based on these field? if not, we can remove them for now.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
capri-xiyue, line 263: Why do we have a `audio` in this case besides `input_audio`?

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
capri-xiyue: I don't think previously the format is used in data. So it should be fine

zetxqx: Image Previously the format is actually used but should be fine I guess? since the data should be already different if format is different.

capri-xiyue: Got it. I think for the content pass to appendMMAsset, we still need both data and format. Unless in extreme case, data is same but format is different.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/mm_estimate.go
```diff
@@ -343,3 +353,78 @@ func (e videoEstimator) tokensPerFrame(meta videoMetadata) int {
...
+	// Use header-provided duration if available
+	duration := meta.duration
+	if duration <= 0 {
+		// Fall back to file-size estimation for base64 data
+		duration = float64(defaultAudioDuration)
+		if isBase64 && len(dataOrURL) > 0 {
+			// Rough estimation: base64 expands 3 bytes to 4 chars
+			// Use mid-range estimate for compressed audio (~8000 bytes/sec)
+			// Base64 size * 3/4 = raw bytes, then / 8000 = seconds
+			rawBytes := float64(len(dataOrURL)) * 0.75
```
capri-xiyue, line 417: Could you verify the accuracy of this estimation using actual data? If the discrepancy is too high, we can simply rely on user-submitted audio metadata; otherwise, we'll stick to the defaults. Similar to how we did it for video.

PR #2603, opened 2026-08-28
fix(chart): allow overriding the EPP Deployment strategy so multi-replica EPP can satisfy helm --wait

### LINKED ISSUE #2568: [Helm] --wait never completes for multi-replica active-passive EPP
**What happened**:

When the gateway chart is configured with `router.epp.replicas: 2`, the chart enables leader election by default. The leader becomes Ready, while the standby intentionally remains NotReady.

As a result, the EPP Deployment remains at `1/2 Ready` and `Available=False` with `MinimumReplicasUnavailable`. A Helm install or upgrade using `--wait` never completes. With Flux Helm Controller, the release remains `pending-upgrade` until the HelmRelease timeout, then enters rollback/remediation. While the release is pending, subsequent EPP configuration or resource updates cannot start a new Helm upgrade.

Observed state:

```text
NAME                         READY   STATUS
<release>-epp-<hash>-leader  1/1     Running
<release>-epp-<hash>-standby 0/1     Running

Deployment:
replicas: 2
updatedReplicas: 2
readyReplicas: 1
availableReplicas: 1
unavailableReplicas: 1
Available=False, reason=MinimumReplicasUnavailable

Helm history:
REVISION  STATUS
N         deployed
N+1       pending-upgrade
```

The standby readiness probe repeatedly returns:

```text
gRPC readiness check not serving
service=readiness
isLeader=false
```

The EPP data path itself remains functional because the Service contains the Ready leader endpoint. The issue is the mismatch between active-passive readiness semantics and Helm's Deployment readiness requirement.

A chart-supported, default multi-replica active-passive deployment should be installable and upgradeable with normal Helm wait behavior.

Possible acceptable outcomes include:

- provide chart/deployment semantics that allow Helm to recognize one active leader plus warm standbys as healthy; or
- explicitly document that Helm `--wait` / Flux wait is incompatible with active-passive replicas, provide a supported no-wait configuration, and define an alternative health gate based on an elected leader and a Ready EPP Service endpoint.

The fix should not make standby replicas receive normal Service traffic before they are elected leader.

Prerequisites: Gateway API Inference Extension CRDs and a valid model-server label selector.

1. Install the gateway chart with one EPP replica and wait enabled:

```bash
helm upgrade --install ha-wait-test \
  oci://ghcr.io/llm-d/charts/llm-d-router-gateway \
  --version v0.9.0 \
  --namespace llm-d-test \
  --create-namespace \
  --wait \
  --timeout 10m \
  --set router.modelServers.matchLabels.app=vllm \
  --set router.epp.replicas=1
```

2. Upgrade to two replicas, leaving leader election at its chart default:

```bash
helm upgrade ha-wait-test \
  oci://ghcr.io/llm-d/charts/llm-d-router-gateway \
  --version v0.9.0 \
  --namespace llm-d-test \
  --wait \
  --timeout 10m \
  --set router.modelServers.matchLabels.app=vllm \
  --set router.epp.replicas=2
```

3. Observe that one EPP pod remains `0/1`, the Deployment remains `Available=False`, and the Helm revision remains `pending-upgrade` until timeout.

The same behavior can be reproduced with Flux Helm Controller by omitting `spec.install.disableWait` and `spec.upgrade.disableWait`, then changing `router.epp.replicas` from 1 to 2.

This appears to be an integration gap rather than an EPP process failure:

- The Endpoint Picker Protocol intentionally defines a follower as `NOT_SERVING` for readiness.
- kubernetes-sigs/gateway-api-inference-extension#1264 introduced the behavior where only the elected leader is Ready and other replicas remain NotReady.
- llm-d/llm-d-router#1884 explicitly keeps readiness and ext_proc health leader-gated; non-leaders remain NotServing.
- Static inspection of the v0.10.0 chart template shows the same relevant combination: leader election is enabled for `replicas > 1`, the multi-replica readiness probe uses the `readiness` gRPC service, and the Deployment strategy remains `Recreate`.

Related links:

- https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/1264
- https://github.com/llm-d/llm-d-router/pull/1884
- https://github.com/llm-d/llm-d-router/blob/v0.10.0/config/charts/routerlib/templates/_deployment.yaml
- https://github.com/llm-d/llm-d-router/issues/1593
- https://github.com/llm-d/llm-d-router/issues/1892

**Environment**:
- Kubernetes version: server v1.33.6, client v1.33.13
- llm-d-router version: gateway chart v0.9.0, EPP image v0.9.0; v0.10.0 chart statically inspected
- Cloud provider or hardware configuration: self-managed Kubernetes cluster
- Install tools: Helm v3.16.1; Flux Helm Controller v1.5.5
- Others: Gateway mode with AgentGateway provider; EPP Deployment uses the chart-default `Recreate` strategy

almogtavor:
**What type of PR is this?**

/kind bug

With `router.epp.replicas > 1` the chart enables leader election. Only the leader
passes its readiness probe; standbys answer the `readiness` gRPC service with
`NOT_SERVING` by design, so they never become Service endpoints. The EPP
Deployment therefore settles at `readyReplicas: 1` of `N`.

Because the chart pins `strategy: Recreate`, Kubernetes forces `maxUnavailable`
to `0`, so the Deployment reports `Available=False` / `MinimumReplicasUnavailable`
permanently. Anything gating on Deployment availability blocks: `helm --wait`
times out, and a Flux `HelmRelease` stays `pending-upgrade` until its timeout and
then rolls back -- and while it is pending, no further EPP change can start a new
upgrade.

This adds `router.epp.deploymentStrategy`, which overrides the Deployment's
`spec.strategy`. **The default is unchanged** (`Recreate`), so this is a no-op for
existing deployments. Operators whose tooling gates on availability can set a
RollingUpdate strategy whose `maxUnavailable` covers the standbys.

Per the issue's constraint, this does not let standbys receive traffic: they stay
`NotReady`, so they remain out of the Service endpoints.

**Verification** (OpenShift 4.21 / k8s v1.32, 2 EPP replicas):

Before, on the unmodified chart:

```console
$ helm install epp-ha . -n epp-ha-repro --set router.epp.replicas=2 --wait --timeout 4m
Error: INSTALLATION FAILED: resource Deployment/epp-ha-repro/epp-ha-epp not ready.
status: InProgress, message: Available: 1/2
context deadline exceeded          # after the full 4m

$ oc get deploy epp-ha-epp -o jsonpath='...'
ready=1 available=1 unavailable=1
Available=False MinimumReplicasUnavailable
```

After, with `deploymentStrategy: {type: RollingUpdate, rollingUpdate: {maxUnavailable: 1, maxSurge: 0}}`:

```console
$ helm install epp-fix . -n epp-verify -f fix-values.yaml --wait --timeout 4m
STATUS: deployed                   # 12 seconds

$ oc get deploy epp-fix-epp -o jsonpath='...'
ready=1/2 available=1
Available=True MinimumReplicasAvailable

$ oc get endpoints epp-fix-epp    # standby 10.130.2.244 is absent
NAME          ENDPOINTS
epp-fix-epp   10.128.11.117:9090,10.128.11.117:9002
```

Rendering is unchanged when the value is unset, and `helm lint` passes for both
the gateway and standalone charts.

**One caveat worth flagging for reviewers**: this remedy works on Helm 3 and Flux,
which compute readiness as `readyReplicas >= replicas - maxUnavailable`. It does
**not** help Helm 4, which uses `kstatus`; `kstatus` compares `availableReplicas`
against `spec.replicas` directly and ignores `maxUnavailable`. I verified both:
Helm 3.21.4 succeeds, Helm 4.1.1 still reports `Available: 1/2`. Helm 4 users need
the second documented option (do not gate on Deployment availability; gate on the
Service endpoint instead).

Fully closing the gap on Helm 4 would mean making standbys `Ready` while keeping
them out of the Service -- e.g. a leader-only label on the pod with the Service
selecting on it. That is an EPP-side change with RBAC implications, so I left it
out of this PR; happy to follow up if maintainers want that direction.

Fixes #2568

**Release note**:
```release-note
Added `router.epp.deploymentStrategy` to override the EPP Deployment's update strategy. The default remains `Recreate`. Multi-replica EPP deployments that must satisfy `helm --wait` or Flux can now set a RollingUpdate strategy whose `maxUnavailable` covers the intentionally-NotReady standby replicas.
```

### DIFF
--- a/config/charts/routerlib/values.yaml
+++ b/config/charts/routerlib/values.yaml
@@ -5,6 +5,22 @@ clusterDomain: cluster.local
 
 epp:
   replicas: 1
+
+  # Overrides the EPP Deployment's spec.strategy. Unset means `type: Recreate`,
+  # which is the recommended default for the stateful single-active-replica pattern.
+  #
+  # With replicas > 1 the chart enables leader election, and standby replicas stay
+  # NotReady by design, so a `Recreate` Deployment never reports Available=True and
+  # `helm --wait` blocks until it times out. Setting a RollingUpdate strategy whose
+  # maxUnavailable covers the standbys lets Helm consider the release ready while
+  # keeping standbys out of the Service (they are still NotReady, so they are not
+  # Endpoints). See "Multi-replica EPP and Helm --wait" in the chart README.
+  #
+  # deploymentStrategy:
+  #   type: RollingUpdate
+  #   rollingUpdate:
+  #     maxUnavailable: 1
+  #     maxSurge: 0
   image:
     registry: ghcr.io/llm-d
     repository: llm-d-router-endpoint-picker
--- a/config/charts/README.md
+++ b/config/charts/README.md
@@ -112,12 +112,75 @@ Core settings for the Endpoint Picker Proxy (EPP) container and pod, including s
 > *   **Active-Active Mode**: You can explicitly disable leader-election by passing `ha-enable-leader-election: false` under `router.epp.flags`. In this mode, all replicas process traffic concurrently.
 >     *   *Warning*: In active-active mode, you **must only use active-active compatible plugins**—specifically plugins that pull real-time metrics/state dynamically from the backend model servers (such as the precise prefix cache, queue, and KV-cache utilization scorers). Avoid plugins that rely on local in-memory routing state, as this state is not synchronized across replicas.
 
+##### Multi-replica EPP and Helm `--wait`
+
+In active-passive mode only the elected leader passes its readiness probe; standby
+replicas answer the `readiness` gRPC service with `NOT_SERVING` on purpose, so they
+never become Service endpoints. That is deliberate, but it means the EPP Deployment
+settles at `readyReplicas: 1` out of `replicas: N`.
+
+With the default `Recreate` strategy, Kubernetes forces `maxUnavailable` to `0`, so
+the Deployment reports `Available=False` / `MinimumReplicasUnavailable` and never
+becomes available. Anything that gates on Deployment availability blocks:
+
+```console
+$ helm install my-release . --set router.epp.replicas=2 --wait --timeout 4m
+Error: INSTALLATION FAILED: resource Deployment/.../my-release-epp not ready.
+status: InProgress, message: Available: 1/2
+```
+
+With Flux, the `HelmRelease` stays `pending-upgrade` until its timeout and then
+rolls back, and while it is pending no further EPP configuration change can start
+a new upgrade.
+
+Pick whichever of these fits your tooling:
+
+**1. Let Helm treat the standbys as expected-unavailable** (Helm 3 and Flux only)
+
+```yaml
+router:
+  epp:
+    replicas: 2
+    deploymentStrategy:
+      type: RollingUpdate
+      rollingUpdate:
+        maxUnavailable: 1   # must cover the standby replicas, i.e. >= replicas - 1
+        maxSurge: 0
+```
+
+Helm 3 computes readiness as `readyReplicas >= replicas - maxUnavailable`, so the
+release converges as soon as the leader is ready. Standbys remain `NotReady` and
+stay out of the Service, so no traffic reaches them before they are elected.
+
+> [!IMPORTANT]
+> This does **not** help Helm 4, which uses `kstatus`. `kstatus` compares
+> `availableReplicas` against `spec.replicas` directly and ignores `maxUnavailable`,
+> so a multi-replica EPP still reports `Available: 1/2` there. Use option 2 on Helm 4.
+>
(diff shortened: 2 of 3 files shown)


config/charts/routerlib/values.yaml
```diff
@@ -5,6 +5,22 @@ clusterDomain: cluster.local
 
 epp:
   replicas: 1
+
+  # Overrides the EPP Deployment's spec.strategy. Unset means `type: Recreate`,
+  # which is the recommended default for the stateful single-active-replica pattern.
+  #
+  # With replicas > 1 the chart enables leader election, and standby replicas stay
```
ahg-g, line 12: We should change this behavior and make the selection explicit:
- active-active
- active-passive
  -  leader election
  - priority with fallback

ahg-g (approved):

PR #2608, opened 2026-08-28
test: move generic Pod builder into test/framework/k8s

chethanuk:
**What type of PR is this?**

/kind cleanup
/kind testing

Next in the #1188 split, after #2121 and #2122. It moves the fluent Pod builders out of
`pkg/epp/util/testing/wrappers.go` into `test/framework/k8s`.

The builders are plain Kubernetes object helpers with nothing EPP-specific about them, and
they already have callers outside `pkg/epp`. Once they leave, `wrappers.go` holds only the GAIE
wrappers it is named for; those move next.

The move is 1:1. `DefaultTestPort`, `PodWrapper`, `FromBase`, `MakePod` and the seven
`PodWrapper` methods change only their package clause. `wrappers.go` drops its `corev1`
import because what remains never used it.

The rest of the diff is not part of the move. `test/framework/k8s/builders_test.go` is new,
because the builders had no direct tests and `test/framework/k8s` would otherwise land
uncovered. The R1 line in `doc.go` and the `importas` block in `.golangci.yml` drop their
`test/framework/context` entries: that package is out of scope for the issue, so both named
a package that will not exist.

**Test plan**:

- [x] `go vet ./...` clean across the repo, plus `-tags=integration_tests` on
      `./test/integration/...`
- [x] `go test ./test/framework/... ./pkg/epp/controller/... ./pkg/epp/util/...` passes
- [x] `golangci-lint` v2.10.0, the version `Dockerfile.builder` pins, reports `0 issues.` over
      `./test/framework/...` and the touched `pkg/epp` packages, matching the same run on `main`
- [x] Pod builder behavior at the new path: the defaults `Complete()` applies, the ready
      condition, the deletion timestamp, and the label, IP and namespace setters

Part of #1188

```release-note
NONE
```


chethanuk:

chethanuk:
@elevran or @ahg-g Please review and merge the PR [part 1 of #1188] - Thank you

PR #2609, opened 2026-08-28
test: move GAIE CRD builders into test/framework/gaie

chethanuk:
**What type of PR is this?**

/kind cleanup
/kind testing

Continues the #1188 split, after #2121, #2122 and the k8s builder move. It is the
last piece of the `pkg/epp/util/testing` removal.

This branch is based on #2608, so the diff currently includes that PR's commit as
well. Once #2608 merges this becomes a single commit touching only the
`test/framework/gaie` files and their callers.

`wrappers.go` holds the GAIE CRD wrapper builders, `MakeInferenceObjective` and
`MakeInferencePool` with their fluent setters. They construct GAIE and apix API
objects and read the API group from `pkg/common/routing`; nothing in them touches
`pkg/epp`. `test/framework/gaie` is the tier `test/framework/doc.go` describes for
that import set. The Pod builder having gone with the previous move, taking these
empties the file, so `wrappers.go`, the `pkg/epp/util/testing` package and the
`testutil` alias in `.golangci.yml` all go.

The move is verbatim. The builder bodies are unchanged; the deltas are the package
clause and a package comment. Callers switch to the `fwkgaie` alias that
`.golangci.yml` already declares. One of them is `test/utils/server.go`, which is
outside the scope of #1188 but calls `MakeInferencePool`, so its import moves with
the symbols and nothing else in the file changes.

`builders_test.go` covers what the builders produce: the default target port and
TypeMeta, each setter including the repeated and zero-value cases, and that `ObjRef`
aliases the wrapped object.

**Test plan**:

- [x] `gofmt -l .` reports nothing
- [x] `go vet ./...` and `go vet -tags integration_tests ./test/...` clean, so the
      delete left no dangling reference and no import cycle
- [x] `go test ./test/framework/...` passes, including the new builder tests
- [x] `go test ./pkg/epp/controller/... ./pkg/epp/datastore/... ./pkg/epp/server/...
      ./pkg/epp/requestcontrol/...` passes, covering every caller that switched alias

Part of #1188

```release-note
NONE
```


chethanuk:

chethanuk:
@elevran or @ahg-g Please review and merge the PR [part 2 of #1188] - Thank you

PR #2610, opened 2026-08-28
test: move ext_proc builders into test/framework/epp

chethanuk:
**What type of PR is this?**

/kind cleanup
/kind testing

Part of the #1188 split. It follows the k8s and GAIE builder moves.

This branch is based on #2609, so the diff currently includes that PR's commit and
#2608's as well. Once both merge this becomes a single commit touching only the
ext_proc helper files and their callers.

`test/integration/util.go` is a 686-line grab bag of ext_proc request builders and live
stream executors, and the suite under `test/integration/epp` imports the directory above
it as a package. That is the cyclic-dependency risk #1188 names: helpers living inside
the tree that consumes them. The file moves to `test/framework/epp`, split into
`builders_extproc.go` for the request and response builders and `client_extproc.go` for
the stream executors and the readiness wait. What remains in `test/integration` sits
behind the `integration_tests` build tag, so the root package is no longer importable.
Callers switch to the `fwkepp` alias that `.golangci.yml` already declares.

Five symbols are dropped rather than moved, which the maintainer ratified.
`ReqLLMWithStream`, `ReqLLMUnary` and `StartExtProcServer` have no callers, and
`GenerateRequestWithStream` is reachable only from `ReqLLMWithStream`, so carrying them
over would put zero-caller exports in the new package. `SendRequest` had one caller, in
the test that moves with it, and that call site uses the `client.Send` and `client.Recv`
pair `SendRequest` wrapped.

Every function that does move has a byte-identical body; the deltas are the package
clause, a package comment and the split across two files. `util_test.go` moves with the
code it covers, which is why the root package goes away entirely rather than being left
importable but empty. Its three tests carry over with their assertions unchanged. The
stub ext_proc server they shared becomes an echo server, which also backs the new
builder and stream client coverage in `builders_extproc_test.go` and
`client_extproc_test.go`.

`extPorcConnSetupPollInterval` is misspelled and moves that way so the diff reads as a
move. Worth a separate cleanup.

**Test plan**:

- [x] `go vet ./...` and `go vet -tags integration_tests ./test/...` clean at this commit
- [x] `gofmt -l .` reports nothing
- [x] `go test ./test/framework/...` passes, including the three moved tests and the new
      builder and stream client coverage against an in-process echo server
- [x] `go test ./pkg/epp/...` passes, unchanged by the move

Part of #1188

```release-note
NONE
```


chethanuk:

chethanuk:
@elevran or @ahg-g Please review and merge the PR [part 3 of #1188] - Thank you

PR #2611, opened 2026-08-28
test: export EPP harness API and root discovery

chethanuk:
**What type of PR is this?**

/kind cleanup
/kind testing

Part of the #1188 split, following the ext_proc builders move. Nothing moves here.
The hermetic EPP harness gets the surface it will be moved with, and it stops
locating the repository by counting parent directories.

This branch is based on #2610, so the diff currently includes that PR's commit and
the two before it. Once those merge this becomes a single commit touching only
`test/integration/epp`.

`test/integration/epp/harness.go` gains `Run(m *testing.M) int`, which absorbs the
`TestMain` body from `hermetic_test.go` in its existing order, along with
`repoRootPath` and `loadBaseResources` so the harness travels complete. `TestMain`
becomes a one-line delegation. `K8sClient`, `Config`, `Logger` and `Scheme` wrap the
same package vars; `GRPCConn` and `SetPodMetrics` expose the harness fields the tests
already reach into; `runMode`, `standaloneStrategy`, `testPoolName`, `cleanMetric`,
`metricReqTotal` and `metricReadyPods` become their exported forms. Callers in the
package follow. `withGPUPods` in
`attributeweight_scenario_test.go` takes a `*testing.T` and reads the harness through
those accessors, so it still compiles from outside the package once the harness moves.

Root discovery does change behavior. `TestMain` took `runtime.Caller(0)` and joined
three `..` onto the source file's directory for the repository root, then shelled
`go list -m` separately for the gateway-api-inference-extension CRD directory. A
`moduleDir` helper does both, resolving each module path through
`go list -m -f {{.Dir}}`. The old walk was source-tree relative and the replacement is
module relative; they agree with this repository as the main module and would differ
under a `go.work` replace.

That count of `..` is why the two halves are separated. The harness lands in
`test/framework/epp/harness`, at a different depth, where a stale `../../..` still
compiles and resolves to a wrong directory that envtest reports as a missing CRD path.
Discovery is fixed ahead of the move so the move itself stays mechanical, under the
rule that no PR mixes a move with a behavior change.

**Test plan**:

- [x] `go vet ./...`, `go vet -tags integration_tests ./test/...` and `gofmt -l .`
      clean at this commit
- [x] `go test ./test/framework/...` and `go test ./pkg/epp/server/...` pass
- [x] `go list -m -f {{.Dir}}` resolves both the llm-d module root and
      `sigs.k8s.io/gateway-api-inference-extension`, and the llm-d root holds
      `config/crd/bases`
- [x] `make test-integration-hermetic` passes, covering `Run(m)` as the harness
      entry point and module-relative root discovery

Part of #1188

```release-note
NONE
```


chethanuk:

chethanuk:
@elevran or @ahg-g Please review and merge the PR [part 4 of #1188] - Thank you

PR #2612, opened 2026-08-28
test: extract EPP envtest harness into test/framework/epp/harness

### LINKED ISSUE #1188: [Testing] Consolidate Hermetic Integration Test Helpers into Structured test/framework

chethanuk:
**What type of PR is this?**

/kind cleanup
/kind testing

Last of the #1188 moves, and it closes the issue. It follows the harness API PR,
which exported the harness surface and replaced the parent-directory count in root
discovery with a module lookup. That work is why this one is a mechanical move: the
names the integration suites call are already the names the harness exports, and the
embed and CRD paths no longer depend on where the file sits.

This branch is based on #2611, so the diff currently includes that PR's commit and
the three before it. Once those merge this becomes a single commit moving the
harness and repointing its callers.

`test/integration/epp/harness.go` becomes `test/framework/epp/harness/harness.go`,
with `testdata/datalayer-config.yaml` alongside it so the `go:embed` still resolves.
Fifteen files under `test/integration/epp` repoint at it through the `eppharness`
alias that `.golangci.yml` gains here.

The pod-state and metric helpers leave `common_tests.go` with it: `PodState`, `P`,
`label`, `labelsToString`, `MetricReqTotal`, `MetricReadyPods` and `CleanMetric`.
They describe harness inputs and the metric text the harness serves, and `WithPods`
and `WaitForReadyPodsMetric` call them, so leaving them in the integration package
would mean the harness importing its own callers.

One change is not part of the move. `framework_TEST_PACKAGES` filters the harness
package out of `./test/framework/...`, in the same `go list | grep -v` form that
`epp_TEST_PACKAGES` already uses. `make test-unit-framework` starts no API server, so
it would report the harness at 0% while `make test-integration-hermetic` is what
exercises it.

**Test plan**:

- [x] `gofmt -l .` empty, `go vet ./...` and `go vet -tags integration_tests ./test/...` clean
- [x] `go test ./test/framework/... ./pkg/epp/...` passes
- [x] `go list ./test/framework/... | grep -vE '/test/framework/epp/harness($|/)'` returns the five other framework packages and drops the harness

The hermetic suite needs envtest and runs in CI under `make test-integration-hermetic`;
that is what covers the moved embed path.

Fixes #1188

```release-note
NONE
```


chethanuk:
@elevran or @ahg-g Please review and merge the PR [part 5 of #1188] - Thank you

PR #2616, opened 2026-08-29
feat(coordinator): add pipeline amplification and media metrics

### LINKED ISSUE #2482: Add native Coordinator pipeline metrics

Extend the native Coordinator Prometheus instrumentation introduced by #2435 with metrics that are not currently covered there.

Part of #2276.

The scope includes:

* `llm_d_coordinator_encode_fanout_size` (Histogram): number of Encode subrequests produced by one client request, including zero for requests that do not require Encode.
* `llm_d_coordinator_orchestration_overhead_seconds{route}` (Histogram): Coordinator wall-clock time spent outside request parsing and pipeline step execution.
* `llm_d_coordinator_media_items{media_type}` (Histogram): number of media items of each type found in one client request.
* `llm_d_coordinator_media_download_duration_seconds{result}` (Histogram): duration of each media download attempt.
* `llm_d_coordinator_response_bytes{stream}` (Histogram): total number of bytes written to the client for streaming and non-streaming responses.

The metrics should reuse the registry and `/metrics` endpoint provided by #2435 rather than introduce a separate metrics server.

All labels must use bounded enumerations:

* `route` must identify a normalized Coordinator route, not a raw URL or request path.
* `media_type` must use the fixed set of media types supported by the Coordinator.
* `result` should use bounded values such as `success`, `error`, and `cancelled`.
* `stream` should distinguish streaming and non-streaming responses using a fixed boolean-like value.

Request IDs, raw URLs, error strings, prompt content, and other unbounded values must not be used as metric labels.

Acceptance criteria:

* The new metrics are registered with the Coordinator metrics infrastructure from #2435 and exposed through its existing `/metrics` endpoint.
* Encode fan-out is observed once per client request after the fan-out size is known.
* Media item counts and download durations are recorded with bounded labels.
* Response bytes are counted correctly for both streaming and non-streaming responses, including partial writes when a request is cancelled or the client disconnects.
* Orchestration overhead has a clearly documented observation boundary and does not double-count parsing or pipeline step execution time.
* Focused tests verify metric observations, error and cancellation paths, streaming and non-streaming byte accounting, and bounded label values.
* Existing request handling and streaming behavior remain unchanged.
* Metric help text documents the observation boundary and units.

**Why is this needed**:

A Coordinator request can expand into multiple EPP requests. Multimodal traffic may additionally fan out into multiple Encode subrequests. EPP request-level metrics therefore describe traffic after Coordinator orchestration and do not fully represent client-facing request behavior.

#2435 provides the base Coordinator metrics infrastructure and covers request volume, pipeline steps, execution paths, and input tokens. However, it does not expose Encode fan-out size, media processing volume and latency, outbound response size, or Coordinator orchestration overhead.

These additional metrics make request amplification, multimodal processing cost, response volume, and Coordinator-specific overhead directly observable without duplicating the instrumentation already provided by #2435.

Here is a detailed breakdown of the functional value and production use cases for the 5 proposed metrics. These metrics are designed to provide essential observability for multimodal traffic and disaggregated routing behavior:

### 1. `llm_d_coordinator_encode_fanout_size`

* **Definition:** The number of Encode subrequests produced by a single client request (0 for text-only requests).
* **Use Cases (Traffic Amplification & Capacity Planning):**
* In a disaggregated architecture, this metric reveals the actual concurrency pressure on the Vision Encoder cluster, independent of client QPS. It helps in making informed scaling decisions for vision nodes.
* It provides the necessary data to implement complexity-based rate limiting to protect the system from extreme multimodal payloads (e.g., a single request with 50+ images).

### 2. `llm_d_coordinator_orchestration_overhead_seconds{route}`

* **Definition:** The pure orchestration overhead of the Coordinator (calculated as Total Duration - Request Parsing - Upstream Execution).
* **Use Cases (Fault Isolation & Performance Benchmarking):**
* **Latency Attribution:** When P99 latency spikes, this metric helps determine if the bottleneck lies within the Router itself (e.g., lock contention, Go scheduler overhead) or if it's due to upstream vLLM saturation/network I/O.
* It establishes a baseline for measuring the performance impact of future architectural optimizations within the Coordinator.

### 3. `llm_d_coordinator_media_items{media_type}`

* **Definition:** The number of media items included in a single request, categorized by bounded types (e.g., `image`, `audio`).
* **Use Cases (Context Estimation & Capacity-Aware Routing):**
* **KV Cache Routing:** High-resolution images consume significant KV cache during the Prefill phase. This metric allows the Router to estimate the memory footprint beforehand and intelligently route heavy multimodal requests to nodes with sufficient VRAM, preventing OOM errors.
* Provides traffic profiling to understand the distribution of multimodal vs. text-only interactions.

### 4. `llm_d_coordinator_media_download_duration_seconds{result}`

* **Definition:** The duration of each attempt to download external media URLs, labeled by `result` (`success`, `error`, `cancelled`).
* **Use Cases (External I/O Isolation):**
* **Decoupling Network Latency:** Multimodal requests often suffer from slow external image hosts. This metric isolates public network download times from actual AI inference latency, preventing misattribution of performance issues.
* Helps operators tune HTTP client timeout thresholds and retry policies based on the real-world long-tail distribution of media downloads.

### 5. `llm_d_coordinator_response_bytes{stream}`

* **Definition:** The actual bytes written to the client network interface, including partial writes. The `stream` label distinguishes between SSE streaming and non-streaming responses.
* **Use Cases (Accurate Accounting & Bandwidth Monitoring):**
* **Precise Accounting:** In streaming scenarios, clients frequently disconnect mid-generation. Tracking actual bytes written (including partial writes) ensures accurate resource accounting and billing, unlike relying solely on generated tokens.
* **Outbound Bandwidth:** Provides visibility into the actual network throughput consumed by long-lived streaming connections, aiding in network capacity planning.

zhouyou9505:
**What type of PR is this?**
/kind feature

Extends the native Coordinator Prometheus instrumentation from #2435 with the pipeline amplification and media metrics scoped in #2482. Reuses the same `llm_d_coordinator` registry, `/metrics` endpoint, record helpers, and cardinality-bounded labels. Does not duplicate series already covered by #2435 (`request_*`, `step_*`, `upstream_*`, `execution_path_total`, etc.).

Adds five histograms:
- `llm_d_coordinator_encode_fanout_size` — Encode subrequests per client request (including 0)
- `llm_d_coordinator_orchestration_overhead_seconds{route}` — wall time outside parse + pipeline steps
- `llm_d_coordinator_media_items{media_type}` — media count per bounded type
- `llm_d_coordinator_media_download_duration_seconds{result}` — per HTTP download attempt
- `llm_d_coordinator_response_bytes{stream}` — bytes written to the client, including partial writes

Fixes #2482
Part of #2276

**Release note**:
```release-note
Coordinator Prometheus metrics now include encode fan-out size, orchestration overhead, media item counts, media download duration, and response bytes written to the client.
```

## Test plan
- [x] Unit tests for the five new histograms and observation points (metrics, pipeline, handlers, encode, replace-media-urls, render)
- [x] Local smoke against mock gateway/media: scrape `/metrics` and confirm all five series have samples under mixed text/image traffic

### DIFF
--- a/pkg/coordinator/metrics/llm_d_coordinator_metrics.go
+++ b/pkg/coordinator/metrics/llm_d_coordinator_metrics.go
@@ -175,11 +175,11 @@ var (
 // media_download_duration_seconds, which is one observation per download
 // attempt.
 var (
-	encodeFanoutSize = prometheus.NewHistogramVec(
+	encodeSubrequests = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Subsystem: LLMDRouterCoordinatorSubsystem,
-			Name:      "encode_fanout_size",
-			Help:      metricsutil.HelpMsgWithStability("Number of Encode subrequests produced by one client request. Observed once per pipeline execution after fan-out size is known; 0 when Encode does not run or is skipped. Unit: subrequests.", compbasemetrics.ALPHA),
+			Name:      "encode_subrequests",
+			Help:      metricsutil.HelpMsgWithStability("Number of Encode subrequests produced by one client request. Observed once per pipeline execution after fan-out is known; 0 when Encode does not run or is skipped. Unit: subrequests.", compbasemetrics.ALPHA),
 			Buckets:   CountBuckets,
 		},
 		[]string{},
@@ -215,10 +215,10 @@ var (
 		resultLabel,
 	)
 
-	responseBytes = prometheus.NewHistogramVec(
+	responseSize = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Subsystem: LLMDRouterCoordinatorSubsystem,
-			Name:      "response_bytes",
+			Name:      "response_size_bytes",
 			Help:      metricsutil.HelpMsgWithStability("Total bytes written to the client for one request, including partial writes on cancellation or disconnect. Observed once per client request. Unit: bytes.", compbasemetrics.ALPHA),
 			Buckets:   metricsutil.RequestSizeBuckets,
 		},
--- a/pkg/coordinator/metrics/llm_d_coordinator_metrics.go
+++ b/pkg/coordinator/metrics/llm_d_coordinator_metrics.go
@@ -169,3 +169,59 @@ var (
 		[]string{"result"},
 	)
 )
+
+// Per-request pipeline amplification, orchestration cost, media processing,
+// and outbound response size. Recorded once per client request except
+// media_download_duration_seconds, which is one observation per download
+// attempt.
+var (
+	encodeFanoutSize = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: LLMDRouterCoordinatorSubsystem,
+			Name:      "encode_fanout_size",
+			Help:      metricsutil.HelpMsgWithStability("Number of Encode subrequests produced by one client request. Observed once per pipeline execution after fan-out size is known; 0 when Encode does not run or is skipped. Unit: subrequests.", compbasemetrics.ALPHA),
+			Buckets:   CountBuckets,
+		},
+		[]string{},
+	)
+
+	orchestrationOverhead = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: LLMDRouterCoordinatorSubsystem,
+			Name:      "orchestration_overhead_seconds",
+			Help:      metricsutil.HelpMsgWithStability("Coordinator wall time outside request parsing and pipeline step execution. Observed once per client request as max(0, total-parse-sum(steps)). Unit: seconds.", compbasemetrics.ALPHA),
+			Buckets:   metricsutil.GeneralLatencyBuckets,
+		},
+		routeLabel,
+	)
+
+	mediaItems = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: LLMDRouterCoordinatorSubsystem,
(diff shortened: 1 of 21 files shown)


pkg/coordinator/metrics/llm_d_coordinator_metrics.go
```diff
@@ -169,3 +169,59 @@ var (
...
+
+// Per-request pipeline amplification, orchestration cost, media processing,
+// and outbound response size. Recorded once per client request except
+// media_download_duration_seconds, which is one observation per download
+// attempt.
+var (
+	encodeFanoutSize = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Subsystem: LLMDRouterCoordinatorSubsystem,
+			Name:      "encode_fanout_size",
```
revit13: On encode_fanout_size: "size" reads as bytes, but this counts sub-requests. Prometheus convention ends a count metric in the plural noun, as llm_d_coordinator_request_input_tokens already does here. docs/metrics.coord.md  (from https://github.com/llm-d/llm-d-router/pull/2277) also proposed a name for this metric, under "Per-request image visibility":

request_images: Histogram of multimodal entries per request.

Could we use request_images, or encode_subrequests?

Same idea for response_bytes: response_size_bytes would match the existing request_size_bytes.

zhouyou9505: thanks for suggestion. encode_subrequests , response_size_bytes would be better.

Issue #1091, opened 2026-05-13
[Disaggregation]:  Define a standard approach for plugins to identify prefill and decode endpoints


ahg-g:
xref: https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2080

What would you like to be added:

Define a standard approach for plugins to identify prefill and decode endpoints

Why is this needed:

Some plugins need to be p/d aware, one example is https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2021

One suggestion is to have the SchedulingResult define optional PrefillProfileName and DecodeProfileName in addition to the PrimaryProfileName we currently have. The argument for this is that p/d is a core feature of LLM serving, and so having such explicit definitions is reasonable especially when defined as optional fields.

Please suggest other options to address this requirement.

PR #2627, opened 2026-08-31
fix(epp): record encode-stage endpoint in multimodal encoder-cache

### LINKED ISSUE #2626: mm-embeddings-cache-producer records decode pod instead of encode pod in E/PD disaggregation
**What happened**

In encode-disaggregated (E/PD, E/P/D) serving, the `mm-embeddings-cache-scorer` always scores `0` for every encode pod, so encoder-cache affinity never routes repeated media to the pod that already holds the embedding. `encoder_cache_hit_ratio` stays at 0 even when the same image is sent repeatedly.

**Root cause**

The `mm-embeddings-cache-producer` splits work across two hooks:
- `Produce()` (scheduling): reads the per-pod LRU keyed by `endpoint.ID` and stashes `EncoderCacheMatchInfo` on each candidate; the scorer reads that for encode pods.
- `PreRequest()` (post-scheduling): writes the LRU for the pod that handled the request.

`PreRequest` records the endpoints from `schedulingResult.ProfileResults[PrimaryProfileName]`. `disagg-profile-handler` sets `PrimaryProfileName` to the **decode** profile, so the producer records the **decode** pod ID. But encoding actually happens on the **encode** profile pod. The writer stores decode-pod IDs while the reader queries encode-pod IDs, so the keys never match and the affinity score is always 0.

A secondary effect: recording the decode pod also inflates `encoder_cache_hits_total`, which counts a hit for every pod in the LRU holding the hash, even though decode pods do not hold the encoder cache.

**What you expected to happen**

The producer should record the encode-stage endpoint(s), so repeated media sticks to the pod that already encoded it and the hit metrics reflect only encode pods.

**How to reproduce it (as minimally and precisely as possible)**

Deploy the E/PD guide (`guides/multimodal-serving/e-disaggregation`) with two or more encode pods, send the same image repeatedly, and observe that `mm-embeddings-cache-scorer` scores stay 0 for all encode pods (EPP `-v=4` logs) and `encoder_cache_hit_ratio` stays 0.

**Environment**

- EPP v0.10.0
- E/PD disaggregation, `mm-embeddings-cache-producer` + `mm-embeddings-cache-scorer` in the encode profile

acardace:
**What type of PR is this?**
/kind bug

In encode-disaggregated (E/PD, E/P/D) serving, `mm-embeddings-cache-scorer` always scored 0 for every encode pod, so encoder-cache affinity never routed repeated media to the pod that already held the embedding, and `encoder_cache_hit_ratio` stayed at 0.

The `mm-embeddings-cache-producer` reads the per-pod LRU in `Produce()` (keyed by `endpoint.ID`, scoring encode pods) and writes it in `PreRequest()`. `PreRequest` recorded the endpoints from `ProfileResults[PrimaryProfileName]`, which `disagg-profile-handler` sets to the decode profile. So the producer stored decode-pod IDs while scoring queried encode-pod IDs: the keys never matched and affinity was always 0. Recording the decode pod also inflated `encoder_cache_hits_total`, which counts a hit for every pod in the LRU holding the hash even though decode pods do not hold the encoder cache.

This PR records the encode profile target endpoints instead, falling back to the primary profile when no encode profile ran (aggregated serving). The encode profile name is configurable via a new `encodeProfile` parameter (default `encode`) to match `disagg-profile-handler`.

Validated on a 1 encode-per-pod and 2 encode-pod E/PD deployment (Qwen2.5-VL-7B): with the fix, the first request scores 0 (cold), subsequent same-image requests score 1 on the encoding pod and 0 on the other, and all repeated requests route to that pod. `encoder_cache_hits_total` records only the encode pod.

Fixes #2626

```release-note
Fixed multimodal encoder-cache affinity in encode-disaggregated (E/PD, E/P/D) serving. The mm-embeddings-cache-producer now records the encode-stage endpoint that holds the embeddings, rather than the decode endpoint, so repeated media routes to the pod that already encoded it. Adds an optional encodeProfile parameter (default "encode") to match the profiles.encode value on disagg-profile-handler.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/prerequest.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/prerequest.go
@@ -40,7 +40,7 @@ func (p *Producer) PreRequest(ctx context.Context, request *scheduling.Inference
 		return nil
 	}
 
-	targets := targetEndpoints(schedulingResult)
+	targets := p.targetEndpoints(schedulingResult)
 	if len(targets) == 0 {
 		logger.Info("No target endpoints found, skipping encoder-cache update")
 		return nil
@@ -65,13 +65,23 @@ func (p *Producer) PreRequest(ctx context.Context, request *scheduling.Inference
 	return nil
 }
 
-func targetEndpoints(schedulingResult *scheduling.SchedulingResult) []scheduling.Endpoint {
-	if schedulingResult == nil || schedulingResult.PrimaryProfileName == "" || schedulingResult.ProfileResults == nil {
+// targetEndpoints returns the endpoints that computed and hold the encoder cache
+// for the request. In disaggregated serving the encode profile selects those
+// pods; its target is recorded rather than the primary (decode) profile's, whose
+// pod does not hold the encoder cache. In aggregated serving no encode profile
+// runs and the primary profile's pod both encodes and serves.
+func (p *Producer) targetEndpoints(schedulingResult *scheduling.SchedulingResult) []scheduling.Endpoint {
+	if schedulingResult == nil || schedulingResult.ProfileResults == nil {
 		return nil
 	}
-	result := schedulingResult.ProfileResults[schedulingResult.PrimaryProfileName]
-	if result == nil {
+	if result := schedulingResult.ProfileResults[p.encodeProfile]; result != nil {
+		return result.TargetEndpoints
+	}
+	if schedulingResult.PrimaryProfileName == "" {
 		return nil
 	}
-	return result.TargetEndpoints
+	if result := schedulingResult.ProfileResults[schedulingResult.PrimaryProfileName]; result != nil {
+		return result.TargetEndpoints
+	}
+	return nil
 }
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
@@ -51,6 +51,11 @@ const (
 
 	// bytesPerImage is the assumed memory per tracked image.
 	bytesPerImage = 2 * 1024 * 1024
+
+	// defaultEncodeProfile is the scheduling profile that selects encode-stage
+	// endpoints. It matches the disagg-profile-handler default and is
+	// overridable via the encodeProfile parameter.
+	defaultEncodeProfile = "encode"
 )
 
 var (
@@ -67,6 +72,11 @@ var (
 type Parameters struct {
 	// CacheSizeInMBPerServer is the per-endpoint LRU memory budget in mebibytes (MiB).
 	CacheSizeInMBPerServer int `json:"cacheSizeInMBPerServer"`
+
+	// EncodeProfile names the scheduling profile that selects encode-stage
+	// endpoints in disaggregated serving. Must match the profiles.encode value
+	// configured on disagg-profile-handler. Empty selects the default "encode".
+	EncodeProfile string `json:"encodeProfile,omitempty"`
 }
 
 // lruCapacityFromCacheSizeMB converts a MiB budget to a maximum LRU entry count.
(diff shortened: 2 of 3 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/prerequest.go
```diff
@@ -65,13 +65,23 @@ func (p *Producer) PreRequest(ctx context.Context, request *scheduling.Inference
...
+// runs and the primary profile's pod both encodes and serves.
+func (p *Producer) targetEndpoints(schedulingResult *scheduling.SchedulingResult) []scheduling.Endpoint {
+	if schedulingResult == nil || schedulingResult.ProfileResults == nil {
 		return nil
 	}
-	result := schedulingResult.ProfileResults[schedulingResult.PrimaryProfileName]
-	if result == nil {
+	if result := schedulingResult.ProfileResults[p.encodeProfile]; result != nil {
+		return result.TargetEndpoints
+	}
```
zetxqx, line 79: do we want to add a V(logging.DEBUG) after this to warn that the encoderProfile results might be warning?

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
```diff
@@ -67,6 +72,11 @@ var (
 type Parameters struct {
 	// CacheSizeInMBPerServer is the per-endpoint LRU memory budget in mebibytes (MiB).
 	CacheSizeInMBPerServer int `json:"cacheSizeInMBPerServer"`
+
+	// EncodeProfile names the scheduling profile that selects encode-stage
+	// endpoints in disaggregated serving. Must match the profiles.encode value
+	// configured on disagg-profile-handler. Empty selects the default "encode".
+	EncodeProfile string `json:"encodeProfile,omitempty"`
```
zetxqx, line 79: How about dropping the encodeProfile parameter for now and hardcoding the default like approx-prefix-cache-producer does with experimentalDefaultPrefillProfile. https://github.com/llm-d/llm-d-router/blob/faaeff2672cada6c54d40e6a0cd5b414d155a9e8/pkg/epp/framework/plugins/requestcontrol/dataproducer/approximateprefix/types.go#L94 ?

I just feel adding one more parameter could be harder to drop. And based on the configuration, I feel in general we should pass down the profile name while parsing the config to initialize the plugins.

```
  schedulingProfiles:
  - name: encode
    plugins:
    - pluginRef: encode-filter
    - pluginRef: mm-embeddings-cache-scorer
      weight: 4
    - pluginRef: queue-scorer              
      weight: 1
    - pluginRef: max-score-picker
  - name: decode
    plugins:
    - pluginRef: decode-filter
      weight: 2
    - pluginRef: queue-scorer
      weight: 1
    - pluginRef: max-score-picker
```

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/prerequest.go
```diff
@@ -65,13 +65,23 @@ func (p *Producer) PreRequest(ctx context.Context, request *scheduling.Inference
...
 }
 
-func targetEndpoints(schedulingResult *scheduling.SchedulingResult) []scheduling.Endpoint {
-	if schedulingResult == nil || schedulingResult.PrimaryProfileName == "" || schedulingResult.ProfileResults == nil {
+// targetEndpoints returns the endpoints that computed and hold the encoder cache
+// for the request. In disaggregated serving the encode profile selects those
+// pods; its target is recorded rather than the primary (decode) profile's, whose
+// pod does not hold the encoder cache. In aggregated serving no encode profile
+// runs and the primary profile's pod both encodes and serves.
+func (p *Producer) targetEndpoints(schedulingResult *scheduling.SchedulingResult) []scheduling.Endpoint {
```
capri-xiyue, line 73: Maybe rename it to targetEncodeEndpoints to avoid confusion

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/prerequest.go
```diff
@@ -65,13 +65,23 @@ func (p *Producer) PreRequest(ctx context.Context, request *scheduling.Inference
...
+	if schedulingResult == nil || schedulingResult.ProfileResults == nil {
```
capri-xiyue, line 74: Should we put as a top check? I think the logic here should be 

1. First, determine the profile to use by checking p.encodeProfile first, and falling back to schedulingResult.PrimaryProfileName if the first is unavailable. 
2. Next, check if the resolved profile name is empty or not; if it is valid (not empty), proceed to find the corresponding endpoints, otherwise return nil.

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer.go
capri-xiyue: +1. And could you link https://github.com/llm-d/llm-d-router/issues/1091 in the default encode profile so that we know the context

pkg/epp/framework/plugins/requestcontrol/dataproducer/multimodal/producer_test.go
```diff
@@ -146,6 +146,48 @@ func TestProduceMatchesMultiplePodsAndPreRequestUpdatesPlacement(t *testing.T) {
...
+		},
+	}
+
+	_ = producer.PreRequest(context.Background(), request, result)
+	producer.wg.Wait()
+
+	cache := producer.cacheSnapshot()
+	assert.Contains(t, cache["hash-a"], encodePod.String())
+	assert.NotContains(t, cache["hash-a"], decodePod.String())
+}
```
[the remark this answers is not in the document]
capri-xiyue: Please fix it

PR #2631, opened 2026-08-31
fix(scheduling): classify unavailable endpoints

### LINKED ISSUE #2628: P/D disaggregation returns 429 when no decode endpoints are available
**What happened**:
In a P/D disaggregated deployment, a request returns
429/rejected-saturated when prefill endpoints exist but no decode
endpoints are available.

This is misleading because the required decode stage is unavailable,
rather than saturated.

- No decode endpoints: 503/rejected-no-endpoints.
- Decode endpoints exist but are saturated: 429/rejected-saturated.

nicole-lihui:
**What type of PR is this?**

/kind bug

A drained filter chain may mean required endpoint roles are absent, not that eligible workers are saturated. Re-evaluate declared eligibility filters against the original candidates to preserve that distinction.

Fixes #2628

```release-note
P/D requests with no eligible decode endpoint now return 503 with `rejected-no-endpoints`. 
Label-based eligibility constraints use the same classification, while capacity-related filter drains continue to return 429 with `rejected-saturated`.
```

### DIFF
--- a/pkg/epp/scheduling/scheduler_profile.go
+++ b/pkg/epp/scheduling/scheduler_profile.go
@@ -127,14 +127,10 @@ func (p *SchedulerProfile) String() string {
 func (p *SchedulerProfile) Run(ctx context.Context, request *fwksched.InferenceRequest, candidateEndpoints []fwksched.Endpoint) (*fwksched.ProfileRunResult, error) {
 	endpoints := p.runFilterPlugins(ctx, request, candidateEndpoints)
 	if len(endpoints) == 0 {
-		// Filters draining a non-empty candidate set means the pool is busy, not
-		// broken: an empty pool is rejected in the director before scheduling
-		// runs. Report it with the same status and drop-reason vocabulary as a
-		// flow control capacity rejection.
 		return nil, errcommon.Error{
-			Code:    errcommon.ResourceExhausted,
+			Code:    errcommon.ServiceUnavailable,
 			Msg:     "no endpoints available for the given request",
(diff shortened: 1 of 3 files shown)


pkg/epp/scheduling/scheduler_profile.go
```diff
@@ -127,10 +127,16 @@ func (p *SchedulerProfile) String() string {
 func (p *SchedulerProfile) Run(ctx context.Context, request *fwksched.InferenceRequest, candidateEndpoints []fwksched.Endpoint) (*fwksched.ProfileRunResult, error) {
 	endpoints := p.runFilterPlugins(ctx, request, candidateEndpoints)
 	if len(endpoints) == 0 {
-		// Filters draining a non-empty candidate set means the pool is busy, not
-		// broken: an empty pool is rejected in the director before scheduling
-		// runs. Report it with the same status and drop-reason vocabulary as a
-		// flow control capacity rejection.
+		eligibleEndpoints, hasEligibilityFilters := p.runEndpointEligibilityFilters(ctx, request, candidateEndpoints)
```
ahg-g: Filters are not meant to implement filter out "Resource Exhausted" endpoints, this is done by the saturation detector. 

I think we should treat "zero endpoints" as service unavailable all the time. So we don't need a new interface, we should just need to update the error code we are returning below.

nicole-lihui: @ahg-g  Thanks, this helps clear up something I’d been unsure about. I was thinking ahead to `Decode fallback` #2636 and added some extra logic here, but I’d also been feeling that this PR was trying to do a bit too much.

PTAL

ahg-g (approved):

PR #2639, opened 2026-09-01
test(sidecar): add audio and video multimodal test coverage

### LINKED ISSUE #2640: Add sidecar and e2e test coverage for audio and video multimodal requests

Extend sidecar and e2e test coverage so that `video_url`, `audio_url`, and `input_audio` multimodal items are exercised through the encode-cache paths at the same level `image_url` already is.

Concretely:
- Unit tests in `pkg/sidecar/proxy/` for both encode-cache connectors (NIXL and shared_storage) that assert `video_url` and `input_audio` items fan out to encoders correctly — one encoder request per distinct video URL / per inline audio item, `ec_transfer_params` threaded onto the prefill body for NIXL, and no `ec_transfer_params` leakage for shared_storage.
- e2e helpers in `test/e2e/` for multi-video, multi-audio, and mixed-media (image + audio + video) chat completion requests, wired into the encode-prefill-decode combined-deployment path, the encode+prefill/decode split path, and the single-deployment path.

**Why is this needed**:

`connector_ec_common.go` treats `image_url`, `video_url`, `audio_url`, and `input_audio` uniformly under `mmTypes`, but the existing sidecar unit tests and e2e flows only cover the image case. That leaves a gap where a regression in the encoder fan-out for audio or video items — e.g., an item type getting silently dropped from the primer path, or `ec_transfer_params` not being threaded back for inline audio — could ship without any test catching it.

Addressed by #2639.

dmitripikus:
**What type of PR is this?**
/kind test

Extends test coverage for audio and video multimodal requests, matching the coverage image requests already have.

- `pkg/sidecar/proxy/connector_ec_nixl_test.go`: `TestHandleECNIXL_Video` and `TestHandleECNIXL_Audio` assert that `video_url` and `input_audio` items flow through the NIXL encoder fan-out the same way image items do — one encoder request per distinct URL / per inline audio item, with `ec_transfer_params` threaded through to the prefill body under the per-item hash key.
- `pkg/sidecar/proxy/connector_ec_shared_storage_test.go`: `TestHandleECSharedStorage_Video` and `TestHandleECSharedStorage_Audio` cover the primer fire-and-forget path — encoder called once per item, and no `ec_transfer_params` leak onto the outgoing prefill body (shared_storage discards encoder responses).
- `test/e2e/requests_test.go`: adds `runChatCompletionWithVideos`, `runChatCompletionWithAudios`, and `runChatCompletionWithMixedMedia` helpers that build multi-item and mixed-modality chat completion payloads.
- `test/e2e/e2e_test.go`: exercises the new helpers in the encode/prefill/decode combined-deployment test and adds a single-audio request to the E+PD and single-deployment test paths. Encode-decode counter expectation is updated from 5 to 8 for the added multi-audio / multi-video / mixed-media traffic.

Fixes #2640

**Release note**:
```release-note
NONE
```


pkg/sidecar/proxy/connector_ec_multimedia_test.go
```diff
@@ -0,0 +1,157 @@
...
+			wantEncCalls: 2,
+		},
+		{
+			name:    "nixl audio",
+			handler: (*Server).handleECNIXL,
+			items: []map[string]any{
+				inlineAudioItem("aaa"),
+				inlineAudioItem("bbb"),
+			},
+			wantECParams: true,
```
roytman, line 82: Both audio rows in this table use `inlineAudioItem` (inline input_audio), there's no row for audio_url. `mmTypes` in `connector_ec_common.go` treats `audio_url` as a distinct, URL-based, dedup-eligible type in the same bucket as `image_url/video_url`, so video gets full URL+dedup coverage here but audio only gets the inline path. Worth adding `audio_url rows` (nixl + shared_storage) mirroring the video ones, using two distinct URLs the way the video rows do.

test/e2e/e2e_test.go
```diff
@@ -616,29 +616,37 @@ var _ = ginkgo.Describe("Run end to end tests", func() {
...
+				// Multi-image request: two images in one request.
+				func() (string, string) { return runChatCompletionWithImages(testImageURL, testImageURL2) },
+				// Video request: video_url triggers encode stage.
+				func() (string, string) { return runChatCompletionWithVideos() },
+				// Audio request: input_audio triggers encode stage.
+				func() (string, string) { return runChatCompletionWithAudios() },
+				// Multi-audio request: two input_audio blocks in a single request.
+				func() (string, string) { return runChatCompletionWithAudios(testAudioData, testAudioData) },
+				// Multi-video request: two video_url blocks in a single request.
+				func() (string, string) { return runChatCompletionWithVideos(testVideoURL, testVideoURL) },
```
roytman: Same URL passed twice, `mmItemsForFanout`'s URL dedup collapses this to a single encoder item, so it doesn't exercise multi-item video fan-out the way the "Multi-image" entry above it does (two distinct URLs: `testImageURL`, `testImageURL2`). Doesn't break anything since `len(mmRequests)` still holds, but the comment "two video_url blocks" overstates what's actually covered here,  it's re-exercising dedup, not fan-out.

test/e2e/generate_endpoint_test.go
```diff
@@ -21,28 +21,35 @@ const (
...
-// twoImages is the canonical two-image spec for fan-out coverage.
+// twoImages is the canonical two-item spec for fan-out coverage.
 // Distinct lengths so the two placeholder spans aren't accidentally identical.
-var twoImages = []imageSpec{
+var twoImages = []mmSpec{
 	{Hash: "e2e-image-hash-0", Offset: 1, Length: 3},
 	{Hash: "e2e-image-hash-1", Offset: 4, Length: 5},
 }
 
+// mmModalities lists the modality labels each /inference/v1/generate
```
roytman, line 48: `mmTypes` is a `map[string]bool`, it has no order to match and its keys (`image_url`, `audio_url`, `video_url`, `input_audio`) aren't the same strings as these modality labels (`image/video/audio`) anyway; these are the coarser vLLM mm_hashes bucket names, not the OpenAI content-type strings `mmTypes` keys on. Suggest dropping the "order matches" sentence, or replacing it with something like "modality names as vLLM's mm_hashes buckets them — distinct from the content-type strings in mmTypes."

pkg/sidecar/proxy/connector_ec_multimedia_test.go
dmitripikus: Added nixl audio_url and shared_storage audio_url rows with two distinct URLs, mirroring the video rows.
Thanks!

test/e2e/e2e_test.go
dmitripikus: Added testVideoURL2 (same URL as testVideoURL with ?v=2 on the end). The router only compares URL strings to decide duplicates, and the mock server never actually downloads the video, so a different string is all that's needed. Passing both URLs to the row makes it a real multi-video test instead of one that gets collapsed to a single video.
Thanks!

test/e2e/generate_endpoint_test.go
dmitripikus: Reworded the comment to say these are vLLM mm_hashes bucket names, distinct from the OpenAI content-type keys in mmTypes.
Thanks!

roytman (approved):

elevran (approved):
approving on @roytman's LGTM

PR #2643, opened 2026-09-01
feat: emit OpenTelemetry structured JSON logs to stdout

den-rgb:
**What type of PR is this?**
/kind feature

JSON logs currently use zap keys (`msg`, `level`). This emits the OpenTelemetry Logs Data Model on stdout (`body`, `severity_text`, `severity_number`, `timestamp`, `service.name`) and injects `trace_id`/`span_id` on the request path when a span is active, so collectors can correlate logs with traces.

 As an Inference Component Developer
 I want MaaS, llm-d EPP, KServe controllers, and vLLM to emit structured JSON logs to stdout adhering to the OpenTelemetry Logs Data Model and including trace_id and span_id

So that all inference log records are standardized, machine-parsable, and directly correlated with active distributed traces.

### Acceptance Criteria

Standardized JSON Schema: All four components (MaaS, llm-d EPP, KServe controllers, and vLLM) emit logs to stdout formatted as structured JSON conforming to the OpenTelemetry Logs Data Model.

Trace Context Injection: When a trace context is active, the logger automatically extracts and includes trace_id and span_id within the log payload.

Severity Mapping: Standardizes log levels (e.g., INFO, WARN, ERROR) to standard OTel severity representation.

Related: https://redhat.atlassian.net/browse/RHOAIENG-83718

```release-note
JSON logs use OpenTelemetry field names (`body`, `severity_text`, `severity_number`) and include `trace_id`/`span_id` when a trace is active. Collectors that match on `msg` need to match `body` instead.
```

### DIFF
--- a/pkg/common/observability/logging/logger.go
+++ b/pkg/common/observability/logging/logger.go
@@ -33,38 +33,16 @@ import (
 // level can be adjusted after the controller-runtime delegation is fulfilled.
 var atomicLevel = uberzap.NewAtomicLevelAt(zapcore.InfoLevel)
 
-// LevelEncoder maps negative Zap levels to human-readable names that match
-// the project's verbosity constants (VERBOSE=3, DEBUG=4, TRACE=5). Without
-// this, controller-runtime's zap bridge emits all V(n) calls as "debug" in
-// JSON output, which is misleading for V(1)-V(3) (verbose info).
+// LevelEncoder maps zap / logr verbosity levels to OTel severity_text.
 func LevelEncoder(l zapcore.Level, enc zapcore.PrimitiveArrayEncoder) {
-	if l >= 0 {
-		zapcore.LowercaseLevelEncoder(l, enc)
-		return
-	}
-
-	switch l {
-	case zapcore.Level(-1 * DEBUG): // V(4) -> "debug"
-		enc.AppendString("debug")
-	case zapcore.Level(-1 * TRACE): // V(5) -> "trace"
-		enc.AppendString("trace")
-	default:
-		if l >= zapcore.Level(-1*VERBOSE) { // V(1)-V(3) -> "info"
-			enc.AppendString("info")
-		} else { // V(6+) -> "trace"
-			enc.AppendString("trace")
-		}
-	}
+	enc.AppendString(SeverityText(l))
 }
 
-func InitSetupLogging() {
-	config := uberzap.NewProductionEncoderConfig()
-	config.EncodeLevel = LevelEncoder
-
-	logger := zap.New(
+func InitSetupLogging(serviceName string) {
+	logger := NewLogger(
+		serviceName,
 		zap.Level(atomicLevel),
 		zap.RawZapOpts(uberzap.AddCaller()),
-		zap.Encoder(zapcore.NewJSONEncoder(config)),
 	)
 	ctrl.SetLogger(logger)
 }
--- a/pkg/common/observability/logging/otel.go
+++ b/pkg/common/observability/logging/otel.go
@@ -1,0 +1,181 @@
+package logging
+
+import (
+	"context"
+	"os"
+
+	"github.com/go-logr/logr"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/sdk/resource"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	crzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+const (
+	otelTimestampKey      = "timestamp"
+	otelSeverityTextKey   = "severity_text"
+	otelLoggerKey         = "logger"
+	otelCallerKey         = "caller"
+	otelBodyKey           = "body"
+	otelStacktraceKey     = "stacktrace"
+	otelServiceNameKey    = "service.name"
+	otelSeverityNumberKey = "severity_number"
+
+	severityTrace  = "TRACE"
+	severityDebug  = "DEBUG"
+	severityInfo   = "INFO"
+	severityWarn   = "WARN"
+	severityError  = "ERROR"
+	severityDPanic = "DPANIC"
+	severityPanic  = "PANIC"
+	severityFatal  = "FATAL"
+
+	severityNumberTrace  = 1
+	severityNumberDebug  = 5
+	severityNumberInfo   = 9
+	severityNumberWarn   = 13
+	severityNumberError  = 17
+	severityNumberDPanic = 18
+	severityNumberPanic  = 19
+	severityNumberFatal  = 21
+)
+
+// HTTPBodyKey identifies a log field containing an HTTP payload.
+const HTTPBodyKey = "http_body"
+
+// NewLogger returns a logger with OpenTelemetry field names and severity
+// fields. Additional options customize the logger for each service.
+func NewLogger(serviceName string, opts ...crzap.Opts) logr.Logger {
+	defaultOpts := []crzap.Opts{
+		crzap.WriteTo(os.Stdout),
+		crzap.Encoder(zapcore.NewJSONEncoder(EncoderConfig())),
+		crzap.RawZapOpts(
+			zap.WrapCore(WrapCore),
+			zap.Fields(zap.String(otelServiceNameKey, ServiceName(serviceName))),
+		),
+	}
+	return crzap.New(append(defaultOpts, opts...)...)
+}
+
+// NewLoggerWithOptions returns an OpenTelemetry logger configured from
+// controller-runtime logging options.
+func NewLoggerWithOptions(serviceName string, options *crzap.Options) logr.Logger {
+	configured := *options
+	if configured.DestWriter == nil {
+		configured.DestWriter = os.Stdout
+	}
+	configured.Encoder = zapcore.NewJSONEncoder(EncoderConfig())
+	configured.ZapOpts = append(append([]zap.Option(nil), configured.ZapOpts...),
+		zap.WrapCore(WrapCore),
+		zap.Fields(zap.String(otelServiceNameKey, ServiceName(serviceName))),
+	)
+	return crzap.New(crzap.UseFlagOptions(&configured))
+}
+
+// EncoderConfig returns a zap encoder config that emits OTel Logs Data Model
+// field names on stdout JSON records.
+func EncoderConfig() zapcore.EncoderConfig {
+	config := zap.NewProductionEncoderConfig()
+	config.TimeKey = otelTimestampKey
+	config.LevelKey = otelSeverityTextKey
+	config.NameKey = otelLoggerKey
+	config.CallerKey = otelCallerKey
+	config.MessageKey = otelBodyKey
+	config.StacktraceKey = otelStacktraceKey
+	config.EncodeTime = zapcore.RFC3339NanoTimeEncoder
+	config.EncodeLevel = LevelEncoder
+	return config
+}
+
+// SeverityText maps zap / logr verbosity levels to OTel severity_text.
+func SeverityText(l zapcore.Level) string {
+	if l >= 0 {
+		switch {
+		case l >= zapcore.FatalLevel:
+			return severityFatal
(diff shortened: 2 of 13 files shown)


pkg/common/observability/logging/otel.go
```diff
@@ -0,0 +1,119 @@
...
+
+// EncoderConfig returns a zap encoder config that emits OTel Logs Data Model
+// field names on stdout JSON records.
+func EncoderConfig() zapcore.EncoderConfig {
+	config := zap.NewProductionEncoderConfig()
+	config.TimeKey = "timestamp"
+	config.LevelKey = "severity_text"
+	config.NameKey = "logger"
+	config.CallerKey = "caller"
+	config.MessageKey = "body"
```
gyliu513: `MessageKey = "body"` collides with the many existing call sites that already use `"body"` as a field key for raw HTTP request/response payloads, which do `v.Info("request body", ..., "body", RedactBody(bodyBytes))`.

The resulting JSON line ends up with two `"body"` keys, e.g. `{"body":"request body","body":"<payload>",...}`. Any strict JSON parser (including most log pipelines/OTel collectors) keeps only the last key, silently dropping the log message and surfacing the raw (possibly large/PII-containing) payload as the log's semantic body.

Could you rename the existing `"body"` field (used for raw payloads) to something like `"http_body"` or `"payload"` to avoid the collision?

pkg/epp/requestcontrol/director.go
```diff
@@ -201,7 +201,7 @@ func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestCo
 		span.End()
 	}()
 
-	logger := log.FromContext(ctx)
+	logger := logutil.WithTrace(ctx, log.FromContext(ctx))
```
gyliu513: `server.go:485` already calls `tracing.LoggerWithSpanContext(ctx, span)` once at request entry, guarded by an internal `correlatedKey{}` so nested spans don't repeat the trace fields (see that function's comment and `TestLoggerWithSpanContextEnrichesOncePerContext`).

`HandleRequest` starts a child span and then calls the new `logutil.WithTrace(ctx, log.FromContext(ctx))`, which has no knowledge of that guard. This produces two `trace_id` keys (same value) and two `span_id` keys (parent vs. child) in the same JSON line, since JSON-map consumers keep the last key, logs end up correlated to the child span_id instead of the entry span_id that `LoggerWithSpanContext` was designed to guarantee.

`WithTrace` is only called here, and this ctx already carries trace_id/span_id from `tracing.LoggerWithSpanContext` at request entry (server.go:485). Since `log.FromContext(ctx)` already returns the correlated logger, could this line just be:

and the `logutil.WithTrace` function be dropped, rather than teaching it to share the `correlatedKey{}` guard? That avoids introducing a second correlation mechanism (and a new cross-package coupling) for what the existing one already covers.

pkg/common/observability/logging/otel.go
gyliu513: Can you put all those values to const above?

pkg/common/observability/logging/otel.go
```diff
@@ -0,0 +1,119 @@
...
+// SeverityText maps zap / logr verbosity levels to OTel severity_text.
+func SeverityText(l zapcore.Level) string {
+	if l >= 0 {
+		switch {
+		case l >= zapcore.DPanicLevel:
+			return "FATAL"
+		case l >= zapcore.ErrorLevel:
+			return "ERROR"
+		case l >= zapcore.WarnLevel:
+			return "WARN"
```
gyliu513: Can you put all those values to const above?

pkg/common/observability/logging/otel.go
```diff
@@ -0,0 +1,119 @@
...
+	config.EncodeTime = zapcore.RFC3339NanoTimeEncoder
+	config.EncodeLevel = LevelEncoder
+	return config
+}
+
+// SeverityText maps zap / logr verbosity levels to OTel severity_text.
+func SeverityText(l zapcore.Level) string {
+	if l >= 0 {
+		switch {
+		case l >= zapcore.DPanicLevel:
```
gyliu513, line 99: `SeverityText` maps `DPanicLevel`, `PanicLevel`, and `FatalLevel` all to `"FATAL"`/21, losing the previous distinction between a recoverable dev-mode `DPanic`/`Panic` and a genuine process-terminating `Fatal` (`os.Exit`).

This means alerting keyed on OTel FATAL severity can no longer distinguish an actual crash from a recoverable assertion — worth double-checking this collapse is intentional, since it's not covered by any test (logger_test.go / otel_test.go only exercise Info/Warn/Error/Debug/Trace).

pkg/sidecar/proxy/options.go
```diff
@@ -697,11 +697,14 @@ func validatePortRange(startPort, rangeSize int) error {
 // with a custom level encoder that maps verbosity levels to their semantic
```
gyliu513: This doc comment on `NewLogger` still describes the pre-PR implementation ("custom level encoder that maps verbosity levels to their semantic names instead of always rendering V(n) as debug"), but the diff replaced that direct `EncodeLevel` assignment with `logutil.EncoderConfig()` (which also renames `msg`→`body`, `ts`→`timestamp`, uppercases severity text) plus the new `WrapCore`/`service.name` wiring — none of which the comment mentions. Could this be updated to describe the current behavior?

pkg/common/observability/logging/logger.go
```diff
@@ -33,38 +33,21 @@ import (
...
-	config := uberzap.NewProductionEncoderConfig()
-	config.EncodeLevel = LevelEncoder
-
 	logger := zap.New(
+		zap.WriteTo(os.Stdout),
 		zap.Level(atomicLevel),
-		zap.RawZapOpts(uberzap.AddCaller()),
-		zap.Encoder(zapcore.NewJSONEncoder(config)),
+		zap.RawZapOpts(
+			uberzap.AddCaller(),
```
gyliu513: The OTel wiring (`WrapCore` + `Fields(service.name)` + `Encoder(EncoderConfig())`) is duplicated between `InitSetupLogging` here and `NewLogger` in `pkg/sidecar/proxy/options.go`, with no shared constructor. Also, `AddCaller()` is present in one but not the other, is that intentional?

Worth extracting a shared helper so future fixes (e.g. the two issues above) only need to be applied in one place.

pkg/common/observability/logging/otel.go
den-rgb: good catch, done

pkg/epp/requestcontrol/director.go
den-rgb: removed WithTrace now using the context logger directly

pkg/sidecar/proxy/options.go
den-rgb: fixed, updated NewLogger

pkg/common/observability/logging/logger.go
den-rgb: moved the shared OTel logger construction into logging.NewLogger. AddCaller stays only on the router logger to preserve the sidecar’s existing behavior.

pkg/sidecar/proxy/options.go
```diff
@@ -693,15 +691,12 @@ func validatePortRange(startPort, rangeSize int) error {
...
-// names instead of always rendering V(n) as "debug".
+// NewLogger returns a logger configured from the Options logging flags with
+// OpenTelemetry field names and severity fields.
 func (opts *Options) NewLogger() logr.Logger {
-	config := uberzap.NewProductionEncoderConfig()
-	config.EncodeLevel = logutil.LevelEncoder
-	return zap.New(
+	return logutil.NewLogger(
+		"llm-d-router-disagg-sidecar",
 		zap.UseFlagOptions(&opts.loggingOptions),
```
gyliu513: zap.UseFlagOptions(&opts.loggingOptions) is passed as the last option into logutil.NewLogger, and UseFlagOptions does *o = *in , a full struct overwrite. That wipes out the encoder/writer/zap-opts defaults NewLogger just set (OTel JSON encoder, stdout, service.name field), so the sidecar falls back to plain zap JSON on stderr with legacy msg/level/ts fields. The PR's core logging feature doesn't actually take effect for this binary.

pkg/common/observability/logging/logger.go
```diff
@@ -33,38 +33,16 @@ import (
...
-	}
+	enc.AppendString(SeverityText(l))
 }
 
 func InitSetupLogging() {
-	config := uberzap.NewProductionEncoderConfig()
-	config.EncodeLevel = LevelEncoder
-
-	logger := zap.New(
+	logger := NewLogger(
```
gyliu513, line 42: InitSetupLogging() hardcodes "llm-d-router" as the service name, and it's called identically from cmd/coordinator/main.go:59 and cmd/epp/runner/runner.go:215. Both processes end up emitting the same service.name, so a log pipeline can't distinguish coordinator logs from EPP logs, this defeats the per-component correlation this PR is meant to enable. The sidecar already threads its own distinct name ("llm-d-router-disagg-sidecar") into logutil.NewLogger, so the fix pattern (pass a per-binary service name) already exists in this diff.

pkg/common/observability/logging/otel.go
```diff
@@ -0,0 +1,156 @@
...
+		return severityNumberDebug
+	case severityTrace:
+		return severityNumberTrace
+	default:
+		return severityNumberInfo
+	}
+}
+
+// ServiceName returns OTEL_SERVICE_NAME or fallback.
+func ServiceName(fallback string) string {
```
gyliu513, line 146: ServiceName() only checks OTEL_SERVICE_NAME. The existing tracing package (newResource() in pkg/common/observability/tracing/telemetry.go) resolves service name via resource.WithFromEnv(), which also honors OTEL_RESOURCE_ATTRIBUTES. An operator who sets service name only through OTEL_RESOURCE_ATTRIBUTES=service.name=foo (a supported convention, e.g. via the K8s downward API) will get traces tagged service.name=foo but logs falling back to the hardcoded default — logs and traces for the same process disagree, breaking the correlation this PR exists for.

pkg/coordinator/gateway/client.go
```diff
@@ -117,7 +117,7 @@ func (c *Client) Request(ctx context.Context, method, path string, body []byte,
 		if err != nil {
 			return nil, fmt.Errorf("reading response from gateway: %w", err)
 		}
-		v.Info("response body", "status", resp.StatusCode, "body", RedactBody(respBody))
+		v.Info("response body", "status", resp.StatusCode, "http_body", RedactBody(respBody))
```
gyliu513: I'd prefer a const value for `http_body`

pkg/common/observability/logging/otel.go
den-rgb: ServiceName now reads service.name through resource.WithFromEnv, with OTEL_SERVICE_NAME precedence

pkg/common/observability/logging/logger.go
den-rgb: coordinator and EPP now pass distinct service names

pkg/sidecar/proxy/options.go
den-rgb: UseFlagOptions overwrote the defaults, fixed locally with NewLoggerWithOptions, which reapplies the OTel logger configuration

pkg/common/observability/logging/otel_test.go
```diff
@@ -0,0 +1,130 @@
...
+	logger.Info("request body", HTTPBodyKey, `{"prompt":"hello"}`)
+
+	var rec map[string]any
+	if err := json.Unmarshal(buf.Bytes(), &rec); err != nil {
+		t.Fatal(err)
+	}
+	if rec["body"] != "request body" {
+		t.Errorf("body = %v, want request body", rec["body"])
+	}
+	if rec["http_body"] != `{"prompt":"hello"}` {
```
gyliu513: nit: may also need to use `logutil.HTTPBodyKey`?

ahg-g (approved):

PR #2644, opened 2026-09-01
fix(tokenizer): preserve render request key order - fixes #2641

### LINKED ISSUE #2641: /render breaking prefix affinity for precise KV cache routing
**What happened**:

After building a new version of EPP based on upstream main, we noticed a persistent decrease in prefix cache hit rate (all configurations stayed exactly the same) while using precise KV cache routing. The degradation has been noticeable and sustained for a few days, eliminating the hypothesis this is due to traffic changes.

An isolated reproduction indicates that, after #2318, the token sequence computed through `/render` can differ from the token sequence generated by vLLM for the request it actually receives.

This affects precise KV cache aware routing but not approx (as approx will use its own hashing and not rely on vLLM tokens received through ZMQ, so it remains self-consistent).

We expected no degradation of prefix cache hit rate.

I'm attaching a reproduction Python script. The experiment consists in setting up two llm-d deployments, one 
built from a version prior to PR #2318  being merged (we'll call it Control), and one built from the tip of upstream main (we'll call it Treatment). The underlying LM is irrelevant for the reproduction, we used vLLM and `qwen-3-5-9b`.

We then fire a sample request to both llm-d deployments:

```
{
    "model": <model>,
    "messages": [
      {
        "role": "user",
        "content": "Case unsorted_<nonce>: use the weather tool for Paris. This sentence pads the prompt for deterministic cache testing. <repeated 30 times>"
      }
    ],
    "tool_choice": "none",
    "tools": [
      {
        "type": "function",
        "function": {
          "name": "weather_unsorted_<nonce>",
          "description": "Look up current weather for a city.",
          "parameters": {
            "type": "object",
            "properties": {
              "city": {
                "type": "string",
                "description": "City name"
              },
              "units": {
                "type": "string",
                "enum": ["celsius", "fahrenheit"],
                "description": "Temperature units"
              }
            },
            "required": ["city"]
          }
        }
      }
    ],
    "max_tokens": 1,
    "stream": false
  }
```

The first time the request is sent (assuming vLLM pods are cold), we don't get any cache hits as expected.

We then wait a few seconds, and send the same payload again. We observe an EPP cache-affinity match for Control, but not for Treatment.

Finally, we send a sorted version of the request for Treatment, wait a few seconds, and send the same payload again. We then observe cache affinity matches on Treatment.

Our understanding is that this indicates PR #2318 should also prevent  reordering the request when `/render` is called. I believe what happens is somewhere in the `/render` code path some JSON reordering is done, so the token bytes received don't exactly match the ones from the original request.

Appendix:

To check whether llm-d had a prefix cache affinity match or not, we inspect DEBUG logs eg:

```
{
    "msg": "PrefixCacheAffinityFilter: narrowed to sticky",
    "x-request-id": "2f94b3a9-e944-441b-9968-71a715ed5de8",
    "affinityThreshold": 0.8,
    "sticky": 1,
    "total": 2
  }
```
To use the reproduction script run once for each llm-d deployment (Control and Treatment):

```
python3 reproduce_qwen_render_parity_share_publicly.py \
    --llm-d-endpoint https://router.example.test \
    --model <model-name> \
    --engine-block-tokens <engine-block-size> \
    --router-block-tokens <router-block-size>
```

[reproduce_qwen_render_parity_share_publicly.py](https://github.com/user-attachments/files/31685809/reproduce_qwen_render_parity_share_publicly.py)

**Environment**:
- Kubernetes version (use `kubectl version`): v1.36.1
- llm-d-router version (use `git describe --tags --dirty --always` if you built from source, or specify the tag if you used a tagged version or image): image built on top of upstream main (commit 618eb0a99470edd7afdc8afced5df0cbef18e163)
- Cloud provider or hardware configuration: Non-applicable
- Install tools: Non-applicable
- Others:

Lucas-Fernandes-Martins:
**What type of PR is this?**

Fixes #2641, which seemingly breaks precise KV cache as the `/render` code path marshals and unmarshals the raw body, potentially leading to llm-d index seeing different token ID's than those received via ZMQ messages.

Tentative solutuion through which we check if EPP has not mutated the body, and if so, we use the `rawBody` directly to call `/render`

#2641
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/interface/scheduling/types.go
+++ b/pkg/epp/framework/interface/scheduling/types.go
@@ -58,6 +58,8 @@ type InferenceRequest struct {
 	FairnessID string
 	// RequestSizeBytes is the size of the raw request body in bytes when available.
 	RequestSizeBytes int
+	// RawBody contains the request body exactly as received from the downstream proxy.
+	RawBody []byte
 	// SchedulingResult captures the scheduling decisions made during the cycle.
 	SchedulingResult *SchedulingResult
 
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
@@ -27,12 +27,19 @@ import (
 
 	logutil "github.com/llm-d/llm-d-router/pkg/common/observability/logging"
 	fwkrh "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/requesthandling"
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/scheduling"
 )
 
 // tokenInputProducer turns a request body into a TokenizedRequest. Backends vary
 // in fidelity (render vs estimate); callers never branch on which produced it.
 type tokenInputProducer interface {
 	produce(ctx context.Context, body *fwkrh.InferenceRequestBody) (*fwkrh.TokenizedRequest, error)
+}
+
+// requestInputProducer is implemented by backends that need request metadata
+// in addition to the parsed body.
+type requestInputProducer interface {
+	produceRequest(ctx context.Context, request *scheduling.InferenceRequest) (*fwkrh.TokenizedRequest, error)
 }
 
 // timeoutAware is implemented by backends (and the tokenizers they wrap) whose
@@ -102,6 +109,30 @@ func warmupChat(imageURLs ...string) *fwkrh.InferenceRequestBody {
 // the pre-tokenized (Generate) passthrough.
 type renderBackend struct {
 	tk tokenizer
+}
+
+// produceRequest sends the original chat-completions JSON to vLLM when it is
+// also the body that will be forwarded downstream, preserving object key order.
+func (b renderBackend) produceRequest(ctx context.Context, request *scheduling.InferenceRequest) (*fwkrh.TokenizedRequest, error) {
+	body := request.Body
+	renderer, ok := b.tk.(*vllmHTTPRenderer)
+	if !ok || body.ChatCompletions == nil || body.Mutated || len(request.RawBody) == 0 ||
+		request.TargetModel != renderer.modelName || body.Model != renderer.modelName || body.Payload == nil {
+		return b.produce(ctx, body)
+	}
+
+	payload, ok := body.Payload.AsMap()
+	if !ok {
+		return b.produce(ctx, body)
+	}
+	tokenIDs, mmFeatures, err := renderer.postRawChatRender(ctx, request.RawBody, renderer.chatTimeout(payload))
+	if err != nil {
+		return nil, fmt.Errorf("tokenization failed: %w", err)
+	}
+	return &fwkrh.TokenizedRequest{Prompts: []fwkrh.PromptTokens{{
+		TokenIDs:           tokenIDs,
+		MultiModalFeatures: convertMMFeaturesToUpstream(mmFeatures),
+	}}}, nil
 }
 
 func (b renderBackend) produce(ctx context.Context, body *fwkrh.InferenceRequestBody) (*fwkrh.TokenizedRequest, error) {
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
@@ -180,6 +180,14 @@ func (r *vllmHTTPRenderer) postChatRender(ctx context.Context, body any, timeout
 	return resp.TokenIDs, toKVCacheMM(resp.Features), nil
 }
 
+func (r *vllmHTTPRenderer) postRawChatRender(ctx context.Context, body []byte, timeout time.Duration) ([]uint32, *tokenization.MultiModalFeatures, error) {
+	var resp renderResponse
+	if err := r.postJSONPayload(ctx, chatRenderPath, body, timeout, &resp); err != nil {
+		return nil, nil, err
+	}
+	return resp.TokenIDs, toKVCacheMM(resp.Features), nil
+}
+
 func (r *vllmHTTPRenderer) chatTimeout(payload fwkrh.PayloadMap) time.Duration {
 	messages, ok := payload["messages"].([]any)
 	if !ok {
@@ -349,7 +357,10 @@ func (r *vllmHTTPRenderer) postJSON(ctx context.Context, path string, body any, 
 	if err != nil {
 		return fmt.Errorf("marshal request: %w", err)
 	}
+	return r.postJSONPayload(ctx, path, payload, timeout, out)
+}
 
+func (r *vllmHTTPRenderer) postJSONPayload(ctx context.Context, path string, payload []byte, timeout time.Duration, out any) error {
 	reqCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http_test.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http_test.go
@@ -248,6 +248,96 @@ func TestProduce_ChatCompletionsVLLMHTTPUsesRawPayload(t *testing.T) {
 	assert.Equal(t, testHTTPModel, sent["model"])
 }
 
+func TestProduce_ChatCompletionsVLLMHTTPPreservesRequestBodyOrder(t *testing.T) {
+	srv, captured := httpFixture(t, nil, renderResponse{TokenIDs: []uint32{9, 10}})
+	defer srv.Close()
+
+	rawBody := []byte(`{
+  "model": "test-model",
+  "messages": [{"role": "user", "content": "hello"}],
+  "tools": [{"type": "function", "function": {"name": "lookup", "parameters": {
+    "type": "object",
+    "properties": {"zeta": {"type": "string", "description": "last"}, "alpha": {"type": "string", "description": "first"}}
+  }}}]
+}`)
+	var payload fwkrh.PayloadMap
+	require.NoError(t, json.Unmarshal(rawBody, &payload))
+	req := &scheduling.InferenceRequest{
+		TargetModel: testHTTPModel,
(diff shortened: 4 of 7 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
```diff
@@ -104,6 +111,30 @@ type renderBackend struct {
 	tk tokenizer
 }
 
+// produceRequest sends the original chat-completions JSON to vLLM when it is
+// also the body that will be forwarded downstream, preserving object key order.
+func (b renderBackend) produceRequest(ctx context.Context, request *scheduling.InferenceRequest) (*fwkrh.TokenizedRequest, error) {
+	body := request.Body
+	renderer, ok := b.tk.(*vllmHTTPRenderer)
+	if !ok || body.ChatCompletions == nil || body.Mutated || len(request.RawBody) == 0 ||
+		request.TargetModel != renderer.modelName || body.Model != renderer.modelName || body.Payload == nil {
```
nilig, line 120: The raw path only turns on when the model name in the request exactly matches the tokenizer's configured `modelName`. If clients call the model by a different accepted name (a vLLM `--served-model-name` alias, or a LoRA adapter name), the request is not considered mutated, so the engine still gets the original bytes while `/render` gets the reordered copy. For those deployments the #2641 bug stays. That case was broken before #2318 too, so I don't think it needs to hold up this PR. Suggest we file a follow-up issue for it (a likely shape: replace just the model value inside the raw bytes without reordering anything else, the way #2511 keys purely off `Mutated` on the Anthropic path), and in this PR just note the limitation in a comment on `produceRequest` and in the README's `modelName` row so the difference between the two paths is on record.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
```diff
@@ -104,6 +111,30 @@ type renderBackend struct {
...
+		return b.produce(ctx, body)
```
nilig, line 121: There is no way to see which path a request took. No log line, no metric, and both paths produce the same trace span. So if a deployment fails one of the conditions, it looks exactly like the original bug: routing quietly stops matching and nothing tells you why. The original issue took days to notice for the same reason. A DEBUG log on the fallback saying which condition failed (and printing both model names when they differ) would fix that.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
```diff
@@ -104,6 +111,30 @@ type renderBackend struct {
 	tk tokenizer
 }
 
+// produceRequest sends the original chat-completions JSON to vLLM when it is
+// also the body that will be forwarded downstream, preserving object key order.
+func (b renderBackend) produceRequest(ctx context.Context, request *scheduling.InferenceRequest) (*fwkrh.TokenizedRequest, error) {
+	body := request.Body
+	renderer, ok := b.tk.(*vllmHTTPRenderer)
```
nilig, line 118: nit: this is a concrete-type check; everywhere else this package uses small interfaces (`timeoutAware`, `anthropicRenderer`). Same behavior either way today, so style only.

pkg/epp/framework/interface/scheduling/types.go
```diff
@@ -58,6 +58,8 @@ type InferenceRequest struct {
 	FairnessID string
 	// RequestSizeBytes is the size of the raw request body in bytes when available.
 	RequestSizeBytes int
+	// RawBody contains the request body exactly as received from the downstream proxy.
+	RawBody []byte
```
nilig, line 62: The whole fix assumes nothing modifies the request body after the tokenizer runs. That's true today, but only by accident: `MutatePayloadMap` is public, and admission and PreRequest plugins run after the tokenizer, so a third-party plugin that adds any field to the body would quietly reintroduce this exact bug, and no test would catch it. Let's write the rule down: the `RawBody` doc should say the director forwards these bytes unchanged only while `Mutated` is false, and `produceRequest` should get one sentence explaining why it needs the same bytes.

nit: `RequestSizeBytes` and `RawBody` now store the same information separately. Setting one from the other would keep them from disagreeing.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http_test.go
```diff
@@ -248,6 +248,96 @@ func TestProduce_ChatCompletionsVLLMHTTPUsesRawPayload(t *testing.T) {
...
+			Model:   testHTTPModel,
+		},
+	}
+
+	p := newTestPlugin(newHTTPRenderer(t, srv))
+	require.NoError(t, p.Produce(context.Background(), req, nil))
+	assert.Equal(t, rawBody, captured.chat)
+}
+
+func TestProduce_ChatCompletionsVLLMHTTPUsesMutatedPayload(t *testing.T) {
```
nilig, line 282: Each of the fallback conditions should have its own test. Right now this test doesn't set `Body.Model`, so it falls back because of the model check, not the mutated check; removing the `Mutated` condition wouldn't fail any test. The `TargetModel` check, empty `RawBody`, nil payload, and protobuf payload have no tests at all. A small table test that flips one condition at a time would cover all of them.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http_test.go
```diff
@@ -248,6 +248,96 @@ func TestProduce_ChatCompletionsVLLMHTTPUsesRawPayload(t *testing.T) {
...
+				Messages: []fwkrh.Message{{Role: "user", Content: fwkrh.Content{Raw: "hello"}}},
+			},
+			Payload: payload,
+			Model:   testHTTPModel,
+		},
+	}
+
+	p := newTestPlugin(newHTTPRenderer(t, srv))
+	require.NoError(t, p.Produce(context.Background(), req, nil))
+	assert.Equal(t, rawBody, captured.chat)
```
nilig, line 279: nit: this test checks the bytes that were sent but never checks the tokens that came back, so the response handling of `postRawChatRender` is untested. Copying the token assertions from the neighboring tests would close that.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
```diff
@@ -180,6 +180,14 @@ func (r *vllmHTTPRenderer) postChatRender(ctx context.Context, body any, timeout
 	return resp.TokenIDs, toKVCacheMM(resp.Features), nil
 }
 
+func (r *vllmHTTPRenderer) postRawChatRender(ctx context.Context, body []byte, timeout time.Duration) ([]uint32, *tokenization.MultiModalFeatures, error) {
```
nilig, line 183: nit: this duplicates `postChatRender` minus the marshal; `postChatRender` could just marshal and call `postRawChatRender`. Same for `produceRequest` copying the body of `produce`'s chat branch.

pkg/epp/requestcontrol/director.go
```diff
@@ -227,6 +227,7 @@ func (d *Director) HandleRequest(ctx context.Context, reqCtx *handlers.RequestCo
 		RequestID:        reqCtx.Request.Headers[reqcommon.RequestIDHeaderKey],
 		TargetModel:      reqCtx.TargetModelName,
 		Body:             inferenceRequestBody,
+		RawBody:          reqCtx.Request.RawBody,
```
nilig, line 230: nit: `RawBody` is attached even for mutated requests, where nothing ever reads it, and it keeps the original body in memory until the stream ends. Wonder if we could set it only when the body wasn't mutated.

nilig (commented):
The fix goes in the right direction: the only reliable way to keep the router and vLLM in agreement is to send both of them the exact same bytes. My comments are about the cases the fix doesn't cover, the fact that you can't tell when it's active, and test coverage. None of them block the approach. Details are inline.

Two points on the PR itself:

- The release note says NONE, but this fixes routing behavior operators can see, so it should get a line. Otherwise the fix won't appear in the release notes at all.
- nit: "solutuion" typo in the description; the description says the body is "marshaled and unmarshaled", but what actually happens is the parsed map gets re-serialized with sorted keys while the engine gets the original bytes; and `/kind bug` is still inside the HTML comment, so the PR never got its kind label.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
```diff
@@ -104,6 +111,30 @@ type renderBackend struct {
...
+	if !ok || body.ChatCompletions == nil || body.Mutated || len(request.RawBody) == 0 ||
+		request.TargetModel != renderer.modelName || body.Model != renderer.modelName || body.Payload == nil {
+		return b.produce(ctx, body)
+	}
+
+	payload, ok := body.Payload.AsMap()
```
vMaroon, line 124: nit: neither the `body.Payload == nil` guard nor the `payload.AsMap()` failure fallback (lines 124-127) is covered. A case with a nil or non-map payload asserting the fallback to `produce` would lock that in.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
```diff
@@ -104,6 +111,30 @@ type renderBackend struct {
...
+	if !ok {
+		return b.produce(ctx, body)
+	}
+	tokenIDs, mmFeatures, err := renderer.postRawChatRender(ctx, request.RawBody, renderer.chatTimeout(payload))
+	if err != nil {
+		return nil, fmt.Errorf("tokenization failed: %w", err)
+	}
+	return &fwkrh.TokenizedRequest{Prompts: []fwkrh.PromptTokens{{
```
vMaroon, line 132: nit: this `TokenizedRequest` construction is now the third verbatim copy of the same `{Prompts: []PromptTokens{{TokenIDs, MultiModalFeatures: convertMMFeaturesToUpstream(...)}}}` shape (also at lines 150-153 and 159-162 in `produce`). Wonder if we could extract a helper, e.g. `newRenderResult(tokenIDs []uint32, mm *tokenization.MultiModalFeatures) *fwkrh.TokenizedRequest`, and reuse it in all three.

PR #2645, opened 2026-09-01
fix: warn when refresh-metrics-interval exceeds metricsStalenessThres…

### LINKED ISSUE #2514: flowcontrol: warn when metrics refresh interval exceeds staleness threshold
## What would be useful

Emit a startup warning when the data-layer `--refresh-metrics-interval` is longer than a metrics-based detector's staleness threshold, such as `utilization-detector.parameters.metricsStalenessThreshold`.

## Why this is needed

The router defaults are aligned (`refresh-metrics-interval=50ms`, with a 50ms clamp floor, and `metricsStalenessThreshold=200ms`). Deployments can override the refresh interval independently, however, with no warning that the detector threshold is now shorter than the polling cadence.

We observed a deployment configured with:

```text
--refresh-metrics-interval=15s
utilization metricsStalenessThreshold=200ms
```

In a single-endpoint pool, each sample was fresh only briefly after a successful poll and stale for about 98.7% of each cycle. With the default `stalenessPolicy=saturated`, that repeatedly produced `saturation=1`, HoL blocking, and 11-13 second gateway/EPP request latency while the runtime itself remained responsive. Configuring `metricsStalenessThreshold=20s` removed the periodic stale condition and returned request latency to the millisecond range.

## Possible shape

During startup, after plugin defaults and the data-layer polling interval are known, log a warning when:

```text
refreshMetricsInterval > metricsStalenessThreshold
```

The warning should identify the detector and both effective values, and explain that metrics will be stale between polls. This should remain a warning rather than validation failure because deployments may intentionally choose this posture or use another data source cadence.

The comparison should use effective values after the refresh-interval floor clamp and plugin defaulting. If a data source has its own polling interval, that effective source cadence may be more accurate than the runtime base tick.

This is tracked separately from #2408 / #2492, which address how stale samples contribute to saturation rather than detecting a cadence mismatch.

varad-ahirwadkar:
**What type of PR is this?**
/kind bug

When `--refresh-metrics-interval` exceeds a saturation detector's `metricsStalenessThreshold`, endpoint metrics may become stale between polls. This can cause the detector to treat healthy endpoints as saturated, leading to backpressure and increased request latency.

This PR adds a startup warning that identifies the detector and both effective values, helping operators detect and correct the configuration before serving traffic.

Fixes #2514

```release-note
Added a startup warning when `--refresh-metrics-interval` exceeds a saturation detector's `metricsStalenessThreshold`.
```

### DIFF
--- a/pkg/epp/framework/interface/plugin/handle.go
+++ b/pkg/epp/framework/interface/plugin/handle.go
@@ -19,6 +19,7 @@ package plugin
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"k8s.io/apimachinery/pkg/types"
 )
@@ -44,6 +45,11 @@ type Handle interface {
 	// SetCrossReplicaSyncer makes the configured cross-replica syncer available
 	// to plugins through the handle.
 	SetCrossReplicaSyncer(Plugin)
+
+	// RefreshMetricsInterval returns the base polling tick configured by
+	// --refresh-metrics-interval, after the minimum-interval clamp. Data sources
+	// that set their own interval may poll less often than this. Zero when unset.
+	RefreshMetricsInterval() time.Duration
 }
 
 // HandlePlugins defines a set of APIs to work with instantiated plugins
@@ -68,9 +74,10 @@ type PodListFunc func() []types.NamespacedName
 type eppHandle struct {
 	ctx context.Context
 	HandlePlugins
-	podList            PodListFunc
-	metricsRecorder    MetricsRecorder
-	crossReplicaSyncer Plugin
+	podList                PodListFunc
+	metricsRecorder        MetricsRecorder
+	crossReplicaSyncer     Plugin
+	refreshMetricsInterval time.Duration
 }
 
 // Context returns a context the plugins can use, if they need one
@@ -130,6 +137,11 @@ func (h *eppHandle) SetCrossReplicaSyncer(syncer Plugin) {
 	h.crossReplicaSyncer = syncer
 }
 
+// RefreshMetricsInterval returns the data-layer polling cadence.
+func (h *eppHandle) RefreshMetricsInterval() time.Duration {
+	return h.refreshMetricsInterval
+}
+
 // HandleOption configures an eppHandle constructed via NewEppHandle.
 type HandleOption func(*eppHandle)
 
@@ -140,6 +152,14 @@ func WithMetricsRecorder(recorder MetricsRecorder) HandleOption {
 		if recorder != nil {
 			h.metricsRecorder = recorder
 		}
+	}
+}
+
+// WithRefreshMetricsInterval sets the base polling tick advertised to plugins
+// via Handle.RefreshMetricsInterval.
+func WithRefreshMetricsInterval(interval time.Duration) HandleOption {
+	return func(h *eppHandle) {
+		h.refreshMetricsInterval = interval
 	}
 }
 
--- a/pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/detector.go
+++ b/pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/detector.go
@@ -64,7 +64,15 @@ func UtilizationDetectorFactory(
 	if err != nil {
 		return nil, err
 	}
-	return NewDetector(name, *cfg, log.FromContext(handle.Context())), nil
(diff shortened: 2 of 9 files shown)


ahg-g (commented):
I don't recommend expanding and adding yet another interface, instead we can put this in the handle within a perhaps metadata struct that the plugin can check and do whatever it sees fit (fail creation, warn etc.).

pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/detector.go
```diff
@@ -64,7 +64,14 @@ func UtilizationDetectorFactory(
 	if err != nil {
 		return nil, err
 	}
-	return NewDetector(name, *cfg, log.FromContext(handle.Context())), nil
+	logger := log.FromContext(handle.Context())
+	if refresh := handle.RefreshMetricsInterval(); refresh > cfg.MetricsStalenessThreshold {
+		logger.Info("Warning: refresh-metrics-interval exceeds metricsStalenessThreshold; "+
```
LukeAVanDrie, line 69: nit: This logger has no name scope. `NewDetector` adds `WithName(typedName)` after this call returns, so the warning does not say which detector it is about. #2514 asks for the detector and both values. Adding the name as a key covers it:


E.g.,
```
		logger.Info("Warning: refresh-metrics-interval exceeds metricsStalenessThreshold; "+
			"metrics will be stale between polls",
			"detector", name,
```

pkg/epp/framework/interface/plugin/handle.go
```diff
@@ -44,6 +45,11 @@ type Handle interface {
 	// SetCrossReplicaSyncer makes the configured cross-replica syncer available
 	// to plugins through the handle.
 	SetCrossReplicaSyncer(Plugin)
+
+	// RefreshMetricsInterval returns the data-layer polling cadence for endpoint
```
LukeAVanDrie: Non-blocking, but I would change it before merge. The value is the `--refresh-metrics-interval` flag after the floor clamp. That is the runtime's base tick. 

Endpoint metrics can arrive less often as the metrics data source has its own `interval` parameter and `Runtime.Configure` lowers the base tick to the smallest source interval and rounds each source's period up to a base-tick multiple.

A plugin reading this value at creation time sees neither adjustment. Suggested wording:

```
	// RefreshMetricsInterval returns the base polling tick configured by
	// --refresh-metrics-interval, after the minimum-interval clamp. Data sources
	// that set their own interval may poll less often than this. Zero when unset.
```

A follow-up issue for running the comparison where the effective per-source cadence is known (`Runtime.Configure`) would cover the case #2514 mentions in its last paragraph. I can file it if you prefer to keep this PR small.

LukeAVanDrie (dismissed):
LGTM with a couple nits.

@ahg-g You suggested a metadata struct on the handle. This PR adds a direct method which is the same shape as `Metrics()` and `SetCrossReplicaSyncer`. Your call on whether a struct is required now or this is sufficient

pkg/epp/framework/plugins/flowcontrol/saturationdetector/utilization/detector.go
varad-ahirwadkar: thanks! Updated to include the `detector` name in the log context.

pkg/epp/framework/interface/plugin/handle.go
varad-ahirwadkar: Thanks! Updated the docs as suggested.

varad-ahirwadkar: > A follow-up issue for running the comparison where the effective per-source cadence is known (Runtime.Configure) would cover the case https://github.com/llm-d/llm-d-router/issues/2514 mentions in its last paragraph. I can file it if you prefer to keep this PR small.

Agreed. I think it's best to keep this PR scoped.

varad-ahirwadkar:
Hi @LukeAVanDrie, I've addressed review comments and updated the PR. When you have a chance, could you please take another look? Thanks!

ahg-g:
@varad-ahirwadkar sorry we missed this one, pls rebase

varad-ahirwadkar:
Hi @ahg-g , 
No worries! I've rebased the PR and pushed the updated changes.

PR #2649, opened 2026-09-01
feat(flowcontrol): support scoped request queue TTLs

### LINKED ISSUE #2316: [Flow-control]: Support per-priority-band and per-request flow control queue TTL'

Flow control queue TTL can currently only be set globally, across all queues, via
`flowControl.defaultRequestTTL`:

```yaml
flowControl:
  maxBytes: "10Gi"
  maxRequests: "1k"
  defaultRequestTTL: "60s"
```

There should be two more levels at which the queue-wait TTL can be specified:

1. Per priority band — each entry in flowControl.priorityBands may set its own
defaultRequestTTL, alongside the existing maxBytes / maxRequests:
```yaml
flowControl:
  defaultRequestTTL: "300s"
  priorityBands:
    - priority: 10
      defaultRequestTTL: "5s"    # latency-sensitive band, shed quickly
    - priority: -10
      defaultRequestTTL: "300s"  # latency-insensitive band, tolerate long queue waits
```

2. Per request — an individual inference request may set its own queue TTL with a
request header:

```yaml
x-llm-d-inference-ttl: 2s
```

*Resolution rule:* the TTL applied to a request is the minimum of the values that are
set among (global defaultRequestTTL, priority-band defaultRequestTTL, request header).
Taking the lowest value keeps TTL a shedding bound: no level can extend a request's queue-wait
budget beyond a tighter bound set at another level, and a caller can only ask to be shed
sooner, never later.

**Why is this needed**:

Inference requests routed through the same llm-d-router endpoint can have very different
shedding requirements. A single global TTL forces one queue-wait budget on all traffic, so
either latency-sensitive requests wait too long before being shed, or latency-insensitive requests
are shed too aggressively.

Per-band TTL lets operators tie shedding behavior to priority (shed low-priority traffic fast,
let high-priority traffic wait), and per-request TTL lets callers with their own deadlines opt
into being shed sooner rather than served late. This makes flow control usable for
mixed-workload endpoints without requiring a separate endpoint per shedding class.

nt591:
/kind feature

Allow operators to select queue-wait bounds per priority band and callers to shorten that bound through the inference TTL header.

My problem: I want to be able to buffer latency-insensitive work longer than latency-sensitive work, so I'd like to use higher priority bands with lower TTLs to serve interactive traffic, with low priority / higher TTL options for more async workflows.

Side fix: commit `e0db7a7` addresses what looks like a potential regression in https://github.com/llm-d/llm-d-router/pull/2284, where I THINK a request can pass controller submission, get buffered, expire, but the processor doesn't double check expiry time. If that's by design, I can back out my change.

Fixes #2316

```release-note
Allow operators to configure a queue TTL per flow-control priority band and clients to request a shorter TTL with `x-llm-d-inference-ttl`. These TTLs apply while the pool has endpoints; `noEndpointRequestTTL` governs an empty pool.
```


pkg/epp/flowcontrol/controller/internal/processor.go
```diff
@@ -306,6 +306,13 @@ func (p *Processor) enqueue(item *FlowItem) {
 		return
 	}
 
+	// The active queue-wait budget includes time spent in the processor's enqueue buffer.
+	regime := p.regime.Load()
+	if isExpired(item, p.clock.Now(), regime, p.noEndpointRequestTTL) {
+		p.finalizeAndRecordDrop(item, expiryError(regime.empty))
+		return
+	}
```
ahg-g, line 314: was this a bug?

pkg/epp/flowcontrol/registry/connection.go
```diff
@@ -40,3 +42,18 @@ func (c *connection) GetDataPlane() contracts.FlowRegistryDataPlane {
 func (c *connection) FlowKey() flowcontrol.FlowKey {
 	return c.key
 }
+
+// DefaultRequestTTL returns the queue-wait bound configured for the leased priority band and whether it was set.
+func (c *connection) DefaultRequestTTL() (time.Duration, bool) {
+	c.registry.mu.RLock()
+	defer c.registry.mu.RUnlock()
```
ahg-g: accessing the lock of another entity here is error prone, I think we should put the part that requires locking in a function implemented by the registry, and we call the func here.

pkg/epp/flowcontrol/controller/internal/processor.go
ahg-g: sorry, you mention this already in the issue description.

nt591: yes, I understand the code it's a latent bug that probably only matters under very tight TTLs

pkg/epp/flowcontrol/registry/connection.go
nt591: addressed in [763f151](https://github.com/llm-d/llm-d-router/pull/2649/commits/763f15149b745ed44b55833bbc69376e8bcfa4d2), thanks

apix/config/v1alpha1/endpointpickerconfig_types.go
```diff
@@ -564,6 +564,12 @@ type PriorityBandConfig struct {
 	// effectively remove the bound, set an explicit large value.
 	MaxRequests *resource.Quantity `json:"maxRequests,omitempty"`
 
+	// +optional
+	// DefaultRequestTTL bounds how long a request may wait in this priority band before it is evicted.
```
LukeAVanDrie: This reads as the whole queue-wait bound, but it only replaces `DefaultRequestTTL`, which applies while the pool has endpoints. `NoEndpointRequestTTL` stays global and still governs the empty-pool regime for every band, and `"0s"` here does not make waiting unbounded while the pool is empty. Suggest mirroring the scope language the global field uses:

```suggestion
	// DefaultRequestTTL replaces the global DefaultRequestTTL for this priority band: the queue-wait bound
	// while the candidate pool has endpoints. NoEndpointRequestTTL is not band-scoped and still governs
	// queue wait while the pool is empty. If omitted, the global DefaultRequestTTL is used; "0s" disables
	// eviction in this band while the pool has endpoints.
```

pkg/epp/flowcontrol/README.md
```diff
@@ -64,7 +64,9 @@ Tuning knobs, all under the `flowControl:` config section:
   behavior for sheddable traffic).
 * `defaultRequestTTL` — the queue-wait budget against a pool that has endpoints, and the other way a
   request is shed. Keep it under the client or gateway deadline, and size it to the time-to-first-token
-  budget you are willing to spend waiting on a saturated pool.
+  budget you are willing to spend waiting on a saturated pool. Priority-band entries and templates
+  may replace the global value, including with `0s` for unbounded queue wait. Clients may shorten the
```
LukeAVanDrie: Same scope point for the operator-facing doc. The header shortens the saturation budget only; a client asking for `2s` during a scale-from-zero still waits up to `noEndpointRequestTTL`. Worth a sentence so nobody reads the header as a hard queue deadline.

E.g., 

> Clients may shorten the selected operator bound with `x-llm-d-inference-ttl` using Go duration syntax. Both scopes narrow the saturation budget only; `noEndpointRequestTTL` stays global and is not shortened by the header.

pkg/epp/flowcontrol/controller/controller.go
```diff
@@ -502,6 +514,11 @@ func (fc *FlowController) distributeRequest(
 	item *internal.FlowItem,
 ) error {
 	reqID := item.OriginalRequest().ID()
+	select {
+	case <-ctx.Done():
```
LukeAVanDrie, line 519: nit (non-blocking): The reason this early exit exists is not obvious from the code: `Submit` is non-blocking and would otherwise hand an already-expired item to the processor.

```
	// Submit does not observe ctx, so an item whose budget expired during lease acquisition must be
	// rejected here rather than handed to the processor.
	select {
	case <-ctx.Done():
...
```

pkg/epp/requestcontrol/admission.go
```diff
@@ -163,15 +165,24 @@ func (fcac *FlowControlAdmissionController) Admit(
 	logger.V(logutil.TRACE).Info("Executing FlowControlAdmissionController",
 		"requestID", reqCtx.SchedulingRequest.RequestID, "priority", priority, "fairnessID", reqCtx.SchedulingRequest.FairnessID)
 
+	initialEffectiveTTL := time.Duration(0)
+	if rawTTL, ok := metadata.GetLowerCaseHeaderValue(reqCtx.Request.Headers, metadata.InferenceTTLHeaderKey); ok {
+		parsedTTL, err := time.ParseDuration(strings.TrimSpace(rawTTL))
+		if err == nil && parsedTTL > 0 {
```
LukeAVanDrie, line 171: nit (non-blocking): A malformed or non-positive value is dropped with no signal, so an operator debugging why their TTL header did nothing has nowhere to look. The predicted-latency plugin logs its unparseable SLO headers at DEBUG and continues; same pattern would fit here:

```
		if err == nil && parsedTTL > 0 {
			initialEffectiveTTL = parsedTTL
		} else {
			logger.V(logutil.DEBUG).Info("Ignoring invalid request TTL header",
				"requestID", reqCtx.SchedulingRequest.RequestID, "value", rawTTL, "err", err)
		}
```

LukeAVanDrie (commented):
Thanks for driving this!

Both new scopes only narrow the saturation-regime budget. `isExpired` switches to the global `noEndpointRequestTTL` whenever the pool is empty, and the context backstop is `max(saturationTTL, noEndpointRequestTTL)`, so a request carrying `x-llm-d-inference-ttl: 2s` during a scale-from-zero still waits up to `noEndpointRequestTTL`. 

That is consistent for the band value, since it overrides `defaultRequestTTL` which has the same scope. For the request header it is less obvious: the header expresses caller intent, which does not depend on the regime. I do not think this PR needs to change behavior, but the apix field doc and README should say what the scope is.

Also, the release notes only mentions the band-level `defaultRequestTTL`. The `x-llm-d-inference-ttl` header is user-facing too, so let's include it there.

pkg/epp/flowcontrol/controller/internal/processor.go
LukeAVanDrie: +1; thanks for closing this!

LukeAVanDrie (dismissed):
Approved with some small doc / logging nits; thanks!

LukeAVanDrie (approved):

ahg-g (approved):

PR #2651, opened 2026-09-01
docs(epp): add affinity filter composition requirements

LukeAVanDrie:
**What type of PR is this?**

/kind documentation

Adds a composition-requirements section to the `prefix-cache-affinity-filter` README.

The TTFT load gate's break keeps all endpoints; whether a request moves off the sticky set is decided by the scorers and picker that run after the filter. Scorer outputs are clamped to [0, 1] before weighting (`pkg/epp/scheduling/scheduler_profile.go`), which bounds how far in prefix cache score a break can move a request by the ratio of the load scorers' weight sum to the prefix scorer's weight. A prefix weight at or above the load weight sum makes the gate unable to move a request off a fully matched sticky endpoint; at or above the load weight sum divided by `affinityThreshold` it cannot move any request. No documentation stated this requirement. The inert composition is reachable by adding the filter to a weighted-blend profile (weighted `prefix-cache-scorer` + load scorers + `max-score-picker`), and the filter does not require a prefix scorer in the profile (the default `approx-prefix-cache` producer supplies `PrefixCacheMatchInfo`), so nothing structural prevents it.

The new section states the load-only requirement on downstream scorers with the weight arithmetic, names the inert composition, lists the recommended compositions (`token-load-scorer` prefill-bound, `active-request-scorer` decode-bound, with `max-score-picker`), notes that the picker choice is orthogonal to the scorer requirement, and links the optimized-baseline guide and the Sticky Until Saturated post. Also corrects a stale picker name in the same file (`max-picker` to `max-score-picker`).

Possible follow-up outside this repo: the llm-d guides that ship this filter (optimized-baseline and related) state the recommended composition but not the constraint; a one-line note there would cover users editing guide values files directly.

None.

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/README.md
+++ b/pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/README.md
@@ -72,7 +72,8 @@ Can be instantiated multiple times with different thresholds (e.g., 0.99 for glo
 The filter narrows the candidate set or leaves it unchanged; the routing decision is made
 by the scorers and picker that run after it. When the TTFT load gate breaks stickiness, the
 filter keeps all endpoints, and the request moves off the sticky set only if downstream
-scoring prefers the less-loaded endpoints.
+scoring prefers the less-loaded endpoints. Nothing tells the scorers that a break happened;
+they apply the same weights to whatever candidates arrive.
 
 Scorers running after this filter must not include a separately weighted prefix-affinity
 scorer. Each scorer's output is clamped to [0, 1] before weighting, so the other scorers can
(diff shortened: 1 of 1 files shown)


pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/README.md
```diff
@@ -67,6 +67,33 @@ Can be instantiated multiple times with different thresholds (e.g., 0.99 for glo
   the TTFT load gate is skipped. If no endpoints have `PrefixCacheMatchInfo`, all prefix
   scores default to 0 and no endpoints pass the affinity threshold, so all are kept (no-op)
 
+## Composition requirements
+
+The filter narrows or widens the candidate set; the routing decision is made by the scorers
```
kaushikmitr: does it ever widens?

pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/README.md
```diff
@@ -67,6 +67,33 @@ Can be instantiated multiple times with different thresholds (e.g., 0.99 for glo
...
+and picker that run after it. When the TTFT load gate breaks stickiness, the filter keeps
+all endpoints, and the request moves off the sticky set only if downstream scoring prefers
+the less-loaded endpoints.
+
+Scorers running after this filter must score on load only. Each scorer's output is clamped
```
kaushikmitr: I dont think it is correct to say scorers running after this filter must score on load only. Example if the stickiness is broken then we want to still keep some amount of prefix affinity. In the token aware scorers the way its done is the score is calculated based on both current uncached tokens in flight + additinal uncached token from this request (note that uncached essentially means that we substrat the tokens already cached so affinity is still somewhat considered)

LukeAVanDrie: Agreed, "score on load only" is overstated. Reworded to "must not include a separately weighted prefix-affinity scorer" and added a sentence on the uncached-token path.

pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/README.md
LukeAVanDrie: No. The sticky branch returns the sticky subset; the break and no-op branches return the input set unchanged, so the output is never a superset of the input. Reworded to "narrows the candidate set or leaves it unchanged".

ahg-g (approved):

PR #2652, opened 2026-09-01
fix(epp): Add check for unpopulated load attributes for TTFT load gate

### LINKED ISSUE #2650: prefix-cache-affinity-filter: TTFT load gate compares observed values against substituted defaults when an endpoint lacks its load attribute
**What happened**:

The TTFT load gate keeps all endpoints when the best sticky endpoint's estimated TTFT exceeds the best non-sticky endpoint's by more than `maxTTFTPenaltyMs` (`pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go:205-215`). When an endpoint lacks the attribute backing the configured `ttftSource`, `endpointTTFT` substitutes a default: `math.MaxFloat64` on the `latencyPredictor` path, 0 ms on the `prefillThroughput` path (`plugin.go:266-286`). A gate decision can then compare an observed estimate against a substituted one. The two paths fail in opposite directions:

| `ttftSource` | missing signal on | substituted TTFT | gate effect |
|---|---|---|---|
| `latencyPredictor` | every sticky endpoint | `MaxFloat64` | fires against any non-sticky endpoint with a prediction; affinity lost |
| `latencyPredictor` | every non-sticky endpoint | `MaxFloat64` | never fires; stickiness holds under arbitrary overload |
| `prefillThroughput` | any sticky endpoint | 0 ms | best sticky TTFT pinned to 0; never fires |
| `prefillThroughput` | any non-sticky endpoint | 0 ms | best non-sticky TTFT pinned to 0; fires once every sticky endpoint's estimate exceeds the penalty |

The `latencyPredictor` path is reachable in production. `LatencyPredictionInfo` is written per request in `Produce`, only when `PredictInProduce` is enabled and prediction succeeds for the full candidate set (`pkg/epp/framework/plugins/requestcontrol/dataproducer/predictedlatency/dataproducer_hooks.go:82-108`); values persist on endpoint objects across requests.

- An endpoint created or recreated after the last successful prediction carries no value and contributes `MaxFloat64`.
- During a predictor outage, established endpoints keep stale predictions while new endpoints have none, so mixed comparisons persist for the length of the outage.
- With `PredictInProduce: false`, no endpoint ever carries the attribute and `maxTTFTPenaltyMs` is inert. Init validation passes: `Consumes()` marks the attribute required, the framework verifies at init that a producer exists for the key (`pkg/epp/framework/interface/plugin/plugins.go:42`), and the predictedlatency producer declares the key in `Produces()` unconditionally. No check covers whether endpoints carry values at decision time.

The `prefillThroughput` substitution has no known production trigger. The in-flight-load producer injects `InFlightLoad` as a dynamic attribute on the endpoint AddOrUpdate event (`pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go:364-374`), which `Runtime.NewEndpoint` dispatches synchronously during endpoint construction (`pkg/epp/datalayer/runtime.go:431,450`), so a schedulable endpoint carries the attribute and reads live counters. The 0-token fallback (`plugin.go:279-286`) is reachable only if injection fails. The fix covers it for consistency; no production reproduction is claimed for it.

The gate decides from observed values. When either side of the comparison has no observed value, the filter keeps the sticky set, matching the behavior when the gate is disabled, and the skip is observable.

Unit level, predictor path: `ttftSource: latencyPredictor`, `affinityThreshold: 0.80`, `maxTTFTPenaltyMs: 100`. Endpoint A: sticky, no `LatencyPredictionInfo`. Endpoint B: non-sticky, predicted TTFT 50 ms. `Filter()` returns all endpoints and records `load_override` on a penalty of `MaxFloat64` minus 50 ms.

Unit level, throughput path: `ttftSource: prefillThroughput`, `peakPrefillThroughput: 1000`, `maxTTFTPenaltyMs: 100`. Endpoint A: sticky, `InFlightLoad.Tokens = 500` (500 ms estimate). Endpoint B: non-sticky, attribute absent (0 ms). `Filter()` returns all endpoints on a 500 ms penalty computed against an endpoint that reported nothing.

Proposed fix: exclude endpoints lacking the configured attribute from both minimum computations; when either side ends up with no observed value, keep the sticky set and record a `missing_signal` outcome in `prefix_cache_affinity_filter_decisions_total`, which labels decisions by outcome (`metrics.go:48-55`). Exclusion keeps the gate live while a single endpoint lags; the skip applies only when a whole side is blind.

```mermaid
flowchart TD
    S["split candidates into sticky / non-sticky"] --> O{"observed TTFT on both sides?"}
    O -- yes --> G{"bestSticky - bestNonSticky > maxTTFTPenaltyMs?"}
    G -- yes --> ALL["keep all (load_override)"]
    G -- no --> STICKY["keep sticky (sticky)"]
    O -- no --> SKIP["keep sticky (missing_signal)"]
```

Missing data fails toward stickiness: an overloaded sticky endpoint with a data gap keeps traffic until its signal returns. The `missing_signal` count makes that state visible to operators. Under `PredictInProduce: false` the gate skips every decision and the metric shows it.

Precedent: the sgl-model-gateway power-of-two policy compares token loads only when both candidates report them, and otherwise falls back to local request counts for both (`sgl-model-gateway/src/policies/power_of_two.rs`; regression test `test_reproduce_incompatible_metric_bug`). Degrading both sides to a shared coarser signal requires a fallback available on both paths, which the filter lacks; skip-and-count stands in until one exists.

The README behavior list documents the case where no endpoint has the attribute (gate skipped) but not the partial case; the fix should state both.

**Environment**:
- Kubernetes version (use `kubectl version`): n/a (logic-level, environment-independent)
- llm-d-router version: main at c99864c5
- Install tools: n/a
- Others: n/a

d0w:
**What type of PR is this?**

/kind bug

Previously, with the prefix-cache-affinity-filter, the load gate compared loaded values against defaults that would cause the paths to fail in opposite directions. When endpoints (whether sticky or not) were missing certain attributes such as `LatencyPredictionInfo` or `InFlightLoad.Tokens` the substituted value would be arbitrarily 0 or `MaxFloat64` causing the gate to unwillingly fire or not fire.

The implementation changes the TTFT and inFlightToken computation functions to return booleans to determine if the computation is ok or not. If not ok (meaning there was no data), the endpoint is excluded from the bestTTFT calculation. If there was no data from either sticky/non-sticky set, then default to the sticky set. When there is missing data causing a failure toward stickiness, a `missing_signal` count is incremented. 

Tests were added as well as an adjustment to the README to note this behavior.

Fixes https://github.com/llm-d/llm-d-router/issues/2650

```release-note
NONE
```

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
+++ b/pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
@@ -300,8 +300,7 @@ func (p *Plugin) bestTTFT(endpoints []fwksched.Endpoint) (float64, bool) {
 // endpointTTFT returns the predicted TTFT (ms) for an endpoint, either from the
 // latency predictor or estimated from in-flight tokens and peak prefill
 // throughput. Endpoints missing the required attribute contribute no signal:
-// MaxFloat64 on the predictor path (never the fastest), 0 in-flight tokens on
-// the throughput path (no observed load). Returns the ttft and false if predictiond data was missing.
+// the function returns false and the endpoint is excluded from the bestTTFT calculation.
 func (p *Plugin) endpointTTFT(ep fwksched.Endpoint) (float64, bool) {
 	if p.config.usesLatencyPredictor() {
 		if raw, ok := ep.Get(p.latencyPredictionInfoDataKey); ok {
@@ -317,8 +316,8 @@ func (p *Plugin) endpointTTFT(ep fwksched.Endpoint) (float64, bool) {
 	return float64(tokens) / p.config.PeakPrefillThroughput * 1000, true
 }
 
-// inFlightTokens returns an endpoint's in-flight token count, or 0 when the
-// attribute is absent (no observed load).
+// inFlightTokens returns an endpoint's in-flight token count. If the attribute is
+// absent, it returns ok=false so the caller can treat the signal as missing.
 func (p *Plugin) inFlightTokens(ep fwksched.Endpoint) (int64, bool) {
 	if raw, ok := ep.Get(p.inFlightLoadDataKey); ok {
 		if load, ok := raw.(*attrconcurrency.InFlightLoad); ok && load != nil {
--- a/pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
+++ b/pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
@@ -230,8 +230,14 @@ func (p *Plugin) Filter(ctx context.Context, request *fwksched.InferenceRequest,
 
 	// TTFT load gate: break stickiness if sticky endpoints are too slow.
 	if p.config.MaxTTFTPenaltyMs > 0 && len(nonSticky) > 0 {
-		bestStickyTTFT := p.bestTTFT(sticky)
-		bestNonStickyTTFT := p.bestTTFT(nonSticky)
+		bestStickyTTFT, stickyOk := p.bestTTFT(sticky)
+		bestNonStickyTTFT, nonStickyOk := p.bestTTFT(nonSticky)
+		if !stickyOk || !nonStickyOk {
+			logger.V(logutil.DEBUG).Info("PrefixCacheAffinityFilter: TTFT load gate skipped, missing signal",
+				"stickyObserved", stickyOk, "nonStickyObserved", nonStickyOk)
+			recordDecision(p.typedName.Name, outcomeMissingSignal)
+			return sticky
+		}
 		penalty := bestStickyTTFT - bestNonStickyTTFT
 		span.SetAttributes(semconv.LLMDEPPFilterTTFTPenaltyMs(penalty))
 		if penalty > p.config.MaxTTFTPenaltyMs {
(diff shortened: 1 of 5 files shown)


d0w:
@ahg-g Hello, sorry to ping but am just wondering on if this PR is good to be merged? Thanks.

pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
```diff
@@ -230,8 +230,14 @@ func (p *Plugin) Filter(ctx context.Context, request *fwksched.InferenceRequest,
...
 	if p.config.MaxTTFTPenaltyMs > 0 && len(nonSticky) > 0 {
-		bestStickyTTFT := p.bestTTFT(sticky)
-		bestNonStickyTTFT := p.bestTTFT(nonSticky)
+		bestStickyTTFT, stickyOk := p.bestTTFT(sticky)
+		bestNonStickyTTFT, nonStickyOk := p.bestTTFT(nonSticky)
+		if !stickyOk || !nonStickyOk {
+			logger.V(logutil.DEBUG).Info("PrefixCacheAffinityFilter: TTFT load gate skipped, missing signal",
+				"stickyObserved", stickyOk, "nonStickyObserved", nonStickyOk)
+			recordDecision(p.typedName.Name, outcomeMissingSignal)
+			return sticky
```
ahg-g, line 239: should we just return all endpoint here?

d0w: The original issue (https://github.com/llm-d/llm-d-router/issues/2650) notes
> Missing data fails toward stickiness: an overloaded sticky endpoint with a data gap keeps traffic until its signal returns. The missing_signal count makes that state visible to operators. Under PredictInProduce: false the gate skips every decision and the metric shows it.

In my opinion, it should return sticky since the case this fix solves involves a lack of telemetry data (and therefore no way to make a comparison) and not that sticky endpoints are slower. The addition of the `missing_signal` outcome should also make the case more distinct. Let me know what you think.

kaushikmitr: i am not sure if missing_signal should return sticky. I think missing_signal means something is not setup right, and we should just skip the filter.

pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
```diff
@@ -230,8 +230,14 @@ func (p *Plugin) Filter(ctx context.Context, request *fwksched.InferenceRequest,
...
 	// TTFT load gate: break stickiness if sticky endpoints are too slow.
 	if p.config.MaxTTFTPenaltyMs > 0 && len(nonSticky) > 0 {
-		bestStickyTTFT := p.bestTTFT(sticky)
-		bestNonStickyTTFT := p.bestTTFT(nonSticky)
+		bestStickyTTFT, stickyOk := p.bestTTFT(sticky)
+		bestNonStickyTTFT, nonStickyOk := p.bestTTFT(nonSticky)
+		if !stickyOk || !nonStickyOk {
+			logger.V(logutil.DEBUG).Info("PrefixCacheAffinityFilter: TTFT load gate skipped, missing signal",
+				"stickyObserved", stickyOk, "nonStickyObserved", nonStickyOk)
+			recordDecision(p.typedName.Name, outcomeMissingSignal)
```
kaushikmitr, line 238: other cases does span.SetAttributes. That is missign here. We should add here as well

d0w: thanks for catching that, added in my latest changes

pkg/epp/framework/plugins/scheduling/filter/prefixcacheaffinity/plugin.go
d0w: got it. it could definitely go either way but I leaned towards the ticket's request. I changed the filter case to return all endpoints instead of just sticky ones

Issue #790, opened 2026-03-31
Makefile/script environment (e.g., version) defaults are scattered and diverge


elevran:
The repo has one logical set of project-level configurations used in make and scripts: image names, registries, default tags, etc. These are scattered and there is no single canonical file that "owns" them. The same defaults are declared independently in `Makefile`, `scripts/pull_images.sh`, `scripts/kubernetes-dev-env.sh`, and `test/e2e/e2e_suite_test.go` and they have drifted out of sync. 
In addition, the use of `dev` (or `latest`) image tags resulted in wasted effort in local and CICD testing failures.

Some examples:
- pull_images.sh defaults VLLM_SIMULATOR_TAG to latest, so make image-pull pulls a different image than make env-dev-kind
- kubernetes-dev-env.sh defaults EPP_TAG to v0.1.0; a developer who forgets to export it silently deploys a stale release
- pull_images.sh hardcodes ghcr.io/llm-d in image refs, so setting IMAGE_REGISTRY has no effect on that script
- DEVELOPMENT.md documents KIND_PROM_HOST_PORT=30091 make env-dev-kind, but Makefile.kind.mk never forwards that variable to the script and the flag is silently discarded.
- Makefile.cluster.mk passes $$VERSION to four envsubst calls; VERSION is never defined anywhere, so Kubernetes manifests get silent empty substitution
- Two bare envsubst calls (no variable list) risk substituting any matching shell variable ($HOME, $PATH, etc.) into YAML manifests

The goal is to consolidate all configurations to single Makefile that owns all variables such as tag and image defaults, with scripts and tests deriving their values from there rather than each carrying private copies. Scripts should also print the variables they use so it is easy to catch discrepencies.

See also #774

PR #2654, opened 2026-09-02
fix(kind): forward Prometheus host port

Alex-ai-future:
**What type of PR is this?**

/kind bug

`DEVELOPMENT.md` documents `PROM_ENABLED=true KIND_PROM_HOST_PORT=30091 make env-dev-kind`, but the `env-dev-kind` target did not forward `KIND_PROM_HOST_PORT` to `scripts/kind-dev-env.sh`. The script therefore used its default host port, `30090`.

This change forwards `KIND_PROM_HOST_PORT` as `PROM_HOST_PORT`. The Prometheus service NodePort remains `30090`, and direct invocations using `PROM_HOST_PORT` remain supported.

This is a development-only fix. It does not change production deployments, request routing, or the in-cluster Prometheus port. It lets local Kind users select an available host port when `30090` is already in use, while preserving the existing default.

Related to #790

**Local before/after verification**:

Command:

```bash
PROM_ENABLED=true KIND_PROM_HOST_PORT=30091 make --no-print-directory -n env-dev-kind
```

Before this change, the generated command omitted `PROM_HOST_PORT`:

```text
CLUSTER_NAME=llm-d-router-dev \
GATEWAY_HOST_PORT=30080 \
IMAGE_REGISTRY=ghcr.io/llm-d \
./scripts/kind-dev-env.sh; \
```

After this change, the generated command forwards the requested port:

```text
CLUSTER_NAME=llm-d-router-dev \
GATEWAY_HOST_PORT=30080 \
PROM_HOST_PORT="30091" \
IMAGE_REGISTRY=ghcr.io/llm-d \
./scripts/kind-dev-env.sh; \
```

No Kind cluster was created during the comparison.

```release-note
Fix the Prometheus host-port override for the Kind development environment.
```

### DIFF
--- a/Makefile.kind.mk
+++ b/Makefile.kind.mk
@@ -2,6 +2,9 @@
 
 KIND_CLUSTER_NAME ?= $(PROJECT_NAME)-dev
 KIND_GATEWAY_HOST_PORT ?= 30080
+# KIND_PROM_HOST_PORT is the documented Make variable. Fall back to
+# PROM_HOST_PORT so direct callers keep working through this target.
+KIND_PROM_HOST_PORT ?= $(PROM_HOST_PORT)
 
 .PHONY: image-kind
 image-kind: ## Reload the EPP and SideCar images into the kind cluster $(KIND_CLUSTER_NAME)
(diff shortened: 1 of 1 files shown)


Makefile.kind.mk
```diff
@@ -2,6 +2,9 @@
 
 KIND_CLUSTER_NAME ?= $(PROJECT_NAME)-dev
 KIND_GATEWAY_HOST_PORT ?= 30080
+# KIND_PROM_HOST_PORT is the documented Make variable. Fall back to
+# PROM_HOST_PORT so direct callers keep working through this target.
+KIND_PROM_HOST_PORT ?= $(PROM_HOST_PORT)
```
ahg-g: this is a bit confusing, we default KIND_PROM_HOST_PORT to PROM_HOST_PORT then set the value back to PROM_HOST_PORT below. What is the actual default value?

Alex-ai-future: it's right, updated~

ahg-g (approved):

PR #2657, opened 2026-09-02
feat(chart): support vLLM Rust renderer flavor for tokenizer sidecar

### LINKED ISSUE #2440: Support vLLM Rust Renderer
It was 10x faster on codex traces and enables 40x lighter docker image, based on the prototype used in the proposal.

ref:
https://github.com/vllm-project/vllm/issues/49047

Madhumasa84:
**What type of PR is this?**
    /kind feature

  Adds `router.tokenizer.flavor` to configure the tokenizer sidecar backend in the Helm charts.

    - Set `router.tokenizer.flavor: "python"` (default) to run `vllm launch render`.
    - Set `router.tokenizer.flavor: "rust"` to run the standalone `vllm-rs render` binary (~40x lighter image and 10x–14x
  higher preprocessing throughput).

 Key changes:
    - `_deployment.yaml`: Sets default container command to `["vllm-rs", "render"]` when `flavor` is `"rust"`, while
  retaining `["vllm", "launch", "render"]` for `"python"`. User-defined `command` overrides still take precedence.
    - `_helpers.tpl`: Adds validation enforcing that `flavor` must be either `"python"` or `"rust"`.
    - `config/charts/README.md` & `pkg/.../tokenizer/README.md`: Documents the new `flavor` setting and deployment
  configurations.
    - `hack/verify-helm.sh`: Adds positive verification test cases for `tokenizer-python` and `tokenizer-rust`, along
  with negative tests rejecting invalid flavors.
    - `vllm_http_test.go`: Adds `TestVLLMHTTPRenderer_RustStandaloneRenderer` verifying wire-format compatibility with
  the Rust renderer response shape.

  Fixes #2440

    Support vLLM Rust renderer flavor (`router.tokenizer.flavor: "rust"`) in Helm charts for tokenizer sidecars.

### DIFF
--- a/config/charts/routerlib/values.yaml
+++ b/config/charts/routerlib/values.yaml
@@ -95,12 +95,16 @@ epp:
 proxy:
   enabled: false    
 
-# Tokenizer sidecar. Runs vLLM's `vllm launch render <modelName>` and exposes
-# `/render` over loopback HTTP to EPP. modelName is passed as the first
+# Tokenizer sidecar. Runs vLLM's `vllm launch render <modelName>` (Python) or
+# `vllm-rs render <modelName>` (Rust) and exposes `/v1/completions/render` and
+# `/v1/chat/completions/render` over loopback HTTP to EPP. modelName is passed as the first
 # positional arg to the render command (required when enabled). EPP is wired to
 # the sidecar via `router.epp.pluginsCustomConfig`.
+# When flavor is "rust", router.tokenizer.image must be set to an image containing the vllm-rs binary.
 tokenizer:
   enabled: false
+  # Backend renderer flavor: "python" (default, runs vllm launch render) or "rust" (runs vllm-rs render; requires an image that includes vllm-rs).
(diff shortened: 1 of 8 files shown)


config/charts/routerlib/values.yaml
zdtsw, line 112: how the new image is going to be used here when flavor is set to "rust"

Madhumasa84: The chart change controls the container command only. When flavor rust is set, users must also override router.tokenizer.image.repository (and tag) to an image that ships the vllm-rs binary, since vllm/vllm-openai-cpu does not include it. I can add a comment in values.yaml to make this explicit. The appropriate image reference will come from vllm-project/vllm#51264 once the standalone Rust image is published. 
Happy to update this in a follow up commit.

sagearc (commented):
@zdtsw wdyt? should we wait for official vllm rust image?

ahg-g (approved):

PR #2659, opened 2026-09-02
coordinator: extend audio/video media support alongside images

### LINKED ISSUE #2648: [Coordinator] Extend audio/video media support alongside images
## Summary

The coordinator is image-only today: `replace_media_urls` drops
non-image parts, `encode` hardcodes `ModalityImage`, `MultimodalEntry` has no
modality field, and downstream code assumes image by convention. EPP already
supports audio/video end-to-end (tokenizer, OpenAI parser, disagg profile
handler, sidecar EC connectors) — the gap is coordinator-side only.

This is a tracking issue for extending the coordinator to fully support audio
and video media alongside images. 

(Anthropic-shape support is explicitly out of scope)

Related: #2640 (sidecar and e2e test coverage for audio and video multimodal
requests) — that issue covers test-side gaps; this issue covers the
coordinator-side production changes that would let those requests actually
flow through.

## Baseline (why the gap exists today)

- When a request arrives with audio or video media, the coordinator silently
  drops those parts before forwarding the request downstream — only images
  make it through.
- The internal record that tracks each media item has no field for its
  modality, so once a piece of media enters the pipeline, downstream steps
  have no way to tell an image from audio or video.
- Every pipeline stage that handles media — ingestion, encoding, rendering,
  and decoding — is hardcoded to the image case. There is no per-modality
  branching anywhere in the coordinator.

### 1. Type foundation (prerequisite)
Add a modality field to the multimodal-entry record, and add audio and video
peer constants next to the existing image constant. Nothing else can land
until these exist.

### 2. Request ingestion
Today the media-download step accepts only image URLs and rejects everything
else. Widen it to accept audio URLs, video URLs, and inline audio payloads —
each with its own MIME allowlist and its own size cap. Audio and video files
are much larger than images, so a single global limit no longer fits.

### 3. Encode, render, and decode pipeline
Every stage that reads or writes multimodal features is keyed on "image"
today. Generalize each of them to be keyed by modality. The render-response
length check assumes a single flat list of images — it needs per-modality
accounting once entries can span multiple modalities.

### 4. Config
Raise or split the request-body cap (64 MB is fine for images, not for video).
Add per-modality download caps and per-modality MIME allowlists. Update the
config docs that currently talk only about images.

### 5. Unchanged
Metrics, SSRF and address guards, and the gateway proxy path are already
modality-opaque and need no changes.

## Explicitly out of scope

- Anthropic-format audio and video (Anthropic's API doesn't ship those block types).
- End-to-end checks of what the backend actually decoded — see #2640.
- A separate latent question about the image-download Content-Type check; worth resolving but tracked elsewhere.

dmitripikus:
**What type of PR is this?**

/kind feature

## Summary

Adds full audio and video support to the coordinator alongside images. Covers every part of the tracking issue: type foundation, request handling, pipeline plumbing, per-kind config, and encoder path.

What each part of the branch does:

- **Media kind on each item.** Every downloaded media item now carries a label saying what kind it is (image, audio, or video). New names for audio and video sit next to the existing name for image so the rest of the coordinator can tell them apart.

- **Downloading audio and video.** The step that downloads media links used to accept only image links and drop the rest. It now also accepts audio links, video links, and audio sent directly inside the request. Each kind has its own list of allowed file types and its own size limit.

- **Pipeline stages handle every kind.** The steps that pass media between stages of the pipeline no longer assume everything is an image. They now handle audio and video too. Requests that only carry images behave exactly as before — every existing test still passes.

- **Per-kind operator settings.** New settings let an operator set a different download size limit for each kind and a different list of allowed file types for each kind. Images can stay tight while audio and video are allowed to be much larger. Bad settings (zero, negative, or too large) are rejected at startup instead of being silently ignored.

- **Sending audio and video to the encoder.** Audio and video now go through the encoder the same way images do. Each item in the request is paired with the matching content when the encoder is called.

## Design decisions

- **Encoder wire contract**: coordinator sends the encoder an OpenAI chat-completions envelope with the media part in its native shape (`{type: "audio_url", audio_url: {...}}`, `{type: "input_audio", input_audio: {data, format}}`, etc.); features under `mm_hashes[<modality>]`. Encoder is expected to return the same shape with new modality keys.
- **Render service response**: coordinator reads `Features.MM*[<modality>]` per-modality; mismatched responses fail loudly with clear per-modality error messages, never silently.
- **Walker-order pairing**: `MultimodalEntry` is appended in the same request-walker order the encode and decode steps use, so entry `i` of a modality always pairs with the `i`-th content part of that modality. Image-only requests behave identically to the pre-PR indexing; every existing test passes unchanged.
- **Security note**: the PR widens the coordinator's attack surface. Each concern and how the PR handles it:
  - **Wider codec CVE surface.** Audio and video decoders (ffmpeg, libavcodec, and their bindings) have a longer security-fix history than image decoders. The coordinator itself does not decode media — it only downloads, size-checks, and forwards — so decoder exposure lives in the model server. For audio and video, the coordinator also checks the origin's `Content-Type` against the per-modality MIME allowlist on HTTP downloads (not only on `data:` URIs). One normalizer reduces a header or `data:` media type to its bare lowercase type before both the allowlist lookup and the `data:` URI the step writes, cutting at the first `;` (MIME parameters like `;codecs=...`) or `,` (a header holding more than one value). Cutting at the comma is what keeps the emitted URI parseable, since RFC 2397 ends the metadata at the first comma. Operators can lock the backend down to a minimal codec set (e.g. `audio/wav` only) when warranted.
  - **Denial of service through large requests.** Audio and video files are much bigger than images. The per-modality download caps default unset, falling back to the tight 10 MB global limit — audio and video are as constrained as images out of the box. `max_concurrent_downloads` bounds peak memory the same way for all media. An operator who wants to accept 200 MB videos must raise the video cap explicitly, so the peak-memory implications are a conscious choice.
  - **Denial of service through many media parts.** Before this PR a body listing hundreds of `audio_url` or `video_url` endpoints downloaded nothing, because non-`image_url` parts were skipped. Each one is now fetched, so a few KB of request body can pull far more than one per-item cap allows. `max_multimodal_entries` is the gate that bounds the sum, and the shipped `config/coordinator/coordinator.yaml` sets it to 8. The code default stays unlimited, so an existing config that omits the key is unchanged; operators running their own config should set it.
  - **Long downloads under load.** The download code has been in production for 10 MB image downloads. Larger downloads (100+ MB videos) push more bytes through the same SSRF guard, HTTP client, and timeouts. If caps are raised, timeout should be updated.
  - **Documentation.** YAML comments next to the audio and video size caps represent the guidance so an operator sees it right where they'd tune the setting.

## Config changes

New YAML params in the `replace-media-urls` step (all optional):

- `max_image_download_size`, `max_audio_download_size`, `max_video_download_size` — per-modality overrides of the global `max_download_size`
- `allowed_image_content_types`, `allowed_audio_content_types`, `allowed_video_content_types` — override the built-in MIME allowlist per modality

One existing param changes value in the shipped config: `max_multimodal_entries` is set to 8 rather than left commented out, because turning `audio_url` and `video_url` into real downloads makes the entry count the only thing bounding what one request pulls into memory. The code default is untouched, so a config that does not name the key behaves as before.

`allowed_audio_content_types` interacts with `input_audio` through one canonical MIME per format (`mp3` is checked as `audio/mpeg`, not `audio/mp3`). The YAML comment next to the param lists the mapping, and a rejection names the format alongside the MIME it mapped to.

## Out of scope

- Audio and video blocks in Anthropic's format. Their API does not have them.
- End-to-end tests that check decoded audio or video against a real backend (see #2640).
- Checking the `Content-Type` that an image origin returns. Image downloads work as before: the origin's type is not compared against the allowlist. Setups that rely on the `application/octet-stream` fallback keep working. Audio and video downloads are checked. Only `image_url` stays permissive.
- **A limit on the total bytes one request may download.** Each item has its own cap, but nothing in code caps the sum, and every payload is kept in memory until the step finishes. So the real limit is `max_multimodal_entries` times the largest cap in use. The shipped config sets that count to 8, which bounds the product for anyone using it, and the `max_download_size` comments in `coordinator.yaml` tell operators to raise the count and a per-modality cap together. A byte-denominated total limit in code would be better than a count. That is for a later PR.
- **Encoding a download while it is being read.** Today the step reads the whole file and then encodes it, so one download in flight holds the raw bytes and the base64 text at the same time, about 2.4x the file size. This PR removed the copy that was kept for the whole request. Removing this one means rewriting the download loop, and `max_concurrent_downloads` already caps it, so it can wait.
- **Giving `max_multimodal_entries` a default in code.** The shipped config sets it, but the code default stays unlimited, so an existing config that omits the key accepts exactly what it accepted before. Changing the code default would change behavior for deployments that never edited their config, which belongs in its own PR.
- **An escape hatch for `input_audio.format`.** The format-to-MIME map is a closed set (`wav`, `mp3`, `flac`, `opus`, `ogg`, `webm`). A backend that decodes `aac`, `m4a`, or `pcm16` cannot reach it through `input_audio`, and no param extends the map. `audio_url` with a `data:` URI is the workaround.
- **Consolidating per-modality policy into one table.** The policy lives in several package-level maps plus two `modality != ModalityImage` predicates, so a fourth modality would inherit inline-size and download-`Content-Type` enforcement without anyone choosing it. One struct per modality would make an omission a compile error. Worth doing when a fourth modality arrives, not before.

Fixes #2648

**Release note**:

```release-note
coordinator: `replace-media-urls` now accepts `audio_url`, `video_url`, and inline `input_audio` content parts alongside `image_url`, and audio/video are routed through the same encoder pipeline as images. Adds new step params: `max_image_download_size` / `max_audio_download_size` / `max_video_download_size` (per-modality overrides of `max_download_size`) and `allowed_image_content_types` / `allowed_audio_content_types` / `allowed_video_content_types` (override the built-in per-modality MIME allowlist). For audio and video, the origin's `Content-Type` is checked against the per-modality allowlist on HTTP downloads; `image_url` retains its pre-existing permissive behavior on download. The per-modality cap bounds `data:` URIs in `audio_url` / `video_url` and `input_audio` inline payloads as well as HTTP downloads, measured on base64 length before decoding; an `image_url` `data:` URI is size-checked only when `max_image_download_size` is set; left unset, image inline data is bounded by `server.max_request_body_size` as before, so a config setting only `max_download_size` is unaffected. A `data:` URI is recognized whatever letter case its scheme is written in (`DATA:image/png;base64,...`), as RFC 3986 allows; only lower case `data:` counted as inline data before, and any other casing was treated as a link and refused. This applies to `image_url` as well as to the new audio and video paths. An `allowed_image_content_types` / `allowed_audio_content_types` / `allowed_video_content_types` key present with a null value instead of a list is rejected at startup rather than ignored, matching `allowed_domains`. Entries in those lists are normalized the same way a checked type is, so an entry written with a MIME parameter attached still matches the bare type it names. A `Content-Type` header carrying more than one value is reduced to its first type before both the allowlist check and the `data:` URI the step writes, so a multi-value header can no longer put a comma inside the emitted URI. `input_audio` is matched against the audio allowlist through one canonical MIME per format (`mp3` as `audio/mpeg`), and a rejection names the format alongside that MIME. On the generate path, a `features` map whose `mm_placeholders` or `kwargs_data` names a modality absent from `mm_hashes`, or whose per-modality lengths disagree with an empty `mm_hashes` list, is rejected as a client error instead of having that modality silently dropped from the prefill and decode bodies. A `features` map naming a modality with the empty string is likewise rejected as a client error, rather than having its entries treated as `image`. The shipped `config/coordinator/coordinator.yaml` sets `max_multimodal_entries: 8`; the code default remains unlimited, so a config that does not set the key is unaffected.
```

### DIFF
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -35,7 +35,7 @@ server:
   read_timeout: 30s
 
   # write_timeout caps the entire response. For streaming completions this
-  # bounds the total in-flight time of a single chat — set generously.
+  # bounds the total in-flight time of a single chat, set generously.
   write_timeout: 120s
   
   # Kubernetes terminationGracePeriodSeconds must exceed shutdown_timeout; SSE
@@ -108,7 +108,7 @@ pipeline:
     # audio_url, and video_url content parts (base64-inlining them as data:
     # URIs), validates input_audio inline items, and seeds MultimodalEntries
     # for every recognized media part. All modalities feed into the encode
-    # step's fanout — the encoder pod must accept audio/video content-part
+    # step's fanout, the encoder pod must accept audio/video content-part
     # types.
     # -------------------------------------------------------------------
     - type: replace-media-urls
@@ -149,7 +149,7 @@ pipeline:
         # max_video_download_size overrides max_download_size for video_url
         # items. Videos routinely need 100+ MB per clip; leave commented out
         # only if your deployment truly does not accept video. Same codec
-        # CVE caveat as audio applies — treat pending security review as a
+        # CVE caveat as audio applies, treat pending security review as a
         # gate on raising this in production.
         # max_video_download_size: 200
 
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -107,8 +107,9 @@ pipeline:
     # replace-media-urls: first step. Downloads http(s) URLs from image_url,
     # audio_url, and video_url content parts (base64-inlining them as data:
     # URIs), validates input_audio inline items, and seeds MultimodalEntries
-    # for image parts. Audio and video are validated + inlined but do not
-    # enter MultimodalEntries in this release.
+    # for every recognized media part. All modalities feed into the encode
+    # step's fanout — the encoder pod must accept audio/video content-part
+    # types (see H1 in coordinator-audio-video-plan.md).
     # -------------------------------------------------------------------
     - type: replace-media-urls
       params:
@@ -139,12 +140,17 @@ pipeline:
 
         # max_audio_download_size overrides max_download_size for audio_url
         # items and for input_audio inline payloads (checked against the
-        # decoded byte count implied by the base64 length).
+        # decoded byte count implied by the base64 length). Audio codec
+        # decoders historically carry more CVEs than image decoders; keep
+        # this cap tight until your security team has reviewed the audio
+        # decode path end-to-end (H2 in coordinator-audio-video-plan.md).
         # max_audio_download_size: 60
 
         # max_video_download_size overrides max_download_size for video_url
         # items. Videos routinely need 100+ MB per clip; leave commented out
-        # only if your deployment truly does not accept video.
+        # only if your deployment truly does not accept video. Same codec
+        # CVE caveat as audio applies — treat pending security review as a
+        # gate on raising this in production.
         # max_video_download_size: 200
 
         # allowed_image_content_types overrides the built-in image MIME
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -45,6 +45,9 @@ server:
   # max_request_body_size caps the request body in megabytes. The default is
   # sized for multimodal requests that inline images as data: URIs, which land
   # in the body verbatim and are not subject to replace-media-urls.max_download_size.
+  # Deployments that also accept inline audio or video (input_audio parts or
+  # data:audio/data:video URIs in audio_url/video_url slots) may need a much
+  # higher value: a 100 MB video inlines as ~135 MB of base64 in the body.
   # A text-only deployment can set a much smaller value. Megabytes; default 64.
   # env: COORDINATOR_SERVER_MAX_REQUEST_BODY_SIZE
   # max_request_body_size: 64
@@ -101,36 +104,75 @@ pipeline:
 
   steps:
     # -------------------------------------------------------------------
-    # replace-media-urls: first step. Downloads any http(s) image_url
-    # references, base64-inlines them as data: URIs, and seeds
-    # MultimodalEntries on the request context.
+    # replace-media-urls: first step. Downloads http(s) URLs from image_url,
+    # audio_url, and video_url content parts (base64-inlining them as data:
+    # URIs), validates input_audio inline items, and seeds MultimodalEntries
+    # for image parts. Audio and video are validated + inlined but do not
+    # enter MultimodalEntries in this release.
     # -------------------------------------------------------------------
     - type: replace-media-urls
       params:
-        # download_timeout caps each individual image download.
+        # download_timeout caps each individual media download.
         download_timeout: 10s
 
         # max_concurrent_downloads bounds the errgroup that fans out
-        # downloads. Useful when a single request references many images.
+        # downloads. Useful when a single request references many media items.
         max_concurrent_downloads: 10
 
-        # max_multimodal_entries rejects requests with more than N image_url
-        # parts BEFORE any download happens. Cheapest admission gate.
+        # max_multimodal_entries rejects requests with more than N media
+        # parts (image_url + audio_url + video_url + input_audio combined)
+        # BEFORE any download happens. Cheapest admission gate.
         # 0 (or omitted) = unlimited.
         # max_multimodal_entries: 8
 
-        # max_download_size caps the megabytes read per image download. A response
+        # max_download_size caps the megabytes read per media download. A response
         # whose Content-Length or body exceeds the cap is rejected as a client
         # error (4xx). Bounds peak memory together with the two limits above:
         # roughly max_concurrent_downloads x max_download_size x ~2.3 MB.
-        # data: URIs do not hit the network and are unaffected. Default 10 MB.
+        # data: URIs do not hit the network and are unaffected. This value is
+        # the global default; each modality below may override it. Default 10 MB.
         # max_download_size: 10
+
+        # max_image_download_size overrides max_download_size specifically for
+        # image_url items. Same units (MB), same overflow validation.
+        # max_image_download_size: 10
+
+        # max_audio_download_size overrides max_download_size for audio_url
+        # items and for input_audio inline payloads (checked against the
+        # decoded byte count implied by the base64 length).
+        # max_audio_download_size: 60
+
+        # max_video_download_size overrides max_download_size for video_url
+        # items. Videos routinely need 100+ MB per clip; leave commented out
+        # only if your deployment truly does not accept video.
+        # max_video_download_size: 200
+
+        # allowed_image_content_types overrides the built-in image MIME
+        # allowlist for image_url data URIs. When set, ONLY these types are
+        # accepted. Default: image/jpeg, image/png, image/gif, image/webp.
+        # allowed_image_content_types:
+        #   - image/jpeg
+        #   - image/png
+
+        # allowed_audio_content_types overrides the built-in audio MIME
+        # allowlist. Applies to audio_url data URIs and to input_audio parts.
+        # Default: audio/wav, audio/x-wav, audio/mpeg, audio/mp3, audio/flac,
+        # audio/x-flac, audio/ogg, audio/opus, audio/webm.
+        # allowed_audio_content_types:
+        #   - audio/wav
+        #   - audio/mpeg
+
+        # allowed_video_content_types overrides the built-in video MIME
+        # allowlist for video_url data URIs. Default: video/mp4, video/webm,
+        # video/quicktime, video/mpeg, video/ogg.
+        # allowed_video_content_types:
+        #   - video/mp4
 
         # allow_private_networks permits downloads whose resolved IP falls in
         # an RFC1918 private range (10/8, 172.16/12, 192.168/16). Loopback,
         # link-local (including the 169.254.169.254 cloud metadata endpoint),
         # CGNAT, and unique-local addresses stay blocked regardless. Enable
-        # only when image origins are inside the cluster network.
+        # only when media origins are inside the cluster network.
         # Default false.
         # allow_private_networks: false
 
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -109,7 +109,7 @@ pipeline:
     # URIs), validates input_audio inline items, and seeds MultimodalEntries
     # for every recognized media part. All modalities feed into the encode
     # step's fanout — the encoder pod must accept audio/video content-part
-    # types (see H1 in coordinator-audio-video-plan.md).
+    # types.
     # -------------------------------------------------------------------
     - type: replace-media-urls
       params:
@@ -143,7 +143,7 @@ pipeline:
         # decoded byte count implied by the base64 length). Audio codec
         # decoders historically carry more CVEs than image decoders; keep
         # this cap tight until your security team has reviewed the audio
-        # decode path end-to-end (H2 in coordinator-audio-video-plan.md).
+        # decode path end-to-end.
         # max_audio_download_size: 60
 
         # max_video_download_size overrides max_download_size for video_url
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -108,7 +108,7 @@ pipeline:
     # audio_url, and video_url content parts (base64-inlining them as data:
     # URIs), validates input_audio inline items, and seeds MultimodalEntries
     # for every recognized media part. All modalities feed into the encode
-    # step's fanout, the encoder pod must accept audio/video content-part
+    # step's fanout. The encoder pod must accept audio/video content-part
     # types.
     # -------------------------------------------------------------------
     - type: replace-media-urls
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -136,15 +136,16 @@ pipeline:
         # Peak memory is per REQUEST, not per concurrent download. Every
         # downloaded payload is retained until the step returns, so the bound is
         #
-        #   (media parts in the request) x (that modality's cap) x ~2.7
+        #   (media parts in the request) x (that modality's cap) x ~1.4
         #
-        # plus up to max_concurrent_downloads x cap of raw bytes in flight.
-        # max_concurrent_downloads throttles how fast bytes arrive; it does not
-        # cap how many are held at once.
+        # where the 1.4 is the base64 form the payload is inlined as, plus up to
+        # max_concurrent_downloads x cap x ~2.4 for the downloads still in
+        # flight. max_concurrent_downloads throttles how fast bytes arrive; it
+        # does not cap how many are held at once.
         #
         # That makes max_multimodal_entries above the limit that actually bounds
-        # this, and it is unlimited by default. 8 parts x 10 MB is ~220 MB; 8
-        # parts at max_video_download_size: 200 is ~4.3 GB; the same request with
+        # this, and it is unlimited by default. 8 parts x 10 MB is ~110 MB; 8
+        # parts at max_video_download_size: 200 is ~2.2 GB; the same request with
         # max_multimodal_entries unset is unbounded -- a request body of a few KB
         # can name hundreds of URLs. Raise a per-modality cap below only together
         # with max_multimodal_entries.
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -125,9 +125,11 @@ pipeline:
         # BEFORE any download happens. Cheapest admission gate, and the only
         # thing bounding how many bytes one request can pull into memory --
         # see max_download_size for why raising any per-modality cap means
-        # setting this one too.
-        # 0 (or omitted) = unlimited.
-        # max_multimodal_entries: 8
+        # raising this one too.
+        #
+        # 0 (or omitted) = unlimited, which lets a request body of a few KB
+        # name hundreds of URLs and turn each one into a download. Set it.
+        max_multimodal_entries: 8
 
         # max_download_size caps the megabytes read per media download. A response
         # whose Content-Length or body exceeds the cap is rejected as a client
@@ -152,12 +154,10 @@ pipeline:
         # to the encode step's max_parallel). Size the memory limit for several
         # times the number above, not for the number itself.
         #
-        # That makes max_multimodal_entries the limit that actually bounds this,
-        # and it is unlimited by default. 8 parts x 10 MB is ~110 MB in this step;
-        # 8 parts at max_video_download_size: 200 is ~2.2 GB in this step and
-        # multiples of that across the pipeline; the same request with
-        # max_multimodal_entries unset is unbounded -- a request body of a few KB
-        # can name hundreds of URLs. Raise a per-modality cap below only together
+        # That makes max_multimodal_entries the limit that actually bounds this.
+        # 8 parts x 10 MB is ~110 MB in this step; 8 parts at
+        # max_video_download_size: 200 is ~2.2 GB in this step and multiples of
+        # that across the pipeline. Raise a per-modality cap below only together
         # with max_multimodal_entries.
         #
         # This value is the global default; each modality below may override
@@ -232,6 +232,18 @@ pipeline:
         # image, the download path is checked whether or not this param is set.
         # Default: audio/wav, audio/x-wav, audio/mpeg, audio/mp3, audio/flac,
         # audio/x-flac, audio/ogg, audio/opus, audio/webm.
+        #
+        # An input_audio part names a format, not a MIME type, and the two are
+        # matched through one canonical MIME per format:
+        #
+        #   wav -> audio/wav      mp3  -> audio/mpeg   flac -> audio/flac
+        #   opus -> audio/opus    ogg  -> audio/ogg    webm -> audio/webm
+        #
+        # List the canonical type when narrowing this param, or input_audio
+        # parts of that format are rejected. audio/mp3 alone does not admit
+        # format "mp3"; audio/mpeg does. The aliases in the default list
+        # (audio/x-wav, audio/mp3, audio/x-flac) only ever match a data URI or
+        # a download's Content-Type.
         # allowed_audio_content_types:
         #   - audio/wav
         #   - audio/mpeg
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -175,24 +175,20 @@ pipeline:
         # decode, so an oversized item is rejected without being allocated.
         # max_download_size: 10
 
-        # max_image_download_size overrides max_download_size specifically for
-        # image_url items. Same units (MB), same overflow validation. Applies
-        # to http(s) downloads only; see max_download_size for which payloads
-        # each modality's cap covers.
+        # max_image_download_size overrides max_download_size for image_url
+        # items. Same units (MB), same overflow validation.
         # max_image_download_size: 10
 
         # max_audio_download_size overrides max_download_size for audio_url
-        # items, including audio_url data: URIs, and for input_audio inline
-        # payloads. Audio codec decoders historically carry more CVEs than
-        # image decoders; keep this cap tight until your security team has
-        # reviewed the audio decode path end-to-end.
+        # items. Audio codec decoders historically carry more CVEs than image
+        # decoders; keep this cap tight until your security team has reviewed
+        # the audio decode path end-to-end.
         # max_audio_download_size: 60
 
         # max_video_download_size overrides max_download_size for video_url
-        # items, including video_url data: URIs. Videos routinely need 100+ MB
-        # per clip; leave commented out only if your deployment truly does not
-        # accept video. See max_audio_download_size for the codec-CVE caveat
-        # that also applies here.
+        # items. Videos routinely need 100+ MB per clip; leave commented out
+        # only if your deployment truly does not accept video. The codec-CVE
+        # caveat on max_audio_download_size applies here too.
         # max_video_download_size: 200
 
         # allowed_{image,audio,video}_content_types below each REPLACE the
@@ -231,15 +227,12 @@ pipeline:
         # Default: audio/wav, audio/x-wav, audio/mpeg, audio/mp3, audio/flac,
         # audio/x-flac, audio/ogg, audio/opus, audio/webm.
         #
-        # An input_audio part names a format, not a MIME type, and the two are
-        # matched through one canonical MIME per format:
-        #
-        #   wav -> audio/wav      mp3  -> audio/mpeg   flac -> audio/flac
-        #   opus -> audio/opus    ogg  -> audio/ogg    webm -> audio/webm
-        #
-        # List the canonical type when narrowing this param, or input_audio
-        # parts of that format are rejected. audio/mp3 alone does not admit
-        # format "mp3"; audio/mpeg does. The aliases in the default list
+        # An input_audio part carries a format name. Each format is checked
+        # against one canonical MIME type; audioFormatMIME in
+        # pkg/coordinator/steps/replace_media_urls.go lists the pairs.
+        # Narrowing this param to a list that omits a format's canonical type
+        # rejects input_audio parts of that format: audio/mp3 alone does not
+        # admit format "mp3"; audio/mpeg does. The aliases in the default list
         # (audio/x-wav, audio/mp3, audio/x-flac) only ever match a data URI or
         # a download's Content-Type.
         # allowed_audio_content_types:
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -158,9 +158,16 @@ pipeline:
         # applies here.
         # max_video_download_size: 200
 
+        # allowed_{image,audio,video}_content_types below each REPLACE the
+        # built-in allowlist for that modality when set to a non-empty
+        # list. Set to an empty list to accept any content type for that
+        # modality, matching allowed_domains's "empty means unrestricted"
+        # convention.
+
         # allowed_image_content_types overrides the built-in image MIME
-        # allowlist for image_url data URIs. When set, ONLY these types are
-        # accepted. Default: image/jpeg, image/png, image/gif, image/webp.
+        # allowlist for image_url data URIs. When set to a non-empty list,
+        # ONLY these types are accepted. An empty list accepts any type.
+        # Default: image/jpeg, image/png, image/gif, image/webp.
         # allowed_image_content_types:
         #   - image/jpeg
         #   - image/png
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -128,8 +128,13 @@ pipeline:
 
         # max_download_size caps the megabytes read per media download. A response
         # whose Content-Length or body exceeds the cap is rejected as a client
-        # error (4xx). Bounds peak memory together with the two limits above:
-        # roughly max_concurrent_downloads x max_download_size x ~2.3 MB.
+        # error (4xx). Peak memory is bounded by max_concurrent_downloads x
+        # (largest active cap in MB) x ~2.3, where the 2.3 factor covers
+        # the raw bytes and their base64 form living in memory at the same
+        # time during encoding. With the defaults (10 concurrent x 10 MB)
+        # that is ~230 MB. Raising a per-modality cap below raises the
+        # ceiling in proportion: 10 concurrent downloads at
+        # max_video_download_size: 200 peak around 4.6 GB.
         # data: URIs do not hit the network and are unaffected. This value is
         # the global default; each modality below may override it. Default 10 MB.
         # max_download_size: 10
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -146,13 +146,13 @@ pipeline:
         # flight. max_concurrent_downloads throttles how fast bytes arrive; it
         # does not cap how many are held at once.
         #
-        # That figure is a floor for the pod, not the peak. The inlined payloads
+        # That figure is a floor for the pod, well under the peak. The payloads
         # stay on the request body, and each later step that sends the body
         # serializes them again into its own buffer: render, prefill, and decode
         # each hold another full copy while their upstream call is in flight, and
         # the encode fanout holds one media part per concurrent sub-request (up
         # to the encode step's max_parallel). Size the memory limit for several
-        # times the number above, not for the number itself.
+        # times the number above.
         #
         # That makes max_multimodal_entries the limit that actually bounds this.
         # 8 parts x 10 MB is ~110 MB in this step; 8 parts at
@@ -163,8 +163,7 @@ pipeline:
         # This value is the global default; each modality below may override
         # it. Default 10 MB.
         #
-        # Which payloads each cap covers differs by modality, because the image
-        # path predates the audio and video ones and keeps its behavior:
+        # Which payloads each cap covers differs by modality:
         #
         #   image  http(s) downloads only. An image data: URI is not
         #          size-checked; server.max_request_body_size bounds it.
@@ -213,11 +212,10 @@ pipeline:
         # accepted -- for image_url data URIs AND for the Content-Type an
         # http(s) image_url origin returns. An empty list accepts any type.
         #
-        # Leaving this unset is not the same as listing the defaults: the
-        # built-in default is enforced on data URIs only, because plenty of
-        # origins serve a valid image as application/octet-stream or with no
-        # Content-Type at all, and rejecting those would break traffic that
-        # has always worked. Setting the param is read as a deliberate
+        # Leaving this unset differs from listing the defaults: the built-in
+        # default is enforced on data URIs only, because plenty of origins
+        # serve a valid image as application/octet-stream or with no
+        # Content-Type at all. Setting the param is read as a deliberate
         # lockdown and covers both paths. Audio and video downloads are
         # always checked; see max_audio_download_size for why they are
         # treated more strictly.
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -201,6 +201,12 @@ pipeline:
         # built-in default is enforced on data URIs only, because plenty of
         # origins serve a valid image as application/octet-stream or with no
         # Content-Type at all, and rejecting those would break traffic that
+        #
+        # Write the key with a value or leave it out entirely. A key with
+        # nothing after it is a null value, not an empty list, and the
+        # coordinator refuses to start rather than guess which was meant.
+        # A template that renders one of these from an unset variable
+        # produces exactly that shape.
         # has always worked. Setting the param is read as a deliberate
         # lockdown and covers both paths. Audio and video downloads are
         # always checked; see max_audio_download_size for why they are
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -134,7 +134,8 @@ pipeline:
         # error (4xx).
         #
         # Peak memory is per REQUEST, not per concurrent download. Every
-        # downloaded payload is retained until the step returns, so the bound is
+        # downloaded payload is retained until the step returns, so what this
+        # step alone holds is
         #
         #   (media parts in the request) x (that modality's cap) x ~1.4
         #
@@ -143,9 +144,18 @@ pipeline:
         # flight. max_concurrent_downloads throttles how fast bytes arrive; it
         # does not cap how many are held at once.
         #
-        # That makes max_multimodal_entries above the limit that actually bounds
-        # this, and it is unlimited by default. 8 parts x 10 MB is ~110 MB; 8
-        # parts at max_video_download_size: 200 is ~2.2 GB; the same request with
+        # That figure is a floor for the pod, not the peak. The inlined payloads
+        # stay on the request body, and each later step that sends the body
+        # serializes them again into its own buffer: render, prefill, and decode
+        # each hold another full copy while their upstream call is in flight, and
+        # the encode fanout holds one media part per concurrent sub-request (up
+        # to the encode step's max_parallel). Size the memory limit for several
+        # times the number above, not for the number itself.
+        #
+        # That makes max_multimodal_entries the limit that actually bounds this,
+        # and it is unlimited by default. 8 parts x 10 MB is ~110 MB in this step;
+        # 8 parts at max_video_download_size: 200 is ~2.2 GB in this step and
+        # multiples of that across the pipeline; the same request with
         # max_multimodal_entries unset is unbounded -- a request body of a few KB
         # can name hundreds of URLs. Raise a per-modality cap below only together
         # with max_multimodal_entries.
@@ -191,6 +201,12 @@ pipeline:
         # list. Set to an empty list to accept any content type for that
         # modality, matching allowed_domains's "empty means unrestricted"
         # convention.
+        #
+        # Write the key with a value or leave it out entirely. A key with
+        # nothing after it is a null value, not an empty list, and the
+        # coordinator refuses to start rather than guess which was meant.
+        # A template that renders one of these from an unset variable
+        # produces exactly that shape.
 
         # allowed_image_content_types overrides the built-in image MIME
         # allowlist. When set to a non-empty list, ONLY these types are
@@ -201,12 +217,6 @@ pipeline:
         # built-in default is enforced on data URIs only, because plenty of
         # origins serve a valid image as application/octet-stream or with no
         # Content-Type at all, and rejecting those would break traffic that
-        #
-        # Write the key with a value or leave it out entirely. A key with
-        # nothing after it is a null value, not an empty list, and the
-        # coordinator refuses to start rather than guess which was meant.
-        # A template that renders one of these from an unset variable
-        # produces exactly that shape.
         # has always worked. Setting the param is read as a deliberate
         # lockdown and covers both paths. Audio and video downloads are
         # always checked; see max_audio_download_size for why they are
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -35,7 +35,7 @@ server:
   read_timeout: 30s
 
   # write_timeout caps the entire response. For streaming completions this
-  # bounds the total in-flight time of a single chat, set generously.
+  # bounds the total in-flight time of a single chat — set generously.
   write_timeout: 120s
   
   # Kubernetes terminationGracePeriodSeconds must exceed shutdown_timeout; SSE
@@ -122,19 +122,33 @@ pipeline:
 
         # max_multimodal_entries rejects requests with more than N media
         # parts (image_url + audio_url + video_url + input_audio combined)
-        # BEFORE any download happens. Cheapest admission gate.
+        # BEFORE any download happens. Cheapest admission gate, and the only
+        # thing bounding how many bytes one request can pull into memory --
+        # see max_download_size for why raising any per-modality cap means
+        # setting this one too.
         # 0 (or omitted) = unlimited.
         # max_multimodal_entries: 8
 
         # max_download_size caps the megabytes read per media download. A response
         # whose Content-Length or body exceeds the cap is rejected as a client
-        # error (4xx). Peak memory is bounded by max_concurrent_downloads x
-        # (largest active cap in MB) x ~2.3, where the 2.3 factor covers
-        # the raw bytes and their base64 form living in memory at the same
-        # time during encoding. With the defaults (10 concurrent x 10 MB)
-        # that is ~230 MB. Raising a per-modality cap below raises the
-        # ceiling in proportion: 10 concurrent downloads at
-        # max_video_download_size: 200 peak around 4.6 GB.
+        # error (4xx).
+        #
+        # Peak memory is per REQUEST, not per concurrent download. Every
+        # downloaded payload is retained until the step returns, so the bound is
+        #
+        #   (media parts in the request) x (that modality's cap) x ~2.7
+        #
+        # plus up to max_concurrent_downloads x cap of raw bytes in flight.
+        # max_concurrent_downloads throttles how fast bytes arrive; it does not
+        # cap how many are held at once.
+        #
+        # That makes max_multimodal_entries above the limit that actually bounds
+        # this, and it is unlimited by default. 8 parts x 10 MB is ~220 MB; 8
+        # parts at max_video_download_size: 200 is ~4.3 GB; the same request with
+        # max_multimodal_entries unset is unbounded -- a request body of a few KB
+        # can name hundreds of URLs. Raise a per-modality cap below only together
+        # with max_multimodal_entries.
+        #
         # data: URIs do not hit the network and are unaffected. This value is
         # the global default; each modality below may override it. Default 10 MB.
         # max_download_size: 10
@@ -183,7 +197,9 @@ pipeline:
         #   - image/png
 
         # allowed_audio_content_types overrides the built-in audio MIME
-        # allowlist. Applies to audio_url data URIs and to input_audio parts.
+        # allowlist. Applies to audio_url data URIs, to the Content-Type an
+        # http(s) audio_url origin returns, and to input_audio parts. Unlike
+        # image, the download path is checked whether or not this param is set.
         # Default: audio/wav, audio/x-wav, audio/mpeg, audio/mp3, audio/flac,
         # audio/x-flac, audio/ogg, audio/opus, audio/webm.
         # allowed_audio_content_types:
@@ -191,8 +207,10 @@ pipeline:
         #   - audio/mpeg
 
         # allowed_video_content_types overrides the built-in video MIME
-        # allowlist for video_url data URIs. Default: video/mp4, video/webm,
-        # video/quicktime, video/mpeg, video/ogg.
+        # allowlist. Applies to video_url data URIs and to the Content-Type an
+        # http(s) video_url origin returns; as with audio, the download path is
+        # checked whether or not this param is set.
+        # Default: video/mp4, video/webm, video/quicktime, video/mpeg, video/ogg.
         # allowed_video_content_types:
         #   - video/mp4
 
--- a/config/coordinator/coordinator.yaml
+++ b/config/coordinator/coordinator.yaml
@@ -150,27 +150,40 @@ pipeline:
         # can name hundreds of URLs. Raise a per-modality cap below only together
         # with max_multimodal_entries.
         #
-        # data: URIs do not hit the network and are unaffected. This value is
-        # the global default; each modality below may override it. Default 10 MB.
+        # This value is the global default; each modality below may override
+        # it. Default 10 MB.
+        #
+        # Which payloads each cap covers differs by modality, because the image
+        # path predates the audio and video ones and keeps its behavior:
+        #
+        #   image  http(s) downloads only. An image data: URI is not
+        #          size-checked; server.max_request_body_size bounds it.
+        #   audio  http(s) downloads, audio_url data: URIs, and input_audio
+        #          inline payloads.
+        #   video  http(s) downloads and video_url data: URIs.
+        #
+        # Inline payloads are bounded on their base64 length, before any
+        # decode, so an oversized item is rejected without being allocated.
         # max_download_size: 10
 
         # max_image_download_size overrides max_download_size specifically for
-        # image_url items. Same units (MB), same overflow validation.
+        # image_url items. Same units (MB), same overflow validation. Applies
+        # to http(s) downloads only; see max_download_size for which payloads
+        # each modality's cap covers.
         # max_image_download_size: 10
 
         # max_audio_download_size overrides max_download_size for audio_url
-        # items and for input_audio inline payloads (checked against the
-        # decoded byte count implied by the base64 length). Audio codec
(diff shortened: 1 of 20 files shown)


pkg/coordinator/steps/replace_media_urls.go
```diff
@@ -234,24 +311,56 @@ func (s *ReplaceMediaURLsStep) Execute(ctx context.Context, reqCtx *pipeline.Req
 
 	for _, r := range results {
```
roytman: Execute splits media into two separate slices:  `urlRefs` (image_url/audio_url/video_url) and `inlineRefs` (input_audio), walked and appended to `MultimodalEntries` as two separate passes: all results (from `urlRefs`) first, then all `inlineRefs` second. But `collectMediaParts` in `encode.go` and the walker in `decode.injectUUIDs` both group parts by modality in true request/walker order, and `localIndexOf` (`utils.go:163`) computes an entry's local index from its position in `MultimodalEntries`, which is the split-append order, not walker order.

Concretely: a request with `content: [input_audio A, audio_url B]` produces `MultimodalEntries = [B, A]` (URL results are appended before inline refs), but `collectMediaParts`'s `partsByMod["audio"]` is `[A, B]` (true order). `localIndexOf` gives B local index 0, which then looks up `partsByMod["audio"][0] = A`'s content part in` buildSingleMediaContent` (`encode.go:287`), B's hash gets encoded against A's bytes. The same divergence hits `render.go`'s modIndex walk (line 298-308) and `decode.go`'s `injectUUIDs` (line 148-184), which also key off `MultimodalEntries` order vs. true walker order, respectively.

This only manifests when a modality has both a URL-based and an inline part type, today that's audio only (`audio_url`+ `input_audio`), but it's a real cross-wiring, not a hypothetical: any request mixing them gets the wrong bytes hashed/encoded/decoded against the wrong entry, silently. Suggest building one walker-ordered ref list up front (tag each ref with its source kind) instead of two separately-ordered slices, so `MultimodalEntries` append order matches `collectMediaParts/injectUUIDs` walk order.

pkg/coordinator/steps/params.go
```diff
@@ -29,7 +29,18 @@ const (
...
 )
 
-const ModalityImage = "image"
+// Modality* identify the kind of multimodal payload carried by a
+// MultimodalEntry and are used as the key into per-modality feature maps
+// (mm_hashes, mm_placeholders, kwargs_data). Values mirror EPP's canonical
+// modality enum at pkg/epp/framework/interface/requesthandling/types.go.
+// The coordinator today only produces ModalityImage; ModalityAudio and
+// ModalityVideo are reserved for the follow-up widening (see
+// coordinator-audio-video-plan.md).
```
roytman: "The coordinator today only produces ModalityImage... reserved for the follow-up widening (see coordinator-audio-video-plan.md)" is no longer true as of this PR, and `coordinator-audio-video-plan.md` isn't in the repo.

pkg/coordinator/steps/utils.go
```diff
@@ -110,46 +111,85 @@ func capSingleTokenOutput(body map[string]any, format gateway.RequestFormat) {
 
 // buildMMFeatures builds the multimodal features map (mm_hashes, mm_placeholders,
 // and optionally kwargs_data) from the request's multimodal entries. It returns
-// nil when there are no entries.
+// nil when there are no entries. Entries are grouped by Modality so a
+// mixed-modality request produces one key per modality in each feature map
+// (today MultimodalEntries is populated only for image, so the output is a
```
roytman: Same stale claim repeated here ("today MultimodalEntries is populated only for image... identical to prior behaviour"), false as of this PR, and duplicates the fact in params.go. State without "today"/"prior".

pkg/coordinator/steps/replace_media_urls.go
```diff
@@ -234,24 +311,56 @@ func (s *ReplaceMediaURLsStep) Execute(ctx context.Context, reqCtx *pipeline.Req
...
 		if !strings.HasPrefix(r.ref.url, "data:") {
-			r.ref.imageURL["url"] = fmt.Sprintf("data:%s;base64,%s", r.contentType, r.base64Data)
+			r.ref.urlMap["url"] = fmt.Sprintf("data:%s;base64,%s", r.contentType, r.base64Data)
 		}
-
-		appendMultimodalEntry(reqCtx, r.contentType, r.base64Data)
+		// All modalities feed into MultimodalEntries: the encode step's
+		// fanout iterates entries and produces per-modality sub-requests.
+		// See coordinator-audio-video-plan.md for the H1 wire-contract
```
roytman: Dangling reference to `coordinator-audio-video-plan.md` (H1 section) — this file isn't in the repo. Same issue at config/coordinator/coordinator.yaml:112,146. Either add the doc or fold the rationale into the comment itself.

config/coordinator/coordinator.yaml
```diff
@@ -101,36 +104,81 @@ pipeline:
...
     # -------------------------------------------------------------------
-    # replace-media-urls: first step. Downloads any http(s) image_url
-    # references, base64-inlines them as data: URIs, and seeds
-    # MultimodalEntries on the request context.
+    # replace-media-urls: first step. Downloads http(s) URLs from image_url,
+    # audio_url, and video_url content parts (base64-inlining them as data:
+    # URIs), validates input_audio inline items, and seeds MultimodalEntries
+    # for every recognized media part. All modalities feed into the encode
+    # step's fanout — the encoder pod must accept audio/video content-part
+    # types (see H1 in coordinator-audio-video-plan.md).
```
roytman: References `coordinator-audio-video-plan.md` (H1/H2), not present in the repo, so this is a broken pointer for any operator reading the shipped config.

config/coordinator/coordinator.yaml
```diff
@@ -101,36 +104,81 @@ pipeline:
...
+        # max_image_download_size overrides max_download_size specifically for
+        # image_url items. Same units (MB), same overflow validation.
+        # max_image_download_size: 10
+
+        # max_audio_download_size overrides max_download_size for audio_url
+        # items and for input_audio inline payloads (checked against the
+        # decoded byte count implied by the base64 length). Audio codec
+        # decoders historically carry more CVEs than image decoders; keep
+        # this cap tight until your security team has reviewed the audio
+        # decode path end-to-end (H2 in coordinator-audio-video-plan.md).
```
roytman: References `coordinator-audio-video-plan.md` (H2), not present in the repo, so this is a broken pointer for any operator reading the shipped config.

pkg/coordinator/steps/replace_media_urls_test.go
```diff
@@ -1137,3 +1137,629 @@ func TestReplaceMediaURLsStep_CancelledContextSkipsDataURIParse(t *testing.T) {
...
 	}
 }
+
+// ---- Audio / video ingestion (Section 2) -----------------------------------
+
+// TestReplaceMediaURLsStep_AudioURL_Downloads asserts that an audio_url is
+// fetched, size-capped, MIME-checked (against the audio allowlist for data
+// URIs) and inlined as a data URI. Because audio does not enter the encoder
+// pipeline yet, MultimodalEntries stays empty.
+func TestReplaceMediaURLsStep_AudioURL_Downloads(t *testing.T) {
```
roytman, line 1346: These doc comments say audio/video "stay out of" / "does not enter" `MultimodalEntries` yet, but the test bodies assert that entries ARE present (`len(reqCtx.MultimodalEntries) == 1`, etc.). The comments describe an earlier draft of this PR's behavior, not what's being tested now.

pkg/coordinator/steps/replace_media_urls_test.go
```diff
@@ -1137,3 +1137,629 @@ func TestReplaceMediaURLsStep_CancelledContextSkipsDataURIParse(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		},
+	}
+	if err := step.Execute(context.Background(), reqCtx); err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if len(reqCtx.MultimodalEntries) != 1 {
```
roytman, line 1843: These comments say "MultimodalEntries stays empty," but the test checks that `if len(reqCtx.MultimodalEntries) != 1 {`

pkg/coordinator/steps/replace_media_urls_test.go
```diff
@@ -1137,3 +1137,629 @@ func TestReplaceMediaURLsStep_CancelledContextSkipsDataURIParse(t *testing.T) {
...
+		},
+	}
+
+	if err := step.Execute(context.Background(), reqCtx); err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	// Section 5: all three media parts feed into MultimodalEntries.
+	// Entries appear in walker order: URL refs first (image, audio, video),
+	// then any inline refs (none here).
+	if len(reqCtx.MultimodalEntries) != 3 {
```
roytman, line 2310: the comments say "The image feeds into
// MultimodalEntries; audio and video are inlined but stay out."  here we see `if len(reqCtx.MultimodalEntries) != 3 {`

pkg/coordinator/steps/render.go
```diff
@@ -270,33 +270,59 @@ func (s *RenderStep) executeChatCompletions(ctx context.Context, reqCtx *pipelin
...
 
-	imageHashes := renderResp.Features.MMHashes[ModalityImage]
-	imagePlaceholders := renderResp.Features.MMPlaceholders[ModalityImage]
-	imageKwargs := renderResp.Features.KwargsData[ModalityImage]
-
+	// Length check: sum of per-modality slice lengths must match len(entries).
+	// Today entries carry Modality=image and the response has only [image]
+	// keys, so this collapses to the single-key check that was here before.
+	totalHashes, totalPlaceholders, totalKwargs := 0, 0, 0
+	for _, s := range renderResp.Features.MMHashes {
```
roytman: nit: `for _, s := range renderResp.Features.MMHashes` (and the two loops after it) shadow the method's s `*RenderStep` receiver, which is used both above (s.postRender) and below (s.checkPlaceholderLimit) in the same function. Not a bug since `s` isn't read inside these loops, but worth a different loop variable name to avoid confusion.

pkg/coordinator/steps/decode.go
```diff
@@ -155,12 +174,15 @@ func (s *DecodeStep) injectUUIDs(reqCtx *pipeline.RequestContext) {
...
+			modality, isMedia := partTypeModality[partType]
+			if !isMedia {
 				continue
 			}
-			if hashIdx < len(reqCtx.MultimodalEntries) {
-				partMap["uuid"] = reqCtx.MultimodalEntries[hashIdx].Hash
-				hashIdx++
+			localIdx := modCounter[modality]
+			modCounter[modality]++
+			if h, ok := hashByModLocalIdx[modality][localIdx]; ok {
```
roytman: `if h, ok := hashByModLocalIdx[modality][localIdx]; ok { partMap["uuid"] = h }` silently leaves the part untagged on a miss: no log, no error. Given the PR's stated "fail loudly" philosophy elsewhere (`render.go`'s per-modality mismatch errors), a decode-stage pairing miss here is exactly the kind of silent divergence that should at least log at `logged.DEBUG` or count a metric, since it'd otherwise be invisible in production.

roytman (commented):
Solid design and strong config-validation/test coverage for the ingestion side (per-modality size caps, MIME allowlists, the audio-hashes-without-placeholders behavior change is directly tested). 

Unicode symbols: 20 occurrences across new comments, mostly em dashes used as clause separators, plus two arrows:

decode.go:139,148 encode.go:244,278,279,284,285 render.go:299 replace_media_urls.go:284,424 replace_media_urls_test.go:1528,1676,1728,1746 utils.go:117,156 utils_test.go:388,655,698 coordinator.yaml:38,111,152

Suggest a straight-ASCII sweep (commas/periods/parens instead of em dashes, "->" or spelled out instead of →/↔) per project convention.

pkg/coordinator/steps/replace_media_urls.go
dmitripikus: Collected all media parts into a single walker-order slice tagged by kind (URL vs inline) and process/append in that one order, so MultimodalEntries matches the order encode/decode walk.
Thanks!

pkg/coordinator/steps/params.go
dmitripikus: This item is already addressed in previous fix (commit d20fc39)
Thanks!

pkg/coordinator/steps/utils.go
dmitripikus: This item is already addressed in previous fix (commit d20fc39)
Thanks!

dmitripikus: There is also a short follow-up fix in commit 4daff52 .

pkg/coordinator/steps/replace_media_urls.go
dmitripikus: The item is addressed in commit  d20fc39
Thanks!

config/coordinator/coordinator.yaml
dmitripikus: The item is addressed in commit d20fc39
Thanks!

config/coordinator/coordinator.yaml
dmitripikus: The item is addressed in commit  d20fc39
Thanks!

pkg/coordinator/steps/replace_media_urls_test.go
dmitripikus: Rewrote the three stale docstrings (AudioURL_Downloads, AudioDataURI, MixedImageAudioVideo) to match what the bodies actually check.
Thanks!

pkg/coordinator/steps/replace_media_urls_test.go
```diff
@@ -1137,3 +1137,869 @@ func TestReplaceMediaURLsStep_CancelledContextSkipsDataURIParse(t *testing.T) {
...
+	if err := step.Execute(context.Background(), reqCtx); err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if len(reqCtx.MultimodalEntries) != 1 {
+		t.Fatalf("expected 1 audio entry in MultimodalEntries, got %d", len(reqCtx.MultimodalEntries))
+	}
+	if reqCtx.MultimodalEntries[0].Modality != ModalityAudio {
+		t.Fatalf("expected Modality=%q, got %q", ModalityAudio, reqCtx.MultimodalEntries[0].Modality)
+	}
+	if reqCtx.MultimodalEntries[0].ContentType != "audio/wav" {
```
roytman: CI lint failure (blocking merge): golangci-lint `goconst` flags `"audio/wav"` as appearing 3x in this file (here, plus lines ~1433 and ~1709 in the head commit). Extract it to a package-level test const, e.g. `const testAudioWAVMIME = "audio/wav"`.

pkg/coordinator/steps/replace_media_urls.go
```diff
@@ -191,29 +289,48 @@ func (s *ReplaceMediaURLsStep) Execute(ctx context.Context, reqCtx *pipeline.Req
...
 				return fmt.Errorf("downloading %s: %w", ref.url, err)
 			}
-			results[i] = downloadResult{
-				ref:         ref,
+			// Audio and video decoders have historically carried more CVEs
+			// than image decoders, so the origin's Content-Type is checked
+			// against the per-modality allowlist for those two. image_url
+			// keeps its pre-existing permissive treatment on downloads (data
+			// URIs are still checked).
+			if ref.modality != ModalityImage && !s.allowedContentTypeForModality(contentType, ref.modality) {
```
roytman: Possible false rejection of legitimate audio/video downloads: `allowedContentTypeForModality` compares the raw `Content-Type` response header (lowercased/trimmed) directly against the MIME allowlist, without stripping `;`-separated parameters. Contrast with `parseDataURI`, which already does `strings.Cut(meta, ";")` before the same comparison. A real audio/video origin that returns e.g. `Content-Type: video/mp4; codecs="avc1.4D401E"` would get rejected here even though `video/mp4` is in the default allowlist. Since this strict check is new for downloads (image keeps the old permissive path on purpose), worth deciding: strip params like the data-URI path does, or confirm this is intentionally strict.

pkg/coordinator/steps/replace_media_urls_test.go
dmitripikus: The istem is addressed in commit beeeaa1.
Thanks!

pkg/coordinator/steps/replace_media_urls_test.go
dmitripikus: Addressed in commit beeeaa1.

Thanks!

pkg/coordinator/steps/decode.go
dmitripikus: Added a DEBUG log on the miss branch of injectUUIDs with modality, local index, and modality entry count, so the divergence is observable. Request still succeeds; only diagnostics change.

Thanks!

pkg/coordinator/steps/render.go
dmitripikus: The three loop variables are now hashes, placeholders, and kwargs instead of s.
Thanks!

pkg/coordinator/steps/replace_media_urls_test.go
dmitripikus: Extracted the four repeated MIME literals ("audio/wav", "image/png", "image/jpeg", "video/mp4") to file-level test constants and replaced their uses.

Thanks!

pkg/coordinator/steps/replace_media_urls.go
dmitripikus: `allowedContentTypeForModality` now splits on `;` before the allowlist lookup, matching parseDataURI's convention. Legitimate origins returning video/mp4; codecs="avc1.4D401E" or audio/wav; charset=US-ASCII are accepted; text/html; charset=utf-8 still rejects because the media-type portion isn't on the allowlist. 

Thanks!

pkg/coordinator/steps/replace_media_urls.go
```diff
@@ -280,16 +436,16 @@ func (s *ReplaceMediaURLsStep) download(ctx context.Context, rawURL string) ([]b
...
+	data, err := io.ReadAll(io.LimitReader(resp.Body, sizeCap+1))
 	if err != nil {
 		return nil, "", err
 	}
-	if int64(len(data)) > s.maxDownloadSize {
-		return nil, "", fmt.Errorf("response too large: body exceeds max %d: %w", s.maxDownloadSize, pipeline.ErrBadRequest)
+	if int64(len(data)) > sizeCap {
+		return nil, "", fmt.Errorf("response too large: body exceeds max %d: %w", sizeCap, pipeline.ErrBadRequest)
 	}
 	contentType := resp.Header.Get("Content-Type")
```
roytman: Following https://github.com/llm-d/llm-d-router/pull/2659/changes#r3940314232

The `Content-Type` header is stored here without stripping MIME parameters:
```go
contentType := resp.Header.Get("Content-Type")
if contentType == "" {
    contentType = defaultContentType
}
```

`allowedContentTypeForModality` (line 535) strips parameters (`strings.Cut(contentType, ";")`) before comparing against the allowlist, so a header like `video/mp4; codecs="avc1.4D401E,mp4a.40.2"` now correctly passes that check. But the unstripped value is what actually gets returned and stored in `mediaResult.contentType`, it's what gets written into the rewritten data URI at line 391:
```go
ref.urlMap["url"] = fmt.Sprintf("data:%s;base64,%s", r.contentType, r.base64Data)
```
When the `Content-Type`'s parameter value contains a comma (routine for multi-codec video, e.g. the example above), the resulting `data:video/mp4; codecs="avc1.4D401E,mp4a.40.2";base64,...` can't be re-parsed: `parseDataURI` (line 668) cuts on the first comma, landing inside the codecs list instead of at `;base64,`, and the base64-marker check fails.

Suggest normalizing once here, at the source, so every consumer gets the clean value instead of just the allowlist check:

```go
contentType := resp.Header.Get("Content-Type")
if contentType == "" {
    contentType = defaultContentType
}
contentType, _, _ = strings.Cut(contentType, ";")
contentType = strings.ToLower(strings.TrimSpace(contentType))
return data, contentType, nil
```
This mirrors the normalization p`arseDataURI` already does (line 686) and makes `allowedContentTypeForModality`'s own `strings.Cut `at line 535 redundant (harmless, but could be dropped as a followup). 
It also means `MultimodalEntry.ContentType` (set via `appendMultimodalEntry`, line 393) carries a clean MIME type rather than one with stray parameters, which matters if the encoder's wire contract expects a bare type.

`TestReplaceMediaURLsStep_AudioVideo_AcceptsContentTypeWithParams` only asserts `err == nil`, worth adding an assertion that the rewritten URL round-trips through `parseDataURI`, so this doesn't silently regress again.

pkg/coordinator/steps/replace_media_urls.go
roytman: see below

pkg/coordinator/steps/replace_media_urls.go
```diff
@@ -232,26 +349,63 @@ func (s *ReplaceMediaURLsStep) Execute(ctx context.Context, reqCtx *pipeline.Req
...
+			// The inline "data" field already carries the base64 payload
+			// the backend needs, so the body passes through unmodified.
+			appendMultimodalEntry(reqCtx, ref.modality, contentType, ref.data)
+			continue
 		}
-
-		appendMultimodalEntry(reqCtx, r.contentType, r.base64Data)
+		r := results[i]
+		// r.contentType is set on every URL ref the download/parse path
+		// populated. Skip a zero-valued slot defensively so a future path
```
roytman: This comment says the guard exists so "a future path that leaves `results[i]` unset after `g.Wait` cannot panic on a nil-map write," but the condition tested is `ref.urlMap`, a field on ref set at construction time, not on `results[i]`. `ref.urlMap` can't be `nil` for any ref that made it into refs (construction requires a successful map type-assertion first), so as written this branch is unreachable for the failure mode the comment describes. If a future change genuinely does leave `results[i]` unpopulated, this guard wouldn't catch it, and the entry would silently vanish (already counted against `max_multimodal_entries`), shifting every later same-modality index. Either check `results[i]` directly (e.g. `r.contentType == ""`) or drop the guard.

pkg/coordinator/steps/encode.go
```diff
@@ -255,24 +272,59 @@ func collectImageParts(body map[string]any) []map[string]any {
...
+//
+// If localIdx is out of range for the given modality, an empty-shaped
+// URL part of the matching modality is returned as a safe fallback.
+// Reaching that path means the coordinator's entry<->part pairing is
+// broken; the fallback prevents a panic but does not hide the bug, the
+// encoder receives an empty URL of the right modality and rejects the
+// sub-request loudly.
+func buildSingleMediaContent(partsByMod map[string][]map[string]any, modality string, localIdx int) map[string]any {
+	parts := partsByMod[modality]
+	if localIdx < 0 || localIdx >= len(parts) {
```
roytman, line 319: This is the encode-side counterpart of the pairing-miss case that's logged at `logutil.DEBUG` in `decode.go:201`, but here the fallback (`{"type": partType, partType: {"url": ""}}`) is silent,  no log at all. If the entry part pairing ever breaks upstream, this emits an empty-URL content part with no signal pointing back at the coordinator; whoever debugs the resulting encoder-side error has nothing to connect it to. 
Suggest mirroring the log call from `decode.go:201` here (modality, localIdx, parts available).

pkg/coordinator/steps/utils.go
```diff
@@ -110,46 +111,65 @@ func capSingleTokenOutput(body map[string]any, format gateway.RequestFormat) {
...
-	for i, k := range kwargs {
-		if k != "" {
-			items[i] = k
-		}
+// entryModality returns the entry's Modality with an empty-string fallback
+// to ModalityImage. All production entry producers (replace_media_urls,
+// extractMultimodalEntries) set Modality explicitly; the fallback exists so
+// callers constructing entries directly (mostly test fixtures predating the
+// Modality field) do not silently produce a "" modality key.
+func entryModality(entry pipeline.MultimodalEntry) string {
```
roytman: nit: Same silent-default pattern: an empty Modality on a production entry silently becomes "image", with no log. 
The doc comment (line 145: "mostly test fixtures predating the Modality field") also uses the temporal/deictic phrasing this repo's CLAUDE.md asks to avoid ("previously", "predating", etc.,  "a reader with no context for the change must still understand the text"). 
Since this is called O(entries) times, not O(requests), a logger.V(logutil.DEBUG) call on the default branch would be cheap and would make a real occurrence of this discoverable instead of silently plausible.

pkg/coordinator/steps/replace_media_urls.go
```diff
@@ -298,34 +454,215 @@ func (s *ReplaceMediaURLsStep) download(ctx context.Context, rawURL string) ([]b
...
+		if !ok {
+			return nil, fmt.Errorf("%s entries must be strings, got %T", fieldName, e)
+		}
+		set[strings.ToLower(strings.TrimSpace(mime))] = struct{}{}
+	}
+	return set, nil
+}
+
+// audioFormatMIME maps OpenAI's input_audio.format values to canonical MIME
+// types. OpenAI documents "wav" and "mp3" today; the additional entries
```
roytman: nit:  "OpenAI documents "wav" and "mp3" **today**" it's exactly the temporal framing CLAUDE.md's comment conventions ask to avoid, and it silently goes stale if OpenAI's docs change.

pkg/coordinator/steps/replace_media_urls.go
dmitripikus: download() now normalizes the Content-Type at the source (strips MIME params, lowercase/trim), so the allowlist check, the rewritten data URI, and MultimodalEntry.ContentType all see a bare media type. TestReplaceMediaURLsStep_AudioVideo_AcceptsContentTypeWithParams now round-trips the rewritten URL through parseDataURI and asserts the bare media type on the entry, so this can't regress silently.

Thanks!

pkg/coordinator/steps/replace_media_urls.go
dmitripikus: Dropped the dead branch. ref.urlMap is set at construction time and is guaranteed non-nil for any ref in refs, so the guard could never fire; the comment was describing a scenario the guard didn't actually catch.
Thanks!

pkg/coordinator/steps/encode.go
dmitripikus: `buildSingleMediaContent` now returns a usedFallback flag; the encode goroutine logs a DEBUG line on the miss (modality, local index, parts available), mirroring the decode-side miss log in injectUUIDs.

Thanks!

pkg/coordinator/steps/utils.go
dmitripikus: entryModality now takes a logger and writes a DEBUG line when it falls back to image. Also passed a logger through the few helpers that call it. Comment reworded to drop "predating".

Thanks!

pkg/coordinator/steps/replace_media_urls.go
dmitripikus: Rewrote the audioFormatMIME comment without "today".

Thanks!

pkg/coordinator/steps/replace_media_urls.go
```diff
@@ -191,39 +352,64 @@ func (s *ReplaceMediaURLsStep) Execute(ctx context.Context, reqCtx *pipeline.Req
...
+				return fmt.Errorf("data URI at message %d part %d exceeds size limit for %s: %w",
+					ref.msgIdx, ref.partIdx, ref.modality, pipeline.ErrBadRequest)
 			}
-			results[i] = downloadResult{ref: ref, base64Data: b64, contentType: contentType}
 			continue
 		}
+		downloadCount++
 		g.Go(func() error {
-			data, contentType, err := s.download(gCtx, ref.url)
+			data, contentType, err := s.download(gCtx, ref.url, ref.modality)
```
roytman, line 386: `download()` does `io.ReadAll` up to the size cap before the caller checks the allowlist. With a 200 MB video cap, an origin serving `text/html` still costs a full read before rejection, this is the exact DoS/CVE-surface mitigation the PR description claims to provide, and as positioned, it doesn't. 
Move the header (it is available at line 562); check inside `download()`, right after the status check, before `io.ReadAll`.

pkg/coordinator/steps/utils.go
```diff
@@ -110,46 +112,87 @@ func capSingleTokenOutput(body map[string]any, format gateway.RequestFormat) {
...
+// field does not produce a "" modality key.
+//
+// The fallback carries a cost: it groups the entry under image everywhere
+// (mm_hashes, the encode fanout, render's per-modality slots), so a
+// wrongly tagged entry pairs with the wrong part and the failure surfaces far
+// from here, as an encoder rejection. Log through logger.Error, which is
+// emitted at any verbosity, so an occurrence is visible in production
+// rather than only under debug logging.
+func entryModality(entry pipeline.MultimodalEntry, logger logr.Logger) string {
+	if entry.Modality == "" {
```
roytman: Logs an error but returns a guessed modality and lets every caller (render/encode/decode) proceed with it, rather than erroring like `encode.go`'s analogous case does. Inconsistent fail-open/fail-closed behavior for the same bug class in the same PR.

Why entry.Modality == "" should never happen: every MultimodalEntry is created upstream in replace_media_urls.go, which always stamps one of ModalityImage/ModalityAudio/ModalityVideo on it. So an empty Modality reaching this function means an invariant was already broken elsewhere in the coordinator — a real bug, not a client input problem.

What happens on that bug: the function logs at Error level (so it is visible, not literally silent as a log event) — but then returns "image" as if that were the entry's real modality, and every caller proceeds normally with that value.

Why that's dangerous rather than harmless: all four callers use the returned modality as a map key to do positional per-modality pairing:

- `buildMMFeatures` (`utils.go:132`) buckets the entry's hash/placeholder into `hashesByMod["image"]` / `placeholdersByMod["image"]`.
- `encode.go:120` computes `localIdx := modCounter["image"]` and uses it to find the matching image content part in the request.
- `decode.go:154` groups hashes the same way to re-tag response parts with UUIDs.
- `render.go:301` walks `modIndex["image"]` to pull the entry's hash/placeholder/kwargs out of the render response's per-modality slot.
If a genuinely audio/video entry gets mislabeled `"image"` by this fallback, it gets spliced into the image bucket's index sequence at every one of those four independent call sites. Concretely: it can pair with a real image's content part or response slot (wrong bytes tied to the wrong hash), and it shifts the local index for every real image entry that appears later in the request, so those get bumped to the wrong slot too. The request doesn't fail, but it completes and returns a response built on a guess, which is exactly the "data corruption without an error" pattern CLAUDE.md's error-handling conventions and this PR's own design (see below) are trying to avoid.

Why this is inconsistent with the rest of the PR: the exact same failure class - entries and parts out of line - is treated as fatal everywhere else. `encode.go`'s `buildSingleMediaContent` path returns an error when it can't find a matching part, and that propagates up as a failed request (`TestEncodeStep_MissingMediaPartFails` covers it). `render.go's idx >= len(hashes)` check (`render.go:307`) does the same. `entryModality` is the one place in this same invariant-violation family that logs and continues instead of returning an error the caller can fail on.

The fix I'd suggest: change the signature to `func entryModality(entry pipeline.MultimodalEntry, logger logr.Logger) (string, error)`, keep the `logger.Error` call for visibility, and have each of the four callers propagate the error the same way `encode.go` already does for its analogous case - fail the request instead of guessing.

pkg/coordinator/steps/decode.go
```diff
@@ -155,13 +170,31 @@ func (s *DecodeStep) injectUUIDs(reqCtx *pipeline.RequestContext) {
...
+			if localIdx < len(hashes) {
+				partMap["uuid"] = hashes[localIdx]
+				continue
 			}
+			// A miss means entries and parts got out of line upstream
+			// (see mediaPartIsWellFormed). The part still goes to the
+			// backend without its uuid, so this costs a cache lookup
+			// rather than the request. Log at DEBUG so the mismatch is
+			// visible when someone looks.
+			logger.V(logutil.DEBUG).Info("no MultimodalEntry for well-formed media part",
```
roytman, line 197: Pairing-mismatch
Same "entries/parts out of sync" condition that's a hard error in `encode.go`, here it's invisible at default verbosity (DEBUG=4 vs DEFAULT=2).

`TestInjectUUIDs_TagsAllMediaParts`, `TestInjectUUIDs_TagsRepeatedModalityInOrder`, and `TestInjectUUIDs_SkipsMalformedParts` in  (`decode_test.go:458,504,533`) all use exactly-matched entry/part counts. None drives this branch; more well-formed parts than entries for a modality. 
Worth writing a test where the number of well-formed media parts in the request is larger than the number of matching entries for that modality (say, 2 audio parts but only 1 audio entry). Today, when that happens, `decode.go`'s `injectUUID`s doesn't return an error; it just logs at DEBUG and leaves that one part without a uuid tag, and the request goes on to the backend as-is. No test currently exercises that path, so nobody would notice if a future change accidentally turned it into a hard failure, or into something worse (like tagging the part with the wrong hash instead of leaving it untagged).

The contrast with `encode.go`: that step hits the same underlying bug (entries and parts out of sync) but responds completely differently; it returns an error and the whole request fails, rather than continuing with one part silently incomplete. So the same root problem is handled two different ways in the same PR, and neither way is currently tested for `decode.go`.

So, concretely, the test would: build a request with parts/entries deliberately out of sync, call `injectUUID`s, and assert that the mismatched part ends up without a `uuid` key and the function returns without error; documenting today's "degrade gracefully" behavior on purpose, rather than leaving that behavior unverified.

pkg/coordinator/steps/replace_media_urls.go
```diff
@@ -44,10 +45,118 @@ import (
...
+	}
+	return strings.ToLower(strings.TrimSpace(raw))
+}
+
+// dataURIPrefix is the scheme prefix of a data URI. RFC 2397 scheme names
+// are case-insensitive, so every comparison against it goes through
+// isDataURI rather than strings.HasPrefix.
+const dataURIPrefix = "data:"
+
+// isDataURI reports whether s carries the data: scheme, ignoring case.
```
roytman, line 149: Case-insensitive `data:` scheme matching is new here and applies to `image_url` too, not just audio/video: pre-PR, `DATA:image/png;base64,..`. hit `strings.HasPrefix(ref.url, "data:")`, failed the match, and got sent down the download path where it was rejected (scheme guard). Now it's accepted as inline data. Worth confirming this is an intended side effect of introducing `isDataURI` rather than an incidental scope creep beyond #2648 - and if intended, worth a line in the release note, since it's a behavior change for existing image callers, not just new audio/video paths.

Worth validating with the OpenAI/vllm that it is really case-insensitive.

pkg/coordinator/steps/render.go
```diff
@@ -270,33 +270,57 @@ func (s *RenderStep) executeChatCompletions(ctx context.Context, reqCtx *pipelin
...
-		reqCtx.MultimodalEntries[i].KwargsData = imageKwargs[i]
-		reqCtx.MultimodalEntries[i].Placeholder = imagePlaceholders[i]
+		mod := entryModality(reqCtx.MultimodalEntries[i], logger)
+		idx := modIndex[mod]
+		modIndex[mod]++
+		hashes := renderResp.Features.MMHashes[mod]
+		placeholders := renderResp.Features.MMPlaceholders[mod]
+		kwargs := renderResp.Features.KwargsData[mod]
+		if idx >= len(hashes) || idx >= len(placeholders) || idx >= len(kwargs) {
+			return fmt.Errorf("render response missing slot %d for modality %s", idx, mod)
```
roytman, line 314: `TestRenderStep_ChatCompletions_MultipleModalities` (`render_test.go:122`) uses matched per-modality counts throughout. This guard catches a case that slips past the aggregate-count check above it: total hash/placeholder/kwargs counts equal len(entries), but the per-modality split is wrong (e.g. 2 entries: one image, one audio, but the response tags both response hashes as audio). Worth a case constructing exactly that.

roytman (commented):
Thank you for all your fixes. 
Several more inline comments.

A few spots restate the same fact rather than stating it once (CLAUDE.md: "State each fact once, in its canonical location"):

Peak-memory model: derived and restated across `config/coordinator/coordinator.yaml:134-176`, then again in the doc comments on `downloadSizeFor`(`replace_media_urls.go:691-700`), `enforceInlineSize` (`replace_media_urls.go:476-489`), and `enforceDownloadContentType` (`replace_media_urls.go:669-689`).
Per-modality coverage table: `coordinator.yaml:168-175`, restated at :178-196.
Format-to-MIME table: `coordinator.yaml:237` and `replace_media_urls.go:833-848`.
Also, two lines read as aphorisms rather than plain statements — (`replace_media_urls.go:76`) and (`replace_media_urls.go#L116`). Worth a pass to pick one canonical location per fact and have the others point back to it, per the "State each fact once" convention.

pkg/coordinator/steps/encode_prefill_flow_test.go
```diff
@@ -71,8 +71,8 @@ func TestEncodeToPrefill_ECTransferParamsFlow(t *testing.T) {
 		Model:     "llama-3",
 		TokenIDs:  []int{1, 32000, 32000, 32000, 32000, 32000, 32000, 2345},
 		MultimodalEntries: []pipeline.MultimodalEntry{
-			{Index: 0, Hash: "img-hash-1", KwargsData: "dDE=", Placeholder: pipeline.PlaceholderRange{Offset: 1, Length: 3}},
-			{Index: 1, Hash: "img-hash-2", KwargsData: "dDI=", Placeholder: pipeline.PlaceholderRange{Offset: 4, Length: 3}},
+			{Modality: ModalityImage, Hash: "img-hash-1", KwargsData: "dDE=", Placeholder: pipeline.PlaceholderRange{Offset: 1, Length: 3}},
```
capri-xiyue, line 129: Is it worthwhile to add video and audio test case here?

pkg/coordinator/steps/replace_media_urls.go
```diff
@@ -44,10 +45,118 @@ import (
...
+// mediaRef (the caller supplies msgIdx and partIdx). ok is false when the part
+// is missing those fields, and reason then names what was wrong so the caller
+// can log it. partType must already be known to be a media type; modality is
+// the one partTypeModality maps it to.
+//
+// This is the one definition of the well-formedness rule.
+// mediaPartIsWellFormed wraps it for walkers that need only the yes/no answer,
+// so every walker accepts and rejects exactly the same parts. See
+// mediaPartIsWellFormed for why that has to hold.
+func classifyMediaPart(partMap map[string]any, partType, modality string) (ref mediaRef, ok bool, reason string) {
```
capri-xiyue, line 79: Instead of return a reason? Will it be better to return an error?

pkg/coordinator/steps/utils.go
```diff
@@ -110,46 +112,87 @@ func capSingleTokenOutput(body map[string]any, format gateway.RequestFormat) {
 
 // buildMMFeatures builds the multimodal features map (mm_hashes, mm_placeholders,
 // and optionally kwargs_data) from the request's multimodal entries. It returns
-// nil when there are no entries.
-func buildMMFeatures(entries []pipeline.MultimodalEntry, includeKwargs bool) map[string]any {
+// nil when there are no entries. Entries are grouped by Modality so a
+// mixed-modality request produces one key per modality in each feature map.
+func buildMMFeatures(entries []pipeline.MultimodalEntry, includeKwargs bool, logger logr.Logger) map[string]any {
 	if len(entries) == 0 {
```
capri-xiyue, line 117: May need to pay attention to https://github.com/llm-d/llm-d-router/pull/2723/changes#diff-7378c1f0ddb565f20e9e51b7ba6c3f8c2db9551ea0c8d26856a0921a1ec80c0f as conflicts may happen

pkg/coordinator/steps/replace_media_urls.go
dmitripikus: The allowlist check now runs inside `download()` on the response headers, before `io.ReadAll`.
Test is added.
Thanks!

pkg/coordinator/steps/utils.go
dmitripikus: Fixed. You were right that it should fail instead of guessing.

One correction: a client can cause this, not just a coordinator bug. On the generate path `features` comes from the request body, and the modality keys were used as-is, so `mm_hashes: {"": ["h1"]}` created an entry with no modality. That changes the error type, since a plain error becomes a 502 and would let a client trigger 502s.

So the fix has two parts. An empty modality key is now rejected as a bad request where the features map turns into entries. The `image` fallback is gone, and a new check runs at the start of each step, which fails the request the way you asked.

Thanks!

pkg/coordinator/steps/decode.go
dmitripikus: Added. `TestInjectUUIDs_ExtraPartForModalityStaysUntagged` uses 2 audio parts with 1 audio entry and checks the second part has no `uuid` key, so a wrong-hash tag fails the test. 

On the difference from encode: `uuid` is only the cache key, so an untagged part still sends its payload and the answer stays correct. The test comment now says that.

Thanks!

pkg/coordinator/steps/replace_media_urls.go
dmitripikus: It is an intended effect. The change has its own commit (6ba69c86) + test.

It came up because `audio_url` and `video_url` can now carry inline data as well, so the "is this inline data?" test had to become one shared function instead of a separate prefix check for images. While writing that function I made it ignore letter case.

I checked the model server: vLLM parses the URL and compares the scheme to `data`, and Python lowercases the scheme first, exactly as Go does. So `DATA:...` works there too, and we are not letting through something that breaks later.

The comment gave the wrong reason for the old failure. Fixed in 365dba6b.

I added a line to the release note in the description, since this changes behavior for existing image callers too.

Thanks!

pkg/coordinator/steps/render.go
dmitripikus: Added `TestRenderStep_ChatCompletions_RejectsWrongModalitySplit`, with your case (two entries, both response hashes tagged audio) plus two more. Turns out neither check in that block had any test, not just the per-entry one, so there is also a case for the count check to make the difference between the two visible.

Thanks!

pkg/coordinator/steps/encode_prefill_flow_test.go
dmitripikus: Added one image, one audio, and one video entry through both steps. Each step is already tested alone; what was missing is that encode keys EC params by hash while prefill groups features by modality, and nothing checked the two agree for a mixed request.
The fake encoder needed fixing first, since it read the image key from every sub-request and would have panicked on an audio one.

Thanks!

pkg/coordinator/steps/replace_media_urls.go
dmitripikus: I would rather keep the bool here, and I added a comment in bc37ab52 saying why, since the old one explained what the returns mean but not that.

A part failing this check is not a failure we need to handle. It is passed through untouched and the request still succeeds, so there is nothing for a caller to decide. An error would also be risky: four places ask this same question, and all four have to skip the part, because entries and media parts are matched up by position further down the pipeline. An error return reads as license to fail the request, and one caller failing where the others skip would put the two lists out of step and attach the wrong media to the wrong hash.

pkg/coordinator/steps/utils.go
dmitripikus: Yes, thank you, you are right, they do conflict. Both PRs merge cleanly there today, so it is whichever lands second that pays.

The overlap is that #2723 is written on main, where the feature lists are image-only, and this PR replaces those with per-modality lists. In render.go #2723 edits the loop that reads MMHashes[ModalityImage] and fills each entry from it; this PR removed that loop in favor of a per-modality walk. In utils.go, the mmImageArray and mmKwargsField that #2723 builds on no longer exist here. And their context.go doc comment names three fields this PR deleted.

I would suggest that current one goes first and #2723 rebases: 67 commits against 1, and metadata keyed only under image would be wrong as soon as an audio or video entry exists, which is what this PR adds. So rebasing makes their change correct rather than just merged.

I'll notify on #2723 too, so its author sees it before rebasing.

Thanks!

pkg/coordinator/steps/replace_media_urls.go
```diff
@@ -232,26 +410,103 @@ func (s *ReplaceMediaURLsStep) Execute(ctx context.Context, reqCtx *pipeline.Req
...
+// Audio and video are enforced so the per-modality cap holds however the
+// payload arrived. coordinator.yaml's max_download_size comment records the
+// memory bound these caps set. input_audio is only ever inline and is capped by
+// validateInlineAudio, so enforcing here keeps the two ways of sending the same
+// audio in agreement.
+//
+// Images are exempt, on the same grounds as their exemption from the
+// download-path Content-Type check in enforceDownloadContentType.
+// server.max_request_body_size is what bounds an image data URI.
+func (s *ReplaceMediaURLsStep) enforceInlineSize(modality string) bool {
```
roytman, line 482: `enforceInlineSize` exempts images unconditionally, so an explicit `max_image_download_size` has no effect on inline image data.

```go
// Images are exempt, on the same grounds as their exemption from the
// download-path Content-Type check in enforceDownloadContentType.
// server.max_request_body_size is what bounds an image data URI.
func (s *ReplaceMediaURLsStep) enforceInlineSize(modality string) bool {
	return modality != ModalityImage
}
```

The comment claims parity with `enforceDownloadContentType`, but that function's actual image behavior is conditional, not a flat exemption:

```go
// Images are enforced only when the operator set allowed_image_content_types
// explicitly.
...
func (s *ReplaceMediaURLsStep) enforceDownloadContentType(modality string) bool {
	...
	_, explicit := s.contentTypeOverrides[ModalityImage]
	return explicit
}
```

`enforceDownloadContentType` turns strict enforcement on for images only when the operator explicitly configured `allowed_image_content_types`. `enforceInlineSize` has no such opt-in: it is `modality != ModalityImage`, full stop, regardless of whether `max_image_download_size` was set.

Concretely, at the call site:

```go
if s.enforceInlineSize(ref.modality) && s.inlineSizeExceeded(b64, ref.modality) {
```

`enforceInlineSize("image")` is always `false`, so `inlineSizeExceeded` never runs for an image's inline `data:` URI. `downloadSizeFor(modality)` (the configured cap) is still applied to images on the HTTP-download path (`download()`), so the setting is not completely inert. It just silently does nothing for the inline case.

`coordinator.yaml` documents the *unset* default for `max_image_download_size` (falls back to `server.max_request_body_size`), but nothing documents that an explicit value is ignored for image `data:` URIs specifically. An operator who sets `max_image_download_size` to bound memory would reasonably expect it to apply the same way it does for audio and video.

Suggested fix, mirroring `enforceDownloadContentType`'s pattern:

```go
func (s *ReplaceMediaURLsStep) enforceInlineSize(modality string) bool {
	if modality != ModalityImage {
		return true
	}
	_, explicit := s.maxDownloadSizeByMod[ModalityImage]
	return explicit
}
```

This keeps the current default (image inline data unchecked, relying on `server.max_request_body_size`) but makes an explicit `max_image_download_size` actually take effect for inline data too, not just HTTP downloads.

P.S. We have a very simple one-line function and 13 lines of comments ;-)

roytman (commented):
@dmitripikus pls resolve the conflicts 
and optionally a couple of non-blocking nits. 

nit: docs/coordinator_architecture.md:737-738] This still describes replace-media-urls as image-only ("Download image_url references", "per-image hash/placeholder/kwargs"), and the param table doesn't list any of the six new per-modality params this PR adds. Since this is the canonical description of the step being rewritten here, it's worth updating alongside the code rather than in a follow-up.

pkg/coordinator/steps/replace_media_urls.go
dmitripikus: Fixed as you suggested. It opts in rather than always enforcing because max_download_size has never applied to inline images — turning it on there would reject requests current configs accept. max_image_download_size is new here, so keying on it is safe.

config/coordinator/coordinator.yaml
```diff
@@ -152,36 +155,162 @@ pipeline:
...
+        # Content-Type at all. Setting the param is read as a deliberate
+        # lockdown and covers both paths. Audio and video downloads are
+        # always checked; see max_audio_download_size for why they are
+        # treated more strictly.
+        # Default: image/jpeg, image/png, image/gif, image/webp.
+        # allowed_image_content_types:
+        #   - image/jpeg
+        #   - image/png
+
+        # allowed_audio_content_types overrides the built-in audio MIME
```
roytman, line 282: Here and for `allowed_video_content below. 

Recommended: add the missing half-sentence to the yaml, mirroring what's already there for images.

The image comment at `coordinator.yaml:270-274` already says "leaving this unset differs from listing the defaults... plenty of origins serve a valid image as application/octet-stream." Audio/video need the equivalent warning next to their own knobs - currently don't mention it at all. Concretely, add one line to the allowed_audio_content_types / allowed_video_content_types comment blocks:

```go
# allowed_audio_content_types overrides the built-in audio MIME
# allowlist. Applies to audio_url data URIs, to the Content-Type an
# http(s) audio_url origin returns, and to input_audio parts. Unlike
# image, the download path is checked whether or not this param is set --
# including against S3/GCS/presigned-URL origins that serve every object
# as application/octet-stream by default. If audio/video downloads from
# such an origin get rejected as "content type not allowed", either set
# this to an empty list to accept any type, or add
# "application/octet-stream" to the list explicitly.
```
Same addition, adapted, on the video block.

**Why not change the code instead:** adding `application/octet-stream` to the default allowlists would make the default silently accept literally any file type for audio/video, since "the origin declared nothing useful" and "the origin is actively lying" produce the identical header. That defeats the declared-mismatch detection the PR's own security section argues for.

pkg/coordinator/steps/replace_media_urls.go
```diff
@@ -155,32 +310,45 @@ func (s *ReplaceMediaURLsStep) Execute(ctx context.Context, reqCtx *pipeline.Req
...
-			if partMap["type"] != imageURLPartType {
+			partType, _ := partMap["type"].(string)
+			modality, isMedia := partTypeModality[partType]
+			if !isMedia {
 				continue
 			}
-			imageURL, ok := partMap[imageURLPartType].(map[string]any)
+			ref, ok, reason := classifyMediaPart(partMap, partType, modality)
 			if !ok {
+				logger.V(logutil.DEBUG).Info("skipping malformed media part",
```
roytman, line 320: I'd increase the log level to `Info`
```go
logger.Info("skipping malformed media part",
        "reason", reason, "msg_index", msgIdx, "part_index", partIdx, "part_type", partType)
```

pkg/coordinator/steps/prefill_test.go
```diff
@@ -759,3 +760,44 @@ func TestPrefillStep_CoercesInvalidKVTransferParams(t *testing.T) {
 		})
 	}
 }
+
+// TestPrefillStep_EntryWithoutModalityFails covers the invariant guard at a step
```
roytman, line 764: `validateEntryModalities` is wired into five steps `Execute` (encode, decode, render, prefill, conditional_decode), but only prefill has a wiring test (`TestPrefillStep_EntryWithoutModalityFails`) proving the step actually calls the guard and stops before the upstream server. The other four, including `conditional_decode` which this PR didn't otherwise touch, have no equivalent. A regression that dropped or moved the guard in any of those four would pass CI today. Suggest one `TestXStep_EntryWithoutModalityFails` per step, with the same shape as the prefill one.

roytman (commented):
Several small (hopefully last) inline comments plus
 `docs/communication.md:108-110` - This still shows `MultimodalEntry{Index, Base64Data, ContentType}` - this PR removes all three fields. `coordinator_architecture.md` got updated for the new shape; this one was missed. Suggest {Modality: "image"} / {Modality: "audio"} and a note that the bytes stay on the request body.

pkg/coordinator/steps/replace_media_urls.go
dmitripikus: Raised the level, but to V(logutil.DEFAULT) rather than bare Info.

logutil.DEFAULT is 2 and is also the default verbosity, so it shows without anyone raising the log level, which is what you asked for. Bare Info is V(0), one notch above the step's own "complete" line — a skipped part should not be louder than that. It is also the common form here: 24 call sites use V(logutil.DEFAULT).Info against 5 bare Info.
If you meant V(0) specifically, I'll update.

Thanks!

pkg/coordinator/steps/prefill_test.go
dmitripikus: Added all four -  same shape as the prefill one.
To check they test the wiring, I removed each guard in turn and made sure the matching test fails. All four do.
The encode test uses the generate format for that reason. On the chat-completions format the entry fails later at part pairing, so the test passed with the guard gone.

Thanks!

dmitripikus:
Regarding @roytman 's comment for docs/communication.md:108-110
Fixed. Stage 1 now shows {Modality: "image"}, plus a mixed-request block, and says the bytes stay on the request body.

I fixed the other stale spots too: render pairs per modality now, encode fans out per entry, and the overview diagram and text-only section no longer say image only.

The examples still use images, with one line saying audio and video use their own part types and keys.

Thanks!

config/coordinator/coordinator.yaml
dmitripikus: Addressed as you proposed.

Thanks!

pkg/coordinator/steps/replace_media_urls.go
dmitripikus: Changed to bare Info level.

roytman (dismissed):
Thank you.

ahg-g:
@dmitripikus what is the status of this PR, should we move it forward?

PR #2664, opened 2026-09-02
feat: add custom TLS support for vLLM render endpoint

zdtsw:
**What type of PR is this?**
/kind feature

Allows the `token-producer` plugin to connect to render endpoints over HTTPS using a custom CA bundle or mTLS client certificates for clusters with private PKI.

related to https://github.com/llm-d/llm-d-router/issues/2677 but not exatly for sidecar

```release-note
The token-producer plugin now supports HTTPS for the vLLM render endpoint. Configure `vllm.caCertPath` for custom CA verification, `vllm.clientCertPath`/`vllm.clientKeyPath` for mTLS, or `vllm.insecureSkipVerify` to skip server certificate verification.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
@@ -20,12 +20,14 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"maps"
 	"net/http"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
```diff
@@ -120,7 +140,44 @@ func newRenderTransport() *http.Transport {
...
+	if cfg.hasTLS() {
+		tlsCfg, err := renderTLSConfig(cfg)
+		if err != nil {
+			return nil, err
+		}
+		t.TLSClientConfig = tlsCfg
+	}
+	return t, nil
+}
+
```
vMaroon, line 153: nit: `hasTLS()` plus the `if cfg.hasTLS()` gate is a nil-vs-empty optimization only. `renderTLSConfig(cfg)` with no TLS fields returns an empty `*tls.Config` (nil `RootCAs`, `InsecureSkipVerify` false) that behaves identically to leaving `TLSClientConfig` nil, and the transport already disables HTTP/2. Could just call `renderTLSConfig(cfg)` unconditionally and drop the helper. Non-blocking.

vMaroon (approved):

Issue #2390, opened 2026-08-15
EPP precise prefix cache: request-path CPU and allocations at fleet scale


nilig:
The precise-prefix-cache request path does work proportional to
`blockKeys x candidate pods` in string-keyed maps and materialized
slices, plus trace-string construction that runs even when TRACE is
disabled. At agentic-workload scale (240K-token prompt, `blockSize: 64`
= 3,750 block keys, 96 candidate pods, warm index) one scoring pass
costs **674 ms and 13 GB allocated** on an M4 Max
(`go test -bench BenchmarkScoreTokens`); the same shape at 8 pods costs
82 ms / 1.2 GB. Event ingestion and every scored request pay related
costs. Profiled with pprof; each item below is a separate mechanism.

1. **Eager trace construction.** `Indexer.ScoreTokens` and
   `InMemoryIndex.Lookup` build `podsPerKeyPrintHelper(...)` strings as
   log arguments unconditionally; the `ScoreTokens` variant concatenates
   with `+=`, quadratic in keys. This is 652 ms of the 674 ms.
   `InMemoryIndex.Add` boxes per-key trace args on the ingestion path.
2. **Lookup materialization and string-keyed folds.** `Lookup` returns
   `map[BlockHash][]PodEntry` built from per-key LRU `Keys()` calls
   (~18.9 MB per request at 96 pods) which the longest-prefix scorer
   immediately folds through maps keyed by `PodIdentifier` strings
   (~360K string hashes per request); the producer then re-walks the
   same map per endpoint for `matchedBlockCount`/`ByTier`.
3. **Block hashing.** Each block marshals `[parent, tokens, extra]`
   through the CBOR encoder and hashes the buffer, on top of per-block
   chunk slices: 2.77 ms / 1.62 MB / 18,775 allocs per 240K-token
   derivation.
4. **Dead `podLabelSelector`.** `PodDiscoveryConfig.PodLabelSelector` is
   parsed into the config but ignored on the endpoint-notification
   path, so the EPP subscribes to every pool endpoint - decode ranks
   included - and indexes events precise prefill routing never reads.
5. **Chat render round trip.** Non-map chat bodies (gRPC, warmup) and
   every Anthropic Messages request are projected to the typed render
   request, marshaled to JSON, unmarshaled into a generic map, and
   marshaled again for the wire - three JSON passes per request body.
6. **No cancellation checks** in the multi-thousand-iteration index
   lookup/match loops.

[...]

PR #2669, opened 2026-09-03
refactor(kvcache): fold prefix matching into one accumulator

nilig:
**What type of PR is this?**
/kind cleanup

Longest-prefix matching is implemented three times: `LongestPrefixScorer` over a materialized `Lookup` for the weighted score, and the precise-prefix producer's `matchedBlockCount` and `matchedBlockCountByTier` for the block and per-tier counts it publishes, re-walking the lookup result once per endpoint. No single place defines the contract, and the re-walk grows with fleet size.

`Indexer.MatchBlockKeys` becomes the one entry point. A pooled accumulator implements the rules once (candidates are the pods holding the first key, a chain ends at the pod's first missing key, duplicate entries at a key take the highest weight, every tier tracks its own chain) and folds each key's entries in one call. `ScoreTokens` and `LongestPrefixScorer.Score` are projections of it; the producer only aggregates per prompt and publishes, all or nothing. `KVBlockScorer` and `NewTracedScorer` stay as deprecated compatibility surfaces; existing `score_tokens` span attributes keep their meaning, and the matcher's own `match_block_keys` span reports the longest chain.

Ownership after this change: kvblock owns storage and reads; kvcache owns the one matcher, its fallback, and match telemetry; the producer owns prompt aggregation and publication. Index reads are unchanged: this PR materializes `Lookup`. A follow-up adds an ordered index walk that feeds the same accumulator without materializing.

AMD EPYC 9655P node, 16 exclusive CPUs, test binaries built once, six paired rounds in AB/BA order, n=6, per request:

| Row | main | this PR |
|---|---|---|
| `ScoreTokens`, 240K tokens, 96 pods | 1,185.5ms / 12.2 GiB | 68.9ms / 90.7 MiB |
| `ScoreTokens`, multi-rank shape (1,034 blocks held by 40 endpoints with eight rank entries each) | 551.1ms / 4.6 GiB | 85.6ms / 111 MiB |
| precise producer `Produce`, serial | 152.2ms / 163 MiB | 94.4ms / 143 MiB |
| precise producer `Produce`, 16-way | 37.3ms | 22.5ms |

The materialized `Lookup` owns what remains; the follow-up walk is #2680.

Part of #2390, part of #2670

**Release note**:
```release-note
The precise prefix-cache producer and `kvcache.Indexer.ScoreTokens` share one prefix matcher (`Indexer.MatchBlockKeys`), which emits a `match_block_keys` span. `ScoreTokens` no longer emits `compute_scores`; `kvcache.KVBlockScorer` and `kvcache.NewTracedScorer` are deprecated.
```

Test plan:
- [x] matcher parity with the replaced scorer and producer counters on 500 random fixtures (filters, gaps, duplicate pod and rank entries, unconfigured tiers, speculative entries)
- [x] producer aggregation across prompts, multimodal match, candidate filtering, error and cancellation publish nothing
- [x] `score_tokens` telemetry keeps its meaning (span test)
- [x] `make presubmit`


pkg/kvcache/kvblock_scorer.go
```diff
@@ -45,8 +45,11 @@ func DefaultKVBlockScorerConfig() *KVBlockScorerConfig {
 	}
 }
 
-// KVBlockScorer defines the interface for implementing a KV block scoring
-// strategy.
+// KVBlockScorer scores a materialized lookup result.
+//
+// Deprecated: the prefix matcher (Indexer.MatchBlockKeys) is the single
```
vMaroon, line 50: Can we track this for removal

pkg/kvcache/kermit_fixture_bench_test.go
```diff
@@ -0,0 +1,131 @@
...
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/llm-d/llm-d-router/pkg/kvcache"
+	"github.com/llm-d/llm-d-router/pkg/kvcache/kvblock"
+)
+
+// Kermit-shaped fixture: 1,034 blocks held by 40 endpoints, each endpoint
```
vMaroon: Kermit should be dropped from these files IMO.

vMaroon (commented):
This looks great 💯 

cc @bongwoobak @yankay for more eyes

pkg/kvcache/prefix_match.go
```diff
@@ -0,0 +1,419 @@
...
+	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
+		return nil, 0, err
+	}
+	matches, err = matchMaterialized(ctx, keys, keyToPods, k.tierWeights, podFilter)
+	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
+		return nil, 0, err
+	}
+	keysFound = len(keyToPods)
```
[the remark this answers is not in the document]
vMaroon: This is right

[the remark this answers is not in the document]
nilig: Fixed in c4e31fd7, counting only keys with at least one entry. This restores what `ScoreTokens` did before the matcher took over the count, and #2670 commits that the `score_tokens` attributes keep their meaning, so the drift was mine.

On whether the shape is reachable: it is, though only under a race, and it predates this PR. `InMemoryIndex.Lookup` reads `cache.Len()` and `cache.Keys()` as separate locked calls without holding `PodCache.mu`, so an eviction of the last entry between them leaves the unfiltered branch assigning an empty slice. What changed here is that `len(keyToPods)` would have counted that as a hit, where the old path guarded on `len(pods) > 0`. The stub index in the test makes the case deterministic rather than inventing it.

Tracked the `Lookup` window itself as #2698 rather than folding it in here. Correcting something I first got wrong there: #2680 already closes it, since its rewritten `Lookup` reads the entries and their count in one critical section under `PodCache.mu` and stores the result only when the filtered slice is non-empty. So the issue is a record for the current head rather than work to schedule, and a standalone patch would only collide with that rewrite.

pkg/kvcache/kermit_fixture_bench_test.go
nilig: Agreed, dropped in c4e31fd7. The files are now `multi_rank_prefix_bench_test.go` and `multi_rank_scoretokens_bench_test.go`, with `BenchmarkMatchBlockKeysMultiRank`, `BenchmarkMatchBlockKeysMultiRankParallel` and `BenchmarkScoreTokensMultiRank`, and the identifiers and prose went with them. The PR description carried it too, so that is reworded.

The fixture itself is unchanged: 1,034 blocks held by 40 endpoints with eight rank entries each, so every block carries 320 entries collapsing to 40 matched pods. That shape is what separates the three variants, so it seemed worth keeping under a name that describes it.

pkg/kvcache/kvblock_scorer.go
nilig: Tracked in #2697. Keeping the removal out of this PR since #2670 commits that exported API is deprecated and not removed within a release line, so `KVBlockScorer`, `NewKVBlockScorer`, `LongestPrefixScorer`, `NewTracedScorer`, `KVBlockScorerConfig` and the compatibility config path in `indexer.go` go together at the next boundary rather than piecemeal. The issue notes that dropping `KVBlockScorerConfig` changes the accepted `indexerConfig` shape, which is the user-facing part worth checking at that point.

vMaroon (approved):

PR #2676, opened 2026-09-03
feat(sidecar): add TLS security profile flags

### LINKED ISSUE #2677: Expose TLS security profile flags on pd-sidecar

Add `--tls-min-version` and `--tls-cipher-suites` options to the pd-sidecar secure proxy, with equivalent sidecar YAML configuration fields. Values should use Go `crypto/tls` names and invalid values should fail during startup.

The defaults should preserve the pd-sidecar's existing minimum TLS version and cipher suite list when these options are omitted.

**Why is this needed**:

The EPP server already exposes these TLS security profile controls, but the pd-sidecar does not. Deployments that apply a consistent TLS policy to all LLM routing components therefore cannot configure the pd-sidecar equivalently.

KServe is adding these settings to its LLMInferenceService templates in https://github.com/kserve/kserve/pull/5924, and its image CLI validation identified that the current pd-sidecar image does not expose the required flags.

andresllh:
**What type of PR is this?**

/kind feature

Adds `--tls-min-version` and `--tls-cipher-suites` flags to the pd-sidecar secure proxy, matching the TLS security profile controls already available on the EPP server.

The options are also supported in sidecar YAML configuration. Invalid TLS versions and cipher suite names fail during option completion. Existing TLS 1.2 and cipher suite defaults remain unchanged when the new options are omitted.

Validation:
- `go test ./pkg/sidecar/proxy -count=1`
- `make presubmit`

Fixes #2677

Related downstream change: https://github.com/kserve/kserve/pull/5924

```release-note
Add pd-sidecar flags for configuring the minimum TLS version and TLS 1.2 cipher suites.
```


zdtsw (commented):
following the same design in PR2230, LGTM

pkg/sidecar/proxy/options.go
```diff
@@ -354,6 +365,44 @@ func validateStages(stages []string, supportedStages map[string]struct{}, flagNa
 	return nil
 }
 
+var tlsVersions = map[string]uint16{
```
elevran, line 366: nit: 
`k8s.io/component-base/cli/flag` is already a direct dependency in go.mod and provides `TLSVersion()` and `TLSCipherSuites()` parsing and validation equivalents, used by kube-apiserver's flags use. 
Since `pkg/epp/server/options.go` already hand-rolls this same map and functions, so not asking to change it here, but worth a followup issue to consolidate both onto the imported package implementation instead of maintaining two copies.

elevran (commented):
Thanks! this is a clean PR, and closely mirrors the existing EPP TLS profile handling.

One doc gap outside this diff: 
docs/disaggregation.md's "General Sidecar Flags" table (lines 684-687) lists `--enable-tls` and `--tls-insecure-skip-verify` but hasn't been updated with the two new flags this PR adds. Worth adding rows there for discoverability.

pkg/sidecar/proxy/options.go
andresllh: created the issue: https://github.com/llm-d/llm-d-router/issues/2783

elevran (approved):

PR #2678, opened 2026-09-03
fix(epp): support vLLM render endpoints started with --api-key

### LINKED ISSUE #2671: token-producer cannot authenticate to vLLM /render endpoints when --api-key is enabled
**What happened**:

I am running vLLM with API key authentication enabled:

```yaml
args:
  - /models/Qwen3.6-35B-A3B
  - --host
  - 0.0.0.0
  - --port
  - "8000"
  - --api-key
  - $(VLLM_API_KEY)
  - --served-model-name
  - Qwen/Qwen3.6-35B-A3B
```

The EPP token-producer uses the vLLM HTTP render backend:
```
- type: token-producer
  parameters:
    modelName: Qwen/Qwen3.6-35B-A3B
    vllm:
      url: http://qwen36-35b-a3b-vllm.llm-d-agentgateway.svc.cluster.local:8000
      timeout: 5s
      mmTimeout: 30s
```
When a chat completion request is processed, token-producer sends a separate request to the vLLM /v1/chat/completions/render endpoint. This request does not contain an Authorization header, so vLLM returns HTTP 401:
```
{"error":"Unauthorized"}
```
As a result, tokenization fails and downstream functionality that depends on accurate token IDs, such as precise prefix-cache routing, cannot work correctly.

The vLLM HTTP render backend should support authenticating to vLLM instances that are started with --api-key.
Preferably, the API key should be loaded from an environment variable or Kubernetes Secret rather than stored directly in EndpointPickerConfig, because the plugin configuration is commonly rendered into a ConfigMap.
For example, a possible configuration could be:
```
- type: token-producer
  parameters:
    modelName: Qwen/Qwen3.6-35B-A3B
    vllm:
      url: http://qwen36-35b-a3b-vllm.llm-d-agentgateway.svc.cluster.local:8000
      timeout: 5s
      mmTimeout: 30s
      apiKeyEnv: VLLM_API_KEY
```
The EPP container could receive the secret through its environment:
```
router:
  epp:
    env:
      - name: VLLM_API_KEY
        valueFrom:
          secretKeyRef:
            name: vllm-api-key
            key: api-key
```
When configured, token-producer should add the following header to both render requests:
```
Authorization: Bearer <API key>
```
This should apply to:
- POST /v1/completions/render
- POST /v1/chat/completions/render

1. Start a vLLM server with API key authentication enabled:
```
vllm serve Qwen/Qwen3.6-35B-A3B \
  --port 8000 \
  --api-key "${VLLM_API_KEY}" \
  --served-model-name Qwen/Qwen3.6-35B-A3B
```
2. Configure the EPP with a token-producer that points to this server:
```
- type: token-producer
  parameters:
    modelName: Qwen/Qwen3.6-35B-A3B
    vllm:
      url: http://vllm.default.svc.cluster.local:8000
      timeout: 5s
      mmTimeout: 30s
```

3. Send a chat completion request through the inference gateway:
```
curl http://<gateway-address>/v1/chat/completions \
  -H "Authorization: Bearer ${VLLM_API_KEY}" \
  -d '{
    "model": "Qwen/Qwen3.6-35B-A3B",
    "messages": [
      {
        "role": "user",
        "content": "Hello"
      }
    ]
  }'
```
4. Check the EPP logs.
5. Observe that the internally generated render request returns HTTP 401:

no 
**Environment**:
- Kubernetes version (use `kubectl version`):  v1.33.6
- llm-d-router version (use `git describe --tags --dirty --always` if you built from source, or specify the tag if you used a tagged version or image):  v0.9.0
- Install tools: 
- Others:

Iceber:
**What type of PR is this?**

/kind bug

The token-producer's vLLM HTTP render backend sends `/v1/completions/render` and `/v1/chat/completions/render` requests without an `Authorization` header. A vLLM instance started with `--api-key` rejects them with 401, so tokenization fails and every consumer of token IDs (precise prefix-cache routing among them) silently degrades.

The render backend forwards the inbound request's `Authorization` header verbatim on render requests. The render endpoint serves the same model as the pods the request is routed to, so the credential the client presents for inference is the credential the render endpoint expects. Requests without an `Authorization` header send none. 

The warmup probe runs at plugin load with no inbound request and therefore no credential. Against an endpoint started with `--api-key` it gets 401, treats it the same as "backend not yet reachable", and spends the full retry budget (24 attempts, 5s apart) before logging the generic `warmup did not complete` at DEBUG. An authentication rejection cannot clear on retry, so the probe returns on the first 401/403 and logs at Info that warmup was skipped and the first request pays the cold-start cost. `postJSON` returns a typed `renderStatusError` so the status code is available to the caller; the error text is unchanged.

Warmup is still not possible against an authenticated render endpoint. Supporting that would require a plugin-level credential (for example `vllm.apiKeyEnv`) used only by the warmup probe; that is left for a follow-up.

Fixes https://github.com/llm-d/llm-d-router/issues/2671

```release-note
token-producer forwards the inbound Authorization header on vLLM render requests, so render endpoints started with --api-key are supported. The startup warmup probe skips such endpoints instead of exhausting its retry budget.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
@@ -65,14 +65,23 @@ type warmer interface {
 
 // warmup primes the render path so the first request does not pay the cold-start
 // cost. It retries a text render until the backend responds, then issues a
-// best-effort multimodal render. It returns on success, on the attempt cap, or
-// on context cancellation.
+// best-effort multimodal render. It returns on success, on the attempt cap, on
+// an authentication rejection, or on context cancellation.
 func (b renderBackend) warmup(ctx context.Context) {
-	logger := log.FromContext(ctx).V(logutil.DEBUG)
+	logger := log.FromContext(ctx)
 	for i := 0; i < warmupAttempts; i++ {
-		if _, err := b.produce(ctx, warmupChat()); err == nil {
+		_, err := b.produce(ctx, warmupChat())
+		if err == nil {
 			_, _ = b.produce(ctx, warmupChat(warmupImage))
-			logger.Info("token-producer backend warmed up", "attempts", i+1)
+			logger.V(logutil.DEBUG).Info("token-producer backend warmed up", "attempts", i+1)
+			return
+		}
+		// Warmup carries no credentials; an auth rejection will not clear on retry.
+		if isRenderAuthError(err) {
+			logger.V(logutil.DEFAULT).Info(
+				"token-producer backend requires authentication, skipping warmup; "+
+					"the first request pays the cold-start cost",
+				"err", err)
 			return
 		}
 		select {
@@ -81,7 +90,7 @@ func (b renderBackend) warmup(ctx context.Context) {
 			return
 		}
 	}
-	logger.Info("token-producer backend warmup did not complete")
+	logger.V(logutil.DEBUG).Info("token-producer backend warmup did not complete")
 }
 
 // warmupChat builds a single-message chat body carrying the given image URLs.
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/README.md
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/README.md
@@ -106,7 +106,10 @@ The plugin calls `POST {http}/v1/completions/render` and
 `vllm serve <model>` and by the GPU-less `vllm launch render <model>`.
 Any reachable HTTP endpoint serving the same model the scheduler tokenizes
 for will work — sidecar in the EPP pod (loopback) or a dedicated Service
-shared by multiple EPP replicas.
+shared by multiple EPP replicas. When the inbound request carries an
+`Authorization` header, it is forwarded verbatim on render requests, so an
+endpoint started with `--api-key` accepts them; the startup warmup probe
+sends no `Authorization` header.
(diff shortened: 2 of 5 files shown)


pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
```diff
@@ -65,14 +65,20 @@ type warmer interface {
...
+		_, err := b.produce(ctx, warmupChat())
+		if err == nil {
 			_, _ = b.produce(ctx, warmupChat(warmupImage))
-			logger.Info("token-producer backend warmed up", "attempts", i+1)
+			logger.V(logutil.DEBUG).Info("token-producer backend warmed up", "attempts", i+1)
+			return
+		}
+		// Warmup carries no credentials; an auth rejection will not clear on retry.
+		if isRenderAuthError(err) {
+			logger.Info("token-producer backend requires authentication, skipping warmup; the first request pays the cold-start cost", "err", err)
```
ahg-g: pls use a specific logging level

ahg-g (commented):
@vMaroon  fyi

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
Iceber: updated, Thanks!

vMaroon (approved):
Thanks for catching this.

Issue #2670, opened 2026-09-03
Precise prefix matching: separation of concerns and refactor plan


nilig:
Precise prefix matching is implemented in three places that each read the same rules, and the fused `ScoredLookupIndex` in #2402 fixed the cost by putting scoring policy inside the index, which gave the algorithm several owners (https://github.com/llm-d/llm-d-router/pull/2402#issuecomment-5462210694). This issue is the end-to-end plan for the refactor that replaces it, so each PR can be reviewed against one design. The cost measurements are in #2390; this issue is about structure.

## Separation of concerns

One owner per layer on the matching read path: `kvblock` hands out entries, `kvcache` matches, the producer aggregates and publishes. On that path each layer uses the one below it and must not know the one above. The producer keeps its direct `kvblock.Index` use for KV-event ingestion and speculative writes; those are storage operations, not matching, and this issue does not change them.

| Layer | Owns | Must not know |
|---|---|---|
| `kvblock` | storage; ordered access to the requested keys (`Index.Lookup`, optional `KeyWalker`); entry identity and the pod/tier ordinals it hands out; per-key locking; LRU recency; storage telemetry (requests, latency, keys present) | tier weights, scores, chains, the speculative tier name, endpoints |
| `kvcache` | the one prefix matcher (`Indexer.MatchBlockKeys` and its accumulator); choosing the walk when the backend has it and materialized `Lookup` otherwise; match telemetry (longest chain, hit metrics); `ScoreTokens` as a projection of the match | index internals, producer attributes |
| `preciseprefixcache` producer | calling the matcher once per prompt, summing across prompts, the multimodal match, publishing `PrefixCacheMatchInfo` all or nothing | any matching rule |

A new index backend supports the matcher by implementing `Lookup`; it gets the fast path by implementing `KeyWalker`. Neither requires knowing how scoring works.

## The contract, in one place

`MatchBlockKeys(ctx, keys, podFilter) -> map[pod]PodMatch{WeightedScore, MatchedBlocks, BlocksByTier}`:

1. Candidates are the pods holding the first key, after the filter.
2. A pod's chain ends at its first missing key.
3. Duplicate entries for one pod at one key take the highest weight; another rank of the same endpoint adds nothing.
4. Each tier tracks its own contiguous prefix.
5. Unconfigured tiers weigh 1.0.
6. Speculative entries count under the speculative tier.
[...]

PR #2680, opened 2026-09-03
perf(kvcache): match prefixes over an ordered index walk

nilig:
**What type of PR is this?**
/kind feature

The matcher materializes `Lookup`: every request copies the entries of every key it reads into a map before scoring, which at fleet scale is tens of megabytes and most of the request time (#2669 on the multi-rank shape: 84ms and 111 MiB per request on an EPYC node).

`kvblock.KeyWalker` is an optional `Index` capability that visits the requested keys in order and lends each key's records, with pod and tier ordinals, under that key's lock. Its contract is stated on the interface: borrowed-entry lifetime, no re-entry from `visit`, ordinal identity and stability across `Evict` and `Clear`, duplicate and missing positions, no filtering, per-key consistency without a cross-key snapshot, the cancellation polling points, and exactly which keys a walk promotes. `InMemoryIndex` implements it and stores `EntryRef` records, so nothing is hashed per entry; the traced and instrumented decorators expose the capability exactly when the wrapped index has it. `Indexer.MatchBlockKeys` walks when it can and materializes `Lookup` for other backends; both paths feed the same accumulator, and the property test checks both against the replaced algorithms. Request-local state is keyed by ordinal and never sized by it; the index caps the distinct pod identifiers and tiers it interns, and an `Add` past a cap fails before writing anything.

Recency stays exact and gets cheaper: the index owns the request-key LRU lock, reads peek under the shared lock, and every visited key is promoted under one exclusive acquisition when the walk or lookup ends, on every exit path. The contiguous-chain hit metrics are recorded by the matcher under `EnableMetrics` for successful matches rather than direct `Lookup` calls; `score_tokens` `blocks_found` and `block_hit_ratio` report the longest chain one candidate holds, which is as far as a walk reads.

AMD EPYC 9655P node, 16 exclusive CPUs, test binaries built once, six paired rounds in AB/BA order, n=6, per request:

| Row | main | #2669 | this PR |
|---|---|---|---|
| `ScoreTokens`, 240K tokens, 96 pods | 1,185.5ms / 12.2 GiB | 68.9ms / 90.7 MiB | 6.11ms / 1.6 MiB |
| `ScoreTokens`, multi-rank shape (1,034 blocks, 40 endpoints, eight rank entries each) | 551.1ms / 4.6 GiB | 85.6ms / 111 MiB | 1.56ms / 418 KiB |
| precise producer `Produce`, serial | 152.2ms / 163 MiB | 94.4ms / 143 MiB | 1.56ms / 430 KiB |
| precise producer `Produce`, 16-way | 37.3ms | 22.5ms | 230us |
| precise producer `Produce`, eight writers paced at 4,000 adds/s | not run | 101.6ms | 2.02ms |

Against the head of the closed #2402 (`bc40b33b`) in the same paired rounds: `ScoreTokens` at 96 pods -9.8%, multi-rank shape -36.0%; matcher on the multi-rank fixture -58.2% serial and -33.6% 16-way; producer -36.6% serial, -29.4% 16-way, -25.6% with eight paced writers at equal achieved load (3,958 vs 3,952 adds/s). All p=0.002. At an 8,000 adds/s offered rate this PR's writers reach 4% less load than #2402's, because the request-key LRU is one lock where #2402 shards it; recorded as a follow-up. Protocol and raw files: `test/precise-perf-run1/AUDIT-RESULTS-2026-09-03.md` on the fork.

Part of #2670

**Release note**:
```release-note
The precise prefix matcher walks the in-memory KV-block index in place (`kvblock.KeyWalker`) instead of materializing `Lookup`. The `score_tokens` span attributes `blocks_found` and `block_hit_ratio` report the longest contiguous prefix one candidate endpoint holds, in block positions, instead of the number of distinct requested keys any candidate holds; `lookup_hits_total` and `max_pod_hit_count_total` record the longest contiguous prefix once per successful match under `EnableMetrics`. Direct `Lookup` calls and cancelled matches do not increment these counters. Index reads refresh key recency once per read instead of once per key.

`lookup_latency_seconds` measures the full `Lookup` or `WalkKeys` call, including walk callbacks. The matcher span uses `llm_d.kv_cache.prefix_match.longest_chain` for contiguous prefix length; `llm_d.kv_cache.prefix_match.keys_found` is not emitted.

An omitted or non-positive `inMemoryConfig.podCacheSize` selects 10 pod entries per key, so a configuration containing only `size` supports admissions.
```

Test plan:
- [x] in-memory clears finish under a cancelled context and preserve other pods
- [x] recovered walk callback panics leave the key usable for subsequent writes and walks
- [x] size-only in-memory configuration admits entries and enforces the default per-key capacity
- [x] walk contract: positions in order, misses, duplicate positions, early stop, cancellation before, during, and on short walks, ordinal stability across `Clear` and `Evict`, decorator truthfulness, concurrent readers and writers under `-race`
- [x] retention: a key that is only read stays resident; unvisited tails and keys after a cancellation checkpoint are not promoted
- [x] parity: walked and materialized paths equal the replaced algorithms on 500 random fixtures with random rank groups
- [x] hit metrics recorded once per match on both feeders and not at all when metrics are off; nil index config
- [x] cold state: request allocation independent of pod and tier ordinal history
- [x] `make presubmit`


vMaroon (approved):
Reviewed and tested `9ce2fa7ef60e7848d847df6c97e9b3851950bb03`. The `Clear` cancellation and callback panic-unlock concerns are addressed. No remaining merge-blocking findings from this review.

Local verification passed:

```sh
go test -race -count=1 ./pkg/kvcache/... ./pkg/kvevents/... \
  ./pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache \
  ./pkg/epp/framework/plugins/scheduling/scorer/preciseprefixcache
```

- Regression coverage verifies that cancelled clears remove the target pod across tiers while preserving other pods, recovered callback panics permit subsequent writes and walks, and size-only configuration enforces the default per-key capacity.
- Hermetic EPP integration tests passed with `-race`: endpoint-score metadata and full-duplex requests/responses in standard mode and standalone modes with and without CRDs.
- An additional review-only integration test sent 29 tokenized requests over gRPC to a running EPP and delivered synthetic KV events over real ZeroMQ. Scores and selected endpoints matched expectations after stores, cache-group-specific removals, middle-block eviction, CPU-tier stores/removals, endpoint clears, and rebuilding after clear. Tracing confirmed the ordered-walk path was exercised.

Runtime validation used synthetic backend fixtures and events. It did not include Envoy forwarding to live vLLM/GPU servers, model generation, or events emitted by a live inference engine.

PR #2692, opened 2026-09-04
feat: add modality-aware endpoint filtering and multimodal API support to EPP

OwenCorrigan76:
## Summary

Add modality-filter plugin to the EPP that selects endpoints by their  llm-d.ai/model-arch label based on request path. Routes TTS requests to autoregressive-tts or omni-llm pods, STT to encoder-decoder-stt pods, and image generation to diffusion pods. Also extends OpenAI parser to recognize audio and image API endpoints and skip JSON parsing on binary responses.

## Changes

- pkg/common/paths.go: Shared path constants for multimodal APIs (future proofing)
- pkg/common/multimodal.go: PathToModelArch mapping, architecture labels
- pkg/epp/framework/plugins/scheduling/filter/modality/: New modality-filter 
  plugin with test cases covering TTS, STT, image, and unknown paths
- cmd/epp/runner/runner.go: Register modality-filter plugin
- pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go: 
  Add audioTranscriptionsAPI, inferenceAPI claims; skip JSON parsing 
  for audio/* and image/* responses
- pkg/epp/handlers/response.go: isBinaryResponseEndpoint() to gate 
  response parsing for WAV files

## Design Notes

The filter passes unknown paths through unfiltered (e.g., /v1/chat/completions). 
This allows LLM and inference requests to select any endpoint while TTS/STT/image requests are constrained to compatible architectures. 

Query strings are stripped before path matching. For example: the filter strips the ?model=tts-1&voice=alloy part, leaving just /v1/audio/speech to look up in PathToModelArch

## Testing

- Unit tests for modality-filter: unknown paths, query strings, empty endpoints, 
  audio/speech, audio/transcriptions, images/generations
- Unit tests for OpenAI parser: new API type claims, binary response handling
- Tested against cluster with labeled endpoints (llm-d.ai/model-arch)

Linked to RFC: https://github.com/llm-d/llm-d/pull/2240


pkg/common/paths.go
```diff
@@ -0,0 +1,30 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+// API request paths shared by the coordinator, EPP, and sidecar.
+const (
+	ChatCompletionsPath     = "/v1/chat/completions"
```
ahg-g: @zetxqx where should we define these constants so that the parser plugins also consume them.

pkg/common/multimodal.go
```diff
@@ -0,0 +1,39 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+const (
+	// ModelArchLabel is the pod label indicating the model architecture category.
+	ModelArchLabel = "llm-d.ai/model-arch"
```
ahg-g: if these constants are only consumed by the filter plugin, I recommend to define them in the plugin pkg.

ahg-g (commented):
Thanks!

pkg/common/multimodal.go
zetxqx: +1

pkg/common/multimodal.go
```diff
@@ -0,0 +1,39 @@
...
+	// EnvoyPathHeader is the pseudo-header Envoy uses to pass the request path.
+	EnvoyPathHeader = ":path"
+)
+
+// PathToModelArch maps API request paths to compatible model architecture labels.
+var PathToModelArch = map[string][]string{
+	AudioSpeechPath:         {ModelArchOmniLLM, ModelArchAutoRegressTTS},
+	AudioTranscriptionsPath: {ModelArchEncoderDecSTT},
+	ImagesGenerationsPath:   {ModelArchDiffusion},
+}
```
zetxqx: instead of using the path to modelArch mapping, have considered using the parsed request format to determine the label?  https://github.com/llm-d/llm-d-router/blob/24e1090694433e456f00522e78c96bd649e3b087/pkg/epp/framework/interface/requesthandling/types.go#L94-L114

in the modality filter, we can do something like this

```
func modalityArchs(body *fwkrh.InferenceRequestBody) ([]string, bool) {
      switch {
      case body == nil:
          return nil, false
      case body.TextToSpeech != nil:
          return []string{ModelArchOmniLLM, ModelArchAutoRegressTTS}, true
      case body.Images != nil:
          return []string{ModelArchDiffusion}, true
      // needs a new Transcriptions field, see below
      default:
          return nil, false // text request, pass all endpoints through
      }
  }
```

then paths.go is not needed at all.  the benefit is that we don't need to care about the path at all in the filter plugin.

pkg/common/paths.go
zetxqx: See comment https://github.com/llm-d/llm-d-router/pull/2692/changes#r3973470210 . I think if we use the inferenceRequestBody switchCase to determine what label we want to filter. we don't need to define it here at all.

pkg/epp/handlers/response.go
```diff
@@ -75,7 +76,7 @@ func (s *StreamingServer) HandleResponseBody(ctx context.Context, reqCtx *Reques
 	parser, err := s.getOrResolveParser(ctx, reqCtx)
 	if err != nil {
 		logger.Error(err, "parsing response: failed to resolve parser")
-	} else {
+	} else if !isBinaryResponseEndpoint(reqCtx) {
```
zetxqx: Which backend did you see sending binary audio with a missing or wrong Content-Type? If the backend sets Content-Type: audio/* or application/octet-stream, the parser already handles the binary body and extracts usage from response headers, so this gate is not needed for that case. And when ParseResponse does fail on a binary body, the handler only logs the error and the response still reaches the client, so what this gate prevents is log noise and wasted parsing, not a broken response.

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
```diff
@@ -116,6 +118,8 @@ func (p *OpenAIParser) Claims() fwkrh.Claims {
 			imagesGenerationsAPI,
 			imagesEditsAPI,
 			audioSpeechAPI,
+			audioTranscriptionsAPI,
+			inferenceAPI,
```
zetxqx: what is this endpoint? is this a custom endpoint?

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
```diff
@@ -430,6 +440,10 @@ func extractRequestBody(apiType string, rawBody []byte) (*fwkrh.InferenceRequest
 			return nil, validationErr
 		}
 		return &fwkrh.InferenceRequestBody{Images: &images}, nil
+
+	case audioTranscriptionsAPI, inferenceAPI:
+		return &fwkrh.InferenceRequestBody{}, nil
```
zetxqx: can we add a new type for the audioTranscriptionAPI inside InferenceRequestBody to populate it in the parser?

pkg/common/paths.go
OwenCorrigan76: Hey guys. Thanks a million for the prompt and consise review.
File fully removed based on suggestion.

pkg/common/multimodal.go
OwenCorrigan76: llm-d.ai/model-arch moved from pkg/common/multimodal.go into the modality package itself and modality.go deleted.

pkg/common/multimodal.go
OwenCorrigan76: Suggestion implemented and paths.go deleted

pkg/epp/handlers/response.go
OwenCorrigan76: Yes that comment makes sense. `isBinaryResponseEndpoint` removed.

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
OwenCorrigan76: It is not a custom endpoint. It was a leftover from earlier work. Aplolgies for the sloppiness. Removed.

pkg/epp/framework/plugins/requesthandling/parsers/openai/openai.go
OwenCorrigan76: This is done. Could you confirm if it is done the way you recommend? Thanks

zetxqx: thanks! looks good to me for the audioTranscriptionsAPI now

pkg/epp/framework/plugins/scheduling/filter/modality/modality.go
```diff
@@ -0,0 +1,114 @@
...
+// ModelArchLabel is the pod label indicating the model architecture category.
+const ModelArchLabel = "llm-d.ai/model-arch"
+
+// Model architecture values.
+const (
+	ModelArchOmniLLM        = "omni-llm"
+	ModelArchDiffusion      = "diffusion"
+	ModelArchAutoRegressTTS = "autoregressive-tts"
+	ModelArchEncoderDecSTT  = "encoder-decoder-stt"
+	ModelArchAutoRegressLLM = "autoregressive-llm"
```
ahg-g, line 39: pls add a short description of each and examples

ahg-g:
@DeanKelly751 pls let us know if this impacts your changes or not, if not I suggest to close

OwenCorrigan76:
Thanks maintainers. Closing as unresolved as the direction of using hetrogeneous InferencePools is not desired at this time and the `modality-filter` becomes obsolete.

Issue #1109, opened 2026-05-14
Long-context requests see >1s TTFT overhead in EPP request path


yankay:
**What happened**:

Long-context OpenAI-compatible streaming completion requests (~220 KB prompt) can incur **more than 1s time-to-first-byte (TTFT)** while traversing the llm-d-router **Endpoint Picker (EPP)** request path, even when the backend model server returns almost instantly.

With `llm-d-inference-sim` as the backend, throughput plateaus around ~128-132 req/s while TTFT keeps growing with concurrency, which points to queueing / CPU pressure inside EPP rather than backend prefill latency.

Reproduction harness:

- Branch/path: `https://github.com/yankay/llm-d-router/tree/bench/bench/pure-router`
- Raw text results: `bench/pure-router/results/sweep.txt`, `bench/pure-router/results/c150-steady-30s.txt`, `bench/pure-router/results/stable-1s-summary.txt`, `bench/pure-router/results/epp-metrics-steady.txt`

Observed TTFT vs concurrency on a KIND host (220 KB prompt, streaming `/v1/completions`):

| Concurrency | TTFT mean | TTFT P50 | TTFT P99 | Throughput |
|---:|---:|---:|---:|---:|
| 20 | 153.84 ms | 152.62 ms | 221.73 ms | 121.74 req/s |
| 60 | 426.15 ms | 442.48 ms | 496.03 ms | 132.32 req/s |
| 100 | 724.27 ms | 764.72 ms | 815.23 ms | 129.44 req/s |
| 150 | 1095.71 ms | 1144.54 ms | 1239.49 ms | 128.71 req/s |
| 200 | 1452.77 ms | 1516.71 ms | 1635.23 ms | 129.16 req/s |

A steady c=150 run produced mean TTFT `1151.02 ms`, P50 `1167.11 ms`, P99 `1277.72 ms`, and `127.96 req/s` throughput.

The stable gate also reproduced the issue:

```text
env CONCURRENCY=1200 WARMUP_REQS=300 ROUND_REQS=1500 ROUNDS=3 THRESHOLD_MS=1100 ./stable-1s.sh
per_round_ttft_mean_ms: 4701.59 4327.51 4555.61
median_ttft_mean_ms: 4555.61
PASS: median TTFT mean 4555.61 ms >= threshold 1100 ms
```

EPP metrics from the c=150 steady run show scheduling is not the bottleneck:

```text
inference_objective_request_duration_seconds:        1087.58 ms/request
[...]

PR #2699, opened 2026-09-04
bench: exercise 220 KiB prompts through EPP

0z5a:
**What type of PR is this?**
/kind test
/kind bug

The existing long-prompt tokenizer benchmark used a short paragraph, so it did not reproduce the 220 KiB request size reported in #1109. This change builds an exact 225,280-byte UTF-8 prompt, reports prompt bytes alongside token counts, and lets package unit tests run without requiring a live EPP.

The dev KV-cache profile hard-codes TinyLlama even when `MODEL_NAME` selects a different model. This change substitutes the selected model into the token-producer configuration and configures the render-only frontend to accept up to 131,072 tokens. The benchmark setup command also gives the simulator enough context and KV-cache capacity for the long request.

Refs #1109

**Validation**:

- [x] Exact-size and UTF-8 regression test: `go test -race ./test/profiling/tokenizerbench/`
- [x] EPP, sidecar, and framework unit packages with `-race`
- [x] `make presubmit`
- [x] `hack/verify-manifests.sh` with `kubectl-validate v0.0.4`
- [x] CPU-only kind E2E with vLLM render `v0.21.0`, inference simulator `v0.10.2`, and the KV-cache EPP profile:

```bash
E2E_PORT=30109 MODEL_NAME=TinyLlama/TinyLlama-1.1B-Chat-v1.0 \
  go test -run='^$' \
  -bench='Benchmark(Completion|ChatCompletion)_LongPrompt$' \
  -benchmem -count=3 -benchtime=3x -timeout=10m \
  ./test/profiling/tokenizerbench/
```

```text
BenchmarkCompletion_LongPrompt-152  3  332489536 ns/op  225280 prompt_bytes  37652 prompt_tokens  4696821 B/op  276 allocs/op
BenchmarkCompletion_LongPrompt-152  3  359006520 ns/op  225280 prompt_bytes  37652 prompt_tokens  4232208 B/op  269 allocs/op
BenchmarkCompletion_LongPrompt-152  3  334592885 ns/op  225280 prompt_bytes  37652 prompt_tokens  4780218 B/op  278 allocs/op
BenchmarkChatCompletion_LongPrompt-152  3  401234262 ns/op  225280 prompt_bytes  37657 prompt_tokens  5700608 B/op  298 allocs/op
BenchmarkChatCompletion_LongPrompt-152  3  422133210 ns/op  225280 prompt_bytes  37657 prompt_tokens  5784000 B/op  300 allocs/op
BenchmarkChatCompletion_LongPrompt-152  3  417800320 ns/op  225280 prompt_bytes  37657 prompt_tokens  5619421 B/op  301 allocs/op
PASS
```

The benchmark window produced 36 vLLM render requests, all with HTTP 200. The EPP log contained zero `failed to prepare per request data` entries, confirming the requests used token-producer instead of its fail-open path.

```release-note
NONE
```

### DIFF
--- a/test/profiling/tokenizerbench/benchmark_test.go
+++ b/test/profiling/tokenizerbench/benchmark_test.go
@@ -28,12 +28,15 @@ limitations under the License.
 //
 // Or manually:
 //
-//	KV_CACHE_ENABLED=true make env-dev-kind
+//	KV_CACHE_ENABLED=true \
+//	VLLM_EXTRA_ARGS_D='--max-model-len=131072 --kv-cache-size=8192' \
+//	make env-dev-kind
 //	MODEL_NAME="TinyLlama/TinyLlama-1.1B-Chat-v1.0" go test -bench=. -benchmem -count=5 -timeout=5m ./test/profiling/tokenizerbench/
 package tokenizerbench
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"net/http"
 	"os"
@@ -52,6 +55,11 @@ var (
 )
 
 func TestMain(m *testing.M) {
+	flag.Parse()
+	if flag.Lookup("test.bench").Value.String() == "" {
+		os.Exit(m.Run())
+	}
+
 	port := envOrDefault("E2E_PORT", "30080")
 	baseURL = fmt.Sprintf("http://localhost:%s/v1", port)
 	model = envOrDefault("MODEL_NAME", "TinyLlama/TinyLlama-1.1B-Chat-v1.0")
@@ -95,7 +103,9 @@ func waitForReady(timeout time.Duration) error {
 
 const shortPrompt = "What is the capital of France?"
 
-const longPrompt = `You are an expert food critic. You have been asked to review the following restaurant.
+const longPromptBytes = 220 * 1024
+
+const longPromptParagraph = `You are an expert food critic. You have been asked to review the following restaurant.
 Please provide a detailed analysis of the food quality, service, ambiance, and overall experience.
 Consider the menu variety, ingredient freshness, presentation, and value for money.
 Your review should be comprehensive and cover all aspects of the dining experience.
@@ -103,6 +113,8 @@ The restaurant is located in downtown San Francisco and specializes in modern Ca
 with influences from Japanese and Mediterranean cooking traditions. The chef has over 20 years
 of experience working in Michelin-starred restaurants across Europe and Asia. The menu changes
 seasonally to reflect the freshest local ingredients available from nearby farms and fisheries.`
+
+var longPrompt = repeatPromptToSize(longPromptParagraph, longPromptBytes)
 
 var sharedPrefixSuffixes = []string{
 	" What appetizer would you recommend?",
@@ -193,7 +205,7 @@ func benchCompletion(b *testing.B, prompt string) {
 		b.Fatalf("warmup: %v", err)
 	}
 	promptTokens := resp.Usage.PromptTokens
-	b.Logf("prompt_tokens=%d prompt_chars=%d", promptTokens, len(prompt))
+	b.Logf("prompt_tokens=%d prompt_bytes=%d", promptTokens, len(prompt))
 
 	b.ResetTimer()
 	for i := range b.N {
@@ -202,6 +214,7 @@ func benchCompletion(b *testing.B, prompt string) {
 		}
 	}
 	b.ReportMetric(float64(promptTokens), "prompt_tokens")
+	b.ReportMetric(float64(len(prompt)), "prompt_bytes")
 }
 
 func benchChatCompletion(b *testing.B, prompt string) {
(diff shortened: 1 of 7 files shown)


elevran:
@0z5a please fix failing checks on this PR

0z5a:
> @0z5a please fix failing checks on this PR

@elevran done. all passed and PTAL.

elevran (commented):
This CI artifact-retention change (raising retention-days from 1 to 7 in ci-coordinator.yaml and ci-pr-checks.yaml) is unrelated to the benchmark work in this PR and isn't explained in the description or commit message. Please explain why it is needed and if it needs to land together with the benchmark change.

Two minor optional suggestions, not blocking:
- vllm-render.yaml: the comment "This service only tokenizes requests. It does not execute model positions." is unclear; consider clarifying it means the service does not run inference.
- benchmark_test.go's repeatPromptToSize truncates by byte offset, which could split a UTF-8 rune if the source paragraph ever contains non-ASCII characters. Worth a one-line comment noting the ASCII assumption.

0z5a:
Thanks for the review. Both points are addressed in 099d559f.

**CI artifact retention**: you are right that it does not belong here. I dropped the `retention-days` change, so `ci-coordinator.yaml` and `ci-pr-checks.yaml` are now exactly as they are on main. The motivation was to keep the E2E images around for delayed job reruns, and it will go in a separate change if it is still wanted.

**vllm-render.yaml**: the comment now says what the service does rather than what it does not:

```yaml
# The render service tokenizes the prompt without running inference, so the model length
# limit of the serving engine does not apply to it.
```

That also makes the reason for `VLLM_ALLOW_LONG_MAX_MODEL_LEN=1` explicit.

**repeatPromptToSize**: added the ASCII note you suggested. The function truncates at a byte offset, so a non-ASCII paragraph could be split mid-rune; `longPromptParagraph` is ASCII, and the existing `utf8.ValidString(longPrompt)` assertion in `benchmark_unit_test.go` already guards that invariant. No logic change.

`go test ./test/profiling/tokenizerbench/` passes and `gofmt`/`go vet` are clean on the touched files.

0z5a:
Correction to my previous comment: the fixes are prepared but not yet on the branch.

I pushed them as a commit built through the API, which CI rejected with `Check signed commits in PR` ("Unsigned commits detected"). That check passes on the current head, so I reverted the branch to `a6407d42` rather than leave an unsigned commit in the PR. The PR is back to its 3 commits, 9 files and both `Check signed commits in PR` and `dco-check` are green.

The changes themselves are ready and verified locally:

- `ci-coordinator.yaml`, `ci-pr-checks.yaml`: artifact retention restored to 1, so both files match main again
- `vllm-render.yaml`: comment now states the render service tokenizes without running inference, and why the model length limit does not apply
- `benchmark_test.go`: `repeatPromptToSize` documents the ASCII assumption behind its byte-offset truncation

I will push them as a signed commit shortly.

elevran:
> I will push them as a signed commit shortly.

@0z5a waiting on commits

0z5a:
> > I will push them as a signed commit shortly.
> 
> 
> 
> @0z5a waiting on commits

Ohhh, sorry for missing!

test/profiling/tokenizerbench/benchmark_test.go
```diff
@@ -227,6 +240,18 @@ func benchChatCompletion(b *testing.B, prompt string) {
 		}
 	}
 	b.ReportMetric(float64(promptTokens), "prompt_tokens")
+	b.ReportMetric(float64(len(prompt)), "prompt_bytes")
+}
+
+// repeatPromptToSize repeats paragraph until it can be truncated to exactly size bytes.
+// Truncation happens at a byte offset, which can split a UTF-8 rune unless paragraph is
+// ASCII, as longPromptParagraph is.
```
elevran, line 248:
```suggestion
// ASCII.
```
nit: 
the trailing "as longPromptParagraph is" is editorial. The byte-splice-ASCII-precondition note is enough on its own.

test/profiling/tokenizerbench/benchmark_test.go
```diff
@@ -227,6 +240,18 @@ func benchChatCompletion(b *testing.B, prompt string) {
...
+func repeatPromptToSize(paragraph string, size int) string {
+	if size <= 0 {
```
elevran, line 250: Guard empty paragraph to avoid divide-by-zero. No current caller hits this, but the boundary should be defended.

```suggestion
	if size <= 0 || len(paragraph) == 0 {
```

elevran (commented):
PR closes the gap between the existing tokenizer benchmark and the 220 KiB reproduction in #1109. `TestLongPromptMatchesIssueScale` is a good CI gate.

Two follow-ups:

1. `BenchmarkCompletion_SharedPrefix` (line 150 in the pre-PR file) still logs `prefix_chars`. The PR renames `prompt_chars` to `prompt_bytes` in the other two benchmarks; same rationale applies here.

2. The Makefile now tells users to set `VLLM_EXTRA_ARGS_D="--max-model-len=131072 --kv-cache-size=8192"`. That value is substituted as a single YAML list element in `deploy/environments/dev/epd/patch-decode.yaml:26`, so the simulator receives one argv containing a space-separated string. CI passes so the simulator handles it; is that intentional, or should `VLLM_EXTRA_ARGS_D` become a list so each flag is its own argv?

0z5a:
Updated the shared-prefix log to bytes and guarded the empty paragraph case (tokenizerbench tests pass). The simulator flags are intentionally supplied as a space-separated value: kind-dev-env.sh splits them into separate YAML args after envsubst, so each flag reaches the container as its own argv. Thanks for the review.

elevran (approved):

PR #2701, opened 2026-09-04
Add discovery-backed vLLM render balancing

### LINKED ISSUE #2696: token-producer: address render across discovered engine endpoints instead of one URL
`token-producer` takes a single `vllm.url`, so render capacity is bounded by what one Service can expose, which is one endpoint per pod regardless of how many render-capable processes the fleet actually runs.

The wide-EP precise cell makes the gap concrete. Two prefill pods and one decode pod, DP8 per role, so 24 vLLM API servers, every one of them able to serve `/v1/*/render`. The guide's render Service selects the prefill pods and forwards to the rank-0 port, so it reaches 2 of those 24. A Service cannot do better: it has one `targetPort`, and a named port resolves per pod, so each pod contributes exactly one endpoint. Selecting decode as well would make it 3. Ranks 1 through 7 are unreachable through any single-URL Service.

This is the ceiling #2025 describes from the other side. That issue asks for observability and sizing guidance because "whatever the replica count, the ceiling exists" - a fixed pool moves the cliff without removing it. Addressing more endpoints is the other half: capacity that scales with the fleet rather than with a replica count someone has to predict.

Wonder if we could reuse the discovery the precise producer already runs. `precise-prefix-cache-producer` builds a `kvevents.SubscriberManager` whose `podDiscoveryConfig` watches pods by label selector and connects to a per-rank port derived by convention. Render needs the same two things: the set of engine pods, and a port per rank. Pointing the tokenizer at discovered endpoints instead of a Service name would take the reference cell from 2 render processes to 24 without deploying anything extra, and would keep working as the fleet scales.

Design points worth settling before code:

- **Which ports.** ~~For decode the engine ports are not the pool's `targetPorts`.~~ **Corrected:** they are usable. The pool's decode endpoint is the routing sidecar (`--port=8000 --vllm-port=8200` in the wide-EP guide), whose catch-all (`mux.Handle("/", s.decoderProxy)` in `pkg/sidecar/proxy`) reverse-proxies unregistered paths to the local engine, so a render sent to the pool port is served. I originally cited `pkg/coordinator/server/passthrough.go`, which forwards to the gateway instead, but that belongs to `cmd/coordinator`, a different binary. So the inference port is a workable default even on a disaggregated pool. An explicit port mapping is still needed where render is on a separate listener, where the sidecar terminates TLS since discovery addresses endpoints over HTTP, or where a sidecar does not forward render at all.
- **Readiness.** A Service only sends to Ready endpoints. Discovery-based addressing has to reproduce that, or render calls land on pods still loading weights.
- **Balancing.** Render cost scales with prompt length, so least-outstanding beats round-robin on mixed traffic. With per-connection balancing through a Service today this is implicit; addressing endpoints directly makes it a choice we own.
- **Failure handling.** One slow or dead endpoint should cost one retry elsewhere, not the full `vllm.timeout`. This is where the fail-fast idea in #2025 fits naturally.
- **Keep the single-URL mode.** The sidecar, a dedicated `vllm launch render` pool, and SGLang backends that do not serve the render endpoints all still need it, so this should be an additional addressing mode rather than a replacement.
- **Connection pooling.** The render transport is HTTP/1.1 with `MaxIdleConnsPerHost` 16. Fanning out across many endpoints changes what that bound means per endpoint.

Related: llm-d/llm-d#2439 and llm-d/llm-d#2440, where the same guide's render Service was returning 404 because vLLM now gates its scale-out endpoints, which silently disabled precise routing with requests still returning 200. Same class of failure, in that the tokenization path degrades without saying so.

cc @vMaroon

vMaroon:
**What type of PR is this?**

/kind feature

`token-producer` can select vLLM render targets from endpoints published by data-layer discovery instead of sending every render request through one configured URL.

- Adds `vllm.endpointDiscovery`, mutually exclusive with `vllm.url`.
- Tracks endpoint additions, updates, and deletions through the data-layer endpoint notification source.
- Selects render targets through a pluggable load-balancer interface. `round-robin` is the default and currently supported algorithm.
- Keeps the existing `vllm.url` mode available for sidecars and dedicated render Services.
- Supports optional model-limit discovery and label-based minimum renderer capacity filtering.
- Supports `vllm.prefillOnly` to reduce the render-only output budget without changing the inference payload. Requests with explicit prompt truncation retain their original budget.

Kubernetes discovery supplies Ready `InferencePool` endpoints and their target ports. Other discovery plugins can supply explicit render addresses and ports.

Fixes #2696

```release-note
Set `vllm.endpointDiscovery: {}` on `token-producer` to render against discovered endpoints with round-robin balancing. This option is mutually exclusive with `vllm.url` and uses inference ports by default. Under `vllm.endpointDiscovery`, optional `portRules` map pod labels to render ports; nonempty rule lists exclude unmatched endpoints. Optional `attemptTimeout` limits each render attempt. Transient failures permit one retry on another endpoint within the request deadline. Discovered render URLs use HTTP; use `vllm.url` for HTTPS.

Optional `discoverModelLimits`, `minModelLen`, and `contextLimitLabel` settings under `vllm.endpointDiscovery` filter render targets by context capacity. Set `vllm.prefillOnly: true` to use a one-token render-only output budget without changing the inference payload. Requests with explicit prompt truncation retain their original output budget. These settings are opt-in. Producer failures are logged and routing continues; inference validation remains with vLLM.
```

### DIFF
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery.go
@@ -37,6 +37,7 @@ type endpointDiscoveryConfig struct {
 	LoadBalancer *loadBalancerConfig `json:"loadBalancer,omitempty"`
 }
 
+// loadBalancerType returns the configured algorithm or the default round-robin algorithm.
 func (c *endpointDiscoveryConfig) loadBalancerType() string {
 	if c == nil || c.LoadBalancer == nil || c.LoadBalancer.Type == "" {
 		return roundRobinLoadBalancerType
@@ -44,26 +45,32 @@ func (c *endpointDiscoveryConfig) loadBalancerType() string {
 	return c.LoadBalancer.Type
 }
 
+// renderEndpointPicker resolves the base URL for one render request.
 type renderEndpointPicker interface {
 	Pick() (string, error)
 }
 
+// fixedEndpointPicker always returns the statically configured render URL.
 type fixedEndpointPicker string
 
+// Pick returns the static render URL.
 func (p fixedEndpointPicker) Pick() (string, error) {
 	return string(p), nil
 }
 
+// endpointLoadBalancer selects from the current discovered endpoint snapshot.
 type endpointLoadBalancer interface {
 	Pick(endpoints []string) (string, error)
 }
 
 type endpointLoadBalancerFactory func() endpointLoadBalancer
 
+// endpointLoadBalancerFactories maps configuration names to algorithm constructors.
 var endpointLoadBalancerFactories = map[string]endpointLoadBalancerFactory{
 	roundRobinLoadBalancerType: func() endpointLoadBalancer { return &roundRobinLoadBalancer{} },
 }
 
+// newEndpointLoadBalancer constructs the configured selection algorithm.
 func newEndpointLoadBalancer(loadBalancerType string) (endpointLoadBalancer, error) {
 	factory, ok := endpointLoadBalancerFactories[loadBalancerType]
 	if !ok {
@@ -72,11 +79,13 @@ func newEndpointLoadBalancer(loadBalancerType string) (endpointLoadBalancer, err
 	return factory(), nil
 }
 
+// roundRobinLoadBalancer rotates requests through the endpoint snapshot.
 type roundRobinLoadBalancer struct {
 	mu   sync.Mutex
 	next int
 }
 
+// Pick returns the next endpoint and advances the concurrency-safe cursor.
 func (b *roundRobinLoadBalancer) Pick(endpoints []string) (string, error) {
 	if len(endpoints) == 0 {
 		return "", errors.New("no vLLM render endpoints discovered")
@@ -92,6 +101,7 @@ func (b *roundRobinLoadBalancer) Pick(endpoints []string) (string, error) {
 	return endpoint, nil
 }
 
+// discoveredEndpointPicker maintains render URLs keyed by endpoint identity.
 type discoveredEndpointPicker struct {
 	mu           sync.RWMutex
 	endpoints    map[string]string
@@ -99,6 +109,7 @@ type discoveredEndpointPicker struct {
 	loadBalancer endpointLoadBalancer
 }
 
+// newDiscoveredEndpointPicker constructs a picker with the configured algorithm.
 func newDiscoveredEndpointPicker(loadBalancerType string) (*discoveredEndpointPicker, error) {
 	loadBalancer, err := newEndpointLoadBalancer(loadBalancerType)
 	if err != nil {
@@ -110,12 +121,14 @@ func newDiscoveredEndpointPicker(loadBalancerType string) (*discoveredEndpointPi
 	}, nil
 }
 
+// Pick selects a render URL from a stable endpoint snapshot.
 func (p *discoveredEndpointPicker) Pick() (string, error) {
 	p.mu.RLock()
 	defer p.mu.RUnlock()
 	return p.loadBalancer.Pick(p.ordered)
 }
 
+// Upsert validates endpoint metadata and stores its HTTP render URL by identity.
 func (p *discoveredEndpointPicker) Upsert(meta *fwkdl.EndpointMetadata) error {
 	if meta == nil {
 		return errors.New("discovered endpoint metadata is nil")
@@ -138,6 +151,7 @@ func (p *discoveredEndpointPicker) Upsert(meta *fwkdl.EndpointMetadata) error {
 	return nil
 }
 
+// Delete removes an endpoint from render selection by identity.
 func (p *discoveredEndpointPicker) Delete(meta *fwkdl.EndpointMetadata) {
 	if meta == nil || meta.ID.Name == "" {
 		return
@@ -149,6 +163,7 @@ func (p *discoveredEndpointPicker) Delete(meta *fwkdl.EndpointMetadata) {
 	p.rebuildOrdered()
 }
 
+// rebuildOrdered sorts by endpoint identity so event arrival order does not affect selection.
 func (p *discoveredEndpointPicker) rebuildOrdered() {
 	ids := make([]string, 0, len(p.endpoints))
 	for id := range p.endpoints {
--- a/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery.go
+++ b/pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery.go
@@ -1,0 +1,164 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tokenizer
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"sort"
+	"strconv"
+	"sync"
+
+	fwkdl "github.com/llm-d/llm-d-router/pkg/epp/framework/interface/datalayer"
+)
+
+const roundRobinLoadBalancerType = "round-robin"
+
+type loadBalancerConfig struct {
+	Type string `json:"type,omitempty"`
+}
+
+type endpointDiscoveryConfig struct {
+	LoadBalancer *loadBalancerConfig `json:"loadBalancer,omitempty"`
+}
+
+func (c *endpointDiscoveryConfig) loadBalancerType() string {
+	if c == nil || c.LoadBalancer == nil || c.LoadBalancer.Type == "" {
+		return roundRobinLoadBalancerType
+	}
+	return c.LoadBalancer.Type
+}
+
+type renderEndpointPicker interface {
+	Pick() (string, error)
+}
+
+type fixedEndpointPicker string
+
+func (p fixedEndpointPicker) Pick() (string, error) {
+	return string(p), nil
+}
+
+type endpointLoadBalancer interface {
+	Pick(endpoints []string) (string, error)
+}
+
+type endpointLoadBalancerFactory func() endpointLoadBalancer
+
+var endpointLoadBalancerFactories = map[string]endpointLoadBalancerFactory{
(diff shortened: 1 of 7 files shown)


nilig (dismissed):
LGTM

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery.go
```diff
@@ -0,0 +1,333 @@
...
+type fixedEndpointPicker string
+
+// Pick returns the static render URL.
+func (p fixedEndpointPicker) Pick() (string, error) {
+	return string(p), nil
+}
+
+// endpointLoadBalancer selects a non-excluded URL from the full endpoint snapshot.
+type endpointLoadBalancer interface {
+	Pick(endpoints []string, excluded map[string]struct{}) (string, error)
```
elevran, line 88: nit: not for this PR, but worth a follow-up issue. 

Round-robin ignores that render cost varies with prompt length, and a load-aware algorithm (least-outstanding, power-of-two-choices) is cheaper to add here than it looks: Runtime.NewEndpoint hands the same fwkdl.Endpoint instance to both the metrics collector and the EndpointEvent dispatched to Extract, so the live RunningRequestsSize/WaitingQueueSize on that endpoint is already reachable from endpointDiscoveryHandler.Extract -- it's just discarded down to a bare URL string in discoveredEndpointPicker.endpoints today.

A future change could keep the fwkdl.Endpoint (or its *Metrics) alongside the URL and change Pick's signature to take those richer records instead of []string, letting a new factory entry read GetMetrics() at pick time without touching discovery/retry plumbing at all.

elevran (commented):
Left one non-blocking suggestion below on the load-balancer seam so it can be later extended to support selection informed by additional information such as metrics.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery.go
elevran: let me know if you want to merge as is or with the revised interface.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery.go
```diff
@@ -0,0 +1,333 @@
...
+		return fmt.Errorf("discovered endpoint %s has an empty address", meta.ID)
+	}
+	port, err := p.resolveRenderPort(meta)
+	if err != nil {
+		return err
+	}
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.endpoints[meta.ID.String()] = "http://" + net.JoinHostPort(meta.Address, port)
```
nilig, line 202: The scheme is fixed to `http` here, but `newVLLMHTTPRenderer` still accepts `caCertPath`, `clientCertPath`, `clientKeyPath`, and `insecureSkipVerify` together with `endpointDiscovery`, and the TLS config they build never engages on an `http` URL. A TLS render port, which is the sidecar's `--secure-proxy` default, then fails every render at runtime with no startup signal, so routing quietly runs without tokens. Wonder if we could reject those fields when `endpointDiscovery` is set, the same way `url` is rejected, until discovery carries a `scheme`. https discovery is its own piece of work since pod-IP URLs need IP SANs in the server cert.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
```diff
@@ -437,11 +454,15 @@ func (r *vllmHTTPRenderer) postJSON(ctx context.Context, path string, body any,
 	if err != nil {
 		return fmt.Errorf("marshal request: %w", err)
 	}
+	baseURL, err := r.endpointPicker.Pick()
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
vMaroon: Endpoint-selection errors include `pick render endpoint` context and preserve the underlying error through `%w`. Regression coverage checks both.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery.go
vMaroon: Reused `hasTLS()` from Wen's #2664 in ab51e92d to reject nonempty certificate paths or `insecureSkipVerify: true` with `endpointDiscovery` at startup. Added regression coverage and documented the restriction. The existing HTTPS, custom CA, and mTLS tests pass. HTTPS discovery remains a separate follow-up.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery.go
vMaroon: I think we should keep round-robin and the current interface for this PR. We can carry `fwkdl.Endpoint` alongside the render URL when we add cost-aware selection. Nilig's measurements support evaluating render cost in that follow-up.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery.go
```diff
@@ -0,0 +1,179 @@
...
+		loadBalancer: loadBalancer,
+	}, nil
+}
+
+// Pick selects a render URL from a stable endpoint snapshot.
+func (p *discoveredEndpointPicker) Pick() (string, error) {
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+	return p.loadBalancer.Pick(p.ordered)
+}
```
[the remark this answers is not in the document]
vMaroon: The picker copies the endpoint snapshot under the read lock and releases it before calling the balancer. A regression test verifies the lock is available during selection.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery.go
```diff
@@ -0,0 +1,375 @@
...
+
+// TypedName returns the per-instance extractor identity used by the data layer.
+func (h *endpointDiscoveryHandler) TypedName() plugin.TypedName {
+	return h.typedName
+}
+
+// Extract keeps the picker synchronized with endpoint lifecycle events.
+func (h *endpointDiscoveryHandler) Extract(_ context.Context, event fwkdl.EndpointEvent) error {
+	if event.Endpoint == nil || event.Endpoint.GetMetadata() == nil {
+		return nil
```
elevran: nit: 
returns nil silently here but errors on the Upsert path. Pick one consistently.
Right now the caller cannot distinguish "data layer sent garbage" from "nothing to do".

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery_test.go
```diff
@@ -0,0 +1,831 @@
...
+			}
+			return &http.Response{StatusCode: http.StatusOK, Body: body}, nil
+		})},
+	}
+	tokens, _, err := renderer.Render(context.Background(), fwkrh.PayloadMap{"prompt": "hello"})
+	require.NoError(t, err)
+	assert.Equal(t, [][]uint32{{7}}, tokens)
+	assert.Equal(t, []string{"a", "b"}, hosts)
+}
+
```
elevran, line 475: nit: 
the name says "PreservesFullTimeoutByDefault" but the assertion only proves "no retry on success" (since the first call returns 200, retry path never triggers). Either rename to match what it asserts, or add a slow-endpoint case that proves the 5s budget survives a 3s first attempt and that the retry path actually fires.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery_test.go
```diff
@@ -0,0 +1,831 @@
...
+		}
+		tokens, _, err := renderer.Render(context.Background(), fwkrh.PayloadMap{"prompt": "hello"})
+		require.NoError(t, err)
+		assert.Equal(t, [][]uint32{{7}}, tokens)
+		assert.Equal(t, 1, calls)
+	})
+}
+
+func TestVLLMHTTPRenderer_DiscoveryAttemptTimeoutConfiguration(t *testing.T) {
+	for _, value := range []string{"", "100ms", "0s", "-1s", "invalid"} {
```
elevran, line 505: nit: 
`t.Run(value, ...)` with `value == ""` produces a subtest with an empty name (`...AttemptTimeoutConfiguration/`). Use `t.Run("<empty>", ...)` or split the empty-value case out of the loop.

elevran (dismissed):
The core discovery-backed balancing work is solid and CI is green. One scope concern plus a few small nits inline.

**Scope**: #2696 open for addressing render across discovered endpoints with a pluggable balancer. 
The PR now also ships other unrelated features (model-limit discovery and `vllm.prefillOnly`).
While both are real, well-tested features they
- are also feature creep and can be broken into their own PRs
- documented in release note if you prefer keeping them in

**Earlier suggestion (just FYI)**: my previous nit about enriching the load-balancer interface with fwkdl.Endpoint metadata so a future least-outstanding algorithm could read RunningRequestsSize/WaitingQueueSize was not adopted; ok merge-as-is, so it remains a future issue. The new model-limit work uses a different signal (capacity, not queue depth), so the earlier point still stands and can be tracked by an issue.

A few small nits inline.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery.go
vMaroon: Addressed in ba8b3a1d. Missing endpoints or metadata return an error, consistent with invalid metadata on the Upsert path. Tests cover both add/update and delete notifications. The data layer logs these errors and continues; they do not reject client requests.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery_test.go
vMaroon: Added a retry case in ba8b3a1d: the first endpoint returns a 503 after three seconds, then another endpoint succeeds after one second. It verifies that both attempts share the original five-second deadline. The existing slow-success case remains to cover the half-budget regression.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/endpoint_discovery_test.go
vMaroon: Named the empty-value case `unset` in ba8b3a1d.

elevran (dismissed):
Re-reviewed with focus on whether prior feedback was addressed. All items from both review rounds are resolved:
No further concerns. Approving.

elevran (approved):
all feedback addressed in code

PR #2702, opened 2026-09-04
feat(epp): complete W3C trace propagation for OpenTelemetry export

dayakar349:
**What type of PR is this?**

/kind feature

Completes W3C trace propagation on the Endpoint Picker and P/D sidecar when OpenTelemetry tracing is enabled.
- Treat `traceparent`, `tracestate`, and `baggage` as system-owned headers so client values are not re-forwarded after `generateHeaders` injects the active span context.
- Add `InitTextMapPropagator()` and call it at EPP and P/D sidecar startup so extract/inject work when OTLP export is disabled.
- Add unit and hermetic integration tests for propagation header handling and downstream trace injection.
With `--tracing=true` and standard `OTEL_*` env vars, EPP joins upstream traces, records routing spans, injects a single updated `traceparent` for downstream workloads, and exports spans to the configured collector.

## Test plan

*  `make presubmit` with `CONTAINER_RUNTIME=podman`
*  Unit tests for W3C trace header propagation and `InitTextMapPropagator` startup
*  Hermetic integration: single downstream `traceparent` after span injection
*  Manual cluster: `--tracing=true`, OTLP export, ext_proc path, Tempo traces for `llm-d-epp`

Fixes #

```release-note
NONE
```


cmd/pd-sidecar/main.go
```diff
@@ -53,6 +53,8 @@ func main() {
 		return
 	}
 
+	tracing.InitTextMapPropagator()
+
 	// Initialize tracing conditionally using config
 	if opts.Tracing {
 		shutdown, err := tracing.InitTracing(ctx, logger, "llm-d-disagg-sidecar")
```
cgoodfred, line 60: Question: should this be behind the same `if opts.Tracing` gate as `InitTracing` or is the current unconditional `InitTextMapPropegator` what we want?

cmd/epp/runner/runner.go
```diff
@@ -241,6 +241,7 @@ func (r *Runner) Run(ctx context.Context) error {
 	setupLog.Info("Flags processed", "flags", flags)
 
 	logutil.InitLogging(&opts.ZapOptions)
+	tracing.InitTextMapPropagator()
 
 	if opts.Tracing {
 		shutdown, err := tracing.InitTracing(ctx, setupLog, "llm-d-epp")
```
cgoodfred, line 249: Question: should this be behind the same `if opts.Tracing` gate as `InitTracing` or is the current unconditional `InitTextMapPropegator` what we want?

pkg/common/observability/tracing/telemetry.go
```diff
@@ -90,7 +96,7 @@ func InitTracing(ctx context.Context, logger logr.Logger, defaultServiceName str
 
 	tracerProvider := sdktrace.NewTracerProvider(opt...)
 	otel.SetTracerProvider(tracerProvider)
-	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}))
+	InitTextMapPropagator()
```
cgoodfred, line 99: Question: in `pd-sidecar/main.go` and `epp/runner/runner.go` there is a direct call to `InitTextPropagator` then have a conditional that calls `tracing.InitTracing` which now also includes `InitTextMapPropagator`. Whats the reason for the duplication and is it necessary?

cmd/pd-sidecar/main.go
dayakar349: Unconditional is intentional. Propagation needs to work when --tracing=false so EPP can still read and forward upstream traceparent headers. Gating it behind opts.Tracing would only enable that when OTLP export is on.

pkg/common/observability/tracing/telemetry.go
dayakar349: The startup call covers the tracing-off path. InitTracing() calls it again when tracing is on, so there is duplication on that path. It is idempotent. I can remove the call from InitTracing() and keep only the startup init if you prefer.

cmd/epp/runner/runner.go
dayakar349: Unconditional is intentional. Propagation needs to work when --tracing=false so EPP can still read and forward upstream traceparent headers. Gating it behind opts.Tracing would only enable that when OTLP export is on.

pkg/epp/handlers/request_test.go
```diff
@@ -158,6 +162,50 @@ func TestHandleRequestHeaders(t *testing.T) {
 	}
 }
 
+// generateHeaders must inject W3C trace context for Envoy to forward to vLLM,
+// and must not re-forward a client-supplied traceparent after injection.
+func TestGenerateHeaders_InjectsTraceContext(t *testing.T) {
```
gyliu513, line 167: The justification for calling `InitTextMapPropagator()` unconditionally is that EPP must still read and forward an upstream `traceparent` when `--tracing=false`. That path has no test: `TestGenerateHeaders_InjectsTraceContext` installs an SDK tracer provider with `AlwaysSample`, and the hermetic test runs with `WithTracing()`. The tracing-off behaviour depends on a non-obvious detail of the global noop tracer, which returns a `nonRecordingSpan` that carries the extracted remote span context, so `Inject` re-emits the client's values unchanged. I verified locally that this holds today. A test pins it so an OpenTelemetry upgrade or a change to the tracing-off setup cannot silently drop the client's `traceparent`.

Suggest adding to `request_test.go`:

```go
// With tracing disabled the noop tracer keeps the extracted remote span context,
// so the client's trace headers are forwarded unchanged.
func TestGenerateHeaders_ForwardsTraceContextWhenTracingDisabled(t *testing.T) {
	prevTP, prevProp := otel.GetTracerProvider(), otel.GetTextMapPropagator()
	otel.SetTracerProvider(noop.NewTracerProvider())
	tracing.InitTextMapPropagator()
	t.Cleanup(func() {
		otel.SetTracerProvider(prevTP)
		otel.SetTextMapPropagator(prevProp)
	})

	const clientTraceparent = "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0c9902b7-01"
	req := &extProcPb.ProcessingRequest_RequestHeaders{RequestHeaders: &extProcPb.HttpHeaders{
		Headers: &configPb.HeaderMap{Headers: []*configPb.HeaderValue{
			{Key: "traceparent", RawValue: []byte(clientTraceparent)},
			{Key: "tracestate", RawValue: []byte("vendor=abc")},
		}},
	}}
	ctx := extractTraceContext(context.Background(), req)
	ctx, span := tracing.Tracer("test").Start(ctx, "request", trace.WithSpanKind(trace.SpanKindServer))
	defer span.End()

	server := &StreamingServer{}
	reqCtx := &RequestContext{
		TargetEndpoint: "1.2.3.4:8080",
		Request: &Request{Headers: map[string]string{"traceparent": clientTraceparent, "tracestate": "vendor=abc"}},
	}
	got := map[string]string{}
	for _, h := range server.generateHeaders(ctx, reqCtx) {
		got[strings.ToLower(h.Header.Key)] = string(h.Header.RawValue)
	}
	assert.Equal(t, clientTraceparent, got["traceparent"])
	assert.Equal(t, "vendor=abc", got["tracestate"])
}
```

This needs `go.opentelemetry.io/otel/trace/noop` added to the imports.

dayakar349: @gyliu513 Added the test as suggested. PTAL when you have a moment. Thanks.

dayakar349:
@gyliu513 Could you please take a look again?

gyliu513:
@dayakar349 lgtm, thanks for the update.

@ahg-g I think this is ready to be merged, thanks!

ahg-g (approved):

dayakar349:
@ahg-g Code review is already approved, but after rebasing onto main and workflow runs are stuck on action_required again. Could you Approve and run workflows for this new head? Thanks.

dayakar349:
@ahg-g  Some required checks show Skipped. Is there anything still needed from me before merge? Thanks.

dayakar349:
@gyliu513 Could you please help me to get this merge? Some required checks show Skipped blocking the merge.

gyliu513:
@dayakar349 I do not have permission to merge this, but I think @ahg-g will merge it soon, good luck!

PR #2707, opened 2026-09-05
Pass native vLLM render requests through without content conversion

vMaroon:
**What type of PR is this?**

/kind bug
/kind cleanup
/kind deprecation

Scope: EPP HTTP rendering and deprecated Messages compatibility. Sidecar serialization and coordinator rendering are separate follow-ups.

Native rendering forwards each request to its matching vLLM endpoint without translating content or substituting the model:

| Inference protocol | Render endpoint |
| --- | --- |
| Completions | `/v1/completions/render` |
| Chat Completions | `/v1/chat/completions/render` |
| Anthropic Messages (`native` mode) | `/v1/messages/render` |

Native requests without envelope mutations use their original JSON bytes. Routing projections provide read-only inputs to scheduling. Nested objects, arrays, Completions `prompt`, and Messages `system` stay opaque in the mutable envelope. Model rewrites and late routing metadata therefore do not round-trip content through Go values, and the envelope does not decode large prompt strings.

Native rendering does not use Messages-to-Chat conversion, Python JSON emulation, typed render reconstruction, or renderer-side model stamping. Completions token-array inputs use the renderer for truncation and other preprocessing, with a render round trip and availability dependency.

Messages compatibility:

- `vllm.messagesRenderMode: legacy` is the default during the deprecation window. It converts Messages to Chat render input using the configured `modelName` and logs a startup deprecation warning. Conversion does not guarantee token parity with inference.
- `vllm.messagesRenderMode: native` forwards Messages to `/v1/messages/render` without conversion. Invalid mode values are rejected. Neither mode falls back to the other.
- `legacy_messages.go` owns conversion, Python-style JSON formatting, compatibility wire types, configured-model handling, mode validation, and the warning. Its tests are in `legacy_messages_test.go`.
- Compatibility connects through one Messages dispatch branch, a backend flag, and constructor/configuration wiring. Removal consists of deleting the two legacy files and those hooks. Native rendering, estimation, and supplied-token handling do not depend on legacy-owned helpers.

Token production uses supplied final tokens or the selected render backend's output:

- Direct render requests pass through without producing scheduling tokens.
- Supplied final tokens, including SGLang Generate, vLLM HTTP Generate, and tokenized vLLM gRPC input, reach scheduling without rendering. Cache salt, multimodal metadata, and batch boundaries remain attached to the tokens; the wire request is unchanged.
- Completions and Chat use native render tokens. Messages uses the configured render mode. Completions token arrays require renderer preprocessing.

There is no `RawPayload`-based tokenization bypass. The producer preserves pre-populated tokens or obtains them from the selected backend.

The estimate backend is separate. A missing selected endpoint or render failure does not trigger a mode change or estimation. Native gRPC text uses a compatibility path because it has no HTTP JSON envelope. Direct render requests retain model routing while skipping token production and response parsing.

Deployment requirements:

- Native Messages mode requires a vLLM build with [native Messages rendering](https://github.com/vllm-project/vllm/pull/45803). Legacy mode requires `/v1/chat/completions/render`.
- Startup warmup is an unauthenticated, best-effort Chat probe. It does not validate native Messages endpoint capability.
- Native render and serve deployments must accept the same effective model, including aliases and adapters. `modelName` supplies startup probes, native gRPC text compatibility, and legacy Messages conversion. Native pass-through preserves the effective inference model; it performs no adapter loading or base-model fallback.
- Chat and Messages use `max(vllm.timeout, vllm.mmTimeout)` without inspecting content.
- Exact token parity depends on matching model, tokenizer, template, processor, and parser configuration, deterministic upstream rendering, and no later prompt-affecting mutation.

Related work:

- [#2644](https://github.com/llm-d/llm-d-router/pull/2644) addresses unmutated Chat Completions requests. The native path covers all three protocols, model rewrites, and content preservation through EPP envelope mutations.
- The native path supersedes the conversion work in [#2414](https://github.com/llm-d/llm-d-router/pull/2414) and [#2511](https://github.com/llm-d/llm-d-router/pull/2511). Regression inputs cover reverse-ordered schemas and tool inputs, inline systems, billing headers, thinking, tool results, images, and template controls without reproducing vLLM's conversion rules.
- [#2701](https://github.com/llm-d/llm-d-router/pull/2701) remains separate target-selection work. Both changes touch the HTTP renderer and require reconciliation.

Separate follow-ups, including Chat Completions:

- Preserve content through NIXL, P2P, Mooncake, and shared-storage request serialization.
- Audit prefill changes to output limits: `truncate_prompt_tokens=-1` can depend on the output-token budget.
- Audit coordinator rendering independently. In the target topology, coordinator preprocessing precedes EPP; EPP preserves the request it receives.
- Validate render tokens against actual serving across those forwarding paths. The live checks exercise the parser/render backend directly and exclude the complete forwarding path.

Local benchmark:

The measured unchanged-request path improves, but model-rewritten Chat has a reproducible processing-time regression.

Parser plus render backend/client, median of six alternating rounds on Apple M4 Pro, Go 1.26.6, `GOMAXPROCS=1`. Base `4f8e57118894` versus `ef24d91c`. Benchmark coverage excludes `a56e0f99` and deprecated Messages mode. Sizes refer to synthetic prompt/system text. Both revisions receive the same incoming bytes; an in-memory transport returns eight fixed token IDs. Networking, production transport instrumentation, vLLM execution, director repackaging, sidecar/coordinator processing, and generation are excluded.

Unchanged model:

| Workload | Base us/op | `ef24d91c` us/op | Time change |
| --- | ---: | ---: | ---: |
| Completions, 1 KiB | 24.03 | 20.32 | -15.5% |
| Chat, 1 KiB | 35.97 | 28.85 | -19.8% |
| Messages, 1 KiB | 48.09 | 26.66 | -44.6% |
| Completions, 240 KiB | 3704.60 | 2809.40 | -24.2% |
| Chat, 240 KiB | 3729.78 | 3263.63 | -12.5% |
| Messages, 240 KiB | 6000.81 | 2816.70 | -53.1% |

At 240 KiB, allocated bytes/op fall 60.1% for Completions, 60.1% for Chat, and 71.5% for Messages.

With model rewriting, 240 KiB Chat increases from 3.724 ms to 4.258 ms (+14.3%, about 0.533 ms), Completions increases 2.5%, and Messages decreases 36.5%. Rewritten Chat also increases 11.8% at 24 KiB. Rewritten Completions increases 5.9% at 1 KiB and 0.6% at 24 KiB. Profiling identifies validation/compaction of preserved `json.RawMessage` values during mutated-envelope serialization as a cost; optimizing that remains separate work.

These are local processing/allocation measurements, not inference latency or production throughput. The [validation comment](https://github.com/llm-d/llm-d-router/pull/2707#issuecomment-5551350352) records the broader matrix, profiles, and test boundaries. The pipeline benchmark harness and profiles are local artifacts; the focused envelope benchmark is included in this branch.

Verification:

- [x] Native endpoint selection and byte-for-byte forwarding, including unknown fields, aliases, numeric representations, and all completion prompt shapes.
- [x] Nested content preservation through model rewrites and late metadata, including final director repackaging.
- [x] Direct render bypass, native-endpoint failures without conversion fallback, and refusal to reconstruct missing HTTP input.
- [x] Opaque scalar content, escaped-string and numeric-value preservation, and bounded allocation growth for large prompt strings.
- [x] Native HTTP adapter names survive a different configured probe model and an explicit model rewrite; direct render requests retain model routing; native gRPC text produces tokens through its compatibility path.
- [x] Real SGLang, vLLM HTTP Generate, and tokenized vLLM gRPC parsers publish supplied tokens through both render and estimate backends without renderer calls or wire mutations; salts and multimodal metadata are preserved.
- [x] Pre-populated tokens retain their identity and receive a missing cache salt with a raw wire payload.
- [x] Batched Completions token arrays reach the native renderer unchanged, and scheduling uses the returned tokens, batch boundaries, and cache salt.

- [x] Legacy defaults and explicit native selection, invalid mode rejection, and one startup deprecation warning per legacy plugin instance.
- [x] Legacy system/tool/image conversion and Python-style argument formatting, including key order, Unicode escaping, tool results, and thinking content.
- [x] Mode isolation for Chat, Completions token arrays, Generate, direct render requests, and pre-populated tokens; both Messages modes preserve wire requests and propagate renderer failures without fallback.
- [x] Deletion of the legacy files and configuration/dispatch hooks through a build overlay: native, estimate, and supplied-token tests pass, and EPP builds.

For `a56e0f99`, race tests for the tokenizer, all request parsers, and request-control director pass. `make test-integration-hermetic` and `make presubmit` pass; all five commits pass signature/DCO checks. The EPP hermetic integration run completes in 88.922 seconds. Presubmit reports zero lint issues with a worktree-diff warning and no called vulnerabilities.

Evidence for `ef24d91c`: the broader race-enabled unit/integration suite records 149 successful package runs and 2,160 top-level test/example passes; changed-path tests pass 50 shuffled repetitions, and guided envelope fuzzing covers 126,639 executions without failure. This broader evidence is scoped to `ef24d91c`; its configured skips and scope are recorded in the validation comment.

Live endpoint verification:

Ten small, serial cases pass at `ef24d91c` against a provisioned `zai-org/GLM-5.2-FP8` deployment using custom vLLM `0.27.1` builds. Stock vLLM is outside this verification. The harness uses the actual parser and render backend with a native renderer and GPU serving endpoint. Live endpoint coverage excludes `a56e0f99` and deprecated Messages mode.

- [x] Seven Chat/Completions cases match EPP-produced render IDs, direct native render IDs, and the GPU server's exact prompt IDs. Coverage includes text, batches, token-array truncation, tools, and model rewrites and rendered-request repackaging with routing metadata.
- [x] Three Messages cases match EPP/native-render IDs and serving input-token counts, including tool-use/tool-result history and envelope edits. Exact Messages serving-ID parity is unverified because the deployed API does not expose those IDs. Matching counts does not prove matching IDs.

These checks exclude a deployed EPP, sidecar, and coordinator. Renderer and server context limits differ, so no context-limit-dependent truncation parity claim is made. Model-rewrite cases resolve an alias to the base model; actual LoRA adapter parity is unverified. Multimodal, streaming, and long prompts are outside the live coverage. The harness and raw results are local artifacts, not included in this branch.

The checks use provisioned endpoints without deployment or GPU allocation changes. Ten inference requests account for at most eleven generated tokens. These results establish correctness smoke coverage; inference performance is outside the measurement scope.

Addresses [#2641](https://github.com/llm-d/llm-d-router/issues/2641). Follow-up to [#1861](https://github.com/llm-d/llm-d-router/issues/1861). Forwarding parity is outside the implemented scope.

```release-note
The `token-producer` vLLM backend renders Anthropic Messages natively via `/v1/messages/render` (requires a vLLM build with that endpoint). New setting `vllm.messagesRenderMode`: `auto` (default) detects renderer support and falls back to the deprecated Messages-to-Chat conversion when the endpoint is missing; `native` and `legacy` force a mode. Render requests keep the request's own model, so the render endpoint must accept the same model names, aliases, and adapters as inference. Chat and Messages render calls use the larger of `vllm.timeout` and `vllm.mmTimeout` (30s default) for all requests. Completions requests with token arrays require an available renderer.
```


vMaroon:
Fixed the supplied-token regression in `76070955` by deleting the three-line `RawPayload` early return. No new protocol branches or content transformations.

The bypass introduced in this PR skipped SGLang token publication and cache-salt handling for pre-populated tokens with raw payloads. vLLM HTTP Generate and tokenized gRPC already worked; the new tests cover them alongside SGLang.

The contract is:

- Direct render requests pass through without producing scheduling tokens.
- Supplied final tokens reach scheduling without rendering, independent of wire encoding.
- Completions token arrays still go through native rendering because preprocessing can change them. Scheduling uses the renderer's returned tokens.

The correction preserves cache salt, multimodal metadata, batch boundaries, existing token objects, and wire payloads.

Verification with the correction:

- [x] Real-parser regression tests for SGLang, vLLM HTTP Generate, and tokenized vLLM gRPC through render and estimate backends.
- [x] Raw-payload cache-salt handling and batched Completions renderer results.
- [x] Targeted race suites for token production, all request parsers, and request control.
- [x] Hermetic integration tests and `make presubmit`.
- [x] Signature and DCO checks for all four commits.

The SGLang and raw-payload cache-salt cases failed before the fix and passed afterward. Presubmit reported zero lint issues with a worktree-diff warning and no called vulnerabilities.

The [benchmark and live-endpoint results above](https://github.com/llm-d/llm-d-router/pull/2707#issuecomment-5551350352) remain measurements of `ef24d91c`, not `76070955`. They were not rerun for this correction. The model-rewrite serialization regression remains: 240 KiB Chat measured +14.3% local processing time (+0.533 ms) versus base. This fix does not address that cost.

Updated the description to state this pipeline and distinguish current regression checks from earlier benchmark/live evidence.

nilig (dismissed):
LGTM. nit: could we mention in the release note that Completions token-array inputs also call the renderer? This lets vLLM apply truncation and preprocessing correctly, but adds a render round trip and availability dependency for those clients.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
```diff
@@ -247,153 +233,27 @@ func (r *vllmHTTPRenderer) postCompletionsRender(ctx context.Context, body any)
...
-				return r.mmTimeout
-			}
-		}
-	}
-	return r.timeout
-}
-
-// produceTimeout returns the worst-case configured render timeout (multimodal),
-// surfaced so the data-producer executor extends its budget past the default.
+// produceTimeout permits multimodal rendering without inspecting content.
```
elevran, line 252: produceTimeout() always returns max(timeout, mmTimeout), replacing the removed content-aware chatTimeout(payload), which returned the short timeout for text-only messages and only used the long timeout when multimodal content was present. A plain text chat request now waits up to mmTimeout (30s default) before a stalled backend is detected, instead of failing fast at the text timeout (5s default).

pkg/epp/framework/plugins/requesthandling/parsers/util/json.go
```diff
@@ -22,10 +22,55 @@ import (
...
+		SkipResponseProcessing: true,
+	}, nil
+}
+
+// UnmarshalEnvelope keeps objects, arrays, and rawFields opaque so routing
+// metadata edits cannot round-trip content through Go values.
+func UnmarshalEnvelope(data []byte, rawFields ...string) (map[string]any, error) {
+	var fields map[string]json.RawMessage
+	if err := json.Unmarshal(data, &fields); err != nil {
+		if fallbackErr := Unmarshal(data, &fields); fallbackErr != nil {
```
elevran, line 50: When the primary json.Unmarshal fails but the stricter fallback Unmarshal succeeds, this still returns (nil, primary error) instead of the fallback's result. Any input where the fallback would have succeeded gets rejected as a parse error even though it decoded fine. Same pattern exists in UnmarshalMapWithRawField below (line 96).

pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go
```diff
@@ -85,9 +85,11 @@ func (p *AnthropicParser) WithName(name string) *AnthropicParser {
 
 func (p *AnthropicParser) ParseRequest(_ context.Context, body []byte, headers map[string]string) (*fwkrh.ParseResult, error) {
 	path := request.GetRequestPath(headers)
+	if strings.HasSuffix(strings.TrimRight(path, "/"), "/"+messagesAPI+"/render") {
```
elevran: This hand-rolled strings.HasSuffix(strings.TrimRight(path, "/"), ...) duplicates request.MatchPathSuffix, already used elsewhere in this parser. openai.go:136 has the same duplication. Worth using the shared helper so trailing-slash/case handling can't drift between the two.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/legacy_messages.go
```diff
@@ -0,0 +1,416 @@
...
+		return nil, fmt.Errorf("tokenization failed: %w", err)
+	}
+	return &fwkrh.TokenizedRequest{Prompts: []fwkrh.PromptTokens{{
+		TokenIDs:           tokenIDs,
+		MultiModalFeatures: convertMMFeaturesToUpstream(mmFeatures),
+	}}}, nil
+}
+
+// Pre-encoded messages and schemas preserve key order in the converted payload.
+func legacyMessagesPayload(msg *fwkrh.MessagesRequest) fwkrh.PayloadMap {
```
elevran, line 126: This marshals each message individually inside the loop instead of marshaling rr.Messages once as an array, multiplying encode/compaction cost by message count on every legacy-mode Messages request (legacy mode is the default).

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/backend.go
```diff
@@ -93,35 +92,33 @@ func (b renderBackend) warmup(ctx context.Context) {
 	logger.V(logutil.DEBUG).Info("token-producer backend warmup did not complete")
 }
 
-// warmupChat builds a single-message chat body carrying the given image URLs.
-func warmupChat(imageURLs ...string) *fwkrh.InferenceRequestBody {
-	blocks := make([]fwkrh.ContentBlock, 0, 1+len(imageURLs))
-	blocks = append(blocks, fwkrh.ContentBlock{Type: "text", Text: "warmup"})
+// warmupChat constructs a probe, not an inference request.
```
elevran: State what it is now: "warmupChat constructs a probe, not an inference request."

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/legacy_messages.go
```diff
@@ -0,0 +1,416 @@
...
+	tokenizerTypes "github.com/llm-d/llm-d-router/pkg/kvcache/tokenization/types"
+)
+
+const (
+	messagesRenderModeLegacy = "legacy"
+	messagesRenderModeNative = "native"
+)
+
+// Deprecated: use native Messages rendering. Conversion does not guarantee
+// token parity with inference.
```
elevran: The token-parity disclaimer for legacy Messages conversion is restated near-verbatim in four places: vllm_http.go:93 (field doc), here (function doc), the runtime log string a few lines below, and README.md.

elevran (commented):
Two correctness issues (timeout regression, dropped fallback-decode result), plus a duplicated path-matching helper, an avoidable per-message marshal loop, and two style items.

One more thing outside the diff so it doesn't attach as an inline comment: openai.go's tokenInputField (line 178-187) has a `body.Completions != nil` branch that looks unreachable now, since ParseRequest's new control flow (line 144 onward) routes completions/chat-completions requests through a separate `UnmarshalEnvelope` branch before `tokenInputField` is called for those types. Not a runtime bug today, but worth removing or confirming it's still reachable.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/legacy_messages.go
```diff
@@ -0,0 +1,416 @@
...
+func legacyAnthropicToolCall(b fwkrh.AnthropicContentBlock) map[string]any {
+	id := b.ID
+	if id == "" {
+		id = "call_0000000000"
+	}
+	return map[string]any{
+		"id":   id,
+		"type": "function",
+		"function": map[string]any{
+			"name":      b.Name,
```
elevran, line 230: `chatRenderRequest`, `chatMessage`, `chatContent`, `chatPart`, `chatImageURL` were deleted from `vllm_http.go` and reintroduced byte-for-byte here as `legacyChatRenderRequest` and friends.
Keep the canonical types in vllm_http.go and have this file consume them. This assumes we'd follow up with 'deleting legacy_messages.go removes them'.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/legacy_messages.go
```diff
@@ -0,0 +1,416 @@
...
+			"arguments": legacyPythonArguments(b.Input),
+		},
+	}
+}
+
+func legacyPythonArguments(raw json.RawMessage) string {
+	switch string(bytes.TrimSpace(raw)) {
```
elevran, line 238: nit:
WHY comments lost and gratuitous renames in the move. 
For example, `legacyPythonArguments` no longer explains why arguments are CPython-formatted; `legacyConvertAnthropicTools` no longer explains why `input_schema` stays as raw JSON; `legacyAppendAnthropicMessage` and `legacyAppendAnthropicToolResult` no longer explain role ordering, ...

Restore those comments. Also consider dropping the `legacy` prefix on these helpers entirely; the file boundary (*legacy*_messages.go) is already the deprecation marker, and prefixing `legacy` for symbols requires renames outside what the issue requires.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/estimate.go
```diff
@@ -36,6 +37,78 @@ import (
...
+	blockTypeText       = "text"
+	blockTypeImage      = "image"
+	blockTypeImageURL   = "image_url"
+	blockTypeThinking   = "thinking"
+	blockTypeToolUse    = "tool_use"
+	blockTypeToolResult = "tool_result"
+)
+
+const (
+	anthropicBillingHeaderPrefix = "x-anthropic-billing-header"
```
elevran, line 52: nit:
WHY comments lost in the move from tokenizer.go. `anthropicBillingHeaderPrefix` and
`defaultImageMediaType` both still apply here (vLLM strips the billing header server-side so estimation must too; the jpeg default matches vLLM's conversion). Restore the comments.

pkg/epp/framework/plugins/requesthandling/parsers/anthropic/anthropic.go
```diff
@@ -85,9 +85,11 @@ func (p *AnthropicParser) WithName(name string) *AnthropicParser {
...
+		return parserutil.ParseRenderRequest(body)
+	}
```
elevran, line 91: nit:
The new comment 'Auxiliary endpoints need neither local tokenization nor response interception' sits above a check that handles only count_tokens. The plural misrepresents the code. Replace with a count_tokens-specific rationale.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/README.md
```diff
@@ -25,12 +25,72 @@ Backend selection:
...
-  (e.g. `grpc`) can be added alongside `url` under the same `vllm` block.
+  `vllm.clientCertPath`/`vllm.clientKeyPath` for mTLS.
+
+## Deprecated Messages conversion
+
+`vllm.messagesRenderMode: legacy` converts Anthropic Messages into Chat
+Completions render input and uses the configured `modelName`. The mode logs a
+deprecation warning once per plugin instance. Conversion does not guarantee
+token parity with inference. It does not mutate the request forwarded to
+inference.
```
elevran: nit:
'It does not mutate the request forwarded to inference' negates the previous sentence and compresses two facts. Replace with: 'The forwarded request is unchanged.'

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/README.md
```diff
@@ -25,12 +25,72 @@ Backend selection:
...
+
+Set `vllm.messagesRenderMode: native` to forward Messages to
+`/v1/messages/render` without conversion. This mode requires a renderer with
+that endpoint.
+
+The compatibility implementation and tests are contained in
+`legacy_messages.go` and `legacy_messages_test.go`. Its integration points are
+the `MessagesRenderMode` configuration field, `configureLegacyMessages` in the
+plugin constructor, and the `legacyMessages` field and branch in Messages
+dispatch. Removing these files and integration points leaves native rendering
```
elevran: nit:
Future-tense framing for a current behavior. 'Removing these files and integration points leaves native rendering and token production intact' reads as a hypothetical future action. Replace with: 'Native rendering and token production do not depend on legacy_messages.go.'

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/README.md
```diff
@@ -25,12 +25,72 @@ Backend selection:
...
+HTTP requests or the HTTP JSON embedded in Vertex AI gRPC requests.
+
+Chat and Messages requests use the larger of `vllm.timeout` and
+`vllm.mmTimeout`. The renderer does not inspect content to select a timeout.
+
+Token parity requires matching render/serve model, tokenizer, template,
+processor, and parser configuration, plus deterministic upstream rendering.
+The serving path must preserve the effective request after EPP. Sidecar
+prefill/decode mutations and coordinator rendering are separate paths; this
+EPP contract does not establish parity for them.
```
elevran: Deictic reference the project style guide forbids. 'this EPP contract does not establish parity for them' should drop 'this EPP contract'; 'parity is not established for them' suffices on its own terms.

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http.go
```diff
@@ -247,153 +233,27 @@ func (r *vllmHTTPRenderer) postCompletionsRender(ctx context.Context, body any)
...
+// RenderMessages leaves Anthropic conversion to vLLM.
+func (r *vllmHTTPRenderer) RenderMessages(ctx context.Context, payload fwkrh.RequestPayload) ([]uint32, *tokenization.MultiModalFeatures, error) {
+	return r.renderConversation(ctx, messagesRenderPath, payload)
+}
+
+func (r *vllmHTTPRenderer) renderConversation(ctx context.Context, path string, payload fwkrh.RequestPayload) ([]uint32, *tokenization.MultiModalFeatures, error) {
 	var resp renderResponse
-	if err := r.postJSON(ctx, chatRenderPath, body, timeout, &resp); err != nil {
+	if err := r.postJSON(ctx, path, payload, r.produceTimeout(), &resp); err != nil {
 		return nil, nil, err
```
elevran, line 247: nit:
The new comment 'RenderChat forwards the native Chat Completions body without model substitution' paraphrases a one-line method body. Either delete or replace with a WHY (e.g., why model stamping was removed; how this differs from the prior renderer's contract).

pkg/epp/framework/plugins/requestcontrol/dataproducer/tokenizer/vllm_http_test.go
```diff
@@ -516,24 +442,6 @@ func TestVLLMHTTPRenderer_RenderPropagatesTraceContext(t *testing.T) {
...
-	assert.Equal(t, 5*time.Second, r.chatTimeout(textOnly))
-
-	multimodal := fwkrh.PayloadMap{"messages": []any{
-		json.RawMessage(`{"role":"user","content":"hi"}`),
-		json.RawMessage(`{"role":"user","content":[{"type":"image_url","image_url":{"url":"data:image/png;base64,abc"}}]}`),
-	}}
-	assert.Equal(t, 30*time.Second, r.chatTimeout(multimodal))
-}
-
 func TestVLLMHTTPRenderer_TLSInsecureSkipVerify(t *testing.T) {
```
elevran, line 519: nit:
The two-line docstring above TestVLLMHTTPRenderer_RenderSpanName was deleted. It carries the WHY of the assertion (an unnamed 'HTTP POST' span would be unsearchable in traces). Restore it.

pkg/epp/requestcontrol/director_test.go
```diff
@@ -67,6 +68,54 @@ var (
 	mockProducedDataKey = fwkplugin.NewDataKey("producedDataKey", "mock-producer")
 )
 
+func TestRepackagePreservesNativeRenderContent(t *testing.T) {
```
elevran, line 71: nit:
`TestRepackagePreservesNativeRenderContent` sets `body.Mutated = true` after `RewriteModelName` and asserts `renderBody == reqCtx.Request.RawBody`. If RewriteModelName silently no-ops, both `rewrite=false` and `rewrite=true` paths produce identical equality and the test passes. Add an assertion that the model field actually changed (e.g., parse `renderBody` and check model) before asserting repackage preserves it, so a regression where rewrite becomes a no-op fails loudly.

pkg/epp/framework/interface/requesthandling/types.go
```diff
@@ -116,6 +117,11 @@ type InferenceRequestBody struct {
 	// If the payload is unmarshaled, we can perform advanced processing (like prefix cache aware routing).
 	// If it remains as raw bytes, such processing may not be supported.
 	Payload RequestPayload `json:"-"`
+	// RawBody is the original HTTP JSON body (including JSON embedded in gRPC).
+	// Routing projections must not reconstruct content sent to a native renderer.
+	RawBody []byte `json:"-"`
```
elevran, line 123: Two copies of the request body now exist: handlers.Request.RawBody (at handlers/server.go:187) and this new InferenceRequestBody.RawBody. They are equal today only because every parser sets both from the same bytes. 

Either document the invariant on this field (e.g., 'equal to handlers.Request.RawBody when set by a parser'), or have parsers populate it from handlers.Request.RawBody directly so the equality is structural and not accidental.

elevran (commented):
additional nits below.

elevran (commented):
Solid change that fixes the JSON round-trip/reordering issue from #2641 and the missing Messages tokenization from #1861. 

One thing to **fix before merge**: the PR description and the release-note block both say `vllm.messagesRenderMode: legacy` is the default during the deprecation window, but the code (`configureLegacyMessages`, case ""/"auto") and the README default to `auto` (auto-discovery), not `legacy`. 
The release note gets published verbatim at tag time, so this would misinform operators about what they get on upgrade. Please correct the wording (or confirm auto isn't actually the intended default).

elevran (approved):
@vMaroon can merge after fixing release note in PR description

PR #2709, opened 2026-09-05
fix(epp): rotate only equal top-scoring endpoints in maxscore picker

quangh33:
**What type of PR is this?**
/kind bug

Fixes a routing imbalance in `MaxScorePicker` when equal top-scoring endpoints are accompanied by lower-scoring candidates.
Previously, `RotateScoredEndpoints` rotated the entire slice of `n` candidate endpoints before `slices.SortStableFunc`. Because stable sorting preserves the relative order of equal-scoring elements from the rotated slice, only 1 out of `n` rotation shifts separated the equal top-scoring endpoints; in all other `n - 1` shifts, the first candidate in the original list remained ahead of the second.

#### Example:
Given candidate endpoints `[pod0(50), pod1(50), pod2(0), pod3(0), pod4(0), pod5(0), pod6(0), pod7(0)]`:
- When rotation shift = 1: `[pod1(50), pod2(0), ..., pod0(50)]` $\rightarrow$ `pod1` is ahead $\rightarrow$ `pod1` wins (1 of 8 shifts).
- When rotation shift = 2, 3, 4, 5, 6, 7, or 0: the cut point falls among the score-0 pods, leaving `pod0` ahead of `pod1` in the rotated slice. `SortStableFunc` preserves this order $\rightarrow$ `pod0` wins (7 of 8 shifts).

This change:
1. Sorts candidate endpoints by score first.
2. Draws a monotonically increasing counter once per Pick() call via PickerRand.NextCounter().
3. Rotates candidates only within equal-score tiers up to maxNumOfEndpoints using `counter % groupSize`, ensuring:
   - Balanced round-robin distribution among equal top-scoring endpoints regardless of lower-scoring candidates.
   - Correct tie-breaking when a tie group spans the cutoff boundary (maxNumOfEndpoints > 1).
   - Independent rotation across multiple tie groups in the same request without groups interfering with each other's counter state.

**Test plan**:
- [x] Verified equal distribution (40 requests each out of 80) across equal top-scoring candidates in the presence of lower-scoring candidates.
- [x] Equal round-robin rotation (30 requests each out of 90) across tied candidates straddling the cutoff boundary when maxNumOfEndpoints > 1 (maxNumOfEndpoints=2).
- [x] Independent rotation across multiple tie groups of different sizes in a single request without shared counter lockup (maxNumOfEndpoints=3).
- [x] All existing picker unit tests pass.

```release-note
Fixed routing imbalance in MaxScorePicker where equal top-scoring endpoints received skewed traffic when lower-scoring candidates were present.

### DIFF
--- a/pkg/epp/framework/plugins/scheduling/picker/maxscore/picker.go
+++ b/pkg/epp/framework/plugins/scheduling/picker/maxscore/picker.go
@@ -88,10 +88,6 @@ func (p *MaxScorePicker) Pick(ctx context.Context, scoredEndpoints []*fwksched.S
 	log.FromContext(ctx).V(logutil.DEBUG).Info("Selecting endpoints from candidates sorted by max score", "max-num-of-endpoints", p.maxNumOfEndpoints,
 		"num-of-candidates", len(scoredEndpoints), "scored-endpoints", scoredEndpoints)
 
-	// RotateScoredEndpoints provides deterministic round-robin
-	// tie-breaking instead of random shuffle (see picker/common.go)
-	picker.RotateScoredEndpoints(scoredEndpoints)
-
 	slices.SortStableFunc(scoredEndpoints, func(i, j *fwksched.ScoredEndpoint) int { // highest score first
 		if i.Score > j.Score {
 			return -1
@@ -101,6 +97,21 @@ func (p *MaxScorePicker) Pick(ctx context.Context, scoredEndpoints []*fwksched.S
 		}
 		return 0
 	})
+
+	// RotateScoredEndpoints provides deterministic round-robin tie-breaking
+	// for equal-score candidates. Rotating within each equal-score tier ensures
+	// traffic is distributed evenly without distortion from lower-scoring endpoints.
+	counter := picker.PickerRand.NextCounter()
+	for start := 0; start < len(scoredEndpoints) && start < p.maxNumOfEndpoints; {
+		end := start + 1
+		for end < len(scoredEndpoints) && scoredEndpoints[end].Score == scoredEndpoints[start].Score {
+			end++
+		}
+		if end-start > 1 {
+			picker.RotateScoredEndpoints(scoredEndpoints[start:end], counter)
+		}
+		start = end
+	}
 
 	// if we have enough endpoints to return keep only the "maxNumOfEndpoints" highest scored endpoints
 	if p.maxNumOfEndpoints < len(scoredEndpoints) {
--- a/pkg/epp/framework/plugins/scheduling/picker/common.go
+++ b/pkg/epp/framework/plugins/scheduling/picker/common.go
@@ -23,6 +23,7 @@ limitations under the License.
 package picker
 
 import (
+	"math"
 	"math/rand/v2"
 	"sync"
 	"time"
@@ -44,7 +45,7 @@ type PickerParameters struct {
 type lockedRand struct {
 	mu      sync.Mutex
 	rand    *rand.Rand
-	counter int // round-robin counter for deterministic tie-breaking, reset when it reaches endpoint count
+	counter int // round-robin counter for deterministic tie-breaking
 }
 
 func newLockedRand() *lockedRand {
@@ -69,6 +70,17 @@ func (r *lockedRand) Shuffle(n int, swap func(i, j int)) {
 	r.rand.Shuffle(n, swap)
 }
 
+// NextCounter returns a monotonically increasing counter for deterministic
+// round-robin tie-breaking across requests.
+func (r *lockedRand) NextCounter() int {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	c := r.counter
+	r.counter = (r.counter + 1) % math.MaxInt
+	return c
+}
+
 // PickerRand is a thread-safe random number generator shared by all pickers to avoid seeding
 // overhead and ensure controlled randomization.
 var PickerRand = newLockedRand()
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: pkg/epp/framework/plugins/scheduling/picker/maxscore/picker.go [hunk 1]
```diff
@@ -102,6 +98,20 @@ func (p *MaxScorePicker) Pick(ctx context.Context, scoredEndpoints []*fwksched.S
...
+	for start := 0; start < len(scoredEndpoints) && start < p.maxNumOfEndpoints; {
+		end := start + 1
+		for end < len(scoredEndpoints) && scoredEndpoints[end].Score == scoredEndpoints[start].Score {
+			end++
+		}
+		if end-start > 1 {
+			picker.RotateScoredEndpoints(scoredEndpoints[start:end])
+		}
+		start = end
+	}
```
elevran, line 114: while this correctly generalizes to ties spanning the `maxNumOfEndpoints` boundary , that path isn't covered by a test, only the maxNumOfEndpoints=1 case is.
Worth adding a case with maxNumOfEndpoints>1 and a tie straddling the cutoff.

elevran (commented):
Nice catch. 
One gap: the loop generalizes to handle ties that straddle the maxNumOfEndpoints boundary when maxNumOfEndpoints>1, but there's no test for that case (only maxNumOfEndpoints=1 is tested). Worth adding before merge.

pkg/epp/framework/plugins/scheduling/picker/maxscore/picker.go
quangh33: added a test case with maxNumOfEndpoints>1 and a tie straddling the cutoff. tks!

pkg/epp/framework/plugins/scheduling/picker/maxscore/picker.go
```diff
@@ -102,6 +98,20 @@ func (p *MaxScorePicker) Pick(ctx context.Context, scoredEndpoints []*fwksched.S
...
+	// RotateScoredEndpoints provides deterministic round-robin tie-breaking
+	// for equal-score candidates. Rotating within each equal-score tier ensures
+	// traffic is distributed evenly without distortion from lower-scoring endpoints.
+	for start := 0; start < len(scoredEndpoints) && start < p.maxNumOfEndpoints; {
+		end := start + 1
+		for end < len(scoredEndpoints) && scoredEndpoints[end].Score == scoredEndpoints[start].Score {
+			end++
+		}
+		if end-start > 1 {
+			picker.RotateScoredEndpoints(scoredEndpoints[start:end])
```
elevran: `RotateScoredEndpoints` stores `counter = (counter+1) % n` back into the shared `PickerRand.counter`, using each call's own group size as `n`. When a single `Pick()` call rotates more than one tie group of size > 1, the shared counter gets refolded through a different modulus on each call and settles into a fixed point after the first request. Every call after that reuses the same shift for each group, so rotation effectively stops.

For example, using scores `[50,50,30,30,10]` with `maxNumOfEndpoints=3` (tie groups of size 2 and 3). The counter locks after round 1, and the same endpoint in each tier wins every request from then on.

This doesn't affect the tests added here (each has a single tie group of size>1), but it reintroduces the kind of skew this PR is fixing for any setup with more than one qualifying tie tier per pick.

Possible solution: draw the counter once per `Pick()` call rather than once per group, and have each group compute its own shift as `counter % groupSize`, so groups stop feeding into each other's stored state. Open to other approaches.

elevran (commented):
Thanks for adding the maxNumOfEndpoints>1 boundary test, that covers what I flagged earlier. Found one more issue tracing through the multi-group case, left inline with a root cause and a possible direction.

pkg/epp/framework/plugins/scheduling/picker/maxscore/picker.go
quangh33: thanks for pointing out a subtle math bug. I implemented your suggested approach:
- Draws a monotonically increasing counter once per Pick()
- Have each group compute its own shift as counter % groupSize

quangh33: also added a test case and updated the PR description

pkg/epp/framework/plugins/scheduling/picker/maxscore/picker.go
```diff
@@ -88,10 +88,6 @@ func (p *MaxScorePicker) Pick(ctx context.Context, scoredEndpoints []*fwksched.S
 	log.FromContext(ctx).V(logutil.DEBUG).Info("Selecting endpoints from candidates sorted by max score", "max-num-of-endpoints", p.maxNumOfEndpoints,
 		"num-of-candidates", len(scoredEndpoints), "scored-endpoints", scoredEndpoints)
 
-	// RotateScoredEndpoints provides deterministic round-robin
-	// tie-breaking instead of random shuffle (see picker/common.go)
-	picker.RotateScoredEndpoints(scoredEndpoints)
-
 	slices.SortStableFunc(scoredEndpoints, func(i, j *fwksched.ScoredEndpoint) int { // highest score first
 		if i.Score > j.Score {
 			return -1
```
elevran, line 93: nit: 
when a single `Pick()` call has multiple tie tiers of the same size, they all rotate by the same counter % n shift, so their winners move in lockstep across requests rather than independently. Not wrong for the bug this PR fixes, but worth a one-line comment noting the tradeoff since the PR description implies full independence between tiers.

elevran (commented):
Nice root-cause writeup and the new tests directly verify the fix.

One doc gap worth calling out: pkg/epp/framework/plugins/scheduling/picker/maxscore/README.md (steps 2-3) still describes the old shuffle-then-sort behavior; since this PR changes that exact mechanism and order (now sort-then-rotate-within-tiers), please update the README to match. See also the inline note on picker.go. Neither point blocks approval from my side.

elevran (approved):

pkg/epp/framework/plugins/scheduling/picker/maxscore/picker.go
elevran: approving. Will merge or reapprove (if comment is updated) tomorrow

Issue #1979, opened 2026-07-12
[RFC] Session-centric prefix-cache affinity and KV-lifecycle orchestration


vMaroon:
Tracker doc https://docs.google.com/document/d/1tM2qI_ElFaIvBy5p564OmrqhLzFzB89P-N7Onhu3UbI/edit?tab=t.0

**What would you like to be added**:

A tracking issue for making a session the addressable unit the router and engine share, so prefix-cache affinity, eviction visibility, and retention all become expressible over one identity. This is the umbrella; each step below lands as its own issue and PR, and each is independently useful.

The building blocks are two identifiers and three contracts, all computed by the router from request bytes — no tokenization, no engine hashes on the wire:

- `session_tag` — policy scope. A router-minted handle for a lineage; declared client ids (`prompt_cache_key`, `previous_response_id`, `session` headers) alias it, content seeds it where they are absent.
- `continuation_id` — the addressable node. A content-derived chain position; hash-prefix equality holds exactly when content prefixes are byte-identical, the condition for KV reuse. Recomputable from any resent prefix.
- Signals in (`application -> llm-d`): declared identity where present, derived from content where absent; external APIs untouched.
- Visibility up (`engine -> llm-d`): session labels ride block-store events, so the index learns what the engine actually holds instead of guessing.
- Directives down (`llm-d -> engine`): `retain` first, then `offload` / `move` / `prefetch`, all addressed in the same coordinates.

Roadmap:

- Stage 1 — Affinity (router-only, stock engine)
  - [ ] L1 + L2: content-derived session identity, an estimate-seeded `session -> {engine -> extent}` index, and covered-fraction affinity through the existing `prefix-cache-scorer` slot. Tracked in #<L1L2-issue>.
- Stage 2 — Truth and keeping
  - [ ] L3: fold engine-labeled KV events into the index so extents are truthful under eviction.
  - [ ] L4: retention directives keyed on session identity — keep continuing and declared sessions warm through pauses.
  - [ ] V1a (engine, vLLM): `session_tag` + `continuation_id` on the request, echoed on block-store events.
  - [ ] V2 (engine, vLLM): TTL-bounded eviction-priority input keyed by tag.
- Stage 3 — Session graph
  - [ ] V1b (engine, vLLM): removals reported in node coordinates so visibility survives recompute.
  - [ ] Session graph and program-aware prioritization over the index — retention that reads lifecycle state, co-location that reads program structure.

[...]

PR #2716, opened 2026-09-06
Expose session-managed prefix-cache lookup from KV events

vMaroon:
**What type of PR is this?**
/kind feature

Session-aware routing needs to find a request's cached prefix on compatible workers when `token-producer` supplies either real or estimated tokens. We expose `SessionCacheManager` and extend `precise-prefix-cache-producer` to query observed engine blocks through that interface, reusing the existing index and `prefix-cache-scorer`.

- The explicitly configured manager owns session identity, request association, continuation matching, and retention. It can consume the existing identity and token producers. A request without a manager stamp skips session lookup and mutation; this PR does not provide a default manager.
- The manager's request stamp travels through the request body's `session_id` and returns in vLLM KV events. The consumer delivers ordered events with their source and sequence to the manager and maintains physical residency separately. Logical paths can resolve across workers in an explicitly compatible cache namespace. Eviction removes the reported location, and disconnect or failed replay invalidates the affected source.
- Estimated associations produce affinity scores with zero exact cached-token coverage. Exact associations can supply coverage to P/D consumers. Concurrent branches, ranks, and cache groups are queried separately. Reconnection resets the replay cursor so buffered events can rebuild availability without another live request.

The engine path requires vLLM's session-addressable KV events (`vllm-project/vllm#51381`). Without `sessionManager`, the producer retains its token-based lookup path. Configuration and the integration contract are documented in the precise producer README.

Related to #1979 for session-addressable event visibility and #2524 / #2531 for integration with external session-state producers. This supplies the cache/event interface; it does not complete either issue's broader session policy work.

**Validation**:

Every `make presubmit LINT_NEW_ONLY=false` target, the full EPP unit suite with the race detector, and the hermetic integration suite passed on the head rebased onto current main. Three unit tests requiring external prediction or multimodal services skipped. The EPP suite ran two packages at a time after a fully parallel attempt was killed during KV-index stress testing.

- [x] Verify session stamping, both token backends, token-free event consumption, namespace/rank/group isolation, branch handling, and eviction.
- [x] Compare residency and token coverage against a separate reference model; exercise concurrent event updates and lookups. Ten shuffled repetitions of session, consumer, and reconnect tests passed with the race detector.
- [x] Reproduce the reconnect failure before the fix and verify replay recovery without new live events afterward.
- [x] Verify endpoint removal and replacement with blocked socket teardown, including rejection of late events and resets from the retired subscriber.
- [x] Reject Redis in session mode before connecting, preserve token-based Redis support, and verify that resetting one EPP leaves another replica's residency intact.
- [x] Verify the replay test buffer keeps serving after a requester aborts an attempt. The ROUTER send fails with a broken pipe in that case, and exiting on it left later replay requests unanswered, which failed a replay test in roughly every other race run of the package.
- [x] Exercise 62 assertions in a separate local deployment harness using Envoy, a race-instrumented EPP, two real vLLM CPU workers, and test session managers. This covered concurrent requests, event loss and duplication, unavailable replay, discovery changes, restarts, chat/SSE, cache salts, and zero rendering calls in estimate mode, including with the renderer stopped.

Live validation used `HuggingFaceTB/SmolLM2-135M-Instruct` and vLLM commit `f4eccdadefc6501fafeb1a0bf7f171ff24f984b0`. It validates routing and event-derived residency on CPU with one KV group and TP/DP of one. GPU/distributed-rank validation and independent reconciliation with engine cache-hit counters remain pending. The engine can delay publishing an administrative cache-clear event until its next execution step; the clear tests explicitly trigger that step. SSE checks cover content and routing, not chunk timing.

```release-note
Set `sessionManager` on `precise-prefix-cache-producer` to a configured external session-manager plugin to enable session-aware routing from observed vLLM cache blocks with real or estimated tokens. The option is unset by default; no default session manager is supplied. This mode requires per-endpoint vLLM KV-event discovery and session-addressable vLLM events, and is incompatible with `speculativeIndexing: true`. Estimated matches provide routing affinity without exact cached-token coverage. Session mode requires a process-local in-memory index and rejects redisConfig. External managers supply any shared session associations across EPP replicas.

Precise prefix-cache routing recovers from truncated KV-event replay buffers by clearing stale residency and rebuilding from available history, and an out-of-order replay frame resumes the replay instead of ending it. Both apply with sessionManager unset. Endpoint removal and replacement can proceed while an old session KV-event connection shuts down.
```


pkg/kvevents/zmq_subscriber.go
```diff
@@ -388,8 +413,16 @@ func (z *zmqSubscriber) requestReplay(ctx context.Context, startSeq uint64) bool
 				terminalErr = fmt.Errorf("malformed replay frame with %d frames", len(frames))
 				break
 			}
+			if attemptReplayed == 0 && seq > expectedSeq {
```
bongwoobak, line 416: This branch and the `receiveErr` change below have no `p.consumer != nil` guard, so they change the token path too. Before, a replay buffer that started past the requested sequence hit `invalidateReplay`, which cleared the pod and then dropped every live event for the cooldown, and the next attempt failed the same way, so a pod behind a truncated buffer never got indexed. Now it clears and rebuilds from the retained suffix and keeps the live stream.

That is a fix on the default path, `TestZMQSubscriber_ProactiveReplayRebuildsFromTruncatedHistory` (renamed from `...RejectsTruncatedHistory`) pins it with a token-mode harness. The release note only mentions `sessionManager`, could you add a line for this?

pkg/kvevents/subscriber_manager.go
```diff
@@ -96,6 +97,10 @@ func (sm *SubscriberManager) EnsureSubscriber(
 	// Create a context and start subscriber
 	subCtx, cancel := context.WithCancel(ctx)
 	done := make(chan struct{})
+	if sm.pool.consumer != nil && sourceEndpoint != "" {
+		// Availability retained by an index backend predates this stream.
+		sm.pool.resetForSource("", sourceEndpoint)
```
bongwoobak, line 102: With a shared Redis backend this reset is cross-replica. Replica B attaching to a pod clears the availability replica A built for it, and reconnects do the same. Since `sessionEventConsumer.locations` is per-process anyway, B can't read A's entries even when they survive. So in session mode the Redis backend buys nothing and multi-replica actively fights.

Either the README should state that session mode is in-memory, single-replica, or this reset needs to be scoped to what this process wrote. The README currently only says to use a separate Redis database.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/session.go
```diff
@@ -0,0 +1,282 @@
...
+	if xargs == nil {
+		xargs = map[string]any{}
+	}
+	mode := "incremental"
+	if lookup.FullReport {
+		mode = "full"
+	}
+	xargs["kv_cache_report_mode"] = mode
+	request.Body.MutatePayloadMap(func(payload fwkrh.PayloadMap) {
+		payload["session_id"] = lookup.Stamp
```
bongwoobak, line 278: This overwrites whatever the client sent, `TestSessionCacheStampEnvelope` pins it. On the vLLM side `_get_session_id` reads the body field first, then the session header, then `vllm_xargs.session_id`, so writing the body field also shadows a client that used the header. vLLM documents the field as "stable session identity shared by related requests", and today v1 only echoes it into block_pool events, so nothing breaks today.

How about having the manager derive `Stamp` from the incoming value when present, or at least calling the overwrite out in the README?

pkg/kvevents/subscriber_manager.go
```diff
@@ -131,11 +136,22 @@ func (sm *SubscriberManager) RemoveSubscriber(ctx context.Context, podIdentifier
...
 
+func (sm *SubscriberManager) resetConsumer(entry *subscriberEntry) {
+	if sm.pool.consumer == nil {
+		return
+	}
+	// Drain the subscriber before queuing the reset so old events cannot
+	// restore availability after a detach or a replacement subscription.
+	<-entry.done
+	sm.pool.resetForSource(entry.subscriber.podIdentifier, entry.sourceEndpoint)
+}
```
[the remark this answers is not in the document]
vMaroon: The manager no longer waits. `retire()` marks the subscriber under its queue mutex and queues the source reset right there, so the reset lands after every event the subscriber had already accepted, and later events or resets from the retired subscriber are dropped at `enqueue`. `EnsureSubscriber` and `RemoveSubscriber` return without touching socket teardown; `TestConsumerSubscriberRetirementDoesNotWaitForSocketClose` covers both paths.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/session.go
vMaroon: Called it out rather than deriving in the producer, since the stamp is the manager's to choose. README step 3 and the `Stamp` doc state that the body `session_id` is replaced and that vLLM reads the body field ahead of the header and `vllm_xargs.session_id`, with a link to the resolver. A manager that honors incoming identity reads it before stamping and can use it as `Stamp` when it can tell concurrent observations apart. The overwrite is what makes the stamp the identity the events echo, so I'd keep it unconditional.

pkg/kvevents/zmq_subscriber.go
vMaroon: Agreed, both changes are on the default path. The release note now has a line for the truncated-buffer rebuild and for resuming after an out-of-order frame, with `sessionManager` unset called out. No code change beyond that; the renamed test and `TestZMQSubscriber_ReplayResumesAfterMissingFrame` pin it in token mode. One related fix in the same push: the replay test buffer exited its goroutine when a ROUTER send hit a broken pipe after the subscriber aborted an attempt, which failed a replay test in roughly every other race run of the package. It keeps serving now.

pkg/epp/framework/interface/requestcontrol/session_cache.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"github.com/llm-d/llm-d-router/pkg/epp/framework/interface/plugin"
+	"github.com/llm-d/llm-d-router/pkg/kvevents"
+)
+
+// SessionCacheManager binds requests to externally managed sessions. Produce
+// publishes SessionCacheRequest under its own producer name. Event callbacks
+// may run concurrently with Produce. Session identity, continuation matching,
+// report policy, and association retention belong to the manager.
+type SessionCacheManager interface {
```
ahg-g, line 28: Framework interfaces define a contract between the runtime (in this case the code that executes request handling) and the plugins that implement extension points exposed by the runtime. If I understand correctly, this doesn't seem to be exactly the case here, from the runtime perspective, SessionCacheManager is just DataProducer; the EventConsumer and CacheNamespace interfaces are called by the precise token producer, another plugin, not the runtime.

pkg/epp/framework/interface/requestcontrol/session_cache.go
```diff
@@ -0,0 +1,68 @@
...
+	// stable until the source resets; group numbers are local to each source.
+	CacheNamespace(kvevents.EventSource, *int) string
+}
+
+// SessionCacheRequestDataKey has no default producer; the manager is explicit.
+var SessionCacheRequestDataKey = plugin.NewDataKey("SessionCacheRequest", "")
+
+// SessionCacheRequest carries a manager's event stamp and cache lookup for one
+// prompt. Absence or an empty stamp skips session lookup and request mutation.
+type SessionCacheRequest struct {
```
ahg-g, line 42: ditto, this and the type below don't seem to be defining a contract between the runtime and the plugins.

pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
```diff
@@ -53,6 +53,8 @@ type PrefixCacheMatchInfo struct {
 	// per tier. Speculative index entries count under SpeculativeTierKey.
 	// Nil when the producer supplies no tier data.
 	cachedBlocksByTier map[string]int
+	inputTokenCount    *int
```
ahg-g, line 56: pls document the new parameters, perhaps we call this totalTokens to align with totalBlocks which track the same thing in blocks.

pkg/epp/framework/plugins/datalayer/attribute/prefix/data_types.go
```diff
@@ -53,6 +53,8 @@ type PrefixCacheMatchInfo struct {
...
+	observedTokenCount *int
```
ahg-g, line 57: this is set by doesn't seem to be consumed anywhere in the PR.

pkg/epp/framework/plugins/requestcontrol/dataproducer/inflightload/producer.go
```diff
@@ -686,16 +686,16 @@ func addedTokensKey(endpointID, profileName string) string {
 	return endpointID + "|" + profileName + "|added"
 }
 
-// uncachedInputTokens returns the prompt tokens this endpoint must actually compute,
-// excluding any prefix already cached on it.
```
ahg-g: did we change this behavior?

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/extractor.go
```diff
@@ -54,7 +54,7 @@ func (p *Producer) Extract(ctx context.Context, event fwkdl.EndpointEvent) error
 		logger.V(logging.DEBUG).Info("Adding subscriber", "endpoint", endpointKey)
 	case fwkdl.EventDelete:
 		p.subscribersManager.RemoveSubscriber(ctx, endpointKey)
-		if meta.Address != "" {
+		if meta.Address != "" && p.sessionManager == nil {
```
ahg-g, line 57: pls document why the cache is not cleared when a session manager exists.

pkg/epp/framework/interface/requestcontrol/session_cache.go
```diff
@@ -0,0 +1,68 @@
...
+	// Stamp replaces the request body's session_id and returns in vLLM's KV
+	// events. The manager must distinguish concurrent requests whose block
+	// observations cannot be combined.
+	Stamp string
```
ahg-g, line 46: why do we call it stamp? is this a SessionIdOverride?

pkg/epp/framework/interface/requestcontrol/session_cache.go
```diff
@@ -0,0 +1,68 @@
...
+// publishes SessionCacheRequest under its own producer name. Event callbacks
+// may run concurrently with Produce. Session identity, continuation matching,
+// report policy, and association retention belong to the manager.
+type SessionCacheManager interface {
+	DataProducer
+	kvevents.EventConsumer
+	// CacheNamespace resolves configured engine-hash compatibility for a
+	// source and cache group. Empty means unconfigured. The mapping must be
+	// stable until the source resets; group numbers are local to each source.
+	CacheNamespace(kvevents.EventSource, *int) string
```
ahg-g, line 34: pls name the parameters, it is not clear what the int is supposed to represent

pkg/kvevents/consumer.go
```diff
@@ -0,0 +1,45 @@
...
+
+// EventConsumer receives decoded events before token processing or filtering.
+// Calls are ordered per source endpoint and concurrent across endpoints.
+// Implementations must not mutate batches. An error invalidates that source.
+type EventConsumer interface {
+	// ProcessEvents includes wire AllBlocksCleared events; consumers must
+	// invalidate report completeness when handling them.
+	ProcessEvents(context.Context, EventSource, EventBatch) error
+	// Reset invalidates availability and report completeness for an endpoint.
+	Reset(context.Context, string) error
```
ahg-g, line 36: pls name the parameters, one can guess that string is supposed to represent the endpoint from the comment, but better to name them.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -146,6 +166,28 @@ func PluginFactory(name string, rawParameters *json.Decoder, handle plugin.Handl
 // The kvcache indexer, KV-events pool, and any local ZMQ subscriber start
 // in background goroutines bound to ctx.
 func New(ctx context.Context, name string, config PluginConfig) (*Producer, error) {
+	if config.SessionManager != "" {
```
ahg-g, line 169: This is confusing, why would we fail to instantiate if sessionManager is set?!

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -271,9 +324,13 @@ func (p *Producer) Produces() map[plugin.DataKey]any {
...
 
-// Consumes declares the TokenizedRequest dependency from token-producer so
-// the data-layer DAG orders tokenization before this producer runs.
+// Consumes orders lookup after the configured session manager or token producer.
 func (p *Producer) Consumes() plugin.DataDependencies {
+	if p.sessionManager != nil {
+		return plugin.DataDependencies{Required: map[plugin.DataKey]any{
+			p.sessionDataKey: requestcontrol.SessionCacheRequest{},
+		}}
+	}
```
ahg-g, line 333: This is another indication that we better create a different producer, not merge it into the precise one

pkg/epp/framework/interface/requestcontrol/session_cache.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// SessionCacheRequestDataKey has no default producer; the manager is explicit.
+var SessionCacheRequestDataKey = plugin.NewDataKey("SessionCacheRequest", "")
```
ahg-g, line 38: defining a key in an interface is not a common pattern, the key is owned by a concrete implementation of the plugins, and so it is expected to be defined under the datalayer attributes.

pkg/kvcache/kvblock/engine_keys.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/cespare/xxhash/v2"
+)
+
+// EngineScope projects a logical cache namespace into a physical location.
+// Nil rank or group identifies a legacy event without that dimension.
+type EngineScope struct {
+	CacheNamespace   string
+	Endpoint         string
+	DataParallelRank *int
+	GroupIdx         *int
```
ahg-g, line 31: pls document what CacheNamespace and GroupIdx represent

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/producer.go
```diff
@@ -146,6 +166,28 @@ func PluginFactory(name string, rawParameters *json.Decoder, handle plugin.Handl
...
 // in background goroutines bound to ctx.
 func New(ctx context.Context, name string, config PluginConfig) (*Producer, error) {
+	if config.SessionManager != "" {
+		return nil, errors.New("sessionManager requires PluginFactory to resolve its dependency")
+	}
+	return newProducer(ctx, name, config, nil)
+}
+
+func newProducer(ctx context.Context, name string, config PluginConfig, manager requestcontrol.SessionCacheManager) (*Producer, error) {
+	if manager != nil {
```
ahg-g, line 176: This producer now almost behaves completely differently when the manager is set, how it is instantiated, configured, produces its data and its dependencies. I think this indicates it should be a different producer plugin.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/session.go
```diff
@@ -0,0 +1,282 @@
...
+		return false
+	}
+}
+
+func localGPUEvent(tier, locality, ownership string) bool {
+	return strings.EqualFold(tier, "gpu") && ownership == "" &&
+		(locality == "" || strings.EqualFold(locality, "local"))
+}
+
+func (p *Producer) produceFromSession(ctx context.Context, request *scheduling.InferenceRequest, endpoints []scheduling.Endpoint) error {
```
ahg-g, line 163: pls document the semantics of each of the functions in this file, especially this one as this seems like the core part of session-based matching.

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/session.go
```diff
@@ -0,0 +1,282 @@
...
+	matches, err := p.kvCacheIndexer.MatchBlockKeys(ctx, scope.Keys(hashes), sets.New(scope.Endpoint))
+	if err != nil {
+		return nil, fmt.Errorf("match session cache prefix: %w", err)
+	}
+	match := matches[scope.Endpoint]
+	matchedTokens := min(totalTokens, match.MatchedBlocks*prefix.BlockSizeTokens)
+	score := min(totalTokens, int(match.WeightedScore*float64(prefix.BlockSizeTokens)))
+	// Unit-size blocks preserve token coverage across engine block sizes.
+	info := attrprefix.NewPrefixCacheMatchInfo(score, totalTokens, 1).
+		WithCachedBlockCount(0).
```
ahg-g, line 227: So only if we have the exact flag set we actually set the cachedblockcount ? why wouldn't we have the exact flag set in the first place if the system is configured with precise?

pkg/epp/framework/plugins/requestcontrol/dataproducer/preciseprefixcache/session.go
```diff
@@ -0,0 +1,282 @@
...
+	if prefix.Exact {
+		hashes = hashes[:min(len(hashes), totalTokens/prefix.BlockSizeTokens)]
+	}
+	matches, err := p.kvCacheIndexer.MatchBlockKeys(ctx, scope.Keys(hashes), sets.New(scope.Endpoint))
+	if err != nil {
+		return nil, fmt.Errorf("match session cache prefix: %w", err)
+	}
+	match := matches[scope.Endpoint]
+	matchedTokens := min(totalTokens, match.MatchedBlocks*prefix.BlockSizeTokens)
+	score := min(totalTokens, int(match.WeightedScore*float64(prefix.BlockSizeTokens)))
```
ahg-g, line 224: why do we call this a score? isn't this matched prefix length in blocks?

pkg/epp/framework/interface/requestcontrol/session_cache.go
```diff
@@ -0,0 +1,68 @@
...
+	// events. The manager must distinguish concurrent requests whose block
+	// observations cannot be combined.
+	Stamp string
+	// FullReport requests reused blocks as well as newly stored blocks.
+	FullReport bool
+	// TotalTokens is the full prompt length in engine-token units, measured
+	// or estimated by the manager. It includes content beyond the known prefixes
+	// and is authoritative for affinity and prefill-load accounting.
+	TotalTokens int
+	// Prefixes are alternative observations, never a union of session branches.
```
ahg-g, line 53: the comment is not very informative, what does "prefixes are alternative observations" mean? alternative to what?

ahg-g (commented):
Thanks @vMaroon. One main architectural concern I have is how the relationship between SessionManager and PrecisePrefixCacheProducer is structured here.

Under our producer-consumer model, producers publish data that downstream components look up via data keys without taking direct dependencies on concrete producer instances. In this PR, however, PrecisePrefixCacheProducer couples directly to the SessionCacheManager instance—invoking its methods directly in addition to reading its produced data.

Furthermore:
- Tangled Event Processing: Both act as EventConsumers, with PrecisePrefixCacheProducer delegating event handling to SessionManager. I would expect these two plugins to remain completely decoupled and independent.
- Separation of Concerns: PrecisePrefixCacheProducer now builds and maintains its own session-based index, blurring the boundary of what belongs in the session manager versus the cache producer.

Given that the producer's behavior, dependencies, and configuration diverge significantly when a session manager is present, this suggests session-managed matching should likely be extracted into a separate, dedicated producer plugin rather than overloading the existing precise prefix cache producer.

bongwoobak:
I looked at this from the precise producer side, and I agree with splitting it out. Two things that may make the split cheaper than it looks.

The index is not really shared between the two modes. With `sessionManager` set, `Produce` never reaches the token path and the pool runs in consumer mode, so the token key space stays empty and the session path only uses `EngineScope` keys inside the same object. A separate producer can own its own in-memory index and nothing changes functionally. What it would carry over is the consumer-mode pool, the subscriber manager, `Extract`, `PreRequest` and the `PrefixCacheMatchInfo` publish. `prefix-cache-scorer` and the inflight producer already pick the producer by name, so they don't move.

The part the producer/consumer model does not cover is the event side. KV events are not request scoped, there is no attribute the manager could read them from, and I think that is why the direct `ProcessEvents` / `Reset` / `CacheNamespace` calls ended up here. But looking at #2852, the v1 manager uses `ProcessEvents` for metrics only, `CacheNamespace` returns "" and `Prefixes` is nil. So a first session producer can skip event fan-out to the manager entirely without losing anything, and `CacheNamespace` can be a config on the producer rather than a manager method. Once the manager starts learning block paths, a consumer registration on the pool at wiring time would keep the per-event coupling out of the request path. Also, the kvevents pieces here (consumer mode, adapter fields, `InputTokenCount`) are separable from the lookup, and part of the subscriber/replay changes already landed via #2824 and #2860 so a rebase touches them anyway. `ObservedTokenCount` has no reader in this PR or in #2852.

vMaroon:
pivoted PR

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionprefixcache/README.md
```diff
@@ -0,0 +1,157 @@
...
+## Residency
+
+Each producer instance serves one `cacheNamespace`. Candidate prefixes with
+a missing or different namespace are ignored. Configure pod discovery to
+select engines with compatible hash semantics; the namespace is an operator
+assertion, not something the producer can infer from a block hash. Use separate
+producer instances for incompatible engines.
+
+Residency is recorded per physical cache: endpoint, data-parallel rank, and
+KV-cache group. A block hash reported by two caches has two entries. A prefix
```
hyeongyun0916: Doesn't cross-worker matching also require the workers to share a hash seed?

pkg/epp/framework/plugins/requestcontrol/dataproducer/sessionprefixcache/README.md
```diff
@@ -0,0 +1,157 @@
...
+  newly stored ones.
+- `TotalTokens`: the full prompt length in engine-token units, measured or
+  estimated. It bounds coverage and is the prompt length the inflight-load
+  producer accounts for this request.
+- `Prefixes`: candidate prompt prefixes as ordered engine block hashes with
+  their block size and cache namespace. Each is resolved on its own;
+  candidates are never concatenated. `Exact` marks a prefix the prompt is
+  known to begin with.
+
+A session producer learns block hashes from the engine's KV events. It can
```
hyeongyun0916: With this pattern, each consumer seems to open its own per-pod ZMQ subscription and replay requests (the precise producer, each session-prefix-cache-producer instance, and the session producer). With three consumers, the engine publishes every event on three sockets, the EPP decodes each batch three times, and a reconnect or a new pod triggers three replay requests. Is that intended?

pkg/kvevents/zmq_subscriber.go
```diff
@@ -418,8 +443,16 @@ func (z *zmqSubscriber) requestReplay(ctx context.Context, startSeq uint64) bool
...
+				// A bounded replay buffer can lose removals for indexed blocks.
+				z.resetForSource(topic)
+				logger.Info("Rebuilding from retained replay history",
+					"requestedSeq", expectedSeq, "firstAvailableSeq", seq,
+					"replayEndpoint", z.replayEndpoint)
+				expectedSeq = seq
+			}
 			if seq != expectedSeq {
-				terminalErr = fmt.Errorf("incomplete replay: expected sequence %d, got %d", expectedSeq, seq)
+				receiveErr = fmt.Errorf("incomplete replay: expected sequence %d, got %d", expectedSeq, seq)
```
hyeongyun0916: This also changes precise prefix-cache routing without a session producer. When `replaySocketPort` is set, an out-of-order replay frame now retries from the next expected sequence, instead of clearing the endpoint's residency and waiting out the replay cooldown. Should the release note(PR Description) mention it?

PR #2719, opened 2026-09-06
fix(kvevents): wait for old subscriber to exit on endpoint change

quangh33:
**What type of PR is this?**
/kind bug

This fix is similar to https://github.com/llm-d/llm-d-router/pull/2717.

When `SubscriberManager.EnsureSubscriber` replaces an existing subscriber due to an endpoint change, it cancels the old subscriber's context but does not wait for its goroutine to terminate before spawning the replacement subscriber. This allows the old and new subscribers for the same pod to run concurrently, potentially interleaving events out-of-order into the worker pool.

This change waits on `entry.done` before creating and starting the replacement subscriber, and honors context cancellation during the wait.

Fixes #

**Test plan**:
- [x] Verified `TestSubscriberManager_EndpointChange_WaitsForOldSubscriberExit` confirms the old listening socket is closed before `EnsureSubscriber` returns.
- [x] Verified `TestSubscriberManager_EndpointChange_HonorsContextCancellation` cleans up state when the context is cancelled.
- [x] Verified all tests in `pkg/kvevents/...` pass with race detector enabled (`go test -v -race ./pkg/kvevents/...`).

**Release note**:
```release-note
Fix a race condition where replacing a subscriber on endpoint change could run two subscriber goroutines concurrently for the same pod.

### DIFF
--- a/pkg/kvevents/subscriber_manager.go
+++ b/pkg/kvevents/subscriber_manager.go
@@ -83,7 +83,16 @@ func (sm *SubscriberManager) EnsureSubscriber(
 			"oldReplayEndpoint", entry.replayEndpoint,
 			"newReplayEndpoint", replayEndpoint)
 		entry.cancel()
+		select {
+		case <-entry.done:
+		case <-ctx.Done():
+		}
 		delete(sm.subscribers, podIdentifier)
+		if err := ctx.Err(); err != nil {
+			metrics.SubscriberActive.Set(float64(len(sm.subscribers)))
+			cleanupSubscriberMetrics(podIdentifier, entry.done)
+			return err
+		}
 		// The replacement subscriber below reuses podIdentifier, so its series
 		// are kept rather than cleaned up.
 	}
--- a/pkg/kvevents/subscriber_manager_test.go
+++ b/pkg/kvevents/subscriber_manager_test.go
@@ -280,3 +280,151 @@ func TestSubscriberManager_Shutdown_HonorsContextCancellation(t *testing.T) {
 	identifiers, _ := sm.GetActiveSubscribers()
 	assert.Empty(t, identifiers)
 }
+
+func TestSubscriberManager_EndpointChange_WaitsForOldSubscriberExit(t *testing.T) {
+	ctx := context.Background()
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: pkg/kvevents/subscriber_manager.go [hunk 1]
```diff
@@ -83,6 +83,14 @@ func (sm *SubscriberManager) EnsureSubscriber(
 			"oldReplayEndpoint", entry.replayEndpoint,
 			"newReplayEndpoint", replayEndpoint)
 		entry.cancel()
+		select {
```
vMaroon, line 86: Thanks for adding the teardown wait and tests. There's one cancellation case we should cover: if `entry.done` and `ctx.Done()` are both ready, `select` can choose `entry.done`, create a replacement with an already-canceled context, and return `nil`.

Could we check `ctx.Err()` after the wait and apply the cancellation cleanup before creating the replacement? A test with both channels ready would cover this path.

quangh33: nice catch. i updated the code to honor context cancelation in case both entry.done() and ctx.Done() are both ready.

pkg/kvevents/subscriber_manager_test.go
```diff
@@ -280,3 +280,151 @@ func TestSubscriberManager_Shutdown_HonorsContextCancellation(t *testing.T) {
...
+	require.Eventually(t, func() bool {
+		conn, err := net.Dial("tcp", addr1)
+		if err == nil {
+			_ = conn.Close()
+			return true
+		}
+		return false
+	}, 2*time.Second, 10*time.Millisecond)
+
+	// Cancel the subscriber and wait until its socket is released, guaranteeing
```
vMaroon, line 406: nit: Rebinding the socket does not guarantee that `entry.done` is closed: socket cleanup happens inside `runSubscriber`, before `Start` returns and the outer goroutine closes `done`. Could we synchronize on `entry.done` directly in a package-internal test so the both-ready setup is deterministic?

vMaroon (approved):
Thanks for following up. Checking `ctx.Err()` after the wait handles cancellation regardless of which channel wins, and cleanup happens before a replacement can be created. This addresses my concern.

PR #2723, opened 2026-09-07
feat(coordinator): prefer mm_metadata on prefill with EC

### LINKED ISSUE #2722: Coordinator: use render mm_metadata on prefill instead of full kwargs_data
## What would you like to be added

After vLLM [#54659](https://github.com/vllm-project/vllm/pull/54659), render returns `features.mm_metadata` (metadata-only `MultiModalKwargsItem` blobs: `placeholder_metadata_fields` / `keep_on_cpu`, without encoder tensors such as `pixel_values`). Prefill may omit `kwargs_data` and send `mm_metadata` when `ec_transfer_params` is set.

Update the coordinator to:

1. Parse optional `mm_metadata` from the render response into each `MultimodalEntry`.
2. Keep encode on full `kwargs_data` (ViT still needs pixel tensors).
3. On prefill, when `ec_transfer_params` is non-empty and every entry has `MMMetadata`, emit `mm_metadata` and omit `kwargs_data`; otherwise keep sending `kwargs_data` for older renderers and non-EC paths.

```text
Encode  = kwargs_data
Prefill = mm_metadata + ec_transfer_params
no EC / old renderers = kwargs_data unchanged
```

Documented as the follow-up to the vLLM half in `docs/communication.md` (Optimization: avoid sending pixel data to prefill).

## Why is this needed

Prefill only needs lightweight multimodal metadata (for example `image_grid_thw` for mRoPE) plus EC transfer params. Forwarding full `kwargs_data` re-ships ~1 MB/image of `pixel_values` that encode already consumed. Dropping those blobs cuts coordinator-to-prefill payload size on E/P/D multimodal traffic.

## Success criteria

- [ ] Render stores `mm_metadata` when present; absent field remains backward compatible.
- [ ] Prefill with NIXL EC + metadata uses `mm_metadata` and omits `kwargs_data`.
- [ ] Prefill without EC (or without metadata) still sends `kwargs_data`.
- [ ] Encode unchanged.
- [ ] Unit tests cover the prefer-metadata and fallback paths.
- [ ] `docs/communication.md` describes the `mm_metadata` contract (not the earlier plain `image_grid_thw` sketch).

zhouyou9505:
**What type of PR is this?**
/kind feature

vLLM [#54659](https://github.com/vllm-project/vllm/pull/54659) adds render `features.mm_metadata` so prefill can drop full `kwargs_data` when embeddings arrive via `ec_transfer_params`. This wires that path in the coordinator: parse `mm_metadata` from render, keep encode on `kwargs_data`, and on prefill emit `mm_metadata` (omit `kwargs_data`) when EC params are present and every entry has metadata. Older renderers and non-EC paths keep sending `kwargs_data`.

Fixes #2722

```release-note
Coordinator prefill for native /inference/v1/generate endpoints sends multimodal mm_metadata instead of full kwargs_data when EC transfer params are available.
```

## Test plan
- [x] Unit tests for prefer-metadata prefill (generate + chat tokens.features)
- [x] Unit tests for kwargs_data fallback without EC / partial metadata
- [x] Render parses optional mm_metadata; extractMultimodalEntries parallel/null rules
- [x] `go test ./pkg/coordinator/steps/ ./pkg/coordinator/pipeline/`

### DIFF
--- a/docs/communication.md
+++ b/docs/communication.md
@@ -825,10 +825,12 @@ Prefill only needs the metadata sibling plus `ec_transfer_params` to load those
 embeddings. For large images the pixel tensors dominate payload size, so dropping
 `kwargs_data` on prefill cuts coordinator-to-prefill traffic.
 
+The multimodal data each stage sends:
+
 ```text
-Encode  = kwargs_data
-Prefill = mm_metadata + ec_transfer_params
-no EC / old renderers = kwargs_data unchanged
+Encode              = kwargs_data
+Prefill (optimized) = mm_metadata + ec_transfer_params
+Prefill (fallback)  = kwargs_data
 ```
 
 **Coordinator behavior:**
--- a/docs/communication.md
+++ b/docs/communication.md
@@ -814,37 +814,48 @@ No `features` or `ec_transfer_params` (no images); `prompt` contains the token a
 
 ### Optimization: avoid sending pixel data to prefill
 
-Currently the full `kwargs_data` blobs (containing both `pixel_values` and `image_grid_thw`) are forwarded to the prefill worker. The prefill worker only needs `image_grid_thw` for mRoPE -- the `pixel_values` are redundant since the encoder already consumed them. For large images, the pixel tensors dominate the payload size, so stripping them would significantly reduce the data sent to prefill.
+With a renderer that emits `features.mm_metadata` (see vLLM
+[#54659](https://github.com/vllm-project/vllm/pull/54659)), the coordinator can
+omit `kwargs_data` on the prefill request when `ec_transfer_params` is present.
 
-**Required changes:**
+`kwargs_data` mixes encoder tensors (`pixel_values`) with lightweight fields
+(`image_grid_thw` and other `keep_on_cpu` / placeholder-metadata fields). Encode
+already consumes the tensors and publishes embeddings through the EC connector.
+Prefill only needs the metadata sibling plus `ec_transfer_params` to load those
+embeddings. For large images the pixel tensors dominate payload size, so dropping
+`kwargs_data` on prefill cuts coordinator-to-prefill traffic.
 
-1. **vLLM render endpoint** (`vllm/entrypoints/openai/render/serving.py`): return `image_grid_thw` as a separate top-level field in the render response, alongside `kwargs_data`. The render step already computes it during image preprocessing (`get_image_grid_thw()` in the vision processor). Example response:
-   ```json
-   {
-     "token_ids": [1, 32000, 32000, 32000, ...],
-     "features": {
-       "mm_hashes": {"image": ["abc123hash", "def456hash"]},
-       "mm_placeholders": {"image": [{"offset": 1, "length": 3}, {"offset": 4, "length": 3}]},
-       "kwargs_data": {"image": ["<full-msgpack-blob-1>", "<full-msgpack-blob-2>"]},
-       "image_grid_thw": {"image": [[1, 24, 24], [1, 16, 16]]}
-     }
-   }
-   ```
+```text
+Encode  = kwargs_data
+Prefill = mm_metadata + ec_transfer_params
+no EC / old renderers = kwargs_data unchanged
+```
 
-2. **vLLM prefill worker**: accept `image_grid_thw` directly in the features dict (as plain JSON arrays) instead of extracting it from the msgpack `kwargs_data` blob.
+**Coordinator behavior:**
 
-3. **Coordinator render step** (`pkg/steps/render.go`): parse `image_grid_thw` from the render response and store it per `MultimodalEntry`.
+1. **Render step** (`pkg/coordinator/steps/render.go`): parse optional
+   `features.mm_metadata` from the render response (same per-modality item order
+   as `mm_hashes` / `kwargs_data`) and store it on each `MultimodalEntry`.
+2. **Encode step** (`pkg/coordinator/steps/encode.go`): unchanged; encode still
+   sends full `kwargs_data`.
+3. **Prefill step** (`pkg/coordinator/steps/prefill.go`): when
+   `ec_transfer_params` is non-empty and every multimodal entry has
+   `MMMetadata`, emit `mm_metadata` and omit `kwargs_data`. Otherwise keep
+   sending `kwargs_data` so older renderers and non-EC paths (for example
+   generate-path encode skip) stay compatible.
 
-4. **Coordinator prefill step** (`pkg/steps/prefill.go`): send `image_grid_thw` instead of `kwargs_data` in the prefill request features:
-   ```json
-   "features": {
-     "mm_hashes": {"image": ["abc123hash", "def456hash"]},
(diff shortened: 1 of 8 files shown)


docs/communication.md
```diff
@@ -814,37 +814,48 @@ No `features` or `ec_transfer_params` (no images); `prompt` contains the token a
...
+omit `kwargs_data` on the prefill request when `ec_transfer_params` is present.
+
+`kwargs_data` mixes encoder tensors (`pixel_values`) with lightweight fields
+(`image_grid_thw` and other `keep_on_cpu` / placeholder-metadata fields). Encode
+already consumes the tensors and publishes embeddings through the EC connector.
+Prefill only needs the metadata sibling plus `ec_transfer_params` to load those
+embeddings. For large images the pixel tensors dominate payload size, so dropping
+`kwargs_data` on prefill cuts coordinator-to-prefill traffic.
+
+```text
```
capri-xiyue, line 830: Could you add description regarding this block I think this block is for input of encode stage and prefill stage?

pkg/coordinator/steps/utils.go
```diff
@@ -112,46 +112,85 @@ func capSingleTokenOutput(body map[string]any, format gateway.RequestFormat) {
...
+}
+
+// buildPrefillMMFeatures builds features for the prefill request.
+//
+// When preferMetadata is true and every entry carries non-empty MMMetadata,
+// the result includes mm_metadata and omits kwargs_data. That path is only
+// valid together with ec_transfer_params (vLLM rejects metadata-only features
+// without EC params). Otherwise kwargs_data is included for backward
+// compatibility with renders that do not emit mm_metadata.
+func buildPrefillMMFeatures(entries []pipeline.MultimodalEntry, preferMetadata bool) map[string]any {
```
capri-xiyue, line 125: Is there any other stage also need MMFeaturesBuild, should we name it as buildMMFeatures?

zhouyou9505: Good catch on the naming. buildMMFeatures already exists in the same file and is used by the encode step — it builds features with kwargs_data only. buildPrefillMMFeatures adds the EC+metadata gating logic on top (checking preferMetadata and allEntriesHaveMMMetadata). They serve different stages with different behavior, so keeping the Prefill prefix makes the call site in prefill.go self-documenting without needing to read the function body to understand which path it takes.

docs/communication.md
zhouyou9505: ok. I adjust the description to make it clearer.

pkg/coordinator/steps/utils.go
```diff
@@ -112,46 +112,85 @@ func capSingleTokenOutput(body map[string]any, format gateway.RequestFormat) {
...
+	items := make([]any, len(values))
+	for i, v := range values {
+		if v != "" {
+			items[i] = v
 		}
 	}
 	return map[string][]any{ModalityImage: items}
 }
 
+// mmKwargsField is kept as a thin alias for tests and call sites that name the
```
roytman, line 207: **nit**: I understood the idea of a thin alias, but it has exactly one call site (`encode.go:230`), which could just call `mmBase64Field` directly. Not a correctness issue, just an unnecessary indirection introduced by this PR. It's not blocking the PR.

roytman (changes requested):
## `mm_metadata` vs `kwargs_data` is decided per-request, not per-entry

`buildPrefillMMFeatures` (and the `preferMetadata` bool it takes) makes a single yes/no decision for the whole request:

```go
// prefill.go:140
features := buildPrefillMMFeatures(reqCtx.MultimodalEntries, len(ecParams) > 0)

// utils.go:125-130
func buildPrefillMMFeatures(entries []pipeline.MultimodalEntry, preferMetadata bool) map[string]any {
	useMetadata := preferMetadata && allEntriesHaveMMMetadata(entries)
	...
}
```

`len(ecParams) > 0` only answers "did **any** image get an EC descriptor?" - it says nothing about **which** images. `allEntriesHaveMMMetadata` checks metadata presence per entry, but nothing checks that the same entry also has a corresponding key in `ecParams`.

**Concrete failure case:** a 2-image batch on `ec-nixl`. Image A's encoder returns an `ec_transfer_params` descriptor; image B's doesn't (a real, already-handled path, see `nixlEC.MergeEncodeResponse` in `pkg/coordinator/connectors/ec/nixl.go:38-46`, which logs a warning and continues rather than erroring when an encoder response carries no descriptor). Both A and B have `MMMetadata` from render. Result: `ecParams` is non-empty (from A) -> `useMetadata = true` -> `kwargs_data` is omitted **for the whole request**, including B. B now has no tensor data (kwargs_data dropped) and no transfer path (no EC entry) — silently wrong/missing output for that image, no error anywhere in this pipeline.

The encode step fans out one HTTP sub-request per multimodal entry (`encode.go`, the `g.Go` loop), and merges each response independently. This code path exists specifically to handle one encoder response coming back without a descriptor while others do; it logs a warning and moves on; it doesn't fail the request. That's a designed-for degraded case, e.g. one encoder replica hiccups, or one image type doesn't produce a NIXL descriptor for whatever reason. Nothing else in the pipeline reconciles that image's missing coverage before prefill runs.

Before this PR, `kwargs_data` was sent unconditionally for every entry so this failure mode couldn't occur.

### The fix should be per-entry, not a request-wide fallback

I checked how vLLM's side (vllm-project/vllm#54659) actually merges these fields — it's genuinely per-item, not a request-level mode switch:

```python
# vllm/entrypoints/scale_out/token_in_token_out/mm_features.py
def merge_mm_kwargs_items(kwargs_item, metadata_item):
    if kwargs_item is None:
        return metadata_item
    if metadata_item is None:
        return kwargs_item
    return MultiModalKwargsItem({**kwargs_item, **metadata_item})

def mm_kwargs_from_features(features):
    ...
    mm_kwargs[modality] = [
        merge_mm_kwargs_items(k, m)
        for k, m in zip(kwargs_items, metadata_items, strict=True)
    ]
```

vLLM zips `kwargs_data[i]` and `mm_metadata[i]` index by index and merges whichever side is non-null. A single request can carry `kwargs_data[A] = null, kwargs_data[B] = <blob>` alongside `mm_metadata[A] = <blob>, mm_metadata[B] = null` and vLLM reconstructs each item correctly. (Its own request validator, `_require_ec_for_metadata_only` in `protocol.py`, is also a whole-request boolean check — it doesn't catch this partial-coverage case either, so this really does need to be handled coordinator-side.)

So rather than falling the entire request back to `kwargs_data` when even one entry lacks EC coverage, `buildMMFeaturesOpts` could send both fields together with complementary per-index nulls: `mm_metadata[i]` for entries whose hash is in `ecParams` (and that have metadata), `kwargs_data[i]` for every other entry. That's a smaller change than it sounds - the null-elision convention (`mmBase64Field`) already exists for the cache-hit sentinel; it would just need to run for both fields simultaneously instead of picking one for the whole map.

pkg/coordinator/steps/utils.go
zhouyou9505: ok. understood.

Issue #2665, opened 2026-09-02
[Observability] Measure observability (tracing) performance overhead for llm-d-router


nicolexin:
## Parent
llm-d/llm-d#2389.

## Goal

Measure EPP's tracing overhead under load and publish a short report, so the SIG
can decide whether tracing should default on and at what sampling rate.

## Runs

Using `test/perf/` with `optimized-baseline.yaml`, same cluster and machine
family for all runs:

| Run | Config |
|---|---|
| 1 | tracing off (baseline) |
| 2 | tracing on, `samplerArg: "0.01"` |
| 3 | tracing on, `samplerArg: "0.1"` (chart default) |
| 4 | tracing on, `samplerArg: "1.0"` |

## Measure

- p50 / p95 / p99 latency and throughput at fixed offered load
- EPP CPU and memory
- CPU profile for run 4 (`--collect-pprof-profiles`), noting time spent in the
  OTel SDK and exporter path

## Deliverable

Report under `test/perf/config/studyanalysis/` following the existing format:
methodology, results table, pprof artifacts, and a recommendation on default
sampling rate.

satyamg1620:
@nicolexin , i would like to work on this issue. Can you assign it to me

nicolexin:
Hi @satyamg1620, this is already being worked on by @albertoperdomo2 and @rameshdoddaiah. They have some updates in the parent issue https://github.com/llm-d/llm-d/issues/2389.

PR #2725, opened 2026-09-07
fix(perf): honor tracing configuration and report scheduler p99

0z5a:
**What type of PR is this?**

/kind bug
/kind test

The performance runner writes its Helm overrides after the selected router configuration. Its hard-coded tracing flag disables requested tracing, and setting only that flag does not enable the Chart's OTEL environment configuration.

Resolve `router.tracing.enabled`, fall back to the legacy EPP tracing flag when unset, and render both switches consistently. Preserve the configured sampler and exporter settings and the default-off behavior. Invalid tracing values and malformed YAML fail before Helm runs.

Carry scheduler histogram p99 through calculation and report output alongside p50 and p95. Append a new table when the existing report uses the older header, retaining historical rows. The change touches only the performance runner, its README and a new offline test file.

Preparation for #2665 and llm-d/llm-d#2389. The requested tracing-overhead experiment remains open: no cluster measurement or sampling recommendation is included.

**Test plan**:

- [x] Verify tracing precedence, accepted flag values, default-off behavior, invalid-input rejection and preservation of unrelated overrides. The 20 offline tests also cover scheduler p99 interpolation, report layout/history and main-to-report wiring. All 20 passed on Python 3.10.20 and 3.12.13 with PyYAML 6.0.3, with zero skips or errors.
- [x] Inspect 18 source-Chart Helm renders generated from the actual runner overrides, covering off/1%/10%/100%, conflicting and missing settings, and the five checked-in recipes. All 18 passed with Helm v3.17.1 and a null kubeconfig. These local render checks do not validate the released OCI Chart or a live collector.

Reproduce the checked-in offline tests:

```bash
python3 -m unittest discover -s test/perf -p test_run_nightly_perf.py -v
```

Full `make presubmit` passed on the three-file snapshot based on `d8d22ea8f7d412f2a7e61ec415d11b24322a7938`. Lint reported zero issues; govulncheck found no called vulnerable symbols, while retaining its package/module-level findings. Publication commit `4fb7cc0d63eb1a08c66385d41539f71c1fe370ce` contains the exact same three validated file hashes. The repository's signature/DCO check and independent local `git verify-commit` both passed on this commit. GitHub currently reports `unknown_key` for the existing SSH signing key; account-level signing-key registration remains pending.

Validation used an isolated CPU-only builder, with no dependency changes or cluster access. No live EPP startup, exported traces, tracing-overhead measurements or full benchmark results are claimed. Development and self-checks were AI-assisted.

```release-note
The performance runner honors router tracing settings and includes scheduler histogram p99 in its reports.
```

### DIFF
--- a/test/perf/run_nightly_perf.py
+++ b/test/perf/run_nightly_perf.py
@@ -114,6 +114,22 @@ def deploy_epp(ns, chart_path, chart_version, router_config_path, epp_cpu="2", e
     
     if not os.path.exists(router_config_path):
         raise FileNotFoundError(f"Router config file not found at: {router_config_path}")
+
+    with open(router_config_path, "r") as f:
+        guide_data = yaml.safe_load(f) or {}
+    guide_router = guide_data.get("router") or {}
+    tracing = (guide_router.get("tracing") or {}).get("enabled")
+    if tracing is None:
+        guide_epp = guide_router.get("epp") or {}
+        tracing = (guide_epp.get("flags") or {}).get("tracing")
+        if tracing is None:
+            tracing = False
+        elif str(tracing) in ("1", "t", "T", "true", "TRUE", "True"):
+            tracing = True
+        elif str(tracing) in ("0", "f", "F", "false", "FALSE", "False"):
+            tracing = False
+    if not isinstance(tracing, bool):
+        raise ValueError("tracing requires a boolean router.tracing.enabled or a boolean EPP flag")
         
     release_name = os.path.splitext(os.path.basename(router_config_path))[0]
     
@@ -136,7 +152,7 @@ def deploy_epp(ns, chart_path, chart_version, router_config_path, epp_cpu="2", e
                 "flags": {
                     "v": 4,
                     "enable-pprof": "true",
-                    "tracing": "false"
+                    "tracing": str(tracing).lower()
                 },
                 "resources": {
                     "requests": {
@@ -148,6 +164,9 @@ def deploy_epp(ns, chart_path, chart_version, router_config_path, epp_cpu="2", e
                         "memory": mem_limit
                     }
                 }
+            },
+            "tracing": {
+                "enabled": tracing
             },
             "monitoring": {
                 "prometheus": {
@@ -167,18 +186,11 @@ def deploy_epp(ns, chart_path, chart_version, router_config_path, epp_cpu="2", e
         }
     }
 
-    # Extract guide specific model server labels and nullify them in overrides to avoid helm deep merge keeping them
-    try:
-        with open(router_config_path, "r") as f:
-            guide_data = yaml.safe_load(f)
-            if (guide_data and "router" in guide_data 
-                    and "modelServers" in guide_data["router"] 
-                    and "matchLabels" in guide_data["router"]["modelServers"]):
-                for key in guide_data["router"]["modelServers"]["matchLabels"].keys():
-                    if key != "app":
-                        overrides["router"]["modelServers"]["matchLabels"][key] = None
-    except Exception as e:
-        print(f"Warning: Could not parse router config to extract modelServers labels for nullification: {e}")
+    # Clear model-server labels replaced by the simulator.
+    match_labels = (guide_router.get("modelServers") or {}).get("matchLabels") or {}
+    for key in match_labels:
+        if key != "app":
+            overrides["router"]["modelServers"]["matchLabels"][key] = None
     
     if machine_family:
         overrides["router"]["epp"]["affinity"] = {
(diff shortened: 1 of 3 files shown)


test/perf/run_nightly_perf.py
```diff
@@ -114,6 +114,22 @@ def deploy_epp(ns, chart_path, chart_version, router_config_path, epp_cpu="2", e
     
     if not os.path.exists(router_config_path):
         raise FileNotFoundError(f"Router config file not found at: {router_config_path}")
+
+    with open(router_config_path, "r") as f:
+        guide_data = yaml.safe_load(f) or {}
+    guide_router = guide_data.get("router") or {}
+    tracing = (guide_router.get("tracing") or {}).get("enabled")
```
gyliu513: The tracing resolution here assumes `router.tracing` and `router.epp.flags` are always mappings:

    tracing = (guide_router.get("tracing") or {}).get("enabled")
    ...
    tracing = (guide_epp.get("flags") or {}).get("tracing")

If someone writes `router.tracing: true` instead of the correct `router.tracing.enabled: true` (or `router.epp.flags: true`), `.get("enabled")` is called on a bool, raising an unhandled `AttributeError: 'bool' object has no attribute 'get'` instead of the intended `ValueError("tracing requires a boolean ...")` a few lines below. This defeats the "invalid tracing values fail cleanly" contract the PR's own tests assert for other malformed inputs.

Suggest checking the type before calling `.get()`, so this case falls through to the existing ValueError instead of crashing with an AttributeError:

    tracing_cfg = guide_router.get("tracing")
    if tracing_cfg is not None and not isinstance(tracing_cfg, dict):
        raise ValueError("tracing requires a boolean router.tracing.enabled or a boolean EPP flag")
    tracing = (tracing_cfg or {}).get("enabled")
    if tracing is None:
        guide_epp = guide_router.get("epp") or {}
        flags_cfg = guide_epp.get("flags")
        if flags_cfg is not None and not isinstance(flags_cfg, dict):
            raise ValueError("tracing requires a boolean router.tracing.enabled or a boolean EPP flag")
        tracing = (flags_cfg or {}).get("tracing")

test/perf/run_nightly_perf.py
```diff
@@ -167,18 +186,11 @@ def deploy_epp(ns, chart_path, chart_version, router_config_path, epp_cpu="2", e
...
-            if (guide_data and "router" in guide_data 
-                    and "modelServers" in guide_data["router"] 
-                    and "matchLabels" in guide_data["router"]["modelServers"]):
-                for key in guide_data["router"]["modelServers"]["matchLabels"].keys():
-                    if key != "app":
-                        overrides["router"]["modelServers"]["matchLabels"][key] = None
-    except Exception as e:
-        print(f"Warning: Could not parse router config to extract modelServers labels for nullification: {e}")
+    # Clear model-server labels replaced by the simulator.
+    match_labels = (guide_router.get("modelServers") or {}).get("matchLabels") or {}
```
gyliu513: This refactor moved the config-file read earlier in `deploy_epp`, but in doing so also dropped the `try/except` that used to wrap the `modelServers`/`matchLabels` extraction and print a warning on parse failure:

    match_labels = (guide_router.get("modelServers") or {}).get("matchLabels") or {}

If `router.modelServers` is malformed (e.g. a list instead of a mapping), this now raises an unhandled `AttributeError: 'list' object has no attribute 'get'` and aborts the entire nightly run, whereas previously it printed "Warning: Could not parse router config to extract modelServers labels for nullification: ..." and continued.

This behavior change doesn't seem related to what the PR is actually for (tracing + p99 reporting), so unless it was intentional, suggest restoring the fallback:

    try:
        match_labels = (guide_router.get("modelServers") or {}).get("matchLabels") or {}
    except AttributeError as e:
        print(f"Warning: Could not parse router config to extract modelServers labels for nullification: {e}")
        match_labels = {}

test/perf/run_nightly_perf.py
0z5a: Thanks for catching this. Fixed in 310ab672eeec3504bac9a8929d7904743550a07a. The tracing, EPP and EPP-flags containers are checked for mapping types before calling `.get()`, including false and empty malformed values. Invalid containers raise `ValueError` before any Helm or kubectl command; null defaults and explicit tracing precedence are preserved. The regression suite passes on Python 3.10 and 3.12 (21 offline tests on each), and `make presubmit` passes.

test/perf/run_nightly_perf.py
0z5a: Fixed in 310ab672eeec3504bac9a8929d7904743550a07a. Restored warning-and-continue behavior for malformed `modelServers` or `matchLabels`, while retaining the simulator `app: llm-d-sim` override and resolved tracing setting. Regression coverage checks seven malformed label shapes, one warning, preserved overrides and unchanged input YAML.

ahg-g (approved):

PR #2726, opened 2026-09-07
fix: omit unexecuted pipeline step timings

learner0810:
**What type of PR is this?**

/kind cleanup

When `conditional-decode` completes a request, the pipeline exits early without running the remaining EPD steps. The timing log still included zero-value entries for those unexecuted steps, producing empty fields such as:

```json
{"level":"info","ts":1788752454.3713067,"logger":"handler","caller":"pipeline/pipeline.go:111","msg":"pipeline step timings","x-request-id":"c66d3013-da9b-94d2-ba40-b72ed5a2b340","parse":"91.851µs","replace-media-urls":"224.247897ms","render":"18.292683ms","conditional-decode":"277.689852ms","":"0s","":"0s","":"0s"}
```

Fixes #

```release-note
NONE
```


roytman (approved):
Thank you for the fix

